PR #1648, opened 2018-08-27
build: move unused-variable check from tslint to tsc

bajtos:
Now that we no longer need additional imports of referenced types to satisfy the compiler (see #1641), we have the opportunity to enable unused-variable checking at TypeScript level.

This pull request has two parts:

**build: move unused-variable check from tslint to tsc** (94c4440)

Disable tslint's "no-unused-variable" rule, it has been deprecated.

Enable TypeScript rule "noUnusedLocals", rework all places violating
this rule to use a different mechanism for turning this rule off.

Checking unused variables at compile time has the following benefits:

- Unused variables (locals) are highlighted directly in VS Code.
  Before, we had to run tslint to discover unused variables, since
  the VS Code extension for tslint was not able to detect unused vars.

- The output of tslint is clean again with no deprecation warnings
  cluttering the output.

**build: fix tslint issues related to promises** (fd3317e)

After I disabled no-unused-variable check, tslint started to complain about places incorrectly using promises. Some of the warnings were legitimate and I fixed the code. Other cases were not valid, I fixed them by modifying tslint config and adding comments to disable the problematic rule for few source code lines.

## Checklist

- [x] `npm test` passes on your machine
- New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `examples/*` were updated

PR #1651, opened 2018-08-28
build: fix tslint issues related to promises

bajtos:
A follow-up for https://github.com/strongloop/loopback-next/pull/1648.

> After I disabled no-unused-variable check, tslint started to complain about places incorrectly using promises. Some of the warnings were legitimate and I fixed the code. Other cases were not valid, I fixed them by modifying tslint config and adding comments to disable the problematic rule for few source code lines.

It makes me wonder why `tslint` is not reporting these promise-related problems when `no-unused-variable` is enabled, but I can't be bothered to investigate.

## Checklist

### DIFF
--- a/packages/rest/test/integration/rest.server.integration.ts
+++ b/packages/rest/test/integration/rest.server.integration.ts
@@ -45,7 +45,7 @@ describe('RestServer (integration)', () => {
 
   it('honors port binding after instantiation', async () => {
     const server = await givenAServer({rest: {port: 80}});
-    await server.bind(RestBindings.PORT).to(0);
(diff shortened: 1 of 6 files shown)


packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -45,7 +45,7 @@ describe('RestServer (integration)', () => {
 
   it('honors port binding after instantiation', async () => {
     const server = await givenAServer({rest: {port: 80}});
-    await server.bind(RestBindings.PORT).to(0);
```
hacksparrow: @bajtos why was `await` removed here and other places as well?

PR #1649, opened 2018-08-27
feat(service-proxy): add service mixin

bajtos:
Implement "ServiceMixin" for applications. This mixin enhances component
registration so that service providers exported by a component are
automatically registered for dependency injection; and adds a new sugar
API for registering service providers manually:

```ts
app.serviceProvider(MyServiceProvicer);
```

The method name "serviceProvider" was chosen deliberately to make it
clear that we are binding a Provider, not a class constructor. Compare
this to `app.repository(MyRepo)` that accepts a class construct. In the
future, we may add `app.service(MyService)` method if there is enough
user demand.

This pull request is the first step towards automated service registration via boot, see #1439.

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [x] Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

Issue #1439, opened 2018-06-19
Sugar API for registering services + automated registration via boot


bajtos:
At the moment, `@loopback/boot` is able to automatically load datasources and model repositories from project files. We should add support for loading services too.

See  #1347 for an example implementation.

## Acceptance criteria

- [x] Implement a mixin adding `app.service` API, this mixin should be provided by `@loopback/service-proxy`. Proposed implementation (see #1347):

    ```ts
    service<T>(provider: Constructor<Provider<T>>) {
      const key = `services.${provider.name.replace(/Provider$/, '')}`;
      this.bind(key).toProvider(provider);
    }
   ```

    (PR #1649, the method name is called `serviceProvider`.)

- [x] Update the CLI template to apply this new mixin to all newly created applications - PR #1649

- [x]  Add a new booter for loading services from files matching the following naming convention: `services/${name}.service.js` - PR #1652

- Update/simplify `examples/todo` to use this new functionality 
  - [x] use the new mixin - PR #1649 
  - [x] use automatic registration via boot

- [x] Document the new booter in [Booting an application](http://loopback.io/doc/en/lb4/Booting-an-Application.html)

- Update/simplify the content in the relevant [Todo Tutorial](http://loopback.io/doc/en/lb4/todo-tutorial.html) section to leverage the new booter
  - [x] use the new mixin - PR #1649 
  - [x] use automatic registration via boot

bajtos:
> Proposed implementation (see #1347):
>
> ```ts
> service<T>(provider: Constructor<Provider<T>>) {
>   const key = `services.${provider.name.replace(/Provider$/, '')}`;
>   this.bind(key).toProvider(provider);
> }
> ```

I am proposing to call the new method `serviceProvider` in order to avoid possible confusion about the type passed to the sugar API (a Provider vs. a service class), because other sugar-API methods like `controller`, `repository` and `dataSource` are accepting a class constructor.

PR #992, opened 2018-02-13
[WIP] feat(context): add `@injectable` to decorate bindable classes

raymondfeng:
Implements https://github.com/strongloop/loopback-next/issues/958

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] Related API Documentation was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `packages/example-*` were updated

PR #1652, opened 2018-08-28
feat(boot): implement Service booter

bajtos:
Implement automatic discovery and registration of Service providers - see #1439 for the full context and #1649 for the first step on that journey.

This pull request has three parts/commits, the first two are preparations making the actual implementation easier to make and troubleshoot.

**feat(testlab): expose "sandbox.path" property** (4e86b08)

Allow TestSandbox consumers to access the path of the sandbox directory.

Clean up the code by moving "validateInst" check into "path" getter,
thus avoiding the need to explicitly call "validateInst" in every
public API method.

**feat(boot): add debug logs for better troubleshooting** (f88e776)

Make it easier to troubleshoot the situation when a booter is not
recognizing artifact files and/or classes exported by those files.

To make debug logs easy to read, absolute paths are converted to
project-relative paths in debug logs. This change required a refactoring
of Booter design, where "projectRoot" becomes a required constructor
argument.

While making these changes, I changed "options" to be a required
constructor argument too, and chaged both "options" and "projectRoot"

**feat(boot): implement Service booter**

The main change.

## Checklist


packages/boot/src/booters/service.booter.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Default ArtifactOptions for DataSourceBooter.
+ */
+export const ServiceDefaults: ArtifactOptions = {
+  dirs: ['services'],
+  extensions: ['.service.js'],
+  nested: true,
+};
+
+function isServiceProvider(cls: Constructor<{}>): cls is ServiceProviderClass {
+  return /Provider$/.test(cls.name);
```
raymondfeng: I think we should check the existence of `value()`.

packages/boot/test/fixtures/service-class.artifact.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/boot
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+export class GreetingService {
+  greet(whom: string = 'world') {
+    return Promise.resolve(`Hello ${whom}`);
+  }
+}
```
raymondfeng: Are we going to bind plain service class during boot? I don't see the logic.

packages/boot/src/booters/service.booter.ts
bajtos: Sounds good, I add that check to the existing name check.

(IMO, checking just for `value` is not enough, as there may be other classes/interfaces that contain `value` method too.)

packages/boot/test/fixtures/service-class.artifact.ts
bajtos: Not yet. The logic here is to ensure that the booter is picking only Providers and excludes regular classes, there is an integration test for that.

https://github.com/strongloop/loopback-next/blob/9098ad3ade0b226e90291c0f3365b290bd508a66/packages/boot/test/integration/service.booter.integration.ts#L23-L26

Suggestions are welcome on how to make this more obvious to people reading this code.

Maybe it's a sign that I should enhance `ServiceMixin` and the booter to handle both `app.serviceProvider(GeocoderServiceProvider)` and `app.service(GreeterService)` cases, even though `app.service` may not be used by any real LB4 app? IDK.

raymondfeng: For now, let's just add a comment to `GreeterService` saying it will be excluded from discovery as we only allow providers so far.

packages/boot/src/booters/service.booter.ts
virkt25: There should be documentation stating that the Booter is expecting `Provider` to be a part of the class name.

packages/boot/src/booters/base-artifact.booter.ts
```diff
@@ -78,7 +95,25 @@ export class BaseArtifactBooter implements Booter {
...
+    debug(
+      'Discovering %s artifacts in %j using glob %j',
+      this.artifactName,
+      this.projectRoot,
+      this.glob,
+    );
+
     this.discovered = await discoverFiles(this.glob, this.projectRoot);
+
+    if (debug.enabled) {
```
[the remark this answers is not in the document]
raymondfeng: Because JSON.stringify is expensive. The parameters will be calculated before invoking debug itself.

bajtos:
I have addressed your review comments and finished updates of example apps and docs - see 36a8eac.

The pull request is ready for the final review and landing.

@raymondfeng @virkt25 PTAL again.

docs/site/Booting-an-Application.md
```diff
@@ -237,6 +237,33 @@ The `datasources` object support the following options:
...
+### Service Booter
+
+#### Description
+
+Discovers and binds Service providers using `app.serviceProvider()` (Application
+must use `ServiceMixin` from `@loopback/service-proxy`).
+
+**IMPORTANT:** For a class to be recognized by `ServiceBooter` as a service
+provider, its name must end with `Provider` suffix and its prototype must have a
+`value()` method.
```
raymondfeng: To me, this is a sign that we should have something like https://github.com/strongloop/loopback-next/pull/992 to make it possible to explicit.

docs/site/soap-calculator-tutorial-scaffolding.md
```diff
@@ -12,12 +12,14 @@ Let's start by creating the initial application by running the following
 command:
 
 ```sh
-lb4 app soap-calculator --enableRepository
+lb4 app soap-calculator --enableRepository --enableServices
```
raymondfeng: Hmm, now we have `enableRepository` vs `enableServices`. The naming is inconsistent.

docs/site/Booting-an-Application.md
virkt25: +1 for #992, it's certainly needed.

Issue #1531, opened 2018-07-15
Unable to Explore API (generated OpenAPI spec is not valid)


Vandivier:
## Description / Steps to reproduce / Feature proposal

1. `lb4 example`, then pick the Todo app.
2. install it and get it running.
3. visit http://localhost:3000/openapi.yaml
4. paste the yaml within http://editor.swagger.io

## Current Behavior

get errors:
![image](https://user-images.githubusercontent.com/5559355/42729349-11e8839a-87a3-11e8-9111-9c393e4259d1.png)

## Expected Behavior

don't get errors :)

virkt25:
Able to reproduce. Looks like we aren't generating `responses` property and a few other issues possibly from JSON to YAML conversion. Great catch!

dhmlau:
FYI - https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#response-object

dhmlau:
There are 2 kinds of errors when opening our generated openapi.json in editor.swagger.io:
1. responses object is empty
   - According to the spec (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#response-object), at least `description` should exist.
   - A quick change for users (to just get rid of the error) is to modify `responses: {}` to:
   ```
   responses: {
        "200": {
          description: "some description"
        }
      }
   ```

2. If `in` field in the parameter object is `path`, the following needs to exist:

```
required: true
```
   
    i.e. 
   ```
    parameters:
        - name: id
          in: path
          schema:
            type: string
          required: true 
    ```

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object

bajtos:
> I am proposing to modify our @param.path.* decorators to always set required: true. See packages/openapi-v3/src/decorators/parameter.decorator.ts.

Never mind :) A different solution has been already landed via https://github.com/strongloop/loopback-next/pull/1633 👍

virkt25:
## Acceptance Criteria

- Introduce a default `responses` object on the generated OpenAPI Spec (`responses: {200: {description: ''}}`
- Introduce a `@responses` decorator which takes in an OpenAPI responses object for a given controller method. This will take the exact spec. Should be able to resolve {schema: MODEL} to a model in the spec.
- Update CLI to automatically decorate CRUD APIs with `@responses()` decorator
[...]

PR #1655, opened 2018-08-29
feat(openapi-v3): responses decorator

virkt25:
connected to #1531 

---

This PR:
- Ensures we set a default `responses` property on a Controller methods OpenAPI Spec if one isn't given
- Extends the OpenAPI spec by introducing a `x-ts-type` property that can be used to specify the schema of a response object. Can also be used to specify the type of array items. 
```
content: {'x-ts-type': Todo} => content: {schema: {$ref: '#components/schemas/Todo'}}
content: {schema: {type: 'array', items: {x-ts-type: Todo}}} => content: {schema: {type: 'array', items: {$ref: '#components/schemas/Todo'}}}
```
- Refactors `generate-schema` to have a single function to resolve schema from a Function.
- Updates CLI Controller template to set a `responses` object based on the Model
- Updates `example-todo` / `example-todo-list` to set `responses` object

~This PR introduces:~
~- a new `@responses` decorator that can be used to decorate a Controller method with the appropriate OpenAPI responses object.~
~- for controller's not decorated with this property, or via the `@operation` and family (get, post, etc.) of decorators, we generate a default responses object to be OpenAPI 3.0 compliant~
~- `controller-spec` generator can set Model as `$ref` for `{schema: Model}` or {schema: {type: 'array', items: Model}}}` in the responses object.~
~- Decorate `example-todo` with `@responses()` decorator as an example. Once this PR is agreed upon, a follow up PR/commit will add decoration to CLI template + remaining examples + docs~

## Checklist

### DIFF
--- a/examples/todo-list/src/controllers/todo-list-todo.controller.ts
+++ b/examples/todo-list/src/controllers/todo-list-todo.controller.ts
@@ -13,33 +13,68 @@ export class TodoListTodoController {
     @repository(TodoListRepository) protected todoListRepo: TodoListRepository,
   ) {}
 
-  @post('/todo-lists/{id}/todos')
-  async create(@param.path.number('id') id: number, @requestBody() todo: Todo) {
+  @post('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: 'TodoList.Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
+      },
+    },
+  })
+  async create(
+    @param.path.number('id') id: number,
+    @requestBody() todo: Todo,
+  ): Promise<Todo> {
     return await this.todoListRepo.todos(id).create(todo);
   }
 
-  @get('/todo-lists/{id}/todos')
+  @get('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: "Array of Todo's belonging to TodoList",
+        content: {
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Todo}},
+          },
+        },
+      },
+    },
+  })
   async find(
     @param.path.number('id') id: number,
     @param.query.string('filter') filter?: Filter,
-  ) {
+  ): Promise<Todo[]> {
     return await this.todoListRepo.todos(id).find(filter);
   }
 
-  @patch('/todo-lists/{id}/todos')
+  @patch('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: 'TodoList.Todo PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
   async patch(
     @param.path.number('id') id: number,
     @requestBody() todo: Partial<Todo>,
     @param.query.string('where') where?: Where,
-  ) {
+  ): Promise<number> {
     return await this.todoListRepo.todos(id).patch(todo, where);
   }
 
-  @del('/todo-lists/{id}/todos')
+  @del('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: 'TodoList.Todo DELETE success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
   async delete(
     @param.path.number('id') id: number,
     @param.query.string('where') where?: Where,
-  ) {
+  ): Promise<number> {
     return await this.todoListRepo.todos(id).delete(where);
   }
 }
--- a/examples/todo/src/controllers/todo.controller.ts
+++ b/examples/todo/src/controllers/todo.controller.ts
@@ -16,8 +16,15 @@ export class TodoController {
     @inject('services.GeocoderService') protected geoService: GeocoderService,
   ) {}
 
-  @post('/todos')
-  async createTodo(@requestBody() todo: Todo) {
+  @post('/todos', {
+    responses: {
+      '200': {
+        description: 'Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
+      },
+    },
+  })
+  async createTodo(@requestBody() todo: Todo): Promise<Todo> {
     if (todo.remindAtAddress) {
       // TODO(bajtos) handle "address not found"
       const geo = await this.geoService.geocode(todo.remindAtAddress);
@@ -26,11 +33,17 @@ export class TodoController {
       // https://gis.stackexchange.com/q/7379
       todo.remindAtGeo = `${geo[0].y},${geo[0].x}`;
     }
-
     return await this.todoRepo.create(todo);
   }
 
-  @get('/todos/{id}')
+  @get('/todos/{id}', {
+    responses: {
+      '200': {
+        description: 'Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
+      },
+    },
+  })
   async findTodoById(
     @param.path.number('id') id: number,
     @param.query.boolean('items') items?: boolean,
@@ -38,12 +51,30 @@ export class TodoController {
     return await this.todoRepo.findById(id);
   }
 
-  @get('/todos')
+  @get('/todos', {
+    responses: {
+      '200': {
+        description: 'Array of Todo model instances',
+        content: {
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Todo}},
+          },
+        },
+      },
+    },
+  })
   async findTodos(): Promise<Todo[]> {
     return await this.todoRepo.find();
   }
 
-  @put('/todos/{id}')
+  @put('/todos/{id}', {
+    responses: {
+      '200': {
+        description: 'Todo PUT success',
+        content: {'application/json': {'x-ts-type': Boolean}},
+      },
+    },
+  })
   async replaceTodo(
     @param.path.number('id') id: number,
     @requestBody() todo: Todo,
@@ -51,7 +82,14 @@ export class TodoController {
     return await this.todoRepo.replaceById(id, todo);
   }
 
-  @patch('/todos/{id}')
+  @patch('/todos/{id}', {
+    responses: {
+      '200': {
+        description: 'Todo PATCH success',
+        content: {'application/json': {'x-ts-type': Boolean}},
+      },
+    },
+  })
   async updateTodo(
     @param.path.number('id') id: number,
     @requestBody() todo: Todo,
@@ -59,7 +97,14 @@ export class TodoController {
     return await this.todoRepo.updateById(id, todo);
   }
 
-  @del('/todos/{id}')
+  @del('/todos/{id}', {
+    responses: {
+      '200': {
+        description: 'Todo DELETE success',
+        content: {'application/json': {'x-ts-type': Boolean}},
+      },
+    },
+  })
   async deleteTodo(@param.path.number('id') id: number): Promise<boolean> {
     return await this.todoRepo.deleteById(id);
   }
--- a/packages/openapi-spec-builder/src/openapi-spec-builder.ts
+++ b/packages/openapi-spec-builder/src/openapi-spec-builder.ts
@@ -120,7 +120,7 @@ export class OpenApiSpecBuilder extends BuilderBase<OpenApiSpec> {
 export class OperationSpecBuilder extends BuilderBase<OperationObject> {
   constructor() {
     super({
-      responses: {},
+      responses: {'200': {description: 'An undocumented response body.'}},
     });
   }
 
--- a/packages/openapi-v3/src/controller-spec.ts
+++ b/packages/openapi-v3/src/controller-spec.ts
@@ -16,8 +16,9 @@ import {getJsonSchema} from '@loopback/repository-json-schema';
 import {OAI3Keys} from './keys';
 import {jsonToSchemaObject} from './json-to-schema';
 import * as _ from 'lodash';
+import {resolveSchema} from './generate-schema';
 
-const debug = require('debug')('loopback:openapi3:metadata');
+const debug = require('debug')('loopback:openapi3:metadata:controller-spec');
 
 // tslint:disable:no-any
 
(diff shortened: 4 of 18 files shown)


examples/todo/src/controllers/todo.controller.ts
```diff
@@ -16,8 +25,14 @@ export class TodoController {
     @inject('services.GeocoderService') protected geoService: GeocoderService,
   ) {}
 
+  @responses({
+    '200': {
+      description: 'Todo model instance',
+      content: {'application/json': {schema: Todo}},
+    },
+  })
```
raymondfeng: Looks good. Very intuitive to folks who are familiar with OpenAPI spec.

examples/todo/src/controllers/todo.controller.ts
```diff
@@ -38,11 +58,25 @@ export class TodoController {
...
   @get('/todos')
   async findTodos(): Promise<Todo[]> {
     return await this.todoRepo.find();
   }
 
+  @responses({
+    '200': {
+      description: 'Put success',
+      content: {'application/json': {schema: {type: 'boolean'}}},
```
raymondfeng: We might want to allow `{schema: Boolean}`.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -93,12 +94,57 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+        } else if (
+          operationSpec.responses[code].content[c].schema.type === 'array' &&
+          typeof operationSpec.responses[code].content[c].schema.items ===
+            'function'
+        ) {
+          debug(
+            'responses schema of type array detected with items as a model / function',
+          );
+          operationSpec.responses[code].content[c].schema.items = getModelRef(
+            operationSpec.responses[code].content[c].schema.items,
```
raymondfeng: Refactor `operationSpec.responses[code].content[c].schema` to be a variable.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -239,3 +285,7 @@ export function getControllerSpec(constructor: Function): ControllerSpec {
   }
   return spec;
 }
+
+function getModelRef(fn: Function) {
+  return {$ref: `#/components/schemas/${fn.name}`};
```
raymondfeng: If `fn` is a Model class with `@model({name: 'my-model'})`, we should use `my-model`.

bajtos:
The proposed decorator API looks great for advanced users. I have a question though - how is `@responses` different from specifying the `responses` object via `@post` options? To me, it does not bring enough benefits to justify the implementation/maintenance cost and more importantly possible confusion between `@responses` and `@reponse` that is described as the next decorator to be implemented.

Current proposal:

```ts
  @post('/todos')
  @responses({
     '200': {
       description: 'Todo model instance',
       content: {'application/json': {schema: Todo}},
     },
   })
```

Existing solution working OOTB right now:

```ts
  @post('/todos', {
    responses: {
      description: 'Todo model instance',
      content: {'application/json': {schema: Todo}},
    },
  })
```
}
If our main goal is to get a valid OpenAPI spec regardless of the developer experience, then I am proposing to simply fix our example apps and CLI templates to use the existing solution described above.

When it comes to adding a new decorator API, I would like to see a simpler API that will hide away the complexity of OpenAPI format and allow developers to focus on the important parts only.

For example:

```ts
@post()
@returnsJson(Todo, 'Todo model instance')
async createTodo(@requestBody() todo: Todo): Promise<Todo> {
  // ...
}
```

I picked `returnsJson` as the first name that comes to my mind, we can find a more suited one that supports all different flavors we want to provide. I can imagine the following signatures:

 - `@returnsJson(201, Todo, 'Todo model instance')`
 - `@returnsJson(201, Todo, {description: 'Todo model instance', headers: {/*...*/}, links: {/*...*/}})`
 - `@response(201, {/* ResponseObject *})`

---

I am proposing to remove `@responses` decorator from this pull request and focus on 
 1. Modifying examples and CLI templates so that controller methods define responses via `@get`/`@post`/etc. options as shown below.
 2. Implementing the logic inferring the schema from the return value type provided by TypeScript.

I am not sure whether schema inference is usable. IIUC, we are not able to infer `Todo` type from the controller method return value `Promise<Todo>`. Since most controller methods are going to return promises, they will have to provide the schema explicitly and thus the schema inference is not worth the implementation & maintenance costs IMO.

On the other hand, if we are able to infer the schema from existing controller methods in our examples, then I would expect that no changes to our examples and CLI templates should be needed?

Either way, it seems to me that we need to implement only one of these two changes.

Thoughts?

packages/openapi-spec-builder/src/openapi-spec-builder.ts
```diff
@@ -120,7 +120,7 @@ export class OpenApiSpecBuilder extends BuilderBase<OpenApiSpec> {
 export class OperationSpecBuilder extends BuilderBase<OperationObject> {
   constructor() {
     super({
-      responses: {},
+      responses: {'200': {description: ''}},
```
bajtos: Let's use a non-empty string please. For example `description: 'An undocumented response body.'`

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -93,12 +94,57 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
     if (!operationSpec) {
       // The operation was defined via @operation(verb, path) with no spec
       operationSpec = {
-        responses: {},
+        responses: {
+          '200': {
+            description: '',
```
bajtos: Let's use a non-emptyh default value please. For example, we can use the controller name and method name to build a more descriptive text:

```ts
description: `Return value of controller method ${fullMethodName}`
```

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -93,12 +94,57 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+    debug('  processing responses for method %s', op);
+    let responses = MetadataInspector.getMethodMetadata<ResponsesObject>(
+      OAI3Keys.RESPONSES_KEY,
+      constructor.prototype,
+      op,
+    );
+
+    debug('  decorated responses for method %s: %o', op, responses);
+
+    const defaultResponse = {
```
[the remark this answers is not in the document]
bajtos: As I commented earlier, I am proposing to remove `@responses` decorator from this pull request.

If we decide to keep it around, then I agree too that the default response spec should be reused between these two places.

packages/openapi-v3/src/controller-spec.ts
bajtos: It would be great if we could share the implementation of "Model class to $ref" algorithm with openapi-v3 package, see https://github.com/strongloop/loopback-next/blob/8499862ed7bbb615d8bcdf2c2590fcdfaa624e8f/packages/openapi-v3/src/generate-schema.ts#L54-L65

raymondfeng:
Please note that `@operation` accepts `OperationObject`, which can contain `ResponsesObject`. The `SchemaObject` does not allow usage of TypeScript classes.

Here is an idea I have to mitigate the issue:

1. For a return type such as `Promise<Customer>`, we can use `{$ref: '#components.schemas.Customer'}`.

2. Introduce a `@schemas` decorator for the controller class to register model classes for schemas. Such as:
```ts
@schemas({Customer, MyOrder: Order})
export class MyController {}
``` 
Or a `@schema` decorator to map between TS and OpenAPI spec for parameters or return value, for example:
```ts
@schema(Customer)
async findById(id: string) {}

async create(@schema(Customer) customer: DeepPartial<Customer>) {}
```

With this approach, we can have an OpenAPI compatible spec for parameters and responses while allowing developers to map TS types to schema definitions.

In the long run, we can possibly use `AST` to pre-process the TypeScript file to automate these if necessary.

raymondfeng:
+1 to have `modelToRef` or `typeRef` utility.  But I don't think it will work well as part of the decoration, which happens when the class is built.

- `Todo` can be undefined. We might have to use typeResolver such as `modelRef(() => Todo)`.
- `modelRef` does not have access to `x-typescript-type`

virkt25:
UPDATED: 

This PR:
- Ensures we set a default `responses` property on a Controller methods OpenAPI Spec if one isn't given
- Extends the OpenAPI spec by introducing a `x-ts-type` property that can be used to specify the schema of a response object. Can also be used to specify the type of array items. 
```
content: {'x-ts-type': Todo} => content: {schema: {$ref: '#components/schemas/Todo'}}
content: {schema: {type: 'array', items: {x-ts-type: Todo}}} => content: {schema: {type: 'array', items: {$ref: '#components/schemas/Todo'}}}
```
- Refactors `generate-schema` to have a single function to resolve schema from a Function.
- Updates CLI Controller template to set a `responses` object based on the Model
- Updates `example-todo` / `example-todo-list` to set `responses` object

---

- I am purposely choosing to not introduce a `modelRef()` function as I think it's easier for `controller-spec` to resolve the schema from `x-ts-type`. 
- This is a first step, next we will need to look into adding sugar response decorators as proposed in #1672

examples/todo-list/src/controllers/todo-list-todo.controller.ts
```diff
@@ -13,33 +13,68 @@ export class TodoListTodoController {
     @repository(TodoListRepository) protected todoListRepo: TodoListRepository,
   ) {}
 
-  @post('/todo-lists/{id}/todos')
-  async create(@param.path.number('id') id: number, @requestBody() todo: Todo) {
+  @post('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: 'TodoList.Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
```
bajtos: Just in case it was not clear, if we added `modelRef(ctor)` helper as a sugar API returning `{'x-ts-type': ctor}`, then the line above could look like this:

```ts
content: {'application/json': modelRef(Todo)},
```

Having wrote that, I am not sure how much better that is and whether the benefits justify maintenance costs. I guess what we really want is to implement the new decorator.

examples/todo-list/src/controllers/todo-list-todo.controller.ts
```diff
@@ -13,33 +13,68 @@ export class TodoListTodoController {
...
+  ): Promise<Todo[]> {
     return await this.todoListRepo.todos(id).find(filter);
   }
 
-  @patch('/todo-lists/{id}/todos')
+  @patch('/todo-lists/{id}/todos', {
+    responses: {
+      '200': {
+        description: 'TodoList.Todo PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
```
bajtos: Hmm, a response returning just a number is not very usual in REST/JSON world, is it?

I think it would be better to return an object with a `count` property instead, such response is easier to extend with additional properties while preserving backwards compatibility.

Having said that, such change is out of scope of this pull request.

packages/openapi-spec-builder/src/openapi-spec-builder.ts
bajtos: PTAL, this comment hasn't been addressed yet.

> Let's use a non-empty string please. For example description: 'An undocumented response body.'

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -89,16 +90,55 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
       endpoint.spec = operationSpec;
     }
     debug('  operation for method %s: %j', op, endpoint);
 
+    debug('  spec responses for method %s: %o', op, operationSpec.responses);
+
+    const TS_TYPE_KEY = 'x-ts-type';
+
+    for (const code in operationSpec.responses) {
+      for (const c in operationSpec.responses[code].content) {
```
bajtos: IIRC, `content` is an optional property. Please add a test to verify what happens when the OAI spec does not provide any value for `content`, and fix the implementation if needed.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -89,16 +90,55 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+        debug('  evaluating response code %s with content: %o', code, c);
+        const tsType = operationSpec.responses[code].content[c][TS_TYPE_KEY];
```
bajtos: Please extract `operationSpec.responses[code].content[c]` into a shared variable to avoid repetition.

packages/openapi-v3/test/integration/controller-spec.integration.ts
```diff
@@ -41,7 +41,11 @@ describe('controller spec', () => {
       paths: {
         '/foo': {
           post: {
-            responses: {},
+            responses: {
+              '200': {
+                description: 'Return value of FooController.create',
+              },
```
bajtos: Interesting. Is this the test verifying how a response status with no `content` is handled?

> IIRC, `content` is an optional property. Please add a test to verify what happens when the OAI spec does not provide any value for `content`, and fix the implementation if needed.

It makes me wonder why this test passes, when the code processing the schema is not checking whether `.content` is set. Are we filling the missing `content` property somewhere else?

examples/todo-list/src/controllers/todo-list-todo.controller.ts
virkt25: I'd rather see new decorators than introduce a helper utility that we have to maintain.

examples/todo-list/src/controllers/todo-list-todo.controller.ts
virkt25: I agree it's unusual but it's what out Repository returns. I'll make a follow up issue for us to return an object instead.

packages/openapi-v3/src/controller-spec.ts
virkt25: `content` is an optional property and definitely took me a minute to realize what was happening here since I coded it. 

The `for (const c in operationSpec.responses[code].content)` loop is only entered if the `operationSpec.responses[code]` object had a `content` property. If the property is not defined, the loop is never entered as there are no properties on `undefined` -- making the implementation correct and sound. 

I also verified this behaviour on this page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

By changing the code to be as follows: 

```
var string1 = "";
var object1 = {a: 1, b: 2, c: 3};
// var object1 = {a: 1, b: 2, c: 3, d: {x:1, y:2}}; 

for (var property1 in object1.d) {
  console.log('in loop')
  string1 = string1 + property1;
}

console.log(string1);
// expected output: "123"
```

The above will log `""` as is and uncommenting line 3 should give `"xy"` after 2x `in loop`

packages/openapi-v3/test/integration/controller-spec.integration.ts
virkt25: See https://github.com/strongloop/loopback-next/pull/1655#discussion_r216884048

PR #1617, opened 2018-08-17
[WIP] fix: fix improper response code

hacksparrow:
For fixing https://github.com/strongloop/loopback-next/issues/1472 and https://github.com/strongloop/loopback-next/issues/1469.

PR #1621, opened 2018-08-20
feat: default 404 for delete request to non-existent resource

hacksparrow:
Send 404 for delete requests to non-existent resource by default. Addresses issues like #1472

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

Issue #1472, opened 2018-06-25
[REST] DELETE with invalid ID is successful


virkt25:
## Description / Steps to reproduce / Feature proposal

- In `examples/todo`, using the API Explorer make a request to the `DELETE` endpoint. Use `id` 2 / something that doesn't exist.
- You get a success message ... but it should've been a Bad Request since the `id` doesn't exist.

## Current Behavior

- Get a success message when deleting a resource that doesn't exist.

## Expected Behavior

- Return a non-success message when attempting to delete a resource that doesn't exist

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

marioestradarosa:
Actually, if you call the DELETE operation with no parameters, it will remove ALL records. I think the juggler Delete All is not checking the empty where clause either. For security this should be disable. What we do for now is that we override these delete methods and place our own validations.

bajtos:
Is this perhaps related to _PersistedModel.destroyAll has very dangerous behaviour if the filter is not passed in correctly https://github.com/strongloop/loopback/issues/3094_?

marioestradarosa:
It is not related, The DEL operation on endpoint /mymodel/{id} (deleteById) where ID doesn't exist returns an http 200 status with empty response body instead of 404 probably. If {id} exists it returns 200 status with true in the response body.

marioestradarosa:
However, on DeleteAll() I believe which is the main problem because users can delete all their records in one operation, we should read the record first, even if that means a double operation on DeleteAll.

if (find) {
    Delete()
}

  This way if the user sends an invalid where clause we don't issue the delete operation at all.

bajtos:
> It is not related, 

@marioestradarosa In that case, let's discuss `deleteAll` elsewhere please, to keep this issue focused on deleting a record by id that does not exist. I think strongloop/loopback#3094 is the best place for such discussion, we have already covered many aspects there.

bajtos:
> Return a non-success message when attempting to delete a resource that doesn't exist

Let's discuss this, I can see good reasons why it may be actually a good thing to return 200 even when the requested id does not exist.

Here are two use cases I have in mind:

- Multiple clients are trying to delete the same record. This creates a race condition - the first HTTP request to arrive is going to delete the record (and return 200), the second HTTP request will become a no-op.

[...]

PR #1658, opened 2018-08-30
fix: return 404 when a model was not found

### LINKED ISSUE #1469: Incorrect response codes for invalid HTTP calls

From #1206 

- In `examples/todo` make a **GET** request to `/todo/id` using API Explorer. `id` shouldn't have a value (`undefined`).
- You get back a 500 - Server Error **BUT** it should've been a 400 (Bad Request) since the request is missing the `id`. Same goes for an invalid `id` value such as `-10`, `ten`, etc.  

## Current Behavior

- Invalid Response Code

## Expected Behavior

- Correct HTTP Response Code as per http://www.restapitutorial.com/httpstatuscodes.html

## Acceptance Criteria

- [ ] Fix response codes as described above, see https://github.com/strongloop/loopback-next/issues/1469#issuecomment-405324333
- [ ] Appropriate test cases

bajtos:
_This is an alternative solution to #1617 that fixes #1469._

- test: fix incorrect usage of rejectedWith assertion. I discovered this problem while fixing tests added as part of this pull request.
- feat(repository): change Model.modelName to a computed property. This ensures `modelName` is always available and match model definition.
- feat(repository): implement EntityNotFoundError. This error builds a descriptive error message and sets `error.code` to `ENTITY_NOT_FOUND` string.
- refactor: rework "findById" to leverage EntityNotFoundError. As a result, "model not found" scenario can be detected by checking whether the error `code` is `ENTITY_NOT_FOUND`.
- fix(rest): return 404 when a model was not found. Here we modify the `reject` action to convert `ENTITY_NOT_FOUND` error code to response status code 404.
- docs: add Error-handling.md, describe error codes
- feat(rest): add error codes for REST validation errors
    `VALIDATION_FAILED`, `INVALID_PARAMETER_VALUE`, `MISSING_REQUIRED_PARAMETER`

Out of scope of this pull request:
 - Update updateById,deleteById to throw EntityNotFound error for consistency.
 - Add a factory method allowing people to obtain a modified Repo instance that returns undefined if not found, as described in https://github.com/strongloop/loopback-next/pull/1658#issuecomment-420318023.

## Checklist

### DIFF
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -182,6 +182,10 @@ children:
       url: Decorators_repository.html
       output: 'web, pdf'
 
+  - title: 'Error handling'
+    url: Error-handling.html
+    output: 'web, pdf'
+
 - title: 'Booting an Application'
   url: Booting-an-Application.html
   output: 'web, pdf'
--- a/examples/todo-list/src/controllers/todo-list.controller.ts
+++ b/examples/todo-list/src/controllers/todo-list.controller.ts
@@ -3,8 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Filter, Where, repository} from '@loopback/repository';
-import {post, param, get, patch, del, requestBody} from '@loopback/rest';
+import {Filter, repository, Where} from '@loopback/repository';
+import {del, get, param, patch, post, requestBody} from '@loopback/rest';
 import {TodoList} from '../models';
 import {TodoListRepository} from '../repositories';
 
--- a/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
+++ b/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
@@ -1,4 +1,5 @@
-import {Filter, Where, repository} from '@loopback/repository';
+import {Filter, repository, Where} from '@loopback/repository';
+
 import {
   post,
   param,
--- a/packages/repository/src/errors/entity-not-found.error.ts
+++ b/packages/repository/src/errors/entity-not-found.error.ts
@@ -1,0 +1,40 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity} from '../model';
+
+export class EntityNotFoundError<ID, Props extends object = {}> extends Error {
+  code: string;
+  entityName: string;
+  entityId: ID;
+
+  constructor(
+    entityOrName: typeof Entity | string,
+    entityId: ID,
+    extraProperties?: Props,
+  ) {
+    const entityName =
+      typeof entityOrName === 'string'
+        ? entityOrName
+        : entityOrName.modelName || entityOrName.name;
+
+    const quotedId = JSON.stringify(entityId);
+
+    super(`Entity not found: ${entityName} with id ${quotedId}`);
+
+    Error.captureStackTrace(this, this.constructor);
+
+    this.code = 'ENTITY_NOT_FOUND';
+    this.entityName = entityName;
+    this.entityId = entityId;
+
+    Object.assign(this, extraProperties);
+  }
+}
+
+// tslint:disable-next-line:no-any
+export function isEntityNotFoundError(e: any): e is EntityNotFoundError<any> {
+  return e instanceof EntityNotFoundError;
+}
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -3,26 +3,26 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import * as legacy from 'loopback-datasource-juggler';
-
-import * as assert from 'assert';
 import {isPromiseLike} from '@loopback/context';
+import * as assert from 'assert';
+import * as legacy from 'loopback-datasource-juggler';
 import {
-  Options,
   AnyObject,
   Command,
-  NamedParameters,
-  PositionalParameters,
   DataObject,
+  NamedParameters,
+  Options,
+  PositionalParameters,
 } from '../common-types';
+import {HasManyDefinition} from '../decorators/relation.decorator';
+import {EntityNotFoundError} from '../errors';
 import {Entity, ModelDefinition} from '../model';
 import {Filter, Where} from '../query';
-import {EntityCrudRepository} from './repository';
 import {
   createHasManyRepositoryFactory,
   HasManyRepositoryFactory,
 } from './relation.factory';
-import {HasManyDefinition} from '../decorators/relation.decorator';
+import {EntityCrudRepository} from './repository';
 
 export namespace juggler {
   export import DataSource = legacy.DataSource;
@@ -211,7 +211,7 @@ export class DefaultCrudRepository<T extends Entity, ID>
       this.modelClass.findById(id, filter, options),
     );
     if (!model) {
-      throw new Error(`no ${this.modelClass.name} found with id "${id}"`);
+      throw new EntityNotFoundError(this.entityClass, id);
     }
     return this.toEntity(model);
   }
--- a/packages/repository/src/repositories/repository.ts
+++ b/packages/repository/src/repositories/repository.ts
@@ -5,7 +5,6 @@
 
 import {Entity, ValueObject, Model} from '../model';
 import {
-  Class,
   DataObject,
   Options,
   AnyObject,
@@ -16,6 +15,7 @@ import {
 import {DataSource} from '../datasource';
 import {CrudConnector} from '../connectors';
 import {Filter, Where} from '../query';
+import {EntityNotFoundError} from '../errors';
 
 // tslint:disable:no-unused-variable
 
@@ -134,8 +134,10 @@ export interface EntityCrudRepository<T extends Entity, ID>
   delete(entity: DataObject<T>, options?: Options): Promise<boolean>;
 
   /**
-   * Find an entity by id
+   * Find an entity by id, return a rejected promise if not found.
    * @param id Value for the entity id
+   * @param filter Additional query options. E.g. `filter.include` configures
+   * which related models to fetch as part of the database query (or queries).
    * @param options Options for the operations
    * @returns A promise of an entity found for the id
    */
@@ -206,16 +208,20 @@ export class CrudRepositoryImpl<T extends Entity, ID>
   implements EntityCrudRepository<T, ID> {
   private connector: CrudConnector;
 
-  constructor(public dataSource: DataSource, public model: Class<T>) {
+  constructor(
+    public dataSource: DataSource,
+    // model should have type "typeof T", but that's not supported by TSC
+    public model: typeof Entity & {prototype: T},
+  ) {
     this.connector = dataSource.connector as CrudConnector;
   }
 
   private toModels(data: Promise<DataObject<Entity>[]>): Promise<T[]> {
-    return data.then(items => items.map(i => new this.model(i)));
+    return data.then(items => items.map(i => new this.model(i) as T));
   }
 
   private toModel(data: Promise<DataObject<Entity>>): Promise<T> {
-    return data.then(d => new this.model(d));
+    return data.then(d => new this.model(d) as T);
   }
 
   create(entity: DataObject<T>, options?: Options): Promise<T> {
@@ -250,16 +256,18 @@ export class CrudRepositoryImpl<T extends Entity, ID>
     return this.toModels(this.connector.find(this.model, filter, options));
   }
 
-  findById(id: ID, options?: Options): Promise<T> {
+  async findById(id: ID, filter?: Filter, options?: Options): Promise<T> {
     if (typeof this.connector.findById === 'function') {
       return this.toModel(this.connector.findById(this.model, id, options));
     }
     const where = this.model.buildWhereForId(id);
-    return this.connector
-      .find(this.model, {where: where}, options)
-      .then((entities: T[]) => {
-        return entities[0];
-      });
+    const entities = await this.toModels(
+      this.connector.find(this.model, {where: where}, options),
+    );
+    if (!entities.length) {
+      throw new EntityNotFoundError(this.model, id);
+    }
+    return entities[0];
   }
 
   update(entity: DataObject<T>, options?: Options): Promise<boolean> {
--- a/packages/rest/src/providers/reject.provider.ts
+++ b/packages/rest/src/providers/reject.provider.ts
@@ -9,6 +9,12 @@ import {HttpError} from 'http-errors';
 import {RestBindings} from '../keys';
 import {writeErrorToResponse, ErrorWriterOptions} from 'strong-error-handler';
 
+// TODO(bajtos) Make this mapping configurable at RestServer level,
+// allow apps and extensions to contribute additional mappings.
+const codeToStatusCodeMap: {[key: string]: number} = {
+  ENTITY_NOT_FOUND: 404,
+};
+
 export class RejectProvider implements Provider<Reject> {
   constructor(
     @inject(RestBindings.SequenceActions.LOG_ERROR)
@@ -23,6 +29,14 @@ export class RejectProvider implements Provider<Reject> {
 
   action({request, response}: HandlerContext, error: Error) {
     const err = <HttpError>error;
+
+    if (!err.status && !err.statusCode && err.code) {
+      const customStatus = codeToStatusCodeMap[err.code];
+      if (customStatus) {
+        err.statusCode = customStatus;
+      }
+    }
+
     const statusCode = err.statusCode || err.status || 500;
     writeErrorToResponse(err, request, response, this.errorWriterOptions);
     this.logError(error, statusCode, request);
--- a/packages/rest/src/rest-http-error.ts
+++ b/packages/rest/src/rest-http-error.ts
@@ -7,12 +7,22 @@ export namespace RestHttpErrors {
     extraProperties?: Props,
   ): HttpErrors.HttpError & Props {
     const msg = `Invalid data ${JSON.stringify(data)} for parameter ${name}!`;
-    return Object.assign(new HttpErrors.BadRequest(msg), extraProperties);
+    return Object.assign(
+      new HttpErrors.BadRequest(msg),
+      {
+        code: 'INVALID_PARAMETER_VALUE',
+        parameterName: name,
+      },
+      extraProperties,
+    );
   }
 
   export function missingRequired(name: string): HttpErrors.HttpError {
     const msg = `Required parameter ${name} is missing!`;
-    return new HttpErrors.BadRequest(msg);
+    return Object.assign(new HttpErrors.BadRequest(msg), {
+      code: 'MISSING_REQUIRED_PARAMETER',
+      parameterName: name,
+    });
   }
 
   export function invalidParamLocation(location: string): HttpErrors.HttpError {
@@ -23,7 +33,12 @@ export namespace RestHttpErrors {
   export const INVALID_REQUEST_BODY_MESSAGE =
     'The request body is invalid. See error object `details` property for more info.';
   export function invalidRequestBody(): HttpErrors.HttpError {
-    return new HttpErrors.UnprocessableEntity(INVALID_REQUEST_BODY_MESSAGE);
+    return Object.assign(
+      new HttpErrors.UnprocessableEntity(INVALID_REQUEST_BODY_MESSAGE),
+      {
+        code: 'VALIDATION_FAILED',
+      },
+    );
   }
 
   /**
--- a/packages/rest/test/unit/router/reject.provider.unit.ts
+++ b/packages/rest/test/unit/router/reject.provider.unit.ts
@@ -28,6 +28,18 @@ describe('reject', () => {
     expect(result).to.have.property('statusCode', 500);
   });
 
+  it('converts error code ENTITY_NOT_FOUND to status code 404', async () => {
+    const reject = new RejectProvider(noopLogger).value();
+
+    const notFoundError: Error & {code?: string} = new Error('not found');
+    notFoundError.code = 'ENTITY_NOT_FOUND';
+
+    reject(contextStub, notFoundError);
+    const result = await contextStub.result;
+
+    expect(result.statusCode).to.equal(404);
+  });
+
   it('logs the error', async () => {
     const logger = sinon.spy() as LogError & SinonSpy;
     const reject = new RejectProvider(logger).value();
--- a/docs/site/Error-handling.md
+++ b/docs/site/Error-handling.md
@@ -11,9 +11,12 @@ permalink: /doc/en/lb4/Error-handling.html
 In order to allow clients to reliably detect individual error causes, LoopBack
 sets the error `code` property to a machine-readable string.
 
-| Error code       | Description                                                                                                                                                                                                              |
-| :--------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| ENTITY_NOT_FOUND | The entity (model) was not found. This error is returned for example by [`EntityCrudRepository.prototype.findById`](http://apidocs.loopback.io/@loopback%2fdocs/repository.html#EntityCrudRepository.prototype.findById) |
+| Error code                 | Description                                                                                                                                                                                                              |
+| :------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ENTITY_NOT_FOUND           | The entity (model) was not found. This error is returned for example by [`EntityCrudRepository.prototype.findById`](http://apidocs.loopback.io/@loopback%2fdocs/repository.html#EntityCrudRepository.prototype.findById) |
+| VALIDATION_FAILED          | The data provided by the client is not a valid entity.                                                                                                                                                                   |
+| INVALID_PARAMETER_VALUE    | The value provided for a parameter of a REST endpoint is not valid. For example, a string value was provided for a numeric parameter.                                                                                    |
+| MISSING_REQUIRED_PARAMETER | No value was provided for a required parameter.                                                                                                                                                                          |
 
 Besides LoopBack-specific error codes, your application can encounter low-level
 error codes from Node.js and the underlying operating system. For example, when
--- a/docs/site/Error-handling.md
+++ b/docs/site/Error-handling.md
@@ -1,0 +1,26 @@
+---
+lang: en
+title: Error Handling
+keywords: LoopBack 4.0
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Error-handling.html
+---
+
+## Known Error Codes
+
+In order to allow clients to reliably detect individual error causes, LoopBack
+sets the error `code` property to a machine-readable string.
+
+| Error code       | Description                                                                                                                                                                                                              |
+| :--------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ENTITY_NOT_FOUND | The entity (model) was not found. This error is returned for example by [`EntityCrudRepository.prototype.findById`](http://apidocs.loopback.io/@loopback%2fdocs/repository.html#EntityCrudRepository.prototype.findById) |
+
+Besides LoopBack-specific error codes, your application can encounter low-level
+error codes from Node.js and the underlying operating system. For example, when
+a connector is not able to reach the database, an error with code `ECONNREFUSED`
+is returned.
+
+See the following resources for a list of low-level error codes:
+
+- [Common System Errors](https://nodejs.org/api/errors.html#errors_common_system_errors)
+- [Node.js Error Codes](https://nodejs.org/api/errors.html#errors_node_js_error_codes)
--- a/examples/todo-list/test/acceptance/todo-list.acceptance.ts
+++ b/examples/todo-list/test/acceptance/todo-list.acceptance.ts
@@ -3,6 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {EntityNotFoundError} from '@loopback/repository';
 import {createClientForHandler, expect, supertest} from '@loopback/testlab';
 import {TodoListApplication} from '../../src/application';
 import {TodoList} from '../../src/models/';
@@ -110,7 +111,7 @@ describe('Application', () => {
         .expect(200);
       await expect(
         todoListRepo.findById(persistedTodoList.id),
-      ).to.be.rejectedWith(/no TodoList found with id/);
+      ).to.be.rejectedWith(EntityNotFoundError);
     });
   });
 
--- a/examples/todo-list/test/acceptance/todo-list.acceptance.ts
+++ b/examples/todo-list/test/acceptance/todo-list.acceptance.ts
@@ -92,6 +92,10 @@ describe('Application', () => {
       expect(result.body).to.deepEqual(expected);
     });
 
+    it('returns 404 when a todo does not exist', () => {
+      return client.get('/todo-lists/99999').expect(404);
+    });
+
     it('updates a todoList by ID', async () => {
       const updatedTodoList = givenTodoList({
         title: 'A different title to the todo list',
--- a/examples/todo-list/test/acceptance/todo.acceptance.ts
+++ b/examples/todo-list/test/acceptance/todo.acceptance.ts
@@ -3,6 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {EntityNotFoundError} from '@loopback/repository';
 import {createClientForHandler, expect, supertest} from '@loopback/testlab';
 import {TodoListApplication} from '../../src/application';
 import {Todo} from '../../src/models/';
@@ -96,7 +97,7 @@ describe('Application', () => {
         .send()
         .expect(200);
       await expect(todoRepo.findById(persistedTodo.id)).to.be.rejectedWith(
-        /no Todo found with id/,
+        EntityNotFoundError,
       );
     });
   });
--- a/examples/todo-list/test/acceptance/todo.acceptance.ts
+++ b/examples/todo-list/test/acceptance/todo.acceptance.ts
@@ -64,6 +64,10 @@ describe('Application', () => {
       expect(result.body).to.deepEqual(expected);
     });
 
+    it('returns 404 when a todo does not exist', () => {
+      return client.get('/todos/99999').expect(404);
+    });
+
     it('replaces the todo by ID', async () => {
       const updatedTodo = givenTodo({
         title: 'DO SOMETHING AWESOME',
--- a/examples/todo/test/acceptance/todo.acceptance.ts
+++ b/examples/todo/test/acceptance/todo.acceptance.ts
@@ -3,16 +3,17 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {EntityNotFoundError} from '@loopback/repository';
 import {createClientForHandler, expect, supertest} from '@loopback/testlab';
 import {TodoListApplication} from '../../src/application';
 import {Todo} from '../../src/models/';
 import {TodoRepository} from '../../src/repositories/';
 import {
-  HttpCachingProxy,
   aLocation,
   getProxiedGeoCoderConfig,
   givenCachingProxy,
   givenTodo,
+  HttpCachingProxy,
 } from '../helpers';
 
 describe('Application', () => {
@@ -128,7 +129,7 @@ describe('Application', () => {
         .send()
         .expect(200);
       await expect(todoRepo.findById(persistedTodo.id)).to.be.rejectedWith(
-        /no Todo found with id/,
+        EntityNotFoundError,
       );
     });
   });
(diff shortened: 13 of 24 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -211,7 +211,12 @@ export class DefaultCrudRepository<T extends Entity, ID>
       this.modelClass.findById(id, filter, options),
     );
     if (!model) {
-      throw new Error(`no ${this.modelClass.name} found with id "${id}"`);
+      const quotedId = JSON.stringify(id);
+      const err: Error & {code?: string} = new Error(
```
hacksparrow: Maybe declare the type `Error & {code?: string}` somewhere? We will be using it in other places too.

packages/rest/src/providers/reject.provider.ts
```diff
@@ -11,6 +11,12 @@ import {RestBindings} from '../keys';
 
 const debug = require('debug')('loopback:rest:reject');
 
+// TODO(bajtos) Make this mapping configurable at RestServer level,
+// allow apps and extensions to contribute additional mappings.
+const codeToStatusCodeMapping: {[key: string]: number} = {
+  MODEL_NOT_FOUND: 404,
```
hacksparrow: This is out custom error code name, right? We'll have to document this and other possible error types. Eg:

1. Database serve not reachable
2. Authentication failure
3. Authorization failure
4. Runtime errors

packages/rest/src/providers/reject.provider.ts
```diff
@@ -22,7 +28,15 @@ export class RejectProvider implements Provider<Reject> {
   }
 
   action({request, response}: HandlerContext, error: Error) {
-    const err = <HttpError>error;
+    const err = error as HttpError & {code?: string};
```
hacksparrow: Should we declare the `HttpError & {code?: string}` type somewhere?

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: Good point 👍

packages/rest/src/providers/reject.provider.ts
bajtos: Yes, this is our custom error code. We are using the same `code` property as Node.js core does, as a result we "inherit" all error codes from Node.js core too - see https://nodejs.org/api/errors.html#errors_node_js_error_codes and https://nodejs.org/api/errors.html#errors_common_system_errors

> Database serve not reachable

I think this will typically manifests as `ECONNREFUSED`,  `ECONNRESET` or `ETIMEDOUT`.

> Authentication failure
> Authorization failure
> Runtime errors

Agreed. See existing auth error codes in LB 3.x - e.g. [common/models/user.js](https://github.com/strongloop/loopback/blob/master/common/models/user.js#L245)

raymondfeng:
While I support the idea to have `error mapping` capability, I'm not very keen to have `findById` throws an error if the id does not match any records for a few reasons:

1. Other `*ById` methods do not throw, such as `deleteById` and `updateById`.
2. There should be options for consumers of `findById` API to handle the `not found` situation, such as:
  - Return `null` or `undefined`
  - Throw NotFound
  - Use a default value

See https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html#findById-ID-

virkt25:
> Other *ById methods do not throw, such as deleteById and updateById.

I would hope / think that the intention would be to keep all the other methods consistent and this is just to demonstrate the idea. 

> There should be options for consumers of findById API to handle the not found situation, such as:

I'm confused ... how does this approach not allow the consumers of findById API to handle the not found situation? Wouldn't a consumer of this API now be able to try / catch this error and know the issue vs. just getting a `null` and dealing with it?

joeytwiddle:
The mongoose library returns `null` if [findById](https://mongoosejs.com/docs/api.html#model_Model.findById) did not find the document.

As a result, we either had to check the result of every `findById` call, and throw a suitable error each time, or just accept that sometimes we will get an error from the processing code: `Cannot read property 'foo' of null`

After this experience, I feel that having a 404 be the default behaviour would be quite convenient, reducing boilerplate and uncertainty in the code.

But I can see the benefits of Raymond's suggestion to optionally return `null` or a default value. In some code, especially when using `findOne()`, we might _expect_ the document to sometimes be missing. (I assume we are aiming for parity between `findById()` and `findOne()` anyway.)

However, a function that can sometimes return `null` might complicate the return type. Will TypeScript start warning us to check for `null` results even when the option is set/defaulting to throw? If so, would it be worth having a separate function `findByIdMaybe()` for that situation, just to improve type assistance?

bajtos:
Thank you @joeytwiddle for chiming in! It's always great to hear feedback from users 👍 

> Other *ById methods do not throw, such as deleteById and updateById.

We have recently landed https://github.com/strongloop/loopback-next/pull/1621 which changed the default behavior of `deleteById` to throw 404 error when the target model was not found. See also https://github.com/strongloop/loopback-next/issues/1472 for context.

I think we need to rethink "not found" handling in all these three methods and implement a consistent solution.

> There should be options for consumers of findById API to handle the not found situation, such as:
> - Return null or undefined
> - Throw NotFound
> - Use a default value
> See https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html#findById-ID-

I fully agree we should provide API that makes it easy for consumers to handle "not found" situation. At the same time, I'd like our API to make it easy for consumers to not have to deal with "not found" situation.

In the past, I had success with providing two methods: `get` and `find`. In this case, `getById` would throw when the model was not found and `findById` returns a special value indicating unknown model id (e.g. `null`).

The link to Spring framework is a great discussion starter. Notice that Spring's `findById` does not return `null` to signal "not found" case (which would be a perfectly valid API in Java), but uses `Optional<T>` (also known as `Maybe<T>` in other languages).  This new type has few benefits: 
 - It forces developers to think about the case when no value was returned. This is enforced by the compiler, because you cannot assign `Optional<T>` to a variable/argument expecting `T`.
 - Provides a convenient way of detecting "not found" case.
 - Provides a value getter that throws an error when no value was set (to avoid "NullReferenceException").
  - Can integrate with other language constructs via monadic behavior, e.g. one can define `+` operator that can add two `Optional<number>` values in such way that "not found" is treated as "NaN" type, etc.

An example to illustrate the point:

```ts
const result = repo.findById(123);

// We are pretty sure the id exist
// Let the framework throw an error if not found
// (the check is implemented in "value" getter)
console.log(result.value);

// We don't know if the id exists
// Handle "not found" case ourselves
if (result.hasValue) {
  console.log(result.value);
} else {
  console.log('not found');
}
```

`Optional<T>` is a great solution in typed languages where the compiler can ensure that `Optional<T>` is not accidentally passed to functions expecting `T`. I am not sure how much useful it is in JavaScript though, because there is no compiler to tell the developer that they forget to handle "not found" case. (We still do want to support LB4 apps written in pure JavaScript, right?)

Consider the following controller method:

```js
async findTodo(id) {
  return await this.repo.findById(id);
}
```

The code looks fine, only an experienced LB4 user with a keen eye may notice that this method is going to return model data wrapped in `value` property, e.g. `{value: {id: 1, name: 'Clean my room'}}`, instead of the correct payload `{id: 1, name: 'Clean my room'}`.

> I don't see a lot of JS developers use try-catch for NotFoundError as testing the error type is not very easy in JS. The null check is much simpler.

I agree that using try-catch to specifically handle "not found" error can be cumbersome.

I disagree that returning `null` makes code simpler. In TypeScript with `strictNullChecks` enabled, a `findById` method returning `Promise<T | null>` forces every caller to deal with the "not found" (`null`) case, even if the caller is pretty sure the model exists (typically in unit tests). We are effectively trading ease of use in code that wants to handle "not found" case for ease of use in code that wants to treat "not found" case as any other exceptional situation (e.g. a database connection problem).

> However, a function that can sometimes return null might complicate the return type. Will TypeScript start warning us to check for null results even when the option is set to throw?

This can be handled by function overloads, we already do something similar for `context.get`, see [context.ts](https://github.com/strongloop/loopback-next/blob/ac011f80b229b5dd5c69851ab14b39705b852629/packages/context/src/context.ts#L236-L299).

The idea is to tell TypeScript that the same function returns different types depending on the input arguments.

A mock-up to illustrate my point, it may not work exactly this way.

```ts
findById(id: ID): Promise<T>;
findById(id: ID, {optional: false}): Promise<T>;
findById(id: ID, {optional: true}): Promise<T | null>;

findById(id: ID, options: {optional: boolean}): Promise<T | null> {
  const result: T | null = // get data via juggler
  if (result) {
    return result;
  }

  if (options && options.optional) {
      return null;
  }
  
  throw /* not found error */
}
```

hacksparrow:
In the the callback pattern `(err, result)`, it is probably the most common practice to pass errors in `err` and return the results of a query in `result`. If nothing was found `result` would likely be `undefined` or `null`, depending on the implementation of the querying function. In this pattern conveying a 404 via `error` would be confusing, even if possible.

Since we are moving on to `async`-`await`, rejecting via any means (throwing or calling `reject()`) should be the convention for conveying the non-finding of a resource. The rejection message can have more details about why it was not found - the resource doesn't actually exist, network failed, not authorized etc.

Ideally, resolution should always return a resource of a single expected type.

raymondfeng:
For references:

TypeORM (https://github.com/typeorm/typeorm/blob/master/src/repository/Repository.ts#L268)
  - `findOne` returns `Promise<Entity|undefined>`
  - `findOneOrFail` returns `Promise<Entity>` and throws if not found

Sails.js (https://sailsjs.com/documentation/reference/waterline-orm/models/find-one)
  - `findOne` returns `undefined` if not found

BookShelf (http://bookshelfjs.org/#Model-instance-fetch)
  - Option `(require=false] Boolean`: Reject the returned response with a NotFoundError if results are empty.

marioestradarosa:
Why not delegate this error type transformation to the controller  and leave the current undefined if not found to the artifacts .
 
```ts
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    return await this.daisyRepository.findById(id);
  }
```  
like such:

```ts
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    let result =  await this.daisyRepository.find(filter);
    if (result === undefined || result === null) { 
             throw new HttpErrors. NotFound(`id ${id} was not found`);
       } 
    return result;
  }
```  

At the end, 404 status is really something expected from a client application such as the following. The problem is that if we don't receive an http status error, then the logic should change in this snippet.

```ts
      this.dummyService.findDaisy(arg).toPromise().then(data => {
         this.totalRecords = data;
        // this.paginator.length = data;
      }).catch(reason => { // do something }) ;
```

raymondfeng:
> Why not delegate this error type transformation to the controller and leave the current undefined if not found to the artifacts .

That's how I would draw the line too. The repository API just returns `null` (or a better object) if not found and we leave the mapping to controllers, such as `null --> 404`.

bajtos:
> > Why not delegate this error type transformation to the controller and leave the current undefined if not found to the artifacts .

> That's how I would draw the line too. The repository API just returns `null` (or a better object) if not found and we leave the mapping to controllers, such as `null --> 404`.

While a null-returning API is practical for users that are expecting the model instance to not exist (e.g. a Controller method), it's very impractical for users that are certain the instance should exist (e.g. unit/integration tests). See my earlier comment:

_I fully agree we should provide API that makes it easy for consumers to handle "not found" situation. At the same time, I'd like our API to make it easy for consumers to not have to deal with "not found" situation._

In the light of the discussion above, I am proposing the following plan going forward:
1. Add a new method `getById` that throws an error when the model was not found.
2. Modify `findById` to return `null` instead of throwing an error. Modify tests expecting the model instance exists to use `getById` instead of `findById`.
3. Add a helper for constructing "model instance id XYZ was not found"
4. Modify controller template & examples to handle "not found" error by leveraging the two mechanisms described above

hacksparrow:
@marioestradarosa we considered that too. The problem with that we will have repetitive code in controllers - all `*ById` methods have to implement 404 handlers of their own. Even if we create a common helper function for that purpose, we will end up with an odd-looking experience in the controller methods.

```js
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    return await this.daisyRepository.findById(id);
  }
```

vs

```js
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    return callThe404Helper(await this.daisyRepository.findById(id));
  }
```
That's not intuitive.

Also, it requires a change in `EntityCrudRepository `:

```js
findById(id: ID, filter?: Filter, options?: Options): Promise<T>;
```
should become

```js
findById(id: ID, filter?: Filter, options?: Options): Promise<T | null>;
```

hacksparrow:
> Modify controller template & examples to handle "not found" error by leveraging the two mechanisms described above

So, if we use `getById()`, our code will look like this:

```js
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    return await this.daisyRepository.findById(id);
  }
```

and if we use `findById()`, it looks something like this:

```js
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    let result =  await this.daisyRepository.find(filter);
    if (result === undefined || result === null) { 
             throw new HttpErrors. NotFound(`id ${id} was not found`);
       } 
    return result;
  }
```

or

```js
@get('/daisies/{id}')
  async findById(@param.path.number('id') id: number): Promise<Daisy> {
    return callThe404Helper(await this.daisyRepository.findById(id));
  }
```

Is my understanding correct?

bajtos:
@hacksparrow here is an example from the code I am working on:

```ts
export class TodoController {
  // ...

  @get('/todos/{id}')
  async findTodoById(
    @param.query.boolean('items') items?: boolean,
  ): Promise<Todo> {
    const maybeResult = await this.todoRepo.findById(id);
    if (maybeResult) return maybeResult;
    throw new EntityNotFoundError(Todo, id, {statusCode: 404});
  }
}
```

bajtos:
I reworked my pull request per my previous comment. I am keeping the original code in the git history for posterity, I'll rebase the first two commits away before landing this pull request.

@raymondfeng @hacksparrow @marioestradarosa @joeytwiddle please review

marioestradarosa:
@hacksparrow **You analysis is good and valid**,  but let me share with you that we are implementing some batch processes that will run periodically from unix cron. 

We are using the context and the Juggler from **LB4** as they are independent packages, for this scenario developers are used to check the result for null or undefined and it is not related to controllers.

 The code in the controller  as @bajtos  is proposing below makes sense for keeping both worlds happy.  It is repetitive ?, yes but it is generated and developers can changed it if they want to, however, I don't see any reason so far why they should remove this `throw new EntityNotFoundError` as it is related to web client apps in need of it. 

Wrapping it up inside the internal libraries doesn't allow this flexibility mentioned above.

```ts
if (maybeResult) return maybeResult;
    throw new EntityNotFoundError(Todo, id, {statusCode: 404});
  }
```

hacksparrow:
@bajtos this looks like a sensible solution, LGTM.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -206,16 +206,27 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
     );
     if (!model) {
-      throw new Error(`no ${this.modelClass.name} found with id "${id}"`);
+      // entity instance not found
+      return undefined;
     }
     return this.toEntity(model);
   }
 
+  async getById(id: ID, filter?: Filter, options?: Options): Promise<T> {
```
raymondfeng: While I fully understand the motivation to add `getById`, I do have a few concerns:

- `getById` is not a typical method name for CRUD as `get` is usually for accessors or REST `GET`.
- Adding a method with similar function as `findById` might bite us in the long run as we see in LB 3.x. 

I suggest that we add a `helper` method such as `assertFound(entityOrUndefined: Promise<T | undefined>): Promise<T>` instead. 

Case 1: We're sure `c01` exists.
```ts
const customer = await customerRepo.findById('c01');
console.log(customer!.name);
```

Case 2: Convert `undefined` to NotFoundError
```
const customer = await assertFound(customerRepo.findById('c01'));
console.log(customer.name);
```
Case 3: Check the result
```
const customer = await customerRepo.findById('c01');
if (customer) console.log(customer.name);
else throw new Error(...);
```

joeytwiddle: The `assertFound()` helper is concise. I like that.

My concern (which I also raised in my review) is that sleepy developers might simply forget to use the assertion (or the correct method call) and so end up with runtime errors about undefined customers.

So I guess my preference would be to make `findById()` _throw by default_, and have users opt-in to the allow-undefined behaviour. That would make for stricter/tighter code. But if you have rejected that path for sure, then ok never mind me!

One other thing: It might be difficult to translate `assertFound()` over to other functions like `updateById()` and `deleteById()`, if we want to give them a similar and consistent mechanism.

virkt25:
Is there any reason to not consider a flag on the model such as `strictFind` that can be leveraged by the `juggler-bridge` to throw an error or not for the `null` case which keeps the config clean and if the return is an error the sequence should be able to handle that.

examples/todo-list/src/controllers/todo-list.controller.ts
```diff
@@ -41,7 +46,9 @@ export class TodoListController {
 
   @get('/todo-lists/{id}')
   async findById(@param.path.number('id') id: number): Promise<TodoList> {
-    return await this.todoListRepository.findById(id);
+    const maybeResult = await this.todoListRepository.findById(id);
+    if (maybeResult) return maybeResult;
+    throw new EntityNotFoundError(TodoList, id, {statusCode: 404});
```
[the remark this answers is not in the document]
bajtos: Good catch!

I tend to prefer reverse logic when the happy path is short and error handling requires several lines of code. I think I started with such code initially, but then simplified the error path to a single line.

I agree the code should be re-ordered per your suggestion.

virkt25:
Thanks for such a detailed discussion summary and comment @bajtos !! 

> From what I've seen so far, different callers need different behavior. A single model-level flag cannot provide us that feature.

Makes sense. 

---

Hmm, this is a tough one. I think the biggest thing here will be to pick an approach using our tooling for the templates as that's the default most people will stick with IMO unless they have different requirements ... so I'm thinking what would be easier to deal with for a user looking to modify what the CLI generated. 

I like both the proposed solutions of either always throwing or defaulting to a given value OR introducing a new `OrThrow` option. 

### OrThrow
**PRO**
- Simple, distinct UX and each method has a clear, distinct intent
**CON**
- User may not be aware of the other method ... depending on the default. 

### default option
**PRO**
- Only one API for a user to interact with
- Might be easier to test as we can allow a default field to be passed into the controller to be passed on
**CON**
- Explaining a `default` being able to override an error seems weird (but maybe it's just me)
- Multiple intents / return types possible from the `default` property. 

I can sort of make an argument for both in my mind but I think I would personally like to see two different methods to separate the concerns at a first glance. But something is definitely to be said about keeping it simple by only having a single API which is super flexible and as a result better / easier. Either way I think documentation around this will become key!

Want to hear what others think of the options ... hopefully someone can convince me to lean more one way or the other.

hacksparrow:
**A reminder**

Any change we want to make to `findById()` will require change in `EntityCrudRepository`'s `findById()` too, because `DefaultCrudRepository` implements `EntityCrudRepository`. Anything else implementing `EntityCrudRepository` will be affected, whether it is required for them or not.

---

@bajtos The above something to consider for `await repo.findById('id-does-not-exist', {default: undefined});`.

Can user set `default` to a function, a `Date`, etc.? Or do we have a restriction to what those values can be?

If `default` can be set to anything, we will end up with something not very TS-like and a change with complicated side-effects.

hacksparrow:
> From that perspective, it may be ok to keep findById as-is and introduce findByIdOrThrow as a new method to be used in our CRUD Controller template. Once the users are familiar with findByIdOrThrow, I think the other method findById should be easy to discover.

My personal preference would be for `*ById` to throw by default, and introduce a new method which doesn't.

Since we are in `async`-`await` context, throwing should be perceived as rejecting the promise, not the conventional error throwing. Resolving a promise with something else other than what I queried for feels illogical.

hacksparrow:
```ts
findById('does-not-exist'); // returns a rejected promise

findByIdWithDefault('id-does-not-exist'); // returns promise resolved as `undefined`
findByIdWithDefault('id-does-not-exist', undefined); // same as above

findByIdWithDefault('id-does-not-exist', new Product({name: 'Pencil'}));
// returns a promise resolved with the given product instance
```

Lovin' it!

Btw, we'll have to do it for all `*ById` methods.

raymondfeng:
The fact that we cannot reach an agreement on the API style is a warning sign that we try to do too much for something that doesn't have a consistent expectation from the API consumers.

One lesson I learned from LB 2.x/3.x is that we try to provide too many variants of the CRUD APIs. Starting from LB4, I was trying hard to trim down the number of methods in base Repositories to fully reflect the common operations that most developers need and reasonably supported by various databases (for example, `findById` can be possibly optimized by a DB than a regular `findOne` and `upsert` can be atomic). Application developers now have the flexibility to extend our base Repositories to add more flavors of operations. In this context, I propose to keep it simple as follows:

1. Fix findById() to be `findById(id: ID, filter?: Filter, options?: Options): Promise<T | undefined>;`. For TypeScript developers, the TSC compiler forces them to deal with `undefined`. But if we choose to throw an error, developers are not obligated to catch and handle it gracefully.

2. Leave `undefined` to `404` mapping to controllers.

3. No magic via `options.throwIfNotFound` or `options.default`. For example, `options.default` is often not very useful if it only accepts a constant value. 

4. Document how application developers can extend base Repositories to provide their own variants of the CRUD APIs and helper methods if necessary.

bajtos:
@raymondfeng 

> The fact that we cannot reach an agreement on the API style is a warning sign that we try to do too much for something that doesn't have a consistent expectation from the API consumers.

I think the expectation is pretty consistent here. We have two use cases to support and each one has different requirements from the API (see https://github.com/strongloop/loopback-next/pull/1658#issuecomment-417568919).

- Consumers that WANT to handle "not found" situation explicitly. For example, to throw HTTP Not Found error from a Controller method.
- Consumers that DO NOT WANT to handle "not found" situation explicitly and want the framework to throw a descriptive error. For example, in our tests, we are pretty sure the model instance exists, so we don't want to deal with "not found" case in every test. At the same time, if something goes wrong, we want the test to fail with a descriptive error. An example of a helpful failure: `Cannot find User id 123`. An example of an unhelpful message, produced e.g. after using TypeScript's `!` operator to silence null checks: `TypeError: Cannot read property 'email' of undefined`.

> One lesson I learned from LB 2.x/3.x is that we try to provide too many variants of the CRUD APIs. Starting from LB4, I was trying hard to trim down the number of methods in base Repositories to fully reflect the common operations that most developers need and reasonably supported by various databases

I agree that in LB 2.x/3.x, we were providing too many variants of CRUD APIs, for example `findOrCreate`, `findOrCreateWithWhere`, `updateOrCreate`, `replaceOrCreate`. From my experience, the problem was in the fact that the implementation of each variant was pretty involved, each one was e.g. executing Operation Hooks with a different shape of the `context` object, performing validations or not, etc.

However, I find the situation here different. We have a single CRUD API variant, "find a model with the given id", and need to find a method signature that will be easy to use. 

If we were in Java or .NET world, I would go for a method returning `Optional<Model>`. Because the `Optional` (`Maybe`) design pattern does not map well to JavaScript, we need to look for a different solution.

I am proposing to have two methods, e.g. `findById` and `findByIdOrThrow`, where one is implemented as a thin wrapper around the other one. I consider this as very different from the problem of different CRUD variants described above and therefore don't see how the difficulties of supporting too many different CRUD variants are relevant to my proposal.

> Application developers now have the flexibility to extend our base Repositories to add more flavors of operations.

We need both flavors (throw when not found, return undefined when not found) in our Todo and TodoList example applications. Now I could implement a custom Repository adding `findByIdOrThrow` method in our example applications, it would it look silly to implement `getById` in every repository-based example app we are writing? To me, this is a clear sign that the base Repository API needs to support both API flavors out of the box.

> Leave undefined to 404 mapping to controllers.

I am fine with that, as long as we are ok with the code bloat introduced to every controller class, see https://github.com/strongloop/loopback-next/pull/1658#discussion_r215121384

> Must we do this in every controller?
>
> I thought the idea was to use `getById()` and then, if it throws a MODEL_NOT_FOUND error, it would be automatically converted to a 404 **in the sequence**.
>
> Or is that planned for a later iteration?
>
> That would reduce controller bloat. This function would just have one line.

Perhaps we can write a short documentation explaining how to move 404 handling from individual controller methods to the Sequence?

raymondfeng:
If we decide to add `findByIdOrThrow`, we should also add `updateByIdOrThrow` and `deleteByIdOrThrow`. The number of methods will grow quickly with these variants.

bajtos:
> If we decide to add `findByIdOrThrow`, we should also add `updateByIdOrThrow` and `deleteByIdOrThrow`. 

I agree that we should add those other methods too. Maybe `deleteByIdOrThrow` is not necessary, because when the model was not found, then it means it no longer exists and the intention to delete it has been fulfilled. 

> The number of methods will grow quickly with these variants.

That is true. What other options are there? The discussion above mentions:
 - `Optional` return type. It does not map well to JavaScript and I also don't see how it would apply to `updateById` and `deleteById`.
 - A new argument `options`, with a property like `default` or `throwIfNotFound` allowing the developer to pick the error handling mode they prefer. This has been rejected: _No magic via `options.throwIfNotFound` or `options.default`. For example, `options.default` is often not very useful if it only accepts a constant value._


As as I see it, we have to make a trade-off here and accept that neither solution is going to be perfect.

 - A single `findById` method, always returning `undefined` when the model was not found. Downside: test code has to explicitly throw an error or rely on not very helpful TypeError.
 - Two methods `findById` and `findByIdOrThrow`. The downside: the number of repository methods can grow too much.
 - A single method `findById` with an `options` object controlling how "not found" is handled. The downside: @raymondfeng considers this as magic that should be avoided.

As I wrote the list above, I actually like the last variant based on `options` object most.

It is following the pattern we are already using in Context, where one can use `ctx.get(KEY)` or `ctx.get(KEY, {optional: true})`, where the first one throws on not found and the second returns `undefined` when not found. 

https://github.com/strongloop/loopback-next/blob/13e8f99402c9b6f43a61a30acd2e7237e20d468b/packages/context/src/context.ts#L258

https://github.com/strongloop/loopback-next/blob/13e8f99402c9b6f43a61a30acd2e7237e20d468b/packages/context/src/context.ts#L281-L284

It is also easy to apply to other methods like `updateById` and can be easily extended in the future by adding more options properties (as opposed to more explicitly named methods like `findByIdOrThrow`).

hacksparrow:
Or can we have a property in the app context, which configures how `*ById` behaves?

bajtos:
In a chat with @raymondfeng @hacksparrow and @virkt25, we proposed the following way forward:


(1) `findById` is going to throw if not found.

(2) We will add a factory method allowing people to obtain a modified Repo instance that returns  if not found. The mockup bellow illustrates the idea, we need to find a better name than "relaxed":

```ts
const repo = new ProductRepository(Product, db);
// findById throws by default

const repo = new ProductCrudRepo(repo); 
const relaxedRepo: RelaxedCrudRepository<ProductRepository> 
  = repo.createRelaxedRepo(); // find a better name!
// relaxedRepo.findById(id) -> returns undefined not found
```

We should also update `updateById`,`deleteById(id)` for consistency.

(3) Add a type-guard for detecting whether an error is an `EntityNotFoundError` instance, to make it easier for consumers to detect the "not found error" situation.

(4) Keep controller methods simple, just `return this.repo.findById()` and iImplement the status code lookup based on `err.code` in `reject` action as was shown in the first commit 6891c34c9.

bajtos:
I have reworked the pull request to apply the direction outlined in the previous comment. 

To get these important changes landed sooner, I decided to leave the following feature out of scope of this pull request:

- Update `updateById` and `deleteById` to throw `EntityNotFound` error for consistency with `findById`.
 - Add a factory method allowing people to obtain a modified Repo instance that returns undefined if not found.

My intention is to start working on these two items as soon as this initial pull request is merged.

As far as I am concerned, this pull request is ready for final review and landing. The changes are split into 5 commits, the review may be easier if you follow commit-by-commit.

@raymondfeng @strongloop/sq-lb-apex @hacksparrow @marioestradarosa @joeytwiddle PTAL, LGTY now?

bajtos:
> coverage/coveralls — Coverage decreased (-0.04%) to 89.569%

The decrease is caused by my changes in `CrudRepositoryImpl<T>` which does not come with any test coverage from the beginning. Please note this repository is not used anywhere. Our templates, examples and docs are using juggler-backed `DefaultCrudRepository` which does have test coverage.

packages/repository/src/repositories/repository.ts
```diff
@@ -250,16 +251,18 @@ export class CrudRepositoryImpl<T extends Entity, ID>
...
-    return this.connector
-      .find(this.model, {where: where}, options)
-      .then((entities: T[]) => {
-        return entities[0];
-      });
+    const entities = await this.toModels(
+      this.connector.find(this.model, {where: where}, options),
+    );
+    if (!entities.length) {
+      throw new EntityNotFoundError(this.model.name, id);
```
b-admike: can we leverage `this.model.modelName()` here?

packages/rest/test/unit/router/reject.provider.unit.ts
```diff
@@ -28,6 +28,18 @@ describe('reject', () => {
     expect(result).to.have.property('statusCode', 500);
   });
 
+  it('converts error code ENTITY_NOT_FOUND to status code 404', async () => {
+    const reject = new RejectProvider(noopLogger).value();
+
+    const notFoundError: Error & {code?: string} = new Error('not found');
```
b-admike: I thought the `Error` type def would have `code` key in there. Maybe it's not optional and that's why we are defining the type of `notFoundError` like this?

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -182,6 +182,10 @@ children:
       url: Decorators_repository.html
       output: 'web, pdf'
 
+  - title: 'Error handling'
+    url: Error-handling.html
```
b-admike: The `.md` file is called `Errors-handling` whereas here we have `Error-handling.html`. I think the better option is to rename the `.md` file to align with this and the title of the page.

packages/rest/src/rest-http-error.ts
```diff
@@ -7,12 +7,22 @@ export namespace RestHttpErrors {
...
+      },
+      extraProperties,
+    );
   }
 
   export function missingRequired(name: string): HttpErrors.HttpError {
     const msg = `Required parameter ${name} is missing!`;
-    return new HttpErrors.BadRequest(msg);
+    return Object.assign(new HttpErrors.BadRequest(msg), {
+      code: 'MISSING_REQUIRED_PARAMETER',
```
b-admike: Can we add these new error codes to the mapping you created for the `404` case?

docs/site/sidebars/lb4_sidebar.yml
virkt25: +1. Please rename `docs/site/Errors-handling.md` => `docs/site/Error-handling.md` for consistency even though this will render fine as-is (or should) since the permalink in the file is correct.

docs/site/Errors-handling.md
```diff
@@ -0,0 +1,29 @@
+---
+lang: en
+title: Error Handling
+keywords: LoopBack 4.0
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Error-handling.html
```
virkt25: Rename this file for consistency please.

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
```diff
@@ -81,8 +82,8 @@ export class <%= className %>Controller {
       },
     },
   })
-  async findById(@param.path.<%= idType %>('id') id: <%= idType %>): Promise<<%= modelName %>> {
-    return await this.<%= repositoryNameCamel %>.findById(id);
+  findById(@param.path.<%= idType %>('id') id: <%= idType %>): Promise<<%= modelName %>> {
```
virkt25: I'm confused about the need for this change. All other controller methods are still using async/await. Why does this one need to be treated differently?

packages/repository/src/repositories/repository.ts
```diff
@@ -250,16 +251,18 @@ export class CrudRepositoryImpl<T extends Entity, ID>
     return this.toModels(this.connector.find(this.model, filter, options));
   }
 
-  findById(id: ID, options?: Options): Promise<T> {
+  async findById(id: ID, filter?: Filter, options?: Options): Promise<T> {
```
virkt25: What filter would `findById` method possibly take? I think the interface should be updated to remove this method instead of adding it to the implementation. 

If it is supposed to be here then can you add the `@param` annotation to the interface since it is missing.

packages/repository/src/errors/entity-not-found.error.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const quotedId = JSON.stringify(entityId);
+
+    super(`Entity not found: ${entityName} with id ${quotedId}`);
+
+    Error.captureStackTrace(this, this.constructor);
+
+    this.code = 'ENTITY_NOT_FOUND';
+    this.entityName = entityName;
+    this.entityId = entityId;
```
virkt25: Should we set this to `quotedId` or not?

packages/rest/test/unit/router/reject.provider.unit.ts
virkt25: `code` isn't part of a standard Node `Error` object but something we add to it ourselves IIUC.

packages/repository/src/repositories/repository.ts
bajtos: I am going to fix the type of the constructor argument `model` and use `this.model` (let EntityNotFoundError to obtain the model name).

packages/repository/src/repositories/repository.ts
hacksparrow: I am also curious about the filter for `findById`.

packages/rest/src/providers/reject.provider.ts
```diff
@@ -9,6 +9,12 @@ import {HttpError} from 'http-errors';
 import {RestBindings} from '../keys';
 import {writeErrorToResponse, ErrorWriterOptions} from 'strong-error-handler';
 
+// TODO(bajtos) Make this mapping configurable at RestServer level,
+// allow apps and extensions to contribute additional mappings.
```
hacksparrow: > allow apps and extensions to contribute additional mappings.

Allow add, disallow edit and delete.

packages/rest/test/unit/router/reject.provider.unit.ts
bajtos: As @virkt25 wrote, JavaScript `Error` class does not have any `code` property by default.

hacksparrow: Yes, no `code` in JavaScript `Error`.

docs/site/sidebars/lb4_sidebar.yml
bajtos: Thank you for spotting this. I think I started with `Errors.md`, then decided to go with "Error Handling" and forgot to switch the first word back to singular.

I am going to apply your suggestion.

hacksparrow:
Looks pretty good to land.

packages/rest/src/rest-http-error.ts
bajtos: These errors are created inside REST layer, where we know about HTTP status codes and thus `error.statusCode` is already set. No need to set `statusCode` in the `reject` action. In fact, the mapping preserves `statusCode` if it's already set, and uses the lookup table only when no status code was provided by the error. Adding these new error codes to the lookup table would be a no-op that may confuse users IMO.

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
bajtos: +1 to be consistent.

I am going to revert this change and keep the original `async` method with `return await` statement.

packages/repository/src/repositories/repository.ts
bajtos: If you have a `Customer` model that has many `Order` instances, then you can fetch a customer with all their orders.

```
GET /api/customers/123?filter[include]=orders
```

AFAIK, juggler supports only a subset of Filter properties. For example, `fields` and `include` are most likely supported, but `where` is not.

On the second thought, using the `Filter` type was a bad idea, we should use a different type that describes only the properties applicable to `findById`. Let's leave this change out of this initial pull request. I opened a new issue to keep track of it, see https://github.com/strongloop/loopback-next/issues/1721

packages/repository/src/errors/entity-not-found.error.ts
bajtos: No. The purpose of `quotedId` is to have a value that's easy to parse in a string, to make it easy for readers to distinguish between e.g. a number `1` and a string `"1"`.

Error properties like `entityId` and `entityName` can be serialized in different ways in the HTTP response. When using JSON, the distinction between different types (`undefined`, `null`, a number, a string, an object/array) is preserved.

packages/repository/src/repositories/repository.ts
bajtos: Never mind. The `filter` argument of `findById` has been added as part of the initial repository implementation in e92cc244a.

Removing `filter` argument from `findById` signature is a change not related to handling "not found" situation.

I think this particular source code line is highlighted by git because previously, the default repository implementation (that nobody uses) did not have `filter` argument and as I was fixing `findById` signatures, I made the signature consistent with the rest.

I'll add a minimal `@param` annotation to the interface then.

bajtos:
Landed 🎉

Issue #1588, opened 2018-08-02
[CLI] add `lb4 repository` command


hacksparrow:
## Description / Steps to reproduce / Feature proposal
`lb4 repository` command will generate the basic repository file for a given name. It is better than copy-pasting and editing files.

## Current Behavior
There is no `lb4 repository` command.

## Expected Behavior
We should have `lb4 repository` command.

## Acceptance Criteria

- [ ] Add `lb4 repository` command to `@loopback/cli`. Works as follows:
  - [ ] DataSource Selection (Detect in project using https://github.com/strongloop/loopback-next/blob/master/packages/cli/lib/utils.js#L210). (Infer binding key based on Boot Conventions).
  - [ ] Detect the type of Repository (DefaultCrudRepository / KVRepository) based on the DataSource (No prompt).
  - [ ] Model Selection (Detect in project using https://github.com/strongloop/loopback-next/blob/master/packages/cli/lib/utils.js#L210). This will be a multiple choice question allowing a user to select multiple models to back to same DataSource. We will generate a new class for each selected model.
  - [ ] Prompt for type of `id` field in Model ... if we can't infer by inspecting the `Model` class ourselves (Ex: Check if `id` property exists and set as `typeof Todo.prototype.id` in model or check metadata for which field is the `id` ... if both fail then prompt the user.

  - [ ] Use templates to generate the appropriate repository artifact as `/src/repositories/<modelName>.<dataSourceName>repository.ts`
- [ ] Tests + Docs
  - [ ] Ensure `--config` mode is supported (Can pass in answers to all prompts using JSON).
- [ ] Blog

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

virkt25:
- Should allow selection of datasource from available datasources, model from available models and then the type of repository (CRUD, relation, etc).

marioestradarosa:
Relates #1359 . This feature is complex but needed.  It will have to generate the appropriate repository and the controller based upon a relationship (if it exists). Perhaps two PRS, a simple one (_without relations_) first?

bajtos:
Instead of adding a new `lb4` command, I am proposing to extend `lb4 model` to ask the user whether they want to create a repository to accompany the new model. I expect most models to be created together with a repository, let's make this most common scenario easy to follow.

Prompts to ask:
 - Would you like to create a Repository class for your model? (Y/n)
[...]

PR #1659, opened 2018-08-30
feat(cli): lb4 repository implementation

marioestradarosa:
**Done**
- **Filtering** out the _soap and rest connectors_ from the data source list
- Inferring the repository class (**KV** or **CRUD**) based on the connector property for JSON datasource
- Allow _multiple selection_ of models 
- Infers the **id property name** from the model Class based on _3 scenarios_
   - The current template generated by `lb4 model` command 
   - A JSON object inside the @model decorator
   - A JSON definition inside the class model
- **Prompts for the property name i**n case the infer process _couldn't find the primary key property_
- Currently it is generating the repository based on the two repository types (KV or CRUD)
- support for --config is inherited from the base generators 
- properly display an error message if it can't fine a src/datasources or src/models directory
- Add basic documentation for the feature
- Use name argument as the repository name
- add shared functions in utils to define the path and artifact names for repositories and datasources
- fixed lib/ArtifactGenerator to support multiple index files

**Command line feature**
- Now we can supply the arguments from the command line. The arguments are checked against all the artifacts implemented already (ie: id inferring list, datasource list, model list) and if not valid, it falls back to the prompt. 

**Syntax**: _Use them interchangeably_
`lb4 repository  name  --datasource --model  --id`

**Example:** 
`lb4 repository  --datasource dbmem --model=pluto --id=IDPK`

It will generate the repository for pluto with id property name IDPK. It is checked if dbmem is a valid datasource, pluto (changed to Pluto) is a valid model and if its id can't be inferred then it falls back to use the id argument supplied. All validations are performed against the list of artifacts already loaded form their respective directories.

### Test Status
- [x] Test on Default generated models  using CRUD connector
- [x] Test on models defined by @model decorator using CRUD connector
- [x] Test on Default  generated models using KV connector
- [x] Test on models defined by @model decorator using KV connector
- [x] Test on failing providing or selecting a model. 
   - at this point the user should have selected a valid datasource, otherwise this point will never get reached
- [x] Test on multi-word models
- [x] Test on command line interface --model --id from a none-id inferred model file
- [x] Test failing  when the datasource list is empty.
   -  the datasources are not KV or CRUD, and thus it returns an empty list
   -  or the datasource directory is empty
- [x] Add test for supporting config file 
- [x] Add test for support of name argument
- [x] Add test for multiple repository generation

### Preconditions checked
check the following directories in this order

- check if the src/datasources exists, displaying an approriate error
- check if the src/models exists 

Implements #1588

## Checklist


marioestradarosa:
render1535648525827

marioestradarosa:
**# lb4 repository**

render1535696680437

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,294 @@
...
+
+module.exports = class RepositoryGenerator extends ArtifactGenerator {
+  // Note: arguments and options should be defined in the constructor.
+  constructor(args, opts) {
+    super(args, opts);
+
+    /** instance helper method isolated from the execution loop
+     * @connectorType: can be a single or a comma separated string list
+     */
+    this.isConnectorType = async function(connectorType, dataSourceClassName) {
```
bajtos: Helper methods should be defined as regular class methods and their name should start with `_`:

```ts
module.exports = class RepositoryGenerator extends ArtifactGenerator {
  constructor(args, opts) {
    // ...
  }

  private async _isConnectorType(connectorType, dataSourceClassName) {
    // ...
  }
}
```

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,294 @@
...
+    } else {
+      // iterate thru each selected model, infer or ask for the ID type
+      for (let item of this.artifactInfo.modelNameList) {
+        debug(`iterating thru ${item}`);
+        this.artifactInfo.modelName = item;
+        const prompts = [
+          {
+            type: 'list',
+            name: 'idType',
+            message: `What is the ID type of ${item} `,
```
bajtos: Can you use `typeof ${model}.prototype.id`? See e.g. TodoRepository: https://github.com/strongloop/loopback-next/blob/ac011f80b229b5dd5c69851ab14b39705b852629/examples/todo/src/repositories/todo.repository.ts#L10-L13

In the future, we may want to ask the user for the _name_ of the id property to use.

Regardless of that, I think we should not be duplicating the id type name in the repository class if possible.

What do you think?

packages/cli/generators/repository/templates/src/repositories/repository-crud-default-template.ts.ejs
```diff
@@ -0,0 +1,12 @@
+import {<%= repositoryTypeClass %>}, juggler} from '@loopback/repository';
+import {<%= modelName %>} from '../models';
+import {inject} from '@loopback/core';
+
+export class <%= className %>Repository extends <%= repositoryTypeClass %><
+  <%= modelName %>, <%= idType %>> {
```
bajtos: I think this is not going to pass `prettier` linter and should be changed to the following:

```ejs
export class <%= className %>Repository extends <%= repositoryTypeClass %><
  <%= modelName %>, 
  <%= idType %>,
> {
```

bajtos:
> we have two type of locations for the model
> 
> - A JSON definition file used as parameter to the @model decorator
> - A TS class with properties
> 
> I am trying to find an reflection solution to inquire the TS class property or ultimately go with a simple read of the file. For the first option (JSON definition file) is easy since it is a JSON file. Once that is covered, then we will go with the test.

For the initial implementation, I am proposing to use the same solution as `lb4 controller` is using.

https://github.com/strongloop/loopback-next/blob/ac011f80b229b5dd5c69851ab14b39705b852629/packages/cli/generators/controller/index.js#L105-L108

AFAICT, you are already doing that :+1:

packages/cli/generators/repository/index.js
marioestradarosa: yes, I will use  `typeof ${model}.prototype.id` if the model has the id field and if it is set to the primary key.  To do that, I am trying to see the best way to do it, probably just read the file and inspect it or use ATS to analyze it more in depth and even get the id name from it.

thoughts?

marioestradarosa:
@bajtos I addressed your suggestions 👍  thanks.
 
PTAL

marioestradarosa:
render1535961909485

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+'use strict';
+const _ = require('lodash');
+const ArtifactGenerator = require('../../lib/artifact-generator');
+const debug = require('../../lib/debug')('repository-generator');
+const inspect = require('util').inspect;
+const path = require('path');
+const chalk = require('chalk');
+const utils = require('../../lib/utils');
+const connectors = require('../datasource/connectors.json');
+const tsquery = require('@phenomnomnominal/tsquery').tsquery;
```
bajtos: I am afraid `tsquery` is not properly licensed as open-source, which means we cannot add it to our dependencies.

I opened an issue in the project repository - see https://github.com/phenomnomnominal/tsquery/issues/25, hopefully the project maintainers will fix it soon.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+    const nodes = tsquery(ast, ':matches([kindName="PropertyDeclaration"])');
+
+    for (let x of nodes) {
+      retArray.push(x.name.escapedText);
+    }
+
+    return retArray;
+  }
+
+  async _getModelIdType(modelName) {
```
bajtos: Could you please move the code detecting the "id" property into as standalone helper file that can be used by other generators? It will also make it easier to test this functionality.

A few non-obvious test cases that come to my mind:

The id property is defined via `@model` definition:

```ts
@model({
  name: 'Product',
  properties: {
    id: {type: 'number', id: true}
    name: {type: 'string', required: true},
  },
})
class Product extends Entity {
  id: number;
  name: string;
}
```

The model does not use any decorator:

```ts
class Product extends Entity {
  static definition = {
    name: 'Product',
    properties: {
      id: {type: 'number', id: true}
      name: {type: 'string', required: true},
    },
  };

  id: number;
  name: string;
}
```

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+    try {
+      jsonFileContent = this.fs.readJSON(datasourceJSONFile, {});
+    } catch (err) {
+      debug(`${ERROR_READING_FILE} ${datasourceJSONFile}: ${err.message}`);
+      return this.exit(err);
+    }
+
+    for (let connector of Object.values(connectors)) {
+      if (
+        jsonFileContent.connector === connector.name &&
```
bajtos: I believe we allow two options how to specify a connector used by a datasource, at least that's the case of LB 3.x CLI and both LB 3.x/4.x runtime:

1. Using the full name, e.g. `loopback-connector-mysql`.
2. Using a short name without `loopback-connector-` prefix, i.e. `mysql`.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+      let result = await this._isConnectorOfType(
+        VALID_CONNECTORS_FOR_REPOSITORY,
+        item,
+      );
+      debug(
+        `${item} has keyword ${VALID_CONNECTORS_FOR_REPOSITORY} is ${result}`,
+      );
+      if (!result) {
+        // remove from original list
+        _.remove(datasourcesList, e => e == item);
```
bajtos: Please rewrite this look using `Array.prototype.filter` function.

```ts
const availableDatasources = datasourcesList.filter(item => {
  const result = await this._isConnectorOfType(
     VALID_CONNECTORS_FOR_REPOSITORY,
     item,
   );
   debug(/*...*/);
   return result;
});
```

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+      }
+    }
+
+    if (_.isEmpty(datasourcesList)) {
```
bajtos: How about using `!datasourcesList.length`, since we know the list always contain an array value?

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+        this.artifactInfo.modelName = item;
+        let modelPropertyList = await this._getModelFields(item);
+
+        const prompts = [
+          {
+            type: 'list',
+            name: 'propertyName',
+            message: `Please, select the ID property for ${item} from the list `,
+            choices: modelPropertyList,
+            default: modelPropertyList[0],
```
bajtos: Can we perhaps leave AST-based detection of model properties out of the first version of `lb4 repository` to simplify the initial implementation and get it landed sooner?

I think the following prompt should be good enough for now:

```
const prompts = [
  {
    type: 'input',
    name: 'propertyName'
    message: `Please enter the name of the ID property of ${item}:`,
    default: 'id',
  },
];
```

Thoughts?

cc @raymondfeng @virkt25

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,393 @@
...
+
+        let idType = await this._getModelIdType(item);
+
+        if (idType === 'none') {
+          const answer = await this.prompt(prompts);
+          this.artifactInfo.idType = `typeof ${item}.prototype.${
+            answer.propertyName
+          }`;
+        } else {
+          this.artifactInfo.idType = idType;
```
bajtos: Please use `typeof ${item}.prototype${idType}` in this branch too.

The benefit is that when people rename models and/or change the `id` property type in their application later in the future, the Controller code will automatically track those changes.

When you put a hard-coded ID type in the generated controller source, users have to update that ID type when it changes, which can be tedious.

packages/cli/package.json
```diff
@@ -39,9 +39,11 @@
     "sinon": "^4.5.0",
     "yeoman-assert": "^3.1.1",
     "yeoman-environment": "^2.0.6",
-    "yeoman-test": "^1.7.0"
+    "yeoman-test": "^1.7.0",
+    "typescript": "^3.0.1"
```
bajtos: I am curious, what do we need `typescript` for? Why is it a direct dev-dependency?

packages/cli/generators/repository/index.js
marioestradarosa: Nice, they added the License already 👍

packages/cli/generators/repository/index.js
marioestradarosa: Yes, that's true, there are many variations. One that is a little bit more complex is when you import the JSON definition file from somewhere else (ie: from `@yourcompany/models`) . I believe we should tackle the most common scenarios and then fine tune it to land the PR?

packages/cli/generators/repository/index.js
marioestradarosa: So we should ask for both, short as of now and the longer version right?

packages/cli/generators/repository/index.js
marioestradarosa: @bajtos , sure, I will update it, but what would be the benefit ? .

packages/cli/generators/repository/index.js
bajtos: >  I believe we should tackle the most common scenarios and then fine tune it to land the PR?

Yes please!

Just make sure there is a workaround solution for people using an uncommon scenario. I think your current implementation, where the user can enter a custom value if the AST-based reader did not find anything, is a great solution for that 👍

packages/cli/generators/repository/index.js
marioestradarosa: I used `_.isEmpty` because it appears in the generators, but again what would be the benefit, is there an overhead for the lodash method?

packages/cli/generators/repository/index.js
marioestradarosa: Even if they already updated the license in their repo, shall we remove the AST-based detection for this PR ?

packages/cli/generators/repository/index.js
bajtos: The prompt for a connector to use is handled by `lb4 datasource` command, no need to change that.

I am saying that you should improve this check to correctly handle both long and short form of the `connector` property specified in the datasource configuration.

E.g.

```ts
const connector = jsonFileContent.connector;

if (
  connector === connector.name ||
  `loopback-connector-${connector}` === connector.name
  // ...
) {
}
```

Also please refactor this multi-line `if` condition to fit into a single line, see [Indentation of multi-line expressions in if](https://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if), the example called "Best".

packages/cli/package.json
marioestradarosa: I am not  sure if it is my machine only. I ran npm install `@phenomnomnominal/tsquery --save` only, and it complained about the `typescript` missing. They just specify this dependency for dev only. But in runtime, they still need it. The only way to get rid of this error was to place the `typescript` dependency in our CLI package.json .

packages/cli/generators/repository/index.js
bajtos: The code is easier to reason about because you are not modifying the array items in the middle of the loop iterating through it, and you also don't need `_.remove` from an external library.

marioestradarosa:
@bajtos , 

I created the helper library and discovered a better way to work with ts query. We just define the patterns and can really inspect the code in the TS class. so I refactored into a single method exposed in the helper library. The method now works and identifies the three scenarios as follow:

- The default model generated by `lb4 repository`
- A JSON definition passed to the @model decorator
- A static definition inside the class 

It fallbacks to ask for the property name for the ID field.  I believe the tsquery looks promising, with a little coding and the pattern constants we can create more scenarios easily, and take into account that I am not expert on the library, so the library can serve very well for future updates such as relations if we get acquainted to its concept.

The multi-line IF statement is saved in that format from my VS Code automatically, I have the two plugins specified in the loopback4 developer manual.

The constans in the helper library are strings concatenated by + symbol, I couldn't get a better way to keep them in the appropriate width.

Let me know if it is feasible to still use this tsquery, the other alternative is to use a parser to make a tree and inspect it. But I don't think supporting the 3 mentioned scenarios would be easy.

marioestradarosa:
I forgot, the primary key discovered can be any name. ID,  DATAPK, id , anything. inside the code you will see that we ask for id but that's is just the id: true property.

packages/cli/package.json
marioestradarosa: Also, I found in their tests they import some constants also from typescript, especially to identify the type of node etc.

marioestradarosa: @bajtos , they answered my issue and will fix it today. They will apply typscript as a `peerDependcy`  rather than the current scenario `devDependency`.

hacksparrow:
How about just `--model` instead of `--modelName`, and `--id` instead of `--idName`?

marioestradarosa:
> How about just --model instead of --modelName, and --id instead of --idName?

@hacksparrow thank for your observation. I think it is better this way 👍,  changes have been done to these two argument names.

marioestradarosa:
@bajtos PTAL

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,398 @@
...
+    return result;
+  }
+
+  _setupGenerator() {
+    this.artifactInfo = {
+      type: 'repository',
+      rootDir: 'src',
+    };
+
+    this.artifactInfo = {
```
bajtos: IIUC, this line is overriding `this.artifactInfo` defined with `type: repository` above, and changes it to `type:datasource`. Is intentional?

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,398 @@
...
+      rootDir: 'src',
+    };
+
+    this.artifactInfo.outDir = path.resolve(
+      this.artifactInfo.rootDir,
+      'repositories',
+    );
+    this.artifactInfo.datasourcesDir = path.resolve(
+      this.artifactInfo.rootDir,
+      'datasources',
```
bajtos: It would be great if we could share the code building `datasourcesDir`, `modelDir` and other convention-based directories between generators, so that there is a single place defining our convention.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,398 @@
...
+        `${ERROR_NO_MODELS_FOUND} ${this.artifactInfo.modelDir}.
+        ${chalk.yellow(
+          'Please visit http://loopback.io/doc/en/lb4/Repository-generator.html for information on how models are discovered',
+        )}`,
+      );
+    }
+
+    return this.prompt([
+      {
+        type: 'checkbox',
```
bajtos: I am little bit confused about this - why would we allow the developer to pick multiple models for a single repository? What use case do you have in mind?

I would expect that a single repository class is tied to a single model class. (Relations are out of scope of this pull request, we don't know yet whether they will use the same `lb4 relation` command or perhaps a different command specific to relations.)

Let's discuss!

packages/cli/lib/ast-helper.js
```diff
@@ -0,0 +1,149 @@
...
+ * Parse the file using the possible formats specified in the arrays
+ * rootNodesFindID and childNodesFindID
+ * @param {string} fileContent with a model.ts class
+ */
+exports.getIdFromModel = async function(fileContent) {
+  let nodePos = 0;
+  let retVal = null;
+
+  for (let rootNodes of RootNodesFindID) {
+    const ast = tsquery.ast(fileContent);
```
bajtos: Is it necessary to build a new `ast` object for every `rootNodes` entry? Cannot we share the same `ast` objectt for all queries?

I am concerned about performance, parsing source code into AST representation takes non-negligible time from my experience with running `prettier` on loopback-next.

packages/cli/lib/ast-helper.js
```diff
@@ -0,0 +1,149 @@
...
+    let propertyArr = [];
```
bajtos: Can we use `const` here?

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -0,0 +1,343 @@
...
+const MODEL_APP_PATH = 'src/models';
+const REPOSITORY_APP_PATH = 'src/repositories';
+const INDEX_FILE = path.join(SANDBOX_PATH, REPOSITORY_APP_PATH, 'index.ts');
+
+const SANDBOX_FILES = [
+  {
+    fixturePath: DATASOURCE_FIXTURES_PATH,
+    path: DATASOURCE_APP_PATH,
+    file: 'dbkv.datasource.json',
+    content: `
```
bajtos: Please don't use string literals to define JSON objects, use `JSON.stringify` instead.

```ts
content: JSON.stringify({
  name: "dbkv",
  // ...
});
```

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -0,0 +1,343 @@
...
+    {
+      "name": "dbkv",
+      "connector": "kv-redis",
+      "url": "",
+      "host": "localhost",
+      "port": 3306,
+      "password": "test",
+      "db": "test"
```
bajtos: Are these options (url, host, ..., db) required for the test to work? Can we leave them out and keep only the properties that are required by the test, i.e. `name` and `connector`?

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -0,0 +1,343 @@
...
+
+// Sandbox fixture files definition
+const DATASOURCE_FIXTURES_PATH = 'datasources';
+const MODEL_FIXTURES_PATH = 'models';
+const DATASOURCE_APP_PATH = 'src/datasources';
+const MODEL_APP_PATH = 'src/models';
+const REPOSITORY_APP_PATH = 'src/repositories';
+const INDEX_FILE = path.join(SANDBOX_PATH, REPOSITORY_APP_PATH, 'index.ts');
+
+const SANDBOX_FILES = [
```
bajtos: Please move the definition of these sandbox files after the actual test cases, see https://loopback.io/doc/en/contrib/style-guide.html#layout-of-test-files

packages/cli/generators/repository/templates/src/repositories/repository-crud-default-template.ts.ejs
```diff
@@ -0,0 +1,14 @@
+import {<%= repositoryTypeClass %>}, juggler} from '@loopback/repository';
```
bajtos: I think you have an extra `}` here. Should be:

```
import {<%= repositoryTypeClass %>, juggler} from '@loopback/repository';
```

bajtos:
I also noticed that when `lb4 repository` finishes, it prints the following message:

```
Datasource undefined was created in src/repositories/
```

I would expect the artifact type to be `Repository` (not `Datasource`) and to be followed by the actual repository name (instead of `undefined`).

bajtos:
@marioestradarosa I am done with the review and manual testing, please take a look at my comments above.

I hope the long list of issues to address will not discourage you from finishing the work. 🤞 We are looking forward to work together with you to get this great contribution landed.

marioestradarosa:
> I hope the long list of issues to address will not discourage you from finishing the work
@bajtos , not at all, I am happy to brush up my coding 👍  and learning from **all** of you.

packages/cli/generators/repository/index.js
marioestradarosa: for now, these are only used to read the ts files in the specified directory.  you mean having these definitions as export members in a library ?

packages/cli/generators/repository/index.js
bajtos: Oh, I see.

I would not choose this UX design myself, but I am fine to follow the opinion of the rest of the team here.

packages/cli/generators/repository/index.js
bajtos: I mean that `lb4 datasource` must be using this same approach based on the magic string `datasources` to find the directory where to put the newly generated `.ts` and `.json` files. Maybe it's handled by one of the base generator classes?

Ideally, the code determining where to write a file like `src/datasource/db.datasource.ts`  should be the same that's used to determine where to look for those datasource files.

I am not sure if I have explained myself well, please do ask more questions if needed!

At the same time, this is not a big deal in the context of this pull request, we can defer such cleanup for later, after this patch is landed.

marioestradarosa:
> I also noticed that when lb4 repository finishes, it prints the following message:

Datasource undefined was created in src/repositories/

Done !.  now it says Repository and the repository or a list of repositories created.

packages/cli/generators/repository/index.js
marioestradarosa: > I would not choose this UX design myself

in theory this PR hide this detail from the app developer, in that he/she can pass in the command line a single repository.

marioestradarosa:
> An attempt to create a repository in an app that has a model but no datasources (lb4 app + lb4 model) fails with an ugly error. Please include a test to verify the fix (unless it's too difficult to write one).


addressed already. now it provides the appropriate error message and a test was added.

marioestradarosa:
> The code scaffolded by lb4 repository for a CRUD repository does not compile.

I tested and created multi-word, simple ones and it compiles now. probably the problem was the additional } character in the templates.

marioestradarosa:
👍 @virkt25  I was waiting for your review 😄

hacksparrow:
Created a repository successfully. The end message says: "Datasource to use Animal was created in src/repositories/".

Maybe it should be something like "Repository for Animal model created". 

Is there a possibility that repositories can be created elsewhere apart from `src/repositories/`. If not, then, maybe we should not mention the dir?

marioestradarosa:
Thanks @hacksparrow  👍 it is fixed now.

marioestradarosa:
> Is there a possibility that repositories can be created elsewhere apart from src/repositories/

I believe that it doens't hurt to have the layout in the way it is now. I like the oppinionated way in this case so all  participants (_app developers, framework developers)_ are in the same page, and app developers follow a standard and best practices.

Allowing to change all dirs (ie: datasources, services etc.) would imply also to change how the booters load these files.   And too add a more complex scenario, imaging developers being able to have artifacts everywhere such as some repositories in src/repos and some in src/myrepos  in the same app :-)

However, if app developers should ask for this feature, I believe that in the app generator we should ask for these locations, presented as questions with default values and then from there it should be the developr responsability to modify it and no other generator should ask again for these locations but rather fed from the config file.  This option shouldn't be an array but a literal string again to make the booter life simpler and functional.

marioestradarosa:
I was able to generate and compile succesfully a KV repository until I modified the template as indicated in this doc section:  https://loopback.io/doc/en/lb4/Repositories.html

packages/cli/generators/repository/index.js
bajtos: > ok, one single place to define the directory locations and then import them from all generators , correct?

Yes, exactly!

docs/site/Repository-generator.md
```diff
@@ -0,0 +1,97 @@
...
+{% include content/generator-create-app.html lang=page.lang %}
+
+### Synopsis
+
+Adds a new [Repository or Multiple Repositories](Repositories.md) class to a
+LoopBack application with one single command.
+
+### Notes
+
+The optional argument `[<name>]`, referes to a valid datasource already created
```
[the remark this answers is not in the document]
marioestradarosa: > Either way, I think the datasource should be specified via --dataSource argument?

I agree. 

> I find it confusing that names is referring to datasource in lb4 repository command.

It is extending from ArtifactGenerator and because of the multiple selection of models and thus mulitple repository generation feature:

1. I couldn't see the name as the repository name because it can generate multiple ones
2. ArtifactGenerator still was showing the name in the help see `lb4 controller --help` for instance 
3. I agree that it should be datasource name and this can be discarded
4. I will be placeing the option --dataSource instead

docs/site/Repository-generator.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Configuration file
+
+This generator supports a config file with the following format, see the
+Standard options below to see different ways you can supply this configuration
+file.
+
+```ts
+{
+  "name": "validDataSourceName",
```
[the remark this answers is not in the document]
marioestradarosa: Sure. see above comment.

docs/site/Repository-generator.md
```diff
@@ -0,0 +1,97 @@
...
+lb4 repository [options] [<name>]
+```
+
+### Options
+
+`--model` : _(Optional)_ name of a valid model already created in src/models
+
+`--id` : _(Optional)_ name of the property serving as **ID** in the selected
+model. Keep in mind that CLI will still check and try to infer a valid ID from
+the model file and give it priority over the one supplied here.
```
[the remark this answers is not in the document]
marioestradarosa: how about adding a boolean switch `--force`  , that could allow the user to generate without inspecting or validating a datasource, a model and the id name , plus , with this switch we can use the name for the repository name also ?

packages/cli/generators/repository/index.js
```diff
@@ -29,7 +30,8 @@ const PROMPT_MESSAGE_DATA_SOURCE = 'Please select the datasource';
 const ERROR_READING_FILE = 'Error reading file';
 const ERROR_NO_DATA_SOURCES_FOUND = 'No datasources found in';
 const ERROR_NO_MODELS_FOUND = 'No models found in';
-const ERROR_NO_MODEL_SELECTED = 'You did not select any model';
+const ERROR_NO_MODEL_SELECTED = 'You did not select a valid model';
+const ERROR_NO_DIRECTORY = "couldn't find the directory";
```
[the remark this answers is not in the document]
marioestradarosa: 👍 ok

packages/cli/generators/repository/index.js
```diff
@@ -45,7 +47,7 @@ module.exports = class RepositoryGenerator extends ArtifactGenerator {
     let fileContent = '';
     let modelFile = path.join(
       this.artifactInfo.modelDir,
-      modelName.toLowerCase() + '.model.ts',
+      `${utils.kebabCase(modelName)}.model.ts`,
```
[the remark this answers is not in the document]
marioestradarosa: ok

packages/cli/generators/repository/templates/src/repositories/repository-kv-template.ts.ejs
```diff
@@ -1,10 +1,13 @@
-import {<%= repositoryTypeClass %>}} from '@loopback/repository';
+import {<%= repositoryTypeClass %>} from '@loopback/repository';
 import {<%= modelName %>} from '../models';
+import {<%= dataSourceImportName %>} from '../datasources';
```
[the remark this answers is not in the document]
marioestradarosa: Maybe @raymondfeng  can help us here 😄 .   A generated KV repository using the generic juggler.DataSource didn't compile until I followed the template in the tutorial  see  https://github.com/strongloop/loopback-next/pull/1659#issuecomment-421040192

docs/site/Repository-generator.md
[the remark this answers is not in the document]
bajtos: Maybe your generator should be inheriting from a different base generator class?

docs/site/Repository-generator.md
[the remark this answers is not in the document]
bajtos: We already have a switch called `--yes` that allow prompts to be skipped if a prompt has default value or the corresponding anwser exists in options.

I am concerned that having both `--force` and `--yes` would be confusing to our users.

docs/site/Repository-generator.md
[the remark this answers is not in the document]
marioestradarosa: probably the option is BaseGenerator, I just followed up the other options initially. let me change it and try it.

docs/site/Repository-generator.md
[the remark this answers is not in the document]
marioestradarosa: 👍 , agree. i din't see --yes ,but that make sense. however, the --yes skips the prompts and that should be treated internally in the Generator as skipping the checkings also right?.

packages/cli/generators/repository/index.js
marioestradarosa: I think we should address this in a separate PR and consolidate in one place these feature for all generators, what do you think?

packages/cli/generators/repository/index.js
marioestradarosa: I left Artifact Generator and include support for name argument to follow the standards. However, the name is only used for the first model in a multi-select model generation process.

marioestradarosa:
@bajtos  PTAL

packages/cli/generators/repository/index.js
bajtos: Sure, I am fine with addressing this in a separate PR.

See my earlier comment above.
> At the same time, this is not a big deal in the context of this pull request, we can defer such cleanup for later, after this patch is landed.

docs/site/Repository-generator.md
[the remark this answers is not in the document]
bajtos: Yes, I believe so.

packages/cli/lib/utils.js
```diff
@@ -370,3 +370,24 @@ function validateValue(name, unallowedCharacters) {
   }
   return true;
 }
+/**
+ *  Returns the modelName in the directory file format for the model
+ * @param {string} modelName
+ */
+exports.toModelFileName = function(modelName) {
```
bajtos: Perhaps `getModelFileName` would be a better function name?

packages/cli/lib/utils.js
```diff
@@ -370,3 +370,24 @@ function validateValue(name, unallowedCharacters) {
...
+  return `${_.kebabCase(modelName)}.model.ts`;
+};
+
+/**
+ * Returns the repositoryName in the directory file format for the repository
+ * @param {string} repositoryName
+ */
+exports.toRepositoryFileName = function(repositoryName) {
```
bajtos: Perhaps `getRepositoryFileName` would be a better function name?

packages/cli/generators/repository/templates/src/repositories/repository-crud-default-template.ts.ejs
```diff
@@ -0,0 +1,14 @@
+import {<%= repositoryTypeClass %>, juggler} from '@loopback/repository';
+import {<%= modelName %>} from '../models';
+import {inject} from '@loopback/core';
+
+export class <%= className %>Repository extends <%= repositoryTypeClass %><
+  <%= modelName %>,
+  typeof <%= modelName %>.prototype.<%= idType %>
+> {
+  constructor(
+    @inject('datasources.<%= dataSourceName %>') protected datasource: juggler.DataSource,
```
bajtos: Can we make this consistent with `repository-kv-template.ts.ejs` please and use `dataSourceImportName` instead of `juggler.DataSource`?

Also `dataSourceImportName` does not sound as a correct description to me, how about `dataSourceClassName`?

packages/cli/lib/utils.js
marioestradarosa: @bajtos , I thought about this name, however, I wanted to keep consistency with the other method toClassName in the same utils.js file. so we have toClassName, toRepositoryFileName etc.

packages/cli/generators/repository/templates/src/repositories/repository-crud-default-template.ts.ejs
marioestradarosa: dataSourceClassName sounds perfect.

packages/cli/lib/utils.js
marioestradarosa: There was method called toClassName already in utils.js , so I wanted to keep the same line with toModelFileName. But I believe we can change here toModelFileName and then probably open another PR to change the current one toClassName .

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,469 @@
...
+      });
+    }
+
+    debug(
+      `artifactInfo.dataSourceClassName ${
+        this.artifactInfo.dataSourceClassName
+      }`,
+    );
+
+    if (availableDatasources.length === 0) {
```
jannyHou: would the length be checked before [assigning the `dataSourceClassName` to `this.artifactInfo`](https://github.com/strongloop/loopback-next/pull/1659/files#diff-d7868e84823c8c505dac05e06ae1d44aR251)?

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+    const availableDatasources = datasourcesList.filter(item => {
+      debug(`data source unfiltered list: ${item}`);
+      const result = this._isConnectorOfType(
+        VALID_CONNECTORS_FOR_REPOSITORY,
+        item,
+      );
+      return result;
+    });
+
+    debug(`artifactInfo.dataSourceClass ${this.artifactInfo.dataSourceClass}`);
```
virkt25: This seems out of place. Did you mean to have it after the `if` statement on Line 264?

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+        this.options.name,
+      );
+
+      // make sure the name supplied from cmd line is only used once
+      delete this.options.name;
+    } else {
+      this.artifactInfo.className = utils.toClassName(
+        this.artifactInfo.modelName,
+      );
+      this.artifactInfo.outFile = utils.getRepositoryFileName(
```
virkt25: Overwriting `outFile` each time a model from `modelNameList` is processed will result in only the last generated `repository` file to be exported. 

To write multiple files to `index.ts`, the following must be done:
- Create an empty property in `_setupGenerator()` on `artifactInfo. `this.artifactInfo.indexesToBeUpdated = []`.
- After setting `outFile` here, add the following: `this.artifactInfo.indexedToBeUpdated.push({dir: this.artifactInfo.outDir, file: this.artifactInfo.outFile});`

For implementation of updating indexes see: https://github.com/strongloop/loopback-next/blob/master/packages/cli/lib/artifact-generator.js#L112-L127

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+        this.artifactInfo.modelName,
+      );
+    }
+
+    if (debug.enabled) {
```
virkt25: nitpick: this if is not needed as there is nothing process heavy about `outFile`.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+
+    const source = this.templatePath(
+      path.join(
+        utils.sourceRootDir,
+        utils.repositoriesDir,
+        this.artifactInfo.defaultTemplate,
+      ),
+    );
+
+    if (debug.enabled) {
```
virkt25: nitpick: this if is not needed as `source` is just a string.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+      debug(`Using template at: ${source}`);
+    }
+    const dest = this.destinationPath(
+      path.join(this.artifactInfo.outDir, this.artifactInfo.outFile),
+    );
+
+    if (debug.enabled) {
```
virkt25: This is needed due to the `inspect` 👍 💯

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,467 @@
...
+           */
+          delete this.options.id;
+        } else {
+          idType = await this._getModelIdType(item);
+          if (idType === null) {
+            const answer = await this.prompt(prompts);
+            idType = answer.propertyName;
+          }
+        }
+        this.artifactInfo.idType = idType;
```
virkt25: nitpick: Can we rename this to `idProp` or `idProperty` -- While looking at the template I was very confused because `idType` to me implied this was a type .. such as `string`, `number`, etc. but IIUC it is the name of the id property. --- I'd say it only needs to be changed on `artifactInfo` and in the `*.ejs` templates.

packages/cli/generators/repository/index.js
marioestradarosa: Nice 👍 . However, I had to change the artifact-generator.js in order to work. if you take a look at the else on line 141, would reset ` this.artifactInfo.indexesToBeUpdated = [];`  the array in our case. I will do another commit with the changes in a minute.

packages/cli/generators/repository/index.js
marioestradarosa: I changed it to idProperty and also change the method named to getModelIdProperty.

packages/cli/lib/utils.js
```diff
@@ -370,3 +370,24 @@ function validateValue(name, unallowedCharacters) {
...
+/**
+ * Returns the repositoryName in the directory file format for the repository
+ * @param {string} repositoryName
+ */
+exports.getRepositoryFileName = function(repositoryName) {
+  return `${_.kebabCase(repositoryName)}.repository.ts`;
+};
+
+// literal strings with artifacts directory locations
+exports.repositoriesDir = 'repositories';
```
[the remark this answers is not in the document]
marioestradarosa: Yes, I think this should be fixed in a separate PR and involve the other generators in order to standardized it.

marioestradarosa:
> One more thing I would like to see done apart from the comments would be a test to cover the multiple repository generation -- however as this may involve some effort I'd be ok if this was done in a follow up PR.

I added a test for multiple repository generation and fixed the lib/ArtifactGenerator.js in order to support the `indexesToBeUpdated` feature. I checked and no other generator was using this feature, please check this file.

I also fixed the nitpicks.

@virkt25 PTAL

packages/cli/lib/utils.js
bajtos: I see. In that case I am ok with either way.

packages/cli/generators/repository/index.js
```diff
@@ -0,0 +1,469 @@
...
+  setOptions() {
+    return super.setOptions();
+  }
+
+  async checkPaths() {
+    // check for datasources
+    if (!fs.existsSync(this.artifactInfo.datasourcesDir)) {
+      return this.exit(
+        new Error(
+          `${ERROR_NO_DIRECTORY} ${this.artifactInfo.datasourcesDir}.
```
raymondfeng: I think a better message could be `No datasources found at ${this.artifactInfo.datasourcesDir}`.

marioestradarosa: I changed it for both models and repositories. @raymondfeng  PTAL

virkt25:
@marioestradarosa Looks like we have all the signoffs to land this PR, congrats 🎉 

To land this can you please rebase this PR against `master` as well as squash the 17 commits into 1 (or as few as needed) to keep the history clean. Then You can just go ahead and merge it :)

PR #1611, opened 2018-08-15
feat(rest): allow static assets to be served by a rest server

raymondfeng:
This PR allows static assets to be served from a REST server by calling `static()`.

## Missing tests identified by @bajtos

- [ ] What happens when there is both a LB4 route (e.g. a controller method) and a static file mapped at the same URL.
- [x] New static assets can be registered after the app was started.
- ~~When a file cannot be read (e.g. because of permission problems), the error response is sent by the Sequence (as provided/customized by the app).~~

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #1637, opened 2018-08-23
feat(rest): provide more options to customize api explorer and openapi spec

raymondfeng:
The generated openapi spec has a dummy server url at the moment: `servers: [url: '/']`. The PR adds more options to rest server config.

See https://github.com/strongloop/loopback-next/blob/c78d636b9cccf66bc843ec49b203b0a6c319f78e/packages/rest/README.md#configuration

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #1664, opened 2018-08-31
[WIP][RFC] Add local swagger ui for LoopBack 4 API explorer

raymondfeng:
Depends on https://github.com/strongloop/loopback-next/pull/1611

## Checklist


hacksparrow:
Yes, `@loopback/explorer` should be a LB4 extension. Baking it in with `@loopback/rest` doesn't sound right.

raymondfeng:
@bajtos @hacksparrow I agree that it should be an extension to `rest`. 

I'm just taking baby steps to get there. :-).

raymondfeng:
I took a step to create `ExplorerComponent`, which contributes an Express middleware and expects to receive the configuration from the context.

There seems to be few gaps to have `RestComponent` to collaborate with `ExplorerComponent`. See `FIXME` sections on https://github.com/strongloop/loopback-next/blob/cba9e21efb4593c025df707b7c35514e77727379/packages/explorer/test/integration/rest.integration.ts. It boils down the question of how to divide responsibility between the two components.

packages/explorer/test/integration/rest.integration.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  before(async () => {
+    server = await givenAServer({
+      rest: {},
+    });
+
+    // FIXME: How do we configure the API Explorer UI?
+    server.bind(ExplorerBindings.CONFIG).to({});
+
+    // FIXME: Should the ExplorerComponent contribute a handler or route?
```
virkt25: I think it might be better to contribute a route. Do we even support multiple handlers within an application?

hacksparrow:
Btw, what do we want the url to be? `/swagger-ui` or `/api-explorer`? `/api-explorer` sounds more intuitive to me, besides being the same as in LB3.

packages/rest/src/rest.server.ts
```diff
@@ -232,25 +245,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
   protected _handleHttpRequest(request: Request, response: Response) {
-    const mapping = this.config.openApiSpec!.endpointMapping!;
-    if (request.method === 'GET' && request.url && request.url in mapping) {
-      // NOTE(bajtos) Regular routes are handled through Sequence.
-      // IMO, this built-in endpoint should not run through a Sequence,
-      // because it's not part of the application API itself.
-      // E.g. if the app implements access/audit logs, I don't want
-      // this endpoint to trigger a log entry. If the server implements
-      // content-negotiation to support XML clients, I don't want the OpenAPI
-      // spec to be converted into an XML response.
```
bajtos: Please preserve this comment.

packages/rest/src/rest.server.ts
```diff
@@ -202,11 +202,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
     this._expressApp.use(cors(corsOptions));
 
     // Place the assets router here before controllers
     this._setupRouterForStaticAssets();
 
+    const mapping = this.config.openApiSpec!.endpointMapping!;
+    // Serving OpenAPI spec
+    for (const p in mapping) {
+      this._expressApp.use(p, (req, res) =>
+        this._serveOpenApiSpec(req, res, mapping[p]),
```
bajtos: First of all, please move this new setup code to a new method, similar to `_setupRouterForStaticAssets`.

Secondly, I think this change makes sense on its own and should have been made as part of #1637. Could you please move it to #1637 or perhaps open a new PR after #1637 is landed? I'd like to get this cleanup landed quickly and fear that the Explorer pull request will require longer to get into a mergeable state.

packages/explorer/test/integration/rest.integration.ts
```diff
@@ -0,0 +1,45 @@
...
+    const handler = await server.get(ExplorerBindings.HANDLER);
+
+    // FIXME: How to register routes to the rest server by phase or order
+    // FIXME: Should REST server automatically mounts the route based on registration
+    // via bindings such as `rest.server.routes`?
+    server.staticRouter.use('/explorer', handler);
```
bajtos: I find this very hacky.

Just the fact that `staticRouter` is a public property is alarming to me, as it opens door to all kinds of hacks that people can start making with our internal implementation details of `app.static()`. Let's fix that please and make `server.staticRouter` a private property.

Secondly, I thought the entire purpose of `app.static` was to allow the explorer component to serve swagger-ui static assets. I don't fully understand why you don't want to use it?

IMO, we should focus on `RestApplication` scenario in the first iteration because multi-server apps are not officially supported yet, and call `app.static()` in component constructor.

Thoughts?

packages/rest/src/rest.server.ts
```diff
@@ -561,6 +555,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._routerForStaticAssets.use(path, express.static(rootDir, options));
   }
 
+  get staticRouter() {
```
bajtos: Please don't, see my comment above.

packages/explorer/test/integration/rest.integration.ts
```diff
@@ -0,0 +1,45 @@
...
+      .get('/explorer')
+      .expect(200, /\<title\>LoopBack API Explorer<\/title\>/)
+      .expect('content-type', /text\/html.*/);
+  });
+});
+
+async function givenAServer(options?: {rest: RestServerConfig}) {
+  const app = new Application(options);
+  app.component(RestComponent);
+  // FIXME: Can we mount the ExplorerComponent to a given server?
```
bajtos: I think this is a missing extension point?

At the moment, components can customize only the entire `Application`, they don't have a good access to individual `Server` instances. 

However, for 4.0 GA, we support only single-server applications using `RestApplication`, so I would not worry about this too much. I am proposing to create a follow-up issue for post-GA and move on.

hacksparrow:
> I'd like to propose an entirely different approach.

I like that approach - less disruptive and simpler to implement.

Things to consider:

1. Make Explorer optional
2. Maybe run on a separate process?

raymondfeng:
> IMO, LoopBack is primarily intended for REST API servers, it's not a universal HTTP server good for serving single-page applications. API Explorer is a single page application.

In reality, some of our users use LoopBack to serve both APIs and web assets. We should exclude developers from adopting LoopBack for such use cases.

> What if we leverage vanilla express to build API Explorer and serve the explorer on a different port?Or perhaps use middleware composition to add LoopBack application as a middleware in the top-level express app?

We should make it possible if an application decides to use `Express` as the container to mount multiple routes. But if our developers prefers the LoopBack way, we should have such capabilities.

packages/explorer/test/integration/rest.integration.ts
virkt25: I guess this is a missing piece that needs to be figured out. I think ideally a component should be allowed to contribute a route ... similar to how it can contribute a component and then rest will figure out the mapping of the files to the route.

virkt25:
@raymondfeng Can you please rebase this PR? Something seems off as I'm seeing 28 commits & 105 file changes and can't do a review at this time as a result.

packages/rest/src/rest.server.ts
raymondfeng: I agree. I'm playing with various options here to surface the problems.

packages/explorer/test/integration/rest.integration.ts
bajtos: At the moment, extensions can contribute controller (classes). So instead of contributing a single route, Explorer can contribute controller class with a single method (HTTP endpoint). That should work OOTB right now, BUT this new endpoint will be included in OpenAPI spec documenting application's API. I think we should add an option for hiding certain endpoints from the documentation, e.g. `@undocumented` or `@private`.

bajtos: > call APIs on RestServer such as route(...)? This will create a dependency from explorer to `rest.

I am not really concerned about this. IMO, the API explorer cannot function with the REST component, because it's tightly coupled with OpenAPI.

If you are concerned about coupling at the level of Node.js dependencies, then I am proposing to define a minimal `RestApplication` interface inside the Explorer extension, this interface can describe the `route` API we need.

raymondfeng:
@bajtos @virkt25 I reworked the PR to allow explorer component to contribute a middleware to the REST server. PTAL.

bajtos:
I am not happy at all about allowing third party components and user applications to access the internal Express router.

1. Ensuring the correct order of middleware registration has been always a difficult problem. We have somewhat improved the situation by using middleware phases in LoopBack 3.x, but it was far from perfect.

2. One of the main design goals of LoopBack Next was to abandon the concept of Express(-like) middleware and use better design for composing different request processing pieces.

3. By exposing the built-in express router, we are putting ourselves into a corner. It will be difficult to replace Express with a better solution in the future.

In one of my earlier comments, I was proposing to use `app.static()` API that was added primarily to support API Explorer use case.
> Secondly, I thought the entire purpose of `app.static` was to allow the explorer component to serve swagger-ui static assets. I don't fully understand why you don't want to use it?
> 
> IMO, we should focus on RestApplication scenario in the first iteration because multi-server apps are not officially supported yet, and call app.static() in component constructor.
>
> At the moment, components can customize only the entire Application, they don't have a good access to individual Server instances.
>
> However, for 4.0 GA, we support only single-server applications using RestApplication, so I would not worry about this too much. I am proposing to create a follow-up issue for post-GA and move on.

I am not entirely against using a different approach, but need to hear an explanation why we cannot use `app.static()` in the first place - what problems would that introduce?

Considering that an explorer hosted at `explorer.loopback.io` is good enough for 4.0 GA, and the pull request https://github.com/strongloop/explorer.loopback.io/pull/1 should hopefully land in few days time, I think we should avoid shortcuts in this pull request and aim for a well-designed solution.

For example:
1. Use `app.static` to serve static swagger-ui assets
2. Define a new flag (OpenAPI OperationObject extension) that will mark certain HTTP endpoints as hidden from the documentation (the OpenAPI spec generated for the app).
3. Define a hidden route (either via `app.route` or by exporting a controller from the component) that will return JSON file with the dynamic configuration needed by swagger-ui, see how loopback-component-explorer works. Use the new flag (see item 2) to hide this route from API spec & docs.

raymondfeng:
> I am not entirely against using a different approach, but need to hear an explanation why we cannot use app.static() in the first place - what problems would that introduce?

Please note we allow customization of the home page and render it as an EJS view while keeping relative paths to js assets from `swagger-ui-dist`. `app.static()` does not support views.

> Considering that an explorer hosted at explorer.loopback.io is good enough for 4.0 GA, and the pull request strongloop/explorer.loopback.io#1 should hopefully land in few days time, I think we should avoid shortcuts in this pull request and aim for a well-designed solution.

Sure, `explorer.loopback.io` will buy us some buffer so that we can flush out a reasonable approach.

raymondfeng:
> I am not happy at all about allowing third party components and user applications to access the internal Express router.

In the latest code, I have removed the exposure of the internal Express router. Instead, we allow extensions to bind an Express middleware to the extension point.

> Ensuring the correct order of middleware registration has been always a difficult problem. We have somewhat improved the situation by using middleware phases in LoopBack 3.x, but it was far from perfect.

> https://github.com/strongloop/loopback-next/pull/1671 is my experiment to address the ordering issue. 
- It adds phase-based relative ordering to register handlers
- It allows `(ctx, chain) => Promise` style handler, which has much better control for one-way (express style) or cascading (chain.next(), Koa style). Also special `$final` and `$error` phases are added to support `try-catch-finally`.
- It allows plug-in of existing Express middleware as-is, but wrap them internally to be consistent with our new handler design/signature.

> One of the main design goals of LoopBack Next was to abandon the concept of Express(-like) middleware and use better design for composing different request processing pieces.

Yes and No. Yes - We want to have a better handler/chain design. No - One of the major reason of re-adopting Express as the http container is to allow the pluggability of existing Express middleware modules as-is. 

> By exposing the built-in express router, we are putting ourselves into a corner. It will be difficult to replace Express with a better solution in the future.

Again, we don't expose Router directly. Instead, we'll offer an API such as `server.middleware` to register Express middleware modules. The Express routing is not necessarily used behind the scene. Please also note that Express Router is different from its routing. I see a Router as a grouping of middleware handlers.

We should open a new issue to discuss how to position Express with LoopBack 4 and what's the ideal programming model to get the best of both breeds.

packages/rest/src/rest.server.ts
```diff
@@ -272,13 +282,29 @@ export class RestServer extends Context implements Server, HttpServerLike {
     return this.httpHandler.handleRequest(request, response);
   }
 
+  protected _registerMiddleware() {
```
virkt25: 👍

packages/rest/src/rest.server.ts
```diff
@@ -233,6 +225,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+    const corsOptions = this.config.cors || {
+      origin: '*',
+      methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
+      preflightContinue: false,
+      optionsSuccessStatus: 204,
+      maxAge: 86400,
+      credentials: true,
+    };
+    // Set up CORS
+    this._expressApp.use(cors(corsOptions));
```
virkt25: Can we register this using `this._registerMiddlerware()` to serve as an example.

bajtos:
I opened a new PR inspired by this work, see https://github.com/strongloop/loopback-next/pull/2014

Issue #1665, opened 2018-09-03
[CLI] `lb model` makes confusing prompts


hacksparrow:
## Description / Steps to reproduce / Feature proposal

There are two prompts which make no sense.

*Is ID field*?
*Required?*

![kapture 2018-09-03 at 12 29 32](https://user-images.githubusercontent.com/950112/44972041-6e380680-af75-11e8-813a-de9c749c6a0c.gif)

Is the intended prompt *Is ID field Required?*. We should fix is ASAP.

## Current Behavior

Confusing prompts.

## Expected Behavior

No confusing prompts.

bajtos:
My understanding of these prompts - please note these prompts are related to a model _property_ being defined.

> Is ID field?

Is this property the identification of the model (a primary key)?

> Required?

Is this property required, i.e. must it always have a value set?

@strongloop/sq-lb-apex could you please confirm?

virkt25:
> could you please confirm?

That was certainly the correct interpretation of those prompts.

dhmlau:
PR #1666

PR #1666, opened 2018-09-03
fix: make `lb model` prompts clearer

hacksparrow:
Fixes https://github.com/strongloop/loopback-next/issues/1665

## Checklist

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -133,7 +133,7 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
         },
         {
           name: 'id',
-          message: 'Is ID field?',
+          message: `Is ${chalk.yellow(this.propName)} the ID property?`,
           type: 'confirm',
           default: false,
           when: answers => {
@@ -145,7 +145,7 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
         },
         {
           name: 'required',
-          message: 'Required?:',
+          message: 'Is it required?:',
           type: 'confirm',
           default: false,
         },


hacksparrow:
What we should have is consistency is the format.

```
Is ID?
Is required?
```

is better than

```
Is ID?
Required?
```

@bajtos thoughts?

marioestradarosa:
Coming from a **RDBMS** world I can see the primary key word is no longer used but rather the **ID** field, resembling a unique value inside the table or model or document :-). Albeit we still design composed primary keys and use sometimes an internal one offered by the DBMS such as the (Informix rowid or Mongo DB _id)  or make our own ID fields to align with frameworks like **LoopBack 4**. 

I am writing this so you can understand the scope and the importance of the questions since we will have developers coming from the Enterprise DBs also.

In this case the _field_ is also no longer used but rather the word **property** which is really mapped to a table field or document property depending upon database management software used.

So, asking if the property we've just typed its name is the ID one, should be simply identified by current developers in the form of the following question, note the use of the `noun`.

- **Is it the ID property?**  

Which can also be presented by dynamically show the property name entered in the previous step. So it could easily become:

- Is **socialSecurity** the ID property?

The good news for `redundancy`  is that it will never ask ths question once the user has confirmed one property as the ID one.

**Now here comes the second question**. We want to know if when passing the value into the data manipulation layers in Loopback 4 (ie: insert operations) the framework should check if it contains a value before the operation (_if it is required_).

**- Is it required?**

The question uses the `noun` again to emphasize that we are still talking about the property .

hacksparrow:
Thanks for the insight @marioestradarosa.

I like Mario's suggestion. @virkt25 @bajtos what are your thoughts?

virkt25:
Thanks for the great summary @marioestradarosa! I think I agree with your proposal but think there is most certainly a fine line to balance between too much text vs. not enough text. That said I think as long as we're consistent between the two prompts, the users will learn with experience and not be as bothered by the prompts. @hacksparrow feel free to update the PR with a consistent message and we can go from there.

bajtos:
I don't really mind to be honest, as long as the prompts are easy to understand and consistent.

I think I like most the variant where the property name is included in the prompt, ideally highlighted using a different text-color. The other option, "Is it the ID property"/"Is it a required property" looks good too.

Please pick the option you like most and move forward, let's avoid bikeshedding.

Issue #1604, opened 2018-08-14
Document how to deploy LB4  app to IBM Cloud via CloudFoundry


hacksparrow:
## Description / Steps to reproduce / Feature proposal

Add documentation for deploying LB4 app to IBM Cloud's Cloud Foundry.

Related to: https://github.com/strongloop/loopback-next/pull/1574
Information to document can be inferred from the PR for the Spike.

## Acceptance Criteria
- [ ] Document Cloud Foundry artifacts creation (manifest.yml, .cfignore)
- [ ] Document service provisioning and binding (links to IBM Cloud Docs)
- [ ] Document sample code for using the provisioned service in the app's datasource

dhmlau:
@hacksparrow , as part of the grooming exercise, could you please provide more details on:
- where this information will be added in documentation
- what needs to be included, i.e. what are the things a user needs to do in order to deploy an LB app to IBM Cloud

After we have that, we can then move for estimation.  Thanks!

dhmlau:
Cross posting comments from @raymondfeng https://github.com/strongloop/loopback-next/issues/1614#issuecomment-413567914:
> We need to find out how much IBM Cloud CLI can do against a LB4 project and try to reuse it as much as possible. Check out https://console.bluemix.net/docs/cli/reference/ibmcloud/download_cli.html#install_use
>
> Having some commands in the CLI to enable/configure cloud deployment/integration is good if necessary.

bajtos:
> where this information will be added in documentation

IMO, we should create a new page called along the lines of a "Deployment guide".

> what needs to be included, i.e. what are the things a user needs to do in order to deploy an LB app to IBM Cloud

The spike pull request #1574 already contains a draft of a deployment guide, see https://github.com/strongloop/loopback-next/blob/3f55d2d819c0da5e18b7e06d41c78da4817d11a4/examples/todo/DEPLOYING.md. This guide does not describe the changes needed in the application implementation though, these changes are contained in the spike pull request.

dhmlau:
Closing as resolved. PR #1668 has landed.

PR #1668, opened 2018-09-04
docs: document deployment to Cloud Foundry

hacksparrow:
Document deployment to Cloud Foundry. Addresses https://github.com/strongloop/loopback-next/issues/1604.

## Checklist

### DIFF
--- a/docs/site/Deploying-to-IBM-Cloud.md
+++ b/docs/site/Deploying-to-IBM-Cloud.md
@@ -1,0 +1,218 @@
+---
+lang: en
+title: 'Deploying to IBM Cloud'
+keywords: LoopBack
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Deploying-to-IBM-Cloud.html
+---
+
+This is a basic guide for deploying a LoopBack 4 (LB4) app to IBM Cloud. In the
+setup explained below, your app will use a local instance of Cloundant when
+running locally, and a provisioned Cloudant service when running on the cloud.
+
+**NOTE**: Production deployment to IBM Cloud is a much bigger topic with many
+possible options, refer to
+"[IBM Cloud Continuous Delivery: Build, deploy, and manage apps with toolchains](https://www.ibm.com/cloud/garage/content/deliver/tool_continuous_delivery/)"
+for the details.
+
+Before we begin make sure the following are installed:
+
+1.  [Cloud Foundy CLI](https://docs.cloudfoundry.org/cf-cli/install-go-cli.html)
+2.  [Docker](https://www.docker.com/)
+
+Consider the
+["todo" example](https://github.com/strongloop/loopback-next/tree/master/examples/todo)
+from the [loopback-next repository](https://github.com/strongloop/loopback-next)
+as a basis for the instruction.
+
+You can quickly clone the "todo" example app by running the command:
+
+```sh
+$ lb4 example todo
+```
+
+We will be replacing the default memory-based connector of the app with a
+Cloudant connector, so data is persisted.
+
+## Cloud setup
+
+### Database
+
+1.  Provision a Cloudant database service, and name it `myCloudant`.
+2.  Log on to the Cloudant Dashboard and create a database named `todo`.
+
+### Connecting the database to app
+
+Before we can use the provisioned database, we have to connect the database to
+an app on Cloud Foundry.
+
+So, create a
+[Node.js SDK app](https://console.bluemix.net/catalog/starters/sdk-for-nodejs)
+and connect it to the provisioned Cloudant service.
+
+Make a note of the name of the app. Let's say you named it `mylb4app`. We will
+be using this name as a placeholder for the unique name you will choose yourself
+eventually. Replace `mylb4app` with your app's name in all instances in the doc.
+
+## Local setup
+
+### Database
+
+1.  [Download Cloudant image](https://hub.docker.com/r/ibmcom/cloudant-developer/).
+2.  Ensure port 8080 is free and start Cloudant:
+
+```
+docker run \
+      --volume cloudant:/srv \
+      --name cloudant-developer \
+      --publish 8080:80 \
+      --hostname cloudant.dev \
+      ibmcom/cloudant-developer
+```
+
+3.  Log on to the local Cloudant Dashboard using `admin`:`pass` as the
+    crendentials.
+4.  Create a database named `todo`.
+
+### Updates to the app
+
+#### Updating datasource
+
+Update `db.datasource.json` to use the Cloudant connector.
+
+```js
+{
+  "name": "db",
+  "connector": "cloudant",
+  "url": "http://admin:pass@localhost:8080",
+  "database": "todo",
+  "modelIndex": ""
+}
+```
+
+Install the `loopback-connector-cloudant` package.
+
+```sh
+$ npm i loopback-connector-cloudant
+```
+
+#### Updating npm script
+
+Remove the `prestart` script from `package.json`, since we don't want to do any
+building on the cloud.
+
+#### Updating application
+
+We will use the `cfenv` module to simplify some of the Cloud Foundry related
+operations. Install `cfenv` in the project directory.
+
+```sh
+$ npm i cfenv
+```
+
+Update the `src/index.ts` file to the following to enable service binding.
+
+```ts
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+const datasourceDb = require('./datasources/db.datasource.json');
+const cfenv = require('cfenv');
+const appEnv = cfenv.getAppEnv();
+
+export async function main(options?: ApplicationConfig) {
+  // Set the port assined for the app
+  if (!options) options = {};
+  if (!options.rest) options.rest = {};
+  options.rest.port = appEnv.isLocal ? options.rest.port : appEnv.port;
+
+  const app = new TodoListApplication(options);
+  // If running on IBM Cloud, we get the Cloudant service details from VCAP_SERVICES
+  if (!appEnv.isLocal) {
+    // 'myCloudant' is the name of the provisioned Cloudant service
+    const updatedDatasourceDb = Object.assign({}, datasourceDb, {
+      url: appEnv.getServiceURL('myCloudant'),
+    });
+    app.bind('datasources.config.db').to(updatedDatasourceDb);
+  }
+  await app.boot();
+  await app.start();
+
+  const url = app.restServer.url;
+  console.log(`Server is running at ${url}`);
+  return app;
+}
+```
+
+## Running the app locally
+
+1.  Install all dependencies:
+
+```sh
+$ npm i
+```
+
+2.  Build the app:
+
+```sh
+$ npm run build
+```
+
+3.  Ensure Cloudant is running locally.
+4.  Start the app:
+
+```sh
+$ npm start
+```
+
+5.  Go to http://localhost:3000/swagger-ui to load API Explorer and add a todo
+    or two.
+
+6.  Go to http://localhost:3000/todos to see the todos.
+
+## Deploying the app
+
+Authenticated to IBM Cloud using the `cf login` command before proceeding
+further.
+
+We'll need to create the following Cloud Foundry artifacts to deploy our app to
+IBM Cloud:
+
+1.  `.cfignore` - not mandatory, prevents `cf` from uploading the listed files
+    to IBM Cloud.
+2.  `manifest.yml` - details about the app.
+
+Create a file named `.cfignore` in the app directory with the following content.
+
+```
+.git/
+node_modules/
+vcap-local.js
+.vscode/
+```
+
+To create the `manifest.yml` file, run `cf create-app-manifest` and specify the
+name of the app that was created earlier, viz.: `mylb4app`.
+
+```sh
+$ cf create-app-manifest mylb4app -p manifest.yml
+```
+
+Now we are ready to deploy the app to IBM Cloud.
+
+1.  Build the app locally:
+
+```sh
+npm run build
+```
+
+2.  Push the app to Cloud Foundry:
+
+```
+cf push mylb4app
+```
+
+3.  Go to https://mylb4app.eu-gb.mybluemix.net/swagger-ui to load the API
+    Explorer and add a todo or two. The host `mylb4app.eu-gb.mybluemix.net` is
+    derived from the `name` and `domain` values in the `manifest.yml` file.
+
+4.  Go to https://mylb4app.eu-gb.mybluemix.net/todos to see the todos.
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -98,6 +98,9 @@ children:
     - title: 'Run and Test it'
       url: soap-calculator-tutorial-run-and-test.html
       output: 'web, pdf'
+  - title: 'Deploying to IBM Cloud'
+    url: Deploying-to-IBM-Cloud.html
+    output: 'web, pdf'
 
 - title: 'Key concepts'
   url: Concepts.html


docs/site/Deployment-guide-Cloud-Foundy.md
```diff
@@ -0,0 +1,244 @@
+---
+lang: en
+title: 'How to deploy LoopBack 4 app to Cloud Foundry'
```
dhmlau: If putting this as the root layer, the title should be consistent with the rest of the titles.  
Suggestion: Deploying to IBM Cloud (or something along that line)

After the title has been finalized, might want to change the file name as well.  (typo in the file name as well)

docs/site/Deployment-guide-Cloud-Foundy.md
```diff
@@ -0,0 +1,244 @@
...
+keywords: LoopBack
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Deployment-guide-Cloud-Foundry.html
```
dhmlau: make it consistent with the title after it's been finalized.

docs/site/Deployment-guide-Cloud-Foundy.md
```diff
@@ -0,0 +1,244 @@
...
+npm run build
+```
+
+2.  Push the app to Cloud Foundry:
+
+```
+cf push mylb4app
+```
+
+3.  Go to https://mylb4app.eu-gb.mybluemix.net/swagger-ui to load the API
```
dhmlau: - might want to explain a bit that the link is derived from the name and domain in the manifest.yml file.

dhmlau:
@hacksparrow , @raymondfeng mentioned some time ago that there's IBM Cloud command we can use instead of the `cf` command.  Have you had a chance to take a look to see whether that's a better approach?

hacksparrow:
@dhmlau I think that was not a separate tool, it was `cf`'s own command, if I am not wrong. @raymondfeng was it something else?

I have updated the doc accordingly.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+
+3.  Log on to the local Cloudant Dashboard using `admin`:`pass` as the
+    crendentials.
+4.  Create a database named `todo`.
+
+### Updates to the app
+
+#### Updating datasource
+
+Update `db.datasource.json` to use the Cloudan connector.
```
bajtos: typo: `Cloudan`

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+  "url": "http://admin:pass@localhost:8080",
+  "database": "todo",
+  "modelIndex": ""
+}
+```
+
+Install the `loopback-connector-cloudant` package.
+
+```sh
+$ npm i loopback-connector-cloudant -S
```
bajtos: `-s` to save this connector as a regular (production) dependency.

With the proposed `-S`, the connector will be available only at dev-time.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+Remove the `prestart` script from `package.json`, since we don't want to do any
+building on the cloud.
+
+#### Updating application
+
+We will use the `cfenv` module to simplify some of the Cloud Foundry related
+operations. Install `cfenv` in the project directory.
+
+```sh
+$ npm i cfenv -S
```
bajtos: Use `-s` for `--save` for production.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+}
+
+export async function main(options?: ApplicationConfig) {
+  // Set the port assined for the app
+  if (!options) {
+    options = {rest: {port: appEnv.port}};
+  } else {
+    if (!options.rest) options.rest = {};
+    options.rest.port = appEnv.port || options.rest.port;
+  }
```
bajtos: Let's simplify this block a little bit. How about the following?

```ts
if (!options) options = {};
if (!options.rest) options.rest = {};
options.rest.port = appEnv.port || options.rest.port;
```

Also: do we need to check `appEnv.isLocal` flag before applying `appEnv.port`?

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+Update the `index.ts` file to the following to enable service binding.
+
+```ts
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+const datasourceDb = require('./datasources/db.datasource.json');
+const cfenv = require('cfenv');
+const appEnv = cfenv.getAppEnv();
+// 'myCloudant' is the name of the provisioned Cloudant service
+const dbUrl = appEnv.getServiceURL('myCloudant');
```
bajtos: I would personally structure the code slightly differently:

```ts
const appEnv = cfenv.getAppEnv();
if (!appEnv.isLocal) {
  // 'myCloudant' is the name of the provisioned Cloudant service
  datasourceDb.url = appEnv.getServiceURL('myCloudant');
}
```

What do you think?

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+
+if (!appEnv.isLocal) {
+  datasourceDb.url = dbUrl;
```
bajtos: Modifying the object returned by `require('./datasources/db.datasource.json')` is a working solution, but I find it rather hacky. When writing a code calling `require` on a JSON file, most people expect the `require` call to return the exact data as stored in the file, not a value that was modified by somebody.

A better solution is to create a new object and set the modified properties there.

```ts
const appEnv = cfenv.getAppEnv();
if (!appEnv.isLocal) {
  // 'myCloudant' is the name of the provisioned Cloudant service
  datasourceDb = Object.assign({}, datasourceDb, {
    url: appEnv.getServiceURL('myCloudant')
  });
  app.bind('datasources.config.db').to(datasourceDb);
}
```

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,209 @@
...
+npm run build
+```
+
+2.  Push the app to Cloud Foundry:
+
+```
+cf push mylb4app
+```
+
+3.  Go to https://mylb4app.eu-gb.mybluemix.net/swagger-ui to load the API
```
bajtos: I have a question regarding a host name and the application nam `mylb4app`. Shouldn't users pick a unique application name? I don't see how the same domain `mylb4app.eu-gb.mybluemix.net` can serve every LB4 user trying out cloud deployment by following this guide.

Am I missing something here?

I think the guide should be more clear about the fact that `mylb4app` is just a placeholder that people need to change to an actual unique application name.

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: ooops!

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: It was hacky indeed. Let's have that better implementation.

hacksparrow:
@bajtos I made the changes you suggested. Thanks!

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,211 @@
...
+
+Install the `loopback-connector-cloudant` package.
+
+```sh
+$ npm i loopback-connector-cloudant -s
+```
+
+#### Updating npm script
+
+Remove the `prestart` script from `package.json`, since we don't want to do any
```
virkt25: I'm not sure I agree with this part! Building locally and deploying (without the `node_modules`) folder seems weird. What if there is a new dependency released vs. the target your app was built with? Not to mention it seems like a pain to have to build the app yourself before deploying it when the build folder isn't stored in GitHub. 

The way I personally see builds working is as follows: 
- BASIC: Build the app on the cloud as you would locally. 
- Intermediate: Set up a CI/CD pipeline to build / deploy the app, something like Travis / IBM Cloud DevOps offering. This would help limit downtime during a switch if not using blue-green deployments.

Building locally should not be a thing and the deploy should be with the dependencies it's built with IMO.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,211 @@
...
+
+Update the `index.ts` file to the following to enable service binding.
+
+```ts
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+const datasourceDb = require('./datasources/db.datasource.json');
+const cfenv = require('cfenv');
+const appEnv = cfenv.getAppEnv();
+if (!appEnv.isLocal) {
```
virkt25: I think it might be worthwhile to explain in a comment here as to what this is doing aka overriding / re-binding the config if on IBM Cloud from `cfenv`.

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: Looks like building on IBM Cloud will require `lb-tsc` to be installed globally. Moving `@loopback/build` to `dependencies` does not make `lb-tsc` available to npm script like it was installed globally.

Any suggestions?

virkt25: Really? Because I don't have typescript installed globally and have never had issues with `lb-tsc` not working locally. @raymondfeng any ideas as to why this might happen? Only thing that comes to my mind is it might be related to Node 8?

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,211 @@
...
+operations. Install `cfenv` in the project directory.
+
+```sh
+$ npm i cfenv -s
+```
+
+Update the `index.ts` file to the following to enable service binding.
+
+```ts
+import {TodoListApplication} from './application';
```
dhmlau: I got errors when somewhat copying and pasting your code snippet.  The 2 errors:
1.  In the line of ` app.bind('datasources.config.db').to(datasourceDb);`, `app` hasn't been defined yet
2. `const datasourceDb` meaning it's not changeable but indeed we are assigning another value a few lines after. 

For your reference, here is my index.ts:
```

import {ShoppingApplication} from './application';
import {ApplicationConfig} from '@loopback/core';

// IBM Cloud deployment
let datasourceDb = require('./datasources/user.datasource.json');
const cfenv = require('cfenv');
const appEnv = cfenv.getAppEnv();

export {ShoppingApplication};

export async function main(options?: ApplicationConfig) {
  // Set the port assigned for the app
  if (!options) options = {};
  if (!options.rest) options.rest = {};

  const app = new ShoppingApplication(options);
  if (appEnv.isLocal) options.rest.port = options.rest.port;
  else {
    datasourceDb = Object.assign({}, datasourceDb, {
      url: appEnv.getServiceURL('myCloudant'),
    });
    app.bind('datasources.config.db').to(datasourceDb);
    options.rest.port = appEnv.port;
  }

  await app.boot();
  await app.start();

  const url = app.restServer.url;
  console.log(`Server is running at ${url}`);
  console.log(`Try ${url}/ping`);

  return app;
}
```

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,211 @@
...
+```sh
+npm run build
+```
+
+2.  Push the app to Cloud Foundry:
+
+```
+cf push mylb4app
+```
+
```
dhmlau: I think there's a missing step of binding the service in the app (from the IBM Cloud console)

docs/site/Deploying-to-IBM-Cloud.md
dhmlau: I think there's one more glitches here.  After going to that URL, it got redirect to another url that won't work. I have to remove the "8080" port from the redirected url and change to `https`. for example: 
https://loopback.io/api-explorer/?url=https://loopback4-example-shopping.mybluemix.net/openapi.json
instead of the original:
https://loopback.io/api-explorer/?url=http://loopback4-example-shopping.mybluemix.net:8080/openapi.json

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: Maybe something to do with the environment on IBM Cloud. I tried logging in to the shell, there's no `node` or `npm`.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,211 @@
...
+  "url": "http://admin:pass@localhost:8080",
+  "database": "todo",
+  "modelIndex": ""
+}
+```
+
+Install the `loopback-connector-cloudant` package.
+
+```sh
+$ npm i loopback-connector-cloudant -s
```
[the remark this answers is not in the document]
bajtos: Maybe use a more explicit `--save` instead of `-s`, to make the command easier to understand for people that are not very familiar with `npm`.

docs/site/Deploying-to-IBM-Cloud.md
bajtos: > Building locally and deploying (without the node_modules) folder seems weird. What if there is a new dependency released vs. the target your app was built with? 

Or what if the public npm registry goes down and our production box cannot install the dependencies?

IMO, if we deploy from a local computer (instead of using a build/deploy pipeline), then we should run the build (TypeScript compilation) locally AND also deploy `node_modules` dependencies as we have installed them locally.

At least this was our conclusion when we were discussing different deployment options back in 2014/2015. We created https://github.com/strongloop/strong-build to make that process easy to automate.

@rmg could you please chime in? What's your opinion on the deployment process for Node.js apps requiring a build step?

virkt25:
So I tried to deploy `loopback4-example-shopping` to IBM Cloud as a regular app and this was my experience: 

- Forked the repo so I can make changes
- Goal was to swap the database in production from mongodb to Cloudant since it offers a free tier
- **Step 1** was to deploy as is without creating an app on IBM Cloud first. The goal was simple, push to Cloud as is, no changes. `cf push t-shopping-app` failed.
  - This created an app in IBM Cloud but failed to deploy. Missing buildpack. Specified `sdk-for-nodejs` as the buildpack via command-line (only needs to be done once) -- But I added it to my `manifest.yml` later as well.
  - Attempt to deploy again failed. Missing `lb-tsc`. IBM Cloud is a production environment and doesn't install `devDependencies`. 
  - Moved `@loopback/build` to be a regular dependency. Got errors about `tsc` missing `@types/*` packages as they are a `devDependency`. 

- I decided to build locally and deploy the app to make sure I can do that. Failed to do so since I hadn't installed before and doing so now was attempting to push `node_modules` folder which isn't allowed. I created a `.cfignore` and excluded `node_modules`.
  - Pushed and app failed as I forgot to set port from `env` in prod. Fixed that.
  - Push finally worked and I got the `ping` endpoint working.

Now to work out the database issues.

- Update code to swap out the database connection info and connector on a production environment
- Got annoyed by forgetting to `re-build` the app before a push. Got sidetracked to find a better solution and decided to use `IBM Toolchains`. 

IBM Toolchain Setup
- It created it's own private Git repo. I already had GitHub setup -- so I had to re-configure some stuff and install the GitHub ToolChain thingy (Not sure if it's called an app / plugin / extension  for the toolchain). 
- IBM Toolchain for `npm` in the build stage offers Node 0.10.x as a default! This caused failures. The max Node version that is pre-installed is Node 6.9.x. Had to figure out how to install `nvm` and my own version of `Node` since LB4 requires Node 8 or higher. Came across: https://gist.github.com/DMeechan/2f6e962622c3387785bb2e44298a153c
  - The script didn't work out of the box as is. Here's my final `Build` Stage script I used. 
```
#!/bin/bash
# Credit: http://gh-blog.mybluemix.net/blogs/cokeSchlumpf/rethink-it/posts/bluemix/node-buildpipeline.md?cm_sp=dw-bluemix-_-nospace-_-answers
# Paste the script below into IBM Cloud Delivery Pipeline Build process (shell script)
# Set your Node and NVM version:
export NODE_VERSION=8.11.4
export NVM_VERSION=0.33.11
export NVM_DIR=/home/pipeline/nvm

mkdir $NVM_DIR

npm config delete prefix \
  && curl https://raw.githubusercontent.com/creationix/nvm/v${NVM_VERSION}/install.sh | sh \
  && . $NVM_DIR/nvm.sh \
  && nvm install $NODE_VERSION \
  && nvm alias default $NODE_VERSION \
  && nvm use default \
  && node -v \
  && npm -v

export NVM_DIR="$HOME/nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"

npm install
npm run build
```
- Had to configure the rest of the toolchain (namely the `Deploy` step) to use the output of the previous build step to deploy.
- Toolchain works every time a commit is merged to `master`

Back to DataBase.
- Configured a Cloudant instance in the dashboard and bound it to the application. Set authentication as `Legacy and IAM`. 
- Was getting `404` errors from trying to POST to `/users`. 
- Figured it might be because I was missing a database. Created a `user` and `users` database. Didn't work. 
- Explicitly set the database as `user` in `user.datasource.json`. 
- Errors showed that the connector was trying to save data to a database named `test` despite the config saying `user`. Created a database named `test` in cloudant ... and everything just worked. 

--- 

I'm not a fan of trying to remember to build locally before deploying. I know one can setup local commit hooks or something but I personally like the Toolchain.

raymondfeng:
@virkt25 Do you want to try the following?
- https://docs.cloudfoundry.org/buildpacks/node/index.html#vendoring
- https://github.com/cloudfoundry/nodejs-buildpack/issues/16

hacksparrow:
Thanks for the findings @virkt25.

For someone who just wants to see a basic LB4 app running on IBM Cloud, the overall experience seems overly complex. Thoughts?

rmg:
@virkt25 to specify the node and npm versions that the buildpack should use you specify them in `engines` in `package.json`.

virkt25:
> to specify the node and npm versions that the buildpack should use you specify them in engines in package.json

This works for the actual deployed Application to IBM Cloud but IBM Toolchain doesn't seem to honour this. Below is the default build for `npm` in IBM Toolchain

image

virkt25:
> Do you want to try the following?
> 
> https://docs.cloudfoundry.org/buildpacks/node/index.html#vendoring
> cloudfoundry/nodejs-buildpack#16

The issue is that the IBM Toolchain environment is very different from IBM Cloud where the app is deployed and hence the need for that script.

dhmlau:
I think we can make this round of documentation to be just deploying a LB4 app to IBM Cloud without all the toolchain setup.  Keep it minimal and easy to follow so that users know what are the steps involved.  As a follow up, we can create an additional section on the toolchain.  Even that, I'm thinking whether we can piggyback on some of the existing docs, meaning we only add the instructions specific or additional to LB app deployment.

Not sure if it helps in the discussion, with the help of @virkt25, I was able to deploy my own LB4 app to IBM Cloud.  Similar to the starting point @virkt25 had above:
- clone the app
- make it to be cloudant connector, i.e. modify the datasource.json
- modify index.ts (https://github.com/virkt25/loopback4-example-shopping/blob/master/src/index.ts)'
- Call `cf login` to login first
- push the app using `cf push` command
- Go to IBM Cloud, create the cloudant service and name it to match with what's used in index.ts
- In IBM Cloud, go to the app, create connection to bind the cloudant service with the app 
- go to browser to open up API Explorer (with modified url that we've discussed above)

I think if we have the above steps (as an outline), it is good for the purpose of telling the users how to deploy to IBM Cloud.  

Thoughts?

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: It is mentioned under "Connecting the database to app".

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: Fixed in the latest commit. Thanks.

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: Good find. This is a bug either in the `rest` package or `loopback.io`. Will investigate further and open an issue accordingly.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,215 @@
...
+keywords: LoopBack
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Deploying-to-IBM-Cloud.html
+---
+
+This is a guide for deploying LoopBack 4 (LB4) apps to IBM Cloud. In the setup
+explained below, your app will use a local instance of Cloundant when running
+locally, and a provisioned Cloudant service when running on the cloud.
+
+Before we begin make sure the following are installed.
```
virkt25: nitpick: sentence should end in `:` instead of `.`

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,215 @@
...
+
+1.  [Cloud Foundy CLI](https://docs.cloudfoundry.org/cf-cli/install-go-cli.html).
```
virkt25: nitpick: lets not add `.` at the end of the list items.

docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -0,0 +1,215 @@
...
+
+  const url = app.restServer.url;
+  console.log(`Server is running at ${url}`);
+  return app;
+}
+```
+
+## Deploying the app
+
+### Local
```
virkt25: I'm not sure how I feel about the `local` step -- it's not necessary for a deploy of the application.

docs/site/Deploying-to-IBM-Cloud.md
virkt25: Here's my `index.ts` that I think we can leverage as I think it's simpler. Not the `default` value being set in the `main` function.

```ts
export async function main(options: ApplicationConfig = {}) {
  options.rest = options.rest || {};
  options.rest.port = appEnv.isLocal ? options.rest.port : appEnv.port;

  const app = new ShoppingApplication(options);

  if (!appEnv.isLocal) {
    userDataSource.url = appEnv.getServiceURL('t-shopping-cloudant');
    userDataSource.connector = 'cloudant';

    app.bind('datasources.config.user').to(userDataSource);
  }

  console.log('cloudant url: ', userDataSource.url);

  await app.boot();
  await app.start();

  const url = app.restServer.url;
  console.log(`Server is running at ${url}`);
  console.log(`Try ${url}/ping`);

  return app;
}
```

I am making two PRs to propose reading the port from env as a default and setting the options as a `{}` by default in hopes of simplifying this step.

hacksparrow: Writing to directly `userDataSource` will have side-effects for other files loading the datasource JSON file. I have uptimized, though.

Good proposals.

docs/site/Deploying-to-IBM-Cloud.md
hacksparrow: Removed.

bajtos:
I quickly skimmed through the current version of the deployment guide and it looks good enough to me.

I agree the build process and/or deployment pipeline is out of scope of this pull request (and the related user story). Having said that, can we add a note to tell readers that a manual build is not the best approach, they should consider setting up a build pipeline and give them a link to an existing documentation for build pipelines? I think it's good to let them be aware of other solutions, even if we don't provide any additional LoopBack-specific guidelines (yet).

>  Also ... why are we discussing cloudant integration / local setup? I find there to be a bit of a disconnect in the flow of the document. If we want to keep things simple then we can do a tutorial along the following lines:
> 
> - Deploy the app as is. Point the limitation that data may be lost between restarts. Gets users comfortable with the idea of deploying.
>  
> - Explain why we are adding Cloudant (persistance). Explain the local setup ... get the app working locally.
>
> - Show how to deploy with Cloudant on BlueMix.

I love this new outline!

On the other hand, this pull request has been opened for 9 days already, maybe it's best to land it in the current form and then rework it to the new outline in a follow-up pull request (or even a follow-up user story)?

dhmlau:
I'm good with landing this PR as-is too.  

+1 on the new outline of removing the local setup.  As I've mentioned before, it is a bit confusing to have the local set up / deploy instructions because we can assume the prereq is the app running locally already.

PR #1671, opened 2018-09-05
[RFC] Port loopback-phase to TypeScript for LoopBack 4

raymondfeng:
This PR ports `loopback-phase` to TypeScript with improvements:

- Promise based handler
- Allow the handler to control how to proceed if it's invoked as part of the invocation chain (supporting both express and koa styles)
- Special error and final phase to mimic `try - catch - finally`

This package can be used as the foundation for the use cases such as:
- Express middleware registration/execution by phase
- Sequence of action for req/res processing
- Sequence of boot/lifecycle steps

## Checklist


packages/phase/README.md
```diff
@@ -0,0 +1,134 @@
...
+  console.log(ctx.req.url);
+};
+```
+
+- Invoke downstream handlers within the method in cascading style
+
+```ts
+async (ctx, chain) => {
+  const start = process.hrtime();
+  await chain!.next();
```
bajtos: Please find a better type description for the callback function, one that does not require users to use `!` operator. If the framework always provide `chain` value, then this fact should be encoded in the type signature of the callback.

packages/phase/README.md
```diff
@@ -0,0 +1,134 @@
...
+};
+```
+
+- Terminate the handler chain and return
+
+```ts
+async (ctx, chain) => {
+  if (ctx.req.url === '/status') {
+    ctx.response = {status: 'ok'};
+    chain!.return();
```
bajtos: I find `return` as ambiguous.  Can we use a name that better expresses the intent please? For example: `abort`, `terminate` or `stop`.

packages/phase/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The handlers within the same subphase will be executed in serial or parallel
+depending on the `options.parallel` flag, which is default to `false`.
+
+The three sub-phases within the same phase will be executed in the order of
+`before`, `use`, and `after`.
+
+There is also a `failFast` option to control how to handle errors. If `failFast`
+is set to `false`, errors will be caught and set on the `ctx.error` without
+aborting the handler chain.
```
bajtos: I find this rather dangerous. I expect that most people writing phase-handlers are not going to be aware of the `failFast:false` mode. As a result, `failFast:false` mode can execute handlers that are not prepared to handle the situation left after one of the previous handlers failed.

I think a better solution would be to enable `failFast:false` mode on a per-handler basis - i.e. allow each handler to signal whether it wants to be executed even if there was an error up in the chain.

What is the use case you are envisioning for `failFast:false` mode?

Can we perhaps leave this feature out of this initial pull request, to get it landed faster?

packages/phase/README.md
```diff
@@ -0,0 +1,134 @@
...
+```ts
+try {
+  // await run handlers for phase 1
+  // await run handlers for phase 2
+  // ...
+  // await run handlers for phase N
+} catch (e) {
+  // await run handlers for errorPhase
+} finally {
+  // await run handlers for finalPhase
```
bajtos: Are we enforcing the order of `$error` and `$final` phase? Can users insert additional phases before, in-between or after these two built-in phases?

packages/phase/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@loopback/phase",
+  "version": "0.1.0",
+  "description": "LoopBack's API Explorer",
```
bajtos: Please fix the package description.

packages/phase/src/handler.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param handlers Array of handlers
+ */
+export function asHandler(
+  handlers: Handler[],
+  options: PhaseOptions = {},
+): Handler {
+  return async (ctx: Context) => {
+    if (options.parallel) {
+      debug('Executing handlers in parallel');
+      await Promise.all(handlers.map(h => h(ctx)));
```
bajtos: I don't think it's a good idea to run all handlers in parallel.

What if we have 3 handlers registered and the handler no.2 calls `ctx.return()`? It would be surprised to learn that the 3rd handler was executed to completion too!

What if we have 2 handlers registered and the first handler throws an error? I would find it surprising that the second handler was still executed. Also, are we going to wait until that second handler finishes, or will we return the error immediately?

Also, what's the point of `Phase#parallel` option when we are always running all handlers in parallel?

Please add tests to capture the expected behavior and verify the implementation.

packages/phase/src/phase-list.ts
```diff
@@ -0,0 +1,298 @@
...
+      await this.finalPhase.run(ctx);
+    }
+  }
+
+  /**
+   * Get an array of phase identifiers.
+   * @returns phaseNames
+   */
+  getPhaseNames(): string[] {
+    return this.phases.map(p => p.id);
```
bajtos: Should we include `$error` and `$final` in this list?

packages/phase/src/phase-list.ts
```diff
@@ -0,0 +1,298 @@
...
+   * // register via phase.after()
+   * phaseList.registerHandler('auth:after', async (ctx, chain) => { //... });
+   * ```
+   *
+   * @param phaseName Name of an existing phase, optionally with
+   *   ":before" or ":after" suffix.
+   * @param handler The handler function to register
+   *   with the given phase.
+   */
+  registerHandler(phaseName: string, handler: Handler) {
```
bajtos: What's the API for registering handlers for `$error` and `$final` phases?

packages/phase/test/unit/phase.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+      it('should execute phase handlers in parallel', async () => {
+        const phase = new Phase({parallel: true});
+
+        phase
+          .before(mockHandler('b1'))
+          .before(mockHandler('b2'))
+          .use(mockHandler('h1'))
+          .after(mockHandler('a1'))
+          .after(mockHandler('a2'))
+          .use(mockHandler('h2'));
```
bajtos: Please use a more descriptive names, I find it difficult to follow these terse two-letter identifiers.

For example: `before-handlerA-1`, `before-handlerA-2`, `handlerA`, `after-handlerA-1`, etc. `handlerB`.

packages/phase/README.md
raymondfeng: I introduce `failFast:false` mode mostly for error phase so that we can have a list of error handlers to mediate errors.

packages/phase/README.md
raymondfeng: Yes, we enforce them as `catch` and `finally`. No additional phases can interfere with them.

packages/phase/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```sh
+npm install --save @loopback/phase
+```
+
+## Basic Use
+
+### Handler
+
+A handler is a function that takes a `Context` object, an optional
```
virkt25: Till I dug into the code, I found the usage of the work `Context` here confusing as I thought this was referring to `@loopback/context`.

bajtos:
@raymondfeng what's the status of this pull request - is it still relevant?

raymondfeng:
Closing it to favor middleware chain.

Issue #100, opened 2017-03-29
Router: coercion of complex parameters from string sources


bajtos:
In loopback/strong-remoting 3.x, we allow remote methods to accept parameters of complex types and have the runtime to convert string inputs (e.g. in a query string or formdata) to such complex type.

- consider whether we want to implement this feature in loopback-next too
- figure how to extend ~~Swagger~~OpenAPI specification to allow developers to specify a complex parameter type to be coerced from a string
- implement the proposal

strong-remoting 3.x has pretty extensive test suite that we can reuse here - see [strong-remoting:test/rest-coercion/](https://github.com/strongloop/strong-remoting/tree/24dbcc75669ddce9d4d0aabce94563fdc7882ce6/test/rest-coercion)

## Acceptance criteria  (Scope for 4.0 GA)

 - [x]  Support "type: object" only
 - [x] Leverage OpenAPI spec parameter "style: deepObject"
 - [x] For deepObject parameters, support both `color[R]=100&color[G]=200&color[B]=150` and `color={"R":100,"G":200,"B":150}` encodings (the latter is our own extension).
 - [x] Test + Docs

Everything else (other styles than deepObject, array values, etc.) is out of scope of GA.

The spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterObject

## Examples

1. Send a "filter" query as JSON-encoded value:

    ```
    GET /api/products?filter={"where":{"name":"Pen"},"limit":10}
    ```

    Sets `filter` argument to `{where: {name: 'Pen'}, limit: 10}`.

    Alternative encoding:

    ```
    GET /api/products?filter[where][name]=Pen&filter[limit]=10
    ```

2. Encode an array of strings as a comma-separated list (OUT OF SCOPE OF 4.0 GA)

    ```
    GET /api/domains?tlds=com,net,eu
    ```

    Sets `tlds` argument to `['com', 'net', 'eu']`.

dhmlau:
Need that for REST enablement

bajtos:
I think this is a super-set of #394 (Filter in query of type object treated as string).

bajtos:
This is also related to our planned upgrade to OpenaAPI Spec v3, see https://github.com/strongloop/loopback-next/issues/182

virkt25:
+1 for implementing in LB4. I don't think this should be in `Needs Discussion` as this is missing acceptance criteria.

bajtos:
[...]

PR #1679, opened 2018-09-07
REST API to find models using a filter 

bajtos:
This pull request leverages the recently added `@param.query.object()` decorator (see #1667) to add support for querying a subset of models using a Filter object.

For example:

```
GET /todos?filter[where][isCompleted]=false
```

The OpenAPI spec contains a reasonably-minimal description of the Filter object.

screen shot 2018-09-07 at 15 42 29

**Done**
 - infrastructure to describe `filter` and `where` parameters
 - updated `todo` example 
 - update `todo-list` example
 - update CRUD Controller template in CLI
 - doc updates

See also #100

## Checklist

### DIFF
--- a/docs/site/Controller-generator.md
+++ b/docs/site/Controller-generator.md
@@ -96,8 +96,17 @@ Here's an example of what the template will produce given a `Todo` model and a
 `TodoRepository`:
 
 ```ts
-import {Filter, Where, repository} from '@loopback/repository';
-import {post, param, get, patch, del, requestBody} from '@loopback/rest';
+import {Filter, repository, Where} from '@loopback/repository';
+import {
+  post,
+  param,
+  get,
+  getFilterSchemaFor,
+  getWhereSchemaFor,
+  patch,
+  del,
+  requestBody,
+} from '@loopback/rest';
 import {Todo} from '../models';
 import {TodoRepository} from '../repositories';
 
@@ -106,43 +115,98 @@ export class TodoController {
     @repository(TodoRepository) public todoRepository: TodoRepository,
   ) {}
 
-  @post('/todos')
-  async create(@requestBody() obj: Todo): Promise<Todo> {
-    return await this.todoRepository.create(obj);
+  @post('/todos', {
+    responses: {
+      '200': {
+        description: 'Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
+      },
+    },
+  })
+  async create(@requestBody() data: Todo): Promise<Todo> {
+    return await this.todoRepository.create(data);
   }
 
-  @get('/todos/count')
-  async count(@param.query.string('where') where?: Where): Promise<number> {
+  @get('/todos/count', {
+    responses: {
+      '200': {
+        description: 'Todo model count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async count(
+    @param.query.object('where', getWhereSchemaFor(Todo)) where?: Where,
+  ): Promise<number> {
     return await this.todoRepository.count(where);
   }
 
-  @get('/todos')
-  async find(@param.query.string('filter') filter?: Filter): Promise<Todo[]> {
+  @get('/todos', {
+    responses: {
+      '200': {
+        description: 'Array of Todo model instances',
+        content: {
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Todo}},
+          },
+        },
+      },
+    },
+  })
+  async find(
+    @param.query.object('filter', getFilterSchemaFor(Todo)) filter?: Filter,
+  ): Promise<Todo[]> {
     return await this.todoRepository.find(filter);
   }
 
-  @patch('/todos')
+  @patch('/todos', {
+    responses: {
+      '200': {
+        description: 'Todo PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
   async updateAll(
-    @requestBody() obj: Todo,
-    @param.query.string('where') where?: Where,
+    @requestBody() data: Todo,
+    @param.query.object('where', getWhereSchemaFor(Todo)) where?: Where,
   ): Promise<number> {
-    return await this.todoRepository.updateAll(obj, where);
+    return await this.todoRepository.updateAll(data, where);
   }
 
-  @get('/todos/{id}')
+  @get('/todos/{id}', {
+    responses: {
+      '200': {
+        description: 'Todo model instance',
+        content: {'application/json': {'x-ts-type': Todo}},
+      },
+    },
+  })
   async findById(@param.path.number('id') id: number): Promise<Todo> {
     return await this.todoRepository.findById(id);
   }
 
-  @patch('/todos/{id}')
+  @patch('/todos/{id}', {
+    responses: {
+      '204': {
+        description: 'Todo PATCH success',
+      },
+    },
+  })
   async updateById(
     @param.path.number('id') id: number,
-    @requestBody() obj: Todo,
+    @requestBody() data: Todo,
   ): Promise<void> {
-    await this.todoRepository.updateById(id, obj);
+    await this.todoRepository.updateById(id, data);
   }
 
-  @del('/todos/{id}')
+  @del('/todos/{id}', {
+    responses: {
+      '204': {
+        description: 'Todo DELETE success',
+      },
+    },
+  })
   async deleteById(@param.path.number('id') id: number): Promise<void> {
     await this.todoRepository.deleteById(id);
   }
--- a/docs/site/todo-list-tutorial-controller.md
+++ b/docs/site/todo-list-tutorial-controller.md
@@ -98,47 +98,119 @@ Using our constraining factory as we did with the `POST` request, we'll define
 the controller methods for the rest of the HTTP verbs for the route. The
 completed controller should look as follows:
 
-#### src/controllers/todo-list-todo.controller.ts
+#### src/controllers/todo-list.controller.ts
 
 ```ts
-import {repository, Filter, Where} from '@loopback/repository';
+import {Filter, repository, Where} from '@loopback/repository';
+import {
+  del,
+  get,
+  getFilterSchemaFor,
+  getWhereSchemaFor,
+  param,
+  patch,
+  post,
+  requestBody,
+} from '@loopback/rest';
+import {TodoList} from '../models';
 import {TodoListRepository} from '../repositories';
-import {post, get, patch, del, param, requestBody} from '@loopback/rest';
-import {Todo} from '../models';
 
-export class TodoListTodoController {
+export class TodoListController {
   constructor(
-    @repository(TodoListRepository) protected todoListRepo: TodoListRepository,
+    @repository(TodoListRepository)
+    public todoListRepository: TodoListRepository,
   ) {}
 
-  @post('/todo-lists/{id}/todos')
-  async create(@param.path.number('id') id: number, @requestBody() todo: Todo) {
-    return await this.todoListRepo.todos(id).create(todo);
+  @post('/todo-lists', {
+    responses: {
+      '200': {
+        description: 'TodoList model instance',
+        content: {'application/json': {'x-ts-type': TodoList}},
+      },
+    },
+  })
+  async create(@requestBody() obj: TodoList): Promise<TodoList> {
+    return await this.todoListRepository.create(obj);
   }
 
-  @get('/todo-lists/{id}/todos')
+  @get('/todo-lists/count', {
+    responses: {
+      '200': {
+        description: 'TodoList model count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async count(
+    @param.query.object('where', getWhereSchemaFor(TodoList)) where?: Where,
+  ): Promise<number> {
+    return await this.todoListRepository.count(where);
+  }
+
+  @get('/todo-lists', {
+    responses: {
+      '200': {
+        description: 'Array of TodoList model instances',
+        content: {'application/json': {'x-ts-type': TodoList}},
+      },
+    },
+  })
   async find(
-    @param.path.number('id') id: number,
-    @param.query.string('filter') filter?: Filter,
-  ) {
-    return await this.todoListRepo.todos(id).find(filter);
+    @param.query.object('filter', getFilterSchemaFor(TodoList)) filter?: Filter,
+  ): Promise<TodoList[]> {
+    return await this.todoListRepository.find(filter);
   }
 
-  @patch('/todo-lists/{id}/todos')
-  async patch(
-    @param.path.number('id') id: number,
-    @requestBody() todo: Partial<Todo>,
-    @param.query.string('where') where?: Where,
-  ) {
-    return await this.todoListRepo.todos(id).patch(todo, where);
+  @patch('/todo-lists', {
+    responses: {
+      '200': {
+        description: 'TodoList PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async updateAll(
+    @requestBody() obj: Partial<TodoList>,
+    @param.query.object('where', getWhereSchemaFor(TodoList)) where?: Where,
+  ): Promise<number> {
+    return await this.todoListRepository.updateAll(obj, where);
   }
 
-  @del('/todo-lists/{id}/todos')
-  async delete(
+  @get('/todo-lists/{id}', {
+    responses: {
+      '200': {
+        description: 'TodoList model instance',
+        content: {'application/json': {'x-ts-type': TodoList}},
+      },
+    },
+  })
+  async findById(@param.path.number('id') id: number): Promise<TodoList> {
+    return await this.todoListRepository.findById(id);
+  }
+
+  @patch('/todo-lists/{id}', {
+    responses: {
+      '204': {
+        description: 'TodoList PATCH success',
+      },
+    },
+  })
+  async updateById(
     @param.path.number('id') id: number,
-    @param.query.string('where') where?: Where,
-  ) {
-    return await this.todoListRepo.todos(id).delete(where);
+    @requestBody() obj: TodoList,
+  ): Promise<void> {
+    await this.todoListRepository.updateById(id, obj);
+  }
+
+  @del('/todo-lists/{id}', {
+    responses: {
+      '204': {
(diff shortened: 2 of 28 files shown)


bajtos:
Regarding API Explorer/swagger-ui - I am not able to send a request with a `filter` value, it looks like the UI is somehow not passing the filter through. I'll have to investigate. According to https://github.com/swagger-api/swagger-ui/issues/4216, `style: deepObject` should be working.

packages/repository-json-schema/test/unit/filter-json-schema.unit.ts
```diff
@@ -0,0 +1,158 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, Filter, hasMany, model, property} from '@loopback/repository';
+import {expect} from '@loopback/testlab';
+import * as Ajv from 'ajv';
+import {getFilterJsonSchemaFor} from '../../src/filter-json-schema';
+import {JsonSchema} from '../../src';
+
+describe('getFilterJsonSchemaFor', () => {
+  @model()
```
b-admike: Can we define models at the bottom of the test file?

packages/repository-json-schema/test/unit/filter-json-schema.unit.ts
```diff
@@ -0,0 +1,158 @@
...
+        dataPath: '.where',
+        message: 'should be object',
+      },
+    ]);
+  });
+
+  it('describes "fields" as an object', () => {
+    const filter = {fields: 'invalid-fields'};
+    ajv.validate(customerFilterSchema, filter);
+    expect(ajv.errors || []).to.containDeep([
```
b-admike: I'm curious; why are we doing `ajv.errors || []` here?

bajtos: When `ajv.validate` returns `true`, `ajv.errors` is set to `null`. As a result, `shouldjs` produced error message that seemed to me as more difficult to understand.

I am open to suggestions on how to improve this assertion and/or make the code easier to understand!

raymondfeng:
> Regarding API Explorer/swagger-ui - I am not able to send a request with a filter value

What about stringified json? For qs style such as `filter[where][isCompleted]=false`, I don't know if possible to type into the `filter` text in the form.

bajtos:
> Regarding API Explorer/swagger-ui - I am not able to send a request with a filter value

I did some more investigation. It seems that we need to upgrade our api-explorer with the latest swagger-ui version to bring in deepObject fix from https://github.com/swagger-api/swagger-js/pull/1235. I opened a pull request for that - see https://github.com/strongloop/loopback.io/pull/736

We should also set `explode: true` in a deep-object parameter definition, I'll fix that as part of this pull request.

With those changes in place, I am able to send a request like `?filter[where]=true` by entering `{"where": true}` to the input box:

screen shot 2018-09-10 at 12 37 50

Unfortunately, when I provide a value with nested properties, e.g. `{"where":{"id: 1}}`, swagger-ui ignores such parameter value :(

Based on the following discussions, I think it may be up to us to contribute serialization of deeply nested objects:

https://github.com/swagger-api/swagger-js/pull/1140
> Limitations:
> deepObject does not handle nested objects. The specification and swagger.io documentation does not provide an example for serializing deep objects. Flat objects will be serialized into the deepObject style just fine.

https://github.com/swagger-api/swagger-js/pull/1140#issuecomment-333017825
> As for deepObject and nested objects - that was explicitly left out of the spec, and it's ok to just Not Support It™.

Related issue: https://github.com/swagger-api/swagger-ui/issues/4216

> What about stringified json? For qs style such as `filter[where][isCompleted]=false`, I don't know if possible to type into the `filter` text in the form.

Unfortunately not. Because the OAS says `filter` is an object, the user has to enter valid JSON.

When I tried to enter a string containing the JSON, e.g. `"{\"where\":{\"id\":1}}`, it triggered a validation error in swagger-ui's user interface.

bajtos:
I left a comment in a swagger-ui issue to discuss this problem further, see https://github.com/swagger-api/swagger-ui/issues/4064#issuecomment-419874700

@raymondfeng @strongloop/sq-lb-apex what's your opinion on the next steps in this pull request? Should I continue adding `filter` parameter described as `{type: 'object', style: 'deepObject', explode: true}`, despite the fact that swagger-ui does not support such `filter` values at the moment? Should we put this pull request on hold until swagger-ui supports our parameter style?

Here is my concern with landing these changes: users trying out LoopBack 4 and the API Explorer will quickly discover that they cannot send `filter` values. Because swagger-ui does not provide any hints on what's wrong, I expect a lot of confusion among our users, which may cast bad light on our new major version.

Thoughts?

virkt25:
It will definitely be an annoyance of not being able to test a `filter` from the explorer because it doesn't support deepObject serialization. I think the next steps should be to put this PR on hold ... possibly contribute a patch to `swagger-ui` and then continue work on this PR. Without support for this in swagger-ui this change to our code doesn't necessarily help users looking to test changes via the explorer.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (modelCtor !== EmptyModel) {
+    schema.properties!.include = {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          // TODO(bajtos) restrict values to relations defined by "model"
+          relation: {type: 'string'},
+          // TODO(bajtos) describe the filter for the relation target model
+          scope: scopeFilter,
```
jannyHou: Not involved in this PR: 
I think we can use a `$ref` here and append the related model's filter schema in `spec.components.schemas` when work on describing the filter for the related model.

examples/todo/test/acceptance/todo.acceptance.ts
```diff
@@ -179,4 +193,14 @@ describe('Application', () => {
...
     return await todoRepo.create(givenTodo(todo));
   }
+
+  // JSON transport does not preserve properties that are undefined
+  // As a result, deepEqual checks fail because the expected model
+  // value contains these undefined property values, while the actual
+  // result returned by REST API does not.
+  // Use this function to convert a model instance into a data object
+  // as returned by REST API
+  function deserializedFromJson<T extends object>(value: T): T {
```
jannyHou: nitpick: not a necessary part of this PR, we can update [this test](https://github.com/strongloop/loopback-next/blob/bc9e2f6e511180e7b21975d57de0480c72c1d534/examples/todo/test/acceptance/todo.acceptance.ts#L94) to call `deserializedFromJson` as well :)

raymondfeng:
Maybe we should start with what LB3 allows - when the openapi spec is generated, we mark `filter` as a string with `x-format` or `x-style` as `json`.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -0,0 +1,80 @@
...
+        items: {
+          type: 'string',
+        },
+      },
+    },
+  };
+
+  // TODO(bajtos) Rework this condition to check whether "model has any
+  // relations defined
+  if (modelCtor !== EmptyModel) {
```
virkt25: When do we expect a `modelCtor` to be an `EmptyModel`. Not sure I follow this logic.

packages/testlab/src/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './test-sandbox';
 export * from './skip-travis';
 export * from './request';
 export * from './http-server-config';
+export * from './misc';
```
bajtos: This is from https://github.com/strongloop/loopback-next/pull/1733, `git` should remove it after #1733 lands and this patch is rebased on top of the new master.

packages/testlab/src/misc.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * JSON transport does not preserve properties that are undefined
+ * As a result, deepEqual checks fail because the expected model
+ * value contains these undefined property values, while the actual
+ * result returned by REST API does not.
+ * Use this function to convert a model instance into a data object
+ * as returned by REST API
+ */
+export function deserializedFromJson<T extends object>(value: T): T {
```
bajtos: Ditto: This is from https://github.com/strongloop/loopback-next/pull/1733, `git` should remove it after #1733 lands and this patch is rebased on top of the new master.

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
```diff
@@ -68,10 +68,10 @@ export class <%= className %>Controller {
     },
   })
   async updateAll(
-    @requestBody() <%= name %>: <%= modelName %>,
+    @requestBody() data: <%= modelName %>,
```
virkt25: Not a big deal but I personally prefer not using a generic name like `data` since it would repeat multiple times and as such would much rather have it be the model name ... `todo: Todo`.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+ * a generic json schema allowing any "where" condition.
+ *
+ * @param modelCtor The model constructor to build the filter schema for.
+ */
+export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
+  const schema: JsonSchema = {
+    properties: {
+      where: {
+        type: 'object',
+        // TODO(bajtos) enumberate "model" properties and operators like "and"
```
virkt25: WIll this be a part of this PR?

docs/site/Controller-generator.md
```diff
@@ -106,43 +114,96 @@ export class TodoController {
...
-  @get('/todos/count')
+  @get('/todos/count', {
+    responses: {
+      '200': {
+        description: 'Todo model count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
   async count(@param.query.string('where') where?: Where): Promise<number> {
```
virkt25: Shouldn't this be changed to `@param.query.object('where')`?

docs/site/todo-list-tutorial-controller.md
```diff
@@ -102,43 +102,105 @@ completed controller should look as follows:
...
-    return await this.todoListRepo.todos(id).create(todo);
+  @post('/todo-lists', {
+    responses: {
+      '200': {
+        description: 'TodoList model instance',
+        content: {'application/json': {'x-ts-type': TodoList}},
+      },
+    },
+  })
+  async create(@requestBody() obj: TodoList): Promise<TodoList> {
```
virkt25: `obj` => `data` as per the new CLI template.

docs/site/todo-list-tutorial-controller.md
```diff
@@ -102,43 +102,105 @@ completed controller should look as follows:
...
+    responses: {
+      '200': {
+        description: 'Array of TodoList model instances',
+        content: {'application/json': {'x-ts-type': TodoList}},
+      },
+    },
+  })
   async find(
-    @param.path.number('id') id: number,
     @param.query.string('filter') filter?: Filter,
```
virkt25: This should be `@param.query.object`

docs/site/Controller-generator.md
```diff
@@ -106,43 +114,96 @@ export class TodoController {
...
+        content: {
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Todo}},
+          },
+        },
+      },
+    },
+  })
+  async find(
+    @param.query.object('filter', getFilterSchemaFor(Todo)) filter?: Filter,
```
virkt25: Out of scope of this PR I'd say but what is your thought on simplifying the DX here by creating the following:

```
@param.query.filter(Todo) => @param.query.object('filter', getFilterSchemaFor(Todo))
@param.query.filter('filterObj', Todo) => @param.query.object('filterObj', getFilterSchemaFor(Todo))
```

I know it goes against the normal `type` flavours we provide for param decorators but I do think it provides a cleaner developer experience.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+        additionalProperties: true,
+      },
+
+      fields: {
+        type: 'object',
+        // TODO(bajtos) enumberate "model" properties
```
virkt25: Will this be a part of this PR?

docs/site/Controller-generator.md
```diff
@@ -106,43 +114,96 @@ export class TodoController {
...
+      '200': {
+        description: 'Todo PATCH success count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
   async updateAll(
-    @requestBody() obj: Todo,
+    @requestBody() data: Todo,
     @param.query.string('where') where?: Where,
```
jannyHou: shouldn't `where` an object instead of string?

docs/site/Controller-generator.md
jannyHou: same here `@param.query.object('where')`

docs/site/todo-list-tutorial-controller.md
```diff
@@ -102,43 +102,105 @@ completed controller should look as follows:
...
+
+  @get('/todo-lists/count', {
+    responses: {
+      '200': {
+        description: 'TodoList model count',
+        content: {'application/json': {'x-ts-type': Number}},
+      },
+    },
+  })
+  async count(@param.query.string('where') where?: Where): Promise<number> {
```
jannyHou: `@param.query.object('where')`

packages/repository/test/unit/decorator/relation.decorator.unit.ts
```diff
@@ -124,3 +124,43 @@ describe('relation decorator', () => {
...
+    @model()
+    class Customer extends User {
+      @hasMany(Order)
+      orders: Order[];
+    }
+
+    const relations = getModelRelations(Customer);
+    expect(relations).to.deepEqual({
+      accessTokens: {
+        keyTo: 'userId',
```
jannyHou: wouldn't it be `customerId`?
Otherwise when we have multiple models extend `User`, they will share the same fk which is `userId`...

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
bajtos: Makes sense, will go with `todo: Todo`.

Before my change, we were using the controller name as the argument name, e.g. `TodoController: Todo`.

docs/site/Controller-generator.md
bajtos: Sounds good to me, I have created a follow-up issue here: https://github.com/strongloop/loopback-next/issues/1749

packages/repository-json-schema/src/filter-json-schema.ts
bajtos: No, see #1748

docs/site/Controller-generator.md
bajtos: > shouldn't `where` an object instead of string?

Good point, will fix.

packages/repository/test/unit/decorator/relation.decorator.unit.ts
bajtos: No, it's `userId` because Customer inherited the relation the User class.

In LB 3.x, we call this a "polymorphic relation" and add another property to `accessToken` to distinguish between different targets that can have the same `userId`. For example, `accessToken` can have two properties: `userId` for the id (PK/FK) value, `userModel` for a name of the target model class (User, Customer, etc.)

The purpose of the test is just to verify that the new function returns all relations - defined on the model and inherited from parents.

bajtos:
@jannyHou @virkt25 thank you for a thorough review, you spotted important places to fix!

In order to support `where` parameters, I added a new helper `getWhereSchemaFor`.

Then I used this new helper in all controller-related code and docs, plus addressed your other comments too.

Could you please take another look and let me know if the changes LGTY now?

bajtos:
I have also created two follow-up stories (post 4.0 GA):
- Enumerate model properties in Filter/Where schema #1748
- Sugar decorators for Filter and Where parameters #1749

packages/openapi-v3/src/filter-schema.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Build a OpenAPI schema describing the format of the "where" object
+ * used to filter model instances to query, update or delete.
+ *
+ * Note we don't take the model properties into account yet and return
+ * a generic json schema allowing any "where" condition.
+ *
+ * @param modelCtor The model constructor to build the filter schema for.
+ */
+export function getWhereSchemaFor(modelCtor: typeof Model): SchemaObject {
```
virkt25: Since we include `getWhereSchemaFor` in this file, would it rename this file to `get-partial-schema`? or something along those lines?

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Build a JSON schema describing the format of the "where" object
+ * used to filter model instances to query, update or delete.
+ *
+ * Note we don't take the model properties into account yet and return
+ * a generic json schema allowing any "where" condition.
+ *
+ * @param modelCtor The model constructor to build the filter schema for.
+ */
+export function getWhereJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
```
virkt25: ditto: Should we consider renaming the file?

packages/repository/test/unit/decorator/relation.decorator.unit.ts
jannyHou: > we call this a "polymorphic relation"
oh ok I got it 👍 it's polymorphic.

packages/openapi-v3/src/filter-schema.ts
bajtos: This file contains functions to build schema of `Filter` and `Where` objects. I feel `get-partial-schema` is misleading, because both `Filter` and `Where` are much more than just an object allowing a subset of model properties. For example, `Where` can contain conditions like `and` and custom operators like `neq`, `gte`, etc.

As I see it, both `Filter` and `Where` are related to querying/filtering model instances. `Where` describes condition(s) to apply, `Filter` provides additional query instructions like pagination and which fields & relations to fetch.

That's why I think `filter-schema` is a good file name.

hadasiddhant:
It would be of great help if anyone here can answer this question for me. Thanks a lot
https://stackoverflow.com/questions/55996451/how-do-i-query-a-particular-field-in-loopback-4-through-the-repository

PR #1680, opened 2018-09-07
fix: soap example

nabdelgadir:
* If using lb4 example to clone soap example, the calculator ds json file gets created

Using lb4 example for the soap calculator would not produce the calculator.datasource.json file in dist8 so the example would not run. This PR is created to fix the issue. 

## Checklist


examples/soap-calculator/index.d.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-todo-list
```
virkt25: The module name should be updated to `@loopback/example-soap-calculator`

examples/soap-calculator/tsconfig.build.json
```diff
@@ -1,14 +1,12 @@
 {
   "$schema": "http://json.schemastore.org/tsconfig",
   "extends": "./node_modules/@loopback/build/config/tsconfig.common.json",
+  "compilerOptions": {
+    "rootDir": "."
+  },
   "include": [
+    "index.ts",
```
virkt25: nitpick: the 3 entries of this array can be on the same line.

nabdelgadir:
> Is the change to .vscode/settings.json intentional?

No, it wasn't intentional, just changed it back.

.vscode/settings.json
```diff
@@ -23,7 +23,6 @@
   },
   "files.insertFinalNewline": true,
   "files.trimTrailingWhitespace": true,
-
```
nabdelgadir: Is the empty line important/necessary here?
It's removed by autosave, but if yes, I can revert the change.

virkt25: No it's not necessary. Probably best to remove it.

examples/soap-calculator/package.json
```diff
@@ -43,14 +43,6 @@
...
   },
   "license": "MIT",
-  "files": [
-    "README.md",
-    "index.js",
-    "index.d.ts",
-    "dist*/src",
-    "dist*/index*",
-    "src"
-  ],
```
raymondfeng: Have we decided to remove the explicit listing of `files` in `package.json` now? IIRC, it was there to exclude certain files to be published. 

If so, we should clean up all packages.

virkt25: None of the other example packages have a `files` field.

dhmlau:
@nabdelgadir, thanks for your first contribution to LB!

PR #1686, opened 2018-09-08
docs: split decorators topic

bschrammIBM:
## Checklist


docs/site/Server.md
```diff
@@ -21,13 +21,12 @@ listening on different ports and working with different protocols.
 
 LoopBack 4 offers the
 [`@loopback/rest`](https://github.com/strongloop/loopback-next/tree/master/packages/rest)
-package out of the box, which provides HTTP/HTTPS based server called
+package out of the box, which provides an HTTP/HTTPS-based server called
```
virkt25: `an` => `a`

docs/site/Decorators.md
```diff
@@ -22,647 +22,7 @@ tie into an API.
...
-- `@referencesOne`
-- `@referencesMany`
-
-Register a specific relation.
-
-_This feature has not yet been released in alpha form. Documentation will be
-added here as this feature progresses._
+- [Route Decorators](Decorators_route.md)
+- [Dependency Injection Decorator](Decorators_inject.md)
+- [Authentication Decorator](Decorators_auth.md)
```
virkt25: I wonder if it's better for us to use `Decorators_authenticate` and `Decorators_repository` as that is the full name of the actual decorator being described by the doc page.

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -162,6 +162,22 @@ children:
     url: Decorators.html
     output: 'web, pdf'
```
virkt25: After this line you need to add a `children: ` to make sure Jekyll knows the pages indented below are child pages. It's also why `verify-docs` is failing on Travis.

docs/site/Decorators.md
bschrammIBM: Renamed

docs/site/sidebars/lb4_sidebar.yml
bschrammIBM: fixed, thanks

docs/site/Server.md
bschrammIBM: "A HTTP" vs "An HTTP" It can go either way, depending upon your pronunciation. I say "aitch" for H; I guess there is an alternate pronunciation that says "haitch". This is from the Chicago Manual of Style "If ... the acronym is treated as a series of letters, the choice of the article depends on the pronunciation of the first letter."

I prefer "An" but really could change it to "a" if you want me to!

PR #1687, opened 2018-09-08
chore: add node version check to fail gracefully

raymondfeng:
Check required Node.js versions and print out meaningful error message if fails.

## Checklist

### DIFF
--- a/packages/dist-util/index.js
+++ b/packages/dist-util/index.js
@@ -6,15 +6,28 @@
 'use strict';
 
 const path = require('path');
+const util = require('util');
+const semver = require('semver');
+
+/**
+ * Make sure node version meets the requirement. This file intentionally
+ * only uses ES5 features so that it can be run with lower versions of Node
+ * to report the version requirement.
+ */
+function checkNodeVersion(range) {
+  const nodeVer = process.versions.node;
+  const requiredVer = range || require('./package.json').engines.node;
+  const ok = semver.satisfies(nodeVer, requiredVer);
+  if (!ok) {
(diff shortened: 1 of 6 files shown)


bajtos:
I find it difficult to review this change because there is no easy way how to show possible changes made while moving most of the code from `bin/cli.js` to `bin/main.js`. This also looses git history of the changes made in this part of CLI code base, rendering `git blame` rather useless.

Have you considered creating a new file, e.g. `cli-with-check.js` or `cli-entry.js` that would contain the newly-added version-check, and keep most of the existing code inside `bin/cli.js` with no (or very little) changes?

raymondfeng:
@bajtos I created `bin/cli-main.js` for `lb4` and moved `bin/cli.js` to `lib/cli.js`. Version histories are now preserved. PTAL.

packages/dist-util/index.js
```diff
@@ -6,15 +6,28 @@
...
 const path = require('path');
+const util = require('util');
+const semver = require('semver');
+
+/**
+ * Make sure node version meets the requirement. This file intentionally
+ * only uses ES5 features so that it can be run with lower versions of Node
+ * to report the version requirement.
+ */
+function checkNodeVersion(range) {
```
bajtos: AFAICT, `range` is not used anywhere. Should we remove it and simplify the implementation a tiny bit?

raymondfeng: It's used here - https://github.com/strongloop/loopback-next/blob/ece15edca2fb3fd5c177967b737bdc9e54365647/packages/cli/bin/cli-main.js#L12-L13

Issue #1688, opened 2018-09-09
Module has no exported member ServiceMixin


siosio34:
## Description / Steps to reproduce / Feature proposal
Hi. i report npm run start error when make todo tutorial in loopback-next document.
this error message is
```
error TS2305 Module has no exported member ServiceMixin.
```

so i enter node_modules/@loopback/service-proxy then no named exported ServiceMixin.  
## Expected Behavior

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

marioestradarosa:
Hi @siosio34, 
The ServiceMixin has just been introduced in one of the recent releases. Do you have the latest @loopback/cli package in your environment?.  

I currently have @loopback/cli version: 0.22.0

Sarh:
Hello, 
I am having this same error. I currently have @loopback/cli version: 0.22.0

marioestradarosa:
I could reproduce it. Even if we have 0.22.0 , the lb4 app is not including the latest package from the npm repository. we´ll fix it. 

For now, can you install it manually in your project? , it should be @0.8.0 , which is the one that contains the ServiceMixin addition.

` npm install @loopback/service-proxy@latest --save`

marioestradarosa:
**As an additional note:** If you run the following command to scaffold the example.

`lb4 example todo`

and then inside the `loopback4-example-todo` directory you try to run it `npm start`.  You will see that it complains with the following error:

`error TS2307: Cannot find module '@loopback/openapi-v3-types/node_modules/openapi3-ts/dist/model/OpenApi` 

I had to install manuall the openapi3-ts dependency inside the openapi-v3-types as follows:

`cd ./node_modules/@loopback/openapi-v3-types`
and then 
`npm install`

After that the scaffolded example will run properly. 
I will open an issue on this one now.

raymondfeng:
@marioestradarosa See https://github.com/Microsoft/TypeScript/issues/26985 and https://github.com/strongloop/loopback-next/issues/1643.

raymondfeng:
@marioestradarosa Pls review https://github.com/strongloop/loopback-next/pull/1689

siosio34:
Ok. this issue is just temporary bug because of loopback cli and service-proxy version difference.  it will be fixed after latest published to npm. i'll wait ~

raymondfeng:
Please try `@loopback/cli@0.22.1`.

[...]

PR #1689, opened 2018-09-09
fix(rest): use direct import to work around a TS bug

raymondfeng:
See https://github.com/Microsoft/TypeScript/issues/26985

Work around https://github.com/strongloop/loopback-next/issues/1688#issuecomment-419688949

## Checklist

### DIFF
--- a/packages/rest/src/keys.ts
+++ b/packages/rest/src/keys.ts
@@ -5,7 +5,12 @@
 
 import {CoreBindings} from '@loopback/core';
 import {BindingKey, Context} from '@loopback/context';
-import {OpenApiSpec} from '@loopback/openapi-v3-types';
+
+/**
+ * See https://github.com/Microsoft/TypeScript/issues/26985
+ */
+// import {OpenApiSpec} from '@loopback/openapi-v3-types';
+import {OpenAPIObject as OpenApiSpec} from 'openapi3-ts';
 
 import {HttpHandler} from './http-handler';
 import {SequenceHandler} from './sequence';
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -42,6 +42,7 @@
     "js-yaml": "^3.11.0",
     "lodash": "^4.17.5",
     "openapi-schema-to-json-schema": "^2.1.0",
+    "openapi3-ts": "^1.0.0",
     "parseurl": "^1.3.2",
     "path-to-regexp": "^2.2.0",
     "qs": "^6.5.2",


marioestradarosa:
@raymondfeng , the current `lb4 app` scaffolds the app correctly with ServiceMixin, howeve it needs `service-proxy": "^0.8.0`   but it writes currently `service-proxy": "^0.7.1`.

 I noticed that the package.json for CLI under templateDependencies is correct `0.8.0` but wasn't published to NPM , it was updated the next day.

 Is that the reason why we are experiencing this effect ?  `0.7.1` instead of  `0.8.0`

marioestradarosa:
I see this PR fixes the other issue with `openapi3-ts`.  but please check my other comment, I think based on the first issue opened, I could see two problems.

raymondfeng:
I suspect that one of the publish failed half way and left the versions inconsistent

packages/rest/src/keys.ts
```diff
@@ -5,7 +5,12 @@
 
 import {CoreBindings} from '@loopback/core';
 import {BindingKey, Context} from '@loopback/context';
-import {OpenApiSpec} from '@loopback/openapi-v3-types';
+
+/**
+ * See https://github.com/Microsoft/TypeScript/issues/26985
+ */
+// import {OpenApiSpec} from '@loopback/openapi-v3-types';
+import {OpenAPIObject} from 'openapi3-ts';
```
bajtos: Can you please rename `OpenAPIObject` to `OpenApiSpec` to avoid changing the places using this type below? We don't want to rename them from `OpenApiSpec` to `OpenAPIObject` now and then rename them back to `OpenApiSpec` when the typescript bug is fixed.

```ts
import {OpenAPIObject as OpenApiSpec} from 'openapi3-ts';
```

raymondfeng: Good idea.

raymondfeng:
@bajtos PTAL

packages/rest/src/keys.ts
```diff
@@ -5,7 +5,12 @@
 
 import {CoreBindings} from '@loopback/core';
 import {BindingKey, Context} from '@loopback/context';
-import {OpenApiSpec} from '@loopback/openapi-v3-types';
+
+/**
+ * See https://github.com/Microsoft/TypeScript/issues/26985
+ */
+// import {OpenApiSpec} from '@loopback/openapi-v3-types';
```
b-admike: can we remove this import statement?

raymondfeng: I intentionally kept it here so that we don't forget to revert it once the TS issue is fixed.

PR #1699, opened 2018-09-11
fix(rest): tidy up host/port parsing and client url building

raymondfeng:
This PR improves host/port parsing from http headers. If the `host` is set and `port` is default, we honor it now.

@virkt25 I think it will fix the issue you have with bluemix where `8080` is added.

## Checklist

### DIFF
--- a/packages/rest/test/integration/rest.server.integration.ts
+++ b/packages/rest/test/integration/rest.server.integration.ts
@@ -391,15 +391,6 @@ paths:
     const app = new Application();
     app.component(RestComponent);
     const server = await app.getServer(RestServer);
-    const greetSpec = {
-      responses: {
-        200: {
-          schema: {type: 'string'},
-          description: 'greeting of the day',
-        },
-      },
-    };
-    server.route(new Route('get', '/greet', greetSpec, function greet() {}));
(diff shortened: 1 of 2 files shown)


packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -416,6 +407,45 @@ paths:
...
+
+    const response = await createClientForHandler(server.requestHandler)
+      .get('/swagger-ui')
+      .set('x-forwarded-proto', 'http')
+      .set('x-forwarded-host', 'example.com:8080,my.example.com:9080');
+    await server.get(RestBindings.PORT);
+    const expectedUrl = new RegExp(
+      [
+        'https://loopback.io/api-explorer',
+        '\\?url=http://example.com:8080/openapi.json',
```
virkt25: Shouldn't this not include the `:8080`?

raymondfeng: It should for this case. For bluemix, I think you use `https://<your-bm-host>/...` with the default port `443` for https. The new code won't accidentally add `8080` as it's from the config, not the header.

PR #983, opened 2018-02-10
[WIP] feat(context): a better/simpler way for binding configuration and @inject.config

raymondfeng:
This PR provides a simpler and better way (IMO) over https://github.com/strongloop/loopback-next/pull/872.

1. Add `context.configure()` to configure bindings
2. Add `context.getConfig()` to look up config for a binding
3. Add `@injection.config()` to receive injection of config

Key design:

1. `configuration` for each binding is now another binding, for example: `servers.rest.server1 --> $config:servers.rest.server1`. The previous design uses `binding.options`, which doesn't allow us to bind configuration before the binding itself is created.

2. Allow hierarchical resolution following the key namespaces. 
  For example, if the binding key is `servers.rest.server1`, we'll try the following entries:
   - `servers.rest.server1:$config#host` (namespace: server1)
   - `servers.rest:$config#server1.host` (namespace: rest)
   - `servers.$config#rest.server1.host` (namespace: server)
   - `$config#servers.rest.server1.host` (namespace: '' - root)
 

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] Related API Documentation was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `packages/example-*` were updated

PR #1705, opened 2018-09-12
feat(rest): set port from env by default

virkt25:
Based on a discussion with @rmg and the experience of deploying the app to cloud, I think we should offer a default based on the `env`. Most cloud providers set `$PORT` as the environment variable. 

Users can still override if needed. 

## Checklist

### DIFF
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -155,8 +155,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
 
     // Can't check falsiness, 0 is a valid port.
     if (config.port == null) {
-      config.port = 3000;
+      config.port = process.env.PORT ? +process.env.PORT : 3000;
     }
+
     if (config.host == null) {
       // Set it to '' so that the http server will listen on all interfaces
       config.host = undefined;


hacksparrow:
LGTM. What about tests?

bajtos:
> Based on a discussion with @rmg and the experience of deploying the app to cloud, I think we should offer a default based on the env. 

+1 for honoring `process.env.PORT` out of the box, as we are already doing in LoopBack 3.x

As I commented above, I would like to see a different implementation that's not hard-coded inside our `@loopback/rest` component.

Nice to have: let the new host/port booter(s) provide a configuration option allowing app developers to provide their own environment variable names to be used for configuring host/port.

bajtos:
If the rest of the team is convinced that `@loopback/rest` should apply `PORT` environment variable, then at minimum, I'd like to see a design where`process.env` is accessed via dependency injection.

I think it's a good idea to access `process.env` via dependency injection in all cases, including a port/host booters described above.

raymondfeng:
From the architecture perspective, I agree with @bajtos:

1. We should have a single source of truth for the resolved configuration. IMO, it's the `Context` in LoopBack 4, which can be used for dependency injection. I see `Context` as the knowledge base.

2. The consumer of the configuration will receive values from the Context while the provider (or resolver) of the configuration should bind/resolve its values to the Context. As a result, the configuration provider and consumer are nicely decoupled.

3. There are various ways to provide the configuration, such as a file, an object for the constructor, and environment variables. We should have a consistent policy on how to consolidate the configuration values from these different sources.

4. We can have booters to proactively populate the `Context` for configuration as constant values or a dynamic `ConfigurationProvider`. I wonder if we should allow a `ConfigurationResolver` to handle all configuration value resolution by namespace and key at low level so that it becomes pluggable.

Now for `PORT`, I see a few options to achieve the separation of concerns:

1. Having the application to honor `process.env.PORT`:
```ts
export class MyApplication extends RestApplication {
  constructor() {
    const config = {
      rest: {
        port: process.env.PORT || 3000,
      }
    }
  }
}
``` 

2. Having the booter to read `process.env.PORT` and bind it to Context as @bajtos suggests.

3. Having a bound service such as `ConfigurationResolver` that controls all configuration access. See some ideas from https://github.com/strongloop/loopback-next/pull/983.

But for the short term, I'm fine to have RestServer checks `process.env.port` directly or via DI.

bajtos:
> Having the application to honor process.env.PORT:
> 
> ```ts
> export class MyApplication extends RestApplication {
>   constructor() {
>    const config = {
>      rest: {
>        port: process.env.PORT || 3000,
>        }
>     }
>  }
> ```

I am ok with this approach as long as `process.env` is injected to the application, for example from the `main` function (see [src/index.ts](https://github.com/strongloop/loopback-next/blob/111442b313ce8d3ba3061afade800ae08e000fc0/examples/todo/src/index.ts#L9-L17)).

That way we can write a unit/integration test for this functionality without having to modify global `process.env`.

```ts
// src/index.ts
export async function main(options?: ApplicationConfig, env?: NodeJS.ProcessEnv) {
  const app = new TodoListApplication(options, env);
  await app.boot();
  await app.start();

  const url = app.restServer.url;
  console.log(`Server is running at ${url}`);
  return app;
}

// src/application.ts
export class MyApplication extends RestApplication {
  constructor(options?: ApplicationConfig = {}, env: NodeJS.ProcessEnv = {}) {
   options.rest = options.res || {};
   options.rest.port = env.PORT || options.rest.port || 3000;
   super(options);
  }
}

// /index.js
if (require.main === module) {
  // Run the application
  application.main(process.env).catch(err => {
    console.error('Cannot start the application.', err);
    process.exit(1);
  });
}
```

Alternatively, we can push the `process.env.PORT` code all the way up to the top-level `index.js` file, but then we won't be able to test this functionality in-process. The benefit I see in this approach compared to the original proposal, is that the test suite is not affected by `process.env` at all.

```js
if (require.main === module) {
  // Run the application
  const config = {
    rest: {
      port: process.env.PORT, 
      host: process.env.HOST
    },
  };
  application.main(config).catch(err => {
    console.error('Cannot start the application.', err);
    process.exit(1);
  });
}

rmg:
This is more of a meta comment on my experience that precipitated this than the implementation..

When I was testing out the deployment workflow, the lack of using `$PORT` was pretty minor compared to not being able to figure out how to actually specify the port. I'm not sure if it's part of some scaffold generation or just a quirk of the example app, but I couldn't find any explicit config being passed anywhere.

I like the idea of sane defaults, but any scaffolded/example code should at least pass in some sort of annotated/commented empty object so it is clear where those defaults should be overridden.

raymondfeng:
> I like the idea of sane defaults, but any scaffolded/example code should at least pass in some sort of annotated/commented empty object so it is clear where those defaults should be overridden.

I always advocate that we generate the complete object of configuration, even for these with default values. It's nice to be self documentary without requiring a developer to search for the reference guide.

raymondfeng:
Close in favor of https://github.com/strongloop/loopback-next/pull/1900

PR #1708, opened 2018-09-13
fix(cli): git-ignore all "dist*" dirs in scaffolded projects

bajtos:
Fix the project template to configure Git to ignore all dist files created by the TypeScript compiler, e.g. `dist8` for Node.js 8.x and `dist10` for Node.js 10.x

Before this change, only the directory `dist` was ignored.

I discovered this problem while testing #1659 

## Checklist

### DIFF
--- a/packages/cli/generators/project/templates/_.gitignore
+++ b/packages/cli/generators/project/templates/_.gitignore
@@ -61,4 +61,4 @@ typings/
 api-docs/
 
 # Transpiled JavaScript files from Typescript
-dist/
+/dist*/


hacksparrow:
Unless there is the possibility of directories which match the pattern `dist*`, this looks fine to me.

bajtos:
> It should be `/dist*/` to only exclude matching root level folders, ideally.

Applied this suggestion.

Please note that we are not using `/` prefix in other places (e.g. `coverage`, etc.)

bajtos:
The Travis CI build is failing on a single platform only because of a timeout during `npm install` or `lerna bootstrap` phase. I am going to overrule the checks and land this PR as-is.

PR #1711, opened 2018-09-14
fix(repository): handle conversion of property values to plain json

raymondfeng:
This PR adds the ability to recursively convert model instances to plain json object

## Checklist

### DIFF
--- a/packages/repository/src/model.ts
+++ b/packages/repository/src/model.ts
@@ -156,7 +156,11 @@ export abstract class Model {
 
     for (const p in def.properties) {
       if (p in this) {
-        json[p] = (this as AnyObject)[p];
+        let val = (this as AnyObject)[p];
+        if (val != null && typeof val.toJSON === 'function') {
+          val = val.toJSON();
+        }
+        json[p] = val;
       }
     }
     return json;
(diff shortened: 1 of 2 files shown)


packages/repository/src/model.ts
```diff
@@ -170,7 +174,13 @@ export abstract class Model {
     if (options && options.ignoreUnknownProperties === false) {
       obj = {};
       for (const p in this) {
-        obj[p] = this[p];
+        let val = (this as AnyObject)[p];
+        if (val != null && typeof val.toObject === 'function') {
+          val = val.toObject();
+        } else if (val != null && typeof val.toJSON === 'function') {
```
b-admike: I guess `toJSON` is our fallback from the `toObject` method? I.e. why do we want to have this `else if` block?

raymondfeng: Use `toJSON()` as the fallback

packages/repository/src/model.ts
```diff
@@ -156,7 +156,11 @@ export abstract class Model {
 
     for (const p in def.properties) {
       if (p in this) {
-        json[p] = (this as AnyObject)[p];
+        let val = (this as AnyObject)[p];
```
jannyHou: @raymondfeng What is the difference between `toJSON` and `toObject` function? I couldn't find related document talking about this 2 methods, shall we add some doc to explain the difference and their usage?

raymondfeng: `toObject()` accepts some options which `toJSON()` does not. I improved the test to show that.

jannyHou: thanks got it.

PR #1712, opened 2018-09-14
fix(repository): convert array items to json

raymondfeng:
A follow-up to #1711. This PR fixes the array conversion.

## Checklist

### DIFF
--- a/packages/repository/src/model.ts
+++ b/packages/repository/src/model.ts
@@ -137,6 +137,32 @@ export class ModelDefinition {
   }
 }
 
+function asJSON(value: any): any {
+  if (value == null) return value;
+  if (typeof value.toJSON === 'function') {
+    return value.toJSON();
+  }
+  // Handle arrays
+  if (Array.isArray(value)) {
+    return value.map(item => asJSON(item));
+  }
+  return value;
+}
+
+function asObject(value: any, options?: Options): any {
+  if (value == null) return value;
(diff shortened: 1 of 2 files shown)


packages/repository/src/model.ts
```diff
@@ -137,6 +137,32 @@ export class ModelDefinition {
...
+  }
+  // Handle arrays
+  if (Array.isArray(value)) {
+    return value.map(item => asJSON(item));
+  }
+  return value;
+}
+
+function asObject(value: any, options?: Options): any {
+  if (value == null) return value;
```
jannyHou: Does `asObject` need check the flag for unknown properties as well like `toObject`?

`if (options && options.ignoreUnknownProperties === false) { // do the convertion}`

packages/repository/src/model.ts
```diff
@@ -137,6 +137,32 @@ export class ModelDefinition {
   }
 }
 
+function asJSON(value: any): any {
+  if (value == null) return value;
```
virkt25: Should there be a similar check for `undefined`?

packages/repository/src/model.ts
raymondfeng: No. The function is only used to handle 1st level properties.

packages/repository/src/model.ts
raymondfeng: Please note `== null` is `true` for both `undefined` and `null`.

Issue #1714, opened 2018-09-15
Error: Cannot find module '/home/aleksei/LoopBack4/getting-started/dist10'


aleksei-timofeyev:
## Description / Steps to reproduce / Feature proposal
Scaffold a new app
lb4 app
cd getting-started
npm start

Possibly related to #1557

## Current Behavior
After scaffolding a new app in lb4 version 0.22.10 and run npm start it complains with the following message:
```
internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module '/home/aleksei/LoopBack4/getting-started/dist10'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.loadDist (/home/aleksei/LoopBack4/getting-started/node_modules/@loopback/dist-util/index.js:36:10)
    at Object.<anonymous> (/home/aleksei/LoopBack4/getting-started/index.js:1:114)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! getting-started@1.0.0 start: `npm run build && node .`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the getting-started@1.0.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/aleksei/.npm/_logs/2018-09-15T13_39_22_695Z-debug.log
```
node --version
v10.10.0

npm --version
6.4.1

## Expected Behavior
The app should start without any problems.

## Temporary workaround 
Make the symlink with name dist10 pointing to dist.

raymondfeng:
npm run build should fix it. Did u create the app using lb4 cli?

aleksei-timofeyev:
> Did u create the app using lb4 cli?
Yes, I did.
npm run build creates dist folder:
```
 npm run build

> getting-started@1.0.0 build /home/aleksei/LoopBack4/getting-started
[...]

PR #1715, opened 2018-09-16
fix(cli): generate correct index.js without loopback build

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/1714

## Checklist

### DIFF
--- a/packages/cli/generators/project/templates/package.plain.json.ejs
+++ b/packages/cli/generators/project/templates/package.plain.json.ejs
@@ -14,7 +14,7 @@
     "build": "tsc --outDir dist --target es2017",
     "build:all-dist": "npm run build",
     "build:watch": "tsc --outDir dist --target es2017 --watch",
-    "clean": "rm -rf dist",
+    "clean": "rimraf dist",
 <% if (project.prettier && project.tslint) { -%>
     "lint": "npm run prettier:check && npm run tslint",
     "lint:fix": "npm run tslint:fix && npm run prettier:fix",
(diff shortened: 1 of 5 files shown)


packages/cli/generators/project/templates/package.plain.json.ejs
```diff
@@ -71,7 +71,6 @@
     "@loopback/context": "<%= project.dependencies['@loopback/context'] -%>",
 <% if (project.projectType === 'application') { -%>
     "@loopback/core": "<%= project.dependencies['@loopback/core'] -%>",
-    "@loopback/dist-util": "<%= project.dependencies['@loopback/dist-util'] -%>",
```
bajtos: This change looks suspicious. For applications scaffolded with `loopbackBuild` feature enabled, the index file is using dist-util to load the correct `dist*` version (both in dev and in production), but `@loopback/dist-util` is not specified as a production dependency.

I think the following scenario will stop working:

```shell
# create an app and verify it works
$ lb4 app
$ npm it

# imagine we are deploying to production now
# and excluding node_modules directory
$ rm -rf node_modules

# now on the production box (or in Docker)
$ npm install --production
$ npm start
```

Could you please check?

IMO, we need to change the template to add dist-util when loopbackBuild is enabled. I believe this is needed for both application and extension projects.

raymondfeng: You probably didn't realize that we use `packages/cli/generators/project/templates/package.plain.json.ejs` when `loopbackBuild` is disabled.

bajtos: You are right, I did not realize we have two copies of `package.json` template. The proposed change makes sense now👍 

For the longer term, it would be great to unify these two templates to reduce the amount of duplicated information.

PR #1723, opened 2018-09-18
feat(repository): rework *ById methods to throw if id not found

bajtos:
_This is a follow-up for #1658 related to #1469._

Rework Repository interface and implementations to throw EntityNotFound error from `updateById`, `replaceById` and `deleteById` when there is no entity with the provided id. 

Update related controller methods to return 204 No Content on success, because there is no boolean flag to indicate success anymore. Clients should check the response status code instead (204 vs 404).

Improve our REST layer to set the response status code to 204 when body is undefined.

This patch is reverting the implementation details of #1621 while keeping the observed REST API behavior.

See the discussion in #1658 and https://github.com/strongloop/loopback-next/pull/1658#issuecomment-421966796 in particular to better understand the reasoning for this change. Also note that soon we will add a new Repository implementation that will provide `*ById` variant using a a special return value to indicate "not found" case. 

## Checklist

### DIFF
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -125,7 +125,14 @@ export class DefaultCrudRepository<T extends Entity, ID>
     this.modelClass = dataSource.createModel<juggler.PersistedModelClass>(
       definition.name,
       properties,
-      Object.assign({strict: true, strictDelete: true}, definition.settings),
+      Object.assign(
+        // settings that users can override
+        {strict: true},
+        // user-defined settings
+        definition.settings,
+        // settings enforced by the framework
+        {strictDelete: false},
+      ),
     );
     this.modelClass.attachTo(dataSource);
   }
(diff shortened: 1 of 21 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -235,21 +240,33 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+  async updateById(
+    id: ID,
+    data: DataObject<T>,
+    options?: Options,
+  ): Promise<void> {
     const idProp = this.modelClass.definition.idName();
     const where = {} as Where;
     where[idProp] = id;
-    return this.updateAll(data, where, options).then(count => count > 0);
+    const count = await this.updateAll(data, where, options);
```
jannyHou: we don't return `count` anymore?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -216,11 +221,11 @@ export class DefaultCrudRepository<T extends Entity, ID>
     return this.toEntity(model);
   }
 
-  update(entity: T, options?: Options): Promise<boolean> {
+  update(entity: T, options?: Options): Promise<void> {
```
jannyHou: I think it should call `return await this.updateById(entity.getId(), entity, options)`, same for the next function `delete`.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -216,11 +221,11 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
     return this.updateById(entity.getId(), entity, options);
   }
 
-  delete(entity: T, options?: Options): Promise<boolean> {
+  delete(entity: T, options?: Options): Promise<void> {
```
jannyHou:
```ts
return await this.deleteById(entity.getId(), options);
```

packages/repository/src/repositories/legacy-juggler-bridge.ts
raymondfeng: It's not necessary.

PR #1731, opened 2018-09-19
[WIP] feat: support built-in JavaScript/Node schema types

hacksparrow:
Add support for built-in JavaScript/Node schema types in Model
properties

## Checklist


packages/repository-json-schema/test/unit/build-schema.unit.ts
```diff
@@ -88,24 +88,72 @@ describe('build-schema', () => {
...
+    it('converts Number', () => {
       expect(metaToJsonProperty({type: Number})).to.eql({
         type: 'number',
       });
     });
 
-    it('converts arrays', () => {
+    it('converts Date', () => {
+      expect(metaToJsonProperty({type: Date})).to.eql({
+        type: 'date',
```
bajtos: I think JavaScript `Date` should be described as `date-time`, because `Date` holds both date *and* time.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#data-types


Common Name | type | format | Comments
-- | -- | -- | --
date | `string` | `date` | As defined by full-date - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
dateTime | `string` | `date-time` | As defined by date-time - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)

LoopBack 3.x describes `Date` properties as `{type: 'string', format: 'date-time'}` too, see loopback-swagger's [lib/specgen/schema-builder.js](https://github.com/strongloop/loopback-swagger/blob/c40420ecb2da2044202dda6d332f46377e318f0d/lib/specgen/schema-builder.js#L103-L106):

```js
    case 'date':
      schema.type = 'string';
      schema.format = 'date-time';
      break;
```

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -125,6 +126,10 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
 
   if (isComplexType(propertyType)) {
     Object.assign(propDef, {$ref: `#/definitions/${propertyType.name}`});
+  } else if (propertyType === Date) {
+    Object.assign(propDef, {
+      type: 'date-time',
```
bajtos:
```js
type: 'string', 
format: 'date-time',
```

bajtos: @hacksparrow before you fix this: please look around the tests and try to find a test that runs JSON Schema validation against the JSON/OpenAPI schema we are producing.

Ideally, if our schema builder emits an invalid schema (as you did here), then there should be a test that fails and reports the problem.

bajtos: Maybe we have an acceptance test that needs to be expanded with the newly supported types?

hacksparrow: Noted.

hacksparrow:
@bajtos made the changes you suggested.

raymondfeng:
Please fix the lint issue before merging - https://travis-ci.org/strongloop/loopback-next/jobs/431549108

PR #1733, opened 2018-09-20
feat(testlab): add a new helper `toJSON`

bajtos:
Implement a helper function to convert an object (e.g. a Model instance) to a data object representing JSON version of the input object.

Specifically, properties set to a value that cannot be represented in
JSON (e.g. `undefined`) are removed from the result.

## Checklist

- [x] Documentation in ~~[/docs/site](../tree/master/docs/site)~~ testlab's README was updated

### DIFF
--- a/packages/testlab/README.md
+++ b/packages/testlab/README.md
@@ -53,6 +53,7 @@ Table of contents:
 - [givenHttpServerConfig](#givenhttpserverconfig) - Generate HTTP server config.
 - [httpGetAsync](#httpgetasync) - Async wrapper for HTTP GET requests.
 - [httpsGetAsync](#httpsgetasync) - Async wrapper for HTTPS GET requests.
+- [toJSON](#toJSON) - A helper to obtain JSON data representing a given object.
 
 ### `expect`
 
@@ -132,6 +133,24 @@ Async wrapper for making HTTPS GET requests.
 ```ts
 import {httpsGetAsync} from '@loopback/testlab';
 const response = await httpsGetAsync('https://example.com');
+```
+
+### `toJSON`
+
+JSON encoding does not preserve properties that are undefined. As a result,
+`deepEqual` checks fail because the expected model value contains these
+undefined property values, while the actual result returned by REST API does
+not. Use this function to convert a model instance into a data object as
+returned by REST API.
+
+```ts
+import {createClientForHandler, toJSON} from '@loopback/testlab';
+
+it('gets a todo by ID', () => {
+  return client
+    .get(`/todos/${persistedTodo.id}`)
+    .expect(200, toJSON(persistedTodo));
+});
 ```
 
 #### Test request parsing
--- a/packages/testlab/src/index.ts
+++ b/packages/testlab/src/index.ts
@@ -12,3 +12,4 @@ export * from './test-sandbox';
 export * from './skip-travis';
 export * from './request';
 export * from './http-server-config';
+export * from './to-json';
--- a/packages/testlab/src/to-json.ts
+++ b/packages/testlab/src/to-json.ts
@@ -1,0 +1,39 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/testlab
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// Important! Date.prototype.toJSON() returns a string.
+export function toJSON(value: Date): string;
+
+// Important! Functions cannot be encoded in JSON.
+export function toJSON(value: Function): undefined;
+
+// Distinguish arrays from objects (an array is an object too)
+export function toJSON(value: unknown[]): unknown[];
+
+/**
+ * JSON encoding does not preserve properties that are undefined
+ * As a result, deepEqual checks fail because the expected model
+ * value contains these undefined property values, while the actual
+ * result returned by REST API does not.
+ * Use this function to convert a model instance into a data object
+ * as returned by REST API
+ */
+export function toJSON(value: object): object;
+
+// The following overloads are provided for convenience.
+// In practice, they should not be necessary, as they simply return the input
+// value without any modifications.
+
+// tslint:disable-next-line:unified-signatures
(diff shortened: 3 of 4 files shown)


packages/testlab/src/misc.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/testlab
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * JSON transport does not preserve properties that are undefined
```
raymondfeng: We should say `JSON format` or `JSON encoding` instead of `transport`.

packages/testlab/src/misc.ts
```diff
@@ -0,0 +1,16 @@
...
+ * As a result, deepEqual checks fail because the expected model
+ * value contains these undefined property values, while the actual
+ * result returned by REST API does not.
+ * Use this function to convert a model instance into a data object
+ * as returned by REST API
+ */
+export function deserializedFromJson<T extends object>(value: T): T {
```
raymondfeng: The method name is a bit misleading. It tricks me to think about the argument is a JSON string.

What about `asJSON()` or `toJSON()`?

Do we want to restrain the argument to be an `object`? What about `any`?

We need to deal with the case where value is `undefined`. 
```ts
JSON.parse(JSON.stringify(undefined)); // SyntaxError: Unexpected token u in JSON at position 0
```
The return type is incorrect, for example:

```ts
// Customer is a class with `getFullName()` method.
const customer: Customer = await this.repo.findById('c001');
const json = deserializedFromJson(customer); // TypeScript infers the return type as `Customer`
json.getFullName(); // The function does not exist but TypeScript cannot catch it.
```

I propose that we change the method to be:
```
export function toJSON<T extends object>(value: T): DeepPartial<T> { // ... }
```

virkt25: +1 to using either `asJSON` or `toJSON` ... with a slight preference for `toJSON`. 

```
// Customer is a class with `getFullName()` method.
const customer: Customer = await this.repo.findById('c001');
const json = deserializedFromJson(customer); // TypeScript infers the return type as `Customer`
json.getFullName(); // The function does not exist but TypeScript cannot catch it.
```

I don't think `json.getFullName();` is a valid use of this as the intention of this method is to act as a helper to allow comparison of an object to the value you would get from a rest api ... which is the use case for testing (and the inspiration for this helper method). One should just do `customer.getFullName()` instead of converting to JSON and then trying to treat it as a class. 

Considering the above maybe a better name we can use is `modelToJSON`?

raymondfeng: @virkt25 Please note by the current signature, `deserializedFromJson(customer); ` returns a `Customer`, which has `getFullName()`. As a result, TypeScript thinks `json.getFullName()` is valid use but the `getFullName` does not exist on the object.

bajtos:
> `export function toJSON<T extends object>(value: T): DeepPartial<T> { // ... }`

I think this isn't going to really fix the problem, because `DeepPartial` will only make all methods optional - it is not going to hide them from the result.

I solved the problem by returning a generic `object` type. This type does not have any properties nor methods beyond built-in `Object.prototype`. This is good enough for the use case of comparing data via deep-equal.

I have also renamed the helper function to `toJSON` and added support for other value types, from `undefined` and `null` to scalar types like numbers & strings.

PTAL again, LGTY now?

packages/testlab/test/unit/misc.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expectBooleanValue(value, true);
+  });
+
+  it('handles dates', () => {
+    const value = toJSON(new Date('2018-01-01T00:00:00.000Z'));
+    expectStringValue(value, '2018-01-01T00:00:00.000Z');
+  });
+});
+
+// Helpers to enforce compile-time type checks
```
bajtos: Should I move these helpers to `src/expect.js` and export them as part of testlab's public API?

packages/testlab/test/unit/misc.test.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/testlab
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '../../src/expect';
+import {toJSON} from '../../src/misc';
+
+describe('toJSON', () => {
```
raymondfeng: Please add the following test cases:
- class Customer: a customer instance => plain json
- class Customer and Address: a customer with address instance => plain json
- Arrays (top level or embedded)
- A function, such as `function x() {}` - I think it will throw `SyntaxError: Unexpected token u in JSON at position 0` as `JSON.stringify(x)` produces `undefined`.

packages/testlab/test/unit/misc.test.ts
raymondfeng: Let's keep the list of helpers internal for now. Too many public helpers will be hard to maintain.

raymondfeng:
@bajtos Please address my latest comments.

packages/testlab/test/unit/misc.test.ts
bajtos: > class Customer: a customer instance => plain json
> class Customer and Address: a customer with address instance => plain json

So far, `@loopback/testlab` does not depend on `@loopback/repository`. Are we ok with adding this dependency?

> Arrays (top level or embedded)

Makes sense 👍 

> A function, such as function x() {}

I think this will be rejected by TypeScript compiler, but I can add a check for this to accommodate for people calling this API from JavaScript.

raymondfeng:
> So far, @loopback/testlab does not depend on @loopback/repository. Are we ok with adding this dependency?

No need to use `@model`. Just plain JS classes will do.

bajtos:
@raymondfeng added more tests as you requested and simplified the actual implementation by wrapping the input value in an object property. LGTY now?

PR #1734, opened 2018-09-20
feat(testlab): add createRestAppClient(), simplify usage in tests

bajtos:
While reviewing https://github.com/strongloop/loopback4-example-shopping/pull/12, I noticed the test code is depending on low-level types from `supertest` that are difficult to discover (in my experience). I decided to fix this and introduce a type alias, only to find that we already have `Client` alias defined! As I was replacing supertest type with `Client` alias,  I found few places calling `supertest` directly instead of using our `createClient*` helper APIs and decided to improve that too. Here is the outcome:

- Clean up all test code (including examples in the documentation) to use `Client` instead of `supertest.SuperTest<supertest.Test>`.
- Remove `createClientForRestServer` because it was not used anywhere and had the issue of leaking a listening server (not stopping the server after the test is done).
- Add a new helper `createRestAppClient` instead, rework acceptance tests (including the templates) to use this new helper.

## Checklist

- [x] Documentation in ~~[/docs/site](../tree/master/docs/site)~~ testlab's README was updated

### DIFF
--- a/docs/site/Implementing-features.shelved.md
+++ b/docs/site/Implementing-features.shelved.md
@@ -40,11 +40,11 @@ Create `test/acceptance/product.acceptance.ts` with the following contents:
 ```ts
 import {HelloWorldApp} from '../..';
 import {RestBindings, RestServer} from '@loopback/rest';
-import {expect, supertest} from '@loopback/testlab';
+import {Client, expect, supertest} from '@loopback/testlab';
 
 describe('Product (acceptance)', () => {
   let app: HelloWorldApp;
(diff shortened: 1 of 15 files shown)


bajtos:
> It's marked as a feature ... I wonder if it should also be marked as a breaking change in the commit message? ... that said the effect is the same for feat and breaking change while on 0.x.y versioning.

Unfortunately, when `lerna publish` encounters a breaking changes, it bump the version to 1.0.0 :(

See _Build: `lerna publish` upgrades from `0.x.y` to `1.0.0` for breaking changes #1068_

PR #1735, opened 2018-09-20
[WIP] feat(repository): add UncheckedCrudRepository

bajtos:
_This is a follow-up for #1658 related to #1469. See the discussion in #1658 and https://github.com/strongloop/loopback-next/pull/1658#issuecomment-421966796 in particular to better understand the reasoning for this change._

Add a new Repository interface and implementation that provides an alternative "not found error" handling. Instead of throwing `EntityNotFound` error, repository methods use return value (`undefined` or `false`) to signal "not found" situation.

I am opening this pull request to get early feedback. The initial commit contains the proposed new interface and the first cut on the implementation providing `repository.save` method only.

## Checklist

### DIFF
--- a/packages/repository/src/repositories/unsafe-crud-repository.ts
+++ b/packages/repository/src/repositories/unsafe-crud-repository.ts
@@ -1,0 +1,186 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  DataObject,
+  Options,
+  Command,
+  NamedParameters,
+  PositionalParameters,
+  AnyObject,
+} from '../common-types';
+import {Entity} from '../model';
+import {Filter, Where} from '../query';
+import {
+  CrudRepository,
+  EntityCrudRepository,
+  EntityRepository,
+} from './repository';
+import {isEntityNotFoundError} from '../errors';
+
+/**
+ * CRUD operations for a repository of entities.
+ *
+ * Operations affecting a single entity signal "not found" error by returning
+ * `undefined`, as opposed to EntityCrudRepository that throws EntityNotFound
+ * error.
+ */
+export interface UnsafeCrudRepository<T extends Entity, ID>
+  extends EntityRepository<T, ID>,
+    CrudRepository<T> {
+  /**
+   * Save an entity. If no id is present, create a new entity
+   * @param entity Entity to be saved
+   * @param options Options for the operations
+   * @returns A promise resolved with the updated entity or `undefined` if `entity`
+   * has ID set but there is no record with such ID in the database.
+   */
+  save(entity: DataObject<T>, options?: Options): Promise<T | undefined>;
+
+  /**
+   * Update an entity
+   * @param entity Entity to be updated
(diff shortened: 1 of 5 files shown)


packages/repository/src/repositories/unsafe-crud-repository.ts
```diff
@@ -0,0 +1,186 @@
...
+import {isEntityNotFoundError} from '../errors';
+
+/**
+ * CRUD operations for a repository of entities.
+ *
+ * Operations affecting a single entity signal "not found" error by returning
+ * `undefined`, as opposed to EntityCrudRepository that throws EntityNotFound
+ * error.
+ */
+export interface UnsafeCrudRepository<T extends Entity, ID>
```
raymondfeng: We should name it something like `UncheckedEntityRepository`. `Unsafe` really scares me :-).

bajtos: Good suggestion, I agree `Unchecked` is better than `Unsafe`.

Here is my reasoning why I prefer to exclude `Entity` and include `Crud` in the interface name:
 - I feel that different requirements for handling "does not exist" are specific to entities, because they have an `id` that's used to lookup an existing record. I don't see how a "not found" situation can happen for a model that does not have any unique identifier. As a result, all "unchecked" repositories are going to deal with Entities.
 - I am assuming there is a reason why we have `EntityCrudRepository` instead of shorter `EntityRepository` and we are envisioning to have non-CRUD entity repositories in the future (or in 3rd party code). Maybe KeyValue is a good example? As a result, I want to make it clear that the interface I am implementing here is providing CRUD access.

Going forward, I'll use the name `UncheckedCrudRepository`.

hacksparrow:
I am assuming unit tests would be added later?

packages/repository/test/integration/repositories/unchecked-crud-repository.integration.ts
```diff
@@ -0,0 +1,78 @@
...
+  DefaultCrudRepository,
+  Entity,
+  juggler,
+  UncheckedCrudRepository,
+  UncheckedCrudRepositoryImpl,
+  model,
+} from '../../..';
+import {property} from '../../../src';
+
+describe('UnsafeCrudRepository', () => {
```
raymondfeng: Please change the test name too.

packages/repository/test/integration/repositories/unchecked-crud-repository.integration.ts
```diff
@@ -0,0 +1,78 @@
...
+  const DUMMY_OPTIONS = {test: true};
+
+  let unsafeRepository: UncheckedCrudRepository<
```
raymondfeng: var name => uncheckedRepo

raymondfeng:
Please replace all `unsafe` to `unchecked`

bajtos:
> I am assuming unit tests would be added later?

Well I started to write a unit test using mocks, but ended up with mocking so much behavior that the test became rather useless. That's why I decided to go with an integration test.

bajtos:
@raymondfeng what's your opinion about the priority of this feature? Is it something deserving to be finished reasonably soon, or should we close this pull request for now, until there is more user demand?

bajtos:
I am closing this pull request for now.

Issue #1312, opened 2018-05-09
CLI: a command to create a new service


bajtos:
_This is a follow-up for #1267_

There should be a new CLI command for creating services, this command should:
 - (MUST HAVE) Create the datasource.
 - (MUST HAVE) Create any artifacts needed to register the service with the app for dependency injection. Leverage the Service booter implemented by #1439
 - (SHOULD HAVE) Create any artifacts needed to write an integration test for the service as I shown in #1311.
 - (NICE TO HAVE) Ideally, include a scaffolding making it easier for app developers to start writing smoke tests.

bajtos:
> The only artifact we need for service proxy is the data source for a service-oriented connector. I don't see the immediate requirement to make service declarative beyond the data source itself.

See the example in #1311. When writing tests of a service proxy, users should not need to bootstrap the full app in order to access the service proxy instance. Ideally, the tests should not need to know how the service proxy is created at all - the creation aspect should be captured in one place only.

```ts
import {GeoService} from '../services/geo.service';
import {expect} from '@loopback/testlab';

describe('GeoService', () => {
  let service: GeoService;
  beforeEach(() => service = new GeoService());

  describe('geocode', () => {
    it('returns geo coords for a given address', () => {
      const result = await service.geocode('107 S B St', 'San Mateo', '94401');
      // { lat: 37.5669986, lng: -122.3237495 }
     expect(result.lat).approximately(37.5669986, 0.5);
     expect(result.lng).approximately(-122.3237495, 0.5);
    });
  });
});
```

bajtos:
I think this task should wait until https://github.com/strongloop/loopback-next/issues/1439 (Sugar API for registering services + automated registration via boot) is done. See the code in `examples/todo` added by https://github.com/strongloop/loopback-next/pull/1347 for a reference implementation that our CLI template should scaffold.

jannyHou:
cc @strongloop/sq-lb-apex @raymondfeng @hacksparrow @bajtos 
The Geo services is added to `example/todo` and creating/booting a service is also supported now.
I am good with using the story description as acceptance criteria.

Any more comments before getting it ready for estimation?

virkt25:
Is the story just creating a datasource file (json/ts) ... ?

bajtos:
I have slightly updated the issue description to mention that Service booter is already implemented, and also marked the last two items are should-have/nice-to-have.

[...]

PR #1736, opened 2018-09-21
feat(cli) add basic lb4 service 

marioestradarosa:
This PR implements the basic generation of a service class, as discussed in https://github.com/strongloop/loopback-next/issues/1312#issuecomment-423224230.  

## Done 
- **Generates** a **basic** _service class_
- Add **Integration test** 
- Add **Documentation**  for `lb4 service`
- **Move** the datasource file inspector to the utils library to be consumed by other artifacts such as lb4 repository, _this way we don't duplicate code_
- **Add* a function to return the connector type, which is basic for future inferring of remote methods, so the function returns soap or rest based on the DS name.
- **Share** `addGivenLB4Project` now with repository and service generators. This way we separate concerns and make the code reading and reviewing more efficient
- **Move** the fixtures to their respective `fixtures` directory for both `repository` and `service` generators
- **Install** the @loopback/proxy-service package if missing from package.json

## Checklist

### DIFF
--- a/packages/cli/generators/service/index.js
+++ b/packages/cli/generators/service/index.js
@@ -80,13 +80,14 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
     if (this.options.name) {
       Object.assign(this.artifactInfo, {name: this.options.name});
     }
+
     const prompts = [
       {
         type: 'input',
         name: 'name',
         // capitalization
         message: utils.toClassName(this.artifactInfo.type) + ' name:',
-        when: this.artifactInfo.name === undefined,
+        when: !this.artifactInfo.name,
         validate: utils.validateClassName,
       },
     ];
@@ -198,14 +199,13 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
   scaffold() {
     if (this.shouldExit()) return false;
 
-    this.artifactInfo.dataSourceClassName = this.artifactInfo.dataSourceClass.replace(
-      'Datasource',
-      'DataSource',
+    this.artifactInfo.dataSourceName = utils.getDataSourceName(
+      this.artifactInfo.datasourcesDir,
+      this.artifactInfo.dataSourceClass,
     );
 
-    this.artifactInfo.dataSourceName = utils.lowerCase(
-      this.artifactInfo.dataSourceClass.replace('Datasource', ''),
-    );
+    this.artifactInfo.dataSourceClassName =
+      this.artifactInfo.dataSourceName + 'DataSource';
 
     // Setting up data for templates
     this.artifactInfo.className = utils.toClassName(this.artifactInfo.name);
--- a/packages/cli/generators/service/index.js
+++ b/packages/cli/generators/service/index.js
@@ -77,7 +77,6 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
    * Ask for Service Name
    */
   async promptArtifactName() {
-    if (this.shouldExit()) return false;
     debug('Prompting for service name');
 
     if (this.options.name) {
@@ -100,8 +99,6 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
   }
 
   async promptDataSourceName() {
-    if (this.shouldExit()) return false;
-
     debug('Prompting for a datasource ');
     let cmdDatasourceName;
     let datasourcesList;
--- a/packages/cli/generators/service/index.js
+++ b/packages/cli/generators/service/index.js
@@ -63,10 +63,11 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
     if (!fs.existsSync(this.artifactInfo.datasourcesDir)) {
       return this.exit(
         new Error(
-          `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.
-        ${chalk.yellow(
-          'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
-        )}`,
+          `${ERROR_NO_DATA_SOURCES_FOUND} ${
+            this.artifactInfo.datasourcesDir
+          }.${chalk.yellow(
+            'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
+          )}`,
         ),
       );
     }
@@ -142,10 +143,11 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
     if (availableDatasources.length === 0) {
       return this.exit(
         new Error(
-          `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.
-        ${chalk.yellow(
-          'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
-        )}`,
+          `${ERROR_NO_DATA_SOURCES_FOUND} ${
+            this.artifactInfo.datasourcesDir
+          }.${chalk.yellow(
+            'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
+          )}`,
         ),
       );
     }
@@ -183,7 +185,9 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
   /**
    * this method will be in charge of inferring and create
    * the remote interfaces from either SOAP wsdl or REST openApi json
-   * TODO: @marioestradarosa
+   *
+   * TODO: add functionality to inspect service specs to generate
+   * strongly-typed service proxies and corresponding model definitions.
    */
   async inferInterfaces() {
     let connectorType = utils.getDataSourceConnectorName(
--- a/packages/cli/generators/service/index.js
+++ b/packages/cli/generators/service/index.js
@@ -61,14 +61,12 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
   async checkPaths() {
     // check for datasources
     if (!fs.existsSync(this.artifactInfo.datasourcesDir)) {
-      return this.exit(
-        new Error(
-          `${ERROR_NO_DATA_SOURCES_FOUND} ${
-            this.artifactInfo.datasourcesDir
-          }.${chalk.yellow(
-            'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
-          )}`,
-        ),
+      new Error(
+        `${ERROR_NO_DATA_SOURCES_FOUND} ${
+          this.artifactInfo.datasourcesDir
+        }. ${chalk.yellow(
+          'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
+        )}`,
       );
     }
   }
@@ -142,7 +140,7 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
         new Error(
           `${ERROR_NO_DATA_SOURCES_FOUND} ${
             this.artifactInfo.datasourcesDir
-          }.${chalk.yellow(
+          }. ${chalk.yellow(
             'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
           )}`,
         ),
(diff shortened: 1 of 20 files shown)


packages/cli/generators/service/index.js
```diff
@@ -0,0 +1,242 @@
...
+    return super.setOptions();
+  }
+
+  async checkPaths() {
+    // check for datasources
+    if (!fs.existsSync(this.artifactInfo.datasourcesDir)) {
+      return this.exit(
+        new Error(
+          `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.
+        ${chalk.yellow(
```
bajtos: I find this line break rather suspicions, because the indenting of the second line is going to be preserved in the message.

```
No datasources found at src/datasources.
         Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered
```

Please find a better way how to compose this error message. For example:

```ts
const message = [
  `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.`,
  chalk.yellow(
    'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information ' + 
      'on how datasources are discovered'
  ),
].join('\n');
return this.exit(new Error(message));
```

packages/cli/generators/service/index.js
```diff
@@ -0,0 +1,242 @@
...
+      .catch(err => {
+        debug(`Error during datasource prompt: ${err}`);
+        return this.exit(err);
+      });
+  }
+
+  /**
+   * this method will be in charge of inferring and create
+   * the remote interfaces from either SOAP wsdl or REST openApi json
+   * TODO: @marioestradarosa
```
bajtos: What is remaining to do? Could you please add some content to the `TODO` line to explain what needs to be done?

packages/cli/generators/service/templates/service-template.ts.ejs
```diff
@@ -0,0 +1,24 @@
+import {getService, juggler} from '@loopback/service-proxy';
+import {inject, Provider} from '@loopback/core';
+import {<%= dataSourceClassName %>} from '../datasources';
+
+/**
+ * this is where you define the Node.js methods that will be
```
bajtos: Could you please move this comment into the class and also don't use `/**` because that's for `tsdoc`.

```
export interface <%= className %>Service {
  // Declare methods provided by the remote service here.
  // These methods should map 1:1 to operations configured 
  // in and provided by the datasource (e.g. SOAP operations).
}
```

packages/cli/lib/utils.js
```diff
@@ -386,8 +388,96 @@ exports.getRepositoryFileName = function(repositoryName) {
...
+    datasourcesDir,
+    dataSourceClass.replace('Datasource', '.datasource.json').toLowerCase(),
+  );
+
+  debug(`reading ${datasourceJSONFile}`);
+  try {
+    jsonFileContent = JSON.parse(fs.readFileSync(datasourceJSONFile, 'utf8'));
+  } catch (err) {
+    debug(`Error reading file ${datasourceJSONFile}: ${err.message}`);
+    throw new Error(err);
```
bajtos: Don't create a new Error object, it will discard the original stack trace and other useful information from `err`. Just `throw err`.

packages/cli/lib/utils.js
```diff
@@ -386,8 +388,96 @@ exports.getRepositoryFileName = function(repositoryName) {
...
```
bajtos: Ditto, just `throw err`.

marioestradarosa:
@bajtos PTAL

marioestradarosa:
>> I am little bit concerned about the amount of code needed to setup the test environment (sandbox)

Yes, that will become a shared function with lb4 repository in a separate PR.

packages/cli/generators/service/index.js
```diff
@@ -63,10 +63,11 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
     if (!fs.existsSync(this.artifactInfo.datasourcesDir)) {
       return this.exit(
         new Error(
-          `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.
-        ${chalk.yellow(
-          'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
-        )}`,
+          `${ERROR_NO_DATA_SOURCES_FOUND} ${
+            this.artifactInfo.datasourcesDir
+          }.${chalk.yellow(
```
bajtos: Missing space after the dot character? `}. ${`

packages/cli/generators/service/index.js
```diff
@@ -142,10 +143,11 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
     if (availableDatasources.length === 0) {
       return this.exit(
         new Error(
-          `${ERROR_NO_DATA_SOURCES_FOUND} ${this.artifactInfo.datasourcesDir}.
-        ${chalk.yellow(
-          'Please visit https://loopback.io/doc/en/lb4/DataSource-generator.html for information on how datasources are discovered',
-        )}`,
+          `${ERROR_NO_DATA_SOURCES_FOUND} ${
+            this.artifactInfo.datasourcesDir
+          }.${chalk.yellow(
```
bajtos: ditto

packages/cli/test/test-utils.js
```diff
@@ -63,10 +63,13 @@ exports.executeGenerator = function(GeneratorOrNamespace, settings) {
  * @property {boolean} excludeDataSourcesDir Excludes the datasources directory
  * @property {boolean} includeDummyModel Creates a dummy model file in /src/models/product-review.model.ts
  * @property {boolean} includeDummyRepository Creates a dummy repository file in /src/repositories/bar.repository.ts
+ * @property {boolean} includeSandboxFilesFixtures creates files specified in SANDBOX_FILES array
+ * @param {Object[]} sandBoxFiles specify files, directories and their content to be included as fixtures
  */
-exports.givenLBProject = function(rootDir, options) {
+exports.givenLBProject = function(rootDir, options, sandBoxFiles) {
```
bajtos: I find it redundant that we have to specify both `options.includeSandboxFilesFixture` and `sandBoxFiles`. If `sandBoxFiles` is not provided or empty, then `options.includeSandboxFilesFixture` can be considered as `false`. Similarly, if `sandBoxFiles` are provided and not empty, then `options.includeSandboxFilesFixture` can be considered as true. Any other combination does not make sense to me and I would consider it as a programmer's error.

I am proposing to merge `options.includeSandboxFilesFixture` and `sandBoxFiles` into `options.additionalFiles`.

Example usage:

```ts
testUtils.givenLBProject(
  SANDBOX_PATH,
  {additionalFiles: SANDBOX_FILES}
);
```

packages/cli/test/fixtures/service/index.js
```diff
@@ -0,0 +1,61 @@
+const DATASOURCE_APP_PATH = 'src/datasources';
+const CONFIG_PATH = '.';
+const DUMMY_CONTENT = '--DUMMY VALUE--';
+
+exports.SANDBOX_FILES = [
+  {
+    path: CONFIG_PATH,
+    file: 'mysoapconfig.json',
+    content: `{
```
bajtos: Let's use `JSON.stringify` instead of a string template literal.

Ideally, the code creating these sandbox files should be smart enough to call `JSON.stringify(content, null, 2)` when `typeof content === 'object'`.

packages/cli/test/fixtures/service/index.js
```diff
@@ -0,0 +1,61 @@
...
+      "name": "MultiWordService",
+      "datasource": "myds"
+    }`,
+  },
+  {
+    path: CONFIG_PATH,
+    file: 'myrestconfig.json',
+    content: `{
```
bajtos: Ditto. Let's use `JSON.stringify` instead of a string template literal.

packages/cli/test/fixtures/repository/index.js
```diff
@@ -0,0 +1,161 @@
...
+      constructor(data?: Partial<MultiWord>) {
+        super(data);
+      }
+    }
+    `,
+  },
+  {
+    path: MODEL_APP_PATH,
+    file: 'invalid-id.model.ts',
+    content: `
```
bajtos: Maintaining source code in a string template literal is cumbersome. Can you move the source code into a standalone file please, e.g. `packages/cli/test/fixtures/repository/models/invalid-id.model.ts`.

Then you can e.g. read the file here:

```ts
content: fs.readFileSync(require.resolve('./models/invalid-id.model.ts'), {encoding: 'utf-8'}),
```

It would be even better if the code creating sandbox directory was able to copy a tree of files from fixtures directory, so that these instructions in `SANDBOX_FILES` can say something along the lines of "recursively copy content of the directory `packages/cli/test/fixtures/repository`".

marioestradarosa:
@raymondfeng  PTAL

raymondfeng:
@marioestradarosa great work as usual!

I found a bug which can be reproduced as follows:

1. Use `lb4 datasource` to create REST ds named `MapDS`. It creates `src/datasources/map-ds.datasource.json`
2. Run `lb4 service` and it crashes:
```
Error: ENOENT: no such file or directory, open '/private/tmp/t2/src/datasources/mapds.datasource.json'
    at Object.openSync (fs.js:436:3)
    at Object.readFileSync (fs.js:341:35)
    at Object.exports.isConnectorOfType (/Users/rfeng/Projects/loopback4/loopback-next/packages/cli/lib/utils.js:458:37)
    at datasourcesList.filter.item (/Users/rfeng/Projects/loopback4/loopback-next/packages/cli/generators/service/index.js:130:28)
    at Array.filter (<anonymous>)
    at ServiceGenerator.promptDataSourceName (/Users/rfeng/Projects/loopback4/loopback-next/packages/cli/generators/service/index.js:126:50)
```

packages/cli/generators/service/templates/service-template.ts.ejs
```diff
@@ -0,0 +1,21 @@
+import {getService, juggler} from '@loopback/service-proxy';
```
raymondfeng: The `juggler` import is not used in the code.

packages/cli/lib/utils.js
```diff
@@ -386,8 +388,96 @@ exports.getRepositoryFileName = function(repositoryName) {
...
+exports.getDataSourceConnectorName = function(datasourcesDir, dataSourceClass) {
+  if (!dataSourceClass) {
+    return false;
+  }
+  let result;
+  let jsonFileContent;
+
+  let datasourceJSONFile = path.join(
+    datasourcesDir,
+    dataSourceClass.replace('Datasource', '.datasource.json').toLowerCase(),
```
raymondfeng: Some name mangling is needed to be consistent with datasource.json.

raymondfeng:
Since you are here, the generated `datasource.ts` file has formatting issues:
```
constructor(
    @inject('datasources.config.map', {optional: true})
    dsConfig: AnyObject = config // missing a trialing ,
  )
```

packages/cli/lib/utils.js
marioestradarosa: I have created a helper function to return the file name correctly and will serve to these two functions for now in utils.js . Thank you for finding that bug @raymondfeng

marioestradarosa:
> I found a bug which can be reproduced as follows:
Use lb4 datasource to create REST ds named MapDS. It creates src/datasources/map-ds.datasource.json
Run lb4 service and it crashes:

The problem is with getArtifactList , in that it only read files not its content, so it can't tell with 100%  accuracy the model class name being exported in a specific file, rather, it attempts to conver the file name (ie: `map-ds.datasource`) to the class name by using the function `pascalCase`, so the result is MapDs  and not MapDS. The source of truth is either the class name being exported in that file or the name property in its .json file.  I think all these artifacts need to be revised and improved. In the mean time I will fix this BUG by going to the name property.

marioestradarosa:
@raymondfeng  PTAL, I addressed the problem with the Datasource in the last commit. The problem covered also the repository, I am now taking the name property directly and placed that in a function  in utils. refactor repository as well.  Also, placed the missing , in the datasource.ts file

packages/cli/generators/repository/templates/src/repositories/repository-crud-default-template.ts.ejs
```diff
@@ -1,14 +1,14 @@
 import {<%= repositoryTypeClass %>, juggler} from '@loopback/repository';
 import {<%= modelName %>} from '../models';
-import {<%= dataSourceClassName %>} from '../datasources';
+import {<%= dataSourceName %>DataSource} from '../datasources';
```
raymondfeng: I don't think it's a good idea to concatenate strings in the EJS template. We should calculate the `dataSourceClassName` in the params passed to the EJS.

If we want to have both `dataSourceName` and `dataSourceClassName`, that's perfectly fine.

marioestradarosa: Ok. I will create dataSourceClassName.

marioestradarosa:
@raymondfeng  PTAL

marioestradarosa:
Thanks @raymondfeng  👍

PR #1743, opened 2018-09-21
feat(cli): add update notifier to remind cli upgrade

raymondfeng:
Add [update-notifier](https://github.com/yeoman/update-notifier) to remind users of newer CLI versions.

## Checklist

### DIFF
--- a/packages/cli/bin/cli-main.js
+++ b/packages/cli/bin/cli-main.js
@@ -8,8 +8,9 @@
 
 const checkNodeVersion = require('@loopback/dist-util').checkNodeVersion;
 
+const pkg = require('../package.json');
 try {
-  const range = require('../package.json').engines.node;
+  const range = pkg.engines.node;
   checkNodeVersion(range);
 } catch (e) {
   console.error(e.message);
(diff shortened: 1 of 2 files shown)


packages/cli/bin/cli-main.js
```diff
@@ -26,4 +27,13 @@ const opts = minimist(process.argv.slice(2), {
     commands: 'l', // --commands or -l: print commands
   },
 });
+
+const updateNotifier = require('update-notifier');
+// Force version check with `lb4 --version`
+const interval = opts.version ? 0 : undefined;
```
b-admike: curious: what is the behaviour of `updateNotifier` use when this value is `0` or `undefined`? I'm thinking that if `lb4 --version` is run, then it checks right away and if not, then it won't at all.

raymondfeng: Running `lb4 --version` forces the check immediately with `0`. If the interval is not number, it's default to one day.

Please note `update-notifier` uses a background process and it only prompts upgrade for next run.

PR #1746, opened 2018-09-23
feat(cli): add lb4 model option to select base model class

### LINKED ISSUE #1698: Cannot choose a base model class in `lb4 model`
When creating a new model, it is desirable to choose the base class from which the new model should inherit.

Possible use cases:

- inherit from `Entity` - this is typical for models persisted in a CRUD database
- inherit from `Model` - this is typical for models persisted in a KeyValue storage like Redis, but also for models describing data schemas used by nested (embedded) properties of other models (e.g. an `Address` model).
- inherit from a model defined in the application - this can be useful to design object hierarchy of models, e.g. a `Customer` is a `User`, but also an `Administrator` is a `User`.

/cc @raymondfeng if there is already an existing issue keeping track of this work, then feel free to close this one as a duplicate.

## Acceptance criteria

 - [ ] MUST HAVE: `lb4 model` asks the user whether the new model class should inherit from `Entity` or `Model`.
 - [ ] MUST HAVE: `--base` option to specify the base model via CLI arguments
 - [ ] MUST HAVE: update `lb4 repository` (added by https://github.com/strongloop/loopback-next/pull/1659) to offer only models inheriting from a base model compatible with the selected datasource (CRUD requires Entity, KeyValue can work with both Entity and Model).
 - [ ] SHOULD HAVE: `lb4 model` should offer all application model classes in addition to built-in Model and Entity classes in the prompt.
- [ ] Extract the functions like listing available models to utils.

marioestradarosa:
This PR adds the option to select a base model class to `lb4 model` **CLI** command.

**Done** 
- User can specify --base command line option with either Entity , Model or any valid model  custom class name. The command will validate if any custom class provided exists in `src/models`
- User is presented a list of options including Entity, Model and any valid model already created in `src/models` directory
- The **CLI** command will generate the file from the appropriate model template.
- Tests with invalid model name and valid model names were added
- Add documentation about the new option
- Thanks to @raymondfeng  suggestion the separator (base classes and custom one) was added in the model list.

**Excluded**
- `lb4 repository` update to filter models based on the baseModel  type 
   - PersistedModel works with Entity 
   - KeyValueModel works with Entity and Model
   - The reason is that this addition can make this PR more complex to digest for review
      - We have to navigate from child to father models to find out if it is Entity or Model for filtering, whenver an extended model from a custom model is selected.

close #1698

## Checklist

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -7,9 +7,15 @@
 
 const ArtifactGenerator = require('../../lib/artifact-generator');
 const debug = require('../../lib/debug')('model-generator');
+const inspect = require('util').inspect;
 const utils = require('../../lib/utils');
 const chalk = require('chalk');
 const path = require('path');
+
+const PROMPT_BASE_MODEL_CLASS = 'Please select the model base class';
+const ERROR_NO_MODELS_FOUND = 'Model was not found in';
+const BASE_MODELS = ['Entity', 'Model'];
+const MODEL_TEMPLATE_PATH = 'model.ts.ejs';
 
 /**
  * Model Generator
@@ -54,6 +60,17 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
     this.artifactInfo.properties = {};
     this.propCounter = 0;
 
+    this.artifactInfo.modelDir = path.resolve(
+      this.artifactInfo.rootDir,
+      utils.modelsDir,
+    );
+
+    this.option('base', {
+      type: String,
+      required: false,
+      description: 'A valid based model',
+    });
+
     return super._setupGenerator();
   }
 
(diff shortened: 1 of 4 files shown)


packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -1,7 +1,7 @@
 import {Entity, model, property} from '@loopback/repository';
```
bajtos: I think `Entity` needs to be changed to `<%= modelBaseClass %>`.

packages/cli/generators/model/templates/model-extended.ts.ejs
```diff
@@ -0,0 +1,20 @@
+import {model, property} from '@loopback/repository';
+import {<%= modelBaseClass %>} from './';
+
+@model()
+export class <%= className %> extends <%= modelBaseClass %> {
+<% Object.entries(properties).forEach(([key, val]) => { -%>
+  @property({
+  <%_ Object.entries(val).forEach(([propKey, propVal]) => { -%>
+    <%_ if (!['tsType'].includes(propKey)) { -%>
```
bajtos: I am not very happy that we have this complex logic duplicated in two templates. Can we us a simple `if` in the original template to decide what to import?

```ejs
<% if(isModelBaseBuiltin) -%>
import {<%= modelBaseClass %>, model, property} from '@loopback/repository';
<% } else { -%>
import {model, property} from '@loopback/repository';
<% } -%>
```

(Note that `require('./')` can be shortened to `require('.')` in recent Node.js versions.)

packages/cli/generators/model/templates/model.ts.ejs
marioestradarosa: Will these be valid?
`import {Entity, property} from '@loopback/repository';`
and
`import {Model, property} from '@loopback/repository';`

Wen use the `<%= modelBaseClass %>` ?

packages/cli/generators/model/templates/model-extended.ts.ejs
marioestradarosa: Ok. The idea was to separate them at this stage in case in the future both receive different complex logic. But at this point it makes sense to place the if as you indicated 👍

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -1,7 +1,12 @@
-import {Entity, model, property} from '@loopback/repository';
+<% if (isModelBaseBuiltin) { -%>
+import {<%= modelBaseClass %>, model, property} from '@loopback/repository';
+<% } else { -%>
+import {model, property} from '@loopback/repository';
+import {<%= modelBaseClass %>} from '.';
```
raymondfeng: nitpick: Would it be better to import from the base class file directly?

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -1,7 +1,12 @@
-import {Entity, model, property} from '@loopback/repository';
+<% if (isModelBaseBuiltin) { -%>
+import {<%= modelBaseClass %>, model, property} from '@loopback/repository';
```
raymondfeng: We should skip `property` import if no properties are defined for the model to avoid tslint errors.

marioestradarosa: What if the developer creates an empty model and then he starts adding the properties manually in the generated class?

packages/cli/generators/model/templates/model.ts.ejs
marioestradarosa: Isn't it the reason for the index.ts file in src/models?

packages/cli/generators/model/index.js
```diff
@@ -69,14 +86,82 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+        );
+      }
+    }
+
+    return this.prompt([
+      {
+        type: 'list',
+        name: 'modelBaseClass',
+        message: PROMPT_BASE_MODEL_CLASS,
+        choices: availableModelBaseClasses,
```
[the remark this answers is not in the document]
marioestradarosa: 👍 for the separator, I will include it.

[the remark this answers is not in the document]
marioestradarosa: @raymondfeng  PTAL the separator as suggested.

marioestradarosa:
@bajtos PTAL

packages/cli/generators/model/index.js
```diff
@@ -200,7 +288,11 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       this.artifactInfo.outFile,
     );
 
-    const modelTemplatePath = this.templatePath('model.ts.ejs');
+    this.artifactInfo.isModelBaseBuiltin = BASE_MODELS.includes(
+      this.artifactInfo.modelBaseClass,
+    )
+      ? true
+      : false;
```
bajtos: Why are these true/false values needed? Doesn't `include` already return a boolean? If it does not, then you case add `!!` before the call to convert a value to a boolean.

```ts
!!BASE_MODELS.include(/*...*/)
```

packages/cli/generators/model/templates/model.ts.ejs
bajtos: There actually is a good reason for importing from the model file instead of the index file.

Think about cyclic dependencies. For example, if `Customer` is extending `User` model, and the index file has re-exports sorted alphabetically (`Customer` first, `User` second), then the line in `Customer.ts` calling `require('.').User` will not be accessing an export that was not defined yet.

Having said that, I think this is not very likely because we are not sorting index files now, thus I am ok to fix this later (or never).

marioestradarosa:
thank you @bajtos !

packages/cli/generators/model/index.js
marioestradarosa: you were right, I simplified it. 👍

packages/cli/generators/model/templates/model.ts.ejs
marioestradarosa: Ok. I will leave it out of the scope of this PR, but will work on it.

PR #1750, opened 2018-09-24
fix(rest): coerce string parameters (reject object values)

bajtos:
Add a missing coercion step for string parameters and prevent the situation when a string value provided by the HTTP client and converted into an object or an array by query-string parser is passed as-is to the controller method, violating the endpoint contract and TypeScript declaration.

I discovered this problem while working on #1679. Without this change, endpoints like `TodoListController#count` are incorrectly accepting object values for their `where` parameter described as `type: string`. 

https://github.com/strongloop/loopback-next/blob/d65a17fab7ce44b7977e9c03388f5684558475f2/examples/todo-list/src/controllers/todo-list.controller.ts#L37-L39

## Checklist

### DIFF
--- a/packages/rest/test/unit/coercion/paramStringToString.unit.ts
+++ b/packages/rest/test/unit/coercion/paramStringToString.unit.ts
@@ -1,0 +1,87 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {ParameterObject} from '@loopback/openapi-v3-types';
+import {RestHttpErrors} from '../../..';
+import {test} from './utils';
+
+const OPTIONAL_STRING_PARAM: ParameterObject = {
+  in: 'path',
+  name: 'aparameter',
+  schema: {type: 'string'},
+};
+
+const REQUIRED_STRING_PARAM = {
+  ...OPTIONAL_STRING_PARAM,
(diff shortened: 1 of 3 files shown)


packages/rest/test/unit/coercion/paramStringToString.unit.ts
```diff
@@ -0,0 +1,88 @@
...
+    test(
+      REQUIRED_STRING_PARAM,
+      '',
+      RestHttpErrors.missingRequired(REQUIRED_STRING_PARAM.name),
+    );
+  });
+});
+
+describe('coerce param from string to string - optional', () => {
+  context('valid values', async () => {
```
jannyHou: nitpick: no need to add `async` I think

packages/rest/test/unit/coercion/paramStringToString.unit.ts
```diff
@@ -0,0 +1,88 @@
...
+    test(OPTIONAL_STRING_PARAM, '1.234e+30', '1.234e+30');
+    test(OPTIONAL_STRING_PARAM, '-1.234e+30', '-1.234e+30');
+  });
+
+  context('empty collection converts to undefined', () => {
+    // [], {} sent from request are converted to raw value undefined
+    test(OPTIONAL_STRING_PARAM, undefined, undefined);
+  });
+
+  // @jannyhou For review: shall we convert empty value to 0 or throw error?
```
jannyHou: 🙈 the comment can be removed. thanks!

PR #1618, opened 2018-08-17
feat(repository): introduce resolver and hasMany key inferrence with belongsTo decorator AND add belongsTo repository

shimks:
### **1st and 2nd commit: Type resolver and `keyTo` inferrence**

There is an issue with relation and decorators where model resolution doesn't occur properly. A short example to illustrate the problem:
```ts
class Customer {
id: number;
@hasMany(Order) orders: Order[];
}
// in another file
class Order {
@belongsTo(Customer) customerId: number;
}
```
When Customer is being defined, Order will be read and be attempted to be defined. As Order is being defined, it comes across Customer which has yet to be defined, so the metadata being stored by the decorators is incomplete at the time it's being stored.

The work around for this is to introduce a resolver; a resolver is a function that would return the given model when called. So instead of passing in the Models themselves these decorators, you would pass in their resolver instead: `@hasMany(() => Order)` and `@belongsTo(() => Customer)`.

Another feature included in these commits is automatic building of hasMany metadata based on existing metadata on belongsTo. More specifically, `keyTo` is inferred from the name of the property key in which belongsTo metadata exists.

In summary, this PR introduces:
- logic to detect circular dependency (only with `@hasMany` and `@property.array`)
- `keyTo` inference from the key decorated by `@belongsTo`

### **3rd commit: belongsTo repository**

There is a problem where if belongsTo and hasMany relations are defined on the repositories, our context system will throw an error due to the cyclic dependence the relations create. To illustrate the example in a dirty dirty code snippet:
```js
// ... customer repo constructor
constructor(@repository(OrderRepository) orderRepo)
// ... order repo constructor
constructor(@repository(CustomerRepository) customerRepo)
```
Take an event where customer repository needs to be instantiated. An injection on `orderRepo` will attempted to be made and an order repository will be instantiated. When an order repository is being instantiated, a customer repository needs to be injected, and therefore a circular dependency is established.

The solution to solving the dependency is to use a resolver like we did for the commits before. Take a look at [this code snippet](https://github.com/strongloop/loopback-next/blob/3acb81d62709613eff424b749f9b75115c751913/packages/repository/test/fixtures/repositories/customer.repository.ts) and [this](https://github.com/strongloop/loopback-next/blob/3acb81d62709613eff424b749f9b75115c751913/packages/repository/test/fixtures/repositories/order.repository.ts) to see how the UX would look like. I'll probably work on introducing a decorator that won't promisify the repositories here.

related to #1361

### Things that have not been done:
[...]

PR #1751, opened 2018-09-24
feat: type resolver for property decorators

bajtos:
Implement a concept of deferred type resolution, where the decorator metadata provides a TypeResolver function to obtain the actual type to use. For example:

```ts
@property(() => Product
```

Deferred resolution is a solution for avoiding require() loops that are frequently encountered when defining relations like hasMany + belongsTo.

I have extracted this pull request from #1618, please refer to that pull request to better understand the context of this change and previous discussions.

In #1618, I was advocating against using `Type | TypeResolver` union. As I was thinking more about other approaches, the impacts on developer experience (LB4 app authors) and how little time we have until 4.0 GA, I decided to continue down the road started by @shimks.

## Checklist


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -78,22 +77,6 @@ export function stringTypeToWrapper(type: string | Function): Function {
   return wrapper;
 }
 
-/**
- * Determines whether the given constructor is a custom type or not
- * @param ctor Constructor
- */
-export function isComplexType(ctor: Function) {
```
bajtos: Moved to `@loopback/repository` as `isBuiltinType`.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -122,19 +105,20 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
...
     });
-  } else {
+  } else if (isBuiltinType(resolvedType)) {
     Object.assign(propDef, {
-      type: <JSONSchemaTypeName>propertyType.name.toLowerCase(),
+      type: resolvedType.name.toLowerCase(),
     });
+  } else {
+    Object.assign(propDef, {$ref: `#/definitions/${resolvedType.name}`});
   }
```
bajtos: I re-organized the branches in a different order, that's why the diff looks bigger than the changes actually are.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -176,34 +160,40 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
...
-        if (propSchema.definitions) {
-          for (const key in propSchema.definitions) {
-            result.definitions[key] = propSchema.definitions[key];
-          }
-          delete propSchema.definitions;
-        }
+        resolveType(metaProperty.itemType as string | Function)
+      : resolvedType;
 
-        result.definitions[referenceType.name] = propSchema;
```
bajtos: I added a reverse check + `continue`, this part is best reviewed with white-space changes ignored. (See "Diff settings" at the top of the page.)

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -176,34 +160,40 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
     // populating "properties" key
     result.properties[p] = metaToJsonProperty(metaProperty);
 
+    // handling 'required' metadata
```
bajtos: Moved here from the bottom of the loop body to allow me to use early `continue`.

packages/repository/test/unit/type-resolver.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('returns false when the arg is Number type', () => {
+    expect(isTypeResolver(Number)).to.be.false();
+  });
+
+  it('returns false when the arg is Boolean type', () => {
+    expect(isTypeResolver(Boolean)).to.be.false();
+  });
+
+  it('returns false when the arg is Object type', () => {
```
b-admike: Nitpick: since the logic in these tests is the same, can you create a helper function which takes in the parameter to the `isTypeResolver` function and creates a test case with the type name as well as the expected result?

packages/repository/test/unit/type-resolver.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('supports Date type', () => {
+    const ctor = resolveType(Date);
+    expect(ctor).to.eql(Date);
+
+    const inst = new ctor(A_DATE_STRING);
+    expect(inst.toISOString()).to.equal(A_DATE_STRING);
+  });
+
+  it('supports Date provider', () => {
```
b-admike: Nitpick: would `supports Date resolver` be better than `provider`?

packages/repository/test/unit/type-resolver.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+    expect(inst.toISOString()).to.equal(A_DATE_STRING);
+  });
+});
+
+describe('isBuiltinType', () => {
+  it('returns true for Number', () => {
+    expect(isBuiltinType(Number)).to.eql(true);
+  });
+
+  it('returns true for String', () => {
```
b-admike: Nitpick: ditto re organizing the test cases in a clear way, while utilizing a helper function like I mentioned above. With that said, I'm still fine with these explicit tests.

packages/repository/src/type-resolver.ts
```diff
@@ -0,0 +1,72 @@
...
+  // 3. Built-in types like Date & Array are not type providers
+  if (isBuiltinType(fn)) return false;
+
+  // TODO(bajtos): support model classes defined via ES5 constructor function
+
+  return true;
+}
+
+/**
+ * Check if the provided function is a built-in type: a constructor-like
```
b-admike: Nitpick: can we say `built-in type or a constructor-like...` instead of using a colon?

packages/repository/test/unit/type-resolver.unit.ts
bajtos: Creating parameterized test cases is tricky. If done wrong, then test failures don't point to the source code of the individual test case, only to the same helper function. In coercion tests, we found a solution where we intercept errors and append extra stack trace to them in order to preserve location of where the test case was defined.

In this particular case, I think such extra complexity outweighs benefits.

packages/repository/test/unit/type-resolver.unit.ts
bajtos: Good catch! Will fix.

Issue #788, opened 2017-12-11
Response code for POST method should be 201 not 200


jannyHou:
Response code for POST method should be 201 not 200

An old issue from LoopBack3: https://github.com/strongloop/loopback/issues/360

We should fix it in LoopBack4 as well.

### Notes from the spike

- Users should be able to customize the status code from the response object returned by the controller method, otherwise we should default to 201 - this will be implemented by #436.

- There are two approaches we can choose as implementation, but based on the spike findings and the consensus, pursue option 1 from below:
  - option 1: return the status code and header from the result of controller method directly
  - ~option 2: read the status code from OpenAPI spec and write it into the response in `rest/src/writer.ts`~, see the spike PR https://github.com/strongloop/loopback-next/pull/1260
  - There is already some comparison and discussion for the two approaches in PR https://github.com/strongloop/loopback-next/pull/1260 to start with, chose one approach and implement it. 

### Acceptance criteria

- [ ] Modify controllers in the `example/*` projects so that `create*` methods (typically mapped to `POST` verb) return `201 Created` instead of `200.
  - Update the OpenAPI Spec description of operation result, replace 200 to 201
  - Update the controller method implementation to return status code 201 and set `Location` header to the URL where the newly created model instance can be found
- [ ] Update CRUD Controller template used by CLI similarly
- [ ] If https://github.com/strongloop/loopback-next/issues/2736 has been done, then update the extension accordingly too.

Cooker-Monster:
Thanks to reactivate this important point !

bajtos:
I think this has two parts:

 - We need to figure out how to allow controllers to specify response status code (and response headers, while we are at this)

 - The CRUD controller template (see https://github.com/strongloop/loopback-next/issues/727) and the CRUD controller base class (see https://github.com/strongloop/loopback-next/pull/740) should return 201 from `create` operation

jannyHou:
hi team, we are upgrading from swagger2 to openapi3, PR is ready for it https://github.com/strongloop/loopback-next/pull/916
There is an interface change of the `responseObject` in openapi3, I think it's better to implement this story for v3 directly instead of implement it with v2 then take another effort to upgrade to v3, thanks!

kjdelisle:
[...]

Issue #436, opened 2017-07-13
Configurable response types/formats


bajtos:
At the moment, the return value of a controller method/router handler is converted to the response using the following hard-coded algorithm:

 - Object types (including arrays, but also Date, RegExp, Buffer) are converted to JSON and sent with "application/json" content-type.
 - String results are sent as-is with "text/plain" content-type.
 - Anything else is passed to `response.write()` as-is and no content-type is set.

We need a more flexible setup, we should at least honor "produces" configuration from the API specification.

## Acceptance Criteria

1. A controller method should be able to specify `media type` for the `responses` object with `@operation`. See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#response-object

    _Note by @bajtos: I believe this is already supported, see e.g. the TodoController in our examples: https://github.com/strongloop/loopback-next/blob/cb308add60bbe938c8e85812676eb817e9e0efc9/examples/todo/src/controllers/todo.controller.ts#L32_

2. A controller method should be able to return JS representation of the response body or a wrapper corresponding to the response objects. The goal: allow controller methods to control response status code and headers. E.g. `Location` for `201 Created` responses to POST requests creating new model instances, `Content-Type` and `Content-Disposition` for file downloads.

3. LoopBack should be able to match the response media types to the `Accept` request header to determine what content type to be produced for a given request. It will set `Content-Type` response header.

4. Have built in support for `application/json` and `text/plain` media-types.

**EDIT: Move to another task** 
>4. LoopBack runtime should define an extension point to register http response serializers that can be used to serialize the JS plain/wrapper object into a http response. A serializer should be able to handle certain media types, for example, json, xml, or protocol buffer. The serializer just has to deal with the http response object as a canonical representation.

raymondfeng:
We probably should start to refactor https://github.com/strongloop/loopback-next/tree/master/packages/repository/src/types into its own package to form the LB.next typing system and leverage it for parameter parsing/conversion/serialization.

bajtos:
Labelling as non-MVP, because we have a simple workaround available - users can use their own implementation of `send` sequence action.

dustinbarnes:
@bajtos can you give an example of that? I'm currently fighting the content-type header (it really needs to send application/json when we return a single string), and using the sugar for @get(path, spec), but it seems to ignore whatever is put into the "produces" section of the spec object.

bajtos:
[...]

PR #1753, opened 2018-09-24
[WIP] feat(rest): allow bypassing http response writing and custom content type

raymondfeng:
We discover an issue to return html from a controller:
See https://github.com/strongloop/loopback4-example-shopping/pull/16

## Checklist

### DIFF
--- a/docs/site/Controllers.md
+++ b/docs/site/Controllers.md
@@ -214,6 +214,55 @@ export class HelloController {
 - `@param.query.number` specifies in the spec being generated that the route
   takes a parameter via query which will be a number.
 
+## Custom Response Writing
+
+By default, LoopBack serializes the return value from the controller methods
+into HTTP response based on the data type. For example, a string return value
+will be written to HTTP response as follows:
+
+status code: `200` Content-Type header: `text/plain` body: `the string value`
+
+In some cases, it's desirable for a controller method to customize how to
+produce the HTTP response. For example, we want to set the content type as
+`text/html`.
+
+```ts
+import {get} from '@loopback/openapi-v3';
+import {inject} from '@loopback/context';
+import {RestBindings, Response} from '@loopback/rest';
+
+export class HomePageController {
+  // ...
+  constructor(@inject(RestBindings.Http.RESPONSE) private res: Response) {
+    // ...
+  }
+  @get('/')
+  homePage() {
+    const homePage = '<html><body><h1>Hello</h1></body></html>';
+    this.res
+      .status(200)
+      .contentType('html')
+      .send(homePage);
+    // Return the `response` object itself to bypass further writing to HTTP
+    return this.res;
+  }
+}
+```
+
+Alternatively, the controller method can set status code and content type but
+leave the body to be populated by the LoopBack HTTP response writer.
+
+```ts
+@get('/')
+  homePage() {
+    const homePage = '<html><body><h1>Hello</h1></body></html>';
+    this.res
+      .status(200)
+      .contentType('html');
+    return homePage;
+  }
+```
+
 ## Handling Errors in Controllers
 
 In order to specify errors for controller methods to throw, the class
--- a/packages/rest/src/types.ts
+++ b/packages/rest/src/types.ts
@@ -51,8 +51,13 @@ export type InvokeMethod = (
  *
  * @param response The response the response to send to.
  * @param result The operation result to send.
+ * @param contentType The content type to send
  */
-export type Send = (response: Response, result: OperationRetval) => void;
+export type Send = (
+  response: Response,
+  result: OperationRetval,
+  contentType?: string,
+) => void;
 
 /**
  * Reject the request with an error.
--- a/packages/rest/src/writer.ts
+++ b/packages/rest/src/writer.ts
@@ -12,21 +12,37 @@ import {Readable} from 'stream';
  *
  * @param response HTTP Response
  * @param result Result from the API to write into HTTP Response
+ * @param contentType Optional content type for the response
  */
 export function writeResultToResponse(
   // not needed and responsibility should be in the sequence.send
   response: Response,
   // result returned back from invoking controller method
   result: OperationRetval,
+  // content type for the response
+  contentType?: string,
 ): void {
-  if (!result) {
-    response.statusCode = 204;
+  // Check if response writing should be bypassed. When the return value
+  // is the response object itself, no writing should happen.
+  if (result === response) return;
+
+  if (result === undefined) {
+    response.status(204);
     response.end();
     return;
   }
 
-  if (result instanceof Readable || typeof result.pipe === 'function') {
-    response.setHeader('Content-Type', 'application/octet-stream');
+  function setContentType(defaultType: string = 'application/json') {
+    if (response.getHeader('Content-Type') == null) {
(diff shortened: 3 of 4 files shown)


raymondfeng:
@virkt25 I refined the PR to check if a controller method returns the `response` object itself as the flag to skip writing. PTAL.

docs/site/Controllers.md
```diff
@@ -214,6 +214,41 @@ export class HelloController {
 - `@param.query.number` specifies in the spec being generated that the route
   takes a parameter via query which will be a number.
 
+## Custom Response Writing
+
+By default, LoopBack serialize the return value from the controller methods into
```
virkt25: LoopBack will serialize
**OR**
LoopBack serializes the ...

packages/rest/src/writer.ts
```diff
@@ -12,21 +12,37 @@ import {Readable} from 'stream';
...
+
+  if (result === undefined) {
+    response.status(204);
     response.end();
     return;
   }
 
-  if (result instanceof Readable || typeof result.pipe === 'function') {
-    response.setHeader('Content-Type', 'application/octet-stream');
+  function setContentType(defaultType: string = 'application/json') {
```
virkt25: the variable name should be changed to `type` as it can be overridden.

bajtos:
I am rather unhappy about the design changes proposed in this pull request.

So far, the framework was offering pretty strong guarantees to LoopBack users: every HTTP response was produced either by `send` or `reject`. If an application or an extension wants to intercept or modify _all_ responses, and it provides custom `send` & `reject` actions, then it can be pretty certain it has all cases covered.

The solution proposed by this pull request, where the controller sends the response directly via Express `response` API, is throwing all guarantees out.

I understand we need a quick workaround to enable our shopping-cart example to serve a home page (https://github.com/strongloop/loopback4-example-shopping/pull/16) and I am fine to add an undocumented fix as a short-term workaround.

I would personally use the following check to determine whether the response have been already produced:

 - The controller method returned `undefined` AND
 - The response headers have been already sent (see Node.js API [response.headersSent](https://nodejs.org/api/http.html#http_response_headerssent))

What's wrong with letting the controller method return `response`? I don't like the fact that a controller method can return the response and yet don't send any data back. We will end up with a client connection hanging forever in such case.

For a longer term solution that will be documented, I'd like us to use the approach described in https://github.com/strongloop/loopback-next/issues/436, https://github.com/strongloop/loopback-next/issues/436#issuecomment-388879072 and https://github.com/strongloop/loopback-next/issues/788#issuecomment-381639062, where the controller method can return an object describing the response to be produced, e.g. `{statusCode: 201, headers: {/*...*/}, body: data}`.

To move this pull request forward, I am asking for the following changes:
 - Reduce the amount of functionality changes. Just add a check to `send` action to return immediately when the response have been already send by the controller.
 - Either keep this feature undocumented, or else make it very clear that this feature is a short-term hack.

packages/rest/src/types.ts
```diff
@@ -51,8 +51,13 @@ export type InvokeMethod = (
  *
  * @param response The response the response to send to.
  * @param result The operation result to send.
+ * @param contentType The content type to send
  */
-export type Send = (response: Response, result: OperationRetval) => void;
+export type Send = (
+  response: Response,
+  result: OperationRetval,
+  contentType?: string,
```
bajtos: I don't understand the purpose of this new parameter. I don't see it used anywhere in production code, only in unit tests.

`send` is typically invoked from a Sequence. There is usually the same sequence of actions implemented for all endpoints. IMO, the decision which content-type to send back should be made in individual controller methods (or routes), not in the Sequence.

I am proposing to revert this change.

packages/rest/src/writer.ts
```diff
@@ -12,21 +12,37 @@ import {Readable} from 'stream';
...
  * @param result Result from the API to write into HTTP Response
+ * @param contentType Optional content type for the response
  */
 export function writeResultToResponse(
   // not needed and responsibility should be in the sequence.send
   response: Response,
   // result returned back from invoking controller method
   result: OperationRetval,
+  // content type for the response
+  contentType?: string,
```
bajtos: Similarly here, let's revert this change please.

packages/rest/src/writer.ts
```diff
@@ -12,21 +12,37 @@ import {Readable} from 'stream';
...
+  if (result === undefined) {
+    response.status(204);
     response.end();
     return;
   }
 
-  if (result instanceof Readable || typeof result.pipe === 'function') {
-    response.setHeader('Content-Type', 'application/octet-stream');
+  function setContentType(defaultType: string = 'application/json') {
+    if (response.getHeader('Content-Type') == null) {
```
bajtos: Personally, I prefer to revert this change.

Ideally, there should be only two ways how a content-type is determined:

1. The controller method/route returns data to be serialized to the response. The content-type is determined by the framework depending on the type of data (object vs. string vs. stream or buffer), request headers like `Accepts` and endpoint OpenAPI spec. Right now, we have a very limited version of this algorithm implemented in the current `writeResultToResponse` code (objects are sent as JSON, streams as binary, strings as text).

2. The controller wants to set an explicit content type, status code or change any other aspect of the HTTP response. It returns an object describing response properties (status code, headers, body) and the `send` action applies these instructions to the `response` object. This will be enabled by #436 in the future.

There is of course also the fourth mode being added here, I think it should be the only scope of this pull request,  and I consider it as a short-term workaround:

4. The controller (or a sequence action in general) takes full control of the response sent: it sets the status code & headers and writes the response body. The `send` action becomes a no-op.

packages/rest/src/writer.ts
```diff
@@ -12,21 +12,37 @@ import {Readable} from 'stream';
...
+      response.setHeader('Content-Type', contentType || defaultType);
+    }
+  }
+
+  if (
+    result instanceof Readable ||
+    typeof (result && result.pipe) === 'function'
```
bajtos: https://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

```js
const isStream = result instanceof Readable || 
  typeof (result && result.pipe) === 'function';

if (isStream) {
  // ...
}
```

raymondfeng:
@bajtos Thank you for the feedback. A few clarifications:

1. I introduced the extra `contentType` argument in the hope of being able to infer the value from the response object. But that seems to be a bit involved to implement at the moment. As you mentioned, the final content type can be a result of content negotiation.

2. I was thinking about to introduce a new `HttpResponse` return type with the fluent API so that a controller method can return a fully-populated response so that our writer can honor it. But I realized most of the APIs are already available on `Response` and I didn't want to reinvent the wheel.

3. There is a `pull` and `push` to write the http response. 
- pull: The writer pulls data from the controller method return value and serialize them into the response
- push: The controller method serializes data into the response

I see use cases for both modes and would love to allow both flavors. We can treat the `push` mode as the controller method has its own writer.

The issues above need to be addressed in the long run. To enable our shopping example, I'll create a different (much simpler) PR to only override the content type if it's not set.

virkt25:
Isn't it better to allow the controller to just set the response type / response header? ... How much work is it for us to implement pulling the controller metadata to see what the response header should be and set it based on that?

raymondfeng:
Close it to favor #1760

Issue #1754, opened 2018-09-24
GET /ping is missing content object in the OAS


Alan-Cha:
## Description

The a new LoopBack app automatically creates a `GET /ping` operation but the operation is missing a content object, aka [media type object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#mediaTypeObject).

## Step to reproduce

Create a new Loopback project, start the API, and find the OAS (go to [http://localhost:3000/openapi.json](http://localhost:3000/openapi.json)). 

```
lb4 app
cd [app name]
npm start
```

## Current behavior

```
"/ping": {
      "get": {
        "x-controller-name": "PingController",
        "x-operation-name": "ping",
        "tags": [
          "PingController"
        ],
        "responses": {
          "200": {
            "description": "Return value of PingController.ping"
          }
        }
      }
    }
  }
```

## Expected behavior

```
"/ping": {
  "get": {
    "x-controller-name": "PingController",
    "x-operation-name": "ping",
    "tags": [
      "PingController"
    ],
    "responses": {
      "200": {
        "description": "Return value of PingController.ping",
        "content": {
          "application/json": {
            "schema": {
              "type": "object",
              "properties": {
                "greeting": {
                  "type": "string"
                },
                "date": {
                  "type": "string"
                },
                "url": {
                  "type": "string"
                },
                "headers": {
                  "type": "object",
                  "properties": {
                    "cache-control": {
                      "type": "string"
                    },
                    "postman-token": {
                      "type": "string"
                    },
                    "user-agent": {
                      "type": "string"
                    },
                    "accept": {
                      "type": "string"
                    },
                    "host": {
                      "type": "string"
                    },
                    "accept-encoding": {
                      "type": "string"
                    },
                    "connection": {
                      "type": "string"
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

In theory you should just be able to copy and paste the content object into the template but I am unsure of the data types for `cache-control`, `postmanToken`, and `userAgent`.

raymondfeng:
See https://github.com/strongloop/loopback4-example-shopping/pull/16/commits/2258e9876dd4ebed608589348f0b3a17447bedab

Issue #301, opened 2017-05-25
Drop our openapi-spec packages, use openapi3-ts instead


bajtos:
https://github.com/metadevpro/openapi3-ts looks more feature complete than our own openapi packages. Should we use openapi3-ts instead?

If we want to support both Swagger v2 and OpenAPI v3 then we probably cannot drop our packages, unless there is a replacement available.

UPDATE:

Looks like we have already dropped most of our custom type definitions. As of today ([0411811aa](https://github.com/strongloop/loopback-next/tree/0411811aa8348e31636fb0fe07b44dd18469280f/packages/openapi-v3-types/src)), openapi-v3-spec-types provides only few helpers:
 - `OpenApiSpec` as an alias for `OpenAPIObject`
 - `createEmptyApiSpec()`
 - `isSchemaObject()` and `isReferenceObject()` (type guards)

## Acceptance criteria

- [x] contribute type guards to `openapi3-ts` - see https://github.com/metadevpro/openapi3-ts/pull/46 and https://github.com/strongloop/loopback-next/pull/2974
- ~~try to upstream `createEmptyApiSpec`, e.g. to [src/dsl/OpenApiBuilder.ts](https://github.com/metadevpro/openapi3-ts/blob/ab997f12a63fa215e3b0c08cc293429b97ce0a44/src/dsl/OpenApiBuilder.ts)~~
- [x] move the rest to our `openapi-v3` package ~~(`createEmptyApiSpec` should go to `openapi--spec-builder` package)~~ - see https://github.com/strongloop/loopback-next/pull/3220
- [ ] and finally remove `openapi-v3-types` - see https://github.com/strongloop/loopback-next/pull/3385

bajtos:
Related: #182.

dhmlau:
@jannyHou , with your openapiv3 work, is this ticket still valid?

bajtos:
From what I remember, our OpenAPI v3 packages are using a 3rd party dependency under the hood, but adding some extra features/types on top of that dependency. I would personally like more if we could contribute our improvements to that 3rd party package and start using it directly. IIRC, @jannyHou had good arguments against that, could you @jannyHou  please re-post them here please?

virkt25:
@jannyHou any updates for this?

jannyHou:
@bajtos @virkt25 Nice to know we already have an issue to continue the discussion. Copy paste from my V3 prototype PR:

---

From @bajtos 
[...]

PR #1755, opened 2018-09-24
feat(cli): add responses for PingController.ping()

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/1754

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
+++ b/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
@@ -1,5 +1,31 @@
-import {Request, RestBindings, get} from '@loopback/rest';
+import {Request, RestBindings, get, ResponseObject} from '@loopback/rest';
 import {inject} from '@loopback/context';
+
+/**
+ * OpenAPI response for ping()
+ */
+const PING_RESPONSE: ResponseObject = {
+  description: 'Ping Response',
+  content: {
+    'application/json': {
+      schema: {
+        type: 'object',
+        properties: {
+          greeting: {type: 'string'},
+          date: {type: 'string'},
+          url: {type: 'string'},
+          headers: {
+            type: 'object',
+            patternProperties: {
+              '^.*$': {type: 'string'},
+            },
+            additionalProperties: false,
+          },
+        },
+      },
+    },
+  },
+};
 
 /**
  * A simple controller to bounce back http requests
@@ -8,8 +34,13 @@ export class PingController {
   constructor(@inject(RestBindings.Http.REQUEST) private req: Request) {}
 
   // Map to `GET /ping`
-  @get('/ping')
+  @get('/ping', {
+    responses: {
+      '200': PING_RESPONSE,
+    },
+  })
   ping(): object {
+    // Reply with a greeting, the current time, the url, and request headers
     return {
       greeting: 'Hello from LoopBack',
       date: new Date(),
--- a/packages/cli/generators/app/templates/test/acceptance/ping.controller.acceptance.ts.ejs
+++ b/packages/cli/generators/app/templates/test/acceptance/ping.controller.acceptance.ts.ejs
@@ -2,6 +2,7 @@ import {
   Client,
   createRestAppClient,
   givenHttpServerConfig,
+  expect,
 } from '@loopback/testlab';
 import {<%= project.applicationName %>} from '../..';
 
@@ -25,7 +26,8 @@ describe('PingController', () => {
   });
 
   it('invokes GET /ping', async () => {
-    await client.get('/ping?msg=world').expect(200);
+    const res = await client.get('/ping?msg=world').expect(200);
+    expect(res.body).to.containEql({greeting: 'Hello from LoopBack'});
   });
 
   function givenAnApplication() {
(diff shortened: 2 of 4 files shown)


packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
```diff
@@ -1,14 +1,41 @@
 import {Request, RestBindings, get} from '@loopback/rest';
 import {inject} from '@loopback/context';
 
+const PING_RESPONSE = {
```
virkt25: Can we move this to be below the `controller` **and** add a comment above it stating that it's the response object schema. As is, it's very overwhelming to look at.

raymondfeng: I don't think the decorator can forward reference the constant.

bajtos: Please add an explicit type `SchemaObject` to let the compiler verify correctness of the schema here, where we are defining `PING_RESPONSE`, instead of deferring the check until `PING_RESPONSE` is assigned to `responses` in `@get` below.

```ts
const PING_RESPONSE: SchemaObject = {
  // ...
};
```

packages/cli/generators/app/templates/test/acceptance/ping.controller.acceptance.ts.ejs
```diff
@@ -25,7 +26,47 @@ describe('PingController', () => {
   });
 
   it('invokes GET /ping', async () => {
-    await client.get('/ping?msg=world').expect(200);
+    const res = await client.get('/ping?msg=world').expect(200);
+    expect(res.body).to.containEql({greeting: 'Hello from LoopBack'});
+  });
+
+  it('exposes OpenAPI spec at /openapi.json', async () => {
```
bajtos: The test name feels misleading to me. We are not testing the entire OpenAPI spec, only the spec of `GET /ping` endpoint here. How about:

```ts
it('provides a valid OpenAPI spec for "GET /ping"', async () => {
 // ...
});
```

Personally, I don't think we have to make an HTTP request to test that, I would move this test to `src/test/integration` and obtain the spec via `app.restServer.getApiSpec()`. Such test will be much more resilient against possible changes of `/openapi.json` location in the future.

I see the value in a test obtaining the spec via HTTP. I think such test should go to `test/acceptance/application.acceptance.ts` and verify that the returned spec is valid.

```ts
import {validateApiSpec} from '@loopback/testlab';

describe('TodoApplication' /* use the real app class name */, () => {
  it('exposes valid OpenAPI spec at /openapi.json', async () => {
    const res = await client.get('/openapi.json').expect(200);
    await validateApiSpec(res.body);
  });
});
```

I am ok with leaving such test out of scope of this pull request if you prefer.

raymondfeng:
@bajtos I added typing for PING_RESPONSE and removed the `/openapi.json` test from the PR. PTAL.

packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
```diff
@@ -1,5 +1,32 @@
 import {Request, RestBindings, get} from '@loopback/rest';
 import {inject} from '@loopback/context';
+import {ResponseObject} from '@loopback/openapi-v3-types';
```
bajtos: Can we import `ResponseObject` from `@loopback/rest` directly? I believe our rest package is re-exporting OAS types for ease of use.

We are going to remove `@loopback/openapi-v3-types` soon (before 4.0 GA), see https://github.com/strongloop/loopback-next/issues/301.

packages/cli/generators/project/templates/package.json.ejs
```diff
@@ -78,6 +78,7 @@
     "@loopback/core": "<%= project.dependencies['@loopback/core'] -%>",
     "@loopback/dist-util": "<%= project.dependencies['@loopback/dist-util'] -%>",
     "@loopback/openapi-v3": "<%= project.dependencies['@loopback/openapi-v3'] -%>",
+    "@loopback/openapi-v3-types": "<%= project.dependencies['@loopback/openapi-v3-types'] -%>",
```
bajtos: I prefer to import OAS types from `@loopback/rest`, see my comment above.

packages/cli/generators/project/templates/package.plain.json.ejs
```diff
@@ -72,6 +72,7 @@
 <% if (project.projectType === 'application') { -%>
     "@loopback/core": "<%= project.dependencies['@loopback/core'] -%>",
     "@loopback/openapi-v3": "<%= project.dependencies['@loopback/openapi-v3'] -%>",
+    "@loopback/openapi-v3-types": "<%= project.dependencies['@loopback/openapi-v3-types'] -%>",
```
bajtos: Ditto.

packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
raymondfeng: I was concerned about https://github.com/Microsoft/TypeScript/issues/26985. Will double-check.

raymondfeng: I add the code to re-export `@loopback-v3-types` from `@loopback-rest`.

Issue #1590, opened 2018-08-02
Cache pre-compiled AJV validators


bajtos:
The initial implementation of request-body validation uses the more convenient but significantly less-performant API from AJV:

https://github.com/strongloop/loopback-next/blob/34eba3498528fe70fa9e64e8582d30a5b255e35b/packages/rest/src/validation/request-body.validator.ts#L93-L103

As a result, a typical POST (create) request is 10x slower than a GET (find) request.

We should rework our validation to cache pre-compiled AJV validators, see [docs](https://github.com/epoberezkin/ajv#compileobject-schema---functionobject-data).

- To avoid slow application start, validators should be precompiled lazily (on the first request).
- Precompiled validators can be stored for example as a new property of the [`RouteEntry`](https://github.com/strongloop/loopback-next/blob/34eba3498528fe70fa9e64e8582d30a5b255e35b/packages/rest/src/router/routing-table.ts#L165-L202) class.

## Acceptance criteria

 - [ ] Modify request-body validator to cache pre-compiled validators
 - [ ] Re-run our [benchmark](https://github.com/strongloop/loopback-next/tree/master/benchmark) and update its [README](https://github.com/strongloop/loopback-next/blob/master/benchmark/README.md) with the newly observed data

PR #1757, opened 2018-09-25
perf: improve schema validation peformance

hacksparrow:
Improve schema validation performance. Fixes https://github.com/strongloop/loopback-next/issues/1590.

Was:

```
find all todos: { requestsPerSecond: 1737.57, latency: 5.23 }
create a new todo: { requestsPerSecond: 249.67, latency: 39.6 }
```

Is:

```
find all todos: { requestsPerSecond: 2244.7, latency: 3.89 }
create a new todo: { requestsPerSecond: 1209.37, latency: 7.74 }
```

## Checklist

### DIFF
--- a/packages/rest/src/validation/request-body.validator.ts
+++ b/packages/rest/src/validation/request-body.validator.ts
@@ -87,6 +87,9 @@ function convertToJsonSchema(openapiSchema: SchemaObject) {
  * @param schema The JSON schema used to perform the validation.
  * @param globalSchemas Schema references.
  */
+
+const compiledSchemaCache = new WeakMap();
+
 function validateValueAgainstJsonSchema(
   // tslint:disable-next-line:no-any
   body: any,
@@ -102,22 +105,23 @@ function validateValueAgainstJsonSchema(
     allErrors: true,
   });
 
-  try {
-    if (ajv.validate(schemaWithRef, body)) {
-      debug('Request body passed AJV validation.');
-      return;
-    }
-  } catch (err) {
-    debug('Fails to execute AJV validation:', err);
-    // TODO: [rfeng] Do we want to introduce a flag to disable validation
-    // or sink validation errors?
-    throw err;
+  if (!compiledSchemaCache.has(jsonSchema)) {
+    const compiled = ajv.compile(schemaWithRef);
+    compiledSchemaCache.set(jsonSchema, compiled);
   }
 
-  debug('Invalid request body: %s', util.inspect(ajv.errors));
+  const validate = compiledSchemaCache.get(jsonSchema);
+  if (validate(body)) {
+    debug('Request body passed AJV validation.');
+    return;
+  }
+
+  const validationErrors = validate.errors;
+
+  debug('Invalid request body: %s', util.inspect(validationErrors));
 
   const error = RestHttpErrors.invalidRequestBody();
-  error.details = _.map(ajv.errors, e => {
+  error.details = _.map(validationErrors, e => {
     return {
       path: e.dataPath,
       code: e.keyword,


packages/rest/src/validation/request-body.validator.ts
```diff
@@ -34,7 +34,6 @@ export function validateRequestBody(
   globalSchemas?: SchemasObject,
 ) {
   if (!requestBodySpec) return;
-
```
bajtos: Please preserve these empty lines, they serve as visual delimiter between different sections of this long(ish) method.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -51,7 +51,8 @@ export function validateRequestBody(
   if (!schema) return;
 
   const jsonSchema = convertToJsonSchema(schema);
-  validateValueAgainstJsonSchema(body, jsonSchema, globalSchemas);
+  const schemaTitle = schema.title || '';
```
hacksparrow: Can title ever be `undefined` in our case? Temporarily doing `schema.title || ''` to stop TS from complaining, this will be fixed later.

raymondfeng: See https://json-schema.org/understanding-json-schema/reference/generic.html#metadata. I think `title` is optional.

raymondfeng: We probably should generate uuid as the title to make it a unique key for caching.

hacksparrow:
@raymondfeng I found a workaround for the possibility that `title` could be undefined. Updated the PR with the new implementation.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -102,22 +109,22 @@ function validateValueAgainstJsonSchema(
...
   }
 
-  debug('Invalid request body: %s', util.inspect(ajv.errors));
+  const validate = validateSchema[schemaTitle];
+  if (validate(body)) {
+    debug('Request body passed AJV validation.');
+    return;
+  }
+
+  const validationErrors = validate.errors;
```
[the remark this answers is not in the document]
hacksparrow: `validate.errors` is where the errors are located for a compiled function like `validate`.
`ajv.errors` contains the errors for from the less-performant `ajv.validate` method.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -47,11 +47,18 @@ export function validateRequestBody(
...
 
   const jsonSchema = convertToJsonSchema(schema);
-  validateValueAgainstJsonSchema(body, jsonSchema, globalSchemas);
+  validateValueAgainstJsonSchema(schemaRef, body, jsonSchema, globalSchemas);
+}
+
+function getSchemaRef(requestBodySpec: RequestBodyObject): string {
+  const content = requestBodySpec.content;
+  const schema = content[Object.keys(content)[0]].schema as SchemaObject;
+  return schema[Object.keys(schema)[0]];
```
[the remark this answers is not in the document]
hacksparrow: `Map`/`WeakMap` is exactly what we need. I was also a little doubtful about `getSchemaRef ()`, but needed someone with more context to point out the possible problems, if any.

hacksparrow:
@b-admike this is a performance patch, so the best way to see the outcome is to run https://github.com/strongloop/loopback-next/tree/master/benchmark and see the results.

hacksparrow:
@raymondfeng implemented using `WeakMap`, PTAL.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -87,10 +87,17 @@ function convertToJsonSchema(openapiSchema: SchemaObject) {
  * @param schema The JSON schema used to perform the validation.
  * @param globalSchemas Schema references.
  */
+export interface AnyObject {
+  // tslint:disable-next-line:no-any
+  [property: string]: any;
+}
+
+const validateSchema = new WeakMap();
```
raymondfeng: Maybe a better name such as `cacheForCompiledSchemas`?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -102,22 +109,23 @@ function validateValueAgainstJsonSchema(
...
-      return;
-    }
-  } catch (err) {
-    debug('Fails to execute AJV validation:', err);
-    // TODO: [rfeng] Do we want to introduce a flag to disable validation
-    // or sink validation errors?
-    throw err;
+  if (!validateSchema.has(jsonSchema)) {
+    const compiled = ajv.compile(schemaWithRef);
+    validateSchema.set(jsonSchema, compiled);
```
virkt25: Wouldn't `validateSchema` become unnecessarily large (which may hurt performance?) if the key we are setting is the entire schema itself? ... is it not possible to have a simpler key?

raymondfeng: No, the key is just the internal object id. And it's a weak map which won't prevent the schema from being GCed if it's no longer referenced anywhere else.

packages/rest/test/unit/request-body.validator.test.ts
```diff
@@ -333,9 +333,6 @@ function verifyValidationRejectsInputWithError(
 ) {
   try {
     validateRequestBody(body, spec, schemas);
-    throw new Error(
```
jannyHou: Any reason to remove this line? 
Without it if `validateRequestBody(body, spec, schemas)` doesn't throw error, the test will pass while it's expected to fail.

hacksparrow: That `Error` was to  "emit" a wrong error if the `try`-`catch` in previous code did not throw. We are not `try`-`catching` anymore.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -87,10 +87,17 @@ function convertToJsonSchema(openapiSchema: SchemaObject) {
  * @param schema The JSON schema used to perform the validation.
  * @param globalSchemas Schema references.
  */
+export interface AnyObject {
```
bajtos: AFAICT, this interface is not used anywhere. Can you remove it please?

Also please check the tsdoc comment above - it belongs to `validateValueAgainstJsonSchema`, not AnyObject!

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -87,10 +87,17 @@ function convertToJsonSchema(openapiSchema: SchemaObject) {
...
+  [property: string]: any;
+}
+
+const compiledSchemaCache = new WeakMap();
+
 function validateValueAgainstJsonSchema(
   // tslint:disable-next-line:no-any
   body: any,
-  jsonSchema: object,
+  jsonSchema: SchemaObject,
```
bajtos: Please note that `SchemaObject` is referring to OpenAPI Spec schema, while this method is accepting JSON Schema. These two schemas are slightly different. Please revert this line.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -87,10 +87,17 @@ function convertToJsonSchema(openapiSchema: SchemaObject) {
  * @param schema The JSON schema used to perform the validation.
  * @param globalSchemas Schema references.
  */
+export interface AnyObject {
+  // tslint:disable-next-line:no-any
+  [property: string]: any;
+}
+
+const compiledSchemaCache = new WeakMap();
```
bajtos: Singletons are usually an anti-pattern to avoid.

How is this cache going to behave in regards to memory management and garbage collection? For example, when a route (API endpoint) or perhaps an entire LB4 app is deleted (removed from the process), when will the cached validators be released by V8?

Have you considered storing the compiled validators in Route objects as have been proposed in the issue description?

Let's discuss!

packages/rest/test/unit/request-body.validator.test.ts
bajtos: > We are not try-catching anymore.

This does not match my understanding of this test code, I still see try/catch in place.

Please keep revert the changes in this test.

packages/rest/src/validation/request-body.validator.ts
hacksparrow: WeakMap keys are just references to object IDs, so we don't have to worry about memory management and garbage collection.

I opted for a caching object right in the validator for the sake of simplicity. Would this cache be ever accessed from outside the validator? If not, then I am not very convinced it should be located outside the validator.

packages/rest/test/unit/request-body.validator.test.ts
hacksparrow: Updating to be consistent.

hacksparrow:
@bajtos @jannyHou 😭I removed `throw new Error("expected Function { name: 'validateRequestBody' } to throw exception")` because we are not running `ajv.validate()` in a `try`-`catch` block anymore, but forgot we still have to use a `try`-`catch` in our test since our validator still throws. I have put it back, because without it we won't know the full truth. Thanks for pointing out.

packages/rest/src/validation/request-body.validator.ts
bajtos: > WeakMap keys are just references to object IDs, so we don't have to worry about memory management and garbage collection.

How about the values stored in the map? When are those garbage collected?

I'd like to avoid the situation when removing a controller from the app does not release all associated memory. This used to be a problem in LoopBack 2.x/3.x and it was non-trivial to discover all sources of leaks.

Issue #1745, opened 2018-09-21
Scaffolded LB4 app reports 404 for the home page


raymondfeng:
## Description / Steps to reproduce / Feature proposal

```
lb4 my-app
cd my-app
npm start
```

It displays:
```
Server is running at http://[::1]:3000
Try http://[::1]:3000/ping
```

## Current Behavior

Open your browser and point to http://[::1]:3000, you'll see 404 Error.

## Expected Behavior

We should have a default home page to promote LoopBack and guide the developer to use the API explorer and other resources.

Adding this feature will help `loopback4-example-shopping` too before we create a UI.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

raymondfeng:
Possible quick fixes for the generated app:

1. Add `public/index.html` (Either simple page or redirect to /explorer). This also allows our users to add static assets into `public` by convention.

2. Call `app.static('/', path.join(__dirname, '../../public');`  ==> At this moment `/` is not allowed.

virkt25:
> Possible quick fixes

Since we don't allow static file mounting to `/` path intentionally, maybe the above won't be a quick fix. How about if we using a controller, return some `html` or `text` saying `Welcome to your application`. Try your API's by visiting (link to explorer)?

Alternatively if we had some way of checking to see if a controller has been defined or not for `/` ... and if not then we should just redirect the user to the explorer?

bajtos:
> How about if we using a controller, return some html or text saying Welcome to your application. Try your API's by visiting (link to explorer)?

I like this approach. Similarly to how we add a `PingController` to a newly created app, we can also add e.g. `HomePageController` that will serve the root endpoint `/`.

I am proposing to serve a simple self-contained HTML page that includes all necessary CSS styles, etc., so that we don't need any other static middleware to have it correctly rendered in the browser. The page should contain the application name and description (from `package.json`), a link to OpenAPI JSON specification (e.g. `/openapi.json`) and a link to API explorer.

dhmlau:
> I am proposing to serve a simple self-contained HTML page that includes all necessary CSS styles, etc., so that we don't need any other static middleware to have it correctly rendered in the browser. The page should contain the application name and description (from package.json), a link to OpenAPI JSON specification (e.g. /openapi.json) and a link to API explorer.
[...]

PR #1763, opened 2018-09-26
feat(cli): add code template for default home page controller

raymondfeng:
This PR adds code template for a default home page controller to be generated as part of `lb4 app`.

Fixes https://github.com/strongloop/loopback-next/issues/1745

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/public/index.html.ejs
+++ b/packages/cli/generators/app/templates/public/index.html.ejs
@@ -1,0 +1,73 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title><%= project.description %></title>
+
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="shortcut icon" type="image/x-icon" href="//v4.loopback.io/favicon.ico">
+
+  <style>
+    h3 {
+      margin-left: 25px;
+      text-align: center;
+    }
+
+    a, a:visited {
+      color: #3f5dff;
+    }
+
+    h3 a {
+      margin-left: 10px;
+    }
+
+    a:hover, a:focus, a:active {
+      color: #001956;
+    }
+
+    .power {
+      position: absolute;
+      bottom: 25px;
+      left: 50%;
+      transform: translateX(-50%);
+    }
+
+    .info {
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%)
+    }
+
+    .info h1 {
+      text-align: center;
+      margin-bottom: 0px;
+    }
+
+    .info p {
+      text-align: center;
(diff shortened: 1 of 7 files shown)


packages/cli/test/integration/generators/app.integration.js
```diff
@@ -63,6 +63,18 @@ describe('app-generator specific files', () => {
...
     );
+    assert.fileContent(
+      'src/controllers/home-page.controller.ts',
+      /export class HomePageController/,
+    );
+    assert.fileContent(
+      'src/controllers/home-page.controller.ts',
+      /homePage\(\)/,
+    );
+    assert.fileContent(
```
virkt25: Can we also assert `./test-helper.ts` file

packages/cli/generators/app/templates/test/acceptance/test-helper.ts.ejs
```diff
@@ -0,0 +1,24 @@
+import {<%= project.applicationName %>} from '../..';
+import {
+  createRestAppClient,
+  givenHttpServerConfig,
+  Client,
+} from '@loopback/testlab';
+
+export interface AppWithClient {
```
virkt25: Can we move this to be below the function?

packages/cli/generators/app/templates/public/index.html.ejs
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title><%= project.description %></title>
+
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
```
virkt25: Update to the newest design so we have a favicon (otherwise loading this throws a console error in the browser (if someone was to check)).

raymondfeng:
@virkt25 Updated. PTAL.

packages/cli/generators/app/templates/public/index.html.ejs
```diff
@@ -0,0 +1,73 @@
...
+      margin-bottom: 3em;
+      margin-top: 1em;
+    }
+  </style>
+</head>
+
+<body>
+  <div class="info">
+    <h1><%= project.name %></h1>
+    <p>Version <%= project.version || '1.0.0' %></p>
```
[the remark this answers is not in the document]
raymondfeng: This was intentional as the first step:

We wanted to keep the HomePageController implementation simple and easy to understand.
  - The generated `public/index.html` is a plain html instead of a template. We set the name/version during code gen but it's up to developers to maintain it
  - There is no need to use a template engine to render the view

That's been said, I'm fine to upgrade it to an EJS view once https://github.com/strongloop/loopback-next/pull/1764 is landed.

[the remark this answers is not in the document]
bajtos: Can we remove the Version from the static page then? Application name is unlikely to change, but the version is going to change often.

PR #1764, opened 2018-09-26
feat(boot): bind content of package.json to app context

raymondfeng:
See https://github.com/strongloop/loopback4-example-shopping/pull/16#discussion_r220140265

## Checklist

### DIFF
--- a/packages/boot/src/mixins/boot.mixin.ts
+++ b/packages/boot/src/mixins/boot.mixin.ts
@@ -8,6 +8,8 @@ import {Booter, BootOptions, Bootable} from '../interfaces';
 import {BootComponent} from '../boot.component';
 import {Bootstrapper} from '../bootstrapper';
 import {BootBindings} from '../keys';
+import {CoreBindings} from '@loopback/core';
+import * as path from 'path';
 
 // Binding is re-exported as Binding / Booter types are needed when consuming
 // BootMixin and this allows a user to import them from the same package (UX!)
@@ -53,6 +55,11 @@ export function BootMixin<T extends Constructor<any>>(superClass: T) {
       this.bind(BootBindings.BOOT_OPTIONS).toDynamicValue(
         () => this.bootOptions,
       );
+
+      this.bind(CoreBindings.APPLICATION_PACKAGE_JSON).toDynamicValue(() => {
+        const pkgFile = path.resolve(this.projectRoot, 'package.json');
+        return require(pkgFile);
+      });
     }
 
     /**
--- a/packages/boot/test/acceptance/controller.booter.acceptance.ts
+++ b/packages/boot/test/acceptance/controller.booter.acceptance.ts
@@ -7,9 +7,11 @@ import {
   createRestAppClient,
   givenHttpServerConfig,
   TestSandbox,
+  expect,
 } from '@loopback/testlab';
 import {resolve} from 'path';
 import {BooterApp} from '../fixtures/application';
+import {CoreBindings} from '@loopback/core';
 
 describe('controller booter acceptance tests', () => {
   let app: BooterApp;
@@ -20,6 +22,15 @@ describe('controller booter acceptance tests', () => {
   beforeEach(getApp);
 
   afterEach(stopApp);
+
+  it('binds package.json', async () => {
+    const pkg = await app.get(CoreBindings.APPLICATION_PACKAGE_JSON);
+    expect(pkg).containEql({
+      name: 'boot-test-app',
+      version: '1.0.0',
+      description: 'boot-test-app',
+    });
+  });
 
   it('binds controllers using ControllerDefaults and REST endpoints work', async () => {
     await app.boot();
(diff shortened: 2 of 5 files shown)


packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -53,6 +55,11 @@ export function BootMixin<T extends Constructor<any>>(superClass: T) {
       this.bind(BootBindings.BOOT_OPTIONS).toDynamicValue(
         () => this.bootOptions,
       );
+
+      this.bind(CoreBindings.APPLICATION_PACKAGE_JSON).toDynamicValue(() => {
```
virkt25: Would it not be better for this to be done via a `Booter`? This is not the intended usage of `bootmixin` and seems like concerns are being mixed.

raymondfeng: I don't mind moving it to a `application` booter in the future once we figure out our configuration loading story, such as how to deal with env vars.

packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -53,6 +55,11 @@ export function BootMixin<T extends Constructor<any>>(superClass: T) {
...
+        const pkgFile = path.resolve(this.projectRoot, 'package.json');
+        return require(pkgFile);
```
virkt25: Should we do a try/catch here in case `package.json` path is resolved incorrectly or the file doesn't exist (not sure why that may happen ... but in case it does).

packages/boot/test/acceptance/controller.booter.acceptance.ts
```diff
@@ -21,6 +23,15 @@ describe('controller booter acceptance tests', () => {
 
   afterEach(stopApp);
 
+  it('binds package.json', async () => {
```
bajtos: Please move this test to a more appropriate file, maybe create a new test file in `test/acceptance` directory?

It looks wrong to me to see package.json-related tests inside `controller booter acceptance tests`.

(I think this is another clue that package.json should be handled by a new booter.)

packages/core/src/application.ts
```diff
@@ -192,6 +192,10 @@ export class Application extends Context {
     const instance = this.getSync<Component>(componentKey);
     mountComponent(this, instance);
   }
+
+  public packageJson(pkg: PackageJson) {
```
bajtos: I find this name confusing for users and readers.

```ts
app.packageJson(/*...*/);
```

Can we find a better name, for example `setApplicationMetadata`? (I do realize this is not consistent with other API like `app.component` or `app.controller`.)

packages/core/src/application.ts
```diff
@@ -207,3 +211,14 @@ export interface ApplicationConfig {
 
 // tslint:disable-next-line:no-any
 export type ControllerClass = Constructor<any>;
+
+/**
+ * Type description for `package.json`
+ */
+export interface PackageJson {
```
bajtos: JSON is a serialization format. Can we replace it with something else in the interface name please? How about `PackageManifest`, `PackageMetadata` or even `ApplicationManifest`/`ApplicationMetadata`?

I think I like `ApplicationMetadata` most.

packages/core/src/application.ts
```diff
@@ -207,3 +211,14 @@ export interface ApplicationConfig {
...
+  name: string;
+  version: string;
+  description: string;
+  // tslint:disable-next-line:no-any
+  [name: string]: any;
```
bajtos: If this data object is typically loaded from JSON, then I think we should restrict the allowed property types to `number | string | object | array | null`. JSON does not support functions and Dates for example. The current type definition makes it easy to accidentally treat a package.json property as a function and call it.

packages/core/src/application.ts
```diff
@@ -192,6 +192,16 @@ export class Application extends Context {
     const instance = this.getSync<Component>(componentKey);
     mountComponent(this, instance);
   }
+
+  /**
+   * Set application metadata. By default, the content of `package.json` is
```
bajtos: "By defaults" implies to me that `package.json` metadata is populated by Application constructor, which is not true.

Can we make it more clear please that this metadata is empty by default but `@loopback/boot` populates is from `package.json`?

packages/core/src/application.ts
bajtos: > If this data object is typically loaded from JSON, then I think we should restrict the allowed property types to `number | string | object | array | null`. JSON does not support functions and Dates for example. The current type definition makes it easy to accidentally treat a package.json property as a function and call it.

This comment hasn't been addressed nor responded to, PTAL.

raymondfeng: Fixed

raymondfeng:
@bajtos Fixed. PTAL.

packages/boot/test/acceptance/controller.booter.acceptance.ts
```diff
@@ -33,6 +33,7 @@ describe('controller booter acceptance tests', () => {
   });
 
   async function getApp() {
+    await sandbox.copyFile(resolve(__dirname, '../fixtures/package.json'));
```
bajtos: I think this will not work after the changes  I made in https://github.com/strongloop/loopback-next/pull/1824) - the `package.json` file is not going to be copied over from `test` to `dist/test`.

I think the easiest solution is to load the package json file from `test/fixtures/application.ts`:

```ts
import * as pkg from './package.json'
```

PR #1765, opened 2018-09-26
feat(rest): Implement trie-based routing and tidy up path validation

raymondfeng:
The PR introduces the following improvements:

- validate and normalize paths for openapi
- add benchmark for routing
- make the router pluggable
- simplify app routing to plain functions
- optimize simple route mapping and use regexp as default
- add `rest.router` binding to allow customization
- add docs for routing requests

3. Optimize routes without variables to use a direct lookup 

## Checklist


packages/rest/src/router/routing-table.ts
```diff
@@ -127,6 +128,8 @@ export class RoutingTable {
       );
     }
     this._routes.push(route);
+    // Sort routes
+    this._routes.sort(compareRoute);
```
virkt25: What benefit do we get by sorting the routes?

packages/openapi-v3/test/unit/to-openapi-path.unit.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import {toOpenApiPath} from '../../src/controller-spec';
+
+describe('toOpenApiPath', () => {
+  it('allows /:foo/bar', () => {
+    const path = toOpenApiPath('/:foo');
+    expect(path).to.eql('/{foo}');
+  });
+
+  it('allows /:foo/bar', () => {
```
virkt25: We're going to start supporting express style of parameters? If so, we should document it.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -80,7 +81,7 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
 
     const endpoint = endpoints[op];
     const verb = endpoint.verb!;
-    const path = endpoint.path!;
+    const path = toOpenApiPath(endpoint.path!);
```
virkt25: Hmm ... why is `path` not a required property? (If it was we wouldn't need the `!`.

packages/rest/src/router/routing-table.ts
raymondfeng: Think about this example:
```ts
/**
 * A simple controller to bounce back http requests
 */
export class PingController {
  constructor(@inject(RestBindings.Http.REQUEST) private req: Request) {}

  @get('/ping/:me')
  pingMe(@param.path.string('me') me: string): object {
    return {
      greeting: `Hello from ${me}`,
    };
  }

  // Map to `GET /ping`
  @get('/ping')
  ping(): object {
    return {
      greeting: 'Hello from LoopBack',
      date: new Date(),
      url: this.req.url,
      headers: Object.assign({}, this.req.headers),
    };
  }

  @get('/ping/xyz')
  pingXyz(): object {
    return {
      greeting: 'Hello from XYZ',
    };
  }
}
```

I would expect `GET /ping/xyz` is routed to `pingXyz()` instead of `pingMe()`. Without this PR, it matches `GET /ping/:me`.

raymondfeng: I'll add an integration test.

packages/openapi-v3/test/unit/to-openapi-path.unit.ts
raymondfeng: We didn't do any check before this PR. I think it will be good to allow express style too.

packages/openapi-v3/src/controller-spec.ts
raymondfeng: If `path` is missing, it's default to `/`'.

packages/rest/src/router/routing-table.ts
```diff
@@ -469,3 +477,73 @@ export function createControllerFactoryForInstance<T>(
...
+
+  // Longer path comes before shorter path
+  if (tokensForPath1.length < tokensForPath2.length) {
+    return 1;
+  } else if (tokensForPath1.length > tokensForPath2.length) {
+    return -1;
+  }
+
+  // Now check token by token
+  for (let i = 0; i < tokensForPath1.length; i++) {
```
b-admike: Do we have test coverage for this code path?

packages/openapi-v3/test/unit/to-openapi-path.unit.ts
bajtos: > We didn't do any check before this PR. I think it will be good to allow express style too.

I would rather not introduce multiple parameter styles, I am concerned about the confusion it can create and the extra work it will add for any consumer of routes.

raymondfeng:
@bajtos I reverted the sorting based approach and implemented trie-based routing based on `wayfarer`. PTAL.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -275,3 +276,38 @@ export function getControllerSpec(constructor: Function): ControllerSpec {
   }
   return spec;
 }
+
+function validatePath(path: string = '/') {
+  let tokens = pathToRegExp.parse(path);
+  if (tokens.some(t => typeof t === 'object')) {
+    throw new Error(
+      `Invalid path template: '${path}'. Please use {param} instead of ':param'`,
```
bajtos: 👍

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -275,3 +276,38 @@ export function getControllerSpec(constructor: Function): ControllerSpec {
...
+
+/**
+ * Validate and normalize the path to OpenAPI path template. No parameter
+ * modifier, custom pattern, or unnamed parameter is allowed.
+ *
+ * @param path Express style path (https://github.com/pillarjs/path-to-regexp)
+ */
+export function toOpenApiPath(path: string = '/') {
+  validatePath(path);
+  return path.replace(/\:([A-Za-z0-9_]+)/g, '{$1}');
```
bajtos: Is this `replace` call still needed, considering that `validatePath` rejects values containing express-style parameters? Maybe I am missing something?

packages/openapi-v3/test/unit/to-openapi-path.unit.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('disallows /:foo/bar', () => {
+    disallows('/:foo/bar');
+  });
+
+  it('disallows /:foo/:bar', () => {
+    disallows('/:foo/:bar');
+  });
+
+  it('disallows /:foo+', () => {
```
bajtos: Should we also test `disallows /{foo+}` and similar?

packages/openapi-v3/test/unit/to-openapi-path.unit.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('disallows /:foo*', () => {
+    disallows('/:foo*');
+  });
+
+  it('disallows /:foo(\\d+)', () => {
+    disallows('/:foo(\\d+)');
+  });
+
+  it('disallows /foo/(.*)', () => {
```
bajtos: Should we also test `disallows /foo/{.*}`?

packages/rest/src/router/routing-table.ts
```diff
@@ -127,6 +126,10 @@ export class RoutingTable {
       );
     }
     this._routes.push(route);
+    // Add the route to the trie
+    const path = route.path.startsWith('/') ? route.path : `/${route.path}`;
+    const verb = route.verb.toLowerCase() || 'get';
+    this._router.on(`/${verb}${path}`, () => {});
```
bajtos: Can we store `route` on the function passed to `this._router.on` to avoid costly lookup (iteration through `this._routes` later on? E.g.

```ts
// store
this._router.on(`/${verb}${path}`, Object.assign(() => {}, {route}));

// pick up - I hope that's the Wayfarer API
const found = this._router.match(path);
const route = found.route;
```

packages/rest/src/router/routing-table.ts
```diff
@@ -148,11 +151,26 @@ export class RoutingTable {
...
+        // Normalize the path
+        const p = r.path.startsWith('/') ? r.path : `/${r.path}`;
+        return method === r.verb && `/${method}${p}` === found.route;
+      });
+      if (route) {
+        debug('Route found: %s', inspect(route, {depth: 5}));
+        return createResolvedRoute(route, found.params);
+      }
+    } catch (err) {
+      // Ignore the error, will report as NotFound
```
bajtos: Please add a debug log to make troubleshooting easier when something unexpected happens inside wayfarer.

packages/rest/test/acceptance/routing/routing.acceptance.ts
```diff
@@ -700,6 +700,48 @@ describe('Routing', () => {
...
+
+      app.api(spec);
+      app.controller(MyController);
+
+      await whenIMakeRequestTo(app)
+        .get('/greet/john')
+        .expect(200, 'hello john');
+      await whenIMakeRequestTo(app)
+        .get('/greet/world')
+        .expect(200, 'hello WORLD');
```
bajtos: I like this test much more than the previous version 👍

bajtos:
Could you also update the benchmark to show the new performance with Wayfarer under the hood? See https://github.com/strongloop/loopback-next/tree/master/benchmark

raymondfeng:
@b-admike @bajtos I pushed more changes to use _router to store routes per @bajtos suggestion. PATL.

raymondfeng:
I added optimization for those routes that don't have path variables. The `RoutingTable` by default uses `RegExpRouter` now.

b-admike:
@raymondfeng I looked at your recent changes at a high level and they look great 👍. I'll look into them in detail tomorrow.

packages/rest/src/rest.server.ts
```diff
@@ -281,7 +282,14 @@ export class RestServer extends Context implements Server, HttpServerLike {
     // See https://github.com/strongloop/loopback-next/issues/433
     if (this._httpHandler) return;
 
-    this._httpHandler = new HttpHandler(this);
+    const router = this.getSync(RestBindings.ROUTER, {optional: true});
+    let routingTable: RoutingTable;
+    if (router) {
```
bajtos: Can we move this `if` to `RoutingTable` constructor so that RestServer can be simplified as follows:

```ts
const routingTable = new RoutingTable(router);
```

packages/rest/src/router/index.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+export {
+  RouteEntry,
```
bajtos: Considering how many items we are exporting, I think it may be best to simply export everything.

```ts
export * from `./routing-table';
```

Thoughts?

packages/rest/src/router/routing-table.ts
```diff
@@ -57,11 +55,28 @@ export type ControllerFactory<T extends ControllerInstance> = (
...
+}
+
 /**
  * Routing table
  */
 export class RoutingTable {
-  private readonly _routes: RouteEntry[] = [];
+  private readonly _router: RestRouter = new RegExpRouter();
+
+  constructor(router?: RestRouter) {
```
bajtos: How about:

```ts
constructor(private readonly router: RestRouter = new RegExpRouter()) {
  // ...
}
```

raymondfeng:
@bajtos Addressed your coding comments and added docs. PTAL.

I would like to have it landed with our GA as the mapping of routes without variables are much faster now.

docs/site/Routing-requests.md
```diff
@@ -0,0 +1,87 @@
...
+By default, `@loopback/rest` uses `RegExpRouter` as it performs well with up to
+a few hundred of routes. There is a simple benchmarking for `RegExpRouter` and
+`TrieRouter` at
+https://githhub.com/strongloop/loopback-next/benchmark/src/rest-routing/routing-table.ts.
+
+To change the router for REST routing, we can bind the router class as follows:
+
+```ts
+import {RestBindings} from '@loopback/rest';
+app.bind(RestBindings.ROUTER).toClass(TrieRouter);
```
b-admike: nitpick: Which file would this be in and where can users import `TrieRouter` from?

packages/rest/src/rest.server.ts
```diff
@@ -281,7 +282,13 @@ export class RestServer extends Context implements Server, HttpServerLike {
     // See https://github.com/strongloop/loopback-next/issues/433
     if (this._httpHandler) return;
 
-    this._httpHandler = new HttpHandler(this);
+    /**
+     * Check if there is custom router in the context
+     */
+    const router = this.getSync(RestBindings.ROUTER, {optional: true});
```
b-admike: Do we need a `try/catch` block for the `getSync` call here?

packages/rest/test/acceptance/routing/routing.acceptance.ts
```diff
@@ -701,7 +702,16 @@ describe('Routing', () => {
...
     });
 
-    it('sorts routes based on their specifics', async () => {
+    it('allows pluggable router', async () => {
+      const app = new RestApplication();
+      app.bind(RestBindings.ROUTER).toClass(TrieRouter);
+      const server = await app.getServer(RestServer);
+      const handler = await server.get(RestBindings.HANDLER);
+      // tslint:disable-next-line:no-any
+      expect((handler as any)._routes._router).to.be.instanceof(TrieRouter);
```
b-admike: nitpick: why do we need to cast handler as `any` here? can we do `const handler = await server.get<HttpHandler>(RestBindings.HANDLER);` above?

packages/rest/src/rest.server.ts
raymondfeng: No, as it uses an `optional` flag.

packages/rest/test/acceptance/routing/routing.acceptance.ts
raymondfeng: It's a hacky way to test non-public properites.

docs/site/Routing-requests.md
raymondfeng: Good catch.

packages/rest/test/acceptance/routing/routing.acceptance.ts
bajtos: @raymondfeng Please add a comment to the code to preserve this explanation for future readers.

packages/rest/package.json
```diff
@@ -44,7 +44,8 @@
     "path-to-regexp": "^2.2.0",
     "qs": "^6.5.2",
     "strong-error-handler": "^3.2.0",
-    "validator": "^10.4.0"
+    "validator": "^10.4.0",
+    "wayfarer": "^6.6.4"
```
bajtos: AFAICT, `wayfarer` is not used by our code, can you remove this dependency please?

packages/rest/src/rest.application.ts
```diff
@@ -141,23 +142,44 @@ export class RestApplication extends Application implements HttpServerLike {
...
+   * }
+   * app.route('get', '/', operationSpec, greet);
+   * ```
+   */
+  route(
+    verb: string,
+    path: string,
+    spec: OperationObject,
+    handler: Function,
+  ): Binding;
```
bajtos: Please use this new API to simplify `Routing` doc page - see https://loopback.io/doc/en/lb4/Routes.html#using-partial-openapi-spec-fragments

packages/rest/src/rest.server.ts
```diff
@@ -281,7 +282,13 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+    const routingTable = new RoutingTable(router);
+
+    this._httpHandler = new HttpHandler(this, routingTable);
```
bajtos: I am wondering if it's a good idea for HttpHandler to accept RoutingTable instance. Have you considered an alternative design where HttpHandler accepts a router instance only?

```ts
const router = this.getSync(RestBindings.ROUTER, {optional: true});
this._httpHandler = new HttpHandler(this, router);
```

IMO, the RoutingTable class is an implementation detail of HttpHandler that's not open for extensibility to consumers.

packages/rest/src/router/openapi-path.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+/**
+ * Validate and normalize the path to OpenAPI path template. No parameter
+ * modifier, custom pattern, or unnamed parameter is allowed.
+ *
+ * @param path Path template with `{var}`
+ */
+export function toOpenApiPath(path: string = '/') {
```
bajtos: I still don't understand the purpose of this `toOpenApiPath` function - what are the benefits of calling `toOpenApiPath` as compared to calling `validatePath` directly?

packages/rest/src/router/openapi-path.ts
```diff
@@ -0,0 +1,72 @@
...
+ * allows `[A-Za-z0-9_]`
+ */
+const POSSIBLE_VARNAME_PATTERN = /\{([^\}]+)\}/g;
+const INVALID_VARNAME_PATTERN = /\{([^\}]*[^\w\}][^\}]*)\}/;
+
+/**
+ * Validate the path to be compatible with OpenAPI path template. No parameter
+ * modifier, custom pattern, or unnamed parameter is allowed.
+ */
+export function validatePath(path: string = '/') {
```
bajtos: Perhaps `validateApiPath` would be a better name, to make it clear that we are validating REST API/URL paths and not filesystem paths?

packages/rest/src/router/regexp-router.ts
```diff
@@ -0,0 +1,94 @@
...
+  private _sorted: boolean;
+  private _sort() {
+    if (!this._sorted) {
+      this.routes.sort(compareRoute);
+      this._sorted = true;
+    }
+  }
+
+  add(route: RouteEntry) {
+    if (super.add(route)) return true;
```
bajtos: I find this design brittle, as it's relying on developers implementing custom Router classes to not forget to handle this special case.

Have you considered using Template Design Pattern?

```ts
export class BaseRouter {
  add(route: RouteEntry) {
     if (getPathVariables(route.path)) {
       this.addTemplatedRoute(route);
     } else {
       const key = this.getKeyForRoute(route);
       this.simpleRoutes[key] = route;
     }
  }

  protected addTemplatedRoute(route: RouteEntry) {
    throw new Error('This method is abstract');
  }

  // ...
}

export class RegExpRouter extends BaseRouter {
   private routes: RegExpRouteEntry[] = [];

  addTemplatedRoute(route: RouteEntry) {
     const entry = route as RegExpRouteEntry;
     // ...
  }

  // ...
}
```

Similarly for find/list.

packages/rest/src/router/regexp-router.ts
```diff
@@ -0,0 +1,94 @@
...
+      const pathParams = this._buildPathParams(r, match);
+      debug(' -> found with params: %j', pathParams);
+
+      return createResolvedRoute(r, pathParams);
+    }
+    return undefined;
+  }
+
+  list() {
+    const basicRoutes = Object.values(this.simpleRoutes);
```
bajtos:
```ts
const basicRoutes = Object.values(this.simpleRoutes);
```

Let's use a single name (basic vs simple) everywhere please.

I personally dislike terms "basic" and "simple" because they don't explain what makes the target basic or simple.

How about `staticRoutes` our `paramlessRoutes`?

packages/rest/src/router/router-base.ts
```diff
@@ -0,0 +1,49 @@
...
+  protected simpleRoutes: {[path: string]: RouteEntry} = {};
+
+  protected getKeyForRoute(route: RouteEntry) {
+    const path = route.path.startsWith('/') ? route.path : `/${route.path}`;
+    const verb = route.verb.toLowerCase() || 'get';
+    return `/${verb}${path}`;
+  }
+
+  add(route: RouteEntry) {
+    const key = this.getKeyForRoute(route);
```
bajtos: Let's avoid preliminary pessimization and obtain the key only if we are going to register this route ourselves.

packages/rest/src/router/router-base.ts
```diff
@@ -0,0 +1,49 @@
...
+import {getPathVariables} from './openapi-path';
+
+/**
+ * Base router implementation that only handles path without variables
+ */
+export abstract class BaseRouter implements RestRouter {
+  /**
+   * A map to optimize matching for routes without variables in the path
+   */
+  protected simpleRoutes: {[path: string]: RouteEntry} = {};
```
bajtos: Let's find a better name please, e.g. `staticRoutes` our `paramlessRoutes`.

packages/rest/src/router/routing-table.ts
```diff
@@ -126,13 +147,16 @@ export class RoutingTable {
         describeOperationParameters(route.spec),
       );
     }
-    this._routes.push(route);
+
+    validatePath(route.path);
+    // Add the route to the trie
```
bajtos: I find this comment misleading, we don't use trie by default. Let's remove it completely.

packages/rest/src/router/trie-router.ts
```diff
@@ -0,0 +1,53 @@
...
+import {Request} from '../types';
+import {inspect} from 'util';
+import {BaseRouter} from './router-base';
+
+const debug = require('debug')('loopback:rest:router:trie');
+
+/**
+ * Router implementation based on trie
+ */
+export class TrieRouter extends BaseRouter {
```
bajtos: AFAICT, there are no test for TrieRouter and the code coverage of this file is ~20% - see https://coveralls.io/builds/19364099/source?filename=packages/rest/src/router/trie-router.ts

Please add some tests to ensure this code works as intended.

Considering how little time we have before 4.0 GA, it may be better to leave `TrieRoute` out of scope of this pull request.

packages/rest/src/router/trie-router.ts
```diff
@@ -0,0 +1,53 @@
...
+    const path = `/${method}${request.path}`;
+
+    let route = this.simpleRoutes[path];
+    if (route) return createResolvedRoute(route, {});
+
+    const found = this.trie.match(path);
+    debug('Route matched: %j', found);
+
+    if (found) {
+      route = found.node.value!;
```
bajtos: Can you fix the type definition of `trie.match` to communicate the fact that `value` is always set when a route was found?

For example:

```ts
export interface ResolvedNode<T> {
   node: Node<T> & {value: T}; // the value is always set
   params?: {[name: string]: any};
 }
```

packages/rest/src/router/trie.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Match a route path against the trie
+   * @param path The route path, such as `/customers/c01`
+   */
+  match(path: string): ResolvedNode<T> | undefined {
+    const keys = path.split('/').filter(Boolean);
+    const params = {};
+    const resolved = search(keys, 0, params, this.root);
+    if (resolved && resolved.node.value != null) return resolved;
```
bajtos:
```ts
if (resolved && this.isNodeWithValue(resolved.node)) return resolved;
```

packages/rest/src/router/trie.ts
```diff
@@ -0,0 +1,221 @@
...
+  for (const k in parent.children) {
+    child = parent.children[k];
+    if (!child.names || !child.regexp) continue;
+    const match = child.regexp.exec(key);
+    if (match) {
+      const resolved: ResolvedNode<T> = {params: {}, node: child};
+      let i = 0;
+      for (const n of child.names) {
+        const val = match[++i];
+        resolved.params![n] = decodeURIComponent(val);
```
bajtos: Can we avoid relying on `!` parameter?

E.g.

```ts
// you should probably define a new type for param map (don't we already have one?)
const params: {[name: string]: any} = {};
let i = 0;
for (const n of child.names) {
  const val = match[++i];
  params[n] = decodeURIComponent(val);
}
return {params, node: child};
```

packages/rest/test/unit/router/routing-table.regexp.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  it('finds simple "GET /hello" endpoint', () => {
+    const spec = anOpenApiSpec()
+      .withOperationReturningString('get', '/hello', 'greet')
+      .build();
+
+    class TestController {}
+
+    const table = givenRoutingTable();
+    table.registerController(spec, TestController);
```
bajtos: Have you considered using route-base approach to avoid the complexity of dealing with controllers?

packages/rest/test/unit/router/routing-table.regexp.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+    const route = table.find(request);
+
+    expect(route).to.be.instanceOf(ControllerRoute);
+    expect(route)
+      .to.have.property('spec')
+      .containEql(spec.paths['/hello'].get);
+    expect(route).to.have.property('pathParams');
+    expect(route.describe()).to.equal('TestController.greet');
+  });
+
```
bajtos: Please add tests to verify how templated paths like `/hello/{name}` are handled.

raymondfeng:
@bajtos I cleaned up the code based on your feedback.

The final benchmark now shows `trie` based router performs much better than `regexp`. I must have messed up some of the testing before.

```
name                 duration    count    found   missed
TrieRouter          0,1453883       10        8        2
RegExpRouter        0,1220030       10        8        2

name                 duration    count    found   missed
TrieRouter          0,2109957       40       35        5
RegExpRouter        0,8762936       40       35        5

name                 duration    count    found   missed
TrieRouter          0,4895252      160      140       20
RegExpRouter       0,52156699      160      140       20

name                 duration    count    found   missed
TrieRouter         0,16065852      640      560       80
RegExpRouter      0,304921026      640      560       80

name                 duration    count    found   missed
TrieRouter         0,60165877     2560     2240      320
RegExpRouter      4,592089555     2560     2240      320
```
I now make `TrieRouter` the default and have test/benchmark coverage for both `TrieRouter` and `RegExpRouter`.

packages/rest/src/rest.server.ts
raymondfeng: Please note that using the constructor the pass down a version of RestRouter is still internal implementation details, which we can change later. The public contract is to bind a custom RestRouter to the given key.

Issue #1471, opened 2018-06-25
[REST] Consistent response format for PATCH/PUT


virkt25:
## Description / Steps to reproduce / Feature proposal

From #1206 

A `PATCH` / `PUT` request returns `true` while other REST endpoints return a JSON output. Should these methods also return a JSON response to be consistent?

## Current Behavior

- Returns a `boolean` for PATCH / PUT

## Expected Behavior

- Return `JSON` ... something like `{success: true}` or `{result: <value-returned-by-juggler>}`
- Update existing code / templates / tests

cc: @strongloop/sq-lb-apex @raymondfeng @bajtos Thoughts? Is this the intended behaviour or a bug? If it's the intended behaviour, do we want to look into changing it for LB4?

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

bajtos:
+1 for returning a JSON like `{success: true}` or perhaps `{result: <value-returned-by-juggler>}`.

Please note that we cannot return full model data in most cases. It would require another database command to fetch the updated model instance, which would introduce race condition.

I am going to label this as a bug and assign to Core-GA backlog for consideration.

jannyHou:
+1 for returning a JSON result. Agree with Miroslav's comment. And it also allows we to customize result details, e.g. Cloudant includes which data fails update and which one passes.

bajtos:
In #1723, I reworked methods returning `boolean` to return `undefined` and report "model not found" via 404 Not Found error.

However, the following endpoints are still returning a naked number:

- `PATCH /my-models` (`repo.updateAll(data, where)`) returns a number of updated instances.
- `GET /my-models/count` (`repo.count(where)`) returns a number of instance found.

I think we should rework these two endpoints to wrap the number in an object:

```json
{
  "count": 123
}
```

For reference, here is the LoopBack 3.x behavior:
 - [count](https://loopback.io/doc/en/lb3/PersistedModel-REST-API.html#get-instance-count) returns an object with `count` property, e.g. `{count: 6}`.
 - [updateAll](http://apidocs.loopback.io/loopback/#persistedmodel-updateall) returns an object with `count` property too.

PR #1770, opened 2018-09-28
fix(repository): return an object for count and updateAll

virkt25:
fixes https://github.com/strongloop/loopback-next/issues/1471

- update repository to return a {count: number} instead of just a `number`
- update `@responses`, examples, cli, docs.

## Checklist

### DIFF
--- a/packages/repository/src/common-types.ts
+++ b/packages/repository/src/common-types.ts
@@ -82,3 +82,20 @@ export type NamedParameters = AnyObject;
  */
 // tslint:disable-next-line:no-any
 export type PositionalParameters = any[];
+
+/**
+ * Count of Model instances that were successful for methods like `updateAll`,
+ * `deleteAll`, etc.
+ */
+export interface Count {
+  count: number;
+}
+
+/**
+ * JSON Schema describing the Count interface. It's the response type for
+ * REST calls to APIs which return Count
+ */
+export const CountSchema = {
+  type: 'object',
+  properties: {count: {type: 'number'}},
+};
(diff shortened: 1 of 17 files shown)


packages/repository/src/common-types.ts
```diff
@@ -82,3 +82,12 @@ export type NamedParameters = AnyObject;
  */
 // tslint:disable-next-line:no-any
 export type PositionalParameters = any[];
+
+export interface Count {
```
b-admike: I know this is a straightforward interface, but can we add TSDocs to it to make it clear what it is?

packages/repository/src/repositories/relation.repository.ts
```diff
@@ -99,11 +99,12 @@ export class DefaultHasManyEntityCrudRepository<
...
     options?: Options,
-  ): Promise<number> {
-    return await this.targetRepository.updateAll(
+  ): Promise<Count> {
+    return this.targetRepository.updateAll(
       constrainDataObject(dataObject, this.constraint),
       constrainWhere(where, this.constraint),
       options,
     );
+    // return result;
```
b-admike: nitpick: no need for commented line :)

packages/repository/src/common-types.ts
virkt25: @b-admike Good call. Added comments.

raymondfeng:
In the future, maybe we should consider to allow a normalized json response, such as:
```js
{
  data: {}, // normal result from the handler
  error: {} // error from the handler
}

Issue #1773, opened 2018-09-28
compile error: cannot find module ....@loopback/metadata/src/types


gms1:
## Description / Steps to reproduce / Feature proposal

compile error:
```
> tsc

../node_modules/@loopback/authentication/dist8/src/keys.d.ts:89:99 - error TS2307: Cannot find module '../../context/node_modules/@loopback/metadata/src/types'.

89 export declare const AUTHENTICATION_METADATA_KEY: MetadataAccessor<AuthenticationMetadata, import("../../context/node_modules/@loopback/metadata/src/types").DecoratorType>;
     
```
looks indeed weird:
```
gms@orion:~/tmp$ wget -q https://registry.npmjs.org/@loopback/authentication/-/authentication-0.11.16.tgz
gms@orion:~/tmp$ tar -xzf authentication-0.11.16.tgz 
gms@orion:~/tmp$ find package/ -name "*.d.ts" -exec grep -Hw node_modules {} \;
package/dist8/src/keys.d.ts:export declare const AUTHENTICATION_METADATA_KEY: MetadataAccessor<AuthenticationMetadata, import("../../context/node_modules/@loopback/metadata/src/types").DecoratorType>;
package/dist10/src/keys.d.ts:export declare const AUTHENTICATION_METADATA_KEY: MetadataAccessor<AuthenticationMetadata, import("../../context/node_modules/@loopback/metadata/src/types").DecoratorType>;
gms@orion:~/tmp$ 

```

## Current Behavior
cannot find module
## Expected Behavior
should find module

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

raymondfeng:
It's similar as https://github.com/strongloop/loopback-next/issues/1643. The root problem is https://github.com/Microsoft/TypeScript/issues/26985.

PR #1774, opened 2018-09-28
fix(metadata): remove the default type to work around a TS bug

raymondfeng:
See https://github.com/Microsoft/TypeScript/issues/26985
Fixes https://github.com/strongloop/loopback-next/issues/1773

## Checklist


raymondfeng:
There is no need to revert. The new version is more explicit.

PR #1779, opened 2018-09-29
refactor(cli): update the response OAS for ping controller

marioestradarosa:
This PR removes the invalid headers sent from ping controller related to OAS Graph.

@raymondfeng ,  I am not sure why we had to place headers in the response object the first time. I think they are needed for custom headers only.  we placed a basic header for content-type and additionalProperties set to true.

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
+++ b/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
@@ -16,10 +16,10 @@ const PING_RESPONSE: ResponseObject = {
           url: {type: 'string'},
           headers: {
             type: 'object',
-            patternProperties: {
-              '^.*$': {type: 'string'},
+            properties: {
+              'Content-Type': {type: 'string'},
             },
-            additionalProperties: false,
+            additionalProperties: true,
           },
         },
       },


marioestradarosa:
### Post Man

screen shot 2018-09-29 at 09 56 41

screen shot 2018-09-29 at 09 56 17

### Internal GUI 

screen shot 2018-09-29 at 09 56 30

marioestradarosa:
@raymondfeng , I place now  the following property and it works fine too. 
thought?

```ts
headers: {
            type: 'object',
            properties: {
              'Content-Type': {type: 'string'},
            },
            additionalProperties: true,
          },
```

raymondfeng:
Can u make sure the oas valudator is happy with addtionalProperties?

raymondfeng:
Please note such schema only matters for oas validation. It won’t impact the rest response.

marioestradarosa:
> Can u make sure the oas validator is happy with addtionalProperties?

It is working with no complains. However, if I remove this property still works. 
thoughts?

marioestradarosa:
with no additionalProperties

screen shot 2018-09-29 at 11 13 56

packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
```diff
@@ -16,10 +16,10 @@ const PING_RESPONSE: ResponseObject = {
           url: {type: 'string'},
           headers: {
             type: 'object',
-            patternProperties: {
-              '^.*$': {type: 'string'},
+            properties: {
```
raymondfeng: nitpick: I think we can remove `properties`.

marioestradarosa: No, we can't.  it accepts the removal of headers completely, but it needs the properties: .   what we can remove is the additionalProperties or the headers.

marioestradarosa:
@raymondfeng  I will merge the change since it is small and you approve it !.

PR #1781, opened 2018-09-29
feat(cli): check project deps against cli template

raymondfeng:
Sometimes a user try to run newer version of `lb4` against existing
applications. This feature will prompt users to confirm if some of
the project deps do not satisify the cli template.

It also checks exit status to avoid unnessary prompts.

## Checklist

### DIFF
--- a/packages/cli/lib/base-generator.js
+++ b/packages/cli/lib/base-generator.js
@@ -10,9 +10,8 @@ const chalk = require('chalk');
 const {StatusConflicter, readTextFromStdin} = require('./utils');
 const path = require('path');
 const fs = require('fs');
-const readline = require('readline');
 const debug = require('./debug')('base-generator');
-const assert = require('assert');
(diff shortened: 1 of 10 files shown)


packages/cli/lib/base-generator.js
```diff
@@ -297,8 +296,48 @@ module.exports = class BaseGenerator extends Generator {
...
+    const prompts = [
+      {
+        name: 'ignoreIncompatibleDependencies',
+        message: `Continue to run the command?`,
+        type: 'confirm',
+        default: false,
+      },
+    ];
+    const answers = await this.prompt(prompts);
+    if (answers && answers.ignoreIncompatibleDependencies) {
```
bajtos: Will this work in non-interactive mode when the answers are provided via JSON and optional prompts are skipped?

raymondfeng: yes

PR #1782, opened 2018-09-30
feat(repository): type-safe queries

### LINKED ISSUE #1777: Add support for type-safe queries
## Feature proposal

Hallo loopback team,

please provide a mechanism for building type-safe queries (e.q Where/Filter objects), which would make refactoring much easier.

Building type-safe queries could be done by the use of the TypeScript 'keyof' operator (validity of the property names) in combination with the TypeScript indexed access operator 'T[K]' (validity of the property types) 

sample:
https://github.com/gms1/node-sqlite3-orm/blob/master/src/sqlite3orm/query/Where.ts

## Current Behavior

*  Where/Filter are not type-safe
*  refactoring of entity properties (names/types) is not supported 

## Expected Behavior

*  Where/Filter or similar objects should be type-safe
*  refactoring of entity properties (names/types) should be easy

gms1:
add type-safety to the query definition to have
full IntelliSense and refactoring support

BREAKING CHANGE: logical and comparison conditions must be separated into different objects

fixes  #1777

### about the breaking change, or is it a fix?

To make the compiler happy I had to separate the conditions defined on
properties (see 'PredicateCondition<MT>') from the logical conditions (see first
part of the union type 'Condition<MT>')

currently, neither juggler nor lb4 has stopped us from doing wrong things like:

```TypeScript
{title: 't1', or: [{content: 'Note 2'}, {content: 'Note 3'}]}
```

which results to:

```sql
WHERE title=:1 AND (content=:2) OR (content=:3)
```

while indentional we wanted to do this:

```TypeScript
{and: [{title: 't1'}, {or: [{content: 'Note 2'}, {content: 'Note 3'}]}]}
```

which results to:

```sql
WHERE (title=:1) AND ((content=:2) OR (content=:3))
```

please see how the logical operators are supposed to be used:
https://loopback.io/doc/en/lb3/Where-filter.html#and--or

### WhereBuilder

The WhereBuilder has been changed accordingly, please see the effects in
'query-builder.unit.ts'

regarding the WhereBuilder additional tests are required!

## Checklist

### DIFF
--- a/packages/repository/src/query.ts
+++ b/packages/repository/src/query.ts
@@ -31,21 +31,38 @@ export enum Operators {
 }
 
 /**
+ * Matching predicate comparison
+ */
+export type PredicateComparison<PT> = {
+  eq?: PT;
+  neq?: PT;
+  gt?: PT;
+  gte?: PT;
+  lt?: PT;
+  lte?: PT;
+  inq?: PT;
+  between?: [PT, PT];
+  exists?: boolean;
+};
+
+type ShortHandEqualType = string | number | boolean | Date;
+
+/**
+ * Matching predicate condition
+ */
+export type PredicateCondition<MT> = {
+  [P in keyof MT]?: PredicateComparison<MT[P]> | (MT[P] & ShortHandEqualType)
+} & {and?: never; or?: never};
+
+/**
  * Matching condition
  */
-export interface Condition {
-  eq?: any;
-  neq?: any;
-  gt?: any;
-  gte?: any;
-  lt?: any;
-  lte?: any;
-  inq?: any[];
-  between?: any[];
-  exists?: boolean;
-  and?: Where[];
-  or?: Where[];
-}
+export type Condition<MT> =
+  | {
+      and?: Condition<MT>[]; // AND
+      or?: Condition<MT>[]; // OR
+    }
+  | PredicateCondition<MT>;
 
 /**
  * Where object
@@ -55,11 +72,7 @@ export interface Condition {
  * `{and: [{fieldone: 'one'}, {fieldtwo: 'two'}]}`
  * `{or: [{fieldone: 'one'}, {fieldtwo: 'two'}]}`
  */
-export interface Where {
-  and?: Where[]; // AND
-  or?: Where[]; // OR
-  [property: string]: Condition | any; // Other criteria
-}
+export type Where<MT = AnyObject> = Condition<MT>;
 
 /**
  * Order by direction
@@ -72,9 +85,7 @@ export type Direction = 'ASC' | 'DESC';
  * Example:
  * `{afieldname: 'ASC'}`
  */
-export interface Order {
-  [property: string]: Direction;
-}
+export type Order<MT = AnyObject> = {[P in keyof MT]: Direction};
 
 /**
  * Selection of fields
@@ -82,9 +93,7 @@ export interface Order {
  * Example:
  * `{afieldname: true}`
  */
-export interface Fields {
-  [property: string]: boolean;
-}
+export type Fields<MT = AnyObject> = {[P in keyof MT | string]?: boolean};
 
 /**
  * Inclusion of related items
@@ -94,23 +103,23 @@ export interface Fields {
  * Example:
  * `{relation: 'aRelationName', scope: {<AFilterObject>}}`
  */
-export interface Inclusion {
+export interface Inclusion<MT = AnyObject> {
   relation: string;
-  scope?: Filter;
+  scope?: Filter<MT>;
 }
 
 /**
  * Query filter object
  */
-export interface Filter {
+export interface Filter<MT = AnyObject> {
   /**
    * The matching criteria
    */
-  where?: Where;
+  where?: Where<MT>;
   /**
    * To include/exclude fields
    */
-  fields?: Fields;
+  fields?: Fields<MT>;
   /**
    * Sorting order for matched entities. Each item should be formatted as
    * `fieldName ASC` or `fieldName DESC`.
@@ -135,10 +144,10 @@ export interface Filter {
   /**
    * To include related objects
    */
-  include?: Inclusion[];
+  include?: Inclusion<MT>[];
 }
 
-export function isFilter(arg: any): arg is Filter {
+export function isFilter<MT>(arg: any): arg is Filter<MT> {
   if (typeof arg === 'object') {
     const filterFields = [
       'where',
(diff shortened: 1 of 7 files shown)


gms1:
signing the CLA was not possible (504 error), will try it again tomorrow

packages/repository/src/query.ts
```diff
@@ -31,21 +31,38 @@ export enum Operators {
...
+};
+
+type ShortHandEqualType = string | number | boolean | Date;
+
+/**
+ * Matching predicate condition
+ */
+export type PredicateCondition<MT> = {
+  [P in keyof MT]?: PredicateComparison<MT[P]> | (MT[P] & ShortHandEqualType)
+} & {and?: never; or?: never};
```
raymondfeng: Does this prevent us from using nesting `and` and `or`?

packages/repository/test/unit/query/query-builder.unit.ts
```diff
@@ -100,10 +110,13 @@ describe('WhereBuilder', () => {
...
-      b: {gt: 2},
-      c: {lt: 2},
-      and: [{x: 'x'}, {y: {gt: 1}}, {a: 1}, {b: 2}],
+      and: [
+        {a: 1, b: {gt: 2}, c: {lt: 2}},
+        {x: 'x'},
+        {y: {gt: 1}},
+        {a: 1},
+        {b: 2},
+      ],
```
raymondfeng: Why do we remove the ability to include multiple causes at top level, including shorthand and `and`?

packages/repository/test/unit/query/query-builder.unit.ts
```diff
@@ -115,20 +128,12 @@ describe('WhereBuilder', () => {
...
-      and: [
-        {
-          a: 1,
-          and: [{x: 'x'}, {y: {gt: 1}}],
-        },
-        {
-          and: [{b: 'b'}, {c: {lt: 1}}],
-        },
-      ],
+      and: [{a: 1}, {x: 'x'}, {y: {gt: 1}}, {b: 'b'}, {c: {lt: 1}}],
```
raymondfeng: Sam question, why do you have to flatten it now?

packages/repository/src/query.ts
```diff
@@ -173,14 +182,26 @@ export function isFilter(arg: any): arg is Filter {
...
+export class WhereBuilder<MT = AnyObject> {
+  where: Where<MT>;
 
-  constructor(w?: Where) {
+  constructor(w?: Where<MT>) {
     this.where = w || {};
   }
 
-  private add(w: Where): this {
+  private add(w: Where<MT>): this {
```
gms1: the logical conditions are handled by 'and' and 'or', so we may want to rename this method to something like 'addPredicate' or 'addComparison',

bajtos:
Hello @gms1, thank you for the pull request! I am happy to see changes allowing the compiler to better check the where conditions 👍 

I am concerned about the effect of the current proposal on the generated `where` objects. As I understand our intentions (going back to LoopBack 1.0 days), top-level key/value entries are always joined using `and` statement. In other words, the following to conditions are equivalent:

```ts
{a: 1, x: 'x', y: {gt: 1}}

{and: [{a: 1}, {x: 'x'}, {y: {gt: 1}}]}
```

IMO, our builder should be using the first version because it's the simpler option.

IIUC, with your changes in place, we will start emitting the second (longer) option - that's a change for worse in my opinion. Is there a way how to preserve the shorter form?

> `{title: 't1', or: [{content: 'Note 2'}, {content: 'Note 3'}]}`

This is an interesting example. I am not familiar with all details of our querying language, but I would fall to the same trap as you did and expect this condition to work as `HERE (title=:1) AND ((content=:2) OR (content=:3))`

To me, this looks like a bug in the current implementation of our SQL connectors. Clearly, the code processing the condition does understand that we have AND and OR operators involved, but does not take into account the precedence of AND and OR operators in SQL.

It makes me wonder - how is this condition handled by our NoSQL connectors like MongoDB?

Personally, I am proposing to address this problem in the following way:
 - Add a test to our test suite executed by all connectors (see [test/basic-querying.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/fcac276955a58c4aabcd2b602af046f45cfc73fe/test/basic-querying.test.js). This test should run a query similar to `{title: 't1', or: [{content: 'Note 2'}, {content: 'Note 3'}]}` and verify that only the expected records were returned back. (You may need to create additional model instances to enable this scenario.)
 - Open a new pull request with this change. Our CI infrastructure will run the new test for all connectors. Check how each connector is dealing with this tricky condition. 
 - Depending on the outcome (the actual behavior of the connectors), we can discuss what's the best way forward.

gms1:
Hello bajtos,

thank you for this comprehensive answer and the clarification of your concerns

> Personally, I am proposing to address this problem in the following way:

I am with you, so I will start soon, however, I am in the next days under heavy use, so please be patient
To be on the safe side, I will test both where-objects with this proposed test.
```
{title: 't1', or: [{content: 'Note 2'}, {content: 'Note 3'}]}
{and: [{title: 't1'}, {or: [{content: 'Note 2'}, {content: 'Note 3'}]}]}
```
With that, we can also see if all the connectors would be able to work with the new form of the where-object

> To me, this looks like a bug in the current implementation of our SQL connectors.

I also think this is a bug, which must be around for a long time. I think the people have stuck to the documentation, which is why this bug has not really been noticed

Please see the samples from the documentation: (https://loopback.io/doc/en/lb3/Where-filter.html#and--or)
```
{and: [{title: 'My Post'}, {content: 'Hello'}]
{or: [{title: 'My Post'}, {content: 'Hello'}]
{or: [
     { and: [{ field1: 'foo' }, { field2: 'bar' }] },
     { field1: 'morefoo' }
   ]
 }
```
> I am concerned about the effect of the current proposal on the generated where objects. As I understand our intentions (going back to LoopBack 1.0 days), top-level key/value entries are always joined using and statement. In other words, the following to conditions are equivalent:

```
{a: 1, x: 'x', y: {gt: 1}}

{and: [{a: 1}, {x: 'x'}, {y: {gt: 1}}]}
```

but this is exactly the same now:

```whereBuilder.eq(a,1).eq(x,'x').gt(y,'1).build()``` would build the first form of the Where-object
```whereBuilder.and([{a: 1}, {x: 'x'}, {y: {gt: 1}}]).build()``` would build the second form of the Where object

and most important, both where-objects can be expressed in a type-safe manner!

The new WhereBuilder will only separate logical conditions from predicates (comparison conditions)
So the expressions that would not be allowed anymore, are expressions like:
```
{a: 1, x: 'x', y: {gt: 1}, and: [{z: 42}]} 
```
The proposed WhereBuilder would instead generate a where object like this:
```
{and: [{a: 1, x: 'x', y: {gt: 1}}, {z: 42}]} 
```

or in the case of 'or', like this:
```
{and: [{a: 1, x: 'x', y: {gt: 1}}], {or: [{z: 42}]}]}
```
I see this "and" in front is really unusual until now

> IIUC, with your changes in place, we will start emitting the second (longer) option - that's a change for worse in my opinion. Is there a way how to preserve the shorter form?

Here I do not understand yet, why this may be a problem, because no matter if my Where object looks like this:

```
{ and: [{ field1: 'foo' }, { field2: 'bar' }]
```
or that
```
{field1: 'foo' , field2: 'bar' }
```
the generated where-clause should mostly look like that:
```
WHERE field1 = :foo AND field2 = :bar
```
apart from the field names and host variable names, these queries are generated by the juggler and/or the SQLConnector base class, but regarding the non-SQL connectors I unfortunately have no experience ... so the additional tests are appropriate in any case

> IMO, our builder should be using the first version because it's the simpler option.

We must not even adapt the WhereBuilder output at all, we could leave it as it is. The user can still use ```WhereBuilder<Model>``` and will benefit from the type safety, only the generated where-object must then be of type ```Where<AnyObject>``` or  ```juggler.Where```, because casting the output to ```Where<Model>``` would be a very dirty hack

raymondfeng:
@gms1 Sorry for the delay. I have refined your patch to keep it backward compatible and fixed a few bugs. Please review https://github.com/gms1.

PR #1787, opened 2018-10-01
[WIP]: Break cyclic dependencies in relations

bajtos:
Rework HasMany-related code to avoid cyclic dependencies after introducing `belongsTo` relation. This pull request is another step on my journey to land #1618 in several smaller chunks.

Done:

 - [x] Modify `@hasMany` decorator to always accept TypeResolver for the target model (instead of the target model class as before). See the comment below for detailed reasoning why this change is made.

 - [x] Modify HasManyRepository implementation and factory to accept the target model via a Getter (instead of a model class as before).

Todo:
 - [ ] Update repository-json-schema to understand `TypeResolver` properties.
 - [ ] Review the code coverage report and add missing tests

## Checklist


bajtos:
> Modify `@hasMany` decorator to always accept TypeResolver for the target model (instead of the target model class as before).

Before this change, a relation property was decorated as follows:

```ts
@model()
class Customer extends Entity { 
  // ...

  @hasMany(Order)
  orders: Order[];
}
```

My pull request is changing this API as follows:

```ts
@model()
class Customer extends Entity { 
  // ...

  @hasMany(() => Order)
  orders: Order[];
}
```

I am cross-posting https://github.com/strongloop/loopback-next/pull/1618#issuecomment-418113328 below to explain why I am forcing users to always use TypeProvider even in cases where a model constructor would work too.

My argument: While `@belongsTo(User)` may look like a simpler solution, it also means users will have to rework their codebase as soon as they introduce a cyclic relation (e.g. add `@hasMany(Order)`.

IMO, most relations are bidirectional and come in a pair (User has many Orders, thus Order belongs to User. Category has many and belongs to many Products, thus Product has many and belongs to many Categories).

I personally prefer to start with the more flexible solution from the beginning, instead of allowing the users to pick a simpler path that will backfire on them in the next step.

---

> Modify HasManyRepository implementation and factory to accept the target model via a Getter (instead of a model class as before).

This is a similar change.

Before:

```ts
class CustomerRepository extends DefaultCrudRepository<
   Customer,
   typeof Customer.prototype.id
 > {
   public orders: HasManyRepositoryFactory<Order, typeof Customer.prototype.id>;
   constructor(
     @inject('datasources.db') protected db: juggler.DataSource,
     @repository(OrderRepository) orderRepository: OrderRepository,
   ) {
     super(Customer, db);
     this.orders = this._createHasManyRepositoryFactoryFor(
       'orders',
       orderRepository,
     );
   }
```

After:

```ts
class CustomerRepository extends DefaultCrudRepository<
   Customer,
   typeof Customer.prototype.id
 > {
   public orders: HasManyRepositoryFactory<Order, typeof Customer.prototype.id>;
   constructor(
     @inject('datasources.db') protected db: juggler.DataSource,
     @repository.getter(OrderRepository)
     getOrderRepository: Getter<OrderRepository>,
   ) {
     super(Customer, db);
     this.orders = this._createHasManyRepositoryFactoryFor(
       'orders',
       getOrderRepository,
     );
   }
```

My reasoning is similar too, see https://github.com/strongloop/loopback-next/pull/1618#discussion_r212276590:

That way people don't have to upgrade their code from `EntityCrudRepository` to `Getter</*...*/>` at the point when they add a new relation introducing cyclic dependency.

packages/repository/src/decorators/relation.decorator.ts
```diff
@@ -72,41 +70,32 @@ export function hasOne(definition?: Object) {
...
+export function hasMany<T extends Entity>(
+  targetResolver: EntityResolver<T>,
   definition?: Partial<HasManyDefinition>,
 ) {
   // todo(shimks): extract out common logic (such as @property.array) to
   // @relation
-  return function(target: Object, key: string) {
-    property.array(targetModel)(target, key);
-
-    const defaultFkName = camelCase(target.constructor.name + '_id');
```
b-admike: Why do we want to get rid of the default foreign key name generation?

packages/repository/src/model.ts
```diff
@@ -119,6 +119,15 @@ export class ModelDefinition {
     return this;
   }
 
+  /**
+   * Define a new relation.
+   * @param definition The definition of the new relation.
+   */
+  addRelation(definition: RelationMetadata): this {
+    this.relations[definition.name] = definition;
+    return this;
```
b-admike: I'm curious; why do we return the model definition here? wouldn't the first line suffice?

packages/repository/src/repositories/relation.factory.ts
```diff
@@ -35,21 +40,76 @@ export function createHasManyRepositoryFactory<
...
     return new DefaultHasManyEntityCrudRepository<
       Target,
       TargetID,
       EntityCrudRepository<Target, TargetID>
-    >(targetRepository, constraint as DataObject<Target>);
+    >(targetRepositoryGetter, constraint as DataObject<Target>);
   };
 }
+
+export type HasManyResolvedDefinition = HasManyDefinition & {keyTo: string};
```
b-admike: 👍

packages/repository/src/repositories/relation.factory.ts
```diff
@@ -35,21 +40,76 @@ export function createHasManyRepositoryFactory<
...
+ * Resolves given hasMany metadata if target is specified to be a resolver.
+ * Mainly used to infer what the `keyTo` property should be from the target's
+ * belongsTo metadata
+ * @param relationMeta hasMany metadata to resolve
+ */
+export function resolveHasManyMetadata(
+  relationMeta: HasManyDefinition,
+): HasManyResolvedDefinition {
+  if (typeof relationMeta.target !== 'function') {
+    const reason = 'target model is required.';
```
b-admike: I might be mistaken, but isn't this check similar to the one below?

packages/repository/src/repositories/relation.factory.ts
```diff
@@ -35,21 +40,76 @@ export function createHasManyRepositoryFactory<
...
+
+  const sourceModel = relationMeta.source;
+  if (!sourceModel || !sourceModel.modelName) {
+    const reason = 'source model must be defined';
+    throw new Error(invalidDefinition(relationMeta, reason));
+  }
+
+  const targetModel = relationMeta.target();
+  debug(
+    'Resolved model %s from given metadata: %o',
```
b-admike: 👍 for adding these debug statements.

packages/repository/src/repositories/relation.factory.ts
```diff
@@ -35,21 +40,76 @@ export function createHasManyRepositoryFactory<
...
+    targetModel.modelName,
+    targetModel,
+  );
+  const defaultFkName = camelCase(sourceModel.modelName + '_id');
```
b-admike: Okay, disregard my comment above :-)

packages/repository/src/type-resolver.ts
```diff
@@ -10,7 +10,10 @@ import {Class} from './common-types';
  * Intended to be used for cases when the JS engine is unable to fully define
  * a given type (require() loops).
  */
-export type TypeResolver<T extends Object> = () => Class<T>;
+export type TypeResolver<
+  Type extends Object,
+  StaticMembers = Function
```
b-admike: Can you clarify what `StaticMembers = Function` does? (sorry I havent seen it before)

packages/repository/test/integration/repositories/relation.factory.integration.ts
```diff
@@ -178,21 +167,28 @@ describe('HasMany relation', () => {
...
-          keyTo: 'authorId',
-        },
-        reviewsApproved: {
-          type: RelationType.hasMany,
-          keyTo: 'approvedId',
-        },
-      },
-    });
+    })
+      .addRelation({
```
b-admike: Nice refactor here!

packages/repository/test/unit/decorator/relation.decorator.unit.ts
```diff
@@ -147,19 +133,21 @@ describe('getModelRelations', () => {
...
-      @hasMany(Order)
+      @hasMany(() => Order)
       orders: Order[];
     }
 
     const relations = getModelRelations(Customer);
-    expect(relations).to.deepEqual({
+    expect(relations).to.containDeep({
       accessTokens: {
-        keyTo: 'userId',
```
b-admike: Is the reason we don't have `keyTo` here because the class resolution hasn't happened yet or we don't have use for it at this point?

packages/repository/src/model.ts
bajtos: See https://en.wikipedia.org/wiki/Fluent_interface

> A fluent interface is normally implemented by using method chaining to implement method cascading (in languages that do not natively support cascading), concretely by having each method return this (self).

packages/repository/src/type-resolver.ts
bajtos: Let's say we have the following resolver:

```ts
const resolver: TypeResolver<Entity> = () => Customer;
const type = resolver();
```

Before my change, `type` was set to `Class<Entity>`, which is defined as a constructor function with arbitrary static properties.

```ts
export interface Class<T> {
  // new MyClass(...args) ==> T
  new (...args: any[]): T;
  // Other static properties/operations
  [property: string]: any;
}
```

These arbitrary static properties are effectively disabling compiler type checks and preventing VSCode from showing hints for static properties.

Consider the following snippet:

```ts
type.modelName
```

When `type` is just `Class<Entity>`, `modelName` is treated as an arbitrary indexer property of type `any`.

This pull request is making two changes:
1. Adding a `StaticMembers` template argument allowing users of `TypeProvider` to specify what static properties are expected on the resolved type. For example:

    ```ts
    const resolver: TypeResolver<Entity, typeof Entity> = () => Customer;
    const type = resolver();
    type.modelName 
    // ^-- TypeScript knows this is a `string` property
    // and VSCode shows a hint
    ```

2. Providing a default value for the new `StaticMembers` template argument. We use `Function` as the default value, because `Function` is the common denominator of all types (every class is a Function). As a result, the value returned by a type provider has only few static properties defined, e.g. `name: string`.

@b-admike Does it make sense now? I am going to capture this information in a code comment for future readers.

packages/repository/test/unit/decorator/relation.decorator.unit.ts
bajtos: With the current codebase, we could infer `keyTo` inside the `@hasMany` decorator and set it in relation metadata. However, one of the next steps is to change the way how the foreign key is inferred and leverage `@belongsTo` decorator (see https://github.com/strongloop/loopback-next/pull/1618/files#diff-444ac7611514971eb1c8e5790e78035cR135). Inspecting `@belongsTo` properties on the target model is the part that cannot be done from `@hasMany` but must wait until all classes are created and the target type can be resolved.

Setting aside `@belongsTo`, I think it's still a good idea keep `keyTo` optional, because it simplifies relation configuration for people not using decorators but writing the model definition manually via `ModelDefinition` class.

packages/repository/src/repositories/relation.factory.ts
bajtos: Good catch 👍

packages/repository/src/type-resolver.ts
bajtos: What is also important: before my change, you could treat `modelName` as a function, write `model.modelName()`, the compiler will not complain and you will discover the mistake later at runtime. With my change in place, the compiler is able to type-check known static properties and point out that `modelName` is a string that cannot be invoked like a function.

bajtos:
I have addressed the comments and added more tests in 537644e. I am going to proceed with landing this PR as-is, I am happy to address any further feedback in follow-up PRs.

Issue #1361, opened 2018-05-24
[Relation] Support BelongsTo relation


dhmlau:
## Description / Steps to reproduce / Feature proposal
Add support for `BelongsTo` relation between two models. Stretch goal for `DP3`.

### Acceptance Criteria
- [x] Update [factory method](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repositories/relation.factory.ts) to handle this relation
- [x] create interface for CRUD methods available for the relation.  See [this](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repositories/relation.repository.ts) as an example.
    - [x] create default class which implements the interface
- [x] Implement the `belongsTo` decorator
  - [x] model decorator: infers the metadata for the relation
  - [x] repository decorator: return the constrained target crud repository (possible name, `belongsToRepository`)
- [x] Address dual usage of `hasMany` and `belongsTo` for a two way relation definition
- [ ] tests / docs / blog post

dhmlau:
I think we need to have an example of using BelongsTo, though the acceptance criteria does not include it.  Feel free to open a separate task.

Issue #1179, opened 2018-03-23
OpenAPI decorator does not properly generate schemas of type `Partial`


shimks:
## Description / Steps to reproduce / Feature proposal
From `todo.controllers.ts` in `example-getting-started`:
```ts
    async createTodo(@requestBody() todo: Todo) {
    // TODO(bajtos) This should be handled by the framework
    // See https://github.com/strongloop/loopback-next/issues/118
      if (!todo.title) {
        return Promise.reject(new HttpErrors.BadRequest('title is required'));
      }
      return await this.todoRepo.create(todo);
    }
```

It should be noted that the spec generated by `@requestBody` is not correct since the argumnet that the operation takes in _does not_ need an `id` property. One may think that `Partial<Todo>` may be the next best thing, but it's not possible at the moment to specify which properties are optional and which aren't.

We should find a way to generate the schema for a model when used as a parameter so that id property is not included in the generated schema

### Proposed behavior:
```ts
@post('/')
makeStuff(@requestBody(Note) note: Partial<Note>) {}
```
- Whenever `requestBody` is used on a model that extends `Entity`, always generate (or modify) its spec without the property that's been assigned as `id`.
- pros:
  - simple ux
- cons:
  - Not very flexible
  - `requestBody` becomes very overloaded

### Acceptance Criteria:
```ts
@post('/exclude')
postWithout(@requestBody(Note, {excludeProperties: ['id']}) note: Partial<Note>) {}
@post('/include')
PostWith(@requestbody(Note, {includeProperties: ['title', 'description']) note: Partial<Note>) {}
```
- [ ] explore more options from TypeScript (like Partial) to see if they can infer or provide us with additional data/metadata
- [ ] allow `requestBody` (or other new decorator to create the appropriate metadata) to take in the model type and schema generation options to produce a custom schema
  - [ ] specifying `excludeProperties` should generate an OpenAPI Schema that lacks the given keys and their values when converting from the JSON Schema generated in `rest`
  - [ ] specifying `includeProperties` should generate an OpenAPI Schema that contains only the given keys and their values when converted from the JSON Schema generated in `rest`
- [ ] Backwards compatibility must be preserved (`@requestBody(spec)`)
      
- related to https://github.com/strongloop/loopback-next/pull/1121

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

bajtos:
[...]

PR #1794, opened 2018-10-02
feat(repository): implement belongsTo relation

bajtos:
Implement missing pieces for belongsTo relation:
 - define a new interface describing belongsTo relation metadata
 - fix the decorator to set all required metadata
 - define an accessor function used in repositories in lie of a repository class 
 - create a factory function for obtaining the accessor function
 - etc.

This pull request is another step on my journey to land #1618 in several smaller chunks.

_This pull request is related to #1361._

## Checklist


bajtos:
@raymondfeng @b-admike I am not able to make the following test pass right now:

```
  1) TodoListApplication
       creates todo for a todoList:
     Error: expected 200 "OK", got 422 "Unprocessable Entity"
```

Source code:

```ts
    const todo = givenTodo({todoListId: undefined});
    const response = await client
      .post(`/todo-lists/${persistedTodoList.id}/todos`)
      .send(todo)
      .expect(200);
```

The situation is as follows:
 - `@belongsTo` decorator marks the foreign key property (`todoListId`) as required
 - The client making the request wants to omit that value and let the server set it from the path parameter.
 - Our REST API is not flexible enough to describe the endpoint as accepting `Todo` body without the required `todoListId` property (this is a well known issue https://github.com/strongloop/loopback-next/issues/1179).

I'd like to ask you for your opinion on how to address this problem. I see two options:
 - Don't mark the foreign key as a required property. Let the repository and/or database server to check and enforce its presence.
 - Modify the failing test to send `todoListId` in the request body. This means that all clients of LB4 REST APIs will have to include the foreign key value when creating related model instance.

Thoughts?

bajtos:
@raymondfeng @b-admike I just realized another problem. Notice that in some tests, we are applying `@belongsTo` decorator on the foreign key property, e.g.

https://github.com/strongloop/loopback-next/blob/d407ea3efdc0c7788a0e9def5d9f6c41012efab6/packages/repository/test/fixtures/models/order.model.ts#L29-L30

but in other places we apply the decorator on the target model getter property, e.g.

https://github.com/strongloop/loopback-next/blob/d407ea3efdc0c7788a0e9def5d9f6c41012efab6/packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts#L101-L105

The second approach have two problems:
 - It is not enough to define `customer` property, `belongsTo` always required the foreign key to be defined on the source model too (e.g. `customerId`).
 - A getter property is introducing a dependency cycle that's not possible to break as long as TypeScript is setting `design:type` metadata. In our test showing `@belongsTo` on ` customer`, we use `ICustomer` instead of `Customer` as the property type.

I think we should unify the code in this pull request to use only one of this approaches; and I am leaning towards decorating the foreign key property.

Thoughts?

b-admike:
> The situation is as follows:
> 
> @belongsTo decorator marks the foreign key property (todoListId) as required
> The client making the request wants to omit that value and let the server set it from the path parameter.
> Our REST API is not flexible enough to describe the endpoint as accepting Todo body without the required todoListId property (this is a well known issue #1179).
> I'd like to ask you for your opinion on how to address this problem. I see two options:
> 
> Don't mark the foreign key as a required property. Let the repository and/or database server to check and enforce its presence.
> Modify the failing test to send todoListId in the request body. This means that all clients of LB4 REST APIs will have to include the foreign key value when creating related model instance.

Thanks for explaining the issue in detail @bajtos. My vote is for not marking the foreign key as a required property and to let the database check its presence. But I guess that may not be true for all the databases we support (not sure if referential integrity is enforced in NoSQL dbs). I am okay with the other approach if there is strong reason to do so, but I see this as a temporary issue since it'll be resolved in #1179.

b-admike:
> @raymondfeng @b-admike I just realized another problem. Notice that in some tests, we are applying `@belongsTo` decorator on the foreign key property, e.g.
> 
> [loopback-next/packages/repository/test/fixtures/models/order.model.ts](https://github.com/strongloop/loopback-next/blob/d407ea3efdc0c7788a0e9def5d9f6c41012efab6/packages/repository/test/fixtures/models/order.model.ts#L29-L30)
> 
> Lines 29 to 30 in [d407ea3](/strongloop/loopback-next/commit/d407ea3efdc0c7788a0e9def5d9f6c41012efab6)
> 
>  @belongsTo(() => Customer) 
>  customerId: number; 
> but in other places we apply the decorator on the target model getter property, e.g.
> 
> [loopback-next/packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts](https://github.com/strongloop/loopback-next/blob/d407ea3efdc0c7788a0e9def5d9f6c41012efab6/packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts#L101-L105)
> 
> Lines 101 to 105 in [d407ea3](/strongloop/loopback-next/commit/d407ea3efdc0c7788a0e9def5d9f6c41012efab6)
> 
>  @belongsTo(() => Customer) 
>  // TypeScript does not allow us to reference Customer here, the compiled 
>  // code throws at runtime when trying to set design:type metadata to Customer 
>  // class that is not defined yet. 
>  customer: ICustomer; 
> The second approach have two problems:
> 
> * It is not enough to define `customer` property, `belongsTo` always required the foreign key to be defined on the source model too (e.g. `customerId`).
> * A getter property is introducing a dependency cycle that's not possible to break as long as TypeScript is setting `design:type` metadata. In our test showing `@belongsTo` on `customer`, we use `ICustomer` instead of `Customer` as the property type.
> 
> I think we should unify the code in this pull request to use only one of this approaches; and I am leaning towards decorating the foreign key property.
> 
> Thoughts?

I was not aware of the places where we use the `@belongsTo` decorator for the target model getter property. I think that might have been an oversight on our part and agree to unify it to be on the foreign key property.

marioestradarosa:
> not sure if referential integrity is enforced in NoSQL dbs

The ref integrity has to be enforced at the client level, the document based DB has no rules for this at the DB Engine level.

bajtos:
> A getter property is introducing a dependency cycle that's not possible to break as long as TypeScript is setting design:type metadata. In our test showing @belongsTo on customer, we use ICustomer instead of Customer as the property type.

I filled a TypeScript issue for this. IMO, the compiler should report compile-time error if it cannot emit code that works. See https://github.com/Microsoft/TypeScript/issues/27519

bajtos:
Thank you for the comments. I am going to make the following changes:
 - Keep the foreign key as an optional property.
 - Move `@belongsTo` decorator to the foreign key property. I am not sure if this is the right design decision, I feel it's inconsistent with `@hasMany` that's decorating a property for accessing the related model instances, but I don't have any better solution right now (especially because of Microsoft/TypeScript#27519).

@marioestradarosa 
> The ref integrity has to be enforced at the client level, the document based DB has no rules for this at the DB Engine level.

When you say "client level", do you mean our ORM layer (`@loopback/repository`) as the database client, or a web/mobile/IoT application acting as the client sending REST requests to an LB4 API server?

Ideally, I'd like LoopBack to enforce referential integrity inside our ORM layer (`@loopback/repository`). That is difficult right now because of #1179.

When a model is created via a relational API, e.g. `customerRepo.order(customerId).create()`, then the foreign key is always set. However, the regular CRUD API (`orderRepo.create()`) is not checking for the presence of the foreign key (because it's an optional property now) and even if the foreign key is set, it does not check whether it is referring to an existing target model instance.

Are you ok with proceeding with the currently proposed solution and add LoopBack-level checks of foreign key values later?

bajtos:
@raymondfeng @b-admike @jannyHou @marioestradarosa the pull request is done and ready for the final review. PTAL.

I am leaving documentation out of scope of this pull request in order to get it lander sooner. I'll open a new PR to document this new relation type.

packages/repository/test/unit/decorator/relation.decorator.unit.ts
```diff
@@ -109,6 +118,85 @@ describe('relation decorator', () => {
...
+        addressBookId: number;
+      }
+      const jugglerMeta = MetadataInspector.getAllPropertyMetadata(
+        MODEL_PROPERTIES_KEY,
+        Address.prototype,
+      );
+      expect(jugglerMeta).to.eql({
+        addressBookId: {
+          type: Number,
+          required: true,
```
b-admike: Is `required: true` set by juggler because this is the only property in the model def?

packages/repository/test/integration/repositories/relation.factory.integration.ts
```diff
@@ -227,3 +151,131 @@ describe('HasMany relation', () => {
...
+
+describe('BelongsTo relation', () => {
+  let findCustomerOfOrder: BelongsToAccessor<
+    Customer,
+    typeof Order.prototype.id
+  >;
+
+  before(givenCrudRepositories);
+  before(givenAccessor);
+  afterEach(async function resetRepositories() {
```
b-admike: Dont we want to clean up the repos in the `before()` handler so that we can inspect the records after if tests fail and to start with clean repository instances?

packages/repository/test/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -0,0 +1,71 @@
...
+  juggler,
+  repository,
+  RepositoryMixin,
+  ApplicationWithRepositories,
+} from '../..';
+import {CustomerRepository, OrderRepository} from '../fixtures/repositories';
+import {expect} from '@loopback/testlab';
+import {Application} from '@loopback/core';
+
+describe('HasMany relation', () => {
```
b-admike: `HasMany relation` -> `belongsTo relation`

packages/repository/src/decorators/relation.decorator.ts
```diff
@@ -112,7 +112,9 @@ export function belongsTo<T extends Entity>(
         decoratedTarget,
         decoratedKey,
       ),
-      required: true,
+      // TODO(bajtos) Make the foreign key required once our REST API layer
+      // allows controller methods to exclude required properties
+      // required: false,
```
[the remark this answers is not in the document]
bajtos: Good catch. I'll keep the commented line to make it clear what is TODO, but change it to `required: true`.

packages/repository/test/unit/decorator/relation.decorator.unit.ts
bajtos: No, `required: true` used to be set by `@belongsTo` decorator before I changed this behavior based on our discussion.

packages/repository/test/integration/repositories/relation.factory.integration.ts
bajtos: Sure! I was copying this code from elsewhere, will fix it.

PR #1636, opened 2018-08-22
chore: add script to build typescript project references

raymondfeng:
Start to add support for TypeScript project references.

## Checklist

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [x] Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

PR #1803, opened 2018-10-04
build: drop multiple dist targets, always use es2017

bajtos:
Based on the work in https://github.com/strongloop/loopback-next/pull/1636, it's clear that we cannot continue building for multiple ES targets (ES2017 for Node.js 8.x and ES2018 for Node.js 10.x) if we want to use TypeScript Project References. (And we do want to!)

This pull request is removing multi-dist build and switching all packages (including project templates) to always build for ES2017 (Node.js 8.x) only.

 - `dist10` is removed completely
 - `dist8` is moved to `dist`
 - `build:dist8`, `build:dist10` and `build:dist-all` scripts are removed in favor of `build:dist`
 - remove all usages of `lb-mocha`'s `DIST` template, use `dist` instead
 - remove all usages of `@loopback/dist-util`, load `./dist` directly. Keep using `@loopback/dist-util` inside CLI - we need the version check
 - fix `compile-package` to correctly treat `--outDir` path as relative to CWD


bajtos:
Grr. Apparently `tsc es2017 --outdir dist` does not work because `dist` is considered as absolute to monorepo root, not to the package that involved `tsc`.

bajtos:
Uff, we have so many places affected by the design decision to support multiple dist targets! I hope that I found them all, unfortunately it makes the change-set huge.

The pull request is ready for review.

PR #1810, opened 2018-10-05
feat: deprecate dist-util package

bajtos:
As a follow-up for  #1803 and #1809, this pull request is removing remaining usages of dist-util package and marking the package as deprecated. My intention is to remove dist-util from loopback-next *before* 4.0 GA.

## Checklist

### DIFF
--- a/packages/cli/bin/cli-main.js
+++ b/packages/cli/bin/cli-main.js
@@ -6,14 +6,18 @@
 
 'use strict';
 
-const checkNodeVersion = require('@loopback/dist-util').checkNodeVersion;
-
 const pkg = require('../package.json');
-try {
-  const range = pkg.engines.node;
-  checkNodeVersion(range);
-} catch (e) {
-  console.error(e.message);
+const semver = require('semver');
+
+// Make sure node version meets the requirement. This code intentionally only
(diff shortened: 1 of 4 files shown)


mightytyphoon:
Hi, I come from : #1817
I think this broke the management of dist folders somehow. I could fix it but now tests are not working.

raymondfeng:
@mightytyphoon I don't this PR is the root cause as it has not been merged yet.

The issue is that some existing projects have left-over `dist8` or `dist10` folders before we switch a single `dist` mode. Now `npm run clean` does not remove such files and they may cause issues and confusions.

Try to run `npx lerna run clean -- dist*` to remove such left-over dirs.

mightytyphoon:
@raymondfeng Hi, thanks for your answer, this was actually a problem with the 0.29.0 version, someone had the same problem here : #1817 but 0.30.0 is running well, tests also. So it's all good. There is just a warning at the end of the tests on a clean cli install : 

`no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.`

But this deals with tslint apparently [some infos there](https://github.com/palantir/tslint/pull/3919)

Thank you for the support.

Regards.

PR #1816, opened 2018-10-06
feat(repository): add type safety for query filter/where

raymondfeng:
Supersedes https://github.com/strongloop/loopback-next/pull/1782

@gms1 Please review.

## Checklist

### DIFF
--- a/packages/repository/src/query.ts
+++ b/packages/repository/src/query.ts
@@ -16,49 +16,115 @@ import * as assert from 'assert';
 /**
  * Operators for where clauses
  */
-export enum Operators {
-  eq, // Equal
-  neq, // Not Equal
-  gt, // >
-  gte, // >=
-  lt, // <
-  lte, // <=
-  inq, // IN
-  between, // BETWEEN [val1, val2]
-  exists,
-  and, // AND
-  or, // OR
-}
+export type Operators =
+  | 'eq' // Equal
+  | 'neq' // Not Equal
+  | 'gt' // >
+  | 'gte' // >=
+  | 'lt' // <
+  | 'lte' // <=
+  | 'inq' // IN
+  | 'between' // BETWEEN [val1, val2]
+  | 'exists'
+  | 'and' // AND
+  | 'or' // OR
+  | 'like' // LIKE
+  | 'nlike' // NOT LIKE
+  | 'ilike' // ILIKE'
+  | 'nilike' // NOT ILIKE
+  | 'regexp'; // REGEXP'
 
 /**
- * Matching condition
+ * Matching predicate comparison
  */
-export interface Condition {
-  eq?: any;
-  neq?: any;
-  gt?: any;
-  gte?: any;
-  lt?: any;
-  lte?: any;
-  inq?: any[];
-  between?: any[];
+export type PredicateComparison<PT> = {
+  eq?: PT;
+  neq?: PT;
+  gt?: PT;
+  gte?: PT;
+  lt?: PT;
+  lte?: PT;
+  inq?: PT[];
+  between?: [PT, PT];
   exists?: boolean;
-  and?: Where[];
-  or?: Where[];
+  like?: PT;
+  nlike?: PT;
+  ilike?: PT;
+  nilike?: PT;
+  regexp?: string | RegExp;
+  // [extendedOperation: string]: any;
+};
+
+/**
+ * Value types for `{propertyName: value}`
+ */
+export type ShortHandEqualType = string | number | boolean | Date;
+
+/**
+ * Key types of a given model, excluding operators
+ */
+export type KeyOf<MT extends object> = Exclude<
+  Extract<keyof MT, string>,
+  Operators
+>;
+
+/**
+ * Condition clause
+ * For example:
+ * ```ts
+ * {
+ *   name: {inq: ['John', 'Mary']},
+ *   status: 'ACTIVE',
+ *   age: {gte: 40}
+ * }
+ * ```
+ */
+export type Condition<MT extends object> = {
+  [P in KeyOf<MT>]?:
+    | PredicateComparison<MT[P]> // {x: {lt: 1}}
+    | (MT[P] & ShortHandEqualType) // {x: 1},
+};
+
+/**
+ * Where clause
+ * For example:
+ * ```ts
+ * {
+ *   name: {inq: ['John', 'Mary']},
+ *   status: 'ACTIVE'
+ *   and: [...],
+ *   or: [...],
+ * }
+ * ```
+ */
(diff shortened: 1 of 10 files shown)


gms1:
@raymondfeng 
I had not checked this variant and would have hoped that it would work that way, but unfortunately the same problem with which I've already messed around also occurs here

```
interface Post {
  id: number;
  title: string;
  author: string;
  likes: number;
}

// as soon as there is a mixture of predicates and logical conditions in the same object, 
// the typescript compiler accepts incorrect data types for the predicates

const t1: Where<Post> = {
  id: 'foo',  // wrong type
  or: [{id: 42}]
};

const t2: Where<Post> = {
  id: 42,
  or: [{id: 'foo'}] // wrong type
};
```

packages/repository/src/query.ts
```diff
@@ -55,11 +89,7 @@ export interface Condition {
  * `{and: [{fieldone: 'one'}, {fieldtwo: 'two'}]}`
  * `{or: [{fieldone: 'one'}, {fieldtwo: 'two'}]}`
  */
-export interface Where {
-  and?: Where[]; // AND
-  or?: Where[]; // OR
-  [property: string]: Condition | any; // Other criteria
-}
+// export type Where<MT = AnyObject> = Condition<MT>;
```
bajtos: Commented out code - please remove or fix.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -220,20 +220,24 @@ export class DefaultCrudRepository<T extends Entity, ID>
     }
   }
 
-  async find(filter?: Filter, options?: Options): Promise<T[]> {
-    const models = await ensurePromise(this.modelClass.find(filter, options));
+  async find(filter?: Filter<T>, options?: Options): Promise<T[]> {
+    const models = await ensurePromise(
+      this.modelClass.find(filter as legacy.Filter, options),
```
bajtos: I am surprised that `Filter<T>` is not compatible with `legacy.Filter`.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -267,8 +271,8 @@ export class DefaultCrudRepository<T extends Entity, ID>
     options?: Options,
   ): Promise<void> {
     const idProp = this.modelClass.definition.idName();
-    const where = {} as Where;
-    where[idProp] = id;
+    const where = {} as Where<T>;
+    (where as AnyObject)[idProp] = id;
```
bajtos: How about:

```ts
Object.assign(where, {[idProp]: id});
```

packages/repository/test/unit/query/query-builder.unit.ts
```diff
@@ -5,6 +5,7 @@
 
 import {expect} from '@loopback/testlab';
 import {FilterBuilder, WhereBuilder, filterTemplate, isFilter} from '../../../';
+import {AnyObject} from '../../..';
```
bajtos: Please join this import statement with the one on the previous line. Ideally, run the task "Organize imports" in VSCode.

packages/repository/test/unit/repositories/constraint-utils.unit.ts
```diff
@@ -7,14 +7,14 @@ import {expect} from '@loopback/testlab';
...
   Filter,
-  Where,
   WhereBuilder,
   constrainFilter,
   constrainWhere,
   constrainDataObject,
   constrainDataObjects,
   Entity,
 } from '../../..';
+import {AnyObject} from '../../../src';
```
bajtos: Please import from `../../..`, not `../../../src`.

Also join this import statement with the one on the previous line. Ideally, run the task "Organize imports" in VSCode.

packages/repository/test/unit/repositories/constraint-utils.unit.ts
```diff
@@ -127,7 +127,7 @@ describe('constraint utility functions', () => {
     return builder.build();
   }
 
-  function whereBuilderHelper(where: Where) {
+  function whereBuilderHelper(where: AnyObject) {
```
bajtos: Can we use a stricter type here? E.g.

```ts
function whereBuilderHelper<T>(where: Where<T>) {
  // ...
}
```

raymondfeng:
@gms1 I have the types to catch errors for your cases. PTAL.

gms1:
wow, great, somehow ```& Partial <AndClause <MT >> & Partial <OrClause <MT >>``` did the trick.
Do we want to exclude all Operators by 'KeyOf' or only the logical operators 'and' and 'or'?

packages/repository/src/query.ts
```diff
@@ -135,11 +197,17 @@ export interface Filter {
...
-export function isFilter(arg: any): arg is Filter {
-  if (typeof arg === 'object') {
+/**
+ * TypeGuard for Filter
+ * @param candidate
+ */
+export function isFilter<MT extends object>(
+  candidate: any,
+): candidate is Filter<MT> {
+  if (typeof candidate === 'object') {
```
bajtos: Let's use reverse logic + early return to keep the indentation level smaller.

```ts
if (typeof candidate !== 'object') return false;
```

packages/repository/src/query.ts
```diff
@@ -194,91 +265,120 @@ export class WhereBuilder {
...
   /**
    * Add an `=` condition
    * @param key Property name
    * @param val Property value
    */
-  eq(key: string, val: any): this {
-    return this.add({[key]: val});
+  eq<K extends KeyOf<MT>>(key: K, val: MT[K]): this {
+    const w: Where<MT> = {};
+    w[key] = val;
```
bajtos: I think this can be simplified as follows - see the original code which was already using this simpler form!

```ts
return this.add(w, {[key]: val});
```

The same comment applies to builder methods below too.

packages/repository/test/unit/query/query-builder.unit.ts
```diff
@@ -139,10 +147,44 @@ describe('WhereBuilder', () => {
...
+      likes: number;
+      gte: number;
+      regexp: unknown;
+    }
+
+    type Key = KeyOf<Post>;
+
+    const k1: Key = 'id';
+    // const k2: Key = 'regexp'; // fail to compile, regexp is a keyword
+    // const k3: Key = 'gte'; // fail to compile, gte is a keyword
```
b-admike: I guess we don't really have a way to "catch" compiler errors, but if these were to be uncommented, we can see them right?

raymondfeng: Yes

PR #1824, opened 2018-10-08
refactor: let TypeScript do resource copying

bajtos:
The first commit enables TypeScript flag `resolveJsonModule` and leverages this feature to load datasource configurations in a type-safe way using regular "import" statements. As a result, the TypeScript compiler is taking care of copying datasource JSON files to `dist` now.

The second commit reorganizes test fixtures so that the build step no longer needs to copy resources to `dist`.

The third commit renames `lb-tsc` flag `--ignore-resources` to `--copy-resources`, changing the behavior of `lb-tsc` from "copy by default" to "copy only if asked to". This is a breaking change affecting existing LB4 projects. After this change is published, new projects are going to be scaffolded in such way that they don't need resource copying.

I have two goals in mind:
 - Reduce the amount of work needed to switch our monorepo to TypeScript project references (see https://github.com/strongloop/loopback-next/pull/1636).
 - Speed up our build (especially in watch mode) by avoiding unnecessary build steps (like resource copying).

## Checklist


docs/site/Repositories.md
```diff
@@ -326,7 +326,7 @@ context.
 ```ts
 import {inject} from '@loopback/core';
 import {juggler, AnyObject} from '@loopback/repository';
-const config = require('./redis.datasource.json');
+import * as config from './redis.datasource.json';
```
raymondfeng: +1 to use TS import for json files.

packages/http-server/test/integration/http-server.integration.ts
```diff
@@ -211,13 +211,14 @@ describe('HttpServer (integration)', () => {
     host?: string;
   }): HttpServer {
     const options: HttpServerOptions = {protocol: 'https', host};
+    const certDir = path.resolve(__dirname, '../../../fixtures');
```
raymondfeng: This is what annoys me.The directory structure is different between source code and the distribution. In this case, developers can be confused as they might assume `../../../fixtures` resolve to `packages/fixtures` based on the source code directory chain.

marioestradarosa:
👍  for `If we decide to follow your approach where __tests__ are nested under src, the problem goes away too, `

Issue #1797, opened 2018-10-03
Missing doc about form-data parameters in controller


JesusTheHun:
In the documentation about route parameters ( https://loopback.io/doc/en/lb4/Routes.html#parameters ) it is mentioned "form Data". However, no where we can see a use of it. And right now, I'm unable to catch any submitted data from the form.

In my case, in the sequence, the call of `this.parseParams(request, route)` return an array of undefined values. So obviously I'm missing something, and document it would be great !

Exemple : 

```typescript
export class RegistrationController {

  // stuff 

  @post('/register', {
    responses: {
      '200': {
        description: 'Registration information page',
        content: {'text/html': {schema: {type: 'string'}}},
      },
    },
  })
  create(
    @param.query.string('email') email: string,
    @param.query.string('password') password: string,
  ): Promise<string> {
    // email & password are empty
  }
}
```

JesusTheHun:
Hi @marioestradarosa , have you been able to perform some tests ?

raymondfeng:
You probably should use `@requestBody` to decorate the body parameter and use `body.email` and `body.password`. Something like:

```ts
create(@requestBody({
    content: {
      'application/x-www-form-urlencoded': {
        schema: {
          type: 'object',
          properties: {
            email: {
              description: 'email',
              type: 'string',
            },
            password: {
              description: 'password',
              type: 'string',
            },
          },
        },
      },
    },
  })
  body: {
    email: string;
    password: string;
  }) {
    console.log(body.email);
    return 'good';
  }
```

dhmlau:
Moving it to `In Progress` because of PR #1838.

dhmlau:
Closing as done, because PR #1838 has landed.

Issue #1452, opened 2018-06-21
[EPIC] REST layer improvements (post-GA)


bajtos:
- [ ] configurable response types/formats #436
- [ ] serve OpenAPI spec from a controller method
- [ ] etag-based caching
- [ ] Allow controller methods to specify custom status code and contribute additional response headers
- [x] streaming requests
- [ ] streaming responses #2230
- [ ] handling file uploads https://github.com/strongloop/loopback-next/issues/1873
- etc.

PR #1838, opened 2018-10-11
feat(rest): add support for form request body

raymondfeng:
See https://github.com/strongloop/loopback-next/issues/1797

## Checklist

### DIFF
--- a/docs/site/Parsing-requests.md
+++ b/docs/site/Parsing-requests.md
@@ -179,6 +179,77 @@ in/by the `@requestBody` decorator. Please refer to the documentation on
 [@requestBody decorator](Decorators.md#requestbody-decorator) to get a
 comprehensive idea of defining custom validation rules for your models.
 
+We support `json` and `urlencoded` content types. The client should set
+`Content-Type` http header to `application/json` or
+`application/x-www-form-urlencoded`. Its value is matched against the list of
+media types defined in the `requestBody.content` object of the OpenAPI operation
+spec. If no matching media types is found or the type is not supported yet, an
+UnsupportedMediaTypeError (http statusCode 415) will be reported.
+
+Please note that `urlencoded` media type does not support data typing. For
+example, `key=3` is parsed as `{key: '3'}`. The raw result is then coerced by
+AJV based on the matching content schema. The coercion rules are described in
+[AJV type coercion rules](https://github.com/epoberezkin/ajv/blob/master/COERCION.md).
+
+The [qs](https://github.com/ljharb/qs) is used to parse complex strings. For
+example, given the following request body definition:
+
+```ts
+const requestBodyObject = {
+  description: 'data',
+  content: {
+    'application/x-www-form-urlencoded': {
+      schema: {
+        type: 'object',
+        properties: {
+          name: {type: 'string'},
+          location: {
+            type: 'object',
+            properties: {
+              lat: {type: 'number'},
+              lng: {type: 'number'},
+            },
+          },
+          tags: {
+            type: 'array',
+            items: {type: 'string'},
+          },
+        },
+      },
+    },
+  },
+};
+```
+
+The encoded value
+`'name=IBM%20HQ&location[lat]=0.741895&location[lng]=-73.989308&tags[0]=IT&tags[1]=NY'`
+is parsed and coerced as:
+
+```ts
+{
+  name: 'IBM HQ',
+  location: {lat: 0.741895, lng: -73.989308},
+  tags: ['IT', 'NY'],
+}
+```
+
+The request body parser options (such as `limit`) can now be configured by
+binding the value to `RestBindings.REQUEST_BODY_PARSER_OPTIONS`
+('rest.requestBodyParserOptions'). For example,
+
+```ts
+server
+  .bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS)
+  .to({limit: 4 * 1024 * 1024}); // Set limit to 4MB
+```
+
+The list of options can be found in the [body](https://github.com/Raynos/body)
+module.
+
+By default, the `limit` is `1024 * 1024` (1MB). Any request with a body length
+exceeding the limit will be rejected with http status code 413 (request entity
+too large).
+
 A few tips worth mentioning:
 
 - If a model property's type refers to another model, make sure it is also
--- a/packages/openapi-v3/src/decorators/request-body.decorator.ts
+++ b/packages/openapi-v3/src/decorators/request-body.decorator.ts
@@ -81,7 +81,9 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
   return function(target: Object, member: string, index: number) {
     debug('@requestBody() on %s.%s', target.constructor.name, member);
     debug('  parameter index: %s', index);
-    debug('  options: %s', inspect(requestBodySpec, {depth: null}));
+    /* istanbul ignore if */
+    if (debug.enabled)
+      debug('  options: %s', inspect(requestBodySpec, {depth: null}));
 
     // Use 'application/json' as default content if `requestBody` is undefined
     requestBodySpec = requestBodySpec || {content: {}};
@@ -95,7 +97,9 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
 
     const paramType = paramTypes[index];
     const schema = resolveSchema(paramType);
-    debug('  inferred schema: %s', inspect(schema, {depth: null}));
+    /* istanbul ignore if */
+    if (debug.enabled)
+      debug('  inferred schema: %s', inspect(schema, {depth: null}));
     requestBodySpec.content = _.mapValues(requestBodySpec.content, c => {
       if (!c.schema) {
         c.schema = schema;
@@ -109,7 +113,9 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
       requestBodySpec[REQUEST_BODY_INDEX] = index;
     }
 
-    debug('  final spec: ', inspect(requestBodySpec, {depth: null}));
+    /* istanbul ignore if */
+    if (debug.enabled)
+      debug('  final spec: ', inspect(requestBodySpec, {depth: null}));
     ParameterDecoratorFactory.createDecorator<RequestBodyObject>(
       OAI3Keys.REQUEST_BODY_KEY,
       requestBodySpec as RequestBodyObject,
--- a/packages/rest/src/parser.ts
+++ b/packages/rest/src/parser.ts
@@ -8,6 +8,8 @@ import {
   isReferenceObject,
   OperationObject,
   ParameterObject,
+  ReferenceObject,
+  SchemaObject,
   SchemasObject,
 } from '@loopback/openapi-v3-types';
 import * as debugModule from 'debug';
@@ -19,8 +21,15 @@ import {promisify} from 'util';
 import {coerceParameter} from './coercion/coerce-parameter';
 import {RestHttpErrors} from './index';
 import {ResolvedRoute} from './router/routing-table';
-import {OperationArgs, PathParameterValues, Request} from './types';
+import {
+  OperationArgs,
+  PathParameterValues,
+  Request,
+  RequestBodyParserOptions,
+} from './types';
 import {validateRequestBody} from './validation/request-body.validator';
+import {is} from 'type-is';
+import * as qs from 'qs';
 
 type HttpError = HttpErrors.HttpError;
 
@@ -29,26 +38,30 @@ const debug = debugModule('loopback:rest:parser');
 export const QUERY_NOT_PARSED = {};
 Object.freeze(QUERY_NOT_PARSED);
 
-// tslint:disable-next-line:no-any
-type MaybeBody = any | undefined;
+// tslint:disable:no-any
+export type RequestBody = {
+  value: any | undefined;
+  coercionRequired?: boolean;
+  mediaType?: string;
+  schema?: SchemaObject | ReferenceObject;
+};
 
-const parseJsonBody: (req: IncomingMessage) => Promise<MaybeBody> = promisify(
-  require('body/json'),
-);
+const parseJsonBody: (
+  req: IncomingMessage,
+  options: {},
+) => Promise<any> = promisify(require('body/json'));
+
+const parseFormBody: (
+  req: IncomingMessage,
+  options: {},
+) => Promise<any> = promisify(require('body/form'));
 
 /**
  * Get the content-type header value from the request
  * @param req Http request
  */
 function getContentType(req: Request): string | undefined {
-  const val = req.headers['content-type'];
-  if (typeof val === 'string') {
-    return val;
-  } else if (Array.isArray(val)) {
-    // Assume only one value is present
-    return val[0];
-  }
-  return undefined;
+  return req.get('content-type');
 }
 
 /**
@@ -61,11 +74,12 @@ function getContentType(req: Request): string | undefined {
 export async function parseOperationArgs(
   request: Request,
   route: ResolvedRoute,
+  options: RequestBodyParserOptions = {},
 ): Promise<OperationArgs> {
   debug('Parsing operation arguments for route %s', route.describe());
   const operationSpec = route.spec;
   const pathParams = route.pathParams;
-  const body = await loadRequestBodyIfNeeded(operationSpec, request);
+  const body = await loadRequestBodyIfNeeded(operationSpec, request, options);
   return buildOperationArguments(
     operationSpec,
     request,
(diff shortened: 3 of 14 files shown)


packages/rest/src/parser.ts
```diff
@@ -83,6 +87,18 @@ async function loadRequestBodyIfNeeded(
 
   const contentType = getContentType(request);
   debug('Loading request body with content type %j', contentType);
+
+  if (
+    contentType &&
+    contentType.startsWith('application/x-www-form-urlencoded')
```
marioestradarosa: I was playing around today with this, and I found that there can also be `multipart/form-data`  ?

raymondfeng: `multipart/form-data` is usually used for file upload. See https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.

marioestradarosa: correct, I was trying to upload a file from a form, but I find this very uncommon, as we can have a service on the client to communicate to the rest api.

raymondfeng: We need to add `multipart/form-data` support for sure to enable file uploads.

raymondfeng: The issue was reported by a user who wants to use simple html form. I think `application/x-www-form-urlencoded` is good enough for that.

Supporting `multipart/form-data` would be much more involved. We had `loopback-component-storage` to allow that in the past. We will need to use a module like https://github.com/mscdex/busboy#readme

raymondfeng: See https://github.com/expressjs/body-parser for more details.

marioestradarosa: I agree 100%, thanks @raymondfeng and good luck today with your presentation!.

bajtos: FYI: handling file uploads is tracked by this GH issue: https://github.com/strongloop/loopback-next/issues/1452

bajtos:
Correctly supporting `application/x-www-form-urlencoded` is more difficult than it seams, because this encoding does not preserve any type information. We have been bitten by this in strong-remoting, plus changes made in the coercion algorithm later are usually a breaking change requiring a semver-major version.

Let's make sure we have a good understanding of how different values are treated by our body parser and that we are ok with any limitations discovered.

For example, let's consider the following schema - I am typing it from top of my head, I hope I won't make any mistakes.

```ts
const schema ={
  type: 'object',
  properties: {
    id: {type: 'number'},
    created: {type: 'string', format: 'date-time'},
    tagIds: {type: 'array', items: {type: 'number'}}
  }
};

const spec = givenOperationWithRequestBody({
  description: 'data',
  content: {
    'application/x-www-form-urlencoded': {schema}
  },
});
```

Things to consider:
 - will the value provided by our parser pass AJV validation?
 - the type of `data.id` property - a number or a string?
 - the type of `data.created` property - a string or a `Date` instance?
 - the type of items in `data.tagIds` - strings or numbers?

Please add more tests to demonstrate how our REST layer is going to treat these scenarios.

bajtos:
FWIW, I wrote pretty extensive (possibly too extensive) test suite for strong-remotin 3.0, I think many of the test cases are relevant for loopback-next too and we should incorporate them into our test suite too.

- [jsonform-array.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-array.suite.js)
- [jsonform-boolean.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-boolean.suite.js)
- [jsonform-date.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-date.suite.js)
- [jsonform-geopoint.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-geopoint.suite.js)
- [jsonform-integer.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-integer.suite.js)
- [jsonform-number.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-number.suite.js)
- [jsonform-object-type.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-object-type.suite.js)
- [jsonform-object.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-object.suite.js)
- [jsonform-string.suite.js](https://github.com/strongloop/strong-remoting/blob/5812e0d442d84782aca7a268a85186d7404ffdb4/test/rest-coercion/jsonform-string.suite.js)

JesusTheHun:
1. Yes, definitly.

2. It should not be a change, but a new feature.

Feature v1 :
```typescript
@formData()
myMethod(formData: object)

// or

@formData({
    properties: ['date', 'username']
    // a shorthand for
    properties: [
        {fieldName: 'date', name: 'date'},
        {fieldName: 'username', name: 'username'}
    ]
})
myMethod(date: string, username: string)
```

Feature v2 :
```typescript
// The key 'user' is registred in the application, and binded to the data transformer
// that returns an instance of MyUserModel.
@formData({
    properties: [
        'foo', // shorthand still works
        {fieldName: 'date', type: 'date'},
        {fieldName: 'username', name: 'user', type: 'user'}]
})
myMethod(date: Date, user: MyUserModel)
```

Does that make sense ?

raymondfeng:
@bajtos PTAL

packages/rest/src/parser.ts
```diff
@@ -78,29 +85,47 @@ export async function parseOperationArgs(
...
+  ) {
+    const body = await parseFormBody(request).catch((err: HttpError) => {
+      debug('Cannot parse request body %j', err);
+      err.statusCode = 400;
```
bajtos: This branch is still not covered by any test, see https://coveralls.io/builds/19498795/source?filename=packages%2Frest%2Fsrc%2Fparser.ts

Please add a new test. AFAICT, `body/form` returns error only when the request body is larger than the configured limit (1MB by default) or when an unsupported content-encoding is used. I guess our best option is to explicitly set `Content-Length` header of the outgoing request to a value larger than 1MB.

Relevant source code:
- https://github.com/Raynos/body/blob/master/index.js
- https://github.com/stream-utils/raw-body/blob/master/index.js

packages/rest/src/parser.ts
```diff
@@ -78,29 +85,47 @@ export async function parseOperationArgs(
...
+    });
+    // form parser returns an object with prototype
+    return {
+      value: Object.assign({}, body),
+      coercionRequired: true,
+    };
+  }
+
   if (contentType && !/json/.test(contentType)) {
     throw new HttpErrors.UnsupportedMediaType(
```
bajtos: This branch is not covered by any test, see https://coveralls.io/builds/19498795/source?filename=packages%2Frest%2Fsrc%2Fparser.ts

Please add a new test, for example sending a `text/plain` payload (request body).

raymondfeng:
@bajtos More tests are added to cover the error cases.

There are a few things that we need to fix as follow-ups:
1. Why do we use `body` instead of `body-parser`? 
2. We should allow `text` and `stream` formats too
3. We need to make it configurable for `limit`. At the moment, the body size cannot go beyond 1MB.

bajtos:
@raymondfeng Thank you for addressing my comments.

I really dislike pull requests combining multiple changes that could have been done in standalone pull requests. Such large pull requests are difficult to review and take long to land. It would be much better if this PR stayed focused on supporting `application/x-www-form-urlencoded`.

> Why do we use body instead of body-parser?

`body-parser` is a middleware that does not provide promise API. 

Please note that we are parsing the request body only if the operation is accepting a parameter parsed from the request body, therefore we cannot simply mount `body-parser` on our internal Express app.

Also at the time we implemented minimal body parsing, we were not using Express yet (IIRC).

I believe that both `body-parser` and `body` share a lot of the implementation under the hood, e.g. `raw-body`, thus I think there are not many (if any) benefits from migrating to `body-parser`.

> We should allow text and stream formats too

+1

That's out of scope of this pull request IMO.

> We need to make it configurable for limit. At the moment, the body size cannot go beyond 1MB.

Agreed. Again, this is best left for a follow-up pull request.

packages/rest/src/types.ts
```diff
@@ -77,6 +78,14 @@ export type LogError = (
 ) => void;
 
 // tslint:disable:no-any
+
+/**
+ * Options for request body parsing
+ */
+export type RequestBodyParserOptions = OptionsJson &
```
bajtos: Is it a good idea to mix all content-type-specific options at the top level?

What if a user wants to configure different limits for JSON and Text payloads? Is it a valid requirement to support?

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -272,6 +264,78 @@ describe('HttpHandler', () => {
...
+          error: {
+            statusCode: 413,
+            name: 'PayloadTooLargeError',
+            message: 'request entity too large',
+          },
+        })
+        .catch(err => {
+          // The server side can close the socket before the client
+          // side can send out all data
+          if (err && err.code !== 'EPIPE') throw err;
```
bajtos: Please extract this code into a shared helper function.

Example usage:

```ts
.catch(rethrowUnlessPipeError)
```

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -272,6 +264,78 @@ describe('HttpHandler', () => {
...
+        .expect(415, {
+          error: {
+            message: 'Content-type application/xml is not supported.',
+            name: 'UnsupportedMediaTypeError',
+            statusCode: 415,
+          },
+        });
+    });
+
+    it('rejects over-limit request form body', () => {
```
bajtos: Uh, I did not anticipate how complex this test will become when I asked to add it. Considering that we didn't have this kind of tests before this pull request (in the current `master`), I am ok to leave them out of this PR.

In case you prefer to keep these two new tests, then please:
 - make it more clear that `EPIPE` is one of the expected outcomes of the test
 - add a check to verify that the controller method/route handler was not invoked

raymondfeng:
@bajtos I move the switch of `body-parser` to a separate PR.

packages/rest/src/parser.ts
```diff
@@ -78,29 +93,58 @@ export async function parseOperationArgs(
...
+  debug('Request body parser options: %j', options);
 
   const contentType = getContentType(request);
   debug('Loading request body with content type %j', contentType);
+
+  if (
+    contentType &&
+    contentType.startsWith('application/x-www-form-urlencoded')
+  ) {
+    const body = await parseFormBody(request, options).catch(
```
bajtos: I find it rather weird that you are mixin `await` and `.catch()` style. Can you rewrite this using try/catch please?

```ts
try {
  const body = await parseFormBody(request, options)
  return { /* ... */ };
} catch (err) {
  // handle the error, use a helper function shared by JSON & form parsers
  throw normalizeBodyParserError(err);
}
```

packages/rest/src/parser.ts
```diff
@@ -78,29 +93,58 @@ export async function parseOperationArgs(
...
       `Content-type ${contentType} is not supported.`,
     );
   }
 
-  return await parseJsonBody(request).catch((err: HttpError) => {
-    debug('Cannot parse request body %j', err);
-    err.statusCode = 400;
-    throw err;
-  });
+  const jsonBody = await parseJsonBody(request, options).catch(
```
bajtos: Ditto, please use `try/catch` instead of `.catch()`.

packages/rest/src/parser.ts
```diff
@@ -78,29 +93,58 @@ export async function parseOperationArgs(
...
+    },
+  );
+  return {value: jsonBody};
 }
 
 function buildOperationArguments(
   operationSpec: OperationObject,
   request: Request,
   pathParams: PathParameterValues,
-  body: MaybeBody,
```
bajtos: Please check if `MaybeBody` still used elsewhere - maybe it's time to remove this type alias entirely?

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -272,7 +264,83 @@ describe('HttpHandler', () => {
...
+        .set('content-type', 'application/json')
+        .send({key: givenLargeRequest()})
+        .expect(413, {
+          error: {
+            message: 'request entity too large',
+            name: 'Error',
+            statusCode: 413,
+          },
+        })
+        .catch(catchEpipeError)
```
bajtos: When I read this line, I have no idea what is the intent. What happens when `EPIPE` error is caught? What happens when a different error was thrown?

Also I believe `E` in `EPIPE` is an abbreviation for `Error` (`EPIPE` = error pipe), thus `EpipeError` feels rather redundant to me ("error pipe error").

I am proposing to rename `catchEpipeError` to `ignorePipeError`.

bajtos:
Besides the comments posted above (I hope they were persisted by GitHub despite the current outage), I'd like to discuss the following point I raised above:

> IMO, LB4 should leverage the information provided in OpenAPI spec and reject requests with content-type not allowed by the spec early on. There is no need to parse the request body (e.g. text or html) when the endpoint is not prepared to accept it.

IIRC, our `@requestBody` decorator describes the OpenAPI spec operation as accepting `application/json` only by default, see here:
https://github.com/strongloop/loopback-next/blob/bf329719db08cacc011aaaf88bf29aad6bed391d/packages/openapi-v3/src/decorators/request-body.decorator.ts#L90

If we are accepting `application/x-www-form-urlencoded` now too, then the decorator should be changed to correctly announce that fact in the OpenAPI spec we are emitting. I think we should also look into ways how to allow app developers to control which content-types they are willing to accept, but that's out of scope of this pull request - let's open a new issue to discuss the options.

---

In the documentation for the new features, please include a short section describing the problem of coercion from urlencoded format that supports string type only into JavaScript types like `number` and `boolean`, explain that we are using AJV coercion under the hood and point readers to the relevant section of AJV docs. ideally point readers to AJV docume

raymondfeng:
@bajtos I believe that I have addressed your comments and added more information to the docs. PTAL.

packages/rest/src/types.ts
```diff
@@ -77,6 +77,16 @@ export type LogError = (
 ) => void;
 
 // tslint:disable:no-any
+
+/**
+ * Options for request body parsing
+ */
+export type RequestBodyParserOptions = {
+  limit?: number;
```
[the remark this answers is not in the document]
raymondfeng: For `body` module, it's `number`.

docs/site/Parsing-requests.md
```diff
@@ -179,6 +179,77 @@ in/by the `@requestBody` decorator. Please refer to the documentation on
...
+          },
+        },
+      },
+    },
+  },
+};
+```
+
+The encoded value
+`'name=IBM%20HQ&location[lat]=0.741895&location[lng]=-73.989308&tags[0]=IT&tags[1]=NY'`
```
bajtos: Great example 👍

packages/rest/src/parser.ts
```diff
@@ -129,9 +223,11 @@ function buildOperationArguments(
   }
 
   debug('Validating request body - value %j', body);
-  validateRequestBody(body, operationSpec.requestBody, globalSchemas);
+  validateRequestBody(body, operationSpec.requestBody, globalSchemas, {
+    coerceTypes: body.coercionRequired,
```
bajtos: Why are we passing `coerceTypes` in options, when `validateRequestBody` can infer the value from `body.coercionRequired`?

packages/rest/src/types.ts
```diff
@@ -77,6 +77,27 @@ export type LogError = (
...
 // tslint:disable:no-any
+
+/**
+ * Options for request body parsing
+ * See https://github.com/Raynos/body
+ */
+export type RequestBodyParserOptions = {
+  /**
+   * The limit of request body size. By default it is 1MB (1024 * 1024). If a
+   * stream contains more then 1MB, it returns an error. This prevents someone
```
bajtos: typo: `more then`. should be `more than`.

packages/openapi-v3/src/decorators/request-body.decorator.ts
```diff
@@ -81,7 +81,8 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
   return function(target: Object, member: string, index: number) {
     debug('@requestBody() on %s.%s', target.constructor.name, member);
     debug('  parameter index: %s', index);
-    debug('  options: %s', inspect(requestBodySpec, {depth: null}));
+    if (debug.enabled)
```
bajtos: Can you add a comment please to tell our code coverage tool to ignore the fact that this block is intentionally skipped by tests and thus should be not be counted towards code coverage?

packages/openapi-v3/src/decorators/request-body.decorator.ts
```diff
@@ -95,7 +96,8 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
 
     const paramType = paramTypes[index];
     const schema = resolveSchema(paramType);
-    debug('  inferred schema: %s', inspect(schema, {depth: null}));
+    if (debug.enabled)
```
bajtos: Can you add a comment please to tell our code coverage tool to ignore the fact that this block is intentionally skipped by tests and thus should be not be counted towards code coverage?

packages/openapi-v3/src/decorators/request-body.decorator.ts
```diff
@@ -109,7 +111,8 @@ export function requestBody(requestBodySpec?: Partial<RequestBodyObject>) {
       requestBodySpec[REQUEST_BODY_INDEX] = index;
     }
 
-    debug('  final spec: ', inspect(requestBodySpec, {depth: null}));
+    if (debug.enabled)
```
bajtos: Can you add a comment please to tell our code coverage tool to ignore the fact that this block is intentionally skipped by tests and thus should be not be counted towards code coverage?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -46,24 +48,13 @@ export function validateRequestBody(
     throw err;
   }
 
-  const schema = getRequestBodySchema(requestBodySpec);
-  debug('Request body schema: %j', util.inspect(schema, {depth: null}));
+  const schema = body.schema;
+  if (debug.enabled) {
```
bajtos: Can you add a comment please to tell our code coverage tool to ignore the fact that this block is intentionally skipped by tests and thus should be not be counted towards code coverage?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -73,10 +64,12 @@ function getRequestBodySchema(
 function convertToJsonSchema(openapiSchema: SchemaObject) {
   const jsonSchema = toJsonSchema(openapiSchema);
   delete jsonSchema['$schema'];
-  debug(
-    'Converted OpenAPI schema to JSON schema: %s',
-    util.inspect(jsonSchema, {depth: null}),
-  );
+  if (debug.enabled) {
```
bajtos: Can you add a comment please to tell our code coverage tool to ignore the fact that this block is intentionally skipped by tests and thus should be not be counted towards code coverage?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -111,7 +105,9 @@ function validateValueAgainstSchema(
 
   const validationErrors = validate.errors;
 
-  debug('Invalid request body: %s', util.inspect(validationErrors));
+  if (debug.enabled) {
```
bajtos: ditto, let's disable code coverage for this block please

bajtos:
> >  The request body content Notice that requestBody > content describes only application/json type and does not allow urlencoded.
> We're going to have long list of media types supported with extensions. Defaulting to only application/json makes sense to me. The requestBody content can always be customized via `@requestBody`.

I am not sure I fully understand what you are proposing.

Here is my opinion:
 - Any endpoint should be accepting only content types described in its OpenAPI spec.
 - If the spec says we accept JSON only, then urlencoded bodies should be rejected.
 - I don't have a strong opinion on whether urlencoded should be enabled by default or not. 

Does this match your vision too?

If yes, then I would like this pull request to include the following tests (some of them may already exist, IDK):
 - Verify that when the request uses a content-type not described in request-body spec, then the request is rejected.
 - Verify what content types are described by `@requestBody` decorator in the spec it emits.

If no, then let's discuss more.

raymondfeng:
@bajtos We are on the same page.

> Verify that when the request uses a content-type not described in request-body spec, then the request is rejected.

We have a test to cover that.

> Verify what content types are described by @requestBody decorator in the spec it emits.

Added.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -128,6 +128,7 @@ function validateValueAgainstSchema(
 function createValidator(
   schema: SchemaObject,
   globalSchemas?: SchemasObject,
+  options?: AJV.Options,
```
jannyHou: I think you created a type `RequestBodyValidationOptions` to be consistent.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -92,15 +88,16 @@ const compiledSchemaCache = new WeakMap();
 function validateValueAgainstSchema(
   // tslint:disable-next-line:no-any
   body: any,
-  schema: SchemaObject,
+  schema: SchemaObject | ReferenceObject,
```
jannyHou: Hmm, any reason remove the `ReferenceObject` here?

packages/rest/src/parser.ts
```diff
@@ -75,32 +89,112 @@ export async function parseOperationArgs(
...
+    try {
+      const jsonBody = await parseJsonBody(request, options);
+      requestBody.value = jsonBody;
+      return requestBody;
+    } catch (err) {
+      throw normalizeParsingError(err);
+    }
+  }
+
+  throw new HttpErrors.UnsupportedMediaType(
```
jannyHou: There is a file to organize the HttpErrors we defined: https://github.com/strongloop/loopback-next/blob/a478e567daaf732098ca2c1b9994a479146208fe/packages/rest/src/rest-http-error.ts
Defining errors there makes error messages re-usable, e.g. assert the error and message in test.

packages/rest/src/validation/request-body.validator.ts
raymondfeng: It's adding `ReferenceObject` back.

packages/rest/test/unit/parser.unit.ts
```diff
@@ -58,6 +59,223 @@ describe('operationArgsParser', () => {
     expect(args).to.eql([{key: 'value'}]);
   });
 
+  it('parses body parameter for urlencoded', async () => {
```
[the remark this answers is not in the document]
raymondfeng: I think we do.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -194,6 +194,44 @@ describe('RestServer.getApiSpec()', () => {
     });
   });
 
+  it('emits all media types for request body', () => {
+    const opSepc = anOperationSpec()
```
bajtos: Typo: `opSepc`, should be `opSpec`. Since this is the spec we are expecting to be created, and not the spec used to define the API, please rename the variable to `expectedOpSpec` or `expectedSpec` to make that distinction clear.

JesusTheHun:
Congratulations gentlemen ! I'll get my hands on it as soon as the doc is online and give you a feedback

Issue #1785, opened 2018-10-01
Invoke serve-static after API router


bajtos:
In #1611, we added API allowing app developers to serve static assets from their applications. To merge the pull request sooner, we have intentionally accepted a design with suboptimal performance:
 - All serve-static middleware is mounted before API routes, violating the best practices.
 - Users are forbidden to mount static content at root `/` to avoid the biggest performance penalty.

Let's rethink the way how static files are served, follow the best practices and get the best performance.

See also #559

## Acceptance criteria

- [ ] `serve-static` middleware is invoked only after no other route (an API endpoint, an Express-like route) matched the request
- [ ] applications can mount static files at root `/`

bajtos:
In https://github.com/strongloop/loopback-next/pull/1611#issuecomment-415677600 and https://github.com/strongloop/loopback-next/issues/1784, I proposed the following solution for consideration:

Can we define a special kind of a catch-all route that will:
 - match any URL that did not match any API endpoint; i.e. the sequence action `findRoute` returns this catch-all route if no API endpoint matched the requested URL
 - run the express Router where static assets were mounted; i.e. the sequence action `invoke` runs express routing to handle static assets
 - throw `HttpError.NotFound` when no static asset was found, i.e. the sequence action `invoke` throws the 404 error or the express Router for static assets called `next()`.

In my mind, such solution would solve most if not all issues:
 - static middleware is invoked after any routes
 - static middleware is invoked as part of the sequence
 - 404 errors are still handled by the sequence

hacksparrow:
Fixed via https://github.com/strongloop/loopback-next/pull/1848

Issue #1784, opened 2018-10-01
Error handling for non-API routes


bajtos:
In #1611, we added API allowing app developers to serve static assets from their applications. The implementation was focused on enabling the happy path (serve a file that exists and can be fully read) but neglected error handling.

When a static asset is not found and/or cannot not be read, the `reject` error handler is not invoked and a very bare-bone fallback handler is used instead. 

https://github.com/strongloop/loopback-next/blob/0364b59409ce3691babf5f17b5cd444b4e7f72aa/packages/rest/src/rest.server.ts#L654-L666

Let's find a way that enables LB4 app developers to control the error responses for static assets (and any non-API routes in general).

bajtos:
I have already spent some time thinking about this problem, here are few solutions for consideration.

#### 1. Implement static middleware as a virtual route

https://github.com/strongloop/loopback-next/pull/1611#issuecomment-415677600

In LB4, API requests are handled by a Sequence, which is designed with the assumption that the Sequence is going to handle all requests and implement error handling too (including 404 not found).

IMO, removing this assumption and changing the design of Sequence to add a third possible outcome - let somebody else handle the request - would be a big step backwards and we should avoid that.

Also mounting serve-static middleware at Express layer is bypassing the sequence entirely. This is likely to have unintended consequences that may bite us hard. For example, happens serve-static cannot read a file, e.g. because of a permission problem, then it forward the error via `next(err)` for the error-handling middleware to take care of that (see [here](https://github.com/expressjs/serve-static/blob/f287bd6c26ad2bfd0422c533b0358f2f4b16f7db/index.js#L115-L119)). The problem is that we don't have a way how to invoke the `reject` sequence action to handle that error! In fact, the sequence may not even use `reject` and can handle errors differently.

I am wondering: Can we define a special kind of a catch-all route that will:
 - match any URL that did not match any API endpoint; i.e. the sequence action `findRoute` returns this catch-all route if no API endpoint matched the requested URL
 - run the express Router where static assets were mounted; i.e. the sequence action `invoke` runs express routing to handle static assets
 - throw `HttpError.NotFound` when no static asset was found, i.e. the sequence action `invoke` throws the 404 error or the express Router for static assets called `next()`.

In my mind, such solution would solve most if not all issues:
 - static middleware is invoked after any routes
 - static middleware is invoked as part of the sequence
 - 404 errors are still handled by the sequence

#### 2. Move `reject` from Sequence to Server

[...]

PR #1848, opened 2018-10-15
fix: optimize serving static files

hacksparrow:
Optimize serving static files

See #1784 and #1785

## Checklist

### DIFF
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -18,7 +18,6 @@ import * as express from 'express';
 import {PathParams} from 'express-serve-static-core';
 import {IncomingMessage, ServerResponse} from 'http';
 import {safeDump} from 'js-yaml';
-import * as pathToRegExp from 'path-to-regexp';
 import {ServeStaticOptions} from 'serve-static';
 import {HttpHandler} from './http-handler';
 import {RestBindings} from './keys';
@@ -33,7 +32,7 @@ import {
   Route,
   RouteEntry,
   RoutingTable,
-  StaticRoute,
+  StaticAssetsRoute,
 } from './router/routing-table';
 
 import {DefaultSequence, SequenceFunction, SequenceHandler} from './sequence';
@@ -47,7 +46,6 @@ import {
   Send,
 } from './types';
 import {ServerOptions} from 'https';
-import * as HttpErrors from 'http-errors';
 
 const debug = require('debug')('loopback:rest:server');
 
@@ -58,6 +56,10 @@ export type HttpRequestListener = (
 
 export interface HttpServerLike {
   requestHandler: HttpRequestListener;
+}
+
+interface ExpressRouter extends express.Router {
+  handle: express.RequestHandler;
 }
 
 const SequenceActions = RestBindings.SequenceActions;
@@ -129,7 +131,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
   protected _httpServer: HttpServer | undefined;
 
   protected _expressApp: express.Application;
-  protected _routerForStaticAssets: express.Router;
+  protected _routerForStaticAssets: ExpressRouter;
 
   get listening(): boolean {
     return this._httpServer ? this._httpServer.listening : false;
@@ -192,25 +194,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._setupRequestHandler();
 
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
-
-    // LB4's static assets serving router
-    const staticAssetsRouter = new StaticRoute(
-      (req: Request, res: Response) => {
-        return new Promise((resolve, reject) => {
-          const onFinished = () => resolve();
-          res.once('finish', onFinished);
-          this._routerForStaticAssets.handle(req, res, (err: Error) => {
-            if (err) {
-              return reject(err);
-            }
-            // Express router called next, which means no route was matched
-            return reject(new HttpErrors.NotFound());
-          });
-        });
-      },
-    );
-
-    this.httpHandler.registerRoute(staticAssetsRouter);
   }
 
   protected _setupRequestHandler() {
@@ -238,9 +221,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
     };
     this._expressApp.use(cors(corsOptions));
 
-    // Place the assets router here before controllers
-    this._setupRouterForStaticAssets();
-
     // Set up endpoints for OpenAPI spec/ui
     this._setupOpenApiSpecEndpoints();
 
@@ -255,6 +235,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
         this._onUnhandledError(req, res, err);
       },
     );
+
+    // LB4's static assets serving router
+    this._setupRouterForStaticAssets();
   }
 
   /**
@@ -263,7 +246,13 @@ export class RestServer extends Context implements Server, HttpServerLike {
    */
   protected _setupRouterForStaticAssets() {
     if (!this._routerForStaticAssets) {
-      this._routerForStaticAssets = express.Router();
+      this._routerForStaticAssets = express.Router() as ExpressRouter;
+
+      const staticAssetsRouter = new StaticAssetsRoute(
+        this._routerForStaticAssets,
+      );
+
+      this.route(staticAssetsRouter);
     }
   }
 
@@ -646,7 +635,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * See https://expressjs.com/en/4x/api.html#express.static
    * @param path The path(s) to serve the asset.
    * See examples at https://expressjs.com/en/4x/api.html#path-examples
-   * To avoid performance penalty, `/` is not allowed for now.
    * @param rootDir The root directory from which to serve static assets
    * @param options Options for serve-static
    */
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -33,6 +33,7 @@ import {
   Route,
   RouteEntry,
   RoutingTable,
+  StaticRoute,
 } from './router/routing-table';
 
 import {DefaultSequence, SequenceFunction, SequenceHandler} from './sequence';
@@ -46,6 +47,7 @@ import {
   Send,
 } from './types';
 import {ServerOptions} from 'https';
+import * as HttpErrors from 'http-errors';
 
 const debug = require('debug')('loopback:rest:server');
 
@@ -190,6 +192,25 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._setupRequestHandler();
 
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
+
+    // LB4's static assets serving router
+    const staticAssetsRouter = new StaticRoute(
+      (req: Request, res: Response) => {
+        return new Promise((resolve, reject) => {
+          const onFinished = () => resolve();
+          res.once('finish', onFinished);
+          this._routerForStaticAssets.handle(req, res, (err: Error) => {
+            if (err) {
+              return reject(err);
+            }
+            // Express router called next, which means no route was matched
+            return reject(new HttpErrors.NotFound());
+          });
+        });
+      },
+    );
+
+    this.httpHandler.registerRoute(staticAssetsRouter);
   }
 
   protected _setupRequestHandler() {
@@ -243,7 +264,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
   protected _setupRouterForStaticAssets() {
     if (!this._routerForStaticAssets) {
       this._routerForStaticAssets = express.Router();
-      this._expressApp.use(this._routerForStaticAssets);
     }
   }
 
@@ -631,12 +651,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param options Options for serve-static
    */
   static(path: PathParams, rootDir: string, options?: ServeStaticOptions) {
-    const re = pathToRegExp(path, [], {end: false});
-    if (re.test('/')) {
-      throw new Error(
-        'Static assets cannot be mount to "/" to avoid performance penalty.',
-      );
-    }
     this._routerForStaticAssets.use(path, express.static(rootDir, options));
   }
 
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -14,9 +14,11 @@ import {
 } from '@loopback/openapi-v3-types';
 import {AssertionError} from 'assert';
 import * as cors from 'cors';
+import * as debugFactory from 'debug';
 import * as express from 'express';
 import {PathParams} from 'express-serve-static-core';
 import {IncomingMessage, ServerResponse} from 'http';
+import {ServerOptions} from 'https';
 import {safeDump} from 'js-yaml';
 import {ServeStaticOptions} from 'serve-static';
 import {HttpHandler} from './http-handler';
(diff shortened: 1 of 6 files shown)


packages/rest/src/router/routing-table.ts
```diff
@@ -290,7 +290,7 @@ export class Route extends BaseRoute {
     verb: string,
     path: string,
     public readonly spec: OperationObject,
-    protected readonly _handler: Function,
+    protected readonly _handler: Function, // <-- doesn't this Function have a signature?
```
bajtos: The handler function accepts a list of parameters as described in its OpenAPI specification.

See `invokeHandler` below and our docs: https://loopback.io/doc/en/lb4/Routes.html#using-partial-openapi-spec-fragments

packages/rest/src/rest.server.ts
```diff
@@ -190,6 +191,15 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._setupRequestHandler();
 
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
+
+    // LB4's static assets serving router
+    const staticAssetsRouter = new Route('get', '*', {responses: {}}, () => {
+      console.log('>> staticAssetsRouter HANDLER');
+      // @bajtos How do we access req and res here?
```
bajtos: You cannot access the request/response from a handler route (a `Route` instance). In my comment https://github.com/strongloop/loopback-next/issues/1784#issuecomment-428916153, I was showing how to create a new kind of `RouteEntry` (a new class implementing `RouteEntry` interface). RouteEntry's method `invokeHandler` does have access to the request/response objects.

```ts
  // somewhere in the code setting up RoutingTable
  // the anonymous RouteEntry object should be refactored into a class
  this.routingTable.catchAllRoute = <RouteEntry>{
    verb: 'get',
    path: '*',
    spec: { /* TODO: this route should be excluded from the spec */ },
    updateBindings(requestContext: Context) {
      // no-op
    },
    
    invokeHandler(
      {request, response}: Context,
      args: OperationArgs,
    ): Promise<OperationRetval> {
      return this._handleStaticAssetRequest(request, response);
    }

    describe(): string {
      return "final route to handle static assets";
    }
  };
```

packages/rest/src/router/routing-table.ts
```diff
@@ -238,6 +240,31 @@ export interface ResolvedRoute extends RouteEntry {
...
+  updateBindings(requestContext: Context): void {
+    // no-op
+  }
+
+  async invokeHandler(
+    requestContext: Context,
+    args: OperationArgs,
+  ): Promise<OperationRetval> {
+    const req = await requestContext.get(RestBindings.Http.REQUEST);
+    const res = await requestContext.get(RestBindings.Http.RESPONSE);
```
bajtos: Let's use `context.request` and `context.response`, it should be faster than obtaining the values via Context bindings.

packages/rest/src/router/routing-table.ts
```diff
@@ -238,6 +240,31 @@ export interface ResolvedRoute extends RouteEntry {
...
+    return this._handler(req, res);
```
bajtos: I am proposing to simplify the design here. Since there will be always at most one StaticRoute instance registered with a server, there is no need to make the handler configurable. Let's move the code dealing with `_routerForStaticAssets` here into `invokeHandler`.

Ideally, I'd like to see `_routerForStaticAssets` to be a member property of `StaticRoute`.

packages/rest/src/rest.server.ts
```diff
@@ -190,6 +192,25 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+      (req: Request, res: Response) => {
+        return new Promise((resolve, reject) => {
+          const onFinished = () => resolve();
+          res.once('finish', onFinished);
+          this._routerForStaticAssets.handle(req, res, (err: Error) => {
+            if (err) {
+              return reject(err);
+            }
+            // Express router called next, which means no route was matched
+            return reject(new HttpErrors.NotFound());
```
bajtos: Let's be more verbose here. See how we deal with 404 errors now:

https://github.com/strongloop/loopback-next/blob/a2ab86d3a3c721f28159585ca4e87bd8504476c4/packages/rest/src/router/routing-table.ts#L183-L186

packages/rest/src/rest.server.ts
```diff
@@ -190,6 +192,25 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+          });
+        });
+      },
+    );
+
+    this.httpHandler.registerRoute(staticAssetsRouter);
```
bajtos: IIUC, this is relying on the fact that the code paths invoked here is bypassing path validations and that our routing table will pass the path `*` to Express router without modifications. 

I find that rather brittle, let's find a different solution please.

I am thinking about the following implementation of [`RoutingTable#find`](https://github.com/strongloop/loopback-next/blob/a2ab86d3a3c721f28159585ca4e87bd8504476c4/packages/rest/src/router/routing-table.ts#L173-L188):

```ts
class RoutingTable {
  // ...

  find(request: Request): ResolvedRoute {
    debug('Finding route %s for %s %s', request.method, request.path);

    const found = this._router.find(request);

    if (found) {
      debug('Route matched: %j', found);
      return found;
    }

    debug('No API route found for %s %s, trying to find a static assets', request.method, request.path);
    return this._staticAssetsRoute;
  }
}
```

packages/rest/src/router/routing-table.ts
```diff
@@ -238,6 +240,31 @@ export interface ResolvedRoute extends RouteEntry {
   readonly schemas: SchemasObject;
 }
 
+export class StaticRoute implements RouteEntry {
```
bajtos: Let's call this `StaticAssetsRoute`. The name "static route" can be also interpreted as a route that's static (e.g. does not accept any path parameters).

packages/rest/src/router/routing-table.ts
```diff
@@ -150,6 +157,9 @@ export class RoutingTable {
 
     validateApiPath(route.path);
     this._router.add(route);
+    if (route.path === '*') {
```
raymondfeng: This is a bit hacky to test `path === '*'`. The special route is also added to the underlying router, which might cause unexpected results.

bajtos:
@hacksparrow please let me know when this pull request is ready for another review. Also the title says `[WIP]`, is it still true?

hacksparrow:
@bajtos I will let you now. There is API spec generation side-effect, please consider it WIP till I remove it.

hacksparrow:
@bajtos yes that's what I am referring too. Let me add that feature first then. Breakpoints jumping locations and disappearing is making debugging really hard.

packages/rest/src/router/routing-table.ts
```diff
@@ -150,6 +160,9 @@ export class RoutingTable {
 
     validateApiPath(route.path);
     this._router.add(route);
+    if (route.path === '/*') {
```
hacksparrow: @raymondfeng you had some reservations about this.
> This is a bit hacky to test path === '*'. The special route is also added to the underlying router, which might cause unexpected results.

Can you elaborate the concerns, so that we can start a discussion?

packages/rest/src/rest.server.ts
```diff
@@ -58,6 +58,10 @@ export interface HttpServerLike {
   requestHandler: HttpRequestListener;
 }
 
+interface ExpressRouter extends express.Router {
```
hacksparrow: `ExpressRouter` is used in `rest.server.ts` and `routing-table.ts`, and defined in both the files. What is a good location to export it from to avoid duplication?

Having to extend `express.Router` because its definition file does not include the `handle` method for it. I wonder why.

raymondfeng: Please note that `express.Router` itself is  the `handle` function.

raymondfeng: `src/types.ts` is a good place to define common interfaces

packages/rest/src/router/routing-table.ts
hacksparrow: I am still not able to see what the potential problem is. We run `validateApiPath(route.path)` and `/*` is a valid key to the static assets router. What am I missing?

raymondfeng: It's not helpful to have `/*` in the router which won't match any requests.

packages/rest/src/rest.server.ts
```diff
@@ -242,8 +246,13 @@ export class RestServer extends Context implements Server, HttpServerLike {
    */
   protected _setupRouterForStaticAssets() {
     if (!this._routerForStaticAssets) {
-      this._routerForStaticAssets = express.Router();
-      this._expressApp.use(this._routerForStaticAssets);
+      this._routerForStaticAssets = express.Router() as ExpressRouter;
```
raymondfeng: I don't think this is a valid cast as `express.Router()` returns a function which is a handler.

packages/rest/src/rest.server.ts
bajtos: As @raymondfeng said, `express.Router` is a callable function. It extends `IRouter` which extends `RequestHandler` which is defined as follows:

```ts
export interface RequestHandler {
    // tslint:disable-next-line callable-types (This is extended from and can't extend from a type alias in ts<2.2
    (req: Request, res: Response, next: NextFunction): any;
}
```

packages/rest/src/rest.server.ts
```diff
@@ -127,7 +131,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
   protected _httpServer: HttpServer | undefined;
 
   protected _expressApp: express.Application;
-  protected _routerForStaticAssets: express.Router;
+  protected _routerForStaticAssets: ExpressRouter;
```
bajtos: Let's move the implementation details of dealing with static assets down to `StaticAssetsRoute` please. 

Ideally, I'd like to see `RestServer#static` implemented as a thin wrapper delegating the registration of static-asset middleware to StaticAssetsRoute and/or our `HttpHandler` class.

See also my earlier https://github.com/strongloop/loopback-next/pull/1848#discussion_r225476644

packages/rest/src/router/routing-table.ts
bajtos: +1 on what Raymond said.

>  the static asset route is a special one outside of the Router responsibility. It serves as the last matching candidate.
> I'm fine to have `/*` as the path value, but don't call `this._router.add(route);`.

Let me cross-post my earlier comment https://github.com/strongloop/loopback-next/pull/1848#discussion_r225478536 too:

> IIUC, this is relying on the fact that the code paths invoked here is bypassing path validations and that our routing table will pass the path `*` to Express router without modifications. 
>
> I find that rather brittle, let's find a different solution please.
>
> I am thinking about the following implementation of [`RoutingTable#find`](https://github.com/strongloop/loopback-next/blob/a2ab86d3a3c721f28159585ca4e87bd8504476c4/packages/rest/src/router/routing-table.ts#L173-L188):

```ts
class RoutingTable {
  // ...

  find(request: Request): ResolvedRoute {
    debug('Finding route %s for %s %s', request.method, request.path);

    const found = this._router.find(request);

    if (found) {
      debug('Route matched: %j', found);
      return found;
    }

    debug('No API route found for %s %s, trying to find a static assets', request.method, request.path);
    return this._staticAssetsRoute;
  }
}
```
---

Nitpick: I would personally use just `*` as the path value, or even an empty string, because this is just a placeholder value that should be never used, but this is a detail not worth long discussions - pick whatever makes most sense to you.

packages/rest/src/router/routing-table.ts
```diff
@@ -271,6 +295,50 @@ export abstract class BaseRoute implements RouteEntry {
...
+    responses: {},
+  };
+  private _handler: Function;
+  constructor(private _routerForStaticAssets: ExpressRouter) {
+    this._handler = (request: Request, response: Response) => {
+      return new Promise((resolve, reject) => {
+        const onFinished = () => resolve();
+        response.once('finish', onFinished);
+        this._routerForStaticAssets.handle(request, response, (err: Error) => {
+          if (err) {
```
bajtos: When the router called us back, then please remove `onFinished` listener to avoid resolving/rejecting the promise twice and also to avoid a potential (memory) leak.

packages/rest/src/router/routing-table.ts
```diff
@@ -271,6 +295,50 @@ export abstract class BaseRoute implements RouteEntry {
...
+            );
+          }
+        });
+      });
+    };
+  }
+  updateBindings(requestContext: Context): void {
+    // no-op
+  }
+  async invokeHandler(
```
bajtos: No need to use `async` if you are not calling `await`.

packages/rest/src/router/routing-table.ts
```diff
@@ -271,6 +295,50 @@ export abstract class BaseRoute implements RouteEntry {
...
+    requestContext: RequestContext,
+    args: OperationArgs,
+  ): Promise<OperationRetval> {
+    const {request, response} = requestContext;
+    return this._handler(request, response);
```
bajtos: Have you considered inlining `this._handler` inside `invokeHandler`?

```ts
class StaticAssetsRoute /*...*/ {
//...

  invokeHandler(
    {request, response}: RequestContext,
    args: OperationArgs,
  ): Promise<OperationRetval> {
    return new Promise((resolve, reject) => {
      const onFinished = () => resolve();
      response.once('finish', onFinished);
     // ...
    });
  }

//...
}
```

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -164,7 +147,7 @@ describe('RestServer (integration)', () => {
       .expect(200, 'Hello');
   });
 
-  it('serve static assets if matches before other routes', async () => {
+  it('does not serve static assets even if matches before other routes', async () => {
```
bajtos: I am proposing to rephrase this test name as follows:

```ts
it('gives precedence to API routes over static assets')
```

packages/rest/src/router/routing-table.ts
hacksparrow: It's a `once`, do we still need to remove it?

packages/rest/src/router/routing-table.ts
hacksparrow: Got it now.

packages/rest/src/router/routing-table.ts
bajtos: Yes, because when the static asset cannot be found or a ReadStream cannot be created, then the `finish` event is not fired within the timeframe of this async method.

I guess one could argue that the `finish` method will be eventually fired because the HTTP response will be finished by other parts of the framework eventually, but I find it very brittle and hacky to rely on some other part of our codebase to trigger "finish". See also [Action at a distance](https://en.wikipedia.org/wiki/Action_at_a_distance_%28computer_programming%29) anti-pattern.

packages/rest/src/rest.server.ts
hacksparrow: @bajtos Calling `app.static` after the server has started should still set up the static assets route successfully. I tried to understand what you are suggesting and moved some pieces around, but resulted in `app.static` losing functionality after the app started.

Can you elaborate more on your suggestion, please? What does the outline of `RestServer#static` look like?

bajtos: Let's wait until you have addressed the other comments and then I can try to shuffle the code around myself, I think it will be the fastest option.

hacksparrow: I had addressed all others, except this one. Please refresh to see the changes.

bajtos:
@hacksparrow I pushed a new commit f516b31 to further clean up the implementation, PTAL.

bajtos:
@hacksparrow hmm, the code coverage data suggests that we are missing the following test:

- Create an app with static assets configured. 
- Request an URL that does not exist.
- Verify that 404 was returned. 

(This triggers a different code path than for a pure API server with no static assets.)

Could you please add it as part of this pull request?

hacksparrow:
Will do.

raymondfeng:
@hacksparrow I merged the PR so that we can do a release soon.

PR #1856, opened 2018-10-15
fix(cli): add more exit checks to fail fast

raymondfeng:
Add more checks to make sure CLI fails fast. For example, if `--config` has an invalid value/file, we should abort before further prompting.

## Checklist

### DIFF
--- a/packages/cli/test/integration/generators/model.integration.js
+++ b/packages/cli/test/integration/generators/model.integration.js
@@ -65,7 +65,7 @@ describe('lb4 model integration', () => {
       testUtils
         .executeGenerator(generator)
         .inDir(SANDBOX_PATH, () =>
-          testUtils.givenLBProject(SANDBOX_PATH, {excludeKeyword: true}),
+          testUtils.givenLBProject(SANDBOX_PATH, {excludeKeyword: false}),
         )
         .withArguments('myNewModel --base InvalidModel'),
     ).to.be.rejectedWith(/Model was not found in/);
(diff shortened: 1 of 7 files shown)


raymondfeng:
@b-admike Added tests.

packages/cli/test/integration/lib/base-generator.js
```diff
@@ -66,13 +66,27 @@ module.exports = function(generator) {
         assert.equal(gen.options.name, 'abc');
         assert.equal(gen.options.description, 'Test');
       });
+
+      it('fails fast if --config has invalid value', async () => {
```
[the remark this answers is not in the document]
raymondfeng: This is a base test run by all generators

packages/cli/test/integration/generators/model.integration.js
```diff
@@ -65,7 +65,7 @@ describe('lb4 model integration', () => {
       testUtils
         .executeGenerator(generator)
         .inDir(SANDBOX_PATH, () =>
-          testUtils.givenLBProject(SANDBOX_PATH, {excludeKeyword: true}),
+          testUtils.givenLBProject(SANDBOX_PATH, {excludeKeyword: false}),
```
b-admike: for my knowledge, what does this change entail?

raymondfeng: We now fail fast. With the `excludeKeyword` is set `true` (no `loopback` keyword in package.json), the first error is surfaced. Before this change, the 2nd error overrides the 1st one.

PR #1859, opened 2018-10-16
docs: update LB4 status

dhmlau:
In the [LB4 docs main page](https://loopback.io/doc/en/lb4/), remove the note about LB4 is still in early development and is not yet released.

PR #1862, opened 2018-10-16
Update index.md

Edo78:
It looks like LB4 was officially released so it's time to remove the `It is still in early development and is not yet released.` warning


raymondfeng:
@Edo78 Thank you for catching it. The issue is being fixed by https://github.com/strongloop/loopback-next/pull/1859.

Issue #1422, opened 2018-06-12
[Relation] Supports HasOne relation


dhmlau:
## Description / Steps to reproduce / Feature proposal
Add support for `HasOne` relation between two models.

### Acceptance Criteria
- [x] Update [factory method](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repositories/relation.factory.ts) to handle this relation
- [x] create interface for CRUD methods available for the relation.  See [this](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repositories/relation.repository.ts) as an example.
    - [x] create default class which implements the interface
- [x] Implement the `hasOne` decorator
  - [x] model decorator: infers the metadata
- [x] create test cases
- [x] update docs
- [x] create a follow up issue to demonstrate using multiple relations in a new/existing example

dhmlau:
Cross posting in https://github.com/strongloop/loopback-next/issues/1449#issuecomment-425139538 from @elv1s:
> We have a requirement for hasOne in addition to hasMany and belongsTo

Currently `hasOne` is not part of our Q1 roadmap. So I have 2 questions:
1. @elv1s, what is the timeline you're thinking in terms of having `hasOne` support? 
2. @strongloop/sq-lb-apex @raymondfeng @bajtos, after we have the 2 model relations, do you think it's a relatively small effort for us to get `hasOne` in?

elv1s:
We are developing a POC with LB3 right now. If we move to LB4 we would need `hasOne` relationships early 2019.

b-admike:
Closing as done.

PR #1879, opened 2018-10-19
feat(repository): hasOne relation

b-admike:
First iteration of `hasOne` relation. This PR includes the the repository factory, repository interface and default hasOne repository, and has one decorator implementations (mostly taken from our hasMany implementation since they're very similar in nature). It also includes an acceptance test with a Customer has one Address relation to drive it and a unit test as well. Please note that only the **EDIT** `get` and `create` methods are implemented at the moment and tests covering them, so that we can focus on that and incrementally add patch and delete methods to the hasOne repository interface.

**Todos**
- Add documentation for `hasOne` relation
- Add example that uses `hasOne` relation

Fixes https://github.com/strongloop/loopback-next/issues/1422

## Checklist

### DIFF
--- a/packages/repository/src/relations/has-one/has-one-repository.factory.ts
+++ b/packages/repository/src/relations/has-one/has-one-repository.factory.ts
@@ -1,0 +1,103 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/example-todo
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import * as debugFactory from 'debug';
+import {camelCase} from 'lodash';
+import {DataObject} from '../../common-types';
+import {InvalidRelationError} from '../../errors';
+import {Entity} from '../../model';
+import {EntityCrudRepository} from '../../repositories/repository';
+import {isTypeResolver} from '../../type-resolver';
+import {Getter, HasOneDefinition} from '../relation.types';
+import {DefaultHasOneRepository, HasOneRepository} from './has-one.repository';
+
+const debug = debugFactory('loopback:repository:has-many-repository-factory');
+
+export type HasOneRepositoryFactory<Target extends Entity, ForeignKeyType> = (
+  fkValue: ForeignKeyType,
+) => HasOneRepository<Target>;
+
+/**
+ * Enforces a constraint on a repository based on a relationship contract
+ * between models. For example, if a Customer model is related to an Address model
+ * via a HasOne relation, then, the relational repository returned by the
+ * factory function would be constrained by a Customer model instance's id(s).
+ *
+ * @param relationMeta The relation metadata used to describe the
+ * relationship and determine how to apply the constraint.
+ * @param targetRepo The repository which represents the target model of a
+ * relation attached to a datasource.
+ * @returns The factory function which accepts a foreign key value to constrain
+ * the given target repository
+ */
+export function createHasOneRepositoryFactory<
+  Target extends Entity,
+  TargetID,
+  ForeignKeyType
+>(
+  relationMetadata: HasOneDefinition,
+  targetRepositoryGetter: Getter<EntityCrudRepository<Target, TargetID>>,
+): HasOneRepositoryFactory<Target, ForeignKeyType> {
+  const meta = resolveHasOneMetadata(relationMetadata);
+  debug('Resolved HasOne relation metadata: %o', meta);
+  return function(fkValue: ForeignKeyType) {
+    // tslint:disable-next-line:no-any
+    const constraint: any = {[meta.keyTo]: fkValue};
+    return new DefaultHasOneRepository<
+      Target,
+      TargetID,
+      EntityCrudRepository<Target, TargetID>
+    >(targetRepositoryGetter, constraint as DataObject<Target>);
+  };
+}
+
+type HasOneResolvedDefinition = HasOneDefinition & {keyTo: string};
+
+/**
+ * Resolves given hasMany metadata if target is specified to be a resolver.
+ * Mainly used to infer what the `keyTo` property should be from the target's
+ * belongsTo metadata
+ * @param relationMeta hasMany metadata to resolve
+ */
+function resolveHasOneMetadata(
+  relationMeta: HasOneDefinition,
+): HasOneResolvedDefinition {
+  if (!isTypeResolver(relationMeta.target)) {
+    const reason = 'target must be a type resolver';
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  if (relationMeta.keyTo) {
+    // The explict cast is needed because of a limitation of type inference
+    return relationMeta as HasOneResolvedDefinition;
+  }
+
+  const sourceModel = relationMeta.source;
+  if (!sourceModel || !sourceModel.modelName) {
+    const reason = 'source model must be defined';
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  const targetModel = relationMeta.target();
+  debug(
+    'Resolved model %s from given metadata: %o',
+    targetModel.modelName,
+    targetModel,
+  );
+  const defaultFkName = camelCase(sourceModel.modelName + '_id');
+  const hasDefaultFkProperty =
+    targetModel.definition &&
+    targetModel.definition.properties &&
+    targetModel.definition.properties[defaultFkName];
+
+  if (!hasDefaultFkProperty) {
+    const reason = `target model ${
+      targetModel.name
+    } is missing definition of foreign key ${defaultFkName}`;
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  return Object.assign(relationMeta, {keyTo: defaultFkName});
+}
--- a/packages/repository/src/relations/has-one/has-one-repository.factory.ts
+++ b/packages/repository/src/relations/has-one/has-one-repository.factory.ts
@@ -99,5 +99,12 @@ function resolveHasOneMetadata(
     throw new InvalidRelationError(reason, relationMeta);
   }
 
+  if (
+    !targetModel.definition.properties[defaultFkName].id === true &&
+    !targetModel.definition.properties[defaultFkName].generated === false
+  ) {
+    // throw InvalidRelationError('property must be a generated id field')
+  }
+
   return Object.assign(relationMeta, {keyTo: defaultFkName});
 }
--- a/packages/repository/src/relations/has-one/has-one.repository.ts
+++ b/packages/repository/src/relations/has-one/has-one.repository.ts
@@ -4,13 +4,12 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {Getter} from '@loopback/context';
-import {Count, DataObject, Options} from '../../common-types';
+import {DataObject, Options} from '../../common-types';
 import {Entity} from '../../model';
 import {Filter, Where} from '../../query';
 import {
   constrainDataObject,
   constrainFilter,
-  constrainWhere,
 } from '../../repositories/constraint-utils';
 import {EntityCrudRepository} from '../../repositories/repository';
 
@@ -28,21 +27,17 @@ export interface HasOneRepository<Target extends Entity> {
     targetModelData: DataObject<Target>,
     options?: Options,
   ): Promise<Target>;
-  /**
-   * Find target model instance
-   * @param Filter A filter object for where, order, limit, etc.
-   * @param options Options for the operation
-   * @returns A promise which resolves with the found target instance(s)
-   */
-  find(filter?: Filter<Target>, options?: Options): Promise<Target[]>;
 
   /**
    * Find the only target model instance that belongs to the declaring model.
-   * @param filter Query filter
+   * @param filter Query filter without a Where condition
    * @param options Options for the operations
    * @returns A promise of the target object or null if not found.
    */
-  findOne(filter?: Filter<Target>, options?: Options): Promise<Target | null>;
+  get(
+    filter?: Exclude<Filter<Target>, Where<Target>>,
+    options?: Options,
+  ): Promise<Target | undefined>;
 }
 
 export class DefaultHasOneRepository<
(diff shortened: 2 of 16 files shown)


packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -0,0 +1,97 @@
...
+   */
+  find(filter?: Filter<Target>, options?: Options): Promise<Target[]>;
+
+  /**
+   * Find the only target model instance that belongs to the declaring model.
+   * @param filter Query filter
+   * @param options Options for the operations
+   * @returns A promise of the target object or null if not found.
+   */
+  findOne(filter?: Filter<Target>, options?: Options): Promise<Target | null>;
```
bajtos: Let's replace `find` and `findOne` methods with a single method called `get`.

```ts
get(options?: Options): Promise<Target | undefined>;
```

Since there is always at most one target model instance associated with the source model instance, it feels more natural for me to `get` that single instance than `findOne`. I am also not sure what the `filter` argument would be good for, because I see very little value in filtering a set of at most single item.

On the second thought, maybe we should use `find(option)` instead of `get(options)` to indicate that the method may return `undefined` when the target does not exist, as opposed to throwing an EntityNotFound exception.

I think we should also consider the alternative where `get(options)` actually throws when the target model was not found, because that's what CRUD repository methods `findById`, `patchById`, etc. do.

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -0,0 +1,97 @@
...
+    public constraint: DataObject<TargetEntity>,
+  ) {}
+
+  async create(
+    targetModelData: DataObject<TargetEntity>,
+    options?: Options,
+  ): Promise<TargetEntity> {
+    const targetRepository = await this.getTargetRepository();
+    return targetRepository.create(
+      constrainDataObject(targetModelData, this.constraint),
```
bajtos: I am afraid this is not enough to ensure there is never more than one model linked.

Please add a test that calls `.create()` two times and verifies that the second create is rejected.

This is the main difference between HasOne and HasMany.

packages/repository/src/relations/relation.types.ts
```diff
@@ -70,12 +70,26 @@ export interface BelongsToDefinition extends RelationDefinitionBase {
   keyTo?: string;
 }
 
+export interface HasOneDefinition extends RelationDefinitionBase {
+  type: RelationType.hasOne;
+
+  /**
+   * The foreign key used by the target model.
+   *
+   * E.g. when a Customer has many Order instances, then keyTo is "customerId".
```
bajtos: "has one", not "has many"

packages/repository/src/repositories/repository.ts
```diff
@@ -185,6 +185,14 @@ export interface EntityCrudRepository<T extends Entity, ID>
...
    */
   exists(id: ID, options?: Options): Promise<boolean>;
+
+  /**
+   * Find one matching record. Same as 'find', but limited to one result.
+   * @param filter Query filter
+   * @param options Options for the operations
+   * @returns A promise of the target object or null if not found.
+   */
+  findOne(filter?: Filter<T>, options?: Options): Promise<T | null>;
```
bajtos: We should be using `undefined` instead of `null` because it works better with ES6 default parameters.

bajtos:
> coverage/coveralls — Coverage decreased (-55.1%) to 34.7%

This is scary! I don't think your changes would account for such a huge drop in code coverage numbers, but at the same time we will need to fix this before landing your PR.

packages/repository/src/relations/has-one/has-one.repository.ts
b-admike: Yeah I thought about not needing `filter` argument too, then thought that someone might want to use the `fields` key in `filter` to get the instance with certain properties (maybe not a good use case). 

>I think we should also consider the alternative where get(options) actually throws when the target model was not found, because that's what CRUD repository methods findById, patchById, etc. do.

These methods throw the `EntityNotFound` error which would require an `id` to be present and in this case we won't be given one for the method. For that reason, I think naming it `get(options)` and returning `undefined` sounds like a solid approach IMO.

bajtos: > Yeah I thought about not needing filter argument too, then thought that someone might want to use the fields key in filter to get the instance with certain properties (maybe not a good use case).

That's a good use case. Can we mark `filter` as accepting `Filter` properties without `where`? See `Exclude<T, U>` in [TypeScript's Advanced Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#predefined-conditional-types).

> These methods throw the EntityNotFound error which would require an id to be present

I have run into this problem while encountering `belongsTo` relation too. I think it's a clear sign that we need a way how to improve `EntityNotFoundError` to support a where filter in addition to the id value. Such change would be out of scope of this pull request though.

b-admike:
> > coverage/coveralls — Coverage decreased (-55.1%) to 34.7%
> 
> This is scary! I don't think your changes would account for such a huge drop in code coverage numbers, but at the same time we will need to fix this before landing your PR.

Yeah that is huge indeed 😨. I will take a look at the details.

packages/repository/src/relations/has-one/has-one.repository.ts
b-admike: Great catch; I'm working on this right now, and was wondering, where we should have validation in place to reject subsequent requests. If we do it at this level, then we'd probably have to make a query first to see if there is an existing instance IIUC. Thoughts?

bajtos: Could you please check how is HasOne implemented in LB 3.x? 

https://github.com/strongloop/loopback-datasource-juggler/blob/f7036b2837578348207999d8c31d88d624af4251/lib/relation-definition.js#L1791-L1834

packages/repository/src/relations/has-one/has-one.repository.ts
b-admike: > That's a good use case. Can we mark filter as accepting Filter properties without where? See Exclude<T, U> in TypeScript's Advanced Types.

Yes, this approach sounds good to me 👍 and for `EntityNotFound`, we can create a follow-up task for those enhancements.

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -0,0 +1,93 @@
...
+    // performance here?
+    const found = await targetRepository.find(
+      constrainFilter({}, this.constraint),
+    );
+    if (found.length > 0) {
+      throw new Error(
+        'HasOne relation does not allow creation of more than one target model instance',
+      );
+    } else {
+      return await targetRepository.create(
```
bajtos: It is crucial to leverage an atomic implementation of `findOrCreate` provided by our connectors. The current proposal is prone to race conditions, where to instances of the target "hasOne" model can be created.

Consider the following scenario: the LB4 server is handling two incoming HTTP requests to create a hasOne target and the code is executed in the following way by Node.js runtime:

1. Request 2 arrives, `DefaultHasOneRepository#create` is invoked
2. `targetRepository.find` is called. It's an async function so other stuff happens while the query is in progress.
3. Request 1  arrives, `DefaultHasOneRepository#create` is invoked.
3. `targetRepository.find` is called. It's an async function so other stuff happens while the query is in progress.
4. `targetRepository.find` for Request 1 returns, no model was found. `targertRepository.create` in called.
5. `targetRepository.find` for Request 2 returns, no model was found. `targertRepository.create` in called.
6.  Both `create` requests eventually finish. We have two models that are a target of `hasOne` relation. 💥 💣 💥  

I am proposing to put this pull request on hold and open a new pull request to add `findOrCreate` method to `EntityCrudRepository` (and the implementations).

packages/repository/src/relations/has-one/has-one-repository.factory.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  if (relationMeta.keyTo) {
+    // The explict cast is needed because of a limitation of type inference
+    return relationMeta as HasOneResolvedDefinition;
+  }
+
+  const sourceModel = relationMeta.source;
+  if (!sourceModel || !sourceModel.modelName) {
+    const reason = 'source model must be defined';
```
jannyHou: Is it ok to define `reason` twice in one function?

raymondfeng: Yes, it's block-scoped.

packages/repository/src/relations/has-one/has-one-repository.factory.ts
```diff
@@ -0,0 +1,103 @@
...
+    targetModel.definition.properties[defaultFkName];
+
+  if (!hasDefaultFkProperty) {
+    const reason = `target model ${
+      targetModel.name
+    } is missing definition of foreign key ${defaultFkName}`;
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  return Object.assign(relationMeta, {keyTo: defaultFkName});
```
jannyHou: Do we allow override the default fk in the `relationMeta`? This line implies we don't?

packages/repository/src/relations/has-one/has-one.repository.ts
b-admike: Thank you for explaining the need for an atomic implementation with a great example. I agree, and I agree with doing that first.

b-admike:
@jannyHou Thank you for your valuable feedback. I do intend to refactor the code since these two relations have lots of similarities, but thought it would be better to start off like this and think about ways we can re-use common code.

> Some proposal:
> 
> - Add a configure property in the hasMany relation metadata to specify one-to-one relation.
> - Apply additional check in the hasManyRepository's create method to make sure one source instance only HAS ONE target instance.

I like this proposal. Are you thinking that the `hasOne` decorator set this configuration property and the rest of the flow is the same as `hasMany` relation? The only thing that is separate is the repository interface which would be its own standalone interface which would call on `findOrCreate` for `create` for instance.

packages/repository/src/relations/has-one/has-one-repository.factory.ts
bajtos: Good catch. I believe we should allow users to provide their own `keyTo` value. @b-admike please add a test to ensure this use-case is supported.

bajtos:
> The implementation looks similar to hasMany relation, from the design's perspective, would it be possible to make hasOne a more constrained version of hasMany? IMO essentially hasOne = hasMany + one more constraint in the create method.

Let's explore!

Few constraints to keep in mind:

 - The constraint in `create` must be enforced in an atomic way (e.g. `findOrCreate`)
 - Many `HasOneRepository` methods have different signature from `HasManyRepository`, because there is always at most one model on the other side of the relation. Where has-many relation is loading/updating/deleting possibly multiple target instances, has-one always returns/updates a single instance only.

packages/repository/src/relations/has-one/has-one-repository.factory.ts
bajtos: Actually: this line is executed only when `relationMeta.keyTo` was falsey, see lines 72-75 above.

```ts
  if (relationMeta.keyTo) {
    // The explict cast is needed because of a limitation of type inference
    return relationMeta as HasOneResolvedDefinition;
  }
```

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: I opened a new issue to keep track of "findOrCreate" story, see https://github.com/strongloop/loopback-next/issues/1956

raymondfeng:
Cross-posting from https://github.com/strongloop/loopback-next/issues/1956#issuecomment-436380749

RaphaelDrai:
Hello,
I have tested the "hasOne" feature from the github branch "feat/hasone-relation" in order to evaluate what is working and missing. Below my report:

I simulated a use case where we have a "hasOne" relation connection from a source model (parent) to a target  model (child).
The relation indicates that a source model of a given instance (parent_id) cannot have more than one instance to the target model (child_id).
The feature is working as expected and we get an error if we attempt to create two children bind to a same parent.

**Missing features**:
	When we attempt to create two targets to a same source instance the application generates response and log with following error:
	Server side -> 500 Error: HasOne relation does not allow creation of more than one target model instance
	Client side -> receives response with "statusCode": 500 and "message": "Internal Server Error"
	**I recommend** to remove the server log and replace the current client error as following:
		error code: 400 "Bad request"
		message: source model cannot have two targets
	
	_CRUD operations and features missing_:
		FIND
		DELETE
		PATCH

I am currently preparing a short documentation that details how to code an application example of "hasOne" relation. Please let me know where do you think that I can post it.
I need also to know if the missing features are completed in any given github branch and if not what is the plan for it.
Thanks you,
Raphael

b-admike:
hi @RaphaelDrai thank you for taking the time to test out this branch and compiling your results. I've actually continued my work on the `spike/hasone-relation` branch (see #2005) instead of here since we decided to take the approach from there and it was easier for me to compare the changes there to master. 

> When we attempt to create two targets to a same source instance the application generates response and log with following error:
> Server side -> 500 Error: HasOne relation does not allow creation of more than one target model instance
> Client side -> receives response with "statusCode": 500 and "message": "Internal Server Error"

This should now be a `Duplicate Entry` Error since we use the database to determine the uniqueness of the foreign key for us (it is also the primary key on the target model) and marked with a `409` error code.

> _CRUD operations and features missing_:
> 	FIND
> 	DELETE
> 	PATCH

`find` or `get` is currently there, but `PATCH` and `DELETE` are not. I think we can add those two APIs. I've added a `hasOne` relation to the `todo-list` example to demonstrate the changes, so you can also test with that. 

>I am currently preparing a short documentation that details how to code an application example of "hasOne" relation. Please let me know where do you think that I can post it.

That's awesome. We can possibly use the documentation as a follow-up or part of #2005 for the relation I added to `todo-list`.

b-admike:
Closing this in favour of #2005. Feel free to also copy/link your comment there @RaphaelDrai.

RaphaelDrai:
Hello @b-admike ,
It sounds great, thanks you :-).
I will follow-up the todo-list documentation part that you have added in #2005 case and we will publish the documentation and code example of hasOne relation.
**Two questions please**:
1) Actually hasOne feature is considered as a show stopper in our loopback-4 application. Can you provide an estimation when the missing CRUD operations will be implemented and then merged ?
2) For help, I will be glad to contribute with the hasOne CRUD coding on any required part of the code that it is not covered yet.  If it is possible, can you suggest me what contribution do you need that I can help ?
Kind regards,
Raphael

PR #1888, opened 2018-10-20
refactor(build): honors project mocha.opts file

mrbatista:
After generating a new loopback 4 project with cli generator, a new `mocha.opts` file is created by default. When I run the `npm test` command, all options defined in the `mocha.opts` file should be loaded. This PR loads your default built-in mocha options only if the corresponding parameter `--use-default-options` is added to the `npm test` command.

If this logic is accepted, all affected packages should be modified accordingly. I look forward to knowing from you which of them should I update.

## Checklist

### DIFF
--- a/packages/build/bin/run-mocha.js
+++ b/packages/build/bin/run-mocha.js
@@ -17,25 +17,19 @@ Usage:
 
 function run(argv, options) {
   const utils = require('./utils');
-  const path = require('path');
 
   // Substitute the dist variable with the dist folder
   const dist = utils.getDistribution();
   const mochaOpts = argv.slice(2).map(a => a.replace(/\bDIST\b/g, dist));
 
-  // Add default options
-  if (mochaOpts.indexOf('--opts') === -1) {
-    const optsPath = require.resolve('../mocha.opts');
-    mochaOpts.unshift('--opts', optsPath);
-  }
+  const setMochaOpts =
+    !utils.isOptionSet(mochaOpts, '--opts') &&
+    !utils.mochaOptsFileProjectExists();
 
-  // Add source map support
-  if (mochaOpts.indexOf('source-map-support/register') === -1) {
-    // Resolve source-map-support so that the path can be used by mocha
-    const sourceMapRegisterPath = require.resolve(
-      'source-map-support/register',
-    );
-    mochaOpts.unshift('--require', sourceMapRegisterPath);
+  // Add default options
+  if (setMochaOpts) {
+    const mochaOptsFile = utils.getConfigFile('mocha.opts');
+    mochaOpts.unshift('--opts', mochaOptsFile);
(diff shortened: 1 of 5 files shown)


marioestradarosa:
@mrbatista  👍 for this PR.

>  I look forward to knowing from you which of them should I update.

The examples should be updated, in order to reuse this mocha.opts file to keep consistency with the new generated applications.

mrbatista:
> The examples should be updated, in order to reuse this mocha.opts file to keep consistency with the new generated applications.

@marioestradarosa The existing applications should work because they load the autogenerated `mocha.opts` file. With the previous sentence I meant the packages in this repository.

bajtos:
@mrbatista thank you for opening the pull request! I have re-read the code you have changed and I agree our current implementation is confusing for users. However, I am not sure if your proposed solution is going in the right direction.

> After generating a new loopback 4 project with cli generator, a new mocha.opts file is created by default. 

I think our intention here is to provide our recommended Mocha configuration (options) and make make it easy for developers (LoopBack consumers) to tweak it.

All is good so far.

> When I run the npm test command, all options defined in the mocha.opts file should be loaded. 

Agreed! Any other behavior is confusing and I would consider it as a bug.

>  This PR loads your default built-in mocha options only if the corresponding parameter --use-default-options is added to the npm test command.

I think the idea of `lb-mocha` is to simplify life of projects that want to test TypeScript codebase, don't want to deal with Mocha options and want to use the defaults offered by LoopBack. 

A prime example of such consumer is `loopback-next`, where we have "test" scripts defined in the top-level `package.json` and also in package.json files of all subpackages contained by this monorepo. We don't want to clutter our repository with per-package `test/mocha.opts` files sharing the same content, and neither we want to add `--use-default-options` to every `package.json` file in our monrepo.

So, I am proposing the following algorithm:
 - When invoked with `--opts` argument, the path to the configuration is honored. (No change).
 - When the project has `test/mocha.opts` file (relative to current working directory), then we don't interfere and invoke mocha without any `--opts` argument to let it pick the default opts file.
 - Otherwise we add `--opts` pointing to `mocha.opts` provided by `@loopback/build`.

I believe this is the best solution both for us, developers working in side loopback-next, and our users building LoopBack applications/extensions.

Thoughts?

cc @raymondfeng

raymondfeng:
> When invoked with --opts argument, the path to the configuration is honored. (No change).
When the project has test/mocha.opts file (relative to current working directory), then we don't interfere and invoke mocha without any --opts argument to let it pick the default opts file.
Otherwise we add --opts pointing to mocha.opts provided by @loopback/build.

+1.

jannyHou:
+2 for @bajtos 's plan.

mrbatista:
@bajtos @raymondfeng PTAL

Issue #1722, opened 2018-09-18
Partial update (PATCH) over REST


bajtos:
## Description / Steps to reproduce / Feature proposal

Consider the following acceptance tests for Todo controller in our example apps.

```ts
it('updates the todo by ID ', async () => {
  await client
    .patch(`/todos/${persistedTodo.id}`)
    .send({isComplete: true})
    .expect(204);
  const result = await todoRepo.findById(persistedTodo.id);
  const expected = Object.assign(persistedTodo.toJSON(), {isComplete: true};
  expect(result.toJSON()).to.eql(expected);
});

it('returns 404 when updating a todo that does not exist', () => {
  return client
    .patch('/todos/99999')
    .send({isComplete: true})
    .expect(404);
});

```
## Current Behavior

Both tests fail because the PATCH request fails with 422 validation error. The request body is rejected because it's not a valid Todo instance (missing required title property, etc.)

## Expected Behavior

The PATCH operation behaves like a real patch, accepting a subset of Todo properties to update. Both tests pass.

Related issues:
 - OpenAPI decorator does not properly generate schemas of type `Partial` #1179
 - [API Explorer] Fields shown but ignored #1470

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

raymondfeng:
I'm proposing to introduce an `@schema` decorator or extend existing decorators so that a schema can be something like:
```js
{
  schema: {
    'x-ts-type': Order,
    'x-ts-type-options': {
      partial: true,
      includes: ['p1', 'p2'],
      excludes: []
    }
  }

bajtos:
> I'm proposing to introduce an @schema decorator or extend existing decorators so that a schema can be something like.

The discussion about excluding/including certain properties from parameter schema has already started in #1179, let's keep it in a single place there. I'll cross-post your comment @raymondfeng.

IMO, this issue should be focused on leveraging #1179 in our example projects and CLI controller templates.

Alternatively, we can close is as a duplicate of #1179.

bajtos:
On the second thought, maybe this issue is slightly different from #1179.

In #1179, we want to allow `create` operation to exclude certain model properties from the payload, for example `id`.

In this issue, we want to allow all model properties in the payload, but make them all optional.

Ideally, I'd like us to fine a mechanism (decorator API) that can support both use cases in a consistent way.

[...]

PR #1894, opened 2018-10-22
fix(rest): allow @patch update obj to be a Partial

joeytwiddle:
This changes the type of the `obj` argument in `@patch` endpoints from `Model` to `Partial<Model>`.

Since the route checking has become more strict, requests to patch only some fields of a document are now being rejected, because the validation expects to see _all_ required fields in the patch.

In our codebase, we have made the type of the update object `Partial<Model>` to allow partial updates, and that has been working well.

This is only a quick fix. I believe you may settle on a more comprehensive solution later, as discussed in https://github.com/strongloop/loopback-next/issues/1722

- I used `Partial<Model>` because I thought that was more appropriate than `DeepPartial<Model>` or `DataObject<Model>` but I could be wrong.

If this PR is not a step in the desired direction, then please feel free to close it.

## Checklist

  I'm afraid the develop branch is currently failing for me with: `Conflicting definitions for 'express-serve-static-core'`

### DIFF
--- a/docs/site/Controller-generator.md
+++ b/docs/site/Controller-generator.md
@@ -168,7 +168,7 @@ export class TodoController {
     },
   })
   async updateAll(
-    @requestBody() data: Todo,
+    @requestBody() data: Partial<Todo>,
     @param.query.object('where', getWhereSchemaFor(Todo)) where?: Where,
   ): Promise<number> {
     return await this.todoRepository.updateAll(data, where);
@@ -195,7 +195,7 @@ export class TodoController {
   })
   async updateById(
     @param.path.number('id') id: number,
-    @requestBody() data: Todo,
+    @requestBody() data: Partial<Todo>,
   ): Promise<void> {
     await this.todoRepository.updateById(id, data);
   }
--- a/docs/site/todo-list-tutorial-controller.md
+++ b/docs/site/todo-list-tutorial-controller.md
@@ -197,7 +197,7 @@ export class TodoListController {
   })
   async updateById(
     @param.path.number('id') id: number,
-    @requestBody() obj: TodoList,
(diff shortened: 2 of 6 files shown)


bajtos:
@joeytwiddle thank you for opening this pull request. The proposed changes are going in the right direction, but I am don't think we have the required underlying block already in place.

```diff
-    @requestBody() data: Todo,
+    @requestBody() data: Partial<Todo>,
```

The way how `@requestBody()` currently works: it's using the so-called `design:type` metadata produced by TypeScript to learn about the type of the argument value (a `Todo` constructor). When this design:type is a model, then the decorator emits a JSON/OpenAPI schema describing model properties as the accepted value.

When you change the type from `Todo` to `Partial<Todo>`, this metadata is lost and `@requestBody()` decorator sees only a generic `object` type. As a result, the OpenAPI spec describing application's REST API is describing the endpoint as accepting an empty object. At least that's what I remember - could you please check the actual behavior yourself?

If my description above still holds, then we need to find a way how to explicitly tell `requestBody` decorator what type is the parameter accepting, before we can change the CRUD Controller template used by our CLI. See the discussion in https://github.com/strongloop/loopback-next/issues/1722 for implementation proposals.

Would you like to take a look at this underlying feature first?

ambrt:
```
-    @requestBody() data: Todo,
+    @requestBody() data: Partial<Todo>
```

> As a result, the OpenAPI spec describing application's REST API is describing the endpoint as accepting an empty object. At least that's what I remember - could you please check the actual behavior yourself?

I can confirm -  `Partial<Todo>` describes endpoint as accepting empty object.

bajtos:
I am closing this pull request as abandoned, feel free to reopen if you get time to work on this again.

Issue #1874, opened 2018-10-18
Add new extension property to hide controllers and routes from openapi.json


hacksparrow:
## Description / Steps to reproduce / Feature proposal

We should be able to mark some controllers and routes as internal/private, meaning they are not listed in `openapi.json`.

Suggested property name is `x-internal`.

Example usecase is https://github.com/strongloop/loopback-next/pull/1848.

## Current Behavior

There is no way to mark controllers and routes as private.

## Expected Behavior

Should be able to mark controllers and routes as private.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

hacksparrow:
Featured landed via https://github.com/strongloop/loopback-next/pull/1896.

PR #1896, opened 2018-10-22
feat: add 'x-visibility' extension property to OpenAPI spec

hacksparrow:
Add `x-visibility` extension property to OpenAPI spec.

Addresses https://github.com/strongloop/loopback-next/issues/1874.

## Checklist

### DIFF
--- a/packages/rest/src/router/routing-table.ts
+++ b/packages/rest/src/router/routing-table.ts
@@ -156,6 +156,7 @@ export class RoutingTable {
     const paths: PathObject = {};
 
     for (const route of this._router.list()) {
+      if (route.spec['x-visibility'] === 'undocumented') continue;
       if (!paths[route.path]) {
         paths[route.path] = {};
       }
--- a/packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
+++ b/packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
@@ -86,6 +86,27 @@ describe('RestServer.getApiSpec()', () => {
     });
   });
 
+  it('ignores routes marked as "x-visibility" via app.route(route)', () => {
+    function greet() {}
+    function meet() {}
+    server.route(
+      'get',
+      '/greet',
+      {'x-visibility': 'undocumented', responses: {}, spec: {}},
+      greet,
+    );
+    server.route('get', '/meet', {responses: {}, spec: {}}, meet);
+    const spec = server.getApiSpec();
+    expect(spec.paths).to.eql({
+      '/meet': {
+        get: {
+          responses: {},
+          spec: {},
+        },
+      },
+    });
+  });
+
   it('returns routes registered via app.route(..., Controller, method)', () => {
     class MyController {
       greet() {}
@@ -108,6 +129,46 @@ describe('RestServer.getApiSpec()', () => {
           'x-controller-name': 'MyController',
           'x-operation-name': 'greet',
           tags: ['MyController'],
+        },
+      },
+    });
+  });
+
+  it('ignores routes marked as "x-visibility" via app.route(..., Controller, method)', () => {
+    class GreetController {
+      greet() {}
+    }
+
+    class MeetController {
+      meet() {}
+    }
+
+    server.route(
+      'get',
+      '/greet',
+      {'x-visibility': 'undocumented', responses: {}},
+      GreetController,
+      createControllerFactoryForClass(GreetController),
+      'greet',
+    );
+
+    server.route(
+      'get',
+      '/meet',
+      {responses: {}},
+      MeetController,
+      createControllerFactoryForClass(MeetController),
+      'meet',
+    );
+
+    const spec = server.getApiSpec();
+    expect(spec.paths).to.eql({
+      '/meet': {
+        get: {
+          responses: {},
+          'x-controller-name': 'MeetController',
+          'x-operation-name': 'meet',
+          tags: ['MeetController'],
         },
       },
     });


packages/rest/src/router/routing-table.ts
```diff
@@ -156,11 +156,13 @@ export class RoutingTable {
     const paths: PathObject = {};
 
     for (const route of this._router.list()) {
-      if (!paths[route.path]) {
-        paths[route.path] = {};
-      }
+      if (!route.spec['x-internal']) {
```
bajtos: Please use reverse logic + early return to make the code easier to follow and less nested.

```ts
if (route.spec['x-internal']) continue;
```

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -89,6 +89,29 @@ describe('RestServer.getApiSpec()', () => {
     });
   });
 
+  it('ignores routes marked as "x-internal" via app.route(route)', () => {
+    function greet() {}
+    function meet() {}
+    server.route(
+      new Route(
```
bajtos: Let's use the recently introduced sugar API instead of calling `new Route`.

```ts
server.route('get', '/greet', {'x-internal': true, /*...*/}, greet);
```

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -89,6 +89,29 @@ describe('RestServer.getApiSpec()', () => {
...
+        'get',
+        '/greet',
+        {'x-internal': true, responses: {}, spec: {}},
+        greet,
+      ),
+    );
+    server.route(new Route('get', '/meet', {responses: {}, spec: {}}, meet));
```
bajtos: This check feels rather redundant to me - we already have tests that are verifying how regular (non-internal) routes are published to the spec, no need to duplicate those test again.

In case you convince me otherwise, then replace `server.route(new Route(...))` to `server.route(...)`.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -89,6 +89,29 @@ describe('RestServer.getApiSpec()', () => {
...
+    const spec = server.getApiSpec();
+    expect(spec.paths).to.eql({
+      '/meet': {
```
bajtos: See my comment above. In this test, it should be enough to verify that when the app has only a single route that's internal, then the spec object has no paths.

```ts
expect(spec.paths).to.eql({});
```

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -116,6 +139,46 @@ describe('RestServer.getApiSpec()', () => {
...
+      {'x-internal': true, responses: {}},
+      GreetController,
+      createControllerFactoryForClass(GreetController),
+      'greet',
+    );
+
+    server.route(
+      'get',
+      '/meet',
+      {responses: {}},
```
bajtos: Same comment as above - let's keep this test focused on the x-internal method only and don't duplicate the tests verifying how non-internal routes are handled.

packages/rest/src/router/routing-table.ts
```diff
@@ -202,7 +208,7 @@ export interface RouteEntry {
   /**
    * OpenAPI operation spec
    */
-  readonly spec: OperationObject;
+  readonly spec: LB4OperationObject;
```
bajtos: This change is not strictly necessary because `OperationObject` type already supports arbitrary extension (additional properties). Having said that, I see how `LB4OperationObject` can be useful because it allows IDEs like VisualStudio Code to offer auto-completion and also the compiler can check that developers don't assign a non-boolean value to this extension by a mistake.

So, if we want to keep `LB4OperationObject`, then please find other places in our public API that are accepting `OperationObject` and change them accordingly. Few examples to get you started:
 - `@operation` decorator and verb-based shortcuts like `@get`
 - `server.route()` method
 - `Route` and `ControllerRoute` classes (and their constructors)

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
hacksparrow: VSC is complaining `No overload expects 4 arguments, but overloads do exist that expect either 1 or 6 arguments.`.

Did you mean something like this?

```
server.route('get', '/greet', spec, MyController, factory, 'greet');
```

We'll have to create `MyController` and `factory`, in that case. It adds more to the test code.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
hacksparrow: In my first bad attempt some days ago, setting `x-internal` hid all the routes. Let's ensure `x-internal` does not have side effects.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
hacksparrow: Same as above, verify for no side-effects.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
hacksparrow: Same thing, verify no side-effects.

packages/rest/src/router/routing-table.ts
hacksparrow: Let's keep it and update the references.

hacksparrow: Replacing `OperationObject ` with `LB4OperationObject ` in all instances is not easy, let's not do it in this PR. So, for not let's keep it as such. Reverting to `readonly spec: OperationObject`.

packages/rest/test/unit/rest.server/rest.server.open-api-spec.unit.ts
hacksparrow: Same thing as above.

packages/rest/src/router/routing-table.ts
```diff
@@ -156,13 +156,12 @@ export class RoutingTable {
     const paths: PathObject = {};
 
     for (const route of this._router.list()) {
+      if (route.spec['x-internal']) continue;
```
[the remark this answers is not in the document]
hacksparrow: We just want to hide the route from OpenAPI spec, so unlisted is what we want.

What purpose could a route that's hidden from the spec and 404s, serve?

@bajtos, thoughts on @raymondfeng's suggestions?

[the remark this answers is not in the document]
bajtos: We want to hide the route from the generated OpenAPI spec, but still be available via the REST API for clients that know the route. Example use cases:
 - a health endpoint used for DevOps monitoring
 - an internal endpoint used by API Explorer to load dynamic configuration for a static-only front-end (we do this in LB 3.x now)
 - a catch-all route to serve static files - see https://github.com/strongloop/loopback-next/pull/1848

Personally, I prefer `x-hidden: true` or `x-documented: false` most.

hacksparrow:
So @raymondfeng what about `x-hidden: true` or `x-documented: false`?

raymondfeng:
Maybe something like `x-visibility: unlisted` or `x-visibility: undocumented`? This way, we can extend the values in the future.

bajtos:
Let's go with `x-visibility: undocumented` then 👍

Issue #1775, opened 2018-09-29
[Docs] Add docs for exposing GraphQL APIs


dhmlau:
## Description / Steps to reproduce / Feature proposal
Currently user can do the following steps to create GraphQL APIs within a LoopBack app:

1. Run npm i --save express express-graphql oasgraph
2. Make sure the LB app has started. 
3. From the browser, go to http://localhost:3000/openapi.json, copy the openapi.json, and save it under the root of the LoopBack application, say `todo-openapi.json`
4. Run `./node_modules/.bin/oasgraph todo-openapi.json`
5. Test it by going to http://localhost:3001/graphql

## Acceptance Criteria
- [ ] Determine where to put the docs.  
   - Perhaps as one of the tutorials in https://loopback.io/doc/en/lb4/Examples-and-tutorials.html
- [ ] Create the docs page with the above steps 
   - Verify the steps first as things might've changed by the time of writing.

marioestradarosa:
> create GraphQL APIs

For what I understood, please correct me, it can only expose GraphQL Schemas from the OpenAPI spec file, but not it won't create any API.  If this is correct ,then I think you should change the title, otherwise, ignore this message 😄

marioestradarosa:
@dhmlau , Here is my summary for this weekend. I played with OASGraph as a library and wrote this small presentation. We need to check what to add/remove etc. I see a potential way to mount the /graphql using the internal express. So, the developer can configure the option if he/she wants a graphql server , something like `{ graphql : { available: true, mountPoint: '/graphql' } }`  passing this to the application, the internal express can be used to create it using oasgraph as the translator. 

At this moment, I just played with the two dependencies and could manage to make a working loopback 4 application with no problems. I think the only problem at this moment is the the Delete operation not being sent the correct format to **OASGraph**, I will open an issue on this, aside from that I could test the relationships on the todo-list querying them and works fine. 

Here is a small video presentation summarizing my findings this weekend. Note that the current oasgraph CLI is only used to test an openapi.json spec file, this won't work in production, probably the best way is to use the mechanism I placed on the video, this way, I didn't have to install the oasgraph CLI or run npm link etc.

https://www.youtube.com/watch?v=s2hScucNmaA

dhmlau:
@marioestradarosa , yes, but I haven't given too much thought on where it should be. :)   Any suggestion? 
[...]

Issue #656, opened 2017-10-17
Extension: GraphQL


dhmlau:
# Description/Steps to reproduce

| Extension        | Descriptions           | Developers  |  LB4 Maintainer
| ------------- |:-------------:| -----:|-----:|
| GraphQL | Component to provide GraphQL Endpoint for Models. | @virkt25 , @mpicard, Help Welcome | Taranveer  (@virkt25) |
|       |      |    ||

# Additional information
This is originated from Call for Contribution on LB Extensions [issue](https://github.com/strongloop/loopback-next/issues/647).

mpicard:
Looking forward to working on this! 😄

dhmlau:
added @beeman.  Welcome aboard!

beeman:
Thanks @dhmlau!

raymondfeng:
FYI: Here are prior arts that integrate GraphQL with LoopBack 3.x:

- https://github.com/yahoohung/loopback-graphql-server
- https://github.com/Tallyb/loopback-graphql
- https://github.com/BlueEastCode/loopback-graphql-relay

mpicard:
I'd like to look at https://github.com/apollographql/apollo-server as well as incorporating the https://github.com/apollographql/apollo-tracing interface for metrics! Would be cool to enable that.

dhmlau:
@mpicard , now that you've mentioned apollo, here is another reference too. 
https://github.com/gramps-graphql/gramps-express

mpicard:
@dhmlau very cool project! I wasn't aware of it. Thanks for sharing, the modular aspects of it look great and would be a great for lb!

dhmlau:
Still needed!
I wonder if there's any news on this extension.  And would it be possible if we can use  https://github.com/yahoohung/loopback-graphql-server (just an example) as a reference for implementation?

peanutenthusiast:
@dhmlau Any updates with this issue? My team is looking into using LB4 and GraphQL in production, so I'd be interested in providing a helping hand.

dhmlau:
@peanutenthusiast, thanks for your interest!  We’re looking at an approach that would convert your REST APIs into GraphQL APIs.  Would that be something fit your need?  
It is a bit different from https://github.com/yahoohung/loopback-graphql-server which takes LB artifacts to convert into GraphQL APIs.  I can provide more details once we're (or close to) ready.

peanutenthusiast:
[...]

PR #1899, opened 2018-10-22
docs: add tutorial to expose GraphQL APIs

dhmlau:
Fixes https://github.com/strongloop/loopback-next/issues/1775

Based on the [QuickLab](https://developer.ibm.com/tutorials/create-rest-apis-minutes-with-loopback-4/) we've created for Node+JS Interactive, this PR added the GraphQL APIs part as one of the tutorials.


docs/site/exposing-graphql-api.md
```diff
@@ -0,0 +1,106 @@
...
+
+Make sure you have a running LoopBack 4 application. In this tutorial, we'll use
+the `todo` example. You can create this application by running the command
+below:
+
+```sh
+lb4 example todo
+```
+
+### Install OASGraph and required dependencies
```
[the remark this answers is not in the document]
dhmlau: OK. Will change it. 
Actually our docs are not quite inconsistent on the capitalization of the titles. :(

marioestradarosa:
@dhmlau,

I think this line is missing from inside the src/application.ts file, or can it be resolved differently?

```ts
options = Object.assign({}, {
       rest: {
         openApiSpec: {
           servers: [{url:'http://localhost:3000'}],
         }
       }
    });
```

otherwise, the query in the doc will return

```
{
  "errors": [
    {
      "message": "Invalid URI \"/todos\"",
      "locations": [
```

marioestradarosa:
> Thanks @marioestradarosa for catching that. Somehow I thought it's no longer needed! PTAL again.

@dhmlau , I have just open a PR #1900  that I think can make things easier for the application in this matter.  If it lands, then I believe we should then modify the examples accordingly.

thoughts?

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,131 @@
+---
+lang: en
+title: 'Exposing GraphQL APIs'
+keywords: LoopBack 4.0, LoopBack 4
```
bajtos: Should we add "GraphQL" keyword too?

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+### Install OASGraph and Required Dependencies
+
+From your LoopBack application, run the following command to install OASGraph
+and the required dependencies:
+
+```sh
+npm i --save oasgraph express-graphql
```
bajtos: I am curious, why do we have to install `express-graphql` ourselves? I would expect the CLI command `oasgraph` to be self-contained and come with all dependencies.

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,131 @@
...
+express on port 3001. Specifying the OpenAPI spec generated by the
+todo-application as the parameter, start up the server by running the following
+command:
+
+```sh
+node_modules/.bin/oasgraph http://localhost:3000/openapi.json
+```
+
+That’s it! You’re now ready to try out some tests and requests in the browser at
+http://localhost:3001/graphql.
```
bajtos: While this is okay for building prototypes and exploring GraphQL API, I find it cumbersome to have the GraphQL endpoint running on a different port than the main REST API. It complicates configuration of gateways, load balancers and other DevOps infrastructure. Can we find a way how to server GraphQL at the same port as the REST API please? I think this work may be best left out of scope of this blog post, so let's create a new GitHub issue to keep track of this work and explain in this blog post that we are looking into ways how to server both REST and GraphQL at the same port & link to the new GitHub issue.

Thoughts?

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,131 @@
...
+     title
+     desc
+   }
+ }
+```
+
+The expected output looks like this:
+
+```json
+{ "data": { "todos": [ { "id": 1, "title":
```
bajtos: Can you reformat the data to make it easier to read please? E.g. use `JSON.stringify(data, null, 2)`, or perhaps put each todos array item on its own line.

```ts
{ "data": { "todos": [
  {"id": 1, /*...*/},
  {"id": 2, /*...*/},
  // etc.
]}}
```

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -98,6 +98,11 @@ children:
     - title: 'Run and Test it'
       url: soap-calculator-tutorial-run-and-test.html
       output: 'web, pdf'
+
+  - title: 'Exposing GraphQL APIs'
```
bajtos: Is this still a Tutorial? I feel that "Exposing GraphQL APIs" and "Deploying to IBM Cloud" don't belong to Tutorials section, I would certainly not look for these pages in Tutorials myself!

Can we introduce a new section "Guides" please and move both pages there?

docs/site/exposing-graphql-apis.md
marioestradarosa: The **OASGraph** is meant to be used as a library to transform from **OpenAPI** to **GraphQL** schema, It doesn't need `express-graphql`, since you can use any other package once you have your schema converted.  The main idea of the library  is to be used inside the application.

However the **CLI** is used to test the implementation and it requires `express-graphql`. It is not a separate package _(like monorepo)_, it is just a single file that implements the **GraphQL** server with the generated schema. IMO, it is not meant for production.

docs/site/exposing-graphql-apis.md
marioestradarosa: >  I find it cumbersome to have the GraphQL endpoint running on a different port than the main REST API. It 

Correct, the main idea of **GraphQL** library is to be used by the application. I opened #1900 for a starting point, since first we need to be able to generate the server property adequately. However, the next step is to probably :

1.  Create a CLI option `--enable-graphql` ?
2. Used the generated **OpenAPI** _schema_ by the library _(already installed by --enable-graphql in our application)_  and get an internal representation of the GraphQL Schema
3.  Use that generated GraphQL Schema to create an internal end-point (ie: /lb4graphql)  by the internal express and use the `express-graphql`  to consume that schema

That way, it will be in the same port number and u can consume graphql api from /lb4graphql .

docs/site/sidebars/lb4_sidebar.yml
marioestradarosa: How about turning them into a **How to** sections ?

docs/site/exposing-graphql-apis.md
bajtos: Thanks @marioestradarosa. @dhmlau has pointed me to https://github.com/strongloop/oasgraph/issues/30, I have a better understanding now.

So, assuming oasgraph is not going to change, I am proposing the following changes:

Include `express` in the modules to install, we must not rely on `express` being installed in top-level `node_modules` by being a dependency of `@loopback/rest`.

```shell
$ npm i --save oasgraph express express-graphql
```

Explain the users why we are asking them to install those two extra express modules and point them to the discussion in https://github.com/strongloop/oasgraph/issues/30 to learn more and/or give us feedback.

docs/site/exposing-graphql-apis.md
bajtos: Thank you for opening #1900.

As I see it, the first step is to find out how to integrate `oasgraph` into LB4 application - what kind of HTTP request handler API is the module exposing? How can we plug it into an LB4 application? Do we need any any new extension points to enable this integration?

IMO, the `oasgraph` integration should live in a new extension package independent of `@loopback/core`. This package should be probably living in loopback-next alongside other LB4 modules.

I guess this is mostly what you have described in your steps 2 & 3.

Adding a CLI option like `--enable-graphql` should be the last step, the icing on a cake so to speak.

bajtos:
Hmm, we may want to use a deep-assign implementation instead of `Object.assign`- `Object` assign will overwrite the entire `rest` configuration with the object provided by the caller.

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,131 @@
...
+`http://localhost:3000/openapi.json`. If not, you can start it by running the
+`npm start` command.
+
+Now we will use the oasgraph CLI to set up a GraphQL HTTP Server backed by
+express on port 3001. Specifying the OpenAPI spec generated by the
+todo-application as the parameter, start up the server by running the following
+command:
+
+```sh
+node_modules/.bin/oasgraph http://localhost:3000/openapi.json
```
raymondfeng: This can be simplified as `npx oasgraph http://localhost:3000/openapi.json`

marioestradarosa:
> `Object` assign will overwrite the entire `rest` configuration with the object provided by the caller.

I thought that adding options as the end parameter prevents this overwrite.

docs/site/exposing-graphql-apis.md
marioestradarosa: Just need to mention the user that `npx`  comes bundled with npm version >5.2 . it works fine, I have just tested, didn't know about `npx` 😄

docs/site/exposing-graphql-apis.md
marioestradarosa: > IMO, the oasgraph integration should live in a new extension package independent of @loopback/core 

I don't see the integration too complicated.  It only consumes the **openAPISpec** and transforms it to a **graphQLSchema** (of course more complex validations sanitations etc.), with that all we need is to use `express-graphql` as a middleware and probably reuse the same express used to expose the `openapi.json` file. 

So, in options you can disable it by setting  {  rest: {... }  ,  **oasgraph** : false }   and if that's true then we register the middleware. The middleware `express-graphql` only receives the generated graphQL Schema and take charge of everything else.

marioestradarosa: Does it make sense to have it in a separate package?, it takes two packages and a couple of lines to implement it.

dhmlau: > That way, it will be in the same port number and u can consume graphql api from /lb4graphql 

+1.  Perhaps simply /graphql?

docs/site/sidebars/lb4_sidebar.yml
dhmlau: `How to` sounds good to me.

dhmlau:
Thanks for your feedback! To move this PR forward, I'm going to incorporate the feedback with the major changes:
- Remove `Customize the Server URL of the REST Server` section to modify the application.ts after @marioestradarosa's PR (#1900 )has landed
  - I think the PR includes the template changes in the application.ts, and supposedly users don't need to do anything else.  
- Move the instructions for GraphQL (this PR) and "Deploying to Cloud" to a new section `How Tos`. 

Will create a new task for having the oasgraph and LB rest server running on the same port.

docs/site/exposing-graphql-apis.md
bajtos: Node.js 8.9.0, the minimal Node.js version required by LoopBack 4, comes with npm version 5.5.1, which includes `npx`. IMO, there is no need to explain how to run oasgraph on npm version older than 5.2.

Having said that, I think most LoopBack users are probably not familiar with npx yet, so we should give them a pointer on where to learn more.

For example, add the following paragraph after the code snippet showing how to invoke `npx`:

```markdown
_Haven't heard about `npx` yet? It's a cool helper provided by `npm` 
and available out of the box since Node.js 8.x. Learn more in their announcement blog post:
[Introducing npx: an npm package runner](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b))_
```

docs/site/exposing-graphql-apis.md
bajtos: > Does it make sense to have it in a separate package?, it takes two packages and a couple of lines to implement it.

If we keep graphql built into our `@loopback/rest` package, then we end up with a tight coupling to a specific semver-major version of `oasgraph `. Whenever `oasgraph` makes breaking changes, we need to release a new semver-major vresion of `@loopback/rest`. 

We have been bitten by this kind of tight coupling many times in LB 3.x. I don't remember a single time when we would wish an independent extension was built into the framework. There were many times we wished a built-in feature was packaged as a standalone extension from the beginning: offline sync, authentication & authorization layer, etc.

Also by forcing all `@loopback/rest` consumers to depend on `oasgraph`, we make it harder for people not using GraphQL to avoid possible security vulnerabilities in `oasgraph` and its dependencies.

Last but not least, while `oasgraph` may be a good solution for prototyping GraphQL support, I think it's not suitable for production deployment on sites with a decent amount of traffic, because it cannot optimize database queries to fetch data from multiple related models in a single SQL query. I'd like us to look into ways how to allow LB4 app developers to leverage full power of database access and relations in the GraphQL API, see https://github.com/strongloop/loopback-next/issues/656 and https://github.com/strongloop/loopback4-extension-graphql. If we bundle `oasgraph` as the default GraphQL implementation, we are nudging app developers to use a tool that may shoot them in their foot in a near future of their project.

bajtos: Let's move this discussion to https://github.com/strongloop/loopback-next/issues/1905 please.

docs/site/exposing-graphql-apis.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+If you're using npm version lower than 5.2, run the command below instead:
+
+```sh
+node_modules/.bin/oasgraph http://localhost:3000/openapi.json
+```
+
+That’s it! You’re now ready to try out some tests and requests in the browser at
+http://localhost:3001/graphql.
```
bajtos: Let's start a new conversation focused on the scope of this pull request.

I am proposing to add a short explanation of why GraphQL is served at a different port and point people to #1905 to get updates on our progress towards a better solution.

For example:

```markdown
That’s it! You’re now ready to try out some tests and requests in the browser at
http://localhost:3001/graphql.

{% include note.html content="
We are looking into ways how to expose the GraphQL endpoint alongside the main REST API,
on the same HTTP host and port. See
[issue #1905](https://github.com/strongloop/loopback-next/issues/1905).
" %}
```

marioestradarosa:
@dhmlau , I think your PR is now ready to land 👍

PR #1900, opened 2018-10-23
refactor(cli): use options to specify port and OpenAPI server url

marioestradarosa:
This PR only applies the options object in the application to benefit the app developer in two areas:

1. For new developers change the application port number with ease
2. Avoid the step to specify the server url + port number when used by OASGraph _(if not used, then I believe it won't hurt? )_

This pull request supersedes #1900.

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/index.js.ejs
+++ b/packages/cli/generators/app/templates/index.js.ejs
@@ -4,7 +4,17 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
-  application.main().catch(err => {
+  const config = {
+    rest: {
+      port: +process.env.PORT || 3000,
+      host: process.env.HOST || 'localhost',
+      openApiSpec: {
+        // useful when used with OASGraph to locate your application
+        setServersFromRequest: true,
+      },
+    },
+  };
+  application.main(config).catch(err => {
     console.error('Cannot start the application.', err);
     process.exit(1);
   });
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -40,6 +40,10 @@ describe('app-generator specific files', () => {
     assert.fileContent('src/application.ts', /constructor\(/);
     assert.fileContent('src/application.ts', /this.projectRoot = __dirname/);
 
+    assert.file('index.js');
+    assert.fileContent('index.js', /openApiSpec: {/);
+    assert.fileContent('index.js', /setServersFromRequest: true/);
+
     assert.file('src/index.ts');
     assert.fileContent('src/index.ts', /new MyAppApplication/);
     assert.fileContent('src/index.ts', /await app.start\(\);/);
--- a/examples/hello-world/index.js
+++ b/examples/hello-world/index.js
@@ -9,7 +9,17 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
-  application.main().catch(err => {
+  const config = {
+    rest: {
+      port: +process.env.PORT || 3000,
+      host: process.env.HOST || 'localhost',
+      openApiSpec: {
+        // useful when used with OASGraph to locate your application
+        setServersFromRequest: true,
+      },
+    },
+  };
+  application.main(config).catch(err => {
     console.error('Cannot start the application.', err);
     process.exit(1);
   });
--- a/examples/soap-calculator/index.js
+++ b/examples/soap-calculator/index.js
@@ -4,7 +4,17 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
-  application.main().catch(err => {
+  const config = {
+    rest: {
+      port: +process.env.PORT || 3000,
+      host: process.env.HOST || 'localhost',
+      openApiSpec: {
+        // useful when used with OASGraph to locate your application
+        setServersFromRequest: true,
+      },
+    },
+  };
+  application.main(config).catch(err => {
     console.error('Cannot start the application.', err);
     process.exit(1);
   });
--- a/examples/todo-list/index.js
+++ b/examples/todo-list/index.js
@@ -9,6 +9,16 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
+  const config = {
+    rest: {
+      port: +process.env.PORT || 3000,
+      host: process.env.HOST || 'localhost',
+      openApiSpec: {
+        // useful when used with OASGraph to locate your application
+        setServersFromRequest: true,
+      },
+    },
+  };
   application.main().catch(err => {
     console.error('Cannot start the application.', err);
     process.exit(1);
--- a/examples/todo/index.js
+++ b/examples/todo/index.js
@@ -9,7 +9,17 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
-  application.main().catch(err => {
+  const config = {
+    rest: {
+      port: +process.env.PORT || 3000,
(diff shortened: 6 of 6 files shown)


packages/cli/generators/app/templates/src/application.ts.ejs
```diff
@@ -19,6 +19,21 @@ export class <%= project.applicationName %> extends BootMixin(
 export class <%= project.applicationName %> extends BootMixin(RestApplication) {
 <% } -%>
   constructor(options: ApplicationConfig = {}) {
+    options = Object.assign(
+      {
+        rest: {
+          // your application listening port number, you can change this once
+          // in production
+          port: 3000,
```
raymondfeng: Does it make sense now to have `port: +process.env.PORT || 3000` now?

packages/cli/generators/app/templates/src/application.ts.ejs
```diff
@@ -19,6 +19,21 @@ export class <%= project.applicationName %> extends BootMixin(
...
+          openApiSpec: {
+            // useful when used with OASGraph to locate your application when
+            // resolving GraphQL schemas
+            servers: [{url: 'http://localhost:3000'}],
```
raymondfeng: Is `setServersFromRequest: true` better as we don't always know the port in advance?

packages/cli/generators/app/templates/src/application.ts.ejs
marioestradarosa: Yes, you are right, I forgot about the process.env.PORT support

marioestradarosa: is the plus symbol `+process.env.PORT` intentional ?

packages/cli/generators/app/templates/src/application.ts.ejs
raymondfeng: If you have `setServersFromRequest: true`, what's `servers` from the generated openapi spec?

Please note that browsers may cache the response. I had to use private browsing sometimes to work around it.

marioestradarosa: > If you have `setServersFromRequest: true`, what's `servers` from the generated openapi spec?

```
"servers": [
    {
      "url": "http://[::1]:3010"
    }
  ]
```
However, the `node_modules/oasgraph/cli.js http://localhost:3010/openapi.json` still doesn't work. I guess the problem is with OASGraph not liking the format [::1]  ?

marioestradarosa: Neither `node_modules/.bin/oasgraph http://localhost:3000/openapi.json`  or probably the problem is the `express-graphql` since this is the one connecting the resolver from the GraphQL schema.

packages/cli/generators/app/templates/src/application.ts.ejs
raymondfeng: Yes. It converts the value to `number`.

packages/cli/generators/app/templates/src/application.ts.ejs
raymondfeng: `http://[::1]:3010` uses ipv6 address.

marioestradarosa:
@raymondfeng `setServersFromRequest: true` works fine.  I was querying the wrong schema from my browser. PTAL.

packages/cli/generators/app/templates/src/application.ts.ejs
```diff
@@ -19,6 +19,20 @@ export class <%= project.applicationName %> extends BootMixin(
 export class <%= project.applicationName %> extends BootMixin(RestApplication) {
 <% } -%>
   constructor(options: ApplicationConfig = {}) {
+    options = Object.assign(
+      {
+        rest: {
+          // your application listening port number, you can change this once
+          // in production
+          port: process.env.PORT || 3000,
```
bajtos: would you mind adding a `host` configuration too?

```
host: process.env.HOST || undefined
```

marioestradarosa:
@bajtos, @raymondfeng  PTAL

packages/cli/test/integration/generators/app.integration.js
```diff
@@ -39,6 +39,9 @@ describe('app-generator specific files', () => {
     );
     assert.fileContent('src/application.ts', /constructor\(/);
     assert.fileContent('src/application.ts', /this.projectRoot = __dirname/);
+    assert.fileContent('src/application.ts', /openApiSpec: {/);
+    assert.fileContent('src/application.ts', /port: process.env.PORT || 3000/);
```
raymondfeng: Is this still true after moving `process.env.PORT` to `index.js`?

marioestradarosa: You are right!, but it didn't give any error, it should have returned false, what would be happening?

marioestradarosa: do you think the `||`  is being evaluated ?

raymondfeng: Yeah, you regular expression might have wildcard matches.

marioestradarosa:
@raymondfeng I addressed the test evaluation. PTAL

packages/cli/generators/app/templates/index.js.ejs
```diff
@@ -4,7 +4,13 @@ module.exports = application;
 
 if (require.main === module) {
   // Run the application
-  application.main().catch(err => {
+  const config = {
+    rest: {
+      port: process.env.PORT,
```
raymondfeng: `+process.env.PORT || 0`?

marioestradarosa: fixed!, thanks @raymondfeng

marioestradarosa:
@bajtos  PTAL

packages/cli/generators/app/templates/src/application.ts.ejs
```diff
@@ -19,6 +19,17 @@ export class <%= project.applicationName %> extends BootMixin(
 export class <%= project.applicationName %> extends BootMixin(RestApplication) {
 <% } -%>
   constructor(options: ApplicationConfig = {}) {
+    options = Object.assign(
```
bajtos: Could you please verify that `Object.assign` is preserving deep values provided by `options`? I believe it will replace `rest` object as a whole, i.e. changing `{rest: {port: 123, host: 'name'}}` provided by the top-level `index.js` into `{rest: {openApiSpec: {setServersFromRequest: true}}}`. A possible solution is to use a deep-merge algorithm, there are many implementations available on npmjs. [deepmerge](https://www.npmjs.com/package/deepmerge) seems like a popular one that's still actively maintained. `lodash` provides deep-merge or deep-assign API too.

raymondfeng: Why not move the config to `index.js.ejs`?

marioestradarosa: That could also be a good solution @raymondfeng , simplifies everything for this context. @bajtos  thoughts?

marioestradarosa:
> Please update all example projects to be consistent with the new index.js template

@bajtos / @raymondfeng  could you check the new application.ts ?, and I also placed a default value for host on index.js 

Now it works with lodash merge method and is preserving the port and host passed thru the index.js file. After that,  I will update the examples too.

raymondfeng:
I still prefer to move it into index.js.ejs. Using lodash to deep merge in application ctor is a bit overwhelming.

PR #1903, opened 2018-10-23
feat: import strong-docs as @loopback/tsdocs

raymondfeng:
- Importing strong-docs into the monorepo as `@loopback/tsdocs`
- Removing support for plain JS

The history of `strong-docs` is imported using `lerna import`. Do we want to preserve these commits? The downside is that such commits didn't follow our message conventions.

## Checklist


bajtos:
> The history of strong-docs is imported using lerna import. Do we want to preserve these commits? The downside is that such commits didn't follow our message conventions.

I don't have a strong opinion. I guess it would be ok to structure this pull request as the following list of commits:
 1. Import the current codebase from strong-docs. Ideally, there should be no changes made compared to the original codebase - just squash the commits produces by `lerna import` into a single one.
 2. Update the codebase to work with our monorepo (pass the tests and linter checks)
 3. Remove support for plain JavaScript
 4. Upgrade the codebase to TypeScript, make sure it passes all of our linter rules

docs/site/DEVELOPING.md
```diff
@@ -144,9 +144,9 @@ test/unit/application.unit.ts
 
 ## API Documentation
 
-We use [strong-docs](https://github.com/strongloop/strong-docs) to generate API
-documentation for all our packages. This documentation is generated when
-publishing new releases to npmjs.org and it's picked up by
+We use [@loopback/tsdocs](https://github.com/strongloop/@loopback/tsdocs) to
```
bajtos: Is it still our intention to allow LB4 developers to build applications and extension in plain JavaScript? If yes, then I think we will eventually need to bring back support for JavaScript. To keep that option open, I am proposing to use a different name than `@loopback/tsdocs`, because that one is tied to TypeScript. Have you considered my proposal for `@loopback/apidocs`?

examples/strong-docs/.github/ISSUE_TEMPLATE.md
```diff
@@ -1,37 +0,0 @@
-<!--
-Questions:
-  https://groups.google.com/forum/#!forum/loopbackjs
-  https://gitter.im/strongloop/loopback
-Immediate support:
-  https://strongloop.com/api-connect-faqs/
-  https://strongloop.com/node-js/subscription-plans/
--->
-
-# Description/Steps to reproduce
```
bajtos: Huh, I am confused - did you accidentally import `strong-docs` into `examples` directory? It looks like you are moving the imported source code from `examples/strong-docs` to `packages/strong-docs`. 

The pull request diff is humongous and very difficult to review as a result :(

I am also concerned about making code archaeology very difficult in the future.

Can you please re-run the import again (or simply copy the code from strong-docs over as I suggested in my comment above) and ensure the files are added to `packages/{new package name}` from the beginning?

bajtos:
@raymondfeng please make sure to add a new entry to `CODEOWNERS` and `MONOREPO.md` files as part of this work.

lerna.json
```diff
@@ -1,11 +1,11 @@
...
   "packages": [
-    "benchmark",
-    "docs",
-    "examples/*",
     "packages/*",
-    "sandbox/*"
+    "examples/*",
+    "sandbox/*",
+    "benchmark",
+    "docs"
```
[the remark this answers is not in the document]
raymondfeng: For some reason, `lerna import` always uses `examples/*` as the target and it cannot be customized. Initially I thought the order matters but it does not.

docs/site/DEVELOPING.md
raymondfeng: I rename the module as `tsdocs` to make it clear that only TS is supported. If we want to keep JS support, I would rather not to import it into our monorepo as it adds overhead and slows down the build.

bajtos: Fair enough. We can start worrying about JS support when it becomes relevant.

lerna.json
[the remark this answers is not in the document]
bajtos: That's really bad :( 

How about the following recipe?

```
$ lerna import (...)
$ mv examples/strong-docs packages/tsdocs
$ git add .
$ git commit --amend
```

Sorry for spending so much of our time on this topic. It's a huge patch that's difficult enough on it's own, without having to deal with file renames between commits.

bajtos:
@raymondfeng what's the status of this pull request?

Maybe we should consider using https://github.com/TypeStrong/typedoc instead of maintaining our own API doc generator? See also https://blog.cloudflare.com/generating-documentation-for-typescript-projects/ and the discussion under the post.

raymondfeng:
@bajtos After some thoughts, I'm not very keen on adding `@loopback/tsdocs` into the monorepo any more for a few reasons:

1. It creates a bit circular deps
2. It loses plain JS support
3. It loses commit history

BTW, I have trouble running `typedoc` cli against our modules.

bajtos:
Fair enough. Let's close this pull request then.

Issue #1892, opened 2018-10-21
x-ts-type requires model pointed in #/components/schemas


chunghha:
## Description / Steps to reproduce / Feature proposal
To reproduce, you may checkout develop branch of this repo first,
https://github.com/chunghha/lb-poc/commit/5aa2da90281484b581a64dbea0872d21d8fc406f
then uncomment the line of 'x-ts-type': Country in the country.controller.ts
## Current Behavior
Once the above steps done, open http://explorer.loopback.io/?url=http://localhost:3000/openapi.json#/CountryController/get_country
to see the error, "Could not resolve reference because of: Could not resolve pointer: /components/schemas/Country does not exist in document"
 
## Expected Behavior
Country model (country.model.ts) exists in src/models so expect that model pointed with 'x-ts-type' without re-writing properties of Country to #/components/schemas.

dhmlau:
It seems like your generated code is a bit out-of-date. 
I'm trying to recreate the app based on your commit provided above, and there is ` "$ref": "#/components/schemas/Country"` in the openapi.json

You can see my version of the app here: https://github.com/dhmlau/temp-country-poc:

```
"get": {
        "x-controller-name": "CountryController",
        "x-operation-name": "find",
        "tags": [
          "CountryController"
        ],
        "responses": {
          "200": {
            "description": "Array of Country model instances",
            "content": {
              "application/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/Country"
                  }
                }
              }
            }
          }
        },
```

Hope it helps.

dhmlau:
@chunghha, sorry that I missed that you've commented out the `x-ts-type` in the controller.  Yes, it does make a difference if you commented it, because in that case, there is no way to know what will get returned when generating the openapi.json.

The questions to ask would be:
Does the service provider return the same model (data structure) as yours? 
What do you want your API to return?

chunghha:
When the x-ts-type line is uncommented, I don't see #/components/schemas/Country generated, only $ref line is generated so answers to your questions.

1. The service provider returns more than what I want. I like to control response properties defined in my country.model.ts. When I need to add more properties I will add them to the Country model.

2. My country service's "countries" service endpoint will return response as [Country].

[...]

PR #1907, opened 2018-10-24
fix(openapi-v3): generate schemas for x-ts-type

raymondfeng:
https://github.com/strongloop/loopback-next/issues/1892

## Checklist

### DIFF
--- a/packages/openapi-v3/src/controller-spec.ts
+++ b/packages/openapi-v3/src/controller-spec.ts
@@ -11,6 +11,10 @@ import {
   PathObject,
   ComponentsObject,
   RequestBodyObject,
+  ResponseObject,
+  ReferenceObject,
+  SchemaObject,
+  isReferenceObject,
 } from '@loopback/openapi-v3-types';
 import {getJsonSchema} from '@loopback/repository-json-schema';
 import {OAI3Keys} from './keys';
@@ -49,6 +53,8 @@ export interface RestEndpoint {
   path: string;
   spec?: OperationObject;
 }
+
+export const TS_TYPE_KEY = 'x-ts-type';
 
 /**
  * Build the api spec from class and method level decorations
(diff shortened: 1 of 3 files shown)


packages/openapi-v3/src/controller-spec.ts
```diff
@@ -214,47 +209,84 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+  debug('  evaluating content: %j', content);
+  const tsType = content[TS_TYPE_KEY];
+  debug('  %s => %o', TS_TYPE_KEY, tsType);
+  if (tsType) {
+    content.schema = resolveSchema(tsType, content.schema);
+    if (content.schema.$ref) generateOpenAPISchema(spec, tsType);
+
+    // We don't want a Function type in the final spec.
+    delete content[TS_TYPE_KEY];
+  } else {
```
bajtos: Let's return early so that the next block of code can be indented by one less level.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -214,47 +209,84 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+    const schema = content.schema;
+    if (!schema || isReferenceObject(schema)) return;
+    if (schema.type === 'array') {
```
bajtos: So IIUC we are not doing any work if the schema is not array? Can we use early return with a comment to explain? E.g.

```ts
if (schema.type !== 'array') {
  // the schema should be well-specified by now
  return;
}
```

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -214,47 +209,84 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+      const itemSchema = schema.items;
+      if (!itemSchema || isReferenceObject(itemSchema)) return;
+      const itemTsType = itemSchema[TS_TYPE_KEY];
+      debug('  %s => %o', TS_TYPE_KEY, itemTsType);
+      if (itemTsType) {
+        schema.items = resolveSchema(itemTsType, itemSchema);
+        if (schema.items.$ref) generateOpenAPISchema(spec, itemTsType);
+        // We don't want a Function type in the final spec.
+        delete schema.items[TS_TYPE_KEY];
```
bajtos: AFAICT, these three lines (or even more than three) are copy&pasted from the case when schema is not an array. Can we share this code please?

Also is it possible for an array to contain another nested array? E.g. an array of arrays of x-ts-type schemas.

```yaml
schema:
 type: array
 items: 
   type: array
   items:
    x-ts-type: Number
```

I think it would be best if we could recursively call `resolveTSType` to handle items.

```ts
resolveTSType(spec, schema.items);
```

packages/openapi-v3/test/integration/controller-spec.integration.ts
```diff
@@ -219,7 +219,48 @@ describe('controller spec', () => {
     });
   });
 
-  it('generates schema for an array from `x-ts-type`', () => {
+  @model()
```
bajtos: I find it rather unexpected to encounter a helper model being defined in the middle of the test cases. Please move the class definition to the bottom of this `describe` block.

raymondfeng:
@bajtos I have improved the PR per your comments. PTAL.

PR #1912, opened 2018-10-25
fix(cli): allow other connectors to be used for repositories

raymondfeng:
The problem was reported by Sam Ruby.

1. `lb4 datasource` to add a DS with `loopback-connector-sqlite3`
2. `lb4 repository` fails to list `loopback-connector-sqlite3` as it's NOT on the list of connectors we pull from `loopback-workspace`

## Checklist

### DIFF
--- a/packages/cli/generators/datasource/index.js
+++ b/packages/cli/generators/datasource/index.js
@@ -134,7 +134,7 @@ module.exports = class DataSourceGenerator extends ArtifactGenerator {
       const prompts = [
         {
           name: 'customConnector',
-          message: "Enter the connector's module name",
+          message: "Enter the connector's package name:",
           validate: utils.validate,
         },
       ];
@@ -309,6 +309,10 @@ module.exports = class DataSourceGenerator extends ArtifactGenerator {
       }
 
       debug(`npmModule - ${pkgs[0]}`);
+    } else {
+      const connectorName = this.artifactInfo.connector;
+      // Other connectors
+      if (!deps[connectorName]) pkgs.push(connectorName);
     }
 
     if (!deps['@loopback/repository']) {
--- a/packages/cli/test/integration/generators/repository.integration.js
+++ b/packages/cli/test/integration/generators/repository.integration.js
@@ -31,7 +31,7 @@ describe('lb4 repository', function() {
 
   // special cases regardless of the repository type
   describe('generate repositories on special conditions', () => {
-    it('generates multipe crud repositories', async () => {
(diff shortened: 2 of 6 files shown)


packages/cli/generators/datasource/index.js
```diff
@@ -134,7 +134,7 @@ module.exports = class DataSourceGenerator extends ArtifactGenerator {
       const prompts = [
         {
           name: 'customConnector',
-          message: "Enter the connector's module name",
+          message: "Enter the connector's module name:",
```
jannyHou: Would it be better to emphasize "enter the FULL name"?
I remember in lb3 w/o the prefix `loopback-connector` confused people.

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -289,6 +289,29 @@ describe('lb4 repository', function() {
       );
     });
 
+    it('allows other connectors', async () => {
```
jannyHou: maybe add some check to verify the custom connector is installed or get added into the package.json file?

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -289,6 +289,29 @@ describe('lb4 repository', function() {
...
+          .executeGenerator(generator)
+          .inDir(SANDBOX_PATH, () =>
+            testUtils.givenLBProject(SANDBOX_PATH, {
+              // Only use the sqlite3 datasource
+              additionalFiles: files,
+            }),
+          )
+          .withPrompts(basicPrompt)
+          .withArguments(' --model Defaultmodel'),
+      ).to.not.be.rejectedWith(/No datasources found/);
```
jannyHou: I am wondering if the command exits with another error, would the test pass?
If yes I think omit the message here is better than specify it.

raymondfeng:
> What version spec are you going to use for custom connectors? Could you please add a test to verify what is written to package.json?

Without a version, we run `npm i loopback-connector-sqlite3`, which should brings the latest version.

> Ideally, we should consult the npm registry to find the version tagged as latest and perhaps fall back to asking the user to tell us the version they would like to use.

Maybe we should not bother as the developer can always override it in package.json using `npm i`. Unless we go to full extent to discover connectors from npm.

bajtos:
> Without a version, we run npm i loopback-connector-sqlite3, which should brings the latest version.

IIUC, you are saying we are not adding the dependency directly to `package.json`, but rely on `npm install` to add it? I suppose that's good enough for now 👍

PR #1915, opened 2018-10-25
 style(boot): remove workaround warning comment 

Yaty:
https://github.com/nodejs/node/issues/20278 was solved so Object.entries/values can now be used.

## Checklist

### DIFF
--- a/packages/boot/src/booters/booter-utils.ts
+++ b/packages/boot/src/booters/booter-utils.ts
@@ -44,6 +44,7 @@ export function isClass(target: any): target is Constructor<any> {
  * and then testing each exported member to see if it's a class or not.
  *
  * @param files An array of string of absolute file paths
+ * @param projectRootDir The project root directory
  * @returns {Constructor<{}>[]} An array of Class constructors from a file
  */
 export function loadClassesFromFiles(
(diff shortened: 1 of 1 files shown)


packages/boot/src/booters/booter-utils.ts
```diff
@@ -54,10 +55,7 @@ export function loadClassesFromFiles(
   for (const file of files) {
     debug('Loading artifact file %j', path.relative(projectRootDir, file));
     const moduleObj = require(file);
-    // WORKAROUND: use `for in` instead of Object.values().
-    // See https://github.com/nodejs/node/issues/20278
-    for (const k in moduleObj) {
-      const exported = moduleObj[k];
+    for (const [k, exported] of Object.entries(moduleObj)) {
```
raymondfeng: `for-in` loop is fine for this usage. I don't think we need to go back to `Object.entries`.

Yaty: Ok ! So what about removing the comment as it is now resolved in V8 ? This is what confused me in the first place.

Yaty:
Ooops formatOnSave was activated, wip :) ...

PR #1916, opened 2018-10-25
docs: add link to cloudant dashboard

mhdawson:
Add link to cloudant dashboard in example. I had to google how to access the dashboard so
having the link would be useful.

## Checklist

- [X] Documentation in [/docs/site](../tree/master/docs/site) was updated


mhdawson:
@raymondfeng done.  I guess I should have looked more closely after the timeout when I ran npm test as I guess it did not run the following steps.

PR #929, opened 2018-01-30
[WIP] feat(context): Allow components to expose a list of bindings

raymondfeng:
The PR adds a few APIs:

- Context.bind(Binding) - bind an binding to the current context.
- Component.classes and Component.bindings - allow a component to declare an array of bindings or a map of classes.

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] Related API Documentation was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `packages/example-*` were updated

PR #1924, opened 2018-10-26
feat(core): improve component bindings

raymondfeng:
Reactivation of https://github.com/strongloop/loopback-next/pull/929

- Improve the component design to allow contribution of various bindings
- Introduce a `@component` decorator to mark a component class (inspired by https://docs.nestjs.com/modules)

A component class can be declared as follows:
```ts
     @component({
      controllers: [MyController],
      bindings: [binding],
      classes: {'my-class': MyClass},
      providers: {'my-provider': MyProvider},
    })
    class MyComponent implements Component {
      bindings = [];
      constructor() {
        // Dynamically add more bindings
        this.bindings.push(...);
      }
    }
```
## Checklist

### DIFF
--- a/docs/site/Creating-components.md
+++ b/docs/site/Creating-components.md
@@ -24,8 +24,8 @@ export class MyComponent implements Component {
       'my-validator': MyValidator,
     };
 
-    const bindingX = new Binding('x').to('Value X');
-    const bindingY = new Binding('y').toClass(ClassY);
+    const bindingX = Binding.bind('x').to('Value X');
+    const bindingY = Binding.bind('y').toClass(ClassY);
     this.bindings = [bindingX, bindingY];
   }
 }
@@ -45,7 +45,7 @@ Please note that `providers` and `classes` are shortcuts for provider and class
 `bindings`.
 
 The example `MyComponent` above will add `MyController` to application's API and
-create the following bindings to the application context:
+create the following bindings in the application context:
 
 - `my-value` -> `MyValueProvider` (provider)
 - `my-validator` -> `MyValidator` (class)
--- a/docs/site/Creating-components.md
+++ b/docs/site/Creating-components.md
@@ -20,6 +20,13 @@ export class MyComponent implements Component {
     this.providers = {
       'my-value': MyValueProvider,
     };
+    this.classes = {
+      'my-validator': MyValidator,
+    };
+
+    const bindingX = new Binding('x').to('Value X');
+    const bindingY = new Binding('y').toClass(ClassY);
+    this.bindings = [bindingX, bindingY];
   }
 }
 ```
@@ -28,10 +35,22 @@ When a component is mounted to an application, a new instance of the component
 class is created and then:
 
 - Each Controller class is registered via `app.controller()`,
-- Each Provider is bound to its key in `providers` object.
+- Each Provider is bound to its key in `providers` object via
+  `app.bind(key).toProvider(providerClass)`
+- Each Class is bound to its key in `classes` object via
+  `app.bind(key).toClass(cls)`
+- Each Binding is added via `app.add(binding)`
+
+Please note that `providers` and `classes` are shortcuts for provider and class
+`bindings`.
 
 The example `MyComponent` above will add `MyController` to application's API and
-create a new binding `my-value` that will be resolved using `MyValueProvider`.
+create the following bindings to the application context:
+
+- `my-value` -> `MyValueProvider` (provider)
+- `my-validator` -> `MyValidator` (class)
+- `x` -> `'Value X'` (value)
+- `y` -> `ClassY` (class)
 
 ## Providers
 
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -42,16 +42,26 @@ export class Context {
    * Create a binding with the given key in the context. If a locked binding
    * already exists with the same key, an error will be thrown.
    *
-   * @param key Binding key
+   * @param keyOrBinding Binding key or a binding
    */
   bind<ValueType = BoundValue>(
-    key: BindingAddress<ValueType>,
+    keyOrBinding: BindingAddress<ValueType> | Binding,
   ): Binding<ValueType> {
+    let key: string;
+    let binding: Binding<ValueType>;
+    if (keyOrBinding instanceof Binding) {
+      key = keyOrBinding.key;
+      binding = keyOrBinding;
+    } else {
+      key = keyOrBinding.toString();
+      binding = new Binding<ValueType>(key);
+    }
+
     /* istanbul ignore if */
     if (debug.enabled) {
       debug('Adding binding: %s', key);
     }
-    key = BindingKey.validate(key);
+
     const keyExists = this.registry.has(key);
     if (keyExists) {
       const existingBinding = this.registry.get(key);
@@ -59,8 +69,6 @@ export class Context {
       if (bindingIsLocked)
         throw new Error(`Cannot rebind key "${key}" to a locked binding`);
     }
-
-    const binding = new Binding<ValueType>(key);
     this.registry.set(key, binding);
     return binding;
   }
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -42,21 +42,23 @@ export class Context {
    * Create a binding with the given key in the context. If a locked binding
    * already exists with the same key, an error will be thrown.
    *
-   * @param keyOrBinding Binding key or a binding
+   * @param key Binding key
    */
   bind<ValueType = BoundValue>(
-    keyOrBinding: BindingAddress<ValueType> | Binding,
+    key: BindingAddress<ValueType>,
   ): Binding<ValueType> {
-    let key: string;
-    let binding: Binding<ValueType>;
-    if (keyOrBinding instanceof Binding) {
-      key = keyOrBinding.key;
-      binding = keyOrBinding;
-    } else {
-      key = keyOrBinding.toString();
-      binding = new Binding<ValueType>(key);
-    }
+    const binding = new Binding<ValueType>(key.toString());
+    this.add(binding);
+    return binding;
+  }
 
+  /**
+   * Add a binding to the context. If a locked binding already exists with the
+   * same key, an error will be thrown.
+   * @param binding The configured binding to be added
+   */
+  add<ValueType = BoundValue>(binding: Binding<ValueType>): this {
+    const key = binding.key;
     /* istanbul ignore if */
     if (debug.enabled) {
       debug('Adding binding: %s', key);
@@ -70,7 +72,7 @@ export class Context {
         throw new Error(`Cannot rebind key "${key}" to a locked binding`);
     }
     this.registry.set(key, binding);
-    return binding;
+    return this;
   }
 
   /**
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -57,7 +57,7 @@ export class Context {
    * same key, an error will be thrown.
    * @param binding The configured binding to be added
    */
-  add<ValueType = BoundValue>(binding: Binding<ValueType>): this {
+  add(binding: Binding<unknown>): this {
     const key = binding.key;
     /* istanbul ignore if */
     if (debug.enabled) {
@@ -85,7 +85,7 @@ export class Context {
    * @param key Binding key
    * @returns true if the binding key is found and removed from this context
    */
-  unbind<ValueType = BoundValue>(key: BindingAddress<ValueType>): boolean {
+  unbind(key: BindingAddress<unknown>): boolean {
     key = BindingKey.validate(key);
     const binding = this.registry.get(key);
     if (binding == null) return false;
@@ -99,7 +99,7 @@ export class Context {
    * delegating to the parent context
    * @param key Binding key
    */
-  contains<ValueType = BoundValue>(key: BindingAddress<ValueType>): boolean {
+  contains(key: BindingAddress<unknown>): boolean {
     key = BindingKey.validate(key);
     return this.registry.has(key);
   }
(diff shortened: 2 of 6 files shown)


bajtos:
> Improve the component design to allow contribution of various bindings

+1, this part makes sense to me. Can we have a dedicated pull request for this feature please? We should discuss documentation updates, descriptive API naming, etc. as part of the review.

> Introduce a `@component` decorator to mark a component class

I don't understand the benefits of this change. I believe the example shown above can be easily rewritten without `@component` decorator as follows:

```ts
    class MyComponent implements Component {
      controllers: [MyController],
      bindings = [binding];
      classes: {'my-class': MyClass},
      providers: {'my-provider': MyProvider},

      constructor() {
        // Dynamically add more bindings
        this.bindings.push(...);
      }
    }
```

What benefits do you @raymondfeng see in the `@component()` based solution? 

I don't want to introduce two ways how to accomplish the same thing, because that usually confuse users. Let's pick one way that we think is the best, make it easy to use and well documented.

BTW I think the current approach for building components is too cumbersome and we should look into ways how to leverage `@loopback/boot` to load component artifacts based on conventions. For example:

```
// provided by LB
class ComponentBase {
  // implements Application-like APIs for registering artifacts
  controller(ctor) {
    this.controllers.push(ctor);
  }
}

// in an extension
import {boot} from '@loopback/boot';

class MyExtension extends BootMixin(RepositoryMixin(ServiceMixin(ComponentBase)) {
  boot() {
    boot(this, {/* boot options */});
  }
}
```

As long as both `Application` and `ComponentBase` provide the same API for registering artifacts, we should be able to use the same Booter implementations for both applications and components. How awesome would that be for extension developers!

Let me open a new issue to discuss that idea: https://github.com/strongloop/loopback-next/issues/1968

bajtos:
As I was thinking about this some more and in the light of https://github.com/strongloop/loopback-next/issues/1968, I think it would be best if the code example shown in the PR description was instead written this way:

```ts
    class MyComponent extends Component {
      constructor() {
        this.controller(MyController);
        this.bind('foo').to('bar');
        this.bind('my-class').toClass(MyClass);
        this.bind('my-provider').toProvider(MyProvider);
     }
  }
```

I see few important benefits in such approach:
 - Extension developers can use the APIs they are already familiar with from building Applications. Less APIs to learn make LB users more productive.
 - I find the originally proposed extension properties like `classes` and `providers` difficult to understand. In my proposal, we reuse the existing Binding APIs - no need to invent new terminology.
 - By leveraging existing Application and Binding APIs, we have less new APIs to document, test and maintain for the future.
 - Later on, we can leverage `@loopback/boot` to load and register artifacts from source files automatically, based on project layout conventions.

raymondfeng:
@bajtos I think the different options boil down to `declarative` vs `imperative`. I like the new style of using `Component` as a base class. I assume the `Component` itself is `Context`.

IIRC, I proposed before that each component becomes a context to group its contributions. If that happens, the application will becomes:

- app (Application extends Component)
- rest component
- explorer component

We also need to find a way to aggregate all contexts, either by merging or delegating.

bajtos:
> I think the different options boil down to `declarative` vs `imperative`. I like the new style of using `Component` as a base class. I assume the `Component` itself is `Context`.
>
> IIRC, I proposed before that each component becomes a context to group its contributions. If that happens, the application will becomes:
>
> - app (Application extends Component)
> - rest component
> - explorer component
>
> We also need to find a way to aggregate all contexts, either by merging or delegating.

I don't think it's a good idea for Component to be also a Context:
 - IMO, merging/aggregation of multiple context hierarchies is a complex problem that's difficult to implement correctly (think about multiple inheritance and the diamond problem). We have discussed this in the PR where you proposed an implementation merging contexts.
 - By giving Components access to APIs for binding resolution (e.g. `ctx.get()`), we make it easy for developers to shoot into their own foot. Typically, the Component context need to inherit bindings like `CoreBinding.APPLICATION_INSTANCE` from the Application context. Until the Component context is attached to the Application, these bindings cannot be resolved though. I expect there may be more similar edge cases.

Here is what I am proposing:

(1) Let's keep the contract between Application and Component simple. The component exports an array of `Binding` instances. As I see it, this is the simplest solution that provides all the power components may need. In my experience, simpler is usually better.

(2) Give us and the community freedom to experiment with different ways of building/setting-up components. Let the solution described in my comment above (and little bit later here as well) be only one of the possible ways.

(3) As for my proposal: my vision is to conceptually extract Context + Application APIs into two groups:
 1. APIs for defining/registering new bindings: `ctx.bind()`, `app.controller()`, etc.
 2. APIs for consuming/resolving bindings: `ctx.get()`, `@inject`, etc.

So far, `app.controller()` and friends are just sugar methods wrapping `ctx.bind()`. Let's extract them from the Application class into a new mixin.

As for components, let's implement a base component class that provides `bind()` API required by `controller()` but also `repository()`, `dataSource` and other helpers contributed by mixins. Under the hood, `bind()` can push a new binding to `this.bindings[]` array instead of using Context-based registration.

```ts
interface Component {
  // ...
  bindings?: Binding[];
}

interface BindingBuilder {
  bind(key: string): Binding;
}

function AppArtifactRegistrationMixin<T extends Class<any>>(superClass: T) {
  return class extends superClass {
    controller(controllerCtor: ControllerClass, name?: string): Binding {
      name = name || controllerCtor.name;
      return this.bind(`controllers.${name}`)
        .toClass(controllerCtor)
        .tag('controller');
    }
    // etc.
  };
}

class ComponentBindingBuilder implements BindingBuilder {
  bindings: Binding[] = [];

  bind(key: string): Binding {
    const binding = new Binding(key);
    this.bindings.push(binding);
    return binding;
  }
}

class ComponentBase implements Component
  extends AppArtifactRegistrationMixin(ComponentBindingBuilder) {
}

// and also in the Application class:
class Application extends AppArtifactRegistrationMixin(Context) {
  // ...
}
```

bajtos:
> BTW, we can do something similar today:
> ```ts
> export class MyComponent {
>      constructor(@inject(CoreBinding.APPLICATION_INSTANCE) app: Application) {
> //...
> ```

Honestly, this seems a bit hacky to me, I consider this as a (hopefully short-term) workaround until we enhance Component interface to support all kinds of artifacts.

On the other hand, maybe this is a sign that we need to revisit Component design? In the current design, whenever we add a new artifact to Application/RestApplication/etc. (e.g. static assets), we also need to find a way how to enable components to contribute these artifacts too.

In the example you have shown, the components are manipulating the application directly, thus there is no need to enhance the Component API contract to support new features.

I guess it boils down to _declarative vs imperative_ as you wrote yourself. 

If we feel that declarative support is something we should support, then I think we should improve our design to make this more obvious/easier to use. For example, the component constructor can receive the Application instance directly in the constructor without any DI.

This raises a question though: do we want to support both declarative and imperative approaches to defining components? Personally, I prefer to choose only one approach, to keep the maintenance costs low. If we support both approaches, then we have more documentation to write and maintain, more questions to answer because people will be confused about which approach to choose, etc.

@strongloop/loopback-maintainers thoughts? (See also my previous comment proposing a solution how to allow extensions to contribute arbitrary bindings.)

bajtos:
@raymondfeng To move this pull request forward: I think we both agree that it's a good idea to allow components to export `bindings: Binding[]`. Let's land this feature first, in isolation from the other parts proposed in this pull request, and then discuss how to continue building on top of `bindings: Binding[]`.

raymondfeng:
@bajtos I reduced the scope of this PR to add `bindings` support. PTAL.

packages/context/src/context.ts
```diff
@@ -47,22 +47,32 @@ export class Context {
...
+    this.add(binding);
+    return binding;
+  }
+
+  /**
+   * Add a binding to the context. If a locked binding already exists with the
+   * same key, an error will be thrown.
+   * @param binding The configured binding to be added
+   */
+  add<ValueType = BoundValue>(binding: Binding<ValueType>): this {
```
bajtos: A nitpick: I think `= BoundValue` is not needed, because the specialization of `add` can be always inferred by the compiler from the specialization of the `binding` argument.

packages/context/src/context.ts
```diff
@@ -47,22 +47,32 @@ export class Context {
...
+    const key = binding.key;
     /* istanbul ignore if */
     if (debug.enabled) {
       debug('Adding binding: %s', key);
     }
-    key = BindingKey.validate(key);
```
bajtos: IIUC, this validation can be removed because `Binding`'s constructor has already validated the key 👍

packages/core/src/component.ts
```diff
@@ -27,13 +31,21 @@ export interface Component {
    * A map of name/class pairs for binding providers
    */
   providers?: ProviderMap;
+
+  classes?: ClassMap;
```
bajtos: I find this name confusing. What does it mean for a component to export a class? Some classes are exported via JavaScript/TypeScript exports, e.g. `export class FooBar`. Why should be some classes exported via JS/TS and some others exported via Component properties?

Personally, I prefer to remove this API entirely, keep `providers` for backwards compatibility (possibly even deprecate it) and ask users to leverage `providers` API for all bindings.

If you feel it's important to keep this short-cut, then let's find a better name please. For example:`classBindings` or `classesToBind`.

packages/core/src/component.ts
```diff
@@ -49,15 +61,33 @@ export interface Component {
...
+      const binding = Binding.bind(classKey).toClass(
+        component.classes[classKey],
+      );
+      app.add(binding);
     }
   }
 
   if (component.providers) {
     for (const providerKey in component.providers) {
-      app.bind(providerKey).toProvider(component.providers[providerKey]);
```
bajtos: I don't understand why is this change necessary? What's wrong with the old implementation?

packages/core/test/unit/application.unit.ts
```diff
@@ -35,9 +42,26 @@ describe('Application', () => {
...
+      }
+    }
     class MyComponent implements Component {
       controllers = [MyController];
+      bindings = [binding];
+      classes = {'my-class': MyClass};
+      providers = {'my-provider': MyProvider};
+    }
+
+    class MyComponentWithDI implements Component {
```
bajtos: Please define this class inside the test that is using it (i.e. `it('binds from a component constructor'`) to avoid future temptation to add more stuff into this class.

packages/core/test/unit/application.unit.ts
```diff
@@ -35,9 +42,26 @@ describe('Application', () => {
...
+    class MyProvider implements Provider<string> {
+      value() {
+        return 'my-str';
+      }
+    }
     class MyComponent implements Component {
       controllers = [MyController];
+      bindings = [binding];
+      classes = {'my-class': MyClass};
+      providers = {'my-provider': MyProvider};
```
bajtos: Instead of treating `MyComponent` as a sort of a kitchen sink, please create a minimal component class in each test case.

For example:

```ts
    it('binds providers from a component', () => {
       class ComponentWithProvider implements Component {
         providers = {'my-provider': MyProvider};
       }

       app.component(ComponentWithProvider);

       expect(app.contains('my-provider')).to.be.true();
       expect(app.getSync('my-provider')).to.be.eql('my-str');
     });
```

My proposed design makes the tests much easier to understand and modify.

packages/core/src/component.ts
bajtos: Oops, I made a typo in my comment above.

```diff
- ask users to leverage `providers` API for all bindings.
+ ask users to leverage `bindings` API for all bindings.
```

raymondfeng:
@bajtos PTAL.

packages/core/src/component.ts
raymondfeng: I added jsdocs to make it clear.

bajtos: I am afraid jsdocs are not good enough. Consider the following code snippet from your pull request:

```ts
      class MyComponentWithClasses implements Component {
         classes = {'my-class': MyClass};
       }
```

The person reading the code (e.g. while reviewing a pull request on GitHub) does not see tscode comments.

packages/context/src/context.ts
bajtos: @raymondfeng PTAL - is the default value for `ValueType` needed?

raymondfeng:
I updated to docs and tsdocs.

I still prefer to keep `classes` to be consistent with `providers` and `controllers` as such concepts are relative to context bindings.

packages/context/src/context.ts
raymondfeng: I simplified it as `add(binding: Binding<unknown>): this {`. The same is also applied to a few other methods.

raymondfeng:
@bajtos PTAL

docs/site/Creating-components.md
```diff
@@ -28,10 +35,22 @@ When a component is mounted to an application, a new instance of the component
...
+- Each Class is bound to its key in `classes` object via
+  `app.bind(key).toClass(cls)`
+- Each Binding is added via `app.add(binding)`
+
+Please note that `providers` and `classes` are shortcuts for provider and class
+`bindings`.
 
 The example `MyComponent` above will add `MyController` to application's API and
-create a new binding `my-value` that will be resolved using `MyValueProvider`.
+create the following bindings to the application context:
```
bajtos: I am not sure if `create ... to` is grammatically correct. I think `create ... in` or `add ... to` would be better.
```suggestion
create the following bindings in the application context:
```

ping @bschrammIBM @dhmlau

packages/core/src/component.ts
```diff
@@ -23,17 +30,46 @@ export interface Component {
...
    * A map of name/class pairs for servers
    */
   servers?: {
     [name: string]: Constructor<Server>;
   };
 
+  /**
+   * An array of bindings to be aded to the application context. For example,
+   * ```ts
+   * const bindingX = new Binding('x').to('Value X');
```
bajtos: Let's use the newly added sugar API please.
```suggestion
   * const bindingX = Binding.bind('x').to('Value X');
```

Issue #1845, opened 2018-10-12
Allow components to contribute start actions


mrbatista:
When bootstrap LB4 application I need to execute/load some actions on startup.
In the example above I need to register job task.
The `CustomComponent` is added to `QueueJobApplication` through QueueMixin.
The `CustomComponent` throw error becase don't have the repository bound to the context.
Is it possible to add `QueueMixin` after the RepositoryMixin are loaded all repositories to context?

## Example:
```typescript
export class CustomComponent implements Component {
  constructor(
    @repository(QueueRepository) private queueRepository: QueueRepository,
  ) {
    this.registerProcess();
  }
}

export function QueueMixin<T extends Class<any>>(superClass: T) {
  return class extends superClass {
    // A mixin class has to take in a type any[] argument!
    // tslint:disable-next-line:no-any
    constructor(...args: any[]) {
      super(...args);
      this.component(CustomComponent);
    }
  }
}

export class QueueJobApplication extends BootMixin(
  QueueMixin(ServiceMixin(RepositoryMixin((RestApplication))),
) {
  constructor(options: ApplicationConfig = {}) {
    super(options);

    // Set up the custom sequence
    this.sequence(MySequence);
    this.projectRoot = __dirname;
  }
}
```

LB3 Example (script added under `server/boot`):
```javascript
module.export = app => {
  const {QueueRepository} = app.models;
  QueueRepository.once('dataSourceAdded', () => {
     // Execute/load some logic
  }); 
}
```

mrbatista:
For now I have loaded component after boot:

```typescript
import {ApplicationConfig} from '@loopback/core';
import {QueueJobApplication} from './application';
import { CustomComponent } from './components';

export {QueueJobApplication};

export async function main(options: ApplicationConfig = {}) {
  const app = new QueueJobApplication(options);
  await app.boot();

  app.component(CustomComponent);

  await app.start();

  const url = app.restServer.url;
  console.log(`Server is running at ${url}`);
  console.log(`Try ${url}/ping`);

  return app;
}
```

hacksparrow:
@mrbatista unless you are having any problems, loading the component after boot should be fine. You might want to check out [BootComponent](https://loopback.io/doc/en/lb4/Booting-an-Application.html#bootcomponent) too.

bajtos:
> When bootstrap LB4 application I need to execute/load some actions on startup.
In the example above I need to register job task.

[...]

Issue #1919, opened 2018-10-26
Spike: Operation hooks for models/repositories


bajtos:
In LoopBack 3.x, we have a concept of [Operation Hooks](https://loopback.io/doc/en/lb3/Operation-hooks.html) allowing models to register handlers for common events: access, before/after save, before/after delete, load/persist.

We should provide similar functionality in LoopBack 4 too.

See also the discussion in #1857 which triggered this feature request.

/cc @vvdwivedi @David-Mulder @marioestradarosa @raymondfeng 

## Acceptance criteria 
- [ ] A draft pull request that:
  - finds out the minimum infrastructure and what other use cases would use this hook
  - see if use cases can be handled by life cycle observers(for data events) or interceptors
  - high level proposal for the hooks in the repository level
  - finds out that which kind of hooks are more suitable for LB4: a set of hooks that is universal for all possible implementations of a given repository interface, or the hooks coupled with a particular repository implementation. ( e.g different sets of hooks for key/value and CRUD)

bajtos:
I have created this story as a Spike, because I feel we need to do a bit of research first.  For example, where to register hook handlers - at model level or at repository level? If the hooks are registered at model level, how can we ensure that all repository implementations are honoring them? Can we find a set of hooks that is universal for all possible implementations of a given repository interface, or are the hooks coupled with a particular repository implementation and/or interface? For example, I can imagine that a KeyValue model/repository needs a different set of hooks than a CRUD model/repository.

marioestradarosa:
👍 to implement them at the repository level.  Keep in mind that by design it is the repository the one that intermediates between both artifacts. Remember that I mentioned you in our conference call?. Usually the hooks will interact on **CRUD** operations on **Before/After** _events_, they will receive a record, they can mutate it and then call next hook. So it makes sense in the repository since these CRUD operations occur at this level.

On WANG os, we called them updateExits 😄.

David-Mulder:
My understanding of the new design was that it's perfectly okay for two repositories to access the same models (in the case of more complex relational data), would this mean that a third abstraction layer would need to be added to share the hooks between those repositories? Or is it bad design ifo two different repositories refer to the same model?

marioestradarosa:
>  would this mean that a third abstraction layer would need to be added to share the hooks between those repositories?

By design you shouldn't have more than one repository that belongs to the same datasource pointing to the same model if we use these hooks implementation, I can't figure out a real case scenario for now.  

[...]

PR #1928, opened 2018-10-26
feat(core): introduce basic life cycle support for LoopBack applications

raymondfeng:
Implements:
- https://github.com/strongloop/loopback-next/issues/1845
- https://github.com/strongloop/loopback-next/issues/2034

The PR implements the following features:

- Allow registration and actions of life cycle observers using context bindings - See https://github.com/strongloop/loopback-next/blob/lifecycle/docs/site/Life-cycle.md
- Discover and boot life cycle scripts
- Add `lb4 observer` command to generate life cycle scripts - See https://github.com/strongloop/loopback-next/blob/lifecycle/docs/site/Life-cycle-observer-generator.md

## Test `observer` cli

```sh
cd loopback-next
git checkout lifecycle
npm run build
cd sandbox
../packages/cli/bin/cli-main.js app life
cd life
../../packages/cli/bin/cli-main.js observer
cd ../..
npm i
cd sandbox/life
npm start
```

## Checklist

### DIFF
--- a/docs/site/Extension-life-cycle.md
+++ b/docs/site/Extension-life-cycle.md
@@ -1,0 +1,80 @@
+---
+lang: en
+title: 'Extension life cycle'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Extension-life-cycle.html
+---
+
+## Extension life cycle
+
+As described in [Life cycle](Life-cycle.md), a LoopBack
+[Application](Application.md) has its own life cycles at runtime. Corresponding
+events such as `start` and `stop` are emitted upon the state change. Please note
+that LoopBack only support `start` and `stop` events for an application's life
+cycles at this moment.
+
+Extension modules for LoopBack often contribute artifacts such as servers,
+datasources, and connectors to the application. They typically provide a
+component to bind such artifacts to the context together. Being able to listen
+on life cycle events is important for extension modules to collaborate with the
+application.
+
+An extension module follows the same way as applications to implement and
+register life cycle observers.
+
+### Implement a life cycle observer
+
+A life cycle observer class optionally implements `start` and `stop` methods to
+be invoked upon `start` and `stop` events emitted by an application's life cycle
+respectively.
+
+```ts
+import {LifeCycleObserver} from '@loopback/core';
+
+export class MyLifeCycleObserver implements LifeCycleObserver {
+  start() {
+    // It can return `void` or `Promise<void>`
+  }
+  stop() {
+    // It can return `void` or `Promise<void>`
+  }
+}
+```
+
+A life cycle observer can be tagged with `CoreTags.LIFE_CYCLE_OBSERVER_GROUP` to
+indicate its group to be invoked for ordering. We can decorate the observer
+class with `@lifeCycleObserver` to provide more metadata for the binding.
+
+```ts
+import {lifeCycleObserver} from '@loopback/core';
+
+@lifeCycleObserver('g1')
+export class MyLifeCycleObserver {
+  // ...
+}
+```
+
+### Register a life cycle observer
+
+A life cycle observer can be registered by calling `lifeCycleObserver()` of the
+application. It binds the observer to the application context with a special
+tag - `CoreTags.LIFE_CYCLE_OBSERVER`.
+
+```ts
+app.lifeCycleObserver(MyObserver);
+```
+
+Life cycle observers can be declared via a component class too. when the
+component is mounted to an application, the observers are automatically
+registered.
+
+```ts
+export class MyComponentWithObservers implements Component {
+  lifeCycleObservers = [XObserver, YObserver];
+}
+
+// Mount the component
+app.mount(MyComponentWithObservers);
+// Now `XObserver` and `YObserver` are registered in the application.
+```
--- a/docs/site/Life-cycle-observer-generator.md
+++ b/docs/site/Life-cycle-observer-generator.md
@@ -1,0 +1,83 @@
+---
+lang: en
+title: 'Life cycle observer generator'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Life-cycle-observer-generator.html
+---
+
+{% include content/generator-create-app.html lang=page.lang %}
+
+### Synopsis
+
+Adds a new [LifeCycleObserver](Life-cycle.md) class to a LoopBack application.
+
+```sh
+lb4 observer [--group <group>] [<name>]
+```
+
+### Arguments and options
+
+`<name>` - Required name of the observer to create as an argument to the
+command. If provided, the tool will use that as the default when it prompts for
+the name.
+
+`--group <group>` - Optional name of the observer group to sort the execution of
+observers by group.
+
+### Interactive Prompts
+
+The tool will prompt you for:
+
+- **Name of the observer.** _(observerName)_ If the name had been supplied from
+  the command line, the prompt is skipped.
+
+- **Group of the observer.** _(groupName)_ If the group had been supplied from
+  the command line, the prompt is skipped.
+
+### Output
+
+Once all the prompts have been answered, the CLI will do the following:
+
+- Create a LifeCycleObserver class as follows:
+  `/src/observers/${observerName}.observer.ts`
+- Update `/src/observers/index.ts` to export the newly created LifeCycleObserver
+  class.
+
+The generated class looks like:
+
+```ts
+import {
+  /* inject, Application, CoreBindings, */
+  lifeCycleObserver, // The decorator
+  CoreTags,
+  LifeCycleObserver, // The interface
+} from '@loopback/core';
+
+/**
+ * This class will be bound to the application as a `LifeCycleObserver` during
+ * `boot`
+ */
+@lifeCycleObserver('observer-group-name')
+export class HelloObserver implements LifeCycleObserver {
+  /*
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
+  ) {}
+  */
+
+  /**
+   * This method will be invoked when the application starts
+   */
+  async start(): Promise<void> {
+    // Add your logic for start
+  }
+
+  /**
+   * This method will be invoked when the application stops
+   */
+  async stop(): Promise<void> {
+    // Add your logic for start
+  }
+}
+```
--- a/docs/site/Life-cycle.md
+++ b/docs/site/Life-cycle.md
@@ -1,0 +1,282 @@
+---
+lang: en
+title: 'Life cycle events and observers'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Life-cycle.html
+---
+
+## Overview
+
+A LoopBack application has its own life cycles at runtime. There are two methods
+to control the transition of states of `Application`.
+
+- start(): Start the application
+- stop(): Stop the application
+
+It's often desirable for various types of artifacts to participate in the life
+cycles and perform related processing upon `start` and `stop`. Good examples of
+such artifacts are:
+
+- Servers
+
+  - start: Starts the HTTP server listening for connections.
+  - stop: Stops the server from accepting new connections.
+
+- Components
+
+  - A component can register life cycle observers
+
+- DataSources
+
+  - connect: Connect to the underlying database or service
+  - disconnect: Disconnect from the underlying database or service
+
+- Custom scripts
+  - start: Custom logic to be invoked when the application starts
+  - stop: Custom logic to be invoked when the application stops
+
+## The `LifeCycleObserver` interface
+
+To react on life cycle events, a life cycle observer implements the
+`LifeCycleObserver` interface.
+
+```ts
+import {ValueOrPromise} from '@loopback/context';
+
+/**
+ * Observers to handle life cycle start/stop events
+ */
+export interface LifeCycleObserver {
+  start?(): ValueOrPromise<void>;
+  stop?(): ValueOrPromise<void>;
+}
+```
+
+Please note all methods are optional so that an observer can opt in certain
+events. Each main events such as `start` and `stop` are further divided into
+three sub-phases to allow the multiple-step processing.
+
+## Register a life cycle observer
+
+A life cycle observer can be registered by calling `lifeCycleObserver()` of the
+application. It binds the observer to the application context with a special
+tag - `CoreTags.LIFE_CYCLE_OBSERVER`.
+
+```ts
+app.lifeCycleObserver(MyObserver);
+```
+
+Please note that `app.server()` automatically registers servers as life cycle
+observers.
+
+Life cycle observers can be registered via a component too:
+
+```ts
+export class MyComponentWithObservers {
+  lifeCycleObservers: [XObserver, YObserver];
+}
+```
+
+## Discover life cycle observers
+
+The `Application` finds all bindings tagged with `CoreTags.LIFE_CYCLE_OBSERVER`
+within the context chain and resolve them as observers to be notified.
+
+## Notify life cycle observers of start/stop related events by order
+
+There may be dependencies between life cycle observers and their order of
+processing for `start` and `stop` need to be coordinated. For example, we
+usually start a server to listen on incoming requests only after other parts of
+the application are ready to handle requests. The stop sequence is typically
+processed in the reverse order. To support such cases, we introduce
+two-dimension steps to control the order of life cycle actions.
+
+### Observer groups
+
+First of all, we allow each of the life cycle observers to be tagged with a
+group. For example:
+
+- datasource
+
+  - connect/disconnect
+  - mongodb
+  - mysql
+
+- server
+  - rest
+  - gRPC
+
+We can then configure the application to trigger observers group by group as
+configured by an array of groups in order such as `['datasource', 'server']`.
+Observers within the same group can be notified in parallel.
+
+For example,
+
+```ts
+app
+  .bind('observers.MyObserver')
+  .toClass(MyObserver)
+  .tag({
+    [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
+  })
+  .apply(asLifeCycleObserverBinding);
+```
+
+The observer class can also be decorated with `@bind` to provide binding
+metadata.
+
+```ts
+@bind(
+  {
+    tags: {
+      [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
+    },
+  },
+  asLifeCycleObserverBinding,
+)
+export class MyObserver {
+  // ...
+}
+
+app.add(createBindingFromClass(MyObserver));
+```
+
+The order of observers are controlled by a `groupsByOrder` property of
+`LifeCycleObserverRegistry`, which receives its options including the
+`groupsByOrder` from `CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS`. Thus the
+initial `groupsByOrder` can be set as follows:
+
+```ts
+app
+  .bind(CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS)
+  .to({groupsByOrder: ['g1', 'g2', 'server']});
+```
+
+Or:
+
+```ts
+const registry = await app.get(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY);
+registry.setGroupsByOrder(['g1', 'g2', 'server']);
+```
+
+Observers are sorted using `groupsByOrder` as the relative order. If an observer
+is tagged with a group that are not in `groupsByOrder`, it will come before any
+groups within `groupsByOrder`. Such custom groups are also sorted by their names
+alphabetically.
+
+In the example below, `groupsByOrder` is set to `['g1', 'g2']`. Given the
+following observers:
+
+- 'my-observer-1' ('g1')
+- 'my-observer-2' ('g2')
+- 'my-observer-4' ('2-custom-group')
+- 'my-observer-3' ('1-custom-group')
+
+The sorted observer groups will be:
+
+```ts
+{
+  '1-custom-group': ['my-observer-3'],
+  '2-custom-group': ['my-observer-4'],
+  'g1': ['my-observer-1'],
+  'g2': ['my-observer-2'],
+}
+```
+
+### Event phases
+
+It's also desirable for certain observers to do some processing before, upon, or
+after the `start` and `stop` events. To allow that, we notify each observer in
+three phases:
+
+- start: preStart, start, and postStart
+- stop: preStop, stop, and postStop
+
+Combining groups and event phases, it's flexible to manage multiple observers so
+that they can be started/stopped gracefully in order.
+
+For example, with a group order as `['datasource', 'server']` and three
+observers registered as follows:
+
+- datasource group: MySQLDataSource, MongoDBDataSource
+- server group: RestServer
+
+The start sequence will be:
+
+1. MySQLDataSource.preStart
+2. MongoDBDataSource.preStart
+3. RestServer.preStart
+
+4. MySQLDataSource.start
+5. MongoDBDataSource.start
+6. RestServer.start
+
+7. MySQLDataSource.postStart
+8. MongoDBDataSource.postStart
+9. RestServer.postStart
+
+## Add custom life cycle observers by convention
+
+Each application can have custom life cycle observers to be dropped into
+`src/observers` folder as classes implementing `LifeCycleObserver`.
+
+During application.boot(), such artifacts are discovered, loaded, and bound to
+the application context as life cycle observers. This is achieved by a built-in
+`LifeCycleObserverBooter` extension.
+
+## CLI command
+
+To make it easy for application developers to add custom life cycle observers,
+we introduce `lb4 observer` command as part the CLI.
+
+To add a life cycle observer:
+
+1. cd <my-loopback4-project>
+2. lb4 observer
+
+```
+? Observer name: Hello
+   create src/observers/my.hello-observer.ts
+   update src/observers/index.ts
+
+Observer Hello was created in src/observers/
+```
+
+The generated class looks like:
+
+```ts
+import {bind} from '@loopback/context';
+import {
+  /* inject, Application, */
+  CoreBindings,
+  LifeCycleObserver,
+} from '@loopback/core';
+
+/**
+ * This class will be bound to the application as a `LifeCycleObserver` during
+ * `boot`
+ */
+@bind({tags: {[CoreBindings.LIFE_CYCLE_OBSERVER_GROUP]: ''}})
+export class HelloObserver implements LifeCycleObserver {
+  /*
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
+  ) {}
+  */
+
+  /**
+   * This method will be invoked when the application starts
+   */
+  async start(): Promise<void> {
+    // Add your logic for start
+  }
+
+  /**
+   * This method will be invoked when the application stops
+   */
+  async stop(): Promise<void> {
+    // Add your logic for start
+  }
+}
+```
--- a/docs/site/Life-cycle.md
+++ b/docs/site/Life-cycle.md
@@ -25,7 +25,8 @@ such artifacts are:
 
 - Components
 
-  - A component can register life cycle observers
+  - A component itself can be a life cycle observer and it can also contribute
+    life cycle observers
 
 - DataSources
 
@@ -53,9 +54,8 @@ export interface LifeCycleObserver {
 }
 ```
 
-Please note all methods are optional so that an observer can opt in certain
-events. Each main events such as `start` and `stop` are further divided into
-three sub-phases to allow the multiple-step processing.
+Both `start` and `stop` methods are optional so that an observer can opt in
+certain events.
 
 ## Register a life cycle observer
 
@@ -73,8 +73,12 @@ observers.
 Life cycle observers can be registered via a component too:
 
 ```ts
-export class MyComponentWithObservers {
-  lifeCycleObservers: [XObserver, YObserver];
+export class MyComponentWithObservers implements Component {
+  /**
+   * Populate `lifeCycleObservers` per `Component` interface to register life
+   * cycle observers
+   */
+  lifeCycleObservers = [XObserver, YObserver];
 }
 ```
 
@@ -97,9 +101,8 @@ two-dimension steps to control the order of life cycle actions.
 First of all, we allow each of the life cycle observers to be tagged with a
 group. For example:
 
-- datasource
+- datasource (connect/disconnect)
 
-  - connect/disconnect
   - mongodb
   - mysql
 
@@ -109,7 +112,6 @@ group. For example:
 
 We can then configure the application to trigger observers group by group as
 configured by an array of groups in order such as `['datasource', 'server']`.
-Observers within the same group can be notified in parallel.
 
 For example,
 
@@ -120,20 +122,23 @@ app
   .tag({
     [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
   })
-  .apply(asLifeCycleObserverBinding);
+  .apply(asLifeCycleObserver);
 ```
 
 The observer class can also be decorated with `@bind` to provide binding
 metadata.
 
 ```ts
+import {bind, createBindingFromClass} from '@loopback/context';
+import {CoreTags, asLifeCycleObserver} from '@loopback/core';
+
 @bind(
   {
     tags: {
       [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
     },
   },
-  asLifeCycleObserverBinding,
+  asLifeCycleObserver,
 )
 export class MyObserver {
   // ...
@@ -142,34 +147,65 @@ export class MyObserver {
 app.add(createBindingFromClass(MyObserver));
 ```
 
-The order of observers are controlled by a `groupsByOrder` property of
+Or even simpler with `@lifeCycleObserver`:
+
+```ts
+import {createBindingFromClass} from '@loopback/context';
+import {lifeCycleObserver} from '@loopback/core';
+
+@lifeCycleObserver('g1')
+export class MyObserver {
+  // ...
+}
+
+app.add(createBindingFromClass(MyObserver));
+```
+
+The order of observers is controlled by a `orderedGroups` property of
 `LifeCycleObserverRegistry`, which receives its options including the
-`groupsByOrder` from `CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS`. Thus the
-initial `groupsByOrder` can be set as follows:
+`orderedGroups` from `CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS`.
+
+```ts
+export type LifeCycleObserverOptions = {
+  /**
+   * Control the order of observer groups for notifications. For example,
+   * with `['datasource', 'server']`, the observers in `datasource` group are
+   * notified before those in `server` group during `start`. Please note that
+   * observers are notified in the reverse order during `stop`.
+   */
+  orderedGroups: string[];
+  /**
+   * Notify observers of the same group in parallel, default to `true`
+   */
+  parallel?: boolean;
+};
+```
+
+Thus the initial `orderedGroups` can be set as follows:
 
 ```ts
 app
   .bind(CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS)
-  .to({groupsByOrder: ['g1', 'g2', 'server']});
+  .to({orderedGroups: ['g1', 'g2', 'server']});
 ```
 
 Or:
 
 ```ts
 const registry = await app.get(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY);
-registry.setGroupsByOrder(['g1', 'g2', 'server']);
+registry.setOrderedGroups(['g1', 'g2', 'server']);
 ```
 
-Observers are sorted using `groupsByOrder` as the relative order. If an observer
-is tagged with a group that are not in `groupsByOrder`, it will come before any
-groups within `groupsByOrder`. Such custom groups are also sorted by their names
-alphabetically.
+Observers are sorted using `orderedGroups` as the relative order. If an observer
+is tagged with a group that is not defined in `orderedGroups`, it will come
+before any groups included in `orderedGroups`. Such custom groups are also
+sorted by their names alphabetically.
 
-In the example below, `groupsByOrder` is set to `['g1', 'g2']`. Given the
-following observers:
+In the example below, `orderedGroups` is set to
+`['setup-servers', 'publish-services']`. Given the following observers:
 
-- 'my-observer-1' ('g1')
-- 'my-observer-2' ('g2')
+- 'my-observer-1' ('setup-servers')
+- 'my-observer-2' ('publish-services')
 - 'my-observer-4' ('2-custom-group')
 - 'my-observer-3' ('1-custom-group')
 
@@ -177,44 +213,20 @@ The sorted observer groups will be:
 
 ```ts
 {
-  '1-custom-group': ['my-observer-3'],
-  '2-custom-group': ['my-observer-4'],
-  'g1': ['my-observer-1'],
-  'g2': ['my-observer-2'],
+  '1-custom-group': ['my-observer-3'], // by alphabetical order
+  '2-custom-group': ['my-observer-4'], // by alphabetical order
+  'setup-servers': ['my-observer-1'], // by orderedGroups
+  'publish-services': ['my-observer-2'], // orderedGroups
 }
 ```
 
-### Event phases
+The execution order of observers within the same group is controlled by
+`LifeCycleObserverOptions.parallel`:
 
-It's also desirable for certain observers to do some processing before, upon, or
-after the `start` and `stop` events. To allow that, we notify each observer in
-three phases:
-
-- start: preStart, start, and postStart
-- stop: preStop, stop, and postStop
-
-Combining groups and event phases, it's flexible to manage multiple observers so
-that they can be started/stopped gracefully in order.
-
-For example, with a group order as `['datasource', 'server']` and three
-observers registered as follows:
-
-- datasource group: MySQLDataSource, MongoDBDataSource
-- server group: RestServer
-
-The start sequence will be:
-
-1. MySQLDataSource.preStart
-2. MongoDBDataSource.preStart
-3. RestServer.preStart
-
-4. MySQLDataSource.start
-5. MongoDBDataSource.start
-6. RestServer.start
-
-7. MySQLDataSource.postStart
-8. MongoDBDataSource.postStart
-9. RestServer.postStart
+- `true` (default): observers within the same group are notified in parallel
+- `false`: observers within the same group are notified one by one. The order is
+  not defined. If you want to have one to be invoked before the other, mark them
+  with two distinct groups.
 
 ## Add custom life cycle observers by convention
 
@@ -225,58 +237,20 @@ During application.boot(), such artifacts are discovered, loaded, and bound to
 the application context as life cycle observers. This is achieved by a built-in
 `LifeCycleObserverBooter` extension.
 
-## CLI command
+## CLI command to generate life cycle observers
 
 To make it easy for application developers to add custom life cycle observers,
 we introduce `lb4 observer` command as part the CLI.
 
-To add a life cycle observer:
-
-1. cd <my-loopback4-project>
-2. lb4 observer
-
-```
-? Observer name: Hello
-   create src/observers/my.hello-observer.ts
+```sh
+$ lb4 observer
+? Observer name: test
+? Observer group: g1
+   create src/observers/test.observer.ts
    update src/observers/index.ts
 
-Observer Hello was created in src/observers/
+Observer test was created in src/observers/
 ```
 
-The generated class looks like:
-
-```ts
-import {bind} from '@loopback/context';
-import {
-  /* inject, Application, */
-  CoreBindings,
-  LifeCycleObserver,
-} from '@loopback/core';
-
-/**
- * This class will be bound to the application as a `LifeCycleObserver` during
- * `boot`
- */
-@bind({tags: {[CoreBindings.LIFE_CYCLE_OBSERVER_GROUP]: ''}})
-export class HelloObserver implements LifeCycleObserver {
-  /*
-  constructor(
-    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
-  ) {}
-  */
-
-  /**
-   * This method will be invoked when the application starts
-   */
-  async start(): Promise<void> {
-    // Add your logic for start
-  }
-
-  /**
-   * This method will be invoked when the application stops
-   */
-  async stop(): Promise<void> {
-    // Add your logic for start
-  }
-}
-```
+See [Life cycle observer generator](Life-cycle-observer-generator.md) for more
+details.
--- a/packages/boot/src/__tests__/fixtures/lifecycle-observer.artifact.ts
+++ b/packages/boot/src/__tests__/fixtures/lifecycle-observer.artifact.ts
@@ -1,0 +1,30 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/boot
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {LifeCycleObserver} from '@loopback/core';
+
+/**
+ * An mock-up `LifeCycleObserver`. Please note that `start` and `stop` methods
+ * can be async or sync.
+ */
+export class MyLifeCycleObserver implements LifeCycleObserver {
+  status = '';
+
+  /**
+   * Handling `start` event asynchronously
+   */
+  async start() {
+    // Perform some work asynchronously
+    // await startSomeAsyncWork(...)
+    this.status = 'started';
+  }
+
+  /**
+   * Handling `stop` event synchronously.
+   */
+  stop() {
+    this.status = 'stopped';
+  }
+}
--- a/packages/cli/generators/observer/templates/observer-template.ts.ejs
+++ b/packages/cli/generators/observer/templates/observer-template.ts.ejs
@@ -1,0 +1,33 @@
+import {
+  /* inject, Application, CoreBindings, */
+  lifeCycleObserver, // The decorator
+  CoreTags,
+  LifeCycleObserver, // The interface
+} from '@loopback/core';
+
+/**
+ * This class will be bound to the application as a `LifeCycleObserver` during
+ * `boot`
+ */
+@lifeCycleObserver('<%= group %>')
+export class <%= className %> implements LifeCycleObserver {
+  /*
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
+  ) {}
+  */
+
+  /**
+   * This method will be invoked when the application starts
+   */
+  async start(): Promise<void> {
+    // Add your logic for start
+  }
+
+  /**
+   * This method will be invoked when the application stops
+   */
+  async stop(): Promise<void> {
+    // Add your logic for start
+  }
+}
--- a/packages/core/src/__tests__/unit/application-lifecycle.unit.ts
+++ b/packages/core/src/__tests__/unit/application-lifecycle.unit.ts
@@ -1,0 +1,235 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/core
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  bind,
+  BindingScope,
+  Constructor,
+  Context,
+  createBindingFromClass,
+} from '@loopback/context';
+import {expect} from '@loopback/testlab';
+import {
+  Application,
+  Component,
+  CoreBindings,
+  CoreTags,
+  LifeCycleObserver,
+  lifeCycleObserver,
+  Server,
+} from '../..';
+
+describe('Application life cycle', () => {
+  describe('start', () => {
+    it('starts all injected servers', async () => {
+      const app = new Application();
+      app.component(ObservingComponentWithServers);
+      const component = await app.get<ObservingComponentWithServers>(
+        `${CoreBindings.COMPONENTS}.ObservingComponentWithServers`,
+      );
+      expect(component.status).to.equal('not-initialized');
+      await app.start();
+      const server = await app.getServer(ObservingServer);
+
+      expect(server).to.not.be.null();
+      expect(server.listening).to.equal(true);
+      expect(component.status).to.equal('started');
+      await app.stop();
+    });
+
+    it('starts servers bound with `LIFE_CYCLE_OBSERVER` tag', async () => {
+      const app = new Application();
+      app
+        .bind('fake-server')
+        .toClass(ObservingServer)
+        .tag(CoreTags.LIFE_CYCLE_OBSERVER, CoreTags.SERVER)
+        .inScope(BindingScope.SINGLETON);
+      await app.start();
+      const server = await app.get<ObservingServer>('fake-server');
+
+      expect(server).to.not.be.null();
+      expect(server.listening).to.equal(true);
+      await app.stop();
+    });
+
+    it('starts/stops all registered components', async () => {
+      const app = new Application();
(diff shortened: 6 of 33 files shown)


davidxiao:
when will this PR be merged? :)

raymondfeng:
@bajtos Please review.

packages/core/src/lifecycle.ts
```diff
@@ -0,0 +1,23 @@
...
+// Node module: @loopback/core
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * Action to handle life cycle events
+ */
+export interface LifeCycleAction {
+  start(): Promise<void> | void;
+  stop(): Promise<void> | void;
```
bajtos: Let's discuss this API. Shouldn't we provide some kind of a context in which the event is happening? Without a context, how can a life-cycle-action handler access relevant data like the application and/or the server being started?

packages/core/src/application.ts
```diff
@@ -130,6 +131,9 @@ export class Application extends Context {
    * @memberof Application
    */
   public async start(): Promise<void> {
+    await this._forEachComponent(c => {
+      if (isLifeCycleAction(c)) return c.start();
```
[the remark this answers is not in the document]
raymondfeng: Ultimately we'll switch the `observer` pattern so that components can subscribe to the `start/stop` events. See #1972.

Do we want to wait unless we settle on #1972?

packages/core/src/lifecycle.ts
raymondfeng: If we switch to `observer` pattern, yes. For simple `start/stop`, I don't see the need to have an argument for context. Other information can be injected from the application context.

packages/core/src/application.ts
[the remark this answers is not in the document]
bajtos: I think that even with the observer pattern from #1972, we should still allow components to export lifecycle observers in an indirect way, i.e. letting `mountComponent` to handle the actual registration.

It is not clear to me how #1972 would affect the design of lifecycle events and whether the solution based purely on low-level observer APIs would be enough easy to use.

My preferred approach is to use composition instead of inheritance, allow components to export an array of life-cycle listener/observe classes and get this pull request landed.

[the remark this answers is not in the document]
bajtos: Later on, when #1972 is implemented, we can look into ways how to leverage the observer pattern as a low-level implementation detail to power "LifeCycleObserver".

packages/core/src/application.ts
```diff
@@ -130,6 +131,9 @@ export class Application extends Context {
    * @memberof Application
    */
   public async start(): Promise<void> {
+    await this._forEachComponent(c => {
```
bajtos: This opens an interesting question: should we invoke "start" observers before the servers are started or after?

 - When the observers are invoked before, they have a chance to modify server configuration.
 - When the observers are invoked after, they are able to access the information created during startup, e.g. the port number assigned by the operation system to build the URL of a server listening on `port: 0`.

I think the second use case is more important because I don't see how it could be implemented otherwise? The first use case (modifying config before start) seems redundant to me, because application setup can already be customized during `app.boot()`.

bajtos: Please make sure there is a test capturing this nuance!

raymondfeng: It's always tricky to make sure artifacts are started/stopped in the right order. For example, we usually want servers to be started last and stopped first so that requests can only be handled when the system is ready.

A way to describe the dependency is needed to allow life cycle actions to be triggered by order. For example, each action can declare its dependencies or desired phase.

raymondfeng:
@bajtos PTAL

packages/core/src/keys.ts
```diff
@@ -38,7 +38,29 @@ export namespace CoreBindings {
    */
   export const SERVERS = 'servers';
 
+  /**
+   * Binding tag for servers
+   */
+  export const SERVER_TAG = 'server';
```
bajtos: Instead of using `_TAG` suffix, can we create a new namespace instead?

```ts
CoreBindings.Tags.SERVER
```

packages/core/src/keys.ts
```diff
@@ -66,4 +88,9 @@ export namespace CoreBindings {
    * context
    */
   export const CONTROLLER_CURRENT = BindingKey.create('controller.current');
+
+  /**
+   * Binding tag for life cycle observers
+   */
+  export const LIFE_CYCLE_OBSERVER_TAG = 'lifeCycleObserver';
```
bajtos: Should we keep all tags grouped together?

packages/core/src/keys.ts
bajtos: Or perhaps better: `CoreTags.SERVER`.

So far, namespaces like `CoreBindings`, `RestBindings`, etc. were holding binding keys (addresses) only.

packages/core/src/application.ts
```diff
@@ -139,25 +166,50 @@ export class Application extends Context {
    * @memberof Application
    */
   public async stop(): Promise<void> {
-    await this._forEachServer(s => s.stop());
+    debug('Stopping the application...');
+    const bindings = this._findLifeCycleObserverBindings();
+    // Stop in the reverse order
+    for (const binding of bindings.reverse()) {
```
bajtos: Can we move the call of `.reverse()` out of the `for` loop? E.g.

```ts
const bindings = this._findLifeCycleObserverBindings().reverse();
```

packages/boot/src/booters/lifecyle-script.booter.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+}
+
+/**
+ * Default ArtifactOptions for DataSourceBooter.
+ */
+export const LifeCycleScriptDefaults: ArtifactOptions = {
+  dirs: ['scripts'],
+  extensions: ['.script.js'],
```
b-admike: `script.ts` as well?

packages/core/src/application.ts
```diff
@@ -130,7 +147,17 @@ export class Application extends Context {
    * @memberof Application
    */
   public async start(): Promise<void> {
-    await this._forEachServer(s => s.start());
+    debug('Starting the application...');
+    const bindings = this._findLifeCycleObserverBindings();
+    for (const binding of bindings) {
```
bajtos: Have you considered extracting this loop into a function accepting 1) a list of bindings 2) a string name of the method to invoke?

```ts
notifyLifeCycleObservers(bindings, 'start');
```

packages/core/src/application.ts
```diff
@@ -41,9 +55,11 @@ export class Application extends Context {
    */
   controller(controllerCtor: ControllerClass, name?: string): Binding {
     name = name || controllerCtor.name;
-    return this.bind(`controllers.${name}`)
+    const key = `controllers.${name}`;
+    debug('Adding controller %s', name);
+    return this.bind(key)
       .toClass(controllerCtor)
-      .tag('controller');
+      .tag(CoreBindings.CONTROLLER_TAG);
```
bajtos: It would be great if you could extract the part reworking existing binding tags from strings to constants and send a standalone pull request.

packages/core/src/application.ts
```diff
@@ -3,17 +3,31 @@
...
 import {Component, mountComponent} from './component';
 import {CoreBindings} from './keys';
+import {
+  asLifeCycleObserverBinding,
+  isLifeCycleObserver,
+  isLifeCycleObserverClass,
+  LifeCycleObserver,
+} from './lifecycle';
+import {Server} from './server';
+import debugModule = require('debug');
```
bajtos: In other places, I was already using `debugFactory` instead of `debugModule`. Let's pick one consistent way going forward please.

packages/context/src/binding.ts
```diff
@@ -423,6 +423,22 @@ export class Binding<T = BoundValue> {
     return this;
   }
 
+  /**
+   * Apply a template function to set up the binding with scope, tags.
+   *
+   * For example,
+   * ```ts
+   * const serverTemplate = (binding: Binding) =>
+   *   binding.inScope(BindingScope.SINGLETON).tag('server');
```
bajtos: Should we use the new `SERVER_TAG` constant instead of a hard-coded string?

packages/boot/src/booters/lifecyle-script.booter.ts
```diff
@@ -0,0 +1,70 @@
...
+        .apply(asLifeCycleObserverBinding);
+    }
+  }
+}
+
+/**
+ * Default ArtifactOptions for DataSourceBooter.
+ */
+export const LifeCycleScriptDefaults: ArtifactOptions = {
+  dirs: ['scripts'],
```
bajtos: I find the directory name `scripts` (and also the CLI command `lb4 script`) as not descriptive enough. Every JavaScript source code is essentially a script.

Let's find a name that better conveys the intention please. How about `lifecycle` directory and `lb4 lifecycle`? (`lifecycle-script` seems too long to me.)

packages/core/src/application.ts
```diff
@@ -14,13 +14,13 @@ import {Component, mountComponent} from './component';
...
 import {
   asLifeCycleObserverBinding,
-  isLifeCycleObserver,
   isLifeCycleObserverClass,
   LifeCycleObserver,
 } from './lifecycle';
 import {Server} from './server';
-import debugModule = require('debug');
-const debug = debugModule('loopback:core:application');
+const CoreTags = CoreBindings.Tags;
```
bajtos: There are quite few  places where we are aliasing `CoreBindings.Tags` to `CoreTags`. I see it as a clear sign that tags should be a new top-level export and not a nested namespace inside `CoreBindings`. I.e.

```ts
import {CoreBindings, CoreTags} from './keys';
```

Thoughts?

bajtos:
@raymondfeng I am afraid I have few more concerns about the high-level design.

(1)
Observers are sorted by the framework and servers get a special treatment:
- This is not extensible. When an extension comes with another artifact that needs special position in the list of lifecycle observers, then there is no mechanism allowing the extension to do that. We would have to add another exception to the sorting routine used by core.
- The order of observers is quite non-deterministic for app and especially for extension developers. It's easy to make incorrect assumptions about the order in which observers are executed at one point in application development lifecycle, and then be surprised that this order was changed when a new observer was added via a component or even explicitly by the app.

(2)
> It's always tricky to make sure artifacts are started/stopped in the right order. For example, we usually want servers to be started last and stopped first so that requests can only be handled when the system is ready.

Sure, this makes sense in a lot of cases. However, there are valid use cases where we want to observe the fact that the application has started (all servers are up & listening). Besides the example of printing application's URL to the console, there is also a more compelling scenario - registering the started application with a Service Registry and unregistering it *before* shutdown. 

For example, I can imagine a 3rd-party extension providing start/stop observers that will automatically register the application with a specific Service Registry, e.g. `loopback4-extension-consul` or `loopback4-extension-etcd`. This is not possible to implement with the current design.

I have few ideas for possible solutions, can you come up with few of your own too?
- I am proposing to run all observers in parallel. Conceptually, there is no need to start multiple servers one by one, they can be started in parallel.
- Introduce more lifecycle events. For example, `starting` can start the servers, `started` can register the server with Service Registry. Similarly for `stop`. Alternatively, we can have `beforeStart`, `start` and `afterStart`, but that feels a bit too much to me.
- Support server-specific observers. Instead of waiting for the entire app to start, we can have an observer registered for a RestServer only.
- Don't treat servers as lifecycle observers, but start them explicitly. For example, `app.start` can be implemented as three steps:
  1. Notify all observers - call `start` event (in parallel)
  2. Start all servers
  3. Notify all observers - call `started` event (in parallel)

Anyhow. For me, the important requirement is that we need to give observers the option to decide whether they should run before or after the servers are started. The rest are implementation details...

raymondfeng:
@bajtos I introduced a `group` concept to organize observer.  The order of groups can be configured so that we notify them by the order of its group. Observers within the same group are triggered in parallel. For a composite event such as `start`, we emit `preStart`, `start` and `postStart`.

Ultimately such infrastructure should be refactored into https://github.com/strongloop/loopback-next/pull/1972 and possibly align with https://github.com/strongloop/loopback-next/pull/1671.

From user experience perspective, we need to:

1. Allow observers/listeners/handlers to be registered/subscribed against their interests (such as topics/events/phases)
2. Allow observers/listeners/handlers to be discovered for a given event
3. Allow observers/listeners/handlers to be executed in order (for example, by group or phase)
4. Allow observers/listeners/handlers to be unregistered/unsubscribed

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (!group) {
+      // Fall back to a tag that matches one of the groups
+      group = this.groupsByOrder.find(g => binding.tagMap[g] === g);
+    }
+    return group || '';
+  }
+
+  /**
+   * Sort the life cycle observer bindings so that we can start/stop them
+   * in the right order. By default, we can start other observers before servers
```
bajtos: I don't understand why is this sorting still required when we already have `preStart`/`start`/`postStart` groups allowing observers to decide whether they want to run before or after the server is started. Could you please explain? What am I missing?

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,211 @@
...
+    protected ctx: Context,
+    protected options: LifeCycleObserverOptions = {parallel: true},
+  ) {}
+  /**
+   * Configures the ordering of life cycle observers by group names
+   */
+  protected groupsByOrder: string[] = ['server'];
+
+  setGroupsByOrder(groups: string[]) {
+    this.groupsByOrder = groups || ['server'];
```
bajtos: I have difficulties to comprehend this group ordering concept and how can application & extension users leverage this configuration point. Could you please explain, perhaps in the form of a documentation as part of this pull request?

packages/context/test/unit/binding.unit.ts
```diff
@@ -173,6 +173,16 @@ describe('Binding', () => {
     });
   });
 
+  describe('apply(templateFunction)', () => {
```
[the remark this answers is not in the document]
raymondfeng: See https://github.com/strongloop/loopback-next/pull/2119

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,211 @@
...
+   * binding tagged with `CoreBindings.Tags.LIFE_CYCLE_OBSERVER` or
+   * `CoreBindings.Tags.SERVER`.
+   */
+  findObserverBindings() {
+    return this.ctx.find<LifeCycleObserver>(
+      binding =>
+        (binding.type === BindingType.CONSTANT ||
+          binding.scope === BindingScope.SINGLETON) &&
+        (binding.tagMap[CoreTags.LIFE_CYCLE_OBSERVER] != null ||
+          binding.tagMap[CoreTags.SERVER]),
```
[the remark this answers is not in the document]
raymondfeng: We can get rid of `SERVER` tag as we tag servers during `app.server`.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,140 @@
...
+5. MongoDBDataSource.start
+6. RestServer.start
+7. MySQLDataSource.postStart
+8. MongoDBDataSource.postStart
+9. RestServer.postStart
+
+## Add custom life cycle observers by convention
+
+Each application can have custom life cycle observers to be dropped into
+`src/observers` folder as classes implementing `LifeCycleObserver`.
```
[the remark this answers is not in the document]
raymondfeng: I would like to use https://github.com/strongloop/loopback-next/pull/2129

raymondfeng:
> I am proposing to simplify the design as follows:

> Keep an ordered list of named groups (start datasource first, server second), this gives us a lot of flexibility & extensibility.

+1. 

> For each group, support only two events: start and stop. Users can implement post/after hooks by defining new named groups if they like to.

I'm fine to only allow `start` and `stop` for now. IMO, named groups solve the ordering problem, but it might force us to create multiple classes for `preStart`, `start`, and `postStart`. For example:

```
pre-server: [MyPreServer]
server: [MyServer]
post-server: [MyPostServer]
```

With the current design, we allow:
```
server: [MyServer] (allows preStart, start, and postStart in MyServer
```

> Eventually, we should refactor the implementation of an ordered list of named groups into @loopback/phase.

+1.

raymondfeng:
BTW, I would like to use https://github.com/strongloop/loopback-next/pull/2122 to simplify the discovery of life cycle observers.

raymondfeng:
#2122 is probably implementation detail for this PR. We can merge this one and refine it after #2122 is landed.

bajtos:
> I'm fine to only allow start and stop for now. IMO, named groups solve the ordering problem, but it might force us to create multiple classes for preStart, start, and postStart. For example:
>
> ```
> pre-server: [MyPreServer]
> server: [MyServer]
> post-server: [MyPostServer]
> ```
> With the current design, we allow:
> ```
> server: [MyServer] (allows preStart, start, and postStart in MyServer)
> ```

IMO, instead of generic pre-server and post-server groups, we should use (and advocate for) more descriptive names. E.g. `pre-server` can become `server-setup` and `post-server` can become `service-registration` (with Consule/Etcd) or `status-broadcast`.

The most important part is an easy and well understood solution allowing both extensions and app developers to introduce new groups and control their order relatively to other (well known?) groups.

raymondfeng:
PTAL

bajtos:
I did not manage to find time to review this pull request again 😞 I'll take a look when I get back from vacation in January.

bajtos:
I would like us to improve our development process so that more people review foundational features like the one proposed in this pull request. I have to goals in mind:
 - Get more opinions on the design & architecture decisions we are making - more eyes see more.
 - Spread the knowledge about how things are designed and implemented, so that people making subsequent changes can get meaningful feedback from more people than just me and @raymondfeng 

In that light, I'd kindly like to ask members of @strongloop/loopback-maintainers, especially (but not only) @hacksparrow and @jannyHou, to review the changes proposed in this pull request. Please don't focus on coding style and implementation details only, try to build understanding of the big picture and the high-level design. If there are anything that's not clear to you from the documentation, pull request discussion and/or the code, then please ask.

Few suggestions:
- Start at high level and ask yourself: do you understand the use cases/scenarios the PR is addressing?
- Put yourself in the shoes of LB4 users building applications. Would you find the proposed feature and APIs easy to discover and use? Do they make it easy to write application code that's clean, brief and easy to understand?
- Do a similar assessment from the point of view of extension developers.
- Think about our existing conventions and design decisions made in the past. Are the newly proposed APIs consistent with the rest of the framework?
- Implementation wise: is the code easy to read, understand and reason about?
- Testing: are all major scenarios and edge cases covered by tests? Are the individual tests easy to understand? Can you tell what is being tested by each individual test? Will the test print a helpful assertion error on failure?

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,212 @@
...
+two-dimension steps to control the order of life cycle actions.
+
+### Observer groups
+
+First of all, we allow each of the life cycle observers to be tagged with a
+group. For example:
+
+- datasource
+
+  - connect/disconnect mongodb
```
jannyHou: the tag here is a little bit weird, did you mean something like
```
- connect/disconnect
- mongodb
- mysql
```
?

packages/cli/generators/observer/templates/observer-template.ts.ejs
```diff
@@ -0,0 +1,33 @@
+import {bind} from '@loopback/context';
+import {
+  /* inject, Application, */
```
hacksparrow: Let's add `CoreBindings`, since the commented code below (`CoreBindings.APPLICATION_INSTANCE`) refers to it.

packages/core/src/application.ts
```diff
@@ -140,18 +148,37 @@ export class Application extends Context {
...
+    });
   }
 
   /**
-   * Helper function for iterating across all registered server components.
+   * Discover bindings matching the key namespace or tag
+   * @param keyNamespace Binding key prefix
+   * @param tag Tag name
+   */
+  private _findByNamespaceOrTag(keyNamespace: string, tag?: string) {
```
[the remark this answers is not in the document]
bajtos: Great idea, +1 from me

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,212 @@
...
+such artifacts are:
+
+- Servers
+
+  - start: Starts the HTTP server listening for connections.
+  - stop: Stops the server from accepting new connections.
+
+- Components
+
+  - A component can register life cycle observers
```
hacksparrow: "life cycle observers" or "life cycle events"? If "life cycle observers", can only components register them?

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,212 @@
...
+`LifeCycleObserver` interface.
+
+```ts
+import {ValueOrPromise} from '@loopback/context';
+
+/**
+ * Observers to handle life cycle start/stop events
+ */
+export interface LifeCycleObserver {
+  preStart?(): ValueOrPromise<void>;
```
hacksparrow: Wouldn't a dynamic interface (no hard-coded methods) be more flexible? Or do we plan support events only around start and stop?

Some other events that I can think of "ready", "connected", "online" (which was "offline" a while ago), "message"; especially if a component is dependent on other components asynchronously.

Something to think about, since we are considering LB4 as a messaging hub too (can't find your PR).

This makes we wonder if a "global event listener" is what we are looking for.

The interface of `LifeCycleObserver` is "set in stone", but with a "global event listener", we have the flexibility to accommodate an infinite type of events, not just those based around start and stop.

hacksparrow:
The concept seems to be based around the start and stop events of the main Application. How about we make it a general purpose implementation? That way we will be able to listen for lifecycle events on other components as well.

@raymondfeng

hacksparrow:
Making it a general purpose implementation will make "[Adding logic to models](https://loopback.io/doc/en/lb3/Adding-logic-to-models.html)" (maybe LB4 has a different approach, just an example), and other event/situation based handlers a lot easier.

bajtos:
> The concept seems to be based around the start and stop events of the main Application. How about we make it a general purpose implementation? That way we will be able to listen for lifecycle events on other components as well.

@hacksparrow see https://github.com/strongloop/loopback-next/pull/1928#discussion_r230709564

This pull request is implementing start/stop events only to avoid over-engineering and have a specific use case to focus on. I believe the pull request https://github.com/strongloop/loopback-next/pull/1972 is implementing a more generic infrastructure that can be used for other purposes, the plan is to rework start/stop implementation on top of that PR once it lands.

hacksparrow:
Thanks for the details, @bajtos. As long as we eventually switch this implementation to be based on #1972, I am fine with it. Will be looking deeper into the implementation details today.

hacksparrow:
@bajtos @raymondfeng we should have a unified means of event notification. The implementation in this PR and #2122 are different, although the basic requirements are the same; this is a hint at fragmentation and impending technical debt.

I am inclined towards making https://github.com/strongloop/loopback-next/pull/1972 a priority and landing it first.

This PR and #2122 should be using a common low-level notification infrastructure.

hacksparrow:
@bajtos is there a reason why we might want to land this implementation as-is, without waiting for #1972?

hacksparrow:
> The worst outcome of our discussion would be to combine this pull request and #1972 into one huge change-set.

Nah, #1972 should land on its own. This PR should then be changed to use the infrastructure created by #1972, so should #2122 be. The size of the new PRs would then be smaller than the current ones, since the common implementation details would reside somewhere else.

@raymondfeng thoughts? I can actively help to land #1972.

hacksparrow:
Another place where we would be using #1972 - https://github.com/strongloop/loopback-next/issues/1919.

raymondfeng:
@hacksparrow I wanted to land in the following order:
- #2122: make it simpler to implement extension point/extension pattern.
- #2259: make it simpler to configure bound services
- #2249: illustrate the usage of #2122 and #2259 
- #1928: organize life cycle observers as extensions
- #1972: extract the common code from #2122, #1928 as refinement

hacksparrow:
Thanks @raymondfeng for the update.

bajtos:
@raymondfeng what's the status of this pull request? Is it ready for another round of reviews?

Going forward, could you please use GitHub's recently added feature [Re-request review on a pull request](https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/) to let reviewers know that it's time for another review?

bajtos:
@raymondfeng AFAICT from what GitHub is showing to me, the following comments were not addressed yet: 

https://github.com/strongloop/loopback-next/pull/1928#issuecomment-451933919
(1)
@raymondfeng in https://github.com/strongloop/loopback-next/pull/1928#issuecomment-445047426, you wrote:

> I'm fine to only allow start and stop for now.

However, the pull request still uses preStart/start/postStart hooks. Is it intentional? What's the reasoning behind the decision?

(2)
Also AFAICT, the discussion in https://github.com/strongloop/loopback-next/pull/1928#discussion_r230709564 has not been resolved yet. Besides the important design decision we need to make, it reminds me of a missing documentation for extension authors.

(3)
Please add a new page nested under https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html and show to extension developers what is the idiomatic way for an extension to be notified about lifecycle events. As I see it, the page `docs/site/Life-cycle.md` is focused on application developers only, which is good and needed, but we also need docs focused on extension developers.

https://github.com/strongloop/loopback-next/pull/1928#pullrequestreview-189799320
(4)
And few more minor comments related to implementation details:

https://github.com/strongloop/loopback-next/pull/1928#discussion_r245646461
Now that #2129 has landed, could you expand this section of the documentation and show how lifecycle observes should specify the group they want to belong to? AFAICT, the template used by the CLI tool has been already updated to leverage @bind() decorator.

https://github.com/strongloop/loopback-next/pull/1928#discussion_r245646804
Should the CLI ask for the observer group via an interactive prompt and then set the group here?

https://github.com/strongloop/loopback-next/pull/1928#discussion_r245647310
I believe we are using this form in other places:

```ts
import * as debugFactory from 'debug';
```

Let's stay consistent please.

https://github.com/strongloop/loopback-next/pull/1928#discussion_r245647743
Please make the patch smaller by preserving the place where start/stop methods are defined.

You pull request is moving them from lines 136-172 to lines 200-218.

https://github.com/strongloop/loopback-next/pull/1928#discussion_r245647995
Consider extracting this line into a shared helper function, e.g.

```ts
    const lifecycle = await this.getLifeCycleObserverRegistry();
```

---

I see also more comments from other team members that does not seem to be addressed or responded to. Please review them all.

bajtos:
It's annoying that GitHub is hiding away comments that were not resolved, please keep in mind that you need to expand the hidden comments:

Screen Shot 2019-03-21 at 13 19 20

packages/boot/src/booters/lifecyle-script.booter.ts
raymondfeng: No, we only scan `*.js` files from `dist`.

raymondfeng:
> #1928 (comment)
(1)
@raymondfeng in #1928 (comment), you wrote:
>> I'm fine to only allow start and stop for now.

> However, the pull request still uses preStart/start/postStart hooks. Is it intentional? What's the reasoning behind the decision?

Fixed.

> (2)
Also AFAICT, the discussion in #1928 (comment) has not been resolved yet. Besides the important design decision we need to make, it reminds me of a missing documentation for extension authors.

Added Component.lifeCycleObservers as an array of observer classes.

> (3)
Please add a new page nested under https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html and show to extension developers what is the idiomatic way for an extension to be notified about lifecycle events. As I see it, the page docs/site/Life-cycle.md is focused on application developers only, which is good and needed, but we also need docs focused on extension developers.

TBD.

> #1928 (review)
(4)
And few more minor comments related to implementation details:

> #1928 (comment)
Now that #2129 has landed, could you expand this section of the documentation and show how lifecycle observes should specify the group they want to belong to? AFAICT, the template used by the CLI tool has been already updated to leverage @Bind() decorator.

Fixed.

> #1928 (comment)
Should the CLI ask for the observer group via an interactive prompt and then set the group here?

Fixed.

> #1928 (comment)
I believe we are using this form in other places:
import * as debugFactory from 'debug';
Let's stay consistent please.

Fixed.

> #1928 (comment)
Please make the patch smaller by preserving the place where start/stop methods are defined.

See #2625

> You pull request is moving them from lines 136-172 to lines 200-218.

Fixed.

> #1928 (comment)
Consider extracting this line into a shared helper function, e.g.
    const lifecycle = await this.getLifeCycleObserverRegistry();

Fixed.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -0,0 +1,149 @@
...
+      expect(observer.status).to.equal('stopped');
+    });
+
+    it('does not attempt to start poorly named bindings', async () => {
+      const app = new Application();
+      app.component(FakeComponent);
+
+      // The app.start should not attempt to start this binding.
+      app.bind('controllers.servers').to({});
+      await app.start();
```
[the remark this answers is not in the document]
emonddr: I had the same request. The test is passing with app.start() and then app.stop() and no assertions.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,282 @@
...
+
+A life cycle observer can be registered by calling `lifeCycleObserver()` of the
+application. It binds the observer to the application context with a special
+tag - `CoreTags.LIFE_CYCLE_OBSERVER`.
+
+```ts
+app.lifeCycleObserver(MyObserver);
+```
+
+Please note that `app.server()` automatically registers servers as life cycle
```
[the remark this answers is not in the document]
raymondfeng: Yes.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,282 @@
...
+  - gRPC
+
+We can then configure the application to trigger observers group by group as
+configured by an array of groups in order such as `['datasource', 'server']`.
+Observers within the same group can be notified in parallel.
+
+For example,
+
+```ts
+app
```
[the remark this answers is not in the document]
raymondfeng: Use `@bind` decoration.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,282 @@
...
+- start: preStart, start, and postStart
+- stop: preStop, stop, and postStop
+
+Combining groups and event phases, it's flexible to manage multiple observers so
+that they can be started/stopped gracefully in order.
+
+For example, with a group order as `['datasource', 'server']` and three
+observers registered as follows:
+
+- datasource group: MySQLDataSource, MongoDBDataSource
```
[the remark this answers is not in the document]
raymondfeng: There is no ordering within a group. It's probably by the order of registration.

packages/core/src/__tests__/unit/lifecycle-registry.unit.ts
```diff
@@ -0,0 +1,99 @@
...
+  function givenContext() {
+    context = new Context('app');
+  }
+
+  async function givenLifeCycleRegistry() {
+    context
+      .bind(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY)
+      .toClass(LifeCycleObserverRegistry)
+      .inScope(BindingScope.SINGLETON);
+    registry = await context.get(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY);
```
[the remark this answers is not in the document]
raymondfeng: No. Please note `CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY` is already strongly-typed.

packages/core/src/__tests__/unit/lifecycle-registry.unit.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function givenObserver(name: string, group = '') {
```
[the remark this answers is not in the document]
raymondfeng: Yes.

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,241 @@
...
+   * Control the order of observer groups for notifications. For example,
+   * with `['datasource', 'server']`, the observers in `datasource` group are
+   * notified before those in `server` group during `start`. Please note that
+   * observers are notified in the reverse order during `stop`.
+   */
+  groupsByOrder: string[];
+  /**
+   * Notify observers of the same group in parallel, default to `true`
+   */
+  parallel?: boolean;
```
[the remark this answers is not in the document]
raymondfeng: I don't think `type` allows initializer.

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,241 @@
...
+    this.options.groupsByOrder = groups || ['server'];
+  }
+
+  /**
+   * Get observer groups ordered by the group
+   */
+  protected getObserverGroupsByOrder(): LifeCycleObserverGroup[] {
+    const bindings = this.observersView.bindings;
+    const groups = this.sortObserverBindingsByGroup(bindings);
+    if (debug.enabled) {
```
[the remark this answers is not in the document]
raymondfeng: It's controlled by `DEBUG` env var. Checking `debug.enabled` to avoid unnecessary calculation if the debug statement uses parameters with expressions.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,282 @@
...
+To make it easy for application developers to add custom life cycle observers,
+we introduce `lb4 observer` command as part the CLI.
+
+To add a life cycle observer:
+
+1. cd <my-loopback4-project>
+2. lb4 observer
+
+```
+? Observer name: Hello
```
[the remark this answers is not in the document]
raymondfeng: We added.

raymondfeng:
> Please add a new page nested under https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html and show to extension developers what is the idiomatic way for an extension to be notified about lifecycle events. As I see it, the page docs/site/Life-cycle.md is focused on application developers only, which is good and needed, but we also need docs focused on extension developers.

Added.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,217 @@
...
+  - mongodb
+  - mysql
+
+- server
+  - rest
+  - gRPC
+
+We can then configure the application to trigger observers group by group as
+configured by an array of groups in order such as `['datasource', 'server']`.
+Observers within the same group can be notified in parallel.
```
bajtos: > Can be notified in parallel

I find this part ambiguous. Are you trying to say that observers are notified in series by default, but there is an option to enable parallel processing instead? How can readers of this paragraph find how to set that option?

Or did you perhaps want to say that _observers within the same group **are** notified in parallel_?

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,217 @@
...
+      [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
+    },
+  },
+  asLifeCycleObserver,
+)
+export class MyObserver {
+  // ...
+}
+
+app.add(createBindingFromClass(MyObserver));
```
bajtos: Probably out of scope of this pull request: would it make sense to introduce Context-level sugar API for adding a binding created from a class?

```diff
- app.add(createBindingFromClass(MyObserver));
+ app.bindFromClass(MyObserver);
```

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,217 @@
...
+
+Or:
+
+```ts
+const registry = await app.get(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY);
+registry.setGroupsByOrder(['g1', 'g2', 'server']);
+```
+
+Observers are sorted using `groupsByOrder` as the relative order. If an observer
+is tagged with a group that are not in `groupsByOrder`, it will come before any
```
bajtos:
```suggestion
is tagged with a group that is not in `groupsByOrder`, it will come before any
```

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,217 @@
...
+- 'my-observer-3' ('1-custom-group')
+
+The sorted observer groups will be:
+
+```ts
+{
+  '1-custom-group': ['my-observer-3'],
+  '2-custom-group': ['my-observer-4'],
+  'g1': ['my-observer-1'],
+  'g2': ['my-observer-2'],
```
bajtos: This example would work for multiple different sorting rules. For example, `1-custom-group` is alphabetically preceding `g1`, also `g1` is before `g2`. Your entire list is sorted in the alphabetical order, it's not clear which part is controller by specific rules for unknown groups, which part is controlled by `groupsByOrder` and which is controlled by alphabetical sorting.


I am proposing to use the following example instead:

- `groupsByOrder `is set to `['setup-servers', 'announce-services']`
- 'my-observer-1' in group 'setup-servers'
- 'my-observer-2' in group 'announce-services'
- 'my-observer-4' in 'my-custom-group'
- 'my-observer-3' in 'another-custom-group'

Also please check how is this section rendered from Markdown to HTML, I think it may need a bit of polishment.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+The execution order of observers within the same group is not defined. If you
+want to have one to be invoked before the other, mark them with two groups.
```
bajtos: 👍

docs/site/Life-cycle.md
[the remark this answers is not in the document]
bajtos: > could you expand this section of the documentation and show how lifecycle observes should specify the group they want to belong to? AFAICT, the template used by the CLI tool has been already updated to leverage `@bind()` decorator.

☝️

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -0,0 +1,148 @@
...
+      await app.stop();
+    });
+  });
+});
+
+class FakeComponent implements Component {
```
bajtos: What aspects of this component make it a fake component? To me, this looks like a legitimate component implementing LifeCycleObserver interface and providing two "dummy" servers.

```suggestion
class ObservingComponentWithServers implements Component, LifeCycleObservers {
```

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+  start() {
+    this.status = 'started';
+  }
+  stop() {
+    this.status = 'stopped';
+  }
+}
+
+class FakeServer extends Context implements Server {
```
bajtos: Similarly here.

We are using Sinon to build test doubles in other places. If you are using "fake" as a synonym to a test double, then please use the name `ServerStub` instead, it's more consistent with Sinon terminology.

Personally, I'd use a name that conveys the intention to observe lifecycle events, e.g. `ObservingServer`.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -0,0 +1,148 @@
...
+      expect(observer.status).to.equal('not-initialized');
+      await app.start();
+      expect(observer.status).to.equal('started');
+      await app.stop();
+      expect(observer.status).to.equal('stopped');
+    });
+
+    it('does not attempt to start poorly named bindings', async () => {
+      const app = new Application();
+      app.component(FakeComponent);
```
bajtos: I don't see how is FakeComponent relevant to this particular test case. Can you remove this line please?

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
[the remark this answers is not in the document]
bajtos: +1

packages/core/src/application.ts
```diff
@@ -216,6 +222,26 @@ export class Application extends Context {
...
+    ctor: Constructor<T>,
+    name?: string,
+  ): Binding<T> {
+    debug('Adding life cycle observer %s', name || ctor.name);
+    const binding = createBindingFromClass(ctor, {
+      name,
+      namespace: CoreBindings.LIFE_CYCLE_OBSERVERS,
+      type: CoreTags.LIFE_CYCLE_OBSERVER,
+      defaultScope: BindingScope.SINGLETON,
+    }).apply(asLifeCycleObserver);
```
bajtos: Can we move the default namespace, type and defaultScope to `asLifeCycleObserver` template?

```ts
const binding = createBindingFromClass(ctor, {name})
  .apply(asLifeCycleObserver);
```

packages/core/src/lifecycle-registry.ts
```diff
@@ -0,0 +1,243 @@
...
+    groups: LifeCycleObserverGroup[],
+    reverse = false,
+  ) {
+    const observers = await this.observersView.values();
+    const bindings = this.observersView.bindings;
+    if (reverse) groups = groups.reverse();
+    for (const group of groups) {
+      const observersForGroup: LifeCycleObserver[] = [];
+      const bindingsInGroup = reverse
+        ? group.bindings.reverse()
```
bajtos: > https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse
> The `reverse()` method reverses an array in place. The first array element becomes the last, and the last array element becomes the first.

I don't think it's a good idea to modify `group.bindings` array.

```suggestion
        ? [...group.bindings].reverse()
```

Please add a test that fails in the current implementation and passes with the fix in place.

packages/core/src/lifecycle.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export function asLifeCycleObserver<T = unknown>(binding: Binding<T>) {
+  return binding.tag(CoreTags.LIFE_CYCLE_OBSERVER);
+}
+
+/**
+ * Find all life cycle observer bindings. By default, a binding tagged with
+ * `CoreTags.LIFE_CYCLE_OBSERVER`
+ */
+export const lifeCycleObserverFilter: BindingFilter = binding =>
```
bajtos: Let's use a regular named function please.

```ts
export function lifeCycleObserverFilter(binding: Binding): boolean {
  // ...
}
```

packages/core/src/application.ts
bajtos: Or even better:

```ts
app.bindFromClass(ctor, {name}, asLifeCycleObserver);
```

docs/site/Life-cycle.md
raymondfeng: +1.

packages/core/src/application.ts
raymondfeng: > Can we move the default namespace, type and defaultScope to asLifeCycleObserver template?

In fact, no. Such information is used to infer binding attributes but the template function is applied after the binding is created.

packages/core/src/lifecycle.ts
raymondfeng: I was trying to make it clear that `lifeCycleObserverFilter` is a `BindingFilter` function.

hacksparrow:
@bajtos @raymondfeng  Is there a reason why `packages/boot/test` hasn't been moved to `packages/boot/src/__test__`?

docs/site/Extension-life-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+```ts
+app.lifeCycleObserver(MyObserver);
+```
+
+Life cycle observers can be registered via a component too:
+
+```ts
+export class MyComponentWithObservers {
+  lifeCycleObservers = [XObserver, YObserver];
```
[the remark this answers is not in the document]
hacksparrow: @bajtos your thoughts?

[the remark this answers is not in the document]
raymondfeng: I'm adding `implements Component` and some comments to clarify.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,237 @@
...
+app
+  .bind(CoreBindings.LIFE_CYCLE_OBSERVER_OPTIONS)
+  .to({groupsByOrder: ['g1', 'g2', 'server']});
+```
+
+Or:
+
+```ts
+const registry = await app.get(CoreBindings.LIFE_CYCLE_OBSERVER_REGISTRY);
+registry.setGroupsByOrder(['g1', 'g2', 'server']);
```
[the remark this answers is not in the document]
raymondfeng: Good idea. Changed.

raymondfeng:
> @bajtos @raymondfeng Is there a reason why packages/boot/test hasn't been moved to packages/boot/src/__test__?

Good catch. The PR was created before we moved test to `src/__tests__` and these are left-over files. Fixed now.

raymondfeng:
@bajtos PTAL.

bajtos:

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -13,6 +19,7 @@ import {
   LifeCycleObserver,
   Server,
 } from '../..';
+import {lifeCycleObserver} from '../../lifecycle';
```
bajtos: Please import from `../..`, just add the new imported symbol to the existing import above.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -153,6 +198,37 @@ class MyObserver implements LifeCycleObserver {
...
-  lifeCycleObservers = [MyObserver];
+@bind({
+  tags: {
+    [CoreTags.LIFE_CYCLE_OBSERVER]: CoreTags.LIFE_CYCLE_OBSERVER,
+    [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'my-group',
+    namespace: CoreBindings.LIFE_CYCLE_OBSERVERS,
+  },
+  scope: BindingScope.SINGLETON,
+})
+class MyObserverWithBind implements LifeCycleObserver {
```
bajtos: This class is used from a single test only, please move its definition to that test case. I think it will make the test easier to read.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -153,6 +198,37 @@ class MyObserver implements LifeCycleObserver {
...
+  start() {
+    this.status = 'started';
+  }
+  stop() {
+    this.status = 'stopped';
+  }
+}
+
+@lifeCycleObserver('my-group', BindingScope.SINGLETON)
+class MyObserverWithDecorator implements LifeCycleObserver {
```
bajtos: This class is used from a single test only, please move its definition to that test case. I think it will make the test easier to read.

packages/core/src/lifecycle.ts
```diff
@@ -57,3 +65,17 @@ export function asLifeCycleObserver<T = unknown>(binding: Binding<T>) {
 export function lifeCycleObserverFilter(binding: Binding<unknown>): boolean {
   return binding.tagMap[CoreTags.LIFE_CYCLE_OBSERVER] != null;
 }
+
+/**
+ * Sugar decorator to mark a class as life cycle observer
+ * @param group Optional observer group name
+ * @param scope Optional binding scope
+ */
+export function lifeCycleObserver(group = '', scope?: BindingScope) {
```
bajtos: I am concerned about flexibility of this API. Have you considered to accept the same arguments as `@bind` does?

```ts
export function lifeCycleObserver(group = '', ...specs: BindingSpec[]) {
  return bind(
    asLifeCycleObserver,
    {tags: [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: group},
    ...specs
  );
}
```

docs/site/Extension-life-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+import {bind, createBindingFromClass} from '@loopback/context';
+import {CoreTags, asLifeCycleObserver} from '@loopback/core';
+
+@bind(
+  {
+    tags: {
+      [CoreTags.LIFE_CYCLE_OBSERVER_GROUP]: 'g1',
+    },
+  },
+  asLifeCycleObserver,
```
bajtos: I find this example pretty complex. Can we show the new `lifeCycleObserver` decorator instead?

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
bajtos: > What aspects of this component make it a fake component? To me, this looks like a legitimate component implementing LifeCycleObserver interface and providing two "dummy" servers.

@raymondfeng PTAL ☝️

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
bajtos: ☝️ PTAL

packages/core/src/application.ts
bajtos: Does it mean that if the lifecycle class is decorated with `@bind` and sets the type to something else than `CoreTags.LIFE_CYCLE_OBSERVER`, then the result of calling `app.lifeCycleObserver` is a binding with a type that's not LIFE_CYCLE_OBSERVER? Similarly for `namespace`. Are we okay with that? I find it surprising to be honest.

docs/site/Life-cycle-observer-generator.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+{% include content/generator-create-app.html lang=page.lang %}
+
+### Synopsis
+
+Adds a new [LifeCycleObserver](Life-cycle.md) class to a LoopBack application.
+
+```sh
+lb4 observer [--group <group>] [<name>]
```
bajtos: Please describe the new generator in CLI README too, see https://github.com/strongloop/loopback-next/blob/master/packages/cli/README.md

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
raymondfeng: No, it's used in more than one tests.

packages/core/src/lifecycle.ts
raymondfeng: 👍

docs/site/Life-cycle-observer-generator.md
raymondfeng: Good catch.

raymondfeng:
> Does it mean that if the lifecycle class is decorated with @bind and sets the type to something else than CoreTags.LIFE_CYCLE_OBSERVER, then the result of calling app.lifeCycleObserver is a binding with a type that's not LIFE_CYCLE_OBSERVER? Similarly for namespace. Are we okay with that? I find it surprising to be honest.

Please note that `type` and `namespace` options in `createBindingFromClass()` overrides `@bind`.

docs/site/Extension-life-cycle.md
```diff
@@ -0,0 +1,71 @@
...
+tag - `CoreTags.LIFE_CYCLE_OBSERVER`.
+
+```ts
+app.lifeCycleObserver(MyObserver);
+```
+
+Life cycle observers can be registered via a component too:
+
+```ts
+export class MyComponentWithObservers {
```
hacksparrow: @raymondfeng ` implements Component` not added.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,256 @@
...
+
+Life cycle observers can be registered via a component too:
+
+```ts
+export class MyComponentWithObservers implements Component {
+  /**
+   * Populate `lifeCycleObservers` per `Component` interface to register life
+   * cycle observers
+   */
+  lifeCycleObservers = [XObserver, YObserver];
```
hacksparrow: The current documentation does not contain any information about `lifeCycleObservers`. Opened an issue to document Components - https://github.com/strongloop/loopback-next/issues/2701

packages/boot/src/__tests__/fixtures/lifecycle-observer.artifact.ts
```diff
@@ -0,0 +1,18 @@
...
+import {LifeCycleObserver} from '@loopback/core';
+
+export class MyLifeCycleObserver implements LifeCycleObserver {
+  status = '';
+
+  async start() {
+    this.status = 'started';
+  }
+
+  stop() {
```
hacksparrow: Any reason for not making this async as well?

bajtos:
> > Does it mean that if the lifecycle class is decorated with @bind and sets the type to something else than CoreTags.LIFE_CYCLE_OBSERVER, then the result of calling app.lifeCycleObserver is a binding with a type that's not LIFE_CYCLE_OBSERVER? Similarly for namespace. Are we okay with that? I find it surprising to be honest.

> Please note that `type` and `namespace` options in `createBindingFromClass()` overrides `@bind`.

Well in that case, what are the arguments against setting the tag & namespace via the `asLifeCycleObserver` template?

What I am looking for: an easy way how to bind a life-cycle observer without using `createBindingFromClass` or `app.lifeCycleObserver`.

packages/core/src/application.ts
```diff
@@ -77,12 +90,13 @@ export class Application extends Context {
...
     name?: string,
   ): Binding<T> {
+    debug('Adding server %s', name || ctor.name);
     const binding = createBindingFromClass(ctor, {
       name,
       namespace: CoreBindings.SERVERS,
       type: CoreTags.SERVER,
       defaultScope: BindingScope.SINGLETON,
-    });
+    }).apply(asLifeCycleObserver);
```
bajtos: A typical LifeCycleObserver is bound with the following metadata

```ts
    const binding = createBindingFromClass(ctor, {
	      name,
	      namespace: CoreBindings.LIFE_CYCLE_OBSERVERS,
	      type: CoreTags.LIFE_CYCLE_OBSERVER,
	      defaultScope: BindingScope.SINGLETON,
	    }).tag(CoreTags.LIFE_CYCLE_OBSERVER);
```

When a server is registered as an observer, some of the metadata will be different:

- namespace: `SERVERS` vs. `LIFE_CYCLE_OBSERVERS`
- type: `SERVER` vs. `LIFE_CYCLE_OBSERVER`

Is it intended and acceptable? I am concerned about possible confusion it may create.

packages/core/src/application.ts
```diff
@@ -216,6 +222,26 @@ export class Application extends Context {
...
+    ctor: Constructor<T>,
+    name?: string,
+  ): Binding<T> {
+    debug('Adding life cycle observer %s', name || ctor.name);
+    const binding = createBindingFromClass(ctor, {
+      name,
+      namespace: CoreBindings.LIFE_CYCLE_OBSERVERS,
+      type: CoreTags.LIFE_CYCLE_OBSERVER,
+      defaultScope: BindingScope.SINGLETON,
+    }).tag(CoreTags.LIFE_CYCLE_OBSERVER);
```
bajtos: Can we use `.apply(asLifeCycleObserver)` here? Would the following work?

```suggestion
    }).apply(asLifeCycleObserver)
```

docs/site/Extension-life-cycle.md
```diff
@@ -0,0 +1,71 @@
...
+on life cycle events is important for extension modules to collaborate with the
+application.
+
+An extension module follows the same way as applications to implement and
+register life cycle observers.
+
+### Implement a life cycle observer
+
+A life cycle observer class optionally implements `start` and `stop` methods to
+be invoked upon `start` and `stop` events.
```
hacksparrow: It'll add to the clarity if it is mentioned `start` and `stop` events of what.

docs/site/Extension-life-cycle.md
hacksparrow: The following code doesn't make sense in the context of the statement "_Life cycle observers can be registered via a component too:_" There's no registration going on.

docs/site/Life-cycle.md
```diff
@@ -0,0 +1,256 @@
...
+  - stop: Stops the server from accepting new connections.
+
+- Components
+
+  - A component itself can be a life cycle observer and it can also contribute
+    life cycle observers
+
+- DataSources
+
+  - connect: Connect to the underlying database or service
```
hacksparrow: Are these events supported? If yes, we need to show an example.

docs/site/Extension-life-cycle.md
```diff
@@ -0,0 +1,71 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Extension-life-cycle.html
+---
+
+## Extension life cycle
+
+As described in [Life cycle](Life-cycle.md), a LoopBack
+[Application](Application.md) has its own life cycles at runtime. Corresponding
+events such as `start` and `stop` are emitted upon the state change.
```
hacksparrow: Are there any events apart from `start` and `stop` that can be observed? If no, should make it very clear about the fact.

packages/boot/src/__tests__/fixtures/lifecycle-observer.artifact.ts
raymondfeng: Added some comments in the code.

packages/core/src/application.ts
raymondfeng: The only requirement is to be tagged as `CoreTags.LIFE_CYCLE_OBSERVER`. I have updated code accordingly.

raymondfeng:
The observer pattern can be used to implement life cycle observers but the proposed apis can stay. If we have to change, we can always release a new major version. Release often, release fast :-)

PR #1931, opened 2018-10-27
docs: add webstorm

Yaty:
This PR adds documentation for Webstorm IDE.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -33,7 +33,8 @@ and tools contained in LoopBack repositories.
 
 - [Monorepo overview](./site/MONOREPO.md)
 - [Setting up development environment](./site/DEVELOPING.md#setting-up-development-environment)
-  - [Developing with VisualStudio Code](./site/VSCODE.md)
+  - [Developing with VisualStudio Code (recommended)](./site/VSCODE.md)
+  - [Developing with WebStorm](./site/WEBSTORM.md)
 - [How to contribute the code](http://loopback.io/doc/en/contrib/code-contrib.html#how-to-contribute-to-the-code)
 - [Building the project](./site/DEVELOPING.md#building-the-project)
 - [Running tests](./site/DEVELOPING.md#running-tests)
--- a/docs/site/WEBSTORM.md
+++ b/docs/site/WEBSTORM.md
@@ -1,0 +1,85 @@
+# Developing with Webstorm
+
+There is currently an issue opened with Webstorm :
+[WEB-34416](https://youtrack.jetbrains.com/issue/WEB-34416)
+
+It should be fixed in `2018.3`, WebStorm is indexing symlinks create by Lerna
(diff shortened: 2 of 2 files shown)


raymondfeng:
@bajtos Please review it too.

docs/site/WEBSTORM.md
```diff
@@ -0,0 +1,85 @@
...
+There is currently an issue opened with Webstorm :
+[WEB-34416](https://youtrack.jetbrains.com/issue/WEB-34416)
+
+It should be fixed in `2018.3`, WebStorm is indexing symlinks create by Lerna
+inside `node_modules`, so it takes ages. As a workaround you need to exclude
+every `node_modules` directory or you can use [VisualStudio Code](./VSCODE.md).
+
+We recommend our contributors to use
+[WebStorm](https://www.jetbrains.com/webstorm/) with the following extensions
+installed:
```
bajtos: As I am reading this sentence, it's saying that the LoopBack project recommends people contributing code to use WebStorm. I am rather reluctant to recommend WebStorm because most of the team is using VS Code now.  I expect this will change in the future as we recruit more maintainers from our community, but we are not there yet.

So, I am proposing to use the following phrasing:

```markdown
For contributors using [WebStorm](https://www.jetbrains.com/webstorm/) as their preferred 
development environment, we recommend the following extensions to be installed:
```

docs/CONTRIBUTING.md
```diff
@@ -33,7 +33,8 @@ and tools contained in LoopBack repositories.
 
 - [Monorepo overview](./site/MONOREPO.md)
 - [Setting up development environment](./site/DEVELOPING.md#setting-up-development-environment)
-  - [Developing with VisualStudio Code](./site/VSCODE.md)
+  - [Developing with VisualStudio Code (recommended)](./site/VSCODE.md)
```
bajtos: +1 for adding `(recommended)` here 👍

docs/site/WEBSTORM.md
```diff
@@ -0,0 +1,85 @@
...
+1.  Open any existing TypeScript file, e.g. `packages/src/index.ts`
+
+2.  Add a small bit of code to break TypeScript's type checks, for example:
+
+    ```ts
+    const foo: number = 'bar';
+    ```
+
+3.  Verify that WebStorm editor has marked `foo` with a red underscore. Hover
+    over `foo` and check the problem message. It should mention `[ts]` source,
```
bajtos: The text says "it should mention `[ts]` source", but the text message shown as an example does not contain `[ts]` string. I am not familiar with WebStorm anymore, is the `[ts]` source shown somewhere else in the UI? Should we change the proposed docs to say something else, e.g. "It should start with `TS` prefix"?

Yaty: Nice catch ! I missed that :)

bajtos:
Landed, thank you @Yaty for the contribution 🎉

PR #1936, opened 2018-10-29
feat(rest): switch to body parser and add extensibility for parsers

raymondfeng:
Recreation of https://github.com/strongloop/loopback-next/pull/1887

This PR adds the following features:

switches to express body-parser:
- body-parser is more actively maintained that body
- body-parser supports compression options

Allow body parsers to be plugged in
- define an interface for BodyParser
- allow body parser extensions to be bound to the context with REQUEST_BODY_PARSER_TAG.

## Checklist

Documentation TODO as seen by @bajtos:
- [ ] How to build extensions providing custom body parsers. 
  - Show them how to contribute custom body parsers, the guide should mention the proposed helper function for creating bindings (`createBodyParserBinding`).
  - Describe `executeBodyParserMiddleware`, this function can be very useful to extension authors building custom body parsers, because it effectively allows them to take any Express middleware performing body parsing and wrap it into LB4 RequestParser. It would be great to change it into a well-documented public API.
- [ ] How to consume request body as a stream
  - In the documentation for this new feature, use the scenario where the app/extension developer wants to consume the request body as a stream to drive the narrative.
  - I can imagine this new documentation can take a form of a new page in the section "How tos", plus a short mention of the new extension flag in apidocs and/or `@requestBody` overview.


bajtos:
@raymondfeng  what's the status of this pull request, is it ready for review? I see a conflict in `packages/rest/package.json` reported by GitHub.

raymondfeng:
@bajtos I resolved the conflict and it's ready for review.

raymondfeng:
@bajtos PTAL

bajtos:
Sorry, the GitHub notification got buried in my inbox, I'll review early next week.

packages/rest/src/body-parser.ts
```diff
@@ -0,0 +1,287 @@
...
+
+/**
+ * Parse the body asynchronously
+ * @param handle The express middleware handler
+ * @param request Http request
+ */
+function parse(
+  handle: BodyParserWithCallback,
+  request: Request,
+): Promise<void> {
```
bajtos: Can we change `parse` to return the parsed body please?

I find such design more cohesive: the details about request parsing and storing the parsed body on `request.body` property stay inside `parse`. The rest of this file can stay free from this (not particularly nice) detail.

packages/rest/src/body-parser.ts
```diff
@@ -0,0 +1,287 @@
...
+// Default limit of the body length
+const DEFAULT_LIMIT = '1mb';
+
+type ParserOption<T extends 'json' | 'urlencoded' | 'text'> = T extends 'json'
+  ? OptionsJson
+  : T extends 'urlencoded'
+    ? OptionsUrlencoded
+    : T extends 'text' ? OptionsText : Options;
+
+function getParserOptions<T extends 'json' | 'urlencoded' | 'text'>(
```
bajtos: (1) I find this notation as rather complex/convoluted Have you considered using function overloads instead?

```ts
function getParserOptions(type: 'json', options: RequestBodyParserOptions): OptionsJson;
function getParserOptions(type: 'urlencoded', options: RequestBodyParserOptions): OptionsUrlencoded;
function getParserOptions(type: 'text', options: RequestBodyParserOptions): OptionsText;

function getParserOptions(type: 'json' | 'urlencoded' | 'text', options: RequestBodyParserOptions) {
  // the implementation
}
```

(2) I also find it difficult to build understanding of what is this function actually doing. Could you please add some comments explaining the purpose and showing example input & outputs?

(3) How do you expect custom body-parsers to parse the options? Should we improve `getParserOptions` to allow extension to leverage the same convention for processing parser options as the built-in extensions do?

packages/rest/src/body-parser.ts
```diff
@@ -0,0 +1,287 @@
...
+  for (const k of ['json', 'urlencoded', 'text']) {
+    delete opts[k];
+  }
+  return opts as ParserOption<T>;
+}
+
+/**
+ * Interface to be implemented by body parser extensions
+ */
+export interface BodyParser {
```
bajtos: Please move this interface definition to the top of this source file. It's one of the most important public artifacts provided by this file. It's a popular convention to put public artifacts at the top and implementation details below.

Even better, create a new directory for all things related to body parsing and put the interface into a new "types" file used by body parsing, i.e. `packages/rest/src/body-parser/body-parser.types.ts`

packages/rest/src/body-parser.ts
```diff
@@ -0,0 +1,287 @@
...
+    );
+    this.urlencodedParser = urlencoded(urlencodedOptions);
+
+    const textOptions = Object.assign(
+      {type: 'text/*', limit: DEFAULT_LIMIT},
+      getParserOptions('text', options),
+    );
+    this.textParser = text(textOptions);
+
+    this.parsers = parsers.concat([
```
bajtos: 1. Can applications disable certain body parsers? If not, can they provide custom implementation for given media types?

2. I find this implementation a bit hacky in the sense that it has different internal structure from what parsers contributed by extensions would use. Please rework it so that internal parsers look the same way as external parsers would do. Ideally, each built-in parser should have its own source code file, e.g. `packages/rest/src/body-parser/json.body-parser.ts`.

I hate to repeat myself so often, but: it would be much easier for both of us if one pull request contained one logical change only.

packages/rest/src/body-parser.ts
```diff
@@ -0,0 +1,287 @@
...
+    // Check of the request content type matches one of the expected media
+    // types in the request body spec
+    let matchedMediaType: string | false = false;
+    for (const type in content) {
+      matchedMediaType = is(contentType, type);
+      if (matchedMediaType) {
+        debug('Matched media type: %s -> %s', type, contentType);
+        requestBody.mediaType = type;
+        requestBody.schema = content[type].schema;
+        // Skip body parsing as the controller method wants to have full control
```
bajtos: Now that we allow custom body parsers, I am no longer convinced it's a good idea to allow controllers to have full control over body parsing.

What uses case do you have for this feature? In what circumstances would an application or an extension need to parse the body directly and would not be able to install custom body parser instead?

packages/rest/src/providers/parse-params.provider.ts
```diff
@@ -17,12 +22,18 @@ import {Request, ParseParams, RequestBodyParserOptions} from '../types';
...
+
   constructor(
     @inject(RestBindings.REQUEST_BODY_PARSER_OPTIONS, {optional: true})
-    private options?: RequestBodyParserOptions,
-  ) {}
+    options?: RequestBodyParserOptions,
+    @inject.tag(REQUEST_BODY_PARSER_TAG, {optional: true})
+    parsers: BodyParser[] = [],
+  ) {
+    this.requestBodyParser = new RequestBodyParser(options, parsers);
```
bajtos: Hang on. Why are we instantiating BodyParser directly when the class is already wired up for dependency injection?

IMO, we should receive `RequestBodyParser` via dependency injection. I suspect this may simplify unit testing?

Alternatively, we can consider `RequestBodyParser` as an implementation detail that is not customizable, in which case I don't see why should RequestBodyParser be decorated for dependency injection.

Let's discuss.

packages/rest/test/acceptance/validation/validation.acceptance.ts
```diff
@@ -65,7 +65,10 @@ describe('Validation at REST level', () => {
     it('rejects requests with no (empty) body', async () => {
       // NOTE(bajtos) An empty body cannot be parsed as a JSON,
       // therefore this test request does not even reach the validation logic.
-      await client.post('/products').expect(400);
+      await client
+        .post('/products')
+        .type('json')
+        .expect(422);
```
bajtos: With this change in place, the comment at the top of the test is inconsistent with what is the test asserting. Please take a closer look, I would like us to understand how are empty payloads handled with the new request handler in place.

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -393,6 +450,83 @@ describe('HttpHandler', () => {
     }
   });
 
+  context('multipart/form-data', () => {
```
bajtos: As I mentioned before, IMO the test for multipart is rather complex and unnecessarily complex for integration tests. I would like to see an acceptance-level test for this feature.

Also, I don't think we should be telling LB4 users to handle file uploads by calling `multer` directly from the controller. Instead, we should teach them how to build a custom body parser using multer under the hood.

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -324,14 +347,45 @@ describe('HttpHandler', () => {
...
+        supports(mediaType: string) {
+          return !!is(mediaType, 'xml');
+        }
+
+        async parse(request: Request) {
+          return {value: {key: 'value'}};
+        }
+      }
+
+      // Register a request body parser for xml
```
bajtos: (1) Please add a sugar API for registering body parsers. 

Typical usage in a REST application:

```ts
class TodoApplication extends /*mixins*/ RestApplication {
  constructor() {
    // ...
  
    this.bodyParser(XmlBodyParser);
  }
}
```

(2) Add a new test (or extend this one) to show the idiomatic solution for passing configuration options (e.g. max body size) to custom request parsers.

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -393,6 +450,83 @@ describe('HttpHandler', () => {
...
+      const spec = anOpenApiSpec()
+        .withOperation('post', '/show-body', {
+          'x-operation-name': 'showBody',
+          requestBody: <RequestBodyObject>{
+            description: 'multipart/form-data value.',
+            required: true,
+            content: {
+              'multipart/form-data': {
+                // Skip body parsing
+                'x-skip-body-parsing': true,
```
bajtos: Please add a simpler test to verify `x-skip-body-parsing` feature. One that has less moving parts and will be easier to debug if anything goes wrong.

docs/site/Parsing-requests.md
```diff
@@ -179,12 +179,13 @@ in/by the `@requestBody` decorator. Please refer to the documentation on
...
-`application/x-www-form-urlencoded`. Its value is matched against the list of
-media types defined in the `requestBody.content` object of the OpenAPI operation
-spec. If no matching media types is found or the type is not supported yet, an
-UnsupportedMediaTypeError (http statusCode 415) will be reported.
+We support `json`, `urlencoded`, and `text` content types. The client should set
+`Content-Type` http header to `application/json`,
+`application/x-www-form-urlencoded`, or `text/plain`. Its value is matched
+against the list of media types defined in the `requestBody.content` object of
+the OpenAPI operation spec. If no matching media types is found or the type is
+not supported yet, an UnsupportedMediaTypeError (http statusCode 415) will be
```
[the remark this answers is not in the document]
raymondfeng: Fixed

packages/rest/src/body-parser.ts
raymondfeng: This is just for internal code reuse. Fixed

packages/rest/src/body-parser.ts
raymondfeng: Fixed

packages/rest/src/body-parser.ts
raymondfeng: 1 & 2 are fixed.

I'm sorry that I had to combine a few things in a single PR as they build on top of each other. Maintaining 3 PRs (such as pr3 -> pr2 --> pr1) is hard for me too. I intentionally keep them as separate commits.

packages/rest/src/body-parser.ts
raymondfeng: I don't think the parsers can handle all cases or are optimal. Allowing a controller method to fully control how it's handled is desirable, for example, a method may want to streamline large json request processing.

packages/rest/src/providers/parse-params.provider.ts
raymondfeng: Switching to DI.

packages/rest/test/integration/http-handler.integration.ts
raymondfeng: This is a just a starting point to provide `multipart/form-data` integration.

raymondfeng:
@bajtos PTAL

packages/rest/src/body-parser.ts
bajtos: I was thinking about this some more and I agree with you that this feature makes sense. I'd like us to reframe it a bit though. Instead of focusing on implementation details (skipping body parsing) & describing what the app/extension developer does not want to, can we describe this feature from the point of what they actually want to achieve? I.e. receive the request body as a stream.

In that light, I am proposing to:

Find a different way how developers can tell the framework that they want to receive the request body as a (raw/original) stream. 
- At minimum, we should rename `x-skip-body-parsing`, e.g. to something like `x-stream: true`. I think it will be even better to use an enum instead of a boolean flag, e.g. `x-process-as: stream`. It will allow us to provide other ways for consuming request bodies, e.g. consume large JSON arrays via async iterators (`x-process-as: iterator`). I cannot come up with a good name for this new option right now, please try to find a better one yourself.
- Consider adding a higher-level sugar API at `@requestBody` decorator. I am not sure how easy this would be to design and I am fine to leave it out of scope of this pull request, but we should at least consider the user/developer experience of people using this feature in their projects.

In the documentation for this new feature (which you still need to add AFAICT), use the scenario where the app/extension developer wants to consume the request body as a stream to drive the narrative. I can imagine this new documentation can take a form of a new page in the section "How tos", plus a short mention of the new extension flag in apidocs and/or `@requestBody` overview.

packages/rest/test/integration/http-handler.integration.ts
bajtos: I think we may not understand each other.

As I understand this pull request, it's adding the following two extension points:
- pluggable body parsers
- consume request body as a stream
while using "file upload" scenario as an acceptance criteria to verify the implementation.

I am asking you to:
1. Implement the simplest possible integration test for consuming request body as a stream. One that does not rely on complex stuff like multipart processing. When this integration test fail, it must be easy to identify the source of the problem. With multipart in play, it's difficult to tell whether the problem is in our framework or in the interaction with multer.

2. Write an acceptance test to show how this new feature can be used to implement file uploads. Essentially, move the test code inside `context('multipart/form-data')` into a new file in the [test/acceptance](`https://github.com/strongloop/loopback-next/tree/b090e0706c451df57deebc2bd92595787e0606d2/packages/rest/test/acceptance) directory and beef little bit up to work at RestApplication level.

bajtos: I am totally fine to keep `multipart/form-data` integration very minimal in this pull request. I actually love it how this pull request is reframing file-upload support from a built-in feature (as proposed in https://github.com/strongloop/loopback-next/pull/1989) into an acceptance test serving as a possible reference implementation allowing anybody to implement any custom file-upload handing.

bajtos: As I am thinking about this more, it would be even more cool if we could have two acceptance tests for file upload:
- one showing how to handle file uploads by registering a custom request body parser
- the other showing how to handle file uploads in individual controller methods by receiving the request body as a stream.

packages/rest/src/body-parsers/body-parser.json.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {RequestBodyParserOptions} from '../types';
+
+import {json} from 'body-parser';
+import {inject} from '@loopback/context';
+import {Request} from '../..';
```
bajtos: I am worrying about circular dependencies. Can re-arrange type definitions so that deep files like this one can import `Request` from a non-index file, e.g. `../types`?

packages/rest/src/body-parsers/body-parser.json.ts
```diff
@@ -0,0 +1,45 @@
...
+import {RestBindings} from '../keys';
+import {is} from 'type-is';
+
+import {RequestBody, BodyParser} from './types';
+import {
+  BodyParserWithCallback,
+  DEFAULT_LIMIT,
+  getParserOptions,
+  parseRequest,
+} from './helper';
```
bajtos: I believe we are trying to keep filenames unique so that we don't depend on file paths to distinguish between them. Can you rename this file to something like `body-parser.helpers.ts` please?

packages/rest/src/body-parsers/body-parser.json.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class JsonBodyParser implements BodyParser {
+  name = 'json';
+  private jsonParser: BodyParserWithCallback;
+
+  constructor(
+    @inject(RestBindings.REQUEST_BODY_PARSER_OPTIONS, {optional: true})
+    options: RequestBodyParserOptions = {},
+  ) {
+    const jsonOptions = Object.assign(
```
bajtos: Can you move this part into `getParserOptions` too, to simplify our parsers into something like this:

```ts
const jsonOptions = getParserOptions(
  'json', // the key (required)
  {type: 'json', limit: DEFAULT_LIMIT}, // the defaults (required)
  options // user-provided options (optional)
);
```

Do we actually have to explicitly set `{type: 'json'}`? Can `getParserOptions` set that property for us? And how about `limit`, isn't this something that `getParserOptions` can provide default value for too?

```ts
const jsonOptions = getParserOptions(
  'json', // the key (required)
  {limit: DEFAULT_LIMIT}, // the defaults (required)
  options // user-provided options (optional)
);

File: packages/rest/src/body-parsers/body-parser.text.ts [hunk 15]
```diff
@@ -0,0 +1,45 @@
...
+export class TextBodyParser implements BodyParser {
+  name = 'text';
+  private textParser: BodyParserWithCallback;
+
+  constructor(
+    @inject(RestBindings.REQUEST_BODY_PARSER_OPTIONS, {optional: true})
+    options: RequestBodyParserOptions = {},
+  ) {
+    const textOptions = Object.assign(
+      {type: 'text/*', limit: DEFAULT_LIMIT},
```
bajtos: Interesting. JsonParser is using `type: 'json'` but TextParse is using `type: 'text/*'`. Cannot we use `type: 'text'` too? I am not familiar with body-parser internals and may be wrong, but still wants to ask.

packages/rest/src/body-parsers/body-parser.text.ts
```diff
@@ -0,0 +1,45 @@
...
+      getParserOptions('text', options),
+    );
+    this.textParser = text(textOptions);
+  }
+
+  supports(mediaType: string) {
+    return !!is(mediaType, 'text/*');
```
bajtos: Similarly here - why is JsonParser using `json` but TextParser `text/*`?

packages/rest/src/body-parsers/body-parser.ts
```diff
@@ -0,0 +1,115 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import * as debugModule from 'debug';
+import {Request, RequestBodyParserOptions} from '../types';
+import {inject} from '@loopback/context';
+import {isReferenceObject, OperationObject} from '../..';
```
bajtos: Let's not import from index files please, we don't want to run into cyclic dependencies. Import from a specific non-index file instead. I think these imports should be from `@loopback/openapi-v3` or similar?

packages/rest/src/body-parsers/helper.ts
```diff
@@ -0,0 +1,97 @@
...
+export function parseRequest(
+  handle: BodyParserWithCallback,
+  request: Request,
+): Promise<any> {
+  // A hack to fool TypeScript as we don't need `response`
+  const response = ({} as any) as Response;
+  return new Promise<void>((resolve, reject) => {
+    handle(request, response, err => {
+      if (err) {
+        reject(normalizeParsingError(err));
```
bajtos: Do we actually need to normalize the error here? Personally, I'd prefer to rely on `RequestBodyParser` to do the normalization, to ensure we apply the same normalization on errors thrown by custom parsers that don't call `parseRequest` under the hood.

Also, please add a test to verify that when a custom parser throws an error, it is normalized as expected.

packages/rest/src/body-parsers/helper.ts
```diff
@@ -0,0 +1,97 @@
...
+  request: Request,
+  response: Response,
+  callback: (err: HttpError) => void,
+) => void;
+
+/**
+ * Normalize parsing errors as `4xx`
+ * @param err
+ */
+export function normalizeParsingError(err: HttpError) {
```
bajtos: As I mentioned in the other comment, I think error normalization should be handled in a single place and that's in `RequestBodyParser`. Let's move this helper function to `body-parser.ts`.

packages/rest/src/body-parsers/helper.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// tslint:disable:no-any
+
+/**
+ * Parse the body asynchronously
+ * @param handle The express middleware handler
+ * @param request Http request
+ */
+export function parseRequest(
```
bajtos: I find this name misleading because the function is not doing any body parsing, just executing a body-parsing middleware in a way that's async/await friendly. Let's find a name that described this intent better. 

For example: `executeBodyParserMiddleware`.

I think this function can be very useful to extension authors building custom body parsers, because it effectively allows them to take any Express middleware performing body parsing and wrap it into LB4 RequestParser. It would be great to change it into a well-documented public API.

packages/rest/src/body-parsers/helper.ts
```diff
@@ -0,0 +1,97 @@
...
+  type: 'urlencoded',
+  options: RequestBodyParserOptions,
+): OptionsUrlencoded;
+export function getParserOptions(
+  type: 'text',
+  options: RequestBodyParserOptions,
+): OptionsText;
+
+export function getParserOptions<T extends 'json' | 'urlencoded' | 'text'>(
+  type: T,
```
bajtos: Nitpick - I think these two lines can be simplified as follows:

```ts
export function getParserOptions(
  type: 'json' | 'urlencoded' | 'text',
```

packages/rest/src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './router';
 export * from './providers';
 
 export * from './parser';
+export * from './body-parsers';
```
bajtos: I have mixed feelings about using plural form here, perhaps `./body-parser` or `./body-parsing` would be better? Just an idea to consider, I am not sure what's the best name.

packages/rest/src/keys.ts
```diff
@@ -85,10 +86,32 @@ export namespace RestBindings {
...
   >('rest.requestBodyParserOptions');
 
+  /**
+   * Binding key for request body parser
+   */
+  export const REQUEST_BODY_PARSER = BindingKey.create<RequestBodyParser>(
+    'rest.requestBodyParser',
+  );
+
+  export const REQUEST_BODY_PARSER_JSON = BindingKey.create<BodyParser>(
```
bajtos: Is it a good idea to turn binding keys of individual parsers into a public API? Personally, I would expect that these keys should be treated as an implementation detail, similarly to how we are not making much guarantees about the binding keys for controllers, repositories and routes.

packages/rest/src/parser.ts
```diff
@@ -89,106 +53,6 @@ export async function parseOperationArgs(
...
 
-function normalizeParsingError(err: HttpError) {
-  debug('Cannot parse request body %j', err);
-  if (!err.statusCode || err.statusCode >= 500) {
-    err.statusCode = 400;
-  }
-  return err;
-}
-
-export async function loadRequestBodyIfNeeded(
```
bajtos: I love how shorter and simpler this file has become 👍 ❤️

packages/rest/src/rest.component.ts
```diff
@@ -37,6 +44,26 @@ export class RestComponent implements Component {
...
   };
+  /**
+   * Add built-in body parsers
+   */
+  bindings = [
+    Binding.bind(RestBindings.REQUEST_BODY_PARSER)
+      .toClass(RequestBodyParser)
+      .inScope(BindingScope.SINGLETON),
+    Binding.bind(RestBindings.REQUEST_BODY_PARSER_JSON)
+      .toClass(JsonBodyParser)
```
bajtos: +1 for using the new `bindings` API.

I wish we could reuse the implementation from `RestServer#bodyParser` here. Can we perhaps at least re-implement `RestServer#bodyParser` as a private helper method in the component class (just copy&paste the implementation details) and simplify this long registration block as follows?

```ts
export class RestComponent implements Component {
  // ...
  bindings: [];

  constructor(/*...*/) {
    // ...

    this.bodyParser(JsonBodyParser);
    this.bodyParser(TextBodyParser);
    this.bodyParser(UrlEncodedBodyParser);
  }
}
```


Also, how do we expect 3rd-party extensions to contribute custom body parsers? I think we should provide them with a helper that will construct a binding with the correct tag, scope and inferred key for them. To keep this pull request simple, I am proposing to write a new helper function that will accept a parser constructor and return a Binding. This function should be used by `RestServer#bodyParser`. As for `RestComponent`, we can provide a helper method `bodyParser` calling this helper function under the hood, or simply call the helper to create Binding instances to initialize `bindings` array. For example:

```ts
export class RestComponent implements Component {
  // ...
  bindings: [
    Binding.bind(RestBindings.REQUEST_BODY_PARSER)
       .toClass(RequestBodyParser)
       .inScope(BindingScope.SINGLETON),
    createBodyParserBinding(JsonBodyParser),
    createBodyParserBinding(TextBodyParser),
    createBodyParserBinding(UrlEncodedBodyParser),
  ];

  // ...
}
```

I think I like this last approach most, because it's using the same constructs that any 3rd-party extension author can use too.

We need to start writing a guide for extension developers to show them how to contribute custom body parsers, the guide should mention this new helper function for creating bindings too.

packages/rest/src/rest.server.ts
```diff
@@ -716,6 +717,16 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this.sequence(SequenceFromFunction);
   }
 
+  /**
+   * Bind a body parser to the server context
+   * @param bodyParserClass
+   */
+  bodyParser(bodyParserClass: Constructor<BodyParser>) {
+    this.bind(`${RestBindings.REQUEST_BODY_PARSER}.${bodyParserClass.name}`)
```
bajtos: Let's extract this logic into a public helper function that 3rd-party components can use too.

```ts
bodyParser(bodyParserClass: Constructor<BodyParser>) {
  this.add(createBodyParserBinding(bodyParserClass));
}
```

packages/rest/test/acceptance/validation/validation.acceptance.ts
bajtos: PTAL ☝️

bajtos:
@raymondfeng I would like you to consider the following high-level scenarios, I am not sure if the current design for body-parser registration is flexible enough to support them.

While the canonical content-type for JSON payloads is `application/json`, the alternative `text/json` used to be pretty popular. (See https://stackoverflow.com/q/477816/69868 for more alternative types). Should LB4 parse `text/json` as JSON, not as plain text? If yes, then please add a test for this.

This scenario came to my attention because I incorrectly assumed that `text/json` would be accepted by both JsonParser and TextParser. It turns out that `typeis.is(contentType, 'json')` does not recognize `text/json`, so this is not really a concern right now. But:

**It opens a more generic question of the priority of body parsers.**

Let's say the user has two parsers, one recognizing `*/*+json` (a generic JSON parser) and the other one recognizing `application/vnd.api+json` (a custom parser for JSON-API payloads). How can the developer control which parser will take precedence? What if the generic JSON parser is contributed by `@loopback/rest` and the JSON-API parser is contributed by another component (e.g. `@loopback/jsonapi`)? What if both parsers are contributed by 3rd party components that are not aware of each other?

Do we need to sort the parsers from most-specific to most-generic?

To support all JSON-based content-types, should we broaden content types accepted by our built-in parser to allow any of `['*/json', 'application/*+json']`, for example `text/json` and `application/vnd.api+json`?

packages/rest/src/body-parsers/body-parser.json.ts
raymondfeng: > The code in individual built-in parser classes seems to be duplicated a lot, have you considered using a Strategy or Template design pattern to extract a shared base class?

The size of each class is fairly small as most of the common logic have been pushed into the helper class. Introducing a new base class seems to overkill and add complexity, IMO.

raymondfeng:
@bajtos We can possibly use the http `accept` header format to define the q-factor of accepted media types for any given body parser - see https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html. For example, a JSON body parser can declare:

```
accept: 'application/vnd.x.json; q=0.5, application/json'
```

Such features can be done outside of this PR.

Please note that http content negotiation utilizes the `Accept` header for matching response media types. We can borrow the idea.

bajtos:
> We can possibly use the http accept header format to define the q-factor of accepted media types for any given body parser.

I am not sure if that will work, to be honest. 

> Such features can be done outside of this PR.

I am fine to leave it outside of this PR, but I am concerned that before we do at least some minimal implementation, then we will not know whether the proposed approach is actually feasible :(

Can we at least put the built-in parsers to the end of the list, so that user-defined parsers always take precedence? It may be enough to sort the list of injected parsers in reverse, so that the last parser registered will be the first parser to have a chance to process the request.

bajtos:
I was thinking about this problem more and have another idea to consider.

In Express, middleware preprocessing requests (e.g. parsing the body) does not have any route-specific metadata to use as a guide to decide which behavior to use (which parser to execute). In LoopBack 4, routes provide OpenAPI spec which is allows us to provide additional instructions for body parsing (accepted content types + extra metadata for each type, e.g. schema).

So I was thinking about the following solution, which will cover both selection of the right parser and the ability to skip body parsing.

1. Restrict the built-in body parsers to a single type only (`text/plain`, `application/javascript`, etc.).
2. Define a new OpenAPI extension allowing developers to choose which parser to use. This extension can be called e.g. `x-parser` and allow string values (parser names) & parser classes (constructor functions). We can define a special parser called e.g. `stream` that will be a no-op.
3. When the operation provides `x-parser` setting, we skip content-type matching algorithm and use the selected parser regardless of whether it is accepting the actual content-type.

Few examples:

```ts
// parse text/json as JSON
@requestBody({
  content: {
    'text/json': {
      'x-parser': 'json',
      schema: {/*...*/},
    },
  }
})

// parse custom type as text
@requestBody({
  content: {
    'text/calendar': {
      'x-parser': 'text',
      schema: {/*...*/},
    },
  },
})

// consume unsupported type as a stream (skip body parsing)
@requestBody({
  content: {
    'multipart/form-data': {
      'x-parser': 'stream',
      schema: {(/*...*/},
    },
  },
});

// override handling of a known type - consume the body as a stream
@requestBody({
  content: {
    'application/json': {
      'x-parser': 'stream',
      schema: {(/*...*/},
    },
  },
});

// provide a custom parser class
@requestBody({
  content: {
    'multipart/form-data': {
      'x-parser': FileUploadParser,
      schema: {(/*...*/},
    },
  },
});

class FileUploadParser implements BodyParser {
  // ...
}
```

I find this approach very elegant, because we can leverage the same mechanism (`x-parser` extension) for both customizing which parser is used and disabling body parsing entirely. It also avoids automagic algorithm for deciding which parser to use, thus it should make life easier both for our users (it's easier to learn and understand how parser selection works, because it's explicit) and us, the maintainers (less confused users = less time spent on support, no need to build & maintain complex code like handling of weight parameter `q`).

We can even push this approach one step further and always use "stream" (no-op) body parser for content types that don't have any matching parser. I.e. introduce a new default behavior where bodies with unsupported content types are automatically passed to controller methods/route handlers as streams.

This is based on the assumption that our body parsing step is accepting only the content types described in the OpenAPI spec, and returning "Unsupported Media Type" for requests with other content types.

@raymondfeng @strongloop/loopback-next thoughts?

raymondfeng:
@bajtos I have implemented the `x-parser` proposal. PTAL.

packages/rest/src/rest.component.ts
```diff
@@ -37,6 +49,21 @@ export class RestComponent implements Component {
...
     [RestBindings.SequenceActions.SEND.key]: SendProvider,
   };
+  /**
+   * Add built-in body parsers
+   */
+  bindings = [
+    Binding.bind(RestBindings.REQUEST_BODY_PARSER)
+      .toClass(RequestBodyParser)
+      .inScope(BindingScope.SINGLETON),
+    createBodyParserBinding(JsonBodyParser).inScope(BindingScope.SINGLETON),
```
bajtos: IMO, body parsers should be singletons by default and thus `createBodyParserBinding` should be setting the scope for all bindings it creates.

What use cases do you see for non-singleton parsers? 

Even when we change `createBodyParserBinding` to set the scope to SINGLETON, non-singleton parsers can still override it:

```ts
createBodyParserBinding(TransientParser).inScope(BindingScope.TRANSIENT)
```

packages/rest/src/types.ts
```diff
@@ -80,23 +86,15 @@ export type LogError = (
...
-   */
-  encoding?: string;
-  [property: string]: any;
-};
+export interface RequestBodyParserOptions extends Options {
+  json?: OptionsJson;
+  urlencoded?: OptionsUrlencoded;
+  text?: OptionsText;
+  raw?: Options;
+  [name: string]: any;
```
bajtos: Have you considered using a more specific type, e.g. `object` or `unknown`?

packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts
```diff
@@ -0,0 +1,159 @@
...
+  beforeEach(givenClient);
+  beforeEach(givenBodyParamController);
+
+  it('supports file uploads', async () => {
+    const FIXTURES = path.resolve(__dirname, '../../../../fixtures');
+    const res = await client
+      .post('/show-body')
+      .field('user', 'john')
+      .field('email', 'john@example.com')
+      .attach('certFile', path.resolve(FIXTURES, 'cert.pem'), {
```
bajtos: I find it disturbing to see a code uploading private certificate to a server.

Have you considered using a hard-coded buffer instance instead? For example:

```ts
const TEST_DATA = Buffer.from([0x01, 0x02, 0x03]);

// ...
.attach('data.bin', TEST_DATA, { /* ... */ })

File: packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts [hunk 30]
```diff
@@ -0,0 +1,159 @@
...
+        filename: 'cert.pem',
+        contentType: 'multipart/form-data',
```
bajtos: Interesting, is the `pem` format a `multipart/form-data` too? Shouldn't we specify `application/octet-stream` instead?

packages/rest/src/body-parsers/body-parser.raw.ts
```diff
@@ -0,0 +1,42 @@
...
+    options: RequestBodyParserOptions = {},
+  ) {
+    const rawOptions = getParserOptions('raw', options);
+    this.rawParser = raw(rawOptions);
+  }
+
+  supports(mediaType: string) {
+    // Return false so that it won't be used by matching
+    // only `x-parser: 'raw' will trigger this parser
+    return false;
```
bajtos: An idea to consider: should we parse `application/octet-stream` types into buffers by default, similarly to how we are parsing `text/plain` into strings?

packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts
```diff
@@ -0,0 +1,159 @@
...
+    rootContext
+      .bind(SequenceActions.LOG_ERROR)
+      .to(createUnexpectedHttpErrorLogger());
+    rootContext.bind(SequenceActions.SEND).to(writeResultToResponse);
+    rootContext.bind(SequenceActions.REJECT).toProvider(RejectProvider);
+
+    rootContext.bind(RestBindings.SEQUENCE).toClass(DefaultSequence);
+    rootContext
+      .bind(RestBindings.REQUEST_BODY_PARSER)
+      .toClass(RequestBodyParser);
```
bajtos: Wow, that's a lot of work to setup the context. Since this is an acceptance-level test, can we use `RestApplication` please?

That way we can use regular `@requestBody` decorator and `app.controller` instead of `givenControllerClass`

packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    handler = new HttpHandler(rootContext);
+    rootContext.bind(RestBindings.HANDLER).to(handler);
+  }
+
+  function givenClient() {
```
bajtos: Similarly here. Since this is an acceptance test, if you use `RestApplication` as the root context, then you can use testlab's client factory functions too.

The original setup code was used in integration-level tests where I did not want to use full application originally. Maybe it's time to revisit that decision and simplify our integration-level tests too, but such refactoring/cleanup out of scope of this pull request. However, that does not mean that new test code should be repeating old mistakes and introducing new tech debt. So let's use RestApplication + testlab helpers in the acceptance tests please.

packages/rest/test/acceptance/validation/validation.acceptance.ts
bajtos: Still not addressed ☝️

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -274,12 +280,28 @@ describe('HttpHandler', () => {
...
+            statusCode: 415,
+          },
+        });
+    });
+
+    it('rejects unmatched request body', () => {
+      logErrorsExcept(415);
+      return client
+        .post('/show-body')
+        .set('content-type', 'application/x-xml')
```
bajtos: `application/x-xml` is so close to `application/xml` that it's difficult to spot this difference when reading the code. Please use a distinctively different content type, e.g. `text/plain`. This will also verify that we are checking the spec to determine which types are accepted, even for types that we know how to parse.

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -324,14 +346,45 @@ describe('HttpHandler', () => {
...
+          return {value: {key: 'value'}};
+        }
+      }
+
+      // Register a request body parser for xml
+      rootContext
+        .bind('rest.requestBodyParsers.xml')
+        .toClass(XmlBodyParser)
+        // Make sure the tag is used
+        .tag(REQUEST_BODY_PARSER_TAG);
```
bajtos: Please leverage `createBodyParserBinding` here.

```ts
rootContext.add(createBodyParserBinding(XmlBodyParser));
```

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -393,6 +449,115 @@ describe('HttpHandler', () => {
     }
   });
 
+  context('x-skip-body-parsing', () => {
```
bajtos: I believe `x-skip-body-parsing` was superseded by `x-parser: stream`. The tests below seem to be valid, so I think you just need to change the name of this test context?

```suggestion
  context('x-parser flag', () => {
```

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -324,14 +346,45 @@ describe('HttpHandler', () => {
...
-        .to({limit: 4 * 1024 * 1024}); // Set limit to 4MB
+        .to({limit: '4mb'}); // Set limit to 4MB
       return client
         .post('/show-body')
         .set('content-type', 'application/json')
         .send(body)
         .expect(200, body);
     });
 
+    it('allows request body parser extensions', () => {
```
bajtos: Since most of the body parsing behavior was extracted from HttpHandler into a new BodyParser class, you should also move the relevant tests from `test/integration/http-handler.integration.ts` to a new test file specific for the body-parser implementation. 

It makes sense to keep few body-parsing tests in the integration test suite for HttpHandler to ensure that all parts fit together well, but most of the tests should be at body-parser level.

I think it may be possible to test a lot of body-parsing functionality at unit-test level now, that should simplify the test code a lot.

Paraphrasing https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html:

To create a great test suite, think smaller and favor fast, focused unit-tests over slow application-wide end-to-end tests.

- One “acceptance test”, where you start the application, make an HTTP request to execute the tested use case and verify the expected HTTP response was returned. This verifies that all parts of the implementation are correctly wired together. (You will probably need more than on acceptance-level test for a feature this big, but I hope you see the point.)

- Few “integration tests” to test integration between a subset of components involved. E.g. BodyParser + individual parser implementations.

- Many “unit tests” where you test individual classes in isolation and verify they handle all different situations, including error paths and edge cases.

packages/rest/test/unit/coercion/utils.ts
```diff
@@ -15,6 +15,7 @@ import {
   ResolvedRoute,
   Route,
 } from '../../..';
+import {RequestBodyParser} from '../../../src';
```
bajtos: Please merge this import with the previous one (import from `../../..`).

packages/rest/test/unit/parser.unit.ts
```diff
@@ -54,11 +57,51 @@ describe('operationArgsParser', () => {
     });
     const route = givenResolvedRoute(spec);
 
-    const args = await parseOperationArgs(req, route);
+    const args = await parseOperationArgs(req, route, requestBodyParser);
 
     expect(args).to.eql([{key: 'value'}]);
   });
 
+  it('parses body string as json', async () => {
```
bajtos: This is effectively an integration test, because you are checking how `parseOperationArgs` integrates with `requestBodyParser`. Please move it to `test/integration` directory.

Please review your overall testing strategy and decide what should be covered at what level. For example, I think the next test "parses body number as json" does not need to go through `parseOperationArgs` at all, it should be enough to test this at requestBodyParser level, or even in a unit-test for JsonBodyParser.

packages/rest/test/unit/parser.unit.ts
```diff
@@ -363,6 +393,75 @@ describe('operationArgsParser', () => {
     }
   });
 
+  describe('body parser', () => {
```
bajtos: please move these tests into a new test file, e.g. `packages/rest/test/unit/request-body-parser.unit.ts`

packages/rest/src/index.ts
bajtos: ☝️

packages/rest/src/body-parsers/body-parser.json.ts
bajtos: @raymondfeng your answer above seems to be related to a different comment, could you PTAL?

packages/rest/src/body-parsers/body-parser.text.ts
bajtos: > If you decide to keep `text/*`, then please add a test showing what happens when the user registers a parser that accepts a more specific text-based type, e.g. `text/json` or `text/html`.

This is basically what I repeated in my review comment later, where I proposed `text/json`.

bajtos:
Cross-posting a part of one of the line comments that I consider as still relevant:

It would be even more cool if we could have two acceptance tests for file upload:

- one showing how to handle file uploads by registering a custom request body parser
- the other showing how to handle file uploads in individual controller methods by receiving the request body as a stream.

bajtos:
@raymondfeng To make my further reviews easier, please don't amend any existing commits and push new commits only. (Rebasing on top of master is fine.) That way I can review only new increments and don't have to re-read all ~2k lines again.

Considering the scope of the changes proposed in this pull request, I am not sure if it still makes sense to preserve the current series of smaller commits. If it's easier for you to stop splitting the changes into small commits, with the vision of landing this entire PR in one or two big commits, then feel free to do so. (As long as I have an easy way how to see what has changed since my last review.)

docs/site/Parsing-requests.md
```diff
@@ -260,6 +269,126 @@ A few tips worth mentioning:
...
+      required: true,
+      content: {
+        'multipart/form-data': {
+          // Skip body parsing
+          'x-parser': 'stream',
+          schema: {type: 'object'},
+        },
+      },
+    })
+    request: Request,
```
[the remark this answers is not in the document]
raymondfeng: I would keep `request: Request` as is. Most express middleware expects `Request`.

packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts
raymondfeng: Please note `cert.pem` is the public key.

packages/rest/src/body-parsers/body-parser.helpers.ts
```diff
@@ -0,0 +1,107 @@
...
+import * as debugModule from 'debug';
+import {HttpError} from 'http-errors';
+import {Request, Response, RequestBodyParserOptions} from '../types';
+
+import {
+  OptionsJson,
+  OptionsUrlencoded,
+  OptionsText,
+  Options,
+} from 'body-parser';
```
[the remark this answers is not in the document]
raymondfeng: Fixed

packages/context/src/binding.ts
```diff
@@ -443,7 +443,7 @@ export class Binding<T = BoundValue> {
    * easy to read.
    * @param key Binding key
    */
-  static bind(key: string): Binding {
-    return new Binding(key);
+  static bind(key: BindingAddress<unknown>): Binding {
```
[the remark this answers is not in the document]
raymondfeng: Fixed

packages/rest/test/integration/http-handler.integration.ts
raymondfeng: Fixed

packages/rest/test/unit/parser.unit.ts
raymondfeng: Refactored as body-parser unit tests

packages/rest/test/unit/parser.unit.ts
raymondfeng: Fixed

raymondfeng:
> It would be even more cool if we could have two acceptance tests for file upload:
one showing how to handle file uploads by registering a custom request body parser
the other showing how to handle file uploads in individual controller methods by receiving the request body as a stream.

See `test/acceptance/file-upload`.

packages/rest/src/body-parsers/body-parser.text.ts
raymondfeng: I extended json parser to deal with `*/json` and `*/*+json`.

raymondfeng:
@bajtos PTAL.

packages/rest/src/body-parsers/body-parser.raw.ts
```diff
@@ -30,13 +30,13 @@ export class RawBodyParser implements BodyParser {
   }
 
   supports(mediaType: string) {
-    // Return false so that it won't be used by matching
-    // only `x-parser: 'raw' will trigger this parser
+    // Return `false` to only allow `{x-parser: 'raw'} to trigger this parser
+    // It won't be used for
```
bajtos:
```ts
// It won't be used for
```

It won't be used for ... what?

packages/rest/src/body-parsers/body-parser.ts
```diff
@@ -50,18 +53,60 @@ export class RequestBodyParser {
...
+      operationSpec,
+      request,
+    );
+    if (!operationSpec.requestBody) return requestBody;
+    const matchedMediaType = requestBody.mediaType!;
+    try {
+      if (customParser) {
+        // Invoke the custom parser
+        const body = await this._invokeCustomParser(customParser, request);
+        if (body !== undefined) return Object.assign(requestBody, body);
```
bajtos: Please see my earlier comment on this topic.

Why are you falling back to iterate through `this.parsers` when `_invokeCustomParser` returns `undefined`? I find it very confusing.

packages/rest/src/keys.ts
bajtos: @raymondfeng you haven't addressed nor responded to this comment. PTAL again.

packages/rest/src/rest.application.ts
```diff
@@ -97,10 +97,14 @@ export class RestApplication extends Application implements HttpServerLike {
...
    * Bind a body parser to the server context
-   * @param bodyParserClass
+   * @param parserClass Body parser class
+   * @param address Optional binding address
    */
-  bodyParser(bodyParserClass: Constructor<BodyParser>) {
-    this.restServer.bodyParser(bodyParserClass);
+  bodyParser(
+    bodyParserClass: Constructor<BodyParser>,
+    address?: BindingAddress<BodyParser>,
```
bajtos: Can we use `key` instead of `address` for consistency with other places please? Most notably `Context` class uses `key` in most of places.

packages/rest/test/acceptance/file-upload/file-upload-with-parser.acceptance.ts
```diff
@@ -0,0 +1,123 @@
...
+        description: 'multipart/form-data value.',
+        required: true,
+        content: {
+          [FORM_DATA]: {
+            schema: {type: 'object'},
+          },
+        },
+      })
+      body: unknown,
+      @inject(RestBindings.Http.RESPONSE) response: Response,
```
bajtos: AFAICT, `response` is no longer used, can you remove it please?

packages/rest/test/acceptance/request-parsing/request-parsing.acceptance.ts
```diff
@@ -0,0 +1,134 @@
...
+              schema: {type: 'object'},
+            },
+          },
+        })
+        request: // tslint:disable-next-line:no-any
+        any,
+      ): Promise<object> {
+        if (parser === 'stream') {
+          // Request body parsing is skipped
+          expect(request.body).to.be.undefined();
```
bajtos: Having assertions buried so deeply in a test helper makes the test difficult to understand and modify.

Please change this controller to store the `request` value into a shared variable, e.g. `actualRequest` or `requestReceived` and move assertions back to individual test cases.

packages/rest/test/integration/http-handler.integration.ts
```diff
@@ -449,40 +449,28 @@ describe('HttpHandler', () => {
     }
   });
 
-  context('x-skip-body-parsing', () => {
+  context('x-parser extension', () => {
```
bajtos: As I commented before, is this behavior really specific to HttpHandler? IMO, this should be tested in unit or integration tests for RequestParser class.

packages/rest/src/body-parsers/body-parser.text.ts
bajtos: > I extended json parser to deal with */json and */*+json.

I am afraid you missed my point.

I was using `text/json` as an example use case (test case), where the user wants to register a specialized parser for a media type that's already covered by a built-in parser. 

I see that the current implementation sorts the parsers and invokes built-in parsers at the end, so I suppose this is solved (although I didn't check if there is an accompanying test).

docs/site/Parsing-requests.md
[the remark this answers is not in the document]
bajtos: It leaves less maneuvering space for us in the future when the controller methods are expecting a full request object instead of a readable stream.

Note that controllers wishing to receive the full request can already do that via `RestBindings.Http.Request`.

packages/rest/test/acceptance/file-upload/file-upload.acceptance.ts
bajtos: Ah, good point. I still think it's better to build a simple buffer inside the test, rather than relying on external fixtures.

Anyhow, not a blocker.

packages/rest/src/rest.application.ts
raymondfeng: I just checked. The `Context` class `BindingAddress` for most cases.

If you refer to `app.controller(controllerClass, name)` or similar methods, I intentionally make it different so that we can use the well-know binding address to override built-in body parsers. 

BTW, `export const REQUEST_BODY_PARSER_JSON` in `keys.ts` is by design to support the acceptance test of overriding built-in body parsers.

packages/rest/test/integration/http-handler.integration.ts
raymondfeng: Moved to `body-parser.unit.ts`

packages/rest/test/acceptance/request-parsing/request-parsing.acceptance.ts
raymondfeng: Fixed

packages/rest/src/body-parsers/body-parser.ts
raymondfeng: Fixed

packages/rest/src/body-parsers/body-parser.raw.ts
raymondfeng: Fixed

raymondfeng:
> I was using text/json as an example use case (test case), where the user wants to register a specialized parser for a media type that's already covered by a built-in parser.
I see that the current implementation sorts the parsers and invokes built-in parsers at the end, so I suppose this is solved (although I didn't check if there is an accompanying test).

Yes. It's fixed.

1. `text/json` is now parsed by json paraser. See https://github.com/strongloop/loopback-next/blob/switch-to-body-parser/packages/rest/test/unit/body-parser.unit.ts#L91
2. See https://github.com/strongloop/loopback-next/blob/switch-to-body-parser/packages/rest/test/acceptance/request-parsing/request-parsing.acceptance.ts

raymondfeng:
> Also: Documentation wise, I am (still) missing a guide for extension developers showing them how to write an extension contributing custom parsers.

Do you think we need to refactor this section to a separate guide? https://github.com/strongloop/loopback-next/blob/switch-to-body-parser/docs/site/Parsing-requests.md#adding-parsers

raymondfeng:
@bajtos I believe that I have addressed most of your comments. PTAL. If you think there are more blocking issues, let me know.

raymondfeng:
@bajtos ping.

packages/rest/src/body-parsers/body-parser.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  constructor(
+    @inject(RestBindings.REQUEST_BODY_PARSER_OPTIONS, {optional: true})
+    options: RequestBodyParserOptions = {},
+    @inject.tag(REQUEST_BODY_PARSER_TAG, {optional: true})
+    parsers?: BodyParser[],
+    @inject.context() private readonly ctx?: Context,
+  ) {
+    if (!parsers || parsers.length === 0) {
+      this.parsers = [
```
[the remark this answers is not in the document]
bajtos: @raymondfeng let's discuss this part please before landing. If we land this list of default parsers, then it would require a semver-major version to remove it - i.e. it's a change that cannot be easily undone.

bajtos:
> > Also: Documentation wise, I am (still) missing a guide for extension developers showing them how to write an extension contributing custom parsers.
>
> Do you think we need to refactor this section to a separate guide? https://github.com/strongloop/loopback-next/blob/switch-to-body-parser/docs/site/Parsing-requests.md#adding-parsers

Eventually, I think we will need to split "Parsing requests" into multiple pages, because it's growing too long. BUT: As far as this PR is concerned, I am fine with keeping everything in a single page.

When I put myself in the shoes of an extension developer, I see the following shortcomings:

1. Discoverability. When I go to docs > LoopBack 4 > [Extending LoopBack 4](https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html), I don't see any mention of how to write an extension contributing custom body parsers.
2. The content in "Parsing-requests" seems to be geared towards app developers. I does not show how an extension can export body parsers and have them picked up by the target app. Also how to allow apps to set the configuration options for the custom parser.

Here is what I have in mind:
1. Add a new page, e.g. "Creating custom body parsers", nest it inside "Extending LoopBack 4" section in the sidebar, alongside "Creating Decorators" and other guides for extension developers.
2. In the new page: 

    1. Tell people to read "Parsing-requests" to learn how to write a class implementing a body parser. (Keep it simple, no need to repeat what's written elsewhere. The important part is to make it easier for people to discover the relevant content.) 

    2. Explain how to wire up the extension (a Component class) so that the custom parser is picked up by target applications. Include a code snippet showing how to use `bodyParserBindingKey` helper to create the binding.

    3. Explain how to allow applications to configure the custom body parser (e.g. set the `limit`, etc.). Include a code snippet with an example implementation. (Do you expect extensions to leverage `getParserOptions` helper?)

packages/rest/src/keys.ts
bajtos: Let's discuss this point before landing the pull request. If we wanted to remove these public keys later, then we would have to wait until we are ready for a semver-major version.

bajtos:
@raymondfeng Here is the list of comments where I feel we did not reach consensus yet and where it would be difficult to change the code/APIs later:

- The built-in hard-coded list of parser instances bypassing Dependency Injection. https://github.com/strongloop/loopback-next/pull/1936/files#r233875934
- Exporting binding keys of individual parsers. https://github.com/strongloop/loopback-next/pull/1936/files#r233875934
- The type used by custom extensions in the body parser options. https://github.com/strongloop/loopback-next/pull/1936/files#r233878521
- A possibly breaking change in the way how empty bodies are handled. https://github.com/strongloop/loopback-next/pull/1936/files#r232701765
- A test where it's difficult to spot what's being tested. https://github.com/strongloop/loopback-next/pull/1936/files#r233884268

packages/rest/test/integration/http-handler.integration.ts
raymondfeng: Switch to `text/plain`.

packages/rest/test/acceptance/validation/validation.acceptance.ts
raymondfeng: It turns out that express json body parser creates `{}` for empty body. I added some logic in json parser to treat empty body as `undefined`.

raymondfeng:
@bajtos I have addressed all of your pending comments. PTAL.

raymondfeng:
> The new doc page "Extending-request-body-parsing" is missing instructions on how a component can contribute a new/custom parser binding.

Added.

bajtos:
Yay 🕺 🎉

Issue #1909, opened 2018-10-24
hasMany relations broken with SQL datasources


gerbus:
## Description / Steps to reproduce / Feature proposal
I built a simple API modelled very closely after the ToDo and ToDoList example models. However I used a mysql datasource, and I suspect this caused my relations between repositories/models not to work. I had in my ToDoList model a proper `@hasMany(() => ToDo` decorator. Wired it up in the repository as per documentation instructions. Then I used the API Explorer to do a GET /todolist in the ToDoListController.

## Current Behavior

```
Error: ER_BAD_FIELD_ERROR with 'ToDo' field
```

I think something is looking to the ToDoList db table looking for the ToDo field. Obviously this doesn't make sense, right? I want the model to represent that, but I have a different table in my database for ToDo.

## Expected Behavior
Relation on model works

Related question: Is relational-database discovery available in LB4? Didn't really see a concrete answer in the docs. 

I don't have any experience with previous releases of LoopBack.

ZeroCoolJr:
This can be easily reproduced by running the lb4 example todo-list and changed it to use a different data source, in my case "postgres", I'm getting the "500 error: column "todos" does not exist". In-memory works fine.

dougal83:
> This can be easily reproduced by running the lb4 example todo-list and changed it to use a different data source, in my case "postgres", I'm getting the "500 error: column "todos" does not exist". In-memory works fine.

I managed to get postgres to work.  Used cli to create a postgres datasource which installed dependencies.  Have you created the db table with automigrate/autoupdate script?  Unlike tools like typeORM, I think you have to manually initiate db changes.

The following worked for me: (WARNING: example uses automigrate that will destroy existing table)
created ./src/bin/todo.automigrate.ts

import { TodoRepository } from '../repositories/todo.repository';
import { DataSource } from 'loopback-datasource-juggler';
import * as config from '../datasources/db.datasource.json';

const ds = new DataSource(config);
const repo = new TodoRepository(ds);

ds.attach(repo.modelClass);

ds.automigrate('Todo', err => {
  if (err) throw err;
  ds.disconnect();
});

executed using ts-node, it works for now.  I don't understand how to setup a table with relations currently.  Also this script is a bodge as implementation is incomplete.

ZeroCoolJr:
> > This can be easily reproduced by running the lb4 example todo-list and changed it to use a different data source, in my case "postgres", I'm getting the "500 error: column "todos" does not exist". In-memory works fine.
[...]

PR #1939, opened 2018-10-29
fix(repository): build relations based on their names

raymondfeng:
The current implementation uses property names as the key of
relations and it's not compatible with legacy juggler.

See https://github.com/strongloop/loopback-next/issues/1909

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -130,7 +130,7 @@ export class OrderRepository extends DefaultCrudRepository<
   ) {
     super(Order, db);
     this.customer = this._createBelongsToAccessorFor(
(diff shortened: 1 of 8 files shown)


raymondfeng:
I realized this PR is just a starting point to fully support `include`. There are a few issues:

1. LB4 adds a property for `hasMany` relation. For example: A customer class will have `orders` property but loopback-datasource-juggler does not like that for relational databases
2. We don't convert LB4 relations into juggler model definition - see https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repositories/legacy-juggler-bridge.ts#L131. Mapping LB4 relations into juggler is difficult as it needs to remap the source/target model too.

PR #1957, opened 2018-11-01
[WIP] feat: local API Explorer

hacksparrow:
Add local API Explorer

## Checklist

### DIFF
--- a/packages/core/src/component.ts
+++ b/packages/core/src/component.ts
@@ -15,15 +15,6 @@ export interface ProviderMap {
 }
 
 /**
- * Static assets settings
- */
-export type staticAssetsSettings = {
-  path: string;
-  rootDir: string;
-  options?: object;
-};
-
-/**
  * A component declares a set of artifacts so that they cane be contributed to
  * an application as a group
  */
@@ -42,10 +33,6 @@ export interface Component {
   servers?: {
     [name: string]: Constructor<Server>;
   };
-  /**
-   * An array of directory to serve static assets
-   */
-  staticAssets?: staticAssetsSettings[];
 
   /**
    * Other properties
--- a/packages/core/src/component.ts
+++ b/packages/core/src/component.ts
@@ -15,6 +15,15 @@ export interface ProviderMap {
 }
 
 /**
+ * Static assets settings
+ */
+export type staticAssetsSettings = {
+  path: string;
+  rootDir: string;
+  options?: object;
+};
+
+/**
  * A component declares a set of artifacts so that they cane be contributed to
  * an application as a group
  */
@@ -33,6 +42,10 @@ export interface Component {
   servers?: {
     [name: string]: Constructor<Server>;
   };
+  /**
+   * An array of directory to serve static assets
+   */
+  staticAssets?: staticAssetsSettings[];
 
   /**
    * Other properties
--- a/packages/rest/src/explorer.component.ts
+++ b/packages/rest/src/explorer.component.ts
@@ -3,13 +3,14 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Component} from '@loopback/core';
+import {Component} from './types';
+import * as path from 'path';
 
 export class ExplorerComponent implements Component {
   staticAssets = [
     {
       path: '/explorer1',
-      rootDir: './public',
+      rootDir: './public', //path.resolve(__dirname, '../../../explorer'),
       options: {},
     },
   ];
--- a/packages/rest/src/explorer.component.ts
+++ b/packages/rest/src/explorer.component.ts
@@ -3,10 +3,10 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Component} from './types';
+import {StaticComponent} from './types';
 import * as path from 'path';
 
-export class ExplorerComponent implements Component {
+export class ExplorerComponent implements StaticComponent {
   staticAssets = [
     {
       path: '/explorer1',
--- a/packages/rest/src/explorer.component.ts
+++ b/packages/rest/src/explorer.component.ts
@@ -1,0 +1,16 @@
+// Copyright IBM Corp. 2017. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Component} from '@loopback/core';
+
+export class ExplorerComponent implements Component {
+  staticAssets = [
+    {
+      path: '/explorer1',
+      rootDir: './public',
+      options: {},
+    },
+  ];
+}
--- a/packages/rest/src/rest.application.ts
+++ b/packages/rest/src/rest.application.ts
@@ -8,7 +8,6 @@ import {
   Application,
   ApplicationConfig,
   Server,
-  Component,
   mountComponent,
   BindingScope,
 } from '@loopback/core';
@@ -18,7 +17,6 @@ import {ServeStaticOptions} from 'serve-static';
 import {format} from 'util';
 import {RestBindings} from './keys';
 import {RestComponent} from './rest.component';
-import {ExplorerComponent} from './explorer.component';
 import {HttpRequestListener, HttpServerLike, RestServer} from './rest.server';
 import {
   ControllerClass,
@@ -26,6 +24,7 @@ import {
   RouteEntry,
 } from './router/routing-table';
 import {SequenceFunction, SequenceHandler} from './sequence';
+import {Component, StaticAssetsConfig} from './types';
 
 export const ERR_NO_MULTI_SERVER = format(
   'RestApplication does not support multiple servers!',
@@ -73,22 +72,26 @@ export class RestApplication extends Application implements HttpServerLike {
     return this.restServer.requestHandler;
   }
 
-  constructor(config: ApplicationConfig = {}) {
+  constructor(readonly config: ApplicationConfig = {}) {
     super(config);
     this.component(RestComponent);
-    if (config.enableExplorer) {
-      const name = ExplorerComponent.name;
-      const componentKey = `components.${name}`;
-      this.bind(componentKey)
-        .toClass(ExplorerComponent)
-        .inScope(BindingScope.SINGLETON)
-        .tag('component');
-      // Assuming components can be synchronously instantiated
-      const explorerInstance = this.getSync<Component>(componentKey);
-      mountComponent(this, explorerInstance);
-      console.log(explorerInstance.staticAssets);
-      const explorer = explorerInstance.staticAssets![0];
-      this.static(explorer.path, explorer.rootDir, explorer.options);
+  }
+
+  component(componentCtor: Constructor<Component>, name?: string) {
+    name = name || componentCtor.name;
+    const componentKey = `components.${name}`;
+    this.bind(componentKey)
+      .toClass(componentCtor)
+      .inScope(BindingScope.SINGLETON)
+      .tag('component');
+    // Assuming components can be synchronously instantiated
+    const instance = this.getSync<Component>(componentKey);
+    mountComponent(this, instance);
+    console.log(instance.staticAssets);
+    if (instance.staticAssets) {
+      instance.staticAssets.forEach((config: StaticAssetsConfig) => {
+        this.static(config.path, config.rootDir, config.options);
+      });
     }
   }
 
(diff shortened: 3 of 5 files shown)


packages/core/src/component.ts
```diff
@@ -14,6 +14,15 @@ export interface ProviderMap {
   [key: string]: Constructor<Provider<BoundValue>>;
 }
 
+/**
+ * A map of name/class pairs for binding static assets
+ */
+export type staticAssetsMap = {
```
bajtos: Uff, then name and comment seems very misleading to me, because there is no map involved AFAICT.

Also please use `interface` instead of `type`.

packages/rest/src/explorer.component.ts
```diff
@@ -0,0 +1,16 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Component} from '@loopback/core';
+
+export class ExplorerComponent implements Component {
+  staticAssets = [
+    {
+      path: '/explorer1',
+      rootDir: './public',
```
bajtos: How is this relative path going to be resolved into an absolute one? IMO, it should be the responsibility of the component to resolve the path. E.g. `rootDir: path.resolve(__dirname, '../../public')`

packages/rest/src/rest.application.ts
```diff
@@ -4,13 +4,21 @@
...
+  mountComponent,
+  BindingScope,
+} from '@loopback/core';
 import {OpenApiSpec, OperationObject} from '@loopback/openapi-v3-types';
 import {PathParams} from 'express-serve-static-core';
 import {ServeStaticOptions} from 'serve-static';
 import {format} from 'util';
 import {RestBindings} from './keys';
 import {RestComponent} from './rest.component';
+import {ExplorerComponent} from './explorer.component';
```
bajtos: Don't bundle the explorer component inside `@loopback/rest`. The coupling between the API explorer (e.g. a specific swagger-ui version used) and the runtime (`@loopback/rest`) is the thing we want to avoid by moving explorer into a new component.

The API Explorer should be added to applications as part of our project scaffolding provided by `lb4 app`. Nice to have: `lb4 app` can ask the user whether they want to add API Explorer to their project or not.

packages/rest/src/rest.application.ts
```diff
@@ -68,6 +76,20 @@ export class RestApplication extends Application implements HttpServerLike {
...
+      this.bind(componentKey)
+        .toClass(ExplorerComponent)
+        .inScope(BindingScope.SINGLETON)
+        .tag('component');
+      // Assuming components can be synchronously instantiated
+      const explorerInstance = this.getSync<Component>(componentKey);
+      mountComponent(this, explorerInstance);
+      console.log(explorerInstance.staticAssets);
+      const explorer = explorerInstance.staticAssets![0];
+      this.static(explorer.path, explorer.rootDir, explorer.options);
```
bajtos: This block needs to be split into two parts:

`RestApplication` class should override `component` method provided by `Application`, iterate over `component.staticAssets` and call `this.static` to register them.

The scaffolded project-specific application class should register the component in the constructor, e.g.

```ts
class TodoApplication extends RepositoryMixin(BootMixin(RestApplication)) {
  constructor () {
    // ...
    this.component(ExplorerComponent);
  }
}
```

packages/core/src/component.ts
```diff
@@ -33,6 +42,10 @@ export interface Component {
   servers?: {
     [name: string]: Constructor<Server>;
   };
+  /**
+   * An array of directory to serve static assets
+   */
+  staticAssets?: staticAssetsSettings[];
```
bajtos: I think these REST-specific extensions should be defined in a new interface inside `@loopback/rest`, because applications using different transports (e.g GraphQL or gRPC) cannot support static assets.

packages/core/src/component.ts
```diff
@@ -14,6 +14,15 @@ export interface ProviderMap {
   [key: string]: Constructor<Provider<BoundValue>>;
 }
 
+/**
+ * Static assets settings
+ */
+export type staticAssetsSettings = {
```
bajtos: `export interface StaticAssetsSettings` (use interface, not type; interface names start with a capital letter)

I would also pick a slightly different name, e.g. `staticAssetsDefinition` or `staticAssetsConfig`.

raymondfeng:
@hacksparrow Are you proposing it as a replacement to https://github.com/strongloop/loopback-next/pull/1664?

packages/rest/src/types.ts
```diff
@@ -4,11 +4,28 @@
...
+/**
+ * Static assets settings
+ */
+export interface StaticAssetsConfig {
+  path: string;
+  rootDir: string;
+  options?: object;
+}
+
+export interface Component extends CoreComponent {
```
raymondfeng: It's probably better to name it as `ComponentWithStaticAssets`.

packages/rest/src/rest.application.ts
```diff
@@ -65,11 +72,29 @@ export class RestApplication extends Application implements HttpServerLike {
     return this.restServer.requestHandler;
   }
 
-  constructor(config: ApplicationConfig = {}) {
+  constructor(readonly config: ApplicationConfig = {}) {
     super(config);
     this.component(RestComponent);
   }
 
+  component(componentCtor: Constructor<Component>, name?: string) {
```
raymondfeng: What if an application does not extend from `RestApplication` but mount the `RestComponent`?

packages/rest/src/rest.application.ts
```diff
@@ -65,11 +72,29 @@ export class RestApplication extends Application implements HttpServerLike {
...
+    name = name || componentCtor.name;
+    const componentKey = `components.${name}`;
+    this.bind(componentKey)
+      .toClass(componentCtor)
+      .inScope(BindingScope.SINGLETON)
+      .tag('component');
+    // Assuming components can be synchronously instantiated
+    const instance = this.getSync<Component>(componentKey);
+    mountComponent(this, instance);
+    console.log(instance.staticAssets);
```
raymondfeng: Remove `console.log`

packages/core/src/component.ts
raymondfeng: Should we generalize it to allow components to contribute `middleware` based routes?

packages/rest/src/types.ts
bajtos: We my want to add additional exports besides static assets later in the future. I think `RestComponent` would be a better name?

I also think the interface should not be inheriting from `CoreComponent` to allow composition of multiple transport-specific interfaces, e.g. `Component & RestComponent & GrapQLComponent`.

packages/rest/src/rest.application.ts
bajtos: If the app is extending from `Application`, then there is no REST server configured by default and the rest-specific exports are silently ignored.

raymondfeng: What I'm trying to say is that `RestComponent` should be responsible for mounting such routes. In this case, other components contribute routes for the `RestComponent`.

packages/rest/src/types.ts
raymondfeng: Don't we have a class named as `RestComponent`?

hacksparrow: > Don't we have a class named as RestComponent
@raymondfeng Yes, we have that already. For now I have renamed it to `StaticComponent`.

> I also think the interface should not be inheriting from CoreComponent ...

@bajtos in that case we'll have to do `componentCtor: Constructor<Component | StaticComponent>`, etc. Is that OK?

I think this should be an independent interface from `RestComponent`, `StaticComponent` does not need any of the properties defined in `RestComponent`. Clubbing them together is not efficient.

packages/rest/src/rest.application.ts
hacksparrow: @raymondfeng can you elaborate on the above statement, please?

hacksparrow:
> Are you proposing it as a replacement to #1664?

@raymondfeng yes, a different approach.

packages/core/src/component.ts
hacksparrow: Sounds like a separate feature to me.

raymondfeng:
> @raymondfeng yes, a different approach.

Does it assume that API explorer UI is purely based on static assets without any server-side templating/rendering?

packages/rest/src/types.ts
```diff
@@ -9,6 +9,22 @@ import {Request, Response} from 'express';
...
+/**
+ * Static assets settings
+ */
+export interface StaticAssetsConfig {
+  path: string;
+  rootDir: string;
+  options?: object;
+}
+
+export interface StaticComponent {
```
raymondfeng: The name `StaticComponent` does not make sense to me. It's really a component that contributes static assets. I propose to consider https://github.com/strongloop/loopback-next/pull/1924 so that we can give more flexibility to component providers. For example, the ExplorerComponent can expose a binding for its static assets so that the RestComponent can discover that and mount it.

packages/rest/src/rest.application.ts
raymondfeng: A developer can do the the following:

```ts
const app = new Application();
app.component(RestComponent);
app.component(ExplorerComponent);
```
With the current implementation, no `/explorer` route will be mounted.

hacksparrow:
> Does it assume that API explorer UI is purely based on static assets without any server-side templating/rendering?

Yes.

packages/rest/src/types.ts
hacksparrow: Agreed `StaticComponent` isn't the best name. We can decide on the appropriate name later, for now I chose something to get moving.

packages/rest/src/rest.application.ts
hacksparrow: Thanks, that explains it.

@bajtos that would require us to leave`RestApplication.component` alone and let `Application.component` take care of mounting components. However, now we'll have to have a list of supported components (`Component | RestComponent` etc.) in the parameters, IF we don't want `RestComponent` to extend `Component`.

bajtos:
```ts
const app = new Application();
app.component(RestComponent);
app.component(ExplorerComponent);
```

Good catch! I did not realize we have `RestComponent`.

I need to think this through, will post a follow-up comment later.

bajtos:
Based on the thread https://github.com/strongloop/loopback-next/pull/1957#discussion_r230118996, it looks like I underestimated the complexity of allowing components to contribute static assets in an indirect way (by exporting an `staticAssets` array). I am proposing to abandon that approach for now and focus on how to implement self-hosted API explorer using the tools that LB provides right now.

In an older chat with @hacksparrow, we were discussing the following approach:

```ts
class ExplorerComponent {
  constructor(
    @inject(CoreBindings.APPLICATION_INSTANCE) app: Application
  ) {
   app.static(/*...*/);
  }
}
```

I opened a new issue to discuss the best way how to allow extensions to contribute static assets: https://github.com/strongloop/loopback-next/issues/1967

packages/rest/src/rest.application.ts
bajtos: As I commented at the bottom, I think this is a wider problem that we should address outside of this pull request, I opened a new issue for that - see https://github.com/strongloop/loopback-next/issues/1967

packages/rest/src/types.ts
bajtos: > Don't we have a class named as RestComponent?

We actually do! It's a component adding REST servers to an app. 

The intention here was to describe a component that's contributing REST-specific artifacts like static assets.

What a confusion 🙈

packages/core/src/component.ts
bajtos: +1 to leave middleware-based routes out of scope.

Personally, I would like to avoid middleware-based routes for as long as we can. LB4 already supports handler-based routes that play nice with OpenAPI, I would prefer extensions to use that approach.

hacksparrow:
Although we did not achieve what we wanted to with this approach at this moment, at least we discovered an important missing feature - contributing static assets from components.

I will be opening a new PR for local API Explorer with the approach @bajtos described above.

raymondfeng:
I think that would be just simplified version of https://github.com/strongloop/loopback-next/pull/1664 based on the assumption that ExplorerComponent only requires static assets without server-side rendering/templating.

hacksparrow:
> I think that would be just simplified version of #1664 based on the assumption that ExplorerComponent only requires static assets without server-side rendering/templating

Yes.

hacksparrow:
Closing this in favor of https://github.com/strongloop/loopback-next/pull/1976.

PR #1964, opened 2018-11-02
Fix trie routing

raymondfeng:
1. Handle overlapping paths with different vars

For example:

- GET /users/{id}/products
- GET /users/{id}
- GET /users/{userId}/orders

The issue was discovered with https://github.com/strongloop/loopback4-example-shopping/pull/23

2. Improve route sorting to group by path and verb. 

For example:
```
      {verb: 'get', path: '/orders/count'},
      {verb: 'get', path: '/orders/{id}/exists'},
      {verb: 'put', path: '/orders/{id}'},
      {verb: 'patch', path: '/orders/{id}'},
      {verb: 'get', path: '/orders/{id}'},
      {verb: 'delete', path: '/orders/{id}'},
      {verb: 'post', path: '/orders'},
      {verb: 'patch', path: '/orders'},
      {verb: 'get', path: '/orders'},
      {verb: 'delete', path: '/orders'},
      {verb: 'get', path: '/users/{userId}/orders'},
      {verb: 'get', path: '/users/{id}/products'},
      {verb: 'get', path: '/users/{id}'},
```

We currently sort by verb/number of parts/path.

## Checklist

### DIFF
--- a/benchmark/src/rest-routing/README.md
+++ b/benchmark/src/rest-routing/README.md
@@ -11,29 +11,39 @@ https://loopback.io/doc/en/lb4/Routing-requests.html for more information.
 
 ```sh
 npm run -s benchmark:routing // default to 1000 routes
-npm run -s benchmark:routing -- <number-of-routes>
+npm run -s benchmark:routing -- <number-of-routes> <number-of-routes>
+```
+
+For example:
+
+```
+npm run -s benchmark:routing -- 10 40 160 640 2560
 ```
 
 ## Base lines
 
 ```
 name                 duration    count    found   missed
-TrieRouter          0,1453883       10        8        2
-RegExpRouter        0,1220030       10        8        2
+TrieRouter          0,1038244       10        8        2
+RegExpRouter        0,1002559       10        8        2
+
 
 name                 duration    count    found   missed
-TrieRouter          0,2109957       40       35        5
-RegExpRouter        0,8762936       40       35        5
+TrieRouter           0,661415       40       35        5
+RegExpRouter        0,5797990       40       35        5
+
 
 name                 duration    count    found   missed
-TrieRouter          0,4895252      160      140       20
-RegExpRouter       0,52156699      160      140       20
+TrieRouter          0,2851976      160      140       20
+RegExpRouter       0,27040556      160      140       20
+
 
 name                 duration    count    found   missed
-TrieRouter         0,16065852      640      560       80
-RegExpRouter      0,304921026      640      560       80
+TrieRouter          0,8914005      640      560       80
+RegExpRouter      0,330599418      640      560       80
+
 
 name                 duration    count    found   missed
-TrieRouter         0,60165877     2560     2240      320
-RegExpRouter      4,592089555     2560     2240      320
+TrieRouter         0,45302321     2560     2240      320
+RegExpRouter      5,471787942     2560     2240      320
 ```
--- a/packages/rest/src/router/route-sort.ts
+++ b/packages/rest/src/router/route-sort.ts
@@ -28,41 +28,30 @@ export function compareRoute(
   route1: Pick<RouteEntry, 'verb' | 'path'>,
   route2: Pick<RouteEntry, 'verb' | 'path'>,
 ): number {
-  // First check verb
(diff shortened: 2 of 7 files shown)


bajtos:
> Improve route sorting to group by path and verb.

Is this new sort changing observable behavior or is it a performance improvement only?

Either way, please re-run both bechmarks and update the results in READMEs as part of this patch.

bajtos:
I think the change in sorting will also modify the order in which OpenAPI endpoints are declared in the spec. Should we communicate this change to LB users?

raymondfeng:
> Is this new sort changing observable behavior or is it a performance improvement only?

It's mostly for OpenAPI spec to better group routes. There is probably some minor performance improvement too. Trie-based router does not rely the order for matching.

Please also note that OpenAPI uses `tags` to group routes 1st.

packages/rest/test/unit/router/trie-router.unit.ts
```diff
@@ -36,30 +46,76 @@ describe('trie router', () => {
...
+    const req = givenRequest({method: 'get', url: '/users/123/products'});
+    const route = router.find(req);
+    expect(getRouteInfo(route)).to.containEql({
+      verb: 'get',
+      path: '/users/{id}/products',
+      params: {id: '123'},
+    });
+  });
+
+  it('finds route for GET /users/{userId}/orders', () => {
```
bajtos: I find these two tests ("finds route for GET /users/{id}/products" and "finds route for GET /users/{userId}/orders") very difficult to understand. The paths looks pretty much the same, why would one test fail when the other is passing?

Please find a way how to capture the problem you are solving in the structure of the test suite. For example, the test can make it more clear that parameterized paths using different path parameter names are correctly "merged" together and point endpoints work as expected.

It would be also great to move the setup of routes closer to the tests. Right now it's not clear at all which test is relying on which routes defined by `givenRoutes`.

A good test  uses only the minimal setup is strictly required by the tested scenario. By creating a new test setup razor-focused on needs of each test case, the tests become easier to understand, since it’s immediately clear what parts of the setup are relevant to the tests. If the tests share a large setup with many routes, it is difficult to tell which routes are actually relevant to the tested scenario.

raymondfeng:
@bajtos PTAL

benchmark/src/rest-routing/README.md
```diff
@@ -11,29 +11,39 @@ https://loopback.io/doc/en/lb4/Routing-requests.html for more information.
...
-RegExpRouter       0,52156699      160      140       20
+TrieRouter          0,2851976      160      140       20
+RegExpRouter       0,27040556      160      140       20
+
 
 name                 duration    count    found   missed
-TrieRouter         0,16065852      640      560       80
-RegExpRouter      0,304921026      640      560       80
+TrieRouter          0,8914005      640      560       80
+RegExpRouter      0,330599418      640      560       80
```
bajtos: Weird. If I am reading these numbers correctly, then TrieRouter is about 2.5 times slower than RegExpRouter in this scenario. Is that expected?

packages/rest/src/router/route-sort.ts
```diff
@@ -28,41 +28,30 @@ export function compareRoute(
...
+    if (token1 === undefined) return 1;
+    if (token2 === undefined) return -1;
+    if (token1 < token2) return -1;
+    if (token1 > token2) return 1;
   }
+  // Then check verb
+  const verb1 = HTTP_VERBS[route1.verb.toLowerCase()] || HTTP_VERBS.get;
+  const verb2 = HTTP_VERBS[route2.verb.toLowerCase()] || HTTP_VERBS.get;
+  if (verb1 !== verb2) return verb1 - verb2;
   return 0;
```
bajtos: FWIW, this line is not executed by our test suite according to https://coveralls.io/builds/19876890/source?filename=packages/rest/src/router/route-sort.ts

Please consider adding a test to demonstrate what happens when two routes with the same path are configured. (It's probably better to open a new PR for that, so that this patch can be landed ASAP.)

benchmark/src/rest-routing/README.md
raymondfeng: No, TrieRouter takes `8,914,005ns (8.9ms)` while RegExpRouter takes `330,599,418ns (330ms)` to process 640 requests.

PR #1965, opened 2018-11-02
refactor(rest): merge RoutingTable into HttpHandler

bajtos:
Initially, the RoutingTable class was created to leave router complexity out of the HttpHandler class. This has its own problems too: most of HttpHandler's public API was just delegating the implementation to RoutingTable, which was making API changes cumbersome.

After recent changes that extracted route registration and lookup into standalone RestRouter classes, there is no longer need for this split and the two classes can be joined together.

## Checklist


bajtos:
Initially, I made this changes while working on https://github.com/strongloop/loopback-next/pull/1955. On the second thought, I am not entirely sure if it's a good idea to merge RoutingTable and HttpHandler into a single class: before my changes, the RoutingTable was independent from Context and could be used to build an alternate REST framework that does not use dependency injection, or maybe extracted into a router module that could be used by people building express apps, etc. With my changes in place, routing is coupled with Context and Dependency Injection, because `HttpHandler` requires the top-level context as the first constructor argument.

Let's discuss - should we proceed with this pull request? Should I look into other code designs that would allow us to get rid dummy methods in HttpHandler that are just forwarding the implementation to RoutingTable?

hacksparrow:
> With my changes in place, routing is coupled with Context and Dependency Injection, because HttpHandler requires the top-level context as the first constructor argument.

`HttpHandler` was already dependent on the top-level context, right?

benchmark/src/rest-routing/routing-table.ts
```diff
@@ -75,7 +76,7 @@ function givenRouter(router: RestRouter, spec: OpenApiSpec, count: number) {
     log('Creating %s, %d', name, count);
     let start = process.hrtime();
 
-    const table = new RoutingTable(router);
+    const table = new HttpHandler(new Context(), router);
```
jannyHou: nitpick: maybe rename `table` to `handler` as well? since it's not a routing table anymore.

raymondfeng:
> Although, HttpHandler was mostly just forwarding the implementation to the RoutingTable, it acts as a place to fine tune some aspects (if required), apart from enabling RoutingTable to be a specialized entity.

> RoutingTable also helps create a good mental image of a map of paths and routes that are responsible for setting up the routing in the framework; clubbing everything under HttpHandler, not so much. Not that HttpHandler does not give the idea of routing, it is just that it a very open-ended idea; RoutingTable is more specific.

> I am not in favor of this change, because if we need to add more implementation code (you never know what may come up) in the HttpHandler as proposed, it will start to get cluttered and we may end up re-creating RoutingTable again. Also, general "intuition" 😀

I agree with @hacksparrow. -1 to merge two classes.

jannyHou:
I am ok to merge the `routing-table` class into `http-handler`:

> before my changes, the RoutingTable was independent from Context and could be used to build an alternate REST framework that does not use dependency injection

My understanding is people can provide a dummy `new Context()` if they don't need DI. And usually the limitation for users would be "requiring context but the class doesn't support it", not the other way around. 

Another reason is the current `http-handler` mostly wraps functions in `routing-table` without additional code, and I couldn't think of example functionalities that have to be decoupled into two classes instead of adding them directly in the http-handler class 🤔

b-admike:
> Initially, I made this changes while working on #1955. On the second thought, I am not entirely sure if it's a good idea to merge RoutingTable and HttpHandler into a single class: before my changes, the RoutingTable was independent from Context and could be used to build an alternate REST framework that does not use dependency injection, or maybe extracted into a router module that could be used by people building express apps, etc. With my changes in place, routing is coupled with Context and Dependency Injection, because `HttpHandler` requires the top-level context as the first constructor argument.
> 
> Let's discuss - should we proceed with this pull request? Should I look into other code designs that would allow us to get rid dummy methods in HttpHandler that are just forwarding the implementation to RoutingTable?
> 
> /cc @strongloop/loopback-maintainers

I am not too familiar with this codebase, but from what I've read in this thread, I think it's worth it to explore one alternative and see why we should proceed with this PR. With that said, IMO, decoupling Context/DI from RoutingTable is worth it for some dummy methods in HttpHandler which call RoutingTable's methods.

PR #1976, opened 2018-11-05
[WIP]feat: add local API Explorer

hacksparrow:
Add local API Explorer

## Checklist

### DIFF
--- a/packages/rest/src/explorer.component.ts
+++ b/packages/rest/src/explorer.component.ts
@@ -1,0 +1,50 @@
+// Copyright IBM Corp. 2017. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Component, CoreBindings} from '@loopback/core';
+import {inject} from '@loopback/context';
+import {RestApplication, RestBindings, Request, Response} from './';
+import {get} from '@loopback/openapi-v3';
+const swaggerUI = require('swagger-ui-dist');
+const path = require('path');
+const readFileSync = require('fs').readFileSync;
+const writeFileSync = require('fs').writeFileSync;
+
+export class ExplorerComponent implements Component {
+  controllers = [ExplorerController];
+
+  constructor(@inject(CoreBindings.APPLICATION_INSTANCE) app: RestApplication) {
+    if (app.config.enableExplorer) {
+      //console.log('COMPONENT INIT');
+      const openApiSpecUrl = app.config.openApiSpecUrl || '/openapi.json';
+      const explorerPath = app.config.explorerPath || '/explorer';
+      const explorerLocalPath = swaggerUI.getAbsoluteFSPath();
+      // const explorerIndexFilePath = path.join(explorerLocalPath, 'index.html');
+      // const templateFilePath = path.join(__dirname, 'index.html.tpl');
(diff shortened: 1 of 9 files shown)


packages/rest/src/explorer.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      const explorerPath = app.config.explorerPath || '/explorer';
+      const explorerLocalPath = swaggerUI.getAbsoluteFSPath();
+      // const explorerIndexFilePath = path.join(explorerLocalPath, 'index.html');
+      // const templateFilePath = path.join(__dirname, 'index.html.tpl');
+      // const templateContent = readFileSync(templateFilePath, 'utf8').replace(
+      //   '<%- openApiSpecUrl %>',
+      //   openApiSpecUrl,
+      // );
+      // writeFileSync(explorerIndexFilePath, templateContent);
+      app.static(explorerPath, explorerLocalPath);
```
hacksparrow: @bajtos `app.static` seems to be overriding the `ExplorerController `. Any idea what's going on? Or did I misunderstand something?

packages/rest/test/integration/rest.application.integration.ts
```diff
@@ -71,6 +80,19 @@ describe('RestApplication (integration)', () => {
       .expect(content);
   });
 
+  it('gives precedence to API controllers over static assets', async () => {
```
hacksparrow: @bajtos here is the test which catches the problem. Calling `restApp.static('/html', root);` after  `restApp.controller(DummyController);` passes the test. That means, something's wrong in the order of registration /look up.

raymondfeng:
@hacksparrow Please rebase this PR off master.

hacksparrow:
_Notes on the current state of the PR_

- Apps can be configured to enable the local API Explorer by passing these two config options: `enableExplorer` and `explorerPath`. Eg:

```ts
  options.enableExplorer = true;
  options.explorerPath = '/explorer1';
```
- On making a GET request to the root of the API Explorer as defined in `explorerPath`, we intent to serve a customized `index.html` file. However due to a current limitation, the file is not being served as `text/html`. Raymond is working on this issue - https://github.com/strongloop/loopback-next/tree/honor-response-content-type.

- What do we do with the currently baked-in router `/explorer`?
  - Do we remove it altogether?
  - What do we do if user set `explorerPath` to `/explorer`?

bajtos:
> Apps can be configured to enable the local API Explorer by passing these two config options: enableExplorer and explorerPath.

I would personally prefer a slightly different configuration:
- To enable local API explorer, apps have to add ExplorerComponent.
- To disable local API explorer, apps should skip registration of ExplorerComponent.
That way we can keep ExplorerComponent simpler. However, this solution may not be feasible depending on what we do with the built-in redirect to swagger-ui.

> What do we do with the currently baked-in router for /explorer?
> 
> - Do we remove it altogether?
> - What do we do if user set explorerPath to /explorer?

I feel that removing this feature would be a breaking change that requires semver-major. Are we ok with publishing `@loopback/rest@5.0.0` just a month after v4 was released?

IMO, a better solution is to introduce a configuration flag allowing applications to disable this built-in redirect. When the Explorer component is mounted, it can set this flag to `true` to ensure the built-in redirect is disabled.

bajtos:
I opened a new PR inspired by this work, see https://github.com/strongloop/loopback-next/pull/2014

Issue #433, opened 2017-07-13
Dynamic binding/rebinding of controllers after app start


bajtos:
Support hot-reloading of controllers after the app started. See [`Application.prototype._setupHandlerIfNeeded()`](https://github.com/strongloop/loopback-next/blob/19604dbc1dcaa8eed7c79c0e46e5aa3f5fb3b8b9/packages/core/src/application.ts#L85-L89).

Example test case:

```ts
const app = new Application();
await app.start();
app.controller(MyController);
// MyController is available via REST API now
```

## Acceptance criteria

When the application binds a controller after the app has started, the application picks up the endpoints contributed by the new controller.
- [ ] REST server is updated with new endpoints, so that clients can access the new API.
- [ ] OpenAPI specification of the application is updated to describe the new endpoints.

Nice to have:

- [ ] When a controller is removed from the application (e.g. via `app.unbind()` API), the endpoints are removed from the REST server and the OpenAPI spec.

lindoelio:
Hey guys! I'll working on this issue. As soon as possible I will send my suggestion on a PR. Let me know if you have any advise for me about it :-)

dhmlau:
@lindoelio, are you still interested in submitting a PR?

dhmlau:
@bajtos , what is the use case for hot-reloading?  I'm thinking users would add controller during development time, so it's not so critical to have the hot reloading feature. 

Discussed with @virkt25, this is a nice-to-have feature, but we can leave it post-GA given there are higher priority items for GA.

bajtos:
I am fine with removing this item from 4.0 GA scope.

> what is the use case for hot-reloading?

For example, IBM AppConnect uses LoopBack to define models & REST APIs on the fly. Their LoopBack server provides REST APIs to define (create) and remove connections to 3rd-party databases/services. When a new 3rd party database/service is connected, AppConnect need to create a set of models & controllers allowing its clients to access the new database/service through the LoopBack runtime.

oleg269:
Hi,

I'll working on this issue. 
 
I tried load controller after the app started and used ``app.controller(myController)``.

Controller was added to app registry, but routes of controller were not created.
I found the solution to add routes of controller:

Change in ``loopback-next/packages/rest/src/rest.server.ts``

    protected get httpHandler(): HttpHandler {

to 

    public get httpHandler(): HttpHandler {

and add:

    app.controller(TestController);
    var restServer: RestServer = await app.get<RestServer>('servers.RestServer');
[...]

PR #1977, opened 2018-11-05
fix: fix static assets router blocking controller registration

hacksparrow:
This PR is to fix a regression, which prevented controllers and routes from being registered if `app.static()` was called before them.

`app.static()` was calling `this.httpHandler.registerStaticAssets(path, rootDir, options);`, which caused `_setupHandlerIfNeeded()` to be called and `this._httpHandler` being set. Once `this._httpHandler` is set, none of the bound controllers or routes were registered.

The fix is to collect the details of `app.static()` in an array `this.staticAssets`, and process them as part of `_setupHandlerIfNeeded()`.

## Checklist


packages/rest/test/integration/rest.application.integration.ts
```diff
@@ -92,6 +127,19 @@ describe('RestApplication (integration)', () => {
       .expect('Hello');
   });
 
+  it('gives precedence to API controllers over static assets', async () => {
```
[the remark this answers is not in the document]
hacksparrow: > ... there is no need to duplicate RestServer tests in RestApplication integration test suite ...

If the implementation of `app.static` is broken (for whatever reason) and we don't have tests to determine `app.static` actually work, the problem will be found out only after trying to call `app.static` in a project.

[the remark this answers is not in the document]
bajtos: True. On the other hand, `app.static` is implemented as a simple wrapper delegating all work to `this.restServer.static`. I agree it's important to have basic integration tests in place, but I don't think it's necessary to duplicate tests for all edge cases. Having too many tests is a maintenance burden.

packages/rest/test/integration/rest.application.integration.ts
```diff
@@ -92,6 +127,19 @@ describe('RestApplication (integration)', () => {
       .expect('Hello');
   });
 
+  it('registers controllers defined later than static assets', async () => {
```
bajtos: It's important to run this test against `RestServer` because that's the place the bug was caused.

If you feel it's important to duplicate this test for RestApplication too, then I can live with that, but please test this scenario for bare RestServer.

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -114,6 +114,49 @@ describe('RestServer (integration)', () => {
...
+
+    server.static('/html', dirA);
+    server.static('/html', root);
+    const content = fs
+      .readFileSync(path.join(dirA, 'index.html'))
+      .toString('utf-8');
+    await createClientForHandler(server.requestHandler)
+      .get('/html/index.html')
+      .expect('Content-Type', /text\/html/)
+      .expect(200, content);
```
bajtos: It would be great to also verify that different file names are correctly merged. For example, when the first static assets directory contains `/index.html` (only) and the second directory contains `/robots.txt`, then requests for `/robots.txt` should be served from the second directory.

In the context of this test, we can change "dir-a" fixture to contain `robots.txt` file in addition to `index.html`, and then make another request in this test to verify that `robots.txt` can be loaded from the second directory.

packages/rest/test/integration/rest.application.integration.ts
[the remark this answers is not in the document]
hacksparrow: 👍 removed them.

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -114,6 +114,49 @@ describe('RestServer (integration)', () => {
...
+    await createClientForHandler(server.requestHandler)
+      .get('/html-1/index.html')
+      .expect('Content-Type', /text\/html/)
+      .expect(200, content);
+  });
+
+  it('allows static assets to be mounted on the same path multiple times', async () => {
+    const root = FIXTURES;
+    const dirA = path.join(root, 'dir-a');
```
bajtos: Let's find a more descriptive directory name than `dir-a` please.

I think it's time to reorganize the `fixtures` directory, I am proposing the following structure:

```
fixtures
 - cert.pem
 - key.pem
 - pfx.pfx
 - assets
    - index.html # moved from fixtures/index.html
 - other-assets
    - index.html
```

In the long term, I think it would be best to leverage `TestSandbox` to create temporary asset directories for the tests, but I guess that's out of scope of this pull request.

packages/rest/test/integration/rest.application.integration.ts
hacksparrow: Moved to `RestServer ` integration test.

packages/rest/src/rest.server.ts
```diff
@@ -614,7 +621,15 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param options Options for serve-static
    */
   static(path: PathParams, rootDir: string, options?: ServeStaticOptions) {
-    this.httpHandler.registerStaticAssets(path, rootDir, options);
+    if (this._httpHandler) {
+      this.httpHandler.registerStaticAssets(path, rootDir, options);
```
bajtos: I find this a bit weird - you are checking for the existence of `this._httpHandler` and then calling `this.httpHandler.registerStaticAssets`. Shouldn't you use the same property in both places?

packages/rest/src/rest.server.ts
```diff
@@ -310,6 +310,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
         this._setupOperation(verb, path, routeSpec);
       }
     }
+
+    this.staticAssets.forEach(assetEntry => {
+      const {path, rootDir, options} = assetEntry;
+      this.httpHandler.registerStaticAssets(path, rootDir, options);
```
bajtos: I think you should call `this._httpHandler.registerStaticAssets` here, instead of `this.httpHandler....`?

packages/rest/src/rest.server.ts
```diff
@@ -614,7 +621,15 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param options Options for serve-static
    */
   static(path: PathParams, rootDir: string, options?: ServeStaticOptions) {
-    this.httpHandler.registerStaticAssets(path, rootDir, options);
+    if (this._httpHandler) {
+      this.httpHandler.registerStaticAssets(path, rootDir, options);
+    } else {
+      this.staticAssets.push({
```
bajtos: I think you should always create a new entry in `this.staticAssets`, so that all static assets are preserved whenever the HTTP handler is rebuilt.

See #433 and 
https://github.com/strongloop/loopback-next/blob/1bcdb5b69221501b3952bd70c0c78409b600d1ab/packages/rest/src/rest.server.ts#L262-L267

I think this is the simplest version of this method:

```ts
this.staticAssets.push(/*...*/);
delete this._httpHandler;
```

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -114,8 +115,61 @@ describe('RestServer (integration)', () => {
...
+      .get('/html-0/index.html')
+      .expect('Content-Type', /text\/html/)
+      .expect(200, content);
+    await createClientForHandler(server.requestHandler)
+      .get('/html-1/index.html')
+      .expect('Content-Type', /text\/html/)
+      .expect(200, content);
+  });
+
+  it('allows static assets to be mounted on the same path multiple times', async () => {
```
bajtos: Let's improve this test name:

`merges different static asset directories when mounted on the same path`

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -114,8 +115,61 @@ describe('RestServer (integration)', () => {
...
+        port: 0,
+      },
+    });
+
+    server.static('/html', root);
+    server.static('/html', otherAssets);
+
+    let content = fs
+      .readFileSync(path.join(root, 'index.html'))
+      .toString('utf-8');
```
bajtos: This can be simplified, see https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_fs_readfilesync_path_options

```ts
let content = fs.readFileSync(path.join(root, 'index.html'), {encoding: 'utf-8'});
```

It would be even better to extract a shared helper function to avoid repetition.

```ts
let content = await readFileFromDirectory(root, 'index.html');

// impl
const readFileAsync = util.promisify(fs.readFile);
function readFileFromDirectory(dirname: string, filename: string): Promise<string> {
  return readFileAsync(path.join(dirname, filename), {encoding: 'utf-8'});
}
```

raymondfeng:
@hacksparrow Is this a regression? Please add a bit more context in the PR.

hacksparrow:
@raymondfeng updated:

> This PR is to fix a regression, which prevented controllers and routes from being registered if app.static() was called before them.

> app.static() was calling this.httpHandler.registerStaticAssets(path, rootDir, options);, which caused _setupHandlerIfNeeded() to be called and this._httpHandler being set. Once this._httpHandler is set, none of the bound controllers or routes were registered.

> The fix is to collect the details of app.static() in an array this.staticAssets, and process them as part of _setupHandlerIfNeeded().

packages/rest/src/rest.server.ts
```diff
@@ -310,6 +310,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
         this._setupOperation(verb, path, routeSpec);
       }
     }
+
+    this.staticAssets.forEach(assetEntry => {
+      const {path, rootDir, options} = assetEntry;
+      this._httpHandler.registerStaticAssets(path, rootDir, options);
```
raymondfeng: Why don't we create an Express Router as the container so that `registerStaticAssets` can just add a new entry to the router without rebuilding the handler?

hacksparrow: I am responding which some assumptions, please clarify if I misunderstood something.

We are ultimately using an Express Router - https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/router/static-assets-route.ts#L35.

We have to register the static assets router with the Routing Table so that we have the ability to check the matching route for a request in the Routing Table, and return the static assets router, if none was matched.

If I misunderstood your statement, can you clarify what you mean by "container" and "rebuilding the handler".

raymondfeng: I'm saying we always create an instance of `StaticAssetsRoute` as the place holder and use it to register static assets. This way we don't have to do the following:
```
this.staticAssets.push(/*...*/);
delete this._httpHandler; // I assume it's used to remove the cached _httpHandler to pick up new static assets
```

hacksparrow: I am still not getting it 😬.

One of the requirements of `app.static()` is that it should be able to add new routes even after the app has started. Will it work with what you are suggesting?

packages/rest/src/rest.server.ts
```diff
@@ -844,3 +848,12 @@ export interface RestServerOptions {
  * @interface RestServerConfig
  */
 export type RestServerConfig = RestServerOptions & HttpServerOptions;
+
+/**
+ * Options for handling static assets
+ */
+export interface StaticAssetsEntry {
```
bajtos: AFAICT, this interface is no longer used. Please remove.

hacksparrow: Removed.

PR #1984, opened 2018-11-06
docs: how to serve static files

hacksparrow:
Document how to serve static files.

## Checklist

### DIFF
--- a/docs/site/Serving-static-files.md
+++ b/docs/site/Serving-static-files.md
@@ -1,0 +1,64 @@
+---
+lang: en
+title: 'Serving static files in LoopBack 4'
+keywords: LoopBack
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Serving-static-files.html
+---
+
+One of the basic requirements of a web app is the ability to serve static files.
+Serving static files from a LoopBack is very simple - just call the
+`app.static(path, rootDir[, options])` method. The variables in the API are
+explained below.
+
+- `app`: An instance of a LoopBack application.
+- `path`: The path where the static assets are to be served from. Refer to
+  https://expressjs.com/en/4x/api.html#path-examples for possible values.
+- `rootDir`: The directory where the static assets are located on the file
+  system.
+- `options`: An otional object for configuring the underlying
+  [express.static](https://expressjs.com/en/4x/api.html#express.static).
+
+Here is an example of configuring an app to serve the static assets from a
+directory named `public` at `/`.
+
+```ts
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+const path = require('path');
+
+export async function main(options?: ApplicationConfig) {
+  const app = new TodoListApplication(options);
+  app.static('/', path.join(__dirname, 'public'));
+  await app.boot();
+  await app.start();
+
+  const url = app.restServer.url;
+  console.log(`Server is running at ${url}`);
+  return app;
+}
+```
+
+You can call `app.static()` multiple times to configure the app to serve static
+assets from different drectories.
+
+```ts
+app.static('/files', path.join(__dirname, 'files'));
+app.static('/downloads', path.join(__dirname, 'mp3s'));
+```
+
+You can also call `app.static()` multiple times on the same mounth path to serve
+static assets from different directories on the same path.
+
+```ts
+app.static('/files', path.join(__dirname, 'files'));
+app.static('/files', path.join(__dirname, 'other-files'));
+```
+
+And `app.static()` can be called even after the app have started.
+
+```ts
+await app.boot();
+await app.start();
+app.static('/files', path.join(__dirname, 'files'));
+```
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -202,6 +202,9 @@ children:
     - title: 'Deploying to IBM Cloud'
       url: Deploying-to-IBM-Cloud.html
       output: 'web, pdf'
+    - title: 'Serving static files'
+      url: Serving-static-files.html
+      output: 'web, pdf'
 
 - title: 'Booting an Application'
   url: Booting-an-Application.html


docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,64 @@
...
+directory named `public` at `/`.
+
+```ts
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+const path = require('path');
+
+export async function main(options?: ApplicationConfig) {
+  const app = new TodoListApplication(options);
+  app.static('/', path.join(__dirname, 'public'));
```
bajtos: When reading this code, it's not clear to me whether `__dirname` is project root or `src`. 

AFAICT, the `main` function is exported by `src/index.ts`, which is compiled to `dist/src/index.js`. As a result, your example is mounting the directory `dist/src/public` as the static assets. I believe this is not going to work in most projects because static assets are not copied over by `tsc`.

Please fix the path to point to the `public` directory in the project root.

Also to make this easier to understand for readers, add a [Code caption](https://loopback.io/doc/en/contrib/code_samples.html#code-caption) to tell readers what file should contain this code snippet.

```
{% include code-caption.html content="src/index.ts" %}
```

---

Personally, I think it's the responsibility of the `Application` class to lead the static assets. Consider the following acceptance test: [examples/todo/test/acceptance/home-page.controller.acceptance.ts](https://github.com/strongloop/loopback-next/blob/81c85d3659b7009f4437ab2795379ac2e38db19f/examples/todo/test/acceptance/home-page.controller.acceptance.ts#L6-L16):

```ts
import {Client} from '@loopback/testlab';
import {TodoListApplication} from '../..';
import {setupApplication} from './test-helper';


describe('HomePageController', () => {
  let app: TodoListApplication;
  let client: Client;

  before('setupApplication', async () => {
    ({app, client} = await setupApplication());
  });

  // ...
});
```

Because the static assets are configured only in the `main` function, this acceptance test will not see the static assets and thus cannot verify that static assets were correctly configured.

My proposal:

Move the code setting up static assets to Application's constructor.

```ts
export class TodoListApplication extends BootMixin(
  ServiceMixin(RepositoryMixin(RestApplication)),
) {
  constructor(options: ApplicationConfig = {}) {
    super(options);

    // ...

    this.static('/', path.join(__dirname, '../../public'));
  }
}
```

hacksparrow:
@bajtos I have made the changes as per your suggestions, PTAL.

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,64 @@
...
+`app.static(path, rootDir[, options])` method. The variables in the API are
+explained below.
+
+- `app`: An instance of a LoopBack application.
+- `path`: The path where the static assets are to be served from. Refer to
+  [path examples](https://expressjs.com/en/4x/api.html#path-examples) in the
+  Express docs for possible values.
+- `rootDir`: The directory where the static assets are located on the file
+  system.
+- `options`: An otional object for configuring the underlying
```
dhmlau: typo: `otional` -> `optional`

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,64 @@
...
+```ts
+export class TodoListApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options: ApplicationConfig = {}) {
+    super(options);
+
+    // ...
+
+    this.static('/', path.join(__dirname, '../../public'));
```
dhmlau: trying to verify your instruction.  Is it `options.path.join...` instead of `path.join`?  Because when i copied and pasted this line, it says "cannot find path".

jannyHou: @dhmlau `path` should be a node module. The sample code could import it to fix your error.

hacksparrow: Added `path` dependency for clarity.

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,66 @@
...
+permalink: /doc/en/lb4/Serving-static-files.html
+---
+
+One of the basic requirements of a web app is the ability to serve static files.
+Serving static files from a LoopBack is very simple - just call the
+`app.static(path, rootDir[, options])` method. The variables in the API are
+explained below.
+
+- `app`: An instance of a LoopBack application.
+- `path`: The path where the static assets are to be served from. Refer to
```
bajtos: Should we perhaps use `urlPath` instead of `path`, to avoid possible confusion between `path` as a static asset configuration and `path` as the built-in Node.js module?

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,66 @@
...
+You can call `app.static()` multiple times to configure the app to serve static
+assets from different drectories.
+
+```ts
+app.static('/files', path.join(__dirname, 'files'));
+app.static('/downloads', path.join(__dirname, 'mp3s'));
+```
+
+You can also call `app.static()` multiple times on the same mounth path to serve
+static assets from different directories on the same path.
```
bajtos: Can we make it more clear that this is effectively merging the content of multiple directories?

For example:

You can also call `app.static()` multiple times on the same mounth path to merge files from multiple filesystem directories and expose them on the same URL path. When a file with the same name is present in multiple directories mounted at the same URL path, then the precedence is given the file from the directory that was registered earlier.

docs/site/Serving-static-files.md
hacksparrow: Makes sense.

docs/site/Serving-static-files.md
hacksparrow: Much better.

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,69 @@
+---
+lang: en
+title: 'Serving static files in LoopBack 4'
+keywords: LoopBack
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Serving-static-files.html
+---
+
+One of the basic requirements of a web app is the ability to serve static files.
+Serving static files from a LoopBack is very simple - just call the
```
b-admike: nitpick: `from a LoopBack` -> `from a LoopBack application`

docs/site/Serving-static-files.md
```diff
@@ -0,0 +1,69 @@
...
+`app.static(urlPath, rootDir[, options])` method. The variables in the API are
+explained below.
+
+- `app`: An instance of a LoopBack application.
+- `urlPath`: The path where the static assets are to be served from. Refer to
+  [path examples](https://expressjs.com/en/4x/api.html#path-examples) in the
+  Express docs for possible values.
+- `rootDir`: The directory where the static assets are located on the file
+  system.
+- `options`: An optional object for configuring the underlying
```
b-admike: `underlying express static` -> `underlying express static middleware`?

Issue #1956, opened 2018-11-01
Atomic "find or create" Repository API


bajtos:
While implementing `hasOne` relation in https://github.com/strongloop/loopback-next/pull/1879, we discovered a requirement for an atomic "find or create" operation that's needed to avoid race conditions in `hasOne`'s implementation of `create` operation.

Cross-posting https://github.com/strongloop/loopback-next/pull/1879#discussion_r227363386

It is crucial to leverage an atomic implementation of `findOrCreate` provided by our connectors. The current proposal is prone to race conditions, where to instances of the target "hasOne" model can be created.

Consider the following scenario: the LB4 server is handling two incoming HTTP requests to create a hasOne target and the code is executed in the following way by Node.js runtime:

1. Request 1 arrives, `DefaultHasOneRepository#create` is invoked
2. `targetRepository.find` is called. It's an async function so other stuff happens while the query is in progress.
3. Request 2  arrives, `DefaultHasOneRepository#create` is invoked.
3. `targetRepository.find` is called. It's an async function so other stuff happens while the query is in progress.
4. `targetRepository.find` for Request 1 returns, no model was found. `targetRepository.create` in called.
5. `targetRepository.find` for Request 2 returns, no model was found. `targetRepository.create` in called.
6.  Both `create` requests eventually finish. We have two models that are a target of `hasOne` relation. 💥 💣 💥  

## Acceptance criteria

- [ ] Add a new Repository-like interface describing atomic operations.

    ```ts
   export interface FindOrCreateResult<T extends Entity> {
     entity: T;
     found: boolean;
   }

    export interface AtomicCrudRepository<T extends Entity, ID>
       extends EntityCrudRepository<T, ID> {
        findOrCreate(
         filter: Filter<T>,
         entity: DataObject<T>,
         options?: Options,
       ): Promise<FindOrCreateResult>
    }
    ```

- [ ] Add a new repository class implementing this new interface while extending `DefaultCrudRepository`. Either the constructor or the `findOrCreate` method should verify that the connector provides atomic implementation of `findOrCreate` and throw an error with a machine-readable code if it does not.

- [ ] Modify the table mapping known error codes to HTTP status codes to map the new error code to `501 Not Implemented`.

Nice to have but not required:

- [ ] A new repository class implementing `AtomicCrudRepository` on top of `CrudRepositoryImpl`.

## Out of scope

- Support for other atomic operations like `patchOrCreate`, `replaceOrCreate`, etc.

[...]

PR #2005, opened 2018-11-09
feat(repository): hasOne relation

b-admike:
**EDIT - Continuation of #1879 **

> First iteration of `hasOne` relation. This PR includes the the repository factory, repository interface and default hasOne repository, and has one decorator implementations (mostly taken from our hasMany implementation since they're very similar in nature). It also includes an acceptance test with a Customer has one Address relation to drive it and a unit test as well. Please note that only the **EDIT** `get` and `create` methods are implemented at the moment and tests covering them, so that we can focus on that and incrementally add patch and delete methods to the hasOne repository interface.
> 
> **Todos**
> - [ ] Add documentation for `hasOne` relation
> - [x] Add example that uses `hasOne` relation

See discussion in https://github.com/strongloop/loopback-next/issues/1956. @bajtos proposed using the PK of a target model in a `hasOne` relation to enforce uniqueness and 1-to-1 relation between source and target model instances (see https://github.com/strongloop/loopback-next/issues/1956#issuecomment-437360566). ~This is just a quick spike to see the feasibility of the proposal and get more feedback.~ 

## Checklist


bajtos:
Looks good to me at high level 👍 But! One of the has-many tests is failing, so your implementation may need few tweaks?

I think the next task is to verify this implementation against real databases, e.g. MongoDB and MySQL. It would be great to have a full example app with hasOne relation configured for that. Can we come up with a meaningful entity that we could add to the existing todo-list example?

packages/repository/src/relations/has-one/index.ts
```diff
@@ -4,3 +4,5 @@
 // License text available at https://opensource.org/licenses/MIT
 
 export * from './has-one.decorator';
+export * from './has-one.decorator';
```
bajtos: a duplicated import?

packages/repository/test/acceptance/has-one.relation.acceptance.ts
```diff
@@ -58,21 +58,17 @@ describe('hasOne relation', () => {
     expect(
       controller.createCustomerAddress(existingCustomerId, {
         street: '456 test street',
-        zipcode: '44012',
+        customerId: 44012,
       }),
-    ).to.be.rejectedWith(
-      /does not allow creation of more than one target model instance/,
-    );
+    ).to.be.rejectedWith(/Property "customerId" cannot be changed!/);
```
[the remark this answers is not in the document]
b-admike: I wrote the test wrong :(. I've updated it!

examples/todo-list/src/models/author.model.ts
```diff
@@ -0,0 +1,18 @@
+import {Entity, model, property, belongsToUniquely} from '@loopback/repository';
+import {TodoList} from './todo-list.model';
+
+@model()
+export class Author extends Entity {
+  @belongsToUniquely(() => TodoList)
```
bajtos: This scenario does not make sense to me. In a typical application, a single author has many todos.

What are the arguments for using Author instead of other alternatives you have considered?

examples/todo-list/src/controllers/author.controller.ts
```diff
@@ -0,0 +1,122 @@
...
+  getFilterSchemaFor,
+  getWhereSchemaFor,
+  patch,
+  del,
+  requestBody,
+} from '@loopback/rest';
+import {Author} from '../models';
+import {AuthorRepository} from '../repositories';
+
+export class AuthorController {
```
bajtos: Since every author belongs to a single todo, does it actually make sense to expose authors directly? Isn't it better to expose only the has-one relation API?

Just an idea to consider.

examples/todo-list/src/repositories/todo-list.repository.ts
```diff
@@ -21,17 +23,27 @@ export class TodoListRepository extends DefaultCrudRepository<
...
+    protected todoListImageRepositoryGetter: Getter<AuthorRepository>,
   ) {
     super(TodoList, dataSource);
     this.todos = this._createHasManyRepositoryFactoryFor(
       'todos',
       todoRepositoryGetter,
     );
+    this.author = this._createHasOneRepositoryFactoryFor(
+      'author',
+      todoListImageRepositoryGetter,
```
bajtos: `todoListImageRepositoryGetter` is inconsistent with the target model name `Author`. Did you mean `authorRepositoryGetter` instead?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -0,0 +1,87 @@
...
+    filter?: Exclude<Filter<TargetEntity>, Where<TargetEntity>>,
+    options?: Options,
+  ): Promise<TargetEntity> {
+    const targetRepository = await this.getTargetRepository();
+    const found = await targetRepository.find(
+      Object.assign({limit: 1}, constrainFilter(filter, this.constraint)),
+      options,
+    );
+    if (found.length < 1) {
+      // We don't have a direct access to the foreign key value here :(
```
bajtos: Should we leave a TODO comment here? Maybe open a new issue to follow-up later?

packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts
```diff
@@ -7,7 +7,7 @@ import {MetadataInspector} from '@loopback/context';
 import {expect} from '@loopback/testlab';
 import {RelationMetadata} from '../../..';
 import {
-  belongsTo,
+  belongsToUniquely,
```
bajtos: I think we should preserve existing tests for `belongsTo` and create new tests verifying specific differences of `belongsToUniquely`. Thoughts?

packages/repository/test/fixtures/repositories/customer.repository.ts
```diff
@@ -10,8 +10,10 @@ import {
   juggler,
   repository,
 } from '../../..';
-import {Customer, Order} from '../models';
+import {Customer, Order, Address} from '../models';
 import {OrderRepository} from './order.repository';
+import {HasOneRepositoryFactory} from '../../../src';
```
bajtos: Don't import from `src`, just `../../..`

examples/todo-list/src/models/author.model.ts
b-admike: Well I was thinking of "a todo-list can only have one owner/author" and thus this model. But yea, as discussed, I'm going to revert to `todoListImage` model and describe the limitation for the binary data property in a follow-up issue and leave it as string for now.

examples/todo-list/src/controllers/author.controller.ts
b-admike: Yeah I was contempleting that as well and I think it makes sense just to expose the `has-one` relation APIs.

examples/todo-list/src/controllers/todo-list-image.controller.ts
```diff
@@ -0,0 +1,37 @@
+import {TodoListRepository} from '../repositories';
+import {repository, Filter} from '@loopback/repository';
+import {param, post, requestBody, get} from '@loopback/rest';
+import {TodoListImage} from '../models';
+
+export class TodoListAuthorController {
```
bajtos: forgotten `Author`

packages/repository/test/acceptance/has-one.relation.acceptance.ts
```diff
@@ -0,0 +1,165 @@
...
+      customerId: existingCustomerId,
+      street: '123 test avenue',
+    });
+
+    const persisted = await addressRepo.findById(address.customerId);
+    expect(persisted.toObject()).to.deepEqual(address.toObject());
+  });
+
+  it("doesn't allow to create related model instance twice", async () => {
+    const address = await controller.createCustomerAddress(existingCustomerId, {
```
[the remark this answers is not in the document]
b-admike: Yes the `existingCustomerId` is created once for the whole test suite, but the address model instances are cleaned up in the `beforeEach()` block, so running this test with `it.only()` works the same as running the full test suite.

packages/repository/test/acceptance/has-one.relation.acceptance.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  it('does not allow where filter to find related model instance', async () => {
+    const address = await controller.createCustomerAddress(existingCustomerId, {
+      street: '123 test avenue',
+    });
+
+    const foundAddress = await controller.findCustomerAddressWithFilter(
+      existingCustomerId,
+      {where: {street: '123 test avenue'}},
```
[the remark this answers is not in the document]
b-admike: The purpose of the `filter` argument here is to accept things like `fields` in it, but it also ignores `where` clauses since there is only one instance (the type for it is `filter?: Exclude<Filter<TargetEntity>, Where<TargetEntity>>`). This test is to ensure that it ignores the `where` object.

packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts
```diff
@@ -93,14 +94,30 @@ describe('model decorator', () => {
...
     id: string;
 
-    @belongsTo(() => Customer)
+    @belongsTo(
+      () => Customer,
+      {},
+      {
+        id: true,
+        generated: false,
+      },
```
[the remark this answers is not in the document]
b-admike: Actually the current test case is still there at https://github.com/strongloop/loopback-next/pull/2005/files/3925655eb77bdbcf911976a93ee86aa82775bcab#diff-2ccb195898f4b3a07caf05f6b513a9c1L272.

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
```diff
@@ -11,8 +11,11 @@ import {BelongsToDefinition, RelationType} from '../relation.types';
 
 /**
  * Decorator for belongsTo
- * @param targetResolver
- * @param definition
+ * @param targetResolver A resolver function that retruns the target model for
```
bajtos: typo: `retruns`

packages/repository/test/acceptance/has-one.relation.acceptance.ts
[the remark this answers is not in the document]
bajtos: 👍

packages/repository/test/unit/repositories/has-one-repository-factory.unit.ts
```diff
@@ -132,7 +132,7 @@ describe('createHasOneRepositoryFactory', () => {
     province: String;
   }
 
-  class Order extends Entity {
+  class ModelWithoutIndexFK extends Entity {
     static definition = new ModelDefinition('Order')
```
bajtos: Please fix the model name string to match the class name.

packages/repository/test/unit/repositories/has-one-repository-factory.unit.ts
```diff
@@ -142,18 +142,23 @@ describe('createHasOneRepositoryFactory', () => {
         type: 'string',
         required: true,
       })
+      .addProperty('id', {
+        type: 'number',
+        id: true,
```
bajtos: I am little bit confused. The model name is `ModelWithoutIndexFK`, but it has a regular `id` property. What am I missing?

b-admike: Good question. I added an `id` property because the model itself extends from `Entity`, but the purpose here is the decorated relational property is not part of the composite index, thus the name `ModelWithoutIndexFK`. Thoughts on improving it if it's confusing? I can extend from `Model` and remove the `id` property as well.

bajtos: >  I can extend from Model and remove the id property as well.

If this works, then I think it would be a good solution.

If not, then please capture the information from your comment above in a code comment.

b-admike:
>I have a question regarding the BelongsToUnique decorator: My understanding is, regardless it's a hasMany relation or hasOne relation, one child instance could only have one parent, why would we create a specific belongsToUnique decorator for hasOne?

Great question @jannyHou. The reason we created a sugar syntax decorator `belongsToUniquely` is so that the decorated property is designated as a PK for the target model (needed for hasOne relation). It is to save time to label it as an `id` property that is not `generated`. It makes the UX better to not manually write those out.

packages/repository/test/unit/decorator/model-and-relation.decorator.unit.ts
```diff
@@ -93,14 +94,30 @@ describe('model decorator', () => {
...
 
+  @model()
+  class RegistrationDate extends Entity {
+    @belongsToUniquely(() => Customer)
+    customerId: number;
+
+    @property()
+    registeredOn: Date;
+  }
+
```
raymondfeng: It's not so obvious that `RegistrationDate` deserves a `hasOne` relation from the `Customer`. Something like `A customer has a public profile` is probably better.

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
```diff
@@ -53,3 +58,22 @@ export function belongsTo<T extends Entity>(
...
+ * @returns {(target: Object, key:string)}
+ */
+export function belongsToUniquely<T extends Entity>(
+  targetResolver: EntityResolver<T>,
+  definition?: Partial<BelongsToDefinition>,
+) {
+  const propertyMetaDefaults: Partial<PropertyDefinition> = {
+    id: true,
+    generated: false,
+  };
```
raymondfeng: Is the intention to make the FK unique? If so, I think it's wrong to set `id` to `true` as it will add the FK to the PK. For example, `(id, customerId)` becomes the PK.

I think we need to mark it as `unique`. IIRC, we can mark it uniquely indexed.

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
```diff
@@ -53,3 +58,22 @@ export function belongsTo<T extends Entity>(
...
+
+/**
+ * Sugar syntax for belongsTo decorator for hasOne relation. Calls belongsTo
+ * with property definition defaults for non-generated id field
+ * @param targetResolver A resolver function that returns the target model for
+ * a belongsTo relation
+ * @param definition Optional metadata for setting up a belongsTo relation
+ * @returns {(target: Object, key:string)}
+ */
+export function belongsToUniquely<T extends Entity>(
```
raymondfeng: The name is not easy to understand. The purpose is to say the FK should be unique.

raymondfeng: Maybe it's a warning sign that we should not overload `@belongTo` for responsibility of `@property`.

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
bajtos: > Is the intention to make the FK unique?

Yes.

> If so, I think it's wrong to set id to true as it will add the FK to the PK. For example, (id, customerId) becomes the PK.

The way we have envisioned the current hasOne implementation, there is only one PK property, and that's `customerId`, there is no property called `id`. I.e. the primary key is a foreign key too.

> I think we need to mark it as unique. IIRC, we can mark it uniquely indexed.

I am not sure if all databases support `unique` constraint. I am also pretty sure our auto-migration implementation does not correctly define `unique` constraint either.

As a result, by using `unique` instead of `id`, the default implementation of `hasOne` in LB4 would not work correctly out of the box.

On the other hand, all databases are enforcing the uniqueness constraint on primary keys (id property), otherwise primary keys could not be used as unique identifiers.

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
bajtos: I agree that ideally, relation decorators like `@belongsTo` and `@hasMany` should not call `@property` decorator under the hood. To get there, we would have to change the current design in a non-trivial way and break backwards compatibility along the way.

Do we want to delay `hasOne` implementation?

Personally, I'd prefer to finish `hasOne` implementation using the current approach, and then look into ways how to change the overall design.

> The name is not easy to understand. The purpose is to say the FK should be unique.

Can you propose a better name that's easier to understand?

Our intention here is to provide a syntactic sugar that will allow users to tell what they want to accomplish (setup the other side of hasOne relation) without going into low-level details (setting up a PK property as a way to get uniqueness constraint + make queries fast).

packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
raymondfeng: I think it's too restrictive to promote the FK for `hasOne` to be part of PK. At least we should check if there is an existing PK.

BTW, marking a property `id: true` does not guarantee it will be enforced by the database. I think we should generate an index with `{unique: true}` option per https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#indexes.

bajtos: Not only we are promoting FK to be the only PK, but we also require the target od HasOne to have the same PK as the model it relates to. E.g. when a Customer has one PublicProfile, then both the customer instance and the profile instance share the same primary key value.

This is the only way I am aware of that offers strong guarantees about consistency and avoid possible race conditions in all databases we support in LoopBack.

I am more than happy to consider different approaches, as long as they work similarly well across the databases we have connectors for.

Should we open a new issue to continue this discussion?

RaphaelDrai:
Hello @b-admike ,
Great work with the merge of hasOne, thanks you so much !
As discussed in #1879 we are currently working on hasOne example and documentation as per todos requirement above. 
Is there an estimation when the missing PATCH and DELETE operations will be covered? If not, and that no one is actually working on PATCH and DELETE, we can contribute to provide these features. Let me know if we can start this task?
Raphael

b-admike:
hey @RaphaelDrai, sorry just saw your comment; based on the comments from @raymondfeng and @bajtos and discussion with @raymondfeng, I've decided to try the `unique index` approach for the FK relation property on the target model for a `hasOne` relation. With it, I plan to add the PATCH and DELETE operations and docs for the `hasOne` feature. Are you able to open a PR with your work on the example and its documentation? If so, we can definitely help you land it 👍

RaphaelDrai:
Hello @b-admike,
thanks you for the update about the DELETE and PATCH operations, this is great... 🥇 
Just for our internal knowledge, do you have a time estimation when they will be available?
Yes, we will be very happy to get your help of how creating a PR with our work and how to publish our hasOne example and documentation.

RaphaelDrai:
> > Hello @b-admike,
> > thanks you for the update about the DELETE and PATCH operations, this is great... 🥇
> > Just for our internal knowledge, do you have a time estimation when they will be available?
> > Yes, we will be very happy to get your help of how creating a PR with our work and how to publish our hasOne example and documentation.
> 
> @RaphaelDrai Right now I'm using #2126 for enforcement of uniqueness on the target model instead of marking the foreign key as a primary key in the current design. @bajtos @raymondfeng and I had a discussion and since the memory connector and cloudant connector do not support unique indexes, we've decided to have the first release of `hasOne` provide "weak" relations i.e. ease of navigation and aggregation instead of enforcement of referential integrity and document it so users are aware of the limitations. We will set the metadata for the foreign key to be unique index, but it is up to users to run automigrate and if not supported, to set up the referential integrity in their collections/tables. I'm working on DELETE and PATCH operations at the moment in another branch as well as supporting unique indexes for memory connector (see [strongloop/loopback-datasource-juggler#1672](https://github.com/strongloop/loopback-datasource-juggler/pull/1672)) and I would like to ideally wrap them up this week. Can you let us know a bit more about your use case for the `hasOne` relation? I.e. what are your tables defined as and what is your expectation to enforce unique constraint?

Thanks you for your detailed and helpful answer.
Our particular use case for the 'hasOne' relation is creating a connection from a model let's say "Country" with another model let's say "Capital City". The relation should follow the "one-to-one" relation where a given "Country" can have only one "Capital City" and vice versa, a given "Capital City" can have (belongs to) only one "Country". 

I read the previous messages about "belongsToUnique" decorator. I think that it could be helpful to have the option to use "belongsTo" or "belongsToUnique" in 'hasOne' relation.
For example if we use the decorator "belongsToUnique" it will address our use case above where in both directions the one-to-one relation is preserved and ensured.
But if we use the decorator "belongsTo" it will address the use case of one-to-one in one direction, and in the opposite direction we can allow one-to-many. 
Example of this use case is a data model "Customer" connected to a data model "Orders" using the 'hasOne' relation. A given "Order" can have only one Customer, this is enforced with the 'belongsToUnique', but in the opposite direction, "Customer" can have many orders.

b-admike:
> Thanks you for your detailed and helpful answer.
> Our particular use case for the 'hasOne' relation is creating a connection from a model let's say "Country" with another model let's say "Capital City". The relation should follow the "one-to-one" relation where a given "Country" can have only one "Capital City" and vice versa, a given "Capital City" can have (belongs to) only one "Country".
> 
> I read the previous messages about "belongsToUnique" decorator. I think that it could be helpful to have the option to use "belongsTo" or "belongsToUnique" in 'hasOne' relation.
> For example if we use the decorator "belongsToUnique" it will address our use case above where in both directions the one-to-one relation is preserved and ensured.
> But if we use the decorator "belongsTo" it will address the use case of one-to-one in one direction, and in the opposite direction we can allow one-to-many.
> Example of this use case is a data model "Customer" connected to a data model "Orders" using the 'hasOne' relation. A given "Order" can have only one Customer, this is enforced with the 'belongsToUnique', but in the opposite direction, "Customer" can have many orders.

Thanks for your feedback @RaphaelDrai. Those sound like good use cases. Unfortunately, I'm going to revert `belongsToUnique` implementation since it marks the foreign key as a primary key on the target model (in this case `Capital City`). We have agreed to let our users/DBAs set up the unique constraint(s) for the relation and provide the "weak" relation I've mentioned above. I have a PR on the way and I will ping you as well to give your feedback/review.

PR #2016, opened 2018-11-12
feat(rest): add config option to disable API Explorer redirects

bajtos:
To allow users to mount self-hosted API Explorer at `/explorer`, we need to allow them to disable the redirect pointing `/explorer` to the externally hosted swagger-ui instance.

In this pull request, I am modifying the REST component to allow the config option `rest.apiExplorer` to be configured to `false`, thus disabling redirects.

## Checklist

- n/a API Documentation in code was updated
- n/a Affected artifact templates in `packages/cli` were updated


packages/rest/src/rest.server.ts
```diff
@@ -837,7 +835,7 @@ export interface ApiExplorerOptions {
 export interface RestServerOptions {
   cors?: cors.CorsOptions;
   openApiSpec?: OpenApiSpecOptions;
-  apiExplorer?: ApiExplorerOptions;
+  apiExplorer?: ApiExplorerOptions | false;
```
raymondfeng: Maybe a flag inside `ApiExplorerOptions` is better off. For example:
```ts
const app = new RestApplication({
  rest: {
    apiExplorer: {disabled: true},
  },
});
```

We can make it consistent across all options. Such a convention will help tooling and documentation.

bajtos: What's wrong with the convention of using `false` instead of `{disabled: true}`? What problems do you see in my proposal, and what benefits does your alternative brings? Let's discuss.

raymondfeng: IMO, having it as an attribute is simpler for the code and tooling:

1. TypeScript can simply declare a property as `disabled?: boolean` instead of a union type. No code is needed for normalization.

2. Building a UI for the config is simpler as the front-end does not have to handle a `false` value. I have seen a few cases that UI fails due to `false` value for a typical object.

raymondfeng: Anyway, it's not a blocker.

bajtos: Your arguments make sense 👍

Issue #2033, opened 2018-11-16
Unable to deploy successfully to Cloud Foundry


cloudwheels:
## Description / Steps to reproduce / Feature proposal
- Create node sdk app on Cloud Foundry
- Generate lb4 starter app & test 'locally' (I am using Google Cloud Shell, which has some other implications e.g. inability to use cloudant in docker as per deployment docs, but the app works, as do other starter apps, including samples using cloudant connector to access the IBM hosted cloudant)
- remove pre-start from package.json (as per deployment docs - also tried with leaving this)
- login to ibm cloud : I found all the various CLI flavours of this very confusing as someone new to the whole IBM eco system and have worked with `cf` (as per deployment docks), `ibmcloud cf` `bx` (as per the cloud foundry node.js documentation after setting up an app. There are very few examples of this process around and those that exist seem to be conflicting, although they all seem to achieve the same results.
Depending on which you choose and what prompts you receive, it can take some time to work out exactly what the required 'endpoint', 'org' and 'space' parameters for logging in and settings app endpoints. This is not covered at all in the deployment documentation.
- possibly create a manifest - using `bx` and specifying app name seemed to negate the need for this.
- push to the cf app
## Current Behavior
The app seems to build, start successfully and then crash, most lately / consistently with exit code 137 (which may possibly indicate memory problems with the cf container?).
The app's routes are not available in the cloud.
Extract of log, my package.json & node version below.

Please help - this has been driving me mental for 24hours, I've had plenty of stabs at different approaches but it just doesn't seem to work. I must be missing something really obvious but I have the fear that this version is all a bit bleeding edge right now and things seems to be in a state of flux with the IBM CLI tools!

```
...
2018-11-16T00:20:11.38+0000 [APP/PROC/WEB/0] OUT Server is running at http://127.0.0.1:8080
   2018-11-16T00:20:11.38+0000 [APP/PROC/WEB/0] OUT Try http://127.0.0.1:8080/ping
   2018-11-16T00:21:05.96+0000 [CELL/0] ERR Failed to make TCP connection to port 8080: connection refused
   2018-11-16T00:21:05.96+0000 [CELL/0] ERR Timed out after 1m0s: health check never passed.
   2018-11-16T00:21:05.97+0000 [CELL/SSHD/0] OUT Exit status 0
[...]

PR #2039, opened 2018-11-16
docs: update steps to deploy to IBM Cloud

dhmlau:
Originated from https://github.com/strongloop/loopback-next/issues/2033

In this PR, 
- fix the code snippet in `index.ts` to get it working
- add the instruction to sign up for IBM Cloud and include information on creating a Cloudant service. 

Will submit another PR for some more improvements, but would like to get this PR up first so that users won't get stuck on the outdated instruction.


docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -120,11 +147,12 @@ const datasourceDb = require('./datasources/db.datasource.json');
 const cfenv = require('cfenv');
 const appEnv = cfenv.getAppEnv();
 
-export async function main(options?: ApplicationConfig) {
+export async function main(options: ApplicationConfig = {}) {
```
b-admike: Hmm why do we want to make `options` a required parameter here?

dhmlau: that's good point. That's what I got after generating the app, but I just updated to the latest @loopback/cli and seems like the original line is correct. Will change it back.

Issue #2002, opened 2018-11-09
Regression: Model names not used for queries.


jgutix:
## Description / Steps to reproduce / Feature proposal
Having this model definition:
```
@model({
    name: 'orden'
})
export class Order extends Entity {
```

It no longer uses the `name` attribute in the model decorator for building the queries to the database.

## Current Behavior
The queries will use the class name as the table name, resulting in errors like this: `Unhandled error in GET /orders: 500 Error: ER_NO_SUCH_TABLE: Table 'mydb.orders' doesn't exist`

## Expected Behavior
The queries will be built using the `name` property set in the model decorator.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

themadrix:
In our case:
```
@model({
	name: "articles"
})
export class Article extends Entity {
```
The following code in the controller queries the table `Article` in the database, resulting an error:
```
@get('/articles')
  async find(
    @param.query.object('filter', getFilterSchemaFor(Article)) filter?: Filter,
  ): Promise<Article[]> {
    return await this.articleRepository.find(filter);
  }
```
Otherwise, if we exclude the filter in the controller, it queries the right table `articles`:
```
@get('/articles')
  async find(
  ): Promise<Article[]> {
    return await this.articleRepository.find();
  }
```

themadrix:
We are not experiencing this issue when using `@loopback/repository-json-schema@1.0.0`, but it appears with versions `1.2.0` and higher (haven't tried with versions in between).

PR #2044, opened 2018-11-16
fix(repository): make sure model definition is built correctly

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/2002

## Checklist


raymondfeng:
> can we add a test which simulates #2002 with a different model name specified in the model decorator than the TS model class?

MyUserModel is decorated with `@model({name: 'my-user-model')`.

PR #2055, opened 2018-11-19
docs: fix image links in Cloud Deployment guide

dhmlau:
In the [IBM Cloud deployment guide](https://loopback.io/doc/en/lb4/Deploying-to-IBM-Cloud.html), the image links are broken.  
It is because the relative path that the image refers to will bring it to: https://loopback.io/doc/en/lb4/imgs/deploytocloud-mycloudant.png, where the actual image is found in https://loopback.io/pages/en/lb4/imgs/deploytocloud-mycloudant.png.

### DIFF
--- a/docs/site/Deploying-to-IBM-Cloud.md
+++ b/docs/site/Deploying-to-IBM-Cloud.md
@@ -53,7 +53,7 @@ Cloudant connector, so data is persisted.
   and resource group. Select "Use both legacy credentials and IAM" as the
   available authentication methods
 
-  <img src="./imgs/deploytocloud-mycloudant.png" alt="myCloudant service name" style="width: 300px"/>
+  ![myCloudant service name](./imgs/deploytocloud-mycloudant.png)
 
 - Click Create.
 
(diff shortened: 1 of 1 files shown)


docs/site/Deploying-to-IBM-Cloud.md
```diff
@@ -53,7 +53,7 @@ Cloudant connector, so data is persisted.
   and resource group. Select "Use both legacy credentials and IAM" as the
   available authentication methods
 
-  <img src="./imgs/deploytocloud-mycloudant.png" alt="myCloudant service name" style="width: 300px"/>
+  <img src="https://loopback.io/pages/en/lb4/imgs/deploytocloud-mycloudant.png" alt="myCloudant service name" style="width: 300px"/>
```
raymondfeng: I wonder why the relative path does not work. If we keep the same directory structure for loopback.io and github, it should work for both.

raymondfeng: FYI: on https://loopback.io/doc/en/lb4/Deploying-to-IBM-Cloud.html, the broken image is https://loopback.io/doc/en/lb4/imgs/deploytocloud-mycloudant.png.

dhmlau: I remember having the same problem when adding a diagram in todo example (https://raw.githubusercontent.com/strongloop/loopback-next/master/examples/todo/README.md).  TV was looking at it and couldn't get it to work, and finally change the image url to `todo-tutorial-overview`.

raymondfeng: Ideally, we should make it work for both github and loopback.io.

Issue #487, opened 2017-08-03
[LB-Next] Migration Management Framework


kjdelisle:
_From @kjdelisle on July 28, 2017 18:37_

## Scenario
FooCorp wants to deploy version 2.0 of their LoopBack application on their production server, but doing so will require some alteration of the schema in their database. Some of these changes require a more delicate touch that cannot necessarily be performed by the `autoupdate` function.

Currently, FooCorp would have to create their own scripts and execute them to modify their database schema to match their expectations, which may require them to more directly interface with drivers that are used by their chosen LoopBack connector, or using another implementation entirely!

## Proposal

Create a component for use in loopback-next that handles the idea of production-system migration within a LoopBack application as it is upgraded from one release to another.

Originally, I had a design in my mind for how this would work, but I noticed that there are some generic migration tools on npm (https://www.npmjs.com/package/migration), and I was wondering if there was anything additional that we'd want to add to make it easier to do this sort of thing.

**Just to be clear, this should not be a part of LoopBack core, or any of the offerings we intend to finish before GA**

@ritch @raymondfeng @bajtos @superkhau

_Copied from original issue: strongloop-internal/scrum-apex#290_

kjdelisle:
_From @bajtos on July 31, 2017 14:38_

I quite like the approach of Knex - see http://perkframework.com/v1/guides/database-migrations-knex.html. It's similar e.g. to [EntityFramework](https://msdn.microsoft.com/en-us/library/jj591621(v=vs.113).aspx). I have personally used this approach at my previous job and our experience was great.

 - Migrations are granular and version-controlled together with application code.
 - Each code change in the app is accompanied by a small migration script.
 - Each migration script has a unique id.
 - The database contains a special table containing list of migration script ids that have been applied.
 - There is tooling to automatically run missing migration scripts to update the database from the current schema to the latest (or any other) version.

Since we already have autoupdate functionality, our tooling should be able to generate the initial version of a migration script (create newly added columns, remove deleted columns, etc.)

kjdelisle:
_From @bajtos on July 31, 2017 14:39_

Can we move this issue to loopback-next repository please, so that we can discuss this feature in public and get early feedback?

ritch:
In the spirit of LoopBack.next, we should build the foundational features for a framework like this, document those really well, and support 3rd party implementations. If one of those implementations gains a lot of adoption, we can pull it into the modules in this repo.

[...]

PR #2059, opened 2018-11-20
Easy-to-use database migrations at application level

bajtos:
I find the current instructions for database migration as too involved for practical use, for example when running our example applications against a real database (SQL, MongoDB). I would like to propose a better user experience as a short-term solution to scratch this itch. Please see https://github.com/strongloop/loopback-next/issues/487 for the discussion about a more comprehensive solution for long term.

In this pull request, I am introducing new APIs to simplify database migrations:

- ~~A new interface `MigrateableRepository` describing repositores that know how to migrate their database schema.~~

- A new Application-level method `app.migrateSchema()` provided by `RepositoryMixin` and running schema migration provided by all registered ~~repositories~~ datasources under the hood.

Using these two new APIs, I also

- ~~Implement `MigrateableRepository` in `DefaultCrudRepository` using autoupdate/automigrate APIs provided by legacy-juggler's  DataSource class.~~

- Simplify the instructions shown in `Database-migrations.db`

- Add an example `migrate.ts` script to our Todo example to verify that the code snippet shown in the docs works as intended.

## Checklist

- out of scope: Affected artifact templates in `packages/cli` were updated

### DIFF
--- a/docs/site/Database-migrations.md
+++ b/docs/site/Database-migrations.md
@@ -20,123 +20,111 @@ LoopBack offers two ways to do this:
 - **Auto-update**: Change database schema objects if there is a difference
   between the objects and model definitions. Existing data will be kept.
 
-## Implementation Example
+{% include warning.html content="Auto-update will attempt to preserve data while
+updating the schema in your target database, but this is not guaranteed to be
+safe.
 
-Below is an example of how to implement
-[automigrate()](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-automigrate)
-and
-[autoupdate()](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-autoupdate),
-shown with the
-[TodoList](https://loopback.io/doc/en/lb4/todo-list-tutorial.html) example.
+Please check the documentation for your specific connector(s) for a detailed
+breakdown of behaviors for automigrate! " %}
 
-Create a new file **src/migrate.ts** and add the following import statement:
+## Examples
 
-```ts
-import {DataSource, Repository} from '@loopback/repository';
-```
+LoopBack applications are typically using `RepositoryMixin` to enhance the core
+`Application` class with additional repository-related APIs. One of such methods
+is `migrateSchema`, which iterates over all registered repositories and asks
+them to migrate their schema. Repositories that do not support schema migrations
+are silently skipped.
 
-Import your application and your repositories:
+In the future, we would like to provide finer-grained control of database schema
+updates, learn more in the GitHub issue
+[#487 Database Migration Management Framework](https://github.com/strongloop/loopback-next/issues/487)
 
-```ts
-import {TodoListApplication} from './index';
-import {TodoRepository, TodoListRepository} from './repositories';
-```
+### Auto-update database at start
 
-Create a function called _dsMigrate()_:
+To automatically update the database schema whenever the application is started,
+modify your main script to execute `app.migrateSchema()` after the application
+was bootstrapped (all repositories were registered) but before it is actually
+started.
 
-```ts
-export async function dsMigrate(app: TodoListApplication) {}
-```
-
-In the _dsMigrate()_ function, get your datasource and instantiate your
-repositories by retrieving them, so that the models are attached to the
-corresponding datasource:
+{% include code-caption.html content="src/index.ts" %}
 
 ```ts
-const ds = await app.get<DataSource>('datasources.db');
-const todoRepo = await app.getRepository(TodoRepository);
-const todoListRepo = await app.getRepository(TodoListRepository);
-```
-
-Then, in the same function, call _automigrate()_:
-
-```ts
-await ds.automigrate();
-```
-
-This call to automigrate will migrate all the models attached to the datasource
-db. However if you want to only migrate some of your models, add the names of
-the classes in the first parameter:
-
-```ts
-// Migrate a single model
-ds.automigrate('Todo');
-```
-
-```ts
-// Migrate multiple models
-ds.automigrate(['Todo', 'TodoList']);
-```
-
-The implementation for _autoupdate()_ is similar. Create a new function
-_dsUpdate()_:
-
-```ts
-export async function dsUpdate(app: TodoListApplication) {
-  const ds = await app.get<DataSource>('datasources.db');
-  const todoRepo = await app.getRepository(TodoRepository);
-  const todoListRepo = await app.getRepository(TodoListRepository);
-
-  await ds.autoupdate();
-}
-```
-
-The completed **src/migrate.ts** should look similar to this:
-
-```ts
-import {DataSource, Repository} from '@loopback/repository';
-import {TodoListApplication} from './index';
-import {TodoRepository, TodoListRepository} from './repositories';
-
-export async function dsMigrate(app: TodoListApplication) {
-  const ds = await app.get<DataSource>('datasources.db');
-  const todoRepo = await app.getRepository(TodoRepository);
-  const todoListRepo = await app.getRepository(TodoListRepository);
-
-  await ds.automigrate();
-}
-
-export async function dsUpdate(app: TodoListApplication) {
-  const ds = await app.get<DataSource>('datasources.db');
-  const todoRepo = await app.getRepository(TodoRepository);
-  const todoListRepo = await app.getRepository(TodoListRepository);
-
-  await ds.autoupdate();
-}
-```
-
-Finally, in **src/index.ts**, import and call the _dsMigrate()_ or _dsUpdate()_
-function:
-
-```ts
-import {TodoListApplication} from './application';
-import {ApplicationConfig} from '@loopback/core';
-
-// Import the functions from src/migrate.ts
-import {dsMigrate, dsUpdate} from './migrate';
-
-export {TodoListApplication};
-
 export async function main(options: ApplicationConfig = {}) {
   const app = new TodoListApplication(options);
   await app.boot();
+  await app.migrateSchema();
   await app.start();
 
   const url = app.restServer.url;
   console.log(`Server is running at ${url}`);
 
-  // The call to dsMigrate(), or replace with dsUpdate()
-  await dsMigrate(app);
   return app;
+}
+```
+
+### Auto-update the database explicitly
+
+It's usually better to have more control about the database migration and
+trigger the updates explicitly. To do so, you can implement a custom script as
+shown below.
+
+{% include code-caption.html content="src/migrate.ts" %}
+
+```ts
+import {TodoListApplication} from './application';
+
+export async function migrate(args: string[]) {
+  const dropExistingTables = args.includes('--rebuild');
+  console.log('Migrating schemas (%s)', rebuild ? 'rebuild' : 'update');
+
+  const app = new TodoListApplication();
+  await app.boot();
+  await app.migrateSchema({dropExistingTables});
+}
+
+migrate(process.argv).catch(err => {
+  console.error('Cannot migrate database schema', err);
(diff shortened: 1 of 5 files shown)


packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -163,6 +170,34 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
...
+     * while updating the schema in your target database, but this is not
+     * guaranteed to be safe.
+     *
+     * Please check the documentation for your specific connector(s) for
+     * a detailed breakdown of behaviors for automigrate!
+     *
+     * @param options Migration options, e.g. whether to update tables
+     * preserving data or rebuild everything from scratch.
+     */
+    async migrateSchema(options?: SchemaMigrationOptions): Promise<void> {
```
raymondfeng: I would prefer to have two methods and use `updateSchema` as default.

- migrateSchema(...)
- updateSchema(...)

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -332,4 +336,9 @@ export class DefaultCrudRepository<T extends Entity, ID>
   protected toEntities(models: juggler.PersistedModel[]): T[] {
     return models.map(m => this.toEntity(m));
   }
+
+  async migrateSchema(options?: SchemaMigrationOptions): Promise<void> {
```
raymondfeng: Same here - `automigrate` and `autoupdate`.  At repository level, maybe it's better to keep the names as `automigrate` and `autoupdate`.

packages/repository/src/repositories/migrateable.repository.ts
```diff
@@ -0,0 +1,49 @@
...
+import {Options} from '../common-types';
+import {Model} from '../model';
+import {Repository} from './repository';
+
+export interface SchemaMigrationOptions extends Options {
+  /**
+   * When set to true, schema migration will drop existing tables and recreate
+   * them from scratch, removing any existing data along the way.
+   */
+  rebuild?: boolean;
```
raymondfeng: We should have an option to dry run - returning a bunch of SQL statements.

raymondfeng: The name `rebuild` is not clear. Maybe something like `dropThenRecreate`?

packages/repository/src/mixins/repository.mixin.ts
raymondfeng: Or at least an array of repo names.

packages/repository/src/repositories/migrateable.repository.ts
bajtos: +1

Unfortunately, our current juggler API does not provide dry-run feature, therefore I don't have any easy means for enabling dry run option. Let's leave it out of scope of this pull request.

In the future, I would like us to implement a mode where we create a SQL file containing the migration script, so that the user can customize it. See https://github.com/strongloop/loopback-next/issues/487.

bajtos:
@raymondfeng 

> I would prefer to have two methods and use updateSchema as default.
>
> - `migrateSchema(...)`
> - `updateSchema(...)`

Let's discuss.

I personally prefer a single method with an option to choose between an incremental update or destructive drop + create. I makes it easier to implement wrapper functions delegating the implementation to lower-level layers, e.g. a single `app.migrateSchema` calling a single `repository.migrateSchema` & forwarding the entire `options` argument.

When there are two methods, then we also need multiple wrapper functions (`app.migrateSchema` and `app.updateSchema`.

What are your arguments for having two methods? 

I also find the names "update" and "migrate" not descriptive enough, the term "migrate" does not imply drop+create in my mind. AFAIK, the term "migration" is usually used for non-destructive upgrades. Few references:
 - Ruby on Rails https://guides.rubyonrails.org/v3.2/migrations.html.
 - Martin Fowler on Evolutionary Database Design: https://www.martinfowler.com/articles/evodb.html#AllDatabaseChangesAreMigrations
 - .NET EntityFramework's Code First Migrations: https://docs.microsoft.com/en-us/aspnet/mvc/overview/getting-started/getting-started-with-ef-using-mvc/migrations-and-deployment-with-the-entity-framework-in-an-asp-net-mvc-application

> It should be possible to filter what repos should be migrated/updated.

I don't understand what's the use case for such filter.

Here is the user story I am trying to address: _As a developer wanting to run a LB4 app, I want an easy way how to update my database schema to match application's domain model._  In this scenario, we always want to update all tables.

What scenario do you @raymondfeng have in mind for choosing a subset of tables only? I think this filter is easy to add later as an incremental improvement of my MVP solution, thus I am proposing to leave it out of this initial pull request.

bajtos:
In d8c52a6, I reworked the underlying implementation to leverage DataSource-level `autoupdate`/`automigrate` API and thus allow connectors to set up foreign keys.

The pull request is ready for another round of review. @strongloop/loopback-maintainers PTAL.

I tested the migration with `examples/todo-list` and MySQL connector. The tables were created correctly, but unfortunately the foreign key constraint was not set up. I don't know whether this is a limitation of our MySQL connector, or whether it's a problem of the relation implementation in LB4. Either way, I consider this issue out of scope of this (initial) pull request.

packages/repository/src/datasource.ts
```diff
@@ -17,3 +17,11 @@ export interface DataSource {
   // tslint:disable-next-line:no-any
   [property: string]: any; // Other properties that vary by connectors
 }
+
+export interface SchemaMigrationOptions extends Options {
+  /**
+   * When set to true, schema migration will drop existing tables and recreate
+   * them from scratch, removing any existing data along the way.
+   */
+  dropExistingSchema?: boolean;
```
raymondfeng: Maybe instead of `boolean`, we should have an `enum`:

```ts
policyForExistingSchema: 'drop' | 'alter' | 'ignore' | 'fail';
```

raymondfeng:
> The tables were created correctly, but unfortunately the foreign key constraint was not set up. I don't know whether this is a limitation of our MySQL connector, or whether it's a problem of the relation implementation in LB4.

I think it's a limitation of LB3 `automigrate/autoupdate`.

packages/repository/src/datasource.ts
bajtos: I am not sure what `ignore` would mean, but other than that I like the idea to use an enum that's easier to extend in the future.

The current juggler implementation supports `drop` and `alter` only, so that's what I am going to use in this pull request.

I find `policyForExistingSchema` too long, how about shorter `existingSchema`?

bajtos:
> I think it's a limitation of LB3 automigrate/autoupdate.

Good to know, so I am not going to worry about this for now. I think the most important thing is that the current high-level design does not prevent the connector to setup referential integrity and possibly other constraints in the future.

raymondfeng:
> What scenario do you @raymondfeng have in mind for choosing a subset of tables only? I think this filter is easy to add later as an incremental improvement of my MVP solution, thus I am proposing to leave it out of this initial pull request.

Developers want to control what models that need to be migrated. For example, our UI allows people to select a list of models and click `migrate`.

In real world, some models are created newly while others are discovered from existing DB schema. We don't want run `migrate` for such models discovered.

We can simply add `models` to the `options` and pass it down to datasource.

bajtos:
> Developers want to control what models that need to be migrated. For example, our UI allows people to select a list of models and click migrate.
> 
> In real world, some models are created newly while others are discovered from existing DB schema. We don't want run migrate for such models discovered.
> 
> We can simply add models to the options and pass it down to datasource.

I feel it's a bit of a scope creep here, but I am ok to add models to the options in the TS/JS API as you suggested :+1:

bajtos:
@raymondfeng I added support for `options.models` (defaulting to migrate all models) and reworked a boolean option `dropExistingSchema` into `existingSchema` accepting a string enum.

LGTY now?

---

I would also like to hear opinions from @strongloop/loopback-maintainers on the following topic I mentioned earlier:

Right now, the pull request contains what I consider as a minimal viable implementation/improvement. Besides the changes proposed, I would like to discuss few more improvements we may or may not want to do, either as part of this pull request or in follow-ups:

- Add `src/migrate.ts` to all example repositories. I think it would be very useful to have automigration available in our TodoList example too. /cc @b-admike 
- Introduce an npm/package.json script, e.g. `npm run update-schema`, as an alias for `node dist/src/migrate`.
- Modify the CLI template to include `src/migrate.ts` and `npm run update-schema` in all new projects.

packages/repository/src/datasource.ts
raymondfeng: We can start with `'drop' | 'alter'`. I'm fine with `existingSchema`.

dhmlau:
IMHO, it is a good idea to have a more "automatic" way to create the database tables for users, especially for new users.    

Another question or discussion point is that what we want for the default behaviour for someone creating a LB app for real use.  Maybe it's just me, I might just want the `migrate` script to be run for the first time to create the db definitions or whenever that's needed rather than on every application start.

docs/site/Database-migrations.md
```diff
@@ -89,6 +93,35 @@ your database by running `node dist/src/migrate` and rebuild it from scratch by
...
+In some scenarios, the application may need to define additional schema
+constraints or seed the databse with predefined model instances. This can be
+achieved by overriding the `migrateSchema` method provided by the mixin.
+
+The example below shows how to do so in our Todo example application.
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+export class TodoListApplication extends BootMixin(
```
b-admike: nitpick: maybe add a comment for imports on top?

examples/todo/src/migrate.ts
```diff
@@ -6,12 +6,20 @@
...
+    dropExistingSchema ? 'rebuild' : 'update',
+  );
 
   const app = new TodoListApplication();
   await app.boot();
-  await app.migrateSchema({rebuild});
+  await app.migrateSchema({dropExistingSchema});
+
+  // Connectors usually keep a pool of opened connections,
+  // this keeps the process running even after all works is done.
```
b-admike: nitpick: `works` -> `work`

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -184,17 +180,30 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
...
+        : 'autoupdate';
 
-      for (const b of repoBindings) {
-        const repo = await this.get(b.key);
+      // Instantiate all repositories to ensure models are registered & attached
+      // to their datasources
+      const repoBindings: Readonly<Binding<unknown>>[] = this.findByTag(
+        'repository',
+      );
+      await Promise.all(repoBindings.map(b => this.get(b.key)));
```
b-admike: I'm wondering how this will affect performance if we're loading up all the repository instances? Should we do some sort of filtering now that we support select models in the `migrateSchema` function?

bajtos: Great question!

We are creating instances of repository classes. This is a lightweight task - a repository instance is created and a new PersistedModel class is registered with the appropriate datasource. I am not anticipating any performance issues. 

Migration is a one-time action so even if its performance is not optimal, the affect on user applications should be minimal.

To allow datasource migration to define foreign key constraints, it may be necessary to setup related models even if they are not migrated. For example, if we are migrating only an Order model which has a belongs to relation to User, then we should instantiate UserRepository too, so that the datasource/connector knows about both sides of the relation.

Finally, I am not aware of a reasonably easy & sufficiently robust way of how to filter repositories based on model names, because the repository binding does not provide any information about the model it is providing data access to.

b-admike: Thank you for clarifying 💯!

Issue #1884, opened 2018-10-19
[EPIC] Support messaging/eventing style APIs


raymondfeng:
Currently LoopBack supports RPC style APIs such as REST or gRPC. We would like to extend that with messaging/eventing capabilities.

# Use cases
- Expose endpoints to accept a stream of messages/events and/or return a stream of messages/events (such as SSE)
- Connect to backend resources (such as Kafka or MQTT Broker) to produce/consume (including pub/sub) messages/events
- Define business functions to produce/consume a stream of messages/events (synchronously or asynchronously)
- Create serverless actions (such as IBM Cloud Functions and AWS Lambda) with LoopBack 4 as event handlers
- Observe events and react on them (hooks? TBD)
  - https://github.com/ReactiveX/rxjs?
  - https://rxjs-dev.firebaseapp.com/guide/observable
- Make it easy to implement web hooks

# Protocols
- WebSocket
- HTTP/2 with SSE
- MQTT

# Mapping to LoopBack 4 concepts
- `Servers`: listen on WebSocket, MQTT, and HTTP/2 protocols
-  `Connectors`: subscribe to an event source or message broker or publish messages to backend systems
- `Controllers`: create methods (with decorators) to produce/consume/process a stream of events. 
- `Events`: define models representing various types of events

# Examples

```ts
app.server(WebSocketServer, config);

app.datasource({name: 'kafka', ...};

export class MoveEvent {
  gameId: string;
  sequence: number;
  player: string;
  shape: 'o' | 'x';
  move: [number, number];
}

export class GameStatus {
  id: string;
  board: ('o' | 'x' )[][];
  winner: string;
  status: string;
}

export class TicTakToeController {
  async createGame(player1: string, player2: string): number {}

  // Should we support event by event, a stream of events, or both
  handleMoves(move: AsyncIterator<MoveEvent>): AsyncIterator<GameStatus> {}
 
  async handleMove(move: MoveEvent): GameStatus {}
}

export class MyEventController {
  @subscribe('order.created')
  async onNewOrders(@event('order.created') orderCreatedEvent: OrderCreated) {}

  @publish('invoice.created')
  generateInvoices(): AsyncIterator<InvoiceCreated> {}
}
```

mikaeljensen:
I think this is going to be very important for the future of LB4.

[...]

PR #2065, opened 2018-11-21
feat(http-server): add http/2 support based on spdy

raymondfeng:
Adds http/2 support using https://github.com/spdy-http2/node-spdy. Even Node.js core now has `http2` support, the `spdy` module is express-compatible.

## Checklist

### DIFF
--- a/packages/http-server/src/http-server.ts
+++ b/packages/http-server/src/http-server.ts
@@ -3,12 +3,12 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {IncomingMessage, ServerResponse} from 'http';
 import * as http from 'http';
+import {IncomingMessage, ServerResponse} from 'http';
 import * as https from 'https';
 import {AddressInfo} from 'net';
-
 import * as pEvent from 'p-event';
+import * as spdy from 'spdy';
 
 export type RequestListener = (
   req: IncomingMessage,
@@ -47,12 +47,19 @@ export interface HttpsOptions extends ListenerOptions, https.ServerOptions {
 }
 
 /**
+ * HTTP/2 server options based on `spdy`
+ */
+export interface Http2Options extends ListenerOptions, spdy.ServerOptions {
+  protocol: 'http2' | 'https';
+}
+
+/**
  * Possible server options
  *
  * @export
  * @type HttpServerOptions
  */
-export type HttpServerOptions = HttpOptions | HttpsOptions;
+export type HttpServerOptions = HttpOptions | HttpsOptions | Http2Options;
 
 /**
  * Supported protocols
@@ -60,7 +67,7 @@ export type HttpServerOptions = HttpOptions | HttpsOptions;
  * @export
  * @type HttpProtocol
  */
-export type HttpProtocol = 'http' | 'https'; // Will be extended to `http2` in the future
+export type HttpProtocol = 'http' | 'https' | 'http2';
 
 /**
  * HTTP / HTTPS server used by LoopBack's RestServer
@@ -84,14 +91,22 @@ export class HttpServer {
    */
   constructor(
     requestListener: RequestListener,
-    serverOptions?: HttpServerOptions,
+    serverOptions: HttpServerOptions = {},
   ) {
     this.requestListener = requestListener;
     this.serverOptions = serverOptions;
     this._port = serverOptions ? serverOptions.port || 0 : 0;
     this._host = serverOptions ? serverOptions.host : undefined;
     this._protocol = serverOptions ? serverOptions.protocol || 'http' : 'http';
-    if (this._protocol === 'https') {
+    if (
+      this._protocol === 'http2' ||
+      (this._protocol === 'https' && serverOptions.hasOwnProperty('spdy'))
+    ) {
+      this.server = spdy.createServer(
+        this.serverOptions as spdy.ServerOptions,
+        this.requestListener,
+      );
+    } else if (this._protocol === 'https') {
       this.server = https.createServer(
         this.serverOptions as https.ServerOptions,
         this.requestListener,
@@ -116,8 +131,14 @@ export class HttpServer {
    */
   public async stop() {
     if (!this.server) return;
-    this.server.close();
-    await pEvent(this.server, 'close');
+    if (!this.server.listening) return;
+    await new Promise<void>((resolve, reject) => {
+      this.server.close((err: unknown) => {
+        if (!err) resolve();
+        else reject(err);
+      });
+    });
+
     this._listening = false;
   }
 
@@ -153,7 +174,8 @@ export class HttpServer {
     } else if (host === '0.0.0.0') {
       host = '127.0.0.1';
     }
-    return `${this._protocol}://${host}:${this.port}`;
+    const protocol = this._protocol === 'http2' ? 'https' : this._protocol;
+    return `${protocol}://${host}:${this.port}`;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


bajtos:
+1 for adding support for HTTP/2.

Personally, I'd prefer to use the built-in HTTP/2 implementation provided by Node.js core, I expect it will have better integration with tooling like traces, monitoring, debugging, etc. and also get security issues fixed more quickly.

The built-in `http2` module is marked as Stable in all Node.js releases we support (8.x, 10.x), see [the docs](https://nodejs.org/docs/latest-v8.x/api/http2.html#http2_http_2).

@raymondfeng What are your arguments for using a user-land module instead?

packages/http-server/src/http-server.ts
```diff
@@ -84,14 +91,22 @@ export class HttpServer {
...
   constructor(
     requestListener: RequestListener,
-    serverOptions?: HttpServerOptions,
+    serverOptions: HttpServerOptions = {},
   ) {
     this.requestListener = requestListener;
     this.serverOptions = serverOptions;
     this._port = serverOptions ? serverOptions.port || 0 : 0;
     this._host = serverOptions ? serverOptions.host : undefined;
     this._protocol = serverOptions ? serverOptions.protocol || 'http' : 'http';
```
bajtos: Since `serverOptions` are always defined, the checks `serverOptions ?:` can be removed now. (This applies to all three lines above.)

bajtos:
Also note that the built-in `http2` modules supports HTTP/2 over uncencrypted connection too, which is useful for local testing.

bajtos:
See also http2's [Compatibility API](https://nodejs.org/docs/latest-v8.x/api/http2.html#http2_compatibility_api)

bajtos:
@raymondfeng I'd like to better understand the motivations behind this change and the goals you have in mind.

- What are the scenarios you would like to enable by this pull request? 
- HTTP/2 has a different programming offering a richer set of features compared to HTTP/1. What subset of these features are trying to provide by this pull request? I'd like us to be explicit about which parts of HTTP/2 we do and do not support.
- What is the priority of this work? Who is waiting for this feature?

In general, I am very reluctant to force node-spdy dependency on all LB4 users.

Also, because HttpServer is exposing the underlying HTTP/1 or HTTP/2 server as a public API, we must carefully consider ramifications of committing to node-spdy server API (as opposed to core `http2` APIs).

From what I was able to find about `http2` and `express` integration, express@4 is not compatible with http2 because it manipulates internals of the built-in request/response objects in an unsupported way. The upcoming express@5 version is aiming to fix that problem, unfortunately the release is probably months to years in the future.

On the other hand, from what I remember, we use only a small subset of Express functionality. I believe we are relying on two features only:
 - enhanced request/response objects (as expected by 3rd-party middleware handlers)
 - a basic router capable of executing middleware handlers in the order they were registered

If that's the case, then I am proposing to rework our internal implementation, drop full Express and replace it with our own composition of lower-level Express components (see https://github.com/pillarjs) from the upcoming version express@5. That way we can stay Express compatible and be able to leverage native `http2` module.

Let's discuss!

bajtos:
> HTTP/2 has a different programming offering a richer set of features compared to HTTP/1. What subset of these features are trying to provide by this pull request? I'd like us to be explicit about which parts of HTTP/2 we do and do not support.

Besides the obvious push-based approach for serving files, do we want to support ALPN-based support for falling back to HTTP/1.1? See https://nodejs.org/api/http2.html#http2_alpn_negotiation

bajtos:
@raymondfeng if your goal is to quickly add an experimental support for HTTP/2 based on  https://github.com/spdy-http2/node-spdy, then I am proposing to implement it as a new extension and modify `@loopback/rest` to allow app/extensions to provide a custom server factory (provider).

I am thinking about changing the following line:

https://github.com/strongloop/loopback-next/blob/b9570c8f9e6b5b9e57271605e8e132b5815a2198/packages/rest/src/rest.server.ts#L742

into something like this:

```ts
this._httpServer = this.httpServerFactory(this.requestHandler, serverOptions);
```

By default, `httpServerFactory` can be implemented via `new HttpServer()`, but users are allowed to provide a custom factory function via Dependency Injection.

```ts
app.bind(RestBindings.SERVER_FACTORY).to((handler, options) => {
  return new SpdyServer(
    this.serverOptions as spdy.ServerOptions,
    this.requestListener,
  );
});

class SpdyServer implements HttpServer {
 // use spdy.createServer() under the hood
 // provide HttpServer features like port/host/url, start & stop, etc.
}
```

Just an idea to kickstart the discussion.

raymondfeng:
I'm exploring the opportunity to leverage `HTTP/2 + SSE` to support https://github.com/strongloop/loopback-next/issues/1884. At this moment, I'm mostly trying to use `http/2` as a transparent drop-in enhancement to `http/https` without touching the http/2 specific programming model level features yet.

The primary motivtion to use `spdy` is its express compatibility.

bajtos:
> I'm exploring the opportunity to leverage HTTP/2 + SSE to support #1884. At this moment, I'm mostly trying to use http/2 as a transparent drop-in enhancement to http/https without touching the http/2 specific programming model level features yet.

Makes sense 👍 

I believe HTTP/2 is not required for SSE ([server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events)), SSE works with HTTP/1 too. For example, LB3 (strong-remoting) is using SSE to stream notification about model updates - see [http-context.js](https://github.com/strongloop/strong-remoting/blob/e10c5345402a4066ac9614b412276a3b2a8ed212/lib/http-context.js#L400-L431) for the implementation details, the [docs](https://loopback.io/doc/en/lb3/Realtime-server-sent-events.html) and the [example client](https://github.com/strongloop/angular-live-set).

Also as far as I understand HTTP/2, it does not really provide any benefits for server-sent events compared to HTTP/1. It's always the client who has to initiate the request. The server can speculatively send content for a request that will be requested soon (e.g. css/js assets for a HTML page), but it's still up to the client to request that content. I.e. there is no way for the HTTP/2 server to actively push a new event (a notification) to the client, SSE is using the same never-ending stream as HTTP/1.

I am trying to say that missing support for HTTP/2 should not be a blocker for your work on  #1884.

> The primary motivtion to use spdy is its express compatibility.

I understand that very well after my short research on express+HTTP/2 compatibility on GitHub. What a mess :(

I see `@loopback/rest` as a core component of LB4, one that will be used by most LB4 apps. I am very reluctant to force a new dependency on `node-spdy` to all LB4 users that are happy with HTTP/1 and don't ask for HTTP/2.  I strongly believe that a built-in support for HTTP/2 should be implemented on top of Node.js core module `http2`.

At the same time, I am perfectly fine to provide HTTP/2 support based on `node-spdy` via a new extension package. Because this package is versioned independently and used by a small fraction of our user base, we have more space for iterating and learning from mistakes.

Also by adding an extension point allowing applications/extensions to contribute custom HttpServer implementation, we make LoopBack even more future proof. When HTTP/3 comes closer, we can create another extension to run LoopBack on HTTP/3, even before there is a rock-solid HTTP/3 implementation available for general use.

raymondfeng:
Superseded by #2078

bajtos:
@raymondfeng FYI, a Node.js core maintainer is warning against using node-spdy.

https://twitter.com/addaleax/status/1068063898228727814
> Please do *not* use the `spdy` module on npm, if at all possible.
>
> Its code depends on Node's internals in an incredibly fragile way, and hasn't worked on Node 10 for most of its lifetime.
>
> SPDY itself is deprecated in favour of HTTP/2; modern browsers have dropped support.

PR #2070, opened 2018-11-23
docs: add note to create table in todo example

dhmlau:
When running the `todo` or `todo-list` example, if I choose to use a relational database as datasource instead of the in-memory database, I will need to create the table.  As a new user, it might not be trivial. 

In this PR, I've added a note to remind users to have the table created or follow the [database migration docs](https://loopback.io/doc/en/lb4/Database-migrations.html).

### DIFF
--- a/docs/site/todo-tutorial-datasource.md
+++ b/docs/site/todo-tutorial-datasource.md
@@ -67,6 +67,11 @@ Create a `data` folder in the applications root and add a new file called
 }
 ```
 
+{% include note.html content="If you are using a relational database as the
+datasource, don't forget to create the corresponding table or follow the
+[Database migration instruction](https://loopback.io/doc/en/lb4/Database-migrations.html) to get it created programmatically.
+" %}
+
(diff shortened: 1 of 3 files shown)


docs/site/todo-tutorial-datasource.md
```diff
@@ -67,6 +67,11 @@ Create a `data` folder in the applications root and add a new file called
 }
 ```
 
+> **NOTE**: If you are using a relational database as the datasource, don't
```
bajtos: Please use `{% include note.html ...%}`, see https://loopback.io/doc/en/contrib/alerts.html

docs/site/todo-tutorial-datasource.md
```diff
@@ -67,6 +67,13 @@ Create a `data` folder in the applications root and add a new file called
 }
 ```
 
+> **NOTE**:
```
bajtos: I think this should be removed, now that you are using `{% include note.html ... %}`

Issue #1866, opened 2018-10-17
How to use regular expressions in route path and how to make parameters optional


hbakhtiyor:
* how to use regular expression in route path?

for example like expressjs
```
app.get('/assets/:uuid([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})', function (req, res) {
}) 
```

* how to make route param is optional?

for example

`/assets/eed761cf-5418-53d3-9f67-c279f58ac875.png`
`/assets/eed761cf-5418-53d3-9f67-c279f58ac875`, `.format` is optional

```
import { get, param } from '@loopback/rest';

export class AssetController {
  @get('/assets/{uuid}.{format}')
  assets(@param.path.string('uuid') uuid: string,
             @param.path.string('format') format?: string): object {
      return {};
  }
}
```

hbakhtiyor:
cc @raymondfeng

hacksparrow:
@hbakhtiyor LB4 uses Express' [serve-static](https://www.npmjs.com/package/serve-static) middleware for serving static files, you can see how to see regular expression and other patterns for a path at http://expressjs.com/en/4x/api.html#path-examples.

hbakhtiyor:
i can't use like this in lb4, which valid in expressjs

```
@get('/assets/{uuid}([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})')
```

hacksparrow:
At the moment, static files can be served from a single mount point only. We don't have a controller interface yet. Having said that, you LB4 provides an Express-like method (`app.static()`) for setting up statics assets, here is an example:

```
  const app = new Lb4appApplication(options);

  const publicDir = path.join(__dirname, 'public');

  await app.boot();
  await app.start();

  app.static(/\/abc|\/xyz/, publicDir); // <-- regex path

  const url = app.restServer.url;
  console.log(`Server is running at ${url}`);
  console.log(`Try ${url}/ping`);

  return app;
```

hbakhtiyor:
i don't need exactly for static serve, not related to my questions

hacksparrow:
LB4 does not support regex based paths.

@bajtos @raymondfeng @strongloop/sq-lb-apex how about supporting regex-based paths?

marioestradarosa:
> how to use regular expression in route path?

[...]

PR #2077, opened 2018-11-26
fix(rest): allow `.` to be used in openapi path template

raymondfeng:
See https://github.com/strongloop/loopback-next/issues/1866

## Checklist

### DIFF
--- a/packages/rest/src/router/openapi-path.ts
+++ b/packages/rest/src/router/openapi-path.ts
@@ -34,7 +34,7 @@ export function validateApiPath(path: string = '/') {
     );
   }
 
-  const regexpPath = path.replace(POSSIBLE_VARNAME_PATTERN, ':$1');
+  const regexpPath = toExpressPath(path);
   tokens = pathToRegExp.parse(regexpPath);
   for (const token of tokens) {
     if (typeof token === 'string') continue;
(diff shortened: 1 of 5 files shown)


packages/rest/src/router/openapi-path.ts
```diff
@@ -62,5 +62,7 @@ export function getPathVariables(path: string) {
  * @param path OpenAPI path with optional variables as `{var}`
  */
 export function toExpressPath(path: string) {
-  return path.replace(POSSIBLE_VARNAME_PATTERN, ':$1');
+  // Convert to `.` to `\\.` so that path-to-regexp will treat it as the plain
```
jannyHou: I think you meant

> Convert `.` to `\\.`

packages/rest/src/router/openapi-path.ts
```diff
@@ -62,5 +62,7 @@ export function getPathVariables(path: string) {
  * @param path OpenAPI path with optional variables as `{var}`
  */
 export function toExpressPath(path: string) {
-  return path.replace(POSSIBLE_VARNAME_PATTERN, ':$1');
+  // Convert `.` to `\\.` so that path-to-regexp will treat it as the plain
+  // `.` character
+  return path.replace(POSSIBLE_VARNAME_PATTERN, ':$1').replace('.', '\\.');
```
jannyHou: just curious: why `.` is special here? I realized that we don't do the conversion for other special characters like `@`.

raymondfeng: `.` is special char for regexp.

PR #2078, opened 2018-11-26
feat(http-server): make http2 server creation extensible

raymondfeng:
Supercedes https://github.com/strongloop/loopback-next/pull/2065

## Checklist

### DIFF
--- a/packages/http-server/src/http-server.ts
+++ b/packages/http-server/src/http-server.ts
@@ -3,64 +3,17 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {IncomingMessage, ServerResponse} from 'http';
 import * as http from 'http';
 import * as https from 'https';
 import {AddressInfo} from 'net';
-
 import * as pEvent from 'p-event';
-
-export type RequestListener = (
-  req: IncomingMessage,
-  res: ServerResponse,
-) => void;
-
-/**
- * Basic HTTP server listener options
- *
- * @export
- * @interface ListenerOptions
- */
-export interface ListenerOptions {
-  host?: string;
-  port?: number;
-}
-
-/**
- * HTTP server options
- *
- * @export
- * @interface HttpOptions
- */
-export interface HttpOptions extends ListenerOptions {
-  protocol?: 'http';
-}
-
-/**
- * HTTPS server options
- *
- * @export
- * @interface HttpsOptions
- */
-export interface HttpsOptions extends ListenerOptions, https.ServerOptions {
-  protocol: 'https';
-}
-
-/**
- * Possible server options
- *
- * @export
- * @type HttpServerOptions
- */
-export type HttpServerOptions = HttpOptions | HttpsOptions;
-
-/**
- * Supported protocols
- *
- * @export
- * @type HttpProtocol
- */
-export type HttpProtocol = 'http' | 'https'; // Will be extended to `http2` in the future
+import {
+  HttpProtocol,
+  HttpServer,
+  HttpServerOptions,
+  RequestListener,
+  ProtocolServerFactory,
+} from './types';
 
 /**
  * HTTP / HTTPS server used by LoopBack's RestServer
@@ -68,15 +21,18 @@ export type HttpProtocol = 'http' | 'https'; // Will be extended to `http2` in t
  * @export
  * @class HttpServer
  */
-export class HttpServer {
+export class DefaultHttpServer implements HttpServer {
   private _port: number;
   private _host?: string;
   private _listening: boolean = false;
-  private _protocol: HttpProtocol;
+  protected _protocol: string;
+  private _urlScheme: string;
   private _address: AddressInfo;
-  private requestListener: RequestListener;
-  readonly server: http.Server | https.Server;
-  private serverOptions?: HttpServerOptions;
+  protected readonly requestListener: RequestListener;
+  protected _server: http.Server | https.Server;
+  protected readonly serverOptions: HttpServerOptions;
+
+  protected protocolServerFactories: ProtocolServerFactory[];
 
   /**
    * @param requestListener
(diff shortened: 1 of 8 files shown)


packages/http-server/src/http-server.ts
```diff
@@ -3,80 +3,32 @@
...
+} from './types';
 
 /**
  * HTTP / HTTPS server used by LoopBack's RestServer
  *
  * @export
  * @class HttpServer
  */
-export class HttpServer {
+export class DefaultHttpServer implements HttpServer {
```
bajtos: What's the impact of this change on possible users of `@loopback/http-server`?

For example, the following code creates a fully working HttpServer class in the current version (before your changes):

```ts
import * as turbo from 'turbo-http';

class TurboHttpServer extends HttpServer {
  constructor(
    serverOptions?: HttpServerOptions,
  ) {
    super(requestListener, serverOptions);
    this.server = turbo.createServer(requestListener);
  }
}
```

IMO, we should preserve backwards compatibility in our API. Otherwise a new major version must be published, plus we will have to keep maintaining the old version according to our LTS policy...

packages/http-server/src/types.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * HTTP/2 server options
+ *
+ * @export
+ * @interface Http2Options
+ */
+export interface Http2Options extends ListenerOptions {
+  protocol: 'http2';
```
bajtos: The built-in `http2` module support HTTP/2 protocol over plain unencrypted socket (similar to HTTP) or over TLS socket (similar to HTTPS). How do you envision encoding this configuration flag when there is a single protocol value `http2`?

packages/http-server/test/integration/http-server.integration.ts
```diff
@@ -184,18 +195,51 @@ describe('HttpServer (integration)', () => {
...
     // Windows does not support http://0.0.0.0:3000/
-    server = new HttpServer(dummyRequestHandler, {host: '0.0.0.0'});
+    server = givenHttpServer(dummyRequestHandler, {host: '0.0.0.0'});
     await server.start();
     expect(server.url).to.equal(`http://127.0.0.1:${server.port}`);
   });
 
+  it('supports HTTP/2 protocol with key and certificate files', async () => {
+    // For some reason, spdy fails with Node 11
+    if (+process.versions.node.split('.')[0] > 10) return;
```
bajtos: This early return will mark the test as passed, leading to false sense of security.

Please mark the test as skipped on Node.js 11.x. If we don't have a helper like "it.skipIf", then please consider creating one in `@loopback/testlab`.

```ts
it.skipIf(
  // For some reason, spdy fails with Node 11
  +process.versions.node.split('.')[0] > 10),
  'supports HTTP/2 protocol with key and certificate files',
  async () => {
    // ...
  }
);
```

packages/http-server/test/integration/http-server.integration.ts
```diff
@@ -184,18 +195,51 @@ describe('HttpServer (integration)', () => {
...
+    const serverOptions: Http2Options = Object.assign(
+      {
+        protocol: 'http2' as 'http2',
```
bajtos:
```ts
'http2' as 'http2'
```

This looks weird, are you sure it's needed?

packages/http-server/test/integration/http-server.integration.ts
```diff
@@ -184,18 +195,51 @@ describe('HttpServer (integration)', () => {
...
+        rejectUnauthorized: false,
+        spdy: {protocols: ['h2' as 'h2']},
```
bajtos: similarly here: `'h2' as 'h2'`?

packages/http-server/test/integration/http-server.integration.ts
```diff
@@ -184,18 +195,51 @@ describe('HttpServer (integration)', () => {
...
+        plain: false,
+        ssl: true,
+      },
+    }) as spdy.Agent;
+    const response = await httpsGetAsync(server.url, agent);
+    expect(response.statusCode).to.equal(200);
+    // We need to close the agent so that server.close() returns
+    // `@types/spdy@3.x` is not fully compatible with `spdy@4.0.0`
+    // tslint:disable-next-line:no-any
+    (agent as any).close();
```
bajtos: Please close the agent before making assertion.

In your current version, when `expect(response.statusCode)` throw an assertion error, then the agent is never closed.

packages/http-server/src/http-server.ts
```diff
@@ -163,10 +150,26 @@ export class HttpServer {
...
    */
   public get address(): AddressInfo | undefined {
     return this._listening ? this._address : undefined;
   }
 }
+
+/**
+ * Default implementation of HttpServerFactory
+ */
+export class DefaultHttpServerFactory {
```
bajtos: Do we have to use a class here? Can we use a plain function instead?

```ts
export function defaultHttpServerFactory(
  serverOptions?: HttpServerOptions,
): HttpServer {
  return new DefaultHttpServer(requestListener, serverOptions);
}
```

See also [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).

packages/http-server/src/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  readonly address: AddressInfo | undefined;
+
+  start(): Promise<void>;
+  stop(): Promise<void>;
+}
+
+/**
+ * A factory to create http servers
+ */
+export interface HttpServerFactory {
```
bajtos: Can we use a plain function instead of an interface please?

packages/testlab/src/request.ts
```diff
@@ -22,10 +22,15 @@ export function httpGetAsync(urlString: string): Promise<IncomingMessage> {
...
-export function httpsGetAsync(urlString: string): Promise<IncomingMessage> {
-  const agent = new https.Agent({
-    rejectUnauthorized: false,
-  });
+export function httpsGetAsync(
+  urlString: string,
+  agent?: https.Agent,
+): Promise<IncomingMessage> {
+  agent =
+    agent ||
```
bajtos: I find the formatting of this code block rather weird. Have you considered the following alternative?

```ts
const defaultAgent = new https.Agent({
 rejectUnauthorized: false,
});

export function httpsGetAsync(
   urlString: string,
   agent: https.Agent = defaultAgent,
): Promise<IncomingMessage> {
  // agent is always set now
}
```

packages/http-server/src/http-server.ts
raymondfeng: We can release `@loopback/http-server@2.0.0` if we have to break the compatibility due to the nature of this work.

packages/http-server/test/integration/http-server.integration.ts
raymondfeng: If we don't cast `'http2'` to `'http2'`, TS complains as it infers the type of `http2` as `string`, which is not compatible with `'http' | 'https' | 'http2'`.

packages/http-server/src/types.ts
raymondfeng: I wanted to make it possible to use dependency injection for the implementation.

packages/http-server/src/http-server.ts
bajtos: > We can release @loopback/http-server@2.0.0 if we have to break the compatibility due to the nature of this work.

Fair enough. In that case:

- The commit message in this PR should clearly describe the breaking changes so that our release tooling can pick it up. Ideally, there should be a short migration guide too.
- We need to update http-server's README with a table listing supported major versions (LTS terms). See https://github.com/strongloop/loopback#module-long-term-support-policy for inspiration. This should be done *before* 2.0.0 release, so that the new content is displayed on https://www.npmjs.com/package/@loopback/http-server

packages/http-server/test/integration/http-server.integration.ts
bajtos: Weird.

Can you add a comment to explain to readers please? (Similarly for 'h2' below.)

packages/http-server/src/types.ts
bajtos: Makes sense. Can you enhance the TSDoc comment to explain why we choose an interface instead of a function?

bajtos:
One more reason why I don't like HttpServerFactory is that HttpServer itself is implementing as a factory that accepts `protocol` and creates a different underlying server instance based on the configured value (`http.createServer`, `https.createServer`, etc.)

packages/http-server/src/http-server.ts
bajtos: See also https://github.com/strongloop/loopback-next/pull/2103

packages/http-server/src/types.ts
```diff
@@ -0,0 +1,110 @@
...
+export interface ProtocolServer {
+  server: http.Server | https.Server;
+  /**
+   * Scheme for the URL
+   */
+  urlScheme: string;
+}
+
+export interface ProtocolServerFactory {
+  supports(protocol: string, serverOptions: HttpServerOptions): boolean;
```
bajtos: Interesting. The good part about this design is that it follows what we have already in place for body parsers. However, I feel it's unnecessarily complicated for the case of http servers.

For long term, I think we need to provide `http2` support based on Node.js core module `http2`. From what I read, `spdy` does not work on Node.js 10+. Once this is done, most users should not need to change the way how HTTP servers are created at all.

raeef-refai:
any news about this feature?

vaibhavkumar-sf:
Hey, we need http2 support in LB4,

Issue #2088, opened 2018-11-27
RestBindings.Http.Request.query property always 'undefined'


larsm11:
I have been using LB4 since July, and until recently I've been able to grab the query parameters from the RestBindings.Http.Request.query using the following injection statement in my controller constructor

`@inject(RestBindings.Http.REQUEST) private req: Request,`

Followed by `this.req.query.<property>` somewhere in my code.

This no longer works, even if I create a brand new application using the latest **lb4 app** and try to expose the query via the default Ping-controller. When trying the following section of code
```
      query: 'Query response: ' + this.req.query.start,
      greeting: 'Hello from LoopBack',
      date: new Date(),
      url: this.req.url,
      headers: Object.assign({}, this.req.headers),
```
using this request **localhost:3000/ping?start=2018-08-25&end=2018-09-09&user=larsm** 
reponse now returns 'undefined' (rather than returning the query parameters):

```
{
  "query": "Query response: undefined",
  "greeting": "Hello from LoopBack",
  "date": "2018-11-27T23:21:53.142Z",
  "url": "/ping?start=2018-08-25&end=2018-09-09&user=larsm",
  "headers": {
    "host": "localhost:3000",
    "connection": "keep-alive",
    "cache-control": "max-age=0",
    "upgrade-insecure-requests": "1",
    "user-agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36",
    "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8",
    "accept-encoding": "gzip, deflate, br",
    "accept-language": "nb-NO,nb;q=0.9,en-GB;q=0.8,en;q=0.7,no;q=0.6,nn;q=0.5,en-US;q=0.4"
  }
}

```

raymondfeng:
Fixed by https://github.com/strongloop/loopback-next/pull/2089

PR #2089, opened 2018-11-28
fix(rest): parse query string even when there is no rest query param

raymondfeng:
See https://github.com/strongloop/loopback-next/issues/2088

Now we use Express body parsers instead of `body` module. I think there was a side effect of `body` module that parses `request.query` beyond Express.

## Checklist


packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -52,6 +53,17 @@ describe('RestServer (integration)', () => {
     expect(server.url).to.be.undefined();
   });
 
+  it('parses query without decorated rest query params', async () => {
+    // See https://github.com/strongloop/loopback-next/issues/2088
+    const server = await givenAServer({rest: {port: 0, host: '127.0.0.1'}});
+    server.handler(queryRequestHandler);
```
bajtos: Please inline the handler in this test case.

I find it difficult to understand what's going on in this test case when the query handler is defined elsewhere and the name does not explain what is the handler actually doing.

Alternatively, find a better name for this handler function, e.g. `respondWithParsedQueryObject`.

bajtos:
-1 from me, I have intentionally disabled query parsing when no query arguments are described in the operation spec.

@larsm11 @raymondfeng Let's discuss.

What is the use case for having query arguments not documented in the OpenAPI spec? I find the  example provided in #2088 as too artificial. IMO, a real-world endpoint will want `start` and `end` query arguments described in the documentation so that API consumers are aware of them, client generators like swagger-codegen can include them in the client API, etc.

raymondfeng:
> -1 from me, I have intentionally disabled query parsing when no query arguments are described in the operation spec.

I think disabling query parsing is over-restrictive. Parsing query string is not a heavy operation. There are valid use cases that we want to have query string parsed regardless of being referenced in a controller method argument.

1. There are non-application specific query params, such as:
  - `access_token`.
  - https://github.com/expressjs/method-override
  - `_format` to override/specify `Accept` header

2. We allow routes beyond `ControllerRoute`

bajtos:
There is also the question of consistency. So far, LB4 applications were usually parsing the query by calling `qs` with default arguments, see 
https://github.com/strongloop/loopback-next/blob/8156f9d32aac3a0461b617a01cde9c319a1cceca/packages/rest/src/parser.ts#L122

With your change in place, this code will be effectively never called, because the query will be always parsed by the Express layer.

If we agree that the query should be parsed inside Express, then please clean up the code in `packages/rest/src/parser.ts` as part of that change. I think you will need to update few tests as well.

> There are non-application specific query params, such as access_token, ...
> We allow routes beyond ControllerRoute

Agreed.

My main objection is that all of these query parameters should be included in the OpenAPI spec for documentation purposes and for consumption by client generators.

Once these params are documented, we will automatically parse the query as part of argument processing.

On the other hand, I see that the current implementation does not work when the parsed query needs to be accessed before the parameter parser did it job, e.g. from a middleware or a custom sequence action called early in the sequence flow.

I guess that means we do need to change the behavior as you are proposing here.

raymondfeng:
> If we agree that the query should be parsed inside Express, then please clean up the code in packages/rest/src/parser.ts as part of that change. I think you will need to update few tests as well.

I switch to `express.query` to parse query string for consistency.

raymondfeng:
@bajtos PTAL

packages/rest/src/parser.ts
```diff
@@ -106,22 +103,23 @@ function getParamFromRequest(
...
     // TODO(jannyhou) to support `cookie`,
     // see issue https://github.com/strongloop/loopback-next/issues/997
     default:
       throw RestHttpErrors.invalidParamLocation(spec.in);
   }
 }
 
+/**
+ * This method is mostly used for unit testing where Express is not set up
+ * to parse query string into `request.query` object
```
bajtos: IIRC, most of our tests are using Shot-based mock to build the request/response objects.

https://github.com/strongloop/loopback-next/blob/cb57a4211f098fce68cfa5e534bd2df3d84eb64f/packages/testlab/src/shot.ts#L108-L109

Can you please update the code building the mock to parse the query string the same way as Express (LoopBack4) does, and then remove `ensureRequestQueryWasParsed` completely?

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -52,6 +53,28 @@ describe('RestServer (integration)', () => {
     expect(server.url).to.be.undefined();
   });
 
+  it('parses query without decorated rest query params', async () => {
+    // This handler responds with the query object (which is expected to
+    // be parsed by Express)
+    function requestWithQueryHandler(handler: {
```
bajtos: The first argument is context, not handler, and we have a type defined for that: `RequestContext`. 

https://github.com/strongloop/loopback-next/blob/80ba08145df8b7620ea2639af2b6ff67c513d026/packages/rest/src/rest.server.ts#L707

https://github.com/strongloop/loopback-next/blob/80ba08145df8b7620ea2639af2b6ff67c513d026/packages/rest/src/sequence.ts#L14-L21

I am proposing to simplify this part as follows

```ts
function requestWithQueryHandler({request, response}: RequestContext) {
  response.json(request.query);
  response.end();
}
```

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -52,6 +53,28 @@ describe('RestServer (integration)', () => {
...
+      request: Request;
+      response: Response;
+    }) {
+      const {request, response} = handler;
+      response.json(request.query);
+      response.end();
+    }
+
+    // See https://github.com/strongloop/loopback-next/issues/2088
+    const server = await givenAServer({rest: {port: 0, host: '127.0.0.1'}});
```
bajtos: Please use `givenHttpServerConfig`, it will take care of setting the correct host & port for our CI.

packages/rest/src/parser.ts
raymondfeng: Fixed

packages/rest/test/integration/rest.server.integration.ts
raymondfeng: Fixed

raymondfeng:
@bajtos Good feedback! PTAL.

packages/rest/test/integration/rest.server.integration.ts
```diff
@@ -702,6 +688,8 @@ paths:
   });
 
   async function givenAServer(options?: {rest: RestServerConfig}) {
+    options = options || {rest: {port: 0}};
```
bajtos: Please note that `givenHttpServerConfig` sets the port to 0 by default. I think the following simpler version should work well too:

```ts
async function givenAServer(options: {rest?: RestServerConfig} = {}) {
   options.rest = givenHttpServerConfig(options.rest);
   // ...
}
```

PR #2094, opened 2018-11-29
feat: scaffold DB migration script for new app projects

bajtos:
As a follow-up for https://github.com/strongloop/loopback-next/pull/2059:

- Add a template of `migrate.ts` to `lb4 app` generator. All new app project will be created with `src/migrate.ts` now.

- Further extend `lb4 app` generator to include the following npm/package script:

    ```
    "migrate": "node ./dist/src/migrate"
    ```

- Apply these changes to `examples/todo` and `examples/todo-list`

- Update `Database migrations` documentation page accordingly.

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/src/migrate.ts.ejs
+++ b/packages/cli/generators/app/templates/src/migrate.ts.ejs
@@ -1,0 +1,20 @@
+import {<%= project.applicationName %>} from './application';
+
+export async function migrate(args: string[]) {
+  const existingSchema = args.includes('--rebuild') ? 'drop' : 'alter';
+  console.log('Migrating schemas (%s existing schema)', existingSchema);
+
+  const app = new <%= project.applicationName %>();
+  await app.boot();
+  await app.migrateSchema({existingSchema});
+
+  // Connectors usually keep a pool of opened connections,
+  // this keeps the process running even after all work is done.
+  // We need to exit explicitly.
+  process.exit(0);
+}
+
+migrate(process.argv).catch(err => {
+  console.error('Cannot migrate database schema', err);
+  process.exit(1);
+});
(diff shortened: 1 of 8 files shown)


packages/cli/generators/app/templates/src/migrate.ts.ejs
```diff
@@ -0,0 +1,20 @@
...
+  console.log('Migrating schemas (%s existing schema)', existingSchema);
+
+  const app = new <%= project.applicationName %>();
+  await app.boot();
+  await app.migrateSchema({existingSchema});
+
+  // Connectors usually keep a pool of opened connections,
+  // this keeps the process running even after all work is done.
+  // We need to exit explicitly.
+  process.exit(0);
```
raymondfeng: On a side note, we probably should hook up `dataSource.disconnect` with `app.stop`.

jannyHou:
Do we have any plan to update other example repos like `todo-list`?

packages/cli/generators/app/templates/src/migrate.ts.ejs
bajtos: That's a great idea! Should we also call `dataSource.connect()` from `app.start`?

Is this something to fix first, before we change the project template?

bajtos:
> Do we have any plan to update other example repos like todo-list?

- I am adding `src/migrate.ts` and `npm run migrate` to `examples/todo-list` as part of this PR.
- `examples/todo` already contains `src/migrate.ts`, I added it as part of the previous PR. In this PR, I am adding `npm run migrate` script.

Did I miss anything?

packages/cli/generators/app/templates/src/migrate.ts.ejs
bajtos: Hmm, migrate does not call start/stop, only boot.

Should I change this block from `process.exit` to iterate over all registered datasources and call `disconnect` on them? Maybe do that as part of `app.migrate()`?

raymondfeng: I'm fine to use `process.exit()` for this PR since connectors are `started/connected` on demand even `app.start` is not involved. We can revisit it after the life cycle feature is ready.

PR #2096, opened 2018-11-29
fix(cli): replace PATCH method by missing PUT in rest controller template

ludohenin:
Looks like there's a mistake in the rest controller template.

## Checklist


raymondfeng:
@ludohenin Thanks for the patch.

We have two methods in the repository:

- updateById (PATCH)
- replaceById (PUT)

Can you adjust your patch accordingly?

ludohenin:
@raymondfeng sure I'll do that.

Before pushing the modifications, I have some questions:

1. does it then make sense to keep the PATCH method for a single object update `/my-model/{id}` (here I simply replaced PATCH by PUT)?
2. Regarding response code, I replace by `204` by `200`. Is `204` actually more appropriate ?
3. Changing the response code in the method definition still produces a `204` response from the server, is that the intended behavior ?

ludohenin:
Build is not passing due to `tsc` [errors ](https://travis-ci.org/strongloop/loopback-next/jobs/461479951#L153-L161)not related to the proposed changes. No idea how to fix that.

bajtos:
@ludohenin thank you for the pull request and a great discussion!

> does it then make sense to keep the PATCH method for a single object update /my-route/{id} (here I simply replaced PATCH by PUT)?

Yes please. Patch is making a partial update, e.g. you can update a name without touching any other properties. Replace is replacing all properties, it's useful when you want to set certain properties to `undefined`. Patch cannot set properties to `undefined`.

> Regarding response code, I replaced 204 by 200. Is 204 actually more appropriate ?

Please check out what's the actual behavior of the endpoints (what does the repository method return) and use the appropriate status code. 204 is used when the operation does not return anything (`Promise<void>`).

> Changing the response code in the method definition still produces a 204 response from the server, is that the intended behavior ?

I don't fully understand your question, but I think the answer is yes. The `responses` section is serving as a documentation only, the actual response is driven by what your code returns.

ludohenin:
@bajtos I checked `replaceById` and `updateById` method of the repository and both return `Promise<void>` so kept 204 as the response code for the method documentation (that was actually my question 3).

Should be OK, let me know otherwise.

ludohenin:
Sorry I messed up with git, will fix that.

ludohenin:
@raymondfeng I've added an integration test for the added controller method. I'm not quite sure what to do with the unit test (https://github.com/strongloop/loopback-next/blob/master/packages/cli/test/unit/openapi/controller-spec.unit.js), don't see anything about the patch method here as well.

PR #2098, opened 2018-11-29
chore(build): upgrade to TypeScript 3.2.1

raymondfeng:
See https://blogs.msdn.microsoft.com/typescript/2018/11/29/announcing-typescript-3-2/

## Checklist

### DIFF
--- a/packages/repository/src/query.ts
+++ b/packages/repository/src/query.ts
@@ -103,9 +103,10 @@ export type Condition<MT extends object> = {
  * }
  * ```
  */
-export type Where<MT extends object = AnyObject> = Condition<MT> &
-  Partial<AndClause<MT>> &
-  Partial<OrClause<MT>>;
+export type Where<MT extends object = AnyObject> =
+  | Condition<MT>
+  | AndClause<MT>
+  | OrClause<MT>;
 
 /**
  * And clause
@@ -249,8 +250,7 @@ export class WhereBuilder<MT extends object = AnyObject> {
       if (k in this.where) {
         // Found conflicting keys, create an `and` operator to join the existing
         // conditions with the new one
-        const where: Where<MT> = {};
-        where.and = [this.where, w];
+        const where = {and: [this.where, w]};
         this.where = where;
         return this;
       }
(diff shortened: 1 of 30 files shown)


packages/repository/src/query.ts
```diff
@@ -265,15 +265,7 @@ export class WhereBuilder<MT extends object = AnyObject> {
...
-    const w: Where<MT> = {};
-    if ('and' in clause) {
-      w.and = clause.and;
-    } else if ('or' in clause) {
-      w.or = clause.or;
-    } else {
-      Object.assign(w, clause);
-    }
-    return w;
+    return clause;
```
bajtos: Hmm, are you sure this change is preserving current behavior?

A test case to consider:

```ts
cast({and: [{x: 1}, {y: 2}], z: 3})
```

Before your change, cast returns

```
{and: [{x: 1}, {y: 2}]}
```

After your change, cast returns

```
{and: [{x: 1}, {y: 2}], z: 3}
```

I am typing these examples from my head based on how I understand the code. I may be wrong.

@raymondfeng please take a second look at this change.

If the new implementation of `cast` is just an identity function (returning the input without any change), then I think we should mark `cast` as deprecated (at minimum in TSDocs; ideally also at runtime using [depd](https://www.npmjs.com/package/depd)) and remove it in the next major version.

This does not necessarily has to happen in this pull request, but it would be great to make those changes soon, otherwise we are increasing our technical debt.

Thoughts?

packages/repository/src/query.ts
```diff
@@ -480,13 +472,14 @@ export class FilterBuilder<MT extends object = AnyObject> {
     if (!this.filter.fields) {
       this.filter.fields = {};
     }
+    const _fields = this.filter.fields;
```
bajtos: Why underscore? Can we use `fields` or perhaps `allFields`/`filterFields` to better distinguish from `field` name used later?

bajtos:
Nice!

I am proposing to make the following additional changes while upgrading to TS 3.2:

1. Enable `strictBindCallApply` ([docs](https://blogs.msdn.microsoft.com/typescript/2018/11/29/announcing-typescript-3-2/#strictbindcallapply))
2. Leverage _Configuration inheritance via node_modules packages_ ([docs](https://blogs.msdn.microsoft.com/typescript/2018/11/29/announcing-typescript-3-2/#configuration-inheritance-via-node_modules-packages)) to simplify `tsconfig.build.json` in our packages & examples, see e.g. [packages/core](https://github.com/strongloop/loopback-next/blob/master/packages/core/tsconfig.build.json#L3) and [examples/todo](https://github.com/strongloop/loopback-next/blob/ad0229b6711184da5a00c66ddeba585afe89eb79/examples/todo/tsconfig.build.json#L3). 

    I think we should use the following form everywhere:

    ```
    "extends": "@loopback/build/config/tsconfig.common.json",
    ```

PR #2109, opened 2018-12-03
RFC: helpful error message for deferred bindings [WIP]

bajtos:
When using our authentication package, many users try to inject the current user directly via `@inject()`. Because this injection is resolved before the sequence is executed, and the current-user value is set only inside the sequence, such attempt fails with an error that does not provide any guide on how to fix it.

In this pull request, I am introducing a new kind of bound value: deferred. It is serving as a placeholder for bindings that will be set later via `@inject.setter`. When users try to access such binding too early, we can provide a helpful error message pointing them to `@inject.getter`.

Such improvement is relying on the good will of extension authors to create a deferred binding for a value they will set later using `@inject.setter()`. I think that's too much to ask and we should provide a way to enforce this behavior.

The second part if this pull request is modification of `@inject.setter` to report a warning when the binding key is now know (was not registered as deferred). Originally, I want to `throw new Error`, but that would break backwards compatibility (which I don't want to right now).

**TODO**

 - [ ] tests
 - [ ] optional injection - add a test, decide on the expected behavior
 - [ ] remove warnings triggered by existing code (e.g. tests)
 - [ ] documentation

@strongloop/loopback-next what do you think? If you have encountered  _Error: The key authentication.currentUser was not bound to any value._, would you find it useful if the error message told you to use `@inject.getter()`?

## Checklist

### DIFF
--- a/packages/authentication/src/authentication.component.ts
+++ b/packages/authentication/src/authentication.component.ts
@@ -4,11 +4,12 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {AuthenticationBindings} from './keys';
-import {Component, ProviderMap} from '@loopback/core';
+import {Component, ProviderMap, Binding} from '@loopback/core';
 import {AuthenticateActionProvider, AuthMetadataProvider} from './providers';
 
 export class AuthenticationComponent implements Component {
   providers?: ProviderMap;
+  bindings: Binding[] = [];
 
   // TODO(bajtos) inject configuration
   constructor() {
@@ -16,5 +17,9 @@ export class AuthenticationComponent implements Component {
       [AuthenticationBindings.AUTH_ACTION.key]: AuthenticateActionProvider,
       [AuthenticationBindings.METADATA.key]: AuthMetadataProvider,
     };
+
+    this.bindings.push(
+      Binding.bind(AuthenticationBindings.CURRENT_USER).toDeferred(),
+    );
   }
 }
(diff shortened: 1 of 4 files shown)


codejamninja:
Ah yes I ran into this problem and figured it out after banging my head on the table.

raymondfeng:
@bajtos I don't think we have to introduce a new API. It can be achieved using `apply()` from https://github.com/strongloop/loopback-next/blob/lifecycle/packages/context/src/binding.ts#L437.

```ts
export const toBeBound = binding => {
  binding.toDynamicValue(() => Promise.reject(
        new Error(
          `There was no value provided for "${
            binding.key
          }" yet. Consider using \`@inject.getter()\`.`,
        ),
      );
}

ctx.bind(AuthenticationBindings.CURRENT_USER).apply(toBeBound);
```

Or maybe it can be even more simpler by defaulting `Binding._getValue()` to reject until the value is configured with `to`, `toClass`, etc. This way, we can simply do:

```ts
ctx.bind(AuthenticationBindings.CURRENT_USER); // without further calls, it will be resolved to a rejected promise with meaning error information.
```

bajtos:
@raymondfeng I prefer the code to be easy to write & understand. 

When I read `ctx.bind(AuthenticationBindings.CURRENT_USER)`, it feels incomplete - we are defining a binding but not saying anything about the value. Also, we may want to distinguish between the case when the developer forgot to provide the bound value at binding time (no setter will be called later) and the case where the bound value is intentionally not provided at binding time and a setter will provide it later.

The option `ctx.bind(AuthenticationBindings.CURRENT_USER).apply(toBeBound);` seems unnecessary long to me.

> I don't think we have to introduce a new API.

IMO: Bindings with a value set later are an integral part of Context, similarly to `@inject.getter` and `@inject.setter`. The API for defining deferred bindings early in the application setup is just a missing piece needed to improve user experience.

PR #2112, opened 2018-12-04
refactor: removed underscore from protected methods in DefaultCrudRep…

### LINKED ISSUE #2091: Why does public exposed function start with underscore?
Out of curiosity, why does the `_createHasManyRepositoryFactoryFor` function start with an underscore even though it is publically exposed?

codejamninja:
Removed underscore from protected methods

Fixes #2091 

## Checklist

### DIFF
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -147,6 +147,29 @@ export class DefaultCrudRepository<T extends Entity, ID>
   }
 
   /**
+   * @deprecated
+   * Function to create a constrained relation repository factory
+   *
+   * Use `this.createHasManyRepositoryFactoryFor()` instaed
+   *
+   * @param relationName Name of the relation defined on the source model
+   * @param targetRepo Target repository instance
+   */
+  protected _createHasManyRepositoryFactoryFor<
+    Target extends Entity,
+    TargetID,
+    ForeignKeyType
+  >(
+    relationName: string,
+    targetRepoGetter: Getter<EntityCrudRepository<Target, TargetID>>,
+  ): HasManyRepositoryFactory<Target, ForeignKeyType> {
+    return this.createHasManyRepositoryFactoryFor(
+      relationName,
+      targetRepoGetter,
+    );
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -184,9 +207,31 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+   * Use `this.createBelongsToAccessorFor()` instaed
+   *
+   * @param relationName Name of the relation defined on the source model
+   * @param targetRepo Target repository instance
+   */
   protected _createBelongsToAccessorFor<Target extends Entity, TargetId>(
     relationName: string,
     targetRepoGetter: Getter<EntityCrudRepository<Target, TargetId>>,
+  ): BelongsToAccessor<Target, ID> {
+    return this.createBelongsToAccessorFor(relationName, targetRepoGetter);
```
ludohenin: Shouldn't it make sense to log the deprecation warning ?
What's the overall project strategy to deal with this ?

bajtos: Good question. We don't have a deprecation strategy yet.

I guess the question is how much we want to force our users to update their code.

A slow process:
 1. Soft deprecations (docs only)
 2. Hard deprecation (warnings printed at runtime) - maybe wait until the next semver-major?
 3. API removed (in a semver-major release)

A faster process:
1. Hard deprecation (warnings printed at runtime)
2. API removed (in a semver-major release)

In the past, we were using the fast process. From what I heard about React, their users love their slower process for API deprecations, maybe we can do the same in LoopBack?

Implementation wise, we have had good experience using [depd](https://www.npmjs.com/package/depd) module for dealing with deprecation warnings in LoopBack 2.x/3.x

bajtos: On the second thought, since this is a simple rename that has a trivial upgrade path, I am fine with a hard deprecation from the beginning.

bajtos:
Rebased on top of the latest master, squashed the commits into a single one and slightly improved the commit message. Let's wait for CI results before landing.

bajtos:
Landed 🎉 Thank you @codejamninja for the contribution! ❤️

Issue #1952, opened 2018-10-31
[Spike] How to do relation traversal


dhmlau:
**Timebox to 5 days.**
## Description / Steps to reproduce / Feature proposal
Originated from https://github.com/strongloop/loopback-next/issues/1352

Cross posting the description from @bajtos (https://github.com/strongloop/loopback-next/issues/1352#issuecomment-433300937): 

> I think we need a spike to determine how to actually implement relation traversal, because right now, a repository for source model does not necessarily know how to obtain an instance of a repository for accessing the target model.

## Acceptance Criteria
Investigate how to implement relation traversal.  Implementation of the following DefaultCrudRepository methods to correctly support inclusion of related models as configured via `filter.include` property.
- [ ] find
- [ ] findOne
- [ ] findById

bajtos:
I have posted a comment in the other issue with a proposal how to implement inclusion of related models at Repository level, see https://github.com/strongloop/loopback-next/issues/1352#issuecomment-436197152

jannyHou:
Thanks for all the feedback! 

I understand that the best outcome of a spike is a concrete solution, but I am afraid the inclusion story is more complicated than we expect,  and it would be better to create separate stories to explorer different potential approaches.

Like what @raymondfeng points out, GraphQL's query system has two features we could borrow:

- use resolver to describe a complicated object and calculate its value
- use different query types to describe the filter of a model

And a overall design question to consider (from @bajtos ) :

How are we going to generate JSON Schema and OpenAPI schema for payloads containing both model properties (Customer's id, name, email) and data of related model (Customer's orders, addresses, etc.). At the moment, we are leveraging @property metadata. If we choose a different approach for describing model+relations in TypeScript, then we also need a new way how to obtain JSON Schema for such new types.

Here is a  proposal of the next steps:

- [ ] Explorer the approach "Define relation property as a resolver"
  - The prototype would be:
    ```js
       import { TodoList } from './TodoList';
       class Todo extends Entity {
          @inclusion()
          TodoList: ()=>TodoList
        }
    ```
- [ ] Explorer the approach "Use a type wrapper to prevent the circular reference problem"
  - The prototype would be: 
    ```js
        export type Related<T> = T | undefined;
        class Todo extends Entity {
          // ...
          todoList: Related<TodoList>;
        }
        class TodoList extends Entity {
         // ...
          todos: Related<Todo>;
        }
    ```
- [ ] Explorer "Generate decorator based filter types"
[...]

Issue #1352, opened 2018-05-24
Inclusion of related models [MVP]


dhmlau:
## Description / Steps to reproduce / Feature proposal
Follow up task for PR #1342 

Inclusion of related models (same as LB3)
For example, `Customer` model has `{include: ['order']}`.  When query on `Customer`, the result should includes the `Order` array. 

Duplicates:
 - Include related models in "GET /mymodels" results #1889
 - https://stackoverflow.com/questions/53162562/loopback-4-hasmany-include-filter-error-500/53172949

Follow-up stories:
 - Add support for "include" and "fields" to findById (REST API) #1721
 
### MVP scope

2019Q3
- [x] Add `keyFrom` to resolved relation metadata #3441
- [x] Test relations against databases #3442
- [x] Add `findByForeignKeys` helper (initial version) #3443
- [x] Introduce `InclusionResolver` concept #3445
- [x] Include related models in `DefaultCrudRepository` #3446
- [x] Implement InclusionResolver for hasMany relation #3447
- [x] Implement InclusionResolver for belongsTo relation #3448
- [x] Implement InclusionResolver for hasOne relation #3449
- [x] Update `todo-list` example to use inclusion resolver #3450

2019Q4
- [x] Reject create/update requests when data contains navigational properties #3439
- [x] Add inclusion resolvers to `lb4 relation` CLI #3451
- [x] Verify relation type in `resolve{Relation}Metadata` #3440
- [x] Run repository tests for PostgreSQL #3436
- [x] Run repository tests for Cloudant #3437
- [x] Blog post: announce Inclusion of related models #3452

### Out of MVP scope

- [ ] Support `inq` splitting in `findByForeignKeys` #3444
- [ ] Include related models with a custom scope #3453
- [ ] Recursive inclusion of related models #3454
- [ ] Reject queries with incompatible "filter.fields" and "filter.include" #3455
- [ ] Spike: robust handling of ObjectID type for MongoDB #3456

See https://github.com/strongloop/loopback-next/issues/3585 for the full list.

bajtos:
I added the following acceptance criteria:

> - [ ] Modify the implementation of the following `DefaultCrudRepository` methods to correctly support inclusion of related models as configured via `filter.include` property.
>   - [ ] `find`
>   - [ ] `findOne`
>   - [ ] `findById`
> - [ ] Include integration-level tests using in-memory database to verify the implementation.
[...]

PR #2124, opened 2018-12-05
[Spike] feat: inclusion

jannyHou:
Related to https://github.com/strongloop/loopback-next/issues/1952

Copy the discussion here: 

This PR is a PoC of the inclusion handler proposed by @batjos in [comment](https://github.com/strongloop/loopback-next/issues/1352#issuecomment-436197152). 

Takes `findById` as an example method, after adding the code that fetches the included items, `TodoList.findById(1, {include: [{relation: 'todos'}]})` returns 

screen shot 2018-11-29 at 11 18 04 pm

The PoC contains:

- A `InclusionHandlerFactory` is created to return a handler function that fetches the included items.

- When a relation is created, a corresponding handler will be registered in the source repository.

- ~TBD~ DONE: retrieve the relation metadata in `InclusionHandlerFactory` function:
  - return different handler functions according to the relation type
  - find the fkName in the relation definition

- ~TBD(improve)~ DONE: Turn the factory to a class, each repository has a `InclusionHandler` class instance as property:
  - call `this._inclusionHandler.register<TargetEntity>(targetRepositoryGetter)` to register the relation handler.
  - call `this._inclusionHandler.searchHandler(relationName)` for each entry in the filter.include.

I also discovered a few problems worth discussion:

- The `Filter` interface is not designed to describe related entities, see [its definition(and `Inclusion`'s definition)](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/query.ts#L165-L206).

- I feel the inclusion handler's implementation would have overlapping code with relation repository's find method, I will investigate how to leverage those relation repositories when register the handler.

### DIFF
--- a/examples/todo-list/src/repositories/todo-list.repository.ts
+++ b/examples/todo-list/src/repositories/todo-list.repository.ts
@@ -44,6 +44,10 @@ export class TodoListRepository extends DefaultCrudRepository<
       'image',
       todoListImageRepositoryGetter,
     );
+    this._inclusionHandler.registerHandler<Todo, typeof Todo.prototype.id>(
+      'todos',
+      todoRepositoryGetter,
+    );
   }
 
   public findByTitle(title: string) {
--- a/packages/repository/src/repositories/inclusion.ts
+++ b/packages/repository/src/repositories/inclusion.ts
@@ -1,0 +1,96 @@
+import {Entity} from '../model';
+import {Getter} from '@loopback/core';
+import {
+  Filter,
+  Where,
+  resolveHasManyMetadata,
+  resolveBelongsToMetadata,
+  RelationMetadata,
+  RelationType,
+  constrainWhere,
+} from '../';
+import {AnyObject} from '..';
+import {DefaultCrudRepository} from './legacy-juggler-bridge';
+import {inspect} from 'util';
+import {
+  HasManyDefinition,
+  BelongsToDefinition,
+  HasManyResolvedDefinition,
+  BelongsToResolvedDefinition,
+} from '../relations';
+
+type ResolvedRelationMetadata =
+  | HasManyResolvedDefinition
+  | BelongsToResolvedDefinition;
+
+// SE: the source entity
+// TE: the target entity
+// SID: the ID of source entity
+// TID: the ID of target entity
+
+export class InclusionHandler<SE extends Entity, SID> {
+  _handlers: {[relation: string]: Function} = {};
+  constructor(public sourceRepository: DefaultCrudRepository<SE, SID>) {}
+
+  registerHandler<TE extends Entity, TID>(
+    relationName: string,
+    targetRepoGetter: Getter<DefaultCrudRepository<TE, TID>>,
+  ) {
+    this._handlers[relationName] = fetchIncludedItems;
+    const self = this;
+
+    async function fetchIncludedItems(
+      fks: SID[],
+      filter?: Filter<TE>,
+    ): Promise<TE[]> {
+      const targetRepo = await targetRepoGetter();
+      const relationDef: ResolvedRelationMetadata = self.getResolvedRelationDefinition(
+        relationName,
+      );
+      filter = filter || {};
+      filter.where = self.buildConstrainedWhere<TE>(
+        fks,
+        filter.where || {},
+        relationDef,
+      );
+      console.log(`inclusion filter: ${inspect(filter)}`);
+
+      return await targetRepo.find(filter);
+    }
+  }
+
+  findHandler(relationName: string) {
+    const errMsg =
+      `The inclusion handler for relation ${relationName} is not found!` +
+      `Make sure you defined ${relationName} properly.`;
+
+    return this._handlers[relationName] || new Error(errMsg);
+  }
+
+  buildConstrainedWhere<TE extends Entity>(
+    ids: SID[],
+    whereFilter: Where<TE>,
+    relationDef: ResolvedRelationMetadata,
+  ): Where<TE> {
+    const keyPropName: string = relationDef.keyTo;
+    const where: AnyObject = {};
+    where[keyPropName] = {inq: ids};
+    return constrainWhere(whereFilter, where as Where<TE>);
+  }
+
+  getResolvedRelationDefinition(name: string): ResolvedRelationMetadata {
+    const relationMetadata: RelationMetadata = this.sourceRepository.entityClass
+      .definition.relations[name];
+
+    switch (relationMetadata.type) {
+      case RelationType.hasMany:
+        return resolveHasManyMetadata(relationMetadata as HasManyDefinition);
+      case RelationType.belongsTo:
+        return resolveBelongsToMetadata(
+          relationMetadata as BelongsToDefinition,
+        );
+      default:
+        throw new Error(`Unsupported relation type ${relationMetadata.type}`);
+    }
+  }
+}
--- a/examples/todo-list/src/controllers/todo-list.controller.ts
+++ b/examples/todo-list/src/controllers/todo-list.controller.ts
@@ -93,7 +93,6 @@ export class TodoListController {
     },
   })
   async findById(
-    return await this.todoListRepository.findById(id);
     @param.path.number('id') id: number,
     @param.query.object('filter') filter?: Filter<TodoList>,
   ): Promise<TodoList> {
--- a/examples/todo-list/src/controllers/todo-list.controller.ts
+++ b/examples/todo-list/src/controllers/todo-list.controller.ts
@@ -92,8 +92,18 @@ export class TodoListController {
       },
     },
   })
-  async findById(@param.path.number('id') id: number): Promise<TodoList> {
+  async findById(
     return await this.todoListRepository.findById(id);
+    @param.path.number('id') id: number,
+    @param.query.object('filter') filter?: Filter<TodoList>,
+  ): Promise<TodoList> {
+    // somehow the filter sent in the request query is undefined
+    // will dig more.
+    // hardcoded the inclusion filter in the PoC PR
+    const hardcodedFilterForPoC = {
+      include: [{relation: 'todos'}],
+    };
+    return await this.todoListRepository.findById(id, hardcodedFilterForPoC);
   }
 
   @patch('/todo-lists/{id}', {
--- a/examples/todo-list/src/controllers/todo.controller.ts
+++ b/examples/todo-list/src/controllers/todo.controller.ts
@@ -43,8 +43,12 @@ export class TodoController {
   async findTodoById(
     @param.path.number('id') id: number,
     @param.query.boolean('items') items?: boolean,
+    @param.query.object('filter') filter?: Filter<Todo>,
   ): Promise<Todo> {
-    return await this.todoRepo.findById(id);
+    const hardcodedFilterForPoC = {
+      include: [{relation: 'todoList'}],
+    };
+    return await this.todoRepo.findById(id, hardcodedFilterForPoC);
   }
 
   @get('/todos', {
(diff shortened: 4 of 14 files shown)


jannyHou:
From @bajtos 

@jannyHou 

> The Filter interface is not designed to describe related entities, see its definition(and Inclusion's definition).

Good catch, can you work with @raymondfeng to improve the type definition please?

We should eventually take a look at the JSON Schema/OpenAPI parameter schema emitted for `filter` too.

examples/todo-list/src/repositories/todo-list.repository.ts
```diff
@@ -44,6 +44,10 @@ export class TodoListRepository extends DefaultCrudRepository<
       'image',
       todoListImageRepositoryGetter,
     );
+    this._inclusionHandler.registerHandler<Todo, typeof Todo.prototype.id>(
```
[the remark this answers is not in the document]
jannyHou: @bajtos 

I added the belongsTo handler and it does fetch the parent instance for the child. BUT there is a problem when construct the returned data:

The `Todo` model only has property `todoListId` as the foreign key, but not `todoList` as its parent property. Therefore given the object 
```js
const todo = {
    id: 1,
    title: 'a todo item',
    todoListId: 1,
    todoList: {
      id: 1,
      color: red
    }
}
```
 method `toEntity(todo)` won't convert property `todoList` and removes it in the constructed todo entity.

packages/repository/src/repositories/inclusion.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor(public sourceRepository: DefaultCrudRepository<SE, SID>) {}
+
+  registerHandler<TE extends Entity, TID>(
+    relationName: string,
+    targetRepoGetter: Getter<DefaultCrudRepository<TE, TID>>,
+  ) {
+    this._handlers[relationName] = fetchIncludedItems;
+    const self = this;
+
+    async function fetchIncludedItems(
```
raymondfeng: It can be simplified as:
```ts
this._handlers[relationName] = async (
      fks: SID[],
      filter?: Filter<TE>,
    ): Promise<TE[]> => {...}
```

packages/repository/src/repositories/inclusion.ts
```diff
@@ -0,0 +1,96 @@
...
+type ResolvedRelationMetadata =
+  | HasManyResolvedDefinition
+  | BelongsToResolvedDefinition;
+
+// SE: the source entity
+// TE: the target entity
+// SID: the ID of source entity
+// TID: the ID of target entity
+
+export class InclusionHandler<SE extends Entity, SID> {
```
raymondfeng: The name is a bit confusing as `InclusionHandler` contains a list of `_handlers`.

packages/repository/src/repositories/inclusion.ts
```diff
@@ -0,0 +1,96 @@
...
+      return await targetRepo.find(filter);
+    }
+  }
+
+  findHandler(relationName: string) {
+    const errMsg =
+      `The inclusion handler for relation ${relationName} is not found!` +
+      `Make sure you defined ${relationName} properly.`;
+
+    return this._handlers[relationName] || new Error(errMsg);
```
raymondfeng: It's suspicious to return an `Error` but later on it checks:
```ts
const handler = this._inclusionHandler.findHandler(relation);
    if (!handler) {
      throw new Error('Fetch included items is not supported');
    }
```

raymondfeng: It's suspicious to return an `Error` but later on it checks:
```ts
const handler = this._inclusionHandler.findHandler(relation);
    if (!handler) {
      throw new Error('Fetch included items is not supported');
    }
```

packages/repository/src/repositories/inclusion.ts
raymondfeng: Maybe we should define a type such as:

```
export type IncludedItemsFetcher<SID, TE> = (
      fks: SID[],
      filter?: Filter<TE>,
    ) => Promise<TE[]>
```

packages/repository/src/repositories/inclusion.ts
```diff
@@ -0,0 +1,96 @@
...
+// SE: the source entity
+// TE: the target entity
+// SID: the ID of source entity
+// TID: the ID of target entity
+
+export class InclusionHandler<SE extends Entity, SID> {
+  _handlers: {[relation: string]: Function} = {};
+  constructor(public sourceRepository: DefaultCrudRepository<SE, SID>) {}
+
+  registerHandler<TE extends Entity, TID>(
```
raymondfeng: I like the idea to delegate resolution of `inclusion` to a list of functions.

raymondfeng: I think we can borrow some ideas from https://graphql.org/learn/execution/. It will allow us even define custom relations.

raymondfeng: https://www.apollographql.com/docs/graphql-tools/resolvers.html

bajtos:
> Which means a Todo model cannot describe an instance that includes its parent

Good catch!

When building belongsTo relation, I had a gut feeling that it's not a good idea to decorate the foreign key (e.g. `Order.prototype.customerId`) and don't provide relation property (e.g. `Order.prototype.customer`), but didn't investigate further. Inclusion of related models is a good use case to drive improvements in the design for belongsTo.

I think the current design is a workaround for TypeScript limitation explained in https://github.com/Microsoft/TypeScript/issues/27519

A possible solution is to define a type wrapper that will prevent the circular reference problem at the cost of losing design: type metadata (the metadata will say "Object" instead of the actual type).

```ts
export type Related<T> = T | undefined;

class Todo extends Entity {
  // ...

  todoList: Related<TodoList>;
}

class TodoList extends Entity {
  // ...

  todos: Related<Todo>;
}
```

_Note: AFAICT, it's important to use a named type. Replacing `Related<Todo>` with `Todo | undefined` does not break the circular reference._

For longer term, I'd like to change the way how relations are defined: relation decorators provide relation metadata only, foreign keys are defined explicitly. For example:

```ts
class Todo extends Entity {
  // ...

  @foreignKey(() => TodoList, 'id') // target model, target property
  @property({type: 'number', required: true})
  todoListId: number;

  @belongsTo(()=> TodoList)
  todo: Related<Todo>;
}
```

bajtos:
One more idea to consider - perhaps we can define two model classes, the first describing only the properties actually stored in the database, the second including relations?

```ts
class Customer extends Entity {
  // ...
}

class CustomerWithRelations extends Customer {
  orders?: Order[]
}
```

Create/update methods will greatly benefit from this distinction:
 - at the moment, it's not possible to update related models together with the source model (e.g. create a Customer and the related Order in one call)
 - at the same time, our TypeScript API (and possible OpenAPI spec too) allow relational properties in the create/update/patch input data.

packages/repository/src/repositories/inclusion.ts
jannyHou: yeah it should be plural

jannyHou:
Thanks for all the feedback! 

I understand that the best outcome of a spike is a concrete solution, but I am afraid the inclusion story is more complicated than we expect,  and it would be better to create separate stories to explorer different potential approaches.

Like what @raymondfeng points out, GraphQL's query system has two features we could borrow:

- use type resolver to describe a complicated object and calculate its value
- use different query types to describe the filter of a model


Here is a  proposal of the next steps:

- [ ] Explorer the approach "Define relation property as a type resolver"
  - The prototype would be:
    ```js
       import { TodoList } from './TodoList';
       class Todo extends Entity {
          @inclusion()
          TodoList: ()=>TodoList
        }
    ```
- [ ] Explorer the approach "Use a type wrapper to prevent the circular reference problem"
  - The prototype would be: 
    ```js
        export type Related<T> = T | undefined;
        class Todo extends Entity {
          // ...
          todoList: Related<TodoList>;
        }
        class TodoList extends Entity {
         // ...
          todos: Related<Todo>;
        }
    ```
- [ ] Explorer "Generate decorator based filter types"
  - Inspired by the query/mutation type in GraphQL, I am trying to define different partial model types described by decorators
  - The prototype:
     ```js
         // model.ts file
        class Todo extends Entity {
          @property({name: 'id', type: 'string'})
          id: string,
          @property({name: 'desc', type: 'string'})
          desc?: string
          @inclusion()
          TodoList: ()=>TodoList
        }
        // model.query.ts file
        type TodoMutation = Pick<Todo, a union of properties decorated by @property except id>
        type TodoQuery = TodoMutation | Inclusion<Todo>
     ```
- Explorer the approach "Define multiple model classes for different purpose"
  - Prototype:
    ```js
      class Customer extends Entity {
        // ...
      }

      class CustomerWithRelations extends Customer {
       orders?: Order[]
      }
    ```

bajtos:
> Explorer the approach "Define relation property as a type resolver"
> The prototype would be:

```ts
   import { TodoList } from './TodoList';
   class Todo extends Entity {
      @inclusion()
      TodoList: ()=>TodoList
    }
```

How do you envision serializing such structure to JSON, for transmission over HTTP?

> Explorer "Generate decorator based filter types"
> `type TodoQuery = TodoMutation | Inclusion<Todo>`

I like that approach! Although I am not sure how will `Inclusion<Todo>` know what is the property name where to store the target Todo instance? 

Also I think you need to use `&` operator, not `|`:

```ts
type TodoQuery = TodoMutation & Inclusion<Todo>
```

jannyHou:
> Also I think you need to use & operator, not |

Good catch!

> Although I am not sure how will Inclusion<Todo> know what is the property name where to store the target Todo instance?

That's something I am trying out on local but haven't got it work :(


> How do you envision serializing such structure to JSON, for transmission over HTTP?

Hmm could you elaborate more about it? Do you mean infer the JSON schema of a model?

bajtos:
> > How do you envision serializing such structure to JSON, for transmission over HTTP?
> > ```ts
> > class Todo extends Entity {
> >       @inclusion()
> >       TodoList: ()=>TodoList
> >     }
> > ```
> Hmm could you elaborate more about it? Do you mean infer the JSON schema of a model? 

IIUC your proposal, when we create an instance of `Todo` model, `todo.TodoList` will contain a function (the resolver).

When creating the HTTP response body, we want JSON data of the related TodoList instance, e.g.

```ts
{
  id: '1',
  desc: 'todo description',
  TodoList: {
    id: '1',
    title: 'list name',
  }
}
```

My question is how do you envision converting the value from a resolver function into a JSON object? Maybe I misunderstood your proposal?

packages/repository/src/repositories/inclusion.ts
jannyHou: yeah the code here should definitely be refined ^

b-admike:
I really like the last two proposals from https://github.com/strongloop/loopback-next/pull/2124#issuecomment-445655354 (so if we create the follow up stories and had to prioritize them, I'd like to see those two higher in the list) and the idea of moving definition of relations to repository level.

bajtos:
Another aspect to consider: how are we going to generate JSON Schema and OpenAPI schema for payloads containing both model properties (Customer's id, name, email) and data of related model (Customer's orders, addresses, etc.). At the moment, we are leveraging `@property` metadata. If we choose a different approach for describing model+relations in TypeScript, then we also need a new way how to obtain JSON Schema for such new types.

jannyHou:
@bajtos @b-admike Thanks for the feedback! I post your comment with the proposals in a new story https://github.com/strongloop/loopback-next/issues/2152. Let's continue the discussion there. 

I am closing this PoC PR. Anyone works on the follow up feel free to reopen/cherry-pick the commit if you think useful :)

AnanthGopal:
Hi Guys, 
I need this type of output.
{
  id: '1',
  desc: 'todo description',
  TodoList: {
    id: '1',
    title: 'list name',
  }
}

Did you find any solution for the related model issue?

PR #2126, opened 2018-12-06
fix(repository): use unique indexes for hasOne FK constraint

b-admike:
re-work `hasOne` relation based on https://github.com/strongloop/loopback-next/pull/2005#discussion_r238540913 and consequent comments. See if this approach is plausible. Memory connector doesn't have concept of unique indexes AFAIK, so I'll have to make a fix there to get the uniqueness on create test passing. @bajtos @raymondfeng PTAL.
## Checklist

### DIFF
--- a/packages/repository/test/acceptance/has-one.relation.acceptance.ts
+++ b/packages/repository/test/acceptance/has-one.relation.acceptance.ts
@@ -44,7 +44,7 @@ describe('hasOne relation', () => {
       street: '123 test avenue',
     });
 
-    const persisted = await addressRepo.findById(address.customerId);
+    const persisted = await addressRepo.findById(address.zipcode);
     expect(persisted.toObject()).to.deepEqual(address.toObject());
   });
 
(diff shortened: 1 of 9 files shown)


packages/repository/test/acceptance/has-one.relation.acceptance.ts
```diff
@@ -44,25 +44,27 @@ describe('hasOne relation', () => {
...
-    const persisted = await addressRepo.findById(address.customerId);
+    const persisted = await addressRepo.findById(address.zipcode);
     expect(persisted.toObject()).to.deepEqual(address.toObject());
   });
 
   it('refuses to create related model instance twice', async () => {
     const address = await controller.createCustomerAddress(existingCustomerId, {
       street: '123 test avenue',
     });
+    // FIXME(b-admike): memory connector doesn't support unique indexes
```
bajtos: Please check other commonly used connectors too! For example:
 - MongoDB
 - CouchDB/Cloudant
 - MySQL
 - PostgreSQL
 - Oracle
 - MSSQL

packages/repository/test/fixtures/models/order.model.ts
```diff
@@ -27,5 +27,6 @@ export class Order extends Entity {
   isShipped: boolean;
 
   @belongsTo(() => Customer)
+  @property()
```
bajtos: IMO, this is a breaking change, since you are requiring all existing users of `belongsTo` to add `@property` decorator to their models. Making a breaking change comes with (maintenance) costs, see https://github.com/strongloop/loopback-next/blob/8394c74d5369d7dcce861c0ed30e033a3e7ce3ae/docs/site/DEVELOPING.md#making-breaking-changes

bajtos:
I feel we need to step back a bit and review our overall design for relations and how to map it to what different SQL/NoSQL databases support.

packages/repository/test/acceptance/has-one.relation.acceptance.ts
b-admike: So far from what I checked, MongoDB, MysQL, PostgreSQL, and Oracle support unique indexes. Will check the others and report back 👍

packages/repository/test/fixtures/models/order.model.ts
b-admike: Good point. After reading the guide about frequency of breaking changes and what warrants them/support model, I have decided to revert these changes for now and just keep this PR about usage of unique indexes. @raymondfeng thoughts?

bajtos:
Based on the discussion we had in https://github.com/strongloop/loopback-next/issues/2127 and the plan outlined in https://github.com/strongloop/loopback-next/issues/2127#issuecomment-460181198, I think it makes a lot of sense to use `unique` constraint for `hasOne` relation and therefore we should resume the work on this pull request.

PR #2129, opened 2018-12-06
feat(context): add `@bind` to decorate classes with binding attributes

raymondfeng:
This PR is a reactivation of https://github.com/strongloop/loopback-next/pull/992.

## Checklist

### DIFF
--- a/packages/context/src/binding-decorator.ts
+++ b/packages/context/src/binding-decorator.ts
@@ -1,0 +1,105 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {ClassDecoratorFactory} from '@loopback/metadata';
+import {
+  asBindingTemplate,
+  asProvider,
+  BindingMetadata,
+  BindingSpec,
+  BINDING_METADATA_KEY,
+  isProviderClass,
+  asClassOrProvider,
+  removeNameAndKeyTags,
+} from './binding-inspector';
+import {Provider} from './provider';
+import {Constructor} from './value-promise';
+
+/**
+ * Decorator factory for `@bind`
+ */
+class BindDecoratorFactory extends ClassDecoratorFactory<BindingMetadata> {
+  mergeWithInherited(inherited: BindingMetadata, target: Function) {
+    if (inherited) {
+      return {
+        templates: [
+          ...inherited.templates,
+          removeNameAndKeyTags,
+          ...this.spec.templates,
+        ],
+        target: this.spec.target,
+      };
+    } else {
+      this.withTarget(this.spec, target);
+      return this.spec;
+    }
+  }
+
+  withTarget(spec: BindingMetadata, target: Function) {
+    spec.target = target as Constructor<unknown>;
+    return spec;
+  }
+}
+
+/**
+ * Decorate a class with binding configuration
+ *
+ * @example
+ * ```ts
+ * @bind((binding) => {binding.inScope(BindingScope.SINGLETON).tag('controller')}
+ * )
+ * export class MyController {
+ * }
+ * ```
+ *
+ * @param specs A list of binding scope/tags or template functions to
+ * configure the binding
+ */
+export function bind(...specs: BindingSpec[]): ClassDecorator {
+  const templateFunctions = specs.map(t => {
+    if (typeof t === 'function') {
+      return t;
+    } else {
+      return asBindingTemplate(t);
+    }
+  });
+
+  return (target: Function) => {
+    const cls = target as Constructor<unknown>;
+    const spec: BindingMetadata = {
+      templates: [asClassOrProvider(cls), ...templateFunctions],
+      target: cls,
+    };
+
+    const decorator = BindDecoratorFactory.createDecorator(
+      BINDING_METADATA_KEY,
+      spec,
+    );
+    decorator(target);
+  };
+}
+
+export namespace bind {
+  /**
+   * `@bind.provider` to denote a provider class
+   *
+   * A list of binding scope/tags or template functions to configure the binding
+   */
+  export function provider(
+    ...specs: BindingSpec[]
+  ): ((target: Constructor<Provider<unknown>>) => void) {
+    return (target: Constructor<Provider<unknown>>) => {
+      if (!isProviderClass(target)) {
+        throw new Error(`Target ${target} is not a Provider`);
+      }
+      bind(
+        // Set up the default for providers
+        asProvider(target),
+        // Call other template functions
+        ...specs,
+      )(target);
+    };
+  }
+}
--- a/packages/context/src/index.ts
+++ b/packages/context/src/index.ts
@@ -4,29 +4,14 @@
 // License text available at https://opensource.org/licenses/MIT
 
 export * from '@loopback/metadata';
-
-export {
-  isPromiseLike,
-  BoundValue,
-  Constructor,
-  ValueOrPromise,
-  MapObject,
-  resolveList,
-  resolveMap,
-  resolveUntil,
-  transformValueOrPromise,
-  tryWithFinally,
-  getDeepProperty,
-} from './value-promise';
-
-export {Binding, BindingScope, BindingType, TagMap} from './binding';
-
-export {Context} from './context';
-export {BindingKey, BindingAddress} from './binding-key';
-export {ResolutionSession} from './resolution-session';
-export {inject, Setter, Getter, Injection, InjectionMetadata} from './inject';
-export {Provider} from './provider';
-
-export {instantiateClass, invokeMethod} from './resolver';
-// internals for testing
-export {describeInjectedArguments, describeInjectedProperties} from './inject';
+export * from './binding';
+export * from './binding-decorator';
+export * from './binding-inspector';
+export * from './binding-key';
+export * from './context';
+export * from './inject';
+export * from './keys';
+export * from './provider';
+export * from './resolution-session';
+export * from './resolver';
+export * from './value-promise';
--- a/packages/context/test/acceptance/bind-decorator.acceptance.ts
+++ b/packages/context/test/acceptance/bind-decorator.acceptance.ts
@@ -1,0 +1,70 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {Context, bind, BindingScope, Provider} from '../..';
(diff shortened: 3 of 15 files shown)


packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+import {ClassDecoratorFactory, MetadataInspector} from '@loopback/metadata';
+import {Binding, BindingScope, BindingTemplate, TagMap} from './binding';
+import {Provider} from './provider';
+import {Constructor} from './value-promise';
+
+const BINDING_METADATA_KEY = 'binding.metadata';
+
+type BindingMetadata = {
+  templates: BindingTemplate[];
+  target?: Constructor<unknown>;
```
bajtos: What is the scenario where `target` ends up undefined? 

I think you need to define two interfaces:
 - configuration properties accepted by `@bind` decorator
 - the metadata object created by `@bind` decorator (config properties + target)

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  withTarget(spec: BindingMetadata, target: Function) {
+    spec.target = target as Constructor<unknown>;
+    return spec;
+  }
+}
+
+export type BindingScopeAndTags = {
+  scope?: BindingScope;
+  tags?: string | TagMap | (string | TagMap)[];
```
bajtos: This type definition looks long:

```ts
string | TagMap | (string | TagMap)[];
```

Can we use a named type please? Please check for other places that may be using this union type, and update them to use the new named type too.

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+   * A list of binding scope/tags or template functions to configure the binding
+   */
+  export function provider(
+    ...templates: (BindingTemplate | BindingScopeAndTags)[]
+  ): ((target: Constructor<Provider<unknown>>) => void) {
+    return (target: Constructor<Provider<unknown>>) => {
+      bind(
+        // Set up the default for providers
+        binding =>
+          binding.toProvider(target).tag('provider', {type: 'provider'}),
```
bajtos: Should we define a constant for `provider`, e.g. `ContextTags.PROVIDER`?

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+ */
+export function bindingTemplateFor(cls: Constructor<unknown>): BindingTemplate {
+  const spec = getBindingMetadata(cls);
+  const templateFunctions = (spec && spec.templates) || [];
+  return binding => {
+    for (const t of templateFunctions) {
+      binding.apply(t);
+    }
+    if (spec && spec.target !== cls) {
+      // Remove name/key from base classes
```
bajtos: This comment can be read in two ways:
 1. modify base classes and remove name/key
 2. modify binding and remove name/key values set by base classes

I think the second option is describing your intent, in which case I am proposing slightly different wording:

```
Remove name & key tags that were configured by base classes
```

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+      binding.apply(removeNameAndKey);
+    }
+  };
+}
+
+export const TYPE_NAMESPACES: {[name: string]: string} = {
```
bajtos: Let's move this map into a standalone source code file please.

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+  controller: 'controllers',
+  repository: 'repositories',
```
bajtos: ❗️ Context should not need to know about concepts from dependants (repository, controller, etc.), otherwise we have a cyclic dependency, at least at conceptual level: repository depends on context, but context needs to understand repository concepts.

Also, we this hard-coded map is not extensible. How can 3rd-party extensions register new mappings?

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+ * @param cls A class
+ * @param key Optional binding key. If not present, infer it from the class
+ */
+export function createBindingFromClass<T = unknown>(
+  cls: Constructor<unknown>,
+  key?: string,
+): Binding<T> {
+  const templateFn = bindingTemplateFor(cls);
+  // Create a dummy binding if key is not provided
+  if (!key) {
```
bajtos: Please extract the code building the key into a standalone function and use early returns to keep the indentation level low.

```ts
if (key) return key;
// ...
key = bindingTemplate.tagMap['key'];
if (key) return key;
// ...
key = `${getNamespace(type)}.${name}`;
return key;
```

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+ * Create a binding from a class with decorated metadata
+ * @param cls A class
+ * @param key Optional binding key. If not present, infer it from the class
+ */
+export function createBindingFromClass<T = unknown>(
+  cls: Constructor<unknown>,
+  key?: string,
+): Binding<T> {
+  const templateFn = bindingTemplateFor(cls);
+  // Create a dummy binding if key is not provided
```
bajtos: Please describe the algorithm for building binding keys in the docs. 

1. Framework users need to understand how this "automagic" works, what keys to expect.
2. People reviewing this pull request will have simpler life as they won't need to build the understanding of the algorithm from code.

packages/context/src/index.ts
```diff
@@ -19,12 +19,19 @@ export {
...
+  BindingType,
+  TagMap,
+  BindingTemplate,
+} from './binding';
 
 export {Context} from './context';
 export {BindingKey, BindingAddress} from './binding-key';
 export {ResolutionSession} from './resolution-session';
 export {inject, Setter, Getter, Injection, InjectionMetadata} from './inject';
+export * from './binding-decorator';
```
bajtos: Let's try to keep these exports sorted by file names please. I.e. this new line should be added after `export /*...*/ from './binding-key'`.

packages/context/src/value-promise.ts
```diff
@@ -28,7 +28,8 @@ export type BoundValue = any;
  */
 export type ValueOrPromise<T> = T | PromiseLike<T>;
 
-export type MapObject<T> = {[name: string]: T};
+// tslint:disable-next-line:no-any
+export type MapObject<T = any> = {[name: string]: T};
```
bajtos: AFAICT, this change is not necessary - I don't see any place in this pull request using `MapObject` without providing `T` value.

Please revert.

packages/context/src/binding-decorator.ts
```diff
@@ -0,0 +1,215 @@
...
+ * )
+ * export class MyController {
+ * }
+ * ```
+ *
+ * @param templates A list of binding scope/tags or template functions to
+ * configure the binding
+ */
+export function bind(
+  ...templates: (BindingTemplate | BindingScopeAndTags)[]
```
bajtos: Personally, I would restrict the allowed signatures to `BindingTemplate[] | BindingScopeAndTags`.

Either way, please make sure there are tests covering all different combinations allowed by this API, both for `@bind()` and `@bind.provider()`. For example: 
 - multiple templates
 - multiple scope+tags objects
 - combinations of the above

packages/context/src/binding-decorator.ts
bajtos: IMO, Context should provide API allowing other modules to register custom extensions. `@loopback/core`, `@loopback/repository` and `@loopback/service-proxy` should leverage this API to register their type namespaces. The list of namespaces built into `@loopback/context` should contain only `class` and `provider`.

Alternatively, replace `TYPE_NAMESPACES` map with a different mechanism that's easier to extend.

packages/context/src/binding-decorator.ts
raymondfeng: Added tests

raymondfeng:
@bajtos PTAL

packages/context/test/acceptance/bind-decorator.acceptance.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
```
b-admike: Nitpick: I wonder if we should use `Copyright IBM Corp. 2018.` here.

packages/context/test/acceptance/bind-decorator.acceptance.ts
```diff
@@ -0,0 +1,70 @@
...
+      const binding = createBindingFromClass(c, {
+        typeNamespaceMapping: {
+          controller: 'controllers',
+          service: 'services',
+        },
+      });
+      ctx.add(binding);
+    });
+    expect(ctx.findByTag('provider').map(b => b.key)).eql(['my-date-provider']);
+    expect(ctx.getBinding('services.MyService').scope).to.eql(
```
b-admike: To show the customization, can we use different names in `typeNameSpaceMapping` object above for `createBindingFromClass`? i.e. something like 
```
     typeNamespaceMapping: {
          controller: 'myControllers',
          service: 'myServices',
        }
```
my reasoning is from the test above, the name spaces by default are plural of the artifact type.

packages/context/test/acceptance/binding-decorator.feature.md
```diff
@@ -0,0 +1,52 @@
...
+
+Developers can help the bootstrapper by decorating these classes with `@bind`.
+
+```ts
+@bind({tags: ['log']})
+export class LogController {}
+
+export const LOG_LEVEL = 'info';
+
+@bind({type: 'provider', tags: ['log']})
```
b-admike: Nitpick: maybe we can showcase the sugar `bind.provider` here instead :).

packages/context/test/unit/binding-decorator.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    @bind.provider(spec)
+    class MyProvider implements Provider<string> {
+      value() {
+        return 'my-value';
+      }
+    }
+
+    expect(inspectScopeAndTags(MyProvider)).to.eql({
+      tags: {rest: 'rest', type: 'provider', provider: 'provider'},
```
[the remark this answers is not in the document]
raymondfeng: `type: provider` is mostly used to infer the binding key. The `provider` tag itself is often used to find providers.

packages/context/test/acceptance/binding-decorator.feature.md
```diff
@@ -0,0 +1,52 @@
+# Feature: @bind for classes representing various artifacts
+
+- In order to automatically bind classes for various artifacts to a context
+- As a developer
```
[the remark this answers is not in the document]
raymondfeng: I follow other acceptance md files.

raymondfeng:
@b-admike Thank you for the review comments. I have addressed most of them.

PR #2136, opened 2018-12-09
feat(cli): allow annonymous schemas in openapi to be mapped to models

raymondfeng:
This PR adds `--promote-anonymous-schemas` option to map anonymous schemas in OpenAPI specs to LoopBack 4 models. Without the flag, we generate such schemas as TypeScript type literals.

## Checklist

### DIFF
--- a/docs/site/OpenAPI-generator.md
+++ b/docs/site/OpenAPI-generator.md
@@ -19,7 +19,9 @@ lb4 openapi [<url>] [options]
 ### Options
 
 - `--url`: URL or file path of the OpenAPI spec.
-- `--validate`: Validate the OpenAPI spec. Default: false.
+- `--validate`: Validate the OpenAPI spec. Default: `false`.
+- `--promote-anonymous-schemas`: Promote anonymous schemas as models classes.
+  Default: `false`.
 
 ### Arguments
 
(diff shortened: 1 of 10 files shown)


bajtos:
@raymondfeng could you please update the documentation in [OpenAPI-generator.html](https://loopback.io/doc/en/lb4/OpenAPI-generator.html) to explain the new CLI flag in more details? For example, show what happens when the flag is turned off vs. when it's turned on. As a person not very familiar with `lb4 openapi`, I have difficulties understanding when would I want to use this flag and when not.

@jannyHou Could you please review this pull request? You are quite familiar with LB3 CLI and a code-owner of many OpenAPI-spec related modules in loopback-next, I think that makes you a well-qualified person for reviewing this change :)

packages/cli/generators/openapi/schema-helper.js
```diff
@@ -334,6 +317,58 @@ function generateModelSpecs(apiSpec, options) {
...
+      // Skip anonymous schemas
+      return;
+    }
+    // Infer a schema name
+    if (schema.title) {
+      schemaName = camelCase(schema.title);
+    } else {
+      if (typeRegistry.nameIndex == null) typeRegistry.nameIndex = 0;
+      typeRegistry.nameIndex++;
+      schemaName = 'AnonymousType_' + typeRegistry.nameIndex;
```
bajtos: Can we please infer a more descriptive schema name than "Anonymous_XYZ", perhaps using operation name and argument name.

For example, in the USPTO spec, the operation `list-searchable-fields` uses several anonymous schemas, they can be called:
 - the first parameter: `ListSearchableFields_Dataset`
 - the second parameter: `ListSearchableFields_Version`
 - the third parameter: `ListSearchableFieldsResponse` (not using `_` to avoid clash with a parameter called "Response")

packages/cli/test/unit/openapi/schema-model.unit.js
```diff
@@ -63,6 +69,140 @@ describe('schema to model', () => {
...
+      },
+      {
+        description: 'AnonymousType_2',
+        name: 'AnonymousType_2',
+        className: 'AnonymousType2',
+        fileName: 'anonymous-type-2.model.ts',
+        properties: [],
+        imports: [],
+        import: "import {AnonymousType2} from './anonymous-type-2.model';",
+        declaration: 'string',
```
bajtos: IIUC, we are defining a custom type (a custom class) to represent a string type? I feel that's an overkill. Also, from my experience, built-in types like `String` and `Number` are difficult to subclass in JavaScript because of edge cases where `String` instance behaves differently from `string` value.

I am proposing to create anonymous types only for objects and arrays, and leave scalar values (strings, numbers, booleans) as anonymous schemas.

Thoughts?

packages/cli/test/unit/openapi/schema-model.unit.js
```diff
@@ -63,6 +69,140 @@ describe('schema to model', () => {
...
+        className: 'AnonymousType7',
+        fileName: 'anonymous-type-7.model.ts',
+        properties: [],
+        imports: [],
+        import: "import {AnonymousType7} from './anonymous-type-7.model';",
+        declaration: '{\n  \n}[]',
+        signature: 'AnonymousType7',
+        itemType: {
+          imports: [],
+          declaration: '{\n  \n}',
```
bajtos: IIUC, this is defining an empty object type that does not have any properties, is my understanding correct? What is the intended usage of such type? 

Assuming this type is representing the following response schema:

```yaml
              schema:
                type: array
                items:
                  type: object
                  additionalProperties:
                    type: object
```

According to JSON Schema (https://json-schema.org/understanding-json-schema/reference/object.html#properties), `additionalProperties: {type: 'object'}` means that the item objects can contain any additional properties holding values of object type. For example:

```ts
response = [
  {foo: {flag: true, name: 'first'}},
  {bar: {flag: false}}
];
```

I think you need to take `additionalProperties` into consideration when emitting TypeScript declaration:

```ts
{
  [additionalProperty: string]: object
}
```

This may be out of scope of this pull request, IDK.

packages/cli/test/unit/openapi/schema-model.unit.js
raymondfeng: +1, that was a mistake.

packages/cli/test/unit/openapi/schema-model.unit.js
raymondfeng: Fixed.

raymondfeng:
@bajtos Good catches. I have addressed your comments and added an integration tests to verify. PTAL.

docs/site/OpenAPI-generator.md
```diff
@@ -135,10 +137,148 @@ import {NewPet} from './new-pet.model.ts';
...
+```ts
+/* tslint:disable:no-any */
+import {model, property} from '@loopback/repository';
+
+/**
+ * The model class is generated from OpenAPI schema - performSearchBody
+ * performSearchBody
+ */
+@model({name: 'performSearchBody'})
+export class PerformSearchBody {
```
bajtos: Should we include "Request" in the name, to distinguish from "Response" body? I.e. `PerformSearchRequestBody`. (PerformSearchResponseBody is already used later in this file).

packages/cli/generators/openapi/spec-helper.js
```diff
@@ -182,15 +223,23 @@ function buildMethodSpec(controllerSpec, op, options) {
...
-    const jsonType = content && content['application/json'];
-    if (jsonType && jsonType.schema) {
-      bodyType = mapSchemaType(jsonType.schema, options);
-      addImportsForType(bodyType);
-    }
+    const contentType =
+      content &&
+      (content['application/json'] || content[Object.keys(content)[0]]);
+
     let bodyName = 'body';
```
bajtos: `request-body`? (see my previous comment)

bajtos:
> coverage/coveralls — Coverage decreased (-0.09%) to 90.179%

@raymondfeng please check the coverage report, there may be a valid scenario/use-case that's not covered by tests now.

packages/cli/test/integration/generators/openapi-uspto-with-anonymous.integration.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    assert.fileContent(
+      performSearchRequestBodyModel,
+      'export class PerformSearchRequestBody {',
+    );
+
+    assert.fileContent(
+      performSearchResponseBodyModel,
+      'export type PerformSearchResponseBody = {',
```
jannyHou: How do we determine export a type or a class?

raymondfeng: Schemas with `object` type are usually generated as classes but other types such as `array` or union types are generated as TypeScript types.

Issue #1718, opened 2018-09-17
lb4 relations referential integrity for hasMany/belongsTo


virkt25:
## Description / Steps to reproduce / Feature proposal

As I was working on the user order history component for https://github.com/strongloop/loopback4-example-shopping I ran into an issue I thought LoopBack would've taken care of for me -- referential integrity with the `@hasMany` relation. 

## Current Behavior

I am able to create an order using `/post/{userId}/orders` for a userId which is not that of a user. I would've expected some sort of a check that would've ensure that the `userId` exists.

## Expected Behavior

Throw an error for an `userId` not in the database. 

## Acceptance Criteria
For longer term, we can look into a better way to handle referential integrity.  For now:
- [ ] check the database whether the userId exist and then make the post call

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

dhmlau:
In discussion with @raymondfeng @bajtos: 
Proposed solution: check the database whether the userId exist and then make the post call
In the future, we can look into a better way to handle it.

jannyHou:
Defer the estimation until we figure out at which level are we going to apply the constraint.
There are different approaches:
- repository level: the framework handles it for users, enforce on the database side
- controller level: fix the shopping example, enforce on the client side

bajtos:
In our recent discussion about HasOne's unique constraint for the foreign key, @raymondfeng @b-admike and me discovered a new way how to look at relations and constraints. Essentially, we see two levels of relations:
 - Weak relations, where the developer primarily wants to navigate related models (think of GraphQL) and referential integrity is either not important or already enforced by the (SQL) database.
 - Strong relations, where the referential integrity is guaranteed by the framework together with the database.

I think this issue (#1718) is applicable to Strong relations only.

IMO, referential integrity must be guaranteed at repository level (or lower). For example, when a test helper invokes repository APIs to seed the database with sample data, then we want the referential integrity to be preserved. If we implemented referential integrity at controller level, then consumers of Repository API would be bypassing these checks and would be able to created corrupted data.

Implementation wise, it's important to implement referential integrity in way that does not create any race conditions. For example, a naive implementation calling `findById` on the source model before creating a new instance of the target model creates a race condition - if another request deletes the source model instance after `findById` received database response but before `create` command is set, then we end up with corrupted data.

[...]

PR #2147, opened 2018-12-12
fix(repository): revert hasOne target FK as PK implementation

b-admike:
Remove attempt at guaranteeing the foreign key of a has one relation to be unique by using it
as the primary key of the target model. This allow users to set up their own unique contstraints. @RaphaelDrai FYI (what I mentioned in https://github.com/strongloop/loopback-next/pull/2005#issuecomment-445991523)

Motivation:
>@bajtos @raymondfeng and I had a discussion and since the memory connector and cloudant connector do not support unique indexes, we've decided to have the first release of hasOne provide "weak" relations i.e. ease of navigation and aggregation instead of enforcement of referential integrity and document it so users are aware of the limitations. 

TODO: docs for the `hasOne` relation (most likely in another PR, so we can do a release).

## Checklist

### DIFF
--- a/examples/todo-list/src/models/todo-list-image.model.ts
+++ b/examples/todo-list/src/models/todo-list-image.model.ts
@@ -1,9 +1,15 @@
-import {Entity, model, property, belongsToUniquely} from '@loopback/repository';
+import {Entity, model, property, belongsTo} from '@loopback/repository';
 import {TodoList} from './todo-list.model';
 
 @model()
 export class TodoListImage extends Entity {
-  @belongsToUniquely(() => TodoList)
+  @property({
+    type: 'number',
+    id: true,
+  })
+  id: number;
+
+  @belongsTo(() => TodoList)
   todoListId?: number;
 
   @property({
--- a/examples/todo-list/src/repositories/index.ts
+++ b/examples/todo-list/src/repositories/index.ts
@@ -5,3 +5,4 @@
 
 export * from './todo.repository';
 export * from './todo-list.repository';
+export * from './todo-list-image.repository';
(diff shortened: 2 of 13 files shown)


RaphaelDrai:
Hello @b-admike, 
I understand that the change will guarantee the uniqueness relation from the source to the target but, the uniqueness of the relation in the opposite direction (from the target to the source)  is upon the user decision, this is right?
If the answer is positive, can you please elaborate the two way configurations  in the @model() for the user to guarantee or ignore the uniqueness from the target to the source direction? 
I mean how to configure the @hasOne and @belongsTo decorators in the @model().
Thanks.

b-admike:
> Hello @b-admike,
> I understand that the change will guarantee the uniqueness relation from the source to the target but, the uniqueness of the relation in the opposite direction (from the target to the source) is upon the user decision, this is right?
> If the answer is positive, can you please elaborate the two way configurations in the @model() for the user to guarantee or ignore the uniqueness from the target to the source direction?
> I mean how to configure the @hasone and @belongsTo decorators in the @model().
> Thanks.

Actually the changes do not guarantee uniqueness of the relation in any way, it is up to our users to set up the underlying database to be used instead of LoopBack doing so since we couldn't find a generic approach for it and need to investigate how to deliver a robust solution. You can find more info at https://github.com/strongloop/loopback-next/issues/1718#issuecomment-446906249 and #2161.

gczobel-f5:
But what about in-memory? In this case the is no DB...

RaphaelDrai:
Hello,
What is the status of the missing implementation of DELETE and PATCH CRUD operations?
Thanks,
Raphael

bajtos:
@RaphaelDrai Good catch! I think those two operations should have been implemented as part of https://github.com/strongloop/loopback-next/issues/1422 but somehow that task slipped out.

Please create a new GitHub issue for this missing feature.

I am going to lock this pull request to collaborators only, please open new issues for any new discussions.

PR #2148, opened 2018-12-13
fix: cli/generators/openapi: required is not generated in properties …

appcoders:
…decoration

When using openapi generator a schema with required properties won't issue a 400 error on omitted properties.

Currently the following is generated for a required property:

```ts
  @property({name: 'id'})
  id: number;  // ? is omitted as it is not optional
```

The generator should produce this:

```ts
  @property({name: 'id', required: true})
  id: number;
```

## Checklist

- [n/a] API Documentation in code was updated
- [n/a] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [n/a] Affected artifact templates in `packages/cli` were updated
- [n/a] Affected example projects in `examples/*` were updated


appcoders:
I accepted the CLA no.

bajtos:
Thank you @appcoders for the pull request, the change makes sense to me 👍 

I am not sure why `clahub` is complaining, could you please try to sign our CLA again?

Notice that the tests are failing now. Please fix the existing tests to match the new behavior.

Ideally, we should end up with two tests:
 - One defining a required property and verifying that `required: true` was added.
 - Another defining an optional property and verifying tat `required: false` was added.

Considering that properties are optional by default, I think it would make for better developer experience if the decorator for optional properties was leaving out `required: false` entirely.

packages/cli/generators/openapi/schema-helper.js
```diff
@@ -160,7 +160,13 @@ function mapObjectType(schema, options) {
       );
       // The property name might have chars such as `-`
       const propName = escapePropertyOrMethodName(p);
+      
       let propDecoration = `@property({name: '${p}'})`;
+   
+      if (required.includes(p)) {
+		propDecoration = `@property({name: '${p}', required: true})`;
```
raymondfeng: Please run `npm run lint:fix` to fix style issues.

appcoders: done

raymondfeng:
@appcoders I have squashed the commits and merged it into master - https://github.com/strongloop/loopback-next/commit/3c372868196ee7e18d8532f00c624c3f72378bf0.

Thanks!

PR #2159, opened 2018-12-14
Extract tslint config into a standalone package

bajtos:
See https://github.com/strongloop/loopback-next/pull/2156

By keeping the tslint config in an independently versioned package `@loopback/tslint-config`, we can properly communicate breaking changes without having to worry about long-term support for multiple versions of `@loobpack/build`.

## Checklist

### DIFF
--- a/examples/hello-world/tslint.build.json
+++ b/examples/hello-world/tslint.build.json
@@ -1,4 +1,4 @@
 {
   "$schema": "http://json.schemastore.org/tslint",
-  "extends": ["./node_modules/@loopback/build/config/tslint.build.json"]
+  "extends": ["@loopback/tslint-config/tslint.build.json"]
 }
--- a/packages/build/package.json
+++ b/packages/build/package.json
@@ -42,5 +42,8 @@
   "scripts": {
     "test": "npm run mocha",
     "mocha": "node bin/run-mocha --timeout 30000 \"test/integration/*.js\""
+  },
+  "devDependencies": {
+    "@loopback/tslint-config": "^1.0.0-1"
   }
 }
--- a/packages/build/package.json
+++ b/packages/build/package.json
@@ -13,6 +13,7 @@
   "copyright.owner": "IBM Corp.",
   "license": "MIT",
   "dependencies": {
+    "@loopback/tslint-config": "^1.0.0-1",
     "@types/mocha": "^5.0.0",
     "@types/node": "^10.11.2",
     "cross-spawn": "^6.0.5",
--- a/packages/tslint-config/package.json
+++ b/packages/tslint-config/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@loopback/tslint-config",
+  "version": "1.0.0-1",
+  "description": "",
+  "engines": {
+    "node": ">=8.9"
+  },
+  "author": "IBM",
+  "copyright.owner": "IBM Corp.",
+  "license": "MIT",
+  "peerDependencies": {
+    "tslint": ">=5.11.0"
+  },
+  "files": [
+    "README.md",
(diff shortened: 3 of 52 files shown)


packages/build/package.json
```diff
@@ -13,6 +13,7 @@
   "copyright.owner": "IBM Corp.",
   "license": "MIT",
   "dependencies": {
+    "@loopback/tslint-config": "^1.0.0-1",
```
b-admike: Hmm why is this dependency`^1.0.0-1` as opposed to `^1.0.0`?

examples/hello-world/tslint.build.json
```diff
@@ -1,4 +1,4 @@
 {
   "$schema": "http://json.schemastore.org/tslint",
-  "extends": ["./node_modules/@loopback/build/config/tslint.build.json"]
+  "extends": ["./node_modules/@loopback/tslint-config/tslint.build.json"]
```
raymondfeng: We should use the module name instead:
```
"extends": ["@loopback/tslint-config/tslint.build.json"]
```
See https://palantir.github.io/tslint/usage/configuration/

raymondfeng:
Do we want to extend this module to include `tsconfig` too? Please note that TypeScript 3.2.x allows references to module names for `tsconfig` now. If so, the module name can be `@loopback/ts-config`.

examples/hello-world/tslint.build.json
bajtos: Ah, of course, good catch!

bajtos:
> Do we want to extend this module to include tsconfig too? Please note that TypeScript 3.2.x allows references to module names for tsconfig now. If so, the module name can be @loopback/ts-config.

No, I want to decouple tslint config as much as possible. We can take a look at a shared tsconfig later, I don't have a strong opinion whether it would be useful to have a standalone package for tsconfig or not.

examples/hello-world/tslint.build.json
bajtos: Fixed.

packages/tslint-config/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "1.0.0-1",
+  "description": "",
+  "engines": {
+    "node": ">=8.9"
+  },
+  "author": "IBM",
+  "copyright.owner": "IBM Corp.",
+  "license": "MIT",
+  "peerDependencies": {
+    "tslint": ">=5.11.0"
```
bajtos: Hmm. The idea behind this peer dependency is to specify the minimum version of `tslint` that can understand our rules. Unfortunately, a typical LB4 project does not install `tslint` directly but gets it from `@loopback/build` instead.

As a result, `npm install` can print the following warning:

```
npm WARN @loopback/tslint-config@1.0.0-1 requires a peer of tslint@>=5.11.0 but none is installed. You must install peer dependencies yourself.
```

I am not sure what to do about it. We can remove this peer dependency, but that way we loose control over `tslint` version used with our config. Or we can modify our project templates and example projects to install `tslint` directly, kind of duplicating `@loopback/build`'s dependency on tslint. 

I think in the long term, it may be best to get rid of `lb-tslint` and use https://github.com/Microsoft/tslint-microsoft-contrib to invoke tslint as part of `tsc` build.

@strongloop/loopback-maintainers thoughts?

bajtos: What's worse, some 3rd-party tslint rules like tslint-consistent-codestyle (see https://github.com/strongloop/loopback-next/pull/2156) are loading `tslint` directly and fail if it's not installed as a peer dependency (e.g. because it's sourced from` @loopback/build`). I'll have to deal with this problem in #2156.

PR #2166, opened 2018-12-17
refactor: remove dependency of package.json keywords

### LINKED ISSUE #2163: Loopback CLI depends on "loopback" keyword being present in package.json
I removed the `keywords` section of my `package.json` and the next time I ran `lb4 model`, I received this error: 

```
Generation is aborted: Error: No `loopback` keyword found in /Users/dcain/Code/meal-api/package.json. The command must be run in a LoopBack project.
```

I add "loopback" as a keyword in my `package.json` and the problem goes away. I see [here](https://github.com/strongloop/loopback-next/blob/0a42541143636208832207669c0a2c36b7c9c1de/packages/cli/lib/base-generator.js#L326) that this is to make sure that we are in a loopback project.

I mean, this is an approach that possibly works, but it does feel kinda weird to have to keep the "loopback" keyword in my package.json. Instead, why don't we check the `dependencies` for a loopback package? Right below the code in the link above, the dependencies are parsed so it would be easy to look there. Or, what happens if we just skip that check? Is it really needed?

dericgw:
When the CLI performed an artifact generation, it would validate that the CLI's current working directory was a directory in which a Loopback application is installed. In order to do this, it looked at the package.json's "keywords" section and looked to see if the "loopback" keyword was present in that section. So that we do not have to force users to have the "loopback" keyword in their package.json, we can simply check the package.json dependencies to make sure that the `@loopback/core` package is installed. 

Fixes #2163 

## Checklist

### DIFF
--- a/packages/cli/test/test-utils.js
+++ b/packages/cli/test/test-utils.js
@@ -54,7 +54,7 @@ exports.executeGenerator = function(GeneratorOrNamespace, settings) {
  *
  * @param {string} rootDir Root directory in which to create the project
  * @param {Object} options
- * @property {boolean} excludeKeyword Excludes the 'loopback' keyword in package.json
+ * @property {boolean} excludeLoopbackCore Excludes the '@loopback/core' dependency in package.json
  * @property {boolean} excludePackageJSON Excludes package.json
  * @property {boolean} excludeYoRcJSON Excludes .yo-rc.json
  * @property {boolean} excludeControllersDir Excludes the controllers directory
@@ -70,9 +70,16 @@ exports.givenLBProject = function(rootDir, options) {
   options = options || {};
   const sandBoxFiles = options.additionalFiles || [];
 
-  const content = {};
-  if (!options.excludeKeyword) {
-    content.keywords = ['loopback'];
+  const content = {
+    dependencies: {
+      '@loopback/core': '*',
+    },
+  };
+
+  // We infer if a project is loopback by checking whether its dependencies includes @loopback/core or not.
+  // This flag is created for testing invalid loopback projects.
+  if (options.excludeLoopbackCore) {
+    delete content.dependencies['@loopback/core'];
   }
 
   if (!options.excludePackageJSON) {
--- a/packages/cli/lib/base-generator.js
+++ b/packages/cli/lib/base-generator.js
@@ -305,14 +305,14 @@ module.exports = class BaseGenerator extends Generator {
 
   /**
    * Checks if current directory is a LoopBack project by checking for
-   * keyword 'loopback' under 'keywords' attribute in package.json.
-   * 'keywords' is an array
+   * "@loopback/core" package in the dependencies section of the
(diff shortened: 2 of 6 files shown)


packages/cli/test/test-utils.js
```diff
@@ -71,8 +71,9 @@ exports.givenLBProject = function(rootDir, options) {
   const sandBoxFiles = options.additionalFiles || [];
 
   const content = {};
-  if (!options.excludeKeyword) {
-    content.keywords = ['loopback'];
+  if (!options.excludeDependencies) {
```
jannyHou: How about we change the logic to remove the `@loopback/core` from `package.json` when the flag enabled? 

```js
if (options.excludeDependencies) {
  // may need some extra check to make sure the `dependencies` exist
  delete content.dependencies['@loopback/core']
}
```
And maybe rename the flag to be more specific, like `excludeLoopbackCore`. 
Also add a few comments to explain its usage: We infer if a project is loopback by checking its dependencies includes `@loopback/core` or not. This flag is created for testing invalid loopback projects.

sandbox/example/node_modules/@loopback/tslint-config
```diff
@@ -0,0 +1 @@
+../../../../packages/tslint-config
```
bajtos: I think this file does not belong to this pull request, could you remove it please?

packages/cli/test/test-utils.js
```diff
@@ -70,9 +70,16 @@ exports.givenLBProject = function(rootDir, options) {
   options = options || {};
   const sandBoxFiles = options.additionalFiles || [];
 
-  const content = {};
-  if (!options.excludeKeyword) {
-    content.keywords = ['loopback'];
+  const content = {
+    dependencies: {
+      '@loopback/core': 'X.*',
```
bajtos: Can we simplify the version spec as follows?
```suggestion
      '@loopback/core': '*',
```

dericgw:
@bajtos Can’t you just squash when merging? That is what our team always does because it’s a lot easier than requesting the contributor to rebase each time we request changes.

bajtos:
> Can’t you just squash when merging?

We prefer to run a full CI build of the final commit(s) before landing. Without that, it's too easy to accidentally land changes that does not pass our commit message linter.

> It’s a lot easier than requesting the contributor to rebase each time we request changes.

I agree requesting contributor to rebase & clean commit messages adds extra time to the process. Fortunately, GitHub allows us to do the cleanup ourselves and push to your feature branch, that's what I do if the contributor does not have bandwidth to do the final cleanup.

raymondfeng:
LGTM, assuming the CI failure will be fixed by rebasing - https://travis-ci.org/strongloop/loopback-next/jobs/476321979

raymondfeng:
@dericgw I have squashed the commits.

bajtos:
Landed, thank you @dericgw for the contribution!

Issue #1517, opened 2018-07-11
[CLI] Add `--config` support for `lb4 model`


virkt25:
## Acceptance Criteria

Add support for `--config` for `lb4 model` as follows:

- [ ] `--config` accepts an Object with a `properties` object that describes the properties for a model
  - [ ] If there are properties present, skip the prompts for properties and just generate the file.

- [ ] This will allow us to test `property` generation logic (not the prompts) BY feeding in the properties via `json` and then checking the generated model. WRITE tests for testing the different types of properties and options available for properties to ensure we generate correct files.

raymondfeng:
I don't think there is need to introduce `--json` as we already have `--config` which takes a json file or stringified json.

virkt25:
Oh, I might've confused the flags ... but it's basically adding support for the `--config` flag (I thought it was `--json`) ... by skipping over the prompts for properties if supplied by a user.

marioestradarosa:
Adding a use case for @virkt25  proposal. We use models that have more than 20,30 columns. We create manually our models and for this scenario we would not be using an lb4 model wizard. But we would like an option that could generate a basic repository out of this definition file. 

`lb4 model --config src/models/large.definition.ts``

And then it would create the src/models/large.model.ts and src/repositories/large.repository.ts automatically, that could help :-)

dhmlau:
@marioestradarosa, It might be a stupid question... what would be the difference between `large.definition.ts` that user provides and `large.model.ts` that LB CLI generates?

marioestradarosa:
@dhmlau  😄 forgive me, but I think I am the one who made the mistake for the .ts extension. I meant .json extension and I meant having a large table defined in json format (**LB3** ), we should provide a way to migrate to **LB4** with single command

 `lb4 model --config-json="anydirectory/large.table.json`

or

`lb4 model --config-json="anydirectory/large.table.json --with-repository`

and it would create the **LB4** artifacts in `src/models`

By large, I meant a table with several columns. I believe for demos is fine to have the `lb4 model` to ask for properties etc., but in large applications that wouldn't work.  So this feature makes sense for two scenarios:

1. As part of a migration path from LB3 to LB4
2. As an option for enterprise applications where we have the database designed already and in production.

Developers can create these  json files compatible with juggler easily out of their production schema (reverse-engineer more or less)

I hope this clarifies what I meant and **thank you** for answering. Please let me know if my comment makes sense and is aligned with @virkt25 request.

[...]

PR #2171, opened 2018-12-18
fix(cli): allow base class exist for model config option

dhmlau:
When using `lb4 model --config` option, if the json file I passed in contains `base` attribute/class, it gives an error saying `Model was not found`.   In a previous version of `@loopback/cli`, it was working with the following json as stated in https://github.com/strongloop/loopback-next/issues/1517#issuecomment-433522659:
```
{
    "name": "Customer",
    "base": "Entity",
    "properties": {
        "id": {
            "type": "number",
            "id": true,
            "required": false
        },
        "name": {
            "type": "string",
            "length": 200,
            "id": false,
            "required": true
        }
    }
}
```

It is caused by the check being incorrect in this line:  https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/model/index.js#L131

cc @b-admike 

Fixes https://github.com/strongloop/loopback-next/issues/2257

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -80,6 +80,13 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       description: 'A valid based model',
     });
 
+    // The base class can be specified:
+    // 1. From the prompt
+    // 2. using the --base flag
+    // 3. in the json when using the --config flag
+    // This flag is to indicate whether the base class has been validated.
+    this.isBaseClassChecked = false;
+
     return super._setupGenerator();
   }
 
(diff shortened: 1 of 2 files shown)


raymondfeng:
CI is failing.

dhmlau:
Yes, 5 test cases are failing.  Trying to resolve it locally first.  Marking it as WIP. Thanks.

packages/cli/generators/model/index.js
```diff
@@ -154,8 +164,30 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       },
     ])
       .then(props => {
-        if (typeof props.modelBaseClass === 'object')
-          props.modelBaseClass = props.modelBaseClass.value;
+        if (!this.isBaseClassChecked) {
```
bajtos: Nitpick: use reverse logic and early return to reduce indentation level and make the code easier to read.

```suggestion
        if (this.isBaseClassChecked) return;
```

packages/cli/generators/model/index.js
```diff
@@ -154,8 +164,30 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+          if (typeof props.modelBaseClass === 'object')
+            props.modelBaseClass = props.modelBaseClass.value;
+          // Find whether the specified base class is one of the available base
+          // class list
+          if (
+            !props.modelBaseClass &&
+            !this.isValidBaseClass(
```
bajtos: Please extract this long multi-line condition into a helper variable.

```ts
const isValidBase = this.isValidBaseClass(
  // ...
);

if (!props.modelBaseClass && !isValidBase) {
  // etc.
}
```

See also https://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

dhmlau:
@slnode test please

dhmlau:
@slnode test please

PR #2179, opened 2018-12-20
fix: make the tslint-config repo public

jannyHou:
## Description

Turn on the public config for module tslint-config

## Checklist


jannyHou:
There is a version check doesn't match after the new release. fixing it.

jannyHou:
@dhmlau Oh you are right, the version doesn't match because the example versions are updated in the package.json file, but their publish failed. This is caused by the failed installation of `@loopback/tslint-config`.

We need to merge this PR first.

Issue #1844, opened 2018-10-12
[Docs] Example config.json for `lb4 model` shows the app-config.json


dhmlau:
## Description / Steps to reproduce / Feature proposal
In the docs https://loopback.io/doc/en/lb4/Model-generator.html, the example `--config` option shows the `lb4 app` example.  

## Expected Behavior
We should show the model-config.json that this command is expecting.

dhmlau:
- example for `lb4 model` (cross posting from https://github.com/strongloop/loopback-next/issues/1517#issuecomment-433522659)
```
{
    "name": "Customer",
    "base": "Entity",
    "properties": {
        "id": {
            "type": "number",
            "id": true,
            "required": false
        },
        "name": {
            "type": "string",
            "length": 200,
            "id": false,
            "required": true
        }
    }
}
```
- example for `lb4 controller` (cross posting from https://github.com/strongloop/loopback-next/pull/2181#issuecomment-449485669 from @elv1s ): 
```
{
  "name": "Widget",
  "modelName": "Widget",
  "controllerType": "REST Controller with CRUD functions",
  "repositoryName": "WidgetRepository",
  "idType": "string",
  "httpPathName": "/widget"
}
```

bajtos:
> ```
>  "controllerType": "REST Controller with CRUD functions",
> ```

This looks like an unintentional bug to me. We should use a languange-independent key, e.g. `REST` or `RestController` instead of a full name that will be (most likely) subject to localization into different languages in the future.

yanamg7:
What parameters I have to use to create model with annotation of the property like: 
"properties": {
          "id": {
            "type": "string",
            "mcpAnnotations": [
              {
                "destinationClass": "defClass",
                "destinationField": "name",
                "prefix": "/Common/profile_for_"
              }
            ]
          }
}
In the model I get: 

@property({
    type: 'string',
    mcpAnnotations: [object Object],
    required: true,
    id: true,
  })
  id: string;
What wrong I do?

bajtos:
> if lb4 JSON model property has an object or object array it doesn't work. For example:

@yanamg7 This is a different problem, possibly related to https://github.com/strongloop/loopback-next/issues/2953. Please open a new issue to discuss it.

yanamg7:
The issue about array of objects into the property is already fixed. Now this is just on one line instead of columns exactly related to  #2953.

PR #2181, opened 2018-12-21
fix(docs): fix example for using --config option

dhmlau:
According to [CLI docs](https://loopback.io/doc/en/lb4/Application-generator.html), one of the ways to use `--config` is:
```
lb4 app --config {"name":"my-app"}
```
I've tried it on Mac and it didn't work.  It works for me with quotes added, i.e. 
```
lb4 app --config '{"name":"my-app"}'
```

### DIFF
--- a/docs/site/includes/CLI-std-options.md
+++ b/docs/site/includes/CLI-std-options.md
@@ -20,7 +20,7 @@
 For example,
 ```sh
 lb4 app --config config.json
-lb4 app --config {"name":"my-app"}
+lb4 app --config '{"name":"my-app"}'
 cat config.json | lb4 app --config stdin
 lb4 app --config stdin < config.json
 lb4 app --config stdin << EOF


elv1s:
Is there an example anywhere of a more complete config that can be passed in?

dhmlau:
@elv1s, I've tried the `lb4 app` and `lb4 model` ones, will need to dig up the example.  Any one in particular you're looking for? 
There's some issue for using `lb4 model --config`.  I'm working on a PR https://github.com/strongloop/loopback-next/pull/2171 (still need to add some tests) to fix it.

elv1s:
@dhmlau `lb4 controller` seems to generate a controller with only a constructor. Not sure if this is a bug or if the config is invalid?

dhmlau:
@elv1s , it seems to be a bug, I've tried:
```

**Comment by elv1s** [comment]
Actually I found this config works. The controller type needs to be verbose "REST Controller with CRUD functions"

```json
{
  "name": "Widget",
  "modelName": "Widget",
  "controllerType": "REST Controller with CRUD functions",
  "repositoryName": "WidgetRepository",
  "idType": "string",
  "httpPathName": "/widget"
}
```

dhmlau:
Thanks @elv1s! We should put it as part of this task: https://github.com/strongloop/loopback-next/issues/1844.

Issue #1298, opened 2018-05-01
Description of Options -- loopbackBuild / mocha


virkt25:
## Description / Steps to reproduce / Feature proposal

From #1206 

> What is loopbackBuild?

Related Note:
> No description in the CLI tool.

---

> What is `Enable mocha`

Related Note:
> Asked if this means that it will generate the tests for the project for him.

## Acceptance Criteria

- [ ] Add a description to the Todo tutorial explaining what CLI options we are asking them to select mean such as `loopbackBuild` / `mocha`
- [ ] A good / easy to understand description of options in CLI tool when prompted to select them
    - `loopbackBuild` sets up typescript compiler with other features such as (copy non-ts files to dist)
    - `mocha` -- installs mocha as a dependency (not auto-generate tests)
- [ ] Create an issue based on the discrepency between the CLI tooling and the documentation for it (if any)

### Discussion Point
- Should we keep `loopbackBuild` as an option or just remove the option entirely? ... `lb-tsc` provides copying of `non-ts` files which is essential for `@loopback/boot`'s declarative support!
  - we have decided on __keeping the option__

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

raymondfeng:
1. +1 to make the options more descriptive.
2. -1 to remove the possibility to opt in build scripts from `@loopback/build`.
3. We need to find a way to make it easy to understand the prompts.

dhmlau:
Besides making the prompt more descriptive, perhaps add in more details in the docs as well? http://loopback.io/doc/en/lb4/Application-generator.html

bajtos:
I am of the opinion that less options are better, especially for novice-to-intermediate users, and that our CLI should be tailored for the majority of our users base that is happy to follow our recommended setup.

As for advanced users that understand reasoning behind our recommended setup and ramifications of disabling things like loopbackBuild and tslint, if they understand so much about the tooling, then surely they are able to tweak project infrastructure manually, without our CLI tooling. In fact, I'd expect that once our users outgrow the default infrastructure scaffolded by our CLI, they will need custom configuration that's different from the current options offered by CLI anyways.

I would like to see only one option offered by our CLI tooling:

 - _Do you want to use JavaScript or TypeScript?_ 

Everything else should be driven by the answer to this single question.

 - JavaScript: enable eslint, prettier, mocha.
 - TypeScript: enable tslint, prettier, mocha, loopbackBuild

[...]

PR #2186, opened 2018-12-21
fix(cli): add descriptions to features

nabdelgadir:
Connect to https://github.com/strongloop/loopback-next/issues/1298.

Add descriptions from https://github.com/strongloop/loopback-next/issues/1298#issuecomment-449027445 to features in the CLI when creating an app so users can understand what the features do. 

## Checklist

### DIFF
--- a/docs/site/Getting-started.md
+++ b/docs/site/Getting-started.md
@@ -42,13 +42,13 @@ Answer the prompts as follows:
 ? Project root directory: (getting-started)
 ? Application class name: StarterApplication
 ? Select features to enable in the project:
-❯◉ Enable tslint
- ◉ Enable prettier
- ◉ Enable mocha
- ◉ Enable loopbackBuild
- ◉ Enable vscode
- ◉ Enable repositories
- ◉ Enable services
+❯◉ Enable tslint: add a linter with pre-configured lint rules
+ ◉ Enable prettier: add new npm scripts to facilitate consistent code formatting
+ ◉ Enable mocha: install mocha to assist with running tests
+ ◉ Enable loopbackBuild: use @loopback/build helpers (e.g. lb-tslint)
+ ◉ Enable vscode: add VSCode config files
+ ◉ Enable repositories: include repository imports and RepositoryMixin
+ ◉ Enable services: include service-proxy imports and ServiceMixin
 ```
 
 ### Starting the project
--- a/packages/cli/lib/project-generator.js
+++ b/packages/cli/lib/project-generator.js
@@ -6,6 +6,7 @@
 'use strict';
 const BaseGenerator = require('./base-generator');
 const utils = require('./utils');
+const chalk = require('chalk');
 
 module.exports = class ProjectGenerator extends BaseGenerator {
   // Note: arguments and options should be defined in the constructor.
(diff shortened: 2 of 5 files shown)


docs/site/Getting-started.md
```diff
@@ -42,13 +42,13 @@ Answer the prompts as follows:
...
 ? Application class name: StarterApplication
 ? Select features to enable in the project:
-❯◉ Enable tslint
- ◉ Enable prettier
- ◉ Enable mocha
- ◉ Enable loopbackBuild
- ◉ Enable vscode
- ◉ Enable repositories
- ◉ Enable services
+❯◉ Enable tslint: adds a linter with pre-configured lint rules
```
bajtos: I think it would be better to use the same tense in both parts of the "sentence", i.e.

```
Enable tslint: add a linter with pre-configured lint rules
```

docs/site/Getting-started.md
```diff
@@ -42,13 +42,13 @@ Answer the prompts as follows:
...
+ ◉ Enable prettier: installs prettier for consistent code formatting
```
bajtos: I think this is not only installing prettier, but also adding new npm scripts and hooking them into `npm test` and `npm lint`.

Can we explain the feature at high-level, maybe explain why users may want to use this feature. Don't explain implementation details.

docs/site/Getting-started.md
```diff
@@ -42,13 +42,13 @@ Answer the prompts as follows:
...
+ ◉ Enable mocha: installs mocha as a dependency
```
bajtos: Ditto.

docs/site/Getting-started.md
```diff
@@ -42,13 +42,13 @@ Answer the prompts as follows:
...
+ ◉ Enable loopbackBuild: sets up TypeScript compiler with other features
```
bajtos: I believe we are always setting TypeScript compilation. This flag determines whether we use `@loopback/build` helpers or call `tsc` directly. IIRC, this affects other build tooling like `mocha` (vs. `lb-mocha`) and `tslint` (vs. `lb-tslint`) too.

packages/cli/lib/project-generator.js
```diff
@@ -177,11 +190,11 @@ module.exports = class ProjectGenerator extends BaseGenerator {
       },
     ];
     return this.prompt(prompts).then(props => {
-      const settings = props.settings || choices.map(c => c.name);
+      const settings = props.settings || choices.map(c => c.name.split(':')[0]);
```
bajtos: I find this string manipulation rather hacky. Isn't the original option name available as `c.key`?

packages/cli/lib/project-generator.js
```diff
@@ -177,11 +190,11 @@ module.exports = class ProjectGenerator extends BaseGenerator {
...
       const features = choices.map(c => {
         return {
           key: c.key,
-          value: settings.indexOf(c.name) !== -1,
+          value: settings.indexOf(c.name.split(':')[0]) !== -1,
```
bajtos: Ditto.

packages/cli/lib/project-generator.js
nabdelgadir: It has to include the word `Enable` but I'll replace it with `Enable ${c.key}` so it's less hacky.

packages/cli/lib/project-generator.js
```diff
@@ -177,11 +194,11 @@ module.exports = class ProjectGenerator extends BaseGenerator {
       },
     ];
     return this.prompt(prompts).then(props => {
-      const settings = props.settings || choices.map(c => c.name);
+      const settings = props.settings || choices.map(c => `Enable ${c.key}`);
```
bajtos: I find this suspicious. Here is what the docs is saying:

https://www.npmjs.com/package/inquirer#question
> `choices`: `(Array|Function)` Choices array or a function returning a choices array. If defined as a function, the first parameter will be the current inquirer session answers. Array values can be simple strings, or objects containing a `name` (to display in list), a `value` (to save in the answers hash) and a `short` (to display after selection) properties.

I think we should change the way we are building choices as follows:

 - Set `value` to `f.name` (e.g. value = `tslint`)
 - Keep `name` as it is now
 - Add `short` set to `Enable ${f.name}`

With those changes in place, I believe that `props.settings` will use the `f.name` as the key and no string manipulation (`Enable ${c.key}`) will be needed in this anonymous function.

What do you think?

nabdelgadir: I think that's a better approach because it's clearer. I'll change it to this one. Thank you!

nabdelgadir: On second thought, should we keep `value` to stay the same as it is now? It's currently a boolean to determine whether or not to enable a setting. We can use `short` in ` const settings = props.settings || choices.map(c => c.short);` as it's already `Enable ${c.key}`, but what do you think?

Issue #2170, opened 2018-12-18
[CLI] cli datasource should show the path of the database file


hacksparrow:
## Description / Steps to reproduce / Feature proposal
The memory datasource should show the path of database file it created.

## Current Behavior

```sh
$ lb4 datasource
? Datasource name: memory
? Select the connector for memory: In-memory db (supported by StrongLoop)
? window.localStorage key to use for persistence (browser only):
? Full path to file for persistence (server only): /database/file
   create src/datasources/memory.datasource.json
   create src/datasources/memory.datasource.ts
Datasource memory was created in src/datasources/
```
## Expected Behavior

```sh
$ lb4 datasource
? Datasource name: memory
? Select the connector for memory: In-memory db (supported by StrongLoop)
? window.localStorage key to use for persistence (browser only):
? Full path to file for persistence (server only): git clean -fd
   create src/datasources/memory.datasource.json
   create src/datasources/memory.datasource.ts
Database file created at /database/file
Datasource memory was created in src/datasources/
```

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

hacksparrow:
We do :)

hacksparrow:
Printing the location of the database file was a bad idea. The path can be very long and displaying it on the terminal will not look good. Instead, I have made some changes to how the file is generated and have updated the doc.

PR: https://github.com/strongloop/loopback-next/pull/2190

PR #2190, opened 2018-12-24
feat: add clarity about memory connector's database file

hacksparrow:
Adds clarity about the behavior and documentation of the database
file generated for the "In memory database" connector.

Fixes https://github.com/strongloop/loopback-next/issues/2170.

## Checklist


packages/cli/generators/datasource/index.js
```diff
@@ -287,6 +288,27 @@ module.exports = class DataSourceGenerator extends ArtifactGenerator {
...
+          this.artifactInfo.settings.file = './' + generateRandomFileName();
+        } else {
+          this.artifactInfo.settings.file = suggestedFile;
+        }
+      }
+      const dbFilePath = path.resolve(this.artifactInfo.settings.file);
+      fse.ensureFileSync(dbFilePath);
+    }
+
+    function generateRandomFileName() {
```
dhmlau: I'm wondering instead of just a random file, would it be better to have the name prefixed by the artifact name? Just a thought.

docs/site/DataSource-generator.md
```diff
@@ -44,6 +44,17 @@ The tool will prompt you for:
 - **Connector configuration.** If the connector is not a custom connector, the
   CLI will prompt for the connector configuration information.
 
+In case of "In-memory db" connector, you will be prompted for these additional
```
dhmlau: since this is applicable for LB3 as well, and the LB3 docs has a memory connector page, would it make more sense to put it in https://loopback.io/doc/en/lb3/Memory-connector.html?

packages/cli/generators/datasource/index.js
hacksparrow: Good suggestion.

docs/site/DataSource-generator.md
hacksparrow: The LB3 doc has the info already.

The behavior of the memory connector in `lb4 connector` and `lb connector` are different, so I guess this PR should suffice. Differences:

1. In LB4, the database directory will be created by the generator. In LB3, the running app will throw an error.
2. In LB4, if no name is specified, the generator will generate one based on the name of the connector and create the file in the application directory. In LB3, the behavior is "unknown" and undocumented.

docs/site/DataSource-generator.md
```diff
@@ -44,6 +44,17 @@ The tool will prompt you for:
...
+details:
+
+- **window.localStorage key** The name of the key for persistence in the
+  browser.
+- **Path to file for persistence** The path of the file where the connector
+  should store its data. If no path is specified, a `.db.json` file derived from
+  the name of the connector will be created in the application directory. If a
+  file already exists with such a name, a random `.db.json` file will be created
+  in the application directory.
```
bajtos: > In LB4, if no name is specified, the generator will generate one based on the name of the connector and create the file in the application directory

I consider this as incorrect behavior. As I see it, the `memory` connector should be primarily used for in-memory storage that does not survive app restarts. This is intentional, as it simplifies scenarios where the app (or perhaps a test suite) wants to start with a clean database.

IMO, we should fix `lb4 datasource` to behave the same way as `lb datasource`, i.e. set "file" to an empty value (or leave it out completely).

> In LB3, the behavior is "unknown" and undocumented.

The behavior is well-know and documented. Quoting from https://loopback.io/doc/en/lb3/Memory-connector.html:

> (...) the memory connector does not persist data after an application stops.

packages/cli/generators/datasource/index.js
```diff
@@ -287,6 +288,30 @@ module.exports = class DataSourceGenerator extends ArtifactGenerator {
     // Copy Templates
     this.fs.writeJSON(jsonPath, ds);
     this.copyTemplatedFiles(classTemplatePath, tsPath, this.artifactInfo);
+
+    if (ds.connector === 'memory') {
+      if (!this.artifactInfo.settings.file) {
```
bajtos: As I commented above, it's perfectly fine and often desirable to leave the `file` option not set.

bajtos:
I can see how the current prompts can be confusing to users new to LoopBack. I am proposing two alternative improvements for the consideration.

**Option A**

Before asking for connector-specific options, the CLI tool can print a link to our docs where users can read more about the individual options. For most connectors, this link can point to npmjs.org using the package name to build the URL, e.g. https://www.npmjs.com/package/loopback-connector-mysql. For built-in connectors like `memory` and `kv-memory`, we can add a new field to [available-connectors.json](https://github.com/strongloop/loopback-workspace/blob/master/available-connectors.json) to specify URL of the documentation.

This approach is generic and will work for all connectors, it does not require any specific overrides for the memory connector in particular.

**Option B**

When configuring the memory connector, rephrase the prompts as follows:

- Question: _Do you want to persist the data in a file? y/N_ (Default: No)
- If the answer was yes:
- Question: _file (Full path to file for persistence (server only))_
- Skip "localStorage" question, LoopBack 4 does not support running in a browser

hacksparrow:
**Option A** is convenient for us, but inconvenient for our users. As a developer, I'd want the system to cater to my intent, not ask me to go read the docs 😂

**Option B** is more intuitive.

bajtos:
> Option A is convenient for us, but inconvenient for our users. As a developer, I'd want the system to cater to my intent, not ask me to go read the docs 

Sure, I agree it's not the best user experience. On the other hand, we don't want our CLI code to have hard-coded custom prompts for every connector we support. What we want instead, is a generic framework that will allow each connector to describe their prompts in a way that allows our CLI to run them. 

While Option A is far from ideal from UX perspective, it's a solution that works for all connectors, including those maintained by the community, while adding very little (if any) maintenance overhead.

bajtos:
BTW, the two options I proposed are not mutually exclusive, we can implement them both.

Anyhow, that's just my thoughts, feel free to come up with a different solution.

bajtos:
@hacksparrow what is the status of this pull request?

hacksparrow:
Will revisit this when I get some time.

dhmlau:
@hacksparrow, are you still pursuing this PR?

hacksparrow:

Issue #2175, opened 2018-12-19
Examples tsconfig filename resolution issue /w VSCode


dougal83:
## Description / Steps to reproduce

1. Install any example via cli, e.g ''lb4 example todo-list''
2. Open folder in editor, e.g. VSCode
3. Trigger linter, e.g Open src/controllers/todo.controller.ts

(n.b. VSCode is only client tested)

## Current Behavior

Warns for example below (which is set via tsconfig.build.json):
[ts] Experimental support for decorators is a feature that is subject to change in a future release. Set the 'experimentalDecorators' option to remove this warning.

## Expected Behavior

The tsconfig rules to be applied correctly.  Renaming 'tsconfig.build.json' to 'tsconfig.json' resolves the config correctly in VSCode.

dhmlau:
@nabdelgadir, could you please take a look? Thanks.

PR #2201, opened 2018-12-27
refactor: rename tsconfig.build files in examples

nabdelgadir:
Connect to https://github.com/strongloop/loopback-next/issues/2175.

Renamed the `tsconfig.build.ts` files in the examples to `tsconfig.ts` because, when using the examples from the `lb4 example` command, they weren't extending [`tsconfig.common.json`](https://github.com/strongloop/loopback-next/blob/master/packages/build/config/tsconfig.common.json) until after the rename. They also now match the [`rpc-server`](https://github.com/strongloop/loopback-next/blob/master/examples/rpc-server/tsconfig.json) example. 

## Checklist

### DIFF
--- a/docs/site/DEVELOPING.md
+++ b/docs/site/DEVELOPING.md
@@ -510,3 +510,25 @@ configuration, it's important to verify that all usage scenarios keep working.
 5.  Test integration with supported IDEs:
 
     - [VS Code](./VSCODE.md#how-to-verify-tslint-setup)
+
+### tsconfig files
+
+In the [`loopback-next`](https://github.com/strongloop/loopback-next) monorepo,
+`TypeScript` is set up in two places:
+
+1. When using VS Code, the `TypeScript` engine views `loopback-next` as a single
+   big project.
+
+   This enables the "refactor - rename" command to change all places using the
+   renamed symbol, and also makes "go to definition" command jump to `.ts` files
+   containing the original source code. Otherwise "refactor - rename" works
+   within the same package only and "go to definition" jumps to `.d.ts` files.
+
+2. When building the monorepo, we need to build the packages individually, so
+   that one `dist` directory is created for each package.
+
+This is why we have two sets of `tsconfig` files:
+
+- At monorepo root, there is `tsconfig.json` used by VS Code.
+- Inside each package, there is `tsconfig.build.json` used by `npm run build`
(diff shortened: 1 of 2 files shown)


bajtos:
I believe this change is going to break "refactor - rename" and "go to definition" commands in VS Code. Let me explain the background. It would be great if you could capture this information in our docs, e.g. [DEVELOPING.md](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md).

Right now, our monorepo has two TypeScript setups in place.

1. Inside VS Code, we want the TypeScript engine to view loopback-next as a single big project. This enables "refactor - rename" command to change all places using the renamed symbol, and also makes "go to definition" jump to `.ts` files containing the original source code. Otherwise "refactor - rename" works within the same package only and "go to definition" jumps to `.d.ts` files.

2. When building the monorepo, we need to build the packages individually, so that one `dist` directory is created for each package.

That's why we have two sets of `tsconfig` files:
- At monorepo root, there is `tsconfig.json` used by VSCode
- Inside each package, there is `tsconfig.build.json` used by `npm run build` command.

@raymondfeng and I are looking into ways how to leverage TypeScript's Project References to allow us to have the same tsconfig for both build & VSCode, but that's a longer term work. See https://github.com/strongloop/loopback-next/pull/1636.

docs/site/DEVELOPING.md
```diff
@@ -510,3 +510,25 @@ configuration, it's important to verify that all usage scenarios keep working.
...
+
+### tsconfig files
+
+In the [`loopback-next`](https://github.com/strongloop/loopback-next) monorepo,
+`TypeScript` is set up in two places:
+
+1. When using VS Code, the `TypeScript` engine views `loopback-next` as a single
+   big project.
+
+- This enables the "refactor - rename" command to change all places using the
```
bajtos: This content does not render well.

screen shot 2019-01-08 at 16 14 58

I am proposing to indent this paragraph with four spaces, so that the text stays inside the list item.

```markdown
1. When using VS Code, the `TypeScript` engine views `loopback-next` as a single
    big project.

    This enables the "refactor - rename" command to change all places using the
    renamed symbol, and also makes "go to definition" command jump to `.ts` files
    (etc.)
```

Issue #1530, opened 2018-07-13
[CLI] `lb4 model` strict prompt


virkt25:
## Acceptance Criteria

- [ ] Add a prompt to explain / ask user if they would like the model to be `static` or not (default to current behaviour). Update template to include that when not the default.

bajtos:
Cross-posting from our [Model](https://github.com/strongloop/loopback-next/blob/master/docs/site/Model.md) docs:

To create a model that allows both well-defined but also arbitrary extra properties, you need to disable strict mode in model settings and tell TypeScript to allow arbitrary additional properties to be set on model instances.

```ts
@model({settings: {strict: false}})
class MyFlexibleModel extends Entity {
  @property({id: true})
  id: number;

  // Define well-known properties here

  // Add an indexer property to allow additional data
  [prop: string]: any;
}
```

nabdelgadir:
Closing as done https://github.com/strongloop/loopback-next/pull/2204.

PR #2204, opened 2018-12-31
feat(cli): add strict prompt to model generator

nabdelgadir:
Connect to https://github.com/strongloop/loopback-next/issues/1530.

Add a prompt to model generator that asks the user if they want to allow additional properties (disable the `strict` setting). And add `modelSettings` to `this.artifactInfo` in `ModelGenerator` that contains settings such as `strict`.

## Checklist

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -154,20 +154,42 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       },
     ])
       .then(props => {
-        if (typeof props.modelBaseClass === 'object')
+        if (typeof props.modelBaseClass === 'object') {
           props.modelBaseClass = props.modelBaseClass.value;
+        }
 
         Object.assign(this.artifactInfo, props);
         debug(`props after model base class prompt: ${inspect(props)}`);
+        return props;
+      })
+      .catch(err => {
+        debug(`Error during model base class prompt: ${err}`);
+        return this.exit(err);
+      });
+  }
+
+  async promptStrictMode() {
+    if (this.shouldExit()) return false;
+    return this.prompt([
+      {
+        name: 'allowAdditionalProperties',
+        message: 'Allow additional (free-form) properties?',
+        type: 'confirm',
+        default: false,
+        when: !this.artifactInfo.allowAdditionalProperties,
+      },
+    ])
+      .then(setting => {
+        Object.assign(this.artifactInfo, setting);
+
         this.log(
           `Let's add a property to ${chalk.yellow(
             this.artifactInfo.className,
           )}`,
         );
-        return props;
       })
       .catch(err => {
-        debug(`Error during model base class prompt: ${err}`);
+        debug(`Error during model strict mode prompt: ${err}`);
         return this.exit(err);
       });
   }
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -66,6 +66,7 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
     ];
 
     this.artifactInfo.properties = {};
+    this.artifactInfo.modelSettings = {};
     this.propCounter = 0;
 
     this.artifactInfo.modelDir = path.resolve(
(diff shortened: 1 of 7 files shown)


packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -5,7 +5,11 @@ import {model, property} from '@loopback/repository';
 import {<%= modelBaseClass %>} from '.';
 <% } -%>
 
+<% if(disabledStrictMode) { -%>
```
jannyHou: I think a more general way to get the model settings from the CLI is passing in a settings object, like

```ts
 <% if(settings) { -%>
 @model(<%= settings %>)
 <% } else { -%>
 @model()
 <% } -%>
```

just a suggestion :) can do it in a separate PR.

packages/cli/generators/model/index.js
```diff
@@ -154,20 +154,43 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+        type: 'confirm',
+        default: false,
+        when: !this.artifactInfo.disabledStrictMode,
+      },
+    ])
+      .then(setting => {
+        Object.assign(this.artifactInfo, setting);
+
         this.log(
           `Let's add a property to ${chalk.yellow(
```
jannyHou: Any reason why we log this message here? I feel it should be in the `promptPropertyName` function. While I also noticed that this code was in the `promptModelBaseClass` function, so I am a little bit confused...

packages/cli/generators/model/index.js
```diff
@@ -154,20 +154,43 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+    return this.prompt([
+      {
+        name: 'disabledStrictMode',
+        message: 'Disable strict mode for this model?',
+        type: 'confirm',
+        default: false,
+        when: !this.artifactInfo.disabledStrictMode,
+      },
+    ])
+      .then(setting => {
```
jannyHou: Correct me if wrong:
I assume the setting should be an object with the `disabledStrictMode` property like:
`{disabledStrictMode: true}`

I prefer to add a property called `modelSettings` in `this.artifactInfo` in case we support more configurable setting attributes in the future, like:

```ts
artifactInfo: {
  modelSettings: {
    disabledStrictMode: true
  }
}
```

And btw, why do we return the `setting` at the end of the prompt 🤔

packages/cli/generators/model/index.js
nabdelgadir: It's so that it appears in this order: 
```
lb4 model
? Model class name: model
? Please select the model base class Entity (A persisted model with an ID)
? Disable strict mode for this model? No
Let's add a property to Model
```
instead of:
```
lb4 model
? Model class name: model
? Please select the model base class Entity (A persisted model with an ID)
Let's add a property to Model
? Disable strict mode for this model? No
```

packages/cli/generators/model/index.js
nabdelgadir: Thanks, I'll update it accordingly and remove the return!

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -5,7 +5,11 @@ import {model, property} from '@loopback/repository';
...
+@model({settings: {strict: false}})
+<% } else { -%>
 @model()
+<% } -%>
 export class <%= className %> extends <%= modelBaseClass %> {
```
bajtos: Please note it's not enough to disable strict mode in model setting, the TypeScript class must be changed to allow additional properties too.

Cross-posting the relevant part from https://github.com/strongloop/loopback-next/issues/1530#issuecomment-434203916

```ts
  // Add an indexer property to allow additional data
  [prop: string]: any;
```

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -18,6 +22,15 @@ export class <%= className %> extends <%= modelBaseClass %> {
   <%= key %><%if (!val.required) {%>?<% } %>: <%= val.tsType %>;
 
 <% }) -%>
+
```
bajtos: I think this will create an empty line in the generated output, even when `allowAdditionalProperties` was set to `false`. Please move this empty line inside `<% if(allowAdditionalProperties) { -%>` branch.

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -18,6 +22,15 @@ export class <%= className %> extends <%= modelBaseClass %> {
...
+<% if(allowAdditionalProperties) { -%>
+  // Define well-known properties here
+
+
```
bajtos: Do we need two empty lines here? I think prettier will remove the second line anyways, won't it? Can we keep a single empty line in this template?

Issue #2027, opened 2018-11-13
[Spike] Token based authentication in API Explorer


dhmlau:
**Timebox to 5 days**
### User experience
Go to API Explorer, login and set the token so that API Explorer can use the token for subsequent request. 

Originated from https://github.com/strongloop/loopback-next/issues/1035#issuecomment-438387429

### Acceptance Criteria
- [x] Investigate whether swagger ui support token based authentication
- [x] and what changes/workaround needed in order to make it work
- [x] create a list of follow-up tasks (user stories) describing changes we need to make in order to make token based authentication work in our API Explorer

### References
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#jwt-bearer-sample
- https://github.com/swagger-api/swagger-ui/issues/818#issuecomment-113185459
- https://swagger.io/docs/specification/authentication/bearer-authentication/
- https://github.com/swagger-api/swagger-ui/issues/5004

hacksparrow:
Conclusion: Swagger UI provides UI for setting the token, we don't have to create additional UI. Links to example and discussion - https://github.com/strongloop/loopback-next/pull/2210#issuecomment-451866876.

bajtos:
I disagree with closing this story. A spike it's done when there is a list of follow-up stories describing what needs to happen next.

In this particular case, we need user stories describing changes that we need to make to enable token-based authentication in API Explorer rendered for LB4 applications. Based on the discussion in #2210, I think we will need to describe `security` schemas in the OAI spec generated for our apps, but that's something to figure out as part of this spike.

lygstate:
when this got to be fixed

jotamora:
Hello, 

Anybody can said to me if there is someone working in that? If there is someone working, anybody knows, more less, when can be ready? 

Thanks.

shendkardevesh:
hi @bajtos,
i tried to set headers by  @param.header.string('token') token?: string in controller which gives me option to enter token for a api, which get's set in header.
which in ui gives me -
![image](https://user-images.githubusercontent.com/19186558/62427250-ed960300-b70d-11e9-87e3-869f87d7f86a.png)

doing this is a correct way or we need something exactly same as in swagger-ui.

nflaig:
Hi @shendkardevesh,

right now I also face the problem that the explorer is completely unusable for me because there is no option to set headers so I have to use Postman instead.

[...]

Issue #1997, opened 2018-11-08
Groundwork: Access tokens, login & logout


nabdelgadir:
## Description
Step 2 from https://github.com/strongloop/loopback-next/issues/1035#issuecomment-429774436.

Define a new `AccessToken` model and a relation to `Customer` (a customer has many tokens, a token belongs to a customer).

Implement login and logout, expose them via REST API. Let's create a new Controller class for these endpoints!

The goal of this step is to show a reference implementation in our Shopping App and prepare ground for the actual auth-related work in the next iterations. Later, we may decide to extract AccessToken & Password models together with the relevant repositories and Controller classes into a shareable LB4 component (e.g. `@loopback/local-credentials`).

Ideally, there should be a documentation and/or a blog-post and/or a reference implementation to make it easier for LB4 users to implement similar functionality in their project.

Previous step: https://github.com/strongloop/loopback-next/issues/1996
Next step: https://github.com/strongloop/loopback-next/issues/1998

David-Mulder:
Is the idea here that tokens are stored in the database or to use JWTs?

dhmlau:
@David-Mulder , we're planning to implement a simple authentication provider that uses JWT access token under the hood.

dhmlau:
Extracted from the discussion notes captured in https://github.com/strongloop/loopback-next/issues/1035#issuecomment-438387429:

## Acceptance Criteria
- [ ] Implement simple authentication provider that users JWT access token under the hood, and contain hardcoded list of user credentials (could be in-memory in the shopping card example). It has 2 APIs:
   - [ ] login: input: userid+password.  output: Object that contains "yes/no", JWT access token, and user info. 
   - [ ] logout: invalidate the access token
- [ ] Modify the shopping app accordingly to make use of it
- [ ] docs/blog post, if applicable

bajtos:
> 1. Authentication package needs to be refactored to extract the minimal contract for authentication providers
> 2. Implement simple authentication provider that users JWT access token under the hood, and contain hardcoded list of user credentials (could be in-memory in the shopping card example). 
> 3. Modify the shopping app accordingly to make use of it

I am little bit concerned about practical aspects of this approach. My original vision was to start from outside (the example app) and drive the internal design & implementation based on what we need to provide to API consumers.

[...]

Issue #1035, opened 2018-02-22
Authentication and Authorization extension points


dhmlau:
Per discussion with @raymondfeng @kjdelisle:

Need extension point implementation and extension interfaces for authentication and authorization
  - make sure we have the infrastructure for users to build extensions
  - if we have solid reference implementation for Passport strategy, it is not a must to provide other reference implementation for GA

  - Extension: Authentication
    - [OAuth2 Strategy support for the authentication package](https://github.com/strongloop/loopback-next/issues/230) <-- not a must
    - [Spike: Use express constructs / enable express shimming for Authentication](https://github.com/strongloop/loopback-next/issues/191)
    - [Test other Passport Strategies](https://github.com/strongloop/loopback-next/issues/160)
  - Extension: Authorization
    - [Extension: Add Authorization component](https://github.com/strongloop/loopback-next/issues/538)

## User Experience
- For a given controller method, we should be able to use decorators to express the authorization requirement. e.g. we allow / deny certain roles. 
- need to have the ability to decorate the methods to indicate the scope of the resource access.  i.e. group the methods for access.  e.g. write access for Order. 
   - describe your protection scope for the method
- example: when cancelling order, only the customer and customer rep can cancel. 

- can possibly use json / yaml files to define the access rules, besides using decorators.  
- Need infrastructure to help us to make the yes/no decision - decision could be no-go, consult external provider (e.g. enterprise might have a centralized authorization management system).  This can be expressed in extension points to plug in other authorization providers.
- from admin perspective, we need a way to specify the role mapping. i.e. which user belongs to which role.  -- 1) static mapping table.  2) some role has to be dynamically determined per request, e.g. cancelling order. need to figure out for the given orderid, it matches with the customerid to place the order. 
- authorization has the dependencies on authentication, because we need to know who is making the call.   Two pieces of identity 1) the user itself, 2) application id making the request <--- think FB   app. 
- also look at what we have in LB3 for functional parity
- we also allow developers to write code to override our authorization decisions (added business logics in the controller)

### Questions to answer
**Timeboxed max to 2 weeks**
- mockup decorators to show what metadata is required for controller/method
- need to understand the interface for authorization provider
- need to understand the roles (for grouping users)/scope (for grouping resources)
- what customization we allow developers to do at the controller level
- need to understand what json/yaml externalization for access rules and resource definitions and other metadata
- need to understand what tooling to provide to harness the authorization requirement
- identify the dependencies to the authentication component

[...]

PR #2210, opened 2019-01-03
[WIP] feat: add access token UI

hacksparrow:
Add UI for specifying the access token in API Explorer. Also makes additional UI improvements.

**Old**
image

**New**
image

### DIFF
--- a/packages/rest-explorer/templates/index.html.ejs
+++ b/packages/rest-explorer/templates/index.html.ejs
@@ -13,6 +13,7 @@
       box-sizing: border-box;
       overflow: -moz-scrollbars-vertical;
       overflow-y: scroll;
+      font-family: sans-serif;
     }
 
     *,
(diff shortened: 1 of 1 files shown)


bajtos:
I am confused. How do you envision to use the access token set in the UI? In your pull request, I don't see how the value is passed to the underlying swagger-ui or swagger-js implementation. 

Also note that by default, LoopBack 4 applications don't have authentication enabled, they don't accept any access tokens!

bajtos:
IMO, we need to implement first few parts of Authentication epic https://github.com/strongloop/loopback-next/issues/1035, most notably https://github.com/strongloop/loopback-next/issues/1997, before it makes sense to work on API Explorer enablement.

hacksparrow:
> I don't see how the value is passed to the underlying swagger-ui or swagger-js implementation.

Work in progress.

After discussing with you on Slack, looks like this is indeed early.

sbacem:
I think is better to use like this in SwaggerUI

capture
 
and 

capture2

bajtos:
> I think is better to use like this in SwaggerUI

+1 to use swagger-ui's built-in controls and don't reinvent the wheel.

hacksparrow:
@sbacem is that an in-built feature in SwaggerUI? How do you bring it up?

sbacem:
@hacksparrow you can see this [https://editor.swagger.io](https://editor.swagger.io/?_ga=2.136414640.1865821691.1546850370-1844356384.1546850370)

In first step is to define your security strategy

```
securityDefinitions:
  petstore_auth:
    type: "oauth2"
    authorizationUrl: "http://petstore.swagger.io/oauth/dialog"
    flow: "implicit"
    scopes:
      write:pets: "modify pets in your account"
      read:pets: "read your pets"
  api_key:
    type: "apiKey"
    name: "api_key"
    in: "header"
```
After this  you can use this definition in path like this 
```
      security:
      - petstore_auth:
        - "write:pets"
        - "read:pets"

```

bajtos:
Thank you @sbacem for an example of how to configure auth/auth in OpenAPI. The full spec can be found here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#securitySchemeObject

Here are the valid auth types: "apiKey", "http", "oauth2", "openIdConnect". We have been using "apiKey" type in LoopBack 3.x, I think that's the direction we are investigating for LB4 too.

cc @jannyHou

hacksparrow:
Thanks @sbacem!

raymondfeng:
As we improve the authentication/authorization support, security related metadata should be generated into the OpenAPI spec.

hacksparrow:
Closing this as we won't need to work on additional UI.

shendkardevesh:
hi , confuse how to get ui changed for accepting authentication key

dougal83:
@shendkardevesh The UI is not ready yet, for implementation progress see issue: https://github.com/strongloop/loopback-next/issues/2027 

You can use a tool like [Postman](https://www.getpostman.com/) to query the API with an access token in the meantime.

PR #2218, opened 2019-01-07
fix: optional auth metadata

jannyHou:
### Description

static endpoint like the homepage in the shopping example doesn't have a controller, thus the metadata injected for the auth action should be optional.
Related to https://github.com/strongloop/loopback4-example-shopping/pull/26

## Checklist

### DIFF
--- a/packages/authentication/test/unit/providers/auth-metadata.provider.unit.ts
+++ b/packages/authentication/test/unit/providers/auth-metadata.provider.unit.ts
@@ -65,6 +65,17 @@ describe('AuthMetadataProvider', () => {
         );
         expect(authMetadata).to.be.undefined();
       });
+
+      it('returns undefined when the class or method is missing', async () => {
+        const context: Context = new Context();
+        context
+          .bind(CoreBindings.CONTROLLER_METHOD_META)
+          .toProvider(AuthMetadataProvider);
+        const authMetadata = await context.get(
+          CoreBindings.CONTROLLER_METHOD_META,
+        );
+        expect(authMetadata).to.be.undefined();
+      });
     });
   });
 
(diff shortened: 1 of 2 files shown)


bajtos:
> static endpoint like the homepage in the shopping example doesn't have a controller, thus the metadata injected for the auth action should be optional.

How about handler-based routes? There is not controller name & method name for these routes too, but I think there should be a way how to invoke authentication for them.

```ts
const spec: OperationObject = { /*...*/ };
function greet(name: string) {
  return `hello ${name}`;
}

app.route('get', '/', spec, greet);
```

This is just something to consider, no need to fix this as part of this pull request.

The changes looks reasonable to me when I consider them as a quick fix, please add some tests to verify.

bajtos:
@jannyHou based on the comments above, I think this pull request is going in the right direction. The remaining task is to add tests to verify the changes.

packages/authentication/test/unit/providers/auth-metadata.provider.unit.ts
```diff
@@ -65,6 +65,17 @@ describe('AuthMetadataProvider', () => {
         );
         expect(authMetadata).to.be.undefined();
       });
+
+      it('returns undefined when the class or method is missing', async () => {
```
b-admike: Can we separate this test into two? i.e. one for class missing and the other class defined but method missing? or is that not useful?

jannyHou: @b-admike The method lives in the controller, so if the controller class is missing then the method will definitely be missing. Which means the only valid scenario is no controller class + no method.

What we could improve in the next PR is the scenario that Miroslav brings in his comment:

```ts
const spec: OperationObject = { /*...*/ };
function greet(name: string) {
  return `hello ${name}`;
}

app.route('get', '/', spec, greet);
``` 
And as Raymond said, the auth metadata could be included in the OpenAPI spec.

PR #2226, opened 2019-01-09
docs: use markdown image tags

hacksparrow:
Use Markdown image tags for LB4 docs.

The `{% include image.html file="lb4/10000000.png" alt="" %}` format is a legacy remnant from the time when LoopBack docs were hosted using Confluence. 

I considered the possibility of replacing all the instances of `{% include image.html }` tag with Markdown tags, but they are way too may in older docs. Replacing them may have unforeseen side-effects, so left them alone.

Good thing is, there were only three instances in the LB4 docs. 

## Checklist

### DIFF
--- a/docs/site/Preparing-the-API-for-consumption.shelved.md
+++ b/docs/site/Preparing-the-API-for-consumption.shelved.md
@@ -37,11 +37,11 @@ environment to test the API endpoints defined by the raw spec found at
 flavour at <http://localhost:3000/openapi.yaml>
 " %}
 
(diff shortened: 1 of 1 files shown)


dhmlau:
Quick question: we shouldn't be using `{% include image.html }` anymore?

hacksparrow:
Not for new content in LB4 docs. `{% include image.html }` adds additional markup and styling, which I don't think need anymore at all.

Getting rid of the old tag altogether is a problem because we cannot detect the images it will break, without having to go through all image pages; there are too many.

Issue #2224, opened 2019-01-08
[Spike] Drop LB3 model files into LB4 app


bajtos:
The migration approaches described in #485 and #1849 have a major problem: they require developers to migrate everything in one big change, there is no way how to migrate in smaller self-contained steps. I would like to look into ways how to make the migration more incremental.

**Background**

Essentially, LB3 consists of three parts: `loopback-datasource-juggler` providing ORM (data modelling and database access), `strong-remoting` providing REST API layer, and `loopback` glueing all things together. In LB4, we are still using the same juggler under the hood. `strong-remoting` was replaced with `@loopback/rest` and the "glue" has been completely rewritten.

**The idea**

Carry over LB3 models (`.json` and `.js` files) into LB4. Use juggler v4 under the hood, there were very little changes from LB3 there. Write a compatibility layer that will process strong-remoting metadata and build LB4 controllers and/or handler routes to expose LB3-flavored models via the new REST API layer.

**The envisioned user experience**

- Take the existing LB3 app
- Create a new LB4 project, or perhaps modify existing LB3 project infrastructure to LB4
- Add `@loopback/compat-v3` extension to the LB4 app
- Done. All models are exposed via LB4 app.

**The plan**

Things we will need to implement (or copy from LB3):
- registry of models (`app.registry`)
- base models (`Model` and `PersistedModel`)
- parts of `loopback-boot` responsible for model definition & registration

Things to implement anew:
- Code converting strong-remoting metadata into LB4 routes. We may leverage `loopback-swagger` to convert strong-remoting metadata into OpenAPI spec that's used as the remoting format by LB4.
 
**Things that will not work (in the first version)**

- Authorization and related built-in models like User and AccessTokens
- Boot scripts
- More items may be discovered (and documented!) during the spike

**Task list**

- [x] Registry of models and datasources, configure the model in the app
- [x] Infrastructure for defining remote methods (strong-remoting metadata), define few CRUD endpoints
- [x] Expose remote methods via `@loopback/rest`
- [x] Boot models from JSON + JS files
- [x] Example app: migrated [loopback-getting-started](https://github.com/strongloop/loopback-getting-started)
- [x] Summarize the work, write a high-level description of the PoC implementation.
- [x] Open a pull request to discuss the findings: https://github.com/strongloop/loopback-next/pull/2274
[...]

Issue #2090, opened 2018-11-28
[CLI] Generate full REST API (model+repository+controller) from model schema


orshlom:
## Description / Steps to reproduce / Feature proposal
The intention of this proposal is to create a fully functional REST API with ease imported from a schema and using an existing CLI tool `lb4 openapi`.

## Current Behavior
Currently, the OpenAPI generator `lb4 openapi` is missing several functionalities:

- The command creates only partial properties information for models listed in the `yaml` schema.
For example:
```
// IN CUSTOMER.JSON MODEL FILE (LB3):
...
"properties": {
    "id": {
        "type": "string",
        "id": true,
        "description": "Customer ID",
        "defaultFn": "uuid",
        "postgresql": {
            "owner": "public",
            "tableName": "customer",
            ...
        }
    ...
}
...
 
// TRANSLATES INTO THIS IN CUSTOMER.MODEL.TS MODEL FILE (LB4):
...
@property({name: 'id'})
id: string;
...
```
- No repositories or datasources are created with the command.
- All controller CRUD methods are implemented with throwing a "Not implemented" error.
## Expected Behavior
In order to provide basic but functional application, this is my proposal after reading [this comment](https://github.com/strongloop/loopback-next/issues/1849#issuecomment-429729819): 
- Create a default repository file to provide default CRUD operations (same as using the CLI command lb4 repository):
For each model provided in the yaml schema, create a `<model_name>.repository.ts` file that contain basic CRUD operations.
- Use the newly created repository to provide basic functionality in the Controller file (instead of throwing "not implemented" errors).
- Add support for custom attributes in the yaml schema for relations and datasource:
current generator does not create a datasource connector, and there is no relations creation (although there is an issue [#1359](https://github.com/strongloop/loopback-next/issues/1359) regarding this).
The datasource file should be generated the same as creating one using the CLI command lb4 datasource.
It is possible using the [OpenAPI extentions](https://swagger.io/docs/specification/openapi-extensions/) support to define properties for relations and datasource. For example, in the OpenAPI `yaml` file:
```
...
Customer:
    properties:
        id:
            description: Customer ID
            type: string
        name:
            type: string
        required:
            - id
    additionalProperties: false
    x-lb-relations:                 <==== NEW ATTRIBUTE HERE
        orders:
            type: hasMany
            model: Order
    x-lb-datasource: postgreSQL     <==== NEW ATTRIBUTE HERE
Order:
    properties:
        id:
            description: Customer ID
            type: string
[...]

Issue #1921, opened 2018-10-26
[User Scenario] Create APIs with existing database


dhmlau:
## User Scenario
I'd like to create REST APIs with an existing database.  

## Current Behavior
Steps involved:
1. Scaffold the app using `lb4 app` command
2. Discover schema
In `src/application.ts` file, use [DataSource.discoverSchemas](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-discoverschemas) to discover the mode definitions
3. Create an empty model with `Entity` as baseClass, using `lb4 model` command.
Copy and paste the json from step 2 in the generated model.ts located in `src/models` folder.

See details in: 
https://github.com/strongloop/v4.loopback.io/issues/50#issuecomment-433191443
https://github.com/strongloop/v4.loopback.io/issues/50#issuecomment-433236523

## To-dos 
- [ ] Document this model discovery path possibly in the "How tos" section in the docs

## Potential Enhancement
- @marioestradarosa suggested to discover the model through the `lb4 model` command such that:
```
lb4 model --table mysqlTable --datasource myDS
```
- Proposed by @bajtos, a `lb4 discover` command would be helpful - see https://github.com/strongloop/loopback-next/issues/1949

See [his comment](https://github.com/strongloop/v4.loopback.io/issues/50#issuecomment-433236523) for more details.

See also https://github.com/strongloop/loopback-next/issues/1949

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

marvinirwin:
This is my favorite part of loopback3, being able to create and configure every model from a data source.

Using the [typescript compiler API](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API) I've hacked together something which mimics the behavior of the old loopback which produces the following typescript. 

```
import { model as model, property as property, Entity as Entity } from "@loopback/repository";
@model({ settings: { postgresql: { schema: "public", table: "v_nested_sets_graph" } } })
export class VNestedSetsGraph extends Entity {
    @property({ type: "Number", postgresql: { columnName: "source_id", dataType: "bigint", nullable: "YES" } })
    sourceId: Number;
    @property({ type: "Number", postgresql: { columnName: "node_id", dataType: "bigint", nullable: "YES" } })
    nodeId: Number;
    @property({ type: "Number", postgresql: { columnName: "lft", dataType: "bigint", nullable: "YES" } })
    lft: Number;
[...]

PR #2245, opened 2019-01-13
Model Discovery

marvinirwin:
Adds the command `lb4 discover` command and its documentation.

The command has a `schemaDefs` option (the path to a file which contains a schema definition, or array of lb3 schema definitions) which overrides everything and simply creates the corresponding typescript file.  It's used for the integration test, and could also solve #2224

It's a piece of #2090
Implements #1921

## Checklist

### DIFF
--- a/docs/site/Discovering-models.md
+++ b/docs/site/Discovering-models.md
@@ -1,0 +1,45 @@
+---
+lang: en
+title: 'Discovering models from relational databases'
+keywords: LoopBack 4.0, LoopBack-Next
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Discovering-models.html
+---
+
+## Synopsis
+
+LoopBack makes it simple to create models from an existing relational database.
+This process is called _discovery_ and is supported by the following connectors:
+
+- Cassandra
+- MySQL
+- Oracle
+- PostgreSQL
+- SQL Server
+- IBM DB2
+- IBM DashDB
+- IBM DB2 for z/OS
+- [SAP HANA](https://www.npmjs.org/package/loopback-connector-saphana) - Not
+  officially supported;
+
+## Overview
+
+Models can be discovered from a supported datasource by running the
+`lb4 discover` command.
+
+**The LoopBack project must be built and contain the built datasource files in
+`PROJECT_DIR/dist/datasources/*.js`**
+
+### Options
+
+`--dataSource`: Put a valid datasource name here to skip the datasource prompt
+
+`--views`: Choose whether to discover views. Default is true
+
+`--all`: Skips the model prompt and discovers all of them
+
+`--outDir`: Specify the directory into which the `model.model.ts` files will be
+placed. Default is `src/models`
+
+`--schema`: Specify the schema which the datasource will find the models to
+discover
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -97,6 +97,12 @@ export class Customer {
 }
 ```
 
+## Model Discovery
+
+LoopBack can automatically create model definitions by discovering the schema of
+your database. See [Discovering models](Discovering-models.md) for more details
+and a list of connectors supporting model discovery.
+
 ## Using the Juggler Bridge
 
 To define a model for use with the juggler bridge, extend your classes from
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -275,9 +275,35 @@ Run the following command to install the CLI.
 
     Arguments:
       name  # Name for the observer  Type: String  Required: false
+
     ```
 
-11. To list available commands
+11. To discover a model from a supported datasource
+
+    ```sh
+    cd <your-project-directory>
+    lb4 discover
+      lb4 discover [<name>] [options]
+
+    Options:
+      -h,    --help           # Print the generator's options and usage
+             --skip-cache     # Do not remember prompt answers                                              Default: false
+             --skip-install   # Do not automatically install dependencies                                   Default: false
+             --force-install  # Fail on install dependencies error                                          Default: false
+      -c,    --config         # JSON file name or value to configure options
+      -y,    --yes            # Skip all confirmation prompts with default or provided value
+             --format         # Format generated code using npm run lint:fix
+      -ds,   --dataSource     # The name of the datasource to discover
+             --views          # Boolean to discover views                                                   Default: true
+             --schema         # Schema to discover
+             --all            # Discover all models without prompting users to select                       Default: false
+             --outDir         # Specify the directory into which the `model.model.ts` files will be placed
+
+    Arguments:
+      name  # Name for the discover  Type: String  Required: false
+    ```
+
+12. To list available commands
 
     `lb4 --commands` (or `lb4 -l`)
 
@@ -296,7 +322,7 @@ Run the following command to install the CLI.
 
     Please note `lb4 --help` also prints out available commands.
 
-12. To print out version information
+13. To print out version information
 
     `lb4 --version` (or `lb4 -v`)
 
--- a/packages/cli/generators/discover/index.js
+++ b/packages/cli/generators/discover/index.js
@@ -1,0 +1,255 @@
+path = require('path');
+const fs = require('fs');
+const ArtifactGenerator = require('../../lib/artifact-generator');
+const modelMaker = require('../../lib/model-discoverer');
+const debug = require('../../lib/debug')('discover-generator');
+const utils = require('../../lib/utils');
+const modelDiscoverer = require('../../lib/model-discoverer');
+const rootDir = 'src';
+
+module.exports = class DiscoveryGenerator extends ArtifactGenerator {
+  constructor(args, opts) {
+    super(args, opts);
+
+    this.option('dataSource', {
+      type: String,
+      alias: 'ds',
+      description: 'The name of the datasource to discover',
+    });
+
+    this.option('views', {
+      type: Boolean,
+      description: 'Boolean to discover views',
+      default: true,
+    });
+
+    this.option('schema', {
+      type: String,
+      description: 'Schema to discover',
+      default: '',
+    });
+
+    this.option('all', {
+      type: Boolean,
+      description: 'Discover all models without prompting users to select',
+      default: false,
+    });
+
+    this.option('outDir', {
+      type: String,
+      description:
+        'Specify the directory into which the `model.model.ts` files will be placed',
+      default: undefined,
+    });
+  }
+
+  _setupGenerator() {
+    this.artifactInfo = {
+      type: 'discover',
+      rootDir,
+      outDir: path.resolve(rootDir, 'models'),
+    };
+
+    return super._setupGenerator();
+  }
+
+  /**
+   * If we have a dataSource, attempt to load it
+   * @returns {*}
+   */
+  setOptions() {
+    if (this.options.dataSource) {
+      debug(`Data source specified: ${this.options.dataSource}`);
+      this.artifactInfo.dataSource = modelMaker.loadDataSourceByName(
+        this.options.dataSource,
+      );
+    }
+
+    return super.setOptions();
+  }
+
+  /**
+   * Ensure CLI is being run in a LoopBack 4 project.
+   */
+  checkLoopBackProject() {
+    if (this.shouldExit()) return;
+    return super.checkLoopBackProject();
+  }
+
+  /**
+   * Loads all datasources to choose if the dataSource option isn't set
+   */
+  async loadAllDatasources() {
+    // If we have a dataSourcePath then it is already loaded for us, we don't need load any
+    if (this.artifactInfo.dataSource) {
+      return;
+    }
+    const dsDir = modelMaker.DEFAULT_DATASOURCE_DIRECTORY;
+    const datasourcesList = await utils.getArtifactList(
+      dsDir,
+      'datasource',
+      false,
+    );
+    debug(datasourcesList);
+
+    this.dataSourceChoices = datasourcesList.map(s =>
+      modelDiscoverer.loadDataSource(
+        path.resolve(dsDir, `${utils.kebabCase(s)}.datasource.js`),
+      ),
+    );
+    debug(`Done importing datasources`);
+  }
+
+  /**
+   * Ask the user to select the data source from which to discover
+   */
+  promptDataSource() {
+    if (this.shouldExit()) return;
+    const prompts = [
+      {
+        name: 'dataSource',
+        message: `Select the connector to discover`,
+        type: 'list',
+        choices: this.dataSourceChoices,
+        when:
+          this.artifactInfo.dataSource === undefined &&
+          !this.artifactInfo.modelDefinitions,
+      },
+    ];
+
+    return this.prompt(prompts).then(answer => {
+      if (!answer.dataSource) return;
+      debug(`Datasource answer: ${JSON.stringify(answer)}`);
+
+      this.artifactInfo.dataSource = this.dataSourceChoices.find(
+        d => d.name === answer.dataSource,
+      );
+    });
+  }
+
+  /**
+   * Puts all discoverable models in this.modelChoices
+   */
+  async discoverModelInfos() {
+    if (this.artifactInfo.modelDefinitions) return;
+    debug(`Getting all models from ${this.artifactInfo.dataSource.name}`);
+
+    this.modelChoices = await modelMaker.discoverModelNames(
+      this.artifactInfo.dataSource,
+      {views: this.options.views, schema: this.options.schema},
+    );
+    debug(
+      `Got ${this.modelChoices.length} models from ${
+        this.artifactInfo.dataSource.name
+      }`,
+    );
+  }
+
+  /**
+   * Now that we have a list of all models for a datasource,
+   * ask which models to discover
+   */
+  promptModelChoices() {
+    // If we are discovering all we don't need to prompt
+    if (this.options.all) {
+      this.discoveringModels = this.modelChoices;
+    }
+
+    const prompts = [
+      {
+        name: 'discoveringModels',
+        message: `Select the models which to discover`,
+        type: 'checkbox',
+        choices: this.modelChoices,
+        when:
+          this.discoveringModels === undefined &&
+          !this.artifactInfo.modelDefinitions,
+      },
+    ];
+
+    return this.prompt(prompts).then(answers => {
+      if (!answers.discoveringModels) return;
+      debug(`Models chosen: ${JSON.stringify(answers)}`);
+      this.discoveringModels = [];
+      answers.discoveringModels.forEach(m => {
+        this.discoveringModels.push(this.modelChoices.find(c => c.name === m));
+      });
+    });
+  }
+
+  /**
+   * Using artifactInfo.dataSource,
+   * artifactInfo.modelNameOptions
+   *
+   * this will discover every model
+   * and put it in artifactInfo.modelDefinitions
+   * @return {Promise<void>}
+   */
+  async getAllModelDefs() {
+    this.artifactInfo.modelDefinitions = [];
+    for (let i = 0; i < this.discoveringModels.length; i++) {
+      const modelInfo = this.discoveringModels[i];
+      debug(`Discovering: ${modelInfo.name}...`);
+      this.artifactInfo.modelDefinitions.push(
+        await modelMaker.discoverSingleModel(
+          this.artifactInfo.dataSource,
+          modelInfo.name,
+          {schema: modelInfo.schema},
+        ),
+      );
+      debug(`Discovered: ${modelInfo.name}`);
+    }
+  }
+
+  /**
+   * Iterate through all the models we have discovered and scaffold
+   */
+  async scaffold() {
+    this.artifactInfo.indexesToBeUpdated =
+      this.artifactInfo.indexesToBeUpdated || [];
+
+    // Exit if needed
+    if (this.shouldExit()) return false;
+
+    for (let i = 0; i < this.artifactInfo.modelDefinitions.length; i++) {
+      const modelDefinition = this.artifactInfo.modelDefinitions[i];
+      Object.entries(modelDefinition.properties).forEach(([k, v]) =>
+        modelDiscoverer.sanitizeProperty(v),
+      );
+      modelDefinition.isModelBaseBuiltin = true;
+      modelDefinition.modelBaseClass = 'Entity';
+      modelDefinition.className = utils.pascalCase(modelDefinition.name);
+      // These last two are so that the templat doesn't error out of they aren't there
+      modelDefinition.allowAdditionalProperties = true;
+      modelDefinition.modelSettings = modelDefinition.settings || {};
+      debug(`Generating: ${modelDefinition.name}`);
+
+      const fullPath = path.resolve(
+        this.options.outDir || this.artifactInfo.outDir,
+        utils.getModelFileName(modelDefinition.name),
+      );
+      debug(`Writing: ${fullPath}`);
+
+      this.copyTemplatedFiles(
+        modelDiscoverer.MODEL_TEMPLATE_PATH,
+        fullPath,
+        modelDefinition,
+      );
+
+      this.artifactInfo.indexesToBeUpdated.push({
+        dir: this.options.outDir || this.artifactInfo.outDir,
+        file: utils.getModelFileName(modelDefinition.name),
+      });
+    }
+
+    // This part at the end is just for the ArtifactGenerator
+    // end message to output something nice, before it was "Discover undefined was created in src/models/"
+    this.artifactInfo.name = this.artifactInfo.modelDefinitions
+      .map(d => utils.getModelFileName(d.name))
+      .join(',');
+  }
+
+  async end() {
+    await super.end();
+  }
+};
--- a/packages/cli/lib/model-discoverer.js
+++ b/packages/cli/lib/model-discoverer.js
@@ -1,0 +1,108 @@
+const debug = require('./debug')('model-discoverer');
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Given a datasource and discovery options,
+ * return a list of objects {table: 'foo', schema: 'bar}
+ */
+async function discoverModelNames(ds, options) {
+  if (!ds.connected) {
+    await new Promise(resolve => {
+      ds.on('connected', resolve);
+    });
+  }
+  return await ds.discoverModelDefinitions(options);
+}
+
+/**
+ * Returns the schema definition for a model
+ * @param ds {Juggler.DataSource}
+ * @param modelName {string}
+ * @param options {object}
+ * @return {Promise<Juggler.SchemaDefinition>}
+ */
+async function discoverSingleModel(ds, modelName, options) {
+  const schema = await ds.discoverSchema(modelName, options);
+  if (schema) {
+    schema.settings = schema && schema.options;
+  }
+  return schema;
+}
+
+/**
+ * Loads a DataSource from a file
+ * If the path provided is a JSON, it instantiates a juggler.DataSource with the config as the only argument
+ * Else it requires it like a compiled loopback datasource
+ * @param path
+ * @returns juggler.DataSource
+ */
+function loadDataSource(path) {
+  const ds = require(path);
+  const key = Object.keys(ds)[0];
+  const val = new ds[key]();
+  return val;
+}
+
+/**
+ * Loads a compiled loopback datasource by name
+ * @param name {string}
+ * @returns {*}
+ */
+function loadDataSourceByName(name) {
+  debug(`Searching for specified dataSource ${name}`);
+  const dataSourceFiles = getAllDataSourceFiles();
+  debug(`Loaded ${dataSourceFiles.length} dataSource files`);
+
+  for (let i = 0; i < dataSourceFiles.length; i++) {
+    const f = dataSourceFiles[i];
+    const ds = loadDataSource(path.resolve(DEFAULT_DATASOURCE_DIRECTORY, f));
+    if (ds.name === name) {
+      debug(`Found dataSource ${name}`);
+      return ds;
+    } else {
+      debug(`Did not match dataSource ${name} !== ${ds.name}`);
+    }
+  }
+  throw new Error(
+    `Cannot find datasource "${name}" in ${DEFAULT_DATASOURCE_DIRECTORY}`,
+  );
+}
+
+const DEFAULT_DATASOURCE_DIRECTORY = './dist/datasources';
+
+const MODEL_TEMPLATE_PATH = path.resolve(
+  __dirname,
+  '../generators/model/templates/model.ts.ejs',
+);
+
+const sanitizeProperty = function(o) {
+  Object.entries(o).forEach(([k, v]) => {
+    // Delete the null properties so the template doesn't spit out `key: ;`
+    if (v === null) {
+      delete o[k];
+    }
+
+    // If you are an object or array, stringify so you don't appear as [object [object]
+    if (v === Object(v)) {
+      o[k] = JSON.stringify(o[k]);
+    }
+  });
+
+  o.tsType = o.type;
+};
+
+function getAllDataSourceFiles(dir = DEFAULT_DATASOURCE_DIRECTORY) {
+  return fs.readdirSync(dir).filter(s => s.endsWith('.datasource.js'));
+}
+
+module.exports = {
+  getAllDataSourceFiles,
+  sanitizeProperty,
+  discoverModelNames,
+  discoverSingleModel,
+  loadDataSource,
+  loadDataSourceByName,
+  DEFAULT_DATASOURCE_DIRECTORY,
+  MODEL_TEMPLATE_PATH,
+};
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -109,6 +109,16 @@
         "through": ">=2.2.7 <3"
       }
     },
+    "accept-language": {
+      "version": "3.0.18",
+      "resolved": "https://registry.npmjs.org/accept-language/-/accept-language-3.0.18.tgz",
+      "integrity": "sha1-9QJfF79lpGaoRYOMz5jNuHfYM4Q=",
+      "dev": true,
+      "requires": {
+        "bcp47": "^1.1.2",
+        "stable": "^0.1.6"
+      }
+    },
     "agent-base": {
       "version": "4.2.1",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
@@ -330,6 +340,12 @@
         }
       }
     },
+    "bcp47": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/bcp47/-/bcp47-1.1.2.tgz",
+      "integrity": "sha1-NUvjMH/9CEM6ePXh4glYRfifx/4=",
+      "dev": true
+    },
     "bcrypt-pbkdf": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/bcrypt-pbkdf/-/bcrypt-pbkdf-1.0.2.tgz",
@@ -357,6 +373,16 @@
       "version": "2.1.2",
       "resolved": "https://registry.npmjs.org/binaryextensions/-/binaryextensions-2.1.2.tgz",
       "integrity": "sha512-xVNN69YGDghOqCCtA6FI7avYrr02mTJjOgB0/f1VPD3pJC8QEvjTKWc4epDx8AqxxA75NI0QpVM2gPJXUbE4Tg=="
+    },
+    "bl": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/bl/-/bl-2.2.0.tgz",
+      "integrity": "sha512-wbgvOpqopSr7uq6fJrLH8EsvYMJf9gzfo2jCsL2eTy75qXPukA4pCgHamOQkZtY5vmfVtjB+P3LNlMHW5CEZXA==",
+      "dev": true,
+      "requires": {
+        "readable-stream": "^2.3.5",
+        "safe-buffer": "^5.1.1"
+      }
     },
     "bluebird": {
       "version": "3.5.3",
@@ -566,6 +592,12 @@
       "resolved": "https://registry.npmjs.org/chardet/-/chardet-0.7.0.tgz",
       "integrity": "sha512-mT8iDcrh03qDGRRmoA2hmBJnxpllMR+0/0qlzjqZES6NdiWDcZkCNAk4rPFZ9Q85r27unkiNNg8ZOiwZXBHwcA=="
     },
+    "charenc": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/charenc/-/charenc-0.0.2.tgz",
+      "integrity": "sha1-wKHS86cJLgN3S/qD8UwPxXkKhmc=",
+      "dev": true
+    },
     "chownr": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/chownr/-/chownr-1.1.1.tgz",
@@ -596,6 +628,12 @@
           }
         }
       }
+    },
+    "cldrjs": {
+      "version": "0.5.1",
+      "resolved": "https://registry.npmjs.org/cldrjs/-/cldrjs-0.5.1.tgz",
+      "integrity": "sha512-xyiP8uAm8K1IhmpDndZLraloW1yqu0L+HYdQ7O1aGPxx9Cr+BMnPANlNhSt++UKfxytL2hd2NPXgTjiy7k43Ew==",
+      "dev": true
     },
     "cli-boxes": {
       "version": "1.0.0",
@@ -818,6 +856,12 @@
         }
       }
     },
+    "crypt": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/crypt/-/crypt-0.0.2.tgz",
+      "integrity": "sha1-iNf/fsDfuG9xPch7u0LQRNPmxBs=",
+      "dev": true
+    },
     "crypto-random-string": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/crypto-random-string/-/crypto-random-string-1.0.0.tgz",
@@ -919,6 +963,12 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/delayed-stream/-/delayed-stream-1.0.0.tgz",
       "integrity": "sha1-3zrhmayt+31ECqrgsp4icrJOxhk=",
+      "dev": true
+    },
+    "depd": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/depd/-/depd-2.0.0.tgz",
+      "integrity": "sha512-g7nH6P6dyDioJogAAGprGpCtVImJhpPk/roCzdb3fIh61/s/nPsfR6onyMwkCAR/OlC3yBC0lESvUoQEAssIrw==",
       "dev": true
     },
     "detect-conflict": {
@@ -1531,6 +1581,15 @@
         "ini": "^1.3.4"
       }
     },
+    "globalize": {
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/globalize/-/globalize-1.4.2.tgz",
+      "integrity": "sha512-IfKeYI5mAITBmT5EnH8kSQB5uGson4Fkj2XtTpyEbIS7IHNfLHoeTyLJ6tfjiKC6cJXng3IhVurDk5C7ORqFhQ==",
+      "dev": true,
+      "requires": {
+        "cldrjs": "^0.5.0"
+      }
+    },
     "globby": {
       "version": "8.0.2",
       "resolved": "https://registry.npmjs.org/globby/-/globby-8.0.2.tgz",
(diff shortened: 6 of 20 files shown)


packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,111 @@
...
+const path = require('path');
+
+/**
+ * Given a datasource and discovery options,
+ * return a list of objects {table: 'foo', schema: 'bar}
+ */
+function discoverModelNames(ds, options) {
+  return new Promise((resolve, reject) => {
+    function discover() {
+      ds.discoverModelDefinitions(options, (err, schemas) => {
```
bajtos: Discovery APIs support Promises, no need to use callbacks.

```ts
const schemas = await ds.discoverModelDefinitions(options);
// etc.
```

Please rewrite all helpers in this file to use `await` instead of callbacks.

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,41 @@
+---
+lang: en
+title: "Discovering models from relational databases"
+keywords: LoopBack
+sidebar: lb3_sidebar
```
bajtos: Please fix to `lb4_sidebar`.

This new file needs to be added to the sidebar, see https://github.com/strongloop/loopback-next/blob/master/docs/site/sidebars/lb4_sidebar.yml

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -5,7 +5,7 @@ import {model, property} from '@loopback/repository';
 import {<%= modelBaseClass %>} from '.';
 <% } -%>
 
-@model()
+@model(<%- locals.settings ? JSON.stringify({settings}) : '' %>)
```
bajtos: Where is `locals.settings` coming from? Did you perhaps mean to check `settings`?

packages/cli/test/fixtures/discover/index.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+
+exports.SANDBOX_FILES = [
+  {
+    path: '',
+    file: 'test-schema-defs.json',
+    content: fs.readFileSync(
+      require.resolve('./models/test-schema-defs.json.txt'),
```
bajtos: I am confused, why are you using `.json.txt` extension? Can we store JSON content in files with `.json` extension please?

packages/cli/test/fixtures/discover/index.js
```diff
@@ -0,0 +1,23 @@
...
+    ),
+  },
+  {
+    path: 'dist/src/datasources',
+    file: 'mem.datasource.json',
+    content: fs.readFileSync(
+      require.resolve('./models/mem.datasource.json.txt'),
```
bajtos: Ditto, can we use `mem.datasource.json`?

packages/cli/test/fixtures/discover/index.js
```diff
@@ -0,0 +1,23 @@
...
+    ),
+  },
+  {
+    path: 'dist/src/datasources',
+    file: 'mem.datasource.js',
+    content: fs.readFileSync(require.resolve('./models/mem.datasource.js.txt')),
```
bajtos: And similarly here, can we use `mem.datasource.js` please?

packages/cli/generators/model/templates/model.ts.ejs
marvinirwin: Yes, this was put in to check if the settings object existed.  Is there a more idiomatic way to do this?

bajtos: Have you tried to following?

```suggestion
@model(<%- settings ? JSON.stringify({settings}) : '' %>)
```

marvinirwin: It will throw a reference error if it's not there, I got my solution from here

https://stackoverflow.com/questions/5372559/what-is-the-proper-way-to-check-for-existence-of-variable-in-an-ejs-template-us

raymondfeng: `locals` is special variable for EJS and it's safer to use `locals.settings` if `settings` can be `undefined` when the EJS template is rendered.

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,292 @@
...
+    this.option('schema', {
+      type: String,
+      description: 'Schema to discover, required for postgresql',
+      default: 'public',
+    });
+
+    this.option('discoverAll', {
+      type: Boolean,
+      description:
+        'Avoids the prompt asking for specific models and just discovers them all',
```
raymondfeng: Is `Discover all models without prompting users to select` a better description?

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,292 @@
...
+      );
+      debug(`Discovered: ${modelInfo.name}`);
+    }
+  }
+
+  promptController() {
+    if (this.shouldExit()) return;
+    const prompts = [
+      {
+        name: 'generateRepository',
```
raymondfeng: The name should be `generateController`?

raymondfeng: Where are these answers being used? I don't see code to generate controller/repository classes based on this prompt.

raymondfeng:
@marvinirwin Great effort! Please see my comments.

packages/cli/test/fixtures/discover/index.js
marvinirwin: Changing this extension to `.js` causes something to try and load the file, which then gets a `Cannot find module '@loopback/core'` error when I run `npm test`.  

The reason I had them all with the `.txt` extension is because they did it in `fixtures/repositories/models/`, is avoiding this error why?

packages/cli/generators/model/templates/model.ts.ejs
bajtos: Makes sense, I was not aware of this aspect of EJS.

Thank you for being patient with me and explaining the details, I learned something new today 😄

packages/cli/test/fixtures/discover/index.js
bajtos: Fair enough, let's follow the current convention of using `.txt` files for now. Dealing with issues caused by using a different extension is out of scope of this pull request, I think.

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,41 @@
...
+## Overview
+Models can be discovered from a supported datasource by running the `lb4 discover` command.  *The loopback project must be built and contain the built datasource files in `PROJECT_DIR/dist/src/datasources.js*
+
+### Options
+
+`--dataSource` : Put a valid datasource here to skip the datasource prompt
+
+`--views`: Choose whether to discover views, default true
+
+`-- discoverAll`: Skips the model prompt and discovers all of them
```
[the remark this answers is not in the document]
bajtos: +1, I like `lb4 discover --all`

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,292 @@
...
+      description:
+        'Avoids the prompt asking for specific models and just discovers them all',
+      default: false,
+    });
+
+    this.option('schemaDefs', {
+      type: String,
+      description:
+        'The path to a .json file with an array of schema definition',
+      default: undefined,
```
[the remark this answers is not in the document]
marvinirwin: Yes.  Is there a way I can mark that fact for yeoman?

packages/cli/generators/discover/index.js
marvinirwin: Fixed!  I had briefly toyed with entire api generation.

docs/site/Discovering-models-from-relational-databases.md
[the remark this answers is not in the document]
marvinirwin: Done

packages/cli/test/fixtures/discover/index.js
marvinirwin: Done

packages/cli/test/fixtures/discover/index.js
marvinirwin: Done

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Done

packages/cli/lib/model-discoverer.js
marvinirwin: Done

packages/cli/generators/discover/index.js
marvinirwin: Yes, that's much better.

marvinirwin:
@raymondfeng  @bajtos   Is there anything else I can do to help this get merged?  I think I've addressed all the reviewers' concerns.

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,43 @@
...
+* [IBM DB2](DB2-connector.html)
+* [IBM DashDB](DashDB.html)
+* [IBM DB2 for z/OS](DB2-for-z-OS.html)
+* [SAP HANA](https://www.npmjs.org/package/loopback-connector-saphana) - Not officially supported; see [Community connectors](Community-connectors.html).
+
+
+## Overview
+Models can be discovered from a supported datasource by running the `lb4 discover` command.
+
+**The loopback project must be built and contain the built datasource files in `PROJECT_DIR/dist/src/datasources.js**
```
dhmlau: `loopback` -> `LoopBack`

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Options
+
+`--dataSource`: Put a valid datasource name here to skip the datasource prompt
+
+`--views`: Choose whether to discover views, default true
```
dhmlau: `Choose whether to discover views, default true` -> `Choose whether to discover views.  Default is true`

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,43 @@
...
+
+`--all`: Skips the model prompt and discovers all of them
+
+`--outDir`: Specify the directory into which the `model.model.ts` files will be placed
+
+`--schema`: Specify the schema which the datasource will find the models to discover
+
+`--schemaDefs`: The path to a `.json` file containing a model definition, like those used loopback 2 and 3.  Used to make a typescript model out of an old loopback model without any prompts.
```
dhmlau: - `loopback` -> `LoopBack`
- For the second sentence, I'm wondering if it would be better:
   ```
   LoopBack 4 models will be created out of the LoopBack 3 models specified in `.json` file without any prompts.
   ```

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,272 @@
...
+
+    this.option('views', {
+      type: Boolean,
+      description: 'Boolean to discover views',
+      default: true,
+    });
+
+    this.option('schema', {
+      type: String,
+      description: 'Schema to discover, required for postgresql',
```
dhmlau: I'm not sure if there's a need for "required for postgresql".  Thoughts?

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Fixed!

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Significantly better, fixed.

packages/cli/generators/discover/index.js
marvinirwin: You're right, removed.  

I had it in there because the postgres connector will discover `pg_` tables (along with `public` tables) without a specified schema, on reconsideration, it's not worth warning about.

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Done

marvinirwin:
@raymondfeng  @dhmlau  @bajtos Got anything else?

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,49 @@
...
+- [SAP HANA](https://www.npmjs.org/package/loopback-connector-saphana) - Not
+  officially supported; see [Community connectors](Community-connectors.html).
+
+## Overview
+
+Models can be discovered from a supported datasource by running the
+`lb4 discover` command.
+
+**The LoopBack project must be built and contain the built datasource files in
+`PROJECT_DIR/dist/src/datasources.js**
```
dhmlau: I think there's a missing backtick.

dhmlau:
@marvinirwin, thanks for updating the PR.  I've added one minor comment. 
It looks good to me in terms of the documentation.  We need to get the technical reviews from @bajtos @raymondfeng and team.  Thanks!

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: fixedd

bajtos:
> Merge branch 'master' into pr-model-discovery

We don't merge branches and use rebase instead. That way the git history remains neatly linear. Could you please run `git rebase master` in your feature branch? See [How to rebase your branch](https://loopback.io/doc/en/contrib/code-contrib.html#how-to-rebase-your-branch)).

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,49 @@
...
+### Options
+
+`--dataSource`: Put a valid datasource name here to skip the datasource prompt
+
+`--views`: Choose whether to discover views. Default is true
+
+`--all`: Skips the model prompt and discovers all of them
+
+`--outDir`: Specify the directory into which the `model.model.ts` files will be
+placed
```
bajtos: I would expect `outDir` to default to `models`. Can we mention that in the docs and make sure the implementation does provide this default?

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,49 @@
...
+
+`--schema`: Specify the schema which the datasource will find the models to
+discover
+
+`--schemaDefs`: The path to a `.json` file containing a model definition, like
```
bajtos: Can we find a better name for this option, what that will convey the fact that we are loading models from JSON files instead of the database?

Few ideas to consider:

- `--import`
- `--import-definition`
- `--from-definition`
- `--from-file`
- `--from-lb3-models`

Also please make mention that the JSON file can contain either a single model definition object or an array with multiple definitions.

docs/site/Model.md
```diff
@@ -65,6 +65,11 @@ decorators for adding metadata to your TypeScript/JavaScript classes in order to
 use them with the legacy implementation of the
 [datasource juggler](https://github.com/strongloop/loopback-datasource-juggler).
 
+## Model Discovery
+
+Some connectors support the ability to
+[Discover](Discovering-models-from-relational-databases.md) models.
```
bajtos: Let's move this section _after_ "Definition of a Model" please.

Here is how I am thinking about the layout of this page:

- Overview
- Definition of a Model
- Model Discovery
- Using the Juggler Bridge

I am also proposing to slightly rephrase the text.

```markdown
LoopBack can automatically create model definitions by discovering the schema 
of your database. See 
[Discovering models](Discovering-models-from-relational-databases.md)
for more details and a list of connectors supporting model discovery.
```

docs/site/Discovering-models-from-relational-databases.md
```diff
@@ -0,0 +1,49 @@
+---
+lang: en
+title: 'Discovering models from relational databases'
+keywords: LoopBack 4.0, LoopBack-Next
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Discovering-models-from-relational-databases.html
```
bajtos: IIRC, Juggler supports several discovery modes. Besides parsing SQL schema, it can also define models based on example data. I think it would be great to implement support for this feature at some point in the future (but not in this pull request).

To keep our doors open, I am proposing to rename this page and leave out the part about relational databases.
- file name: `Discovering models.md`
- permalink: `/doc/en/lb4/Discovering-models.md`

Or perhaps also rename the page to "Model discovery" (`Model-discovery.md)? IDK.

packages/cli/generators/discover/index.js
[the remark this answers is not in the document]
bajtos: Probably not. Can you mention the exclusivity in `description` please?

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,279 @@
...
+      default: false,
+    });
+
+    this.option('schemaDefs', {
+      type: String,
+      description:
+        'The path to a .json file with an array of schema definition',
+      default: undefined,
+    });
+  }
```
bajtos: I don't see `--outDir` defined in generator options. It's defined in the docs though. What am I missing?

docs/site/Discovering-models-from-relational-databases.md
bajtos: On the second thought, I am not sure if it's a good idea to mix `lb4 discover` with importing models from JSON definitions. I find it little bit confusing as a user, plus it makes the implementation more complex because of all those `if (importing from file) return` statements at the top of many generator steps.

Have you considered moving this functionality into `lb4 model` generator? I think it will give us CLI that feels pretty natural too.

```sh
lb4 model --import definition.json
# or perhaps
lb4 model --from-definition definition.json
```

packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,94 @@
...
+    if (v === null) {
+      delete o[k];
+    }
+
+    // If you are an object or array, stringify so you don't appear as [object [object]
+    if (v === Object(v)) {
+      o[k] = JSON.stringify(o[k]);
+    }
+  });
+  // o.id could be 1 or 0, typescript will be angry if it is not a boolean
```
bajtos: This is a bug in our type definitions. Can we fix the typedefs to correctly describe `id` as `boolean | number` please?

packages/cli/package-lock.json
```diff
@@ -0,0 +1,9825 @@
+{
+  "name": "@loopback/cli",
+  "version": "1.2.2",
+  "lockfileVersion": 1,
```
bajtos: I believe we are intentionally not using package-lock files. Please remove it from the pull request.

packages/cli/test/fixtures/discover/mem.datasource.js.txt
```diff
@@ -0,0 +1,29 @@
+"use strict";
+var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
```
bajtos: Hmm, keeping versioning JS files is not pretty. It makes me wonder how are tests for other generators (e.g. `lb4 repository`) dealing with datasource discovery? Could you PTAL and see if there is a way how to use an existing approach? Do we actually need this `.js` file? Isn't it enough to have datasource `.json` file?

packages/cli/test/test-utils.js
```diff
@@ -3,7 +3,9 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-'use strict';
+const fsextra = require('fs-extra');
+
+('use strict');
```
bajtos: Please keep `use strict` at the top.

```js
'use strict';

const fsextra = require('fs-extra');
```

packages/cli/test/test-utils.js
```diff
@@ -124,8 +126,14 @@ exports.givenLBProject = function(rootDir, options) {
...
-      if (!fs.existsSync(fullPath)) {
-        fs.writeFileSync(fullPath, theFile.content);
+      const datasourcespath = path.join(rootDir);
+      try {
+        if (!fs.existsSync(fullPath)) {
+          fsextra.ensureDirSync(path.dirname(fullPath));
+          fs.writeFileSync(fullPath, theFile.content);
+        }
+      } catch (e) {
+        console.error(e);
```
bajtos: When the file cannot be created then the test should fail on that error.

Please remove the `try`/`catch` block and let the test runner handle any error thrown by this helper.

packages/cli/test/fixtures/discover/mem.datasource.js.txt
marvinirwin: You're right, I'm going to try and do this with the `legacy.Datasource` constructor and a JSON file.  The only downside I can see is that if any config is defined in`server/datasources/datasource.ts` it won't be loaded properly

packages/cli/test/test-utils.js
marvinirwin: fixed

packages/cli/test/test-utils.js
marvinirwin: fixed

packages/cli/package-lock.json
marvinirwin: fixed

packages/cli/lib/model-discoverer.js
marvinirwin: I'm not sure how to do this

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Yes it's dramatically more intuitive, I'll move it to the model generator.  

Slightly related:  The `.json` loading was used to give me a way of testing the model discovery, which I still have a bit of trouble doing thoroughly.  A couple of solutions:

The memory datasource could be given `discoverModelDefinitions` and `discoverSchema`, but that might be out of the scope of this PR.  

Docker could also be used in the unit tests, but that definitely out of scope of this PR

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: Added test for custom --outDir and mentioned default in docs

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: @bajtos After looking around the code a bit I'm proposing keeping the --schemaDefs option.  And that I submit a seperate PR for adding the `--import` option to the model generator.

The current `discover --schemaDefs path` accepts an array of objects, which mimics how the `discover` command discovers multiple models.  I also need a way to mock data for the integration test.

packages/cli/test/fixtures/discover/mem.datasource.js.txt
marvinirwin: I've tried this, but if I only use a .json file it will crash because the `lb4` command doesn't have the connector.  Is it worth it to try and bootstrap the connectors found in the discovering `node_modules`?

bajtos:
@marvinirwin I see that you pushed new commits. Is the patch ready for another review? Most of us are not receiving notifications when a new commit is pushed into a pull request, please post a comment when it's time to review the changes again.

marvinirwin:
@bajtos  New fixes

1. mentioned exclusivity of --schemaDefs
2. added --import option to the model cli tool
3. renamed from Discovering-models-from-relational-databases.md to Discovering-models.md
4. moved Discovering Models below Definition of a model.

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: done

docs/site/Model.md
marvinirwin: done

packages/cli/generators/discover/index.js
marvinirwin: Woops, fixed

docs/site/Discovering-models-from-relational-databases.md
bajtos: Let's discuss testing and user experience independently.

For testing, I think you should mock `discoverModelDefinitions` and/or `discoverSchema` in such way that your tests sees the model schema you are interested in. I'll try to find time to create a code snippet showing how to mock those two methods.

For user experience, I still prefer importing via `lb4 model` over `--schemaDefs`. I am fine with leaving this option out of this initial pull request, as long as `--schemaDefs` is removed too.

packages/cli/lib/model-discoverer.js
bajtos: I think we need to change the following line:

https://github.com/strongloop/loopback-next/blob/6f62444601faba789b798df953d14c2aababf26b/packages/repository/src/model.ts#L31

```diff
-  id?: boolean;
+  id?: boolean | number;
```

packages/cli/package-lock.json
bajtos: I still see the package-lock file included in the PR. PTAL again.

packages/cli/test/fixtures/discover/mem.datasource.js.txt
bajtos: Could you please create a new feature branch in your repo where you will commit the code showing what have you tried, so that I can run that version myself and better understand what's going on under the hood? I am afraid I am not able to build a good picture in my head just from our discussion.

packages/cli/lib/model-discoverer.js
bajtos: It may be best to submit this fix as a standalone pull request, that way we can get it landed quickly.

docs/site/Discovering-models-from-relational-databases.md
bajtos: I am thinking about the following solution for testing:

- Modify the datasource class used by your tests (`mem.datasource.js.txt`) to override discovery methods provided by `juggler.DataSource`.
- In these overridden methods, load schema from `test-schema-defs.json` and return the relevant parts depending on which discovery method was called.

A mock-up to illustrate what I mean:

```ts
export class MemDataSource extends juggler.DataSource {
  async discoverSchema(name, options) {
    await ensureSchemaDefsLoaded();
    const schema = this.schemas.find(s => s.name === name);
    if (schema) return schema;
    throw new Error(`Schema not found: ${name}`);
  }

  private schemas;
  private async ensureSchemaDefsLoaded() {
    if (schemas) return;
    const schemaFile = path.resolve(__dirname, 'test-schema-defs.json');
    _schemas = JSON.parse(await fs.readFile(schemaFile, 'utf-8'));
  }
}    
```

packages/cli/test/fixtures/discover/mem.datasource.js.txt
marvinirwin: ### Compare loading from `name.datasource.json`, to loading `name.datasource.js`

(requires lb4 project with non-memory datasource in `dist`, it doesn't have to connect successfully)

```
git clone https://github.com/marvinirwin/loopback-next --single-branch --branch discovering-from-json  .;
cd packages/cli;
npm run build;
npm link;
cd SOME_LB4_PROJECT;
//  new DataSource(JSON.parse(fs.readFileSync(pathToJson).toString()))
// This should tell you to npm install the connector
LOAD_JSON_DATASOURCES=true lb4 discover;
// new require(pathToJavascript);
// This will not tell you to npm install the connector
LOAD_JSON_DATASOURCES= lb4 discover;
```

marvinirwin:
@bajtos New commit!

1. Removed `--import` and `--schemaDefs`
2. Testing `lb4 discover` is now done with a custom `DataSource`
3. Submitted https://github.com/strongloop/loopback-next/pull/2406

docs/site/Discovering-models-from-relational-databases.md
marvinirwin: I'm pretty embarrassed I didn't think of implementing my own DataSource.  The testing is now much easier.    

https://github.com/strongloop/loopback-next/blob/5445c58c09760df39f5d010a136979f36c57b217/packages/cli/test/fixtures/discover/mem.datasource.js.txt

packages/cli/lib/model-discoverer.js
marvinirwin: I've submitted it here https://github.com/strongloop/loopback-next/pull/2406

Royedc4:
Hey guys, I've been a long term user of 2.X and this feature is important for us. How could we try this PR version to use the `lb4 discover`?

marvinirwin:
@Royedc4 

> Hey guys, I've been a long term user of 2.X and this feature is important for us.  How could we try this PR version to use the `lb4 discover`?

I created this PR and I'm waiting on @bajtos for feedback right now, but I use it often.

```
git clone -b pr-model-discovery --single-branch https://github.com/marvinirwin/loopback-next;
cd loopback-next/packages/cli;
npm install && npm run build && npm run link;
cd `LOOPBACK_4_PROJECT`;
npm run build;
lb4 discover;
```

packages/cli/package.json
```diff
@@ -48,6 +48,7 @@
     "fs-extra": "^7.0.1",
     "json5": "^2.0.1",
     "lodash": "^4.17.11",
+    "loopback-datasource-juggler": "^4.5.3",
```
bajtos: IIUC, `loopback-datasource-juggler` is used only by the test suite. If that's true, then please move it from `dependencies` to `devDependencies`.

packages/cli/test/fixtures/discover/index.js
```diff
@@ -0,0 +1,19 @@
...
+  /*  {
+    path: '',
+    file: 'test-schema-defs.json',
+    content: fs.readFileSync(require.resolve('./test-schema-defs.json')),
+  },*/
+  /*  {
+    path: 'dist/src/datasources',
+    file: 'mem.datasource.json',
+    content: fs.readFileSync(require.resolve('./mem.datasource.json')),
+  },*/
```
bajtos: Please remove commented-out code. Also if `test-schema-defs.json` and `mem.datasource.json` are no longer used, then remove those files too.

docs/site/Discovering-models.md
```diff
@@ -0,0 +1,45 @@
...
+- [SAP HANA](https://www.npmjs.org/package/loopback-connector-saphana) - Not
+  officially supported; see [Community connectors](Community-connectors.html).
+
+## Overview
+
+Models can be discovered from a supported datasource by running the
+`lb4 discover` command.
+
+**The LoopBack project must be built and contain the built datasource files in
+`PROJECT_DIR/dist/src/datasources.js`**
```
bajtos: I don't think LB4 projects have a file called `dist/src/datasources.js`. Did you mean `dist/src/datasources/*.js` instead?

Also please note that we have recently changed the organization of project files in the `dist` folder, the compiler is configured to compile `src/*` to `dist/*`, i.e. `src/datasources/db.datasource.ts` is compiled to `dist/datasources/db.datasource.js` now.

Please update the README (and the implementation?) accordingly.

packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,105 @@
...
+      ds.on('connected', resolve);
+    });
+  }
+  return await ds.discoverModelDefinitions(options);
+}
+
+/**
+ * Returns the entire model definition for a model
+ */
+async function discoverSingleModel(ds, options, name) {
```
bajtos:
```suggestion
async function discoverSingleModel(ds, modelName, options) {
```

`options` is usually the last argument. I find the argument name `name` as too generic, let's use `modelName` to make it obvious what kind of a name we are expecting.

packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      debug(`Did not match dataSource ${name} !== ${ds.name}`);
+    }
+  }
+  throw new Error(
+    `Cannot find datasource "${name}" in ${DEFAULT_DATASOURCE_DIRECTORY}`,
+  );
+}
+
+const DEFAULT_DATASOURCE_DIRECTORY = './dist/src/datasources';
```
bajtos:
```suggestion
const DEFAULT_DATASOURCE_DIRECTORY = './dist/datasources';
```

I thought we already have code to discover datasources elsewhere in the CLI codebase, isn't such functionality required by `lb4 repository` command? Can we find a way how to extract a shared helper and avoid code duplication?

packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // If you are an object or array, stringify so you don't appear as [object [object]
+    if (v === Object(v)) {
+      o[k] = JSON.stringify(o[k]);
+    }
+  });
+  // o.id could be 1 or 0, typescript will be angry if it is not a boolean
+  if (o.id) {
+    o.id = !!o.id;
```
bajtos: #2406 has been landed, I think we can remove this workaround now?

```diff
- if (o.id) {
-   o.id = !!o.id;
- }
```

packages/cli/lib/model-discoverer.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+  o.tsType = o.type;
+};
+
+function getAllDataSourceFiles(dir = DEFAULT_DATASOURCE_DIRECTORY) {
```
bajtos: Here is an existing implementation for looking up all datasources. Please re-use the same low-level building blocks instead of creating new ones.

https://github.com/strongloop/loopback-next/blob/71292e9ac1b3d89ebfe284a659264cbb83dbe814/packages/cli/generators/repository/index.js#L232-L310

I think it would be great to extract `promptDataSourceName` generator action into a shared helper that can be called from multiple generators. 

Example usage of such helper in `lb4 datasource`:

```ts
const filter = item => {
    debug(`data source inspecting item: ${item}`);
    const result = utils.isConnectorOfType(
      VALID_CONNECTORS_FOR_REPOSITORY,
      this.artifactInfo.datasourcesDir,
      item,
    );
    return result !== false;
  });
};

return promptDataSourceName(this.options.datasource, this.artifactInfo, filter)
  .then({dataSourceClass} => {
    Object.assign(this.artifactInfo, {dataSourceClass});
    })
  .catch(err => {
    debug(`Error during datasource prompt: ${err}`);
    return this.exit(err);
  });
```

bajtos: > I think it would be great to extract promptDataSourceName generator action into a shared helper that can be called from multiple generators.

Feel free to leave out this refactoring out of scope of this pull request if it's easier that way.

docs/site/Discovering-models.md
marvinirwin: Yes, that's what I meant.  Fixed

packages/cli/package.json
marvinirwin: WOops, fixed

packages/cli/lib/model-discoverer.js
marvinirwin: I'll do this in another PR

marvinirwin: > Here is an existing implementation for looking up all datasources. Please re-use the same low-level building blocks instead of creating new ones.

Fixed, I use [getArtifactList](https://github.com/strongloop/loopback-next/blob/master/packages/cli/lib/utils.js#L211) now.  I'll make the general prompt in another PR

packages/cli/lib/model-discoverer.js
marvinirwin: done

packages/cli/lib/model-discoverer.js
marvinirwin: done

packages/cli/test/fixtures/discover/index.js
marvinirwin: done

packages/cli/lib/model-discoverer.js
marvinirwin: done

marvinirwin:
@bajtos  I've made your changes, now waiting for @jannyHou, @marioestradarosa , @raymondfeng

raymondfeng:
@marvinirwin Thank you for the contribution! I'm rebasing the PR against latest master and squashing them into one commit. Will push back to your branch if it's permitted.

raymondfeng:
@marvinirwin I have pushed my changes back to your branch. It is up to date with upstream master and fixes an issue related to `dist` (not more `dist/src`).

packages/cli/test/test-utils.js
```diff
@@ -2,9 +2,9 @@
 // Node module: @loopback/cli
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
-
 'use strict';
 
+const fsextra = require('fs-extra');
```
bajtos: AFAIK, `fsextra` is exposing all `fs` APIs and adds some new ones. There is no need to have both `fs` and `fsextra` variables. Here is what we usually do:

```js
- const fs = require('fs');
+ const fs = require('fs-extra');
```

packages/cli/lib/model-discoverer.js
bajtos: I am afraid you addressed my comment partially. It's good to see that generators are using the shared `getArtifactList` function now, but we need to change other places like `loadDataSourceByName` too, so that `getAllDataSourceFiles` and `DEFAULT_DATASOURCE_DIRECTORY` can be removed completely.

My concern is that right now, the convention for layout of datasource files is duplicated in multiple places and when we change it in the future, we won't update all of these places. 

Anyhow, I am ok to leave further improvements in this area out of scope of this initial pull request.

marvinirwin:
@bajtos  I think I understand what you mean about the duplication of logic in loading datasources now.  

`DEFAULT_DATASOURCE_DIRECTORY` will be replaced with `utils.distRootDir = 'dist'`.  
`loadDataSourceByName` will be replaced with loading via path with [dataSourceToJSONFilename] (https://github.com/strongloop/loopback-next/blob/861256c4e893f6b26e1d2b81d69e614ab5de9e9f/packages/cli/lib/utils.js#L507)

Once these changes are implemented both the `Repository` and `Discovery` generator can create their datasource prompts in the exact same way.

 I'll begin that PR once this one has been merged.

bajtos:
@marvinirwin 

> I'll begin that PR once this one has been merged.

Sounds good.

I noticed that our code coverage dropped by 0.4% with your patch in place, which is more than we usually accept.

I reviewed the code coverage data, it looks like you pull request is missing tests for datasource prompt when discovering a single model (see e.g. https://coveralls.io/builds/22183916/source?filename=packages/cli/generators/discover/index.js)) and for the code mapping discovered types to TypeScript snippets (see https://coveralls.io/builds/22183916/source?filename=packages/cli/generators/model/index.js#L457).

How easy or difficult it is to write tests covering these two areas?

packages/cli/test/test-utils.js
bajtos: > AFAIK, `fsextra` is exposing all `fs` APIs and adds some new ones. There is no need to have both `fs` and `fsextra` variables.

@marvinirwin PTAL ☝️

marvinirwin: woops, fixed

marvinirwin:
@bajtos Thanks for reminding me about the discovering single models.  I added tests for that and some tests ensuring nicer error messages.    

I'm not sure about the loopback->typescript mapping parts.  I think I can hit them all in a more exhaustive test, but I think that's out of the scope of this PR.

packages/cli/test/integration/generators/discover.integration.js
```diff
@@ -10,6 +10,7 @@ const path = require('path');
 const assert = require('yeoman-assert');
 const testlab = require('@loopback/testlab');
 
+const expect = testlab.expect;
 const TestSandbox = testlab.TestSandbox;
```
bajtos: Nitpick: these two lines can be simplified as follows:

```js
const {expect, TestSanbox} = testlab;
```

packages/cli/test/fixtures/discover/mem.datasource.js.txt
```diff
@@ -0,0 +1,106 @@
+const DataSource = require('loopback-datasource-juggler').DataSource;
```
bajtos: I would prefer to avoid direct dependency on loopback-datasource-juggler and import `DataSource` class constructor from `@loopback/repository`.

Not a big deal, we can improve this in a follow-up pull request.

packages/cli/test/fixtures/discover/mem.datasource.js.txt
```diff
@@ -0,0 +1,106 @@
...
+        'length': null,
+        'precision': null,
+        'scale': 0,
+        'id': 1,
+      },
+    },
+  },
+];
+
+// TODO add .txt to this extension so it doesn't get called by global tests`
```
bajtos: This TODO does not seem to be relevant any more, the file already has `.txt` extension.

docs/site/Discovering-models.md
```diff
@@ -0,0 +1,45 @@
...
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Discovering-models.html
+---
+
+## Synopsis
+
+LoopBack makes it simple to create models from an existing relational database.
+This process is called _discovery_ and is supported by the following connectors:
+
+- [Cassandra](Cassandra-connector.html)
```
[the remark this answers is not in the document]
marvinirwin: Huh, I forgot about those.  From a cursory look, I'm not sure if I can relative paths with those.

packages/cli/generators/discover/index.js
```diff
@@ -0,0 +1,256 @@
...
+      rootDir,
+      outDir: path.resolve(rootDir, 'models'),
+    };
+
+    return super._setupGenerator();
+  }
+
+  /**
+   * If we have a dataSource, attempt to load it
+   * If we have a schemaDef path, attempt to load it
```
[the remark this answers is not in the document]
marvinirwin: good catch, `schemaDefs` had been removed

docs/site/Discovering-models.md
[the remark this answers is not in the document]
dhmlau: @marvinirwin , I believe you're getting the list from the LB3 page: https://loopback.io/doc/en/lb3/Discovering-models-from-relational-databases.html.  We have [a work item ](https://github.com/strongloop/loopback-next/issues/2598) to add those connector docs pages back.  
For this PR, I'd like to propose to simply remove the links of the connectors.  When we get to https://github.com/strongloop/loopback-next/issues/2598, we'll add back the links.

bajtos:
@marvinirwin could you please address the comments raised by @nabdelgadir above, and also add this new command to the [CLI README](https://github.com/strongloop/loopback-next/blob/master/packages/cli/README.md)? I'd like to get them fixed before we land this patch.

marvinirwin:
@bajtos done

raymondfeng:
> add this new command to the CLI README?

You can run `packages/cli/bin/cli-main.js discover --help` to print out the command help that can be added to README.

marvinirwin:
@raymondfeng  

The first line of my output is not correct, but I'm not sure how to change it.  (discover does not have a name parameter)

```
Usage:
  lb4 discover [<name>] [options]

Options:
  -h,    --help           # Print the generator's options and usage
         --skip-cache     # Do not remember prompt answers                                              Default: false
         --skip-install   # Do not automatically install dependencies                                   Default: false
         --force-install  # Fail on install dependencies error                                          Default: false
  -c,    --config         # JSON file name or value to configure options
  -y,    --yes            # Skip all confirmation prompts with default or provided value
         --format         # Format generated code using npm run lint:fix
  -ds,   --dataSource     # The name of the datasource to discover
         --views          # Boolean to discover views                                                   Default: true
         --schema         # Schema to discover
         --all            # Discover all models without prompting users to select                       Default: false
         --outDir         # Specify the directory into which the `model.model.ts` files will be placed

Arguments:
  name  # Name for the discover  Type: String  Required: false
```

packages/cli/README.md
```diff
@@ -251,7 +251,27 @@ Run the following command to install the CLI.
...
+    Options:
+            --dataSource     # The name of the datasource to use
+            --views          # Whether the prompt should list views                                     default: true
+            --all            # Skip the model selection prompt and discover all of them                 default: true
+            --outDir         # The directory in which the `<model>.<model>.ts files will be created     default: src/models
+            --schema         # The schema with which the datasource will discover models
+
+    ```
+
+11) To list available commands
```
nabdelgadir: Can you change this to `11.` instead of `11)` (same for `12.`)?

marvinirwin: nice catch

bajtos:
> The first line of my output is not correct, but I'm not sure how to change it. (discover does not have a name parameter)

@raymondfeng could you please help here? I think you now the internal CLI architecture best.

Personally, I can live with incorrect help message for the time being, I see much more value in having the command `lb4 discover` available to all users.

marvinirwin:
@bajtos Alright I've put the mildly incorrect output from `lb4 discover --help` in the readme.  The Travis is failing right now, though I'm not sure if it's my fault or not

bajtos:
> The Travis is failing right now, though I'm not sure if it's my fault or not

See https://travis-ci.org/strongloop/loopback-next/jobs/516258539. Your feature branch seems to contain a bunch of commits that are violating our [Commit Message Guidelines](https://loopback.io/doc/en/contrib/code-contrib-lb4.html#commit-message-guidelines).

Could you please try to rebase your feature branch on top of the latest master? See https://loopback.io/doc/en/lb4/submitting_a_pr.html#6-rebasing

We can help you with this last step if you like, just let us know.

marvinirwin:
@bajtos Done, I think

dhmlau:
@slnode test please

b-admike:
@marvinirwin I've rebased your branch and resolved some conflicts, PTAL and if you are satisfied, we can land your patch.

marvinirwin:
@b-admike 

LGTM

Royedc4:
Awwwwwesome, Finally 🎉
When will this be available in loopback for everyone?

dhmlau:
There're some changes we want to go in for the same release.  Once that's approved, we'll release.  Hopefully very soon 🤞

PR #2122, opened 2018-12-04
feat(context): add support to watch context bindings

raymondfeng:
Recreated from https://github.com/strongloop/loopback-next/pull/2111 due to branch renaming.

- ContextView - watches matching bindings of a context and maintain
  a live collection of values
- `@inject.view` - injects an array of values resolved from bindings
  that match the filter function
- Enhance `@inject` and `@inject.getter` to accept a binding filter function

See [docs](https://github.com/strongloop/loopback-next/blob/context-watcher/docs/site/Context.md#context-watcher).

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #2249, opened 2019-01-15
feat: add greeter extension to illustrate extension point/extension pattern

raymondfeng:
The PR adds an example project to demonstrate how to implement extensibility using the extension point/extension pattern on top of LoopBack 4's IoC and DI foundation.

See https://github.com/strongloop/loopback-next/blob/greeter-extension/examples/greeter-extension/README.md.

## Checklist


jannyHou:
@raymondfeng Good to have an example repository for extension. What's the difference between this example repo and the log extension in https://github.com/strongloop/loopback-next/tree/master/examples/log-extension?

bajtos:
> What's the difference between this example repo and the log extension in https://github.com/strongloop/loopback-next/tree/master/examples/log-extension?

+1, I'd like to better understand the differences too.

bajtos:
Also IIUC, this pull request contains commits from other pull requests, for example #2122. Am I assuming correctly that you will rebase those commit away once the other PRs are landed?

raymondfeng:
> Also IIUC, this pull request contains commits from other pull requests, for example #2122. Am I assuming correctly that you will rebase those commit away once the other PRs are landed?

Yes. The PR is against `context-watcher` branch.

raymondfeng:
> What's the difference between this example repo and the log extension in https://github.com/strongloop/loopback-next/tree/master/examples/log-extension?

The log-extension focuses on demonstrating the common LB4 constructs for extensions, such as decorators, mixins, providers, and components. 

The greeter-extension is dedicated to the extension point/extension pattern. It illustrates how to solve the common extensibility issue by implementing the pattern on top of LB4 constructs. See README for more details.

bajtos:
@raymondfeng it makes me wonder. Instead of building a made-up example, would it be better to update our `@loopback/authentication` or `@loopback/boot` packages to use these new features first? That way we can test the new extensibility features in a more realistic scenarios and thus have a higher chance of discovering missing pieces or rough edges.

I agree it does make sense to have an example showing extension points in isolation, making it easier to understand for readers. I am just not sure if now is the right time to work on that example, before we reworked our existing packages to use these new mechanisms?

raymondfeng:
My intention is to make the business logic very simple to avoid any noises. The example represents the minimal requirements for any module that needs to implement `extension point/extension` pattern with LB4 core:

1. The extension point defines an interface for extensions to implement
2. Extensions can be added independent of the extension point
3. The extension point need to access all registered extensions
4. Extensions can come and go
5. The extension point and extensions should be able to accept configuration

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,160 @@
...
+}
+```
+
+### Define class for the extension point
+
+```ts
+/**
+ * An extension point for greeters that can greet in different languages
+ */
+export class GreeterExtensionPoint {
```
bajtos: I find this class name problematic. This class is not the extension point, it's a service providing multilingual `greet` function and tapping into the `greeter` extension point. The actual extension point is just a convention on how greeter implementations should be bound on the context and how the greeter service is looking them up.

If we want to better show the implementation details of an extension point, then I am proposing a slightly different design:

- `Greeter` interface remains without changes, this part is already designed wells.
- `GreeterExtensionPoint` that accepts the requested language and returns `Greeter` instance matching the language.
- `GreeterService` that provides `greet` method and uses `GreeterExtensionPoint` to obtain the greeter instance to use.

```ts
export class GreeterExtensionPoint {
  constructor(
    @inject.getter(bindingTagFilter({extensionPoint: 'greeter'}))
    private greeters: Getter<Greeter[]>,
  ) {}

  findByLanguage(language: string): Greeter | undefined {
    const greeters = await this.greeters();
    return greeters.find(g => g.language === language);
  }
}

class GreeterService
  constructor(
    @inject('extensions.greeter)
    private greeterRegistry: GreeterExtensionPoint
    // ...
  ) {}

  async greet(language: string, name: string): Promise<string> {
    const greeter = await greeterRegistry.findByLanguage(language);
    const greeting = greeter ? greeter.greet(name) : `Hello, ${name}`;
    // etc.
  }
}
```

examples/greeter-extension/src/component.ts
```diff
@@ -0,0 +1,20 @@
...
+import {EnglishGreeter} from './greeters/greeter-en';
+import {GreeterExtensionPoint} from './greeter-extension-point';
+
+export class GreeterComponent implements Component {
+  bindings = [
+    createBindingFromClass(GreeterExtensionPoint, {
+      key: 'greeter-extension-point',
+    }),
+    createBindingFromClass(EnglishGreeter, {namespace: 'greeters'}),
+    createBindingFromClass(ChineseGreeter, {namespace: 'greeters'}),
```
bajtos: Can we use `asGreeter` template please?

examples/greeter-extension/src/types.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+export const asGreeter: BindingTemplate = extensionFor('greeter');
+
+/**
+ * Shortcut to inject extensions for the given extension point. To be promoted
+ * as `@extensions` in `@loopback/core` module.
+ *
+ * @param extensionPoint Name/id of the extension point
+ */
+export function extensions(extensionPoint: string) {
```
bajtos: Perhaps `injectExtensions` to be promoted to `@inject.extensions` may be a better name?

examples/greeter-extension/README.md
bajtos: I feel it would be even better if we could find a way how to refactor GreeterExtensionPoint into something more generic that can be configured via `@inject` decorator. Below is the first version that comes to my mind. It's not as pretty as I would like it to be, we may need to look for a different approach.

```ts
type GreeterLocator = (lang: string) => Greeter;

class GreeterService {
  constructor(
    @inject.extensionLocator(
       'extensions.greeter', 
       (lang: string) => (ext: Greeter) => ext.language === lang)
    private findGreeterByLanguage: GreeterLocator) {}

   async greet(language: string, name: string): Promise<string> {
    const greeter = await findGreeterByLanguage(language);
    // etc.
  }
}

// resolver for extensionLocator
// the next two values will be obtained from injection metadata
const extensionPointName = // 'extensions.greeter';
const filterFactory = // (lang: string) => (ext: Greeter) => ext.language === lang);
const extensions = await new ContextView(ctx, filterByExtension(extensionPointName)).values();
return (...args) => {
  const filter = filterFactory(...args);
  return extensions.find(filter);
};
```

Maybe the complexity is not worth the benefits? IDK.

Anyways, this is probably out of scope of this pull request.

examples/greeter-extension/src/component.ts
bajtos: I mean why the built-in greeters are created with `{namespace: 'greeters'}` but without applying the changes made by `asGreeter` and why `asGreeter` does not set any namespace?

examples/greeter-extension/README.md
raymondfeng: I was experimenting generalization of extension points into a base class at https://github.com/strongloop/loopback-next/tree/extension-point. My current assessment is that the benefit is not significant enough.

examples/greeter-extension/src/types.ts
raymondfeng: Yes. Once we are settled on the final approach.

raymondfeng:
> I find this class name problematic. This class is not the extension point, it's a service providing multilingual `greet` function and tapping into the `greeter` extension point. The actual extension point is just a convention on how greeter implementations should be bound on the context and how the greeter service is looking them up.

Decoupling `GreeterService` from `GreeterDispatcher` is fine if you see the extension point is only responsible for dispatching requests to corresponding extensions. But it should also be possible to combine the responsibility of two into one class.

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,304 @@
...
+ */
+export interface ChineseGreeterOptions {
+  // Name first, default to `true`
+  nameFirst: boolean;
+}
+
+/**
+ * A greeter implementation for Chinese
+ */
+@bind(asGreeter)
```
emonddr: Any chance we can describe what `asGreeter` is earlier? It is only explained at line 223.

emonddr: image

If `@bind(asGreeter)` is : 

image

then what does

image
(adjusted for ChineseGreeter)

do exactly?

raymondfeng: I made some adjustment.

bajtos:
> > I find this class name problematic. This class is not the extension point, it's a service providing multilingual `greet` function and tapping into the `greeter` extension point. The actual extension point is just a convention on how greeter implementations should be bound on the context and how the greeter service is looking them up.
>
> Decoupling `GreeterService` from `GreeterDispatcher` is fine if you see the extension point is only responsible for dispatching requests to corresponding extensions. But it should also be possible to combine the responsibility of two into one class.

I am ok to allow users to combine the responsibility of two into one class, if they wish so.

However, I am arguing that such design
- is a bad practice violating [SOLID principles](https://en.wikipedia.org/wiki/SOLID), in particular [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle),
- and makes the example unnecessarily difficult to understand for new readers.

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,328 @@
...
+/**
+ * Greet in the given language
+ * @param language Language code
+ * @param name Name
+ */
+async greet(language: string, name: string): Promise<string> {
+  let greeting: string = '';
+  // Get the latest list of greeters
+  const greeters = await this.greeters();
+  // Find a greeter that can speak the given language
```
bajtos: If we end up with a single class implementing both greeter lookup and invocation of the `greet`, then we should at least split the implementation of this method into two parts.

```ts
export class GreeterExtensionPoint {
  // ...

async greet(language: string, name: string): Promise<string> {
  const greeter = this.findGreeterByLanguage(language);

  const greeting = greeter
    ? greeter.greet(name):
    : `Hello, ${name}`;  // Fall back to English

  if (this.options && this.options.color) {
    greeting = chalk.keyword(this.options.color)(greeting);
  }
  return greeting;
}
```

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+```ts
+import {createBindingFromClass} from '@loopback/context';
+import {Component} from '@loopback/core';
+import {GreeterExtensionPoint} from './greeter-extension-point';
+
+export class GreeterComponent implements Component {
+  bindings = [
+    createBindingFromClass(GreeterExtensionPoint, {
+      key: 'greeter-extension-point',
```
bajtos: Is there any particular reason for using manually constructed binding key in camel case, instead of using extension-point-class-name as the binding key?

```ts
createBindingFromClass(GreeterExtensionPoint, {
  key: 'GreeterExtensionPoint',
});
```

Ideally, I'd like LB4 to take care of the binding key for me.

```ts
createBindingFromClass(GreeterExtensionPoint)
```

Alternatively, we can provide `asExtensionPoint` template. I am not sure what's the right way for creating a binding from a class using a template, is it the following?

```ts
createBindingFromClass(GreeterExtensionPoint)
  .apply(asExtensionPoint);
```

examples/greeter-extension/src/greeter-extension-point.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Greet in the given language
+   * @param language Language code
+   * @param name Name
+   */
+  async greet(language: string, name: string): Promise<string> {
+    let greeting: string = '';
+    // Get the latest list of greeters
+    const greeters = await this.greeters();
```
bajtos: _Re-posting https://github.com/strongloop/loopback-next/pull/2249/files#r267735009:_

If we end up with a single class implementing both greeter lookup and invocation of the `greet`, then we should at least split the implementation of this method into two parts.

```ts
export class GreeterExtensionPoint {
  // ...

async greet(language: string, name: string): Promise<string> {
  const greeter = this.findGreeterByLanguage(language);

  const greeting = greeter
    ? greeter.greet(name):
    : `Hello, ${name}`;  // Fall back to English

  if (this.options && this.options.color) {
    greeting = chalk.keyword(this.options.color)(greeting);
  }
  return greeting;
}
```

examples/greeter-extension/README.md
raymondfeng: I improved the code to have `findGreeter` and `greet`. The class is renamed to `GreetingService`.

examples/greeter-extension/README.md
raymondfeng: Fixed.

raymondfeng:
@bajtos PTAL

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,352 @@
...
+export class GreetingService {
+  constructor(
+    // ...
+    private greeters: Getter<Greeter[]>,
+    /**
+     * An extension point should be able to receive its options via dependency
+     * injection.
+     */
+    @configuration() // Sugar for @inject('services.GreetingService.options', {optional: true})
+    private options?: GreeterExtensionPointOptions,
```
bajtos: Should we rename the options type to `GreeterServiceOptions` for consistency?

examples/greeter-extension/src/__tests__/greeter-extension.acceptance.ts
```diff
@@ -0,0 +1,80 @@
...
+    let msg = await greetingService.greet('en', 'Raymond');
+    expect(msg).to.eql('Hello, Raymond');
+    msg = await greetingService.greet('zh', 'Raymond');
+    expect(msg).to.eql('Raymond，你好！');
+  });
+
+  it('supports options for the extension point', async () => {
+    // Configure the extension point
+    app.bind('services.GreetingService.options').to({color: 'blue'});
+    const _greetingService = await app.get(GREETING_SERVICE);
```
bajtos: Why a name starting with an underscore?

```suggestion
    greetingService = await app.get(GREETING_SERVICE);
```

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,352 @@
...
+     */
+    @extensions() // Sugar for @inject.getter(filterByTag({extensionPoint: 'greeter'}))
+    private greeters: Getter<Greeter[]>, // ...
+  ) {}
+}
+```
+
+Please note that it's possible to add/remove greeters after the extension point
+is instantiated. The injected getter function always pick up the latest list of
+greeters. For example:
```
bajtos: I find this text slightly confusing. I got the impression that `inject.getter` will not pick up the latest list 😱

I am proposing to rephrase this part as follows:

```markdown
Please note that it's possible to add/remove greeters after the extension point
is instantiated. If we had used `@inject` to receive the list of greeters, our service
would not be able to pick up those changes. With `@inject.getter`, LoopBack injects
a getter function that will return the actual data, picking up any changes made
since the last call. 
```

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,352 @@
...
+```ts
+@extensionPoint('greeter')
+export class GreetingService {
+  constructor(
+    /**
+     * Inject a getter function to fetch greeters (bindings tagged with
+     * 'greeter')
+     */
+    @extensions() // Sugar for @inject.getter(filterByTag({extensionPoint: 'greeter'}))
+    private greeters: Getter<Greeter[]>, // ...
```
bajtos: Nouns are usually used for properties holding data. Functions/methods typically start with verbs.

```suggestion
    private getGreeters: Getter<Greeter[]>, // ...
```

examples/greeter-extension/src/decorators.ts
```diff
@@ -0,0 +1,89 @@
...
+        target,
+      );
+      extensionPointName =
+        (meta && meta.name) || injection.target.constructor.name;
+    }
+
+    const bindingFilter = filterByTag({extensionPoint: extensionPointName});
+    const view = new ContextView(ctx, bindingFilter);
+    view.open();
+    return view.asGetter(session);
```
bajtos: Can we somehow leverage the implementation from `inject.getter` here? Perhaps by extracting the last three lines into a shared helper function?

examples/greeter-extension/src/greeting-service.ts
```diff
@@ -0,0 +1,69 @@
...
+     * 'greeter')
+     */
+    @extensions() // Sugar for @inject.getter(filterByTag({extensionPoint: 'greeter'}))
+    private greeters: Getter<Greeter[]>,
+    /**
+     * An extension point should be able to receive its options via dependency
+     * injection.
+     */
+    @configuration() // Sugar for @inject('services.GreetingService.options', {optional: true})
+    private options?: GreetingServiceOptions,
```
bajtos: +1 for using `GreetingServiceOptions` name. Please update the docs too.

examples/greeter-extension/src/greeting-service.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+@extensionPoint('greeter')
+export class GreetingService {
+  constructor(
+    /**
+     * Inject a getter function to fetch greeters (bindings tagged with
+     * 'greeter')
+     */
+    @extensions() // Sugar for @inject.getter(filterByTag({extensionPoint: 'greeter'}))
+    private greeters: Getter<Greeter[]>,
```
bajtos:
```suggestion
    private getGreeters: Getter<Greeter[]>,
```

examples/greeter-extension/src/decorators.ts
raymondfeng: I added a helper function to `@loopback/context`.

examples/greeter-extension/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) Author 2018. All Rights Reserved.
```
nabdelgadir: `Author 2018` -> `IBM Corp. 2019`

examples/greeter-extension/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/example-log-extension
```
nabdelgadir: `example-log-extension` -> `example-greeter-extension`

examples/greeter-extension/src/index.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/example-log-extension
```
nabdelgadir: `example-log-extension` -> `example-greeter-extension`

examples/greeter-extension/src/keys.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/example-log-extension
```
nabdelgadir: `example-log-extension` -> `example-greeter-extension`

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,355 @@
...
+ * An extension point for greeters that can greet in different languages
+ */
+export class GreetingService {
+  constructor(
+    /**
+     * Inject a getter function to fetch greeters (bindings tagged with
+     * 'greeter')
+     */
+    @inject.getter(bindingTagFilter({extensionPoint: 'greeter'}))
+    private greeters: Getter<Greeter[]>,
```
nabdelgadir: Should this one and https://github.com/strongloop/loopback-next/pull/2249/files#diff-a0070b7f3505cc0476e42fd94aa48883R293 be `getGreeters` as well?

docs/site/MONOREPO.md
```diff
@@ -15,6 +15,7 @@ The [loopback-next](https://github.com/strongloop/loopback-next) repository uses
...
 | [example-express-composition](https://github.com/strongloop/loopback-next/tree/master/examples/express-composition) | @loopback/example-express-composition | A simple Express application that uses LoopBack 4 REST API                                                                                                                                                                                                                                       |
+| [example-greeter-extension](https://github.com/strongloop/loopback-next/tree/master/examples/greeter-extension)     | @loopback/example-greeter-extension   | An example showing how to implement extension point/extension pattern using LoopBack 4                                                                                                                                                                                                           |
```
nabdelgadir: `how to implement *an* extension point...etc`?

examples/greeter-extension/src/__tests__/greeter-extension.acceptance.ts
```diff
@@ -0,0 +1,80 @@
...
+    msg = await greetingService.greet('zh', 'Raymond');
+    expect(msg).to.eql('Raymond，你好！');
+  });
+
+  it('supports options for the extension point', async () => {
+    // Configure the extension point
+    app.bind('services.GreetingService.options').to({color: 'blue'});
+    greetingService = await app.get(GREETING_SERVICE);
+    const msg = await greetingService.greet('en', 'Raymond');
+    expect(msg).to.eql(chalk.keyword('blue')('Hello, Raymond'));
```
emonddr: if we change 
> expect(msg).to.eql(chalk.keyword('blue')('Hello, Raymond'));
to
```
expect(msg).to.eql(chalk.keyword('green')('Hello, Raymond'));
```

the test still passes since the `msg` string variable is still 'Hello, Raymond' and doesn't contain any `chalk` notation.

The usage of chalk is usually within a console.log() statement : 
```
console.log(chalk.blue('Hello') + ' World' + chalk.red('!'));
```
In a command window, we can see the effect of `chalk` colours.

So perhaps this test should check that the greeter extension point actually had the options set on it, that we requested?

examples/greeter-extension/src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+export function extensionFor(extensionPoint: string): BindingTemplate {
+  return binding =>
+    binding.inScope(BindingScope.SINGLETON).tag({extensionPoint});
+}
+
+/**
+ * A binding template for greeter extensions
+ */
+export const asGreeter: BindingTemplate = binding => {
+  extensionFor('greeter')(binding);
```
jannyHou: Something maybe worth mentioning here:
The extension name here should be consistent with the one in the [extensionPoint decorator](https://github.com/strongloop/loopback-next/blob/988dbb62511c61183b7ad3c3b305b8007fee52b7/examples/greeter-extension/src/greeting-service.ts#L21)

examples/greeter-extension/src/decorators.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * TODO: to be promoted to `@loopback/core` module possibly as
+ * `@inject.extensions`.
+ *
+ * @param extensionPoint Name of the extension point. If not supplied, we use
+ * the name from `@extensionPoint` or the class name of the extension point.
+ */
+export function extensions(extensionPointName?: string) {
+  return inject('', {decorator: '@extensions'}, (ctx, injection, session?) => {
+    if (!session) return undefined;
```
jannyHou: I might miss the implementation of session in other PRs, while wondering what does a `session` do?

examples/greeter-extension/src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  binding.tag({namespace: 'greeters'});
```
jannyHou: Hmm, why the `namespace` is needed here?

examples/greeter-extension/src/__tests__/greeter-extension.acceptance.ts
raymondfeng: Which OS do you test? I see a failure as follows:
```
AssertionError: expected '\u001b[38;5;21mHello, Raymond\u001b[39m' to equal '\u001b[38;5;34mHello, Raymond\u001b[39m'
      + expected - actual
```

raymondfeng: I'll add a test for `greetingService.options` anyway.

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+## Register extensions
+
+To connect an extension to an extension point, we just have to bind the
+extension to the `Context` and tag the binding with
+`{extensionPoint: 'greeter'}`.
+
+```ts
+app
```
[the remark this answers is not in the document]
raymondfeng: The `@bind` metadata is only honored by `createBindingFromClass`. It's not checked for `binding.toClass`.

examples/greeter-extension/src/decorators.ts
raymondfeng: `session` is to track the dependency injection chain of bindings.

examples/greeter-extension/src/types.ts
raymondfeng: Make it easy to build binding key as `<namespace>.<name>`.

examples/greeter-extension/README.md
```diff
@@ -0,0 +1,357 @@
...
+the given language.
+
+```ts
+export class GreetingService {
+  // ...
+  /**
+   * Find a greeter that can speak the given language
+   * @param language Language code for the greeting
+   */
+  async findGreeter(language: string) {
```
bajtos: Perhaps explicitly describe the return type to make it more obvious to readers?

```suggestion
  async findGreeter(language: string): Greeter | undefined {
  ```
```

packages/context/src/context-view.ts
```diff
@@ -152,3 +152,19 @@ export class ContextView<T = unknown> extends EventEmitter
...
   }
 }
+
+/**
+ * Create a context view as a getter
+ * @param ctx Context object
+ * @param bindingFilter A function to match bindings
+ * @param session Resolution session
+ */
+export function createViewGetter<T = unknown>(
```
bajtos: +1 for making these `@loopback/context` changes in a standalone commit, so that release notes are populated with a meaningful entry.

PR #2259, opened 2019-01-16
feat(context): add support for binding configuration and injection

raymondfeng:
Reactivation of #983 based on requirements illustrated by #2249.

1. Add context.configure() to configure bindings
2. Add context.getConfig/getConfigSync() to look up configuration for a binding
3. Add `@config`, `@config.getter`, and `@config.view` to receive injection of configuration

## Checklist

### DIFF
--- a/docs/site/Context.md
+++ b/docs/site/Context.md
@@ -531,3 +531,135 @@ class MyController {
   }
 }
 ```
+
+## Configuration by convention
+
+To allow bound items in the context to be configured, we introduce some
+conventions and corresponding APIs to make it simple and consistent.
+
+We treat configurations for bound items in the context as dependencies, which
+can be resolved and injected in the same way of other forms of dependencies. For
+example, the `RestServer` can be configured with `RestServerConfig`.
+
+Let's first look at an example:
+
+```ts
+export class RestServer {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) app: Application,
+    @inject(RestBindings.CONFIG, {optional: true})
+    config: RestServerConfig = {},
+  ) {
+    // ...
+  }
+  // ...
+}
+```
+
+The configuration (`RestServerConfig`) itself is a binding
+(`RestBindings.CONFIG`) in the context. It's independent of the binding for
+`RestServer`. The caveat is that we need to maintain a different binding key for
+the configuration. Referencing a hard-coded key for the configuration also makes
+it impossible to have more than one instances of the `RestServer` to be
+configured with different options, such as `protocol` or `port`.
+
+To solve these problems, we introduce an accompanying binding for an item that
+expects configuration. For example:
+
+- `servers.RestServer.server1`: RestServer
+- `servers.RestServer.server1:$config`: RestServerConfig
+
+- `servers.RestServer.server2`: RestServer
+- `servers.RestServer.server2:$config`: RestServerConfig
+
+The following APIs are available to enforce/leverage this convention:
+
+1. `ctx.configure('servers.RestServer.server1')` => Binding for the
+   configuration
+2. `Binding.configure('servers.RestServer.server1')` => Creates a accompanying
+   binding for the configuration of the target binding
+3. `ctx.getConfig('servers.RestServer.server1')` => Get configuration
+4. `@config` to inject corresponding configuration
+5. `@config.getter` to inject a getter function for corresponding configuration
+6. `@config.view` to inject a `ContextView` for corresponding configuration
+
+The `RestServer` can now use `@config` to inject configuration for the current
+binding of `RestServer`.
+
+```ts
+export class RestServer {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) app: Application,
+    @config()
+    config: RestServerConfig = {},
+  ) {
+    // ...
+  }
+  // ...
+}
+```
+
+The `@config.*` decorators can take an optional `configPath` parameter to allow
+the configuration value to be a deep property of the bound value. For example,
+`@config('port')` injects `RestServerConfig.port` to the target.
+
+```ts
+export class MyRestServer {
+  constructor(
+    @config('host')
+    host: string,
+    @config('port')
+    port: number,
+  ) {
+    // ...
+  }
+  // ...
+}
+```
+
+Now we can use `context.configure()` to provide configuration for target
+bindings.
+
+```ts
+const appCtx = new Context();
+appCtx.bind('servers.RestServer.server1').toClass(RestServer);
+appCtx
+  .configure('servers.RestServer.server1')
+  .to({protocol: 'https', port: 473});
+
+appCtx.bind('servers.RestServer.server2').toClass(RestServer);
+appCtx.configure('servers.RestServer.server2').to({protocol: 'http', port: 80});
+```
+
+Please note that `@config.*` is different from `@inject.*` as `@config.*`
+injects configuration based on the current binding where `@config.*` is applied.
+No hard-coded binding key is needed. The `@config.*` also allows the same class
+such as `RestServer` to be bound to different keys with different configurations
+as illustrated in the code snippet above.
+
+All configuration accessors or injectors (such as `ctx.getConfig`, `@config`) by
+default treat the configuration binding as optional, i.e. return `undefined` if
+no configuration was bound. This is different from `ctx.get` and `@inject` APIs,
+which require the binding to exist and throw an error when the requested binding
+is not found. The behavior can be customized via `ResolutionOptions.optional`
+flag.
+
+### Allow configuration to be changed dynamically
+
+Some configurations are designed to be changeable dynamically, for example, the
+logging level for an application. To allow that, we introduce `@config.getter`
+to always fetch the latest value of the configuration.
+
+```ts
+export class Logger {
+  @config.getter()
+  private getLevel: Getter<string>;
+
+  async log(level: string, message: string) {
+    const currentLevel = await getLevel();
+    if (shouldLog(level, currentLevel)) {
+      // ...
+    }
+  }
+}
+```
--- a/packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
@@ -1,0 +1,199 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {config, configBindingKeyFor, Context, ContextView, Getter} from '../..';
+
+describe('Context bindings - injecting configuration for bound artifacts', () => {
+  let ctx: Context;
+
+  beforeEach(givenContext);
+
+  it('binds configuration independent of binding', async () => {
+    // Bind configuration
+    ctx.configure('servers.rest.server1').to({port: 3000});
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServer);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServer>('servers.rest.server1');
+
+    expect(server1.configObj).to.eql({port: 3000});
+  });
+
+  it('configures an artifact with a dynamic source', async () => {
+    // Bind configuration
+    ctx
+      .configure('servers.rest.server1')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServer);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServer>('servers.rest.server1');
+    expect(server1.configObj).to.eql({port: 3000});
+  });
+
+  it('configures an artifact with alias', async () => {
+    // Configure rest server 1 to reference `rest` property of the application
+    // configuration
+    ctx
+      .configure('servers.rest.server1')
+      .toAlias(configBindingKeyFor('application', 'rest'));
+
+    // Configure the application
+    ctx.configure('application').to({rest: {port: 3000}});
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServer);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServer>('servers.rest.server1');
+    expect(server1.configObj).to.eql({port: 3000});
+  });
+
+  it('allows configPath for injection', async () => {
+    class RestServerWithPort {
+      constructor(@config('port') public port: number) {}
+    }
+
+    // Bind configuration
+    ctx
+      .configure('servers.rest.server1')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServerWithPort);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  const LOGGER_KEY = 'loggers.Logger';
+  it('injects a getter function to access config', async () => {
+    class Logger {
+      constructor(
+        @config.getter()
+        public configGetter: Getter<LoggerConfig | undefined>,
+      ) {}
+    }
+
+    // Bind logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'INFO'});
+
+    // Bind Logger
+    ctx.bind(LOGGER_KEY).toClass(Logger);
+
+    const logger = await ctx.get<Logger>(LOGGER_KEY);
+    let configObj = await logger.configGetter();
+    expect(configObj).to.eql({level: 'INFO'});
+
+    // Update logger configuration
+    const configBinding = ctx.configure(LOGGER_KEY).to({level: 'DEBUG'});
+
+    configObj = await logger.configGetter();
+    expect(configObj).to.eql({level: 'DEBUG'});
+
+    // Now remove the logger configuration
+    ctx.unbind(configBinding.key);
+
+    // configGetter returns undefined as config is optional by default
+    configObj = await logger.configGetter();
+    expect(configObj).to.be.undefined();
+  });
+
+  it('injects a view to access config', async () => {
+    class Logger {
+      constructor(
+        @config.view()
+        public configView: ContextView<LoggerConfig>,
+      ) {}
+    }
+
+    // Bind logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'INFO'});
+
+    // Bind Logger
+    ctx.bind(LOGGER_KEY).toClass(Logger);
+
+    const logger = await ctx.get<Logger>(LOGGER_KEY);
+    let configObj = await logger.configView.singleValue();
+    expect(configObj).to.eql({level: 'INFO'});
+
+    // Update logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'DEBUG'});
+
+    configObj = await logger.configView.singleValue();
+    expect(configObj).to.eql({level: 'DEBUG'});
+  });
+
+  it('injects a view to access config with path', async () => {
+    class Logger {
+      constructor(
+        @config.view('level')
+        public configView: ContextView<string>,
+      ) {}
+    }
+
+    // Bind logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'INFO'});
+
+    // Bind Logger
+    ctx.bind(LOGGER_KEY).toClass(Logger);
+
+    const logger = await ctx.get<Logger>(LOGGER_KEY);
+    let level = await logger.configView.singleValue();
+    expect(level).to.eql('INFO');
+
+    // Update logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'DEBUG'});
+
+    level = await logger.configView.singleValue();
+    expect(level).to.eql('DEBUG');
+  });
+
+  it('rejects injection of config view if the target type is not ContextView', async () => {
+    class Logger {
+      constructor(
+        @config.view()
+        public configView: object,
+      ) {}
+    }
+
+    // Bind logger configuration
+    ctx.configure(LOGGER_KEY).to({level: 'INFO'});
+
+    // Bind Logger
+    ctx.bind(LOGGER_KEY).toClass(Logger);
+
+    await expect(ctx.get<Logger>(LOGGER_KEY)).to.be.rejectedWith(
+      'The type of Logger.constructor[0] (Object) is not ContextView',
+    );
+  });
+
+  function givenContext() {
+    ctx = new Context();
+  }
+
+  interface RestServerConfig {
+    host?: string;
+    port?: number;
+  }
+
+  class RestServer {
+    constructor(@config() public configObj: RestServerConfig) {}
+  }
+
+  interface LoggerConfig {
+    level: 'DEBUG' | 'INFO' | 'WARNING' | 'ERROR';
+  }
+});
--- a/packages/context/src/__tests__/acceptance/binding-config.feature.md
+++ b/packages/context/src/__tests__/acceptance/binding-config.feature.md
@@ -1,0 +1,84 @@
+# Feature: Context bindings - injecting configuration for bound artifacts
+
+- In order to receive configuration for a given binding
+- As a developer
+- I want to set up configuration for a binding key
+- So that configuration can be injected by the IoC framework
+
+# Scenario: configure an artifact before it's bound
+
+- Given a `Context`
+- Given a class RestServer with a constructor accepting a single argument
+  `config: RestServerConfig`
+- Given `RestServer` ctor argument is decorated with `@config()`
+- When I bind a configuration object `{port: 3000}` to
+  `servers.rest.server1:$config`
+- And bind the rest server to `servers.rest.server1`
+- And resolve the binding for `servers.rest.server1`
+- Then I get a new instance of `RestServer`
+- And the instance was created with `config` set to `{port: 3000}`
+
+```ts
+class RestServer {
+  constructor(@config() public config: RestServerConfig) {}
+}
+
+const ctx = new Context();
+
+// Bind configuration
+ctx.configure('servers.rest.server1').to({port: 3000});
+
+// Bind RestServer
+ctx.bind('servers.rest.server1').toClass(RestServer);
+
+// Resolve an instance of RestServer
+// Expect server1.config to be `{port: 3000}
+const server1 = await ctx.get('servers.rest.server1');
+```
+
+# Scenario: configure an artifact with a dynamic source
+
+- Given a `Context`
+- Given a class RestServer with a constructor accepting a single argument
+  `config: RestServerConfig`
+- Given `RestServer` ctor argument is decorated with `@config()`
+- When I bind a configuration factory of `{port: 3000}` to
+  `servers.rest.server1:$config`
+- And bind the rest server to `servers.rest.server1`
+- And resolve the binding for `servers.rest.server1`
+- Then I get a new instance of `RestServer`
+- And the instance was created with `config` set to `{port: 3000}`
+
+```ts
+class RestServer {
+  constructor(@config() public config: RestServerConfig) {}
+}
+
+const ctx = new Context();
+
+// Bind configuration
+ctx
+  .configure('servers.rest.server1')
+  .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+// Bind RestServer
+ctx.bind('servers.rest.server1').toClass(RestServer);
+
+// Resolve an instance of RestServer
+// Expect server1.config to be `{port: 3000}
+const server1 = await ctx.get('servers.rest.server1');
+```
+
+# Notes
+
+This document only captures the expected behavior at context binding level. It
+establishes conventions to configure and resolve binding.
+
+Sources of configuration can be one or more files, databases, distributed
+registries, or services. How such sources are discovered and loaded is out of
+scope.
+
+See the following modules as candidates for configuration management facilities:
+
+- https://github.com/lorenwest/node-config
+- https://github.com/mozilla/node-convict
--- a/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
@@ -9,11 +9,13 @@ import {
   BindingCreationPolicy,
   BindingKey,
   BindingScope,
+  config,
   Constructor,
   Context,
   Getter,
   inject,
   Injection,
+  instantiateClass,
   invokeMethod,
   Provider,
   ResolutionSession,
@@ -456,14 +458,14 @@ describe('Context bindings - Injecting dependencies of classes', () => {
 
   it('injects a nested property', async () => {
     class TestComponent {
-      constructor(@inject('config#test') public config: string) {}
+      constructor(@inject('config#test') public configForTest: string) {}
     }
 
     ctx.bind('config').to({test: 'test-config'});
     ctx.bind('component').toClass(TestComponent);
 
     const resolved = await ctx.get<TestComponent>('component');
-    expect(resolved.config).to.equal('test-config');
+    expect(resolved.configForTest).to.equal('test-config');
   });
 
   describe('@inject.binding', () => {
@@ -691,6 +693,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
       .toDynamicValue(() => Promise.reject(new Error('Bad')))
       .tag('store:location');
     await expect(ctx.get(STORE_KEY)).to.be.rejectedWith('Bad');
+  });
+
+  it('injects a config property', () => {
+    class Store {
+      constructor(
+        @config('x') public optionX: number,
+        @config('y') public optionY: string,
+      ) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.optionX).to.eql(1);
+    expect(store.optionY).to.eql('a');
+  });
+
+  it('injects a config property with promise value', async () => {
+    class Store {
+      constructor(@config('x') public optionX: number) {}
+    }
+
+    ctx.configure('store').toDynamicValue(() => Promise.resolve({x: 1}));
+    ctx.bind('store').toClass(Store);
+    const store = await ctx.get<Store>('store');
+    expect(store.optionX).to.eql(1);
+  });
+
+  it('injects a config property with a binding provider', async () => {
+    class MyConfigProvider implements Provider<{}> {
+      constructor(@inject('prefix') private prefix: string) {}
+      value() {
+        return {
+          myOption: this.prefix + 'my-option',
+        };
+      }
+    }
+
+    class Store {
+      constructor(@config('myOption') public myOption: string) {}
+    }
+
+    ctx.bind('config').toProvider(MyConfigProvider);
+    ctx.configure('store').toProvider(MyConfigProvider);
+    ctx.bind('prefix').to('hello-');
+    ctx.bind('store').toClass(Store);
+
+    const store = await ctx.get<Store>('store');
+    expect(store.myOption).to.eql('hello-my-option');
+  });
+
+  it('injects a config property with a rejected promise', async () => {
+    class Store {
+      constructor(@config('x') public optionX: number) {}
+    }
+
+    ctx
+      .configure('store')
+      .toDynamicValue(() => Promise.reject(Error('invalid')));
+
+    ctx.bind('store').toClass(Store);
+
+    await expect(ctx.get('store')).to.be.rejectedWith('invalid');
+  });
+
+  it('injects a config property with nested property', () => {
+    class Store {
+      constructor(@config('x.y') public optionXY: string) {}
+    }
+
+    ctx.configure('store').to({x: {y: 'y'}});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.optionXY).to.eql('y');
+  });
+
+  it('injects config if the configPath is not present', () => {
+    class Store {
+      constructor(@config() public configObj: object) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.configObj).to.eql({x: 1, y: 'a'});
+  });
+
+  it("injects config if the configPath is ''", () => {
+    class Store {
+      constructor(@config('') public configObj: object) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.configObj).to.eql({x: 1, y: 'a'});
+  });
+
+  it('injects config with configPath', () => {
+    class Store {
+      constructor(@config('x') public optionX: number) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.optionX).to.eql(1);
+  });
+
+  it('injects undefined option if configPath not found', () => {
+    class Store {
+      constructor(@config('not-exist') public option: string | undefined) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.option).to.be.undefined();
+  });
+
+  it('injects a config property for different bindings with the same class', async () => {
+    class Store {
+      constructor(
+        @config('x') public optionX: number,
+        @config('y') public optionY: string,
+      ) {}
+    }
+
+    ctx.configure('store1').to({x: 1, y: 'a'});
+    ctx.configure('store2').to({x: 2, y: 'b'});
+
+    ctx.bind('store1').toClass(Store);
+    ctx.bind('store2').toClass(Store);
+
+    const store1 = await ctx.get<Store>('store1');
+    expect(store1.optionX).to.eql(1);
+    expect(store1.optionY).to.eql('a');
+
+    const store2 = await ctx.get<Store>('store2');
+    expect(store2.optionX).to.eql(2);
+    expect(store2.optionY).to.eql('b');
+  });
+
+  it('injects undefined config if no binding is present', async () => {
+    class Store {
+      constructor(@config('x') public settings: object | undefined) {}
+    }
+
+    const store = await instantiateClass(Store, ctx);
+    expect(store.settings).to.be.undefined();
+  });
+
+  it('injects config from config binding', () => {
+    class MyStore {
+      constructor(@config('x') public optionX: number) {}
+    }
+
+    ctx.configure('stores.MyStore').to({x: 1, y: 'a'});
+    ctx.bind('stores.MyStore').toClass(MyStore);
+
+    const store: MyStore = ctx.getSync('stores.MyStore');
+    expect(store.optionX).to.eql(1);
   });
 
   function createContext() {
--- a/packages/context/src/__tests__/unit/context-config.unit.ts
+++ b/packages/context/src/__tests__/unit/context-config.unit.ts
@@ -1,0 +1,142 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {
+  BindingAddress,
+  BindingKey,
+  ConfigurationResolver,
+  Context,
+  ContextTags,
+  DefaultConfigurationResolver,
+  ResolutionOptions,
+  ValueOrPromise,
+} from '../..';
+
+describe('Context binding configuration', () => {
+  /**
+   * Create a subclass of context so that we can access parents and registry
+   * for assertions
+   */
+  class TestContext extends Context {
+    public configResolver: ConfigurationResolver;
+  }
+
+  let ctx: TestContext;
+  beforeEach(createContext);
+
+  describe('configure()', () => {
+    it('configures options for a binding before it is bound', () => {
+      const bindingForConfig = ctx.configure('foo').to({x: 1});
+      expect(bindingForConfig.key).to.equal(
+        BindingKey.buildKeyForConfig('foo'),
+      );
+      expect(bindingForConfig.tagMap).to.eql({
+        [ContextTags.CONFIGURATION_FOR]: 'foo',
+      });
+    });
+
+    it('configures options for a binding after it is bound', () => {
+      ctx.bind('foo').to('bar');
+      const bindingForConfig = ctx.configure('foo').to({x: 1});
+      expect(bindingForConfig.key).to.equal(
+        BindingKey.buildKeyForConfig('foo'),
+      );
+      expect(bindingForConfig.tagMap).to.eql({
+        [ContextTags.CONFIGURATION_FOR]: 'foo',
+      });
+    });
+  });
+
+  describe('getConfig()', () => {
+    it('gets config for a binding', async () => {
+      ctx.configure('foo').toDynamicValue(() => Promise.resolve({x: 1}));
+      expect(await ctx.getConfig('foo')).to.eql({x: 1});
+    });
+
+    it('gets config for a binding with configPath', async () => {
+      ctx
+        .configure('foo')
+        .toDynamicValue(() => Promise.resolve({a: {x: 0, y: 0}}));
+      ctx.configure('foo.a').toDynamicValue(() => Promise.resolve({x: 1}));
+      expect(await ctx.getConfig<number>('foo.a', 'x')).to.eql(1);
+      expect(await ctx.getConfig<number>('foo.a', 'y')).to.be.undefined();
+    });
+
+    it('defaults optional to true for config resolution', async () => {
+      // `servers.rest` does not exist yet
+      let server1port = await ctx.getConfig<number>('servers.rest', 'port');
+      expect(server1port).to.be.undefined();
+
+      // Now add `servers.rest`
+      ctx.configure('servers.rest').to({port: 3000});
+      server1port = await ctx.getConfig<number>('servers.rest', 'port');
+      expect(server1port).to.eql(3000);
+    });
+
+    it('throws error if a required config cannot be resolved', async () => {
+      await expect(
+        ctx.getConfig('servers.rest', 'host', {
+          optional: false,
+        }),
+      ).to.be.rejectedWith(/The key 'servers\.rest\:\$config' is not bound/);
+    });
+  });
+
+  describe('getConfigSync()', () => {
+    it('gets config for a binding', () => {
+      ctx.configure('foo').to({x: 1});
+      expect(ctx.getConfigSync('foo')).to.eql({x: 1});
+    });
+
+    it('gets config for a binding with configPath', () => {
+      ctx.configure('foo').to({x: 1});
+      expect(ctx.getConfigSync('foo', 'x')).to.eql(1);
+      expect(ctx.getConfigSync('foo', 'y')).to.be.undefined();
+    });
+
+    it('throws a helpful error when the config is async', () => {
+      ctx.configure('foo').toDynamicValue(() => Promise.resolve('bar'));
+      expect(() => ctx.getConfigSync('foo')).to.throw(
+        /Cannot get config for foo synchronously: the value is a promise/,
+      );
+    });
+  });
+
+  describe('configResolver', () => {
+    class MyConfigResolver implements ConfigurationResolver {
+      getConfigAsValueOrPromise<ConfigValueType>(
+        key: BindingAddress<unknown>,
+        configPath?: string,
+        resolutionOptions?: ResolutionOptions,
+      ): ValueOrPromise<ConfigValueType | undefined> {
+        return (`Dummy config for ${key}` as unknown) as ConfigValueType;
+      }
+    }
+    it('gets default resolver', () => {
+      ctx.getConfigSync('xyz');
+      expect(ctx.configResolver).to.be.instanceOf(DefaultConfigurationResolver);
+    });
+
+    it('allows custom resolver', () => {
+      ctx.configResolver = new MyConfigResolver();
+      const config = ctx.getConfigSync('xyz');
+      expect(config).to.equal('Dummy config for xyz');
+    });
+
+    it('allows custom resolver bound to the context', () => {
+      ctx
+        .bind(`${BindingKey.CONFIG_NAMESPACE}.resolver`)
+        .toClass(MyConfigResolver);
+      const config = ctx.getConfigSync('xyz');
+      expect(config).to.equal('Dummy config for xyz');
+      expect(ctx.configResolver).to.be.instanceOf(MyConfigResolver);
+    });
+  });
+
+  function createContext() {
+    ctx = new TestContext();
+  }
+});
--- a/packages/context/src/__tests__/unit/context-view.unit.ts
+++ b/packages/context/src/__tests__/unit/context-view.unit.ts
@@ -45,6 +45,17 @@ describe('ContextView', () => {
     expect(await taggedAsFoo.asGetter()()).to.eql(['BAR', 'FOO']);
   });
 
+  it('reports error on singleValue() if multiple values exist', async () => {
+    return expect(taggedAsFoo.singleValue()).to.be.rejectedWith(
+      /The ContextView has more than one value\. Use values\(\) to access them\./,
+    );
+  });
+
+  it('supports singleValue() if only one value exist', async () => {
+    server.unbind('bar');
+    expect(await taggedAsFoo.singleValue()).to.eql('FOO');
+  });
+
   it('reloads bindings after refresh', async () => {
     taggedAsFoo.refresh();
     const abcBinding = server
--- a/packages/context/src/__tests__/unit/context.unit.ts
+++ b/packages/context/src/__tests__/unit/context.unit.ts
@@ -538,6 +538,21 @@ describe('Context', () => {
     });
   });
 
+  describe('getOwnerContext', () => {
+    it('returns owner context', () => {
+      ctx.bind('foo').to('bar');
+      expect(ctx.getOwnerContext('foo')).to.equal(ctx);
+    });
+
+    it('returns owner context with parent', () => {
+      ctx.bind('foo').to('bar');
+      const childCtx = new Context(ctx, 'child');
+      childCtx.bind('xyz').to('abc');
+      expect(childCtx.getOwnerContext('foo')).to.equal(ctx);
+      expect(childCtx.getOwnerContext('xyz')).to.equal(childCtx);
+    });
+  });
+
   describe('get', () => {
     it('returns a promise when the binding is async', async () => {
       ctx.bind('foo').toDynamicValue(() => Promise.resolve('bar'));
--- a/packages/context/src/binding-config.ts
+++ b/packages/context/src/binding-config.ts
@@ -1,0 +1,73 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {BindingAddress, BindingKey} from './binding-key';
+import {Context} from './context';
+import {ResolutionOptions} from './resolution-session';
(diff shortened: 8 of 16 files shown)


bajtos:
@raymondfeng In  #983, we had a long discussion about different aspects and user requirements for configuration system. How is this pull request different from #983? Can you also explain which comments you have addressed (and how), and which parts you decided to leave unaddressed?

raymondfeng:
I removed the magic to resolve config values following the namespace hierarchy. The implementation now only uses a naming convention to find the bound config for a given binding key.

bajtos:
> I removed the magic to resolve config values following the namespace hierarchy. The implementation now only uses a naming convention to find the bound config for a given binding key

Cool, that should make the review easier.

How do you imagine the bigger picture and the end-to-end usage of this new feature? 

Few stories to consider:

 - As an LB4 app developer, I'd like to have a single JSON file to define all configuration entries. How are we going to map from a JSON format to the API proposed in this pull request? Maybe you have a different approach in mind?

 - As an extension developer, I may want my extension to provide default configuration options that can be later overridden by app-level config provided by the user.

 - As an extension developer, I may want to contribute configuration for artifacts bound by a different extension. A contrived example: `@loopback/rest-explorer` may want to override certain configuration of `@loopback/rest`.

 - As an extension developer, I want to read configuration of bindings bound by a different extension. For example, `@loopback/rest-explorer` needs to find out the configuration controlling `/openapi.json` URL.


I think it's important to start from the outside (work top-to-bottom) and think about the intended end-to-end user experience first.

It may be best to start with a spike that will allow you to iterate faster, make it easier for all of us to keep the discussion at high level & focused on the intended user experience and internal design.

packages/context/src/resolution-session.ts
```diff
@@ -342,4 +342,9 @@ export interface ResolutionOptions {
    * will return `undefined` instead of throwing an error.
    */
   optional?: boolean;
+
+  /**
+   * Environment for resolution, such as `dev`, `test`, `staging`, and `prod`
+   */
+  environment?: string;
```
bajtos: I am not happy about introducing `environment` to LoopBack runtime. From what I know, this approach is considered as an anti-pattern. It couples internal implementation of our framework with the environment-based decision making. Also in many cases the same environment may need different configuration, e.g. the application can be deployed in `production` environment to different geographic zones and thus will need different database connection strings.

IMO, all environment-specific aspects should be handled in a single place, and that's the single place where we are loading the configuration - e.g. from env variables, JSON files, Consul service, whatever. The concept of per-environment config should stay inside that layer and should not leak to our IoC implementation.

This is another reason why I want to address the "configurability" feature from a high-level perspective of users first (see my comment above).

raymondfeng: I removed the `env`.

raymondfeng:
> * As an LB4 app developer, I'd like to have a single JSON file to define all configuration entries. How are we going to map from a JSON format to the API proposed in this pull request? Maybe you have a different approach in mind?
> * As an extension developer, I may want my extension to provide default configuration options that can be later overridden by app-level config provided by the user.
> * As an extension developer, I may want to contribute configuration for artifacts bound by a different extension. A contrived example: `@loopback/rest-explorer` may want to override certain configuration of `@loopback/rest`.
> * As an extension developer, I want to read configuration of bindings bound by a different extension. For example, `@loopback/rest-explorer` needs to find out the configuration controlling `/openapi.json` URL.

I agree with the use cases but they are out of the scope of this PR.

The purpose of this PR should be very simple (precondition: there are other tiers that load/populate the configuration into the context chain) based on requirements from #2249: 

1. Define the convention to configure a bound entry in the context (the configuration should be decoupled from the target, both of them can be bound independently).

2. Define sugar APIs on `Context` to provide/consume configuration for a given binding (configure() and getConfig())

3. Define a `@inject.config` decorator to inject corresponding configuration into a class without the need to know the binding key.

bajtos:
> I agree with the use cases but they are out of the scope of this PR.
>
> The purpose of this PR should be very simple (precondition: there are other tiers that load/populate the configuration into the context chain) based on requirements from #2249: 
>
> 1. Define the convention to configure a bound entry in the context (the configuration should be decoupled from the target, both of them can be bound independently). This allows us to leverage our IoC/DI container for configuration resolution.
>
> 2. Define sugar APIs on `Context` to provide/consume configuration for a given binding (configure() and getConfig())
>
> 3. Define a `@inject.config` decorator to inject corresponding configuration into a class without the need to know the binding key.

Sure, I agree it's good to work in small increments and thus keep the scope of this pull request small. My concern is that you are defining foundational building blocks for the app/extension configuration, but we don't know if they will be able to support our requirements. Once this new feature is published as part of our public API, it will become difficult to make breaking changes.

I am not asking you to create a full implementation for the use cases I have described above, a high-level description of the indented solution is good enough.

For example:

_As an LB4 app developer, I'd like to have a single JSON file to define all configuration entries. How are we going to map from a JSON format to the API proposed in this pull request?_

We can define a convention where the binding key is used as the key in the JSON file too. E.g.

```json
{
  "rest.port": 3000,
  "rest.host": "localhost",
}
```

Is it just me who see a similarity with Java property files? Maybe we should use nesting instead of dot-separated property names? Anyhow, that's just an implementation detail we can figure out later.

_As an extension developer, I may want my extension to provide default configuration options that can be later overridden by app-level config provided by the user._

Can we rely on the order in which the configuration bindings are created?

```ts
class MyApp {
  constructor() {
    app.component(RestComponent); // defines the default configuration
   
    // custom configuration - overrides the config values bound by components
  }
}
```

Would this become a pain point and we will need to look for different ways how to allow extensions to provide default values?

_As an extension developer, I may want to contribute configuration for artifacts bound by a different extension. A contrived example: `@loopback/rest-explorer` may want to override certain configuration of `@loopback/rest`._

And here it becomes tricky. If we rely on the order in which the configuration was defined, then we need to ensure that `@loopback/rest` component is mounted first, `@loopback/rest-explorer` second.

Thoughts?

_As an extension developer, I want to read configuration of bindings bound by a different extension. For example, `@loopback/rest-explorer` needs to find out the configuration controlling `/openapi.json` URL.)

I think this is a similar problem. If we rely on the order in which components are registered, then the default configuration provided by `@loobpack/rest` will be available at the time when `@loobpack/rest-explorer` is being mounted. However, this will not take into account configuration supplied by the application for `@loopback/rest` after `@loopback/rest-explorer` was mounted. I think that means we need to use a getter function instead of obtaining the configured value directly, similarly to how we are using `@inject.getter` for current authenticated user.

**Now we have a question very relevant to your pull request: do we want to support both flavors of config injection (value vs getter)?**

I am arguing that we should always inject a getter function to support configuration changes made after the config was resolved.

For example, in your greeter extension example shown in https://github.com/strongloop/loopback-next/pull/2249, `GreeterExtensionPoint` is bound in `SINGLETON` scope. Once the binding is resolved and the singleton is created, there is no way how to update it when the configuration changes. The following code will not work:

```ts
app.configure('greeter-extension-point').to({color: false});
(await app.get('greeter-extension-point')).greet('en', 'monochrome');
app.configure('greeter-extension-point').to({color: true});
(await app.get('greeter-extension-point')).greet('en', 'rainbow');
// ^^ this does not pick up the new setting!!!
```

If we change `@inject.config` to always resolve to a getter function, we force all configuration consumers to support config updates from the start.

Let's discuss.

packages/context/src/inject.ts
```diff
@@ -331,21 +395,28 @@ export function describeInjectedArguments(
...
-  const bindings = ctx.findByTag(tag);
-
   return resolveList(bindings, b => {
     // We need to clone the session so that resolution of multiple bindings
     // can be tracked in parallel
     return b.getValue(ctx, ResolutionSession.fork(session));
   });
 }
 
+function resolveByTag(
```
bajtos: Wasn't this resolver already added by a previous pull request? I remember seeing this change in other places. Or is it just the way how git is computing the diff?

packages/context/src/inject.ts
```diff
@@ -331,21 +395,28 @@ export function describeInjectedArguments(
   return meta || [];
 }
 
-function resolveByTag(
+function resolveBindings(
```
bajtos: Similarly here, I don't see how is this change relevant to binding configuration. Can you revert it please?

packages/context/src/inject.ts
```diff
@@ -45,7 +46,7 @@ export interface InjectionMetadata {
    */
   decorator?: string;
   /**
-   * Control if the dependency is optional, default to false
+   * Control if the dependency is optional. Default to `false`.
```
bajtos: I find both the old and the new versions difficult to parse. How about the following?

```suggestion
   * Control if the dependency is optional. Default value: `false`.
```

packages/context/src/context.ts
```diff
@@ -340,6 +350,105 @@ export class Context extends EventEmitter {
...
+   * Create a corresponding binding for configuration of the target bound by
+   * the given key in the context.
+   *
+   * For example, `ctx.configure('controllers.MyController').to({x: 1})` will
+   * create binding `controllers.MyController:$config` with value `{x: 1}`.
+   *
+   * @param key The key for the binding to be configured
+   */
+  configure<ConfigValueType = BoundValue>(
+    key: BindingAddress<unknown> = '',
```
bajtos: IIRC, `BindingAddress` has the template argument set to `unknown` by default, thus we should be able to simplify this line as follows:

```suggestion
    key: BindingAddress = '',
```

Same comment applies to other places below that are using `BindingAddress<unknown>` too.

packages/context/src/context.ts
```diff
@@ -340,6 +350,105 @@ export class Context extends EventEmitter {
...
+    resolutionOptions?: ResolutionOptions,
+  ): ValueOrPromise<ConfigValueType | undefined> {
+    return this.getConfigResolver().getConfigAsValueOrPromise(
+      key,
+      configPath,
+      resolutionOptions,
+    );
+  }
+
+  private getConfigResolver() {
```
bajtos: I find it ugly to have both `configResolver` as a property and `getConfigResolver` as a getter function. Have you considered initializing `this.configResolver` right in the constructor?

packages/context/src/binding-config.ts
```diff
@@ -0,0 +1,63 @@
...
+    configPath?: string,
+    resolutionOptions?: ResolutionOptions,
+  ): ValueOrPromise<ConfigValueType | undefined> {
+    configPath = configPath || '';
+    const configKey = BindingKey.create<ConfigValueType>(
+      BindingKey.buildKeyForConfig(key),
+      configPath,
+    );
+
+    const options: ResolutionOptions = Object.assign(
```
bajtos: How about the following?

```ts
const options: ResolutionOptions = {optional: true,  ...resolutionOptions};
```

packages/context/src/__tests__/acceptance/binding-config.feature.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+# Scenario: configure values for different environments
+
+- Given a `Context`
+- Given a class RestServer with a constructor accepting a single argument
+  `config: RestServerConfig`
+- Given `RestServer` ctor argument is decorated with `@inject.config()`
+- When I bind a configuration object `{port: 3000}` to
+  `$config.test:servers.rest.server1`
```
bajtos: Please update the binding keys in this markdown file to follow the format used by the real implementation. Do we actually need this markdown file at all?

bajtos:
@raymondfeng I see you pushed some new changes, thus many of my line-comments above are outdated now.

However, I don't see any response to my high-level concerns described in https://github.com/strongloop/loopback-next/pull/2259#issuecomment-464064374. What's your opinions on them?

raymondfeng:
@bajtos It's still WIP. I'll address your comments later.

xc2:
@bajtos

> I am arguing that we should always inject a getter function to support configuration changes made after the config was resolved.
> If we change @inject.config to always resolve to a getter function, we force all configuration consumers to support config updates from the start.

I think a sync configuration getter function is ok, but not the async one.

### Synchronous vs Asynchronous

**Configurations may change, but not often change.**

Always injecting a async getter means it is supported to require a config on demand. As a result:

1. Tons of simple sync function will become a async one just because it is asking for a nearly invariable value.
2. It is impossible to use configuration to initialize self because class constructor is never async.

### Getter or Value

It just depends.

- For TRANSIENT bindings, instances are created and gone in seconds, so injecting a value is enough.
- For SINGLETON bindings, configurations are often resolved to create heavy resources. Always using the getter function may cause inconsistent result. For example, in `RestServer`, you may get `port` (via getter) and create http server (which is heavy resource), while you may get `port` (via getter) and log http server information (which is a light task). When `port` changes, the http server is still running in old port, while the new port will logged after.

So the configuration getter may not be a good solution to respond to configuration changes.

### Configurations not often change, but may change

Yes, and most of those who managing these changed configurations are supposed to make quick respond to these changes.

#### Scenario

- A series of long alive mysql connections to `config.mysql.url` was created.
- The old mysql server would be under maintaining for a while, so we need to connect to another server.
- `config.mysql.url` was changed to a temporary server via `zookeeper` on the fly.
- **all alive mysql connection to old server should be closed and new connections should be created.**

Any solutions?

bajtos:
@xc2 thank you for a thoughtful comment! I was thinking about those concerns too.

I think we should provide three flavors of config injection:
- Inject the config object as-is: good for TRANSIENT consumers) 
- Inject a getter function: good for stateless SINGLETON consumers like the GreeterService from #2249)
- Inject a ContextView instance allowing SINGLETON consumers to subscribe to configuration changes. Good for stateful SINGLETON consumers like data-sources, servers, etc.

To keep the number of differen `inject.*` helpers small, I think we should remove `inject.config` and use a different design based on functional composition.

```ts
@inject(configFor(BindingKeys.GREETER_SERVICE), {optional: true})
@inject.getter(configFor(BindingKeys.GREETER_SERVICE), {optional: true})
@inject.view(configFor(BindingKeys.GREETER_SERVICE), {optional: true})

function configFor<ConfigValueType = unknown>(binding: BindingAddress, configPath?: string) {
  return BindingKey.create<ConfigValueType>(
    BindingKey.buildKeyForConfig(key),
    configPath,
  );
}
```

Thoughts?

/cc @strongloop/loopback-next

raymondfeng:
> `@inject(configFor(BindingKeys.GREETER_SERVICE), {optional: true})`
`@inject.getter(configFor(BindingKeys.GREETER_SERVICE), {optional: true})`
`@inject.view(configFor(BindingKeys.GREETER_SERVICE), {optional: true})`

I like the idea to use a binding filter for configurations. And I would like to have the following sugar decorators:
- `@config(BindingKeys.GREETER_SERVICE)` 
- `@config.getter(BindingKeys.GREETER_SERVICE)`
- `@config.view(BindingKeys.GREETER_SERVICE)`

raymondfeng:
@bajtos To allow a child configuration to be resolved from its parent, for example, configuration for `api explorer` can be resolved as `servers.RestServer.options#apiExplorer`, I introduce `Binding.toAlias`. See https://github.com/strongloop/loopback-next/pull/2259/commits/bec1a72035e241c3aadc6a8ebe819c1a883b0e48.

raymondfeng:
FYI: I have implemented the following decorators:

1. `@config` (replacing `@inject.config`)
2. `@config.getter`
3. `@config.view`

bajtos:
> @bajtos To allow a child configuration to be resolved from its parent, for example, configuration for `api explorer` can be resolved as `servers.RestServer.options#apiExplorer`, I introduce `Binding.toAlias`. See https://github.com/strongloop/loopback-next/pull/2259/commits/bec1a72035e241c3aadc6a8ebe819c1a883b0e48.
>
> I created a separate PR - https://github.com/strongloop/loopback-next/pull/2635

Excellent!

I am seeing commits from #2635 in this pull request too, which makes the review less practical. I am proposing to put this patch on hold until #2635 is landed and this patch rebased on top of the new master.

Thoughts?

> FYI: I have implemented the following decorators:
>
> 1 . `@config` (replacing `@inject.config`)
> 2. `@config.getter`
> 3. `@config.view`

raymondfeng:
> I am seeing commits from #2635 in this pull request too, which makes the review less practical. I am proposing to put this patch on hold until #2635 is landed and this patch rebased on top of the new master.

I removed some commits for now as there is no hard dependency.

raymondfeng:
@emonddr Good catch. I'm working with too many branches and sometimes I might make mistakes during rebase. Fixed.

packages/context/src/__tests__/unit/context-config.unit.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    it('throws error if a required config cannot be resolved', async () => {
+      expect(
+        ctx.getConfig('servers.rest', 'host', {
+          optional: false,
+        }),
+      )
+        .to.be.rejectedWith(
+          `Configuration 'servers.rest#host' cannot be resolved`,
```
bajtos: I find this slightly confusing. It's `servers.rest` that could not be resolved, the property accessor `#host` does not play any role here, correct? I would expect to see the message above if the `host` property was not set.

packages/context/src/binding-config.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Resolve config for the binding key
+   *
+   * @param key Binding key
+   * @param configPath Property path for the option. For example, `x.y`
+   * requests for `<config>.x.y`. If not set, the `config` object will be
+   * returned.
+   * @param resolutionOptions Options for the resolution.
+   * - optional: if not set or set to `true`, `undefined` will be returned if
+   * no corresponding value is found. Otherwise, an error will be thrown.
```
bajtos: I think this detailed tsdoc comment should be moved to `ConfigurationResolver` description, so that people implementing `ConfigurationResolver` better understand what to expect in the arguments.

packages/context/src/binding-key.ts
```diff
@@ -103,4 +103,17 @@ export class BindingKey<ValueType> {
       keyWithPath.substr(index + 1),
     );
   }
+
+  static CONFIG_NAMESPACE = '$config';
+  /**
```
bajtos: Please add an empty line to visually delimit individual members.

packages/context/src/binding-key.ts
```diff
@@ -103,4 +103,17 @@ export class BindingKey<ValueType> {
...
+   * Build a binding key for the configuration of the given binding.
+   * The format is `<key>:$config`
+   *
+   * @param key The binding key that accepts the configuration
+   */
+  static buildKeyForConfig(key: BindingAddress<unknown> = '') {
```
bajtos: Can we modify this function to support type checks and inference?

```ts
export class BindingKey<ValueType> {
  // ...

  static buildKeyForConfig<T>(key: BindingAddress<unknown> = ''): BindingKey<T> {
    // ...
    return BindingKey.create<T>(bindingKey);
  }
}
```

Also since this method is part of `BindingKey`, can we perhaps shorten it's name to `forConfig`, with the following use?

```ts
const key = BindingKey.forConfig(...);
```

packages/context/src/context-view.ts
```diff
@@ -151,6 +151,19 @@ export class ContextView<T = unknown> extends EventEmitter
...
   }
+
+  /**
+   * Get the single value
+   */
+  async singleValue(session?: ResolutionSession): Promise<T | undefined> {
+    const values = await this.values(session);
+    if (values.length === 0) return undefined;
+    if (values.length === 1) return values[0];
+    else
```
bajtos: `else` is not needed here, previous branches always return from the function. Please remove this line, it will make the code less indented and probably easier to read too.

packages/context/src/context.ts
```diff
@@ -349,6 +355,120 @@ export class Context extends EventEmitter {
...
+   * For example, `ctx.configure('controllers.MyController').to({x: 1})` will
+   * create binding `controllers.MyController:$config` with value `{x: 1}`.
+   *
+   * @param key The key for the binding to be configured
+   */
+  configure<ConfigValueType = BoundValue>(
+    key: BindingAddress = '',
+  ): Binding<ConfigValueType> {
+    const keyForConfig = BindingKey.buildKeyForConfig(key);
+    const bindingForConfig = this.bind<ConfigValueType>(keyForConfig).tag({
```
bajtos: See my comments above, I think we should pass `ConfigValueType` to the function creating the binding key. Thoughts?

```ts
const keyForConfig = BindingKey.buildKeyForConfig< ConfigValueType>(key);
const bindingForConfig = this.bind(keyForConfig)./*etc*/
```

packages/context/src/context.ts
```diff
@@ -349,6 +355,120 @@ export class Context extends EventEmitter {
...
+      const configResolver = this.getSync<ConfigurationResolver>(
+        `${BindingKey.CONFIG_NAMESPACE}.resolver`,
+        {
+          optional: true,
+        },
+      );
+      if (configResolver) {
+        this.configResolver = configResolver;
+      } else {
+        // Fallback to DefaultConfigurationResolver
```
bajtos: Please add `debug` logs to tell the user what resolver we are using.

packages/context/src/context.ts
```diff
@@ -349,6 +355,120 @@ export class Context extends EventEmitter {
...
+  /**
+   * Resolve configuration synchronously for the binding by key
+   *
+   * @param key Binding key
+   * @param configPath Property path for the option. For example, `x.y`
+   * requests for `config.x.y`. If not set, the `config` object will be
+   * returned.
+   * @param resolutionOptions Options for the resolution.
+   */
+  getConfigSync<ConfigValueType>(
```
bajtos: Is it a good idea to introduce the sync version? I mean what are the use cases where `await ctx.getConfig()` cannot be used?

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,167 @@
...
+import {ContextView} from './context-view';
+import {
+  inject,
+  Injection,
+  InjectionMetadata,
+  inspectTargetType,
+} from './inject';
+import {ResolutionSession} from './resolution-session';
+import {getDeepProperty, ValueOrPromise} from './value-promise';
+/**
```
bajtos: Please add an empty line to visually delimit the block of `import` statements from the actual code.

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,167 @@
...
+  };
+}
+
+function resolveFromConfig(
+  ctx: Context,
+  injection: Injection,
+  session: ResolutionSession,
+): ValueOrPromise<unknown> {
+  if (!session.currentBinding) {
+    // No binding is available
```
bajtos: How can this situation happen? Do we have a test to verify?

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,167 @@
...
+  });
+}
+
+function resolveAsGetterFromConfig(
+  ctx: Context,
+  injection: Injection,
+  session: ResolutionSession,
+) {
+  if (!session.currentBinding) {
+    // No binding is available
```
bajtos: How can this situation happen? Do we have a test to verify?

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+  }
+  if (!session.currentBinding) {
+    // No binding is available
+    return undefined;
+  }
+  const bindingKey = session.currentBinding.key;
+  const view = new ConfigView(
+    ctx,
+    binding =>
```
bajtos: Can you cache this filter function in `Injection` please? So that we create the filter function only once when `@config.view()` is invoked, instead of re-creating it every time a dependency is resolved/injected.

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,167 @@
...
+    filter: BindingFilter,
+    private configPath?: string,
+  ) {
+    super(ctx, filter);
+  }
+
+  async values(session?: ResolutionSession) {
+    const configValues = await super.values(session);
+    return configValues.map(v =>
+      this.configPath ? getDeepProperty(v, this.configPath) : v,
```
bajtos: Performance micro-optimization:

```ts
return this.configPath 
  ? configValues.map(v => getDeepProperty(v, this.configPath)) 
  : configValues.map;
```

packages/context/src/inject.ts
```diff
@@ -318,15 +318,15 @@ export namespace inject {
...
   ctx: Context,
   injection: Readonly<Injection>,
-  session?: ResolutionSession,
+  session: ResolutionSession,
 ) {
   assertTargetIsGetter(injection);
   const bindingSelector = injection.bindingSelector as BindingAddress;
   // We need to clone the session for the getter as it will be resolved later
-  session = ResolutionSession.fork(session);
+  const forkedSession = ResolutionSession.fork(session);
```
bajtos: I see many similarities between `resolveAsGetter` and `resolveAsConfigGetter`, it would be great to share the common bits via a template function. Same for `*View` versions.

bajtos:
> coverage/coveralls — Coverage decreased (-0.01%) to 91.229%

@raymondfeng PTAL at the coverage report, are there any important use-cases/scenarios that are missing test coverage?

packages/context/src/inject-config.ts
raymondfeng: Please note the filter function depends on the current binding key.

docs/site/Context.md
```diff
@@ -530,3 +530,98 @@ class MyController {
...
+
+### Allow configuration to be changed dynamically
+
+Some configurations are designed to be changeable dynamically, for example, the
+logging level for an application. To allow that, we introduce `@configGetter` to
+always fetch the latest value of the configuration.
+
+```ts
+export class Logger {
+  @configGetter()
```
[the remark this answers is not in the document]
raymondfeng: The doc was out of date. It's fixed now.

packages/context/src/inject-config.ts
raymondfeng: It happens when `instantiateClass` is called directly and we already have a test to cover it.

packages/context/src/inject-config.ts
raymondfeng: Fixed.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
```diff
@@ -0,0 +1,155 @@
...
+  });
+
+  it('configures an artifact with alias', async () => {
+    const ctx = new Context();
+
+    // Configure rest server 1 to reference `rest` property of the application
+    // configuration
+    ctx
+      .configure('servers.rest.server1')
+      .toAlias(configBindingKeyFor('application', 'rest'));
```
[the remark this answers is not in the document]
raymondfeng: Please note `@inject` does not leverage `session.currentBinding`.

raymondfeng:
> I see quite few small improvements and additions to various context-related classes. It would be great to extract these small bits into a new pull request. 

See https://github.com/strongloop/loopback-next/pull/2745

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/2747

raymondfeng:
The PR contains commits from #2748, which can be reviewed and landed 1st.

raymondfeng:
FYI: #2745 and #2748 have landed.

docs/site/Context.md
```diff
@@ -530,3 +530,98 @@ class MyController {
...
+    @inject(RestBindings.CONFIG, {optional: true})
+    config: RestServerConfig = {},
+  ) {
+    // ...
+  }
+  // ...
+}
+```
+
+The configuration itself is a binding (`RestBindings.CONFIG`) in the context.
```
hacksparrow: The configuration itself ...  -> The configuration (`RestServerConfig `) itself ...?

docs/site/Context.md
```diff
@@ -530,3 +530,98 @@ class MyController {
...
+maintain a different binding key for the configuration. Referencing a hard-coded
+key for the configuration also makes it impossible to have more than one
+instances of the `RestServer` to be configured with different options, such as
+`protocol` or `port`.
+
+To solve these problems, we introduce an accompanying binding for an item that
+expects configuration. For example:
+
+- `servers.RestServer.server1`: RestServer
+- `servers.RestServer.server1.$config`: RestServerConfig
```
hacksparrow: Why not simply `servers.RestServer.server1.config`?

docs/site/Context.md
```diff
@@ -530,3 +530,98 @@ class MyController {
...
+  private getLevel: Getter<string>;
+
+  async log(level: string, message: string) {
+    const currentLevel = await getLevel();
+    if (shouldLog(level, currentLevel)) {
+      // ...
+    }
+  }
+}
+```
```
hacksparrow: Should document `@config.view`.

hacksparrow: Probably `@config` should have a page of its own.

docs/site/Context.md
```diff
@@ -530,3 +530,98 @@ class MyController {
...
+
+The following APIs are available to enforce/leverage this convention:
+
+1. ctx.configure('servers.RestServer.server1') => Binding for the configuration
+2. ctx.getConfig('servers.RestServer.server1') => Get configuration
+3. `@config` to inject corresponding configuration
+4. `@config.getter` to inject a getter function for corresponding configuration
+5. `@config` to inject a `SingleValuedContextView` for corresponding
+   configuration
+
```
hacksparrow: Should add `@config.view()` too. Probably `@config` should have a page of its own.

hacksparrow: Should document `configPath`.

docs/site/Context.md
raymondfeng: We use `$config` as a special suffix to avoid possible conflict.

hacksparrow: Doesn't feel consistent with the rest of the API - the only place where you access the value with a key prefixed with a `$`. Also, technically it doesn't prevent conflict.

What could possibly set a `config` value on `server1`, apart from its own configuration? Maybe `config` should be implemented as a getter-setter?

docs/site/Context.md
```diff
@@ -530,3 +530,101 @@ class MyController {
...
+  }
+  // ...
+}
+```
+
+The `@config.*` decorators can take an optional `configPath` parameter to allow
+the configuration value to be a deep property of the bound value. For example,
+`@config('port')` injects `RestServerConfig.port` to the target.
+
+```ts
```
hacksparrow: Is this example supposed to show `@config('port')` in action? If yes, it doesn't contain relevant code.

hacksparrow:
@raymondfeng `@config()` looks like it is a decorator for setting the **configuration** object of a class (I think this requirement was the inspiration for the PR). However, it is able to to add any root-level methods and properties to a class. The ability to configure options for a class is just a subset of the functionality of this decorator.

From the behavior of this decorator, it looks like it would be named `@extend()` or something along that line, rather than `@config()`.

raymondfeng:
>  However, it is able to to add any root-level methods and properties to a class. The ability to configure options for a class is just a subset of the functionality of this decorator.

Where do you get that impression? @config is just a sugar to inject for configuration values. It cannot change the class.

hacksparrow:
> @config is just a sugar to inject for configuration values.

I agree that is the intent. However, our classes don't have a configuration property which would hold the configuration values. In our case, configurations are instance properties.

Here is an example of adding a new method on an instance of a class.

```ts
class RestServerWithPort {
  constructor(@config('hello') public hello: Function) {}
}

ctx.configure('servers.rest.server1').to({hello: () => console.log('Hello World!')});
ctx.bind('servers.rest.server1').toClass(RestServerWithPort);

const server = await ctx.get<RestServerWithPort>('servers.rest.server1');
server.hello();
```

hacksparrow:
I am wondering if there should be an intermediary between `@config` values and the properties being affected on the instance.

docs/site/Context.md
hacksparrow: @raymondfeng @bajtos thoughts on the comment above?

raymondfeng:
@hacksparrow The example you described can be achieved by `@inject` itself. In fact, a JavaScript class or object is open by default. You add additional methods by setting a property with function value. Why is it a concern for you?

docs/site/Context.md
bajtos: My understanding:

The idea is to treat `$` as a reserved character that should not be used by user-defined bindings, only by the framework itself. 

- `servers.RestServer.server1.config` is regular binding key that anybody can define and bind values to.
- `servers.RestServer.server1.$config` is an internal key used by the framework only. App & extension developers should never use this key directly, but use helpers like `@config` and `ctx.configure` instead.

docs/site/Context.md
```diff
@@ -530,3 +530,101 @@ class MyController {
...
+
+- `servers.RestServer.server1`: RestServer
+- `servers.RestServer.server1.$config`: RestServerConfig
+
+- `servers.RestServer.server2`: RestServer
+- `servers.RestServer.server2.$config`: RestServerConfig
+
+The following APIs are available to enforce/leverage this convention:
+
+1. ctx.configure('servers.RestServer.server1') => Binding for the configuration
```
bajtos: I am concerned that in order to contribute configuration, the code needs a reference to the target Context instance. It creates a coupling that makes it difficult for extensions to contribute configuration bindings.

Can you add API for creating standalone configuration bindings (no context instance required)? Perhaps a short-cut leveraging `configBindingKeyFor` helper?

For example, we can add a static method `Binding.configure` that will create a new `Binding` instance with the correct configuration key:

```ts
// in a component, export the following binding:
class MyComponent {
  bindings: [
    // some other bindings

    Binding.configure('servers.RestServer.server1').to({port: 0})
  ],
});

// in mountComponent - already in place
if (component.bindings) {
  for (const binding of component.bindings) {
    app.add(binding);
  }
}
```

I am not sure if `Binding.configure` is the right name, feel free to pick a better one!

I am also fine to leave out such new helper out of scope of this pull request, as long as the documentation explains how to create standalone Binding instances for configuration (e.g. using `configBindingKeyFor`)

docs/site/Context.md
bajtos: +1

The example is missing code showing how `RestServer` is calling `@config('port')`.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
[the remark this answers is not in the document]
bajtos: > Please note `@inject` does not leverage `session.currentBinding`.

This is very important, please describe the difference in the documentation.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
```diff
@@ -0,0 +1,192 @@
...
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  const LOGGER_KEY = 'loggers.Logger';
+  it('injects a getter function to access config', async () => {
+    class Logger {
+      constructor(
+        @config.getter()
+        public configGetter: () => Promise<LoggerConfig | undefined>,
```
bajtos: Would the following work?

```suggestion
        public configGetter: Getter<LoggerConfig | undefined>, 
```

Also why are we expecting to receive `undefined` too? Is `@config.getter` enabling `{optional: true}` under the hood? If that's the case, then please 
- add a code comment here to explain this fact to readers
- add a new test (or enhance this test) to test what happens when `configGetter` is called but no configuration was bound

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
```diff
@@ -0,0 +1,192 @@
...
+import {config, configBindingKeyFor, Context, ContextView} from '../..';
+
+describe('Context bindings - injecting configuration for bound artifacts', () => {
+  let ctx: Context;
+
+  beforeEach(givenContext);
+
+  it('binds configuration independent of binding', async () => {
+    // Bind configuration
+    ctx.configure('servers.rest.server1').to({port: 3000});
```
bajtos: The way how this test is written, it's not clear at all that RestServer is using `@config` with no binding key set, therefore it will pick up configuration associated with the current RestServer binding being resolved.

Let's add a new test to verify this specific scenario.

```ts
class RestServer {
  constructor(
    // Notice there is no binding key specified for the @config() decorator
    // @config will pick configuration binding associated with the binding key 
    // of the server being instantiated (resolved)
    @config() public configObj: RestServerConfig
  ) {}
}

ctx.bind('server1').toClass(RestServer);
ctx.bind('server2').toClass(RestServer);

ctx.configure('server1').to({port: 3000});
ctx.configure('server2').to({port: 8000});

// server1 picks up config for server1
const server1 = ctx.get<RestServer>('server1');
expect(server1.configObj).to.eql({port: 3000});

// server2 picks up config for server2
const server2 = ctx.get<RestServer>('server2');
expect(server2.configObj).to.eql({port: 8000});
```

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -456,14 +458,14 @@ describe('Context bindings - Injecting dependencies of classes', () => {
 
   it('injects a nested property', async () => {
     class TestComponent {
-      constructor(@inject('config#test') public config: string) {}
+      constructor(@inject('config#test') public configVal: string) {}
```
bajtos: How about renaming `configVal` to `settings` (everywhere in this file).

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      constructor(@config('x.y') public optionXY: string) {}
+    }
+
+    ctx.configure('store').to({x: {y: 'y'}});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.optionXY).to.eql('y');
+  });
+
+  it('injects config if the binding key is not present', () => {
```
bajtos: The test name seems to describe a scenario that's different from the actual test code. Is this testing `@config()` invocation with no property accessor? 

The first `@config` argument is called `configPath` and described as `Optional property path of the config.`. Let's use "config property path" instead of "binding key" in the test names.

```suggestion
  it('injects entire config object if there is no config property path present', () => {
```

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      constructor(@config() public configObj: object) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.configObj).to.eql({x: 1, y: 'a'});
+  });
+
+  it("injects config if the binding key is ''", () => {
```
bajtos:
```suggestion
  it("injects entire config object if the config property path is ''", () => {
```

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      constructor(@config('') public configObj: object) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.configObj).to.eql({x: 1, y: 'a'});
+  });
+
+  it('injects config if the binding key is a path', () => {
```
bajtos: This does not make much sense to me. The first argument is always a property path and we already have tests to verify that the correct config property is injected. Please reword the test name or remove the test entirely.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      constructor(@config('x') public optionX: number) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.optionX).to.eql(1);
+  });
+
+  it('injects undefined option if key not found', () => {
```
bajtos:
```suggestion
  it('injects undefined if config property path not found', () => {
```

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      constructor(@config('not-exist') public option: string | undefined) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.option).to.be.undefined();
+  });
+
+  it('injects a config property based on the parent binding', async () => {
```
bajtos: What is the parent binding? The only parent-child relation I am aware of is between contexts, e.g. Application context is a parent of Request context.

Please improve the test name to better describe the scenario being tested.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
bajtos: Maybe the test I am asking for is already implemented as `it('injects a config property based on the parent binding'`?

packages/context/src/__tests__/unit/context-config.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+  });
+
+  describe('getConfig()', () => {
+    it('gets config for a binding', async () => {
+      ctx.configure('foo').toDynamicValue(() => Promise.resolve({x: 1}));
+      expect(await ctx.getConfig('foo')).to.eql({x: 1});
+    });
+
+    it('gets local config for a binding', async () => {
```
bajtos: What is "local config"? IIUC, you are testing property access, let's use that term in the test name please.

```suggestion
    it('gets a property of the config object for a binding', async () => {
```

docs/site/Context.md
```diff
@@ -530,3 +530,101 @@ class MyController {
...
+2. ctx.getConfig('servers.RestServer.server1') => Get configuration
+3. `@config` to inject corresponding configuration
+4. `@config.getter` to inject a getter function for corresponding configuration
+5. `@config.view` to inject a `ContextView` for corresponding configuration
```
bajtos: Please explain that:

All configuration accessors (`ctx.getConfig`, `@config*`) consider the configuration binding as optional, i.e. return `undefined` if no configuration was bound. This is different from `ctx.get` and `@inject` APIs, which treat bindings as required and throw an error when the requested binding is not found.

packages/context/src/__tests__/unit/context-config.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+  describe('getConfigSync()', () => {
+    it('gets config for a binding', () => {
+      ctx.configure('foo').to({x: 1});
+      expect(ctx.getConfigSync('foo')).to.eql({x: 1});
+    });
+
+    it('throws a helpful error when the config is async', () => {
+      ctx.configure('foo').toDynamicValue(() => Promise.resolve('bar'));
+      expect(() => ctx.getConfigSync('foo')).to.throw(
+        /Cannot get config\[\] for foo synchronously: the value is a promise/,
```
bajtos: Where is the `[]` coming from? Is it because there is no property path specified? Can we leave it out in such case?

Please consider adding a test for `ctx.getConfigSync('key', 'path')` too.

packages/context/src/__tests__/unit/context-config.unit.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  describe('configResolver', () => {
+    class MyConfigResolver implements ConfigurationResolver {
+      getConfigAsValueOrPromise<ConfigValueType>(
+        key: BindingAddress<unknown>,
+        configPath?: string,
+        resolutionOptions?: ResolutionOptions,
+      ): ValueOrPromise<ConfigValueType | undefined> {
+        return ((key + '.config') as unknown) as ConfigValueType;
```
bajtos: I find this confusing. We use `.$config` suffix for config binding keys. When I see `.config` here, I have to think hard whether it's a type (missing `$` character), or whether this is just a random string.

Let's do this:

```suggestion
        return (`dummy config for ${key}`as unknown) as ConfigValueType;
```

packages/context/src/__tests__/unit/context-view.unit.ts
```diff
@@ -35,6 +35,17 @@ describe('ContextView', () => {
...
+  it('reports error on singleValue() if multiple values exist', async () => {
+    return expect(taggedAsFoo.singleValue()).to.be.rejectedWith(
+      /The ContextView has more than one values\. Use values\(\) to access them\./,
+    );
+  });
+
+  it('supports singleValue() if only one value exist', async () => {
+    server.unbind('bar');
+    expect(await taggedAsFoo.singleValue()).to.eql('FOO');
+  });
```
bajtos: Let's land this improvement in a standalone pull request please. I'd like to see a dedicated entry in the CHANGELOG.

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -533,6 +533,21 @@ describe('Context', () => {
     });
   });
 
+  describe('getOwnerContext', () => {
```
bajtos: Ditto: Let's land this improvement in a standalone pull request please. I'd like to see a dedicated entry in the CHANGELOG.

packages/context/src/binding-key.ts
bajtos: > Please add an empty line to visually delimit individual members.

☝️

packages/context/src/binding-key.ts
bajtos: > Can we modify this function to support type checks and inference?
> (...)
> Also since this method is part of BindingKey, can we perhaps shorten it's name to forConfig, with the following use?
> (...)

☝️

packages/context/src/context-view.ts
bajtos: > `else` is not needed here, previous branches always return from the function. Please remove this line, it will make the code less indented and probably easier to read too.

☝️

packages/context/src/context.ts
bajtos: > See my comments above, I think we should pass ConfigValueType to the function creating the binding key. Thoughts?

☝️

packages/context/src/context.ts
bajtos: > Is it a good idea to introduce the sync version? I mean what are the use cases where `await ctx.getConfig()` cannot be used?

☝️

docs/site/Context.md
raymondfeng: Good idea. I'll add `Binding.configure`.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
raymondfeng: Yes.

raymondfeng:
> Is it a good idea to introduce the sync version? I mean what are the use cases where await ctx.getConfig() cannot be used?

I was trying to make it symmetric with `get/getSync` for bindings.

docs/site/Context.md
bajtos: > Good idea. I'll add `Binding.configure`

Please describe the new API in the documentation too. (E.g. add it to the list here.)

docs/site/Context.md
bajtos: > >  Please explain that:
>
> All configuration accessors (`ctx.getConfig`, `@config*`) consider the configuration binding as optional, i.e. return `undefined` if no configuration was bound. This is different from `ctx.get` and `@inject` APIs, which treat bindings as required and throw an error when the requested binding is not found.

I see that you have documented this binding in API docs. Can you please describe it in `docs/site/Context.md` too? In my experience, most people are not going to read our API docs.

packages/context/src/keys.ts
```diff
@@ -29,6 +29,11 @@ export namespace ContextTags {
    */
   export const KEY = 'key';
 
+  /**
+   * Binding tag to associate a configuration binding with the target binding key
+   */
+  export const CONFIGURATION_OF = 'configurationOf';
```
bajtos: `CONFIGURATION_OF` looks a bit weird to me, have you considered any alternative names? E.g. `CONFIGURATION_FOR`?

(Nitpick - feel free to ignore.)

packages/context/src/context-view.ts
```diff
@@ -165,10 +165,9 @@ export class ContextView<T = unknown> extends EventEmitter
     const values = await this.values(session);
     if (values.length === 0) return undefined;
     if (values.length === 1) return values[0];
-    else
-      throw new Error(
-        'The ContextView has more than one values. Use values() to access them.',
-      );
+    throw new Error(
+      'The ContextView has more than one values. Use values() to access them.',
```
bajtos: "one value", not "one values"

```diff
- The ContextView has more than one values.
+ The ContextView has more than one value.
```

packages/context/src/context.ts
```diff
@@ -356,11 +362,125 @@ export class Context extends EventEmitter {
...
+      const configResolver = this.getSync<ConfigurationResolver>(
+        `${BindingKey.CONFIG_NAMESPACE}.resolver`,
+        {
+          optional: true,
+        },
+      );
+      if (configResolver) {
+        this.configResolver = configResolver;
+      } else {
+        // Fallback to DefaultConfigurationResolver
```
bajtos: Please add `debug` logs to tell the user what resolver we are using.

packages/context/src/binding-config.ts
```diff
@@ -0,0 +1,68 @@
...
```
bajtos: Re-posting an older comment.

I think this detailed tsdoc comment should be moved to `ConfigurationResolver` description, so that people implementing `ConfigurationResolver` better understand what to expect in the arguments.

raymondfeng:
> Besides the few remaining comments below, please structure the commit history in such way that there are two standalone commits for the addition of ContextView.prototype.singleValue() and Context.prototype.getOwnerContext() APIs. I'd like to see dedicated CHANGELOG entries.

I'll do during squashing.

docs/site/Context.md
```diff
@@ -574,11 +574,14 @@ expects configuration. For example:
...
 The following APIs are available to enforce/leverage this convention:
 
-1. ctx.configure('servers.RestServer.server1') => Binding for the configuration
-2. ctx.getConfig('servers.RestServer.server1') => Get configuration
-3. `@config` to inject corresponding configuration
-4. `@config.getter` to inject a getter function for corresponding configuration
-5. `@config.view` to inject a `ContextView` for corresponding configuration
+1. `ctx.configure('servers.RestServer.server1')` => Binding for the
+   configuration
+2. `binding.configure(servers.RestServer.server1)` => Creates a accompanying
```
bajtos: I believe the syntax is slightly different?

```ts
Binding.configure('servers.RestServer.server1')
```

packages/context/src/context.ts
```diff
@@ -410,15 +411,20 @@ export class Context extends EventEmitter {
...
-        `${BindingKey.CONFIG_NAMESPACE}.resolver`,
+        ContextBindings.CONFIGURATION_RESOLVER,
         {
           optional: true,
         },
       );
       if (configResolver) {
+        debug(
+          'Custom ConfigurationResolver is loaded from %s.',
+          ContextBindings.CONFIGURATION_RESOLVER.toString(),
```
bajtos: Is it necessary to call `.toString()` here? Wouldn't `debug` call it automatically for `%s`?

raymondfeng: Here is my test:
```
debug('%s', {x: 1, toString: () => 'y'})
  x { x: 1, toString: [Function: toString] } +55s
```

docs/site/Context.md
raymondfeng: Fixed.

packages/context/src/__tests__/acceptance/binding-config.feature.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+# Scenario: configure an artifact with a dynamic source
+
+- Given a `Context`
+- Given a class RestServer with a constructor accepting a single argument
+  `config: RestServerConfig`
+- Given `RestServer` ctor argument is decorated with `@config()`
+- When I bind a configuration factory of `{port: 3000}` to
+  `servers.rest.server1:$config`
```
b-admike: Is it `servers.rest.server1.$config` here instead of `:`?

packages/context/src/__tests__/acceptance/binding-config.feature.md
```diff
@@ -0,0 +1,84 @@
...
+// Bind configuration
+ctx
+  .configure('servers.rest.server1')
+  .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+// Bind RestServer
+ctx.bind('servers.rest.server1').toClass(RestServer);
+
+// Resolve an instance of RestServer
+// Expect server1.config to be `{port: 3000}
```
b-admike: From my previous comment, are we defaulting to `server1.config` or `server1.configObj`?

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+
+  it('injects undefined option if configPath not found', () => {
+    class Store {
+      constructor(@config('not-exist') public option: string | undefined) {}
+    }
+
+    ctx.configure('store').to({x: 1, y: 'a'});
+    ctx.bind('store').toClass(Store);
+    const store: Store = ctx.getSync('store');
+    expect(store.option).to.be.undefined();
```
b-admike: From a UX perspective, wouldn't it be better if we somehow alerted developers that the `configPath` is not found? I understand that `undefined` makes sense because it's optional, but it'll be good (or maybe add a `debug` statement in the source code?) to have IMO.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+    expect(store2.optionX).to.eql(2);
+    expect(store2.optionY).to.eql('b');
+  });
+
+  it('injects undefined config if no binding is present', async () => {
+    class Store {
+      constructor(@config('x') public settings: object | undefined) {}
+    }
+
+    const store = await instantiateClass(Store, ctx);
```
b-admike: Wondering why we're using `instantiateClass` here as opposed to binding the class to a context and using `ctx.get` as previous tests? Also, got a similar question regarding UX as mentioned above.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -693,6 +695,168 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+    expect(store.settings).to.be.undefined();
+  });
+
+  it('injects config from config binding', () => {
```
b-admike: How is this test different than https://github.com/strongloop/loopback-next/pull/2259/files#diff-cf5075a2ae88e5e65480d792a8a2b738R698?

packages/context/src/__tests__/unit/context-view.unit.ts
b-admike: +1 (ignore if feedback applied).

packages/context/src/__tests__/unit/context.unit.ts
b-admike: +1 (ignore if feedback applied).

packages/context/src/__tests__/unit/context-view.unit.ts
```diff
@@ -45,6 +45,17 @@ describe('ContextView', () => {
     expect(await taggedAsFoo.asGetter()()).to.eql(['BAR', 'FOO']);
   });
 
+  it('reports error on singleValue() if multiple values exist', async () => {
+    return expect(taggedAsFoo.singleValue()).to.be.rejectedWith(
```
b-admike: I'm curious why we wanted to have the `singleValue()` method in the first place?

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (!bindingKey) return undefined;
+    return ctx.getConfigAsValueOrPromise(bindingKey, meta.configPath, {
+      session: forkedSession,
+      optional: meta.optional,
+    });
+  };
+}
+
+/**
+ * Resolver for `@config.view`
```
b-admike: I hope our TSDocs generator doesn't pick `@config` as some kind of keyword (from review POV it looks like one).

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+      .configure('servers.rest.server1')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServer);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServer>('servers.rest.server1');
+    expect(server1.configObj).to.eql({port: 3000});
```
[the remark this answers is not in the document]
raymondfeng: No, `server1.configObj` is not a key. It's the typescript reference to `configObj` property injected.

packages/context/src/inject-config.ts
```diff
@@ -0,0 +1,182 @@
...
+    binding =>
+      binding.key === BindingKey.buildKeyForConfig(bindingKey).toString(),
+    injection.metadata.configPath,
+  );
+  view.open();
+  return view;
+}
+
+/**
+ * A Context to handle dynamic configuration
```
b-admike: Nitpick: can we add a bit more info on what the class does?

packages/context/src/__tests__/acceptance/binding-config.feature.md
raymondfeng: `servers.rest.server1:$config` is the binding key for config.

packages/context/src/__tests__/acceptance/binding-config.feature.md
raymondfeng: Against, they are not binding keys.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: The default behavior is documented. It can be overridden with `{optional: false}`.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: Both options are fine. `instantiateClass` allows us to create an instance without binding itself to the context.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: It's pretty similar. In this case, only one property `x` is used.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
b-admike: I see it now at https://github.com/strongloop/loopback-next/pull/2259/files#diff-3735d35dac13ab389eebd5a3dac04521R88.

packages/context/src/context.ts
```diff
@@ -356,11 +363,130 @@ export class Context extends EventEmitter {
...
+  ): ConfigValueType | undefined {
+    const valueOrPromise = this.getConfigAsValueOrPromise<ConfigValueType>(
+      key,
+      configPath,
+      resolutionOptions,
+    );
+    if (isPromiseLike(valueOrPromise)) {
+      const prop = configPath ? ` property ${configPath}` : '';
+      throw new Error(
+        `Cannot get config${prop} for ${key} synchronously: the value is a promise`,
```
b-admike: Nitpick: should we point devs to use `getConfig` instead?

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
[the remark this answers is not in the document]
b-admike: Oh I see the definition of `RestServer` below now :). Got it!

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
b-admike: Ok was wondering if we could just reduce it to one test, but ok with both.

packages/context/src/__tests__/acceptance/binding-config.feature.md
b-admike: It's written as `.$config` from https://github.com/strongloop/loopback-next/pull/2259/files/28db47cd98ed35288764fd47c33b8c806e87cd3c#diff-3ecbb2bbc45d751a5c7190943c36f277R570.

packages/context/src/inject-config.ts
raymondfeng: It will not as it's quoted with `.

packages/context/src/__tests__/unit/context-view.unit.ts
raymondfeng: It's for dynamic configuration - `@config.view`. See acceptance tests for refs.

b-admike: Okay, but we can still use `values()` for that too no?

docs/site/Context.md
hacksparrow: @raymondfeng there's not mention of `@config('port')` in the example.

packages/context/src/__tests__/unit/context-view.unit.ts
raymondfeng: Yes, we can still use `values()`. `singleValue()` is just a helper.

Issue #1144, opened 2018-03-19
The key controller.current.ctor was not bound to any value.


NiklasA:
## Description / Steps to reproduce / Feature proposal:
Hi everyone, 
I added @loopback/authentication to my LoopBack 4 application (like descriped in [README.md](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/README.md) and I always get the error: "The key controller.current.ctor was not bound to any value.". Does anyone know what I have to do, any hint, where I have to look? Thanks in advance!

## Current Behavior:
Every time, when I call my TestController: `curl -u username:password http://localhost:3000/whoami`. I get the following error: Error: The key controller.current.ctor was not bound to any value.

raymondfeng:
@NiklasA It would be helpful if you can either describe steps to reproduce the problem or having a sandbox repo to contain the test app.

NiklasA:
Hi @raymondfeng,

I found my issue. The problem was that I integrated `@loopback/authentication` and `example-log-extension` both and that I merged both snippets wrong in `sequence.ts`. I had to move `await this.authenticateRequest(req);` into try / catch block.

**Before:**
```javascript
  async handle(req: ParsedRequest, res: ServerResponse) {
    // we define these variable outside of try / catch so that they can be accessed by logger.
    let args: any = [];
    let result: any;

    // this is the important line added to the default sequence implementation.
    await this.authenticateRequest(req);

    // optionally start timing the sequence using the timer function available via LogFn.
    const start = this.logger.startTimer();

    try {
      const route = this.findRoute(req);
``` 

**After:**
```javascript
  async handle(req: ParsedRequest, res: ServerResponse) {
    // we define these variable outside of try / catch so that they can be accessed by logger.
    let args: any = [];
    let result: any;

    // optionally start timing the sequence using the timer function available via LogFn.
    const start = this.logger.startTimer();

    try {
      const route = this.findRoute(req);

      // this is the important line added to the default sequence implementation.
      await this.authenticateRequest(req);
```

The second mistake was, that I injected `@inject(RestBindings.Http.CONTEXT) public ctx: Context` in constructor of MySequence. Do you know, why it is not possible to inject Context, when I use @loopback/authentication?

Thank you in advance.
Niklas

raymondfeng:
[...]

PR #2271, opened 2019-01-21
docs: update docs for serving static files

joeytwiddle:
Document two common gotchas for the static router.

Today I had to set up static routing in our app. I stumbled over two gotchas on the way. To make it easier for others to set up in future, I suggest adding them to the docs.

1. Putting `this.static('/', ___)` in the constructor will break all controllers! We must leave it until after boot.
2. If authentication is enabled, it conflicts with static requests. (The fix here https://github.com/strongloop/loopback-next/issues/1144#issuecomment-438359985 is popular, so I added it to the docs.)

## Checklist

## Version

I am using:
@loopback/authentication@1.0.2
@loopback/boot@1.0.2
@loopback/core@1.0.0
@loopback/http-server@1.0.0
@loopback/rest@1.1.0

### DIFF
--- a/docs/site/Serving-static-files.md
+++ b/docs/site/Serving-static-files.md
@@ -37,9 +37,21 @@ export class TodoListApplication extends BootMixin(
 
     // ...
 
-    this.static('/', path.join(__dirname, '../../public'));
+    this.static('/public', path.join(__dirname, '../../public'));
   }
 }
+```
+
+If the path prefix overlaps with controller routes (e.g. `'/'`), then you should
+only add the static route _after_ the controllers have been loaded.
+
+```ts
+  async boot() {
+    await super.boot();
+
+    // After all the controllers have loaded, add this fallback route
+    this.static('/', path.join(__dirname, '../../public'));
+  }
 ```
 
 You can call `app.static()` multiple times to configure the app to serve static
@@ -67,4 +79,25 @@ And `app.static()` can be called even after the app have started.
 await app.boot();
 await app.start();
 app.static('/files', path.join(__dirname, 'files'));
+```
+
+### Authentication
+
+The static router
+[does not](https://github.com/strongloop/loopback-next/issues/1144#issuecomment-438359985)
+work smoothly with authentication. If you are using authentication, then you
+should change the way `authenticateRequest` is called in `sequence.ts` to
+operate like this:
+
+```ts
+  async handle(context: RequestContext) {
+    // ...
+
+      // Only attempt authentication on non-static routes, because authenticateRequest fails on static routes
+      if (!(route instanceof StaticAssetsRoute)) {
+        await this.authenticateRequest(request);
+      }
+
+    // ...
+  }
 ```


docs/site/Serving-static-files.md
```diff
@@ -37,11 +37,23 @@ export class TodoListApplication extends BootMixin(
...
+
+```ts
+  async boot() {
+    await super.boot();
+
+    // After all the controllers have loaded, add this fallback route
+    this.static('/', path.join(__dirname, '../../public'));
+  }
+```
+
```
joeytwiddle: This is the solution to the aforementioned issue. It might be simpler if we just recommend this approach by default, and drop the earlier docs which use the constructor. Would you like me to do that?

docs/site/Serving-static-files.md
```diff
@@ -68,3 +80,24 @@ await app.boot();
...
+    // ...
+
+      // Only attempt authentication on non-static routes, because authenticateRequest fails on static routes
+      if (!(route instanceof StaticAssetsRoute)) {
+        await this.authenticateRequest(request);
+      }
+
+    // ...
+  }
+```
```
joeytwiddle: An alternative solution might be to wrap this check into `authenticateRequest`...

docs/site/Serving-static-files.md
```diff
@@ -37,11 +37,23 @@ export class TodoListApplication extends BootMixin(
 
     // ...
 
-    this.static('/', path.join(__dirname, '../../public'));
+    this.static('/public', path.join(__dirname, '../../public'));
```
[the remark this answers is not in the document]
bajtos: That's weird. What is the version of your `@loopback/rest` dependency? I believe we have fixed this issue. Static routes should be always defined at the end, after any Controller-contributed routes. This is important for performance reasons too.

docs/site/Serving-static-files.md
```diff
@@ -68,3 +80,24 @@ await app.boot();
...
+work smoothly with authentication. If you are using authentication, then you
+should change the way `authenticateRequest` is called in `sequence.ts` to
+operate like this:
+
+```ts
+  async handle(context: RequestContext) {
+    // ...
+
+      // Only attempt authentication on non-static routes, because authenticateRequest fails on static routes
+      if (!(route instanceof StaticAssetsRoute)) {
```
bajtos: Can we instead fix `@loopback/authentication` please? I understand this doc change is a quick improvement, but I think everybody would be better of if we fixed the problem properly.

bajtos: Re-posting a part of https://github.com/strongloop/loopback-next/issues/1144#issuecomment-438368225

Most importantly, I think we need to fix our reference `authenticateRequest` function to correctly support `StaticAssetsRoute` and any other custom route that apps or extensions my register via `app.route()` API.

bajtos:
@joeytwiddle ping 👋 What's the status of this pull request? Are you still keen to finish it?

joeytwiddle:
Apologies, I have not had time to check this.

It sounds like `this.static()` has been fixed.

I'm not sure about authentication. But if you plan to fix that inside the authentication module, then these docs won't be needed.

I will close this PR, and hopefully check again when I next update our loopback dependencies.

Issue #1920, opened 2018-10-26
Feature parity with LoopBack 3.x (and the lack of it)


bajtos:
This epic is keeping a track of features that are available in LoopBack 3.x but don't have an easy-to-use alternative in LoopBack 4.x.

- [ ] More relations #1450
  - [x] HasOne #1422
  - [x] HasManyThrough #2264
  - [ ] 🚶HasAndBelongsToMany #2308 
  - [ ] Polymorphic #2487
  - [ ] EmbedsOne 
  - [ ] EmbedsMany
  - [ ] ReferencesMany
- [x] CLI for model relation https://github.com/strongloop/loopback-next/issues/1359
- [ ] Further REST layer improvements #1452
  - [ ] streaming requests/responses
  - [x] handling file uploads #1873
  - [x] Express Middleware:  #1293, #2035, #1982. Possibly more.
- [x] 🚶Authentication #1035 (plus a bunch of other stories, e.g. #1334)
- [x] Authorization #538
- [x] Include related models in query results (#1352, #1721, #1889)
  - _A note on `findById`: filtering was enabled by #4114. The task #1721 somewhat optional, it mostly improves the type descriptions (TypeScript, OpenAPI schema)._
- [x] Model migration (in LB 3.x scope) https://github.com/strongloop/loopback-next/pull/2059 (further work beyond LB3 scope: https://github.com/strongloop/loopback-next/issues/487)
- [x] Model discovery (in LB 3.x scope) https://github.com/strongloop/loopback-next/issues/1949
- [x] From model definition to REST API with no repository/controller classes #2036 #565
- [x] Boot scripts #2034
- [ ] Operation hooks: #1919
- Plain JavaScript experience [ON HOLD] #560

See also a loosely related [Migration from LoopBack 3.x](https://github.com/strongloop/loopback-next/issues/1849)

dhmlau:
Hi everyone, we've identified the feature parity gap as listed above and created separate issues in `loopback-next` repo.  We'd like to hear from you about which one you'd like to see first, so that it helps us decide. :)  

### How it works
1. Review the feature parity gap list in the [original description of this issue](https://github.com/strongloop/loopback-next/issues/1920#issue-374262517)
2. For the one you'd like to see first, upvote it.  Of course, feel free to comment.

### What if my wishlist is not on this list? 
Create a new issue in this repo and put a comment in this issue.  

### How do I see which ones are popular requests from the community? 
[...]

PR #2282, opened 2019-01-24
docs: remove broken links

dhmlau:
When looking up some information, I found out some broken links and irrelevant information.  
In this PR, I'm proposing to remove links that no longer exist or relevant.

### DIFF
--- a/docs/site/FAQ.md
+++ b/docs/site/FAQ.md
@@ -17,36 +17,22 @@ summary: LoopBack 4 is a completely new framework, also known as LoopBack-Next.
 
 See [Crafting LoopBack 4](Crafting-LoopBack-4.md) for more details.
 
-### What’s the timeline for LoopBack 4?
-
-See
-[Upcoming releases](https://github.com/strongloop/loopback-next/wiki/Upcoming-Releases).
-
 ### Where are the tutorials?
 
 See [Examples and tutorials](Examples-and-tutorials.md).
 
-### What features are planned ?
+### What features are planned?
 
-- 100% promise-based APIs and async and await as first-class keywords.
-- Being able to choose to use JavaScript or TypeScript.
-- Better extensibility, ability to override any aspect of the framework (for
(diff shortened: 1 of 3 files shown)


docs/site/FAQ.md
```diff
@@ -17,16 +17,11 @@ summary: LoopBack 4 is a completely new framework, also known as LoopBack-Next.
...
-
-See
-[Upcoming releases](https://github.com/strongloop/loopback-next/wiki/Upcoming-Releases).
-
 ### Where are the tutorials?
 
 See [Examples and tutorials](Examples-and-tutorials.md).
 
-### What features are planned ?
+### What features are planned?
```
bajtos: Should we split this section in two parts - what has been done and what is being planned? We can also point readers to "LB3 Feature Parity" issue.

dhmlau: Actually, I was thinking whether it's better to remove this list completely because features in plan are dynamic and changes over time.  Maybe it's better to put something like: 
Our roadmap is captured in <link to github search with "roadmap" label>.  You can also check out our monthly milestone plans: https://github.com/strongloop/loopback-next/labels/Monthly%20Milestone.  

For LoopBack 3 users, here is the list of LB3 Feature parity we plan to implement in the future: https://github.com/strongloop/loopback-next/issues/1920.

What do you think?

bajtos: Your proposal looks good 👍

hacksparrow:
We should actually run an automated tool to find all the broken links and fix/remove them.

bajtos:
> We should actually run an automated tool to find all the broken links and fix/remove them.

It would be even better to run such tool as part of our CI builds, similarly to how we run commit-lint and code-lint. Such setup is out of scope of this pull request though.

dhmlau:
@hacksparrow, thanks for the suggestion.  Do you have some tools in mind that we can add to CI builds?  Either way, could you please open a new issue? Thanks.

PR #2283, opened 2019-01-24
build(testlab): move test files to `src/__tests__`

bajtos:
I'd like to resume work on migrating our monorepo to TypeScript Project References, as investigated in https://github.com/strongloop/loopback-next/pull/1636. As I see it, the next step is to simplify our project setup by moving our test files from `test` to `src/__test__`.

- Source files are compiled from `src/{foo}` to `dist/{foo}`, the same pattern is applied to test files too.

- Both TypeScript sources and JavaScript output are stored in the same path relative to project root. This makes it much easier to refer to test fixtures.

- ~~There is no need for hacky manually-mainitained index files in package root. We can directly use TypeScript output: `dist/index.js` as the main package file, `dist/index.d.ts` as the type definitions.~~

This change is also enabling future improvements, for example TypeScript project references and migration to jest test runner.

**IMPORTANT NOTE**

I am intentionally migrating only one of our packages in this pull request. By limiting the changes to a single package, the pull request is smaller and easier to review in detail. Once we agree on the best way how to make these changes, I'll open follow-up pull requests to migrate other monorepo packages accordingly.

## Checklist


lerna.json
```diff
@@ -23,8 +23,7 @@
       "prefix": false,
       "loglevel": "silent",
       "stream": true,
-      "concurrency": 8,
-      "sort": false
```
bajtos: It is important to compile dependencies first, dependents second. A package cannot be consumed before its compiled, because `types` and `main` are pointing to files that don't exist yet.

packages/testlab/package.json
```diff
@@ -39,10 +41,10 @@
     "README.md",
     "index.js",
     "index.d.ts",
-    "dist/src",
-    "dist/index*",
+    "dist",
     "should-as-function.d.ts",
-    "src"
+    "src",
+    "!*/__tests__"
```
bajtos: This excludes both `src/__tests__` and `dist/__tests__` from the npm package tarball.

lerna.json
raymondfeng: We should keep `"concurrency": 8,` as it applies to modules that can be built in parallel. For example, with the following deps:
```
a -> b -> d
   -> c -> d
```
Module `b` and `c` can be complied independently.

bajtos:
I just realized that I forgot to check that developer experience in VS Code is preserved with these changes in place, per https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#how-to-test-infrastructure-changes

This PR should not be landed until I verify those aspects.

lerna.json
bajtos: I am keeping that setting, here is the full diff:

```diff
- "concurrency": 8,
- "sort": false
+ "concurrency": 8
```

bajtos:
As kind of expected, "Go to definition" in VSCode is jumping to `.d.ts` files after this change :(

bajtos:
>  "Go to definition" in VSCode is jumping to `.d.ts` files 

This can be easily fixed by adding `"declarationMap": true` to compiler options in `packages/build/config/tsconfig.common.json`.

Unfortunately, that does not solve the second important feature: refactorings, e.g. "rename symbol".

bajtos:
To make refactorings work, I had to bring package-level index files back. I hope we will be able to remove them when we switch to project references and abandon the current approach where we pretend for VSCode that the entire monorepo is a single typescript project.

bajtos:
@raymondfeng the patch is ready for final review and landing, PTAL.

raymondfeng:
@bajtos The following settings in `package.json` seem to be working for me if we remove `index.*` at root level.
```json
"main": "./dist/index.js",
"types": "./src/index.ts",
```

Can you verify?

raymondfeng:
What's going to happen for `fixtures`? I'm seeing two `fixtures` under `testlab`:

- /fixtures
- /src/__tests__/fixtures

bajtos:
@raymondfeng 

> The following settings in `package.json` seem to be working for me if we remove  `index.*` at root level.
> ```json
> "main": "./dist/index.js",
> "types": "./src/index.ts",
> ```
>
> Can you verify?

Nice hack! It works for me well 👍 

I am concerned about the ramifications of such change for our consumers though, I posted a question on StackOverflow to get a qualified answer (hopefully ): https://stackoverflow.com/questions/54399058/ramifications-of-exporting-original-ts-files-instead-of-d-ts-files-as-module-t. Feel free to post an answer yourself.

Considering that switching from hand-written `/index*` files to `"types": "./src/index.ts"` is a very small change that should not introduce any merge conflicts, I prefer to leave it out of scope of this pull request and wait until we have better understanding of the problem.

I think that once we switch to project references, we may not need this hack anymore 🤞 , so it may be better to not bother at all and remove `/index*` files as part of switching to project references.

Thoughts?

> What's going to happen for `fixtures`? I'm seeing two `fixtures` under `testlab`:
> 
> - /fixtures
> - /src/__tests__/fixtures

The directory `/fixtures` contains files that we don't want TypeScript to compile and files that TypeScript cannot compile (e.g. `*.json`, `*.txt`, etc.).

The directory `/src/__tests__/fixtures` contains TypeScript sources we want to compile to JavaScript.

What do you think, how can we make this distinction easier to understand for people contributing to LoopBack (including ourselves)? Few alternatives that come to my mind:
 - Create two README files, one in each `fixtures` directory, and explain what kind of files belongs to each directory.
 - Add an entry to README's section `Testing`.
 - Add a new section to `DEVELOPING.md`
 - Any other ideas?

raymondfeng:
Before this PR, we have root level `index.ts` (`export * from './src';`) and `index.d.ts` (export * from './dist';`). Without running `npm build`, `./dist` does not exist. If VSCode is happy with our project layout, I think it gets types from `index.ts`.

Please also note that `sourcemap` files do not contain TypeScript source code as far as I see from `./dist/**/*.map`.

packages/testlab/fixtures/README.md
```diff
@@ -0,0 +1,9 @@
+# Static fixtures
+
+This directory contains test fixture files that we want to use in verbatim and
+also files that TypeScript compiler would not copy from `src` to `dist`.
+
+For example:
+
+- text files
+- JSON files not imported by any `.ts` file, and so one.
```
b-admike: typo: `so one` -> `so on`

bajtos: Good catch 👍  Fixed.

PR #2291, opened 2019-01-25
feat(context): introduce context events and observers for bind/unbind

raymondfeng:
Spin-off from #2122 

- Add support for context to emit bind and unbind events
- Allow observers to register with the context chain and respond to come and go of bindings of interest

Followed-up PRs:

#2122 (depends on #2291)

- Introduce ContextView to dynamically resolve values of matching bindings
- Introduce @inject.view to support dependency injection of multiple bound values
- Extend @inject and @inject.getter to support multiple bound values

#2249
- Add an example package to illustrate how to implement extension point/extension pattern using LoopBack 4’s IoC and DI container

#2259
- Propose new APIs for Context to configure bindings
- Add @inject.config to simplify injection of configuration

## Checklist

### DIFF
--- a/docs/site/Context.md
+++ b/docs/site/Context.md
@@ -229,3 +229,195 @@ class HelloController {
 These "sugar" decorators allow you to quickly build up your application without
 having to code up all the additional logic by simply giving LoopBack hints (in
 the form of metadata) to your intent.
+
+## Context events
+
+The `Context` emits the following events:
+
+- `bind`: Emitted when a new binding is added to the context.
+  - binding: the newly added binding object
+  - context: Owner context of the binding object
+- `unbind`: Emitted when an existing binding is removed from the context
+  - binding: the newly removed binding object
+  - context: Owner context of the binding object
+- `error`: Emitted when an observer throws an error during the notification
+  process
+  - err: the error object thrown
+
+When an existing binding key is replaced with a new one, an `unbind` event is
+emitted for the existing binding followed by a `bind` event for the new binding.
+
+If a context has a parent, binding events from the parent are re-emitted on the
+context when the binding key does not exist within the current context.
+
+## Context observers
+
+Bindings can be added or removed to a context object. With emitted context
+events, we can add listeners to a context object to be invoked when bindings
+come and go. There are a few caveats associated with that:
+
+1. The binding object might not be fully configured when a `bind` event is
+   emitted.
+
+   For example:
+
+   ```ts
+   const ctx = new Context();
+   ctx
+     .bind('foo')
+     .to('foo-value')
+     .tag('foo-tag');
+   ctx.on('bind', binding => {
+     console.log(binding.tagNames); // returns an empty array `[]`
+   });
+   ```
+
+   The context object emits a `bind` event when `ctx.bind` method is called. It
+   does not control the fluent apis `.to('foo-value').tag('foo-tag')`, which
+   happens on the newly created binding object. As a result, the `bind` event
+   listener receives a binding object which only has the binding key populated.
+
+   A workaround is to create the binding first before add it to a context:
+
+   ```ts
+   const ctx = new Context();
+   const binding = Binding.create('foo')
+     .to('foo-value')
+     .tag('foo-tag');
+   ctx.add(binding);
+   ctx.on('bind', binding => {
+     console.log(binding.tagMap); // returns `['foo-tag']`
+   });
+   ```
+
+2. It's hard for event listeners to perform asynchronous operations.
+
+To make it easy to support asynchronous event processing, we introduce
+`ContextObserver` and corresponding APIs on `Context`:
+
+1. `ContextObserverFn` type and `ContextObserver` interface
+
+```ts
+/**
+ * Listen on `bind`, `unbind`, or other events
+ * @param eventType Context event type
+ * @param binding The binding as event source
+ * @param context Context object for the binding event
+ */
+export type ContextObserverFn = (
+  eventType: ContextEventType,
+  binding: Readonly<Binding<unknown>>,
+  context: Context,
+) => ValueOrPromise<void>;
+
+/**
+ * Observers of context bind/unbind events
+ */
+export interface ContextObserver {
+  /**
+   * An optional filter function to match bindings. If not present, the listener
+   * will be notified of all binding events.
+   */
+  filter?: BindingFilter;
+
+  /**
+   * Listen on `bind`, `unbind`, or other events
+   * @param eventType Context event type
+   * @param binding The binding as event source
+   */
+  observe: ContextObserverFn;
+}
+
+/**
+ * Context event observer type - An instance of `ContextObserver` or a function
+ */
+export type ContextEventObserver = ContextObserver | ContextObserverFn;
+```
+
+If `filter` is not required, we can simply use `ContextObserverFn`.
+
+2. Context APIs
+
+- `subscribe(observer: ContextEventObserver)`
+
+  Add a context event observer to the context chain, including its ancestors
+
+- `unsubscribe(observer: ContextEventObserver)`
+
+  Remove the context event observer from the context chain
+
+- `close()`
+
+  Close the context and release references to other objects in the context
+  chain. Please note a child context registers event listeners with its parent
+  context. As a result, the `close` method must be called to avoid memory leak
+  if the child context is to be recycled.
+
+To react on context events asynchronously, we need to implement the
+`ContextObserver` interface or provide a `ContextObserverFn` and register it
+with the context.
+
+For example:
+
+```ts
+const app = new Context('app');
+server = new Context(app, 'server');
+
+const observer: ContextObserver = {
+  // Only interested in bindings tagged with `foo`
+  filter: binding => binding.tagMap.foo != null,
+
+  observe(event: ContextEventType, binding: Readonly<Binding<unknown>>) {
+    if (event === 'bind') {
+      console.log('bind: %s', binding.key);
+      // ... perform async operation
+    } else if (event === 'unbind') {
+      console.log('unbind: %s', binding.key);
+      // ... perform async operation
+    }
+  },
+};
+
+server.subscribe(observer);
+server
+  .bind('foo-server')
+  .to('foo-value')
+  .tag('foo');
+app
+  .bind('foo-app')
+  .to('foo-value')
+  .tag('foo');
+
+// The following messages will be printed:
+// bind: foo-server
+// bind: foo-app
+```
+
+Please note when an observer subscribes to a context, it will be registered with
+all contexts on the chain. In the example above, the observer is added to both
+`server` and `app` contexts so that it can be notified when bindings are added
+or removed from any of the context on the chain.
+
+- Observers are called in the next turn of
+  [Promise micro-task queue](https://jsblog.insiderattack.net/promises-next-ticks-and-immediates-nodejs-event-loop-part-3-9226cbe7a6aa)
+
+- When there are multiple async observers registered, they are notified in
+  series for an event.
+
+- When multiple binding events are emitted in the same event loop tick and there
+  are async observers registered, such events are queued and observers are
+  notified by the order of events.
+
+### Observer error handling
+
+It's recommended that `ContextEventObserver` implementations should not throw
+errors in their code. Errors thrown by context event observers are reported as
+follows over the context chain.
+
+1. Check if the current context object has `error` listeners, if yes, emit an
+   `error` event on the context and we're done. if not, try its parent context
+   by repeating step 1.
+
+2. If no context object of the chain has `error` listeners, emit an `error`
+   event on the current context. As a result, the process exits abnormally. See
+   https://nodejs.org/api/events.html#events_error_events for more details.
--- a/packages/context/src/__tests__/unit/context-observer.unit.ts
+++ b/packages/context/src/__tests__/unit/context-observer.unit.ts
@@ -1,0 +1,365 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {promisify} from 'util';
+import {
+  Binding,
+  Context,
+  ContextEventType,
+  ContextObserver,
+  filterByTag,
+} from '../..';
+
+const pEvent = require('p-event');
+const setImmediateAsync = promisify(setImmediate);
+
+/**
+ * Create a subclass of context so that we can access parents and registry
+ * for assertions
+ */
+class TestContext extends Context {
+  get parentEventListeners() {
+    return this._parentEventListeners;
+  }
+  /**
+   * Wait until the context event queue is empty or an error is thrown
+   */
+  waitUntilObserversNotified(): Promise<void> {
+    return this.waitUntilPendingNotificationsDone(100);
+  }
+}
+
+describe('Context', () => {
+  let ctx: TestContext;
+  beforeEach('given a context', createContext);
+
+  describe('observer subscription', () => {
+    let nonMatchingObserver: ContextObserver;
+
+    beforeEach(givenNonMatchingObserver);
+
+    it('subscribes observers', () => {
+      ctx.subscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.true();
+    });
+
+    it('unsubscribes observers', () => {
+      ctx.subscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.true();
+      ctx.unsubscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+    });
+
+    it('allows subscription.unsubscribe()', () => {
+      const subscription = ctx.subscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.true();
+      subscription.unsubscribe();
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+      expect(subscription.closed).to.be.true();
+    });
+
+    it('registers observers on context with parent', () => {
+      const childCtx = new TestContext(ctx, 'child');
+      expect(childCtx.parentEventListeners).to.be.undefined();
+      childCtx.subscribe(nonMatchingObserver);
+      expect(childCtx.parentEventListeners!.has('bind')).to.be.true();
+      expect(childCtx.parentEventListeners!.has('unbind')).to.be.true();
+      expect(childCtx.isSubscribed(nonMatchingObserver)).to.true();
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+    });
+
+    it('un-registers observers on context chain', () => {
+      const childCtx = new Context(ctx, 'child');
+      childCtx.subscribe(nonMatchingObserver);
+      expect(childCtx.isSubscribed(nonMatchingObserver)).to.true();
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+      childCtx.unsubscribe(nonMatchingObserver);
+      expect(childCtx.isSubscribed(nonMatchingObserver)).to.false();
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+    });
+
+    it('un-registers observers on context chain during close', () => {
+      const childCtx = new TestContext(ctx, 'child');
+      childCtx.subscribe(nonMatchingObserver);
+      const parentEventListeners = new Map(childCtx.parentEventListeners!);
+      childCtx.close();
+      for (const [event, listener] of parentEventListeners) {
+        expect(ctx.listeners(event)).to.not.containEql(listener);
+      }
+      expect(childCtx.parentEventListeners).to.be.undefined();
+      expect(childCtx.isSubscribed(nonMatchingObserver)).to.false();
+    });
+
+    function givenNonMatchingObserver() {
+      nonMatchingObserver = {
+        filter: binding => false,
+        observe: () => {},
+      };
+    }
+  });
+
+  describe('event notification', () => {
+    const events: string[] = [];
+    let nonMatchingObserverCalled = false;
+
+    beforeEach(givenObservers);
+
+    it('emits one bind event to matching observers', async () => {
+      ctx.bind('foo').to('foo-value');
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+      ]);
+      expect(nonMatchingObserverCalled).to.be.false();
+    });
+
+    it('emits multiple bind events to matching observers', async () => {
+      ctx.bind('foo').to('foo-value');
+      ctx.bind('xyz').to('xyz-value');
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+        'SYNC:xyz:xyz-value:bind',
+        'ASYNC:xyz:xyz-value:bind',
+      ]);
+    });
+
+    it('emits unbind event to matching observers', async () => {
+      ctx.bind('foo').to('foo-value');
+      await ctx.waitUntilObserversNotified();
+      ctx.unbind('foo');
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+        'SYNC:foo:foo-value:unbind',
+        'ASYNC:foo:foo-value:unbind',
+      ]);
+      expect(nonMatchingObserverCalled).to.be.false();
+    });
+
+    it('emits bind/unbind events for rebind to matching observers', async () => {
+      ctx.bind('foo').to('foo-value');
+      await ctx.waitUntilObserversNotified();
+      ctx.bind('foo').to('new-foo-value');
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+        'SYNC:foo:foo-value:unbind',
+        'ASYNC:foo:foo-value:unbind',
+        'SYNC:foo:new-foo-value:bind',
+        'ASYNC:foo:new-foo-value:bind',
+      ]);
+      expect(nonMatchingObserverCalled).to.be.false();
+    });
+
+    it('does not trigger observers if affected binding is the same', async () => {
+      const binding = ctx.bind('foo').to('foo-value');
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+      ]);
+      ctx.add(binding);
+      await ctx.waitUntilObserversNotified();
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+      ]);
+    });
+
+    it('reports error if an observer fails', () => {
+      ctx.bind('bar').to('bar-value');
+      return expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
+        'something wrong',
+      );
+    });
+
+    it('does not trigger observers registered after an event', async () => {
+      ctx.bind('foo').to('foo-value');
+      process.nextTick(() => {
+        // Register a new observer after 1st event
+        ctx.subscribe((event, binding, context) => {
+          const val = binding.getValue(context);
+          events.push(`LATE:${binding.key}:${val}:${event}`);
+        });
+      });
+
+      await ctx.waitUntilObserversNotified();
+      // The late observer is not triggered
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+      ]);
+      // Emit another event
+      ctx.bind('xyz').to('xyz-value');
+      await ctx.waitUntilObserversNotified();
+      // Now the late observer is triggered
+      expect(events).to.eql([
+        'SYNC:foo:foo-value:bind',
+        'ASYNC:foo:foo-value:bind',
+        'SYNC:xyz:xyz-value:bind',
+        'ASYNC:xyz:xyz-value:bind',
+        'LATE:xyz:xyz-value:bind',
+      ]);
+    });
+
+    function givenObservers() {
+      nonMatchingObserverCalled = false;
+      events.splice(0, events.length);
+      // An observer does not match the criteria
+      const nonMatchingObserver: ContextObserver = {
+        filter: binding => false,
+        observe: () => {
+          nonMatchingObserverCalled = true;
+        },
+      };
+      // A sync observer matches the criteria
+      const matchingObserver: ContextObserver = {
+        observe: (event, binding, context) => {
+          // Make sure the binding is configured with value
+          // when the observer is notified
+          const val = binding.getValue(context);
+          events.push(`SYNC:${binding.key}:${val}:${event}`);
+        },
+      };
+      // An async observer matches the criteria
+      const matchingAsyncObserver: ContextObserver = {
+        filter: binding => true,
+        observe: async (event, binding, context) => {
+          await setImmediateAsync();
+          const val = binding.getValue(context);
+          events.push(`ASYNC:${binding.key}:${val}:${event}`);
+        },
+      };
+      // An async observer matches the criteria that throws an error
+      const matchingAsyncObserverWithError: ContextObserver = {
+        filter: binding => binding.key === 'bar',
+        observe: async () => {
+          await setImmediateAsync();
+          throw new Error('something wrong');
+        },
+      };
+      ctx.subscribe(nonMatchingObserver);
+      ctx.subscribe(matchingObserver);
+      ctx.subscribe(matchingAsyncObserver);
+      ctx.subscribe(matchingAsyncObserverWithError);
+    }
+  });
+
+  describe('event notification for context chain', () => {
+    let app: Context;
+    let server: Context;
+
+    let contextObserver: MyObserverForControllers;
+    beforeEach(givenControllerObserver);
+
+    it('receives notifications of matching binding events', async () => {
+      const controllers = await getControllers();
+      // We have server: 1, app: 2
+      // NOTE: The controllers are not guaranteed to be ['1', '2'] as the events
+      // are emitted by two context objects and they are processed asynchronously
+      expect(controllers).to.containEql('1@server');
+      expect(controllers).to.containEql('2@app');
+      server.unbind('controllers.1');
+      // Now we have app: 2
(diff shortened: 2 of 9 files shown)


packages/context/src/context.ts
```diff
@@ -96,10 +129,75 @@ export class Context {
...
+   *
+   * @param event Event names: `bind` or `unbind`
+   * @param binding Binding bound or unbound
+   */
+  protected notifyListeners(
+    event: ContextEventType,
+    binding: Readonly<Binding<unknown>>,
+  ) {
+    // Notify listeners in the next tick
+    process.nextTick(async () => {
```
bajtos: This isn't going to work well. The promise returned by the async function is ignored by `process.nextTick`. When an error happens, a warning "unhandled promise rejection" is triggered. In the future versions of Node, this can crash the app the same way as "uncaught error" does already.

I'd like us to consider two aspects here.

1. Do we want `notifyListeners` to wait until all listeners processed the notification? In the proposal here, we treat notifications as "fire and forget" and run them kind of in background.

2. If we decide for "fire and forget" mode, how should we handle errors reported by listeners? IMO, application developers should have a way how to catch such errors and handle them in their own way.

I am imagining that the currently proposed solution will make it more difficult to write a reliable test for ContextView.

```ts
const context = new Context();
const view = new ContextView(view);

context.bind('foo').to('bar');
// ^^ triggers notification at some point in the future

const values = await view.resolve();
// ^^ how to ensure this is run only after the view was notified
// about the new binding?

expect(values).to.containEqual('bar');
```

I wanted to propose to wait until all listeners processed the notification:
- Error handling is much easier, we can propagate any errors thrown by listeners up the call chain (down the call stack).
- We avoid subtle errors that can happen when the notification of a listener is executed later than code relying on the outcome of that listener.

I was thinking about the following, just to realize that it won't work either:

```ts
const context = new Context();
const view = new ContextView(view);

await context.bind('foo').to('bar');
// ^^ uh oh, what are we trying to await?
// The outcome of `Binding.prototype.to()`?!

const values = await view.resolve();
expect(values).to.containEqual('bar');
```

Also changing the act of creating a new binding from sync to async is a big change that's possibly backwards incompatible.

My conclusion is that we should keep "fire and forget" approach but need to rework the way how these delayed notifications work. Instead of creating a new `process.nextTick` callback for each new binding, we should maintain a queue of pending notifications and a single `process.nextTick` callback to process the entire queue. 

```ts
/*** schedule a new notification ***/
this._pendingNotifications.push({listener, event, args: [binding]});
if (!this._queueHandlerScheduled) {
  process.nextTick(this._processQueueCallback, this);
  this._queueHandlerScheduled = true;
}

/*** process notifications ***/
_processQueueCallback() {
  this._queueHandlerScheduled = false;
  this._processQueueAsync.then(
    success => {
      // let Context consumers know that we are done, e.g.
      this.emit('NotificationsProcessed');
    },
    error => {
      // signal the error to Context consumers, e.g.
      this.emit('error', error);
    }
  );
}

async _processQueueAsync() {
  const notifications = this. _pendingNotifications;
  this._pendingNotifications = [];

  return Promise.all(notifications.map(
    n => n.listener.listen(n.event, ...n.args)
  ));
}

/*** usage in the tests ***/
context.bind('foo').to('bar');
await pEvent(context, 'NotificationsProcessed');
```

@raymondfeng @hacksparrow @jannyHou what do you think?

Regardless of the solution we decide to implement, I think it's very important to document the reasons why we decided for a particular design & implementation, the alternatives we have rejected and why they were rejected.

I am thinking about _Lightweight Architecture Decision Records_ as recommended by [ThoughWork's technology radar](https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records):

> Much documentation can be replaced with highly readable code and tests. In a world of evolutionary architecture, however, it's important to record certain design decisions for the benefit of future team members as well as for external oversight. Lightweight Architecture Decision Records is a technique for capturing important architectural decisions along with their context and consequences. We recommend storing these details in source control, instead of a wiki or website, as then they can provide a record that remains in sync with the code itself. For most projects, we see no reason why you wouldn't want to use this technique.

If it makes it easier for us, then I am also ok with keeping the documentation inside the source code as a (somewhat longer) code comment.

packages/context/src/context.ts
```diff
@@ -150,6 +248,19 @@ export class Context {
...
+  ): Readonly<Binding<ValueType>>[];
+
+  /**
+   * Find bindings using a filter function
+   * @param filter A function to test on the binding. It returns `true` to
+   * include the binding or `false` to exclude the binding.
+   */
+  find<ValueType = BoundValue>(
+    filter: BindingFilter,
+  ): Readonly<Binding<ValueType>>[];
```
bajtos: @raymondfeng I believe these two overloads are not needed. VSCode has suggested to combine them into a single method signature, which we have already in `master`:

```ts
find<ValueType = BoundValue>(
  pattern?: string | RegExp | BindingFilter,
```

Can you revert these changes in `find` declaration please?

packages/context/test/unit/context.unit.ts
```diff
@@ -681,6 +685,109 @@ describe('Context', () => {
...
+    let matchingListener: ContextEventListener;
+    let nonMatchingListener: ContextEventListener;
+    const events: string[] = [];
+    let nonMatchingListenerCalled = false;
+
+    beforeEach(givenListeners);
+
+    it('emits bind event to matching listeners', async () => {
+      ctx.bind('foo').to('foo');
+      await nextTick();
```
bajtos: I find this rather brittle, it's relying on the internal implementation details of Context class. Can we find a more robust solution please? See my first comment above for a possible option.

packages/context/test/unit/context.unit.ts
```diff
@@ -681,6 +685,109 @@ describe('Context', () => {
...
+      events.splice(0, events.length);
+      nonMatchingListener = {
+        filter: binding => false,
+        listen: (event, binding) => {
+          nonMatchingListenerCalled = true;
+        },
+      };
+      matchingListener = {
+        filter: binding => true,
+        listen: (event, binding) => {
```
bajtos: Please add and verify how async listeners are handled too.

```ts
asyncMatchingListener = {
  filter: binding => true,
  listen: async (event, binding) => { 
    await nextTick(); 
    events.push(`ASYNC:${binding.key}:${event}`); 
  },
};
```

Besides the basic scenarios shown in the existing tests, please cover the following more complex situation too.

There are two listeners (sync first, async second) registered with the context. Now we add two bindings "A" and "B". In which order are notifications delivered?

I believe your current implementation will do the following:
 - tick 0: bindings are created
 - tick 1: listeners are notified about both events. The first (sync) listeners pushes both A and B notifications to `events`. The second listener (async) is waiting for the next tick.
 - tick 2: The async listener handles notifications, and pushes A and B notifications to `events`.
Outcome: `['A', 'B', 'ASYNC:A', 'ASYNC:B']`.

I think the following order of calls will less surprising to LB4 users?
- tick 0: bindings are created
- tick 1: notification about A is processed. The sync listener pushes notification `A` to `events`. The async listeners starts the wait.
- tick 2: The async listener processes notification about A and pushes an entry to `events`. The sync listener receives notification about B and pushes an entry to `events`. The async listener receives notification about B and starts the wait.
- tick 3: The async listener pushes `B` to `events`.
Outcome: `['A', 'ASYNC:A', 'B', 'ASYNC:B']`.

Let's discuss.

bajtos:
My first and my last comments are related, the order in which notifications are processed depends on the way how we schedule queue-processing callbacks.

Besides the impact on the behavior observed by Context consumers, we should also consider performance implications. Scheduling a new `process.nextTick` callback for every new binding is not very efficient.

packages/context/src/context.ts
bajtos: Based on my last comment below, it may be better to process notifications in series.

```diff
 async _processQueueAsync() {
   const notifications = this. _pendingNotifications;
   this._pendingNotifications = [];

-  return Promise.all(notifications.map(
-    n => n.listener.listen(n.event, ...n.args)
-  ));
+  for (const n of notifications) {
+    await n.listener.listen(n.event, ...n.args)
+  }
 }
```

hacksparrow: > As an app developer, I would want to register my error listener on the application context and receive errors from child contexts too, e.g. from per-request contexts.

Yes. We should also we able to find out which context triggered the error. 

> Do we really need to support async listeners? 

I think we should. We should not enforce blocking code in our framework. Developers should have a choice to use sync or async version of an API.

hacksparrow:
The interface looks good. However, as pointed out by Miroslav, notification processing needs some more work, maybe resulting in additions to to the API. Let's the keep the discussion going.

raymondfeng:
@bajtos @hacksparrow Thank you for the feedback. I now use https://github.com/jessetane/queue to handle event notifications. PTAL.

hacksparrow:
Good update @raymondfeng. Some more feedback.

The `filter` method of `ContextEventListener` should also be optionally async and tested. Eg usecase: I want to check the existence of a file asynchronously and listen for binding events.

Test cases should be also added to document how errors are handled in sync and sync `filter` methods.

We should be able to tell where in the context chain, an error originated from.

bajtos:
> I now use https://github.com/jessetane/queue to handle event notifications.

What were the criteria on which you chose this particular module? I did a quick search on npm (https://www.npmjs.com/search?q=queue) and [p-queue](https://www.npmjs.com/package/p-queue) seems to be more popular (p-queue: 283k downloads/month vs queue: 57k downloads/months) and also has more ergonomic API based on promises, not events.

```ts
class Context {
  // ...
  async waitUntilEventsProcessed() {
    await this.eventQueue.onIdle();
  }
}
```

The downside I see is that `p-queue` does forward errors from individual tasks to the queue instance :(

---

Setting the choice of a queue library aside. It's not clear to me what are the benefits of adding a queue library, when we are still calling `process.nextTick` for every binding added? I'd prefer to see a solution where we call `process.nextTick` only once for all bindings created from a single tick of event loop.

packages/context/src/context.ts
bajtos: > I'd prefer to start with implementation that's simpler - i.e. sync listeners only, and wait for user feedback & demand for async listeners before we add all the complexity.

On the other hand, because we want to run even synchronous listeners later, on a clean stack started by `process.nextTick()`, we have to address all issues caused by asynchronicity anyways. At that point the matter of adding support for async listeners is trivial and makes sense to get done.

bajtos:
@raymondfeng please take a look at the following use-case/scenario, I believe it's not covered by your patch yet:

> As an app developer, I want to register my error listener on the application context and receive errors from child contexts too, e.g. from per-request contexts.

hacksparrow:
> FWIW, I was proposing to remove filter property entirely in our earlier discussions.

Hmm, I like this approach. @raymondfeng what do you say?

raymondfeng:
> The filter method of ContextEventListener should also be optionally async and tested. Eg usecase: I want to check the existence of a file asynchronously and listen for binding events.

It is intentional to have `filter` functions to be sync. The purpose is to match bindings of interest for notifications. I don't see a need to perform any async operations.

raymondfeng:
> What were the criteria on which you chose this particular module?

I only found `queue` but I'm open to other modules. As you pointed out, `p-queue` does not emit error events at queue level and it makes error handling very hard.

> Setting the choice of a queue library aside. It's not clear to me what are the benefits of adding a queue library, when we are still calling process.nextTick for every binding added? I'd prefer to see a solution where we call process.nextTick only once for all bindings created from a single tick of event loop.

I'm adding the queue based on your comments (`we should maintain a queue of pending notifications`). Considering most bindings are added at application boot/start, it should not matter too much whether we batch the notifications or not.

raymondfeng:
The key challenge here is to decide when to notify listeners of binding events because our Context APIs are synchronous and fluent, such as `ctx.bind('foo').to('bar').tag('t1');`. A listener needs to know the fully populated binding upon notification. The proposed design in this PR is to use a background task queue to perform notifications in ticks after the binding is added/removed from the context via sync apis.

packages/context/test/unit/context.unit.ts
```diff
@@ -25,6 +29,24 @@ class TestContext extends Context {
     const map = new Map(this.registry);
     return map;
   }
+  /**
+   * Wait until the context event queue is empty
+   */
+  waitUntilEventsProcessed() {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
raymondfeng: We have to await two events: `end` and `error`.

packages/context/test/unit/context.unit.ts
```diff
@@ -5,14 +5,18 @@
...
   BindingScope,
   BindingType,
+  Context,
+  ContextEventListener,
   isPromiseLike,
-  BindingKey,
 } from '../..';
 
+import {promisify} from 'util';
+const setImmediatePromise = promisify(setImmediate);
```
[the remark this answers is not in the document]
raymondfeng: I copied it from https://nodejs.org/api/timers.html#timers_setimmediate_callback_args but I'm open to any convention.

packages/context/test/unit/context.unit.ts
raymondfeng: Added.

packages/context/test/unit/context.unit.ts
raymondfeng: Fixed

raymondfeng:
> As an app developer, I want to register my error listener on the application context and receive errors from child contexts too, e.g. from per-request contexts.

I wonder if we can expose `Context.prototype.eventQueue` as `eventEmitter` so that error handlers can be registered. For example:

```ts
ctx.eventEmitter.on('error', ...);
```

raymondfeng:
@bajtos This is an attempt to make error handling a bit simpler - https://github.com/strongloop/loopback-next/pull/2291/commits/0550e8fdce7dac30fefbd4c5fc32c31ecbcd9e83. We can revert this commit if you guys don't like it.

packages/context/test/unit/context.unit.ts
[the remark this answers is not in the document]
bajtos: By default, `p-event` awaits `error` and converts it to rejected promise. That's why I recommend it so often!

> https://www.npmjs.com/package/p-event#peventemitter-event-options
> Returns a Promise that is fulfilled when emitter emits an event matching event, or rejects if emitter emits any of the events defined in the rejectionEvents option.

packages/context/src/context.ts
```diff
@@ -34,16 +43,90 @@ export class Context {
...
+    this.setupEventHandlers();
+  }
+
+  /**
+   * Set up an internal listener to notify registered observers asynchronously
+   * upon `bind` and `unbind` events
+   */
+  private setupEventHandlers() {
+    // Ideally p-event should allow multiple event types in an iterator
+    this.observeEvent('bind');
```
bajtos: This function is observing multiple event instances, I think `observeEvents` or `observeEventInstances` would be a more appropriate name.

`observeEvent` is returning a promise that you are ignoring here. I find this difficult to spot and possibly misleading. At minimum, please add a comment to explain that `observeEvent` is an async function that we are going to leave running in background. Can we come up with a name that will convey this information without the need for an additional comment?

packages/context/test/unit/context.unit.ts
```diff
@@ -681,7 +694,223 @@ describe('Context', () => {
...
+          nonMatchingObserverCalled = true;
+        },
+      };
+      // A sync observer matches the criteria
+      const matchingObserver: ContextObserver = {
+        observe: (event, binding, context) => {
+          // Make sure the binding is configured with value
+          // when the observer is notified
+          const val = binding.getValue(context);
+          events.push(`1:${binding.key}:${val}:${event}`);
```
bajtos: I find the prefixes `1` and `2` as too cryptic. Can you use `SYNC` and `ASYNC` please?

bajtos:
> Considering most bindings are added at application boot/start, it should not matter too much whether we batch the notifications or not.

Few comments:
- There are few per-request bindings too: current request, response, controller name/method, current user, etc.
- I agree that from user perspective, it should not matter too much whether the notifications are batched or not.
- Regardless of how we decide on batching, the behavior should be clearly described in our docs and verified by our test suite to prevent regression.
- For runtime performance, I think it's important to optimize the number of times we call `process.nextTick`.

bajtos:
> The key challenge is two-fold:
>
> 1.  To decide when to notify listeners of binding events because our Context APIs are synchronous and fluent, such as `ctx.bind('foo').to('bar').tag('t1');`. A listener needs to know the fully populated binding upon notification. 
> 
> 2. How to receive feedback (errors) from context event listeners which are notified asynchronously.

Agreed, you have described it very well 👍

raymondfeng:
> Considering the difficulty of implementing good error handling, should we perhaps mandate that event observers must not throw, and report any unhandled errors via console.warn, or even crash the process the hard way?

+💯.

packages/context/src/context.ts
```diff
@@ -96,10 +186,68 @@ export class Context {
...
+   * @param eventType Event names: `bind` or `unbind`
+   * @param binding Binding bound or unbound
+   */
+  protected async notifyObservers(
+    eventType: ContextEventType,
+    binding: Readonly<Binding<unknown>>,
+  ) {
+    if (this.observers.size === 0) return;
+
+    for (const observer of this.observers) {
```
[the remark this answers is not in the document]
raymondfeng: Fixed.

packages/context/src/context.ts
```diff
@@ -34,16 +43,90 @@ export class Context {
...
+        this.pendingEvents--;
+        this.emit('error', err);
+      }
+    }
+  }
+
+  /**
+   * Wait until event notification is idle
+   */
+  protected async waitForIdle() {
```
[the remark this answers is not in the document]
raymondfeng: Fixed.

packages/context/src/context.ts
```diff
@@ -34,16 +43,90 @@ export class Context {
...
+      eventType,
+    );
+    for await (const binding of bindings) {
+      try {
+        await this.notifyObservers(eventType, binding);
+        this.pendingEvents--;
+        this.emit('idle');
+      } catch (err) {
+        this.pendingEvents--;
+        this.emit('error', err);
```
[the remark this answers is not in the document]
raymondfeng: If no error listeners are registered with the context, the process will crash.

packages/context/src/context.ts
```diff
@@ -34,16 +43,90 @@ export class Context {
...
+    // Create an async iterator from the given event type
+    const bindings: AsyncIterable<Readonly<Binding<unknown>>> = pEvent.iterator(
+      this,
+      eventType,
+    );
+    for await (const binding of bindings) {
+      try {
+        await this.notifyObservers(eventType, binding);
+        this.pendingEvents--;
+        this.emit('idle');
```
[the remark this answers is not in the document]
raymondfeng: fixed.

raymondfeng:
> I noticed you no longer need to call process.nextTick. How are the notifications deferred now? Is this handled by the async iterator provided by p-event? Do we use Promise micro-queue instead of Node.js event-loop-ticks now?

Now it happens within Promise microtask queue as implemented by `p-event`.

packages/context/test/unit/context.unit.ts
raymondfeng: fixed

packages/context/src/context.ts
raymondfeng: fixed

raymondfeng:
@bajtos PTAL

packages/context/src/context.ts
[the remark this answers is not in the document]
bajtos: I find this approach slightly problematic.

As an application developer, I can easily attach `on('error')` handler to the root (application-level) context, but there is no easy way how to attach error handler to child contexts like per-request contexts created for each incoming request. As a result, we end up in a setup where failure of certain observers will trigger our error handler while failure of other observers will not.

I think this is an edge case that will be interesting to very few people, so it's not worth to make our code too complex. At the same time, I don't want our users to be surprised about inconsistency in LB behavior.

I am proposing one of the following changes:

(1) When a new child context is created, it should propagate error events to its parent. Conceptually:

```ts
class Context {
  constructor(parent?: Context) {
    if (parent) this.on('error', (...args) => parent.emit('error', ...args));
  }
}
```

This way it's easy for application developers to setup error handler that will be called for errors from all context observers.

(2) Don't use `error` event at all, print errors to console or crash the process. This may be too drastic, the first option described above is probably a better one.

packages/context/src/context.ts
```diff
@@ -34,16 +52,142 @@ export class Context {
...
+  /**
+   * Wrap the debug statement so that it always print out the context name
+   * as the prefix
+   * @param formatter
+   * @param args
+   */
+  // tslint:disable-next-line:no-any
+  private _debug(formatter: any, ...args: any[]) {
+    /* istanbul ignore if */
+    if (debug.enabled) {
```
bajtos:
```suggestion
    if (!debug.enabled) return;
```

packages/context/src/context.ts
```diff
@@ -34,16 +52,142 @@ export class Context {
...
+      // No need to schedule notifications if no observers are present
+      if (this.observers.size === 0) return;
+      // Track pending events
+      this.pendingEvents++;
+      // Take a snapshot of current observers to ensure notifications of this
+      // event will only be sent to current ones. Emit a new event to notify
+      // current context observers.
+      this.emit(notificationEvent, {
+        binding,
+        observers: new Set(this.observers),
```
bajtos: In your implementation, you are trading off performance of event notification for performance of adding a new observer:
 - when a new observer is added, we preserve the Set instance in `this.observers`
 - when a new event is fired, a new Set instance is created

I expect that events will be triggered more often than new observers are added, therefore we should do a different trade-off:
 - when a new observer is added, a new Set instance is created, it replaces the value stored in `this.observers`
 - when a new event is fired, we take a reference to the Set instance in `this.observers` because this instance is considered as immutable

Feel free to fix this in a follow-up pull request if you prefer.

packages/context/src/context.ts
```diff
@@ -34,16 +52,142 @@ export class Context {
...
+      });
+    });
+    // FIXME(rfeng): p-event should allow multiple event types in an iterator.
+    // Create an async iterator from the given event type
+    const events: AsyncIterable<{
+      binding: Readonly<Binding<unknown>>;
+      observers: Set<ContextObserver>;
+    }> = pEvent.iterator(this, notificationEvent);
+    for await (const event of events) {
```
bajtos:
```suggestion
    for await (const {binding, observers} of events) {
```

packages/context/src/context.ts
```diff
@@ -34,16 +52,142 @@ export class Context {
...
+        // The execution of observers happen in the Promise micro-task queue
+        const binding = event.binding;
+        await this.notifyObservers(eventType, binding, event.observers);
+        this.pendingEvents--;
+        this._debug(
+          'Observers notified for %s of binding %s',
+          eventType,
+          binding.key,
+        );
+        this.emit('observersNotified', {eventType, binding});
```
bajtos: This allocates a new object for each event triggered. It would be more performant to use multiple positional arguments instead:

```ts
this.emit('observersNotified', eventType, binding);
```

Having said that, I don't know how well this works with `p-event` and whether the performance impact is significant enough.

packages/context/test/unit/context-observer.unit.ts
```diff
@@ -0,0 +1,319 @@
...
+    it('reports error if an observer fails', () => {
+      ctx.bind('bar').to('bar-value');
+      return expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
+        'something wrong',
+      );
+    });
+
+    it('does not trigger observers registered after an event', async () => {
+      ctx.bind('foo').to('foo-value');
+      process.nextTick(() => {
```
bajtos: Can you use promise micro-queue instead of next tick?

```suggestion
await Promise.resolve();
```

If not, then please use a promisified version of nextTick.

```ts
// at the top
const nextTickAsync = promisify(process.nextTick);

// here
await nextTickAsync();
```

packages/context/test/unit/context-observer.unit.ts
```diff
@@ -0,0 +1,319 @@
...
+        const anotherObserver: ContextObserver = {
+          observe: (event, binding, context) => {
+            const val = binding.getValue(context);
+            events.push(`LATE:${binding.key}:${val}:${event}`);
+          },
+        };
+        ctx.subscribe(anotherObserver);
+      });
+
+      await ctx.waitUntilObserversNotified();
```
bajtos: I am proposed to rewrite this test case as follows.


```ts
const notified = ctx.waitUntilObserversNotified(); // don't wait for this 

await nextTickAsync(); // or Promise.resolve()
const anotherObserver: ContextObserver = {
  // ...
};
ctx.subscribe(anotherObserver);

await notified;
// verify the expectations now
```

packages/context/test/unit/context-observer.unit.ts
```diff
@@ -0,0 +1,319 @@
...
+      return expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
+        'something wrong',
+      );
+    });
+
+    it('does not trigger observers registered after an event', async () => {
+      ctx.bind('foo').to('foo-value');
+      process.nextTick(() => {
+        // Register a new observer after 1st event
+        const anotherObserver: ContextObserver = {
```
bajtos: I find this rather cumbersome. Can we add an overload to `ctx.subscribe`, one that will accept the observe function directly?

User experience I am looking for:

```ts
ctx.subscribe((event, binding, context) => {
  const val = binding.getValue(context);
  events.push(`LATE:${binding.key}:${val}:${event}`);
});
```

Feel free to leave this out and make the improvement in a follow-up PR.

docs/site/Context.md
```diff
@@ -229,3 +229,150 @@ class HelloController {
...
+   ctx.add(binding);
+   ctx.on('bind', binding => {
+     console.log(binding.tagMap); // returns `['foo-tag']`
+   });
+   ```
+
+2. It's hard for event listeners to perform asynchronous operations.
+
+To make it easy to support asynchronous event processing, we introduce
+`ContextObserver` and corresponding APIs on `Context:
```
bajtos:
```suggestion
`ContextObserver` and corresponding APIs on `Context`:
```

docs/site/Context.md
```diff
@@ -229,3 +229,150 @@ class HelloController {
...
+
+// The following messages will be printed:
+// bind: foo-server
+// bind: foo-app
+```
+
+It's recommended that `ContextObserver` implementations should not throw errors
+in `observe` method. Uncaught errors will be reported as `error` events of the
+context object. If there is no error listener registered, the node process will
+crash.
```
bajtos: Let's add a new heading before this section, e.g. `### Error handling`. That way we can create a URL pointing directly to this part.

Can you also expand this paragraph to describe how are errors affected by context hierarchy (child vs. parent context)? If you implement my proposal to propagate errors from child contexts to the parent context, then we should describe that behavior here.

docs/site/Context.md
```diff
@@ -229,3 +229,150 @@ class HelloController {
...
+
+- `close()`
+
+  Close the context and release references to other objects in the context
+  chain. Please note observers of a child context are also registered with its
+  parent context. As a result, the `close` method must be called to avoid memory
+  leak if the child context is to be recycled.
+
+To react on context events asynchronously, we need to implement the
+`ContextObserver` interface and register it with the context.
```
bajtos: Please add a short description to explain the difference in how are ContextObserver instances invoked when compared to regular events:
- observers are called in the next turn of Promise micro-queue
- when there are multiple async observers registered, do they run in parallel or in series?
- when there are multiple events triggered in the same event loop and there are async observers registered, in which order will they be triggered?

packages/context/src/context.ts
```diff
@@ -34,16 +52,142 @@ export class Context {
...
+
+  /**
+   * Set up an internal listener to notify registered observers asynchronously
+   * upon `bind` and `unbind` events
+   */
+  private setupEventHandlers() {
+    // The following are two async functions. Returned promises are ignored as
+    // they are long-running background tasks.
+    this.startNotificationTask('bind').catch(err => {
+      // Catch error to avoid lint violations
```
bajtos: Please extract the error handler into a shared function to avoid duplication. The source code file `context.ts` is already way too long :(

Also this comment is no longer relevant, we are not catching the error only to avoid lint violations, we are forwarding it via `error` event.

```ts
const reportError = err => {
  debug(err);
  this.emit('error', err);
};

this.startNotificationTask('bind').catch(reportError);
this.startNotificationTask('unbind').catch(reportError);
```

packages/context/test/unit/context-observer.unit.ts
bajtos: I'd like to hear from @hacksparrow and @jannyHou - which version of the test would you find easier to read? One that's using process.nextTick with a callback, or the one I proposed where promises & await are used instead?

bajtos: Feel free to fix this in a follow-up pull request if you prefer.

bajtos:
A general observation: the file `context.ts` is almost 700 lines long. I feel that's too much, I read it as a sign that our Context class is doing too much. It would be great to find a way how to move some parts of the functionality into smaller blocks that can live in a different file, and then use composition to leverage those building blocks in Context.

Let's leave such refactoring out of scope of this pull request though, we should focus on getting the initial implementation finished and merged now.

packages/context/src/context.ts
raymondfeng: `p-event` allows `multiArgs` but it ends up creating `[eventType, binding]` as the return value from the promise. I think the overhead is similar.

packages/context/src/context.ts
raymondfeng: A new notification event is only fired when there are observers. I'll defer this micro-optimization.

packages/context/src/context.ts
raymondfeng: Good destructing.

packages/context/test/unit/context-observer.unit.ts
bajtos: On the second thought, let's keep the current test as-is. I'd like to investigate this in more depth to better understand how this test work, I'll do that after the PR is landed.

docs/site/Context.md
raymondfeng: Added

packages/context/src/context.ts
raymondfeng: Fixed.

docs/site/Context.md
raymondfeng: Fixed

packages/context/test/unit/context-observer.unit.ts
raymondfeng: Done

raymondfeng:
@bajtos PTAL

docs/site/Context.md
```diff
@@ -372,7 +388,31 @@ app
...
+It's recommended that `ContextEventObserver` implementations should not throw
+errors in their code. Errors thrown by context event observers are reported as
+follows over the context chain.
+
+1. Check if the current context object has `error` listeners, if yes, emit an
+   `error` event on the context and we're done. if not, try its parent context
+   by repeating step 1.
+
+2. If no context object of the chain has `error` listeners, emit an `error`
+   event on the current context. As a result, the process exits abnormally. See
```
bajtos: Personally, I would choose a slightly different behavior: _If no context object of the chain has `error` listeners, emit an `error` event on the root context  (typically the Application class)._

I.e. let the error bubble up the context chain util we either find a context with error handlers or reach the top.

Just an idea to consider.

packages/context/src/__tests__/unit/context-observer.unit.ts
```diff
@@ -270,6 +268,30 @@ describe('Context', () => {
...
+        if (binding.key === 'bar') {
+          return Promise.reject(err);
+        }
+      });
+      server.bind('bar').to('bar-value');
+      const obj = await pEvent(server, 'error');
+      expect(obj).to.equal(err);
+    });
+
+    it('reports error on parent context if an observer fails', async () => {
```
bajtos: I see two test cases with different observed behavior here, but the test names don't explain why is the behavior different.
- it reports error on current context if an observer fails
- reports error on parent context if an observer fails

I am proposing to rename this test case to something like:

```ts
it('reports error on parent context when there are no error listeners');
```

packages/context/src/context.ts
```diff
@@ -116,18 +117,32 @@ export class Context extends EventEmitter {
...
+      this._debug(...args);
+      // Bubbling up the error event over the context chain
+      // until we find an error listener
+      let ctx: Context | undefined = this;
+      while (ctx) {
+        if (ctx.listenerCount('error')) {
+          this._debug('Emitting error to context %s', ctx.name);
+          ctx.emit('error', ...args);
+          break;
```
bajtos: Have you considered using early `return` here? Then you don't have have to check `if (ctx === undefined)` on L135 below.

packages/context/src/context.ts
```diff
@@ -116,18 +117,32 @@ export class Context extends EventEmitter {
    * upon `bind` and `unbind` events
    */
   private setupEventHandlers() {
+    // tslint:disable-next-line:no-any
+    const notificationErrorHandler = (...args: any[]) => {
+      // Catch error to avoid lint violations
```
bajtos: I am confused about this comment. Which linting rule would be violated without the debug statement?

docs/site/Context.md
raymondfeng: I leave it to the current context so that we know which one causes the crash. The end result should be the same.

packages/context/src/context.ts
raymondfeng: That was a left-over from refactoring. Removed.

packages/context/src/context.ts
```diff
@@ -34,10 +54,57 @@ export class Context {
   protected _parent?: Context;
 
   /**
-   * Create a new context
+   * Event listeners for parent context keyed by event names. It keeps track
+   * of listeners from this context against its parent so that we can remove
+   * these listeners when this context is closed.
+   */
+  protected _parentEventListeners:
+    | Map<
```
jannyHou: an extra `|`?

packages/context/src/context.ts
```diff
@@ -46,6 +113,194 @@ export class Context {
...
+    // Listen on the parent context events
+    this._parent.on(event, parentEventListener);
+  }
+
+  /**
+   * Handle errors caught during the notification of observers
+   * @param err Error
+   */
+  private handleNotificationError(err: unknown) {
+    this._debug(err);
```
bajtos: I think this debug statement will be confusing to our users - it prints the error only but does not give any context where is the error coming from (a context-event listener).

I am proposing to remove this `_debug` statement and include the error in the debug statements on lines 215 and 220 below.

packages/context/src/context.ts
raymondfeng: That's added by `prettier`.

packages/context/src/__tests__/unit/context-observer.unit.ts
```diff
@@ -0,0 +1,365 @@
...
+      ctx.subscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.true();
+      ctx.unsubscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.false();
+    });
+
+    it('allows subscription.unsubscribe()', () => {
+      const subscription = ctx.subscribe(nonMatchingObserver);
+      expect(ctx.isSubscribed(nonMatchingObserver)).to.true();
+      subscription.unsubscribe();
```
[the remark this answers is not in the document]
raymondfeng: For `subscription.subscribe()`, we don't need to remember the observer. It follows https://github.com/tc39/proposal-observable

Issue #2285, opened 2019-01-25
rest-explorer doesn't work behind path-mapping proxy


mgabeler-lee-6rs:
## Description / Steps to reproduce / Feature proposal
The `rest-explorer` component doesn't work if the app is behind a proxy that does path mapping -- e.g. external requests made to `https://example.com/servicename/api-endpoint`, which is mapped to `http://localhost:3000/api-endpoint` in a docker container or the like.  There is no way to make the explorer component or the external explorer redirect work in this environment that I can find.

## Current Behavior
In the normal setup, the explorer tells the client to load `/openapi.json`, which doesn't work, because it needs to load `https://example.com/servicename/openapi.json`.  The external redirect has the same problem.

## Expected Behavior
1. It should be possible to configure things such that the `openapi.json` correctly references paths when behind a proxy that inserts a path prefix
2. It should be possible to configure the `rest-explorer` component to work in this context too

## Other Notes
This is loosely related to https://github.com/strongloop/loopback-next/issues/1702 but that only dealt with host/port issues, not path issues.

This is more closely related to https://github.com/strongloop/loopback-next/issues/914

Internally we do not want to use the externally hosted API explorer, but the issues with the base path in the API spec serving mean that I couldn't make that work either even if I wanted to.

### Workaround & possible fix?

I was able to work around this, but it is not pretty:
1. Set `rest.openApiSpec.servers` to reflect the proxy path mapping -- this seems to be roughly equivalent to the LB2 config item, and seems reasonable
2. Override the `ExplorerController` class with a local sub-class that patches its private `openApiSpecUrl` member to be aware of the `rest.openApiSpec.servers` value, and setup the app binding to use this class instead of the normal one.  Similarly also had to patch up the `indexRedirect` method to be aware of this path prefix.

The second item is rather icky given the class I'm extending is not exported, and the member I'm modifying is private.

However, making the normal `ExplorerController` aware of the `openApiSpec.servers` list seems like it _might_ be a reasonable fix for this?  However doing that "for real" has some issues I'm not sure how to address:
1. What if there are multiple servers defined?  Which one should it use?  Using the first one is a simple answer, but it's not at all clear to me that it is the correct answer, esp. since I'm not clear on any precedence rules that might be implied by order in that list.
[...]

PR #2293, opened 2019-01-25
fix: make rest-explorer aware of openapi server root path

mgabeler-lee-6rs:
The `rest-explorer` component needs to be aware of any "external" base paths that are in use.  This PR is at least a start to that by making it prefix any paths it sends to clients with the first path from the `openApiSpec.servers` list (if any).

Fixes: #2285

## Checklist

  - N/A?
  - N/A
  - N/A
  - N/A

Having some problems with `test` and `lint`, mainly related to `npm install` not having installed `prettier` for some reason.  I have no reason to believe that this is caused by my changes, and I added tests for my changes, which pass, but I've been unable to run the linter or get the full test suite to pass on my system yet.

### DIFF
--- a/packages/rest-explorer/src/rest-explorer.controller.ts
+++ b/packages/rest-explorer/src/rest-explorer.controller.ts
@@ -21,6 +21,7 @@ const template = fs.readFileSync(indexHtml, 'utf-8');
 const templateFn = ejs.compile(template);
 
 export class ExplorerController {
+  private serverRootPath: string;
   private openApiSpecUrl: string;
 
   constructor(
@@ -29,11 +30,12 @@ export class ExplorerController {
     @inject(RestBindings.Http.REQUEST) private request: Request,
     @inject(RestBindings.Http.RESPONSE) private response: Response,
   ) {
+    this.serverRootPath = this.getServerRootPath(restConfig);
     this.openApiSpecUrl = this.getOpenApiSpecUrl(restConfig);
   }
 
   indexRedirect() {
-    this.response.redirect(301, this.request.url + '/');
+    this.response.redirect(301, this.serverRootPath + this.request.url + '/');
   }
 
   index() {
@@ -48,13 +50,24 @@ export class ExplorerController {
       .send(homePage);
   }
 
+  private getServerRootPath(restConfig: RestServerConfig): string {
+    const openApiConfig = restConfig.openApiSpec || {};
+    const servers = openApiConfig.servers || [];
+    let url = servers[0] ? servers[0].url : '';
+    // trim trailing '/' so we can safely append rooted paths to it
+    if (url.endsWith('/')) {
+      url = url.substring(0, url.length - 1);
+    }
+    return url;
+  }
+
   private getOpenApiSpecUrl(restConfig: RestServerConfig): string {
     const openApiConfig = restConfig.openApiSpec || {};
     const endpointMapping = openApiConfig.endpointMapping || {};
     const endpoint = Object.keys(endpointMapping).find(k =>
       isOpenApiV3Json(endpointMapping[k]),
     );
-    return endpoint || '/openapi.json';
+    return this.serverRootPath + (endpoint || '/openapi.json');
   }
 }
 
--- a/packages/rest-explorer/test/acceptance/rest-explorer.acceptance.ts
+++ b/packages/rest-explorer/test/acceptance/rest-explorer.acceptance.ts
@@ -71,13 +71,6 @@ describe('API Explorer (acceptance)', () => {
       const body = response.body;
       expect(body).to.match(/^\s*url: '\/apispec',\s*$/m);
     });
-
-    async function givenAppWithCustomRestConfig(config: RestServerConfig) {
-      app = givenRestApplication(config);
-      app.component(RestExplorerComponent);
-      await app.start();
-      request = createRestAppClient(app);
-    }
   });
 
   context('with custom RestExplorerConfig', () => {
@@ -109,8 +102,53 @@ describe('API Explorer (acceptance)', () => {
     }
   });
 
+  context('with custom server root', async () => {
+    beforeEach(async () => {
+      await givenAppWithCustomRestConfig({
+        openApiSpec: {
+          servers: [{url: '/external/proxy/root'}],
+        },
+      });
+    });
+
+    it('prepends server root path to OpenAPI path', async () => {
+      const response = await request.get('/explorer/').expect(200);
+      const body = response.body;
+      expect(body).to.match(
+        /^\s*url: '\/external\/proxy\/root\/openapi.json',\s*$/m,
+      );
+    });
+
+    it('prepends server root path to /explorer redirect', async () => {
+      await request
+        .get('/explorer')
+        .expect(301)
+        .expect('location', '/external/proxy/root/explorer/');
+    });
+
+    it('handles server root with trailing slash', async () => {
+      await givenAppWithCustomRestConfig({
+        openApiSpec: {
+          servers: [{url: '/external/proxy/root/'}],
+        },
+      });
+      const response = await request.get('/explorer/').expect(200);
+      const body = response.body;
+      expect(body).to.match(
+        /^\s*url: '\/external\/proxy\/root\/openapi.json',\s*$/m,
+      );
+    });
+  });
+
   function givenRestApplication(config?: RestServerConfig) {
     const rest = Object.assign({}, givenHttpServerConfig(), config);
     return new RestApplication({rest});
+  }
+
+  async function givenAppWithCustomRestConfig(config: RestServerConfig) {
+    app = givenRestApplication(config);
+    app.component(RestExplorerComponent);
+    await app.start();
+    request = createRestAppClient(app);
   }
 });


packages/rest-explorer/src/rest-explorer.controller.ts
```diff
@@ -29,11 +30,12 @@ export class ExplorerController {
     @inject(RestBindings.Http.REQUEST) private request: Request,
     @inject(RestBindings.Http.RESPONSE) private response: Response,
   ) {
+    this.serverRootPath = this.getServerRootPath(restConfig);
     this.openApiSpecUrl = this.getOpenApiSpecUrl(restConfig);
   }
 
   indexRedirect() {
-    this.response.redirect(301, this.request.url + '/');
+    this.response.redirect(301, this.serverRootPath + this.request.url + '/');
```
raymondfeng: I'm not sure if inferring from `openapi.servers[0].url` is a feasible solution. Please note that LoopBack app is not aware how its endpoints are proxies/exposed. The same endpoint can be possibly exposed as many different urls, such as:

- https://a.example.com/api/customer
- https://b.example.com/api/user
- http://internal.example.com:8080/customer

When it comes to redirect via `Location` header, my understanding is that proxy/load balancer is responsible for rewriting `Location` to match its clients.

mgabeler-lee-6rs: The redirect could be made to depend on the proxy (though that requires config on the proxy).  But the same question would then apply to the `/openapi.json` reference too.

I was concerned about this in my original bug report, but @bajtos had a different perspective.

A brainstorm: I think LB2 dealt with this by having the explorer be the one that serves the `openapi.json`, at least when "talking to itself".  This could be replicated within `rest-explorer` with a little work I think, though it would require the `RestServer` to expose `_serveOpenApiSpec` as a public method.  But I think then the explorer could be made to use exclusively relative URLs, avoiding the path rewriting issue in a more general way?

raymondfeng: Some references:
- http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect

mgabeler-lee-6rs: I was hoping that using a relative `Location` url would mean the proxy doesn't _need_ to rewrite it :)

mgabeler-lee-6rs:
@raymondfeng @bajtos -- What do you think about having the explorer serve its own copy of the `openapi.json` so that it can be loaded from a URL relative to the explorer itself, and thus hopefully avoid any issues of computing base paths or the like?

raymondfeng:
> What do you think about having the explorer serve its own copy of the openapi.json so that it can be loaded from a URL relative to the explorer itself, and thus hopefully avoid any issues of computing base paths or the like?

Can you elaborate?

mgabeler-lee-6rs:
Currently the explorer tries to figure out the URL it should send to the client for it to fetch the `openapi.json` file that the `RestServer` is already serving.  But trying to figure out that URL, as noted above, can be complex, if even impossible, in more complex proxy & custom config cases.

So, instead of trying to find the URL it's being served on for other purposes, what if it adds its own URL on which it serves the OAS material.  I.e. if you have the explorer mounted at `.../explorer/`, then it could serve the OAS JSON to itself at `.../explorer/openapi.json`, regardless of what other endpoints you've defined for the OAS content to be served.  That way it can safely know that, whatever URL is serving the explorer, `./openapi.json` will work as a relative URL to get the API spec, and it doesn't have to figure out what proxy mapping or other configuration is active to find the client-view of the URL for `/openapi.json` -- i.e. it doesn't have to have any idea what that `...` is.  It also means the explorer can still work if the user has customized away the default `RestServer` to e.g. only serve the OAS file in YAML.

This is (roughly) how it seems to have worked in LB2 -- the OAS file was at `.../explorer/swagger.json` IIRC, so the explorer didn't need to worry about absolute paths to load that file.

raymondfeng:
I think it's worth exploring. Maybe we should expose an API from [`@loopback/rest`](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L241).

mgabeler-lee-6rs:
> I think it's worth exploring.

I see what you did there :wink: 

> Maybe we should expose an API from [`@loopback/rest`](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L241).

Indeed ... I was thinking of exposing a version of [`serveOpenApiSpec`](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L367), with suitable adjustments for being a public method (if any).  Then I could add a `openApiSpec` method to `ExplorerControler`, wire it up as `explorerPath + '/openapi.json'` with `registerControllerRoute` in `RestExplorerComponent`, and update `index.html.ejs` accordingly.

If this sounds sensible, I can update this PR to go down that path.

packages/rest-explorer/src/rest-explorer.controller.ts
bajtos: I believe that `this.request.url` contains only the path portion of the requested URL _as seen by the LB4 server_, after it has been rewritten by the reverse proxy. The current implementation is correct, it's redirecting from `/explorer` to `/explorer/index.html`.

I am not able to fully understand implications of the change proposed here. More importantly, I don't understand why this particular change is needed.

Can we revert it please?

bajtos:
+1 to modify our REST API explorer to use a relative URL for the OpenAPI spec if that's going to fix the problem.

I am not convinced that we need a new explorer-specific endpoint for that though. A typical LoopBack 4 application comes with the following endpoints:

 - `{basePath}/openapi.json` serves the spec
 - `{basePath}/explorer` servers Explorer UI

IMO, it should be enough to figure out what's the relative path from `/explorer` (which can be customized via REST API Explorer config option `path`) to `/openapi.json` (which can be customized via `restConfig.openApiSpec.endpointMapping`.

`basePath` is `/` by default, but can be changed by `app.basePath` API or rewritten by a reverse proxy as described by @mgabeler-lee-6rs.

Can we assume that the base path for openapi and explorer endpoints will be always the same? I believe the answer is true when LB4 is called directly. I am not sure about reverse proxies. How likely it is that somebody will configure the reverse proxy to expose `/openapi.json` at let's say `https://example.com/servicename/api-endpoint/openapi.json` and `/explorer` at `https://example.com/servicename/api-explorer`?

I think our implementation should:
 - Work out of the box for the default setup, where both `/openapi.json` and `/explorer` endpoints are exposed at the same base url. This can be fixed either by computing a relative path, or it may be enough to fix the implementation of `getOpenApiSpecUrl` to take `basePath` config into account. (/cc @nabdelgadir, I think you run into the problem of `basePath` support in REST API Explorer too)
 - Provide a new REST API Explorer configuration option allowing advanced users to specify a custom URL where to find the spec. For example:

    ```
    {
      path: '/explorer',
      specUrl: 'https://example.com/servicename/api-explorer',
    }
    ```

That way we can keep our implementation simple and easy to maintain and still support advanced use cases.

bajtos:
@mgabeler-lee-6rs I feel that I may not fully understand what's happening under the hood when a reverse proxy is configured in front of a LB4 app and REST API Explorer is loaded. Could you please create a setup we could use to reproduce the problem you are encountering on our machines? I'd like to investigate this more in depth myself to build a better understanding on how individual parts interact together. A reverse proxy configuration for our [todo example](https://github.com/strongloop/loopback-next/tree/master/examples/todo) would be ideal.

raymondfeng:
@mgabeler-lee-6rs It would be great to create such scenario based on https://loopback.io/doc/en/lb4/deploying-with-pm2-and-nginx.html to illustrate how the `Location` header is impacted by `nginx` reverse proxy.

raymondfeng:
This seems to be simple way to reproduce the problem too - https://github.com/strongloop/loopback-next/pull/2306#discussion_r252321716

bajtos:
> #### Conflicting files
> `packages/rest-explorer/test/acceptance/rest-explorer.acceptance.ts`

In https://github.com/strongloop/loopback-next/pull/2330, we have moved test files from `test` to `src/__tests__`. Please run follow the steps in [How to rebase your branch](https://github.com/strongloop/loopback.io/blob/26bc4788d818662a4e8d4e4c19803ab38ed75655/pages/en/contrib/code-contrib.md#how-to-rebase-your-branch) to resolve the merge conflict. Please do not run `git merge`, it's important to use `git rebase master`.

Sorry for the inconvenience.

bajtos:
@mgabeler-lee-6rs ping 👋 what's the status of this pull request? Are you still keen to get it finished?

mgabeler-lee-6rs:
I am still keen to finish it up, yes.  I've just been diverted by some higher priorities temporarily.

raymondfeng:
I created https://github.com/strongloop/loopback-next/pull/2387 to address the issue when rest-server is mounted to express using a path such as `expressApp.use('/api', restServer.requestHandler);`.

ludohenin:
@raymondfeng when do you see that one landing ?

bajtos:
Is this pull request still relevant, considering the work done in https://github.com/strongloop/loopback-next/pull/2856 and previous pull requests?

mgabeler-lee-6rs:
I'd need to do some testing.  I think `basePath` handling doesn't cover my use case, where the reverse proxy is stripping off the path prefix.  In that case, setting up a `basePath` actually _breaks everything_, at at least that's what I recall from the last time I tried it -- we need the controllers to all be blissfully unaware of the path prefix the reverse proxy is externally applying.  Trying to put a `basePath` into things caused it to want to have _two_ copies of that in the path, which just made things more broken.

@dkrantsberg's reverse-proxy setup described in the comments on #2554 is not dissimilar to mine, though simpler.  I'm curious if he has a working setup for comparison.

The workaround I'm still monkey-patching into my apps is to sub-class the explorer and give it an extra config setting for the "external" base path, which is built based on known patterns for our reverse proxy when running in production.  This is ... not pretty to say the least, but it gets the job done for me.

I think the simplest way to be able to handle this in a portable fashion is to make the rest explorer exclusively use relative URLs as described earlier -- have the explorer serve its own copy of the openapi spec using the format it needs, at a path relative to the explorer.  Then we never need to worry about absolute paths in any of the requests, and most of the path logic _should_ go away?

bajtos:
@mgabeler-lee-6rs thank you for the response.

> I think the simplest way to be able to handle this in a portable fashion is to make the rest explorer exclusively use relative URLs as described earlier -- have the explorer serve its own copy of the openapi spec using the format it needs, at a path relative to the explorer. Then we never need to worry about absolute paths in any of the requests, and most of the path logic should go away?

That sounds like a reasonable solution to me. Would you like to open a new pull request for this proposal?

Another alternative that comes to my mind: modify Explorer to use a relative path to access OpenAPI spec. I.e. when serving `/mountPath/explorer/`, then use `../openapi.json` instead of `/mountPath/openapi.json`.

mgabeler-lee-6rs:
> Would you like to open a new pull request for this proposal?

Yes, I'll give that a shot :slightly_smiling_face: 

> Another alternative that comes to my mind: modify Explorer to use a relative path to access OpenAPI spec. I.e. when serving `/mountPath/explorer/`, then use `../openapi.json` instead of `/mountPath/openapi.json`.

I thought about that, but then the problem space of correctly computing the relative path to the OAS for all the possible ways users might modify it starts to feel ... complicated.  Much more so than having our own copy that is always at a fixed path relative to the explorer.

mgabeler-lee-6rs:
I've started a draft of the relative path hosting in #3133.  That PR is definitely in a WIP state right now, but there is enough there to take a look at the idea at least.  I'll be fleshing it out with tests and whatnot of course.

Closing this PR in favor of that new one.

Issue #1981, opened 2018-11-05
Cannot start the application. Error: Unsupported type: geopoint


JorisPalings:
# Description/Steps to reproduce
I'm attempting to start my project by using `npm start` and get the following output:
```
> project-api@1.0.0 prestart /Users/joris/Documents/Personal/Projects/project/project-api
> npm run build


> project-api@1.0.0 build /Users/joris/Documents/Personal/Projects/project/project-api
> lb-tsc es2017 --outDir dist


> project-api@1.0.0 start /Users/joris/Documents/Personal/Projects/project/project-api
> node .

Cannot start the application. Error: Unsupported type: geopoint
    at stringTypeToWrapper (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/repository-json-schema/dist/src/build-schema.js:68:19)
    at metaToJsonProperty (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/repository-json-schema/dist/src/build-schema.js:101:25)
    at modelToJsonSchema (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/repository-json-schema/dist/src/build-schema.js:151:32)
    at Object.getJsonSchema (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/repository-json-schema/dist/src/build-schema.js:22:27)
    at generateOpenAPISchema (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/openapi-v3/dist/src/controller-spec.js:181:49)
    at resolveControllerSpec (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/openapi-v3/dist/src/controller-spec.js:128:17)
    at Object.getControllerSpec (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/openapi-v3/dist/src/controller-spec.js:205:16)
    at RestServer._setupHandlerIfNeeded (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/rest/dist/src/rest.server.js:195:42)
    at RestServer.start (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/rest/dist/src/rest.server.js:498:14)
    at _forEachServer.s (/Users/joris/Documents/Personal/Projects/project/project-api/node_modules/@loopback/core/dist/src/application.js:123:42)
[...]

PR #2299, opened 2019-01-28
fix(repository): add near operator for where filters

### LINKED ISSUE #1419: Using near filter in where query cause toEntity(model) bug
## Description

I have Address model:

`Address extends Entity {`

    @property({
        type: 'number',
        id: true,
        generated: true,
        mysql: {
            "columnName": "m_id"
        }
    })
    mId?: number;

    @property({
        type: 'number',
        mysql: {
            "columnName": "m_belong_to_id"
        }
    })
    mBelongToId?: number;

    @property({
        type: 'GeoPoint',
        required: true,
        mysql: {
            "columnName": "m_geo_point"
        }
    })
    mGeoPoint: GeoPoint;
`}`

## Current Behavior
I try this query: 
`let nearByAddressList = await this.addressRepo.find({`

                where: {
                    and: [
                        {
                            mGeoPoint: {
                                near: location,
                                maxDistance: this.maxDistance,
                                unit: this.unit
                            }
                        }
                    ]
                }
            });
and the console show me the bug: _model.toObject() is not a function on file node_modules/@loopback/repository/dist8/src/repositories/legacy-juggler-bridge.js:160_
If I remove the _where_ filter, the code run normally. 

I checked the code and couldn't find _near_ properties on Where filter class. 
Can I use _near_ in _where_ filter on Lp4 now?
## Expected Behavior
Query run normally with _near_ in query. 

b-admike:
Fixes #1419. Currently, I'm not able to reproduce the error message seen in the issue itself. https://loopback.io/doc/en/lb2/Where-filter.html#near shows that the `near` operator includes additional properties like `maxDistance` and `unit`, but I see them being specified concurrently or at the same level as `near` and not nested inside, so I've left them out of its type definition.
## Checklist

### DIFF
--- a/packages/repository/src/query.ts
+++ b/packages/repository/src/query.ts
@@ -39,7 +39,8 @@ export type Operators =
   | 'nlike' // NOT LIKE
   | 'ilike' // ILIKE'
   | 'nilike' // NOT ILIKE
-  | 'regexp'; // REGEXP'
+  | 'regexp' // REGEXP'
+  | 'near'; // NEAR
 
 /**
  * Matching predicate comparison
@@ -60,6 +61,7 @@ export type PredicateComparison<PT> = {
   ilike?: PT;
   nilike?: PT;
   regexp?: string | RegExp;
+  near?: object;
   // [extendedOperation: string]: any;
 };
 
--- a/packages/repository/test/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/test/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -62,6 +62,33 @@ describe('DefaultCrudRepository', () => {
     }
   }
 
+  interface GeoPoint {
+    /**
+     * latitude
+     */
+    lat: number;
+
+    /**
+     * longitude
+     */
+    lng: number;
+  }
+  class Address extends Entity {
+    static definition = new ModelDefinition({
+      name: 'Address',
+      properties: {
+        mGeoPoint: {type: 'GeoPoint', required: true},
+        mBelongToId: 'number',
+        mId: {name: 'id', type: 'number', id: true, generated: true},
+      },
+    });
+    mId?: number;
+
+    mBelongToId?: number;
+
+    mGeoPoint: GeoPoint;
+  }
+
   beforeEach(() => {
     ds = new juggler.DataSource({
       name: 'db',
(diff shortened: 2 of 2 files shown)


packages/repository/src/query.ts
```diff
@@ -60,6 +61,7 @@ export type PredicateComparison<PT> = {
   ilike?: PT;
   nilike?: PT;
   regexp?: string | RegExp;
+  near?: object;
```
b-admike: Need a more specific type here to show it is a `GeoPoint` object. We have a definition in our `geo-services` example. Maybe we can extract that out to `repository` layer types. Thoughts?

b-admike:
@hosiduy @susannaHayrapetyan This PR attempts to fix your issue in #1419. If you have suggestions on how to reproduce the error in the test case I created, please let me know. Right now, it is behaving correctly.

b-admike:
I was able to reproduce the issue once I switched the datasource to use MySQL with the test case. It seems that we return a model class (legacy) instance in the case of a `find` operation for a `Note` model without a GeoPoint property whereas we return a plain JSON object of the instance with `find` for the `Address` model for the test I added.

packages/repository/src/query.ts
bajtos: IIRC, the type used by `geo-services` is not exactly the same one as we use in LoopBack and connectors. I made the type closer to the external webservice called by the example app to make the code simpler.

Please see https://github.com/strongloop/loopback-next/issues/1981#issuecomment-436966284 and the related discussion for different options.

For now, I am proposing to use `any` or an inline type.

An example of an inline type to illustrate what I mean is below. The type definition is probably not correct, please don't take it as-is.

```ts
near?: {lat: number, lng: number} | [number, number];
```

packages/repository/test/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -62,6 +62,33 @@ describe('DefaultCrudRepository', () => {
...
+     */
+    lng: number;
+  }
+  class Address extends Entity {
+    static definition = new ModelDefinition({
+      name: 'Address',
+      properties: {
+        mGeoPoint: {type: 'GeoPoint', required: true},
+        mBelongToId: 'number',
+        mId: {name: 'id', type: 'number', id: true, generated: true},
```
bajtos: I am curious, where is the `m` prefix coming from? I believe we don't use such prefixes elsewhere in LB4 codebase. Can you rename the properties please?

See also https://en.wikipedia.org/wiki/Hungarian_notation#Disadvantages

b-admike: It was coming from the model definition that our user @hosiduy had when he encountered the issue. I will rename them 👍.

dhmlau:
@b-admike, since #1419 has been resolved by another PR, are we going to abandon this PR?

Issue #1982, opened 2018-11-06
Tutorial/Blog: how to mount LB app as Express/Koa middleware


dhmlau:
## Description / Steps to reproduce / Feature proposal
Per our discussion with @raymondfeng @bajtos and others, users can mix and match using LoopBack and Express, taking the best of both framework. 
LB can have a facade application that can continue to use express for hosting capability but can compose different segment of the route.  LB app can be mounted to express / koa as middleware.

See also https://github.com/strongloop/loopback-next/issues/1293 Allow app developers to configure custom Express middleware

### Acceptance Criteria
* [ ] Create a sample app built from the tutorial (see TodoList tutorial for reference)

bajtos:
See https://github.com/strongloop/loopback-next/issues/691 and https://github.com/strongloop/loopback-next/issues/1818 for related discussions we have had in the past.

With the current code base and APIs, it's already possible to mount LB4 REST API as an Express middleware.

```ts
const mainApp = express();
const apiApp = new MyLb4Application();

mainApp.use('/api', apiApp.requestHandler);
mainApp.use(express.static('./assets'));

mainApp.listen(3000);
```

drouillard:
This would be helpful. I know target market is API developers, but when getting started it is nice to be able to plug in views without having to launch a separate React App or similar. 

I ended up doing this which I can't imagine is your desired solution :) 

```ts
// @ts-ignore
  const expressApp = app.restServer._expressApp;

  expressApp.set('view engine', 'ejs');
  expressApp.set('views', path.join(__dirname, 'views'));
```

I looked into overriding Rest Server and Rest Application but the learning curve was too much for a days worth of trial / error. The above solution of mounting inside of an Express works well and is cleaner. I wanted to try out loopback-authentication though hence the hack I did.

jannyHou:
@drouillard Thank you for sharing a solution from a different perspective! I think your approach and the one in the original comment serves different requirements:

### launching an express app 

Developers can take this approach when they only want to mount loopback app's APIs as a whole without interfering with its implementation details.

### build routes on top of express app in rest server

As you explained, it's good for adding plugin like views on top of an existing loopback app.

We could investigate and document both.

drouillard:
[...]

PR #2306, opened 2019-01-29
feat: add express example

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/1982.

## Checklist

### DIFF
--- a/docs/site/express-with-lb4-rest-tutorial.md
+++ b/docs/site/express-with-lb4-rest-tutorial.md
@@ -1,0 +1,308 @@
+---
+lang: en
+title: 'Creating an Express Application with LoopBack REST API'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/express-with-lb4-rest-tutorial.html
+summary: A simple Express application with LoopBack 4 REST API
+---
+
+## Overview
+
+[Express](https://expressjs.com) is an unopinionated Node.js framework. LoopBack
+REST API can be mounted to an Express application and be used as middleware.
+This way the user can mix and match features from both frameworks to suit their
+needs.
+
+This tutorial assumes familiarity with scaffolding a LoopBack 4 application,
+[`Models`](Model.md), [`DataSources`](DataSources.md),
+[`Repositories`](Repositories.md), and [`Controllers`](Controllers.md). To see
+how they're used in a LoopBack application, please see the
+[`Todo` tutorial](todo-tutorial.md).
+
+## Try it out
+
+If you'd like to see the final results of this tutorial as an example
+application, follow these steps:
+
+1.  Run the `lb4 example` command to select and clone the express-composition
+    repository:
+
+    ```sh
+    $ lb4 example
+    ? What example would you like to clone? (Use arrow keys)
+    todo: Tutorial example on how to build an application with LoopBack 4.
+    todo-list: Continuation of the todo example using relations in LoopBack 4.
+    hello-world: A simple hello-world Application using LoopBack 4.
+    log-extension: An example extension project for LoopBack 4.
+    rpc-server: A basic RPC server using a made-up protocol.
+    > express-composition: A simple Express application that uses LoopBack 4 REST API.
+    ```
+
+2.  Switch to the directory.
+
+    ```sh
+    cd loopback4-example-express-composition
+    ```
+
+3.  Finally, start the application!
+
+    ```sh
+    $ npm start
+
+    Server is running at http://127.0.0.1:3000
+    ```
+
+### Create your LoopBack Application
+
+#### Scaffold your Application
+
+Run `lb4 app note` to scaffold your application and fill out the following
+prompts as follows:
+
+```sh
+$ lb4 app
+? Project description: An application for recording notes.
+? Project root directory: (note)
+? Application class name: (NoteApplication)
+❯◉ Enable tslint: add a linter with pre-configured lint rules
+ ◉ Enable prettier: add new npm scripts to facilitate consistent code formatting
+ ◉ Enable mocha: install mocha to assist with running tests
+ ◉ Enable loopbackBuild: use @loopback/build helpers (e.g. lb-tslint)
+ ◉ Enable vscode: add VSCode config files
+ ◉ Enable repositories: include repository imports and RepositoryMixin
+ ◉ Enable services: include service-proxy imports and ServiceMixin
+ # npm will install dependencies now
+ Application note was created in note.
+```
+
+#### Add Note Model
+
+Inside the project folder, run `lb4 model` to create the `Note` model with
+`Entity` model base class. Add an `id` property with type `number`, a required
+`title` property with type `string`, and a `content` property of type `string`.
+
+#### Add a DataSource
+
+Now, let's create a simple in-memory datasource by running the
+`lb4 datasource ds` command and the following full path to file:
+`./data/ds.json`.
+
+Similar to the `Todo` example, let's create the `ds.json` by creating a data
+folder at the application's root.
+
+```sh
+$ mkdir data
+$ touch data/ds.json
+```
+
+Then copy and paste the following into the `ds.json` file:
+
+```json
+{
+  "ids": {
+    "Note": 3
+  },
+  "models": {
+    "Note": {
+      "1": "{\"title\":\"Things I need to buy\",\"content\":\"milk, cereal, and waffles\",\"id\":1}",
+      "2": "{\"title\":\"Great frameworks\",\"content\":\"LoopBack is a great framework\",\"id\":2}"
+    }
+  }
+}
+```
+
+#### Add Note Repository
+
+To create the repository, run the `lb4 repository` command and choose the
+`DsDataSource` as the datasource and `Note` model as the model.
+
+#### Add Note Controller
+
+To complete the `Note` application, create a controller using the
+`lb4 controller note` command, with the `REST Controller with CRUD functions`
+type, `Note` model, and `NoteRepository` repository. The `id`'s type will be
+`number` and base HTTP path name is the default `/notes`.
+
+### Create a Facade Express Application
+
+Let's start by installing dependencies for the `express` module:
+
+```sh
+npm install --save express
+npm install --save-dev @types/express
+```
+
+Create a new file **src/server.ts** to create your Express class:
+
+{% include code-caption.html content="src/server.ts" %}
+
+```ts
+import * as express from 'express';
+
+export class ExpressServer {
+  constructor() {}
+}
+```
+
+Create two properties, the Express application instance and LoopBack application
+instance:
+
+```ts
+import {NoteApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+import * as express from 'express';
+
+export class ExpressServer {
+  private app: express.Application;
+  private lbApp: NoteApplication;
+
+  constructor(options: ApplicationConfig = {}) {
+    this.app = express();
+    this.lbApp = new NoteApplication(options);
+  }
+}
+```
+
+Now, inside the constructor, we're going to add the basepath and expose the
+front-end assets via Express:
+
+```ts
+this.app.use('/api', this.lbApp.requestHandler);
+```
+
+Let's also modify **public/index.html** to update the base path:
+
+{% include code-caption.html content="public/index.html" %}
+
+```html
+<h3>OpenAPI spec: <a href="/api/openapi.json">/openapi.json</a></h3>
+<h3>API Explorer: <a href="/api/explorer">/explorer</a></h3>
+```
+
+Then, we can add some custom Express routes, as follows:
+
+```ts
+import {Request, Response} from 'express';
+import * as path from 'path';
+
+export class ExpressServer {
+  private app: express.Application;
+  private lbApp: NoteApplication;
+
+  constructor(options: ApplicationConfig = {}) {
+    // earlier code
+
+    // Custom Express routes
+    this.app.get('/', function(_req: Request, res: Response) {
+      res.sendFile(path.resolve('public/express.html'));
+    });
+    this.app.get('/hello', function(_req: Request, res: Response) {
+      res.send('Hello world!');
+    });
+  }
+}
+```
+
+And add the
+[public/express.html](https://github.com/strongloop/loopback-next/blob/master/examples/express-composition/public/express.html)
+file to your project.
+
+Let's also install [`p-event`](https://www.npmjs.com/package/p-event) to make
+sure the server is listening:
+
+```sh
+npm install --save p-event
+```
+
+Finally, we can add functions to boot the `Note` application and start the
+Express application:
+
+```ts
+import * as pEvent from 'p-event';
+
+export class ExpressServer {
+  private app: express.Application;
+  private lbApp: NoteApplication;
+
+  constructor(options: ApplicationConfig = {}) {
+    //...
+  }
+
+  async boot() {
+    await this.lbApp.boot();
+  }
+
+  async start() {
+    await this.lbApp.start();
+    const server = this.app.listen(3000);
+    await pEvent(server, 'listening');
+  }
+}
+```
+
+Now that our **src/server.ts** file is ready, then we can modify our
+**src/index.ts** file to start the application:
+
+{% include code-caption.html content="src/index.ts" %}
+
+```ts
+import {ExpressServer} from './server';
+import {ApplicationConfig} from '@loopback/core';
+
+export {ExpressServer};
+
+export async function main(options: ApplicationConfig = {}) {
+  const server = new ExpressServer(options);
+  await server.boot();
+  await server.start();
+  console.log('Server is running at http://127.0.0.1:3000');
+}
+```
+
+Now let's start the application and visit <http://127.0.0.1:3000>:
+
+```sh
+npm start
+
+Server is running at http://127.0.0.1:3000
+```
+
+If we go to the [Explorer](http://127.0.0.1:3000/api/explorer), we can make
+requests for our LoopBack application. Notice how the server is
+<http://127.0.0.1:3000/api>.
+
+To view our custom `/hello` Express route, go to <http://127.0.0.1:3000/hello>
+and you should see 'Hello world!'.
+
+To serve static files in your application, add the following to the end of your
+constructor:
+
+{% include code-caption.html content="src/server.ts" %}
+
+```ts
+export class ExpressServer {
+  private app: express.Application;
+  private lbApp: NoteApplication;
+
+  constructor(options: ApplicationConfig = {}) {
+    // earlier code
+
+    // Serve static files in the public folder
+    this.app.use(express.static('public'));
+  }
+
+  // other functions
+}
+```
+
+Now, you can load any static files in the **public/** folder. For example, add
+the following
+[public/notes.html](https://github.com/strongloop/loopback-next/blob/master/examples/express-composition/public/notes.html)
+file to your project, run `npm start` again, and visit
+<http://127.0.0.1:3000/notes.html>. You can now see a static file that will
+display your Notes in a table format. For more information, please visit
+[Serving static files in Express](https://expressjs.com/en/starter/static-files.html).
+
+Congratulations, you just mounted LoopBack 4 REST API onto a simple Express
+application.
--- a/examples/express-composition/package.json
+++ b/examples/express-composition/package.json
@@ -1,0 +1,72 @@
+{
+  "name": "express-composition",
+  "version": "1.0.0",
+  "description": "LoopBack 4 REST API on Express",
+  "keywords": [
+    "loopback-application",
+    "loopback",
+    "example",
+    "tutorial",
+    "express"
+  ],
+  "main": "index.js",
+  "engines": {
+    "node": ">=8.9"
+  },
+  "scripts": {
+    "build:apidocs": "lb-apidocs",
+    "build": "lb-tsc es2017 --outDir dist",
+    "build:watch": "lb-tsc --watch",
+    "clean": "lb-clean dist",
+    "lint": "npm run prettier:check && npm run tslint",
+    "lint:fix": "npm run tslint:fix && npm run prettier:fix",
+    "prettier:cli": "lb-prettier \"**/*.ts\" \"**/*.js\"",
+    "prettier:check": "npm run prettier:cli -- -l",
+    "prettier:fix": "npm run prettier:cli -- --write",
+    "tslint": "lb-tslint",
+    "tslint:fix": "npm run tslint -- --fix",
+    "pretest": "npm run clean && npm run build",
+    "test": "lb-mocha --allow-console-logs \"dist/src/__tests__\"",
+    "posttest": "npm run lint",
+    "test:dev": "lb-mocha --allow-console-logs dist/src/__tests__/**/*.js && npm run posttest",
+    "migrate": "node ./dist/src/migrate",
+    "prestart": "npm run build",
+    "start": "node .",
+    "prepublishOnly": "npm run test"
+  },
+  "repository": {
+    "type": "git"
+  },
+  "author": "IBM Corp.",
+  "license": "MIT",
+  "files": [
+    "README.md",
+    "index.js",
+    "index.d.ts",
+    "dist/src",
+    "dist/index*",
+    "src"
+  ],
+  "dependencies": {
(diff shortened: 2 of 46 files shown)


nabdelgadir:
To be able to use `express.index.ts`, I'd need to modify `examples/todo/dist/src/index.d.ts` because https://github.com/strongloop/loopback-next/blob/master/examples/todo/index.js#L6 which means `application.main` needs to use `index.ts` as `dist` regenerates so I can't add `express.index.ts`'s `main` function there. 

Another option would be to add the code to `index.ts` and comment out the express parts (but commented code wouldn't look nice). I can also not add the code to the example at all and keep it simply in the tutorial doc, but then I wouldn't be able to add tests.

I also tried creating a new class called `ExpressApplication` which would be something like: 
```ts
import * as exp from 'express';
const express = require('express');

export class ExpressApplication {
  // doesn't recognize express unless it's imported
  app: exp.Application;
  constructor() {
    this.app = new express();
  }
}

```
But then calling `const app = new ExpressApplication();` from `main` and using it would result in the same commented code. 

Any thoughts?

dhmlau:
I wonder if it would be better to separate this from the todo tutorial.  IIUC, the `todo` tutorial is meant to be the very first tutorial someone would follow.  It should contain the minimal set of instruction in order to get things started.  If we keep adding customized/advanced stuff in it, I think it defeats the purpose of the todo tutorial.

nabdelgadir:
@dhmlau yes, I agree. Originally I wanted to make a separate tutorial, but there's only: 
```ts
const app = new express();
const lbApp = new TodoListApplication(options);

// Mount LB4 as our REST API
await lbApp.boot();
lbApp.basePath('/api');

// Expose the front-end assets via Express, not as LB4 route
app.use(lbApp.requestHandler);
app.use(express.static('./public'));

// Add any custom Express routes
app.get('/', function(_req: Request, res: Response) {
  res.sendFile(path.resolve('public/index.html'));
});
app.get('/get', function(_req: Request, res: Response) {
  res.send('Get!');
});

app.listen(3000);
```
to add. That's why I wasn't sure if it warranted its own tutorial, since it would be really simple.

dhmlau:
What we can do is to have a page under `How to` section, doesn't need to be part of the todo app code.  Just a thought.

nabdelgadir:
@dhmlau that's also what I was considering. I think that's the approach that I will go with, thanks!

examples/todo/src/express.index.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/example-todo
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {TodoListApplication} from './application';
+import {ApplicationConfig} from '@loopback/core';
+import {Request, Response} from '@loopback/rest';
+const express = require('express');
```
raymondfeng:
```ts
import * as express from `express`;
```

examples/todo/package.json
```diff
@@ -51,9 +51,12 @@
     "@loopback/http-caching-proxy": "^1.0.5",
     "@loopback/testlab": "^1.0.5",
     "@loopback/tslint-config": "^2.0.0",
+    "@types/express": "^4.16.1",
     "@types/lodash": "^4.14.109",
     "@types/node": "^10.11.2",
     "lodash": "^4.17.10",
+    "p-event": "^2.2.0",
+    "path": "^0.12.7",
```
raymondfeng: There is a built-in module named `path` from node.js itself. Is it intentional?

examples/todo/src/express.index.ts
```diff
@@ -0,0 +1,43 @@
...
+const path = require('path');
+// const pEvent = require('p-event');
+
+export async function main(options: ApplicationConfig = {}) {
+  const app = new express();
```
raymondfeng: It should be `const app = express();`. See http://expressjs.com/en/4x/api.html

examples/todo/src/express.index.ts
```diff
@@ -0,0 +1,43 @@
...
+  const lbApp = new TodoListApplication(options);
+
+  // Mount LB4 as our REST API
+  await lbApp.boot();
+  lbApp.basePath('/api');
+
+  // Expose the front-end assets via Express, not as LB4 route
+  app.use(lbApp.requestHandler);
```
raymondfeng: Should it be the following?
```ts
app.use('/api', lbApp.requestHandler);
```

examples/todo/src/express.index.ts
```diff
@@ -0,0 +1,43 @@
...
+    res.sendFile(path.resolve('public/index.html'));
+  });
+  app.get('/get', function(_req: Request, res: Response) {
+    res.send('Get!');
+  });
+
+  app.listen(3000);
+
+  // This line doesn't work right now
+  // await pEvent(app, 'listening');
```
raymondfeng: It should be:
```ts
const server = app.listen(3000);
await pEvent(server, 'listening');
```

nabdelgadir: Thanks!

examples/todo/src/express.index.ts
nabdelgadir: It works when it's `app.use(lbApp.requestHandler);` because otherwise explorer fails because of `Fetch error: Failed to fetch http://127.0.0.1:3000/openapi.json`, and using `app.use('/api', lbApp.requestHandler);` along `lbApp.basePath('api');` makes the path for the LB app `/api/api` instead of `/api`.

examples/todo/package.json
nabdelgadir: Oh no, you're right.

examples/todo/src/express.index.ts
raymondfeng: See discussion at https://github.com/strongloop/loopback-next/pull/2293

nabdelgadir: Then I think I can work on this PR right now as is, and modify once that PR lands. What do you think?

raymondfeng: Sure.

raymondfeng:
At high level, I think it's better to have a separate package for the express example, such as `examples/express-composition`. The `todo` application can be a dependency of the new module and imported into the express app.

nabdelgadir:
@raymondfeng on second thought, that might be a better approach than just adding it to the `How to` section because, although it would be a simple app, then I can at least add tests.

bajtos:
> on second thought, that might be a better approach than just adding it to the How to section because, although it would be a simple app, then I can at least add tests.

Yes please, it's important to have automated tests in place that will verify LB4-mounted-in-Express setup. @nabdelgadir has already discovered a problem in this integration, there may be more tricky aspects we are not aware of now or that we may accidentally break in the future. Automated test cases will prevent such regressions.

While it's tempting to require existing `todo` application into the new Express project, I think such setup will make the example less useful to our users. I expect most of our users to have both LB4 artifacts and Express code inside the same package (project), thus our example should show the same.

I think it will be best to create a new example project (package) as suggested by Raymond, and use a simpler LB4 setup than we have in the Todo app, e.g. configure a dummy Note model as proposed by Janny.

examples/express-composition/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "@loopback/boot": "^1.0.11",
+    "@loopback/context": "^1.5.0",
+    "@loopback/core": "^1.1.5",
+    "@loopback/openapi-v3": "^1.2.0",
+    "@loopback/repository": "^1.1.4",
+    "@loopback/rest": "^1.5.4",
+    "@loopback/rest-explorer": "^1.1.7",
+    "@loopback/service-proxy": "^1.0.7",
+    "@types/p-event": "^1.3.0",
+    "memory": "0.0.3"
```
raymondfeng: This is a wrong dependency.

raymondfeng:
@nabdelgadir We should not duplicate the `Todo` code here. This example should declare a dependency of `@loopback/example-todo` and import the app to be composed with the express part.

nabdelgadir:
@raymondfeng although the `todo` and `note` examples are similar, I assume the user won't import their own LB app as a dependency. From @bajtos: 

> While it's tempting to require existing `todo` application into the new Express project, I think such setup will make the example less useful to our users. I expect most of our users to have both LB4 artifacts and Express code inside the same package (project), thus our example should show the same.

raymondfeng:
> While it's tempting to require existing `todo` application into the new Express project, I think such setup will make the example less useful to our users. I expect most of our users to have both LB4 artifacts and Express code inside the same package (project), thus our example should show the same.

I see two flavors:

1. The composite express app has all local modules within the same project structure. Ideally, such layout is supported by `lerna` as a monorepo.

2. The composite express app has dependencies to LB4 apps as modules from npm.

Personally, I think 2 is a good starting point so that we can decompose the Express (web) and LB4 (api).

nabdelgadir:
@raymondfeng I'm okay with either approach personally, but @bajtos what do you think?

bajtos:
I am afraid I don't fully understand the difference between the two flavors described by @raymondfeng in https://github.com/strongloop/loopback-next/pull/2306#issuecomment-461829656.

As a LB4 user, I would like to see an example that will teach me how can I build an application that uses top-level Express for certain parts and LB4 for other parts (most notably the  REST API layer) and show me a project layout I can follow in my own project.

I am concerned that an example based on package composition would be too complex for an average user, both to understand the concept while learning and to apply it to their own project.

> The composite express app has all local modules within the same project structure. Ideally, such layout is supported by `lerna` as a monorepo.

To me, this is a good setup for advanced users. I don't think it's possible to setup a lerna monorepo (example project) within another monorepo (loopback-next). To provide a meaningful example, we should implement this variant as a standalone GitHub repo.

Unfortunately, that makes it difficult to test Express+LB4 integration within loopback-next.

> The composite express app has dependencies to LB4 apps as modules from npm.

I find this overly complicated. In real world, it requires the users to setup private npm packages (e.g. a paid account on npmjs.org or their own private registry) and orchestrate a deployment pipeline that will deploy a new version of the top-level Express app when a new version of a dependent LB4 app is published to the registry.

I am proposing the following steps forward:
- As part of this user story, create a new example project (package) that will create a top-level Express app, mount LB4 app with a single Note model, and keep all artifacts in the same npm package.
- Create a follow-up story to create a more advanced example showing how to compose Express and LB4 components via monorepo layout.

raymondfeng:
> I am proposing the following steps forward:
> * As part of this user story, create a new example project (package) that will create a top-level Express app, mount LB4 app with a single Note model, and keep all artifacts in the same npm package.
> * Create a follow-up story to create a more advanced example showing how to compose Express and LB4 components via monorepo layout.

I'm good with the proposal. Keep the LB4 app as simple as possible instead of duplicating `examples/todo`.

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,342 @@
...
+   create src/controllers/note.controller.ts
+   update src/controllers/index.ts
+
+Controller note was created in src/controllers/
+```
+
+### Express Application
+
+Rename **src/application.ts** to **src/note.application.ts** and create a new
+file **src/application.ts**. In the new file, create your Express class:
```
bajtos: I find it slightly confusing to have `note.application.ts` and `application.ts` next to each other. Can we use a different name for the file hosting Express code? For example `src/server.ts`?

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,342 @@
...
+
+```sh
+npm i --save express
```
bajtos: I am proposing to either abbreviate all npm flags (`npm i -s`) or none (`npm install --save`).

I think we also need to `npm i -S "@types/express`?

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,342 @@
...
+```
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import * as express from 'express';
+
+export class ExpressApplication {
```
bajtos: Have you considered exporting the Express app directly? I find `ExpressApplication` as unnecessary.

```ts
export const app = express();
```

examples/express-composition/src/__tests__/acceptance/test-helper.ts
```diff
@@ -0,0 +1,60 @@
+import {givenHttpServerConfig, Client, supertest} from '@loopback/testlab';
+import {Request, Response} from 'express';
+import {NoteApplication} from '../../note-application';
+import {ExpressApplication} from '../../application';
+import {Note} from '../../models/note.model';
+import * as path from 'path';
+import * as http from 'http';
+import * as pEvent from 'p-event';
+
+export async function setupExpressApplication(): Promise<AppWithClient> {
```
bajtos: 👎 

IIUC, you are duplicating the code setting up the Express application here in tests again.

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,358 @@
...
+
+    ```sh
+    $ npm start
+
+    Server is running at http://127.0.0.1:3000
+    ```
+
+### Create your LoopBack Application
+
+#### Scaffold your Application
```
raymondfeng: For the record: I'm wondering if we should omit the steps to build up a `Note` application with LoopBack 4 (we already have tutorial for that) and go directly to `Express Class`.

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,358 @@
...
+? What is the name of your CRUD repository? NoteRepository
+? What is the type of your ID? number
+? What is the base HTTP path name of the CRUD operations? /notes
+   create src/controllers/note.controller.ts
+   update src/controllers/index.ts
+
+Controller note was created in src/controllers/
+```
+
+### Express Class
```
raymondfeng: The title should be more meaningful, such as `Create a facade Express application`

docs/site/express-with-lb4-rest-tutorial.md
dhmlau: +1 on just showing the delta 
Perhaps use `todo` application (don't think we have `Note` application)?

nabdelgadir: @raymondfeng I tried to omit the explanations and only put the CLI parts, but I can try to trim that down even further (e.g. create a `Note` repository with `DsDataSource` and the `Note` model). What do you think?

@dhmlau we had a discussion above about using the `todo` application, but ultimately decided to use a new, simple `Note` one (see https://github.com/strongloop/loopback-next/pull/2306#issuecomment-459277370 and its replies).

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,358 @@
...
+}
+```
+
+Now, inside the constructor, we're going to add the basepath and expose the
+front-end assets via Express:
+
+```ts
+this.lbApp.basePath('/api');
+this.app.use(this.lbApp.requestHandler);
+```
```
raymondfeng: Now we should be able to use `this.app('/api', this.lbApp.requestHandler)`. The bug was fixed by https://github.com/strongloop/loopback-next/pull/2387

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,51 @@
...
+  private server: http.Server;
+
+  constructor(options: ApplicationConfig = {}) {
+    this.app = express();
+    this.lbApp = new NoteApplication(options);
+
+    // Expose the front-end assets via Express, not as LB4 route
+    this.lbApp.basePath('/api');
+    this.app.use(this.lbApp.requestHandler);
+
```
raymondfeng: Let's switch to `this.app.use('/api', this.lbApp.requestHandler);`

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  async start() {
+    this.server = this.app.listen(3000);
+    await pEvent(this.server, 'listening');
+  }
+
+  // For testing purposes
+  async stop() {
+    this.server.close();
```
jannyHou: Is `this.server.close()` an async function? if not then `stop` doesn't need to be an async function :)

You can refer to the code in [rpc server](https://github.com/strongloop/loopback-next/blob/2b7f425b2fcab9f2b87f04c5f0190cc4986323a0/examples/rpc-server/src/rpc.server.ts#L39-L42) about how to stop the server properly. We need to close until event `close` is emitted.

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  async getLBApp() {
```
jannyHou: it doesn't need to be an async function.

examples/express-composition/src/server.ts
raymondfeng: See https://github.com/strongloop/loopback-next/blob/master/packages/http-server/src/http-server.ts#L117

examples/express-composition/src/server.ts
raymondfeng: We don't usually use `getXXX()`. Make the property `public readonly` or create a `get` function such as:
```ts
public readonly lbApp;
```
or

```ts
get loopBackApp() { return this.lbApp; } 
```

docs/site/express-with-lb4-rest-tutorial.md
dhmlau: What I meant was to use todo app as the starting point to build this tutorial, and we just need to show the additional steps required to mount Express.

docs/site/express-with-lb4-rest-tutorial.md
nabdelgadir: @raymondfeng I updated it to use `this.app('/api', this.lbApp.requestHandler)`, but the `/api/explorer` endpoint still leads to `Fetch error Not Found /openapi.json`. But it works if I go to: http://explorer.loopback.io/?url=http://127.0.0.1:3000/api/openapi.json.

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+#### Add Note Controller
+
+To complete the `Note` application, create a controller using the
+`lb4 controller note` command, with the `REST Controller with CRUD functions`
+type, `Note` model, and `NoteRepository` repository. The `id`'s type will be
+`number` and base HTTP path name is the default `/notes`.
+
+### Create a Facade Express Application
+
+Create a new file **src/server.ts** to create your Express class:
```
b-admike: Nitpick: since we show installing dependencies first, can we mention `Let's install dependencies for express module and create a new file...` here?

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+
+And add the
+[public/express.html](https://github.com/strongloop/loopback-next/blob/master/examples/express-composition/public/express.html)
+file to your project.
+
+Finally, we can add functions to boot the `Note` application and start the
+Express application:
+
+```sh
+npm install --save p-event
```
b-admike: Nitpick: ditto i.e. can we mention why we are installing `p-event`?

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+}
+```
+
+Now that our **src/server.ts** file is ready, then we can modify our
+**src/index.ts** file to start the application:
+
+{% include code-caption.html content="src/index.ts" %}
+
+```ts
+import {ExpressServer} from './server';
```
b-admike: `ExpressApplication` here?

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+import {ApplicationConfig} from '@loopback/core';
+
+export {ExpressServer};
```
b-admike: ditto

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+
+export async function main(options: ApplicationConfig = {}) {
+  const server = new ExpressServer(options);
```
b-admike: ditto `ExpressApplication`. Might want to rename the assigned `server` variable too.

docs/site/express-with-lb4-rest-tutorial.md
```diff
@@ -0,0 +1,304 @@
...
+  constructor(options: ApplicationConfig = {}) {
+    //...
+  }
+
+  async boot() {
+    await this.lbApp.boot();
+  }
+
+  async start() {
+    const server = this.app.listen(3000);
```
b-admike: This probably has been asked before, but why don't we need to start the `lb4app` instance as well?

examples/express-composition/src/migrate.ts
```diff
@@ -0,0 +1,20 @@
+import {NoteApplication} from './application';
+
+export async function migrate(args: string[]) {
+  const existingSchema = args.includes('--rebuild') ? 'drop' : 'alter';
+  console.log('Migrating schemas (%s existing schema)', existingSchema);
+
+  const app = new NoteApplication();
```
b-admike: Should we use `const app = new ExpressApplication()` here instead? Since it's a wrapper/facade of our Note application IIUC.

docs/site/express-with-lb4-rest-tutorial.md
nabdelgadir: The class is called `ExpressServer` currently.

b-admike: Hmm according to https://github.com/strongloop/loopback-next/pull/2306/files#diff-27f00d9e00bf82aa50db95dca9d2292fR141 I see it as `ExpressApplication`. As long as we are consistent with the naming convention, I'm fine with either.

nabdelgadir: Oh, good catch! Thanks :)

docs/site/express-with-lb4-rest-tutorial.md
nabdelgadir: Before it was causing some errors with routing (and it worked without it), but looks like it was resolved so I'll add it back.

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+  public async start() {
+    await this.lbApp.start();
+    this.server = this.app.listen(3000);
+    await pEvent(this.server, 'listening');
+  }
+
+  // For testing purposes
+  public async stop() {
+    if (!this.server) return;
+    this.server.close();
```
b-admike: If we decide to start `lbApp` in `start()` above, then we should also stop it IMO.

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Serve static files in the public folder
+    this.app.use(express.static('public'));
+  }
+
+  public async boot() {
+    await this.lbApp.boot();
+  }
+
+  public async start() {
+    await this.lbApp.start();
```
b-admike: Sorry, my previous comment was more of a question instead of a request. Do we have configuration changes in place to make sure that LoopBack and Express are listening on different ports? @raymondfeng @bajtos @strongloop/loopback-maintainers thoughts?

examples/express-composition/src/server.ts
nabdelgadir: Will add this in a follow-up PR or just remove starting the LoopBack app depending on https://github.com/strongloop/loopback-next/pull/2306#discussion_r261250257

examples/express-composition/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.server = this.app.listen(3000);
```
bajtos: IMO, we should honor `rest.port` and `rest.host` provided via `ApplicationConfig`.

Without that, most of the configuration code in `index.js` is effectively ignored.

```js
  // Run the application
  const config = {
    rest: {
      port: +process.env.PORT || 3000,
      host: process.env.HOST || 'localhost',
      openApiSpec: {
        // useful when used with OASGraph to locate your application
        setServersFromRequest: true,
      },
    },
  };
```

What's worse, the application cannot be deployed to http://12factor.net environment like https://www.heroku.com

/cc @nabdelgadir

nabdelgadir: I'll make a new PR for this, thanks!

Issue #1573, opened 2018-07-30
Validate parameter values against their schema


bajtos:
OpenAPI spec allows developers to provide additional restrictions on input parameters via parameter schema.

For example, a `pageSize` parameter can be defined as follows:

```yaml
parameters:
 - name: pageSize
   in: query
   description: Number of records to return.
   required: false
   schema:
     type: integer
     format: int32      
     minimum: 0
     maximum: 100
```

At the moment, `@loopback/rest`'s sequence action `parseParams` validates parameter values against their schema for `in: body` arguments only.

We should modify `parseParams`to validate all parameter values against their schema. Leverage AJV to perform the actual validation.

hbakhtiyor:
@bajtos, need to create another `packages/rest/src/validation/params.validator.ts` ?

bajtos:
> need to create another packages/rest/src/validation/params.validator.ts ?

Possibly. I think we should refactor the code in `request-body.validator.ts` and extract pieces that will be shared with parameter validator, e.g. `convertToJsonSchema`, `validateValueAgainstSchema` and `createValidator`, so there may be more new files to add in the end.

YaelGit:
@bajtos is there more information you can share in regards to how we can reproduce the issue described above.
If issue can be reproduced using the example of todo-list, or other ways
Please share if available,

jannyHou:
I think PR https://github.com/strongloop/loopback-next/pull/2059 is relevant.

YaelGit:
Hello @bajtos, please note i would like to start handling this issue #1573. According to my understanding the needed is to add similar functions of validation as there are for validation of parameters of the body (according to the schema), for the parameters of the query.

YaelGit:
Hello @bajtos, according to my understanding this is rather a feature and not a minor bug therefore will need a solution design. Do you think this is a good stage to perform the pull request ? or should i do that only after the deign is complete?

bajtos:
> according to my understanding this is rather a feature and not a minor bug 

Makes sense, I updated the issue labels accordingly 👍 

> therefore will need a solution design. Do you think this is a good stage to perform the pull request ? or should i do that only after the deign is complete?

I think the most important the design work has been already done while implementing request-body validation.

@YaelGit Please go ahead and open a pull request. Feel free to open it early, even before the implementation and tests are complete. That way we can provide you with early feedback and you can save time by keeping possible reworks smaller in scope.

YaelGit:
Hi Bajtos,

[...]

PR #2307, opened 2019-01-30
feat(rest): implement query parameter validation (issue 1573) 

YaelGit:
Implements #1573 

## Checklist


packages/rest/src/parser.ts
```diff
@@ -38,11 +39,18 @@ export async function parseOperationArgs(
     operationSpec,
     request,
   );
+
+  const query = await requestBodyParser.loadRequestBodyIfNeeded(
+    operationSpec,
+    request,
+  );
+
```
raymondfeng: This is suspicious. Why do we parse the request body again into `query`?

packages/rest/src/validation/request-query.validator.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Check whether the request query is valid according to the provided OpenAPI schema.
+ * The JSON schema is generated from the OpenAPI schema which is typically defined
+ * by `@requestQuery()`.
+ * The validation leverages AJS schema validator.
+ * @param query The request query parsed from an HTTP request.
+ * @param requestQuerySpec The OpenAPI requestQuery specification defined in `@requestQuery()`.
+ * @param globalSchemas The referenced schemas generated from `OpenAPISpec.components.schemas`.
+ */
+export function validateRequestQuery(
+  query: RequestBody,
```
raymondfeng: If we want to reuse `RequestBody` for query, we should rename the type as something like `ValueWithSchema`.

packages/rest/src/validation/request-query.validator.ts
```diff
@@ -0,0 +1,150 @@
...
+      {},
+      {
+        allErrors: true,
+      },
+      options,
+    ),
+  );
+
+  return ajv.compile(schemaWithRef);
+}
```
raymondfeng: There seem to be quite a bit duplicate code as request-body-validator. Can we refactor the code for better reuse?

YaelGit: @raymondfeng -Thanks for the review and comments.
I will try to reply: I was actually looking to differentiate the 2 in order to allow/take into consideration future changes and needs-do you think this is un-needed? will the only difference be - the extraction of the query params and the structure to be sent to same validation function to leverage the AJV ? we need not forget, we also need different error messages.  Please share your thoughts.

packages/rest/src/validation/request-query.validator.ts
```diff
@@ -0,0 +1,150 @@
...
+  options: RequestQueryValidationOptions = {},
+) {
+  const required = requestQuerySpec && requestQuerySpec.required;
+
+  if (required && query.value == undefined) {
+    const err = Object.assign(
+      new HttpErrors.BadRequest('Request query is required'),
+      {
+        code: 'MISSING_REQUIRED_PARAMETER',
+        parameterName: 'request query',
```
raymondfeng: It would be nice to see which `query` param.

packages/rest/src/validation/request-query.validator.ts
raymondfeng: I would create a utility function that validates `ValueWithSchema` using AJV.

packages/rest/src/validation/request-query.validator.ts
YaelGit: is this sufficient?
query_validation_errors

packages/rest/src/validation/request-query.validator.ts
YaelGit: @raymondfeng - do you mean we should duplicate the type requestBody in types.ts and then use this new type (ValueWithSchema) in request-query.validator.ts?  
should we also duplicate the BodyParser, BodyParserFunction? and then duplicate the code in body-parser.ts?

raymondfeng: No duplication. What I meant is to refactor `RequestBody` to `ValueWithSchema` into its own file and use it for both body and query validation.

packages/rest/src/validation/request-query.validator.ts
raymondfeng: Sure.

bajtos:
@YaelGit ping 👋 

What's the status of this pull request? Are you still keen to get it finished?

> Merge branch 'master' into issue-1573

Please note we are avoiding merge commits to keep our git history neatly linear. We use `git rebase` instead, see [How to rebase your branch](https://loopback.io/doc/en/contrib/code-contrib.html#how-to-rebase-your-branch).

packages/rest/src/parser.ts
```diff
@@ -77,11 +92,30 @@ function buildOperationArguments(
...
+      (<any>schemasValue)[paramName] = paramSchema;
+    }
+  }
 
+  //if query parameters from URL - send to query validation
+  if (isQuery) {
+    query.value = queryValue;
+    globalSchemas = {properties: schemasValue};
+    query.schema = globalSchemas;
+    validateRequestQuery(query, operationSpec.requestBody, globalSchemas);
```
bajtos: IIUC, you are building an object with all parameters from the query string, and then validate this object in one pass.  I have two reservations about this approach:
 - It seems overly complicated to me.
 - There are more parameter sources besides `query` and `body`, we need to validate parameters from all sources (e.g. `path` and `header`).

Have you considered to call `ajv` validator repeatedly, once for each parameter?

YaelGit: Hello @bajtos ,
sorry for the delay - yes i do plan to close the issue soon and in regards to your comments:
1. I will do the rebase as requested.
2. Regarding the validation of all parameters all together - i did it in this way due to following same validation standard done for body parameters.
3. Regrading the request for handling other parameter types, i think a new issue should be opened for other parameter types as the initial issue was for query only.

YaelGit:
Unfortunately due to some technical issue with my Laptop - things went wrong while trying to rebase.
I will need to reopen a new PR to re-enter my code.

YaelGit:
Hello @bajtos @raymondfeng,
Due to technical issues, I have opened a new PRin the following link:
https://github.com/strongloop/loopback-next/pull/2382
The PR includes implementation regarding all of the comments above.

PR #2281, opened 2019-01-24
build(build): upgrade Mocha to v6 prerelease

boneskull:
also replaces legacy mocha.opts with new .mocharc.json

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

PR #1431, opened 2018-06-14
ci: add Node.js 10 to AppVeyor config (Windows)

bajtos:
This finishes the revert of commit a36e257 (see 5e605c05d from https://github.com/strongloop/loopback-next/pull/1350) and configures AppVeyor CI to use the latest Node.js 10.x version, since the problem we were experiencing in 10.1.0 has been fixed in 10.2.0.

## Checklist

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #2309, opened 2019-01-30
chore(build): use new config format for mocha v6.x

raymondfeng:
Supersedes https://github.com/strongloop/loopback-next/pull/2281.

## Checklist


raymondfeng:
Need to investigate why `(node:1888) ExperimentalWarning: The http2 module is an experimental API.
` is printed on Windows. Maybe the node version is too low.

bajtos:
> Need to investigate why (node:1888) ExperimentalWarning: The http2 module is an experimental API. is printed on Windows. Maybe the node version is too low.

IIRC, AppVeyor is running an old version of Node.js 8.x. HTTP2 API was marked as stable at the time when 8.x was already in LTS state. We are also not able to upgrade to the latest version of Node.js 10.x because the tests are freezing - see https://github.com/strongloop/loopback-next/pull/1431

packages/build/package.json
```diff
@@ -20,7 +20,7 @@
     "debug": "^4.0.1",
     "fs-extra": "^7.0.0",
     "glob": "^7.1.2",
-    "mocha": "^5.1.1",
+    "mocha": "^6.0.0-1",
```
bajtos: As commented in the original PR, I am reluctant to use pre-release versions. Can we wait until `6.0.0` is released please, and then use that version instead?

packages/build/bin/run-mocha.js
```diff
@@ -31,8 +31,8 @@ function run(argv, options) {
   if (typeof options === 'boolean') options = {dryRun: options};
   options = options || {};
   if (setMochaOpts) {
```
bajtos: setMochaOpts is checking for presence of `--opts`.  So IIUC the current implementation, if `--opts` is not provided, we add `--config`. Which means that `lb-mocha --config my.json` will add `--config .mocharc.json` to the command line. Please add a test to verify this scenario.

How about backwards compatibility? If we replace `--opts` with `--config` in our check and stop detecting `--opts`, then I think this is a breaking change requiring a semver-major release. It may be more conventient for our users if we keep support for both `--opts` and `--config` options.

packages/build/package.json
raymondfeng: Sure. The PR was created to test out 6.x.

packages/build/bin/run-mocha.js
raymondfeng: Maybe it's better to bump `@loopback/build` to the next major version. Using `--opts` for mocha@6.x seems to be add confusion.

bajtos: I am fine with publishing a new semver-major version of `@loopback/build`. See https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#making-breaking-changes for a check-list.

bajtos:
FWIW, `mocha@6.0.0` is public: https://github.com/mochajs/mocha/releases/tag/v6.0.0

bajtos:
@raymondfeng what's the status of this pull request? How is it different from #2422 which you have already landed?

raymondfeng:
I would like to land this PR, which converts to the new `--config` and `mocharc.json`.

bajtos:
> I would like to land this PR

Sounds good to me. We need to follow the process for a new major version now. 

Quoting from https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#making-breaking-changes
> - Describe incompatibilites for release notes
> - Look for more breaking changes to include in the release
> - Update list of supported versions

Most importantly, the commit message introducing the breaking change must include `BREAKING CHANGE: ` footer - see https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#footer-optional.

We will need to update `build`'s README to add a table listing supported major versions, I am ok to do that outside of this pull request if you prefer.

As for other breaking changes to include in the release, I am not really aware of any.

bajtos:
> We will need to update `build`'s README to add a table listing supported major versions.

Let's apply the same format we use in LB3 repositories, see e.g.
https://github.com/strongloop/loopback#module-long-term-support-policy

packages/build/bin/run-mocha.js
raymondfeng: It turns out that mocha 6.x continues to support `mocha.opts`. I have improved the PR to respect `--opts` and `mocha.opts` in addition to `.mocharc.*` formats. It should be now backward compatible.

raymondfeng:
@bajtos I think I have made it backward compatible except breaking changes from `mocha 6.x` itself.

packages/build/bin/utils.js
```diff
@@ -214,9 +214,18 @@ function isOptionSet(opts, ...optionNames) {
...
-function mochaOptsFileProjectExists() {
-  const configFile = path.join(getPackageDir(), 'test/mocha.opts');
-  return fs.existsSync(configFile);
+function mochaConfiguredForProject() {
+  const configFiles = [
+    '.mocharc.js',
+    '.mocharc.json',
+    '.mocharc.yaml',
+    '.mocharc.yml',
+    'mocha.opts',
```
bajtos: I believe the default config file is `test/mocha.opts`, not `mocha.opts`. Has this changed in `mocha@6`?

raymondfeng: Good catch. I fixed it.

PR #2310, opened 2019-01-30
chore: fix test cases to be compatible with mocha@6.x

raymondfeng:
See https://github.com/mochajs/mocha/pull/3550

## Checklist

### DIFF
--- a/packages/rest/test/unit/rest.component.unit.ts
+++ b/packages/rest/test/unit/rest.component.unit.ts
@@ -24,22 +24,30 @@ import {
 const SequenceActions = RestBindings.SequenceActions;
 describe('RestComponent', () => {
   describe('Providers', () => {
-    describe('Default implementations are bound', async () => {
-      const app = new Application();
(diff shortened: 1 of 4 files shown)


packages/rest/test/unit/rest.component.unit.ts
```diff
@@ -24,22 +24,30 @@ import {
...
 
-      const comp = await app.get<Component>('components.RestComponent');
-      for (const key in comp.providers || {}) {
-        it(key, async () => {
-          const result = await app.get(key);
-          const expected: Provider<BoundValue> = new comp.providers![key]();
-          expect(result).to.deepEqual(expected.value());
-        });
-      }
+      it('', async () => {
```
b-admike: Do we get a deprecation warning if we don't wrap this block of code with the empty `it` block?

raymondfeng: Yes. It's also required to trigger `before` and run multiple tests in a loop.

PR #2314, opened 2019-01-30
feat(rest): add redirect route

### LINKED ISSUE #2022: Sugar API for defining URL redirects
_This is a follow-up for https://github.com/strongloop/loopback-next/pull/2014._

When using LB4+ to serve HTML pages, it's useful to add a trailing slash to the URL when serving a folder, for example redirect `/explorer` to `/explorer/`. Without this redirect, relative URLs to assets like CSS & JS files are incorrectly resolved. For example, when served from `/explorer`, relative links like `./swagger-ui.css` are resolved in the parent directory, e.g. `/swagger-ui.css` instead of `/explorer/swagger-ui.css`.

Right now, a redirect can be implemented using a controller route that's hidden from the documentation and uses HTTP `response` object to send back the redirect. Such solution requires a lot of code and feels a bit hacky.

Let's make redirects a first-class feature in LB4 and provide a high-level API that's easy to use.

For example:

```ts
restApp.redirect('/explorer', '/explorer/');
restServer.redirect('/explorer', '/explorer/');
```

Under the hood, this can be implemented as a new Route class, for example:

```ts
app.route(new RedirectRoute('/explorer', '/explorer/'));
```

A mock-up implementation of RedirectRoute:

```ts
export class RedirectRoute implements RouteEntry, ResolvedRoute {
  // ResolvedRoute API
  readonly pathParams: PathParameterValues = [];
  readonly schemas: SchemasObject = {};

  // RouteEntry implementation
  readonly verb: string = 'get';
  readonly get path(): string { return this.sourcePath; }
  // ...

  constructor(
    public readonly sourcePath: string, 
    public readonly targetPath: string, 
    public statusCode: number = 303,
  ) {
    this.path = sourcePath;
  }

  async invokeHandler(
    {response}: RequestContext,
    args: OperationArgs,
  ): Promise<OperationRetval> {
    response.redirect(this.statusCode, this.targetPath);
  }

  // ...
}
```

## Acceptance criteria

https://github.com/strongloop/loopback-next/pull/2512
- [x] The implementation, including unit/integration/acceptance tests 
  - Most of the tests should be written for `RestServer`.
  - Add one or few tests for `RestApplication` at integration or acceptance level, just to ensure the new RestApplication API is covered.
- [x] Documentation

**TODO**
- [ ] Redirect to dynamically computed location (see the discussion below)
- [ ] Search for all places calling `.redirect(` and consider updating them to use the new route and/or the new RestServer/RestApplication sugar APIs. E.g. REST API Explorer.
  - Redirect to externally hosted swagger-ui
  - Redirect from `/explorer` to `/explorer/`
- [ ] Redirect to a location that's full URL (`http://example.com`) instead of a local path (`/home`). The trick is to skip appending `basePath`.
- [ ] Honor [`req.baseUrl`](https://expressjs.com/en/4x/api.html#req.baseUrl) when the LB4 app is mounted on an external Express application.

sanadHaj:
Fixes #2022 

Implement RedirectRoute to redirect request route , beside SwaggerRoute that extend from RedirectRoute to redirect swagger ui including some refactoring of rest server to support the new swagger route instead of express middleware   
The implementation, including 12 (unit/integration/acceptance) tests.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
@@ -735,6 +735,53 @@ describe('Routing', () => {
         .get('/greet/world')
         .expect(200, 'HELLO WORLD');
     });
+
+    it('gives precedence to redirect routes over controller methods', async () => {
+      class MyController {
+        @get('/hello', {
+          responses: {},
+        })
+        hello(): string {
+          return 'hello';
+        }
+        @get('/hello/world')
+        helloWorld() {
+          return `hello world`;
+        }
+      }
+      const app = givenAnApplication();
+      const server = await givenAServer(app);
+      server.basePath('/api');
+      server.redirect('/test/hello', '/hello/world');
+      givenControllerInApp(app, MyController);
+      const response = await whenIMakeRequestTo(server)
+        .get('/api/test/hello')
+        .expect(303);
+      // new request to verify the redirect target
+      await whenIMakeRequestTo(server)
+        .get(response.header.location)
+        .expect(200, 'hello world');
+    });
+
+    it('gives precedence to redirect routes over route methods', async () => {
+      const app = new RestApplication();
+      app.route(
+        'get',
+        '/greet/{name}',
+        anOperationSpec()
+          .withParameter({name: 'name', in: 'path', type: 'string'})
+          .build(),
+        (name: string) => `hello ${name}`,
+      );
+      app.redirect('/hello/john', '/greet/john');
+      const response = await whenIMakeRequestTo(app)
+        .get('/hello/john')
+        .expect(303);
+      // new request to verify the redirect target
+      await whenIMakeRequestTo(app)
+        .get(response.header.location)
+        .expect(200, 'hello john');
+    });
   });
 
   /* ===== HELPERS ===== */
--- a/packages/rest/src/__tests__/integration/rest.application.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.application.integration.ts
@@ -7,7 +7,7 @@ import {anOperationSpec} from '@loopback/openapi-spec-builder';
 import {Client, createRestAppClient, expect} from '@loopback/testlab';
 import * as fs from 'fs';
 import * as path from 'path';
-import {RestApplication, RestServer, RestServerConfig} from '../..';
+import {RestApplication, RestServer, RestServerConfig, get} from '../..';
 
 const ASSETS = path.resolve(__dirname, '../../../fixtures/assets');
 
@@ -145,8 +145,28 @@ describe('RestApplication (integration)', () => {
     });
   });
 
+  it('can app redirect routes with custom status', async () => {
+    givenApplication();
+    restApp.controller(DummyController);
+    restApp.redirect('/fake/html', '/html', 304);
+    await restApp.start();
+    client = createRestAppClient(restApp);
+    const response = await client.get('/fake/html').expect(304);
+    await client.get(response.header.location).expect(200, 'Hi');
+  });
+
   function givenApplication(options?: {rest: RestServerConfig}) {
     options = options || {rest: {port: 0, host: '127.0.0.1'}};
     restApp = new RestApplication(options);
   }
 });
+
+class DummyController {
+  constructor() {}
+  @get('/html', {
+    responses: {},
+  })
+  ping(): string {
+    return 'Hi';
+  }
+}
--- a/packages/rest/src/__tests__/integration/rest.server.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.server.integration.ts
@@ -735,6 +735,39 @@ paths:
       );
       expect(response.body.servers).to.containEql({url: '/api'});
     });
+
+    it('can redirect routes when it does not exist', async () => {
+      server.controller(DummyController);
+      server.redirect('/page/html', '/html');
+      const response = await createClientForHandler(server.requestHandler)
+        .get('/api/page/html')
+        .expect(303);
+      await createClientForHandler(server.requestHandler)
+        .get(response.header.location)
+        .expect(200, 'Hi');
+    });
+
+    it('can redirect routes', async () => {
+      server.controller(DummyController);
+      server.redirect('/hello', '/endpoint');
+      const response = await createClientForHandler(server.requestHandler)
+        .get('/api/hello')
+        .expect(303);
+      await createClientForHandler(server.requestHandler)
+        .get(response.header.location)
+        .expect(200, 'hello');
+    });
+
+    it('can server redirect routes with custom status', async () => {
+      server.controller(DummyController);
+      server.redirect('/hello', '/endpoint', 304);
+      const response = await createClientForHandler(server.requestHandler)
+        .get('/api/hello')
+        .expect(304);
+      await createClientForHandler(server.requestHandler)
+        .get(response.header.location)
+        .expect(200, 'hello');
+    });
   });
 
   async function givenAServer(
@@ -762,6 +795,12 @@ paths:
     })
     ping(): string {
       return 'Hi';
+    }
+    @get('/endpoint', {
+      responses: {},
+    })
+    hello(): string {
+      return 'hello';
     }
   }
 
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -640,8 +640,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param source URL path of the redirect endpoint
    * @param target URL path of the endpoint
    */
-  redirect(source: string, target: string): Binding {
-    return this.route(new RedirectRoute(source, this._basePath + target));
+  redirect(source: string, target: string, statusCode?: number): Binding {
+    return this.route(
+      new RedirectRoute(source, this._basePath + target, statusCode),
+    );
   }
 
   // The route for static assets
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -42,6 +42,7 @@ import {
   RouteEntry,
   RoutingTable,
   StaticAssetsRoute,
+  RedirectRoute,
 } from './router';
 import {DefaultSequence, SequenceFunction, SequenceHandler} from './sequence';
 import {
@@ -627,6 +628,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
         methodName,
       ),
     );
+  }
+
+  /**
+   * Redirect route by invoking a handler function.
+   *
+   * ```ts
+   * server.redirect('/explorer', '/explorer/');
+   * ```
+   *
+   * @param source URL path of the redirect endpoint
+   * @param target URL path of the endpoint
+   */
+  redirect(source: string, target: string): Binding {
+    return this.route(new RedirectRoute(source, this._basePath + target));
   }
 
   // The route for static assets
--- a/packages/rest/src/router/redirect-route.ts
+++ b/packages/rest/src/router/redirect-route.ts
@@ -37,6 +37,6 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
   }
 
   describe(): string {
-    return `RedirectRoute.${this.sourcePath}.${this.targetPath}`;
+    return `RedirectRoute from "${this.sourcePath}" to "${this.targetPath}"`;
   }
 }
--- a/packages/rest/src/router/redirect-route.ts
+++ b/packages/rest/src/router/redirect-route.ts
@@ -1,0 +1,42 @@
+import {RouteEntry, ResolvedRoute} from '.';
+import {RequestContext} from '../request-context';
+import {OperationObject, SchemasObject} from '@loopback/openapi-v3-types';
+import {OperationArgs, OperationRetval, PathParameterValues} from '../types';
+
+export class RedirectRoute implements RouteEntry, ResolvedRoute {
+  // ResolvedRoute API
+  readonly pathParams: PathParameterValues = [];
+  readonly schemas: SchemasObject = {};
+
+  // RouteEntry implementation
+  readonly verb: string = 'get';
+  readonly path: string = this.sourcePath;
+  readonly spec: OperationObject = {
+    description: 'LoopBack Redirect route',
+    'x-visibility': 'undocumented',
+    responses: {},
+  };
+
+  constructor(
+    public readonly sourcePath: string,
+    public targetPath: string,
+    public statusCode: number = 303,
+  ) {
+    this.sourcePath = sourcePath;
+  }
+
+  async invokeHandler(
+    {response}: RequestContext,
+    args: OperationArgs,
+  ): Promise<OperationRetval> {
+    response.redirect(this.statusCode, this.targetPath);
+  }
+
+  updateBindings(requestContext: RequestContext) {
+    // no-op
+  }
+
+  describe(): string {
+    return `RedirectRoute.${this.sourcePath}.${this.targetPath}`;
+  }
+}
--- a/packages/rest/src/__tests__/unit/rest.server/rest.server.redirect.unit.ts
+++ b/packages/rest/src/__tests__/unit/rest.server/rest.server.redirect.unit.ts
@@ -1,0 +1,19 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {Application} from '@loopback/core';
+import {RestServer, RestComponent} from '../../..';
+
+describe('RestServer.redirect()', () => {
+  it('binds the redirect route', async () => {
+    const app = new Application();
+    app.component(RestComponent);
+    const server = await app.getServer(RestServer);
+    server.redirect('/test', '/test/');
+    let boundRoutes = server.find('routes.*').map(b => b.key);
+    expect(boundRoutes).to.containEql('routes.get %2Ftest');
+  });
+});
--- a/packages/rest/src/rest.application.ts
+++ b/packages/rest/src/rest.application.ts
@@ -238,8 +238,8 @@ export class RestApplication extends Application implements HttpServerLike {
    * @param source URL path of the redirect endpoint
    * @param target URL path of the endpoint
    */
-  redirect(source: string, target: string): Binding {
-    return this.restServer.redirect(source, target);
+  redirect(source: string, target: string, statusCode?: number): Binding {
+    return this.restServer.redirect(source, target, statusCode);
   }
 
   /**
--- a/packages/rest/src/rest.application.ts
+++ b/packages/rest/src/rest.application.ts
@@ -229,6 +229,20 @@ export class RestApplication extends Application implements HttpServerLike {
   }
 
   /**
+   * Redirect route by invoking a handler function.
+   *
+   * ```ts
+   * app.redirect('/explorer', '/explorer/');
+   * ```
+   *
+   * @param source URL path of the redirect endpoint
+   * @param target URL path of the endpoint
+   */
+  redirect(source: string, target: string): Binding {
+    return this.restServer.redirect(source, target);
+  }
+
+  /**
    * Set the OpenAPI specification that defines the REST API schema for this
    * application. All routes, parameter definitions and return types will be
    * defined in this way.
--- a/packages/rest/src/router/index.ts
+++ b/packages/rest/src/router/index.ts
@@ -9,6 +9,8 @@ export * from './base-route';
 export * from './controller-route';
 export * from './handler-route';
 export * from './static-assets-route';
+export * from './redirect-route';
+export * from './swagger-route';
 
 // routers
 export * from './rest-router';
--- a/packages/rest/src/router/index.ts
+++ b/packages/rest/src/router/index.ts
@@ -10,7 +10,6 @@ export * from './controller-route';
 export * from './handler-route';
 export * from './static-assets-route';
 export * from './redirect-route';
-export * from './swagger-route';
 
 // routers
 export * from './rest-router';


sanadHaj:
@bajtos  @raymondfeng

packages/rest/test/unit/rest.server/rest.server.redirect.unit.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+});
+
+describe('RestServer.redirect()', () => {
+  it('binds the redirect route', async () => {
+    const app = new Application();
+    app.component(RestComponent);
+    const server = await app.getServer(RestServer);
+    server.redirect('/test', '/test/');
+    let boundRouets = server.find('routes.*').map(b => b.key);
```
raymondfeng: Typo: it should be `boundRoutes`

packages/rest/test/unit/rest.application/rest.application.unit.ts
```diff
@@ -50,5 +50,13 @@ describe('RestApplication', () => {
         ERR_NO_MULTI_SERVER,
       );
     });
+
+    it('when attempting to redirect endpoint', () => {
+      const app = new RestApplication();
+      const server = app.restServer;
+      server.redirect('/test', '/test/');
+      let boundRouets = server.find('routes.*').map(b => b.key);
```
raymondfeng: Typo: it should be `boundRoutes`.

packages/rest/src/router/swagger-route.ts
```diff
@@ -0,0 +1,44 @@
...
+    const config = this.config.apiExplorer!;
+    if (config.disabled) {
+      response.statusCode = 404;
+      response.send();
+      return;
+    }
+    this.targetPath = this.extarctFullUri(request, config);
+    // https://github.com/strongloop/loopback-next/issues/1623
+    response = response.set('location', this.targetPath);
+    response.redirect(this.statusCode, this.targetPath);
```
raymondfeng: `response.redirect` sets up `Location` header, doesn't it?

packages/rest/src/router/swagger-route.ts
```diff
@@ -0,0 +1,44 @@
+import {RedirectRoute} from '.';
+import {RequestContext} from '../request-context';
+import {OperationArgs, OperationRetval, Request} from '../types';
+import {RestServerConfig, ApiExplorerOptions} from '../rest.server';
+import {getUrlForClient, getProtocolForRequest} from '../rest-utils';
+
+export class SwaggerRoute extends RedirectRoute {
```
raymondfeng: Maybe `SwaggerUIRoute` is a better name?

packages/rest/src/router/swagger-route.ts
```diff
@@ -0,0 +1,44 @@
+import {RedirectRoute} from '.';
```
raymondfeng:
```ts
import {RedirectRoute} from './redirect-route';
```

packages/rest/src/router/swagger-route.ts
sanadHaj: out of the blue no ! i was expect to be included it seems express was doing the job .

packages/rest/src/router/swagger-route.ts
sanadHaj: +1

packages/rest/src/router/swagger-route.ts
sanadHaj: +1

packages/rest/src/router/swagger-route.ts
raymondfeng: I confirm that Express does the job - https://github.com/expressjs/express/blob/master/lib/response.js#L922

sanadHaj: using express middileware was set the header location, it's doesn't appear since using the route.
perhaps there something  wrong in LB4 Routing ,  reset the headers ?

sanadHaj: @raymondfeng  
anyway removing this line , will redirect the request without location header,

sanadHaj:
@bajtos  review please

packages/rest/src/router/swagger-route.ts
raymondfeng: We should have a test to check the `location` header to ensure it's set correctly.

sanadHaj: @raymondfeng 
we already have more than one. that check the header location content. 
 https://github.com/strongloop/loopback-next/blob/4d37ff1d1757d4d0fa3677a069c4fbaca051803f/packages/rest/test/integration/rest.server.integration.ts#L459 
https://github.com/strongloop/loopback-next/blob/4d37ff1d1757d4d0fa3677a069c4fbaca051803f/packages/rest/test/integration/rest.server.integration.ts#L493
https://github.com/strongloop/loopback-next/blob/4d37ff1d1757d4d0fa3677a069c4fbaca051803f/packages/rest/test/integration/rest.server.integration.ts#L512
https://github.com/strongloop/loopback-next/blob/4d37ff1d1757d4d0fa3677a069c4fbaca051803f/packages/rest/test/integration/rest.server.integration.ts#L652
etc ..

packages/rest/src/rest.server.ts
```diff
@@ -629,6 +630,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+   *
+   * ```ts
+   * server.redirect('/explorer', '/explorer/');
+   * ```
+   *
+   * @param source URL path of the redirect endpoint
+   * @param target URL path of the endpoint
+   */
+  redirect(source: string, target: string): Binding {
+    return this.route(new RedirectRoute(source, this._basePath + target));
```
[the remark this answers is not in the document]
sanadHaj: > I find this surprising. Why are we prepending `this._basePath` to `target`, but not to `source`? I would expect basePath to be prepended to both or to none. What am I missing?
>
1. adding the the base path to the source will cause to double the base path so the result will be 
something like basePath/basePath/redirectRoute.
2. the reason of adding base path to source to avoid 404 not found cause the location header not holding the base path so when the request come back during the redirect process express/lb4 router will return 404.   

eventually  i wanna to achieve to redirect 
route if not exist this will help developers and give them some flexibility of using redirect api

lets assume i have configured the lb4 app like this ; 
```ts
    restApp.basePath('/api');
    await restApp.controller(barController).lock();
    await restApp.controller(fooController).lock();
    restApp.restServer.redirect('/foo','/bar');
    restApp.restServer.redirect('/notExistRoute','/bar');
    await restApp.start();
```

in the above example the user will not figure out what happening behind the scene , removing the base path to be added as target `/api/bar` a litile bit confusing me . 

the result of requesting `/api/foo` and `/api/notExistRoute` it will be the bar controller response. 
 even when the `api/notExistRoute` missing as route .

packages/rest/src/rest.server.ts
```diff
@@ -199,6 +200,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
 
     this.bind(RestBindings.BASE_PATH).toDynamicValue(() => this._basePath);
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
+    // mount SwaggerUI
```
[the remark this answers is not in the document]
sanadHaj: the logic of handling disable has been moved to swaggerRoute 
i think this unrelated piece  of code that shouldn't be at rest server  level 
https://github.com/strongloop/loopback-next/blob/83cf8a4d6b14475837ec57d2a77ba03c1969f76c/packages/rest/src/router/swagger-route.ts#L22

packages/rest/src/router/redirect-route.ts
```diff
@@ -0,0 +1,42 @@
...
+  ): Promise<OperationRetval> {
+    response.redirect(this.statusCode, this.targetPath);
+  }
+
+  updateBindings(requestContext: RequestContext) {
+    // no-op
+  }
+
+  describe(): string {
+    return `RedirectRoute.${this.sourcePath}.${this.targetPath}`;
```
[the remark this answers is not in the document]
sanadHaj: +1

packages/rest/src/router/redirect-route.ts
```diff
@@ -0,0 +1,42 @@
...
+  readonly spec: OperationObject = {
+    description: 'LoopBack Redirect route',
+    'x-visibility': 'undocumented',
+    responses: {},
+  };
+
+  constructor(
+    public readonly sourcePath: string,
+    public targetPath: string,
+    public statusCode: number = 303,
```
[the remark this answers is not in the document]
sanadHaj: yep, as optional parameter.

sanadHaj:
good point 
> I would like to propose a different solution that's more generic and can be useful to more consumers.
> The built-in redirect needs to obtain the full URL (including protocol, hostname and port) before the redirect URL can be constructed. Inside `@loopback/rest-explorer`, we want the path portion of the URL. I feel these two uses cases may be pretty common and it would be great to support them as first-class features. 

> I am proposing to modify `app.redirect` to and introduce a new variant that will accept a function instead of a string as the "target" argument.
> 
disagree with this line of code the base bath should not be set as part of openapi.json 
to prevent => Not Found http://127.0.0.1:8080/api/openapi.json if baseUrl /api
cause the openapi.json path under root
e.g http://127.0.0.1:8080/openapi.json
for more further detail check this issue  
> ```ts
>   const openApiUrl = `${protocol}://${host}${basePath}/openapi.json`;

it should be 

 ```ts
   const openApiUrl = `${protocol}://${host}/openapi.json`;
```
please see the bellow comment for more information , i think we should support the two use cases 
so we can make redirect accept string as part of path , and accept function as full url when redirect  function requested for full url  we make sure the base path not applied . 
> Besides the comment above and the comments below, I'd like to discuss the impact of `basePath` on redirects. Should we apply basePath to `source`, `target` or both? I'd like us to find a solution that will be least confusing to our users. Personally, I am leaning towards applying `basePath` to both of them. Obviously, when the `target` is a full URL like `https://explorer.loopback.io/...`, then `basePath` must not be applied.

**Comment by bajtos** [comment]
> #### Conflicting files
> `packages/rest/test/acceptance/routing/routing.acceptance.ts`
> `packages/rest/test/integration/rest.application.integration.ts`
> `packages/rest/test/integration/rest.server.integration.ts`
> `packages/rest/test/unit/rest.application/rest.application.unit.ts`

In https://github.com/strongloop/loopback-next/pull/2330, we have moved test files from `test` to `src/__tests__`. Please follow the steps in [How to rebase your branch](https://github.com/strongloop/loopback.io/blob/26bc4788d818662a4e8d4e4c19803ab38ed75655/pages/en/contrib/code-contrib.md#how-to-rebase-your-branch) to resolve the merge conflict. Please do not run `git merge`, it's important to use `git rebase master`.

Sorry for the inconvenience.

File: packages/rest/src/rest.server.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (sanadHaj) [reply packages/rest/src/rest.server.ts]: not relevant after rolling back and discard swagerRoute

File: packages/rest/src/router/redirect-route.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (sanadHaj) [reply packages/rest/src/router/redirect-route.ts]: fixed

File: packages/rest/src/router/redirect-route.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (sanadHaj) [reply packages/rest/src/router/redirect-route.ts]: fixed

File: packages/rest/src/router/swagger-route.ts (hunk 5 above)
Comment (sanadHaj) [reply packages/rest/src/router/swagger-route.ts]: not relevant after rolling back

File: packages/rest/src/router/swagger-route.ts (hunk 4 above)
Comment (sanadHaj) [reply packages/rest/src/router/swagger-route.ts]: not relevant after rolling back

File: packages/rest/src/router/swagger-route.ts (hunk 3 above)
Comment (sanadHaj) [reply packages/rest/src/router/swagger-route.ts]: not relevant after rolling back

File: packages/rest/test/unit/rest.application/rest.application.unit.ts (hunk 2 above)
Comment (sanadHaj) [reply packages/rest/test/unit/rest.application/rest.application.unit.ts]: fixed

File: packages/rest/test/unit/rest.server/rest.server.redirect.unit.ts (hunk 1 above)
Comment (sanadHaj) [reply packages/rest/test/unit/rest.server/rest.server.redirect.unit.ts]: fixed

**Comment by bajtos** [comment]
Good discussion!

> > I am proposing to modify `app.redirect` to and introduce a new variant that will accept a function instead of a string as the "target" argument.
> 
> how you intend to get the host and protocol , currently we extract them from request .

As I am envisioning this functionality, it will be the responsibility of LB4 framework to obtain the host and protocol from the incoming request _before_ calling user-provided function to build the target path/url. This way we can keep the complex logic for obtaining the right host/protocol/etc. inside the framework, individual redirects don't have to repeat this logic and will avoid accidental errors.

> may we need to take the (base path + traget) as target to bind the route , and when invoking the invokeHandler to redirect we can add host and protocol to the target then redirect to the full url
> (...)
> the user only should provide the api path and the full url should be generated by lb4 .

I think this is not going to work for the redirecting to externally-hosted swagger-ui, where the target is pointing to that external URL and using the location of the LB4 app as a query string parameter.

Different applications will have different needs when it comes to target URLs. For example, when running behind a url-rewriting reverse proxy, it's probably easiest to redirect to a path only (e.g. `Location: /base-path/target`) and let the reverse proxy figure out the actual URL the client should see.

That's why I proposed that LB4 runtime should provide `{protocol, host, basePath, urlPath}` to the handler function and let the handler to decide how exactly it want to construct the redirect target.

> disagree with this line of code the base bath should not be set as part of openapi.json
> to prevent => Not Found http://127.0.0.1:8080/api/openapi.json if baseUrl /api
> cause the openapi.json path under root
> e.g http://127.0.0.1:8080/openapi.json
> for more further detail check this issue #2329 ,

I am of a different opinion, I think the openapi endpoint should honor `basePath` setting. Let's discuss this aspect in #2329 please and keep this pull request focused on implementing a redirect route.

To move this work forward, I am proposing to split this pull request into multiple smaller steps:
1. Implement simple redirect accepting source & target strings only: `app.redirect('/from', '/to')`. I think this is a useful addition on its own. Please open a new PR for that.
2. Once that first PR is landed, let's rebase this PR on top of the new master and resume the discussion about supporting more advanced use cases.

Thoughts?

**Comment by bajtos** [comment]
> To move this work forward, I am proposing to split this pull request into multiple smaller steps:
> 
> 1. Implement simple redirect accepting source & target strings only: `app.redirect('/from', '/to')`. I think this is a useful addition on its own. Please open a new PR for that.
> 2. Once that first PR is landed, let's rebase this PR on top of the new master and resume the discussion about supporting more advanced use cases.

I see that you already took the opposite approach and simplified this pull request to implement only string-based redirect route. That's fine with me too, I am going to review the code changes now.

File: packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts [hunk 10]
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
         .get('/greet/world')
         .expect(200, 'HELLO WORLD');
     });
+
+    it('redirect swagger routes', async () => {
```
bajtos:
```suggestion
    it('redirects to external Swagger UI', async () => {
```

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
...
+      const app = new RestApplication();
+      await whenIMakeRequestTo(app)
+        .get('/swagger-ui')
+        .expect(308);
+      await whenIMakeRequestTo(app)
+        .get('/explorer')
+        .expect(308);
+    });
+
+    it('redirect controller routes', async () => {
```
bajtos: I am confused about this test. Are you trying to verify that redirects take precedence over controller methods? Let's capture the intention in the test name please. For example:

```ts
it('gives precedence to redirect routes over controller methods', async () => {
```

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
...
+          return `hello world`;
+        }
+      }
+      const app = givenAnApplication();
+      const server = await givenAServer(app);
+      server.redirect('/hello', '/hello/world');
+      givenControllerInApp(app, MyController);
+      const response = await whenIMakeRequestTo(server)
+        .get('/hello')
+        .expect(303);
```
bajtos: Please verify the redirect target too (i.e. the value returned in `Location` header).

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
         .get('/greet/world')
         .expect(200, 'HELLO WORLD');
     });
+
+    it('redirect swagger routes', async () => {
+      const app = new RestApplication();
+      await whenIMakeRequestTo(app)
+        .get('/swagger-ui')
+        .expect(308);
```
bajtos: Please verify the redirect target too (i.e. the value returned in `Location` header).

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
...
+      await whenIMakeRequestTo(app)
+        .get('/explorer')
+        .expect(308);
```
bajtos: Please verify the redirect target too (i.e. the value returned in `Location` header).

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -735,6 +735,61 @@ describe('Routing', () => {
...
+      givenControllerInApp(app, MyController);
+      const response = await whenIMakeRequestTo(server)
+        .get('/hello')
+        .expect(303);
+      await whenIMakeRequestTo(server)
+        .get(response.header.location)
+        .expect(200, 'hello world');
+    });
+
+    it('redirect routes based on their specifics', async () => {
```
bajtos: Similarly here. What scenario are you trying to verify in this test? It's not clear to me.

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -145,8 +145,44 @@ describe('RestApplication (integration)', () => {
...
+    givenApplication();
+    restApp.controller(DummyController);
+    restApp.redirect('/page/html', '/html');
+    await restApp.start();
+    client = createRestAppClient(restApp);
+    const response = await client.get('/page/html').expect(303);
+    await client.get(response.header.location).expect(200, 'Hi');
+  });
+
+  it('controls redirect routes', async () => {
```
bajtos: Please use unique test names that describe the scenario being tested: what are the conditions, what is being done, what is the expected result.

I am confused about these two tests. What are you trying to cover and how are these two tests different from the acceptance-level tests added earlier?

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -145,8 +145,44 @@ describe('RestApplication (integration)', () => {
     });
   });
 
+  it('controls redirect routes', async () => {
+    givenApplication();
+    restApp.controller(DummyController);
+    restApp.redirect('/page/html', '/html');
+    await restApp.start();
+    client = createRestAppClient(restApp);
+    const response = await client.get('/page/html').expect(303);
```
bajtos: Please verify the value of `Location` header too.

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -646,7 +646,6 @@ paths:
         cert: fs.readFileSync(certPath),
       },
     });
-    server.handler(dummyRequestHandler);
```
bajtos: This looks suspicious to me. Could you please explain the reasoning for this change?

packages/rest/src/__tests__/unit/rest.application/rest.application.unit.ts
```diff
@@ -50,5 +50,13 @@ describe('RestApplication', () => {
         ERR_NO_MULTI_SERVER,
       );
     });
+
+    it('when attempting to redirect endpoint', () => {
```
bajtos: This test is nested under `describe('throws')`, which is a block of tests verifying conditions where `RestApplication` is expected to throw an error. Your test does not throw any error though, thus it should be moved outside of the "throws" block.

Also please improve the test name to say what's the expected result. For example:

```ts
it('binds a redirect route for each "server.redirect()" call', () => {
```

I also find it suspicious that you are calling `server.redirect` in a test for `RestApplication`. I think you should call `app.redirect` instead and update the test name accordingly.

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
sanadHaj: after rolling back the swagerRedirect this test removed its not relevant .

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
sanadHaj: https://github.com/strongloop/loopback-next/blob/78f70a1b49ac295b77336a1b5eb8f119abfd1815/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts#L781

 make new request to the redirect route based on location  and verifying the result 
its more idealistic.

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
sanadHaj: test removed

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
sanadHaj: https://github.com/strongloop/loopback-next/blob/78f70a1b49ac295b77336a1b5eb8f119abfd1815/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts#L781

 make new request to the redirect route based on location  and verifying the result 
its more idealistic.

packages/rest/src/__tests__/integration/rest.application.integration.ts
sanadHaj: this is clearly code duplication .. tests replaced with custom status code .

packages/rest/src/__tests__/integration/rest.application.integration.ts
sanadHaj: test replaced with custom status code 
similarly here i make new request to the redirect route to check the result 
based on location header no need to check  the header manually 
the request will check that if its fail it means the location is not sett correctly

packages/rest/src/__tests__/integration/rest.server.integration.ts
sanadHaj: this line removed cause the previous PR introduced SwaggerRedirect , 
removing it make this line come back again

packages/rest/src/__tests__/unit/rest.application/rest.application.unit.ts
sanadHaj: my mistake , a new file created to unit test for redirect . 
i think this will be helpful for the next pull request when we gonna test
 the new functionality

packages/rest/src/__tests__/integration/rest.application.integration.ts
sanadHaj: >     * What happens when `basePath` is set - how is the target URL constructed. Must have: server-level test. Nice to have: acceptance-level test against an applicaition.

i think we have enough test that covers all the redirect angel . theirs no need to check the test base path if it is set . in server and acceptance test i make a new request to check location and the location is based on base path . two birds one stone 
lets assume some one delete base path from redirect method all redirect test that depends on base path will fail .

bajtos: > two birds one stone

In my experience, it's better to have more smaller tests where each test is focused on a single scenario. See https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html

bajtos:
Hmm, I am not able to push into your feature branch :( I'll have to open a new PR to get this landed.

bajtos:
Closing in favor of #2512.

PR #2330, opened 2019-02-04
build: move test files to `src/__tests__` in packages

bajtos:
- Source files are compiled from `src/{foo}` to `dist/{foo}`, the same
  pattern is applied to test files too.

- Both TypeScript sources and JavaScript output are stored in the same
  path relative to project root. This makes it much easier to refer
  to test fixtures.

This is a follow-up for 066d52510cda (#2283) and 91a37dca4214 (#2316),  see also #1636.

Out of scope: refactor project (app & extension) templates, I'll open another PR for that.

## Checklist


.nycrc
```diff
@@ -5,10 +5,9 @@
       "examples/*/dist"
     ],
     "exclude": [
-      "packages/*/dist/test/",
       "packages/*/dist/__tests__/",
       "packages/cli/test/",
-      "examples/*/dist/test/",
+      "examples/*/dist/__tests__/",
```
bajtos: I should have made this change in the previous pull request. I think the drop in code coverage is caused by the fact that before, we were including test files in example projects in lines-to-be-covered. Test files usually have 100% coverage, which is skewing the coverage value upwards.

> **coverage/coveralls ** — Coverage decreased (-0.8%) to 90.429%

PR #2347, opened 2019-02-06
feat(repository): delete operation in hasOne relation and example hasOne

### LINKED ISSUE #2233: hasOne relation DELETE, PATCH and PUT CRUD operations are missing
Add support hasOne CRUD Operations DELETE and PATCH

## Initial reference:
https://github.com/strongloop/loopback-next/pull/2147#issuecomment-453047863

## Acceptance Criteria
- It requires a feature enhancement to support DELETE and PATCH CRUD operations in hasOne relation.
- create interface for CRUD methods for DELETE and PATCH hasOne relation
- create test cases
- update docs
- create a follow up issue to demonstrate using multiple relations in a new/existing example

RaphaelDrai:
implemented delete new feature on hasone relation. created an example of one to one application into
examples folder

"fix #2233"

Implemented the DELETE operation in hasOne relation. which is part of the issue #2233.
I have added an example application for the hasOne relation under the folder:
examples/one-2-one
The next step, I will implement the PATCH then the PUT operations new features in the hasOne relation.

Fixes #2233

## Checklist


examples/one-2-one/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@loopback/example-one-2-one",
+  "version": "1.3.1",
+  "description": "Continuation of the hasOne example using relations in LoopBack 4.",
```
bajtos: The description says "Continuation of the hasOne example" - which example do you mean by "hasOne example"?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -39,6 +40,14 @@ export interface HasOneRepository<Target extends Entity> {
     filter?: Pick<Filter<Target>, Exclude<keyof Filter<Target>, 'where'>>,
     options?: Options,
   ): Promise<Target>;
+
+  /**
+   * Delete multiple target model instances
```
bajtos: This does not make sense to me. In HasOne relation, there is always either no target model or one target model. I don't see why would the caller want to delete multiple instances or provide a filter?

IMO, the API should look like this:

```ts
delete(options?: Options): Promise<Count>;
```

I am also not sure whether to return `Count` (`0` = not found;  `1` = deleted), or whether to use `Boolean` (`false` = not found; `true` = deleted). I guess it's good to stay consistent with other `delete` methods and return `Count`.

examples/one-2-one/package.json
RaphaelDrai: Hello @bajtos , thanks you for your comments. I will add shortly the test of the delete operation. Just for information I already completed the PATCH operation as well, and I will commit it on next week. About the one-2-one example: the example implementation is fitting the documentation in LB4 for hasOne (https://loopback.io/doc/en/lb4/hasOne-relation.html) that describes Supplier and Account (by the way same as in LB3). "Todo-List has one" could make confusion with the existing "Todo-List" that implement hasMany relation. In my opinion and from my own experience as loopback new learner, better to have each relation with its own example to ease the learning and fit the documentation.

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: I agree hasOne is always either no target model or one target model. The problem is that actually there is no any check to avoid multiple targets as we had in the past. Adding multiple targets check was removed from the initial version that I tested a coupe of months ago. This is the reason why I kept the Count. Let me know if you want me to add the check that avoid to add multiple targets then I will return true or false as per your recommendation.
Do you agree with a separated example one-2-one application as I did?

examples/one-2-one/package.json
bajtos: Makes sense 👍 

Let's find a new better example name please. How about `@loopback/example-hasOne-relation` or maybe just `@loopback/example-hasOne`?

bajtos: Or maybe `@loopback/example-supplier-account`?

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: > I agree hasOne is always either no target model or one target model. The problem is that actually there is no any check to avoid multiple targets as we had in the past. Adding multiple targets check was removed from the initial version that I tested a coupe of months ago. This is the reason why I kept the Count.

I think we need to decouple the delete API (what parameters is the method accepting) from the value it returns.

Do we both agree that it does not make sense to provide a `filter` or `where` argument where deleting the other side of `hasOne` relation?

As for the return value, let's keep returning `Count` for consistency with other `delete` methods.

examples/one-2-one/.gitignore
```diff
@@ -0,0 +1,64 @@
+# Logs
```
bajtos: We don't keep package-level `.gitignore` files and use a top-level `.gitignore` file in loopback-next monorepo instead. Please remove this file.

examples/one-2-one/.yo-rc.json
```diff
@@ -0,0 +1 @@
+{}
```
bajtos: We don't have `.yo-rc.json` in other example packages, can you please remove this file too?

examples/one-2-one/package.json
bajtos: Let's use a clearer description. How about "An example showing hasOne relation in LoopBack 4"?

examples/one-2-one/README.md
```diff
@@ -0,0 +1,3 @@
+# one2one
```
bajtos: The README needs to be improved before this patch is landed. See our [README guidelines](https://loopback.io/doc/en/contrib/README-guidelines.html).

examples/one-2-one/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "dependencies": {
+    "@loopback/boot": "^1.0.8",
+    "@loopback/context": "^1.4.0",
+    "@loopback/core": "^1.1.3",
+    "@loopback/openapi-v3": "^1.1.5",
+    "@loopback/openapi-v3-types": "^1.0.3",
+    "@loopback/repository": "^1.1.1",
+    "@loopback/rest": "^1.5.1",
+    "@loopback/rest-explorer": "^1.1.4",
+    "@loopback/service-proxy": "^1.0.5",
```
bajtos: I think this dependency is not needed by this example app, can you remove it please?

examples/one-2-one/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "loopback-connector-rest": "^3.1.1"
+  },
+  "devDependencies": {
+    "@loopback/build": "^1.1.0",
+    "@loopback/http-caching-proxy": "^1.0.3",
```
bajtos: Ditto, can we remove this dependency please?

examples/one-2-one/public/index.html
```diff
@@ -0,0 +1,73 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>hasOne relation example between supplier and account</title>
```
bajtos: Please use the same description in `index.html`, `README.md` and `package.json` too. Feel free to use this description instead of what I proposed above, just make sure it's used consistently in all three places.

examples/one-2-one/src/application.ts
```diff
@@ -0,0 +1,42 @@
+import {BootMixin} from '@loopback/boot';
+import {ApplicationConfig} from '@loopback/core';
+import {
+  RestExplorerBindings,
+  RestExplorerComponent,
+} from '@loopback/rest-explorer';
+import {RepositoryMixin} from '@loopback/repository';
+import {RestApplication} from '@loopback/rest';
+import {ServiceMixin} from '@loopback/service-proxy';
```
bajtos: I believe we don't need ServiceMixin for this example, can you remove it please?

examples/one-2-one/src/controllers/account.controller.ts
```diff
@@ -0,0 +1,120 @@
...
+  param,
+  patch,
+  post,
+  put,
+  requestBody,
+} from '@loopback/rest';
+import {Account, Supplier} from '../models';
+import {AccountRepository} from '../repositories';
+
+export class AccountController {
```
bajtos: Do we want to expose accounts directly? Have you considered removing `AccountController` and exposing only `SupplierAccountController` + `SupplierController`? The less files we have in the example, the easier it is to read for our users and the less files we have to maintain ourselves.

examples/one-2-one/src/repositories/supplier.repository.ts
```diff
@@ -0,0 +1,31 @@
...
+  HasOneRepositoryFactory,
+} from '@loopback/repository';
+import {Supplier, Account} from '../models';
+import {AccountRepository} from './account.repository';
+
+export class SupplierRepository extends DefaultCrudRepository<
+  Supplier,
+  typeof Supplier.prototype.id
+> {
+  public readonly accounts: HasOneRepositoryFactory<
```
bajtos: Since the relation is 1-1, each supplier has only one account. Let's rename this property to `account` please.

examples/one-2-one/src/controllers/supplier-account.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+} from '@loopback/rest';
+import {Account} from '../models';
+import {SupplierRepository} from '../repositories';
+
+export class SupplierAccountController {
+  constructor(
+    @repository(SupplierRepository) protected supplierRepo: SupplierRepository,
+  ) {}
+
+  @post('/suppliers/{id}/accounts', {
```
bajtos: Each supplier has a single account only, please fix the REST URL paths in all controller methods to use `/account` (singular form).

examples/one-2-one/src/controllers/supplier-account.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+        description: "Array of Account's belonging to Supplier",
+        content: {
+          'application/json': {
+            schema: {type: 'array', items: {'x-ts-type': Account}},
+          },
+        },
+      },
+    },
+  })
+  async find(
```
bajtos: Let's call this method `get` for consistency with the method name used by the repository.

examples/one-2-one/src/controllers/supplier-account.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+    responses: {
+      '200': {
+        description: 'Supplier.Account DELETE success count',
+        content: {'application/json': {schema: CountSchema}},
+      },
+    },
+  })
+  async delete(
+    @param.path.string('id') id: string,
+    @param.query.object('where', getWhereSchemaFor(Account)) where?: Where,
```
bajtos: As commented earlier, I think HasOne's `delete` should not accept any `where` condition. Let's simplify the REST API too.

examples/one-2-one/test/acceptance/test-helper.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  const client = createRestAppClient(app);
+
+  return {app, client};
+}
+
+export interface AppWithClient {
+  app: One2oneApplication;
+  client: Client;
+}
```
bajtos: Please add acceptance-level tests for each controller you are adding to this example application:
 - `SupplierController`
 - `SupplierAccountController`
 - `AccountController` (if you decide to keep it)

examples/one-2-one/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json.schemastore.org/tsconfig",
+  "extends": "@loopback/build/config/tsconfig.common.json",
+  "include": [
+    "src",
+    "test",
+    "index.ts"
```
bajtos: We have recently reworked our project layout so that test files are kept in `src/__tests__`. Please rebase your pull request on top of the latest master and update your new example app to follow the new convention. See the following pull request for the changes you need to make: https://github.com/strongloop/loopback-next/pull/2316/files.

It basically boils down to:
 - move `test` to `src/__tests__`
 - update `package.json` scripts & files
 - update `tsconfig.build.json`
 - fix paths in `import` statements

I apologize for inconvenience I caused here.

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +96,12 @@ export class DefaultHasOneRepository<
     }
     return found[0];
   }
+
+  async delete(where?: Where<TargetEntity>, options?: Options): Promise<Count> {
```
bajtos: Please add tests to cover this new repository method, see [`packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts)

bajtos:
@RaphaelDrai I just found that our `example/todo-list` already comes with a has-one relation: a TodoList has one Image, see e.g. here:

https://github.com/strongloop/loopback-next/blob/d1d1995615c354e1ebf11062a4bced1b1a57f03c/examples/todo-list/src/repositories/todo-list.repository.ts#L43-L46

Given that we already have an example for HasOne, I am taking back my earlier comments where I said it's ok to create a new example for HasOne-only. Let's keep building on top of the existing example please.

examples/one-2-one/.gitignore
RaphaelDrai: ok

examples/one-2-one/.yo-rc.json
RaphaelDrai: ok

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: ok

examples/one-2-one/README.md
RaphaelDrai: I will remove my hasOne example as per your notification below.

bajtos:
@RaphaelDrai I see a lot of unrelated commits from our `master` in your feature branch, it looks like GitHub thinks your pull request (feature branch) was started from an older commit in `master` and the history somehow diverged. I suspect you have incorrectly rebased your branch. Could you please fix it? See our instructions in [How to rebase your branch](https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/code-contrib.md#how-to-rebase-your-branch).

Since most of the review comments were related to the new example app that's no longer part of this pull request, it may be easier for you to start fresh, create a new feature branch and open a new pull request. You can use `git cherry-pick <sha>` to move the relevant commits from this old branch to the new one. In general, we don't like when an old PR is closed a new one is opened, because it loses discussion history, but I feel this case is different because most of the comments are no longer relevant.

RaphaelDrai:
Hi @bajtos and @jannyHou,
I will try again to to proceed with a new rebase of my branch.
If it will fail again I will create a new PR.
Thanks,
Raphael

RaphaelDrai:
Hi @bajtos,
I created a new PR and committed the implementation.
Reference: https://github.com/strongloop/loopback-next/pull/2414

RaphaelDrai:
This PR was replaced with : https://github.com/strongloop/loopback-next/pull/2414

bajtos:
Closing in favor of #2414

Issue #2188, opened 2018-12-23
Routes with parameters are not working in some cases


ambienthack:
## Description / Steps to reproduce / Feature proposal

Create two routes like this

<pre>
  @patch('/users/{userId}/meals/{id}')
  async updateById(
    @param.path.string('userId') userId: string,
    @param.path.string('id') id: string
  ): Promise<void> {
    //anyting
  }

  @patch('/users/{id}')
  async updateById(
    @param.path.string('id') id: string
  ): Promise<void> {
    //anything
  }

</pre>

Make a request PATCH /users/5c1d78842566764efc8b9605

## Current Behavior

<pre>
{
  "error": {
    "statusCode": 404,
    "name": "NotFoundError",
    "message": "Endpoint \"PATCH /users/5c1d78842566764efc8b9605\" not found."
  }
}
</pre>

If the error doesn't arise, try changing parameter names in path patterns, like this:
`@patch('/users/{id}/meals/{mealId}')`
`@patch('/users/{userId}')`

## Expected Behavior

Controller method should be invoked

I believe there is a bug [here](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/router/trie.ts#L164)

<pre>
    for (const child of children) {
        const result = search(keys, index + 1, params, child.node);
        if (result) {
            Object.assign(params, child.params);
            return result;
        }
    }
</pre>

Should be `if (result && isNodeWithValue(result.node)) {`.

When multiple children are found, if the first child checked is `@patch('/users/{userId}/meals/{id}')`, search() method assumes route is found & doesn't check other routes.

Joel-Raju:
I seem to have run into the same issue. These are my controller actions

#### Action 1


    @get('/restaurants/{id}')
    async getRestaurantById(
    @param.path.number('id') id: number,
      ): Promise<Restaurant> {
        console.log(id);
        return await this.restaurantRepository.findById(id);
      }

#### Action 2

    @get('/restaurants/', {
    async find(
    @param.query.object('filter', getFilterSchemaFor(Restaurant))
    filter?: Filter,
    ): Promise<Restaurant[]> {
      return await this.restaurantRepository.find(filter);
    }

****Current Behavior****
Action 1 throws a 404, but Action 2 works fine

    {
      "error": {
        "statusCode": 404,
        "name": "NotFoundError",
        "message": "Endpoint \"GET /restaurants/1\" not found."
      }
    }

## Expected Behavior
Controller method should be invoked

[...]

PR #2355, opened 2019-02-08
fix(rest): remove trailing slash for routing match

raymondfeng:
Handle trailing slash for urls.

## Checklist

### DIFF
--- a/packages/rest/src/router/router-base.ts
+++ b/packages/rest/src/router/router-base.ts
@@ -7,7 +7,7 @@ import {Request} from '../types';
 import {getPathVariables} from './openapi-path';
 import {createResolvedRoute, ResolvedRoute, RouteEntry} from './route-entry';
 import {compareRoute} from './route-sort';
-import {RestRouter} from './rest-router';
+import {RestRouter, RestRouterOptions} from './rest-router';
 
 /**
  * Base router implementation that only handles path without variables
@@ -16,12 +16,12 @@ export abstract class BaseRouter implements RestRouter {
   /**
    * A map to optimize matching for routes without variables in the path
    */
-  protected routesWithoutPathVars: {[path: string]: RouteEntry} = {};
+  protected routesWithoutPathVars: {[key: string]: RouteEntry} = {};
+
+  constructor(protected options: RestRouterOptions = {strict: false}) {}
 
   protected getKeyForRoute(route: RouteEntry) {
-    const path = route.path.startsWith('/') ? route.path : `/${route.path}`;
-    const verb = route.verb.toLowerCase() || 'get';
-    return `/${verb}${path}`;
+    return this.getKey(route.verb, route.path);
   }
 
   add(route: RouteEntry) {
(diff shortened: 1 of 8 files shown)


raymondfeng:
Please note the explorer redirect test is failing as we now treat `/explorer/` the same as `/explorer`. What's the best practice for trailing `/`? Google seems to have this view - https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html.

What do you think? Should we fix the explorer redirect for `/explorer/` as it's not needed now?

raymondfeng:
I added commit https://github.com/strongloop/loopback-next/pull/2355/commits/402a80db2fea254653c10a8f40b41ac7e9ccf6d7 to make sure all tests are passing.

packages/rest-explorer/src/__tests__/acceptance/rest-explorer.acceptance.ts
```diff
@@ -36,11 +36,12 @@ describe('API Explorer (acceptance)', () => {
         .expect(/<title>LoopBack API Explorer/);
     });
 
-    it('redirects from "/explorer" to "/explorer/"', async () => {
+    it('exposes API Explorer at "/explorer"', async () => {
```
bajtos: This redirect is important because the way how relative paths are resolved in the browser. When the SPA UI is served from `/explorer`, a path like `src="swagger-ui.js"` is resolved to `/swagger-ui.js` instead of `/explorer/swagger-ui.js`.

I don't see how your patch is fixing this problem.

Can you revert the changes in REST Explorer please?

packages/rest/src/router/router-base.ts
```diff
@@ -18,12 +18,29 @@ export abstract class BaseRouter implements RestRouter {
...
 
-  protected getKeyForRoute(route: RouteEntry) {
-    const path = route.path.startsWith('/') ? route.path : `/${route.path}`;
-    const verb = route.verb.toLowerCase() || 'get';
+  /**
+   * Build a key for verb+path
+   * @param verb HTTP verb/method
+   * @param path URL path
+   */
+  private getKey(verb: string, path: string) {
```
bajtos: Since this method is not using anything from the class instance (no `this`), can you move it outside of the class into a regular function?

raymondfeng:
@bajtos Without https://github.com/strongloop/loopback-next/commit/402a80db2fea254653c10a8f40b41ac7e9ccf6d7, CI is failing. See https://travis-ci.org/strongloop/loopback-next/jobs/490404103. 

After we normalizing the key for route matching, `/explorer` matches `/explorer/` and vice versa. As a result, `/explorer` returns `200`.

raymondfeng:
@bajtos I improved `rest-explorer` to honor `explorerPath` in index.html. PTAL.

packages/rest-explorer/src/__tests__/acceptance/rest-explorer.acceptance.ts
raymondfeng: Instead of depending on redirect, I updated `index.html.ejs` to honor `explorerPath` and it works now for `http://localhost:3000/explorer`.

raymondfeng:
@bajtos PTAL

packages/rest-explorer/templates/index.html.ejs
```diff
@@ -5,9 +5,9 @@
 <head>
   <meta charset="UTF-8">
   <title>LoopBack API Explorer</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css">
-  <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
-  <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
+  <link rel="stylesheet" type="text/css" href="<%- explorerPath %>/swagger-ui.css">
```
bajtos: Are you sure this is going to work when the app is running behind a reverse proxy like nginx?

bajtos:
> @bajtos Without https://github.com/strongloop/loopback-next/commit/402a80db2fea254653c10a8f40b41ac7e9ccf6d7, CI is failing. See https://travis-ci.org/strongloop/loopback-next/jobs/490404103. 
>
> After we normalizing the key for route matching, `/explorer` matches `/explorer/` and vice versa. As a result, `/explorer` returns `200`.

That looks like a breaking change to me!

I agree that `/explorer` should match `/explorer/` and vice versa but **only when there is no handler for the other path registered**.

I am concerned that the behavior you are proposing here will be confusing to our users: they register handlers for two different paths, the framework happily accepts then and then only one handler is invoked for both paths.

raymondfeng:
FYI: Express has the following [setting](http://expressjs.com/en/4x/api.html#app.set:
) to control the trialing `/` :

strict routing | Boolean | Enable strict routing. When enabled, the router treats "/foo" and "/foo/" as different. Otherwise, the router treats "/foo" and "/foo/" as the same.NOTE: Sub-apps will inherit the value of this setting. | N/A (undefined)
-- | -- | -- | --

hacksparrow:
I think we should have an option to configure this behavior.

Btw, the the setting is simply `strict` in case of a `Router` - http://expressjs.com/en/4x/api.html#express.router.

raymondfeng:
@bajtos @hacksparrow I extracted the fix for #2188 into a new PR - #2375 and leave the trailing slash handling in this PR, which may require more discussions.

packages/rest-explorer/templates/index.html.ejs
raymondfeng: It's the same deal with or without this change. See https://github.com/strongloop/loopback-next/pull/2293.

raymondfeng:
Superseded by #2419

PR #2356, opened 2019-02-08
feat(repository): created hasManyThrough following convention pattern

codejamninja:
I created some boilerplate for the hasThroughMany relation using the hasMany relation. Currently, this contains the identical functionality as the hasMany relation.

See also #2264

## Checklist

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

Issue #2264, opened 2019-01-17
Model relation: hasManyThrough


dhmlau:
Open a separate issue to track `hasManyThrough` per request from @RaphaelDrai in 
https://github.com/strongloop/loopback-next/issues/2043#issuecomment-448912973

> Hello @bajtos,
It is important for us to have in LB4 the hasAndBelongsToMany (the direct many-to-many relation) that we are using in our LB3 application. We are actually in a transition process to move our LB3 application to LB4 and hasAndBelongsToMany is considered as an high priority.
May I suggest to have a new and separated case for the hasManyThrough in order to avoid confusion with this current case?

RaphaelDrai:
Hi @dhmlau ,
Thanks you for creating this separate issue.
I am adding the design proposal of the SW implementation for review and comment.
[Design_HasAndBelongsToMany_Contrib.docx](https://github.com/strongloop/loopback-next/files/2776434/Design_HasAndBelongsToMany_Contrib.docx)

codejamninja:
As suggested by @bajtos, I'm going to start working on the hasManyThrough before finishing the implementation of hasAndBelongsToMany.

#2308

codejamninja:
I attempted to reimplement hasManyThrough using the hasMany relation with a `through` property. After implementing it, I now think it might not be the best approach. Below are some of the issues I ran into.

The hasManyThrough has to make a database query before creating a constraint. This forces the `createHasManyRepositoryFactory` to return an asynchronous function, which breaks the existing API.

https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-repository.factory.ts#L66-L79 

To keep backwards compatibility, I tried to make the `createHasManyRepositoryFactory` only return a promise if `through` is set. This allows the return type to be the following. I should mention that the following return type does give some errors currently.

```ts
HasManyRepository<Target> | Promise<HasManyRepository<Target>>
```
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-repository.factory.ts#L24

codejamninja:
Another issue I ran into is the generics. hasManyThrough needs access to the through repository. This requires passing the repository type using generics, very similar to the way the target repository type gets passed through the generic.

[...]

Issue #2308, opened 2019-01-30
[Model Relation] hasAndBelongsToMany


dhmlau:
## Description / Steps to reproduce / Feature proposal
Originated from https://github.com/strongloop/loopback-next/issues/2043

This issue is created specific for `hasAndBelongsToMany` because the original issue contains other types of model relation. 

A few comments from #2043 that are related to `hasAndBelongsToMany`.
- @bajtos gave some info and starting point on creating a new relation type: https://github.com/strongloop/loopback-next/issues/2043#issuecomment-441176508
- @RaphaelDrai posted a design doc on `hasAndBelongsToMany`: https://github.com/strongloop/loopback-next/issues/2043#issuecomment-447065436
   - I believe @bajtos suggested to make the design docs in `.md` or in gist for easier review (?)

cc @RaphaelDrai @TomerSalton @codejamninja  - because you've expressed interests to implement this mdoel relation type.  Please let me know if there's anything I can help facilitating.

codejamninja:
@RaphaelDrai I'm assuming you want this code in the `has-and-belongs-to-many.decorator.ts` file, not the `has-and-belongs-to-many.repository.ts`

codejamninja:
I've started work on this. You can see my progress at #2315.

RaphaelDrai:
@codejamninja Hello, yes it is a mistake in my document. The file name in bullet 8 is indeed "has-and-belongs-to-many.decorator.ts" and not "has-and-belongs-to-many.repository.ts"

codejamninja:
One challenge I ran into that I would like some feedback on is where and when to create the junction table schemes. The other relations only use schemas a user defines in models. In this particular relation, we will need schemas created outside of the user defined models.

codejamninja:
I just realized I don't have to worry about creating the junction table with the hasManyThrough relation because the junction table is defined by the user in a model. I'm going to start implementing hasManyThrough before I finish this implementation. Thanks for the suggestion @bajtos.

TomerSalton:
Hey,

I would like to discuss the definition of the default API that will be provided by this relation.

Assuming STUDENTS hasAndBelongsToMany CLASS, my opinion is to provide the following API:

GET ALL THE CLASSES OF A SPECIFIC STUDENT:

HTTP Method | GET
-- | --
URL | http://root/students/{student_id}/classes/
Returns | Array of Classes

CREATE A CLASS AND CONNECT IT TO THE STUDENT

HTTP Method | POST
[...]

Issue #2043, opened 2018-11-16
Many to many relation (hasAndBelongsToMany)


jiwantmg:
## Description / Steps to reproduce / Feature proposal
I want to implement many to many relation in loopback4 but I couldn't see the article or any documentation regarding many to many relation

nm:
+ 1

dhmlau:
Thanks for the feedback!  It seems like there are more users who are looking for the other model relation types than what we have so far.  As you might've known already, support of `hasOne` https://github.com/strongloop/loopback-next/issues/1422 is in progress.  Let me discuss with @raymondfeng and team and see if there's some document to get others started on adding more relation types.

dhmlau:
cc @RaphaelDrai

dhmlau:
Hi all, thanks for your interest in contributing to more relation types!  Would like to see who has a particular interest or need for a certain relation type. Not sure what's the best way to do a poll here, I'm going to add one comment for each relation type below, and if you would like to contribute or need it for your project, please upvote.  Thanks! 

Here are the outstanding relation types:
 HasManyThrough
 HasAndBelongsToMany
 Polymorphic
 EmbedsOne
 EmbedsMany
 ReferencesMany

dhmlau:
HasManyThrough

dhmlau:
2. HasAndBelongsToMany

dhmlau:
3. Polymorphic

dhmlau:
5. EmbedsMany

dhmlau:
6. ReferencesMany

dhmlau:
cc @strongloop/loopback-next @strongloop/loopback-maintainers

koalabi:
HasManyThrough first but interested in other ones too

funkizer:
HasManyThrough, HasAndBelongsToMany. 
Flick-the-switch type of Auto Update would be very nice. One that adds foreign keys automatically (which LB3 never could do), with cascade delete/restrict.

codejamninja:
How can I contribute to the manyToMany ??? I really need it.

dhmlau:
@b-admike @bajtos , what do you think is the best way to let others join forces?  some content around how to get started or how the code are being organized?  Thanks.

bajtos:
I am proposing to keep this issue focused on "many to many" relation, and discuss other relation types in _[EPIC] More Relations (post-GA) #1450_.

[...]

PR #2359, opened 2019-02-09
HasManyThrough using HasMany

codejamninja:
This pull request attempts to set up the hasManyThrough relation using the hasMany relation. It does it using a `through` property that can be provided to the `@hasMany()` decorator.

I chose to setup the hasManyThrough this way in an attempt to stick to the existing loopback hasManyThrough spec.

https://loopback.io/doc/en/lb3/HasManyThrough-relations.html#defining-a-hasmanythrough-relation

```ts
import { Entity, model, hasMany, property } from '@loopback/repository';
import { Patient, Appointment } from '../models';

@model()
export class Physician extends Entity {
  @property({
    type: 'string',
    id: true
  })
  id?: string;

  @hasMany(() => Patient, { through: () => Appointment })
  patients: Patient[];
}
```

See also #2264 
See also #2356

## Checklist


codejamninja:
I've separated the hasMany and hasManyThrough repository factories. Please reference the following two comments.

https://github.com/strongloop/loopback-next/issues/2264#issuecomment-462585511

https://github.com/strongloop/loopback-next/issues/2264#issuecomment-462588068

bajtos:
@codejamninja thank you for the pull request, this is great effort!

I'll need to set aside larger chunk of time to review the changes, hopefully later this week 🤞 

> I've separated the hasMany and hasManyThrough repository factories.

Please add integration-level tests for the new repository factory to https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts

I think we should also copy https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/has-many.relation.acceptance.ts into `has-many-through.relation.acceptance.ts` and write tests to cover the new "through" mode.

codejamninja:
Ok, thanks @bajtos

codejamninja:
I just added integration level tests and acceptance tests. They are all passing for me.

codejamninja:
Anyone wanting to get early access to the hasManyThrough relation (or anyone wanting to test it), you can install it the following way.

```sh
npm install --save @loopback/repository@git+https://git@github.com/codejamninja/loopback-next.git#npm/has-many-through
```

or add the following to your package.json file

```json
{
  "dependencies": {
    "@loopback/repository": "git+https://git@github.com/codejamninja/loopback-next.git#npm/has-many-through"
  }
}
```

codejamninja:
I just added some basic usage docs for hasManyThrough

raymondfeng:
@codejamninja Thank you for the contribution. I'll try to find some time to go over the PR.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -17,21 +15,31 @@ import {
...
   ModelDefinition,
   RelationType,
+  createBelongsToAccessor,
+  createHasManyRepositoryFactory,
+  juggler,
 } from '../../..';
+import {
+  createHasManyThroughRepositoryFactory,
+  HasManyThroughRepository,
+} from '../../../relations';
```
bajtos: Is there any particular reason for importing from `../../../relations` instead of `../../..` as we do for most other types?

```suggestion
} from '../../..';
```

packages/repository/src/__tests__/unit/repositories/has-many-through-repository-factory.unit.ts
```diff
@@ -0,0 +1,152 @@
...
+  DefaultCrudRepository,
+  Entity,
+  HasManyThroughDefinition,
+  ModelDefinition,
+  RelationType,
+  createHasManyThroughRepositoryFactory,
+  juggler,
+} from '../../..';
+
+describe('createHasManyThroughRepositoryFactory', () => {
```
bajtos: I am proposing to add the following two test cases to this test suite:

- `it('rejects relations with missing "through"')`
- `it('rejects relations with "through" that is not a type resolver')`

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+export type HasManyThroughRepositoryFactory<
+  Target extends Entity,
+  Through extends Entity,
+  ForeignKeyType
+> = (fkValue: ForeignKeyType) => HasManyThroughRepository<Target, Through>;
+
+export interface AdvancedConstraint<T extends Entity> {
+  dataObject: DataObject<T>;
+  filter: Filter<T>;
+  where: Where<T>;
```
bajtos: Can you add TSDoc comment to explain the purpose of this interface please?

I find it suspicious that we have both a `filter` and a `where` object for the same type here. A `filter` already contains `where` portion, what is the purpose of the other `where` then? (`constraint.filter.where` vs `constraint.where`)

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+  ForeignKeyType
+>(
+  relationMetadata: HasManyThroughDefinition,
+  targetRepositoryGetter: Getter<EntityCrudRepository<Target, TargetID>>,
+  throughRepositoryGetter: Getter<EntityCrudRepository<Through, ThroughID>>,
+): HasManyThroughRepositoryFactory<Target, Through, ForeignKeyType> {
+  const meta = resolveHasManyThroughMetadata(relationMetadata);
+  debug('Resolved HasMany relation metadata: %o', meta);
+  return function(fkValue?: ForeignKeyType) {
+    async function getAdvancedConstraint(
```
bajtos: The name "advanced" does not tell us much about the intent. How about `getHasManyThroughConstraint`?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+      targetInstance: Target,
+    ): Promise<AdvancedConstraint<Target | Through>> {
+      return createAdvancedConstraint<
```
bajtos: Similarly here, how about `createHasManyThroughConstraint`?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+  targetFkName: string;
+  targetPrimaryKey: string;
+};
+
+async function createAdvancedConstraint<
+  Target extends Entity,
+  Through extends Entity,
+  ThroughID,
+  ForeignKeyType
+>(
```
bajtos: Please specify the return type of this function explicitly.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+    advancedConstraint.where = {
+      or: throughInstances.map((throughInstance: Through) => {
+        return {id: throughInstance[meta.targetFkName as keyof Through]};
+      }),
+    };
+  }
+  advancedConstraint.filter = {
+    where: advancedConstraint.where,
+  };
+  return advancedConstraint as AdvancedConstraint<Target | Through>;
```
bajtos: I find this weird. What model (Target or Through) is this object constraining? How are the consumers supposed to know that?

Maybe the `AdvancedConstraint` type need more template arguments, e.g. `interface AdvancedConstraint<Through, Target> {...}`?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+    );
+    if (!throughInstances.length) {
+      const id =
+        'through constraint ' + JSON.stringify(advancedConstraint.dataObject);
+      throw new EntityNotFoundError(throughRepo.entityClass, id);
+    }
+    advancedConstraint.dataObject = {};
+    advancedConstraint.where = {
+      or: throughInstances.map((throughInstance: Through) => {
+        return {id: throughInstance[meta.targetFkName as keyof Through]};
```
bajtos: Please remove the hard-coded `id` column name and use the name of the primary key column as provided by the Entity. LB4 allows models to use custom PK names.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (targetInstance) {
+    // through constraint
+    advancedConstraint.dataObject[meta.targetFkName] =
+      targetInstance[meta.targetPrimaryKey as keyof Target];
+    advancedConstraint.where = {where: advancedConstraint.dataObject};
+  } else {
+    // target constraint
+    const throughRepo = await throughRepositoryGetter();
+    const throughInstances = await throughRepo.find(
+      constrainFilter(undefined, advancedConstraint.dataObject),
```
bajtos: I am concerned about this design, where we are making SQL queries as part of the step creating the constraint.

In the future, I would like to implement relational queries using SQL JOINs to get better performance. To be able to do that in a backwards-compatible way, we need to keep all database calls inside the relation repository classes.

Can we come up with such design that `HasManyThroughRepository` accepts a set of constraints (one for the "through" model and another for the "target" model) and all SQL queries are executed by the repository only?

I understand that to build the "target" constraint, we need the data returned by the query of the "through" table. The challenge is to come up with a declarative format for the "target" constraint that will tell the HasManyThroughRepository how to build "target" constraint from "through" query results.

For example, we may decide to compose the constraint from the following fields:
 - a `where` condition for the "through" table (e.g. `{customerId: 1}`)
 - an object mapping "through" columns to "target" columns (e.g. `{sellerId: 'id'}`)
 - an optional `where` condition for the "target" table with additional constraints (e.g. `{country: 'us'}` to constraint the repository to US sellers only)

The HasManyThroughRepository can works as follows:

```ts
const throughInstances = await throughRepo.find(...);
const targetWheres = throughInstances.map(inst => {
  const where = {};
  for (const throughProp in mappingProvidedByConstraint) {
    const targetProp = mappingProvidedByConstraint[throughProp];
    where[targetProp] = inst[throughProp];
  }
  return where;
});
const targetFilter = {where: {
  and: [
    {or: targetWheres}, 
    {...whereProvidedByConstraint},
  ],
  // (I am not sure if I got this filter syntax right)
}};
const result = await targetRepo.find(targetFilter);
```

Another benefit of this approach is that we can get rid of constraint getter and pass the directly the constraint object, the same way as we do for other relation types.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,151 @@
...
+   * @param getTargetRepository the getter of the related target model repository instance
+   * @param constraint the key value pair representing foreign key name to constrain
+   * the target repository instance
+   */
+  constructor(
+    public getTargetRepository: Getter<TargetRepository>,
+    public getThroughRepository: Getter<ThroughRepository>,
+    public getAdvancedConstraint: (
+      targetInstance?: TargetEntity,
+    ) => Promise<AdvancedConstraint<TargetEntity | ThroughEntity>>,
```
bajtos: As commented above, I think the constraint type should be parameterized by both "target" and "through" classes, and use a more descriptive name.

```suggestion
    ) => Promise<HasManyThroughConstraint<TargetEntity, ThroughEntity>>,
```

packages/repository/src/__tests__/unit/repositories/has-many-through-repository-factory.unit.ts
codejamninja: Good idea

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: I don't technically need to add the filter and where. I will remove the filter since it is confusing. I built the advanced constraint because the code originally consuming the constraint assumed the constraint would be a `DataObject<Entity>`. The hasManyThrough relation needs a constraint built out using `or` in the `where` filter to prevent the software from making too many database queries.

Basically, the following is how the constraint is built given a sourceId.
  1. The junction table is filtered by records with a sourceId
  2. The results are reduced to a where filter using or (this is why I have to make a custom constraint)
  3. The target table is filtered by records that match the where filter

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: The confusing thing about this constraint is that it is used for multiple queries. First for the junction table query and then for the target query. Maybe we should just separate them and have `junctionConstraint` and `targetConstraint`???

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Refer to the previous comment

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: 👍

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Refer to the previous comment

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: I accidently am using the term junction when I mean through, so `throughConstraint` and `sourceConstraint`

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Ah, yes, another oversight, thanks!

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Interesting idea. I will need to look into this more.

packages/repository/src/relations/has-many/has-many-through.repository.ts
codejamninja: 👍 will do

docs/site/HasManyThrough-relation.md
```diff
@@ -0,0 +1,200 @@
...
+The definition of the `hasManyThrough` relation is inferred by using the `@hasMany`
+decorator with a `through` property. The decorator takes in a function resolving
+the target model class constructor. The `through` property takes in a function
+resolving the through model class constructor.
+
+The decorated property name is used as the relation name and stored as part of
+the source model definition's relation metadata. The property type metadata is
+also preserved as an array of type `Seller` as part of the decoration.
+
+## Configuring a hasMany relation
```
raymondfeng: `Configuring a hasManyThrough relation`?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
raymondfeng: The resolution of hasManyThrough relation for `find` requires the following input:

1. A filter to constrain the source model instances
2. A PK/FK constraint to connect source instances to through instances
3. A PK/FK constraint to connect through instances to target instances
4. A filter to constrain the target model instances

The implementation can be a waterfall of multiple queries or a single join if such models are backed by the same relational database.

TomerSalton:
Hey,
I wanted to ask what is the difference in the functionality (not implementation) between hasManyThrough and hasAndBelongsToMany?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
bajtos: > The implementation can be a waterfall of multiple queries or a single join if such models are backed by the same relational database.

+1

What's important for me: we need to design our API to allow the implementation to choose whether the operation is implemented via multiple queries or a single JOIN. The currently proposed solution does not offer the flexibility needed, because a query is executed while building one of the constraints.

I'd like us to find a solution how to express constraints declaratively and defer any database queries to the actual repository operation. My comment above offers one of possible solutions.

bajtos:
> I wanted to ask what is the difference in the functionality (not implementation) between hasManyThrough and hasAndBelongsToMany?

This is my understanding:

- `hasManyThrough` requires the user to explicitly provide the "through" (junction) model.
- `hasAndBelongsToMany` creates the "through" model automatically for the user.

@raymondfeng is that correct?

TomerSalton:
> > I wanted to ask what is the difference in the functionality (not implementation) between hasManyThrough and hasAndBelongsToMany?
> 
> This is my understanding:
> 
> * `hasManyThrough` requires the user to explicitly provide the "through" (junction) model.
> * `hasAndBelongsToMany` creates the "through" model automatically for the user.
> 
> @raymondfeng is that correct?

You are right in that aspect. What about the API though?

bajtos:
> What about the API though?

I think that's something we need to find out while working on `hasAndBelongsToMany` relation. Do you anticipate any problems if we don't consider `hasAndBelongsToMany` requirements while working on `hasManyThrough`? If not, then let's move discussion about `hasAndBelongsToMany` to a more appropriate place please, to keep this PR focused on `hasManyThrough`.

codejamninja:
I don't anticipate any problems.

bajtos:
@codejamninja what's the status of this pull request? Do you need any additional information or feedback from us? I just want to ensure the progress is not blocked by us.

codejamninja:
@bajtos, thanks for asking. I am not blocked. The holdup is my schedule. I should get back to this sometime towards the end of the week.

codejamninja:
Ooops, I made some git mistakes. I merged in the master branch, but after merging realized a rebase was probably a better way of handling it. Of course, that created a fork in the git history which required a force push. If someone could re-open the pull request, that would be great.

codejamninja:
I noticed in the loopback 3 docs, it uses the name `keyThrough` instead of what I call the `targetFkName`. I think I will update my code to use `keyThrough` as the name.

https://loopback.io/doc/en/lb3/HasManyThrough-relations.html#defining-the-foreign-key-property

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Fixed with e9a38d47

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: https://github.com/strongloop/loopback-next/pull/2359/files/ecbbba2468f702ec3469aa325c368bad7a46c5ae#diff-a35f109a989bf5b2ac1206db18756a0cR99

docs/site/HasManyThrough-relation.md
codejamninja: a48ffb740a7c3b061b896b0b3f2fee90e40e4b4d

codejamninja:
I noticed that lockfiles somehow got into the commit. I just removed them.

bajtos:
> I noticed in the loopback 3 docs, it uses the name `keyThrough` instead of what I call the `targetFkName`. I think I will update my code to use accept `keyThrough`.

+1 to support `keyThrough` from LB3

> I'm going to keep the same naming I have in the logic because I find it hard to keep track of the variables without prepending `target` and `source`.

Makes sense to me 👍

codejamninja:
I did a major refactor to the way I handle the constraint. I think y'all will like this refactor. It certainly makes the logic a lot cleaner. It leverages the original constraint logic and I think it does a better job with separation of concerns. Below is the basic rundown of how it originally worked and how the new refactor works.

There are two challenges with the constraint in the hasManyThrough relation that the `createAdvancedConstraint` solved.
https://github.com/codejamninja/loopback-next/blob/d3c770e080546eabc71d479544363b89898a86d5/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts#L89-L138

1. There are two constraints that are used (one for the through repository and one for the target repository)
2. Some of the constraints (primarily the one for the through repository) are dependant on a database operation.

My original approach had the database call in `createAdvancedConstraint`, the function that created the constraint. This forced the function that created the constraint to be asynchronous.
https://github.com/codejamninja/loopback-next/blob/d3c770e080546eabc71d479544363b89898a86d5/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts#L117-L119
The fact that this function was asynchronous and that it created constraints for two different repositories made it very difficult to read.

The two challenges I mentioned will not go away, but they can certainly be handled in a more straightforward way that tells a better story of what is happening.

To make the first challenge more readable, I separated the functions that created the constraint into two separate functions, `createTargetConstraint` and `createThroughConstraint`.
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts#L98-L113
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts#L115-L133

To make the second challenge more readable, I refactored the functions that create the constraint to accept the entities from the database rather than handling the database operations themselves. This allows the database calls to happen in the repository, which is where they belong. This also allows the function that creates the constraint to be synchronous, which is more consistent with the original implementation of the constraint.

This new implementation requires the database calls to happen explicitly in the repository before creating the constraint, which I believe is a lot more readable and explains what is going on better.
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-through.repository.ts#L99-L103
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-through.repository.ts#L154-L158

So, in summary, I pass along `getTargetConstraint` and `getThroughConstraint` to the repositories constructor. Any time the constraint requires a database operation, the database operation is explicitly executed and the result passed into the function that creates the constraint. The constraint is then built and can be consumed like any constraint used in previous relations.
https://github.com/codejamninja/loopback-next/blob/codejamninja/has-many-through-using-has-many/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts#L86-L87

codejamninja:
You can see the refactored changes in the following commit.

e450bc90915eafca990f7f66fe37d259dc6be090

codejamninja:
I noticed I accitendly pushed a `yarn.lock` file. Is it better to remove it with a force push, or to remove it with a new commit?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: This is irrelevant now with the new constraint refactor. Read the following comment.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: I have separated the constraint creator into `createTargetConstraint` and `createThroughConstraint`. Read the following comment.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Again, this is now irrelevant with the constraint refactor.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: Take a look at my following comment. Perhaps it is best to come back to this later.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

packages/repository/src/relations/has-many/has-many-through.repository.ts
codejamninja: Done :)

Take a look at the following comment.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

bajtos:
> I noticed I accitendly pushed a yarn.lock file. Is it better to remove it with a force push, or to remove it with a new commit?

Either way is fine in a pull request.

bajtos:
> The two challenges I mentioned will not go away, but they can certainly be handled in a more straightforward way that tells a better story of what is happening.

I appreciate the effort you put to clean up this part and move execution of database queries to repositories 👍 

As I mentioned in https://github.com/strongloop/loopback-next/pull/2359#discussion_r257276433, my main concern is about the ability to turn constraints into SQL JOIN statements. 

Could you please outline how do you envision to convert the proposed implementation based on `getTargetConstraint` and `getThroughConstraint` into a single SQL JOIN call? In particular, how can we convert "target" constraint into SQL before any database query was made? I am not asking for a fully working solution, just an outline demonstrating that SQL JOINs are achievable while preserving backwards compatibility.

To drink my own medicine, here is an example how to build SQL JOIN query for the design in proposed in https://github.com/strongloop/loopback-next/pull/2359#discussion_r257276433:

> - a `where` condition for the "through" table (e.g. `{customerId: 1}`)
> - an object mapping "through" columns to "target" columns (e.g. `{sellerId: 'id'}`)
> - an optional `where` condition for the "target" table with additional constraints (e.g. `{country: 'us'}` to constraint the repository to US sellers only)

Using the examples from above, it is possible to create a following SQL JOIN query (assuming source model "Customer", through model "CustomerSeller" and target model "Seller"):

```sql
SELECT target.*
   FROM CustomerSeller as through
   # this ON statement is inferred from relation definition
   # maybe we don't need the mapping object at all?
   LEFT JOIN Seller as target ON through.sellerId = target.id
   # this condition is built from "through" constraint 
   WHERE through.customerId = 1 
   # this condition is built from "target" constraint
   AND target.country = 'us';
```

derdeka:
@bajtos @codejamninja Is this PR still valid? Is this feature parked for some reason?

codejamninja:
@derdeka I've been a little busy lately, but I have full intentions on completing this feature. My goal is to have this pull request accepted before the end of the summer. Of course if I can get it in sooner that would be great too!

I think the biggest "holdup" right now is building a way to query in a single round trip for sql supported databases.

Currently this code can be "monkey-patched" into existing loopback projects. I've written about it at the link below. However there are no guarantees that things won't change.

https://github.com/strongloop/loopback-next/pull/2359#issuecomment-475979639

Please let me know if you have any trouble getting it working or discover any bugs.

derdeka:
@codejamninja
Thanks for your great work on this topic. I patch your PR each time on top of each loopback-next release with [patch-package](https://www.npmjs.com/package/patch-package).

I totally agree to query sql like databases with a single roundtrip, but i would personally prefer faster feature parity with lb3 over the perfect solution in one shot. Is it possible to add the performance improvement later as the public api keeps stable? (afaik lb3 is also doing multiple queries)

codejamninja:
I don't have a problem with that, but I'd like to know what @bajtos thinks.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -236,6 +240,88 @@ export class DefaultCrudRepository<T extends Entity, ID>
     );
   }
 
+  /**
+   * @deprecated
```
bajtos: Let's not add code that's already deprecated at the time of writing. Please remove this deprecated method from the pull request

bajtos:
> I totally agree to query sql like databases with a single roundtrip, but i would personally prefer faster feature parity with lb3 over the perfect solution in one shot. Is it possible to add the performance improvement later as the public api keeps stable? (afaik lb3 is also doing multiple queries)

Let me clarify my previous comments. I am not asking for implementing SQL JOINs in the initial version, that's a lot of work and I agree we should prefer feature parity over perfect solution.

What I am asking for is to design public APIs in such way that will allow us to rework internal implementation details into SQL JOINs with no breaking changes required in the public APIs.

---

I would love to see this this work moving forward. In my experience, long-living pull requests are not very practical because one has to often resolve merge conflicts after rebasing on changes made to `master`.

Let's find a way how to split this pull request into smaller incremental chunks that can be landed independently.

For example, the interface `HasManyThroughDefinition`, possibly together with `resolveHasManyThroughMetadata` and `HasManyThroughResolvedDefinition`, look like good candidates to me.

The interface `HasManyThroughRepository` looks like another good candidate for a standalone pull request.

The next step can be adding `createHasManyThroughRepositoryFactoryFor` with an empty implementation throwing "not implemented" error and the acceptance test suite in `packages/repository/src/__tests__/acceptance/has-many-through.relation.acceptance.ts` - the test should be skipped via `describe.skip` for now, until we add the necessary implementation bits.

With this infrastructure in place, we can focus on the implementation details - `createHasManyThroughRepositoryFactory`, interfaces for defining and resolving contraints, etc. - without being distracted by the large amount of supporting code not really relevant to the implementation bits.

Thoughts?

packages/repository/src/__tests__/unit/repositories/has-many-through-repository-factory.unit.ts
codejamninja: I have added those tests.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
codejamninja: I suggest adding the PK/FK constraints in a separate pull request. I created a feature request for it here #3545.

packages/repository/src/repositories/legacy-juggler-bridge.ts
codejamninja: Sounds great

codejamninja:
I am ready for this pull request to be accepted

codejamninja:
@bajtos this pull request is ready.

For those interested in trying it out while waiting on the pull request getting accepted, you can install it the following way.

```sh
npm install --save @loopback/repository@git+https://git@github.com/codejamninja/loopback-next.git#npm/codejamninja/has-many-through-using-has-many@1.11.0-rc.1
```

or add the following to your package.json file

```json
{
  "dependencies": {
    "@loopback/repository": "git+https://git@github.com/codejamninja/loopback-next.git#npm/codejamninja/has-many-through-using-has-many@1.11.0-rc.1"
  }
}
```

mamiller93:
@codejamninja Just to be clear, this hasManyThrough decorator requires an Entity as your map, correct? 

Customer => Order => Seller, each of them are entities. If you simply have a mapping table in between, say:
Customer <=> Map <=> AccountManager (mapping account managers to customers, with the map not needing a unique ID but the columns themselves create a unique pair), this hasManyThrough would not work?

codejamninja:
> @codejamninja Just to be clear, this hasManyThrough decorator requires an Entity as your map, correct?
> 
> Customer => Order => Seller, each of them are entities. If you simply have a mapping table in between, say:
> Customer <=> Map <=> AccountManager (mapping account managers to customers, with the map not needing a unique ID but the columns themselves create a unique pair), this hasManyThrough would not work?

So you are saying that the Map/Join/Junction table should be allowed to not be an entity?

mamiller93:
> > @codejamninja Just to be clear, this hasManyThrough decorator requires an Entity as your map, correct?
> > Customer => Order => Seller, each of them are entities. If you simply have a mapping table in between, say:
> > Customer <=> Map <=> AccountManager (mapping account managers to customers, with the map not needing a unique ID but the columns themselves create a unique pair), this hasManyThrough would not work?
> 
> So you are saying that the Map/Join/Junction table should be allowed to not be an entity?

Correct - the Mapping table that maps customers to multiple account managers or account managers to multiple customers doesn't need an ID. The pair (account manager, customer) becomes the unique identifier itself. And an 'Entity without an ID' (this beast doesn't exist) is just a Model. So my question is whether your implementation of hasManyThrough would work for Models, not necessarily need Entities?

codejamninja:
@mamiller93 could having the Join table not be an Entity be filed as another issue for another pull request? This pull request is so huge and is getting challenging for me to maintain. I'm constantly having to keep it in sync with the master branch and spend a lot of time rewriting things due to changes in the master.

The current implementation can definitely be improved (for example #3545), but it also works well as it is and the code/api is stable. It should not require a large amount of refactoring to improve upon this implementation.

mamiller93:
@codejamninja Oh for sure! I genuinely was just curious whether this PR would satisfy something I need in my project or whether I'd need to concoct my own implementation :)

codejamninja:
@bajtos any update?

bajtos:
@codejamninja thank you for the updates and sorry for the delay.

This PR is indeed large and difficult to keep in sync in master. Have you considered my proposal to split it into smaller chunks that can be landed faster?

> For example, the interface `HasManyThroughDefinition`, possibly together with `resolveHasManyThroughMetadata` and `HasManyThroughResolvedDefinition`, look like good candidates to me.
>
> The interface `HasManyThroughRepository` looks like another good candidate for a standalone pull request.
>
> The next step can be adding `createHasManyThroughRepositoryFactoryFor` with an empty implementation throwing "not implemented" error and the acceptance test suite in `packages/repository/src/__tests__/acceptance/has-many-through.relation.acceptance.ts` - the test should be skipped via `describe.skip` for now, until we add the necessary implementation bits.
>
> With this infrastructure in place, we can focus on the implementation details - `createHasManyThroughRepositoryFactory`, interfaces for defining and resolving contraints, etc. - without being distracted by the large amount of supporting code not really relevant to the implementation bits.

I am going to review the pull request as it is now, to see if it goes in the right direction at the high level.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,87 @@
...
+      TargetID,
+      EntityCrudRepository<Target, TargetID>,
+      Through,
+      ThroughID,
+      EntityCrudRepository<Through, ThroughID>
+    >(
+      targetRepositoryGetter,
+      throughRepositoryGetter,
+      getTargetContraint,
+      getThroughConstraint,
```
bajtos: Re-starting the thread https://github.com/strongloop/loopback-next/pull/2359#discussion_r257276433 here.

> Take a look at my following comment. Perhaps it is best to come back to this later.
> I suggest adding the PK/FK constraints in a separate pull request. I created a feature request for it here #3545.

My problem with this proposal is that the way how you designed the API, #3545 will require breaking changes and thus a new semver-major release.

My comment https://github.com/strongloop/loopback-next/pull/2359#discussion_r257276433 includes a proposal that will be forward-compatible with SQL-JOIN-based implementation and can be implemented using multiple DB queries too, but I understand you may be reluctant to follow that proposal.

I feel we have to make a trade-off here:
1. We can accept the current proposal and (hopefully) get HasManyThrough relation available sooner, at the future expense of breaking the API & releasing a semver-major version when we get to implement SQL JOINs.
2. Alternatively, we can insist on a forward-compatible API at the cost of delaying the delivery of HasManyThrough relation.

Considering that SQL JOINs are not on our roadmap yet, I think we won't get to implement SQL JOINs in the next 6 months, thus it should be ok to bump the major version when we eventually get to work on SQL JOINs. Unless there is a community contribution for this feature coming sooner.

I'd like to hear from @strongloop/loopback-maintainers, especially @raymondfeng and @dhmlau: what is your opinion?

packages/repository/src/__tests__/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -0,0 +1,247 @@
...
+  RepositoryMixin,
+} from '../..';
+import {Order, Seller} from '../fixtures/models';
+import {
+  CustomerRepository,
+  OrderRepository,
+  SellerRepository,
+} from '../fixtures/repositories';
+
+describe('HasManyThrough relation', () => {
```
bajtos: We are in the process of migrating relation acceptance tests to `packages/repository-tests` so that we can run the tests against different databases (`memory`, `mysql` and `mongodb` ATM, more databases will come in the future) - see https://github.com/strongloop/loopback-next/issues/3442 and https://github.com/strongloop/loopback-next/pull/3538.

Can you please add these new tests to `repository-tests` from the beginning, so that we don't have to move them later?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
dhmlau: Just my 2 cents.

It's tempting to have HasManyThrough relation available sooner, given the number of our users have indicated they want this feature back, see https://github.com/strongloop/loopback-next/issues/2043. I think if we won't need to have a semver-major version released in the next 6 months (at least) because of this, it's reasonable.

Also, @raymondfeng @bajtos, I remember we were once discussing there are some breaking changes that we might need to make too, though I couldn't recall which area or feature. 

Would definitely like to hear opinions from @raymondfeng and @strongloop/loopback-maintainer because I'm not as close to the implementation as the rest. :)

codejamninja:
@bajtos I would prefer to have this entire pull request go in at once because all the work is already done. If you think it's not possible to review a commit this large then I will split it up, but if you can review I can wait a little bit longer.

bajtos:
> I would prefer to have this entire pull request go in at once because all the work is already done. If you think it's not possible to review a commit this large then I will split it up, but if you can review I can wait a little bit longer.

I am able to review it, it's just requires much more effort on my side. As a result, I tend to avoid coming back to the patch and often need several days to build enough energy to do another round of review. 🤷‍♂

codejamninja:
Is there anything on my plate I need to do before this gets merged in?

bajtos:
The file `docs/site/HasManyThrough-relation.md` seems to be empty, could you @codejamninja please take a look?

packages/repository/src/relations/relation.types.ts
```diff
@@ -68,6 +68,42 @@ export interface HasManyDefinition extends RelationDefinitionBase {
   keyTo?: string;
 }
 
+export interface HasManyThroughDefinition extends RelationDefinitionBase {
+  type: RelationType.hasMany;
```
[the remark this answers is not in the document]
bajtos: > I think we should use a new relation type here, e.g. `RelationType.hasAndBelongsToMany`?

☝️ that does not make sense, this PR is adding "has many through", not "has and belongs to many".

Having said that, is it a good idea to keep using `hasMany` type or should we introduce a new relation type `RelationType.hasManyThrough`?

packages/repository/src/relations/relation.types.ts
```diff
@@ -68,6 +68,42 @@ export interface HasManyDefinition extends RelationDefinitionBase {
...
+   */
+  keyTo?: string;
+
+  /**
+   * The foreign key used by the through model to reference the target model.
+   *
+   * E.g. when a Customer has many Order instances and a Seller has many Order instances,
+   * then keyThrough is "sellerId".
+   */
+  keyThrough?: string;
```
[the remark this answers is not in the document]
bajtos: > I find the names `through`, `keyTo` and `keyThrough` a bit difficult to comprehend.
> (...)
> Have you considered using a nested object to define the "through" parameters, for example:
> (...)

☝️ @codejamninja PTAL, the comment above was up-voted but was not addressed yet.

codejamninja:
> The file `docs/site/HasManyThrough-relation.md` seems to be empty, could you @codejamninja please take a look?

Fixed

codejamninja:
> I have reviewed the pull request again. I think the overall direction is good, but I am not confident enough about all implementation details.
> 
> At the same time, I don't want to block progress. I think it will be best to land this PR soon, even if it's not perfect yet, and incrementally improve it in subsequent pull requests. To leave us room to fix possible design mistakes, I would like to mark this new relation type as EXPERIMENTAL. Let's add a warning to the documentation that the underlying implementation of the relation (including public APIs like relation repository) can change in the near future and we won't be releasing a semver-major version even if some of the changes could break backward-compatibility.
> 
> @strongloop/loopback-maintainers thoughts?

I added the experimental warning

packages/repository/src/__tests__/unit/repositories/has-many-through-repository-factory.unit.ts
```diff
@@ -0,0 +1,185 @@
...
+      ),
+    ).to.throw(/target must be a type resolver/);
+  });
+
+  it('rejects relations with a target that is not a type resolver', () => {
+    const relationMeta = givenHasManyThroughDefinition({
+      // tslint:disable-next-line:no-any
+      target: undefined,
+      // the cast to any above is necessary to disable compile check
+      // we want to verify runtime assertion
```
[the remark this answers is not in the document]
codejamninja: I moved the test e75e97c2f72a8e5a347d699e2e601dff764edd84

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,164 @@
...
+  relationMeta: HasManyThroughResolvedDefinition,
+  throughInstances: Through[],
+): DataObject<Target> {
+  const {targetPrimaryKey} = relationMeta;
+  const targetFkName = relationMeta.keyThrough;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const constraint: any = {
+    or: throughInstances.map((throughInstance: Through) => {
+      return {
+        [targetPrimaryKey]: throughInstance[targetFkName as keyof Through],
```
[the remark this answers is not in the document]
codejamninja: I chose to fix with `inq` instead of using `findByForeignKeys` because it keeps the `constraint` logic consistent.

f72b958e98d6b074e437c18b95ee045571545b7b

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,167 @@
...
+   * @param throughModelData - The through model data
+   * @param options - Options for the operation
+   * @param throughOptions - Options passed to create through
+   * @returns A promise which resolves to the newly created target model instance
+   */
+  create(
+    targetModelData: DataObject<Target>,
+    throughModelData?: DataObject<Through>,
+    options?: Options,
+    throughOptions?: Options,
```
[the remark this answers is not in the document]
codejamninja: I added `throughOptions` to all operations

cdefaf776cf9117361d317f637ee3835d62bf972

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
codejamninja: I think if we shouldn't add `RelationType.hasManyThrough`. If it's added, what's the point of hasManyThrough being part of hasMany?

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
codejamninja: Would it be ok to overload through? If you're not trying to specify the through `keyFrom` or `keyTo` I find it awkward to write . . .

```ts
  through: {
    model: () => Appointment
  }
```

and find the following less awkward

```ts
  through: () => Appointment
```

codejamninja:
@bajtos can we prioritize getting the pull request in soon? I am spending way too much time rewriting things just to keep it in sync with the master.

codejamninja:
I see you have added it to the September goal. Thank you :)

raymondfeng:
@codejamninja Can you add me `raymondfeng` as a collaborator to your folked repo? I have rebased the brach against upstream master and squashed all commits into one. It would be nice to push back to your repo to make sure I did it correctly.

codejamninja:
Sure thing. @raymondfeng I just added you as a collaborator.

raymondfeng:
@codejamninja I pushed my result to `rfeng` branch of your repo? Can you have a quick check to make sure the merge is done correctly?

In the future, I suggest that we use the following flow to keep track of upstream to make your life easier:

```
1. git remote add upstream git@github.com:strongloop/loopback-next.git
2. git fetch --all
3. git rebase upstream master
```

codejamninja:
@raymondfeng aside from the fact that all the tests are messed up, it looks good. If someone wouldn't mind fixing the tests that would be great. I have no clue what you are doing with the tests and it looks halfway done, so when I try and move them, I just make things worse.

raymondfeng:
@codejamninja Can you elaborate on the issue with tests? Is my merge missing some tests from your original PR? Or are they merged incorrectly? I think there was a landed PR that moved some tests from `repository` to `repository-tests`.

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/3538

raymondfeng:
@codejamninja I cleaned up the tests. Please verify.

bajtos:
> I am spending way too much time rewriting things just to keep it in sync with the master.

That's the cost of large pull requests. I have repeatedly suggested to split this PR into smaller changes that could be reviewed & landed more quickly, see e.g. https://github.com/strongloop/loopback-next/pull/2359#issuecomment-502097909 and https://github.com/strongloop/loopback-next/pull/2359#issuecomment-524209067 🤷‍♂

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
bajtos: So far, we are using `type` to discriminate between different interfaces inheriting from `RelationDefinitionBase`. See https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions

If we want to keep using `hasMany` value for has-many-through relations, then I am fine with that, but let's capture that in the type system too.

We can either get rid of `HasManyThroughDefinition` completely and move all members to `HasManyDefinition`, or else modify `HasManyThroughDefinition` to inherit from `HasManyDefinition`.

I think we need to fix `@hasMany` decorator too, to ensure it describes  "through" fields too. This is important to allow TypeScript to verify metadata related to "through" part and allow IDEs to offer auto-completion.

It's possible you already modified `@hasMany` decorator, I didn't checked - the patch is too large to navigate around.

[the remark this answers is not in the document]
bajtos: So far, we are using `type` to discriminate between different interfaces inheriting from `RelationDefinitionBase`. See https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions

If we want to keep using `hasMany` value for has-many-through relations, then I am fine with that, but let's capture that in the type system too.

We can either get rid of `HasManyThroughDefinition` completely and move all members to `HasManyDefinition`, or else modify `HasManyThroughDefinition` to inherit from `HasManyDefinition`.

I think we need to fix `@hasMany` decorator too, to ensure it describes  "through" fields too. This is important to allow TypeScript to verify metadata related to "through" part and allow IDEs to offer auto-completion.

It's possible you already modified `@hasMany` decorator, I didn't checked - the patch is too large to navigate around.

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
bajtos: Makes sense. Ideally, I think I'd like to have the following options:

When defining the relation via `@hasMany`, I want the short-cut syntax `through: () => Appointment` but also the full-powered variant `through:  {model: () => Appointment, ... }`.

When the relation metadata is resolved, I want `through: () => Appointment` to be expanded into `{model: () => Appointment}` so that we don't have to deal with the shortcut syntax in all places consuming the metadata.

Perhaps the short syntax can be handled by the decorator, so that it expands `@belongsTo(() => Patient, {through: () => Appointment})` to the longer version?

Thoughts?

packages/repository-tests/src/crud/relations/fixtures/models/order.model.ts
```diff
@@ -38,6 +39,9 @@ export class Order extends Entity {
   @belongsTo(() => Customer)
   customerId: MixedIdType;
 
+  @belongsTo(() => Seller)
+  sellerId: number;
```
bajtos:
```suggestion
  sellerId: MixedIdType;
```

We are using `string` when running on NoSQL databases like MongoDB and Cloudant.

codejamninja:
> > I am spending way too much time rewriting things just to keep it in sync with the master.
> 
> That's the cost of large pull requests. I have repeatedly suggested to split this PR into smaller changes that could be reviewed & landed more quickly, see e.g. [#2359 (comment)](https://github.com/strongloop/loopback-next/pull/2359#issuecomment-502097909) and [#2359 (comment)](https://github.com/strongloop/loopback-next/pull/2359#issuecomment-524209067)

@bajtos you are correct. Either way, it's painful. For now, I will just try and get this pull request in.

raymondfeng:
@codejamninja If you can confirm the `rfeng` branch is clean, please replace this one with the `rfeng` branch for the PR to avoid further efforts for rebasing.

codejamninja:
@raymondfeng ok sure

bajtos:
It's annoying that GitHub is collapsing comment threads that haven't been resolved yet. Here are links to the major discussion points needing resolution:

- https://github.com/strongloop/loopback-next/pull/2359#discussion_r317011708
- https://github.com/strongloop/loopback-next/pull/2359#discussion_r317012754

derdeka:
@codejamninja @bajtos Just another month waiting for this feature. Will this PR ever get merged?

codejamninja:
@derdeka hopefully soon. There are 2 discussion points that need to be resolved. I'll try and get around to it this week.

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
codejamninja: > hasAndBelongsToMany

I thought this will be added in a separate pull request? `hasAndBelongsToMany` is dependant on `hasManyThrough`.

https://github.com/strongloop/loopback-next/issues/2308

packages/repository/src/relations/relation.types.ts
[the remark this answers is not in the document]
codejamninja: > I find the names `through`, `keyTo` and `keyThrough` a bit difficult to comprehend.
> 
> Let's take a look at a "simple" relation type where only two models are involved:
> 
> * One side of the relation is described by `source` (an Entity) and `keyFrom` (a property name)
> * The other side of the relation is described by `target` (an Entity) and `keyTo` (a property name)
> 
> In that light, when defining an N-N relation, I would find it more consistent if `keyTo` refered to a property name in the `target` entity (not in the `through` model).
> 
> Have you considered using a nested object to define the "through" parameters, for example:
> 
> ```ts
> {
>   type: RelationType.hasMany, // or hasAndBelongsToMany?
>   targetsMany: true,
> 
>   source: Patient,
>   keyFrom: 'id',
> 
>   target: () => Doctor,
>   keyTo: 'id',
> 
>   through: {
>     model: () => Appointment,
>     keyFrom: 'patientId',
>     keyTo: 'doctorId',
>   }
> }
> ```
> 
> @raymondfeng @strongloop/loopback-next thoughts?

Can this be tabled as a separate pull request?

codejamninja:
> It's annoying that GitHub is collapsing comment threads that haven't been resolved yet. Here are links to the major discussion points needing resolution:
> 
> * [#2359 (comment)](https://github.com/strongloop/loopback-next/pull/2359#discussion_r317011708)
> * [#2359 (comment)](https://github.com/strongloop/loopback-next/pull/2359#discussion_r317012754)

https://github.com/strongloop/loopback-next/pull/2359\#discussion_r331771823
https://github.com/strongloop/loopback-next/pull/2359\#discussion_r331771896

codejamninja:
@bajtos @raymondfeng the tests on this commit are all messed up because the approach to testing was completely redone. It seems like tests are halfway moved and halfway not. I have no idea where to put tests or how the new and "improved" testing works.

bajtos:
>  It seems like tests are halfway moved and halfway not. I have no idea where to put tests or how the new and "improved" testing works.

Can you be more specific please? E.g. show us which of your existing tests you don't know where to put.

I am sure our developer documentation around repository & relation testing can be improved, what kind of information and in which form would help you?

Maybe you can ignore tests for now and figure out the implementation first? Once we agree on the overall direction & implementation design, then you can treat this pull request as a throw-away prototype and start sending smaller pull request to land this big feature in smaller incremental blocks.

codejamninja:
> > It seems like tests are halfway moved and halfway not. I have no idea where to put tests or how the new and "improved" testing works.
> 
> Can you be more specific please? E.g. show us which of your existing tests you don't know where to put.
> 
> I am sure our developer documentation around repository & relation testing can be improved, what kind of information and in which form would help you?
> 
> Maybe you can ignore tests for now and figure out the implementation first? Once we agree on the overall direction & implementation design, then you can treat this pull request as a throw-away prototype and start sending smaller pull request to land this big feature in smaller incremental blocks.

Ok, I think I'll move in that direction then

iury123:
Hello, @bajtos, I'm working on a loopback 4 project and my database's schema has several many-to-many relations. I haven't found anything on lb4 documentation and I was wondering if there is a good solution to handle many-to-many on lb4. What can you suggest me to solve this issue?

bajtos:
> Hello, @bajtos, I'm working on a loopback 4 project and my database's schema has several many-to-many relations. I haven't found anything on lb4 documentation and I was wondering if there is a good solution to handle many-to-many on lb4. What can you suggest me to solve this issue? How can I include many-to-many related objects in the response body of my @get method.

LB4 does not support many-to-many relations yet, this pull request is trying to implement such feature. There may be workarounds possible, I don't have bandwidth to look for them though. Sorry.

CedBadarou:
@iury123 

From my understanding, you're gonna have to wait, because loopback doesn't handle relations good enough right now. They are working on it though.
As a workaround, I use transactions to make all my queries and return the complete object  my client needs. You can also write the whole sql (or whatever) query to fetch all the data you need.

iury123:
@bajtos @CedBadarou Thank you for answering me. Nonetheless, lb4 is a great framework and I really hope this feature will be implemented asap. Why wasn't it implemented yet since many-to-many is so ordinary?

KalleV:
Hi. Is this PR still being worked on or is someone else picking up the feature implementation?

codejamninja:
This pull request forms a decent start / reference for implementing hasManyThrough using HasMany. However, this will need to be submitted in multiple pull requests. At this point, I am not able to spend the time on this necessary to push it through to completion. This implementation works, but is not compatible with the updated way loopback is writing tests. If someone wants to dig through this code and submit PRs to get this implemented, that would be grate.

For now, I have moved on to using https://www.prisma.io as my database layer. You can see an example of it below.

https://github.com/codejamninja/lb4-prisma/blob/master/example/src/controllers/user.controller.ts

bajtos:
@codejamninja Thank you for letting us know :+1: Would you mind to keep this pull request open for now, to give opportunity to other community members (including us, in the core team) to pick up the work where you have left it?

> For now, I have moved on to using https://www.prisma.io as my database layer. You can see an example of it below.
> 
> https://github.com/codejamninja/lb4-prisma/blob/master/example/src/controllers/user.controller.ts

Nice! It's great to see that LB4's REST API layer is flexible enough to support other ORMs like Prisma under the hood. It was one of our design goals for LB4, it feels good to see it validated in practice.

I would love to learn more about the experience you had with integrating LB4 with Prisma - what works great, where did you encounter friction, etc. Have you considered documenting your journey in a blog post (or a series of posts)?

codejamninja:
@bajtos sure

codejamninja:
@bajtos I created an issue to write observations on using prisma with loopback 4.

https://github.com/strongloop/loopback-next/issues/4195

KalleV:
Thank you for the updates. I’m glad to hear this might get picked up by the core team or community!

Is there a list of missing requirements before this feature can be merged to master and a plan for breaking it up into smaller PRs?

bajtos:
> Is there a list of missing requirements before this feature can be merged to master and a plan for breaking it up into smaller PRs?

Re-posting https://github.com/strongloop/loopback-next/pull/2359#issuecomment-502097909:

For example, the interface `HasManyThroughDefinition`, possibly together with `resolveHasManyThroughMetadata` and `HasManyThroughResolvedDefinition`, look like good candidates to me.

The interface `HasManyThroughRepository` looks like another good candidate for a standalone pull request.

The next step can be adding `createHasManyThroughRepositoryFactoryFor` with an empty implementation throwing "not implemented" error and the acceptance test suite in `packages/repository/src/__tests__/acceptance/has-many-through.relation.acceptance.ts` - the test should be skipped via `describe.skip` for now, until we add the necessary implementation bits.

With this infrastructure in place, we can focus on the implementation details - `createHasManyThroughRepositoryFactory`, interfaces for defining and resolving contraints, etc. - without being distracted by the large amount of supporting code not really relevant to the implementation bits.

koliberr136a1:
Does this bug fixed ?

bajtos:
> When this feature will merge to master ?

At the moment, this work is a community effort. @KalleV was kind enough to contribute the first step - see #4247.

Would you @koliberr136a1 like to join the forces and help @KalleV with the next steps?

koliberr136a1:
> > When this feature will merge to master ?
> 
> At the moment, this work is a community effort. @KalleV was kind enough to contribute the first step - see #4247.
> 
> Would you @koliberr136a1 like to join the forces and help @KalleV with the next steps?

I like a lot, but unfortunately I haven't the required informations about how to implement this feature

MacGyver26:
Hello guys !

Do you know when this feature is going to be available in the framework ? thanks in advance.

dhmlau:
@MacGyver26, I believe https://github.com/strongloop/loopback-next/pull/4438 takes over the implementation of the hasManyThrough relation.

agnes512:
The implementation and documentation are merged, closing the issue.

Issue #2020, opened 2018-11-13
Add Component page under Key Concepts


dhmlau:
## Description / Steps to reproduce / Feature proposal
In https://loopback.io/doc/en/lb4/Crafting-LoopBack-4.html#rebuilding-loopback-experience-on-top-of-the-new-core, the link for `Component` is broken. 

Talked to @raymondfeng , we should have a page for `Component` under concepts.  

Related to https://github.com/strongloop/loopback-next/issues/1908

## Acceptance Criteria
- [ ] Create a `Component` page under `Key concepts`. Describe the functionality and interface of a `Component` class on it.
- [ ] Refactor the `Using components` page by moving relevant content to `Component` page.
- [ ] Link the `Component` page from `Concepts` page.

bajtos:
As a short-term solution to fix the broken link, I am proposing to point readers to https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html for now as part of this story, and create a new story to write `Component` page under `Concepts`.

dhmlau:
Created PR https://github.com/strongloop/loopback-next/pull/2361 to fix the broken link, and updated the Acceptance Criteria.

dhmlau:
@raymondfeng , what content do you have in mind for that `Component` page?  given that we have https://loopback.io/doc/en/lb4/Using-components.html.  Thanks.

dhmlau:
Fixed the broken link through PR https://github.com/strongloop/loopback-next/pull/2361.  Moving this out of `TOB`.

dhmlau:
I forgot that we need to estimate this task first. 
@hacksparrow, could you please edit the acceptance criteria? Thanks!

emonddr:
@dhmlau 
> In https://loopback.io/doc/en/lb4/Crafting-LoopBack-4.html#rebuilding-loopback-experience-on-top-of-the-new-core, the link for Component is broken. 

I don't see the word `Component` on the url you mentioned. Do you mean it is missing somewhere on the left side bar?

dhmlau:
Sorry.. I think I was referring to the section above. Should be here: https://loopback.io/doc/en/lb4/Crafting-LoopBack-4.html#deep-dive-into-loopback-4-extensibility. But I fixed the broken link to point to the `using-component.md`.

dhmlau:
Closing as resolved because PR https://github.com/strongloop/loopback-next/pull/2702 has merged.

PR #2361, opened 2019-02-10
fix(docs): broken link in Crafting in LB

dhmlau:
See https://github.com/strongloop/loopback-next/issues/2020

This PR is a quick fix in the `Crafting in LoopBack` page for the broken link. 
As suggested by @bajtos, we'll create a dedicated Component page as a longer term fix. 
 See https://github.com/strongloop/loopback-next/issues/2020#issuecomment-461822363.

### DIFF
--- a/docs/site/Crafting-LoopBack-4.md
+++ b/docs/site/Crafting-LoopBack-4.md
@@ -473,7 +473,7 @@ There are several key pillars to make extensibility a reality for LoopBack 4.
 - [Context](Context.md), the IoC container to manage services
 - [Dependency injection](Dependency-injection.md) to facilitate composition
 - [Decorators](Decorators.md) to supply metadata using annotations
-- [Component](Component.md) as the packaging model to bundle extensions
+- [Component](Using-components.md) as the packaging model to bundle extensions
 
 Please check out [Extending LoopBack 4](Extending-LoopBack-4.md).
 


docs/site/Crafting-LoopBack-4.md
```diff
@@ -473,7 +473,8 @@ There are several key pillars to make extensibility a reality for LoopBack 4.
 - [Context](Context.md), the IoC container to manage services
 - [Dependency injection](Dependency-injection.md) to facilitate composition
 - [Decorators](Decorators.md) to supply metadata using annotations
-- [Component](Component.md) as the packaging model to bundle extensions
+- [Component](Extending-LoopBack-4.md) as the packaging model to bundle
+  extensions
 
 Please check out [Extending LoopBack 4](Extending-LoopBack-4.md).
```
nabdelgadir: Should we still keep this?

dhmlau: I think we should still keep this bullet, because it belongs to `There are several key pillars to make extensibility a reality for LoopBack 4.` 
An alternative link for `Component` could be https://loopback.io/doc/en/lb4/Using-components.html. then it won't be duplicated what we have after the bullets.

what do you think?

nabdelgadir: Yeah, I think that's better so we can avoid duplication.

PR #2375, opened 2019-02-12
fix(rest): only return matched trie nodes with values

### LINKED ISSUE #2188: Routes with parameters are not working in some cases

raymondfeng:
Fixes #2188 

This is a spin-off from #2355.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/router/routing-table.unit.ts
+++ b/packages/rest/src/__tests__/unit/router/routing-table.unit.ts
@@ -214,6 +214,44 @@ function runTestsWithRouter(router: RestRouter) {
     expect(route.pathParams).to.containEql({userId: '1', format: 'json'});
   });
 
+  it('finds "GET /orders, /orders/{id}, /orders/{orderId}/shipments" endpoints', () => {
+    class TestController {
+      @get('/orders/{id}')
+      async getOrderById(@param.path.number('id') id: number): Promise<object> {
+        return {id};
+      }
+      @get('/orders')
+      async findOrders(): Promise<object[]> {
+        return [];
+      }
+      // A path that overlaps with `/orders/{id}`. Please note a different var
+      // name is used - `{orderId}`
+      @get('/orders/{orderId}/shipments')
+      async getShipmentsForOrder(
+        @param.path.number('orderId') id: number,
+      ): Promise<object> {
+        return [];
+      }
+    }
+
+    const table = givenRoutingTable();
+    const spec = getControllerSpec(TestController);
+    table.registerController(spec, TestController);
+
+    const findAndCheckRoute = (url: string, expectedPath: string) => {
+      let request = givenRequest({
+        method: 'get',
+        url,
+      });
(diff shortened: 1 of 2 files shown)


packages/rest/src/__tests__/unit/router/routing-table.unit.ts
```diff
@@ -214,6 +214,44 @@ function runTestsWithRouter(router: RestRouter) {
     expect(route.pathParams).to.containEql({userId: '1', format: 'json'});
   });
 
+  it('finds "GET /orders/ and GET /orders/{id}" endpoints', () => {
```
bajtos:
```suggestion
  it('finds "GET /orders and GET /orders/{id}" endpoints', () => {
```

The test names should match the code.

packages/rest/src/__tests__/unit/router/routing-table.unit.ts
```diff
@@ -214,6 +214,44 @@ function runTestsWithRouter(router: RestRouter) {
...
+    class TestController {
+      @get('/orders/{id}')
+      async getOrderById(@param.path.number('id') id: number): Promise<object> {
+        return {id};
+      }
+      @get('/orders')
+      async findOrders(): Promise<object[]> {
+        return [];
+      }
+      // A path overlaps with `/order/{id}`. Please note a different var name
```
bajtos:
```suggestion
      // A path overlaps with `/orders/{id}`. Please note a different var name
```

b-admike: nitpick: `A path overlaps` -> `A path that overlaps`

PR #2378, opened 2019-02-12
chore: ignore @types/node upgrade by greenkeeper

raymondfeng:
Add `@types/node` to `ignore` list so that Greenkeeper skips PRs such as #2377. 

## Checklist

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -33,5 +33,6 @@
         "sandbox/example/package.json"
       ]
     }
-  }
+  },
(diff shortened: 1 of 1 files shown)


bajtos:
If we don't want to use the latest version of `@types/nodes`, then I think we should downgrade from 10.x to 8.x. That way we can avoid accidental call of APIs that were added by Node.js 10 and are not available in Node.js 8 yet.

Thoughts?

PR #2379, opened 2019-02-12
[WIP] fix: set minimum to 1 for filter.limit

raymondfeng:
API Explorer proposes the following input for `filter` based on the json schema:

```json
{
  "where": {},
  "fields": {
    "id": true,
    "title": true,
    "desc": true,
    "isComplete": true,
    "remindAtAddress": true,
    "remindAtGeo": true
  },
  "offset": 0,
  "limit": 0,
  "skip": 0,
  "order": [
    "string"
  ]
}
``` 

Please note `limit` and `order` contain invalid values.

This PR changes `limit` to be `>=1`. But the explorer does not honor it. We'll continue to investigate how to instruct the explorer to use examples or other constraints from the OpenAPI spec.

## Checklist

### DIFF
--- a/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
@@ -3,16 +3,16 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {getFilterSchemaFor} from '../..';
 import {Entity, model, property} from '@loopback/repository';
 import {expect} from '@loopback/testlab';
+import {getFilterSchemaFor} from '../..';
 
 describe('filterSchema', () => {
   @model({
     name: 'my-user-model',
   })
   class MyUserModel extends Entity {
-    @property() id: string;
+    @property({id: true}) id: string;
 
     @property() age: number;
   }
@@ -33,7 +33,7 @@ describe('filterSchema', () => {
           },
         },
         offset: {type: 'integer', minimum: 0},
-        limit: {type: 'integer', minimum: 0},
+        limit: {type: 'integer', minimum: 1, examples: [100]},
         skip: {type: 'integer', minimum: 0},
         order: {type: 'array', items: {type: 'string'}},
       },
--- a/packages/repository-json-schema/src/filter-json-schema.ts
+++ b/packages/repository-json-schema/src/filter-json-schema.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Model, model, getModelRelations} from '@loopback/repository';
+import {getModelRelations, Model, model} from '@loopback/repository';
 import {JSONSchema6 as JsonSchema} from 'json-schema';
 
 @model({settings: {strict: false}})
@@ -34,7 +34,8 @@ export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
 
       limit: {
         type: 'integer',
-        minimum: 0,
+        minimum: 1,
+        examples: [100],
       },
 
       skip: {


bajtos:
@raymondfeng have you tried to set the `example` field in `filter` parameter spec? See  https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#fixed-fields-10 I am not sure if swagger-ui supports it though.

I am thinking about something like:

```json5
"parameters": [
  {
    "name": "filter",
      "in": "query",
      "style": "deepObject",
      "explode": true,
      "schema": {
        // ...
      },
      "example": {
        "offset": 0,
        "limit": 10
      }
  }
]

**Comment by raymondfeng** [comment]
Unfortunately, adding `examples` and `default` does not help. I'll search existing issues in swagger-ui to see what we can do.

**Comment by derdeka** [comment]
@raymondfeng
Setting the limit for each request on explorer drives me crazy, thats why i also did some experiments - and i think i found a solution. You need to pass "example" value for "limit".

This is the content of a simplified openapi.json:
```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "Test API",
    "version": "1.0.0"
  },
  "paths": {
    "/users": {
      "get": {
        "x-controller-name": "UserController",
        "x-operation-name": "userFind",
        "tags": [
          "UserController"
        ],
        "responses": {
          "200": {
            "description": "Return value of UserController.userFind"
          }
        },
        "parameters": [
          {
            "name": "filter",
            "in": "query",
            "style": "deepObject",
            "explode": true,
            "schema": {
              "properties": {
                "where": {
                  "type": "object"
                },
                "fields": {
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "boolean"
                    },
                    "email": {
                      "type": "boolean"
                    },
                    "password": {
                      "type": "boolean"
                    },
                    "firstname": {
                      "type": "boolean"
                    },
                    "lastname": {
                      "type": "boolean"
                    }
                  }
                },
                "offset": {
                  "type": "integer",
                  "minimum": 0
                },
                "limit": {
                  "type": "integer",
                  "minimum": 1,
                  "example": 100
                },
                "skip": {
                  "type": "integer",
                  "minimum": 0
                },
                "order": {
                  "type": "array",
                  "items": {
                    "type": "string"
                  }
                }
              },
              "type": "object"
            }
          }
        ],
        "operationId": "UserController.userFind"
      }
    }
  },
  "servers": [
    {
      "url": "/"
    }
  ],
  "components": {
    "securitySchemes": {
    },
    "schemas": {
    }
  }
}
```

Serving this file with `http-server --cors .` and opening http://explorer.loopback.io/?url=http://localhost:8080/openapi.json

This is the result calculated by the explorer:
```json
{
  "where": {},
  "fields": {
    "id": true,
    "email": true,
    "password": true,
    "firstname": true,
    "lastname": true
  },
  "offset": 0,
  "limit": 100,
  "skip": 0,
  "order": [
    "string"
  ]
}
```

Hope this helps to update the PR.

bajtos:
@derdeka thank you for the example. I am not sure how much time @raymondfeng has to work on this pull request, would you perhaps like to open your own pull request to apply the changes you have suggested?

raymondfeng:
> You need to pass "example" value for "limit".

@derdeka I believe I have tried that. It's great now you confirmed the `example` helps.

raymondfeng:
> "limit": {
                  "type": "integer",
                  "minimum": 1,
                  "example": 100
                },

It seems that JSON schema only allows `examples` instead `example`.

derdeka:
@raymondfeng 
`example` is working with ui, `examples` not.

Which JSON schema file are you using for validation? I did not find an official schema on https://github.com/OAI/OpenAPI-Specification/tree/master/schemas, but only this one https://raw.githubusercontent.com/kogosoftwarellc/open-api/master/packages/openapi-schema-validator/resources/openapi-3.0.json which allows only `example`.

raymondfeng:
@derdeka I'm referring to https://github.com/strongloop/loopback-next/blob/master/packages/repository-json-schema/src/filter-json-schema.ts#L7. It's the TypeScript definition that prevents us from setting `example`. Of course, we can hack it by casting it to `any`.

bajtos:
Is it perhaps possible that JSON Schema uses `examples` field, but OpenAPI spec allows `example` only? In that case we need to fix our code converting from JSON Schema to OpenAPI, see here:

https://github.com/strongloop/loopback-next/blob/e61565770b674fc8957e9e317632de71dcd39de5/packages/openapi-v3/src/json-to-schema.ts#L75-L78

bajtos:
Quoting from https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#fixed-fields-20:

> Other than the JSON Schema subset fields, the following fields MAY be used for further schema documentation:
>
> Field Name | Type | Description
> -- | -- | --
> (...)||
> example | Any | A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.

See also https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#model-with-example

> Model with example
>
> ```json
> {
>   "type": "object",
>   "properties": {
>     "id": {
>       "type": "integer",
>       "format": "int64"
>     },
>     "name": {
>       "type": "string"
>     }
>   },
>   "required": [
>     "name"
>   ],
>   "example": {
>     "name": "Puma",
>     "id": 1
>   }
> }
> ```

derdeka:
@raymondfeng With #3911 I've found a way to set limit to the explorer. Can this PR closed or do you want to add something?

derdeka:
@raymondfeng PTAL

PR #2382, opened 2019-02-13
feat(rest): add query parameter validation

### LINKED ISSUE #1573: Validate parameter values against their schema

YaelGit:
fix #1573

## Checklist

- [X] New tests added or existing tests modified to cover all changes
- [X] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/packages/rest/src/parser.ts
+++ b/packages/rest/src/parser.ts
@@ -11,12 +11,13 @@ import {
   SchemasObject,
 } from '@loopback/openapi-v3-types';
 import * as debugFactory from 'debug';
-import {RequestBody, RequestBodyParser} from './body-parsers';
+import {ValueWithSchema, RequestBodyParser} from './body-parsers';
 import {coerceParameter} from './coercion/coerce-parameter';
 import {RestHttpErrors} from './rest-http-error';
 import {ResolvedRoute} from './router';
 import {OperationArgs, PathParameterValues, Request} from './types';
 import {validateRequestBody} from './validation/request-body.validator';
+import {validateRequestQuery} from './validation/request-query.validator';
 const debug = debugFactory('loopback:rest:parser');
 
 /**
@@ -38,11 +39,16 @@ export async function parseOperationArgs(
     operationSpec,
     request,
   );
+  const query = await requestBodyParser.loadRequestBodyIfNeeded(
+    operationSpec,
+    request,
+  );
   return buildOperationArguments(
     operationSpec,
     request,
     pathParams,
     body,
+    query,
     route.schemas,
   );
 }
@@ -51,7 +57,8 @@ function buildOperationArguments(
   operationSpec: OperationObject,
   request: Request,
   pathParams: PathParameterValues,
-  body: RequestBody,
+  body: ValueWithSchema,
+  query: ValueWithSchema,
   globalSchemas: SchemasObject,
 ): OperationArgs {
   let requestBodyIndex: number = -1;
@@ -67,6 +74,12 @@ function buildOperationArguments(
 
   const paramArgs: OperationArgs = [];
 
+  let isQuery = false;
+  let paramName = '';
+  let paramSchema = {};
+  let queryValue = {};
+  let schemasValue = {};
+
   for (const paramSpec of operationSpec.parameters || []) {
     if (isReferenceObject(paramSpec)) {
       // TODO(bajtos) implement $ref parameters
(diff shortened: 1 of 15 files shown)


packages/rest/src/parser.ts
```diff
@@ -38,11 +39,16 @@ export async function parseOperationArgs(
     operationSpec,
     request,
   );
+  const query = await requestBodyParser.loadRequestBodyIfNeeded(
```
jannyHou: I doubt whether we can get the query using `requestBodyParser`, according to https://github.com/strongloop/loopback-next/blob/2e12925f2dcb89d2ef0262873615336446390386/packages/rest/src/body-parsers/body-parser.helpers.ts#L54, it eventually focuses and returns the `body` field in a request.

I also understand that there would be a good amount of logic that could be shared among body parsers and query parsers, I would suggest:
- We start from creating a working `query-parser.json.ts` like https://github.com/strongloop/loopback-next/blob/2e12925f2dcb89d2ef0262873615336446390386/packages/rest/src/body-parsers/body-parser.json.ts
- Refactor the common logic into utils so that body/query/header parser could share
- Implement the query parser and call its functions here.

bajtos:
Thank you @YaelGit for the pull request.

I am afraid you are going in a wrong direction here. Conceptually, I would like our validation logic to be independent from the parameter source (body vs. query vs. headers vs. other sources).

Let's take a look at the following code in our parameter parser:

https://github.com/strongloop/loopback-next/blob/5211e474660a0e90d2d84c29f2a1ffe2b2261451/packages/rest/src/parser.ts#L70-L80

What I would like to see is a solution that simply adds validation step to that algorithm, e.g.

```ts
   const spec = paramSpec as ParameterObject; 
   const rawValue = getParamFromRequest(spec, request, pathParams); 
   const coercedValue = coerceParameter(rawValue, spec); 
   validateParameterValue(coercedValue, spec, globalSchemas);  // <--- ADDED
   paramArgs.push(coercedValue); 
```

The function `validateParameterValue` should invoke `ajv` under the hood, similarly to how `validateRequestBody` works now. We cannot use `validateRequestBody` directly, because it uses "request body" name in many places. I think we should refactor `validateRequestBody` into a more generic function that should accept parameter name instead of assuming "request body".

This is the first step. The next step is to group all validation errors into a single error response. The idea is that when there are multiple parameters with invalid values, the response should describe validation errors for all of them. To keep things simple, I am proposing to leave this usability improvement out of scope of the initial pull request.

YaelGit:
@bajtos - i would only like to verify that this is the exact requirement (described by you above). 
For this request: validation of each parameter separately, there is of course no need to rename 'RequestBody' with `ValueWithSchema` (requested on comment by Raymond Feng in the initial issue #2307) for example. Please assist,

bajtos:
I apologize for a late reply.

> For this request: validation of each parameter separately, there is of course no need to rename 'RequestBody' with ValueWithSchema (requested on comment by Raymond Feng in the initial issue #2307) for example. 

As I understand @raymondfeng's https://github.com/strongloop/loopback-next/pull/2307/files/8e3248c1e2930db1867bb27961cddef8cbbe500d#r252310803, he was asking to use a more descriptive name for the type that's used to validate request query.

So yes, there is no need to rename that type. OTOH, depending on how you implement `validateParameterValue`, you may end up with a need to introduce a new type similar to `ValueWithSchema` proposed in that other pull request.

YaelGit:
Hello @bajtos , if i understood you correctly the requested addition you require is a simple function which validates each parameter - validateRequestParameter(paramName, coercedValue, spec, paramSchema); and the removal of usage of current body validation. Plus a modification of the type requestBody in types.ts into 'ValueWithSchema' name.

bajtos:
Hi @YaelGit, what's the status of this pull request? Are you still keen to work on this improvement?

PR #2387, opened 2019-02-14
feat: allow rest-server to be mounted on a path to express

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/2306#discussion_r252321716

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/integration/express.integration.ts
+++ b/packages/rest/src/__tests__/integration/express.integration.ts
@@ -1,0 +1,89 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Application} from '@loopback/core';
+import {get} from '@loopback/openapi-v3';
+import {
+  Client,
+  createClientForHandler,
+  expect,
+  givenHttpServerConfig,
+} from '@loopback/testlab';
+import * as express from 'express';
+import {RestComponent} from '../../rest.component';
+import {
+  HttpRequestListener,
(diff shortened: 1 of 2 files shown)


greenkeeper.json
```diff
@@ -33,5 +33,6 @@
         "sandbox/example/package.json"
       ]
     }
-  }
+  },
+  "ignore": ["@types/node"]
```
bajtos: > chore: ignore @types/node upgrade by greenkeeper

What is the reasoning for this exclusion? Can you move this commit to a new pull request please? I don't see how it's related to RestServer improvements.

packages/rest/src/__tests__/integration/express.integration.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  function givenControllerClass(
+    // tslint:disable-next-line:no-any
+    ctor: new (...args: any[]) => Object,
+  ) {
+    server.controller(ctor);
+  }
+
+  function givenClient() {
+    const app = express();
```
bajtos: I find this very confusing. The function name says that it's creating a client, but it is creating an Express server too. 

I am proposing to rework the suite setup as follows:

```ts
beforeEach(givenLoopBackApp);
beforeEach(givenExpressServer);
beforeEach(givenClient);
```

greenkeeper.json
raymondfeng: Sorry, I guess I accidentally created this branch from the wrong base.

raymondfeng:
@bajtos PTAL.

packages/rest/src/__tests__/integration/express.integration.ts
raymondfeng: Fixed.

PR #2395, opened 2019-02-14
refactor(context): extract async event notification into new class

bajtos:
Cross-posting 
https://github.com/strongloop/loopback-next/pull/2291#issuecomment-460648648:

> A general observation: the file `context.ts` is almost 700 lines long. I feel that's too much, I read it as a sign that our Context class is doing too much. It would be great to find a way how to move some parts of the functionality into smaller blocks that can live in a different file, and then use composition to leverage those building blocks in Context.

In this pull request, I am extracting most of the code related to asynchronous event notifications into a new class `ContextEventNotifier`.

As a result, the size of `context.ts` went from 816 lines down to 711.

## Checklist

### DIFF
--- a/packages/context/src/context-event-notifier.ts
+++ b/packages/context/src/context-event-notifier.ts
@@ -1,0 +1,189 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {EventEmitter} from 'events';
+import {
+  ContextEventObserver,
+  ContextEventType,
+  Notification,
+  Subscription,
+  ContextObserver,
+} from './context-observer';
+import {Context} from './context';
+import {ContextSubscription} from './context-subscription';
+import {Binding} from './binding';
+
+// FIXME: `@types/p-event` is out of date against `p-event@2.2.0`
+const pEvent = require('p-event');
+
+export class ContextEventNotifier extends EventEmitter {
+  /**
+   * A list of registered context observers. The Set will be created when the
+   * first observer is added.
+   */
+  protected observers: Set<ContextEventObserver> | undefined;
+
+  /**
+   * Internal counter for pending notification events which are yet to be
+   * processed by observers.
+   */
+  private pendingNotifications = 0;
+
+  /**
+   * Queue for background notifications for observers
+   */
+  private notificationQueue: AsyncIterableIterator<Notification> | undefined;
+
+  constructor(
+    // tslint:disable-next-line:no-any
+    private _debug: (...args: any[]) => void,
+  ) {
+    super();
+  }
+
+  /**
+   * Add a context event observer to the context
+   * @param observer Context observer instance or function
+   */
+  subscribe(observer: ContextEventObserver): Subscription {
+    this.observers = this.observers || new Set();
+    this.observers.add(observer);
+    return new ContextSubscription(this, observer);
+  }
+
+  /**
+   * Remove the context event observer from the context
+   * @param observer Context event observer
+   */
+  unsubscribe(observer: ContextEventObserver): boolean {
+    if (!this.observers) return false;
+    return this.observers.delete(observer);
+  }
+
+  /**
+   * Check if an observer is subscribed to this context
+   * @param observer Context observer
+   */
+  isSubscribed(observer: ContextObserver) {
+    if (!this.observers) return false;
+    return this.observers.has(observer);
+  }
+
+  /**
+   * Start a background task to listen on context events and notify observers
+   */
+  start(): Promise<void> {
+    // Set up listeners on `bind` and `unbind` for notifications
+    this.setupNotification('bind', 'unbind');
+
+    // Create an async iterator for the `notification` event as a queue
+    this.notificationQueue = pEvent.iterator(this, 'notification');
+
+    return this.processNotifications();
+  }
+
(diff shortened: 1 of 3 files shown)


packages/context/src/context.ts
```diff
@@ -295,9 +231,13 @@ export class Context extends EventEmitter {
    * notified for relevant events.
    */
   protected async waitUntilPendingNotificationsDone(timeout?: number) {
-    const count = this.pendingNotifications;
-    if (count === 0) return;
-    await pEvent.multiple(this, 'observersNotified', {count, timeout});
+    if (!this.eventNotifier) return;
+    const errorHandler = () => {
+      // no-op, eventNotifier will report this error as a rejected promise
```
bajtos: This is a bit ugly.

1. Context is listening for ContextEventNotifier `error` event and re-emits it on the context object itself.
2. `pEvent.multiple` installs the error handler on the notifier object, not the context
3. Because there is no error event handler installed on the context, it triggers unhandled error
3. Before that error is detected by Mocha, `waitUntilPendingNotificationsDone` return a promise rejected because of a timeout.

Before my refactoring, `pEvent.multiple` installed the error handler on the context and thus no uncaught error was reported.

If we don't like the solution proposed here, then we can keep the entire implementation of `waitUntilPendingNotificationsDone` in the Context class only, make sure that `observersNotified` are re-emitted on `Context`, and expose `ContextEventNotifier#pendingNotifications` as a public read-only property that can be consumed by `Context` to setup `pEvent.multiple`. Having said that, this alternative approach seems even more complicated than what I have now.

Thoughts?

packages/context/src/context.ts
```diff
@@ -190,6 +183,15 @@ export class Context extends EventEmitter {
...
+
+      if (this.eventNotifier) {
+        this._debug('  calling evenNotifier to dispatch the event');
+        this.eventNotifier.emit(event, binding, context);
+      } else {
+        this._debug('  evenNotifier was not setup, discarding the event');
+      }
+
+      // For backwards compatibility
       this.emit(event, binding, context);
```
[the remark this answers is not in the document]
raymondfeng: I would prefer to emit such events from `Context` itself, especially for `bind` and `unbind` as `Context` is the event source.

packages/context/src/context.ts
```diff
@@ -135,21 +119,30 @@ export class Context extends EventEmitter {
...
+    if (this.eventNotifier) return;
 
-    this.addParentEventListener('bind');
-    this.addParentEventListener('unbind');
+    this.eventNotifier = new ContextEventNotifier(this._debug.bind(this));
+    this.eventNotifier.on('error', err => this.handleNotificationError(err));
+
+    // For backwards compatibility
+    this.eventNotifier.on(
+      'observersNotified',
```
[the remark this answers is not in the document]
raymondfeng: We can remove `observersNotified` from `Context`.

packages/context/src/context.ts
raymondfeng: This is a warning sign that composing `ContextEventNotifier` into `Context` as a property might not be the best approach. Have you considered to have `Context` extend from `ContextEventNotifier` or model `ContextEventNotifier` as a mixin?

raymondfeng: What about simply moving `waitUntilPendingNotificationsDone` into `ContextEventNotifier`? We don't really care about context level error events for `waitUntilPendingNotificationsDone`, which is used mostly for test.

raymondfeng: Please note that `eventNotifier` is a `protected` property of `Context`. We should be able to use it as follows in the `TestContext`:

```ts
class TestContext extends Context {
  waitUntilPendingNotificationsDone(timeout?: number) {
    return this.eventNotifier.waitUntilPendingNotificationsDone(timeout);
  }
}
```

packages/context/src/context-event-notifier.ts
```diff
@@ -0,0 +1,189 @@
...
+  private pendingNotifications = 0;
+
+  /**
+   * Queue for background notifications for observers
+   */
+  private notificationQueue: AsyncIterableIterator<Notification> | undefined;
+
+  constructor(
+    // tslint:disable-next-line:no-any
+    private _debug: (...args: any[]) => void,
```
raymondfeng: The constructor is a bit ugly. We can get rid of it as the `Context` object is available for the two usages of this function. For example:
```ts
await this.notifyObservers(eventType, binding, context, observers);
        this.pendingNotifications--;
        debug(
          '[%s] Observers notified for %s of binding %s',
          eventType,
          binding.key,
          context.name,
        );
```

raymondfeng: BTW, it might be desirable to change the constructor to be:
```ts
constructor(private context: Context) {}
```

packages/context/src/context-subscription.ts
```diff
@@ -0,0 +1,27 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {ContextEventObserver, Subscription} from './context-observer';
+
+/**
+ * An implementation of `Subscription` interface for context events
+ */
+export class ContextSubscription implements Subscription {
+  constructor(
+    protected notifier: {unsubscribe(observer: ContextEventObserver): boolean},
```
raymondfeng: Why not `ContextEventNotifier`? It might be overkill to have anonymous interface here for `unsubscribe`.

bajtos:
I am giving up on this refactoring for now, I need to focus on other things.

Issue #1978, opened 2018-11-05
Spike: How to write a LB3-like application in JavaScript


dhmlau:
## Description / Steps to reproduce / Feature proposal
In the spike, 
1. Find out the following, focus on what can be done right now without changing the code too much:
- how to create and boot application class in javascript
- how to register route
- how to customize the sequence without dependency injection
- data access should not depend on LB Repository.  Options: 1/ Juggler + connectors. 2/ use db driver directly (e.g. mysql/mongodb modules) 

2. Identify pain points

### Out of Scope
- Dependency injection
- use of decorators
- use of Repository and service proxy

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

rubys:
Suggestion to make this a bit more concrete: add a `--js` option or the like to the `lb4` command.  So, for example, `lb4 app --js`.  The generated code might be less developer friendly, and there may be a minor loss of feature (e.g. dependency injection), but this (a) eliminates the objection, and (b) enables developers to see the benefits of typescript.

jamilhassanspain:
I agree, I would not mind helping out on this one

hacksparrow:
Created a POC - https://github.com/strongloop/loopback4-example-javascript.

hacksparrow:
Some notes from the spike meeting. It was decided that we focus on usecase scenarios instead of 100% porting LB4 to JS.

Plain JS for SOME parts for the app:

1. Compose LB4 app into existing Express app
2. LB4 controller written in JS
3. Front facing LB4 API for a backend API
4. OpenAPI specs for Express routes - request argument parsing with LB4

Technical implementations to figure out:

1. How to define models, additional metadata for model properties?
2. How to define controllers, how to provide OpenAPI spec for controller methods?
3. How to do dependency injection, what and to provide arguments?

bajtos:
Just a quick summary of a chat I had with @hacksparrow on Slack.

My expectations were not matching the goal of this spike as written in the issue description.

> Find out the following, focus on what can be done right now without changing the code too much:
>
> * how to create and boot application class in javascript
> * how to register route
> * how to customize the sequence without dependency injection
> * data access should not depend on LB Repository.  Options: 1/ Juggler + connectors. 2/ use db driver directly (e.g. mysql/mongodb modules)
>
> ### Out of Scope
> * Dependency injection
> * use of decorators
> * use of Repository and service proxy

[...]

PR #2404, opened 2019-02-15
[DO NOT MERGE] Proposed spikes for JSLB4

hacksparrow:
As suggested by @bajtos - https://github.com/strongloop/loopback-next/issues/1978#issuecomment-463960626.

### DIFF
--- a/jslb4-spikes.md
+++ b/jslb4-spikes.md
@@ -1,0 +1,60 @@
+Here is the proposal for the followup tasks for
+https://github.com/strongloop/loopback-next/issues/1978. Anything needs to be
+clarified or you'd like to be added, do let me know.
+
+**1. Spike: Dependency injection or its alternative in JSLB4**
+
+_Acceptance criteria:_
+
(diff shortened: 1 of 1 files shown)


dhmlau:
Capturing my comments in [here](https://github.com/strongloop/loopback-next/issues/1978#issuecomment-463901205).

My thought process is:
1. Based on what we know about the JS experience, can we create a tutorial to let our users know how it can be done?   It doesn't need to be the perfect DX but at least we can show them a way.  Or we should do some enhancements before the tutorial?
2. If there's some enhancement needs to be done, what would be the MVP?

raymondfeng:
@hacksparrow Why don't we use a github issue or wiki instead of PR to discuss the spike?

hacksparrow:
@raymondfeng @bajtos suggested to create a markdown file and commit it.

bajtos:
The spike has been already finished, this PR is no longer relevant.

PR #2406, opened 2019-02-15
fix: model id type is now boolean|number instead of boolean

marvinirwin:
model id type is now boolean|number instead of boolean

https://github.com/strongloop/loopback-next/pull/2245#discussion_r256881910

## Checklist


raymondfeng:
@marvinirwin At least my intention was to discontinue the usage of `number` for compound keys. A better way is to do so at model class level (to be implemented), such as:
```
@model({id: ['p1', 'p2']})
```
WDYT?

marvinirwin:
@raymondfeng I made this at the request of @bajtos 

see https://github.com/strongloop/loopback-next/pull/2245#discussion_r256881910

bajtos:
>  At least my intention was to discontinue the usage of number for compound keys. A better way is to do so at model class level (to be implemented).

In the pull request #2245, @marvinirwin is implementing `lb4 discovery` command. As I understand the situation, juggler and/or connectors are returning a numeric value for the `id` flag of discovered model properties. See https://github.com/strongloop/loopback-next/pull/2245/files#r255073664

I proposed to fix our type definitions to allow both boolean & number values for the `id` flag, to make this compatible with the underlying juggler implementation.

I understand that you @raymondfeng want to move LB4+ in a different direction. What do you propose as a short-term solution to allow `lb4 discovery` to emit correct TypeScript model files? Are you fine with converting the id value from a number to a boolean, loosing any information about composite primary keys along the way?

This is the currently proposed implementation:

```ts
  // o.id could be 1 or 0, typescript will be angry if it is not a boolean
  if (o.id) {
    o.id = !!o.id;
  }
```

raymondfeng:
I'm fine to allow `number` for now and worry about it when we rewrite juggler.

bajtos:
Landed, thank you @marvinirwin for the contribution 🎉

Issue #2333, opened 2019-02-04
Memory connector should enforce referential integrity


bajtos:
The memory connector should enforce foreign key and uniqueness constraints for HasMany, HasOne and BelongsTo relations.

For example, when creating an order using `/post/{userId}/orders` for a `userId` which is not that of a user, I would've expected some sort of a check that would've ensure that the `userId` exists. The connector should throw an error for an `userId` not in the database. 

See https://github.com/strongloop/loopback-next/issues/1718 for background and https://github.com/strongloop/loopback-next/issues/2332 for the SQL story.

See https://github.com/strongloop/loopback-connector-rest/issues/134 for a `Connector` helper to parse different styles of index/fk definitions into a format that's easy to consume by connectors.

## Acceptance criteria

- [ ] Modify the built-in `memory` connector to enforce unique indexes defined at model and property level, leverage the new helper implemented by https://github.com/strongloop/loopback-connector-rest/issues/134. See #1672  for the description of different ways on how model indexes can be defined. See https://github.com/strongloop/loopback-datasource-juggler/pull/1672 for prior implementation that was not finished.
- [ ] Modify the buit-in `memory` connector to enforce FOREIGN KEY constraints defined at model and property level, leverage the new helper implemented by https://github.com/strongloop/loopback-connector-rest/issues/134. See #1672 for the description of different ways on how model indexes can be defined.
- [ ] Warn the user about index/foreign-key options that are not supported by `memory` connector (see out-of-scope below).

### Out of scope

- "Plain" indexes without uniqueness check
- `onUpdate` and `onDelete` options for foreign keys

bajtos:
The pull request https://github.com/strongloop/loopback-datasource-juggler/pull/1672 contains an early version of UNIQUE index support for our memory connector.

jannyHou:
@strongloop/loopback-next Let's add the acceptance criteria for this story.

bajtos:
I think we need to work on #2606 first and the implement any follow-up stories.

Issue #2332, opened 2019-02-04
Database migration should create SQL constraints for relations


bajtos:
_See https://github.com/strongloop/loopback-next/issues/1718 for background context._

As a user building an LB4 application with HasMany BelongsTo or HasOne relation, I want `lb4 automigrate` to define SQL constraints (foreign key, unique) in such way that referential integrity will be correctly enforced by the database.

## Acceptance criteria

- [ ] Improve the base SQL connector [loopback-connector](https://github.com/strongloop/loopback-connector) to include CONSTRAINT definitions when updating tables.
- [ ] Find out how to test this functionality in automated way in our connectors. It's important to test both autoupdate (incremental changes to existing schema) and automigrate (creating new tables).

Ensure the new tests are executed as part of connector test suite and that they pass on our CI for the following connectors
- [ ] [loopback-connector-mysql](https://github.com/strongloop/loopback-connector-mysql)
- [ ] [loopback-connector-postgresql ](https://github.com/strongloop/loopback-connector-postgresql )
- [ ] [loopback-connector-mssql ](https://github.com/strongloop/loopback-connector-mssql )
- [ ] [loopback-connector-oracle ](https://github.com/strongloop/loopback-connector-oracle )
  
Depending on how much work the following connectors will require, either improve them as part of this story or create follow-up stories:
- [ ] [loopback-connector-db2iseries ](https://github.com/strongloop/loopback-connector-db2iseries )
- [ ] [loopback-connector-dashdb ](https://github.com/strongloop/loopback-connector-dashdb )
- [ ] [loopback-connector-informix ](https://github.com/strongloop/loopback-connector-informix )
- [ ] [loopback-connector-db2z ](https://github.com/strongloop/loopback-connector-db2z )

nabdelgadir:
We'll do `mysql` for this story and create follow-up stories for the remaining ones, as it's quite large.

authenticode:
@bajtos @nabdelgadir I am able to create foreign key constraints by adding `settings.foreignKeys` configuration to a model. but the problem is loopback attempts to create foreign keys along with the table creation. A foreign key could reference a column of a table that is yet to be created. In this case, loopback migration script fails. Foreign key creation must be delayed until tables are all created.

klassicd:
[...]

Issue #2331, opened 2019-02-04
Strong relations with referential integrity


bajtos:
Essentially, we see two levels of relations:

- Weak relations, where the developer primarily wants to navigate related models (think of GraphQL) and referential integrity is either not important or already enforced by the (SQL) database.
- Strong relations, where the referential integrity is guaranteed by the framework together with the database.

IMO, referential integrity must be guaranteed at repository level (or lower). For example, when a test helper invokes repository APIs to seed the database with sample data, then we want the referential integrity to be preserved. If we implemented referential integrity at controller level, then consumers of Repository API would be bypassing these checks and would be able to created corrupted data.

Implementation wise, it's important to implement referential integrity in way that does not create any race conditions. For example, a naive implementation calling `findById` on the source model before creating a new instance of the target model creates a race condition - if another request deletes the source model instance after `findById` received database response but before create command is set, then we end up with corrupted data.

With SQL databases, referential integrity is typically implemented by defining FOREIGN KEY constraints. In the context of LB4, this means that `lb4 migrate` must create such constraints based on the relations configured by the app.

With NoSQL databases, users should use different relation types than `hasMany`, for example `referencesMany` or `embedsMany`.

odykyi:
this works for me 
```

@model({
  settings: {
    foreignKeys: {
      fk_todo_todoListId: {
        name: 'fk_todo_todoListId',
        entity: 'TodoList',
        entityKey: 'id',
        foreignKey: 'todolistid',
      },
    },
  },
})
export class Todo extends Entity {
  //etc.
}

```
https://loopback.io/doc/en/lb4/todo-list-tutorial-sqldb.html#specify-the-foreign-key-constraints-in-todo-model

achrinza:
Just to quote from the other issue:

> Unfortunately this isn’t universal across other connectors as LoopBack doesn’t provide a common interface. For example, the MSSQL connector doesn’t support this natively.

I think this is a good opportunity to revisit this soon as it’ll help streamline the use of LoopBack 4.

Issue #2127, opened 2018-12-06
Relations and referential integrity in NoSQL databases


bajtos:
Our current implementation of model relations (has-many, has-one, belongs-to) is deeply rooted in SQL and based on the assumption that the database take care of referential integrity for us.

Example 1: "Customer has many Order instances" and "Order belongs to Customer". When creating a new Order instance, we expect the database to verify that `Order.customerId` is matching the id value of an existing Customer record. We don't have any _reliable & atomic_ way to do this check at LoopBack side.

Example 2: "Customer has one Credentials instance". When creating a new Credentials instance, we expect the database to verify that there are no other Credentials instances already created for the user. We don't have any _reliable & atomic_ way to do this check at LoopBack side.

SQL databases provide FOREIGN KEY and UNIQUE constraints that work great for this flavor of relations.

The situation becomes more tricky when we try to map this approach to NoSQL databases. Many NoSQL databases do not provide FOREIGN KEY and UNIQUE constraints, this is often a constraint caused by [CAP theorem](https://en.wikipedia.org/wiki/CAP_theorem).

For example, it's not possible to enforce UNIQUE constraint when the model data is stored in multiple physical machines and a network partition occurs (a node performing a write operation is not able to reach other nodes because of networking problems, and thus is cannot verify that the new value is not violating uniqueness constraint for records stored on those nodes).

I think we should rethink the way how we are modelling relations and offer different flavors optimized for different backends.

For example, instead of storing a foreign key in the target model, we can store id of related model(s) in the source model and use optimistic locking scheme to enforce the constraints

```ts
class Customer {
  // hasMany relation
  orderIds: string[];

  // hasOne relation
  credentialsId: string;
}

// Algorithm for creating a new Order

// 1. Check that the customer exists
const customer = await customerRepo.findById(customerId);

// 2. Create the order
const order = await orderRepo.create(orderData);
try {
  // 3. Add the new order to the customer
  customer.orderIds.push(order.id);
  await customerRepo.replace(customer);
} catch (err) {
  if (/* a conflict occurred, e.g. somebody deleted the Customer */) {
    // 4. Roll back on conflict
    await orderRepo.deleteById(order.id);
  }
  throw err;
}
```

We can even store the related models as embedded documents, this should work great for Document databases.

```ts
class Customer {
  // hasMany relation
  orders: Order[];

  // hasOne relation
  credentials: Credentials;
}
```

Related issues & discussions:
[...]

PR #2414, opened 2019-02-17
feat(repository): delete & patch operations for HasOne relation

### LINKED ISSUE #2233: hasOne relation DELETE, PATCH and PUT CRUD operations are missing

RaphaelDrai:
implemented the missing crud operations delete, patch and put operations
in relation hasOne.

"fix #2233"

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
+++ b/packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
@@ -7,14 +7,14 @@ import {Application} from '@loopback/core';
 import {expect, toJSON} from '@loopback/testlab';
 import {
   ApplicationWithRepositories,
+  EntityNotFoundError,
+  Filter,
   juggler,
   repository,
   RepositoryMixin,
-  Filter,
-  EntityNotFoundError,
 } from '../..';
 import {Address} from '../fixtures/models';
-import {CustomerRepository, AddressRepository} from '../fixtures/repositories';
+import {AddressRepository, CustomerRepository} from '../fixtures/repositories';
 
 describe('hasOne relation', () => {
   // Given a Customer and Address models - see definitions at the bottom
@@ -115,6 +115,91 @@ describe('hasOne relation', () => {
     ).to.be.rejectedWith(EntityNotFoundError);
   });
 
+  it('can PATCH hasOne instances', async () => {
+    const address = await controller.createCustomerAddress(existingCustomerId, {
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const patchObject = {city: 'Lyon-Genas'};
+    const arePatched = await controller.patchCustomerAddress(
+      existingCustomerId,
+      patchObject,
+    );
+
+    expect(arePatched).to.deepEqual({count: 1});
+    const patchedData = await addressRepo.findById(address.zipcode);
+    expect(toJSON(patchedData)).to.deepEqual({
+      customerId: existingCustomerId,
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Lyon-Genas',
+      province: 'Rhone',
+    });
+  });
+
+  it('patches the related instance only', async () => {
+    const bob = await customerRepo.create({name: 'Bob'});
+    await customerRepo.address(bob.id).create({city: 'Paris'});
+
+    const alice = await customerRepo.create({name: 'Alice'});
+    await customerRepo.address(alice.id).create({city: 'London'});
+
+    const result = await controller.patchCustomerAddress(alice.id, {
+      city: 'New York',
+    });
+
+    expect(result).to.deepEqual({count: 1});
+
+    const found = await customerRepo.address(bob.id).get();
+    expect(toJSON(found)).to.containDeep({city: 'Paris'});
+  });
+
+  it('throws an error when PATCH tries to change the foreignKey', async () => {
+    try {
+      await expect(
+        controller.patchCustomerAddress(existingCustomerId, {
+          customerId: existingCustomerId + 1,
+        }),
+      ).to.be.rejectedWith(/Property "customerId" cannot be changed!/);
+    } catch (err) {}
+  });
+
+  it('can DELETE hasOne relation instances', async () => {
+    await controller.createCustomerAddress(existingCustomerId, {
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const areDeleted = await controller.deleteCustomerAddress(
+      existingCustomerId,
+    );
+    expect(areDeleted).to.deepEqual({count: 1});
+
+    await expect(
+      controller.findCustomerAddress(existingCustomerId),
+    ).to.be.rejectedWith(EntityNotFoundError);
+  });
+
+  it('deletes the related model instance only', async () => {
+    const bob = await customerRepo.create({name: 'Bob'});
+    await customerRepo.address(bob.id).create({city: 'Paris'});
+
+    const alice = await customerRepo.create({name: 'Alice'});
+    await customerRepo.address(alice.id).create({city: 'London'});
+
+    const result = await controller.deleteCustomerAddress(alice.id);
+
+    expect(result).to.deepEqual({count: 1});
+
+    const found = await addressRepo.find();
+    expect(found).to.have.length(1);
+  });
+
   /*---------------- HELPERS -----------------*/
 
   class CustomerController {
@@ -141,6 +226,18 @@ describe('hasOne relation', () => {
       filter: Filter<Address>,
     ) {
       return await this.customerRepository.address(customerId).get(filter);
+    }
+    async patchCustomerAddress(
+      customerId: number,
+      addressData: Partial<Address>,
+    ) {
+      return await this.customerRepository
+        .address(customerId)
+        .patch(addressData);
+    }
+
+    async deleteCustomerAddress(customerId: number) {
+      return await this.customerRepository.address(customerId).delete();
     }
   }
 
--- a/packages/repository/src/relations/has-one/has-one.repository.ts
+++ b/packages/repository/src/relations/has-one/has-one.repository.ts
@@ -4,15 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {Getter} from '@loopback/context';
-import {DataObject, Options} from '../../common-types';
+import {Count, DataObject, Options} from '../../common-types';
+import {EntityNotFoundError} from '../../errors';
 import {Entity} from '../../model';
 import {Filter} from '../../query';
 import {
   constrainDataObject,
   constrainFilter,
+  constrainWhere,
   EntityCrudRepository,
 } from '../../repositories';
-import {EntityNotFoundError} from '../../errors';
 
 /**
  * CRUD operations for a target repository of a HasMany relation
@@ -39,6 +40,21 @@ export interface HasOneRepository<Target extends Entity> {
     filter?: Pick<Filter<Target>, Exclude<keyof Filter<Target>, 'where'>>,
     options?: Options,
   ): Promise<Target>;
+
+  /**
+   * Delete the related target model instance
+   * @param options
+   * @returns A promise which resolves the deleted target model instances
+   */
+  delete(options?: Options): Promise<Count>;
+
+  /**
+   * Patch the  related target model instance
+   * @param dataObject The target model fields and their new values to patch
+   * @param options
+   * @returns A promise which resolves the patched target model instances
+   */
+  patch(dataObject: DataObject<Target>, options?: Options): Promise<Count>;
 }
 
 export class DefaultHasOneRepository<
@@ -86,5 +102,24 @@ export class DefaultHasOneRepository<
       throw new EntityNotFoundError(targetRepository.entityClass, id);
     }
     return found[0];
+  }
+  async delete(options?: Options): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return targetRepository.deleteAll(
+      constrainWhere({}, this.constraint),
+      options,
+    );
+  }
+
+  async patch(
+    dataObject: DataObject<TargetEntity>,
+    options?: Options,
+  ): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return await targetRepository.updateAll(
+      constrainDataObject(dataObject, this.constraint),
+      constrainWhere({}, this.constraint),
+      options,
+    );
   }
 }
--- a/packages/repository/src/relations/has-one/index.ts
+++ b/packages/repository/src/relations/has-one/index.ts
@@ -4,4 +4,5 @@
 // License text available at https://opensource.org/licenses/MIT
 
 export * from './has-one.decorator';
+export * from './has-one.repository';
 export * from './has-one-repository.factory';


packages/repository/src/model.ts
```diff
@@ -262,6 +272,17 @@ export abstract class Entity extends Model implements Persistable {
...
     return entityOrData[idName];
   }
+  /**
+   * Get the identity value of the primary key(s) for a given entity
+   * instance or entity data object.
+   *
+   * @param entityOrData The data object for which to determine the primary
+   * key(s).
+   */
+  static getPrimaryKey(entityOrData: AnyObject): any {
```
elv1s: Since this function can return 1 or more keys I think `getPrimaryKeys` is more clear. Also instead of `any` shouldn't this be a `string[]`

packages/repository/src/model.ts
```diff
@@ -156,6 +156,16 @@ export class ModelDefinition {
     );
     return idProps;
   }
+
+  /**
+   * Returns the primary key(s) of a given model.
+   * Implemented by Raphael Drai for the purpose of PUT operations
```
elv1s: Probably best to leave names out of the source. I've seen others leave usernames with `TODO` comments

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +112,47 @@ export class DefaultHasOneRepository<
...
+  async put(
+    dataObject: DataObject<TargetEntity>,
+    options?: Options,
+  ): Promise<void> {
+    const targetRepository = await this.getTargetRepository();
+    const entity = targetRepository.entityClass;
+    const primaryKeys = targetRepository.entityClass.getPrimaryKey(entity);
+    let id = (this as AnyObject)[''];
+    if (primaryKeys.length === 1 || primaryKeys.length === 0) {
+      id = Reflect.get(dataObject, primaryKeys[0]);
```
elv1s: if `primaryKeys.length === 0` then `primaryKeys[0]` will be `undefined`

It is probably best to throw an error if this is the case, otherwise `replaceById` is called without `id`

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +112,47 @@ export class DefaultHasOneRepository<
...
+    } else if (primaryKeys.length > 1) {
+      for (let idx = 0; idx < primaryKeys.length; idx++) {
```
elv1s: I didn't see a test case for multiple keys

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
     ).to.be.rejectedWith(EntityNotFoundError);
   });
 
+  it('can PATCH hasOne instances', async () => {
+    existingCustomerId = 11;
```
bajtos: Where is the magic number `11` coming from? Can you use the value set from a beforeEach hook on line 34?

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+    const address = await controller.createCustomerAddress(existingCustomerId, {
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const persisted = await addressRepo.findById(address.zipcode);
+    expect(persisted.toObject()).to.deepEqual(address.toObject());
```
bajtos: Can we remove these two lines please? We have other tests in place to verify that `createCustomerAddress` works as expected. This test should focus on verifying PATCH behavior only.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+
+    const patchObject = {city: 'Lyon-Genas'};
+    const arePached = await controller.patchCustomerAddress(
+      existingCustomerId,
+      patchObject,
+    );
+
+    expect(arePached.count).to.equal(1);
```
bajtos: Perhaps `expect(arePatched).to.deepEqual({count: 1})` may be a better assertion?

For example, if `patch` was was returning a boolean instead of an object with `count` property, then your check will fail with "expected undefined to equal 1", which is not very helpful. My proposed check will report "expected true to deep equal {count: 1}".

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+    existingCustomerId = 11;
+    const address = await controller.createCustomerAddress(existingCustomerId, {
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const persisted = await addressRepo.findById(address.zipcode);
+    expect(persisted.toObject()).to.deepEqual(address.toObject());
```
bajtos: Similarly here, I think these two lines do not belong to this test case.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+
+    const putObject = {
+      city: 'Lyon-Genas',
+      province: 'Rhone-Alpe',
+      zipcode: '69740',
+    };
+    await controller.putCustomerAddress(existingCustomerId, putObject);
+    const putData = await addressRepo.findById(address.zipcode);
```
bajtos: I find the name `putData` little bit confusing. How about `replacedInstance`?

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+    expect(putData.city).to.containEql(address.city);
+    expect(putData.province).to.containEql(address.province);
```
bajtos: I think we should also verify `zipcode` property and probably `customerId` too. 

In general, I don't see much value in the `PUT` (`replace`) endpoint. Because the endpoint is performing a full replace, it requires all data including the instance id (`zipcode` here). Once the client has this data, it can call regular (non-relational) endpoint instead.

I am proposing to remove `PUT` endpoint from this pull request. We can discuss later what are the use cases for `PUT` and what is the best way how to support them.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+      patchObject,
+    );
+
+    expect(arePached.count).to.equal(1);
+    const patchedData = await addressRepo.findById(address.zipcode);
+    expect(patchedData.city).to.containEql(address.city);
+  });
+
+  it('can PUT hasOne instances', async () => {
+    existingCustomerId = 11;
```
bajtos: Similarly here, where is `11` coming from? Can you use the value provided from the beforeEach hook?

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+      zipcode: '69740',
+    };
+    await controller.putCustomerAddress(existingCustomerId, putObject);
+    const putData = await addressRepo.findById(address.zipcode);
+    expect(putData.city).to.containEql(address.city);
+    expect(putData.province).to.containEql(address.province);
+  });
+
+  it('can DELETE hasOne instances', async () => {
+    existingCustomerId = 12;
```
bajtos: Similarly here, where is `12` coming from? Can you use the value provided from the beforeEach hook?

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,70 @@ describe('hasOne relation', () => {
...
+    });
+
+    const persisted = await addressRepo.findById(address.zipcode);
+    expect(persisted.toObject()).to.deepEqual(address.toObject());
+
+    const arePached = await controller.deleteCustomerAddress(
+      existingCustomerId,
+    );
+    expect(arePached.count).to.equal(1);
+  });
```
bajtos: Please add a check to verify that the Address instance was deleted in the database.

packages/repository/src/model.ts
```diff
@@ -156,6 +156,16 @@ export class ModelDefinition {
...
+   * in relations models.
+   *
+   */
+  idNames(): string[] {
+    return this.idProperties();
```
bajtos: I am confused about the purpose of this method. Why cannot you call directly `this.idProperties` instead?

I am also arguing that `PUT` operation does not make much sense for relations and that we should remove it from this pull request. At which point `idNames` is not needed at all.

packages/repository/src/model.ts
bajtos: +1

I am little bit confused about the purpose of this function. Are you trying to get primary key names? We already have `idProperties()` for that, plus I don't see why that requires `entityOrData` argument.

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -39,6 +40,30 @@ export interface HasOneRepository<Target extends Entity> {
...
+   */
+  patch(dataObject: DataObject<Target>, options?: Options): Promise<Count>;
+
+  /**
+   * Put related target model instance
+   * @param dataObject The target model fields and their new values to put
+   * @param options
+   * @returns A promise which resolves the patched target model instances
+   */
+  put(dataObject: DataObject<Target>, options?: Options): Promise<void>;
```
bajtos: As mentioned above, I am proposing to remove `put` from this pull request.

If we decide to keep it: please use the name `replace` for consistency with LB 3.x API. `replace` describes the intent better (replace all properties of a model instance), `put` is too REST/HTTP-specific.

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +112,47 @@ export class DefaultHasOneRepository<
     }
     return found[0];
   }
+
+  async delete(where?: Where<TargetEntity>, options?: Options): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return targetRepository.deleteAll(
+      constrainWhere(where, this.constraint as Where<TargetEntity>),
```
bajtos: Why is the explicit cast to `Where<TargetEntity>` needed here? Can we find a way that would not require it?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +112,47 @@ export class DefaultHasOneRepository<
...
+    );
+  }
+
+  async patch(
+    dataObject: DataObject<TargetEntity>,
+    options?: Options,
+  ): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return await targetRepository.updateAll(
+      dataObject,
```
bajtos: I think we need to call `constrainDataObject` here to prevent clients from overwriting properties they are not allowed to change (e.g. foreign key, etc.).

Please add a test to verify, see this HasMany test for inspiration: https://github.com/strongloop/loopback-next/blob/fed9956d760426766190b9c53cf7aaeed12d29f1/packages/repository/src/__tests__/acceptance/has-many.relation.acceptance.ts#L105-L112

packages/repository/src/relations/has-one/index.ts
```diff
@@ -5,3 +5,4 @@
 
 export * from './has-one.decorator';
 export * from './has-one-repository.factory';
+export * from './has-one.repository';
```
bajtos: Let's keep the order of exports consistent with other places please:
 1. decorator
 2. repository
 3. factory

https://github.com/strongloop/loopback-next/blob/fed9956d760426766190b9c53cf7aaeed12d29f1/packages/repository/src/relations/has-many/index.ts#L6-L8

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok, no problems

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok, no problem.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok, no problem.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: As per your proposition, I removed from the PUT operation test; it will be implemented in a later phase.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok, no problem. I removed the PUT from the acceptance test.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok.

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok, done will be in the next commit.

packages/repository/src/model.ts
RaphaelDrai: isNames()  method should exist for the PUT operation due to the abstract class "export abstract class Entity" defined in packages/repository/src/model.ts that invokes this method in line 272, see below the copy/past.
But as per your proposing, I am removing it and it will be handled later when PUT will be defined. I will open a new PR for that purpose.

267   static getIdOf(entityOrData: AnyObject): any {
268      if (typeof entityOrData.getId === 'function') {
269        return entityOrData.getId();
270      }
271
272      const idName = this.definition.idName();
273      return entityOrData[idName];
274   }

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Part of my answer is above with the constrains to implement isNames() method due to calling existing instruction "this.definition.idName()" implemented into the abstract class.
But, as per your proposing, I am removing the PUT operation and these questions will be handled  in the new PR.

packages/repository/src/model.ts
RaphaelDrai: HI, I am removing the PUT operation that will be handled in a separated PR.

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Same answer as above :-)

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Same answer as above related to PUT (removed and will be handled in a new PR)

packages/repository/src/relations/has-one/index.ts
RaphaelDrai: Ok

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,61 @@ describe('hasOne relation', () => {
...
+
+    const patchObject = {city: 'Lyon-Genas'};
+    const arePatched = await controller.patchCustomerAddress(
+      existingCustomerId,
+      patchObject,
+    );
+
+    expect(arePatched).to.deepEqual({count: 1});
+    const patchedData = await addressRepo.findById(address.zipcode);
+    expect(patchedData.city).to.containEql(address.city);
```
bajtos: Since `city` is a string, we should use regular equality comparison.

Also I believe we should be expecting the new (patched) value, not the original?

```suggestion
    expect(patchedData.city).to.equal('Lyon-Genas');
```

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,61 @@ describe('hasOne relation', () => {
...
+  });
+
+  it('can throws error when query tries to change the foreignKey', async () => {
```
bajtos:
```suggestion
  it('throws an error when PATCH tries to change the foreignKey', async () => {
```

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,61 @@ describe('hasOne relation', () => {
...
+    //    existingCustomerId = 12;
+    const address = await controller.createCustomerAddress(existingCustomerId, {
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const newData = await addressRepo.findById(address.zipcode);
+    expect(newData.city).to.containEql(address.city);
```
bajtos: There is no need to verify that `createCustomerAddress` works correctly, we have already verified that in other tests. Please remove these two lines.

packages/repository/src/model.ts
bajtos: Please revert this block, we don't want to add a method that's commented out.

If you want to preserve this code for your future work, then you can create a new `git` branch:

```
# Create a new branch
$ git checkout -b hasOneCRUD-backup
$ git checkout hasOneCRUD
# Remove the commented out code now
$ git commit
```

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -39,6 +40,22 @@ export interface HasOneRepository<Target extends Entity> {
     filter?: Pick<Filter<Target>, Exclude<keyof Filter<Target>, 'where'>>,
     options?: Options,
   ): Promise<Target>;
+
+  /**
+   * Delete multiple target model instances
```
bajtos: I find this confusing. In `HasMany` relation, there should be only **one** related model instance, right?

```suggestion
   * Delete the related target model instance
```

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -39,6 +40,22 @@ export interface HasOneRepository<Target extends Entity> {
...
+   * @param where Instances within the where scope are deleted
+   * @param options
+   * @returns A promise which resolves the deleted target model instances
+   */
+  delete(where?: Where<Target>, options?: Options): Promise<Count>;
```
bajtos: Since there is only one (at most one) target model instance to be deleted, I feel the `where` argument does not make sense. Let's remove it.

```suggestion
  delete(options?: Options): Promise<Count>;
```

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -39,6 +40,22 @@ export interface HasOneRepository<Target extends Entity> {
...
+
+  /**
+   * Patch related target model instance
```
bajtos:
```suggestion
   * Patch the related target model instance
```

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: > Why is the explicit cast to `Where<TargetEntity>` needed here? Can we find a way that would not require it?

☝️

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: > I think we need to call `constrainDataObject` here to prevent clients from overwriting properties they are not allowed to change (e.g. foreign key, etc.).
>
> Please add a test to verify.

☝️

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,61 @@ describe('hasOne relation', () => {
     ).to.be.rejectedWith(EntityNotFoundError);
   });
 
+  it('can PATCH hasOne instances', async () => {
+    //    existingCustomerId = 11;
```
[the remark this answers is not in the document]
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Actually, the hasOne relation allows to create two targets for a same source. This is what I wrote in my previous comments. In my opinion, delete operation shall remove all targets associated to a same source. When we will fix this issue to avoid to create two targets for a same source (new PR) then we can remove the Where as you recommend.
What do you think, can we keep the where till the fix?

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Ok

bajtos:
> Actually, the hasOne relation allows to create two targets for a same source. This is what I wrote in my previous comments. In my opinion, delete operation shall remove all targets associated to a same source. When we will fix this issue to avoid to create two targets for a same source (new PR) then we can remove the Where as you recommend.
What do you think, can we keep the where till the fix?

We were discussing referential integrity for relations in https://github.com/strongloop/loopback-next/issues/2127, with two conclusions:
 - There are two kinds of relations: weak and strong. Weak relations are for navigating between models only (think of GraphQL). Strong relations enforce referential integrity.
- It's too difficult (if possible at all) to enforce referential integrity at LoopBack side. Our relations require the database to provide integrity guarantees. To support different programming models (especially in NoSQL databases), we want to investigate different relation types (HasMany vs. ReferencesMany, etc.). See https://github.com/strongloop/loopback-next/issues/2331 and the related issue (if you don't have ZenHub extension, then you can view the Epic stories [here](https://github.com/strongloop/loopback-next#workspaces/loopback-54ebffa9e8323184150c2125/boards?epics=78452015_2331&filterLogic=any&useDefaultFilterLogic=false&repos=78452015)). I think the following issues may be related for this pull request: 

  - _Memory connector should enforce referential integrity #2333_ 
  - _Database migration should create SQL constraints for relations #2332_
  - _Document hasMany/belongsTo/hasOne limitations for NoSQL databases #2340_

Now if you are using HasOne with PostgreSQL (or any other SQL database), then we require you to configure FOREIGN KEY and UNIQUE constraints to enforce HasOne constraints.

> In my opinion, delete operation shall remove all targets associated to a same source. 

I guess this makes sense. However, I don't understand how is the `where` parameter need for that? How do you envision its usage? IMO, we should simple delete all target models belonging to the source model instance. 

> When we will fix this issue to avoid to create two targets for a same source (new PR) then we can remove the Where as you recommend.

This won't be really fixed in LoopBack. We want to improve our auto-migration to automatically define SQL constraints as needed by relations (see #2332), and also improve our memory connector to enforce such constraints too (see #2333), but I am not sure if that can be considered as a fix?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +104,23 @@ export class DefaultHasOneRepository<
...
+  }
+
+  async patch(
+    dataObject: DataObject<TargetEntity>,
+    options?: Options,
+  ): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return await targetRepository.updateAll(
+      constrainDataObject(dataObject, this.constraint),
+      constrainWhere({}, this.constraint as Where<TargetEntity>),
```
bajtos: Similarly here, why is the explicit cast to `Where<TargetEntity>` needed? Can we find a way that would not require it?

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
```diff
@@ -115,6 +115,53 @@ describe('hasOne relation', () => {
...
+      street: '1 Amedee Bonnet',
+      zipcode: '69740',
+      city: 'Genas',
+      province: 'Rhone',
+    });
+
+    const areDeleted = await controller.deleteCustomerAddress(
+      existingCustomerId,
+    );
+    expect(areDeleted.count).to.equal(1);
```
bajtos:
```suggestion
    expect(areDeleted).to.deepEqual({count: 1});
```

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Ok

packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts
RaphaelDrai: Ok

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -3,17 +3,17 @@
...
 // License text available at https://opensource.org/licenses/MIT
 
-import {Getter} from '@loopback/context';
-import {DataObject, Options, Count} from '../../common-types';
-import {Entity} from '../../model';
-import {Filter, Where} from '../../query';
+import { Getter } from '@loopback/context';
+import { DataObject, Options, Count } from '../../common-types';
+import { Entity } from '../../model';
+import { Filter, Where } from '../../query';
```
bajtos: We don't add spaces around braces, please run `npm run lint:fix`.

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: @RaphaelDrai PTAL, can we remove this explicit type cast here?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -87,4 +103,19 @@ export class DefaultHasOneRepository<
     }
     return found[0];
   }
+  async delete(options?: Options): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    return targetRepository.deleteAll(options);
```
bajtos: This will delete ALL target model instances, not only the instance associated with the source model instance.

Please add a test to verify that `delete` preserves instances that are not related to the source model and then fix the implementation.

I think this is something we should verify for the PATCH operation too. Please add another test to check that PATCH is updating only the single instance on the other end of the relation. 

Yet another two tests to consider: what happens when there is no target model instance? IMO, both `delete` and `patch` should run successfully and return `{count: 0}`.

packages/repository/src/relations/has-one/has-one.repository.ts
bajtos: Do we need to call `constrainWhere` at all? I think we can use `this.constraint` directly.

```suggestion
      this.constraint,
```

packages/repository/src/relations/has-one/has-one.repository.ts
RaphaelDrai: Hi,
I did what you requested to remove the where argument etc... Note that I implemented hasOne same as hasMany is implemented and in hasMany there is no any test to verify the associated target model, the relation already handle it. Please see how hasMany is implement. If your assumption is right the same issue is existing in hasMany. But, when I made tests in hasOne with my initial code including PUT, hasOne worked as expected.
Just to keep you informed, someone else will continue the updates of hasOne with you.
I am moving to other responsibilities.
Kind regards
Raphael

bajtos: I am afraid we misunderstood each other. I was asking to remove `where` argument from the public API, not from the underlying implementation...

> Just to keep you informed, someone else will continue the updates of hasOne with you.
I am moving to other responsibilities.

Thank you for the update, I wish you good luck on your new assignment 🍀

bajtos:
To speed up landing of this pull request, I made the remaining changes myself in 1e199e8.

@raymondfeng @b-admike LGTY?

bajtos:
Landed, thank you for the contribution!

PR #2423, opened 2019-02-18
feat(rest): allow assigning express settings

### LINKED ISSUE #2420: How to configure qs
I understand that LB is using `qs` under the hood to parse query string. Is there a way, through binding or other mean, to configure qs as it only parses object for a depth of 5. The rest is kept stingified (see https://github.com/ljharb/qs#parsing-objects) ?

ludohenin:
Add a new rest configuration for `queryParser`

Fixes #2420 

## Checklist

### DIFF
--- a/docs/site/Server.md
+++ b/docs/site/Server.md
@@ -213,6 +213,25 @@ export async function main() {
 For a complete list of CORS options, see
 https://github.com/expressjs/cors#configuration-options.
 
+### Express settings
+
+Override the default express settings and/or assign your own settings:
+
+```ts
+const app = new RestApplication({
+  rest: {
+    expressSettings: {
+      'x-powered-by': false,
+      env: 'production',
+      ...
+    },
+  },
+});
+```
+
+Checkout `express` [documentation](http://expressjs.com/fr/api.html#app.set) for
+more details about the build-in settings.
+
 ### Configure the Base Path
 
 Sometime it's desirable to expose REST endpoints using a base path, such as
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -214,7 +214,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
   protected _setupRequestHandlerIfNeeded() {
     if (this._expressApp) return;
     this._expressApp = express();
-    this._expressApp.set('query parser', 'extended');
+    this._applyExpressSettings();
     this._requestHandler = this._expressApp;
 
     // Allow CORS support for all endpoints so that users
@@ -243,6 +243,16 @@ export class RestServer extends Context implements Server, HttpServerLike {
         this._onUnhandledError(req, res, err);
       },
     );
+  }
+
+  /**
+   * Apply express settings.
+   */
+  protected _applyExpressSettings() {
+    const settings = this.config.expressSettings || {};
+    for (const key in settings) {
+      this._expressApp.set(key, settings[key]);
+    }
   }
 
   /**
@@ -954,6 +964,8 @@ export interface RestServerOptions {
   apiExplorer?: ApiExplorerOptions;
   requestBodyParser?: RequestBodyParserOptions;
   sequence?: Constructor<SequenceHandler>;
+  // tslint:disable-next-line:no-any
+  expressSettings?: {[name: string]: any};
 }
 
 /**
--- a/packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
+++ b/packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
@@ -3,12 +3,18 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {expect} from '@loopback/testlab';
-import {anOperationSpec} from '@loopback/openapi-spec-builder';
 import {Binding, Context} from '@loopback/context';
 import {Application} from '@loopback/core';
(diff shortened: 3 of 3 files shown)


ludohenin:
@raymondfeng not sure what to test here. I've made this PR to start the discussion.

packages/rest/src/rest.server.ts
```diff
@@ -937,6 +946,7 @@ export interface RestServerOptions {
   openApiSpec?: OpenApiSpecOptions;
   apiExplorer?: ApiExplorerOptions;
   sequence?: Constructor<SequenceHandler>;
+  queryParser?: qs.IParseOptions;
```
raymondfeng: I suggest that we extend it for all express related settings - see http://expressjs.com/en/4x/api.html#app.set.

```ts
expressSettings?: {[name: string]: any};
```

ludohenin: Sound good. question though:

1. do you want to set the query parser custom function from the config?
1. do we really need support for `jsonp`, `views`, `view engine` and `view cache`?

raymondfeng: > 1. do you want to set the query parser custom function from the config?

IMO, it should be possible.

> 2. do we really need support for `jsonp`, `views`, `view engine` and `view cache`?

We can skip such options if they don't make sense to LB4. The other option is to take them as-is if they don't do any harm.

ludohenin: Second try ... I think it's actually fine to leave it up to the app developper to pass whatever he/she wants as settings. Is it more inline with what you had in mind ?

*note: will update the doc once you're ok with the implementation*

packages/rest/src/rest.server.ts
```diff
@@ -937,6 +951,7 @@ export interface RestServerOptions {
   openApiSpec?: OpenApiSpecOptions;
   apiExplorer?: ApiExplorerOptions;
   sequence?: Constructor<SequenceHandler>;
+  expressSettings?: {[name: string]: any};
```
raymondfeng: Please add `// tslint:disable-next-line:no-any` above `expressSettings`.

packages/rest/src/rest.server.ts
```diff
@@ -234,6 +234,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
   }
 
+  /**
+   * Apply express settings. Bind functions to the server context (not express).
+   */
+  protected _applyExpressSettings() {
+    Object.entries(this.config.expressSettings || {}).forEach(
+      ([key, value]) => {
+        if (typeof value === 'function') {
+          value = value.bind(this);
```
raymondfeng: Why `bind` here? I think it's the responsibility of the code that provides the configuration to supply the correct function.

docs/site/Server.md
```diff
@@ -211,6 +211,38 @@ export async function main() {
...
+### Configure query parser
+
+`qs` is used to parse query string with the default package settings.
+
+You can override the defaults by setting your own configuration:
+
+```ts
+// Rest server config
+rest: {
+  queryParser: {...},
```
raymondfeng: Please fix the option here to be `expressSettings`.

raymondfeng:
To unit test the `expressSettings`, you can create a subclass of `RestServer`, which can then access `_expressApp` and use `_expressApp.get(...)` to verify.

docs/site/Server.md
ludohenin: done

packages/rest/src/rest.server.ts
ludohenin: not sure, I was thinking it's better than having express bound to this. Removed it.

packages/rest/src/rest.server.ts
ludohenin: done

ludohenin:
@raymondfeng should be closer now

raymondfeng:
@ludohenin Please run `npm run lint:fix` to fix style/formatting [issues reported by CI](https://travis-ci.org/strongloop/loopback-next/builds/495203052?utm_source=github_status&utm_medium=notification).

ludohenin:
@raymondfeng done but there's still an error with node 10

raymondfeng:
@ludohenin Thanks. The CI failure is not related to your change.

raymondfeng:
@ludohenin I rebased your branch against latest master. Can you grant me the permission per https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ so that I can push such changes to the PR?

packages/rest/src/rest.server.ts
raymondfeng: BTW, we need to keep the default `'query parser': 'extended'` if not present. I fixed it in my clone (to be pushed back to the PR)

ludohenin:
@raymondfeng should be good now

packages/rest/src/rest.server.ts
```diff
@@ -214,7 +214,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
   protected _setupRequestHandlerIfNeeded() {
     if (this._expressApp) return;
     this._expressApp = express();
-    this._expressApp.set('query parser', 'extended');
+    this._applyExpressSettings();
```
bajtos: If I am reading these lines correctly, then:

- before the change, we were explicitly configuring the Express instance to use `extended` query parser
- now we use the default `query parser` configuration as provided by Express

I see that Express uses `extended` query parser by default, so I guess this is ok.

https://expressjs.com/en/4x/api.html#app.settings.table

Issue #1995, opened 2018-11-08
Running unit tests inside docker


ido567:
I have a fresh installation of Loopback V4 which I try to run inside a docker.
To access the `ping` endpoint, I had to set the `HOST` ENV variable to `0.0.0.0`.
My problem is that after the change, I get an error when I run `npm run test`, the log:
```
> myapp@1.0.0 pretest /src
> npm run clean && npm run build


> myapp@1.0.0 clean /src
> lb-clean dist


> myapp@1.0.0 build /src
> lb-tsc es2017 --outDir dist


> myapp@1.0.0 test /src
> lb-mocha --allow-console-logs "dist/test"



  !!

  0 passing (109ms)
  2 failing

  1) HomePageController
       exposes a default home page:
     Error: connect EADDRNOTAVAIL ::1:33967 - Local (:::0)
      at Object._errnoException (util.js:1022:11)
      at _exceptionWithHostPort (util.js:1044:20)
      at internalConnect (net.js:976:16)
      at defaultTriggerAsyncIdScope (internal/async_hooks.js:283:19)
      at net.js:1074:9
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)

  2) PingController
       invokes GET /ping:
     Error: connect EADDRNOTAVAIL ::1:46317 - Local (:::0)
      at Object._errnoException (util.js:1022:11)
      at _exceptionWithHostPort (util.js:1044:20)
      at internalConnect (net.js:976:16)
      at defaultTriggerAsyncIdScope (internal/async_hooks.js:283:19)
      at net.js:1074:9
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)

```

I found few posts about Loopback & docker, but they were all about V3.
Any idea of how to fix it? Or any related materials for docker & V4?

dhmlau:
@ido567 , let me try it out later tonight.  A few things to note:
- HomePageController no longer being generated if you're using the latest @loopback/cli. 
- Regarding the error about :::0, it might be because it's expecting a IPv6 address? There's some change in the generated index.js file to specify the host, see https://github.com/strongloop/loopback-next/blob/master/examples/todo/index.js#L15.

dhmlau:
@ido567, I created the `Dockerfile` below and ran into the same issue:
```
From node:8-slim
EXPOSE 3000
```
[...]

PR #2433, opened 2019-02-20
fix(cli): force test host to be HOST env var or ipv4 interface

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/1995

## Checklist

- [ ] Agree to the CLA (Contributor License Agreement) by [clicking and signing](https://cla.strongloop.com/agreements/strongloop/loopback-next)

### DIFF
--- a/packages/cli/generators/app/templates/src/__tests__/acceptance/test-helper.ts.ejs
+++ b/packages/cli/generators/app/templates/src/__tests__/acceptance/test-helper.ts.ejs
@@ -6,8 +6,17 @@ import {
 } from '@loopback/testlab';
 
 export async function setupApplication(): Promise<AppWithClient> {
+  const config = givenHttpServerConfig();
+
+  // Set host to `HOST` env var or ipv4 loopback interface
+  // By default, docker container has ipv6 disabled.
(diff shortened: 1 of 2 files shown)


bajtos:
@raymondfeng what is the reasoning for adding this new logic to the project template, instead of modifying `givenHttpServerConfig` directly? `givenHttpServerConfig` is a test helper provided by `@loopback/testlab`, it's meant to be use in tests.

We are already detecting Travis CI and enforcing IPv4 there, it would make a lot of sense to me to detect other Docker environments too.

https://github.com/strongloop/loopback-next/blob/af4a55c5a88a3b72a0cdce26669c9b931e2d42a4/packages/testlab/src/http-server-config.ts#L17

raymondfeng:
I had thought about detecting Docker in givenHttpServerConfig() but the logic is not that simple. Docker container by default has ipv6 disabled but it's enabled. Adding the logic in the template makes it easy to customize for user projects.

bajtos:
I see. I am proposing to decouple the Docker fix from the config customization, please review https://github.com/strongloop/loopback-next/pull/2502

Issue #1293, opened 2018-04-27
Allow app developers to configure custom Express middleware


hacksparrow:
Allow Application and Extensions developers to mount Express middleware at application level.

**WORKAROUND:**

Our current recommendation is to create a top-level Express application (where you can easily mount any middleware) and then mount LB4 as a component (middleware) in that application. See https://loopback.io/doc/en/lb4/express-with-lb4-rest-tutorial.html for detailed instructions.

## Acceptance Criteria

- [x] A set of APIs allowing Application developers to mount custom Express middleware
- [x] A set of APIs allowing Extension developers to contribute Express middleware to target application
- [x] Documentation for both App & Extension developers
- [x] A clear guidance on how to ensure correct middleware ordering (this may require a spike)
- [x] Implementation wise, the piece invoking middleware should be implemented as part of the Sequence handling, so that errors reported by Express middleware are routed to `reject` action. (For example, we may introduce a new sequence action to be executed before `findRoute`.)

**Out of scope**
- Register Express middleware per route, controller class or controller method - see #2035

bajtos:
Related: https://github.com/strongloop/loopback-next/issues/559#issuecomment-399077339

> I am proposing to enhance RestServer/RestApplication with an API allowing consumers to mount static files, for example `app.static(path, options)`. This API should leverage Express middleware [server-static](https://www.npmjs.com/package/serve-static) as an implementation detail.
>
> With this new API in place, we can either write an API Explorer component bundling swagger-ui and calling `app.static(pathToSwaggerUI)`, or preferably write a short guide explaining LB4 users how to add swagger-ui to their project themselves.

bajtos:
IMO, this is out of 4.0 GA scope as we have defined it. Relabeling this story as post-GA.

virkt25:
Simplifying usage is probably out of scope of GA but I think we should at the very least have some documentation around how a user can add some express middleware to their application (static file serving for example) by overriding the default `requestHandler`. -- I think that's something important for GA.

bajtos:
I feel the issue description is out of sync with the issue title. I am going to rewrite the description to match the title and focus on LB4 user experience.

Below is the original description for posterity.

---


Since we have decided to use Express as our underlying HTTP framework, `@loopback/rest` and `@loopback/http-server` can be refactored with the use of Express APIs. 

## Current Behavior
Code is not optimized to use Express APIs.

## Expected Behavior
Use Express APIs and remove redundant code.

[...]

Issue #2329, opened 2019-02-03
swagger ui not working when using basepath link is broken


david-sw:
## Description / Steps to reproduce / Feature proposal

in file `src/application.ts`
i added to the contractor 

```ts
this.basePath('/icontrol/v1');
```

so that all controllers will have the base path of `/icontrol/v1`

## Current Behavior

after opening the swagger ui and clicking on the explorer the wrong link is opened 
https://explorer.loopback.io/?url=https://127.0.0.1:5443/icontrol/v1/openapi.json
the file openapi.json is not under the new path and if i change the path manually to 
https://explorer.loopback.io/?url=https://127.0.0.1:5443/openapi.json 
it start working

## Expected Behavior
swagger ui link will not change on the static open-api .json

bajtos:
Possibly related: https://github.com/strongloop/loopback-next/issues/2285

> after opening the swagger ui and clicking on the explorer the wrong link is opened
> https://explorer.loopback.io/?url=https://127.0.0.1:5443/icontrol/v1/openapi.json
> the file openapi.json is not under the new path and if i change the path manually to
> https://explorer.loopback.io/?url=https://127.0.0.1:5443/openapi.json

IIUC, you have configured `basePath` as `/icontrol/v1`, the swagger ui expects the OpenAPI spec to be available at `/icontrol/v1/openapi.json` but the current implementation is mounting the routes at `/openapi.json`.

Is that correct description? Based on the source code, I think the actual behavior is opposite: Explorer expects OpenAPI spec to be available at `/openapi.json` but in fact it's exposed at `/${basePath}/openapi.json`. 

IMO, this is the intended behavior we should eventually get to:
- `basePath` controls the URL of OpenAPI spec endpoints.
- REST API Explorer is aware of `basePath`.

I think we need to fix REST API Explorer to take `basePath` into account when building the spec url here:

https://github.com/strongloop/loopback-next/blob/b5f12be7c2f1f809360122ca72a3e55c5ef92bcb/packages/rest-explorer/src/rest-explorer.controller.ts#L51-L58

Please read through the discussion in https://github.com/strongloop/loopback-next/issues/2285#issuecomment-457573880 too, it contains important information relevant to this issue too.

sanadHaj:
@bajtos  @raymondfeng  
i am aware this not possible solution for this issue , 
getOpenApiSpecUrl method that you mentioned before is only available when the self hosted explorer enabled
[...]

PR #2434, opened 2019-02-20
[RFC] Introduce express middleware extension point

raymondfeng:
This PR introduces an extension point for express middleware so that they can be added and sorted by phase to force a hierarchical router.

Doc: https://github.com/strongloop/loopback-next/blob/express-middleware/packages/express-middleware/README.md

See also #1293

## Checklist

- [ ] Agree to the CLA (Contributor License Agreement) by [clicking and signing](https://cla.strongloop.com/agreements/strongloop/loopback-next)


bajtos:
I am not sure if it's a good idea to re-introduce middleware phases in LB4. From the comments I seen in the past, many LB3 users are not "getting it", they don't understand why should they use middleware phases instead of vanilla express style of registration.

Do we still want to purpose Sequence as the tool for orchestrating request handling pipeline? I think we should focus more on Sequence and treat compatibility with Express middleware as a short-term workaround only.

As I see it, there are principally four middleware phases:
 1. request pre-processing
 2. route handlers
 3. static assets
 4. error handler

In LoopBack 3, the primary motivation for introducing middleware phases was to ensure correct order of invocation of these different middleware types. IIRC, the concept of finer-grained phases for request pre-processing was a kind of a nice side effect.

In LoopBack 4, there is already infrastructure in place to enforce these four phases. What we are missing is API that would allow app developers to register their own request pre-processing middleware.

As I see understand the proposal in your PR, it  is splitting the request pre-processing phase into multiple finer-grained sub-phases. While I see how that can be useful, I am not convinced that the benefits justify the additional complexity (both for our users and the framework codebase).

raymondfeng:
In fact, I would like to refactor this PR into a separate module such as `@loopback/express` to achieve my objectives:

1. Make it easier to leverage LoopBack 4 core (especially the IoC/DI) for existing Express users
2. Make it more scalable for Express applications that consists of multiple modules
  - Discover/register middleware (forcing components to use `app.use` is not feasible)
  - Use IoC/DI for middleware or route handlers
3. Incrementally bring in the full REST experience into Express

It's my intent to create an `Express++` experience using TypeScript and LoopBack 4 core modules.

For LoopBack itself, I personally don't think the sequence of actions is good enough to take over the Express middleware pipeline and we should keep the door open for developers to use existing middleware.

raymondfeng:
I refactored the code into `@loopback/express-middleware`. It's fully functional with documentation. Please review.

raymondfeng:
I would like to port `error` and `final` support from https://github.com/strongloop/loopback-next/pull/1671.

packages/express-middleware/README.md
```diff
@@ -0,0 +1,129 @@
...
+## Basic Use
+
+### Register middleware handler
+
+Handler functions compatible with Express middleware can be registered as
+follows:
+
+```ts
+const expressCtx = new ExpressContext();
+expressCtx.middleware(cors(), {phase: 'cors'});
```
bajtos: Do you expect users to call `middleware` with a single argument only? If not, then I think the following API may be more ergonomic:

```ts
expressCtx.middleware({
  handler: cors(),
  phase: 'cors',
  // path, etc.
});
```

packages/rest/src/rest.server.ts
```diff
@@ -218,6 +219,12 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._applyExpressSettings();
     this._requestHandler = this._expressApp;
 
+    const expressCtx = new ExpressContext(this);
+    expressCtx.setMiddlewareRegistryOptions({
```
bajtos: I am not comfortable forcing Express context on all LoopBack4 users.

Can you find a way how to make this extension an optional addition please?

packages/rest/src/rest.server.ts
```diff
@@ -272,14 +290,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
     const mapping = this.config.openApiSpec!.endpointMapping!;
     // Serving OpenAPI spec
     for (const p in mapping) {
-      this._expressApp.get(p, (req, res) =>
-        this._serveOpenApiSpec(req, res, mapping[p]),
+      expressCtx.middleware(
+        (req, res) => this._serveOpenApiSpec(req, res, mapping[p]),
```
bajtos: I think our REST layer should be providing enough features to allow us to implement `_serverOpenApiSpec` routes as regular LB4 routes with `x-visibility: undocumented`. Instead of relying on your new extension, can we rework this part to use existing LB4 means?

packages/rest/src/rest.server.ts
```diff
@@ -272,14 +290,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+          phase: 'openapi-spec',
+        },
       );
     }
 
     const explorerPaths = ['/swagger-ui', '/explorer'];
-    this._expressApp.get(explorerPaths, (req, res, next) =>
-      this._redirectToSwaggerUI(req, res, next),
+    expressCtx.middleware(
+      (req, res, next) => this._redirectToSwaggerUI(req, res, next),
```
bajtos: Similarly here. We don't support dynamic redirects yet - see https://github.com/strongloop/loopback-next/issues/2022 and the work in progress in https://github.com/strongloop/loopback-next/pull/2529, but I think it should be possible to implement "redirecto to Swagger UI" by implementing a LB4 route that's calling `response.redirect` directly.

packages/rest/src/rest.server.ts
bajtos: Ah, it looks like it was an intentional decision to implement these routes at REST API layer, bypassing the sequence.

https://github.com/strongloop/loopback-next/blob/c3d800700b253e97316fd0871641ea80fcb457f3/packages/rest/src/rest.server.ts#L265-L271

bajtos: Uh oh, this is even more complex than I thought. It turns out that OpenAPI endpoints like `/openapi.json` do not honor `basePath` configuration set at server level.

bajtos: You can see the RouteEntry-based implementation of `/openapi.json` here: 2b5be221

packages/rest/src/rest.server.ts
bajtos: I think this change should be deferred until #2529 and #2547 are landed, at which point we should rework these redirects to leverage `server.redirect` APIs instead of calling `response.redirect` directly.

packages/rest/src/rest.server.ts
bajtos: > Uh oh, this is even more complex than I thought. It turns out that OpenAPI endpoints like `/openapi.json` do not honor `basePath` configuration set at server level.

This problem is already tracked here: https://github.com/strongloop/loopback-next/issues/2329

raymondfeng:
Close to favor https://github.com/strongloop/loopback-next/pull/5118

Issue #2152, opened 2018-12-13
[Spike] Explore the way to resolve inclusion property


jannyHou:
## Description / Steps to reproduce / Feature proposal

A follow-up story of the relation traverse spike https://github.com/strongloop/loopback-next/issues/1952

The previous spike story turns out that we could improve our design of describing&defining model properties.

Copied the discussion and proposal from it:

Thanks for all the feedback! 

I understand that the best outcome of a spike is a concrete solution, but I am afraid the inclusion story is more complicated than we expect,  and it would be better to create separate stories to explorer different potential approaches.

Like what @raymondfeng points out, GraphQL's query system has two features we could borrow:

- use resolver to describe a complicated object and calculate its value
- use different query types to describe the filter of a model

And a overall design question to consider (from @bajtos ) :

How are we going to generate JSON Schema and OpenAPI schema for payloads containing both model properties (Customer's id, name, email) and data of related model (Customer's orders, addresses, etc.). At the moment, we are leveraging @property metadata. If we choose a different approach for describing model+relations in TypeScript, then we also need a new way how to obtain JSON Schema for such new types.

Here is a  proposal of the next steps:

- [ ] Explorer the approach "Define relation property as a resolver"
  - The prototype would be:
    ```js
       import { TodoList } from './TodoList';
       class Todo extends Entity {
          @inclusion()
          TodoList: ()=>TodoList
        }
    ```
- [ ] Explorer the approach "Use a type wrapper to prevent the circular reference problem"
  - The prototype would be: 
    ```js
        export type Related<T> = T | undefined;
        class Todo extends Entity {
          // ...
          todoList: Related<TodoList>;
        }
        class TodoList extends Entity {
         // ...
          todos: Related<Todo>;
        }
    ```
- [ ] Explorer "Generate decorator based filter types"
  - Inspired by the query/mutation type in GraphQL, I am trying to define different partial model types described by decorators
  - The prototype:
     ```js
         // model.ts file
        class Todo extends Entity {
          @property({name: 'id', type: 'string'})
          id: string,
          @property({name: 'desc', type: 'string'})
          desc?: string
          @inclusion()
          TodoList: ()=>TodoList
        }
        // model.query.ts file
        type TodoMutation = Pick<Todo, a union of properties decorated by @property except id>
        type TodoQuery = TodoMutation & Inclusion<Todo>
     ```
- [ ] Explorer the approach "Define multiple model classes for different purpose"
  - Prototype:
    ```js
[...]

PR #2442, opened 2019-02-21
feat(repository): allow optional property definition on belongsTo decorator

### LINKED ISSUE #2256: Allow configuration for relation properties
## Description
When defining a new relation in a model file, the relation property is also decorated with the `@property` decorator called inside the relation decorator.
The problem when creating a relation property this way is that it cannot be configured the same as a standard property.
## Current Behavior
For example in the `Order` model, the property `customerId` from the `@belongsTo` relation cannot be configured to be of type `UUID` (the same type as the defined `Customer` model `id` property):
```
@belongsTo(() => Customer, {keyTo: 'id'})
customerId: string;
```
For example, when trying to create a foreign key for "customerId", it fails with an incompatible type error:
```
detail: 'Key columns "customerid" and "id" are of incompatible types: text and uuid.'
```
## Expected Behavior (proposal)
I think that implementing support for property configuration in relation decorators is required.
Ability to config relation properties, kind of:
```
@belongsTo(() => Customer, {keyTo: 'id'}, {
    type: 'string',
    defaultFn: 'uuid',
    required: true})
customerId: string;
```

### LINKED ISSUE #2345: How to specify database specific property settings when using relation?

How to specify the postgres dataType when using belongsTo-Relation?

Example:
```javascript
@model()
export class ModelA extends Entity {

  @property({
    id: true,
    defaultFn: 'uuidv4',
    mysql: { dataType: 'varchar', dataLength: 36 },
    postgresql: { dataType: 'uuid', columnName: 'id', nullable: 'YES' },
  })
  id?: string;

  @belongsTo(() => ModelB)
  @property({
    required: true,
    mysql: { dataType: 'varchar', dataLength: 36 },
    postgresql: { dataType: 'uuid', columnName: 'modelb_id', nullable: 'NO' },
  })
  modelbId: UUID;
}
```

## Current Behavior

```
Error: Decorator cannot be applied more than once on ModelA.prototype.modelbId
    at PropertyDecoratorFactory.mergeWithOwn (./node_modules/@loopback/metadata/dist/src/decorator-factory.js:321:19)
    at PropertyDecoratorFactory.decorate (./node_modules/@loopback/metadata/dist/src/decorator-factory.js:219:25)
    at ./node_modules/@loopback/metadata/dist/src/decorator-factory.js:327:47
    at ./node_modules/@loopback/repository/dist/src/relations/belongs-to/belongs-to.decorator.js:23:45
    at DecorateProperty (./node_modules/reflect-metadata/Reflect.js:553:33)
    at Object.decorate (./node_modules/reflect-metadata/Reflect.js:123:24)
    at __decorate (./dist/src/models/modela.model.js:4:92)
    at Object.<anonymous> (./dist/src/models/modela.model.js:33:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (./dist/src/models/index.js:86:10)
```

## Expected Behavior

I need a way to specify dataType for postgres for fields which a marked with belongsTo decorator. I think the implicit definition of property by belongsTo should merge with existing property decorator?

derdeka:
allow optional property definition on belongsTo decorator as property decorator can not applied twice

Fixes #2345
Fixes #2256

## Checklist

- [x] Agree to the CLA (Contributor License Agreement) by [clicking and signing](https://cla.strongloop.com/agreements/strongloop/loopback-next)

### DIFF
--- a/packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
+++ b/packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
@@ -198,6 +198,51 @@ describe('relation decorator', () => {
         },
       });
     });
+
+    it('accepts additional property metadata', () => {
+      @model()
+      class AddressBook extends Entity {
+        @property({id: true})
+        id: string;
+      }
+
+      @model()
+      class Address extends Entity {
+        id: string;
+        @belongsTo(
+          () => AddressBook,
+          {},
+          {
+            length: 36,
+            postgresql: {
+              dataType: 'uuid',
+            },
+          },
+        )
+        addressBookId: string;
+      }
+
+      const jugglerMeta = MetadataInspector.getAllPropertyMetadata(
+        MODEL_PROPERTIES_KEY,
+        Address.prototype,
+      );
+      expect(jugglerMeta).to.eql({
+        addressBookId: {
+          type: String,
+          length: 36,
+          postgresql: {
+            dataType: 'uuid',
+          },
+        },
+      });
+      expect(Address.definition.relations).to.containDeep({
+        addressBook: {
+          keyFrom: 'addressBookId',
+          name: 'addressBook',
+          type: 'belongsTo',
+        },
+      });
+    });
   });
 });
 
--- a/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
+++ b/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts
@@ -14,22 +14,30 @@ import {BelongsToDefinition, RelationType} from '../relation.types';
  * @param targetResolver A resolver function that returns the target model for
  * a belongsTo relation
  * @param definition Optional metadata for setting up a belongsTo relation
+ * @param propertyDefinition Optional metadata for setting up the property
  * @returns {(target: Object, key:string)}
  */
 export function belongsTo<T extends Entity>(
   targetResolver: EntityResolver<T>,
   definition?: Partial<BelongsToDefinition>,
+  propertyDefinition?: Partial<PropertyDefinition>,
 ) {
   return function(decoratedTarget: Entity, decoratedKey: string) {
-    const propMeta: PropertyDefinition = {
-      type: MetadataInspector.getDesignTypeForProperty(
-        decoratedTarget,
-        decoratedKey,
-      ),
-      // TODO(bajtos) Make the foreign key required once our REST API layer
-      // allows controller methods to exclude required properties
-      // required: true,
-    };
+    const propMeta: PropertyDefinition = Object.assign(
+      {},
+      // properties provided by the caller
+      propertyDefinition,
+      // properties enforced by the decorator
+      {
+        type: MetadataInspector.getDesignTypeForProperty(
+          decoratedTarget,
+          decoratedKey,
+        ),
+        // TODO(bajtos) Make the foreign key required once our REST API layer
+        // allows controller methods to exclude required properties
+        // required: true,
+      },
+    );
     property(propMeta)(decoratedTarget, decoratedKey);
 
     // @belongsTo() is typically decorating the foreign key property,


raymondfeng:
@derdeka Thank you for the PR. Would you please add a test case to cover such usage?

derdeka:
@raymondfeng
I've added a unit test - hopefully in the correct file.
The build currently fails because of issue #2455

raymondfeng:
ok to test

packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
```diff
@@ -198,6 +198,55 @@ describe('relation decorator', () => {
...
+          () => AddressBook,
+          {},
+          {
+            length: 36,
+            postgresql: {
+              dataType: 'uuid',
+            },
+          },
+        )
+        addressBookId: string;
```
raymondfeng: I'm not very happy that we can have to mix the property metadata with relations.

Ideally, `@belongsTo` should be applied to the `addressBook` relational property and `@property` to the `addressBookId`. Something like:
```ts
      @model()
      class Address extends Entity {
        id: string;
        @belongsTo(
          () => AddressBook,
          {},
        )
        addressBook?: AddressBook;

        @property({
            length: 36,
            postgresql: {
              dataType: 'uuid',
            },
          })
        addressBookId: string;
```

WDYT?

elv1s: What will the `address` table look like in the database? Specifically what is the column type of `addressBook`

Another approach might be to allow multiple decorators on a property, with the constraint that the decorator keys must be unique.

orshlom: I think that using two different properties in a model to define a relation is redundant.
In my opinion, the property `addressBook?: AddressBook;` @raymondfeng suggested should be also the one internally creating the column `addressBookId` in the DB. This way it can get automatically configured to comply with the PK it references.
In addition, property decorator (`@belongsTo()`) may also support additional configuration for the FK column in the DB (`addressBookId`).

derdeka: > ```ts
>       @model()
>       class Address extends Entity {
>         id: string;
>         @belongsTo(
>           () => AddressBook,
>           {},
>         )
>         addressBook?: AddressBook;
> 
>         @property({
>             length: 36,
>             postgresql: {
>               dataType: 'uuid',
>             },
>           })
>         addressBookId: string;
> ```

@raymondfeng I really like this approach as it is clearly seperates concerns and gives the developer more control of just reading the id or loading the related object (with subquery to the database).

Should this work already with the latest release or are changes needed to make this possible? (I tried this approach on my application and got an exception "500 Error: Invalid type for property type")

elv1s: When I run the code it creates a text column for addressBook. This seems confusing to have a `AddressBook` table and `AddressBook` column with the same information. I agree with @orshlom that two properties are redundant and maybe the `@belongsTo` property can automatically create an `addressBookId` property to link to the primary key.

Also I'm wondering what the API for the user will be when they call the `Address` endpoint.  Will the user be able to create an Address and AddressBook in a single POST. How will they add a reference to an existing addressBook?

@derdeka The issue you are running into might be related to `length`. You can move it to `dataLength` within `postgresql`, but when I do that I'm getting an error `type modifier is not allowed for type "uuid"`

drdmitry: As I understood, as `belongsTo` decorator uses `property` decorator inside the code, this means that on every `belongsTo` decorator call - new property will be created, and then `belongsTo` creating the relation for the same.

https://github.com/strongloop/loopback-next/blob/4cee896e7f7b2b503b118e560c31036626587f47/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L19-L56

For the suggested approach of splitting `belongsTo` and `property` decorators for defining the relation - this doesn't work, because it is trying to create the same property 2nd time.
Could we somehow map property which is created/defined in `belongsTo` to the property created with `property` decorator?
Or could we have creating property in `belongsTo` only when it is not created yet?
What do you think? How this could be solved?

elv1s:
Is this PR still valid? I'm interested in this pull request since it is the same fix for #2256

It sounds like we don't want to mix the `property` and `belongsTo` decorators. The solution provided by @raymondfeng might work but it adds redundant info to the db and schema which will be confusing to the user.

Is there a way to prevent the `addressBook` property from being part of the schema or created in the db?

derdeka:
I also tried to map addressBook to the existing addressBookId, but i did not get it working. For me is the PR still a valid workaround.

@raymondfeng @bajtos Can you please advise how to continue?

bajtos:
> It sounds like we don't want to mix the property and belongsTo decorators. The solution provided by @raymondfeng might work but it adds redundant info to the db and schema which will be confusing to the user.

Here is one reason why we should not mix the property and the relation decorator: 
 - When creating a new model instance, or updating an existing model instance, we don't want the data type to include the navigational property. `addressRepo.create(data)` does not support updating the AddressBook instance we are belonging to, therefore  `data` type should exclude `addressBook` property.
- Navigational properties are needed only when querying data via `repo.find`.
- This difference must be propagated to OpenAPI schemas used by REST API endpoints.

For long term, I would like to propose following API at conceptual level (implementation details will be most likely different):

```ts
// Relations are defined at model-level
@belongsTo(() => AddressBook)
@model()
class Address extends Entity {
  // the foreign key is defined explicitly
  @foreignKey(() => AddressBook, 'id')
  @property({
    length: 36,
    postgresql: {
      dataType: 'uuid',
    },
  })
  addressBookId: string;
}

// A custom class or interface describes format
// of data returned by a query
@model()
class AddressWithRelations extends Address {
  @property()
  addressBook?: AddressBook;
}

// Repository API
class DefaultCrudRepository {
  find(filter: Filter<Address>): Promise<AddressWithRelations>;
  create(data: Address): Promise<Address>;
  // ...
}
```

bajtos:
Relatively recently, we changed `@hasMany` decorator so that it no longer calls `@property` decorator under the hood. IMO, we should make the same change to `@belongsTo` and `@hasOne` decorators too, to ensure consistent behavior.

Unfortunately, such change is backwards-incompatible and requires a new semver-major release of `@loopback/repository`.  We need to follow the steps in [Making breaking changes](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#making-breaking-changes) section of our DEV docs.

While we are making a new major version, I think it would be good to use this opportunity to review other existing problems in the current implementation of relations and ideally fix all those that are backwards incompatible. (E.g. rework relation decorators to be applied on model classes, not navigational properties.)

Considering the effort required for that, it may be best to land this pull request as a short-term workaround.

bajtos:
Related: in https://github.com/strongloop/loopback-next/issues/2152, we will be investigating how to best represent navigational properties in model classes.

bajtos:
> Relatively recently, we changed @hasMany decorator so that it no longer calls @property decorator under the hood. IMO, we should make the same change to @belongsTo and @hasOne decorators too, to ensure consistent behavior.
>
> Unfortunately, such change is backwards-incompatible and requires a new semver-major release of @loopback/repository. We need to follow the steps in Making breaking changes section of our DEV docs.
>
> While we are making a new major version, I think it would be good to use this opportunity to review other existing problems in the current implementation of relations and ideally fix all those that are backwards incompatible. (E.g. rework relation decorators to be applied on model classes, not navigational properties.)

On the second thought, I think there is a way how to make these changes in a backwards-compatible way.

1. For existing applications, keep `@belongsTo` calling `@property` under the hood. Land this PR to allow consumers to provide additional property metadata.

2. Going forward, extend `@belongsTo` to support decoration of model classes, as I outlined in 
https://github.com/strongloop/loopback-next/pull/2442#issuecomment-470559945. Keep supporting property-based decoration for backwards compatibility, possibly deprecate it at runtime via [depd](https://www.npmjs.com/package/depd).

```ts
export function belongsTo<T extends Entity>(
  targetResolver: EntityResolver<T>,
  definition?: Partial<BelongsToDefinition>,
) {
  return applyBelongsTo;

  function applyBelongsTo(decoratedTarget: typeof Entity): void;
  function applyBelongsTo(decoratedTarget: Entity, decoratedKey: string): void;
  function applyBelongsTo(
    decoratedTarget: Entity | typeof Entity,
    decoratedKey?: string,
  ) {
    if (typeof decoratedTarget === 'function' && decoratedKey === undefined) {
      // model-level decoration
   } else {
     // property-level decoration
  }
}
```

packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
```diff
@@ -198,6 +198,55 @@ describe('relation decorator', () => {
         },
       });
     });
+
+    it('accepts additional property metadata', () => {
+      @model()
+      class AddressBook extends Entity {
+        @property({
+          id: true,
+          length: 36,
```
bajtos: AFAICT, this additional configuration for `AddressBook#id` property is not relevant to this particular test. Please simplify the property definition to provide only the metadata that's needed to make this test fail now and pass with your fix in place.

I think the following should work?

```ts
@model()
class AddressBook extends Entity {
  @property({id: true})
  id: string;
}
```

packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
```diff
@@ -198,6 +198,55 @@ describe('relation decorator', () => {
...
+          postgresql: {
+            dataType: 'uuid',
+          },
+        })
+        id: string;
+      }
+      @model()
```
bajtos: Please add an empty line to visually delimit the two model classes.

```
}

@model()
```

packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
derdeka: done

packages/repository/src/__tests__/unit/decorator/relation.decorator.unit.ts
derdeka: done

raymondfeng:
> In that light, I'd like us to land this pull request as a short-term fix, and thus buy us more time to address this issue in a wider context.

+1.

derdeka:
@bajtos what is your timeline for merging this PR?

bajtos:
Landed, thank you @derdeka for the contribution 🎉

Issue #1851, opened 2018-10-15
Enable eslint for CLI package


bajtos:
The package [packages/cli](https://github.com/strongloop/loopback-next/tree/master/packages/cli) is the only package written in JavaScript and because it's an exception in our TypeScript-based project, it's not covered by any linter. As a result, this part of our codebase may contain silly bugs that could have been caught by a linter.

I am proposing to start running `eslint` for `packages/cli`.

Alternatively, we can rewrite `@loopback/cli` into TypeScript, but that would be a larger task IMO.

## Acceptance criteria

- [ ] `npm run lint` executes `eslint` for `packages/cli`
- [ ] `npm run lint:fix` executes `eslint --fix`
- [ ] eslint is configured using [eslint-config-loopback](https://www.npmjs.com/package/eslint-config-loopback) with custom configuration to disable eslint rules clashing with our Prettier formatting.
- [ ] all linting errors are fixed - this should be a standalone commit in the pull request.

raymondfeng:
1. +1 to use eslint as the short-term improvement.

2. Converting the package into TypeScript might not be too complicated as we have used ES6 classes for most of the code.

bajtos:
>  eslint is configured using eslint-config-loopback with custom configuration to disable eslint rules clashing with our Prettier formatting.

We can use [eslint-plugin-prettier](https://www.npmjs.com/package/eslint-plugin-prettier) or [eslint-config-prettier](https://www.npmjs.com/package/eslint-config-prettier) to help with that.

bajtos:
Let's wait until https://github.com/strongloop/loopback-next/issues/2608 is done.

dhmlau:
Closing it as done.

PR #2492, opened 2019-02-26
feat: introduce eslint with typescript-eslint to replace tslint

raymondfeng:
See some background at:
- https://www.infoq.com/news/2019/02/tslint-deprecated-eslint
- https://github.com/typescript-eslint/typescript-eslint

## Checklist


raymondfeng:
See https://github.com/strongloop/loopback-next/issues/1851

bajtos:
Yay! Switching from tslint to eslint was on my personal todo list too.

Let me post my notes before I review the changes:

- https://github.com/desktop/desktop/pull/6684 is an example from a different project that has recently switched from tslint to eslint
- can we use `--cache` option to speed up linting?
- there are two eslint plugins to improve integration with prettier: [eslint-plugin-prettier](
https://www.npmjs.com/package/eslint-plugin-prettier) checks formatting and reports violations, [eslint-config-prettier](https://www.npmjs.com/package/eslint-config-prettier) disables any eslint rules that may clash with prettier. Personally, prefer to include prettier linting in eslint run, because standalone `prettier` is rather slow and AFAIK it does not offer any options to cache the results (yet).
- If we move prettier checks to eslint, we must still preserve Prettier checks for our Markdown documents! This utility may be useful: https://github.com/ikatyang/prettylint

.vscode/settings.json
```diff
@@ -24,6 +24,15 @@
   "files.insertFinalNewline": true,
   "files.trimTrailingWhitespace": true,
   "tslint.ignoreDefinitionFiles": true,
```
bajtos: Can we remove this `tslint` config too?

examples/hello-world/package.json
```diff
@@ -43,10 +43,13 @@
...
-    "@loopback/tslint-config": "^2.0.1",
+    "@loopback/eslint-config": "^1.0.0-1",
     "@types/node": "^10.11.2",
-    "tslint": "^5.12.0",
-    "typescript": "^3.3.1"
+    "eslint": "^5.14.1",
+    "typescript": "^3.3.1",
+    "@typescript-eslint/eslint-plugin": "^1.4.2",
+    "@typescript-eslint/parser": "^1.4.2",
+    "eslint-plugin-eslint-plugin": "^2.0.1"
```
bajtos: Is it necessary to duplicate these three dependencies in every package? 

```
    "@typescript-eslint/eslint-plugin": "^1.4.2",
    "@typescript-eslint/parser": "^1.4.2",
    "eslint-plugin-eslint-plugin": "^2.0.1"
```

Can we find a way how to move them into `@loopback/eslint-config` instead?

packages/cli/generators/openapi/spec-helper.js
```diff
@@ -196,7 +196,7 @@ function buildMethodSpec(controllerSpec, op, options) {
   const paramNames = {};
   if (parameters) {
     args = parameters.map(p => {
-      const name = escapeIdentifier(p.name);
+      let name = escapeIdentifier(p.name);
```
bajtos: This is suspicious, why is it necessary to change from `const` to `let`?

packages/cli/generators/project/templates/.eslintrc.js.ejs
```diff
@@ -0,0 +1,50 @@
...
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  plugins: ['eslint-plugin', '@typescript-eslint'],
+  env: {
+    es6: true,
+    node: true,
+  },
+  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
+  rules: {
```
bajtos: Please don't duplicate the content of `@loopback/eslint-config` here, it will quickly get out of sync.

IMO, we should always extend from `@loopback/eslint-config`, it's a shared config that's can be used independently from `@loopback/build`.

packages/eslint-config/index.js
```diff
@@ -0,0 +1,6 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/eslint-config
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+module.exports = require('./.eslintrc');
```
bajtos: I find this rather ugly. Can we use `main` property in `package.json` to specify the file providing the configuration?

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
```
bajtos: I believe the convention is to call this file `eslint.json`, see e.g. https://github.com/strongloop/eslint-config-loopback/blob/master/eslint.json

I am ok to use JavaScript too, in which case I think the file should be called `eslint.js`.

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    '@typescript-eslint/no-var-requires': 'off',
+    '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/no-object-literal-type-assertion': 'off',
+    '@typescript-eslint/no-parameter-properties': 'off',
+    '@typescript-eslint/no-angle-bracket-type-assertion': 'off',
+    '@typescript-eslint/prefer-interface': 'off',
+    '@typescript-eslint/no-namespace': 'off',
+
+    // Disable the base `no-unused-vars` to use `typescript`
+    'no-unused-vars': 'off',
```
bajtos: This is not great, we want to apply `no-unused-vars` to our JavaScript code in `@loopback/cli`.

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+      {
+        vars: 'all',
+        args: 'none',
+        ignoreRestSiblings: false,
+      },
+    ],
+    '@typescript-eslint/ban-types': 'error',
+    '@typescript-eslint/no-triple-slash-reference': 'off',
+    '@typescript-eslint/no-empty-interface': 'off',
+  },
```
bajtos: Have you verified that all rules we are currently using in `@loopback/tslint-config` already have their counterpart in `@typescript-eslint` or other eslint plugins?

If there are any rules that are not available yet, then we should have a discussion whether we are ok to accept that limitation.

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -113,7 +113,7 @@ describe('ClassDecoratorFactory', () => {
     expect(() => {
       @classDecorator({x: 1})
       @classDecorator({y: 2})
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos: Please keep using single-line comments, I believe eslint does support them.

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

Same comment applies to all other places your are changing in this pull request.

packages/metadata/src/inspector.ts
```diff
@@ -11,6 +11,8 @@ import {
   DecoratorType,
 } from './types';
 
+/* eslint-disable @typescript-eslint/ban-types */
```
bajtos: Ditto.

packages/metadata/src/decorator-factory.ts
```diff
@@ -9,7 +9,9 @@ import * as debugModule from 'debug';
 import {MetadataMap, DecoratorType, MetadataKey} from './types';
 const debug = debugModule('loopback:metadata:decorator');
 
-// tslint:disable:no-any
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/ban-types */
```
bajtos: Where is `ban-types` rule coming from? It is not enabled in our current tslint config.

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    '@typescript-eslint/no-unused-vars': [
+      // Turn it off for now before we find out how it applies to decorators
+      'off',
+      {
+        vars: 'all',
+        args: 'none',
+        ignoreRestSiblings: false,
+      },
+    ],
+    '@typescript-eslint/ban-types': 'error',
```
bajtos: Let's remove this rule from this initial pull request. Our current tslint config does not enforce `ban-types`.

examples/hello-world/package.json
raymondfeng: Unfortunately not yet. See https://github.com/eslint/rfcs/tree/master/designs/2018-simplified-package-loading

packages/cli/generators/openapi/spec-helper.js
raymondfeng: I'll check.

packages/cli/generators/project/templates/.eslintrc.js.ejs
raymondfeng: +1.

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
raymondfeng: I noticed that `eslint` only accepts `/* eslint-disable ...  */. 

`// eslint-disable-next-line ...` seems to be fine.

packages/metadata/src/decorator-factory.ts
raymondfeng: This rule actually helps us catch a few issues, such as:

```ts
let str: String; // it should be `let str: string`
```

packages/eslint-config/.eslintrc.js
raymondfeng: Here is my approach:

1. Use the recommended rules from `eslint` and `typescript-eslint` to expose all violations in our code base
2. Fix violations if they make good sense
3. Disable certain rules if we need to discuss or spend more time to fix

examples/hello-world/package.json
bajtos: Yeah, that's rather unfortunate. Thank you for digging the RFC.

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
bajtos: Interesting. 

Anyhow, since `// eslint-disable-next-line` is supported according to your analysis, I think we should keep using it.

I am ok to use `/*` for overrides that apply to more than one line, if that's what eslint forces us to do. I actually quite like that on the second thought, because it will make such statement to stand out from our regular code and make them easier to spot during code reviews.

packages/metadata/src/decorator-factory.ts
bajtos: I see. I don't understand the nuance between `String` and `string`, but I trust the TypeScript & eslint teams that it's a good idea to use `string` instead of `String`.

As I wrote elsewhere, this patch is already too large, let's not make it even larger by enabling extra rules.

I would like to keep this patch focused on the migration from TSLint to ESLint only, and leave out additional rules for follow pull requests. I have my own list of rules to enable too :)

bajtos: > I would like to keep this patch focused on the migration from TSLint to ESLint only

Based on your comment below, I better understand your motivation here and don't mind to enable ban-types rule as part of this PR.

packages/eslint-config/.eslintrc.js
bajtos: > Use the recommended rules from `eslint` and `typescript-eslint` to expose all violations in our code base

I see the merit of this. While I am not very happy that it increases the size of this pull request, I guess I can live with that.

However, I feel it's important to understand which rules enabled by our current configuration will be lost after your change. Quoting from an earlier comment:

> Have you verified that all rules we are currently using in `@loopback/tslint-config` already have their counterpart in `@typescript-eslint` or other eslint plugins?
> 
> If there are any rules that are not available yet, then we should have a discussion whether we are ok to accept that limitation.

The current TSLint config was also based on the setup recommended at the time we created them. What's more important, it contains rules that helper us to detect important bad practices in the past. I don't want to loose them.

What I would like to see:
 - List of rules that are enabled now but won't be applied after your change
 - If there are any GH issues in eslint/tslint projects that are tracking these missing rules, then please reference them in the list.
 - The PR description should be updated with this information.

benchmark/src/__tests__/benchmark.integration.ts
```diff
@@ -17,7 +17,7 @@ const DUMMY_STATS: EndpointStats = {
 
 describe('Benchmark (SLOW)', function() {
   // Unfortunately, the todo app requires one second to start
-  // tslint:disable-next-line:no-invalid-this
+  /* eslint-disable-next-line no-invalid-this */
```
bajtos:
```suggestion
  // eslint-disable-next-line no-invalid-this */
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -352,7 +352,7 @@ describe('PropertyDecoratorFactory', () => {
 
   it('throws if applied more than once on the same property', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars */
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -400,7 +400,7 @@ describe('PropertyDecoratorFactory for static properties', () => {
 
   it('throws if applied more than once on the same static property', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -448,7 +448,7 @@ describe('MethodDecoratorFactory', () => {
 
   it('throws if applied more than once on the same method', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -496,7 +496,7 @@ describe('MethodDecoratorFactory for static methods', () => {
 
   it('throws if applied more than once on the same static method', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars */
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -545,7 +545,7 @@ describe('ParameterDecoratorFactory', () => {
 
   it('throws if applied more than once on the same parameter', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -634,7 +634,7 @@ describe('ParameterDecoratorFactory for a static method', () => {
...
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
```diff
@@ -695,7 +695,7 @@ describe('MethodParameterDecoratorFactory with invalid decorations', () => {
 
   it('reports error if the # of decorations exceeeds the # of params', () => {
     expect(() => {
-      // tslint:disable-next-line:no-unused
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
bajtos: It makes me wonder - if we removed the class name, would eslint still complain about unused variables?

```ts
expect(() => {
  class {
    @propertyDecorator({x: 1})
    @propertyDecorator({y: 2})
    myProp: string;
  }
}).to.throw(
  // ...
);
```

packages/context/src/binding.ts
```diff
@@ -20,6 +20,8 @@ import {
 
 const debug = debugModule('loopback:context:binding');
 
+/* eslint-disable @typescript-eslint/ban-types */
```
bajtos: Why are we disabling this rule instead of fixing the linting violations as we do in other files?

packages/context/src/binding-key.ts
```diff
@@ -5,7 +5,7 @@
 
 export type BindingAddress<T = unknown> = string | BindingKey<T>;
 
-// tslint:disable-next-line:no-unused
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
bajtos:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```

packages/context/src/context.ts
```diff
@@ -34,6 +34,8 @@ const pEvent = require('p-event');
 
 const debug = debugModule('loopback:context');
 
+/* eslint-disable @typescript-eslint/ban-types */
```
bajtos: Why are we disabling this rule instead of fixing the linting violations as we do in other files?

packages/context/src/inject.ts
```diff
@@ -24,6 +24,8 @@ const PROPERTIES_KEY = MetadataAccessor.create<Injection, PropertyDecorator>(
   'inject:properties',
 );
 
+/* eslint-disable @typescript-eslint/ban-types */
```
bajtos: Why are we disabling this rule instead of fixing the linting violations as we do in other files?

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -41,7 +41,7 @@ describe('HttpCachingProxy', () => {
 
   it('proxies HTTP requests', async function() {
     // Increase the timeout to accomodate slow network connections
-    // tslint:disable-next-line:no-invalid-this
+    /* eslint-disable-next-line no-invalid-this */
```
bajtos:
```suggestion
    // eslint-disable-next-line no-invalid-this
```

There are too many places to point out, please search for them yourself. Here is a pattern that works well for me: `/* eslint-disable-next-line`

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+  env: {
+    es6: true,
+    node: true,
+  },
+  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
+  rules: {
+    'comma-dangle': ['error', 'always-multiline'],
+    'no-mixed-operators': 'error',
+    'no-console': 'off',
+    'no-undef': 'off',
```
bajtos: What is this rule about and why are we turning it off, against the recommendations?

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    'no-inner-declarations': 'off',
```
bajtos: What is this rule about and why are we turning it off, against the recommendations?

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    // TypeScript allows method overloading
+    'no-dupe-class-members': 'off',
+    'no-useless-escape': 'off',
```
bajtos: What is this rule about and why are we turning it off, against the recommendations?

packages/eslint-config/.eslintrc.js
```diff
@@ -0,0 +1,66 @@
...
+    '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/explicit-member-accessibility': 'off',
+    '@typescript-eslint/no-var-requires': 'off',
+    '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/no-object-literal-type-assertion': 'off',
+    '@typescript-eslint/no-parameter-properties': 'off',
+    '@typescript-eslint/no-angle-bracket-type-assertion': 'off',
+    '@typescript-eslint/prefer-interface': 'off',
+    '@typescript-eslint/no-namespace': 'off',
```
bajtos: I see a bunch of rules that we are disabling against the recommendations. What is your reasoning for disabling them? Also why is `ban-types` treated differently from these rules? Is it because fixing `ban-types` violations is easier than fixing violations of these other rules? That seems to be a poor excuse to me.

IMO, we should add `ban-type` to the list of disabled rules to keep this PR smaller and then incrementally try to enable most of the recommended rules, one by one. Starting with `ban-types` if you like.

packages/eslint-config/.eslintrc.js
bajtos: BTW, if `ban-types` is a recommended rule, then why do we have to explicitly enable it here? Isn't it automatically inherited from the recommended configuration?

bajtos:
@raymondfeng I would love to see this pull request landed soon, but unfortunately it seems to make too many changes that are very loosely related (if related at all). Can you find a way how to reduce the size of this patch so that we can land it sooner?

I like the idea of inheriting all recommended rules. For this first iteration, I prefer to disable any of these recommendations that we are not already complying with, to avoid large code changes and/or adding `/* eslint-disable */` statements that we probably should want to remove in the near future anyways.

raymondfeng:
@bajtos Let's start with https://github.com/strongloop/loopback-next/pull/2572.

raymondfeng:
@bajtos PTAL.

raymondfeng:
FYI: https://github.com/typescript-eslint/typescript-eslint/issues/571.

dhmlau:
@raymondfeng, I've skimmed through the PR and the changes make sense to me. I have 2 questions:
1. does it mean we can remove `@loopback/tslint-config` eventually?
2. I might've missed the PR to create `@loopback/eslint-config`. The current version of this module is `1.0.0-1`. When we make a new release, would it be `1.0.1` or something else?

raymondfeng:
> does it mean we can remove @loopback/tslint-config eventually?

Yes.

> I might've missed the PR to create @loopback/eslint-config. The current version of this module is 1.0.0-1. When we make a new release, would it be 1.0.1 or something else?

`1.0.0`.

packages/rest/src/__tests__/unit/coercion/utils.ts
```diff
@@ -68,11 +68,13 @@ export async function testCoercion<T>(config: TestArgs<T>) {
         pathParams.aparameter = config.rawValue;
         break;
       case 'query':
-        const q = qs.stringify(
-          {aparameter: config.rawValue},
-          {encodeValuesOnly: true},
-        );
-        url += `?${q}`;
+        {
```
bajtos: Weird, why are we wrapping this code in a `{}` block?

Can you find a different solution please, one that will require less code changes?

packages/rest/src/types.ts
```diff
@@ -105,7 +105,6 @@ export type OperationArgs = any[];
  * operation results from other "any" typed values.
  */
 export type OperationRetval = any;
-// tslint:enable:no-any
```
bajtos: Enable the disabled rules again please.

packages/testlab/should-as-function.d.ts
```diff
@@ -210,8 +213,8 @@ interface Internal extends ShouldInternal {
   inspect(value: any, obj: any): any;
 }
 
-declare var should: Internal;
-declare var Should: Internal;
+let should: Internal;
+// let Should: Internal;
```
bajtos: I am not able to comprehend the impact of this change where you are removing `declare var`. Is it actually a good idea to lint `.d.ts` files?

Anyhow, if this change works well, then I am ok with it, but please remove the commented-out line.

bajtos:
@raymondfeng Travis CI build is failing for `verify-docs` step, PTAL.

https://travis-ci.org/strongloop/loopback-next/jobs/539476548
```
lerna bootstrap --scope @loopback/tsdocs
lerna notice cli v3.14.1
lerna info versioning independent
lerna info ci enabled
lerna notice filter including "@loopback/tsdocs"
lerna info filter [ '@loopback/tsdocs' ]
lerna WARN bootstrap Installing local packages that do not match filters from registry
lerna info Bootstrapping 1 package
lerna info Installing external dependencies
lerna ERR! npm install --no-save exited 1 in '@loopback/tsdocs'
lerna ERR! npm install --no-save stderr:
npm WARN deprecated @types/handlebars@4.1.0: This is a stub types definition. 
    handlebars provides its own type definitions, so you do not need this installed.
npm ERR! code ETARGET
npm ERR! notarget No matching version found for @loopback/eslint-config@^1.0.0
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
npm ERR! notarget 
npm ERR! notarget It was specified as a dependency of '@loopback/tsdocs'
npm ERR! notarget 
```

.vscode/settings.json
```diff
@@ -26,7 +26,15 @@
...
-  "tslint.nodePath": "./packages/build/node_modules",
-  "typescript.tsdk": "./packages/build/node_modules/typescript/lib"
+  "typescript.tsdk": "./packages/build/node_modules/typescript/lib",
+  "eslint.autoFixOnSave": true,
+  "eslint.run": "onSave",
+  "eslint.nodePath": "./packages/build/node_modules",
+  "eslint.validate": [
+    "javascript",
+    {
+        "language": "typescript",
```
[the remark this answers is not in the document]
raymondfeng: See https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint. The setting is an array. It basically says `javascript` is validated and `typescript` is validated and auto-fixed.

packages/booter-lb3app/fixtures/lb3app/common/models/coffee-shop.js
```diff
@@ -2,12 +2,12 @@
 
 module.exports = function(CoffeeShop) {
   CoffeeShop.status = function(cb) {
-    var currentDate = new Date();
-    var currentHour = currentDate.getHours();
-    var OPEN_HOUR = 6;
-    var CLOSE_HOUR = 20;
+    let currentDate = new Date();
```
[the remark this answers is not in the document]
raymondfeng: Good catch. I now turn on `prefer-const` rule.

packages/eslint-config/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@loopback/eslint-config",
-  "version": "1.0.0-2",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
raymondfeng: Unfortunately, lerna publishes it as `1.0.0-2`.

packages/rest/src/__tests__/unit/coercion/utils.ts
raymondfeng: I think it's due to https://eslint.org/docs/rules/no-case-declarations

packages/testlab/should-as-function.d.ts
raymondfeng: I'll skip `*.d.ts`.

packages/eslint-config/eslintrc.js
```diff
@@ -95,6 +101,19 @@ module.exports = {
       {
         vars: 'all',
         args: 'none', // none - do not check arguments
+        /*
+         * The following is a workaround to the issue that parameter decorators
+         * are treated as `unused-vars`.
```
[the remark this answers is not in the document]
raymondfeng: See https://github.com/typescript-eslint/typescript-eslint/issues/571. I updated the comment.

packages/eslint-config/eslintrc.js
```diff
@@ -95,6 +101,19 @@ module.exports = {
...
+         *
+         * @example
+         * ```ts
+         * import {inject} from '@loopback/context';
+         * class MyController {
+         *   constructor(@inject('foo') foo: string) {}
+         * }
+         * ```
+         */
+        varsIgnorePattern: 'inject|(\\w+)Bindings',
```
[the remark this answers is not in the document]
raymondfeng: We use `@inject(RestBindings.Http.REQUEST)`. Both `@inject` and `RestBindings` are treated as no-used due to https://github.com/typescript-eslint/typescript-eslint/issues/571.

raymondfeng:
> Please don't duplicate the content of @loopback/eslint-config here, it will quickly get out of sync.
IMO, we should always extend from @loopback/eslint-config, it's a shared config that's can be used independently from @loopback/build.

Fixed.

packages/context/src/inject.ts
raymondfeng: It does not allow `Object` and there is no configuration to exclude certain types.

packages/metadata/src/__tests__/unit/decorator-factory.unit.ts
raymondfeng: A class declaration without `default` modifier must have a name.

.vscode/settings.json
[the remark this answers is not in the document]
bajtos: Fair enough. Can you capture this in a code comment please? I believe `settings.json` allow comments.

Issue #2495, opened 2019-02-27
Relations' repository bindings are broken after first time access 


shazha:
I'm running the official to-do-list example against the latest loopback4 packages, and the following error is thrown when I tried to access `GET /todo-lists/1/todos`:
```
Unhandled error in GET /todo-lists/1/todos: 500 Error: The key repositories.TodoRepository was not bound to any value.
    at RequestContext.getBinding (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/context.js:493:15)
    at RequestContext.getValueOrPromise (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/context.js:528:30)
    at RequestContext.get (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/context.js:471:27)
    at DefaultHasManyRepository.getter [as getTargetRepository] (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/inject.js:171:20)
    at DefaultHasManyRepository.find (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/repository/dist/relations/has-many/has-many.repository.js:24:45)
    at TodoListTodoController.find (/Users/shazha/github/loopback-next/examples/todo-list/dist/controllers/todo-list-todo.controller.js:31:50)
    at value_promise_1.transformValueOrPromise.args (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/resolver.js:177:41)
    at Object.transformValueOrPromise (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/value-promise.js:231:16)
    at Object.invokeMethod (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/context/dist/resolver.js:172:28)
    at ControllerRoute.invokeHandler (/Users/shazha/github/loopback-next/examples/todo-list/node_modules/@loopback/rest/dist/router/controller-route.js:65:32)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```
It seems  repository bindings for all relations, including HasOne/HasMany/BelongsTo, are broken after I upgraded my loopback4 packages to the latest release.

jeckel:
I just reproduced the same issue after upgrading loopback4

apocaliss92:
[...]

PR #2499, opened 2019-02-28
fix(context): tidy up binding scopes for LB4 constructs to fix issue 2495

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/2495

**Analysis**
1. The Repository is a singleton but it has an injected getter function, which internally references the `request context` that triggers the creation.
2. After the `request context` closes, the getter function stops working.

## Checklist

### DIFF
--- a/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
@@ -5,13 +5,15 @@
 
 import {expect} from '@loopback/testlab';
 import {
+  BindingScope,
   Context,
-  inject,
-  Setter,
   Getter,
-  Provider,
+  inject,
   Injection,
+  invokeMethod,
+  Provider,
   ResolutionSession,
+  Setter,
 } from '../..';
 
 const INFO_CONTROLLER = 'controllers.info';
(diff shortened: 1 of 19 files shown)


bajtos:
>  The `ControllerRoute` binds controller factory as a singleton. As a consequence, controllers become singletons and it's wrong.

As I understand this part, the problem is caused by the fact that singletons are cached by the root context, which typically means application level. What we want is to cache the controller factory at request level.

I'd like us to go deeper here and ask ourselves why did we introduce this bug?

Perhaps the scope values `SINGLETON` and `CONTEXT` don't make it clear what is the distinction and thus it's difficult to spot incorrect scope when reading the code. Can we come up with names that are more descriptive? For example: rename `SINGLETON` to `GLOBAL_SINGLETON` and `CONTEXT` to `LOCAL_SINGLETON`?

Is it actually a good idea to allow binding of global singletons via child context instance? I think we will save a lot of headaches if we required application-level singletons to be bound via the application context, and forced all request-level singletons to stay within request-level context (i.e. force `CONTEXT` scope).

>  The Repository is a singleton but it has an injected getter function, which internally references the `request context` that triggers the creation.

This does not match my understanding of our design & implementation. AFAIK, repositories are transient classes, we create a new repository instance for each request.

As I understand the problem, it's caused by the fact that the **controller** class ends up being created as a global (application-level) singleton, but depends on classes that are in transient request-level scope.

IMO, a global singleton depending on transient dependencies provided by child contexts is an invalid setup that we should reject at binding time.
- We can crawl dependency tree and ensure that that SINGLETON binding depends on SINGLETON dependencies only. CONTEXT bindings may depend on SINGLETON or CONTEXT. This can be too complex though.
- I think a much simpler solution is to reject `SINGLETON` scope when the binding is created on a child context.

As for this pull request, I am proposing to split it into two parts:
1. Fix the scope of controller bindings to quickly address the regression.
2. Improve the context implementation to better handle this special situation.

packages/context/src/context.ts
```diff
@@ -576,7 +575,7 @@ export class Context extends EventEmitter {
    *
    * ```ts
    * // get "rest" property from the value bound to "config"
-   * // use "undefined" when not config was provided
+   * // use "undefined" when not config is provided
```
shazha: caught a typo 😄
```suggestion
   * // use "undefined" when no config is provided
```

raymondfeng:
I'm sorry that I have to take back my false alarm that binding current controller to a request context in `SINGLETON` is wrong. In fact, that's the expected behavior as `SINGLETON` means there is a single instance in the owning context and its descendants and we want to make sure the request context and its child see the same instance.

I'm also fixing the default binding scope for repositories and service proxies to `TRANSIENT`. We already bind data sources as `SINGLETON` and that takes care of shared connections to backend systems. Repositories and service proxies should be able receive request specific contextual data.

packages/context/src/context.ts
raymondfeng: Good catch

raymondfeng:
> IMO, a global singleton depending on transient dependencies provided by child contexts is an invalid setup that we should reject at binding time.

I agreed and it's now fixed in this PR.

> The Repository is a singleton but it has an injected getter function, which internally references the `request context` that triggers the creation.

It's fixed in this PR too, in addition to service proxies. Only `data sources` should be registered as singleton to share connections.

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -71,7 +71,7 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
         name,
         namespace: 'repositories',
         type: 'repository',
-      }).inScope(BindingScope.SINGLETON);
+      });
```
b-admike: by doing this, the default scope is `transient` correct?

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -145,34 +147,132 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+
+    it('throws if a value cannot be resolved by the owning context', async () => {
+      class Store {
+        constructor(@inject('hash') public hash: string) {}
+      }
+
+      // Create a child context of `ctx`
+      const requestCtx = new Context(ctx, 'child');
+      // Bind `hash` to `value`
+      requestCtx.bind('hash').to('123');
```
b-admike: Sorry, what does binding key `has` to `123` in the child context signify to the test?

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -145,34 +147,132 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+
+      // Bind `Store` as a singleton
+      ctx
+        .bind('store')
+        .toClass(Store)
+        .inScope(BindingScope.SINGLETON);
+
+      let value = await invokeMethod(Store, 'getHash', requestCtx);
```
b-admike: Are these two lines (and 257-258) showing how it can be done before the feature to inject method parameters down the context chain?

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -378,7 +373,7 @@ describe('Routing', () => {
         };
       }
     }
-    givenControllerInApp(app, GetCurrentController);
+    givenControllerInApp(app, GetCurrentController, BindingScope.SINGLETON);
```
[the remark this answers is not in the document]
raymondfeng: The binding scope can be set by the framework or by applying `@bind` to the class.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: It verifies that the binding in a child context does not fulfill the dependency for a singleton at parent level.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: It verifies that method parameter injections for singleton are still allowed.

packages/rest/src/rest.component.ts
```diff
@@ -94,6 +60,56 @@ export class RestComponent implements Component {
...
+   * need to be added to the same context.
+   *
+   * @param app Application instance
+   */
+  private bindBodyParsers(app: Application) {
+    // Try to see if there is a RestServer
+    const server = app.getSync<RestServer>('servers.RestServer', {
+      optional: true,
+    });
+    const ctx = server || app;
```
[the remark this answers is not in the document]
raymondfeng: Added. Actually, the RestServer is always there.

packages/service-proxy/src/mixins/service.mixin.ts
```diff
@@ -75,7 +70,7 @@ export function ServiceMixin<T extends Class<any>>(superClass: T) {
         name: serviceName,
         namespace: 'services',
         type: 'service',
-      }).inScope(BindingScope.SINGLETON);
+      });
```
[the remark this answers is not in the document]
bajtos: Please add an acceptance test that uses a service proxy and fails with the current implementation. I am thinking about a test similar to what you wrote for controllers: https://github.com/strongloop/loopback-next/blob/8fc416ceb830a5122b3b0db270b73e9b3d76a6b1/examples/todo-list/src/__tests__/acceptance/todo-list-todo.acceptance.ts#L85-L96

packages/rest/src/__tests__/acceptance/request-parsing/request-parsing.acceptance.ts
```diff
@@ -64,7 +64,7 @@ describe('request parsing', () => {
   });
 
   it('allows built-in body parsers to be removed', async () => {
-    app.unbind(RestBindings.REQUEST_BODY_PARSER_JSON);
+    app.restServer.unbind(RestBindings.REQUEST_BODY_PARSER_JSON);
```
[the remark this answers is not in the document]
b-admike: IIUC the old way still works, and if that's the case, then we can split this in two tests.

packages/rest/src/__tests__/unit/router/controller-route.unit.ts
```diff
@@ -85,6 +86,59 @@ describe('ControllerRoute', () => {
...
+    let appCtx: Context;
+    let requestCtx: Context;
+
+    before(givenContextsAndControllerRoute);
+
+    it('adds bindings to a request context', async () => {
+      expect(requestCtx.contains(CoreBindings.CONTROLLER_CURRENT));
+      expect(
+        requestCtx.getBinding(CoreBindings.CONTROLLER_CURRENT).scope,
+      ).to.equal(BindingScope.SINGLETON);
```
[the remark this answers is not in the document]
b-admike: I thought the `CONTROLLER_CURRENT` key needs to be bound as a SINGLETON so as to make sure that each sequence action works with the same instance of the controller as it goes through the request-response lifecycle of an API call. Maybe I understood it wrong.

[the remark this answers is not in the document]
raymondfeng: `SINGLETON` is really a local singleton for the subtree. For example, with the a context chain `request -> server -> app` and the following setup:

```ts
server.bind('logger').toClass(Logger).inScope(BindingScope.SINGLETON);
```
server.get('logger') and request.get('logger') will always get the same instance of `Logger` as it's cached at ownerContext of `logger`, which is `server`.

raymondfeng:
Close this PR as commits have been refactored into multiple PRs.

PR #2502, opened 2019-02-28
Improve HTTP server config used by tests

bajtos:
#1995 modified the project template to use the following configuration for HTTP server host & port:

```ts
  const config = givenHttpServerConfig();
	
  // Set host to `HOST` env var or ipv4 loopback interface
  // By default, docker container has ipv6 disabled.
  config.host = config.host || process.env.HOST || '127.0.0.1';

  // Set port to `PORT` env var or `0`
  config.port = config.port || +(process.env.PORT || 0);
```

I find such approach problematic.

1. Acceptance tests are sort of self-contained. Both the server and the client are running in the same Node.js process and the server is running only for the duration of the test suite. There is no need to expose the server to public by honoring HOST/PORT environment variables.

2. By honoring HOST/PORT environment variables, we can actually break the test suite for people that are not running inside docker, because there may be already another process listening on that address.

3. I find the configuration rather complex. By placing it into a project template, we are forcing this complexity on all framework users. There are also no tests to verify correctness of the implementation.

In this pull request, I am proposing a different approach:

- The first commit modifies `givenHttpServerConfig` to always use `127.0.0.1` as the host. I believe this should fix the problem with Docker IPv6 setup as described in #1995. 

- `givenHttpServerConfig` is also changed to ignore empty values in custom host/port config, to make the customization even easier.

- The second commit modifies the project template to run the acceptance tests with the default config provided by `givenHttpServerConfig`. The template also contains a code comment showing users how they can customize the host & port values.

## Checklist

### DIFF
--- a/packages/testlab/src/http-server-config.ts
+++ b/packages/testlab/src/http-server-config.ts
@@ -6,14 +6,20 @@
 /**
  * Create an HTTP-server configuration that works well in test environments.
  *  - Ask the operating system to assign a free (ephemeral) port.
- *  - Use IPv4 localhost `127.0.0.1` on Travis-CI to avoid known IPv6 issues.
+ *  - Use IPv4 localhost `127.0.0.1` to avoid known IPv6 issues in Docker-based
+ *    environments like Travis-CI.
  *
  * @param customConfig Additional configuration options to apply.
  */
 export function givenHttpServerConfig<T extends object>(
   customConfig?: T,
-): T & {host?: string; port: number} {
-  const defaults = {port: 0};
-  const hostConfig = process.env.TRAVIS ? {host: '127.0.0.1'} : {};
-  return Object.assign(defaults, hostConfig, customConfig);
+): T & {host: string; port: number} {
+  const defaults = {
+    host: '127.0.0.1',
+    port: 0,
+  };
+  const config = Object.assign({}, defaults, customConfig);
+  if (config.host === undefined) config.host = defaults.host;
+  if (config.port === undefined) config.port = defaults.port;
+  return config;
 }
--- a/packages/cli/generators/app/templates/src/__tests__/acceptance/test-helper.ts.ejs
+++ b/packages/cli/generators/app/templates/src/__tests__/acceptance/test-helper.ts.ejs
@@ -6,17 +6,16 @@ import {
 } from '@loopback/testlab';
 
 export async function setupApplication(): Promise<AppWithClient> {
-  const config = givenHttpServerConfig();
-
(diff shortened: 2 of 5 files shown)


packages/testlab/src/http-server-config.ts
```diff
@@ -6,14 +6,20 @@
...
  */
 export function givenHttpServerConfig<T extends object>(
   customConfig?: T,
-): T & {host?: string; port: number} {
-  const defaults = {port: 0};
-  const hostConfig = process.env.TRAVIS ? {host: '127.0.0.1'} : {};
-  return Object.assign(defaults, hostConfig, customConfig);
+): T & {host: string; port: number} {
+  const defaults = {
+    host: '127.0.0.1',
```
raymondfeng: This is problematic as it pretty much excludes our test coverage on ipv6 unless we explicitly add more tests for both ipv4 and ipv6.

bajtos:
> This is problematic as it pretty much excludes our test coverage on ipv6 unless we explicitly add more tests for both ipv4 and ipv6.

Good point, let's discuss.

My thoughts:

- We should be explicit about what is being tested. At the moment, we are using IPv6 when running the tests locally and on AppVeyor; we use IPv4 when testing on Travis CI. This is not obvious from our test code though.
- By using different IP version in different environments, we can easily end up with test failures that are difficult to troubleshoot. A test failing on Travis may pass locally.
- We should verify IPv6 functionality in our test suite for `@loopback/http-server` package, there is no need to test it again and again in every integration/acceptance test. Here is a test we already have in place: https://github.com/strongloop/loopback-next/blob/fed9956d760426766190b9c53cf7aaeed12d29f1/packages/http-server/src/__tests__/integration/http-server.integration.ts#L175-L183
- In my experience, most applications are still running on IPv4. I think `127.0.0.1` is a reasonable default that will work well for majority of our users.
- `givenHttpServerConfig` is an optional helper, people building IPv6 application can remove it and supply their own server config, e.g. `{host: '::1', port: 0}`.

In a nutshell, I'd like to treat `givenHttpServerConfig` as a test helper making life easier for majority of our users and simplify our test suite to be more consistent & predictable.

@raymondfeng What do you think?

bajtos:
Rebased on top of the latest master and added two more tests in c9dbb3e

raymondfeng:
Please verify with a docker image for LB4 app scaffolded with `--docker`.

bajtos:
> Please verify with a docker image for LB4 app scaffolded with --docker.

I have verified that both `node .` and `npm test` work inside Docker.

@raymondfeng LGTY now?

PR #2505, opened 2019-02-28
fix(repository): support embedded models for nested properties 

### LINKED ISSUE #2130: Support embedded models for nested properties
This issue was discussed at [stackoverflow](https://stackoverflow.com/questions/53554400/exception-in-deserialization-of-nested-objects-for-datamodels-in-loopback-4/). 

An error occurs on create if nested entities are defined. 

## Current Behavior

on create of an entity with a nested child (or more), an error occurs: 

> Unhandled error in POST /myMainEntity: 500 TypeError: Class constructor MySubEntity cannot be invoked without 'new'

if the nested entity was included like this:
```
@property({
    type: MySubEntity, // maybe in aphostrope
  })
  subEntity?: MySubEntity;
```

## Expected Behavior
The objective of this was to get nested entities, which
- are type safe (checks on the properties of the sub object; autocomplete on the sub-object; swagger shows the correct sub-object; ...)
- are invoked correctly with their containing datatypes (e.g. transformation from String to Date if the property is defined as a date)

_Even more helpful would be to use jsonSchema for the datamodel and get the described features like autocomplete, typesafety if this is possible anyhow._

IsCaster:
Fixes #2130

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -10,7 +10,9 @@ import {
   Entity,
   EntityNotFoundError,
   juggler,
+  model,
   ModelDefinition,
+  property,
 } from '../../..';
 
 describe('legacy loopback-datasource-juggler', () => {
@@ -120,6 +122,74 @@ describe('DefaultCrudRepository', () => {
       expect(ShoppingList.definition.properties).to.containDeep(
         originalPropertyDefinition,
       );
+    });
+
+    it('converts PropertyDefinition with model type', () => {
+      @model()
+      class Role {
+        @property()
+        name: String;
+      }
+
+      @model()
+      class Address {
+        @property()
+        street: String;
+      }
+
+      @model()
+      class User extends Entity {
+        @property({
+          type: 'number',
+          id: true,
+        })
+        id: number;
+
+        @property({type: 'string'})
+        name: String;
+
+        @property.array(Role)
+        roles: Role[];
+
+        @property()
+        address: Address;
+      }
+
+      expect(ds.getModel('User')).undefined();
+
+      // tslint:disable-next-line:no-unused-expression
+      new DefaultCrudRepository(User, ds);
+
+      const JugglerUser = ds.getModel('User')!;
+      expect(JugglerUser).to.be.a.Function();
+
+      const addressProperty = JugglerUser.definition.properties.address;
+      const addressModel = addressProperty.type as typeof juggler.ModelBase;
+      expect(addressModel).to.be.a.Function();
+      expect(addressModel).to.equal(ds.getModel('Address'));
+
+      expect(addressModel.name).to.equal('Address');
+      expect(addressModel.definition).to.containDeep({
+        name: 'Address',
+        properties: {street: {type: String}},
+      });
+
+      const rolesProperty = JugglerUser.definition.properties.roles;
+      expect(rolesProperty.type)
+        .to.be.an.Array()
+        .of.length(1);
+
+      // FIXME(bajtos) PropertyDefinition in juggler does not allow array type!
+      // tslint:disable-next-line:no-any
+      const rolesModel = (rolesProperty.type as any)[0] as typeof juggler.ModelBase;
+      expect(rolesModel).to.be.a.Function();
+      expect(rolesModel).to.equal(ds.getModel('Role'));
+
+      expect(rolesModel.name).to.equal('Role');
+      expect(rolesModel.definition).to.containDeep({
+        name: 'Role',
+        properties: {name: {type: String}},
+      });
     });
   });
 
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -16,20 +16,20 @@ import {
   PositionalParameters,
 } from '../common-types';
 import {EntityNotFoundError} from '../errors';
-import {Entity, ModelDefinition} from '../model';
+import {Entity, Model, PropertyType} from '../model';
 import {Filter, Where} from '../query';
 import {
+  BelongsToAccessor,
   BelongsToDefinition,
+  createBelongsToAccessor,
+  createHasManyRepositoryFactory,
+  createHasOneRepositoryFactory,
   HasManyDefinition,
   HasManyRepositoryFactory,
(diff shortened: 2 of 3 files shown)


packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -12,6 +12,8 @@ import {
   juggler,
   ModelDefinition,
 } from '../../..';
+import {Model} from '../../../model';
+import {property, model} from '../../../decorators';
```
bajtos: Let's keep the imports simple and import everything from `../../..`.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -93,26 +97,53 @@ export class DefaultCrudRepository<T extends Entity, ID>
   ) {
     const definition = entityClass.definition;
     assert(
-      !!definition,
-      `Entity ${entityClass.name} must have valid model definition.`,
```
bajtos: Please preserve this assert and add your new assertion as a new `assert` call.

When `definition` is undefined, your check `definition.idProperties().length > 0` will crash on trying to access `idProperties` on an undefined object.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -93,26 +97,53 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+      definition.idProperties().length > 0,
+      `Entity ${entityClass.name} must have at least one id/pk property.`,
     );
 
+    this.setupModel(entityClass);
+    this.setModelClass();
+  }
+  private setupModel(entityClass: typeof Model & {prototype: T}) {
```
bajtos: We are adding an empty line between methods to visually delimit them.

```ts
}

private setupModel(entityClass: typeof Model & {prototype: T}) {
```

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -93,26 +97,53 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+    const definition = entityClass.definition;
+
     assert(
-      definition.idProperties().length > 0,
-      `Entity ${entityClass.name} must have at least one id/pk property.`,
+      !!definition,
```
bajtos: As I wrote above, this check should be run as the first one.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -93,26 +97,53 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+      }
+    }
+
+    this.attachModelToDataSource(definition);
+  }
+
+  private setModelClass() {
+    const model = this.dataSource.getModel(this.entityClass.definition.name);
+    if (model) {
+      // The backing persisted model has been already defined.
```
bajtos: What if it has not been registered? 

I think the following code may work well:

```ts
const model = this.dataSource.getModel(this.entityClass.definition.name);
assert(!!model, 'The backing persisted model should have been defined by setupModel()');
this.modelClass = model as typeof juggler.PersistedModel;
```

Even better, can we modify `setupModel` to return the PersistedModel class created? The we could change the constructor as follows:

```ts
this.modelClass = this.setupModel(entityClass);
```

I would also rename `setupModel` to something like `definePersistedModel`.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -93,26 +97,53 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
   }
 
   // Create an internal legacy Model attached to the datasource
-  private setupPersistedModel(definition: ModelDefinition) {
+  private attachModelToDataSource(definition: ModelDefinition) {
     const dataSource = this.dataSource;
 
     const model = dataSource.getModel(definition.name);
     if (model) {
-      // The backing persisted model has been already defined.
```
bajtos: Please preserve this comment.

packages/repository/src/repositories/legacy-juggler-bridge.ts
IsCaster: This check also runs when it's a embedded model.

IsCaster:
@bajtos @raymondfeng Hi, I had already changed the code. Could you merge it, please?

bajtos:
@IsCaster sorry for the delays. I'll try to take a look at the proposed changes again by the end of this week.

bajtos:
As I was reviewing the changes in more details, I realize we need to improve the implementation and decided it will be faster for me to make the changes rather than try to explain them. See 45e88ce.

I have also rebased the pull request on top of the latest master.

@IsCaster please take a look, are you ok with the changes I made?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -124,14 +141,15 @@ export class DefaultCrudRepository<T extends Entity, ID>
     // the juggler understands
     Object.entries(definition.properties).forEach(([key, value]) => {
       if (value.type === 'array' || value.type === Array) {
-        value = Object.assign({}, value, {type: [resolveType(value.itemType)]});
+        value = Object.assign({}, value, {
+          type: [value.itemType && this.resolvePropertyType(value.itemType)],
```
bajtos: We want juggler to see the new persisted model in the property definition, not a reference to LB4 model.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -124,14 +141,15 @@ export class DefaultCrudRepository<T extends Entity, ID>
...
+        });
         delete value.itemType;
       }
-      value.type = resolveType(value.type);
+      value.type = this.resolvePropertyType(value.type);
```
bajtos: Similarly here: We want juggler to see the new persisted model in the property definition, not a reference to LB4 model.

IsCaster:
> As I was reviewing the changes in more details, I realize we need to improve the implementation and decided it will be faster for me to make the changes rather than try to explain them. See [45e88ce](https://github.com/strongloop/loopback-next/commit/45e88ce994ed734988ebecc34fc9ff72fd9f440c).
> 
> I have also rebased the pull request on top of the latest master.
> 
> @IsCaster please take a look, are you ok with the changes I made?

Ok with me

bajtos:
Landed, thank you @IsCaster for the contribution ❤️

PR #2512, opened 2019-03-01
feat(rest): add support for redirect routes

bajtos:
Implement a new class RedirectRoute to provide redirection from a given URL path to a new URL path.

Add RestServer and RestApplication methods for defining redirect routes:

```ts
redirect(fromPath: string, toPathOrUrl: string, statusCode?: number);
```

This pull request supersedes #2314, see also #2022

## Checklist

### DIFF
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -42,6 +42,7 @@ import {
   RouteEntry,
   RoutingTable,
   StaticAssetsRoute,
+  RedirectRoute,
 } from './router';
 import {DefaultSequence, SequenceFunction, SequenceHandler} from './sequence';
 import {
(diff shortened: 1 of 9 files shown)


packages/rest/src/rest.server.ts
```diff
@@ -666,6 +667,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+   * ```ts
+   * server.redirect('/explorer', '/explorer/');
+   * ```
+   *
+   * @param source URL path of the redirect endpoint
+   * @param target Location (URL path) where to redirect to
+   * @param statusCode HTTP status code to respond with,
+   *   defaults to 303 (See Other).
+   */
+  redirect(source: string, target: string, statusCode?: number): Binding {
```
hacksparrow: The names `source` and `target` sound a little weird. Maybe `fromUrl` and `toUrl` is more descriptive?

packages/rest/src/router/redirect-route.ts
```diff
@@ -0,0 +1,42 @@
...
+  readonly path: string = this.sourcePath;
+  readonly spec: OperationObject = {
+    description: 'LoopBack Redirect route',
+    'x-visibility': 'undocumented',
+    responses: {},
+  };
+
+  constructor(
+    public readonly sourcePath: string,
+    public targetPath: string,
```
raymondfeng: private?

packages/rest/src/router/redirect-route.ts
```diff
@@ -0,0 +1,42 @@
...
+    public statusCode: number = 303,
```
raymondfeng: private?

packages/rest/src/rest.server.ts
bajtos: Good point. Will use the following names:

```ts
class RestServer {
  //...

  redirect(
    fromPath: string,
    toPathOrUrl: string,
    statusCode?: number,
  ): Binding;
}
```

PR #2515, opened 2019-03-01
feat(context): tidy up dependency injection for singleton bindings

raymondfeng:
Spin-off from #2499 

## Checklist


raymondfeng:
FYI: The following diagram illustrates LoopBack 4's binding scopes.

image

bajtos:
Re-posting from our online chat:

Let's put this diagram to our docs please.

Could you also enhance the table to show where is the resolved value cached? I expect the place will be different for different rows in your table.

raymondfeng:
See https://github.com/strongloop/loopback-next/blob/fix-2495-phase-2/docs/site/Dependency-injection.md#dependency-injection-for-bindings-with-different-scopes

docs/site/Dependency-injection.md
```diff
@@ -301,9 +301,118 @@ Error: Circular dependency detected:
...
+```ts
+import {inject, Context, BindingScope} from '@loopback/context';
+import {RestBindings} from '@loopback/rest';
+
+interface Logger() {
+  log(message: string);
+}
+
+class PingController {
+  constructor(@inject('logger') private logger: Logger) {}
```
jannyHou: when injecting the logger using same name `logger`, how would the resolver distinguish between controller's logger and service's logger? According to the context?

raymondfeng: Yes, it’s determined by the resolution context, the owner context, and the binding scope.

raymondfeng:
@bajtos Please review.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -145,34 +148,181 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      class Store {
+        constructor(@inject.getter('hash') public getter: Getter<string>) {}
+      }
+
+      givenRequestContext(Store);
+
+      // Create the store instance from the child context
+      // Now the singleton references a `getter` to `hash` binding from
+      // the `requestCtx`
+      const store = requestCtx.getSync<Store>('store');
```
bajtos: Where is the binding key `store` coming from?

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -145,34 +148,181 @@ describe('Context bindings - Injecting dependencies of classes', () => {
...
+      store.setHash('456');
+
+      // For singleton bindings, the injected setter will set value to the owner
+      // context (`ctx`) instead of the current one (`requestCtx`) used to
+      // resolve the bound `Store`
+      expect(await ctx.get<string>('hash')).to.equal('456');
+      expect(await requestCtx.get<string>('hash')).to.equal('123');
+    });
+
+    function givenRequestContext(storeClass: Constructor<unknown>) {
```
bajtos: The name does not tell the whole store about what the helper does. I am proposing the following improvement:

```suggestion
    function givenRequestContextWithHashAndStore(storeClass: Constructor<unknown>) {
```

packages/rest/src/__tests__/acceptance/request-parsing/request-parsing.acceptance.ts
```diff
@@ -64,7 +64,7 @@ describe('request parsing', () => {
   });
 
   it('allows built-in body parsers to be removed', async () => {
-    app.unbind(RestBindings.REQUEST_BODY_PARSER_JSON);
+    app.restServer.unbind(RestBindings.REQUEST_BODY_PARSER_JSON);
```
bajtos: This looks like a breaking change to me. Can we find a way how to keep body parsers bound at application level?

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -361,30 +356,80 @@ describe('Routing', () => {
...
+        .get('/name')
+        .expect({
+          count: 1, // The count is still 1 as it's from a new instance
+          isSingleton: false,
+          result: true,
+        });
+    });
+
+    it('binds current controller resolved from a singleton binding', async () => {
+      givenControllerAsSingletonInApp(app, controllerClass);
```
bajtos: When reading the code for the first time, I find the overhead of reading yet another helper implementation as too much. Have you considered the following?

```suggestion
      givenControllerInApp(app, controllerClass).inScope(BindingScope.SINGLETON)
```

Considering how little is `givenControllerInApp` doing, I would get rid of it entirely.

```suggestion
      app.controller(controllerClass).inScope(BindingScope.SINGLETON)
```

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
```diff
@@ -361,30 +356,80 @@ describe('Routing', () => {
...
+        .expect({
+          count: 2, // The count increases as the controller is singleton
+          isSingleton: true,
+          result: true,
+        });
+    });
+
+    async function setupController() {
+      app = givenAnApplication();
+      server = await givenAServer(app);
```
bajtos: The helper says `setupController`, but under the hood it's also setting up `app` and `server`. I find that very confusing!

I think you should split this helper into two parts:

- `setupServerInApp`
- `setupController`

We use `given` instead of `setup` prefix in other places, `givenServerInApp` and `givenController` are probably more consistent names.

packages/rest/src/__tests__/unit/router/controller-route.unit.ts
```diff
@@ -85,6 +86,59 @@ describe('ControllerRoute', () => {
     expect(route._controllerName).to.eql('my-controller');
   });
 
+  describe('updateBindings()', () => {
+    let appCtx: Context;
+    let requestCtx: Context;
+
+    before(givenContextsAndControllerRoute);
```
bajtos: Great name for the setup helper 👍

packages/rest/src/__tests__/unit/router/controller-route.unit.ts
```diff
@@ -85,6 +86,59 @@ describe('ControllerRoute', () => {
...
+
+    it('adds bindings to a request context', async () => {
```
bajtos: Since there is always only one request context for every `updateBindings` call, I think `the request context` is more grammatically more correct than `a request context`?

```suggestion
    it('adds bindings to the request context', async () => {
```

packages/rest/src/rest.application.ts
```diff
@@ -87,7 +87,6 @@ export class RestApplication extends Application implements HttpServerLike {
    * See https://expressjs.com/en/4x/api.html#express.static
    * @param path The path(s) to serve the asset.
    * See examples at https://expressjs.com/en/4x/api.html#path-examples
-   * To avoid performance penalty, `/` is not allowed for now.
```
bajtos: The changes in this file look very unrelated to the rest of this pull request. Could you please move them to a new PR? Or at least to a standalone commit.

packages/rest/src/rest.component.ts
```diff
@@ -94,6 +60,53 @@ export class RestComponent implements Component {
       Object.assign(apiSpec, {servers: config.openApiSpec.servers});
     }
     app.bind(RestBindings.API_SPEC).to(apiSpec);
+    app.server(RestServer);
+    this.bindBodyParsers(app);
```
bajtos: 1. As I commented earlier, I believe this is a backwards-incompatible change that requires a semver-major release.

2. Isn't this particular implementation a step backwards? Before, we used declarative approach to export servers and bindings. Now we are moving to imperative style.

packages/rest/src/rest.component.ts
```diff
@@ -94,6 +60,53 @@ export class RestComponent implements Component {
...
+
+  /**
+   * Bind body parsers to the `RestServer`. This is needed because
+   * we bind `RequestBodyParser` as a singleton so that all body parser bindings
+   * need to be added to the same context.
+   *
+   * @param app Application instance
+   */
+  private bindBodyParsers(app: Application) {
+    const server = app.getSync<RestServer>(
```
bajtos: There may be more than a single (REST) server bound in the application. Before your change, all RestServer instance were obtaining the default list of built-in body parsers. With your change in place, non-default RestServers will end up with an empty list.

I feel we need to spend more time looking into different approaches for registrations of body parsers. I think it's a more generic problem that can be phrased as follows:
 - How to allow applications to configure bindings that are shared by all server instances?
 - How to allow applications to override list of bindings for a certain tag so that a selected server instance receives a different list of bindings (e.g. body parsers)?

To move your pull request forward and finish the cleanup of DI for singleton bindings, I am proposing to revert your changes related to body-parser configuration. To fix the problem surfaced by your cleanup, I am proposing to change the scope of body parsers from SINGLETON to TRANSIENT. IMO, body parsers should be lightweight classes that are cheap to instantiate, thus it should be ok to use transient scope for them.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: I extract the binding key to be constants.

packages/rest/src/rest.component.ts
raymondfeng: I reverted body parser changes and set the scope to be `TRANSIENT`.

packages/rest/src/rest.component.ts
raymondfeng: Reverted

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
bajtos: > Considering how little is `givenControllerInApp` doing, I would get rid of it entirely.

☝️

packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts
bajtos: > The helper says setupController, but under the hood it's also setting up app and server. I find that very confusing!
> 
> I think you should split this helper into two parts

☝️

PR #2517, opened 2019-03-01
fix(boot): improve booter bindings

raymondfeng:
The PR introduces two minor enhancements to booter class bindings.

1. Honor `@bind` for booter classes
2. Bind booter classes using `SINGLETON` scope by default

## Checklist

### DIFF
--- a/packages/boot/src/mixins/boot.mixin.ts
+++ b/packages/boot/src/mixins/boot.mixin.ts
@@ -8,6 +8,7 @@ import {
   Constructor,
   Context,
   createBindingFromClass,
+  BindingScope,
 } from '@loopback/context';
 import {BootComponent} from '../boot.component';
 import {Bootstrapper} from '../bootstrapper';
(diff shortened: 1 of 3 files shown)


packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -134,9 +141,11 @@ export function _bindBooter(
...
-  return ctx
-    .bind(`${BootBindings.BOOTER_PREFIX}.${booterCls.name}`)
-    .toClass(booterCls)
-    .inScope(BindingScope.CONTEXT)
-    .tag(BootBindings.BOOTER_TAG);
+  const binding = createBindingFromClass(booterCls, {
+    namespace: BootBindings.BOOTER_PREFIX,
+  })
+    .tag(BootBindings.BOOTER_TAG)
+    .inScope(BindingScope.SINGLETON);
```
bajtos: Hmm, this is making multiple changes in a way that makes it difficult to spot them:
- change the scope from CONTEXT to SINGLETON
- start honouring `@bind` decorator by calling `createBindingFromClass`, at least I think that's the case?

Let's split these changes into two commits please, if not two pull requests.

What's more important, please add tests to verify these changes and prevent regressions in the future.

It would be ideal to have an acceptance-level test that fails when a booter is bound in CONTEXT scope.

raymondfeng: Split the PR into two commits and added corresponding tests.

raymondfeng:
PTAL.

PR #2520, opened 2019-03-01
feat(context): honor binding scope from @bind

raymondfeng:
The PR improves binding attributes as follows:

- Refine visibility for binding attributes
- Introduce `defaultScope` to only override binding scope if not set

Use case:

If a class is decorated with `@bind({scope: BindingScope.SINGLETON})`, we should honor it and not override it with a default scope. For example:

```
@bind({scope: BindingScope.SINGLETON})
class MyController {}

app.controller(MyController); // My controller should be bound as SINGLETON
```

## Checklist

### DIFF
--- a/packages/context/src/__tests__/acceptance/bind-decorator.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/bind-decorator.acceptance.ts
@@ -72,4 +72,42 @@ describe('@bind - customize classes with binding attributes', () => {
       'controllers.my-controller',
     ]);
   });
+
+  it('supports default binding scope in options', () => {
+    const binding = createBindingFromClass(MyController, {
+      defaultScope: BindingScope.SINGLETON,
+    });
+    expect(binding.scope).to.equal(BindingScope.SINGLETON);
+  });
+
+  describe('binding scope', () => {
+    @bind({
+      // Explicitly set the binding scope to be `SINGLETON` as the developer
+      // choose to implement the controller as a singleton without depending
+      // on request specific information
+      scope: BindingScope.SINGLETON,
+    })
+    class MySingletonController {}
+
+    it('allows singleton controller with @bind', () => {
+      const binding = createBindingFromClass(MySingletonController, {
+        type: 'controller',
+      });
+      expect(binding.key).to.equal('controllers.MySingletonController');
+      expect(binding.tagMap).to.containEql({controller: 'controller'});
+      expect(binding.scope).to.equal(BindingScope.SINGLETON);
+    });
+
+    it('honors binding scope from @bind over defaultScope', () => {
+      let binding = createBindingFromClass(MySingletonController, {
+        defaultScope: BindingScope.TRANSIENT,
+      });
+      expect(binding.scope).to.equal(BindingScope.SINGLETON);
+    });
+
+    it('honors binding scope from @bind', () => {
+      const binding = createBindingFromClass(MySingletonController);
+      expect(binding.scope).to.equal(BindingScope.SINGLETON);
+    });
+  });
 });
--- a/packages/context/src/binding-inspector.ts
+++ b/packages/context/src/binding-inspector.ts
@@ -136,7 +136,9 @@ export function removeNameAndKeyTags(binding: Binding<unknown>) {
  *
  * @param cls A class with optional `@bind`
  */
-export function bindingTemplateFor(cls: Constructor<unknown>): BindingTemplate {
+export function bindingTemplateFor<T = unknown>(
+  cls: Constructor<T | Provider<T>>,
+): BindingTemplate<T> {
   const spec = getBindingMetadata(cls);
   const templateFunctions = (spec && spec.templates) || [
     asClassOrProvider(cls),
@@ -198,16 +200,16 @@ export type BindingFromClassOptions = {
  * @param cls A class
  * @param options Options to customize the binding key
  */
-export function createBindingFromClass(
-  cls: Constructor<unknown>,
+export function createBindingFromClass<T = unknown>(
+  cls: Constructor<T | Provider<T>>,
   options: BindingFromClassOptions = {},
-): Binding {
+): Binding<T> {
   const templateFn = bindingTemplateFor(cls);
   let key = options.key;
   if (!key) {
     key = buildBindingKey(cls, options);
   }
-  const binding = Binding.bind(key).apply(templateFn);
+  const binding = Binding.bind<T>(key).apply(templateFn);
   if (options.name) {
     binding.tag({name: options.name});
   }
--- a/packages/context/src/binding-inspector.ts
+++ b/packages/context/src/binding-inspector.ts
@@ -193,11 +193,20 @@ export type BindingFromClassOptions = {
    * Mapping artifact type to binding key namespaces
    */
   typeNamespaceMapping?: TypeNamespaceMapping;
+  /**
+   * Default scope if the binding does not have an explicit scope
+   */
+  defaultScope?: BindingScope;
 };
 
 /**
- * Create a binding from a class with decorated metadata
- * @param cls A class
+ * Create a binding from a class with decorated metadata. The class is attached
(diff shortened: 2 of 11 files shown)


bajtos:
I am overloaded with PR reviews now, I'll try to find some time to review these changes later this week.

packages/context/src/__tests__/acceptance/bind-decorator.acceptance.ts
```diff
@@ -72,4 +78,20 @@ describe('@bind - customize classes with binding attributes', () => {
       'controllers.my-controller',
     ]);
   });
+
+  it('honors binding scope from @bind', () => {
+    let binding = createBindingFromClass(MySingletonController, {
+      defaultScope: BindingScope.TRANSIENT,
```
jannyHou: nitpick: it would be good to mention the precedence in Binding's doc: https://loopback.io/doc/en/lb4/Binding.html#configure-binding-attributes-for-a-class

raymondfeng: Added.

packages/context/src/__tests__/acceptance/bind-decorator.acceptance.ts
```diff
@@ -35,6 +35,12 @@ describe('@bind - customize classes with binding attributes', () => {
   })
   class MyController {}
 
+  @bind({
+    scope: BindingScope.SINGLETON,
+    tags: ['controller', {name: 'my-singleton-controller', type: 'controller'}],
+  })
+  class MySingletonController {}
```
bajtos: What is the minimal `@bind()` configuration needed to create a singleton controller? Would the following work?

```ts
@bind({scope: BindingScope.SINGLETON})
class MySingletonController {}
```

If it is intended to work, could you please add a test that will add such controller class via `app.controller` API and verify that name, type, namespace, etc. was set correctly?

packages/service-proxy/src/__tests__/unit/mixin/service.mixin.unit.ts
```diff
@@ -78,7 +78,7 @@ describe('ServiceMixin', () => {
     const boundServices = myApp.find('services.*').map(b => b.key);
     expect(boundServices).to.containEql('services.GeocoderService');
     const binding = myApp.getBinding('services.GeocoderService');
-    expect(binding.scope).to.equal(BindingScope.TRANSIENT);
+    expect(binding.scope).to.be.undefined();
```
bajtos: This looks like a backwards-incompatible change to me.

I think we should use a different approach:

- Keep the change setting the value of`binding._scope` to `undefined`
- Modify `binding.scope` getter to convert `undefined to `TRANSIENT`
- Add a new public method for applying the default scope, something along the following lines.

```ts
class Binding {
 // ..

  applyDefaultScope(scope: BindingScope) {
    if (this._scope) return;
    this._scope = scope;
  }
}
```

That will give use two Binding methods:
 - `binding.inScope` forces the scope the the value provided
 - `binding.applyDefaultScope` preserves explicitly configured scope and applies the given scope only when no explicit scope was defined

raymondfeng: Good idea.

packages/core/src/application.ts
```diff
@@ -50,6 +50,7 @@ export class Application extends Context {
       name,
       namespace: 'controllers',
       type: 'controller',
+      defaultScope: BindingScope.TRANSIENT,
```
bajtos: Do we have a test to verify this behavior?

Also since TRANSIENT is the default scope anyways, is it necessary to explicitly configure it here again? I am fine either way, just asking.

packages/core/src/application.ts
```diff
@@ -80,7 +81,8 @@ export class Application extends Context {
       name,
       namespace: CoreBindings.SERVERS,
       type: 'server',
-    }).inScope(BindingScope.SINGLETON);
+      defaultScope: BindingScope.SINGLETON,
```
bajtos: Please add a test to verify this change.

I think this is the scenario:
 - A server  class is configured with TRANSIENT scope
 - The class is added to an app via `app.server` API
 - The server binding is expected to have TRANSIENT scope

packages/core/src/application.ts
```diff
@@ -196,7 +198,8 @@ export class Application extends Context {
       name,
       namespace: 'components',
       type: 'component',
-    }).inScope(BindingScope.SINGLETON);
+      defaultScope: BindingScope.SINGLETON,
```
bajtos: Ditto - please add a test to verify this fix.

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -71,6 +71,7 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
         name,
         namespace: 'repositories',
         type: 'repository',
+        defaultScope: BindingScope.TRANSIENT,
```
bajtos: Do we have a test to verify this behavior?

Also since TRANSIENT is the default scope anyways, is it necessary to explicitly configure it here again? I am fine either way, just asking.

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -122,7 +123,8 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
           name: name || dataSource.dataSourceName,
           namespace: 'datasources',
           type: 'datasource',
-        }).inScope(BindingScope.SINGLETON);
+          defaultScope: BindingScope.SINGLETON,
```
bajtos: Please add a test to verify this fix.

packages/service-proxy/src/__tests__/unit/mixin/service.mixin.unit.ts
```diff
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Application, Component, Provider, BindingScope} from '@loopback/core';
+import {Application, BindingScope, Component, Provider} from '@loopback/core';
```
bajtos: Since there are no other changes in this file, I think it would be better to revert this reorder of imports.

packages/core/src/application.ts
raymondfeng: I wanted to make it explicit.

packages/core/package.json
```diff
@@ -12,7 +12,7 @@
     "clean": "lb-clean loopback-core*.tgz dist package api-docs",
     "pretest": "npm run build",
     "integration": "lb-mocha \"dist/__tests__/integration/**/*.js\"",
-    "test": "lb-mocha \"dist/__tests__/unit/**/*.js\" \"dist/__tests__/integration/**/*.js\" \"dist/__tests__/acceptance/**/*.js\"",
+    "test": "lb-mocha \"dist/__tests__/**/*.js\"",
```
bajtos: This is unrelated to this pull request.

I don't have a strong opinion on how to specify test files. However, I would like us to be consistent in all packages, including our examples and CLI project templates.

Please revert this change here and open a new pull request to address the problem in a wider context.

packages/context/src/binding-inspector.ts
```diff
@@ -191,29 +193,36 @@ export type BindingFromClassOptions = {
...
   typeNamespaceMapping?: TypeNamespaceMapping;
+  /**
+   * Default scope if the binding does not have an explicit scope
+   */
+  defaultScope?: BindingScope;
 };
 
 /**
  * Create a binding from a class with decorated metadata
  * @param cls A class
```
bajtos: Ditto.

For example:

```suggestion
  * @param cls A class constructor (e.g. a Controller class) or a value provider (Provider class)
```

packages/core/package.json
raymondfeng: See https://github.com/strongloop/loopback-next/pull/2597

PR #2529, opened 2019-03-04
[WIP] feat(rest) : redirect route accept function

### LINKED ISSUE #2022: Sugar API for defining URL redirects

sanadHaj:
Implements a part of #2022
This is a follow-up for #2512

change redirect method signature to accept function , 
the RedirectRoute class check if target is a function and call it else redirect the path .
with adding some functions to extract basePath,host,protocol from request . 
delete basePath from server and added it later on Redirect-route to avoid if statement in redirect method on server class . 

## Checklist

### DIFF
--- a/packages/rest/src/rest.application.ts
+++ b/packages/rest/src/rest.application.ts
@@ -244,7 +244,7 @@ export class RestApplication extends Application implements HttpServerLike {
    */
   redirect(
     fromPath: string,
-    toPathOrUrl: string,
+    toPathOrUrl: string | Function,
     statusCode?: number,
   ): Binding {
     return this.restServer.redirect(fromPath, toPathOrUrl, statusCode);
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -683,11 +683,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
    */
   redirect(
     fromPath: string,
-    toPathOrUrl: string,
+    toPathOrUrl: string | Function,
     statusCode?: number,
   ): Binding {
     return this.route(
-      new RedirectRoute(fromPath, this._basePath + toPathOrUrl, statusCode),
+      new RedirectRoute(fromPath, toPathOrUrl, statusCode),
     );
   }
 
--- a/packages/rest/src/router/redirect-route.ts
+++ b/packages/rest/src/router/redirect-route.ts
@@ -2,7 +2,7 @@ import {RouteEntry, ResolvedRoute} from '.';
 import {RequestContext} from '../request-context';
 import {OperationObject, SchemasObject} from '@loopback/openapi-v3-types';
 import {OperationArgs, OperationRetval, PathParameterValues} from '../types';
-
+import {Request} from "express";
 export class RedirectRoute implements RouteEntry, ResolvedRoute {
   // ResolvedRoute API
   readonly pathParams: PathParameterValues = [];
@@ -19,15 +19,22 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
 
   constructor(
     private readonly sourcePath: string,
-    private readonly targetLocation: string,
+    private readonly targetLocation: string | Function,
     private readonly statusCode: number = 303,
   ) {}
 
   async invokeHandler(
-    {response}: RequestContext,
+    {response,request}: RequestContext,
     args: OperationArgs,
   ): Promise<OperationRetval> {
-    response.redirect(this.statusCode, this.targetLocation);
+    let basePath = this._getBasePath(request) || '';
+    let targetLocation :string = typeof this.targetLocation === 'function'
+        ? this.targetLocation(
+          this._getProtocolForRequest(request),
+          this._getHost(request),
+          basePath
+      ) : basePath + this.targetLocation;
+    response.redirect(this.statusCode,targetLocation);
   }
 
   updateBindings(requestContext: RequestContext) {
@@ -39,4 +46,24 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
       this.targetLocation
     }"`;
   }
+
+  private _getHost(request: Request) {
+      return request.get('x-forwarded-host') || request.headers.host;
+  }
+
+  private _getProtocolForRequest(request: Request) {
+      return (
+          (request.get('x-forwarded-proto') || '').split(',')[0] ||
+          request.protocol ||
+          'http'
+      );
+  }
+
+   private _getBasePath(request: Request) {
+        let basePath = '';
+        if (request.baseUrl && request.baseUrl !== '/') {
+            basePath = request.baseUrl + basePath;
+        }
+        return basePath;
+    }
 }
--- a/packages/rest/src/run.ts
+++ b/packages/rest/src/run.ts
@@ -1,0 +1,29 @@
+import {RestApplication, RestServer, RestServerConfig, get} from '.';
+
+function givenApplication(options?: {rest: RestServerConfig}) {
+    options = options || {rest: {port: 4000, host: '127.0.0.1'}};
+    return new RestApplication(options);
+}
+
+async function main() {
+    let restApp = givenApplication();
+    class PingController {
+        @get('/ping')
+        ping(): string {
+            return 'Hi';
+        }
+        @get('/bye')
+        bye(): string {
+            return 'Bye';
+        }
+    }
+    restApp.controller(PingController);
+    restApp.redirect('/ping',(protocol:string, host:string, basePath:string)=>{
+        console.log(protocol+host+basePath);
+        return protocol+'://'+host+basePath+'/bye';
+    });
+    //restApp.redirect('/ping','/bye');
+    await restApp.start();
+}
+
+main();


sanadHaj:
@bajtos @raymondfeng  @hacksparrow 
this file will be removed https://github.com/strongloop/loopback-next/blob/35066c7b40c9e50889f5947f8cde9e9fb8cf6aeb/packages/rest/src/run.ts 
its only for testing purpose . as well  the `
import {Request} from "express"; `  

let me know what do you think before start writing tests

bajtos:
@sanadHaj thank you for the pull request. I took a quick look at the proposed changes, my first impression is that your changes are going in a wrong direction, unfortunately. I'll do a more thorough review later this week and post more details.

packages/rest/src/rest.application.ts
```diff
@@ -244,7 +244,7 @@ export class RestApplication extends Application implements HttpServerLike {
    */
   redirect(
     fromPath: string,
-    toPathOrUrl: string,
+    toPathOrUrl: string | Function,
```
bajtos: Please use a more specific type instead of `Function`, for example:

```ts
toPathOrUrl: string | LocationFactory
```

where `LocationFactory` can be defined as follows:

```ts
export type LocationFactory = (ctx: RedirectContext) => string;

export interface RedirectContext {
  request: Request,
  // etc.
}
```

packages/rest/src/router/redirect-route.ts
```diff
@@ -39,4 +46,24 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
...
+      return (
+          (request.get('x-forwarded-proto') || '').split(',')[0] ||
+          request.protocol ||
+          'http'
+      );
+  }
+
+   private _getBasePath(request: Request) {
+        let basePath = '';
+        if (request.baseUrl && request.baseUrl !== '/') {
```
bajtos: `request.baseUrl` is set by Express when a middleware is mounted via `app.use(basePath, handler)`. It's a different concept from LB4's `basePath` option.

Eventually, we need to support both. Let's work in small increments please and leave `baseUrl` out of scope of this pull request.

packages/rest/src/rest.server.ts
```diff
@@ -683,11 +683,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
    */
   redirect(
     fromPath: string,
-    toPathOrUrl: string,
+    toPathOrUrl: string | Function,
```
bajtos: Use `LocationFactory` instead of `Function`.

packages/rest/src/rest.server.ts
```diff
@@ -683,11 +683,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
     statusCode?: number,
   ): Binding {
     return this.route(
-      new RedirectRoute(fromPath, this._basePath + toPathOrUrl, statusCode),
+      new RedirectRoute(fromPath, toPathOrUrl, statusCode),
```
bajtos: I think you should pass `this._basePath` to `RedirectRoute` constructor via a new argument.

packages/rest/src/router/redirect-route.ts
```diff
@@ -19,15 +19,22 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
 
   constructor(
     private readonly sourcePath: string,
-    private readonly targetLocation: string,
+    private readonly targetLocation: string | Function,
```
bajtos: Use `LocationFactory` instead of `Function`.

packages/rest/src/router/redirect-route.ts
```diff
@@ -19,15 +19,22 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
...
-    {response}: RequestContext,
+    {response,request}: RequestContext,
     args: OperationArgs,
   ): Promise<OperationRetval> {
-    response.redirect(this.statusCode, this.targetLocation);
+    let basePath = this._getBasePath(request) || '';
+    let targetLocation :string = typeof this.targetLocation === 'function'
+        ? this.targetLocation(
+          this._getProtocolForRequest(request),
+          this._getHost(request),
```
bajtos: We should not be running this complex logic for simple redirects. Here is how I am envisioning the implementation:

```ts
class {

  async invokeHandler(
    {response,request}: RequestContext,
    args: OperationArgs,
  ): Promise<OperationRetval> {
    if (typeof this.targetLocation === 'string') {
      response.redirect(this.statusCode, this._basePath + this.targetLocation);
      return;
    }
    const ctx: RedirectContext = {
      request,
      basePath: this._basePath,
      // build additional fields like protocol, request, etc.
    };
    const targetLocation = this.targetLocation(ctx);
    response.redirect(this.statusCode, targetLocation);
  }

}
```

See https://github.com/strongloop/loopback-next/issues/2022#issuecomment-468588124 for more details.

packages/rest/src/rest.application.ts
bajtos: On the second thought: let's not invent any new `RedirectContext` class/interface and use the existing RequestContext as defined here:

https://github.com/strongloop/loopback-next/blob/c8f1a71fc9ae7f2331f52fe11af43fab567ba4d1/packages/rest/src/request-context.ts#L15-L21

I feel the additional fields like `basePath`, etc. may be useful in other places besides redirection, therefore it make sense to define them directly on the request context.

bajtos:
I am proposing to make this pull request smaller and focus on introducing `locationFactory` only, leaving out any computed fields like the requested host, etc.

```ts
class {

  async invokeHandler(
    requestContext: RequestContext,
    args: OperationArgs,
  ): Promise<OperationRetval> {
    const target = typeof this.targetLocation === 'string'
      ? this._basePath + this.targetLocation
      : this.targetLocation(requestContext);
    requestContext.response.redirect(this.statusCode, target);
  }

}
```

Please make sure to include automated tests and also update API docs & regular documentation.

packages/rest/src/rest.application.ts
bajtos: >  it make sense to define them directly on the request context.

See https://github.com/strongloop/loopback-next/pull/2547

packages/rest/src/rest.server.ts
bajtos: Even better: wait until https://github.com/strongloop/loopback-next/pull/2547 is landed and them modify RedirectRoute's `invokeHandler` function to obtain the base path from the request context via `requestContext.basePath`.

packages/rest/src/run.ts
```diff
@@ -0,0 +1,29 @@
...
+        bye(): string {
+            return 'Bye';
+        }
+    }
+    restApp.controller(PingController);
+    restApp.redirect('/ping',(protocol:string, host:string, basePath:string)=>{
+        console.log(protocol+host+basePath);
+        return protocol+'://'+host+basePath+'/bye';
+    });
+    //restApp.redirect('/ping','/bye');
```
hacksparrow: Please clean up irrelevant comments from PRs.

hacksparrow:
@sanadHaj thanks for the PR. Please run `npm t` to catch possible linting issues, apart checking implementation integrity.

packages/rest/src/run.ts
sanadHaj: @hacksparrow  this file for testing purpose . 
this pull request seized until #2547 resolved .

bajtos:
FWIW, https://github.com/strongloop/loopback-next/pull/2547 has been landed.

bajtos:
Hi @sanadHaj, what's the status of this pull request? Are you still keen to continue working on this feature?

PR #2533, opened 2019-03-04
feat(openapi-v3): add operationId based on controller/method names

raymondfeng:
While  was working on https://github.com/strongloop/loopback-connector-swagger/pull/52, I realized that LB4 does not generate `operationId` for OpenAPI operation specs.

## Checklist


packages/openapi-v3/src/__tests__/unit/decorators/param/param.decorator.unit.ts
```diff
@@ -32,10 +32,11 @@ describe('Routing metadata for parameters', () => {
 
       const expectedSpec = anOperationSpec()
         .withOperationName('greet')
+        .withControllerName('MyController')
         .withParameter(paramSpec)
         .withResponse(200, {description: 'Return value of MyController.greet'})
         .build();
-      expect(actualSpec.paths['/greet']['get']).to.eql(expectedSpec);
+      expect(actualSpec.paths['/greet']['get']).to.containEql(expectedSpec);
```
jannyHou: nitpick: could you update the `withOperationName` function as well to generate the `operationId` so that the `expectedSpec` contains that field?
The `.eql()` is used intended to make sure the generated spec and `expectedSpec` are exactly the same. Thanks.

raymondfeng: Fixed.

packages/openapi-spec-builder/src/openapi-spec-builder.ts
```diff
@@ -170,7 +170,34 @@ export class OperationSpecBuilder extends BuilderBase<OperationObject> {
...
+    this.setupOperationId();
+    return this;
+  }
+
+  private setupOperationId() {
+    if (this._spec.operationId) return;
+    const controllerName = this._spec['x-controller-name'];
+    const operationName = this._spec['x-operation-name'];
+    if (controllerName && operationName) {
+      this._spec.operationId = controllerName + '$' + operationName;
```
bajtos: This will produce IDs like `Product$create`. Have you considered using `.` separator to produce `Product.create` instead? 

In LoopBack 3, we use the dot character, see
https://github.com/strongloop/loopback-swagger/blob/52ba102c3e5dcd653c6d0a3243b16f4a51ecc169/test/specgen/route-helper.test.js#L501-L504
```js
  it('route operationId DOES include model name.', function() {
    var doc = createAPIDoc({method: 'User.login'});
    expect(doc.operation.operationId).to.equal('User.login');
  });
```

Also what's the impact on client generators like https://github.com/swagger-api/swagger-codegen, is there any difference in how they treat both `$` and `.` ?

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -186,6 +186,15 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
     }
 
     operationSpec['x-operation-name'] = op;
+    operationSpec['x-controller-name'] =
+      operationSpec['x-controller-name'] || constructor.name;
+
+    if (operationSpec.operationId == null) {
+      operationSpec.operationId =
+        operationSpec['x-controller-name'] +
+        '$' +
```
bajtos: Ditto, have you considered using `.` instead?

packages/openapi-spec-builder/src/openapi-spec-builder.ts
raymondfeng: Swagger clients such as https://github.com/swagger-api/swagger-js uses `operationId` as method names. Using `PingController.ping` will be confusing.

raymondfeng:
@jannyHou @bajtos I created two commits for this PR:

1. Improve `openapi-spec-builder` with new methods and add tests
2. Set up `operationId` for controller routes

PTAL.

packages/openapi-v3/src/controller-spec.ts
raymondfeng: See above.

packages/openapi-spec-builder/src/openapi-spec-builder.ts
bajtos: From what I remember from the times where we were looking into swagger-codegen, they were normalizing operation ids and replacing characters like `.` to `_`. As a result, you would end up with a method like `PingController_ping`.

AFAICT from reading swagger-js code, it's doing the same.

https://github.com/swagger-api/swagger-js/blob/0cdc6488372547f69ca7372cc9ad44cf4bf92de5/src/helpers.js#L29-L38
```js
export function opId(operation, pathName, method = '', {v2OperationIdCompatibilityMode} = {}) {
  if (!operation || typeof operation !== 'object') {
    return null
  }
  const idWithoutWhitespace = (operation.operationId || '').replace(/\s/g, '')
  if (idWithoutWhitespace.length) {
    return escapeString(operation.operationId)
  }
  return idFromPathMethod(pathName, method, {v2OperationIdCompatibilityMode})
}
```

https://github.com/swagger-api/swagger-js/blob/0cdc6488372547f69ca7372cc9ad44cf4bf92de5/src/helpers.js#L5-L7

```js
const escapeString = (str) => {
  return str.replace(/[^\w]/gi, '_')
}
```

Could you please verify?

If my understanding is correct, then both `PingController$ping` and `PingController.ping` are going to produce the same method name `PingController_ping`.

The reason why I am going to such length here is that changing the format of generated operation ids would be a breaking change for our consumers.

packages/openapi-spec-builder/src/__tests__/unit/openapi-spec-builder.unit.ts
```diff
@@ -0,0 +1,143 @@
...
+        .build();
+      expect(spec).to.containEql({operationId: 'MyController.greet'});
+    });
+
+    it('sets operationId from controller/operation name', () => {
+      const spec = anOperationSpec()
+        .withControllerName('MyController')
+        .withOperationName('greet')
+        .build();
+      expect(spec).to.containEql({operationId: 'MyController$greet'});
```
bajtos: Depending on how we resolve our discussion above, this may need to be changed to

```suggestion
      expect(spec).to.containEql({operationId: 'MyController.greet'});
```

packages/openapi-spec-builder/src/openapi-spec-builder.ts
raymondfeng: Good finding. I'll use `.` instead of `$` then.

PR #2547, opened 2019-03-07
feat: add computed RequestContext properties like `requestedBaseUrl`

bajtos:
Move the code calculating the requested protocol, base path and base URL from RestServer private methods into RequestContext's public API. This allows Route implementations to access these calculated properties too. For example, it simplifies building target location for HTTP redirect response (see https://github.com/strongloop/loopback-next/issues/2022#issuecomment-468588124) and makes it possible to move the implementation of `/openapi.json` route from Express route into regular LB4 RouteEntry implementation (see https://github.com/strongloop/loopback-next/pull/2434#discussion_r263265038 and a PoC in https://github.com/strongloop/loopback-next/commit/2b5be221cb49e8cc28ca9a4cd4e1d58aa200e204).

This pull request has three parts:
 - The first commit improves `givenHttpServerConfig` to support creation of HTTPS servers by filling-in dummy `key` and `cert` values.
 - The second commit tightens type checks in RestServer config and fixes a problem discovered along the way.
 - The third commit introduces new RequestContext properties and refactors RestServer to use these new properties instead of private helper methods.

## Checklist


packages/rest/src/request-context.ts
```diff
@@ -6,17 +6,90 @@
...
 export class RequestContext extends Context implements HandlerContext {
+  /**
+   * Get the protocol used by the client to make the request.
+   * Please note this protocol may be different from what we are observing
+   * at HTTP/TCP level, because reverse proxies like nginx or sidecars like
+   * Envoy are switching between protocols.
+   */
+  get requestedProtocol(): string {
+    return (
+      (this.request.get('x-forwarded-proto') || '').split(',')[0] ||
```
hacksparrow: Do we wan to check `this.request.get('x-forwarded-proto') || '').split(',')[0]` actually yields `http` or `https`, or do we blindly return whatever was set in the `x-forwarded-proto` header?

bajtos: Good catch.

To be honest, I think the extra complexity of checking whether `x-forwarded-proto` contains http/https is not worth our time right now. Personally, I prefer to wait until there is a user with a real-word scenario where this becomes a problem.

If you feel this edge case is important, then please open a new issue to track it.

This pull request is not modifying LB4's behavior in this particular edge case.

hacksparrow: I see potential problems in code like `foo[protocol].bar`, where `protocol` was expected to be either `http` or `https`.

I am fine with you suggestion, though.

packages/rest/src/request-context.ts
```diff
@@ -6,17 +6,90 @@
...
  * (request, response, etc.).
  */
 export class RequestContext extends Context implements HandlerContext {
+  /**
+   * Get the protocol used by the client to make the request.
+   * Please note this protocol may be different from what we are observing
+   * at HTTP/TCP level, because reverse proxies like nginx or sidecars like
+   * Envoy are switching between protocols.
+   */
+  get requestedProtocol(): string {
```
hacksparrow: Shouldn't we add unit test for this? We want to be able to detect any unexpected change in future.

packages/rest/src/request-context.ts
```diff
@@ -6,17 +6,90 @@
...
+
+  /**
+   * Get the base URL used by the client to make the request.
+   * This URL contains the protocol, hostname, port and base path.
+   * The path of the invoked route and query string is not included.
+   *
+   * Please note these values may be different from what we are observing
+   * at HTTP/TCP level, because reverse proxies like nginx are rewriting them.
+   */
+  get requestedBaseUrl(): string {
```
hacksparrow: Shouldn't we add unit test for this? We want to be able to detect any unexpected change in future.

packages/rest/src/request-context.ts
```diff
@@ -6,17 +6,90 @@
...
+      'http'
+    );
+  }
+
+  /**
+   * Get the effective base path of the incoming request. This base path
+   * combines `baseUrl` provided by Express when LB4 handler is mounted on
+   * a non-root path, with the `basePath` value configured at LB4 side.
+   */
+  get basePath(): string {
```
hacksparrow: Shouldn't we add unit test for this? We want to be able to detect any unexpected change in future.

packages/rest/src/rest.server.ts
```diff
@@ -979,20 +954,24 @@ export interface ApiExplorerOptions {
 }
 
 /**
- * Options for RestServer configuration
+ * RestServer options
  */
-export interface RestServerOptions {
+export type RestServerOptions = Partial<RestServerResolvedOptions>;
+
+export interface RestServerResolvedOptions {
```
raymondfeng: We can leave `RestServerOptions` unchanged and declare resolved options as `Required<RestServerOptions>`. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html.

Or event make it `ReadonlyRequired`:
```ts
type ReadonlyRequired<T> = { +readonly [P in keyof T]-?: T[P] };  // Add `readonly` and remove `?`
```

packages/rest/src/request-context.ts
raymondfeng: The property name should be simplified as `protocol`.

raymondfeng:
Please fix formatting issue:

> loopback-next@0.1.0 prettier:cli /home/travis/build/strongloop/loopback-next
> node packages/build/bin/run-prettier "**/*.ts" "**/*.js" "**/*.md" "-l"
packages/rest/src/__tests__/integration/http-handler.integration.ts

packages/rest/src/rest.server.ts
bajtos: Nice, I was not aware of `Required` type.

Please note that not all fields are required in the resolved config. The following properties are remaining optional: `basePath`, `sequence`, `requestBodyParser`.

Conceptually, it makes more sense to me to use the resolved config as the source of truth:
- RestServer defines a type describing configuration it uses at runtime. The shape of the config object as required by various implementation bits.
- Then we define another interface describing options accepted by the server. This interface uses optional properties for fields that have a reasonable default.

packages/rest/src/request-context.ts
bajtos: Good catch. I was being lazy and relying on the existing integration/acceptance-level tests. Will add unit tests.

bajtos:
@raymondfeng @hacksparrow LGTY now?

packages/rest/src/__tests__/unit/request-context.unit.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  describe('requestedBaseUrl', () => {
+    it('defaults to data from the HTTP connection', () => {
+      const ctx = givenRequestContext(
+        aRestServerConfig({
+          host: undefined,
+          port: 0,
+        }),
+      );
+      expect(ctx.requestedBaseUrl).to.equal('http://localhost');
```
hacksparrow: When port is `0`, doesn't it use a random port? Should't the resulting base URL also contain the port number?

bajtos: Good catch again!

I guess this is a limitation of testing this part in unit-test style. When port is set to `0`, we should be using a random port indeed. I don't know how to mock that up though.

What do you propose?

raymondfeng: Maybe it does not make sense to unit test this function?

raymondfeng: Or generate a random port in `givenRequestContext` if port is 0?

raymondfeng: I wonder why the host is `localhost` instead of `127.0.0.1`.

hacksparrow: > Maybe it does not make sense to unit test this function?

We should be able to detect breaking changes in the implementation/behavior without having a bug let us know about it.

hacksparrow: > I wonder why the host is localhost instead of 127.0.0.1

`shot` is making a request to `localhost`. The value of the `host` is whatever the client is requesting, not what was configured.

hacksparrow: > Or generate a random port in givenRequestContext if port is 0?

Just found out, no matter what `port` value you set, it is always being set to `80` (by `shot`?). Since `80` is the default HTTP port, it is being stripped to return `http://localhost` instead of `http://localhost:80`.

Shouldn't `RequestContext` be tested in the context of an actual `RestServer` instead of a simulator (shot)?

bajtos: > Shouldn't RequestContext be tested in the context of an actual RestServer instead of a simulator (shot)?

Possibly. I wrote the tests at unit-test level because you asked for that:

> Shouldn't we add unit test for this? We want to be able to detect any unexpected change in future.

bajtos: I rewrote all new tests into integration-test style, to verify the real end-to-end behavior.

bajtos:
Reworked new tests from unit-style to integration-style.

@raymondfeng @hacksparrow LGTY now?

packages/rest/src/__tests__/integration/request-context.integration.ts
```diff
@@ -0,0 +1,130 @@
...
+      await client.get('/products').expect(200);
+      expect(observedCtx.basePath).to.equal('');
+    });
+
+    it('honors baseUrl when mounted on a sub-path', async () => {
+      const lbApp = new RestApplication();
+      lbApp.handler(contextObservingHandler);
+
+      const expressApp = express();
+      expressApp.use('/api', lbApp.requestHandler);
```
raymondfeng: What's going to happen if `lbApp.basePath('/v1')` is called?

raymondfeng: We probably need to have a test to cover the case above.

packages/testlab/src/http-server-config.ts
```diff
@@ -3,23 +3,49 @@
...
 
+import * as path from 'path';
+import {readFileSync} from 'fs';
+import {ServerOptions as HttpsServerOptions} from 'https';
+
+const FIXTURES = path.resolve(__dirname, '../fixtures');
+const DUMMY_TLS_CONFIG = {
+  key: readFileSync(path.join(FIXTURES, 'key.pem')),
+  cert: readFileSync(path.join(FIXTURES, 'cert.pem')),
+};
```
raymondfeng: 👍

packages/rest/src/__tests__/integration/request-context.integration.ts
bajtos: I agree it's a use case we should cover by our test suite and possibly fix.

At the same time, I feel it's out of scope of this pull request. My intention is to move certain computed properties from RestServer to RequestContext while preserving their behavior.

I'll open a new PR to take a look at this.

bajtos: See https://github.com/strongloop/loopback-next/pull/2589

PR #2552, opened 2019-03-07
update express-composition example's npm files and include the example in `npm test`

nabdelgadir:
Removed `files` from `package.json` of `express-composition` example so the correct files download when using `lb4 example express-composition`. Then changed the `express-composition` example based on how it's done in https://github.com/strongloop/loopback-next/pull/2316 so they're included when running `npm test` at the root of `loopback-next`.

## Checklist

### DIFF
--- a/examples/express-composition/src/__tests__/acceptance/note.acceptance.ts
+++ b/examples/express-composition/src/__tests__/acceptance/note.acceptance.ts
@@ -12,6 +12,7 @@ describe('NoteApplication', () => {
 
   before('setupApplication', async () => {
     ({server, client, lbApp} = await setupExpressApplication());
+    await changeDataSourceConfig();
     await givenNoteRepository();
   });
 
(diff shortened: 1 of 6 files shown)


examples/express-composition/src/__tests__/acceptance/note.acceptance.ts
```diff
@@ -57,4 +58,15 @@ describe('NoteApplication', () => {
   async function givenNoteRepository() {
     noteRepo = await lbApp.getRepository(NoteRepository);
   }
+
+  async function changeDataSourceConfig() {
+    /**
+     * Override default config for DataSource for testing so we don't write
+     * test data to file when using the memory connector.
```
bajtos: Have you considered using the approach we are recommending in https://loopback.io/doc/en/lb4/Testing-your-application.html#create-a-test-datasource? Maybe it's not a good one, IDK.

It would be great to have a single prescribed way, used consistently both in our docs and the example repos. I think that's out of scope of this pull request though.

nabdelgadir: I was following how it was done in the `Todo` example, but I can definitely look into that way and update the examples/the docs in a different PR.

PR #2553, opened 2019-03-07
feat(context): add events to ContextView

raymondfeng:
Extension points that use ContextView can then listen on events to update
its state/cache beyond the view accordingly.

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/context-view.unit.ts
+++ b/packages/context/src/__tests__/unit/context-view.unit.ts
@@ -90,6 +90,49 @@ describe('ContextView', () => {
     expect(await taggedAsFoo.values()).to.eql(['BAR', 'FOO']);
   });
 
+  describe('EventEmitter', () => {
+    let events: string[] = [];
+
+    beforeEach(setupListeners);
+
+    it('emits close', () => {
+      taggedAsFoo.close();
+      expect(events).to.eql(['close']);
+      // 2nd close does not emit `close` as it's closed
+      taggedAsFoo.close();
+      expect(events).to.eql(['close']);
+    });
+
+    it('emits refresh', () => {
+      taggedAsFoo.refresh();
+      expect(events).to.eql(['refresh']);
+    });
+
+    it('emits resolve', async () => {
+      await taggedAsFoo.values();
+      expect(events).to.eql(['resolve']);
+      // Second call does not resolve as values are cached
+      await taggedAsFoo.values();
+      expect(events).to.eql(['resolve']);
+    });
+
+    it('emits refresh & resolve when bindings are changed', async () => {
+      server
+        .bind('xyz')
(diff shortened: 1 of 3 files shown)


packages/context/src/context-view.ts
```diff
@@ -28,23 +29,36 @@ const nextTick = promisify(process.nextTick);
  * points. For example, the RestServer can react to `controller` bindings even
  * they are added/removed/updated after the application starts.
  *
+ * `ContextView` is an event emitter that emits the following events:
+ * - 'open': when the view is open (starting to observe context events)
```
bajtos: How do you envision to use this event in practice? I believe `view.open` is called by the framework before the view is injected to the receiver and thus before the receiver can install `on('open')` handler.

packages/context/src/__tests__/unit/context-view.unit.ts
```diff
@@ -90,6 +90,59 @@ describe('ContextView', () => {
...
+      expect(events).to.eql(['close']);
+      // 2nd close does not emit `close` as it's closed
+      taggedAsFoo.close();
+      expect(events).to.eql(['close']);
+    });
+
+    it('emits open', () => {
+      // At this point the view is opened and call `open` again won't emit
+      // `open` event
+      taggedAsFoo.open();
```
bajtos: Please add a check that no events were emitted, to make this test code easier to read & understand.

```ts
taggedAsFoo.open();
expect(events).to.eql([]);
```

packages/context/src/context-view.ts
raymondfeng: Good catch. I'll remove `open` and `close` events as I only have use cases for `refresh/resolve` in mind.

raymondfeng: I removed `open` event.

packages/context/src/__tests__/unit/context-view.unit.ts
raymondfeng: `open` is removed.

packages/context/src/context-view.ts
```diff
@@ -28,23 +29,34 @@ const nextTick = promisify(process.nextTick);
...
   protected _cachedBindings: Readonly<Binding<T>>[] | undefined;
   protected _cachedValues: T[] | undefined;
   private _subscription: Subscription | undefined;
 
   constructor(
     protected readonly context: Context,
     public readonly filter: BindingFilter,
-  ) {}
+  ) {
+    super();
```
bajtos: Is it necessary to explicitly call the parent constructor? Isn't it automatically called by JS runtime?

raymondfeng: Yes, it's necessary as the base class's constructor has a different signature.

raymondfeng:
Added docs too.

PR #2558, opened 2019-03-08
Enable package lock to reduce duration & CPU usage of "npm install" 🏎🌱

bajtos:
Historically, we were avoiding package locks to ensure our CI builds are using the latest version of all dependencies, the same versions that our consumers would receive when running `npm install` in their projects.

Now that we have Greenkeeper enabled to detect new versions in our dependency tree and semi-automatically update our package.json & package-lock.json files, I feel it's safe to use package-lock feature.

This pull request enables package-locks in the entire monorepo in the first commit, and adds individual `package-lock.json` files in the second commit - I hope this will make the change easier to review. The third commit removes Node.js version 10.0.0 from the build matrix on Windows, because `npm ci` is not available there. (We cannot upgrade to a newer 10.x version - see https://github.com/strongloop/loopback-next/pull/1431.)

Note: To fully leverage the power of package locks, the CI is using `npm ci` instead of `npm install` now. Similarly, `lerna` is configured to use `npm ci` when installing dependencies in individual monorepo packages.

As a result, the time to install dependencies was cut by half - from 1m44s to 0m52s on my machine.  Not only the installation is faster, it's also burning less CPU cycles (and energy).

**before**

```
# clean all dependencies
$ npm run clean:lerna && rm -rf node_modules 
# measure how long it takes to install them
$ time npm i

real  1m43.710s
user  6m44.097s
sys   1m37.536s
```

**after**

```
# clean all dependencies
$ npm run clean:lerna && rm -rf node_modules 
# measure how long it takes to install them using "npm ci"
$ time npm ci

real  0m51.985s
user  3m56.915s
sys   1m18.982s
```

## Checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-package-lock=false
+package-lock=true
--- a/examples/hello-world/package-lock.json
+++ b/examples/hello-world/package-lock.json
@@ -1,0 +1,361 @@
+{
+	"name": "@loopback/example-hello-world",
+	"version": "1.1.5",
+	"lockfileVersion": 1,
+	"requires": true,
+	"dependencies": {
+		"@types/node": {
+			"version": "10.12.30",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.30.tgz",
+			"integrity": "sha512-nsqTN6zUcm9xtdJiM9OvOJ5EF0kOI8f1Zuug27O/rgtxCRJHGqncSWfCMZUP852dCKPsDsYXGvBhxfRjDBkF5Q==",
+			"dev": true
+		},
+		"ansi-regex": {
+			"version": "2.1.1",
+			"resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+			"integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
+			"dev": true
+		},
+		"ansi-styles": {
+			"version": "2.2.1",
+			"resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-2.2.1.tgz",
+			"integrity": "sha1-tDLdM1i2NM914eRmQ2gkBTPB3b4=",
+			"dev": true
+		},
+		"argparse": {
+			"version": "1.0.10",
+			"resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
+			"integrity": "sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==",
+			"dev": true,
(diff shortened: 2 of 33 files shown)


.npmrc
```diff
@@ -1 +1 @@
-package-lock=false
+package-lock=true
```
bajtos: I am intentionally keeping `.npmrc` and forcing `package-lock` on all loopback-next developers.

Personally, I have a global `.npmrc` file setting `package-lock` to `false`, because that's what I need when working on LB3 repositories. If there was no `.npmrc` file in loopback-next, then I would not be using package-lock functionality because of my global config.

IMO, it's best to not make any assumptions about the global npm configuration and explicitly tell npm that we want to use package-lock files.

bajtos:
Bummer, Node.js 10.0.0 used on AppVeyor does not support `npm ci`. I think it's best to remove 10.0.0 from the build matrix on Windows. The version 10.0.0 is very outdated when compared to current 10.x LTS versions.

bajtos:
@raymondfeng if the patch LGTY and there are no objections from other reviewers, then feel free to land it yourself. The sooner we speed up our CI builds, the better for us and the environment too.

raymondfeng:
We need to try out a few things:

1. Make sure `package-lock.json` works nicely with lerna including our release process.
2. Figure out the impact of released `@loopback/*` modules with `package-lock.json`, such as:
  - `@loopback/rest@2.0.0` depends on `@loopback/context@^2.0.0` (locked)
  - Does `@loopback/rest@2.0.0` pick up `@loopback/context@2.0.1` for LB4 apps if `npm i` is used?

raymondfeng:
I'm seeing a diff in `package-lock.json` such as:
```
-          "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
+          "resolved": "http://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
```

I have this in `~/.npmrc`:

```
//registry.npmjs.org/:_authToken=...
```

examples/hello-world/package-lock.json
```diff
@@ -0,0 +1,361 @@
+{
```
raymondfeng: It's surprising that I don't see entries for `@loopback/*` dependencies in the generated `package-lock.json`. Is it by design as they are symbolically linked? If so, what's the impact for published modules?

raymondfeng:
> Figure out the impact of released @loopback/* modules with package-lock.json

I just realized that `package-lock.json` won't be published to npm. 

- package-lock.json is never published to npm, whereas npm-shrinkwrap is by default
- package-lock.json files that are not in the top-level package are ignored, but shrinkwrap files belonging to dependencies are respected

Basically, `package-lock.json` is used to ensure repeatable build while `shrinkwrap.json` is used for repeatable installation/runtime.

bajtos:
> Basically, `package-lock.json` is used to ensure repeatable build while `shrinkwrap.json` is used for repeatable installation/runtime.

Exactly 👍

PR #2560, opened 2019-03-08
fix(rest): make sure basePath is included in RestServer.url

raymondfeng:
RestServer.url does not reflect `basePath`. The problem can be reproduced by adding basePath: '/api' to [examples/todo/index.js](https://github.com/strongloop/loopback-next/blob/master/examples/todo/index.js#L15) and run `npm build`:

The url is printed as http://127.0.0.1:3000, which returns 404. I think the url should include the basePath, i.e., 'http://127.0.0.1:3000/api'.

See https://github.com/strongloop/loopback-next/pull/2554#issuecomment-471089213

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/integration/rest.server.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.server.integration.ts
@@ -6,12 +6,12 @@
 import {Application} from '@loopback/core';
 import {
   createClientForHandler,
+  createRestAppClient,
   expect,
   givenHttpServerConfig,
   httpsGetAsync,
   itSkippedOnTravis,
   supertest,
-  createRestAppClient,
 } from '@loopback/testlab';
 import * as fs from 'fs';
 import {IncomingMessage, ServerResponse} from 'http';
@@ -26,11 +26,11 @@ import {
   Request,
   requestBody,
   RequestContext,
+  RestApplication,
   RestBindings,
   RestComponent,
   RestServer,
   RestServerConfig,
-  RestApplication,
 } from '../..';
 const readFileAsync = util.promisify(fs.readFile);
 
@@ -38,20 +38,76 @@ const FIXTURES = path.resolve(__dirname, '../../../fixtures');
 const ASSETS = path.resolve(FIXTURES, 'assets');
 
 describe('RestServer (integration)', () => {
-  it('exports url property', async () => {
-    const server = await givenAServer();
-    server.handler(dummyRequestHandler);
-    expect(server.url).to.be.undefined();
-    await server.start();
-    expect(server)
-      .to.have.property('url')
-      .which.is.a.String()
-      .match(/http|https\:\/\//);
-    await supertest(server.url)
-      .get('/')
-      .expect(200, 'Hello');
-    await server.stop();
-    expect(server.url).to.be.undefined();
+  describe('url/rootUrl properties', () => {
+    let server: RestServer;
+
+    afterEach('shuts down server', async () => {
+      if (!server) return;
+      await server.stop();
+      expect(server.url).to.be.undefined();
+      expect(server.rootUrl).to.be.undefined();
+    });
+
+    describe('url', () => {
+      it('exports url property', async () => {
+        server = await givenAServer();
+        server.handler(dummyRequestHandler);
+        expect(server.url).to.be.undefined();
+        await server.start();
+        expect(server)
+          .to.have.property('url')
+          .which.is.a.String()
+          .match(/http|https\:\/\//);
+        await supertest(server.url)
+          .get('/')
+          .expect(200, 'Hello');
+      });
+
+      it('includes basePath in the url property', async () => {
+        server = await givenAServer({rest: {basePath: '/api'}});
+        server.handler(dummyRequestHandler);
+        expect(server.url).to.be.undefined();
+        await server.start();
+        expect(server)
+          .to.have.property('url')
+          .which.is.a.String()
+          .match(/http|https\:\/\//);
+        expect(server.url).to.match(/api$/);
+        await supertest(server.url)
+          .get('/')
+          .expect(200, 'Hello');
+      });
+    });
+
+    describe('rootUrl', () => {
+      it('exports rootUrl property', async () => {
+        server = await givenAServer();
+        server.handler(dummyRequestHandler);
+        expect(server.rootUrl).to.be.undefined();
+        await server.start();
+        expect(server)
+          .to.have.property('rootUrl')
+          .which.is.a.String()
+          .match(/http|https\:\/\//);
+        await supertest(server.rootUrl)
+          .get('/api')
+          .expect(200, 'Hello');
+      });
+
+      it('does not include basePath in rootUrl', async () => {
+        server = await givenAServer({rest: {basePath: '/api'}});
+        server.handler(dummyRequestHandler);
+        expect(server.rootUrl).to.be.undefined();
+        await server.start();
+        expect(server)
+          .to.have.property('rootUrl')
+          .which.is.a.String()
+          .match(/http|https\:\/\//);
+        await supertest(server.rootUrl)
+          .get('/api')
+          .expect(200, 'Hello');
+      });
+    });
   });
 
   it('parses query without decorated rest query params', async () => {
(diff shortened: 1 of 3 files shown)


packages/rest/src/rest.server.ts
```diff
@@ -151,7 +151,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
   }
 
   get url(): string | undefined {
```
bajtos: Please add TSDoc comment to explain what is this property supposed to return. Make it clear that `basePath` is included in the URL.

hacksparrow:
@bajtos I added as a way to access `this._httpServer.url`. Don't remember adding any functionality based around it.

packages/rest/src/rest.server.ts
```diff
@@ -150,8 +150,14 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
   }
 
+  /**
+   * The base url for the server, including the basePath if set
+   */
   get url(): string | undefined {
-    return this._httpServer && this._httpServer.url;
+    let serverUrl = this._httpServer && this._httpServer.url;
+    if (!serverUrl) return serverUrl;
+    serverUrl = serverUrl + (this._basePath || '');
```
hacksparrow: I find it a little strange that a property named `url` in `RestServer` should have anything appended to it.

Shouldn't `basePath` be a setter-getter, instead?

raymondfeng: Are you proposing to have getters for base url and basePath? Url can have path and it matches servers in OpenAPI spec.

raymondfeng: See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#server-object

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -36,20 +36,39 @@ const FIXTURES = path.resolve(__dirname, '../../../fixtures');
...
+      expect(server.url).to.be.undefined();
+      await server.start();
+      expect(server)
+        .to.have.property('url')
+        .which.is.a.String()
+        .match(/http|https\:\/\//);
+      await supertest(server.url)
+        .get('/')
+        .expect(200, 'Hello');
+      await server.stop();
```
bajtos: When the test fails, this line is never called and the server will stay running until the test suite is finished. I am proposing to move this line to `afterEach` hook.

```ts
describe('RestServer (integration)', () => {
  let server: RestServer;

  afterEach('stop the server', async function teardown() {
    if (!server) return;
    await server.stop();
  });

  // the tests
});
```

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -98,7 +98,9 @@ describe('RestApplication (integration)', () => {
     restApp.static('/', ASSETS);
     await restApp.start();
     client = createRestAppClient(restApp);
-    await client.get('/html/index.html').expect(200);
+    // No more `/html` in client as `restApp.restServer.url` already contains
+    // the base path `/html`
```
bajtos: This change will make it more difficult to use `createRestAppClient` for testing endpoints that are ignoring basePath, e.g. `/openapi.json`.

I think this is also a breaking change for consumers of `createRestAppClient`.

Personally, I'd prefer to preserve the current behavior of `reateRestAppClient` where it creates supertest instance for the host+port only.

Maybe it's a sign that we need two server-level properties?
 - protocol+host+port only (e.g. `http://127.0.0.1:3000`)
 - base URL (e.g. `http://127.0.0.1:3000/my-base-path`)

Let's discuss.

packages/rest/src/rest.server.ts
hacksparrow: Oh, I see.

packages/rest/src/__tests__/integration/rest.application.integration.ts
raymondfeng: I introduced `rootUrl` for the protocol + host + port and changed `createRestAppClient` to use `rootUrl` to maintain the backward compatibility.

I use `url` with base path to be consistent with OpenAPI spec server url.

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -26,32 +26,71 @@ import {
...
-    await supertest(server.url)
-      .get('/')
-      .expect(200, 'Hello');
-    await server.stop();
-    expect(server.url).to.be.undefined();
+  describe('url', () => {
+    let server: RestServer;
+
+    afterEach('shuts down server', async () => {
+      if (server) {
```
bajtos:
```suggestion
      if (!server) return;
```

It reduces indentation.

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -26,32 +26,71 @@ import {
...
+        .which.is.a.String()
+        .match(/http|https\:\/\//);
+      expect(server.url).to.match(/api$/);
+      await supertest(server.url)
+        .get('/')
+        .expect(200, 'Hello');
+    });
+
+    it('exports rootUrl property', async () => {
+      server = await givenAServer();
```
bajtos: Please add a test to verify that when `basePath` is set, `rootUrl` is not affected. That's the main point of `rootUrl` property, right?

I think you should move that new test and this existing test into a new describe block called `rootUrl`.

```ts
describe('RestServer (integration)', () => {
  describe('url', () => {
    // ...
  });

  describe('rootUrl', () => {
    // ...
  });

  // ...
});
```

packages/rest/src/rest.server.ts
```diff
@@ -151,7 +151,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
     return this._httpServer ? this._httpServer.listening : false;
   }
 
+  /**
+   * The base url for the server, including the basePath if set
```
bajtos: Let's add an example value please. E.g. `http://127.0.0.1:3000/api`.

packages/rest/src/rest.server.ts
```diff
@@ -151,7 +151,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+   */
   get url(): string | undefined {
+    let serverUrl = this.rootUrl;
+    if (!serverUrl) return serverUrl;
+    serverUrl = serverUrl + (this._basePath || '');
+    return serverUrl;
+  }
+
+  /**
+   * The root url for the server without the basePath
```
bajtos: Let's add an example value please. E.g. `http://127.0.0.1:3000/`Or is it `http://127.0.0.1:3000` (no trailing slash)? Please check and use the actual value.

packages/testlab/src/client.ts
```diff
@@ -33,13 +33,13 @@ export function createClientForHandler(
  * @param app A running (listening) instance of a RestApplication.
  */
 export function createRestAppClient(app: RestApplicationLike) {
-  const url = app.restServer.url;
-  if (!url) {
+  const rootUrl = app.restServer.rootUrl;
```
bajtos: Let's keep the size of this patch smaller. Also to support older version that don't expose `rootUrl` property yet, we should fall back to `url` when `rootUrl` is not set.

```suggestion
  const url = app.restServer.rootUrl || app.restServer.url;
```

packages/rest/src/rest.server.ts
```diff
@@ -151,7 +151,23 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+    if (!serverUrl) return serverUrl;
+    serverUrl = serverUrl + (this._basePath || '');
+    return serverUrl;
+  }
+
+  /**
+   * The root url for the server without the basePath. For example, the value
+   * will be 'http://localhost:3000' regardless of the `basePath`.
+   */
+  get rootUrl(): string | undefined {
```
hacksparrow: Probably `serverUrl` is more descriptive of what it actually is? `rootUrl` sounds more like the functionality provided by `url` above.

raymondfeng: It's a property on `RestServer`. `RestServer.serverUrl` seems to be confusing and it might be misunderstood as OpenAPI server url/

bajtos: Naming is important and often difficult, thank you @hacksparrow for starting this discussion.

I agree with Raymond's arguments here.

raymondfeng: @hacksparrow The `rootUrl` is only used to create the request test client at the moment. Let's go with it for now.

Issue #2491, opened 2019-02-26
Create abstraction for login related operations in authentication module


jannyHou:
### Description

Create abstraction for login service.

- [ ] Create abstraction for login service

```ts
// U stands for User model, in this story we can omit it and use `UserProfile` 
// C stands for Credentials type
export interface LoginService<U, C> {}
```

- Login service contains
  - [ ] login function `login(request: Request): Promise<U>`
    - optional to break it down into 2 functions
    - [ ] `extractCredentials(request: Request): Promise<C>`
    - [ ] `verifyCredentials(credentials: C): Promise<U>`
  - [ ] logout function `invalidateCredentials?(request: Request): Promise<boolean>`
- [ ] Create Markdown file with tests

bajtos:
Similarly to what I wrote in https://github.com/strongloop/loopback-next/issues/2435#issuecomment-469604581, I see two distinctive services here:

1. One service to deal with the business logic of credentials in a way that's independent from the transport
    - `verifyCredentials(credentials: C): Promise<U>`

2. Another service to deal with transport specific details, e.g. where to extract username & password from. 
    - `extractCredentials(request: Request): Promise<C>`
    - `invalidateCredentials?(request: Request): Promise<boolean>`

While it makes sense to me to have an interface describing a service that can convert credentials into a user profile (perform the login), I find it problematic to implement transport specific behavior in a service.

How do we envision that such service would be used from a controller? It's important to describe the `login` endpoint via OpenAPI and include specification of input parameters. Are we expecting the controller method to be decorated with `@param` decorators and then pass the entire request to the login service? That would make it too easy to introduce bugs where the controller is describing different parameters than the service extracts from the request.

IMO, extract & invalidate credentials should not be provided by any service, but should be implemented as controller actions leveraging the services described in #2435 under the hood.

```ts
class UserController {
  constructor (
    // inject response, AccessTokenService, TokenTransportStrategy and LoginService
  ) {}

  @post('/login')
  async login(
    @param() email: string,
    @param() password: string,
  ) {
    const user = await this.loginService.verifyCredentials({email, password});
    const token = await this.tokenService.generateAccessToken(user);
    await this.transportStrategy.serializeAccessToken(this.response);
   // uh oh, what is the return value of this method?
   // if the token is returned in response body, how are we going to describe response schema?
  }
    
  @post('/logout')
  async logout(
    @inject('current access token') token: string
  ) {
    await this.tokenService.invalidateAccessToken(token);
[...]

Issue #2435, opened 2019-02-20
Create abstraction for access token related operations in authentication module


jannyHou:
### Description

Create abstraction for token service

- [ ] Create abstraction for token service

```ts
// U stands for User model, in this story we can omit it and use `UserProfile` 
export interface Token Service<U extends Model> {}
```

- Token service contains
  - [ ] `generateAccessToken(user: U, options: Object): Promise<string>;`
  - [ ] `serializeAccessToken(response: Response): Promise<void>;`
  - [ ] `extractAccessToken(request: Request): Promise<string>;`
  - [ ] `verifyAccessToken(token: string): Promise<U>;`
  - [ ] `invalidateAccessToken?(token: string): Promise<boolean>;`
- [ ] ~Create unit tests~ Create Markdown file for acceptance tests (includes investigation of popular token-based authentication strategies and explain how this interface works with them).

bajtos:
> ```ts
> // U stands for User model, in this story we can omit it and use `UserProfile` 
> export interface Token Service<U extends Model> {}
> ```

I am concerned about requiring the `U` parameter to extend `Model`. What are the reasoning for requiring a Model class, instead of allowing app developers to use any interface?

> Token service contains
> - `generateAccessToken(user: U, options: Object): Promise<string>;`
> - `serializeAccessToken(response: Response): Promise<void>;`
> - `extractAccessToken(request: Request): Promise<string>;`
> - `verifyAccessToken(token: string): Promise<U>;`
> - `invalidateAccessToken?(token: string): Promise<boolean>;`

I see two distinctive services in this list.

1. A service to create/verify/invalidate access tokens. This service is independent on the transport mechanism used, e.g. a JWT token can be exchanged via REST, gRPC, GraphQL or even email!

2. A service to serialize and deserialize the token depending on the transport used (serialize a token to a response, extract a token from a request).

Conceptually, the first service is operating at Data/Repository level, dealing with behavior related to data. The second service is operating at Transport/Controller level, dealing with transport-specific behavior.

I feel these two services are orthogonal and should be implemented independently, to make it easy to compose them in different ways. For example:
- A single JWT service can be combined with different transport mechanisms - the token can be sent to the client via REST response body and read from `Authorization` header, but it can be also sent in `Set-Cookie` response header and read from request cookies.
[...]

Issue #2467, opened 2019-02-25
Refactor the authentication action to be a common one.


jannyHou:
## Description

Refactor the authentication action to be a common one that could perform `authenticate` for both non-passport based strategies and passport based strategies.

### Acceptance criteria 

- [ ] Refactor [`action`](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/authentication.provider.ts#L44) so that it could perform any strategy's authenticate method, not only passport based ones.
  - The returned strategy should be described using the `AuthenticationStrategy` interface created in #2466

bajtos:
Isn't this a necessary part of #2466: Add abstraction for authentication strategy? I mean what is a new abstraction good for, when it's not used by the authentication action?

jannyHou:
I was trying to split the abstraction story into separate ones, honestly without working on the code I couldn't say for sure does the auth action have to be refactored together with the auth strategy interface 😞

emonddr:
May involve the work needed in #2311

dhmlau:
@emonddr @jannyHou, now that PR https://github.com/strongloop/loopback-next/pull/2688 has landed, is this task done?

Issue #2312, opened 2019-01-30
Extension point for registering the authentication strategies


jannyHou:
## Description / Steps to reproduce / Feature proposal

Follow the [extension pattern example](links_to_be_add_after_pr_merged) to create extension point for registering multiple authentication strategies.

Background of the story:
Now we implement a new provider in a hardcoded way to perform the authenticate action of the custom jwt strategy created by ourselves, see [custom authentication provider](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/providers/custom.authentication.provider.ts).
This should be an extensible point that people can plugin the actions for both passport and non passport-based strategies.

## Acceptance Criteria

- [x] Create an extension point for authentication strategies
  - [x] Decorate a strategy provider class `AuthenticationStrategyProvider` with `@extensionPoint`
  - [x] The strategy provider class constructor should have a strategy getter decorated with `@extensions` 
   - [x] update the authentication action provider `AuthenticateActionProvider` to use `AuthenticationStrategy` interface and the new strategy provider class
  - [x] Document how to implement & register an auth strategy. Can be light documentation since we already have some documentation discussing extension points and extensions.
- [x] Use "jwt" and "basic" strategies for tests. Not shipping production ready strategies.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

bajtos:
IMO, we should design the extension point for auth strategies in the following way:
- The extension point is LoopBack specific, the strategy interface is defined to match the needs of LoopBack and custom LoopBack strategies. 
- To support Passport-based strategies, we should provide an adapter that will accept a Passport strategy and convert it into LB strategy (that will be part of https://github.com/strongloop/loopback-next/issues/2311 though).

bajtos:
I think this story depends on #2466: Add abstraction for authentication strategy.

jannyHou:
> To support Passport-based strategies, we should provide an adapter that will accept a Passport strategy and convert it into LB strategy

Sounds good to me. There is already an adapter in `@loopback/authenticate` we can borrow.

dhmlau:
@emonddr , with your PR #2763 merged, does it mean this task is done? Thanks.

PR #2576, opened 2019-03-12
feat: design auth system with user scenario

jannyHou:
### Description

connect to https://github.com/strongloop/loopback-next/issues/2491

(The markdown files should be part of task https://github.com/strongloop/loopback-next/issues/2435 while we wrote them when creating the interface for login service, so connect this PR to https://github.com/strongloop/loopback-next/issues/2491 instead)

This PR 

- explains/proposed a new design of auth module which could potentially support multiple auth strategies. 
- adds abstractions for ~login~ user service

When review PR, please start from the file `authentication-system.md` in the root dir of `@loopback/authentication`, the content on that page will mention about each new created .md file and take you there.

## Checklist


packages/authentication/authentication-system.md
```diff
@@ -0,0 +1,127 @@
+## Multiple Authentication strategies
```
bajtos: Typically, the files in a package root directory contain project configuration like `package.json`, linter config, etc. 

Documentation files are usually stored inside `docs` directory.

Let's move this file to `packages/authentication/docs/authentication-system.md` please.

emonddr: @jannyHou , The 'authentication-strategy' link points to authentication-action.md.

emonddr: @jannyHou , `jwt strategy` and `oauth2 strategy` link give http 404.

packages/authentication/docs/strategies/jwt.md
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      throw new HttpError[401]('User not found.');
+    }
+    return userProfile
+  };
+
+  // 1. Try to find current user
+  // 2. If found, return it
+  // 3. If not found, throw 401:
+  verify(request) {
```
emonddr: Missing complete signature

packages/authentication/docs/strategies/oauth2.md
```diff
@@ -0,0 +1,42 @@
...
+    const credentials: Credentials = await loginService.extractCredentials(request);
+    // if `userProfile` is undefined, it means the verification fails
+    const tokenAndUser: TokenAndUser = await loginService.verifyCredentials(credentials);
+
+    // Should be an extension point for different serializers
+    await tokenService.serializeAccessToken(tokenAndUser, response);
+    return tokenAndUser.userProfile;
+  };
+
+  ensureLoggedIn(request) {
```
emonddr: Missing complete signature

jannyHou:
@bajtos 
> The text seems to be based in the original ideas described in GH issues and does not seem to take into account my comments and concerns about shortcomings of that original proposal. Is that intentional?

do you mean split the abstractions into "transport related" and "transport agnostic" ones?

If you look at the [services folder](https://github.com/strongloop/loopback-next/tree/e32902eff54adfc0114ff2bfa46e3a1ef8f3b8b7/packages/authentication/src/services), we split the login service into two:

- [login service](https://github.com/strongloop/loopback-next/blob/e32902eff54adfc0114ff2bfa46e3a1ef8f3b8b7/packages/authentication/src/services/login.service.ts): agnostic of the transport layer.
- [client service](https://github.com/strongloop/loopback-next/blob/e32902eff54adfc0114ff2bfa46e3a1ef8f3b8b7/packages/authentication/src/services/client.service.ts): related to the transport layer. (The service name may not be a proper one, we can change)

Or did I miss anything?

packages/authentication/docs/authentication-action.md
```diff
@@ -0,0 +1,32 @@
+### Auth action
+
+```ts
+async action(request: Request): Promise<UserProfile | undefined> {
+    const strategy = await this.getStrategy();
+    // read the action metadata from endpoint's auth metadata like 
+    // `@authenticate('strategy_name', {actions: 'authenticate'})`
+    // type ActionType = 'authenticate' | 'ensureLoggedIn'
+    const action = await this.getAction();
```
raymondfeng: Maybe we don't need to differentiate between `login` and `verify` actions. Please note `Passport` just uses different strategies, for example:

- Local strategy for username/password based login
- BearTokenStrategy for token verification

packages/authentication/docs/controller-functions.md
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+```
+- /loginWithFB
+
+```ts
+class UserController {
+  @post('/loginWithFB', ...APISpec)
+  @authenticate('oath2.fb', {action: 'login', session: false})
+  loginWithFB() {}
```
raymondfeng: If we use `login with facebook`, the browser will be redirected to facebook login page to perform the oAuth 2.0 `implicit` or `authorization code` flow. Both of them require a callback URL. When the controller method is decorated as shown, what logic should be in `loginWithFB` method? Is it the callback function or just a place holder route to excise the FB login flow? I imagine that `/loginWithFB` will initiate the FB login.

packages/authentication/authentication-system.md
```diff
@@ -0,0 +1,125 @@
...
+
+An authentication system in a LoopBack 4 application could potentially support multiple popular strategies, including basic auth, oauth2, saml, openid-connect, etc...And also allow programmers to either token based or session based approach to track the logged in user.
+
+The diagram below illustrates the high level abstraction of such an extensible authentication system.
+
+<img src="./docs/imgs/multiple-auth-strategies-login.png" width="1000px" />
```
raymondfeng: 👍

packages/authentication/docs/authentication-action.md
jannyHou: I see, I was coupling them into one strategy.

raymondfeng:
@jannyHou Please run `npm lint:fix` to fix formatting issues reported by https://travis-ci.org/strongloop/loopback-next/jobs/505426718

packages/authentication/src/services/login.service.ts
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Logout a user. 
+   * 
+   * For JWT token: client side clears the cookie
+   * For session: delete the session entry on server/client side
+   * Code example TBD
+   * @param request The incoming Request
+   */
+  // This function should probably moved to session service?
+  invalidateCredentials?(request: Request): Promise<boolean>
```
bajtos: Since the login service is transport agnostic now, I think it should be accepting `credentials: Credentials` instead of `request: Request`?

packages/authentication/src/services/client.service.ts
```diff
@@ -0,0 +1,55 @@
...
+   * from a HTTP request header.
+   * 
+   * Or for third-party access granting, like login facebook account using oauth 2.0, load
+   * the configuration `clientId`, `client`, `scopes`, `callbackURL` from a local file.
+   * @param request The incoming Request
+   */
+  // Here I hardcoded the request type to be HTTP request,
+  // while a generic type R should be used here to support
+  // other clients like SOAP, GRPC
+  extractCredentials(request: Request): Promise<C>;
```
bajtos: What are your thoughts on the following (re-posted from https://github.com/strongloop/loopback-next/issues/2491#issuecomment-469613023)?

I find it problematic to implement transport specific behavior in a service.

How do we envision that such service would be used from a controller? It's important to describe the `login` endpoint via OpenAPI and include specification of input parameters. Are we expecting the controller method to be decorated with `@param` decorators and then pass the entire request to the login service? That would make it too easy to introduce bugs where the controller is describing different parameters than the service extracts from the request.

IMO, extract & invalidate credentials should not be provided by any service, but should be implemented as controller actions leveraging the services described in #2435 under the hood.

```ts
class UserController {
  constructor (
    // inject response, AccessTokenService, TokenTransportStrategy and LoginService
  ) {}

  @post('/login')
  async login(
    @param() email: string,
    @param() password: string,
  ) {
    const user = await this.loginService.verifyCredentials({email, password});
    const token = await this.tokenService.generateAccessToken(user);
    await this.transportStrategy.serializeAccessToken(this.response);
   // uh oh, what is the return value of this method?
   // if the token is returned in response body, how are we going to describe response schema?
  }
    
  @post('/logout')
  async logout(
    @inject('current access token') token: string
  ) {
    await this.tokenService.invalidateAccessToken(token);
  }
}
```

packages/authentication/authentication-system.md
jannyHou: updated 👍 

@bajtos I was about to merge this file into `README.md`. While to keep this PR simpler, I agree with you, removed it to the docs folder.

packages/authentication/docs/controller-functions.md
jannyHou: > Is it the callback function or just a place holder route to excise the FB login flow?

The second one "a place holder route to excise the FB login flow".

We find it pretty hard to illustrate the oauth+session strategies without implementing an example, so I created a story to explorer it then revisit the design here and update accordingly.

https://github.com/strongloop/loopback-next/issues/2624

jannyHou: My proposal would be defining a separate callback endpoint, e.g. `login/facebook` and `login/facebook/callback`

packages/authentication/src/services/login.service.ts
jannyHou: 👍 yep, we redesigned all the transport related logic into the strategy class or the controller method. Now the services are agnostic of the client.

packages/authentication/src/services/client.service.ts
jannyHou: @bajtos Yeah that's the twisted part...the login endpoint is verify special, in the [shopping example](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/controllers/user.controller.ts#L125-L154), login doesn't even invokes the authentication action...

In the new design we come up with a new proposal which moves the credential verification back into the auth strategy. The controller function only generates the access token.

Because the auth strategy's responsibility is taking in a request and returning an authenticated user profile, I suggest let developers decide handling the authentication in the auth action OR in the controller:

- Do authentication in strategy:
  - the strategy will extract and validate the user credentials from request
  - it also verifies the credentials by calling `userService.verifyCredentials()`
  - controller function generates the access token
  - pros
    - consistent with endpoints other than `/login`
  - cons
    - lose the convenience provided by REST decorator(e.g. describing parameters using `maxLength`)

- Do authentication in controller:
  - describe the credentials from request using OpenAPI spec
  - action `parseParam` will parse+validate credentials
  - controller function verify credentials(call `userService.verifyCredentials()`) and generates the access token.
  - pros
    - describe the response and request parameters in a clear way.
  - cons
    - repeating the auth strategy logic(extract+verify credentials) using OpenAPI spec+controller code
     - usually the login endpoints is countable and less than 5 for most of applications, treat them specially wouldn't be a huge effort.

packages/authentication/docs/authentication-action.md
```diff
@@ -0,0 +1,26 @@
...
+      const userProfile: UserProfile = await authStrategy.authenticate(request);
+      this.setCurrentUser(userProfile);
+      // a convenient return for the next request handlers
+      return userProfile;
+    } catch (err) {
+      // interpret the raw error code/msg here and throw the corresponding HTTP error
+      // convert it to http error
+      if (err.code == '401') {
+        throw new HttpErrors.Unauthorized(err.message);
+      }
```
hacksparrow: Shouldn't we throw for the `else` condition as well?

jannyHou: good point 👍 The code in the .md file is pseudo code so I didn't refine too much for the error catch, will handle all the errors in story https://github.com/strongloop/loopback-next/issues/2467

hacksparrow:
I wish the mechanics for enabling authentication was simpler. Like flip a switch, and `@authenticate()` just works, without having to create a provider and a custom sequence.

jannyHou:
@hacksparrow Good feedback

> I wish the mechanics for enabling authentication was simpler. Like flip a switch, and @authenticate() just works, without having to create a provider and a custom sequence.

The design in this PR covers all the details of the auth mechanics, to give extension providers/app developers the max flexibility to plugin/replace their implementation. That's the reason why "making a complete auth system" looks complicated. While we could provide a default auth system that saves people's effort. 

E.g. 

- Add an option to enable authenticate in the sequence when scaffold the app. People don't need to customize the sequence.

- Implement the default auth strategy extension point in https://github.com/strongloop/loopback-next/issues/2312, then people can authenticate an endpoint by ONLY 2 steps: 1. register their auth strategy as an extension-->2.decorate the endpoint with strategy name. Then everything automatically works!

packages/authentication/docs/authentication-action.md
```diff
@@ -0,0 +1,26 @@
+### Auth action
+
+```ts
+import * as HttpErrors from 'http-errors';
+
+async action(request: Request): Promise<UserProfile | undefined> {
+    const authStrategy = await this.getAuthStrategy();
+    if (!authStrategy) {
```
raymondfeng: How do we differentiate between the following two cases?

1. No authentication is required
2. No strategy is found to handle the authentication

packages/authentication/docs/authentication-strategy.md
```diff
@@ -0,0 +1,17 @@
+### Auth strategy interface
+
+```ts
+import {Request} from '@loopback/rest';
+
+interface AuthenticationStrategy {
+  // The resolver will read the options object from metadata, call `strategy.setOptions`
+  options: object;
+  authenticate(request: Request): Promise<UserProfile | undefined>;
+  setOptions(options: object);
```
raymondfeng: It's better to pass in the metadata via `authenticate` method, such as:
```
authenticate(request: Request, options?: ...): Promise<UserProfile | undefined>;
```

packages/authentication/docs/controller-functions.md
```diff
@@ -0,0 +1,106 @@
+## Endpoint definitions
+
+The following decorated controller functions demos the endpoints described at
+the beginning of markdown file
+[authentication-system](./authentication-system.md).
+
+Please note how they are decorated with `@authenticate()`, the syntax is:
+`@authenticate(<strategy_name>, {action: <action_name>, session: <enabled_or_not>})`
```
raymondfeng: Is `action` still needed?

packages/authentication/docs/controller-functions.md
```diff
@@ -0,0 +1,106 @@
...
+
+class LoginController{
+  constructor(
+    @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+    @inject(AuthenticationBindings.SERVICES.JWT_TOKEN) JWTtokenService: TokenService,
+  ) {}
+
+  // Describe the response using OpenAPI spec
+  @post('/loginOAI/local', RESPONSE_SPEC_FOR_JWT_LOGIN)
+  @authenticate('basicAuth')
```
raymondfeng: Do we allow a method to be authenticated by one of the strategies?
```ts
@authenticate(['basicAuth', 'localAuth']) // We allow either basic or local auth
```

packages/authentication/docs/strategies/basic-auth.md
```diff
@@ -0,0 +1,30 @@
+You could find the `AuthenticationStrategy` interface in file
+[authentication-strategy.md](./docs/authentication-strategy.md)
+
+```ts
+import {Request} from '@loopback/rest';
+
+class BasicAuthenticationStrategy implements AuthenticationStrategy {
+  options: object;
+  constructor(
+    @inject(AUTHENTICATION_BINDINGS.SERVICES.USER) userService: UserService,
```
raymondfeng: Maybe `AUTHENTICATION_BINDINGS.USER_SERVICE`? Let's constrain the namespace to be at most one level.

packages/authentication/docs/authentication-action.md
jannyHou: @raymondfeng I would expect the resolver `getAuthStrategy()` to throw an error if the strategy is not found.

We haven't added the resolver, I plan to implement it in the extension story in https://github.com/strongloop/loopback-next/issues/2312

packages/authentication/docs/controller-functions.md
jannyHou: good catch, not really

packages/authentication/docs/authentication-strategy.md
jannyHou: > Your current proposal will mandate a strategy implementation to be bound in TRANSIENT scope.

True....

packages/authentication/docs/strategies/basic-auth.md
jannyHou: makes sense

packages/authentication/docs/controller-functions.md
jannyHou: I thought "local" and "basic" auth are same, which is wrong. So the endpoint here meant to be decorated with only one strategy.

raymondfeng: We leave it for future decisions. It would be nice to support one of the many auths on the same method.

packages/authentication/docs/controller-functions.md
```diff
@@ -5,7 +5,7 @@ the beginning of markdown file
 [authentication-system](./authentication-system.md).
 
 Please note how they are decorated with `@authenticate()`, the syntax is:
-`@authenticate(<strategy_name>, {action: <action_name>, session: <enabled_or_not>})`
+`@authenticate(strategy_name, options)`
```
raymondfeng: In the future, we can allow class-level decoration of `@authenticate`.

packages/authentication/docs/controller-functions.md
```diff
@@ -48,10 +48,16 @@ class LoginController{
...
 
+  // I was about to create a local login example, while if the credentials are
+  // provided in the request body, all the authenticate logic will happen in the
+  // controller, the auth action isn't even involved.
+  // See the login endpoint in shopping example
+  // https://github.com/strongloop/loopback4-example-shopping/blob/master/src/controllers/user.controller.ts#L137
+
   // Describe the response using OpenAPI spec
-  @post('/loginOAI/local', RESPONSE_SPEC_FOR_JWT_LOGIN)
+  @post('/loginOAI/basicAuth', RESPONSE_SPEC_FOR_JWT_LOGIN)
```
raymondfeng: Do we want to clarify that the `login` method implementation can have two choices:

1. `@authenticate + user code`
2. plain `user code`

packages/authentication/docs/controller-functions.md
jannyHou: > We leave it for future decisions. It would be nice to support one of the many auths on the same method.

My understanding is:
- For "login" endpoints:
  - the number of endpoints will be small, e.g. one system supports 5 login ways
  - usually there is only one strategy applied per endpoint
- For "verify" endpoints:
  - the number of endpoints will be large
  - it makes sense to specify a default auth strategy for them instead of repeating the decoration thousands of times, OR, to be more practical, resolve the auth strategy based on the login approach.

There is an issue discussing about resolving the strategy after login:
Shall we continue the discussion in https://github.com/strongloop/loopback-next/issues/2139?

packages/authentication/docs/controller-functions.md
jannyHou: Agree. sounds like an improvement could be applied with https://github.com/strongloop/loopback-next/pull/2576#discussion_r269311285

packages/authentication/docs/controller-functions.md
jannyHou: Sounds good, would like to add https://github.com/strongloop/loopback-next/pull/2576#discussion_r267953354 here.

emonddr:
rebased and squashed commits

PR #2583, opened 2019-03-13
chore: add script to update package-level package-lock.json

raymondfeng:
This PR adds a npm script to update package-level package-lock.json files as `lerna bootstrap` does not do so. See https://github.com/lerna/lerna/issues/1415#issuecomment-455790478 for more details.

## Checklist


package.json
```diff
@@ -26,10 +26,11 @@
     "postinstall": "lerna bootstrap && npm run -s update-greenkeeper-json",
     "prerelease": "npm run build:full && npm run mocha && npm run lint",
     "release": "lerna version && lerna publish from-git --yes",
+    "update-package-lock-json": "lerna exec \"npm install -s --ignore-scripts --package-lock-only --no-audit\"",
     "update-template-deps": "node bin/update-template-deps -f",
     "update-greenkeeper-json": "node bin/update-greenkeeper-json -f",
     "sync-dev-deps": "node bin/sync-dev-deps",
-    "version": "npm run update-template-deps && npm run apidocs",
+    "version": "npm run update-package-lock-json && npm run update-template-deps && npm run apidocs",
```
b-admike: Does running `update-template-deps` before or after `update-package-lock-json` make any difference to the `package-lock.json` files that are updated?

package.json
```diff
@@ -26,10 +26,11 @@
     "postinstall": "lerna bootstrap && npm run -s update-greenkeeper-json",
     "prerelease": "npm run build:full && npm run mocha && npm run lint",
     "release": "lerna version && lerna publish from-git --yes",
+    "update-package-lock-json": "lerna exec \"npm install -s --ignore-scripts --package-lock-only --no-audit\"",
```
b-admike: Can we add `git add packages/*/package-lock.json"` at the end of the script as pointed out in https://github.com/lerna/lerna/issues/1415#issuecomment-455790478?

packages/build/package.json
```diff
@@ -43,8 +43,5 @@
   "scripts": {
     "test": "npm run mocha",
     "mocha": "node bin/run-mocha --timeout 30000 \"test/integration/*.js\""
-  },
-  "devDependencies": {
-    "@loopback/tslint-config": "^2.0.0"
```
nabdelgadir: Just curious, why was this removed?

raymondfeng: It's already listed in `dependencies`.

package.json
raymondfeng: Good catch. I forgot that.

package.json
raymondfeng: Not really.

package.json
raymondfeng: Fixed.

benchmark/package-lock.json
```diff
@@ -1,2704 +1,2798 @@
...
-		"path-is-absolute": {
-			"version": "1.0.1",
-			"resolved": "https://registry.npmjs.org/path-is-absolute/-/path-is-absolute-1.0.1.tgz",
-			"integrity": "sha1-F0uSaHNVNP+8es5r9TpanhtcX18=",
-			"dev": true
-		},
-		"path-key": {
-			"version": "2.0.1",
-			"resolved": "https://registry.npmjs.org/path-key/-/path-key-2.0.1.tgz",
-			"integrity": "sha1-QRyttXTFoUDTpLGRDUDYDMn0C0A=
```
bajtos: I think this is not right. Package lock should not list local dependencies that are linked within monorepo.

In the current setup, `npm i` (`lerna bootstrap`) installs monorepo-local packages from the npm registry and replaces any links created by `lerna`.

bajtos:
@raymondfeng I feel this change is very likely going in a wrong direction. Why is it necessary to update package-lock files when we are making new releases? The lock files are not supposed to contain entries for dependencies resolved via symlinks by lerna.

I am fixing the immediate problem in https://github.com/strongloop/loopback-next/pull/2593.

I think we should also revert the changes made in `package.json` by this pull request. Thoughts?

benchmark/package-lock.json
raymondfeng: For the record, `npm i` does not update package-lock.json for me even after I added new dependencies to a package.json. That's what led me to discover the issue and put a workaround.

raymondfeng: It seems that `lerna clean && lerna bootstrap --no-ci` does update package-lock.json for all leaf packages. I'll open a PR to use it.

raymondfeng:
> I feel this change is very likely going in a wrong direction. Why is it necessary to update package-lock files when we are making new releases? The lock files are not supposed to contain entries for dependencies resolved via symlinks by lerna.

I wanted to make sure package-lock.json files are up-to-date respecting package.json. For example, if I add a new dependency to `packages/rest/package.json`, `npm i` at `loopback-next` level does not automatically update package-lock.json and it fails CI as CI uses `npm ci` and the new dep will be missing.

Issue #1766, opened 2018-09-27
[Docs] Calling other APIs Page


virkt25:
## Description / Steps to reproduce / Feature proposal

As I was working on https://github.com/strongloop/loopback4-example-shopping/pull/20 I was referencing our docs https://loopback.io/doc/en/lb4/Calling-other-APIs-and-web-services.html and I found it to be a bit confusing ... 

- Creating the datasource using `lb4 datasource` prompts for a `baseUrl` which is not respected by `loopback-connector-rest`.
- The Service Interface section was confusing -- we should aim for 1 section showing a weakly typed example first and then an alternative stongly typed example. 
- `@serviceProxy` decorator usage doesn't explain what it takes in -- is the string the name of the datasource (what I got from code), name of service, etc. and how is it determined by the booter?
- `Get an instance of your controller` section is at a really weird place and confusing. What's the intent of this section?
- Under `Make service proxies easier to test`, the code should be reviewed. For example, the following is wrong. -- we are initializing the service instead of letting DI do it. This entire section seems a bit incomplete to me. 
```
@inject('datasources.geoService')
    protected datasource: juggler.DataSource = new GeocoderDataSource(),
```

## Expected Behavior

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

marioestradarosa:
I think this section should be updated. Considering that now we have a basic `lb4 service` CLI option.

dhmlau:
#2128 and #1766 are about the same docs page. @nabdelgadir , could you please take a look and see if there are duplicate requirements and whether we can close one and continue the discussion on one of them? Thanks.

dhmlau:
From what @nabdelgadir described:
#1766 - about fixing/updating the existing content
#2128 - about adding new content for calling other REST APIs.

dhmlau:
Looking at the content in https://loopback.io/doc/en/lb4/Calling-other-APIs-and-web-services.html, it seems to be that we no longer need to install the service-proxy directly?  

I'm thinking that we should rewrite this page as below:
- goal is to call other REST/SOAP API
   - there are lots of REST APIs out there, we can pick one, say https://swapi.co/
   - If we want to stick with SOAP, perhaps we can use the one that @raymondfeng created and deployed.

- the steps would be:
1. scaffold an LB4 app
2. add datasource using `lb4 datasource` command.  Select REST / SOAP connector depends what is decided
[...]

PR #2588, opened 2019-03-14
fix(docs): add steps to call SOAP services

dhmlau:
Fixes https://github.com/strongloop/loopback-next/issues/1766

### UPDATED 3/14/2019
The [Calling other APIs docs page](https://loopback.io/doc/en/lb4/Calling-other-APIs-and-web-services.html) is updated because we no longer need to manually install `@loopback/service-proxy`.  Instead, we can use `lb4 service`. 

The code snippets and instructions are verified with my repo: https://github.com/dhmlau/loopback4-external-apis.  

-----
## Old description 
Since we no longer need to install `@loopback/service-proxy` manually and can use `lb4 service` instead, I'm updating this doc: https://loopback.io/doc/en/lb4/Calling-other-APIs-and-web-services.html. 

_I think_ my changes can be generalized to be used for calling REST APIs as well, just that the datasource generation will be different.  Will need some work on that. 

This is currently a draft PR, because I have a question:
- Do I need to specify the `Maps WSDL binding operations to Node.js methods` when creating datasource? (see my comment in the actual code changes)

### DIFF
--- a/docs/site/Calling-other-APIs-and-Web-Services.md
+++ b/docs/site/Calling-other-APIs-and-Web-Services.md
@@ -11,68 +11,165 @@ Services, gRPC microservices, or other forms of APIs.
 
 To facilitate calling other APIs or web services, we introduce
 `@loopback/service-proxy` module to provide a common set of interfaces for
-interacting with backend services.
+interacting with backend services. You can create a `Service` class to provide a
+common set of interfaces for interacting with backend services.
 
-## Installation
+There are 3 major steps:
 
+1. **Add a datasource**: specify the service you're trying to connect
+2. **Add a service**: define how the operations/methods in the external APIs
+   will be mapped to the service methods.
+3. **Add a controller**: expose the REST APIs that will call the service methods
+
+## Add a datasource
+
+Add a datasource using the [Datasource generator](DataSource-generator.md) and
+select the corresponding connector.
+
+### Datasource for SOAP web service
+
+For calling SOAP web services, you also need to know the URL of the SOAP web
+service endpoint and its corresponding WSDL file.
+
+```sh
+$ lb4 datasource
+? Datasource name: ds
+? Select the connector for ds: SOAP webservices (supported by StrongLoop)
+? URL to the SOAP web service endpoint: http://calculator-webservice.mybluemix.net/calculator
+? HTTP URL or local file system path to the WSDL file: http://calculator-webservice.mybluemix.net/calculator?wsdl
+? Expose operations as REST APIs: Yes
+? Maps WSDL binding operations to Node.js methods:
 ```
-$ npm install --save @loopback/service-proxy
+
+For the last option `Maps WSDL binding operations to Node.js methods`, specify
+the JSON in the format of:
+
+```json
+servicMethodName: {
+  "service": "<WSDL service name>",
+  "port": "<WSDL port name>",
+  "operation": "<WSDL operation name>"
+}
 ```
 
-## Usage
+If you have more than one operations to map, it might be easier to edit the
+DataSource JSON after it's been created. See below for the example of the
+mapping of the WSDL binding operations and Node.js methods.
 
-### Define a data source for the service backend
-
-```ts
-import {juggler} from '@loopback/service-proxy';
-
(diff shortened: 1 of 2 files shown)


docs/site/Calling-other-APIs-and-Web-Services.md
```diff
@@ -9,9 +9,145 @@ permalink: /doc/en/lb4/Calling-other-APIs-and-web-services.html
...
+
+```json
+{
+  "name": "ds",
+  "connector": "soap",
+  "url": "http://calculator-webservice.mybluemix.net/calculator",
+  "wsdl": "http://calculator-webservice.mybluemix.net/calculator?wsdl",
+  "remotingEnabled": true,
+  // ADD THIS SNIPPET
+  "operations": {
```
dhmlau: @strongloop/loopback-maintainers , do I need to add this?  I tried without this snippet and it seems to be working fine.  I initially thought that by specifying the `operations` here, there's less code I need to write in the Service class.

raymondfeng: The `operations` is for customizing the method names. See https://loopback.io/doc/en/lb3/SOAP-connector.html#operations-property.

Otherwise, the method name is built from serviceName + portName + operationName

dhmlau: Thanks @raymondfeng.  The problem I'm having now is that the method name I'm setting here doesn't seem to be used for generating the methods inside the service interface, i.e. no matter what I set under the `operations` attribute, it's not being referenced in the code. Is it where it's being used?

docs/site/Calling-other-APIs-and-Web-Services.md
```diff
@@ -9,9 +9,145 @@ permalink: /doc/en/lb4/Calling-other-APIs-and-web-services.html
...
+      "port": "CalculatorPort",
+      "operation": "subtract"
+    }
+  }
+  // END OF THE SNIPPET
+}
+```
+
+For details, you can refer to the SOAP connector's operations property:
+https://github.com/strongloop/loopback-connector-soap#operations-property
```
dhmlau: will change it to https://loopback.io/doc/en/lb3/SOAP-connector.html#operations-property.

docs/site/Calling-other-APIs-and-Web-Services.md
```diff
@@ -9,9 +9,145 @@ permalink: /doc/en/lb4/Calling-other-APIs-and-web-services.html
...
+
+### Add a Controller
+
+Add a controller using the [Controller generator](Controller-generator.md) with
+the `Empty Controller` option.
+
+#### Inject the Service in the constructor
+
+```ts
+constructor(
```
bajtos:
```suggestion
  constructor(
```

dhmlau:
@bajtos , thanks for the review.  I had a question about the `operations` mapping so think that it might be easier to have a draft PR to show it.  You're right, I haven't done the cleanup yet.

docs/site/Calling-other-APIs-and-Web-Services.md
raymondfeng: It does not automatically update `CalculatorService` interface but the method names of the service proxy are impacted. We have yet to generate strongly-typed interface from the SOAP WS.

dhmlau:
@bajtos, I've restored the `@serviceProxy` and instructions to write integration tests back. 
I've also changes the README on the updated LB3 LTS date.

dhmlau:
@bajtos, I've incorporated your feedback. Also, I've updated the LB3 LTS date as part of this PR.

dhmlau:
@slnode test please
seems to be stuck at CLA check

bajtos:
> Also, I've updated the LB3 LTS date as part of this PR.

It usually better to open a new pull request for such an unrelated change.

docs/site/Calling-other-APIs-and-Web-Services.md
marioestradarosa: > I tried without this snippet and it seems to be working fine.
It would be good to hint the developer at the command line about these properties and mark them as _(optional)_

PR #2593, opened 2019-03-15
Remove monorepo-local dependencies from package locks

bajtos:
Remove local dependencies from all package-lock.json files. These dependencies are installed as symlinks by `lerna` and do not belong to package locks. Before this change, symlinks created by `lerna` were overwritten by published package versions installed by `npm ci`.

Update `DEVELOPING` documentation - explain that we use package lock files and show how to fix the locks when needed.

## Checklist


bajtos:
@raymondfeng if you agree with the changes and there are no other objections, then please go ahead and merge this pull request.

docs/site/DEVELOPING.md
```diff
@@ -126,6 +127,46 @@ npm script `lint:fix`.
 npm run lint:fix
 ```
 
+## Working with dependencies
+
+We use npm'
```
raymondfeng: `npm's`?

docs/site/DEVELOPING.md
```diff
@@ -126,6 +127,46 @@ npm script `lint:fix`.
...
 
+## Working with dependencies
+
+We use npm's
+[package-lock feature](https://docs.npmjs.com/files/package-lock.json) to speed
+up our development workflow and CI builds.
+
+When installing dependencies in individual packages, `lerna bootstrap` calls
+`npm ci` to install (deep) dependencies as specified in `package-lock.json`
+file.
```
bajtos: @raymondfeng If I understood your comments on Slack correctly, you are saying that `lerna boostrap` is not using CI mode by default? Should we expand this paragraph to explain this in more details?

docs/site/DEVELOPING.md
```diff
@@ -126,6 +127,46 @@ npm script `lint:fix`.
...
+```
+$ npx lerna clean && npx lerna bootstrap --no-ci
+```
+
+### Adding or updating dependencies
+
+Use the following command to add or update dependency `dep` in a package `name`:
+
+```
+$ npx lerna add --scope ${name} ${dep}
```
bajtos: I guess we will want to rework this section too, to explain that besides `lerna add`, dependencies can be also added directly by modifying `package.json` files and running `npm i` to update package locks. WDYT?

docs/site/DEVELOPING.md
raymondfeng: Right. See https://github.com/lerna/lerna/tree/master/commands/bootstrap#--no-ci.

> When using the default --npm-client, lerna bootstrap will call npm ci instead of npm install in **CI** environments.

I debugged the issue before and confirmed that the `ci` flag is calculated using `isCI` module if not set.

docs/site/DEVELOPING.md
raymondfeng: I agree. Maybe you can remove this section and I'll add it as part of the follow-up PR to fix the `update-package-lock-json` script.

PR #2596, opened 2019-03-15
build: improve npm scripts to update package-lock.json upon install

raymondfeng:
This is a follow-up improvement to #2593 and #2583:

1. Add `update-package-locks` script
2. Add `update-packages` to update `greenkeeper.json` and `package-lock.json` for individual packages.
3. The release script uses `npm run ci` to capture potential package-lock issues.

## Checklist


package.json
```diff
@@ -23,14 +23,16 @@
     "typescript": "^3.3.1"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap && npm run -s update-greenkeeper-json",
+    "postinstall": "npm run -s update-package-lock && npm run -s update-greenkeeper-json",
```
bajtos: Before this change, `npm install` triggered `lerna bootstrap` which did incremental update of dependencies - this is usually reasonably fast.

After this change, `npm install` triggers `update-package-lock` which cleans all dependencies first and then reinstalls them from scratch. I am concerned about the impact of this change on how long it will take to run `npm install` now.

The typical scenario:
 - Fetch new changes from GitHub: `git pull`
 - Update deps, rebuild and run tests to verify all is good: `npm it`

Can we look for a different solution for updating package locks please, one that will not degrade experience in most common use cases (on the happy path)?

I also think we should open an issue in Lerna to discuss why `lerna bootstrap` is not updating package-lock files by default, and work with the Lerna team to fix it.

To get this pull request landed sooner, I am proposing to revert changes in this line.

package.json
```diff
@@ -23,14 +23,16 @@
...
+    "ci": "npm ci --ignore-scripts && npm run -s postci",
+    "postci": "lerna bootstrap --ci",
     "prerelease": "npm run build:full && npm run mocha && npm run lint",
     "release": "lerna version && lerna publish from-git --yes",
-    "update-package-lock-json": "lerna exec \"npm install -s --ignore-scripts --package-lock-only --no-audit\" && git add \"**/package-lock.json\"",
+    "update-package-lock": "lerna clean && lerna bootstrap --no-ci",
```
bajtos: Since this command is updating many package locks, I am proposing to rename it accordingly.

```suggestion
    "update-package-locks": "lerna clean && lerna bootstrap --no-ci",
```

package.json
raymondfeng: We'll have to pay the price to ensure `package-lock.json` files are up to date so that we benefit of less time consuming on `npm ci` for both build and local development.

There are a few situations:

1. Add or remove dependencies, or update out-of-range versions in `package.json`. We must update `package-lock.json` files.

2. There are new versions of dependencies in range of what's in `package.json` files. It's nice to have to update `package-lock.json` files.

If we don't run `npm run update-package-locks` during `npm install`. We either have to do so manually and trigger it during the release process.

bajtos: I am arguing that the case when no package-lock updates are needed is much more frequent than the case when some of package-lock files need to be updated.

Anyhow, as long as `npm cit` works fast, then I can live with slow `npm install` because I am not going to use it myself.

bajtos:
@raymondfeng What happened to `ci` and `postci` scripts? You have removed them as part of the changes made from 8a4e887 to e5cb985.

```
  "ci": "npm ci --ignore-scripts && npm run -s postci",
  "postci": "lerna bootstrap --ci",
```

I really want to preserve fast install for local development. It was my understanding that the two lines above were ensuring that `npm ci` always runs fast `npm ci` in individual packages too. Without them, `npm ci` calls `postinstall` script which triggers slow clean & full reinstall.

bajtos:
I measured time to install dependencies to see how bad the difference actually is.

Your feature branch as of e5cb985a9:

```
$ time npm i

> loopback-next@0.1.0 postinstall /Users/bajtos/src/loopback/next
> lerna bootstrap

(...)

real	0m5.613s
user	0m6.447s
sys	0m1.190s
```

Current master as of a56e9118

```
$ time npm i

> loopback-next@0.1.0 postinstall /Users/bajtos/src/loopback/next
> lerna bootstrap && npm run -s update-greenkeeper-json

real	0m5.978s
user	0m6.823s
sys	0m1.261s
```

Just now I realized that you have actually reworked `postinstall` script to invoke `lerna bootstrap`, not the full clean+reinstall 🙈

Issue #2609, opened 2019-03-18
Enable TypeScript Project References


bajtos:
At the moment, our monorepo is configured in a hacky way to allow TypeScript to build individual packages independently but still allow the TypeScript Language Service run by VSCode to treat the monorepo as one large project.

Let's abandon our non-standard setup and leverage TypeScript's [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) instead.

Work done so far:
- [x] build: drop multiple dist targets, always use es2017 #1803
- [x] refactor: let TypeScript do resource copying #1824
- [x] move test files to `src/__tests__` #2283 #2316 #2330 #2339

Work remaining:

- [ ] 🚶  remove top-level `index.*` files and use `main` & `types` entries in `package.json` instead #5440 ~~#2613~~ 
- [x]  chore: add script to build typescript project references #5155 ~~#1636~~
- [x] 🚶update example projects and CLI templates (if needed) #5204

Nice to have ~~implement it or create follow-up stories:~~
- Watch mode
- [x] Incremental builds - already done by #2928
- Incremental builds on Travis CI and AppVeyor - cache intermediate files between CI runs
  - Travis CI docs: https://docs.travis-ci.com/user/caching/
  - AppVeyor docs: https://www.appveyor.com/docs/build-cache/

PR #2613, opened 2019-03-19
feat: enable declarationMap, use .d.ts files in monorepo too

bajtos:
Configure TSC to emit declaration maps, providing information needed by IDEs like VSCode to jump from usage of a symbol to its definition in the original source code.

Simplify layout of all packages, remove top-level `index.ts` and `index.d.ts` files in favor of a package.json entry

    "types": "dist/index.d.ts"

Simplify layout of module packages (excluding example applications) and remove the top-level `index.js` file in favor of a package.json entry
    "main": "dist/index.js"

Update developer documentation and project templates accordingly.

**Benefits of these changes**

- It's a preparation step for TypeScript Project References, see https://github.com/strongloop/loopback-next/pull/1636

- Build is consuming less CPU cycles now. User time on my machine used to be 5m21  before, it's 3m13 now.

   UPDATE: With lib check disabled (see #2621), the new CPU  1m48s now and the real time is 0m37s.

    - Simpler project setup in our packages, we no longer need to maintain
      dummy index files at the top level.

    Drawbacks:

    - Refactor-rename does not work across packagas. I believe this is a
      limitation of TypeScript, hopefully it will be eventually fixed.

    - Navigation across packages may not work until the monorepo is built.
- Simpler project setup in our packages, we no longer need to maintain dummy index files at the top level.

**Drawbacks**

- ~~The build is taking slightly more time now. Real time went from 0m49s to 0m56s. This should improve once we enable project references and incremental or watch build.~~

- Refactor-rename does not work across packagas. I believe this is a limitation of TypeScript, hopefully it will be eventually fixed.

- Navigation across packages may not work until the monorepo is built.

- ~~When compiling a class extending from a base class created through mixins, TypeScript emits type definitions that are violating "noImplicitAny" rule.~~

    ```ts
    declare const TodoListApplication_base;
    export declare class TodoListApplication extends TodoListApplication_base {
        constructor(options?: ApplicationConfig);
    }
    ```

  ~~As a result, when our benchmark includes `TodoApplication`, the build fails. This problem does not seem to be present when Project References are enabled. As a temporal workaround, the Todo example is exporting `src/index.ts` for the type definitions.~~

See also #2609

## Checklist


bajtos:
> As a result, when our benchmark includes TodoApplication, the build fails. This problem does not seem to be present when Project References are enabled. As a temporal workaround, the Todo example is exporting `src/index.ts` for the type definitions.

I managed to solve this problem by enabling `noLibCheck` compiler options. See https://github.com/strongloop/loopback-next/pull/2621

> Build is consuming less CPU cycles now. User time on my machine used to be 5m21 before, it's 3m13 now.
> The build is taking slightly more time now. Real time went from 0m49s to 0m56s. This should improve once we enable project references and incremental or watch build.

With `noLibCheck` in place, the compile time is actually faster than it used to be before.

```
real	0m36.980s
user	1m48.239s
sys	0m8.529s
```

raymondfeng:
> The build is taking slightly more time now. Real time went from 0m49s to 0m56s. This should improve once we enable project references and incremental or watch build.

I can live with it.

> Refactor-rename does not work across packagas. I believe this is a limitation of TypeScript, hopefully it will be eventually fixed.

Hmm, this will be annoying.

> Navigation across packages may not work until the monorepo is built.

Understandable.

> When compiling a class extending from a base class created through mixins, TypeScript emits type definitions that are violating "noImplicitAny" rule.
```ts
declare const TodoListApplication_base;
export declare class TodoListApplication extends TodoListApplication_base {
    constructor(options?: ApplicationConfig);
}
```
> As a result, when our benchmark includes TodoApplication, the build fails. This problem does not seem to be present when Project References are enabled. As a temporal workaround, the Todo example is exporting src/index.ts for the type definitions.

Is it fixed by #2621?

bajtos:
> > Refactor-rename does not work across packagas. I believe this is a limitation of TypeScript, hopefully it will be eventually fixed.
>
> Hmm, this will be annoying.

I agree it's annoying. 

As far as I understand Project References, they rely on declaration maps too, and therefore don't support cross-package renames either.

So basically we have to choose between the two options:
- current non-standard layout & build setup, cross-package renames supported
- project references, cross-package renames not supported

Considering how much benefits are unlocked by package references (think about `--watch`, `--incremental`, possibly moving our tests to Jest), and how infrequently we do cross-package renames, I am personally ok to give up cross-package renames. I also hope that this will be eventually fixed in TypeScript.

Thoughts?

> > As a result, when our benchmark includes TodoApplication, the build fails. This problem does not seem to be present when Project References are enabled. As a temporal workaround, the Todo example is exporting src/index.ts for the type definitions.>
> 
> Is it fixed by #2621?

Yes, the error message about noImplictAny violation is fixed by #2621.

The type declaration emitted by TypeScript is still using `any` under the hood, loosing information about the actual methods inherited from the Application class and added by mixins. This part is unrelated to changes presented in this pull request though, I think it's a limitation (or even a bug) in TypeScript.

raymondfeng:
Let me play with this PR with VSCode for some time to give more concrete feedbacks.

raymondfeng:
> Refactor-rename does not work across packagas. I believe this is a limitation of TypeScript, hopefully it will be eventually fixed.

`Find all references` does not work either.

bajtos:
@raymondfeng 
> `Find all references` does not work either.

That's pretty bad :( Do you happen to know if this problem persists with Project References too?

bajtos:
> - Refactor-rename does not work across packages.
> - `Find all references` does not work either.

This is tracked by the following TypeScript issue: https://github.com/microsoft/TypeScript/issues/30823

PR #2635, opened 2019-03-22
feat(context): add `binding.toAlias()`

raymondfeng:
A few improvements for `@loopback/context` related to #2259 

- Add `binding.toAlias()`
Use case:
  ```ts
  ctx.bind('parent.options').to({child: {disabled: true}});
  ctx.bind('child.options').toAlias('parent.options#child');
  const childOptions = ctx.getSync('child.options');
  expect(childOptions).to.eql({disabled: true});
  ```
Related to #2657 
Depends on #2656 

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/binding.unit.ts
+++ b/packages/context/src/__tests__/unit/binding.unit.ts
@@ -191,6 +191,64 @@ describe('Binding', () => {
     });
   });
 
+  describe('toAlias(bindingKeyWithPath)', async () => {
+    it('binds to another binding with sync value', () => {
+      ctx.bind('parent.options').to({child: {disabled: true}});
+      ctx.bind('child.options').toAlias('parent.options#child');
+      const childOptions = ctx.getSync('child.options');
+      expect(childOptions).to.eql({disabled: true});
+    });
+
+    it('creates the alias even when the target is not bound yet', () => {
+      ctx.bind('child.options').toAlias('parent.options#child');
+      ctx.bind('parent.options').to({child: {disabled: true}});
+      const childOptions = ctx.getSync('child.options');
+      expect(childOptions).to.eql({disabled: true});
+    });
+
+    it('binds to another binding with async value', async () => {
+      ctx
+        .bind('parent.options')
+        .toDynamicValue(() => Promise.resolve({child: {disabled: true}}));
+      ctx.bind('child.options').toAlias('parent.options#child');
+      const childOptions = await ctx.get('child.options');
+      expect(childOptions).to.eql({disabled: true});
+    });
+
+    it('reports error if alias binding cannot be resolved', () => {
+      ctx.bind('child.options').toAlias('parent.options#child');
+      expect(() => ctx.getSync('child.options')).to.throw(
+        /The key 'parent.options' is not bound to any value in context/,
+      );
+    });
+
+    it('reports error if alias binding cannot be resolved - async', async () => {
+      ctx.bind('child.options').toAlias('parent.options#child');
+      return expect(ctx.get('child.options')).to.be.rejectedWith(
+        /The key 'parent.options' is not bound to any value in context/,
+      );
+    });
+
+    it('allows optional if alias binding cannot be resolved', () => {
+      ctx.bind('child.options').toAlias('parent.options#child');
+      const childOptions = ctx.getSync('child.options', {optional: true});
+      expect(childOptions).to.be.undefined();
+    });
+
+    it('allows optional if alias binding cannot be resolved - async', async () => {
+      ctx.bind('child.options').toAlias('parent.options#child');
+      const childOptions = await ctx.get('child.options', {optional: true});
+      expect(childOptions).to.be.undefined();
+    });
+
+    it('sets type to ALIAS', () => {
+      const childBinding = ctx
+        .bind('child.options')
+        .toAlias('parent.options#child');
+      expect(childBinding.type).to.equal(BindingType.ALIAS);
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


raymondfeng:
> What happens when somebody attempts to change the value of an alias? Are going to rebind they key or updated the alias target instead? How about @inject.setter, should it be allowed on aliases?

These questions apply to existing binding types too. IMO, It should rebind the same key to whatever we allows, such as:
```ts
ctx.bind('existing-key').to<...>();
```

For `@injecter.setter`, we implement it as follows:
```ts
return function setter(value: unknown) {
    ctx.bind(bindingSelector).to(value);
  };
```

The binding is rebound as a constant value. Maybe `@inject.setter` (or a new `@inject.binding`) should inject an instance of `Binding` instead. This way, we can do:
```ts
binding.to<...>();
```

raymondfeng:
@bajtos Based on the discussion, I added two more commits:

1. Allow `@inject.setter.applyTemplates`
2. Make sure binding cache is cleared if its scope and/or _getValue changes.

docs/site/Decorators_inject.md
```diff
@@ -143,6 +143,29 @@ export class HelloController {
...
+existent) or update the binding. For example:
+
+```ts
+const binding = this.greetingSetter('Greetings!');
+```
+
+or
+
+```ts
+const binding = this.greetingSetter(b => b.toDynamicValue(() => 'Greetings!'));
```
raymondfeng: An alternative approach is:
```ts
this.greetingSetter().apply(b => b.toDynamicValue(() => 'Greetings!'));
```

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
```diff
@@ -110,7 +110,10 @@ describe('AuthenticateActionProvider', () => {
       strategy.setMockUser(mockUser);
       provider = new AuthenticateActionProvider(
         () => Promise.resolve(strategy),
-        u => (currentUser = u),
+        (...u) => {
+          currentUser = u[0] as UserProfile;
```
jannyHou: Is it safe to assume `u` has at least one element? I think `u` could be empty.

raymondfeng: This is just a mock-up for the test. Even when `u` is an empty array, `u[0]` returns `undefined`. The setter should be always injected for regular usage.

packages/context/src/__tests__/unit/binding.unit.ts
```diff
@@ -191,6 +191,57 @@ describe('Binding', () => {
     });
   });
 
+  describe('toAlias(bindingKeyWithPath)', async () => {
+    it('binds to another binding with sync value', () => {
+      ctx.bind('parent.options').to({child: {disabled: true}});
+      ctx.bind('child.options').toAlias('parent.options#child');
```
jannyHou: Does it matter if I bind to alias first 
`ctx.bind('child.options').toAlias('parent.options#child');`
then 
`ctx.bind('parent.options').to({child: {disabled: true}});`
?

raymondfeng: It does NOT matter.

raymondfeng: I added a test for that.

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
jannyHou: Oh I see it's in a test file.

bajtos:
Uh oh, I am not a big fan of pull requests adding multiple features. I would have preferred to see `@inject.setter` changes in a standalone patch.

bajtos:
In general, I agree we should improve `@inject.setter` to allow consumers to use different value types (`toDynamicValue`, `toAlias`, etc.) Let's find a more robust solution now.

Personally, I think I prefer `@inject.binding` most. It fully preserves simplicity and current behavior of `@inject.setter`, it also communicates well what is the intent of the new injection and what value will be injected.

raymondfeng:
I have refactored other related commits to:

- #2657
- #2656

This PR will be only responsible for adding `binding.toAlias()`. Please note it depends #2656

bajtos:

packages/context/src/__tests__/unit/binding.unit.ts
```diff
@@ -191,6 +191,64 @@ describe('Binding', () => {
...
+      const childOptions = ctx.getSync('child.options');
+      expect(childOptions).to.eql({disabled: true});
+    });
+
+    it('binds to another binding before the it is bound', () => {
```
bajtos:
```suggestion
    it('creates the alias even when the target is not bound yet', () => {
```

packages/context/src/binding.ts
```diff
@@ -506,6 +510,24 @@ export class Binding<T = BoundValue> {
...
   }
 
+  /**
+   * Bind the key to an alias of another binding
+   * @param keyWithPath Target binding key with optional path,
+   * such as `servers.RestServer.options#apiExplorer`
+   */
+  toAlias(keyWithPath: BindingAddress<T>) {
+    /* istanbul ignore if */
+    if (debug.enabled) {
```
bajtos: Do we need this check? AFAICT, the debug statement below is not using any expensive computation.

raymondfeng: We use it consistently in this file with `/* istanbul ignore if */` so that test coverage is not dragged down by the `debug` statement.

Issue #2389, opened 2019-02-14
Mount Express router with OpenAPI spec


bajtos:
_This is a follow-up for the spike #2318, see the discussion and PoC code for more context._

Implement `app.mountExpressRouter(basePath, router, spec?)` to allow LB4 app developers to add arbitrary set of Express routes and provide OpenAPI spec

```ts
mountExpressRouter(
  basePath: string,
  router: ExpressRequestHandler,
  spec?: RouterSpec,
): void;
```

The new API should build on the foundation laid down by StaticAssetsRoute and allows LB4 app
developers to add arbitrary set of Express routes and provide OpenAPI spec.

As a result, LB4 will have the following request-handling pipeline:

1. Request-preprocessing middleware like CORS, this is not customizable yet.
2. Native LoopBack 4 routes (controller methods or route handlers).
3. External Express routers (if the request was not handled yet).
4. Static assets (if the request was not handled yet).

Related issues: #1849

## Acceptance criteria

- [x] Refactor `StaticAssetsRoute` into `ExternalExpressRoutes` and `ExternalRoute`, see PoC in [external-express-routes.ts](https://github.com/strongloop/loopback-next/blob/379e0c050c49e70713cabc9aa6d85a892250d375/packages/rest/src/router/external-express-routes.ts) --> #2524
- [ ] Add `mountExpressRouter` to both `RestServer `and `RestApplication` classes.
- [ ] Update API docs
- [ ] Describe the new API in [Routes](https://loopback.io/doc/en/lb4/Routes.html), include a simple example based on `express.Router` (not LB3).

PR #2318, opened 2019-01-31
PoC showing how to mount an LB3 app and include its REST API in OpenAPI spec [DO NOT MERGE]

bajtos:
A proof of concept showing how to expose a LB3 app ([loopback-getting-started](https://github.com/strongloop/loopback-getting-started)) via LB4 REST API (see #2301). The solution presented in this pull request is a somewhat alternate approach to #2274, see the discussion in that PR for context.

Under the hood, the solution has the following major parts:
- `app.mountExpressRouter(basePath, router, spec)` allows LB4 app
developers to add arbitrary set of Express routes and provide OpenAPI spec.
- `Lb3AppBooter` responsible for setting up LoopBack 3 application, obtaining
OpenAPI spec describing app's REST API and mounting the app on LB4.
- Example app

See [SPIKE NOTES](https://github.com/strongloop/loopback-next/blob/spike/add-lb3app-to-lb4/SPIKE-NOTES.md) for further details.

Related: https://github.com/strongloop/loopback-next/issues/1849

## Checklist

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #2643, opened 2019-03-25
feat(rest): add mountExpressRouter

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/2389.

Add `mountExpressRouter` to `RestApplication` and `RestServer` and include in `Routes.md`.

## Checklist

### DIFF
--- a/docs/site/Routes.md
+++ b/docs/site/Routes.md
@@ -226,3 +226,64 @@ export class MyApplication extends RestApplication {
   }
 }
 ```
+
+## Mounting an Express Router
+
+If you have an existing [Express](https://expressjs.com/) application that you
+want to use with LoopBack 4, you can mount the Express application on top of a
+LoopBack 4 application. This way you can mix and match both frameworks, while
+using LoopBack as the host. You can also do the opposite and use Express as the
+host by mounting LoopBack 4 REST API on an Express application. See
+[Creating an Express Application with LoopBack REST API](express-with-lb4-rest-tutorial.md)
+for the tutorial.
+
+Mounting an Express router on a LoopBack 4 application can be done using the
+`mountExpressRouter` function provided by both
+[`RestApplication`](http://apidocs.loopback.io/@loopback%2fdocs/rest.html#RestApplication)
+and
+[`RestServer`](http://apidocs.loopback.io/@loopback%2fdocs/rest.html#RestServer).
+
+Example use:
+
+{% include note.html content="
+Make sure [express](https://www.npmjs.com/package/express) is installed.
+" %}
+
+{% include code-caption.html content="src/express-app.ts" %}
+
+```ts
+import {Request, Response} from 'express';
+import * as express from 'express';
+
+const legacyApp = express();
+
+// your existing Express routes
+legacyApp.get('/pug', function(_req: Request, res: Response) {
+  res.send('Pug!');
+});
+
+export {legacyApp};
+```
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import {RestApplication} from '@loopback/rest';
+
+const legacyApp = require('./express-app').legacyApp;
+
+const openApiSpecForLegacyApp: RouterSpec = {
+  // insert your spec here, your 'paths', 'components', and 'tags' will be used
+};
+
+class MyApplication extends RestApplication {
+  constructor(/* ... */) {
+    // ...
+
+    this.mountExpressRouter('/dogs', legacyApp, openApiSpecForLegacyApp);
+  }
+}
+```
+
+Any routes you define in your `legacyApp` will be mounted on top of the `/dogs`
+base path, e.g. if you visit the `/dogs/pug` endpoint, you'll see `Pug!`.
--- a/packages/rest/src/__tests__/integration/rest.application.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.application.integration.ts
@@ -5,9 +5,17 @@
 
 import {anOperationSpec} from '@loopback/openapi-spec-builder';
 import {Client, createRestAppClient, expect} from '@loopback/testlab';
+import * as express from 'express';
+import {Request, Response} from 'express';
 import * as fs from 'fs';
 import * as path from 'path';
-import {RestApplication, RestServer, RestServerConfig, get} from '../..';
+import {
+  get,
+  RestApplication,
+  RestServer,
+  RestServerConfig,
+  RouterSpec,
+} from '../..';
 
 const ASSETS = path.resolve(__dirname, '../../../fixtures/assets');
 
@@ -161,6 +169,103 @@ describe('RestApplication (integration)', () => {
     client = createRestAppClient(restApp);
     const response = await client.get('/custom/ping').expect(304);
     await client.get(response.header.location).expect(200, 'Hi');
+  });
+
+  context('mounting an Express router on a LoopBack application', async () => {
+    beforeEach('set up RestApplication', async () => {
+      givenApplication();
+      await restApp.start();
+      client = createRestAppClient(restApp);
+    });
+
+    it('gives precedence to an external route over a static route', async () => {
+      const router = express.Router();
+      router.get('/', function(_req: Request, res: Response) {
+        res.send('External dog');
+      });
+
+      restApp.static('/dogs', ASSETS);
+      restApp.mountExpressRouter('/dogs', router);
+
+      await client.get('/dogs/').expect(200, 'External dog');
+    });
+
+    it('mounts an express Router without spec', async () => {
+      const router = express.Router();
+      router.get('/poodle/', function(_req: Request, res: Response) {
+        res.send('Poodle!');
+      });
+      router.get('/pug', function(_req: Request, res: Response) {
+        res.send('Pug!');
+      });
+      restApp.mountExpressRouter('/dogs', router);
+
+      await client.get('/dogs/poodle/').expect(200, 'Poodle!');
+      await client.get('/dogs/pug').expect(200, 'Pug!');
+    });
+
+    it('mounts an express Router with spec', async () => {
+      const router = express.Router();
+      function greetDogs(_req: Request, res: Response) {
+        res.send('Hello dogs!');
+      }
+
+      const spec: RouterSpec = {
+        paths: {
+          '/hello': {
+            get: {
+              responses: {
+                '200': {
+                  description: 'greet the dogs',
+                  content: {
+                    'text/plain': {
+                      schema: {type: 'string'},
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      };
+      router.get('/hello', greetDogs);
+      restApp.mountExpressRouter('/dogs', router, spec);
+      await client.get('/dogs/hello').expect(200, 'Hello dogs!');
+
+      const openApiSpec = restApp.restServer.getApiSpec();
+      expect(openApiSpec.paths).to.deepEqual({
+        '/dogs/hello': {
+          get: {
+            responses: {
+              '200': {
+                description: 'greet the dogs',
+                content: {'text/plain': {schema: {type: 'string'}}},
+              },
+            },
+          },
+        },
+      });
+    });
+
+    it('mounts more than one express Router', async () => {
+      const router = express.Router();
+      router.get('/poodle', function(_req: Request, res: Response) {
+        res.send('Poodle!');
+      });
+
+      restApp.mountExpressRouter('/dogs', router);
+
+      const secondRouter = express.Router();
+
+      secondRouter.get('/persian', function(_req: Request, res: Response) {
+        res.send('Persian cat.');
+      });
+
+      restApp.mountExpressRouter('/cats', secondRouter);
+
+      await client.get('/dogs/poodle').expect(200, 'Poodle!');
+      await client.get('/cats/persian').expect(200, 'Persian cat.');
+    });
   });
 
   function givenApplication(options?: {rest: RestServerConfig}) {
--- a/packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
+++ b/packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
@@ -1,0 +1,151 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {assignRouterSpec, RouterSpec} from '../../../';
+
+describe('assignRouterSpec', () => {
+  it('duplicates the additions spec if the target spec is empty', async () => {
+    const target: RouterSpec = {paths: {}};
+    const additions: RouterSpec = {
+      paths: {
+        '/': {
+          get: {
+            responses: {
+              '200': {
+                description: 'greeting',
+                content: {
+                  'application/json': {
+                    schema: {type: 'string'},
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+      components: {
+        schemas: {
+          Greeting: {
+            type: 'object',
+            properties: {
+              message: {
+                type: 'string',
+              },
+            },
+          },
+        },
+      },
+      tags: [{name: 'greeting', description: 'greetings'}],
+    };
+
+    assignRouterSpec(target, additions);
+    expect(target).to.eql(additions);
+  });
+
+  it('does not assign components without schema', async () => {
+    const target: RouterSpec = {
+      paths: {},
+      components: {},
+    };
+
+    const additions: RouterSpec = {
+      paths: {},
+      components: {
+        parameters: {
+          addParam: {
+            name: 'add',
+            in: 'query',
+            description: 'number of items to add',
+            required: true,
+            schema: {
+              type: 'integer',
+              format: 'int32',
+            },
+          },
+        },
+        responses: {
+          Hello: {
+            description: 'Hello.',
+          },
+        },
+      },
+    };
+
+    assignRouterSpec(target, additions);
+    expect(target.components).to.be.empty();
+  });
+
+  it('uses the route registered first', async () => {
+    const originalPath = {
+      '/': {
+        get: {
+          responses: {
+            '200': {
+              description: 'greeting',
+              content: {
+                'application/json': {
+                  schema: {type: 'string'},
+                },
+              },
+            },
+          },
+        },
+      },
+    };
+
+    const target: RouterSpec = {paths: originalPath};
+
+    const additions: RouterSpec = {
+      paths: {
+        '/': {
+          get: {
+            responses: {
+              '200': {
+                description: 'additional greeting',
+                content: {
+                  'application/json': {
+                    schema: {type: 'string'},
+                  },
+                },
+              },
+              '404': {
+                description: 'Error: no greeting',
+                content: {
+                  'application/json': {
+                    schema: {type: 'string'},
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    };
+
+    assignRouterSpec(target, additions);
+    expect(target.paths).to.eql(originalPath);
+  });
+
+  it('does not duplicate tags from the additional spec', async () => {
+    const target: RouterSpec = {
+      paths: {},
+      tags: [{name: 'greeting', description: 'greetings'}],
+    };
+    const additions: RouterSpec = {
+      paths: {},
+      tags: [
+        {name: 'greeting', description: 'additional greetings'},
+        {name: 'salutation', description: 'salutations!'},
+      ],
+    };
+
+    assignRouterSpec(target, additions);
+    expect(target.tags).to.containDeep([
+      {name: 'greeting', description: 'greetings'},
+      {name: 'salutation', description: 'salutations!'},
+    ]);
+  });
+});
--- a/packages/rest/src/__tests__/unit/router/rebase-openapi-spec.unit.ts
+++ b/packages/rest/src/__tests__/unit/router/rebase-openapi-spec.unit.ts
@@ -1,0 +1,144 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {rebaseOpenApiSpec} from '../../../router';
+
+describe('rebaseOpenApiSpec', () => {
+  it('does not modify an OpenAPI spec if it does not have paths', async () => {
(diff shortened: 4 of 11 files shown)


packages/rest/src/router/external-express-routes.ts
```diff
@@ -42,6 +50,17 @@ export class ExternalExpressRoutes {
...
+    router: ExpressRequestHandler,
+    spec: RouterSpec = {paths: {}},
+  ) {
+    this._externalRoutes.use(basePath, router);
+
+    spec = rebaseOpenApiSpec(spec, basePath);
+    assignRouterSpec(this._specForExternalRoutes, spec);
+  }
+
   find(request: Request): ResolvedRoute {
```
jannyHou: @nabdelgadir When we was debugging it I told you the problem is raised by undefined `spec` while that's not true, don't be mislead by that wrong assersion :) I think the real problem is this:

For the endpoint your are testing `/dog/pug`, the route lookup happens in the following way:
1. in your app/server's sequence: `const route = this.findRoute(request);`
2. goes to the http handler's [`findRoute` function](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/http-handler.ts#L61-L64)
3. calls the find method in routing table and falls to the [external route condition](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/router/routing-table.ts#L138-L145)
4. finally reaches the `find` method at line#64 here

Now if you look at the implementation of the function `find`, it always returns an external route instanciated with `this._staticRoutes`, WHILE your endpoint is registered by calling `mountRouter` method, which calls `this._externalRoutes.user(basePath, router)` to register the router/handler.

To make a non-static external route work, I believe you need to do some change in the `find` method and also the routing table file:

- IMO an external route is NOT equal to a static route, but [our code](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/router/routing-table.ts#L138-L145) assumes it's always static, this could be told from the debug message and also the `find` method here

- Improve the `find` method here to distinguish between static routes and external(non-static) routes, so that `/dog/pug` instantiate `ExternalRoute` with router `_externalRoutes`, instead of `_staticRoutes`

cc @bajtos @hacksparrow may know more about the routing stuff, ^ correct me if wrong :) thanks.

jannyHou: quickly hacked code to 
```ts

  find(request: Request): ResolvedRoute {
    // NOT a solution, just a hack
    return new ExternalRoute(
      this._externalRoutes,
      request.method,
      request.url,
      {
        description: 'LoopBack static assets route',
        'x-visibility': 'undocumented',
        responses: {},
      },
    );
  }
```
And your test pass. So I think your tests are written in the correct way(create a new express router then register routes to it)

bajtos: I don't have time to look into this in details today. If you look at my PoC in https://github.com/strongloop/loopback-next/pull/2318, you can see that the resolved route is accepting both static and external routes.

https://github.com/strongloop/loopback-next/blob/379e0c050c49e70713cabc9aa6d85a892250d375/packages/rest/src/router/external-express-routes.ts#L85-L91

When that route is invoked, it calls the external router to handle the request fist, and if the request is not handled, then it calls static assets router second.

https://github.com/strongloop/loopback-next/blob/379e0c050c49e70713cabc9aa6d85a892250d375/packages/rest/src/router/external-express-routes.ts#L102-L114

bajtos: > Improve the find method here to distinguish between static routes and external(non-static) routes, so that /dog/pug instantiate ExternalRoute with router _externalRoutes, instead of _staticRoutes

I don't think it's feasible to tell at `findRoute` time whether the request will match an external route or a static-assets route.

jannyHou: > When that route is invoked, it calls the external router to handle the request fist, and if the request is not handled, then it calls static assets router second.

> I don't think it's feasible to tell at findRoute time whether the request will match an external route or a static-assets route.

Thanks! Got it.

docs/site/Routes.md
```diff
@@ -226,3 +226,40 @@ export class MyApplication extends RestApplication {
...
+import {RestApplication, RouterSpec} from '@loopback/rest';
+import {Request, Response} from 'express';
+import * as express from 'express';
+
+const app = new RestApplication();
+const router = express.Router();
+
+router.get('/pug', function(_req: Request, res: Response) {
+  res.send('Pug!');
+});
```
bajtos: Since our primary motivation for enabling this feature is to allow existing Express applications to be mounted in LB4, I would prefer if this example was more Express-app centric.

```ts
const legacyApp = express();
// your existing Express routes
legacyApp.get('/pug', function(_req: Request, res: Response) {
  res.send('Pug!');
});

// compose them with the new LB4 app
const spec: RouterSpec = {
  // insert your spec here, your 'paths', 'components', and 'tags' will be used
};
	
app.mountExpressRouter('/dogs', legacyApp, spec);
```

Also, I think `mountExpressRouter` should be typically called from the application constructor. To promote this best practice, I would prefer the example to use the same approach too.

```ts
const legacyApp = require('./express-app');

const legacySpec = {
   // insert your spec here, your 'paths', 'components', and 'tags' will be used
}

class MyApplication extends RestApplication {
  constructor(/*...*/) {
    // ...
   
    this.mountExpressRouter('/dogs', legacyApp, legacySpec);
  }
}
```

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -5,9 +5,12 @@
 
 import {anOperationSpec} from '@loopback/openapi-spec-builder';
 import {Client, createRestAppClient, expect} from '@loopback/testlab';
+import * as express from 'express';
+import {Request, Response} from 'express';
 import * as fs from 'fs';
 import * as path from 'path';
-import {RestApplication, RestServer, RestServerConfig, get} from '../..';
+import {get, RestApplication, RestServer, RestServerConfig} from '../..';
+import {RouterSpec} from '../../router/router-spec';
```
bajtos: Can we import `RouterSpec` from `../..`, similarly as we do for other types?

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -163,6 +166,73 @@ describe('RestApplication (integration)', () => {
     await client.get(response.header.location).expect(200, 'Hi');
   });
 
+  context('mounting an express router on a LoopBack application', async () => {
+    givenApplication();
+
+    const router = express.Router();
+    await restApp.start();
+    client = createRestAppClient(restApp);
```
bajtos: Test setup belongs to a `before` or `beforeEach` hook.

That way the setup code is run only when the tests inside this `context` block are actually executed (think of `it.only` used elsewhere). More importantly, when the setup is executed from a hook, any errors are correctly associated with the relevant test case(s).

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -163,6 +166,73 @@ describe('RestApplication (integration)', () => {
...
+      router.get('/pug', function(_req: Request, res: Response) {
+        res.send('Pug!');
+      });
+      restApp.mountExpressRouter('/dogs', router);
+
+      await client.get('/dogs/poodle/').expect(200, 'Poodle!');
+      await client.get('/dogs/pug').expect(200, 'Pug!');
+    });
+
+    it.skip('mounts an express Router with spec with paths', async () => {
```
bajtos: `it.skip` - I assume this is not implemented yet?

packages/rest/src/rest.application.ts
```diff
@@ -264,4 +270,22 @@ export class RestApplication extends Application implements HttpServerLike {
...
+
+  /**
+   * Mount an Express router to expose additional REST endpoints handled
+   * via legacy Express-based stack.
+   *
+   * @param basePath Path where to mount the router at, e.g. `/` or `/api`.
+   * @param router The Express router to handle the requests.
+   * @param spec A partial OpenAPI spec describing endpoints provided by the router.
+   * LoopBack will prepend `basePath` to all endpoints automatically. Use `undefined`
+   * if you don't want to document the routes.
```
bajtos:
```diff
- Use `undefined` if you don't want to document the routes.
+ This argument is optional. You can leave it out if you don't want to document the routes.
```

packages/rest/src/rest.server.ts
```diff
@@ -827,6 +829,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+
+  /**
+   * Mount an Express router to expose additional REST endpoints handled
+   * via legacy Express-based stack.
+   *
+   * @param basePath Path where to mount the router at, e.g. `/` or `/api`.
+   * @param router The Express router to handle the requests.
+   * @param spec A partial OpenAPI spec describing endpoints provided by the router.
+   * LoopBack will prepend `basePath` to all endpoints automatically. Use `undefined`
+   * if you don't want to document the routes.
```
bajtos:
```diff
- Use `undefined` if you don't want to document the routes.
+ This argument is optional. You can leave it out if you don't want to document the routes.
```

packages/rest/src/router/external-express-routes.ts
```diff
@@ -42,12 +53,29 @@ export class ExternalExpressRoutes {
...
-      'x-visibility': 'undocumented',
-      responses: {},
-    });
+    return new ExternalRoute(
+      this._externalRoutes,
+      this._staticRoutes,
+      request.method,
+      request.url,
+      {
+        description: 'LoopBack static assets route',
```
bajtos: This description is no longer true, please update it. E.g.

```suggestion
        description: 'External route or a static asset',
```

packages/rest/src/router/external-express-routes.ts
```diff
@@ -90,6 +126,24 @@ class ExternalRoute implements RouteEntry, ResolvedRoute {
   }
 }
 
+export function rebaseOpenApiSpec<T extends Partial<OpenApiSpec>>(
```
bajtos: Please add few unit-level tests to verify how `rebaseOpenApiSpec` handles different edge cases. Take a look at `if` conditions below to see what to test.

packages/rest/src/router/router-spec.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {OpenApiSpec} from '@loopback/openapi-v3-types';
+
+export type RouterSpec = Pick<OpenApiSpec, 'paths' | 'components' | 'tags'>;
+
+export function assignRouterSpec(target: RouterSpec, additions: RouterSpec) {
```
bajtos: Similarly here. Please add few unit-level tests to verify how `assignRouteSpec` handles different edge cases. Take a look at `if` conditions and explanatory code comments below to see what to test.

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -163,6 +166,73 @@ describe('RestApplication (integration)', () => {
     await client.get(response.header.location).expect(200, 'Hi');
   });
 
+  context('mounting an express router on a LoopBack application', async () => {
```
bajtos: Please add a test to verify the order of precedence of external vs. static-asset routes.

```ts
app.staticAssets('/test', /* path to some dir */);
app.mountExpressRouter('/test', /* a router */);
// ...

await client.get(
 // a URL matching a static asset and an Express route too
).expect(
  // the express route was invoked
);
```

docs/site/Routes.md
```diff
@@ -226,3 +226,57 @@ export class MyApplication extends RestApplication {
   }
 }
 ```
+
+## Mounting an Express Router
+
+If you have an existing [Express](https://expressjs.com/) application that you
+want to use with LoopBack 4, you can mount the Express application on top of a
+LoopBack 4 application. This way you can mix and match both frameworks, while
+using LoopBack as the host. Mounting an Express router on a LoopBack 4
```
bajtos: Lovely 👏 

Can we also add a link to docs explaining how to compose LB4 and Express the other way, where Express is acting as the host and LB4 routes are mounted?

docs/site/Routes.md
```diff
@@ -226,3 +226,57 @@ export class MyApplication extends RestApplication {
...
+import * as express from 'express';
+
+const legacyApp = express();
+
+// your existing Express routes
+legacyApp.get('/pug', function(_req: Request, res: Response) {
+  res.send('Pug!');
+});
+
+module.exports = legacyApp;
```
bajtos: Since we are TypeScript, would the following work too?

```suggestion
export {legacyApp};
```
```

File: packages/rest/src/__tests__/integration/rest.application.integration.ts [hunk 14]
```diff
@@ -163,6 +171,90 @@ describe('RestApplication (integration)', () => {
...
+                },
+              },
+            },
+          },
+        },
+      };
+      router.get('/hello', greetDogs);
+      restApp.mountExpressRouter('/dogs', router, spec);
+      await client.get('/dogs/hello').expect(200, 'Hello dogs!');
+    });
```
bajtos: In this test, it's important to verify that `GET /hello` endpoint is included in the spec returned by `restApp.restServer.getApiSpec()`.

packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
```diff
@@ -0,0 +1,20 @@
...
+  const target: RouterSpec = {paths: {}, components: {}, tags: []};
+  const additions: RouterSpec = {paths: {}, components: {}, tags: []};
+
+  it('duplicates the additions spec if the target spec is empty', async () => {});
+
+  it('does not assign components without schema', async () => {});
+
+  it('uses the route registered first', async () => {});
+
+  it('does not duplicate tags from the additional spec', async () => {});
```
bajtos: Nice list of test cases :+1:

packages/rest/src/router/routing-table.ts
```diff
@@ -137,7 +137,7 @@ export class RoutingTable {
 
     if (this._externalRoutes) {
       debug(
-        'No API route found for %s %s, trying to find a static asset',
+        'No API route found for %s %s, trying to find external Express route',
```
bajtos:
```suggestion
        'No API route found for %s %s, trying to find an external Express route',
```

docs/site/Routes.md
```diff
@@ -226,3 +226,62 @@ export class MyApplication extends RestApplication {
...
 ```
+
+## Mounting an Express Router
+
+If you have an existing [Express](https://expressjs.com/) application that you
+want to use with LoopBack 4, you can mount the Express application on top of a
+LoopBack 4 application. This way you can mix and match both frameworks, while
+using LoopBack as the host. You can also do the opposite and use Express as the
+host by mounting LoopBack 4 REST API on an Express application. See
+[Creating an Express Application with LoopBack REST API](express-with-lb4-rest-tutorial.md)
```
bajtos: Should we update `express-with-lb4-rest-tutorial.md` to link back to this section too?

docs/site/Routes.md
```diff
@@ -226,3 +226,62 @@ export class MyApplication extends RestApplication {
...
+for the tutorial.
+
+Mounting an Express router on a LoopBack 4 application can be done using the
+`mountExpressRouter` function provided by both `RestApplication` and
+`RestServer`.
```
bajtos: How about replacing naked class names with links to API docs?

- http://apidocs.loopback.io/@loopback%2fdocs/rest.html#RestApplication
- http://apidocs.loopback.io/@loopback%2fdocs/rest.html#RestServer

packages/rest/src/__tests__/unit/router/rebase-openapi-spec.unit.ts
```diff
@@ -0,0 +1,143 @@
...
+              },
+            },
+          },
+        },
+      },
+    };
+
+    rebaseOpenApiSpec(spec, '/greetings');
+
+    expect(spec).to.equal(spec);
```
bajtos: This is a [Tautology](https://en.wikipedia.org/wiki/Tautology_(logic)) - you are asserting that `spec` is a pointer to the same object as pointed to by `spec`.

I think you should check that `spec.paths` has entries `/` and `/hello` after the rebase.

```suggestion
    expect(Object.keys(spec.paths)).to.deepEqual('/', '/hello');
```

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -163,6 +171,103 @@ describe('RestApplication (integration)', () => {
...
+      });
+
+      restApp.static('/dogs', ASSETS);
+      restApp.mountExpressRouter('/dogs', router);
+
+      await client.get('/dogs/').expect(200, 'External dog');
+    });
+
+    it('mounts an express Router without spec', async () => {
+      const router = express.Router();
```
b-admike: What would the default generated(?) spec look like here if one is not provided? Can we add an assertion for it?

packages/rest/src/__tests__/integration/rest.application.integration.ts
```diff
@@ -163,6 +171,103 @@ describe('RestApplication (integration)', () => {
...
+      client = createRestAppClient(restApp);
+    });
+
+    it('gives precedence to an external route over a static route', async () => {
+      const router = express.Router();
+      router.get('/', function(_req: Request, res: Response) {
+        res.send('External dog');
+      });
+
+      restApp.static('/dogs', ASSETS);
```
b-admike: Do we silently replace the static route in this case?

packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
```diff
@@ -0,0 +1,151 @@
...
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    };
+
+    assignRouterSpec(target, additions);
```
b-admike: IIUC, we can make this assertion a little more clearer for others:
```suggestion
    const resultSpec = assignRouterSpec(target, additions);
    expect(resultSpec.paths).to.equal(target.paths);
```

packages/rest/src/__tests__/unit/router/rebase-openapi-spec.unit.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {rebaseOpenApiSpec} from '../../../router';
+
+describe('rebaseOpenApiSpec', () => {
+  it('does not modify the OpenApiSpec if it does not have paths', async () => {
```
b-admike: Nitpick:
```suggestion
  it('does not modify an OpenAPI spec if it does not have paths', async () => {
```

packages/rest/src/__tests__/unit/router/rebase-openapi-spec.unit.ts
```diff
@@ -0,0 +1,144 @@
...
+                description: 'OK',
+              },
+            },
+          },
+        },
+      },
+    };
+
+    let rebasedSpec = rebaseOpenApiSpec(spec, '');
+    expect(spec).to.eql(rebasedSpec);
```
b-admike: nitpick (we usually expect the result to match some expected value):
```suggestion
    expect(rebasedSpec).to.eql(spec);
```

packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
nabdelgadir: `assignRouterSpec` has a `void` return type, but I changed `expect(target.paths).to.equal(target.paths);` to `expect(target.paths).to.eql(target.paths);`

packages/rest/src/__tests__/integration/rest.application.integration.ts
nabdelgadir: The `spec` is optional, so if it isn't provided then it won't get generated or added to `openapi.json`.

packages/rest/src/__tests__/integration/rest.application.integration.ts
nabdelgadir: When it's looking for the route, it first checks if there's an external route, then if it can't find it then it uses the static one (see [here](https://github.com/strongloop/loopback-next/pull/2643/files#diff-4ac1e4b2069b2ff62309cfdd6b9dbe7bR103)), so I believe the static one still exists, but is overshadowed by the external one.

docs/site/Routes.md
```diff
@@ -226,3 +226,64 @@ export class MyApplication extends RestApplication {
...
+```
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import {RestApplication} from '@loopback/rest';
+
+const legacyApp = require('./express-app').legacyApp;
+
+const legacySpec: RouterSpec = {
```
raymondfeng: The variable name `legacySpec` is confusing. It should be something like `openApiSpecForLegacyApp`.

packages/rest/src/__tests__/unit/router/assign-router-spec.unit.ts
```diff
@@ -0,0 +1,151 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {assignRouterSpec, RouterSpec} from '../../../router';
```
raymondfeng: It's better to import from `../../..` (the `@loopback/rest` module).

packages/rest/src/rest.application.ts
```diff
@@ -264,4 +270,23 @@ export class RestApplication extends Application implements HttpServerLike {
...
+   *
+   * @param basePath Path where to mount the router at, e.g. `/` or `/api`.
+   * @param router The Express router to handle the requests.
+   * @param spec A partial OpenAPI spec describing endpoints provided by the
+   * router. LoopBack will prepend `basePath` to all endpoints automatically.
+   * This argument is optional. You can leave it out if you don't want to
+   * document the routes.
+   */
+  mountExpressRouter(
+    basePath: string,
```
raymondfeng: Do we allow `RegExp`?

nabdelgadir: An express router can take in `RegExp` so it's possible, but I'll do this in a different PR so this one can be landed and I can also add tests.

raymondfeng: Maybe it's not practical as OpenAPI spec does not support regexp/wildcard for `path`.

PR #1169, opened 2018-03-22
Introduce typed binding key

bajtos:
Modify Context, Binding, and related methods to allow developers to add a type information to individual binding keys. These binding keys will allow the compiler to verify that a correct value is passed to binding setters (`.to()`, `.toClass()`, etc.) and automatically infer the type of the value returned by `ctx.get()` and `ctx.getSync()`.

~~This is an early prototype to discuss whether we want to follow the proposed direction.~~

@raymondfeng @strongloop/lb-next-dev @strongloop/loopback-next thoughts?

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] Related API Documentation was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `packages/example-*` were updated
- [ ] Documentation updates

PR #2656, opened 2019-03-28
fix(context): clear binding cache upon scope/getValue changes

raymondfeng:
Extracted from #2635:

Make sure binding cache is cleared if the binding is bound to a new value provider or scope
- The binding instance may cache values for contexts based on the scope
- If an existing binding changes its value provider or scope, the cache should be invalidated.

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/binding.unit.ts
+++ b/packages/context/src/__tests__/unit/binding.unit.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {expect} from '@loopback/testlab';
+import {expect, sinon, SinonSpy} from '@loopback/testlab';
 import {
   Binding,
   BindingScope,
@@ -211,6 +211,68 @@ describe('Binding', () => {
       await expect(ctx.get(binding.key)).to.be.rejectedWith(
         /Binding foo is not bound to a value yet/,
       );
+    });
+  });
+
+  describe('cache', () => {
+    let spy: SinonSpy;
+    beforeEach(() => {
+      spy = sinon.spy();
+    });
+
+    it('clears cache if scope changes', () => {
+      const indexBinding = ctx
+        .bind<number>('index')
+        .toDynamicValue(spy)
+        .inScope(BindingScope.SINGLETON);
+
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledOnce(spy);
+      spy.resetHistory();
+
+      // Singleton
+      ctx.getSync(indexBinding.key);
+      sinon.assert.notCalled(spy);
+      spy.resetHistory();
+
+      indexBinding.inScope(BindingScope.CONTEXT);
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledOnce(spy);
+    });
+
+    it('clears cache if _getValue changes', () => {
+      const providerSpy = sinon.spy();
+      class IndexProvider implements Provider<number> {
+        value() {
+          return providerSpy();
+        }
+      }
+      const indexBinding = ctx
+        .bind<number>('index')
+        .toDynamicValue(spy)
+        .inScope(BindingScope.SINGLETON);
+
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledOnce(spy);
+      spy.resetHistory();
+
+      // Singleton
+      ctx.getSync(indexBinding.key);
+      sinon.assert.notCalled(spy);
+      spy.resetHistory();
+
+      // Now change the value getter
+      indexBinding.toProvider(IndexProvider);
+      ctx.getSync(indexBinding.key);
+      sinon.assert.notCalled(spy);
+      sinon.assert.calledOnce(providerSpy);
+      spy.resetHistory();
+      providerSpy.resetHistory();
+
+      // Singleton
+      ctx.getSync(indexBinding.key);
+      sinon.assert.notCalled(spy);
+      sinon.assert.notCalled(providerSpy);
     });
   });
 
--- a/packages/context/src/binding.ts
+++ b/packages/context/src/binding.ts
@@ -212,6 +212,15 @@ export class Binding<T = BoundValue> {
   }
 
   /**
+   * Clear the cache
+   */
+  private _clearCache() {
+    if (!this._cache) return;
+    // WeakMap does not have a `clear` method
+    this._cache = new WeakMap();
+  }
+
+  /**
    * This is an internal function optimized for performance.
    * Users should use `@inject(key)` or `ctx.get(key)` instead.
    *
@@ -346,6 +355,7 @@ export class Binding<T = BoundValue> {
    * @param scope Binding scope
    */
   inScope(scope: BindingScope): this {
+    if (this._scope !== scope) this._clearCache();
     this._scope = scope;
     return this;
   }
@@ -357,7 +367,7 @@ export class Binding<T = BoundValue> {
    */
   applyDefaultScope(scope: BindingScope): this {
     if (!this._scope) {
-      this._scope = scope;
+      this.inScope(scope);
     }
     return this;
   }
@@ -367,6 +377,8 @@ export class Binding<T = BoundValue> {
    * @param getValue getValue function
    */
   private _setValueGetter(getValue: ValueGetter<T>) {
+    // Clear the cache
+    this._clearCache();
     this._getValue = getValue;
   }
 
--- a/packages/context/src/binding.ts
+++ b/packages/context/src/binding.ts
@@ -3,11 +3,16 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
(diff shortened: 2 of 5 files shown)


packages/context/src/__tests__/unit/binding.unit.ts
```diff
@@ -214,6 +214,45 @@ describe('Binding', () => {
...
+        .bind<number>('index')
+        .toDynamicValue(() => {
+          return index++;
+        })
+        .inScope(BindingScope.SINGLETON);
+      expect(ctx.getSync(indexBinding.key)).to.equal(0);
+      // Singleton
+      expect(ctx.getSync(indexBinding.key)).to.equal(0);
+      indexBinding.inScope(BindingScope.CONTEXT);
+      expect(ctx.getSync(indexBinding.key)).to.equal(1);
```
bajtos: I find these test little bit difficult to understand, because I have to map from `0`/`1` values to whether the value getter was called or was not.

Could you please try to rewrite these tests using a Sinon spy? A mock-up to illustrate what I mean:

```ts
  const spy = sinon.spy();

// ...
  .toDynamicValue(spy);

  ctx.getSync(indexBinding.key)
  sinon.assertCalled(spy);

  ctx.getSync(indexBinding.key)
  sinon.assertNotCalled(spy);

  indexBinding.inScope(BindingScope.CONTEXT);
  sinon.assertCalled(spy);
```

The same comment applies to the next test too.

packages/context/src/binding.ts
```diff
@@ -328,11 +349,26 @@ export class Binding<T = BoundValue> {
...
+      this.inScope(scope);
     }
     return this;
   }
 
+  /**
+   * Set the `_getValue` function
+   * @param getValueFn getValue function
+   */
+  private _setGetValue(
```
bajtos: I find this name difficult to understand. How about the following?

```suggestion
  private _setValueGetter(
```

packages/context/src/binding.ts
```diff
@@ -162,9 +166,9 @@ export class Binding<T = BoundValue> {
 
   private _cache: WeakMap<Context, T>;
   private _getValue: (
-    ctx?: Context,
-    session?: ResolutionSession,
-  ) => ValueOrPromise<T>;
+    ctx: Context,
+    optionsOrSession?: ResolutionOptionsOrSession,
+  ) => ValueOrPromise<T | undefined>;
```
bajtos: I am concerned about this change, we don't want to start leaking `undefined` to code that was expecting a `T` value instead. I'll take a closer look tomorrow.

packages/context/src/__tests__/unit/binding.unit.ts
raymondfeng: 👍

packages/context/src/binding.ts
raymondfeng: Please note `_getValue` is an internal method.

raymondfeng: This change is needed to allow an alias binding to resolve to the `undefined` with `optional: true` even when the target binding does not exist.

bajtos: > This change is needed to allow an alias binding to resolve to the `undefined` with `optional: true` even when the target binding does not exist.

Thanks for explaining this scenario, I agree it's something we need to take care of.

So far, our type definitions for Context and Binding methods went a long way to allow the compiler to catch programming errors related to `undefined` values.

Cross-posting https://github.com/strongloop/loopback-next/pull/1169#issuecomment-375362267

> Consider the following REST binding:
> 
> ```ts
> export const HOST = new BindingKey<string | undefined>('rest.host');
> ```
>
> Before my change, there was no indication that undefined is a valid value for the HOST binding. Without that knowledge, one could write following code:
>
> ```ts
> const resolve = promisify(dns.resolve);
>
> const host = await ctx.get<string>(RestBindings.HOST);
> const records = await resolve(host);
> // etc.
> ```
> 
> The compiler will happily accept such code and only at runtime we will learn that HOST may be also undefined in which case our code needs to find the server's host name using a different way.

Now what I would like to avoid is the situation when an alias binding is defined with value of type `T`, but the actual runtime type is `T | undefined`.

Let me review the actual code changes to see if this concern is relevant or not.

bajtos: Hey, binding aliases are not part of this pull request! Let's revert this change please and discuss it in #2635 as part of `toAlias` implementation.

packages/context/src/binding.ts
```diff
@@ -328,11 +349,26 @@ export class Binding<T = BoundValue> {
...
 
+  /**
+   * Set the `_getValue` function
+   * @param getValueFn getValue function
+   */
+  private _setValueGetter(
+    getValueFn: (
+      ctx: Context,
+      optionsOrSession?: ResolutionOptionsOrSession,
+    ) => ValueOrPromise<T | undefined>,
```
bajtos: Have you considered defining a named type, e.g. `BindingValueGetter`? This type definition is repeated in two places (the type of `_getValue` property, the type of `_setValueGetter` argument).

packages/context/src/binding.ts
```diff
@@ -246,11 +262,15 @@ export class Binding<T = BoundValue> {
         }
       }
     }
+    optionsOrSession = asResolutionOptions(optionsOrSession);
```
bajtos:
```suggestion
    const options = asResolutionOptions(optionsOrSession);
```

After this line, we always have the options object, right?

packages/context/src/binding.ts
```diff
@@ -328,11 +349,26 @@ export class Binding<T = BoundValue> {
...
   }
 
+  /**
+   * Set the `_getValue` function
+   * @param getValueFn getValue function
+   */
+  private _setValueGetter(
+    getValueFn: (
+      ctx: Context,
+      optionsOrSession?: ResolutionOptionsOrSession,
```
bajtos: What is the reasoning for asking every value getter variant to support both flavors of `options` and `session`? From what I see in other parts of this patch, it adds unnecessary complexity to getter implementations.

IMO, `ResolutionOptionsOrSession` should be accepted only by `getValue` method for backwards compatibility. Under the hood, `getValue` should "normalize" that argument to `ResolutionOptions` and use that options in all places going forward.

packages/context/src/resolution-session.ts
```diff
@@ -343,3 +339,22 @@ export interface ResolutionOptions {
...
+/**
+ * Normalize ResolutionOptionsOrSession to ResolutionOptions
+ * @param optionsOrSession resolution options or session
+ */
+export function asResolutionOptions(
+  optionsOrSession?: ResolutionOptionsOrSession,
+): ResolutionOptions {
+  // backwards compatibility
+  if (optionsOrSession instanceof ResolutionSession) {
+    optionsOrSession = {session: optionsOrSession};
```
bajtos: Can we avoid mutating arguments please?

```suggestion
    return {session: optionsOrSession};
```

packages/context/src/__tests__/unit/binding.unit.ts
```diff
@@ -214,6 +214,56 @@ describe('Binding', () => {
...
+        .toDynamicValue(spy)
+        .inScope(BindingScope.SINGLETON);
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledOnce(spy);
+      // Singleton
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledOnce(spy);
+      indexBinding.inScope(BindingScope.CONTEXT);
+      ctx.getSync(indexBinding.key);
+      sinon.assert.calledTwice(spy);
```
bajtos: The test is easier to understand now. I am proposing to improve it even more and use [`spy.resetHistory()`](https://sinonjs.org/releases/v7.3.1/spies/#spyresethistory) in between context calls.

```ts
ctx.getSync(indexBinding.key);
sinon.assert.calledOnce(spy);

// Singleton is caching the value
spy.resetHistory();
ctx.getSync(indexBinding.key);
sinon.assert.notCalled(spy);

// Scope change clears the cache
indexBinding.inScope(BindingScope.CONTEXT);
spy.resetHistory();
ctx.getSync(indexBinding.key);
sinon.assert.calledOnce(spy);
```

packages/context/src/binding.ts
raymondfeng: I refactored it into a separate commit. Please also note that `Binding.getValue()` still returns `ValueOrPromise<T>`. Only the `_getValue` is changed.

packages/context/src/binding.ts
```diff
@@ -128,6 +133,11 @@ export type BindingTag = TagMap | string;
  */
 export type BindingTemplate<T = unknown> = (binding: Binding<T>) => void;
 
+type ValueGetter<T> = (
+  ctx: Context,
+  options: ResolutionOptions,
+) => ValueOrPromise<T | undefined>;
```
bajtos: I am really not comfortable with this change and need more time to research possible ramifications.

If you can change this line to return `ValueOrPromise<T>` then this PR could be landed sooner.

```suggestion
) => ValueOrPromise<T>;
```

PR #2657, opened 2019-03-28
feat(context): improve @inject.setter and add @inject.binding

raymondfeng:
Extracted from https://github.com/strongloop/loopback-next/pull/2635

1. Allow `@inject.setter` to set other forms of value providers than constant

```ts
// Set the binding to a constant value, this is the current behavior
injectedSetter('my-value'); 

// or use the returned binding to configure it
const binding = injectedSetter();
binding.toClass(MyClass);
```

2. Allows `bindingCreation` option to control how the underlying binding is resolved/created.

## Checklist

### DIFF
--- a/docs/site/Decorators_inject.md
+++ b/docs/site/Decorators_inject.md
@@ -126,7 +126,19 @@ class MyControllerWithGetter {
 
 `@inject.setter` injects a setter function to set the bound value of the key.
 
-Syntax: `@inject.setter(bindingKey: string)`.
+Syntax: `@inject.setter(bindingKey: BindingAddress, {bindingCreation?: ...})`.
+
+The `setter` function injected has the following signature:
+
+```ts
+export type Setter<T> = (value?: T) => void;
+```
+
+The binding resolution/creation is controlled by `bindingCreation` option. See
+[@inject.binding](#injectbinding) for possible settings.
+
+The following example shows the usage of `@inject.setter` and the injected
+setter function.
 
 ```ts
 export class HelloController {
@@ -141,6 +153,70 @@ export class HelloController {
     return defaultGreet;
   }
 }
+```
+
+Please note the setter function simply binds a `value` to the underlying binding
+using `binding.to(value)`.
+
+To set other types of value providers such as `toDynamicValue`or `toClass`, use
+`@inject.binding` instead.
+
+### @inject.binding
+
+`@inject.binding` injects a binding for the given key. It can be used to bind
+various types of value providers to the underlying binding or configure the
+binding. This is an advanced form of `@inject.setter`, which only allows to set
+a constant value (using `Binding.to(value)` behind the scene) to the underlying
+binding.
+
+Syntax: `@inject.binding(bindingKey: BindingAddress, {bindingCreation?: ...})`.
+
+```ts
+export class HelloController {
+  constructor(
+    @inject.binding('greeting') private greetingBinding: Binding<string>,
+  ) {}
+
+  @get('/hello')
+  async greet() {
+    // Bind `greeting` to a factory function that reads default greeting
+    // from a file or database
+    this.greetingBinding.toDynamicValue(() => readDefaultGreeting());
+    return await this.greetingBinding.get<string>(this.greetingBinding.key);
+  }
+}
+```
+
+The `@inject.binding` takes an optional `metadata` object which can contain
+`bindingCreation` to control how underlying binding is resolved or created based
+on the following values:
+
+```ts
+/**
+ * Policy to control if a binding should be created for the context
+ */
+export enum BindingCreationPolicy {
+  /**
+   * Always create a binding with the key for the context
+   */
+  ALWAYS_CREATE = 'Always',
+  /**
+   * Never create a binding for the context. If the key is not bound in the
+   * context, throw an error.
+   */
+  NEVER_CREATE = 'Never',
+  /**
+   * Create a binding if the key is not bound in the context. Otherwise, return
+   * the existing binding.
+   */
+  CREATE_IF_NOT_BOUND = 'IfNotBound',
+}
+```
+
+For example:
+
+```ts
+@inject.setter('binding-key', {bindingCreation: BindingCreationPolicy.NEVER_CREATES})
 ```
 
 ### @inject.tag
--- a/packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
+++ b/packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
@@ -3,13 +3,13 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {expect} from '@loopback/testlab';
 import {Context, instantiateClass} from '@loopback/context';
 import {Request} from '@loopback/rest';
-import {AuthenticateFn, UserProfile, AuthenticationBindings} from '../../..';
-import {MockStrategy} from '../fixtures/mock-strategy';
+import {expect} from '@loopback/testlab';
 import {Strategy} from 'passport';
+import {AuthenticateFn, AuthenticationBindings, UserProfile} from '../../..';
 import {AuthenticateActionProvider} from '../../../providers';
+import {MockStrategy} from '../fixtures/mock-strategy';
 
 describe('AuthenticateActionProvider', () => {
   describe('constructor()', () => {
--- a/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
@@ -5,6 +5,8 @@
 
 import {expect} from '@loopback/testlab';
 import {
+  Binding,
+  BindingCreationPolicy,
   BindingKey,
   BindingScope,
   Constructor,
@@ -169,6 +171,25 @@ describe('Context bindings - Injecting dependencies of classes', () => {
     expect(await store.getter()).to.equal('456');
   });
 
+  it('creates getter from a value', () => {
+    const getter = Getter.fromValue('data');
+    expect(getter).to.be.a.Function();
+    return expect(getter()).to.be.fulfilledWith('data');
+  });
+
+  it('reports an error if @inject.getter has a non-function target', async () => {
+    ctx.bind('key').to('value');
+
+    class Store {
+      constructor(@inject.getter('key') public getter: string) {}
+    }
+
+    ctx.bind(STORE_KEY).toClass(Store);
+    expect(() => ctx.getSync<Store>(STORE_KEY)).to.throw(
+      'The type of Store.constructor[0] (String) is not a Getter function',
+    );
+  });
+
   describe('in SINGLETON scope', () => {
     it('throws if a getter cannot be resolved by the owning context', async () => {
       class Store {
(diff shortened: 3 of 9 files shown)


docs/site/Decorators_inject.md
```diff
@@ -143,6 +143,66 @@ export class HelloController {
...
+```
+
+```ts
+const binding = this.greetingSetter('Greetings!');
+```
+
+or
+
+```ts
+const binding = this.greetingSetter().toDynamicValue(() => 'Greetings!');
```
bajtos: @raymondfeng could you please remind me and other people reviewing this pull request why are we adding this feature? What use cases are you trying to support?

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
```diff
@@ -110,7 +110,10 @@ describe('AuthenticateActionProvider', () => {
       strategy.setMockUser(mockUser);
       provider = new AuthenticateActionProvider(
         () => Promise.resolve(strategy),
-        u => (currentUser = u),
+        (...u) => {
```
bajtos: I don't see how is this change related to the enhanced setter. Could you please explain the reasoning, or revert this part back?

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -17,6 +17,7 @@ import {
   ResolutionSession,
   Setter,
 } from '../..';
+import {BindingCreationPolicy} from '../../context';
```
bajtos: Please import directly from `../..` and merge the import with the lines above.

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -12,6 +12,7 @@ import {
   Context,
   isPromiseLike,
 } from '../..';
+import {BindingCreationPolicy} from '../../context';
```
bajtos: Ditto, please import from `../..`

packages/context/src/context.ts
```diff
@@ -737,6 +741,30 @@ export class Context extends EventEmitter {
...
+    key: BindingAddress<T>,
+    policy?: BindingCreationPolicy,
+  ) {
+    let binding: Binding<T>;
+    if (policy === BindingCreationPolicy.ALWAYS_CREATE) {
+      binding = this.bind(key);
+    } else if (policy === BindingCreationPolicy.NEVER_CREATE) {
+      binding = this.getBinding(key);
+    } else {
+      if (this.isBound(key)) {
```
bajtos: Let's reduce nesting:

```ts
} else if (this.isBound(key)) {
```

packages/context/src/inject.ts
```diff
@@ -188,7 +188,9 @@ export function inject(
 /**
  * The function injected by `@inject.getter(bindingSelector)`.
  */
-export type Getter<T> = () => Promise<T>;
+export interface Getter<T> {
+  (): Promise<T>;
+}
```
bajtos: What is the reasoning for this change? Do you need an interface so that you can inherit from it?

packages/context/src/inject.ts
```diff
@@ -201,9 +203,32 @@ export namespace Getter {
 }
 
 /**
- * The function injected by `@inject.setter(key)`.
+ * The function injected by `@inject.setter(bindingKey)`.
  */
-export type Setter<T> = (value: T) => void;
+export interface Setter<T> {
```
bajtos: What is the reasoning for this change? Do you need an interface so that you can inherit from it?

Also the API docs seems little weird to me - it says `the function` but describes an interface.

packages/context/src/inject.ts
```diff
@@ -351,15 +376,22 @@ function resolveAsSetter(ctx: Context, injection: Injection) {
       `The type of ${targetName} (${targetType.name}) is not a Setter function`,
     );
   }
-  const bindingSelector = injection.bindingSelector;
-  if (!isBindingAddress(bindingSelector)) {
+  const key = injection.bindingSelector;
+  if (!isBindingAddress(key)) {
```
bajtos: Can we keep `bindingSelector` please? I find that name more descriptive than just `key`.

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
raymondfeng: It was a left-over and should be reverted now.

packages/context/src/inject.ts
raymondfeng: I changed them to interfaces to define `Setter.applyTemplates()`. Now they can be reverted.

docs/site/Decorators_inject.md
raymondfeng: Added more description.

raymondfeng:
> Personally, I prefer UNIX philosophy - many small tools where each tool does one job and does it well. In that light, I would prefer a new decorator for injecting the entire Binding instance. You have mentioned @inject.binding yourself in one of the earlier pull requests, what was the reasoning for your decision to put this new feature into the existing @inject.setter function?

I'm open here but I would like to avoid a new `@inject.*` flavor before it's necessary. Here is my reasoning:

`@inject.setter` was introduced to bind a `value (getter)` to the underlying key. The 1st version only allows constant value (binding.to()) by calling `setter(constantValue)`. This PR extends the capability to have more control by exposing the Binding instance. The initial version was to allow `BindingTemplate` functions to be accepted by the setter function but we cannot differentiate `constant function value` from `template function`.

If we decide to introduce `@inject.binding`, maybe we should deprecate and finally remove `@inject.setter`. It's similar as preferring `@inject(filterByTag('my-tag')` over `@inject.tag('my-tag')`.

raymondfeng:
@bajtos I went ahead to add `@inject.binding` and leave `@inject.setter` to only set constant values.

bajtos:
> I went ahead to add `@inject.binding` and leave `@inject.setter` to only set constant values.

Awesome 👍 

Here is my point of view:

 - I find it confusing to have a decorator called `@inject.setter` to inject a `Binding` instance.
 - Ideally, IoC related code (like `Binding` instances) should not leak into the actual implementation (the code receiving the dependencies via injection).

With `@inject.setter`, the decorated argument/property receives a generic setter function that can be passed to any 3rd-party code (e.g. a package from npm), the consumer does not need to understand anything about our `Context` and `Binding`.

In that sense, `@inject.binding` can be seen as a kind of an anti-pattern. Having said that, I am ok to add that functionality as an advanced tool for situations where `@inject.setter` is not good enough. Once we learn more about use cases where `@inject.binding` is needed, we can look into ways how to provide more clean API(s) to support those cases.

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -342,6 +343,86 @@ describe('Context', () => {
     });
   });
 
+  describe('findOrCreateBinding', () => {
+    describe('bindingCreationPolicy - ALWAYS_CREATE', () => {
```
bajtos: I think we usually use `context` instead of `describe` to group tests of the same subject but under different conditions.

```suggestion
    context('with ALWAYS_CREATE', () => {
```

This way the tests read like:

- findOrCreateBinding with ALWAYS_CREATE creates a new binding even the key is bound
- findOrCreateBinding with ALWAYS_CREATE creates a new binding if not bound
- ...

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -342,6 +343,86 @@ describe('Context', () => {
...
+      it('creates a new binding if not bound', () => {
+        const binding = ctx.findOrCreateBinding(
+          'a-new-key',
+          BindingCreationPolicy.ALWAYS_CREATE,
+        );
+        expect(binding.key).to.eql('a-new-key');
+      });
+    });
+
+    describe('bindingCreationPolicy - NEVER_CREATE', () => {
```
bajtos: Ditto, use `context` instead of `describe`.

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -342,6 +343,86 @@ describe('Context', () => {
...
+        expect(() =>
+          ctx.findOrCreateBinding(
+            'a-new-key',
+            BindingCreationPolicy.NEVER_CREATE,
+          ),
+        ).to.throw(/The key 'a-new-key' is not bound to any value in context/);
+      });
+    });
+
+    describe('bindingCreationPolicy - CREATE_IF_NOT_BOUND', () => {
```
bajtos: And here too - use `context` instead of `describe`.

packages/context/src/__tests__/unit/context.unit.ts
```diff
@@ -342,6 +343,86 @@ describe('Context', () => {
...
+      it('creates a new binding if the key is not bound', () => {
+        const binding = ctx.findOrCreateBinding(
+          'a-new-key',
+          BindingCreationPolicy.CREATE_IF_NOT_BOUND,
+        );
+        expect(binding.key).to.eql('a-new-key');
+      });
+    });
+
+    describe('default bindingCreationPolicy - CREATE_IF_NOT_BOUND', () => {
```
bajtos: Ditto, use `context` instead of `describe`.

packages/context/src/inject.ts
```diff
@@ -249,12 +274,20 @@ export namespace inject {
...
   export const setter = function injectSetter(
     bindingKey: BindingAddress,
-    metadata?: InjectionMetadata,
+    metadata?: InjectBindingMetadata,
   ) {
     metadata = Object.assign({decorator: '@inject.setter'}, metadata);
     return inject(bindingKey, metadata, resolveAsSetter);
   };
 
+  export const binding = function injectBinding(
```
bajtos: Please add a tsdoc comment for this new API.

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
```diff
@@ -110,7 +110,10 @@ describe('AuthenticateActionProvider', () => {
       strategy.setMockUser(mockUser);
       provider = new AuthenticateActionProvider(
         () => Promise.resolve(strategy),
-        u => (currentUser = u),
+        u => {
+          currentUser = u;
+          return new Binding('authentication.currentUser').to(currentUser);
```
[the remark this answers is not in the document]
raymondfeng: It's a mock-up return a binding object.

bajtos:

packages/context/src/inject.ts
```diff
@@ -201,9 +203,31 @@ export namespace Getter {
 }
 
 /**
- * The function injected by `@inject.setter(key)`.
+ * The function injected by `@inject.setter(bindingKey)`. It sets the underlying
+ * binding to a const value. Returns the `Binding` object so that the binding
+ * can be further configured, such as setting it to a class with `toClass()`.
```
bajtos: I am not convinced it's a good idea to return the binding from the setter. I don't want us to leak details about our IoC container (Context/Binding classes) to code receiving dependencies. As I wrote above, I see `Setter` as a concept that's agnostic of DI/IoC and I'd like to preserve that.

For use cases where the developer needs to set more than just the bound value, there is the new `@inject.binding` decorator to give full control over the binding.

I am also concerned about robustness of the proposed solution. It allows developers to invoke the injected setter function without any value provided:

```ts
setter(); 
```

What's the expected outcome? For example, if `T` is `string | undefined`, the call above should be valid and should set the target binding value to `undefined`.

For the reasons outline above, I am against changing the signature of `Setter` function and propose to keep it in the current form.

raymondfeng: I can revert the return type back to `void`. What about the `bindingCreation` option?

raymondfeng: reverted.

raymondfeng:
@bajtos PTAL

raymondfeng:
@bajtos Ping.

bajtos:
> What about the `bindingCreation` option?

This option is configured at DI/IoC level in `@inject.setter` arguments, it makes sense to keep it available. As I see it, it allows the developer that's configuring DI wiring to tweak the behavior of the setter that will be injected.

packages/context/src/inject.ts
```diff
@@ -249,16 +269,49 @@ export namespace inject {
...
+  ) {
+    metadata = Object.assign({decorator: '@inject.binding'}, metadata);
+    return inject(bindingKey, metadata, resolveAsBinding);
+  };
+
   /**
    * Inject an array of values by a tag pattern string or regexp
    *
-   * @example
+   * For example,
```
bajtos:
```suggestion
   * For example:
```

PR #2650, opened 2019-03-27
[greenkeeper] Greenkeeper/default/semver 6.0.0

raymondfeng:
## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #2661, opened 2019-03-29
build: add script to check package-lock files

bajtos:
- Add a new script `bin/check-package-lock.js` to verify that local monorepo dependencies are excluded from package-lock files.
- Run this script as part of `npm run lint`.
- Fix `sandbox/example` package - add `.npmrc` to enable package-lock feature, remove a monorepo-local references from `sandbox/examples/package-lock.json`.

My intention is to prevent accidental addition of monorepo-local references to package locks, see https://github.com/strongloop/loopback-next/pull/2650#pullrequestreview-219976910

## Checklist

### DIFF
--- a/bin/check-package-locks.js
+++ b/bin/check-package-locks.js
@@ -1,0 +1,65 @@
+#!/usr/bin/env node
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback-next
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * This is an internal script to verify that local monorepo dependencies
+ * are excluded from package-lock files.
+ */
+'use strict';
+
+const path = require('path');
+const fs = require('fs');
+const promisify = require('util').promisify;
+
+const readFile = promisify(fs.readFile);
+
+const Project = require('@lerna/project');
+
+async function checkPackageLocks() {
+  const project = new Project(process.cwd());
+  const packages = await project.getPackages();
+  const rootPath = project.rootPath;
+  const lockFiles = packages.map(p =>
+    path.relative(rootPath, path.join(p.location, 'package-lock.json')),
+  );
+
+  const checkResults = await Promise.all(
+    lockFiles.map(async lockFile => {
+      return {lockFile, violations: await checkLockFile(lockFile)};
+    }),
+  );
+  const badPackages = checkResults.filter(r => r.violations.length > 0);
+  if (!badPackages.length) return true;
+
+  console.error('\nInvalid package-lock entries found!\n');
+  for (const {lockFile, violations} of badPackages) {
+    console.error('  %s', lockFile);
+    for (const v of violations) {
+      console.error('    -> %s', v);
+    }
+  }
+
+  console.error('\nRun the following command to fix the problems:');
+  console.error('\n  $ npm run update-package-locks\n');
+  return false;
+}
+
+if (require.main === module) {
+  checkPackageLocks().then(
+    ok => process.exit(ok ? 0 : 1),
+    err => {
+      console.error(err);
+      process.exit(2);
+    },
+  );
+}
+
+async function checkLockFile(lockFile) {
+  const data = JSON.parse(await readFile(lockFile, 'utf-8'));
+  return Object.keys(data.dependencies || []).filter(dep =>
+    dep.startsWith('@loopback/'),
+  );
+}
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "coverage:ci": "node packages/build/bin/run-nyc report --reporter=text-lcov | coveralls",
     "precoverage": "npm test",
     "coverage": "open coverage/index.html",
-    "lint": "npm run prettier:check && npm run tslint",
+    "lint": "npm run prettier:check && npm run tslint && node bin/check-package-locks",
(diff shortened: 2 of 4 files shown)


sandbox/example/package-lock.json
```diff
@@ -1,15 +1,3 @@
...
-  "dependencies": {
-    "@loopback/tslint-config": {
-      "version": "2.0.2",
-      "dev": true,
-      "requires": {
-        "tslint-consistent-codestyle": "^1.14.1"
-      }
-    }
-  }
+  "lockfileVersion": 1
```
bajtos: Unfortunately, lerna does not create package-lock files when a package does not have any external dependencies - see https://github.com/lerna/lerna/issues/2015.

I ran `npm init && npm install` in an empty project to see how a lock file looks for an empty project and used the same content for our sandbox example.

b-admike: Sounds good. Also, would disabling `package-lock` for this sandbox example be a good alternative?

package.json
```diff
@@ -37,7 +37,7 @@
     "coverage:ci": "node packages/build/bin/run-nyc report --reporter=text-lcov | coveralls",
     "precoverage": "npm test",
     "coverage": "open coverage/index.html",
-    "lint": "npm run prettier:check && npm run tslint",
+    "lint": "npm run prettier:check && npm run tslint && node bin/check-package-locks",
```
b-admike: Have you considered having this check in a different build step than `lint`?

bin/check-package-locks.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const checkResults = await Promise.all(
+    lockFiles.map(async lockFile => {
+      return {lockFile, violations: await checkLockFile(lockFile)};
+    }),
+  );
+  const badPackages = checkResults.filter(r => r.violations.length > 0);
+  if (!badPackages.length) return 0;
+
+  console.error();
```
b-admike: Can we not use `\n` for new lines with console.error (I'm assuming that's why we have the two calls before and after the `invalid` lock file message)?

bin/check-package-locks.js
```diff
@@ -0,0 +1,69 @@
...
+  console.error('Run the following command to fix the problems:');
+  console.error();
+  console.error('  $ npm run update-package-locks');
+  console.error();
+}
+
+if (require.main === module) {
+  checkPackageLocks().then(
+    result => process.exit(result),
```
b-admike: To clarify, `result` will be populated if the condition in L35 is satisfied right? Otherwise, whenever we hit the console.error() calls in `checkPackageLocks`, then `err` is populated with them?

bajtos:
Cross-posting from our conversation with @raymondfeng:

> It seems that GK submits package-lock.json that includes `@loopback/*` deps

That's pretty bad! It means that with my check in place, all builds triggered by GreenKeeper will fail and thus GK will start opening new issues whenever a new version of our dependency is published :cry:

OTOH:

- If we disable GreenKeeper, then we can end up running out CI jobs against old version of our dependencies and thus do not discover problems that our users may encounter.
- If we let GreenKeeper to add `@loopback/*` dependencies to our lock files, then `npm i` will install monorepo packages from npm instead of creating symlinks, and thus changes made across multiple packages will not compile.

I really wish npm had support for workspaces - see https://npm.community/t/npm-needs-a-more-robust-story-for-local-multi-module-development/1059/8

bin/check-package-locks.js
bajtos: Are you asking me to use `\n` instead of calling `console.error()`?

bin/check-package-locks.js
bajtos: Oh, good catch. My intention was that `checkPackageLocks` should return `1` when there was a problem. Let me fix that!

package.json
bajtos: Initially, I put the check into `.travis.yml` only. But then I realized that we don't want developers to discover problem with package-locks only after opening a pull request. It's better to find that out while running `npm test`.

Please note that this check is pretty cheap, it takes less than a second on my machine:

```
$ time bin/check-package-locks.js

real	0m0.729s
user	0m0.337s
sys	0m0.191s
```

What are your concerns? Why do you think this should go to a different script? I suppose I can move it to `posttest` if you prefer.

bajtos:
@b-admike PTAL again

sandbox/example/package-lock.json
bajtos: Disabling package-lock would work until we add an external dependency, at which point we will almost certainly forget to enable it.

raymondfeng:
FYI: I was referring to https://github.com/strongloop/loopback-next/pull/2650 (GK adding `@loopback/*` deps in package-lock.json`) but https://github.com/strongloop/loopback-next/pull/2662 seems to be fine. We need to double-check.

raymondfeng:
This is an example PR from greenkeeper that adds `@loopback/*` to package-lock.json files:

https://github.com/strongloop/loopback-next/pull/2664

bajtos:
I am going to land this pull request. IMO, if we want to keep using package-locks, then we need an automated way for excluding monorepo-local dependenices from lock files. If this proves to be a problem, then I think we need to consider disabling package-lock files until there is better support in related tooling.

package.json
b-admike: I was just wondering why we put it there, sounds good. Thank you for providing the tests too!

Issue #2672, opened 2019-04-02
Replace "request" with something else


bajtos:
`request` module is moving to maintenance mode, we should eventually find a new HTTP client. 

- The announcement: https://github.com/request/request/issues/3142
- List of alternatives: https://github.com/request/request/issues/3143
- Comparison: https://www.npmjs.com/package/got#comparison

I tried to upgrade to `got`. `got` works well for simple use cases, but stops working for non-trivial requests (e.g. when a proxy is involved). What's worse, it replaces errors reported by Node.js core with its own error classes, which makes troubleshooting extremely difficult :disappointed: It seems that other popular clients like `node-fetch` and `axios` are doing the same mistake :disappointed: 

(UPDATED on Jun 12, 2019)
### Acceptance Criteria

- [ ] Review `request` alternatives (e.g. `axios` and `node-fetch`), find one that meets the following requirements:
  - It should preserve low-level stack traces reported by Node.js instead of thrown fresh error instances
  - It need to support proxies, ideally `@loopback/http-caching-proxy` too.
   - If it's browser-friendly, then we should choose something that's more compatible with `node-fetch` API.

It can be a reusable component/module where we can customize the behavior.

seiya-git:
for proxy u can use https://www.npmjs.com/package/proxy-agent with got

hacksparrow:
What about https://www.npmjs.com/package/axios?

bajtos:
> for proxy u can use https://www.npmjs.com/package/proxy-agent with got

Good to know! I was trying https://github.com/koichik/node-tunnel and that did not work.

In general, I am very reluctant to use a library that's discarding low-level stack traces reported by Node.js and replacing them with a new stack pointing only to the client library. Unfortunately, that rules out both `got` and `axios`.

dhmlau:
@bajtos @hacksparrow, do you think we should turn this into a spike to find out what's a good substitution? 

I'm thinking of the following acceptance criteria. Please review:

### Acceptance Criteria
- [ ] Review the list of alternatives listed in https://github.com/request/request/issues/3143, and determine which one is the best 
   - Question: what are the criteria to determine the most suited one?

raymondfeng:
I don't have strong opinions here but axios and node-fetch seem to be two popular candidates.

dhmlau:
Thanks @raymondfeng. How do we know which one is better? :)  Any criteria to determine that?

bajtos:
[...]

PR #2673, opened 2019-04-02
refactor(http-server): remove "request" module from tests

bajtos:
See #2672.

This pull request is removing `request` from dev-dependencies of `@loopback/http-server`. The only test using that module is reworked to use `httpGetAsync` from Node.js core instead.

## Checklist

### DIFF
--- a/packages/http-server/src/__tests__/integration/http-server.integration.ts
+++ b/packages/http-server/src/__tests__/integration/http-server.integration.ts
@@ -3,19 +3,18 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {HttpServer, HttpOptions, HttpServerOptions} from '../../';
 import {
-  supertest,
   expect,
-  itSkippedOnTravis,
+  givenHttpServerConfig,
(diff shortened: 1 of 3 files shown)


packages/http-server/src/__tests__/integration/http-server.integration.ts
```diff
@@ -46,11 +45,7 @@ describe('HttpServer (integration)', () => {
     server = new HttpServer(dummyRequestHandler, serverOptions);
     await server.start();
     await server.stop();
-    await expect(
-      makeRequest({
-        uri: server.url,
-      }),
-    ).to.be.rejectedWith(/ECONNREFUSED/);
+    await expect(httpGetAsync(server.url)).to.be.rejectedWith(/ECONNREFUSED/);
```
raymondfeng: Does `supertest` hide `ECONNREFUSED` too?

bajtos: IDK. `httpGetAsync` is using Node.js core API, it's the simplest possible solution. That's why I prefer it over `supertest` or `request`.

Issue #133, opened 2017-04-13
Interceptors


deepakrkris:
- [ ] Must be able to inject request/response interceptors for a Controller
 ```
      @intercept('request',  function(request, next) {
         logMessage(request, next);
      });
      @intercept('response', './scripts/error-handler');
      @intercept('request', './scripts/user-context');
      @api(def) 
      class OrderCancellationController {
      .....
      }
``` 
- [ ] interceptors will be invoked in the order of declaration for the controller
- [ ] interceptor scripts must export a function with a specific signature
- [ ] An interceptor can modify the request/response, headers and body

dhmlau:
need a way to register the decorators for interceptors

bajtos:
I think this is conceptually similar to remoting hooks (beforeRemote, afterRemote, afterRemoteError) in LoopBack 3.x

raymondfeng:
See https://docs.nestjs.com/interceptors

dhmlau:
Closing as done. 
See https://loopback.io/doc/en/lb4/Interceptors.html

PR #2687, opened 2019-04-03
feat(context): add support for method interceptors

raymondfeng:
See https://github.com/strongloop/loopback-next/blob/interceptor/docs/site/Interceptors.md

Implements #133

## Checklist


raymondfeng:
> How would the interceptor affect the sequence? will they be executed when invoking the controller function?

They will be invoked by the `InvokeMethod` action. The hook is on the `invokeMethod()`.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Example interceptors
+
+Here are some example interceptor functions:
+
+```ts
+// A sync interceptor to log method calls
+const logSync: Interceptor = (invocationCtx, next) => {
+  console.log('logSync: before-' + invocationCtx.methodName);
+  // We don't wait for `next()` to return even when if the method or an
```
b-admike: nitpick: `even when if` -> `even when` or `even if`

docs/site/Interceptors.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+Here is the list of interceptors invoked for each method:
+
+- greetStatic: `log`
+- greetStaticWithDI: `log`
+- greetSync: `log`, `logSync`
+- greet: `convertName`, `log`
+
+## Invoke a method with interceptors
```
b-admike: In this section, how are we specifying the `interceptor` methods to use with `invokeMethod`?

raymondfeng: Controller methods decorated with `@intercept` are invoked with `invokeMethod` behind the scene.

raymondfeng: I added a sentence in the docs to make it clear.

raymondfeng:
> I like the idea of interceptors 👍, and I can see how they can be used, but was wondering how we want to use them with LB4. Are they supposed to represent before/after hooks in LB3 for CRUD methods? Are there some use cases we can show users in the context of LB4?

The interceptors can be used as `around` middleware at per method level. Typical usages:

1. Implement caching
2. Add logging/monitoring/..
3. Implement authorization
4. Enforce parameter validation
5. Transform input/output

They are close to remote hooks.

It's also possible for method decorators to return a proxy method. But that won't allow context access for such interceptor functions.

bajtos:
I am not sure if I'll manage to review this pull request by the end of this week, please give me more time to catch up. I feel this is pull request is adding a fundamental building block that will be difficult to change later, therefore a careful review is needed.

raymondfeng:
FYI: I also added an acceptance test for `@loopback/rest` to illustrate a mock-up caching interceptor.

raymondfeng:
Implements https://github.com/strongloop/loopback-next/issues/133

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+interceptor:
+
+```ts
+class NameValidator implements Provider<Interceptor> {
+  constructor(@inject('valid-names') private validNames: string[]) {}
+  value() {
+    return async <T>(
+      invocationCtx: InvocationContext,
+      next: () => ValueOrPromise<T>,
+    ) => {
```
bajtos: Anonymous functions produce stack traces that makes troubleshooting difficult. For Sequence action providers, we are advocating to move the function into a named method. I'd like us to do the same for interceptors.

```ts
class NameValidator implements Provider<Interceptor> {
  constructor(@inject('valid-names') private validNames: string[]) {}

  value() {
    return (ctx, next) => this.intercept(ctx, next);
  }

  async intercept<T>(invocationCtx: InvocationContext, next: () => ValueOrPromise<T>) {
    const name = invocationCtx.args[0];
    // etc.
  }
}
```

Now the stack trace will contain `NameValidator.intercept` entry.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+
+  @intercept(changeName)
+  async helloWithChangeName(name: string) {
+    return `Hello, ${name}`;
+  }
+
+  @intercept(logWithErrorHandling)
+  async helloWithError(name: string) {
+    throw new Error('error: ' + name);
+  }
```
bajtos: I find this example too long. There are about 10 methods in an unstructured list, many of them are slight variations on the same theme. Could you please pick 3-5 most important cases and convert them into the form of a combination of short explanatory text with a code snippet?

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+}
+```
+
+### Class level interceptors
+
+To apply interceptors to be invoked for all methods on a class, we can use
+`@intercept` to decorate the class. When a method is invoked, class level
+interceptors (if not explicitly listed at method level) are invoked before
+method level ones.
```
bajtos: Please clarify whether the interceptors will be applied to protected and private methods too.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+    };
+  }
+}
+```
+
+## Apply interceptors
+
+Interceptors form a cascading chain of handlers around the target method
+invocation. We can apply interceptors by decorating methods/classes with
+`@intercept`.
```
bajtos: I find this misleading. `@intercept` is  not replacing the actual method body, but only creating a new metadata entry. IIUC the source code, interceptors are applied only when the method is invoked via `invokeMethodWithInterceptors`.

Please make this fact very clear in the docs.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+### Global interceptors
+
+Global interceptors are discovered from the `InvocationContext`. They are
+registered as bindings with `interceptor` tag. For example,
+
+```ts
+app
+  .bind('interceptors.MetricsInterceptor')
+  .toProvider(MetricsInterceptorProvider)
+  .tag('interceptor');
```
bajtos: Have you considered creating a tag template `asInterceptor`?

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+Here is the list of interceptors invoked for each method on
+`MyControllerWithClassLevelInterceptors`.
+
+- greetStatic: `log` (from the class)
+- greetStaticWithDI: `log` (from the method, class level `log` is removed)
+- greetSync: `log`, `logSync` (from the method, method level `log` is preserved
+  and class level `log` is removed)
+- greet: `convertName`, `log` (from the method, method level `log` is preserved
+  and class level `log` is removed. This allows a method to explicitly control
+  the order.)
```
bajtos: I find this text difficult to comprehend, it's referring to small bits of a large code example from elsewhere. 

I am proposing to expand each item in this list to show the relevant code example (the class & method definition including any decorators).

It's also not clear to me why are you mentioning both `greet` and `greetSync` when the comment in the brackets is similar: "from the method, method level `log` is preserved and class level `log` is removed."

Please make this part more clear.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Invoke a method with interceptors
+
+Controller methods decorated with `@intercept` are invoked with applied
+interceptors for corresponding routes upon API requests.
```
bajtos: It's great to have a detailed documentation for interceptors :+1:

I feel the current structure is focused too much on low-level details and how things are implemented, rather on what a typical framework user needs to understand.

I am proposing to rework the page "Key concepts >> Interceptors" as follows:

1. Overview - keep it as it is
2. When are interceptors invoked - for controller methods 
3. How to apply interceptors in my code - use `@intercept` decorator
4. How to implement my own interceptor
5. Advanced - move the remaining information here

@strongloop/loopback-maintainers @bschrammIBM thoughts?

docs/site/Interceptors.md
```diff
@@ -0,0 +1,328 @@
...
+export interface Interceptor {
+  <T>(
+    context: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ): ValueOrPromise<T>;
+}
+```
+
+An interceptor is responsible for calling `next()` if it wants to proceed with
+next interceptor or the target method invocation.
```
bajtos: I am afraid this sentence is not enough to explain `next`. I was confused about this concept until I read example interceptors below, at which point I started to understand that you are using `next` the same way as Koa does.

Please add more details here to better explain the concept of `next`, how can interceptors use it to run code after the target methods has finished, etc.

bajtos:
> coverage/coveralls — Coverage decreased (-0.03%) to 91.018%

@raymondfeng please review the code coverage report, there may be valid use cases/scenarios that are not covered by your tests.

raymondfeng:
> (1)
I have one major concern though: you are proposing to apply interceptors on controller methods only. However, LB4 supports other flavors for endpoint implementations, most notable route handler functions:

It's now supported.

docs/site/Interceptors.md
raymondfeng: I did some refactoring and add more comments.

bajtos:
Thank you @raymondfeng for the updates. I'm quite busy with other work now, I'll try to find time to review this pull request in details by Friday this week.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,552 @@
...
+`caching-interceptor` for all methods.
+
+```ts
+app
+  .bind('caching-interceptor')
+  .toProvider(CachingInterceptorProvider)
+  .apply(asInterceptor);
+```
+
+Global interceptors are also executed for handler functions without a controller
```
bajtos:
```suggestion
Global interceptors are also executed for route handler functions without a controller
```

docs/site/Interceptors.md
```diff
@@ -0,0 +1,552 @@
...
+
+/**
+ * A plain function to convert `text` to lower case
+ * @param text
+ */
+function toLowerCase(text: string) {
+  return text.toLowerCase();
+}
+
+app.route('get', '/toLowerCase/{text}', toLowerCaseOperationSpec, toLowerCase);
```
bajtos: Can we use a simpler example please? IIUC, you want to demonstrate that interceptors are invoked for handler functions too. This code contains lot of low-level details related to route definition but unrelated to interceptors. 

I think it's best to not duplicate content we have elsewhere and simply point readers to https://loopback.io/doc/en/lb4/Routes.html#using-partial-openapi-spec-fragments where handler functions are explained in more details.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,552 @@
...
+[proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
+can be created from the target class or object to apply interceptors. This is
+useful for the case that a controller declares dependencies of repositories or
+services and would like to allow repository or service methods to be
+intercepted.
+
+```ts
+import {createProxyWithInterceptors} from '@loopback/context';
+
+const proxy = createProxyWithInterceptors(controllerWithClassInterceptors, ctx);
```
bajtos: Is this accepting a class constructor or a class instance? I assume the latter, in which case I am proposing to rename the variable as follows:

```suggestion
const proxy = createProxyWithInterceptors(controllerInstanceWithClassInterceptors, ctx);
```

docs/site/Interceptors.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+There is also an `asProxyWithInterceptors` option for binding resolution or
+dependency injection to return a proxy to apply interceptors as methods are
+invoked.
+
+```ts
+class DummyController {
+  constructor(
+    @inject('my-controller', {asProxyWithInterceptors: true})
```
bajtos: The addition of proxy support is making this pull request unnecessarily large. Can you please remove proxies from the first iteration, to make it easier for us to review the changes? At the moment, this pull request has 1800 new lines of code :(

raymondfeng: The addition is in its own commit. Would that help?

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  context('toUpperCase with bound caching interceptor', () => {
+    it('invokes the controller method if not cached', async () => {
+      await client.get('/toUpperCase/Hello').expect(200, 'HELLO');
+      expect(status.returnFromCache).to.be.false();
+    });
+
+    it('returns from cache without invoking the controller method', async () => {
+      for (let i = 0; i <= 5; i++) {
+        await client.get('/toUpperCase/Hello').expect(200, 'HELLO');
```
bajtos: What happens when I run this test in isolation and  don't execute `it('invokes the controller method if not cached'` first? Will it still pass?

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+        expect(status.returnFromCache).to.be.true();
+      }
+    });
+
+    it('invokes the controller method after cache is cleared', async () => {
+      clearCache();
```
bajtos: What happens when I run this test in isolation and don't execute any of the tests above? Will it ever fail when the implementation is not correct?

Tests depending on side effects of other tests are fragile and difficult to maintain. Please rework this test suite so that each test can be executed on its own.

1. Move `clearCache` to beforeEach hook
2. If the test is assuming a particular state of the cache, then this state should be (re)created in the test

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  context('toLowerCase with cache interceptor function', () => {
+    it('invokes the controller method if not cached', async () => {
+      await client.get('/toLowerCase/Hello').expect(200, 'hello');
+      expect(status.returnFromCache).to.be.false();
+    });
+
+    it('returns from cache without invoking the controller method', async () => {
+      for (let i = 0; i <= 5; i++) {
+        await client.get('/toLowerCase/Hello').expect(200, 'hello');
```
bajtos: Same comment applies here, rework the test to be independent of other tests.

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+        expect(status.returnFromCache).to.be.true();
+      }
+    });
+
+    it('invokes the controller method after cache is cleared', async () => {
+      cachedResults.clear();
+      await client.get('/toLowerCase/Hello').expect(200, 'hello');
```
bajtos: Same comment applies here, rework the test to be independent of other tests.

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+    toUpperCase(@param.path.string('text') text: string) {
+      return text.toUpperCase();
+    }
+
+    @intercept(cache)
+    @get('/toLowerCase/{text}')
+    toLowerCase(@param.path.string('text') text: string) {
+      return text.toLowerCase();
+    }
+  }
```
bajtos: When we will be adding more `context` blocks in the future, this class will grow into size and become difficult to read and understand.

Please move the definition of controller methods to individual `context` blocks.

```ts
context('toUpperCase with bound caching interceptor', () => {
  before(() => {
    class StringCaseController {
     @intercept('caching-interceptor')
      @get('/toUpperCase/{text}')
      toUpperCase(@param.path.string('text') text: string) {
        return text.toUpperCase();
      }

    app.controller(StringCaseController);
  });

   // the tests
});
```

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export const status = {
+  returnFromCache: false,
+};
+
+/**
+ * An interceptor function that caches results. It uses `invocationContext`
+ * to locate the http request
+ */
+export const cachedResults = new Map<string, unknown>();
```
bajtos: The comment says "An interceptor function" but this is a key/value map. Please fix.

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export async function cache<T>(
+  invocationCtx: InvocationContext,
+  next: () => ValueOrPromise<T>,
+) {
+  status.returnFromCache = false;
+  const req = await invocationCtx.get(RestBindings.Http.REQUEST, {
+    optional: true,
+  });
+  if (!req || req.method.toLowerCase() !== 'get') {
```
bajtos: Why is `cache` interceptor caching GET requests only, but `CachingInterceptorProvider` caching all requests?

Please simplify the implementation of these two alternatives by extracting the common code into a shared function.

packages/rest/src/__tests__/acceptance/caching-interceptor/global-caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,122 @@
...
+describe('global caching interceptor', () => {
+  let client: Client;
+  let app: RestApplication;
+
+  before(givenAClient);
+  after(async () => {
+    await app.stop();
+  });
+
+  context('toUpperCase', () => {
```
bajtos: This is confusing. IIUC, this block is verifying that global interceptors are applied on controller methods.

Please rework `context` name to capture this information and move the definition of the controller from the bottom of the test file into this particular `context` block that's using it.

packages/rest/src/__tests__/acceptance/caching-interceptor/global-caching-interceptor.acceptance.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    it('invokes the controller method after cache is cleared', async () => {
+      clearCache();
+      await client.get('/toUpperCase/Hello').expect(200, 'HELLO');
+      expect(status.returnFromCache).to.be.false();
+    });
+  });
+
+  context('toLowerCase', () => {
```
bajtos: This is confusing. IIUC, this block is verifying that global interceptors are applied on route handlers.

Please rework `context` name to capture this information and move the definition of the route from the bottom of the test file into this particular `context` block that's using it.

docs/site/Interceptors.md
bajtos: Sorry, it's still not clear to me whether the function is accepting a class constructor or a class instance. Please clarify.

docs/site/Interceptors.md
bajtos: Not really, but I'll do my best.

packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
```diff
@@ -239,6 +240,75 @@ describe('Interceptor', () => {
     });
   });
 
+  context('proxy with interceptors', () => {
+    it('invokes async interceptors on an async method', async () => {
+      const proxy = createProxyWithInterceptors(
+        controllerWithClassInterceptors,
+        ctx,
+      );
+      const msg = await proxy.greet('John');
```
bajtos: Please move the controller class & method definition to individual tests, as I explained in one of my earlier comments, so that reader of this code can understand what scenario is being tested, without having to jump to the definition of `controllerWithClassInterceptors`. The same comment applies to other tests added by this commit too.

packages/context/src/binding.ts
```diff
@@ -503,9 +504,18 @@ export class Binding<T = BoundValue> {
       debug('Bind %s to class %s', this.key, ctor.name);
     }
     this._type = BindingType.CLASS;
-    this._setValueGetter((ctx, options) =>
-      instantiateClass(ctor, ctx, options.session),
-    );
+    this._setValueGetter((ctx, options) => {
+      const instOrPromise = instantiateClass(ctor, ctx, options.session);
+      if (!options.asProxyWithInterceptors) return instOrPromise;
+      return transformValueOrPromise(instOrPromise, inst => {
```
bajtos: Please extract this part into a named function and keep `this._setValueGetter` callback size down to 3-5 lines of code.

packages/context/src/binding.ts
```diff
@@ -503,9 +504,18 @@ export class Binding<T = BoundValue> {
...
+        if (typeof inst !== 'object') return inst;
+        return (createProxyWithInterceptors(
+          // Cast inst from `T` to `object`
+          (inst as unknown) as object,
```
bajtos: Why is TypeScript not able to infer that `inst` must be an `object`, considering the `typeof inst !== 'object'` typecheck above? Can you try to rewrite this code in such way that no explicit casts are needed? 

Would this work?

```ts
if (typeof inst !== 'object') return inst;
const obj: object = inst;
const proxy = createProxyWithInterceptors(obj, ctx);
return (proxy as unknown) as T;
```

Also I see that `creatzeProxyWithInterceptors` is already parameterized, why cannot we leverage that generic parameter here? If we cannot leverage it, then what's the point of having that parameter in the first place?

```ts
return createProxyWithInterCeptors<T>(obj, ctx);
```

Is the problem perhaps in the fact that our current type descriptions allow `T` to be any value, not only objects? IIUC, it's possible to create `new Binding<number>().toClass(/*...*/)`, which does not look right to me. Perhaps we need to restrict the types for which `toClass` is available, something along the lines of `T extends object`. Can we change `Constructor<T>` alias for that? I am not sure if that would be considered as a breaking change or not.

packages/context/src/inject.ts
```diff
@@ -413,6 +409,7 @@ function resolveAsGetter(
     return ctx.get(bindingSelector, {
       session: forkedSession,
       optional: injection.metadata.optional,
+      asProxyWithInterceptors: injection.metadata.asProxyWithInterceptors,
```
bajtos: Now that `InjectionMetadata` are extending `ResolutionOptions`, I'd prefer to avoid the need to copy subset of metadata properties and having to update this block whenever we add a new resolution option. I am proposing to  pass all properties from `injection.metadata` instead:

```ts
return ctx.get(bindingSelector, {
  ...(injection.metadata),
  session: forkedSession
});
```

packages/context/src/interceptor.ts
```diff
@@ -315,3 +315,36 @@ function invokeInterceptors(
   };
   return next();
 }
+
+/**
+ * A proxy handler that applies interceptors
+ *
+ * See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Proxy
+ */
+export class InterceptorHandler<T extends object> implements ProxyHandler<T> {
```
bajtos: Please improve the name to describe the intent (apply/invoke interceptors), not implementation details (interceptor-related handler). E.g. `InterceptorInvoker` instead of `InterceptorHandler`.

packages/context/src/interceptor.ts
```diff
@@ -315,3 +315,36 @@ function invokeInterceptors(
...
+  constructor(private context = new Context()) {}
```
bajtos: Under which circumstances would it make sense to use this handler with an empty context? Can we remove the default argument value?

packages/context/src/interceptor.ts
```diff
@@ -315,3 +315,36 @@ function invokeInterceptors(
...
+  get(target: T, p: PropertyKey, receiver: any): any {
```
bajtos: Please use a more descriptive name than `p`, e.g. `prop` or `propKey`.

packages/context/src/interceptor.ts
```diff
@@ -315,3 +315,36 @@ function invokeInterceptors(
...
+/**
+ * Create a proxy that applies interceptors for method invocations
+ * @param target Target class or object
+ * @param context Context object
+ */
+export function createProxyWithInterceptors<T extends object>(
+  target: T,
+  context?: Context,
+) {
+  return new Proxy(target, new InterceptorHandler(context));
```
bajtos: I think you should pass `T` argument to the constructor? Unless TypeScript is clever enough to infer that specialization automatically.

```ts
return new Proxy(target, new InterceptorHandler<T>(context));
```

packages/context/src/resolution-session.ts
```diff
@@ -338,6 +338,13 @@ export interface ResolutionOptions {
    * will return `undefined` instead of throwing an error.
    */
   optional?: boolean;
+
+  /**
+   * A boolean flag to control if a proxy should be created to apply
+   * interceptors for the resolved value. It's only honored for bindings backed
+   * by a class.
```
bajtos: We should let the users know when this flag was ignored because the binding was not backed by a class.

1. You can print a `debug` statement, this way we don't pollute console with excessive output. 
2. Alternatively, I am also fine with throwing an error to let the programmer know about the error they made in their code.

I slightly prefer the second option (throw an error).

packages/context/src/resolver.ts
```diff
@@ -154,6 +154,7 @@ function resolve<T>(
           // If the `optional` flag is set for the injection, the resolution
           // will return `undefined` instead of throwing an error
           optional: injection.metadata.optional,
+          asProxyWithInterceptors: injection.metadata.asProxyWithInterceptors,
```
bajtos: Similar comment here: now that metadata are inheriting from resolution options, can we simply copy all metadata?

```ts
return ctx.getValueOrPromise(key, {
  ...(injection.metadata),
  session: s,
});
```

Please extract all parts related to this inheritance change into a standalone pull request, so that we can land them faster and more importantly to get a proper entry in the change-log.

docs/site/Interceptors.md
bajtos: I am afraid the changes are not enough. We are writing documentation here, the content should read as well-structured text. It's ok to use code snippets for illustrations, but these snippets should not serve as a replacement for the actual human-readable text.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+);
+```
+
+Or:
+
+```ts
+const proxy = await ctx.get<MyControllerWithClassLevelInterceptors>(
+  'my-controller',
+  {asProxyWithInterceptors: true},
+const msg = await proxy.greet('John');
```
bajtos: Syntax error. Should be:

```diff
 const proxy = await ctx.get<MyControllerWithClassLevelInterceptors>(
   'my-controller',
   {asProxyWithInterceptors: true},
+ );
 const msg = await proxy.greet('John');
```

I spotted the error because the markdown source rendered by GitHub in the PR diff is using incorrect colors from the next line onwards: the text is in blue instead of in black. 

When you fix this error then please check the rest of this file to ensure there are no other typos in the code snippets.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+
+Interceptors form a cascading chain of handlers around the target method
+invocation. We can apply interceptors by decorating methods/classes with
+`@intercept`. Please note `@intercept` does **NOT** return a new method wrapping
+the target one. Instead, it adds some metadata instead and such information is
+used by `invokeMethod` or `invokeWithMethodWithInterceptors` functions to
+trigger interceptors around the target method. The original method stays intact.
+Invoking it directly won't apply any interceptors.
+
+### @intercept
```
bajtos:
```suggestion
### `@intercept`
```

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+
+A **public** static or prototype method on a class can be decorated with
+`@intercept` to attach interceptors to the target method. Please note
+interceptors don't apply to protected or private methods.
+
+#### Static methods
+
+```ts
+class MyControllerWithStaticMethods {
+  // Apply `log` to a static method
```
bajtos: Please rework this example from a single code snippet with code comments into proper text with multiple code snippets.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+  static async greetStatic(name: string) {
+    return `Hello, ${name}`;
+  }
+
+  @intercept(log)
+  static async greetStaticWithDI(@inject('name') name: string) {
+    return `Hello, ${name}`;
+  }
+
+  // We can apply `@intercept` multiple times on the same method
```
bajtos: Please rework this example from a single code snippet with code comments into proper text with multiple code snippets.

docs/site/Interceptors.md
bajtos: > Please clarify whether the interceptors will be applied to protected and private methods too.

I don't see where is this question answered, could you please point me to the right place?

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+class MyControllerWithClassLevelInterceptors {
+  // ...
+}
+```
+
+1. greetStatic: [`log`]
+
+   ```ts
+   // No explicit `@intercept` at method level. The class level `log` will
+   // be applied
```
bajtos: Please move the content from code comment into regular documentation text.

Same comment applies to other items below too.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+
+```ts
+// Apply `log` to all methods on the class
+@intercept(log)
+class MyControllerWithClassLevelInterceptors {
+  // ...
+}
+```
+
+1. greetStatic: [`log`]
```
bajtos: What does `greetStatic` mean in the context of "Order of invocation for interceptors" section? It looks like a random JS/TS identifier to me.

Please rework this title to capture the essence of the scenario you are going to show. For example: "A static method with no method-level interceptors".

Same comment applies to other items below too.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+
+The interceptor function has the following signature:
+
+```ts
+export interface Interceptor {
+  <T = unknown>(
+    context: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ): ValueOrPromise<T>;
+}
```
bajtos: We will need to update the signature of this function based on the comments I posted in the TypeScript source code.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+method be executed. It's valid to skip `next()` if by intention, for example, an
+interceptor can fail the invocation early due to validation errors or return a
+response from cache without invoking the target method.
+
+### Example interceptors
+
+Here are some example interceptor functions:
+
+```ts
+// A sync interceptor to log method calls
```
bajtos: Please split this long code snippet into multiple smaller ones. Rework the comment introducing each interceptor implementation into a title. For example:

```markdown
### Example interceptors
	
Here are some example interceptor functions:

#### A sync interceptor to log method calls

``ts
const logSync: Interceptor = (invocationCtx, next) => {
  // ...
};
``

#### An async interceptor to log method calls

etc.
```

docs/site/Interceptors.md
```diff
@@ -0,0 +1,529 @@
...
+[Route Handler](Routes.md#using-partial-openapi-spec-fragments).
+
+### Create a proxy to apply interceptors
+
+A
+[proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
+can be created from the target class or object to apply interceptors. This is
+useful for the case that a controller declares dependencies of repositories or
+services and would like to allow repository or service methods to be
+intercepted.
```
bajtos: While I see how compelling this idea is, I find it also problematic. Consider the situation when a sync function is decorated with an async interceptor. The proxy is changing the API of the decorated class, turning synchronous methods into asynchronous.

bajtos:
@raymondfeng I posted 40+ review comments above, GitHub is collapsing them. Please make sure to expand the collapsed placeholder to see all comments.

docs/site/Interceptors.md
raymondfeng: See https://github.com/strongloop/loopback-next/pull/2687/commits/a419b4fe47d3db95112fae84d596652b87450711

docs/site/Interceptors.md
raymondfeng: See https://github.com/strongloop/loopback-next/blob/interceptor/docs/site/Interceptors.md#method-level-interceptors

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
raymondfeng: I want to start the app once and `app.controller` won't work after the application is started.

packages/context/src/resolver.ts
raymondfeng: Please note we use `es2017` but the object spread syntax is new in `es2018`.

packages/context/src/interceptor.ts
```diff
@@ -0,0 +1,317 @@
...
+}
+
+/**
+ * Interceptor function
+ */
+export interface Interceptor {
+  <T = unknown>(
+    context: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ): ValueOrPromise<T>;
```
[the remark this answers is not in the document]
raymondfeng: Please note `invokeMethod` returns `ValueOrPromise`. If we want to keep backward compatibility, `invokeMethodWithInterceptors` should also return `ValueOrPromise`.

packages/context/src/binding.ts
raymondfeng: TypeScript cannot infer `inst` as `object` after `if (typeof inst !== 'object') return inst;`.

packages/context/src/interceptor.ts
raymondfeng: If the method does not have `interceptors` from binding keys, `context` is not required.

packages/context/src/interceptor.ts
```diff
@@ -0,0 +1,317 @@
...
+  return binding.tag(ContextTags.INTERCEPTOR);
+}
+
+/**
+ * Interceptor function
+ */
+export interface Interceptor {
+  <T = unknown>(
+    context: InvocationContext,
+    next: () => ValueOrPromise<T>,
```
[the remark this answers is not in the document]
raymondfeng: `T` removed.

packages/context/src/resolver.ts
raymondfeng: I tried with Node 8 and 10 and both work. Updated.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,583 @@
...
+Interceptors are supported for public controller methods (including both static
+and prototype) and handler functions for REST routes.
+
+Controller methods decorated with `@intercept` are invoked with applied
+interceptors for corresponding routes upon API requests.
+
+```ts
+import {intercept} from '@loopback/context';
+
+@intercept(log)
```
hacksparrow: We should provide context to where `log` comes from.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,583 @@
...
+controller class. See an example in
+[Route Handler](Routes.md#using-partial-openapi-spec-fragments).
+
+### Create a proxy to apply interceptors
+
+A
+[proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
+can be created from the target class or object to apply interceptors. This is
+useful for the case that a controller declares dependencies of repositories or
+services and would like to allow repository or service methods to be
```
hacksparrow: @raymondfeng Is the interception limited to the instance in the controller or is it a global effect?

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
bajtos: > I want to start the app once

What is the reasoning behind the requirement? Can we find a different solution how to move the definition of controllers into individual tests/context blocks?

bajtos:
@raymondfeng before I review the changes in details, I'd like to reach a shared high-level understanding of how we want to deal with asynchronicity.

At high level:
1. The target function (intercepted controller method) can be sync or async.
2. The interceptor can be sync or async.

This gives us four combinations:
 1. sync target, sync interceptor
 2. sync target, async interceptor
 3. async target, sync interceptor (caveat: even if the interceptor is sync in principle, the outcome is async because `next` returns a promise)
 4. async target, async interceptor

In theory, the first combination (sync target, sync interceptor) can preserve sync behavior of the invoked operation. The remaining three combinations will always cause the operation to be async. (Assuming interceptors wait for the outcome of the intercepted operations.)

As I understand your current proposal, you want to support sync behavior, which is reflected in the Interceptor API:

- The interceptor returns `ValueOrPromise` (i.e. can return synchronously, but does not have to!)
- The `next` callback returns `ValueOrPromise` (i.e. can return synchronously, but does not have to!)

IMO, this is complicating implementation a lot.
1. The code calling an interceptor (or a chain of interceptors for a method) cannot tell in advance whether the outcome will be sync or async and thus must be prepared to handle both cases.
2. Individual interceptor implementations cannot make any assumption on whether `next` is going to return the result (sync) or a promise (async), they must be prepared to handle both cases.

The first point is of lesser importance because interceptors are usually invoked by the framework, thus there are only few places where we need to be careful.

I am mostly concerned about the second point, about interceptors that will be written by regular LB users. IMO, most of these developers won't understand delicate details of sync vs. async flavors of interceptors and target methods. In the better case, they will write all interceptors as `async` functions and thus prevent the framework from leveraging benefits of sync interception. Worse, they will try to write their interceptor to support both sync & async workflows and possibly end up with complex code that does not work as expected.

In that light, I'd like us to reconsider whether the combination of sync target + sync interceptors is so important to justify burden of extra complexity placed on framework users.

Personally, I prefer the simplicity of fully-async approach:

```ts
export interface Interceptor {
  (
    context: InvocationContext,
    next: () => Promise<InvocationResult>,
  ): Promise <InvocationResult>;
}
```

@strongloop/loopback-maintainers thoughts?

docs/site/Interceptors.md
bajtos: > See a419b4f

@raymondfeng the commit does not seem to be a part of this pull request anymore, perhaps after some rebase & force-push changes? I left few comments for you in that commit, please take a look.

raymondfeng:
> Personally, I prefer the simplicity of fully-async approach

I'm not against the idea. Here is the caveat for api compatibility:

1. Currently `invokeMethod` allows both sync/async
2. The PR improved `invokeMethod` to always apply interceptors if present. As a result, `invokeMethod` does not allow sync any more.

docs/site/Interceptors.md
raymondfeng: This is the new commit to address your concerns - https://github.com/strongloop/loopback-next/pull/2687/commits/7b15940b66372c6cf042cd3a8bcf7a80a4d1caa9

bajtos:
> > Personally, I prefer the simplicity of fully-async approach
> > ```ts
> > export interface Interceptor {
> >  (
> >    context: InvocationContext,
> >     next: () => Promise<InvocationResult>,
> >  ): Promise <InvocationResult>;
> > }
> > ```
> 
> I'm not against the idea. Here is the caveat for api compatibility:
>
> Currently invokeMethod allows both sync/async
> The PR improved invokeMethod to always apply interceptors if present. As a result, invokeMethod does not allow sync any more.

I was thinking about this problem during the weekend.

My proposal has an important downside: it's not forward-compatible. If we ever decide to add support for sync methods & sync interceptors in the future, then such change could break existing code. For example:

```ts
function myInterceptor(context, next) {
  // a dummy logger using Promise API instead of async/await
  return next.then(result => {
    console.log('result', result);
    return result;
  });
}

const proxy = // create proxy with interceptors
proxy.someMethod().then(result => {
  // a dummy example invoking an intercepted method
  // while using Promise API instead of async/await
});
```

In that light, I think we should preserve the current implementation that supports both sync and async flavors. BUT! I still would like us to make interceptors easier to understand for new LB developers, I think this can be easily achieved by reworking our documentation to start with simplified information showing fully-async flavor only, and then explaining sync versions later in "advanced" section.

packages/context/src/interception-proxy.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export type AsPromise<T> = T extends Promise<unknown> ? T : Promise<T>;
+
+/**
+ * The async variant of a function to always return Promise<R>. If T is not a
+ * function, the type is `T`.
+ */
+// tslint:disable-next-line:no-unused (possible tslint bug to treat `R` as unused)
+export type AsAsyncFunction<T> = T extends (...args: InvocationArgs) => infer R
+  ? (...args: InvocationArgs) => AsPromise<R>
```
bajtos: Per my earlier comments, it may be better to use `ValueOrPromise` here, to preserve sync behavior when both the intercepted method and all interceptors are synchronous.

IMO, modifying this API from `Promise` to `ValueOrPromise` would be breaking change, therefore it's better to get it right on the first try.

packages/context/src/resolution-session.ts
bajtos: > We should let the users know when this flag was ignored because the binding was not backed by a class.
>
> 1. You can print a `debug` statement, this way we don't pollute console with excessive output. 
> 2. Alternatively, I am also fine with throwing an error to let the programmer know about the error they made in their code.
>
> I slightly prefer the second option (throw an error).

☝️

packages/rest/src/__tests__/acceptance/caching-interceptor/caching-interceptor.acceptance.ts
bajtos: > > I want to start the app once
>
> What is the reasoning behind the requirement? Can we find a different solution how to move the definition of controllers into individual tests/context blocks?

☝️

docs/site/Interceptors.md
```diff
@@ -0,0 +1,586 @@
...
+
+```ts
+{
+  name: string; // the same as MyController
+  greet(name: string): Promise<string>; // the return type becomes `Promise<string>`
+  hello(name: string): Promise<string>; // the same as MyController
+}
+```
+
+### Use `invokeMethod` to apply interceptors
```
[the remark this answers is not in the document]
raymondfeng: Now `invokeMethod` internally uses `invokeMethodWithInterceptors` to allow both method parameter dependency injection and interceptors.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,586 @@
...
+
+```ts
+/**
+ * InvocationContext for method invocations
+ */
+export class InvocationContext extends Context {
+  /**
+   * Construct a new instance
+   * @param parent Parent context
+   * @param target Target class or object
```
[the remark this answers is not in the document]
raymondfeng: Yes, the target will be a class for static methods.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,586 @@
...
+
+```ts
+const proxy = await ctx.get<MyController>('my-controller', {
+  asProxyWithInterceptors: true,
+});
+const msg = await proxy.greet('John');
+```
+
+Please note synchronous methods (which don't return `Promise`) are converted to
+be asynchronous in the proxy so that interceptors can be applied. For example,
```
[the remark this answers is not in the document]
raymondfeng: Great idea!

docs/site/Interceptors.md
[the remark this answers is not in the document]
raymondfeng: I added a note to the doc.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,586 @@
...
+### Interceptor functions
+
+The interceptor function has the following signature:
+
+```ts
+export interface Interceptor {
+  <T = unknown>(
+    context: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ): ValueOrPromise<T>;
```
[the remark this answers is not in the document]
raymondfeng: I made some adjustment.

raymondfeng:
> I slightly prefer the second option (throw an error).

Fixed.

raymondfeng:
> What is the reasoning behind the requirement? Can we find a different solution how to move the definition of controllers into individual tests/context blocks?

Fixed.

raymondfeng:
FYI: I extracted binding sorting functions to https://github.com/strongloop/loopback-next/pull/2848.

docs/site/Interceptors.md
```diff
@@ -0,0 +1,651 @@
...
+
+```ts
+const proxy = await ctx.get<MyController>('my-controller', {
+  asProxyWithInterceptors: true,
+});
+const msg = await proxy.greet('John');
+```
+
+Please note synchronous methods (which don't return `Promise`) are converted to
+be asynchronous in the proxy so that interceptors can be applied. For example,
```
bajtos: > Please note synchronous methods (which don't return `Promise`) are converted to
> be asynchronous in the proxy so that interceptors can be applied.

I think this is not entirely correct. IIUC, synchronous methods may be converted to asynchronous, this happen when there is at least one async interceptor. Because it's not possible to determine whether the proxied method will end up being sync or async, sync methods are converted into methods returning `ValueOrPromise`.

Could you please update the documentation to make the behavior more clear to users?

packages/context/src/interception-proxy.ts
```diff
@@ -5,25 +5,28 @@
...
+  : ValueOrPromise<T>;
 
 /**
- * The async variant of a function to always return Promise<R>. If T is not a
- * function, the type is `T`.
+ * The async variant of a function to always return `ValueOrPromise<R>`. If `T`
+ * is not a function, the type is `T`.
  */
 // tslint:disable-next-line:no-unused (possible tslint bug to treat `R` as unused)
 export type AsAsyncFunction<T> = T extends (...args: InvocationArgs) => infer R
```
bajtos: The name is not very accurate - the function is not always async, it can return the value too. I am proposing to rename this type to `AsInterceptedFunction` (describe the intent, not the implementation details).

raymondfeng:
@strongloop/loopback-maintainers I have squashed all commits into one. PTAL before I merge it.

packages/context/src/interceptor.ts
```diff
@@ -0,0 +1,432 @@
...
+    );
+    this.sortGlobalInterceptorBindings(bindings);
+    return bindings.map(b => b.key);
+  }
+
+  /**
+   * Sort global interceptor bindings by `globalInterceptorGroup` tags
+   * @param bindings An array of global interceptor bindings
+   */
+  private sortGlobalInterceptorBindings(
```
b-admike: Can this leverage new functionality introduced in #2848?

docs/site/Interceptors.md
```diff
@@ -0,0 +1,657 @@
...
+   */
+  (
+    context: InvocationContext,
+    next: () => ValueOrPromise<InvocationResult>,
+  ): ValueOrPromise<InvocationResult>;
+}
+```
+
+An interceptor can be asynchronous (returning a promise) or synchronous
+(returning a value) or. If one of the interceptors or the target method is
```
b-admike: typo: `or synchronous (returning a value) or.` -> `or synchronous (returning a value).`

packages/context/src/interceptor.ts
raymondfeng: Yes. After #2848 is landed.

Issue #2466, opened 2019-02-25
Add abstraction for authentication strategy


jannyHou:
## Feature

Define an interface to describe the authentication strategy. It should contain

- [x] a function `authenticate` that verifies the request.
- [ ] ~function `authenticate` should take in an `option` to perform the auth action accordingly.~
- [x] document if people want to use services in the authentication strategy, they can inject the service in the strategy constructor 

**Acceptance Criteria**
- [x] Verify the interface works by refactoring the Shopping Cart example to utilize it.
- [x] Verify the interface works by refactoring the Shopping Cart example to utilize it and strategies registered via extension points.

Parent story see https://github.com/strongloop/loopback-next/issues/1035#issuecomment-465311972 (step 1b)

bajtos:
How is this story different from #2311 and #2312? Is it a pre-requisite that must be implemented first, before we can work on an extension point?

As I wrote in https://github.com/strongloop/loopback-next/issues/2312#issuecomment-469598086, I think the authentication strategy abstraction should be defined around what LoopBack-specific strategies need and allow us to write an adapter to convert Passport strategies into LoopBack strategies. 

What I am trying to say: The approach shown in the current version of `@loopback/authentication`, where we use Passport's `Strategy` interface as the API contract between LoopBack strategies and the authentication component, is something we should abandon in favor of a more flexible solution.

jannyHou:
@bajtos 
> How is this story different from #2311 and #2312? Is it a pre-requisite that must be implemented first, before we can work on an extension point?

This story simply creates the interface for auth strategy, the interface will be mostly the same as what is proposed in the [design file](https://github.com/strongloop/loopback-next/pull/2576/files#diff-d7368ea1b2895005672ca0963f8ef9e9R6)

samarpanB:
@bajtos we have done oauth 2 integration as a separate component in our LB4 application very recently. We followed the guidelines shared [here](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/README.md). I am not too sure what different aspect is getting implemented here or any of the authentication tasks. But **if there is any way I can help/contribute, do let me know**. I can share our implementation as well, if needed. That's not so robust but it was workable for us. We also did authorisation component but I would mention that in the appropriate issue.

PR #2688, opened 2019-04-03
feat: introduce an authentication strategy interface

emonddr:
Introduce an authentication strategy interface

Implements #2466 

## Checklist

### DIFF
--- a/packages/authentication/docs/authentication-strategy.md
+++ b/packages/authentication/docs/authentication-strategy.md
@@ -1,20 +1,45 @@
-### Auth strategy interface
+### Authentication strategy interface
 
 ```ts
 import {Request} from '@loopback/rest';
 
-interface AuthenticationStrategy {
-  // The resolver will read the `options` object from metadata, then invoke the
-  // `authenticate` with `options` if it exists.
-  authenticate(
-    request: Request,
-    options: object,
-  ): Promise<UserProfile | undefined>;
+/**
+ * An interface that describes the common authentication strategy.
+ *
+ * An authentication strategy is a class with an
+ * 'authenticate' method that verifies a user's credentials and
+ * returns the corresponding user profile.
+ *
+ */
+export interface AuthenticationStrategy {
+  /**
+   * The 'name' property is a unique identifier for the
+   * authentication strategy ( for example : 'basic', 'jwt', etc)
+   */
+  name: string;
 
-  // This is a private function that extracts credential fields from a request,
-  // it is called in function `authenticate`. You could organize the extraction
-  // logic in this function or write them in `authenticate` directly without defining
-  // this extra utility.
-  private extractCredentials?(request: Request): Promise<Credentials>;
+  /**
+   * The 'authenticate' method takes in a given request and returns a user profile
+   * which is an instance of 'UserProfile'.
+   * (A user profile is a minimal subset of a user object)
+   * If the user credentials are valid, this method should return a 'UserProfile' instance.
+   * If the user credentials are invalid, this method should throw an error
+   * If the user credentials are missing, this method should throw an error, or return 'undefined'
+   * and let the authentication 'action' in the 'sequence' deal with it.
+   *
+   * @param request
+   */
+  authenticate(request: Request): Promise<UserProfile | undefined>;
 }
 ```
+
+An authentication strategy resolver can make use of the `name` property to
+`find` the registered authentication strategy.
+
+The authentication strategy interface has an `authenticate` function which takes
+in a request and returns a user profile.
+
+Authentication strategies that implement this interface can use dependency
+injection in the constructor to obtain **global** or **request-specific**
+`options` or any `services` it may require (a service to extract credentials
+from a request, for example).
--- a/packages/authentication/src/types.ts
+++ b/packages/authentication/src/types.ts
@@ -22,3 +22,32 @@ export interface UserProfile {
   name?: string;
   email?: string;
 }
+
+/**
+ * An interface that describes the common authentication strategy.
+ *
+ * An authentication strategy is a class with an
+ * 'authenticate' method that verifies a user's credentials and
+ * returns the corresponding user profile.
+ *
+ */
+export interface AuthenticationStrategy {
+  /**
+   * The 'name' property is a unique identifier for the
+   * authentication strategy ( for example : 'basic', 'jwt', etc)
+   */
+  name: string;
+
+  /**
+   * The 'authenticate' method takes in a given request and returns a user profile
+   * which is an instance of 'UserProfile'.
+   * (A user profile is a minimal subset of a user object)
+   * If the user credentials are valid, this method should return a 'UserProfile' instance.
+   * If the user credentials are invalid, this method should throw an error
+   * If the user credentials are missing, this method should throw an error, or return 'undefined'
+   * and let the authentication action deal with it.
+   *
+   * @param request
+   */
+  authenticate(request: Request): Promise<UserProfile | undefined>;
+}


emonddr:
Implemented the function signature as
> authenticate(request: Request, options: O): Promise<UserProfile>;
instead of 
```
authenticate(request: Request, options: O): Promise<UserProfile | undefined>;
```

because this function should throw an error if authentication fails and a `UserProfile` cannot be returned.

emonddr:
Didn't implement
```
private extractCredentials?(request: Request): Promise<Credentials>;
```
mentioned in the design document since:
- it is not possible to have a private function in an interface 
- we simply wanted to mention that the user `could` implement the `extractCredentials` function in the strategy implementation itself instead of using a separate service.

packages/authentication/src/authentication-strategy.interface.ts
```diff
@@ -0,0 +1,21 @@
+import { Request } from '@loopback/rest';
+import { UserProfile } from './types';
+
+
+/**
+ * Interface for the authentication strategy
+ */
+interface AuthenticationStrategy<O> {
```
jannyHou: nitpick: `export interface AuthenticationStrategy<O> {}`

raymondfeng: 1. I'm not keen on using `O` as a generic type name.
2. The `options` is passed in from authentication action. I don't think it's a good idea to type `authenticate` with specific options.

emonddr: @raymondfeng , yes...Janny and I spoke about this yesterday. Thanks for confirming that it isn't a good idea. There is no way to `test` that the interface `fits in` with all the providers, resolvers, actions etc except to go ahead and adjust all the surrounding files needed for a representative `PoC`. I am doing that at the moment...
This is why I placed [WIP] in the title... because I wasn't done yet.

emonddr: Removed `options` in the interface since the implementation classes of the interface can inject `global` strategy options,  and inject the `authentication metadata` into the class constructor ( thereby obtaining the `overriding options` from the metadata provided by the optional `options` parameter from the @authenticate(type,options) decorator on the controller method). In short, the implementation strategy classes can obtain the global and overriding options via injection, and so the interface doesn't need to know about them.

emonddr: Introduced the `name` property in the interface to help a `strategy resolver` find a specific strategy registered via the extension point mechanism that will be introduced to register various authentication strategies.

emonddr: Not really possible to `test` this interface in a `mocha test` until we implement many other pieces in the authorization package. Instead I refactored the shopping cart example to use this interface, used 2 authentication strategies that implemented this interface, registered them as extensions of an authentication strategy extension point as a `POC`, and everything worked together. So I am confident with this interface at point in time.

packages/authentication/docs/authentication-strategy.md
```diff
@@ -1,20 +1,16 @@
...
-  ): Promise<UserProfile | undefined>;
-
-  // This is a private function that extracts credential fields from a request,
-  // it is called in function `authenticate`. You could organize the extraction
-  // logic in this function or write them in `authenticate` directly without defining
-  // this extra utility.
-  private extractCredentials?(request: Request): Promise<Credentials>;
+export interface AuthenticationStrategy {
+  name: string;
+  authenticate(request: Request): Promise<UserProfile>;
```
bajtos: At the moment, `authenticate` can return either a user profile or `undefined`. I presume `undefined` is indicating there were no credentials provided by the client and thus the request is anonymous. (It's important to distinguish this case from the situation where the client provided some credentials but they were not valid - an error should be thrown in such case.)

In your new proposal, authentication strategies have to either return a user profile or throw an error.

How do you envision to handle anonymous requests with no credentials provided by the client?

I am not familiar with all details of the proposed authentication design. If anonymous requests are meant to be handled elsewhere in the new authentication solution, then please ignore my comment.

packages/authentication/src/types.ts
```diff
@@ -22,3 +22,13 @@ export interface UserProfile {
...
   email?: string;
 }
+
+/**
+ * Authenticates user credentials in the given request and returns a 'UserProfile'.
+ * If 'authenticate' function fails, it should throw an error.
+ */
+export interface AuthenticationStrategy {
+  name: string;
+  authenticate(request: Request): Promise<UserProfile>;
```
bajtos: Please add tsdoc comments for `name` and `authenticate`.

packages/authentication/src/authentication-strategy.interface.ts
bajtos: >  Instead I refactored the shopping cart example to use this interface, used 2 authentication strategies that implemented this interface, registered them as extensions of an authentication strategy extension point as a POC, and everything worked together. 

Please include a link to the pull request making these changes, so that we can review the outcome.

bajtos:
>  Instead I refactored the shopping cart example to use this interface, used 2 authentication strategies that implemented this interface, registered them as extensions of an authentication strategy extension point as a POC, and everything worked together. 

@emonddr Please include a link to the pull request making these changes, so that we can review the outcome.

packages/authentication/src/authentication-strategy.interface.ts
emonddr: The shopping cart example modifications I made are not production yet ready. I only used it as a personal test for myself to play around with the extensionpoint/extension decorators Raymond created.  

This interface we are introducing here identical to the `strategy interface` already working in the shopping example: https://github.com/strongloop/loopback4-example-shopping/blob/master/src/authentication-strategies/authentication.strategy.ts  wiith the exception that there is a `name:string` field to help find a strategy by name in the forthcoming  PR for registering different authentication strategies as extension to an extension point.  

Since the interface is the same as the working one in the shopping cart example, I see no point in providing a PR of the shopping cart example.

We will be revising the entire shopping cart with all the new authentication pieces that Janny and I are working on, in a separate task.

bajtos: In general, it's difficult to review changes in APIs and interfaces without seeing how the new version is going to be used. We usually write unit-tests to demonstrate the usage and verify the implementation. I understand this may be not practical here and that's why I am asking to see code using the new interface.

> This interface we are introducing here identical to the strategy interface already working in the shopping example: https://github.com/strongloop/loopback4-example-shopping/blob/master/src/authentication-strategies/authentication.strategy.ts wiith the exception that there is a name:string field to help find a strategy by name in the forthcoming PR for registering different authentication strategies as extension to an extension point.

Great! This is exactly what I was looking for.

However! In the example, the `authenticate` method can return `undefined`, while your code is requiring the method to always return a user profile.

```ts
export interface AuthenticationStrategy {
  authenticate(request: Request): Promise<UserProfile | undefined>;
}
```

I have already pointed out this aspect, let's continue this discussion in https://github.com/strongloop/loopback-next/pull/2688#pullrequestreview-226512546 please.

packages/authentication/docs/authentication-strategy.md
emonddr: @bajtos , what is meant by `anonymous` request?  In the controller classes, there are endpoints that won't be decorated with `@authenticate( '{authentication strategy name}'  )`, and some endpoints that will be decorated with it.  The authenticate action in the sequence will not attempt any authentication on an endpoint which doesn't have an authentication strategy specified. If a strategy is provided, that strategy will return a user if and only if the user credentials are valid, otherwise the strategy will throw an http error.

emonddr:
> > Instead I refactored the shopping cart example to use this interface, used 2 authentication strategies that implemented this interface, registered them as extensions of an authentication strategy extension point as a POC, and everything worked together.
> 
> @emonddr Please include a link to the pull request making these changes, so that we can review the outcome.

Please see my comment above. thx.

packages/authentication/docs/authentication-strategy.md
bajtos: By `anonymous` request I mean a request that does not include any credentials. For example, by opening http://localhost:3000/orders in your browser.

Let's say this that the controller method implementing `/orders` is decorated with `@authenticate(/*some strategy*/)`.

In the code we have so far, the authentication strategy returns `undefined`, meaning there are no credentials provided for the request, and it's up to `authenticate()` sequence action or even the sequence itself to decide how to handle the situation when an anonymous request is received for a resource that requires authentication.

IIUC your proposal correctly, you are pushing the responsibility of handling this situation to individual authentication strategies.

I have two concerns about that approach:
 - Different strategy can throw different HTTP error. Applications using multiple strategies can end up with inconsistent error responses depending on which strategy was applied.
 - It's more difficult to implement fallbacks, i.e. configure the app so that if one strategy did not find any credentials in the request, then we can try a different strategy to see if it's more successful.

Anyhow, I am not familiar with all details of the solution you are aiming for. If the change of `authenticate` signature is in line with your new overall design, then I am fine with that. As long as it's a conscious decision made after considering pros and cons of different approaches.

raymondfeng:
@emonddr Please fix the CI failure - https://travis-ci.org/strongloop/loopback-next/jobs/520443467

packages/authentication/src/types.ts
```diff
@@ -22,3 +22,30 @@ export interface UserProfile {
...
+   */
+  name: string;
+
+  /**
+   * The 'authenticate' method takes in a given request and returns a user profile
+   * which is an instance of 'UserProfile'.
+   * (A user profile is a minimal subset of a user object)
+   * If the user credentials are valid, this method will return a 'UserProfile' instance.
+   * If the user credentials are invalid, this method will throw an error
+   *
```
bajtos: Please explain what is the expected strategy behavior when no credentials were found in the request.

For example:

```
If no user credentials were found or the user credentials are invalid,
this method will throw an error.
```

emonddr: > Please explain what is the expected strategy behavior when no credentials were found in the request.

Hi @bajtos , the authenticate method on a strategy will throw an error like: https://github.com/strongloop/loopback4-example-shopping/blob/master/src/authentication-strategies/JWT.strategy.ts#L34.

packages/authentication/docs/authentication-strategy.md
b-admike: @emonddr Can we keep the `undefined` in the return signature of the `authenticate` method given the above discussion and the points made by @bajtos? I am also leaning towards letting the sequence action or sequence handle anonymous requests in a unified way as opposed to delegating it to authentication strategies.

emonddr: Currently the shopping cart `sequence` looks like this:

```ts
  constructor(
    @inject(SequenceActions.FIND_ROUTE) protected findRoute: FindRoute,
    @inject(SequenceActions.PARSE_PARAMS) protected parseParams: ParseParams,
    @inject(SequenceActions.INVOKE_METHOD) protected invoke: InvokeMethod,
    @inject(SequenceActions.SEND) public send: Send,
    @inject(SequenceActions.REJECT) public reject: Reject,
    @inject(AuthenticationBindings.AUTH_ACTION)
    protected authenticateRequest: AuthenticateFn,
  ) {}

    try {

      await this.authenticateRequest(request);
      this.send(response, result);
    } catch (err) {
      this.reject(context, err);
    }
  }
}
```
Currently the shopping cart `authentication` action in the sequence looks like this:

```ts
export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
  constructor(
    // The provider is instantiated for Sequence constructor,
    // at which time we don't have information about the current
    // route yet. This information is needed to determine
    // what auth strategy should be used.
    // To solve this, we are injecting a getter function that will
    // defer resolution of the strategy until authenticate() action
    // is executed.
    @inject.getter(AuthenticationBindings.STRATEGY)
    readonly getStrategy: Getter<AuthenticationStrategy>,
    @inject.setter(AuthenticationBindings.CURRENT_USER)
    readonly setCurrentUser: Setter<UserProfile>,
  ) {}

  /**
   * @returns authenticateFn
   */
  value(): AuthenticateFn {
    return request => this.action(request);
  }

  /**
   * The implementation of authenticate() sequence action.
   * @param request The incoming request provided by the REST layer
   */
  async action(request: Request): Promise<UserProfile | undefined> {
    const strategy = await this.getStrategy();
    if (!strategy) {
      // The invoked operation does not require authentication.
      return undefined;
    }
    if (!strategy.authenticate) {
      throw new Error('invalid strategy parameter');
    }
    const user = await strategy.authenticate(request);
    if (user) this.setCurrentUser(user);
    return user;
  }
}
```

emonddr: @jannyHou 's new `authentication` action for `authentication` package will look similar. See : https://github.com/strongloop/loopback-next/issues/2467 .


In the `authentication` action,
If a request comes through from an endpoint that is not decorated with a strategy(in controller via @authenticate('jwt') for example), then the action won't find a strategy and returns. 

If a request comes throw that does have a strategy associated with it, that strategy's authenticate method is called and returns a user profile which is placed on the context. (Or it throws an error, and that error percolates up the try/catch in the sequence which has an error handler. I don't see the point for an authentication strategy to return `undefined`.

emonddr: Are you saying that you want the `authenticate` method of a strategy to return { user: {value or undefined}, error: {value or undefined}}, and never throws an error, and it is up to the action to standardize the error handling in its block of code (regardless of strategy)?  In this case, still... the strategy.authenticate will always return an object, but it is the action that will handle throwing errors.

bajtos: I am asking to preserve the signature used by the current strategy interface in the shopping cart.

```ts
export interface AuthenticationStrategy {
  name: string;
  authenticate(request: Request): Promise<UserProfile | undefined>;
}
```

Take a look at the implementation of AuthenticateFn you posted above - it's already prepared to handle the case when the authentication strategy did not find any user credentials.

```ts
    const user = await strategy.authenticate(request);
    if (user) this.setCurrentUser(user);
    return user;
```

packages/authentication/src/types.ts
bajtos: > > Please explain what is the expected strategy behavior when no credentials were found in the request.
>
> Hi @bajtos , the authenticate method on a strategy will throw an error like: https://github.com/strongloop/loopback4-example-shopping/blob/master/src/authentication-strategies/JWT.strategy.ts#L34.

As I understand that code, a different error will be thrown, see here:
https://github.com/strongloop/loopback4-example-shopping/blob/91805dd8740dce8b19bd7af613c485f7935cc8b5/src/authentication-strategies/JWT.strategy.ts#L21-L22

```ts
    let token = request.query.access_token || request.headers['authorization'];
    if (!token) throw new HttpErrors.Unauthorized('No access token found!');
```

> An error should be thrown by the strategy if the controller method is decorated with an authentication strategy BUT the user has not passed in any credentials

Makes sense (if that's the indented design). I am asking you to capture this information in the tsdoc comment for the new interface.

emonddr: This code is from shopping cart.. it is old code. Going forward the code may be different. The point is that some error is thrown for a given strategy.authenticate().
This PR is about an interface not implementation.

Issue #2628, opened 2019-03-22
Fix repository-json-schema to handle circular references


bajtos:
## Description / Steps to reproduce / Feature proposal

Consider the following test case:

```ts
it('handles circular references', () => {
  @model()
  class Category {
    @property.array(() => Product)
    products?: Product[];
  }

  @model()
  class Product {
    @property(() => Category)
    category?: Category;
  }

  const schema = modelToJsonSchema(Category);
  expect(schema).to.deepEqual({
    title: 'Category',
    properties: {
      products: {
        type: 'array',
        items: {$ref: '#/definitions/Product'},
      },
    },
    definitions: {
      Product: {
        title: 'Product',
        properties: {
          category: {
            $ref: '#/definitions/Category',
          },
        },
      },
    },
  });
});
```

Please note this test case is synthetic and does not represent the actual design we will use for relation navigational properties.

## Current Behavior

The test fails with a stack-overflow error.

```
       handles circular references:
     RangeError: Maximum call stack size exceeded
      at RegExp.test (<anonymous>)
      at isTypeResolver (/Users/bajtos/src/loopback/next/packages/repository/src/type-resolver.ts:44:16)
      at Object.resolveType (/Users/bajtos/src/loopback/next/packages/repository/src/type-resolver.ts:88:10)
      at metaToJsonProperty (src/build-schema.ts:148:24)
      at modelToJsonSchema (src/build-schema.ts:241:28)
      at getJsonSchema (src/build-schema.ts:38:23)
      at modelToJsonSchema (src/build-schema.ts:263:24)
      at getJsonSchema (src/build-schema.ts:38:23)
      at modelToJsonSchema (src/build-schema.ts:263:24)
      at getJsonSchema (src/build-schema.ts:38:23)
      at ` (src/build-schema.ts:263:24)
      ...
```

## Expected Behavior

The tests passes.

## Additional information

See the spike #2592 and the commit https://github.com/strongloop/loopback-next/pull/2592/commits/2256435f2affe7c3962267dc4d2a2148cb15a6d5 for more details.

The idea is to introduce a new `options` argument to `getJsonSchema` and `modelToJsonSchema` functions, this argument will expect the following properties to begin with:

```ts
export interface JsonSchemaOptions {
  visited?: {[key: string]: JSONSchema};
}
```

## Acceptance criteria

- [ ] Make the changes necessary to make the above test pass
- [ ] Add additional test coverage at unit-test and integration level as necessary
[...]

Issue #2653, opened 2019-03-28
Emit schema excluding certain model properties


bajtos:
Allow Controller methods implementing CREATE operation to describe their input data as "a model without `id` and `_rev` properties ". This story requires #2629 and #2631 to be implemented first.

To emit schema with certain properties excluded, we should add a new schema-generation option called `exclude` & accepting a list of property names to remove from the schema. To allow TypeScript compiler to verify that `exclude` items are valid property names, we should leverage TypeScript generics and `keyof` keyword.

```ts
export interface JsonSchemaOptions<T extends object = any>{
  /// List of properties to exclude from the schema
  exclude?: (keyof T)[];
}
```

An example showing a controller method excluding the property `id`:

```ts
class TodoListController {
  // ...

  @post('/todo-lists', {
    responses: {
      // left out for brevity
    },
  })
  async create(
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(TodoList, {exclude: ['id']}),
          /***** ^^^ THIS IS IMPORTANT - OPENAPI SCHEMA ^^^ ****/
        },
      },
    })
    obj: Omit<TodoList, 'id'>,
    /***** ^^^ THIS IS IMPORTANT - TYPESCRIPT TYPE ^^^ ****/
  ): Promise<TodoList> {
    return await this.todoListRepository.create(obj);
  }
}
```

See https://github.com/strongloop/loopback-next/pull/2646/commits/887ae847c04c29fab01735e3c54db141f3cf53ed for a spike showing proposed implementation details.

## Acceptance criteria

- ~~A type alias for `Pick<T, Exclude<keyof TodoList, K>>`, see the discussion in https://github.com/strongloop/loopback-next/pull/2646/commits/887ae847c04c29fab01735e3c54db141f3cf53ed#r269305305~~ - not needed, TypeScript has recently introduced `Omit` type.
- [x] Implementation including test coverage
- [x] API documentation
- [x] `CREATE` endpoints in all example applications are updated to leverage `exclude` option to remove `id` property and use the new type alias as the type of the controller method argument accepting model data.
- [x] CLI templates are updated accordingly
- [ ] Should have - create a new issue if it's too difficult to implement:

    Controller methods working with related models should exclude the foreign key from request body properties. Cross-posting from https://github.com/strongloop/loopback-next/issues/1872#issuecomment-503571828:

[...]

Issue #2652, opened 2019-03-28
Emit schema with all model properties optional


bajtos:
Allow Controller methods implementing PATCH operation to describe their input data as "a model with all properties optional". This story requires #2629 and #2631 to be implemented first.

To emit schema with optional model properties, we should add a new schema-generation option called `partial`.

```ts
export interface JsonSchemaOptions {
  /// Make all properties optional
  partial?: boolean;
}
```

An example showing a controller method accepting a partial model instance:

```ts
class TodoListController {
  // ...

  @patch('/todo-lists/{id}', {
    responses: {
      // left out for brevity
    },
  })
  async updateById(
    @param.path.number('id') id: number,
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(TodoList, {partial: true}),
          /***** ^^^ THIS IS IMPORTANT - OPENAPI SCHEMA ^^^ ****/
       },
     },
    })
    obj: Partial<TodoList>,
    /***** ^^^ THIS IS IMPORTANT - TYPESCRIPT TYPE ^^^ ****/
  ): Promise<void> {
    await this.todoListRepository.updateById(id, obj);
  }
}
```

See https://github.com/strongloop/loopback-next/pull/2646/commits/887ae847c04c29fab01735e3c54db141f3cf53ed for a spike showing proposed implementation details.

## Acceptance criteria

- [x]  Implementation including test coverage - see https://github.com/strongloop/loopback-next/pull/3199
- [x] API documentation - see https://github.com/strongloop/loopback-next/pull/3199
- [x] `PATCH` endpoints in all example applications are updated to leveraga `partial` option and use `Partial<MyModel>` as the type of the controller method argument accepting model data - see https://github.com/strongloop/loopback-next/pull/3199
- [x] CLI templates are updated accordingly - see https://github.com/strongloop/loopback-next/pull/3199
- [x] Review #1722 and #1179, either close them as fixed or post a comment explaining what's remaining to be done.

bajtos:
Done 🎉

PR #2646, opened 2019-03-26
Spike: Constrain the OpenAPI schema emitted for a model [DO NOT MERGE]

bajtos:
This pull request offers a PoC implementation demonstrating how to modify the schema generated from a LB4 model depending on configurable options. In particular, the following two use cases are covered:

- PATCH - all properties are optional
- CREATE - id and possibly _rev must not be present

The first commit imports code from https://github.com/strongloop/loopback-next/pull/2592 and should be ignored.

The second commit contains PoC and SPIKE notes with a high-level overview.

Related stories:
 - the spike story: https://github.com/strongloop/loopback-next/issues/2082
 - #1722
 - #1179
 - #1470
 - #1719
 - #1894

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

PR #2690, opened 2019-04-04
fix(repository-json-schema): resolve the circular reference

jannyHou:
Fixes https://github.com/strongloop/loopback-next/issues/2628

## Description

The solution here is a little bit different than the proposal in the original issue:

- Ignore this item. ~The `definitions` contains both `Category` and `Product`.(This is not true) So that~

```
...
properties: {
    category: {
        $ref: '#/definitions/Category',
    },
},
...
```
~refers to a valid definition entry.~

- The `visited` field is ~an array~ a set tracks the title of visited items instead of caching the schema. This is because `getJsonSchema` is called during each iteration of the property schema generation, not after the completed model schema(without definitions) get built.

## Checklist

- [ ] ~Affected artifact templates in `packages/cli` were updated~
- [ ] ~Affected example projects in `examples/*` were updated~

### DIFF
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -591,6 +591,45 @@ describe('build-schema', () => {
       });
     });
 
+    context('model conversion', () => {
+      @model()
+      class Category {
+        @property.array(() => Product)
+        products?: Product[];
+      }
+
+      @model()
+      class Product {
+        @property(() => Category)
+        category?: Category;
+      }
+
+      const expectedSchema = {
+        title: 'Category',
+        properties: {
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/Product'},
+          },
+        },
+        definitions: {
+          Product: {
+            title: 'Product',
+            properties: {
+              category: {
+                $ref: '#/definitions/Category',
+              },
+            },
+          },
+        },
+      };
+
+      it('handles circular references', () => {
+        const schema = modelToJsonSchema(Category);
+        expect(schema).to.deepEqual(expectedSchema);
+      });
+    });
+
     function expectValidJsonSchema(schema: JsonSchema) {
       const ajv = new Ajv();
       const validate = ajv.compile(
@@ -639,6 +678,55 @@ describe('build-schema', () => {
         newProperty: {
           type: 'string',
         },
+      });
+    });
+    it('does not pollute the JSON schema options', () => {
+      @model()
+      class Category {
+        @property()
+        name: string;
+      }
+
+      const JSON_SCHEMA_OPTIONS = {};
+      getJsonSchema(Category, JSON_SCHEMA_OPTIONS);
+      expect(JSON_SCHEMA_OPTIONS).to.be.empty();
+    });
+    context('circular reference', () => {
+      @model()
+      class Category {
+        @property.array(() => Product)
+        products?: Product[];
+      }
+
+      @model()
+      class Product {
+        @property(() => Category)
+        category?: Category;
+      }
+
+      const expectedSchemaForCategory = {
+        title: 'Category',
+        properties: {
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/Product'},
+          },
+        },
+        definitions: {
+          Product: {
+            title: 'Product',
+            properties: {
+              category: {
+                $ref: '#/definitions/Category',
+              },
+            },
+          },
+        },
+      };
+
+      it('generates the schema without running into infinite loop', () => {
+        const schema = getJsonSchema(Category);
+        expect(schema).to.deepEqual(expectedSchemaForCategory);
       });
     });
   });
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -14,18 +14,27 @@ import {
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY} from './keys';
 
+export interface JsonSchemaOptions {
+  visited?: {[key: string]: JSONSchema};
+}
+
 /**
  * Gets the JSON Schema of a TypeScript model/class by seeing if one exists
  * in a cache. If not, one is generated and then cached.
  * @param ctor Contructor of class to get JSON Schema from
  */
-export function getJsonSchema(ctor: Function): JSONSchema {
-  // NOTE(shimks) currently impossible to dynamically update
-  const jsonSchema = MetadataInspector.getClassMetadata(JSON_SCHEMA_KEY, ctor);
-  if (jsonSchema) {
-    return jsonSchema;
+export function getJsonSchema(
+  ctor: Function,
+  options?: JsonSchemaOptions,
+): JSONSchema {
+  // In the near future the metadata will be an object with
+  // different titles as keys
+  const cached = MetadataInspector.getClassMetadata(JSON_SCHEMA_KEY, ctor);
+
+  if (cached) {
+    return cached;
   } else {
-    const newSchema = modelToJsonSchema(ctor);
+    const newSchema = modelToJsonSchema(ctor, options);
     MetadataInspector.defineMetadata(JSON_SCHEMA_KEY.key, newSchema, ctor);
     return newSchema;
   }
@@ -142,16 +151,26 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
  * reflection API
  * @param ctor Constructor of class to convert from
  */
-export function modelToJsonSchema(ctor: Function): JSONSchema {
+export function modelToJsonSchema(
+  ctor: Function,
+  jsonSchemaOptions: JsonSchemaOptions = {},
+): JSONSchema {
+  const options = {...jsonSchemaOptions};
+  options.visited = options.visited || {};
+
   const meta: ModelDefinition | {} = ModelMetadataHelper.getModelMetadata(ctor);
-  const result: JSONSchema = {};
 
   // returns an empty object if metadata is an empty object
   if (!(meta instanceof ModelDefinition)) {
     return {};
   }
 
-  result.title = meta.title || ctor.name;
+  const title = meta.title || ctor.name;
+
+  if (options.visited[title]) return options.visited[title];
+
+  const result: JSONSchema = {title};
+  options.visited[title] = result;
 
   if (meta.description) {
     result.description = meta.description;
@@ -190,20 +209,24 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
       continue;
     }
 
-    const propSchema = getJsonSchema(referenceType);
+    const propSchema = getJsonSchema(referenceType, options);
 
-    if (propSchema && Object.keys(propSchema).length > 0) {
+    includeReferencedSchema(referenceType.name, propSchema);
+
+    function includeReferencedSchema(name: string, schema: JSONSchema) {
+      if (!schema || !Object.keys(schema).length) return;
       result.definitions = result.definitions || {};
 
-      // delete nested definition
-      if (propSchema.definitions) {
-        for (const key in propSchema.definitions) {
-          result.definitions[key] = propSchema.definitions[key];
+      // promote nested definition to the top level
+      if (schema.definitions) {
+        for (const key in schema.definitions) {
+          if (key === title) continue;
+          result.definitions[key] = schema.definitions[key];
         }
-        delete propSchema.definitions;
+        delete schema.definitions;
       }
 
-      result.definitions[referenceType.name] = propSchema;
+      result.definitions[name] = schema;
     }
   }
   return result;


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -14,18 +14,27 @@ import {
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY} from './keys';
 
+export interface JsonSchemaOptions {
+  // Track the models/titles that have been visited.
+  visited?: string[];
```
raymondfeng: Let's use `Set` instead to include visited TypeScript classes.

raymondfeng:
Please run `npm run update-package-locks` or `npm run update-packages` to fix package-lock.json files.

packages/repository-json-schema/src/build-schema.ts
bajtos: +1 to use a `Set` instead of an array.

FYI: In a very near feature, we will be emitting multiple schemas for a single model class:
 - Vanilla model schema as today
 - Model schema including relations
 - Model schema excluding certain properties (`id`, `_rev`, etc.)

Using TypeScript classes as the key will not work for that. I think it's better to use Schema title because that has to be unique. See https://github.com/strongloop/loopback-next/issues/2653,  https://github.com/strongloop/loopback-next/issues/2652 and https://github.com/strongloop/loopback-next/pull/2646/files#diff-432476dd2d984e441d8b4e0af39a39be.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -148,6 +160,8 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
   }
 
   result.title = meta.title || ctor.name;
+  const isVisited =
+    options && options.visited && options.visited.includes(result.title!);
```
bajtos: Let's set `options` to `{}` by default so that we don't have to check for its existence.

```ts
export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions = {}
): JSONSchema {
  // ...

  const isVisited =
    options.visited && options.visited.includes(result.title!);

  // ...
}
```

bajtos: Please avoid using `!` operator to silence TypeScript checks.

```ts
const title = meta.title || ctor.name;
const isVisited = options.visited && options.visited.includes(title);

result.title = title;
```

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -186,7 +200,18 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
       continue;
     }
 
-    const propSchema = getJsonSchema(referenceType);
+    // Break to avoid the circular reference.
+    // If the model is already visited in the call stack, it implies one or more
+    // of its referenced properties refer back to it.
+
+    if (isVisited) break;
```
bajtos: I am confused. Why are we checking `isVisited` deep inside the loop iterating over model properties?

In the spike https://github.com/strongloop/loopback-next/commit/2256435f2affe7c3962267dc4d2a2148cb15a6d5, I was able to return early in the schema generation. Once we know the schema title, we can check whether we have already seen such schema and immediately return.

What am I missing?

> The visited field is an array tracks the title of visited items instead of caching the schema. This is because `getJsonSchema` is called during each iteration of the property schema generation, not after the completed model schema(without definitions) get built.

This should not matter. In our cache (both `visited` and metadata-based cache used by `getJsonSchema`), we are storing a schema object by reference. Even if we initially store any empty schema object first with no properties, the code executed by `modelToJsonSchema` will eventually fill in those properties directly in the cached object.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -186,7 +200,18 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
...
+
+    // Use object assign to avoid polluting the original `options`.
```
bajtos: +1

Please add a test to verify.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -552,6 +552,54 @@ describe('build-schema', () => {
...
+                type: 'array',
+                items: {$ref: '#/definitions/Product'},
+              },
+            },
+          },
+        },
+      };
+
+      it('handles circular references', () => {
+        const schema = modelToJsonSchema(Category);
```
bajtos: Please add a test to verify `getJsonSchema` too.

.vscode/settings.json
```diff
@@ -1,5 +1,7 @@
 {
-  "editor.rulers": [80],
+  "editor.rulers": [
+    80
```
bajtos: Unrelated change, please revert.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -186,7 +200,18 @@ export function modelToJsonSchema(ctor: Function): JSONSchema {
...
+    const getJsonSchemaOptions = Object.assign({}, options);
+    getJsonSchemaOptions.visited = getJsonSchemaOptions.visited || [];
+    getJsonSchemaOptions.visited.push(result.title!);
+
+    const propSchema = getJsonSchema(referenceType, getJsonSchemaOptions);
 
     if (propSchema && Object.keys(propSchema).length > 0) {
       result.definitions = result.definitions || {};
```
bajtos: One idea: since we are effectively storing the schema of visited models in `result.definitions`; if we need to access schema of a model that has been already visited, we can perhaps obtain it from those definitions object? Or maybe we should remove `options.visited` and pass the `definitions` key-value map in options instead?

.vscode/settings.json
jannyHou: @bajtos The format is changed when auto save. And if you check the other settings, breaking it into multiple the lines is the expected format.

I can try to move the change into a separate commit.

packages/repository-json-schema/src/build-schema.ts
jannyHou: @bajtos The set tracks `title` instead of the class name, see [code](https://github.com/strongloop/loopback-next/pull/2690/files#diff-432476dd2d984e441d8b4e0af39a39beR211)

Or do you want to store the whole generated schema in the `options.visited`?
That would not be possible IMO, since generating the referenced definitions happen in [the iteration](https://github.com/strongloop/loopback-next/blob/f6fe8eb98ef5f8a9310f0e01453cbac3d32dccf8/packages/repository-json-schema/src/build-schema.ts#L168), before all the properties are visited.

packages/repository-json-schema/src/build-schema.ts
jannyHou: > One idea: since we are effectively storing the schema of visited models in result.definitions

Hmm, do you mean storing the schema of the current model in the `definitions`?

Like 
```
{
        title: 'Category',
        properties: {
          name: {
            type: 'string'
          },
        },
        definitions: {
          Category: {
            title: 'Category',
            properties: {
               name: {
                 type: 'string'
               },
            },
          },
        },
}
```

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: WIP, will add it in the next commit.

packages/repository-json-schema/src/build-schema.ts
jannyHou: WIP, will add it in the next commit.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: I didn't realize `getJsonSchema` is the entry point of the schema generation, thought it starts with `modelToJsonSchema`.

Good catch! Tests added.

packages/repository-json-schema/src/build-schema.ts
jannyHou: You are right. My expected JSON schema was wrong, I thought the `definitions` field from `modelToJsonSchema(Category)` should contain the schema of Category itself. 

I refactored the code to return an empty array when the model is visited, which is still slightly different from your PoC.

> In our cache (both visited and metadata-based cache used by getJsonSchema), we are storing a schema object by reference. Even if we initially store any empty schema object first with no properties, the code executed by modelToJsonSchema will eventually fill in those properties directly in the cached object.

Sounds reasonable to me...I am trying the code in PoC while run into the circular error again. I might miss something, need more time to figure it out.

jannyHou:
@bajtos @raymondfeng Thanks for the review, feedback applied.

@bajtos My implementation is still slightly different than your PoC, see my [comment](https://github.com/strongloop/loopback-next/pull/2690#discussion_r274245785)

I may need some time to explore you original approach then compare.

.vscode/settings.json
bajtos: > The format is changed when auto save. And if you check the other settings, breaking it into multiple the lines is the expected format.

Fair enough :+1:

> I can try to move the change into a separate commit.

Yes please. A new PR would be even better, that way we can land this small fix sooner.

packages/repository-json-schema/src/build-schema.ts
bajtos: > Hmm, do you mean storing the schema of the current model in the `definitions`?

No, only schema of models used by properties (i.e. `Product` when emitting schema for `Category`).

packages/repository-json-schema/src/build-schema.ts
bajtos: > The set tracks title instead of the class name

Cool, that should work well :+1:

> do you want to store the whole generated schema in the options.visited?

IIRC, in my PoC, I was storing the whole generated schema in `options.visited` to allow me to load it from `options.visited` and put it to `result.definitions`.

bajtos:
> My implementation is still slightly different than your PoC, see my comment
>
> I may need some time to explore you original approach then compare.

Sure, take your time. My spike code is just a prototype, it's fine to end up with a different implementation. As long as it supports the use cases we need to support.

jannyHou:
@bajtos I figured out the difference. Your PoC skipped prompting visited definition to the top level with [this line of code](https://github.com/strongloop/loopback-next/pull/2690/files#diff-432476dd2d984e441d8b4e0af39a39beR221), which simplifies everything!

I still didn't get the idea of tracking the visited model with its generated schema like `{CategoryModelOnly: ...CategoryModelOnlySchema}`. IMO using a set containing all the visited models(titles to be more accurate) is enough, it's light than storing the entire schema in the option. Thought?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -142,16 +151,24 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
...
 
   // returns an empty object if metadata is an empty object
   if (!(meta instanceof ModelDefinition)) {
     return {};
   }
 
-  result.title = meta.title || ctor.name;
+  const title = meta.title || ctor.name;
+
+  if (options.visited.has(title)) return {};
```
bajtos: IIUC, by returning an empty object, you are trying to satisfy the return value type `JSONSchema` while returning an invalid (or at least useless) value. That's effectively bypassing type checks :(

I think this was the reason why I was keeping the visited schemas in the cache?

Anyhow. Please change this line to `return undefined` to make it clear that when a model has been already visited, then we are not returning it's schema. 

To preserve backwards compatibility, I am proposing the following signature:

```ts
export function modelToJsonSchema(ctor: Function): JSONSchema;
export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions = {},
): JSONSchema | undefined;

export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions = {},
): JSONSchema | undefined {
  // the implementation
}
```

Alternatively, and I think this is a better option when it comes to type safety, promote `visited` to a top-level argument and keep the `JsonSchemaOptions` as an empty interface for now.

```ts
export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions = {}
): JSONSchema;

export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions,
  visited: Set<string>;
): JSONSchema | undefined;

export function modelToJsonSchema(
  ctor: Function,
  options: JsonSchemaOptions = {},
  visited?: Set<string>;
): JSONSchema | undefined {
  // the implementation
}
```

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -142,16 +151,24 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
  * reflection API
  * @param ctor Constructor of class to convert from
  */
-export function modelToJsonSchema(ctor: Function): JSONSchema {
+export function modelToJsonSchema(
+  ctor: Function,
+  options: JsonSchemaOptions = {},
+): JSONSchema {
+  options.visited = options.visited || new Set<string>();
```
bajtos: Please don't modify original `options`.

```suggestion
  if (!options.visited) options = {...options, visited: new Set<string>()};
```

Please add a test to verify.

packages/repository-json-schema/src/build-schema.ts
jannyHou: > That's effectively bypassing type checks :(
I think this was the reason why I was keeping the visited schemas in the cache?

Oh...that's true!

> by returning an empty object, you are trying to satisfy the return value type JSONSchema while returning an invalid (or at least useless) value.

I agree.
And `undefined` and `{}` both return useless value which logically isn't as perfect as returning the cache value.

And since the cached schema is just a reference, tracking with a Set won't save space compared with tracking with cache. I switched to the approach in your PoC.

packages/repository-json-schema/src/build-schema.ts
jannyHou: > Please don't modify original options.

Hmm, any reason?

bajtos: It's confusing for callers when you modify the `options` object. The caller may use the same `options` instance for multiple calls.

For example:

```ts
const SCHEMA_OPTIONS = {includeRelations: true};

const productSchema = getJsonSchema(Product, SCHEMA_OPTIONS);
const categorySchema = getJsonSchema(Category, SCHEMA_OPTIONS);
```

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -193,8 +200,8 @@ export function modelToJsonSchema(
     const resolvedType = resolveType(metaProperty.type) as string | Function;
     const referenceType = isArrayType(resolvedType)
       ? // shimks: ugly type casting; this should be replaced by logic to throw
-      // error if itemType/type is not a string or a function
-      resolveType(metaProperty.itemType as string | Function)
+        // error if itemType/type is not a string or a function
+        resolveType(metaProperty.itemType as string | Function)
```
bajtos: This looks like an unrelated whitespace change to me, PTAL

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -142,16 +151,25 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
  * reflection API
  * @param ctor Constructor of class to convert from
  */
-export function modelToJsonSchema(ctor: Function): JSONSchema {
+export function modelToJsonSchema(
+  ctor: Function,
+  options: JsonSchemaOptions = {},
+): JSONSchema {
+  options.visited = options.visited || {};
```
bajtos: Please don't modify original `options`, see the discussion above.

packages/repository-json-schema/src/build-schema.ts
jannyHou: 👍 got it thank you.

packages/repository-json-schema/src/build-schema.ts
jannyHou: added

.vscode/settings.json
jannyHou: It turns out it's just my vscode's weird settings.
The other settings are broken down into multiple lines due to the max characters per line.
Reverting.

packages/repository-json-schema/src/build-schema.ts
jannyHou: applied and new test case added.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -641,5 +680,55 @@ describe('build-schema', () => {
         },
       });
     });
+    it('does not pollute the JSON schema options', () => {
```
b-admike: I wonder if we can treat this as a unit test instead. Feel free to ignore.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -641,5 +680,55 @@ describe('build-schema', () => {
...
+      @model()
+      class Category {
+        @property()
+        name: string;
+      }
+
+      const JSON_SCHEMA_OPTIONS = {};
+      // tslint:disable-next-line:no-unused
```
nabdelgadir: Question, why do you need the `const categorySchema =` part instead of just calling `getJsonSchema(Category, JSON_SCHEMA_OPTIONS);`?

raymondfeng:
@jannyHou Do we have a test to make sure such json schemas are converted to OpenAPI spec correctly? For example, using:

```yaml
components
  - schemas
     - Product
```

jannyHou:
@raymondfeng Yeah I believe we have it in the `openapi-v3` package, see [test](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts#L71-L85)

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: Good catch!

PR #2704, opened 2019-04-08
feat(build): add more TypeScript "strict" checks

bajtos:
Enable all "strict" checks with the exception of `strictPropertyInitialization` which would require significant changes.

The following additional checks are added:

 - noImplicitThis
 - alwaysStrict
 - strictFunctionTypes

In the future, any checks added to TypeScript "strict" mode will be automatically enabled for LoopBack projects too.

See also [TypeScript Compiler Options](https://www.typescriptlang.org/docs/handbook/compiler-options.html)

## Checklist

### DIFF
--- a/packages/context/src/context-view.ts
+++ b/packages/context/src/context-view.ts
@@ -85,10 +85,11 @@ export class ContextView<T = unknown> extends EventEmitter
   /**
    * Find matching bindings and refresh the cache
    */
-  protected findBindings() {
+  protected findBindings(): Readonly<Binding<T>>[] {
     debug('Finding matching bindings');
-    this._cachedBindings = this.context.find(this.filter);
-    return this._cachedBindings;
+    const found = this.context.find(this.filter);
+    this._cachedBindings = found;
+    return found;
   }
 
   /**
--- a/benchmark/src/benchmark.ts
+++ b/benchmark/src/benchmark.ts
@@ -5,7 +5,7 @@
 
 import * as byline from 'byline';
 import {ChildProcess, spawn} from 'child_process';
-import pEvent from 'p-event';
+import pEvent, {Emitter} from 'p-event';
 import {Autocannon, EndpointStats} from './autocannon';
 import {Client} from './client';
 import {scenarios} from './scenarios';
@@ -111,5 +111,10 @@ function startWorker() {
 
 async function closeWorker(worker: ChildProcess) {
   worker.kill();
-  await pEvent(worker, 'close');
+  await pEvent(
+    // workaround for a bug in pEvent types which makes them
+    // incompatible with "strictFunctionTypes"
+    worker as Emitter<[unknown]>,
+    'close',
+  );
 }
--- a/packages/build/config/tsconfig.common.json
+++ b/packages/build/config/tsconfig.common.json
@@ -3,11 +3,12 @@
   "compilerOptions": {
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "noImplicitAny": true,
-    "strictNullChecks": true,
     "resolveJsonModule": true,
-    "strictBindCallApply": true,
     "skipLibCheck": true,
+    "strict": true,
+
+    // FIXME(bajtos) LB4 is not compatible with this setting yet
+    "strictPropertyInitialization": false,
 
     "lib": ["es2018", "esnext.asynciterable"],
     "module": "commonjs",
--- a/packages/core/src/lifecycle.ts
+++ b/packages/core/src/lifecycle.ts
@@ -62,7 +62,7 @@ export function asLifeCycleObserver<T = unknown>(binding: Binding<T>) {
  * Find all life cycle observer bindings. By default, a binding tagged with
  * `CoreTags.LIFE_CYCLE_OBSERVER`. It's used as `BindingFilter`.
  */
-export function lifeCycleObserverFilter(binding: Binding<unknown>): boolean {
+export function lifeCycleObserverFilter(binding: Readonly<Binding>): boolean {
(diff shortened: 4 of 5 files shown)


packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -458,8 +458,8 @@ describe('Context bindings - Injecting dependencies of classes', () => {
         'location',
         {},
         // Set up a custom resolve() to access information from the session
-        (c: Context, injection: Injection, session: ResolutionSession) => {
-          resolutionPath = session.getResolutionPath();
+        (c: Context, injection: Injection, session?: ResolutionSession) => {
+          resolutionPath = session ? session.getResolutionPath() : undefined;
```
bajtos: I think this is showing a problem in the definition of `ResolveFunction` type. Is it a valid scenario that the resolver is called with no session argument? I though that we always create a new resolution session before calling any resolvers. There are few more places in our tests where I have to add an explicit handling of no resolution session, see below.

WDYT?

raymondfeng: Let's make `session` required. Can you open a PR to address that first?

bajtos: Sure, see https://github.com/strongloop/loopback-next/pull/2711

raymondfeng:
Now #2711 is landed, please rebase against latest master.

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: Inspired by this comment, I also created https://github.com/strongloop/loopback-next/pull/2717.

bajtos:
> Now #2711 is landed, please rebase against latest master.

Done.

I have also changed the way how I address the problem with `BindingFilter<T>` vs. `BindingFilter<unknown>`, see 90117fa57

@raymondfeng LGTY now?

packages/context/src/binding-filter.ts
```diff
@@ -9,6 +9,15 @@ import {BindingAddress} from './binding-key';
...
  * binding.
+ *
+ * Note: originally, we allow filters to be tied with a single value type.
+ * This actually does not make much sense - the filter function is typically
+ * invoked on all bindings to find those matching the given criteria,
+ * therefore filters must be prepared to handle bindings of any value type.
+ * We learned about this problem after enabling TypeScript's `strictFunctionTypes`
+ * check, but decided to preserve `ValueType` argument for backwards compatibility.
+ *
+ * TODO(semver-major): remove ValueType template argument.
```
raymondfeng: See https://github.com/strongloop/loopback-next/pull/2728. The `ValueType` is used internally with a default. Applications or extensions may define their own filter functions and mostly won't cast it to `BindingFilter<...>`. I don't think we have to release a new major.

raymondfeng:
Is there a possibility that this change will break existing applications?

packages/context/src/context-view.ts
```diff
@@ -161,10 +161,10 @@ export class ContextView<T = unknown> extends EventEmitter
  */
 export function createViewGetter<T = unknown>(
   ctx: Context,
-  bindingFilter: BindingFilter<T>,
+  bindingFilter: BindingFilter,
```
bajtos: Before this change, it was possible to infer `T` in `createViewGetter` from `bindingFilter` type.

```ts
const filter: BindingFilter<MyValue> = b => true;
const getter = createViewGetter(ctx, filter);
// ^^^ getter is of type Getter<MyValue[]>
```

With this change in place, users have to explicitly specify the getter type, otherwise they end up with `unknown`.

```ts
const getter = createViewGetter(ctx, filter);
// ^^^ getter is of type Getter<unknown[]>

const getter = createViewGetter<MyValue>(ctx, filter);
// ^^^ getter is of type Getter<MyValue[]>
```

packages/context/src/context-view.ts
```diff
@@ -161,10 +161,10 @@ export class ContextView<T = unknown> extends EventEmitter
...
   session?: ResolutionSession,
 ): Getter<T[]> {
-  const view = new ContextView(ctx, bindingFilter);
+  const view = new ContextView<T>(ctx, bindingFilter);
```
bajtos: Similarly here. Before this change, it was possible to infer `T` of `ContextView` from `bindingFilter` type. After this change, `new ContextView(...)` always treats `T` as `unknown`.

bajtos:
> Is there a possibility that this change will break existing applications?

Good question, see my two comments above.

bajtos:
Also any project using `@loopback/build` and our shared tsconfig may start failing to compile if they are not violating any of the newly enabled checks.

I don't consider this as a change that would require semver-major, because this situation happens regularly with new semver-minor releases of TypeScript. TypeScript releases a new version, e.g. 3.4, consumers upgrade to this new version (because `@loopback/build` uses `^3.3.0` as the dependency spec) automatically, and because the new version of compiler is rejecting code that was accepted by the old version, build suddenly breaks. I am arguing that build broken by a new "strict" rule enabled by `@loopback/build` is conceptually the same as when the build is broken by a new version of `tsc`.

bajtos:
I have extracted the first two commits to get them landed faster: https://github.com/strongloop/loopback-next/pull/2733

bajtos:
@raymondfeng I have rebased the pull request on top of the latest `master`, it's ready for final review & landing. LGTY?

raymondfeng:
@bajtos I landed for you to reduce the backlog of PRs.

Issue #2606, opened 2019-03-18
Spike: Declarative definition of FOREIGN KEY and UNIQUE constraints


bajtos:
To support strong relations with referential integrity enforced by the database, we need a mechanism allowing models to define FOREIGN KEY and UNIQUE constraints in a declarative way.

Let's consider the relation "Customer has one Account", using `account.customerId` as the relational property. As part of the domain model definition, we need to define the following constraints:

- FOREIGN KEY: the value stored in `account.customerId` must be an existing customer id.
- UNIQUE: there can be at most one record (Account instance) with a given customer id.

Things to consider:

- UNIQUE constraint can be composite. E.g. on GitHub, we want the combination of `(user, repo)` to be unique.

- FOREIGN KEY constraints can be composite too. See e.g. [here](https://www.ibm.com/support/knowledgecenter/en/SSGU8G_12.1.0/com.ibm.sqls.doc/ids_sqs_0535.htm):

    ```sql
    FOREIGN KEY (ref_num, ref_type) REFERENCES accounts
      (acc_num, acc_type)
    ```

- In LB3 (and loopback-datasource-juggler), we may already have partial support for foreign keys.

## Acceptance criteria

- [x] Read through juggler code-base and find out what's the current implementation.
- [x] Describe the current implementation - what works, how does it works, what is missing. (E.g. composite constraints).
- [x] Propose a solution for expressing FK/UNIQUE constraints in model/property definitions. (No decorators involved here.)
- [x] Propose a solution for defining FK/UNIQUE constraints using TypeScript decorators. Do we want to define new decorators like `@foreignKey()` and `@unique()`?
- [ ] Create a list of user stories to implement the proposals

There should be a pull request with any spike/PoC bits of code used to verify the proposals, and a `/_SPIKE_.md` file with the spike documentation (description or the current status, high-level overview of the proposals, etc.)

### Out of scope

- Implementation in our connectors - see https://github.com/strongloop/loopback-next/issues/2333 and https://github.com/strongloop/loopback-next/issues/2332

emonddr:
Timebox 2 weeks please.

bajtos:
spike pull request: https://github.com/strongloop/loopback-next/pull/2712

bajtos:
Created the following follow-up stories:

- New syntax for INDEX and FOREIGN KEY definition https://github.com/strongloop/loopback-next/issues/2766
[...]

PR #2712, opened 2019-04-09
RFC: Declarative definition of FOREIGN KEY and UNIQUE constraints

bajtos:
See #2606

To support strong relations with referential integrity enforced by the database, we need a mechanism allowing models to define FOREIGN KEY and UNIQUE constraints in a declarative way.

Please review my proposal outlined in `_SPIKE_.md`, you can also review the changes proposed in model/property definition interfaces (TypeScript types).

## Checklist

### DIFF
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -1,0 +1,209 @@
+# Declarative definition of FOREIGN KEY and UNIQUE constraints
+
+## Current status
+
+### Syntax for `index` creation, including `UNIQUE INDEX`
+
+- Not supported by connectors: Oracle, DB2, DashDB,
+- CouchDB and Cloudant support indexes, but not UNIQUE index - this is a
+  limitation of the database design.
+
+Docs: https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#indexes
+
+#### Property level
+
+- All supported connectors
+
+  ```js
+  {
+    email: {
+      type: 'string',
+      // just an index
+      index: true
+    }
+
+    email: {
+      type: 'string',
+      // UNIQUE constraint - not supported by CouchDB & Cloudant
+      index: {
+        unique: true
+      }
+    }
+  ```
+
+- MySQL, PostgreSQL, MSSQL:
+
+  ```js
+  email: {
+    type: 'string',
+    // custom kind & type
+    index: {
+      kind: 'UNIQUE' // Additional MySQL kinds: FULLTEXT, SPATIAL
+      type: 'btree', // or hash
+
+      name: 'index name', // supported by MSSQL only
+    }
+  }
+  }
+  ```
+
+- MongoDB:
+
+  ```js
+  {
+    email: {
+      type: 'string',
+
+      // unique index
+      index: true,
+      unique: true,
+    },
+
+    email: {
+      type: 'string',
+
+      index: {
+        unique: true,
+        mongodb: {
+          kind: 'text', // or 1, -1, etc.
+        }
+      }
+    },
+  ```
+
+#### Model-settings level
+
+- Common SQL flavor (supported by MySQL, PostgreSQL, MSSQL now):
+
+```js
+{
+  indexes: {
+    // a "plain" index
+    index1: {
+      // a CSV string of SQL-escaped column names,
+      // possibly including additional modifiers
+      columns: 'email, name';
+    }
+
+    // custom index
+    index2: {
+      columns: 'email, name';
+      kind: 'UNIQUE',
+      type: 'hash',
+    }
+  }
+}
+```
+
+- MYSQL extras:
+
+  ```js
+  {
+    indexes: {
+      // unique index
+      index1: {
+        columns: 'email',
+        options: {unique: true}
+      }
+
+      // also
+      index2: {
+        columns: 'email',
+        kind: 'UNIQUE',
+      }
+
+      index3: {
+        // key-value map using colum names (not property names!) as keys
+        keys: {
+          email: 1, // ASCENDING
+          updatedAt: -1, // DESCENDING
+        }
+      }
+
+      index4: {
+        columns: 'email',
+        type: 'btree' // CREATE INDEX ... USING btree
+      }
+    }
+  }
+  ```
+
+  TODO: add support for index comments, e.g. via index.docs or index.comment
+
+- PostgreSQL extras:
+
+  ```js
+  {
+    indexes: {
+      // just an index
+      index0: {
+        // list of SQL colum names, not LB property names!
+        keys: ['email', 'createdByAdmin'];
+      }
+    }
+  }
+  ```
+
+- MongoDB flavor:
+
+  ```js
+  {
+    indexes: {
+      index0: {
+        name: 'optional name to overwrite index0'
+        options: {unique: true},
+        keys: {
+          // keys are MongoDB column names, not LB property names!
+          name: 1, // ascending
+          updatedAt: -1, // descending
+          subject: 'text', // text index
+        }
+      },
+
+      index1: {
+        // keys are specified directly in the index object
+        email: 1,
+        subject: 'text',
+      }
+    }
+  }
+  ```
+
+- CouchDB, Cloudant flavor:
+
+  ```js
+  {
+    index0: {
+      keys: {
+        name: 1, // ascending
+        updatedAt: -1, // descending
+      }
+    }
+  }
+  ```
+
+### Syntax for FOREIGN KEY constraints
+
+Foreign keys are always defined at model-settings level.
+
+- Supported by: MySQL, PostgreSQL
+- Missing support in: MSSQL, Oracle
+- Not possible to implement in: MongoDB, CouchDB, Cloudant
+
+```js
+{
+  foreignKeys: {
+    keyName: {
+      name: 'constraint_name_for_db',
+      entity: 'TargetModel',
+
+      // MySQL: property name PostgreSQL: column name!
+      foreignKey: 'source property or column name',
+      // MySQL & PostgreSQL: column name!
+      entityKey: 'target column name',
+    }
+  }
+}
+```
+
+TODO: support `ON UPDATE` and `ON DELETE` options, e.g. `ON DELETE CASCADE`
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -22,28 +22,12 @@ expressing `ON UPDATE` and `ON DELETE` behavior.
 
 ### Indexes at property level
 
-Keep the current generic form as the baseline:
-
-```js
-{
-  // property definitions:
-  email: {
-    type: 'string',
-    index: {
-      name: 'email_index',
-      kind: 'UNIQUE' // Additional MySQL kinds: FULLTEXT, SPATIAL
-      type: 'btree', // or hash
-    }
-  }
-}
-```
-
-Support the following two short-hand forms:
+Support the following two short-hand forms only. Ask users to use model-level
+form to define indexes that are more complex.
 
 - a "plain" index with no special configuration
 
   ```js
-  // property definitions:
   {
     email: {
       type: 'string',
@@ -55,7 +39,6 @@ Support the following two short-hand forms:
 - UNIQUE index with no special configuration
 
   ```js
-  // property definitions:
   {
     email: {
       type: 'string',
@@ -65,7 +48,7 @@ Support the following two short-hand forms:
   ```
 
 We should also allow models to configure indexes for certain databases only,
-using any of the three forms shown above.
+using any of the forms shown above.
 
 ```js
 {
@@ -91,6 +74,8 @@ two names can be different, especially when using database-specific config to
 map property names to custom column names.
 
 The proposal:
+
+
 
 **TBD**
 
@@ -170,6 +155,8 @@ cannot process. The flag can allow three values:
 - Spike: create a template implementation of index & constraint migration in
   `Connector` and `SqlConnector`, this will be a pre-requisite for later work in
   connectors.
+- Update `examples/todo-list` to define FK and UNIQUE constraints
+- Update CLI templates for relations to define the constraints too
 
 ## Detailed description of the current syntax
 
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -2,6 +2,177 @@
 
 ## Current status
 
+Support for indexes and foreign keys depends on the connector. Some do support
+all features, some support only some of them. Depending on the database used,
+certain features are not possible to implement at all because of database design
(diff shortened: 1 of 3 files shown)


bajtos:
The spike is ready for review.

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+    nameQueries: {
+      // index definition
+    }
+  }
+}
+```
+
+Individual indexes can be defined as follows:
+
+- Add a new field `properties` as a key-value map from property names to
```
jannyHou: I don't quite follow the meaning of "Individual indexes", could you explain more about it?

I understand the example above uses LB3 syntax:

```
{
  strict: false,
  forceID: true,
  indexes: {
    uniqueEmail: {
      // index definition
    },
    nameQueries: {
      // index definition
    }
  }
}
```
While confused about the individual index...is it a syntax supported in LB3 or a proposal that will be supported?

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+connector running automigration, the connector should let the user know about
+the problem. To preserve backwards compatibility, I am proposing to print a
+console warning only.
+
+We can also introduce a new `options` flag for `autoupdate`/`automigrate` to
+control what happens when a model definition specifies metadata that the
+connector cannot process. The flag can allow three values:
+
+- true: abort migration with error
+- 'warn': print a warning and ignore unsupported metadata
```
jannyHou: sounds good, I would suggest `warn` as default.

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+
+1. Describe the new syntax in definition interfaces in
+   [`model.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts),
+   include comprehensive API documentation.
+
+   Modify `DefaultCrudRepository` constructor to process model-level indexes and
+   foreign keys; it needs to fill the corresponding fields in juggler model
+   settings.
+
+   Update `examples/todo-list` to define FK and UNIQUE constraints
```
jannyHou: also the shopping example.

_SPIKE_.md
emonddr: I have same comment as @jannyHou .

Do you mean 

```
uniqueEmail: {
      // index definition
    },
```

can be filled in like this?: 

```
uniqueEmail: {
      properties: {
           email: 1, // ASC
          createdAt: 'DESC', // alias for -1
          bio: 'text', // database-specific value (MongoDB's "text")
       }
    },
```

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+    email: {
+      type: 'string',
+      index: true,
+    }
+  }
+  ```
+
+- UNIQUE index with no special configuration
+
+  ```js
```
dhmlau: In the case of Cloudant and CouchDB, they don't have UNIQUE index as you mentioned... If I'm using Cloudant as the datasource and happen to have `unique` set. What's going to happen? We'll ignore it because it's not supported by the database?

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+      onUpdate: 'CASCADE',
+      onDelete: 'CASCADE',
+    }
+  }
+}
+```
+
+### Decorators
+
+I am proposing to reuse existing `@property` and `@model` decorators. They are
```
dhmlau: +1 on reusing existing decorators from the users' point of view.

_SPIKE_.md
dhmlau: +1 on warning to be the default.

_SPIKE_.md
jannyHou: I think the only unique field in cloudant/couchdb is the document id `_id`, other loopback flavor constrains will be ignored.

See https://forums.couchbase.com/t/can-i-do-unique-constraint-on-some-attribute-not-the-id/5621
and https://stackoverflow.com/questions/1541239/unique-constraints-in-couchdb

_SPIKE_.md
elv1s: What about relationship decorators such as `@belongsTo`, `@hasMany`, `@hasOne`. I know we recently added support to add property data to those decorators, but I don't think it is something we want to encourage.

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+  `columns` field too. This field should contain an array of column-index
+  specification, for example:
+
+  ```js
+  ['email ASC', 'lower(name) DESC'];
+  ```
+
+  In the contrary with the current implementation, I am proposing that `columns`
+  should replace any column list created from `properties` and `keys`, and that
+  `columns` should be nested inside connector-specific options:
```
elv1s: I agree with the syntax for `postgresql.columns = ['lower (email) ASC`]`. Just want to make sure what will happen if a user defines both, `properties` and `postgreqsl.columns`. 

It seems like the connector will just use whatever is defined in the connector property. I don't know how you could know programmatically that `[lower(email) ASC` was intended to replace `email`

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+    postgresql: {
+      // in PostgreSQL, use a different index definition
+      columns: ['lower(email) ASC']
+    },
+  }
+  ```
+
+### Foreign keys at property level
+
+Introduce a new property metadata "references" (inspired by ANSI SQL):
```
elv1s: I like the idea of references, but isn't this duplicating data? How can this example be applied to the TodoList example? In the todo model, we define todoListId as `belongsTo` (aka references) `TodoList`

https://github.com/strongloop/loopback-next/blob/master/examples/todo-list/src/models/todo.model.ts#L33

Maybe there is a way to leverage or combine this syntax with relational decorators?

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+  uniqueness).
+
+- Database-specific options will be stored under a key with the connector name:
+
+  ```js
+  {
+    properties: {
+      email: 'ASC',
+    },
+    mongodb: {
```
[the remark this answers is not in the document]
bajtos: I am talking about database-specific configuration of an index. Please note the code snippet we are commenting on is showing definition of an index, not a model.

```ts
@model({
  indexes: {
    demo: {
      properties: {email: 'ASC'},
      mongodb: {
        sparse: true
      },
    },
  },
})
export class MyModel extends Entity {
  @property()
  email: string;
  // ...
}
```

_SPIKE_.md
bajtos: >  I don't know how you could know programmatically that `lower(email) ASC` was intended to replace` email`.

Exactly! That's why I am proposing that `columns` should take precedence over `keys`/`properties` and replace anything defined via those key/value maps.

- If `postgresql.columns` is defined, then the connector uses `columns` and ignored `keys` & `properties`.
- If `postgresql.columns` is not defined, the connector will merge entries defined by `keys` and `properties`.

_SPIKE_.md
bajtos: > I like the idea of references, but isn't this duplicating data?  How can this example be applied to the TodoList example?

I have modified the TodoList example to show how the new syntax could be used, see the bottom half of this pull request.

I feel we are not duplicating data here, because `belongsTo` decorator can be used without referential integrity, just to navigate to related models in GraphQL style. 

Also as we have discussed in the past, we need to rethink how relation decorators are applied. At the moment, `@hasMany` is applied on the navigational property (`todoList.todos`) while `@belongsTo` is applied on the foreign key (`todo.todoListId`) - this is inconsistent & possibly confusing. The navigational property does not belong to the model base class because relational data is not part of model data, create/update operations are not able to update related data in a single call. So far, we are leaning to rework decorators to be applied on the model constructor, and let the foreign-key property be defined independently of the relation.

Having said that, I can imagine that for now, we can modify `@belongsTo` decorator to add `references` metadata to the property definition it creates. The difficult part is how to infer `keyTo` (the primary key of the target model) at the time the decorator is invoked.

https://github.com/strongloop/loopback-next/blob/cb308add60bbe938c8e85812676eb817e9e0efc9/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L26-L41

```ts
    if (!('references' in propertyDefinition)) {
      propertyDefinition.references = {
         model: targetResolver,
         property: keyTo,
      };
    }

    const propMeta: PropertyDefinition = Object.assign(
      // etc. (no change)
    );
```

_SPIKE_.md
bajtos: See my comment above. The new property-definition field `references` will be usable with the relational decorators supporting custom property definition metadata.

```ts
@model()
class Todo extends Entity {
  // ...

  @belongsTo(
    () => TodoList,
    {},
    {
      /*==== Define a foreign key to enforce referential integrity ===*/
      references: {
        model: () => TodoList,
        property: 'id',
        onDelete: 'CASCADE',
     },
    },
  )
  todoListId: number;
}
```

_SPIKE_.md
bajtos: +1

bajtos: Done in 240fe57

_SPIKE_.md
```diff
@@ -0,0 +1,543 @@
...
+    }
+  }
+  ```
+
+  Important: property names are mapped to database column names when building
+  the index definition.
+
+- Keep supporting `keys` field as a key-value map from database column name to
+  indexing order, see the description of the actual status below. Entries from
+  `keys` should me merged with entries from `properties`, `keys` taking
```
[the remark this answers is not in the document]
bajtos: Fixed in edbcca6

_SPIKE_.md
```diff
@@ -0,0 +1,549 @@
...
+  })
+  email: string;
+  ```
+
+- UNIQUE index with no special configuration
+
+  ```ts
+  @property({
+    type: 'string',
+    unique: true,
```
raymondfeng: I'm wondering if we can just support `index: {unique: true}` to avoid the `unique` keyword. Do we need to support `unique` constraint beyond indexing?

_SPIKE_.md
```diff
@@ -0,0 +1,549 @@
...
+
+Support the following two short-hand forms only. Ask users to use model-level
+form to define indexes that are more complex.
+
+- a "plain" index with no special configuration
+
+  ```ts
+  @property({
+    type: 'string',
+    index: true,
```
raymondfeng: Do we allow `index: boolean | IndexOptions`?

_SPIKE_.md
bajtos: I find the syntax `unique: true` shorter and easier to write than `index: {unique: true}`.

Also the longer form `index: {unique: true}` is opening doors for adding additional index metadata, e.g. `index: {unique: true, sparse: true}`. I am little bit concerned that it makes it easier for users to define something that's not supported.

@raymondfeng What are your concerns, why would you like to avoid adding the new keyword `unique`?

_SPIKE_.md
bajtos: See my previous answer. My intention is to keep property-level indexes simple and ask users to define a model-level index if they need anything more complex. 

Allowed all `IndexOptions` properties is opening doors to complex combinations, for example how do you propose to treat property-level index defined as follows?

```ts
class MyModel extends Entity {
  @property({
    index: {
      properties: {
         // "name" is a different property
        name: 'ASC',
      }
    },
    postgresql: {
      // custom "columns" typically replace
      // configuration provided by properties/keys 
      columns: ['lower(email) ASC'],
    }
  })
  email: string;

  @property()
  name: string;
}
```

_SPIKE_.md
```diff
@@ -0,0 +1,549 @@
...
+
+Introduce a new property metadata "references" (inspired by ANSI SQL):
+
+```ts
+@property({
+  type: 'number',
+  required: true,
+  references: {
+    // a TypeResolver
+    model: () => Category,
```
b-admike: I'm wondering why we'd like to support three ways of declaring the target model here. Can we keep this simple and support the typeResolver flavour only (at least for first iteration)?

_SPIKE_.md
b-admike: > If postgresql.columns is not defined, the connector will merge entries defined by keys and properties.

So the order is `postgresql.columns` -> `keys` -> `properties`, correct?

_SPIKE_.md
```diff
@@ -0,0 +1,549 @@
...
+  ```
+
+### Syntax for FOREIGN KEY constraints
+
+Foreign keys are always defined at model-settings level.
+
+- Supported by: MySQL, PostgreSQL
+- Missing support in: MSSQL, Oracle
+- Not possible to implement in: MongoDB, CouchDB, Cloudant
+- Not supported: `ON UPDATE` and `ON DELETE` options, e.g. `ON DELETE CASCADE`
```
b-admike: Should we add tasks for supporting these features in the connectors as part of next actions?

_SPIKE_.md
b-admike: IMO the shorthand is more convenient - at least for first iteration.

_SPIKE_.md
bajtos: My expectation is that we will create those tasks as part of "Spike: a template implementation of index & constraint migration in `SqlConnector`". I'll update capture this information in the spike acceptance criteria.

_SPIKE_.md
bajtos: Good point :+1:

_SPIKE_.md
bajtos: Yes.
If `columns` are set, then `keys` and `properties` are ignored completely.

Otherwise the content from `keys` and `properties` is merged, e.g. this will create a composite index over two columns.

```js
{
  keys: {
    name: 'ASC',
  },
  properties: {
    email: 'DESC',
  }
}
```

bajtos:
Pushed two new commits:
- address Biniam's feedback 09bc09a
- clarify handling of unsupported options 4029988

bajtos:
FWIW, we already have a community-contributed pull request adding support for `onUpdate` and `onDelete` fields for Foreign Keys, see https://github.com/strongloop/loopback-connector-mysql/pull/370

bajtos:
Created the following follow-up stories:

- New syntax for INDEX and FOREIGN KEY definition https://github.com/strongloop/loopback-next/issues/2766
- Document how to configure indexes and foreign keys https://github.com/strongloop/loopback-next/issues/2767
- Connector helpers to obtain INDEX and FOREIGN KEY configurations from model definitions https://github.com/strongloop/loopback-connector-rest/issues/134
- Spike: a template implementation of index & constraint migration in SqlConnector https://github.com/strongloop/loopback-connector-rest/issues/135

Closing this pull request as done.

PR #2722, opened 2019-04-10
fix(cli): improve openapi code generation for naming and typing

raymondfeng:
Fixing issues to generate code from OpenAPI specs from https://www.berlin-group.org/psd2-access-to-bank-accounts. Here is the [swagger doc](https://github.com/strongloop/loopback-next/files/3064625/berlin-group-swagger-2.yaml.txt).

## Checklist

### DIFF
--- a/packages/cli/generators/openapi/schema-helper.js
+++ b/packages/cli/generators/openapi/schema-helper.js
@@ -9,7 +9,7 @@ const {
   isExtension,
   titleCase,
   kebabCase,
-  escapePropertyOrMethodName,
+  escapeIdentifier,
   toJsonStr,
 } = require('./utils');
 
@@ -159,7 +159,7 @@ function mapObjectType(schema, options) {
         }),
       );
       // The property name might have chars such as `-`
-      const propName = escapePropertyOrMethodName(p);
+      const propName = escapeIdentifier(p);
 
       let propDecoration = `@property({name: '${p}'})`;
 
@@ -171,11 +171,17 @@ function mapObjectType(schema, options) {
         const itemType =
           propertyType.itemType.kind === 'class'
             ? propertyType.itemType.className
-            : getJSType(propertyType.itemType.name);
+            : // The item type can be an alias such as `export type ID = string`
+              getJSType(propertyType.itemType.declaration) ||
+              // The item type can be `string`
+              getJSType(propertyType.itemType.name);
         if (itemType) {
           // Use `@property.array` for array types
           propDecoration = `@property.array(${itemType}, {name: '${p}'})`;
-          collectImports(typeSpec, propertyType.itemType);
+          if (propertyType.itemType.className) {
+            // The referenced item type is either a class or type
+            collectImports(typeSpec, propertyType.itemType);
+          }
         }
       }
       const propSpec = {
@@ -277,7 +283,7 @@ const JSTypeMapping = {
  */
 function getJSType(type) {
   const ctor = JSTypeMapping[type];
-  return (ctor && ctor.name) || type;
+  return ctor && ctor.name;
 }
 
 /**
@@ -319,13 +325,12 @@ function mapSchemaType(schema, options) {
  * @param {object} schema
  */
 function resolveSchema(schemaMapping, schema) {
-  if (schema['x-$ref']) {
-    const resolved = schemaMapping[schema['x-$ref']];
-    if (resolved) {
-      return resolved;
-    }
+  if (!schema['x-$ref']) return schema;
+  let resolved = schema;
+  while (resolved && resolved['x-$ref']) {
+    resolved = schemaMapping[resolved['x-$ref']];
   }
-  return schema;
+  return resolved || schema;
 }
 
 /**
@@ -350,7 +355,10 @@ function generateModelSpecs(apiSpec, options) {
     // `model` is `undefined` for primitive types
     if (model == null) continue;
     if (model.className) {
-      models.push(model);
+      // The model might be a $ref
+      if (!models.includes(model)) {
+        models.push(model);
+      }
     }
   }
   return models;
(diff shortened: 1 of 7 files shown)


raymondfeng:
[berlin-group-swagger-2.yaml.txt](https://github.com/strongloop/loopback-next/files/3064625/berlin-group-swagger-2.yaml.txt)

packages/cli/generators/openapi/schema-helper.js
```diff
@@ -171,11 +171,15 @@ function mapObjectType(schema, options) {
         const itemType =
           propertyType.itemType.kind === 'class'
             ? propertyType.itemType.className
-            : getJSType(propertyType.itemType.name);
+            : getJSType(propertyType.itemType.declaration) ||
```
jannyHou: What does the `declaration` here mean?

raymondfeng: For example, for a type alias `export type IDType = string`:

- name: `IDType
- declaration: `string`

packages/cli/test/unit/openapi/schema-model.unit.js
```diff
@@ -392,17 +410,27 @@ describe('schema to model', () => {
             signature: 'addresses?: AddressList;',
             decoration: "@property.array(Address, {name: 'addresses'})",
           },
+          {
+            name: 'us-office',
+            signature:
+              'usOffice?: {\n  street?: string;\n  city?: string;\n' +
```
b-admike: Wondering if the signature can be `usOffice: USAddress;` or `usOffice: Address` here instead.

raymondfeng: Good question. I'll check.

raymondfeng: It should be fixed. I have to recursively resolve `$ref` for some models.

packages/cli/generators/openapi/schema-helper.js
```diff
@@ -171,11 +171,17 @@ function mapObjectType(schema, options) {
         const itemType =
           propertyType.itemType.kind === 'class'
             ? propertyType.itemType.className
-            : getJSType(propertyType.itemType.name);
+            : // The item type can be an alias such as `export type ID = string`
+              getJSType(propertyType.itemType.declaration) ||
+              // The item type can be `string`
+              getJSType(propertyType.itemType.name);
```
bajtos: Have you considered the following, to avoid duplication of `getJSType`?

```ts
: getJSType( 
  // The item type can be an alias such as `export type ID = string`
  propertyType.itemType.declaration ||
  // The item type can be `string`
  propertyType.itemType.name
)
```

packages/cli/test/unit/openapi/schema-model.unit.js
```diff
@@ -374,14 +385,19 @@ describe('schema to model', () => {
           },
           {
             name: 'first-name',
-            signature: "'first-name'?: string;",
+            signature: 'firstName?: string;',
```
bajtos: 👍

packages/cli/test/unit/openapi/schema-model.unit.js
```diff
@@ -374,14 +385,19 @@ describe('schema to model', () => {
...
           {
             name: 'last-name',
-            signature: "'last-name'?: Name;",
+            signature: 'lastName?: Name;',
             decoration: "@property({name: 'last-name'})",
           },
+          {
+            name: 'profiles',
+            signature: 'profiles?: ProfileId[];',
+            decoration: "@property.array(String, {name: 'profiles'})",
```
bajtos: I am wondering - why are we including `name` in the property metadata? Isn't the decorator inferring the property name automatically? I believe the property name specified in the metadata must match the property name in TypeScript source, otherwise things will break at runtime.

IMO, we should exclude `name` from the generated property metadata.

Feel free to leave such changes out of scope of this pull request.

raymondfeng: Some of the property names such as `last-name` have to be set in the `@property` decoration. I didn't bother to test if the original name is the same as the TS property.

bajtos: > Some of the property names such as `last-name` have to be set in the `@property` decoration. I didn't bother to test if the original name is the same as the TS property.

So how is this going to work in practice? Let's say the OpenAPI schema describes a property called `last-name`. This means the clients will be sending request body payload like `{"last-name": "Bajtos"}`.

At runtime, this will create object `{'last-name': 'Bajtos'}`.

To allow TypeScript code to access this data, the TypeScript property must be called `last-name` too. If it's changed e.g. to `lastName`, then 
 - TypeScript code won't be able to access the real property value.
 - Values set from TypeScript (e.g. `data.lastName = 'Feng'`) will be ignored by model's `toJSON` function, thus they won't be persisted in the database and serialized to HTTP responses.

> Some of the property names such as last-name have to be set in the @property decoration.

Are you saying that our code in incorrectly inferring model property name from the TypeScript/JavaScript property name passed by TypeScript to `@property`? I.e. that the following won't work?

```ts
@model()
class MyModel extends Entity {
  @property()
  'last-name': string;
}
``

Comment (raymondfeng) [reply packages/cli/test/unit/openapi/schema-model.unit.js]: The following code should work:
```ts
@model()
class MyModel extends Entity {
  @property()
  'last-name': string;
}
```

But we prefer to support the following:
```ts
@model('my-model')
class MyModel extends Entity {
  @property('last-name')
  lastName: string;
}
```

Mapping between JS/TS and JSON should be allowed. I'll check if our model class constructor and toJSON() supports so.

raymondfeng: See https://github.com/strongloop/loopback-next/pull/2743

PR #2728, opened 2019-04-11
feat(context): fix generic typing for BindingFilter

raymondfeng:
Fixes the `BindingFilter` generic typing as discovered by https://github.com/strongloop/loopback-next/pull/2704.

## Checklist

### DIFF
--- a/packages/context/src/binding-filter.ts
+++ b/packages/context/src/binding-filter.ts
@@ -9,9 +9,37 @@ import {BindingAddress} from './binding-key';
 /**
  * A function that filters bindings. It returns `true` to select a given
  * binding.
+ *
+ * TODO(semver-major): We might change this type in the future to either remove
+ * the `<ValueType>` or make it as type guard by asserting the matched binding
+ * to be typed with `<ValueType>`.
+ *
+ * **NOTE**: Originally, we allow filters to be tied with a single value type.
+ * This actually does not make much sense - the filter function is typically
+ * invoked on all bindings to find those ones matching the given criteria.
+ * Filters must be prepared to handle bindings of any value type. We learned
+ * about this problem after enabling TypeScript's `strictFunctionTypes` check,
+ * but decided to preserve `ValueType` argument for backwards compatibility.
+ * The `<ValueType>` represents the value type for matched bindings but it's
+ * not used for checking.
+ *
+ * Ideally, `BindingFilter` should be declared as a type guard as follows:
+ * ```ts
+ * export type BindingFilterGuard<ValueType = unknown> = (
+ *   binding: Readonly<Binding<unknown>>,
+ * ) => binding is Readonly<Binding<ValueType>>;
+ * ```
+ *
+ * But TypeScript treats the following types as incompatible and does not accept
+ * type 1 for type 2.
+ *
+ * 1. `(binding: Readonly<Binding<unknown>>) => boolean`
+ * 2. `(binding: Readonly<Binding<unknown>>) => binding is Readonly<Binding<ValueType>>`
+ *
  */
+// tslint:disable-next-line:no-unused
 export type BindingFilter<ValueType = unknown> = (
-  binding: Readonly<Binding<ValueType>>,
+  binding: Readonly<Binding<unknown>>,
 ) => boolean;
 
 /**
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -39,9 +39,9 @@ if (!Symbol.asyncIterator) {
   (Symbol as any).asyncIterator = Symbol.for('Symbol.asyncIterator');
 }
 /**
- * This import must happen after the polyfill.
- *
- * WARNING: VSCode organize import may change the order of this import
+ * WARNING: This following import must happen after the polyfill. The
+ * `auto-import` by an IDE such as VSCode may move the import before the
+ * polyfill. It must be then fixed manually.
  */
 import {iterator, multiple} from 'p-event';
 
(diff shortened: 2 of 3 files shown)


packages/context/src/binding-filter.ts
```diff
@@ -10,16 +10,28 @@ import {BindingAddress} from './binding-key';
  * A function that filters bindings. It returns `true` to select a given
  * binding.
  */
-export type BindingFilter<ValueType = unknown> = (
-  binding: Readonly<Binding<ValueType>>,
-) => boolean;
+export type BindingFilter = (binding: Readonly<Binding<unknown>>) => boolean;
```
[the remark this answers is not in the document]
bajtos: > The `ValueType` is used internally with a default. Applications or extensions may define their own filter functions and mostly won't cast it to `BindingFilter<...>`. I don't think we have to release a new major.

My concern is that if there is any user code explicitly referring to `BindingFilter<..>` type, then this code will stop working when your pull request removes the generic argument.

How about preserving the generic argument as "unused" - would that work?

```ts
export type BindingFilter<_unused = unknown> = (
  binding: Readonly<Binding<ValueType>>,
) => boolean;
```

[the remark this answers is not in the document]
raymondfeng: I decided to keep the generic type.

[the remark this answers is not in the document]
raymondfeng: Can we remove the `breaking-change` label now?

packages/context/src/binding-filter.ts
```diff
@@ -9,11 +9,36 @@ import {BindingAddress} from './binding-key';
 /**
  * A function that filters bindings. It returns `true` to select a given
  * binding.
+ *
+ * **NOTE**: We keep the generic type `T` for backward compatibility. It
+ * represents the value type for matched bindings.
```
bajtos: Would you mind copying my longer explanation from https://github.com/strongloop/loopback-next/pull/2704/files#diff-07c537de62e910016440ad9289c6f53cR12?

I would also like to add `TODO(semver-major)` comment here, to make it easy for us to find breaking changes to include when making a new semver-major release.

packages/context/src/binding-filter.ts
```diff
@@ -9,11 +9,36 @@ import {BindingAddress} from './binding-key';
...
+/**
+ * A type guard that asserts the value type for a binding if the filter function
+ * returns `true`. This type is very much the same as `BindingFilter` except
+ * that it instructs TypeScript compiler that the `binding` parameter is
+ * `Readonly<Binding<T>>` when the return value is `true`. For example:
+ *
+ * ```ts
+ * const interceptorFilter: BindingFilterGuard<Interceptor> = (
+ *   binding,
+ * ): binding is Readonly<Binding<Interceptor>> => binding.tagMap['interceptor'];
```
bajtos: I believe TypeScript is able to infer types in lambda function arguments and it should be possible to simplify this snippet as follows:

```ts
const interceptorFilter: BindingFilterGuard<Interceptor> = binding => !!binding.tagMap['interceptor'];
```

packages/context/src/binding-filter.ts
```diff
@@ -9,11 +9,36 @@ import {BindingAddress} from './binding-key';
...
+ *
+ * const myBinding: Binding<unknown> = ...;
+ * if (interceptorFilter(myBinding)) {
+ *   // Now myBinding is Readonly<Binding<Interceptor>>
+ * }
+ * ```
+ */
+export type BindingFilterGuard<T = unknown> = (
```
bajtos: AFAICT, this new function type is not accepted by any public API, it's used only internally in `Context.prototype.find` below. I am not convinced we actually need it. Would you mind reverting this addition?

packages/context/src/context.ts
```diff
@@ -522,8 +527,8 @@ export class Context extends EventEmitter {
   find<ValueType = BoundValue>(
     pattern?: string | RegExp | BindingFilter,
   ): Readonly<Binding<ValueType>>[] {
-    const bindings: Readonly<Binding>[] = [];
-    const filter = filterByKey(pattern);
+    const bindings: Readonly<Binding<ValueType>>[] = [];
+    const filter = filterByKey(pattern) as BindingFilterGuard<ValueType>;
```
bajtos: Is this type cast necessary?

packages/context/src/binding-filter.ts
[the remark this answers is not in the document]
bajtos: > I decided to keep the generic type.

👏 

> Can we remove the `breaking-change` label now?

Yes, of course!

packages/context/src/binding-filter.ts
raymondfeng: TypeScript treats `(binding) => boolean` and `(binding) => binding is Binding<T>` as incompatible types and they cannot be implicitly casted.

packages/context/src/binding-filter.ts
raymondfeng: If it's not for backward-compatibility reason, the signature for `BindingFilterGuard` is better than `BindingFilter`.

packages/context/src/binding-filter.ts
```diff
@@ -9,9 +9,33 @@ import {BindingAddress} from './binding-key';
 /**
  * A function that filters bindings. It returns `true` to select a given
  * binding.
+ *
+ * **NOTE**: Originally, we allow filters to be tied with a single value type.
+ * This actually does not make much sense - the filter function is typically
+ * invoked on all bindings to find those ones matching the given criteria.
+ * Filters must be prepared to handle bindings of any value type. We learned
+ * about this problem after enabling TypeScript's `strictFunctionTypes` check,
+ * but decided to preserve `ValueType` argument for backwards compatibility.
```
bajtos: Can you add TODO(semver-major) comment please, to make it easy for us to find breaking changes to include when making a new semver-major release.

PR #2754, opened 2019-04-15
fix: relax constrain check to allow input containing constrained values

### LINKED ISSUE #1719: relation expects foreignKey on the Object

When I was working on user order history component for https://github.com/strongloop/loopback4-example-shopping I ran into an issue ...

`/users/{userId}/orders` expects the orders object to have a `userId` field (Since it's a required property of the `Order` model and is enforced by the controller coercion / validation). The issue is that passing in the `userId` on the Order object passes controller validation but causes the relation validation to throw an error saying the foreign key can't be changed -- despite the fact that the foreign key is exactly the same as the one in the URL and the one given to the repository. 

## Current Behavior

Throws error if foreign key on object saying you can't change it even if it's the same one.

## Expected Behavior

Be smart enough to not throw an error if the key given is the same as the one that's expected. 

---

My current solution is to delete the foreign key in the controller. The other alternatives I can think of is to mark the key as an optional property (do we want this?) ... or to mark the `requestBody` type as `Partial<Order>` but this will mean losing type validation of expected properties on the Order type.

### LINKED ISSUE #2658: LB4 todo-list example gives Error: Property "todoListId" cannot  be changed!
I have cloned the project using cli "lb4 example todo-list" without modifying  anything.

After I Inserted new record to "todo-list" model, I got the response.
`{
  "id": 3,
  "title": "newToDoList",
  "color": "red"
}`

then I use id = 3 as an input for `POST /todo-lists/{id}/todos`.

```
Unhandled error in POST /todo-lists/3/todos: 500 Error: Property "todoListId" cannot be changed!
    at Object.constrainDataObject (D:\Projects\NodeJs\lbex\loopback4-example-todo-list\node_modules\@loopback\repository\dist\repositor
ies\constraint-utils.js:49:19)
    at DefaultHasManyRepository.create (D:\Projects\NodeJs\lbex\loopback4-example-todo-list\node_modules\@loopback\repository\dist\rela
tions\has-many\has-many.repository.js:21:59)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

I also try to create the relation model myself following the example.
https://loopback.io/doc/en/lb4/HasMany-relation.html
But it also give the same error.

bajtos:
Consider a Product constraint like `{categoryId: 1}` and a request
to update (patch) Product with the following data:

    {
      name: 'updated',
      categoryId: 1
    }

Before this change, such request would be incorrectly rejected.

With this change in place, such requests are allowed.

Fixes #1719, resolves #2658 

/cc @foobarrrz 

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
@@ -5,14 +5,14 @@
 
 import {expect} from '@loopback/testlab';
 import {
-  FilterBuilder,
-  Filter,
-  Where,
-  constrainFilter,
-  constrainWhere,
   constrainDataObject,
   constrainDataObjects,
+  constrainFilter,
+  constrainWhere,
   Entity,
+  Filter,
+  FilterBuilder,
+  Where,
 } from '../../..';
 
 describe('constraint utility functions', () => {
@@ -87,7 +87,7 @@ describe('constraint utility functions', () => {
       });
     });
 
-    it('throws error when the query changes field in constrain', () => {
+    it('throws error when the query changes field in constraint', () => {
       const input = new Order({id: 1, description: 'order 1'});
       const constraint: Partial<Order> = {id: 2};
       expect(() => {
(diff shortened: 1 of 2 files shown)


packages/repository/src/repositories/constraint-utils.ts
```diff
@@ -71,13 +74,7 @@ export function constrainDataObject<T extends Entity>(
...
+  constraint: DataObject<T>,
 ): DataObject<T>[] {
-  const constrainedData = cloneDeep(originalData);
-  for (let obj of constrainedData) {
-    for (let prop in constraint) {
-      (obj as AnyObject)[prop] = constraint[prop];
-    }
-  }
-  return constrainedData;
+  return originalData.map(obj => constrainDataObject(obj, constraint));
```
bajtos: This is slightly changing the behavior when the input data is trying to modify the constrained properties. Before my change, we would silently replace values provided by the user with values enforced by the constraint. With my change in place, we reject requests trying to modify constrained properties.

AFAICT, the function `constrainDataObjects` (notice the plural name!) is not used anywhere in loopback-next codebase and I would find it surprising if anybody was using it directly from 3rd party projects. I consider this change as a backwards-compatible bug fix (not a breaking change).

packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
```diff
@@ -95,16 +95,50 @@ describe('constraint utility functions', () => {
...
+        new Order({description: 'order 1'}),
+        new Order({description: 'order 2'}),
       ];
       const constraint: Partial<Order> = {id: 3};
       const result = constrainDataObjects(input, constraint);
       expect(result[0]).to.containDeep(Object.assign({}, input[0], constraint));
       expect(result[1]).to.containDeep(Object.assign({}, input[1], constraint));
     });
+
+    it('throws error when the query changes field in constrain', () => {
```
raymondfeng: Please change `constrain` to `constraint` (noun).

packages/repository/src/repositories/constraint-utils.ts
```diff
@@ -55,12 +55,15 @@ export function constrainDataObject<T extends Entity>(
 ): DataObject<T> {
   const constrainedData = cloneDeep(originalData);
   for (const c in constraint) {
-    if (constrainedData.hasOwnProperty(c))
+    if (constrainedData.hasOwnProperty(c)) {
+      if (originalData[c] === constraint[c]) continue;
```
raymondfeng: `===` won't work for objects such as `ObjectId`, but I can live with the limitation. Please add a comment.

bajtos: Good catch! I'll add a comment.

packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
bajtos: Will fix here and in the original test that I copied from too.

PR #2758, opened 2019-04-15
feat(core): extension point/extension decorators and helper functions

raymondfeng:
Add helper/decorator functions for extension point/extension pattern to `@loopback/core`.

## Checklist

### DIFF
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -349,6 +349,10 @@ children:
     url: Creating-servers.html
     output: 'web, pdf'
 
+  - title: 'Extension point and extensions'
+    url: Extension-point-and-extensions.html
+    output: 'web, pdf'
+
   - title: 'Extending request body parsing'
     url: Extending-request-body-parsing.html
     output: 'web, pdf'
--- a/packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
+++ b/packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
@@ -1,0 +1,187 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/core
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  BindingScope,
+  BINDING_METADATA_KEY,
+  Context,
+  createBindingFromClass,
+  Getter,
+  MetadataInspector,
+} from '@loopback/context';
+import {expect} from '@loopback/testlab';
+import {addExtension, extensionPoint, extensions} from '../..';
+import {CoreTags} from '../../keys';
+
+describe('extension point', () => {
+  describe('@extensionPoint', () => {
+    it('specifies name of the extension point as a binding tag', () => {
+      @extensionPoint('greeters', {scope: BindingScope.SINGLETON})
+      class GreetingService {
+        @extensions()
+        public greeters: Getter<Greeter[]>;
+      }
+
+      const bindingMetadata = MetadataInspector.getClassMetadata(
+        BINDING_METADATA_KEY,
+        GreetingService,
+      );
+
+      expect(bindingMetadata).to.not.undefined();
+      expect(bindingMetadata!.templates).to.be.an.Array();
+
+      const binding = createBindingFromClass(GreetingService);
+      expect(binding.tagMap).to.containEql({
+        [CoreTags.EXTENSION_POINT]: 'greeters',
+      });
+      expect(binding.scope).to.eql(BindingScope.SINGLETON);
+    });
+  });
+
+  describe('@extensions', () => {
+    let ctx: Context;
+
+    beforeEach(givenContext);
+
+    it('injects a getter function of extensions', async () => {
+      @extensionPoint('greeters')
+      class GreetingService {
+        @extensions()
+        public greeters: Getter<Greeter[]>;
+      }
+
+      // `@extensionPoint` is a sugar decorator for `@bind`
+      const binding = createBindingFromClass(GreetingService, {
+        key: 'greeter-service',
+      });
+      ctx.add(binding);
+      registerGreeters('greeters');
+      const greeterService = await ctx.get<GreetingService>('greeter-service');
+      const greeters = await greeterService.greeters();
+      assertGreeterExtensions(greeters);
+    });
+
+    it('injects extensions based on `name` tag of the extension point binding', async () => {
+      class GreetingService {
+        @extensions()
+        public greeters: Getter<Greeter[]>;
+      }
+      ctx
+        .bind('greeter-service')
+        .toClass(GreetingService)
+        .tag({name: 'greeters'}); // Tag the extension point with a name
+      registerGreeters('greeters');
+      const greeterService = await ctx.get<GreetingService>('greeter-service');
+      const greeters = await greeterService.greeters();
+      assertGreeterExtensions(greeters);
+    });
+
+    it('injects extensions based on class name of the extension point', async () => {
+      class GreetingService {
+        constructor(
+          @extensions()
+          public greeters: Getter<Greeter[]>,
+        ) {}
+      }
+      ctx.bind('greeter-service').toClass(GreetingService);
+      registerGreeters(GreetingService.name);
+      const greeterService = await ctx.get<GreetingService>('greeter-service');
+      const loadedGreeters = await greeterService.greeters();
+      assertGreeterExtensions(loadedGreeters);
+    });
+
+    it('injects extensions based on class name of the extension point using property', async () => {
+      class GreetingService {
+        @extensions()
+        public greeters: Getter<Greeter[]>;
+      }
+      ctx.bind('greeter-service').toClass(GreetingService);
+      registerGreeters(GreetingService.name);
+      const greeterService = await ctx.get<GreetingService>('greeter-service');
+      const greeters = await greeterService.greeters();
+      assertGreeterExtensions(greeters);
+    });
+
+    it('injects extensions based on extension point name from @extensions', async () => {
+      class GreetingService {
+        @extensions('greeters')
+        public greeters: Getter<Greeter[]>;
+      }
+      ctx.bind('greeter-service').toClass(GreetingService);
+      registerGreeters('greeters');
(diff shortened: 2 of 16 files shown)


docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -353,6 +353,10 @@ children:
     url: Extension-life-cycle.html
     output: 'web, pdf'
 
+  - title: 'Extension point and extensions'
```
bajtos: Since "Extending request body parsing" and "Extension life cycle" is building on top of "Extension point", I feel the sidebar should list "Extension point" first, before the other two pages.

packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    it('injects extensions based on the name tag of the extension point binding', async () => {
+      class GreetingService {
+        @extensions()
+        public greeters: Getter<Greeter[]>;
+      }
+      ctx
+        .bind('greeter-service')
+        .toClass(GreetingService)
+        .tag({name: 'greeters'}); // Tag the extension point with a name
```
bajtos: I find this design where the extension point is specified as binding name, as suboptimal. What if the GreetingService wanted to leverage multiple extension points?

```ts
class GreetingService {
  @extensions('greeters')
  public greeters(): Getter<Greeter[]>;

  @extensions('loggers')
  public loggers(): Getter<Logger[]>;
}
```

Also the tag name `name` is easy to misinterpret, there is no indication that `name` is a name of the extension point, as opposed to e.g. name of the class, name of the binding (the key), etc.

packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
```diff
@@ -0,0 +1,131 @@
...
+      registerGreeters('greeters');
+      const greeterService = await ctx.get<GreetingService>('greeter-service');
+      const greeters = await greeterService.greeters();
+      const languages = greeters.map(greeter => greeter.language);
+      expect(languages).to.containEql('en');
+      expect(languages).to.containEql('zh');
```
bajtos: These three lines are duplicated a lot:

```ts
const languages = greeters.map(greeter => greeter.language);
expect(languages).to.containEql('en');
expect(languages).to.containEql('zh');
```

Please extract them to a shared helper function accepting a list of expected languages. For example:

```ts
expectGreeterLanguages(greeters, ['en', 'zh']);
```

Implementation wise, please find a way how to merge the two assertions into a single one. At the moment, you are allowing the list of greeter to contain additional greeters, is that intentional? Personally, I would expect an assertion like this:

```ts
const actual = [...languages];
actual.sort()
expect(actual).to.deepEqual(['en', 'zh']);
```

packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+
+    it('injects extensions based on extension point name from @extensions', async () => {
+      class GreetingService {
+        @extensions('greeters')
```
bajtos: Please add a test to demonstrate & verify what happens when there are two `@extensions()` properties on the same class.

```ts
class GreetingService {
  @extensions('greeters')
  public greeters(): Getter<Greeter[]>;

  @extensions('loggers')
  public loggers(): Getter<Logger[]>;
}
```

packages/core/src/extension-point.ts
```diff
@@ -0,0 +1,116 @@
...
+ * the name from `@extensionPoint` or the class name of the extension point
+ * class.
+ */
+export function extensions(extensionPointName?: string) {
+  return inject('', {decorator: '@extensions'}, (ctx, injection, session) => {
+    // Find the key of the target binding
+    if (!session.currentBinding) return undefined;
+
+    if (!extensionPointName) {
+      extensionPointName = session.currentBinding.tagMap.name;
```
bajtos: As I commented earlier, I find it confusing to use tag called `name` for the extension point name. Let's use a more descriptive tag name please, e.g. `extensionPointName` or simply `extensionPoint`.

packages/core/src/extension-point.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @param extensionPointName Name of the extension point. If not supplied, we use
+ * the name from `@extensionPoint` or the class name of the extension point
+ * class.
+ */
+export function extensions(extensionPointName?: string) {
+  return inject('', {decorator: '@extensions'}, (ctx, injection, session) => {
+    // Find the key of the target binding
+    if (!session.currentBinding) return undefined;
+
+    if (!extensionPointName) {
```
bajtos: The details of extension point name inference are making the implementation of the resolver function too complex. Please extract  this code into a new function, e.g.

```ts
if (!extensionPointName) {
  extensionPointName = inferExtensionPointName(injection, session);
}
```

packages/core/src/extension-point.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+}
+
+/**
+ * A factory function to create binding template for extensions of the given
+ * extension point
+ * @param extensionPointName Name of the extension point
+ */
+export function extensionFor(extensionPointName: string): BindingTemplate {
+  return binding => binding.tag({extensionPoint: extensionPointName});
```
bajtos: Uh oh, so an extension is decorated with a tag called `extensionPoint` which means that the decorated class is not an extension point itself, but it's an extension for the specified `extensionPoint`. I find it difficult to get my head around this naming convention. Could you please try to find a better one? Perhaps something along the lines of `targetExtensionPoint` or `extensionFor`?

packages/core/src/index.ts
```diff
@@ -3,13 +3,13 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-export {Component, ProviderMap} from './component';
-export * from './server';
+// Re-export public Core API coming from dependencies
+export * from '@loopback/context';
+// Export APIs
```
bajtos: Please add an empty line to visually delimit the block "Re-export..." from the block "Export APIs"

packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
bajtos: In fact, the tag `name` is already defined as "artifact name":

https://github.com/strongloop/loopback-next/blob/c774ed1e7b0168029ca8ec63de5867b475b26e0f/packages/context/src/keys.ts#L18-L21

Is the concept of "artifact name" the same as "extension point name"? They look like two different things to me.

hacksparrow:
@raymondfeng how do we start the `greeter-extension` app? There is no start script in `package.json` and no mention of how to start the app in the `README.md` file.

hacksparrow:
Tests are failing.

```
  1) extension point
       @extensionPoint
         specifies name of the extension point:

      AssertionError: expected Object {
  templates: Array [ Function { name: '' }, Function { name: '' } ],
  target: Function { name: 'GreetingService' }
} to equal Object { tags: Object { name: 'greeters' } }
      + expected - actual

       {
      -  "target": [Function]
      -  "templates": [
      -    [Function]
      -    [Function]
      -  ]
      +  "tags": {
      +    "name": "greeters"
      +  }
       }

      at Assertion.fail (packages/testlab/node_modules/should/as-function.js:275:17)
      at Assertion.value (packages/testlab/node_modules/should/as-function.js:356:19)
      at Context.it (packages/core/src/__tests__/acceptance/extension-point.acceptance.ts:29:12)

  2) extension point
       @extensions
         injects a getter function of extensions:
     AssertionError: expected Array [] to contain 'en'
      at Assertion.fail (packages/testlab/node_modules/should/as-function.js:275:17)
      at Assertion.value [as containEql] (packages/testlab/node_modules/should/as-function.js:356:19)
      at Context.it (packages/core/src/__tests__/acceptance/extension-point.acceptance.ts:50:28)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

packages/core/src/__tests__/acceptance/extension-point.acceptance.ts
raymondfeng: I now use `extensionPoint` tag for the name of an extension point. It falls back to `name` tag or the class name.

packages/core/src/extension-point.ts
raymondfeng: `extensionFor` is now used.

raymondfeng:
> I see why the app is not working. It is not an LB4 app!

Fixed.

hacksparrow:
@raymondfeng I am a little confused about the use case of this feature. Kindly enlighten :D

My impression was that extension points would be something like hooks, with some sort of automatic triggers. Looks like it is not.

Also:

```
  async main() {
    const greetingService = await this.getGreetingService();
    let msg = await greetingService.greet('en', 'Raymond');
    console.log('English:', msg);
    msg = await greetingService.greet('zh', 'Raymond');
    console.log('Chinese:', msg);
  }
```

could be implemented with a simple function as well. Why the added complexity?

raymondfeng:
> I am a little confused about the use case of this feature

You should read https://github.com/strongloop/loopback-next/blob/greeter-extension/examples/greeter-extension/README.md first to understand the use cases.

raymondfeng:
>
```ts 
async main() {
    const greetingService = await this.getGreetingService();
    let msg = await greetingService.greet('en', 'Raymond');
    console.log('English:', msg);
    msg = await greetingService.greet('zh', 'Raymond');
    console.log('Chinese:', msg);
  }
```

The magic is in the GreetingService class which is an extension point to accept greeters for different languages. From an application perspective, it is a service that can greet in various languages. The extensions are decoupled and plugged in. In this case, `GreetingService` represents the business logic to achieve greetings and it opens the door for extensions.

raymondfeng:
@hacksparrow Please note that extension point/extension pattern is often used for the framework to implement hooks, observers, booters, and other building blocks. But application services can benefit from it too. We still need to refactor existing code in LB4 to be consistently use the sugar functions and conventions to improve our extensibility story for various places:

1. RestServer -> Controllers
2. RestServer -> Body parsers
3. Bootstrapper -> Booters

We use the technique to implement life cycle observers, and interceptors.

hacksparrow:
It finally dawned on me that this feature is more for us than our users. Making sense now.

PR #2746, opened 2019-04-12
feat: refactor auth action

jannyHou:
### Description

connect to https://github.com/strongloop/loopback-next/issues/2467

This PR starts to refactor the authentication action to takes in a getter for the newly created AuthenticationStrategy interface in https://github.com/strongloop/loopback-next/pull/2688.

- The new strategy resolver should be agnostic of a strategy is passport based or not.
- The `StrategyAdapter` is also refactored to implement the `AuthenticationStrategy` interface
- Having the PR as a draft to prove the interface in https://github.com/strongloop/loopback-next/pull/2688 is good enough to describe a universal auth strategy. Will turn it to a formal one to finish story #2467 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

Issue #2311, opened 2019-01-30
Creating adapter for the plugged in passport strategies


jannyHou:
## Description / Steps to reproduce / Feature proposal

We should allow users to plugin custom passport strategies like `passport-jwt`. I tried replacing our own implementation of `jwt` strategy with the community one `passport-jwt` in PR https://github.com/strongloop/loopback4-example-shopping/commit/3cbb0a8b8d41a6db5c5a9ecafa78b3aef6b1aad3 to verify it works with our current `@loopback/authentication` module, the missing part is make the strategy registration flexible.

## Acceptance Criteria

- [x] Allow developers to register their own passport strategy
- [ ] ~A reference for the implementation would be how we register a custom body parser in the application, see [this test case](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/__tests__/acceptance/request-parsing/request-parsing.acceptance.ts#L47-L64)~ we have the extension point now 🎉 
- [x] Please note that this extension point only open for [`passport`](https://github.com/jaredhanson/passport) based strategies, which means when create the binding key/resolve the strategy name, make sure it's for `passport` particularly.
- [x] The story owner could explorer different approaches other than the way we register custom body parser.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

bajtos:
As I commented in https://github.com/strongloop/loopback-next/issues/2312#issuecomment-469598086, I think our support for Passport strategies should be built on top of a LoopBack-specific extension point described by #2312.

That means we need to implement #2312 before working on this story.

bajtos:
I think this story depends on #2466: Add abstraction for authentication strategy.

emonddr:
Some of this work seems to be occurring in the PR for https://github.com/strongloop/loopback-next/issues/2467.

jannyHou:
> Some of this work seems to be occurring in the PR for #2467.

+1.

If we agree on the rewrite of adapter in the draft PR https://github.com/strongloop/loopback-next/pull/2688, then this story can be closed along with #2467 after the formal version of #2688 get merged

jannyHou:
The new adapter is published as @loopback/authentication-passport@0.1.0, an experimental package 🎉 https://www.npmjs.com/package/@loopback/authentication-passport

We've tried 2 approaches:
- wrapping a passport strategy module like `passport-http`
- wrapping the express middleware `passport`

[...]

Issue #2246, opened 2019-01-14
[Spike]Make model `UserProfile` more flexible


jannyHou:
## Description / Steps to reproduce / Feature proposal

See the discussion in [comment](), the `UserProfile` interface in `@loopback/authentication` should be more flexible to allow user defined custom properties.

## Current Behavior

Other than the required `id` property, there are 2 optional properties `name` and `email`, while not all the models being authenticated have them. For example, the [user](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/models/user.model.ts) model has `firstName` and `secondName` instead of `name`. 

## Expected Behavior

Other than the required `id` property, `UserProfile` should honour custom properties defined in the authenticated model.

## Acceptance Criteria

- [ ] Explorer the best approach to allow developers provide additional/custom properties in `UserProfile`
- [ ] Since the [`authenticate`](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/authentication.provider.ts#L54) function and the [authenticated method](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/controllers/user.controller.ts#L105) return a user profile instance, the solution should make generating OpenAPI schema for the User model easy enough.
- [ ] Potential suggestions:
  - [ ] Turn `UserProfile` into a LoopBack 4 model
  - [ ] Custom User implements `UserProfile`
  - [ ] Others(for the story owner to explorer more)

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

osmanmesutozcan:
It feels to me `@loopback/authentication` should not even define a `UserProfile` interface. It should let the user define their own interfaces and promote that in usage example. It is simply not possible to guess what every user would like to have in their user profile.

Or maybe I am missing something ?

jannyHou:
@osmanmesutozcan 
The reason why I would like to have a built-in UserProfile model/interface/type is this would simplify the authorization system we are going to build.

While I also understand your concern:

> It feels to me @loopback/authentication should not even define a UserProfile interface. It should let the user define their own interfaces and promote that in usage example. It is simply not possible to guess what every user would like to have in their user profile.

The idea of turning `UserProfile` to a class is to generate the corresponding OpenAPI schema easier. I had a PR on local(haven't published to remote) to create the class and also have the same feeling as yours.

We should explorer more approaches in this issue.

jannyHou:
[...]

PR #2763, opened 2019-04-17
feat: resolve authentication strategy registered via extension point

emonddr:
Resolve authentication strategies registered via extension point

Part of #2312

## Checklist

### DIFF
--- a/packages/authentication/docs/authentication-system.md
+++ b/packages/authentication/docs/authentication-system.md
@@ -163,3 +163,199 @@ And the abstractions for:
     - return user
   - controller function:
     - process the injected user
+
+## Registering an authentication strategy via an extension point
+
+Authentication strategies register themselves to an authentication strategy
+provider using an
+[Extension Point and Extensions](https://loopback.io/doc/en/lb4/Extension-point-and-extensions.html)
+pattern.
+
+The `AuthenticationStrategyProvider` class in
+`src/providers/auth-strategy.provider.ts` (shown below) declares an
+`extension point` named
+`AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME` via the
+`@extensionPoint` decorator. The binding scope is set to **transient** because
+an authentication strategy **may** differ with each request.
+`AuthenticationStrategyProvider` is responsible for finding (with the aid of the
+`@extensions()` **getter** decorator) and returning an authentication strategy
+which has a specific **name** and has been registered as an **extension** of the
+aforementioned **extension point**.
+
+```ts
+@extensionPoint(
+  AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  {scope: BindingScope.TRANSIENT},
+)
+export class AuthenticationStrategyProvider
+  implements Provider<AuthenticationStrategy | undefined> {
+  constructor(
+    @extensions()
+    private authenticationStrategies: Getter<AuthenticationStrategy[]>,
+    @inject(AuthenticationBindings.METADATA)
+    private metadata?: AuthenticationMetadata,
+  ) {}
+  async value(): Promise<AuthenticationStrategy | undefined> {
+    if (!this.metadata) {
+      return undefined;
+    }
+    const name = this.metadata.strategy;
+    const strategy = await this.findAuthenticationStrategy(name);
+    if (!strategy) {
+      // important not to throw a non-protocol-specific error here
+      let error = new Error(`The strategy '${name}' is not available.`);
+      Object.assign(error, {
+        code: AUTHENTICATION_STRATEGY_NOT_FOUND,
+      });
+      throw error;
+    }
+    return strategy;
+  }
+
+  async findAuthenticationStrategy(name: string) {
+    const strategies = await this.authenticationStrategies();
+    const matchingAuthStrategy = strategies.find(a => a.name === name);
+    return matchingAuthStrategy;
+  }
+}
+```
+
+The **name** of the strategy is specified in the `@authenticate` decorator that
+is added to a controller method when authentication is desired for a specific
+endpoint.
+
+```ts
+    class UserController {
+      constructor() {}
+      @get('/whoAmI')
+      @authenticate('basic')
+      whoAmI()
+      {
+        ...
+      }
+    }
+```
+
+An authentication strategy must implement the `AuthenticationStrategy` interface
+defined in `src/types.ts`.
+
+```ts
+export interface BasicAuthenticationStrategyCredentials {
+  email: string;
+  password: string;
+}
+
+export class BasicAuthenticationStrategy implements AuthenticationStrategy {
+  name: string = 'basic';
+
+  constructor(
+    @inject(BasicAuthenticationStrategyBindings.USER_SERVICE)
+    private userService: BasicAuthenticationUserService,
+  ) {}
+
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    const credentials: BasicAuthenticationStrategyCredentials = this.extractCredentals(
+      request,
+    );
+    const user = await this.userService.verifyCredentials(credentials);
+    const userProfile = this.userService.convertToUserProfile(user);
+
+    return userProfile;
+  }
+```
+
+A custom sequence must be created to insert the
+`AuthenticationBindings.AUTH_ACTION` action. The `AuthenticateFn` function
+interface is implemented by the `value()` function of
+`AuthenticateActionProvider` class in `/src/providers/auth-action.provider.ts`.
+
+```ts
+class SequenceIncludingAuthentication implements SequenceHandler {
+  constructor(
+    @inject(SequenceActions.FIND_ROUTE) protected findRoute: FindRoute,
+    @inject(SequenceActions.PARSE_PARAMS)
+    protected parseParams: ParseParams,
+    @inject(SequenceActions.INVOKE_METHOD) protected invoke: InvokeMethod,
+    @inject(SequenceActions.SEND) protected send: Send,
+    @inject(SequenceActions.REJECT) protected reject: Reject,
+    @inject(AuthenticationBindings.AUTH_ACTION)
+    protected authenticateRequest: AuthenticateFn,
+  ) {}
+
+  async handle(context: RequestContext) {
+    try {
+      const {request, response} = context;
+      const route = this.findRoute(request);
+
+      //
+      // The authentication action utilizes a strategy resolver to find
+      // an authentication strategy by name, and then it calls
+      // strategy.authenticate(request).
+      //
+      // The strategy resolver throws a non-http error if it cannot
+      // resolve the strategy. When the strategy resolver obtains
+      // a strategy, it calls strategy.authentication(request) which
+      // is expected to return a user profile. If the user profile
+      // is undefined, then it throws a non-http error.
+      //
+      // It is necessary to catch these errors
+      // and rethrow them as http errors (in our REST application example)
+      //
+      // Errors thrown by the strategy implementations are http errors
+      // (in our REST application example). We simply rethrow them.
+      //
+      try {
+        //call authentication action
+        await this.authenticateRequest(request);
+      } catch (e) {
+        // strategy not found error, or user profile undefined
+        if (
+          e.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
+          e.code === USER_PROFILE_NOT_FOUND
+        ) {
+          throw new HttpErrors.Unauthorized(e.message);
+        } else {
+          // strategy error
+          throw e;
+        }
+      }
+
+      // Authentication successful, proceed to invoke controller
+      const args = await this.parseParams(request, route);
+      const result = await this.invoke(route, args);
+      this.send(response, result);
+    } catch (error) {
+      this.reject(context, error);
+      return;
+    }
+  }
+}
+```
+
+Then custom sequence must be bound to the application, and the authentication
+strategy must be added as an **extension** of the **extension point** using the
+`addExtension` function.
+
+```ts
+export class MyApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options?: ApplicationConfig) {
+    super(options);
+
+    this.component(AuthenticationComponent);
+
+    this.sequence(SequenceIncludingAuthentication);
+
+    addExtension(
+      this,
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      BasicAuthenticationStrategy,
+      {
+        namespace:
+          AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      },
+    );
+  }
+}
+```
--- a/packages/authentication/package.json
+++ b/packages/authentication/package.json
@@ -36,7 +36,8 @@
     "@types/node": "^10.11.2",
     "@types/passport": "^1.0.0",
     "@types/passport-http": "^0.3.6",
-    "passport-http": "^0.3.0"
+    "passport-http": "^0.3.0",
+    "jsonwebtoken": "^8.5.1"
   },
   "keywords": [
     "LoopBack",
--- a/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
@@ -1,0 +1,251 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {inject} from '@loopback/context';
+import {Application} from '@loopback/core';
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {api, get} from '@loopback/openapi-v3';
+import {Request, RestServer} from '@loopback/rest';
+import {Client, createClientForHandler} from '@loopback/testlab';
+import {
+  authenticate,
+  AuthenticationBindings,
+  registerAuthenticationStrategy,
+  UserProfile,
+} from '../..';
+import {AuthenticationStrategy} from '../../types';
+import {
+  createBasicAuthorizationHeaderValue,
+  getApp,
+  getUserRepository,
+} from '../fixtures/helper';
+import {BasicAuthenticationStrategyBindings, USER_REPO} from '../fixtures/keys';
+import {MyAuthenticationSequence} from '../fixtures/sequences/authentication.sequence';
+import {BasicAuthenticationUserService} from '../fixtures/services/basic-auth-user-service';
+import {BasicAuthenticationStrategy} from '../fixtures/strategies/basic-strategy';
+import {User} from '../fixtures/users/user';
+import {UserRepository} from '../fixtures/users/user.repository';
+
+describe('Basic Authentication', () => {
+  let app: Application;
+  let server: RestServer;
+  let users: UserRepository;
+  let joeUser: User;
+  beforeEach(givenAServer);
+  beforeEach(givenControllerInApp);
+  beforeEach(givenAuthenticatedSequence);
+  beforeEach(givenProviders);
+
+  it(`authenticates successfully for correct credentials of user 'jack'`, async () => {
+    const client = whenIMakeRequestTo(server);
+    await client
+      .get('/whoAmI')
+      .set('Authorization', createBasicAuthorizationHeaderValue(joeUser))
+      .expect(joeUser.id);
+  });
+
+  it('returns error for missing Authorization header', async () => {
+    const client = whenIMakeRequestTo(server);
+
+    await client.get('/whoAmI').expect({
+      error: {
+        message: 'Authorization header not found.',
+        name: 'UnauthorizedError',
+        statusCode: 401,
+      },
+    });
+  });
+
+  it(`returns error for missing 'Basic ' portion of Authorization header value`, async () => {
+    const client = whenIMakeRequestTo(server);
+    await client
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBasicAuthorizationHeaderValue(joeUser, {prefix: 'NotB@sic '}),
+      )
+      .expect({
+        error: {
+          message: `Authorization header is not of type 'Basic'.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it(`returns error for too many parts in Authorization header value`, async () => {
+    const client = whenIMakeRequestTo(server);
+    await client
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBasicAuthorizationHeaderValue(joeUser) + ' someOtherValue',
+      )
+      .expect({
+        error: {
+          message: `Authorization header value has too many parts. It must follow the pattern: 'Basic xxyyzz' where xxyyzz is a base64 string.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it(`returns error for missing ':' in decrypted Authorization header credentials value`, async () => {
+    const client = whenIMakeRequestTo(server);
+    await client
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBasicAuthorizationHeaderValue(joeUser, {separator: '|'}),
+      )
+      .expect({
+        error: {
+          message: `Authorization header 'Basic' value does not contain two parts separated by ':'.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it(`returns error for too many parts in decrypted Authorization header credentials value`, async () => {
+    const client = whenIMakeRequestTo(server);
+    await client
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBasicAuthorizationHeaderValue(joeUser, {
+          extraSegment: 'extraPart',
+        }),
+      )
+      .expect({
+        error: {
+          message: `Authorization header 'Basic' value does not contain two parts separated by ':'.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('allows anonymous requests to methods with no decorator', async () => {
+    class InfoController {
+      @get('/status')
+      status() {
+        return {running: true};
+      }
+    }
+
+    app.controller(InfoController);
+    await whenIMakeRequestTo(server)
+      .get('/status')
+      .expect(200, {running: true});
+  });
+
+  it('returns error for unknown authentication strategy', async () => {
+    class InfoController {
+      @get('/status')
+      @authenticate('doesnotexist')
+      status() {
+        return {running: true};
+      }
+    }
+
+    app.controller(InfoController);
+    await whenIMakeRequestTo(server)
+      .get('/status')
+      .expect({
+        error: {
+          message: `The strategy 'doesnotexist' is not available.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('returns error when undefined user profile returned from authentication strategy', async () => {
+    class BadBasicStrategy implements AuthenticationStrategy {
+      name: string = 'badbasic';
+      async authenticate(request: Request): Promise<UserProfile | undefined> {
+        return undefined;
+      }
+    }
+    registerAuthenticationStrategy(server, BadBasicStrategy);
+
+    class InfoController {
+      @get('/status')
+      @authenticate('badbasic')
+      status() {
+        return {running: true};
+      }
+    }
+
+    app.controller(InfoController);
+    await whenIMakeRequestTo(server)
+      .get('/status')
+      .expect({
+        error: {
+          message: `User profile not returned from strategy's authenticate function`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  async function givenAServer() {
+    app = getApp();
+    server = await app.getServer(RestServer);
+  }
+
+  function givenControllerInApp() {
+    const apispec = anOpenApiSpec()
+      .withOperation('get', '/whoAmI', {
+        'x-operation-name': 'whoAmI',
+        responses: {
+          '200': {
+            description: '',
+            schema: {
+              type: 'string',
+            },
+          },
+        },
+      })
+      .build();
+
+    @api(apispec)
+    class MyController {
+      constructor() {}
+
+      @authenticate('basic')
+      async whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ): Promise<string> {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+    app.controller(MyController);
+  }
+
+  function givenAuthenticatedSequence() {
+    // bind user defined sequence
+    server.sequence(MyAuthenticationSequence);
+  }
+
+  function givenProviders() {
+    registerAuthenticationStrategy(server, BasicAuthenticationStrategy);
+
+    server
+      .bind(BasicAuthenticationStrategyBindings.USER_SERVICE)
+      .toClass(BasicAuthenticationUserService);
+
+    users = getUserRepository();
+    joeUser = users.list['joe888'];
+    server.bind(USER_REPO).to(users);
+  }
+
+  function whenIMakeRequestTo(restServer: RestServer): Client {
+    return createClientForHandler(restServer.requestHandler);
+  }
+});
--- a/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
@@ -3,37 +3,34 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {inject} from '@loopback/context';
 import {Application} from '@loopback/core';
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {api, get} from '@loopback/openapi-v3';
 import {
-  RestBindings,
-  ParseParams,
   FindRoute,
   InvokeMethod,
-  Send,
+  ParseParams,
   Reject,
-  SequenceHandler,
-  RestServer,
-  RestComponent,
   RequestContext,
+  RestBindings,
+  RestComponent,
+  RestServer,
+  Send,
+  SequenceHandler,
 } from '@loopback/rest';
-import {api, get} from '@loopback/openapi-v3';
 import {Client, createClientForHandler} from '@loopback/testlab';
-import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
-import {inject, Provider, ValueOrPromise} from '@loopback/context';
 import {
   authenticate,
-  UserProfile,
-  AuthenticationBindings,
   AuthenticateFn,
-  AuthenticationMetadata,
+  AuthenticationBindings,
   AuthenticationComponent,
+  UserProfile,
 } from '../..';
-import {Strategy} from 'passport';
-import {BasicStrategy} from 'passport-http';
 
 const SequenceActions = RestBindings.SequenceActions;
 
-describe('Basic Authentication', () => {
+describe.skip('Basic Authentication', () => {
   let app: Application;
   let server: RestServer;
   let users: UserRepository;
@@ -159,32 +156,34 @@ describe('Basic Authentication', () => {
   }
 
   function givenProviders() {
-    class MyPassportStrategyProvider implements Provider<Strategy | undefined> {
-      constructor(
-        @inject(AuthenticationBindings.METADATA)
-        private metadata: AuthenticationMetadata,
-      ) {}
-      value(): ValueOrPromise<Strategy | undefined> {
-        if (!this.metadata) {
-          return undefined;
+    /**
+        class MyPassportStrategyProvider implements Provider<Strategy | undefined> {
+          constructor(
+            @inject(AuthenticationBindings.METADATA)
+            private metadata: AuthenticationMetadata,
+          ) {}
+          value(): ValueOrPromise<Strategy | undefined> {
+            if (!this.metadata) {
+              return undefined;
+            }
+            const name = this.metadata.strategy;
+            if (name === 'BasicStrategy') {
+              return new BasicStrategy(this.verify);
+            } else {
+              return Promise.reject(`The strategy ${name} is not available.`);
+            }
+          }
+          // callback method for BasicStrategy
+          verify(username: string, password: string, cb: Function) {
+            process.nextTick(() => {
+              users.find(username, password, cb);
+            });
+          }
         }
-        const name = this.metadata.strategy;
-        if (name === 'BasicStrategy') {
-          return new BasicStrategy(this.verify);
-        } else {
-          return Promise.reject(`The strategy ${name} is not available.`);
-        }
-      }
-      // callback method for BasicStrategy
-      verify(username: string, password: string, cb: Function) {
-        process.nextTick(() => {
-          users.find(username, password, cb);
-        });
-      }
-    }
-    server
-      .bind(AuthenticationBindings.STRATEGY)
-      .toProvider(MyPassportStrategyProvider);
+        server
+          .bind(AuthenticationBindings.STRATEGY)
+          .toProvider(MyPassportStrategyProvider);
+    */
   }
 
   function whenIMakeRequestTo(restServer: RestServer): Client {
--- a/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
@@ -1,0 +1,471 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {inject} from '@loopback/context';
+import {Application} from '@loopback/core';
+import {get, post} from '@loopback/openapi-v3';
+import {Request, RestServer} from '@loopback/rest';
+import {Client, createClientForHandler, expect} from '@loopback/testlab';
+import {
+  authenticate,
+  AuthenticationBindings,
+  AuthenticationStrategy,
+  registerAuthenticationStrategy,
+  UserProfile,
+} from '../..';
+import {
+  createBearerAuthorizationHeaderValue,
+  createUserProfile,
+  getApp,
+  getUserRepository,
+} from '../fixtures/helper';
+import {JWTAuthenticationStrategyBindings, USER_REPO} from '../fixtures/keys';
+import {MyAuthenticationSequence} from '../fixtures/sequences/authentication.sequence';
+import {JWTService} from '../fixtures/services/jwt-service';
+import {JWTAuthenticationStrategy} from '../fixtures/strategies/jwt-strategy';
+import {User} from '../fixtures/users/user';
+import {UserRepository} from '../fixtures/users/user.repository';
+
+describe('JWT Authentication', () => {
+  let app: Application;
+  let server: RestServer;
+  let testUsers: UserRepository;
+  let joeUser: User;
+  let token: string;
+
+  beforeEach(givenAServer);
+  beforeEach(givenAuthenticatedSequence);
+  beforeEach(givenProviders);
+
+  it('authenticates successfully with valid token', async () => {
+    class InfoController {
+      constructor(
+        @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+        public tokenService: JWTService,
+        @inject(USER_REPO)
+        public users: UserRepository,
+      ) {}
+
+      @post('/login')
+      async logIn() {
+        //
+        // ...Other code for verifying a valid user (e.g. basic or local strategy)...
+        //
+
+        // Now with a valid userProfile, let's create a JSON web token
+        return await this.tokenService.generateToken(
+          createUserProfile(joeUser),
+        );
+      }
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    token = (await whenIMakeRequestTo(server)
+      .post('/login')
+      .expect(200)).text;
+
+    expect(token).to.be.not.null();
+    expect(token).to.be.String();
+
+    const id = (await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set('Authorization', createBearerAuthorizationHeaderValue(token))
+      .expect(200)).text;
+
+    expect(id).to.equal(joeUser.id);
+  });
+
+  it(`returns error for missing Authorization header`, async () => {
+    class InfoController {
+      constructor(
+        @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+        public tokenService: JWTService,
+        @inject(USER_REPO)
+        public users: UserRepository,
+      ) {}
+
+      @post('/login')
+      async logIn() {
+        //
+        // ...Other code for verifying a valid user (e.g. basic or local strategy)...
+        //
+
+        // Now with a valid userProfile, let's create a JSON web token
+        return await this.tokenService.generateToken(
+          createUserProfile(joeUser),
+        );
+      }
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    token = (await whenIMakeRequestTo(server)
+      .post('/login')
+      .expect(200)).text;
+
+    expect(token).to.be.not.null();
+    expect(token).to.be.String();
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .expect({
+        error: {
+          message: 'Authorization header not found.',
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it(`returns error for invalid 'Bearer ' portion of Authorization header value`, async () => {
+    class InfoController {
+      constructor(
+        @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+        public tokenService: JWTService,
+        @inject(USER_REPO)
+        public users: UserRepository,
+      ) {}
+
+      @post('/login')
+      async logIn() {
+        //
+        // ...Other code for verifying a valid user (e.g. basic or local strategy)...
+        //
+
+        // Now with a valid userProfile, let's create a JSON web token
+        return await this.tokenService.generateToken(
+          createUserProfile(joeUser),
+        );
+      }
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    token = (await whenIMakeRequestTo(server)
+      .post('/login')
+      .expect(200)).text;
+
+    expect(token).to.be.not.null();
+    expect(token).to.be.String();
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBearerAuthorizationHeaderValue(token, 'NotB3ar3r '),
+      )
+      .expect({
+        error: {
+          message: `Authorization header is not of type 'Bearer'.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it(`returns error for too many parts in Authorization header value`, async () => {
+    class InfoController {
+      constructor(
+        @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+        public tokenService: JWTService,
+        @inject(USER_REPO)
+        public users: UserRepository,
+      ) {}
+
+      @post('/login')
+      async logIn() {
+        //
+        // ...Other code for verifying a valid user (e.g. basic or local strategy)...
+        //
+
+        return await this.tokenService.generateToken(
+          createUserProfile(joeUser),
+        );
+      }
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    token = (await whenIMakeRequestTo(server)
+      .post('/login')
+      .expect(200)).text;
+
+    expect(token).to.be.not.null();
+    expect(token).to.be.String();
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set(
+        'Authorization',
+        createBearerAuthorizationHeaderValue(token) + ' someOtherValue',
+      )
+      .expect({
+        error: {
+          message: `Authorization header value has too many parts. It must follow the pattern: 'Bearer xx.yy.zz' where xx.yy.zz is a valid JWT token.`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('returns error due to expired token', async () => {
+    class InfoController {
+      constructor() {}
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    const expiredToken =
+      'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjEiLCJlbWFpbCI6ImpvZUBleGFtcGxlLmNvbSIsIm5hbWUiOiJqb2Ugam9lbWFuIiwiaWF0IjoxNTU1ODY3NDAzLCJleHAiOjE1NTU4Njc0NjN9.QKmO5qDC8Yg-aK3EedLRsXczL7VQDDnWtA-cpyqszqM';
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set('Authorization', createBearerAuthorizationHeaderValue(expiredToken))
+      .expect({
+        error: {
+          message: `Error verifying token : jwt expired`,
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('returns error due to invalid token #1', async () => {
+    class InfoController {
+      constructor() {}
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    const invalidToken = 'aaa.bbb.ccc';
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set('Authorization', createBearerAuthorizationHeaderValue(invalidToken))
+      .expect({
+        error: {
+          message: 'Error verifying token : invalid token',
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('returns error due to invalid token #2', async () => {
+    class InfoController {
+      constructor() {}
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (!userProfile) return 'userProfile is undefined';
+        if (!userProfile.id) return 'userProfile id is undefined';
+        return userProfile.id;
+      }
+    }
+
+    app.controller(InfoController);
+
+    const invalidToken = 'aaa.bbb.ccc.ddd';
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set('Authorization', createBearerAuthorizationHeaderValue(invalidToken))
+      .expect({
+        error: {
+          message: 'Error verifying token : jwt malformed',
+          name: 'UnauthorizedError',
+          statusCode: 401,
+        },
+      });
+  });
+
+  it('creates a json web token and throws error for userProfle that is undefined', async () => {
+    class InfoController {
+      constructor(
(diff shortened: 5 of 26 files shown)


emonddr:
First commit tackles mock `Basic` authentication strategy class which implements `AuthenticationStrategy` interface, and which registers itself as an extension of the authentication strategy extension point.

packages/authentication/src/__tests__/acceptance/authentication-action-provider/authentication-action.provider.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback4-example-shopping
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Getter, inject, Provider, Setter} from '@loopback/context';
```
jannyHou: You can rewrite the [provider](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/authentication.provider.ts) directly.

See [the same file in my PR](https://github.com/strongloop/loopback-next/blob/53cfeebbaf2ebe2897e28f51432f6fb5d1a4e946/packages/authentication/src/providers/authentication.provider.ts), which contains exactly the same code as yours, and also proves this action is agnostic of the strategy type(passport or non-passport).

packages/authentication/src/__tests__/acceptance/strategy-resolver/strategy.resolver.ts
```diff
@@ -0,0 +1,45 @@
...
+import {AuthenticationStrategy} from '../../../types';
+
+@extensionPoint('authentication-strategy')
+export class StrategyResolverProvider
+  implements Provider<AuthenticationStrategy | undefined> {
+  constructor(
+    @inject(AuthenticationBindings.METADATA)
+    private metadata: AuthenticationMetadata,
+    @extensions() // Sugar for @inject.getter(filterByTag({extensionPoint: 'greeter'}))
+    private authenticationStrategies: Getter<AuthenticationStrategy[]>,
```
jannyHou: The getter here only captures an array of our authentication strategies, therefore I think we need another extension which has a getter for passport-strategies, like

```ts
@extensionPoint('passport-strategy')
export class PassportStrategyResolverProvider
  implements Provider<Strategy | undefined> {
  constructor(
    private metadata: AuthenticationMetadata,
    @extensions()
    private passportStrategies: Getter<Strategy[]>,
  ) {}
  value(): ValueOrPromise<Strategy | undefined> {
    if (!this.metadata) {
      return;
    }
    // make sure the endpoint is decorated with a passport based strategy
    const isPassportStrategy =
          this.metadata.options && this.metadata.options.isPassportStrategy;
    if (!isPassportStrategy) return;

    const name = this.metadata.strategy;

    return this.findPassportStrategy(name).then(function(strategy) {
      if (strategy) {
        // Please note we wrap it with the adapter to return a Loopback auth strategy
        return new StrategyAdapter(strategy);
      } else {
        throw new Error(`The strategy '${name}' is not available.`);
      }
    });
  }

  async findPassportStrategy(name: string) {
    const strategies = await this.passportStrategies();
    const matchingAuthStrategy = strategies.find(a => a.name === name);
    return matchingAuthStrategy;
  }
}
```

Then the `StrategyResolverProvider` invokes the `PassportStrategyResolverProvider` when it finds the endpoint is decorated with a passport strategy.

jannyHou: You can find the prototype implementation in [My PR](https://github.com/strongloop/loopback-next/blob/53cfeebbaf2ebe2897e28f51432f6fb5d1a4e946/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts#L163-L214).
There are 2 provider classes and each should be decorated with `extensionPoint()`.

raymondfeng:
Pls fix lint issue - https://travis-ci.org/strongloop/loopback-next/jobs/521377177

packages/authentication/src/__tests__/acceptance/keys.ts
```diff
@@ -0,0 +1,25 @@
+import {BindingKey} from '@loopback/context';
```
jannyHou: Nitpick: I would suggest moving `keys.ts` to a folder called `fixtures`, same for `services` and `strategies`.

packages/authentication/src/__tests__/acceptance/users/user.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+export interface User {
```
jannyHou: Usually the `User` would be a model, can we define a model instead?

packages/authentication/src/types.ts
```diff
@@ -51,3 +51,54 @@ export interface AuthenticationStrategy {
    */
   authenticate(request: Request): Promise<UserProfile | undefined>;
 }
+
+export interface UserService<U, C> {
```
jannyHou: We created the interface for services in https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/user.service.ts
and 
https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/token.service.ts
No need to duplicate the definition here :-p

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -3,207 +3,189 @@
...
-    function search(key: string) {
-      return userList[key].profile.id === username;
-    }
-    const found = Object.keys(userList).find(search);
-    if (!found) return cb(null, false);
-    if (userList[found].password !== password) return cb(null, false);
-    cb(null, userList[found].profile);
-  }
-}
+// import {inject, Provider, ValueOrPromise} from '@loopback/context';
```
emonddr: @jannyHou stated I should comment out these tests for the meantime. Her PR will re-write the test.

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
```diff
@@ -3,116 +3,116 @@
...
 // License text available at https://opensource.org/licenses/MIT
 
-import {Context, instantiateClass} from '@loopback/context';
-import {Request} from '@loopback/rest';
-import {expect} from '@loopback/testlab';
-import {Strategy} from 'passport';
-import {AuthenticateFn, AuthenticationBindings, UserProfile} from '../../..';
-import {AuthenticateActionProvider} from '../../../providers';
-import {MockStrategy} from '../fixtures/mock-strategy';
+// import {Context, instantiateClass} from '@loopback/context';
```
emonddr: @jannyHou stated I should comment out these tests for the meantime. Her PR will re-write the test.

packages/authentication/src/__tests__/acceptance/users/user.ts
emonddr: The `user repository` used in similar, existing tests was a hard-coded list of JSON objects and didn't involve datasources and models...so I figured there is no need to create a model for my acceptance tests.

packages/authentication/src/__tests__/acceptance/strategy-resolver/strategy.resolver.ts
emonddr: I will take a look. My strategy resolver in my PR will not deal with passport strategies at the moment. This will be handled by your PR. You can rebase your PR branch off of my branch as discussed.

emonddr:
#2311 will have changes which will affect the `AuthenticationStrategyProvider` presented in this work item. It will involve resolving passport related authentication strategies, and will also affect some mocha tests that I have commented out for the moment.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,192 @@ And the abstractions for:
...
+resolver using an
+[ExtensionPoint/Extension Pattern](https://wiki.eclipse.org/FAQ_What_are_extensions_and_extension_points%3F)
+as described in the
+[Greeter extension example](https://github.com/strongloop/loopback-next/tree/master/examples/greeter-extension).
+
+The `AuthenticationStrategyProvider` class in
+`src/providers/auth-strategy.provider.ts` declares an `extension point` named
+`AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME` via the
+`@extensionPoint` decorator. `AuthenticationStrategyProvider` is responsible for
+returning an authentication strategy which has a `specific name` and has been
```
hacksparrow: Is `specific name` a programming construct? If not, we should omit the backticks. Likewise, other backticked words in the doc.

packages/authentication/src/__tests__/fixtures/services/jwt-service.ts
```diff
@@ -0,0 +1,86 @@
...
+      );
+    } //if
+
+    if (!userProfile.id) {
+      throw new HttpErrors['Unauthorized'](
+        `Error generating token : userProfile 'id'  is null`,
+      );
+    } //if
+
+    if (!userProfile.email) {
```
jannyHou: I remember the `email` and `name` are optional, see [interface UserProfile](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/types.ts#L20)

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,47 @@
...
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+    const token: string = this.extractCredentals(request);
+    const userProfile: UserProfile = await this.token_service.verifyToken(
+      token,
+    );
+    return userProfile;
+  }
+
+  extractCredentals(request: Request): string {
+    if (!request.headers.access_token) {
```
jannyHou: From what I've read online, the typical way is checking the authorization header like `request.header.authorization`.

See "How do JSON Web Tokens work?" in https://jwt.io/introduction/

The `access_token` field is usually from the query, see our implementation in the [shopping cart example](https://github.com/strongloop/loopback4-example-shopping/blob/ee557f0dee0a9a5bd30e1f6b8f63b467d23162ca/packages/shopping/src/authentication-strategies/JWT.strategy.ts#L21).

packages/authentication/src/types.ts
```diff
@@ -51,3 +51,13 @@ export interface AuthenticationStrategy {
    */
   authenticate(request: Request): Promise<UserProfile | undefined>;
 }
+
+/**
+ * The error thrown when the authentication strategy resolver
+ * cannot find the specified authentication strategy by name.
+ */
+export class AuthenticationStrategyNotFoundError extends Error {
+  constructor(error: string) {
```
jannyHou: You may want to add a `code` property as the machine-readable error code that could be understood by any client(HTTP/GRPC/SOAP...)

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
jannyHou: @emonddr Could you use `describe.skip()` instead of commenting out the code?
It will be much easier for reviewers to compare the difference in the PR recovering the tests. Thanks!

See the current code change comparison: https://github.com/strongloop/loopback-next/pull/2746/commits/92b10a262233d735d022f752cf4ca82d92f38b09#diff-a5eff729d80d870cc338b162f09467be, all the code are changed, it's hard to see the real difference.

packages/authentication/src/__tests__/fixtures/keys.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export namespace JWTAuthenticationStrategyBindings {
+  export const TOKEN_SECRET = BindingKey.create<string>(
+    'authentication.jwt.secret',
+  );
+  export const TOKEN_EXPIRES_IN = BindingKey.create<string>(
+    'authentication.jwt.expires.in.seconds',
+  );
+  export const TOKEN_SERVICE = BindingKey.create<JWTService>(
```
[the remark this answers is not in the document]
emonddr: Because I am not going to create a binding key with generic interface `UserService`, to be consistent, I won't create a binding key using the generic interface `TokenService`.

packages/authentication/src/__tests__/fixtures/keys.ts
```diff
@@ -0,0 +1,26 @@
...
+import {BasicAuthenticationUserService} from './services/basic-auth-user-service';
+import {JWTService} from './services/jwt-service';
+import {UserRepository} from './users/user.repository';
+
+export const USER_REPO = BindingKey.create<UserRepository>(
+  'authentication.user.repo',
+);
+
+export namespace BasicAuthenticationStrategyBindings {
+  export const USER_SERVICE = BindingKey.create<BasicAuthenticationUserService>(
```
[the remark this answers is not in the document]
emonddr:
```ts
  export const USER_SERVICE = BindingKey.create<UserService>(
    'services.authentication.basic.user.service',
  );

```
produces a typescript error most likely because `UserService` is defined like:
```ts
interface UserService<U, C>
```
In my test cases, I only user the `BasicAuthenticationUserService` for my basic authentication strategy mock implementation. I don't see a point in using `UserService` when creating a binding key.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (credentials.password !== foundUser.password) {
+      throw new HttpErrors.Unauthorized('The password is not correct.');
+    } //if
+
+    return foundUser;
+  }
+
+  convertToUserProfile(user: User): UserProfile {
+    if (!user) {
+      throw new HttpErrors.Unauthorized(`'user' is null`);
```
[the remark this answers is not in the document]
emonddr: I personally like adding `//if` and `//endif` and `//try` and `//catch` next to the `}` in complex nested blocks in order to make my code easier to read...but I will remove the `//if` comments in this piece of code because the code isn't complex.

packages/authentication/src/__tests__/fixtures/services/jwt-service.ts
emonddr: @jannyHou:

 I realize that `email` and `name` can be optional in the `UserProfile` interface we have defined in the authentication package, but this is mainly so we don't force the community to use the two fields we are providing as `suggestions`.

 In my acceptance tests  in the `src/__tests__`folder, I create mock implementation classes for two strategies, for a token service, and for a user service, and I personally decided that each field in `UserProfile` (optional or not) is required for creating a JWT Token. The purpose of the mocha tests are to test the authentication strategy interface, and registering authentication strategies via the extensionPoint/extensions pattern.

packages/authentication/src/types.ts
emonddr: > You may want to add a code property as the machine-readable error code that could be understood by any client(HTTP/GRPC/SOAP...)
@jannyHou , not sure what you mean here.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
```diff
@@ -0,0 +1,83 @@
...
+  constructor(
+    @inject(USER_REPO)
+    private userRepository: UserRepository,
+  ) {}
+
+  async verifyCredentials(
+    credentials: BasicAuthenticationStrategyCredentials,
+  ): Promise<User> {
+    if (!credentials) {
+      throw new HttpErrors.Unauthorized(`'credentials' is null`);
```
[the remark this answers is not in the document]
emonddr: @jannyHou , we discussed that the strategy resolver would return a JS error object since, in the future, we may use it to resolve strategies for authenticating protocol requests other than http rest requests, and that the strategy resolver that ships in the main code of the authentication package shouldn't be returning any http specific errors. 

However my mock implementations of token service and user service and strategies don't ship in the main code of the authentication package and only exist in the src/__tests__ folder, and so I have chosen to have then throw http errors since I am testing them with a rest server.

emonddr:
@jannyHou , I have moved my revised authentication action provider from the main code to the src/tests/fixtures/providers folder since you stated you changed your mind about me replacing the original...that it might cause existing users problems. We will need a new authentication action provider going forward, and the authentication component needs to specify an authentication action provider...so which one will we set? 

I am assuming that your PR will set the appropriate one for the component then.

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
emonddr: @jannyHou , I changed it to use the Authorization header with 'Bearer aaa.bbb.ccc' (where aaa.bbb.ccc is the JWT token ) instead.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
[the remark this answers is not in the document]
jannyHou: @emonddr Fair enough 👍

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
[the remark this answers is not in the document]
jannyHou: Oh ok I am ok with either using the comment or not.

packages/authentication/src/__tests__/fixtures/services/jwt-service.ts
jannyHou: > and I personally decided that each field in UserProfile (optional or not) is required for creating a JWT Token. 

Fair enough 👍

packages/authentication/docs/authentication-system.md
hacksparrow: @emonddr ^

jannyHou:
Was discussing with @emonddr , the next step would be:

- Move the action provider from `__test__/fixtures` to `src/providers`
- Rename the current `authentication-action.provider.ts` to `passport-authentication-action.provider.ts`
- Name the new action provider to be `authentication-action.provider.ts`
- Create a binding key for the deprecated passport auth action provider
- Bind `passport-authentication-action.provider.ts` to the deprecated key
- This PR contains breaking change, let's bump up the version of `@loopback/authentication` to 2.0.0

packages/authentication/src/types.ts
jannyHou: @emonddr The same topic as what we discussed here https://github.com/strongloop/loopback-next/pull/2763#discussion_r278605367 :) 

```ts
export class AuthenticationStrategyNotFoundError extends Error {
  // the `code` would be a string that could be understood by different clients
  // like 'INVALID_ACCESS_TOKEN'
  // see https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/authentication-strategies/JWT.strategy.ts#L35
  code: string;
  constructor(error: string) { 
    super(error);
  }
}
```

Also see my example https://github.com/strongloop/loopback-next/pull/2746/commits/9bda8b7852db196b72129ea6323635b3b46a7d71#diff-440925ebccbc865e10eb2da8810e08eaR9

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -3,37 +3,37 @@
...
+  AuthenticationMetadata,
+  UserProfile,
 } from '../..';
-import {Strategy} from 'passport';
-import {BasicStrategy} from 'passport-http';
 
 const SequenceActions = RestBindings.SequenceActions;
 
-describe('Basic Authentication', () => {
+describe.skip('Basic Authentication', () => {
```
emonddr: I only add '.skip(' to the test file, but it also resorted the imports. Not sure how to prevent that.
It shouldn't matter, @jannyHou's PR, https://github.com/strongloop/loopback-next/pull/2746, needs to rewrite this test file.

packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts
```diff
@@ -108,10 +108,10 @@ describe('AuthenticateActionProvider', () => {
     function givenAuthenticateActionProvider() {
       strategy = new MockStrategy();
       strategy.setMockUser(mockUser);
-      provider = new AuthenticateActionProvider(
-        () => Promise.resolve(strategy),
-        u => (currentUser = u),
-      );
+      // provider = new AuthenticateActionProvider(
```
emonddr: Needed to comment out code here. @jannyHou's PR, https://github.com/strongloop/loopback-next/pull/2746, needs to rewrite this test file.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,194 @@ And the abstractions for:
...
+    @extensions()
+    private authenticationStrategies: Getter<AuthenticationStrategy[]>,
+  ) {}
+  value(): ValueOrPromise<AuthenticationStrategy | undefined> {
+    if (!this.metadata) {
+      return;
+    }
+    const name = this.metadata.strategy;
+
+    return this.findAuthenticationStrategy(name).then(function(strategy) {
```
raymondfeng: Let's consistently use `async-await` instead of `then`.

packages/authentication/src/types.ts
emonddr: @jannyHou , I removed the `AuthenticationStrategyNotFoundError` class, and now I just create an instance of Error, and augment it (with Object.assign) with a `code` field  which contains a specific error.

```ts
    return this.findAuthenticationStrategy(name).then(strategy => {
      if (strategy) {
        return strategy;
      } else {
        // important not to throw a non-protocol-specific error here
        let error = new Error(`The strategy '${name}' is not available.`);
        Object.assign(error, {
          code: AUTHENTICATION_STRATEGY_NOT_FOUND,
        });
        throw error;
      }
    });

```

packages/authentication/src/providers/auth-strategy.provider.ts
```diff
@@ -0,0 +1,58 @@
...
+    @extensions()
+    private authenticationStrategies: Getter<AuthenticationStrategy[]>,
+  ) {}
+  value(): ValueOrPromise<AuthenticationStrategy | undefined> {
+    if (!this.metadata) {
+      return;
+    }
+    const name = this.metadata.strategy;
+
+    return this.findAuthenticationStrategy(name).then(strategy => {
```
raymondfeng: Let's use `async-await` instead of `then`.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,194 @@ And the abstractions for:
...
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options?: ApplicationConfig) {
+    super(options);
+
+    this.component(AuthenticationComponent);
+
+    this.sequence(SequenceIncludingAuthentication);
+
+    addExtension(
```
raymondfeng: Maybe we should add a sugar method such as `app.extension()`.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,194 @@ And the abstractions for:
...
+      this,
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      BasicAuthenticationStrategy,
+      {
+        namespace:
+          AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
```
raymondfeng: Usually, the namespace is in plural form, such as `authentication.strategies`

packages/authentication/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@loopback/authentication",
-  "version": "1.2.1",
+  "version": "2.0.0",
```
raymondfeng: The `version` should be changed by release process.

packages/authentication/src/providers/auth-strategy.provider.ts
emonddr: I didn't realize I was allowed to : add `async` to the `value()` function of the `Provider` interface, and change its return type from `ValueOrPromise` to `Promise`.

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
jannyHou: It's ok, should be some auto fix by prettier/tslint

packages/authentication/src/types.ts
jannyHou: @emonddr thank you, that refactor looks perfect to me 👍

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,514 @@
...
+  });
+
+  async function givenAServer() {
+    app = new Application();
+    app.component(AuthenticationComponent);
+    app.component(RestComponent);
+    server = await app.getServer(RestServer);
+  }
+
+  function givenAuthenticatedSequence() {
```
raymondfeng: Can we extract these functions into a separate `helper` ts file so that it can be shared by multiple tests?

raymondfeng:
@emonddr Quite a few files are missing copyright headers. Please add.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: @raymondfeng , please specify which functions should be extracted and re-used by which tests.   I guess `getAuthenticatedSequence()`  and `givenAServer()` and the `user repo` are common for `packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts` and `packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts`.

emonddr:
added copyright where it was missing.
extracted some common code into a separate sequence class and a helper util.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,192 @@ And the abstractions for:
...
+  email: string;
+  password: string;
+}
+
+export class BasicAuthenticationStrategy implements AuthenticationStrategy {
+  name: string = 'basic';
+
+  constructor(
+    @inject(BasicAuthenticationStrategyBindings.USER_SERVICE)
+    private user_service: BasicAuthenticationUserService,
```
nabdelgadir: Question: why aren't you using camel case here?

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,192 @@ And the abstractions for:
...
+      //
+      try {
+        //call authentication action
+        await this.authenticateRequest(request);
+      } catch (e) {
+        // strategy not found error
+        if (e instanceof AuthenticationStrategyNotFoundError) {
+          throw new HttpErrors.Unauthorized(e.message);
+        } //if
+        else {
```
nabdelgadir: Nitpick: is the `else` needed?

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+
+        return await this.tokenService.generateToken(joeUserProfile);
+      }
+
+      @get('/whoAmI')
+      @authenticate('jwt')
+      whoAmI(
+        @inject(AuthenticationBindings.CURRENT_USER) userProfile: UserProfile,
+      ) {
+        if (userProfile) {
```
nabdelgadir: Suggestion (for here and any other instances of this code):
```ts
if(!userProfile) return 'userProfile is undefined';
if(!userProfile.email) return 'userProfile email is undefined';
return userProfile.email;
```

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+          if (userProfile.email) return userProfile.email;
+          else return 'userProfile email is undefined';
+        } //if
+        else return 'userProfile is undefined';
+      }
+    }
+
+    app.controller(InfoController);
+
+    const the_token: string = (await whenIMakeRequestTo(server)
```
nabdelgadir: Can you just call it `token` (in other tests below too)?

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+import {JWTAuthenticationStrategyBindings, USER_REPO} from '../fixtures/keys';
+import {MyAuthenticationSequence} from '../fixtures/sequences/authentication.sequence';
+import {JWTService} from '../fixtures/services/jwt-service';
+import {JWTAuthenticationStrategy} from '../fixtures/strategies/jwt-strategy';
+import {UserRepository} from '../fixtures/users/user.repository';
+
+describe('JWT Authentication', () => {
+  let app: Application;
+  let server: RestServer;
+  let test_users: UserRepository;
```
nabdelgadir: Why not `testUsers`?

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    const invalidToken = 'aaa.bbb.ccc.ddd';
+
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .set('Authorization', 'Bearer ' + invalidToken)
+      .expect(401);
+  });
+
+  it('create a json web token throws error for missing email', async () => {
```
nabdelgadir: Nitpick (same for the test below): `create` -> `creates`, 
`and throws`?

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
raymondfeng: We should use camelCase for variable names consistently.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
raymondfeng: We use singular form for the verb to describe a test.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -1,14 +1,12 @@
...
+import {Getter, inject, Provider, Setter} from '@loopback/context';
 import {Request} from '@loopback/rest';
-import {Strategy} from 'passport';
 import {AuthenticationBindings} from '../keys';
-import {StrategyAdapter} from '../strategy-adapter';
-import {AuthenticateFn, UserProfile} from '../types';
+import {AuthenticateFn, AuthenticationStrategy, UserProfile} from '../types';
 
 /**
  * @description Provider of a function which authenticates
```
raymondfeng: Let's remove `@description`.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -47,12 +45,9 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
...
-    if (!strategy.authenticate) {
-      throw new Error('invalid strategy parameter');
-    }
-    const strategyAdapter = new StrategyAdapter(strategy);
-    const user = await strategyAdapter.authenticate(request);
-    this.setCurrentUser(user);
-    return user;
+
+    const userProfile = await strategy.authenticate(request);
+    if (userProfile) this.setCurrentUser(userProfile);
```
raymondfeng: Do we need error handling if `userProfile` is not returned from the strategy?

packages/authentication/src/keys.ts
```diff
@@ -87,18 +86,22 @@ export namespace AuthenticationBindings {
...
    * // ... routes that may need authentication
+   * ```
    * }
    */
   export const CURRENT_USER = BindingKey.create<UserProfile | undefined>(
     'authentication.currentUser',
   );
+
+  export const AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME =
+    'authentication-strategies';
```
raymondfeng: Maybe `authentication.strategies` to be consistent?

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,192 @@ And the abstractions for:
...
+```ts
+@extensionPoint(
+  AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  {scope: BindingScope.TRANSIENT},
+)
+export class AuthenticationStrategyProvider
+  implements Provider<AuthenticationStrategy | undefined> {
+  constructor(
+    @inject(AuthenticationBindings.METADATA)
+    private metadata: AuthenticationMetadata,
```
raymondfeng: private metadata?: AuthenticationMetadata,

raymondfeng:
Please either remove https://github.com/strongloop/loopback-next/blob/dremond_authentication_2312/packages/authentication/src/services/session.service.ts or make it an empty interface with copyright headers for now.

packages/authentication/src/services/token.service.ts
```diff
@@ -1,3 +1,8 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
 import {UserProfile} from '../types';
 
 export interface TokenService {
```
raymondfeng: Please add jsdocs.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+  it(`authenticates successfully for correct credentials for user 'jack'`, async () => {
+    const client = whenIMakeRequestTo(server);
+    const credential =
+      users.list['joe@example.com'].user.email +
+      ':' +
+      users.list['joe@example.com'].user.password;
+    const hash = Buffer.from(credential).toString('base64');
+    await client
+      .get('/whoAmI')
+      .set('Authorization', 'Basic ' + hash)
```
raymondfeng: Please refactor the logic to build `Authorization` header from user/pass into a helper function and reuse it for all tests.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+      @authenticate('doesnotexist')
+      status() {
+        return {running: true};
+      }
+    }
+
+    app.controller(InfoController);
+    await whenIMakeRequestTo(server)
+      .get('/status')
+      .expect(401);
```
raymondfeng: Should we use a different http status code when a strategy is not available?

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+  function givenProviders() {
+    addExtension(
+      server,
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      BasicAuthenticationStrategy,
+      {
+        namespace:
+          AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      },
+    );
```
raymondfeng: Let's create a sugar function such as:

```ts
function registerAuthenticationStrategy(strategyClass: Constructor<AuthenticationStrategy>) {
  addExtension(
      server,
      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
      strategyClass,
      {
        namespace:
          AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
      },
    );
}
```

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -159,32 +156,32 @@ describe('Basic Authentication', () => {
...
+    //   // callback method for BasicStrategy
+    //   verify(username: string, password: string, cb: Function) {
+    //     process.nextTick(() => {
+    //       users.find(username, password, cb);
+    //     });
+    //   }
+    // }
+    // server
+    //   .bind(AuthenticationBindings.STRATEGY)
+    //   .toProvider(MyPassportStrategyProvider);
```
raymondfeng: It's probably better to use block comment here.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+  it('authenticates successfully with valid token', async () => {
+    class InfoController {
+      constructor(
+        @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+        public tokenService: JWTService,
+        @inject(USER_REPO)
+        public users: UserRepository,
+      ) {}
+
+      @get('/login')
```
raymondfeng: login is usually over http `post`.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+      }
+    }
+
+    app.controller(InfoController);
+
+    const the_token: string = (await whenIMakeRequestTo(server)
+      .get('/login')
+      .expect(200)).text;
+
+    expect(the_token !== null).to.equal(true);
```
raymondfeng: expect(the_token).to.be.not.null();

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+    expect(typeof the_token === 'string').to.equal(true);
```
raymondfeng: expect(the_token).to.be.string();

packages/authentication/src/__tests__/fixtures/users/user.repository.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {User} from './user';
+
+export class UserRepository {
+  constructor(readonly list: {[key: string]: {user: User}}) {}
```
raymondfeng: Let's get rid of `user` property:
```ts
constructor(readonly list: {[key: string]: User})
```

packages/authentication/src/__tests__/fixtures/users/user.repository.ts
```diff
@@ -0,0 +1,22 @@
...
+  find(email: string, password: string): User | undefined {
+    const userList = this.list;
+    function search(key: string) {
+      return userList[key].user.email === email;
+    }
+    const found = Object.keys(userList).find(search);
+    if (found) {
+      return userList[found].user;
+    } //if
```
raymondfeng: No need to have `// if`.

packages/authentication/src/__tests__/fixtures/users/user.repository.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import {User} from './user';
+
+export class UserRepository {
+  constructor(readonly list: {[key: string]: {user: User}}) {}
+  find(email: string, password: string): User | undefined {
+    const userList = this.list;
+    function search(key: string) {
+      return userList[key].user.email === email;
+    }
```
raymondfeng: Use an arrow function instead:
```ts
Object.keys(userList).find(k => this.list[k].email === email);
```

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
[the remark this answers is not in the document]
raymondfeng: Modern IDE such as VSCode can help find matching `{}` or `()`. Adding such comments is not helpful for other developers as `if` can be nested too.

packages/authentication/docs/authentication-system.md
emonddr: @raymondfeng is this something you will add?

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,192 @@ And the abstractions for:
     - return user
   - controller function:
     - process the injected user
+
+## Registering an authentication strategy via an extension point
```
b-admike: I hope in follow-up PRs we can move this awesome doc to the README.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,199 @@
...
+      ':' +
+      users.list['joe@example.com'].user.password;
+    const hash = Buffer.from(credential).toString('base64');
+    await client
+      .get('/whoAmI')
+      .set('Authorization', 'Basic ' + hash)
+      .expect(users.list['joe@example.com'].user.email);
+  });
+
+  it(`authenticates successfully for correct credentials for user 'jill'`, async () => {
```
b-admike: Feel free to ignore, but I think it'd be good to re-assign `credential` to user `jill` in the test above and assert that auth works for both users (have it in one test)

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    //not passing in 'Authorization' header
+    await whenIMakeRequestTo(server)
+      .get('/whoAmI')
+      .expect(401);
+  });
+
+  it(`returns error for invalid 'Bearer ' portion of Authorization header value`, async () => {
+    class InfoController {
```
b-admike: Consider moving `InfoController` to the bottom of the test as a helper if it can be re-used throughout these tests. Feel free to do it in a follow up PR.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+
+        return await this.tokenService.generateToken(joeUserProfile);
+      }
+    }
+
+    app.controller(InfoController);
+
+    await whenIMakeRequestTo(server)
+      .get('/createtoken')
+      .expect(401);
```
b-admike: For a lot of these tests we are asserting the error code, but is it possible to also assert error messages sent back? Not a deal breaker, but it'll make the tests more robust.

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!auth_header_value.startsWith('Bearer')) {
+      //throw an error
+      throw new HttpErrors.Unauthorized(
+        `Authorization header is not of type 'Bearer'.`,
+      );
+    } //if
+
+    //split the string into 2 parts : 'Bearer ' and the `xxx.yyy.zzz`
+    let parts = auth_header_value.split(' ');
```
b-admike: Do we want to assert that length of parts does not exceed 2 here?

packages/authentication/docs/authentication-system.md
emonddr: the `else` is needed. If you notice there is a try/catch within a try/catch .

 `await this.authenticateRequest(request);` can fail due to: an error resolving the strategy by name (in a strategy resolver), or due to a strategy throwing an error. 

I want to catch the error  of the `strategy not found` from the resolver (which doesn't throw http errors. it throws a protocol agnostic error), and I've chosen (for my test bucket) to have the sequence throw this error as http error.  

For all other errors (coming from the strategies themselves (they are http errors), I will throw them as-is.

The outer catch catches errors thrown from inner catch.

This is not production code, the sequence handler is a test fixture.

packages/authentication/src/providers/auth-strategy.provider.ts
```diff
@@ -0,0 +1,50 @@
...
+  AuthenticationStrategy,
+  AUTHENTICATION_STRATEGY_NOT_FOUND,
+} from '../types';
+
+//this needs to be transient, e.g. for request level context.
+@extensionPoint(
+  AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  {scope: BindingScope.TRANSIENT},
+)
+export class AuthenticationStrategyProvider
```
b-admike: Since this is part of the source code, it'd be great to add TSDocs to here for our API docs.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: I personally don't like this way of writing if statements.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
nabdelgadir: +1

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: @raymondfeng , are you agreeing with me or @nora?

packages/authentication/src/providers/auth-action.provider.ts
emonddr: the strategy.authenticate(request) implementations will throw an exception when : userService.verifyCredentials(user creds) fails with 401, or when tokenService.verifyToken(token) fails with 401.

emonddr:
> Please either remove https://github.com/strongloop/loopback-next/blob/dremond_authentication_2312/packages/authentication/src/services/session.service.ts or make it an empty interface with copyright headers for now.

I will make it an empty interface with copyright headers for now.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
raymondfeng: Yes, I agree with @nabdelgadir to use `creates`.

dhmlau:
@emonddr , FYI,  you can install `strong-tools` globally and run `slt copyright` to add the copyright headers.

packages/authentication/docs/authentication-system.md
emonddr: @b-admike , yes. @jannyHou 's PR : https://github.com/strongloop/loopback-next/pull/2822 overlaps with mine, and we have discussed that we will be moving all separate .md files in the authentication package into a comprehensive readme.md file. And this readme file will also be the source of the loopback.io docs that are generated by our doc build process.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: The Info controllers are different for pretty much all tests, so I will leave as is.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
emonddr: After speaking with @b-admike , we decided to remove the 2nd test involving the user Jill.

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
emonddr: I will add a check, but we must remember these are mock implementations in src/__tests__ and not production code. We don't need to test each line of code for these items. This PR has mocha tests that focus on verifying the authentication strategy interface , registering and finding of authentication strategies via extensionPoints, the authentication action, and the authentication strategy provider. The mock implementations for token service and user service and basic strategy and jwt strategy only serve as fixture helpers and don't need to be tested so heavily.

packages/authentication/src/providers/auth-action.provider.ts
emonddr: I will add a check for undefined userProfile.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
emonddr: regarding
```ts
it('returns error for unknown authentication strategy', async () => {
    class InfoController {
      @get('/status')
      @authenticate('doesnotexist')
      status() {
        return {running: true};
      }
    }

    app.controller(InfoController);
    await whenIMakeRequestTo(server)
      .get('/status')
      .expect(401);
  });
```
If the user decorates the controller method with the `@authenticate` decorator, then this means he/she wants the endpoint protected. If the user specifies a `name` that he/she hasn't registered as an extension, then I strongly believe the user should receive a `401 - Not Authorized`. If we don't throw this error, this will allow access the endpoint they were trying to protect...this isn't good.

If something goes wrong through any code that is helping with authenticating an endpoint, I am returning a 401 Not authorized in my mock strategy implementations and mock user service and token service.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: I tried to use something similar a few weeks ago, and a linter didn't like it. This one worked though.

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
emonddr: I will create a utility to base64 encode `email:password` or any combination of strings required for my tests (valid and invalid combinations)

e.g.
```
.set('Authorization', 'Basic ' + createBase64Hash(users.list['joe@example.com'].email, users.list['joe@example.com'].password))
or 
.set('Authorization', 'NotB@s1c ' + createBase64Hash(users.list['joe@example.com'].email, users.list['joe@example.com'].password))
or 
.set('Authorization', 'Basic ' + createBase64Hash(users.list['joe@example.com'].email, users.list['joe@example.com'].password, {alternativeSeparator: '|'}))
or
.set('Authorization', 'Basic ' + createBase64Hash(users.list['joe@example.com'].email, users.list['joe@example.com'].password, {extraSegment: 'extraPart'}))
```

emonddr:
added two more test cases for auth-action.provider.ts since code coverage decreased after I added an error for undefined userProfile return from a strategy.

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
emonddr: @b-admike , I have addressed your comment. I check error code and message in my tests.

emonddr:
rebased...

packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
emonddr: Plus... we don't ship a custom sequence with the authentication action...the user has to create one. They can decide in the sequence code block , what error code to return. It just so happens I decided to return 401...but they can decide otherwise.

packages/authentication/src/__tests__/fixtures/helper.ts
```diff
@@ -0,0 +1,99 @@
...
+    options && options.alternativePrefix ? options.alternativePrefix : 'Basic ';
+  // default separator is ':', unless another is specified
+  const separator =
+    options && options.alternativeSeparator
+      ? options.alternativeSeparator
+      : ':';
+  // extraPart is '' (blank), unless another is specified
+  const extraPart =
+    options && options.extraSegment ? separator + options.extraSegment : '';
+  return (
```
raymondfeng: This can be simplified as:
```ts
options = Object.assign({
  alternativePrefix: 'Basic ',
  alternativeSeparator: ':',
  extraSegment: ''
}, options);
```

packages/authentication/src/__tests__/fixtures/sequences/authentication.sequence.ts
```diff
@@ -0,0 +1,80 @@
...
+        //call authentication action
+        await this.authenticateRequest(request);
+      } catch (e) {
+        // strategy not found error, or user profile undefined
+        if (
+          e.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
+          e.code === USER_PROFILE_NOT_RETURNED
+        ) {
+          throw new HttpErrors.Unauthorized(e.message);
+        } //if
```
raymondfeng: Please remove `//if`.

packages/authentication/src/__tests__/fixtures/sequences/authentication.sequence.ts
```diff
@@ -0,0 +1,80 @@
...
+        else {
+          // strategy error
+          throw e;
+        } //endif
+      } //catch
```
raymondfeng: Please remove `//endif` and `//catch`.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    if (!credentials.email) {
```
raymondfeng: Basic auth allows `username`. Do we want generalize it as `username` instead of `email`?

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    if (!user.surname) {
+      throw new HttpErrors.Unauthorized(`'user.surname' is null`);
+    }
+
+    if (!user.email) {
+      throw new HttpErrors.Unauthorized(`'user.email' is null`);
+    }
+
```
raymondfeng: Do we really want to make `firstname`, `surname`, `email` required? It is desired sometimes to hide fields such as `email` due to privacy.

I'm also not sure if these errors should be mapped to `HttpErrors.Unauthorized`.

packages/authentication/src/__tests__/fixtures/services/jwt-service.ts
```diff
@@ -0,0 +1,88 @@
...
+      );
+    } //if
+
+    let userProfile: UserProfile;
+
+    try {
+      // decode user profile from token
+      userProfile = await verifyAsync(token, this.jwt_secret);
+    } catch (error) {
+      throw new HttpErrors['Unauthorized'](
```
raymondfeng: Let's use `new HttpErrors.Unauthorized` instead.

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+import {AuthenticationStrategy, UserProfile} from '../../../types';
+import {JWTAuthenticationStrategyBindings} from '../keys';
+import {JWTService} from '../services/jwt-service';
+
+export class JWTAuthenticationStrategy implements AuthenticationStrategy {
+  name: string = 'jwt';
+
+  constructor(
+    @inject(JWTAuthenticationStrategyBindings.TOKEN_SERVICE)
+    public token_service: JWTService,
```
raymondfeng: Please use camelCase `tokenService`.

packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts
```diff
@@ -0,0 +1,54 @@
...
+      token,
+    );
+    return userProfile;
+  }
+
+  extractCredentals(request: Request): string {
+    if (!request.headers.authorization) {
+      //throw an error
+      throw new HttpErrors.Unauthorized(`Authorization header not found.`);
+    } //if
```
raymondfeng: Please remove `//if`

packages/authentication/src/__tests__/fixtures/users/user.repository.ts
```diff
@@ -0,0 +1,19 @@
...
+export class UserRepository {
+  constructor(readonly list: {[key: string]: User}) {}
+  find(email: string): User | undefined {
+    const found = Object.keys(this.list).find(
+      k => this.list[k].email === email,
+    );
+    if (found) {
+      return this.list[found];
+    }
+    return undefined;
```
raymondfeng: It can be simplified as `return found ? this.list[found]: undefined;`.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
emonddr: can't people log in with email and password?

emonddr: email can be a person's userid can it not?

emonddr: I can have two gmail addresses associated with my name and surname. So an email can serve as a username.

packages/authentication/src/__tests__/fixtures/helper.ts
emonddr: That's a great replacement for the ternary operator in this case. Thanks. :)

packages/authentication/src/__tests__/fixtures/sequences/authentication.sequence.ts
emonddr: removed all `//if` occurrences.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
emonddr: The token service and user service and strategies in src/__tests__/fixtures are mock implementations not production implementations in 'src'. Is it absolutely necessary for them to be perfect? The users must create their own production ready artifacts. (Just wondering)

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
emonddr: I am trying to make use of
```ts
export interface UserProfile {
  id: string;
  name?: string;
  email?: string;
}
```
and in our user service interface, we have 
```ts
convertToUserProfile(user: U): UserProfile;
```

where we have a `user` object (we obtained by searching a database with user credentials) and then we convert the larger `user` object into a slimmer `user profile` we place on the context.

Why would we have the email field in UserProfile if we cannot use it?
@jannyHou , what are your thoughts? thx

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
emonddr: The code block above is called when the basic strategy is trying to verify that the user's credentials are good.

called from 
```ts
    const credentials: BasicAuthenticationStrategyCredentials = this.extractCredentals(
      request,
    );
    const user = await this.userService.verifyCredentials(credentials);
    const userProfile = this.userService.convertToUserProfile(user);

    return userProfile;
  }

```

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
dhmlau: Not sure how relevant it is, just wanna point out that there is another discussion thread talking about making the UserProfile more flexible: https://github.com/strongloop/loopback-next/issues/2246.

packages/authentication/src/__tests__/fixtures/strategies/basic-strategy.ts
```diff
@@ -36,20 +36,20 @@ export class BasicAuthenticationStrategy implements AuthenticationStrategy {
     if (!request.headers.authorization) {
       //throw an error
```
raymondfeng: This comment does not add any value here.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
raymondfeng: I'm proposing to be consistent with http://www.passportjs.org/docs/basic-digest/. Please `email` id can be used as `username`.

packages/authentication/src/__tests__/fixtures/services/basic-auth-user-service.ts
raymondfeng: In this case, some of the user profile fields are not populated. It's not something wrong with the credentials. For basic auth, credentials are username/password.

packages/authentication/src/__tests__/fixtures/helper.ts
raymondfeng: Thanks for the update. Next time, please click on `Resolve conversation` if it's fixed.

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,194 @@ And the abstractions for:
...
+      try {
+        //call authentication action
+        await this.authenticateRequest(request);
+      } catch (e) {
+        // strategy not found error, or user profile undefined
+        if (
+          e.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
+          e.code === USER_PROFILE_NOT_RETURNED
+        ) {
+          throw new HttpErrors.Unauthorized(e.message);
```
bajtos: I feel this is adding too much friction for application developers. The current proposal is is good enough for now (incremental baby steps FTW!), but I'd like you to create a follow-up SPIKE issue to look into ways how to allow  `@loopback/authentication` (and any other 3rd party extensions) to contribute custom error-code to status-code mapping.  See the current hard-coded map here:
https://github.com/strongloop/loopback-next/blob/22400fe9a40b17042a5df684300f8a007dd50edd/packages/rest/src/providers/reject.provider.ts#L12-L16

packages/authentication/docs/authentication-system.md
```diff
@@ -163,3 +163,194 @@ And the abstractions for:
...
+      // (in our REST application example). We simply rethrow them.
+      //
+      try {
+        //call authentication action
+        await this.authenticateRequest(request);
+      } catch (e) {
+        // strategy not found error, or user profile undefined
+        if (
+          e.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
+          e.code === USER_PROFILE_NOT_RETURNED
```
bajtos: I find the code `USER_PROFILE_NOT_RETURNED` not very clear - what does it mean that user profile was not returned? Did the code looking it up freeze and never return back to the caller?

I am proposing to use `USER_PROFILE_NOT_FOUND` code instead, following the convention used by `AUTHENTICATION_STRATEGY_NOT_FOUND` and `ENTITY_NOT_FOUND` (from `@loopback/repository`).

packages/authentication/docs/authentication-system.md
emonddr: @bajtos :
> I'd like you to create a follow-up SPIKE issue to look into ways how to allow @loopback/authentication (and any other 3rd party extensions) 

@jannyHou  mentioned there was a task to work on something like this.

Do you mean [Enable custom reject implementation to leverage the built-in "code to statusCode" mapping ](https://github.com/strongloop/loopback-next/issues/1942) ?

emonddr: @bajtos 
> I feel this is adding too much friction for application developers

I am not sure what you mean. The code you mention is inside a `custom sequence` that user must write... we are not shipping the sequence which includes the authentication action. The two agnostic errors I am specifying here are thrown from code that cannot throw protocol specific errors since that code may support protocols other than http in the future.  My mock (not production) implementations for strategies, token service and user service need to throw some kind of errors. I decided for my test bucket to have them throw http errors because my test bucket runs a rest server that handles http requests. The custom sequence rethrows these http exceptions as is, and throws the captured agnostic errors as http errors .

Since users will be writing their own production-ready implementations for `sequence`, `token service`, `user service`, and `strategies` , I am having a hard time understanding how the code above is `painting the user into a corner`. If you could clarify. thanks :)

packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
raymondfeng: What's suggested by @nabdelgadir is consistent with our coding style. The key is to return fast to avoid deep if/else nesting/indentation. It's for readability.

packages/authentication/docs/authentication-system.md
bajtos: @emonddr 

> Do you mean Enable custom reject implementation to leverage the built-in "code to statusCode" mapping ?

Not really. The goal of https://github.com/strongloop/loopback-next/issues/1942 is to allow custom implementations of `reject` to read data from the mapping table. I.e. if the application developer decides to write a different error handling routine, they should have a way how to determine the HTTP status code to return based on error code.

In this discussion, I am asking for a mechanism that will allow extension developers to contribute new mapping entries, so that `reject` implementations can pick them up.

> I am not sure what you mean. The code you mention is inside a custom sequence that user must write...

I find the custom sequence you are proposing as too complex. I would like the custom sequence to look like this:

```ts
class SequenceIncludingAuthentication implements SequenceHandler {
  // ...
    try {
      
      await this.authenticateRequest(request);	

      this.send(response, result);
    } catch (error) {
      this.reject(context, error);
      return;
    }
  }
}
```

Under the hood:
- When `@loopback/authentication` is registered as a component via `app.component`, it contributes new mappings to `codeToStatusCodeMap`.
 - The built-in `reject` action picks up the mapping contributed by the authentication extension and converts `AUTHENTICATION_STRATEGY_NOT_FOUND` to `401 Unauthorized`.

---

BTW the current implementation is problematic because it discards the original error stack trace pointing to the place where the error was triggered (e.g. in the authentication strategy) and replaces it with a stack trace pointing to the custom sequence. This makes troubleshooting rather difficult.

Here is a better solution:

```ts
try {
  await this.authenticateRequest(request);
} catch (e) {
  if (
    e.code === AUTHENTICATION_STRATEGY_NOT_FOUND ||
    e.code === USER_PROFILE_NOT_FOUND
  ) {
    err.statusCode = 401;
  }
  throw err;
}
```

emonddr: Good point about the stack trace. I will look into it. :)

PR #2779, opened 2019-04-20
fix(cli): escape char sequences for javascript comments

raymondfeng:
1. Escape `/*` and `*/` inside `description` for generated code comments.
2. Use `_requestBody` to avoid tslint shadowed name complaints

## Checklist

### DIFF
--- a/packages/cli/generators/openapi/utils.js
+++ b/packages/cli/generators/openapi/utils.js
@@ -128,6 +128,11 @@ const JS_KEYWORDS = [
   'false',
 ];
 
+/**
+ * Avoid tslint error - Shadowed name: 'requestBody'
+ */
+const DECORATOR_NAMES = ['operation', 'param', 'requestBody'];
+
 const SAFE_IDENTIFER = /^[a-zA-Z_$][0-9a-zA-Z_$]*$/;
 
 /**
@@ -147,6 +152,9 @@ function escapeIdentifier(name) {
   }
   if (!name.match(SAFE_IDENTIFER)) {
     name = _.camelCase(name);
+  }
+  if (DECORATOR_NAMES.includes(name)) {
+    return '_' + name;
   }
   return name;
 }
--- a/packages/cli/generators/openapi/utils.js
+++ b/packages/cli/generators/openapi/utils.js
@@ -170,6 +170,16 @@ function escapePropertyName(name) {
   return name;
 }
 
+/**
+ * Escape a string to be used as a block comment
+ *
+ * @param {string} comment
+ */
+function escapeComment(comment) {
+  comment = comment || '';
+  return comment.replace(/\/\*/g, '\\/*').replace(/\*\//g, '*\\/');
+}
+
 function toJsonStr(val) {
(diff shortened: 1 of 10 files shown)


packages/cli/generators/openapi/utils.js
```diff
@@ -148,6 +153,9 @@ function escapeIdentifier(name) {
   if (!name.match(SAFE_IDENTIFER)) {
     name = _.camelCase(name);
   }
+  if (DECORATOR_NAMES.includes(name)) {
```
b-admike: Are these changes related to the escape comment fix? It might be good to separate them into a different commit if not.

packages/cli/test/unit/openapi/controller-spec.unit.js
```diff
@@ -44,13 +44,13 @@ describe('openapi to controllers/models', () => {
...
-              '@param requestBody Customer to add',
-              '@param accessToken Access token',
+              '@param _requestBody Customer to add',
+              '@param accessToken Access token (/* access_token */)',
               '@returns customer response',
             ],
             decoration: "@operation('post', '/customers')",
             signature:
-              'async createCustomer(@requestBody() requestBody: Customer, ' +
+              'async createCustomer(@requestBody() _requestBody: Customer, ' +
```
jannyHou: wouldn't the name here be `customer` instead of `requestBody` or `_requestBody`

packages/cli/generators/openapi/utils.js
raymondfeng: Refactored into two commits.

packages/cli/test/unit/openapi/controller-spec.unit.js
raymondfeng: We use parameter names from the openapi spec instead of inferring it from the type.

jannyHou: I see.
Then it will always be `requestBody`, only parameters have a name in the OpenAPI spec.

packages/cli/test/unit/openapi/openapi-utils.unit.js
```diff
@@ -41,4 +41,11 @@ describe('openapi utils', () => {
     expect(utils.escapePropertyName('customer_id')).to.eql('customer_id');
     expect(utils.escapePropertyName('customerid')).to.eql('customerid');
   });
+
+  it('escapes chars for comments', () => {
+    expect(utils.escapeComment('/* abc */')).to.eql('\\/* abc *\\/');
```
b-admike: I wonder if we should escape them only if it's a complete comment like `/* .... */` instead of `/* ....` or `... */`. I'm thinking perhaps if it's a malformed/unfinished comment we can fail fast.

raymondfeng: We should accept what is in the `description` as-is. It's not our responsibility to check the content of OpenAPI spec `description`s.

b-admike: Okay that makes sense.

Issue #2392, opened 2019-02-14
Example: mount LB3 app in LB4


bajtos:
_This is a follow-up for the spike #2318, see the discussion and PoC code for more context._

Create an example app based on loopback-getting-started (@loopback/example-lb3app) to showcase LB3+LB4 integration. Add acceptance tests to verify few REST endpoints, also a test to verify that LB3 endpoints are included in the OpenAPI spec produced by the LB4 app.

Related issues: #1849 #2389 #2391

## Acceptance criteria

- [ ] A new example project, e.g. `examples/lb3app`
- [ ] Copy loopback-getting-started codebase to `legacy` folder, make the following changes:
   - Disable LB3 error handling in favor of LB4 error handler ([PoC](https://github.com/strongloop/loopback-next/pull/2318/commits/450e637b66121c11f44695be0f110df25a86fc75))
   - Disable static assets in LB3 app ([PoC](https://github.com/strongloop/loopback-next/pull/2318/commits/c8433d31147786ea76a9044a7a3638b9eef5aac4))
   - Add callback to the boot script creating sample data ([PoC](https://github.com/strongloop/loopback-next/pull/2318/commits/1ffa7bab676e54cd4d4d2574b90c67def897e5b3))
   - Use `debug` instead of `console.log` ([PoC](https://github.com/strongloop/loopback-next/pull/2318/commits/eb7942fcc721637679391e501235868ed344977a))
    - Use use in-memory database instead of MySQL to avoid race conditions ([PoC](https://github.com/strongloop/loopback-next/pull/2318/commits/379e0c050c49e70713cabc9aa6d85a892250d375))
- [ ] Add `@loopback/booter-v3app` to the project and configure it to load the legacy app
- [ ] Write LB4-style acceptance tests for the REST API provided by the LB3 app. Cover the following REST API endpoints:
    - `CoffeeShop.create`
    - `CoffeeShop.find`
    - `CoffeeShop.status`
    - `User.create`
    - `User.login`
    - `User.findById` + provide a valid access token, expect 200
    - `User.findById` + provide a token of another user, expect authorization error
- [ ] Add an acceptance test to verify the list of REST API endpoints contributed by the LB3 app. Make sure the test reports a useful error message on failure. Doing a deep-equal comparison of the entire OpenAPI spec IS NOT a good solution.
- [ ] Include the new example app in our docs: [Examples and tutorials](https://loopback.io/doc/en/lb4/Examples-and-tutorials.html)

PR #2572, opened 2019-03-11
feat(build): add eslint scripts and default configs

raymondfeng:
Phase 1 for https://github.com/strongloop/loopback-next/pull/2492:

- Add `@loopback/eslint-config`
- Add `lb-eslint` to `@loopback/build`

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

PR #2803, opened 2019-04-26
feat: add lb3 example application

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/2392.

Implements a new example illustrating how the `booter-lb3app` package's `Lb3AppBooterComponent` works for mounting a LoopBack 3 application in a LoopBack 4 project.

## Checklist


examples/lb3application/.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=true
```
raymondfeng: Is it better to use `lb3-application` as the root directory?

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -0,0 +1,190 @@
...
+      });
+
+      const response = await client
+        .get(`/api/CoffeeShops/greet?access_token=${token.id}`)
+        .expect(200);
+
+      expect(response.body.undefined).to.eql('Hello from this Coffee Shop');
+    });
+
+    it.skip("denies request made by another user's access token", async () => {
```
nabdelgadir: This test and https://github.com/strongloop/loopback-next/pull/2803/files#diff-45fa26c4bf0c5f92945dfc0b15bc5982R92 both pass, but they print the error `Unhandled error for request GET` (anyone have any idea on how to suppress the error?)

examples/lb3-application/lb3app/common/models/coffee-shop.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const debug = require('debug')('loopback:example:lb3application');
+
+module.exports = function(CoffeeShop) {
+  CoffeeShop.status = function(cb) {
```
bajtos: Would you mind rewriting this function from callback-style to `async function`? We are not using callbacks in LB4 and async functions (functions returning a promise) have been supported by LB3 for long time.

```ts
 CoffeeShop.status = async function() {
  // ...
  return response;
}
```

examples/lb3-application/lb3app/common/models/coffee-shop.js
```diff
@@ -0,0 +1,39 @@
...
+    var currentDate = new Date();
+    var currentHour = currentDate.getHours();
+    var OPEN_HOUR = 6;
+    var CLOSE_HOUR = 20;
```
bajtos: Let's use `const` instead of `var` please, see https://loopback.io/doc/en/contrib/style-guide-es6.html#variable-declarations

This comment applies to all newly added JS code.

In the future, I'd like us to enable linter rule to automatically detect this old style, but we are not there yet - see #2492 and #2572.

examples/lb3-application/lb3app/common/models/coffee-shop.js
```diff
@@ -0,0 +1,39 @@
...
+    },
+    returns: {
+      arg: 'status',
+      type: 'string',
+    },
+  });
+  CoffeeShop.greet = function(cb) {
+    process.nextTick(function() {
+      cb(null, 'Hello from this Coffee Shop');
+    });
```
bajtos: Let's rework this to an `async function` and remove `process.nextTick` call - I think it's not relevant for this example.

```ts
CoffeeShop.greet = async function() {
  return 'Hello from this Coffee Shop';
};
```

examples/lb3-application/lb3app/server/boot/create-sample-models.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const debug = require('debug')('loopback:example:lb3application');
+
+module.exports = function(app, cb) {
+  app.dataSources.db.automigrate('CoffeeShop', function(err) {
```
bajtos: Similarly here, I'd prefer to use `await` instead of callbacks. I understand this code is imported from an existing LB3 example app, thus if you don't want to change it, then I would not mind too much.

```ts
module.exports = async function(app) {
  await app.dataSources.db.automigrate('CoffeeShop');
  const coffeeShops = await app.models.CoffeeShop.create([
    // ...
  ]);

  debug('Models created: \n', coffeeShops);
};
```

examples/lb3-application/public/lb3-index.html
```diff
@@ -0,0 +1,5 @@
+<head><title>LoopBack</title></head>
+<body>
+    <h1>LoopBack Rocks!</h1>
+    <p>Hello World... </p>
```
bajtos: I am concerned about discoverability of this page - how can app users learn about the existence of `/lb3-index.html` URL? Can you perhaps add a link to this page to the main `index.html`? For example:

```html
    <h3>OpenAPI spec: <a href="/openapi.json">/openapi.json</a></h3>
    <h3>API Explorer: <a href="/explorer">/explorer</a></h3>
    <h3><a href="/lb3-index.html">Home page from the LB3 app</a></h3>
```

examples/lb3-application/src/__tests__/acceptance/home-page.acceptance.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  before('setupApplication', async () => {
+    ({app, client} = await setupApplication());
+  });
+
+  after(async () => {
+    await app.stop();
+  });
+
+  it('exposes a default home page', async () => {
```
bajtos: Have you considered adding a test to verify LB3 home page `/lb3-index.html` too?

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -0,0 +1,190 @@
...
+      const result = await client.get(`/api/CoffeeShops/${response.body.id}`);
+      expect(result.body).to.containDeep(_.pick(coffeeShop, ['name', 'city']));
+    });
+
+    it("gets the CoffeeShop's status", async () => {
+      const currentDate = new Date();
+      const currentHour = currentDate.getHours();
+      const response = await client.get('/api/CoffeeShops/status').expect(200);
+
+      if (currentHour >= 6 && currentHour < 20) {
```
bajtos: This makes the test brittle and prone to timing errors. Consider this timing:
 - in the test, we observe the current date as `19h59m59.999s`
 - by the time the request reaches the status endpoint, current date becomes `20h00m00.001s`

A better solution: use a fake timer and force `new Date` to always return the same value you want. See https://sinonjs.org/releases/v7.3.2/fake-timers/

Alternatively, I think it's ok to simplify this test to simply check that the response is one of the two expected strings.

```ts
const response = await client.get('/api/CoffeeShops/status').expect(200);
expect(response.body.status).to.be.equalOneOf(
   'We are open for business.',
  'Sorry, we are closed. Open daily from 6am to 8pm.',
);
```

See http://shouldjs.github.io/#assertion-equaloneof

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
bajtos: Essentially, you need to modify the wiring of error handler/error logger to suppress console output for expected non-200 status codes.

See how we do that in `@loopback/rest` tests:

- https://github.com/strongloop/loopback-next/blob/97c01b77fa8872af7e9f09c5abc061b6377c29a5/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts#L214-L220

- https://github.com/strongloop/loopback-next/blob/97c01b77fa8872af7e9f09c5abc061b6377c29a5/packages/rest/src/__tests__/acceptance/routing/routing.acceptance.ts#L844-L851

- https://github.com/strongloop/loopback-next/blob/97c01b77fa8872af7e9f09c5abc061b6377c29a5/packages/rest/src/__tests__/helpers.ts#L15-L33

Maybe we can move `createUnexpectedHttpErrorLogger` from `@loopback/rest` test helpers to `@loopback/testlab` to avoid code duplication?

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -0,0 +1,190 @@
...
+      });
+    });
+
+    it('makes an authenticated request', async () => {
+      await User.create({
+        email: 'sample@email.com',
+        password: 'L00pBack!',
+      });
+
+      // await client.get('/api/CoffeeShops/greet').expect(401);
```
bajtos: Please find a way how to re-enable this check. Personally, I'd move this check to a standalone test.

```ts
it('rejects anonymous requests to protected endpoints', async () => {
  await client.get('/api/CoffeeShops/greet').expect(401);
});
```

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  context('OpenAPI spec', () => {
+    let apiSpec: OpenApiSpec;
+    let swaggerSpec: OpenApiSpec;
+    let lb3Spec: OpenApiSpec;
+
+    before(async () => {
+      apiSpec = app.restServer.getApiSpec();
+      swaggerSpec = await generateSwaggerSpec(lb3App);
+      const result = await swagger2openapi.convertObj(swaggerSpec, {});
```
bajtos: I find this approach problematic. You are basically repeating the same steps used by LB3 booter to convert from Swagger to OpenAPI. If these steps are flawed and produce incorrect spec, then your test will not catch that.

A better solution is to explicitly specify what are the expected tags, schema keys and perhaps also endpoint paths?

```ts
expect(apiSpec.tags).to.eql([
  'CoffeeShops',
  'Users',
  // ...
]);

expect(Object.keys(apiSpec.components || {})).eql([
  'CoffeeShop',
  'User',
  // ...
]);
```

We already have tests to verify details of LB3 booter and Swagger-to-OpenAPI conversion, there is no need to repeat those detailed tests here again. IMO, it's important to test the high-level aspects of the generated spec:  Are CoffeeShop endpoints at expected URL paths? Are all model schemas included?

examples/lb3-application/src/migrate.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import {CoffeeShopApplication} from './application';
+
+export async function migrate(args: string[]) {
+  const existingSchema = args.includes('--rebuild') ? 'drop' : 'alter';
+  console.log('Migrating schemas (%s existing schema)', existingSchema);
+
+  const app = new CoffeeShopApplication();
+  await app.boot();
+  await app.migrateSchema({existingSchema});
```
bajtos: IIUC, this is not going to migrate LB3 models like `CoffeeShop`, `User`, `AccessToken`, etc. Auto-migration of LB3 models is out of scope of this pull request, but I feel it's important to clearly spell out the limitation of the current implementation. 

I am proposing to open a follow-up issue to keep track of this enhancement and modify `migrate.ts` to print a `console.warn` message, e.g.

```ts
console.warn('Skipping migration of models from LB3 application (not supported yet).');
console.warn('See https://github.com/strongloop/loopback/issues/<fill the number>');
await app.migrateSchema({existingSchema});

File: examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts [hunk 13]
```diff
@@ -0,0 +1,218 @@
...
+      });
+
+      const response = await client
+        .get(`/api/CoffeeShops/greet?access_token=${token.id}`)
+        .expect(200);
+
+      expect(response.body.undefined).to.eql('Hello from this Coffee Shop');
+    });
+
+    it.skip('rejects anonymous requests to protected endpoints', async () => {
```
nabdelgadir: I'll update the skipped tests after https://github.com/strongloop/loopback-next/pull/2757 and https://github.com/strongloop/loopback-next/pull/2830 are landed.

examples/lb3-application/.vscode/tasks.json
```diff
@@ -0,0 +1,29 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
+  "version": "2.0.0",
+  "tasks": [
```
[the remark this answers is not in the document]
nabdelgadir: All other examples have the `.vscode` folder and although this one isn't specific to this example, wouldn't it be better to keep it if someone downloads the example from npm?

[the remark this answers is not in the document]
bajtos: I did not realize other example apps have `.vscode` too. Let's stay consistent and keep `.vscode` in the new example too.

examples/lb3-application/src/migrate.ts
nabdelgadir: Issue created: https://github.com/strongloop/loopback-next/issues/2825

examples/lb3-application/README.md
```diff
@@ -0,0 +1,115 @@
...
+     server/
+       boot/
+         # LB3 boot scripts
+   public/
+     # front-end assets (LB4 way)
+   src/
+     # LoopBack 4 application in TypeScript
+   ```
+
+3. Move LB3 dependencies to the main package.json file and remove
```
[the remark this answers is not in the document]
bajtos: +1

examples/lb3-application/lb3app/server/boot/create-sample-models.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const debug = require('debug')('loopback:example:lb3application');
+
+module.exports = async function(app) {
+  await app.dataSources.db.automigrate('CoffeeShop');
+
+  const coffeeShops = app.models.CoffeeShop.create([
```
bajtos:
```suggestion
  const coffeeShops = await app.models.CoffeeShop.create([
```

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -0,0 +1,218 @@
...
+    });
+
+    it('transfers the tags from the LB3 spec to the LB4 spec', () => {
+      expect(apiSpec.tags).to.containDeep([
+        {name: 'User', description: undefined, externalDocs: undefined},
+        {name: 'CoffeeShop', description: undefined, externalDocs: undefined},
+      ]);
+    });
+
+    it.skip('transfers the components from the LB3 spec to the LB4 spec', () => {});
```
bajtos: A test waiting to be implemented after #2757 is landed?

examples/lb3-application/package.json
```diff
@@ -36,27 +36,27 @@
...
     "loopback": "^3.25.1",
     "loopback-boot": "^3.3.0"
   },
   "devDependencies": {
-    "@loopback/build": "^1.5.0",
-    "@loopback/http-caching-proxy": "^1.0.15",
-    "@loopback/rest": "^1.10.4",
-    "@loopback/testlab": "^1.2.5",
+    "@loopback/build": "^1.5.2",
+    "@loopback/http-caching-proxy": "^1.0.17",
```
bajtos: Is this module used by your `lb3-application` example at all? I think it's not, could you please remove it?

Issue #2676, opened 2019-04-02
[Spike] Explore how to roll out experimental features


dhmlau:
_Capturing the discussion we had so far regarding this experimental feature and turning it in a discussion thread on GitHub for others to chime in._

### Why experimental features
While we want to make sure high quality of code goes into our codebase which may take many iterations, releasing the features as `experimental` allow the community to try out and we can get early feedback, and improve the code incrementally. 

----

#### Proposal 1: Introduce experimental flag
The experimental flag means we release features with different stages like what Node.js does: https://nodejs.org/api/documentation.html#documentation_stability_index 

#### Proposal 2: Encourage composition packages
Create a separate package within the loopback-next monorepo but release only as `0.x.y` to indicate it's experimental

#### Proposal 3: Use master-experimental branch
Use `master` as the production ready branch (stay unchanged). Create a `master-experimental` branch for the "master" of experimental features. 

Pros: 
- easier for testing because everything is in one monorepo

Cons: 
- need to sync up between `master` and `master-experimental` branch. (might be able to set up a CI job for rebase)
- might skew our code coverage data

#### Proposal 4: a separate monorepo `loopback-labs`
Pros:
- Clear separation between stable (loopback-next) and experimental (loopback-labs)
- All experiments are in a single place, easy to find and combine together
- Less overhead in repo setup, from CI & DCO checks to GitHub permissions. (when all the experimental projects are in one mono repo)

--- 

### Discussion Topic 2 - Acceptance criteria and graduation criteria 
We need to determine on: 
- clear goal/acceptance criteria for the experimental features
    - explain the expectations and limitations
- graduation criteria/expectation

cc @strongloop/loopback-maintainers

dhmlau:
@strongloop/loopback-next, thoughts?

jannyHou:
@dhmlau The experimental flag means we release features with different stages like what Node.js does: https://nodejs.org/api/documentation.html#documentation_stability_index

(@bajtos could correct me if I am wrong) For "Encourage composition packages", my understanding is we treat features as separate modules that's not part of core.

bajtos:
> we'll create a separate package within the loopback-next monorepo but release only as 0.x.y to indicate it's experimental?
> For "Encourage composition packages", my understanding is we treat features as separate modules that's not part of core.

+1, that's my idea too.

> The experimental flag means we release features with different stages like what Node.js does: https://nodejs.org/api/documentation.html#documentation_stability_index

[...]

PR #2822, opened 2019-05-01
Draft/passport strategy

jannyHou:
**The 2nd commit is the real change**

connect to https://github.com/strongloop/loopback-next/issues/2467
and also https://github.com/strongloop/loopback-next/issues/2311

A draft PR based on https://github.com/strongloop/loopback-next/pull/2763.
Discussion see https://github.com/strongloop/loopback-next/commit/53cfeebbaf2ebe2897e28f51432f6fb5d1a4e946#r33364760 and https://github.com/strongloop/loopback-next/commit/53cfeebbaf2ebe2897e28f51432f6fb5d1a4e946#r33347455

This PR adds another extension point for passport based strategies:

- Passport strategy provider as an extension point: [link](https://github.com/strongloop/loopback-next/pull/2822/commits/cc2d82a3c345594210ca661246a8b0b3e4fba5ce#diff-fa8a1fe478cc37407805319350ccf7e1)

- In the strategy provider, it searches for a passport strategy if no result in the non-passport ones: [link](https://github.com/strongloop/loopback-next/pull/2822/commits/cc2d82a3c345594210ca661246a8b0b3e4fba5ce#diff-23da90c38dcfcc46ee5b733f660675a6)

- The passport strategy is registered as a provider, reason and details please check [the acceptance test](https://github.com/strongloop/loopback-next/pull/2822/commits/cc2d82a3c345594210ca661246a8b0b3e4fba5ce#diff-ab40694bfaf16924ff8eb31695302bb0)

- One skipped test to be recovered (need sometime to check the error message, not a big deal), more tests to be added for the new extension point.

## Checklist

### DIFF
--- a/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
@@ -3,37 +3,37 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {inject, Provider, ValueOrPromise} from '@loopback/context';
 import {Application} from '@loopback/core';
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {api, get} from '@loopback/openapi-v3';
 import {
-  RestBindings,
-  ParseParams,
   FindRoute,
   InvokeMethod,
-  Send,
+  ParseParams,
   Reject,
-  SequenceHandler,
-  RestServer,
-  RestComponent,
   RequestContext,
+  RestBindings,
+  RestComponent,
+  RestServer,
+  Send,
+  SequenceHandler,
 } from '@loopback/rest';
-import {api, get} from '@loopback/openapi-v3';
 import {Client, createClientForHandler} from '@loopback/testlab';
-import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
-import {inject, Provider, ValueOrPromise} from '@loopback/context';
-import {
-  authenticate,
-  UserProfile,
-  AuthenticationBindings,
-  AuthenticateFn,
-  AuthenticationMetadata,
-  AuthenticationComponent,
-} from '../..';
 import {Strategy} from 'passport';
 import {BasicStrategy} from 'passport-http';
+import {
+  authenticate,
+  AuthenticateFn,
+  AuthenticationBindings,
+  AuthenticationComponent,
+  AuthenticationMetadata,
+  UserProfile,
+} from '../..';
 
 const SequenceActions = RestBindings.SequenceActions;
 
-describe('Basic Authentication', () => {
+describe.skip('Basic Authentication', () => {
   let app: Application;
   let server: RestServer;
   let users: UserRepository;
--- a/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
@@ -3,8 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {inject, Provider, ValueOrPromise} from '@loopback/context';
-import {Application} from '@loopback/core';
+import {inject} from '@loopback/context';
+import {addExtension, Application, Provider} from '@loopback/core';
 import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
 import {api, get} from '@loopback/openapi-v3';
 import {
@@ -20,20 +20,19 @@ import {
   SequenceHandler,
 } from '@loopback/rest';
 import {Client, createClientForHandler} from '@loopback/testlab';
-import {Strategy} from 'passport';
-import {BasicStrategy} from 'passport-http';
+import {BasicStrategy, BasicVerifyFunction} from 'passport-http';
 import {
   authenticate,
   AuthenticateFn,
   AuthenticationBindings,
   AuthenticationComponent,
-  AuthenticationMetadata,
   UserProfile,
 } from '../..';
-
+import {StrategyAdapter} from '../../strategy-adapter';
+import {AuthenticationStrategy} from '../../types';
 const SequenceActions = RestBindings.SequenceActions;
 
-describe.skip('Basic Authentication', () => {
+describe('Basic Authentication', () => {
   let app: Application;
   let server: RestServer;
   let users: UserRepository;
@@ -41,7 +40,6 @@ describe.skip('Basic Authentication', () => {
   beforeEach(givenUserRepository);
   beforeEach(givenControllerInApp);
   beforeEach(givenAuthenticatedSequence);
-  beforeEach(givenProviders);
 
   it('authenticates successfully for correct credentials', async () => {
     const client = whenIMakeRequestTo(server);
@@ -87,10 +85,45 @@ describe.skip('Basic Authentication', () => {
     });
   }
 
+  // Since it has to be user's job to provide the `verify` function and
+  // instantiate the passport strategy, we cannot add the imported `BasicStrategy`
+  // class as extension directly, we need to wrap it as a strategy provider,
+  // then add the provider class as the extension.
+  // See Line 89 in the function `givenAServer`
+  class PassportBasicAuthProvider implements Provider<AuthenticationStrategy> {
+    value(): AuthenticationStrategy {
+      const basicStrategy = this.configuratedBasicStrategy(verify);
+      return this.convertToAuthStrategy(basicStrategy);
+    }
+
+    configuratedBasicStrategy(verifyFn: BasicVerifyFunction): BasicStrategy {
+      return new BasicStrategy(verifyFn);
+    }
+
+    convertToAuthStrategy(basic: BasicStrategy): AuthenticationStrategy {
+      return new StrategyAdapter(basic, 'basic');
+    }
+  }
+
+  function verify(username: string, password: string, cb: Function) {
+    process.nextTick(() => {
+      users.find(username, password, cb);
+    });
+  }
+
   async function givenAServer() {
     app = new Application();
     app.component(AuthenticationComponent);
     app.component(RestComponent);
+    addExtension(
+      app,
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      PassportBasicAuthProvider,
+      {
+        namespace:
+          AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+      },
+    );
     server = await app.getServer(RestServer);
   }
 
@@ -115,7 +148,7 @@ describe.skip('Basic Authentication', () => {
         @inject(AuthenticationBindings.CURRENT_USER) private user: UserProfile,
       ) {}
 
-      @authenticate('BasicStrategy')
+      @authenticate('basic')
       async whoAmI(): Promise<string> {
         return this.user.id;
       }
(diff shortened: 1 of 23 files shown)


packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -87,10 +83,36 @@ describe.skip('Basic Authentication', () => {
     });
   }
 
+  // Since it has to be user's job to provide the `verify` function and
+  // instantiate the passport strategy, we cannot add the imported `BasicStrategy`
+  // class as extension directly, we need to wrap it as a strategy provider,
+  // then add the provider class as the extension.
+  // See Line 89 in the function `givenAServer`
```
bajtos: Since we are not able to register a Passport strategy class directly and have to create a custom provider class, I would like to propose a slightly different user experience that does not require any new extension points to be created.

Usage:

```ts
import {Strategy, PassportStrategyAdapter} from '@loopback/authentication';

class PassportBasicAuthProvider implements Provider<Strategy> {
    value(): Strategy {
      // create an instance of the Passport Strategy
      // configure it with "verify" and any other options/behavior as needed
      const passportStrategy = new BasicStrategy(this.verify.bind(this));

      // wrap the Passport Strategy class instance and 
      // adapt it to match LB Strategy interface contract
      return new PassportStrategyAdapter(passportStrategy);
    }

  verify(username: string, password: string, cb: Function) {
    process.nextTick(() => {
      users.find(username, password, cb);
    });
  }
}

// register PassportBasicAuthProvider as an extension
// for AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME
// (a regular LB4 Authentication strategy)
```

Implementation wise, I think we just need to export existing `StrategyAdapter` class for public consumption, preferably under a more descriptive name. In my example, I am using the name `PassportStrategyAdapter`. (In the future, we can implement additional adapters. E.g. `Auth0StrategyAdapter` (see https://auth0.com, I am totally making this example up.)

An important feature of this approach is extensibility. When each strategy contract (LB4, Passport, etc.) requires a new extension point, then it's difficult to implement support for new strategy contracts in 3rd-party modules. With the proposed Adapter approach, the core authentication framework does not need to deal with different strategy contracts and thus adding a new strategy contract does not require any changes in the core.

In fact, I think we should move `PassportStrategyAdapter` into a standalone package:
 - It will validate extensibility of our design and enforce proper separation of concerns.
 - People using only LB strategies should not have to install `passport` dependency. By moving the passport adapter to a new package, only consumers of this new package will depend on `passport`.
 - As we learn more about Passport, we may need to introduce breaking change in the passport adapter. Ideally, these breaking changes should affect only Passport users, they should not trigger a semver-major release of the entire authentication framework.

/cc @raymondfeng

jannyHou: > Since we are not able to register a Passport strategy class directly and have to create a custom provider class, I would like to propose a slightly different user experience that does not require any new extension points to be created.


Neat 👍 yeah since the adapter implements `AuthenticationStrategy` it makes sense to only have one extension. And this could also be a good practice for other adapters like you said in 
> we can implement additional adapters, e.g. Auth0StrategyAdapter (see https://auth0.com, I am totally making this example up.)

raymondfeng:
There are two possible options to adapt Passport strategies:

1. Adapt a passport strategy into our AuthenticationStrategy and register it as a regular extension.
2. Adapt `passport` middleware and use existing passport apis to register passport strategies.

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
jannyHou: >  With the proposed Adapter approach, the core authentication framework does not need to deal with different strategy contracts and thus adding a new strategy contract does not require any changes in the core.

+1 Fair enough.

jannyHou:
@raymondfeng Thanks for the suggestion

> 1. Adapt a passport strategy into our AuthenticationStrategy and register it as a regular extension.

I believe this is the approach that this PR is implementing.

> 2. Adapt passport middleware and use existing passport apis to register passport strategies.

👍 I am also interested to explorer it, while it looks more like documenting the best practice of applying an express middleware `passport`, and has nothing to do with our authentication metadata/strategy interface/extension point. And I believe it won't affect the existing authentication action. I can try it in https://github.com/strongloop/loopback-next/issues/2311, but keep it out of the scope of https://github.com/strongloop/loopback-next/issues/2467, are you good with the plan?

jannyHou:
Tried wrapping `passport` on local, some findings:

Here is a prototype of the passport adapter provider:

- It injects a `Passport` instance which is created and bind to app
   
   ```ts
   const passport = new Passport();
   registePassportStrategies(passport);

   // Register the configured passport strategies
   function registePassportStrategies(passport: Passport) {
     passport.use(new BasicStrategy(verify_basic));
     passport.use(new JWTStrategy(verify_jwt));
   }

   app.bind(AuthenticationBindings.PASSPORT).to(passport);
   ```

- (?) Invoke the `passport.authenticate('strategyName')` to authenticate the request. The main blocker is I couldn't imagine how to invoke the `passport.authenticate` method in the action, with the request from context.

```ts
export class PassportAdapterProvider implements Provider<AuthenticateFn> {
  constructor(
    // `Passport` is a function and cannot be used as a type here
    // need to build a type to describe the instance of Passport
    @inject(AuthenticationBindings.PASSPORT)
    readonly passport: Passport,
    private metadata?: AuthenticationMetadata,
  ) { }

  /**
   * @returns authenticateFn
   */
  value(): AuthenticateFn {
    return request => this.action(request);
  }

  /**
   * The implementation of authenticate() sequence action.
   * @param request The incoming request provided by the REST layer
   */
  async action(request: Request): Promise<UserProfile | undefined> {
    // we can infer the strategy name from `metadata.name` as an alternate.
    const strategyName = metadata.options.strategyName;

    const asyncAuthenticate = promisify(passport.authenticate);
    const user = await asyncAuthenticate(strategyName);
    // (Blocker) How to authenticate a request using `passport.authenticate` in the action?

    // We need a function to convert the `user` to `userProfile`
    const userProfile = user;
    return userProfile;
  }
}
```

Since the existing adapter wraps the passport strategy instead of `passport`, story https://github.com/strongloop/loopback-next/issues/2467 and https://github.com/strongloop/loopback-next/issues/2311 are created based on that assumption. 

I am afraid I need more time to flush out the uncertainties of wrapping `passport`, how about I recover the tests:
- https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts#L33
- https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/unit/providers/authentication.provider.unit.ts#L13

in story https://github.com/strongloop/loopback-next/issues/2467.

Then explorer wrapping `passport` in story https://github.com/strongloop/loopback-next/issues/2311? We have more time to decide which one is better or keep both.

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -87,10 +85,45 @@ describe.skip('Basic Authentication', () => {
...
+  class PassportBasicAuthProvider implements Provider<AuthenticationStrategy> {
+    value(): AuthenticationStrategy {
+      const basicStrategy = this.configuratedBasicStrategy(verify);
```
emonddr: @jannyHou , use `configured` instead of `configurated`.

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -87,10 +85,45 @@ describe.skip('Basic Authentication', () => {
...
+      return this.convertToAuthStrategy(basicStrategy);
+    }
+
+    configuratedBasicStrategy(verifyFn: BasicVerifyFunction): BasicStrategy {
+      return new BasicStrategy(verifyFn);
+    }
+
+    convertToAuthStrategy(basic: BasicStrategy): AuthenticationStrategy {
+      return new StrategyAdapter(basic, 'basic');
```
emonddr: @jannyHou , let's store the 'basic' identifier string for the authentication strategy in a constant somewhere else.

packages/authentication/src/__tests__/unit/strategy-adapter.unit.ts
```diff
@@ -16,33 +16,37 @@ describe('Strategy Adapter', () => {
     it('calls the authenticate method of the strategy', async () => {
       let calledFlag = false;
       // TODO: (as suggested by @bajtos) use sinon spy
-      class Strategy extends MockStrategy {
+      class Strategy extends MockPassportStrategy {
```
emonddr: @jannyHou , Please use a custom strategy class other than `Strategy` so it is not confused by the `Strategy` class used by passport-strategy. thx.

emonddr: @jannyHou ^^^ Please see my comments. thx.

emonddr:
@jannyHou , it is possible for you to rebase off of loopback-next master? my changes from my feature branch (from which you based your feature branch) are in master branch now.

For basic-auth-acceptance.ts, please use this sequence : https://github.com/strongloop/loopback-next/tree/master/packages/authentication/src/__tests__/fixtures/sequences .

emonddr:
In basic-auth-acceptance.ts,
we can turn off the `.skip` for this mocha test

```ts
      it('throws an error if the injected passport strategy is not valid', async () => {
        const context: Context = new Context();
        context
          .bind(AuthenticationBindings.STRATEGY)
          .to({} as AuthenticationStrategy);
        context
          .bind(AuthenticationBindings.AUTH_ACTION)
          .toProvider(AuthenticateActionProvider);
        const authenticate = await context.get<AuthenticateFn>(
          AuthenticationBindings.AUTH_ACTION,
        );
        const request = <Request>{};
        let error;
        try {
          await authenticate(request);
        } catch (exception) {
          error = exception;
        }
        expect(error).to.have.property('message', 'strategy.authenticate is not a function');
      });

```
and use the error message listed. 
This catches the missing `authenticate` function error
in line: https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-action.provider.ts#L52  . This is because you used   `{} as AuthenticationStrategy`.

To catch https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-strategy.provider.ts#L45, perhaps add a test like: https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts#L145 and https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts#L390.

jannyHou:
@emonddr Thank you for the comments. This draft PR is a PoC of the new strategy adapter, so some implementation details are not refined yet and will be rewrote/improved in the official feature PR.

The plan of moving forward this PR is:

- [x] Recover the auth action tests, merged PR https://github.com/strongloop/loopback-next/pull/2859.
- [ ] [In progress] Spike https://github.com/strongloop/loopback-next/issues/2676#issuecomment-488746964
- [ ] Create the passport adapter in `loopback-labs`(or the core repo is we decide to have feature branches in the same repo)

I removed the test file `basic-auth-acceptance.ts` since it should be part of the new passport adapter module. Thank you for the very detailed comment in https://github.com/strongloop/loopback-next/pull/2822#issuecomment-493109074 and https://github.com/strongloop/loopback-next/pull/2822#issuecomment-493084698, I will address them in the passport adapter PR.

packages/authentication/src/__tests__/unit/strategy-adapter.unit.ts
jannyHou: @emonddr Thanks good catch. This code change is merged in PR https://github.com/strongloop/loopback-next/pull/2859. I can address it in the coming doc PR that moves https://github.com/strongloop/loopback-next/blob/master/packages/authentication/docs/authentication-system.md to the root level README.md file.

jannyHou:
I am closing the draft to keep open PRs less.
The implementation plan is https://github.com/strongloop/loopback-next/pull/2822#issuecomment-493246380

PR #2827, opened 2019-05-03
fix(rest): aggressive redirection to Swagger UI

hacksparrow:
Send 302, instead of 308 for Swagger UI redirection.

Fixes https://github.com/strongloop/loopback-next/issues/2808

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #2829, opened 2019-05-03
chore(testlab): move from deprecated `shot` to `@hapi/shot`

bajtos:
This should eventually fix most of the deprecation warnings reported by `npm install` executed from our CLI tests (see e.g. https://github.com/strongloop/loopback-next/pull/2827) ~~Please note that we need to publish a new version of `@loopback/testlab` to bring this change into effect.~~

```
npm WARN deprecated shot@4.0.7: This module has moved and is now available at @hapi/shot. Please update your dependencies as this version is no longer maintained an may contain bugs and security issues.
npm WARN deprecated hoek@6.1.3: This module has moved and is now available at @hapi/hoek. Please update your dependencies as this version is no longer maintained an may contain bugs and security issues.
npm WARN deprecated joi@14.3.1: This module has moved and is now available at @hapi/joi. Please update your dependencies as this version is no longer maintained an may contain bugs and security issues.
npm WARN deprecated topo@3.0.3: This module has moved and is now available at @hapi/topo. Please update your dependencies as this version is no longer maintained an may contain bugs and security issues.
```

This patch IS NOT fixing the following deprecation message:

```
npm WARN deprecated @types/handlebars@4.1.0: This is a stub types definition. handlebars provides its own type definitions, so you do not need this installed.
```

Such fix would be more involved because `@types/handlebars` is a deep dependency out of our control.

```
@loopback/build@1.5.0
  └─┬ strong-docs@4.2.0
    └─┬ typedoc@0.14.2
      └── @types/handlebars@4.1.0
```

## Checklist

### DIFF
--- a/packages/testlab/src/shot.ts
+++ b/packages/testlab/src/shot.ts
@@ -10,21 +10,25 @@
 
 // tslint:disable:no-any
 
+import * as express from 'express';
 import {IncomingMessage, ServerResponse} from 'http';
-import * as util from 'util';
-
 import {
+  Listener as ShotListener,
   RequestOptions as ShotRequestOptions,
   ResponseObject,
-  inject,
-} from 'shot';
+} from 'shot'; // <-- workaround for missing type-defs for @hapi/shot
+import * as util from 'util';
 
-import * as express from 'express';
+const inject: (
+  dispatchFunc: ShotListener,
+  options: ShotRequestOptions,
+) => Promise<ResponseObject> = require('@hapi/shot');
+// ^^ workaround for missing type-defs for @hapi/shot
 
 export {inject, ShotRequestOptions};
 
 // tslint:disable-next-line:variable-name
-const ShotRequest: ShotRequestCtor = require('shot/lib/request');
+const ShotRequest: ShotRequestCtor = require('@hapi/shot/lib/request');
 type ShotRequestCtor = new (options: ShotRequestOptions) => IncomingMessage;
 
 export function stubServerRequest(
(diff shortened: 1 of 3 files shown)


bajtos:
@hacksparrow PTAL again, I have to rework the way how we are importing code and types from the new Shot module.

bajtos:
Strangely enough, I am no longer able to reproduce the handlebars warning as can be seen from the checks that are all green.

packages/testlab/src/shot.ts
```diff
@@ -10,21 +10,25 @@
...
 import {IncomingMessage, ServerResponse} from 'http';
-import * as util from 'util';
-
 import {
+  Listener as ShotListener,
   RequestOptions as ShotRequestOptions,
   ResponseObject,
-  inject,
-} from 'shot';
+} from 'shot'; // <-- workaround for missing type-defs for @hapi/shot
```
b-admike: Correct me if I'm wrong, but are we not still importing from `shot` module? Does that mean we should still leave the `shot` dependency in `package.json`?

bajtos: This is importing **type declarations** only. The import statement is used by TypeScript compiler only, it resolves to importing types from `@types/shot` package and does not emit any JavaScript code.

The actual implementation of various Shot functions is always imported from `@hapi/shot` package, as can be seen elsewhere.

It's hacky, I hope somebody will publish new `@types/*` package soon and we will be able to get rid of this hack.

bajtos: > Are we not still importing from shot module?

No, we are not. This is just a way how to tell TypeScript to import type declarations from `@types/shot`, because it's not possible to write `import {...} from '@types/shot'`.

> Does that mean we should still leave the shot dependency in package.json?

No, because we are not importing from `shot` at runtime (nor compile time). If we kept `shot` in our dependencies, then we would still receive the deprecation warning, which would defeat the purpose of this change.

b-admike: Okay thanks for explaining, I see the `@types/shot` package 👍

PR #2830, opened 2019-05-03
refactor: move unexpectedHttpErrorLogger to testlab

nabdelgadir:
See https://github.com/strongloop/loopback-next/pull/2803#discussion_r280324582 for motivation.

Moving the `unexpectedHttpErrorLogger` from `@loopback/rest` to `@loopback/testlab` so it can be used for tests in other packages/examples.

## Checklist

### DIFF
--- a/packages/testlab/README.md
+++ b/packages/testlab/README.md
@@ -54,6 +54,8 @@ Table of contents:
 - [httpGetAsync](#httpgetasync) - Async wrapper for HTTP GET requests.
 - [httpsGetAsync](#httpsgetasync) - Async wrapper for HTTPS GET requests.
 - [toJSON](#toJSON) - A helper to obtain JSON data representing a given object.
+- [createUnexpectedHttpErrorLogger](#createUnexpectedHttpErrorLogger) - An error
+  logger that only logs errors for unexpected HTTP statuses.
 
 ### `expect`
 
@@ -283,6 +285,46 @@ describe('MyApp', () => {
     const app = new MyApp();
     const server = await app.getServer(RestServer);
     await validateApiSpec(server.getApiSpec());
+  });
+});
+```
+
+### `createUnexpectedHttpErrorLogger`
+
+An error logger that logs the error only when the HTTP status code is not the
+expected HTTP status code. This is useful when writing tests for error
+responses:
+
+- When we don't want any error messages printed to the console when the server
+  responds with the expected error and the test passes.
+
+- When something else goes wrong and the server returns an unexpected error
+  status code, and we do want an error message to be printed to the console so
+  that we have enough information to troubleshoot the failing test.
+
+```ts
+import {createUnexpectedHttpErrorLogger} from '@loopback/testlab';
+import {RestApplication} from '@loopback/rest';
+
+describe('MyApp', () => {
+  it('does not log a known 401 error to console', async () => {
+    const app = new RestApplication();
+    const errorLogger = createUnexpectedHttpErrorLogger(401);
+
+    const spec = {
+      responses: {
+        /*...*/
+      },
+    };
+    function throwUnauthorizedError() {
+      throw new HttpErrors.Unauthorized('Unauthorized!');
+    }
+
+    app.route('get', '/', spec, throwUnauthorizedError);
+
+    // binds the custom error logger
+    app.bind(SequenceActions.LOG_ERROR).to(errorLogger);
+    await app.start();
   });
 });
 ```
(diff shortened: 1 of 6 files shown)


packages/testlab/src/http-error-logger.ts
```diff
@@ -0,0 +1,24 @@
+import {Request} from 'express';
+import {IncomingMessage} from 'http';
+
+export function createUnexpectedHttpErrorLogger(
+  expectedStatusCode: number = 0,
```
raymondfeng: Should we use `200` as the default value?

raymondfeng: Can we accept `...expectedStatusCodes: number[]` to allow multiple status codes without breaking existing code?

bajtos: The purpose of this helper is to disable error logs printed by our default error handler for the status code expected by an integration/end-to-end test making HTTP calls to the tested application.

Typically, such test is verifying the response status code against a single value only.

```ts
await client.get('/foo/bar').expect(404);
```

> Should we use `200` as the default value?

That does not make much sense to me - no error is logged for 2xx and 3xx status codes. This helper is useful only for tests expecting 4xx or 5xx response.

> Can we accept `...expectedStatusCodes: number[]` to allow multiple status codes without breaking existing code?

I think so. However, I am not convinced this a good practice to support. Let's defer this enhancement until we have a real-world use case needing it.

packages/testlab/src/http-error-logger.ts
```diff
@@ -0,0 +1,24 @@
+import {Request} from 'express';
+import {IncomingMessage} from 'http';
+
+export function createUnexpectedHttpErrorLogger(
```
bajtos: Please add API documentation explaining how to use this helper and more importantly, describe this new helper in testlab's README: https://github.com/strongloop/loopback-next/blob/master/packages/testlab/README.md

packages/testlab/src/http-error-logger.ts
raymondfeng: > That does not make much sense to me - no error is logged for 2xx and 3xx status codes. This helper is useful only for tests expecting 4xx or 5xx response.

Then why don't we have the following signature? It leaves the default to `undefined` instead of `0` to avoid similar confusions as I had.
```ts
export function createUnexpectedHttpErrorLogger(
  expectedStatusCode? number,
) {}
```

raymondfeng:
Please also check the test coverage - https://coveralls.io/builds/23172908/source?filename=packages%2Ftestlab%2Fsrc%2Fhttp-error-logger.ts#L14

bajtos:
> Please also check the test coverage - https://coveralls.io/builds/23172908/source?filename=packages%2Ftestlab%2Fsrc%2Fhttp-error-logger.ts#L14

Good catch!

Screen Shot 2019-05-03 at 17 57 33

I think we should disable code coverage for the `console.error` branch, because this branch is never executed on successful CI builds.

I think `/* istanbul ignore next */ ` is the tool to use? See https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

nabdelgadir:
The [test](https://github.com/strongloop/loopback-next/pull/2830/commits/30a4989a2f7ecebc89018fc4a79bcb1795239541#diff-bd22a2988bbd52731761228640ba62bdR12) I created initially does pass, but it requires `@loopback/rest` which causes a circular dependency, so I had to remove it.

packages/testlab/src/http-error-logger.ts
bajtos: I don't remember why I picked `0` as the default value. Your proposal to use `undefined` instead of `0` is fine with me 👍

packages/testlab/README.md
```diff
@@ -287,6 +289,39 @@ describe('MyApp', () => {
 });
 ```
 
+### `createUnexpectedHttpErrorLogger`
+
+An error logger that logs the error only when the HTTP status code is not the
+expected HTTP status code. This helps surpress the console from printing an
+expected error.
```
bajtos: > This helps surpress the console from printing an expected error.

Let's improve this part. My proposal - feel free to change it as you like:

```markdown
This is useful when writing tests for error responses:

- We don't want any error message printed to the console when 
  the server responds with the expected error and the test passes.

- When something else goes wrong and the server returns an unexpected 
  error status code, we do want an error message to be printed to the console 
  so that we have enough information to troubleshoot the failing test.
```

packages/testlab/src/http-error-logger.ts
nabdelgadir: Updated to use `undefined`. 👍

PR #2834, opened 2019-05-03
feat(tsdocs): add integration with api-extractor/documenter

raymondfeng:
See #2507 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #2836, opened 2019-05-05
chore: fix invalid tsdoc tags in comments

raymondfeng:
Invalid block tags prevent api-documenter from generating apidocs.

Relates to #2834 

## Checklist

### DIFF
--- a/packages/boot/src/booters/base-artifact.booter.ts
+++ b/packages/boot/src/booters/base-artifact.booter.ts
@@ -27,13 +27,13 @@ const debug = debugFactory('loopback:boot:base-artifact-booter');
  *
  * Currently supports the following boot phases: configure, discover, load
  *
- * @property options Options being used by the Booter
- * @property projectRoot Project root relative to which all other paths are resovled
- * @property dirs Directories to look for an artifact in
- * @property extensions File extensions to look for to match an artifact (this is a convention based booter)
- * @property glob Glob pattern to use to discover artifacts. Takes precedence over (dirs + extensions)
- * @property discovered List of files discovered by the Booter that matched artifact requirements
- * @property classes List of exported classes discovered in the files
+ * - options: Options being used by the Booter
+ * - projectRoot: Project root relative to which all other paths are resovled
+ * - dirs: Directories to look for an artifact in
+ * - extensions: File extensions to look for to match an artifact (this is a convention based booter)
+ * - glob: Glob pattern to use to discover artifacts. Takes precedence over (dirs + extensions)
+ * - discovered: List of files discovered by the Booter that matched artifact requirements
+ * - classes: List of exported classes discovered in the files
  */
 export class BaseArtifactBooter implements Booter {
   /**
--- a/packages/boot/src/interfaces.ts
+++ b/packages/boot/src/interfaces.ts
@@ -32,9 +32,6 @@ export type ArtifactOptions = {
  * load(): Promise<void>
  *
  * A Booter will run through the above methods in order.
- *
- * @export
- * @interface Booter
  */
 export interface Booter {
   /**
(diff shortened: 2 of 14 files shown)


packages/boot/src/interfaces.ts
```diff
@@ -60,10 +57,10 @@ export const BOOTER_PHASES = ['configure', 'discover', 'load'];
...
  * Type Object for Options passed into .boot()
  *
- * @property projectRoot Root of project. All other artifacts are resolved relative to this
- * @property booters An array of booters to bind to the application before running bootstrapper
- * @property filter.booters An array of booters that should be run by the bootstrapper
- * @property filter.phases An array of phases that should be run
+ * - projectRoot: Root of project. All other artifacts are resolved relative to this
+ * - booters: An array of booters to bind to the application before running bootstrapper
+ * - filter.booters: An array of booters that should be run by the bootstrapper
+ * - filter.phases: An array of phases that should be run
```
bajtos: This documentation does not match the described interface.

I believe `booters` and `filters` properties were moved from `BootOptions` to `BootExecutionOptions`; and `projectRoot` was moved to `Bootable` interface.

packages/boot/src/booters/base-artifact.booter.ts
```diff
@@ -27,13 +27,13 @@ const debug = debugFactory('loopback:boot:base-artifact-booter');
...
- * @property glob Glob pattern to use to discover artifacts. Takes precedence over (dirs + extensions)
- * @property discovered List of files discovered by the Booter that matched artifact requirements
- * @property classes List of exported classes discovered in the files
+ * - options: Options being used by the Booter
+ * - projectRoot: Project root relative to which all other paths are resovled
+ * - dirs: Directories to look for an artifact in
+ * - extensions: File extensions to look for to match an artifact (this is a convention based booter)
+ * - glob: Glob pattern to use to discover artifacts. Takes precedence over (dirs + extensions)
+ * - discovered: List of files discovered by the Booter that matched artifact requirements
+ * - classes: List of exported classes discovered in the files
```
bajtos: Describing properties in class documentation is weird. Can you move the documentation of individual properties to property-level tsdoc comments please?

packages/openapi-v3/src/generate-schema.ts
```diff
@@ -10,7 +10,7 @@ import {SchemaObject} from '@loopback/openapi-v3-types';
  * parameter's type.
  * `type` and `format` will be preserved if provided in `schema`
  *
- * @private
+ * @internal
```
bajtos: Does `strong-docs` recognize `@internal` annotation? If not, then I think this change is going to break API docs that we are generating right now :(

packages/repository/src/type-resolver.ts
```diff
@@ -14,10 +14,10 @@ import {Class} from './common-types';
  * the actual reference to the class itself until later, when both sides
  * of the relation are created as JavaScript classes.
  *
- * @template Type The type we are resolving, for example `Entity` or `Product`.
+ * @typeparam Type The type we are resolving, for example `Entity` or `Product`.
```
bajtos: Does `strong-docs` recognize `@typeparam` annotation? If not, then I think this change is going to break API docs that we are generating right now :(

packages/boot/src/interfaces.ts
raymondfeng: This can be fixed in a follow-up PR. The original doc follows js style. For TS, such properties should be documented directly at  `BootExecutionOptions` and `Bootable` level.

packages/repository/src/type-resolver.ts
raymondfeng: TypeDoc already uses `@typeparam` - https://github.com/TypeStrong/typedoc/blob/0014556908f807486d2c1d17812ed472113fe522/examples/basic/src/generics.ts#L4

raymondfeng:
To be assured, `npm run apidocs` on master is successful.

PR #2848, opened 2019-05-07
feat(context): add support to sort bindings for `@inject.*` and `ContextView`

raymondfeng:
This PR introduces BindingSorter to help sort bindings discovered from contexts. Some extension points need to control the order of bound extensions. Such APIs can be used to sort life cycle observers
and global interceptors after the PR is landed. 

It has two commits:

1. Add `BindingComparator` interface and a utility to sort by group tag.
2. Add support of `bindingComparator` for `ContextView` and `@inject.*`.

## Checklist

### DIFF
--- a/docs/site/Decorators_inject.md
+++ b/docs/site/Decorators_inject.md
@@ -83,6 +83,23 @@ class MyControllerWithValues {
 }
 ```
 
+To sort matched bindings found by the binding filter function, `@inject` honors
+`bindingComparator` in `metadata`:
+
+```ts
+class MyControllerWithValues {
+  constructor(
+    @inject(binding => binding.tagNames.includes('foo'), {
+      bindingComparator: (a, b) => {
+        // Sort by value of `foo` tag
+        return a.tagMap.foo.localeCompare(b.tagMap.foo);
+      },
+    })
+    public values: string[],
+  ) {}
+}
+```
+
 A few variants of `@inject` are provided to declare special forms of
 dependencies.
 
--- a/packages/context/src/__tests__/unit/binding-sorter.unit.ts
+++ b/packages/context/src/__tests__/unit/binding-sorter.unit.ts
@@ -1,0 +1,155 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {Binding, compareByOrder, sortBindingsByPhase} from '../..';
+
+describe('BindingComparator', () => {
+  const FINAL = Symbol('final');
+  const orderOfPhases = ['log', 'auth', FINAL];
+  const phaseTagName = 'phase';
+  let bindings: Binding<unknown>[];
+  let sortedBindingKeys: string[];
+
+  beforeEach(givenBindings);
+  beforeEach(sortBindings);
+
+  it('sorts by phase', () => {
+    /**
+     * Phases
+     * - 'log': logger1, logger2
+     * - 'auth': auth1, auth2
+     */
+    assertOrder('logger1', 'logger2', 'auth1', 'auth2');
+  });
+
+  it('sorts by phase - unknown phase comes before known ones', () => {
+    /**
+     * Phases
+     * - 'metrics': metrics // not part of ['log', 'auth']
+     * - 'log': logger1
+     */
+    assertOrder('metrics', 'logger1');
+  });
+
+  it('sorts by phase alphabetically without orderOf phase', () => {
+    /**
+     * Phases
+     * - 'metrics': metrics // not part of ['log', 'auth']
+     * - 'rateLimit': rateLimit // not part of ['log', 'auth']
+     */
+    assertOrder('metrics', 'rateLimit');
+  });
+
+  it('sorts by binding order without phase tags', () => {
+    /**
+     * Phases
+     * - '': validator1, validator2 // not part of ['log', 'auth']
+     * - 'metrics': metrics // not part of ['log', 'auth']
+     * - 'log': logger1
+     */
+    assertOrder('validator1', 'validator2', 'metrics', 'logger1');
+  });
+
+  it('sorts by binding order without phase tags', () => {
+    /**
+     * Phases
+     * - '': validator1 // not part of ['log', 'auth']
+     * - 'metrics': metrics // not part of ['log', 'auth']
+     * - 'log': logger1
+     * - 'final': Symbol('final')
+     */
+    assertOrder('validator1', 'metrics', 'logger1', 'final');
+  });
+
+  /**
+   * The sorted bindings by phase:
+   * - '': validator1, validator2 // not part of ['log', 'auth']
+   * - 'metrics': metrics // not part of ['log', 'auth']
+   * - 'rateLimit': rateLimit // not part of ['log', 'auth']
+   * - 'log': logger1, logger2
+   * - 'auth': auth1, auth2
+   */
+  function givenBindings() {
+    bindings = [
+      Binding.bind('logger1').tag({[phaseTagName]: 'log'}),
+      Binding.bind('auth1').tag({[phaseTagName]: 'auth'}),
+      Binding.bind('auth2').tag({[phaseTagName]: 'auth'}),
+      Binding.bind('logger2').tag({[phaseTagName]: 'log'}),
+      Binding.bind('metrics').tag({[phaseTagName]: 'metrics'}),
+      Binding.bind('rateLimit').tag({[phaseTagName]: 'rateLimit'}),
+      Binding.bind('validator1'),
+      Binding.bind('validator2'),
+      Binding.bind('final').tag({[phaseTagName]: FINAL}),
+    ];
+  }
+
+  function sortBindings() {
+    sortBindingsByPhase(bindings, phaseTagName, orderOfPhases);
+    sortedBindingKeys = bindings.map(b => b.key);
+  }
+
+  function assertOrder(...keys: string[]) {
+    let prev: number = -1;
+    let prevKey: string = '';
+    for (const key of keys) {
+      const current = sortedBindingKeys.indexOf(key);
+      expect(current).to.greaterThan(
+        prev,
+        `Binding ${key} should come after ${prevKey}`,
+      );
+      prev = current;
+      prevKey = key;
+    }
+  }
+});
+
+describe('compareByOrder', () => {
+  it('honors order', () => {
+    expect(compareByOrder('a', 'b', ['b', 'a'])).to.greaterThan(0);
+  });
+
+  it('value not included in order comes first', () => {
+    expect(compareByOrder('a', 'c', ['a', 'b'])).to.greaterThan(0);
+  });
+
+  it('values not included are compared alphabetically', () => {
+    expect(compareByOrder('a', 'c', [])).to.lessThan(0);
+  });
+
+  it('null/undefined/"" values are treated as ""', () => {
+    expect(compareByOrder('', 'c')).to.lessThan(0);
+    expect(compareByOrder(null, 'c')).to.lessThan(0);
+    expect(compareByOrder(undefined, 'c')).to.lessThan(0);
+  });
+
+  it('returns 0 for equal values', () => {
+    expect(compareByOrder('c', 'c')).to.equal(0);
+    expect(compareByOrder(null, '')).to.equal(0);
+    expect(compareByOrder('', undefined)).to.equal(0);
+  });
+
+  it('allows symbols', () => {
+    const a = Symbol('a');
+    const b = Symbol('b');
+    expect(compareByOrder(a, b)).to.lessThan(0);
+    expect(compareByOrder(a, b, [b, a])).to.greaterThan(0);
+    expect(compareByOrder(a, 'b', [b, a])).to.greaterThan(0);
+  });
+
+  it('list symbols before strings', () => {
+    const a = 'a';
+    const b = Symbol('a');
+    expect(compareByOrder(a, b)).to.greaterThan(0);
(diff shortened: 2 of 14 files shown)


packages/context/src/__tests__/unit/binding-sorter.unit.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  beforeEach(givenBindings);
+  beforeEach(sortBindings);
+
+  it('sorts by group', () => {
+    assertOrder('logger1', 'logger2', 'auth1', 'auth2');
+  });
+
+  it('sorts by group - unknown group comes before known ones', () => {
+    assertOrder('metrics', 'logger1');
```
jannyHou: a user experience nitpick:

My intuition is the unknown ones should come after known ones and sorted by alphabet by default.

raymondfeng: For global interceptors, unknown groups should come first as they expect less knowledge from other ones.

raymondfeng: If you want to change the behavior, you can have a custom sorter anyway.

jannyHou: I see, fair enough.

raymondfeng:
> If you could please provide an example of when it helps that bindings are sorted. I thought bindings were stored in a hashmap and were resolved quickly. Is it more for a programmer that is writing a specific piece of logic that needs to pass through a list in a particular fashion?

Bindings are stored in a `Map` by keys. No sorting is involved when a binding is resolved by key. The binding sorter is only allowed for `@inject.*` or `ContextView` that takes a `BindingFilter` function to select multiple bindings, such as:

1. Registered life cycle observers
2. Request body parsers
3. Global interceptors

raymondfeng:
> If you could please provide an example of when it helps that bindings are sorted. I thought bindings were stored in a hashmap and were resolved quickly. Is it more for a programmer that is writing a specific piece of logic that needs to pass through a list in a particular fashion?

Bindings are stored in a `Map` by keys. If a binding is resolved by its key, no binding sorter is involved. Please note we have `@inject.*` and `ContextView` that accept a `BindingFilter` function to select multiple bindings. The binding sorter is introduced to sort matched bindings. Examples are:

1. Life cycle observers to be sorted by group
2. Global interceptors to be sorted by group

raymondfeng:
> If you could please provide an example of when it helps that bindings are sorted. I thought bindings were stored in a hashmap and were resolved quickly. Is it more for a programmer that is writing a specific piece of logic that needs to pass through a list in a particular fashion?

Test

packages/context/src/__tests__/unit/binding-sorter.unit.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it('sorts by group alphabetically without ordered group', () => {
+    assertOrder('metrics', 'rateLimit');
+  });
+
+  it('sorts by binding order without group tags', () => {
+    assertOrder('validator1', 'validator2', 'logger1');
```
b-admike: I'm a bit confused by the name of the test. How is binding order determined in this case? Isn't it just using default group tag of an empty string for `validator1` and `validator2` and alphabetically sorting them followed by the named group tag bindings?

raymondfeng: Yes, without an explicit `group` tag, the group name is `''` and they are sorted based on that.

raymondfeng: I added more comments to the tests to make it easier to understand.

packages/context/src/inject.ts
```diff
@@ -112,6 +117,9 @@ export function inject(
     resolve = resolveValuesByFilter;
   }
   const injectionMetadata = Object.assign({decorator: '@inject'}, metadata);
+  if (injectionMetadata.bindingSorter && !resolve) {
+    throw new Error('Binding sorter is only allowed with a binding filter');
```
b-admike: For my understanding, what's the rationale behind not allowing binding sorting without binding filter?

packages/context/src/context-view.ts
```diff
@@ -158,14 +163,23 @@ export class ContextView<T = unknown> extends EventEmitter
...
  * @param ctx Context object
  * @param bindingFilter A function to match bindings
+ * @param bindingSorter A function to sort matched bindings
  * @param session Resolution session
  */
 export function createViewGetter<T = unknown>(
   ctx: Context,
   bindingFilter: BindingFilter,
+  bindingSorterOrSession?: BindingSorter | ResolutionSession,
   session?: ResolutionSession,
```
bajtos: This API allows the following invocation which does not seem valid to me:

```ts
createViewGetter<T>(ctx, filter, session1, session2);
```

Please use function overloads to better describe the actual API.

```ts
export function createViewGetter<T = unknown>(
  ctx: Context,
  bindingFilter: BindingFilter,
  bindingSorter: BindingSorter,
  session?: ResolutionSession,
): Getter<T[]>;

export function createViewGetter<T = unknown>(
  ctx: Context,
  bindingFilter: BindingFilter,
  session?: ResolutionSession,
): Getter<T[]>;

export function createViewGetter<T = unknown>(
  ctx: Context,
  bindingFilter: BindingFilter,
  bindingSorterOrSession?: BindingSorter | ResolutionSession,
  session?: ResolutionSession,
): Getter<T[]> {
  // the implementation
}
```

packages/context/src/inject.ts
bajtos: When there is no binding filter, then the injection is resolved with a single binding (as described by the provided binding key). It does not make sense to sort a single binding.

At least that's my understanding.

docs/site/Decorators_inject.md
```diff
@@ -83,6 +83,23 @@ class MyControllerWithValues {
...
 ```
 
+To sort matched bindings found by the binding filter function, `@inject` honors
+`bindingSorter` in `metadata`:
+
+```ts
+class MyControllerWithValues {
+  constructor(
+    @inject(binding => binding.tagNames.includes('foo'), {
+      bindingSorter: (a, b) => {
```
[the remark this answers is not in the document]
raymondfeng: I renamed it to bindingComparator.

packages/context/src/binding-sorter.ts
```diff
@@ -0,0 +1,64 @@
...
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Binding} from './binding';
+
+/**
+ * Compare function to sort bindings
+ */
+export type BindingSorter = (
```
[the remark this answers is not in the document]
raymondfeng: `BindingComparator` is used now.

docs/site/Decorators_inject.md
```diff
@@ -83,6 +83,23 @@ class MyControllerWithValues {
...
+        // Sort by value of `foo` tag
+        return a.tagMap.foo.localCompare(b.tagMap.foo);
```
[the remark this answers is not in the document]
raymondfeng: Good catch!

packages/context/src/__tests__/unit/context-view.unit.ts
```diff
@@ -169,14 +195,14 @@ describe('ContextView', () => {
       server
         .bind('bar')
         .toDynamicValue(() => Promise.resolve('BAR'))
-        .tag('foo', 'bar')
+        .tag({foo: 'a'})
```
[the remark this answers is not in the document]
raymondfeng: I fixed by adding `{group: 'a'}` for sorting.

packages/context/src/binding-sorter.ts
```diff
@@ -0,0 +1,64 @@
...
+ * denoted by `orderedGroups`.
+ * 3. If a binding's `group` does not exist in `orderedGroups`, it comes before
+ * the one exists in `orderedGroups`.
+ * 4. If both bindings have `group` values outside of `orderedGroups`, they are
+ * ordered by group names alphabetically.
+ *
+ * @param groupTagName Name of the tag for group
+ * @param orderedGroups An array of group names as predefined orders
+ */
+export function createSorterByGroup(
```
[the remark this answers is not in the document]
raymondfeng: I rename it to `compareBindingsByGroup`.

raymondfeng:
> How do you envision integration & interaction of this new feature with Extension Point/Extension pattern? Can code receiving a list of extension bindings use this new feature out of the box?

Yes, I'll refactor some of the extension points to use this new feature in two flavors:

1. Use `bindingComparator` for `@extensions`
2. Use `bindingComparator` for `ContextView` or `@inject.view`

packages/context/src/binding-sorter.ts
```diff
@@ -0,0 +1,64 @@
...
+  };
+}
+
+/**
+ * Sort bindings by tagged group name
+ * @param bindings An array of bindings
+ * @param groupTagName Tag name for group
+ * @param orderedGroups An array of group names as predefined orders
+ */
+export function sortBindingsByGroup(
```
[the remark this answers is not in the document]
raymondfeng: We now have `compareBindingsByGroup` to create a compare function and `sortBindingsByGroup` to sort bindings.

docs/site/Decorators_inject.md
```diff
@@ -83,6 +83,23 @@ class MyControllerWithValues {
...
 ```
 
+To sort matched bindings found by the binding filter function, `@inject` honors
+`bindingComparator` in `metadata`:
+
+```ts
+class MyControllerWithValues {
+  constructor(
+    @inject(binding => binding.tagNames.includes('foo'), {
+      bindingComparator: (a, b) => {
```
bajtos: I like the new name `bindingComparator` 👍

packages/context/src/binding-sorter.ts
```diff
@@ -0,0 +1,124 @@
...
+ * 3. If a binding's `group` does not exist in `orderedGroups`, it comes before
+ * the one with `group` exists in `orderedGroups`.
+ * 4. If both bindings have `group` value outside of `orderedGroups`, they are
+ * ordered by group names alphabetically and symbol values come before string
+ * values.
+ *
+ * @param groupTagName Name of the binding tag for group
+ * @param orderedGroups An array of group names as the predefined order
+ */
+export function compareBindingsByGroup(
```
bajtos: I find the `ByGroup` part problematic.

So far, bindings have a `key`, `namespace` and zero or more `tag`s. There is no concept of a group AFAICT. As a result, the name `compareBindingsByGroup` does not make it clear what rule is used for comparison. 

What's more, the tag name can be further customized to a different value than `group`, in which case the code makes even less sense:

```ts
compareBindingsByGroup('priority', ['high', 'medium', 'low']);
```

Let's find a name that better reveals the intent please. Personally, I would use `compareBindingsByTag`.

I am also not very happy about the term "group". In my mind, groups are about grouping several things into one group, there is no order involved. In LB3, we were using the term `phase` to sort middleware handlers, would it make sense to use that term here too? 

```ts
export function compareBindingsByGroup(
  tagName: string = 'phase',
  orderedPhases: (string | symbol)[] = [],
): BindingComparator {
  // ...
}
```

packages/context/src/binding-sorter.ts
```diff
@@ -0,0 +1,124 @@
...
+  groupTagName: string = 'group',
+  orderedGroups: (string | symbol)[] = [],
```
bajtos: "orderedGroups" feels weird when using a tag name different from "group". How about `orderedTagValues`?

raymondfeng: Switched to `compareBindingsByTag`.

packages/rest/src/body-parsers/body-parser.ts
```diff
@@ -32,7 +34,7 @@ export class RequestBodyParser {
   readonly parsers: BodyParser[];
 
   constructor(
-    @inject.tag(REQUEST_BODY_PARSER_TAG, {optional: true})
+    @inject(filterByTag(REQUEST_BODY_PARSER_TAG), {optional: true})
```
bajtos: What is the reasoning for this change? Is it just a stylistic refactoring, or does it introduce any changes in the externally-observed behavior, e.g. the order in which the parsers are invoked?

I am ok with the change, but would like to better understand the consequences.

packages/context/src/context-view.ts
```diff
@@ -155,17 +160,54 @@ export class ContextView<T = unknown> extends EventEmitter
...
+): Getter<T[]>;
+
+/**
+ * Create a context view as a getter
+ * @param ctx Context object
+ * @param bindingFilter A function to match bindings
+ * @param bindingComparatorOrSession A function to sort matched bindings or
+ * resolution session if the comparator is not needed
+ * @param session Resolution session if the comparator is provided
+ */
```
bajtos: I think the actual implementation should not have any tsdoc comments. We have already described the two flavors recognized by the compiler when validating places calling this method. WDYT?

packages/context/src/inject.ts
```diff
@@ -52,6 +53,14 @@ export interface InjectionMetadata extends ResolutionOptions {
    * It's usually set by the decorator implementation.
    */
   decorator?: string;
+  /**
+   * Control if the dependency is optional, default to false
+   */
+  optional?: boolean;
```
bajtos: I believe the flag `optional` is already inherited from `ResolutionOptions`. Can you remove the duplicate definition from `InjectionMetadata` please?

packages/rest/src/body-parsers/body-parser.ts
raymondfeng: It's stylistic. IIRC, we would like to deprecate `@inject.tag` in the future and use `@inject` with a filter function instead.

packages/context/src/context-view.ts
raymondfeng: The comments are for maintainers only. I assume strong-docs or tsdocs only expose the method signatures (not implementation).

PR #2854, opened 2019-05-09
fix(rest-explorer): exclude basePath from /openapi URL

bajtos:
Endpoints serving OpenAPI spec ignore basePath setting, i.e. even if basePath is set to `/api`, the spec is served at `/openapi.json`.

This pull request is fixing the problem and supersedes https://github.com/strongloop/loopback-next/pull/2554.

/cc @gordancso @dkrantsberg

## Checklist


bajtos:
On the second thought, this is not a correct fix. While it solves the problem when `basePath` is set, it breaks the scenario when the entire LB4 app is mounted in another Express app. I need to rework the patch.

PR #2859, opened 2019-05-09
refactor: recover tests for auth action

jannyHou:
### Description

Fixes https://github.com/strongloop/loopback-next/issues/2467

See discussion in the draft PR https://github.com/strongloop/loopback-next/pull/2822.

- I deprecated the old strategy adapter in this PR. Will implement different approaches of the new strategy adapter in story #2311.
- The deleted test file will also be added in story #2311, as the acceptance test in the new adapter package.

And since a new package will be created for the new adapter(s), I will work on the spike https://github.com/strongloop/loopback-next/issues/2676#issuecomment-488746964 first, then create the adapter's package in `loopback-labs`.


packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
```diff
@@ -1,208 +0,0 @@
-// Copyright IBM Corp. 2019. All Rights Reserved.
```
b-admike: Are we removing these tests because they are already covered/moved to some other place (Perhaps @emonddr's PR had them)?

packages/authentication/src/__tests__/unit/fixtures/mock-strategy-passport.ts
```diff
@@ -0,0 +1,67 @@
...
+      request.headers.testState === 'error'
+    ) {
+      this.returnError('unexpected error');
+      return;
+    }
+    process.nextTick(this.returnMockUser.bind(this));
+  }
+
+  returnMockUser() {
+    this.success(this.mockUser);
```
b-admike: Where is `success` defined?

packages/authentication/src/__tests__/unit/fixtures/mock-strategy.ts
```diff
@@ -3,26 +3,35 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Strategy, AuthenticateOptions} from 'passport';
-import {Request} from 'express';
+import {Request} from '@loopback/rest';
+import {AuthenticationStrategy, UserProfile} from '../../../types';
+
+class AuthenticationError extends Error {
```
b-admike: Is `Error` imported from somewhere or a general interface?

packages/authentication/src/__tests__/acceptance/basic-auth.acceptance.ts
jannyHou: @b-admike These tests are written using a passport based strategy + passport adapter. While the old adapter will be deprecated and replaced by a new one to be created in a new module, so **acceptance test for the passport strategy** should be rewritten in that coming new module.

The new auth sysmte that Dom created is built with the extension point, so that the new acceptance tests also reflect that change:
see https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
and https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts 

> Are we removing these tests because they are already covered/moved to some other place
So to answer your question, yes or no, the acceptance tests for the new auth system is covered. The new adapter and its corresponding acceptance tests haven't created. The file you are looking at is removed because of the old adapter doesn't work with the new system anymore.

b-admike: Makes sense, thanks for the in-depth explanation @jannyHou :-).

packages/authentication/src/__tests__/unit/fixtures/mock-strategy-passport.ts
jannyHou: @b-admike it's a function from the passport strategy, see https://github.com/jaredhanson/passport/blob/master/lib/middleware/authenticate.js#L212

packages/authentication/src/__tests__/unit/fixtures/mock-strategy.ts
jannyHou: Ah, you can follow the chat in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32176, the `Error` interface doesn't have a `code` or `statusCode` property, and using `ErrnoException` violates everywhere that expects an `Error` object, so I created my own Error class to specify the `code`.

packages/authentication/src/__tests__/unit/strategy-adapter.unit.ts
```diff
@@ -16,11 +16,15 @@ describe('Strategy Adapter', () => {
     it('calls the authenticate method of the strategy', async () => {
       let calledFlag = false;
       // TODO: (as suggested by @bajtos) use sinon spy
-      class Strategy extends MockStrategy {
+      class Strategy extends MockPassportStrategy {
         // override authenticate method to set calledFlag
         async authenticate(req: Request, options?: AuthenticateOptions) {
           calledFlag = true;
-          await MockStrategy.prototype.authenticate.call(this, req, options);
+          await MockPassportStrategy.prototype.authenticate.call(
```
raymondfeng: Please use `await super.authenticate(...)` instead.

packages/authentication/src/__tests__/unit/strategy-adapter.unit.ts
```diff
@@ -16,11 +16,11 @@ describe('Strategy Adapter', () => {
     it('calls the authenticate method of the strategy', async () => {
       let calledFlag = false;
       // TODO: (as suggested by @bajtos) use sinon spy
-      class Strategy extends MockStrategy {
+      class Strategy extends MockPassportStrategy {
         // override authenticate method to set calledFlag
         async authenticate(req: Request, options?: AuthenticateOptions) {
           calledFlag = true;
-          await MockStrategy.prototype.authenticate.call(this, req, options);
+          await super.authenticate.call(this, req, options);
```
raymondfeng: Just `super.authenticate(req, options);`.

jannyHou: I was dumm 😞

emonddr:
Great job, @jannyHou

PR #1205, opened 2018-03-28
feat: add authorization component

raymondfeng:
Connect to #538 

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `packages/example-*` were updated

PR #2866, opened 2019-05-10
feat(context): add more getters for InvocationContext

raymondfeng:
Add more getters for `InvocationContext` as discovered during the development of `authorization` component using `Interceptor`.

See https://github.com/strongloop/loopback-next/pull/1205 

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/invocation-context.unit.ts
+++ b/packages/context/src/__tests__/unit/invocation-context.unit.ts
@@ -1,0 +1,107 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/context
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {Context, InvocationContext} from '../..';
+
+describe('InvocationContext', () => {
+  let ctx: Context;
+  let invocationCtxForGreet: InvocationContext;
+  let invocationCtxForCheckName: InvocationContext;
+  let invalidInvocationCtx: InvocationContext;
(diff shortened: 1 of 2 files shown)


packages/context/src/interceptor.ts
```diff
@@ -104,6 +104,33 @@ export class InvocationContext extends Context {
...
+  }
+
+  /**
+   * Description of the invocation
+   */
+  get description() {
+    return `InvocationContext(${this.name}): ${this.targetName}`;
+  }
+
+  toString() {
```
jannyHou: Isn't `InvocationContext(${this.name}): ${this.targetName}` already a string? What would be the difference between description and `toString()`?

packages/context/src/__tests__/unit/invocation-context.unit.ts
```diff
@@ -0,0 +1,95 @@
...
+        invocationCtxForCheckName.name
+      }): MyController.checkName`,
+    );
+  });
+
+  it('throws error if method does not exist', () => {
+    expect(() => invalidInvocationCtx.assertMethodExists()).to.throw(
+      'Method MyController.prototype.invalid-method not found',
+    );
+  });
```
jannyHou: nitpick: maybe test the error catch for static method as well.

packages/context/src/interceptor.ts
raymondfeng: `description` can be used in debugging, such as `invocationCtx.description`. `toString` overrides the base method so that it uses `description` as the string representation of the invocation context object.

packages/context/src/__tests__/unit/invocation-context.unit.ts
raymondfeng: added

Issue #2541, opened 2019-03-05
Inject repository in oauth provider


DatGuyJonathan:
## Description / Steps to reproduce / Feature proposal

I'm trying to use a couple of repositories in an oauth2 flow. The setup is similar to https://github.com/strongloop/loopback-next/issues/1835 where I have a `verify` function that I want to try calling the repositories. The error is:

`Error: The key repositories.UserInfoRepository was not bound to any value`

The organization and user models have a **has-many/belongs-to** relation so creating a user with the org repository is supposed to invoke the user repository. I followed https://loopback.io/doc/en/lb4/HasMany-relation.html to setup the models & repos. I _think_ I have the repositories and models wired up correctly because I can move the logic into a controller and it works fine.

Any ideas what I need to bind/inject?

```
export class MyAuthenticationProvider
  implements Provider<PassportAuthenticationStrategy> {
  constructor(
    @repository(OrganizationRepository)
    private organizationRepository: OrganizationRepository,
    @inject(SecretBindings.SECRET_SERVICE)
    private secretService: SecretService
  ) {}

  public value(): ValueOrPromise<PassportAuthenticationStrategy> {
    const callbackURL = process.env.DNS_NAME
      ? process.env.DNS_NAME
      : 'http://localhost:3000';
    const clientID = process.env.CLIENT_ID
      ? process.env.CLIENT_ID
      : 'CLIENT_ID_TEST';
    const clientSecret = process.env.CLIENT_SECRET
      ? process.env.CLIENT_SECRET
      : 'CLIENT_SECRET_TEST';
    return new MyOAuth2Strategy(
      {
        authorizationURL:
          '<insert auth url>',
        tokenURL: '<insert token url>,
        clientID,
        clientSecret,
        callbackURL: `${callbackURL}/auth/callback`
      },
      this.verify.bind(this)
    );
  }

  private async verify(
    accessToken: string,
    refreshToken: string,
    user: UserInfo,
    callback: VerifyCallback
  ): Promise<void> {
    if (accessToken && refreshToken) {
      user.refreshToken = refreshToken;

      await this.organizationRepository.create({
        id: user.organizationId
      });

      user.token = await this.secretService.encrypt(accessToken);
      user.refreshToken = await this.secretService.encrypt(user.refreshToken);
      await this.organizationRepository.users(user.organizationId).create(user); // THE ERROR IS HERE
      return callback(null, user);
    }
    return callback(new Error('Invalid tokens'), user);
  }
}
```

[...]

PR #2872, opened 2019-05-11
feat(context): leave local bindings and parent unchanged in context.close()

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/2541

In some cases, the request context is still in use when the http response is finished. This PR changes `close` to only clear subscriptions and remove itself from the parent context so that the context is ready
for GC without destroying bindings.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/integration/request-context.integration.ts
+++ b/packages/rest/src/__tests__/integration/request-context.integration.ts
@@ -9,13 +9,13 @@ import {
   createRestAppClient,
   expect,
   givenHttpServerConfig,
-  supertest,
   httpsGetAsync,
+  supertest,
 } from '@loopback/testlab';
 import * as express from 'express';
 import {RequestContext} from '../../request-context';
 import {RestApplication} from '../../rest.application';
-import {RestServerConfig} from '../../rest.server';
+import {RestServer, RestServerConfig} from '../../rest.server';
 import {DefaultSequence} from '../../sequence';
 
 let app: RestApplication;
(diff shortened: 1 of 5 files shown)


packages/context/src/interceptor.ts
```diff
@@ -390,7 +390,7 @@ export function invokeMethodWithInterceptors(
       const interceptors = invocationCtx.loadInterceptors();
       return invokeInterceptors(invocationCtx, interceptors);
     },
-    () => invocationCtx.close(),
+    () => invocationCtx.clearSubscriptions(),
```
bajtos: Please add a test that fails before this change is made and passes with the fix in place.

Ideally, I'd like to see acceptance-level test demonstrating a somewhat realistic scenario that is broken now.

packages/rest/src/request-context.ts
```diff
@@ -95,7 +95,9 @@ export class RequestContext extends Context implements HandlerContext {
     super(parent, name);
     this._setupBindings(request, response);
     onFinished(this.response, () => {
-      this.close();
+      // See https://github.com/strongloop/loopback-next/issues/2541
+      // We call `clearSubscriptions` instead of `close`
+      this.clearSubscriptions();
```
bajtos: Please add an acceptance-level test demonstrating a somewhat realistic scenario, the test should be failing against the current master and pass with your changes in place.

raymondfeng:
> I am not able to fully understand the difference between clearSubscriptions and close. Could you please explain?

See the description of the PR.

packages/rest/src/__tests__/integration/request-context.integration.ts
```diff
@@ -141,5 +157,9 @@ function contextObservingHandler(
   _sequence: DefaultSequence,
 ) {
   observedCtx = ctx;
+  // Add a subscriber to verify `close()`
+  ctx.subscribe(() => {});
```
emonddr: @raymondfeng 
Just wondering how an empty function can be passed when the function type is:
```
export type ContextObserverFn = (
  eventType: ContextEventType,
  binding: Readonly<Binding<unknown>>,
  context: Context,
) => ValueOrPromise<void>;

```
If you could clarify. Thanks

raymondfeng: The empty function is compatible with `ContextObserverFn` as it does not care about the args and just return `void`.

PR #2879, opened 2019-05-13
docs: add stub service section

### LINKED ISSUE #2101: [Docs] "Testing your application" page contains beta release information
Originated from the docs cleanup spike: https://github.com/strongloop/loopback-next/issues/1908

The ["Testing your application" page](https://loopback.io/doc/en/lb4/Testing-your-application.html) seems to be outdated.  
- ["Create a stub service" section](https://loopback.io/doc/en/lb4/Testing-your-application.html#create-a-stub-service)
   - a WIP note
   - "The initial beta release does not include Services as a first-class feature." <-- we are no longer in beta release. This should be updated. 
- ["Perform an auto-generated smoke test of your REST API" section ](https://loopback.io/doc/en/lb4/Testing-your-application.html#perform-an-auto-generated-smoke-test-of-your-rest-api)
  - it says "It will be revisited after our MVP release".  

@bajtos, I don't have enough knowledge in this area to have a more complete acceptance criteria.  It seems like this page needs to be reviewed more thoroughly, rather than just catching the obvious. Thanks.

b-admike:
Fixes #2101 

## Checklist

### DIFF
--- a/docs/site/Testing-your-application.md
+++ b/docs/site/Testing-your-application.md
@@ -818,8 +818,10 @@ describe('API specification', () => {
 
 {% include important.html content=" The top-down approach for building LoopBack
 applications is not yet fully supported. Therefore, the code outlined in this
-section is outdated and may not work out of the box. It will be revisitedafter
-our MVP release.
+section is outdated and may not work out of the box. Check out
+https://github.com/strongloop/loopback-next/issues/1882 for the epic tracking
+the feature and [OpenAPI generator](OpenAPI-generator.md) page for artifact
+generation from OpenAPI specs.
 " %}
 
 The formal validity of your application's spec does not guarantee that your
--- a/docs/site/Testing-your-application.md
+++ b/docs/site/Testing-your-application.md
@@ -408,7 +408,7 @@ Verify how was the stubbed method executed at the end of your unit test (in the
 ```ts
 // expect that repository.find() was called with the first
 // argument deeply-equal to the provided object
-sinon.assert.calledWithMatch({where: {id: 1}});
+sinon.assert.calledWithMatch(findStub, {where: {id: 1}});
 ```
 
 See [Unit test your controllers](#unit-test-your-controllers) for a full
@@ -416,9 +416,64 @@ example.
 
 #### Create a stub Service
 
-{% include content/tbd.html %}
+If your controller relies on service proxy for service oriented backends such as
+REST APIs, SOAP Web Services, or gRPC microservices, then we can create stubs
+for the service akin to the steps outlined in the above
+[stub repository](#create-a-stub-repository) section. Consider a dependency on a
+`GeoCoder` service that relies on a remote REST API for returning coordinates
+for a specific address.
 
-The initial beta release does not include Services as a first-class feature.
+```ts
+export interface GeocoderService {
+  geocode(address: string): Promise<GeoPoint[]>;
+}
+```
+
+The first step is to create a mocked instance of the `GeocoderService` API and
+configure its `geocode` method as a Sinon stub:
+
+```ts
+describe('GeocoderController', () => {
+  let geoService: GeoCoderService;
+  let geocode: sinon.SinonStub;
+
+  beforeEach(givenMockGeoCoderService);
+
+  // your unit tests
+
+  function givenMockGeoCoderService() {
+    // this creates a stub with GeocoderService API
+    // in a way that allows the compiler to verify type correctness
+    geoService = {geocode: sinon.stub()};
+
+    // this creates a reference to the stubbed "geocode" method
+    // because "geoService.geocode" has type from GeocoderService
+    // and does not provide Sinon APIs
+    geocode = geoService.geocode as sinon.SinonStub;
+  }
+});
+```
+
+Afterwards, we can configure the `geocode` stub's behaviour before the `act`
+phase of our test(s):
+
+```ts
+// geoService.geocode() will return a promise that
+// will be resolved with the provided array
+geocode.resolves([<GeoPoint>{y: 41.109653, x: -73.72467}]);
+```
+
+Lastly, we'll verify how the stub was executed:
+
+```ts
+// expect that geoService.geocode() was called with the first
+// argument equal to the provided address string
+sinon.assert.calledWithMatch(geocode, '1 New Orchard Road, Armonk, 10504');
+```
+
+Check out
+[TodoController unit tests](https://github.com/strongloop/loopback-next/blob/bd0c45033503f631a533ad6176620354d9cf6768/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts#L53-L71)
+illustrating the above points in action for more information.
 
 ### Unit test your Controllers
 
--- a/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts
+++ b/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts
@@ -67,6 +67,7 @@ describe('TodoController', () => {
 
       expect(result).to.eql(expected);
       sinon.assert.calledWith(create, input);
+      sinon.assert.calledWith(geocode, input.remindAtAddress);
     });
   });
 


docs/site/Testing-your-application.md
```diff
@@ -408,17 +408,64 @@ Verify how was the stubbed method executed at the end of your unit test (in the
...
+Configure the `geocode` method as a stub and its behaviour before your test(s):
+
+```ts
+// geoService.geocode() will return a promise that
+// will be resolved with the provided array
+const geocodeStub = geoService.geocode as sinon.SinonStub;
+geocodeStub.resolves([{y: 0, x: 0}]);
+```
+
+Verify how the stub was executed:
```
jannyHou: nitpick: `//` before the sentence.

docs/site/Testing-your-application.md
```diff
@@ -408,17 +408,64 @@ Verify how was the stubbed method executed at the end of your unit test (in the
...
+
+```ts
+// expect that geoService.geocode() was called with the first
+// argument equal to the provided address string
+sinon.assert.calledWithMatch(geocodeStub, '1600 Pennsylvania Ave NW');
```
jannyHou: Hmm, where is `'1600 Pennsylvania Ave NW'` used in the previous code?

nabdelgadir:
I tried it with the following setup:

```ts
import {GeocoderService} from './geocoder.service';
const GeoPoint = require('geopoint');

export class DummyGeocoder implements GeocoderService {
  // in practice, the constructor could take an API key to authenticate with
  // the service proxy provider
  constructor() {}

  geocode(address: string) {
    // this would call the remote API for the real coordinates
    return new GeoPoint(address);
  }
}
```

Test:
```ts
import {DummyGeocoder} from '../../../services/dummy.service';
import {sinon} from '@loopback/testlab';

describe('GeoCoderController', () => {
  let geoService: DummyGeocoder;
  beforeEach(givenStubbedService);

  // your unit tests

  it('', () => {
    const geocodeStub = geoService.geocode as sinon.SinonStub;
    geocodeStub.resolves([{y: 0, x: 0}]);

    sinon.assert.calledWithMatch(geocodeStub, '1600 Pennsylvania Ave NW');
  });

  function givenStubbedService() {
    geoService = sinon.createStubInstance(DummyGeocoder);
  }
});
```

But it fails with `AssertError: expected geocode to be called with match`. `geocode` never gets called, so is there something wrong with the setup?

docs/site/Testing-your-application.md
b-admike: Good question. It's not apparent here, but the way we structure our test is to `arrange, act, then assert`, so somewhere in the `act` phase, we'd call a controller method that eventually calls the `geoService.geocode()` with the given address string. I don't show that here because I thought it would be apparent, but I can add a note for that to make it clear.

b-admike:
> I tried it with the following setup:
> 
> ```ts
> import {GeocoderService} from './geocoder.service';
> const GeoPoint = require('geopoint');
> 
> export class DummyGeocoder implements GeocoderService {
>   // in practice, the constructor could take an API key to authenticate with
>   // the service proxy provider
>   constructor() {}
> 
>   geocode(address: string) {
>     // this would call the remote API for the real coordinates
>     return new GeoPoint(address);
>   }
> }
> ```
> 
> Test:
> 
> ```ts
> import {DummyGeocoder} from '../../../services/dummy.service';
> import {sinon} from '@loopback/testlab';
> 
> describe('GeoCoderController', () => {
>   let geoService: DummyGeocoder;
>   beforeEach(givenStubbedService);
> 
>   // your unit tests
> 
>   it('', () => {
>     const geocodeStub = geoService.geocode as sinon.SinonStub;
>     geocodeStub.resolves([{y: 0, x: 0}]);
> 
>     sinon.assert.calledWithMatch(geocodeStub, '1600 Pennsylvania Ave NW');
>   });
> 
>   function givenStubbedService() {
>     geoService = sinon.createStubInstance(DummyGeocoder);
>   }
> });
> ```
> 
> But it fails with `AssertError: expected geocode to be called with match`. `geocode` never gets called, so is there something wrong with the setup?

Thanks for trying it out @nabdelgadir. This is missing the `act` stage of the test between the `arrange`/ `setup` of the test and the assertion for the expected result. A good example to that would be https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts#L66. That would invoke the `geocode` method with the address string and then the assertion should work. I'll add an assertion in that test and refer to it in the docs.

docs/site/Testing-your-application.md
b-admike: This is explaining the step that is about to happen like `Configure the `geocode` method as a stub and its behaviour before your test(s):` above, so it's not meant to be a comment. WDYT @jannyHou?

docs/site/Testing-your-application.md
```diff
@@ -408,17 +408,68 @@ Verify how was the stubbed method executed at the end of your unit test (in the
...
+Verify how the stub was executed:
+
+```ts
+// expect that geoService.geocode() was called with the first
+// argument equal to the provided address string
+sinon.assert.calledWithMatch(geocodeStub, '1 New Orchard Road, Armonk, 10504');
+```
+
+Check out
+[this](<(https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts#L53-L71)>)
```
dhmlau: From the github view, the `this` doesn't show the link.  do you mean:
```
[this](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts#L53-L71)
```


Also, I'm not sure specifying the line number would be a good solution in general, because they tend to get outdated easily.  But I'm not sure if specifying the test case name would be a good alternative either.  Just wanna bring this up.  Thanks.

docs/site/Testing-your-application.md
```diff
@@ -408,17 +408,68 @@ Verify how was the stubbed method executed at the end of your unit test (in the
...
+describe('GeoCoderController', () => {
+  let geoService: DummyGeocoder;
+  beforeEach(givenStubbedService);
+
+  // your unit tests
+
+  function givenStubbedService() {
+    geoService = sinon.createStubInstance(DummyGeocoder);
+  }
+});
```
bajtos: Your example is showing a different approach from what we have in our Todo example. I find it impractical that we have to create a `DummyGeocoder` class, when all we want is a mocked instance with `GeocoderService` API.

Here is what we do in our example - see https://github.com/strongloop/loopback-next/blob/0cccae6e2a7a9eb08feee6ab57730e8c10e4269c/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts

```ts
describe('TodoController', () => {
  let geoService: GeocoderService;
  let geocode: sinon.SinonStub;
 
  beforeEach(resetRepositories);

  // (the tests)

  function resetRepositories() {
    // ...

    // this creates a stub with GeocoderService API
    // in a way allowing the compiler to verify type correctness
    geoService = {geocode: sinon.stub()};

    // this creates a reference to the stubbed "geocode" method
    // because "geoService.gecode" has type from GeocoderService
    // and does not provide Sinon APIs
    geocode = geoService.geocode as sinon.SinonStub;

    controller = new TodoController(todoRepo, geoService);
  }
});
```

Having said that, your approach has its benefits too. If we use `createStubInstance` from testlab instead of `sinon.createStubInstance`, then we don't have to cast stubbed methods to `sinon.SinonStub`.

```ts
let geoService: StubbedInstanceWithSinonAccessor<GeocoderService>;
// ...
geoService = createStubInstance(DummyGeocoder);
// ...
geoService.stubs.geocode.resolves([<GeoPoint>{y: 41.109653, x: -73.72467}]);
```

See how TodoController tests are stubbing TodoRepository and its methods.

docs/site/Testing-your-application.md
bajtos: 1) Let's use a more descriptive link name than "this". For example, "TodoController unit tests".

2) Either point to the entire file in `master` branch (`https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts`)  or lock down the commit sha in a link pointing to line numbers (`https://github.com/strongloop/loopback-next/blob/bd0c45033503f631a533ad6176620354d9cf6768/examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts#L53-L71`).

b-admike: Yes I was looking for the commit sha, thank you both for pointing this out :-).

docs/site/Testing-your-application.md
b-admike: Yea I looked at those examples (and https://github.com/strongloop/loopback-next/blob/master/packages/service-proxy/src/__tests__/acceptance/service.mixin.acceptance.ts#L18-L28) and decided to make it similar to how we were stubbing repositories. I don't mind changing it back to a mocked service of the `GeoService` API, especially since I'm pointing back to the test case in our Todo example.

bajtos: I think my main concern is about consistency between the code in our examples and the documentation. If you prefer the approach proposed in this PR, then feel free to go with it, just update the examples to use it too.

b-admike:
@bajtos @dhmlau @jannyHou @nabdelgadir I think I've applied all your feedback. LGTY now?

docs/site/Testing-your-application.md
```diff
@@ -408,17 +408,72 @@ Verify how was the stubbed method executed at the end of your unit test (in the
...
+for a specific address.
 
-The initial beta release does not include Services as a first-class feature.
+```ts
+export interface GeocoderService {
+  geocode(address: string): Promise<GeoPoint[]>;
+}
+```
+
+The first step is to create a mocked instance of the `GeoCoderService` API and
```
nabdelgadir: Nitpick: `GeoCoderService` -> `GeocoderService` (there's a few other instances below)

docs/site/Testing-your-application.md
jannyHou: I see, thank you for the explanation and good document 👍

Issue #2629, opened 2019-03-22
Allow controllers to provide definitions of models referenced in operation spec


bajtos:
## Description / Steps to reproduce / Feature proposal

Allow controller methods to reference model schema directly via OpenAPI spec, without requiring them to use `x-ts-type` extension. Extend support for propagating definitions of references models, so that this feature is available to all operations using `$ref` instead of `x-ts-type`.

Consider the following test in `@loopback/openapi-v3`:

```ts
  it('allows operations to provide definition of referenced models', () => {
    class MyController {
      @get('/todos', {
        responses: {
          '200': {
            description: 'Array of Category model instances',
            content: {
              'application/json': {
                schema: {
                  $ref: '#/definitions/Todo',
                  definitions: {
                    Todo: {
                      title: 'Todo',
                      properties: {
                        title: {type: 'string'},
                      },
                    },
                  },
                },
              },
            },
          },
        },
      })
      async find(): Promise<object[]> {
        return []; // dummy implementation, it's never called
      }
    }

    const spec = getControllerSpec(MyController);
    const opSpec: OperationObject = spec.paths['/todos'].get;
    const responseSpec = opSpec.responses['200'].content['application/json'];
    expect(responseSpec.schema).to.deepEqual({
     $ref: '#/definitions/Todo',
    });

    const globalSchemas = (spec.components || {}).schemas;
    expect(globalSchemas).to.deepEqual({
      Todo: {
        title: 'Todo',
        properties: {
          title: {
            type: 'string',
          },
        },
      },
    });
  });
```

## Current Behavior

The test fails:

```
 + expected - actual

       {
         "$ref": "#/components/schemas/Todo"
      -  "definitions": {
      -    "Todo": {
      -      "properties": {
      -        "title": {
      -          "type": "string"
      -        }
      -      }
      -      "title": "Todo"
      -    }
      -  }
       }
```

## Expected Behavior

The test passes.

See the spike #2592 and especially the commit https://github.com/strongloop/loopback-next/pull/2592/commits/d93dff7a264f29cf234b6ad34074743e277555d1  for more details . Please note the commit shows changes before the review feedback was applied. You need to review the final pull request diff of the PoC before copying any code over.

## Acceptance criteria

- [ ] Make the changes necessary to make the above test pass
- [ ] Add additional test coverage at unit-test and integration level as necessary

emonddr:
@bajtos , we would like to know if the `definitions` is actually needed:

```
        responses: {
[...]

PR #2898, opened 2019-05-14
feat(openapi-v3): allow controller to reference models via openapispec

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/2629.

Controller methods can now reference model schema directly via OpenAPI spec and without using the `x-ts-type` extension

## Checklist

### DIFF
--- a/packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
+++ b/packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
@@ -3,10 +3,15 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {ParameterObject, SchemaObject} from '@loopback/openapi-v3-types';
-import {model, property} from '@loopback/repository';
+import {
+  OperationObject,
+  ParameterObject,
+  SchemaObject,
+} from '@loopback/openapi-v3-types';
+import {Entity, model, property} from '@loopback/repository';
 import {expect} from '@loopback/testlab';
 import {
+  api,
   ControllerSpec,
   get,
   getControllerSpec,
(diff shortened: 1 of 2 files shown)


packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -203,6 +207,131 @@ describe('controller spec', () => {
...
+        properties: {
+          title: {
+            type: 'string',
+          },
+        },
+      },
+    });
+  });
+
+  it('returns an empty object when it cannot find definition of referenced model', () => {
```
raymondfeng: Should we report it as an error for unresolved `$ref` or defer it to other definitions outside of the controller?

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -203,6 +207,131 @@ describe('controller spec', () => {
...
+        title: 'Todo',
+        properties: {
+          title: {
+            type: 'string',
+          },
+        },
+      },
+    });
+  });
+
```
raymondfeng: Can you add a test case to allow `@api` at controller class level to supply `definitions` so that they can be shared by multiple methods in the controller?

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -235,12 +237,18 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
 ) {
-  debug('  evaluating schema: %j', schema);
-  if (!schema || isReferenceObject(schema)) return;
+  debug('  processing extensions in schema: %j', schema);
+  if (!schema) return;
+
+  assignRelatedSchemas(spec, schema.definitions);
+  delete schema.definitions;
+
+  if (isReferenceObject(schema)) return;
```
b-admike: Is there a particular reason we split this logic from the early return on line 245?

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
jannyHou: was talking with @nabdelgadir , I think the `@api()` decorator takes in a whole OpenAPI spec described by interface [`ControllerSpec`](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/controller-spec.ts#L29), so if a user wants to provide the model schema, the spec should be sth like:
```
{
  components: {
    schemas: {
      Todo: {
        // The Todo schema
      }
    }
  },
  paths: {
    // some path spec refers to the Todo schema provided in the root level property 
    // `components`, instead of providing the definition in itself.
  }
}
```

nabdelgadir: I added a test in [this commit](https://github.com/strongloop/loopback-next/pull/2898/commits/e738d63d1a10a97d33d5d47b9c0fe4288fbd3b91), although I'm not sure it's what you meant. I thought that OpenAPI spec had shared definitions in `#/components/schemas`? Can you please provide an example of what you meant?

packages/openapi-v3/src/controller-spec.ts
nabdelgadir: Say you have:
```ts
schema: {
  $ref: #/definitions/Todo,
  definitions: {/* ... */}
}
```
If you keep it the same as before then `!schema || isReferenceObject(schema)` will be a true statement since [`isReferenceObject(schema)`](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3-types/src/type-guards.ts#L19-L21) is true and it'll keep the schema as is.

But with the modified check, now it goes through:
```ts
assignRelatedSchemas(spec, schema.definitions);
delete schema.definitions;
```
and assigns related schemas and removes `definitions: {/* ... */}` before returning.

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
raymondfeng: @jannyHou Good point. We already have the ability to define `schema` at `@api` level. With that in mind, what's the additional value to allow `definitions`? Is it just for full support of json schema?

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -332,6 +333,66 @@ describe('controller spec', () => {
     expect(globalSchemas).to.be.empty();
   });
 
+  it('allows a class to provide definitions of referenced models through #/components/schemas', () => {
+    @api({
+      paths: {
```
raymondfeng: What I meant is to only have schemas at `@api` level for the controller class, not individual operations.

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
jannyHou: @raymondfeng I think `definitions` is the way to add schema reference in the method level metadata, the reason is explained in Miroslav's comment: https://github.com/strongloop/loopback-next/issues/2629#issuecomment-484533465. 
`@api()` doesn't have that limitation :)

raymondfeng: Schema references can be used at method level regardless where the target schema is provided. With this PR, we allow the target schema to be defined in two ways:

1. At `@api` level - inside `components.schemas.<target>`
2. Use `definitions` with `$ref`:
```ts
schema: {
  $ref: #/definitions/Todo,
  definitions: {/* ... */} // [raymond] I assume the schema definitions are going to be visible to other methods?
}
```

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -203,6 +208,249 @@ describe('controller spec', () => {
...
+      class MyController {
+        @get('/todos', {
+          responses: {
+            '200': {
+              description: 'Array of Category model instances',
+              content: {
+                'application/json': {
+                  schema: {
+                    $ref: '#/definitions/Todo',
+                    definitions: {},
```
raymondfeng: We should remove `definitions` here.

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -203,6 +208,249 @@ describe('controller spec', () => {
...
+      class MyController {
+        @get('/todos', {
+          responses: {
+            '200': {
+              description: 'Array of Category model instances',
+              content: {
+                'application/json': {
+                  schema: {
+                    $ref: '#/components/schemas/Todo',
+                    definitions: {
```
raymondfeng: Can we add a test case to show that `definitions` in `find` method can be referenced by another method, such as `findById`?

nabdelgadir: I'm not sure how to test this, but since it's verified that the definition is now present in `#/components/schemas`, then is it safe to assume that any method that references `#/components/schemas/Todo` will be able to reference the definition? If not, any suggestions on how to modify [this test](https://github.com/strongloop/loopback-next/pull/2898/commits/3eec6a51b19cc8b775425b2206dc20df3a341049#diff-28766b6f539bb8743b1be35f2ef35b68R310) to test this scenario (I'll remove the test, otherwise)?

packages/openapi-v3/src/controller-spec.ts
b-admike: Thanks for clarifying @nabdelgadir :-). Sounds good to me.

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
```diff
@@ -307,6 +307,71 @@ describe('controller spec', () => {
...
+              content: {
+                'application/json': {
+                  schema: {$ref: '#/components/schemas/Todo'},
+                },
+              },
+            },
+          },
+        })
+        async findById(): Promise<Todo> {
+          return new Todo();
```
b-admike: We can either not label this as an `async` function or do `return await Promise.resolve(new Todo());`

raymondfeng: The current implementation is fine. `async` has two effects:

1. Allows `await`
2. Forces the return type to a Promise.

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
raymondfeng: You can have two methods:
1. The 1st one has `definitions`
2. The 2nd one has `$ref` to a schema in `definitions` from the 1st method

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
b-admike: Oh did not know that, thanks for the clarification :-).

packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
nabdelgadir: Since the `definition` will become a part of `#/components/schemas`, they should be visible to other methods.

Issue #2900, opened 2019-05-14
Create a common layer for authentication & authorization component


dhmlau:
## Description / Steps to reproduce / Feature proposal
_Capturing the information from @raymondfeng in today's meeting and in Slack._

We might need a common layer e.g. `@loopback/security` to serve as a base where `@loopback/authentication` and `@loopback/authorization` are two extensions.  The common layer can possibly include:
- SecurityContext
- binding keys for current user/subject

_From the conversation in Slack:_
- Authentication binds the `subject` (who) to request context so that authorization can use it to make a decision against object + action (what)
- The subject contains a set of principals and other information from the access token (such as scopes)
- Today we build `UserProfile`, which is a subset of `subject`
- For example, facebook supports `user token`, `client app token`, `page token` etc
- The principal can a user, an application, or a device
- Maybe we should have a common module like `@loopback/security` to define the common interfaces/types

dhmlau:
Draft PR #2902.
Recalling the conversation with @raymondfeng, he thinks after PR #1205 is done, we can see what makes sense to put in the common layer.

jannyHou:
Closing as PR merged.

PR #2902, opened 2019-05-15
[WIP] feat(security): add @loopback/security for common interfaces/types

raymondfeng:
Implements #2900 

## Checklist


packages/security/src/types.ts
```diff
@@ -0,0 +1,183 @@
...
+   * Property of a protected resource type/instance, such as `email`
+   */
+  resourceProperty?: string;
+  /**
+   * Identity of a protected resource instance, such as `order-0001` or
+   * `customer-101`
+   */
+  resourceId?: string;
+
+  get [securityId]() {
```
jannyHou: Are we following any standard to generate the `[securityId]`?

My understanding is: we collect permission information(action, resourceType, etc...) from request, then generate `[securityId]` as a string, then sends it to the authorization server. And my question is: how do we know that the auth server understands the string?

packages/security/src/types.ts
```diff
@@ -0,0 +1,183 @@
...
+ * request. It's the `who` for security.
+ *
+ * Such operations include:
+ * - authentication (login)
+ * - authorization (access control)
+ * - session access
+ * - logout
+ *
+ */
+export interface Subject {
```
jannyHou: @raymondfeng should we make the `principals` more strict? Like it can only contain one item with `{type: 'USER'}`(same for other types), since theoretically one request is only from one user/app/device

packages/security/src/types.ts
raymondfeng: Please note that `securityId` is a `Symbol` - a unique key:

```ts
export const securityId = Symbol('securityId');
```

packages/security/src/types.ts
raymondfeng: Like an object as follows?
```ts
{
  user?: Principal,
  application?: Principal,
  device?: Principal, 
  [other:string]: Principal,
}
```

packages/security/src/types.ts
jannyHou: @raymondfeng sorry I didn't put it clearly, I mean the value of key `[securityId]`, not the symbol itself. (And I like the idea of having the security id as a symbol)

raymondfeng:
Close in favor of #3551

PR #2907, opened 2019-05-15
feat(cli): add `lb4 interceptor` to generate interceptors and a booter to load interceptors

raymondfeng:
1. Add `lb4 interceptor` command to generate interceptors.
2. Add `interceptor` booter to load interceptors from `src/interceptors`.

## Checklist

### DIFF
--- a/docs/site/Interceptor-generator.md
+++ b/docs/site/Interceptor-generator.md
@@ -1,0 +1,160 @@
+---
+lang: en
+title: 'Interceptor generator'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Interceptor-generator.html
+---
+
+{% include content/generator-create-app.html lang=page.lang %}
+
+### Synopsis
+
+Adds a new [interceptor](Interceptors.md#global-interceptors) class to a
+LoopBack application.
+
+```sh
+lb4 interceptor [--global] [--group <group>] [<name>]
+```
+
+### Arguments and options
+
+`<name>` - Name of the interceptor to create as an argument to the command. If
+provided, the tool will use that as the default when it prompts for the name.
+
+`--global` - Optional flag to indicate a global interceptor (default to `true`).
+Use `--no-global` to set it to `false`.
+
+`--group <group>` - Optional name of the interceptor group to sort the execution
+of global interceptors by group. This option is only supported for global
+interceptors.
+
+### Interactive Prompts
+
+The tool will prompt you for:
+
+- **Name of the interceptor.** _(interceptorName)_ If the name had been supplied
+  from the command line, the prompt is skipped.
+
+- **Is it a global interceptor.** _(isGlobal)_ If the flag had been supplied
+  from the command line, the prompt is skipped.
+
+- **Group of the interceptor.** _(groupName)_ If the group had been supplied
+  from the command line, the prompt is skipped. See
+  [Order of invocation for interceptors](https://loopback.io/doc/en/lb4/Interceptors.html#order-of-invocation-for-interceptors).
+
+### Output
+
+Once all the prompts have been answered, the CLI will do the following:
+
+- Create an interceptor class as follows:
+  `/src/interceptors/${interceptorName}.interceptor.ts`
+- Update `/src/interceptors/index.ts` to export the newly created global
+  interceptor class.
+
+The generated class looks like:
+
+1. A global interceptor
+
+```ts
+import {
+  /* inject, */
+  globalInterceptor,
+  bind,
+  Interceptor,
+  Provider,
+} from '@loopback/context';
+
+/**
+ * This class will be bound to the application as a global `Interceptor` during
+ * `boot`
+ */
+@globalInterceptor('auth', {tags: {name: 'test'}})
+export class TestInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  /**
+   * This method is used by LoopBack context to produce an interceptor function
+   * for the binding.
+   *
+   * @returns An interceptor function
+   */
+  value() {
+    return this.intercept.bind(this);
+  }
+
+  /**
+   * The logic to intercept an invocation
+   * @param invocationCtx - Invocation context
+   * @param next - A function to invoke next interceptor or the target method
+   */
+  async intercept(
+    invocationCtx: InvocationContext,
+    next: () => ValueOrPromise<InvocationResult>,
+  ) {
+    try {
+      // Add pre-invocation logic here
+      const result = await next();
+      // Add post-invocation logic here
+      return result;
+    } catch (err) {
+      // Add error handling logic here
+      throw err;
+    }
+  }
+}
+```
+
+2. A non-global interceptor
+
+```ts
+import {
+  /* inject, */
+  bind,
+  Interceptor,
+  Provider,
+} from '@loopback/context';
+
+/**
+ * This class will be bound to the application as a global `Interceptor` during
+ * `boot`
+ */
+@bind({tags: {namespace: 'interceptors', name: 'test'}})
+export class TestInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  /**
+   * This method is used by LoopBack context to produce an interceptor function
+   * for the binding.
+   *
+   * @returns An interceptor function
+   */
+  value() {
+    return this.intercept.bind(this);
+  }
+
+  /**
+   * The logic to intercept an invocation
+   * @param invocationCtx - Invocation context
+   * @param next - A function to invoke next interceptor or the target method
+   */
+  async intercept(
+    invocationCtx: InvocationContext,
+    next: () => ValueOrPromise<InvocationResult>,
+  ) {
+    try {
+      // Add pre-invocation logic here
+      const result = await next();
+      // Add post-invocation logic here
+      return result;
+    } catch (err) {
+      // Add error handling logic here
+      throw err;
+    }
+  }
+}
+```
--- a/packages/boot/src/__tests__/fixtures/interceptor.artifact.ts
+++ b/packages/boot/src/__tests__/fixtures/interceptor.artifact.ts
@@ -1,0 +1,54 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/boot
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  globalInterceptor,
+  Interceptor,
+  InvocationContext,
+  InvocationResult,
+  Provider,
+  ValueOrPromise,
+} from '@loopback/context';
+
+/**
+ * This class will be bound to the application as a global `Interceptor` during
+ * `boot`
+ */
+@globalInterceptor('auth', {tags: {name: 'myGlobalInterceptor'}})
+export class MyGlobalInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  /**
+   * This method is used by LoopBack context to produce an interceptor function
+   * for the binding.
+   *
+   * @returns An interceptor function
+   */
+  value() {
+    return this.intercept.bind(this);
+  }
+
+  /**
+   * The logic to intercept an invocation
+   * @param invocationCtx - Invocation context
+   * @param next - A function to invoke next interceptor or the target method
+   */
+  async intercept(
+    invocationCtx: InvocationContext,
+    next: () => ValueOrPromise<InvocationResult>,
+  ) {
+    try {
+      // Add pre-invocation logic here
+      const result = await next();
+      // Add post-invocation logic here
+      return result;
+    } catch (err) {
+      // Add error handling logic here
+      throw err;
+    }
+  }
+}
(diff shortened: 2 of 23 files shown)


docs/site/Global-interceptor-generator.md
```diff
@@ -0,0 +1,79 @@
...
+ * `boot`
+ */
+@globalInterceptor('auth')
+export class TestInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  value() {
+    const interceptor: Interceptor = async (invocationCtx, next) => {
```
bajtos: We have discussed this in the past, when reviewing interceptor pull requests. Please don't use anonymous functions and move the interceptor into a named class method `intercept`.

```ts
@globalInterceptor('auth')
export class TestInterceptor implements Provider<Interceptor> {
  /*
  constructor() {}
  */

  value() {
    return this.intercept.bind(this);
  }

  async intercept(invocationCtx, next) {
    // Add pre-invocation logic here
    const result = await next();
    // Add post-invocation logic here
    return result;
  }
}
```

Should we perhaps generalize this pattern into `InterceptorProvider`?

```ts
// in the same package where Interceptor is defined:
export abstract class InterceptorProvider implements Provider<Interceptor> {
  value() {
    return this.intercept.bind(this);
  }

  abstract intercept(
    invocationCtx: InvocationContext,
    next: () => ValueOrPromise<InvocationResult>
  ): ValueOrPromise<InvocationResult>;
}

// in application code
export class TestInterceptor extends InterceptorProvider {
  /*
  constructor() {}
  */

  abstract intercept(
    invocationCtx: InvocationContext,
    next: () => ValueOrPromise<InvocationResult>
  ): ValueOrPromise<InvocationResult> {
      // Add pre-invocation logic here
    const result = await next();
    // Add post-invocation logic here
    return result;
  }
}
```

packages/boot/src/__tests__/fixtures/interceptor.artifact.ts
```diff
@@ -0,0 +1,27 @@
...
+ * `boot`
+ */
+@globalInterceptor('auth')
+export class MyGlobalInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  value() {
+    const interceptor: Interceptor = async (invocationCtx, next) => {
```
bajtos: Ditto, move the anonymous functions into a named class method.

packages/boot/src/__tests__/integration/interceptor.booter.integration.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    await app.boot();
+
+    const bindings = app
+      .findByTag(ContextTags.GLOBAL_INTERCEPTOR)
+      .map(b => ({key: b.key, tags: b.tagNames, scope: b.scope}));
+    expect(bindings).to.containEql(expectedBinding);
+  });
+
+  async function getApp() {
```
bajtos: Please improve the name to communicate that this is not just any application, but it's an application with an interceptor artifact.

packages/boot/src/booters/interceptor.booter.ts
```diff
@@ -0,0 +1,78 @@
...
+   * creating a DataSourceConstructor and binding it to the application class.
+   */
+  async load() {
+    await super.load();
+
+    this.interceptors = this.classes as InterceptorProviderClass[];
+    for (const interceptor of this.interceptors) {
+      debug('Bind global interceptor: %s', interceptor.name);
+      const binding = createBindingFromClass(interceptor, {
+        namespace: GLOBAL_INTERCEPTOR_NAMESPACE,
```
bajtos: I am concerned about limiting the files in `src/interceptors` directory to global interceptors. Where are developers supposed to place local (non-global) interceptors?

Personally, I think I would prefer to keep all my interceptors in `src/interceptors` and define a mechanism or convention for deciding which interceptors should be bound as global interceptors; and which should be only bound in the context for manual consumption, or perhaps not bound at all and consumed via `import`.

Two options come to my mind:
1. Use nested directories. `/interceptors/foo.interceptor.ts` is a regular interceptor that is only bound to context but not invoked globally. `/interceptors/global/foo.interceptor.ts` is a global interceptor.
2. Use `@bind`metadata in individual source files to control whether the interceptor should be registered as global or not.

IIUC, you are already leveraging the second approach via `@globalInterceptor('<%= group %>')`, which looks very reasonably to me 👍 

Now let's make sure we can boot non-global interceptors too.

packages/cli/generators/interceptor/index.js
```diff
@@ -0,0 +1,132 @@
...
+      Object.assign(this.artifactInfo, {group: this.options.group});
+    }
+
+    const prompts = [
+      {
+        type: 'input',
+        name: 'group',
+        // capitalization
+        message: utils.toClassName(this.artifactInfo.type) + ' group:',
+        default: '',
```
bajtos: How can the person running `lb4 interceptor` find out what are the valid group names?

What happens when they enter a group name that's not recognized by the app?

packages/cli/generators/interceptor/templates/interceptor-template.ts.ejs
```diff
@@ -0,0 +1,28 @@
...
+ * `boot`
+ */
+@globalInterceptor('<%= group %>')
+export class <%= className %> implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  value() {
+    const interceptor: Interceptor = async (invocationCtx, next) => {
```
bajtos: Please rework into named class method.

packages/context/src/interceptor.ts
```diff
@@ -197,15 +203,26 @@ export class InvocationContext extends Context {
...
  * by tagging it with `ContextTags.INTERCEPTOR`
- * @param binding - Binding object
+ * @param group - Group for ordering the interceptor
  */
 export function asGlobalInterceptor(group?: string): BindingTemplate {
   return binding => {
-    binding.tag(ContextTags.GLOBAL_INTERCEPTOR);
+    binding
+      .tag(ContextTags.GLOBAL_INTERCEPTOR)
+      .tag({[ContextTags.NAMESPACE]: GLOBAL_INTERCEPTOR_NAMESPACE});
```
bajtos: I am confused. Why do we need both `GLOBAL_INTERCEPTOR` tag and `GLOBAL_INTERCEPTOR_NAMESPACE`? I mean why is it not enough to use the tag only?

Is it a good idea to use the same namespace for all interceptors, regardless of which component/extension is contributing them? Wouldn't it better to use namespaces similarly to how Java treat packages, i.e. interceptors contributed by `@loopback/authorization` package should use `@loopback/authorization` namespace.

Let's discuss.

raymondfeng:
> I consider anonymous functions in Providers as a bad practice, let's use named class methods instead.
> I'd like us to consider non-global interceptors too and find a design that can accommodate both global and non-global interceptors. Ideally, I'd like the CLI prompt to ask the developer whether they want to register the interceptor-being-created as a global one.

Both are addressed.

packages/context/src/interceptor.ts
raymondfeng: Only `GLOBAL_INTERCEPTOR` is required. `GLOBAL_INTERCEPTOR_NAMESPACE` is to make the binding key more readable.

packages/cli/generators/interceptor/index.js
raymondfeng: Based on our design, unknown group is sorted by alphabetical order before the known groups.

packages/boot/src/booters/interceptor.booter.ts
raymondfeng: Added support for non-global interceptors. I had it in mind but was trying to allow global ones 1st.

packages/cli/generators/interceptor/index.js
bajtos: > Based on our design, unknown group is sorted by alphabetical order before the known groups.

That answered my second question 👍 

The first one is still not answered though:
> How can the person running `lb4 interceptor` find out what are the valid group names?

raymondfeng: Since the valid group names are configured by a binding, I'm not sure how we can check. Maybe we can improve the prompt to explain so.

bajtos: > Maybe we can improve the prompt to explain so.

Sounds good.

Ideally, I'd like the prompt to tell me:
1. What are the usual (default) group names
2. Where can I find the actual list of group names as recognized by my application

packages/context/src/interceptor.ts
bajtos: Please add a code comment to capture this information.

Also, what is your opinion about using the same namespace for all interceptors vs. using a namespace based on the component/extension contributing the interceptor?

packages/cli/test/integration/generators/interceptor.integration.js
```diff
@@ -0,0 +1,129 @@
...
+    it('generates a non-global interceptor from CLI', async () => {
+      await testUtils
+        .executeGenerator(generator)
+        .inDir(SANDBOX_PATH, () =>
+          testUtils.givenLBProject(SANDBOX_PATH, {
+            additionalFiles: SANDBOX_FILES,
+          }),
+        )
+        .withArguments('myInterceptor --no-global');
+      verifyGeneratedScript(false);
```
bajtos: Boolean arguments are evil. What does `false` means here? It's hard to tell from the name of the invoked function (`verifyGeneratedScript`). Please convert it to an options object, e.g.

```suggestion
      verifyGeneratedScript({global: false});
```

packages/context/src/interceptor.ts
raymondfeng: Personally, I prefer to use `globalInterceptors.*`.

docs/site/Interceptor-generator.md
```diff
@@ -0,0 +1,120 @@
...
+ * `boot`
+ */
+@globalInterceptor('auth', {tags: {name: 'test'}})
+export class TestInterceptor implements Provider<Interceptor> {
+  /*
+  constructor() {}
+  */
+
+  value() {
+    const interceptor: Interceptor = async (invocationCtx, next) => {
```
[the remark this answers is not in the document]
raymondfeng: Can you remind me why you are against using an arrow function inside `value()`? Is it about the stack trace? In my test, the stack trace shows `interceptor` (name of the const) as the function name, which is good to me.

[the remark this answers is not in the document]
bajtos: Yes, it's mostly about the stack trace. IMO, `TestInterceptor.intercept` is most helpful. 

`interceptor` is not entirely bad, but I see few problems:
- It's not clear which interceptor is involved, one has to look to the filename to find that. (This works only when there is a single interceptor exported by that source file.)
- It's not clear why we are storing the interceptor in a local variable. Personally, I would simplify the code as follows, with unintended consequence on stack traces:

    ```ts
    value() {
      return async (invocationCtx, next) => {
        // ...
      };
    }
    ```

More subjectively, this is also about [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) - the `value` function should be concerned with obtaining the interceptor function only, the actual interception should be implemented elsewhere.

packages/context/src/interceptor.ts
```diff
@@ -200,15 +206,28 @@ export class InvocationContext extends Context {
...
+      // Tagging with `GLOBAL_INTERCEPTOR` is required.
+      .tag(ContextTags.GLOBAL_INTERCEPTOR)
+      // `GLOBAL_INTERCEPTOR_NAMESPACE` is to make the binding key more readable.
+      .tag({[ContextTags.NAMESPACE]: GLOBAL_INTERCEPTOR_NAMESPACE});
     if (group) binding.tag({[ContextTags.GLOBAL_INTERCEPTOR_GROUP]: group});
   };
 }
 
+/**
+ * `@globalInterceptor` decorator to mark a class a global interceptor
```
bajtos: Missing "as"? English is not my mother tongue, I may be wrong.

```suggestion
 * `@globalInterceptor` decorator to mark a class as a global interceptor
```

bajtos:
@raymondfeng I am not sure if this comment has been addressed yet: https://github.com/strongloop/loopback-next/pull/2907#discussion_r290255878

> > Maybe we can improve the prompt to explain so.
>
> Sounds good.
>
> Ideally, I'd like the prompt to tell me:
> 1. What are the usual (default) group names
> 2. Where can I find the actual list of group names as recognized by my application

raymondfeng:
@bajtos Did you notice that we now have the following prompt to address https://github.com/strongloop/loopback-next/pull/2907#discussion_r290255878?
```
? Interceptor name: test
? Is it a global interceptor? Yes
? Global interceptors are sorted by the order of an array of group names bound t
o ContextBindings.GLOBAL_INTERCEPTOR_ORDERED_GROUPS. See https://loopback.io/doc
/en/lb4/Interceptors.html#order-of-invocation-for-interceptors.
Group name for the global interceptor: ('') 
```

docs/site/Interceptor-generator.md
```diff
@@ -0,0 +1,160 @@
...
+lb4 interceptor [--global] [--group <group>] [<name>]
+```
+
+### Arguments and options
+
+`<name>` - Name of the interceptor to create as an argument to the command. If
+provided, the tool will use that as the default when it prompts for the name.
+
+`--global` - Optional flag to indicate a global interceptor (default to `true`).
+Use `--no-global` to set it to `false`.
```
emonddr: > --no-global 

Is this the standard in CLI options..to place  'no-' in front of an option like `global`?

Perhaps `not-global` or `non-global` ?

raymondfeng: It's a built-in convention for yeoman.

PR #2908, opened 2019-05-16
feat(context): generalize invocation handler chain and use it for interceptors

raymondfeng:
Extract the logic for invocation handler chain from interceptor and generalize it as common facility.

We can potentially use the apis to build a sequence of actions using this style.

## Checklist


packages/context/src/__tests__/unit/handler-chain.unit.ts
```diff
@@ -0,0 +1,191 @@
...
+    givenInvocationHandlerChain({handler1: {}, handler2: {returnValue: 'ABC'}});
+    const result = await handlerChain.invokeHandlers();
+    expect(result).to.eql('ABC');
+  });
+
+  it('skips downstream handlers if next is not invoked', async () => {
+    givenInvocationHandlerChain({handler1: {skipNext: true}, handler2: {}});
+    await handlerChain.invokeHandlers();
+    expect(events).to.eql(['before-handler1', 'after-handler1']);
+  });
```
bajtos: I find these tests are very cryptic, it's difficult to understand what is being tested (e.g. what exactly does `skipNext` mean?) and why are we expecting the outcome encoded in assertions/expectations.

Can you find a better way how to structure these tests please? Even at the expense of a small amount of duplication.

packages/context/src/__tests__/unit/handler-chain.unit.ts
```diff
@@ -0,0 +1,191 @@
...
+  function givenInvocationHandlerChain(
+    options: {
+      handler1: HandlerOptions;
+      handler2: HandlerOptions;
+    } = {handler1: {}, handler2: {}},
+  ) {
+    const handler1 = givenHandler('handler1', options.handler1);
+    const handler2 = givenHandler('handler2', options.handler2);
+    ctx.bind('handler2').to(handler2);
+    const handlers = [handler1, 'handler2'];
```
bajtos: IIUC, this is verifying that handlers can be specified both by a value and by binding key.

I find the current code very problematic:
 - The assertion/expectation is implicit, it's not obvious from the test cases.
 - We are verifying this behavior in every test, which is rather redundant.
 - It makes it difficult to see what the tests are about.

packages/context/src/__tests__/unit/handler-chain.unit.ts
```diff
@@ -0,0 +1,191 @@
...
+      'after-handler1',
+    ]);
+    expect(result).to.be.undefined();
+  });
+
+  it('honors return value', async () => {
+    givenInvocationHandlerChain({handler1: {}, handler2: {returnValue: 'ABC'}});
+    const result = await handlerChain.invokeHandlers();
+    expect(result).to.eql('ABC');
+  });
```
bajtos: Here is an example showing how to rewrite these tests for clarity.

```ts
it('honors return value', async () => {
  new InvocationHandlerChain(ctx, [
    async (context: Context, next: Next) => { return await next(); },
    async (context: Context, next: Next) => { await next(); return 'ABC'; },
  ]);

  const result = await handlerChain.invokeHandlers();
  expect(result).to.eql('ABC');
});
```

The first handler can be extracted into a test helper, e.g.

```ts
async function noopHandler(context: Context, next: Next) { 
  return await next(); 
};
```

packages/context/src/handler-chain.ts
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * `next` function
+ */
+export type Next = () => ValueOrPromise<InvocationResult>;
+
+/**
+ * A handler function to be invoked in a chain for the given context
+ */
+export interface Handler<C extends Context = Context> {
```
bajtos: I find the name `Handler` rather ambiguous. Also in my mind, handler is responsible for handling the entire process, not calling `next`. I.e. when I see the term `Handler`, I expect something closer to Express route handlers (functions call `res.send` and don't call `next`). In your proposal, Handler is closer to Express middleware (functions that manipulate context and call `next`).

Can we come up with a better name? Have you considered using the term `Middleware`? It's the term that Koa is using for this kind of functions. See https://www.npmjs.com/package/koa#middleware

packages/context/src/handler-chain.ts
```diff
@@ -0,0 +1,210 @@
...
+  /**
+   * @param context - Context object
+   * @param next - A function to proceed with downstream handlers or the target
+   *
+   * @returns The invocation result as a value or promise
+   */
+  (context: C, next: Next): ValueOrPromise<InvocationResult>;
```
bajtos: I am not a fan of defining interfaces for function types, I find such type more difficult to understand. I think it's even more confusing to novice TypeScript users. 

Another aspect that I don't like is that you have two TSDoc comments now - one for the interface, another for the actual function.

What are you arguments for preferring interfaces?

My assumption is that the following syntax should work equally well (if not better), please correct me if I am wrong and/or missing something.

```ts
/**
 * A handler function to be invoked in a chain for the given context
 * @param context - Context object
 * @param next - A function to proceed with downstream handlers or the target
 *
 * @returns The invocation result as a value or promise
 */
export type Handler<C extends Context = Context> = 
  (context: C, next: Next) => ValueOrPromise<InvocationResult>;
```

packages/context/src/handler-chain.ts
```diff
@@ -0,0 +1,210 @@
...
+      throw new Error('No more handler is in the chain');
+    }
+    return this.handlers[this._index++];
+  }
+}
+
+/**
+ * A chain of handlers to be invoked for the given context
+ */
+export class InvocationHandlerChain<C extends Context = Context> {
```
bajtos: If we use `Middleware` instead of `Handler,` then I think this can be renamed to `MiddlewareChain`? That reads much better to me than `InvocationHandlerChain`.

packages/context/src/handler-chain.ts
raymondfeng: I'm concerned that our users will mistakenly assume it's Express/Koa middleware.

packages/context/src/handler-chain.ts
raymondfeng: I'm not sure if a `type` definition can have `@param` and `@returns`. That's why I use the interface style.

packages/context/src/__tests__/unit/handler-chain.unit.ts
raymondfeng: I have fixed the tests.

packages/context/src/handler-chain.ts
raymondfeng: `Handler` is used in https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern.

packages/context/src/handler-chain.ts
raymondfeng: I switched to `type`.

packages/context/src/handler-chain.ts
bajtos: > I'm concerned that our users will mistakenly assume it's Express/Koa middleware.

Well, the interface does describe Koa middleware too, doesn't it?

Would `GenericMiddleware<T>` alleviate your concern?

packages/context/src/__tests__/unit/handler-chain.unit.ts
```diff
@@ -0,0 +1,168 @@
...
+      return 'ABC';
+    });
+    const result = await handlerChain.invokeHandlers();
+    expect(result).to.eql('ABC');
+  });
+
+  it('skips downstream handlers if next is not invoked', async () => {
+    givenHandlerChain(async (context, next) => {
+      return 'ABC';
+    }, givenNamedHandler('handler1'));
```
bajtos: Since the named handler is second in the chain, let's use `handler2` as the name.

packages/context/src/__tests__/unit/handler-chain.unit.ts
```diff
@@ -0,0 +1,168 @@
...
+    ]);
+  });
+
+  it('allows discovery of handlers in context', async () => {
+    const handler1 = givenNamedHandler('handler1');
+    const handler2 = givenNamedHandler('handler2');
+    ctx
+      .bind('handler2')
+      .to(handler2)
+      .tag('handler');
```
bajtos: Should we have a binding template `asHandler` to set the tag and possibly other binding metadata as needed? Unless the tag `handler` is arbitrary and used by the test only, in which case please make that more obvious by using a different tag name, e.g. `some-tag`.

packages/context/src/handler-chain.ts
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * A handler function to be invoked in a chain for the given context
+ *
+ * @param context - Context object
+ * @param next - A function to proceed with downstream handlers or the target
+ *
+ * @returns The invocation result as a value or promise
+ */
+export type Handler<C extends Context = Context> = (
```
bajtos: > I'm concerned that our users will mistakenly assume it's Express/Koa middleware.
> `Handler` is used in https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern.

Well, the interface does describe Koa middleware too, doesn't it? I feel it's actually a good thing if people assume this type is for Koa-like Middleware, because it will make it easier for them to understand the concept.

Would `GenericMiddleware<T>` alleviate your concern?

packages/context/src/handler-chain.ts
bajtos: If we end up with `Handler` as the type name, would it make sense to shorten this class name to `HandlerChain`? I don't understand the `Invocation` part.

packages/context/src/handler-chain.ts
bajtos: Can we find any other name, one that has no connotations to "route handlers" and "express middleware"?

raymondfeng:
> If we end up with Handler as the type name, would it make sense to shorten this class name to HandlerChain? I don't understand the Invocation part.

Renamed to `HandlerChain`.

bajtos:
@raymondfeng 

I feel we need to find a better name, as discussed in https://github.com/strongloop/loopback-next/pull/2908#discussion_r286974595 and  https://github.com/strongloop/loopback-next/pull/2908#discussion_r287362752

I find the name `Handler` rather ambiguous. Also in my mind, handler is responsible for handling the entire process, not calling `next`. I.e. when I see the term `Handler`, I expect something closer to Express route handlers (functions call `res.send` and don't call `next`). In your proposal, Handler is closer to Express middleware (functions that manipulate context and call `next`).

You are concerned that
> our users will mistakenly assume it's Express/Koa middleware.
> Handler is used in https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern.

To which I responded:

> Well, the interface does describe Koa middleware too, doesn't it? I feel it's actually a good thing if people assume this type is for Koa-like Middleware, because it will make it easier for them to understand the concept.
>
> Would `GenericMiddleware<T>` alleviate your concern?

Can we find any other name, one that has no connotations to "route handlers" and "express middleware"?

Let's bring more people to this discussion

@strongloop/loopback-next @strongloop/loopback-maintainers can you help us to find a name that will make the concept easy to learn & understand and at the same time won't be confused with "route handler" and "koa/express middleware"?

raymondfeng:
In AOP terms, around advice is what we try to achieve here. Maybe we can call it `AroundHandler` - A handler around the next handler or the target invocation.

emonddr:
@bajtos @raymondfeng ,

I am fine with `GenericMiddleware` or something else than `Handler`. Whatever we decide
we should have a very descriptive js docs so explain what it is, what it isn't, and what it accomplishes.

jannyHou:
> @strongloop/loopback-next @strongloop/loopback-maintainers can you help us to find a name that will make the concept easy to learn & understand and at the same time won't be confused with "route handler" and "koa/express middleware"?

After reading https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern I am good with a name including word "handler", `AroundHandler` or `ContextHandler`, etc...as long as we can distinguish it from the rest handler :)

samarpanB:
@bajtos, @raymondfeng,

I might be completely out of context here. But if I understand it correctly, it’s actually a handler function for interceptor. So can’t we name it InterceptActionHandler ?

raymondfeng:
> I might be completely out of context here. But if I understand it correctly, it’s actually a handler function for interceptor. So can’t we name it InterceptActionHandler ? Or simply InterceptAction

@samarpanB Thank you for chiming in. We use `Interceptor` to represent specific handlers that intercepts method invocations. The proposal here is to generalize the idea so that it can be used build other chain of responsibilities, such as actions for a sequence or hooks for outbound operations backed by connectors.

But your proposed names are inspiring though. Maybe something like `ActionHandler` or `Action`. See https://github.com/strongloop/loopback-next/pull/2927. I use `RestAction` in that PR.

samarpanB:
> > I might be completely out of context here. But if I understand it correctly, it’s actually a handler function for interceptor. So can’t we name it InterceptActionHandler ? Or simply InterceptAction
> 
> @samarpanB Thank you for chiming in. We use `Interceptor` to represent specific handlers that intercept method invocations. The proposal here is to generalize the idea so that it can be used build other chain of responsibilities, such as actions for a sequence or hooks for outbound operations backed by connectors.
> 
> But your proposed names are inspiring though. Maybe something like `ActionHandler` or `Action`. See #2927. I use `RestAction` in that PR.

I think `Action` is good.

raymondfeng:
So far a few options are proposed:

- Handler
- AroundHandler

- Middleware
- GenericMiddleware

- Action

bajtos:
I like find the name `Action` inspiring too. However, we must be careful to avoid confusion with Sequence Actions, see https://loopback.io/doc/en/lb4/Sequence.html#actions

It makes me wonder if we can find more areas to draw inspiration from. For example, workflows use concepts like "activity" and "operation". When discussing processes, people talk about "events" and "tasks". In recipes, we have "steps". Objects like request/response can be manipulated.

Since we are generalizing interceptors, maybe we can use the term `GenericInterceptor`?

```ts
export interface Interceptor extends GenericInterceptor<InvocationContext> {}
```

I wish we used `InvocationInterceptor` as the type name, then we could use `Interceptor` for the generic variant.

samarpanB:
> I like find the name `Action` inspiring too. However, we must be careful to avoid confusion with Sequence Actions, see https://loopback.io/doc/en/lb4/Sequence.html#actions
> 
I think if we are thinking of this as a generalizing concept which can be used for Sequence Actions as well in future, then, it's fine keep it as Action.

Consider `SequenceAction implements Action` . Just a thought.

samarpanB:
> Since we are generalizing interceptors, maybe we can use the term `GenericInterceptor`?
> 
> ```ts
> export interface Interceptor extends GenericInterceptor<InvocationContext> {}
> ```

I like the GenericInterceptor name. Conforms to the purpose. How about GenericAction ?

raymondfeng:
Naming is the most difficult thing in programming :-).

Since TypeScript makes it really easy to create a type alias, I propose that we should just pick a name and move forward. It's always possible to rename it and keep the old one as an alias for backward compatibility.

I vote for two options:
1. Action and ActionChain
2. GenericInterceptor and InterceptorChain

bajtos:
I find the name `Action` problematic. The interface we are creating here is a function that always accepts `context` & `next` and returns the result of the intercepted/wrapped operation. A sequence action is a free-form function. For example, `findRoute` accepts the incoming request and returns the resolved route, but `send` accepts the incoming response & the operation result, and returns `void`.

Let's go with `GenericInterceptor`. I slightly prefer `GenericInterceptorChain` over `InterceptorChain`, I find it more consistent.

samarpanB:
+1 for GenericInterceptor

raymondfeng:
I picked `GenericInterceptor` and `GenericInterceptorChain`. PTAL.

packages/context/src/interceptor-chain.ts
```diff
@@ -0,0 +1,222 @@
...
+ * Return value for a method invocation
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type InvocationResult = any;
+
+/**
+ * Array of arguments for a method invocation
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type InvocationArgs = any[];
```
bajtos: AFAICT, `InvocationArgs` are not used by the generic chain, they are specific to invocation-interceptors. Please move them back to `packages/context/src/interceptor.ts `.

packages/context/src/interceptor-chain.ts
```diff
@@ -0,0 +1,222 @@
...
+import {Context} from './context';
+import {transformValueOrPromise, ValueOrPromise} from './value-promise';
+
+const debug = debugFactory('loopback:context:interceptor-chain');
+
+/**
+ * Return value for a method invocation
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type InvocationResult = any;
```
bajtos: I find it weird to see type name `InvocationResult` alongside `GenericInterceptorChain`, "invocation" seems like a specialization of "interceptor"  to me.

Lets' find a better name please, e.g. `InterceptorResult` or perhaps `NextResult`/`NextReturnValue`.

When I think about this more, maybe `GenericInterceptor` should accept the "result" type as a generic argument too, allowing specializations to provide better type information where possible? (Of course, invocation interceptor will use `any`.)

packages/context/src/interceptor.ts
```diff
@@ -411,66 +392,10 @@ export function invokeMethodWithInterceptors(
   return tryWithFinally(
     () => {
       const interceptors = invocationCtx.loadInterceptors();
+      const targetMethodInvoker = () => invocationCtx.invokeTargetMethod();
+      interceptors.push(targetMethodInvoker);
```
bajtos: Have you considered overwriting GenericInterceptor's `invokeTarget` method instead of pushing a new interceptor to the list?

If we are not overwriting `invokeTarget`, then what is the point of having such extension point in the `GenericInterceptor` class?

packages/context/src/interceptor-chain.ts
raymondfeng: Added a new type param `T`.

packages/context/src/interceptor.ts
raymondfeng: I removed invokeTarget.

packages/context/src/interceptor-chain.ts
raymondfeng: It's now in `invocation.ts`.

packages/context/src/interceptor-chain.ts
```diff
@@ -8,46 +8,57 @@ import {BindingFilter} from './binding-filter';
...
  * @returns The invocation result as a value (sync) or promise (async)
  */
-export type GenericInterceptor<C extends Context = Context> = (
-  context: C,
-  next: Next,
-) => ValueOrPromise<InvocationResult>;
+export type GenericInterceptor<
+  C extends Context = Context,
+  T = InvocationResult
+> = (context: C, next: Next<T | undefined>) => ValueOrPromise<T | undefined>;
```
bajtos: Why do we need `| undefined` part? IMO, it should be up to the user of `GenericInterceptor` to decide if interceptors are allowed to return `undefined`, and they should have the option to force all interceptors to return a defined value.

packages/context/src/interceptor-chain.ts
```diff
@@ -8,46 +8,57 @@ import {BindingFilter} from './binding-filter';
...
-
 const debug = debugFactory('loopback:context:interceptor-chain');
 
+/**
+ * The `next` function that can be used to  invoke next generic interceptor in
+ * the chain
+ *
+ * @typeParam T - Return type of `next()`
+ */
+export type Next<T = InvocationResult> = () => ValueOrPromise<T>;
```
bajtos: Can you leave out the default value for the generic argument `T`? I am concerned that with the default argument present, it's easy to accidentally use `Next` in places where we should use `Next<SomeOtherT>` instead.

packages/context/src/interceptor-chain.ts
```diff
@@ -8,46 +8,57 @@ import {BindingFilter} from './binding-filter';
...
  *
  * @returns The invocation result as a value (sync) or promise (async)
  */
-export type GenericInterceptor<C extends Context = Context> = (
-  context: C,
-  next: Next,
-) => ValueOrPromise<InvocationResult>;
+export type GenericInterceptor<
+  C extends Context = Context,
+  T = InvocationResult
```
bajtos: Similarly here, I prefer `GenericInterceptor` to explicitly require from users to define `C` and `T` parameters. Since `T` is for `InvocationResult`, I think it makes more sense to call it `R` instead.

packages/context/src/interceptor-chain.ts
```diff
@@ -8,46 +8,57 @@ import {BindingFilter} from './binding-filter';
...
  * function
+ * @typeParam C - `Context` class or a subclass of `Context`
+ * @typeParam T - Return type of `next()`
  */
-export type GenericInterceptorOrKey<C extends Context = Context> =
-  | BindingAddress<GenericInterceptor<C>>
-  | GenericInterceptor<C>;
+export type GenericInterceptorOrKey<
+  C extends Context = Context,
+  T = InvocationResult
```
bajtos: Similarly here, let's remove default values for `C` and `T`, and rename `T` to `R`.

packages/context/src/interceptor-chain.ts
```diff
@@ -79,19 +90,22 @@ class InterceptorChainState<C extends Context = Context> {
  *
  * @typeParam C - `Context` class or a subclass of `Context`
  */
-export class GenericInterceptorChain<C extends Context = Context> {
+export class GenericInterceptorChain<
+  C extends Context = Context,
+  T = InvocationResult
```
bajtos: Ditto, what are the arguments for providing default values for `C` and `T`?

Let's rename `T` to `R`.

packages/context/src/interceptor-chain.ts
```diff
@@ -131,40 +145,32 @@ export class GenericInterceptorChain<C extends Context = Context> {
   /**
    * Invoke the interceptor chain
    */
-  invokeInterceptors(): ValueOrPromise<InvocationResult> {
+  invokeInterceptors(): ValueOrPromise<T | undefined> {
```
bajtos: As commented earlier, I think we should not be adding `undefined` to the retval-type provided by the user and use `ValueOrPromise<T>`.

packages/context/src/interceptor-chain.ts
```diff
@@ -0,0 +1,215 @@
...
+import {BindingFilter} from './binding-filter';
+import {BindingAddress} from './binding-key';
+import {BindingComparator} from './binding-sorter';
+import {Context} from './context';
+import {InvocationResult} from './invocation';
+import {transformValueOrPromise, ValueOrPromise} from './value-promise';
+const debug = debugFactory('loopback:context:interceptor-chain');
+
+/**
+ * The `next` function that can be used to  invoke next generic interceptor in
```
bajtos:
```suggestion
 * The `next` function that can be used to invoke next generic interceptor in
```

packages/context/src/interceptor-chain.ts
raymondfeng: `undefined` can happen when there are no interceptors in the chain.

raymondfeng:
The generic typing of `Next` seems to add quite a bit complexity, such as:

1. If the interceptor chain is empty, should we return `undefined` or throw an error?
2. When it reaches the end of a chain, we need to handle `undefined`
3. It's completely valid that we can have a few transformation interceptors that convert return value from one type to another one. It's not always true that we can enforce the typing for next().

I suggest that we remove the generic typing for now and reevaluate it  with more usages happen.

bajtos:
@raymondfeng great discussion!

> - If the interceptor chain is empty, should we return undefined or throw an error?
> - When it reaches the end of a chain, we need to handle `undefined`

I think these issues were introduced by removing `invokeTarget` and assuming that target invocation will be always implemented via an interceptor.

> - It's completely valid that we can have a few transformation interceptors that convert return value from one type to another one. It's not always true that we can enforce the typing for next().

This is a very good point! Based on this observation, I agree with your proposal to remove the generic type :+1:

> I suggest that we remove the generic typing for now and reevaluate it with more usages happen.

PR #2919, opened 2019-05-17
fix(repository): always copy property definition during juggler model build (#2912)

### LINKED ISSUE #2912: Embedded models broken in repository 1.5.5

After upgrading my project's LB4 dependencies, I'm finding that models that have a property that is another model type are breaking.  I've traced this down to a PR that, ironically, was meant to make this work where it wasn't for others before: #2505 fixing #2130.

I think things were working for me before because I was using the `@property({type: AnotherModelType})` as a workaround for not having proper `embedsOne` support (I'm aware of #1450).

The issue I'm finding is that the _first_ attachment of a model to a DataSource works, but this _modifies the model class_ (specifically rebuilds entries in its `definition`) in such a way that a _second_ attachment of that model to a DataSource fails.

This multiple-attachment is not something that happens with my project in production, but it does happen very much when running unit tests.

## Current Behavior

* Attaching the model to a datasource modifies the `.definition`:
  * The `properties.foo.type` is changed from the `class AnotherModel` to `function AnotherModel(data, options)` resulting from the juggler conversion
* Attaching the model to a _second_ datasource results in a bad model:
  * Deep in the model setup, `isTypeResolver` is called on the modified `properties.foo.type` object, and it returns `true`
  * This results in the property `type` being replaced again, this type with a call to that "resolve" ... but it's not a resolver, it's a constructor function
  * Now the property `type` is actually an _instance_ of the model class
  * Usage of the containing model now fails in places where it tries to do `instanceOf` using the _instance_ on the right hand side, which fails.  And if that didn't fail, it would have tried to "call" the instance as a constructor function, which would also have failed.

## Expected Behavior

* Either `@property({type: ModelType})` or `@embedsOne` should "work"
* Attaching models to datasources should not modify the model definition
  * Or if it does, it should do so in a way such that the model can still be attached to other datasources

## Attempted Workarounds

I gave `embedsOne` whirl even though it's not listed as supported and the most obvious issue I hit at first is that it doesn't result in a `property` being declared on the model, which means that the embedded object is excluded from `.toJSON()`.  That breakage was sufficiently fundamental to my use case I didn't get any farther in testing things.

I experimented with modifying the `ModelClass.definition.properties.foo` object to make the `type` property un-settable, and this "seems" to work around my issue, but it's ... a strange and worrying thing to do.

## Possible Fix

In `legacy-juggler-bridge.ts` ~ line 142, it looks like making it always clone the original property definition would be viable:
```typescript
    Object.entries(definition.properties).forEach(([key, value]) => {
      if (value.type === 'array' || value.type === Array) {
        value = Object.assign({}, value, {
          type: [value.itemType && this.resolvePropertyType(value.itemType)],
        });
        delete value.itemType;
// BEGIN CHANGE
      } else {
        value = Object.assign({}, value, {
          type: this.resolvePropertyType(value.type),
        });
      }
      properties[key] = value;
// END CHANGE
    });
```

Applying this change locally gets my unit tests to pass, and causes my "protect the `type` property" workaround code from firing.

mgabeler-lee-6rs:
Avoids an issue where it modifies the property definition in a way such that
it can't build a second juggler model from the same definition, e.g.  from
attaching the model to a second datasource.

Fixes #2912

## Checklist

:heavy_check_mark: 👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

### DIFF
--- a/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -190,6 +190,10 @@ describe('DefaultCrudRepository', () => {
         name: 'Role',
         properties: {name: {type: String}},
       });
+
+      // issue 2912: make sure the juggler leaves the original model definition alone
+      expect(User.definition.properties.roles.itemType).to.equal(Role);
+      expect(User.definition.properties.address.type).to.equal(Address);
     });
   });
 
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -140,13 +140,18 @@ export class DefaultCrudRepository<T extends Entity, ID>
     // We need to convert PropertyDefinition into the definition that
     // the juggler understands
     Object.entries(definition.properties).forEach(([key, value]) => {
+      // always clone value so that we do not modify the original model definition
+      // ensures that model definitions can be reused with multiple datasources
       if (value.type === 'array' || value.type === Array) {
         value = Object.assign({}, value, {
           type: [value.itemType && this.resolvePropertyType(value.itemType)],
         });
         delete value.itemType;
+      } else {
+        value = Object.assign({}, value, {
+          type: this.resolvePropertyType(value.type),
+        });
       }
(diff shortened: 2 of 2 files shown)


raymondfeng:
@mgabeler-lee-6rs Thank you for the patch!

PR #2928, opened 2019-05-21
feat(build): enable incremental compilation

bajtos:
Enable incremental compilation to speed up our builds. Note: this improvement will apply to projects consuming LoopBack too, once they upgrade to the latest version of `@loopback/build`.

**BEFORE**

```
$ time npm run build
  real  0m35.229s
  user  3m51.223s
  sys   0m14.384s
```

**AFTER**

The first build to populate tsbuildinfo file:

```
$ time npm run build
  real  0m35.692s
  user  3m55.728s
  sys   0m13.828s
```

Subsequent builds (in my case, there were no code changes from the last build):

```
$ time npm run build
  real  0m19.510s
  user  2m3.894s
  sys   0m9.581s
```

I am expecting the build times to improve even more when TypeScript 3.5 is released, see https://devblogs.microsoft.com/typescript/announcing-typescript-3-5-rc/

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ packages/tsdocs/fixtures/monorepo/docs
 /docs/site/readmes
 /docs/apidocs/reports-temp
 /docs/apidocs/models
+*.tsbuildinfo
 
 # TBD: Exclude api reports from git for now
 /docs/apidocs/reports
--- a/packages/build/config/tsconfig.common.json
+++ b/packages/build/config/tsconfig.common.json
@@ -10,6 +10,8 @@
     // FIXME(bajtos) LB4 is not compatible with this setting yet
     "strictPropertyInitialization": false,
 
+    "incremental": true,
+
     "lib": ["es2018", "esnext.asynciterable"],
     "module": "commonjs",
     "moduleResolution": "node",
--- a/packages/cli/generators/project/templates/_.gitignore
+++ b/packages/cli/generators/project/templates/_.gitignore
@@ -62,3 +62,6 @@ api-docs/
 
 # Transpiled JavaScript files from Typescript
 /dist
+
+# Cache used by TypeScript's incremental build
+*.tsbuildinfo
--- a/packages/cli/generators/project/templates/tsconfig.json.ejs
+++ b/packages/cli/generators/project/templates/tsconfig.json.ejs
@@ -17,6 +17,8 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
 
+    "incremental": true,
+
     "lib": ["es2018", "esnext.asynciterable"],
     "module": "commonjs",
     "moduleResolution": "node",
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -116,6 +116,10 @@ describe('app-generator specific files', () => {
     const pkg = JSON.parse(await readFile('package.json'));
     expect(pkg.scripts).to.have.property('migrate', 'node ./dist/migrate');
   });
+
+  it('creates .gitignore', () => {
+    assert.fileContent('.gitignore', /^\*\.tsbuildinfo$/m);
+  });
 });
 
 describe('app-generator with docker disabled', () => {


raymondfeng:
Is it extracted from https://github.com/strongloop/loopback-next/pull/1636 (I enabled incremental build there too)?

.gitignore
```diff
@@ -13,6 +13,7 @@ benchmark/dist
 .sandbox
 packages/cli/generators/datasource/connectors.json
 docs/site/readmes
+**/tsconfig.build.tsbuildinfo
```
raymondfeng: Let's use `dist/.tsbuildinfo` by configuring tsconfig compilerOptions with:
```json
{
  "tsBuildInfoFile": ".tsbuildinfo"
}
```

This way, no changes is needed in `clean` script.

packages/build/config/tsconfig.common.json
```diff
@@ -10,6 +10,8 @@
     // FIXME(bajtos) LB4 is not compatible with this setting yet
     "strictPropertyInitialization": false,
 
+    "incremental": true,
+
```
raymondfeng: Please add `"tsBuildInfoFile": ".tsbuildinfo",`.

bajtos: AFAIK, "tsBuildInfoFile": ".tsbuildinfo"` setting in `packages/build/config/tsconfig.common.json` will be resolved as `packages/build/config/.tsbuildinfo`, which isn't what we want.

I think we will have to add `tsBuildInfoFile` setting to every per-package `package.json` file.

I'll take a look.

raymondfeng: IIRC, it's resolved against the `outDir`.

bajtos: Quoting from https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends:

> All relative paths found in the configuration file will be resolved relative to the configuration file they originated in.

See also https://github.com/Microsoft/TypeScript/issues/29172

Quoting also from https://github.com/Microsoft/TypeScript/pull/29813

> * When tsconfig specifies compiler options `incremental` file with `.tsbuildinfo` extension is emitted.
> * Composite projects indirectly specify `incremental` as `true` and its error to specify `incremental` as `false` in the composite project
> * If project specifies `tsBuildInfoFile` options, it uses the path specified by that to write the build information.
> * If project specifies `out` or `outFile` option, the `outFileWithoutExtension.tsbuildinfo` file is written next to output js file.
> * If project specified `outDir`, config files base file name with extension as `.tsbuildinfo` is written in `outDir`
> * Otherswise config files base file name with extension as `.tsbuildinfo` is written next to the config file

AFAICT, we are not setting `outDir` in our `tsconfig` files and configure it via a CLI option instead. This is a legacy from the days where we used to have multiple dist directories (`dist8` for Node.js 8, `dist10` for Node.js 10, etc.)

bajtos:
> Is it extracted from #1636 (I enabled incremental build there too)?

Not at all, I was not aware of that effort.

In this pull request, I am unconditionally enabling incremental build for all mono-repo packages and all newly created projects. Existing projects created before my changes are not affected.

packages/build/config/tsconfig.common.json
bajtos: > If project specified `outDir`, config files base file name with extension as `.tsbuildinfo` is written in `outDir`

This rule does not seem to be triggered in our current build setup, even when I change all `tsconfig.build.json` files to include `"outDir": "dist"` :(

bajtos:
> In this pull request, I am unconditionally enabling incremental build for all mono-repo packages and all newly created projects. Existing projects created before my changes are not affected.

Actually, that's not correct. Because the changes are made in shared `tsconfig.common.json`, they will apply to existing projects too.

bajtos:
@raymondfeng I have updated the pull request, PTAL again.
- added `outDir` and `tsBuildInfoFile` to per-package `tsconfig.build.json` files, including CLI templates
- enabled `incremental` flag in the shared config

The tests are failing because of a bug in `lerna bootstrap`, the problem will be fixed by https://github.com/strongloop/loopback-next/pull/2944 and lerna/lerna#2104.

bajtos:
> added `outDir` and `tsBuildInfoFile` to per-package `tsconfig.build.json` files, including CLI templates

I think we should also modify list of files used by `npm pack`, so that tsbuildinfo is NOT published to npm.

raymondfeng:
Maybe we should keep it simple - leave the defaults as is - `.tsbuildinfo` in the package root directory. We can add `.tsbuildinfo` to `.gitignore`.

bajtos:
> Maybe we should keep it simple - leave the defaults as is - `.tsbuildinfo` in the package root directory. We can add `.tsbuildinfo` to `.gitignore`.

Done.

@raymondfeng LGTY now?

.gitignore
```diff
@@ -16,6 +16,7 @@ packages/tsdocs/fixtures/monorepo/docs
 /docs/site/readmes
 /docs/apidocs/reports-temp
 /docs/apidocs/models
+*.tsbuildinfo
```
raymondfeng: Let's add `*.tsbuildinfo` to our cli templates for `_.gitigore` too.

bajtos: Added `_.gitignore` entry and a test to verify.

bajtos:
@raymondfeng should we modify `npm run clean` to remove `tsconfig.build.tsbuildinfo` files? Should I modify `clean` script in every per-package `package.json` file, or should be perhaps improve `lb-clean` to automatically delete that file?

Issue #2632, opened 2019-03-22
Modify Repository `find*` methods to include navigational properties


bajtos:
Modify Repository `find*` method signatures to include navigational properties in the description of the return type

- Add a new generic parameter `Relations` to CRUD-related Repository interfaces and implementations.
- Modify the signature `find` and `findById` to return `T & Relations` instead of `T`. If this requires too many explicit casts, then consider using `T & Partial<Relations>` instead, assuming it improves the situation.

See the spike #2592 for more details, especially the changes made in [packages/repository/src/repositories/repository.ts](https://github.com/strongloop/loopback-next/pull/2592/files#diff-785622a1e0ea38ec20dc8e91adbf6534) and [packages/repository/src/repositories/legacy-juggler-bridge.ts](https://github.com/strongloop/loopback-next/pull/2592/files#diff-177253a1afbdd8eb83b99efef86dd415).

## Acceptance criteria

- [ ] Modify Repository interfaces and classes as described above
- [ ] Update API documentation
- [ ] Update `lb4 model` to start emitting `{ModelName}Relations` interfaces
- [ ] Update `lb4 repository` to include `{ModelName}Relations` in the generic parameters of `DefaultCrudRepository` we are inheriting from

Issue #2633, opened 2019-03-22
Add navigational properties to examples/todo-list


bajtos:
Update `examples/todo-list` to leverage recently added features moving us closer to full inclusion of related models.

See the spike #2592 for more details.

## Acceptance criteria

- [x] Define `{Model}Relations` interfaces and `{Model}WithRelations` types

- [x] Update `{Model}Repository` implementations to use these new interfaces

- [x] Update repositories to include related models: overwrite `find` and `findById` methods, add a hard-coded retrieval of related models. This part depends on https://github.com/strongloop/loopback-next/issues/2632 

  Include a comment explaining that the example implementation is temporary and add a link to the GitHub issue tracking the work needed to implement production-grade inclusion resolver. 

- [x] Add test coverage at repository level (integration tests)

- [x] Update response schemas for controller methods `find` and `findById` to leverage `getModelSchemaRef` and `includeRelations`. This part depends on https://github.com/strongloop/loopback-next/issues/2631 and https://github.com/strongloop/loopback-next/issues/2630

- [x] Add acceptance-level tests to verify inclusion of related models via REST API

- [x] Update [Todo List Tutorial](https://loopback.io/doc/en/lb4/todo-list-tutorial.html)

PR #2960, opened 2019-05-24
`find*` methods include navigational properties

nabdelgadir:
Resolves https://github.com/strongloop/loopback-next/issues/2632.

- `find` and `findById` now use `T & Relations` instead of `T`
- New prompt for `lb4 model` that asks for relations in order to create the relation interface (e.g. `export interface TodoRelations {}`). It leaves it empty if there are no current relations.
- Newly scaffolded models from `lb4 model` include `{modelName}WithRelations` type (e.g. `export type TodoWithRelations = Todo & TodoRelations;`)
- Generic parameter `Relations` added to `DefaultCrudRepository`
- When scaffolding a repository with `lb4 repository`, it now includes `{modelName}Relations` (e.g. 
```ts
export class TodoListRepository extends DefaultCrudRepository<
  TodoList,
  typeof TodoList.prototype.id
  TodoListRelations
> 
```
)

**Question**: should all models include a relations interface? If so, I'll update the examples/docs.

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -67,6 +67,12 @@ export class Order extends Entity {
     super(data);
   }
 }
+
+export interface OrderRelations {
+  // describe navigational properties here
+}
+
+export type OrderWithRelations = Order & OrderRelations;
 ```
 
 The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
@@ -82,6 +88,10 @@ class Order extends Entity {
   // constructor, properties, etc.
   @belongsTo(() => Customer, {keyTo: 'pk'})
   customerId: number;
+}
+
+export interface OrderRelations {
+  customer?: CustomerWithRelations;
 }
 ```
 
@@ -120,12 +130,13 @@ import {
   juggler,
   repository,
 } from '@loopback/repository';
-import {Customer, Order} from '../models';
+import {Customer, Order, OrderRelations} from '../models';
 import {CustomerRepository} from '../repositories';
 
 export class OrderRepository extends DefaultCrudRepository<
   Order,
-  typeof Order.prototype.id
+  typeof Order.prototype.id,
+  OrderRelations
 > {
   public readonly customer: BelongsToAccessor<
     Customer,
@@ -212,6 +223,13 @@ export class Category extends Entity {
     super(data);
   }
 }
+
+export interface CategoryRelations {
+  categories?: CategoryWithRelations[];
+  parent?: CategoryWithRelations;
+}
+
+export type CategoryWithRelations = Category & CategoryRelations;
 ```
 
 The `CategoryRepository` must be declared like below
@@ -219,7 +237,8 @@ The `CategoryRepository` must be declared like below
 ```ts
 export class CategoryRepository extends DefaultCrudRepository<
   Category,
-  typeof Category.prototype.id
+  typeof Category.prototype.id,
+  CategoryRelations
 > {
   public readonly parent: BelongsToAccessor<
     Category,
--- a/examples/todo-list/src/models/todo-list-image.model.ts
+++ b/examples/todo-list/src/models/todo-list-image.model.ts
@@ -3,8 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Entity, model, property, belongsTo} from '@loopback/repository';
-import {TodoList} from './todo-list.model';
+import {belongsTo, Entity, model, property} from '@loopback/repository';
+import {TodoList, TodoListWithRelations} from './todo-list.model';
 
 @model()
 export class TodoListImage extends Entity {
@@ -29,3 +29,9 @@ export class TodoListImage extends Entity {
     super(data);
   }
 }
+
+export interface TodoListImageRelations {
+  todoList?: TodoListWithRelations;
+}
+
+export type TodoListImageWithRelations = TodoListImage & TodoListImageRelations;
--- a/examples/todo-list/src/models/todo-list.model.ts
+++ b/examples/todo-list/src/models/todo-list.model.ts
@@ -3,9 +3,12 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Entity, model, property, hasMany, hasOne} from '@loopback/repository';
-import {Todo} from './todo.model';
-import {TodoListImage} from './todo-list-image.model';
+import {Entity, hasMany, hasOne, model, property} from '@loopback/repository';
+import {
+  TodoListImage,
+  TodoListImageWithRelations,
+} from './todo-list-image.model';
+import {Todo, TodoWithRelations} from './todo.model';
 
 @model()
 export class TodoList extends Entity {
@@ -36,3 +39,10 @@ export class TodoList extends Entity {
     super(data);
   }
 }
+
+export interface TodoListRelations {
+  todos?: TodoWithRelations[];
+  image?: TodoListImageWithRelations;
+}
+
+export type TodoListWithRelations = TodoList & TodoListRelations;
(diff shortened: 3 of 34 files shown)


packages/cli/generators/model/index.js
```diff
@@ -297,6 +298,52 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+      return this.exit(err);
+    }
+
+    // skip if there are no other models
+    if (!models.length) return;
+
+    return this.prompt([
+      {
+        name: 'relations',
+        message: 'Is this model part of any relations?',
```
bajtos: Very recently, we landed `lb4 relation` command - see https://github.com/strongloop/loopback-next/commit/75939a4a144b3068ff2890394c178faad58d7458.

I find it weird to have `lb4 model` prompting for model relations, when there is a dedicated command that does better job at scaffolding relation-based code.

Is this part addressing the following acceptance criteria?

> Update lb4 model to start emitting `{ModelName}Relations` interfaces

As I was envisioning the solution, an empty interface should be emitted for all newly-created models, so that we can _Update `lb4 repository` to include `{ModelName}Relations` in the generic parameters of `DefaultCrudRepository` we are inheriting from_

Entries for individual relations would be added to `{ModelName}Relations` either manually by the application developer or by `lb4 repository` command.

My conclusions:
- The answer to your _Question: should all models include a relations interface?_ is YES.
- These new `lb4 model` prompts should be removed.
- Ideally, `lb4 repository` should be improved to add entries to relations interfaces. This was not part of the original acceptance criteria and I don't know how much effort would be required, it may be best to leave this part out of scope of this pull request and possibly out of scope of #2960 too.

What do you think?

nabdelgadir: @bajtos thanks for the comment!

> Is this part addressing the following acceptance criteria?
> 
> > Update `lb4 model` to start emitting `{ModelName}Relations` interfaces

Yes, initially I thought calling `lb4 model` should also populate the interface, but I agree that it should be removed and left up to another CLI command.

> The answer to your Question: should all models include a relations interface? is YES.
> These new `lb4 model` prompts should be removed.

👍 

> Ideally, `lb4 repository` should be improved to add entries to relations interfaces. This was not part of the original acceptance criteria and I don't know how much effort would be required, it may be best to leave this part out of scope of this pull request and possibly out of scope of #2960 too.

I originally thought it was, but I misunderstood the acceptance criteria, but I'm good with opening a new issue for this as it's out of scope of this PR.

nabdelgadir: Issue created for adding entries to the interfaces: https://github.com/strongloop/loopback-next/issues/2988

In the new commits, I added empty interfaces because I think they should be populated only after https://github.com/strongloop/loopback-next/issues/2633 is landed. What do you think?

examples/todo-list/src/models/todo-list-image.model.ts
```diff
@@ -29,3 +29,7 @@ export class TodoListImage extends Entity {
     super(data);
   }
 }
+
+export interface TodoListImageRelations {}
```
bajtos: `TodoListImage` already have some relations defined (I think it belongs to TodoList?), can you add navigational properties for these relations to this interface please?

examples/todo-list/src/models/todo-list.model.ts
```diff
@@ -36,3 +36,7 @@ export class TodoList extends Entity {
     super(data);
   }
 }
+
+export interface TodoListRelations {}
```
bajtos: `TodoList` already have some relations defined (I think it has many `TodoListImage` and `Todo` instances), can you add navigational properties for these relations to this interface please?

examples/todo-list/src/models/todo.model.ts
```diff
@@ -41,3 +41,7 @@ export class Todo extends Entity {
     super(data);
   }
 }
+
+export interface TodoRelations {}
```
bajtos: `Todo` already have some relations defined (I think it belongs to TodoList?), can you add navigational properties for these relations to this interface please?

examples/todo/src/models/todo.model.ts
```diff
@@ -44,3 +44,7 @@ export class Todo extends Entity {
     super(data);
   }
 }
+
+export interface TodoRelations {}
```
bajtos: In this example, `Todo` does not have any relations defined, thus `TodoRelations` should be an empty interface :+1:

To reduce noise in `git diff` output when adding new relations in the future, I am proposing to add a line comment into the empty object block as follows.

```ts
export interface TodoRelations {
}
```

This should be applied to all places where we have an empty Relations interface and to the CLI template too.

Thoughts?

packages/cli/generators/model/index.js
```diff
@@ -297,12 +296,6 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
         if (this.artifactInfo.allowAdditionalProperties) {
           Object.assign(this.artifactInfo.modelSettings, {strict: false});
         }
-
-        this.log(
-          `Let's add a property to ${chalk.yellow(
-            this.artifactInfo.className,
```
bajtos: This change seems unrelated to the rest of this pull request, what is the reasoning for removing this message?

packages/cli/generators/model/templates/model.ts.ejs
```diff
@@ -33,3 +33,9 @@ export class <%= className %> extends <%= modelBaseClass %> {
     super(data);
   }
 }
+
+export interface <%= className %>Relations {
+
```
bajtos: As I commented above, let's replace this empty line (that will be removed by prettier) with a line comment.

packages/repository/src/__tests__/fixtures/models/address.model.ts
```diff
@@ -29,3 +29,7 @@ export class Address extends Entity {
   @belongsTo(() => Customer)
   customerId: number;
 }
+
+export interface AddressRelations {}
```
bajtos: Please add navigational properties for existing relations of `Address` model.

packages/repository/src/__tests__/fixtures/models/order.model.ts
```diff
@@ -29,3 +29,7 @@ export class Order extends Entity {
   @belongsTo(() => Customer)
   customerId: number;
 }
+
+export interface OrderRelations {}
```
bajtos: Please add navigational properties for existing relations of `Customer` model.

packages/cli/generators/model/index.js
nabdelgadir: Removed by accident, will add back 👍

examples/todo/src/models/todo.model.ts
nabdelgadir: I agree 👍

nabdelgadir:
@bajtos 
> IMO, when a model already has some relations defined, then the `*Relations` interface should have the navigational properties defined too. We will need these navigational properties for our next steps on the journey to implement the feature "inclusion of related models".
>
> As I see it, #2633 is about automating the task of adding navigational properties. Until it's done, developers will have to define these properties manually, which is exactly what I am asking you to do in this pull request too.

I thought the purpose of #2633 was only to add them to the `TodoList` example and verify it works for that example (and then after it's tested there, then we could add the properties of the `*Relations` interface to the other examples/docs). But I guess adding them here makes more sense, so now they're a part of this PR 👍 

Also, did you mean https://github.com/strongloop/loopback-next/issues/2988 was for automating instead of #2633?

bajtos:
> Also, did you mean #2988 was for automating instead of #2633?

Yes, I meant #2988. Sorry for the confusion!

docs/site/BelongsTo-relation.md
```diff
@@ -212,14 +223,21 @@ export class Category extends Entity {
     super(data);
   }
 }
+
+export interface CategoryRelations {
+  categories?: CategoryWithRelations[];
+}
```
bajtos: I think this interface should include a navigation property for `parent` relation too, WDYT?

```ts
export interface CategoryRelations {
  categories?: CategoryWithRelations[];
  parent?: CategoryWithRelations;
}
```

PR #2966, opened 2019-05-26
Make invocation context parent public and fix optional binding resolution

raymondfeng:
1. InvocationContext.parent (public readonly)
2. Allow `optional` for binding without a value getter

## Checklist

### DIFF
--- a/packages/context/src/interceptor.ts
+++ b/packages/context/src/interceptor.ts
@@ -60,7 +60,9 @@ export class InvocationContext extends Context {
    * @param args - An array of arguments
    */
   constructor(
-    parent: Context,
+    // Make `parent` public so that the interceptor can add bindings to
+    // the request context, for example, tracing id
+    public readonly parent: Context,
     public readonly target: object,
     public readonly methodName: string,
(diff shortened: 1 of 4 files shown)


packages/context/src/interceptor.ts
```diff
@@ -60,7 +60,9 @@ export class InvocationContext extends Context {
    * @param args - An array of arguments
    */
   constructor(
-    parent: Context,
+    // Make `parent` public so that the interceptor can add bindings to
+    // the request context, for example, tracing id
```
bajtos: Is this relying on the fact that interceptor's context has the request context as the parent? I find that rather brittle, what if we introduce another context in between per-request and interceptor/invocation? 

IMO, interceptors wishing to contribute bindings to per-request context should receive that context via DI. See
https://github.com/strongloop/loopback-next/blob/6caceda5a7f8f83725b4fe79447df4d4b9258d1e/packages/rest/src/keys.ts#L229-L231

raymondfeng: Your proposal is better. I can revert the change.

PR #2592, opened 2019-03-15
PoC: TypeScript types and OpenAPI schema to describe navigational properties (inclusion of related models)

bajtos:
This pull request presents a proof-of-concept implementation demonstrating how to describe navigational properties for inclusion of related models.

See the discussion in #2152 for background information and also examples of different approaches I have researched and abandoned before settling down on what's proposed here.

How to review this pull request:

 - Start by reading `_SPIKE_.md`, it contains high-level description of the proposed implementation with code snippets serving as illustrations.
 - Then review the code. The changes are split into multiple commits, you may find it easier to review the patch in smaller chunks.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

PR #2975, opened 2019-05-27
feat(repository-json-schema): enhance getJsonSchema for navig props

### LINKED ISSUE #2630: Enhance `getJsonSchema` to describe navigational properties
Improve the code converting LoopBack Model definitions into JSON Schema documents to optionally describe navigational properties for relations defined by the model. The decision is controlled by a new option passed to `modelToJsonSchema` and
related methods.

```ts
// own properties only
const spec = getJsonSchema(Product);
// include navigational properties too
const spec = getJsonSchema(Product, {includeRelations: true});
```

Add a new property to all relation definitions. This property will signal whether the relations is targeting a single model instance (e.g. BelongsTo, HasOne) or an array of instances (e.g. HasMany).

```diff
 export interface RelationDefinitionBase {
   /**
    * The type of the relation, must be one of RelationType values.
    */
   type: RelationType;

+  /**
+   * True for relations targeting multiple instances (e.g. HasMany),
+   * false for relations with a single target (e.g. BelongsTo, HasOne).
+   * This property is need by OpenAPI/JSON Schema generator.
+   */
+  targetsMany: boolean;

   // ...
}
``` 

Add a new property to `options` object used by `getJsonSchema` and `modelToJsonSchema`:

```diff
 export interface JsonSchemaOptions {
+  includeRelations?: boolean;
   visited?: {[key: string]: JSONSchema};
 }
```

Modify the implementation of `modelToJsonSchema` to include navigational properties when `includeRelations` is set.

See the spike #2592 and especially the commits https://github.com/strongloop/loopback-next/pull/2592/commits/2256435f2affe7c3962267dc4d2a2148cb15a6d5 for more details. Please note the commit shows changes before the review feedback was applied. You need to review the final pull request diff of the PoC before copying any code over.

This story requires us to fix handling of circular references first, see https://github.com/strongloop/loopback-next/issues/2628.

## Acceptance criteria

- [ ] Implement the changes described above
- [ ] Test coverage
- [ ] Update API documentation

samarpanB:
Enhance `getJsonSchema` to describe navigational properties

fix #2630

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -17,6 +17,7 @@ import {
   JsonSchema,
   JSON_SCHEMA_KEY,
   modelToJsonSchema,
+  MODEL_TYPE_KEYS,
 } from '../..';
 import {expectValidJsonSchema} from '../helpers/expect-valid-json-schema';
 
@@ -648,7 +649,7 @@ describe('build-schema', () => {
       };
       MetadataInspector.defineMetadata(
         JSON_SCHEMA_KEY,
-        cachedSchema,
+        {[MODEL_TYPE_KEYS.ModelOnly]: cachedSchema},
         TestModel,
       );
       const jsonSchema = getJsonSchema(TestModel);
@@ -715,6 +716,198 @@ describe('build-schema', () => {
       it('generates the schema without running into infinite loop', () => {
         const schema = getJsonSchema(Category);
         expect(schema).to.deepEqual(expectedSchemaForCategory);
+      });
+    });
+
+    it('converts HasMany and BelongsTo relation links', () => {
+      @model()
+      class Product extends Entity {
+        @property({id: true})
+        id: number;
+
+        @belongsTo(() => Category)
+        categoryId: number;
+      }
+
+      @model()
+      class Category extends Entity {
+        @property({id: true})
+        id: number;
+
+        @hasMany(() => Product)
+        products?: Product[];
+      }
+
+      const expectedSchema: JsonSchema = {
+        definitions: {
+          ProductWithRelations: {
+            title: 'ProductWithRelations',
+            properties: {
+              id: {type: 'number'},
+              categoryId: {type: 'number'},
+              category: {$ref: '#/definitions/CategoryWithRelations'},
+            },
+          },
+        },
+        properties: {
+          id: {type: 'number'},
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/ProductWithRelations'},
+          },
+        },
+        title: 'CategoryWithRelations',
+      };
+      const jsonSchema = getJsonSchema(Category, {includeRelations: true});
+      expect(jsonSchema).to.deepEqual(expectedSchema);
+    });
+
+    it('converts relation links when no other properties there', () => {
+      @model()
+      class Product extends Entity {
+        @property({id: true})
+        id: number;
+
+        @belongsTo(() => CategoryWithoutProp)
+        categoryId: number;
+      }
+
+      @model()
+      class CategoryWithoutProp extends Entity {
+        @hasMany(() => Product)
+        products?: Product[];
+      }
+      const expectedSchema: JsonSchema = {
+        definitions: {
+          ProductWithRelations: {
+            title: 'ProductWithRelations',
+            properties: {
+              id: {type: 'number'},
+              categoryId: {type: 'number'},
+              category: {
+                $ref: '#/definitions/CategoryWithoutPropWithRelations',
+              },
+            },
+          },
+        },
+        properties: {
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/ProductWithRelations'},
+          },
+        },
+        title: 'CategoryWithoutPropWithRelations',
+      };
+
+      // To check for case when there are no other properties than relational
+      const jsonSchemaWithoutProp = getJsonSchema(CategoryWithoutProp, {
+        includeRelations: true,
+      });
+      expect(jsonSchemaWithoutProp).to.deepEqual(expectedSchema);
+    });
+
+    it('gets cached JSON schema with relation links if one exists', () => {
+      @model()
+      class Product extends Entity {
+        @property({id: true})
+        id: number;
+
+        @belongsTo(() => Category)
+        categoryId: number;
+      }
+
+      @model()
+      class Category extends Entity {
+        @property({id: true})
+        id: number;
+
+        @hasMany(() => Product)
+        products?: Product[];
+      }
+
+      const cachedSchema: JsonSchema = {
+        definitions: {
+          ProductWithRelations: {
+            title: 'ProductWithRelations',
+            properties: {
+              id: {type: 'number'},
+              categoryId: {type: 'number'},
+              category: {$ref: '#/definitions/CategoryWithRelations'},
+            },
+          },
+        },
+        properties: {
+          id: {type: 'number'},
+          cachedProp: {type: 'string'},
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/ProductWithRelations'},
+          },
+        },
+        title: 'CategoryWithRelations',
+      };
+      MetadataInspector.defineMetadata(
+        JSON_SCHEMA_KEY,
+        {[MODEL_TYPE_KEYS.ModelWithRelations]: cachedSchema},
+        Category,
+      );
+      const jsonSchema = getJsonSchema(Category, {includeRelations: true});
+      expect(jsonSchema).to.eql(cachedSchema);
+    });
+
+    it('updates same cache with new key if one exists for model', () => {
+      @model()
+      class Product extends Entity {
+        @property({id: true})
+        id: number;
+
+        @belongsTo(() => Category)
+        categoryId: number;
+      }
+
+      @model()
+      class Category extends Entity {
+        @property({id: true})
+        id: number;
+
+        @hasMany(() => Product)
+        products?: Product[];
+      }
+
+      const cachedSchema: JsonSchema = {
+        definitions: {
+          ProductWithRelations: {
+            title: 'ProductWithRelations',
+            properties: {
+              id: {type: 'number'},
+              categoryId: {type: 'number'},
+              category: {$ref: '#/definitions/CategoryWithRelations'},
+            },
+          },
+        },
+        properties: {
+          id: {type: 'number'},
+          cachedProp: {type: 'string'},
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/ProductWithRelations'},
+          },
+        },
+        title: 'CategoryWithRelations',
+      };
+      MetadataInspector.defineMetadata(
+        JSON_SCHEMA_KEY,
+        {[MODEL_TYPE_KEYS.ModelWithRelations]: cachedSchema},
+        Category,
+      );
+      const jsonSchema = getJsonSchema(Category);
+      // Make sure it's not pulling the withrelations key
+      expect(jsonSchema).to.not.eql(cachedSchema);
+      expect(jsonSchema).to.eql({
+        properties: {
+          id: {type: 'number'},
+        },
+        title: 'Category',
       });
     });
   });
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -9,12 +9,14 @@ import {
   ModelDefinition,
   ModelMetadataHelper,
   PropertyDefinition,
+  RelationMetadata,
   resolveType,
 } from '@loopback/repository';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
-import {JSON_SCHEMA_KEY} from './keys';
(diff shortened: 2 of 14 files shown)


samarpanB:
@bajtos Can you please do a quick review and see if I am on right track ? Its a WIP PR. I still need to add new tests.

bajtos:
@samarpanB thank you for the pull request, the changes look reasonable so far. I think you will need to add code setting `targetsMany` property to all relational decorators, see e.g. https://github.com/strongloop/loopback-next/pull/2592/files#diff-64c2bbb6c63151367ef935def7eb54da

examples/log-extension/src/__tests__/acceptance/log.extension.acceptance.ts
```diff
@@ -3,44 +3,43 @@
...
 
 const SequenceActions = RestBindings.SequenceActions;
 
-import {createLogSpy, restoreLogSpy} from '../log-spy';
-import {logToMemory, resetLogs} from '../in-memory-logger';
-
 describe('log extension acceptance test', () => {
   let app: LogApp;
-  let spy: SinonSpy;
+  let spy: SinonSpy<[(string | undefined)?], void>;
```
samarpanB: There was an error showing up while running tests. It needed explicit definition of type for the generic param. So I added it. Rest of the file changes are just reordering of imports done by vscode extension TSHero. I hope that's fine.

examples/log-extension/src/__tests__/unit/providers/log-action.provider.unit.ts
```diff
@@ -3,23 +3,22 @@
...
 } from '../../..';
-import chalk from 'chalk';
-
-import {createLogSpy, restoreLogSpy, createConsoleStub} from '../../log-spy';
 import {logToMemory} from '../../in-memory-logger';
+import {createConsoleStub, createLogSpy, restoreLogSpy} from '../../log-spy';
 
 describe('LogActionProvider with in-memory logger', () => {
-  let spy: sinon.SinonSpy;
+  let spy: sinon.SinonSpy<[(string | undefined)?], void>;
```
samarpanB: There was an error showing up while running tests. It needed explicit definition of type for the generic param. So I added it. Rest of the file changes are just reordering of imports done by vscode extension TSHero. I hope that's fine.

samarpanB:
@bajtos I updated the PR with all the changes and test cases also added now. 

Just a note. I kind of struggled a lot this time to run the tests. 'npm test' kept on failing. First, it was due to missing deps. So, I ran 'npm ci'. But it also failed due to access issues. I am on a linux machine. After a long search/hit-and-trial, I figured that it was not working properly when I was using my own user in the machine. When I switched to root user, it worked perfectly. Do you think this is something we should document somewhere for contributors to save some time in troubleshooting ?

samarpanB:
@bajtos I am not sure why the build is failing. It is showing permission errors. I think that's unrelated.

examples/log-extension/src/__tests__/acceptance/log.extension.acceptance.ts
bajtos: I think these new errors are reported by the newer version of TypeScript compiler that we recently upgraded to. Can you please rebase your patch on top of the latest master? I think the log-extension example should be already fixed by now.

/cc @raymondfeng

bajtos:
> Just a note. I kind of struggled a lot this time to run the tests. 'npm test' kept on failing. First, it was due to missing deps. So, I ran 'npm ci'. 

Yes, you need to run `npm ci` to install dependencies before you can run `npm test`, see https://loopback.io/doc/en/contrib/code-contrib-lb4.html#building-the-project

> But it also failed due to access issues. I am on a linux machine. After a long search/hit-and-trial, I figured that it was not working properly when I was using my own user in the machine. When I switched to root user, it worked perfectly. Do you think this is something we should document somewhere for contributors to save some time in troubleshooting ?

I find that really weird, to be honest. Did you perhaps run `git clone` under root user? I am not aware of any dependencies that would require global install or root privileges. `npm ci` is supposed to install all dependencies locally, inside your checked-out repository.

Besides running `npm install`, we are also running `lerna` to create symlinks between monorepo packages. Is it perhaps possible that these symlinks cannot be created?

Anyhow, it's hard for me to help you without seeing the particular error logs. This is also very off topic in this pull request. Can you open a new issue please where we can take a closer look at the problems you are experiencing?

samarpanB:
> > Just a note. I kind of struggled a lot this time to run the tests. 'npm test' kept on failing. First, it was due to missing deps. So, I ran 'npm ci'.
> 
> Yes, you need to run `npm ci` to install dependencies before you can run `npm test`, see https://loopback.io/doc/en/contrib/code-contrib-lb4.html#building-the-project
> 
> > But it also failed due to access issues. I am on a linux machine. After a long search/hit-and-trial, I figured that it was not working properly when I was using my own user in the machine. When I switched to root user, it worked perfectly. Do you think this is something we should document somewhere for contributors to save some time in troubleshooting ?
> 
> I find that really weird, to be honest. Did you perhaps run `git clone` under root user? I am not aware of any dependencies that would require global install or root privileges. `npm ci` is supposed to install all dependencies locally, inside your checked-out repository.
> 
> Besides running `npm install`, we are also running `lerna` to create symlinks between monorepo packages. Is it perhaps possible that these symlinks cannot be created?
> 
> Anyhow, it's hard for me to help you without seeing the particular error logs. This is also very off topic in this pull request. Can you open a new issue please where we can take a closer look at the problems you are experiencing?
Yes I agree it's off-topic. I'll raise a separate issue if I encounter this again. I just wanted a heads up if I was doing something wrong. Thanks a lot for the direction @bajtos

examples/log-extension/src/__tests__/acceptance/log.extension.acceptance.ts
samarpanB: I already did. Let me run the tests again and check.

samarpanB: works. I removed my changes from here.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -717,5 +718,182 @@ describe('build-schema', () => {
...
+          },
+        },
+        properties: {
+          id: {type: 'number'},
+          products: {
+            type: 'array',
+            items: {$ref: '#/definitions/ProductWithRelations'},
+          },
+        },
+        title: 'CategoryWithRelations',
```
bajtos: Can you list the schema fields in the order `title`, `properties`, `definitions` please? I find it difficult to read `expectedScheme` when the order is reversed.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -717,5 +718,182 @@ describe('build-schema', () => {
...
+      };
+      const jsonSchema = getJsonSchema(Category, {includeRelations: true});
+      expect(jsonSchema).to.deepEqual(expectedSchema);
+
+      // To check for case when there are no other properties than relational
```
bajtos: Please move this check to a new test case. 

It's a best practice to keep one (logical) assertion per test.

packages/repository/src/relations/relation.types.ts
```diff
@@ -22,6 +22,13 @@ export interface RelationDefinitionBase {
    */
   type: RelationType;
 
+  /**
+   * True for relations targeting multiple instances (e.g. HasMany),
+   * false for relations with a single target (e.g. BelongsTo, HasOne).
+   * This property is needed by OpenAPI/JSON Schema generator.
+   */
+  targetsMany?: boolean;
```
nabdelgadir: I think that this should be a required property like in the issue description. What do you think?

packages/repository-json-schema/src/keys.ts
```diff
@@ -6,10 +6,15 @@
 import {MetadataAccessor} from '@loopback/metadata';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 
+export const enum MODEL_TYPE_KEYS {
```
nabdelgadir: nice 👍

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -717,5 +718,182 @@ describe('build-schema', () => {
...
+        },
+        title: 'CategoryWithRelations',
+      };
+      MetadataInspector.defineMetadata(
+        JSON_SCHEMA_KEY,
+        {[MODEL_TYPE_KEYS.ModelWithRelations]: cachedSchema},
+        Category,
+      );
+      const jsonSchema = getJsonSchema(Category);
+      // Make sure its not pulling the withrelations key
```
nabdelgadir: Nitpick: `its` -> `it's`

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
samarpanB: I had to order it like that because the getJsonSchema method is returning the data in that order and if I don't keep it like that, the test fails. Is there any other way out here ?

packages/repository/src/relations/relation.types.ts
samarpanB: I kept it optional for backward compatibility. But if you think otherwise I can make this as required. Thoughts @bajtos ?

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
samarpanB: fixed.

packages/repository-json-schema/src/keys.ts
samarpanB: thanks :)

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
samarpanB: done

packages/repository/src/relations/relation.types.ts
nabdelgadir: That's a good point, let's wait for @bajtos' thoughts, but can you please still add it to the relation definitions? e.g.:

```ts
export interface HasManyDefinition extends RelationDefinitionBase {
  type: RelationType.hasMany;
  targetsMany: true;
```

samarpanB: added to relation definitions.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
bajtos: I see. Let's not worry about this too much, we can keep the current form.

packages/repository/src/relations/relation.types.ts
bajtos: > I kept it optional for backward compatibility. 

Good point!

I agree that making `targetsMany` a required property can break existing applications or 3rd-party extensions. On the other hand, I would be surprised if any such app or extension actually did exist.

Here is the important part: what is the expected behavior when `targetsMany` is not set? I don't think it's a good idea to default to `false`.

I am proposing the following:
- Add a comment starting with `TODO(semver-major)` explaining that we should make `targetsMany` required in LB5
- Modify the code reading `targetsMany` to throw a helpful error when `targetsMany` does not exist or is `undefined`.  The most important place to change is in `modelToJsonSchema`, but there may be other places to update too.

bajtos: Preferably add a unit test to verify that `modeltoJsonSchema` can handle missing `targetsMany` and demonstrate what happens in such case.

samarpanB: > Modify the code reading targetsMany to throw a helpful error when targetsMany does not exist or is undefined

Don't you think throwing error in this case would also mean we are making it backward incompatible ?

bajtos: As I understand it, the error will be thrown only when `modeltoJsonSchema` is called with the new option `includeRelations`. Since this option did not exist before, there shouldn't be any existing code setting it to `true` and thus there shouldn't be any code affected by the new error being thrown.

samarpanB: Hmmm. Got it. Let add some error handling and a test case to verify it as well. This makes sense.

samarpanB: Done. Please check now.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -203,6 +204,31 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
   return result;
 }
 
+/**
+ * Checks and return property schema definition for the relation
+ * @param relMeta Relation metadata object
+ * @param targetRef Schema definition for the target model
+ */
+export function getPropDefForRelatedModel(
```
bajtos: We prefer to use full names instead of abbreviations.

How about the following?

```ts
export function getNavigationalPropertyForRelation(
  relationMeta: RelationMetadata,
  targetSchema: JSONSchema
): JSONSchema {
  // ...
}
```

samarpanB: Yeah. That's better. Changing it. Will rebase and squash as well.

nabdelgadir:
Thanks for the contribution @samarpanB! It's now landed 👏

Issue #2940, opened 2019-05-22
update @loopback\authentication documentation


emonddr:
## Description / Steps to reproduce / Feature proposal

The REAME.md of the @loopback/authentication module needs to be updated
to discuss the new authentication architecture.

## Current Behavior
Shows an authentication example that makes use of npm modules : passport passport-http and the old strategy adapter.

## Expected Behavior
Explain more of the architecture of the authentication module (not specific implementations of strategies), and provide links to examples which provide ready-to-run applications and detailed explanations of implementations.

## Acceptance Criteria 

- [x] Remove information that is not applicable or not relevant in the current `authentication/README.md` file.
- [] ~~Re-work https://github.com/strongloop/loopback-next/blob/master/packages/authentication/docs/authentication-system.md into standalone pages for Application and Extension developers on loopback.io.~~
- [x] Create tutorial on enabling `jwt` authentication with shopping cart example.
- [x] Show how to set up basic authentication in the docs and point to the above tutorial. ~~If we can showcase basic authentication strategy in the README without too much information, then we can include some code snippets there. Otherwise point to docs.~~

- [x] document briefly discusses the `AuthenticationStrategy` interface, and the `extensionPoint/extensions pattern`,  how `custom authentication strategies can be registered`, how a `custom sequence` must be created, and how `controller methods must be decorated` in order to authenticate endpoints in a LoopBack application. ~~(I don't see the point in going into a lot of step by step details if we are providing a link to examples whose README.md files will go into details). Include basic usage as well.~~

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

jannyHou:
@emonddr Thank you for creating the doc issue, for the 4 tasks in the acceptance criteria:

1 - I agree 👍 
2,3 - Are we creating 2 example under https://github.com/strongloop/loopback-next/tree/master/examples?
4 - sounds good, if the doc PR get landed before #2311 closed, we can link to the ongoing feature branch and update the link after the adapter feature done.

emonddr:
@jannyHou 
> 2,3 - Are we creating 2 example under https://github.com/strongloop/loopback-next/tree/master/examples? 

we can either have 2 separate examples showing separate custom authentication strategies or we can have 1 `example\authentication` example which showcases the 2 different strategies.

emonddr:
Other LoopBack4 documents which will need to be revised:
[...]

PR #2977, opened 2019-05-27
docs: update docs on authentication package

emonddr:
Add a new section on loopback.io about AuthenticationComponent

Associated with : https://github.com/strongloop/loopback-next/issues/2940

## Checklist


emonddr:
This is a work on progress and is not ready for review. Just saving my work for today.

emonddr:
I've taken a first pass at explaining the authentication component's main pieces, and what a developer has to do for authentication. I haven't `yet` used any charts from our design document : https://github.com/strongloop/loopback-next/blob/master/packages/authentication/docs/authentication-system.md.

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+the authentication component named `AuthenticationComponent`.
+
+```ts
+export class MyApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options?: ApplicationConfig) {
+    super(options);
+
+    ...
```
dhmlau: @bajtos suggested, at some point, that we should make the `ts` code snippet a valid one. so, my suggested change would be:
```
// ... 
```

instead of simply 
```...```

File: docs/site/Loopback-component-authentication.md [hunk 2]
```diff
@@ -0,0 +1,600 @@
...
+component.
+
+{% include note.html content="
+For a description of an `Authorization` process, please see <a href='https://loopback.io/doc/en/lb4/Loopback-component-authorization.html'>Authorization</a>.
+" %}
+
+## Authentication Component
+
+To utilize `authentication` in an application `application.ts`, you must load
+the authentication component named `AuthenticationComponent`.
```
dhmlau: Apologize that I'm not familiar with this feature. So the name has to be `AuthenticationComponent`? That means we cannot have more than one Authentication component?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+     */
+
+    return creds;
+  }
+}
+```
+
+As you can see in the example, an authentication strategy can inject custom
+services to help it accomplish certain tasks. A complete example of a **basic**
+authentication strategy can be found
```
dhmlau: Suggested change (see if it makes sense to you):
See the complete example for [basic authentication strategy](link1) and [jwt authentication strategy](link2)

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+```
+
+## Authentication Decorator
+
+Once you have a **registered** authentication strategy with a **unique** name,
+then you are ready to use the
+[Authentication Decorator](https://loopback.io/doc/en/lb4/Decorators_authenticate.html)
+on your controller endpoints.
+
+They decorator's syntax is :
```
dhmlau: typo: They -> The

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+
+```ts
+@authenticate(strategyName: string, options?: Object)
+```
+
+The strategy `name` is **mandatory** and the `options` object is **optional**.
```
dhmlau: I think it's kind of understood and no need for this paragraph?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+  whoAmI(): string {
+    return this.userProfile.id;
+  }
+}
+```
+
+An example of the decorator when options **are** specified looks like this:
+
+```ts
+@authenticate('basic', { propertyOne : "foo", propertyTwo:"bar"})
```
dhmlau: nitpick: do you have some more realistic values that our users might want to use the options for?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,600 @@
...
+Parameters of the `@authenticate` decorator can be obtained via dependency
+injection using the `AuthenticationBindings.METADATA` binding key. It returns
+data of type `AuthenticationMetadata` provided by `AuthMetadataProvider`. The
+`AuthenticationStrategyProvider` in the previous section makes use of
+`AuthenticationMetadata` to figure out what **name** you specified as a
+parameter in the `@authenticate` decorator of a specific controller endpoint.
+
+## Adding an Authentication Action to a Sequence
+
+In a LoopBack 4 REST application, each request passes through a stateless
```
dhmlau: does it only applicable to LB4 application with REST server? I find it a bit strange to refer it as "LoopBack 4 REST application".

docs/site/Loopback-component-authentication.md
emonddr: @dhmlau 

> Apologize that I'm not familiar with this feature. So the name has to be `AuthenticationComponent`? That means we cannot have more than one Authentication component?

So most of the logic of @loopback/authentication ( the authentication metadata provider, the authentication action provider, and the authentication strategy provider ) is available in a component named `AuthenticationComponent`, and must be loaded by an application.  

The user needs to create and register their own custom authentication strategies (user can have many of these), and implement their own token service or user service etc, and create a custom sequence that places the authentication action in a specific place.

But the core `authentication framework logic` is stored in 1 component.

docs/site/Loopback-component-authentication.md
emonddr: @dhmlau 

> nitpick: do you have some more realistic values that our users might want to use the options for?

Not really because we don't ship any authentication strategies at all ( just the interface). I could put a made up options object `{ useCookie: true}` or `{session:true}` or `{ enforceLongPassword: true}` but this would give them the wrong impression that our interface supports this kind of option.  @jannyHou @raymondfeng , what are your thoughts on this?

docs/site/Loopback-component-authentication.md
emonddr: @dhmlau , I see what you mean. A user can create a gRPC server I guess, and not a typical REST API server which uses HTTP requests. Are people using LoopBack for more than REST servers at the moment?

docs/site/Loopback-component-authentication.md
emonddr: @dhmlau 

> I think it's kind of understood and no need for this paragraph?

You are right. Sometimes I like writing for the lowest common denominator (somebody new to TypeScript). I will remove it.

docs/site/Loopback-component-authentication.md
emonddr: @dhmlau ^^ :)

docs/site/Loopback-component-authentication.md
dhmlau: i think they can, right? @raymondfeng

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,596 @@
...
+```
+
+An example of the decorator when options **are** specified looks like this:
+
+```ts
+@authenticate('basic', { propertyOne : "foo", propertyTwo:"bar"})
+```
+
+{% include tip.html content="
+To avoid repeating the same options in the <b>@authenticate</b> decorator for many endpoints in a controller, you can instead define global options which can be injected into an authentication strategy thereby allowing you to avoid specifying the options object in the decorator itself. For controller endpoints that need to override a global option, you can specify it in an options object passed into the decorator. Your authentication strategy would need to handle the option overrides.
```
jannyHou: It would be good if we can show a code example here. While I also realized we don't have an auth option sample yet, maybe use the greeter extension https://github.com/strongloop/loopback-next/tree/master/examples/greeter-extension#implement-an-extension ?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+      [AuthenticationBindings.AUTH_ACTION.key]: AuthenticateActionProvider,
+      [AuthenticationBindings.STRATEGY.key]: AuthenticationStrategyProvider,
+      [AuthenticationBindings.METADATA.key]: AuthMetadataProvider,
+    };
+  }
+}
+```
+
+As you can see, there are a few
+[providers](https://loopback.io/doc/en/lb4/Creating-components.html#providers)
```
nabdelgadir: suggestion: `https://loopback.io/doc/en/lb4/Creating-components.html#providers` -> `Creating-components.md#providers`

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+### AuthenticationStrategy Interface
+
+The `AuthenticationComponent` declares a common authentication strategy
+interface named `AuthenticationStrategy`.
+
+```ts
+export interface AuthenticationStrategy {
+  /**
+   * The 'name' property is a unique identifier for the
+   * authentication strategy ( for example : 'basic', 'jwt', etc)
```
nabdelgadir: Nitpick: `( for example :` -> `(for example:`

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+As you can see in the example, an authentication strategy can inject custom
+services to help it accomplish certain tasks. See the complete examples for
+[basic authentication strategy](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/fixtures/strategies/basic-strategy.ts)
+and
+[jwt authentication strategy](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/fixtures/strategies/jwt-strategy.ts).
+
+The `AuthenticationComponent` component also provides two **optional** service
+interfaces which may be of use to your application :
```
nabdelgadir: Nitpick: ditto about space before the colon

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+[UserService](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/user.service.ts)
+and
+[TokenService](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/token.service.ts).
+
+Developers that wish to use [passport](http://www.passportjs.org/)-based
+authentication strategies, please see [Passport Authentication Strategies](TBD).
```
nabdelgadir: Is this link going to be added as part of this PR?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,596 @@
...
+
+Once you have a **registered** authentication strategy with a **unique** name,
+then you are ready to use the
+[Authentication Decorator](https://loopback.io/doc/en/lb4/Decorators_authenticate.html)
+on your controller endpoints.
+
+The decorator's syntax is :
+
+```ts
+@authenticate(strategyName: string, options?: Object)
```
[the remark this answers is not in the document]
emonddr: @jannyHou  thanks for your feedback.

> A more accurate type would be Promise<UserProfile> 

When creating this diagram, space is very limited, so I chose to avoid using stuff like:
`this.` and `Promise<  >` to save space, and give a general idea of the logic.  The documentation text in the sections below the diagram will show the correct code.

> nitpick: 'some' --> 'amodel' or 'somemodel'. Just a suggestion based on personally preferred naming conversion. 

The `lb4 controller` CLI creates controller files is this format: someName.controller.ts . And all the artifacts follow the same pattern. I think my filename is ok.


> Maybe re-organize the rectangles in a more sorted way? 

It took a long time to create the diagram and all its pieces and labels, so I am reluctant iteratively move all the PowerPoint shapes/labels/arrows around  `unless` someone sends me a `pencil and paper picture` of a re-design.

docs/site/Loopback-component-authentication.md
jannyHou: The default sequence mentioned [here](https://github.com/strongloop/loopback-next/pull/2977/files#diff-f95d62f4892a8c070795ee268ae7ca15R335) is for REST server only.

And please note that in the `AuthenticationStrategy` interface, the `Request` type is imported from `@loopback/rest`, see https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/types.ts#L7

So mentioning `REST application` is ok to me here.

Those services like `UserService` and `TokenService` are reusable for other Clients, but the authentication component does target on the REST request.

docs/site/Loopback-component-authentication.md
emonddr: @jannyHou  ^^ ,

Can we point to a `unit test` that passes in options? https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/__tests__/unit/providers/auth-metadata.provider.unit.ts#L16

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+title: 'Authentication'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Loopback-component-authentication.html
+---
+
+## Overview
+
+Security is of paramount importance when developing a web or mobile application
+and usually consists of two distinct pieces :
```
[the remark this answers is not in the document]
emonddr: @nabdelgadir 

> Just some nitpicks but 👍. When you're linking to another docs on the website, if you use {file-name}.md, then, in vscode, it goes to the file instead of the website so that's why I recommend changing those (there are more of them in the file that I didn't comment on).

Umm...I cannot find any use of fileName.md in my document.

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+" %}
+
+This document describes the details of the LoopBack 4 `Authentication`
+component.
+
+![authentication_overview](./imgs/authentication_overview.png) Basically, to
+secure your API endpoints, you need to: create/register a custom authentication
+strategy with a unique **name**, insert the authentication action in a custom
+sequence, and then decorate the endpoints of a controller with the
+`@authenticate(strategyName, options?)` authentication decorator. The
```
raymondfeng: Let's use a bullet list.

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: @nabdelgadir and I spoke offline. I wasn't sure what she was asking in her comment above.

I understand now that Nora was stating that I should not reference `.html` files to loopback.io documentation, but that I should instead try to use their respective `.md` files.

I distinctly remember someone telling me that inside that I `should` reference the loopback.io `.html` files.  I've done it before in : https://raw.githubusercontent.com/strongloop/loopback-next/master/docs/site/submitting_a_pr.md .

Both work...

@jannyHou @raymondfeng @dhmlau @b-admike , is there a preference?

docs/site/Loopback-component-authentication.md
raymondfeng: Agree - now extra spaces are needed after `(`, or before `:`.

```
* authentication strategy (for example: 'basic', 'jwt', etc)
```

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+  /**
+   * The 'authenticate' method takes in a given request and returns a user profile
+   * which is an instance of 'UserProfile'.
+   * (A user profile is a minimal subset of a user object)
+   * If the user credentials are valid, this method should return a 'UserProfile' instance.
+   * If the user credentials are invalid, this method should throw an error
+   * If the user credentials are missing, this method should throw an error, or return 'undefined'
+   * and let the authentication action deal with it.
+   *
+   * @param request
```
raymondfeng: Add some description for `request`, such as:

`@param request - Express request object`

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+- Authorization
+
+`Authentication` is a process of verifying someone's identity before a protected
+endpoint is accessed.
+
+`Authorization` is a process of verifying the actions that an authenticated user
+is allowed to perform on a protected endpoint.
+
+{% include note.html content="
+For a description of an `Authorization` process, please see <a href='https://loopback.io/doc/en/lb4/Loopback-component-authorization.html'>Authorization</a>.
```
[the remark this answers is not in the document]
emonddr: > could be simplified to please see [Authorization](Loopback-component-authorization.md) 

Actually in a note.html the `string value` cannot contain MD-style formatting...it has to be  HTML-style .

docs/site/Loopback-component-authentication.md
jannyHou: @emonddr Cool the unit test is much more relevant. Looks good to me.

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
dhmlau: _I think_ (at least) if the referencing file is coming from loopback-next, it should be `.md`. 
See https://raw.githubusercontent.com/strongloop/loopback-next/master/docs/site/Concepts.md for examples.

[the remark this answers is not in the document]
jannyHou: I prefer to use the relative link. The external url can be changed due to

- A refactor that changes the host: like the PR that Raymond just merged, the apidocs are moved from `apidocs.strongloop.com` to `loopback.io`

- An update of the permanent link metadata like https://github.com/strongloop/loopback-next/pull/2977/files#diff-f95d62f4892a8c070795ee268ae7ca15R6

Pointing to a relative path saves us the updating effort (some times it's huge).

docs/site/Loopback-component-authentication.md
jannyHou: I was also searching for some meaningful options for the basic auth in `passport-http`, but didn't find any...like what Dom said, they are all related to the `passport` middleware instead of the strategy itself.

dhmlau: got it. maybe at least mention those options are specific to the strategy the user select?

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
dhmlau: You can.  Please see https://github.com/strongloop/loopback-next/blob/master/docs/site/tutorials/soap-calculator/soap-calculator-tutorial-scaffolding.md (around line 40) for an example.

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,603 @@
...
+" %}
+
+This document describes the details of the LoopBack 4 `Authentication`
+component.
+
+![authentication_overview](./imgs/authentication_overview.png) Basically, to
```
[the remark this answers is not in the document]
dhmlau: Thanks for adding the diagram. I was thinking of a more higher-level diagram to help users to understand the big picture first. :)
I was thinking something along the line of the diagram below (it's probably not reflecting what's actually happening, but just an illustration). Does it make sense? 
Screen Shot 2019-06-07 at 5 36 07 PM

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
jannyHou: > When creating this diagram, space is very limited, so I chose to avoid using stuff like:
this. and Promise< > to save space

👍 Fair enough.

And the naming and diagram comments are just nitpicks 🙂 feel free to ignore.

docs/site/Loopback-component-authentication.md
emonddr: @nabdelgadir ,

> Is this link going to be added as part of this PR? 

I don't yet know the url or file name that @jannyHou is going to have for her PR... which will cover the passport stuff.  @jannyHou , can you offer a best-guess to your new url?

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
dhmlau: relative link for `.md` file works for me. 👍

docs/site/Loopback-component-authentication.md
dhmlau: If the link doesn't exist, is it better not to mention it until @jannyHou added her docs?

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: I stand corrected...cool. thanks Nora/Diana . :)

docs/site/Loopback-component-authentication.md
emonddr: Removing (for now):

```
Developers that wish to use [passport](http://www.passportjs.org/)-based
authentication strategies, please see [Passport Authentication Strategies](TBD).
```

docs/site/Loopback-component-authentication.md
emonddr: @jannyHou 

> I was also searching for some meaningful options for the basic auth in passport-http, but unfortunately didn't find any...like what Dom said, they are all related to the passport middleware instead of the strategy itself. 

Actually, the focus of the discussion is not on **passport** stuff in this discussion. The `@authenticate(strategyName, options?)` decorator allows for passing in `options`. This gives the developer of custom authentication strategies a way to pass options in...if needed.

(Buy Janny is correct that the passport-based api have options they can pass in too...)

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: I will use references the .md files for docs instead of html :)

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: Perhaps we should have a `whiteboard session` and see how we can move between highlevel and lowlevel diagrams and when...

docs/site/Loopback-component-authentication.md
raymondfeng: We can use something as follows:

```ts
@authenticate('basic', { /* some options for the strategy */})
```

docs/site/Loopback-component-authentication.md
raymondfeng: The sequence so far only applies to RestServer. Maybe we can say `In a LoopBack 4 application with REST API endpoints`.

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
dhmlau: It's just a suggestion.  Since I don't have much background on the detailed design, so a higher level diagram helps me better to understand.  Maybe it's just personal preference.

docs/site/Loopback-component-authentication.md
emonddr: @jannyHou , I've changed my mind about pointing to the unit test. It doesn't really show global options being overridden by `request` options from the decorator. Perhaps after this PR lands, we can have a follow up PR to add a section about injecting global options into the  strategy, and then the strategy checking for overriding options from the decorator and then overriding the global options. I have an example of this I can dig up from one of my workspaces. But I won't be able to prepare it nicely today for @raymondfeng  to review.

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: @dhmlau 

The action provider isn't a composite part of the strategy provider, and the metadata provider can be used from anywhere (actually all of them are set up like this), so it is more accurate to have them separate from each other than contained in each other.

docs/site/Loopback-component-authentication.md
jannyHou: @emonddr Sure it would be good to have an example and point to it. +1 for improving it in another PR.

docs/site/Loopback-component-authentication.md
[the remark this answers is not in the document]
emonddr: @dhmlau @jannyHou how about this for a high level?
image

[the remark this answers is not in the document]
emonddr: 1. wondering if I should show custom authentication strategies at all in this diagram and just leave it as `extensions`
2. wondering if I should specify the parameters of the decorator: strategyName:string, options?:object .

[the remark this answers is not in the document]
emonddr: actually this one is better

image

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+## Overview
+
+Security is of paramount importance when developing a web or mobile application
+and usually consists of two distinct pieces:
+
+- Authentication
+- Authorization
+
+`Authentication` is a process of verifying someone's identity before a protected
+endpoint is accessed.
```
raymondfeng: `endpoint` -> `resource`?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+
+`Authorization` is a process of verifying the actions that an authenticated user
```
raymondfeng: `verifying ...` -> `deciding if a user can perform an action on a protected resource`?

raymondfeng: Maybe add links such as:
- https://en.wikipedia.org/wiki/Authentication
- https://en.wikipedia.org/wiki/Authorization

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+{% include note.html content="
+For a description of an `Authorization` process, please see [Authorization](Loopback-component-authorization.md).
+" %}
+
+This document describes the details of the LoopBack 4 `Authentication`
+component.
+
+Here is a **high level** overview of the authentication component.
+
+![authentication_overview_highlevel](./imgs/authentication_overview_highlevel.png)
```
raymondfeng: Let's crop the diagram so that there is no blank area to the left:

image

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+
+It is made up of a decorator, a few providers, and an extension point to which
+authentication strategies must be registered.
```
raymondfeng: Terms like decorator/provider/extension point seem to be too low level here. We can say:

- Decorators to express authentication requirement on controller methods
- A provider to access method-level authentication metadata
- An action in the REST sequence to enforce authentication
- An extension point to discover all authentication strategies and handle the delegation

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+
+Here is a **detailed** overview of the authentication component.
+
+![authentication_overview_detailed](./imgs/authentication_overview_detailed.png)
```
raymondfeng: 👍

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+
+Basically, to secure your API endpoints, you need to:
+
+- create/register a custom authentication strategy with a unique **name**
+- insert the authentication action in a custom sequence
+- decorate the endpoints of a controller with the
+  `@authenticate(strategyName, options?)` authentication decorator
+
```
raymondfeng: Maybe change the order as follows:
1. decorate (app developer)
2. add authentication action (app developer/cli)
3. create custom authentication strategy (extension developer)
4. register a custom authentication strategy (app developer)

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+      debug('%s result -', route.describe(), result);
+      this.send(response, result);
+    } catch (error) {
+      this.reject(context, error);
+    }
+  }
+}
+```
+
+Authentication is **not** part of the default sequence of actions, so you must
```
raymondfeng: By default, ...

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+    this.providers = {
+      [AuthenticationBindings.AUTH_ACTION.key]: AuthenticateActionProvider,
+      [AuthenticationBindings.STRATEGY.key]: AuthenticationStrategyProvider,
+      [AuthenticationBindings.METADATA.key]: AuthMetadataProvider,
+    };
+  }
+}
+```
+
+As you can see, there are a few [providers](Creating-components.md#Providers)
```
nabdelgadir: `Providers` -> `providers`.

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+
+```ts
+registerAuthenticationStrategy(this, BasicAuthenticationStrategy);
+```
+
+## Authentication Decorator
+
+Once you have a **registered** authentication strategy with a **unique** name,
+then you are ready to use the
+[Authentication Decorator](decorators/Decorators_authenticate.md) on your
```
nabdelgadir: I don't think you need the `decorators/` part

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+    }
+  }
+}
+```
+
+Notice the new dependency injection in the sequence's contructor.
+
+```ts
+    @inject(AuthenticationBindings.AUTH_ACTION)
+    protected authenticateRequest: AuthenticateFn,
```
nabdelgadir: Can you un-indent this code block?

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,611 @@
...
+```ts
+this.sequence(MyAuthenticatingSequence);
+```
+
+## Summary
+
+We've gone through the main steps for adding `authentication` to your LoopBack 4
+application.
+
+Your application `application.ts` should look similar to this:
```
nabdelgadir: remove `application`?

docs/site/Loopback-component-authentication.md
emonddr: > Decorators

A decorator

docs/site/Loopback-component-authentication.md
emonddr: @nabdelgadir 

The MD file is in the `decorators` sub-directory. And my vscode Markdown editor complains with an error, if I don't put the sub-directory name.

nabdelgadir: Okay, as long as it also works on the site 👍

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,631 @@
...
+      controllers: {
+        dirs: ['controllers'],
+        extensions: ['.controller.js'],
+        nested: true,
+      },
+    };
+  }
+```
+
+You can find a working **example** of a LoopBack 4 shopping cart application
```
dhmlau: nitpick: I'm not quite sure if we need to bold the "example" word.. similar to a few incidents towards the end of the last section.  It could be just personal preference.

docs/site/Loopback-component-authentication.md
```diff
@@ -0,0 +1,637 @@
...
+In order for your custom authentication strategy to be **found**, it needs to be
+**registered**.
+
+**Registering** a custom authentication strategy `BasicAuthenticationStrategy`
+as an extension of the
+`AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME` extension
+point in an application `application.ts` is as simple as:
+
+```ts
+registerAuthenticationStrategy(this, BasicAuthenticationStrategy);
```
b-admike: can we show the import for `registerAuthenticationStrategy` here?

Issue #2644, opened 2019-03-25
Wrong Model name in Repository/Controller by CLI when a Model name has number


nabdelgadir:
## Description / Steps to reproduce / Feature proposal

Say we have model `Model1` as follows, generated through `lb4 model`:
```ts
export class Model1 extends Entity {}
```
The file name is `model-1.model.ts`. Then when generating a repository with `lb4 repository`:

```
$ lb4 repository
? Please select the datasource DsDatasource
? Select the model(s) you want to generate a repository *Model_1*
? Please select the repository base class DefaultCrudRepository (Legacy juggler bridge)
   create src/repositories/model-1.repository.ts
   update src/repositories/index.ts
```

## Current Behaviour

The generated `model-1.repository.ts` file uses `Model_1` instead of `Model1`
```ts
import {DefaultCrudRepository} from '@loopback/repository';
import {Model_1} from '../models'; 
import {DsDataSource} from '../datasources';
import {inject} from '@loopback/core';

export class Model_1Repository extends DefaultCrudRepository<
  Model_1,
  typeof Model_1.prototype.id
> {
  constructor(
    @inject('datasources.ds') dataSource: DsDataSource,
  ) {
    super(Model_1, dataSource);
  }
}
```
This also happens in the generated `model-1.controller.ts` file.

## Expected Behaviour

It should match the model class name and use `Model1` instead of `Model_1` in both the repository and controller.

jannyHou:
The model name should be consistent in 4 places:

*Note(naming conversion): The camel case model name like `FooBarBaz` will result in a model file named as `foo-bar-baz`*

Assume the correct name is exactly the input from `lb4 model`: `Model1`
- model file name should be `model-1.model.ts`
- model class name should be `Model1`
- `lb4 repository` prompt: `? Select the model(s) you want to generate a repository Model1`
- repository: `import {Model1} from '../models';`
- `lb4 controller` prompt: `? Select the model(s) you want to generate a repository Model1`
- controller: `import {Model1} from '../models';`

PR #2989, opened 2019-05-29
fix(cli): change model file naming convention

agnes512:
connect to https://github.com/strongloop/loopback-next/issues/2644

- Change the naming convention of Class so that the class name inputs: ```model1, Model1, model_1, Model___1``` (```mode-1``` is an invalid input) will have **the same** model name: ```Model1``` and the same file name ```model1.model.ts```.
- Add prompt warning msg that the input will get modified when it doesn't fully follow the naming convention.
- Add prompt msg that shows which class/file names will get generated  in advance.
- The prompt message now shows the created model name instead of the input.
- Unify naming functions as ```toFileName```  ```toVarName``` instead of `kebabCase` etc. Easier for maintenance.

## Checklist


agnes512:
|   	|   	|   	|   	|   	|
|:-:	|:-:	|:-:	|---	|---	|
|   	|   	|   	|   	|   	|
|   	|   	|   	|   	|   	|
|   	|   	|   	|   	|   	|

jannyHou:
A summary of our discussion yesterday:

Given the model `name` provided when creating the model by `lb4 model`, the class name and file name are determined as:

- class name: the pascal case of `name`
- file name: the kebab case of `name`

When loading the models as artifacts, it's much easier to get the file name instead of parsing the class name, but since `name` is not saved anywhere, it's not possible to restore the class name from the file name. 

So we are thinking of generating the **file name** from **class name** instead of the `name` input from the prompt. @strongloop/loopback-maintainers Any thought?

raymondfeng:
IMO, the simplest solution is to keep `m1` as-is in file names, `m1.model.ts` instead of `m-1.model.ts`.

I also hate the fact that `lb4 app m1` creates a directory as `m-1`.

agnes512:
> IMO, the simplest solution is to keep `m1` as-is in file names, `m1.model.ts` instead of `m-1.model.ts`.
> 
> I also hate the fact that `lb4 app m1` creates a directory as `m-1`.

I think use the input as the its file name is doable. But it might run into some tricky issues like, the generator will take `m___1` as a valid input. Then the file name will be `m___1.model.ts`.
I am thinking to use the same convention to name the files. Then change the first letter to lower case: 
|   	|   Input	|   Model Name	|  Model File Name |  Repo File Name 	|
|:----	|:------:	|:------:	|-------	|-------	|
|   `current`	|   `m1`	|   `M1`	|  `m-1.model.ts` 	|  `m-1.repository.ts` 	|
|   	|  `m_1` 	|  `M_1` 	|   `m-1.model.ts`	|   `m-1.repository.ts`	|
|   `solution B`	|   `m1`	|   `M1`	|   `m1.model.ts`	|   `m-1.repository.ts`	|
|   |  `M_1NN` | `M_1Nn`  |`m_1Nn.model.ts`   | `m_1.repository.ts`  |


And the rest of file naming are using the `kababCase` as well. We can change it too.

dhmlau:
The proposal for `solution B.1` looks good to me. Thanks.

raymondfeng:
@agnes512 Why is this PR closed? If you want to add more commits or push new commits, please keep it open as github will automatically picks up changes from the PR branch.

agnes512:
> @agnes512 Why is this PR closed? If you want to add more commits or push new commits, please keep it open as github will automatically picks up changes from the PR branch.

That was an accident

packages/cli/lib/utils.js
```diff
@@ -113,13 +115,14 @@ exports.validateNotExisting = function(projDir) {
...
-  return pascalCase(name);
+  return pascalCase(camelCase(name));
 };
 
 exports.lowerCase = lowerCase;
-exports.kebabCase = _.kebabCase;
+exports.toFileName = _.kebabCase;
 exports.pascalCase = pascalCase;
 exports.camelCase = camelCase;
+exports.uTilCamelCase = uTilCamelCase;
```
nabdelgadir: Why is it `uTil` instead of `util`?

agnes512: fixed

packages/cli/generators/controller/index.js
```diff
@@ -207,12 +207,11 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
...
     this.artifactInfo.className = utils.toClassName(this.artifactInfo.name);
     this.artifactInfo.outFile =
-      utils.kebabCase(this.artifactInfo.name) + '.controller.ts';
+      utils.toFileName(this.artifactInfo.name) + '.controller.ts';
     if (debug.enabled) {
       debug(`Artifact output filename set to: ${this.artifactInfo.outFile}`);
     }
-
-    this.artifactInfo.modelVariableName = utils.camelCase(
+    this.artifactInfo.modelVariableName = utils.utilCamelCase(
```
raymondfeng: Why do we rename it to `utilCamelCase`?

packages/cli/lib/utils.js
```diff
@@ -17,13 +17,15 @@ const _ = require('lodash');
...
 const promisify = require('util').promisify;
-const camelCase = require('change-case').camelCase;
+const utilCamelCase = require('change-case').camelCase;
 const pluralize = require('pluralize');
 const urlSlug = require('url-slug');
 const validate = require('validate-npm-package-name');
 const Conflicter = require('yeoman-generator/lib/util/conflicter');
 const connectors = require('../generators/datasource/connectors.json');
 const stringifyObject = require('stringify-object');
+const camelCase = require('camelcase');
```
raymondfeng: It's very confusing to have two `camelCase`? What's the difference between `camelCase` and `change-case.camelCase`?

raymondfeng: I believe that `lodash` already supports `camelCase`.

raymondfeng: I found out that `camelCase` is a smaller package extracted from `change-case`. So `camelCase` and `change-case.camelCase` is the same function.

packages/cli/generators/controller/index.js
raymondfeng: I think it should be named `toVarName` to be consistent with:
- toClassName
- toFileName

packages/cli/lib/utils.js
agnes512: But when I tested it yesterday, ```change-case.camelCase``` keeps the underscores while ```camelCase``` doesn't. And we use ```change-case.camelCase``` in ```this.artifactInfo.modelVariableName```, so I kept it and renamed it.

packages/cli/generators/controller/index.js
agnes512: true :+1::skin-tone-3:

packages/cli/lib/utils.js
raymondfeng: > I tested it yesterday, change-case.camelCase keeps the underscores while camelCase doesn't.

I think it's the other way around:
- change-case.camelCase does not use `_`
- camelCase uses `_`

agnes512: Just tested it out. I think we can get rid of the new imported camelCase since``` _.camelCase``` does the same thing. 
```
? Model class name: to1doList_todo
NAME: to1doList_todo
change-case.camel: to1doListTodo
New imported camel: to1DoListTodo
Lodash camle :to1DoListTodo
```
```
? Model class name: to_1doList_todo
NAME: to_1doList_todo
change-case.camel: to_1doListTodo
New imported camel: to1DoListTodo
Lodash camle :to1DoListTodo
```
But there still a slight difference between ```change-case.camel``` and ```_.camel```. I think we should use ```_.camel``` as we use pascal case to convert it to class name.

raymondfeng: Yes.

bajtos:
My two cents, hopefully it's not too late: personally, I prefer a model name like `Model2` to be stored in a file called `model2.ts` (not `model-2.ts`). Having said that, if this would be too difficult to implement, then I can live with `model-2.ts` too. IMO, `Model2` is not a very good name, people should not really use it in production code.

What I see as most important: the conventions should be clearly described to our users. Preferably, the CLI tools should tell the user in advance about the file name it's going to create, so that they can cancel the operation and pick a better model name early on.

For example (I don't have a working LB4 CLI, sorry that the prompts don't match the actual output):

```
? Enter model name: Model2
Will create class Model2 in src/models/model-2.ts
? What is the base class? Entity
Using base class Entity from @loopback/repository
? Allow additional properties
(etc.)
```

agnes512:
I agree that the file name for class name```Model1```  to be ```model1.ts```. But @raymondfeng suggested that it's better to apply the same convention for file naming as we might need to update lots of files.
And I think it's a good point that the CLI tools should prompt messages about the file names, class names etc. Let me try to add it to the generators:smile_cat:

raymondfeng:
> I agree that the file name for class nameModel1 to be model1.ts. But @raymondfeng suggested that it's better to apply the same convention for file naming as we might need to update lots of files.

I'm with @bajtos that `model1.ts` is better than `model-1.ts`. My suggestion actually should make it easier to implement the convention as you just have to change `utils.toFileName` method now. @agnes512 Why do you think my proposal conflicts with `model1.ts` convention?

agnes512:
I was using camelCase instead of kebabCase for the file names before. Then you asked me to change back to the kebab :joy_cat:. But it's fine. Should be easy to change to any cases now since we are using ```toFileName```.

raymondfeng:
> I was using camelCase instead of kebabCase for the file names before. Then you asked me to change back to the kebab (which would create model-1.ts):. But it's fine. Should be easy to change to any cases now since we are using toFileName.

To be precise, I think @bajtos and myself are proposing a slight variant of kebab case.

- `Model1` -> `model1`
- 'myModel` -> `my-model`
- 'your_model` -> `your-model`

agnes512:
I see. My bad. I don't know which library has such kebabCase. I will finish what we discussed first. Thanks for the suggestions.

raymondfeng:
>  I don't know which library has such kebabCase.

Let's do this:

```ts
let fileName = _.kebabCase(str);
fileName = fileName.replace(/\-(\d+)/g, '$1');
```

dhmlau:
> What I see as most important: the conventions should be clearly described to our users. Preferably, the CLI tools should tell the user in advance about the file name it's going to create, so that they can cancel the operation and pick a better model name early on.

+1 in capturing the information.  Since it's going to affect all the CLI (I think), perhaps we can have a "naming convention" section in the CLI main page? https://loopback.io/doc/en/lb4/Command-line-interface.html.  Just a thought.

packages/cli/generators/controller/index.js
```diff
@@ -75,6 +75,33 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
...
+    // inform user what controller/file names will be created
+    this.log(
+      `Controller ${chalk.yellow(
+        utils.toClassName(this.artifactInfo.name),
+      )} will be created in src/controllers/${chalk.yellow(
+        utils.toFileName(this.artifactInfo.name) + '.controller.ts',
+      )}`,
+    );
+    this.log();
+
```
raymondfeng: I'm seeing the prompts for a few generators. Let's refactor it into `ArtifactGenerator`.

packages/cli/generators/model/index.js
```diff
@@ -183,6 +183,23 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
...
+      );
+    }
+    if (this.artifactInfo.name.match(/[\u00C0-\u024F\u1E00-\u1EFF]/)) {
+      this.log(
+        chalk.red('>>> ') +
+          `Accented chars in the class name will get replaced: ${
+            this.artifactInfo.name
+          }`,
+      );
+    }
```
raymondfeng: To be refactored into `ArtifactGenerator `.

bajtos:
> Since it's going to affect all the CLI (I think), perhaps we can have a "naming convention" section in the CLI main page? https://loopback.io/doc/en/lb4/Command-line-interface.html. 

+1

In my experience, most people don't read docs, therefore I consider it more important to surface the naming convention in CLI messages.

> ```ts
> let fileName = _.kebabCase(str);
> fileName = fileName.replace(/\-(\d+)/g, '$1');
> ```

Edge case to consider: `Car4Share` - do we want `car4-share.ts`? I think `car-4-share.ts` makes more sense.

Proposed regex (notice extra `$` character):

```ts
fileName = fileName.replace(/\-(\d+)$/g, '$1');
```

Either way, please add a test case to verify the output of `toFileName('Car4Share')`.

agnes512:
> Either way, please add a test case to verify the output of `toFileName('Car4Share')`.
:ok_hand: I will modify the test case `ModelNameWithNum1` to `Model1NameWithNum1`

agnes512:
@slnode test please

PR #2992, opened 2019-05-29
feat(rest): allows validation (AJV) options to be configured via RequestBodyParserOptions

### LINKED ISSUE #2626: Support nullable properties in OpenAPI & JSON validation
When trying to perform POST request, some of the attributes are `null`.
**order.model.ts**
```
@property({
  type: 'object',
  default: null,
  nullable: true
})
specialRequests?: object;
```
**request body**
```
{
  "orderTime": "2019-01-01T00:00:00.000Z",
  "specialRequests": null
}
```
currently using mysql db with LB4

## Current Behavior
However, POST with such request body will always return validation error.

**Error: Unprocessable Entity
Response body**

```
{
  "error": {
    "statusCode": 422,
    "name": "UnprocessableEntityError",
    "message": "The request body is invalid. See error object `details` property for more info.",
    "code": "VALIDATION_FAILED",
    "details": [
      {
        "path": ".specialRequests",
        "code": "type",
        "message": "should be object",
        "info": {
          "type": "object"
        }
      }
    ]
  }
}
```

## Expected Behavior
Expected to successfully create a row in DB with captioned request body, same behavior as in LB3.

raymondfeng:
Fixes #2626 

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -8,6 +8,7 @@ import {model, property} from '@loopback/repository';
 import {
   Client,
   createRestAppClient,
+  expect,
   givenHttpServerConfig,
 } from '@loopback/testlab';
 import {
@@ -16,7 +17,9 @@ import {
   jsonToSchemaObject,
   post,
   requestBody,
+  RequestBodyValidationOptions,
   RestApplication,
+  RestBindings,
   SchemaObject,
 } from '../../..';
 import {aBodySpec} from '../../helpers';
@@ -95,6 +98,57 @@ describe('Validation at REST level', () => {
         .type('json')
         .send('null')
         .expect(400);
+    });
+  });
+
+  context('with request body validation options', () => {
+    class ProductController {
+      @post('/products')
+      async create(
+        @requestBody({required: true}) data: Product,
+      ): Promise<Product> {
+        return new Product(data);
+      }
+    }
+
+    before(() =>
+      givenAnAppAndAClient(ProductController, {
+        nullable: false,
+        compiledSchemaCache: new WeakMap(),
+      }),
+    );
+    after(() => app.stop());
+
+    it('rejects requests with `null` with {nullable: false}', async () => {
+      const DATA = {
+        name: 'iPhone',
+        description: null,
+        price: 10,
+      };
+      const res = await client
+        .post('/products')
+        .send(DATA)
+        .expect(422);
+
+      expect(res.body).to.eql({
+        error: {
+          code: 'VALIDATION_FAILED',
+          details: [
+            {
+              code: 'type',
+              info: {
+                type: 'string',
+              },
+              message: 'should be string',
+              path: '.description',
+            },
+          ],
+          message:
+            'The request body is invalid. See error object `details` property for more info.',
+          name: 'UnprocessableEntityError',
+          statusCode: 422,
+        },
+      });
     });
   });
 
@@ -245,8 +299,15 @@ describe('Validation at REST level', () => {
       .expect(422);
   }
 
-  async function givenAnAppAndAClient(controller: ControllerClass) {
+  async function givenAnAppAndAClient(
+    controller: ControllerClass,
+    validationOptions?: RequestBodyValidationOptions,
+  ) {
     app = new RestApplication({rest: givenHttpServerConfig()});
+    if (validationOptions)
+      app
+        .bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS)
+        .to({validation: validationOptions});
     app.controller(controller);
     await app.start();
 
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -119,7 +119,7 @@ describe('Validation at REST level', () => {
     );
     after(() => app.stop());
 
-    it('rejects requests with `null` with {nullable: false}', async () => {
+    it('rejects requests containing `null` with {nullable: false}', async () => {
       const DATA = {
         name: 'iPhone',
         description: null,
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -30,8 +30,10 @@ describe('Validation at REST level', () => {
     @property({required: true})
     name: string;
 
-    @property({required: false})
-    description?: string;
+    // NOTE(rfeng): We have to add `type: 'string'` to `@property` as
+    // `string | null` removes TypeScript design:type reflection
+    @property({required: false, type: 'string', jsonSchema: {nullable: true}})
+    description?: string | null;
 
     @property({required: true})
     price: number;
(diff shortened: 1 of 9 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -165,12 +165,12 @@ export function modelToJsonSchema(
     return {};
   }
 
-  const title = meta.title || ctor.name;
+  const title = meta.title || meta.name || ctor.name;
 
-  if (options.visited[title]) return options.visited[title];
+  if (options.visited.has(ctor)) return options.visited.get(ctor)!;
```
bajtos: Please note we are going to create multiple `visited` entries for a single constructor in the very near feature:
 - schema for the model as-is
 - schema for the model with `id` and `_rev` properties excluded
 - schema for the model and navigational properties for relations

See the discussion in the pull request that introduced `options.visited` for more context.

Please revert this change and keep using `title` as the key.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -29,8 +32,14 @@ describe('Validation at REST level', () => {
     @property({required: true})
     name: string;
 
-    @property({required: false})
-    description?: string;
+    // We have to add `type: 'string'` as `string | null` removes TypeScript
+    // design:type reflection
+    @property({required: false, jsonSchema: {nullable: true, type: 'string'}})
```
bajtos: I find this suspicious. Why is `type: 'string'` set at `jsonSchema` level? How is juggler and connectors going to know that the property type is string?

I would expect the following property definition:

```suggestion
    @property({required: false, type: 'string', jsonSchema: {nullable: true}})
```

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -29,8 +32,14 @@ describe('Validation at REST level', () => {
...
+    description?: string | null;
+
+    // The other option is to define `string` instead `string | null`
```
bajtos: Can `null` be assigned to `string` typed variables/properties? Will the following compile?

```ts
product.note = null;
```

If it does not compile, then I think your proposal is wrong.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -93,6 +104,57 @@ describe('Validation at REST level', () => {
...
+
+    before(() =>
+      givenAnAppAndAClient(ProductController, {
+        nullable: false,
+        skipCache: true,
+      }),
+    );
+    after(() => app.stop());
+
+    it('rejects requests nullable option is false', async () => {
```
bajtos: "(it) rejects requests nullable option is false" does not reads like an English sentence to me. Did you mean the following?

```suggestion
    it('rejects requests when nullable option is false', async () => {
```

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -4,21 +4,28 @@
...
 const debug = debugModule('loopback:rest:validation');
 
-export type RequestBodyValidationOptions = AJV.Options;
+export interface RequestBodyValidationOptions extends AJV.Options {
+  /**
+   * Skip cache for AJV validators. This flag is useful to test different
+   * AJV options as the caching key is the SchemaObject/ReferenceObject without
+   * honoring options.
+   */
+  skipCache?: boolean;
```
bajtos: What happens if the application configures two body parsers with different validation options? For example, `JSON` parser allowing nullable fields, `urlencoded-form` parser not allowing nullable fields. IIUC, the first request will populate the cache with a precompiled validator, and then all subsequent requests will use that validator regardless of parser settings.

IMO, we need to fix the caching mechanism to include (custom) validation options in the key.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
raymondfeng: I removed `note`.

packages/rest/src/validation/request-body.validator.ts
raymondfeng: We have only one `RequestBodyValidationOptions` from `RequestBodyParserOptions`. The flag is introduced mostly for testing. In production environment, the `options` is fixed based on current code.

raymondfeng:
> Please split it into two parts:
> 1. Add nullable: true to default AJV options.
> 2. Allow parsers to customize validation options.

The PR now has two commits.

bajtos:
The first commit does not have any tests, please fix that.

packages/rest/src/validation/request-body.validator.ts
bajtos: This is what the documentation says - emphasis is mine:

The request body parser options is bound to `RestBindings.REQUEST_BODY_PARSER_OPTIONS`. To customize request body parser
options, you can simply bind a new value to its key.

Built-in parsers retrieve their own options from the request body parser options. **The parser specific properties override common ones.** For example, given
the following configuration:
```ts
{
  limit: '1MB'
  json: {
    strict: false
  },
  text: {
    limit: '2MB'
  },
  // Validation options for AJV, see https://github.com/epoberezkin/ajv#options
  validation: {nullable: true},
}
```

---

As I understand that, it is possible to customize validation options on per-parser basis, and thus my original comment still applies.

PTAL.

To be honest, I find configurable AJV rules as an overkill that's not needed at this point.

> The flag is introduced mostly for testing. 

I understand that in tests, you don't want the cache to be preserved across individual tests. Let's find a better way how to accomplish that. 

Few examples:
1. We can include important options fields when computing the cache key, so that we have different cached validators for different combination of options.
2. We can expose a new internal function to clean `compiledSchemaCache`, this function will be called in tests from a `beforeEach` hook
3. We can change `compiledSchemaCache` from an internal singleton to a new application-level binding. Then in tests, whenever we create a new application instance, it will start with an empty cache. Tests sharing the same app instance can re-bind the cache.

raymondfeng: I changed the code to allow a weakmap to be passed in via `options` and it falls back to the built-in singleton.

bajtos:
The first commit a279778377b233a8b45 LGTM.

packages/rest/src/index.ts
```diff
@@ -20,6 +20,7 @@ export * from './rest.server';
 export * from './sequence';
 export * from './rest-http-error';
 export * from './parse-json';
+export * from './validation/request-body.validator';
```
bajtos: Is it a good idea to add `request-body.validator` to our public API? It will make it difficult to evolve its API in the future. 

IIUC, you are trying to access `RequestBodyValidationOptions`. Can we perhaps move definition of that interface to `src/types.ts` instead?

packages/rest/src/validation/request-body.validator.ts
bajtos: > I changed the code to allow a weakmap to be passed in via options and it falls back to the built-in singleton.

I am fine with using this approach to enable the tests 👍

bajtos: Having said that, I am not convinced that it addresses my concern about parser-specific options being ignored because of caching. Please a tests to verify, e.g. using the example I posted above for inspiration:

> For example, `JSON` parser allowing nullable fields, `urlencoded-form` parser not allowing nullable fields. IIUC, the first request will populate the cache with a precompiled validator, and then all subsequent requests will use that validator regardless of parser settings.

raymondfeng:
> I am not convinced that you have addressed my concern about parser-specific options being ignored because of caching. Please add a tests to verify, per my comment above.

1. Parser specific options are only honored by parser implementations. They don't impact the AJV validation.

2. TypeScript already prevents `validation` to be specified under `json`, `text`, `urlencoded` properties.
`{json: {validation: {}}` is not a valid `RequestBodyParserOptions`.

3. I have added more description to tsdocs and docs.

bajtos:
> I have added more description to tsdocs and docs.

This is great! Now I see that I was incorrectly assuming that `validation` is a parser-specific option.

PR #3014, opened 2019-05-31
chore: disable GreenKeeper

bajtos:
GreenKeeper does not support lerna monorepos using package-locks, see https://github.com/greenkeeperio/greenkeeper/issues/1080. Let's migrate to RenovateBot, which we are already successfully using in loopback4-example-shopping.

This pull request is removing GreenKeeper configuration, scripts and documentation. 

Before we land this pull request, we should probably remove GreenKeeper from GitHub applications, e.g. here: https://github.com/organizations/strongloop/settings/installations/633361


bajtos:
@jannyHou @nabdelgadir good catch! I search for other possible occurences of the term "greenkeeper" and founds few more places to change - see 63773cb7a

LGTY now?

bajtos:
> Before we land this pull request, we should probably remove GreenKeeper from GitHub applications, e.g. here: https://github.com/organizations/strongloop/settings/installations/633361

Done.

Issue #2953, opened 2019-05-24
[Model discovery] Generated code does not seem to be correct


dhmlau:
## Description / Steps to reproduce / Feature proposal
Using MySQL as the datasource, I created a table called Customer. Command to create the table is as below:

```
create table customer (custid VARCHAR(100), name VARCHAR(100));
ALTER table customer add primary key(custid);
```
The generated customer model looks like:
```ts
   @model([object Object])
   export class Customer extends Entity {
     @property({
       type: String,
       required: true,
       length: 100,
       id: 1,
       mysql: {"columnName":"custid","dataType":"varchar","dataLength":100,"dataPrecision":null,"dataScale":null,"nullable":"N"},
     })
     custid: String;

     @property({
       type: String,
       required: false,
       length: 100,
       mysql: {"columnName":"name","dataType":"varchar","dataLength":100,"dataPrecision":null,"dataScale":null,"nullable":"Y"},
     })
     name?: String;
     ...
   }
   ```

There are a few issues in the code:
1. Model decorator is not correct: `@model([object Object])`
2. For the string type, should it be `string` instead of `String`?
3. Inside the property decorator, should the type be `type: 'string',` instead of `type:String`

### Acceptance Criteria
Fix the generated code (@bajtos had some pointers for each of them, see https://github.com/strongloop/loopback-next/issues/2953#issuecomment-495578669): 
- [ ] Model decorator is not correct: `@model([object Object])`
- [ ] For string type, it should be `string` instead of `String`
- [ ] Inside the property decorator, it should be `type: 'string',` instead of `type:String`
- [ ] For connector specific options, it should be represented as a regular TypeScript object, i.e. for 
```
mysql: {"columnName":"name","dataType":"varchar","dataLength":100,"dataPrecision":null,"dataScale":null,"nullable":"Y"},
     })
```
it should be: 
```{ 
  @property({
    // ...
    mysql: {
      columnName: 'name',
      dataType: 'varchar',
      dataLength:100,
      dataPrecision:null,
      dataScale:null,
      nullable: 'Y'
    },
  })
  name?: String;
}```

dhmlau:
cc @marvinirwin

bajtos:
> Model decorator is not correct: @model([object Object])

Are we perhaps calling `.toString()` instead of converting the object to JSON or TypeScript? Let's use the same approach I introduced for `lb4 model` in https://github.com/strongloop/loopback-next/pull/2843.

[...]

PR #3015, opened 2019-05-31
fix: modelSettings is stringified properly #2915

### LINKED ISSUE #2915: Discover not working
Hello, I am not able to get lb4 discover to work as described for MySQL.  The 2 different results I get are as follows:

1) When using the default "root" user with no passoword and connecting to the DB I want to discover, I get back a list of "models to discover" that do not include the actual table from the DB, but a bunch of other data that looks like it might be from the performance schema or some other ancillary source (host_summary, host_summary_by_file_io....., metrics, processlist, etc)

2) When using the actual DB user/pass (tested on the command line to connect successfully), it gives me back a completely empty list of models, even though there are 18 tables.  And to be sure that it wasn't some sort of "cranking" through all the data, I tried on a version of the DB with all the data gone, but with structure intact.

Any ideas on what I am doing wrong? 

Thanks!

marvinirwin:
modelSettings is stringified ahead of time to prevent `@settings([object object])`

Fixes #2915, see also #2953

## Checklist

### DIFF
--- a/packages/cli/generators/discover/index.js
+++ b/packages/cli/generators/discover/index.js
@@ -219,9 +219,12 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
       modelDefinition.isModelBaseBuiltin = true;
       modelDefinition.modelBaseClass = 'Entity';
       modelDefinition.className = utils.pascalCase(modelDefinition.name);
-      // These last two are so that the templat doesn't error out of they aren't there
+      // These last two are so that the template doesn't error out of they aren't there
       modelDefinition.allowAdditionalProperties = true;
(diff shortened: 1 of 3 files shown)


marvinirwin:
This could have been prevented with a test that checks if a typescript file is syntactically valid.  Does anything like that exist outside of the [compiler api](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API)?

bajtos:
> This could have been prevented with a test that checks if a typescript file is syntactically valid. Is there a graceful way to do this? My first thought is the compiler api, but that's unstable.

I guess we can execute `npm run build` in the scaffolded project? It's a rather heavy solution.

bajtos:
Landed 🎉

PR #3013, opened 2019-05-31
feat(testlab): rework itSkippedOnTravis to support Jest framework

bajtos:
Describe referenced Mocha types directly in testlab to avoid compile-time dependency on Mocha types, some of which are not compatible with Jest -- fix https://github.com/strongloop/loopback-next/issues/2452

Contents of `packages/testlab/dist/skip-travis.d.ts`:

**BEFORE**

```ts
/// <reference types="mocha" />
export declare type TestCallbackRetval = void | PromiseLike<any>;
/**
 * Helper function for skipping tests on Travis env
 * @param expectation
 * @param callback
 */
export declare function itSkippedOnTravis(expectation: string, callback?: (this: Mocha.ITestCallbackContext, done: MochaDone) => TestCallbackRetval): void;
```

Notice `/// <reference types="mocha" />` which is the root cause of the problem reported in #2452

**AFTER**

```ts
export interface TestFn {
    (this: TestContext): PromiseLike<unknown>;
    (this: TestContext, done: Function): void;
}
export interface TestContext {
    skip(): this;
    timeout(ms: number | string): this;
    retries(n: number): this;
    slow(ms: number): this;
    [index: string]: any;
}
/**
 * Helper function for skipping tests on Travis env
 * @param expectation
 * @param callback
 */
export declare function itSkippedOnTravis(expectation: string, callback?: TestFn): void;
```

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3040, opened 2019-06-04
feat(testlab): add generic helper `skipOnTravis`

bajtos:
Introduce a generic helper `skipOnTravis` that supports arbitrary BDD verbs (`it`, `describe`, etc.) in a test-framework-agnostic way. Rework `itSkippedOnTravis` to leverage the new helper and deprecate it (in tsdocs only).

This is a follow-up for #3013.

## Checklist

### DIFF
--- a/packages/testlab/src/skip-travis.ts
+++ b/packages/testlab/src/skip-travis.ts
@@ -3,6 +3,47 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+/**
+ * A function defining a new test case or a test suite, e.g. `it` or `describe`.
+ */
+export type TestDefinition<ARGS extends unknown[], RETVAL> = (
+  name: string,
+  ...args: ARGS
+) => RETVAL;
+
+/**
+ * Helper function for skipping tests on Travis CI.
+ *
+ * @example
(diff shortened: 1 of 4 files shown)


packages/testlab/src/skip-travis.ts
```diff
@@ -3,6 +3,47 @@
...
+ * ```
+ *
+ * @param verb - The function to invoke to define the test case or the test
+ * suite, e.g. `it` or `describe`.
+ * @param name - The test name (the first argument of `verb` function).
+ * @param args - Additional arguments (framework specific), typically a function
+ * implementing the test.
+ */
+export function skipOnTravis<ARGS extends unknown[], RETVAL>(
+  verb: TestDefinition<ARGS, RETVAL> & {skip: TestDefinition<ARGS, RETVAL>},
```
raymondfeng: Can we make `verb` default to `it`?

bajtos: I prefer not to. Without the verb, it would be difficult to understand whether we are defining a test case (`it()`) or a test suite (`describe()`) or something else (`context()`).

```ts
skipOnTravis('returns some value when some conditions', async () => {
 // the test
});
```

Issue #2393, opened 2019-02-14
Migration guide: mount LB3 in LB4


bajtos:
_This is a follow-up for the spike #2318, see the discussion and PoC code for more context._

Create a migration guide for LB3 users showing them how to mount their existing LB3 project into a new LB4 app.

Related issues: #1849 #2389 #2391 #2392

## Acceptance criteria

- [x] Start by reorganizing the [docs for LoopBack 3.x users](https://loopback.io/doc/en/lb4/LoopBack-3.x.html): -> https://github.com/strongloop/loopback-next/pull/3052

  - Keep the current page as a top-level overview.
  - Move most of the content into a new page with a name like "Understanding the differences".

- [x] Create a new page with a name like "Migrating from v3". Describe how to use `Lb3AppBooterComponent`. Explain the differences (breaking changes) introduced by the migration. ATM, we are aware of a single difference: the spec format. LB3 produces Swagger (OpenApi v2), LB4 produces OpenApi v3. -> https://github.com/strongloop/loopback-next/pull/3052

- [x] A blog post to announce this new migration option -> https://github.com/strongloop/strongloop.com/pull/203

nabdelgadir:
Closing as done.

PR #3052, opened 2019-06-05
docs: add mounting lb3 app on lb4 migration guide

nabdelgadir:
Connect to https://github.com/strongloop/loopback-next/issues/2393.

Docs describing using `Lb3AppBooterComponent` to mount a LoopBack 3 application on a LoopBack 4 project.

## Checklist


docs/site/Migrating-from-LoopBack-3.md
```diff
@@ -0,0 +1,160 @@
...
+Explorer.
+
+## Mounting the LoopBack 3 Application in a LoopBack 4 Project
+
+{% include note.html content="
+If you are not familiar with LoopBack 4, visit
+[Getting started](Getting-started.md) for an introduction.
+" %}
+
+1. To mount your LoopBack 3 application, you will need to first create a new
```
nabdelgadir: The following part is the same as https://github.com/strongloop/loopback-next/blob/master/examples/lb3-application/README.md#tutorial, so I'm wondering if I should just link to there instead or keep it here?

docs/site/Understanding-the-differences.md
```diff
@@ -0,0 +1,106 @@
...
+keywords: LoopBack 4.0, LoopBack 4, LoopBack 3
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Understanding-the-differences.html
+---
+
+This article will help existing users understand LoopBack 4:
+
+- How to connect LoopBack 3.x concepts to LoopBack 4 terms
+- What's new and exciting in LoopBack 4
+- What's on the roadmap to achieve functional parity
```
dhmlau: i think roadmap is no longer part of this page?

docs/site/Understanding-the-differences.md
```diff
@@ -0,0 +1,106 @@
...
+title: 'Understanding the differences between LoopBack 3 and LoopBack 4'
+keywords: LoopBack 4.0, LoopBack 4, LoopBack 3
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Understanding-the-differences.html
+---
+
+This article will help existing users understand LoopBack 4:
+
+- How to connect LoopBack 3.x concepts to LoopBack 4 terms
+- What's new and exciting in LoopBack 4
```
dhmlau: While we're in this page.. I'm thinking whether we should have "what's new and exciting in LB4" (this bullet and the section at the bottom), because the list keeps changing.
I'm happy to open another issue to discuss this, and not blocking the landing of this PR.

docs/site/Migrating-from-LoopBack-3.md
jannyHou: I would suggest avoid copy paste duplicate content so that everytime we only update one place, and other places link to the source page.

docs/site/Migrating-from-LoopBack-3.md
```diff
@@ -0,0 +1,160 @@
...
+   LoopBack 4 application.
+
+   ```sh
+   $ lb4 app
```
emonddr: There are a lot of questions posed by this command, what do you suggest they specify?

docs/site/Migrating-from-LoopBack-3.md
nabdelgadir: That's a good point! I'll link then, thank you 👍

docs/site/Understanding-the-differences.md
nabdelgadir: I added [this](https://github.com/strongloop/loopback-next/pull/3052/commits/bdacc61f140369a55ac38b2f9cd1135a30abb6c7#diff-46c6d1cbe6310da7865410e1888a47b3R107) but let me know if I should just remove the whole section instead.

dhmlau: TBH, I personally prefer to have it completely. Let's keep it this way for now and we can improve the docs later. Thanks.

dhmlau:
TBH, I personally prefer to have the section "What's new and exciting in LoopBack 4" completely, but others might have other opinions.  Let's keep it this way for now and we can improve the docs later. Thanks.

PR #3056, opened 2019-06-06
[LABS] feat: adapter that wraps strategy

jannyHou:
Connect to https://github.com/strongloop/loopback-next/issues/2311

This PR contains 2 potential solutions(*you will find TWO adapter files in `src` + TWO unit tests + TWO acceptance tests*) for converting the passport based strategies to the one defined in `@loopback/authentication`:

- wrap the passport strategy, like `BasicStrategy` in [`passport-http`](https://github.com/jaredhanson/passport-http): the 1st commit
- is supposed to to wrap the passport middleware, but due to the type error I explained in https://github.com/strongloop/loopback-next/pull/3056/commits/561f6b5c1f2952c1f23cc83e996c6c845f4cfe65#diff-fb7bcb47eeb0d22961f9f511534a7db6R76, it still takes in a strategy in the constructor
    - so my code is more like a PoC that proves invoking `passport.authenticate()` works
    - if we could solve the type error, I can quickly turn it into the expected adapter
    - if the type error cannot be solved then we have to wrap the strategy not the middleware

Let's discuss which one do we want and discard the unwanted one. If both are good, I will extract one of them to a new package.

---

Update:

We decided to take the simple approach: wrapping the strategy, so please ignore the 2nd proposal above ^

## Checklist


labs/passport-adapter/src/__tests__/acceptance/authentication-with-passport-adapter.acceptance.ts
```diff
@@ -0,0 +1,201 @@
...
+      return new PassportAdapter(basicStrategy, 'basic');
+    }
+
+    configuredBasicStrategy(verifyFn: BasicVerifyFunction): BasicStrategy {
+      return new BasicStrategy(verifyFn);
+    }
+  }
+
+  function verify(username: string, password: string, cb: Function) {
+    process.nextTick(() => {
```
raymondfeng: Why do we need to wrap it into `process.nextTick()`? The `users.find` is already async.

labs/passport-adapter/src/__tests__/acceptance/authentication-with-passport-adapter.acceptance.ts
```diff
@@ -0,0 +1,201 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// FOR REVIWERS: THIS ACCEPTANCE TEST IS FOR `PassportAdapter`
+
+import { authenticate, AuthenticateFn, AuthenticationBindings, AuthenticationComponent, AuthenticationStrategy, UserProfile } from '@loopback/authentication';
```
raymondfeng: The code does not follow `prettier` formatting style.

labs/passport-adapter/README.md
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  // Applies the `StrategyAdapter` to the configured basic strategy instance.
+  // You'd better define your strategy name as a constant, like
+  // `const AUTH_STRATEGY_NAME = 'basic'`
+  // So that you will decorate the APIs later with the same name
+  convertToAuthStrategy(basic: BasicStrategy): AuthenticationStrategy {
+    return new StrategyAdapter(basic, AUTH_STRATEGY_NAME);
+  }
+}
```
raymondfeng: What about the following idea?

1. Have one `PassportAuthenticationStrategy` provider to be registered with LB4 authentication strategies extension point. 

2. The `PassportAuthenticationStrategy` supports `passport:<name>`

3. Inside `PassportAuthenticationStrategy`, we expose another extension point to register all passport strategies (for example, with a tag `passport.strategy`). 
   - Creates one instance of `passport`
   - Discovers all passport strategy extensions 
   - Calls `passport.use()` to register all passport strategies

jannyHou: @raymondfeng Having 2 extensions kind of falls back to the initial implementation in the draft PR https://github.com/strongloop/loopback-next/pull/2822 (which unfortunately was squashed and cannot be recovered...)

The reason why I wrapped the passport strategy directly to a common implementation of `AuthenticationStrategy` is explained in `https://github.com/strongloop/loopback-next/pull/2822/commits/dbc51ce8e07a6c4e30027dfa9146cd6a311e1243#r280376757

jannyHou: And I am able to perform an authentication by invoking `passport.authenticate()`, but not quite sure about the benefit of doing it (compared with invoking `strategy.authenticate()`), and describing the type of a `passport` instance is very problematic, see [explanation](https://github.com/strongloop/loopback-next/pull/3056/files#diff-fb7bcb47eeb0d22961f9f511534a7db6R76)
By saying that I didn't mean we have to define the `passport` in any function's parameter, but in case we have to do that, we have to figure out a solution for the type.

labs/passport-adapter/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2017,2019. All Rights Reserved.
```
raymondfeng: Would it be better to name the package as `authentication-passport`?

labs/passport-adapter/LICENSE
```diff
@@ -0,0 +1,25 @@
...
+Node module: @loopback/authentication
```
raymondfeng: Change the module name here.

labs/passport-adapter/README.md
```diff
@@ -0,0 +1,185 @@
...
+import {Application, CoreTags} from '@loopback/core';
+import {AuthenticationBindings} from '@loopback/authentication';
+
+app
+  .bind('authentication.strategies.basicAuthStrategy')
+  .to(basicAuthStrategy)
+  .tag({
+    [CoreTags.EXTENSION_FOR]:
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  });
```
raymondfeng: Maybe we should have a sugar method such as `usePassportStrategy()` here.

labs/passport-adapter/README.md
```diff
@@ -0,0 +1,185 @@
...
+  instance. Then in the `value()` function, you return the converted strategy.
+
+So a full implementation of the provider is:
+
+```ts
+import {BasicStrategy, BasicVerifyFunction} from 'passport-http';
+import {StrategyAdapter} from `@loopback/passport-adapter`;
+import {AuthenticationStrategy} from '@loopback/authentication';
+
+class PassportBasicAuthProvider implements Provider<AuthenticationStrategy> {
```
raymondfeng: Another option is to use `StrategyAdapter` as the base class.

labs/passport-adapter/LICENSE
bajtos: +1 to use `authentication-passport`

labs/passport-adapter/README.md
```diff
@@ -0,0 +1,185 @@
...
+  constructor(@inject(VERIFY) verifyFn: BasicVerifyFunction);
```
bajtos: Where is `VERIFY` constant defined?

labs/passport-adapter/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+addExtension(
+  app,
+  AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  PassportBasicAuthProvider,
+  {
+    namespace:
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  },
+);
```
bajtos: I find this as not very elegant. Can we find a way how to simplify this registration step? For example, can we leverage `@extension` or `@bind` decorators to specify the extension point name in the code where we are defining `PassportBasicAuthProvider`.

@raymondfeng you are the author of `addExtension` and the extension-point/extension implementation. What is the recommended way for registering extensions? Are we missing any new helpers in `@loopback/core` to simplify this process?

labs/passport-adapter/README.md
jannyHou: Yes we have it in https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/types.ts#L66 
(my code was copied from the out of date draft.)

labs/passport-adapter/README.md
jannyHou: oops...

labs/passport-adapter/README.md
jannyHou: My code is out of date we do have that sugar function :)

https://github.com/strongloop/loopback-next/pull/3056#discussion_r293112596

labs/passport-adapter/README.md
jannyHou: @raymondfeng Hmm is there a big difference between using the provider or using a class that extends `StrategyAdapter`?

labs/authentication-passport/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2017,2019. All Rights Reserved.
```
raymondfeng: It should be `2019`.

bin/mark-non-labs-packages-private.js
```diff
@@ -23,7 +23,7 @@ async function markNonLabsPackagesPrivate() {
     const dir = path.relative(pkg.rootPath, pkg.location);
     if (dir.startsWith('labs')) continue;
     const data = readJsonFile(pkg.manifestLocation);
-    let modified = !data.private;
+    const modified = !data.private;
```
raymondfeng: Can we update the base branch with this change first?

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
+# Passport Strategy Adapter
+
+_Important: We suggest users understand LoopBack's authentication system[Link
```
raymondfeng: We strongly recommend that users ...

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+TBD](some loopback.io link) before using this module_
+
+This is an adapter module created for plugging in
+[`passport`](https://www.npmjs.com/package/passport) base strategies to the
```
raymondfeng: base -> based

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+authentication system in `@loopback/authentication@2.x`.
+
+## Installation
+
+```sh
+npm i @loopback/passport-adapter --save
```
raymondfeng: `@loopback/authentication-passport`

labs/authentication-passport/README.md
jannyHou: And I updated the link as well since https://github.com/strongloop/loopback-next/pull/2977 get merged

bin/mark-non-labs-packages-private.js
jannyHou: good catch! updated in https://github.com/strongloop/loopback-next/pull/3150

jannyHou:
local tests pass

```
 loopback-next@0.1.0 mocha /Users/jannyhou/2019/auth/adapter/loopback-next
> node packages/build/bin/run-mocha "labs/*/dist/__tests__/**/*.js"



  ․․․․․․․

  7 passing (90ms)


> loopback-next@0.1.0 posttest /Users/jannyhou/2019/auth/adapter/loopback-next
> npm run lint


> loopback-next@0.1.0 lint /Users/jannyhou/2019/auth/adapter/loopback-next
> npm run prettier:check && npm run eslint && node bin/check-package-locks


> loopback-next@0.1.0 prettier:check /Users/jannyhou/2019/auth/adapter/loopback-next
> npm run prettier:cli -- -l


> loopback-next@0.1.0 prettier:cli /Users/jannyhou/2019/auth/adapter/loopback-next
> node packages/build/bin/run-prettier "labs/**/*.ts" "labs/**/*.js" "labs/**/*.md" "-l"


> loopback-next@0.1.0 eslint /Users/jannyhou/2019/auth/adapter/loopback-next
> node packages/build/bin/run-eslint --report-unused-disable-directives --cache .

```

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+Since `AuthenticationStrategy` describes a strategy with different contracts
+than the passport
+[`Strategy`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/passport/index.d.ts#L79),
+and we'd like to support the existing 500+ community passport strategies, an
+**adapter class** is created in this package to convert a passport strategy to
+the one that LoopBack 4 authentication system wants.
+
+## Usage
+
+### Simply Usage
```
raymondfeng: Simple Usage or Basic Use

emonddr:
> Created an instance of the passport strategy 

Create an instance of the passport strategy

emonddr:
> Take the basic strategy exported from passport-http as an example, first create an instance of the basic strategy with your verify function

Taking the basic strategy exported from passport-http as an example, first create an instance of the basic strategy with your verify function

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
+# Passport Strategy Adapter
+
+_Important: We suggest users understand LoopBack's authentication system[Link
+TBD](some loopback.io link) before using this module_
+
+This is an adapter module created for plugging in
+[`passport`](https://www.npmjs.com/package/passport) base strategies to the
+authentication system in `@loopback/authentication@2.x`.
+
```
emonddr: Please see two comments I added today, June 14 outside this review.

jannyHou:
For reviewers: the decreased code coverage is complaining about the `packages/*` modules, so I guess the coverage check cmd on branch `labs/base`, I am going to submit a PR to update it.

see report in https://coveralls.io/builds/24001707

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+function verify(username: string, password: string, cb: Function) {
+  users.find(username, password, cb);
+}
+const basicStrategy = new BasicStrategy(verify);
+```
+
+It's the similar configuration as you do when adding a strategy to a `passport`
+by calling `passport.use()`.
+
+But don't provide any passport specific options since we don't support the
```
emonddr: > But don't provide any passport specific options since we don't support the session manager.

Can you give an example of what you mean here? And what happens if that passport strategy requires a session manager? Will it not work?

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+```ts
+const AUTH_STRATEGY_NAME = 'basic';
+
+const basicAuthStrategy = new StrategyAdapter(
+  // The configured basic strategy instance
+  basicStrategy,
+  // Give the strategy a name
+  // You'd better define your strategy name as a constant, like
+  // `const AUTH_STRATEGY_NAME = 'basic'`.
+  // So that you will decorate the APIs later with the same name.
```
emonddr: > // So that you will decorate the APIs later with the same name. 

// You will need to decorate the APIs later with the same name

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+3. Register(bind) the strategy to app
+
+```ts
+import {Application, CoreTags} from '@loopback/core';
+import {AuthenticationBindings} from '@loopback/authentication';
+
+app
+  .bind('authentication.strategies.basicAuthStrategy')
```
emonddr: > .bind('authentication.strategies.basicAuthStrategy') 

Is there a constant that can be used from @authenticate to form the first two segments of this string?

emonddr: e.g. https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/keys.ts#L108

emonddr: I guess the class name of the strategy here 

https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/types.ts#L65 

is appended as the last segment of the string?

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+  .to(basicAuthStrategy)
+  .tag({
+    [CoreTags.EXTENSION_FOR]:
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  });
+```
+
+### With Provider
+
+If you need to inject stuff (e.g. the verify function) when configure the
```
emonddr: > If you need to inject stuff (e.g. the verify function) when configure the
strategy, you may want to provide your strategy as a provider.

If you need to inject stuff (e.g. the verify function) when configuring the
strategy, you may want to supply your strategy as a provider.

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+strategy, you may want to provide your strategy as a provider.
+
+_Note: If you are not familiar with LoopBack provider, check the documentations
```
emonddr: > Note: If you are not familiar with LoopBack provider, check the documentations
in 

Note: If you are not familiar with LoopBack providerS, check the documentation
in

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+  // Takes in the verify callback function and returns a configured basic strategy.
+  configuredBasicStrategy(verifyFn: BasicVerifyFunction): BasicStrategy {
+    return new BasicStrategy(verifyFn);
+  }
+
+  // Applies the `StrategyAdapter` to the configured basic strategy instance.
+  // You'd better define your strategy name as a constant, like
+  // `const AUTH_STRATEGY_NAME = 'basic'`
+  // So that you will decorate the APIs later with the same name
```
emonddr: > // So that you will decorate the APIs later with the same name

// You will decorate the APIs later with the same name

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+first create an instance of the basic strategy with your `verify` function.
+
+```ts
+function verify(username: string, password: string, cb: Function) {
+  users.find(username, password, cb);
+}
+const basicStrategy = new BasicStrategy(verify);
+```
+
+It's the similar configuration as you do when adding a strategy to a `passport`
```
b-admike: nitpick: `the similar` -> `a similar`

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+```ts
+const AUTH_STRATEGY_NAME = 'basic';
+
+const basicAuthStrategy = new StrategyAdapter(
+  // The configured basic strategy instance
+  basicStrategy,
+  // Give the strategy a name
+  // You'd better define your strategy name as a constant, like
+  // `const AUTH_STRATEGY_NAME = 'basic'`.
+  // So that you will decorate the APIs later with the same name.
```
b-admike: I suggest we trim down on this comment to say something like: `use an authentication strategy name to identify the strategy when @authenticate decorator is used` (maybe this is not any shorter :P)

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+  .to(basicAuthStrategy)
+  .tag({
+    [CoreTags.EXTENSION_FOR]:
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  });
+```
+
+### With Provider
+
+If you need to inject stuff (e.g. the verify function) when configure the
```
b-admike: nitpick: `when configure` -> `when configuring` and `check the documentations` -> `check the documentation`

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+import {BasicStrategy, BasicVerifyFunction} from 'passport-http';
+import {StrategyAdapter} from `@loopback/passport-adapter`;
+import {AuthenticationStrategy} from '@loopback/authentication';
+
+class PassportBasicAuthProvider implements Provider<AuthenticationStrategy> {
+  constructor(
+    @inject('authentication.basic.verify') verifyFn: BasicVerifyFunction,
+  );
+  value(): AuthenticationStrategy {
+    const basicStrategy = this.configuredBasicStrategy(verify);
```
b-admike: `this.configuredBasicStrategy(verify);` -> `this.configuredBasicStrategy(verifyFn);`

labs/authentication-passport/README.md
jannyHou: I doubled checked the usage of `passport-http`, and realize the passport specific options are only provided when the strategy is invoked by `passport.authenticate()`, like https://github.com/jaredhanson/passport-http#authenticate-requests.
So, since we don't involve `passport` in this PR, we don't introduce that concern either. I will remove ~"But don't provide any passport specific options since we don't support the session manager."~

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+  configuredBasicStrategy(verifyFn: BasicVerifyFunction): BasicStrategy {
+    return new BasicStrategy(verifyFn);
+  }
+
+  // Applies the `StrategyAdapter` to the configured basic strategy instance.
+  // You'd better define your strategy name as a constant, like
+  // `const AUTH_STRATEGY_NAME = 'basic'`
+  // So that you will decorate the APIs later with the same name
+  convertToAuthStrategy(basic: BasicStrategy): AuthenticationStrategy {
+    return new StrategyAdapter(basic, AUTH_STRATEGY_NAME);
```
b-admike: can we show where `AUTH_STRATEGY_NAME` is declared or imported?

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+const app = new MyApplication();
+
+// In a real app the function would be imported from a community module
+function verify(username: string, password: string, cb: Function) {
+  users.find(username, password, cb);
+}
+
+app.bind('authentication.basic.verify').to(verify);
+registerAuthenticationStrategy(app, PassportBasicAuthProvider);
```
b-admike: Can we include `registerAuthenticationStrategy` and `PassportBasicAuthProvider` imports at the top for this line?

labs/authentication-passport/README.md
jannyHou: @emonddr The key itself could be any name, it is the **tag** that makes the strategy recognized by the extension point, and the **`name`** property defined in the strategy class that needs to match the name in decorator `@authenticate('basic')`

I don't have a strong opinion on using a string or leverage the constant in  https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/keys.ts#L108, so I would respect your preference.

labs/authentication-passport/src/strategy-adapter.ts
```diff
@@ -0,0 +1,68 @@
...
+  authenticate(request: Request): Promise<UserProfile> {
+    return new Promise<UserProfile>((resolve, reject) => {
+      // mix-in passport additions like req.logIn and req.logOut
+      for (const key in passportRequestMixin) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (request as any)[key] = passportRequestMixin[key];
+      }
+
+      // create a prototype chain of an instance of a passport strategy
+      const strategy = Object.create(this.strategy);
```
b-admike: Maybe I missed something, but wouldn't this be `Object.create(this.passportStrategy);`?

labs/authentication-passport/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+1. Created an instance of the passport strategy
+
+Take the basic strategy exported from
+[`passport-http`](https://github.com/jaredhanson/passport-http) as an example,
+first create an instance of the basic strategy with your `verify` function.
+
+```ts
+function verify(username: string, password: string, cb: Function) {
+  users.find(username, password, cb);
```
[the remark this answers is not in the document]
jannyHou: @b-admike this is an example, we don't ship the code with this package. I think it makes sense that some developers want to know all the returned result, like a bank finds two accounts associated with a phone number, then decides how to process multiple returned result.

labs/authentication-passport/README.md
jannyHou: :P (lazy....) would prefer to stay the same

labs/authentication-passport/README.md
jannyHou: `verifyFn` is a *type*, `verify` is the real function.

labs/authentication-passport/src/strategy-adapter.ts
jannyHou: Good catch `passportStrategy` is a dup. Removed.

labs/authentication-passport/README.md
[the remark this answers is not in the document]
b-admike: Ah that is a good point. As long as they can process returned array like you said, that sounds good to me.

labs/authentication-passport/README.md
b-admike: Oh...where do we get `verify` from? I thought we are passing the injected variable here https://github.com/strongloop/loopback-next/pull/3056/files#diff-ac3170e38ad0d9561109e8d873e5e467R122. I thought `verifyFn` is the function and `BasicVerifyFunction` is the type.

labs/authentication-passport/src/__tests__/unit/passport-strategy-adapter.unit.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication-passport
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// FOR REVIWERS: THIS UNIT TEST IS FOR `StrategyAdapter`
```
b-admike: nitpick: `REVIWERS` -> `REVIEWERS`

jannyHou: oh..another good catch, that comment should be moved too!

PR #3079, opened 2019-06-10
Create a shared Repository test suite and run it against different connectors

bajtos:
This pull request creates infrastructure for writing tests for different Repositiory interfaces and executing them against different Repository implementations (e.g. `DefaultCrudRepository` and `CrudRepositoryImpl`) and different databases (e.g. `MongoDB` and `MySQL`). 

This work is based on my experience building similar test suites in the past
- KeyValue test suite: https://github.com/strongloop/loopback-datasource-juggler/blob/acb667b73d54334615bd389f81410ef27df9bd00/test/kvao.suite.js
- Operation hooks test suite: https://github.com/strongloop/loopback-datasource-juggler/blob/acb667b73d54334615bd389f81410ef27df9bd00/test/persistence-hooks.suite.js

The shared test suite has only two very basic test cases to verify the testing infrastructure. My intention is to move relevant tests from `packages/repository/src/__tests__` to the new test suite in follow-up pull requests.

## Checklist


.travis.yml
```diff
@@ -37,6 +37,16 @@ matrix:
...
+    - node_js: "10"
+      os: linux
+      env:
+        - TASK=test-repository-mongodb
+      services:
+        - mongodb
+      script:
+        - npm run postinstall -- --scope "@loopback/test-repository-mongodb" --include-filtered-dependencies --force-local
+        - npm run build -- --scope "@loopback/test-repository-mongodb" --include-filtered-dependencies
+        - cd acceptance/repository-mongodb && npm run mocha
```
bajtos: Note that I am intentionally leaving acceptance tests OUT OF `npm test` and configuring a new Travis CI job to run them in parallel. My intention is to add new scripts allowing developers to run these tests locally, but only when they decide to do so.

packages/repository-tests/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+The module provides test-suite factories to define standardized test suite
+capable of testing any combination of a Repository class and a corresponding
+connector, for example `DefaultCrudRepository` with connectors like `memory`,
+`mysql` and `mongodb`.
+
+## Installation
+
+```sh
+npm install --save @loopback/repository-testsuite
```
raymondfeng: `npm install --save @loopback/repository-tests`

raymondfeng:
@bajtos Great idea! It's nice that we can parameterize resource dependencies for tests.

bajtos:
The pull request is ready for review 💪 

@strongloop/loopback-maintainers PTAL

bajtos:
BTW if you are concerned about CI times, the new tasks finish pretty fast (and run in parallel with existing tasks).

Screen Shot 2019-06-13 at 14 48 56

acceptance/repository-mysql/README.md
```diff
@@ -0,0 +1,66 @@
...
+npm test
+```
+
+### Using Docker (Linux, MacOS, WSL)
+
+If you do not have a local MongoDB instance, you can also run the test suite
+with very minimal requirements.
+
+- Assuming you have [Docker](https://docs.docker.com/engine/installation/)
+  installed, run the following script which would spawn a MongoDB instance on
```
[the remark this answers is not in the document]
bajtos: Good catch! I was copying READMEs between packages and forgot to update all places. Thanks for catching.

bajtos:
The pull request is ready for final review. 

@raymondfeng @strongloop/loopback-maintainers PTAL

packages/repository-tests/src/helpers.repository-tests.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/repository-tests
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {CrudTestContext} from './types.repository-tests';
+
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
```
b-admike: dup copyrights statements

.travis.yml
b-admike: That's good, but these tasks still run and are required for passing builds right?

bajtos: Yes, these tasks are executed and are required to consider a CI build as success.

bajtos:
@b-admike good feedback, I have addressed your comments in 6b874b97. LGTY now?

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -0,0 +1,65 @@
...
+ * DataSource configuration (connector name, connection string, etc.).
+ */
+export type DataSourceOptions = Options;
+
+/**
+ * List of flags describing connector-specific behavior. These flags
+ * are used by the test suite to tweak assertions and skip tests
+ * for scenarios not supported by some connectors.
+ */
+export interface CrudConnectorFeatures {
```
jannyHou: Just curious :) you use `interface` instead of `type` here, is it for the convenience of using declaration merging?

packages/repository-tests/src/__tests__/acceptance/default-repository.memory.acceptance.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
```
jannyHou: (Out of the scope of this PR) Would it make more sense if move the memory acceptance test to the [acceptance dir](https://github.com/strongloop/loopback-next/tree/f1de448ec886c5dca7eb86f019d2fd87048b4565/acceptance) for the sake of consistency?

bajtos:
@jannyHou 

> You may need to add the path for the acceptance folder in the root level package.json's mocha command so that appveyor also runs the tests.

See https://github.com/strongloop/loopback-next/pull/3079#discussion_r292553020:

Note that I am intentionally leaving acceptance tests OUT OF `npm test` and configuring a new Travis CI job to run them in parallel. My intention is to add new scripts allowing developers to run these tests locally, but only when they decide to do so.

It's my intention to NOT RUN the new connector-based tests on AppVeyor.

acceptance/repository-mongodb/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "@loopback/test-repository-mongodb",
+  "version": "0.1.0",
+  "description": "",
+  "private": "true",
+  "engines": {
+    "node": ">=8.9"
+  },
+  "scripts": {
+    "build": "lb-tsc",
+    "clean": "lb-clean loopback-test-mongodb*.tgz dist tsconfig.build.tsbuildinfo package",
```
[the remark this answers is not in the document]
bajtos: Good catch! Will fix.

packages/repository-tests/src/types.repository-tests.ts
bajtos: When describing object properties, `interface` is the keyword to use. It allows other interfaces and classes to inherit from `CrudConnectorFeatures`. The keyword `type` does not allow inheritance, it defines a kind of a pointer/reference.

packages/repository-tests/src/__tests__/acceptance/default-repository.memory.acceptance.ts
bajtos: > Would it make more sense if move the memory acceptance test to the acceptance dir for the sake of consistency?

No.

I am intentionally keeping memory-based tests inside `packages/repository-tests` to ensure we execute the shared test suite as part of regular `npm test` and that the coverage data include lines covered by the shared test suite.

Here is the typical workflow:
- When making changes in loopback-next, we run the shared test suite using the memory connector, because it's relatively fast.
- We don't run the tests again databases, because it's expensive and the probability of breaking functionality in a database-specific way is very very low.
- However, we still want to verify that our code works with real databases, that's why we have dedicated Travis CI jobs running these tests for pull requests.
- In the rare case that we broke some of the shared tests with one or more databases, we can run the tests in `acceptance/repository-{database}` manually, as described in READMEs.

bajtos:
@jannyHou thank you for thoughtful comments. WDYT, should I capture some of my answers in README files or elsewhere in our docs? Please advise where & how.

packages/repository-tests/src/__tests__/acceptance/default-repository.memory.acceptance.ts
jannyHou: 👍 I see fair enough. And I like running those acceptance tests separately.

PR #3102, opened 2019-06-11
chore: remove tslint-config package

bajtos:
This package is no longer used by loopback-next (see https://github.com/strongloop/loopback-next/pull/3063) and we are not going to add or modify any rules going forward.

## Checklist


bajtos:
> Is there something that needs to be done for https://www.npmjs.com/package/@loopback/tslint-config?

That's a good question, indeed. I don't want to deprecate it, it's ok to keep using it in existing projects.

We may add a message to README, publish a new version, and only then delete the package from our repository. I am not sure if it's worth the effort?

nabdelgadir:
> We may add a message to README, publish a new version, and only then delete the package from our repository. I am not sure if it's worth the effort?

Unless there's confusion about it, I think it's fine as is for now 👍

Issue #3104, opened 2019-06-11
docs should contain which versions of LB3 packages are required to mount into LB4


derdeka:
## Description / Steps to reproduce / Feature proposal

I've tried to mount a large LB3 app into LB4. The openapi.json does not contain LB3 models after booting. I looks like LB4 is continuing processing before LB3 is ready.

## Current Behavior

Our LB3 app uses loopback-boot 3.0.1, which does not set the lb3app.booting flag.
That's why [this check](https://github.com/strongloop/loopback-next/blob/72638d28f7ac7b93a4d896b83674eac9afb6fe3e/packages/booter-lb3app/src/lb3app.booter.ts#L78) does not wait for the lb3 app to be completly booted.
I took some time to figure out, that i just need to update loopback-boot to version 3.3.0.

## Expected Behavior

The documentation should contain which LB3 packages are required at minimum. It would be helpful to print some log message if there is a incompatible version.

nabdelgadir:
The [fix](https://github.com/strongloop/loopback-boot/pull/286) that allows it to set the flag immediately was done in `loopback-boot@3.2.1`, so I'll create a PR to let users know to use that version or later. Thanks for letting us know 👍

PR #3122, opened 2019-06-12
fix: mention required loopback-boot version for mounting a lb3 app

nabdelgadir:
Resolves https://github.com/strongloop/loopback-next/issues/3104.

## Checklist

### DIFF
--- a/packages/booter-lb3app/src/lb3app.booter.ts
+++ b/packages/booter-lb3app/src/lb3app.booter.ts
@@ -75,6 +75,10 @@ export class Lb3AppBooter implements Booter {
     debug('Loading LB3 app from', this.appPath);
     const lb3App = require(this.appPath);
 
+    debug(
+      'If your LB3 app does not boot correctly then make sure it is using loopback-boot version 3.2.1 or higher.',
+    );
+
     if (lb3App.booting) {
       debug('  waiting for boot process to finish');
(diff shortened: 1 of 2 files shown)


nabdelgadir:
It works with `loopback-boot@2.x` but since it's EOL that's why I only mentioned `loopback-boot@3.x`.

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -75,6 +75,8 @@ export class Lb3AppBooter implements Booter {
     debug('Loading LB3 app from', this.appPath);
     const lb3App = require(this.appPath);
 
+    debug('Make sure loopback-boot@3.2.1 or higher is installed');
```
bajtos: As a user seeing this message, I think I would be confused. If I am already using supported loopback-boot version, then why am I being told to use it?

Are you trying to offer advice for users that may observe that application never finishes to boot because they are using an old versions of loopback-boot? If that's the case, then I am proposing a slightly different message.

```suggestion
    debug('If your LB3 app does not boot correctly then make sure it is using loopback-boot version 3.2.1 or higher.');
```

Thoughts?

nabdelgadir: Yes, I agree this is better. I'll make the change. 👍

PR #3133, opened 2019-06-12
feat: rest explorer now hosts its own copy of oas by default

### LINKED ISSUE #2285: rest-explorer doesn't work behind path-mapping proxy

mgabeler-lee-6rs:
This PR updates the `rest-explorer` to, by default, cause the main `RestServer` to expose an additional copy of the OpenAPI spec at path that is fixed relative to the explorer itself.  The explorer then gives the path to the spec as a relative one.

The goal of this is to make the explorer work reliably when behind a reverse proxy that does non-trivial URL transforms (such as placing the app under a path prefix).  Since such reverse proxy configurations are generally external to the app, it may be difficult for the app to reliably know the absolute path to itself.

Making it use a relative URL for the OpenAPI spec means that it doesn't need to care at all about path or hostname transformations a proxy might apply.

This behavior is enabled by default, but possible to disable in case some corner case I've not thought of requires the old behavior.

This should fix #2285.
This replaces #2293.  See that PR for useful background, discussion, and links to more of the same.

## Checklist

  - N/A AFAIK
  - I had to update the version in `examples/todo/package.json` to make the tests happy, not sure if this was "proper"
- [x] Verify this with a real reverse proxy configuration equivalent to that which exposed the original issue

### DIFF
--- a/packages/rest-explorer/README.md
+++ b/packages/rest-explorer/README.md
@@ -52,6 +52,77 @@ requesting a configuration option for customizing the visual style, please
 up-vote the issue and/or join the discussion if you are interested in this
 feature._
 
+### Advanced Configuration and Reverse Proxies
+
+By default, the component will add an additional OpenAPI spec endpoint, in the
+format it needs, at a fixed relative path to that of the Explorer itself. For
+example, in the default configuration, it will expose `/explorer/openapi.json`,
+or in the examples above with the Explorer path configured, it would expose
+`/openapi/ui/openapi.json`. This is to allow it to use a fixed relative path to
+load the spec, to be tolerant of running behind reverse proxies.
+
+You may turn off this behavior in the component configuration, for example:
+
+```ts
+this.configure(RestExplorerBindings.COMPONENT).to({
+  useSelfHostedSpec: false,
+});
+```
+
+If you do so, it will try to locate an existing configured OpenAPI spec endpoint
+of the required form in the REST Server configuration. This may be problematic
+when operating behind a reverse proxy that inserts a path prefix.
+
+When operating behind a reverse proxy that does path changes, such as inserting
+a prefix on the path, using the default behavior for `useSelfHostedSpec` is the
+simplest option, but is not sufficient to have a functioning Explorer. You will
+also need to explicitly configure `rest.openApiSpec.servers` (in your
+application configuration object) to have an entry that has the correct host and
+path as seen by the _client_ browser.
+
+Note that in this scenario, setting `rest.openApiSpec.setServersFromRequest` is
+not recommended, as it will cause the path information to be lost, as the
+standards for HTTP reverse proxies only provide means to tell the proxied server
+(your app) about the _hostname_ used for the original request, not the full
+original _path_.
+
+Note also that you cannot use a url-relative path for the `servers` entry, as
+the Swagger UI does not support that (yet). You may use a _host_-relative path
+however.
+
+#### Summary
+
+For some common scenarios, here are recommended configurations to have the
+explorer working properly. Note that these are not the _only_ configurations
+that will work reliably, they are just the _simplest_ ones to setup.
+
+| Scenario                                                                            | `useSelfHostedSpec` | `setServersFromRequest`                | `servers`                                                        |
+| ----------------------------------------------------------------------------------- | ------------------- | -------------------------------------- | ---------------------------------------------------------------- |
+| App exposed directly                                                                | yes                 | either                                 | automatic                                                        |
+| App behind simple reverse proxy                                                     | yes                 | yes                                    | automatic                                                        |
+| App exposed directly or behind simple proxy, with a `basePath` set                  | yes                 | yes                                    | automatic                                                        |
+| App exposed directly or behind simple proxy, mounted inside another express app     | yes                 | yes                                    | automatic                                                        |
+| App behind path-modifying reverse proxy, modifications known to app<sup>1</sup>     | yes                 | no                                     | configure manually as host-relative path, as clients will see it |
+| App behind path-modifying reverse proxy, modifications not known to app<sup>2</sup> | ?                   | ?                                      | ?                                                                |
+| App uses custom OpenAPI spec instead of LB4-generated one                           | no                  | depends on reverse-proxy configuration | depends on reverse-proxy configuration                           |
+
+<sup>1</sup> The modifications need to be known to the app at build or startup
+time so that you can manually configure the `servers` list. For example, if you
+know that your reverse proxy is going to expose the root of your app at
+`/foo/bar/`, then you would set the first of your `servers` entries to
+`/foo/bar`. This scenario also cases where the app is using a `basePath` or is
+mounted inside another express app, with this same reverse proxy setup. In those
+cases the manually configured `servers` entry will need to account for the path
+prefixes the `basePath` or express embedding adds in addition to what the
+reverse proxy does.
+
+<sup>2</sup> Due to limitations in the OpenAPI spec and what information is
+provided by the reverse proxy to the app, this is a scenario without a clear
+standards-based means of getting a working explorer. A custom solution would be
+needed in this situation, such as passing a non-standard header from your
+reverse proxy to tell the app the external path, and custom code in your app to
+make the app and explorer aware of this.
+
 ## Contributions
 
 - [Guidelines](https://github.com/strongloop/loopback-next/blob/master/docs/CONTRIBUTING.md)
--- a/packages/rest-explorer/src/rest-explorer.component.ts
+++ b/packages/rest-explorer/src/rest-explorer.component.ts
@@ -28,6 +28,13 @@ export class RestExplorerComponent implements Component {
 
     this.registerControllerRoute('get', explorerPath, 'indexRedirect');
     this.registerControllerRoute('get', explorerPath + '/', 'index');
+    if (restExplorerConfig.useSelfHostedSpec !== false) {
+      this.registerControllerRoute(
+        'get',
+        explorerPath + '/openapi.json',
+        'spec',
+      );
+    }
 
     application.static(explorerPath, swaggerUI.getAbsoluteFSPath());
 
--- a/packages/rest-explorer/src/rest-explorer.controller.ts
+++ b/packages/rest-explorer/src/rest-explorer.controller.ts
@@ -6,14 +6,16 @@
 import {inject} from '@loopback/context';
 import {
   OpenApiSpecForm,
-  Request,
-  Response,
+  RequestContext,
   RestBindings,
+  RestServer,
   RestServerConfig,
 } from '@loopback/rest';
 import * as ejs from 'ejs';
 import * as fs from 'fs';
 import * as path from 'path';
+import {RestExplorerBindings} from './rest-explorer.keys';
+import {RestExplorerConfig} from './rest-explorer.types';
 
 // TODO(bajtos) Allow users to customize the template
 const indexHtml = path.resolve(__dirname, '../templates/index.html.ejs');
@@ -21,52 +23,80 @@ const template = fs.readFileSync(indexHtml, 'utf-8');
 const templateFn = ejs.compile(template);
 
 export class ExplorerController {
+  static readonly OPENAPI_RELATIVE_URL = 'openapi.json';
+  static readonly OPENAPI_FORM: OpenApiSpecForm = Object.freeze({
+    version: '3.0.0',
+    format: 'json',
+  });
+
   private openApiSpecUrl: string;
+  private useSelfHostedSpec: boolean;
 
   constructor(
     @inject(RestBindings.CONFIG, {optional: true})
     restConfig: RestServerConfig = {},
+    @inject(RestExplorerBindings.CONFIG, {optional: true})
+    explorerConfig: RestExplorerConfig = {},
     @inject(RestBindings.BASE_PATH) private serverBasePath: string,
-    @inject(RestBindings.Http.REQUEST) private request: Request,
-    @inject(RestBindings.Http.RESPONSE) private response: Response,
+    @inject(RestBindings.SERVER) private restServer: RestServer,
+    @inject(RestBindings.Http.CONTEXT) private requestContext: RequestContext,
   ) {
+    this.useSelfHostedSpec = explorerConfig.useSelfHostedSpec !== false;
     this.openApiSpecUrl = this.getOpenApiSpecUrl(restConfig);
   }
 
   indexRedirect() {
-    const url = this.request.originalUrl || this.request.url;
-    this.response.redirect(301, url + '/');
+    const {request, response} = this.requestContext;
+    let url = request.originalUrl || request.url;
+    // be safe against path-modifying reverse proxies by generating the redirect
+    // as a _relative_ URL
+    const lastSlash = url.lastIndexOf('/');
+    if (lastSlash >= 0) {
+      url = './' + url.substr(lastSlash + 1) + '/';
+    }
+    response.redirect(301, url);
   }
 
   index() {
     let openApiSpecUrl = this.openApiSpecUrl;
 
-    // baseURL is composed from mountPath and basePath
-    // OpenAPI endpoints ignore basePath but do honor mountPath
-    let rootPath = this.request.baseUrl;
-    if (
-      this.serverBasePath &&
-      this.serverBasePath !== '/' &&
-      rootPath.endsWith(this.serverBasePath)
-    ) {
-      rootPath = rootPath.slice(0, -this.serverBasePath.length);
-    }
+    // if using self-hosted openapi spec, then the path to use is always the
+    // exact relative path, and no base path logic needs to be applied
+    if (!this.useSelfHostedSpec) {
+      // baseURL is composed from mountPath and basePath
+      // OpenAPI endpoints ignore basePath but do honor mountPath
+      let rootPath = this.requestContext.request.baseUrl;
+      if (
+        this.serverBasePath &&
+        this.serverBasePath !== '/' &&
+        rootPath.endsWith(this.serverBasePath)
+      ) {
+        rootPath = rootPath.slice(0, -this.serverBasePath.length);
+      }
 
-    if (rootPath && rootPath !== '/') {
-      openApiSpecUrl = rootPath + openApiSpecUrl;
+      if (rootPath && rootPath !== '/') {
+        openApiSpecUrl = rootPath + openApiSpecUrl;
+      }
     }
     const data = {
       openApiSpecUrl,
     };
 
     const homePage = templateFn(data);
-    this.response
+    this.requestContext.response
       .status(200)
       .contentType('text/html')
       .send(homePage);
   }
 
+  spec() {
+    return this.restServer.getApiSpec(this.requestContext);
+  }
+
   private getOpenApiSpecUrl(restConfig: RestServerConfig): string {
+    if (this.useSelfHostedSpec) {
+      return './' + ExplorerController.OPENAPI_RELATIVE_URL;
+    }
     const openApiConfig = restConfig.openApiSpec || {};
     const endpointMapping = openApiConfig.endpointMapping || {};
     const endpoint = Object.keys(endpointMapping).find(k =>
@@ -77,5 +107,8 @@ export class ExplorerController {
 }
 
 function isOpenApiV3Json(mapping: OpenApiSpecForm) {
-  return mapping.version === '3.0.0' && mapping.format === 'json';
+  return (
+    mapping.version === ExplorerController.OPENAPI_FORM.version &&
+    mapping.format === ExplorerController.OPENAPI_FORM.format
+  );
 }
--- a/packages/rest-explorer/src/rest-explorer.types.ts
+++ b/packages/rest-explorer/src/rest-explorer.types.ts
@@ -11,4 +11,22 @@ export type RestExplorerConfig = {
    * URL path where to expose the explorer UI. Default: '/explorer'
    */
   path?: string;
+
+  /**
+   * By default, the explorer will add an additional copy of the OpenAPI spec
+   * in v3/JSON format at a fixed url relative to the explorer itself. This
+   * simplifies making the explorer work in environments where there may be
+   * e.g. non-trivial URL rewriting done by a reverse proxy, at the expense
+   * of adding an additional endpoint to the application. You may shut off
+   * this behavior by setting this flag `false`, in which case the explorer
+   * will try to locate an OpenAPI endpoint from the RestServer that is
+   * already in the correct form.
+   *
+   * Note that, if you are behind such a reverse proxy, you still _must_
+   * explicitly set an `openApiSpecOptions.servers` entry with an absolute path
+   * (it does not need to include the protocol, host, and port) that reflects
+   * the externally visible path, as that information is not systematically
+   * forwarded to the application behind the proxy.
+   */
+  useSelfHostedSpec?: false;
 };
--- a/packages/rest/src/__tests__/integration/rest.server.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.server.integration.ts
@@ -506,6 +506,47 @@ paths:
     await test.get('/explorer').expect(404);
   });
 
+  it('can add openApiSpec endpoints before express initialization', async () => {
+    const server = await givenAServer();
+    server.addOpenApiSpecEndpoint('/custom-openapi.json', {
+      version: '3.0.0',
+      format: 'json',
+    });
+
+    const test = createClientForHandler(server.requestHandler);
+    await test.get('/custom-openapi.json').expect(200);
+  });
+
+  // this doesn't work: once the generic routes have been added to express to
+  // direct requests at controllers, adding OpenAPI spec routes after that
+  // no longer works in the sense that express won't ever try those routes
+  // https://github.com/strongloop/loopback-next/issues/433 will make changes
+  // that make it possible to enable this test
+  it.skip('can add openApiSpec endpoints after express initialization', async () => {
+    const server = await givenAServer();
+    const test = createClientForHandler(server.requestHandler);
+    server.addOpenApiSpecEndpoint('/custom-openapi.json', {
+      version: '3.0.0',
+      format: 'json',
+    });
+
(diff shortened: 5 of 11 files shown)


packages/rest-explorer/src/rest-explorer.types.ts
```diff
@@ -11,4 +11,16 @@ export type RestExplorerConfig = {
    * URL path where to expose the explorer UI. Default: '/explorer'
    */
   path?: string;
+
+  /**
+   * By default, the explorer will add an additional copy of the OpenAPI spec
+   * in v3/JSON format at a fixed url relative to the explorer itself.  This
```
bajtos:
```suggestion
   * in v3/JSON format at a fixed url relative to the explorer itself. This
```

packages/rest-explorer/src/rest-explorer.types.ts
```diff
@@ -11,4 +11,16 @@ export type RestExplorerConfig = {
...
+   * simplifies making the explorer work in environments where there may be
+   * e.g. non-trivial URL rewriting done by a reverse proxy, at the expense
+   * of adding an additional endpoint to the application.  You may shut off
```
bajtos:
```suggestion
   * of adding an additional endpoint to the application. You may shut off
```

packages/rest/src/rest.server.ts
```diff
@@ -260,19 +260,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
    */
   protected _setupOpenApiSpecEndpoints() {
     if (this.config.openApiSpec.disabled) return;
-    // NOTE(bajtos) Regular routes are handled through Sequence.
-    // IMO, this built-in endpoint should not run through a Sequence,
-    // because it's not part of the application API itself.
-    // E.g. if the app implements access/audit logs, I don't want
-    // this endpoint to trigger a log entry. If the server implements
-    // content-negotiation to support XML clients, I don't want the OpenAPI
-    // spec to be converted into an XML response.
```
bajtos: Please preserve this comment, I believe it's still relevant.

mgabeler-lee-6rs: I did (and will with further refactors), I just moved it to stick with the code that does the actual express binding -- see line 285 in the same file.

The suggested refactors for changing the spec hosting should result in it moving back here.

packages/rest-explorer/src/rest-explorer.types.ts
mgabeler-lee-6rs: Old typing habits die hard ;)

Will fix.

mgabeler-lee-6rs:
I've updated according to the review notes, and also have done some testing with an actual reverse proxy config, which found two things related to finding the proper server URL for the explorer to talk to when behind a path-modifying proxy:
1. Using `setServersFromRequest` breaks things
2. Still need to manually set `servers[].url` to have the external base path.

I'm working on an update to try to address these by (conditionally) inserting an entry in `servers` that uses a computed relative path.

mgabeler-lee-6rs:
Unfortunately it seems `swagger-ui` / OAS only support "relative" server paths in the sense of them being relative to the server, not relative to the spec path.  And there is no standard I can find for reverse proxies passing the original request _path_ in some `X-Forwarded-*` header.

mgabeler-lee-6rs:
One hack that I considered and explicitly rejected: When loading the spec from the explorer, it is (usually) possible to infer the external URL in use from the `Referrer` header.

I _could_ add logic to infer a `servers` entry from the `Referrer` when present, but that seems brittle and I wouldn't go down that path unless you really think it's worth it.  It would produce some quite strange behavior, where the spec the explorer gets is not the same as if you manually load the same spec URL the explorer uses.  Not to mention the issues of trying to accurately detect if the referrer even _is_ the self-hosted explorer.

mgabeler-lee-6rs:
I gave a whirl at writing the referer thing to see how ugly it is ... here's what I came up with.  I'm not at all sure that it correctly handles all the possible cases of base/mount paths however, and it still seems ... brittle.

```
diff --git a/packages/rest-explorer/src/rest-explorer.controller.ts b/packages/rest-explorer/src/rest-explorer.controller.ts
index 5f6c0b59..b2d1953b 100644
--- a/packages/rest-explorer/src/rest-explorer.controller.ts
+++ b/packages/rest-explorer/src/rest-explorer.controller.ts
@@ -33,6 +33,7 @@ export class ExplorerController {
 
   private openApiSpecUrl: string;
   private useOwnOpenApiSpecEndpoint: boolean;
+  private explorerPath: string;
 
   constructor(
     @inject(RestBindings.CONFIG, {optional: true})
@@ -47,6 +48,10 @@ export class ExplorerController {
   ) {
     this.useOwnOpenApiSpecEndpoint =
       explorerConfig.useOwnOpenApiSpecEndpoint !== false;
+    this.explorerPath = explorerConfig.path || '/explorer';
+    if (!this.explorerPath.endsWith('/')) {
+      this.explorerPath += '/';
+    }
     this.openApiSpecUrl = this.getOpenApiSpecUrl(restConfig);
   }
 
@@ -89,6 +94,13 @@ export class ExplorerController {
 
   spec() {
     const specObj = this.restServer.getApiSpec(this.requestContext);
+    const referer = this.request.headers.referer;
+    if (this.useOwnOpenApiSpecEndpoint && referer && specObj.servers) {
+      // evil: infer the external server url from the referer header, insert it in the servers list
+      if (referer.endsWith(this.explorerPath)) {
+        specObj.servers.unshift({url: referer.slice(0, -this.explorerPath.length + 1)});
+      }
+    }
     const spec = JSON.stringify(specObj, null, 2);
     this.response.setHeader('content-type', 'application/json; charset=utf-8');
     this.response.end(spec, 'utf-8');
```

bajtos:
Thank you for a thorough investigation, @mgabeler-lee-6rs. I really appreciate the effort you are putting into this work, especially considering all different dead-ends we have encountered so far.

> Unfortunately it seems `swagger-ui` / OAS only support "relative" server paths in the sense of them being relative to the server, not relative to the spec path.  

This looks like a bug in swagger-ui to me. Quoting from https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#server-object, emphasis is mine:

_A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that **the host location is relative to the location where the OpenAPI document is being served**._

Would you mind reporting an issue to swagger-ui and/or swagger-js?

Maybe I am reading the spec incorrectly and the term "host location" is referring only to the scheme+hostname+port part of the URL? Let's find out :)

> And there is no standard I can find for reverse proxies passing the original request _path_ in some `X-Forwarded-*` header.

Hmm.

> I could add logic to infer a servers entry from the Referrer when present, but that seems brittle and I wouldn't go down that path unless you really think it's worth it.

I agree it's rather brittle. I guess I can live with it if we don't find any better solution.

---

It makes me wonder. As the developer setting up the LB app & the reverse proxy, do you know the URL where the OpenAPI spec can be loaded from? What if we added a new configuration option allowing developers to tell the Explorer where to load the spec from?

At the moment, Explorer accepts a single option - the path where to server the front-end:

https://github.com/strongloop/loopback-next/blob/ba3f3894ef7e102b0fcd8b9b7a1f1d221bbbf4a4/packages/rest-explorer/src/rest-explorer.types.ts#L9-L14

What if we added a new option, e.g. `specUrl` that will the explorer use, instead of trying to infer the spec URL from the request & server config?

As for the second pard (using `setServersFromRequest` breaks things + still need to manually set `servers[].url` to have the external base path), is this actually a problem? 

My idea is that if the app is running behind a reverse proxy, then the developer should:
- configure REST API Explorer with the URL where clients can load the spec from
- provide `servers[].url` entry with the URL where clients can access the API

Obviously, that won't work well if there are some clients accessing the server through a reverse proxy and other clients that are accessing the server directly. Is this a relevant use case for you?

mgabeler-lee-6rs:
> Thank you for a thorough investigation, @mgabeler-lee-6rs. I really appreciate the effort you are putting into this work, especially considering all different dead-ends we have encountered so far.

You're welcome.  Enlightened self-interest and all that :smiley:

> > Unfortunately it seems `swagger-ui` / OAS only support "relative" server paths in the sense of them being relative to the server, not relative to the spec path.
> 
> This looks like a bug in swagger-ui to me. Quoting from https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#server-object, emphasis is mine:
> 
> _A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that **the host location is relative to the location where the OpenAPI document is being served**._

Interesting, I was reading https://swagger.io/docs/specification/api-host-and-base-path/#relative-urls which explicitly says "The URLs in the servers array can be relative, such as /v2. In this case, **the URL is resolved against the server that hosts the given OpenAPI definition**".  Clearly the spec is the authoritative source of proper behavior.

> Would you mind reporting an issue to swagger-ui and/or swagger-js?

Filed https://github.com/swagger-api/swagger-ui/issues/5401

> It makes me wonder. As the developer setting up the LB app & the reverse proxy, do you know the URL where the OpenAPI spec can be loaded from? What if we added a new configuration option allowing developers to tell the Explorer where to load the spec from?

That's basically what this PR addresses -- making the explorer handle that for itself by having a copy of the spec available to it at a constant relative path.

> At the moment, Explorer accepts a single option - the path where to server the front-end:
> 
> https://github.com/strongloop/loopback-next/blob/ba3f3894ef7e102b0fcd8b9b7a1f1d221bbbf4a4/packages/rest-explorer/src/rest-explorer.types.ts#L9-L14
> 
> What if we added a new option, e.g. `specUrl` that will the explorer use, instead of trying to infer the spec URL from the request & server config?

Well, that's the issue "self-hosting the spec" in this PR resolves -- it allows the explorer to always use a fixed URL to get the spec, and not care for that aspect about any base URLs that are involved, whether they are within loopback or from an external proxy.

> As for the second part (using `setServersFromRequest` breaks things + still need to manually set `servers[].url` to have the external base path), is this actually a problem?

Not really, no.  A minor annoyance only.

> My idea is that if the app is running behind a reverse proxy, then the developer should:
> 
> * configure REST API Explorer with the URL where clients can load the spec from

The relative url hosting of the spec in this PR will allow that to be handled automagically.

> * provide `servers[].url` entry with the URL where clients can access the API

Yup :+1:

> Obviously, that won't work well if there are some clients accessing the server through a reverse proxy and other clients that are accessing the server directly. Is this a relevant use case for you?

Our use case requires supporting both direct & proxied access, but _not at the same time_ -- developers run with direct access, test/integration & production run behind the proxy.  We use the common pattern of a `NODE_ENV` environment variable to tell the app which scenario it's in.

So, for us, having to set `servers[].url` manually when running in the proxied configuration is acceptable.  We already have a shared utility function to do that as part of building the `config` object for the app at launch, and it does not require monkey patching loopback classes.

Fixing the relative path handling in `swagger-ui` would allow us to simplify that helper and remove the environment-conditional for it -- it would simply always be `../` for us.

----

TL;DR: The relative spec hosting in this PR is sufficient to address our use case.

----

Will take a swing through and work on bringing up the coverage, and also will take a pass at adding a section to the documentation on how to configure this stuff for path-prefixing reverse proxies like we have.

bajtos:
@mgabeler-lee-6rs what's the status of this pull request? Is there anything I can help you with to move this work forward?

mgabeler-lee-6rs:
I've got local work to sort out the test coverage, but the test are currently broken in my merge from `master` -- working on sorting that out, I'm sure it's something relatively simple.

mgabeler-lee-6rs:
OK, I think I have this in decent shape ... notes/thoughts:
* The last commit (editing `version` in `examples/todo/package.json` was a bit of hand waving to make the tests pass.  I'm not sure it was the proper way to get that done?
* I had to rebase & force-push this in order to make the Travis commitlint check happy due to the limited git history it fetches.  That may have messed up some of the older review notes in GitHub
* Let me know if you'd like me to squash the commit history for this PR to clean it up
* On the top level TODO item about verifying this with a real nginx reverse-proxy config: I did actually do that with an earlier version of this PR, just not with the latest updates to it, esp. after the rebase.  I have all the setup however so it will be quick & easy to repeat tomorrow.

Here's my nginx config snippet I used for that testing:

```
map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}

server {
    listen       33000;
    server_name  localhost;

    location = /lb4test {
        return 302 /lb4test/;
    }

    location ^~ /lb4test/ {
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
        proxy_http_version 1.1;
        chunked_transfer_encoding off;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_pass http://172.17.0.1:3000/;
        proxy_send_timeout 10s;
        proxy_read_timeout 10s;
    }
}
```

packages/rest-explorer/src/rest-explorer.controller.ts
```diff
@@ -66,7 +87,17 @@ export class ExplorerController {
       .send(homePage);
   }
 
+  spec() {
+    const specObj = this.restServer.getApiSpec(this.requestContext);
+    const spec = JSON.stringify(specObj, null, 2);
+    this.response.setHeader('content-type', 'application/json; charset=utf-8');
+    this.response.end(spec, 'utf-8');
```
bajtos: Is it necessary for this endpoint to deal with JSON serialization to HTTP response? I believe that by default, LoopBack is automatically serializing controller return values as JSON. I think this controller method can be simplified as follows:

```ts
spec() {
  return this.restServer.getApiSpec(this.requestContext);
}
```

mgabeler-lee-6rs: You may well be right.  This code is copy-pasta from `RestServer._serveOpenApiSpec`.

mgabeler-lee-6rs: Doing some git history spelunking, it looks like this code comes from back when it was done at the express level, instead of being a controller level (3286bc65).

A quick local test indicates that your simplification does work, so I'll clean up both copies of this code :+1:

mgabeler-lee-6rs: One thing I notice happens when I do this: the response is no longer _explicitly_ listed (in the HTTP headers) as being in the UTF-8 encoding (no specific encoding is listed).

I suspect this is largely a non-issue, but let me know if you disagree.

packages/rest-explorer/README.md
```diff
@@ -52,6 +52,44 @@ requesting a configuration option for customizing the visual style, please
...
+Note that in this scenario, setting `rest.openApiSpec.setServersFromRequest` is
+not recommended, as it will cause the path information to be lost, as the
+standards for HTTP reverse proxies only provide means to tell the proxied server
+(your app) about the _hostname_ used for the original request, not the full
+original _path_.
+
+Note also that you cannot use a url-relative path for the `servers` entry, as
+the Swagger UI does not support that (yet). You may use a _host_-relative path
+however.
+
```
raymondfeng: To make it easy to understand, maybe we should build a table to show different cases with corresponding configurations, such as:

|Use case|Recommended configuration|Exposed spec/UI URL|
|---------|------------------------------|----------------------|
|Running as standalone server|||
|Running behind a reverse proxy|||

mgabeler-lee-6rs: :+1: will work on that

I _think_ the scenarios to consider are:
* Whether the app is exposed directly or not (i.e. does the request the app received have the original request host directly or in a reverse proxy provided header)
* Whether there is a path-modifying reverse proxy in front of the app
* Whether you have advanced an (undetermined) advanced use case where you need the explorer to use a custom OpenAPI spec instead of the LB4 generated one

mgabeler-lee-6rs:
Note: The commit-lint check in Travis is angry because it only clones a limited history depth ... which is causing the commitlint to not see the branch point for this PR, and thus it's running commit lint against a giant pile of commits that aren't actually part of this PR.  I can rebase this PR (again) if cleaning that up is important.

raymondfeng:
@mgabeler-lee-6rs We use `rebase` to pick up upstream changes into the PR. See https://loopback.io/doc/en/lb4/submitting_a_pr.html#6-rebasing

packages/rest-explorer/README.md
```diff
@@ -52,6 +52,67 @@ requesting a configuration option for customizing the visual style, please
...
 feature._
 
+### Advanced Configuration and Reverse Proxies
+
+By default, the component will add an additional OpenAPI spec endpoint, in the
+format it needs, at a fixed relative path to that of the Explorer itself. For
+example, in the default configuration, it will expose `/explorer/openapi.json`,
+or in the examples above with the Explorer path configured, it would expose
+`/openapi/ui/openapi.json`. This is to allow it to use a fixed relative path to
+load the spec, to be tolerant of running behind reverse proxies.
```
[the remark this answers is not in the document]
mgabeler-lee-6rs: No, it leaves those un-touched

packages/rest-explorer/src/rest-explorer.types.ts
```diff
@@ -11,4 +11,22 @@ export type RestExplorerConfig = {
...
+   * will try to locate an OpenAPI endpoint from the RestServer that is
+   * already in the correct form.
+   *
+   * Note that, if you are behind such a reverse proxy, you still _must_
+   * explicitly set an `openApiSpecOptions.servers` entry with an absolute path
+   * (it does not need to include the protocol, host, and port) that reflects
+   * the externally visible path, as that information is not systematically
+   * forwarded to the application behind the proxy.
+   */
+  useOwnOpenApiSpecEndpoint?: false;
```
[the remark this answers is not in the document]
mgabeler-lee-6rs: `relativeOpenApiSpecEndpoint` sounds good to me, I'll make that change

packages/rest-explorer/src/rest-explorer.controller.ts
bajtos: > One thing I notice happens when I do this: the response is no longer explicitly listed (in the HTTP headers) as being in the UTF-8 encoding (no specific encoding is listed).

That's because of the following line:

https://github.com/strongloop/loopback-next/blob/c944654881140119e44fe9e1db85a0f26e59d658/packages/rest/src/writer.ts#L52

I think the missing charset information is fine as far as this pull request is concerned.

It would be nice to eventually fix the line above to include charset/encoding too - would you like to open a new pull request?

packages/rest/src/rest.server.ts
```diff
@@ -688,6 +710,24 @@ export class RestServer extends Context implements Server, HttpServerLike {
     }
 
     assignRouterSpec(spec, this._externalRoutes.routerSpec);
+
+    if (requestContext) {
```
bajtos: Let's use reverse logic & early return to make the code easier to reason about and keep the level of indentation low.

```ts
if (!requestContext) {
  return spec;
}

if (this.config.openApiSpec.setServersFromRequest)
// ...
return spec;
```

Alternatively, and I think that's a better solution, can you extract this code into a helper function?

```ts
if (requestContext) {
  spec = updateSpecFromRequest(spec, requestContext);
}
return spec;
```

packages/rest-explorer/src/rest-explorer.controller.ts
```diff
@@ -10,26 +10,43 @@ import {
...
   constructor(
     @inject(RestBindings.CONFIG, {optional: true})
     restConfig: RestServerConfig = {},
+    @inject(RestExplorerBindings.CONFIG, {optional: true})
+    explorerConfig: RestExplorerConfig = {},
     @inject(RestBindings.BASE_PATH) private serverBasePath: string,
+    @inject(RestBindings.SERVER) private restServer: RestServer,
     @inject(RestBindings.Http.REQUEST) private request: Request,
+    @inject(RestBindings.Http.CONTEXT) private requestContext: RequestContext,
     @inject(RestBindings.Http.RESPONSE) private response: Response,
```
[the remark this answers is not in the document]
mgabeler-lee-6rs: Fixing

mgabeler-lee-6rs:
Quoting a few since a prior rebase means GitHub won't let me reply to some review comments directly...

> This is a known problem of the current implementation, see #433

Fixing

> Alternatively, and I think that's a better solution, can you extract this code into a helper function.

Fixing

> I'd like to add to more well-known cases to the list

I'm less familiar with these use cases, will read up a bit and update the PR with info in a followup commit soon

raymondfeng:
@mgabeler-lee-6rs Thank you for the update. But CI is failing - https://travis-ci.com/strongloop/loopback-next/jobs/236909000. Please investigate.

mgabeler-lee-6rs:
Yup, working on it -- the bug I fixed related to having the explorer work in some of the base path related scenarios broke some test expectations, fix will be easy

examples/todo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@loopback/example-todo",
-  "version": "1.7.4",
+  "version": "1.8.0",
```
bajtos: We are bumping up version numbers during the release process, please revert this line.

packages/rest-explorer/src/rest-explorer.controller.ts
```diff
@@ -41,18 +39,24 @@ export class ExplorerController {
...
     this.relativeOpenApiSpecEndpoint =
       explorerConfig.relativeOpenApiSpecEndpoint !== false;
     this.openApiSpecUrl = this.getOpenApiSpecUrl(restConfig);
   }
 
   indexRedirect() {
-    const url = this.request.originalUrl || this.request.url;
-    this.response.redirect(301, url + '/');
+    let url =
+      this.requestContext.request.originalUrl ||
```
bajtos: Repeating `this.requestContext.` prefix is tedious, have you considered the following refactoring?

```ts
const {request, response} = requestContext;

let url = request.originalUrl || request.url;
// etc.
```

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -520,6 +520,8 @@ paths:
   // this doesn't work: once the generic routes have been added to express to
   // direct requests at controllers, adding OpenAPI spec routes after that
   // no longer works in the sense that express won't ever try those routes
+  // https://github.com/strongloop/loopback-next/issues/433 will make changes
+  // that make it possible to enable this test
```
bajtos: 👍

packages/rest-explorer/src/rest-explorer.component.ts
```diff
@@ -28,6 +28,13 @@ export class RestExplorerComponent implements Component {
 
     this.registerControllerRoute('get', explorerPath, 'indexRedirect');
     this.registerControllerRoute('get', explorerPath + '/', 'index');
+    if (restExplorerConfig.relativeOpenApiSpecEndpoint !== false) {
```
bajtos: I have mixed feelings about the option name `relativeOpenApiSpecEndpoint`, I feel it's describing implementation details instead of the intent. How about `useSelfHostedSpec` or `selfHostSpec`?

examples/todo/package.json
mgabeler-lee-6rs: Will do -- but note that it will cause CI to fail

packages/rest-explorer/src/rest-explorer.component.ts
mgabeler-lee-6rs: Not sure why I didn't think of this before, `useSelfHostedSpec` is _much_ clearer & simpler :grinning:

examples/todo/package.json
bajtos: That's weird and not expected at all. Can you try to rebase your PR on top of the latest master please?

mgabeler-lee-6rs: Oh, oops ... forgot to switch git reflexes to fork mode and thought I was already up to date there... I suspect that will do the trick

mgabeler-lee-6rs:
Rebased & squashed

Thanks for the patience & feedback!

raymondfeng:
@mgabeler-lee-6rs PR landed 🥇 Thank you for the great contribution!

bajtos:
Thank you @mgabeler-lee-6rs for your perseverance ❤️ It's great to see this work finished and landed 🎉

Issue #3159, opened 2019-06-14
Spike: migrate tests to Jest (ts-jest)


bajtos:
Let's try to speed up our dev process and CI builds by migrating to ts-jest.

1. Jest can run tests in parallel and provides clever watch mode.
2. ts-jest compiles TypeScript files on the fly, removing the need for costly build.
3. Jest provides tools for snapshot-based testing, this would be very useful to test CLI code scaffolding TypeScript files and to test code emitting OpenAPI spec.

Known problems to resolve:

- [ ] `@loopback/metadata` relies on `instanceof Function` check that does not work under Jest, see #3160
- [ ] `@loopback/boot` is not able to find & load artifacts from original `.ts` files
- [ ] Tests using classes defined outside of the scope of test functions fail with ReferenceError, e.g. `ReferenceError: Cannot access 'CustomerController' before initialization`.
- [ ] Integration with express is broken
- [ ] Repository bindings are not picked up (`The key 'repositories.NoteRepository' is not bound to any value in context application`)

See the list of skipped files in https://github.com/strongloop/loopback-next/blob/test/jest/jest.config.js for the up-to-date information.

bajtos:
Work started in https://github.com/strongloop/loopback-next/tree/test/jest

mamiller93:
I'm curious whether this spike has an updated timeline as you removed the 2020Q1 tag about a month ago. From my research, a lot of users have expressed mocha is faster than jest, but we're also running into some slowness when running 300+ tests over 50+ files. I'm curious whether the migration to Jest would prove beneficial and whether we should embark on that adventure by ourselves or hold off and wait for the official loopback merge.

bajtos:
@mamiller93 this spike is not a priority for us in the next few months (at least). The migration would require too many changes.

We are also discussing with @boneskull the option to improve Mocha to better serve our needs: allow the tests to be run in parallel, better support for TypeScript in watch mode.

> I'm curious whether the migration to Jest would prove beneficial and whether we should embark on that adventure by ourselves or hold off and wait for the official loopback merge.

I guess the only way how to answer the first question is to do a spike on your side and see if you can get any speed up by moving to Jest.

Personally, I find Jest much better to use when working in test-driven way, as it has CLI UI allowing me to re-run only a subset of tests (e.g. only the tests that failed in the last run), update snapshots, etc. Having said that, I have only minimal experience with Jest, it's very likely I am not aware of many problems that we would discover if we migrated loopback-next to Jest.

[...]

PR #3160, opened 2019-06-14
fix(metadata): don't rely on `instanceof` checks

bajtos:
The testing framework [Jest](https://jestjs.io) creates a new V8 VM for each test file and as a result, it's not possible to rely on `instanceof Function` checks to work - there may be multiple `Function` constructors present.

This commit reworks `instanceof Function` check to use `typeof` instead.

See also https://github.com/strongloop/loopback-next/issues/3159

## Checklist


bajtos:
I am leaving out testing this change because I don't fully understand how is Jest creating the situation where there is more than one `Function` constructor present and I suspect that recreating this situation in Mocha tests would be too cumbersome.

PR #3184, opened 2019-06-19
Labs dev/passport adapter

jannyHou:
- I setup a sample app on local to test the released lab package, found some missing part, so create a PR to add them
- Update the sequence to be consistent with others in our docs/tests. cc @emonddr 

## Checklist

### DIFF
--- a/labs/authentication-passport/README.md
+++ b/labs/authentication-passport/README.md
@@ -38,6 +38,8 @@ Taking the basic strategy exported from
 first create an instance of the basic strategy with your `verify` function.
 
 ```ts
+// Create a file named `my-basic-auth-strategy.ts` to define your strategy below
+
 import {BasicStrategy} from 'passport-http';
 
 function verify(username: string, password: string, cb: Function) {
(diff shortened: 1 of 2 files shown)


labs/authentication-passport/README.md
```diff
@@ -31,13 +31,15 @@ the one that LoopBack 4 authentication system wants.
 
 ### Simple Usage
 
-1. Create an instance of the passport strategy
+- 1. Create an instance of the passport strategy
```
dhmlau: i think the original numbering looks good. wondering why we need to change it. :)

labs/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-adapter.acceptance.ts
```diff
@@ -162,14 +164,42 @@ describe('Basic Authentication', () => {
...
+          // It is necessary to catch these errors and add HTTP-specific status
+          // code property.
+          //
+          // Errors thrown by the strategy implementations already come
+          // with statusCode set.
+          //
+          // In the future, we want to improve `@loopback/rest` to provide
+          // an extension point allowing `@loopback/authentication` to contribute
+          // mappings from error codes to HTTP status codes, so that application
+          // don't have to map codes themselves.
```
nabdelgadir: nitpick: `so that application don't` -> `so that the application doesn't`

labs/authentication-passport/README.md
jannyHou: @dhmlau If you look at the npm registry page: https://www.npmjs.com/package/@loopback/authentication-passport

They all show as `1.` instead of a series of numbers 😞

dhmlau: _I think_ if you make the code snippet indented one more level, the number should be fine.

labs/authentication-passport/README.md
```diff
@@ -80,6 +91,37 @@ app
   });
 ```
 
+- 4. Decorate your endpoint
+
+To authenticate your request with the basic strategy, decorate your controller
+function like:
+
+```ts
+import {AUTH_STRATEGY_NAME} from './my-basic-auth-strategy';
```
b-admike: nitpick: it seems like we assume `my-basic-auth-strategy` file is in `src/controllers` here and `src/` from the previous statement in the `application.ts` file. can we have it at a consistent place like `src/`?

jannyHou: @b-admike I don't want to have a strong implication of the "right" file structure for users, they should decide it, that's why I just leave all the file import as `./file_name`.

PR #3185, opened 2019-06-19
docs: injection of current user may require optional set to true

emonddr:
Optional flag may need to be set to true when injecting
CURRENT_USER in a controller constructor if only
some controller methods are decorated with the
@authenticate decorator

Related to : https://github.com/strongloop/loopback-next/issues/2940

## Checklist

### DIFF
--- a/docs/site/decorators/Decorators_authenticate.md
+++ b/docs/site/decorators/Decorators_authenticate.md
@@ -40,5 +40,8 @@ export class WhoAmIController {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/site/decorators/Decorators_authenticate.md
```diff
@@ -40,5 +40,8 @@ export class WhoAmIController {
 }
 ```
 
+{% include note.html content="If only <b>some</b> of the controller methods are decorated with the <b>@authenticate</b> decorator, then the injection decorator for CURRENT_USER in the controller's constructor must be specified as <b>@inject(AuthenticationBindings.CURRENT_USER, {optional:true})</b> to avoid a binding error when an unauthenticated endpoint is accessed. Alternatively, do not inject CURRENT_USER in the controller <b>constructor</b>, but in the controller <b>methods</b> which are actually decorated with the <b>@authenticate</b> decorator. See [Method Injection](../Dependency-injection.md#method-injection), [Constructor Injection](../Dependency-injection.md#constructor-injection) and [Optional Dependencies](../Dependency-injection.md#optional-dependencies) for details.
+" %}
```
agnes512: I don't know if you need the double quotation here for the markdown file

emonddr: @agnes512 

Hi Agnes,

I followed the LoopBack documentation concerning documentation : tips, alerts, notes etc
https://loopback.io/doc/en/contrib/alerts.html

image

They use `"`.

But yes in javascript, `'` can be used in place of `"` in most situations.

agnes512: 👍

jannyHou: I think the double quotation is necessary, since `content="` begins with one.

PR #3171, opened 2019-06-17
feat(example-todo-list): add navigational properties to todo-list example

nabdelgadir:
See https://github.com/strongloop/loopback-next/issues/2633.

- Overwrote `find` and `findById` functions in `TodoRepository`, `TodoListRepository`, `TodoListImage` to include a hard-coded retrieval of related models.
  - https://github.com/strongloop/loopback-next/issues/3195 is for improvement of this
- Updated response schemas for controller methods `find` and `findById` to leverage `getModelSchemaRef` and `includeRelations`
- Updated `TodoList` tutorial

Note: since `HasOneRepository<T>` doesn't have a `find` function, the `image` property wasn't included in the `TodoList` instances as part of this PR.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3188, opened 2019-06-20
feat(repository): include navigation properties in JSON

bajtos:
Modify the code converting Model instances to JSON data to include navigational properties for the defined model relations.

See https://github.com/strongloop/loopback-next/pull/3171 and https://github.com/strongloop/loopback-next/issues/2633

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/model/model.unit.ts
+++ b/packages/repository/src/__tests__/unit/model/model.unit.ts
@@ -4,7 +4,13 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {expect} from '@loopback/testlab';
-import {Entity, ModelDefinition, STRING} from '../../../';
+import {
+  Entity,
+  HasManyDefinition,
+  ModelDefinition,
+  RelationType,
+  STRING,
+} from '../../../';
 
 describe('model', () => {
   const customerDef = new ModelDefinition('Customer');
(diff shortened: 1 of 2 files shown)


nabdelgadir:
From https://github.com/strongloop/loopback-next/pull/3171:

```ts
it('includes TodoList in query result', async () => {
    const list = await givenTodoListInstance(todoListRepo);
    const todo = await givenTodoInstance(todoRepo, {todoListId: list.id});
    const filter = JSON.stringify({include: [{relation: 'todoList'}]});
    const response = await client.get('/todos').query({filter: filter});
    expect(response.body).to.have.length(1);
    expect(response.body[0]).to.deepEqual({
      ...toJSON(todo),
      todoList: toJSON(list),
});
```

This test along with the symmetrical `includes Todo` test were both failing because:
- When the code reaches https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/writer.ts#L54 (a call to `JSON.stringify`)
- It comes to this [`toJSON`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L203) function 
- In this function, since the properties in `*Relations` aren't decorated by `@property` and in the model class, then they aren't part of the model definition's properties so they get stripped which is why the test fails.

From @bajtos:

> As for describing navigational properties in model definitions: originally, `@hasMany` was calling `@property` decorator under the hood. This had its own problems though, for example the navigational properties were then described in the request body of "create" and "update" operations, which is not desirable. That's why we moved away from the design where navigational properties are part of the model schema.
> 
> I think we need to find a way allowing Model's `toJSON` (and `toObject`)  methods to understand that even in `strict` mode where no extra properties are allowed, the navigational properties should be still included in the result.

packages/repository/src/__tests__/unit/model/model.unit.ts
```diff
@@ -194,6 +200,102 @@ describe('model', () => {
...
+        name: 'products',
+        type: RelationType.hasMany,
+        targetsMany: true,
+
+        source: Category,
+        keyFrom: 'id',
+
+        target: () => Product,
+        keyTo: 'categoryId',
+      });
```
raymondfeng: nitpick: IIRC, `{...} as HasManyDefinition` is a recommended style over `<HasManyDefinition>{...}`.

bajtos: Makes sense. From what I remember, when using `{...} as HasManyDefinition`, TypeScript is not able to offer correct suggestions (auto-completion) for members of the object being created :(

Issue #3190, opened 2019-06-20
Explain how custom authentication strategy options should be managed


emonddr:
## Description / Steps to reproduce / Feature proposal

Provide an example of how to manage `default` and `request-level` options in

https://loopback.io/doc/en/lb4/Loopback-component-authentication.html

## Expected Behavior

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

dhmlau:
Closing as done.

PR #3194, opened 2019-06-20
docs: show how to process options for an authentication strategy

emonddr:
Show an example of how to pass default and overriding options
into a custom authentication strategy.

Related to: https://github.com/strongloop/loopback-next/issues/3190

## Checklist

### DIFF
--- a/docs/site/Loopback-component-authentication.md
+++ b/docs/site/Loopback-component-authentication.md
@@ -170,7 +170,7 @@ An example of the decorator when options **are** specified looks like this:
 ```
 
 {% include tip.html content="
-To avoid repeating the same options in the <b>@authenticate</b> decorator for many endpoints in a controller, you can instead define global options which can be injected into an authentication strategy thereby allowing you to avoid specifying the options object in the decorator itself. For controller endpoints that need to override a global option, you can specify it in an options object passed into the decorator. Your authentication strategy would need to handle the option overrides.
+To avoid repeating the same options in the <b>@authenticate</b> decorator for many endpoints in a controller, you can instead define global options which can be injected into an authentication strategy thereby allowing you to avoid specifying the options object in the decorator itself. For controller endpoints that need to override a global option, you can specify it in an options object passed into the decorator. Your authentication strategy would need to handle the option overrides. See [Managing Custom Authentication Strategy Options](#managing-custom-authentication-strategy-options) for details.
 " %}
 
 After a request is successfully authenticated, the current user profile is
@@ -517,7 +517,7 @@ interfaces which may be of use to your application:
 and
 [TokenService](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/services/token.service.ts).
 
-### Registering a Custom Authentication Strategy
+## Registering a Custom Authentication Strategy
 
 The **registration** and **discovery** of authentication strategies is possible
 via the [Extension Point and Extensions](Extension-point-and-extensions.md)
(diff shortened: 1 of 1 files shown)


emonddr:
In my example I use [property injection](https://loopback.io/doc/en/lb4/Dependency-injection.html#property-injection) for the `options` variable.

Just wondering if that is ok, or if I should use [constructor injection](https://loopback.io/doc/en/lb4/Dependency-injection.html#constructor-injection) instead.

And what exactly is the difference, and why is constructor injection considered the preferred way between the two?

docs/site/Loopback-component-authentication.md
```diff
@@ -608,6 +608,201 @@ export class MyApplication extends BootMixin(
 }
 ```
 
+## Managing Custom Authentication Strategy Options
```
raymondfeng: It's easy to define a new decorator that accepts default options:

```
export function myAuthenticate(options: ...) {
  options = Object.assign(DEFAULT_OPTIONS, options);
  return authenticate(options);
}
```

Now `@myAuthenticate` is the new decorator.

docs/site/Loopback-component-authentication.md
```diff
@@ -608,6 +608,201 @@ export class MyApplication extends BootMixin(
...
+
+```ts
+export class MyApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options?: ApplicationConfig) {
+    super(options);
+
+    //...
+    this.bind(BasicAuthenticationStrategyBindings.DEFAULT_OPTIONS).to({
```
raymondfeng: We should start to use `this.configure(...)` more.

raymondfeng: See https://loopback.io/doc/en/lb4/Context.html#configuration-by-convention

emonddr: > We should start to use `this.configure(...)` more.

Hi Raymond, in my other PRs for authentication documentation (Authentication, and Shopping Cart), I don't use yet use `this.configure`. For this PR, since it is a new section in the Authentication document, I want to keep things consistent and provide similar examples with this.bind.

docs/site/Loopback-component-authentication.md
bajtos:
```ts
options = Object.assign(DEFAULT_OPTIONS, options);
```

This will modify `DEFAULT_OPTIONS` and thus affect subsequent calls of `myAuthenticate`.

Here is a better solution:

```ts
options = {...DEFAULT_OPTIONS, ...options};
```

emonddr: @raymondfeng ,

I have tried this approach and it works for sure....but only if user has 1 authentication strategy.

If user needs to use various authentication strategies across their different controller endpoints, this new decorator loads the specific default options of 1 authentication strategy, and won't be applicable to the other N strategies.

To load specific default options for a specific strategy, the user could also define a unique decorator for each strategy (which loads the default options for that strategy) 

```
import {authenticate as originalAuthenticate} from '@loopback/authentication';

const BASIC_DEFAULT_OPTIONS = {gatherStatistics: true};
export function basicAuthenticate(strategyName: string, options?: object) {
  console.log('\nInside my authenticate\n');
  const mergedOptions = Object.assign(BASIC_DEFAULT_OPTIONS, options);
  return originalAuthenticate(strategyName, mergedOptions);
}

const JWT_DEFAULT_OPTIONS = {someOtherProperty: 'someValue'};
export function jwtAuthenticate(strategyName: string, options?: object) {
  console.log('\nInside my authenticate\n');
  const mergedOptions = Object.assign(JWT_DEFAULT_OPTIONS, options);
  return originalAuthenticate(strategyName, mergedOptions);
}

```

emonddr: @raymondfeng  @bajtos 

If the user is going to create a custom authenticate decorator to load default options for > 1 authentication strategies, then we need something more like:

```ts
import {authenticate as originalAuthenticate} from '@loopback/authentication';

const JWT_DEFAULT_OPTIONS = {gatherStatistics: true};
const BASIC_DEFAULT_OPTIONS = {someOtherProperty: ''};

export function authenticate(strategyName: string, options?: object) {

  let defaultOptions;

  switch (strategyName) {
    case 'basic':
      console.log('assigning BASIC_DEFAULT_OPTIONS');
      defaultOptions = BASIC_DEFAULT_OPTIONS;
      break;
    case 'jwt':
      console.log('assigning JWT_DEFAULT_OPTIONS');
      defaultOptions = JWT_DEFAULT_OPTIONS;
      break;
    default:
      console.log('assigning {} for default options');
      defaultOptions = {};
      break;
  }
  const mergedOptions = Object.assign(defaultOptions, options);
  return originalAuthenticate(strategyName, mergedOptions);
}
```

raymondfeng: 1. -1 to override `@authenticate`. The sugar decorators should have their own names.
2. -1 to use switch statements. We should define separate decorators such as`@basicAuthenticate` and `@jwtAuthenticate` instead. Or `@authenticate.basic` and `@authenticate.jwt`.

emonddr: > Or @authenticate.basic and @authenticate.jwt 

I like this approach.

emonddr: I tried defining `authenticate.basic` and `authenticate.jwt` but my editor didn't like it.

packages/shopping/src/decorators/my.authenticate.decorator.ts : 

```ts
import {authenticate} from '@loopback/authentication';

const JWT_DEFAULT_OPTIONS = {gatherStatistics: true};
const BASIC_DEFAULT_OPTIONS = {someOtherProperty: ''};

export function authenticateBasic(options?: object) {
  return authenticate('basic', Object.assign({}, BASIC_DEFAULT_OPTIONS, options));
}

export function authenticateJWT(options?: object) {
  return authenticate('jwt', Object.assign({}, JWT_DEFAULT_OPTIONS, options));
}
```

emonddr: I am wondering how all this will affect `specifying authentication decorator at the controller class level`? Didn't someone ask about such a feature?

@jannyHou @raymondfeng   ^

emonddr: and now a strategy doesn't need to inject any DEFAULT options, it only needs to obtain the `merged` options from the decorator metadata:

e.g.

```ts
import {inject} from '@loopback/context';
import {HttpErrors, Request} from '@loopback/rest';
import {
  AuthenticationStrategy,
  UserProfile,
  TokenService,
  AuthenticationMetadata,
  AuthenticationBindings,
} from '@loopback/authentication';
import {Getter} from '@loopback/core';
import {
  TokenServiceBindings,
  BasicAuthenticationStrategyBindings,
  AuthenticationStrategyOptions,
} from '../keys';

export class JWTAuthenticationStrategy implements AuthenticationStrategy {
  name: string = 'jwt';

  options: AuthenticationStrategyOptions = {};

  constructor(
    @inject(TokenServiceBindings.TOKEN_SERVICE)
    public tokenService: TokenService,
    @inject.getter(AuthenticationBindings.METADATA)
    readonly getMetaData: Getter<AuthenticationMetadata>,
  ) {}

    const token: string = this.extractCredentials(request);

    this.options = await this.getOptions();

    if (this.options.gatherStatistics === true) {
      console.log(`\nGathering statistics...\n`);
    } else {
      console.log(`\nNot gathering statistics...\n`);
    }

    const userProfile: UserProfile = await this.tokenService.verifyToken(token);
    return userProfile;
  }

  extractCredentials(request: Request): string {
    if (!request.headers.authorization) {
      throw new HttpErrors.Unauthorized(`Authorization header not found.`);
    }

    // for example : Bearer xxx.yyy.zzz
    const authHeaderValue = request.headers.authorization;

    if (!authHeaderValue.startsWith('Bearer')) {
      throw new HttpErrors.Unauthorized(
        `Authorization header is not of type 'Bearer'.`,
      );
    }

    //split the string into 2 parts : 'Bearer ' and the `xxx.yyy.zzz`
    const parts = authHeaderValue.split(' ');
    if (parts.length !== 2)
      throw new HttpErrors.Unauthorized(
        `Authorization header value has too many parts. It must follow the pattern: 'Bearer xx.yy.zz' where xx.yy.zz is a valid JWT token.`,
      );
    const token = parts[1];

    return token;
  }

  async getOptions(): Promise<AuthenticationStrategyOptions> {
    const controllerMethodMetadata = await this.getMetaData();
    return controllerMethodMetadata.options
      ? controllerMethodMetadata.options
      : {};
  }
}

```

jannyHou: I am still trying to understand why we should provide the default option in decorator `@authentication()`.

This option overriding flow in Dom's doc makes good sense to me:

> As previously mentioned in the
[Using the Authentication Decorator](#using-the-authentication-decorator)
section, a custom authentication strategy should avoid repeatedly specifying its
**default** options in the **@authenticate** decorator. Instead, it should
define its **default** options in one place, and only specify **overriding**
options in the **@authenticate** decorator when necessary.

I agree on Raymond's comment of "using `@config()`" as improvement, BUT let's address it in a separate PR :) The scope in the doc story that Dom works on is already overhead.

So my take would be:

- In this PR, we stay with what Dom has now: 
  - bind each strategy's default option to app/server
  - override the default option in decorator `@authenticate('basic', {...the special option per request})`
  - the `authenticate()` method in each strategy class is responsible to handle the option override.
 
- Apply `@config` in code and update the doc in a separate story.

raymondfeng: @emonddr See https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L238 for an example to use TypeScript namespace to define `@authenticate.*`.

bajtos: I agree with @jannyHou's proposal. It seems like a good middle ground between what we have now (and need to improve) and `@config` based solution we want to implement in the future.

BTW can somebody open a new GH issue for leveraging `@config` in authentication strategies?

emonddr: > @emonddr See https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L238 for an example to use TypeScript namespace to define @authenticate.*

Thanks, Raymond. Good idea about using namespace for this.

We use namespace in a few places in authentication

e.g. 

https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/keys.ts#L14

emonddr: So @raymondfeng , based on @jannyHou  and @bajtos ' comments, are you ok with my approach (in the PR (not using custom decorators) for handling global and request-level options ? I think it is one (of many) acceptable ways of doing it, and if the user is clever, they can decide to implement it any way they choose (custom decorators, or hard-coding (not injecting) the default options in the strategy itself, etc). 

My only remaining question (asked earlier) is what the difference is between doing property injection in a class or constructor injection in a class. In my case, in my strategy class, I am using property injection for the default options. But the docs on injection seem to favor construction injection over property injection. I just want to know why. thx 

@bajtos ^

bajtos: Here is the rule of thumb:
- Constructor-level injection should be used for dependencies that are required by the class to correctly work and that don't have any reasonable default to use.
- Property-level injection should be used for dependencies that have a good default value that will be used by most callers, and we want to expose this dependency to allow advanced users to customize the default behavior.

It would be great to capture this rule in our docs, perhaps in https://loopback.io/doc/en/lb4/Dependency-injection.html#flavors-of-dependency-injection. @emonddr would you like to contribute that change yourself?

docs/site/Loopback-component-authentication.md
```diff
@@ -608,6 +608,201 @@ export class MyApplication extends BootMixin(
...
+
+### Bind the Default Options
+
+Bind the **default** options of the custom authentication strategy to the
+application `application.ts` via the
+`BasicAuthenticationStrategyBindings.DEFAULT_OPTIONS` binding key.
+
+In this hypothetical example, our custom authentication strategy has a
+**default** option of `gatherStatistics` with a value of `true`. (In a real
+custom authentication strategy, the number of options could be more numerous)
```
b-admike: nitpick: `could be more numerous` -> `could be more` or `could be numerous`

docs/site/Loopback-component-authentication.md
b-admike: Maybe we can create a story to follow-up on changing configuration of bindings to use the new way in not just this package but other places too?

docs/site/Loopback-component-authentication.md
b-admike: Well, I'm 50/50 on this one. I don't feel strongly for the two approaches described here. I think both make sense, and since we already have docs on overriding inside the strategy class, we can keep them as is and follow up on them. From a UX perspective, though, I think creating custom decorators to modify the default options is better.

emonddr:
Thanks @raymondfeng ! 

Thanks @b-admike  !

PR #3199, opened 2019-06-21
Emit schema with all model properties optional + allow partial updates via PATCH

bajtos:
- Allow Controller methods implementing PATCH operation to describe their input data as "a model with all properties optional". To emit schema with optional model properties, we are adding a new schema-generation option called `partial`.
- Modify CLI templates to use the new `partial` option to describe request body parameter.
- Modify example applications to leverage this new option too.
- Update code snippets in our docs for consistency.

See https://github.com/strongloop/loopback-next/issues/2652, https://github.com/strongloop/loopback-next/issues/1722 and https://github.com/strongloop/loopback-next/issues/1179

## Checklist


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -16,10 +16,44 @@ import {JSONSchema6 as JSONSchema} from 'json-schema';
...
+): string {
+  // backwards compatibility
+  switch (Object.keys(options).length) {
+    case 0:
+      return MODEL_TYPE_KEYS.ModelOnly;
+    case 1:
+      if (options.includeRelations) return MODEL_TYPE_KEYS.ModelWithRelations;
+  }
+
+  const flags = Object.keys(options);
```
b-admike: nitpick: can we define flags before the switch block and use it throughout?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -16,10 +16,44 @@ import {JSONSchema6 as JSONSchema} from 'json-schema';
...
+   * @private
+   */
   visited?: {[key: string]: JSONSchema};
 }
 
+/**
+ * @private
+ */
+export function buildModelCacheKey(
+  ctor: Function,
```
b-admike: Is there a reason for the `ctor` parameter in this function? I don't see it being used.

bajtos: Good catch, I'll remove it.

packages/repository-json-schema/src/build-schema.ts
bajtos: Do you meant to use `flags.length` instead of `Object.keys(options).length`? Is there any other place where to use `flags`?

b-admike: Yea that's right 👍

packages/repository-json-schema/src/keys.ts
```diff
@@ -15,6 +19,6 @@ export const enum MODEL_TYPE_KEYS {
  * Metadata key used to set or retrieve repository JSON Schema
  */
 export const JSON_SCHEMA_KEY = MetadataAccessor.create<
-  {[options in MODEL_TYPE_KEYS]: JSONSchema},
```
samarpanB: Any specific reason for removing this type check ?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -16,10 +16,44 @@ import {JSONSchema6 as JSONSchema} from 'json-schema';
...
+  options: JsonSchemaOptions = {},
+): string {
+  // backwards compatibility
+  switch (Object.keys(options).length) {
+    case 0:
+      return MODEL_TYPE_KEYS.ModelOnly;
+    case 1:
+      if (options.includeRelations) return MODEL_TYPE_KEYS.ModelWithRelations;
```
samarpanB: Don’t you think cyclomatic complexity increases with if inside switch? Just a thought.

examples/todo-list/src/controllers/todo.controller.ts
```diff
@@ -89,7 +90,14 @@ export class TodoController {
...
     @param.path.number('id') id: number,
-    @requestBody() todo: Todo,
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {partial: true}),
+        },
+      },
+    })
+    todo: Partial<TodoList>,
```
samarpanB: This should be ToDO

examples/todo-list/src/controllers/todo-list.controller.ts
```diff
@@ -80,7 +81,14 @@ export class TodoListController {
...
   async updateAll(
-    @requestBody() obj: Partial<TodoList>,
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {partial: true}),
+        },
+      },
+    })
+    obj: Partial<TodoList>,
```
nabdelgadir: nitpick: `obj` -> `todoList`

examples/todo-list/src/controllers/todo-list.controller.ts
```diff
@@ -108,7 +116,14 @@ export class TodoListController {
...
     @param.path.number('id') id: number,
-    @requestBody() obj: TodoList,
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {partial: true}),
+        },
+      },
+    })
+    obj: Partial<TodoList>,
```
nabdelgadir: nitpick: `obj` -> `todoList`

examples/todo-list/src/controllers/todo.controller.ts
```diff
@@ -89,7 +90,14 @@ export class TodoController {
   })
   async updateTodo(
     @param.path.number('id') id: number,
-    @requestBody() todo: Todo,
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {partial: true}),
```
nabdelgadir: `getModelSchemaRef(TodoList, {partial: true}),` -> `getModelSchemaRef(Todo, {partial: true}),`

packages/repository-json-schema/src/keys.ts
```diff
@@ -6,6 +6,10 @@
 import {MetadataAccessor} from '@loopback/metadata';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 
+/**
+ * TODO(semver-major) remove these constants in the next major version
+ * @deprecated
+ */
```
nabdelgadir: Can I ask why they're being removed?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -16,10 +16,44 @@ import {JSONSchema6 as JSONSchema} from 'json-schema';
...
+   * Set this flag if you want the schema to define navigational properties
+   * for model relations.
+   */
   includeRelations?: boolean;
+
+  /**
+   * Set this flag to mark all model properties as optional. This is typically
+   * used to describe request body of PATCH endpoints.
+   */
+  partial?: boolean;
```
raymondfeng: In the future, we can support more options such as `exclude`.

bajtos: Definitely, see e.g. https://github.com/strongloop/loopback-next/issues/2653:

> Allow Controller methods implementing CREATE operation to describe their input data as "a model without `id` and `_rev` properties ". This story requires #2629 and #2631 to be implemented first.
> 
> To emit schema with certain properties excluded, we should add a new schema-generation option called `exclude` & accepting a list of property names to remove from the schema. To allow TypeScript compiler to verify that `exclude` items are valid property names, we should leverage TypeScript generics and `keyof` keyword.

packages/repository-json-schema/src/keys.ts
bajtos: In the future, we will be adding more schema options like `exclude` (see https://github.com/strongloop/loopback-next/issues/2653). Because options can be combined, e.g. `partial` with `exclude`, the number of possible combinations (and thus cache) keys is growing fast. I find it impractical to enumerate all valid keys in a TypeScript enum type. We will need 4 entries now, 8 entries when we implement `exclude`, 16 entries when we introduce another option, etc.

That's why I switched the indexer type from `options in MODLE_TYPE_KEYS` to `key: string`.

packages/repository-json-schema/src/build-schema.ts
bajtos: Good point, I am going to simplify this function to make it easier to understand and reason about. Thank you for pointing it out! ❤️

packages/repository-json-schema/src/keys.ts
samarpanB: Makes sense.

packages/repository-json-schema/src/keys.ts
bajtos: Of course you can! 😄 

Cross-posting my comment from the discussion thread https://github.com/strongloop/loopback-next/pull/3199#discussion_r296590295:

> In the future, we will be adding more schema options like `exclude` (see https://github.com/strongloop/loopback-next/issues/2653). Because options can be combined, e.g. `partial` with `exclude`, the number of possible combinations (and thus cache) keys is growing fast. I find it impractical to enumerate all valid keys in a TypeScript enum type. We will need 4 entries now, 8 entries when we implement `exclude`, 16 entries when we introduce another option, etc.
>
> That's why I switched the indexer type from `options in MODLE_TYPE_KEYS` to `key: string`.

I am going to improve this comment to tell developers to use `buildModelCacheKey` instead.

bajtos:
@b-admike @nabdelgadir @raymondfeng @samarpanB Thank you for the review. I believe I have addressed your comments in 6e167805a and 59f0dbc67. LGTY now?

nabdelgadir:
> Should we perhaps use `+` instead of `;`?
> 
> ```ts
> expect(key).to.equal('includeRelations+partial');
> ```

I like `+` more than `;` personally

Issue #979, opened 2018-02-08
New Issues Template


virkt25:
I think the current issues template is super complicated and cluttered, making it harder to create an issue. See below for the original template. 

```
<!--
Questions:
  https://groups.google.com/forum/#!forum/loopbackjs
  https://gitter.im/strongloop/loopback
Immediate support:
  https://strongloop.com/api-connect-faqs/
  https://strongloop.com/node-js/subscription-plans/
-->

# Description/Steps to reproduce

<!--
If feature: A description of the feature
If bug: Steps to reproduce
-->

# Link to reproduction sandbox

<!--
Link to an app sandbox for reproduction

Note: Failure to provide a sandbox application for reproduction purposes will result in the issue being closed.
-->

# Expected result

<!--
Also include actual results if bug
-->

# Additional information

<!--
Copy+paste the output of these two commands:
  node -e 'console.log(process.platform, process.arch, process.versions.node)'
  npm ls --prod --depth 0 | grep loopback
-->
```

## Expected Result

Issues template should be simple and easy to understand. Example / Proposed version below. 

```
## Type
- [ ] Question
- [ ] Feature
- [ ] Bug / Issue

<!--
Bug / Issue => Include: Actual Result, Expected Result, Versions, Link to Sandbox (if available)
For Versions => Copy+Paste the output of these commands:
  node -e 'console.log(process.platform, process.arch, process.versions.node)'
  npm ls --prod --depth 0 | grep loopback
-->

## Description / Steps to Reproduce

```

dhmlau:
I wonder if we want our users to include the package name / function area: https://github.com/strongloop/loopback-next/blob/master/MONOREPO.md#packages. 

Just an example.. If someone is trying out the `example-getting-started` and having problem, it would help us by putting the pkg name on the title or the issue description.  For obvious things, we can encourage them to put this info.

raymondfeng:
Let's create a PR to capture the proposed template. It will make the commenting/reviewing simpler.

dhmlau:
I'll leave it as part of the task for https://github.com/strongloop/loopback-next/issues/47 then.

bajtos:
I agree the current template is too complex and should be simplified 👍 

```
## Type
- [ ] Question
- [ ] Feature
- [ ] Bug / Issue
```

[...]

PR #3202, opened 2019-06-21
chore: better templates for GitHub issues

bajtos:
Let's leverage GitHub's ability to offer users multiple issue templates and let us decide which issue labels to apply depending on the template chosen.

The proposed templates are loosely inspired by:
- Our current template in https://github.com/strongloop/loopback-next/blob/8ae8a0a81db205f052b81caaceece5303cd80ff2/.github/ISSUE_TEMPLATE.md
- The template used by LoopBack 3.x: https://github.com/strongloop/loopback/blob/9b1d4885ab869d977bc7d464aa861b4f88691862/.github/ISSUE_TEMPLATE.md
- TypeScript templates: https://github.com/microsoft/TypeScript/tree/master/.github/ISSUE_TEMPLATE

Link to our previous discussion about issue template(s): https://github.com/strongloop/loopback-next/issues/979

## Checklist


bajtos:
You can try out the proposed templates here: https://github.com/bajtos/issue-template-test/issues/new/choose

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,47 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+
+---
+
+<!-- 🚨 STOP 🚨 STOP 🚨 STOP 🚨
```
dhmlau: When comparing this "bug report" and "question", this one is part of the comment block and the other is not.  Do we want to make it consistent?

.github/ISSUE_TEMPLATE/Question.md
```diff
@@ -0,0 +1,25 @@
+---
+name: Question
+about: The issue tracker is not for questions. Please use Stack Overflow or other resources for help.
```
dhmlau: I thought we still are good to allow questions to be submitted here, however, not encouraged?

.github/ISSUE_TEMPLATE/Bug_report.md
bajtos: In "bug report", the "stop" line is just in the comment - we don't want it to be visible when the issue is created.

I thought it may be better to render it visible in "question" and "security vulnerability" issues.

On the second thought, I guess it does not make sense. If the issue is created despite the fact that we asked people not to, then there is little value in having a big "stop" alert at the top.

I'll change the templates to always put the "stop" alert into a comment.

.github/ISSUE_TEMPLATE/Question.md
bajtos: Good point. Can you propose a better version of this text please?

bajtos: Feel free to edit the template directly, it may be the fastest way.

bajtos:
I am going to land this pull request as-is. We can improve the template for Questions in follow-up pull requests.

Issue #2092, opened 2018-11-28
CLI issues: Generated app default Project name invalid if CWD name has hyphens + why not generate in CWD


cloudwheels:
## Current Behavior

`lb4 app` suggests a project name based on CWD. If this has hyphens they are replaced with spaces so the  default suggested name is invalid.

More importantly, why bother suggesting this if it's going to create a sub-directory anyway. 
Why not either:
- create the project in CWD (that's why I created a folder for it!)
- if it must be created in a subdir, just suggest "`package`" or "`code`", since I don't really want a structure `my-project/my-project` by default (From my home the default suggestion creates `~/username` as the project directory.)

I can see the logic of not just spawning into my home directory, for example, without warning and I know other generators follow this pattern, but my preference would be just make directory, copy into that with WARN this is the current directory & make clear the Project name is really for the `package.json`.

See *"Other examples"* below.

**ALSO**
I am asked to enter the 
> Project root directory

but if I enter a path, the generator appears to complete, but doesn't actually create the directory:
```bash
? Project root directory: ~/my-lb4-app/code
create .npmrc
create .prettierignore
create .prettierrc
...
added 528 packages in 18.892s
Application nigel was created in ~/my-lb4-app/code
nigel@cloudshell:~$ cd ~/my-lb4-app/code
-bash: cd: /home/nigel/my-lb4-app/code: No such file or directory
```
Edit: actually it is there:
```
nigel@cloudshell:~ (cw-wadh)$ ls *~*
my-lb4-app
```
Ewww.

*Other examples*:

```bash
nigel@cloudshell:~$ mkdir loopback-next && cd $_ && lb4 app
? Project name: (loopback next) 
>> Invalid npm package name: loopback next
```

```bash
? Project name: loopback-next
Application loopback-next was created in loopback-next.
nigel@cloudshell:~/loopback-next$ ls
loopback-next
```

cloudwheels:
removed
>  + copy error not thrown
See above.

dhmlau:
@cloudwheels, 
> If this has hyphens they are replaced with spaces so the default suggested name is invalid.

I just scaffolded an application with project name `test-1-2-3`, and it seems to be working as expected.  It is being created in the `test-1-2-3` directory.   Is it what you're seeing too? 
```
$ lb4 app
? Project name: test-1-2-3
? Project description: loopback4 example
? Project root directory: test-1-2-3
```

> ? Project root directory: ~/my-lb4-app/code

[...]

PR #3252, opened 2019-06-26
fix(cli): fix app default project name. untildify the project path

agnes512:
## Description 
In the [issue 2092](https://github.com/strongloop/loopback-next/issues/2092), the **default project name** could be generated incorrectly if the cwd folder name has hyphens, fixed. The **input app name** has no such issue.

The user also suggested that `lb4` should allow the project to be generated in the CWD. However, that could be messy and confusing. The other issue is that `lb4` allows user to use any project path as they want. This could be problematic: 
```javascript
(CWD: User/lb4-app)
? Project name: pathFromRoot
? Project root directory:  ~/try/pathFromRoot
...
Application pathFromRoot was created in ~/try/pathFromRoot

lb4-app$ ls
~
lb4-app$ cd ~
[goes to the root folder]
```
The change here is to untildify the input path to make sure that it always uses absolute path.

resolves https://github.com/strongloop/loopback-next/issues/2092

## Checklist

### DIFF
--- a/packages/cli/lib/project-generator.js
+++ b/packages/cli/lib/project-generator.js
@@ -4,14 +4,12 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
+
 const BaseGenerator = require('./base-generator');
 const utils = require('./utils');
 const chalk = require('chalk');
-<<<<<<< HEAD
 const cliVersion = require('../package.json').version;
-=======
 const path = require('path');
->>>>>>> af20999f... fix(cli): fix app default project name. relevant test added
 
 module.exports = class ProjectGenerator extends BaseGenerator {
   // Note: arguments and options should be defined in the constructor.
--- a/packages/cli/lib/project-generator.js
+++ b/packages/cli/lib/project-generator.js
@@ -7,7 +7,11 @@
 const BaseGenerator = require('./base-generator');
 const utils = require('./utils');
 const chalk = require('chalk');
+<<<<<<< HEAD
 const cliVersion = require('../package.json').version;
+=======
+const path = require('path');
+>>>>>>> af20999f... fix(cli): fix app default project name. relevant test added
 
 module.exports = class ProjectGenerator extends BaseGenerator {
   // Note: arguments and options should be defined in the constructor.
@@ -132,7 +136,8 @@ module.exports = class ProjectGenerator extends BaseGenerator {
         name: 'name',
         message: 'Project name:',
         when: this.projectInfo.name == null,
-        default: this.options.name || this.appname,
+        default:
+          this.options.name || utils.toFileName(path.basename(process.cwd())),
         validate: utils.validate,
       },
       {
--- a/packages/cli/lib/utils.js
+++ b/packages/cli/lib/utils.js
@@ -26,6 +26,8 @@ const connectors = require('../generators/datasource/connectors.json');
 const stringifyObject = require('stringify-object');
 const camelCase = _.camelCase;
 const kebabCase = _.kebabCase;
+const untildify = require('untildify');
+const tildify = require('tildify');
 const readdirAsync = promisify(fs.readdir);
 const toFileName = name => {
   return kebabCase(name).replace(/\-(\d+)$/g, '$1');
@@ -127,6 +129,8 @@ exports.camelCase = camelCase;
 exports.toVarName = toVarName;
 exports.pluralize = pluralize;
 exports.urlSlug = urlSlug;
+exports.untildify = untildify;
+exports.tildify = tildify;
 
 exports.validate = function(name) {
   const isValid = validate(name).validForNewPackages;
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -63,6 +63,8 @@
     "unicode-10.0.0": "^0.7.5",
     "update-notifier": "^3.0.1",
     "url-slug": "^2.1.1",
+    "untildify": "^4.0.0",
+    "tildify": "^2.0.0",
     "validate-npm-package-name": "^3.0.0",
     "yeoman-generator": "^4.0.1"
   },
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -13,11 +13,8 @@ const tildify = require('tildify');
 const assert = require('yeoman-assert');
 const helpers = require('yeoman-test');
 const generator = path.join(__dirname, '../../../generators/app');
-<<<<<<< HEAD
 const cliVersion = require('../../../package.json').version;
-=======
 const build = require('@loopback/build');
->>>>>>> af20999f... fix(cli): fix app default project name. relevant test added
 const props = {
   name: 'my-app',
   description: 'My app for LoopBack 4',
@@ -223,6 +220,7 @@ describe('app-generator with default values', () => {
   const defaultValProjPath = path.join(rootDir, 'sandbox/default-value-app');
   const sandbox = path.join(rootDir, 'sandbox');
   const pathToDefValApp = path.join(defaultValProjPath, 'default-value-app');
+  const cwd = process.cwd();
   const defaultValProps = {
     name: '',
     description: 'An app to test out default values',
@@ -249,5 +247,55 @@ describe('app-generator with default values', () => {
   after(() => {
     process.chdir(sandbox);
     build.clean(['node', 'run-clean', defaultValProjPath]);
+    process.chdir(cwd);
+  });
+});
+
+/** For testing the support of tilde path as the input of project path.
+ * Use differnt paths to test out the support of `~` when the test runs outside of home dir.
+ */
+describe('app-generator with tilde project path', () => {
+  const rootDir = path.join(__dirname, '../../../..');
+  // tildify the path:
+  let sandbox = path.join(rootDir, 'sandbox/tilde-path-app');
+  let pathWithTilde = tildify(sandbox);
+  const cwd = process.cwd();
+
+  // If the test runs outside $home directory
+  if (process.env.CI && !process.env.DEBUG && tildify(sandbox) === sandbox) {
+    sandbox = path.join(os.homedir(), '.lb4sandbox/tilde-path-app');
+    pathWithTilde = '~/.lb4sandbox/tilde-path-app';
+  }
+  const tildePathProps = {
+    name: 'tildified-path',
+    description: 'An app to test out tilde project path',
+    outdir: pathWithTilde,
+  };
+
+  before(async function() {
+    // Increase the timeout to accommodate slow CI build machines
+    // eslint-disable-next-line no-invalid-this
+    this.timeout(30 * 1000);
+    // check it with full path. tilde-path-app should not exist at this point
+    assert.equal(fs.existsSync(sandbox), false);
+    await helpers
+      .run(generator)
+      .inDir(sandbox)
+      // Mark it private to prevent accidental npm publication
+      .withOptions({private: true})
+      .withPrompts(tildePathProps);
+  });
+  it('scaffold a new application for tilde-path-app', async () => {
+    // tilde-path-app should be created at this point
+    assert.equal(fs.existsSync(sandbox), true);
+  });
+  after(function() {
+    // Increase the timeout to accommodate slow CI build machines
+    // eslint-disable-next-line no-invalid-this
+    this.timeout(30 * 1000);
+
+    process.chdir(sandbox);
+    build.clean(['node', 'run-clean', sandbox]);
+    process.chdir(cwd);
   });
 });
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -6,12 +6,18 @@
 'use strict';
 
 const fs = require('fs');
+const os = require('os');
 const {promisify} = require('util');
 const path = require('path');
+const tildify = require('tildify');
 const assert = require('yeoman-assert');
 const helpers = require('yeoman-test');
 const generator = path.join(__dirname, '../../../generators/app');
+<<<<<<< HEAD
 const cliVersion = require('../../../package.json').version;
+=======
+const build = require('@loopback/build');
+>>>>>>> af20999f... fix(cli): fix app default project name. relevant test added
 const props = {
   name: 'my-app',
   description: 'My app for LoopBack 4',
@@ -210,3 +216,38 @@ function testFormat() {
 process.env.CI && !process.env.DEBUG
   ? describe.skip
   : describe('app-generator with --format', testFormat);
+
+/** For testing if the generator handles default values properly */
+describe('app-generator with default values', () => {
+  const rootDir = path.join(__dirname, '../../../..');
+  const defaultValProjPath = path.join(rootDir, 'sandbox/default-value-app');
+  const sandbox = path.join(rootDir, 'sandbox');
+  const pathToDefValApp = path.join(defaultValProjPath, 'default-value-app');
+  const defaultValProps = {
+    name: '',
+    description: 'An app to test out default values',
+    outdir: '',
+  };
+
+  before(async () => {
+    // default-value-app should not exist at this point
+    assert.equal(fs.existsSync(defaultValProjPath), false);
+    assert.equal(fs.existsSync(pathToDefValApp), false);
+    return (
+      helpers
+        .run(generator)
+        .inDir(defaultValProjPath)
+        // Mark it private to prevent accidental npm publication
+        .withOptions({private: true})
+        .withPrompts(defaultValProps)
+    );
+  });
+  it('scaffold a new app for default-value-app', async () => {
+    // default-value-app should be created at this point
+    assert.equal(fs.existsSync(pathToDefValApp), true);
+  });
+  after(() => {
+    process.chdir(sandbox);
+    build.clean(['node', 'run-clean', defaultValProjPath]);
+  });
+});
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -4684,6 +4684,11 @@
 					"requires": {
 						"ansi-regex": "^3.0.0"
 					}
+				},
+				"untildify": {
+					"version": "4.0.0",
+					"resolved": "https://registry.npmjs.org/untildify/-/untildify-4.0.0.tgz",
+					"integrity": "sha512-KK8xQ1mkzZeg9inewmFVDNkg3l5LUhoq9kN6iWYB/CC9YMG8HA+c1Q8HwDe6dEX7kErrEVNVBO3fWsVq5iDgtw=="
 				}
 			}
 		},
@@ -4914,6 +4919,11 @@
 				"readable-stream": "~2.3.6",
 				"xtend": "~4.0.1"
 			}
+		},
+		"tildify": {
+			"version": "2.0.0",
+			"resolved": "https://registry.npmjs.org/tildify/-/tildify-2.0.0.tgz",
+			"integrity": "sha512-Cc+OraorugtXNfs50hU9KS369rFXCfgGLpfCfvlc+Ud5u6VWmUQsOAa9HbTvheQdYnrdJqqv1e5oIqXppMYnSw=="
 		},
 		"timed-out": {
 			"version": "4.0.1",
@@ -5150,9 +5160,9 @@
 			}
 		},
 		"untildify": {
-			"version": "3.0.3",
-			"resolved": "https://registry.npmjs.org/untildify/-/untildify-3.0.3.tgz",
-			"integrity": "sha512-iSk/J8efr8uPT/Z4eSUywnqyrQU7DSdMfdqK4iWEaUVVmcP5JcnpRqmVMwcwcnmI1ATFNgC5V90u09tBynNFKA=="
+			"version": "4.0.0",
+			"resolved": "https://registry.npmjs.org/untildify/-/untildify-4.0.0.tgz",
+			"integrity": "sha512-KK8xQ1mkzZeg9inewmFVDNkg3l5LUhoq9kN6iWYB/CC9YMG8HA+c1Q8HwDe6dEX7kErrEVNVBO3fWsVq5iDgtw=="
 		},
 		"unzip-response": {
 			"version": "2.0.1",


packages/cli/lib/utils.js
```diff
@@ -103,8 +103,13 @@ exports.validateClassName = function(name) {
 
 /**
  * Validate project directory to not exist
+ * and check if the directory path is based on CWD
  */
 exports.validateNotExisting = function(projDir) {
+  const invalidPath = projDir.split('/').includes('~');
+  if (invalidPath) {
+    return util.format('The path shold be based on CWD: %s. ', projDir);
```
raymondfeng: This message is confusing.

packages/cli/lib/utils.js
```diff
@@ -103,8 +103,13 @@ exports.validateClassName = function(name) {
 
 /**
  * Validate project directory to not exist
+ * and check if the directory path is based on CWD
  */
 exports.validateNotExisting = function(projDir) {
+  const invalidPath = projDir.split('/').includes('~');
```
raymondfeng: I think we can disable `~` as a whole, for example, not allowing `~xyz`.

raymondfeng:
> The user also suggested that lb4 should allow the project to be generated in the CWD. However, that could be messy and confusing. 

I agree. Can the user type `.` for his case?

packages/cli/lib/utils.js
b-admike: nitpick: since users might not know what CWD is I suggest `current directory`

packages/cli/lib/project-generator.js
```diff
@@ -131,7 +132,7 @@ module.exports = class ProjectGenerator extends BaseGenerator {
         name: 'name',
         message: 'Project name:',
         when: this.projectInfo.name == null,
-        default: this.options.name || this.appname,
+        default: this.options.name || path.basename(process.cwd()),
```
raymondfeng: This might be problematic. For example, if I run `lb4` in `/tmp`, would `tmp` be proposed here with the change?

packages/cli/lib/utils.js
agnes512: I was not sure about that since `~` could be part of the folder name. But yeah if the folder name contains `~`, `lb4 app` would complain about it. What do you think?

agnes512:
> > The user also suggested that lb4 should allow the project to be generated in the CWD. However, that could be messy and confusing.
> 
> I agree. Can the user type `.` for his case?

Yes, `.` and `..` both work. That's why I mention the CWD in the warning msg. I will make the message more clear.

packages/cli/lib/utils.js
agnes512: yes I agree. I wasn't sure about the warning message. And I will add tests later.

packages/cli/lib/project-generator.js
agnes512: If you're in the `/tmp`, then yes the default project name will be `tmp`

packages/cli/lib/utils.js
emonddr: > shold 

should

packages/cli/lib/utils.js
bajtos: The value `~` is a special path interpreted by unix shells as the home directory, which is defined in the environment variable `$HOME`. Instead of rejecting `~` in the path, I'd prefer to interpret it the same way as the shell does:
 - On Unix systems like Linux and MacOS, replace it with `process.env.HOME`.
- On Windows, the value should be replaced with `process.env.userprofile`, see https://stackoverflow.com/questions/9228950/what-is-the-alternative-for-users-home-directory-on-windows-command-prompt

As for Raymond's proposal to disallow `~xyz`, it seems like an arbitrary limitation with no good reason behind it and as such I'd prefer to avoid such limitation.

raymondfeng: I don't think it's our CLI's responsibility to support shell conventions for a few reasons:

1. Windows users might interpret such chars differently, for example, not considering `~` as home directory

2. If we decide to support `~` for Unix, what about other shell features, such as env vars? IMO, `~` is just a shortcut, our users can always use the full path if necessary.

3. I think Unix shell only allows `~` or `~/<other>`.

packages/cli/lib/project-generator.js
raymondfeng: We don't always need to have a default name, right?

packages/cli/test/unit/utils.unit.js
```diff
@@ -469,6 +469,17 @@ describe('Utils', () => {
     });
   });
 
+  describe('validateNotExisting', () => {
+    it("returns true on the path contains the root folder '~'", () => {
+      expect(utils.validateNotExisting('~/try/pathFromRoot')).to.be.eql(
+        'The path should be based on current directory: ~/try/pathFromRoot.',
+      );
+    });
+    it("returns false on the path that folder's name contains hyphens", () => {
```
emonddr: @agnes512 

A hyphen is '-' . Where are the hyphens in this test?

emonddr: and the 'returns false...' name for the test...is incorrect because you are testing for 'True'.

packages/cli/test/unit/utils.unit.js
```diff
@@ -469,6 +469,17 @@ describe('Utils', () => {
     });
   });
 
+  describe('validateNotExisting', () => {
```
emonddr: > utils.validateNotExisting 

Seems strange to use a function with this name, when we are trying to figure out if '~' (root) is in the path...

emonddr: why not isRootInPath() ?

agnes512: it's the name of the function that it tests out

packages/cli/lib/project-generator.js
agnes512: I actually don't know why we need the default name for the project. If I were the user, I would be confused if the default name is the folder name TBH. Maybe we can remove the default name for this step?

packages/cli/lib/utils.js
agnes512: I agree that it's not our CLI's responsibility to support shell conventions. But I will go with whichever @rmg and @bajtos think is more reasonable.
We can warn the user about the usage of `~`. 
Or we can replace it with `process.env.HOME || process.env.userprofile`.

packages/cli/test/unit/utils.unit.js
agnes512: good catch!

packages/cli/lib/utils.js
rmg: Not sure if I was mentioned by typo or not, but I have opinions so I'll share them.

From a UX perspective, I personally would expect to be able to use `~` paths here because I do it in other interactive terminal programs all the time (eg. `vim`).

Proper expansion of `~` is surprisingly limited in scope and defined for only a few cases:

- leading `~`, bare or followed by `/`, expands to current user's home directory
  - eg. `~`, `~/`, `~/foo`
- leading `~` followed by a username expands to the given user's home directory
  - eg. `~rmg`, `~rmg/`, `~rmg/bar`

A `~` appearing anywhere other than as the first character (eg. `/~/foo`, `foo/~/bar`) should be treated as a literal `~` and not expanded or interpreted.

The simple case (current user only) can actually be completely handled inline:

```js
str.replace(/^~(?=\/|$)/, os.homedir())
```

Expanding the `~username` form would require some very OS-specific calls that make it a lot more complicated. But I'm not sure if anyone would ever use that form with this CLI, so I suspect nobody would even notice if it wasn't supported.

(A quick search reveals [untildify](https://github.com/sindresorhus/untildify), which confirms both the simplicity of the easy case and the complexity of the difficult case.)

Simplicity of adding it directly to the CLI aside, I think this sort of thing would actually make the most sense as part of inquirer itself.

agnes512: @raymondfeng

raymondfeng: I'm fine to use https://github.com/sindresorhus/untildify to handle `~`.

packages/cli/package.json
```diff
@@ -63,6 +63,7 @@
     "unicode-10.0.0": "^0.7.5",
     "update-notifier": "^3.0.0",
     "url-slug": "^2.1.1",
+    "untildify": "4.0.0",
```
raymondfeng: Please use a version range such as `"untildify": "^4.0.0",`.

packages/cli/package.json
```diff
@@ -63,7 +63,7 @@
     "unicode-10.0.0": "^0.7.5",
     "update-notifier": "^3.0.0",
     "url-slug": "^2.1.1",
-    "untildify": "4.0.0",
+    "untildify": "3.0.3",
```
raymondfeng: Why downgrade?

agnes512: Cause `yeoman` somehow has the `untildify` of version 3.0.3. It would complain about the conflict if I use 4.0.0 D:

agnes512: [in this file](https://github.com/strongloop/loopback-next/blob/master/packages/cli/package-lock.json#L5045) @raymondfeng

packages/cli/lib/utils.js
bajtos: +1 to use untildify

packages/cli/lib/project-generator.js
bajtos: FWIW, the application generator (`lb4 app`) has `applicationName` option, perhaps it was the intention to use that one? IDK.

https://github.com/strongloop/loopback-next/blob/08bf6cc8c1e1b3bfde74c2f6a6c575e6b6b3acb4/packages/cli/generators/app/index.js#L31-L34

I am little bit reluctant to land this change together with other changes in the same commit. Also, is there any easy way how to test this? E.g. create a new directory in the sandbox with a known name, then verify that the generator picked that directory name as the default name.

packages/cli/lib/project-generator.js
```diff
@@ -210,7 +211,7 @@ module.exports = class ProjectGenerator extends BaseGenerator {
   scaffold() {
     if (this.shouldExit()) return false;
 
-    this.destinationRoot(this.projectInfo.outdir);
+    this.destinationRoot(utils.untildify(this.projectInfo.outdir));
```
bajtos: We discussed on Slack that this is non-trivial to test. Here is an idea to try:

1. Create a sandbox directory in user's home. Pick a name that will make it clear that this is a sandbox for LB4 tests, because this directory may stay around even after the test suite finished (especially when it was aborted, e.g. by pressing `Ctrl+C`). Suggested name: `~/.lb4sandbox`. Use `os.homedir()` as suggested by Ryan to find out what's the user home directory (`path.join(os.homedir(), '.lb4sandbox')`).
2. Tell the generator to scaffold the new project in this special sandbox directory, use `~` in the path.
3. Verify that the generator expanded `~` and created the artifacts in the expected place.

Thoughts?

agnes512: I was thinking to test it by:

1. Check if folder exists: `homeDir/somePath/loopback-next/sandbox/my-app` (should be false)
2. For any local test, get pwd: `homeDir/somePath/loopback-next`

3. Input path: `~/somePath/loopback-next/sandbox/my-app`

4. Then check if folder exists: `homeDir/somePath/loopback-next/sandbox/my-app` (should be true) 

It might need some helper functions to get `somePath`

rmg: Note that `os.homedir()` honours `process.env.HOME`, so you can use it for testing in a sandbox:
```
$ node
> os.homedir()
'/Users/rmg'
> process.env.HOME = '/tmp/foo'
'/tmp/foo'
> os.homedir()
'/tmp/foo'
```

packages/cli/lib/utils.js
```diff
@@ -105,7 +106,7 @@ exports.validateClassName = function(name) {
  * Validate project directory to not exist
  */
 exports.validateNotExisting = function(projDir) {
-  if (fs.existsSync(projDir)) {
+  if (fs.existsSync(untildify(projDir))) {
```
rmg: Support for `~` is an entirely user-facing feature. I don't think it makes sense for an internal `util.validateNotExisting()` function to support it directly, especially when none of the other helpers support it.

rmg: Looking at where this is actually used, I think the `untildify` call should be added to the validator in-line. Something like this:

```diff
--- a/packages/cli/lib/project-generator.js
+++ b/packages/cli/lib/project-generator.js
@@ -159,7 +159,7 @@ module.exports = class ProjectGenerator extends BaseGenerator {
           this.projectInfo.outdir == null ||
           // prompts if option was set to a directory that already exists
           utils.validateNotExisting(this.projectInfo.outdir) !== true,
-        validate: utils.validateNotExisting,
+        validate: input => utils.validateNotExisting(utils.untildify(input)),
         default: utils.toFileName(this.projectInfo.name),
       },
     ];
```

packages/cli/test/acceptance/app-run.acceptance.js
```diff
@@ -14,25 +15,36 @@ const build = require('@loopback/build');
 describe('app-generator (SLOW)', function() {
   const generator = path.join(__dirname, '../../generators/app');
   const rootDir = path.join(__dirname, '../../../..');
-  const sandbox = path.join(rootDir, 'sandbox/sandbox-app');
+  const fullPathSandbox = path.join(rootDir, 'sandbox/sandbox-app');
+  const homeDir = process.env.HOME || process.env.userprofile;
```
rmg: Is there a reason not to use `os.homedir()` here? It will use the appropriate env var for the platform and also fallback to an appropriate user lookup call (yes, it even reads `/etc/passwd` on UNIX).

packages/cli/test/acceptance/app-run.acceptance.js
```diff
@@ -14,25 +15,36 @@ const build = require('@loopback/build');
...
+  const tildeSandbox = fullPathSandbox.replace(homeDir, '~');
```
rmg: I'm not a fan of using a bunch of tiny dependencies.. but the `untildify` dependency being used was named as the opposite of its sister package, [`tildify`](https://www.npmjs.com/package/tildify), which does exactly what you want here.

packages/cli/lib/project-generator.js
agnes512: I didn't know I could pass an input like that. Thanks!

packages/cli/lib/project-generator.js
```diff
@@ -210,7 +212,7 @@ module.exports = class ProjectGenerator extends BaseGenerator {
...
```
rmg: If you add `utils.untildify` as a `filter` to the prompt for `outdir` then this wouldn't be necessary.

Doing it there instead of here would scope these changes squarely in to the user-input processing instead of also changing some of the scaffolding logic.

packages/cli/test/acceptance/app-run.acceptance.js
```diff
@@ -15,15 +17,30 @@ describe('app-generator (SLOW)', function() {
   const generator = path.join(__dirname, '../../generators/app');
   const rootDir = path.join(__dirname, '../../../..');
   const sandbox = path.join(rootDir, 'sandbox/sandbox-app');
+  const sandboxDefaultVal = path.join(rootDir, 'sandbox/sandbox-default-value');
   const cwd = process.cwd();
   const appName = '@loopback/sandbox-app';
+
+  // for testing out the untildified path:
+  // use a path that contains ~ as the outdir of the app
+  const tildifySandbox = tildify(sandbox);
```
rmg: Something to be aware of.. if this test is run outside of the current user's `$HOME` directory, then `tildifySandbox` will be identical to `sandbox`. This means the tests will pass, even if they don't actually test the `~` support. This could come up on CI systems where tests are run inside `/tmp`, for example.

I don't think anything needs to be done about it, I just wanted to point it out.

packages/cli/lib/project-generator.js
rmg: If you want to get really fancy, you might even be able to add a `transformer: tildify`. If I understand the [docs](https://www.npmjs.com/package/inquirer#question), this would make it so if the user inputs `/Users/me/work/myNewApp` then the answer will be shorted to `~/work/myNewApp` when it is printed back to the user but it will be _stored_ as the absolute path.

rmg:
I'm a little confused/concerned about the coveralls failure. It appears that the `validate` callback is not called at all. I can't tell if that is because the helper bypasses the input validation or if there is some bug in istanbul.

agnes512:
> I'm a little confused/concerned about the coveralls failure. It appears that the `validate` callback is not called at all. I can't tell if that is because the helper bypasses the input validation or if there is some bug in istanbul.

XD
It works fine locally. It prompted the warning msg if the file already exists. I don' know how that happened, and also don't know what would happen if the callback doesn't get called at all. Let me try to put the `untildify()` back to `utils.js`.

packages/cli/lib/project-generator.js
agnes512: yeah I was doing the opposite way: stored the shorted path and generate the files by absolute path D: 
I guess the way you suggested is better as absolute paths are more reliable.

rmg:
Turns out it is the yeoman test helpers. Instead of giving the provided answers to the prompts they mock the prompts and inject the provided answers directly :-(

packages/cli/lib/project-generator.js
bajtos: I think this line can be reverted back, since you are calling `untildify` on L171 above. PTAL.

packages/cli/lib/utils.js
bajtos: +1 to what @rmg wrote. Can we revert this line and modify the `validate` function to call untildify before calling `validateNotExisting`?

packages/cli/test/acceptance/app-run.acceptance.js
bajtos: Let' capture this information in a code comment please. If we ever run into the problem described by @rmg, then it would be great to have some help in our code.

packages/cli/lib/utils.js
agnes512: I had `validate: input => utils.validateNotExisting(utils.untildify(input))` before, but the coverage dropped a bit ( like what @rmg said, it mocks the prompts). I guess it's fine to decrease the coverage a bit?

packages/cli/generators/app/index.js
```diff
@@ -148,7 +148,7 @@ module.exports = class AppGenerator extends ProjectGenerator {
     this.log();
     this.log('Next steps:');
     this.log();
-    this.log('$ cd ' + this.projectInfo.outdir);
+    this.log('$ cd ' + utils.tildify(this.projectInfo.outdir));
```
raymondfeng: I don't think we should `tildify` the `outdir` as `~` won't work for Windows.

packages/cli/lib/project-generator.js
```diff
@@ -159,12 +161,14 @@ module.exports = class ProjectGenerator extends BaseGenerator {
...
           utils.validateNotExisting(this.projectInfo.outdir) !== true,
-        validate: utils.validateNotExisting,
+        validate: input => utils.validateNotExisting(utils.untildify(input)),
         default: utils.toFileName(this.projectInfo.name),
       },
     ];
 
     return this.prompt(prompts).then(props => {
+      const fullPath = props.outdir;
+      props.outdir = utils.untildify(fullPath);
```
raymondfeng: It can be simplified as `props.outdir = utils.untildify(props.outdir);`

packages/cli/lib/utils.js
bajtos: > I guess it's fine to decrease the coverage a bit?

Sure, it's ok in cases like this.

packages/cli/test/acceptance/app-run.acceptance.js
```diff
@@ -15,15 +17,34 @@ describe('app-generator (SLOW)', function() {
...
+  without testing the ~ support. This could come up on CI systems where
+  tests are run inside /tmp, for example. */
+  // For testing out the untildified path:
+  // use a path that contains ~ as the outdir of the app
+  const tildifySandbox = tildify(sandbox);
   const props = {
     name: appName,
     description: 'My sandbox app for LoopBack 4',
-    outdir: sandbox,
+    outdir: tildifySandbox,
```
bajtos: I see that you are modifying an acceptance test suite that's primarily testing how a new app is scaffolded, whether dependencies can be installed, etc. I feel these tests should not deal with home-directory specific stuff and should be preserved with no changes.

Instead, you should write a new test that will focus only on dealing with home directories. This test can be much simpler and should go to https://github.com/strongloop/loopback-next/blob/spike%2Fresolve-included-models/packages/cli/test/integration/generators/app.integration.js

Ideally, the new test should also modify `process.env.HOME` to ensure we are testing tildify/untildify behavior on CI too, see https://github.com/strongloop/loopback-next/pull/3252#discussion_r300438024

Sorry for discovering this late.

packages/cli/lib/project-generator.js
bajtos: >  is there any easy way how to test this? E.g. create a new directory in the sandbox with a known name, then verify that the generator picked that directory name as the default name.

PTAL ☝️

agnes512:
@slnode test please

packages/cli/test/integration/generators/app.integration.js
```diff
@@ -165,6 +167,86 @@ describe('app-generator with --applicationName', () => {
...
+  const projPath = path.join(rootDir, 'sandbox/tilde-path-app');
+
+  /** NOTICE: if this test runs outside of the current user's $HOME directory,
+  then pathWithTilde will be identical to projPath. Means the tests will pass
+  without testing the ~ support. This could come up on CI systems where
+  tests are run inside /tmp, for example. */
+  if (process.env.CI && !process.env.DEBUG && tildify(projPath) === projPath) {
+    // warn the user that it doesn't test the support of `~`
+    console.error(
+      "WARNING: couldn't test the support of tilde in path under this test environment even though the test might pass",
```
raymondfeng: This fails CI - see https://ci.appveyor.com/project/strongloop/loopback-next/builds/25882737. We don't expect to have console output.

agnes512: Is there any way to warn the user if they run `npm t` under such test environment?

raymondfeng: I'm not sure why we want to warn the user.

agnes512: Or I can just leave a comment if that's better @raymondfeng

Issue #2634, opened 2019-03-22
Spike: Resolver for inclusion of related models


bajtos:
Continue the works started in #2124, leverage the building blocks implemented as an outcome of #2592 and create a detailed proposal for implementing resolver for navigational properties.

Implementation wise, I think we want to end up with the following user experience:

```ts
export class TodoListRepository extends DefaultCrudRepository<
  TodoList,
  typeof TodoList.prototype.id
> {
  public readonly todos: HasManyRepositoryFactory<
    Todo,
    typeof TodoList.prototype.id
  >;

  constructor(
    @inject('datasources.db') dataSource: juggler.DataSource,
    @repository.getter(TodoRepository)
    protected todoRepositoryGetter: Getter<TodoRepository>,
  ) {
    super(TodoList, dataSource);
    this.todos = this._createHasManyRepositoryFactoryFor(
      'todos',
      todoRepositoryGetter,
    );
   ////// THE FOLLOWING LINE IS NEWLY ADDED
   this._registerHasManyInclusion('todos', todoRepositoryGetter);
  }
}
```

The question is how to implement `_register**Inclusion` APIs and process the registered inclusion handlers in `find`/`findById` implementations provided by `DefaultCrudRepository`. See https://github.com/strongloop/loopback-next/issues/1352#issuecomment-436197152 for more details on that.

It is important to consider the scenario where we want to fetch multiple source models and include their related models in the result, e.g. `TodoList.find(1, {include: [{relation: 'todos'}]})`. Under the hood, we must avoid [SELECT N+1](https://stackoverflow.com/questions/97197/what-is-the-n1-select-query-issue) issue. Inclusion handler must be able to fetch related models for multiple source-model ids in one query (using `inq` operator under the hood). The design proposed in https://github.com/strongloop/loopback-next/issues/1352#issuecomment-436197152 is already taking that into account, so does the existing implementation in juggler.

## Acceptance criteria

- [ ] A draft pull request demonstrating the proposed solution:
  - Use `examples/todo-list` to show the proposed user experience
  - Provide a PoC implementation in `packages/repository`
  - Include a markdown document in the monorepo root  (`/_SPIKE_.md`) describing the proposal at a high level. Keeping the document in git (instead in GH comments) makes it easier to discuss individual parts of the proposal and keep the proposal versioned.
[...]

PR #3283, opened 2019-07-01
feat(testlab): improve typings for `toJSON()` helper

bajtos:
While working on https://github.com/strongloop/loopback-next/issues/2634, I discovered that `findOne` returns `T | null` and this type cannot be passed to `toJSON` helper, because there is no overload accepting such type.

This pull requests improves the type definitions to let the compiler understand toJSON's behavior for commonly used union types like `object | null` or `unknown[] | null`.

## Checklist

### DIFF
--- a/packages/testlab/src/__tests__/unit/to-json.test.ts
+++ b/packages/testlab/src/__tests__/unit/to-json.test.ts
@@ -62,6 +62,42 @@ describe('toJSON', () => {
     expectUndefined(value);
   });
 
+  it('handles `object | null`', () => {
+    const input = createValueOfUnionType<object | null>({});
+    const output = toJSON(input);
+    expectComplexType<object | null>(output, input);
+  });
+
+  it('handles `object | undefined`', () => {
+    const input = createValueOfUnionType<object | undefined>({});
+    const output = toJSON(input);
+    expectComplexType<object | undefined>(output, input);
+  });
+
+  it('handles `object | null | undefined`', () => {
+    const input = createValueOfUnionType<object | null | undefined>({});
+    const output = toJSON(input);
+    expectComplexType<object | null | undefined>(output, input);
+  });
+
+  it('handles `unknown[] | null`', () => {
+    const input = createValueOfUnionType<unknown[] | null>([]);
+    const output = toJSON(input);
+    expectComplexType<unknown[] | null>(output, input);
+  });
+
+  it('handles `unknown[] | undefined`', () => {
+    const input = createValueOfUnionType<unknown[] | undefined>([]);
+    const output = toJSON(input);
+    expectComplexType<unknown[] | undefined>(output, input);
+  });
+
+  it('handles `unknown[] | null | undefined`', () => {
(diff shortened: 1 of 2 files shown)


packages/testlab/src/__tests__/unit/to-json.test.ts
```diff
@@ -62,6 +62,50 @@ describe('toJSON', () => {
     expectUndefined(value);
   });
 
+  it('handles `object | null`', () => {
+    const input: object | null = Math.random() ? {} : null;
```
hacksparrow: Wouldn't this always yield `{}`?

raymondfeng: > The Math.random() function returns a floating-point, pseudo-random number in the range 0–1 (inclusive of 0, but not 1) with approximately uniform distribution over that range

I think `Math.random() ?` returns `true` in most cases.

packages/testlab/src/__tests__/unit/to-json.test.ts
```diff
@@ -62,6 +62,50 @@ describe('toJSON', () => {
...
+    const output = toJSON(input);
+    expectComplexType<object | null>(output, input);
+  });
+
+  it('handles `object | undefined`', () => {
+    const input: object | undefined = Math.random() ? {} : undefined;
```
raymondfeng: Please extract `Math.random()` ? to a helper function, such as: `randomChoice(...values: any[]): any`

packages/testlab/src/__tests__/unit/to-json.test.ts
hacksparrow: Yes, so the behavior of `input = null` will rarely get tested.

packages/testlab/src/__tests__/unit/to-json.test.ts
raymondfeng: For example:
```ts
function randomChoice(...values: any[]): any {
  if (!values.length) return undefined;
  const index = Math.floor(Math.random() * values.length);
  return values[index];
}
```

bajtos:
Thank you @raymondfeng and @hacksparrow for your comments. I think I really need to improve the tests to make it more clear what is going on and what is being tested. In the tests, I don't really care about the actual value being converted via `toJSON`, because the implementation is always the same. What I need: compile-time checks.

bajtos:
The main problem is that from what I can tell by observing type hints offered by VSCode, the following statement is optimized by TypeScript and a simpler type is used instead.

```ts
const input: object | null = {};
const output = toJSON(input);
// VSCode shows the following hint:
// (alias) toJSON(value: object): object (+14 overloads)
// import toJSON
```

Compare to my current version:

```ts
const input: object | null = Math.random() ? {} : null;
const output = toJSON(input);
// VSCode shows the following hint:
// (alias) toJSON(value: object | null): object | null (+14 overloads)
// import toJSON
```

bajtos:
@raymondfeng @hacksparrow updated, LGTY now?

hacksparrow:
LGTM if it's just for compile-time TS checks.

PR #3285, opened 2019-07-01
feat(boot): use @config to inject options for booters

raymondfeng:
Switch to `@config` and `ctx.configure` for booter options.

## Checklist

### DIFF
--- a/packages/boot/src/booters/datasource.booter.ts
+++ b/packages/boot/src/booters/datasource.booter.ts
@@ -3,16 +3,16 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {bind, config, inject} from '@loopback/context';
 import {CoreBindings} from '@loopback/core';
 import {
   ApplicationWithRepositories,
-  juggler,
   Class,
+  juggler,
 } from '@loopback/repository';
-import {inject} from '@loopback/context';
 import {ArtifactOptions} from '../interfaces';
-import {BaseArtifactBooter} from './base-artifact.booter';
 import {BootBindings} from '../keys';
+import {BaseArtifactBooter} from './base-artifact.booter';
 
 /**
  * A class that extends BaseArtifactBooter to boot the 'DataSource' artifact type.
@@ -24,12 +24,13 @@ import {BootBindings} from '../keys';
  * @param projectRoot - Root of User Project relative to which all paths are resolved
  * @param bootConfig - DataSource Artifact Options Object
  */
+@bind({tags: {configPath: 'datasources'}})
 export class DataSourceBooter extends BaseArtifactBooter {
   constructor(
     @inject(CoreBindings.APPLICATION_INSTANCE)
     public app: ApplicationWithRepositories,
     @inject(BootBindings.PROJECT_ROOT) projectRoot: string,
-    @inject(`${BootBindings.BOOT_OPTIONS}#datasources`)
+    @config()
     public datasourceConfig: ArtifactOptions = {},
   ) {
     super(
--- a/packages/boot/src/booters/datasource.booter.ts
+++ b/packages/boot/src/booters/datasource.booter.ts
@@ -3,15 +3,15 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {bind, config, inject} from '@loopback/context';
+import {config, inject} from '@loopback/context';
 import {CoreBindings} from '@loopback/core';
 import {
   ApplicationWithRepositories,
   Class,
   juggler,
 } from '@loopback/repository';
-import {ArtifactOptions} from '../interfaces';
 import {BootBindings} from '../keys';
+import {ArtifactOptions, booter} from '../types';
 import {BaseArtifactBooter} from './base-artifact.booter';
 
 /**
@@ -24,7 +24,7 @@ import {BaseArtifactBooter} from './base-artifact.booter';
  * @param projectRoot - Root of User Project relative to which all paths are resolved
  * @param bootConfig - DataSource Artifact Options Object
  */
-@bind({tags: {configPath: 'datasources'}})
+@booter('datasources')
 export class DataSourceBooter extends BaseArtifactBooter {
   constructor(
     @inject(CoreBindings.APPLICATION_INSTANCE)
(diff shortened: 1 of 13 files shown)


packages/boot/src/booters/datasource.booter.ts
```diff
@@ -24,12 +24,13 @@ import {BootBindings} from '../keys';
  * @param projectRoot - Root of User Project relative to which all paths are resolved
  * @param bootConfig - DataSource Artifact Options Object
  */
+@bind({tags: {configPath: 'datasources'}})
```
bajtos: I find it rather unusual to use `tags` as a mean for providing `configPath` option. My understanding so far was that `tags` are used to group bindings into tagged groups.

Let's find a better solution that will allow `@bind()` users to specify extra metadata.

I'd prefer the following:

```suggestion
@bind({configPath: 'datasources'})
```

If it would be too difficult to allow arbitrary top-level properties in `@bind` object, then let's introduce a new property dedicated to custom options, e.g.

```suggestion
@bind({options: {configPath: 'datasources'}})
```

bajtos: Either way, if this requires changes in `@bind` decorator, then please make those changes in a new pull request, so that we can review them in isolation.

raymondfeng: > I find it rather unusual to use tags as a mean for providing configPath option. My understanding so far was that tags are used to group bindings into tagged groups.

Tags are not only used for grouping but also for metadata. We already use tags such as `name`, `key`, and `namespace` to customize how a binding is created.

raymondfeng:
I added `@booter` to make it the booter binding simpler to configure.

packages/boot/src/booters/datasource.booter.ts
bajtos: > > I find it rather unusual to use `tags` as a mean for providing `configPath` option. My understanding so far was that `tags` are used to group bindings into tagged groups.
> >
> > Let's find a better solution that will allow `@bind()` users to specify extra metadata.
>
> Tags are not only used for grouping but also for metadata. We already use tags such as `name`, `key`, and `namespace` to customize how a binding is created.

Well, `name`, `key` and `namespace` make sense as tags to me, similarly to how physical products can be tagged with an EAN barcode and a price tag.

However, `configPath` seems different to me. It's not a metadata used to customize how a binding is created, it's a metadata specifying configuration options for other parts of our IoC system.

I feel we should separate those two kinds of metadata.

@strongloop/loopback-maintainers @strongloop/loopback-next thoughts?

bajtos: In https://github.com/strongloop/loopback-next/pull/3329, we renamed `configPath` to `propertyPath`. Let's do the same in this pull request.

Should we also allow `@bind()` to customize the target configuration binding too?

E.g.

```
@bind({config: {targetBindingKey: '...', propertyPath: 'datasources'}})
```

I guess this poses a question whether we want to enhance `@loopback/context` to support config customization from `@bind()` out of the box, or whether we keep that feature just inside `@loopback/boot` as and (undocumented) convention.

Thoughts?

packages/boot/src/keys.ts
```diff
@@ -27,6 +27,9 @@ export namespace BootBindings {
     'application.bootstrapper',
   );
 
-  export const BOOTER_TAG = 'booter';
```
bajtos: Please preserve this constant for backwards compatibility.

```ts
/**
 * @deprecated Use `BootTags.BOOTER` instead.
 */
export const BOOTER_TAG = BootTags.BOOTER;
```

packages/boot/src/types.ts
```diff
@@ -125,3 +126,15 @@ export interface Bootable {
    */
   booters(...booterClasses: Constructor<Booter>[]): Binding[];
 }
+
+/**
+ * `@booter` decorator
```
bajtos: Please explain the purpose of this decorator, why should people use it and especially how it affects `@config()` behavior.

packages/boot/src/booters/datasource.booter.ts
raymondfeng: The `configPath` is now changed to `artifactNamespace`. We use it to infer how the configuration for a given booter can be resolved against boot options.

raymondfeng: It's an interesting idea. We can possibly allow a binding to customize where its configuration should be resolved from.

packages/boot/src/types.ts
raymondfeng: Hmm, I just realized that api-extractor/api-documenter has trouble handling two artifacts with case-sensitive names, for examle, `Booter` interface and `@booter` decorator`. As a result, the docs for `@booter` is overridden by `Booter`. Let me open an issue against their repo and see how it goes.

raymondfeng: See https://github.com/microsoft/web-build-tools/issues/1281

packages/boot/src/booters/datasource.booter.ts
agnes512: ( I don't know if my question makes sense.) As we improved `@config` in [#3329](https://github.com/strongloop/loopback-next/pull/3329), does that mean we can customize the target  `booter` (or the modify the config of the booter) via `@config` here?

packages/boot/src/types.ts
```diff
@@ -125,3 +126,33 @@ export interface Bootable {
...
+ *   // ...
+ *   }
+ * }
+ * ```
+ *
+ * @param artifactNamespace - Namespace for the artifact. It will be used to
+ * inject configuration from boot options
+ * @param specs - Extra specs for the binding
+ */
+export function booter(artifactNamespace: string, ...specs: BindingSpec[]) {
```
agnes512: I don't really get why we name it `artifactNamespace` here. But I either don't know how we usually name these properties.. So I guess it's fine. 
Could you make the example above more specific, or add a comment for this change? It would help to understand the code/naming logic. Thanks!

packages/boot/src/booters/datasource.booter.ts
raymondfeng: @agnes512 It means two things:

1. We need to improve `@bind` to honor `config` related settings
2. We need to improve `@config.*()` to use metadata from `@bind` so that no binding key is required for config injection.

packages/boot/src/types.ts
raymondfeng: Improved the tsdocs for the decorator function.

PR #3286, opened 2019-07-01
feat(rest-explorer): use configuration apis/decorators

raymondfeng:
Switch to `@config` and `ctx.configure` for rest-explorer options.

## Checklist

### DIFF
--- a/packages/rest-explorer/src/rest-explorer.component.ts
+++ b/packages/rest-explorer/src/rest-explorer.component.ts
@@ -3,26 +3,28 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {inject} from '@loopback/context';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import {bind, config, ContextTags, inject} from '@loopback/context';
 import {Component, CoreBindings} from '@loopback/core';
-import {RestApplication, createControllerFactoryForClass} from '@loopback/rest';
+import {createControllerFactoryForClass, RestApplication} from '@loopback/rest';
+import {ExplorerController} from './rest-explorer.controller';
 import {RestExplorerBindings} from './rest-explorer.keys';
 import {RestExplorerConfig} from './rest-explorer.types';
-import {ExplorerController} from './rest-explorer.controller';
 
 const swaggerUI = require('swagger-ui-dist');
 
 /**
  * A component providing a self-hosted API Explorer.
  */
+@bind({tags: {[ContextTags.KEY]: RestExplorerBindings.COMPONENT.key}})
 export class RestExplorerComponent implements Component {
   constructor(
     @inject(CoreBindings.APPLICATION_INSTANCE)
     private application: RestApplication,
-    @inject(RestExplorerBindings.CONFIG, {optional: true})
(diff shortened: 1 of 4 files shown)


packages/rest-explorer/src/rest-explorer.keys.ts
```diff
@@ -4,13 +4,17 @@
...
 import {BindingKey} from '@loopback/context';
+import {RestExplorerComponent} from './rest-explorer.component';
 import {RestExplorerConfig} from './rest-explorer.types';
 
 /**
  * Binding keys used by this component.
  */
 export namespace RestExplorerBindings {
-  export const CONFIG = BindingKey.create<RestExplorerConfig>(
-    'rest-explorer.config',
```
bajtos: As discussed before, this is changing the actual string value of the binding key used for REST Explorer config.

Consider the following code:

```ts
this.bind('rest-explorer.config').to({
  path: '/openapi/ui',
});      
```

At the moment, this code works well and configures REST Explorer as intended.

With your changes in place, the code no longer works and the user is not notified about the problem.

Should we treat your pull request as a breaking change then? Perhaps we can claim that the actual key name `rest-explorer.config` was undocumented and therefore changes of that string are not considered as breaking backwards compatibility?

@strongloop/loopback-maintainers @strongloop/loopback-next thoughts?

raymondfeng: I'm leaning toward the claim that `rest-explorer.config` is an internal string.

raymondfeng: The other option is to enhance `@config` to accept the target binding key to override the one from the enclosing binding.

jannyHou: > I'm leaning toward the claim that rest-explorer.config is an internal string.

Yeah I would say +1, did a quick search in our existing doc, we don't mention `rest-explorer.config` anywhere, so unless users jump into the source code, they should always use the imported const `RestExplorerBindings.CONFIG`.

packages/rest-explorer/src/rest-explorer.component.ts
```diff
@@ -3,26 +3,28 @@
...
  * A component providing a self-hosted API Explorer.
  */
+@bind({tags: {[ContextTags.KEY]: RestExplorerBindings.COMPONENT.key}})
 export class RestExplorerComponent implements Component {
   constructor(
     @inject(CoreBindings.APPLICATION_INSTANCE)
     private application: RestApplication,
-    @inject(RestExplorerBindings.CONFIG, {optional: true})
-    config: RestExplorerConfig = {},
+    @config()
```
jannyHou: Just double check: is `{optional: true}` also needed for `@config()`? or `@config()` allows empty value?

bajtos: I believe `@config` is setting `optional: true` by default. Could you please confirm, @raymondfeng?

packages/rest-explorer/src/rest-explorer.keys.ts
bajtos: Fair enough, let's treat `rest-explorer.config` as an internal string :+1:

PR #3297, opened 2019-07-02
Emit schema excluding certain model properties

nabdelgadir:
See https://github.com/strongloop/loopback-next/issues/2653.

- Add `exclude` option to emit model schema excluding certain model properties.
- Update `POST` in example applications to exclude `id`.
  - Update their docs for consistency.
- Update the CLI templates to also exclude `id` from `POST` request body.

## Checklist

### DIFF
--- a/examples/todo-list/src/controllers/todo-list.controller.ts
+++ b/examples/todo-list/src/controllers/todo-list.controller.ts
@@ -38,8 +38,17 @@ export class TodoListController {
       },
     },
   })
-  async create(@requestBody() obj: TodoList): Promise<TodoList> {
-    return await this.todoListRepository.create(obj);
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {exclude: ['id']}),
+        },
+      },
+    })
+    todoList: Omit<TodoList, 'id'>,
+  ): Promise<TodoList> {
+    return await this.todoListRepository.create(todoList);
   }
 
   @get('/todo-lists/count', {
--- a/packages/cli/generators/relation/templates/controller-relation-template-has-many.ts.ejs
+++ b/packages/cli/generators/relation/templates/controller-relation-template-has-many.ts.ejs
@@ -55,7 +55,13 @@ export class <%= controllerClassName %> {
   })
   async create(
     @param.path.<%= sourceModelPrimaryKeyType %>('id') id: typeof <%= sourceModelClassName %>.prototype.<%= sourceModelPrimaryKey %>,
-    @requestBody() <%= targetModelRequestBody %>: <%= targetModelClassName %>,
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(<%= targetModelClassName %>, {exclude: ['id']}),
+        },
+      },
+    }) <%= targetModelRequestBody %>: Omit<<%= targetModelClassName %>, 'id'>,
   ): Promise<<%= targetModelClassName %>> {
     return await this.<%= paramSourceRepository %>.<%= relationPropertyName %>(id).create(<%= targetModelRequestBody %>);
   }
--- a/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
+++ b/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
@@ -232,13 +232,28 @@ describe('build-schema', () => {
       expect(key).to.equal('modelPartial');
     });
 
+    it('returns "excluding[id,_rev]" when a single option "exclude" is set', () => {
+      const key = buildModelCacheKey({exclude: ['id', '_rev']});
+      expect(key).to.equal('modelExcluding[id,_rev]');
+    });
+
+    it('does not include "exclude" in concatenated option names if it is empty', () => {
+      const key = buildModelCacheKey({
+        partial: true,
+        exclude: [],
+        includeRelations: true,
+      });
+      expect(key).to.equal('modelPartialWithRelations');
+    });
+
     it('returns concatenated option names otherwise', () => {
       const key = buildModelCacheKey({
         // important: object keys are defined in reverse order
         partial: true,
+        exclude: ['id', '_rev'],
         includeRelations: true,
       });
-      expect(key).to.equal('modelPartialWithRelations');
+      expect(key).to.equal('modelPartialExcluding[id,_rev]WithRelations');
     });
   });
 });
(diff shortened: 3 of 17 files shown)


packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,18 @@ describe('build-schema', () => {
       expect(key).to.equal('modelPartial');
     });
 
+    it('returns "excluding[]" when a single option "exclude" is set', () => {
+      const key = buildModelCacheKey({exclude: []});
```
bajtos: Please use a non-empty list of excluded properties, to verify how are the properties encoded in the key.

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,18 @@ describe('build-schema', () => {
...
+
     it('returns concatenated option names otherwise', () => {
       const key = buildModelCacheKey({
-        // important: object keys are defined in reverse order
         partial: true,
+        exclude: [],
         includeRelations: true,
       });
-      expect(key).to.equal('modelPartialWithRelations');
+      expect(key).to.equal('modelPartialExcluding[]WithRelations');
```
bajtos: I find this ugly. Let's treat `exclude: []` as if it was not provided at all.

```suggestion
      expect(key).to.equal('modelPartialWithRelations');
```

I am proposing the following:
- Add a new test to verify `buildModelCacheKey({exclude: []})`
- Modify this test to add some items to `exclude` configuration

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -15,7 +15,8 @@ import {
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY, MODEL_TYPE_KEYS} from './keys';
 
-export interface JsonSchemaOptions {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface JsonSchemaOptions<T extends object = any> {
```
bajtos: Let's discuss the default type value. When we use `any`, any code that's not providing `T` is basically throwing away all type checks and there is no warning/error about that problem reported by the compiler.

I am proposing to not provide any default value for `T`, in which case TypeScript will pick `object` as the default. Alternatively, we can also specify `{}` as the default type. To be honest, I am not sure what's the different between `object` and `{}` and cannot tell what is a better option to use.

My expectation is that when `T` is set to `object` or `{}`:
- `keyof T` is resolved as `never`
- an attempt to set `exclude` to a non-empty array fails with a compilation error
- the developer is forced to specify T value for `JsonSchemaOptions`

I see that `export function buildModelCacheKey<T extends object>` already does not provide any default value for `T`, thus falling back to `object` as the default. Since that seems to work well, let's do the same for JsonSchemaOptions too.

```suggestion
export interface JsonSchemaOptions<T extends object> {
```

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
bajtos:
```suggestion
      const key = buildModelCacheKey({exclude: ['id', '_rev']});
```

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,27 @@ describe('build-schema', () => {
...
 
-    it('returns concatenated option names otherwise', () => {
+    it('returns "excluding[id,_rev]" when a single option "exclude" is set', () => {
+      const key = buildModelCacheKey({exclude: ['id', '_rev']});
+      expect(key).to.equal('modelExcluding[id,_rev]');
+    });
+
+    it('does not include "exclude" in concatenated option names if it is empty', () => {
       const key = buildModelCacheKey({
-        // important: object keys are defined in reverse order
```
bajtos: Please preserve this comment, I believe it's still relevant.

packages/cli/generators/relation/templates/controller-relation-template-has-many.ts.ejs
```diff
@@ -55,7 +55,13 @@ export class <%= controllerClassName %> {
   })
   async create(
     @param.path.<%= sourceModelPrimaryKeyType %>('id') id: typeof <%= sourceModelClassName %>.prototype.<%= sourceModelPrimaryKey %>,
-    @requestBody() <%= targetModelRequestBody %>: <%= targetModelClassName %>,
+    @requestBody({
```
jannyHou: The id field is not always called `id`, I think we need to get a model's id name in the template.

examples/todo-list/src/controllers/todo-list.controller.ts
```diff
@@ -38,8 +38,17 @@ export class TodoListController {
...
-    return await this.todoListRepository.create(obj);
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(TodoList, {exclude: ['id']}),
+        },
+      },
+    })
+    todoList: Omit<TodoList, 'id'>,
```
raymondfeng: Should we support `include` for `todoList: Pick<TodoList, 'id'>`?

raymondfeng: I'm fine to do it in a separate PR.

bajtos: I am not sure what would you use `include` for? A kind of a white-list? That would make sense :+1:

Let's discuss `include` and `Pick` in a separate PR. I think it's best to wait until there is (user) demand for that feature.

packages/cli/generators/relation/templates/controller-relation-template-has-many.ts.ejs
bajtos: Good point. AFAICT, `lb4 model` does allow users to create a primary key property with a different name than `id`.

Example prompts & answers to use `pk` instead of `id`:

```
Model Product will be created in src/models/product.model.ts

Let's add a property to Product
Enter an empty property name when done

? Enter the property name: pk
? Property type: string
? Is pk the ID property? Yes
? Is it required?: No
? Default value [leave blank for none]:
```

I agree with @jannyHou that we should honor custom PK name, there should be a new test to verify. If this is too much work then I can live with adding support for custom PK names in a follow-up pull request.

jannyHou: > If this is too much work then I can live with adding support for custom PK names in a follow-up pull request.

Yeah forgot to say, feel free to address it in another PR :)

nabdelgadir: Will do in a separate PR 👍

PR #3309, opened 2019-07-05
feat(repository-json-schema): add an option to make properties optional

nabdelgadir:
Add a new option `optional: []` so that `getJsonSchema` and related helpers can request a model schema that marks specified properties as optional

Connect to https://github.com/strongloop/loopback-next/issues/2653

## Checklist

### DIFF
--- a/packages/repository-json-schema/package.json
+++ b/packages/repository-json-schema/package.json
@@ -26,12 +26,14 @@
     "@loopback/context": "^1.20.2",
     "@loopback/metadata": "^1.2.5",
     "@loopback/repository": "^1.8.2",
-    "@types/json-schema": "^7.0.3"
+    "@types/json-schema": "^7.0.3",
+    "debug": "^4.1.1"
   },
   "devDependencies": {
     "@loopback/build": "^2.0.3",
     "@loopback/eslint-config": "^2.0.0",
     "@loopback/testlab": "^1.6.3",
+    "@types/debug": "^4.1.4",
     "@types/node": "^10.14.12",
     "ajv": "^6.10.2"
   },
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -983,7 +983,7 @@ describe('build-schema', () => {
         );
       });
 
-      it('doesn\'t exclude properties when the option "exclude" is set to exclude no properties', () => {
+      it(`doesn't exclude properties when the option "exclude" is set to exclude no properties`, () => {
         const originalSchema = getJsonSchema(Product);
         expect(originalSchema.properties).to.deepEqual({
           id: {type: 'number'},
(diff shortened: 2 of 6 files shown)


packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -932,5 +932,53 @@ describe('build-schema', () => {
...
+
+        @property({required: true})
+        name: string;
+
+        @property()
+        description: string;
+      }
+
+      it('makes one property optional when the option "optional" includes one property', () => {
+        const originalSchema = getJsonSchema(Product);
```
jannyHou: @nabdelgadir A question for the generic type `T` in this PR: if you don't specify the type when calling `getJsonSchema(ModelCtor)`, then where is it provided?

raymondfeng: I think TS 3.5.x defaults to `unknown`.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -255,11 +262,14 @@ export function getNavigationalPropertyForRelation(
...
   let suffix = '';
   if (options.partial) {
     suffix += 'Partial';
   }
+  if (options.optional && options.optional.length) {
+    suffix += 'Optional[' + options.optional + ']';
+  }
   if (options.includeRelations) {
     suffix += 'WithRelations';
   }
```
raymondfeng: Would it be simpler to have the title as:

```
<originalModelName>:<stringified-json-for-options>
```

bajtos: Please note the suffix is used in JSON/OpenAPI Schema titles. These titles are later used by code generators like swagger-codegen to emit interface or class names in TypeScript/Java/C#/etc. 

While it may be easier for us to build the prefix as stringified json, I believe it will provide suboptimal experience for our users and anybody consuming OpenAPI spec generated by LB4 applications.

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,27 @@ describe('build-schema', () => {
       expect(key).to.equal('modelPartial');
     });
 
-    it('returns concatenated option names otherwise', () => {
+    it('returns "optional[id,_rev]" when a single option "optional" is set', () => {
```
bajtos: The test name is misleading, you are setting two properties as optional.

```suggestion
    it('returns "optional[id,_rev]" when "optional" is set with two items', () => {
```

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,27 @@ describe('build-schema', () => {
...
+      const key = buildModelCacheKey({optional: ['id', '_rev']});
+      expect(key).to.equal('modelOptional[id,_rev]');
+    });
+
+    it('does not include "optional" in concatenated option names if it is empty', () => {
       const key = buildModelCacheKey({
-        // important: object keys are defined in reverse order
```
bajtos: Please preserve this comment in the test `it('returns concatenated option names otherwise', () => {`

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -232,13 +232,27 @@ describe('build-schema', () => {
...
+        optional: [],
         includeRelations: true,
       });
       expect(key).to.equal('modelPartialWithRelations');
     });
+
+    it('returns concatenated option names otherwise', () => {
+      const key = buildModelCacheKey({
+        partial: true,
+        optional: ['id', '_rev'],
```
bajtos: When `partial` is set to `true`, the `optional` config is effectively ignored, because all properties are optional. 

Should we tell the user that they are may be doing something wrong? Personally, I would throw an error if both `partial` and `optional` are set.

Regardless of the solution we choose, please add a test to verify what happens when both `partial` and `optional` are set.

---

On Slack, we discussed an alternative solution: instead of adding `optional?: (keyof T)[];`, we can modify `partial` to accept `boolean | keyof T)[]`. This removes the edge case where both optional and partial are set. My concern was that this solution may be more difficult to understand for our users.

nabdelgadir: The reason I opted to use `optional` is because I felt the name `partial` might be confusing. If you set `partial` to `[prop1, prop2]` then it might make it seem like these are the only two properties required and the rest are optional. What do you think?

raymondfeng: I agree with @nabdelgadir that `partial: ['a', 'b']` can be misleading.

IMO, `partial: true` becomes a shortcut of `optional` to mark all properties optional. My intuitive interpretation is the more specific one wins, i.e., `optional: ['a', 'b']` should override `partial: true`. WDYT?

nabdelgadir: I agree. I think `optional` should override `partial` as well.

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
nabdelgadir: So this is supposed to mean only the `optional` option is set (similar to the `partial` [test](https://github.com/strongloop/loopback-next/blob/e11e73acfd101e34ccc4ca775ef72c3e80d0d3ee/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts#L230) above it: `returns "partial" when a single option "partial" is set'`), but I'll reword this one.

packages/repository-json-schema/package.json
```diff
@@ -26,7 +26,9 @@
     "@loopback/context": "^1.20.2",
     "@loopback/metadata": "^1.2.5",
     "@loopback/repository": "^1.8.2",
-    "@types/json-schema": "^7.0.3"
+    "@types/json-schema": "^7.0.3",
+    "debug": "^4.1.1",
+    "@types/debug": "^4.1.4"
```
raymondfeng: Nitpick: it's good enough to be a dev dep.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -1001,5 +1001,79 @@ describe('build-schema', () => {
...
+        const optionalIdAndNameSchema = getJsonSchema(Product, {
+          optional: ['id', 'name'],
+        });
+        expect(optionalIdAndNameSchema.required).to.equal(undefined);
+        expect(optionalIdAndNameSchema.title).to.equal(
+          'ProductOptional[id,name]',
+        );
+      });
+
+      it('doesn\'t make properties optional when the option "optional" includes no properties', () => {
```
raymondfeng: Nitpick: we can use ` to quote the string to avoid escaping, such as:
```
`doesn't make properties optional when the option "optional" includes no properties`
```

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
bajtos: > I agree. I think `optional` should override `partial` as well.

Makes sense 👍 

I feel it's important to describe this rule in (API) docs and have a test to verify the behavior.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -1001,5 +1001,79 @@ describe('build-schema', () => {
...
+        const originalSchema = getJsonSchema(Product);
+        expect(originalSchema.required).to.deepEqual(['id', 'name']);
+        expect(originalSchema.title).to.equal('Product');
+
+        const optionalNothingSchema = getJsonSchema(Product, {optional: []});
+        expect(optionalNothingSchema.required).to.deepEqual(['id', 'name']);
+        expect(optionalNothingSchema.title).to.equal('Product');
+      });
+
+      it('overrides "partial" option when "optional" options is set', () => {
```
bajtos: 👍

Issue #2169, opened 2018-12-18
Using custom (underscored) keys for Relations : "Cannot read property 'target' of undefined"


esoodev:
## Description
I am using an existing MySQL database and its schemas, and am following [the tutorial](https://loopback.io/doc/en/lb4/Relations.html) to build relations.

## Current Behavior
- User has many Review
- Review belongs to User

Due to the existing schema, Review has 'user_id' field that links to User's 'id.' I coded this relation in review.model.ts accordingly. 
```
@belongsTo(() => User, {keyFrom: 'user_id', keyTo: 'id'})
user_id: number;
```
However, doing so gives me `...Cannot read property 'target' of undefined...` error whenever ReviewRepository is called.
Modifying the code and its relevant parts like below does not cause any issues.
```
@belongsTo(() => User, {keyTo: 'id'})
userId: number;
```

## Comments
How can I keep 'user_id' as a valid field for Review? Any help is appreciated.

esoodev:
It seems this is related to [this issue](https://www.bountysource.com/issues/6164944-provide-a-way-to-specify-table-and-field-naming-convention-with-automigrate-autoupdate) first mentioned back in 2014.

esoodev:
Within ReviewRepository, instead of
`this.user = this.createBelongsToAccessorFor('user', userRepositoryGetter);`

I used

```
this.user = createBelongsToAccessor(
  {
    target: () => User,
    name: 'user',
    type: RelationType.belongsTo,
    source: Review,
    keyFrom: 'user_id',
    keyTo: 'id',
  },
  userRepositoryGetter,
  this,
);
```

This works for now, though not pretty. If anyone can provide an insight for moving this into Review model, it'd be much appreciated!

JuliaRakitina:
+1 there is no clear documentation to build relations. 
And by the way, I did not find anywhere in source code function createBelongsToAccessorFor

JuliaRakitina:
Please improve documentation
here https://loopback.io/doc/en/lb4/BelongsTo-relation.html  you wrote 
`call the createBelongsToAccessorFor function in the constructor of the source repository class with the relation name (**decorated relation property on the source model**) and target repository instance and assign it the property mentioned above.`

In code snippets source model decorated prorepty is **customerId** , but    
this.customer = this.createBelongsToAccessorFor(
      '**customer**', - I suppose it should be   **customerId**
      customerRepositoryGetter,
    );

JuliaRakitina:
[...]

PR #3326, opened 2019-07-09
Custom foreign key name

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/2169.

* Added constraint to make sure the foreign key in `keyTo` actually exists in the target model
* Added tests to see if custom names (not the standard `{ModelName}Id` work e.g. with underscores or a completely different name) for the foreign key also work. 

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/acceptance/belongs-to.relation.acceptance.ts
+++ b/packages/repository/src/__tests__/acceptance/belongs-to.relation.acceptance.ts
@@ -3,15 +3,19 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {Application} from '@loopback/core';
+import {expect} from '@loopback/testlab';
 import {
+  ApplicationWithRepositories,
   juggler,
   repository,
   RepositoryMixin,
-  ApplicationWithRepositories,
 } from '../..';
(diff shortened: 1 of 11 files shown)


packages/repository/src/__tests__/fixtures/repositories/address.repository.ts
```diff
@@ -29,7 +29,7 @@ export class AddressRepository extends DefaultCrudRepository<
     customerRepositoryGetter: Getter<CustomerRepository>,
   ) {
     super(Address, db);
-    this.customer = this._createBelongsToAccessorFor(
+    this.customer = this.createBelongsToAccessorFor(
```
nabdelgadir: Because of this change, there's no longer coverage for `_createBelongsToAccesorFor`: https://coveralls.io/builds/24455533/source?filename=packages/repository/src/repositories/legacy-juggler-bridge.ts#L259. Should I revert it?

b-admike: It's deprecated, so not sure what the best course of action should be. For now, we could duplicate the tests for coverage of both until we remove it. @bajtos thoughts?

nabdelgadir: I reverted it so the function is still covered, but I can open another PR if we want to change it to the non-deprecated version.

PR #3329, opened 2019-07-09
feat(context): allow @config.* to specify the target binding key

raymondfeng:
Sometimes we want to inject configuration from another binding instead of
the current one.

This PR allows the following config injections:

```ts
@config({targetBindingKey: 'parent', configPath: 'child1'})
@config.getter({targetBindingKey: 'parent', configPath: 'child2'})
@config({targetBindingKey: 'anotherBinding'})
```

## Checklist

### DIFF
--- a/packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
@@ -59,7 +59,7 @@ describe('Context bindings - injecting configuration for bound artifacts', () =>
     expect(server1.configObj).to.eql({port: 3000});
   });
 
-  it('allows configPath for injection', async () => {
+  it('allows propertyPath for injection', async () => {
     class RestServerWithPort {
       constructor(@config('port') public port: number) {}
     }
@@ -67,6 +67,69 @@ describe('Context bindings - injecting configuration for bound artifacts', () =>
     // Bind configuration
     ctx
       .configure('servers.rest.server1')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServerWithPort);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  it('allows propertyPath for injection metadata', async () => {
+    class RestServerWithPort {
+      constructor(@config({propertyPath: 'port'}) public port: number) {}
+    }
+
+    // Bind configuration
+    ctx
+      .configure('servers.rest.server1')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServerWithPort);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  it('allows propertyPath & fromBinding for injection metadata', async () => {
+    class RestServerWithPort {
+      constructor(
+        @config({propertyPath: 'port', fromBinding: 'restServer'})
+        public port: number,
+      ) {}
+    }
+
+    // Bind configuration
+    ctx
+      .configure('restServer')
+      .toDynamicValue(() => Promise.resolve({port: 3000}));
+
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServerWithPort);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  it('allows propertyPath parameter & fromBinding for injection metadata', async () => {
+    class RestServerWithPort {
+      constructor(
+        @config('port', {fromBinding: 'restServer'})
+        public port: number,
+      ) {}
+    }
+
+    // Bind configuration
+    ctx
+      .configure('restServer')
       .toDynamicValue(() => Promise.resolve({port: 3000}));
 
     // Bind RestServer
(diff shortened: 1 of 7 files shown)


raymondfeng:
@bajtos PTAL.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
```diff
@@ -78,6 +78,47 @@ describe('Context bindings - injecting configuration for bound artifacts', () =>
...
+    // Bind RestServer
+    ctx.bind('servers.rest.server1').toClass(RestServerWithPort);
+
+    // Resolve an instance of RestServer
+    // Expect server1.config to be `{port: 3000}
+    const server1 = await ctx.get<RestServerWithPort>('servers.rest.server1');
+    expect(server1.port).to.eql(3000);
+  });
+
+  it('allows targetBindingKey for injection metadata', async () => {
```
bajtos: I find the name `targetBindingKey` rather long. Is it important to include "binding" in the name? Would `targetKey` be enough to convey the meaning of this option? 

I am not sure. If you feel it's important to include "binding" in the name then I am ok with that. It's better to have a longer option name than a name that's not clear.

packages/context/src/inject-config.ts
```diff
@@ -4,13 +4,27 @@
...
+/**
+ * Injection metadata for `@config.*`
+ */
+export interface ConfigInjectionMetadata extends InjectionMetadata {
+  /**
+   * Path to retrieve the configuration of the target binding
+   */
+  configPath?: string;
+  /**
+   * Target binding key to override the current binding
```
bajtos:
```suggestion
   * Target binding key to override the current binding key
```

packages/context/src/inject-config.ts
```diff
@@ -4,13 +4,27 @@
...
 import {assertTargetType, inject, Injection, InjectionMetadata} from './inject';
 import {ResolutionSession} from './resolution-session';
 import {getDeepProperty, ValueOrPromise} from './value-promise';
 
+/**
+ * Injection metadata for `@config.*`
+ */
+export interface ConfigInjectionMetadata extends InjectionMetadata {
+  /**
+   * Path to retrieve the configuration of the target binding
```
bajtos: I find this description a little bit cryptic. IIUC, this path allows users to receive a (deep) property of the configuration object, is that correct? Let's make that more clear please.

I think the name `configPath` may not be very descriptive either. Can we find a better one, e.g. `propertyPath`?  We are already using the name `propertyPath` in `BindingKey` class, let's be consistent.

https://github.com/strongloop/loopback-next/blob/4c5cacafb0d66c99da3760cbf2b9083e6be0ddb6/packages/context/src/binding-key.ts#L41-L44

packages/context/src/inject-config.ts
bajtos: Let's make this text more descriptive please and explain that this option allows consumers to receive configuration specified for another binding.

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
bajtos: How about `forBinding`?

```ts
@config({forBinding: 'restServer', configPath: 'port'})
```

Or even `for`? Would that be too cryptic?

```ts
@config({for: 'restServer', configPath: 'port'})
```

packages/context/src/inject-config.ts
```diff
@@ -111,7 +140,8 @@ function resolveFromConfig(
   injection: Injection,
   session: ResolutionSession,
 ): ValueOrPromise<unknown> {
-  const bindingKey = getCurrentBindingKey(session);
+  const bindingKey =
+    injection.metadata.targetBindingKey || getCurrentBindingKey(session);
```
bajtos: This logic is repeated in three places. Let's extract a shared helper function.

E.g.

```ts
const bindingKey = getConfigInjectionKey(injection, session);
```

packages/context/src/__tests__/acceptance/binding-config.acceptance.ts
bajtos: Actually, I have a question: is `targetBindingKey` specifying the key used to resolve the config (e.g. as returned by `buildKeyForConfig`) or the key of the binding that the configuration is for? My understanding is that the second option is correct, my comments above are based on that assumption too.

Let's improve the API docs to make this fact crystal clear. I think it would be great to add a code example showing how to use custom target key.

Related question: does it make any sense to allow `@config()` to provide the custom _config binding key_ (as returned for `buildKeyForConfig`) too? Or can is this use-case something that's already solvable by using `@inject` directly? Either way, I guess this is out of scope of this pull request, but it may be worth capturing in our docs.

raymondfeng: I now use `fromBinding`.

raymondfeng:
@bajtos PTAL.

raymondfeng:
@bajtos LGTY now?

packages/context/src/inject-config.ts
```diff
@@ -159,28 +200,28 @@ function resolveAsViewFromConfig(
...
-  const bindingKey = getCurrentBindingKey(session);
+  const bindingKey = getTargetBindingKey(injection, session);
   // Return `undefined` if no current binding is present
   if (!bindingKey) return undefined;
   const view = new ConfigView(
     ctx,
     binding =>
       binding.key === BindingKey.buildKeyForConfig(bindingKey).toString(),
-    injection.metadata.configPath,
+    injection.metadata.propertyPath,
```
bajtos: Should we keep support for `meta.configPath` for backwards compatibility?

```suggestion
    injection.metadata.propertyPath || injection.metadata.configPath,
```

If yes, then please include also a `TODO(semver-major)` comment to remove that compatibility code in the next major version.

packages/context/src/inject-config.ts
```diff
@@ -111,11 +152,11 @@ function resolveFromConfig(
   injection: Injection,
   session: ResolutionSession,
 ): ValueOrPromise<unknown> {
-  const bindingKey = getCurrentBindingKey(session);
+  const bindingKey = getTargetBindingKey(injection, session);
   // Return `undefined` if no current binding is present
   if (!bindingKey) return undefined;
   const meta = injection.metadata;
-  return ctx.getConfigAsValueOrPromise(bindingKey, meta.configPath, {
+  return ctx.getConfigAsValueOrPromise(bindingKey, meta.propertyPath, {
```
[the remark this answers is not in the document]
raymondfeng: Not really, we don't allow `configPath` to be part of the `options` before this PR. Current applications only use `configPath` as positional argument.

packages/context/src/inject-config.ts
```diff
@@ -56,16 +80,20 @@ export function config(configPath?: string, metadata?: InjectionMetadata) {
 export namespace config {
   /**
    * `@inject.getter` decorator to inject a config getter function
-   * @param configPath - Optional property path of the config object
+   * @param propertyPath - Optional property path of the config object
    * @param metadata - Injection metadata
    */
   export const getter = function injectConfigGetter(
```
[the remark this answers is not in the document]
raymondfeng: Added.

packages/context/src/inject-config.ts
```diff
@@ -40,14 +57,21 @@ import {getDeepProperty, ValueOrPromise} from './value-promise';
...
+ * @param propertyPath - Optional property path of the config. If is `''` or not
  * present, the `config` object will be returned.
  * @param metadata - Optional metadata to help the injection
  */
-export function config(configPath?: string, metadata?: InjectionMetadata) {
-  configPath = configPath || '';
+export function config(
+  propertyPath?: string | ConfigInjectionMetadata,
+  metadata?: ConfigInjectionMetadata,
+) {
```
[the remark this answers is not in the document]
raymondfeng: Added.

packages/context/src/inject-config.ts
```diff
@@ -74,16 +102,20 @@ export namespace config {
...
    * changes in configuration
-   * @param configPath - Optional property path of the config object
+   * @param propertyPath - Optional property path of the config object
    * @param metadata - Injection metadata
    */
   export const view = function injectConfigView(
-    configPath?: string,
-    metadata?: InjectionMetadata,
+    propertyPath?: string | ConfigInjectionMetadata,
+    metadata?: ConfigInjectionMetadata,
```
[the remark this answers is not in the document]
raymondfeng: Added

PR #3339, opened 2019-07-11
feat(metrics): add metrics integration with prometheus

raymondfeng:
PoC for https://prometheus.io/ integration

## Checklist


packages/metrics/README.md
```diff
@@ -0,0 +1,53 @@
...
+import {MetricsComponent} from '@loopback/metrics';
+```
+
+In the constructor, add the component to your application:
+
+```ts
+this.component(MetricsComponent);
+```
+
+By default, Metrics route is mounted at `/metrics`. This path can be customized
```
hacksparrow: How about access control?

raymondfeng: It's similar as how we expose openapi specs. Let's worry about that later.

There are a few things to consider:
1. pull vs push - prometheus prefers `pull`
2. use a different rest endpoint (host/port)
3. ACL

raymondfeng:
FYI: I refactored the module to `extensions/metrics` - to be consistent with #3360 

I had to include https://github.com/strongloop/loopback-next/pull/3339/commits/c74b870022c3b7e0e24f2fdb214c4fe8bf360789 so that CI can run.

bajtos:
> I had to include c74b870 so that CI can run.

As I commented in the other PR, can you please open a new PR to make the necessary changes to allow extensions to be hosted in `extensions/` directory?

bajtos:
I am not familiar with Prometheus. What kind of metrics is it collecting? What kind of metrics does it makes sense to provide from a LoopBack application? What can be exposed out of the box and what requires users to provide explicit configuration for?

For example:
- Request latency - how long does it take to handle a request.
  - From the time we receive request headers until we send the response headers (not waiting for the response body to be sent entirely)
  - From the time we receive request body until we send the response headers
  - From the time we receive request headers until the last byte of the response was sent
- DataSource latency - how long does it take to execute a call to a datasource (make a DB query, call a backend web-service)
- Memory usage
- CPU usage
- Event loop latency
- Anything else?

I'd like the documentation for the extension (the README?) to better describe these aspects and educate LB4 users that are new to Prometheus.

raymondfeng:
@bajtos I have updated README to include more information.

raymondfeng:
@bajtos PTAL.

extensions/metrics/README.md
```diff
@@ -0,0 +1,202 @@
...
+npm run demo
+```
+
+### /metrics endpoint
+
+http://localhost:3000/metrics returns metrics in plain text format. It includes
+information for the Node.js process as well as LoopBack method invocations.
+
+```
+# HELP process_cpu_user_seconds_total Total user CPU time spent in seconds.
```
bajtos: This example is a long wall of text hiding the relevant content following the code, can you hide it using `<details>` tag please?

Syntax:

```markdown
<details>
<summary>Example of plain text data</summary>
<pre><code>
# HELP process_cpu_user_seconds_total Total user CPU time spent in seconds.
# TYPE process_cpu_user_seconds_total counter
process_cpu_user_seconds_total 0.132181 1564508354524
...
</code></pre>
</details>
```

Live demo:

<details>
<summary>Example of plain text data</summary>
<pre><code>
# HELP process_cpu_user_seconds_total Total user CPU time spent in seconds.
# TYPE process_cpu_user_seconds_total counter
process_cpu_user_seconds_total 0.132181 1564508354524
...
</code></pre>
</details>

extensions/metrics/src/__examples__/demo.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# This script runs prometheus to scape and visulize metrics collected for the
+# example LoopBack 4 application.
```
bajtos: Have you considered promoting the example into a regular package in `examples/` directory, e.g. `examples/prometheus-metrics`?

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
```diff
@@ -0,0 +1,74 @@
...
+import {MetricsComponent} from '../..';
+import {MetricsBindings} from '../../keys';
+import {MetricsOptions} from '../../types';
+
+const runShell = require('@loopback/build/bin/utils').runShell;
+
+// Only run the test on Travis with Linux
+const verb =
+  process.env.TRAVIS && os.platform() === 'linux' ? describe : describe.skip;
+verb('Metrics (with push gateway)', function() {
```
bajtos: Please use `skipIf` or `skipOnTravis` from `@loopback/testlab`.

Example usage:

https://github.com/strongloop/loopback-next/blob/6ad0bb59c71ea2356cb951da786bdbff246b47e7/packages/repository-tests/src/crud/freeform-properties.suite.ts#L26-L30

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
```diff
@@ -0,0 +1,74 @@
...
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(30000);
```
bajtos: Is this going to increase the duration of `npm test` by another 10-30 seconds? I am concerned that `npm test` is already taking to long to finish to make TDD practical, I am reluctant to make it even worse.

extensions/metrics/src/interceptors/metrics.interceptor.ts
```diff
@@ -0,0 +1,71 @@
...
+    const endSummary = this.summary.startTimer({
+      targetName: invocationCtx.targetName,
+    });
+    try {
+      this.counter.inc();
+      return await next();
+    } finally {
+      endGauge();
+      endHistogram();
+      endSummary();
```
bajtos: This is not the best way how to measure timing. The times you collecting are excluding Express middleware and also interceptors invoked before `MetricsInterceptor`. Maybe this is an acceptable limitation for v0.1 release? If you think so, then please clearly describe this fact in the documentation (README).

extensions/metrics/src/metrics.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A component providing metrics for Prometheus
+ */
+export class MetricsComponent implements Component {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE)
+    private application: Application,
+    @inject(MetricsBindings.CONFIG, {optional: true})
+    options: MetricsOptions = DEFAULT_METRICS_OPTIONS,
```
bajtos: Should we use a `@config`-based binding key here? I find it confusing that certain parts of the app are configured via `app.configure` and other parts via `app.bind`.

```
    @config()
    options: MetricsOptions = DEFAULT_METRICS_OPTIONS,
```

package-lock.json
```diff
@@ -3574,9 +3574,9 @@
       }
     },
     "eslint-utils": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/eslint-utils/-/eslint-utils-1.4.0.tgz",
-      "integrity": "sha512-7ehnzPaP5IIEh1r1tkjuIrxqhNkzUJa9z3R92tLJdZIVdWaczEhr3EbhGtsMrVxi1KeR8qA7Off6SWc5WNQqyQ==",
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/eslint-utils/-/eslint-utils-1.4.2.tgz",
+      "integrity": "sha512-eAZS2sEUMlIeCjBeubdj45dmBHQwPHWyBcT1VSYB7o9x9WRRqKxyUoiXlRjyAwzN7YEzHJlYg0NmzDRWx6GP4Q==",
```
bajtos: Please revert changes in this file. I believe we are using RenovateBot to bump up dependencies in the root package lock?

bajtos:
Would you like to expose README of this new component in https://loopback.io/doc/en/lb4/Using-components.html?

raymondfeng:
@bajtos PTAL.

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
bajtos: > `skipIf(describe, '...', () => {});` does not compile.

That's a known limitation of the current version and/or TypeScript.

Did you try the slightly-longer form I shown in my comment?

```ts
skipIf<[(this: Suite) => void], void>( 
```

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
bajtos: So how long does it usually take to get `prom/pushgateway` docker container up and running? When I run `npm test` on my local machine for the second time, how long delay will be introduced because of waiting for docker?

> We may have to define a nightly build to run the costly integration tests. What do you think?

I feel it's not enough to run these tests nightly. If a pull request breaks one of these tests, then we will discover the problem too late.

Can we use the approach I have in place for running `repository-test` tests against real databases? Here is the gist:
- These tests ARE NOT run as part of `npm test`.
- There are clear instruction how to run the tests locally - see e.g. [MySQL instructions](https://github.com/strongloop/loopback-next/tree/7752b08664cd077272cada93266cd31a4c7fe199/acceptance/repository-mysql#running-the-test-suite).
-  The tests expect external services like databases to be already running and available. This way we pay the cost of starting the services only once, not for every test run.
- There is a single Travis CI job for each test suite (MongoDB, MySQL, etc.), these jobs are executed in parallel with other jobs like `npm test`, code linting, commit linting, etc. Example job config: [.travis.yml#L53-L69](https://github.com/strongloop/loopback-next/blob/7752b08664cd077272cada93266cd31a4c7fe199/.travis.yml#L53-L69) - it does not use Docker, because native MySQL is faster to setup, but Travis CI does support docker.

bajtos:
> > Would you like to expose README of this new component in https://loopback.io/doc/en/lb4/Using-components.html?

> I'm thinking of deferring that and waiting for the component to be a bit more stable.

I have two concerns about that:

- If the component is not publicized, how are people going to learn about it so that we can get feedback from real-world usage?
- What process do you propose to remind us to add the new component to our docs? I am concerned that if we don't do it now, then we will forget to do it later.

If the component is not stable, then let's make that clear in component's README file. Personally, I would add the warning just after the first paragraph (the overview).

extensions/metrics/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+<details>
+<summary>Example of plain text data</summary>
+<pre><code>
+# HELP process_cpu_user_seconds_total Total user CPU time spent in seconds.
+# TYPE process_cpu_user_seconds_total counter
+process_cpu_user_seconds_total 0.132181 1564508354524
+</code></pre>
+
+<pre><code>
```
bajtos: I find the rendered version a bit weird with all that whitespace between blocks of text data.

Screen Shot 2019-09-13 at 15 42 20

Was that your intention?

Personally, I would remove all `</code></pre>` and `<pre><code>` lines from inside and keep only the first and the last one.

extensions/metrics/src/interceptors/metrics.interceptor.ts
bajtos: Fair enough, please add a code comment to make the limitation obvious also to people that are don't have deep knowledge of the way how we are integrating with Express, and also add a note to README too.

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
bajtos: Can you write a mock-up pushgateway that will call our metrics endpoint the same way as the real gateway does? That way we can verify push functionality from `extensions/metrics` tests.

Then we can add a new package to `acceptance` directory, e.g. `acceptance/push-metrics`, where we will use a real push gateway running in a docker container, to ensure our push implementation works with real gateways too and catch any discrepancies between our mock gateway and real gateways.

extensions/metrics/README.md
raymondfeng: I had to apply `<pre></pre>` around all blocks, otherwise `#` is rendered as headers. It seems that I can remove `<code>` and `</code>` to reduce number of empty lines.

raymondfeng:
@bajtos PTAL

extensions/metrics/README.md
```diff
@@ -0,0 +1,253 @@
+# @loopback/extension-metrics
+
+This module contains a component that reports metrics of Node.js, LoopBack
+framework, and your application to [Prometheus](https://prometheus.io/).
+
+## Stability - Experimental
```
bajtos: Let's use colon `:` to delimit the field name "Stability" from the value "Experimental. 

```suggestion
## Stability: Experimental
```

See e.g. Node.js docs (they use `:` after "Stability" and then `-` to delimit stability index from the description).

Screen Shot 2019-09-19 at 16 36 10

bajtos: It makes we wonder how to better emphasize the experimental status. How about using all-caps and adding an emoji, e.g. 🛑 🚨 or ⚠️?

```suggestion
## Stability: ⚠️ EXPERIMENTAL ⚠️ 
```

extensions/metrics/README.md
```diff
@@ -1,7 +1,14 @@
...
-[Prometheus](https://prometheus.io/).
+This module contains a component that reports metrics of Node.js, LoopBack
+framework, and your application to [Prometheus](https://prometheus.io/).
+
+## Stability - Experimental
+
+> Experimental packages provide early access to advanced or experimental
+> functionality to get community feedback. Such modules are published to npm
+> using `0.x.y` versions. Their APIs may subject to breaking changes in future
+> releases.
```
bajtos: > Their APIs may subject to breaking changes in future releases.

I am not sure if this is correct in English, it think the following is better:

```
Their APIs may be subject to (...)
```

In addition to breaking API changes, I'd like to give us the option to introduce breaking changes in behavior too.

Proposed text:

```
Their APIs and functionality may be subject to breaking changes in future releases.
```

extensions/metrics/README.md
```diff
@@ -80,86 +88,85 @@ information for the Node.js process as well as LoopBack method invocations.
...
-<pre><code>
+<pre>
 # HELP process_cpu_user_seconds_total Total user CPU time spent in seconds.
 # TYPE process_cpu_user_seconds_total counter
 process_cpu_user_seconds_total 0.132181 1564508354524
-</code></pre>
-
-<pre><code>
+</pre>
+<pre>
```
bajtos: This is still introducing extra empty space between code blocks.

Screen Shot 2019-09-19 at 16 40 00

Is it your intention to visually delimit these blocks?

extensions/metrics/src/interceptors/metrics.interceptor.ts
```diff
@@ -14,6 +14,10 @@ import {
 } from '@loopback/context';
 import {Counter, Gauge, Histogram, Summary} from 'prom-client';
 
+/**
+ * This interceptor captures metrics for method invocations. Please note that
+ * metrics can be collected by other places, such as sequence of actions.
```
bajtos: I am proposing the following text:

```
This interceptor captures metrics for method invocations,
excluding sequence actions and middleware executed before
a method is invoked. Please collect metrics at other places
if you want to cover more than just method invocation.
```

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
bajtos: Can we remove this 30 second timeout now, or at least reduce it to something like 5-10s?

Also I see that you introduce a mock push gateway, which is great! But how can we be sure that it's accurately simulating the behavior of a real gateway? Shouldn't we have an acceptance tests using the docker-based gateway as I proposed in my comment above?

extensions/metrics/src/__tests__/acceptance/mock-pushgateway.ts
```diff
@@ -0,0 +1,43 @@
...
+    app.use(express.text({type: '*/*'}));
+    app.get('/metrics', (req, res) => {
+      res.send(this.metrics.join('\n'));
+    });
+
+    app.post('/metrics/job/:jobName', (req, res) => {
+      this.metrics.push(`job="${req.params.jobName}"\n${req.body}`);
+      res.send('\n');
+    });
+    this.server = app.listen(9091);
```
bajtos: We should not use hard-coded port numbers, they can fail when tests are run in parallel or if the developer running the tests locally happen to have another process already listening on the same port.

You can use `app.listen(0)` or perhaps our `HttpServer` class.

```ts
import {HttpServer} from '@loopback/http-server';

this.server = new HttpServer(app, {/* no options are needed */});
await this.server.start();
```

bajtos: Unless the port `9091` is required by Prometheus and cannot be customized?

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
raymondfeng: I think the mock-up push gateway gives us enough confidence as we just to have make sure this component is pushing metrics to the gateway (the correctness should have been covered by the prom-client).

raymondfeng:
@bajtos PTAL

extensions/metrics/src/__tests__/acceptance/metrics-push.acceptance.ts
```diff
@@ -0,0 +1,62 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {RestApplication, RestServerConfig} from '@loopback/rest';
+import {givenHttpServerConfig, supertest} from '@loopback/testlab';
+import {AddressInfo} from 'net';
+import {promisify} from 'util';
+import {MetricsComponent} from '../..';
+import {MetricsBindings} from '../../keys';
+import {MetricsOptions} from '../../types';
```
bajtos: Can this be simplified as follows?

```ts
import {
 MetricsComponent,
 MetricsBindings,
 MetricsOptions,
} from '../..';
```

extensions/metrics/src/observers/pushgateway.observer.ts
```diff
@@ -0,0 +1,35 @@
...
+// Node module: @loopback/extension-metrics
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {config, LifeCycleObserver} from '@loopback/core';
+import {Pushgateway} from 'prom-client';
+import {MetricsBindings} from '../keys';
+import {DEFAULT_METRICS_OPTIONS, MetricsOptions} from '../types';
+
+export class PushGatewayObserver implements LifeCycleObserver {
```
bajtos: I find this name weird. To me, it implies this class is observing a push gateway. Please find a better one, also rename the file accordingly.

It's a good practice to use names that express the intent (what is this trying to achieve) instead of implementation details (it's implemented as a LifeCycleObserver).

IIUC, this code is setting up periodic reporting of metrics to a push gateway. So maybe `MetricPushReporter` would be a good name?

```suggestion
export class MetricPushReporter implements LifeCycleObserver {
```

hacksparrow:
As a POC, this looks good.

However, I am not sure if hardwiring a core extension to a particular service is a good idea. Ideally, its interface should be an adapter - users should be able to use Prometheus alternatives, if they want to.

extensions/metrics/src/interceptors/metrics.interceptor.ts
```diff
@@ -0,0 +1,77 @@
...
+      targetName: invocationCtx.targetName,
+    });
+    const endHistogram = this.histogram.startTimer({
+      targetName: invocationCtx.targetName,
+    });
+    const endSummary = this.summary.startTimer({
+      targetName: invocationCtx.targetName,
+    });
+    try {
+      this.counter.inc();
```
jannyHou: A question for the counter: I run the demo and noticed the method invocation # is 105:
```
# HELP loopback_invocation_total method invocation counts
# TYPE loopback_invocation_total counter
loopback_invocation_total 105
```
The demo app doesn't have any controller/endpoints, so I am wondering... what are the methods being invoked?

raymondfeng: Good question. We have a [built-in controller](https://github.com/strongloop/loopback-next/blob/prometheus/extensions/metrics/src/controllers/metrics.controller.ts) in `extensions/metrics`.

So whenever the `/metrics` is scraped, our metrics interceptor is triggered.

jannyHou: ah, that makes sense 👍

raymondfeng:
> I have a general question for prometheus: is it also aimed to monitor particular endpoints or it's more for monitoring the health of an app?

We already have `@loopback/extension-health` for health checks. The metrics extension is to enable metrics reporting for prometheus. The metrics includes Node.js runtime, LoopBack framework code (TBA), and application logic.

Issue #2133, opened 2018-12-02
Documentation of loopback 4 Model is incorrect


opepermans:
# Description/Steps to reproduce

# Link to reproduction sandbox

# Expected result

# Additional information

emonddr:
We need to groom this to determine if we want an lb4 version of the https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html .

dhmlau:
Instead of finding which content in https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html we can reuse in LB4, I think we can think the other way round to beef up the docs for [Model Decorator](https://loopback.io/doc/en/lb4/Model.html#model-decorator) and [Property Decorator](https://loopback.io/doc/en/lb4/Model.html#property-decorator), and see if there's anything applicable in the LB3 docs we can reuse so that we don't need to write it from scratch. 

- [Model decorator](https://loopback.io/doc/en/lb4/Model.html#model-decorator)
  - what's expected in `properties`?  Is it https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#properties? 
  - what's expected in `settings`?  Is it everything in https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options? 
- [Property Decorator](https://loopback.io/doc/en/lb4/Model.html#property-decorator)
  - what are the valid values? would it be what's in the table for [general properties](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#general-property-properties)?

b-admike:
Based on the discussion in this thread, I've compiled the following criteria for this issue.

# Acceptance Critera
- [ ] Figure out what can be passed into the model decorator (top-level) and how that affects the LoopBack 4 application. Because we are using legacy juggler bridge, we need to find out if the mapping of top-level keys works OOTB. For example, does passing in `@model({options:{...}})` work similar to how it does in LB3? The main keys to examine are `properties`, `settings`, `options`, `relations` because they're used often, but it'd be good to check all with a test app. Explain this in docs for [Model Decorator](https://loopback.io/doc/en/lb4/Model.html#model-decorator). Use https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html as the base to compare to and document differences.
[...]

Issue #2142, opened 2018-12-10
Confusing @model() syntax


bajtos:
In LB3, users could specify model settings at two levels: as a root property or inside `options` property.

See https://stackoverflow.com/q/53307168/69868 for an example of LB3 syntax applied in an LB4 project:

```ts
@model({
  settings: {strict: false},
  name: 'client',
  plural: 'clients',
  options: {
    mongodb: {
      collection: 'clients',
    },
  },
})
export class Client extends Entity {
  // ...
}
```

I am proposing to make two changes in LB4 to help users coming from LB3:

1. Recognize `options` the same way as `settings`. In the example above, `mongodb` settings are not picked by LB4 now. With the proposed change in place, LB4 will set `collection` to `clients` as expected. Alternatively, tell the user setting `options` that they are trying to set an unsupported model-definition property. This can be done at compiler level too.

2. Allow settings to provided as top-level properties, for example:

    ```ts
    @model({
      name: 'client',
      strict: false,
      mongodb: {
        collection: 'clients',
      },
    })
    export class Client extends Entity {
      // ...
    }
    ```

samarpanB:
@bajtos If its fine, I can try help with this feature implementation. Let me know if that's fine ? I have not worked with LB3, so I may need some guidance on exact requirements.

bajtos:
@samarpanB of course it's fine to contribute! Sorry for the delay, I am often overwhelmed by GitHub notifications and I am still learning how to organize them so that I can focus on more important matters first. Shout out to https://octobox.io which helps me a lot!

> I have not worked with LB3, so I may need some guidance on exact requirements.

Here is the code that's "normalizing" model configuration in LoopBack 3.x:

https://github.com/strongloop/loopback/blob/9b1d4885ab869d977bc7d464aa861b4f88691862/lib/registry.js#L207-L227

```js
  // Settings
  var excludedProperties = {
    base: true,
    'super': true,
    relations: true,
    acls: true,
    dataSource: true,
  };
  if (typeof config.options === 'object' && config.options !== null) {
    for (var p in config.options) {
      if (!(p in excludedProperties)) {
        settings[p] = config.options[p];
      } else {
        g.warn('Property `%s` cannot be reconfigured for `%s`',
          p, modelName);
      }
    }
  } else if (config.options != null) {
    g.warn('The options property of `%s` configuration ' +
      'must be an object', modelName);
  }
```

[...]

PR #3354, opened 2019-07-14
docs: update model decorator

agnes512:
Resolves https://github.com/strongloop/loopback-next/issues/2133

Update inconsistent docs for `@model` and `@property` decorators part. 
By testing out the code in `loopback-datasource-juggler/lib/datasource.js`,  `loopback-datasource-juggler/lib/model-builder.js`, I made following changes: 

#### `@model` decorator: 
( I copied the available part of [the table](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#top-level-properties) from lb3 site and made a bit change on the content. Because I think checking sites back and forth is kind of annoying.)
- indicates that lb4 now only supports `settings` entries in model decorator. And `top-level-properties` in lb3 now are passed in `settings`. 
- indicates that properties are defined via `@property` decorator.
- lists features like `options`, `acls` are not available in lb4 yet. ~Not sure why we don't support `options`. Will open an issue for this part.~ We already have that [issue opened](https://github.com/strongloop/loopback-next/issues/2142).
- indicates that lb4 defines relations by relation decorators such as `@hasMany`. 

#### `@property` decorator:
Since this part is the same as lb3, I only made slight change. 

## Checklist

### DIFF
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -157,30 +157,28 @@ class MyFlexibleModel extends Entity {
 ### Model Decorator
 
 The model decorator can be used without any additional parameters, or can be
-passed in a
+passed in a ModelDefinitionSyntax:
 
 <!-- should be replaced with a lb4 example when possible -->
-
-[ModelDefinitionSyntax](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html)
-object which follows the general format provided in LoopBack 3:
+<!-- according to https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L283 and the legacy juggler file-->
 
 ```ts
 @model({
   name: 'Category',
-  properties: {
-    // define properties here.
-  },
   settings: {
     // etc...
   },
+  // define properties by @property decorator below
 })
 class Category extends Entity {
   // etc...
+  @property({type: 'number'})
+  categoryId: number;
 }
 ```
 
-However, the model decorator already knows the name of your model class, so you
-can omit it.
+The model decorator already knows the name of your model class, so you can omit
+it.
 
 ```ts
 @model()
@@ -190,9 +188,197 @@ class Product extends Entity {
 }
 ```
 
-Additionally, the model decorator is able to build the properties object through
-the information passed in or inferred by the property decorators, so the
-properties key value pair can also be omitted.
+However, the model decorator in LoopBack 4 is not exactly the same as what it is
+in LoopBack 3. For example, in
+[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
+pass in a model definition syntax in the model decorator, such as properties,
+options, relation etc. But not all these entries are available in lb4 model
+decorator:
+
+<!-- Please modify this part when options is available -->
+
+NOTICE: in LoopBack 4 we only support `settings` in the ModelDefinitionSyntax
+for now. Those `top-level properties` in lb3 now are passed in `settings`.
+
+- `properties` now are defined in `@property` decorator (see below for more
+  information).
+- [`options`](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options)
+  in lb3 doesn't have the mapping feature in LB4 yet. (see
+  [issue #2142](https://github.com/strongloop/loopback-next/issues/2142) for
+  further discussion.)
+
+As for entries in `settings`, LoopBack 4 supports these built-in entries for
+now:
+
+#### Supported Entries of Settings
+
+<!-- These entries might need to update once we've made some changes:
+  - description [in lb3 we support array or string, but here we documented it as string only]
+  these two don't seem work in lb4. Moved them to unsupported table.
+-->
+
+  <table>
+  <thead>
+  <tr>
+    <th width="160">Property</th>
+    <th width="100">Type</th>
+    <th>Default</th>
+    <th>Description</th>
+  </tr>
+  </thead>
+
+  <tbody>
+
+  <tr>
+  <td><code>description</code></td>
+  <td>String</td>
+  <td>None</td>
+  <td>
+    Optional description of the model. We only support string type for now. (see <a href="https://github.com/strongloop/loopback-next/issues/3428">issue #3428</a> for more discussion.)
+  </td></tr>
+
+  <tr>
+  <td><code>forceId</code></td>
+  <td>Boolean</td>
+  <td><code>true</code></td>
+  <td>
+    If true, prevents clients from setting the auto-generated ID value manually.
+  </td>
+  </tr>
+
+  <tr>
+    <td><code>strict</code></td>
+    <td>Boolean or String</td>
+    <td><code>true</code>.<br/></td>
+    <td>
+      In LB4, the default for this entry is set to be <code>true</code>.<br/>
+      Specifies whether the model accepts only predefined properties or not. One of:
+      <ul>
+      <li><code>true</code>: Only properties defined in the model are accepted. Use if you want to ensure the model accepts only predefined properties.
+      If you try to save a model instance with properties that are not predefined, LoopBack throws a ValidationError.
+      </li>
+      <li><code>false</code>: The model is an open model and accepts all properties, including ones not predefined in the model.
+        This mode is useful to store free-form JSON data to a schema-less database such as MongoDB.
+      </li>
+      <li><code>"filter"</code>: Only properties defined in the model are accepted.
+      If you load or save a model instance with properties that are not predefined, LoopBack will ignore them. This is particularly useful when dealing with old data that you wish to lose without a migration script.
+      </li>
+      </ul>
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>idInjection</code></td>
+    <td>Boolean</td>
+    <td><code>true</code></td>
+    <td>
+      Whether to automatically add an <code>id</code> property to the model:
+    <ul>
+    <li><code>true</code>: <code>id</code> property is added to the model automatically. This is the default.</li>
+    <li><code>false</code>: <code>id</code> property is not added to the model</li>
+    </ul>
+    See <a href="#id-properties">ID properties</a> for more information.  The <code>idInjection</code> property in <code>options</code> (if present) takes precedence.
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>name</code></td>
+    <td>String</td>
+    <td>None</td>
+    <td>Name of the model.</td>
+  </tr>
+
+  <tr>
+    <td><code>scopes</code></td>
+    <td>Object</td>
+    <td>N/A</td>
+    <td>See <a href="https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#scopes">Scopes</a> in lb3 docs.</td>
+  </tr>
+
+  </tbody>
+
+  </table>
+
+#### Unsupported Entries of Settings
+
+<!-- Please update the OPTIONS, ACLS field when they are available -->
+
+<table>
+  <thead>
+    <tr>
+      <th width="160">Property</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+
+  <tbody>
+
+  <tr>
+    <td><code>acls</code></td>
+    <td>
+      (TBD)
+    </td>
+  </tr>
+
+  <tr>
+  <td><code>base</code></td>
+  <td>This entry is no longer being used. This is done by the typical Js/Tsc classes inheritance way in LB4:
+  <pre><code>@model() class MySuperModel extends MyBaseModel {...}</code>
+  </pre></td>
+  </tr>
+
+  <tr>
+  <td><code>excludeBaseProperties</code></td>
+  <td>(TBD)</td>
+  </tr>
+
+  <tr>
+  <td><code>http</code></td>
+  <td> This entry affects HTTP configuration in LB3. Since in LB4 http configurations are inferred from controller members and the rest server, this field is not applicable.</td>
+  </tr>
+
+  <tr>
+    <td><code>options</code></td>
+    <td>
+      (TBD) see <a href="https://github.com/strongloop/loopback-next/issues/2142">issue #2142</a> for further discussion.
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>plural</code></td>
+    <td>This entry is part of HTTP configuration in LB3. So it's not applicable for the same reason as <code>http</code> above.</td>
+  </tr>
+
+  <tr>
+    <td><code>properties</code></td>
+    <td>This entry is no longer being used as we introduced <code>@property</code> decorator in LB4. See <code>@property</code> decorator below to discover moer about how to define properties for your models.</td>
+  </tr>
+
+  <tr>
+    <td><code>relations</code></td>
+    <td>
+      With the introduction of <a href="https://loopback.io/doc/en/lb4/Repositories.html">repositories</a>, now <code>relations</code> are defined by <code>relations decorators</code> in LB4.
+      See <a href="https://loopback.io/doc/en/lb4/Relations.html">Relations</a> for more details.
+    </td>
+  </tr>
+  <tr>
+    <td><code>remoting.<br/>normalizeHttpPath</code></td>
+    <td>
+    This entry is part of HTTP configuration in LB3. So it's not applicable for the same reason as <code>http</code> above.
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>replaceOnPUT</code></td>
+    <td>This entry is no longer supported as LB4 controllers scaffolded by LB4 controller, PUT is always calling replaceById. Users are free to change the generated code to call <code>patchById</code> if needed.</td>
+  </tr>
+  </tbody>
+</table>
+
+To discover more about `Model Decorator` in LoopBack 4, please check
+[legacy-juggler-bridge file](https://github.com/strongloop/loopback-next/blob/2fa5df67181cdcd23a5dce90c9c640fe75943cb8/packages/repository/src/repositories/legacy-juggler-bridge.ts)
+and
+[model-builder file](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js).
 
 #### Hidden properties
 
@@ -221,6 +407,8 @@ class MyUserModel extends Entity {
 
 ### Property Decorator
 
+<!-- Property decorator can reuse lb3 docs -->
+
 The property decorator takes in the same arguments used in LoopBack 3 for
 individual property entries:
 
@@ -233,6 +421,12 @@ class Product extends Entity {
     type: 'string',
   })
   public name: string;
+
+  @property({
+    type: 'number',
+    id: true,
+  })
+  id: number;
 }
 ```
 
--- a/packages/repository/DEVELOPING.md
+++ b/packages/repository/DEVELOPING.md
@@ -1,0 +1,37 @@
+## LB4 vs. LB3
+
+### How LoopBack 4 creates/handles models. [2019 July 24]
+
+In `lb4`, we introduce the concept of
+[`repository`](https://loopback.io/doc/en/lb4/Repositories.html). In order to
+implement this new feature `repository`, we are using legacy juggler bridge as a
+(temporary) solution. Each model has two **parallel classes**:
+
+- The class defined by the app developers using decorators and
+  `class MyModel extends Entity` syntax.
+- The "implementation" class created by legacy-juggler-bridge, this class is
+  inheriting from juggler's Model `base class`.
+
+If you take a deep look into it, you would find that LB4 model handles some data
+before it hands the model to Juggler. That's why `lb4` has some differences
+compared to `lb3` when generating models.
+
+For example, you might notice that some entries in `@model` decorator are not
+available/supported by lb4 even they both run the same lines of code in Juggler.
+For instance, the default values for the `strict` entry in `@model decorator`
+are different:
+
+|           | [in LB3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#top-level-properties)                                                     | [in LB4](https://loopback.io/doc/en/lb4/Model.html#model-decorator)                                                                                                          |
+| --------- | ------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `default` | `false`                                                                                                                                           | `true`                                                                                                                                                                       |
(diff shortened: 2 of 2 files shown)


docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+lb3. For example, in
+[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
+pass in a model definition syntax in model decorator, such as properties,
+options, relation etc. But not all these entries are available in lb4 model
+decorator:
+
+In lb4, we only support `properties` and `setting` in the ModelDefinitionSyntax
+for now. Those `top-level properties` in lb3 now are passed in `settings`.
+[Options](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options)
+in lb3 doesn't have the mapping feature in lb4 yet.
```
bajtos: Can you please mention the issue https://github.com/strongloop/loopback-next/issues/2142 here? In that issue, we are discussing how to improve the current `@model` syntax to make it easier to use.

docs/site/Model.md
```diff
@@ -221,6 +408,8 @@ class MyUserModel extends Entity {
 
 ### Property Decorator
 
+<!-- Property decorator can reuse lb3 docs -->
```
bajtos: Yes please!

Ideally, we should copy information from https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#properties.

I am not sure if that's in scope of this pull request though.

I see that we have the following text after the example code snippet, thus I think we are good for now.

> The complete list of valid attributes for property definitions can be found in LoopBack 3's [Model definition section](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#properties).

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+    <th width="160">Property</th>
+    <th width="100">Type</th>
+    <th>Default</th>
+    <th>Description</th>
+  </tr>
+  </thead>
+
+  <tbody>
+  <tr>
+  <td><code>base</code></td>
```
bajtos: AFAIK, `base` is not supported in LB4. We use JavaScript class inheritance instead.

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+  </tr>
+
+  <tr>
+  <td><code>description</code></td>
+  <td>String or Array</td>
+  <td>None</td>
+  <td>
+    Optional description of the model.<br/><br/>
+    You can split long descriptions into arrays of strings (lines) to keep line lengths manageable; for example:
+  <pre style="font-size: 75%;">[ "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
```
bajtos: I am not sure if we support description yet, for example whether it's copied to OpenAPI/JSON schema emitted for models. Having said that, I think it's a good idea to document that field :+1:

I have mixed feeling about the "arrays of string" flavor. I have introduced it to make it easier to create long descriptions in JSON files, however it was not welcome by groups working on Swagger and OpenAPI specs.

Let's remove this form from our docs for now. We can revisit this decision later, if there is any user demand.

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+
+</pre>
+    </td></tr>
+
+  <tr>
+  <td><code>excludeBaseProperties</code></td>
+  <td>Array</td>
+  <td><code>['id', 'password']</code></td>
+  <td>
+  Excludes the given list of properties from the base model from being visible. Use this instead of the approach documented under 'Exclude properties from base model' section below.
```
bajtos: Hmm, it makes me wonder how is this going to work in LB4. Can we add a note that this feature hasn't been tested in LB4 and may not work?

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+  <td><code>forceId</code></td>
+  <td>Boolean</td>
+  <td><code>true</code></td>
+  <td>
+    If true, prevents clients from setting the auto-generated ID value manually.
+  </td>
+  </tr>
+
+  <tr>
+  <td><code>http</code></td>
```
bajtos: `http` is not supported in LB4. These settings are used by strong-remoting (REST API layer) in LB3, LB4 has a very different approach for REST APIs.

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+  <td>String</td>
+  <td>None</td>
+  <td>Customized HTTP path for REST endpoints of this model.</td>
+  </tr>
+
+  <tr>
+    <td><code>strict</code></td>
+    <td>Boolean or String</td>
+    <td><code>false</code>.<br/>If the data source is backed by a relational database, then default is  <code>true</code>.</td>
```
bajtos: In LB4, `strict` is set to `true` by default for all databases, see https://github.com/strongloop/loopback-next/blame/4a4e182256c080cc094ed23b1b4f2a5e7b9946e0/docs/site/Model.md#L134

> Models are defined primarily by their TypeScript class. By default, classes
forbid additional properties that are not specified in the type definition. The
persistence layer respects this constraint and configures underlying
PersistedModel classes to enforce `strict` mode.
>
> To create a model that allows both well-defined but also arbitrary extra
properties, you need to disable `strict` mode in model settings through the CLI
and tell TypeScript to allow arbitrary additional properties to be set on model
instances.
>
> (code snippet follows)

I am proposing to:
- Add a new header before the paragraph I quoted above, e.g. `### Strict mode`. This will create an anchor we can use in href links.
- Here in the table, add a link to `#strict-mode` so that people can easily find this additional information

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+
+  <tr>
+    <td><code>name</code></td>
+    <td>String</td>
+    <td>None</td>
+    <td>Name of the model.</td>
+  </tr>
+
+  <tr>
+    <td><code>plural</code></td>
```
bajtos: This is used by strong-remoting to figure out URL path, e.g. model `Product` is exposed at `/products`.

This setting does not apply to LB4.

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+    <td>
+      Even though properties can be defined in model decorator, we don't recommend this way since it's inconsistent. <b>Please do define <code>properties</code> by <code>@property Decorator</code></b>.
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>scopes</code></td>
+    <td>Object</td>
+    <td>N/A</td>
+    <td>See <a href="https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#scopes">Scopes</a> in lb3 docs.</td>
```
bajtos: I think this property should be move the "Not Available" section, since it's not supported by LB4 (yet).

WDYT?

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+  </tr>
+  <tr>
+    <td><code>remoting.<br/>normalizeHttpPath</code></td>
+    <td>
+    (TBD)
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>replaceOnPUT</code></td>
```
bajtos: This is used by REST API layer in LB3 to determine whether `PUT` verb should call `patchById` or `replaceById`, it was introduced for backwards compatibility with LB2.

In LB4 controllers scaffolded by `lb4 controller`, `PUT` is always calling `replaceById`. Users are free to change the generated code to call `patchById` if they prefer.

docs/site/Model.md
agnes512: In [model-builder file](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L157), it sets up the base here when I run the debug tool. That's why I think it's available. Could you explain more about this part?

docs/site/Model.md
agnes512: This parts comes from [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L263). I passed the value in, and it got handled. That's why I noted as available.

docs/site/Model.md
agnes512: I am not sure if `http` here is the same as REST APIs'. This value is handled [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L132).

docs/site/Model.md
agnes512: oops I got the `default: false` from [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L150).. It makes me wonder if I used the wrong module to check out decorators :joy:

docs/site/Model.md
agnes512: Same as above: [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L129)

docs/site/Model.md
agnes512: I agree.  But I think it kinda 'works' in model decorator as I tested it out before. I don't know if that's because we handle `properties`[here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L126). It would still get overwritten by property decorators anyways. So it makes more sense for me to remove it from the available list.

docs/site/Model.md
agnes512: I guess it's more proper to have this part:
> In LB4 controllers scaffolded by `lb4 controller`, `PUT` is always calling `replaceById`. Users are free to change the generated code to call `patchById` if they prefer.

instead of TBD for `replacePUT` right?

docs/site/Model.md
agnes512: Thanks for these comments!

docs/site/Model.md
jannyHou: > for example whether it's copied to OpenAPI/JSON schema emitted for models.

Not really. Good to know the model description could map to the schema's description....

docs/site/Model.md
jannyHou: @agnes512 In LB4 the http configurations are inferred from controller members and the rest server.

docs/site/Model.md
```diff
@@ -190,9 +188,198 @@ class Product extends Entity {
...
+In lb4, we only support `properties` and `setting` in the ModelDefinitionSyntax
+for now. Those `top-level properties` in lb3 now are passed in `settings`.
+[Options](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options)
+in lb3 doesn't have the mapping feature in lb4 yet.
+
+As for entries in `settings`, lb4 supports these built-in entries for now:
+
+#### Available Entries of Settings
+
+  <table>
```
jannyHou: *No need to change the current doc* 
Just in case if you haven't tried the Markdown syntax table: https://www.tablesgenerator.com/markdown_tables

It's simpler to use then the html tags :)

docs/site/Model.md
bajtos: Sure. You are right that juggler's model-builder does support `base` setting.

Here is how model inheritance should be used in LB4:

```ts
@model()
class BaseUser extends Entity {
  // properties and methods used by all models
}

@model()
class Customer extends BaseUser {
  // properties specific to customers
}

@model()
class Administrator extends BaseUser {
  // properties specific to site admins
}
```

I would expect that under the hood, the code converting LB4 model to juggler model definition should take care of model inheritance and configure `base` setting as needed. See [`definePersistedModel`](https://github.com/strongloop/loopback-next/blob/2fa5df67181cdcd23a5dce90c9c640fe75943cb8/packages/repository/src/repositories/legacy-juggler-bridge.ts#L135-L187). (I don't see any such code in that method though.)

To be honest, I am not sure if we ever tested model inheritance in LB4, it may not even work properly.

I also don't know what happens when a LB4 model explicitly configures the `base` setting, this is where a unit/integration test would be very useful.

docs/site/Model.md
bajtos: > Let's remove this form from our docs for now. We can revisit this decision later, if there is any user demand.

Just to clarify - I am proposing to remove support for `description` set to an array of strings. The `description` field itself should stay supported & documented.

> > for example whether it's copied to OpenAPI/JSON schema emitted for models.
>
> Not really.

Let's open a new GH issue to describe that feature, then we can mention the issue here in the docs.

docs/site/Model.md
bajtos: I guess this is a part of a bigger question - whether model inheritance works in LB4 and how should users define models inheriting from other models. See the discussion in https://github.com/strongloop/loopback-next/pull/3354#discussion_r306318428

docs/site/Model.md
bajtos: AFAICT, juggler is reading only `pathName` from the `http` field, it's used to set `http` property on the model class elsewhere: https://github.com/strongloop/loopback-datasource-juggler/blob/6b83b6f4d1a4941917ffbaf8cebf476e40a53f83/lib/model-builder.js#L226-L230

The REST API layer in LB4 does not use the `http` field from juggler model classes, it does not even have access to them!

Let's move `http` to "unsupported" list.

docs/site/Model.md
bajtos: Yes, juggler sets `strict` to false by default. However, LB4 sets `strict` to true by default, before  it hands model settings to juggler.

See here:

https://github.com/strongloop/loopback-next/blob/4b435f3df4eb6ecb76647e0ac1fef6d3ff94e96a/packages/repository/src/repositories/legacy-juggler-bridge.ts#L157-L168

docs/site/Model.md
bajtos: See my https://github.com/strongloop/loopback-next/pull/3354#discussion_r306321047

docs/site/Model.md
bajtos: Yes, let's use that text, but also move `replaceOnPUT ` to the list of "unsupported" settings.

I.e. say that `replaceOnPUT ` is not supported and explain why & how to control what happens on PUT the LB4 way.

docs/site/Model.md
bajtos: I have mixed experience with Markdown tables. They force you to put each table row on a single line, which is rather impractical when the table contains long text (e.g. setting description).

bajtos:
@agnes512 great discussion! I think in general, it's important to realize that in LB4, each model has two parallel classes:
- The class defined by the app developers using decorators and `class MyModel extends Entity` syntax.
- The "implementation" class created by legacy-juggler-bridge, this class is inheriting from juggler's `Model` base class.

Because of this design, not all settings offered by juggler will actually work in LoopBack 4.

You may ask why we have this complexity? The legacy juggler bridge is a (temporary) solution allowing us to implement `Repository` interfaces using existing juggler & connector codebases. In the future, we would like to roll out a LB4-native implementation of `Repository` interfaces, this will call connectors directly and won't depend on juggler at all. However, it requires a significant amount of work, thus we decided to use the legacy juggler bridge for now, despite all the complexity it introduces.

BTW I am pretty sure you are not the last person to be caught by surprise about this design, it would be great if you could document what you have learned about this problem domain. For example, we can add `DEVELOPING.md` file to [packages/repository](https://github.com/strongloop/loopback-next/tree/master/packages/repository).

docs/site/Model.md
agnes512: Yeah I use <html> to set the length of each column.

agnes512:
related issues:
- `discussion` field: [link](https://github.com/strongloop/loopback-next/issues/3428)
- `options` field: [link](https://github.com/strongloop/loopback-next/issues/2142)

docs/site/Model.md
```diff
@@ -190,9 +188,188 @@ class Product extends Entity {
...
+  </thead>
+
+  <tbody>
+
+  <tr>
+  <td><code>description</code></td>
+  <td>String</td>
+  <td>None</td>
+  <td>
+    Optional description of the model. We only support string type for now. (see <a href="https://github.com/strongloop/loopback-next/issues/3428">issue #3428</a> for more discussion.)
```
bajtos: I thought that repository-json-schema package is not copying the description field to the emitted schema, but apparently it does so ✌️ 

https://github.com/strongloop/loopback-next/blob/27a7e15d87ab14109dbd03e0556fdda31cdb0a0b/packages/repository-json-schema/src/build-schema.ts#L329-L331

docs/site/Model.md
```diff
@@ -190,9 +188,188 @@ class Product extends Entity {
...
+  <tr>
+    <td><code>acls</code></td>
+    <td>
+      (TBD)
+    </td>
+  </tr>
+
+  <tr>
+  <td><code>base</code></td>
+  <td>(TBD))</td>
```
bajtos: Should we explain that users are supposed to use JS/TS class inheritance instead?

docs/site/Model.md
```diff
@@ -190,9 +188,188 @@ class Product extends Entity {
...
+  <tr>
+    <td><code>options</code></td>
+    <td>
+      (TBD)
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>plural</code></td>
+    <td>(TBD)</td>
```
bajtos: Let's explain that `plural` field is affecting HTTP configuration in LB3 and since in LB4 we are inferring http configuration from controller members, this field is not applicable in LB4. 

I am not sure if we will ever support this field in LB4, the text "TBD" seems misleading to me.

docs/site/Model.md
```diff
@@ -190,9 +188,188 @@ class Product extends Entity {
...
+    <td><code>relations</code></td>
+    <td>
+      In lb4, with the introduction of <a href="https://loopback.io/doc/en/lb4/Repositories.html">repositories</a>, now <code>relations</code> are defined by <code>relations decorators</code>.
+      See <a href="https://loopback.io/doc/en/lb4/Relations.html">Relations</a> for more details.
+    </td>
+  </tr>
+  <tr>
+    <td><code>remoting.<br/>normalizeHttpPath</code></td>
+    <td>
+    (TBD)
```
bajtos: This is again related to HTTP configuration.

I am not sure if we will ever support this field in LB4, the text "TBD" seems misleading to me.

packages/repository/DEVELOPING.md
```diff
@@ -0,0 +1,37 @@
...
+implement this new feature `repository`, we are using legacy juggler bridge as a
+(temporary) solution. Each model has two **parallel classes**:
+
+- The class defined by the app developers using decorators and
+  `class MyModel extends Entity` syntax.
+- The "implementation" class created by legacy-juggler-bridge, this class is
+  inheriting from juggler's Model `base class`.
+
+If you take a deep look on it, you would find that LB4 model handles some data
+beofore it hands the model to Juggler. That's why `lb4` has some differences
```
bajtos:
```suggestion
before it hands the model to Juggler. That's why `lb4` has some differences
```

packages/repository/DEVELOPING.md
```diff
@@ -0,0 +1,37 @@
...
+compared to `lb3` when generating models.
+
+For example, you might notice that some entries in `@model decorator` are not
```
bajtos:
```suggestion
For example, you might notice that some entries in `@model` decorator are not
```

docs/site/Model.md
```diff
@@ -190,9 +188,197 @@ class Product extends Entity {
...
+However, the model decorator in LoopBack 4 is not exactly the same as what it is
+in lb3. For example, in
+[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
+pass in a model definition syntax in model decorator, such as properties,
+options, relation etc. But not all these entries are available in lb4 model
+decorator:
+
+<!-- Please modify this part when options is available -->
+
+NOTICE: in LoopBack 4 we only support `setting` in the ModelDefinitionSyntax for
```
b-admike: typo: `setting` -> `settings`

docs/site/Model.md
```diff
@@ -157,30 +157,28 @@ class MyFlexibleModel extends Entity {
...
-  },
   settings: {
     // etc...
   },
+  // define properties by @property decorator below
 })
 class Category extends Entity {
   // etc...
+  @property({type: 'number'})
+  CategoryId: number;
```
nabdelgadir: `categoryId`?

docs/site/Model.md
```diff
@@ -190,9 +188,197 @@ class Product extends Entity {
 }
 ```
 
-Additionally, the model decorator is able to build the properties object through
-the information passed in or inferred by the property decorators, so the
-properties key value pair can also be omitted.
+However, the model decorator in LoopBack 4 is not exactly the same as what it is
+in lb3. For example, in
```
nabdelgadir: Nitpick: since you used `LoopBack 4` earlier in the sentence, maybe change `lb3` to `LoopBack 3` for consistency?

docs/site/Model.md
```diff
@@ -190,9 +188,197 @@ class Product extends Entity {
...
+[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
+pass in a model definition syntax in model decorator, such as properties,
```
nabdelgadir: Nitpick: `in model decorator` -> `in the model decorator`

packages/repository/DEVELOPING.md
```diff
@@ -0,0 +1,37 @@
+## LB4 v.s LB3
```
nabdelgadir: `v.s` -> `vs.`

packages/repository/DEVELOPING.md
```diff
@@ -0,0 +1,37 @@
...
+| `default` | `false`                                                                                                                                           | `true`                                                                                                                                                                       |
+| code      | [defined here](https://github.com/strongloop/loopback-datasource-juggler/blob/2b8c1ebaeec3d8be87da38236db3d8210bca6230/lib/model-builder.js#L130) | [defined here](https://github.com/strongloop/loopback-next/blob/ba3f3894ef7e102b0fcd8b9b7a1f1d221bbbf4a4/packages/repository/src/repositories/legacy-juggler-bridge.ts#L160) |
+
+Hope this example shows the idea that even they both use the `base model` from
+Juggler, LB4 model doesn't use, or even have access to some fields/entries in
+the Juggler model!
+
+As mentioned before, we are using the legacy juggler bridge as a temporaty
```
nabdelgadir: Typo: `temporaty` -> `temporary`

packages/repository/DEVELOPING.md
```diff
@@ -0,0 +1,37 @@
...
+solution This may change in the future since the Juggler is pretty.. you know,
```
nabdelgadir: Missing `.` after `solution`?

PR #3364, opened 2019-07-15
[RFC WIP] feat(health): add a component to run health checks

raymondfeng:
Expose endpoint points per https://github.com/CloudNativeJS/cloud-health

## Checklist


raymondfeng:
> Do we need actually need the dependency on @cloudnative/health? Is it adding any additional health data that we don't already have available in LB4?

`@cloudnative/health` is an IBM package to define the convention of `health` status.

> How about the package name, should we call it @loopback/extension-health to better communicate the fact that it's an optional extension?

I'm open to the naming convention.

> I also feel the size of packages/ directory is growing too large, perhaps it's time to introduce a new top-level directory extensions/ to host packages like the one you are introducing?

I'm open to the new `extensions` directory.

extensions/health/src/health.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.application
+      .bind(HealthBindings.HEALTH_CHECKER)
+      .to(new HealthChecker());
+    this.application.lifeCycleObserver(HealthObserver);
+
+    const options: HealthOptions = Object.assign(
+      {},
+      DEFAULT_HEALTH_OPTIONS,
+      healthConfig,
+    );
```
bajtos: FYI, this can be simplified as follows:

```ts
const options: HealthOptions = {
  ...DEFAULT_HEALTH_OPTIONS,
  ...healthConfig,
};
```

extensions/health/src/controllers/health.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {HealthChecker} from '@cloudnative/health';
+import {bind, BindingScope, Constructor, inject} from '@loopback/core';
+import {get} from '@loopback/rest';
+import {HealthBindings} from '../keys';
+import {DEFAULT_HEALTH_OPTIONS, HealthOptions} from '../types';
+
+export function healthControllerFactory(
```
bajtos: Function names usually start with a verb.

```suggestion
export function createHealthController(
```

extensions/health/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@loopback/extension-health",
+  "version": "1.0.0",
```
bajtos: Let's start with version `0.1.0` to allow us to do few more iterations (possibly breaking things along the way) before we call the module stable and publish `1.0.0`.

bajtos:
> I'm open to the new extensions directory.

Excellent! Would you mind to create a standalone PR that will modify our monorepo infrastructure to support `extensions/` directory?

raymondfeng:
> Excellent! Would you mind to create a standalone PR that will modify our monorepo infrastructure to support extensions/ directory?

Unfortunately `lerna` does not allow a package location that does not have any packages.

raymondfeng:
> There is one thing that I would like to understand. Ideally, I'd like datasources to automatically contribute health or live checks (I don't understand the difference yet), e.g. using ping method to check the connection status. Does your current design allow such feature?

Yes. I added an example in README.

Issue #2483, opened 2019-02-26
[EPIC] From relation definition to REST API with auto-generated repository/controller classes


bajtos:
In LoopBack 3, it is very easy to get a fully-featured relational REST API with very little code: a model definition describing model relations + a model configuration specifying which datasource to use.

Let's enhance the solution provided by #2036 to include support for relations too.

- User creates a model class and uses decorators to define model relations. (No change here.)
- User declaratively defines what kind of data-access patterns to provide (CRUD, KeyValue, etc.) and what datasource to use under the hood - will be provided by #2036 
- `@loopback/boot` processes this configuration and registers appropriate repositories & controllers with the app.

## Acceptance criteria

- [ ] Enhance the solution implemented in #2036 to support all relations implemented in LB4 at the time of implementation. For example:
  - HasMany
  - BelongsTo
  - HasOne
  - HasMany through
  - ReferencesMany

- [ ] When an unknown relation is encountered while processing model definition, we should print a descriptive warning or throw an error to abort the application.

- [ ] For each relation not supported yet, update the existing GH issue or create a new GH issue to ensure that when we implement a new relation type, then we also implement support for declarative flavor of building the REST API.

- [ ] Documentation

- [ ] Blog post

jannyHou:
This story may need a spike to figure out how to map the relational paths to REST endpoints.

Reference story:
Hierarchical resource locator: https://github.com/strongloop/loopback-next/issues/802

/Users/{id}/orders
- delegate to user controller
- user controller interprets the `/orders`, resolve the url
- delegate to order controller

Proposal:
Delegate the url resolve to controller, instead of defining the path pattern in relation controller

jannyHou:
@jannyHou will create a spike.

dhmlau:
@jannyHou, could you please create a spike? (I thought I've created one but might've mixed it up with something else). Thanks.

jannyHou:
Spike created in https://github.com/strongloop/loopback-next/issues/2855 🙂

bajtos:
We should be using the same approach we already have in place for relations, where a new controller class is created for every relation. E.g. `UserOrdersController` to handle all endpoints in `/Users/{id}/orders` path space.

dhmlau:
Per the discussion during the estimation meeting yesterday, @raymondfeng suggested to make this task as an epic and start off with one relation first.  
[...]

PR #3387, opened 2019-07-18
RFC: Resolver for inclusion of related models

bajtos:
In this pull request, I am proposing a solution for resolving inclusion of related models and demonstrating the viability of the proposed solution.

See https://github.com/strongloop/loopback-next/issues/2634 and https://github.com/strongloop/loopback-next/issues/1352 for the previous discussions.

How to review this pull request:

Start by reading [`_SPIKE_.md`](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/_SPIKE_.md) document that provides a high-level overview of the proposal.

Next, read through the code to see implementation details you are interested in:

 - [examples/todo-list](https://github.com/strongloop/loopback-next/tree/spike/resolve-included-models/examples/todo-list) was updated to show the proposed usage in user applications
 - [packages/repository/src/repositories/legacy-juggler-bridge.ts](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/packages/repository/src/repositories/legacy-juggler-bridge.ts) shows the implementation in the base Repository class
 - Individual relation resolvers can be found here:
   - [packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts)
   - [packages/repository/src/relations/has-many/has-many.inclusion-resolver.ts](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/packages/repository/src/relations/has-many/has-many.inclusion-resolver.ts)
   - [packages/repository/src/relations/has-one/has-one.inclusion-resolver.ts](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/packages/repository/src/relations/has-one/has-one.inclusion-resolver.ts)
   - shared code: [packages/repository/src/relations/relation.helpers.ts](https://github.com/strongloop/loopback-next/blob/spike/resolve-included-models/packages/repository/src/relations/relation.helpers.ts)

### DIFF
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -1,0 +1,77 @@
+- [DONE] how to test inclusion against all connectors
+
+  - Move `@loopback/repository` integration + acceptance tests into a new
+    package, e.g. `@loopback/repository-testsuite`
+  - Basically, every test using `memory` datasource should be refactored into a
+    test that can be executed against any other connector too.
+  - Setup Travis CI jobs to execute those tests for a (sub)set of connectors as
+    part of loopback-next CI
+    - run example-todo & example-todo-list against other connectors too
+  - Execute the new test suite from each connector: add
+    `@loopback/repository-testsuite` into dev-dependencies and `npm test`
+  - Q! How to land changes requiring changes in both the test suite and (all)
+    connectors?
+
+  - How to structure the test suite: when adding a new relation, we should be
+    adding a new test file (possibly starting by copying the test file for an
+    existing relation)
+
+Edge cases to consider:
+
+- query model with relations, then save the instance returned by the query:
+  should repository methods remove navigational properties before saving?
+  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L104-L141
+
+- include nested models (recursive inclusion)
+  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L175-L195
+
+  User hasMany Post hasMany Comment
+
+  - `userRepo.find({include: {posts: 'comments'}})`
+  - `userRepo.find({include: {posts: {relation: 'comments'}}})`
+
+  OUT OF SCOPE OF THE INITIAL VERSION, this is not even supported by our TS
+  iface.
+
+- inclusion with custom scope -- OUT OF SCOPE OF INITIAL IMPL
+  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L247-L253
+
+  - custom "order", "limit", "skip" and "fields"
+  - additional "scope.where" constraint
+
+  OUT OF SCOPE OF THE INITIAL VERSION. We should throw an error when a custom
+  scope is present.
+
+- dataSource.settings.inqLimit
+
+- interaction between "include" and "fields" (when the primary or foreign key is
+  missing in fields, related models are not fetched)
+
+- different syntax flavours
+
+  - userRepo.find({include: ['posts', 'passports']})
+  - userRepo.find({include: [ {relation: 'posts', scope: {where: {title: 'Post
+    A'}}}, 'passports', ]}
+
+  OUT OF SCOPE OF THIS SPIKE See
+  https://github.com/strongloop/loopback-next/issues/3205
+
+- how to prevent certain relations to be traversed (e.g. User -> AccessToken)
+
+  Solved: don't register resolver for those relations in the repository class.
+
+- test how many DB calls are made by the resolvers
+  - including belongsTo should make only ' + nDBCalls + ' db calls'
+    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1317
+  - including hasManyThrough should make only 3 db calls'
+    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1340
+  - including hasMany should make only ' + dbCalls + ' db calls'
+    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1395
+  - should not make n+1 db calls in relation syntax'
+    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1431
+
+Additional things not covered by the initial implementation
+
+- HasAndBelongsToMany - search juggler's test/include.test.js
+- Inclusion for polymorphic relations, see jugglers' test/relations.test.js
+- hasOne with scope (??), see jugglers' test/relations.test.js
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -75,3 +75,74 @@ Additional things not covered by the initial implementation
 - HasAndBelongsToMany - search juggler's test/include.test.js
 - Inclusion for polymorphic relations, see jugglers' test/relations.test.js
 - hasOne with scope (??), see jugglers' test/relations.test.js
+
+
+---
+
+Ideally, I'd like LB4 to define MongoDB PK and FKs as follows:
+- `{type: 'string', mongodb: {dataType: 'ObjectID'}}`
+Even better, `dataType: 'ObjectID'` should be automatically applied by the
+connector for PK and FKs referencing ObjectID PKs.
+
+For example:
+
+```ts
+@model()
+class Product {
+  @property({
+    type: 'string',
+    generated: true,
+    // ^^ implies dataType: 'ObjectID'
+  })
+  id: string;
+
+  @property({
+    type: 'string',
+    references: {
+      model: () => Category,
+      property: 'id',
+    },
+    // ^^ implies dataType: 'ObjectID' when Category is attached to MongoDB
+  })
+  categoryId: string;
+}
+```
+
+For v1, I suppose we can ask developers to provide dataType manually.
+
+```ts
+@model()
+class Product {
+  @property({
+    type: 'string',
+    generated: true,
+    mongodb: {dataType: 'ObjectID'},
+  })
+  id: string;
+
+  @property({
+    type: 'string',
+    mongodb: {dataType: 'ObjectID'},
+  })
+  categoryId: string;
+}
+```
+
+With this setup in place, `id` and `categoryId` properties should be always
+returned as strings from DAO and connector methods.
+
+This is tricky to achieve for the PK property, because juggler overrides
+user-supplied type with connector's default type when the PK is generated
+by the database. See
+[`DataSource.prototype.setupDataAccess()`](https://github.com/strongloop/loopback-datasource-juggler/blob/0c2bb81dace3592ecde8b9eccbd70d589da44d7d/lib/datasource.js#L713-L719)
+
+Can we rework MongoDB connector to hide ObjectID complexity as an internal
+implementation detail and always use string values in public APIs? Accept
+ids as strings and internally convert them to ObjectID. Convert values returned
+by the database from ObjectID to strings.
+
+Downside: this is not going to work for free-form properties that don't have
+any type definition and where the connector does not know that they should
+be converted from string to ObjectID. But then keep in mind that JSON cannot
+carry type information, therefore REST API clients are not able to set
+free-form properties to ObjectID values even today.
--- a/_SPIKE_.md
+++ b/_SPIKE_.md
@@ -1,86 +1,399 @@
-- [DONE] how to test inclusion against all connectors
+# Spike: Resolve included models
 
-  - Move `@loopback/repository` integration + acceptance tests into a new
-    package, e.g. `@loopback/repository-testsuite`
-  - Basically, every test using `memory` datasource should be refactored into a
-    test that can be executed against any other connector too.
-  - Setup Travis CI jobs to execute those tests for a (sub)set of connectors as
-    part of loopback-next CI
-    - run example-todo & example-todo-list against other connectors too
-  - Execute the new test suite from each connector: add
-    `@loopback/repository-testsuite` into dev-dependencies and `npm test`
-  - Q! How to land changes requiring changes in both the test suite and (all)
-    connectors?
+Consider the following domain model(s):
 
-  - How to structure the test suite: when adding a new relation, we should be
-    adding a new test file (possibly starting by copying the test file for an
-    existing relation)
+- A model called `Category` with properties `id`, `name`.
+- A model called `Product` with properties `id`, `name`, `categoryId`
+- A `hasMany` relation (Category has many products)
 
-Edge cases to consider:
+Now consider the following code to retrieve all categories with all related
+products (perhaps to render a home page of a product catalog):
 
-- query model with relations, then save the instance returned by the query:
-  should repository methods remove navigational properties before saving?
-  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L104-L141
+```ts
+categoryRepo.find({include: [{relation: 'products'}]});
+```
 
-- include nested models (recursive inclusion)
-  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L175-L195
+## The problem
 
-  User hasMany Post hasMany Comment
+How to fetch products for each category found?
 
-  - `userRepo.find({include: {posts: 'comments'}})`
-  - `userRepo.find({include: {posts: {relation: 'comments'}}})`
+Additional constraints:
 
-  OUT OF SCOPE OF THE INITIAL VERSION, this is not even supported by our TS
-  iface.
+- The target model (`Product`) can be backed by a different database than the
+  source model (`Category`). For example, we can use MongoDB to store categories
+  and MySQL to store products.
 
-- inclusion with custom scope -- OUT OF SCOPE OF INITIAL IMPL
-  https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L247-L253
+- We need to test relations against real databases to ensure we are correctly
+  handling database-specific quirks like:
+  - a limit on the number of items that can be included in `inq`
+  - coercion between `ObjectID` vs. `string` for MongoDB.
 
-  - custom "order", "limit", "skip" and "fields"
-  - additional "scope.where" constraint
+## Proposed solution
 
-  OUT OF SCOPE OF THE INITIAL VERSION. We should throw an error when a custom
-  scope is present.
+### 1. Introduce the concept of `InclusionResolver`
 
-- dataSource.settings.inqLimit
+An inclusion resolver is a class that can fetch target models for the given list
+of source model instances. The idea is to create a different inclusion resolver
+for each relation type, e.g. `HasManyInclusionResolver`, etc.
 
-- interaction between "include" and "fields" (when the primary or foreign key is
-  missing in fields, related models are not fetched)
+```ts
+export interface InclusionResolver {
+  fetchIncludedModels<SourceWithRelations extends Entity>(
+    // list of source models as returned by the first database query
+    sourceEntities: SourceWithRelations[],
+    // inclusion requested by the user (e.g. scope constraints to apply)
+    inclusion: Inclusion,
+    // generic options object, e.g. carrying the Transaction object
+    options?: Options,
+  ): Promise<void>;
+  // ^^ nothing is returned, the source models are updated in-place
+}
+```
 
-- different syntax flavours
+With a bit of luck, we will be able to use these resolvers for GraphQL too.
+However, such use-case is out of scope of this spike.
 
-  - userRepo.find({include: ['posts', 'passports']})
-  - userRepo.find({include: [ {relation: 'posts', scope: {where: {title: 'Post
-    A'}}}, 'passports', ]}
+### 2. Base repository classes handle inclusions via resolvers
 
-  OUT OF SCOPE OF THIS SPIKE See
-  https://github.com/strongloop/loopback-next/issues/3205
+Each repository class (e.g. `DefaultCrudRepository` from our legacy juggler
+bridge) should maintain a map containing inclusion resolvers for each relation
+that is allowed to be included.
 
-- how to prevent certain relations to be traversed (e.g. User -> AccessToken)
+Conceptually:
 
-  Solved: don't register resolver for those relations in the repository class.
+```ts
+export class DefaultCrudRepository {
+  public inclusionResolvers: {[key: string]: InclusionResolver};
 
-- test how many DB calls are made by the resolvers
-  - including belongsTo should make only ' + nDBCalls + ' db calls'
-    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1317
-  - including hasManyThrough should make only 3 db calls'
-    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1340
-  - including hasMany should make only ' + dbCalls + ' db calls'
-    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1395
-  - should not make n+1 db calls in relation syntax'
-    https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L1431
+  // ...
+}
+```
 
-Additional things not covered by the initial implementation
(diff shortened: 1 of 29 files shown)


_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+    inclusion: Inclusion,
+    // generic options object, e.g. carrying the Transaction object
+    options?: Options,
+  ): Promise<void>;
+  // ^^ nothing is returned, the source models are updated in-place
+}
+```
+
+With a bit of luck, we will be able to use these resolvers for GraphQL too.
+However, such use-case is out of scope of this spike.
```
raymondfeng: See https://www.apollographql.com/docs/graphql-tools/resolvers/

raymondfeng: And https://www.apollographql.com/docs/graphql-tools/resolvers/

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+  fetchIncludedModels<SourceWithRelations extends Entity>(
+    // list of source models as returned by the first database query
+    sourceEntities: SourceWithRelations[],
+    // inclusion requested by the user (e.g. scope constraints to apply)
+    inclusion: Inclusion,
+    // generic options object, e.g. carrying the Transaction object
+    options?: Options,
+  ): Promise<void>;
+  // ^^ nothing is returned, the source models are updated in-place
+}
```
raymondfeng: I'm thinking louder here:

1. The generic resolver which can be backed by a property or a function with (optional) args, for example:

`name` -> `obj.name`, `obj.name(...)`, or resolve(obj, 'name', ...) 

2. InclusionResolver extends/implements the generic resolver

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+## Proposed solution
+
+### 1. Introduce the concept of `InclusionResolver`
+
+An inclusion resolver is a class that can fetch target models for the given list
+of source model instances. The idea is to create a different inclusion resolver
+for each relation type, e.g. `HasManyInclusionResolver`, etc.
+
+```ts
+export interface InclusionResolver {
```
raymondfeng: Can it be simplified as a function?

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+  fetchIncludedModels<SourceWithRelations extends Entity>(
+    // list of source models as returned by the first database query
+    sourceEntities: SourceWithRelations[],
+    // inclusion requested by the user (e.g. scope constraints to apply)
+    inclusion: Inclusion,
+    // generic options object, e.g. carrying the Transaction object
+    options?: Options,
```
raymondfeng: Is it possible to retrieve metadata describing the relation?
Do we want to introduce `InclusionResolutionContext`?

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+    super(Category, dataSource);
+
+    // we already have this line to create HasManyRepository factory
+    this.products = this.createHasManyRepositoryFactoryFor(
+      'todos',
+      productRepositoryGetter,
+    );
+
+    // add this line to register inclusion resolver
+    this.registerHasManyInclusion('products', this.productRepositoryGetter);
```
raymondfeng: As we already set up `this.products` as a factory to create a `HasManyRepository`, can we directly use it for the `inclusionResolver`?

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+
+In the proposed solution, this is solved by NOT REGISTERING any inclusion
+resolver for such relation.
+
+In the future, we can implement a "dummy" resolver that will report a helpful
+error for such relations (rather than a generic "unknown inclusion" error).
+
+```ts
+// usage
+this.registerForbiddenInclusion('accessTokens');
```
raymondfeng: Maybe it's better to name it as something like `this.prohibitInclusion('accessTokens')`?

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+  and MySQL to store products.
+
+- We need to test relations against real databases to ensure we are correctly
+  handling database-specific quirks like:
+  - a limit on the number of items that can be included in `inq`
+  - coercion between `ObjectID` vs. `string` for MongoDB.
+
+## Proposed solution
+
+### 1. Introduce the concept of `InclusionResolver`
```
raymondfeng: +1 to have `InclusionResolver` so that we can build up virtual properties, such as fetching it from a REST api call or a database.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -113,6 +130,10 @@ export class DefaultCrudRepository<
     );
 
     this.modelClass = this.definePersistedModel(entityClass);
+
+    // Important! Create the map object with `null` prototype to avoid
+    // Prototype Poisoning vulnerabilities.
+    this.inclusionResolvers = Object.create(null);
```
raymondfeng: What about using `Map`?

_SPIKE_.md
```diff
@@ -0,0 +1,496 @@
...
+  scope: {
+    where: {active: true},
+    limit: 1
+  }
+}
+```
+
+I am proposing to leave this feature out of scope of the initial release.
+However, we should tell the user when they try to use this feature via a 4xx
+error.
```
raymondfeng: Allowing extra constraint for the `include` is very important and it should be easy to implement as the relational repository already supports the criteria.

_SPIKE_.md
raymondfeng: Maybe the resolution can be handled as follows:

1. Use a registered resolver if exists
2. Check the named property/method on the entity
  - a method - calling the method to resolve the virtual property
  - a property (including getter) - use the value

_SPIKE_.md
bajtos: I'd like to keep InclusionResolver focused on Repository layer and don't deal with specific aspects of different transport protocols (REST, GraphQL, etc.)

To support GraphQL, we can build GraphQL resolvers that use InclusionResolvers as an implementation detail. (I.e. approach this problem in the opposite way from what you suggested.)

You raised a valid point though. I wanted to write an example code showing how a GraphQL resolver can be implemented using `fetchIncludedModels` API and realized that `fetchIncludedModels` is already too coupled with the way how we resolve navigational properties in ORM and REST APIs.

I am going to research this a bit more and try to come up with a simpler & more flexible API.

_SPIKE_.md
bajtos: That's an interesting idea! IIUC, you are suggesting that `this.products` (`HasManyRepositoryFactory`) can implement also `InclusionResolver` API, therefore this new line would become:

```ts
this.registerInclusion(this.products);
```

I like that approach! :+1:

_SPIKE_.md
bajtos: Most likely yes, a resolver can be a simple function too. I'll take a look.

However, if we want `HasManyRepositoryFactory` to provide inclusion resolver too (as discussed in https://github.com/strongloop/loopback-next/pull/3387#discussion_r305025912), then we need to store the resolver function as a property on that factory function.

Let me think about this some more.

_SPIKE_.md
bajtos: Metadata describing the relation is always the same for every resolver invocation, therefore it's not passed in the arguments. Instead, the metadata is provided to the class constructor. Similarly how a getter of a repository for the target model is provided.

> Do we want to introduce `InclusionResolutionContext`

I am not sure.

On the one hand, the current API where `inclusion` and `options` are provided via positional arguments, is simple and easy to understand. `InclusionResolutionContext` feels rather heavy-weight to me.

On the other hand, positional parameters are difficult to extend in the future. `InclusionResolutionContext` makes it easy to add additional context properties.

If we introduce `InclusionResolutionContext`, what properties would you like to see on that context object? Just `inclusion` and `options`?

_SPIKE_.md
bajtos: In https://github.com/strongloop/loopback-next/pull/3387#discussion_r305021466, you proposed to change resolvers from an interface/class to a simple function. To accommodate that, I think we will have to register resolvers slightly differently, e.g.

```ts
this.registerInclusion('products', this.products.inclusionResolver);
```

_SPIKE_.md
bajtos: > Check the named property/method on the entity

That does not make sense to me. A model/entity class represents the data structure, it does not know how to store or load that data. Data access is implemented by Repositories.

For the initial version, let's keep the scope small (think of MVP) and implement only the first step - "Use a registered resolver if exists". We can look into different ways (like using a named property or method) later, when we have a more concrete requirement for that.

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: Good point. I thought that objects are faster than `Map`.

I wrote a simple benchmark (http://jsben.ch/KZ25p), and found that in current Chrome `Map` is about 10% faster than objects.

bajtos: I am going to rework my code to use `Map`.

_SPIKE_.md
bajtos: Agreed, I will be creating a dedicated user story for this feature anyways.

In this spike, I am trying to identify MVP scope. IMO, support for extra constraints is an additive feature that should be implemented after we ship MVP. There will be several features out of MVP scope, we can prioritize and implement them based on their importance, but only after MVP is shipped.

@raymondfeng Do you agree to leave extra constraints out of the MVP scope?

_SPIKE_.md
bajtos: It's also important to ensure GraphQL resolvers are not prone to [SELECT N+1 PROBLEM](https://stackoverflow.com/q/97197/69868). I did a quick search and it looks like the recommended solution is to leverage [DataLoader](https://github.com/graphql/dataloader/) to batch multiple queries into a single one. DataLoader's example showing integration with GraphQL is not trivial: https://github.com/graphql/dataloader#using-with-graphql.

As I see it, implementing inclusion resolvers for GraphQL requires further research that's out of scope of this spike.

Having said that, I still would like to improve the current InclusionResolver API to make it more flexible.

_SPIKE_.md
bajtos: Done in 507820f

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: Done in 659fee6

_SPIKE_.md
bajtos: Done in a7260ff

_SPIKE_.md
bajtos: Addressed in 2f69523e3.

_SPIKE_.md
bajtos: Done in e1f6745a8

bajtos:
@raymondfeng Thank you for a great feedback! ❤️ 

I pushed few commits to improve my proposal, I think all comments should be either addressed in code or answered here on GitHub.

Do you have any more suggestions? Does the proposed design & implementation LGTY now?

_SPIKE_.md
```diff
@@ -0,0 +1,527 @@
...
+
+  constructor(
+    dataSource: juggler.DataSource,
+    productRepositoryGetter: Getter<ProductRepository>,
+  ) {
+    super(Category, dataSource);
+
+    // we already have this line to create HasManyRepository factory
+    this.products = this.createHasManyRepositoryFactoryFor(
+      'todos',
```
nabdelgadir: `products`?

examples/todo-list/src/repositories/todo-list-image.repository.ts
```diff
@@ -39,23 +34,21 @@ export class TodoListImageRepository extends DefaultCrudRepository<
...
-  ): Promise<TodoListImageWithRelations[]> {
-    // Prevent juggler for applying "include" filter
-    // Juggler is not aware of LB4 relations
-    const include = filter && filter.include;
-    filter = {...filter, include: undefined};
-
-    const result = await super.find(filter, options);
+    _options?: Options,
+  ): Promise<(TodoListImage & TodoListImageRelations)[]> {
+    const result = entities as (TodoListImage & TodoListImageRelations)[];
```
nabdelgadir: Why did you change `TodoListImageWithRelations[]` to `(TodoListImage & TodoListImageRelations)[]`?

examples/todo-list/src/repositories/todo.repository.ts
```diff
@@ -36,23 +36,21 @@ export class TodoRepository extends DefaultCrudRepository<
...
-  ): Promise<TodoWithRelations[]> {
-    // Prevent juggler for applying "include" filter
-    // Juggler is not aware of LB4 relations
-    const include = filter && filter.include;
-    filter = {...filter, include: undefined};
-
-    const result = await super.find(filter, options);
+    _options?: Options,
+  ): Promise<(Todo & TodoRelations)[]> {
+    const result = entities as (Todo & TodoRelations)[];
```
nabdelgadir: Same here

packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  const targetModel = relationMeta.target();
+  const targetName = targetModel.modelName;
+  debug('Resolved model %s from given metadata: %o', targetName, targetModel);
+
+  const targetProperties = targetModel.definition.properties;
+  debug('relation metadata from %o: %o', targetName, targetProperties);
+
+  if (relationMeta.keyTo) {
```
nabdelgadir: I know you're getting the content of this file from `packages/repository/src/relations/belongs-to/belongs-to-accessor.ts`, but in the latest version it includes [this commit](https://github.com/strongloop/loopback-next/commit/ce7538511034a3ab3f7a55d780b00c1c08557439) that adds a constraint here (same thing for `HasMany` and `HasOne`).

_SPIKE_.md
raymondfeng: Fair enough.

_SPIKE_.md
```diff
@@ -0,0 +1,527 @@
...
+      productRepositoryGetter,
+    );
+
+    // add this line to register inclusion resolver
+    this.registerInclusion('products', this.products.inclusionResolver);
```
raymondfeng: Does it make sense to have an option for `createHasManyRepositoryFactoryFor` to automate `this.registerInclusion('products', this.products.inclusionResolver)` by default?

_SPIKE_.md
bajtos: Good catch :+1:

examples/todo-list/src/repositories/todo-list-image.repository.ts
bajtos: That was not intentional. Thanks for spotting this change, I am going to revert this code back to use `TodoListImageWithRelations`.

bajtos: In fact, `includeRelatedModels` should not be needed now that the constructor calls `this.registerInclusion('todoList', this.todoList.inclusionResolver)`

examples/todo-list/src/repositories/todo.repository.ts
bajtos: I am going to remove `includeRelatedModels`, it's no longer needed.

packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
bajtos: Yeah, because of that commit, I am getting merge conflicts when trying to rebase. My intention here is just to move `resolveBelongsToMetadata` into a new file, I don't want to revert any changes made on the `master` branch.

_SPIKE_.md
```diff
@@ -0,0 +1,527 @@
...
+}
+```
+
+### MongoDB and `ObjectID` type
+
+MongoDB is tricky.
+
+- It uses a custom `ObjectID` type for primary keys.
+- `ObjectID` is represented as a `string` when converted to JSON
+- In queries, string values must be case to ObjectID, otherwise they are not
```
hacksparrow: case -> cast

_SPIKE_.md
bajtos: Fixed in d0db330a6

examples/todo-list/src/repositories/todo.repository.ts
bajtos: Fixed in d0db330a6

_SPIKE_.md
```diff
@@ -0,0 +1,527 @@
...
+user-supplied type with connector's default type when the PK is generated by the
+database. See
+[`DataSource.prototype.setupDataAccess()`](https://github.com/strongloop/loopback-datasource-juggler/blob/0c2bb81dace3592ecde8b9eccbd70d589da44d7d/lib/datasource.js#L713-L719)
+
+Can we rework MongoDB connector to hide ObjectID complexity as an internal
+implementation detail and always use string values in public APIs? Accept ids as
+strings and internally convert them to ObjectID. Convert values returned by the
+database from ObjectID to strings.
+
+Downside: this is not going to work for free-form properties that don't have any
```
hacksparrow: In that case, the type should be interpreted as "whatever the DB driver returns". So, I don't really see it as a downside.

_SPIKE_.md
```diff
@@ -0,0 +1,527 @@
...
+
+This is tricky to achieve for the PK property, because juggler overrides
+user-supplied type with connector's default type when the PK is generated by the
+database. See
+[`DataSource.prototype.setupDataAccess()`](https://github.com/strongloop/loopback-datasource-juggler/blob/0c2bb81dace3592ecde8b9eccbd70d589da44d7d/lib/datasource.js#L713-L719)
+
+Can we rework MongoDB connector to hide ObjectID complexity as an internal
+implementation detail and always use string values in public APIs? Accept ids as
+strings and internally convert them to ObjectID. Convert values returned by the
+database from ObjectID to strings.
```
hacksparrow: @bajtos  If I am interpreting this correctly, you are suggesting to get rid of the `ObjectID` type altogether from the MongoDB connector. Right?

_SPIKE_.md
bajtos: I did consider that idea briefly.

I find it problematic for a method called `create(...)FactoryFor` to update register inclusion resolvers, it feels like a surprising side effect to me.

I consider `createHasManyRepositoryFactoryFor` and `registerInclusion` as lower-level building blocks that should stay decoupled, and can be used to build higher-level abstractions in the future.

For example, I can imagine creating a new method called e.g. `registerHasManyRelation`, that would call both `createHasManyRepositoryFactoryFor` and `registerInclusion` under the hood and initialize any repository properties (e.g. `this.products`) along the way. 

```ts
// now
this.products = this.createHasManyRepositoryFactoryFor(
  'products',
  productRepositoryGetter,
);
this.registerInclusion('products', this.products.inclusionResolver);

// with the new API
this.registerHasManyRelation(
  'products', 
  productRepositoryGetter, 
  {allowInclusion: true} // optional
);
```

The non-trivial part is how to do this in a type-safe way - we need to let TypeScript understand that the string `'products'` must match a property `this.products` with type `HasManyRepositoryFactory`.

Anyhow, I'd like to leave this out of this initial MVP scope.

I think it may be better to spend our time on https://github.com/strongloop/loopback-next/issues/2483 (_From relation definition to REST API with no repository/controller classes_), that will allow developers to avoid manually written Repository classes altogether.

@raymondfeng If you like, I can create a follow-up task to investigate a more ergonomic API for registering relations inside Repository classes. WDYT?

_SPIKE_.md
bajtos: Well, from the point of view of a REST API client, "whatever the DB driver returns" is eventually converted to "JSON representation of whatever the DB driver returns", which is a string for ObjectID.

If we stay only in JS/TS land, then I agree with you. The DB driver returns the same type that was used to supply the (initial) value of the free-form property, thus all should be good.

_SPIKE_.md
bajtos: I am proposing to modify MongoDB connector to hide `ObjectID` type as an internal implementation detail, an use `string` type in all other LoopBack layers (Models, Repositories, Controllers).

In the database, primary and foreign keys must be still stored as `ObjectID`, this is important e.g. for performance reasons. The connector will still have to deal with `ObjectID` type.

I just don't want `ObjectID` to leak outside of the connector.

BTW this will be a breaking change that will also break compatibility with LB3. I think it is not an issue, just want to be clear about the ramifications.

Also, these changes are not strictly required for the implementation of inclusion. It's a wider problem relevant to relations in general.

packages/repository/src/model.ts
```diff
@@ -263,6 +263,13 @@ export abstract class ValueObject extends Model implements Persistable {}
  * Base class for entities which have unique ids
  */
 export abstract class Entity extends Model implements Persistable {
+  /**
+   * Get the name of the identity property (the primary key).
+   */
+  static getIdProperties(): string[] {
+    return this.definition.idProperties();
+  }
```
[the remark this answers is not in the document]
bajtos: See https://github.com/strongloop/loopback-next/pull/3410

packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
bajtos: See https://github.com/strongloop/loopback-next/pull/3410

_SPIKE_.md
hacksparrow: > I just don't want ObjectID to leak outside of the connector.

Makes sense. There is an existing issue (I encountered while writing tests, but we decided to defer tackling it), in which a property typed as `ObjectID` is returned as string by the MongoDB driver. The proposed change will eliminate such issues.

bajtos:
Rebased on top of the latest master 💪

bajtos:
Rebased the branch one more time, updated the SPIKE document with table-of-contents and a list of implementation tasks.

This spike is ready for final review.

@raymondfeng @strongloop/sq-lb-apex

_SPIKE_.md
bajtos: > `this.registerHasManyRelation(`

Perhaps just `this.hasMany` to keep it similar to what we have in LB3.

```ts
// inside CategoryRepository constructor
this.hasMany('products', productRepositoryGetter);
```

_SPIKE_.md
```diff
@@ -0,0 +1,772 @@
...
+  this file needs to be split into multiple smaller files, one (or more) for
+  each relation type.
+
+I am not sure how much work will be required to make these tests pass against
+MongoDB. If it would require non-trivial amount of time, then:
+
+- Introduce a new `CrudFeature` flag allowing MongoDB test suite to skip these
+  new tests.
+- Open a follow-up issue to investigate & fix the MongoDB test suite later.
+
```
bajtos: FWIW, the test added by this spike pull request is failing on MongoDB.

```
  1) MongoDB + DefaultCrudRepository
       CRUD Repository operations
         retrieve models including relations
           ignores navigational properties when updating model instance:
     AssertionError: expected undefined to have property length
      at Assertion.fail (/home/travis/build/strongloop/loopback-next/packages/testlab/node_modules/should/as-function.js:275:17)
      at Assertion.value (/home/travis/build/strongloop/loopback-next/packages/testlab/node_modules/should/as-function.js:356:19)
      at Context.it (/home/travis/build/strongloop/loopback-next/packages/repository-tests/src/crud/retrieve-including-relations.suite.ts:139:37)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

_SPIKE_.md
```diff
@@ -0,0 +1,772 @@
...
+  include: [{relation: 'products'}],
+});
+```
+
+In LB3, I think we were silently ignoring `include` filter in such case.
+
+I am proposing to be more explicit in LB4 and reject such queries with a 4xx
+(Bad Request) error.
+
+Later, we can improve our implementation to automatically add PK/FK properties
```
b-admike: 👍

hacksparrow:
LGTM. Next we need to update the `repository` package's `README.md` and show usage examples in the website docs.

bajtos:
@hacksparrow 

> LGTM. Next we need to update the repository package's README.md and show usage examples in the website docs.

Can you please clarify what do you mean by updating repository package's README? I mean what kind of content would you like to see there? AFAICT, the current README does not describe relations at all.

+1 to show usage examples in the docs. At the moment, my proposal has the following sub-task for each relation type:

> Update documentation (e.g. [Configuring a hasMany relation](https://loopback.io/doc/en/lb4/HasMany-relation.html#configuring-a-hasmany-relation)) to explain how to enable or disable inclusion.

I'll expand that sub-tasks to cover examples of queries requesting inclusion of related models. Is that what you are asking for?

hacksparrow:
> I'll expand that sub-tasks to cover examples of queries requesting inclusion of related models.

Yes, that's what I actually meant. Please disregard the mention of `repository`'s `README.md`, I got too lost in the implementation details.

bajtos:
I have created the following follow-up stories.

**MVP scope**

- Run repository tests for PostgreSQL #3436
- Run repository tests for Cloudant #3437
- Reject create/update requests when data contains navigational properties #3439
- Verify relation type in `resolve{Relation}Metadata` #3440
- Add `keyFrom` to resolved relation metadata #3441
- Test relations against databases #3442
- Add `findByForeignKeys` helper (initial version) #3443
- Support `inq` splitting in `findByForeignKeys` #3444
- Introduce `InclusionResolver` concept #3445
- Include related models in `DefaultCrudRepository` #3446
- Implement InclusionResolver for hasMany relation #3447
- Implement InclusionResolver for belongsTo relation #3448
- Implement InclusionResolver for hasOne relation #3449
- Update `todo-list` example to use inclusion resolver #3450
- Add inclusion resolvers to `lb4 relation` CLI #3451
- Blog post: announce Inclusion of related models #3452

**Post-MVP**
- Include related models with a custom scope #3453
- Recursive inclusion of related models #3454
- Reject queries with incompatible "filter.fields" and "filter.include" #3455
- Spike: robust handling of ObjectID type for MongoDB #3456

PR #3388, opened 2019-07-18
feat(context): use invocation context for method dependency injection

raymondfeng:
Use case: interceptors may add or change bindings to influence how the
method parameter injection is resolved.

The current implementation does not use the invocation context to resolve parameter injection. As a result, interceptors cannot rebind new values.

## Checklist

### DIFF
--- a/packages/context/src/interceptor.ts
+++ b/packages/context/src/interceptor.ts
@@ -11,6 +11,7 @@ import {
   MetadataMap,
   MethodDecoratorFactory,
 } from '@loopback/metadata';
+import * as assert from 'assert';
 import * as debugFactory from 'debug';
 import {Binding, BindingTemplate} from './binding';
 import {bind} from './binding-decorator';
@@ -26,6 +27,7 @@ import {
 import {
   InvocationArgs,
   InvocationContext,
+  InvocationOptions,
   InvocationResult,
 } from './invocation';
 import {
@@ -284,13 +286,18 @@ export function intercept(...interceptorOrKeys: InterceptorOrKey[]) {
  * @param target - Target class (for static methods) or object (for instance methods)
  * @param methodName - Method name
  * @param args - An array of argument values
+ * @param options - Options for the invocation
  */
 export function invokeMethodWithInterceptors(
   context: Context,
   target: object,
   methodName: string,
   args: InvocationArgs,
+  options: InvocationOptions = {},
 ): ValueOrPromise<InvocationResult> {
+  // Do not allow `skipInterceptors` as it's against the function name
+  // `invokeMethodWithInterceptors`
+  assert(!options.skipInterceptors, 'skipInterceptors is not allowed');
   const invocationCtx = new InterceptedInvocationContext(
     context,
     target,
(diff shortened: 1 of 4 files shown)


packages/context/src/invocation.ts
```diff
@@ -137,44 +148,38 @@ export function invokeMethod(
...
+ * Invoke a method. Method parameter dependency injection is honored.
  * @param target - Target of the method, it will be the class for a static
  * method, and instance or class prototype for a prototype method
  * @param method - Name of the method
  * @param ctx - Context
  * @param nonInjectedArgs - Optional array of args for non-injected parameters
  */
-function invokeMethodWithInvoker(
-  invoker: MethodInvoker,
+function invokeTargetMethodWithInjection(
```
jannyHou: add similar assertion here?

`assert(!options.skipInjection, 'skipInjection is not allowed');`

raymondfeng: No, this method is only for the target. No interceptors are involved and it does not have `options` arg.

emonddr:
@raymondfeng ,

I know this wasn't part of your current commit , but for 
```
it('invokes async interceptors on a sync method', async () => {
    class MyController {
      // Apply `log` to a sync instance method
      @intercept(log)
      greet(name: string) {
        return `Hello, ${name}`;
      }
    }
...
```

which uses a `sync` method.

For consistency, shouldn't we use a method name of `greetSync` like the test above it?
```
it('invokes sync interceptors on a sync method', () => {
    class MyController {
      // Apply `logSync` to a sync instance method
      @intercept(logSync)
      greetSync(name: string) {
        return `Hello, ${name}`;
      }
    }
...
```

packages/context/src/interceptor.ts
```diff
@@ -284,13 +286,16 @@ export function intercept(...interceptorOrKeys: InterceptorOrKey[]) {
...
+ * @param options - Options for the invocation
  */
 export function invokeMethodWithInterceptors(
   context: Context,
   target: object,
   methodName: string,
   args: InvocationArgs,
+  options: InvocationOptions = {},
 ): ValueOrPromise<InvocationResult> {
+  assert(!options.skipInterceptors, 'skipInterceptors is not allowed');
```
bajtos: What is the reasoning behind the decision to disallow `skipInterceptors`? What is that flag good for when it cannot be used? Please add code comments to explain.

raymondfeng: The function name is `invokeMethodWithInterceptors`. That's why I added an assertion to make sure `skipInterceptors` cannot be set to `true`.

raymondfeng:
> @raymondfeng ,
> 
> I know this wasn't part of your current commit , but for
> 
> ```
> it('invokes async interceptors on a sync method', async () => {
>     class MyController {
>       // Apply `log` to a sync instance method
>       @intercept(log)
>       greet(name: string) {
>         return `Hello, ${name}`;
>       }
>     }
> ...
> ```
> 
> which uses a `sync` method.
> 
> For consistency, shouldn't we use a method name of `greetSync` like the test above it?
> 
> ```
> it('invokes sync interceptors on a sync method', () => {
>     class MyController {
>       // Apply `logSync` to a sync instance method
>       @intercept(logSync)
>       greetSync(name: string) {
>         return `Hello, ${name}`;
>       }
>     }
> ...
> ```

I'm fine either way.

PR #3397, opened 2019-07-19
feat(repository): add beginTransaction API

### LINKED ISSUE #2614: Docs: How to create a transaction and pass it around through various repo CRUD operations

Cross posting my comments: 
> Talked to @bajtos @raymondfeng, what we should do regarding transaction support:
> - shorter term: explain to users how to open a transaction and pass the transaction token/handles around in options.
> - longer term: have a spike on a more elegant solution.

There are different workarounds that our users have posted:
- https://github.com/strongloop/loopback-next/issues/1599
- https://github.com/strongloop/loopback-connector-mongodb/issues/489

Loopback 3 provides transaction support at `model` level.

In LoopBack 4 this should be at the `Repository` level.

There will be  `sugar API` in a `repository` to delegate the work to the `datasource` level.
```ts

//obtain a transaction object with  repo.beginTransaction()
let transaction: Transaction =
await customerRepository.beginTransaction({isolationLevel: Post.Transaction.READ_COMMITTED});

// isolationLevel values:
// Transaction.READ_UNCOMMITTED
// Transaction.READ_COMMITTED (default)
// Transaction.REPEATABLE_READ
// Transaction.SERIALIZABLE

//pass the transaction inside the existing options object
await customerRepository.create( customer, { transaction} ) ;

//pass the transaction inside the existing options object
await orderRepository.create( order, { transaction} ); 

// the two repositories above (customer and order) have the same data source, 
// so they can share the transaction.

// finish working with the transaction, either with commit or rollback
await transaction.commit();//  ( OR await transaction.rollback() )
```
The `Transaction` type mentioned above should be a wrapper for the typescript type of the transaction object in Juggler.

When using the juggler bridge, we need to make sure that the options we set in the wrapper type are consistent with what juggler is expecting.

Since LoopBack 4 doesn't support a distributed transaction, all repos using the transaction must be from the same data source. It is the job of `wrapper` `Transaction` to throw an exception if user attempts to use a transaction against disparate data sources.

DefaultCRUDRepository will implement an additional interface to handle transactions. 

## Acceptance criteria
- [x] **Only** covers relational databases (NoSQL databases are out of scope)
- [x] Document relational database `transaction support` in LoopBack 4
    - [x] Transaction wrapper type 
    - [x] How to create a transaction using repo.beginTransaction() and isolation level argument 
    - [x]  How to pass transaction object into the options object of various repository CRUD methods 
    - [x] How to terminate a transaction via commit() or rollback()
- [x] Create mocha test case for 1 SQL relational database (juggler tests have exhaustive transaction tests already)

b-admike:
Closes #2614 

## Checklist

### DIFF
--- a/docs/site/Transactions.md
+++ b/docs/site/Transactions.md
@@ -1,0 +1,200 @@
+---
+title: 'Using database transactions'
+lang: en
+layout: page
+keywords: LoopBack 4.0, LoopBack 4, Transactions, Transaction
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Using-database-transactions.html
+summary: Transactional API usage in LoopBack 4
+---
+
+## Overview
+
+A *transaction* is a sequence of data operations performed as a single logical
+unit of work. Many relational databases support transactions to help enforce
+data consistency and business logic requirements.
+
+A repository can perform operations in a transaction when the backing datasource
+is attached to one of the following connectors:
+
+- [MySQL connector](MySQL-connector.html) (IMPORTANT: Only with InnoDB as the
+  storage engine).
+- [PostgreSQL connector](PostgreSQL-connector.html)
+- [SQL Server connector](SQL-Server-connector.html)
+- [Oracle connector](Oracle-connector.html)
+- [DB2 Connector](DB2-connector.html)
+- [DashDB Connector](DashDB.html)
+- [DB2 iSeries Connector](DB2-iSeries-connector.html)
+- [DB2 for z/OS connector](DB2-for-z-OS.html)
+- [Informix connector](Informix.html)
+
+The repository class needs to extend from `TransactionalRepository` repository
+interface which exposes the `beginTransaction()` method. Note that LoopBack only
+supports database local transactions - only operations against the same
+transaction-capable datasource can be grouped into a transaction.
+
+## Transaction APIs
+
+The `@loopback/repository` package includes `TransactionalRepository` interface
+based on `EntityCrudRepository` interface. The `TransactionalRepository`
+interface adds a `beginTransaction()` API that, for connectors that allow it,
+will start a new Transaction. The `beginTransaction()` function gives access to
+the lower-level transaction API, leaving it up to the user to create and manage
+transaction objects, commit them on success or roll them back at the end of all
+intended operations. See [Handling Transactions](#handling-transactions) below
+for more details.
+
+## Handling Transactions
+
+See
+the [API reference](https://apidocs.strongloop.com/loopback-datasource-juggler/#datasource-prototype-begintransaction) for
+full transaction lower-level API documentation.
+
+Performing operations in a transaction typically involves the following steps:
+
+- Start a new transaction.
+- Perform create, read, update, and delete operations in the transaction.
+- Commit or rollback the transaction.
+
+### Start transaction
+
+Use the `beginTransaction()` method to start a new transaction from a repository
+class using `DefaultTransactionalRepository` as a base class.
+
+Here is an example:
+
+```ts
+import {
+  Transaction,
+  DefaultTransactionalRepository,
+  IsolationLevel,
+} from '@loopback/repository';
+// assuming there is a Note model extending Entity class, and
+// ds datasource which is backed by a transaction enabled
+// connector
+const repo = new DefaultTransactionalRepository(Note, ds);
+// Now we have a transaction (tx)
+const tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
+```
+
+You can also extend `DefaultTransactionalRepository` for custom classes:
+
+```ts
+import {inject} from '@loopback/core';
+import {
+  juggler,
+  Transaction,
+  DefaultTransactionalRepository,
+  IsolationLevel,
+} from '@loopback/repository';
+import {Note, NoteRelations} from '../models';
+
+export class NoteRepository extends DefaultTransactionalRepository<
+  Note,
+  typeof Note.prototype.id,
+  NoteRelations
+> {
+  constructor(@inject('datasources.ds') ds: juggler.DataSource) {
+    super(Note, ds);
+  }
+}
+```
+
+#### Isolation levels
+
+When you call `beginTransaction()`, you can optionally specify a transaction
+isolation level. LoopBack transactions support the following isolation levels:
+
+- `Transaction.READ_UNCOMMITTED`
+- `Transaction.READ_COMMITTED` (default)
+- `Transaction.REPEATABLE_READ`
+- `Transaction.SERIALIZABLE`
+
+If you don't specify an isolation level, the transaction uses READ_COMMITTED .
+
+{% include important.html content="
+
+**Oracle only supports READ_COMMITTED and SERIALIZABLE.**
+
+" %}
+
+For more information about database-specific isolation levels, see:
+
+- [MySQL SET TRANSACTION Syntax](https://dev.mysql.com/doc/refman/5.7/en/set-transaction.html)
+- [Oracle Isolation Levels](http://docs.oracle.com/cd/B14117_01/server.101/b10743/consist.htm#i17856)
+- [PostgreSQL Transaction Isolation](http://www.postgresql.org/docs/9.4/static/transaction-iso.html)
+- [SQL Server SET TRANSACTION ISOLATION LEVEL](https://msdn.microsoft.com/en-us/library/ms173763.aspx)
+
+### Perform operations in a transaction
+
+To perform create, retrieve, update, and delete operations in the transaction,
+add the transaction object to the `Options` parameter of the standard 
+[`create()`](https://loopback.io/doc/en/lb4/apidocs.repository.defaultcrudrepository.create.html),
+[`update()`](https://loopback.io/doc/en/lb4/apidocs.repository.defaultcrudrepository.update.html),
+[`deleteAll()`](https://loopback.io/doc/en/lb4/apidocs.repository.defaultcrudrepository.deleteall.html)
+(and so on) methods.
+
+For example, again assuming a `Note` model, `repo` transactional repository, and
+transaction object `tx` created as demonstrated in
+[Start transaction](#start-transaction) section:
+
+```ts
+const created = await repo.create({title: 'Groceries'}, {transaction: tx});
+const updated = await repo.update(
+  {title: 'Errands', id: created.id},
+  {transaction: tx},
+);
+
+// commit the transaction to persist the changes
+await tx.commit();
+```
+
+Propagating a transaction is explicit by passing the transaction object via the
+options argument for all create, retrieve, update, and delete and relation
+methods.
+
+### Commit or rollback
+
+Transactions allow you either to commit the transaction and persist the CRUD
+behaviour onto the database or rollback the changes. The two methods available
+on transaction objects are as follows:
+
+```ts
+  /**
+   * Commit the transaction
+   */
+  commit(): Promise<void>;
+
+  /**
+   * Rollback the transaction
+   */
+  rollback(): Promise<void>;
+```
+
+## Set up timeout
+
+You can specify a timeout (in milliseconds) to begin a transaction. If a
+transaction is not finished (committed or rolled back) before the timeout, it
+will be automatically rolled back upon timeout by default.
+
+For example, again assuming a `Note` model and `repo` transactional repository,
+the `timeout` can be specified as part of the `Options` object passed into the
+`beginTransaction` method.
+
+```ts
+const tx: Transaction = await repo.beginTransaction({
+  isolationLevel: IsolationLevel.READ_COMMITTED,
+  timeout: 30000, // 30000ms = 30s
+});
+```
+
+## Avoid long waits or deadlocks
+
+Please be aware that a transaction with certain isolation level will lock
+database objects. Performing multiple methods within a transaction
+asynchronously has the great potential to block other transactions (explicit or
+implicit). To avoid long waits or even deadlocks, you should:
+
+1.  Keep the transaction as short-lived as possible
+2.  Don't serialize execution of methods across multiple transactions
--- a/packages/repository-tests/src/crud-test-suite.ts
+++ b/packages/repository-tests/src/crud-test-suite.ts
@@ -32,6 +32,7 @@ export function crudRepositoryTestSuite(
     idType: 'string',
     freeFormProperties: true,
     emptyValue: undefined,
+    supportsTransactions: true,
     ...partialFeatures,
   };
 
--- a/packages/repository-tests/src/crud/transactions.suite.ts
+++ b/packages/repository-tests/src/crud/transactions.suite.ts
@@ -1,0 +1,176 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/repository-tests
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  Entity,
+  IsolationLevel,
+  juggler,
+  model,
+  property,
+  Transaction,
+  TransactionalEntityRepository,
+} from '@loopback/repository';
+import {expect, skipIf, toJSON} from '@loopback/testlab';
+import {Suite} from 'mocha';
+import {withCrudCtx} from '../helpers.repository-tests';
+import {
+  CrudFeatures,
+  CrudTestContext,
+  DataSourceOptions,
+  TransactionalRepositoryCtor,
+} from '../types.repository-tests';
+
+// Core scenarios for testing CRUD functionalities of Transactional connectors
+// Please keep this file short, put any advanced scenarios to other files
+export function transactionSuite(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: TransactionalRepositoryCtor,
+  connectorFeatures: CrudFeatures,
+) {
+  skipIf<[(this: Suite) => void], void>(
+    !connectorFeatures.supportsTransactions,
+    describe,
+    `transactions`,
+    () => {
+      @model()
+      class Product extends Entity {
+        @property({
+          type: connectorFeatures.idType,
+          id: true,
+          generated: true,
+          description: 'The unique identifier for a product',
+        })
+        id: number | string;
+
+        @property({type: 'string', required: true})
+        name: string;
+
+        constructor(data?: Partial<Product>) {
+          super(data);
+        }
+      }
+
+      describe('create-retrieve with transactions', () => {
+        let repo: TransactionalEntityRepository<
+          Product,
+          typeof Product.prototype.id
+        >;
+        let tx: Transaction | undefined;
+        before(
+          withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+            repo = new repositoryClass(Product, ctx.dataSource);
+            await ctx.dataSource.automigrate(Product.name);
+          }),
+        );
+        beforeEach(() => {
+          tx = undefined;
+        });
+        afterEach(async () => {
+          // FIXME: replace tx.connection with tx.isActive when it become
+          // available
+          // see https://github.com/strongloop/loopback-next/issues/3471
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          if (tx !== undefined && (tx as any).connection) {
+            await tx.rollback();
+          }
+        });
+
+        it('retrieves model instance once transaction is committed', async () => {
+          tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
+          const created = await repo.create(
+            {name: 'Pencil'},
+            {transaction: tx},
+          );
+          expect(created.toObject()).to.have.properties('id', 'name');
+          expect(created.id).to.be.ok();
+
+          await tx.commit();
+          const foundOutsideTransaction = await repo.findById(created.id, {});
+          expect(toJSON(created)).to.deepEqual(toJSON(foundOutsideTransaction));
+        });
+
+        it('can rollback transaction', async () => {
+          tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
+
+          const created = await repo.create(
+            {name: 'Pencil'},
+            {transaction: tx},
+          );
+          expect(created.toObject()).to.have.properties('id', 'name');
+          expect(created.id).to.be.ok();
+
+          const foundInsideTransaction = await repo.findById(
+            created.id,
+            {},
+            {
+              transaction: tx,
+            },
+          );
+          expect(toJSON(created)).to.deepEqual(toJSON(foundInsideTransaction));
+          await tx.rollback();
+          await expect(repo.findById(created.id, {})).to.be.rejectedWith({
+            code: 'ENTITY_NOT_FOUND',
+          });
+        });
+
+        it('ensures transactions are isolated', async () => {
+          tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
+          const created = await repo.create(
+            {name: 'Pencil'},
+            {transaction: tx},
+          );
+          expect(created.toObject()).to.have.properties('id', 'name');
+          expect(created.id).to.be.ok();
+
+          const foundInsideTransaction = await repo.findById(
+            created.id,
+            {},
+            {
+              transaction: tx,
+            },
+          );
+          expect(toJSON(created)).to.deepEqual(toJSON(foundInsideTransaction));
+          await expect(repo.findById(created.id, {})).to.be.rejectedWith({
+            code: 'ENTITY_NOT_FOUND',
+          });
+        });
+
+        it('should not use transaction with another repository', async () => {
+          const ds2Options = Object.assign({}, dataSourceOptions);
+          ds2Options.name = 'anotherDataSource';
+          ds2Options.database = ds2Options.database + '-new';
+          const ds2 = new juggler.DataSource(ds2Options);
+          const anotherRepo = new repositoryClass(Product, ds2);
+          await ds2.automigrate(Product.name);
+
+          tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
+
+          // we should reject this call with a clear error message
+          // stating that transaction doesn't belong to the repository
+          // and that only local transactions are supported
+          // expect(
+          //   await anotherRepo.create({name: 'Pencil'}, {transaction: tx}),
+          // ).to.throw(/some error/);
+          // see https://github.com/strongloop/loopback-next/issues/3483
+          const created = await anotherRepo.create(
+            {name: 'Pencil'},
+            {transaction: tx},
+          );
+          expect(created.toObject()).to.have.properties('id', 'name');
+          expect(created.id).to.be.ok();
+
+          // for now, LB ignores the transaction so the instance
+          // should be created regardless
+          await tx.rollback();
+          const foundOutsideTransaction = await anotherRepo.findById(
+            created.id,
+            {},
+          );
+          expect(toJSON(created)).to.deepEqual(toJSON(foundOutsideTransaction));
+        });
+      });
+    },
+  );
+}
--- a/packages/repository-tests/src/types.repository-tests.ts
+++ b/packages/repository-tests/src/types.repository-tests.ts
@@ -8,6 +8,7 @@ import {
   EntityCrudRepository,
   juggler,
   Options,
+  TransactionalEntityRepository,
 } from '@loopback/repository';
 
 /**
@@ -46,6 +47,14 @@ export interface CrudFeatures {
    * Default: `undefined`
    */
   emptyValue: undefined | null;
+  /**
+   * Does the connector support using transactions for performing CRUD
+   * operations atomically and being able to commit or rollback the changes?
+   * SQL databases usually support transactions
+   *
+   * Default: `false`
+   */
+  supportsTransactions: boolean;
 }
 
 /**
@@ -60,6 +69,19 @@ export type CrudRepositoryCtor = new <
   entityClass: typeof Entity & {prototype: T},
   dataSource: juggler.DataSource,
 ) => EntityCrudRepository<T, ID, Relations>;
+
+/**
+ * A constructor of a class implementing TransactionalRepository interface,
+ * accepting the Entity class (constructor) and a dataSource instance.
+ */
+export type TransactionalRepositoryCtor = new <
+  T extends Entity,
+  ID,
+  Relations extends object
+>(
+  entityClass: typeof Entity & {prototype: T},
+  dataSource: juggler.DataSource,
+) => TransactionalEntityRepository<T, ID, Relations>;
 
 /**
  * Additional properties added to Mocha TestContext/SuiteContext.
--- a/packages/repository/src/__tests__/unit/repositories/crud.repository.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/crud.repository.unit.ts
@@ -4,89 +4,14 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {expect} from '@loopback/testlab';
-
 import {
+  CrudConnector,
   CrudRepository,
   CrudRepositoryImpl,
   DataSource,
   Entity,
-  EntityData,
-  Class,
-  Filter,
-  Where,
-  Options,
-  CrudConnector,
-  AnyObject,
-  Count,
 } from '../../..';
-
-/**
- * A mock up connector implementation
- */
-class CrudConnectorStub implements CrudConnector {
-  private entities: EntityData[] = [];
-  name: 'my-connector';
-
-  connect() {
-    return Promise.resolve();
-  }
-
-  disconnect() {
-    return Promise.resolve();
-  }
-
-  ping() {
-    return Promise.resolve();
-  }
-
-  create(
-    modelClass: Class<Entity>,
-    entity: EntityData,
-    options?: Options,
-  ): Promise<EntityData> {
-    this.entities.push(entity);
-    return Promise.resolve(entity);
-  }
-
-  find(
-    modelClass: Class<Entity>,
-    filter?: Filter,
-    options?: Options,
-  ): Promise<EntityData[]> {
-    return Promise.resolve(this.entities);
-  }
-
-  updateAll(
-    modelClass: Class<Entity>,
-    data: EntityData,
-    where?: Where,
-    options?: Options,
-  ): Promise<Count> {
-    for (const p in data) {
-      for (const e of this.entities) {
-        (e as AnyObject)[p] = (data as AnyObject)[p];
-      }
-    }
-    return Promise.resolve({count: this.entities.length});
-  }
-
-  deleteAll(
-    modelClass: Class<Entity>,
-    where?: Where,
-    options?: Options,
-  ): Promise<Count> {
-    const items = this.entities.splice(0, this.entities.length);
-    return Promise.resolve({count: items.length});
-  }
-
-  count(
-    modelClass: Class<Entity>,
-    where?: Where,
-    options?: Options,
-  ): Promise<Count> {
-    return Promise.resolve({count: this.entities.length});
-  }
-}
+import {CrudConnectorStub} from '../crud-connector.stub';
 
 /**
  * A mock up model class
--- a/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -7,6 +7,7 @@ import {expect} from '@loopback/testlab';
 import {
   bindModel,
   DefaultCrudRepository,
+  DefaultTransactionalRepository,
   Entity,
   EntityNotFoundError,
   juggler,
@@ -14,6 +15,8 @@ import {
   ModelDefinition,
   property,
 } from '../../..';
+import {CrudConnectorStub} from '../crud-connector.stub';
+const TransactionClass = require('loopback-datasource-juggler').Transaction;
 
 describe('legacy loopback-datasource-juggler', () => {
   let ds: juggler.DataSource;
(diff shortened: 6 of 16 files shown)


packages/repository-tests/src/transaction/transactions.suite.ts
```diff
@@ -0,0 +1,80 @@
...
+        before(
+          withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+            repo = new repositoryClass(Product, ctx.dataSource);
+            await ctx.dataSource.automigrate(Product.name);
+          }),
+        );
+
+        it('retrieves a newly created model in a transaction', async () => {
+          const tx: juggler.Transaction = await repo.beginTransaction({
+            isolationLevel: 'READ COMMITTED',
```
raymondfeng: Do we have a constant/enum for the isolation level?

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -9,6 +9,7 @@ import {
   juggler,
   Options,
 } from '@loopback/repository';
+import {TransactionRepository} from '@loopback/repository/src';
```
raymondfeng: This import from `src` seems to be problematic.

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -38,6 +39,15 @@ export interface CrudConnectorFeatures {
...
   freeFormProperties: boolean;
+
+  /**
+   * Does the connector support using transactions for performing CRUD
+   * operations atomically and being able to commit or rollback the changes?
+   * SQL databases usually support transactions
+   *
+   * Default: `false`
+   */
+  supportsTransactions: boolean;
```
raymondfeng: `supportsTransaction`?

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -53,6 +63,19 @@ export type CrudRepositoryCtor = new <
   dataSource: juggler.DataSource,
 ) => EntityCrudRepository<T, ID, Relations>;
 
+/**
+ * A constructor of a class implementing CrudRepository interface,
+ * accepting the Entity class (constructor) and a dataSource instance.
+ */
+export type TransactionRepositoryCtor = new <
```
raymondfeng: `TransactionalRepositoryCtor`?

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -53,6 +63,19 @@ export type CrudRepositoryCtor = new <
...
+  T extends Entity,
+  ID,
+  Relations extends object
+>(
+  entityClass: typeof Entity & {prototype: T},
+  dataSource: juggler.DataSource,
+) => TransactionRepository<T, ID, Relations>;
```
raymondfeng: `TransactionalRepository`?

packages/repository/src/__tests__/unit/repositories/crud.repository.unit.ts
```diff
@@ -86,6 +85,10 @@ class CrudConnectorStub implements CrudConnector {
   ): Promise<Count> {
     return Promise.resolve({count: this.entities.length});
   }
+
+  beginTransaction(options: Options, cb: Callback) {
```
raymondfeng: Do we allow Promise here?

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -416,4 +418,21 @@ describe('DefaultCrudRepository', () => {
       'execute() must be implemented by the connector',
     );
   });
+
+  it('throws an error when beginTransaction() not implemented', async () => {
+    const repo = new DefaultCrudRepository(Note, ds);
+    await expect(repo.beginTransaction({})).to.be.rejectedWith(
+      'beginTransaction must be function implemented by the connector',
```
raymondfeng: `beginTransaction must be implemented`?

packages/repository/src/transaction.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Local transaction
+ */
+export interface Transaction {
+  /**
+   * Commit the transaction
+   * @param callback
+   */
+  commit(callback?: Callback<any>): void;
```
raymondfeng: `commit(callback?: Callback<any>): ValueOrPromise<void>`?

acceptance/repository-mysql/package-lock.json
```diff
@@ -1,630 +0,0 @@
-{
-	"name": "@loopback/test-repository-mysql",
```
bajtos: It looks like you are trying to remove this package-lock file entirely. Why?

acceptance/repository-mysql/src/__tests__/mysql.transaction.acceptance.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import {DefaultCrudRepository} from '@loopback/repository';
+import {
+  CrudRepositoryCtor,
+  transactionTestSuite,
+} from '@loopback/repository-tests';
+import {MYSQL_CONFIG, MYSQL_FEATURES} from './mysql.datasource';
+
+describe('DefaultCrudRepository + mysql connector', () => {
+  transactionTestSuite(
```
bajtos: This is not right. `transactionTestSuite` should be executed as part of running `crudRepositoryTestSuite`, which is already called from [`acceptance/repository-mysql/src/__tests__/mysql-default-repository.acceptance.ts`](https://github.com/strongloop/loopback-next/tree/master/acceptance/repository-mysql/src/__tests__/mysql-default-repository.acceptance.ts) See also [`packages/repository-tests/src/crud-test-suite.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository-tests/src/crud-test-suite.ts) which is iterating over all files and automatically running the test suite(s) exported by them.

How to address this issue:
- Remove this file (`mysql.transaction.acceptance.ts`)
- Verify that the transaction-related tests are still executed for MySQL

packages/repository-tests/package-lock.json
```diff
@@ -1,32 +0,0 @@
-{
-	"name": "@loopback/repository-tests",
```
bajtos: Ditto, why are you removing this package-lock file?

packages/repository-tests/package.json
```diff
@@ -13,7 +13,7 @@
     "build": "lb-tsc",
     "clean": "lb-clean loopback-repository-tests*.tgz dist tsconfig.build.tsbuildinfo package",
     "pretest": "npm run build",
-    "test": "lb-mocha \"dist/__tests__/**/*.js\"",
+    "test": "lb-mocha \"dist/__tests__/**/*.js\" --reporter=spec",
```
bajtos: Please revert this change. You can temporarily enable `spec` reporter as follows:

```
$ npm test -- --reporter=spec
```

packages/repository-tests/src/index.ts
```diff
@@ -4,6 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 export * from './crud-test-suite';
+export * from './transaction-test-suite';
```
bajtos: Transactions are CRUD specific, these tests are not going to work for KeyValue repositories.

Please move the new test file `transaction-test-suite` to `packages/repository-tests/src/crud` directory, where it will be automatically picked up by our infrastructure.

It looks like our developer documentation is insufficiently describing how to add new suites, please improve it to help the next person: https://github.com/strongloop/loopback-next/tree/master/packages/repository-tests#developer-guide

Please sync with @hacksparrow, he is working on this part of our docs too.

packages/repository-tests/src/transaction/transactions.suite.ts
```diff
@@ -0,0 +1,80 @@
...
+import {withCrudCtx} from '../helpers.repository-tests';
+import {
+  CrudConnectorFeatures,
+  CrudTestContext,
+  DataSourceOptions,
+  TransactionRepositoryCtor,
+} from '../types.repository-tests';
+
+// Core scenarios around creating new model instances and retrieving them back
+// Please keep this file short, put any advanced scenarios to other files
```
bajtos: Please update the comment to match the intent of this file.

packages/repository-tests/src/transaction/transactions.suite.ts
```diff
@@ -0,0 +1,80 @@
...
+          });
+          const created = await repo.create(
+            {name: 'Pencil'},
+            {transaction: tx},
+          );
+          expect(created.toObject()).to.have.properties('id', 'name');
+          expect(created.id).to.be.ok();
+
+          // const found = await repo.findById(created.id);
+          await tx.commit();
```
bajtos: How do you envision this test to fail when the database and/or the connector does not support transactions, or implement them incorrectly?

It seems to me that this test only verifies that create+retrieve preserves the data when both commands are executed in the same transaction. It will pass for any repository that implements `beginTransaction` API, regardless of whether it actually enforces the transaction or not.

Let's find more robust test scenarios please.

Here are few scenarios that come to my mind:

1. Create a transaction, `create` a new model from that transaction, COMMIT the transaction, `findById` the created model outside of the transaction - model should be found. This is a kind of a smoke test, it passes even if transaction isolation is not implemented.

2. Create a transaction, `create` a new model from that transaction, ROLLBACK the transaction, `findById` the created model outside of the transaction - no model should be returned. This verifies rollback feature, it does not require transactions to be isolated.

3. Create a transaction, `create` a new model from that transaction. `findById` without any transaction - no model should be returned. `findById` from the same transaction - the created model should be returned. This verifies isolation of individual transactions.

packages/repository-tests/src/types.repository-tests.ts
bajtos: +1

```suggestion
import {TransactionRepository} from '@loopback/repository';
```

packages/repository-tests/src/types.repository-tests.ts
bajtos: I personally prefer `supportsTransactions` (plural form).

Singular form reads weird to me - is it "supports _a_ transaction" or "supports _the_ transaction"? Which particular transaction do we have in mind?

packages/repository-tests/src/types.repository-tests.ts
bajtos: +1 for the name proposed by Raymond.

packages/repository-tests/src/types.repository-tests.ts
bajtos: +1 for `TransactionalRepository`

packages/repository/package-lock.json
```diff
@@ -4,23 +4,499 @@
...
+			"dependencies": {
+				"@hapi/hoek": {
+					"version": "8.0.2",
+					"resolved": "https://registry.npmjs.org/@hapi/hoek/-/hoek-8.0.2.tgz",
+					"integrity": "sha512-O6o6mrV4P65vVccxymuruucb+GhP2zl9NLCG8OdoFRS8BEGw3vwpPp20wpAtpbQQxz1CEUtmxJGgWhjq1XA3qw==",
+					"dev": true
+				}
+			}
+		},
+		"@loopback/build": {
```
bajtos: `@loopback/build` is a dependency that should be resolved within the monorepo. Please fix.

packages/repository/src/__tests__/unit/repositories/crud.repository.unit.ts
```diff
@@ -4,26 +4,25 @@
...
-  AnyObject,
-  Count,
+  Where,
 } from '../../..';
-
 /**
  * A mock up connector implementation
  */
-class CrudConnectorStub implements CrudConnector {
+export class CrudConnectorStub implements CrudConnector {
```
bajtos: Please move this stub class into some kind of a helper file, so that tests don't import from each other. I think these stub should be used by unit-level tests only, integration-level and acceptance tests should use real connectors.

Proposed file name: `packages/repository/src/__tests__/unit/crud-connector.stub.ts`

packages/repository/src/__tests__/unit/repositories/crud.repository.unit.ts
bajtos: Not yet, connectors must use callbacks now. See https://github.com/strongloop/loopback-datasource-juggler/issues/1659

Let's add a code comment here to explain & include a link to that GH issue.

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -416,4 +418,21 @@ describe('DefaultCrudRepository', () => {
...
+    );
+  });
+  it('calls connector beginTransaction() when available', async () => {
+    const crudDs = new juggler.DataSource({
+      name: 'db',
+      connector: CrudConnectorStub,
+    });
+
+    const repo = new DefaultCrudRepository(Note, crudDs);
+    const res = await repo.beginTransaction({});
```
bajtos: If we are accepting an empty array then let's make the argument optional please.

```suggestion
    const res = await repo.beginTransaction();
```

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
bajtos: How about this:

```suggestion
      'Cannot start a new transaction - "memory" connector does not support transactions.'
```

(Replace `memory` with the actual connector name.)

packages/repository/src/transaction.ts
bajtos: Let's not introduce callbacks to LB4 APIs please. I believe we have already added support for promises to `Transaction` members in loopback-connector - see https://github.com/strongloop/loopback-connector/pull/147.

IMO, we should use the following form:

```suggestion
  commit(): Promise<void>;
```

Regarding `commit(callback?: Callback<any>): ValueOrPromise<void>`: this does not offer enough type safety. It allows callers to provide a callback and await the returned value at the same time. Here is a more correct type definition:

```ts
commit(callback: Callback<any>): void;
commit(): Promise<void>;
```

packages/repository/src/transaction.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Rollback the transaction
+   * @param callback
+   */
+  rollback(callback?: Callback<any>): void;
```
bajtos:
```suggestion
  rollback(): Promise<void>;
```

b-admike:
Thank you for your thorough and thoughtful reviews @bajtos @raymondfeng 🙇. I have applied most, if not, all your feedback.

packages/repository-tests/src/types.repository-tests.ts
b-admike: I've left it as `supportsTransactions` as in the transaction feature set. I don't mind changing it if we feel strongly about `supportsTransaction`.

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
b-admike: This is coming from `loopback-connector`. See https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L96-L97. Perhaps we can improve the message there?

packages/repository/src/transaction.ts
b-admike: Since we just want Promise style support in LB4, I've only included `commit(): Promise<void>;` signature for the methods.

b-admike:
> I think I may get confused about your intention regarding how to structure the test suite.
> 
> My expectation was that you will simply add or more new files to [packages/repository-tests/src/crud](https://github.com/strongloop/loopback-next/tree/master/packages/repository-tests/src/crud), because Transactions are specific to Crud repository interface - we don't support them for KeyValue repositories.
> 
> I see that you are creating a whole new transaction-only test suite. What's your reasoning for that? What benefits do you see in that approach?
> 
> I am concerned that if the transaction test suite requires a new test file in each per-connector acceptance package, it is difficult to discover by connector authors and makes it easy to forget to run it for certain connectors.
> 
> So far, my vision was to add new tests to the single CRUD suite, so that connector authors are notified about new features required by our ORM via a new test that may start failing for their connector.
> 
> Let's discuss!
> 
> The comments below are based on the assumption that transactions should be tested as part of our CRUD test suite.

I was doing it wrong :-). Thank you for the pointers, I agree with your POV.

packages/repository-tests/src/types.repository-tests.ts
raymondfeng: I'm good with `supportsTransactions`.

bajtos: On the second thought, I would use just `transactions`. The structure is called `CrudFeatures`, it describes what the connector and/or the repository supports - no need to repeat `support` in the flag name again.

Not a big deal though.

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
bajtos: Yes please, let's improve the message there. Ideally, the error object should have also a `code` property set to a reasonably-distinct string, e.g. `'TRANSACTIONS_NOT_SUPPORTED'`. That will allow us to avoid brittle string check and verify just the error `code` in this test.

Anyhow, I am fine to leave this improvement out of scope of this pull request if you prefer.

acceptance/repository-mysql/package-lock.json
```diff
@@ -252,9 +252,9 @@
 			}
 		},
 		"lodash": {
-			"version": "4.17.15",
-			"resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
-			"integrity": "sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A==",
+			"version": "4.17.14",
+			"resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.14.tgz",
+			"integrity": "sha512-mmKYbW3GLuJeX+iGP+Y7Gp1AiGHGbXHCOh/jZmrawMmsE7MS4znI3RL2FsjbqOyMayHInjOeykW7PEajUk1/xw==",
```
bajtos: Would you mind reverting this change? I prefer (deep) dependencies in package-lock files to be updated independently when they are not related to other changes made in the pull request.

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+Here is an example:
+
+```typescript
+import {
+  Transaction,
+  DefaultTransactionalRepository,
+  IsolationLevel,
+} from '@loopback/repository';
+// assuming there is a Note model extending Entity class, and
+// crudDs datasource which is backed by a transaction enabled
```
bajtos: `crudDs` looks like a typo to me. On the third read of this line, it occurred to me that it may be a variable name "crud Dst". Let's use `db`, `ds` or `crudDataSource` instead.

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+// connector
+const repo = new DefaultTransactionalRepository(Note, crudDs);
+// Now we have a transaction (tx)
+const tx: Transaction = await repo.beginTransaction({
```
bajtos: Is the explicit type required? If it's your intention to make the type of `tx` explicitly clear to readers, then I am fine with that.

Otherwise the following code should work equally well as far as the compiler is concerned:

```suggestion
const tx = await repo.beginTransaction({
```

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+
+- [MySQL SET TRANSACTION Syntax](https://dev.mysql.com/doc/refman/5.7/en/set-transaction.html)
+- [Oracle Isolation Levels](http://docs.oracle.com/cd/B14117_01/server.101/b10743/consist.htm#i17856)
+- [PostgreSQL Transaction Isolation](http://www.postgresql.org/docs/9.4/static/transaction-iso.html)
+- [SQL Server SET TRANSACTION ISOLATION LEVEL](https://msdn.microsoft.com/en-us/library/ms173763.aspx)
+
+### Perform operations in a transaction
+
+To perform create, retrieve, update, and delete operations in the transaction,
+add a second argument consisting of the transaction object to the standard
```
bajtos: `options` is not the second argument of all CRUD methods, e.g. it's the third argument for `findById(id, filter, options)`.

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+   */
+  rollback(): Promise<void>;
+```
+
+## Set up timeout
+
+You can specify a timeout (in milliseconds) to begin a transaction. If a
+transaction is not finished (committed or rolled back) before the timeout, it
+will be automatically rolled back upon timeout by default. The timeout event can
+be trapped using the timeout hook.
```
bajtos: > The timeout event can be trapped using the timeout hook.

This may need more explanation - what is the "timeout hook" and how can it be used?

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -0,0 +1,133 @@
...
+
+      describe('create-retrieve with transactions', () => {
+        let repo: TransactionalRepository<Product, typeof Product.prototype.id>;
+        before(
+          withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+            repo = new repositoryClass(Product, ctx.dataSource);
+            await ctx.dataSource.automigrate(Product.name);
+          }),
+        );
+
```
bajtos: What happens when a test opens a transaction and aborts on an assertion error before it has a chance to commit or reject the transaction?

Let's make the tests more robust to ensure they always close the transactions, even on test failure.

- Add a new `let tx: Transaction` variable.
- Modify tests to store the transaction to this variable, instead of creating a new test-local `const tx`.
- Add an `beforeEach` hook to set `tx` to `undefined`
- Add an `afterEach` hook to call `tx.rollback()` if `tx` is still set
- Figure out how to handle the case where the test has committed or rolled back a transaction. To begin with, we can change the tests to set `tx = undefined` after it has been committed or rolled back. A better solution would be to allow the `afterEach` hook to find out whether the transaction is still active, and do nothing if it is not. Since this would require changes in juggler and/or loopback-connector, I am fine to leave such improvement out of scope of this PR.

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -0,0 +1,133 @@
...
+        it('retrieves model instance once transaction is committed', async () => {
+          const tx: juggler.Transaction = await repo.beginTransaction({
```
bajtos: We shouldn't be using `juggler` types here, let's use `Transaction` from `@loopback/repository` or don't specify the type at all.

```suggestion
          const tx = await repo.beginTransaction({
```

Same comment applies to all other tests too.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -461,3 +464,23 @@ export class DefaultCrudRepository<
...
   }
 }
+
+/**
+ * Default implementation of CRUD repository using legacy juggler model
+ * and data source with beginTransaction() method for connectors which
+ * support Transactions
+ */
+
+export class DefaultTransactionalRepository<
```
bajtos: Is it necessary to introduce a new repository class? Can we change `DefaultCrudRepository` class to implement `TransactionalRepository` too?

I can see benefits of having `DefaultCrudRepository` and `DefaultTransactionalRepository`, but it also have downsides.

@raymondfeng WDYT?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -461,3 +464,23 @@ export class DefaultCrudRepository<
...
+  T extends Entity,
+  ID,
+  Relations extends object = {}
+> extends DefaultCrudRepository<T, ID, Relations>
+  implements TransactionalRepository<T, ID, Relations> {
+  async beginTransaction(
+    options?: IsolationLevel | Options,
+  ): Promise<Transaction> {
+    const dsOptions: juggler.IsolationLevel | Options = options || {};
+    return (await this.dataSource.beginTransaction(dsOptions)) as Transaction;
```
bajtos: I find this suspicious, why is the type cast from `juggler.Transaction` to `Transaction` necessary? I expected those interfaces to be the same, what am I missing?

packages/repository/src/repositories/repository.ts
```diff
@@ -37,6 +38,20 @@ export interface ExecutableRepository<T extends Model> extends Repository<T> {
   ): Promise<AnyObject>;
 }
 
+export interface TransactionalRepository<
+  T extends Entity,
+  ID,
+  Relations extends object = {}
+> extends EntityCrudRepository<T, ID, Relations> {
```
bajtos: Discussion point: Should `TransactionalRepository` extend `Repository` or `EntityCrudRepository`?

Personally, I am inclining towards `TransactionalRepository extends Repository`, because it's possible to have transactions in repositories that are backing models with no primary keys (i.e. `CrudRepository`).

@b-admike @raymondfeng thoughts?

docs/site/Transactions.md
b-admike: Yeah it was just to make it explicit. I don't mind removing it too.

packages/repository/src/repositories/repository.ts
b-admike: Yea that sounds good to me too. I am reworking it to extend from `Repository`

packages/repository/src/repositories/legacy-juggler-bridge.ts
b-admike: Yea that was my expectation as well, but the `juggler.Transaction` still have the Promise/Callback variants of the Transaction methods (e.g. https://github.com/strongloop/loopback-datasource-juggler/blob/master/types/transaction-mixin.d.ts#L43-L47).

packages/repository/src/repositories/legacy-juggler-bridge.ts
b-admike: I am also okay with either approach, but this felt cleaner as I was writing the docs. Later on, we can decide to add it to our CLI templates too.

docs/site/Transactions.md
b-admike: Actually I used LB3 docs as a base when writing the docs. After looking at our `TransactionMixin` class in juggler and comparing it to the `Transaction` class in `loopback-connector`, we still have support for `timeout`, but won't have support for hooks on transactions for observing events before `commit`, `rollback`, or `timeout`. Thus, I'll remove this bit.

packages/repository-tests/src/crud/transactions.suite.ts
b-admike: Sounds great, thank you @bajtos!

b-admike: >Figure out how to handle the case where the test has committed or rolled back a transaction. To begin with, we can change the tests to set tx = undefined after it has been committed or rolled back. A better solution would be to allow the afterEach hook to find out whether the transaction is still active, and do nothing if it is not. Since this would require changes in juggler and/or loopback-connector, I am fine to leave such improvement out of scope of this PR.

I added two more properties to the `Transaction` interface, one being `id` and the other `connection`. I don't have any good typing info for `connection` so I've left it as `any`. That way, if there is no connection, we know it's not active. On the other hand, I can add `ensureActive` method from https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/transaction.js#L197 to the type-def for `Transaction` and use that instead. Thoughts?

packages/repository-tests/src/types.repository-tests.ts
b-admike: Okay leaving it as is for now :).

docs/site/Transactions.md
```diff
@@ -0,0 +1,199 @@
...
+---
+
+## Overview
+
+A *transaction* is a sequence of data operations performed as a single logical
+unit of work. Many relational databases support transactions to help enforce
+data consistency and business logic requirements.
+
+A repository can perform operations in a transaction when the backing datasource
+is attached to one of the following connectors:
```
raymondfeng: Let's be clear that we only support database local transactions - only operations against the same transaction-capable datasource can be grouped into a transaction.

docs/site/Transactions.md
```diff
@@ -0,0 +1,199 @@
...
+  Transaction,
+  DefaultTransactionalRepository,
+  IsolationLevel,
+} from '@loopback/repository';
+// assuming there is a Note model extending Entity class, and
+// ds datasource which is backed by a transaction enabled
+// connector
+const repo = new DefaultTransactionalRepository(Note, ds);
+// Now we have a transaction (tx)
+const tx: Transaction = await repo.beginTransaction({
```
raymondfeng: I assume `Transaction` can be inferred from the `beginTransaction` method return type.

raymondfeng:
@b-admike CI fails due to the following:
```
lerna ERR! npm ci exited 1 in '@loopback/test-repository-mysql'
lerna ERR! npm ci stderr:
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! 
npm ERR! Invalid: lock file's loopback-connector-mysql@5.4.1 does not satisfy loopback-connector-mysql@^5.4.2
npm ERR! 
```

packages/repository-tests/src/crud/transactions.suite.ts
bajtos: `ensureActive` does not seem very useful for our use case - we don't want to throw & catch an error in order to detect whether a transaction is active.

Personally, I'd prefer the transaction to provide `isActive` property that returns a boolean.

A mock-up:

```ts
Object.defineProperty(Transaction.prototype, 'isActive', {
  get: function() { return !!this.connection }
}
```

(The code needs to be added to loopback-connector, then we need to add `isActive` property to type defs both in loopback-connector and LB4.)

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: Makes sense. Can you please add a code comment to explain why the cast is needed?

BTW I think you posted a link pointing to a wrong place, did you mean this insted?

https://github.com/strongloop/loopback-datasource-juggler/blob/ad4edc6a1569eace4876c2c3e5b32e4661d4d6eb/types/transaction-mixin.d.ts#L11-L22

acceptance/repository-mysql/package-lock.json
bajtos: You are effectively downgrading `lodash` to an OLDER version, similarly for `loopback-connector-mysql` later in this file.

Please revert all changes in all `package-lock.json` files.

packages/repository-tests/src/crud-test-suite.ts
```diff
@@ -32,6 +32,7 @@ export function crudRepositoryTestSuite(
     idType: 'string',
     freeFormProperties: true,
     emptyValue: undefined,
+    supportsTransactions: false,
```
bajtos: New feature flags should default to `true`. That way we include the new tests in all downstream repos (e.g. `repository-mysql`), which makes it much easier to discover a database/connector that does not support the new feature.

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -0,0 +1,146 @@
...
+            {},
+            {
+              transaction: tx,
+            },
+          );
+          expect(toJSON(created)).to.deepEqual(toJSON(foundInsideTransaction));
+          await tx.rollback();
+          await expect(repo.findById(created.id, {})).to.be.rejectedWith({
+            code: 'ENTITY_NOT_FOUND',
+            message: /Entity not found/,
```
bajtos: It's enough to assert on the `code`. The message can change in the future, we don't want this test to fail if that happens. Please remove this line.

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -411,9 +414,26 @@ describe('DefaultCrudRepository', () => {
   });
 
   it(`throws error when execute() not implemented by ds connector`, async () => {
-    const repo = new DefaultCrudRepository(Note, ds);
+    const repo = new DefaultTransactionalRepository(Note, ds);
```
bajtos: I don't see why is this change necessary - how is `execute` related to transactions? Can you revert this line please?

packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
```diff
@@ -411,9 +414,26 @@ describe('DefaultCrudRepository', () => {
...
     await expect(repo.execute('query', [])).to.be.rejectedWith(
       'execute() must be implemented by the connector',
     );
   });
+
+  it('throws an error when beginTransaction() not implemented', async () => {
```
bajtos: These new tests are inside `describe('DefaultCrudRepository')`, but they are testing `DefaultTransactionalRepository` class instead.

Please move them to a new describe block.

```ts
describe('DefaultCrudRepository', () => {
  // existing test suite
});

describe('DefaultTransactionalRepository' => {
 // your new tests calling `repo.beginTransaction'
});
```

packages/repository/src/repositories/repository.ts
```diff
@@ -37,6 +38,22 @@ export interface ExecutableRepository<T extends Model> extends Repository<T> {
   ): Promise<AnyObject>;
 }
 
+export type TransactionalEntityRepository<
+  T extends Entity,
+  ID,
+  Relations extends object = {}
+> = TransactionalRepository<T> & EntityCrudRepository<T, ID>;
+export interface TransactionalRepository<T extends Entity>
```
bajtos: Please add an empty line to visually delimit the definition of `TransactionalEntityRepository` and `TransactionalRepository`.

Please add a tsdoc comment for each type.

packages/repository/src/repositories/repository.ts
```diff
@@ -37,6 +38,22 @@ export interface ExecutableRepository<T extends Model> extends Repository<T> {
...
+  extends Repository<T> {
+  /**
+   * Begin a new Transaction
+   * @param options - Options for the operations
+   * @returns Promise<true> if an entity exists for the id, otherwise
```
bajtos: The docs says that the functions returns a promise of a boolean, but the actual code uses `Promise<Transaction>`. Please fix the docs.

packages/repository/src/transaction.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * The transaction Identifier
+   */
+  id: string;
+
+  /**
+   * The connection used by the transaction
+   */
+  connection: any;
```
bajtos: Let's use `unknown` instead of any. That way consumers of this property must explicitly cast it to a type they are assuming to see there, instead of being able to treat it as literally anything and thus bypassing compiler checks.

```suggestion
  connection: unknown;
```

Personally, I'd prefer to keep `connection` as an internal implementation detail and DO NOT expose it in this interface. Instead, expose `isActive` property as discussed earlier.

AFAICT, your current test suite does not need the check whether the connection is active, thus I think it will be best to leave both `connection` and `isActive` properties out of scope of this pull request.

docs/site/Transactions.md
b-admike: It was to be explicit, but I will remove that.

packages/repository/src/repositories/legacy-juggler-bridge.ts
b-admike: Yeah I meant to post that sorry. Will do 👍

b-admike:
> You are effectively downgrading lodash to an OLDER version, similarly for loopback-connector-mysql later in this file.

>Please revert all changes in all package-lock.json files.

Yeah I am not sure how I'm dealing with `package-lock.json` files. I always choose the versions in the latest master branch when rebasing. Maybe I should keep the files generated in my branch.

packages/repository/src/transaction.ts
b-admike: Yea I agree that we can revisit this in another PR (I like the idea in https://github.com/strongloop/loopback-next/pull/3397#discussion_r308605303). But the `afterEach()` hook inside `transactions.suite.ts` would need this to function properly (to check active transactions) at the moment.

bajtos: > But the `afterEach()` hook inside `transactions.suite.ts` would need this to function properly (to check active transactions) at the moment.

Ah, I see.

As a temporary solution, can we perhaps disable type checks in that hook? Something along the following lines:

```ts
afterEach(async () => {
  // FIXME: replace tx.connection with tx.isActive when it become available
  if (tx !== undefined && (tx as any).connection) {
    await tx.rollback();
  }
});
```

Thoughts?

b-admike: Nice! I'll try that :D.

b-admike: I've created https://github.com/strongloop/loopback-next/issues/3471 to track this issue.

packages/repository-tests/src/crud/transactions.suite.ts
b-admike: Created #3471

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+
+Performing operations in a transaction typically involves the following steps:
+
+- Start a new transaction.
+- Perform create, read, update, and delete operations in the transaction.
+- Commit or rollback the transaction.
+
+### Start transaction
+
+Use the `beginTransaction()` method to start a new transaction.
```
raymondfeng: Do we allow `dataSource.beginTransaction()`? I don't see any examples for that.

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+  IsolationLevel,
+} from '@loopback/repository';
+// assuming there is a Note model extending Entity class, and
+// ds datasource which is backed by a transaction enabled
+// connector
+const repo = new DefaultTransactionalRepository(Note, ds);
+// Now we have a transaction (tx)
+const tx = await repo.beginTransaction({
+  isolationLevel: IsolationLevel.READ_COMMITTED,
+});
```
raymondfeng: Do we want to simplify it as follows?
```ts
const tx = await repo.beginTransaction(IsolationLevel.READ_COMMITTED);
```

docs/site/Transactions.md
```diff
@@ -0,0 +1,201 @@
...
+);
+
+// commit the transaction to persist the changes
+await tx.commit();
+```
+
+Propagating a transaction is explicit by passing the transaction object via the
+options argument for all create, retrieve, update, and delete and relation
+methods.
+
```
raymondfeng: What's going to happen if the `transaction` is passed to another repository that has a different datasource?

packages/repository/src/repositories/legacy-juggler-bridge.ts
raymondfeng: Should it be the following now?
```ts
return this.dataSource.beginTransaction(dsOptions) as Promise<Transaction>;
```

docs/site/Transactions.md
b-admike: I see that as an internal API because we have repository level `beginTransaction` which calls it. I guess users can technically do `repo.dataSource.beginTransaction()` too.

docs/site/Transactions.md
b-admike: Added the test and as discussed, created https://github.com/strongloop/loopback-next/issues/3483.

b-admike:
I think the coverage drop is because we don't run the transaction tests for all the downstream dependencies i.e. `acceptance/repository-mongodb` doesn't run them whereas `acceptance/repository-mysql` does:
```
COMMITTED 31 JUL 2019 - 17:38 COVERAGE DECREASED (-0.3%) TO 91.401%
```

docs/site/Transactions.md
```diff
@@ -0,0 +1,200 @@
...
+- Commit or rollback the transaction.
+
+### Start transaction
+
+Use the `beginTransaction()` method to start a new transaction from a repository
+class using `DefaultTransactionalRepository` as a base class.
+
+Here is an example:
+
+```typescript
```
raymondfeng: We use ```ts.

Issue #2654, opened 2019-03-28
Spike: simplify requestBody annotation with schema options


bajtos:
This is a follow-up for the discussion started in https://github.com/strongloop/loopback-next/issues/2082, https://github.com/strongloop/loopback-next/issues/2152 and the related issues.

In #2631, we are introducing a new function `getModelSchemaRef` that can be used to obtain OpenAPI schema for a model in a configurable way. Example use:

```ts
class TodoListController {
  // ...

  @patch('/todo-lists/{id}', {
    responses: {
      // left out for brevity
    },
  })
  async updateById(
    @param.path.number('id') id: number,
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(TodoList, {partial: true}),
          /***** ^^^ THIS IS IMPORTANT - OPENAPI SCHEMA ^^^ ****/
       },
     },
    })
    obj: Partial<TodoList>,
    /***** ^^^ THIS IS IMPORTANT - TYPESCRIPT TYPE ^^^ ****/
  ): Promise<void> {
    await this.todoListRepository.updateById(id, obj);
  }
}
```

Eventually, we plan to implement the following schema generator options:
- `includeRelations` in #2629
- `partial` in #2652 
- `exclude` in #2653

The current approach is very verbose. Compare the example above with a controller method that uses default options:

```ts
class TodoController {
  // ...

  @put('/todos/{id}', {
    responses: {
       // left out for brevity
    },
  })
  async replaceTodo(
    @param.path.number('id') id: number,
    @requestBody() todo: Todo,
  ): Promise<void> {
    await this.todoRepo.replaceById(id, todo);
  }
}
```

In this spike, we should research different options for simplifying request-body definitions requiring custom schema options.

Few examples to get started:

- In https://github.com/strongloop/loopback-next/issues/1722#issuecomment-422439405, @raymondfeng proposed:

    ```ts
    {
      schema: {
        'x-ts-type': Order,
        'x-ts-type-options': {
          partial: true,
          excludes: []
          }
      }
    ```

    This addresses only how to provide schema options when using `x-ts-type` extension. It does not address the problem of too complex spec required by `@requestBody` decorator.

- In https://github.com/strongloop/loopback-next/issues/1722#issuecomment-423065901, @bajtos proposed a different solution:

    ```ts
    class TodoController {
        // ...
    
        // `id` property is not allowed for `create`:
        @post('/')
[...]

PR #3398, opened 2019-07-19
[SPIKE] feat: simplify request body decorator

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/2654

## Checklist

### DIFF
--- a/packages/openapi-v3/spike.md
+++ b/packages/openapi-v3/spike.md
@@ -8,7 +8,23 @@ The current @requestBody() can only
 or
 - generate the schema inferred from the parameter type
 
-To simplify the signature, this spike PR introduces a 2nd parameter `schemaOptions` to configure the schema. The new decorator `newRequestBody1` is written in file 'request-body.options1.decorator.ts'
+To simplify the signature, this spike PR introduces two more parameters `modelCtor` and `schemaOptions` to configure the schema. The new decorator `requestBody2()`(let's discuss a better name later, see section [Naming](#Naming) is written in file 'request-body.spike.decorator.ts'
+
+### Signature
+
+The new decorator's signature  is `@requestBody2(spec, modelCtor, schemaOptions)`.
+
+```ts
+export function requestBody2(
+  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
+  modelOrOptions?: Function | SchemaOptions,
+  schemaOptions?: SchemaOptions
+) {
+  // implementation
+}
+```
+
+All the 3 parameters are optional, therefore in the PoC, the real implementation are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different combination of the parameters.
 
 ### Create - exclude properties
 
@@ -23,11 +39,6 @@ const requestBodySpec = {
 
 // Provide the options that configure your schema
 const excludeOptions = {
-  // Using advanced ts types like `Exclude<>`, `Partial<>` results in
-  // `MetadataInspector.getDesignTypeForMethod(target, member)` only
-  // returns `Object` as the param type, which loses the model type's info.
-  // Therefore user must provide the model type in options.
-  [TS_TYPE_KEY]: Product,
   // Make sure you give the custom schema a unique schema name,
   // this name will be used as the reference name
   // like `$ref: '#components/schemas/ProductWithoutID'`
@@ -41,6 +52,11 @@ class MyController1 {
   @post('/Product')
   create(@newRequestBody1(
     requestBodySpec,
+    // Using advanced ts types like `Exclude<>`, `Partial<>` results in
+    // `MetadataInspector.getDesignTypeForMethod(target, member)` only
+    // returns `Object` as the param type, which loses the model type's info.
+    // Therefore user must provide the model type in options.
+    Product
     excludeOptions
   ) product: Exclude<Product, ['id']>) { }
 }
@@ -55,8 +71,6 @@ const requestSpecForUpdate = {
 };
 
 const partialOptions = {
-  [TS_TYPE_KEY]: Product,
-  schemaName: 'PartialProduct',
   partial: true
 }
 
@@ -64,8 +78,12 @@ class MyController2 {
   @put('/Product')
   update(@newRequestBody1(
     requestSpecForUpdate,
+    Product
     partialOptions
   ) product: Partial<Product>) { }
 }
 ```
 
+## Naming
+
+From @jannyHou: I think we can keep the name `requestBody` unchanged. I enabled the existing `@requestBody()`'s tests but applied `requestBody2()` decorator, all tests pass, which means there is no breaking change.
--- a/packages/openapi-v3/spike.md
+++ b/packages/openapi-v3/spike.md
@@ -1,0 +1,71 @@
+## Improve the UX of @requestBody()
+
+The original discussion is tracked in issue [Spike: simplify requestBody annotation with schema options](https://github.com/strongloop/loopback-next/issues/2654).
+
+The current @requestBody() can only
+
+- takes in an entire request body specification with very nested media type objects
+or
+- generate the schema inferred from the parameter type
+
+To simplify the signature, this spike PR introduces a 2nd parameter `schemaOptions` to configure the schema. The new decorator `newRequestBody1` is written in file 'request-body.options1.decorator.ts'
+
+### Create - exclude properties
+
+Take "Create a new product with excluded properties" as an example:
+
+```ts
+// Provide the description as before
+const requestBodySpec = {
+  description: 'Create a product',
+  required: true,
+};
+
+// Provide the options that configure your schema
+const excludeOptions = {
+  // Using advanced ts types like `Exclude<>`, `Partial<>` results in
+  // `MetadataInspector.getDesignTypeForMethod(target, member)` only
+  // returns `Object` as the param type, which loses the model type's info.
+  // Therefore user must provide the model type in options.
+  [TS_TYPE_KEY]: Product,
+  // Make sure you give the custom schema a unique schema name,
+  // this name will be used as the reference name
+  // like `$ref: '#components/schemas/ProductWithoutID'`
+  schemaName: 'ProductWithoutID',
+  // The excluded properties
+  exclude: ['id']
+}
+
+// The decorator takes in the option without having a nested content object
+class MyController1 {
+  @post('/Product')
+  create(@newRequestBody1(
+    requestBodySpec,
+    excludeOptions
+  ) product: Exclude<Product, ['id']>) { }
+}
+```
+
+### Update - partial properties
+
+```ts
+const requestSpecForUpdate = {
+  description: 'Update a product',
+  required: true,
+};
+
+const partialOptions = {
+  [TS_TYPE_KEY]: Product,
+  schemaName: 'PartialProduct',
+  partial: true
+}
+
+class MyController2 {
+  @put('/Product')
+  update(@newRequestBody1(
+    requestSpecForUpdate,
+    partialOptions
+  ) product: Partial<Product>) { }
+}
+```
+
--- a/packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
@@ -3,11 +3,11 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {post, requestBody, getControllerSpec} from '../../../..';
-import {expect} from '@loopback/testlab';
-import {model, property} from '@loopback/repository';
+import { model, property } from '@loopback/repository';
+import { expect } from '@loopback/testlab';
+import { getControllerSpec, post, requestBody2 } from '../../../..';
 
-describe('requestBody decorator', () => {
+describe.only('requestBody decorator', () => {
   context('can build a correct "RequestBody" spec and', () => {
     it('persists "description" and "required" into the generated schema', () => {
       const requestSpec = {
@@ -16,7 +16,7 @@ describe('requestBody decorator', () => {
       };
       class MyController {
         @post('/greeting')
-        greet(@requestBody(requestSpec) name: string) {}
+        greet(@requestBody2(requestSpec) name: string) { }
       }
 
       const requestBodySpec = getControllerSpec(MyController).paths[
@@ -35,7 +35,7 @@ describe('requestBody decorator', () => {
       };
       class MyController {
         @post('/greeting')
-        greet(@requestBody(requestSpec) name: string) {}
+        greet(@requestBody2(requestSpec) name: string) { }
       }
 
       const requestBodySpec = getControllerSpec(MyController).paths[
@@ -47,7 +47,7 @@ describe('requestBody decorator', () => {
     it('infers request body with complex type', () => {
       const expectedContent = {
         'application/text': {
-          schema: {$ref: '#/components/schemas/MyModel'},
+          schema: { $ref: '#/components/schemas/MyModel' },
         },
       };
 
@@ -60,8 +60,8 @@ describe('requestBody decorator', () => {
       class MyController {
         @post('/MyModel')
         createMyModel(
-          @requestBody({content: {'application/text': {}}}) inst: MyModel,
-        ) {}
+          @requestBody2({ content: { 'application/text': {} } }) inst: MyModel,
+        ) { }
       }
 
       const requestBodySpec = getControllerSpec(MyController).paths['/MyModel'][
@@ -73,15 +73,15 @@ describe('requestBody decorator', () => {
     it('preserves user-provided schema in requestBody', () => {
       const expectedContent = {
         'application/json': {
-          schema: {type: 'object'},
+          schema: { type: 'object' },
         },
       };
 
-      class MyModel {}
+      class MyModel { }
 
       class MyController {
         @post('/MyModel')
-        createMyModel(@requestBody({content: expectedContent}) inst: MyModel) {}
+        createMyModel(@requestBody2({ content: expectedContent }) inst: MyModel) { }
       }
 
       const requestBodySpec = getControllerSpec(MyController).paths['/MyModel'][
@@ -93,17 +93,17 @@ describe('requestBody decorator', () => {
     it('preserves user-provided reference in requestBody', () => {
       const expectedContent = {
         'application/json': {
-          schema: {$ref: '#/components/schemas/MyModel'},
+          schema: { $ref: '#/components/schemas/MyModel' },
         },
       };
 
-      class MyModel {}
+      class MyModel { }
 
       class MyController {
         @post('/MyModel')
         createMyModel(
-          @requestBody({content: expectedContent}) inst: Partial<MyModel>,
-        ) {}
+          @requestBody2({ content: expectedContent }) inst: Partial<MyModel>,
+        ) { }
       }
 
       const requestBodySpec = getControllerSpec(MyController).paths['/MyModel'][
@@ -115,7 +115,7 @@ describe('requestBody decorator', () => {
     it('reports error if more than one requestBody are found for the same method', () => {
       class MyController {
         @post('/greeting')
-        greet(@requestBody() name: string, @requestBody() foo: number) {}
+        greet(@requestBody2() name: string, @requestBody2() foo: number) { }
       }
       expect(() => getControllerSpec(MyController)).to.throwError(
         /An operation should only have one parameter decorated by @requestBody/,
--- a/packages/openapi-v3/src/__tests__/unit/decorators/request-body/spike.request-body.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/request-body/spike.request-body.decorator.unit.ts
@@ -52,22 +52,18 @@ describe.only('spike - requestBody decorator', () => {
     type CategoryWithRelations = Category & CategoryRelations;
 
     it('create - generates schema with excluded properties', () => {
-      const requestBodySpec = {
-        description: 'Create a product',
-        required: true,
-      };
-
-      const excludeOptions = {
-        schemaName: 'ProductWithoutID',
-        exclude: ['id']
-      }
-
       class MyController1 {
         @post('/Product')
         create(@requestBody2(
-          requestBodySpec,
+          {
+            description: 'Create a product',
+            required: true,
+          },
           Product,
-          excludeOptions
+          {
+            schemaName: 'ProductWithoutID',
+            exclude: ['id']
+          }
         ) product: Exclude<Product, ['id']>) { }
       }
 
@@ -105,21 +101,17 @@ describe.only('spike - requestBody decorator', () => {
     })
 
     it('update - generates schema with partial properties', () => {
-      const requestSpecForUpdate = {
-        description: 'Update a product',
-        required: true,
-      };
-
-      const partialOptions = {
-        partial: true
-      }
-
       class MyController2 {
         @put('/Product')
         update(@requestBody2(
-          requestSpecForUpdate,
+          {
+            description: 'Update a product',
+            required: true,
+          },
           Product,
-          partialOptions
+          {
+            partial: true
+          }
         ) product: Partial<Product>) { }
       }
 
--- a/packages/openapi-v3/src/__tests__/unit/decorators/request-body/spike.request-body.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/request-body/spike.request-body.decorator.unit.ts
@@ -1,0 +1,161 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import { belongsTo, Entity, hasMany, model, property } from '@loopback/repository';
+import { expect } from '@loopback/testlab';
+import { getControllerSpec, post, put } from '../../../..';
+import { TS_TYPE_KEY } from '../../../../controller-spec';
+import { newRequestBody1 } from '../../../../decorators/request-body.option1.decorator';
+
+describe.only('spike - requestBody decorator', () => {
+  context('proposal 1', () => {
+    @model()
+    class Product extends Entity {
+      @property({
+        type: 'string',
+      })
+      name: string;
+      @belongsTo(() => Category)
+      categoryId: number;
+
+      constructor(data?: Partial<Product>) {
+        super(data);
+      }
+    }
+
+    /**
+    * Navigation properties of the Product model.
+    */
+    interface ProductRelations {
+      category?: CategoryWithRelations;
+    }
+    /**
+    * Product's own properties and navigation properties.
+    */
+    type ProductWithRelations = Product & ProductRelations;
+
+    @model()
+    class Category extends Entity {
+      @hasMany(() => Product)
+      products?: Product[];
+    }
+    /**
+    * Navigation properties of the Category model.
+    */
+    interface CategoryRelations {
+      products?: ProductWithRelations[];
+    }
+    /**
+    * Category's own properties and navigation properties.
+    */
+    type CategoryWithRelations = Category & CategoryRelations;
+
+    it('create - generates schema with excluded properties', () => {
+      const requestBodySpec = {
+        description: 'Create a product',
+        required: true,
+      };
+
+      const excludeOptions = {
+        [TS_TYPE_KEY]: Product,
+        schemaName: 'ProductWithoutID',
+        exclude: ['id']
+      }
+
+      class MyController1 {
+        @post('/Product')
+        create(@newRequestBody1(
+          requestBodySpec,
+          excludeOptions
+        ) product: Exclude<Product, ['id']>) { }
+      }
+
+      const spec1 = getControllerSpec(MyController1)
+
+      const requestBodySpecForCreate = spec1.paths[
+        '/Product'
+      ]['post'].requestBody;
+
+      const referenceSchema = spec1.components!.schemas!.ProductWithoutID;
+
+      expect(requestBodySpecForCreate).to.have.properties({
+        description: 'Create a product',
+        required: true,
+        content: {
+          'application/json': {
+            schema: {
+              $ref: '#/components/schemas/ProductWithoutID'
+            }
+          }
+        }
+      });
+
+      // The feature that generates schemas according to
+      // different options is TO BE DONE and out of the
+      // scope of this spike, so that the schema `PartialProduct`
+      // here is still the same as `Product`
+      expect(referenceSchema).to.have.properties({
+        title: 'ProductWithoutID',
+        properties: {
+          categoryId: { type: 'number' },
+          name: { type: 'string' }
+        }
+      });
+    })
+
+    it('update - generates schema with partial properties', () => {
+      const requestSpecForUpdate = {
+        description: 'Update a product',
+        required: true,
+      };
+
+      const partialOptions = {
+        [TS_TYPE_KEY]: Product,
+        schemaName: 'PartialProduct',
+        partial: true
+      }
+
+      class MyController2 {
+        @put('/Product')
+        update(@newRequestBody1(
+          requestSpecForUpdate,
+          partialOptions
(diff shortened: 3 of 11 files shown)


packages/openapi-v3/src/controller-spec.ts
```diff
@@ -3,29 +3,14 @@
...
-  RequestBodyObject,
-  ResponseObject,
-  SchemaObject,
-  SchemasObject,
-} from './types';
+import { SchemaOptions } from './decorators/request-body.option1.decorator';
+import { resolveSchema } from './generate-schema';
+import { jsonToSchemaObject, SchemaRef } from './json-to-schema';
+import { OAI3Keys } from './keys';
+import { ComponentsObject, ISpecificationExtension, isReferenceObject, OperationObject, ParameterObject, PathObject, ReferenceObject, RequestBodyObject, ResponseObject, SchemaObject, SchemasObject } from './types';
```
raymondfeng: The formatting is broken.

packages/openapi-v3/src/decorators/request-body.option1.decorator.ts
```diff
@@ -0,0 +1,81 @@
...
+export const REQUEST_BODY_INDEX = 'x-parameter-index';
+
+export type SchemaOptions = JsonSchemaOptions & {
+  // If the schema options are provided, it implies
+  // the default paramter type need to be configured,
+  // so that `generateOpenAPISchema` can name the
+  // generated schema as `schemaName`
+  [TS_TYPE_KEY]?: Function,
+  isVisited?: boolean,
+}
```
raymondfeng: Maybe the following is better:
```ts
export interface SchemaOptions extends JsonSchemaOptions {
  tsType: {
    type: Function,
    exclude?: string[],
    optional?: string[],
  }
  isVisited?: boolean,
}
```

packages/openapi-v3/spike.md
```diff
@@ -0,0 +1,71 @@
...
+  // returns `Object` as the param type, which loses the model type's info.
+  // Therefore user must provide the model type in options.
+  [TS_TYPE_KEY]: Product,
+  // Make sure you give the custom schema a unique schema name,
+  // this name will be used as the reference name
+  // like `$ref: '#components/schemas/ProductWithoutID'`
+  schemaName: 'ProductWithoutID',
+  // The excluded properties
+  exclude: ['id']
+}
```
bajtos: I am quite unhappy about this design.

It's seems to be based on the old `ts-type` approach that we are slowly moving away from (see e.g. https://github.com/strongloop/loopback-next/pull/3402) and using `getModelSchemaRef` instead.

Let's focus on a solution that would leverage `getModelSchemaRef` and allow developers using different ORM frameworks like TypeORM to use their own helper building model-schemas from their model definitions.

packages/openapi-v3/spike.md
```diff
@@ -0,0 +1,71 @@
...
+
+// The decorator takes in the option without having a nested content object
+class MyController1 {
+  @post('/Product')
+  create(@newRequestBody1(
+    requestBodySpec,
+    excludeOptions
+  ) product: Exclude<Product, ['id']>) { }
```
bajtos: To follow-up on my comment above, I am proposing to replace `excludeOptions` with calling `getModelSchemaRef` to obtain the schema.

```ts
class MyController1 {
  @post('/Product')
  create(@newRequestBody1(
    requestBodySpec,
    getModelSchemaRef(Product, {exclude: ['id']}),
  ) product: Exclude<Product, ['id']>) { }
}
```

packages/openapi-v3/src/__tests__/unit/decorators/request-body/spike.request-body.decorator.unit.ts
```diff
@@ -0,0 +1,161 @@
...
+          name: { type: 'string' }
+        }
+      });
+    })
+
+    it('update - generates schema with partial properties', () => {
+      const requestSpecForUpdate = {
+        description: 'Update a product',
+        required: true,
+      };
```
bajtos: I find this approach, where we need one new variable (`requestSpecForUpdate`, `requestSpecForCreate`)  for each endpoint accepting a request body, as not very ergonomic.

Please update [examples/todo/src/controllers/todo.controller.ts](https://github.com/strongloop/loopback-next/blob/extract/relation-helpers/examples/todo/src/controllers/todo.controller.ts) to show your proposal in practice.

bajtos:
While this proposal does simplify request body annotation, it seems like a too small improvement to me. It makes me wonder if we could find a more radical simplification?

For example, this could be a super-simple API for people using LB4 with our ORM:

```ts
@requestBody(MyModel, {
  // parameter-definition options
  description: 'description of the request body parameter',
  required: true, // is it required?
  // schema options
  partial: true,
  optional: ['categoryId'],
  exclude: ['id'],
})

// alternatively, nest `schema` options for better readability
@requestBody(MyModel, {
  // parameter-definition options
  description: 'description of the request body parameter',
  required: true, // is it required?
  schemaOptions: {
    partial: true,
    optional: ['categoryId'],
    exclude: ['id'],
  },
})
```



API allowing people to provide their own model schema:

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // schema to use for populating `content` entries
  getModelSchemaRef(Product, {exclude: ['id']}),
)
```

I think this can be easily extended to support `x-ts-type` too, if we like so:

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // schema to use for populating `content` entries
  {'x-ts-type': Todo},
)
```

And finally the current API stays supported too:

```ts
@requestBody({
  description: 'description of the request body parameter',
  required: true, 
  content: {
    'application/json': {schema: {'x-ts-type': Todo}},
  }
)
```

Here is a mock-up declaration of `requestBody` decorator function:

```ts
// Current API
export function requestBody(requestBodySpec?: Partial<RequestBodyObject>): void;

// Simplified API - content is filled from schema
export function requestBody(
  requestBodySpec: Exclude<RequestBodyObject, 'content'>,
  schema: SchemaObject | ReferenceObject,
): void;

// Even simpler API - content schema is created from Model 
export function requestBody<T extends object>(
  modelCtor: Function & {prototype: T},
  paramAndSchemaOptions: Exclude<RequestBodyObject, 'content'> &
    JsonSchemaOptions<T>,
): void;
```

One important aspect to consider: content-specific entries can provide more than just the `schema` property, for example users can provide one or more examples. See [Request Body Object spec](https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#request-body-object) and TypeScript interfaces related to `RequestBodyObject`:

```ts
export interface RequestBodyObject extends ISpecificationExtension {
    description?: string;
    content: ContentObject;
    required?: boolean;
}
export interface ContentObject {
    [mediatype: string]: MediaTypeObject;
}
export interface MediaTypeObject extends ISpecificationExtension {
    schema?: SchemaObject | ReferenceObject;
    examples?: ExamplesObject;
    example?: any;
    encoding?: EncodingObject;
}
```

I feel it's important for the new API to allow developers to provide examples. I am not so sure about the encoding, according to the [example in OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#encoding-object-example), this fields seems to be relevant mostly for multipart (file-upload) requests. I think the short-hard form cannot be used for multipart requests and therefore we don't have to worry about `encoding` field. However, users may want to add custom extensions to `MediaTypeObject` values, so I think we will eventually need to support that too. Not necessarily as part of this PR, but it would be great to come up with a design that's flexible enough to support that use case.

One option that comes to my mind is to accept `MediaTypeObject` instead of `SchemaObject | ReferenceObject`:

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // MediaTypeObject to use for populating `content` entries
  {schema: getModelSchemaRef(Product, {exclude: ['id']})}
)
```

Let's spend more time on experimentation with different API flavors, so that we can find one that will be significantly easier to use.

/cc @strongloop/loopback-maintainers @strongloop/loopback-next

packages/openapi-v3/spike.md
jannyHou: @bajtos It's ok, I was about to switch to your proposal which uses `getModelSchemaRef` to generate the OpenAPI schema
```
class MyController1 {
  @post('/Product')
  create(@newRequestBody1(
    requestBodySpec,
    getModelSchemaRef(Product, {exclude: ['id']}),
  ) product: Exclude<Product, ['id']>) { }
}
```

but now I am wondering, what is the main purpose of SIMPLIFYING the decorator? To "avoid using very nested object to provide the declarative schema config(which is specific to loopback 4 core) - option1" or to "leverage the schema builder to generate the OAI3 schema that any ORM can consume - option2"?

My current design serves option1, but if our purpose is option2, I will try @bajtos's proposal.

cc @strongloop/loopback-maintainers any thought/preference?

raymondfeng: What's the minimum way to have a valid `requestBody` spec? I assume the new decorator is designed to allow so:

- Use a TS class such as `Product` to build the schema
- Use options such as `exclude` or `optional` to tweak the schema

I would expect something like the following to represent `Exclude<Product, ['id']>`:
```ts
class MyController1 {
  @post('/Product')
  create(@requestBody.fromType(Product, {exclude: ['id']}) product: Exclude<Product, ['id']>) { }
}
```
```

Comment (jannyHou) [reply packages/openapi-v3/spike.md]: @raymondfeng see [the spec definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#request-body-object)

At least we need two more fields:
- `description`: currently requires user to provide in `@requestBody`, but in the future it can be read from model's metadata
- `required`: a boolean value

And the nested `content` object contains more fields like `examples` besides `schema`, see comment https://github.com/strongloop/loopback-next/pull/3398#issuecomment-513726039, the discussion of the `content` spec is at the bottom.

File: packages/openapi-v3/src/decorators/request-body.option1.decorator.ts (hunk 2 above)
Comment (jannyHou) [reply packages/openapi-v3/src/decorators/request-body.option1.decorator.ts]: I agree 👍

**Comment by jannyHou** [comment]
> For example, this could be a super-simple API for people using LB4 with our ORM:

> @requestBody(MyModel, {
>   // parameter-definition options
>   description: 'description of the request body parameter',
>   required: true, // is it required?
>   // schema options
>   partial: true,
>   optional: ['categoryId'],
>   exclude: ['id'],
> })

> // alternatively, nest `schema` options for better readability
> @requestBody(MyModel, {
>   // parameter-definition options
>   description: 'description of the request body parameter',
>   required: true, // is it required?
>   schemaOptions: {
>     partial: true,
>     optional: ['categoryId'],
>     exclude: ['id'],
>   },
> })

I am a little bit reluctant to *mix* our options with the request body spec(`description`, `required`)...it means the request body decorator will contain cumbersome logic to decide which ones are options and which are specs.

--- 

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // schema to use for populating `content` entries
  getModelSchemaRef(Product, {exclude: ['id']}),
)
```

I like this proposal if we decide to pass the entire OAI schema into the `requestBody` decorator, and the improved version below is better 👇 

> One important aspect to consider: content-specific entries can provide more than just the schema property....

I agree with "accepting the `MediaTypeObject`" 👍 , if there are multiple media types the decorator should be able to take in an array of the `MediaTypeObject`.

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // MediaTypeObject to use for populating `content` entries
  {schema: getModelSchemaRef(Product, {exclude: ['id']})}
)
```

---

@bajtos I post a comment in https://github.com/strongloop/loopback-next/pull/3398/files#r305981537, let's take some time to decide which direction to go first :)

If we decide to leverage builder functions like `getModelSchemaRef` to give the decorator a complete schema/mediaTypeObject, then I would vote for:

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // MediaTypeObject to use for populating `content` entries
  {schema: getModelSchemaRef(Product, {exclude: ['id']})}
)
```

packages/openapi-v3/spike.md
bajtos: > what is the main purpose of SIMPLIFYING the decorator?

I created that user story with the intention to find a more ergonomic solution for describing request bodies. IMO, controllers methods using the default body parser should not need to repeat the fact that the request body accepts content of type `application/json`.

See e.g. what Raymond proposed in his comment:

```ts
@requestBody.fromType(Product, {exclude: ['id']})
```

and what I proposed in elsewhere in this pull request:

```ts
@requestBody(MyModel, {
  // parameter-definition options
  description: 'description of the request body parameter',
  required: true,
  // schema options
  partial: true,
  optional: ['categoryId'],
  exclude: ['id'],
})
```

The examples show a reasonably complex case. It may be not obvious that they are also making the default use case super simple - and that was the original goal!

```ts
@requestBody.fromType(Product);
// or
@requestBody(Product)
```

---

In the future, we may add support for other content-types, e.g. XML or [JSON API](https://jsonapi.org/format/#content-negotiation). It would be nice if application developers didn't need to edit all their `@requestBody` calls to add those new content types when they become supported. This isn't required, just something to be aware of.

bajtos:
> I am a little bit reluctant to mix our options with the request body spec(description, required)...it means the request body decorator will contain cumbersome logic to decide which ones are options and which are specs.

Sure, that's why I proposed the second option where schema options are nested in a new property.

```ts
@requestBody(MyModel, {
  description: 'description of the request body parameter',
  required: true,
  schemaOptions: {
    partial: true,
    optional: ['categoryId'],
    exclude: ['id'],
  },
})
```

In this design, requestBody provides `spec.schemaOptions` to `getModelSchemaRef` and copies all other properties to the spec.

> I agree with "accepting the MediaTypeObject" 👍 , if there are multiple media types the decorator should be able to take in an array of the MediaTypeObject.

Array of MediaObjectType won't work, each MediaTypeObject requires a key (the content type, e.g. `application/json`). When the developer needs to specify more than a single MediaObjectType, then they should use the current flavor of `@requestBody` and build the entire request-body spec themselves.

> If we decide to leverage builder functions like getModelSchemaRef to give the decorator a complete schema/mediaTypeObject, then I would vote for:

```ts
@requestBody(
  // fields from RequestBodyObject
  {
    description: 'description of the request body parameter',
    required: true,
  },
  // MediaTypeObject to use for populating `content` entries
  {schema: getModelSchemaRef(Product, {exclude: ['id']})}
)
```

I agree with you that this is better than my older proposals.

As discussed in https://github.com/strongloop/loopback-next/pull/3398#discussion_r305747740, I think this it not enough of an improvement. I think it makes a lot of sense to implement this flavor as the first step, I just would like us to implement also the more simple flavor in the second step.

For example:

```ts
// a mock-up, in reality we will merge requestBodyX with requestBody
function requestBodyX(
  modelCtor: Function, 
  options: Exclude<RequestBodyObject, 'content'> & {schemaOptions: JsonSchemaOptions} = {},
) {
  const schema = getModelSchemaRef(modelCtor, options.schemaOptions);

  const spec = {...options};
  delete spec.schemaOptions;

  return requestBody(spec, {schema});
}
```

jannyHou:
@bajtos @raymondfeng according to the discussion, I feel people like the signature as `@requestBody(modelCtor, ...otherSpecsAndSchemaOptions)`, I am thinking, if we adjust the order of those two params: 
`@requestBody(...otherSpcesAndSchemaOptions, modelCtor)`, it simplifies the nested content object and doesn't differ that much from the current syntax 🥇 it's not even a breaking change, the new decorator is compatible with the current one.

The improved decorator would be:
```ts
// We can preserve the current name `requestBody` since it's not a breaking change
function requestBody(
  // Let's don't exclude the `content` object:
  // - user can still provide the entire content, we only generate the schema when content is empty
  // - it doesn't break the current user's code
  options: RequestBodyObject & {schemaOptions: JsonSchemaOptions} = {},
  modelCtor: Function, 
) { }

// add `exclude: string[]` to JsonSchemaOptions
export interface SchemaOptions extends JsonSchemaOptions {
   // preserve the current usage
   `x-ts-type`?: Function
   isVisited?: boolean,
}
```

> As discussed in #3398 (comment), I think this it not enough of an improvement. I think it makes a lot of sense to implement this flavor as the first step, I just would like us to implement also the more simple flavor in the second step.

The "more simple flavor in the second step" is easier to implement actually... lol so if no objections, I will go with the proposal above.

raymondfeng:
I don't think `@requestBody(...otherSpcesAndSchemaOptions, modelCtor)` is valid as the REST parameter has to be the last argument.

`@requestBody(modelCtor?: Constructor<>, ...otherSpcesAndSchemaOptions, modelCtor)` works for me. We can make `modelCtor` optional and only honors it if the 1st arg is `Function`.

jannyHou:
@raymondfeng oops my bad, we don't need the spread operator 😅 , I updated my last comment https://github.com/strongloop/loopback-next/pull/3398#issuecomment-514384814, does it look good now?

raymondfeng:
@jannyHou Can we make the 1st arg optional so that we can do `@requestBody(Product)`?

raymondfeng:
With `otherSpecsAndSchemaOptions?: RequestBodyObject & {schemaOptions: JsonSchemaOptions},`, do we need to delete `schemaOptions` when the request body spec is built?

jannyHou:
Just had a talk with @raymondfeng , here is a summary of the use cases we could support:

- Use case 1: `@requestBody() product: Product`
- Use case 2: `@requestBody(Product, {partial: true}) product: Partial<Product>`
- Use case 3: `@requestBody({description: 'some desc'}) product: Product`
- Use case 4: `@requestBody({description: 'some desc'}, Product, {partial: true}) product: Partial<Product>`
- User case 5: if people want to build the entire request body spec, they can leverage builder functions like `getModelSchemaRef`. We can create a spec builder with fluent APIs.

I am going to implement the new design.

bajtos:
> - Use case 1: `@requestBody() product: Product`
> - Use case 2: `@requestBody(Product, {partial: true}) product: Partial<Product>`
> - Use case 3: `@requestBody({description: 'some desc'}) product: Product`
> - Use case 4: `@requestBody({description: 'some desc'}, Product, {partial: true}) product: Partial<Product>`
> - User case 5: if people want to build the entire request body spec, they can leverage builder 

I agree with the use cases described above 👍

(1) Do we want to support the following variant too?

- `@requestBody({description: 'some desc'}, Product) product: Product`

(2) I am concerned about ergonomic of the proposed API. Please verify how is Prettier formatting the code. I am concerned that the proposal for use case 4 may lead to code that's contains too much whitespace noise compared to the actual information.

```ts
// Your proposed syntax: 10 lines
@requestBody(
  {
   description: 'some desc',
  },
  Product,
  {
    partial: true,
  },
)
product: Partial<Product>,

// Compare with my proposal (7 lines):
@requestBody(Product, {
  description: 'some desc',
  schemaOptions: {
    partial: true,
  },
})
product: Partial<Product>
``` 

(3) In the use case 3, are you expecting that `Product` model will be automatically inferred by the decorator from the `design:type` metadata provided by TypeScript? That would be nice 👍 As far as I am concerned, model inference can be omitted from this story if it turns out to be non-trivial to implement.

jannyHou:
> (1) Do we want to support the following variant too?
> @requestBody({description: 'some desc'}, Product) product: Product


@bajtos This is a valid call, but the 2nd parameter is only needed when it differs from the inferred argument type.

> (2)  Please verify how is Prettier formatting the code.

Good catch, trying it.

> (3) In the use case 3, are you expecting that Product model will be automatically inferred by the decorator from the design:type metadata provided by TypeScript? 

Yep, type inference is something we already support :) and the model type precedence is just one line's change, should be simple to include in this spike.


Since the new signature is 
```ts
requestBody2(
  requestBodySpecification?: Partial<RequestBodyObject>,
  modelCtor?: Function,
  schemaOptions?: SchemaOptions
) {
// implementation
}
```
There are 3 optional parameters, I am adding test cases to verify all the possible combinations, and see what's the refactor effort to upgrade from the current one.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -271,31 +256,62 @@ function processSchemaExtensions(
   }
 }
 
+function processSchemaExtensionsForRequestBody(
```
bajtos: How is `processSchemaExtensionsForRequestBody` different from the current `processSchemaExtensions` function?

I am reluctant to have different levels of support for schema extension depending on where the schema is coming from (`@param` vs. `@requestBody`). 

Can we find a way how leverage `processSchemaExtensions` for the new decorator syntax?

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -271,31 +256,62 @@ function processSchemaExtensions(
...
+  }
+}
+
 /**
  * Generate json schema for a given x-ts-type
  * @param spec - Controller spec
  * @param tsType - TS Type
  */
-function generateOpenAPISchema(spec: ControllerSpec, tsType: Function) {
+function generateOpenAPISchema(spec: ControllerSpec, tsType: Function, options?: SchemaOptions) {
```
bajtos: I find this change weird. `spec` if of type `ControllerSpec`, which means it contains specification of multiple controller methods. Typically, each controller method accepts a slightly different request body: `create` accepts model data excluding the PK property, `patchById` accepts model data with all properties optional, and so on. I don't see how a single `SchemaOptions` is going to work for that.

Could you please help me to better understand your intent here?

bajtos: As I am reading through the code, IIUC, `generateOpenAPISchema` is adding schema for `tsType` to `spec.components.schemas` - I guess that starts to explain why you are adding `options` argument.

IIUC, you are trying to support `x-ts-type` together with `schemaOptions`. This is not something we support now, and personally I'd prefer to not support it in the future either.

packages/openapi-v3/src/generate-schema.ts
```diff
@@ -17,26 +20,31 @@ import {SchemaObject} from './types';
...
-      resolvedSchema = {$ref: `#/components/schemas/${fn.name}`};
+      // schemaName will override the function name
+      const schemaName = options && options.schemaName;
+      const titleFromMeta = fn.name + getTitleSuffix(options);
+      resolvedSchema = { $ref: `#/components/schemas/${schemaName || titleFromMeta}` };
     }
   }
 
-  return Object.assign(schema, resolvedSchema);
+  return Object.assign(schema, resolvedSchema, { options: Object.assign(options, { isVisited: true }) });
```
bajtos: I find it suspicious that you are setting `isVisited: true` here in a function that's not actually reading `isVisited` at all. Please move that change to the function that's reading `isVisited` value.

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
```diff
@@ -0,0 +1,124 @@
...
+      requestBodySpec.content = { 'application/json': {} };
+
+    // Get the design time method parameter metadata
+    const methodSig = MetadataInspector.getDesignTypeForMethod(target, member);
+    const paramTypes = (methodSig && methodSig.parameterTypes) || [];
+
+    const paramType = paramTypes[index];
+
+    // preserve backward compatibility
+    if (modelCtor) schemaOptions = Object.assign({}, schemaOptions, { [TS_TYPE_KEY]: modelCtor });
```
bajtos: Let's not use `TS_TYPE_KEY` please and find a way how to leverage `getModelSchemaRef` instead.

Ideally, I'd like all of the changes to stay inside the request-body decorator, we should not need to change other functions like `resolveSchema`.

Would the following work?

```ts
const ctor = modelCtor || paramCtor;
if (!ctor) {
  // No model to infer the schema from. This means the user is supposed
  // to describe content schema in requestBodySpecification.
  // Print a warning or throw an error if they did not do so,
  // because the OpenAPI spec  is not very useful without content schema
  return;
}
 
const schema = getModelSchemaRef(ctor, schemaOptions);
requestBodySpec.content = _.mapValues(requestBodySpec.content, c => {
  if (!c.schema) {
    c.schema = schema;
  }
 return c;
});
```

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
```diff
@@ -0,0 +1,124 @@
...
+  return (spec as Partial<RequestBodyObject>).description !== undefined
+    || (spec as Partial<RequestBodyObject>).required !== undefined
+    || (spec as Partial<RequestBodyObject>).content !== undefined;
+}
+
+export function requestBody2(
+  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
+  modelOrOptions?: Function | SchemaOptions,
+  schemaOptions?: SchemaOptions
+) {
```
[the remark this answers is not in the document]
bajtos: Also to avoid confusion about which parameter takes precedence (`spec.content['application/json'].schema` vs. `model` + `schemaOptions`), I am proposing to accept `model` only when `spec` DOES NOT contain `content` property.

```ts
export function requestBody2(
  spec: Omit<Partial<RequestBodyObject>, 'content'>,
  model: Function,
  schemaOptions?: SchemaOptions
);
```

jannyHou:
@bajtos 
> Please use function overloads instead.

I first tried function overload but it doesn't recognize the call with omitted params, maybe I missed some signatures, let me try again.

> Let's not use TS_TYPE_KEY please and find a way how to leverage getModelSchemaRef instead.
Ideally, I'd like all of the changes to stay inside the request-body decorator, we should not need to change other functions like resolveSchema.

[getModelSchemaRef](`https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/controller-spec.ts#L376`) is not called anywhere, I guess we should replace [these two lines](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/controller-spec.ts#L291-L292) to call it first.

[processSchemaExtensions](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/controller-spec.ts#L253) reads the `TS_TYPE_KEY`.

So IMO "Let's not use TS_TYPE_KEY", "leverage getModelSchemaRef instead" and "all of the changes to stay inside the request-body decorator" could not be achieved at the same time...but I will try to simplify the change in my PR.

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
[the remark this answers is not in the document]
jannyHou: @bajtos Typescript doesn't support "different number of parameters",
see https://www.tutorialsteacher.com/typescript/function-overloading

So we still need the wrapper function to resolve the caller's arguments.

packages/openapi-v3/src/controller-spec.ts
jannyHou: > spec if of type ControllerSpec, which means it contains specification of multiple controller methods. Typically, each controller method accepts a slightly different request body: create accepts model data excluding the PK property, patchById accepts model data with all properties optional, and so on. I don't see how a single SchemaOptions is going to work for that.

Yep `generateOpenAPISchema` generates json schema for a given `x-ts-type` and add it to `spec.components.schemas`

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
```diff
@@ -3,11 +3,11 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {post, requestBody, getControllerSpec} from '../../../..';
-import {expect} from '@loopback/testlab';
-import {model, property} from '@loopback/repository';
+import { model, property } from '@loopback/repository';
+import { expect } from '@loopback/testlab';
+import { getControllerSpec, post, requestBody2 } from '../../../..';
```
raymondfeng: The formatting changes should be reverted.

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
```diff
@@ -16,7 +16,7 @@ describe('requestBody decorator', () => {
       };
       class MyController {
         @post('/greeting')
-        greet(@requestBody(requestSpec) name: string) {}
+        greet(@requestBody2(requestSpec) name: string) { }
```
raymondfeng: It would be better to use `@requestBody` for all flavors.

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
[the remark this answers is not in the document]
raymondfeng: @jannyHou TypeScript cannot distinguish between `spec` and `options` from function overloading perspective.

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
jannyHou: sure. Sorry I didn't make it clear in the PR title, this is a PoC PR, not the real implementation. I should have created this PR as a draft one.

packages/openapi-v3/src/decorators/request-body.option1.decorator.ts
jannyHou: Eventually I am able to use `JsonSchemaOptions` without creating a new interface/type. So removed it in the new PR https://github.com/strongloop/loopback-next/pull/3466 :)

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.decorator.unit.ts
jannyHou: @raymondfeng see the section "Naming" in the spike.md file, I started a discussion there and proposed staying with `@requestBody()` since it doesn't break the existing usage.

packages/openapi-v3/src/generate-schema.ts
jannyHou: that flag is not needed any more in the new PoC https://github.com/strongloop/loopback-next/pull/3466

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
```diff
@@ -0,0 +1,124 @@
...
+import { resolveSchema } from '../generate-schema';
+import { OAI3Keys } from '../keys';
+import { RequestBodyObject, REQUEST_BODY_INDEX, SchemaObject } from '../types';
+
+const debug = require('debug')('loopback:openapi3:metadata:requestbody');
+
+export type SchemaOptions = JsonSchemaOptions & {
+  // To support the current flavor using 'x-ts-type',
+  // preserve backward compatibility
+  [TS_TYPE_KEY]?: Function,
```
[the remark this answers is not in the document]
jannyHou: > It is really necessary to include TS_TYPE_KEY in the options?

Nope 😅

packages/openapi-v3/src/controller-spec.ts
jannyHou: > How is processSchemaExtensionsForRequestBody different from the current processSchemaExtensions function?

Reverted these change.

jannyHou:
@bajtos All the complicated code change begin with I mixed `getModelSchemaRef()` function and the code [here](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/controller-spec.ts#L291-L292), thought they do the same thing while they are actually different...


I rewrote the code change in a new draft PR https://github.com/strongloop/loopback-next/pull/3466, and all the changes are inside requestBody decorator.

I am closing this PR and let's continue the discussion in the much cleaner PoC https://github.com/strongloop/loopback-next/pull/3466

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
jannyHou: reverted this code as well in https://github.com/strongloop/loopback-next/pull/3466

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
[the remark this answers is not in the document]
jannyHou: > Also to avoid confusion about which parameter takes precedence (spec.content['application/json'].schema vs. model + schemaOptions), I am proposing to accept model only when spec DOES NOT contain content property.

+1, let's implement it in the real PR.

PR #3262, opened 2019-06-27
feat(boot): Add a booter to bind classes with `@bind`

raymondfeng:
Implements https://github.com/strongloop/loopback-next/issues/3260#issuecomment-506438362

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3429, opened 2019-07-24
feat: add app.service() and extend service booter to load classes with `@bind`

raymondfeng:
Replaces https://github.com/strongloop/loopback-next/pull/3262

This PR extends the `service` concept to be not only a service proxy but also a local service - classes decorated with `@bind`.

## Checklist

### DIFF
--- a/packages/boot/src/booters/service.booter.ts
+++ b/packages/boot/src/booters/service.booter.ts
@@ -3,14 +3,20 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {
+  BINDING_METADATA_KEY,
+  Constructor,
+  inject,
+  MetadataInspector,
+} from '@loopback/context';
 import {CoreBindings} from '@loopback/core';
 import {ApplicationWithServices} from '@loopback/service-proxy';
-import {inject, Provider, Constructor} from '@loopback/context';
+import * as debugFactory from 'debug';
 import {ArtifactOptions} from '../interfaces';
-import {BaseArtifactBooter} from './base-artifact.booter';
 import {BootBindings} from '../keys';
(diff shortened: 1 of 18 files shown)


bajtos:
I'd like the idea in general, will try to find time to review the proposal next week.

In the meantime, could you @raymondfeng please add documentation?

packages/core/src/application.ts
```diff
@@ -241,6 +242,43 @@ export class Application extends Context implements LifeCycleObserver {
...
+   *
+   * const uuidv4 = require('uuid/v4');
+   * export class UuidProvider implements Provider<string> {
+   *   value() {
+   *     return uuidv4();
+   *   }
+   * }
+   *
+   * app.service(LogService);
+   * app.service(UuidProvider, 'uuid');
```
bajtos: Please add examples showing how to receive `LogService` and `uuid` via dependency injection. Also show what is the effect of providing a custom service name - where is that name used?

packages/boot/src/booters/service.booter.ts
```diff
@@ -77,8 +69,20 @@ export const ServiceDefaults: ArtifactOptions = {
...
+
+function isBindableClass(cls: Constructor<unknown>) {
+  if (MetadataInspector.getClassMetadata(BINDING_METADATA_KEY, cls)) {
+    return true;
+  }
+  if (isServiceProvider(cls)) {
+    debug('Provider class found: %s', cls.name);
+    return true;
+  }
+  debug('Skip class not decorated with @bind: %s', cls.name);
```
emonddr: I guess it is customary to only output informational messages when debug is turned on?

raymondfeng: Yes. It's not severe enough to use `console.warn/error()`

Issue #3383, opened 2019-07-18
[Docs] Add diagrams for "Key Concepts" page


dhmlau:
## Description
Originated from the proposal: https://github.com/strongloop/loopback-next/pull/3164

> need a diagram that shows the relationship among all the building blocks. it shows:
> - logical composition of the building blocks (diagram 1)
> - how the request flow through different tiers (diagram 2)

## Acceptance Criteria
Add the following diagram to the key concept page:
- [x] logical composition of the building blocks 
- [ ] how the request flow through different tiers 

Note: @raymondfeng might have some variation of the diagrams as a starting point.

dhmlau:
The diagram added to the key concept page is the overall diagram that includes the client, LoopBack and other services.  It's good for now. 
Closing it as done.

PR #3164, opened 2019-06-14
docs: draft proposal for docs improvement 

dhmlau:
Based on @raymondfeng's draft https://github.com/strongloop/loopback-next/pull/2925/, the purpose of this PR is to create a concrete plan on what we should do at a high level and for each section. 

The goal is that we can create smaller tasks out of this proposal.

PR #3462, opened 2019-07-26
fix(docs): add diagrams to concepts

dhmlau:
Got an overview diagram from @raymondfeng and I'm trying to get a snippet for some of the LB concepts. 

This PR implements some of the proposal items discussed in: https://github.com/strongloop/loopback-next/pull/3164. Changes in this PR:
- added the "behind the scene" sidebar item
- refactor some of the concepts to the "behind the scene"  
- added the diagram in the Concepts page

Related to https://github.com/strongloop/loopback-next/issues/3383

## Checklist


docs/site/Concepts.md
```diff
@@ -6,7 +6,9 @@ sidebar: lb4_sidebar
 permalink: /doc/en/lb4/Concepts.html
 ---
 
-LoopBack 4 introduces some new concepts that are important to understand:
+LoopBack 4 introduces some new concepts that are important to understand.
+
+![Key concepts overview diagram](imgs/key-concepts-overview-diagram.png)
 
 - [**Application**](Application.md): The central class for setting up all of
   your module’s components, controllers, servers and bindings. The Application
```
raymondfeng: We may need to refer to a lower layer that shows `context`, `binding`, and `component`.

raymondfeng: LoopBack 4 defines some key building blocks to represent different responsibilities for typical API and/or Microservice applications.

dhmlau: > We may need to refer to a lower layer that shows context, binding, and component.

@raymondfeng , do you want to move those to the "behind the scene" sidebar?

raymondfeng: Maybe.

dhmlau:
This is how the sidebar looks like:

Screen Shot 2019-08-15 at 10 49 24 PM

dhmlau:
@raymondfeng, i think ideally we can include a diagram for the "behind the scene" section. But I don't think I understand enough to create such diagram to illustrate how context/binding/DI/Components work together.  Perhaps we can work on that in a separate PR? Thanks.

docs/site/Behind-the-scene.md
```diff
@@ -0,0 +1,22 @@
+---
+lang: en
+title: 'Components'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Behind-the-scene.html
+---
+
+Here are the infrastructure that get all the artifacts working together:
```
jannyHou: `infrastructure`-->`insfrastructures`?

docs/site/Concepts.md
jannyHou: Should we move "interceptor" to section "Behind-the-scene" as well?

raymondfeng: Interceptor probably should stay outside of Behind-the-scene.

raymondfeng:
@dhmlau We should re-sort key concepts. For example, move all data related ones so that they are next to each other:

- Models (changed from Model)
- Relations
- DataSources
- Repositories

raymondfeng:
@dhmlau Nice diagrams. Can we use smaller fonts and boxes?

docs/site/DataSources.md
```diff
@@ -12,7 +12,7 @@ A `DataSource` in LoopBack 4 is a named configuration for a Connector instance
 that represents data in an external system. The Connector is used by
 `legacy-juggler-bridge` to power LoopBack 4 Repositories for Data operations.
 
-![Datasource diagram](imgs/datasource.png)
+<img src="imgs/datasource.png" alt="Datasource diagram" width="300">
```
raymondfeng: Should the image be aligned to the center?

What if we change the font/shape size in the original diagram instead of having the browser resizing it?

raymondfeng: I think we should highlight the `DataSource` box and make the rest half transparent or lower contrast.

dhmlau: The font and shape size is small because I just do a screen cap of the overview diagram, but somehow the image got expanded.  
There seems to be problems in displaying the image when I'm testing it locally.  So I'm trying to resize the image itself.

dhmlau: @raymondfeng , it seems like making the image at the center is harder than I thought. :-)
We can use the html snippet to specify the alignment, i.e.
```
<center>
  <img src="pages/en/lb4/imgs/datasource.png">
</center>
```
The image appears fine when testing locally, however, the image will appear to be broken if you view within loopback-next. 

I think it's better to be able to view it within github or within the loopback-next repo (i.e.  `Datasource diagram` )  than to make the image to be at the center.  But I don't have a very strong preference. WDYT?

raymondfeng: I can accept the left-aligned diagrams with simple markdown syntax.

PR #3466, opened 2019-07-29
Spike/new requestbody

jannyHou:
## Description

connect to #2654
A new draft PR created after all the discussions in https://github.com/strongloop/loopback-next/pull/3398.
The PoC in #3398 is over complicated, and the rebase effort is also big after introducing the generic type in interface `JsonSchemaOptions`, so I created this PR to replace it.

## Checklist

### DIFF
--- a/packages/openapi-v3/_spike-request-body.md
+++ b/packages/openapi-v3/_spike-request-body.md
@@ -1,0 +1,105 @@
+## Improve the UX of @requestBody()
+
+The original discussion is tracked in issue
+[Spike: simplify requestBody annotation with schema options](https://github.com/strongloop/loopback-next/issues/2654).
+
+The current @requestBody() can only
+
+- takes in an entire request body specification with very nested media type
+  objects or
+- generate the schema inferred from the parameter type
+
+To simplify the signature, this spike PR introduces two more parameters
+`modelCtor` and `schemaOptions` to configure the schema. The new decorator
+`requestBody2()`(let's discuss a better name later, see section
+[Naming](#Naming) is written in file 'request-body.spike.decorator.ts'
+
+### Signature
+
+The new decorator's signature is
+`@requestBody2(spec, modelCtor, schemaOptions)`.
+
+```ts
+export function requestBody2(
+  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
+  modelOrOptions?: Function | SchemaOptions,
+  schemaOptions?: SchemaOptions,
+) {
+  // implementation
+}
+```
+
+All the 3 parameters are optional, therefore in the PoC, the real implementation
+are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different
+combination of the parameters.
+
+_Please note TypeScript doesn't support function overloading with different
+number of parameters(like requestBody(spec), requestBody(model, options)).
+Therefore we have to create a wrapper function to resolve different signatures
+from the caller_
+
+My PoC PR only adds 2 unit tests for different signatures, the real
+implementation should test all the combinations.
+
+### Create - exclude properties
+
+Take "Create a new product with excluded properties" as an example:
+
+```ts
+// The decorator takes in the option without having a nested content object
+class MyController1 {
+  @post('/Product')
+  create(
+    @requestBody2(
+      // Provide the description as before
+      {
+        description: 'Create a product',
+        required: true,
+      },
+      // Using advanced ts types like `Exclude<>`, `Partial<>` results in
+      // `MetadataInspector.getDesignTypeForMethod(target, member)` only
+      // returns `Object` as the param type, which loses the model type's info.
+      // Therefore user must provide the model type in options.
+      Product,
+      // Provide the options that configure your schema
+      {
+        // The excluded properties
+        exclude: ['id'],
+      },
+    )
+    product: Exclude<Product, ['id']>,
+  ) {}
+}
+```
+
+### Update - partial properties
+
+```ts
+class MyController2 {
+  @put('/Product')
+  update(
+    @requestBody2(
+      {description: 'Update a product', required: true},
+      Product,
+      {partial: true},
+    )
+    product: Partial<Product>,
+  ) {}
+}
+```
+
+## Naming
+
+From @jannyHou: I think we can keep the name `requestBody` unchanged. I enabled
+the existing `@requestBody()`'s tests but applied `requestBody2()` decorator,
+all tests pass, which means there is no breaking change.
+
+## Follow-up Stories
+
+_I will create stories if we agree on the plan_
+
+- [ ] Modify the current `@requestBody()` according to the spike code, pass
+      existing tests across all repos.
+- [ ] Add more unit tests to verify all the signatures.
+- [ ] Upgrade examples to provide the descriptive configs(model, options) using
+      the new decorator.
--- a/packages/openapi-v3/_spike-request-body.md
+++ b/packages/openapi-v3/_spike-request-body.md
@@ -78,11 +78,9 @@ class MyController1 {
 class MyController2 {
   @put('/Product')
   update(
-    @requestBody2(
-      {description: 'Update a product', required: true},
-      Product,
-      {partial: true},
-    )
+    @requestBody2({description: 'Update a product', required: true}, Product, {
+      partial: true,
+    })
     product: Partial<Product>,
   ) {}
 }
--- a/packages/openapi-v3/_spike-request-body.md
+++ b/packages/openapi-v3/_spike-request-body.md
@@ -6,40 +6,83 @@ The original discussion is tracked in issue
 The current @requestBody() can only
 
 - takes in an entire request body specification with very nested media type
-  objects or
-- generate the schema inferred from the parameter type
+  objects OR
+- generates the schema inferred from the parameter type
 
-To simplify the signature, this spike PR introduces two more parameters
-`modelCtor` and `schemaOptions` to configure the schema. The new decorator
-`requestBody2()`(let's discuss a better name later, see section
-[Naming](#Naming) is written in file 'request-body.spike.decorator.ts'
+It doesn't give people the convenience to configure the schema spec with `model`
+and `schemaOptions` without defining them in a nested object. To simplify the
+signature, this spike PR introduces two more parameters `model` and
+`schemaOptions` to configure the schema.
+
+Please note the 2nd and 3rd parameters are only used to contribute the spec for
+_default schema_(I will explain what default means with an example), any other
+request body spec are still defined in the 1st parameter.
+
+See the example from the swagger official website:
+https://swagger.io/docs/specification/describing-request-body/
+
+- `application/json` and `application/xml` share the same schema
+  `#/components/schemas/Pet`
+- `application/x-www-form-urlencoded` uses another schema
+  `#/components/schemas/PetForm'
+- `text/plain` receives a string: `type: string`
+
+The new decorator allows
+
+- a default schema (generated from the 2nd and 3rd params, should be `Pet` in
+  this case) _and_
+- custom schemas (provided by the 1st parameter: partial spec, `PetForm` and
+  string schema)
+
+so that developers don't need to repeat defining the same schema in the 1st
+parameter.
+
+The logic of the new decorator is:
+
+- The 1st parameter contains the pure OpenAPI request body spec
+- Other parameters contribute the default schema' config, `model`, `options`.
+- The decorator:
+  - first reads schema config
+  - then generate default schema spec accordingly
+  - if any content type is missing `schema`, then assign the default schema to
+    it(Our current implementation is already doing it, see
+    [code](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/decorators/request-body.decorator.ts#L103-L107))
+
+ANY OTHER PARTS of the spec are read from the 1st parameter, the decorator only
+generate and merge the default `schema`. We can implement helpers/builders for
+developers to build the entire spec, like
+
+```ts
+requestBodyBuilder
+  .addContent('application/x-www-form-urlencoded')
+  .withSchema(model, options)
+  .withExample(exampleSpecs);
+```
+
+The new decorator `requestBody2()`(let's discuss a better name later, see
+section [Naming](#Naming) is written in file 'request-body.spike.decorator.ts'
 
 ### Signature
 
 The new decorator's signature is
 `@requestBody2(spec, modelCtor, schemaOptions)`.
 
+All the 3 parameters are optional, therefore in the PoC, the real implementation
+are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different
+combination of the parameters.
+
+My PoC PR only adds 2 unit tests for different signatures, the real
+implementation should test all the combinations.
+
 ```ts
-export function requestBody2(
-  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
-  modelOrOptions?: Function | SchemaOptions,
+export function _requestBody2<T>(
+  specOrModelOrOptions?: Partial<RequestBodyObject>,
+  modelOrOptions?: Function & {prototype: T},
   schemaOptions?: SchemaOptions,
 ) {
   // implementation
 }
 ```
-
-All the 3 parameters are optional, therefore in the PoC, the real implementation
-are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different
-combination of the parameters.
-
-_Please note TypeScript doesn't support function overloading with different
-number of parameters(like requestBody(spec), requestBody(model, options)).
-Therefore we have to create a wrapper function to resolve different signatures
-from the caller_
-
-My PoC PR only adds 2 unit tests for different signatures, the real
-implementation should test all the combinations.
 
 ### Create - exclude properties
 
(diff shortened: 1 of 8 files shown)


packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (_.isEmpty(requestBodySpec.content))
+      requestBodySpec.content = { 'application/json': {} };
+
+    // Get the design time method parameter metadata
+    const methodSig = MetadataInspector.getDesignTypeForMethod(target, member);
+    const paramTypes = (methodSig && methodSig.parameterTypes) || [];
+
+    const paramType = paramTypes[index];
+
+    // preserve backward compatibility
```
jannyHou: FYI: compared with the current decorator function, the difference if from Line 81 to Line 91, other code stay the same.

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
```diff
@@ -0,0 +1,220 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import { belongsTo, Entity, hasMany, model, property } from '@loopback/repository';
+import { expect } from '@loopback/testlab';
+import { getControllerSpec, post, put } from '../../../..';
+import { requestBody2 } from '../../../../decorators/request-body.spike.decorator';
```
raymondfeng: Formatting is broken.

jannyHou: @raymondfeng Fixed :)

packages/openapi-v3/_spike-request-body.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+All the 3 parameters are optional, therefore in the PoC, the real implementation
+are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different
+combination of the parameters.
+
+_Please note TypeScript doesn't support function overloading with different
+number of parameters(like requestBody(spec), requestBody(model, options)).
+Therefore we have to create a wrapper function to resolve different signatures
+from the caller_
```
bajtos: I am afraid this is not true. The following code seems to compile correctly for me.

```ts

export function requestBody2(
  spec: Partial<RequestBodyObject>,
): (target: object, member: string, index: number) => void;

export function requestBody2(
  spec: Partial<RequestBodyObject>,
  model: Function,
  schemaOptions?: SchemaOptions,
): (target: object, member: string, index: number) => void;

export function requestBody2(
  spec: Partial<RequestBodyObject>,
  schemaOptions?: SchemaOptions,
): (target: object, member: string, index: number) => void;

export function requestBody2(
  schemaOptions?: SchemaOptions,
): (target: object, member: string, index: number) => void;

export function requestBody2(
  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
  modelOrOptions?: Function | SchemaOptions,
  schemaOptions?: SchemaOptions,
): (target: object, member: string, index: number) => void {
  return (target: object, member: string, index: number): void => {
    // implementation
  };
}
```

packages/openapi-v3/_spike-request-body.md
```diff
@@ -0,0 +1,110 @@
...
+  modelOrOptions?: Function | SchemaOptions,
+  schemaOptions?: SchemaOptions,
+) {
+  // implementation
+}
+```
+
+All the 3 parameters are optional, therefore in the PoC, the real implementation
+are in `_requestBody2()`, `requestBody2()` is a wrapper that resolves different
+combination of the parameters.
```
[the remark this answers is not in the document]
bajtos: > (2)
> I am proposing to forbid the following form
> and always require users to provide the model constructor if they want to use JsonSchemaOptions

My point (2) becomes irrelevant if we decide to nest `schemaOptions` as a new property of the first argument, see https://github.com/strongloop/loopback-next/pull/3466#discussion_r308651390,

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
```diff
@@ -0,0 +1,233 @@
...
+          @requestBody2(
+            {
+              description: 'Create a product',
+              required: true,
+            },
+            Product,
+            {
+              exclude: ['id'],
+            },
+          )
```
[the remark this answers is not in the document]
bajtos: Just to make it clear - I am not saying that we must use my proposal, I just want us to find a more ergonomic solution than the current proposal. Feel free to come up with a different one!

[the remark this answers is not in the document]
jannyHou: @bajtos prettier allows to shorten to 3 lines 👇 
```ts
class MyController2 {
      @put('/Product')
      update(
        @requestBody2(
          {description: 'Update a product', required: true},
          Product,
          {partial: true},
        )
        product: Partial<Product>,
        ) {}
}
```
I also updated the  code and doc, does it eliminate your concern?

packages/openapi-v3/_spike-request-body.md
jannyHou: @bajtos This is weird, not sure if you read the comment in https://github.com/strongloop/loopback-next/pull/3398#discussion_r308324740
yesterday I double checked it with Dom and I believe the conclusion is overloading with different number of parameters is not supported, and my editor still complains with your code:
Screen Shot 2019-07-30 at 9 51 53 AM

packages/openapi-v3/_spike-request-body.md
```diff
@@ -0,0 +1,110 @@
...
+`@requestBody2(spec, modelCtor, schemaOptions)`.
+
+```ts
+export function requestBody2(
+  specOrModelOrOptions?: Partial<RequestBodyObject> | Function | SchemaOptions,
+  modelOrOptions?: Function | SchemaOptions,
+  schemaOptions?: SchemaOptions,
+) {
+  // implementation
+}
```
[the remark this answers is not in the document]
jannyHou: The reason why I put the private function `_requestBody2()`'s signature(instead of the wrapper's) here is to avoid the confusion caused by manually resolve the caller's input.

Let me change the name to the private one.

packages/openapi-v3/_spike-request-body.md
bajtos: What version of TypeScript are you using in the editor? I vaguely remember that VS Code sometimes prefers to use the built-in version instead of the version installed in `node_modules`. Could that explain the difference?

bajtos: Screen Shot 2019-07-30 at 16 03 40

Screen Shot 2019-07-30 at 16 04 00

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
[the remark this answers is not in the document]
bajtos: Fair enough 👍

packages/openapi-v3/_spike-request-body.md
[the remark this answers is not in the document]
jannyHou: @bajtos I don't think excluding the `content` for the first parameter`spec` is a good idea, the 2nd and 3rd parameters are there to **describe the schema**, but `content` has more stuff than `schema`, and user should still have their own flexibility to 
- provide an entire `content` including `schema` without using the 2nd and 3rd params
- if the 2nd+3rd params decide the default schema, user can override it for special media types with different schemas, like `application/x-www-form-urlencoded` in [this example](https://swagger.io/docs/specification/describing-request-body/)

So honestly, IMO the request spec(especially content) is nested in nature, unless we provide a bunch of fluent APIs to help users build it, like `requestBodyBuilder.addContent('application/x-www-form-urlencoded').withSchema(model, options).withExample(exampleSpecs)`, there is not that much to simplify for the UX...and if we implement those builders, the purpose is still generating *a complete requestBody spec*, which can be provided as the 1st parameter in `@requestBody()`, and still fits my current proposal.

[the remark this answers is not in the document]
bajtos: My point of view:

- If users want to override special media types, then they provide full `content` object in the first argument.
- The 2nd+3rd params are only for the case when they want LB to generate default `content` object for the given model & schema options.

If you like to go forward with your proposal, then please document the rules we are going to merge the (deep) fields in `spec.content` with the fields created from model + schemaOptions.

I am also interested how are developers going to learn about those merge rules and more importantly, when they do a mistake, how are they going to figure out why they are getting different schema from what they were expected to see?

Let's start with the example I have shown before:

```ts
requestBody({
  content: {
    'application/json': {
      schema: getModelSchemaRef(MyModel)
    }
  },
  MyModelData, // notice a different class!
);
```

Another case to consider:

```ts
requestBody({
  content: {
    'application/json': {
      example: {/* some model data */
    }
  },
  MyModel,
);
```

Are we going to print debug logs or perhaps regular warnings to notify developers when the framework is overwriting spec in a way that may be unexpected?

packages/openapi-v3/_spike-request-body.md
jannyHou: @bajtos I have the same problem with 3.5.3, see:

Screen Shot 2019-07-30 at 12 29 22 PM

packages/openapi-v3/_spike-request-body.md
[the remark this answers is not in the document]
jannyHou: @bajtos 
> 1)
When the spec contains content property, then we should not allow model & schemaOptions.

Let's still discuss with the example from the swagger official website:
https://swagger.io/docs/specification/describing-request-body/

- `application/json` and `application/xml` share the same schema `#/components/schemas/Pet`
- `application/x-www-form-urlencoded` uses another schema `#/components/schemas/PetForm'
- `text/plain` receives a string: `type: string`

My point is we could allow 
  - a default schema (generated from the 2nd and 3rd params, should be `Pet` in this case)
  *and*
  - custom schemas (provided by the 1st parameter: partial spec, `PetForm` and string schema)

so that developers don't need to repeat defining the same schema in the 1st parameter.
And our current implementation is already doing it, see [code](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/decorators/request-body.decorator.ts#L103-L107):

- we generate a default schema according to the interred type
- iterate through the contents, if `schema` is missing, `content.schema = defaultSchema`.

Back to your use case:

```ts
requestBody({
  content: {
    'application/json': {
      schema: getModelSchemaRef(MyModel)
    }
  },
  MyModelData, // notice a different class!
);
```

This is a VALID use case IMO. 
cc @strongloop/sq-lb-apex @raymondfeng more opinions are welcomed here ^^ :)

With your proposal which merges the `JsonSchemaOptions` into the 1st parameter, everything would be the same...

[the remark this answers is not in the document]
jannyHou: @bajtos hope my comment https://github.com/strongloop/loopback-next/pull/3466#discussion_r308941807 answers your question in https://github.com/strongloop/loopback-next/pull/3466#discussion_r308762644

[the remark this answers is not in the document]
jannyHou: > 2) As discussed in #3398 (comment), we may want to add a new overload in the future that will allow users to a different ORM framework supply the MediaTypeObject in requestBody parameters...

@bajtos I would suggest NOT achieve this by adding more signatures by overloading, we can introduce fluent spec builder functions to generate any part of the requestBody spec, they would be much more straightforward to use than *providing partial specs in some parameters and the decorator function resolves different parts from different places. 

The logic in the current decorator, or my new proposal is pretty straightforward to understand IMO:

- The 1st parameter contains the pure OpenAPI request body spec
- Other parameters contribute the default schema' config, `model`, `options`...
- The decorator:
  - reads schema config
  - generate default schema spec accordingly
  - if any content type is missing `schema`, then assign the default schema to it

ANY OTHER PARTS of the spec are read from the 1st parameter, the decorator only generate and merge the default `schema`. We can implement helpers/builders for developers to build the entire spec.

packages/openapi-v3/_spike-request-body.md
raymondfeng: The overloaded functions work for me in VSCode and `npm run build`:

```ts
export function requestBody2(
  spec: Partial<RequestBodyObject>,
): (target: object, member: string, index: number) => void;

export function requestBody2<T extends object>(
  spec: Partial<RequestBodyObject>,
  model: Function,
  schemaOptions?: JsonSchemaOptions<T>,
): (target: object, member: string, index: number) => void;

export function requestBody2<T extends object>(
  spec: Partial<RequestBodyObject>,
  schemaOptions?: JsonSchemaOptions<T>,
): (target: object, member: string, index: number) => void;

export function requestBody2<T extends object>(
  schemaOptions?: JsonSchemaOptions<T>,
): (target: object, member: string, index: number) => void;

export function requestBody2<T extends object>(
  specOrModelOrOptions?:
    | Partial<RequestBodyObject>
    | Function
    | JsonSchemaOptions<T>,
  modelOrOptions?: Function | JsonSchemaOptions<T>,
  schemaOptions?: JsonSchemaOptions<T>,
) {
//
}
```

packages/openapi-v3/_spike-request-body.md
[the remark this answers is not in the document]
agnes512: After discussing with Janny, I think her proposal makes sense to me high-level(ly). Especially the idea of that we can have some build functions that similar to [this](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-spec-builder/src/openapi-spec-builder.ts#L65) that allows users to add/build spec object if they want to pass in more information. 

But I am not sure if @jannyHou 's proposal could address the following requirement that @bajtos mentioned above:

> (2)
> As discussed in [#3398 (comment)](https://github.com/strongloop/loopback-next/pull/3398#issuecomment-514206650), we may want to add a new overload in the future that will allow users to a different ORM framework supply the `MediaTypeObject` in `requestBody` parameters..

If yes, I am fine with the proposal.

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
```diff
@@ -0,0 +1,223 @@
...
+        content: {
+          'application/json': {
+            schema: {
+              $ref: '#/components/schemas/ProductExcluding[id]',
+            },
+          },
+        },
+      });
+
+      // The feature that generates schemas according to
```
b-admike: Can we include this in the list of follow up stories in the markdown file you have (disregard if already included)?

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
```diff
@@ -0,0 +1,223 @@
...
+            Product,
+            {partial: true},
+          )
+          product: Partial<Product>,
+        ) {}
+      }
+
+      const spec2 = getControllerSpec(MyController2);
+
+      const requestBodySpecForCreate =
```
b-admike: nitpick: `requestBodySpecForUpdate`

packages/openapi-v3/_spike-request-body.md
b-admike: Seems like it works for 2/3 of us, so this would be a great addition/change to us having to determine the arguments from the callee.

packages/openapi-v3/_spike-request-body.md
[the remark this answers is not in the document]
b-admike: I agree with @agnes512, but for the first iteration, let's not exclude the content for the first parameter. I am not knowledgable to know if the use-case pointed by @bajtos is either valid or not valid since we have differing opinions, so I think we should explore it further outside the scope of this PR.

packages/openapi-v3/_spike-request-body.md
jannyHou: k...finally get rid of the error after several times' code comment/uncomment 👏  , but still have no clue why that error happened...

packages/openapi-v3/src/__tests__/unit/decorators/request-body/request-body.spike.decorator.unit.ts
jannyHou: @b-admike Seems this feature is already supported.

packages/openapi-v3/src/decorators/request-body.spike.decorator.ts
```diff
@@ -0,0 +1,142 @@
...
+    (spec as Partial<RequestBodyObject>).content !== undefined
+  );
+}
+
+export function requestBody2<T extends object>(
+  specOrModelOrOptions?:
+    | Partial<RequestBodyObject>
+    | Function
+    | JsonSchemaOptions<T>,
+  modelOrOptions?: Function | JsonSchemaOptions<T>,
```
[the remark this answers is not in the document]
jannyHou: I am able to catch the non-existing fields in `exclude` by using `Function & {prototype: T}`, but
```ts
@requestBody<RandomModel>(spec, Product) test: Product
```

I added a test case for it, let's fix it in the implementation.

jannyHou:
@nabdelgadir 
> At first glance, I thought it should be rejected, but from your POC why do you think it should be a valid use case?

See my explanation in https://github.com/strongloop/loopback-next/pull/3466/files#diff-12798cc8f728c5affaa88ab994dbf31fR21, and also [comment](https://github.com/strongloop/loopback-next/pull/3466#discussion_r308941807).

IMO user should have the flexibility to define a default schema using the 2nd and 3rd parameters, but override it with custom ones provided in the 1st parameter.

jannyHou:
Subsequent tasks see https://github.com/strongloop/loopback-next/issues/2654#issuecomment-517448716
Closing.

PR #3469, opened 2019-07-30
feat(eslint-config): enable "no-return-await" rule

bajtos:
See https://eslint.org/docs/rules/no-return-await:

> Since the return value of an async function is always wrapped in
> `Promise.resolve, `return await` doesn’t actually do anything except
> add extra time before the overarching `Promise` resolves or rejects.

- The first commit modifies `@loopback/eslint-config`. This is a breaking change.
- The second commits fixes violations in our code base. This is a backwards-compatible chore.
- The third commit adds a new task to VSCode config to configure eslint problem matcher for `npm run lint:fix`.

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -188,7 +188,7 @@ export class OrderController {
   async getCustomer(
     @param.path.number('id') orderId: typeof Order.prototype.id,
   ): Promise<Customer> {
-    return await this.orderRepository.customer(orderId);
+    return this.orderRepository.customer(orderId);
   }
 }
 ```
--- a/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
+++ b/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
@@ -26,7 +26,7 @@ describe('HttpCachingProxy', () => {
   let proxy: HttpCachingProxy;
   after(stopProxy);
 
-  beforeEach('clean cache dir', async () => await rimraf(CACHE_DIR));
+  beforeEach('clean cache dir', async () => rimraf(CACHE_DIR));
 
   it('provides "url" property when running', async () => {
     await givenRunningProxy();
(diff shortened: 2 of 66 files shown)


packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -26,7 +26,7 @@ describe('HttpCachingProxy', () => {
   let proxy: HttpCachingProxy;
   after(stopProxy);
 
-  beforeEach('clean cache dir', async () => await rimraf(CACHE_DIR));
+  beforeEach('clean cache dir', async () => rimraf(CACHE_DIR));
```
nabdelgadir: Why is this one removed?

bajtos:
```ts
async () => await rimraf(CACHE_DIR)
```

The code above is essentially the same as the following:

```ts
(async function () { return await rimraf(CACHE_DIR); }).bind(this)
```

nabdelgadir: Ah, thank you.

raymondfeng: In this case, `async` is not required either. We can do `beforeEach('clean cache dir', () => rimraf(CACHE_DIR));` as Mocha supports a function that returns a Promise.

docs/site/BelongsTo-relation.md
```diff
@@ -188,7 +188,7 @@ export class OrderController {
   async getCustomer(
     @param.path.number('id') orderId: typeof Order.prototype.id,
   ): Promise<Customer> {
-    return await this.orderRepository.customer(orderId);
+    return this.orderRepository.customer(orderId);
```
raymondfeng: Now no `await` is used. Do we want to remove `async` keyword for the method signature?

raymondfeng: Maybe keep `async` so that the code can be modified to use `await` before the `return` statement.

PR #3470, opened 2019-07-30
fix: enforce JsonSchemaOptions type when building model schema

bajtos:
Modify functions like `getModelSchemaRef` to link the prototype of the constructor function with the `T` argument of `JsonSchemaOptions`.

Before this change, the following code happily compiles:

    schema: getModelSchemaRef(Note, {exclude: ['prop-doesnt-exist'])})

(The reason: compiler resolves `T` as `object`, therefore any properties are allowed in `exclude` list.)

With this change in place, the code above is rejected by the compiler, because the compiler understands that `T` is `Note` and therefore only `Note` properties are allowed in the `exclude` list.

    Argument of type '{ exclude: "prop-doesnt-exist"[]; }' is not
    assignable to parameter of type 'JsonSchemaOptions<Note>'.

      Types of property 'exclude' are incompatible.
        (...)

        Type '"prop-doesnt-exist"' is not assignable to
        type '"content" | "id" | "title" | "getId" | (...)'.

I discovered this problem while reviewing https://github.com/strongloop/loopback-next/pull/3466

## Checklist

### DIFF
--- a/packages/openapi-v3/src/controller-spec.ts
+++ b/packages/openapi-v3/src/controller-spec.ts
@@ -374,7 +374,7 @@ export function getControllerSpec(constructor: Function): ControllerSpec {
  * @param options - Additional options
  */
 export function getModelSchemaRef<T extends object>(
-  modelCtor: Function,
+  modelCtor: Function & {prototype: T},
   options?: JsonSchemaOptions<T>,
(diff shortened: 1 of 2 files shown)


jannyHou:
LGTM 👍

Issue #3443, opened 2019-07-26
Add `findByForeignKeys` helper (initial version)


bajtos:
Implement a new helper function `findByForeignKeys`, it will become a part of a public API of `@loopback/repository`.

Signature:

```ts
function findByForeignKeys<
  Target extends Entity,
  TargetID,
  TargetRelations extends object,
  ForeignKey
>(
  targetRepository: EntityCrudRepository<Target, TargetID, TargetRelations>,
  fkName: StringKeyOf<Target>,
  fkValues: ForeignKey[],
  scope?: Filter<Target>,
  options?: Options,
): Promise<(Target & TargetRelations)[]>;
```

The initial version should be simple, "inq splitting" and additional "scope" constraints are out of scope of this task.

Example implementation - I think it pretty much covers the full scope of this task, but obviously we need to add test coverage as explained in the acceptance criteria. It would be also great to throw an error if `scope` argument was provided, because we don't support scopes yet.

```ts
const where = ({
  [fkName]: fks.length === 1 ? fkValues[0] : {inq: fkValues},
} as unknown) as Where<Target>;
const targetFilter = {where};

return targetRepository.find(targetFilter, options));
```

See #3387 for more details & a prototype implementation.

## Acceptance criteria

Beside the implementation, it's important to create testing infrastructure that will make it easy to add new tests in the future. There should be two kinds of tests:

- Unit-level tests inside `packages/repository`, these tests should mock `targetRepository`. By using a mock repository, we can assert on how many queries are called, inject errors to verify how are they handled, etc.
- Integration-level tests inside `packages/repository-tests`, these tests will call real repository class & database, we will invoke them against different connectors (MySQL, MongoDB, and so on).

### Out of scope

- Support `inq` splitting in `findByForeignKeys` #3444
- Include related models with a custom scope #3453

PR #3473, opened 2019-07-30
feat(repository): add findByForeignKeys function

nabdelgadir:
Implemented initial version of the new helper function `findByForeignKeys`. 

Resolves https://github.com/strongloop/loopback-next/issues/3443.

## Checklist

### DIFF
--- a/packages/repository/src/relations/relation.helpers.ts
+++ b/packages/repository/src/relations/relation.helpers.ts
@@ -1,0 +1,44 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import * as _ from 'lodash';
+import {Entity, EntityCrudRepository, Filter, Options, Where} from '..';
+
+/**
+ * Finds model instances that contain any of the provided foreign key values.
+ *
+ * @param targetRepository - The target repository where the model instances are found
+ * @param fkName - Name of the foreign key
+ * @param fkValues - Array of the values of the foreign keys to be included
+ * @param scope - Additional scope constraints (not currently supported)
+ * @param options - Options for the operations
+ */
+export async function findByForeignKeys<
+  Target extends Entity,
+  TargetID,
+  TargetRelations extends object,
+  ForeignKey
+>(
+  targetRepository: EntityCrudRepository<Target, TargetID, TargetRelations>,
+  fkName: StringKeyOf<Target>,
+  fkValues: ForeignKey[],
+  scope?: Filter<Target>,
+  options?: Options,
+): Promise<(Target & TargetRelations)[]> {
+  // throw error if scope is defined and non-empty
+  // see https://github.com/strongloop/loopback-next/issues/3453
+  if (scope && !_.isEmpty(scope)) {
+    throw new Error('scope is not supported');
+  }
+
+  const where = ({
+    [fkName]: fkValues.length === 1 ? fkValues[0] : {inq: fkValues},
+  } as unknown) as Where<Target>;
+  const targetFilter = {where};
+
+  return targetRepository.find(targetFilter, options);
+}
+
+export type StringKeyOf<T> = Extract<keyof T, string>;
--- a/packages/repository-tests/src/crud/create-retrieve.suite.ts
+++ b/packages/repository-tests/src/crud/create-retrieve.suite.ts
@@ -7,6 +7,7 @@ import {
   AnyObject,
   Entity,
   EntityCrudRepository,
+  findByForeignKeys,
   model,
   property,
 } from '@loopback/repository';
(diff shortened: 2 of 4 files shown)


packages/repository/src/relations/relation.helpers.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export async function findByForeignKeys<
+  Target extends Entity,
+  TargetID,
+  TargetRelations extends object,
+  ForeignKey
+>(
+  targetRepository: EntityCrudRepository<Target, TargetID, TargetRelations>,
+  fkName: StringKeyOf<Target>,
+  fkValues: ForeignKey[],
```
raymondfeng: Can we accept a single value too? For example:

```ts
fkValues: ForeignKey[] | ForeignKey,
```
```

File: packages/repository/src/relations/relation.helpers.ts [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param targetRepository - The target repository where the model instances are found
+ * @param fkName - Name of the foreign key
+ * @param fkValues - Array of the values of the foreign keys to be included
+ * @param scope - Additional scope constraints (not currently supported)
+ * @param options - Options for the operations
+ */
+export async function findByForeignKeys<
+  Target extends Entity,
+  TargetID,
```
raymondfeng: Do we really need this? Use `unknown` is probably good enough.

packages/repository/src/relations/relation.helpers.ts
raymondfeng: Is it possible to have a type param such as `FK extends StringKeyOf<Target>` and use it to constrain fk values, such as `Target[FK]`?

packages/repository/src/relations/relation.helpers.ts
nabdelgadir: `TargetID`? but we use it in [`targetRepository: EntityCrudRepository<Target, TargetID, TargetRelations>,`](https://github.com/strongloop/loopback-next/pull/3473/files#diff-791ac353d48a6890b8c35dfaa1c836a1R24)

raymondfeng: We can remove `TargetID` from the type params and use `unknown`:

 targetRepository: EntityCrudRepository<Target, unknown, TargetRelations>,

nabdelgadir: Ohh I see, thanks! What's the benefit of using `unknown`?

raymondfeng: The `TargetID` does not matter in our usage of `find` as no `findById` is involved.

packages/repository/src/relations/relation.helpers.ts
agnes512: Let me reword it to make sure that I understand the type constrains in `tsc` correctly:

if we do `foreignKey extends StringKeyOf<Target>`, it would constrain the passed in `foreignKey` to be a string-like `<Target>` instance?

raymondfeng: I'm proposing the following:

```ts
export async function findByForeignKeys<
  Target extends Entity,
  FK extends StringKeyOf<Target>,
  TargetRelations extends object,
>(
  targetRepository: EntityCrudRepository<Target, unknown, TargetRelations>,
  fkName: FK,
  fkValues: Target[FK][],
  scope?: Filter<Target>,
  options?: Options,
): Promise<(Target & TargetRelations)[]> {
}
```

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  options?: Options,
+): Promise<(Target & TargetRelations)[]> {
+  // throw error if scope is defined and non-empty
+  // see https://github.com/strongloop/loopback-next/issues/3453
+  if (scope && !_.isEmpty(scope)) {
+    throw new Error('scope is not supported');
+  }
+
+  const value = Array.isArray(fkValues)
+    ? fkValues.length === 1
```
raymondfeng: Let's add a shortcut for empty array to return `[]` immediately. Include a test to cover this corner case too.

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+      : {inq: fkValues}
+    : fkValues;
+
+  const where = ({[fkName]: value} as unknown) as Where<Target>;
+  const targetFilter = {where};
+
+  return targetRepository.find(targetFilter, options);
+}
+
+export type StringKeyOf<T> = Extract<keyof T, string>;
```
raymondfeng: Maybe it's not necessary to export this type?

packages/repository/src/relations/relation.helpers.ts
nabdelgadir: If we pass in an empty array, it complains that `Type 'never[]' is not assignable to type 'number'.` (or whatever the type of the foreign key value is)

raymondfeng: You can do the following:
```ts
const fkIds: number[] = []; 
```

PR #3479, opened 2019-07-31
Give users more info on lb4 discover.

SouthRibbleTech:
## Checklist

### DIFF
--- a/docs/site/Discovering-models.md
+++ b/docs/site/Discovering-models.md
@@ -8,6 +8,8 @@ permalink: /doc/en/lb4/Discovering-models.html
 
 ## Synopsis
 
+Please note that the code below will not work without creating .js connector files. Also not that lb4 will only create .ts connector files. Your on your own, good luck.
+
 LoopBack makes it simple to create models from an existing relational database.
 This process is called _discovery_ and is supported by the following connectors:
 


docs/site/Discovering-models.md
```diff
@@ -8,6 +8,8 @@ permalink: /doc/en/lb4/Discovering-models.html
 
 ## Synopsis
 
+Please note that the code below will not work without creating .js connector files. Also not that lb4 will only create .ts connector files. Your on your own, good luck.
+
```
raymondfeng: Can you elaborate the issue you ran into? Did you refer to an existing issue?

bajtos: +1 to what Raymond said.

The information in the note does not match my mental model of how model discovery works, it does not make sense to me.

Let's build a better understanding of the problem you @SouthRibbleTech are facing. Then we can provide a better advice than just "you are on your own" and perhaps even improve our tooling to support your scenario.

achrinza: I'm making an assumption, but I believe the culprit is this line:

image

While it does say the project must be built, it assumes that the readers know that they should run `npm run build`. Currently, there's no documentation on the default, generated npm scripts. The only documentation on them is on the [`@loopback/build` README ](https://github.com/strongloop/loopback-next/blob/master/packages/build/README.md).

We could potentially create a new doc page explaining the purpose of these scripts and how to use them, then link to it from the `Discovering Models` doc page.

bajtos:
@SouthRibbleTech what's the status of this pull request? Did the explanation in https://github.com/strongloop/loopback-next/pull/3479#discussion_r347719403 clarified and resolved the problem for you? Are you still keen to improve our documentation to help the next person running into the same confusion?

achrinza:
@SouthRibbleTech I believe #4670 provides additional clarity on the prerequisites of `lb4 discover`. If it doesn't, please feel free to open a new issue.

achrinza:
Duplicate of #4670

Issue #2598, opened 2019-03-15
Add "Connectors Reference" docs pages to LB4 docs


dhmlau:
## Description / Steps to reproduce / Feature proposal
When working on https://github.com/strongloop/loopback-next/issues/1766, I realized that the docs pages under [Connectors references](https://loopback.io/doc/en/lb3/Connectors-reference.html) are still very relevant, and also because we're still using the existing connectors.  

I'd like to propose to copy these docs pages to LB4 docs. 
@strongloop/loopback-maintainers, what do you think? 

## Acceptance Criteria
- [ ] Add the [Connectors Reference](https://loopback.io/doc/en/lb3/Connectors-reference.html) including the subtree to LB4 docs. 
  - [ ] memory connector
  - [ ] database connector
      - can we assume all will be working in LB4? 
  - [ ] Other connectors 
      - Are everything under this category works the same way in LB4?  There's gap in email connector #1980 
  - [ ] community connectors

References:
- Sample of how it can be done: https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/lb3/Cloudant-connector.md.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

bajtos:
>  I realized that the docs pages under Connectors references are still very relevant, and also because we're still using the existing connectors.
>
> I'd like to propose to copy these docs pages to LB4 docs.

+1

There are more content in LB3 that's applicable to LB4 too. For example, the documentation for `filter` and `where` syntax, description of LDL (metadata used to define model and properties), etc.

dhmlau:
FYI - @b-admike showed an example of how the content can be "copied" from the README of a repo to loopback.io: https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/lb3/Cloudant-connector.md.

dhmlau:
For `filter` and `where`, we have a task for that already: https://github.com/strongloop/loopback-next/issues/2527.

dhmlau:
@bschrammIBM, thanks for taking up this task. I'm assigning it to you now.  Thanks.

bschrammIBM:
I moved the entire topic from LB3: https://loopback.io/doc/en/lb3/Connectors-reference.html to loopback-next repo and updated the sidebars and the headers.
The branch is `connectors_from_lb3`.
However, there is an error when building the site that needs to be resolved.
[...]

PR #3481, opened 2019-07-31
feat: add Connector docs to LB4 docs

hacksparrow:
Add Connector docs to LB4 docs.

Required to fix https://github.com/strongloop/loopback-next/issues/2598.

### DIFF
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -257,7 +257,6 @@ children:
   url: Crafting-LoopBack-4.html
   output: 'web, pdf'
 
-
 - title: 'Tutorials'
   url: Tutorials.html
   output: 'web, pdf'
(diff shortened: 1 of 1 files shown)


docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -415,6 +414,194 @@ children:
     url: Download-examples.html
     output: 'web, pdf'
 
+- title: 'Connectors reference'
+  url: Connectors-reference.html
```
raymondfeng: https://loopback.io/doc/en/lb4/Connectors-reference.html does not exist. Am I missing something here?

raymondfeng: Just realized https://github.com/strongloop/loopback.io/pull/863

hacksparrow: There are many more "external" files, all of which are present in the `lb3` dir or made available there eventually. Those files are copied over to the `lb4` dir by the build process in `loopback.io`.

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -415,6 +414,194 @@ children:
...
+
+    - title: 'Cassandra connector'
+      url: Cassandra-connector.html
+      output: 'web, pdf'
+
+    - title: 'Cloudant connector'
+      url: Cloudant-connector.html
+      output: 'web, pdf'
+
+    - title: DashDB
```
jannyHou: `DashDB` --> `DashDB connector` to be consistent with others?

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -415,6 +414,194 @@ children:
...
+  - title: 'Database connectors'
+    url: Database-connectors.html
+    output: 'web'
+    children:
+
+    - title: 'Cassandra connector'
+      url: Cassandra-connector.html
+      output: 'web, pdf'
+
+    - title: 'Cloudant connector'
```
jannyHou: nitpick: same as https://github.com/strongloop/loopback.io/pull/863#discussion_r309302677, we also support couchdb2. If it takes time to add the new entry we can do it in a separate PR.

hacksparrow: Let's do it in a separate PR.

PR #3489, opened 2019-08-01
feat(core): add service decorator to inject by service class/interface

raymondfeng:
Sometimes it's desirable to inject an instance of a known class/interface. This PR introduces `@service` as a shortcut for `@inject`.
 
```ts
class MyController {
  constructor(@service(MyService) public myService: MyService) {}
} 
```

Or even simpler:
```ts
class MyController {
  // MyService is inferred from design type
  constructor(@service() public myService: MyService) {}
}
```

Please note `MyService` has to be bound to the context with `toClass()` or have a `serviceInterface` tag that identifies the service interface (string, symbol, or Function). We have to use a unique string or symbol to represent an interface as TypeScript interfaces cannot be reflected in runtime.

```ts
const ctx = new Context();
ctx.bind('my-service').toClass(MyService);
ctx.bind('my-service').toProvider(MyServiceProvider).tag({[CoreTags.SERVICE_INTERFACE]: MyService};
ctx.bind('my-service').toDynamicValue(() => {}).apply(asService(MyService));
```

See https://github.com/hoangleanhtu/loopback4-spring

## Checklist


packages/core/src/service.ts
```diff
@@ -0,0 +1,121 @@
...
+    (ctx, injection, session) => {
+      let serviceType: Function | undefined = serviceClass;
+      if (!serviceType) {
+        if (typeof injection.methodDescriptorOrParameterIndex === 'number') {
+          serviceType = MetadataInspector.getDesignTypeForMethod(
+            injection.target,
+            injection.member!,
+          ).parameterTypes[injection.methodDescriptorOrParameterIndex];
+        } else {
+          serviceType = MetadataInspector.getDesignTypeForProperty(
```
agnes512: This PR looks good to me. 
I just have a question since I am not that familiar with `injection`. 
The case that `injection.methodDescriptorOrParameterIndex === 'number'` is for when the decorator is applied to a method parameter. 
How about `MetadataInspector.getDesignTypeForProperty`? Does it mean that decorator is applied to a method property? Could you explain a bit? thanks!

raymondfeng:
@bajtos Would you please review it before I merge?

packages/core/src/__tests__/acceptance/service.acceptance.ts
```diff
@@ -0,0 +1,143 @@
...
+    await expect(
+      ctx.get<MyController>('controllers.MyController'),
+    ).to.be.rejectedWith(/Service class cannot be inferred from design type/);
+  });
+
+  class MyService {}
+
+  class MySubService extends MyService {}
+
+  function givenContext() {
```
bajtos: I find it surprising that this helper is binding `MyService`. Let's make that clear please.

```suggestion
  function givenContextWithMyService() {
```

packages/core/src/service.ts
```diff
@@ -0,0 +1,122 @@
...
+            injection.target,
+            injection.member!,
+          );
+        }
+      }
+      if (serviceType === Object || serviceType === Array) {
+        throw new Error('Service class cannot be inferred from design type');
```
bajtos: Please improve the error message to tell the user how to fix the problem. Should they provide the service class as the first decorator argument? Something else?

packages/core/src/service.ts
```diff
@@ -0,0 +1,122 @@
...
+        if (values.length === 1) return values[0];
+        if (values.length >= 1) {
+          throw new Error(
+            `More than one bindings found for ${serviceType!.name}`,
+          );
+        } else {
+          if (metadata && metadata.optional) {
+            return undefined;
+          }
+          throw new Error(`No binding found for ${serviceType!.name}`);
```
bajtos: Similarly here, please improve the error message to tell the user how to fix the problem.

packages/core/src/service.ts
```diff
@@ -0,0 +1,122 @@
...
+      const view = new ContextView(
+        ctx,
+        filterByServiceClass(serviceType, metadata),
+      );
+      const result = view.resolve(session);
+      return transformValueOrPromise(result, values => {
+        if (values.length === 1) return values[0];
+        if (values.length >= 1) {
+          throw new Error(
+            `More than one bindings found for ${serviceType!.name}`,
```
bajtos: Please improve the error message to help the user to better understand where are those bindings coming from. A list of binding names would be great. Also tell the user how to fix the problem.

It's a best practice to make error messages actionable.

bajtos:
I like the idea of a `@service` decorator that's inferring the binding key from the service class :+1:

In the tests, you are binding the service via `ctx.bind`. How is the new decorator going to work with `app.service` API? 

Also, what are the similarities and (more importantly) differences from `@repository` and `app.respository()` APIs that we already have in place?

> I'm open to move this code to `@loopback/context` package as `@inject.service`.

-1 from me. Personally, I'd prefer to keep Context as a low-level generic IoC container and implement higher-level concepts like `@repository` and `@service` in 3rd party modules. Considering that `app.service()` API lives in `@loopback/core`, it makes most sense to me to implement `@service` decorator in core too.

packages/core/src/__tests__/acceptance/service.acceptance.ts
```diff
@@ -0,0 +1,143 @@
...
+    expect(controller.myService).to.be.instanceOf(MyService);
+  });
+
+  it('injects a service matching a sub class', async () => {
+    class MyController {
+      constructor(@service(MyService) public myService: MyService) {}
+    }
+    ctx.unbind('services.MyService');
+    ctx.bind('services.MySubService').toClass(MySubService);
+    ctx.bind('controllers.MyController').toClass(MyController);
```
bajtos: I am concerned about the implications of this feature on user (developer) experience. Personally, I would find it surprising that `@service(MyService)` is resolved using the binding at key `services.MySubService`. If you look at the code wiring up the dependencies, there is no indication whatsoever that binding `services.MySubService` will automatically setup resolution of `services.MyService` too.

```ts
ctx.bind('services.MySubService').toClass(MySubService);
```

Let's find a better solution please. For example, we can make it explicit that one service class will be injected in places asking for another class.

```ts
// Resolve `MyService` service using `MySubService` implementation
ctx.service(MyService).toAlias(MySubService);

// Or maybe the other way around?
ctx.service(MySubService, {aliases: [MyService]})
```

Alternatively, let's keep handling of subclasses out of scope of this initial pull implementation and discuss it in a follow-up pull request.

raymondfeng:
@bajtos Thank you for the feedback. I think I might have confused you with the poor choice of `@service`. The PR was intended to add a dependency injection sugar for bindings created with `toClass`. It's the DI flavor for the `instantiateClass` helper method.

What about refactoring the code as follows?

1. Rename `@service` to `@instance` or `@inject.instance`
2. Move it to `@loopback/context`
3. Improve error message
4. Add documentation

For the record, it's a bit hard to have `@service` for bindings created by `app.service()` using classes.

bajtos:
> For the record, it's a bit hard to have `@service` for bindings created by `app.service()` using classes.

I see. Can we perhaps improve `app.service()` to make that easier?

>  The PR was intended to add a dependency injection sugar for bindings created with `toClass`. It's the DI flavor for the `instantiateClass` helper method.

Makes sense :+1: 

The proposed `@instance` decorator makes sense as a low-level tool.

Personally, I'd prefer to take the opportunity to review this area at high level and come up with a proposal how should LB projects deal with different kinds of services. Some services are implemented as proxies calling external (web) services, some services are implemented as classes. Some aspects are common for both approaches, others are different. 

Both classes-based and juggler-based services are "a service". If we use "services" only for service proxies, then we need to find a different term to use for utility classes implemented locally (e.g. JWT token ~~service~~ something). Considering how ambiguous the term "service" is, I think we would be better of to find a solution that will allow us to use the term "service" for both external (web)services and local utility classes.

I don't want to end up in a situation where we (and our users) frequently use the term "service" for something that's very different from what LB4 understands as a service.

raymondfeng:
@bajtos I refactored the code into `@loopback/context` and rename `@service` to `@instance` to avoid confusion. Documentation has been updated too. PTAL.

raymondfeng:
> Personally, I'd prefer to take the opportunity to review this area at high level and come up with a proposal how should LB projects deal with different kinds of services. Some services are implemented as proxies calling external (web) services, some services are implemented as classes. Some aspects are common for both approaches, others are different.

> Both classes-based and juggler-based services are "a service". If we use "services" only for service proxies, then we need to find a different term to use for utility classes implemented locally (e.g. JWT token service something). Considering how ambiguous the term "service" is, I think we would be better of to find a solution that will allow us to use the term "service" for both external (web)services and local utility classes.

> I don't want to end up in a situation where we (and our users) frequently use the term "service" for something that's very different from what LB4 understands as a service.

With the introduction of `lb4 service` and `ServiceBooter`, we have generalized the `service` concept for both local and remote ones. See https://loopback.io/doc/en/lb4/Service-generator.html

docs/site/decorators/Decorators_inject.md
```diff
@@ -178,6 +178,39 @@ using `binding.to(value)`.
...
+
+`@instance` injects an instance of a given class from a binding created using
+`toClass()`.
+
+Syntax:
+`@instance(serviceClass?: Constructor, metadata: {skipSubClasses?: boolean})`.
+
+```ts
+class MyController {
+  constructor(@instance(MyService) public myService: MyService) {}
```
bajtos: What are the pros and cons of `@instance(MyService)` when compared to something like `@inject.instanceOf(MyService)`? Personally, I slightly prefer the longer & more descriptive variant.

docs/site/decorators/Decorators_inject.md
```diff
@@ -178,6 +178,39 @@ using `binding.to(value)`.
...
+
+Please note the class has to be bound to the context with toClass().
+
+```ts
+const ctx = new Context();
+ctx.bind('my-service').toClass(MyService);
+```
+
+To prevent subclasses from being matched, set `skipSubClasses` to `false` for
+the `metadata` argument.
```
bajtos: As mentioned earlier, I find it potentially confusing that sub-classes are searched by default. How about starting with a stricter behavior where only the actual class is considered, and making sub-class search an opt-in feature? I.e. rename `skipSubClasses` to something like `searchSubClasses`.

_I am cross-posting the original comment below._

---

I am concerned about the implications of this feature on user (developer) experience. Personally, I would find it surprising that `@service(MyService)` is resolved using the binding at key `services.MySubService`. If you look at the code wiring up the dependencies, there is no indication whatsoever that binding `services.MySubService` will automatically setup resolution of `services.MyService` too.

```ts
ctx.bind('services.MySubService').toClass(MySubService);
```

Let's find a better solution please. For example, we can make it explicit that one service class will be injected in places asking for another class.

```ts
// Resolve `MyService` service using `MySubService` implementation
ctx.service(MyService).toAlias(MySubService);

// Or maybe the other way around?
ctx.service(MySubService, {aliases: [MyService]})
```

Alternatively, let's keep handling of subclasses out of scope of this initial pull implementation and discuss it in a follow-up pull request.

raymondfeng:
@bajtos PTAL

docs/site/decorators/Decorators_inject.md
bajtos: Let's go with `@inject.instance` then ✅

docs/site/decorators/Decorators_inject.md
bajtos: > We have a default Logger class from a component
> The application wants to have its own flavor.

First of all, I agree this is a valid use case that we should (eventually) support.

IMO, the application should bind `Logger` to `MyLogger` class in that case, to make it clear that we are overriding `Logger` to be resolved as `MyLogger`.

Both your examples are adding additional binding metadata that can be used by the consumer to look up the bound subclass, therefore they look slightly different from the problem I would like to discuss.

Let's take a look at the unit test from your pull request that I find problematic:

```ts
  it('injects an instance matching a sub class', async () => {
    class MyController {
      constructor(@instance(MyService) public myService: MyService) {}
    }
    ctx.unbind('services.MyService');
    ctx.bind('services.MySubService').toClass(MySubService);
    ctx.bind('controllers.MyController').toClass(MyController);
	
    const controller = await ctx.get<MyController>('controllers.MyController');
    expect(controller.myService).to.be.instanceOf(MySubService);
  });
```

When reading the code setting up bindings, it's not clear that `MyService` will be resolved to `MySubService` class.

I would like to have a more explicit setup, for example:

```ts
    ctx.bind('services.MyService').toClass(MySubService);
```

Another aspect I find potentially confusing - instance injection is comparing class constructors, it completely ignores binding keys. I believe the following setup would pass your test too:

```ts
    ctx.bind('repositories.MySubService').toClass(MySubService);
```

Is it a good idea to resolve `services.MyService` via `repositories.MySubService`?

The class-constructor check is also not very flexible. What if the application wants to use an async factory function to obtain the class instance? Because `@instance` is using class-based check, it won't be able to match a binding backed by a factory function (`.toDynamicValue()`).

I feel we need to revisit this pull request at high level. What scenario (user story) do we want to support & simplify by this new decorator? Are the extensibility options (e.g. subclass matching) worth the overhead in implementation complexity and the decrease in clarity of how are bindings resolved in a project using the extensibility options?

raymondfeng:
The feature is mostly inspired by Java and https://github.com/inversify/InversifyJS/blob/master/wiki/classes_as_id.md. The `class` is used as identifier to bindings in addition to keys. With `@loopback/context`, I think we can leverage a tag such as `serviceInterface` (class, symbol, etc) to access bindings.

raymondfeng:
@strongloop/loopback-maintainers I have re-worked the PR to align with `service` concept. Basically, `@service` can now be used to inject services registered with `app.service()` or created using `lb4 service` command. PTAL.

raymondfeng:
@bajtos PTAL

emonddr:
Just a question about the need for `@inject.instance` decorator mentioned above.

 When we use dependency injection:

```
@inject( some binding key) private someVariable someType;
```
 
isn't an instance object (of a type) placed into `someVariable`?

It is not clear what purpose this new decorator serves.

packages/core/src/service.ts
```diff
@@ -0,0 +1,180 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
```
emonddr: In LB4 documentation, we talk about services in many different ways, so i find this PR confusion somewhat.

`web services`: 

https://loopback.io/doc/en/lb4/Calling-other-APIs-and-web-services.html#add-a-service

`Remote and local services`:

https://loopback.io/doc/en/lb4/Service-generator.html#remote-and-local-services

`Connecting to SOAP services` https://loopback.io/doc/en/lb4/Connecting-to-SOAP.html 


`Add a Service` https://loopback.io/doc/en/lb4/soap-calculator-tutorial-add-service.html#add-a-service 
(talks about soap service here)

`Service` is not a key concept listed in our documentation either:

image

bajtos: I think this may be the same question as have been discussed in https://github.com/strongloop/loopback-next/pull/3489#issuecomment-523982645

raymondfeng:
> Just a question about the need for @inject.instance decorator mentioned above.

It's `@service`. No more `@inject.instance`. The motivation is briefly explained at https://github.com/strongloop/loopback-next/pull/3489#issue-303218574. This is very similar as `@repository` decorator for repositories.

emonddr:
> > Just a question about the need for @inject.instance decorator mentioned above.
> 
> It's `@service`. No more `@inject.instance`. The motivation is briefly explained at [#3489 (comment)](https://github.com/strongloop/loopback-next/pull/3489#issue-303218574). This is very similar as `@repository` decorator for repositories.

I realize that we are now going with @service (which I still have questions about), but was wondering why  @inject.instanceOf  was being entertained at all. Aren't we always injecting object instance of a class? 
(This is for my curiosity...to try to understand existing injection decorators and any proposed new ones.)

packages/core/src/__tests__/acceptance/service.acceptance.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    return expect(
+      ctx.get<MyController>('controllers.MyController'),
+    ).to.be.rejectedWith(/No binding found for Symbol\(MyService\)/);
+  });
+
+  it('allows skipSubClasses flag', async () => {
+    class MyController {
+      constructor(
+        @service(MyService, {skipSubClasses: true})
```
bajtos: AFAICT, the flag `skipSubClasses` is no longer implemented. I think this test is no longer relevant. Can you remove it?

raymondfeng:
@bajtos PTAL.

docs/site/Services.md
```diff
@@ -0,0 +1,112 @@
...
+- Bind a service class or provider with an interface identified by a string
+
+  ```ts
+  // Register a service and mark the interface as `MyService`
+  const binding = app.service(MyService, {interface: 'MyService'});
+  ```
+
+  ```ts
+  // Register a service provider and mark the interface as `MyService`
+  const binding = app.service(MyServiceProvider, {interface: 'MyService'});
```
bajtos: Can you please explain the rules we are using to decide whether the supplied constructor is a service class or a provider class?

docs/site/Services.md
```diff
@@ -0,0 +1,112 @@
...
+  ```
+
+- Bind a service with with an interface identified by a symbol
```
bajtos: Typo: `with with`

Issue #3480, opened 2019-07-31
Unable to create a related model


bhupesh-sf:
## Steps to reproduce

1. Create a new LB4 app.
2. Create 2 models employee and employee-salary-packages
3. use LB4 relation command from CLI to define hasMany relation between employee and employee-salary-packages models.
4. Now try to create employee-salary-packages and it throws an exception. 

Following are the models and the request:

## Employee Model
```js
import { model, property, hasMany } from '@loopback/repository';
import { UserModifiableEntity } from '.';
import { Location } from '../modules/employee/location.enum';
import { EmployeeStatus } from '../modules/employee/employee-status.enum';
import { EmployeeSalaryPackage } from './employee-salary-package.model';

@model({ settings: {} })
export class Employee extends UserModifiableEntity {
  @property({
    type: 'number',
    id: true,
  })
  id?: number;

  @property({
    type: 'string',
    required: true,
  })
  name: string;

  @property({
    type: 'string',
    required: true,
  })
  code: string;

  @property({
    type: 'number',
    required: true,
  })
  department_id: number;

  @property({
    type: 'number',
    required: true,
  })
  designation_id: number;

  @property({
    type: 'string',
    required: true,
    default: 'current',
  })
  status: EmployeeStatus;

  @property({
    type: 'string',
  })
  jira_id?: string;

  @property({
    type: 'boolean',
  })
  is_billable?: boolean;

  @property({
    type: 'date',
    required: true,
  })
  joining_date: string;

  @property({
    type: 'date',
  })
  relieving_date?: string;

  @property({
    type: 'string',
    required: true,
  })
  location_id: Location;

  @hasMany(() => EmployeeSalaryPackage, { keyTo: 'employee_id' })
  employeeSalaryPackages: EmployeeSalaryPackage[];

  constructor(data?: Partial<Employee>) {
    super(data);
  }
}

export interface EmployeeRelations {
  // describe navigational properties here
}

export type EmployeeWithRelations = Employee & EmployeeRelations;
```

## Employee Salary Package model
```js
import { model, property } from '@loopback/repository';
import { UserModifiableEntity } from '.';
import { Currency } from '../modules/employee/currency.enum';

@model({ settings: {} })
export class EmployeeSalaryPackage extends UserModifiableEntity {
  @property({
    type: 'number',
    id: true,
  })
  id?: number;

  @property({
    type: 'number',
    required: true,
  })
  employee_id: number;

  @property({
    type: 'number',
    required: true,
[...]

PR #3495, opened 2019-08-01
fix(cli): set foreignKey to be optional in the requestBody of `@hasMany`

agnes512:
connect to https://github.com/strongloop/loopback-next/issues/3480
 will update relevant docs later.

## Checklist


packages/cli/generators/relation/templates/controller-relation-template-has-many.ts.ejs
```diff
@@ -58,7 +58,7 @@ export class <%= controllerClassName %> {
     @requestBody({
       content: {
         'application/json': {
-          schema: getModelSchemaRef(<%= targetModelClassName %>, {exclude: ['<%= targetModelPrimaryKey %>']}),
+          schema: getModelSchemaRef(<%= targetModelClassName %>, {exclude: ['<%= targetModelPrimaryKey %>'], optional: ['<%= foreignKeyName %>']}),
```
nabdelgadir: Would it be possible to break this into new lines? e.g.:
```
schema: getModelSchemaRef(<%= targetModelClassName %>, {
  exclude: ['<%= targetModelPrimaryKey %>'],
  optional: ['<%= foreignKeyName %>'],
}),
```

bhupesh-sf: even putting the **foreignKeyName** in exclude list does the trick. Not sure how this new optional key helps?

raymondfeng: You can try to generate the code with your PR and format it with `npm run prettier:fix`. Use the formatted code as reference to improve the template.

raymondfeng: We can debate if `foreignKeyName` should be marked as `exclude` or `optional`. For example, to create an order for a customer, the FK of order is `customerId`. It is automatically populated/enforced.

I think mark it `optional` and make sure the value does not violate the constrain is reasonable.

nabdelgadir: I think it makes sense to keep it as `optional` and allow the developer to choose whether or not to include it in the request body. 

Currently, if it is included and is different than the one passed in the endpoint, the user gets an error saying they're not allowed to change the foreign key value.

agnes512: fixed

agnes512: I agree that `optional` makes more sense.
But I don't know if `exclude` should work as `optional`.

bhupesh-sf: @agnes512 @raymondfeng When we work through related models general behaviour as per REST conventions also the child is being created for the parent passed in request URL. For e.g.  if url is **customer/<customerId>/order** this means one is trying to create an order with the given customer id in URL. Here the context which we are talking is of the controller created by LB4 relation which is for relational model i.e. **CustomerOrderController** not the **OrderController**. So in this case customer_id which is the foreign key in order must have the same value as of URL.
Right now I have to manually go to the controller and write the code to fetch the ID from URL and attach it to request body which I believe should happen automatically.

raymondfeng: @bhupesh-sf I'm not sure if the LB runtime will have the knowledge to fetch the `customerId` from the url and attach it to the requestBody. What do you in mind to achieve that? BTW, the url can even skip `:customerId` as it can be inferred from the logged user.

The only (theoretical) possibility is to allow request body to have `variables` that can resolved from the context, such as:
```
{
   "customerId": ${user.id},
}
```

PR #3498, opened 2019-08-01
fix(repository-json-schema): make exclude option reject properties

nabdelgadir:
Before this, the `exclude` option would act as the `optional` option and would not reject any properties passed into the request body.

One thing to note is that the `details` of the error message is not helpful when rejecting e.g.:

```json
{
  "error": {
    "statusCode": 422,
    "name": "UnprocessableEntityError",
    "message": "The request body is invalid. See error object `details` property for more info.",
    "code": "VALIDATION_FAILED",
    "details": [
      {
        "path": "",
        "code": "not",
        "message": "should NOT be valid",
        "info": {}
      }
    ]
  }
}
```

## Checklist

### DIFF
--- a/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
+++ b/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
@@ -67,6 +67,15 @@ describe('TodoApplication', () => {
       .expect(422);
   });
 
+  it('rejects requests with input that contains excluded properties', async () => {
+    const todo = givenTodo();
+    todo.id = 1;
+    await client
+      .post('/todos')
+      .send(todo)
+      .expect(422);
+  });
+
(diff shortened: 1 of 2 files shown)


examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -67,6 +67,15 @@ describe('TodoApplication', () => {
       .expect(422);
   });
 
+  it('rejects requests to create a todo with an id', async () => {
```
nabdelgadir: Any suggestions for a better place to test this fix are appreciated 🙇‍♀️

raymondfeng: Maybe something like `rejects requests with input that contains excluded properties`?

Issue #3501, opened 2019-08-03
The documentation is not updated


frbuceta:
The documentation in loopback.io is not being updated according to this repository

frbuceta:
To have the updated version I have done:

* `npm ci` in the repository root
* `npm run build:site` WATCH OUT! [bundler](http://bundler.io) must be installed
* then I have run `npm start` in `/sandbox/loopback.io`

dhmlau:
Closing as resolved because PR #3502 has landed. @frbuceta, thanks for your contribution!

frbuceta:
The page is still not updated

dhmlau:
@frbuceta , my apologies. I thought your PR is fixing the issue you're complaining.  Do you have the link of the documentation?  And are you interested in submitting a PR for that? Thanks.

frbuceta:
No problem. For example `https://loopback.io/doc/en/lb4/Model.html` not equal to `loopback-next/docs/site/Model.md`

I really don't know what the problem is ... I guess you'll have this automated and it won't be working properly

raymondfeng:
For some reason, the build does not pick up latest version of `@loopback/docs` and reverted my manual updates.

dhmlau:
It should be in sync now.

frbuceta:
Yes, now, yes
Good job guys;)

PR #3502, opened 2019-08-03
Check if bundler is installed

frbuceta:
This implementation will return an error message if the bundle command does not exist, therefore it does not execute the entire code until the command does not exist

This comes because it happened to me when I tried to do this https://github.com/strongloop/loopback-next/issues/3501#issuecomment-517951472

## Checklist

### DIFF
--- a/bin/build-docs-site.sh
+++ b/bin/build-docs-site.sh
@@ -11,6 +11,11 @@
 # Set `-e` so that non-zero exit code from any step will be honored
 set -e
 
+if ! [ -x "$(command -v bundle)" ]; then
+  echo 'Error: You must have Bundler installed (http://bundler.io)' >&2
+  exit 1
+fi
+
 # Set `-v` (verbose) for travis build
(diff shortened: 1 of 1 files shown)


bin/build-docs-site.sh
```diff
@@ -11,6 +11,11 @@
 # Set `-e` so that non-zero exit code from any step will be honored
 set -e
 
+if ! [ -x "$(command -v bundle)" ]; then
```
raymondfeng: Is `command` available on Linux? What about `which bundle`?

frbuceta: Answer in stackoverflow: https://stackoverflow.com/a/677212/11729993
Command in my linux (Solus OS): https://gofile.io/?c=zycrvN

frbuceta: I think it's a small modification that will be useful. For example, I didn't know what it was and I had to investigate online

Issue #1466, opened 2018-06-25
[Spike] Deploy LB4 applications to Microsoft Azure


bajtos:
Find out what's needed to deploy LB4 applications to Microsoft Azure and write a short document describing the steps. 

See #1446 for more details and acceptance criteria.

dhmlau:
Moving out of GA scope, keeping deploying to IBM Cloud as the first priority in the same epic.

ericzon:
First of all, thanks for this nice work.
I was able to deploy a LB 4 API in Azure (connected with a CosmosDB nosql) as App Service.
During the process I found two main problems:
- The password that azure generates to connect to db sometimes contains characters that are encoded during the process of initialize the conection in loopback mongodb connector.

- When I connect this API to Azure API Manager, creating a controller with default CRUD operations, in the POST definition, having something like this:
```
@requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(MyData, {exclude: ['id']}),
        },
      },
    })
```
The "exclude" and "optional" keys generates an OpenAPI schema like "MyDataExcluding[id]" that contains brackets. These characters do not pass the validation when you try to add this new API
![image](https://user-images.githubusercontent.com/1809508/62422565-dc6cd800-b6b4-11e9-9a42-100dad59e5f5.png)
I got this message:
Parsing error(s): The key 'MyDataExcluding[id]' in 'schemas' of components MUST match the regular expression '^[a-zA-Z0-9\.\-_]+$'. [#/components]

I suggest you to modify the method build-schema.js / getTitleSuffix to use any valid value or at least make it configurable from the app.

If you need it, I can create ticket for these two issues.

dhmlau:
@ericzon, thanks for trying out the deployment to Azure. :)  
Seems like you were able to find the cause of the problem, are you interested in submitting a PR?

ericzon:
Yes, there's my PR :)

developingjosh:
@ericzon  Can you explain what steps you have used to deploy to azure app service on a windows web app? 

I have tried multiple methods and it all seems to fail. Although I have gotten this to work on a linux web app.

developingjosh:
### Deploying to Azure App Service Linux Web App
Inside the Azure Portal> Configuration > Application Settings add the following:
BUILD_FLAGS : Off

Inside the Azure Portal> Configuration > General Settings add the following to the Startup Command:
pm2 start --no-daemon index.js

atreyarc:
[...]

PR #3504, opened 2019-08-04
fix(repository-json-schema): update to pass openapi validation in azure

ericzon:
## Checklist

- [X] New tests added or existing tests modified to cover all changes
- [X] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [X] API Documentation in code was updated

This PR is related with my comment in https://github.com/strongloop/loopback-next/issues/1466 about problems found during the process of deploy an API in Azure.
With this PR is possible to deploy an API as "App Service" and connect it with the API Manager as OpenAPI.
I added the optional key 'itExtendsCompatibility' (boolean) in JsonSchemaOptions to tackle these cases. By default uses the current behaviour.
Added some tests, applied prettier & linter of the project.

P.S: by the way, trying to sign your CLA gives a 504 Error in Cloudfront.

### DIFF
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -770,6 +770,7 @@ describe('build-schema', () => {
         definitions: {
           ProductWithRelations: {
             title: 'ProductWithRelations',
+            description: `(Schema options: { includeRelations: true })`,
             properties: {
               id: {type: 'number'},
               categoryId: {type: 'number'},
@@ -785,6 +786,7 @@ describe('build-schema', () => {
           },
         },
         title: 'CategoryWithRelations',
+        description: `(Schema options: { includeRelations: true })`,
       };
       const jsonSchema = getJsonSchema(Category, {includeRelations: true});
       expect(jsonSchema).to.deepEqual(expectedSchema);
@@ -809,6 +811,7 @@ describe('build-schema', () => {
         definitions: {
           ProductWithRelations: {
             title: 'ProductWithRelations',
+            description: `(Schema options: { includeRelations: true })`,
             properties: {
               id: {type: 'number'},
               categoryId: {type: 'number'},
@@ -825,6 +828,7 @@ describe('build-schema', () => {
           },
         },
         title: 'CategoryWithoutPropWithRelations',
+        description: `(Schema options: { includeRelations: true })`,
       };
 
       // To check for case when there are no other properties than relational
@@ -988,7 +992,10 @@ describe('build-schema', () => {
           name: {type: 'string'},
           description: {type: 'string'},
         });
-        expect(excludeIdSchema.title).to.equal('ProductExcluding[id]');
+        expect(excludeIdSchema.title).to.equal('ProductExcluding_id_');
+        expect(excludeIdSchema.description).to.endWith(
+          `(Schema options: { exclude: [ 'id' ] })`,
+        );
       });
 
       it('excludes multiple properties when the option "exclude" is set to exclude multiple properties', () => {
@@ -1007,7 +1014,10 @@ describe('build-schema', () => {
           description: {type: 'string'},
         });
         expect(excludeIdAndNameSchema.title).to.equal(
-          'ProductExcluding[id,name]',
+          'ProductExcluding_id-name_',
+        );
+        expect(excludeIdAndNameSchema.description).to.endWith(
+          `(Schema options: { exclude: [ 'id', 'name' ] })`,
         );
       });
 
@@ -1050,7 +1060,10 @@ describe('build-schema', () => {
 
         const optionalIdSchema = getJsonSchema(Product, {optional: ['id']});
         expect(optionalIdSchema.required).to.deepEqual(['name']);
-        expect(optionalIdSchema.title).to.equal('ProductOptional[id]');
+        expect(optionalIdSchema.title).to.equal('ProductOptional_id_');
+        expect(optionalIdSchema.description).to.endWith(
+          `(Schema options: { optional: [ 'id' ] })`,
+        );
       });
 
       it('makes multiple properties optional when the option "optional" includes multiple properties', () => {
@@ -1063,7 +1076,10 @@ describe('build-schema', () => {
         });
         expect(optionalIdAndNameSchema.required).to.equal(undefined);
         expect(optionalIdAndNameSchema.title).to.equal(
-          'ProductOptional[id,name]',
+          'ProductOptional_id-name_',
+        );
+        expect(optionalIdAndNameSchema.description).to.endWith(
+          `(Schema options: { optional: [ 'id', 'name' ] })`,
         );
       });
 
@@ -1087,14 +1103,20 @@ describe('build-schema', () => {
           optional: ['name'],
         });
         expect(optionalNameSchema.required).to.deepEqual(['id']);
-        expect(optionalNameSchema.title).to.equal('ProductOptional[name]');
+        expect(optionalNameSchema.title).to.equal('ProductOptional_name_');
+        expect(optionalNameSchema.description).to.endWith(
+          `(Schema options: { optional: [ 'name' ] })`,
+        );
 
         optionalNameSchema = getJsonSchema(Product, {
           partial: false,
           optional: ['name'],
         });
         expect(optionalNameSchema.required).to.deepEqual(['id']);
-        expect(optionalNameSchema.title).to.equal('ProductOptional[name]');
+        expect(optionalNameSchema.title).to.equal('ProductOptional_name_');
+        expect(optionalNameSchema.description).to.endWith(
+          `(Schema options: { optional: [ 'name' ] })`,
+        );
       });
 
       it('uses "partial" option, if provided, when "optional" option is set but empty', () => {
@@ -1108,6 +1130,23 @@ describe('build-schema', () => {
         });
         expect(optionalNameSchema.required).to.equal(undefined);
         expect(optionalNameSchema.title).to.equal('ProductPartial');
+      });
+
+      it('can work with "optional" and "exclude" options together', () => {
+        const originalSchema = getJsonSchema(Product);
+        expect(originalSchema.required).to.deepEqual(['id', 'name']);
+        expect(originalSchema.title).to.equal('Product');
+
+        const bothOptionsSchema = getJsonSchema(Product, {
+          exclude: ['id'],
+          optional: ['name'],
+        });
+        expect(bothOptionsSchema.title).to.equal(
+          'ProductOptional_name_Excluding_id_',
+        );
+        expect(bothOptionsSchema.description).to.endWith(
+          `(Schema options: { exclude: [ 'id' ], optional: [ 'name' ] })`,
+        );
       });
     });
 
--- a/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
+++ b/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
@@ -269,9 +269,9 @@ describe('build-schema', () => {
       expect(key).to.equal('modelPartial');
     });
 
-    it('returns "excluding[id,_rev]" when a single option "exclude" is set', () => {
+    it('returns "excluding_id-_rev_" when a single option "exclude" is set', () => {
       const key = buildModelCacheKey({exclude: ['id', '_rev']});
-      expect(key).to.equal('modelExcluding[id,_rev]');
+      expect(key).to.equal('modelExcluding_id-_rev_');
     });
 
     it('does not include "exclude" in concatenated option names if it is empty', () => {
@@ -283,9 +283,9 @@ describe('build-schema', () => {
       expect(key).to.equal('modelPartialWithRelations');
     });
 
-    it('returns "optional[id,_rev]" when "optional" is set with two items', () => {
+    it('returns "optional_id-_rev_" when "optional" is set with two items', () => {
       const key = buildModelCacheKey({optional: ['id', '_rev']});
-      expect(key).to.equal('modelOptional[id,_rev]');
+      expect(key).to.equal('modelOptional_id-_rev_');
     });
 
     it('does not include "optional" in concatenated option names if it is empty', () => {
@@ -302,7 +302,7 @@ describe('build-schema', () => {
         partial: true,
         optional: ['name'],
       });
-      expect(key).to.equal('modelOptional[name]');
+      expect(key).to.equal('modelOptional_name_');
     });
 
     it('includes "partial" in option names if "optional" is empty', () => {
@@ -322,7 +322,7 @@ describe('build-schema', () => {
         includeRelations: true,
       });
       expect(key).to.equal(
-        'modelOptional[name]Excluding[id,_rev]WithRelations',
+        'modelOptional_name_Excluding_id-_rev_WithRelations',
       );
     });
 
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -14,6 +14,7 @@ import {
 } from '@loopback/repository';
 import * as debugFactory from 'debug';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
+import {inspect} from 'util';
 import {JSON_SCHEMA_KEY, MODEL_TYPE_KEYS} from './keys';
 const debug = debugFactory('loopback:repository-json-schema:build-schema');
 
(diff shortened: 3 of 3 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -42,6 +42,11 @@ export interface JsonSchemaOptions<T extends object> {
    */
   optional?: (keyof T)[];
 
+  /**
+   * Boolean to generate fieldNames maximizing compatibility (Azure)
+   */
+  itExtendsCompatibility?: boolean;
```
raymondfeng: I think we should just fix the title with compatible chars without introducing an extra flag. We should also populate `description` to have more meaningful details.

raymondfeng:
@ericzon Thank you for the patch. I had a similar concern in the past - see https://github.com/strongloop/loopback-next/pull/3309#discussion_r301258717.

packages/repository-json-schema/src/build-schema.ts
ericzon: I added that flag to give the opportunity to the users to keep the easy to understand format of 'Exclude|Optional[key1,key2...]' that you designed, but if you prefer to keep it simple and leave always a compatible format, ok! I'll check the description too.

dhmlau:
FYI - the failure on Travis is a timeout error:
```
1) cloneExampleFromGitHub (SLOW)
       extracts project files:
     Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/strongloop/loopback-next/packages/cli/test/integration/generators/clone-example.integration.js)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)
```

dhmlau:
@slnode test please

raymondfeng:
@nabdelgadir Please review the PR.

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -251,6 +259,14 @@ describe('build-schema', () => {
       expect(key).to.equal('modelOptional[id,_rev]');
     });
 
+    it('returns "optional[id,_rev]" when "optional" is set with two items (compatible mode)', () => {
```
nabdelgadir: `returns "optional_id-_rev_"`?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -35,10 +42,15 @@ export interface JsonSchemaOptions<T extends object> {
    */
   optional?: (keyof T)[];
 
+  /**
+   * Boolean to generate fieldNames maximizing compatibility (Azure)
```
nabdelgadir: Can we make this more consistent with the others? E.g. `Set this flag to generate fieldNames maximizing compatibility (Azure)`? 

Also question, are `fieldNames` specifically for Azure?

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
ericzon: Yep, good catch, fixed in my last commit, thanks.

packages/repository-json-schema/src/build-schema.ts
ericzon: Description updated. The new format is less expressive but more compatible with all the possible formats in platforms apart from Azure. Checking the OpenAPI 3.0 spec, it should be fully compatible.
https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json#L196

packages/repository-json-schema/src/build-schema.ts
ericzon: @raymondfeng please, check the current status of this PR and decide which option to take:
- I keep along with this flag
or
- apply the compatible format without flag, loosing expressivity

nabdelgadir: I personally think we should keep this as a flag (first option) because the way it is right now is more readable.

raymondfeng: What I propose is as follows:

1. Make schema `title` compatible by default with a bit less expressivity (IIRC, the `title` needs to be unique)
2. Populate schema `description` to make it readable

dhmlau:
@nabdelgadir, is this good to land?

nabdelgadir:
@dhmlau not yet.

@ericzon, what do you think about @raymondfeng's proposal?

> 1. Make schema title compatible by default with a bit less expressivity (IIRC, the title needs to be unique)
> 2. Populate schema description to make it readable

dhmlau:
Thanks @nabdelgadir . Let me assign it to you since you're already helping this PR.

ericzon:
> @dhmlau not yet.
> 
> @ericzon, what do you think about @raymondfeng's proposal?
> 
> > 1. Make schema title compatible by default with a bit less expressivity (IIRC, the title needs to be unique)
> > 2. Populate schema description to make it readable

@nabdelgadir for me it's ok, compatible by default. @raymondfeng in 2 you propose to add some kind of extra & dynamic description like:

"description": "<original description>... Product excluding fields id, name" 

right?

packages/repository-json-schema/src/build-schema.ts
bajtos: Personally, I am against too many feature flags (see [Decision Fatique](https://en.wikipedia.org/wiki/Decision_fatigue)). IMO, a good tool should pick a reasonable behavior so that users don't have to learn about different options to be able to make a qualified decision.

Let's follow Raymond's proposal.

>1. Make schema `title` compatible by default with a bit less expressivity (IIRC, the `title` needs to be unique)
> 2. Populate schema `description` to make it readable

raymondfeng:
@ericzon Would you please make the final update based on the team agreement? Thanks!

ericzon:
Sure! sorry but I had a busy week. Tomorrow I update the PR with the
agreements.

ericzon:
Updated. Now is compatible by default & description includes information about model options.

Examples:

Description about TechnicalData model
Description about TechnicalData model (Model options: Excluding[id,createdAt])
Description about TechnicalData model (Model options: Partial)
Description about TechnicalData model (Model options: WithRelations)

I tested in Azure to import the generated definition and is accepted.
...

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -271,15 +271,54 @@ export function getNavigationalPropertyForRelation(
...
+  options: JsonSchemaOptions<T> = {}
+) {
+  const sep = {
+    start: '[',
+    end: ']',
+    delimiter: ',',
+  };
+  const suffix = getTextSuffix(options, sep);
+
+  return suffix !== '' ?  `(Model options: ${suffix})` : '';
```
bajtos: 1. Let's use `Schema options` (or `Model schema options`) instead of `Model options` please. The options are changing the schema generated from the model, the model itself remains unchanged.

2. I am not sure if the complexity of the current implementation is worth the benefits. Have you considered using JSON instead of a custom text format?

    ```ts
    return options ? JSON.stringify(options, null, 2) : undefined;
   ```

    Alternatively, if we want something slightly prettier than JSON, then let's use [stringify-object](https://www.npmjs.com/package/stringify-object), we have had a good experience with that package in `lb4` CLI.

    https://github.com/strongloop/loopback-next/blob/c944654881140119e44fe9e1db85a0f26e59d658/packages/cli/lib/utils.js#L539-L548

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -1059,14 +1071,16 @@ describe('build-schema', () => {
...
+        expect(optionalNameSchema.description).to.endWith('(Model options: Optional[name])');
 
         optionalNameSchema = getJsonSchema(Product, {
           partial: false,
           optional: ['name'],
         });
         expect(optionalNameSchema.required).to.deepEqual(['id']);
-        expect(optionalNameSchema.title).to.equal('ProductOptional[name]');
+        expect(optionalNameSchema.title).to.equal('ProductOptional_name_');
+        expect(optionalNameSchema.description).to.endWith('(Model options: Optional[name])');
```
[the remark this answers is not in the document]
ericzon: 1. Done although travis build failed (it seems that the error is not related with code "Invalid revision range")

packages/repository-json-schema/src/build-schema.ts
ericzon: 1. Done
2. I took that approach to keep readability reusing logic applied to title suffix. To clarify, you suggest putting the stringified version of options as description suffix?

bajtos: > To clarify, you suggest putting the stringified version of options as description suffix?

Yes, exactly.

bajtos: We created the format `Optional[name]Excluding[id]` as a middle ground between what OpenAPI/JSON Schema supports and what is still readable by humans.

Since the description field is a free-form text, there is no longer need to limit ourselves to a compressed title, and we can use `JSON.stringify` or `stringify-object` for the best readability.

bajtos:
@ericzon I would love to see this pull request landed soon, let me know how can I help.

bajtos:
I would like the code building title suffix to be kept simple with hardcoded start/delimiter/end characters and also improve the code building options description to use `JSON.stringify` or `stringify-object`.

I feel the current proposal is adding too much complexity that will make the code more difficult to maintain for us.

ericzon:
@bajtos pushed a version with stringify-object. I couldn't reuse stringifyModelSettings from utils because it creates a wrapper with settings field and we want the stringified version in a single line.

packages/repository-json-schema/src/build-schema.ts
ericzon: changes applied (stringify-object)

raymondfeng:
@ericzon I have rebased your branch against latest master of strongloop/loopback-next. In the future, please use the following to pull in latest changes from upstream:

```sh
git remote add upstream git@github.com:strongloop/loopback-next.git
git fetch --all
git rebase upstream/master
```

bajtos:
> pushed a simplified version using stringify-object.
> I couldn't reuse stringifyModelSettings from utils because it creates a wrapper with settings field and we want the stringified version in a single line.

@ericzon I don't see this change being applied to the patch. I don't see any new commits pushed around the time you have posted the comment. Did you perhaps forgot to `git push` the change?

ericzon:
@bajtos now the changes are ready to check. I've squashed and rebase too.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -15,6 +15,7 @@ import {
 import * as debugFactory from 'debug';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY, MODEL_TYPE_KEYS} from './keys';
+const utils = require('../../cli/lib/utils');
```
raymondfeng: We cannot use `utils` from `@loopback/cli` as it's not a dependency of this module.

packages/cli/lib/utils.js
```diff
@@ -547,6 +547,14 @@ exports.stringifyModelSettings = function(modelSettings) {
...
 };
 
+exports.stringifyOptions = function(modelSettings = {}) {
+  return stringifyObject(modelSettings, {
+    indent: '  ', // two spaces
+    singleQuotes: true,
+    inlineCharacterLimit: Infinity,
+  });
+};
+
```
raymondfeng: Please move this utility to packages/repository-json-schema.

ericzon: I moved stringifyOptions utils to build-schema.ts (I thing there's not enough code to create its own utils file)

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -765,11 +766,12 @@ describe('build-schema', () => {
         @hasMany(() => Product)
         products?: Product[];
       }
-
+      const options: JsonSchemaOptions<Category> = {includeRelations: true};
```
bajtos: I feel the introduction of `options` property in the tests is adding too much complexity because we need to provide explicit type. I guess I can live with that though 🤷‍♂

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -307,6 +313,37 @@ function getTitleSuffix<T extends object>(options: JsonSchemaOptions<T> = {}) {
   return suffix;
 }
 
+function stringifyOptions(modelSettings: object = {}) {
+  return stringifyObject(modelSettings, {
+    indent: '  ', // two spaces
```
bajtos: Is this setting needed when `inlineCharacterLimit` is set to `Infinity`? I mean there should be never any line breaks, therefore there is nothing to indent.

Thoughts?

ericzon: Your right, in this case is not really needed. Pushed a version without it.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
ericzon: Simplified without options

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -15,6 +15,7 @@ import {
 import * as debugFactory from 'debug';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY, MODEL_TYPE_KEYS} from './keys';
+const stringifyObject = require('stringify-object');
```
raymondfeng: Can we simply use https://nodejs.org/api/util.html#util_util_inspect_object_options?

ericzon: Ok, tried and it works smooth. I put these options:
```
{
	"depth": Infinity,
	"maxArrayLength": Infinity,
	"breakLength": Infinity
}
```
The style is slightly different but is readable too. i.e:
...(Schema options: { optional: [ 'id' ] })

packages/repository-json-schema/package-lock.json
```diff
@@ -4,23 +4,554 @@
...
+		"@hapi/topo": {
+			"version": "3.1.4",
+			"resolved": "https://registry.npmjs.org/@hapi/topo/-/topo-3.1.4.tgz",
+			"integrity": "sha512-aVWQTOI9wBD6zawmOr6f+tdEIxQC8JXfQVLTjgGe8YEStAWGn/GNNVTobKJhbWKveQj2RyYF3oYbO9SC8/eOCA==",
+			"dev": true,
+			"requires": {
+				"@hapi/hoek": "8.x.x"
+			}
+		},
+		"@loopback/build": {
```
bajtos: Please revert changes in `package-lock.json` file. Now that you are not adding new dependencies, there is no need to change the lock file either.

For future reference: https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#working-with-dependencies
Here is the important part: Dependencies resolved locally within the monorepo must be excluded from package-lock files.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -15,6 +15,7 @@ import {
 import * as debugFactory from 'debug';
 import {JSONSchema6 as JSONSchema} from 'json-schema';
 import {JSON_SCHEMA_KEY, MODEL_TYPE_KEYS} from './keys';
+const nativeUtils = require('util');
```
bajtos: Please use ES6 imports.

```suggestion
import {inspect} from 'util';
```

Or if you want to import everything:

```suggestion
import * as nativeUtils from 'util';
```

packages/repository-json-schema/package-lock.json
bajtos: Nice, this problem was catched by our linter, I forget we have one in place!

See https://travis-ci.com/strongloop/loopback-next/jobs/238791798

```
Invalid package-lock entries found!
  packages/repository-json-schema/package-lock.json
    -> @loopback/build
    -> @loopback/context
    -> @loopback/core
    -> @loopback/eslint-config
    -> @loopback/metadata
    -> @loopback/repository
    -> @loopback/testlab

Run the following command to fix the problems:

  $ npm run update-package-locks
```

Please don't run `update-package-locks` and just revert changes in this file. We prefer to update lock files via https://renovatebot.com

ericzon: Sorry, I followed the suggestion of travis.  Reverted.

packages/repository-json-schema/src/build-schema.ts
ericzon: Changes updated

Issue #3445, opened 2019-07-26
Introduce `InclusionResolver` concept


bajtos:
Introduce the concept of a inclusion resolver, which is a function we can use to fetch related models (both in our ORM/REST API but also from GraphQL resolvers in the future).

```ts
/**
 * @returns An array of resolved values, the items must be ordered in the same
 * way as `sourceEntities`. The resolved value can be one of:
 * - `undefined` when no target model(s) were found
 * - `Entity` for relations targeting a single model
 * - `Entity[]` for relations targeting multiple models
 */
export type InclusionResolver = (
  /**
   * List of source models as returned by the first database query.
   */
  sourceEntities: Entity[],
  /**
   * Inclusion requested by the user (e.g. scope constraints to apply).
   */
  inclusion: Inclusion,
  /**
   * Generic options object, e.g. carrying the Transaction object.
   */
  options?: Options,
) => Promise<(Entity | undefined)[] | (Entity[] | undefined)[]>;
```

See #3387 for more details & a prototype implementation.

## Acceptance criteria

- [ ] Introduce a new type `InclusionResolver`
- [ ] Implement a new helper function `includeRelatedModels`, it will become a part of a public API of `@loopback/repository`. Under the hood, this function should leverage inclusion resolvers to fetch related models.
- [ ] Write unit-level tests to verify the implementation, use mocked or stubbed inclusion resolvers.

Note: helper name `includeRelatedModels` is not final, feel free to propose a better one!

PR #3517, opened 2019-08-07
feat(repository): add InclusionResolver type and includeRelatedModels

agnes512:
resolves https://github.com/strongloop/loopback-next/issues/3445

When reviewing, select **`hide white space changes`** would make it easier to review :smile:

- Add InclusionResolver interface, includeRelatedModels and isInclusionAllowed helpers.
- Used stubbed resolvers for the unit tests.

## Checklist

### DIFF
--- a/packages/repository/src/relations/relation.helpers.ts
+++ b/packages/repository/src/relations/relation.helpers.ts
@@ -3,8 +3,18 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import * as debugFactory from 'debug';
 import * as _ from 'lodash';
-import {Entity, EntityCrudRepository, Filter, Options, Where} from '..';
+import {
+  AnyObject,
+  Entity,
+  EntityCrudRepository,
+  Filter,
+  Inclusion,
+  Options,
+  Where,
+} from '..';
+const debug = debugFactory('loopback:repository:relation-helpers');
 
 /**
  * Finds model instances that contain any of the provided foreign key values.
@@ -48,3 +58,77 @@ export async function findByForeignKeys<
 }
 
 type StringKeyOf<T> = Extract<keyof T, string>;
+
+/**
+ * Returns model instances that include related models that have a registered
+ * resolver.
+ *
+ * @param targetRepository - The target repository where the model instances are found
+ * @param entities - An array of entity instances or data
+ * @param include -Inclusion filter
+ * @param options - Options for the operations
+ */
+
+export async function includeRelatedModels<
+  T extends Entity,
+  Relations extends object = {}
+>(
+  targetRepository: EntityCrudRepository<T, unknown, Relations>,
+  entities: T[],
+  include?: Inclusion<T>[],
+  options?: Options,
+): Promise<(T & Relations)[]> {
+  const result = entities as (T & Relations)[];
+  if (!include) return result;
+
+  const invalidInclusions = include.filter(
+    inclusionFilter => !isInclusionAllowed(targetRepository, inclusionFilter),
+  );
+  if (invalidInclusions.length) {
+    const msg =
+      'Invalid "filter.include" entries: ' +
+      invalidInclusions
+        .map(inclusionFilter => JSON.stringify(inclusionFilter))
+        .join('; ');
+    const err = new Error(msg);
+    Object.assign(err, {
+      code: 'INVALID_INCLUSION_FILTER',
+    });
+    throw err;
+  }
+
+  const resolveTasks = include.map(async inclusionFilter => {
+    const relationName = inclusionFilter.relation;
+    const resolver = targetRepository.inclusionResolvers.get(relationName)!;
+    const targets = await resolver(entities, inclusionFilter, options);
+
+    result.forEach((entity, ix) => {
+      const src = entity as AnyObject;
+      src[relationName] = targets[ix];
+    });
+  });
+
+  await Promise.all(resolveTasks);
+
+  return result;
+}
+/**
+ * Checks if the resolver of the inclusion relation is registered
+ * in the inclusionResolver of the target repository
+ *
+ * @param targetRepository - The target repository where the relations are registered
+ * @param include - Inclusion filter
+ */
+function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
+  targetRepository: EntityCrudRepository<T, unknown, Relations>,
+  include: Inclusion,
+): boolean {
+  const relationName = include.relation;
+  if (!relationName) {
+    debug('isInclusionAllowed for %j? No: missing relation name', include);
+    return false;
+  }
+  const allowed = targetRepository.inclusionResolvers.has(relationName);
+  debug('isInclusionAllowed for %j (relation %s)? %s', include, allowed);
+  return allowed;
+}
--- a/packages/repository/src/relations/relation.types.ts
+++ b/packages/repository/src/relations/relation.types.ts
@@ -3,7 +3,9 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
+import {Options} from '../common-types';
 import {Entity} from '../model';
+import {Inclusion} from '../query';
 import {TypeResolver} from '../type-resolver';
 
 export enum RelationType {
@@ -108,3 +110,25 @@ export type RelationMetadata =
 
 // Re-export Getter so that users don't have to import from @loopback/context
 export {Getter} from '@loopback/context';
+
+/**
+ * @returns An array of resolved values, the items must be ordered in the same
+ * way as `sourceEntities`. The resolved value can be one of:
+ * - `undefined` when no target model(s) were found
+ * - `Entity` for relations targeting a single model
+ * - `Entity[]` for relations targeting multiple models
+ */
+export type InclusionResolver<S extends Entity, T extends Entity> = (
+  /**
+   * List of source models as returned by the first database query.
+   */
+  sourceEntities: S[],
+  /**
+   * Inclusion requested by the user (e.g. scope constraints to apply).
+   */
+  inclusion: Inclusion,
+  /**
+   * Generic options object, e.g. carrying the Transaction object.
+   */
+  options?: Options,
+) => Promise<(T | undefined)[] | (T[] | undefined)[]>;
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -28,6 +28,7 @@ import {
   HasManyRepositoryFactory,
   HasOneDefinition,
   HasOneRepositoryFactory,
+  InclusionResolver,
 } from '../relations';
 import {IsolationLevel, Transaction} from '../transaction';
 import {isTypeResolver, resolveType} from '../type-resolver';
@@ -99,6 +100,11 @@ export class DefaultCrudRepository<
   Relations extends object = {}
 > implements EntityCrudRepository<T, ID, Relations> {
   modelClass: juggler.PersistedModelClass;
+
+  public readonly inclusionResolvers: Map<
+    string,
+    InclusionResolver<T, Entity>
+  > = new Map();
 
   /**
    * Constructor of DefaultCrudRepository
--- a/packages/repository/src/repositories/repository.ts
+++ b/packages/repository/src/repositories/repository.ts
@@ -17,6 +17,7 @@ import {DataSource} from '../datasource';
 import {EntityNotFoundError} from '../errors';
 import {Entity, Model, ValueObject} from '../model';
 import {Filter, Where} from '../query';
+import {InclusionResolver} from '../relations/relation.types';
 import {IsolationLevel, Transaction} from '../transaction';
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
@@ -140,6 +141,7 @@ export interface EntityCrudRepository<
 > extends EntityRepository<T, ID>, CrudRepository<T, Relations> {
   // entityClass should have type "typeof T", but that's not supported by TSC
   entityClass: typeof Entity & {prototype: T};
+  inclusionResolvers: Map<string, InclusionResolver<T, Entity>>;
 
   /**
    * Save an entity. If no id is present, create a new entity
@@ -246,6 +248,10 @@ export interface EntityCrudRepository<
 export class CrudRepositoryImpl<T extends Entity, ID>
   implements EntityCrudRepository<T, ID> {
   private connector: CrudConnector;
+  public readonly inclusionResolvers: Map<
+    string,
+    InclusionResolver<T, Entity>
+  > = new Map();
 
   constructor(
     public dataSource: DataSource,
--- a/packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.helpers.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.helpers.unit.ts
@@ -4,9 +4,8 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {expect} from '@loopback/testlab';
-import {DefaultCrudRepository, findByForeignKeys, juggler} from '../../..';
-import {model, property} from '../../../decorators';
-import {Entity} from '../../../model';
+import {findByForeignKeys} from '../../../..';
+import {ProductRepository, testdb} from './relations-helpers-fixtures';
 
 describe('findByForeignKeys', () => {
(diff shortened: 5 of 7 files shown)


packages/repository/src/relations/relation.helpers.ts
```diff
@@ -48,3 +58,77 @@ export async function findByForeignKeys<
...
+    const resolver = targetRepository.inclusionResolvers.get(relationName)!;
+    const targets = await resolver(entities, i, options);
+
+    for (const ix in result) {
+      const src = result[ix];
+      (src as AnyObject)[relationName] = targets[ix];
+    }
+  });
+
+  await Promise.all(resolveTasks);
```
jannyHou: I am not sure is it good to execute them in parallel...since the tasks operates on the same object.

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -48,3 +58,77 @@ export async function findByForeignKeys<
...
+export async function includeRelatedModels<
+  T extends Entity,
+  Relations extends object = {}
+>(
+  targetRepository: EntityCrudRepository<T, unknown, Relations>,
+  entities: T[],
+  filter?: Filter<T>,
+  options?: Options,
+): Promise<(T & Relations)[]> {
+  const result = entities as (T & Relations)[];
```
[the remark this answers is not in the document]
nabdelgadir: Hmm, but it's never modified; it's only read from.

packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts
```diff
@@ -3,95 +3,369 @@
...
+      expect(errorMessage).to.eql(
+        'Invalid "filter.include" entries: {"relation":"products"}',
+      );
+    });
+
+    it('returns an empty array if target model of the source entity does not have any matched instances', async () => {
+      const category = await categoryRepo.create({name: 'category'});
+      // stubbed resolver
+      const resolver: InclusionResolver = async entities => {
+        const c = entities[0] as Category;
```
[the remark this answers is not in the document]
nabdelgadir: If we don't, then it uses type `Entity`.

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -48,3 +58,77 @@ export async function findByForeignKeys<
...
+      'Invalid "filter.include" entries: ' +
+      invalidInclusions.map(i => JSON.stringify(i)).join('; ');
+    const err = new Error(msg);
+    Object.assign(err, {
+      code: 'INVALID_INCLUSION_FILTER',
+    });
+    throw err;
+  }
+
+  const resolveTasks = include.map(async i => {
```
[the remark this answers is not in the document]
nabdelgadir: `ix` is just an index, but we'll rename `i`.

packages/repository/src/relations/relation.helpers.ts
agnes512: > I am not sure is it good to execute them in parallel...since the tasks operates on the same object.

We are just reading it, not modifying it. So I guess it should be fine?

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -48,3 +58,85 @@ export async function findByForeignKeys<
...
+ * @param options - Options for the operations
+ */
+
+export async function includeRelatedModels<
+  T extends Entity,
+  Relations extends object = {}
+>(
+  targetRepository: EntityCrudRepository<T, unknown, Relations>,
+  entities: T[],
+  filter?: Filter<T>,
```
raymondfeng: Should it be `include?: Inclusion,`?

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -48,3 +58,85 @@ export async function findByForeignKeys<
...
+    });
+    throw err;
+  }
+
+  const resolveTasks = include.map(async inclusionFilter => {
+    const relationName = inclusionFilter.relation;
+    const resolver = targetRepository.inclusionResolvers.get(relationName)!;
+    const targets = await resolver(entities, inclusionFilter, options);
+
+    for (const ix in result) {
```
raymondfeng: Don't use `for in` loop for an array!

packages/repository/src/relations/relation.helpers.ts
agnes512: This method will be called by [`find`, `findById`](https://github.com/strongloop/loopback-next/pull/3387/files#diff-177253a1afbdd8eb83b99efef86dd415R393), etc. That's why we pass in `filter` instead of `include`.

packages/repository/src/relations/relation.types.ts
```diff
@@ -108,3 +110,25 @@ export type RelationMetadata =
...
 export {Getter} from '@loopback/context';
+
+/**
+ * @returns An array of resolved values, the items must be ordered in the same
+ * way as `sourceEntities`. The resolved value can be one of:
+ * - `undefined` when no target model(s) were found
+ * - `Entity` for relations targeting a single model
+ * - `Entity[]` for relations targeting multiple models
+ */
+export type InclusionResolver = (
```
jannyHou: Hmm....I couldn't remember is `InclusionResolver` proposed as a type or an interface, the title of PR says it's an interface... 🤔 

As a type, I think it should take a generic type as Source and another type as Return

```ts
export type InclusionResolver<
  S  extends Entity,
  R extends Entity
> = (
  /**
   * List of source models as returned by the first database query.
   */
  sourceEntities: S[],
  /**
   * Inclusion requested by the user (e.g. scope constraints to apply).
   */
  inclusion: Inclusion,
  /**
   * Generic options object, e.g. carrying the Transaction object.
   */
  options?: Options,
  // The return type here is a little bit weird, wouldn't
  // `Promise<R[]>` work?
  // I feel an empty array [] is still a valid value for type `R[]`
) => Promise<(R | undefined)[] | (R[] | undefined)[]>;
```

Thus you don't need to specify the type like https://github.com/strongloop/loopback-next/pull/3517/files#diff-63b076ff3fb794bb126c7d57d33c82bfR323

packages/repository/src/relations/relation.helpers.ts
raymondfeng: Yes, but this method does not use or honor other properties such as `where`, `fields` from the `Filter`, right? The parameter should be minimally constrained to allow `includeRelatedModels` to work, not the callers.

packages/repository/src/relations/relation.helpers.ts
nabdelgadir: Can we do:
```ts
result.forEach((entity, ix) => {
  const src = entity as AnyObject;
  src[relationName] = targets[ix];
});
```
? We need the index to get the element in `targets` that's for that specific source entity.

packages/repository/src/relations/relation.types.ts
raymondfeng: I don't see a need to specify `S` as it does not help the compiler.

nabdelgadir: > Hmm....I couldn't remember is InclusionResolver proposed as a type or an interface, the title of PR says it's an interface... 🤔

It should be a type. We got confused because of the [issue description](https://github.com/strongloop/loopback-next/issues/3445#issue-473228289) where it says interface. We'll fix the commit message/PR title, thanks!

> // The return type here is a little bit weird, wouldn't
>   // `Promise<R[]>` work?
>   // I feel an empty array [] is still a valid value for type `R[]`

So it returns a Promise that has either an array of entities (`(R | undefined)[]`) or an array of an array of entities (`(R[] | undefined)[]`). The former is for something like a `belongsTo` relation where it will return one target entity per source entity. The latter is for something like a `hasMany` relation where it will return multiple target entities per one source entity.

> I don't see a need to specify S as it does not help the compiler.

It helps us with not having to cast in the [case](https://github.com/strongloop/loopback-next/pull/3517/files#diff-63b076ff3fb794bb126c7d57d33c82bfR323) @jannyHou mentioned. I added it in [this commit](https://github.com/strongloop/loopback-next/pull/3517/commits/602aa1bd92af602a72864dacab2c03b0ba41f70b) but if we're fine with casting, I can remove it.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -100,6 +101,8 @@ export class DefaultCrudRepository<
 > implements EntityCrudRepository<T, ID, Relations> {
   modelClass: juggler.PersistedModelClass;
 
+  public inclusionResolvers: Map<string, InclusionResolver<T, Entity>>;
```
raymondfeng:
```ts
public readonly inclusionResolvers: Map<string, InclusionResolver<T, Entity>> = new Map();
```

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -121,6 +124,7 @@ export class DefaultCrudRepository<
     );
 
     this.modelClass = this.definePersistedModel(entityClass);
+    this.inclusionResolvers = new Map<string, InclusionResolver<T, Entity>>();
```
raymondfeng: Remove it if https://github.com/strongloop/loopback-next/pull/3517/files#r313934062 is resolved.

packages/repository/src/repositories/repository.ts
```diff
@@ -249,13 +251,15 @@ export interface EntityCrudRepository<
 export class CrudRepositoryImpl<T extends Entity, ID>
   implements EntityCrudRepository<T, ID> {
   private connector: CrudConnector;
+  public inclusionResolvers: Map<string, InclusionResolver<T, Entity>>;
```
raymondfeng: See https://github.com/strongloop/loopback-next/pull/3517/files#r313934062

packages/repository/src/relations/relation.types.ts
```diff
@@ -108,3 +110,25 @@ export type RelationMetadata =
...
+ * way as `sourceEntities`. The resolved value can be one of:
+ * - `undefined` when no target model(s) were found
+ * - `Entity` for relations targeting a single model
+ * - `Entity[]` for relations targeting multiple models
+ */
+export type InclusionResolver<S extends Entity, T extends Entity> = (
+  /**
+   * List of source models as returned by the first database query.
+   */
+  sourceEntities: S[],
```
raymondfeng: I wonder if `sourceEntities` needs to have a generic type `S`. It's not helping the compiler much.

jannyHou: @raymondfeng I suggested to add a generic type `S` to avoid specifying the source entity's type like

```ts
const resolver: InclusionResolver = async entities => {
        // you need to specify Category....
        const c = entities[0] as Category;
        const product: Product[] = await categoryRepo.products(c.id).find();
        return [product];
      };

      categoryRepo.inclusionResolvers.set('products', resolver);

```

In the resolver function.

See the current [resolver function](https://github.com/strongloop/loopback-next/blob/02fe72714e56e035495ef37fb72ab18922f15ca7/packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts#L321-L332)

vs

[previous resolver function](https://github.com/strongloop/loopback-next/blob/5e5ae8d8ccedfeea0d4c4cf7017ad5be88f45c6f/packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts#L178-L188)

raymondfeng: I see. It sounds good then.

jannyHou: On a second thought...maybe the resolver should be written as
```ts
const hasManyResolver: InclusionResolver = async (entities: Category[]) => {
    const products: Product[] = [];

    for (const category of entities) {
      const product = await categoryRepo.products(category.id).find();
      products.push(product);
    }
    return products;
  };
```
?

raymondfeng: The current signature with `S` seems to be better.

packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts
```diff
@@ -3,113 +3,398 @@
...
+  belongsTo,
+  BelongsToAccessor,
+  Getter,
+  hasMany,
+  HasManyRepositoryFactory,
+  InclusionResolver,
+} from '../../../relations';
 
-describe('findByForeignKeys', () => {
+describe('relation helpers', () => {
```
bajtos: This test file is growing quite large. Can we split it into two files please? One test file for `findByForeignKeys` (this can be hopefully just a rename from `relation.helpers.unit.ts` to something like `find-by-foreing-key.ts` or `relations-helpers/find-by-foreign.keys`, with no code changes needed), another test file for the new helper `includeRelatedModels`.

packages/repository/src/relations/relation.helpers.ts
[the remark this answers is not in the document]
bajtos: We are not modifying the `result` array itself, we are only modifying properties of the object stored in the array.

Conceptually:

```ts
result[0].products = /* list of products */
```

It's ok to use `const` here. TypeScript (or V8 engine) would throw an error if we were trying to re-assign `result` to something else.

packages/repository/src/relations/relation.helpers.ts
bajtos: Each task is resolving a single relation. Even if there were multiple relations configured to use the same navigational property (which is a programmer error, IMO), then this code would be still safe to execute in parallel. JavaScript is single threaded, parallel execution means that we fire all database queries at the same time, the callback (code executed after `await resolver(...)`) is always run one at a time. If we switched to serial execution, then we will be running one DB query at a time, slowing down the process.

packages/repository/src/relations/relation.types.ts
bajtos: IIRC, generic arguments are pretty useless here. At high level, we need a registry of resolvers, mapping from relation name (e.g. `products`) to a resolver. This registry cannot use generic types and has to fall back to `Entity`.

That's why I am using the following definition in my spike:

https://github.com/strongloop/loopback-next/blob/2fa5df67181cdcd23a5dce90c9c640fe75943cb8/packages/repository/src/relations/relation.types.ts#L121-L134

I am concerned that the current design creates a false sense of safety - the resolver is implemented with the assumption that it will be called with certain `S` and `T` types, but in reality the code calling it will always allow any `Entity`.

Having said that, I see how the stubbed resolvers are easier to implement when generic arguments are in place.

I don't have enough insight now to be able to judge what aspect is more important.

bajtos: Ha, I take my comment back. When I reviewed code below, I found that is able to provide a meaningful value for `S`.

```ts
public readonly inclusionResolvers: Map<
  string,
  InclusionResolver<T, Entity>
> = new Map();
```

Nice trick! I agree that we should keep the generic argument for the type of the source entity.

The remaining question is whether we need the second generic argument for the type of the target (related) entity?

packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts
```diff
@@ -3,113 +3,398 @@
...
+    it('returns source model if no filter is passed in', async () => {
+      const category = await categoryRepo.create({name: 'category 1'});
+      await categoryRepo.create({name: 'category 2'});
+      const result = await includeRelatedModels(categoryRepo, [category]);
+      expect(result).to.eql([category]);
+    });
+
+    it('throws error if the target repository does not have the registered resolver', async () => {
+      const category = await categoryRepo.create({name: 'category 1'});
+      await productRepo.create({id: 1, name: 'product1', categoryId: 1});
```
bajtos: Do we need this product to be created? IIUC, this test is never accessing any products, therefore it's safe to remove this line from this test to make it simpler.

packages/repository/src/relations/relation.helpers.ts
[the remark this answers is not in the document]
jannyHou: I see, @nabdelgadir @bajtos Thank you for the explanation~

packages/repository/src/relations/relation.helpers.ts
jannyHou: Thank you @agnes512 @bajtos 👍 fair enough.

packages/repository/src/__tests__/unit/repositories/relation.helpers.unit.ts
nabdelgadir: Removed 👍

packages/repository/src/relations/relation.types.ts
nabdelgadir: It would help the developer with the return type for each resolver (e.g. restricting to type `Product`), but I don't think it's too important.

PR #3520, opened 2019-08-07
fix: server union type contains null

jannyHou:
The `address()` method from `net` module has `null` in its union type, which results in build error in the `http-server` module. See pic 1:
Screen Shot 2019-08-07 at 6 31 20 PM

The type definition of `address` from `net` module:

Screen Shot 2019-08-07 at 6 31 56 PM

*I am not very familiar with the address resolve, so treated `null` as an empty string here. Should we throw error if it's null?*

## Checklist

### DIFF
--- a/packages/http-server/src/__tests__/integration/http-server.integration.ts
+++ b/packages/http-server/src/__tests__/integration/http-server.integration.ts
@@ -135,7 +135,7 @@ describe('HttpServer (integration)', () => {
       .to.have.property('address')
       .which.is.an.Object();
     await server.stop();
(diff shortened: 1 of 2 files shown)


jannyHou:
Closing.
@types/node@10.x passes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v10/net.d.ts#L59

Only @types/node@12.x fails:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/net.d.ts#L176

Issue #3442, opened 2019-07-26
Test relations against databases


bajtos:
Refactor existing integration & acceptance tests for relations, move most (or
all) of them to [repository-tests](https://github.com/strongloop/loopback-next/tree/master/packages/repository-tests) package.

I am not sure how much work will be required to make these tests pass against MongoDB (coercion between `ObjectID` vs. `string`). If it would require non-trivial amount of time, then:

- Introduce one or more new `CrudFeature` flags allowing MongoDB test suite to skip these new tests.
- Open a follow-up issue to investigate & fix the MongoDB test suite later.

See also #3387, and [here](https://github.com/strongloop/loopback-next/pull/3387/files#diff-1ac72022ee3c92074711610d69356271R594).

## Acceptance criteria

Tests to review & move:

- [ ] [`belongs-to.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/belongs-to.relation.acceptance.ts)
- [ ] [`has-many-without-di.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/has-many-without-di.relation.acceptance.ts)
- [ ] [`has-many.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/has-many.relation.acceptance.ts)
- [ ] [`has-one.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/acceptance/has-one.relation.acceptance.ts)
- [ ] [`relation.factory.integration.ts`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts) - this file needs to be split into multiple smaller files, one (or more) for each relation type.

Issue #3456, opened 2019-07-26
Spike: robust handling of ObjectID type for MongoDB


bajtos:
MongoDB is tricky.

- It uses a custom `ObjectID` type for primary keys.
- `ObjectID` is represented as a `string` when converted to JSON
- In queries, string values must be cast to ObjectID, otherwise they are not considered as the same value: `'some-id' !== ObjectID('some-id')`.

As a result, both PK and FK properties must use `ObjectID` as the type, and coercion must be applied where necessary.

Ideally, I'd like LB4 to define MongoDB PK and FKs as follows:

- `{type: 'string', mongodb: {dataType: 'ObjectID'}}`

Even better, `dataType: 'ObjectID'` should be automatically applied by the connector for PK and FKs referencing ObjectID PKs.

For example:

```ts
@model()
class Product {
  @property({
    type: 'string',
    generated: true,
    // ^^ implies dataType: 'ObjectID'
  })
  id: string;

  @property({
    type: 'string',
    references: {
      model: () => Category,
      property: 'id',
    },
    // ^^ implies dataType: 'ObjectID' when Category is attached to MongoDB
  })
  categoryId: string;
}
```

For v1, I suppose we can ask developers to provide dataType manually.

```ts
@model()
class Product {
  @property({
    type: 'string',
    generated: true,
    mongodb: {dataType: 'ObjectID'},
  })
  id: string;

  @property({
    type: 'string',
    mongodb: {dataType: 'ObjectID'},
  })
  categoryId: string;
}
```

With this setup in place, `id` and `categoryId` properties should be always returned as strings from DAO and connector methods.

This is tricky to achieve for the PK property, because juggler overrides user-supplied type with connector's default type when the PK is generated by the database. See
[`DataSource.prototype.setupDataAccess()`](https://github.com/strongloop/loopback-datasource-juggler/blob/0c2bb81dace3592ecde8b9eccbd70d589da44d7d/lib/datasource.js#L713-L719)

Can we rework MongoDB connector to hide ObjectID complexity as an internal implementation detail and always use string values in public APIs? Accept ids as strings and internally convert them to ObjectID. Convert values returned by the database from ObjectID to strings.

Downside: this is not going to work for free-form properties that don't have any type definition and where the connector does not know that they should be converted from string to ObjectID. But then keep in mind that JSON cannot carry type information, therefore REST API clients are not able to set free-form properties to ObjectID values even today.

See also https://github.com/strongloop/loopback-next/pull/3387

[...]

PR #3538, opened 2019-08-12
refactor(repository-tests): move relation tests to repository-tests package

agnes512:
resolves https://github.com/strongloop/loopback-next/issues/3442

- Move these tests to `repository-tests` package so that they would be tested against real databases such as MongoDB and MySQL.
- Introduce a new type `mixedType` for different id types usage
- Refactor Repository classes so that they are inherited from `CrudRepositoryCtor`.

`relation.factory.integration.ts` is split into `has-many.factory.integration.ts` and `belongs-to.factory.integration.ts`
## Checklist


agnes512:
@raymondfeng I don't know if we need to export these tests as functions like what we did in [this file](https://github.com/strongloop/loopback-next/blob/e9dca4cbcc8ebf4e2567cebdc046f3cf1d886060/packages/repository-tests/src/crud/create-retrieve.suite.ts#L19). 
We keep them as they are because they run fine.

packages/repository-tests/src/relations/acceptance/has-many-without-di.relation.acceptance.ts
```diff
@@ -0,0 +1,210 @@
...
+      const customerB = await customerRepo.create({name: 'customer B'});
+      const customerC = await customerRepo.create({name: 'customer c'});
+
+      const order = await createCustomerOrders(customerB.id, {
+        description: 'order 1',
+      });
+      console.log(typeof customerB.id);
+      console.log(customerB);
+      console.log(typeof order.id);
+      console.log(order);
```
bajtos: Forgotten console logs?

packages/repository-tests/src/relations/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    //--- HELPERS ---//
+
+    async function givenBoundCrudRepositories(db: juggler.DataSource) {
+      orderRepo = new OrderRepository(
+        db,
+        async () => customerRepo,
+        async () => shipmentRepo,
+      );
+      customerRepo = new repositoryClass(Customer, db) as CustomerRepository;
```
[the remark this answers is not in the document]
agnes512: We were trying to pass `db` to the `CustomerRepository` in [here](https://github.com/strongloop/loopback-next/pull/3538/files#diff-4015f24415fad8e76eed5dbafb60ec7fR41). I am not sure if it doesn't run on mongo case we solved some mongo failures from this file before.

packages/repository-tests/src/relations/fixtures/models/order.model.ts
```diff
@@ -12,8 +12,9 @@ export class Order extends Entity {
   @property({
     type: 'string',
     id: true,
+    generated: true,
   })
-  id: string;
+  id: number;
```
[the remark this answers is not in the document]
agnes512: Most of acceptance files use models/repos from `fixtures`. We were thinking to add them to test files individually so that we can set the type to `features.idType`. But that would be lots of repeating code.  
So we are using [this](https://github.com/strongloop/loopback-next/pull/3538/commits/9fa74e5abac210e7145d638c0ee09b036d587580#diff-70c77278b0038ca734ce5e04c92d8d24R44) as a workaround.
Do you have any suggestions to modify the types?

acceptance/repository-mongodb/src/__tests__/mongodb-default-repository.acceptance.ts
```diff
@@ -17,4 +18,11 @@ describe('MongoDB + DefaultCrudRepository', () => {
     DefaultCrudRepository as CrudRepositoryCtor,
     MONGODB_FEATURES,
   );
+
+  relationsRepositoryTestSuite(
```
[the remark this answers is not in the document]
nabdelgadir: We'll change these tests to be part of `crudRepositoryTestSuite` once we get the CI to pass 👍

packages/repository-tests/package.json
```diff
@@ -23,12 +23,17 @@
...
   "dependencies": {
-    "@loopback/testlab": "^1.7.3",
+    "loopback-datasource-juggler": "^4.11.0",
+    "lodash": "^4.17.15",
+    "@loopback/testlab": "^1.7.2",
     "@types/debug": "^4.1.5",
-    "debug": "^4.1.1"
+    "debug": "^4.1.1",
+    "@loopback/context": "^1.21.1",
+    "@loopback/core": "^1.9.0"
```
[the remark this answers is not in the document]
nabdelgadir: We used `'@loopback/context'` in our repository fixtures: `import {Getter, inject} from '@loopback/context';`, but I don't think we use `@loopback/core` anywhere so I'll remove it

packages/repository-tests/src/relations/acceptance/belongs-to.relation.acceptance.ts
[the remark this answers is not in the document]
nabdelgadir: We created a [helper file](https://github.com/strongloop/loopback-next/pull/3538/files#diff-5c34a6f28feccb05e86c469f2ed2aa61) to get the repositories. What do you think?

packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts
```diff
@@ -1,33 +1,45 @@
...
+  })
+  id: string;
   @property({
     type: 'string',
   })
   street: string;
   @property({
     type: 'string',
-    id: true,
+    default: '12345',
```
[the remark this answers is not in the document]
agnes512: If we don't set the default value, MySQL somehow would have `undefined` instead of `null`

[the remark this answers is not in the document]
nabdelgadir: > The CI tests are failing and seem related

They were just timeout failures. They're gone now.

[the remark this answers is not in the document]
jannyHou: > MySQL somehow would have undefined instead of null

This behavior sounds reasonable to me, 🤔 , why is it a problem? Some tests expect `null` instead of `undefined`? Or any API calls will be broken?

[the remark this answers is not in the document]
agnes512: If nothing is provided then loopback uses `undefined` but mysql uses `null`. And if we use `null` in lb4, it's not the same thing as it's in MySQL. That's why we use these default values.

packages/repository-tests/src/crud/relations/acceptance/has-many-without-di.relation.acceptance.ts
```diff
@@ -0,0 +1,205 @@
...
+      }
+      const order = await createCustomerOrders(existingCustomerId, {
+        description: 'order 1',
+      });
+      // avoid type problems of BSON type of mongodb
+      if (features.convertIdType) {
+        // eslint-disable-next-line require-atomic-updates
+        existingCustomerId = existingCustomerId.toString();
+      }
+      expect(order.toObject()).containDeep({
```
bajtos:
```ts
// avoid type problems of BSON type of mongodb
if (features.convertIdType) {
```

I find this very problematic. Does it mean that relations are not working with MongoDB?

Have you tried the following assertion instead?

```ts
expect(toJSON(order)).to.containDeep(toJSON({
  customerId: existingCustomerId,
  description: 'order 1',
});
```

If relations are not working with MongoDB, then let's introduce a flag to skip all relation tests. I am concerned that the current flag `features.convertIdType ` is modifying tests to execute scenarios that are different from what LB4 users are going to do.

BTW I think the underlying problem is in `toObject` that's behaving the same way as `toJSON` and converting ObjectID values to strings. I believe that in LB3, `toObject` was implemented differently and preserved `ObjectID` values. I think this is worth fixing, but not as part of this pull request - I opened a new issue https://github.com/strongloop/loopback-next/issues/3607.

packages/repository-tests/src/crud/relations/acceptance/has-many-without-di.relation.acceptance.ts
```diff
@@ -0,0 +1,205 @@
...
+      });
+      expect(persisted).to.deepEqual(orders);
+    });
+
+    //--- HELPERS ---//
+
+    // use strictObjectIDCoercion here to make sure mongo's happy
+    @model({
+      settings: {
+        strictObjectIDCoercion: true,
```
bajtos: Configuring connector-specific settings in the shared test suite creates unwanted coupling between the generic test suite (that should be connector-independent) and individual connectors. It makes it difficult for individual connectors to tweak these settings, because the change must be made in the shared test suite - this is even more difficult for 3rd party connectors. Let's find a better solution please.

I am proposing to introduce a new `features` flag allowing each connector to provide additional model settings that should be used by all models. Example usage here:

```ts
@model({
  settings: features.modelSettings,
})
class Order extends Entity {
  // ...
}
```

However! I believe the MongoDB connector allows users to enable `strictObjectIDCoercion` at dataSource-level, for all models connected to that datasource. I prefer to enable `strictObjectIDCoercion` that way (see [`MONGODB_CONFIG`](https://github.com/strongloop/loopback-next/blob/6ad0bb59c71ea2356cb951da786bdbff246b47e7/acceptance/repository-mongodb/src/__tests__/mongodb.datasource.ts#L10-L15) instead of introducing a new `CrudFeature` flag.

I'd also like to better understand why is it necessary to enable `strictObjectIDCoercion`? What is happening under the hood?

If `strictObjectIDCoercion` is necessary to make relations work, then we need to make this very clear in our documentation! Ideally, we should modify `lb4 datasource` generator to enable that flag in every datasource that's using the MongoDB connector. Docs & CLI updates are out of scope of this pull request, we should open a new issue to keep track of those changes. Ideally, that issue should explain why is that flag necessary (what is happening under the hood).

Very loosely related: In the future, I'd like us to rework MongoDB connector so that ObjectID type is never leaked outside of the connector, so that the Models and any user-facing code is using `string` instead. See https://github.com/strongloop/loopback-next/issues/3456

packages/repository-tests/src/crud/relations/acceptance/repository.acceptance.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export function hasOneRelationAcceptance(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: CrudRepositoryCtor,
+  features: CrudFeatures,
+) {
+  // This test shows the recommended way how to use @loopback/repository
+  // together with existing connectors when building LoopBack applications
+  describe('Repository in Thinking in LoopBack', () => {
+    let repo: ProductRepository;
```
bajtos: I am little bit confused about this file. It's stored in `packages/repository-tests/src/crud/relations/acceptance/repository.acceptance.ts` but it does not seem to leverage any relations. What am I missing?

We already have other test files to verify the core Repository APIs, see e.g. [packages/repository-tests/src/crud/freeform-properties.suite.ts](https://github.com/strongloop/loopback-next/blob/6ad0bb59c71ea2356cb951da786bdbff246b47e7/packages/repository-tests/src/crud/freeform-properties.suite.ts) that's verifying non-strict (free-form) mode.

packages/repository-tests/src/crud/relations/acceptance/repository.acceptance.ts
```diff
@@ -0,0 +1,167 @@
...
+          const created = await flexibleRepo.create({
+            extra: 'additional data',
+          } as AnyObject);
+          const stored = await flexibleRepo.findById(created.id);
+          expect(stored).to.containDeep({extra: 'additional data'});
+        });
+      },
+    );
+
+    it('allows models to allow nested model properties', async () => {
```
bajtos: Let's create a new test file in [`crud/`](https://github.com/strongloop/loopback-next/tree/6ad0bb59c71ea2356cb951da786bdbff246b47e7/packages/repository-tests/src/crud) directory for tests related to nested model properties, e.g. `crud/nested-model-properties.suite.ts`. We want to keep the test files short and each one focused on a single feature.

packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts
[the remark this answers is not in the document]
bajtos: It's a well-know problem (feature?) that NoSQL databases like `memory` and MongoDB represent missing values are `undefined`, while SQL databases use `NULL`. We have `features.emptyValue` to use in tests to deal with that design.

Please remove `default` fields from the property definitions and modify the checks to use `feature.emptyValue` instead of the default or `undefined`. 

See e.g. this test for inspiration:

https://github.com/strongloop/loopback-next/blob/6ad0bb59c71ea2356cb951da786bdbff246b47e7/packages/repository-tests/src/crud/replace-by-id.suite.ts#L73-L81

packages/repository-tests/src/crud/relations/fixtures/models/customer.model.ts
```diff
@@ -1,19 +1,31 @@
...
+  settings: {
+    strictObjectIDCoercion: true,
+  },
+})
 export class Customer extends Entity {
   @property({
-    type: 'number',
+    type: 'string',
     id: true,
+    generated: true,
```
bajtos: It makes me wonder - why is it necessary to set `generated` to `true`? Isn't it implied automatically for `id` properties? Is this needed to workaround ObjectID issues in MongoDB? 

Anyhow, if we need to enable `generated: true` to make relations work, then again this needs to be clearly documented in our docs and ideally `lb4 model` should set that flag automatically. Such change is out of scope of this pull request, as I mentioned earlier. I think we can open just one issue to make all changes needed to make relations work in MongoDB.

packages/repository-tests/src/crud/relations/fixtures/models/order.model.ts
```diff
@@ -24,14 +29,15 @@ export class Order extends Entity {
   @property({
     type: 'boolean',
     required: false,
+    default: false,
   })
   isShipped: boolean;
 
   @belongsTo(() => Customer)
-  customerId: number;
+  customerId: string;
```
bajtos: As I mentioned earlier, this should use `number | string` type, because many databases use numeric ids. Please fix this issue in all other places defining PK/FK properties.

packages/repository-tests/src/crud/relations/fixtures/models/product.model.ts
```diff
@@ -1,18 +1,23 @@
...
-@model()
+@model({
+  settings: {
+    strictObjectIDCoercion: true,
+  },
+})
 export class Product extends Entity {
   @property({
-    type: 'number',
+    type: 'string',
```
bajtos: I am not very happy that we are using a string type that's later changed to a number. It makes it too easy to forget to update the type from a test that's using this fixture. I would really like to the fixtures structured in a way that allows us to leverage `features.idType`.

packages/repository-tests/src/crud/relations/fixtures/repositories/product.repository.ts
```diff
@@ -3,9 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {DefaultCrudRepository, juggler} from '../../..';
+import {DefaultCrudRepository, juggler} from '@loopback/repository';
 import {Product, ProductRelations} from '../models/product.model';
-export {Product};
 
 export class ProductRepository extends DefaultCrudRepository<
```
bajtos: As I commented earlier, all model-specific repository classes must inherit from `repositoryClass` provided to the test suite by the caller. In the (near) future, we will want to run the shared test suite against `CrudRepositoryImpl` in addition to `DefaultCrudRepository`. See https://github.com/strongloop/loopback-next/blob/6ad0bb59c71ea2356cb951da786bdbff246b47e7/packages/repository/src/repositories/repository.ts#L248

packages/repository-tests/src/crud/relations/integration/belongs-to.factory.integration.ts
```diff
@@ -0,0 +1,146 @@
...
+      Customer,
+      typeof Order.prototype.id
+    >;
+
+    before(deleteAllModelsInDefaultDataSource);
+
+    before(
+      withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+        db = ctx.dataSource;
+        givenCrudRepositories();
```
bajtos: As I commented earlier, this needs to accept `repositoryClass` so that the repositories are inheriting from the right Repository implementation.

packages/repository-tests/src/crud/relations/acceptance/has-many-without-di.relation.acceptance.ts
bajtos: > I find this very problematic. Does it mean that relations are not working with MongoDB?

To clarify: IIUC, `covertIdType` transformation is required to make relations work. IMO, applications are not going to apply this `convertIdType` transformation. Does that mean relations won't work for such apps?

packages/repository-tests/src/crud-test-suite.ts
```diff
@@ -50,15 +50,40 @@ export function crudRepositoryTestSuite(
...
+    testFiles = fs.readdirSync(testRoot);
     testFiles = testFiles.filter(function(it) {
       return (
         !!require.extensions[path.extname(it).toLowerCase()] &&
         /\.suite\.[^.]+$/.test(it)
       );
     });
 
+    // relations folder tests
+    const folders = ['acceptance', 'integration'];
```
bajtos: All tests in `repository-tests` are meant to be acceptance-level. Let's keep things simple(r) and "promote" any tests that are considered to be integration-level in their original place to become acceptance-level here.

packages/repository-tests/src/crud/nested-model-properties.suite.ts
```diff
@@ -0,0 +1,94 @@
...
+        typeof User.prototype.id
+      >(User, db);
+
+      const models = [User, Role, Address];
+      await db.automigrate(models.map(m => m.name));
+
+      const user = {
+        name: 'foo',
+        roles: [{name: 'admin'}, {name: 'user'}],
+        address: {street: 'backstreet'},
```
bajtos: Let's split this test into two test cases please, to make it more obvious what are you trying to test. One test case for  `address: Address` (a single model instance), another for `roles: Role[]` (an array of model instances).

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -0,0 +1,80 @@
...
+export function belongsToRelationAcceptance(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: CrudRepositoryCtor,
+  features: CrudFeatures,
+) {
+  describe('BelongsTo relation (acceptance)', () => {
+    before(deleteAllModelsInDefaultDataSource);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let customerRepo: any;
```
bajtos: This is effectively disabling type checks, it's possible to write e.g `customerRepo.danceMakarena()` and the compiler will happily accept it.

Can we use `EntityCrudRepository<Customer, typeof Customer.prototype.id, {/*relations*/}>` instead?

I think it may be best to create a type alias in `fixtures/models` so that we don't have to repeat the long repository type definition.

```ts
// usage
import {Customer, CustomerRepository} from '../fixtures/models';

// implementation
@model()
export class Customer extends Entity {
  // ...
}

export interface CustomerRepository extends EntityCrudRepository<Customer/*...*/> {
  // define additional members like relation methods here
};
```

Just note that `EntityCrudRepository` is a type, therefore you cannot call `new CustomerRepository`. Feel free to find a better name than `CustomerRepository` to avoid possible confusion about whether the repo is a class or an interface.

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -0,0 +1,80 @@
...
+        name: 'Order McForder',
+      });
+      const order = await orderRepo.create({
+        customerId: customer.id,
+        description: 'Order from Order McForder, the hoarder of Mordor',
+      });
+
+      const result = await orderRepo.customer(order.id);
+      // don't need to check parentId at this point, but still need to pass it
+      // in here so that MySQL won't complain
```
bajtos: Interesting. I'd like to better understand what's going under the hood. Which model instance (`customer` vs. `result`) is missing `parentId` property? Maybe we need to fix the implementation of our relation repositories to add FK property?

Anyhow, I guess this can be fixed later, once this PR is done. Can you add `FIXME` keyword to the comment please (to make it easier to find) and perhaps open a follow-up issue to investigate more?

packages/repository-tests/src/crud/relations/acceptance/has-many-without-di.relation.acceptance.ts
```diff
@@ -0,0 +1,194 @@
...
+export function hasManyWithoutDIRelationAcceptance(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: CrudRepositoryCtor,
+  features: CrudFeatures,
+) {
+  describe('HasMany relation without DI (acceptance)', () => {
+    before(deleteAllModelsInDefaultDataSource);
+    // Given a Customer and Order models - see definitions at the bottom
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let existingCustomerId: any;
```
bajtos: Can we use `unknown` instead of `any`?

packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
```diff
@@ -0,0 +1,253 @@
...
+} from '../../../helpers.repository-tests';
+import {Customer, Order} from '../fixtures/models';
+import {givenBoundCrudRepositories, mixedIdType} from '../helpers';
+
+export function hasManyRelationAcceptance(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: CrudRepositoryCtor,
+  features: CrudFeatures,
+) {
+  describe('HasMany relation (acceptance)', () => {
```
bajtos: I am little bit confused about why we have two test files for `HasMany` relation?
- `describe('HasMany relation without DI (acceptance)',`
- `describe('HasMany relation (acceptance)'`

It's not clear to me what is the difference.

I feel `repository-tests` should not be dealing with different ways how relation repositories can be wired up (e.g. with DI or without DI). `repository-tests` should be focused on test cases verifying behavior that can be database specific and thus we want to test it against different databases. Whether the repositories are wired together using DI or not, that's something to test elsewhere - probably back in `repository`. 

Having said that, I am not sure if the tests in this file are still useful - maybe we can simply delete this file?

packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
```diff
@@ -0,0 +1,253 @@
...
+    before(deleteAllModelsInDefaultDataSource);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let customerRepo: any;
```
bajtos: Let's use a real type please, not `any`. (See one of my earlier comments.)

Same comment applies to all other places that are similar.

packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
```diff
@@ -0,0 +1,253 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let orderRepo: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let existingCustomerId: any;
```
bajtos: Can we use `unknown` instead of `any`?

Same comment applies to all other places using `any` for a variable storing an existing id.

packages/repository-tests/src/crud/relations/fixtures/models/customer.model.ts
bajtos: Makes sense :+1:

Does `lb4 model` set `generated: true` too? If not, then can you open a new GH issue to make that change please? Or even open a new PR, it may be faster than going through issue triage+estimation+planning process.

packages/repository-tests/src/crud/relations/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  const shipmentRepo: any = new shipmentRepoClass(db, async () => orderRepo);
+
+  const addressRepoClass = createAddressRepo(repositoryClass);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const addressRepo: any = new addressRepoClass(db, async () => customerRepo);
+
+  return {customerRepo, orderRepo, shipmentRepo, addressRepo};
+}
+
+export type mixedIdType = string | number;
```
bajtos: We prefer to use PascalCase for type names.

```suggestion
export type MixedIdType = string | number;
```

bajtos: Please add a tsdoc comment explaining the reasons for introducing this type. I think this type should be used by all CRUD tests, not only those specific to relations. Let's move this type definition to `packages/repository-tests/src/helpers.repository-tests.ts`.

packages/repository-tests/src/crud/relations/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+} from './fixtures/repositories';
+
+export function givenBoundCrudRepositories(
+  db: juggler.DataSource,
+  repositoryClass: CrudRepositoryCtor,
+) {
+  // get the repository class and create a new instance of it
+  const customerRepoClass = createCustomerRepo(repositoryClass);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const customerRepo: any = new customerRepoClass(
```
bajtos: As I commented earlier, I'd like us to find a way how to avoid `any` and use an interface describing the properties & methods of each model-specific repositories.

packages/repository-tests/src/crud/relations/integration/has-many.factory.integration.ts
```diff
@@ -0,0 +1,259 @@
...
+    before(
+      withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+        db = ctx.dataSource;
+        givenCrudRepositories();
+        await givenPersistedCustomerInstance();
+        givenConstrainedRepositories();
+        givenRepositoryFactoryFunctions();
+        await ctx.dataSource.automigrate(Review.name);
+        await ctx.dataSource.automigrate(Customer.name);
+        await ctx.dataSource.automigrate(Order.name);
```
bajtos: Should we use the same trick with arrays  that your do elsewhere?

```ts
await ctx.dataSource.automigrate([Review, Customer, Order].map(m => m.name));
```

packages/repository-tests/src/crud/relations/integration/has-many.factory.integration.ts
```diff
@@ -0,0 +1,259 @@
...
+  function givenCrudRepositories() {
+    customerRepo = new repositoryClass(Customer, db);
+    orderRepo = new repositoryClass(Order, db);
+    reviewRepo = new repositoryClass(Review, db);
+  }
+
+  //--- HELPERS ---//
+
+  class Order extends Entity {
+    id: string;
```
bajtos: I think this should be using `mixedIdType`?

On the second thought, do we really care about the actual type, considering that it's a union `number | string` and therefore we can do very little with those values? What if we used `unknown` everywhere we need to describe the type of a PK or a FK property?

packages/repository-tests/src/crud/relations/helpers.ts
bajtos: On the second thought, do we really care about the actual type, considering that it's a union `number | string` and therefore we can do very little with those values? What if we used `unknown` everywhere we need to describe the type of a PK or a FK property?

packages/repository-tests/src/crud/relations/integration/has-many.factory.integration.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    static definition = new ModelDefinition('Review')
+      .addProperty('id', {type: features.idType, id: true, generated: true})
+      .addProperty('description', {type: 'string', required: true})
+      .addProperty('authorId', {type: features.idType, required: false})
+      .addProperty('approvedId', {type: features.idType, required: false});
+  }
+
+  class Customer extends Entity {
+    id: string;
```
bajtos: Ditto, let's use `mixedIdType` or `unknown`. Same comment applies to all other similar model-property definitions.

packages/repository-tests/src/crud/nested-model-properties.suite.ts
```diff
@@ -0,0 +1,94 @@
...
+      @model()
+      class Address extends Entity {
+        @property()
+        street: string;
+      }
+
+      @model()
+      class User extends Entity {
+        @property({
+          type: 'number',
```
bajtos: I think this should be `features.idType`?

packages/repository-tests/src/crud/nested-model-properties.suite.ts
```diff
@@ -0,0 +1,94 @@
...
+          id: true,
+          generated: true,
+        })
+        id: number;
```
bajtos: I think this should be `MixedItType` or `unknown`?

packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts
```diff
@@ -1,20 +1,26 @@
...
 
-import {belongsTo, Entity, model, property} from '../../..';
+import {belongsTo, Entity, model, property} from '@loopback/repository';
 import {Customer, CustomerWithRelations} from './customer.model';
+import {mixedIdType} from '../../helpers';
 
 @model()
 export class Address extends Entity {
+  @property({
+    type: 'number',
```
bajtos: I think this needs to be `features.idType`?

packages/repository-tests/src/crud/relations/fixtures/models/customer.model.ts
```diff
@@ -1,19 +1,28 @@
...
+  property,
+} from '@loopback/repository';
 import {Address, AddressWithRelations} from './address.model';
 import {Order, OrderWithRelations} from './order.model';
+import {mixedIdType} from '../../helpers';
 
 @model()
 export class Customer extends Entity {
   @property({
     type: 'number',
```
bajtos: I think this needs to be `features.idType`?

packages/repository-tests/src/crud/relations/fixtures/models/order.model.ts
```diff
@@ -1,19 +1,21 @@
...
+import {belongsTo, Entity, model, property} from '@loopback/repository';
 import {Customer, CustomerWithRelations} from './customer.model';
 import {Shipment, ShipmentWithRelations} from './shipment.model';
+import {mixedIdType} from '../../helpers';
 
 @model()
 export class Order extends Entity {
   @property({
-    type: 'string',
+    type: 'number',
```
bajtos: I think this needs to be `features.idType`?

packages/repository-tests/src/crud/relations/fixtures/models/order.model.ts
```diff
@@ -24,14 +26,15 @@ export class Order extends Entity {
   @property({
     type: 'boolean',
     required: false,
+    default: false,
```
bajtos: Can we remove this default value please? I think it should not be needed after the changes you made recently, based on my older comments.

packages/repository-tests/src/crud/relations/fixtures/models/shipment.model.ts
```diff
@@ -1,18 +1,20 @@
...
 
-import {Entity, hasMany, model, property} from '../../..';
+import {Entity, hasMany, model, property} from '@loopback/repository';
 import {Order, OrderWithRelations} from './order.model';
+import {mixedIdType} from '../../helpers';
 
 @model()
 export class Shipment extends Entity {
   @property({
     type: 'number',
```
bajtos: I think this needs to be `features.idType`?

packages/repository-tests/src/crud/relations/fixtures/repositories/address.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+      typeof Address.prototype.id
+    >;
+
+    constructor(
+      db: juggler.DataSource,
+      customerRepositoryGetter: Getter<typeof repoClass.prototype>,
+    ) {
+      super(Address, db);
+      // cant use the method createHasManyRepositoryFactoryFor from DefaultCrud
+      // cause it's protected.
```
bajtos: Not only the method `createHasManyRepositoryFactoryFor` is protected in the default repository class, but you are also inheriting from `repoClass` which may not provide `createHasManyRepositoryFactoryFor` at all. 

I feel the current comment can be a bit misleading, could you please improve it?

packages/repository-tests/src/crud/relations/fixtures/repositories/address.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+  BelongsToAccessor,
+  juggler,
+  createBelongsToAccessor,
+  BelongsToDefinition,
+} from '@loopback/repository';
+import {Address, AddressRelations, Customer} from '../models';
+import {CrudRepositoryCtor} from '../../../..';
+
+export function createAddressRepo(repoClass: CrudRepositoryCtor) {
+  return class AddressRepository extends repoClass<
```
bajtos: Nice 👏

packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
agnes512: You're right. Before we changing the repositories to functions, those repos were using injection. We removed the injection part so that the repo classes can be created by functions `create{Name}RepoClass`. 
Will move this file back to `repository`.

packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts
agnes512: We didn't change the model files to `create{modelName}model` functions.

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
agnes512: We pass in features.emptyValue to the `customer` now it should be fine.

packages/repository-tests/src/crud/relations/fixtures/models/order.model.ts
agnes512: We decided to use `MixedIdType` instead of `features.idType` or `unknown`. 
The reason that we don't want to use `unknown` is that, ' in relations, we need an actual type for a foreign key type' -- cc @jannyHou

jannyHou: `unknown` would result in incompatible type error when use `Order`(or `Partial<Order>`) as a type to describe the data. My expected solution is using `features.idType` but I also understand the refactor is not trivial, if `MixedIdType` could get the tests pass in this PR I would vote for it :) We can improve it in another story.

bajtos: Fair enough, I am ok to use `MixedIdType` for now.

acceptance/repository-mysql/src/__tests__/mysql.datasource.ts
```diff
@@ -13,7 +13,7 @@ export const MYSQL_CONFIG: DataSourceOptions = {
   port: process.env.MYSQL_PORT || 3306,
   database: process.env.MYSQL_DATABASE || 'repository-tests',
   username: process.env.MYSQL_USER || 'root',
-  password: process.env.MYSQL_PASSWORD || '',
+  password: process.env.MYSQL_PASSWORD || 'pass',
```
bajtos: This change seems unrelated to "move relation tests to repository-tests package". Can you please revert it here and open a new PR to discuss it in isolation?

packages/repository-tests/package-lock.json
```diff
@@ -7,14 +7,116 @@
 		"@types/debug": {
 			"version": "4.1.5",
 			"resolved": "https://registry.npmjs.org/@types/debug/-/debug-4.1.5.tgz",
-			"integrity": "sha512-Q1y515GcOdTHgagaVFhHnIFQ38ygs/kmxdNpvpou+raI9UO3YZcHDngBSYKQklcKlvA7iuQlmIKbzvmxcOE9CQ=="
+			"integrity": "sha512-Q1y515GcOdTHgagaVFhHnIFQ38ygs/kmxdNpvpou+raI9UO3YZcHDngBSYKQklcKlvA7iuQlmIKbzvmxcOE9CQ==",
+			"dev": true
```
bajtos: Are these changes necessary for this PR? We are usually delegating package-lock updates to RenovateBot. Can you revert changes in all package-lock files please?

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -60,13 +70,12 @@ export function belongsToRelationAcceptance(
...
       const result = await orderRepo.customer(order.id);
-      // FIXME(agnes512)
-      // don't need to check parentId at this point, but still need to pass it
-      // in here so that MySQL won't complain
-      expect(toJSON({...result, parentId: 1})).to.deepEqual(
-        toJSON({...customer, parentId: 1}),
+      // adding parentId to customer so MySQL doesn't complain about null vs
+      // undefined
+      expect(toJSON(result)).to.deepEqual(
+        toJSON({...customer, parentId: features.emptyValue}),
```
bajtos: :+1:

packages/repository-tests/package.json
```diff
@@ -23,15 +23,18 @@
...
+    "@loopback/context": "^1.21.3",
     "@loopback/testlab": "^1.7.4",
-    "@types/debug": "^4.1.5",
+    "loopback-datasource-juggler": "^4.12.0",
+    "lodash": "^4.17.15",
     "debug": "^4.1.1"
   },
   "peerDependencies": {
-    "@loopback/repository": "^1.6.1"
+    "@loopback/repository": "^1.11.1"
```
bajtos: Are you using any feature that has been added after `1.6.1` was released? If not, then please revert this line, so that `@loopback/repository-tests` can be used with an older version of `@loopback/repository` too. (It's a good practice to keep the ranges in `peerDependencies` as wide as possible.)

packages/repository-tests/package.json
```diff
@@ -23,15 +23,18 @@
...
     "@types/debug": "^4.1.5",
-    "@types/node": "^10.14.17"
+    "@types/node": "^10.14.17",
+    "@types/lodash": "^4.14.137"
   },
   "dependencies": {
+    "@loopback/context": "^1.21.3",
     "@loopback/testlab": "^1.7.4",
-    "@types/debug": "^4.1.5",
+    "loopback-datasource-juggler": "^4.12.0",
```
bajtos: AFAICT, `loopback-datasource-jugler` is not used directly anywhere in the new code, can we remove it from the dependencies please?

packages/repository-tests/src/crud/relations/fixtures/models/customer.model.ts
bajtos: > Does `lb4 model` set `generated: true` too? If not, then can you open a new GH issue to make that change please? Or even open a new PR, it may be faster than going through issue triage+estimation+planning process.

☝️ PTAL

agnes512: @bajtos https://github.com/strongloop/loopback-next/pull/3643

Issue #2736, opened 2019-04-12
Extension: CrudRestController


bajtos:
_Part of From model definition to REST API with no custom repository/controller classes #2036_

A new extension (npm package living in loopback-next monorepo) providing CrudController class. See #740 for inspiration.

## Acceptance criteria

- [x] A new extension (let's call it `@loopback/rest-crud`) providing a base Controller class (let's call it `CrudRestController`) implementing typical CRUD operations using `CrudRepository` injected via constructor argument.

- [x] The methods provided by the controller should be exactly the same as we are scaffolding via `lb4 controller`

- [x] Minimal documentation - a "Basic Use" section in extension package's README is enough.

- [x] Test coverage, including end-to-end tests (making HTTP requests to an app).

### Out of scope

- Detailed documentation, will be done as part of #2036
- Changes in CLI and Example apps

bajtos:
Started the work here: https://github.com/strongloop/loopback-next/tree/feat/rest-crud

bajtos:
Pull request in progress: #3582 Early feedback welcome!

bajtos:
Closing as done.

Issue #2036, opened 2018-11-16
From model definition to REST API with no custom repository/controller classes


bajtos:
In LoopBack 3, it was very easy to get a fully-featured CRUD REST API with very little code: a model definition describing model properties + a model configuration specifying which datasource to use.

Let's provide the same simplicity to LB4 users too.

 - User creates a model class and uses decorators to define model properties. (No change here.)
 - User declaratively defines what kind of data-access patterns to provide (CRUD, KeyValue, etc.) and what datasource to use under the hood.
 - `@loopback/boot` processes this configuration and registers appropriate repositories & controllers with the app.

For example, to create a `Product` model, attach it to a `db` datasources and expose a CRUD REST API, all that the developer needs is to add the following two files to their project (plus any npm packages that implement this functionality):

`models/product.ts`:
```ts
@model()
export class Product extends Entity {
  @property()
  name: string;
}
```

`public-models/product.json`:
```json
{
  "model": "Product",
  "pattern": "Crud",
  "dataSource": "db",
}
```

Note: having config for multiple models in a single JSON file, as we do in LB 3.x, can quickly become a maintenance nightmare - see https://github.com/strongloop/loopback/issues/1316. It will be better to have one config file per model.

An example file for a model using KeyValue pattern (not CRUD) - `public-models/cache-entry.json`:

```json
{
  "model": "CacheEntry",
  "pattern": "KeyValue",
  "dataSource": "redis"
}
```

## See also

- https://github.com/strongloop/loopback-next/issues/1889:
    > provide a set of extensions in order to keep the simplicity of LB3, without the needed to creat e repository, controller for default CRUD actions (only if we need custom actions) ?

  /cc @GuillaumeJasmin

 - https://github.com/strongloop/loopback-next/issues/565 Declarative Support 

## Acceptance criteria

 - [x] A new extension (npm package living in loopback-next monorepo) providing CrudController class. See  #740 for inspiration. --> SEE https://github.com/strongloop/loopback-next/issues/2736

 - [x] Propose how to configure public models in such way that there is one JSON file for each model, we also need to consider how to access this file from transpiled output (`dist` vs `src`). Discuss with the wider @strongloop/loopback-next team and get an agreement. --> SEE https://github.com/strongloop/loopback-next/issues/2738 (spike)

 - [x] Implement TestSandbox.writeTextFile --> SEE #3731

 - [x] Improve `defineCrudRestController` to create a named controller class --> SEE #3732

[...]

Issue #2738, opened 2019-04-12
Spike: Booter for creating REST APIs from model files


bajtos:
_Part of From model definition to REST API with no custom repository/controller classes #2036_

In https://github.com/strongloop/loopback-next/issues/2736, we will implement a component providing CRUD REST Controller class that can be used to expose models via REST API.

In this spike, we will find out how to implement a booter that will read model configuration from JSON files, load controllers contributed by extensions, and connect those two parts together to expose REST APIs.

**Prerequisites**

(1)
An extension exporting a Controller, e.g. https://github.com/strongloop/loopback-next/issues/2736. We can build a minimal version of such extension as part of this spike if needed.
 
(2)
Configuration format. Let's start with the convention described in #2036. If it turns out to be difficult to implement, then propose a better one.
For example, to create a `Product` model, attach it to a `db` datasources and expose a CRUD REST API, all that the developer needs is to add the following two files to their project (plus any npm packages that implement this functionality):

`src/models/product.ts`:
```ts
@model()
export class Product extends Entity {
  @property()
  name: string;
}
```

`public-models/product.json`:
```json
{
  "model": "Product",
  "pattern": "Crud",
  "dataSource": "db",
}
```

Note: having config for multiple models in a single JSON file, as we do in LB 3.x, can quickly become a maintenance nightmare - see https://github.com/strongloop/loopback/issues/1316. It will be better to have one config file per model.

An example file for a model using KeyValue pattern (not CRUD) - `public-models/cache-entry.json`:

```json
{
  "model": "CacheEntry",
  "pattern": "KeyValue",
  "dataSource": "redis"
}
```
 
IMPORTANT: Keep in mind that TypeScript compiler does not copy arbitrary JSON files from `src` to `dist`. A possible solution is to keep model configuration files in `/public-models` directory (outside of `/src`).

## Acceptance criteria

- [ ] A design proposal and PoC implementation showing how extensions can contribute controllers that can be picked by a Booter. Preferably leverage ExtensionPoint/Extension pattern as shown by our greeter example.

- [ ] A PoC implementation of a booter parsing model configuration files, gathering controllers contributed by extensions, and using this information to build REST APIs for models as configured. Implementation-wise, start with the approach outlined in https://github.com/strongloop/loopback-next/issues/2036#issuecomment-462732901.

[...]

PR #3582, opened 2019-08-20
Extension: CrudRestController

bajtos:
This pull request adds `CrudRestController` extension as described in https://github.com/strongloop/loopback-next/issues/2736. See the top-level EPIC describing the big picture & plan: #2036 From model definition to REST API with no custom repository/controller classes.

Example use of the new controller:

1. Create a base REST CRUD controller class for your model.
	
   ```ts
   const CrudRestController = defineCrudRestController<
     Product,
     typeof Product.prototype.id,
     'id'
   >(Product, {basePath: '/products'});
   ```

2. Create a new subclass of the base controller class to configure repository
   injection.

   ```ts
   class ProductController extends CrudRestController {
     constructor(@repository(ProductRepository) repo: ProductRepository) {
       super(repo);
     }
   }
   ```

3. Register the controller with your application.

   ```ts
   app.controller(ProductController);
   ```

## Checklist


CODEOWNERS
```diff
@@ -26,6 +26,7 @@ packages/repository/* @raymondfeng
 packages/repository-tests/* @bajtos
 packages/repository-json-schema/* @bajtos
 packages/rest/* @bajtos @raymondfeng
+packages/rest-crud/* @bajtos @hacksparrow
```
bajtos: Discussion point: should this new package live in `packages/rest-crud` or `extensions/rest-crud`?

raymondfeng: My personal take: `packages/rest-crud`.

packages/rest-crud/README.md
```diff
@@ -0,0 +1,35 @@
+# @loopback/repository-tests
```
raymondfeng: `@loopback/rest-crud`

packages/rest-crud/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@loopback/rest-crud",
+  "version": "0.1.0",
```
raymondfeng: Let's use `0.0.1` as the starting version as `lerna version` will increment it to `0.1.0` with the 1st commit such as `feat(rest-crud): add rest controller for crud`.

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,94 @@
...
+  IdType,
+  IdName extends keyof T
+>(
+  modelCtor: typeof Entity & {prototype: T & {[key in IdName]: IdType}},
+): CrudRestControllerCtor<T, IdType, IdName> {
+  class CrudRestControllerImpl
+    implements CrudRestController<T, IdType, IdName> {
+    constructor(public readonly repository: EntityCrudRepository<T, IdType>) {}
+
+    @post('/', {
```
raymondfeng: Maybe expose `options` as an optional argument of `defineCrudRestController ` to customize paths and other things.

raymondfeng: Do we want to provide a `mixin` function to include CRUD controller methods?

bajtos: > Maybe expose `options` as an optional argument of `defineCrudRestController` to customize paths and other things.

Adding a new argument it's easy, let's leave that change for the future, if/when we actually need it.

> Do we want to provide a `mixin` function to include CRUD controller methods?

No. IMO, users should keep their controllers small. If they want to mix CRUD methods with some other APIs, then they should create two (or more) controllers.

bajtos: On the other thought, I see how `options` can be a better solution for customizing the `basePath` :+1:

bajtos:
Added docs and implemented the missing operations.

The pull request is ready for review. /cc @raymondfeng @strongloop/loopback-maintainers

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,320 @@
...
+ *
+ * @param modelCtor A model class, e.g. `Product`.
+ * @param options Configuration options, e.g. `{basePath: '/products'}`.
+ */
+export function defineCrudRestController<
+  T extends Entity,
+  IdType,
+  IdName extends keyof T,
+  Relations extends object = {}
+>(
```
raymondfeng: Is it feasible to define a `CrudRestController` from a `Repository` (instance or class) directly? Ideally, if a developer adds a repository using `lb4 repository` to connect a model to a data source, we should allows him/her to expose the repository as CRUD REST APIs out of box.

bajtos: That's an interesting idea, we can explore it later if you like. This new module is in 0.x version, so even if we need to make breaking changes to support your idea, then that's ok.

Our current plan is to allow developers to go from `lb4 model` to REST API with no custom repository class needed - see https://github.com/strongloop/loopback-next/issues/2036

The idea is that the user will tell us that let's say `Product` model should be exposed via CRUD REST API (as opposed to e.g. KeyValue REST API) using datasource `db`, and the framework will take care of the rest.

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/**
+ * Options to configure different aspects of a CRUD REST Controller.
+ */
+export interface CrudRestControllerOptions {
+  /**
+   * The base path where to "mount" the controller.
+   */
+  basePath: string;
```
nabdelgadir: Is it possible to make this optional and use a default `basePath` if one isn't provided?

bajtos: That would require us to implement some logic to convert a model name (e.g. `Product`) to a path (e.g. `/products`). This gets quickly tricky - do we want `/people` or `/persons` for `Person` model?

I'd prefer to keep `basePath` required for this initial implementation, we can make it optional later.

bajtos: Perhaps the purpose of this field is not clear enough and we should improve the api docs? Can you suggest a better wording? Would an example help?

nabdelgadir: > That would require us to implement some logic to convert a model name (e.g. `Product`) to a path (e.g. `/products`). This gets quickly tricky - do we want `/people` or `/persons` for `Person` model?

I was thinking we could use something similar to [how the controller generator does it](https://github.com/strongloop/loopback-next/blob/feat/rest-crud/packages/cli/generators/controller/index.js#L191) which in this scenario would make it `/people`.

> I'd prefer to keep `basePath` required for this initial implementation, we can make it optional later.

That's fine with me 👍 

> Perhaps the purpose of this field is not clear enough and we should improve the api docs? Can you suggest a better wording? Would an example help?

It's clear to me. My suggestion was so that we could make [`options`](https://github.com/strongloop/loopback-next/pull/3582/files#diff-fac7093db3933c6e14d7bd88c107299eR134) optional and be able to exclude it from the function call:
```ts
   const CrudRestController = defineCrudRestController<
     Product,
     typeof Product.prototype.id,
     'id'
   >(Product, {basePath: '/products'});
   ```
vs.
```ts
   const CrudRestController = defineCrudRestController<
     Product,
     typeof Product.prototype.id,
     'id'
   >(Product);
   ```

File: packages/rest-crud/README.md [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+
+```sh
+npm install --save @loopback/rest-crud
+```
+
+## Basic use
+
+1. Define your model class, e.g. using `lb4 model` tool.
+
+2. Create a Repository class, e.g. using `lb4 repository` tool.
```
raymondfeng: Is it required? IIUC, you want to skip Repository class so that a model can be used to expose REST CRUD APIs with a datasource.

packages/rest-crud/README.md
```diff
@@ -0,0 +1,65 @@
...
+     'id'
+   >(Product, {basePath: '/products'});
+   ```
+
+4. Create a new subclass of the base controller class to configure repository
+   injection.
+
+   ```ts
+   class ProductController extends CrudRestController {
+     constructor(@repository(ProductRepository) repo: ProductRepository) {
```
raymondfeng: If we skip the repository, can we inject a datasource instead?

packages/rest-crud/src/crud-rest.controller.ts
raymondfeng: The current implementation requires a repository class to be defined. I added a few comments inline. I'm open to incremental improvements.

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  @api({basePath: options.basePath, paths: {}})
+  class CrudRestControllerImpl
+    implements CrudRestController<T, IdType, IdName> {
+    constructor(
+      public readonly repository: EntityCrudRepository<T, IdType, Relations>,
+    ) {}
+
+    @post('/', {
+      ...response.model(200, `${modelName} instance created`, modelCtor),
```
jannyHou: shall we change `200` to `201`? I remember many people expect to have `201` as the "successful created" code

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,320 @@
...
+        // FIXME(bajtos) Improve repository API to support this use case
+        // with no explicit type-casts required
+        data as DataObject<T>,
+        where,
+      );
+    }
+
+    @patch('/{id}', {
+      responses: {
+        '204': {description: `${modelName} was updated`},
```
jannyHou: `204` should be `200`

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -0,0 +1,320 @@
...
+  const idProp = modelCtor.getIdProperties()[0];
+  const modelSchema = jsonToSchemaObject(
+    getJsonSchema(modelCtor),
+  ) as SchemaObject;
+  return (modelSchema.properties || {})[idProp] as SchemaObject;
+}
+
+// Temporary implementation of a short-hand version of `@requestBody`
+// See https://github.com/strongloop/loopback-next/issues/3493
+function body<T extends Entity>(
```
jannyHou: 👍

packages/rest-crud/src/crud-rest.controller.ts
bajtos: > I was thinking we could use something similar to [how the controller generator does it](https://github.com/strongloop/loopback-next/blob/feat/rest-crud/packages/cli/generators/controller/index.js#L191) which in this scenario would make it `/people`.

Makes sense! I had mixed experience with auto-generated base path for CRUD endpoints in LB3. Some people found the pluralization rules confusing ("people" vs. "persons"), some wanted different mechanism for converting pascal-cased model names (e.g. "ProductReview") to HTTP paths (e.g. "/ProductReviews", "/product-reviews", etc.).

Let's keep `basePath` optional for now and wait until we learn more about this area while working on https://github.com/strongloop/loopback-next/issues/2036.

packages/rest-crud/README.md
bajtos: A custom repository class is not required, this is just to make the documentation easier to follow.

This new component is meant to be a low-level building block to support https://github.com/strongloop/loopback-next/issues/2036, I am expecting that most people most people will use #2036 and don't leverage the current API directly.

For the initial docs, I choose a scenario that's easy to understand (because it's using existing concepts) and that's matching the scenario used for testing.

Depending on the outcome of the next tasks (especially the spike #2738), we may add more "meat" to this component and update or even rework the README along the way.

packages/rest-crud/README.md
bajtos: Yes, I believe injecting a datasource should be a viable option too. The actual implementation does not prescribe any IoC setup, the base controller class is expecting to receive a repository instance. It's up to the code extending the controller class to decide how to obtain the repository.

I believe the following solution should work as an alternative approach too:

```ts
class ProductController extends CrudRestController {
  constructor(
    @inject('datasources.db') dataSource: juggler.DataSource,
  ) {
    const repo = new DefaultCrudRepository<
      Product,
      typeof Product.prototype.id
    >(Product, db);

    super(repo);
  }
}
```

@raymondfeng Would you like me to open a new issue to add a test & documentation for this approach?

packages/rest-crud/src/crud-rest.controller.ts
bajtos: Eventually yes, but not yet. Responses with status code `201` are supposed to include `Location` header pointing to the newly created resource. We don't have a good solution allowing controller methods to specify a custom status code & response headers; plus it's non-trivial to calculate the correct `Location` - e.g. when the LB4 app is mounted on Express, or even worse when it's running behind a reverse proxy. See #2285, #2293 and #3133.

The scope of this task is to create REST API that's matching the API of controllers generated (scaffolded) via `lb4 controller`.

Adding support for 201 is out of scope, we already have a GH issue to track this feature - see https://github.com/strongloop/loopback-next/issues/788

packages/rest-crud/src/crud-rest.controller.ts
bajtos: `PATCH /my-model/{id}` does not return any response data, that's why the status code is 204. See also the controller template that uses 204 too.

https://github.com/strongloop/loopback-next/blob/554c763aa5c8db92bb4c3e5549960b8a99673791/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs#L116-L122

bajtos:
@raymondfeng @nabdelgadir @jannyHou thank you for the review and thoughtful comments. I am going to land this pull request as it is now. Please let me know if you would like me to improve any parts of this initial implementation as part of the story https://github.com/strongloop/loopback-next/issues/2736.

Issue #2794, opened 2019-04-24
Make datasource imports work with ts-node


bforbis:
Currently the `lb4 datasource` command will create a `{name}.datasource.ts` file and a `{name}.datasource.json` file. Since these files have the same base name, it causes issues with imports while running with `ts-node` (useful for development). By default, `ts-node` will import the JSON file over the TS file, causing datasource construction to fail. For more info on how `ts-node` works, see https://github.com/TypeStrong/ts-node/issues/769

I was able to workaround this issue by manually renaming the json file, but it would be good if the datasource generator didn't create two files with the same base name.

dougal83:
Hey, Please excuse my ignorance but why do you use ts-node?  Curious as to the necessity, are you doing something that specifically requires it?

bforbis:
It's not really necessary, but it makes certain things in development have a quicker feedback loop.

1. I can throw together quick ts scripts and test them with a single command
2. It allows me to skip some more complex steps of my normal build phase that involves moving static files around, as executing within ts-node will operate under the same directory structure within src/, instead of in dist/

dougal83:
> quicker feedback loop.

Ah like build times with the watch flag?  I just use ts-node for scripts like building .env files.  Fair dos, I think this'll be low priority until lb4 has feature parity with lb3.

bajtos:
@bforbis thank you for starting this discussion. I was looking into ts-jest and ts-node recently and discovered a similar problem: by default, LoopBack booters look for `.js` files only, and therefore don't discover `.ts` files. 

I think it would be great to improve LoopBack to support `ts-node`. Would you like to help?

For example, we can modify our CLI templates to use the following naming convention: `{name}.datasource-config.json`.

Thoughts?

/cc @strongloop/loopback-maintainers

AaronJan:
I'm just encountering this issue too.

I used `ts-node` for virtually all my projects, naturally I changed `nom run start:dev` to `cross-env NODE_ENV ts-node src/boot-server.ts`.

But when I added a component, one of my component class injection keep getting exception message: `Cannot resolve injected arguments for XXObserver.[0]: The arguments[0] is not decorated for dependency injection, but a value is not supplied`.

Here is my code:

```typescript
import { inject, lifeCycleObserver, LifeCycleObserver } from '@loopback/core';
import { LoggerToken } from '../../bindings'; // Seems any path that is outside `__dirname` will gets `undefined`.

@lifeCycleObserver('xx')
[...]

PR #3617, opened 2019-08-29
Spike: booter for creating REST APIs from model files

bajtos:
Quoting from https://github.com/strongloop/loopback-next/issues/2036:

> In LoopBack 3, it was very easy to get a fully-featured CRUD REST API with
> very little code: a model definition describing model properties + a model
> configuration specifying which datasource to use.
>
> Let's provide the same simplicity to LB4 users too.
>
> - User creates a model class and uses decorators to define model properties.
>   (No change here.)
> - User declaratively defines what kind of data-access patterns to provide
>   (CRUD, KeyValue, etc.) and what datasource to use under the hood.
> - `@loopback/boot` processes this configuration and registers appropriate
>   repositories & controllers with the app.

In this Spike, I am demonstrating a PoC implementation of an extensible booter
that processed model configuration files in JSON formats and uses 3rd-party
plugins to build repository & controller classes at runtime.

See [_SPIKE_.md](https://github.com/strongloop/loopback-next/blob/spike/crud-rest-booter/_SPIKE_.md) for further details.

## Checklist


packages/context/src/binding-decorator.ts
```diff
@@ -21,6 +21,25 @@ import {Constructor} from './value-promise';
  * Decorator factory for `@bind`
  */
 class BindDecoratorFactory extends ClassDecoratorFactory<BindingMetadata> {
+  mergeWithOwn(
```
bajtos: This change is needed to allow `bind` to be called multiple times on the same class, for example:

```ts
@booter('rest')
@extensionPoint(MODEL_API_BUILDER_PLUGINS)
export class RestBooter extends BaseArtifactBooter {
  // ...
}
```

packages/context/src/inject.ts
```diff
@@ -124,7 +124,7 @@ export function inject(
   }
   return function markParameterOrPropertyAsInjected(
     target: Object,
-    member: string,
+    member?: string,
```
bajtos: This change is needed to allow manual invocation of the decorator - I am surprised TypeScript did not catch that.

Example usage that does not work without my patch:

```ts
inject(`repositories.${entityClass.name}Repository`)(
    controllerClass,
    undefined,
    0,
  );
```

packages/context/src/binding-inspector.ts
```diff
@@ -272,34 +279,39 @@ function buildBindingKey(
...
   options: BindingFromClassOptions = {},
 ) {
-  const templateFn = bindingTemplateFor(cls);
-  // Create a temporary binding
-  const bindingTemplate = new Binding('template').apply(templateFn);
-  // Is there a `key` tag?
-  let key: string = options.key || bindingTemplate.tagMap[ContextTags.KEY];
-  if (key) return key;
+  try {
+    const templateFn = bindingTemplateFor(cls);
```
bajtos: I have wrapped most of the code into `try/catch` block, these changes are best viewed with "ignore whitespaces" setting on.

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -209,17 +180,12 @@ describe('TodoApplication', () => {
...
-      .bind('datasources.config.geocoder')
-      .to(getProxiedGeoCoderConfig(cachingProxy));
-
     // Start Application
     await app.start();
   }
 
   async function givenTodoRepository() {
-    todoRepo = await app.getRepository(TodoRepository);
+    todoRepo = await app.get<TodoRepository>('repositories.TodoRepository');
```
bajtos: We don't have a repository class we could use here (`TodoRepository` is just a type alias for `EntityCrudRepository<Todo, number>`.

examples/todo/src/__tests__/unit/controllers/todo.controller.unit.ts
```diff
@@ -1,161 +0,0 @@
...
-  sinon,
-  StubbedInstanceWithSinonAccessor,
-} from '@loopback/testlab';
-import {TodoController} from '../../../controllers';
-import {Todo} from '../../../models/index';
-import {TodoRepository} from '../../../repositories';
-import {GeocoderService} from '../../../services';
-import {aLocation, givenTodo} from '../../helpers';
-
-describe('TodoController', () => {
```
bajtos: I see little value in writing unit tests for controllers contributed by 3rd-party packages. The controller code should have been already tested in that package (`@loopback/rest-crud` in this case).

examples/todo/src/controllers/todo.controller.ts
```diff
@@ -1,138 +0,0 @@
...
-  patch,
-  post,
-  put,
-  requestBody,
-} from '@loopback/rest';
-import {Todo} from '../models';
-import {TodoRepository} from '../repositories';
-import {GeocoderService} from '../services';
-
-export class TodoController {
```
bajtos: No controller files, yay 🎉

examples/todo/src/repositories/todo.repository.ts
```diff
@@ -1,18 +0,0 @@
-// Copyright IBM Corp. 2018,2019. All Rights Reserved.
-// Node module: @loopback/example-todo
-// This file is licensed under the MIT License.
-// License text available at https://opensource.org/licenses/MIT
-
-import {inject} from '@loopback/core';
-import {DefaultCrudRepository, juggler} from '@loopback/repository';
-import {Todo, TodoRelations} from '../models';
-
-export class TodoRepository extends DefaultCrudRepository<
```
bajtos: No repository files, yay 🎉

examples/todo/src/models/todo.model.ts
```diff
@@ -29,17 +29,6 @@ export class Todo extends Entity {
...
-    type: 'string',
-  })
-  remindAtAddress?: string; // address,city,zipcode
-
-  // TODO(bajtos) Use LoopBack's GeoPoint type here
-  @property({
-    type: 'string',
-  })
-  remindAtGeo?: string; // latitude,longitude
-
```
bajtos: In order to implement geocoder integration while using `public-models/product.config.json`, we need to customize the controller method defined by `@loopback/rest-crud`. That's kind of against the spirit of this change, where we want to use the default REST API as provided by LB. At least for the initial release, I'd like us to recommend users to switch from `@loopback/rest-crud` to `lb4 repository` & `lb4 controller` if they need any tweaks in the built-in functionality.

packages/booter-rest/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@loopback/booter-rest",
```
bajtos: If we keep this extension, then I'd like us to rename it to `@loopback/model-api-booter`. I believe the implementation can support any API style, not only REST.

Having said that, the booter is very simple and pretty un-opinionated, I think we should add it to core `@loopback/boot` package. All opinionated parts are handled by model-api-builders like `@loopback/rest-crud`.

bajtos:
The spike is done and ready for review.

packages/context/src/binding-decorator.ts
bajtos: Fixed by https://github.com/strongloop/loopback-next/pull/3618

packages/context/src/inject.ts
raymondfeng: IIRC, we can use `''` as `member` for constructor parameters.

packages/metadata/src/decorator-factory.ts
```diff
@@ -368,8 +368,9 @@ export class ClassDecoratorFactory<T> extends DecoratorFactory<
     descriptorOrIndex?: TypedPropertyDescriptor<any> | number,
   ) {
     if (ownMetadata != null) {
+      const decoratorName = this.constructor.name.replace(/Factory$/, '');
       throw new Error(
-        'Decorator cannot be applied more than once on ' +
+        `${decoratorName} cannot be applied more than once on ` +
```
raymondfeng: See https://github.com/strongloop/loopback-next/pull/3625

packages/context/src/binding-inspector.ts
raymondfeng: Maybe this logic should be moved to `createBindingFromClass()`. 

Is there a better way than mutating the error message?

bajtos: > Maybe this logic should be moved to `createBindingFromClass()`.

That's possible, my solution here is a just a quick hack to give me the information I need for troubleshooting.

> Is there a better way than mutating the error message?

I am not aware of any. It's important to preserve the original stack trace and AFAIK, that requires us to mutate the original error object.

packages/context/src/inject.ts
bajtos: I am reluctant to set `member= ''` because that's different from what TypeScript does.

Here is the code emitted by `tsc` for a `TodoRepository`:

```js
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
  // [removed for brevity]
};

// [removed for brevity]

let TodoRepository = class TodoRepository 
  extends repository_1.DefaultCrudRepository 
{
  constructor(dataSource) {
    super(models_1.Todo, dataSource);
  }
};
TodoRepository = __decorate([
  __param(0, core_1.inject('datasources.db')),
  __metadata("design:paramtypes", [repository_1.juggler.DataSource])
], TodoRepository);
```

Notice that `__decorate` is called with `key` set to `undefined`.

bajtos: See https://github.com/strongloop/loopback-next/pull/3634

bajtos: See https://github.com/strongloop/loopback-next/pull/3634

_SPIKE_.md
```diff
@@ -0,0 +1,163 @@
...
+
+  - Do we want to introduce `@model()` decorator for configuring dependency
+    injection? (Similar to `@repository`.)
+  - Do we want to rework scaffolded repositories to receive the model class via
+    DI?
+
+- Implement model booter to scan `dist/models/**/*.model.js` files and register
+  them by calling `app.model`.
+
+- Implement `sandbox.writeJsonFile` in `@loopback/testlab`.
```
bajtos: See #3635

_SPIKE_.md
```diff
@@ -0,0 +1,163 @@
...
+`ProductController.replaceById` with `CrudRestControllerImpl.replaceById` -
+which one is more useful?)
+
+In my proposal, model-config files are in JSON format to make programmatic edits
+easier. This has a downside in TypeScript projects - these config files must
+live outside `src` because TypeScript does not copy arbitrary JSON files.
+
+### Extensibility & customization options
+
+The proposed design enables the following opportunities to extend and customize
```
jannyHou: Could you elaborate more about how to customize the controller functions? Like applying interceptors and authentication/authorization? By creating plugins?

Discussed with @emonddr the other day he also has similar question.

_SPIKE_.md
```diff
@@ -0,0 +1,163 @@
...
+I have updated `examples/todo` application to leverage this new functionality,
+you can look around the updated code to see what the user experience will look
+like. Take a look at how acceptance level tests are accessing the repository
+created by `rest-crud` plugin.
+
+The booter is implemented in
+[`packages/booter-rest/src/rest.booter.ts`](https://github.com/strongloop/loopback-next/blob/spike/crud-rest-booter/packages/booter-rest/src/rest.booter.ts).
+Initially, I created a new package for this booter, but on the second thought, I
+think this booter is generic enough to be included in `@loopback/boot` under a
+different name `ModelApiBooter`.
```
jannyHou: > Initially, I created a new package for this booter, but on the second thought, I
think this booter is generic enough to be included in `@loopback/boot` under a
different name `ModelApiBooter`.

+1 Or merge it with the `@loopback/model-api-builder` and export everything as a component.

packages/context/src/inject.ts
bajtos: And also https://github.com/microsoft/TypeScript/issues/33260

_SPIKE_.md
bajtos: > Could you elaborate more about how to customize the controller functions? Like applying interceptors and authentication/authorization? By creating plugins?

Sure.

First of all, if you want to customize controller functions, then your project has most likely outgrown the simplicity offered by `@loopback/rest-crud` and it's time to scaffold a controller class using `lb4 controller`.

Having said that, if you want to apply the same customization to multiple models/controllers, for example if you want to apply the same authentication/authorization rules, then you can:
1. Fork `@loopback/rest` into your own package, or perhaps just copy the few relevant files into your project.
2. Modify the controller class as you like.
3. You can also change the way how repositories are created (if needed).
4. Modify the ModelApiBuilder copied from `@loobpack/rest` - give it a different pattern name (not `CrudRest`).
5. Bind your modified `ModelApiBuilder` to your app, so that the booter can find it.
6. In your model-config files, replace the `pattern` value from `CrudRest` to the new builder name you choose in the step 4.

_SPIKE_.md
bajtos: I don't want to merge it with `@loopback/model-api-builder` because I don't want `@loopback/model-api-builder` to be coupled with `@loopback/boot`.

`@loopback/model-api-builder` is going to be used a dependency of all builder implementations, I want to keep it as lightweight as possible.

_SPIKE_.md
```diff
@@ -0,0 +1,163 @@
...
+better support this spike.
+
+## Open questions:
+
+- Where to keep model config files?
+
+  - `/public-models/product.config.json` (JSON, must be outside src)
+  - `/src/public-models/product-config.ts` (TS, can be inside src, more
+    flexible)
+
```
raymondfeng: We can follow the same convention as `datasources`:

- src/resources/my-model.resource.json
- src/resources/my-model.resource.ts

We can discuss what's the best name for `resources`.

_SPIKE_.md
jannyHou: > because I don't want @loopback/model-api-builder to be coupled with @loopback/boot.

I see fair enough 👍

_SPIKE_.md
bajtos: The approach we are using for datasources is actually problematic, see https://github.com/strongloop/loopback-next/issues/2794

> Currently the `lb4 datasource` command will create a `{name}.datasource.ts` file and a `{name}.datasource.json` file. Since these files have the same base name, it causes issues with imports while running with `ts-node` (useful for development). By default, `ts-node` will import the JSON file over the TS file, causing datasource construction to fail. For more info on how `ts-node` works, see https://github.com/TypeStrong/ts-node/issues/769
>
> I was able to workaround this issue by manually renaming the json file, but it would be good if the datasource generator didn't create two files with the same base name.

Also datasources are different from model configs. A datasource file exports a datasource class, therefore we need a TS file. The configuration json file is kind of optional for datasources.

Model configs are exporting a config object only, therefore we have the option to decide whether to use JSON or TS.

As I was thinking about this more, I am leaning towards using TS files.
- It's more consistent with the approach we have for other artifacts, most notably model definitions.
- It gives a lot of flexibility to application developers, they can e.g. parse env variables to decide certain aspects of model configuration.

packages/context/src/binding-inspector.ts
bajtos: See https://github.com/strongloop/loopback-next/pull/3676

_SPIKE_.md
bajtos: Unfortunately, the convention used by datasources does not work well.

Quoting from https://github.com/strongloop/loopback-next/issues/2794:
> Currently the `lb4 datasource` command will create a `{name}.datasource.ts` file and a `{name}.datasource.json` file. Since these files have the same base name, it causes issues with imports while running with `ts-node` (useful for development). By default, `ts-node` will import the JSON file over the TS file, causing datasource construction to fail.

In my view, datasources are different from model-config files.
- For datasources, we primarily need the `.ts` file that is exporting the datasource class. The `.json` file with the configuration is just an optional support.
- For model config files, there is no class to export, the booter needs a data object with configuration options.

As I was thinking about this more, I am leaning towards using TS files in `src` folder. I feel this is more consistent with the approach we use for all other artifacts (models, repositories, etc.). It also enables application developers to conditionally customize model config, e.g. depending on `process.env` variables.

I am going to update this part of the spike.

bajtos:
I have updated the spike as follows:
- Rebased on top of the latest master
- Reworked the implementation to load model configs from JS files in `dist/public-models/*.config.js` instead of `public-models/*.config.json`
- Moved `RestBooter` to `@loopback/boot` and renamed it to `ModelApiBooter`
- Improved SPIKE docs based on review comments, updated list of implementation stories

I consider this spike as done and ready for final review & approval.

@strongloop/loopback-next PTAL.

_SPIKE_.md
bajtos: I updated `_SPIKE_.md` to include the information from my comment above.

jannyHou: Thank you 👍 very detailed.

sanadHaj:
@bajtos Great effort, 👍
Honestly if feel loopback should also boot controllers repository from open api files 
Is there any consideration of that ?

bajtos:
> Honestly if feel loopback should also boot controllers repository from open api files

@sanadHaj Possibly, but not as part of this initiative. Can you open a new issue please to discuss the details of your use case?

_SPIKE_.md
```diff
@@ -0,0 +1,244 @@
...
+> - `@loopback/boot` processes this configuration and registers appropriate
+>   repositories & controllers with the app.
+
+In this Spike, I am demonstrating a PoC implementation of an extensible booter
+that processed model configuration files in JSON formats and uses 3rd-party
+plugins to build repository & controller classes at runtime.
+
+## Basic use
+
+Create `src/public-models` directory in your project. For each model you want to
```
raymondfeng: I still think the name `public-models` is misleading as `public` is usually a visibility modifier. Maybe something like `model-endpoints` or `model-resources` is better.

_SPIKE_.md
```diff
@@ -0,0 +1,244 @@
...
+expose via REST API, add a new `.config.ts` file that's exporting the model
+configuration.
+
+Example (`src/public-models/product.config.ts`):
```
raymondfeng: We can use `product.resource.ts` or `product.endpoint.ts` once we decide the proper name.

_SPIKE_.md
raymondfeng: Or `rest-models`

_SPIKE_.md
```diff
@@ -0,0 +1,244 @@
...
+I feel it's important to use model-named classes (e.g. `ProductController`)
+instead of reusing the same controller/repository class name for all models.
+Model-named classes make error troubleshooting and debugging easier, because
+stack traces include model name in function names. (Compare
+`ProductController.replaceById` with `CrudRestControllerImpl.replaceById` -
+which one is more useful?)
+
+In my proposal, model-config files are in JSON format to make programmatic edits
+easier. This has a downside in TypeScript projects - these config files must
+live outside `src` because TypeScript does not copy arbitrary JSON files.
```
raymondfeng: IIRC, we can configure the tsconfig to allow json files inside `src`.

_SPIKE_.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+## Implementation
+
+The solution has the following high-level parts:
+
+1. A new package `@loopback/model-api-builder` defines the contract for plugins
+   (extensions) contributing repository & controller builders.
+
+2. A new booter `RestBooter` that loads all JSON files from
```
raymondfeng: RestBooter -> ModelApiBooter?

packages/boot/src/booters/model-api.booter.ts
```diff
@@ -0,0 +1,104 @@
...
+    // TODO assert that `app` has RepositoryMixin members
+
+    super(
+      projectRoot,
+      // Set booter options if passed in via bootConfig
+      Object.assign({}, RestDefaults, booterConfig),
+    );
+  }
+
+  async load(): Promise<void> {
```
raymondfeng: How do we make sure `model` classes are bound to the application context first during boot?

packages/model-api-builder/src/model-api-builder.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Extension Point name for Model API builders.
+ */
+export const MODEL_API_BUILDER_PLUGINS = 'model-api-builders';
+
+/**
+ * Interface for extensions contributing custom API flavors.
+ */
+export interface ModelApiBuilder {
+  readonly pattern: string; // e.g. CrudRest
+  setup(
```
raymondfeng: `setup` -> `build`?

raymondfeng:
@bajtos Great work! 🥇 I love the general direction to allow pluggable builders to create/derive/compose new artifacts from existing ones. 

See my comments inline.

_SPIKE_.md
bajtos: > I still think the name `public-models` is misleading as `public` is usually a visibility modifier.

I am not entirely sold on `public-models` either, but I am not sure if we can come up with a better alternative.

In LB3, we have a single `model-config.json` file where each model has a field called `public` that controls whether the model is exposed via (REST) API or not. That's how I come with `public-models` name. I agree that `public` is usually a visibility modifier - I am interpreting the visibility as whether the model is visible in the server API nor not.

One important aspect to keep in mind is that these config files are not specific to REST APIs, we can use the same files to expose models via gRPC or GraphQL as long as they use a repository & a controller to build the API. I'd like to use a generic directory name, not `rest` specific.

From you suggestions, @raymondfeng, I like most `model-endpoints`.

Few more directory names to consider, examples are for a hypothetical OrderItem model:

- `model-api-config` - e.g. `src/model-api-config/order-item.rest-config.ts`
- `api-config` - e.g. `src/api-config/order-item.rest-config.ts`
- `model-api` (or `model-apis`?) -  e.g. `src/model-api/order-item.rest-config.ts`
- `server-api` - e.g. `src/server-api/order-item.rest-config.ts`

@strongloop/loopback-next @strongloop/loopback-maintainers What do you think? Can you please help us choose the best name, one that will make the structure of LB4 projects easy to understand?

_SPIKE_.md
bajtos: +1

Personally, I'd use `product.rest-config.ts` and if we add support for gRPC controllers later, then `product.grpc-config.ts`.

_SPIKE_.md
bajtos: Sure. However, AFAIK, vanilla `tsc` copies to `dist` only those JSON files that are directly `import`ed from a `.ts` file. Model config files are not imported, they are processed at runtime.

Anyhow, I'd like to start with `.ts` files for the initial version. We can add support for other flavors like `JSON` files later, I believe such improvement should be a small & backwards compatible change.

packages/boot/src/booters/model-api.booter.ts
bajtos: > How do we make sure `model` classes are bound to the application context first during boot? Do we anticipate that model classes are bound during `configure` phase? Do we have to introduce a new `build` phase for booters?

Good catch!

Binding model classes during `configure` phase seems hacky to me. I was expecting that the model booter will be implemented the same way as other artifact booters, e.g. [RepositoryBooter](https://github.com/strongloop/loopback-next/blob/cc5cd20ae11e58c07a336774474ff6d7fe381a06/packages/boot/src/booters/repository.booter.ts).

Here are few options to consider:

(1)
Can we introduce a concept of dependencies to booters, allowing `ModelApiBooter` to request that `ModelBooter` is executed first?

```ts
@booter('modelApi', {dependsOn: [ModelBooter]})
@extensionPoint(MODEL_API_BUILDER_PLUGINS)
export class ModelApiBooter extends BaseArtifactBooter {
  // ...
}
```

(2)
Instead of introducing a Model booter, perhaps we can modify `ModelApiBooter` to load model classes, effectively merging Model + API booters into a single class.

- In the configuration, `ModelApiBooter` will require two sets of `ArtifactOptions` - one for model classes, another for configuration files.
- At discovery, the booter will discover both model and config files.
- When loading, we don't even have to bind model classes to context, API booter can look-up model classes from the array provided by `BaseArtifactBooter`.

(3)
Abandon the idea of booting model classes. Require the api-config object to provide a class constructor.

```ts
import {ModelApiConfig} from '@loopback/rest-crud';
import {Todo} from '../models';

module.exports = <ModelApiConfig>{
  model: Todo,
  pattern: 'CrudRest',
  dataSource: 'db',
  basePath: '/todos',
};
```

This may simplify consumption of models from 3rd-party packages (npm modules) - instead of having to tweak booter configuration, developers can decide in each config file where to load the model from.

```ts
import {Todo} from '@acme/process-models';
import {ModelApiConfig} from '@loopback/rest-crud';

module.exports = <ModelApiConfig>{
  model: Todo,
  pattern: 'CrudRest',
  dataSource: 'db',
  basePath: '/todos',
};
```

---

Personally, I am leaning towards the last option, as it's the easier to implement. We can look into other options later, after the v1 implementation has been shipped.

Thoughts?

/cc @strongloop/loopback-next @strongloop/loopback-maintainers

bajtos:
@raymondfeng thank you for a great feedback! Let's keep the discussion going in the inline comments.

bajtos:
Over the weekend, I realized that if we want to keep `ModelApiBuilder` generic and support more API flavors besides REST, then `basePath` configuration does not belong the the generic config interface. In 4f74b98f, I introduced a new config interface `CrudRestApiConfig` (inheriting from `ModelApiConfig`) and moved `basePath` there.

I am not sure if `CrudRestApiConfig` is the best name (it does not mention "model"), suggestions are welcome.

_SPIKE_.md
agnes512: I prefer `model-endpoints` as the directory name for the models.
As for config files, I think names with `config` makes more sense to me. I am ok with `model-api-config` or `api-config`.

hacksparrow: @bajtos How about `src/models/configs/product.rest-config.ts`? Models related stuff within the  `src/models/` dir seem more intuitive.

bajtos: @hacksparrow 
> How about `src/models/configs/product.rest-config.ts`? Model related stuff within the  `src/models/` dir seem more intuitive.

I see your point. I would choose `src/models/endpoints/product.rest-config.ts` in that case.

However, what if the model is coming from a 3rd-party package, e.g. `import {User} from 'lb4-local-auth'`? Is it ok to have a file in `src/models/endpoints` but no corresponding file in `src/models`?

hacksparrow: I like `src/models/endpoints/product.rest-config.ts` too.

> Is it ok to have a file in src/models/endpoints but no corresponding file in src/models?

It is ok, since the source of `User` is explicitly stated. Also, this question is applicable even if the config files in kept in `src/public-models`, since there will be no corresponding model file for `User` unlike others.

Coming up with a good name for the config directory is not easy. My thoughts on looking the considered dir names:

- `public-models` - are there private models too?
- `model-resources` - what exactly are these resources? Artifacts to help create models? Commonly shared artifacts?
- `rest-models` - will `graphql-models` be supported too?
- `model-endpoints` - aren't model endpoints generated by the framework using the controllers?
- `model-api-config` - this comes quite close to describing what it actually does. But why is it a separate dir? It should probably go inside the `models` dir. I like the name `endpoints` as suggested by @bajtos.
- `api-config` - this is the LoopBack config dir?
- `model-api` - isn't the model API defined by the model files?
- `server-api` - is this where you define the server's API endpoints? Then what is `controllers` for?

bajtos:
Thank you all for feedback. I decided to address it the following way:

- I have renamed `public-models` to `model-endpoints` as suggested in the comments.
- I am using the naming convention `{model-name}.rest-config.ts`, this will allow us to introduce e.g. `{model-name}.grpc-config.ts` in the future.
- The API config is referencing the model class directly now, setting `model` field to an imported class constructor.
- I have removed `app.model()` API that's no longer needed.
- I have update the SPIKE document accordingly.

Let's do a final round of reviews and then call the spike done.

_SPIKE_.md
bajtos: Thank you @hacksparrow for a detailed comment on different directory names. I feel the distinction between `models/endpoints` and `model-endpoints` is small enough to ignore it, considering that both Raymond and Agnes are fine with `model-endpoints`.

packages/boot/src/booters/model-api.booter.ts
bajtos: I have updated the spike with the 3rd option, The ModelApiConfig object is referencing the model class directly now, setting model field to an imported class constructor.

_SPIKE_.md
jannyHou: > model-endpoints

+1

bajtos:
Follow-up stories:
- Implement TestSandbox.writeTextFile #3731
- Improve `defineCrudRestController` to create a named controller class #3732
- Add `defineCrudRepositoryClass` - a helper to create a named repository class #3733
- Model API booter & builder #3736
- Add CrudRestApiBuilder to `@loopback/rest-crud` #3737
- Example app showing CrudRestApiBuilder #3738

I am closing the spike as done.

PR #3622, opened 2019-08-30
fix(repository-json-schema): guards against undefined in schema builder

treffynnon:
Guards the includeReferencedSchemas function against assigning to
undefined.

This appears to happen when a model refers to itself and the delete
keyword wipes out the objects property.

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -1082,5 +1082,27 @@ describe('build-schema', () => {
         expect(optionalNameSchema.title).to.equal('ProductPartial');
       });
     });
+
+    it('can generate a schema involving recursive entities', () => {
+      @model()
+      class Category extends Entity {
+        @property({id: true})
+        id: number;
+
+        @hasMany(() => Category)
+        subCategories?: Category[];
+      }
+      const schema = getJsonSchema(Category, {includeRelations: true});
+      expect(schema.properties).to.deepEqual({
+        id: {type: 'number'},
+        subCategories: {
+          type: 'array',
+          items: {
+            $ref: '#/definitions/CategoryWithRelations',
+          },
+        },
+      });
+      expect(schema.title).to.equal('CategoryWithRelations');
+    });
   });
 });
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -405,7 +405,9 @@ export function modelToJsonSchema<T extends object>(
       delete schema.definitions;
     }
 
-    result.definitions[name] = schema;
+    if (result.definitions) {
+      result.definitions[name] = schema;
+    }
   }
   return result;
 }


raymondfeng:
@treffynnon Thank you for the contribution. Would you please add a test case to cover the issue you have experienced without the proposed change?

treffynnon:
@raymondfeng I have added an integration test for this problem now. Essentially, recursive or self-referential entities cause a type error to be emitted from the code because `result.definitions` ends up being `undefined`. The code then attempts to assign a value to `result.definitions`, which throws an error.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -1082,5 +1082,20 @@ describe('build-schema', () => {
...
+      class Category extends Entity {
+        @property({id: true})
+        id: number;
+
+        @hasMany(() => Category)
+        subCategories?: Category[];
+      }
+
+      expect(getJsonSchema(Category, {includeRelations: true})).not.throwError(
+        "Cannot set property 'CategoryWithRelations' of undefined",
```
bajtos: Let's improve this test to check for the desired outcome (how should the JSON schema look like) instead of merely checking that a specific error was not thrown.

```ts
const schema = await getJsonSchema(Category, {includeRelations: true});
expect(schema).to.deepEqual({
  // what should the schema look like?
});
```

Maybe we don't need to assert on the entire schema object, just check how `subCategories` property is defined and how does it reference the root schema.

bajtos:
Thank you @treffynnon for the pull request. I am not sure if your change is enough to fix the problem - how does the JSON schema for your test model looks like? Let's make sure we fix the problem properly and emit a valid schema.

treffynnon:
## Introduction

@bajtos and @raymondfeng this PR fixes the issue that I am having with recursive entities in my system and appears to allow a valid schema to be built.

## A little background

I am using `@loopback/openapi-v3`, but from there on it is a custom framework - I am not using loopback's repository handling, routing or controllers. I just use it to generate an OAS document from which my framework builds its own routes out. This is to allow for multiple communication "protocols" - REST, GraphQL, gRPC, etc all via the same route handling.

## Entities

Entities in my system can refer to each other, themselves or circuitously via any number of entities (for example `A --> B --> C --> A`).

This is not my actual schema, but it is a simplified example of the same problem in a different domain.

```
+--------------+            +--------------+
|              |            |              |
|  Customers   |            |  Categories  +---+
|              |            |              |   |
+------+-------+            +------+----^--+   | Sub-categories
       |                           |    |      | Parent category
       |                           |    +------+
       |                           |
       |                           |
+------v-------+            +------v-------+
|              |            |              |
|    Orders    +------------>   Products   +---+
|              |            |              |   |
+--------------+            +-----------^--+   | Accessories
                                        |      | Related products
                                        +------+ etc
```

What we're focussed on here is the self-referential relationships of `Categories` and `Products`. Without this change the schema will not build as the code attempts to assign a value to something that is `undefined` - it should be generating a reference to itself instead. Once this change is merged it will do just that as the new test clarifies.

The relationship between `Customers` and `Orders` however currently works and continues to work after this change as well.

## A note on openapi-v3 package

To make this all work for me I've had to clone the `controller-spec.ts` file from the openapi-v3 package into my project such that I can change the call in `generateOpenAPISchema()` to `getJsonSchema()` to use the `includeRelations` option. Without this entities like `Categories` and `Products` do not come through in the schema that is built.

## Conclusion

I hope that this makes the context of this change clearer and that the new test covers enough of the change to make this mergeable.

treffynnon:
@bajtos Is there anything further you need before merging this PR?

treffynnon:
@raymondfeng so... you fixed this already or no? https://github.com/strongloop/loopback-next/pull/3674?

What is happening? Should I close this PR? Is it still useful?

raymondfeng:
@treffynnon I just realized the issue has been fixed in https://github.com/strongloop/loopback-next/pull/3674.

raymondfeng:
@treffynnon Thank you for the patch!

treffynnon:
@raymondfeng and @bajtos ummmm......

> @treffynnon I just realized the issue has been fixed in #3674.

What? You realised? Really? It wasn't that I dug through merged PRs and pointed you right at it?

Perhaps this would make for a more accurate response:

> Thank you for bringing that duplication to my attention - you're right it has been fixed by another PR that I prepared whilst your PR was left pending.
>
> Thank you for all the additional effort you put into describing the problem and producing meaningful tests.
>
> On this occasion though we are unable to merge your pull request as I decided to go in another direction.

I have got to say that I've found the contribution process here rather slow, time consuming and ultimately completely exasperating.

PR #3623, opened 2019-08-30
feat(booter-lb3app): process generated OpenApiSpec of lb3 app

derdeka:
While migration from lb3 to lb4 I had several issues with the converted openapi spec of my lb3app, e.g.:
  - incomplete return types which results in unuseable angular model interfaces
  - renaming of models in the spec to avoid name collisions between lb3 and lb4 implementations

We solved this by post processing the generated openapi spec before it gets published into lb4.
If you think this is the way to go, we can add some tests.

## Checklist

### DIFF
--- a/packages/booter-lb3app/src/lb3app.booter.ts
+++ b/packages/booter-lb3app/src/lb3app.booter.ts
@@ -101,7 +101,11 @@ export class Lb3AppBooter implements Booter {
       // swagger2openapi options
     });
 
-    return result.openapi as OpenApiSpec;
+    let spec = result.openapi as OpenApiSpec;
+    if (typeof this.options.specTransformer === 'function') {
+      spec = this.options.specTransformer(spec);
+    }
+    return spec;
   }
 
   private mountFullApp(lb3App: Lb3Application, spec: OpenApiSpec) {
@@ -128,6 +132,7 @@ export interface Lb3AppBooterOptions {
   path: string;
   mode: 'fullApp' | 'restRouter';
   restApiRoot: string;
+  specTransformer?: (spec: OpenApiSpec) => OpenApiSpec;
 }
 
 interface Lb3Application extends ExpressApplication {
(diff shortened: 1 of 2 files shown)


raymondfeng:
@derdeka Thank you for the patch!

> incomplete return types which results in unusable angular model interfaces
> renaming of models in the spec to avoid name collisions between lb3 and lb4 implementations

Do these fixes apply to most use cases? If so, we should fix them in the booter implementation itself.

I'm open to the idea to allow `postProcessOpenApiSpec` option.

Please add some tests.

bajtos:
Thank you @derdeka for the pull request.

> incomplete return types which results in unuseable angular model interfaces

I agree with what Raymond wrote above - it would be best to fix the code that's generating OpenAPI spec, so that no further post-processing is needed. It makes me wonder - are you using Swagger with your LB3 application? Is that swagger spec correct? 
- If yes, then the problem could be in the conversion from Swagger to OpenAPI - see [swagger2openapi](https://www.npmjs.com/package/swagger2openapi).
- If not, then we should fix [loopback-swagger](https://github.com/strongloop/loopback-swagger) to emit correct Swagger spec in the first place!

> renaming of models in the spec to avoid name collisions between lb3 and lb4 implementations

I am not sure how easy would be to support this use case at loopback-swagger and/or the booter level. Your proposal to allow applications to contribute a custom post-processing function (handler) looks like a great solution :+1:

Can we find a shorter name for `processOpenApiSpec` option? How about `specTransformer` or something similar?

derdeka:
I've renamed and added a simple test.

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -128,6 +132,7 @@ export interface Lb3AppBooterOptions {
   path: string;
   mode: 'fullApp' | 'restRouter';
   restApiRoot: string;
+  specTransformer?: (spec: OpenApiSpec) => OpenApiSpec;
```
raymondfeng: I'm good to have `specTransformer` to be synchronous function as a starting point. We should consider generalizing the idea:

1. Allow asynchronous transformation
2. Create an extension point for spec transformers/processors to transform OpenAPI specs generated from controller metadata.

bajtos: Please note this transformer is applied only for LB3 applications imported to LB4.

PR #3634, opened 2019-09-02
fix(context): allow `inject` to be explicitly invoked for class ctor args

bajtos:
See the discussion in https://github.com/strongloop/loopback-next/pull/3617

## Checklist

### DIFF
--- a/packages/context/src/inject.ts
+++ b/packages/context/src/inject.ts
@@ -124,7 +124,7 @@ export function inject(
   }
   return function markParameterOrPropertyAsInjected(
     target: Object,
-    member: string,
+    member: string | undefined,
     methodDescriptorOrParameterIndex?:
       | TypedPropertyDescriptor<BoundValue>
       | number,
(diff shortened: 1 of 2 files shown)


packages/context/src/inject.ts
```diff
@@ -124,7 +124,7 @@ export function inject(
   }
   return function markParameterOrPropertyAsInjected(
     target: Object,
-    member: string,
+    member: string | undefined,
```
raymondfeng: The `ParameterDecorator` type from TypeScript is declared as follows:
```ts
declare type ParameterDecorator = (target: Object, propertyKey: string | symbol, parameterIndex: number) => void;
```

Can we use `''` for constructor parameters?

raymondfeng: Interestingly, at runtime, the `propertyKey` is `undefined` in constructor cases.

bajtos: > Interestingly, at runtime, the `propertyKey` is `undefined` in constructor cases.

That's the weird thing!

bajtos:
> Maybe we should report the consistency to TypeScript.

I was considering that, but was lazy to fill the report 🙈 

Here we go: https://github.com/microsoft/TypeScript/issues/33260

PR #3643, opened 2019-09-03
fix(cli): add 'generated: true' entry to id property when generating

agnes512:
connect to https://github.com/strongloop/loopback-next/pull/3538

We found in [refactor test PR](https://github.com/strongloop/loopback-next/pull/3538#discussion_r317624487) that id needs to be generated for any instances for the mysql tests, otherwise it complains with Error: `ER_NO_DEFAULT_FOR_FIELD: Field 'id' doesn't have a default value`.
It's not required for relations. But it might be problematic if models are migrated to databases such `MySQL`.
## Checklist

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -411,6 +411,13 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
           default: false,
         },
         {
+          name: 'generated',
+          message: `Is ${chalk.yellow(this.propName)} generated automatically?`,
+          type: 'confirm',
+          default: true,
+          when: answers => answers.id,
+        },
+        {
           name: 'default',
           message: `Default value ${chalk.yellow('[leave blank for none]')}:`,
           when: answers => {
@@ -431,6 +438,8 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       Object.assign(this.artifactInfo.properties[this.propName], answers);
 
       // We prompt for `id` only once per model using idFieldSet flag.
+      // and 'generated' flag makes sure id is defined, especially for database like MySQL
+      // Skipped the test for `generated` for now.
       if (answers.id) {
         this.idFieldSet = true;
       }


agnes512:
> Is there a (reasonably easy) way how to write a test for this new execution path?

The test for this functionality only makes sense for MySQL since in-memory and mongo generate ids automatically without  `generated: true`. 
It can only be tested in the `repository-test` folder. I don't think it makes sense to add a generator test in that folder though. Any suggestions?

packages/cli/generators/model/index.js
```diff
@@ -420,6 +420,11 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
       ];
 
       const answers = await this.prompt(prompts);
+
+      if (answers.id) {
+        // 'generated' makes sure id is defined, especially for database like MySQL
+        answers['generated'] = true;
```
raymondfeng: Should we have a prompt to ask if the `id is automatically generated by the database`?

bajtos:
> >  Is there a (reasonably easy) way how to write a test for this new execution path?
>
> The test for this functionality only makes sense for MySQL since in-memory and mongo generate ids automatically without  `generated: true`. 
> It can only be tested in the `repository-test` folder. I don't think it makes sense to add a generator test in that folder though. Any suggestions?

I agree with you that we don't want to run `lb4` from `repository-test` tests, that would make them super slow!

I was thinking about a test that will call `lb4 model` and then check the generated model `.ts` file to verify that `generated: true` is included in id's `@property` metadata.

packages/cli/generators/model/index.js
bajtos: +1 to add a prompt. The topic is more complex though because of the model-level setting `forceId`.

Let's land this PR as an incremental improvement and open a new issue to investigate what changes are needed to support models where the PK is provided by clients (not generated by the DB).

packages/cli/test/integration/generators/model.integration.js
```diff
@@ -205,6 +205,51 @@ describe('lb4 model integration', () => {
       assert.fileContent(expectedModelFile, /\[prop: string\]: any;/);
     });
 
+    it('scaffolds model with generated flag set to true for id property', async () => {
+      const property = {
```
agnes512: I tried to test if the `generated` flag is added to the id property. But it doesn't generate the id property at all. Is there any way to test it out? Or we can skip the test for now? @raymondfeng  @bajtos

bajtos: Weird. I am ok to skip the test for now, unless @raymondfeng objects.

raymondfeng: As we don't have a good way to mock prompts for composed generators, I'm fine to skip the test for now.

packages/cli/generators/model/index.js
```diff
@@ -410,6 +410,13 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
           type: 'confirm',
           default: false,
         },
+        {
+          name: 'generated',
+          message: 'Is id generated automatically?:',
```
raymondfeng: Please use the id property name for the message:

```js
message: `Is ${chalk.yellow(this.propName)} generated automatically?`,
```

Issue #3645, opened 2019-09-03
why Loopback + mongodb is not working?


rsa408:
Hi guys,

it is 5 day that I try to start lb4 with MongoDB all the movies and tutorials I looked at the net but my problem was solved. I made git repo and uploaded a very basic code.[https://github.com/RsaLB4/test](https://github.com/RsaLB4/test)

I will be appreciated if you can help me. 
screenshots also are there.

Great thanks
------------------------------------------------------------------
The steps I have followed to create project:

```
C:>lb4 test
? Project description: test
? Project root directory: test
? Application class name: TestApplication
? Select features to enable in the project (Press <space> to select, <a> to toggle all, <i> to invert selection)Enable eslint, Enable prettier, Enable mocha, Enable loopbackBuild, Enable vscode, Enable docker, Enable repositories, Enable services

   create .eslintignore
   create .eslintrc.js
   create .mocharc.json
   create .npmrc
   create .prettierignore
   create .prettierrc
   create DEVELOPING.md
   create index.ts
   create package.json
   create README.md
   create tsconfig.json
   create .vscode\settings.json
   create .vscode\tasks.json
   create .gitignore
   create .dockerignore
   create Dockerfile
   create index.js
   create public\index.html
   create src\application.ts
   create src\index.ts
   create src\migrate.ts
   create src\sequence.ts
   create src\controllers\index.ts
   create src\controllers\ping.controller.ts
   create src\controllers\README.md
   create src\datasources\README.md
   create src\models\README.md
   create src\repositories\README.md
   create src_tests_\README.md
   create src_tests_\acceptance\home-page.acceptance.ts
   create src_tests_\acceptance\ping.controller.acceptance.ts
   create src_tests_\acceptance\test-helper.ts

npm WARN deprecated superagent@3.8.3: Please note that v5.0.1+ of superagent removes User-Agent header by default, therefore you may need to add it yourself (e.g. GitHub blocks requests without a User-Agent header).  This notice will go away with v5.0.2+ once it is released.

core-js@2.6.9 postinstall C:\test\node_modules\core-js
node scripts/postinstall || echo "ignore"
Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!
The project needs your help! Please consider supporting of core-js on Open Collective or Patreon:

https://opencollective.com/core-js
https://www.patreon.com/zloirock

Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)
[...]

PR #3667, opened 2019-09-04
Fix/model with id required

jannyHou:
Fixes https://github.com/strongloop/loopback-next/issues/3645

When `id` is required in a model, the expected `create` method in REST controller should be 
```ts
  @post('/todos', {
    responses: {
      '200': {
        description: 'Todo model instance',
        content: {'application/json': {schema: getModelSchemaRef(Todo)}},
      },
    },
  })
  async createTodo(
    @requestBody({
      content: {
        'application/json': {
          // DIFFERENCE
          // instead of `getModelSchemaRef(Todo, {exclude: ['id']})
          schema: getModelSchemaRef(Todo),
        },
      },
    })
   // DIFFERENCE
   // instead of `Omit<Todo, 'id'>`
    todo: Todo,
  ): Promise<Todo> {
    return this.todoRepo.create(todo);
  }
```

This PR generates the right signature for `create` in the controller generator's template.

At first I didn't know if `create` method is the only one affected, so I added another model with id required in the example `todo` to figure out the fix's scope. It turns out other methods are fine. 

I temporary keep the new `TodoWithId` model and its related tests in the example package so reviewers know what's the expected controller file content, and in case you have questions regarding other methods.

Will remove the changes in `examples/todo` when land the PR.
-->

## Checklist

### DIFF
--- a/packages/cli/generators/controller/index.js
+++ b/packages/cli/generators/controller/index.js
@@ -182,6 +182,12 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
         default: 'number',
       },
       {
+        type: 'confirm',
+        name: 'idOmitted',
+        message: 'Is the id omitted when creating a new instance?',
+        default: true,
+      },
+      {
         type: 'input',
         name: 'httpPathName',
         message: 'What is the base HTTP path name of the CRUD operations?',
--- a/packages/cli/test/integration/generators/controller.integration.js
+++ b/packages/cli/test/integration/generators/controller.integration.js
@@ -100,7 +100,7 @@ describe('lb4 controller', () => {
       restCLIInput,
     );
 
-    it('creates REST CRUD template with valid input', async () => {
+    it('creates REST CRUD template with valid input - id omitted', async () => {
       await testUtils
         .executeGenerator(generator)
         .inDir(SANDBOX_PATH, () =>
@@ -111,7 +111,23 @@ describe('lb4 controller', () => {
         )
         .withPrompts(restCLIInputComplete);
 
-      checkRestCrudContents();
+      checkRestCrudContents({idOmitted: true});
+    });
+
+    it('creates REST CRUD template with valid input', async () => {
+      await testUtils
+        .executeGenerator(generator)
+        .inDir(SANDBOX_PATH, () =>
+          testUtils.givenLBProject(SANDBOX_PATH, {
+            includeDummyModel: true,
+            includeDummyRepository: true,
+          }),
+        )
+        .withPrompts(
+          Object.assign({}, restCLIInputComplete, {idOmitted: false}),
+        );
+
+      checkRestCrudContents({idOmitted: false});
     });
 
     describe('HTTP REST path', () => {
@@ -228,21 +244,7 @@ function checkBasicContents() {
   assert.fileContent(expectedFile, /constructor\(\) {}/);
 }
 
-/**
- * Assertions against the template to determine if it contains the
- * required signatures for a REST CRUD controller, specifically to ensure
- * that decorators are grouped correctly (for their corresponding
- * target functions)
- */
-function checkRestCrudContents() {
-  assert.fileContent(expectedFile, /class ProductReviewController/);
-
-  // Repository and injection
-  assert.fileContent(expectedFile, /\@repository\(BarRepository\)/);
-  assert.fileContent(expectedFile, /barRepository \: BarRepository/);
-
-  // Assert that the decorators are present in the correct groupings!
-  // @post - create
+function checkCreateContentsWithIdOmitted() {
   const postCreateRegEx = [
     /\@post\('\/product-reviews', {/,
     /responses: {/,
@@ -254,6 +256,45 @@ function checkRestCrudContents() {
   postCreateRegEx.forEach(regex => {
     assert.fileContent(expectedFile, regex);
   });
+}
+
+/**
+ * Check the contents for operation 'create' when id is not required.
+ */
+function checkCreateContents() {
+  const postCreateRegEx = [
+    /\@post\('\/product-reviews', {/,
+    /responses: {/,
+    /'200': {/,
+    /description: 'ProductReview model instance'/,
+    /content: {'application\/json': {schema: getModelSchemaRef\(ProductReview\)}},\s{1,}},\s{1,}},\s{1,}}\)/,
+    /async create\(\s+\@requestBody\({\s+content: {\s+'application\/json': {\s+schema: getModelSchemaRef\(ProductReview, {\s+title: 'NewProductReview',\s+}\),\s+},\s+},\s+}\)\s+productReview: ProductReview,\s+\)/,
+  ];
+  postCreateRegEx.forEach(regex => {
+    assert.fileContent(expectedFile, regex);
+  });
+}
+
+/**
+ * Assertions against the template to determine if it contains the
+ * required signatures for a REST CRUD controller, specifically to ensure
+ * that decorators are grouped correctly (for their corresponding
+ * target functions)
+ */
+function checkRestCrudContents(options) {
+  assert.fileContent(expectedFile, /class ProductReviewController/);
+
+  // Repository and injection
+  assert.fileContent(expectedFile, /\@repository\(BarRepository\)/);
+  assert.fileContent(expectedFile, /barRepository \: BarRepository/);
+
+  // Assert that the decorators are present in the correct groupings!
+  // @post - create
+  if (options && options.idOmitted) {
+    checkCreateContentsWithIdOmitted();
+  } else {
+    checkCreateContents();
+  }
 
   // @get - count
   const getCountRegEx = [
--- a/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
+++ b/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
@@ -40,12 +40,12 @@ export class <%= className %>Controller {
         'application/json': {
           schema: getModelSchemaRef(<%= modelName %>, {
             title: 'New<%= modelName %>',
-            exclude: ['<%= id %>'],
+            <%if (idOmitted) {%>exclude: ['<%= id %>'],<% } %>
           }),
         },
       },
     })
-    <%= modelVariableName %>: Omit<<%= modelName %>, '<%= id %>'>,
+    <%= modelVariableName %>: <% if (!idOmitted) { -%><%= modelName %><% } else { -%>Omit<<%= modelName %>, '<%= id %>'><% } -%>,
   ): Promise<<%= modelName %>> {
     return this.<%= repositoryNameCamel %>.create(<%= modelVariableName %>);
   }


packages/cli/generators/controller/index.js
```diff
@@ -181,6 +181,12 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
         when: this.artifactInfo.idType === undefined,
         default: 'number',
       },
+      {
+        type: 'confirm',
+        name: 'idRequired',
+        message: 'Is the id required when creating an instance?',
+        default: false,
```
raymondfeng: Can we check the `generated` flag for the id property? See https://github.com/strongloop/loopback-next/pull/3643

jannyHou: @raymondfeng The fix in this PR is in the **controller** generator, so it doesn't have the `generated` flag (#3643 is modifying **model** generator)

raymondfeng: Can the controller code generator have access the model definition to check?

bajtos: It would be great if `lb4 controller` could automatically infer whether the `id` property is required or forbidden when creating new models. At the same time, I think the current proposal is a meaningful improvement, therefore it should be landed. We can improve the generator to parse `id` property definition later.

IIRC, there are multiple ways how to end up with a model that requires `id` values on creation. The model-level setting `forceId` plays an important role.

```ts
@model({settings: {forceId: false}})
export class TodoWithId extends Entity {
  @property({
    type: 'number',
    id: true,
  })
  id?: number;

  // ...
}
```

When `forceId` is disabled, clients have can either supply their own id, or let the database generate a new unique id for them. (In the schema, the `id` property should be shown as optional.) That's my understanding of how things should work, I did not run a test to verify the actual behavior.

packages/cli/test/integration/generators/controller.integration.js
```diff
@@ -114,6 +114,22 @@ describe('lb4 controller', () => {
...
+          testUtils.givenLBProject(SANDBOX_PATH, {
+            includeDummyModel: true,
+            includeDummyRepository: true,
+          }),
+        )
+        .withPrompts(
+          Object.assign({}, restCLIInputComplete, {idRequired: true}),
+        );
+
+      checkRestCrudContents(1);
```
bajtos: Boolean arguments are difficult to understand. What does `1` mean in this context? I have to look up the definition of `checkRestCrudContents` to understand that.

Let's use named flags (an `options` arg) please.

```suggestion
      checkRestCrudContents({idRequired: true});
```

examples/todo/src/models/todo-with-id.model.ts
```diff
@@ -0,0 +1,37 @@
...
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class TodoWithId extends Entity {
+  @property({
+    type: 'number',
+    id: true,
+    required: true,
+  })
+  id?: number;
```
bajtos: Since `id` is `required: true`, I think we should tell TypeScript to make it require too?

```suggestion
  id: number;
```

packages/cli/generators/controller/index.js
jannyHou: @raymondfeng 
> Can the controller code generator have access the model definition to check?

It should have access to the model def if the file exist. Then use `ts-morph` to read the element. 

And I agree with @bajtos that we probably can land this PR first, since in some situations, if `idRequired` cannot be inferred from the model class, a prompt is still needed.

Like the [id name prompt](https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/controller/index.js#L170), when `id` cannot be inferred from the `artifactInfo`, the prompt will ask for it.

I will submit another PR to read the `idRequired` from model file and add a `when` to the prompt. I believe there would be more discussion about how we infer the flag :)

bajtos:
@jannyHou one more thing: please update the `rest-crud` package to stay in sync with the template. See [packages/rest-crud/src/crud-rest.controller.ts](https://github.com/strongloop/loopback-next/blob/0475fe660cd2eba43f314b568596aa4b65f66e97/packages/rest-crud/src/crud-rest.controller.ts).

I am not sure if it's possible to apply both changes (JSON schema, TypeScript type). To me, it's important to emit the correct JSON schema, so that the REST API validation does not reject valid requests. As for the TypeScript type, we can start with something like `Todo & {id?: ID}` - I think that should cover all cases (id required, id forbidden).

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -70,11 +70,11 @@ export interface CrudRestController<
...
-  /**
-   * Implementation of the endpoint `POST /`.
-   * @param data Model data
-   */
-  create(data: Omit<T, IdName>): Promise<T>;
+  // /**
+  //  * Implementation of the endpoint `POST /`.
+  //  * @param data Model data
+  //  */
+  // create(data: Omit<T, IdName>): Promise<T>;
```
jannyHou: cc @bajtos Since the type of `data` is not determined (w/ or w/o id), I removed the create function in this interface. Thought?

examples/todo/src/__tests__/acceptance/todo-with-id.acceptance.ts
```diff
@@ -0,0 +1,198 @@
...
+import {TodoWithId} from '../../models';
+import {TodoWithIdRepository} from '../../repositories';
+import {
+  getProxiedGeoCoderConfig,
+  givenCachingProxy,
+  givenTodoWithId,
+  HttpCachingProxy,
+} from '../helpers';
+
+describe('TodoApplication - test TodoWithId', () => {
```
bajtos: How does `TodoWithId` fit into the narrative of our [Todo tutorial](https://loopback.io/doc/en/lb4/todo-tutorial.html)?

In my mind, the primary purpose of Todo example app is to support the tutorial and show a simple & meaningful LB4 application. It's not the best place for adding acceptance tests that are not related to the Todo tutorial.

packages/rest-crud/src/__tests__/acceptance/default-model-crud-rest.acceptance.ts
```diff
@@ -60,29 +76,58 @@ describe('CrudRestController for a simple Product model', () => {
...
-        // FIXME: POST should return 201
-        // See https://github.com/strongloop/loopback-next/issues/788
-        .expect(200);
-
-      const created = response.body;
-      expect(created).to.containEql({name: 'A pen'});
-      expect(created)
-        .to.have.property('id')
-        .of.type('number');
+    context('id not required', () => {
```
bajtos:
```suggestion
    context('id generated by the database', () => {
```

packages/rest-crud/src/__tests__/acceptance/default-model-crud-rest.acceptance.ts
```diff
@@ -60,29 +76,58 @@ describe('CrudRestController for a simple Product model', () => {
...
+          .expect(422);
+      });
     });
 
-    it('rejects request with `id` value', async () => {
-      await client
-        .post('/products')
-        .send({id: 1, name: 'a name'})
-        .expect(422);
+    context('id required', () => {
```
bajtos:
```suggestion
    context('id required (supplied by the user)', () => {
```

packages/rest-crud/src/crud-rest.controller.ts
bajtos: I am quite unhappy with the changes made in this file, I feel they are introducing too much complexity. We have started with one option - id is required or generated - and have two controller classes. Imagine what is going happen when the number of options grows. Are we going to have 8 different classes when the number of options grows to 3?

I prefer to sacrifice compile-time safety and keep the implementation simple. I expect there will be very few people consuming these runtime-generated controllers directly from TypeScript code.

I tried my suggestions to replace `Omit<T, IdName>` with `Todo & {id?: ID}` and I see it does not work. Can we use the following type instead? 

```ts
   async create(
      data: T | Omit<T, IdName>,
    ): Promise<T>;
```

packages/rest-crud/src/crud-rest.controller.ts
```diff
@@ -254,8 +241,43 @@ export function defineCrudRestController<
     }
   }
 
-  // See https://github.com/microsoft/TypeScript/issues/14607
-  return CrudRestControllerImpl;
+  if (options && options.idRequired) {
```
bajtos: At runtime, we already know whether the model requires the id to be provided by the user or whether it will be generated by the database. Let's use that information please and don't ask the user to supply it to us again.

Relevant code from LB3:
- https://github.com/strongloop/loopback-swagger/blob/fe5cfd7a11fb4cb0c9804db510d461a3d2ad53a6/lib/specgen/model-helper.js#L46-L60
- https://github.com/strongloop/loopback/blob/aad97c20360de236af6afcb4e364f42d071664b0/lib/model.js#L794-L799
- https://github.com/strongloop/loopback-datasource-juggler/blob/de4718d5b81d29ac23597172d8ee591617924fe0/lib/model.js#L841-L844
- https://github.com/strongloop/loopback-datasource-juggler/blob/de4718d5b81d29ac23597172d8ee591617924fe0/lib/model-builder.js#L343-L362

Please note that we need to support more update-only properties, for example the Cloudant connector uses property `_rev` that is always generated by the database.

I see this is topic is more complex than I initially thought :(

If you prefer, then I am ok to leave the changes in `packages/rest-crud` out of scope of this pull request, so that it can be landed sooner. Just open a follow-up issue and add it to the epic #2036 _From model definition to REST API with no custom repository/controller classes_ please.

bajtos: Sounds good. Please open two stories - one to update `rest-crud`, another to improve the CLI tool.

packages/cli/generators/controller/index.js
```diff
@@ -181,6 +181,12 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
         when: this.artifactInfo.idType === undefined,
         default: 'number',
       },
+      {
+        type: 'confirm',
+        name: 'idRequired',
+        message: 'Is the id required when creating an instance?',
```
bajtos: I am not sure if this question makes it clear what are the implications of `yes` and `no` answers. Every model instance must eventually have some id value set in the database, thus one can argue that the id property is always required (in a way).

I think it may be better to instead ask the user if the `id` property is generated by the database. When the id is generated, then it must be excluded from `create` requests - I thinks this implication is easier to understand.

```suggestion
        message: 'Is the id (primary key) value generated by the database?',
```

If you decide to accept my suggestion, then please rename the question/answer name from `idRequired` to something more sensible.

jannyHou: @bajtos Hmm, I have a different opinion on this, I am thinking of being consistent with the LB 3 options.

See https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html

`required`: Implies "Whether a value for the property is required. If true, then adding or updating a model instance requires a value for the property." and this is my purpose of adding the prompt.

`generated`: "The generated property indicates the ID will be automatically generated by the database. If true, the connector decides what type to use for the auto-generated key. For relational databases, such as Oracle or MySQL, it defaults to number. If your application generates unique IDs, set it to false." It describes a more complicated scenario than whether a property is required or not in the request payload...

bajtos: I see your point.

There is still one missing piece though: the id property is special, there are three possible variants:

1. The id is required when creating an instance. (It's always generated by the client, useful e.g. for offline-first.) TS type to use: `T` (e.g. `Product`)
2. The id is not required but can be provided when creating an instance. (The client can decide whether they want to supply their own id or let the DB create one.) TS type to use: `T & {id?: <number|string>}` (I am not entirely sure about the right type)
3. The id must not be provided and will be always generated by the database. TS type to use: `Omit<T, 'id'>`.

By default, LB apps are configured for the option 3 - the `id` property is not allowed in "create" requests.

That's why I find your prompt confusing, because to me, it selects between the first and second variant, but then emits TypeScript type for the third variant.

jannyHou: @bajtos I see, maybe let me change the prompt to be "Is the id omitted when creating an instance?"

- Yes: use `Omit<T, 'id'>`
- No: use `T`

For the 2nd scenario, I think it could be same as 1 if `id` is defined as optional property in the model, like:

```ts
// scenario 1
@model()
class Product{
  @property()
   id: string;
}

// scenario 2
@model()
class Product{
  @property()
   id?: string;
}
```

jannyHou:
@bajtos see my comment and I applied the change accordingly: https://github.com/strongloop/loopback-next/pull/3667#discussion_r324699118, thank you

packages/cli/generators/controller/index.js
```diff
@@ -183,9 +183,9 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
       },
       {
         type: 'confirm',
-        name: 'idRequired',
-        message: 'Is the id required when creating an instance?',
-        default: false,
+        name: 'idOmitted',
+        message: 'Is the id omitted when creating an instance?',
```
bajtos: Nitpick: I would use "creating a new instance" to make it even more obvious that this prompt is related to new instances only

PR #3676, opened 2019-09-06
feat(context): add more logs to simplify troubleshooting

bajtos:
Extracted from https://github.com/strongloop/loopback-next/pull/3617

## Checklist

### DIFF
--- a/packages/context/src/binding-inspector.ts
+++ b/packages/context/src/binding-inspector.ts
@@ -4,11 +4,14 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {MetadataAccessor, MetadataInspector} from '@loopback/metadata';
+import * as debugFactory from 'debug';
 import {Binding, BindingScope, BindingTag, BindingTemplate} from './binding';
 import {BindingAddress} from './binding-key';
 import {ContextTags} from './keys';
 import {Provider} from './provider';
 import {Constructor} from './value-promise';
+
+const debug = debugFactory('loopback:context:binding-inspector');
 
 /**
  * Binding metadata from `@bind`
@@ -63,10 +66,11 @@ export function isProviderClass(
 export function asProvider(
   target: Constructor<Provider<unknown>>,
 ): BindingTemplate {
-  return binding =>
+  return function bindAsProvider(binding) {
     binding.toProvider(target).tag(ContextTags.PROVIDER, {
       [ContextTags.TYPE]: ContextTags.PROVIDER,
     });
+  };
 }
 
 /**
@@ -78,7 +82,7 @@ export function asClassOrProvider(
   target: Constructor<unknown>,
 ): BindingTemplate {
   // Add a template to bind to a class or provider
-  return binding => {
+  return function bindAsClassOrProvider(binding) {
     if (isProviderClass(target)) {
       asProvider(target)(binding);
     } else {
@@ -94,7 +98,7 @@ export function asClassOrProvider(
 export function asBindingTemplate(
   scopeAndTags: BindingScopeAndTags,
 ): BindingTemplate {
-  return binding => {
+  return function applyBindingScopeAndTag(binding) {
     if (scopeAndTags.scope) {
       binding.inScope(scopeAndTags.scope);
     }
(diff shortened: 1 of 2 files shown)


packages/context/src/binding-inspector.ts
```diff
@@ -215,10 +220,16 @@ export function createBindingFromClass<T = unknown>(
...
 ): Binding<T> {
+  debug('create binding from class %s with options', cls.name, options);
   const templateFn = bindingTemplateFor(cls);
   let key = options.key;
   if (!key) {
-    key = buildBindingKey(cls, options);
+    try {
+      key = buildBindingKey(cls, options);
+    } catch (err) {
+      err.message += ` (while building binding key for class ${cls.name})`;
```
raymondfeng: Is mutating the error message too invasive?

packages/context/src/binding-inspector.ts
```diff
@@ -215,10 +220,16 @@ export function createBindingFromClass<T = unknown>(
...
+      throw err;
```
raymondfeng: Can we move the `try-catch` to `createBindingFromClass` so that we can have the `class` information for all errors?

packages/context/src/binding-inspector.ts
bajtos: Cross-posting from https://github.com/strongloop/loopback-next/pull/3617#discussion_r319084318:

> > Is there a better way than mutating the error message?
> 
> I am not aware of any. It's important to preserve the original stack trace and AFAIK, that requires us to mutate the original error object.

Do you have any specific concern in mind? I am not aware of any problems that could be caused by editing error messages.

packages/context/src/binding-inspector.ts
bajtos: This `try-catch` is already inside `createBindingFromClass` function.

```ts
export function createBindingFromClass<T = unknown>(
  cls: Constructor<T | Provider<T>>,
  options: BindingFromClassOptions = {},
): Binding<T> {
  debug('create binding from class %s with options', cls.name, options);
  const templateFn = bindingTemplateFor(cls);
  let key = options.key;
  if (!key) {
    try {
      key = buildBindingKey(cls, options);
    } catch (err) {
      err.message += ` (while building binding key for class ${cls.name})`;
      throw err;
    }
  }

  // ...
}
```

Are you suggesting to wrap the entire body of `createBindingFromClass` into a try-catch blog?

raymondfeng: Yes. As the method knows the class.

packages/context/src/binding-inspector.ts
raymondfeng: I was looking for a more systematic solution like https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace. But I'm fine to change the message for this PR.

packages/context/src/binding-inspector.ts
bajtos: Done.

While making this change, I did a small refactoring of `createBindingFromClass` and extracted a new helper function `applyClassBindingOptions`. This makes the patch smaller (less lines of code are changing the indentation level).

packages/context/src/binding-inspector.ts
```diff
@@ -215,12 +220,23 @@ export function createBindingFromClass<T = unknown>(
   cls: Constructor<T | Provider<T>>,
   options: BindingFromClassOptions = {},
 ): Binding<T> {
-  const templateFn = bindingTemplateFor(cls);
-  let key = options.key;
-  if (!key) {
-    key = buildBindingKey(cls, options);
```
bajtos: There is no need to check `options.key` here, it's already handled by `buildBindingKey`.

packages/context/src/binding-inspector.ts
```diff
@@ -272,11 +287,13 @@ function buildBindingKey(
   cls: Constructor<unknown>,
   options: BindingFromClassOptions = {},
 ) {
+  if (options.key) return options.key;
```
bajtos: Small optimization - we don't need to call `buildingTemplateFor` and create a new `Binding` when `options.key` is set. Return as early as possible in that case.

Issue #3675, opened 2019-09-06
app.stop() should close all database connections


bajtos:
## Suggestion

When the user calls `app.stop()`, all database connections should be closed and any other handles should be released. This way the Node.js process should naturally exit because there is no more work to be done.

See the discussion in https://github.com/strongloop/loopback-next/issues/3672 for more details.

## Use Cases

- Run `mocha` tests without `--exit` flag.
- Graceful shutdown.

## Acceptance criteria

TBD - will be filled by the team.

raymondfeng:
I see a few options to implement this feature:

1. Add `stop` alias to juggler datasource to call `disconnect` and register all datasources as life cycle observers. We probably don't want to call `connect` during `start` to honor `lazyConnect`.

2. Change the CLI template for datasource classes with lifecycle observer signature. This is probably more flexible as application developers can decide if they to override the start/stop.

bajtos:
> Change the CLI template for datasource classes with lifecycle observer signature. This is probably more flexible as application developers can decide if they to override the start/stop.

+1

On the second thought, I see also a third option - modify our connector to unref the TCP connection handles they hold in the connection pool, so that Node.js don't consider them as "active" when deciding whether the process has finished.

derdeka:
@bajtos @raymondfeng
Is this issue resolved as #3678 is merged? (lifecycle support to close the connection)

raymondfeng:
@derdeka Thanks for the reminder.

PR #3678, opened 2019-09-06
feat(cli): add lifecycle support for datasources

raymondfeng:
Implements https://github.com/strongloop/loopback-next/issues/3675

## Checklist

### DIFF
--- a/packages/cli/generators/datasource/templates/datasource.ts.ejs
+++ b/packages/cli/generators/datasource/templates/datasource.ts.ejs
@@ -1,8 +1,15 @@
-import {inject} from '@loopback/core';
+import {
+  inject,
+  lifeCycleObserver,
+  LifeCycleObserver,
+  ValueOrPromise,
+} from '@loopback/core';
 import {juggler} from '@loopback/repository';
 import * as config from './<%= jsonFileName %>';
 
-export class <%= className %>DataSource extends juggler.DataSource {
+@lifeCycleObserver('datasource')
+export class <%= className %>DataSource extends juggler.DataSource
+  implements LifeCycleObserver {
   static dataSourceName = '<%= name %>';
 
   constructor(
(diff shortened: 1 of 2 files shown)


packages/cli/generators/datasource/templates/datasource.ts.ejs
```diff
@@ -1,7 +1,8 @@
-import {inject} from '@loopback/core';
+import {inject, lifeCycleObserver} from '@loopback/core';
 import {juggler} from '@loopback/repository';
 import * as config from './<%= jsonFileName %>';
 
+@lifeCycleObserver('datasource')
 export class <%= className %>DataSource extends juggler.DataSource {
```
derdeka: should the class also implement LifeCycleObserver?
```typescript
export class <%= className %>DataSource extends juggler.DataSource implements LifeCycleObserver {
```

bajtos: +1

raymondfeng: It's not required to be functional but we can add it to be explicit.

bajtos: By adding `implements LifeCycleObserer`, we allow TypeScript compiler to catch certain kind of errors and also make future refactorings easier because of the extra type information.

packages/cli/generators/datasource/templates/datasource.ts.ejs
```diff
@@ -11,4 +12,19 @@ export class <%= className %>DataSource extends juggler.DataSource {
...
+   */
+  start() {
+    // Add your logic here to be invoked when the application is started
+  }
+
+  /**
+   * Disconnect the datasource when application is stopped. This allows the
+   * application to be shut down gracefully.
+   */
+  stop() {
```
derdeka: When I apply this to by lb4 project, i get:
`The inferred type of 'stop' cannot be named without a reference to '@loopback/repository/node_modules/loopback-datasource-juggler'. This is likely not portable. A type annotation is necessary.`

For me it works only that way:
```typescript
stop(): ValueOrPromise<void> {
```
What I'm doing wrong?

raymondfeng: I'm adding the return type.

Issue #2480, opened 2019-02-26
Import LB3 model JSON file to LB4 model class


bajtos:
As an LB3 user migrating my model definition files to LB4, I would like to call `lb4` CLI to generate an LB4 model class from my LB3 model JSON file.

```
lb4 import-lb3-models legacy-lb3app/server/server.js
# load the LB3 app
# ask the user which models to import
# create src/models/product.ts etc.
```

The initial version should handle property definitions only.

For example, let's take an LB3 Product model with a `name` property:

```
{
  "name": "Product",
  "base": "PersistedModel",
  "properties": {
    "name": {
      "type": "string",
      "required": true
    }
  }
}
```

This LB3 definition should be translated into an LB4 TypeScript model class:

```ts
@model()
class Product extends Entity {
  @property({required: true})
  name: string;
}
```

## Acceptance criteria

- [x] A new CLI flag for `lb4 model`, or perhaps a new command, accepting a path to LB3 model file, that will scaffold (code-generate) an LB4 model file from LB3 definition. See #2245 for related prior work. (/cc @marvinirwin)
  --> see https://github.com/strongloop/loopback-next/pull/3688
- [x] The following base model classes should be recognized and mapped to LB4:
   - `Model` -> `Model`
   - `PersistedModel` -> `Entity`
   - `KeyValueModel` -> (to be determined)
 When the base model class is not recognized, the CLI should abort with a descriptive error.
  --> see https://github.com/strongloop/loopback-next/pull/3688
- [x] The following model settings should trigger a warning when a non-empty value is provided in LB3 model file:
   - `relations`
   - `method`
   - `mixins`
   - `acls`
  --> see https://github.com/strongloop/loopback-next/pull/3688
- [x] All other model settings should be copied as-is to LB4 model definition. IMPORTANT: settings can be specified either at top-level or inside `options` property, we need to support both flavors. Notable settings to test & support:
    - `forceId`
    - `strict`
    - connector-specific config like SQL table name and MongoDB collection name
  --> see https://github.com/strongloop/loopback-next/pull/3688
- [x] Documentation at CLI level
   --> see https://github.com/strongloop/loopback-next/pull/3688
- [x] A blog post
- [x] Follow-up stories for things that were left out of the initial implementation. Most notably support for non-default `base` model classes and relations.

See #2036 for the next step that will allow developers to take LB4 model and expose it via REST with little to no configuration.

**Open discussion points & todos**
[...]

PR #3688, opened 2019-09-09
CLI: Import LB3 model(s) into LB4 project [EXPERIMENTAL]

bajtos:
See https://github.com/strongloop/loopback-next/issues/2480

This pull request is adding a new CLI command `lb4 import-lb3-models`:

- The command requires a path to a LB3 application (either the top-level project directory or the server file).
- After the app is loaded via `require()`, the user is presented with a list of models to import.
- Important: built-in models like `User` and `AccessToken` are included in that list.
- For each selected model, a corresponding `src/models/{model-name}.ts` file is created.

Known limitations (features _intentionally_ left out of this initial pull request):

- Connector-specific metadata in property definitions is not imported. #3810
- When a property is defined with a complex object type, the nested property definitions are not converted from LB3 to LB4 format. #3811
- Model relations are not imported. #3812
- Models inheriting from application-specific models (including LB3 built-in   models like `User`) cannot be imported yet. #3813
- For models attached to MongoDB datasource, the imported LB4 model contains incorrect definition of the primary key property of `ObjectID` type. #3814

/cc @strongloop/loopback-next 

## Checklist


packages/cli/generators/import-model/index.js
```diff
@@ -0,0 +1,175 @@
...
+module.exports = class ModelImporter extends BaseGenerator {
+  constructor(args, opts) {
+    super(args, opts);
+
+    this.argument('modelFile', {
+      type: String,
+      required: true,
+      description:
+        'Path to the model JSON file to import, ' +
+        'e.g. "lb3app/common/models/{model-name}.json"',
```
raymondfeng: Some metadata are in `model-config.json`, we should read them too.

docs/site/Importing-LB3-models.md
```diff
@@ -0,0 +1,26 @@
...
+LoopBack 3 models into your LoopBack 4 project.
+
+## Overview
+
+A single model can be imported by running `lb4 import-model` command.
+
+### Arguments
+
+`modelFile`: Path to the JSON file with model definition, e.g.
+`lb3app/common/models/my-model.json`.
```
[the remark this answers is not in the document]
bajtos: I agree your proposal offers much better UX and additional flexibility. In LB3, certain aspects depend on the connector the model is attached to, e.g. `strict` mode (enabled for SQL databases, disabled for NoSQL) and `id` type (typically a `number` for `memory` & SQL, a `string` for NoSQL)..

However, I am concerned about the implementation complexity. In LB3, model definitions can be stored in multiple directories. By default, projects are configured with two locations `common/models` and `server/models`, but users can add more directories via `server/model-config.json`.

I'd prefer to start with the current workflow (it's already enough work for a single pull request) and open a new issue to implement your proposal as an incremental enhancement.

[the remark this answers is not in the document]
bajtos: See also https://github.com/strongloop/loopback-next/issues/2480, it has smaller scope compared to what you are proposing.

packages/cli/generators/import-model/index.js
bajtos: Let's leave that out of the initial implementation, I'd like to avoid scope creep. See the acceptance criteria https://github.com/strongloop/loopback-next/issues/2480.

bajtos:
Based on feedback from @raymondfeng and the fact that many important parts of model definitions are contributed by runtime (e.g. PK type and `strict` mode is decided by the connector), I decided to rework `lb4 import-model` to load the entire LB3 application and obtain model definitions from runtime metadata.

Screenshot of the new version at work, using our [examples/lb3-application/lb3app](https://github.com/strongloop/loopback-next/tree/master/examples/lb3-application/lb3app):

```
$ lb4 import-model lb3app/server/server.js

**WARNING WARNING WARNING**
This command is experimental and not feature-complete yet.
Watch the progress at https://github.com/strongloop/loopback-next/issues/2480

? Select models to import: CoffeeShop
Model CoffeeShop will be created in src/models/coffee-shop.model.ts

   create src/models/coffee-shop.model.ts
   update src/models/index.ts
```

Here is the content of generated `src/models/coffee-shop.model.ts`:

```ts

@model({
  settings: {
    strict: false,
    forceId: false,
    base: 'PersistedModel',
    replaceOnPUT: true,
    validateUpsert: true,
    idInjection: true
  }
})
export class CoffeeShop extends Entity {
  @property({
    type: 'number',
    id: 1,
    generated: true,
    updateOnly: false,
  })
  id?: number;

  @property({
    type: 'string',
    required: true,
  })
  name: string;

  @property({
    type: 'string',
  })
  city?: string;

  // Define well-known properties here

  // Indexer property to allow additional data
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  [prop: string]: any;

  constructor(data?: Partial<CoffeeShop>) {
    super(data);
  }
}

export interface CoffeeShopRelations {
}

export type CoffeeShopWithRelations = CoffeeShop & CoffeeShopRelations;
```

@raymondfeng what do you think, is this closer to the user experience you are envisioning for `lb4 import-model`?

@strongloop/loopback-next WDYT?

docs/site/Importing-LB3-models.md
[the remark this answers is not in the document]
bajtos: Reworked the code to load the entire app, see https://github.com/strongloop/loopback-next/pull/3688#issuecomment-533070495

> Build a list of model names for the user to select (all selected by default)

I am not selecting all models by default. I am not sure if that's a good idea, considering that the current list includes built-in models too. Let's leave this out of scope of this initial PR please (but possibly still in scope of #2480).

packages/cli/generators/import-model/index.js
bajtos: Reworked the code to load the entire app, see https://github.com/strongloop/loopback-next/pull/3688#issuecomment-533070495

This way we take into account any extra configuration provided in `model-config.json`.

docs/site/Importing-LB3-models.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+## Synopsis
+
+To simplify migration from LoopBack 3, LoopBack 4 provides a CLI tool to import
+LoopBack 3 models into your LoopBack 4 project.
+
+## Overview
+
+A single model can be imported by running `lb4 import-model` command.
```
raymondfeng: I wonder if the command name can be better as `lb4 import-lb3-models` or `lb4 import-v3-models`?

bajtos: Good point. Let's go with `lb4 import-lb3-model`.

raymondfeng:
@bajtos Please investigate CI failures.

bajtos:
@raymondfeng thank you for the initial approval. The pull request was not done yet. Today, I added many new tests and fixed handling of edge cases. Now I consider the pull request ready for final review & landing.

Does it still LGTY?

packages/cli/generators/import-lb3-model/model-names.js
```diff
@@ -0,0 +1,32 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+module.exports = {
+  canImportModelName,
+};
+
+// List of built-in LB models to exclude from the prompt
+const EXCLUDED_MODEL_NAMES = [
```
agnes512: I am a bit lost here. In the docs, it says
> you can modify your LB3 model to inherit from `Model`,
>  `PersistedModel` or `KeyValueModel`, then import the model and finally update
>  the generated model

So the `EXCLUDED_MODEL_NAMES` here means, if a model extends from the list, can it be imported or not?

docs/site/Importing-LB3-models.md
```diff
@@ -0,0 +1,79 @@
...
+LoopBack 3 models into your LoopBack 4 project.
+
+{% include warning.html content="
+This command is experimental and not feature-complete yet.
+See the list of known limitations below.
+" %}
+
+## Overview
+
+A single model can be imported by running `lb4 import-lb3-model` command.
```
hacksparrow: Is it assumed that the command for multiple models is `lb4 import-lb3-models`?

hacksparrow: Looks like there is no `lb4 import-lb3-models`.

The statement "A single model can be imported by running `lb4 import-lb3-model` command.", makes it sound like there may be something for multiple models.

So do we have to execute `lb4 import-lb3-model` as many times as there are model that one wants to import?

Why not prompt, "Do you want to import <model>?" one after another?

raymondfeng: I think the agreement and implementation allow multiple models to be imported in one run. Let's rename `import-lb3-model` to `import-lb3-models` across the board.

docs/site/tables/lb4-artifact-commands.html
```diff
@@ -23,6 +23,11 @@
       <td><a href="DataSource-generator.html">DataSource generator</a></td>
     </tr>
 
+    <tr>
+      <td><code>lb4 import-lb3-model</code></td>
+      <td>Import a LoopBack 3 model to a LoopBack 4 application</td>
```
raymondfeng: `Import LoopBack 3 models to a LoopBack 4 application`

docs/site/tables/lb4-artifact-commands.html
```diff
@@ -23,6 +23,11 @@
...
+      <td><a href="Importing-LB3-models.html">Model generator</a></td>
```
raymondfeng: Importer for LoopBack 3 models

packages/cli/generators/import-lb3-model/index.js
```diff
@@ -0,0 +1,163 @@
...
+      this.destinationPath(),
+      this.artifactInfo.outDir,
+    );
+    return super.setOptions();
+  }
+
+  async logExperimentalStatus() {
+    this.log(
+      chalk.red(
+        '\n**WARNING WARNING WARNING**\n' +
```
raymondfeng: Let's use one `WARNING` as the text is already in RED.

packages/cli/generators/import-lb3-model/model-names.js
bajtos: > if a model extends from the list, can it be imported or not?

TL;DR: These two aspects are orthogonal. If a model extends from `Model`, `PersistedModel` or `KeyValueModel` (all of them are in the list), then it can be imported. If it inherits from a different model class, then it cannot be imported yet.

We have two list of model names:

1. List of model names to hide from the CLI prompt when asking which app models to import. This list is `EXCLUDED_MODEL_NAMES` plus any model name starting with `AnonymousModel_`.

    The purpose of this filter is to hide models that cannot be imported (e.g. `Model` and `PersistedModel` are built-in in both LB3 and LB4) or which does not make sense to import (e.g. `Email`).

    The filter does not take into account model inheritance.

    I suppose I could improve the filter to take into account model inheritance. The difficult part is how to explain the user why some models are included in the list and some are not. I would prefer to invest our time into implementing import of models inheriting from other app models instead.

2. List of base models that we know how to handle at LB4 side, see https://github.com/strongloop/loopback-next/blob/42ec9ed2b972fec9ae25b974b65c268a6b2ed34b/packages/cli/generators/import-lb3-model/migrate-model.js#L120-L124

    I created this list by picking built-in LB3 models that have their direct counter-part in LB4.

I'll be opening a new GH issue to implement support for importing models that are inheriting from a model that's not built into LB4. There are two major use cases - import from an LB3 built-in model like `User` and import from an application-specific model. There are few difficult aspects to figure out, that's why I am leaving this feature out of scope of the initial version.
- How to know where to import the base model from (what path to use in the `import` statement)?
- How to ensure that the base LB3 model has been either already imported to the LB4 app or is part of the import being in progress?

docs/site/Importing-LB3-models.md
bajtos: > I think the agreement and implementation allow multiple models to be imported in one run.

+1

My very first draft implementation was for a single model only. I forgot to update the documentation page after I reworked the code 🙈 

> Let's rename `import-lb3-model` to `import-lb3-models` across the board.

Will do.

bajtos:
@agnes512 @raymondfeng @hacksparrow thank you for the feedback, I pushed a new commit to address your comments.

Does the patch LGTY now?

bajtos:
One more commit 644d33b - I reworked formatting of "Known limitations" in the docs and added links to follow-up GitHub issues.

Issue #3671, opened 2019-09-05
Unable to create Recursive Model (Non Entity)


mmussche2:
## Steps to reproduce
I have a non Entity Model ReportState which needs to contain an array of ReportStates, It has no ID and is not persisted, it is just part of another model. When starting the API it crashes. Here is my Model : 
```
export class ReportState extends Model {

  @property.array(ReportState, {})
  states: ReportState[];

  @property({
    type: 'string',
  })
  benchmarkId?: string;

  @property({
    type: 'string',
  })
  color?: string;

  constructor(data?: Partial<ReportState>) {
    super(data);
  }
```

## Current Behavior

```
TypeError: Cannot set property 'ReportState' of undefined
    at includeReferencedSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:321:38)
    at modelToJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:292:9)
    at getJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:40:18)
    at modelToJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:291:28)
    at getJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:40:18)
    at modelToJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:291:28)
    at getJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:40:18)
    at modelToJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:291:28)
    at Object.getJsonSchema (C:\Users\mmussche\Documents\workspace\Boccard\ta-api-core-2\node_modules\@loopback\repository-json-schema\dist\build-schema.js:40:18)
[...]

PR #3722, opened 2019-09-12
fix(repository): allow model classes with recursive type references

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/3671#issuecomment-529376105

## Checklist

### DIFF
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -148,6 +148,10 @@ export class DefaultCrudRepository<
       return model as typeof juggler.PersistedModel;
     }
 
+    // To handle circular reference back to the same model,
+    // we create a placeholder model that will be replaced by real one later
+    dataSource.getModel(definition.name, true /* forceCreate */);
+
     // We need to convert property definitions from PropertyDefinition
     // to plain data object because of a juggler limitation
     const properties: {[name: string]: object} = {};
--- a/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -210,6 +210,35 @@ describe('DefaultCrudRepository', () => {
       expect(User.definition.properties.roles.itemType).to.equal(Role);
       expect(User.definition.properties.address.type).to.equal(Address);
     });
+
+    it('handles recursive model references', () => {
+      @model()
+      class ReportState extends Entity {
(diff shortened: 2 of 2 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -127,68 +127,95 @@ export class DefaultCrudRepository<
...
     );
 
-    this.modelClass = this.definePersistedModel(entityClass);
+    this.modelClass = this.definePersistedModel(entityClass, new Map());
   }
 
   // Create an internal legacy Model attached to the datasource
   private definePersistedModel(
     entityClass: typeof Model,
+    visited: Map<typeof Model, typeof juggler.PersistedModel>,
```
bajtos: Can we leverage DataSource's `ModelBuilder` instance that's already keeping track of models that have been defined, instead of creating a new map?

raymondfeng: Probably not as `ModelBuilder` has no idea of LB4 entity class.

raymondfeng:
@bajtos PTAL

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: `ModelBuilder` is keeping track of all models that are backing LB4 entity classes. Whenever we create a new Repository for an LB4 entity class, a new (Persisted)Model is registered with the ModelBuilder behind the datasource.

bajtos:
I am not very happy about the current implementation, because it introduces multiple places that are keeping track of visited models (`visited`, `dataSource.modelBuilder.models`).

Juggler already has means for handling circular references in models. I simplified your solution to leverage the concept of "unresolved models", now the fix consists of one new line added.

@raymondfeng  PTAL.

If you agree with my proposed direction, then I am proposing the following next steps:
1. Fix juggler typings to include `forceCreate` parameter for `dataSource.getModel()`. See [the source](https://github.com/strongloop/loopback-datasource-juggler/blob/ad1777fcd3d1dd120634b7c140c3313559c93965/lib/datasource.js#L914-L916).
2. Publish a new juggler version.
3. In this patch: remove the cast to `any`, squash the commits. Please preserve my co-authorship via `Co-authored-by` trailer, see [GitHub docs](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line).

raymondfeng:
@bajtos Please approve https://github.com/strongloop/loopback-datasource-juggler/pull/1778 first.

Issue #3441, opened 2019-07-26
Add `keyFrom` to resolved relation metadata


bajtos:
Add `keyFrom` to HasOne/HasMany resolved metadata. This enables a more generic implementation of inclusion resolvers, because we can use `keyFrom` instead of having to find out what is the name of the primary key (which I found difficult to implement inside inclusion resolvers because of TypeScript limitations).

See #3387 and https://github.com/strongloop/loopback-next/commit/a624d970 in particular

## Acceptance criteria

Implementation + test coverage, please create new unit test files to test metadata resolution in an isolated way.

- [ ] HasMany
- [ ] HasOne

(BelongsTo already has `keyFrom` in the metadata.)

PR #3726, opened 2019-09-13
feat(repository): add keyFrom to resolved relation metadata

nabdelgadir:
Resolves https://github.com/strongloop/loopback-next/issues/3441.

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
+++ b/packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
@@ -1,0 +1,164 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {
+  Entity,
+  HasManyDefinition,
+  ModelDefinition,
+  RelationType,
+} from '../../..';
+import {resolveHasManyMetadata} from '../../../relations/has-many/has-many.helpers';
+
+describe('keyTo and keyFrom with resolveHasManyMetadata', () => {
+  it('resolves metadata using keyTo and keyFrom', () => {
+    const meta = resolveHasManyMetadata(Category.definition.relations[
+      'products'
+    ] as HasManyDefinition);
+
+    expect(meta).to.eql({
+      name: 'products',
+      type: 'hasMany',
+      targetsMany: true,
+      source: Category,
+      keyFrom: 'id',
+      target: () => Product,
+      keyTo: 'categoryId',
+    });
+  });
+
+  it('resolves metadata using keyTo, but not keyFrom', () => {
+    const meta = resolveHasManyMetadata(Category.definition.relations[
+      'items'
+    ] as HasManyDefinition);
+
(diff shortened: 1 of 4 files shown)


packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
```diff
@@ -0,0 +1,164 @@
...
+      keyTo: 'categoryId',
+    });
+  });
+
+  it('resolves metadata using keyTo, but not keyFrom', () => {
+    const meta = resolveHasManyMetadata(Category.definition.relations[
+      'items'
+    ] as HasManyDefinition);
+
+    expect(meta).to.not.have.property('keyFrom');
```
jannyHou: If `keyFrom` is not a "must have" property, shall we define it as optional in the type `hasOneResolvedDefinition`?
The corresponding code: https://github.com/strongloop/loopback-next/pull/3726/files#diff-cbf88c051be861b9a242238ab4c6576cR19

packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
```diff
@@ -0,0 +1,164 @@
...
+      type: 'hasMany',
+      targetsMany: true,
+      source: Category,
+      keyFrom: 'id',
+      target: () => Thing,
+      keyTo: 'categoryId',
+    });
+  });
+
+  it('throws if both keyFrom and keyTo are not provided', async () => {
```
agnes512: Does it thorw error because both`keyFrom` and `keyTo` are not provided and the default name for `keyTo` is not defined in `Category` as well? 
I know we probably cover the default name somewhere else. Can we still add it here? I was a bit confused at the beginning. 
 Besides this, the rest look good to me 👍

packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
agnes512: If it doesn't have `keyFrom`, `keyFrom` won't be included in the metadata. Do we need to note it as optional property in this case?

nabdelgadir: I think it's fine as it is now. Like @agnes512 said, if we don't have it then it just won't include it in the metadata.

packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
nabdelgadir: Ah, when I was testing it before I only added `categoryId` to the class but not to the model definition, but I see that when I add it now, it works. Let me modify the tests 👍

packages/repository/src/__tests__/integration/repositories/resolve-has-many-metadata.integration.ts
nabdelgadir: @jannyHou I realized I made a mistake before, but now when `keyFrom` isn't included, then it gets [inferred](https://github.com/strongloop/loopback-next/pull/3726/files#diff-cbf88c051be861b9a242238ab4c6576cR48) based on the source model's `id` property.

Issue #3448, opened 2019-07-26
Implement InclusionResolver for belongsTo relation


bajtos:
## Acceptance criteria

- [ ] Implement a factory function for creating an inclusion resolver for the given belongsTo relation (see `createBelongsToInclusionResolver` in the spike - #3387).
- [ ] Enhance `BelongsToAccessor` to provide `resolver` property (see the spike).
- [ ] Write integration-level tests in `packages/repository-tests` to verify that resolver works for real databases.
- [ ] Feel free to write unit-level tests using mocked repositories too, as needed.
- [ ] Update documentation (e.g. [Configuring a belongsTo relation](https://loopback.io/doc/en/lb4/BelongsTo-relation.html#configuring-a-belongsto-relation) to explain how to enable or disable inclusion. Include examples showing how to build a query to include related model(s).

Issue #3447, opened 2019-07-26
Implement InclusionResolver for hasMany relation


bajtos:
## Acceptance criteria

- [x] Implement a factory function for creating an inclusion resolver for the given hasMany relation (see `createHasManyInclusionResolver` in the spike - #3387, and [here](https://github.com/strongloop/loopback-next/pull/3387/files#diff-d5bb0fde53b1b79c7048d9b43f3f2e8aL7))
- [x] Enhance `HasManyRepositoryFactory` to provide `resolver` property (see the spike).
- [ ] Write integration-level tests in `packages/repository-tests` to verify that resolver works for real databases.
- [x] Feel free to write unit-level tests using mocked repositories too, as needed.
- [ ] Update documentation (e.g. [Configuring a hasMany relation](https://loopback.io/doc/en/lb4/HasMany-relation.html#configuring-a-hasmany-relation) to explain how to enable or disable inclusion. Include examples showing how to build a query to include related model(s).

PR #3431, opened 2019-07-25
feat(repository-tests): test `replaceById`, verify plain data handling + more

bajtos:
While working on the spike for Inclusion of related models (see #3387), I discovered that `replaceById` does not behave as expected when using MongoDB database and when implementing `fromEntity` repository method, extra care must be taken to not break MongoDB support.

In this pull request, I am adding new tests to verify that `replaceById` works correctly when called from JS/TS code (using a model instance as the data parameter) and via REST API (using a plain data object as the data parameter).

- The first commit fixes a wrong `import` statement. While it's not directly related to my changes, I feel it's not worth opening a new PR for that.
- The second commit introduces a new helper `deleteAllModelsInDefaultDataSource` that leverages the recently introduced juggler API (see https://github.com/strongloop/loopback-datasource-juggler/pull/1759)
- The third commits adds the new tests.

This pull request requires https://github.com/strongloop/loopback-datasource-juggler/pull/1763 to be landed & published first. Before that happens, our test suite fails with console warnings that are detected & rejected by our test suite runner.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3727, opened 2019-09-13
feat(repository): implement inclusion resolver helpers

agnes512:
connects to https://github.com/strongloop/loopback-next/issues/3447, https://github.com/strongloop/loopback-next/issues/3448, https://github.com/strongloop/loopback-next/issues/3449

- Add basic helpers for inclusion resolver
- Add unit tests
- Use self-defined `objectId` in `repository` folder to simulate MondoDB environment to test `isBsonType`, `deduplicate`, `normalizeKey`, `buildLookupMap`.
- Unifies unit test files name

- Refactor tests for `findByForeignKeys`: 
  we were testing it with real repos, which it was problematic. For example, we had [code](https://github.com/strongloop/loopback-next/blob/149f1a37ecccd59a58be6083539a33e9af9e96b0/packages/repository/src/relations/relation.helpers.ts#L37)
```
{inq: fkValues} 
```
  instead of `{inq: [...fkValues]}` before. The wrong version passed in the old unit test that ran on real 
  repositories. But when we were implementing `hasMany inclusion resolver`, we realized that 
  `findByForeignKeys` wasn't correct. 
  Now with stub repositories, we are able to check the queries it sends out, which is difficult to reproduce with real repositories. 

## Checklist


packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,202 @@
...
+    it('normalizeKey', async () => {
+      const id = new ObjectID();
+      expect(normalizeKey(id)).to.eql(id.toHexString());
+    });
+    // the tests below simulate mongodb environment.
+    it('simple normalizeKey + buildLookupMap', async () => {
+      const id = new ObjectID();
+      expect(normalizeKey(id)).to.eql(id.toHexString());
+    });
+    it('normalizeKey + buildLookupMap: returns multiple instances in an array', async () => {
```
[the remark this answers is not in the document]
bajtos: > `import {testdb} from './relations-helpers-fixtures';`

Are you sharing the same datasource instance between multiple tests? That would be the most likely cause of the behavior your are observing. Juggler does not support redefinition of a model with the same name in the same datasource.

Solutions:
- Rework these tests to unit-test style, don't use any repositories.
- Don't share datasource between tests, create a new one in `before` hook. Just call
    ```ts
    new juggler.DataSource({connector: 'memory'})
    ```
- Call `testdb.deleteAllModels()` before defining any new models (in a `before` hook). This is the least preferred option to me. It has its place in acceptance-level tests in `repository-tests` - we want to reuse the datasource to avoid re-opening database connections, because that would slow down the tests. Everywhere else, where we are testing against `memory` connector only, we should create a new datasource in each test suite.

See also https://github.com/strongloop/loopback-next/pull/3431/commits/38b3f906b891520437b61436d12674d711638a15 and https://github.com/strongloop/loopback-datasource-juggler/pull/1759

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,202 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/repository-tests
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect, toJSON} from '@loopback/testlab';
+import {ObjectID} from 'bson';
```
bajtos: 👏

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
```diff
@@ -38,35 +38,35 @@ describe('findByForeignKeys', () => {
   });
 
   it('returns all instances that have the foreign key value', async () => {
-    const pens = await productRepo.create({name: 'pens', categoryId: 1});
-    const pencils = await productRepo.create({name: 'pencils', categoryId: 1});
+    const pen = await productRepo.create({name: 'pen', categoryId: 1});
```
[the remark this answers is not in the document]
agnes512: I think we still need repos to test `findByForeignKey`

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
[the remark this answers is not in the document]
agnes512: yes! there is where the problem lies. Thanks!

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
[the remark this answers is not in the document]
bajtos: Yes and no. You can create a stub repository instance and pass it to `findByForeignKey`. That way you can assert on what parameters were passed to target repository `find` method and how many times it was called.

It also makes sense to test how `findByForeignKey` works with real repositories and datasources, that's what we have `repository-test` for.

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+  reduceAsArray,
+} from '../../../../relations';
+
+describe('unit tests, simulates mongodb env for helpers of ininclusion resolver ', () => {
+  describe('helpers for formating instances', async () => {
+    it('isBsonType', async () => {
+      const id1 = new ObjectID();
+      const id2 = new ObjectID();
+      expect(isBsonType(id1)).to.be.true();
+      expect(isBsonType(id2)).to.be.true();
```
bajtos: I feel it's redundant to call `isBsonType` twice. If it works for the instance created by the first `new ObjectID()` call, then why would it not work for the second and subsequent calls?

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+    it('simple deduplcate + isBsonType', async () => {
```
bajtos:
```suggestion
    it('simple deduplicate + isBsonType', async () => {
```

Also please add an empty line to visually delimit every two test cases - this comment applies to all new code.

```ts
});

it(...
```

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+      const id1 = new ObjectID();
+      const id2 = new ObjectID();
+
+      const result = deduplicate([id1, id2]);
+
+      expect(result).to.deepEqual([id1, id2]);
+    });
+    it('multiple items deduplcate + isBsonType', async () => {
```
bajtos:
```suggestion
    it('multiple items deduplicate + isBsonType', async () => {
```

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  describe('helpers for generating inclusion resolvers', async () => {
+    it('normalizeKey', async () => {
+      const id = new ObjectID();
+      expect(normalizeKey(id)).to.eql(id.toHexString());
+    });
+    // the tests below simulate mongodb environment.
+    it('simple normalizeKey + buildLookupMap', async () => {
+      const id = new ObjectID();
+      expect(normalizeKey(id)).to.eql(id.toHexString());
```
bajtos: This test looks like a duplicate of the previous test on lines 45-49. What am I missing? Can we keep just one of them please?

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+        'categoryId',
+        reduceAsArray,
+      );
+      // expects this map to have String/Product pair
+      const expected = new Map<String, Array<Product>>();
+      const strId = categoryId.toString();
+      expected.set(strId, [pen, pencil]);
+      expect(result).to.eql(expected);
+    });
+    it('normalizeKey + buildLookupMap: return instances in multiple arrays', async () => {
```
bajtos: When I read the test names, I find it difficult to understand what' the difference between this test (return instances in multiple arrays) and the previous one (return instances in multiple arrays). Can you please find more descriptive names?

Also if you have a group of tests sharing the same context (normalizeKey + buildLookupMap), then we usually put them in a `context` block.

```ts
context('normalizeKey + buildLookupMap', () => {
  it('returns multiple instances in an array', async () => {
    // ...
  });
  // ...
});
```

Last but not least - if your test code is synchronous (no `await` calls), then it's better to use a regular function `() => {}` instead of `async () => {}`.

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,139 @@
...
+      expected.set(strId2, [eraser]);
+      expect(result).to.eql(expected);
+    });
+  });
+
+  //** helpers
+  @model()
+  class Product extends Entity {
+    // uses unknown for id type in this test to get rid of type error:
+    // 'objectId' is not compatable to 'string'
```
bajtos: `compatable` - did you mean "compatible" or "comparable"?

I find the error message suspicious, why would the compiler try to compare or convert between ObjectID and a string? Maybe your test data builders like `createPen` need to convert the supplied string value to ObjectID instance?

packages/repository/src/__tests__/unit/repositories/relations-helpers/relations-helpers-fixtures.ts
```diff
@@ -14,36 +14,89 @@ import {
...
+  @property()
+  name: string;
+  @belongsTo(() => Product)
+  productId: number;
+}
+interface ManufacturerRelations {
+  products?: Product;
+}
+
+export class ManufacturerRepository extends DefaultCrudRepository<
```
bajtos: I am not very happy that unit tests are calling relations, can we move such tests to `integration` folder? Sorry for not catching this issue earlier. If you prefer, I am ok to leave this change out of scope of this pull request, to get this building block landed sooner.

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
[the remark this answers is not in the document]
agnes512: It makes more sense to me as it takes a repository as part of input.

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -132,3 +133,137 @@ function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
...
+ * @returns {Array} an array with unique items
+ */
+export function deduplicate<T>(input: T[]): T[] {
+  const uniqArray: T[] = [];
+  if (!input) {
+    return uniqArray;
+  }
+  assert(Array.isArray(input), 'array argument is required');
+
+  const comparableA = input.map(item =>
```
bajtos: `comparableA` - is `A` an abbreviation for "array"? We try to avoid abbreviations in names, they make to code difficult to read.

There is no need to encode the type in the variable name, we have type system to capture that information. [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) is considered as anti-pattern.

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -132,3 +133,137 @@ function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
...
+ * Workaround for MongoDB, where the connector returns ObjectID
+ * values even for properties configured with "type: string".
+ *
+ * @param rawKey
+ */
+export function normalizeKey(rawKey: unknown) {
+  if (
+    typeof rawKey === 'object' &&
+    rawKey &&
+    rawKey.constructor.name === 'ObjectID'
```
bajtos: It makes me wonder - should we perhaps use `isBsonType` helper here and find a way how unify `normalizeKey` logic with the mapper used by `deduplicate`? WDYT?

packages/repository/src/__tests__/unit/repositories/relations-helpers/build-lookup-map.unit.ts
```diff
@@ -0,0 +1,105 @@
...
+      );
+      const expected = new Map<Number, Array<Product>>();
+      expected.set(1, [pen, pencil]);
+      expect(result).to.eql(expected);
+    });
+
+    it('return instances in multiple arrays', async () => {
+      const pen = createPen(1);
+      const pencil = createPencil(1);
+      const eraser = createPEraser(2);
```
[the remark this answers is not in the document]
agnes512: Oops, typo, it shuold be createEraser.

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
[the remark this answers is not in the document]
bajtos: Sure, I see your point.

The idea of unit tests is to test individual components (classes, functions) in isolation and replace dependencies with test-doubles (spies or mocks). Such tests often run much faster than integration tests (no datasource is involved) and allow us to test scenarios that would be difficult to reproduce with real repositories. For example, a unit test can test how many database calls are made by `findByForeignKey`, which is difficult to do with real repositories.

Let's get in touch on Monday and take a look at this together.

packages/repository/src/__tests__/unit/repositories/relations-helpers/relations-helpers-fixtures.ts
agnes512: These are just fixtures for test usage. But I will move them to `belongsTo`.

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
agnes512: Sorry the error message is `type 'ObjectId' is not assignable to parameter of type 'string'`. Got this error when it tries to assign generated ObjectId to string type `categoryId`.

packages/repository/src/relations/relation.helpers.ts
agnes512: yes, reuse code is good :smile:

packages/repository/src/relations/relation.helpers.ts
agnes512: I convert the type back to `T` because I misunderstood the intention of this helper. I thought it would convert the type if it's ObjectId.

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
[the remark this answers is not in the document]
agnes512: I just tried to do something similar to [this](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/unit/repositories/has-many-repository-factory.unit.ts#L100) to generate a stub repo, but it doesn't work (I might do it wrong). 
I don't know if it's possible to traverse instances through a stubbed repo in this case

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+import {belongsTo, Entity, hasMany, model, property} from '../../../..';
+import {
+  isBsonType,
+  deduplicate,
+  normalizeKey,
+  buildLookupMap,
+  reduceAsArray,
+} from '../../../../relations';
+
+describe('unit tests, simulates mongodb env for helpers of ininclusion resolver ', () => {
```
nabdelgadir: typo: `inclusion`

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+  describe('helpers for formating instances', () => {
+    it('isBsonType', () => {
```
nabdelgadir: can you also test if it's not a `bson` type?

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+      const id = new ObjectID();
+      expect(isBsonType(id)).to.be.true();
+    });
+
+    context('deduplicate + isBsonType', () => {
```
nabdelgadir: how are you testing `isBsonType` in these tests?

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+        const result = deduplicate([id3, id1, id1, id3, id2]);
+        expect(result).to.deepEqual([id3, id1, id2]);
+      });
+    });
+  });
+
+  describe('helpers for generating inclusion resolvers', () => {
+    // the tests below simulate mongodb environment.
+    context('normalizeKey + buildLookupMap', () => {
+      it('checks type', async () => {
```
nabdelgadir: I think this title could be more descriptive, e.g. `checks if normalized key is type string` or something else.

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+        const id = new ObjectID();
+        expect(normalizeKey(id)).to.eql(id.toString());
+      });
+
+      it('lookup map with a single key', () => {
```
nabdelgadir: first word should be a verb, e.g. `'creates a lookup map with a single key'`

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+        const categoryId = new ObjectID();
+        const anotherCategoryId = new ObjectID();
+        const pen = createPen({categoryId: categoryId});
+        const pencil = createPencil({categoryId: categoryId});
+        createEraser({categoryId: anotherCategoryId});
```
nabdelgadir: why is this created?

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -132,3 +140,133 @@ function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
...
+ * @param _acc
+ * @param it
+ */
+export function reduceAsSingleItem<T>(_acc: T | undefined, it: T) {
+  return it;
+}
+
+/**
+ * Dedupe an array
+ * @param {Array} input an array
```
nabdelgadir: I think you have to put a `-` after the param's name e.g. `@param {Array} input - an array`. Also `an array` of what?

packages/repository/src/__tests__/unit/repositories/relations-helpers/mongo-related-helpers.unit.ts
agnes512: `normalizeKey` calls `isBsontype`

nabdelgadir: where are you calling `normalizeKey` in these tests?

packages/repository/src/__tests__/unit/repositories/relations-helpers/find-by-foreign-keys.unit.ts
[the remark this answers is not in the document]
agnes512: update the PR description with my understanding of this issue :muscle: thank @bajtos

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -132,3 +133,137 @@ function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
...
+    }
+  }
+  return uniqArray;
+}
+
+/**
+ * checkd of the valus is BsonType (mongodb)
+ * @param value
+ */
+export function isBsonType(value: unknown): value is object {
```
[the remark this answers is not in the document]
agnes512: @jannyHou I update the api documentation of this function. Does it look good to you?

packages/repository/src/relations/relation.helpers.ts
```diff
@@ -132,3 +140,137 @@ function isInclusionAllowed<T extends Entity, Relations extends object = {}>(
...
+ */
+export function deduplicate<T>(input: T[]): T[] {
+  const uniqArray: T[] = [];
+  if (!input) {
+    return uniqArray;
+  }
+  assert(Array.isArray(input), 'array argument is required');
+
+  const comparableArray = input.map(item =>
+    isBsonType(item) ? item.toString() : item,
```
bajtos: Can we replace this lambda function with `normalizeKey`? IIUC, it is the same code.

packages/repository/src/relations/relation.helpers.ts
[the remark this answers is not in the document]
jannyHou: Cool 👍 @agnes512

PR #3729, opened 2019-09-13
feat(eslint-config): enable "no-misused-promises" rule

joeytwiddle:
# Proposal

I suggest we add this useful new rule [`no-misused-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md) to core. It catches a common mistake of forgetting to await a `Promise<boolean>`:

```js
const isAllowed = checkIsAllowed(user);
if (isAllowed) {
  await doSomethingPrivileged();
}

// Mistake: checkIsAllowed() returns a promise, but we forgot to await it!
//
// So the condition will always pass, because `isAllowed: Promise<boolean>`
// is an object so is always truthy.
//
// This is not currently caught by eslint, until the rule is enabled.
```

It also complains if you pass an `async` function when a synchronous function is expected.

# Results

With default settings, the rule found three files of concern in the `loopback-next` project:

```
/Users/joey/src/loopback-next/examples/greeting-app/src/caching-service.ts
   33:31  error  Promise returned in function argument where a void return was expected  @typescript-eslint/no-misused-promises
  119:30  error  Promise returned in function argument where a void return was expected  @typescript-eslint/no-misused-promises

/Users/joey/src/loopback-next/packages/cli/generators/example/index.js
  120:10  error  Expected non-Promise value in a boolean conditional  @typescript-eslint/no-misused-promises

/Users/joey/src/loopback-next/packages/rest/src/__tests__/integration/rest.application.integration.ts
  174:67  error  Promise returned in function argument where a void return was expected  @typescript-eslint/no-misused-promises
```

## rest.application.integration.ts

The last error was easily fixed. (There was just no need for [that context function](https://github.com/strongloop/loopback-next/blob/a6ef8c15c35605455c08bbb1d1ae5f6cefc6ab28/packages/rest/src/__tests__/integration/rest.application.integration.ts#L174) to be async.)

## caching-service.ts

The first two errors indicate an `async` function being used as an easy way to get `await`, but without handling rejections. There is a danger here that an uncaught rejection could crash the node process.

In [the first case](https://github.com/strongloop/loopback-next/blob/a6ef8c15c35605455c08bbb1d1ae5f6cefc6ab28/examples/greeting-app/src/caching-service.ts#L33-L37), perhaps crashing the process is desirable behaviour, since the service failing to restart is a critical problem. We could disable the rule locally for this function, but instead I opted to handle any rejection explicitly with `.catch(err => { console.error(err); process.exit(1); });`

[The second case](https://github.com/strongloop/loopback-next/blob/a6ef8c15c35605455c08bbb1d1ae5f6cefc6ab28/examples/greeting-app/src/caching-service.ts#L119-L121) is just cleaning the cache. If a rejection happens in this case, I think it is non critical, so I just opted to log a warning message, and continue.

@raymondfeng may wish to approve/reject these resolutions.

## cli/generators/example/index.js

The final warning has detected some messy code:

The class [overrides](https://github.com/strongloop/loopback-next/blob/a6ef8c15c35605455c08bbb1d1ae5f6cefc6ab28/packages/cli/generators/example/index.js#L119-L124) the `async end()` function of the [parent](https://github.com/strongloop/loopback-next/blob/a6ef8c15c35605455c08bbb1d1ae5f6cefc6ab28/packages/cli/lib/base-generator.js#L437-L446) class with a _synchronous_ `end()` function. And when it calls `super.end()` it forgets to await!

```js
    // This condition will never be met
    // because super.end() is a promise
    if (!super.end()) return false;
```

Furthermore, the parent method can never actually resolve true, since its return type is `Promise<void>`.

My recommendation is to remove that redundant if, and make the function `async`:

```js
    await super.end();
```

This seems to be how all other calls to `super.end()` are made in the code.

@bajtos may like to confirm this resolution.

## Checklist

- [x] [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next)

### DIFF
--- a/examples/greeting-app/src/caching-service.ts
+++ b/examples/greeting-app/src/caching-service.ts
@@ -30,10 +30,12 @@ export class CachingService {
   ) {
     // Use a view so that we can listen on `refresh` events, which are emitted
     // when the configuration binding is updated in the context.
-    optionsView.on('refresh', async () => {
+    optionsView.on('refresh', () => {
       debug('Restarting the service as configuration changes...');
-      await this.stop();
-      await this.start();
+      this.restart().catch(err => {
+        console.error('Cannot restart the caching service.', err);
+        process.exit(1);
+      });
     });
   }
 
@@ -116,8 +118,8 @@ export class CachingService {
     await this.clear();
     const ttl = await this.getTTL();
     debug('TTL: %d', ttl);
-    this.timer = setInterval(async () => {
-      await this.sweep();
+    this.timer = setInterval(() => {
+      this.sweep().catch(console.warn);
     }, ttl);
   }
 
@@ -131,5 +133,14 @@ export class CachingService {
       clearInterval(this.timer);
     }
     await this.clear();
+  }
+
+  /**
+   * This method may be used to restart the service (and may be triggered by a
+   * 'refresh' event)
+   */
+  async restart(): Promise<void> {
+    await this.stop();
+    await this.start();
   }
 }
--- a/packages/cli/generators/example/index.js
+++ b/packages/cli/generators/example/index.js
@@ -116,8 +116,8 @@ module.exports = class extends BaseGenerator {
     return super.install();
   }
 
-  end() {
-    if (!super.end()) return false;
+  async end() {
+    await super.end();
     this.log();
     this.log(`The example was cloned to ${chalk.green(this.outDir)}.`);
     this.log();
--- a/packages/eslint-config/eslintrc.js
+++ b/packages/eslint-config/eslintrc.js
@@ -146,6 +146,7 @@ module.exports = {
 
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/consistent-type-assertions': 'off',
+    '@typescript-eslint/no-misused-promises': 'error',
   },
 };
 
--- a/packages/rest/src/__tests__/integration/rest.application.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.application.integration.ts
@@ -171,7 +171,7 @@ describe('RestApplication (integration)', () => {
     await client.get(response.header.location).expect(200, 'Hi');
   });
 
-  context('mounting an Express router on a LoopBack application', async () => {
+  context('mounting an Express router on a LoopBack application', () => {
     beforeEach('set up RestApplication', async () => {
       givenApplication();
       await restApp.start();


examples/greeting-app/src/caching-service.ts
```diff
@@ -30,10 +30,15 @@ export class CachingService {
   ) {
     // Use a view so that we can listen on `refresh` events, which are emitted
     // when the configuration binding is updated in the context.
-    optionsView.on('refresh', async () => {
-      debug('Restarting the service as configuration changes...');
-      await this.stop();
-      await this.start();
+    optionsView.on('refresh', () => {
+      (async () => {
```
bajtos: Hmm, this does not look very pretty, I'd like to take a closer look on what other alternatives we have.

packages/cli/generators/example/index.js
```diff
@@ -116,8 +116,8 @@ module.exports = class extends BaseGenerator {
     return super.install();
   }
 
-  end() {
-    if (!super.end()) return false;
+  async end() {
+    await super.end();
```
bajtos: I am surprised `super.end()` is causing linting problems in `example/index.js` only. I'd like to review other generators to see how are they implementing `end` and check if we can unify the code.

examples/greeting-app/src/caching-service.ts
raymondfeng: Does it imply that event listeners cannot be async functions?

raymondfeng:
@joeytwiddle Thank you for the contribution. I like most of fixes in this PR triggered by the new rule.

examples/greeting-app/src/caching-service.ts
joeytwiddle: @raymondfeng No, it should allow for async listeners. If the framework _did_ accept an async function here, the linter would see the expected type of the callback function was async/promise-returning, and it would not complain here.

The danger with using an async function when it's not expected, is that any rejection will go unhandled, causing the node process to abort. (Unless `onunhandledrejection` has been used.)

This linting helps us to avoid accidentally introducing that danger.

joeytwiddle: @batjos I can see four alternatives:

1. What I did above, with an [async wrapper](https://twitter.com/joeytwiddle/status/1142286305805533185).

2. Same solution but different syntax. (Sometimes I use this syntax, when it looks better, but in this case it looks worse!)

        debug('Restarting the service as configuration changes...');
        this.stop().then(
            () => this.start()
        ).catch(err => ...);

3. Disable the rule for this function only. Without `onunhandledrejection` that would likely have the same result (crashing the process) but it would be less explicit.

4. Disable this part of the rule globally. This would open our code up to the aforementioned danger of unexpected crashing. (But it would keep the checking for promises in conditions, which is valuable enough for me.)

       '@typescript-eslint/no-misused-promises': [
          'error',
          {checksVoidReturn: false},
        ],

packages/cli/generators/example/index.js
joeytwiddle: Perhaps I should do like the others, and check `if (this.shouldExit()) {`

Screenshot 2019-09-16 10 38 18

examples/greeting-app/src/caching-service.ts
bajtos: Thank you @joeytwiddle for a detailed explanation!

I am proposing the following solution:
- Modify `CacheService` to implement `restart` method - this can be an `async` function using `await` keywords
- Modify the event listener as follows:

    ```ts
    optionsView.on('refresh', () => {
      debug('Restarting the service as configuration changes...');
      this.restart().catch(err => {
        console.error('Cannot restart the caching service.', err);
        process.exit(1);
      });
    });
    ```

In the future, we should probably modify `this.restart` to detect the case when another restart is triggered before the first restart has been finished. That's out of scope of this PR though.

packages/cli/generators/example/index.js
bajtos: I checked other generators, this change LGTM.

joeytwiddle:
Updated as requested, and rebased.

examples/greeting-app/src/caching-service.ts
joeytwiddle: Hmm yes, async code introduces an entire class of problems and concerns!

Sometimes I find the easiest way to deal with that is to use a queue as a form of locking. In your example, the two restarts would happen one after the other.

In this case it might look something like:

```ts
  private cycleQueue = new TaskQueue();

  // ...

    optionsView.on('refresh', () => {
      this.cycleQueue.queueTask(async () => {
        debug('Restarting the service as configuration changes...');
        await this.restart();
      }).catch(err => {
        console.error('Cannot restart the caching service.', err);
        process.exit(1);
      });
    });
```

(`TaskQueue.ts` is only 38 lines long. Note that it passes the error back out of the queued task to the original caller.)

bajtos:
Landed 🎉 Thank you @joeytwiddle for the contribution! 🙇

Issue #2146, opened 2018-12-12
Customizing error messages


jangeorgiev:
## Feature proposal
I have noticed that currently there are no way of customizing error messages returned from [AJV](https://github.com/epoberezkin/ajv) in LB4.
I have found that there is implementation of handling error message already - [AJV-Errors](https://github.com/epoberezkin/ajv-errors).
I'll be very happy to be able to customize error messages, because some of them are extremely ugly. 

## Example:
[See code here](https://codepen.io/ddesolator/pen/maJgJM)

The code above in case of violation of patter will return
 "should match pattern \"^[a-z]+$\""

shadyanwar:
Related to https://github.com/strongloop/loopback-next/issues/1867

dhmlau:
@jangeorgiev , is #1867 that @shadyanwar mentioned above something you're looking for?  If yes, I'd like to continue the discussion over there instead.  Thanks.

jangeorgiev:
No, it's not related to custom errors that are returned from HttpErrors.UnprocessableEntity().
I talk about customizing error messages of predefined validation rules [see all rules here](https://github.com/epoberezkin/ajv/blob/master/KEYWORDS.md#format)

bajtos:
Based on the issue description and the discussion, I see two incremental steps we can make:

1. Change our validation layer to use [ajv-errors](https://github.com/epoberezkin/ajv-errors) to produce the error message, instead of relying on the built-in ajv messages.

2. Create an extensibility point allowing LB4 applications to further customize the error messages.

3. Probably out of scope of this issue, just something to keep in consideration: localize error messages to a different langauge based on server ENV variables and/or request headers.

Does it look like a reasonable plan to you?

I don't think this will become a priority for us in the next 3-6 months, so it's up to you to contribute these improvements. I am happy to help you along the way if you decide so.

See https://loopback.io/doc/en/contrib/code-contrib.html to get started.

shadyanwar:
> Change our validation layer to use ajv-errors to produce the error message, instead of relying on the built-in ajv messages.

@bajtos I hope to find time next weekend to work on this. It would be nice if you tell me where to start.

bajtos:
@shadyanwar that's great to hear!

Here is the code converting AJV validation errors to an HttpError instance:

https://github.com/strongloop/loopback-next/blob/27b6accace55761d8db15442e5ca7705bab7d9fc/packages/rest/src/validation/request-body.validator.ts#L128-L137

[...]

PR #3746, opened 2019-09-16
feat(rest): allow AJV error messages customization

AaronJan:
I'm currently doing some localization works, it should be easy by just using [`ajv-i18n`](https://github.com/epoberezkin/ajv-i18n):

```typescript
const localize_zh = require('ajv-i18n/localize/zh');

// option `i18n` is required for this package to work
var ajv = Ajv({ allErrors: true });
var validate = ajv.compile(schema);
var valid = validate(data);

if (!valid) {
    localize_zh(validate.errors);
    console.log(ajv.errorsText(validate.errors, { separator: '\n' }));
}
```

But I could find a way to do that.

After some searching & reading, I found #2146, I really need this feature, so I'm submitting a minimal change.

Advice needed (we can discus in #2146 ), thank you!

> See also #2146 

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -246,6 +246,64 @@ describe('Validation at REST level', () => {
     });
   });
 
+  context(
+    'with request body validation options - {ajvErrorTransformer: [Function]}',
+    () => {
+      class ProductController {
+        @post('/products')
+        async create(
+          @requestBody({required: true}) data: Product,
+        ): Promise<Product> {
+          return new Product(data);
+        }
+      }
+
+      before(() =>
+        givenAnAppAndAClient(ProductController, {
+          nullable: false,
+          compiledSchemaCache: new WeakMap(),
+          $data: true,
+          ajvErrorTransformer: errors => {
+            return errors.map(e => ({
+              ...e,
+              message: `(translated) ${e.message}`,
+            }));
+          },
+        }),
+      );
+      after(() => app.stop());
+
+      it('transforms error messages provided by AJV', async () => {
+        const DATA = {
+          name: 'iPhone',
+          description: 'iPhone',
+        };
+        const res = await client
+          .post('/products')
+          .send(DATA)
+          .expect(422);
+
+        expect(res.body).to.eql({
+          error: {
+            statusCode: 422,
+            name: 'UnprocessableEntityError',
+            message:
+              'The request body is invalid. See error object `details` property for more info.',
+            code: 'VALIDATION_FAILED',
+            details: [
+              {
+                path: '',
+                code: 'required',
+                message: "(translated) should have required property 'price'",
+                info: {missingProperty: 'price'},
+              },
+            ],
+          },
+        });
+      });
+    },
+  );
+
   // A request body schema can be provided explicitly by the user
   // as an inlined content[type].schema property.
   context('for fully-specified request body', () => {
--- a/packages/rest/src/types.ts
+++ b/packages/rest/src/types.ts
@@ -107,6 +107,11 @@ export interface RequestBodyValidationOptions extends ajv.Options {
    * - `string[]`: Add an array of keywords from `ajv-keywords`
    */
   ajvKeywords?: true | string[];
+  /**
+   * A function that transform the `ErrorObject`s reported by AJV.
+   * This could be used for error messages customization, localization, etc.
+   */
+  ajvErrorTransformer?: (errors: ajv.ErrorObject[]) => ajv.ErrorObject[];
 }
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
--- a/packages/rest/src/validation/request-body.validator.ts
+++ b/packages/rest/src/validation/request-body.validator.ts
@@ -137,7 +137,7 @@ function validateValueAgainstSchema(
     return;
   }
 
-  const validationErrors = validate.errors;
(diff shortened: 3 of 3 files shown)


packages/rest/src/types.ts
```diff
@@ -107,6 +107,10 @@ export interface RequestBodyValidationOptions extends ajv.Options {
    * - `string[]`: Add an array of keywords from `ajv-keywords`
    */
   ajvKeywords?: true | string[];
+  /**
+   * Enable error message customizing.
+   */
+  ajvTranslate?: (errors: ajv.ErrorObject[]) => ajv.ErrorObject[];
```
raymondfeng: Should we allow asynchronous translation here?

AaronJan: @raymondfeng Yes, it would be more unified, but I think I have to touch a couple more files to do that.

Is that OK?

bajtos: So far, we support synchronous validation only, therefore `validateValueAgainstSchema` is sync too. I feel introducing async translation would require too many changes, I am not convinced the (speculative) benefits are worth the cost.

Let's start with sync translations for the first version, we can look into async translations later, if there is need.

bajtos: I like the idea of allowing app developers to transform the errors reported by AJV. I am little bit reluctant to limit this new config option to translations only, can find a more generic name please? For example: `ajvErrorTransformer`.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -246,6 +246,64 @@ describe('Validation at REST level', () => {
...
+            return errors.map(e => ({
+              ...e,
+              message: `(translated) ${e.message}`,
+            }));
+          },
+        }),
+      );
+      after(() => app.stop());
+
+      it('rejects requests with translated error message', async () => {
```
bajtos: I read this test name as "the framework reject requests when the client provided a translated error message". Let's fine a better one please. For examples:

```suggestion
      it('transforms error messages provided by AJV', async () => {
```

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -137,7 +137,11 @@ function validateValueAgainstSchema(
...
 
-  const validationErrors = validate.errors;
+  let validationErrors = validate.errors as AJV.ErrorObject[];
+
+  if (options.ajvTranslate) {
+    validationErrors = options.ajvTranslate(validationErrors);
+  }
 
   /* istanbul ignore if */
   if (debug.enabled) {
```
bajtos: The next line prints a debug log with `util.inspect(validationErrors)`.

In your current proposal, these debug logs will contain translated error message. This is nice when there is a single language involved only and the developer of the application understand the language used for messages returned to clients.

It does not work well for applications supporting multiple languages - for example, if I was developing an app available in English, Czech and Chinese, then I won't be able to understand debug logs translated to Chinese. Personally, I'd prefer to see all debug logs in English using the original message reported by AJV, before any translation or other transformations is applied.

Please run the transformation after the debug log is printed.

If you feel it's useful to see translated messages in the debug log too, then add a new debug log that's printed after the AJV errors were modified.

```ts
if (options.ajvTranslate) {
  validationErrors = options.ajvTranslate(validationErrors);
  /* istanbul ignore if */
  if (debug.enabled) {
    debug(...);
  }
}
```

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -137,7 +137,11 @@ function validateValueAgainstSchema(
     return;
   }
 
-  const validationErrors = validate.errors;
+  let validationErrors = validate.errors as AJV.ErrorObject[];
+
+  if (options.ajvTranslate) {
```
bajtos: Let's be little bit more defensive and check that `ajvTranslate` is actually a function. We don't want to see `TypeError: options.ajvTranslate is not a function`.

```suggestion
  if (typeof options.ajvTranslate === 'function') {
```

packages/rest/src/types.ts
AaronJan: I was uncertain about using verb or noun, I think `ajvErrorTransformer ` is good. 👍

AaronJan: I was worried about the probably low ROI of this async change, but I don't like to support both sync & async transformation in the future either.

But I think minimizing changes is better for now.

AaronJan:
I made a new commit according @bajtos 's suggestions.

Let me know if there is anything can be improved or rebasing. 😄

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -246,6 +246,64 @@ describe('Validation at REST level', () => {
     });
   });
 
+  context(
+    'with request body validation options - {ajvTranslate: [Function]}',
```
raymondfeng: `ajvTranslate` => `ajvErrorTransformer`

bajtos:
Could you @AaronJan rebase your patch on top of the latest master and squash all commits into a single one please? See [Rebasing](https://loopback.io/doc/en/lb4/submitting_a_pr.html#6-rebasing) and [Final rebase and squashing of commits](https://loopback.io/doc/en/lb4/submitting_a_pr.html#10-final-rebase-and-squashing-of-commits) in our guide. See also our [Commit message guidelines](https://loopback.io/doc/en/lb4/code-contrib-lb4.html#commit-message-guidelines).

AaronJan:
Done. @bajtos

PR #3771, opened 2019-09-19
Spike more flexible user profile

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/2246
Reviewers can read [spike-user-profile.md](https://github.com/strongloop/loopback-next/pull/3771/files#diff-f35e305427c53a74556505e9920f786e) to start.

## Checklist

### DIFF
--- a/extensions/authentication-passport/spike-user-profile.md
+++ b/extensions/authentication-passport/spike-user-profile.md
@@ -1,0 +1,22 @@
+# Spike for More Flexible UserProfile
+
+connect to story https://github.com/strongloop/loopback-next/issues/2246
+
+I picked the `authentication-passport` module to start the spike for more flexible user profile because compared with the custom authentication strategies, users have less control to the returned user when using the passport adapter. I believe if we could find a solution for the passport based strategies, applying similar approach to a custom strategy would be easy.
+
+# Solution
+
+A converter function is introduced to be passed into the `StrategyAdapter`'s constructor. It takes in a custom user, converts it to a user profile described by `UserProfile` then returns it.
+
+# Example
+
+See the corresponding change made in file 'authentication-with-passport-strategy-adapter.acceptance.ts':
+
+- Type `UserProfileInDB` is defined to describe the custom user. In a real application it should be a custom User model.
+- Define a converter function `converter` that turns an `UserProfileInDb` instance into a user profile. It's provided in the constructor when create the adapter.
+- The converter is invoked in the strategy's `authentication()` function to make sure it returns a user profile in type `UserProfile`
+- If the strategy is returned in a provider, you can inject the converter.
+
+# To Be Done
+
+Add a field `additionalProperties` in type `AnyObject` in the `UserProfile` interface to allow people add additional fields as the minimum set of identification info. 
--- a/extensions/authentication-passport/spike-user-profile.md
+++ b/extensions/authentication-passport/spike-user-profile.md
@@ -2,17 +2,17 @@
 
 connect to story https://github.com/strongloop/loopback-next/issues/2246
 
-I picked the `authentication-passport` module to start the spike for more
-flexible user profile because compared with the custom authentication
-strategies, users have less control to the returned user when using the passport
-adapter. I believe if we could find a solution for the passport based
-strategies, applying similar approach to a custom strategy would be easy.
+I picked the `authentication-passport` module to start the spike for a more
+flexible user profile because, compared with custom authentication strategies,
+users have less control over the returned user when using the passport adapter.
+I believe that if we could find a solution for the passport based strategies,
+applying a similar approach to a custom strategy would be easy.
 
 # Solution
 
-A converter function is introduced to be passed into the `StrategyAdapter`'s
-constructor. It takes in a custom user, converts it to a user profile described
-by `UserProfile` then returns it.
+A converter function is passed into the `StrategyAdapter`'s constructor. It
+takes in a custom user, converts it to a user profile described by
+`UserProfile`, and then returns it.
 
 # Example
 
@@ -25,8 +25,11 @@ See the corresponding change made in file
   instance into a user profile. It's provided in the constructor when creating
   the adapter.
 - The converter is invoked in the strategy's `authentication()` function to make
-  sure it returns a user profile in type `UserProfile`
+  sure it returns a user profile of type `UserProfile`
 - If the strategy is returned in a provider, you can inject the factory.
+  - I added an example in the
+    [create passport based strategy with provider](./README.md#with-provider)
+    section.
 
 # Token Based Authentication
 
@@ -35,8 +38,10 @@ profile factory in a token based authentication. Here is the scenario:
 
 - The app has a user model called `MyUser`, which has
 
-  - a property defined in `UserProfile` with same type
-  - a property defined in `UserProfile` with a different type
+  - a property which is also defined in `UserProfile` in the same type. E.g.
+    They both have `email` as string.
+  - a property which is also defined in `UserProfile` but in a different type.
+    E.g. `MyUser` has `id` as number while `UserProfile` has it as string.
   - a required property not defined in `UserProfile`
   - an optional property not defined in `UserProfile`
 
--- a/extensions/authentication-passport/spike-user-profile.md
+++ b/extensions/authentication-passport/spike-user-profile.md
@@ -2,21 +2,77 @@
 
 connect to story https://github.com/strongloop/loopback-next/issues/2246
 
-I picked the `authentication-passport` module to start the spike for more flexible user profile because compared with the custom authentication strategies, users have less control to the returned user when using the passport adapter. I believe if we could find a solution for the passport based strategies, applying similar approach to a custom strategy would be easy.
+I picked the `authentication-passport` module to start the spike for more
+flexible user profile because compared with the custom authentication
+strategies, users have less control to the returned user when using the passport
+adapter. I believe if we could find a solution for the passport based
+strategies, applying similar approach to a custom strategy would be easy.
 
 # Solution
 
-A converter function is introduced to be passed into the `StrategyAdapter`'s constructor. It takes in a custom user, converts it to a user profile described by `UserProfile` then returns it.
+A converter function is introduced to be passed into the `StrategyAdapter`'s
+constructor. It takes in a custom user, converts it to a user profile described
+by `UserProfile` then returns it.
 
 # Example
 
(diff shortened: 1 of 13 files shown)


packages/authentication/src/types.ts
```diff
@@ -16,6 +16,10 @@ export interface AuthenticateFn {
   (request: Request): Promise<UserProfile | undefined>;
 }
 
+export interface convertUserToUserProfileFn<U> {
```
bajtos: We use PascalCase for type names and usually use a noun, not a verb.

```suggestion
export interface UserToUserProfileConverterFn<U> {
```

extensions/authentication-passport/src/strategy-adapter.ts
```diff
@@ -44,8 +50,16 @@ export class StrategyAdapter implements AuthenticationStrategy {
       const strategy = Object.create(this.strategy);
 
       // add success state handler to strategy instance
-      strategy.success = function(user: UserProfile) {
-        resolve(user);
+      // as a generic adapter, it is agnostic of the type of
+      // the custom user, so loose the type restriction here
+      // to be `any`
+      strategy.success = function(user: any) {
```
bajtos: Can we use `unknown` instead of `any` please?

extensions/authentication-passport/src/strategy-adapter.ts
```diff
@@ -44,8 +50,16 @@ export class StrategyAdapter implements AuthenticationStrategy {
...
+        if (self.userConverter) {
```
bajtos: Let's supply a default converter function in the constructor, so that we don't have to check if `self.userConverter` is set.

```ts
// In the constructor
constructor(
    private readonly strategy: Strategy, 
    readonly name: string,
    private userConverter: convertUserToUserProfileFn<U> = u => u
  ) {}

// usage
const userProfile = self.userConverter(user);
resolve(userProfile);
```

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-adapter.acceptance.ts
```diff
@@ -98,10 +98,18 @@ describe('Basic Authentication', () => {
   function verify(username: string, password: string, cb: Function) {
     users.find(username, password, cb);
   }
+
+  function converter(user: UserProfileInDB): UserProfile {
+    let userProfile = Object.assign({}, user, {[securityId]: user.id});
+    delete userProfile.id;
```
bajtos: What if `securityId` is the string `id`? I think the following solution would be safer:

```ts
const userProfile = Object.assign({}, user);
const id = userProfile.id;
delete userProfile.id;
userProfile.securityId = id;
```

Is it actually necessary to remove the id property? Cannot we preserve it so that clients can use `id` instead of `securityId` if they like to?

jannyHou: @bajtos The converter function is not a generic one, it's defined for the test case only(for the mock db attached to the user repository), that's why we know the id field is called `id` and is different than symbol  `securityId`. 

The purpose of having a converter function is allowing developers define their custom conversion according to their User model.

packages/authentication/src/types.ts
```diff
@@ -35,6 +35,10 @@ export interface AuthenticateFn {
   (request: Request): Promise<UserProfile | undefined>;
 }
 
+export interface UserToUserProfileConverterFn<U> {
```
raymondfeng: Maybe name it as `UserProfileBuilder`?

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -0,0 +1,22 @@
...
+# Solution
+
+A converter function is introduced to be passed into the `StrategyAdapter`'s constructor. It takes in a custom user, converts it to a user profile described by `UserProfile` then returns it.
+
+# Example
+
+See the corresponding change made in file 'authentication-with-passport-strategy-adapter.acceptance.ts':
+
+- Type `UserProfileInDB` is defined to describe the custom user. In a real application it should be a custom User model.
+- Define a converter function `converter` that turns an `UserProfileInDb` instance into a user profile. It's provided in the constructor when create the adapter.
```
raymondfeng: UserProfileInDB?

packages/authentication/src/types.ts
bajtos: +1 to find a better name.

Personally, I would use `UserProfileFactory` (see [Factory pattern](https://en.wikipedia.org/wiki/Factory_method_pattern)).

The name `UserProfileBuilder` suggest a well-known [Builder pattern](https://en.wikipedia.org/wiki/Builder_pattern), which we are NOT following in this API.

extensions/authentication-passport/src/strategy-adapter.ts
```diff
@@ -3,27 +3,36 @@
...
   /**
    * @param strategy instance of a class which implements a passport-strategy;
    * @description http://passportjs.org/
    */
-  constructor(private readonly strategy: Strategy, readonly name: string) {}
+  constructor(
+    private readonly strategy: Strategy,
+    readonly name: string,
+    // The default converter returns an user as user profile
+    private userConverter: UserToUserProfileConverterFn<U> = (u: unknown) => {
```
raymondfeng: This should be optional.

jannyHou: @raymondfeng a default converter is provided (return the `user` as it is) therefore the converter here is not marked as optional :)

See discussion in https://github.com/strongloop/loopback-next/pull/3771#discussion_r326611630

packages/authentication/src/__tests__/fixtures/user-profile-spike/user.repository.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class MyUserRepository {
+  constructor(readonly list: {[key: string]: MyUser}) {}
+  find(name: string): MyUser | undefined {
+    const found = Object.keys(this.list).find(k => this.list[k].name === name);
+    return found ? this.list[found] : undefined;
+  }
+}
+
+export const SAMPLE_USER_MIM_SET = {
```
emonddr: > SAMPLE_USER_MIM_SET 

What's a MIM set?  Did you mean MIN as in minimum?

jannyHou: oh good catch on the typo

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
 
 connect to story https://github.com/strongloop/loopback-next/issues/2246
 
-I picked the `authentication-passport` module to start the spike for more flexible user profile because compared with the custom authentication strategies, users have less control to the returned user when using the passport adapter. I believe if we could find a solution for the passport based strategies, applying similar approach to a custom strategy would be easy.
+I picked the `authentication-passport` module to start the spike for more
```
emonddr: I picked the `authentication-passport` module to start the spike for `a` more flexible user profile because`,` compared with ~the~ custom authentication
strategies, users have less control ~to~ `over` the returned user when using the passport
adapter. I believe `that` if we could find a solution for the passport based
strategies, applying `a` similar approach to a custom strategy would be easy.

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
...
+flexible user profile because compared with the custom authentication
+strategies, users have less control to the returned user when using the passport
+adapter. I believe if we could find a solution for the passport based
+strategies, applying similar approach to a custom strategy would be easy.
 
 # Solution
 
-A converter function is introduced to be passed into the `StrategyAdapter`'s constructor. It takes in a custom user, converts it to a user profile described by `UserProfile` then returns it.
+A converter function is introduced to be passed into the `StrategyAdapter`'s
```
emonddr: > A converter function is introduced to be passed into the `StrategyAdapter`'s

A converter function is ~introduced to be~ passed into the `StrategyAdapter`'s

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
...
+constructor. It takes in a custom user, converts it to a user profile described
```
emonddr: It takes in a custom user, converts it to a user profile described by `UserProfile` `, and` then returns it.

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
...
-- Define a converter function `converter` that turns an `UserProfileInDb` instance into a user profile. It's provided in the constructor when create the adapter.
-- The converter is invoked in the strategy's `authentication()` function to make sure it returns a user profile in type `UserProfile`
-- If the strategy is returned in a provider, you can inject the converter.
+- Type `UserProfileInDB` is defined to describe the custom user. In a real
+  application it should be a custom User model.
+- Define a factory function `userProfileFactory` that turns an `UserProfileInDB`
+  instance into a user profile. It's provided in the constructor when creating
+  the adapter.
+- The converter is invoked in the strategy's `authentication()` function to make
+  sure it returns a user profile in type `UserProfile`
```
emonddr: sure it returns a user profile ~in~ `of`  type `UserProfile`

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
...
+- If the strategy is returned in a provider, you can inject the factory.
```
emonddr: Can you provide a tiny example of what you mean?

extensions/authentication-passport/spike-user-profile.md
```diff
@@ -2,21 +2,77 @@
...
+# Token Based Authentication
 
-Add a field `additionalProperties` in type `AnyObject` in the `UserProfile` interface to allow people add additional fields as the minimum set of identification info. 
\ No newline at end of file
+I created a test case called "test user profile factory" to apply the user
+profile factory in a token based authentication. Here is the scenario:
+
+- The app has a user model called `MyUser`, which has
+
+  - a property defined in `UserProfile` with same type
```
emonddr: same type as what?

jannyHou:
I am closing this pull request as the spike is done. Creating follow-up stories:

- https://github.com/strongloop/loopback-next/issues/3845
- https://github.com/strongloop/loopback-next/issues/3846

Issue #2825, opened 2019-05-02
Database migrations for mounted LoopBack 3 Models


nabdelgadir:
## Description / Steps to reproduce / Feature proposal

When mounting a LoopBack 3 application in a LoopBack 4 project using `Lb3AppBooterComponent`, running `npm run migrate` will not migrate the LoopBack 3 application's models.

## Current Behaviour

By running `npm run migrate`, LoopBack 3 models are not migrated.

## Expected Behaviour

By running `npm run migrate`, LoopBack 3 models are migrated in addition to LoopBack 4 models.

## Acceptance Criteria
- [ ] Use a non-in-memory datasource (e.g. MySQL, MongoDB) and run `npm run migrate` on a mounted application that has both LoopBack 3 and LoopBack 4 models. Make sure both LoopBack 3 and LoopBack 4 models are migrated properly.
   - [ ] currently it discovers all LB4 datasources in global LB4 context. since the LB3 app is  bound to the context as a whole, you can access all the datasource objects that belong to it by doing `lb3App.datasources`

dhmlau:
@nabdelgadir, could you please add acceptance criteria? Thanks!

bajtos:
Cross-posting from slack.

Take a look at how migrate calls app.migrateSchema and how is the latter implemented in https://github.com/strongloop/loopback-next/blob/7bf6f12f1da721a78071904fa02252577f4d1cba/packages/repository/src/mixins/repository.mixin.ts#L194-L219:

It's iterating over all known datasources and calls `ds.automigrate` or `ds.autoupdate` (all known datasources = all datasources bound to the LB4 app).

To migrate LB3 models, we need to execute automigrate/autoupdate also for datasources in the LB3 app. 

I see two options how to approach this problem:
1. Modify `Lb3AppBooter.load` to bind all LB3 datasources into LB4 app context (via `app.bind`, I don't think we can rely on `app.dataSource` being added via `RepositoryMixin`). That way `migrateSchema` will find both LB4 and LB3 dataSource and migrate them all. The difficult part is how to choose the binding key for LB3 datasources to avoid clashes with datasources defined from LB4. Perhaps it's enough to add a prefix, e.g. `db` -> `datasources.lb3-db`? IDK 🤷‍♂ 
[...]

PR #3779, opened 2019-09-20
fix: migrate LB3 models mounted on LB4 app

hacksparrow:
Migrates LB4 models mounted on LB4 an app.

Fixes https://github.com/strongloop/loopback-next/issues/2825.

## Checklist

### DIFF
--- a/packages/booter-lb3app/fixtures/app-with-model.js
+++ b/packages/booter-lb3app/fixtures/app-with-model.js
@@ -1,0 +1,18 @@
+'use strict';
+
+const loopback = require('loopback');
+const boot = require('loopback-boot');
+
+const app = (module.exports = loopback());
+
+app.dataSource('memory', {connector: 'memory'});
+const Color = app.registry.createModel('Color', {name: String});
+app.model(Color, {dataSource: 'memory'});
+
+app.get('/hello', (req, res) => {
+  res.send('hello');
+});
+
+boot(app, __dirname, function(err) {
+  if (err) throw err;
+});
--- a/packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
+++ b/packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {OperationObject, OpenApiSpec} from '@loopback/rest';
+import {OpenApiSpec, OperationObject} from '@loopback/rest';
 import {Client, expect} from '@loopback/testlab';
 import * as _ from 'lodash';
 import {
@@ -213,6 +213,23 @@ describe('booter-lb3app', () => {
       const spec = app.restServer.getApiSpec();
       const createOp: OperationObject = spec.paths['/api/CoffeeShops'].post;
       expect(createOp.summary).to.eql('just a very simple modification');
+    });
+  });
+
+  context('binding LoopBack 3 datasources', () => {
+    before(async () => {
+      ({app, client} = await setupApplication({
+        lb3app: {path: '../fixtures/app-with-model'},
+      }));
+    });
+
+    it('binds datasource to the context', async () => {
+      const expected = require('../../../fixtures/app-with-model').dataSources
+        .memory;
+      const dsBindings = app.findByTag('datasource');
+      const key = dsBindings[0].key;
+      const ds = await app.get(key);
+      expect(ds).to.eql(expected);
     });
   });
 });
--- a/packages/booter-lb3app/src/lb3app.booter.ts
+++ b/packages/booter-lb3app/src/lb3app.booter.ts
@@ -62,6 +62,20 @@ export class Lb3AppBooter implements Booter {
       this.mountRoutesOnly(lb3App, spec);
     }
 
+    const dataSources = lb3App.dataSources;
+    if (dataSources) {
+      const visited: unknown[] = [];
+      Object.keys(dataSources).forEach(key => {
+        const ds = dataSources[key];
+        if (visited.includes(ds)) return;
+        visited.push(ds);
+        this.app
+          .bind(`datasources.lb3-${key}`)
+          .to(ds)
+          .tag('datasource');
+      });
+    }
+
     // TODO(bajtos) Listen for the following events to update the OpenAPI spec:
     // - modelRemoted
     // - modelDeleted
@@ -137,4 +151,5 @@ export interface Lb3AppBooterOptions {
 
 interface Lb3Application extends ExpressApplication {
   handler(name: 'rest'): ExpressRequestHandler;
+  dataSources?: {[name: string]: unknown};
 }
--- a/docs/site/Migrating-from-LoopBack-3.md
+++ b/docs/site/Migrating-from-LoopBack-3.md
@@ -38,6 +38,10 @@ application to OpenAPI v3 and adds it to the LoopBack 4 project's OpenAPI v3
 spec.
 " %}
 
+Scaffolded LoopBack 4 apps come with a `migrate` npm script, which can be run
+using `npm run migrate`. This command will migrate all the LoopBack 3 models
+along with the LoopBack 4 models.
+
 ### Options
 
 [`Lb3AppBooterComponent`](https://loopback.io/doc/en/lb4/apidocs.booter-lb3app.lb3appbootercomponent.html)
--- a/examples/lb3-application/src/migrate.ts
+++ b/examples/lb3-application/src/migrate.ts
@@ -11,10 +11,6 @@ export async function migrate(args: string[]) {
 
   const app = new CoffeeShopApplication();
   await app.boot();
-  console.warn(
-    'Skipping migration of models from LB3 application (not supported yet).',
-  );
-  console.warn('See https://github.com/strongloop/loopback-next/issues/2825');
   await app.migrateSchema({existingSchema});
 
   // Connectors usually keep a pool of opened connections,


packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -62,6 +63,22 @@ export class Lb3AppBooter implements Booter {
       this.mountRoutesOnly(lb3App, spec);
     }
 
+    const dataSources = lb3App.dataSources;
+    if (dataSources) {
+      const processedDs: string[] = [];
+      Object.keys(dataSources).forEach(key => {
+        key = key.toLowerCase();
+        if (!processedDs.includes(key)) {
```
bajtos: Let's use reverse logic please.

```ts
if (processedDs.includes(key)) continue;
```

Is it safe to use `key.toLowerCase()` to decide whether two entries are unique? I think it would be much safer to compare the actual datasource instances (by reference).

```ts
Object.values(dataSources).forEach(ds => {
  if (visited.includes(ds)) return;
  visited.push(ds);

  // ...
});
```

packages/booter-lb3app/fixtures/app-with-model.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const loopback = require('loopback');
+const boot = require('loopback-boot');
+
+const app = (module.exports = loopback());
+
+app.dataSource('memory', {connector: 'memory'});
+const Color = app.registry.createModel('Color', {name: String});
+app.model(Color, {dataSource: 'memory'});
```
agnes512: Can we also add a non-in-memory datasource test case?

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -62,6 +63,22 @@ export class Lb3AppBooter implements Booter {
...
+          const ds = dataSources[key];
+          processedDs.push(ds.name);
+          this.app
+            .bind(`datasources.lb3-${ds.name}`)
```
bajtos: I would like to discuss the binding naming convention with @raymondfeng and possibly with the wider community @strongloop/loopback-next .

Aspects to consider:
- There should be very low probability of a naming conflict between a key generated for a LB3 datasource and a key created by LB4 API `app.dataSource`.
- The key should make it obvious what's the datasource name and that this datasource is coming from a LB3 app.

I think the current proposal `datasources.lb3-${ds.name}` is reasonable 👍 

Few other options to consider:
- `lb3-datasources.${ds.name}` (e.g. `lb3-datasources.db`)
- `datasources.\$lb3\$${ds.name}` (e.g. `datasources.$lb3$db`)

packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
```diff
@@ -215,4 +215,17 @@ describe('booter-lb3app', () => {
...
+  context('binding LoopBack 3 datasources', () => {
+    before(async () => {
+      ({app, client} = await setupApplication({
+        lb3app: {path: '../fixtures/app-with-model'},
+      }));
+    });
+
+    it('binds datasource to the context', async () => {
+      const dsBindings = app.findByTag('datasource');
+      expect(dsBindings[0].key).to.equal('datasources.lb3-memory');
```
bajtos: Can you verify that the binding value is equal to lb3app's `memory` datasource?

I think you can obtain that datasource instance by loading the app via `require()`.

```ts
const expected = require('../fixtures/app-with-model').dataSources.memory;
// ...
```

packages/booter-lb3app/fixtures/app-with-model.js
hacksparrow: AFAIK, it won't add value to the test, it will rather add a dependency. We just want to check if a connector is added to the context at all.

Thoughts @strongloop/loopback-next?

hacksparrow:
> Did you verify this scenario manually?

Yes, in a standalone app.

> Is there any reasonable way how to automate such test?

Will work on this next and propose something.

dhmlau:
@hacksparrow, there's a decrease in the test coverage.  Could you please look into it? Thanks.

hacksparrow:
@dhmlau the decrease is not in the related package (`booter-lb3app`). What do we do?

@raymondfeng @bajtos

raymondfeng:
@hacksparrow I don't see obvious cause of coverage to worry about in this case.

packages/booter-lb3app/src/lb3app.booter.ts
jannyHou: I am good with either 
`datasources.lb3-${ds.name}` (pro: the prefix is consistent with other datasources)
or 
`lb3-datasources.${ds.name}` (pro: we can apply the same naming convention for other lb3 artifacts like models)

packages/booter-lb3app/fixtures/app-with-model.js
agnes512: What if we add the test in `repository-test` folder? Will it be able to test if the model is migrated to mongoDB or MySQL? It's part of the criteria. 
I am ok to skip the test if @strongloop/loopback-maintainers think it's fine.

jannyHou: Is this PR aimed to fix all issues in #2825? IIUC there will be another PR to make sure the `npm run migrate` migrates LB3 models and that is where we can test different datasources.

I feel the test here is just to make sure lb3 datasources are loaded properly and bound to the lb4 app, so it doesn't matter which database it uses :)

hacksparrow:
@bajtos I have been considering some ways for verifying datasources contributed by LB3 are picked up, but they all involve `app.findByTag('datasource')`, which is already covered by the unit test in the PR.

An approach which is more appropriate to the reported issue ([Database migrations for mounted LoopBack 3 Models](https://github.com/strongloop/loopback-next/issues/2825)) would be an integration test case, which verifies the state of the models in the database with no knowledge of the datasources used. But this will require a more elaborate setup (will need an SQL database) and dependencies. Is it worth it? 

In my opinion, the unit test should suffice because the implementation of migration is done by enumerating the datasources bound to the app. If we can verify that a datasource is bound to the app, we can be sure all models attached to it will be migrated if the corresponding connector supports migration.

bajtos:
> An approach which is more appropriate to the reported issue (Database migrations for mounted LoopBack 3 Models) would be an integration test case, which verifies the state of the models in the database with no knowledge of the datasources used. But this will require a more elaborate setup (will need an SQL database) and dependencies. Is it worth it?
>
> In my opinion, the unit test should suffice because the implementation of migration is done by enumerating the datasources bound to the app. If we can verify that a datasource is bound to the app, we can be sure all models attached to it will be migrated if the corresponding connector supports migration.

Fair enough. We can introduce an integration or end-to-end test later in the future, if needed. :+1:

For posterity, I think it should be possible to write a test that does not require the real database, the test needs to replace the real `automigrate` or `autoupdate` method on a LB3 datasource with a [sinon](https://sinonjs.org) stub, and then verify the stub was called from `app.migrateSchema`.

packages/booter-lb3app/fixtures/app-with-model.js
bajtos: Here is my thinking:

- This pull request is the complete implementation of #2825
- We trust the LB3 runtime to correctly migrate LB3 models when `dataSource.automigrate` is called, this does not need to be tested in LB4 again.
- What we need to ensure and test: `npm run migrate` will call `automigrate`/`autoupdate` on datasources defined in the LB3 app.
- Yaapa run a manual test to verify that `npm run migrate` migrates LB3 models.

> I feel the test here is just to make sure lb3 datasources are loaded properly and bound to the lb4 app, so it doesn't matter which database it uses :)

+1

> What if we add the test in `repository-test` folder?

These tests are verifying `@loopback/repository` functionality. In this pull request, we want to test integration between LB3 and LB4, I feel that does not belong to `@loopback/repository-test`.

We can add a new package to the top-level `acceptance` directory, alongside `repository-mysql` and `repository-mongodb`. However, I feel the cost of running such acceptance test is not worth the benefits. It should be the responsibility of LB3 to test and verify that automigration works as expected.

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -5,6 +5,7 @@
 
 import {BootBindings, Booter} from '@loopback/boot';
 import {CoreBindings, inject} from '@loopback/core';
+import {AnyObject, DataSource} from '@loopback/repository';
```
bajtos: You cannot import from `@loopback/repository` because it is not a dependency of `@loopback/booter-lb3app`, and IMO it should not be. Let's remove this import and use different types, see my next comment.

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -137,4 +151,5 @@ export interface Lb3AppBooterOptions {
 
 interface Lb3Application extends ExpressApplication {
   handler(name: 'rest'): ExpressRequestHandler;
+  dataSources?: AnyObject;
```
bajtos: Can we use a stricter type definition please?

For example:

```suggestion
  dataSources?: {[name: string}: object};
```

Or perhaps:

```suggestion
  dataSources?: {[name: string}: {name: string}};
```

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -62,6 +63,19 @@ export class Lb3AppBooter implements Booter {
...
     }
 
+    const dataSources = lb3App.dataSources;
+    if (dataSources) {
+      const visited: DataSource[] = [];
+      Object.values(dataSources).forEach(ds => {
+        if (visited.includes(ds)) return;
+        visited.push(ds);
+        this.app
+          .bind(`datasources.lb3-${ds.name}`)
```
bajtos: On the second thought, I am not sure if it's a good idea to use `ds.name`. From what I remember, the code creating `ds.name` is messy and does not always honor the name provided by the user, replacing it with connector name sometimes.

Can we use the key in the `dataSource` object as the data-source name? As a nice side-effect, this will allow us to treat data-sources as opaque `object` type.

Not a big deal though, feel free to ignore this comment.

derdeka:
Is it possible to disable this feature by options somehow (migration of lb3 models)? What happens if lb3 models and lb4 models have the same names or use the same database tables? Not all users might want to mix up lb3 and lb4 applications in that way.

packages/booter-lb3app/fixtures/app-with-model.js
agnes512: I see, thank you for the clarification!

packages/booter-lb3app/src/lb3app.booter.ts
hacksparrow: Going with `dataSources?: {[name: string]: DataSource};`.

hacksparrow:
@derdeka mounting LB3 app on a LB4 requires multiple steps to be followed. If one does not want their LB3 models to be mounted, those models should be removed from the LB3 app while the LB3 app being set up to be mounted in the LB4 app.

hacksparrow:
@bajtos thanks for the nice suggestions, I have made the changes accordingly.

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -5,6 +5,7 @@
 
 import {BootBindings, Booter} from '@loopback/boot';
 import {CoreBindings, inject} from '@loopback/core';
+import {DataSource} from '@loopback/repository';
```
bajtos: As I commented before, `@loopback/repository` is not in regular/peer dependencies of `booter-lb3app`, therefore it may not be available at runtime, when this booter is installed in a LB project.

Also the `DataSource` type is not the right one to use, it described LB4 objects but your code is using LB3 instances! The correct type is `juggler.DataSource`.

Please remove this import and find a different solution how to describe the shape (interface) of LB3 datasource objects.

bajtos:
@derdeka
> Is it possible to disable this feature by options somehow (migration of lb3 models)? What happens if lb3 models and lb4 models have the same names or use the same database tables? Not all users might want to mix up lb3 and lb4 applications in that way.

Thank you for raising this concern!

I see only one scenario where import of LB3 & LB4 models can create conflict, and that's when there is a model with the same name attached to a datasource configured to use the same database, and thus share the same table. Additionally, the LB3 and LB4 models must defined differently.

It makes me wonder, what is the intent of application developer in such case. If the model definition is different between LB3 and LB4 applications, isn't this going to create inconsistencies in data? E.g. LB3 endpoints writing table rows in different format than expected by the LB4 model.

Setting that aside, I think it makes sense to give developers more control over which models are migrated. I don't think the decision should be based on LB3 or LB4, I'd prefer a more generic solution allowing developers to specify which datasources to migrate or even which models to migrate.

It turns out this is already supported via `SchemaMigrationOptions` and its `models` field. My recommendation is to overwrite `migrateSchema` in your application class and supply the list of models to migrate.

```ts
class MyApplication extends BootMixin(
  RepositoryMixin(RestApplication)
) {
  // ...

  migrateSchema(options: SchemaMigrationOptions = {}): Promise<void> {
    if (!options.models) {
      options = {...options, models: [/* put your model names here */]};
    }
    return super.migrateSchema(options);
  }
}
```

I think this is not ideal, let's discuss how to improve it here: https://github.com/strongloop/loopback-next/issues/3819

bajtos:
Please consider adding a short paragraph to [Mounting the LoopBack 3 Application in a LoopBack 4 Project](https://loopback.io/doc/en/lb4/Migrating-from-LoopBack-3.html#mounting-the-loopback-3-application-in-a-loopback-4-project) and mention that `npm run migrate` will migrate all models from the LB3 application too.

emonddr:
Great stuff @hacksparrow :)

PR #3805, opened 2019-09-25
docs: update keyFrom, keyTo for hasMany and hasOne relations

agnes512:
- Enhance the  explanations for relation metadata
- Add 2-3 cases of customizing id/foreign key names via the metadata for `hasMany` and `hasOne` relations.

## Checklist

### DIFF
--- a/docs/site/HasMany-relation.md
+++ b/docs/site/HasMany-relation.md
@@ -44,7 +44,8 @@ routes, you need to perform the following steps:
 This section describes how to define a `hasMany` relation at the model level
 using the `@hasMany` decorator. The relation constrains the target repository by
 the foreign key property on its associated model. The following example shows
-how to define a `hasMany` relation on a source model `Customer`.
+how to define a `hasMany` relation on a source model `Customer` and a target
+model `Order`.
 
 {% include code-caption.html content="/src/models/customer.model.ts" %}
 
@@ -85,7 +86,9 @@ inferred properly as an array of the target model instances.
 
 The decorated property name is used as the relation name and stored as part of
 the source model definition's relation metadata. The property type metadata is
-also preserved as an array of type `Order` as part of the decoration.
+also preserved as an array of type `Order` as part of the decoration. (Check
+[Relation Metadata](HasMany-relation.md#relation-metadata) section below for
+more details)
 
 A usage of the decorator with a custom foreign key name for the above example is
 as follows:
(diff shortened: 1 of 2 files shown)


docs/site/HasMany-relation.md
```diff
@@ -178,6 +180,42 @@ export interface OrderRelations {
...
+
+We can specify the `keyTo` field via `@hasMany` decorator if we have customized
+the foreign key name as `my_customer_id` instead of `customerId`:
+
+```ts
+@model()
+export class Order extends Entity {
+  // constructor, properties, etc.
+  @belongsTo(() => Customer)
+  my_customer_id: number; // customized foreign key name
```
bajtos: In this example, I believe the relation definition will use the name `my_customer_id`. This will create problems when including related models, because we will replace `my_customer_id` property from the numeric id to the actual customer instance.

I am proposing to fix the relation name.

```ts
@belongsTo(() => Customer, {name: 'customer'})
my_customer_id: number; // customized foreign key name
```

agnes512: Right mb I forgot to add `name`. 

I am thinking that to have two cases in this doc:
- models only have one relation: hasMany or hasOne, no belongsTo
  - the foreign key is not decorated by `@belongsTo`, we don't need to modify property metadata in another model (`Order` in this case)
  - simple usage to show the idea of `keyFrom, keyTo`
- models have hasMany/hasOne and belongsTo:
  - modify metadata in both models to customize the names.
  - the case to show what the changes it needs to make if we have such relations

I don't know if it's necessary to have two cases, thoughts? @bajtos

dhmlau: I'm wondering... if we have case #2 you described above, doesn't it mean we have case #1 covered too? Or is there anything we'd need to do differently?

docs/site/HasMany-relation.md
```diff
@@ -178,6 +180,55 @@ export interface OrderRelations {
...
+  // constructor, properties, etc.
+  @property({
+    type: 'number',
+  })
+  my_customer_id: number; // customized foreign key name
+}
+```
+
+Notice that if you decorate the foreign key of the target model with
+`@belongsTo`, you also need specify the relation name in it's metadata. See
```
dhmlau: 2 suggested changes in this sentence. 
From:
> you also need specify the relation name in it's metadata
to:

> you also need _to_ specify the relation name in ~it's~ its metadata

docs/site/hasOne-relation.md
```diff
@@ -145,6 +145,55 @@ class Supplier extends Entity {
...
+  // constructor, properties, etc.
+  @property({
+    type: 'number',
+  })
+  supplier_id: number; // customized foreign key name
+}
+```
+
+Notice that if you decorate the foreign key of the target model with
+`@belongsTo`, you also need specify the relation name in it's metadata. See
```
dhmlau: same comment as above. suggestion change:
> you also need to specify the relation name in its metadata

docs/site/HasMany-relation.md
agnes512: They are a bit different from my understanding. 
For case #1, the change is only needed within `hasMany` relation. 
For case #2, we are modifying `belongsTo` relation as well ([here](https://github.com/strongloop/loopback-next/pull/3805/files#diff-65c8c6f0fd2bb35d13770c5822272740R227)). 

When I just started using LB4, it was not clear to me that if I can use hasMany without belongsTo cause we use them together in every example. So I want to make it clear in the docs. I am ok to just keep #2 if it's not necessary.

bajtos: My opinion: If we want to show how to use HasMany without BelongsTo, then we should make that intent explicit. Create a new section, explain that it's possible to use one relation without the other and offer an example how to do it.

Then in the rest of the documentation, I'd prefer to stay consistent with the prevalent approach, which is to describe both relations in tandem.

agnes512: yes! in `hasMany`, we introduced the usages of `hasMany` without and with `belongsTo`. So I think it should be fine to have two `keyTo` cases mapping with those two usages.

docs/site/tables/lb4-relation-metadata-hasmany.html
```diff
@@ -0,0 +1,36 @@
+<head>
+  {% include analytics.html %}
+</head>
```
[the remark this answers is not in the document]
agnes512: removed

docs/site/HasMany-relation.md
```diff
@@ -85,7 +85,9 @@ inferred properly as an array of the target model instances.
 
 The decorated property name is used as the relation name and stored as part of
 the source model definition's relation metadata. The property type metadata is
-also preserved as an array of type `Order` as part of the decoration.
+also preserved as an array of type `Order` as part of the decoration. (Check
+[Relation Metadata](https://loopback.io/doc/en/lb4/HasMany-relation.html#relation-metadata)
```
nabdelgadir: https://loopback.io/doc/en/lb4/HasMany-relation.html#relation-metadata -> HasMany-relation.md#relation-metadata

docs/site/HasMany-relation.md
```diff
@@ -178,6 +180,55 @@ export interface OrderRelations {
...
+- `name`: the name of the relation, which is the `orders` property of `Customer`
+  in our example.
+
+We can specify the `keyTo` field via `@hasMany` decorator if we have customized
+the foreign key name as `my_customer_id` instead of `customerId`:
+
+```ts
+class Customer extends Entity {
+  // constructor, properties, etc.
+  @hasMany(() => Order, {keyTo: 'my_custom_id'})
```
nabdelgadir: Should this be 'my_customer_id' to match below?

docs/site/HasMany-relation.md
```diff
@@ -178,6 +180,55 @@ export interface OrderRelations {
...
+  @property({
+    type: 'number',
+  })
+  my_customer_id: number; // customized foreign key name
+}
+```
+
+Notice that if you decorate the foreign key of the target model with
+`@belongsTo`, you also need to specify the relation name in its metadata. See
+[BelongsTo](BelongsTo-relation.md) for more details.
```
nabdelgadir: Should this link to the relation metadata part of BelongsTo?

agnes512: There is no certain section in belongsTo

agnes512: BelongsTo doesn't have such a section.

dhmlau:
Force merge this PR because the CI failure is due to unavailability of the geocoding service.

Issue #2639, opened 2019-03-23
[Docs] belongsTo() not working in loopback4 - Cannot read property 'target' of undefined


Ajtak:
I have create api which fetch the data from database(PostgreSQL). I have same problem as #2209  

 Table 1 : type(id, name)
 Table 2 : tournament(id, type_id, title)
![obrazek](https://user-images.githubusercontent.com/4963819/54869076-053ea180-4d94-11e9-8c44-6ed6fc976fe2.png)
_PostgreSQL structure: https://github.com/Ajtak/lb4-target-error/blob/master/POSTGRESTRUCTURE.sql_

Table `type` is belongs to table `tournament`. I want fetch data from `type` table then respective `tournament` table data will fetch.

It will return error : 500 TypeError: Cannot read property 'target' of undefined

Sample repo replicate problem:
`git clone https://github.com/Ajtak/lb4-target-error.git`

Just I want to get `tournament` with our `types`.
Thx.

nabdelgadir:
I tried out your application and it seems the problem is coming from `tournament.model.ts`, specifically:
```ts
@belongsTo(() => Type)
person_id: number;
```
I tested it with an in-memory connector, but it seems to work if you change `person_id` to something like `type`/`typeId` (`type_id` or `personId` won't work either).

Can you try changing it and see if it solves your problem?

Ajtak:
If I changed type_id to type, i see only 

```
 [
     {
          "id":1,
          "type":1,
          "title":"XXXX"
    }
 ]
```
without type's name. What is problem?

nabdelgadir:
If you want the actual `Type` object instead of its `id`, change
```ts
@belongsTo(() => Type)
type: number;
```
to
```ts
@belongsTo(() => Type)
type: Type;
```

Ajtak:
If I changed datatype, id only is returned again instead of object.
Is need some change in tournament controller?
```javascript
// Uncomment these imports to begin using these cool features!
// import {inject} from '@loopback/context';
import {Filter, repository} from '@loopback/repository';
import {param,  get, getFilterSchemaFor} from '@loopback/rest';
import {TournamentRepository} from "../repositories";
import {Tournament} from "../models";

export class TournamentController {
    constructor(
        @repository(TournamentRepository)
        public tournamentRepository: TournamentRepository,
    ) { }
    @get('/tournaments', /*{...}*/)
    async find(
[...]

PR #3772, opened 2019-09-19
docs: update belongsTo docs with decorator syntax

deepakrkris:
Fixes https://github.com/strongloop/loopback-next/issues/2639

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3721, opened 2019-09-12
feat(repository): implement inclusionResolver for belongsTo relation 

agnes512:
resolves https://github.com/strongloop/loopback-next/issues/3448

This PR implements the inclusionResolver for `belongsTo` relation:
- Add a helper `flattenTargetsOfOneToOneRelation` and its unit tests for `belongsTo` and `hasOne` relation ( since they both use this helper to implement inclusion resolver)
- Add a new function `createBelongsToInclusionResolver` into the belongsTo factory `createBelongsToAccessorFor `
- Add tests for `createBelongsToInclusionResolver` in `repository-tests`.
  - introduce a new test-suite flag `supportsInclusionResolvers`.
- Add docs to explain the idea of `belongsToInclusionResolver`, and the basic setup/usages.

**Suggested way to review:**  
-> start with the document file and `belongs-to-inclusion-resolver.acceptance.ts` to get the idea of the `inclusion resolver`.
->  review `belongs-to.inclusion-resolver.ts` to see the implementation details.
-> review the rest of files.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #3808, opened 2019-09-25
docs: update belongsTo docs with decorator syntax

deepakrkris:
Fixes https://github.com/strongloop/loopback-next/issues/2639

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -43,6 +43,31 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined.
+
+The `@belongsTo` decorator takes three parameters:
+
+- `target model class` (required)
+- `relation definition` (optional) - has three attributes, keyFrom, keyTo, name
+  - `keyFrom` is the property name of the foreign key on the "source" model. It
+    is always set to the decorated property name (in the given example it is,
+    "customerId" property on Order model).
+  - `keyTo` is the property name of the foreign key on the "target" model, it is
+    typically the primary key of the "target" model. `keyTo` attribute defaults
+    to the id property on the target model (in the given example, "id" property
+    on Customer).
+  - `name` is the name of the relation as defined in the repository. The
+    relation name is used in the repository constructor to define a
+    [BelongsToAccessor](#configuring-a-belongsto-relation) and map it to the
+    relation using a
+    [inclusion resolver](#enabledisable-the-inclusion-resolvers).
+- `property definition` (optional) - creates a property decorator implicitly.
+  The name attribute in the definition can be used to customize datasource
+  column name.
+
 {% include code-caption.html content="/src/models/order.model.ts" %}
 
 ```ts
@@ -75,18 +100,51 @@ export interface OrderRelations {
 export type OrderWithRelations = Order & OrderRelations;
 ```
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+The standard naming convention for the foreign key property in the source model
+is `relation name` + `Id` (for example, Order.customerId).
 
-A usage of the decorator with a custom primary key of the target model for the
-above example is as follows:
+- If the foreign key property name in the source model has to be customized, the
+  relation name has to be explicitly specified in the `name` attribute of the
+  relation definition. In addition, if you have a corresponding `hasMany` or
+  `hasOne` relation in the target model (for example, a Customer has many
+  Orders), the `keyTo` attribute of that corresponding relation needs to be
+  stated explicitly. Check the relation metadata in
+  [hasMany](https://loopback.io/doc/en/lb4/HasMany-relation.html#relation-metadata)
+  and
+  [hasOne](https://loopback.io/doc/en/lb4/hasOne-relation.html#relation-metadata)
+  for more details.
+- In the following example, the foreign key property name is customized as
+  `cust_Id` instead of `customerId`. so the relation definition in the second
+  argument is explicitly passed to the `belongsTo` decorator.
 
 ```ts
 class Order extends Entity {
   // constructor, properties, etc.
-  @belongsTo(() => Customer, {keyTo: 'pk'})
+  @belongsTo(() => Customer, {keyFrom: 'cust_Id', name: 'customer'})
+  cust_Id: number;
+}
+```
+
+In the following example, the db column name of the foreign key is customized by
+passing the property definition in the third argument to the `belongsTo`
+decorator.
+
+```ts
+class Order extends Entity {
+  // constructor, properties, etc.
+  @belongsTo(() => Customer, {keyFrom: 'customerId'}, {name: 'customer_id'})
+  customerId: number;
+}
+```
+
+The `keyTo` attribute in the relation definition has to be stated explicitly,
+when the property of the target model for the belongsTo relation is not the id
+property in the target model.
+
+```ts
+class Order extends Entity {
+  // constructor, properties, etc.
+  @belongsTo(() => Customer, { keyTo: 'customized_target_property' }),
   customerId: number;
 }
 


deepakrkris:
@agnes512 agree on your comment to add example for modifying foreign key property in source model on the previous PR. I have added in this PR.

deepakrkris:
closed previous PR https://github.com/strongloop/loopback-next/pull/3772 because of travis job errrors.

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of belongsTo
+relation is similar to legacy datasource juggler.
+
+The `@belongsTo` decorator takes three parameters :
+
+- the `target model class`, is a mandatory field
+- the `relation definition`, is not a mandatory field - has three attributes,
+  keyFrom, keyTo, name
+  - `keyFrom` is a property name of the the foreign key on the "source" model
```
bajtos:
```suggestion
  - `keyFrom` is a property name of the foreign key on the "source" model
```

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+    (e.g. Order.customerId). keyFrom is always set to the decorated property
```
bajtos:
```suggestion
    (e.g. `customerId` on `Order` model). `keyFrom` is always set to the decorated property
```

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+    name. keyFrom attribute has become a dummy parameter and is not relevant any
```
bajtos:
```suggestion
    name. `keyFrom` attribute has become a dummy parameter and is not relevant any
```

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+    more with the recent changes.
+  - `keyTo` is a property name on the "target" model, typically the primary key
+    of the "target" model. keyTo attribute defaults to the value `id`.
```
bajtos:
```suggestion
    of the "target" model. `keyTo` attribute defaults to the value `id`.
```

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+The above example can also be expanded with explicit relation and property
+definitions as follows,
+
+- The standard naming convention for the foreign key property in the source
+  model is `relation name` + `Id`. (e.g. Order.customerId).
```
agnes512: `relation name` + `Id` -> `source model name + Id`?

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
+    customized name (`cust_Id`).
+  - a property definition is added with the customized db column name
+    (`customer_id`).
+
+```ts
+class Order extends Entity {
+  // constructor, properties, etc.
+  @belongsTo(
+    () => Customer,
+    {keyFrom: 'customerId', keyTo: 'id', name: 'customer'},
```
agnes512: `keyFrom: 'customerId' `-> `keyFrom: 'cust_Id'` ?

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
+    {name: 'customer_id'},
```
agnes512: I think {`keyFrom`: '.., `name`: ..} is enough for customizing the name ( or even just to have the relation name {`name`: ..})

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
+```
 
 A usage of the decorator with a custom primary key of the target model for the
 above example is as follows:
 
 ```ts
 class Order extends Entity {
   // constructor, properties, etc.
-  @belongsTo(() => Customer, {keyTo: 'pk'})
+  @belongsTo(() => Customer, { keyTo: 'customized_target_property' }),
```
agnes512: 👍

docs/site/BelongsTo-relation.md
deepakrkris: keyFrom is dummy anyway

docs/site/BelongsTo-relation.md
deepakrkris: this example is to provide all the parameters for the decorator

docs/site/BelongsTo-relation.md
deepakrkris: @agnes512 do you mean `target model name + Id` ?

agnes512: ah, yes, I was thinking `Customer` is the source of this foreign key. But I think `target model` is more accurate according to the context.

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
agnes512: I don't know. It would be really confusing if I didn't know LB4 that much. 
If `keyFrom` is always set to the decorated property name, why not set it to `cust_id` or even remove it?

or you can add a simplified one after it as:

``` 
or we can simplified it as 
@belongsTo(
    () => Customer,
    {name: 'customer'},
    {name: 'different_property_name_for_db'} // pass in the third param if you'd like to use different name for db column 
)
my_customer_id: number;
```

Personally I'd like to make docs easy to understand at the first glance instead of having everything up all at once. Like the `getting started` part of LB4, it is simple and also has enough material to start. It's good to have the explanation about using different name for database, appreciate that. So I would probably to break it down as: 
```
here's an example of having customized name
...
you can simplify the above example to ...
...
if you would need to use names of db columns, you can do
...
```

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of belongsTo
+relation is similar to legacy datasource juggler.
+
+The `@belongsTo` decorator takes three parameters :
+
+- `target model class`, is a mandatory field
+- `relation definition`, is not a mandatory field - has three attributes,
```
dhmlau: `not a mandatory field` -> `an optional field`?
similarly for the subsequent occurrance..

deepakrkris: agreed

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+The above example can also be expanded with explicit relation and property
+definitions as follows,
+
+- The standard naming convention for the foreign key property in the source
+  model is `target model name` + `Id`. (e.g. Order.customerId).
```
emonddr: In our documentation, we try to stay away from using `e.g.` since it doesn't translate well. It is best to use `for example`. See https://loopback.io/doc/en/contrib/Doc-style-guide.html#spelling-grammar-and-style

emonddr:
Check one or both of these, depending on your changes..

image

docs/site/BelongsTo-relation.md
deepakrkris: The idea is to explain the overall syntax to the developer. we should be able to explain in as few examples as possible, instead of providing multiple combinations.

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+relation is similar to legacy datasource juggler.
+
+The `@belongsTo` decorator takes three parameters :
+
+- `target model class`, is a mandatory field
+- `relation definition`, is not a mandatory field - has three attributes,
+  keyFrom, keyTo, name
+  - `keyFrom` is a property name of the foreign key on the "source" model (e.g.
+    `customerId` on `Order` model). `keyFrom` is always set to the decorated
+    property name. `keyFrom` attribute has become a dummy parameter and is not
```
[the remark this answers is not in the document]
deepakrkris: no special focus needed here, since the indentation and bullets take care of readability.

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
agnes512: Alright, I will respect your decision on this.

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+relation is similar to legacy datasource juggler.
+
+The `@belongsTo` decorator takes three parameters :
+
+- `target model class`, is a mandatory field
+- `relation definition`, is an optional field - has three attributes, keyFrom,
+  keyTo, name
+  - `keyFrom` is a property name of the foreign key on the "source" model (for
+    example, `customerId` on `Order` model). `keyFrom` is always set to the
+    `decorated property name`. `keyFrom` attribute has become a dummy parameter
```
dhmlau: Sorry that I miss this sentence.  Is it the intended behavior?  If it's a bug, I think we should open a new ticket to address that rather than putting it in the documentation.

bajtos: From what I remember from my spike work, `keyFrom` is still used by inclusion resolvers. It's an optional field though because we can infer it automatically in most cases.

deepakrkris: @bajtos it is always automatically inferred from the decorated target. The passed in parameter is never used by inclusion resolvers.

deepakrkris: @dhmlau it is not a bug. Intuitively the keyFrom attribute is not relevant anymore, because the property name is available via the decorator anyway (as the decorated target).

deepakrkris: we can clean up by removing the keyFrom attribute or make the decorator signature more understandable, later. for now the docs state things as they are.

agnes512: Do you mean to remove `keyFrom` from metadata? I don't think it's a good idea...
Could you remove the "keyFrom attribute has become a dummy parameter" part or reword it? It's confusing.

jannyHou: @deepakrkris I think `keyFrom` is still honored, see the test case https://github.com/strongloop/loopback-next/pull/3726/files#diff-6795fbd89427a8971f81ebe5f4487536R16.

And that PR added `keyFrom` field to the metadata.

deepakrkris: @jannyHou do you have an end to end test to validate that keyFrom is honoured ? because the checks on metadata only proves that what is passed in is available in the metadata.

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +98,44 @@ export interface OrderRelations {
...
+The above example can also be expanded with explicit relation and property
+definitions as follows,
+
+- The standard naming convention for the foreign key property in the source
+  model is `target model name` + `Id`. (for example, Order.customerId).
+- If the property name in the source model has to be customized then the `name`
+  attribute in the relation definition (the target model name) has to be stated
+  explicitly. Also, the corresponding `hasMany` relation in the `target` model
+  has to be modified with the correct `keyTo` value.
+- In the following example :
```
nabdelgadir: nitpick: remove space before `:`

docs/site/BelongsTo-relation.md
jannyHou: @deepakrkris When looking at the [decorator's implementation](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L23), I feel the 3rd parameter(property definition) is more for configuring the property, like:

- Is it required/generated/its lengh/default value, see [property configuration](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#general-property-properties), 
- And json schema config, see [type PropertyDefinition](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L29-L36)

`name` should be inferred from the property `cust_Id` itself.

docs/site/BelongsTo-relation.md
dhmlau: @agnes512 @jannyHou, I think I might be seeing what @deepakrkris described. I've created my own app (also with Customer and Order). Even I put some irrelevant characters in `keyTo` _and_ `keyFrom`, it seems to be working fine. i.e. 
```
  @belongsTo(() => Customer, {keyFrom: 'aaaaaa', keyTo: 'aaaaaa'})
  customer_id: string;
```
I'm not sure if it is because I have the `hasMany` relations defined already and somehow it got magically discovered.

To move this task and PR forward, I'd like to suggest:
- defer the task and include the scope to investigate why the values in `keyFrom` and `keyTo` are not honored. Either it got somehow discovered or it's a bug, we will document this properly. 
- keep this PR and modify as needed.

@strongloop/loopback-maintainers (especially @deepakrkris @agnes512 @jannyHou), what do you think?

jannyHou: sure let's address it in a separate story. It shouldn't be a blocker for this PR.

deepakrkris: @dhmlau , work on accessors seems to be still in progress , so I can remove references to keyFrom and land this PR.

agnes512: @dhmlau
by 'it works fine' you mean post/get requests? what's the value of `keyFrom` and `keyTo` in your case?
 Have you tried inclusion resolver for these relations? cause they are actually using `keyFrom` and `keyTo` (sorry I didn't realize earlier that these fields are also related to other part of features): [belongTo](https://github.com/strongloop/loopback-next/pull/3721/files#diff-ee138af2e31d5dd9d5c43732a35a0d04R52)
[hasMany](https://github.com/strongloop/loopback-next/pull/3595/files#diff-d6e69145b3b8445f547492d818ef8d6fR53)
I haven't tested it out with inclusion resolver what would happen if we put random names here.

agnes512: @jannyHou @dhmlau @deepakrkris I just tested it real quick. 

I don't know if the app works fine with @dhmlau means we can do post/get requests. I tested with inclusion resolver. 
For `belongsTo` relation, if you have a customized name `cust_id`, you have to specify {`name`} field in the second parameter in decorator `@belongsTo`. `keyFrom` is optional but it **cannot be** a random name. Or it couldn't get related models through the relation.
I think the `keyFrom` field is honored from the decorated property name, see [here](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L50).

So I think `keyFrom` shouldn't be removed. And we need to make it clear in the docs.

dhmlau: @agnes512, yes, and do post and get. I'm using postgresql, not sure if it matters. 
Could you please share your app in github? thanks!

agnes512: I will also rework my [update keyFrom, keyTo for hasMany and hasOne relations](https://github.com/strongloop/loopback-next/pull/3805) PR to make sure that these explanations will be consistent with inclusion resolver.

agnes512: @dhmlau I use my tests in `packages/repository` :joy: cause I can run it on MySQL and MongoDB. I modified these fields to test it around. I had a feeling that post/get may work cause they don't relate to `keyFrom` or `keyTo`. But I am sure that the inclusion resolver will be affected if it has incorrect `keyFrom`/`keyTo`.

agnes512: I think we can explain what these fields represent. And mention that to customize the property name decorated by `@belongsTo`, you will have to specify the `name` field. Just an idea~

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of belongsTo
+relation is similar to legacy datasource juggler.
```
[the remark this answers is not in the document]
deepakrkris: agree

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+
+The `@belongsTo` decorator takes three parameters :
```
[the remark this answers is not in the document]
deepakrkris: agree

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of belongsTo
```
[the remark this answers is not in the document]
deepakrkris: agree

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,29 @@ related routes, you need to perform the following steps:
...
+relation is similar to legacy datasource juggler.
+
+The `@belongsTo` decorator takes three parameters :
+
+- `target model class`, is a mandatory field
```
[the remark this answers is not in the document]
deepakrkris: agree on the first. disagreeing on the second, do not see the need for explaining what model class is or how to obtain a reference to it.

docs/site/BelongsTo-relation.md
deepakrkris: keyFrom and keyTo are working as expected with the recent published lb4 version. updating the docs as per current status. I will remove the dummy param comment as well.

docs/site/BelongsTo-relation.md
deepakrkris: agree

docs/site/BelongsTo-relation.md
agnes512: Yeah just make sure that next time you use the latest code when you're testing something via `lb4 app` :smile:

docs/site/BelongsTo-relation.md
deepakrkris: @jannyHou that is very strange, PropertyDefinition does not have a `name` key defined, how is typescript accepting the `name` attribute ? also I am using the explicit `name` key to customize db column name and it works perfectly fine !

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
...
+- `relation definition` (optional) - has three attributes, keyFrom,
+  keyTo, name
+  - `keyFrom` is a property name of the foreign key on the "source" model (for
+    example, `customerId` on `Order` model). `keyFrom` is always set to the
+    `decorated property name`.
+  - `keyTo` is a property name on the "target" model, typically the primary key
+    of the "target" model. `keyTo` attribute defaults to the value `id`.
+  - `name` is the target model name. name attribute defaults to
+    `{decorated-property-name}` after stripping the trailing `Id` suffix
+- `property definition`, is an optional field - creates a property decorator
```
nabdelgadir: for consistency: `property definition`, is an optional field -> `property definition` (optional)

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
...
 
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of
+belongsTo relation is similar to legacy [datasource juggler](https://github.com/strongloop/loopback-datasource-juggler).
+
+The `@belongsTo` decorator takes three parameters:
+
+- `target model class` (mandatory)
```
nabdelgadir: Can you change `mandatory` to `required`? We usually use `required`.

deepakrkris: agree

docs/site/BelongsTo-relation.md
deepakrkris: agree

docs/site/BelongsTo-relation.md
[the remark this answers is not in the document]
deepakrkris: lint is not agreeing to wrap belongsTo

docs/site/BelongsTo-relation.md
deepakrkris: @agnes512 I agree with your comment https://github.com/strongloop/loopback-next/pull/3808#discussion_r328731959 , so I have updated the docs as per the suggestions.

docs/site/BelongsTo-relation.md
deepakrkris: @jannyHou , also one more thing, `cust_Id` being a customized name (per standard naming convention it should be `customerId`), the `name` attribute cannot be inferred from the customized name, it has to be explicitly stated.

docs/site/BelongsTo-relation.md
deepakrkris: also comment https://github.com/strongloop/loopback-next/pull/3772#discussion_r326426738 is addressed in this PR .

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +97,45 @@ export interface OrderRelations {
...
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+The standard naming convention for the foreign key property in the source model
+is `target model name` + `Id` (for example, Order.customerId).
+
+- If the foreign key property name in the source model has to be customized,
+  then the `name` attribute in the relation definition (the target model name)
```
agnes512: Same as above, the `name` here is relation name. Users can name the relation based on their preference. For instance, they can name it `orderCustomer` instead of `customer` in the `OrderRelations`. 

If the foreign key property name in the source model has to be customized, the relation name has to be specified in  the `name` attribute of the relation definition. In addition, if you have a the corresponding `hasMany` or `hasOne` relation in the `target` model, e.g a `Customer` has many `Order`s, the `keyTo` field of that corresponding property needs to be specify. Check the `#Relation Metadata` in [`hasMany`](https://loopback.io/doc/en/lb4/HasMany-relation.html) and [`hasOne`](https://loopback.io/doc/en/lb4/HasOne-relation.html) for more details.
(Should either add an example of `hasMany` or a link here.)

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +97,45 @@ export interface OrderRelations {
...
+class Order extends Entity {
+  // constructor, properties, etc.
+  @belongsTo(() => Customer, {keyFrom: 'customerId'}, {name: 'customer_id'})
+  customerId: number;
+}
+```
 
-A usage of the decorator with a custom primary key of the target model for the
-above example is as follows:
+The standard naming convention for the primary key in the target model is `id`.
```
agnes512: Same as above, `keyTo` has a default value (not just a naming convention), which is the primary key of the source model. 
By default, no matter what the name of the id property is in the source mode, `keyTo` would set to be that name automatically. In you case, the default would be `customized_target_property` directly. [See file](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.helpers.ts#L60).

If the primary key we use in this `belongsTo` relation is still be the id property of source model, `keyTo` should be optional here. 

We probably had `id` as a default value before. So it was documented this way. I'd suggest to remove this example or modify it to suit our code base more.

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
...
+belongsTo relation is similar to legacy [datasource juggler](https://github.com/strongloop/loopback-datasource-juggler).
+
+The `@belongsTo` decorator takes three parameters:
+
+- `target model class` (mandatory)
+- `relation definition` (optional) - has three attributes, keyFrom,
+  keyTo, name
+  - `keyFrom` is a property name of the foreign key on the "source" model (for
+    example, `customerId` on `Order` model). `keyFrom` is always set to the
+    `decorated property name`.
```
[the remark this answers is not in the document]
deepakrkris: agreed

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
...
+  - `keyTo` is a property name on the "target" model, typically the primary key
+    of the "target" model. `keyTo` attribute defaults to the value `id`.
```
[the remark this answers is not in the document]
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
...
+  - `name` is the target model name. name attribute defaults to
```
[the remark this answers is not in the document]
deepakrkris: agree

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,28 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
+decorator. The decorator takes in a function resolving the target model class
+constructor and designates the relation type. It also calls `property()` to
+ensure that the decorated property is correctly defined. Usage of
+belongsTo relation is similar to legacy [datasource juggler](https://github.com/strongloop/loopback-datasource-juggler).
```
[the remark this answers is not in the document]
deepakrkris: agree

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +100,51 @@ export interface OrderRelations {
 export type OrderWithRelations = Order & OrderRelations;
 ```
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+The standard naming convention for the foreign key property in the source model
+is `relation name` + `Id` (for example, Order.customerId).
```
agnes512: The explanation here looks weird. It seems like that we define the relation name first, then we decide the default value of our foreign key name. This is because `keyFrom` defaults to the decorated property name, not `targetModel` + `Id`.

I wanted to point out in this PR that, in our code, the way we get the default relation name is by modifying the decorated property name:
```
customerId -> (get rid of `id`) -> customer
```
I found this code is really confusing. For example, if we decorate the foreign key name as
```ts
@belongsTo(..)
anyRandomNameId
```
the default relation name will be `anyRandomName`, which doesn't make sense at all. I guess it was designed to use the default foreign key `customerId`. But the foreign key doesn't default to `targetModel` + `Id`. It defaults to the decorated name. That's the reason makes it hard to document. In short, the code doesn't fully deal with default values as we expected.
 
We might want to improve the way to get the relation name in `belongTo` in the future.

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +100,51 @@ export interface OrderRelations {
...
+
+- If the foreign key property name in the source model has to be customized, the
+  relation name has to be explicitly specified in the `name` attribute of the
+  relation definition. In addition, if you have a corresponding `hasMany` or
+  `hasOne` relation in the target model (for example, a Customer has many
+  Orders), the `keyTo` attribute of that corresponding relation needs to be
+  stated explicitly. Check the relation metadata in
+  [hasMany](https://github.com/strongloop/loopback-next/blob/master/docs/site/HasMany-relation.md)
```
agnes512: Suggestion: 
https://loopback.io/doc/en/lb4/HasMany-relation.html#relation-metadata

nabdelgadir: Or just `[hasMany](HasMany-relation.md#relation-metadata)`

docs/site/BelongsTo-relation.md
deepakrkris: yes the code is not very self explanatory in terms of defining relations as of now

docs/site/BelongsTo-relation.md
deepakrkris: agreed

docs/site/BelongsTo-relation.md
```diff
@@ -75,18 +100,51 @@ export interface OrderRelations {
...
+  and
+  [hasOne](https://github.com/strongloop/loopback-next/blob/master/docs/site/hasOne-relation.md)
```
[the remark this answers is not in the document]
deepakrkris: agreed

docs/site/BelongsTo-relation.md
raymondfeng: Well, the FK usually refers to the target model, such as `customerId`. `anyRandomNameId` does not name the FK meaningfully. In such cases, the options.name should be used.

docs/site/BelongsTo-relation.md
jannyHou: @deepakrkris sorry for replying late (refreshed the PR several times last week but somehow git didn't show your message, weird...)

> PropertyDefinition does not have a name key defined, but typescript is accepting the name attribute

It's because type `PropertyDefinition` allows arbitrary properties, see [code](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L36)

> Also I am explicitly using the name key to customize db column names and it works perfectly fine for the todo-list example

Thank you got it, so it's used to configure the db column name.

PR #3828, opened 2019-09-27
fix: emit "type: 'object'," for schema definitions

noels:
- JSON Schema makes provision for structuring a complex schema into re-usable elements.
- Loopback makes use of reference schema re-use through the '$ref' keyword.
- in the JSON Schema documentation for complex schemas https://json-schema.org/understanding-json-schema/structuring.html the complex schema definitions are declared as type: 'object'.
- Loopback generated JSON Schema omits this declaration. build-schema.ts:353.
- Omitting this declaration causes ajv validation to pass when a required element (which is a reference to a different object) is set to null.
- Adding the declaration: "type: 'object'" to generated schemas enables ajv validation to correctly fail when a required referenced property is present but the value set to null.

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -391,6 +391,7 @@ export function modelToJsonSchema<T extends object>(
   } else if (descriptionSuffix) {
     result.description = descriptionSuffix;
   }
+  result.type = 'object';
 
   for (const p in meta.properties) {
     if (options.exclude && options.exclude.includes(p as keyof T)) {
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -29,6 +29,15 @@ describe('Validation at REST level', () => {
   let client: Client;
 
   @model()
+  class Category {
+    @property({required: true, type: 'string'})
+    name: string;
+
+    @property({required: false, type: 'string'})
+    description?: string;
+  }
+
+  @model()
   class Product {
     @property({required: true})
     name: string;
@@ -41,12 +50,18 @@ describe('Validation at REST level', () => {
     @property({required: true, jsonSchema: {range: [0, 100]}})
     price: number;
 
+    @property({required: true, type: Category})
+    category: Category;
+
     constructor(data: Partial<Product>) {
       Object.assign(this, data);
     }
   }
 
   const PRODUCT_SPEC: SchemaObject = jsonToSchemaObject(getJsonSchema(Product));
+  const CATEGORY_SPEC: SchemaObject = jsonToSchemaObject(
+    getJsonSchema(Category),
+  );
 
   // Add a schema that requires `description`
   const PRODUCT_SPEC_WITH_DESCRIPTION: SchemaObject = {
@@ -125,6 +140,7 @@ describe('Validation at REST level', () => {
         name: 'iPhone',
         description: null,
         price: 10,
+        category: {name: 'Phones'},
       };
       const res = await client
         .post('/products')
@@ -157,6 +173,7 @@ describe('Validation at REST level', () => {
         name: 'iPhone',
         description: 'iPhone',
         price: 200,
+        category: {name: 'Phones'},
       };
       const res = await client
         .post('/products')
@@ -214,6 +231,7 @@ describe('Validation at REST level', () => {
         name: 'iPhone',
         description: 'iPhone',
         price: 200,
+        category: {name: 'Phones'},
       };
       const res = await client
         .post('/products')
@@ -277,6 +295,7 @@ describe('Validation at REST level', () => {
         const DATA = {
           name: 'iPhone',
           description: 'iPhone',
+          category: {name: 'Phones'},
         };
         const res = await client
           .post('/products')
@@ -370,6 +389,7 @@ describe('Validation at REST level', () => {
       components: {
         schemas: {
           Product: PRODUCT_SPEC,
+          Category: CATEGORY_SPEC,
         },
       },
     })
@@ -400,6 +420,7 @@ describe('Validation at REST level', () => {
       name: 'Pencil',
       description: 'An optional description of a pencil',
       price: 10,
+      category: {name: 'Stationary'},
     };
     await client
       .post('/products')
@@ -412,6 +433,7 @@ describe('Validation at REST level', () => {
       name: 'Pencil',
       description: null,
       price: 10,
+      category: {name: 'Stationary'},
     };
     await client
       .post('/products')
@@ -421,7 +443,7 @@ describe('Validation at REST level', () => {
 
   async function serverAcceptsValidRequestBodyForUrlencoded() {
     const DATA =
-      'name=Pencil&price=10&description=An optional description of a pencil';
+      'name=Pencil&price=10&description=An optional description of a pencil&category[name]=Stationary';
     await client
       .post('/products')
       .set('Content-Type', 'application/x-www-form-urlencoded')
@@ -430,6 +452,7 @@ describe('Validation at REST level', () => {
         name: 'Pencil',
         description: 'An optional description of a pencil',
         price: 10,
+        category: {name: 'Stationary'},
       });
   }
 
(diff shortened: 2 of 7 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -348,6 +348,7 @@ export function modelToJsonSchema<T extends object>(
   if (meta.description) {
     result.description = meta.description;
   }
+  result.type = 'object';
```
raymondfeng: Isn't the `type` default to `object` if `properties` is present?

raymondfeng:
@noels Please also fix https://travis-ci.com/strongloop/loopback-next/jobs/239863212

packages/repository-json-schema/src/build-schema.ts
noels: I am not sure, but the test fails when the element is not present so the validation does not work as expected.

noels:
> Thank you for the pull request. Please add a new test to verify the scenario you have described in PR description:
> 
> * Loopback generated JSON Schema omits this declaration. build-schema.ts:353.
> * Omitting this declaration causes ajv validation to pass when a required element (which is a reference to a different object) is set to null.
> * Adding the declaration: "type: 'object'" to generated schemas enables ajv validation to correctly fail when a required referenced property is present but the value set to null.
> 
> This test should fail with a descriptive error message on the current `master` and pass with your changes in place.

Please can you give me some guidance as to where the test would be required. The framework tests the validation of the body in the Rest package tests (which I have now updated to work with my changes), but the schema is supplied to the validation requests as a test fixture. If I added a test there I think I would just be testing something that should already be covered in the ajv test suite. I have added tests that test for the presence of the 'type: "object" in the generated schema to the repository-json-schema package. If you can point me to where there are currently tests that verify that the generated JSONSchema works correctly with ajv validation I can add a new test there to ensure the behaviour is correct.

bajtos:
@noels thanks for asking. I was thinking about adding a new test to `packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts`. Apparently, we already have a test to verify that request bodies containing only `null` are rejected:

https://github.com/strongloop/loopback-next/blob/512d37d4f8165ab02c522f071eb1862bf57125b8/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts#L95-L101

If I understand your scenario correctly, then you want to reject requests when a required property is set to `null`. I think the simplest option is to define such property in the existing `Product` model and then add a new test to verify what happens when it's set to `null`. Just make sure to make the new property optional so that existing tests pass.

Alternatively, you can create a new `context()` block where you define a model & a controller that are specific to your scenario only. This is better in the long run, because we don't want the existing `Product` controller to grow super large to cover all different property types & definitions.

noels:
Thanks for the guidance. I have added an acceptance test to test for an object property being passed a null value. I have now fixed the prettier and commit message issues, and squashed the previous commits into a single commit for clarity.  
I have expanded the product controller to have a Category member. I didn't think that the test was sufficiently different to have to duplicate the product and create a separate controller and helper functions. Let me know if you don't think this is okay.  It looks like all the test will now pass.

noels:
It seems I forgot to include the new test when I copied everything over to a single commit. I've added the test. Please see if that's what you wanted.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -84,6 +99,21 @@ describe('Validation at REST level', () => {
...
+        price: 10,
+        category: null,
+      };
+
+      await client
+        .post('/products')
+        .type('json')
+        .send(DATA)
+        .expect(422);
+    });
```
bajtos: I am afraid this test is passing on the current master too. Here is the error reported:

```js
{
  statusCode: 422,
  name: 'UnprocessableEntityError',
  message: 'The request body is invalid. See error object `details` property for more info.',
  code: 'VALIDATION_FAILED',
  details: [
    {
      path: '',
      code: 'additionalProperties',
      message: 'should NOT have additional properties',
      info: [Object]
    }
  ]
}
```

@noels PTAL. 

We need a test that fails on the current master and thus reproduces the bug you are trying to fix. Without such test, we have very little confidence that your changes are actually achieving the desired goal, plus it's easy to introduce unwanted regressions in the future.

bajtos:
@noels Last but not least, please rebase your patch on top of the latest `master` branch.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
noels: HI @bajtos Thanks for having a look.  I reset my local copy to the upstream master, replaced the validation.acceptance.ts file with the one from my branch, then ran mocha in loopback-next/packages/rest
and got:
Error: expected 422 "Unprocessable Entity", got 200 "OK"
    at Test._assertStatus (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:268:12)
    at Test._assertFunction (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:283:11)
    at Test.assert (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:173:18)
    at localAssert (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:131:12)
    at /home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:128:5
    at Test.Request.callback (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:716:12)
    at parser (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:916:18)
    at IncomingMessage.res.on (/home/noels/Documents/workspace/ho/GEFEAPI/loopback-next/packages/testlab/node_modules/superagent/lib/node/parsers/json.js:19:7)
    at endReadableNT (_stream_readable.js:1059:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
Did you npm run build before running the test? Am I missing something?

bajtos: That's weird, let me try again.

bajtos: I created a feature branch based on the current master (as of today morning), adding your commit with the test (b44cac5).

You can find the feature branch here:

https://github.com/strongloop/loopback-next/tree/test-pr-3828

When I run `npm it` (install & test), all tests are green.

I think the change in behavior may be caused by https://github.com/strongloop/loopback-next/pull/3895, where we started to add `additionalProperties: false` to schema emitted for models in strict mode (which is the default).

When I change the definition of `Product` model in the acceptance test and disable strict mode, then your new test fails as expected. (Unfortunately, there is one more test that fails too.)

```patch
diff --git a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
index 19697b83f..9070273fa 100644
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -28,7 +28,7 @@ describe('Validation at REST level', () => {
   let app: RestApplication;
   let client: Client;

-  @model()
+  @model({settings: {strict: false}})
   class Product {
     @property()
     id: number;
```

Obviously, we don't want to change all existing tests to start testing the scenario with `strict: false` and no longer test the default setup with `strict: true`. To support non-strict mode, we need to create a new group of tests in `validation.acceptance.ts` that will use a new model with `strict: false`.

Let me ask you a question: how are your models configured? Do you use the default "strict" model behavior, or are you disabling it via model settings? If you are using strict mode, then I think you should be find a this pull request is not needed for your app.

bajtos:
@noels what's the status of this pull request? Are you still keen to continue working on this improvement?

Issue #3732, opened 2019-09-13
Improve `defineCrudRestController` to create a named controller class


bajtos:
_This is a follow-up for the spike #3617._

Improve `@loopback/rest-crud` - modify `defineCrudRestController` to create a named controller class, where the name is based on the model name we are creating the controller for.

Example implementation:

https://github.com/strongloop/loopback-next/blob/2848a1b3e4c62da88cfd3edc966e8267768cf326/packages/rest-crud/src/crud-rest.controller.ts#L260-L267

## Acceptance criteria

- [x] Implementation
- [x] Automated tests
- [x] Update API docs
- [x] Update the example code in `rest-crud` README

hacksparrow:
Closed via https://github.com/strongloop/loopback-next/pull/3842

PR #3842, opened 2019-09-30
feat: generate controller based on Model name

hacksparrow:
defineCrudRestController generates controller based on the Model's name.

Addresses https://github.com/strongloop/loopback-next/issues/3732.

## Checklist

### DIFF
--- a/packages/rest-crud/README.md
+++ b/packages/rest-crud/README.md
@@ -19,25 +19,20 @@ npm install --save @loopback/rest-crud
 
 2. Create a Repository class, e.g. using `lb4 repository` tool.
 
-3. Create a base REST CRUD controller class for your model.
+3. Create a REST CRUD controller class for your model.
 
    ```ts
-   const CrudRestController = defineCrudRestController<
+   const ProductController = defineCrudRestController<
      Product,
      typeof Product.prototype.id,
      'id'
    >(Product, {basePath: '/products'});
    ```
 
-4. Create a new subclass of the base controller class to configure repository
-   injection.
+4. Set up dependency injection for the ProductController.
 
    ```ts
-   class ProductController extends CrudRestController {
-     constructor(@repository(ProductRepository) repo: ProductRepository) {
-       super(repo);
-     }
-   }
+   inject('repositories.ProductRepository')(ProductController, undefined, 0);
    ```
 
 5. Register the controller with your application.
--- a/packages/rest-crud/src/__tests__/unit/crud-rest-controller.unit.ts
+++ b/packages/rest-crud/src/__tests__/unit/crud-rest-controller.unit.ts
@@ -1,0 +1,26 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest-crud
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+import {expect} from '@loopback/testlab';
+import {defineCrudRestController} from '../..';
+
+describe('defineCrudRestController', () => {
+  it('should generate controller based on Model name', async () => {
+    @model()
+    class Product extends Entity {
+      @property({id: true})
+      id: number;
+    }
+
+    const CrudRestController = defineCrudRestController<
+      Product,
+      typeof Product.prototype.id,
+      'id'
+    >(Product, {basePath: '/products'});
+
+    expect(CrudRestController.name).to.equal('ProductController');
+  });
+});
--- a/packages/rest-crud/src/crud-rest.controller.ts
+++ b/packages/rest-crud/src/crud-rest.controller.ts
@@ -32,6 +32,7 @@ import {
   ResponsesObject,
   SchemaObject,
 } from '@loopback/rest';
+import assert = require('assert');
 
 // Ideally, this file should simply `export class CrudRestController<...>{}`
 // Unfortunately, that's not possible for several reasons.
@@ -107,17 +108,17 @@ export interface CrudRestControllerOptions {
  * Example usage:
  *
  * ```ts
- * const CrudRestController = defineCrudRestController<
- *   Product,
- *   typeof Product.prototype.id,
- *   'id'
+ * const ProductController = defineCrudRestController<
+ * Product,
+ * typeof Product.prototype.id,
+ * 'id'
  * >(Product, {basePath: '/products'});
  *
- * class ProductController extends CrudRestController {
- *   constructor() {
- *    super(repo);
- *   }
- * }
+ * inject('repositories.ProductRepository')(
+ *  ProductController,
+ *   undefined,
+ *   0,
+ * );
  *
  * app.controller(ProductController);
  * ```
@@ -257,8 +258,14 @@ export function defineCrudRestController<
     }
   }
 
-  // See https://github.com/microsoft/TypeScript/issues/14607
-  return CrudRestControllerImpl;
+  const controllerName = modelName + 'Controller';
+  const defineNamedController = new Function(
+    'controllerClass',
+    `return class ${controllerName} extends controllerClass {}`,
+  );
+  const controller = defineNamedController(CrudRestControllerImpl);
+  assert.equal(controller.name, controllerName);
+  return controller;
 }
 
 function getIdSchema<T extends Entity>(


bajtos:
LGTM so far.

Please update API docs and the example code in rest-crud README per acceptance criteria in the linked issue.

IMO, this is a feature, not a bug fix, because the current code works exactly as intended. Let's change the commit message type to `feat` please.

bajtos:
I can see how `CrudRestController` can be confusing. Should we use `CrudRestControllerImpl` instead? I feel that `controllerClassName` is misleading because the argument is set to a class constructor, not the class name. `controllerClass` is fine with me though.

packages/rest-crud/README.md
```diff
@@ -19,25 +19,20 @@ npm install --save @loopback/rest-crud
...
+   const ProductController = defineCrudRestController<
      Product,
      typeof Product.prototype.id,
      'id'
    >(Product, {basePath: '/products'});
    ```
 
-4. Create a new subclass of the base controller class to configure repository
-   injection.
+4. Inject the model's repository via Dependency Injection.
```
raymondfeng: It should be `Set up dependency injection for the ProductController`.

packages/rest-crud/README.md
```diff
@@ -19,25 +19,20 @@ npm install --save @loopback/rest-crud
...
 
    ```ts
-   class ProductController extends CrudRestController {
-     constructor(@repository(ProductRepository) repo: ProductRepository) {
-       super(repo);
-     }
-   }
+   inject('repositories.ProductRepository')(ProductController, undefined, 0);
```
raymondfeng: Why not using the subclassing for dependency injection? It seems to be easier to understand.

hacksparrow: > It seems to be easier to understand.

Agreed. However, there some redundancy going on now.

Earlier `defineCrudRestController` used to return a `CrudRestControllerImpl` class which was assigned to `CrudRestController`, so it made sense to do `class ProductController extends CrudRestController`.

Now `defineCrudRestController` returns a named controller class like `ProductController`. Doing `class MyProductController extends ProductController`, looks weird. "Why can't I just use `ProductController`?"

Alternative would something like:
```
const ProductControllerBase = defineCrudRestController<
     Product,
     typeof Product.prototype.id,
     'id'
>(Product, {basePath: '/products'});

class ProductController extends ProductControllerBase {}
```

bajtos: The problem with subclassing is naming - the class created by `defineCrudRestController` already has the right name (`ProductController`). How should the new class be called? Should we use an anonymous class?

```ts
const controller = const ProductController = defineCrudRestController<
  Product,
  typeof Product.prototype.id,
  'id'
>(Product, {basePath: '/products'});

const ProductController = class extends controller {
  constructor(
    // args decorated with @inject
  ) {
    super(/*forward args*/);
  }
}
```

I am also concerned about micro-performance - by creating a new class, we are adding more code to be executed for every incoming request.

Anyhow, these instructions will soon be moved to "Advanced use", most LB users are going to use ModelApiBooter that will be introduced by one of the next tasks in this Epic.

raymondfeng: If we always expect to inject a repository, we can either make the repository class an argument of `defineCrudRestController()` or create a helper function such as:

```ts
setupRepositoryInjection(ProductController, 'repositories.ProductRepository');
```

hacksparrow: I am in favor of making the repository class an argument of `defineCrudRestController` makes sense to me. @bajtos thoughts?

hacksparrow: @raymondfeng I experimented the "repository class argument" approach:

1. Passing a binding key would have been best, but since we have no access to the context, it won't work.
2. Passing a repository instance requires the repository to be setup with datasource. It brings in the datasource factor, meaning more work and complexity.

We don't always expect to inject a repository, though. Eg: `class ProductController extends ProductControllerBase {}`.

hacksparrow: I am neutral about `setupRepositoryInjection`, will go with @bajtos's opinion.

raymondfeng: @hacksparrow Two points:

1. No context is needed. We only require the binding key or class for the repository so that we can have:

```ts
export function defineCrudRestController<
  T extends Entity,
  IdType,
  IdName extends keyof T,
  R extends EntityCrudRepository<T, IdType, Relations>,
  Relations extends object = {}
>(
  modelCtor: typeof Entity & {prototype: T & {[key in IdName]: IdType}},
  repoCtor: R,
  options: CrudRestControllerOptions,
): CrudRestControllerCtor<T, IdType, IdName, Relations> {
  // ...
  @api({basePath: options.basePath, paths: {}})
  class CrudRestControllerImpl
    implements CrudRestController<T, IdType, IdName> {
    constructor(
      public readonly @repository(ProductRepository): R,
    ) {}
```

2. Without the repository injection setup, the controller class is not functional.

bajtos: I am strongly against setting up DI in `defineCrudRestController`. It's a code smell when DI-related constructs are littered across the code base. The class created by `defineCrudRestController` should be usable without DI too, it is the responsibility of the caller to decide how they want to supply the needed repository instance.

In some cases, the user may not want to inject the repository via DI, they may want to receive the datasource instance via DI instead and create a Repository instance on the fly, for example by calling `new DefaultCrudRepository(Product, ds)`.

IMO, this discussion has uncovered a major design limitation of our IoC container, where it's not possible to wire up dependencies from outside. In a well-designed app, it's the app that defines how are dependencies resolved, not the individual classes!

Ideally, I would like to wire my IoC setup along the following lines (a low-level example, we would want a nicer API in a real implementation):

```ts
const ProductController = defineCrudRestController(...);
const binding = app.controller(ProductController);
binding.injectConstructorArg(0, 'repositories.ProductRepository');
```

bajtos: I am proposing to land the current version and open a new issue or a new pull request to discuss how to improve the example code snippet.

As I mentioned above, this code snippet will be used by a very small fraction of our users, let's invest our time into improvements with bigger impact.

hacksparrow:
Follow up issue for doc improvement - https://github.com/strongloop/loopback-next/issues/3862

raymondfeng:
> I am strongly against setting up DI in defineCrudRestController. It's a code smell when DI-related constructs are littered across the code base. The class created by defineCrudRestController should be usable without DI too, it is the responsibility of the caller to decide how they want to supply the needed repository instance.

Setting up DI does not mandate DI to use the controller class. It's just the metadata to enable DI. The `defineCrudRestController` already leverages decorators in the base class for REST mapping. My expectation is that once the controller class is created via `defineCrudRestController`, it should be ready to be mounted using `app.controller(...)`. That's the whole purpose to generate REST CRUD APIs by convention from a model definition and corresponding datasource.

raymondfeng:
> IMO, this discussion has uncovered a major design limitation of our IoC container, where it's not possible to wire up dependencies from outside. In a well-designed app, it's the app that defines how are dependencies resolved, not the individual classes!

Not really. The `@inject` is used to declare dependencies to be injected, NOT the how the dependencies should be resolved. We use binding keys as the connection for most cases, but other criteria (such as a binding filter for the interface) can be leveraged to further decouple. Only the individual classes has the knowledge about required dependencies for the business logic. It's the responsibility for the application to fulfill such dependencies. Our IoC framework magically glues the provider and consumer together leveraging the DI metadata.

bajtos:
> >```ts
> > const binding = app.controller(ProductController);
> > binding.injectConstructorArg(0, 'repositories.ProductRepository');
> > ```
>
> This is wrong. Please note `ProductController` can be bound to many bindings. Changing the `@inject` from a binding will impact other other bindings that use the same class.

Please note that my example code is modifying the injected constructor argument only for the single binding returned by `app.controller`. Other bindings using `ProductController` won't be affected.

Here is another example to illustrate what I mean:

```ts
// setup logger used during request handling
app.bind('request.logger').to(ConsoleLogger).injectConstructorArg(0, {
  // print info, warnings and errors 
  level: LogLevel.INFO,
});

// setup logger used during app bootstrap
app.bind('app.logger').to(ConsoleLogger).injectConstructorArg(0, {
  // we don't care about INFO level, want to see only problems
  level: LogLevel.WARN,
});
```

Issue #3436, opened 2019-07-26
Run repository tests for PostgreSQL


bajtos:
Add `acceptance/repository-postgresql` package, modeled after MySQL tests in `acceptance/repository-mysql`. The tests should be run as a new Travis CI job, see the existing setup for other connectors.

See also https://github.com/strongloop/loopback-next/pull/3387

## Acceptance criteria

- [ ] A new directory `acceptance/repository-postgresql` running all CRUD tests from `repository-tests` packages
- [ ] A new matrix entry in Travis CI config to run PostgreSQL tests as part of our CI builds
- [ ] All tests are passing

agnes512:
see @emonddr for help :D

PR #3853, opened 2019-10-01
feat(test-repository-postgresql): run repository tests for postgresql

nabdelgadir:
Resolves https://github.com/strongloop/loopback-next/issues/3436.

Added `acceptance/repository-postgresql` package in order for `repository-tests` to be run for PostgreSQL.

## Checklist


acceptance/repository-postgresql/src/__tests__/postgresql.datasource.ts
```diff
@@ -0,0 +1,25 @@
...
+  username: process.env.POSTGRESQL_USER || 'root',
+  password: process.env.POSTGRESQL_PASSWORD || 'pass',
+  createDatabase: true,
+};
+
+export const POSTGRESQL_FEATURES: Partial<CrudFeatures> = {
+  idType: 'number',
+  freeFormProperties: false,
+  emptyValue: null,
+  supportsTransactions: false,
```
bajtos: Are you sure that our PostgreSQL connector does not support transactions? The database certainly does support them.

nabdelgadir: I was using `DefaultCrudRepository` before by accident, but I've changed this to `true` now.

acceptance/repository-postgresql/src/__tests__/postgresql.datasource.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import {CrudFeatures, DataSourceOptions} from '@loopback/repository-tests';
+
+const connector = require('loopback-connector-postgresql');
+
+export const POSTGRESQL_CONFIG: DataSourceOptions = {
+  connector,
+  host: process.env.POSTGRESQL_HOST || 'localhost',
+  port: process.env.POSTGRESQL_PORT || 5432,
+  database: process.env.POSTGRESQL_DATABASE || 'testdb',
```
bajtos: Please use `repository-tests`, the same value we use in other LB4 repository-acceptance tests.

nabdelgadir: According to the [documentation](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), it can't include a `-` so I'll use `repository_tests`. Should I modify the others to be consistent with this one?

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -140,13 +162,19 @@ export function transactionSuite(
...
           ds2Options.name = 'anotherDataSource';
-          ds2Options.database = ds2Options.database + '-new';
+          ds2Options.database = ds2Options.database + '_new';
+
+          // need to manually create the database for PostgreSQL otherwise it
+          // complains that database doesn't exist
+          if (ds.connector && ds.connector.name === 'postgresql') {
+            await ds.execute(`DROP DATABASE IF EXISTS ${ds2Options.database};`);
+            await ds.execute(`CREATE DATABASE ${ds2Options.database};`);
+          }
```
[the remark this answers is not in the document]
nabdelgadir: I believe it's because it's using the same model. If we create a new repository class with the same datasource and migrate it, it'll use the same table for both repositories, from my understanding.

That being said, I'll move the database setup to the travis setup.

nabdelgadir:
Re the coverage drop: the transactions test suite isn't run by the memory database, but it's run by `acceptance/mysql` and `acceptance/postgresql`.

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -89,6 +91,16 @@ export function transactionSuite(
...
+
+          // temporary workaround for "Release called on client which has
+          // already been released to the pool." for PostgreSQL until
+          // tx.isActive is available
+          //FIXME: remove this once tx.isActive becomes available
+          // see https://github.com/strongloop/loopback-next/issues/3471
+          if (ds.connector && ds.connector.name === 'postgresql') {
+            // eslint-disable-next-line require-atomic-updates
+            tx = undefined;
+          }
```
raymondfeng: Can we wrap this code into a function as it happens 3 times?

packages/repository-tests/src/crud/transactions.suite.ts
```diff
@@ -169,7 +174,20 @@ export function transactionSuite(
...
         });
+
+        // temporary workaround for "Release called on client which has
+        // already been released to the pool." for PostgreSQL until
+        // tx.isActive is available
+        //FIXME: remove this once tx.isActive becomes available
+        // see https://github.com/strongloop/loopback-next/issues/3471
+        function setTxToUndefined(t: Transaction | undefined) {
+          if (ds.connector && ds.connector.name === 'postgresql') {
+            t = undefined;
```
raymondfeng: This is not working correctly as `t = undefined` inside the function won't change the value of `tx` argument from the caller. You can return `undefined` or `t` conditionally and use it  as `tx = setTxToUndefined(tx)`.

raymondfeng: FYI: this can be further simplified as:
```ts
return ds.connector && ds.connector.name === 'postgresql' ? undefined: t;
```

acceptance/repository-postgresql/src/__tests__/postgresql.datasource.ts
bajtos: > I'll use `repository_tests`. 

👍 

> Should I modify the others to be consistent with this one?

I don't mind either way, please decide yourself.

.travis.yml
```diff
@@ -67,6 +67,20 @@ matrix:
...
+        - TASK=test-repository-postgresql
+        - POSTGRESQL_USER=postgres
+      services:
+        - postgresql
+      script:
+        - npm run postinstall -- --scope "@loopback/test-repository-postgresql" --include-filtered-dependencies
+        - npm run build -- --scope "@loopback/test-repository-postgresql" --include-filtered-dependencies
+        - psql -U postgres -c "CREATE DATABASE repository_tests;"
+        - psql -U postgres -c "CREATE DATABASE repository_tests_new;"
+        - psql -U postgres -c "\c repository_tests;"
```
bajtos: What happens when we run the test suite locally? Will the databases `repository_tests` and `repository_tests_new` be created too?

IIRC, the `setup` script will run all commands in `acceptance/repository-postgresql/test/schema.sql`, should we perhaps move creation of databases to that file and then execute it from `.travis.yml` too, to avoid duplication?

packages/repository-tests/src/crud/transactions.suite.ts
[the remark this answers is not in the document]
bajtos: I still find this test confusing, but I guess it's not the job of this pull request to fix that. I like that you moved the database setup to Travis CI build steps, I think this thread can be considered as resolved 👍

acceptance/repository-postgresql/README.md
```diff
@@ -0,0 +1,66 @@
...
+  installed, run the following script which would spawn a PostgreSQL instance on
+  your local:
+
+  ```bash
+  source setup.sh <HOST> <PORT> <USER> <PASSWORD> <DATABASE>
+  ```
+
+  Where `<HOST>`, `<PORT>`, `<USER>`, `<PASSWORD>` and `<DATABASE>` are optional
+  parameters. The default values are `localhost`, `5432`, `root`, `pass` and
+  `testdb` respectively.
```
bajtos: I think `testdb` is not accurate, because we use `repository_tests`? How about the second database, `repository_tests_new`?

bajtos:
BTW I find it weird that the test is failing with a timeout, I would expect it to throw an error saying that the configured database does not exist. This is probably a bug in our PostgreSQL connector, therefore out of scope of this pull request, but possible worth fixing or at least opening a GH issue for.

acceptance/repository-postgresql/README.md
nabdelgadir: I think when you're running it locally via the setup script, it uses `testdb`. It only creates `repository_tests` if the environment variable isn't set through the setup script. Also, this is in line with [`mysql`](https://github.com/strongloop/loopback-next/commit/e9dca4cbcc8ebf4e2567cebdc046f3cf1d886060#diff-7ec3ca7c559cb1be02f07d4bf8ccef34R51).

emonddr:
> BTW I find it weird that the test is failing with a timeout, I would expect it to throw an error saying that the configured database does not exist. This is probably a bug in our PostgreSQL connector, therefore out of scope of this pull request, but possibly worth fixing or at least opening a GH issue for.

@nabdelgadir  ^ please respond to this comment

emonddr:
@nabdelgadir 

Regarding

> Re the coverage drop: the transactions test suite isn't run by the memory database, but it's run by acceptance/mysql and acceptance/postgresql. 

Do you have a proposed fix (new set of tests) to bring the coverage back up to an appropriate level?

Right now, the CI is complaining

image

acceptance/repository-postgresql/README.md
nabdelgadir: I'll update this to `repository_tests` as it's set up in `schema.sql`

nabdelgadir:
> BTW I find it weird that the test is failing with a timeout, I would expect it to throw an error saying that the configured database does not exist. This is probably a bug in our PostgreSQL connector, therefore out of scope of this pull request, but possibly worth fixing or at least opening a GH issue for.

@bajtos I tried creating an app to reproduce this error, and it actually does throw a descriptive error (e.g. `error: database "testdb" does not exist`) and it throws it if I run that test individually. So the connector is throwing the error properly. It might be hard to resolve this issue if it needs this specific setup in order to be tested.

nabdelgadir:
> Regarding
> 
> > Re the coverage drop: the transactions test suite isn't run by the memory database, but it's run by acceptance/mysql and acceptance/postgresql.
> 
> Do you have a proposed fix (new set of tests) to bring the coverage back up to an appropriate level?
> 
> Right now, the CI is complaining
> 
> image

@emonddr Not really sure how to bring it back up because as I said the coverage only accounts for tests run by the memory database and the memory database doesn't support transactions. So the coverage for the [file](https://github.com/strongloop/loopback-next/pull/3853/files#diff-5b8b9c9868638cbf9ea2dfbbee7b1593) it's complaining about can't be brought back up. But do you have something in mind?

.travis.yml
```diff
@@ -30,7 +30,7 @@ matrix:
       env: TASK=code-lint
       # Running Code Linter -- Requires @loopback/build so it's bootstrapped
       script:
-        - lerna bootstrap --scope @loopback/build --include-filtered-dependencies
+        - lerna bootstrap --scope @loopback/build --include-dependencies
```
nabdelgadir: Screen Shot 2019-10-18 at 12 21 21 AM

Replacing because `--include-filtered-dependencies` has been renamed.

acceptance/repository-postgresql/test/schema.sql
```diff
@@ -0,0 +1,8 @@
+-- Seed database schema & data
+-- See https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/schema.sql
+
+DROP DATABASE IF EXISTS repository_tests;
+CREATE DATABASE repository_tests;
+
+DROP DATABASE IF EXISTS repository_tests_new;
```
agnes512: Do we still need the second database if we can have `repository_tests` created?

nabdelgadir: Yeah, we still need to manually create it for [this test]( https://github.com/strongloop/loopback-next/blob/postgresql-tests/packages/repository-tests/src/crud/transactions.suite.ts#L144-L149).

emonddr: > > Regarding
> > > Re the coverage drop: the transactions test suite isn't run by the memory database, but it's run by acceptance/mysql and acceptance/postgresql.
> > 
> > 
> > Do you have a proposed fix (new set of tests) to bring the coverage back up to an appropriate level?
> > Right now, the CI is complaining
> > image
> 
> @emonddr Not really sure how to bring it back up because as I said the coverage only accounts for tests run by the memory database and the memory database doesn't support transactions. So the coverage for the [file](https://github.com/strongloop/loopback-next/pull/3853/files#diff-5b8b9c9868638cbf9ea2dfbbee7b1593) it's complaining about can't be brought back up. But do you have something in mind?

@bajtos @nabdelgadir , should we use

https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

statements for the code block you are mentioning in the file link above?

Or is there a better solution?

nabdelgadir: Thanks, I updated to use this!

emonddr:
@slnode test please

emonddr:
@nabdelgadir , did you run the tests locally and all tests passed? Referring to @bajtos comment in https://github.com/strongloop/loopback-next/pull/3853#pullrequestreview-299993986 .

nabdelgadir:
> @nabdelgadir , did you run the tests locally and all tests passed? Referring to @bajtos comment in [#3853 (review)](https://github.com/strongloop/loopback-next/pull/3853#pullrequestreview-299993986) .

Yes, they work locally.

PR #3858, opened 2019-10-02
[WIP] get ready for Typescript 3.7

raymondfeng:
See https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/

Blocked by https://github.com/microsoft/TypeScript/issues/33739

## Checklist

### DIFF
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -104,9 +104,7 @@ export class DefaultCrudRepository<
 
   public readonly inclusionResolvers: Map<
     string,
-    // FIXME(rfeng): TypeScript does not allow us to define any subtype of Entity
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    InclusionResolver<T, any>
+    InclusionResolver<T, Entity>
   > = new Map();
 
   /**
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -104,7 +104,9 @@ export class DefaultCrudRepository<
 
   public readonly inclusionResolvers: Map<
     string,
-    InclusionResolver<T, Entity>
+    // FIXME(rfeng): TypeScript does not allow us to define any subtype of Entity
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    InclusionResolver<T, any>
   > = new Map();
 
   /**
@@ -505,9 +507,9 @@ export class DefaultCrudRepository<
    * @param relationName - Name of the relation defined on the source model
    * @param resolver - Resolver function for getting related model entities
    */
-  registerInclusionResolver(
+  registerInclusionResolver<R extends Entity>(
     relationName: string,
-    resolver: InclusionResolver<T, Entity>,
+    resolver: InclusionResolver<T, R>,
   ) {
     this.inclusionResolvers.set(relationName, resolver);
   }
--- a/docs/site/Testing-your-application.md
+++ b/docs/site/Testing-your-application.md
@@ -129,7 +129,7 @@ helper method; for example:
 {% include code-caption.html content="src/__tests__/helpers/database.helpers.ts" %}
 
 ```ts
-import {ProductRepository, CategoryRepository} from '../../src/repositories';
+import {ProductRepository, CategoryRepository} from '../../repositories';
 import {testdb} from '../fixtures/datasources/testdb.datasource';
 
 export async function givenEmptyDatabase() {
@@ -145,7 +145,7 @@ belongs to Category, include it in the repository call, for example:
 
 ```ts
 import {Getter} from '@loopback/context';
-import {ProductRepository, CategoryRepository} from '../../src/repositories';
+import {ProductRepository, CategoryRepository} from '../../repositories';
 import {testdb} from '../fixtures/datasources/testdb.datasource';
 
 export async function givenEmptyDatabase() {
@@ -503,8 +503,8 @@ import {
   sinon,
   StubbedInstanceWithSinonAccessor,
 } from '@loopback/testlab';
-import {ProductRepository} from '../../../src/repositories';
-import {ProductController} from '../../../src/controllers';
+import {ProductRepository} from '../../../repositories';
+import {ProductController} from '../../../controllers';
 
 describe('ProductController (unit)', () => {
   let repository: StubbedInstanceWithSinonAccessor<ProductRepository>;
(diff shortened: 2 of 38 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -104,7 +104,9 @@ export class DefaultCrudRepository<
 
   public readonly inclusionResolvers: Map<
     string,
-    InclusionResolver<T, Entity>
+    // FIXME(rfeng): TypeScript does not allow us to define any subtype of Entity
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    InclusionResolver<T, any>
```
bajtos: This is weird, can you (or somebody else) take another look please? I'd like us to better understand why is TypeScript not allowing such construct.

raymondfeng: What we really wants is `InclusionResolver<T, any type that extends Entity>` but I cannot find such typing in TypeScript. 

With `InclusionResolver<T, Entity>`, `this.inclusionResolvers.set('orders', resolver)` fails to compile if `resolver` is `InclusionResolver<T, Order>` as it cannot be assigned to `InclusionResolver<T, Entity>`.

bajtos: Here is the error message:

```
src/repositories/legacy-juggler-bridge.ts(514,47): error TS2345: Argument of type 'InclusionResolver<T, R>' is not assignable to parameter of type 'InclusionResolver<T, Entity>'.
  Types of parameters 'inclusion' and 'inclusion' are incompatible.
    Type 'Inclusion<Entity>' is not assignable to type 'Inclusion<R>'.
      Types of property 'scope' are incompatible.
        Type 'Filter<Entity> | undefined' is not assignable to type 'Filter<R> | undefined'.
          Type 'Filter<Entity>' is not assignable to type 'Filter<R>'.
```

This looks like a regression in TypeScript to me. 

Notice that on the first line, we are trying to assign `InclusionResolver<T, R>` to `InclusionResolver<T, Entity>`. On the third line, the order of `R` and `Entity` is swapped and the compiler complains about assigning `Inclusion<Entity>` to `Inclusion<R>`!

bajtos: I managed to reproduce the problem, it's triggered by code in `FilterBuilder.impose`. Long live spooky actions at a distance ☹️

bajtos: Here is the code reproducing the error (compile it via `tsc --strict`). I think there is a bug in the compiler that causes a misleading error to be reported. At the same time, I believe the compiler has discovered a genuine problem in our type definitions. I'll be pushing a new commit to address that.

```ts
// IMPORTANT! This function must be processed by the compiler first,
// before DefaultCrudRepository is compiled

function breakTheBuild<MT extends object>(
  constraint: Filter<MT> | Condition<MT>,
) {
  // COMMENT-OUT THE FOLLOWING LINE TO MAKE THE COMPILER HAPPY
  if (!isFilter(constraint)) constraint = {where: constraint};
}


class DefaultCrudRepository<
  T extends Entity,
  ID,
  Relations extends object = {}
> {
  public readonly inclusionResolvers: Map<
    string,
    InclusionResolver<T, Entity>
  > = new Map();

  registerInclusionResolver<R extends Entity>(
    relationName: string,
    resolver: InclusionResolver<T, R>,
  ) {
    // THE COMPILER ERROR IS TRIGGERED HERE
    // error TS2345: Argument of type 'InclusionResolver<T, R>' is not
    // assignable to parameter of type 'InclusionResolver<T, Entity>'.
    //   Types of parameters 'inclusion' and 'inclusion' are incompatible.
    //     Type 'Inclusion<Entity>' is not assignable to type 'Inclusion<R>'.
    //       Types of property 'scope' are incompatible.
    //         Type 'Filter<Entity> | undefined' is not assignable to type 'Filter<R> | undefined'.
    //           Type 'Filter<Entity>' is not assignable to type 'Filter<R>'.
    //             Type 'Entity' is not assignable to type 'R'.
    //                'Entity' is assignable to the constraint of type 'R',
    //                but 'R' could be instantiated with a different subtype of constraint 'Entity'.
    this.inclusionResolvers.set(relationName, resolver);
  }
}



class Entity {
  // base class for ORM entities (models)
  // ...
}

type InclusionResolver<S extends Entity, T extends Entity> = (
  /**
   * List of source models as returned by the first database query.
   */
  sourceEntities: S[],
  /**
   * Inclusion requested by the user (e.g. scope constraints to apply).
   */
  inclusion: Inclusion<T>,
) => Promise<(T | undefined)[] | (T[] | undefined)[]>;


/**
 * Operators for where clauses
 */
export type Operators =
  | 'eq' // Equal
  | 'neq' // Not Equal
  // etc.
;

/**
 * Matching predicate comparison
 */
export type PredicateComparison<PT> = {
  eq?: PT;
  neq?: PT;
  // etc.
};

/**
 * Key types of a given model, excluding operators
 */
export type KeyOf<MT extends object> = Exclude<
  Extract<keyof MT, string>,
  Operators
>;

/**
 * Condition clause
 *
 * @example
 * ```ts
 * {
 *   name: {inq: ['John', 'Mary']},
 *   status: 'ACTIVE',
 *   age: {gte: 40}
 * }
 * ```
 */
export type Condition<MT extends object> = {
  [P in KeyOf<MT>]?: PredicateComparison<MT[P]> // {x: {lt: 1}}
};


/**
 * Inclusion of related items
 *
 * Note: scope means filter on related items
 *
 * Example:
 * `{relation: 'aRelationName', scope: {<AFilterObject>}}`
 */
export interface Inclusion<MT extends object> {
  relation: string;
  scope?: Filter<MT>;
}

/**
 * Query filter object
 */
export interface Filter<MT extends object> {
  /**
   * The matching criteria
   */
  where?: Condition<MT>;

  /**
   * To include related objects
   */
  include?: Inclusion<MT>[];
}

/**
 * TypeGuard for Filter
 * @param candidate
 */
export function isFilter<MT extends object>(
  candidate: any,
): candidate is Filter<MT> {
  return true; // the actual implementation is not relevant
}
```

bajtos: See 40e4bd4

bajtos: Opened a PR to fix this problem independently, see #4037

bajtos:
@raymondfeng I think the following two commits should be easy to cherry-pick and land on master independently of this patch: https://github.com/strongloop/loopback-next/pull/3858/commits/3f1fed03103c3c80d8c403fb10fb243fe56b0963 and https://github.com/strongloop/loopback-next/pull/3858/commits/005e7f131b6f447bc0c6dbba26f496bbac1c267b.

The third commit https://github.com/strongloop/loopback-next/pull/3858/commits/46132b72e4c151074267475038cac207deadd191 is tricky because it mixes different fixes, I think we need to split it in smaller parts and land only those that are relevant.

raymondfeng:
I'm going to open a new PR.

Issue #3733, opened 2019-09-13
Add `defineCrudRepositoryClass` - a helper to create a named repository class


bajtos:
_This is a follow-up for the spike #3617._

Add `defineCrudRepositoryClass` to `@loopback/rest-crud`, this function should create a named repository class for the given Model class.

Example implementation (notice wrong function name):

https://github.com/strongloop/loopback-next/blob/2848a1b3e4c62da88cfd3edc966e8267768cf326/packages/rest-crud/src/repository-builder.ts#L14-L36

## Acceptance criteria

- [ ] Implementation
- [ ] Automated tests
- [ ] API docs
- [ ] Add a code example to `rest-crud` README

hacksparrow:
Closed via https://github.com/strongloop/loopback-next/pull/3867

PR #3867, opened 2019-10-04
feat: add defineCrudRepositoryClass

hacksparrow:
Add `defineCrudRepositoryClass` - a helper to create named repository classes.

Addresses https://github.com/strongloop/loopback-next/issues/3733.

## Checklist

### DIFF
--- a/packages/rest-crud/README.md
+++ b/packages/rest-crud/README.md
@@ -4,8 +4,8 @@ REST API controller implementing default CRUD semantics.
 
 ## Overview
 
-This module allows applications to quickly expose a model via REST API without
-having to implement a custom controller class.
+This module allows applications to quickly expose models via REST API without
+having to implement custom controller or repository classes.
 
 ## Installation
 
@@ -15,11 +15,19 @@ npm install --save @loopback/rest-crud
 
 ## Basic use
 
-1. Define your model class, e.g. using `lb4 model` tool.
+`@loopback/rest-crud` exposes two helper methods (`defineCrudRestController` and
+`defineCrudRepositoryClass`) for creating controllers and respositories using
+code.
 
-2. Create a Repository class, e.g. using `lb4 repository` tool.
+For the examples in the following sections, we are assuming a model named
+`Product`, and a datasource named `db` have already been created.
 
-3. Create a REST CRUD controller class for your model.
+### Creating a CRUD Controller
+
+Here is how you would use `defineCrudRestController` for exposing the CRUD
+endpoints of an existing model with a respository.
+
+1. Create a REST CRUD controller class for your model.
 
    ```ts
    const ProductController = defineCrudRestController<
@@ -29,17 +37,64 @@ npm install --save @loopback/rest-crud
    >(Product, {basePath: '/products'});
    ```
 
-4. Set up dependency injection for the ProductController.
+2. Set up dependency injection for the ProductController.
 
    ```ts
    inject('repositories.ProductRepository')(ProductController, undefined, 0);
    ```
 
-5. Register the controller with your application.
+3. Register the controller with your application.
 
    ```ts
    app.controller(ProductController);
    ```
+
+### Creating a CRUD repository
+
+Use the `defineCrudRepositoryClass` method to create named repositories (based
+on the Model) for your app.
+
+Usage example:
+
+```ts
+const ProductRepository = defineCrudRepositoryClass(Product);
+this.repository(ProductRepository);
+inject('datasources.db')(ProductRepository, undefined, 0);
+```
+
+### Integrated example
+
+Here is an example of an app which uses `defineCrudRepositoryClass` and
+`defineCrudRestController` to fulfill its repository and controller
+requirements.
+
+```ts
+export class TryApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options: ApplicationConfig = {}) {
+    ...
+  }
+
+  async boot():Promise<void> {
+    await super.boot();
+
+    const ProductRepository = defineCrudRepositoryClass(Product);
+    const repoBinding = this.repository(ProductRepository);
+
+    inject('datasources.db')(ProductRepository, undefined, 0);
+
+    const ProductController = defineCrudRestController<
+    Product,
+    typeof Product.prototype.id,
+    'id'
+    >(Product, {basePath: '/products'});
+
+    inject(repoBinding.key)(ProductController, undefined, 0);
+    this.controller(ProductController);
+  }
+}
+```
 
 ## Contributions
 
--- a/packages/rest-crud/src/repository-builder.ts
+++ b/packages/rest-crud/src/repository-builder.ts
@@ -1,0 +1,55 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest-crud
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  DefaultCrudRepository,
+  Entity,
+  EntityCrudRepository,
+  juggler,
+} from '@loopback/repository';
+import * as assert from 'assert';
+
+/**
+ * Create (define) a repository class for the given model.
+ *
+ * Example usage:
+ *
+ * ```ts
+ * const ProductRepository = defineCrudRepositoryClass(Product);
+ * ```
+ *
+ * @param modelCtor A model class, e.g. `Product`.
+ */
+export function defineCrudRepositoryClass<
+  T extends Entity,
+  IdType,
+  Relations extends object = {}
+>(
+  entityClass: typeof Entity & {prototype: T},
+): RepositoryClass<T, IdType, Relations> {
+  const repoName = entityClass.name + 'Repository';
+  const defineNamedRepo = new Function(
+    'EntityCtor',
+    'BaseRepository',
+    `return class ${repoName} extends BaseRepository {
+      constructor(dataSource) {
+        super(EntityCtor, dataSource);
+      }
+    };`,
+  );
+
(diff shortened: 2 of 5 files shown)


bajtos:
@hacksparrow I am confused. The pull request says "Add defineCrudRepositoryClass - a helper to create named repository classes", but then it changes 24 files and introduces also a booter class. Can you please check?

hacksparrow:
@bajtos PTAL, updated.

packages/rest-crud/src/repository-builder.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import {
+  DefaultCrudRepository,
+  Entity,
+  EntityCrudRepository,
+  juggler,
+} from '@loopback/repository';
+import * as assert from 'assert';
+
+export function defineCrudRepositoryClass<
```
bajtos: Missing API docs, please add.

packages/rest-crud/README.md
```diff
@@ -41,6 +41,19 @@ npm install --save @loopback/rest-crud
...
+
+A helper method `defineCrudRepositoryClass` is exported by this module, which
+can be used for creating named repositories (based on the Model).
+
+Usage example:
+
+```ts
+const db = new juggler.DataSource({connector: 'memory'});
+const ProductRepository = defineCrudRepositoryClass(Product);
+const repo = new ProductRepository(db);
```
bajtos: Can you please show how to register the created repository class for dependency injection?

I think we need to:

1. Setup dependency injection by calling `inject` - see the instructions for the controller
2. Call `app.repository(ProductRepository)`

I think it would be better to merge this new content with the steps 1-5 above, perhaps rework the step 2 "Create a Repository class, e.g. using `lb4 repository` tool." to show how to use `defineCrudRepositoryClass`?

If you prefer to keep the instructions for the controller as they are now, then I am fine with that too, but please improve the overall structure, e.g. by introducing two subsections to `Basic use`, one for controllers and another for repositories.

packages/rest-crud/src/repository-builder.ts
bajtos: I am afraid this is not resolved yet. By API docs I mean a `tsdoc` comment, see what we have for the controller factory.

packages/rest-crud/src/repository-builder.ts
```diff
@@ -0,0 +1,55 @@
...
+  Relations extends object = {}
+>(
+  entityClass: typeof Entity & {prototype: T},
+): RepositoryClass<T, IdType, Relations> {
+  const repoName = entityClass.name + 'Repository';
+  const defineNamedRepo = new Function(
+    'EntityCtor',
+    'BaseRepository',
+    `return class ${repoName} extends BaseRepository {
+      constructor(dataSource) {
```
raymondfeng: I would like to make it possible to set up `@inject` here too.

hacksparrow: What should it be able to `@inject`?

packages/rest-crud/README.md
hacksparrow: @bajtos I have added a new section **Integrated example**.

packages/rest-crud/src/repository-builder.ts
raymondfeng: Injection of the `datasource`, for example:

```ts
class ${repoName} extends BaseRepository {
      constructor(@inject(${dataSourceBinding}) dataSource: DataSource) {}
}

File: packages/rest-crud/README.md [hunk 4]
```diff
@@ -41,6 +41,67 @@ npm install --save @loopback/rest-crud
...
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options: ApplicationConfig = {}) {
+    ...
+  }
+
+  async boot():Promise<void> {
+    await super.boot();
+
+    const db = new juggler.DataSource({connector: 'memory'});
```
bajtos: Please remove, the code should use the datasource loaded by `super.boot()`.

packages/rest-crud/README.md
```diff
@@ -41,6 +41,67 @@ npm install --save @loopback/rest-crud
...
+    const ProductRepository = defineCrudRepositoryClass(Product);
+    const repo = new ProductRepository(db);
```
bajtos: Please remove, there should be a new repository instance created for each incoming request, the instance will be created by LoopBack Context as part of dependency resolution/injection.

packages/rest-crud/README.md
```diff
@@ -41,6 +41,67 @@ npm install --save @loopback/rest-crud
...
+
+    inject('datasources.db')(ProductRepository, undefined, 0);
+    this.repository(ProductRepository);
+    this.bind('repositories.ProductRepository').to(repo);
```
bajtos: Please remove this line, the binding is already created by `this.repository(ProductRepository)`.

To avoid magic strings, you may want to get the binding key from the binding created by `this.repository` when setting up dependency injection for the controller.

```ts
const repoBinding = this.repository(ProductRepository);

// ...

inject(repoBinding.key)(ProductController, undefined, 0);

File: packages/rest-crud/README.md [hunk 7]
```diff
@@ -41,6 +41,67 @@ npm install --save @loopback/rest-crud
...
+
+    const ProductController = defineCrudRestController<
+    Product,
+    typeof Product.prototype.id,
+    'id'
+    >(Product, {basePath: '/products'});
+
+    inject('repositories.ProductRepository')(ProductController, undefined, 0);
+    this.bind('controllers.ProductController').toClass(ProductController);
```
bajtos:
```suggestion
    this.controller(ProductController);
```

packages/rest-crud/README.md
hacksparrow: @bajtos the app is complaining `The key 'datasources.db' is not bound to any value in context application`.

`repoBinding.key` is good suggestion.

packages/rest-crud/README.md
hacksparrow: The app is complaining `The key 'datasources.db' is not bound to any value in context application`.

packages/rest-crud/README.md
hacksparrow: The app is complaining `The key 'datasources.db' is not bound to any value in context application`.

packages/rest-crud/README.md
```diff
@@ -4,22 +4,21 @@ REST API controller implementing default CRUD semantics.
...
+having to implement custom controller or repository classes.
 
 ## Installation
 
 ```sh
 npm install --save @loopback/rest-crud
 ```
 
-## Basic use
+## Creating a CRUD Controller
```
bajtos: Please preserve `Basic use` section, it's part of our README guidelines - see https://loopback.io/doc/en/contrib/README-guidelines.html

Proposed structure:

```
## Basic use

(Put a short paragraph here to quickly introduce the sections below.)

### Creating a CRUD Controller

(...)

### Creating a CRUD repository

(...)

### Integrated example

(...)
```

packages/rest-crud/README.md
```diff
@@ -29,18 +28,79 @@ npm install --save @loopback/rest-crud
...
+const ProductRepository = defineCrudRepositoryClass(Product);
+const repo = new ProductRepository(db);
+```
+
+## Integrated example
+
+Here is an example of an app which uses `defineCrudRepositoryClass` and
+`defineCrudRestController` to fulfill its repository and controller
+requirements. You don't need to run `lb4 model` and `lb4 repository` to generate
+the model and respository classes.
```
bajtos: Interesting. In my mind, I wanted the example to keep as much of existing CLI & Boot setup as possible, i.e. use `lb4 model` and `lb4 datasource`, and write custom code for the repository & controller only. That's why I proposed to remove creation of the datasource from the boot method and assume it will be created by `@loopback/boot` from `src/datasources/db.datasource.{js,json}`.

If you prefer to create the model & datasource via code, then I am fine with that too, but we need to find out how to deal with `@loopback/boot`. Do we want to mix artifacts loaded by boot with artifacts defined from code? Do we want those two kinds of artifacts to share the same datasource?

packages/rest-crud/README.md
```diff
@@ -29,18 +28,79 @@ npm install --save @loopback/rest-crud
...
+
+```ts
+class Product extends Entity {
+  static definition = new ModelDefinition('Product')
+    .addProperty('id', {type: 'number', id: true})
+    .addProperty('name', {type: 'string'});
```
bajtos: This is duplicating the information provided via `@property` decorators. Please remove the static property `definition` and decorate the class with `@model` decorator instead. That will take care of populating `Product.definition` from metadata provided via `@property` decorators.

packages/rest-crud/README.md
```diff
@@ -29,18 +37,66 @@ npm install --save @loopback/rest-crud
...
+  }
+
+  async boot():Promise<void> {
+    await super.boot();
+
+    const ProductRepository = defineCrudRepositoryClass(Product);
+    const repoBinding = this.repository(ProductRepository);
+
+    inject('datasources.db')(ProductRepository, undefined, 0);
+    this.repository(ProductRepository);
```
bajtos: This is a duplicate of L83.

packages/rest-crud/README.md
```diff
@@ -29,18 +37,66 @@ npm install --save @loopback/rest-crud
...
+
+Use the `defineCrudRepositoryClass` method to create named repositories (based
+on the Model) for your app.
+
+Usage example:
+
+```ts
+const db = new juggler.DataSource({connector: 'memory'});
+const ProductRepository = defineCrudRepositoryClass(Product);
+const repo = new ProductRepository(db);
```
bajtos: I find it confusing that the Controller section is showing DI + app.controller, but the Repository section is showing manual approach, which also does not take into account different lifetimes of a datasource (one instance for entire app life) vs. a repository (new repository created for each request).

PR #3880, opened 2019-10-05
test(cli): remove redundant tests and fix typos/titles

agnes512:
Remove some redundant tests in belongsTo and hasMany cli generator tests
## Checklist

### DIFF
--- a/packages/cli/test/integration/generators/relation.belongs-to.integration.js
+++ b/packages/cli/test/integration/generators/relation.belongs-to.integration.js
@@ -115,25 +115,22 @@ describe('lb4 relation', function() {
     assert.fileContent(expectedFile, relationalPropertyRegEx);
   });
 
-  context('generate model relation', () => {
+  context('generates model relation with default values', () => {
     const promptArray = [
       {
         relationType: 'belongsTo',
         sourceModel: 'Order',
         destinationModel: 'Customer',
-        relationName: 'myCustomer',
       },
       {
         relationType: 'belongsTo',
         sourceModel: 'OrderClass',
         destinationModel: 'CustomerClass',
-        relationName: 'myCustomer',
       },
       {
         relationType: 'belongsTo',
         sourceModel: 'OrderClassType',
         destinationModel: 'CustomerClassType',
-        relationName: 'myCustomer',
       },
     ];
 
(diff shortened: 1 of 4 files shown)


packages/cli/test/integration/generators/relation.has-many.integration.js
```diff
@@ -356,7 +307,7 @@ describe('lb4 relation HasMany', function() {
           .withPrompts(multiItemPrompt);
       });
 
-      it('new controller file created', async () => {
+      it('new controller file has benn created', async () => {
```
nabdelgadir: typo: `benn` -> `been`

packages/cli/test/integration/generators/relation.belongs-to.integration.js
```diff
@@ -169,81 +166,57 @@ describe('lb4 relation', function() {
...
-        await testUtils
-          .executeGenerator(generator)
-          .inDir(SANDBOX_PATH, () =>
-            testUtils.givenLBProject(SANDBOX_PATH, {
-              additionalFiles: SANDBOX_FILES,
-            }),
-          )
-          .withPrompts(multiItemPrompt);
-      });
+    it('Verifies if property name that already existed will be overwritten ', async () => {
```
nabdelgadir: `verifies that a preexisting property will be overwritten`?

packages/cli/test/integration/generators/relation.belongs-to.integration.js
```diff
@@ -265,7 +238,7 @@ describe('lb4 relation', function() {
           .withPrompts(multiItemPrompt);
       });
 
-      it('relation name should be ' + defaultRelationName[i], async () => {
+      it('relation name should be customerPK', async () => {
```
jannyHou: hmm, I am not very familiar with the test here, is it necessary to replace the computed relation name `defaultRelationName[i]` with a fixed one(`customerPK`)?

agnes512: The default values are tested in the test `'generates model relation with default values'`. So I changed it to test `relationName` for 2 reasons:
1. The CLI seems to allow users to set up any relation name:
```
? Source property name for the relation getter: randomm
```
this would be 
```
this.cat = this.createBelongsToAccessorFor('randomm', catRepositoryGetter); // this doesn't seem right to me
```
( I think we actually have a bug here...)

2. Lineup with hasMany as it has this test. 


But as I mentioned above. I think our code has bugs, and this test should also check `repository.ts` as well to make sure it uses the correct relation name. But I decide to remove the redundant tests and move on the inclusion CLI task first, them to see if it _does_ have bugs

Issue #3854, opened 2019-10-01
Contribute OpenAPI spec pieces from extensions


jannyHou:
## Suggestion

This is inspired by [spike](https://github.com/strongloop/loopback4-example-shopping/pull/267). Authentication strategies should be able to contribute security schemas and operation security config into the app's openapi spec.

Let's start from creating an extension point for adding OpenAPI specs. It will read specs from extensions. Like `ControllerSpecBuilder` generates the operation specs, `AuthenticationSpecBuilder` reads security specs from all registered auth strategies(here it means register to another extension point for auth strategy).

And the rest server will load all registered OpenAPI spec pieces and merge them together.

## Use Cases

When people want to contribute OpenAPI spec pieces, they can register an extension that provides the spec.

A proposal:

start with app.api spec.
get controller to add operation spec
then pass the resulting open api spec to extensions of the extensionPoint to contribute

Contract for the extension point

transform( openApiSpec) : ValueOrPromise<OpenAPISpec> 

Then the extension point has a loader to find all extensions and call `extensionName.merge()` to keep adding more specs. 

## Examples

E.g. `ControllerSpecBuilder` generates the operation specs. `AuthenticationSpecBuilder` reads security specs from all registered auth strategies(here it means register to another extension point for auth strategy).

## Acceptance criteria

- [ ] Create an extension point for contributing OpenAPI spec
- [ ] Tests
- [ ] Documents

jannyHou:
cc @raymondfeng @emonddr I created this task as we discussed in the estimation meeting. Feel free to edit the description. Thanks.

dhmlau:
@jannyHou, assigning this to you per our discussion with you and @emonddr.

emonddr:
@jannyHou , on this pass, are we really going deliver AuthenticationSpecBuilder ? Or is the ControllerSpecBuilder sufficient for now, and in 2020 we deal with AuthenticationSpecBuilder?

jannyHou:
Close after PR merged

Issue #3669, opened 2019-09-05
propagate configured AuthenticationStrategy to OpenApiSpec


derdeka:
## Suggestion

The configured `AuthenticationStrategy` should propagate the `securityschema` into the generated openapi spec. Currently each endpoint need several configuration options to make this possible. 

## Use Cases

With configured securityschemas the api explorer handles authentication information automatically and sends it to the lb4 server.

## Examples

Currently I'm doing something like this:

application.ts
```typescript
import { merge } from 'lodash';
// ...
constructor(options: ApplicationConfig = {}) {
	// ...
	const spec = this.getSync(RestBindings.API_SPEC);
	merge(spec, {
	  components: {
		securitySchemes: {
		  BasicAuth: {
			type: 'http',
			scheme: 'basic',
		  },
		  BearerAuth: {
			type: 'http',
			scheme: 'bearer',
		  },
		  ApiKeyAuth: {
			type: 'apiKey',
			in: 'header',
			name: 'X-API-Key',
		  },
		},
	  },
	});
	// ...
}
```

user.controller.ts
```typescript
...
@authorize('jwt')
@get('/users', {
  security: [{
    BearerAuth: [],
  }],
  responses: {
    [STATUS_CODE.OK]: {
      content: { [CONTENT_TYPE.JSON]: { schema: { type: 'array', items: getModelSchemaRef(User) } } },
    },
  },
})
async userFind(
...
```
Please note, that `@authorize('jwt')` and `security` is some kind of redundant and needs to be configured for each endpoint.

## Acceptance criteria

- [ ] Authentication strategy can contribute security schemas when it gets registered to an application. The security schema specs will be merged into `OpenAPISpec.components.schemas`. Modify the [`registerAuthenticationStrategy()`](https://github.com/strongloop/loopback-next/blob/2be7e67f6f685795920341f7762fbdd8da38f7b2/packages/authentication/src/types.ts#L56) method to handle the spec merge. 

- [ ] Update `loopback4-shopping-example` to leverage the new change.

derdeka:
I found a simpler way without modifying all endpoints:

constants.ts
```typescript
export namespace SECURITY_REQUIREMENT {
  export const DEFAULT: SecurityRequirementObject[] = [
    {
      BearerAuth: [],
    },
  ]

  export const NONE: SecurityRequirementObject[] = [
  ]
}
```

application.ts
```typescript
import { merge } from 'lodash';
// ...
[...]

PR #3884, opened 2019-10-06
Initial openapi schema

frbuceta:
Implement initial openapi info

Captura de pantalla 2019-10-06 a las 19 18 19

Captura de pantalla 2019-10-06 a las 19 22 46

##  Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
+++ b/packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
@@ -201,6 +201,25 @@ describe('RestServer', () => {
         const expressApp = server.expressApp;
         expect(expressApp.get('strict routing')).to.equal(true);
       });
+      it('honors openApiSpec.initialObject in config', async () => {
+        const spec = {
+          info: {
+            name: 'Test',
+            version: '1.0.0-beta.0',
+          },
+        };
+
+        const app = new Application({
+          rest: {
+            openApiSpec: {
+              initialObject: spec,
+            },
+          },
+        });
+        app.component(RestComponent);
+        const server = await app.getServer(RestServer);
+        expect(server.getSync(RestBindings.API_SPEC).info).to.equal(spec.info);
+      });
     });
   });
 
--- a/packages/rest/src/rest.component.ts
+++ b/packages/rest/src/rest.component.ts
@@ -14,12 +14,12 @@ import {
 import {createEmptyApiSpec} from '@loopback/openapi-v3';
 import {
   JsonBodyParser,
+  RawBodyParser,
   RequestBodyParser,
   StreamBodyParser,
   TextBodyParser,
   UrlEncodedBodyParser,
 } from './body-parsers';
-import {RawBodyParser} from './body-parsers/body-parser.raw';
 import {RestBindings} from './keys';
 import {
   BindElementProvider,
@@ -89,6 +89,10 @@ export class RestComponent implements Component {
   ) {
     app.bind(RestBindings.SEQUENCE).toClass(DefaultSequence);
     const apiSpec = createEmptyApiSpec();
+    // Merge the OpenAPI spec from the config into the empty one
+    if (config && config.openApiSpec && config.openApiSpec.initialObject) {
+      Object.assign(apiSpec, config.openApiSpec.initialObject);
+    }
     // Merge the OpenAPI `servers` spec from the config into the empty one
     if (config && config.openApiSpec && config.openApiSpec.servers) {
       Object.assign(apiSpec, {servers: config.openApiSpec.servers});
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -973,6 +973,12 @@ export interface OpenApiSpecOptions {
    * Configure servers for OpenAPI spec
    */
   servers?: ServerObject[];
+
+  /**
+   * Configure initial object for OpenAPI spec
+   */
+  initialObject?: OpenAPIObject;
+
   /**
    * Set this flag to disable the endpoint for OpenAPI spec
    */


frbuceta:
Can you help me with coverage and tests? How should I proceed?

derdeka:
@frbuceta This can also be done via app.api() from your application. Why do you need to extend the rest server to pass this values?

frbuceta:
> @frbuceta This can also be done via app.api() from your application. Why do you need to extend the rest server to pass this values?

I think that would be the cleanest code

frbuceta:
@bajtos, how do you see it?

packages/rest/src/rest.component.ts
```diff
@@ -89,7 +89,10 @@ export class RestComponent implements Component {
   ) {
     app.bind(RestBindings.SEQUENCE).toClass(DefaultSequence);
     const apiSpec = createEmptyApiSpec();
-    // Merge the OpenAPI `servers` spec from the config into the empty one
+    // Merge the OpenAPI `info` & `servers` spec from the config into the empty one
+    if (config && config.openApiSpec && config.openApiSpec.info) {
```
derdeka: Instead of copying just the info object, maybe it makes sense to pass a partial schemaobject for initialisation?

derdeka:
An alternative to this approach could be #3854.

emonddr:
> An alternative to this approach could be #3854.








There are plans to allow contributions to the OpenAPI specification via an
extensionPoint/extensions pattern (
[Issue #3854](https://github.com/strongloop/loopback-next/issues/3854) );
including having authentication strategies automatically contribute security
scheme/requirement object information (
[Issue #3669](https://github.com/strongloop/loopback-next/issues/3669) ).

emonddr:
I noticed in a recent PR I worked on, https://github.com/strongloop/loopback4-example-shopping/pull/301, that if the `this.api({ ... })` method is used, it is necessary to pass in the `servers` value, otherwise the `Servers` combo box doesn't appear in the API explorer.

packages/rest/src/rest.component.ts
frbuceta: Yes, the whole object could also be passed

frbuceta: should I delete `openapi: {{version}}`?

packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
```diff
@@ -201,6 +201,25 @@ describe('RestServer', () => {
...
+          info: {
+            name: 'Test',
+            version: '1.0.0-beta.0',
+          },
+        };
+
+        const app = new Application({
+          rest: {
+            openApiSpec: {
+              schema: spec,
```
raymondfeng: The property name should NOT be `schema`. Let's use something like `initialObject`.

derdeka: Why not just merge the whole openApiSpec object?

frbuceta: > Why not just merge the whole openApiSpec object?

Not the whole object is the schema openapi

frbuceta: > The property name should NOT be `schema`. Let's use something like `initialObject` or `baseDocument`.

Solved

bajtos:
> I noticed in a recent PR I worked on, https://github.com/strongloop/loopback4-example-shopping/pull/301, that if the `this.api({ ... })` method is used, it is necessary to pass in the `servers` value, otherwise the `Servers` combo box doesn't appear in the API explorer; because it is also missing from the final/generated OpenAPI.

I think we need to fix the implementation of `app.api()` to be clever and merge the spec provided by the user with the `servers` value created by the Rest Server.

I see two use cases/scenarios to support:

1. The application is built design-first. It has an external OpenAPI spec document that's complete and should not be modified by the app. This document already contains information how to map API operations to LB4 Controller methods. 

2. The application is built code-first. The OpenAPI spec is built from metadata supplied via decorators on controller classes & methods and other sources. In this case, we need a way how to combine:
   - operation definitions from controllers (and routes)
   - global metadata contributed by extensions (see https://github.com/strongloop/loopback-next/issues/3854)
   - global metadata contributed by app developer, typically app name & version

IIRC, we designed the method `app.api` to support the first use case (design-first approach). I am not sure how easy it is to extend this method to support code-first design too. Maybe it's time to introduce a new API that will allow callers to provide only a subset of OpenAPI document and let the app to merge it with whatever spec it already has?

For example:

```ts
class RestApplication {
  // ...

  public mergeSpec(spec: Partial<OpenApiSpec>): void {
    // the implementation
  }
}
```

Alternative name: `addSpec`. Suggestions for an even better name are welcome!

packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
```diff
@@ -201,6 +201,25 @@ describe('RestServer', () => {
         const expressApp = server.expressApp;
         expect(expressApp.get('strict routing')).to.equal(true);
       });
+      it('honors openApiSpec.info in config', async () => {
```
raymondfeng: The title should be `openApiSpec.initialObject` now.

frbuceta: ohh, I fix it now

frbuceta:
> -1 from me.
> 
> As we have discussed in the past, application (and rest server) config is for things that change between environments, things that need to be injected into the application from outside. The port number where to listen on is a config value. The list of components to mount in the application is not a config value.
> 
> I feel that OpenAPI spec is not something that application consumers are supposed to inject, therefore it does not belong to application config.
> 
> My recommendation is to provide the "initial object" from your application constructor as follows:
> 
> ```ts
> export class TodoListApplication extends BootMixin(
>   ServiceMixin(RepositoryMixin(RestApplication)),
> ) {
>   constructor(options: ApplicationConfig = {}) {
>     super(options);
> 
>     this.api({
>       info: {
>         name: 'Example API',
>         version: require('../package.json').version,
>       }
>     });
> 
>     
>     // Set up the custom sequence
>     this.sequence(MySequence);
> 
>     // etc.
>   }
> }
> ```
> 
> Now it's possible this proposal does not work out of the box, in which case let's find a way how to make it work. Instead of adding `initialObject` to the application (rest server) config.

`this.api` forces you to put:

```
openapi: string(version),
info: {
  title: string,
  version: string
},
paths: {}
```

frbuceta:
> > Noté en un PR reciente en el que trabajé, [strongloop / loopback4-example-shopping # 301](https://github.com/strongloop/loopback4-example-shopping/pull/301) , que si `this.api({ ... })`se usa el método, es necesario pasar el `servers`valor; de lo contrario, el `Servers`cuadro combinado no aparece en el explorador de API; porque también falta en el OpenAPI final / generado.
> 
> Creo que debemos corregir la implementación de `app.api()`ser inteligentes y fusionar las especificaciones proporcionadas por el usuario con el `servers`valor creado por el servidor Rest.
> 
> Veo dos casos de uso / escenarios para apoyar:
> 
> 1. La aplicación está construida primero en diseño. Tiene un documento de especificaciones de OpenAPI externo que está completo y no debe ser modificado por la aplicación. Este documento ya contiene información sobre cómo asignar operaciones de API a los métodos del controlador LB4.
> 2. La aplicación está construida primero en código. La especificación OpenAPI está construida a partir de metadatos suministrados a través de decoradores en clases y métodos de controlador y otras fuentes. En este caso, necesitamos una forma de combinar:
>    
>    * definiciones de operación de controladores (y rutas)
>    * metadatos globales aportados por extensiones (ver [# 3854](https://github.com/strongloop/loopback-next/issues/3854) )
>    * metadatos globales aportados por el desarrollador de la aplicación, generalmente el nombre y la versión de la aplicación
> 
> IIRC, diseñamos el método `app.api`para admitir el primer caso de uso (enfoque de diseño primero). No estoy seguro de lo fácil que es extender este método para admitir el diseño de código primero también. ¿Tal vez es hora de presentar una nueva API que permita a las personas que llaman proporcionar solo un subconjunto de documentos OpenAPI y dejar que la aplicación lo combine con cualquier especificación que ya tenga?
> 
> Por ejemplo:
> 
> ```ts
> class  RestApplication {
>    // ... 
> 
>   public mergeSpec ( spec :  Partial < OpenApiSpec >) :  void {
>      // the implementación 
>   } 
> }
> ```
> 
> Nombre alternativo: `addSpec`. ¡Sugerencias para un nombre aún mejor son bienvenidas!

Okay, I'm going to do a little research to understand everything

packages/rest/src/rest.component.ts
```diff
@@ -89,6 +89,10 @@ export class RestComponent implements Component {
   ) {
     app.bind(RestBindings.SEQUENCE).toClass(DefaultSequence);
     const apiSpec = createEmptyApiSpec();
+    // Merge the OpenAPI spec from the config into the empty one
+    if (config && config.openApiSpec && config.openApiSpec.initialObject) {
```
dougal83: FYI: can now be: `config?.openApiSpec?.initialObject`
See [TS docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

bajtos:
> Okay, I'm going to do a little research to understand everything

@frbuceta Ping, have you managed to find some time to look into this area?

frbuceta:
> > Okay, I'm going to do a little research to understand everything
> 
> @frbuceta Ping, have you managed to find some time to look into this area?

Sorry for the delay, this weekend I start working on it. I had completely forgotten. I sign it up, in 3 days I am 100%

dougal83:
FYI: an [example info OASEnhancer](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/info.spec.extension.ts) is available.  Hopefully when #4554 lands we can get cracking with the updates.

frbuceta:
> FYI: an [example info OASEnhancer](https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/info.spec.extension.ts) is available. Hopefully when #4554 lands we can get cracking with the updates.

What is a replacement for this implementation?

dougal83:
@frbuceta The above example [spec. enhancer](https://loopback.io/doc/en/lb4/Extending-OpenAPI-specification.html) is a test that you can pretty much copy and alter in line with this PR.  For implementation, it could potentially be a `core enhancer` like the [proposed schema consolidator implementation](https://github.com/strongloop/loopback-next/pull/4365) but that depends on reviews in due course.

frbuceta:
I close this PR in favor of the `Spec Enhancer`

PR #3895, opened 2019-10-08
Simplify model schema with excluded properties

bajtos:
In a recent discussion about using [openapi-to-graphql](https://loopback.io/openapi-to-graphql.html) with LoopBack apps, we found that the schema created by LoopBack 4 is difficult to interpret. Schema for creating new model instances (e.g. `NewTodo`) is excluding the PK property (e.g. `id`) using `not` and `anyOf` operators, see #3297.

IMO, we should not need to use such construct, it should be enough to omit the excluded property from the schema.

While investigating such fix, I found out that JSON Schema allows arbitrary additional properties by default (`additionalProperties: true`). At the same time, LoopBack 4 models are strict by default and don't allow arbitrary properties.

The first commit removes this inconsistency by modifying `repository-json-schema` to correctly set `additionalProperties` based on `strict` model setting.

The second commits fixes `openapi-v3` to preserve boolean values of `additionalProperties` field. (Before my change, only objects were honored.)

The last commit finally simplifies the schema with excluded properties.

## Checklist

### DIFF
--- a/packages/openapi-v3/src/json-to-schema.ts
+++ b/packages/openapi-v3/src/json-to-schema.ts
@@ -75,7 +75,9 @@ export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
         break;
       }
       case 'additionalProperties': {
-        if (typeof json.additionalProperties !== 'boolean') {
+        if (typeof json.additionalProperties === 'boolean') {
+          result.additionalProperties = json.additionalProperties;
+        } else {
           result.additionalProperties = jsonToSchemaObject(
             json.additionalProperties!,
           );
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -30,6 +30,9 @@ describe('Validation at REST level', () => {
 
   @model()
   class Product {
+    @property()
+    id: number;
+
     @property({required: true})
     name: string;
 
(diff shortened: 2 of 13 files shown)


emonddr:
Coverage is slightly lower than before. Please introduce additional tests. thx.

packages/openapi-v3/src/json-to-schema.ts
```diff
@@ -75,7 +75,9 @@ export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
         break;
       }
       case 'additionalProperties': {
-        if (typeof json.additionalProperties !== 'boolean') {
+        if (typeof json.additionalProperties === 'boolean') {
```
agnes512: IIUC `additionalProperties` indicates that if the additional properties are allowed, it also converts JSON to object. 

Just out of curiosity, why can't we use `strict` field to decide if additional properties are allowed and `additionalProperties` to convert types when `strict` is true?

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -56,6 +59,47 @@ describe('Validation at REST level', () => {
     'description',
   );
 
+  context('with properties excluded from schema', () => {
+    before(() => {
+      const schema: SchemaObject = jsonToSchemaObject(
+        getJsonSchema(Product, {exclude: ['id']}),
+      );
+
```
raymondfeng: Can we add an assertion here for the schema with excluded properties?

bajtos: I feel that's redundant, we already have tests elsewhere to verify schema with excluded properties.

What kind of assertion do you have in mind?

bajtos:
> Coverage is slightly lower than before. Please introduce additional tests. thx.

@emonddr good point. Unfortunately I am not able to figure out which new lines are not covered now! 

The coverage report is confusing. It says that overall coverage fell down, but then it also says that coverage in `examples` and `extensions` stayed the same and went up in `packages`.

Screen Shot 2019-10-08 at 17 02 51

Maybe this is some kind of a rounding error, considering that the decrease is `0.0006%`?

Can you perhaps spot which line is not covered?

packages/openapi-v3/src/json-to-schema.ts
agnes512: I see, thanks for the clarification!

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
bajtos: I see what you mean now! I was a bit puzzled by this too. Apparently the original tests did not account for `not` and `anyOf`:

https://github.com/strongloop/loopback-next/pull/3297/files#diff-06ce773c57624b3b68d652f875f88ebbR949-R984

I'll rework those two test to verify bigger portion of the generated schema.

bajtos: I have opened a new PR to improve the tests in the current master. Once that PR is landed, I'll update this patch, which will make the removal of `not` + `anyOf` more visible.

https://github.com/strongloop/loopback-next/pull/3906

Issue #3706, opened 2019-09-11
Unable to POST on endpoint with recursive model


mmussche2:
Related to #3674
I have updated my openapi-v3 dependency, now my API starts correctly, but I can't send a post on the associated endpoint, I get an internal server error :

```
{
    "error": {
        "statusCode": 500,
        "name": "RangeError",
        "message": "Maximum call stack size exceeded",
[...]

PR #3897, opened 2019-10-08
fix: allow json schema with circular refs to be converted to OpenAPI schema

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/3706

## Checklist

### DIFF
--- a/packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
@@ -139,6 +139,135 @@ describe('jsonToSchemaObject', () => {
     expect(jsonToSchemaObject(inputDef)).to.eql(expectedDef);
   });
 
+  it('handles circular references with $ref', () => {
+    const schemaJson: JsonSchema = {
+      title: 'ReportState',
+      properties: {
+        // ReportState[]
+        states: {type: 'array', items: {$ref: '#/definitions/ReportState'}},
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+    };
+    const schema = jsonToSchemaObject(schemaJson);
+    expect(schema).to.eql({
+      title: 'ReportState',
+      properties: {
+        states: {
+          type: 'array',
+          items: {$ref: '#/components/schemas/ReportState'},
+        },
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+    });
+  });
+
+  it('handles circular references with object', () => {
+    const schemaJson: JsonSchema = {
+      title: 'ReportState',
+      properties: {
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+    };
+    // Add states: ReportState[]
+    schemaJson.properties!.states = {type: 'array', items: schemaJson};
+    const schema = jsonToSchemaObject(schemaJson);
+    expect(schema).to.eql({
+      title: 'ReportState',
+      properties: {
+        states: {
+          type: 'array',
+          items: {$ref: '#/components/schemas/ReportState'},
+        },
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+    });
+  });
+
+  it('handles indirect circular references with $ref', () => {
+    const schemaJson: JsonSchema = {
+      title: 'ReportState',
+      properties: {
+        parentState: {$ref: '#/definitions/ParentState'},
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+      definitions: {
+        ParentState: {
+          title: 'ParentState',
+          properties: {
+            timestamp: {type: 'string'},
+            state: {$ref: '#/definitions/ReportState'},
+          },
+        },
+      },
+    };
+    const schema = jsonToSchemaObject(schemaJson);
+    expect(schema).to.eql({
+      title: 'ReportState',
+      properties: {
+        parentState: {$ref: '#/components/schemas/ParentState'},
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+      definitions: {
+        ParentState: {
+          title: 'ParentState',
+          properties: {
+            timestamp: {type: 'string'},
+            state: {$ref: '#/components/schemas/ReportState'},
+          },
+        },
+      },
+    });
+  });
+
+  it('handles indirect circular references with object', () => {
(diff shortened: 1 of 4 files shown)


packages/openapi-v3/src/json-to-schema.ts
```diff
@@ -32,9 +37,25 @@ export type SchemaRef = ReferenceObject & {definitions: SchemasObject};
...
+ * circular references
  */
-export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
-  const result: SchemaObject = {};
+export function jsonToSchemaObject(
+  json: JsonSchema,
+  visited: Map<JsonSchema, SchemaObject | SchemaRef> = new Map(),
+): SchemaObject | SchemaRef {
+  // A flag to check if a schema object is fully converted
+  const converted = 'x-loopback-converted';
```
bajtos: Should we use a private Symbol instead of a string?

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -234,6 +234,8 @@ describe('build-schema', () => {
         benchmarkId: {type: 'string'},
         color: {type: 'string'},
       });
+      // No circular references in definitions
+      expect(schema.definitions).to.not.have.property('ReportState');
```
bajtos: Please assert that `schema` does not have any `definitions` property.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -467,7 +467,9 @@ export function modelToJsonSchema<T extends object>(
       delete schema.definitions;
     }
 
-    result.definitions[name] = schema;
+    if (result !== schema) {
```
bajtos: IIUC, because this check is executed after we initialize `result.definitions`, models with circular references may end up with an empty object in `definitions` property, which I find a bit ugly.

Let's try to find a way how to avoid having `definitions` set to an empty object.

packages/openapi-v3/src/json-to-schema.ts
raymondfeng: I thought about that too but decided to use open api extension mechanism to be consistent.

packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
```diff
@@ -139,6 +139,31 @@ describe('jsonToSchemaObject', () => {
...
+      title: 'ReportState',
+      properties: {
+        states: {
+          type: 'array',
+          items: {$ref: '#/components/schemas/ReportState'},
+        },
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+      definitions: {ReportState: {$ref: '#/components/schemas/ReportState'}},
```
bajtos: I find it weird to have a definition that's referencing itself, it's effectively creating a cyclic reference/infinite loop. Can we remove this entry please?

packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
```diff
@@ -139,6 +139,31 @@ describe('jsonToSchemaObject', () => {
...
+  it('handles circular references', () => {
+    const schemaJson: JsonSchema = {
+      title: 'ReportState',
+      properties: {
+        states: {type: 'array', items: {$ref: '#/definitions/ReportState'}},
+        benchmarkId: {type: 'string'},
+        color: {type: 'string'},
+      },
+    };
+    schemaJson.definitions = {ReportState: schemaJson};
```
bajtos: Is it necessary for `schemaJson.definitions` to contain an entry for `ReportState`? What will happen when there is no such entry? Can you add a test for that please?

raymondfeng: I improved the tests with different combinations.

raymondfeng:
@bajtos PTAL.

packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
```diff
@@ -139,6 +140,135 @@ describe('jsonToSchemaObject', () => {
...
+            timestamp: {type: 'string'},
+            state: {$ref: '#/components/schemas/ReportState'},
+          },
+        },
+      },
+    });
+  });
+
+  it('handles indirect circular references with object', () => {
+    const parentStateSchema: JSONSchema6Definition = {
```
agnes512: Why do we name it as `JSONSchema_6_Definition`?

packages/openapi-v3/src/json-to-schema.ts
```diff
@@ -32,9 +37,25 @@ export type SchemaRef = ReferenceObject & {definitions: SchemasObject};
...
+  const schema = visited.get(json);
+  if (schema != null && isSchemaObject(schema) && schema[converted] === false) {
+    return {$ref: `#/components/schemas/${json.title}`};
+  }
+  if (schema != null) return schema;
+
+  const result: SchemaObject = {
+    [converted]: false,
```
deepakrkris: 'x-loopback-converted' is never set to true. It is just a placeholder to verify the result object. why not directly declare it as a field here ?

packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
raymondfeng: It's based on JSON Schema draft 6 - https://json-schema.org/draft-06/json-schema-release-notes.html

packages/openapi-v3/src/json-to-schema.ts
raymondfeng: `x-*` is extensibility elements for OpenAPI specs.

packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
bajtos: It makes me wonder - why do we need `JSONSchema6Definition` interface here and cannot use `JsonSchema` that we are using in all other places?

AFAICT:

- `JSONSchema6Definition` is defined as `JSONSchema6 | boolean`
- `JsonSchema` is our alias for `JSONSchema6`

So IIUC, the difference between `JSONSchema6Definition` and `JsonSchema` is that the former also allows naked boolean values (`parentSchema = true`) while the latter requires an object.

Do you really need the `| boolean` part here?

PR #3909, opened 2019-10-10
chore: ensure packages are built in order of dependencies

raymondfeng:
We have `sort: "true"` in [lerna.json](https://github.com/strongloop/loopback-next/blob/master/lerna.json#L29) to maximize concurrency for `lerna run`. But for build, it should be run by the order of dependencies.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "prettier:fix": "npm run prettier:cli -- --write",
     "clean": "lerna run clean && node packages/build/bin/run-clean \"packages/*/dist\" \"extensions/*/dist\" \"examples/*/dist\" \"benchmark/dist\"",
     "clean:lerna": "lerna clean",
-    "build": "node bin/run-lerna run build",
+    "build": "node bin/run-lerna run build --sort",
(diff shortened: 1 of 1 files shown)


derdeka:
https://github.com/strongloop/loopback-next/blob/9a2aa4b4bb1cd0c44681eddbd573f1c4bff37585/lerna.json#L29

raymondfeng:
> What other action are invoked via lerna run? What is a more sensible default - run in parallel or in order of dependencies?

For example, `npm run clean` uses `lerna run clean`.

> I believe our build setup does not require dependencies to be compiled first, because our packages are set up in such way that the compiler sees the original .ts files anyways. This is reducing performance, because TypeScript has to re-parse those .ts files again and again whenever a dependant is built, but there is little we can do about it until we switch to project references.

I think it's always good to build packages based on the order of dependencies.

raymondfeng:
I also have a use case the prefers to build by the order of dependencies:

> Run `npm run build` but it fails in `examples/express-composition`. This leaves other `packages/*` modules unbuilt.

bajtos:
So IIUC, you want to keep `npm run clean` to run in parallel, but have `npm run build` to run in the order of dependencies.

I guess that's fair enough.

PR #3911, opened 2019-10-10
feat(openapi-v3): copy first example from examples to schema

derdeka:
Follow up to #2379

Copy first example from examples in JsonSchema to SchemaObject to be used in api explorer. This is a kind of workaround until the api explorer uses one of the listed examples or the default object.

## Checklist

### DIFF
--- a/packages/openapi-v3/src/json-to-schema.ts
+++ b/packages/openapi-v3/src/json-to-schema.ts
@@ -96,6 +96,13 @@ export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
         );
         break;
       }
+      case 'examples': {
+        if (Array.isArray(json.examples)) {
+          result.examples = json.examples;
+          result.example = json.examples[0];
+        }
+        break;
+      }
       default: {
         result[property] = json[property as keyof JsonSchema];
         break;
--- a/packages/repository-json-schema/src/filter-json-schema.ts
+++ b/packages/repository-json-schema/src/filter-json-schema.ts
@@ -34,7 +34,8 @@ export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
 
       limit: {
         type: 'integer',
-        minimum: 0,
+        minimum: 1,
+        examples: [100],
       },
 
       skip: {
--- a/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
@@ -35,7 +35,7 @@ describe('filterSchema', () => {
           additionalProperties: false,
         },
         offset: {type: 'integer', minimum: 0},
-        limit: {type: 'integer', minimum: 0},
+        limit: {type: 'integer', minimum: 1, example: 100, examples: [100]},
         skip: {type: 'integer', minimum: 0},
         order: {type: 'array', items: {type: 'string'}},
       },
--- a/packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts
@@ -149,6 +149,19 @@ describe('jsonToSchemaObject', () => {
     );
   });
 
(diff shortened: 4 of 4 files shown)


packages/openapi-v3/src/json-to-schema.ts
```diff
@@ -94,6 +94,13 @@ export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
         );
         break;
       }
+      case 'examples': {
+        if (Array.isArray(json.examples)) {
```
bajtos: Should we also check that there is at least one item in the array?

packages/openapi-v3/src/json-to-schema.ts
```diff
@@ -94,6 +94,13 @@ export function jsonToSchemaObject(json: JsonSchema): SchemaObject | SchemaRef {
...
+          result.examples = json.examples;
+          result.example = json.examples[0];
```
bajtos: I am not familiar with JSON Schema. Is it possible to specify a single example in JSON Schema too? Should we check if `result.example` has not been set by other means so that we don't overwrite it with the first item of examples array?

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -34,7 +34,8 @@ export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
 
       limit: {
         type: 'integer',
-        minimum: 0,
+        minimum: 1,
+        examples: [100],
```
bajtos: Nice. Can you please add examples for `offset` and `skip` properties too?

packages/openapi-v3/src/json-to-schema.ts
derdeka: If accessing an empty array `json.examples[0]` the result is `undefined`, which is a valid result. This fact is also used by other transformations in `jsonToSchemaObject`.

packages/openapi-v3/src/json-to-schema.ts
derdeka: `example` is not allowed in `JsonSchema`, so i think this can not happen. If this would be allowed we would not need this lines / workaround at all.

packages/repository-json-schema/src/filter-json-schema.ts
derdeka: We could specify `0` as example, but this is assumed by the explorer by default. Other values would break the user experience of the explorer as it sets this values as default, which might not be useful.

packages/openapi-v3/src/json-to-schema.ts
bajtos: Makes sense, thank you for the explanation 👍

packages/repository-json-schema/src/filter-json-schema.ts
bajtos: Makes sense 👍

packages/openapi-v3/src/json-to-schema.ts
bajtos: Fair enough 👍

bajtos:
@derdeka thank you for responses to my comments. Can you please add a test to verify handling of examples field in unit-test style? see https://github.com/strongloop/loopback-next/blob/master/packages/openapi-v3/src/__tests__/unit/json-to-schema.unit.ts

bajtos:
Landed, thank you for the contribution!

stoffason:
Hi, When using openapi-generator, this update breaks the generated code with error :
'....parameters.[xxxx].schemas.examples is unexpected'
Any workaround to disable the generated examples array?
Thanks!

derdeka:
@stoffason Please raise a new bug report if you found something which does not behave like specified in the OpenAPI specification.

bajtos:
> Hi, When using openapi-generator, this update breaks the generated code with error :
'....parameters.[xxxx].schemas.examples is unexpected'
Any workaround to disable the generated examples array?

The discussion continues here: https://github.com/strongloop/loopback-next/issues/4043

PR #3923, opened 2019-10-14
[RFC] feat(testlab): add docker support

raymondfeng:
This PR adds a few helper functions to use `docker` for tests. It's based on `dockerode` that uses REST APIs to interact with the `docker` daemon.

## Checklist

### DIFF
--- a/packages/testlab/src/__tests__/integration/docker.integration.ts
+++ b/packages/testlab/src/__tests__/integration/docker.integration.ts
@@ -1,0 +1,56 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/testlab
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {DockerClient} from '../..';
+import {expect} from '../../expect';
+
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
+describe('docker', async function() {
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(30000);
+
+  let verb: typeof it | typeof it.skip = it;
+  const client = new DockerClient();
+  const result = await client.isAvailable();
+  if (!result) {
+    verb = it.skip;
+  }
+
+  before(async () => {
+    if (result) await client.pull('alpine:latest');
+  });
+
+  verb('inspects an image by name', async () => {
+    const image = client.docker.getImage('alpine:latest');
+    const info = await image.inspect();
+    expect(info).containEql({RepoTags: ['alpine:latest']});
+  });
+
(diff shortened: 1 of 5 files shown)


bajtos:
I find it difficult to review this pull request because I have no idea what do you want to use the new APIs for.

Have you considered creating a new package instead of adding this functionality to `@loopback/testlab`? I am concerned that testlab is becoming too large. The initial idea was to create a thin wrapper around testing tools we use in many LB4 packages, especially to smooth out integration with TypeScript.

In my mind, Docker client is a very specialized feature that will be used only by few tests, therefore it's better to put in into a new package.

packages/testlab/src/__tests__/integration/docker.integration.ts
```diff
@@ -0,0 +1,51 @@
...
+import {expect} from '../../expect';
+
+describe('docker', function() {
+  let client: DockerClient;
+
+  // eslint-disable-next-line no-invalid-this
+  this.timeout(30000);
+
+  before(async () => {
+    client = new DockerClient();
```
bajtos: How is this going to run on machines that don't have Docker daemon running? For example Windows build machines like AppVeyor? Personally, I am not running the Docker daemon unless I really need it. I would consider it as a degradation of developer experience if `npm test` in loopback-next started to require a running Docker server.

bajtos:
@raymondfeng Have you considered to contribute this higher-level abstraction directly to `dockerode`?

packages/testlab/src/__tests__/integration/docker.integration.ts
bajtos: Travis CI builds on MacOS are failing too.

https://travis-ci.com/strongloop/loopback-next/jobs/245526182
```
 1) docker
       "before all" hook for "inspects an image by name":
     Error: connect ENOENT /var/run/docker.sock
      at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1128:14)
```

raymondfeng:
@bajtos Thank you for the feedback. 

I see the need to have simple Docker APIs that allow us to pull/start a Docker container for popular external resources such as Redis, MongoDB, and Kafka for integration/acceptance tests while I'm experimenting Cloud Deployment and Pub/Sub. 

I see your points that it does not fit `testlab` well. I'll try to find a different avenue for it.

- Contribute to dockerode
- Create a separate package

raymondfeng:
@bajtos Does it make sense to have docker-related helpers to be part of our `@loopback/build` module?

raymondfeng:
I just discovered https://github.com/testcontainers/testcontainers-node. It seems to a good fit for my use cases.

Issue #3437, opened 2019-07-26
Run repository tests for Cloudant


bajtos:
Add `acceptance/repository-cloudant` package, modeled after MongoDB & MySQL tests in `acceptance/repository-mongodb` and `acceptance/repository-mysql`. The tests should be run as a new Travis CI job, see the existing setup for other connectors.

See also https://github.com/strongloop/loopback-next/pull/3387

## Acceptance criteria

- [ ] A new directory `acceptance/repository-cloudant` running all CRUD tests from `repository-tests` packages
- [ ] A new matrix entry in Travis CI config to run Cloudant tests as part of our CI builds
- [ ] All tests are passing. If it's too difficult to make all tests to pass, then leverage `CrudFeatures` flags to skip failing tests and open follow-up issues to fix those tests that we want to pass.

PR #3968, opened 2019-10-18
feat(repository-tests): run repository-tests on cloudant

agnes512:
closes https://github.com/strongloop/loopback-next/issues/3437

Create `acceptance-repository-cloudant` folder.
- Modify tests in `replace-by-id` need to be modified to run the replacement operation on Cloudant ( it needs an extra property `_rev` to do so).
- Set up Travis task: it needs to run the script(`docker.setup.js`) to set up the docker image.
- the `docker.setup.js` file is based on @jannyHou 's [PR](https://github.com/strongloop/loopback-connector-cloudant/pull/227#pullrequestreview-317888999). I modified it a bit, now it can be used to set up the local test as well with customized names. ( see README).
- Add local test setup and README file to allow run repo tests on Cloudant ( or on docker) locally.

## Checklist


.travis.yml
```diff
@@ -1,9 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-  - "8"
-  - "10"
-  - "12"
+  - '8'
+  - '10'
+  - '12'
```
bajtos: Please revert unrelated changes, or alternatively open a standalone pull request to make them.

What VSCode plugin do you use to format the YAML file? Maybe we need to tell Prettier to ignore `.travis.yml`?

acceptance/repository-cloudant/src/__tests__/cloudant-default-repository.acceptance.ts
```diff
@@ -0,0 +1,20 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {DefaultTransactionalRepository} from '@loopback/repository';
+import {
+  CrudRepositoryCtor,
+  crudRepositoryTestSuite,
+} from '@loopback/repository-tests';
+import {CLOUDANT_CONFIG, CLOUDANT_FEATURES} from './cloudant.datasource';
+
+describe('CLOUDANT + DefaultTransactionalRepository', () => {
```
bajtos: Since Cloudant does not support transactions, let's use `DefaultCrudRepository` as we already do in MongoDB tests.

https://github.com/strongloop/loopback-next/blob/9def622d7bafb7eb2470cd896372f030998e3fd4/acceptance/repository-mongodb/src/__tests__/mongodb-default-repository.acceptance.ts#L6-L20

.travis.yml
agnes512: Yes, I use Prettier

packages/repository-tests/src/crud/replace-by-id.suite.ts
```diff
@@ -34,6 +34,11 @@ export function createSuiteForReplaceById(
     })
     id: MixedIdType;
 
+    // cloudant needs this property to do replacement method
+    // see cloundant README file for more details
```
agnes512: [Cloudant README](https://github.com/strongloop/loopback-connector-cloudant#_rev-property)

.travis.yml
agnes512: ps: `.prettierignore` doesn't work.

acceptance/repository-cloudant/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "verify": "npm pack && tar xf loopback-test-repository-cloudant*.tgz && tree package && npm run clean"
+  },
+  "author": "IBM Corp.",
+  "copyright.owner": "IBM Corp.",
+  "license": "MIT",
+  "devDependencies": {
+    "@loopback/build": "^1.7.1",
+    "@loopback/eslint-config": "^4.1.2",
+    "@loopback/repository": "^1.15.2",
+    "@loopback/repository-tests": "^0.5.2",
```
rmg: The version inside this repo is currently `0.6.1` which is not matched by `^0.5.2` so when `npm run build -- --scope "@loopback/test-repository-cloudant" --include-dependencies` is run it doesn't detect that it should be building `packages/test-repositories` and then when it goes to run the tests using that package, it has no `./dist` directory.

.travis.yml
```diff
@@ -80,6 +80,20 @@ matrix:
         - npm run build -- --scope "@loopback/test-repository-postgresql" --include-dependencies
         - psql -U postgres -a -f acceptance/repository-postgresql/test/schema.sql
         - cd acceptance/repository-postgresql && npm run mocha
+    - node_js: '10'
+      os: linux
+      env:
+        - TASK=test-repository-cloudant
+      install:
+        # - rm -rf node_modules && yarn cache clean
+        - npm ci && npm run docker:setup && source cloudant-config.sh && pwd
```
rmg: This should be done as the first `script` step, I think.

.travis.yml
```diff
@@ -80,6 +80,20 @@ matrix:
...
+
+      script:
+        - source cloudant-config.sh
```
rmg: This line alone does nothing as far as I can tell since each `script` command should be run in a new shell so the env vars exported by `cloudant-config.sh` just go away after this line.

.travis.yml
```diff
@@ -80,6 +80,20 @@ matrix:
...
+        - npm run postinstall -- --scope "@loopback/test-repository-cloudant" --include-dependencies
+        - npm run postinstall -- --include-dependencies
```
rmg: Not necessary once dependencies are fixed

.travis.yml
```diff
@@ -80,6 +80,20 @@ matrix:
...
+        - npm run build #-- --scope "@loopback/test-repository-cloudant" --include-dependencies
```
rmg: `npm run build -- --scope "@loopback/test-repository-cloudant" --include-dependencies` should be sufficient here

.travis.yml
```diff
@@ -80,6 +80,20 @@ matrix:
...
+        - cd acceptance/repository-cloudant && npm run mocha
```
rmg: This is where you should source your config.. `source cloudant-config.sh && cd acceptance/repository-cloudant && npm run mocha`

acceptance/repository-cloudant/package.json
agnes512: Yes! it works once I updated it. Thank you so much!!

.travis.yml
agnes512: You're right. And after removing it, it takes less time to run the tests.

.travis.yml
```diff
@@ -80,6 +80,16 @@ matrix:
         - npm run build -- --scope "@loopback/test-repository-postgresql" --include-dependencies
         - psql -U postgres -a -f acceptance/repository-postgresql/test/schema.sql
         - cd acceptance/repository-postgresql && npm run mocha
+    - node_js: "10"
+      os: linux
+      env:
+        - TASK=test-repository-cloudant
+      install:
```
rmg: No need for any `install` steps here at all. Just need to add `npm run docker:setup` to `scripts` any time before the last step.

package.json
```diff
@@ -62,7 +66,8 @@
     "docs:prepare": "./docs/bin/build-preview-site.sh",
     "docs:start": "cd docs/_preview && bundle exec jekyll serve --no-w --i",
     "mocha": "node packages/build/bin/run-mocha \"packages/*/dist/__tests__/**/*.js\" \"extensions/*/dist/__tests__/**/*.js\" \"examples/*/dist/__tests__/**/*.js\" \"packages/cli/test/**/*.js\" \"packages/build/test/*/*.js\"",
-    "posttest": "npm run lint"
+    "posttest": "npm run lint",
+    "docker:setup": "node ./docker.setup.js"
```
bajtos: Since the Docker setup is specific to Cloudant, can we move this npm script to `acceptance/repository-cloudant/package.json` please?

packages/repository-tests/src/crud/replace-by-id.suite.ts
bajtos: I find it suboptimal to force all connector tests (SQL, MongoDB, etc.) to use `_rev` property. Let's find a way how to allow individual connectors to provide any extra properties via `CrudFeatures`.

packages/repository-tests/src/crud/replace-by-id.suite.ts
```diff
@@ -71,7 +76,9 @@ export function createSuiteForReplaceById(
       await repo.replaceById(created.id, created);
 
       const found = await repo.findById(created.id);
-      expect(toJSON(found)).to.deepEqual(
+      // use containDeep instead of deepEqual because cloudant changes _rev value after
+      // replacement
+      expect(toJSON(found)).to.containDeep(
```
bajtos: I am concerned that by changing from `deepEqual` to `containDeep` we are allowing many more changes than just a different `_rev` value. For example, if the connector and/or the database adds new properties, this assertion will happily accept that, even though such behavior is most likely not wanted.

Let's find a way how to deal with `_rev` changes via a new `CrudFeatures` flag in such way that only changes in the `_rev` property is allowed and only when testing against Cloudant/CouchDB.

packages/repository-tests/src/crud/replace-by-id.suite.ts
```diff
@@ -100,7 +107,9 @@ export function createSuiteForReplaceById(
...
```
bajtos: Ditto.

acceptance/repository-cloudant/README.md
```diff
@@ -0,0 +1,77 @@
...
+NOTICE: we use `couchDB3` docker image for testing Cloudant because Cloudant
+doesn't have a maintained image, and most of the their functionalities are the
+same (except couchDB doesn't support geosearch).
+
+- Install [Docker](https://docs.docker.com/engine/installation/).
+
+- Use the `couchDB3:latest` image, and rename the container as
+  `cloudant-testdb`.
+
+- Start the container, and create an admin and password for this image at
```
bajtos: Can we automate this and the previous step please? In other repository acceptance packages, the docker-based setup requires the developer to run a single script only. Can we perhaps tell people to use `npm docker:setup`? (See also my other comments suggesting to move that script to this package.)

docker.setup.js
```diff
@@ -0,0 +1,310 @@
...
+const docker = new require('dockerode')();
+const fmt = require('util').format;
+const http = require('http');
+const ms = require('ms');
+const fs = require('fs-extra');
+
+// // we don't pass any node flags, so we can call _mocha instead the wrapper
+// const mochaBin = require.resolve('mocha/bin/_mocha');
+
+process.env.CLOUDANT_DATABASE = 'testdb';
```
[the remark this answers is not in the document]
agnes512: This file will be removed once the release is done. The script is from https://github.com/strongloop/loopback-next/pull/3968. I use it here to check if it sets up travis correctly and takes reasonable time to finish.

docker.setup.js
```diff
@@ -0,0 +1,310 @@
+// Copyright IBM Corp. 2017,2019. All Rights Reserved.
+// Node module: loopback-connector-cloudant
+// This file is licensed under the Artistic License 2.0.
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+const _ = require('lodash');
+const async = require('async');
```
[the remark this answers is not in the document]
agnes512: We can make changes in that script if needed.

package.json
agnes512: Yes :smile: I mentioned in the description above that once the script is released in `connector-cloudant`, the path will be changed to `"docker:setup": "node node_modules/loopback-connector-cloudant/docker.setup.js" `

packages/repository-tests/src/crud/replace-by-id.suite.ts
agnes512: Updated in the fifth commit.

acceptance/repository-cloudant/docker.setup.js
```diff
@@ -0,0 +1,321 @@
...
+// License text available at https://opensource.org/licenses/Artistic-2.0
+
+'use strict';
+
+const _ = require('lodash');
+const async = require('async');
+const docker = new require('dockerode')();
+const http = require('http');
+const ms = require('ms');
+const fs = require('fs-extra');
```
rmg: `fs-extra` isn't listed in `package.json`.. but it also doesn't appear to really be fully utilized here since the only method used is in `fs` from core.

acceptance/repository-cloudant/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@loopback/eslint-config": "^4.1.2",
+    "@loopback/repository": "^1.15.2",
+    "@loopback/repository-tests": "^0.6.1",
+    "@types/node": "^10.17.5",
+    "loopback-connector-cloudant": "2.3.2",
+    "typescript": "~3.7.2",
+    "async": "^3.1.0",
+    "debug": "^4.1.1",
+    "lodash": "^4.17.11",
+    "dockerode": "^2.4.3"
```
rmg: `ms` is missing from this list - it is used directly by `docker.setup.js`

agnes512: yeah I forgot.. but why is it still working? 0.0

acceptance/repository-cloudant/docker.setup.js
jannyHou: true `fs` is enough

docker.setup.js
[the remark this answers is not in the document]
jannyHou: @bajtos Thank you for the technical feedback for the setup script, it's actually added in https://github.com/strongloop/loopback-connector-cloudant/blob/master/docker.setup.js, @agnes512 put it in her PR before that script merged into the master.

To be honest to save the time, the code in that script is mostly copied from an existing script that has been running for test from 2 years ago 🙈 , so for some of your questions I don't quite know the answer, but I'll try :p

docker.setup.js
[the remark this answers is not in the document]
jannyHou: @bajtos I understand that an top-level async function returns a promise which needs to be handled as https://stackoverflow.com/questions/46515764/how-can-i-use-async-await-at-the-top-level, but `async.waterfall()` doesn't return a promise, it just executes functions in sequence then returns `undefined` IIUC...

package.json
jannyHou: it will be loaded from dependency `loopback-connector-cloudant`.

acceptance/repository-cloudant/package.json
rmg: `ms` is a pretty common dependency. If any of the other dependencies depend on it then they would have pulled it in and due to dependency tree flattening, it would be installed as a top-level dependency.

package.json
agnes512: @jannyHou I moved the whole script to the acceptance/repo-cloudant. But I am okay with either way~

acceptance/repository-cloudant/README.md
agnes512: Done, and README is updated. ( the script is run in this folder now)

docker.setup.js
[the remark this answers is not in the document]
bajtos: > Thank you for the technical feedback of the setup script, it's actually added in https://github.com/strongloop/loopback-connector-cloudant/blob/master/docker.setup.js, @agnes512 put it in her PR before that script merged into the master.

I see, makes sense.

> To be honest, to save the time, the code in that script is mostly copied from an existing script that has been running for test from 2 years ago

Sure, no need to refactor the script if it's working well as it is now. I am going to mark all my comments for `docker.setup.js` as resolved then.

acceptance/repository-cloudant/README.md
```diff
@@ -45,27 +39,25 @@ same (except couchDB doesn't support geosearch).
...
-
-- Start the container, and create an admin and password for this image at
-  `http://localhost:{YOURPORT}/_utils/` if it doesn't have any yet.
-
-- Then run the following script
+- Run the following script
 
 ```bash
-source setup.sh <HOST> <PORT> <USER> <PASSWORD> <DATABASE>
+CLOUDANT_USER=<USER> CLOUDANT_PASSWORD=<PASSWORD> CLOUDANT_DATABASE=<DATABASE> node docker.setup.sh
```
bajtos: `node docker.setup.sh` looks suspicious - node typically expects `.js` files, not `.sh`. I think it's just a typo?

```suggestion
CLOUDANT_USER=<USER> CLOUDANT_PASSWORD=<PASSWORD> CLOUDANT_DATABASE=<DATABASE> node docker.setup.js
```

BTW why `node docker.setup.js` and not `npm run docker:setup`? Should we modify Travis CI config to call `node docker.setup.js` too?

packages/repository-tests/src/crud/replace-by-id.suite.ts
bajtos: I see that you added `revisionTokenRequired` flag to fix test failures.

Personally, I'd prefer to find a way how to define `_rev` property only when running on Cloudant, so that we are not using `_rev` property when testing e.g. MongoDB.

Having said that, I don't have any solution I could offer from the top of my head, so it's probably better to accept the current proposal for now and get this pull request landed sooner rather than later 👍

packages/repository-tests/src/types.repository-tests.ts
```diff
@@ -64,6 +64,14 @@ export interface CrudFeatures {
...
    */
   supportsInclusionResolvers: boolean;
+
+  /**
+   * Does the connector require the revision token for replacement operations?
+   * (Cloudant)
+   *
+   * Default: `false`
+   */
+  revisionTokenRequired: boolean;
```
bajtos: Nitpick - feel free to ignore.

The `_rev` value is not always required, is it? For example, when creating a new model instance, the `_rev` must not be provided, at least AFAIK.

Maybe `hasRevisionToken` would be a better name?

acceptance/repository-cloudant/README.md
agnes512: Good catch!

And right, I think 
```
CLOUDANT_USERNAME=<USER> CLOUDANT_PASSWORD=<PASSWORD> CLOUDANT_DATABASE=<DATABASE> npm run docker:setup
```
 also allows users to have their own configurations 👍 changed.

packages/repository-tests/src/types.repository-tests.ts
agnes512: make sense

acceptance/repository-cloudant/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@loopback/test-repository-cloudant",
+  "version": "1.1.1",
```
nabdelgadir: I think it should be `1.0.0`

acceptance/repository-cloudant/src/__tests__/cloudant-default-repository.acceptance.ts
```diff
@@ -0,0 +1,20 @@
...
+  CrudRepositoryCtor,
+  crudRepositoryTestSuite,
+} from '@loopback/repository-tests';
+import {CLOUDANT_CONFIG, CLOUDANT_FEATURES} from './cloudant.datasource';
+
+describe('CLOUDANT + DefaultCrudRepository', () => {
+  crudRepositoryTestSuite(
+    CLOUDANT_CONFIG,
+    // Workaround for https://github.com/microsoft/TypeScript/issues/31840
+    DefaultTransactionalRepository as CrudRepositoryCtor,
```
nabdelgadir: Since it [doesn't support transactions](https://github.com/strongloop/loopback-next/pull/3968/files#diff-b64c9eeda3125036291ac7b928f708a6R22), I think you should change this to `DefaultCrudRepository` like in the MongoDB one.

packages/repository-tests/src/crud/replace-by-id.suite.ts
nabdelgadir: typo: `cloundant` -> `cloudant`

acceptance/repository-cloudant/package.json
jannyHou: +1

acceptance/repository-cloudant/src/__tests__/cloudant-default-repository.acceptance.ts
jannyHou: 👍 Cloudant uses `_rev` to handle the update conflict, it doesn't handle transaction like sql dbs.

agnes512: right, I just copied it from MySQL :p

PR #3979, opened 2019-10-21
Improve http-caching-proxy and geocoder test

raymondfeng:
The examples/todo depends on https://geocoding.geo.census.gov/geocoder/, which can be down once a while.

The PR improves http-caching-proxy with two new options:

1. timeout - timeout connecting to the target endpoint
2. logError - a flag to control if errors should be logged to console

It also makes the todo tests conditional on the availability of the geocoding service.

## Checklist

### DIFF
--- a/examples/todo/src/__tests__/integration/services/geocoder.service.integration.ts
+++ b/examples/todo/src/__tests__/integration/services/geocoder.service.integration.ts
@@ -7,9 +7,11 @@ import {expect} from '@loopback/testlab';
 import {GeocoderDataSource} from '../../../datasources/geocoder.datasource';
 import {GeocoderService, GeocoderServiceProvider} from '../../../services';
 import {
+  aLocation,
   getProxiedGeoCoderConfig,
   givenCachingProxy,
   HttpCachingProxy,
+  isGeoCoderServiceAvailable,
 } from '../../helpers';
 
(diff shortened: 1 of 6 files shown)


examples/todo/src/__tests__/integration/services/geocoder.service.integration.ts
```diff
@@ -23,15 +25,16 @@ describe('GeoLookupService', function() {
...
-    const points = await service.geocode('1 New Orchard Road, Armonk, 10504');
-
-    expect(points).to.deepEqual([
-      {
-        y: 41.109653,
-        x: -73.72467,
-      },
-    ]);
+    if (!available) return;
+    const points = await service.geocode(aLocation.address);
```
jannyHou: The geocode for a certain address is already computed when checking the availability, it would be better to only run `service.geocode()` once, like returning the result OR true/false/error so we don't need to do it again here.

dhmlau:
Not sure why coveralls is complaining about decreased in coverage. It seems like it's complaining about files that are not part of the PR?

examples/todo/src/__tests__/integration/services/geocoder.service.integration.ts
raymondfeng: It should be the responsibility of http-caching-proxy.

PR #4028, opened 2019-10-31
test(example-todo): mark geocoding tests as skipped when the service is down

bajtos:
I find the solution introduced by https://github.com/strongloop/loopback-next/pull/3979 very problematic. When the geocoder service is not available, we skip the test body and yet mark the test as passed. This creates a false sense of security, leading us to believe that the geocoding functionality is covered by tests and working, despite the fact that eventually we may get to the point when the test is always skipped.

In this pull request, I am improving the code to report the tests as skipped when the Geocoder service is not available.

This is not perfect, because we won't have any automated notification to let us know when the geocoder service becomes permanently broken, but at least we have a chance to notice the problem.

## Checklist

### DIFF
--- a/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
+++ b/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
@@ -87,7 +87,8 @@ describe('TodoApplication', () => {
   });
 
   it('creates an address-based reminder', async function() {
-    if (!available) return;
+    // eslint-disable-next-line no-invalid-this
+    if (!available) return this.skip();
     // Increase the timeout to accommodate slow network connections
     // eslint-disable-next-line no-invalid-this
     this.timeout(30000);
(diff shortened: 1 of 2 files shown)


examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -87,7 +87,8 @@ describe('TodoApplication', () => {
   });
 
   it('creates an address-based reminder', async function() {
-    if (!available) return;
+    // eslint-disable-next-line no-invalid-this
+    if (!available) return this.skip();
```
nabdelgadir: While we won't be notified, would it be worth it to do something like `console.warn('service down, skipping test')`?

raymondfeng: Good trick. I didn't know we can mark the test skipped this way. 

I was trying to use it.skip with the `available` flag calculated in the `before` hook but that is too late.

bajtos: IDK, to be honest 🤷‍♂

PR #4037, opened 2019-11-01
fix(repository): improve typings for model inclusion

bajtos:
Fix a problem discovered by the upcoming TypeScript compiler version 3.7, see the discussion in https://github.com/strongloop/loopback-next/pull/3858 for more details.

Here is gist of the problem:

Before my change, when querying an `Author` model and including let's say `BlogPost` models written by the author, the `include` part of the filter was expecting `scope` property with `Filter<Author>` instead of `Filter<BlogPost>`. The latter is difficult to express, because Filter is recursive and when a `BlogPost` has many comments, then `Filter<BlogPost>` needs to accept `Filter<Comment>` in its nested `include[].scope`.

In this pull request, I am changing `include[].scope` property to accept `Filter` for any model. This new style is consistent with `include[].relation` property which is accepting any string (as opposed to accepting only one of the known relations of the source models).

## Checklist

### DIFF
--- a/packages/repository/src/query.ts
+++ b/packages/repository/src/query.ts
@@ -170,8 +170,18 @@ export type Fields<MT = AnyObject> = {[P in keyof MT]?: boolean};
  * `{relation: 'aRelationName', scope: {<AFilterObject>}}`
  */
 export interface Inclusion<MT extends object = AnyObject> {
+  // ^^^ TODO(semver-major): remove the generic argument <MT>
+
   relation: string;
-  scope?: Filter<MT>;
+
+  // Technically, we should restrict the filter to target model.
+  // That is unfortunately rather difficult to achieve, because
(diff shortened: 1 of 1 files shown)


bajtos:
@raymondfeng I think the following two commits should be easy to cherry-pick and land on master independently of this patch: https://github.com/strongloop/loopback-next/pull/3858/commits/3f1fed03103c3c80d8c403fb10fb243fe56b0963 and https://github.com/strongloop/loopback-next/pull/3858/commits/005e7f131b6f447bc0c6dbba26f496bbac1c267b.

The third commit https://github.com/strongloop/loopback-next/pull/3858/commits/46132b72e4c151074267475038cac207deadd191 is tricky because it mixes different fixes, I think we need to split it in smaller parts and land only those that are relevant.

Issue #2556, opened 2019-03-08
Variable substitution in DataSource JSON files


bajtos:
To support https://12factor.net/ application loading their database connection strings from ENV variables, we should enhance our DataSource booter to provide variable substitution similar to what https://github.com/strongloop/loopback-boot already provides for LB3 applications.

Example LB3 config file:

```json5
// server/datasources.json
{
  "db": {
    "connector": "mysql",
    "url": "${MYSQL_URL}",
  }
}
```

Proposed format for LB4:

```json5
// src/datasources/db.json
{
  "name": "db",
  "connector": "mysql",
  "url": "${ENV.MYSQL_URL}",
}
```

By using `ENV.` prefix, we are keeping doors open to add other variable sources in the future, e.g. `APP.` prefix to access application-level bindings.

To support numeric values, allow the following format (notice the `+` sign after the opening bracket):
```json5
{
  "port": "${+ENV.PORT}",
}
```

## Acceptance criteria

 - [ ] A new helper function accepting configuration object loaded from the JSON file, an object with values to use for substitutions (conceptually `{ENV: process.env}`), and returning modified configuration object with substitutions resolved according to rules described above. Include comprehensive test coverage.
 - [ ] Substitution should be applied to values at all levels, including deeply nested properties.
 - [ ] Support conversion to numeric values as described above.
 - [ ] `@loopback/boot` recognizes and replaces variables in datasource configuration. Leverage the already implemented helper function.
 - [ ] `process.env` must not be accessed directly from individual booters. Instead, the environment object should be provided via `app.boot()` arguments. For example, we can add a new property to `BootExecutionOptions`.
 - [ ] Modify examples and CLI templates to pass `process.env` value to `app.boot()`, so that variable substitution can be performed.
 - [ ] Documentation
 - [ ] Blog post

**Out of scope**

- Allow extensions to contribute custom convention for environment-specific operational configuration - see the spike #1464

marioestradarosa:
@bajtos , I was analyzing the datasource booter and found that this can be done before it calls the repositoryMixin.dataSource easily. I am suggesting that we use a convention based on the datasource name, without specifying any variable in the .json file.

```ts
// src/datasources/db.json
{
  "name": "myDBName",
  "connector": "mysql",
  "host": "xxxx",
  "port": "yyyy",
  "user": "xxxxxkkd",
...
}

````
[...]

Issue #2400, opened 2019-02-14
What environment variables should we use for database configuration?


ri8ika:
After following [medium blog](https://itnext.io/loopback-4-database-configuration-8f085399268), I set the following variables:
```
DATABASE_URL
DB_DATABASE
DB_HOST
DB_PASSWORD
DB_PORT
DB_USER
```

But this is not working.

After some research I found this issue:

https://github.com/strongloop/loopback-boot/issues/4

But I am still not sure what database environment variables are available to use in loopback 4.

---

For further information, I have used the followings inside application.ts inside the constructor:
```
this.bind('datasources.config.db').to({
      name: 'postgres',
      connector: 'postgres',
      hostname: process.env.DB_HOST,
      port: process.env.DB_PORT,
      user: process.env.DB_USER,
      password: process.env.DB_PASSWORD,
      database: process.env.DB_DATABASE,
    });
    this.bind('datasources.db').toClass(PostgresDataSource);
```

I think this is not getting bound, because when try to run migrate on heroku, I get the following error:

```
Connection fails: Error: connect ECONNREFUSED 127.0.0.1:5432
```

dhmlau:
I'm transferring your issue to `loopback-next` repo. That's where it should be anyway.

dhmlau:
You might also wanna see:
https://github.com/strongloop/loopback-next/issues/2176

Hope it helps.

ri8ika:
Hmm, I'm still getting the error even after using Procfile to run index.js when I try to migrate:
```
Error: connect ECONNREFUSED 127.0.0.1:5432
```

dhmlau:
Sorry, i misunderstood your issue and it's actually complaining about cannot connect to the Postgresql.  Looking at the code snippet in the original comment, should `hostname` be `host` instead?  It's because in the datasource.json, it's in the form of:
```
{
  "name": "ds",
  "connector": "postgresql",
  "url": "",
  "host": "localhost", <---- HERE
  "port": 5432,
  "user": "root",
  "password": "pass",
  "database": "testdb"
}

```

ri8ika:
Good catch. But I'm still having same issue.

dhmlau:
@ri8ika, one last thing I can think of.  In this tutorial https://loopback.io/doc/en/lb4/Deploying-to-IBM-Cloud.html#updating-application, it binds the new value in `src/index.ts`:
```
app.bind('datasources.config.db').to(updatedDatasourceDb);
```

[...]

Issue #782, opened 2017-12-07
Easy deploy to Cloud/Production Epic


bajtos:
We should make it easy to deploy LoopBack4 applications to most common production types:

 - IBM Cloud & API Connect
 - [Twelve-factor](https://12factor.net/) platforms (e.g. Heroku), where configuration is read from environment variables (e.g. PORT, database connection strings, etc.)
 - Docker and Kubernetes
 - http://istio.io/

virkt25:
not-stale

bajtos:
Related PRs:

- improve states and add shutdown hooks for applications #4145
- add integration with winston and fluentd logging #4117
- add a new package to enable open tracing #3389
- handle http keepalive connections upon application.stop() #4146

PR #4051, opened 2019-11-05
chore: upgrade to typescript@3.7.2

raymondfeng:
~~Blocked by https://github.com/typescript-eslint/typescript-eslint/issues/815~~ (fixed by https://github.com/strongloop/loopback-next/pull/4051/commits/b21c2ce42070aa4f6416070a4ebf6f6949e75c42)
Worked around https://github.com/microsoft/TypeScript/issues/34761

## Checklist


dhmlau:
@raymondfeng , any reasons that we are renaming the `*.datasource.json` to `*.config.json`?

raymondfeng:
> any reasons that we are renaming the *.datasource.json to *.config.json?

It's a workaround/fix for https://github.com/microsoft/TypeScript/issues/34761. Considering we have:

```
db.datasource.json
db.datasource.ts
```

TypeScript 3.7.x tries to emit db.datasource.d.ts for both files and complains about the conflict.

marioestradarosa:
@raymondfeng nice catch on that. I read the whole issue on typescript and now I understand better. LGTM !. 

I see conflicts of course with current applications in that they will need to manually change the `datasource.json` to `datasource.config` correct?.  

Just a question:  Isn't it possible to just move it to its own config directory ie: `config/myds.datasource.json` and put all the files there ?

docs/site/DataSource-generator.md
```diff
@@ -51,7 +51,7 @@ Once all the prompts have been answered, the CLI will do the following:
 - Install `@loopback/repository` and the connector package (if it's not a custom
   connector).
 - Create a file with the connector configuration as follows:
-  `/src/datasources/${dataSource.dataSourceName}.datasource.json`
+  `/src/datasources/${dataSource.dataSourceName}.config.json`
```
bajtos: This is a pretty disruptive change, let's open a new pull request to discuss it in more details please.

Personally, if we keep the current design, then I prefer the suffix `.datasource-config.json`.

However, since we are already changing datasource-related file convention, I would like to consider a more radical change - let's remove the JSON configuration files altogether and move the configuration options to the `.ts` files.

Here is my reasoning:

- We used JSON files for datasource config because that's what we do in LB3. I think we hoped that eventually there will be tooling to read & update datasource config programmatically, e.g. from a GUI tool. Since this didn't happen yet and we don't have any short-term plans to introduce such functionality, I think we can safely get rid of the complexity of JSON config files.
- We have already implemented editing of TypeScript files via AST manipulation. If we want to read & update datasource config programmatically in the future, it should be feasible to use the same AST-based approach instead of relying on JSON.
- Other places, like model definitions, don't use JSON files either. I find it inconsistent that we use TypeScript of almost everything, but datasources use JSON.

Proposed structure of a datasource `.ts` file:

```ts
// imports

export const CONFIG = {
  connector: 'memory',
  // etc.
  // (content from the JSON file goes here)
};

export class DbDataSource extends juggler.DataSource {
  static dataSourceName = 'db';

  constructor(
    @inject('datasources.config.db', {optional: true})
    config: object = CONFIG,
  ) {
    super(config);
  }
}
```

As an important (and wanted) side effect, this new layout will allow developers to tweak datasource configuration based on the environment. For example:

```ts
export const CONFIG = {
  connector: 'mysql',
  url: process.env.MYSQL_URL,
  // etc.
};
```

While I consider accessing ENV variables deeply from application code as an anti-pattern to avoid, I also think it's a reasonable workaround for now, until we figure out a cleaner way how to apply environment configuration (see #2400, #2556 and #782).

bajtos: Here is the PR: https://github.com/strongloop/loopback-next/pull/4060

bajtos:
> chore: refresh package-lock.json

Can we leave this change out of this pull request please and let RenovateBot do the job?

packages/cli/package.json
```diff
@@ -85,16 +85,16 @@
...
       "debug": "^4.1.1",
       "fs-extra": "^8.1.0",
       "mocha": "^6.2.2",
       "nyc": "^14.1.1",
-      "prettier": "^1.18.2",
+      "prettier": "^1.19.1",
       "rimraf": "^3.0.0",
       "source-map-support": "^0.5.16",
-      "typescript": "~3.6.4",
+      "typescript": "~3.7.2",
```
bajtos: I am surprised to see this change as part of the commit `chore: refresh package-lock.json`, I would expect it to be a part of `chore: upgrade to TypeScript 3.7.2`.

jomarocas:
Good day

i update typescript only for test error

first error is

```
error TS5056: Cannot write file '/dist/datasources/emptor.datasource.d.ts' because it would be overwritten by multiple input files.
error TS5056: Cannot write file '/dist/datasources/reactions.datasource.d.ts' because it would be overwritten by multiple input files.
```

Second error later of rename from reactions.datasource.ts to reactions.datasource.d.ts

```
500 Error: The key 'datasources.reactions' is not bound to any value in context application
 at ReactionsApplication.getBinding (/node_modules/@loopback/context/src/context.ts:889:11)
```

how can solve this?

dhmlau:
@jomarocas, the naming convention for the datasource has been changed.  Your problem should be fixed by renaming the datasource json files  `*.datasource.json` to `*.datasource.config.json`. 

For details, please see https://github.com/strongloop/loopback-next/issues/4173#issuecomment-555970616.

PR #4080, opened 2019-11-08
feat(cli): add lb4 update command to check/update project dependencies

raymondfeng:
Adding `lb4 version` for https://loopback.io/doc/en/lb4/Command-line-interface.html#upgrading-loopback-dependencies:

- `lb4 update` runs against a LB4 project and checks dependencies against the installed `@loopback/cli`.
- By default, `lb4 update` checks for exact match. Use `--semver` option to check compatibility using semver semantics. 

## Checklist


MarkOfSine:
@raymondfeng this will be useful feature.
I would suggest an enhancement to include version checking of typescript. i.e. in particular noting where certain features are not yet compatible with a particular version of typescript.
Mainly forward looking (i.e. latest or more current versions of Typescript)

raymondfeng:
@MarkOfSine Thank you for chiming in.

> I would suggest an enhancement to include version checking of typescript. i.e. in particular noting where certain features are not yet compatible with a particular version of typescript.
Mainly forward looking (i.e. latest or more current versions of Typescript)

The version of `typescript` is part of CLI template for dependencies too. As a result. `lb4 version` checks `typescript`. But we don't have enough knowledge to tell if it's fine to upgrade to latest version TypeScript though. Our experience is that it's usually safe to stay within patch versions, but often breaking with new minor. For example `~3.6.4` is good while `^3.6.4` is risky.

docs/site/Version-generator.md
```diff
@@ -0,0 +1,64 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Version-generator.html
+---
+
+{% include content/generator-create-app.html lang=page.lang %}
+
+### Synopsis
+
+The `lb4 version` command runs against a LoopBack 4 project and checks
```
bajtos: I find the name `version` problematic, it does not communicate the fact that it can be used to update dependencies. We already have `lb4 --version` to print the version of LoopBack dependencies installed in the project.

I am proposing:

(1)
Improve `lb4 --version` to include information about packages that are outdated. At the moment, we print output like this:

```
$ lb4 --version
@loopback/cli version: 1.24.0

@loopback/* dependencies:
  - @loopback/authentication: ^3.2.1
  - @loopback/boot: ^1.5.10
...
```

I am proposing to add a short text after each outdated version, possibly highlighted using `chalk.red`:

```
$ lb4 --version
@loopback/cli version: 1.24.0

@loopback/* dependencies:
  - @loopback/authentication: ^3.0.1 (outdated)
  - @loopback/boot: ^1.5.10
```

(2)
Rename the new command to something like `lb4 update` or `lb4 update-deps`. Personally, I prefer `lb4 update`, because it's short and in line with `npm update` command that most developers are hopefully familiar with.

This opens a question though - what is the benefit of this LB-specific update command when compared to `npm update`? Why (or when?) should our users use this new command instead of `npm update`?

bajtos: > What is the benefit of this LB-specific update command when compared to npm update? Why (or when?) should our users use this new command instead of npm update?

It would be awesome if `lb4 update` could run scripts to automatically apply changes necessary to migrate the codebase to newer conventions. That would be a perfect explanation why to run `lb4 update` instead of `npm update`.

raymondfeng: I'm good to have `lb4 update`. The key benefit of `lb4 update` over `npm update` is that `lb4 update` honors the collection of compatible LB4 packages that released with the current `lb4` command from `@loopback/cli`.

raymondfeng: Please `--version` is implemented at `cli.js` level without a yeoman generator.

raymondfeng:
@bajtos I have refactored the code to use `lb4 update`.

docs/site/Update-generator.md
```diff
@@ -0,0 +1,64 @@
...
+`--semver` : _(Optional)_ Use semver semantics to check version compatibility
+for project dependencies of LoopBack modules
+
+### Output
+
+The following is a sample output from `lb4 update` against a LoopBack 4 project
+generated by older version of CLI:
+
+```
+The project was originally generated by @loopback/cli@<unknown>.
```
bajtos: I find the `<unknown>` version suspicious. Is it a temporary thing because the current CLI versions are not writing the version info? Can we print a nicer error message in such case? For example:

```suggestion
The project was originally generated by an older version of @loopback/cli.
```

Also what happens after we run `npm update` and then a new version of `@loopback/cli` is published - are we going to print the CLI version that was used to generate the project or the last CLI version that was executed to update dependencies?

docs/site/Update-generator.md
```diff
@@ -0,0 +1,64 @@
...
+semantics.
+
+```sh
+lb4 update [options]
+```
+
+If the command is run without a LoopBack 4 project, it prints the version of the
+current CLI and versions of LoopBack packages that are released with the
+`@loopback/cli`. This is the same as `lb4 -v`.
+
```
bajtos: Please explain when and why should users run `lb4 update` instead of `npm update`. I'd like this information to be easy to find, not buried in a pull request discussion.

docs/site/Update-generator.md
raymondfeng: The `unknown` is only there because my very old project does not have such information.

raymondfeng:
@bajtos PTAL.

docs/site/Update-generator.md
```diff
@@ -0,0 +1,68 @@
...
+released with the current lb4 command from `@loopback/cli`.
+
+If the command is run without a LoopBack 4 project, it prints the version of the
+current CLI and versions of LoopBack packages that are released with the
+`@loopback/cli`. This is the same as `lb4 -v`.
+
+### Options
+
+`--semver` : _(Optional)_ Use semver semantics to check version compatibility
+for project dependencies of LoopBack modules
```
bajtos: I find this little bit puzzling. What does semver semantics mean in this case? What is the default behavior (when the option is not present), how is it different from semver? What criteria can our users apply to decide which mode to choose?

docs/site/Update-generator.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Output
+
+The following is a sample output from `lb4 update` against a LoopBack 4 project
+generated by older version of CLI:
+
+```
+The project was originally generated by @loopback/cli@<1.0.1>.
```
bajtos: I'd like you to clarify this point:

What happens after we run `npm update` and then a new version of `@loopback/cli` is published - are we going to print the CLI version that was used to generate the project or the last CLI version that was executed to update dependencies?

For example:

- Project created using `@loopback/cli@<1.0.1>`.
- Later updated by running `@loopback/cli` version `1.2.0`.
- Now I have `@loopback/cli` version `1.3.0` installed and run `lb4 update`.

What will the tool print?

BTW what do the angle brackets around the version mean?

packages/cli/test/integration/lib/check-loopback-project.suite.js
```diff
@@ -9,6 +9,44 @@ const sinon = require('sinon');
...
+
+/***
+ * Parse version range (for example, `'^1.24.3'`) to an object such as
+ * `{major: 1, minor: 24, patch: 3}`
+ *
+ * @param {string} versionRange
+ */
+function parseVersion(versionRange) {
+  const result = /(\^|\~)(\d+)\.(\d+)\.(\d+)/.exec(versionRange);
+  return {major: +result[2], minor: +result[3], patch: +result[4]};
```
bajtos: Have you considered using `semver.parse`?

Quoting from https://www.npmjs.com/package/semver#functions:

> `parse(v)`: Attempt to parse a string as a semantic version, returning either a `SemVer` object or `null`.

raymondfeng: Unfortunately `parse` does not allow a version range such as `^1.0.0`.

docs/site/Update-generator.md
raymondfeng: By default, `lb4 update` checks exact match of versions between the project dependencies and the ones from the running cli template. Settting `--semver` relaxes it by semver checks, for example, `^1.0.0` won't be updated as it's compatible with `1.0.1`.

PR #4056, opened 2019-11-07
docs:  migrate Express middleware from LB3 to LB4

hacksparrow:
Document how to migrate middleware from LB3 to LB4. Addresses https://github.com/strongloop/loopback-next/issues/3947.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4091, opened 2019-11-09
chore: enable esModuleInterop compiler option

raymondfeng:
The flag allows `import fs from 'fs'`. It's highly recommended by TypeScript.
See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html

## Checklist

### DIFF
--- a/packages/build/config/tsconfig.common.json
+++ b/packages/build/config/tsconfig.common.json
@@ -14,6 +14,7 @@
 
     "lib": ["es2018", "esnext.asynciterable"],
     "module": "commonjs",
+    "esModuleInterop": true,
     "moduleResolution": "node",
     "target": "es2017",
     "sourceMap": true,
--- a/packages/cli/generators/datasource/templates/datasource.ts.ejs
+++ b/packages/cli/generators/datasource/templates/datasource.ts.ejs
@@ -5,7 +5,7 @@ import {
   ValueOrPromise,
 } from '@loopback/core';
 import {juggler} from '@loopback/repository';
-import * as config from './<%= jsonFileName %>';
+import config from './<%= jsonFileName %>';
 
 @lifeCycleObserver('datasource')
 export class <%= className %>DataSource extends juggler.DataSource
--- a/benchmark/src/__tests__/benchmark.integration.ts
+++ b/benchmark/src/__tests__/benchmark.integration.ts
@@ -4,7 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {expect} from '@loopback/testlab';
-import * as request from 'request-promise-native';
+import request from 'request-promise-native';
 import {Benchmark} from '..';
 import {Autocannon, EndpointStats} from '../autocannon';
 
--- a/benchmark/src/autocannon.ts
+++ b/benchmark/src/autocannon.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import * as assert from 'assert';
+import assert from 'assert';
 const autocannon = require('autocannon');
 
 export interface EndpointStats {
--- a/benchmark/src/benchmark.ts
+++ b/benchmark/src/benchmark.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import * as byline from 'byline';
+import byline from 'byline';
 import {ChildProcess, spawn} from 'child_process';
 import pEvent, {Emitter} from 'p-event';
 import {Autocannon, EndpointStats} from './autocannon';
--- a/benchmark/src/client.ts
+++ b/benchmark/src/client.ts
@@ -3,8 +3,8 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import * as request from 'request-promise-native';
 import {Todo} from '@loopback/example-todo';
+import request from 'request-promise-native';
 
 export class Client {
   constructor(private url: string) {}
--- a/docs/site/DataSources.md
+++ b/docs/site/DataSources.md
@@ -37,7 +37,7 @@ Example DataSource Class:
 ```ts
 import {inject} from '@loopback/core';
 import {juggler} from '@loopback/repository';
-import * as config from './db.datasource.config.json';
+import config from './db.datasource.config.json';
 
 export class DbDataSource extends juggler.DataSource {
   static dataSourceName = 'db';
--- a/docs/site/Defining-the-API-using-design-first-approach.shelved.md
+++ b/docs/site/Defining-the-API-using-design-first-approach.shelved.md
@@ -235,7 +235,7 @@ npm install --save lodash
 {% include code-caption.html content="/apidefs/product.api.ts" %}
 
 ```ts
-import * as _ from 'lodash';
+import _ from 'lodash';
 
 // Assuming you have created the "base" schema elsewhere.
 // If there are no common properties between all of the endpoint objects,
@@ -336,8 +336,8 @@ put them all together to produce the final OpenAPI spec.
 
 ```ts
 import {ProductAPI, DealAPI, CategoryAPI} from '../apidefs';
-import * as OpenApiSpec from '@loopback/openapi-spec';
-import * as _ from 'lodash';
+import OpenApiSpec from '@loopback/openapi-spec';
+import _ from 'lodash';
 
 // Import API fragments here
 
--- a/docs/site/Repositories.md
+++ b/docs/site/Repositories.md
@@ -335,7 +335,7 @@ dependency injection.
    ```ts
    import {inject} from '@loopback/core';
    import {juggler, AnyObject} from '@loopback/repository';
-   import * as config from './redis.datasource.config.json';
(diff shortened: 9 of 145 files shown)


packages/build/config/tsconfig.common.json
```diff
@@ -14,6 +14,7 @@
 
     "lib": ["es2018", "esnext.asynciterable"],
     "module": "commonjs",
+    "esModuleInterop": true,
```
bajtos: Quoting from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop:

> Note: The new behavior is added under a flag to avoid unwarranted breaks to existing code bases. We highly recommend applying it both to new and existing projects. For existing projects, namespace imports (`import * as express from "express"; express();`) will need to be converted to default imports (`import express from "express"; express();`).

By enabling `esModuleInterop` in the shared tsconfig file, we are forcing this breaking change on all existing LoopBack projects. At minimum, I think this change should be released in a semver-major version of `@loopback/build`.

I have also a question about the impact of this change on code consuming LoopBack modules, especially when the consumer (the app project) and the producer (us, loopback-next) have different setting of `esModuleInterop`.

- Use case 1:

  The application is using an older version of LB4 packages, where the javascript files in `dist` were created using our current TypeScript config (`esModuleInterop` disabled); and uses the current shared TypeScript config.

  Now the application decides to enable `esModuleInterop`, but does not update LB packages. 

  What will happen? Will existing imports in the application continue to work?

- Use case 2:

  Same initial setup as above, but now the application decides to keep `esModuleInterop` disabled, but update LB packages to the latest versions, where `dist` is created using `esModuleInterop` enabled.

  What will happen? Will existing imports in the application continue to work?

Depending on the answer to these questions, we may need to bump semver-major version in all packages, in addition to `@loopback/build`.

If this change is backwards-compatible as far as the consumers of our packages are concerned, then we need to split the change in two commits: a semver-major change in `@loopback/build` as the first commit, and then a standalone "chore" or "refactor" (?) commit updating our codebase to accommodate the new compiler settings.

packages/cli/generators/datasource/templates/datasource.ts.ejs
```diff
@@ -5,7 +5,7 @@ import {
   ValueOrPromise,
 } from '@loopback/core';
 import {juggler} from '@loopback/repository';
-import * as config from './<%= jsonFileName %>';
+import config from './<%= jsonFileName %>';
```
bajtos: What will happen when we run `lb4 datasource` using the updated `lb4` CLI in and application that was created from an older CLI version and therefore has `esModuleInterop` disabled?

raymondfeng:
> Use case 1:

> The application is using an older version of LB4 packages, where the javascript files in dist were created using our current TypeScript config (esModuleInterop disabled); and uses the current shared TypeScript config.

> Now the application decides to enable esModuleInterop, but does not update LB packages.

> What will happen? Will existing imports in the application continue to work?

The `esModuleInterop` only changes the JS code emitted by the TypeScript compiler to ensure `default import` works with non-esModule, such as `debug`.

```js
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
const debug_1 = __importDefault(require("debug"));
const debug = debug_1.default('loopback:context');
```

I'm fairly sure it does not impact the consumers of our modules.

> Use case 2:

> Same initial setup as above, but now the application decides to keep esModuleInterop disabled, but update LB packages to the latest versions, where dist is created using esModuleInterop enabled.

> What will happen? Will existing imports in the application continue to work?

That should be fine. Once the LB4 packages are published to npm, they won't impact how they are consumed.

raymondfeng:
> What will happen when we run lb4 datasource using the updated lb4 CLI in a project that was created from an older CLI version and therefore has esModuleInterop disabled?

Upgrading to newer versions of LB4 modules for a project generated by an old CLI version is not guaranteed to be compatible.

raymondfeng:
Based on the analysis above, I think we can do the following:

1. Separate the `@loopback/build` change to a commit marked as `BREAKING CHANGES`
2. Leave other changes in one commit - non-breaking

bajtos:
Great discussion!

> > What will happen when we run lb4 datasource using the updated lb4 CLI in a project that was created from an older CLI version and therefore has esModuleInterop disabled?
>
> Upgrading to newer versions of LB4 modules for a project generated by an old CLI version is not guaranteed to be compatible.

I find such approach problematic for user experience. How are our users supposed to know when it's ok to use newer versions of LB4 modules and when it's not ok? In case it's not ok, then how are they supposed to know how to upgrade LoopBack dependencies in their projects ?

I think we need to release the changes in CLI templates as semver-major release too. That way we can clearly communicate to our users that the new CLI version does not work with existing projects (scaffolded by an older CLI version) and offer them a migration guide for upgrading their existing CLI projects to support the new CLI version.

> Based on the analysis above, I think we can do the following:
> 
> 1. Separate the `@loopback/build` change to a commit marked as `BREAKING CHANGES`
> 2. Leave other changes in one commit - non-breaking

I am proposing a slightly different plan:

1. Separate the `@loopback/build` change to a commit marked as `BREAKING CHANGES`
2. Separate the `@loopback/cli` change to a commit marked as `BREAKING CHANGES`
3. Leave other changes in one commit - non-breaking

Thoughts?

/cc @strongloop/loopback-next

raymondfeng:
Update to now, we have released `@loopback/cli` with patch/minor versions as the compatibility expectation is not set above projects generated by different versions of `lb4` command. Please also note that `@loopback/cli` is installed globally for most users and any upgrade requires the user to explicitly run `npm i -g @loopback/cli`. 

The strict semver requirement for `@loopback/cli` should be determined by `packages/cli` itself as well as changes made to `packages/cli/package.json[config][templateDependencies]`.

raymondfeng:
@jannyHou The commit message already have the accurate link.

bajtos:
> Update to now, we have released `@loopback/cli` with patch/minor versions as the compatibility expectation is not set above projects generated by different versions of `lb4` command. Please also note that `@loopback/cli` is installed globally for most users and any upgrade requires the user to explicitly run `npm i -g @loopback/cli`. 
> 
> The strict semver requirement for `@loopback/cli` should be determined by `packages/cli` itself as well as changes made to `packages/cli/package.json[config][templateDependencies]`. 

I am not sure if I fully understand what you are saying and if we are on the same page.

Let's discuss the developer experience our users are going to encounter.

1. We have the latest version of `@loopback/cli` (as of now) and our LB4 project has been updated to use the latest version of runtime dependencies.

2. LB4 releases new versions and we want to upgrade. To keeps things simple, let's assume we upgrade the global install of `@loopback/cli` first and then run `lb4 update` (see https://github.com/strongloop/loopback-next/pull/4080) to update dependencies in our projects.

3. I believe we end up with the following setup:

   - `@loopback/build` is updated to the new version which turns on `esModuleInterop` flag.
   - Our existing code is still using the old syntax `import * as foo from 'foo'`.

4. Now if we happen to use express (e.g. as recommded in https://loopback.io/doc/en/lb4/express-with-lb4-rest-tutorial.html and #4056), then the application will stop working.

How is the user going to find out what's wrong and how to fix the problem?

As I was thinking about this aspect more, I agree that a semver-major version of loopback-cli is probably not needed, because if the user happens to run a new version of loopback-cli that assumes `esModuleInterop` to be enabled, in a project that's using the old TypeScript config with `esModuleInterop` disabled (via `@loopback/build` template), then the user will get a warning about possible incompatibilities between the CLI version and the installed dependencies.

It would be better if this was not just a generic warning that is usually a false alert to be ignored, but an actionable message pointing users to some online resource (release notes, migration guide in README, etc.) that will explain what has changed, when and how to upgrade.

bajtos:
If we don't get this right, then we will end up with confused users, as can be already seen here: https://github.com/strongloop/loopback-next/issues/4156

bajtos:
@raymondfeng I see that you made some changes, but don't see any comment. Can you please explain what did you change since the last time I reviewed this patch and how did you address my comments above?

packages/cli/generators/datasource/templates/datasource.ts.ejs
raymondfeng: The `@loopback/build` module will be released as a new major. Running `lb4 datasource` against an old project prompts users about the compatibility issue.

raymondfeng:
@bajtos The latest version of this PR has 3 commits:

1. Enable `esModuleInterop` for `@loopback/build` - BREAKING CHANGE
2. Update cli templates
3. Update existing code

bajtos:
@raymondfeng the split into three commits have been already discussed in the past. I am concerned about impact on existing projects and how are users going to learn about the migration path required after `npm update` or `lb4 update`. See my comment https://github.com/strongloop/loopback-next/pull/4091#issuecomment-555537424, especially this part:

> I believe we end up with the following setup:
>
>    - `@loopback/build` is updated to the new version which turns on `esModuleInterop` flag.
>   - Our existing code is still using the old syntax `import * as foo from 'foo'`.
>
>  Now if we happen to use express (e.g. as recommded in https://loopback.io/doc/en/lb4/express-with-lb4-rest-tutorial.html and #4056), then the application will stop working.
>
> **How is the user going to find out what's wrong and how to fix the problem?**

PR #4112, opened 2019-11-12
fix(cli): handle missing target artifact dir

bajtos:
When creating a new artifact (e.g. Model or Repository) while the target directory does not exist yet (e.g. `src/models` or `src/repositories`), we used to abort CLI with a ENOENT error.

This commit is fixing our CLI to treat the missing directory as if it was empty and continue the operation if possible.

## Checklist

### DIFF
--- a/packages/cli/lib/utils.js
+++ b/packages/cli/lib/utils.js
@@ -224,13 +224,21 @@ exports.findArtifactPaths = async function(dir, artifactType, reader) {
   const readdir = reader || readdirAsync;
   debug(`Finding artifact paths at: ${dir}`);
 
-  // Wrapping readdir in case it's not a promise.
-  const files = await readdir(dir);
-  return _.filter(files, f => {
-    return (
-      _.endsWith(f, `${artifactType}.js`) || _.endsWith(f, `${artifactType}.ts`)
+  try {
+    // Wrapping readdir in case it's not a promise.
+    const files = await readdir(dir);
+    return files.filter(
+      f =>
+        _.endsWith(f, `${artifactType}.js`) ||
+        _.endsWith(f, `${artifactType}.ts`),
     );
-  });
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      // Target directory was not found (e.g. "src/models" does not exist yet).
+      return [];
+    }
+    throw err;
+  }
 };
 /**
  * Parses the files of the target directory and returns matching JavaScript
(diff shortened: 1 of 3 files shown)


packages/cli/lib/utils.js
```diff
@@ -224,13 +224,21 @@ exports.findArtifactPaths = async function(dir, artifactType, reader) {
...
+    // Wrapping readdir in case it's not a promise.
+    const files = await readdir(dir);
+    return files.filter(
+      f =>
+        _.endsWith(f, `${artifactType}.js`) ||
+        _.endsWith(f, `${artifactType}.ts`),
     );
-  });
+  } catch (err) {
+    if (err.code === 'ENOENT') {
```
raymondfeng: I would prefer to check if the directory exists in advance instead of using the err code.

bajtos: I find that approach problematic, because it introduces a race condition - between the time we check if the directory exists, and the time we read the entries, the directory can be deleted or removed.

What are your concerns about my proposed approach?

raymondfeng: I don't think we need to deal with the edge case where the directory is removed when the command in running. If so, we'll have to use sync fs apis.

My main concern is to rely on the error code. It would be more authoritative by calling an api to check existence of the directory.

It's not a blocker for the PR, anyway.

packages/cli/test/integration/generators/model.integration.js
```diff
@@ -119,6 +117,21 @@ describe('lb4 model integration', () => {
       basicModelFileChecks(expectedModelFile, expectedIndexFile);
     });
 
+    it('creates "src/models" directory if it does no exist', async () => {
```
dhmlau: typo: `does no exist` -> `does not exist`

packages/cli/lib/utils.js
```diff
@@ -224,13 +224,21 @@ exports.findArtifactPaths = async function(dir, artifactType, reader) {
...
+      // Target directory was not found (e.g. "src/models" does not exist yet).
+      return [];
```
dhmlau: Question: I assume if the target directory does not exist, it will get created when creating the artifact?

bajtos: Yes, that's correct.

packages/cli/test/integration/generators/model.integration.js
raymondfeng: I fixed it as part of rebase.

Issue #3336, opened 2019-07-11
Filter on findById is ignoring the where clause


mmbabel:
## Steps to reproduce

Submit to any findById call the optional filter argument with a where clause.

## Current Behavior

The where clause is beeing ignored and the item is returned in all cases.

## Expected Behavior

The item will only be returned if it matches the where clause.

## Additional information

This is needed to add item level security based on filters by an Interceptor.

## Related Issues

https://github.com/strongloop/loopback-next/issues/1721
Although this issue is about the include and fields of the filter

## Acceptance Criteria
_Getting from @bajtos' [comment](https://github.com/strongloop/loopback-next/issues/3336#issuecomment-510797763)
- [ ] fix TypeScript definitions and our API docs to make it clear which filter properties are accepted by findById function.

dougal83:
I think you'll find that is by design.  `findById` is simply finding the item with the supplied `Id`.  I suggest that you use `find` and supply the where clause along with the item `Id` for your case.

mmbabel:
When it takes a filter as argument I expect it uses this filter object.

I don't know how findById works internally, but from the signature it looks like a shortened version for find with the where, which has the benefit, that you do not need to know the name of the id.

For now I am doing excatly that overwriting the findById function in all my repos with a version that uses find with the id added to the where clause.

dhmlau:
@mmbabel , are you using the API Explorer to test?

mmbabel:
No, I am using Postman to issue the HTTP requests. 
The filter is also not coming by the request at all, the filter gets populated by an Interceptor and I tested that the filter is there.

As said, as a workaround I am now overwriting the standard findById in all repositories like in this example:

```typescript
async findById(
  id: typeof Firm.prototype.id,
  filter?: Filter<Firm>,
  options?: any,
): Promise<Firm> {
    if(filter === undefined) {
      filter = (new FilterBuilder()).build() as Filter<Firm>;
    }

    if(filter.where === undefined) {
      filter.where = (new WhereBuilder({FirmID: id})).build();
    } else {
      filter.where = (new WhereBuilder(filter.where)).and({FirmID: id}).build();
    }

    const res  = await this.findOne(filter, options);
    if(!res) { console.log(id);
      try {
        await super.findById(id);
      } catch (e) {
        throw e;
      }
      throw new HttpErrors.Forbidden("You do not have permissions to view this item.");
    }
    return res;
  }
```

this works but is ugly,

[...]

Issue #1721, opened 2018-09-18
Add support for "include" and "fields" to findById (REST API)


bajtos:
Let's say our application has a `Customer` model that has many `Order` instances.

In LoopBack 3.x, one can can fetch a customer with all their orders.

```
GET /api/customers/123?filter[include]=orders
```

We should support this feature in LoopBack 4 too.

## Acceptance criteria

 - [ ] Find which `Filter` properties are supported by `findById`
 - [ ] Create a new TypeScript interface describing these properties. This new interface should be most likely provided by loopback-datasource-juggler.
 - [ ] Modify `EntityCrudRepository.prototype.findById` - fix the type of the `filter` argument to use the new TypeScript type describing only the properties supported by `findById`. Fix all implementations of this interface, include tests. 
 - [ ] Create a function for obtaining JSON Schema and OpenAPI schema for this new interface. Build on top of #1679
- [ ] Update [Sequence](https://loopback.io/doc/en/lb4/Sequence.html) that we support these two fields for `findById`.
- [ ] Throw errors ( or warning) when use `order`, `skip`, `limit`, and `where` clauses with `findById`. See https://github.com/strongloop/loopback-next/issues/3336.
--------------------
Done in https://github.com/strongloop/loopback-next/pull/4114
 - Update the REST Controller template and all example controllers to support this new `filter` parameter.
 - Update any documentation pages  and snapshot tests showing `findById` implementation to include the new `filter` argument.

_Note: the argument name `filter` may not be descriptive enough, feel free to find a better name._

**IMPORTANT: This feature depends on https://github.com/strongloop/loopback-next/issues/1352.**

mrbatista:
@bajtos is duplicate of #1352?
In loopback 3.x is it possible to use include for `find()`, `findById()`, `findOne()`.
Does loopback 4 support only `findById()`?

bajtos:
> In loopback 3.x is it possible to use include for find(), findById(), findOne().
> Does loopback 4 support only findById()?

LoopBack 4 does not support inclusion of related models yet. The issue #1352 is keeping track of making `filter[include]` work for `find` and `findOne` and it's a pre-requisite for this issue.

tayyabjamil:
any updates?

PR #4114, opened 2019-11-12
feat(cli): update controller template to enable filter for findById endpoint

agnes512:
As LB4 supports inclusion, I think it's fair to allow users to traverse related models with `findById()` endpoint with CLI generated controllers.

Also, the [todoList example](https://github.com/strongloop/loopback-next/blob/master/examples/todo-list/src/controllers/todo-list.controller.ts#L129) enables such filters. I think they should be consistent as we have a note in [the site](https://loopback.io/doc/en/lb4/todo-list-tutorial.html):
```
This page was generated from the loopback-next/examples/todo-list/README.md
``` 

## Checklist

### DIFF
--- a/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
+++ b/packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
@@ -70,7 +70,10 @@ export class <%= className %>Controller {
         description: 'Array of <%= modelName %> model instances',
         content: {
           'application/json': {
-            schema: {type: 'array', items: getModelSchemaRef(<%= modelName %>)},
+            schema: {
+              type: 'array',
+              items: getModelSchemaRef(<%= modelName %>, {includeRelations: true}),
+            },
           },
         },
       },
@@ -108,12 +111,19 @@ export class <%= className %>Controller {
     responses: {
       '200': {
         description: '<%= modelName %> model instance',
-        content: {'application/json': {schema: getModelSchemaRef(<%= modelName %>)}},
+        content: {
+          'application/json': {
+            schema: getModelSchemaRef(<%= modelName %>, {includeRelations: true}),
+          },
+        },
       },
     },
   })
-  async findById(@param.path.<%= idType %>('id') id: <%= idType %>): Promise<<%= modelName %>> {
-    return this.<%= repositoryNameCamel %>.findById(id);
+  async findById(
+    @param.path.<%= idType %>('id') id: <%= idType %>,
+    @param.query.object('filter', getFilterSchemaFor(<%= modelName %>)) filter?: Filter<<%= modelName %>>
+  ): Promise<<%= modelName %>> {
+    return this.<%= repositoryNameCamel %>.findById(id, filter);
   }
 
   @patch('<%= httpPathName %>/{id}', {
--- a/packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
+++ b/packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
@@ -80,7 +80,10 @@ export class ProductReviewController {
         description: 'Array of ProductReview model instances',
         content: {
           'application/json': {
-            schema: {type: 'array', items: getModelSchemaRef(ProductReview)},
+            schema: {
+              type: 'array',
+              items: getModelSchemaRef(ProductReview, {includeRelations: true}),
+            },
           },
         },
       },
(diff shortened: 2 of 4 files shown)


agnes512:
I also have a question, in https://github.com/strongloop/loopback-next/blob/master/examples/todo-list/src/controllers/todo-list.controller.ts#L80, we use the flag `includeRelations`. @nabdelgadir told me that this is for openAPI spec. But I don't see any documentation for it. Do we want to add it somewhere?

jannyHou:
@agnes512 see code in [buildSchema](https://github.com/strongloop/loopback-next/blob/cd5ca92c368ae35bc10d8847b3b0d379f7196544/packages/repository-json-schema/src/build-schema.ts#L309-L310)

When that flag is added, the openapi spec of the response schema will include relational schemas, otherwise the included items in response will be treated as unknown properties :p

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
```diff
@@ -112,8 +112,11 @@ export class <%= className %>Controller {
       },
     },
   })
-  async findById(@param.path.<%= idType %>('id') id: <%= idType %>): Promise<<%= modelName %>> {
-    return this.<%= repositoryNameCamel %>.findById(id);
+  async findById(
```
jannyHou: based on your question in [comment](https://github.com/strongloop/loopback-next/pull/4114#issuecomment-553091517), the response schema should also be updated to 
```
content: {
  `application/json`: {
      schema: getModelSchemaRef(<%= modelName %>, includeRelations),
    }
}
```

packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
```diff
@@ -122,8 +122,11 @@ export class ProductReviewController {
       },
     },
   })
-  async findById(@param.path.number('id') id: number): Promise<ProductReview> {
-    return this.barRepository.findById(id);
+  async findById(
```
jannyHou: ditto https://github.com/strongloop/loopback-next/pull/4114/files#r346005700

packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
```diff
@@ -291,8 +294,11 @@ export class ProductReviewController {
...
```
jannyHou: ditto https://github.com/strongloop/loopback-next/pull/4114/files#r346005700

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
```diff
@@ -70,7 +70,10 @@ export class <%= className %>Controller {
         description: 'Array of <%= modelName %> model instances',
         content: {
           'application/json': {
-            schema: {type: 'array', items: getModelSchemaRef(<%= modelName %>)},
+            schema: {
+              type: 'array',
+              items: getModelSchemaRef(<%= modelName %>, {includeRelations: true}),
```
bajtos: 👍

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
```diff
@@ -108,12 +111,19 @@ export class <%= className %>Controller {
...
+          },
+        },
       },
     },
   })
-  async findById(@param.path.<%= idType %>('id') id: <%= idType %>): Promise<<%= modelName %>> {
-    return this.<%= repositoryNameCamel %>.findById(id);
+  async findById(
+    @param.path.<%= idType %>('id') id: <%= idType %>,
+    @param.query.object('filter', getFilterSchemaFor(<%= modelName %>)) filter?: Filter<<%= modelName %>>
```
bajtos: Do we support all filter properties in `findById`? For example, I find it weird to combine `findById` with `where` or `limit`/`skip` filter properties, because `findById` always returns a single record matching the given id only. Should we perhaps describe only `filter.include` and `filter.fields` properties in the schema?

bajtos: See also https://github.com/strongloop/loopback-next/issues/3336 _Filter on findById is ignoring the where clause_.

I am concerned that if we describe `filter` parameter of `findById` method as accepting all filter properties, then our users will try to set those (unsupported) properties and get surprised/confused.

bajtos:
@agnes512 I think this pull request is implementing some parts of https://github.com/strongloop/loopback-next/issues/1721, could you please check and update that issue to make it clear what remains to be done?

packages/cli/generators/controller/templates/src/controllers/controller-rest-template.ts.ejs
agnes512: @bajtos I think `findById` only works for inclusion atm.

And I agree that the filter for `findById` should work for `inclusion` and `field` as you described in https://github.com/strongloop/loopback-next/issues/1721. Order, limit, where, etc. don't quite make sense to me. 

I'd like to land this PR first, then we can limit it in #1721 to fix #3336. 

I will try to document this issue in our doc ( I still couldn't find a best place for it) when I update inclusion docs.

Issue #4089, opened 2019-11-09
log extension within the loopback/extensions


fidgi:
## Suggestion

It would be great to have a log extension (something similar to the [log-extension example](https://github.com/strongloop/loopback-next/tree/master/examples/log-extension) within the supported loopback extensions

## Use Cases

Just to be able manage log level indevelopment and production mode

## Examples

see  [log-extension example](https://github.com/strongloop/loopback-next/tree/master/examples/log-extension)

## Acceptance criteria

TBD - will be filled by the team.

marioestradarosa:
@fidgi , you mean something like this ?  

https://github.com/strongloop/loopback-next/issues/3975#issuecomment-549667967

fidgi:
@marioestradarosa What I'm looking for is a mixin to ease the integration of a logging framework such as winston and an option in loopback/cli to enable this mixin at app creation time .

marioestradarosa:
@fidgi correct!, I just wanted to make sure that's final log functionality you were looking for. Indeed that could be a good feature "external extension/package"  to leverage the extensibility of `looback 4` framework.  I will try to write an external extension similar to the example I provided for winston.

fidgi:
@raymondfeng thank for working on this

bajtos:
_Cross-posted from https://github.com/strongloop/loopback-next/pull/4117#issuecomment-554256489_

I'd like to start the discussion with building a better understanding of different loggers and log collectors that we can choose from. 

@raymondfeng Why did you pick winston and fluentd in particular?

Personally, I prefer to use `pino` for logging, it's more modern than winston/bunyan and is significantly more performant (2x faster according to pino's [benchmarks](https://github.com/pinojs/pino/blob/HEAD/docs/benchmarks.md)).

I am not familiar with log collectors. I see that [fluentd](https://www.fluentd.org) is a part of Cloud Native Computing Foundation, I guess that makes it a good candidate.

@strongloop/loopback-next thoughts?

fidgi:
I'm fine with any logging framework that is extensible and that provide a way to hide sensible data in the logs depending of the log level. For what I know winston is ok.
fluentd  is also a good choice for us as we planned to use it to collect logs from our loopback-based microservices running in Kubernetes.

raymondfeng:
For the record, [Pino](https://github.com/pinojs/pino) seems to focus more on efficiency but its extensibility has room for improvement, for example, formatter to transform logs and transports to collect logs to physical destinations.

[...]

PR #4117, opened 2019-11-13
feat(extension-logging): add integration with winston and fluentd logging

raymondfeng:
Experimental feature for https://github.com/strongloop/loopback-next/issues/4089

See docs at https://github.com/strongloop/loopback-next/blob/logging/extensions/logging/README.md.

## Checklist


bajtos:
I'd like to start the discussion with building a better understanding of different loggers and log collectors that we can choose from. 

@raymondfeng Why did you pick winston and fluentd in particular?

Personally, I prefer to use `pino` for logging, it's more modern than winston/bunyan and is significantly more performant (2x faster according to pino's [benchmarks](https://github.com/pinojs/pino/blob/HEAD/docs/benchmarks.md)).

I am not familiar with log collectors. I see that [fluentd](https://www.fluentd.org) is a part of Cloud Native Computing Foundation, I guess that makes it a good candidate.

extensions/logging/src/__tests__/fixtures/fluentd.docker.ts
```diff
@@ -0,0 +1,63 @@
...
+import {GenericContainer, StartedTestContainer} from 'testcontainers';
+
+export const ROOT_DIR = path.join(__dirname, '../../../fixtures');
+export const ETC_DIR = path.join(ROOT_DIR, 'etc');
+
+async function startFluentd() {
+  const container = await new GenericContainer(
+    'fluent/fluentd',
+    'v1.7.4-debian-1.0',
+  )
```
bajtos: Please don't start Docker containers from the tests. It slows down the test suite that's already too slow. If our tests need an external service to be running, then please provide a script that the developers can run ONCE to start all services at the beginning of their coding session. See what we have already in place in `acceptance/repository-*` directories and in individual connectors like [loopback-connector-mysql](https://github.com/strongloop/loopback-connector-mysql/blob/9860f0339f171423c4260ae999cabcc36cb0c1ce/setup.sh).

As we discussed when working on the monitoring/Prometheus integration, it's often a good practice to mock the external dependency (i.e. the log collector) for integration tests to make them faster and then write only  few acceptance/end-to-end tests that are talking to the real service started in Docker.

bajtos:
Also, is it a good idea to have our component coupled to Winston and Fluent? The tricky part is to strike the good balance between making the framework easy to use by prescribing the recommended way, but stay flexible enough to support different use cases.

I think it should be ok to hard-code the logging framework we use under the hood, we can be prescriptive here. On the other hand, I'd like the log outputs (transports) to be as flexible as possible - ideally our users should have the option to choose any transport they like, and have an easy (and well-documented) way how to configure this aspect of logging.

Typically, I would like to print my logs to stdout (console) while running the service under `npm test`. When running the service locally on my developer machine, I would ideally like to see the logs in my console too, but I am not sure if that's feasible when running the service via Docker or in a Kubernetes cluster.

raymondfeng:
The PoC starts with `Fluentd` and `Winston` for a few reasons:

1. Fluentd is part of CNCF - which aligns with our Cloud Native integration direction, such as `health` endpoints for k8s, `metrics` for `prometheus`, and `distributed tracing` with `jaeger`. 

2. Winston is a mature logging framework in Node and it has the key extensibility such as formatters and transports. We leverage them to enable the extension point. 

3. Fluentd Node.js client happens to have a transport for Winston.

4. I was peeking `Pino` as well.

I didn't want to start our abstraction for the common logging building blocks. That's why I picked popular frameworks to test drive.

raymondfeng:
@bajtos I have refactored the docker-based acceptance tests into its own package. A separate task is added for travis to run the fluentd tests. PTAL.

acceptance/extension-logging-fluentd/src/__tests__/fixtures/fluentd.docker.ts
```diff
@@ -0,0 +1,66 @@
...
+  return container;
+}
+
+let fluentd: StartedTestContainer | undefined;
+
+/* eslint-disable no-invalid-this */
+/**
+ * Root-level before hook to start Fluentd container
+ */
+before(async function() {
```
bajtos: While this is better in the sense that we pay the price of Docker container/image startup only when explicitly running acceptance tests for fluentd, I find it still as a suboptimal developer experience.

Let's say there is a problem in the integration between our logging extension and the real Fluentd service that I want to troubleshoot. To do so, I may try different approaches:

1. Start Fluentd in a container, modify one of our example apps to use our logging extension, debug the app & extension to see what's going one.

2. Add additional logging, perhaps enable debug logs, and then run the acceptance tests in this directory. Tweak the code, re-run the tests to see if the problem is fixed.

In the first example, I don't have any easy & standard way how to start the container, because I cannot run the `before` hook outside of the test suite.

In the second example, I am waiting for Fluentd to start and stop every time I run the test suite. What a waste of my time (and CPU cycles of my computer).

To wrap it up: it's great that you moved Fluentd tests into `acceptance` folder 👍 Now let's extract the code starting (and stopping?) the docker container into a standalone script that we can run only once.

bajtos:
@raymondfeng What's the relation between this work and [examples/log-extension](https://github.com/strongloop/loopback-next/tree/master/examples/log-extension)? I find it confusing to have a "real" log extension and also an "example" log extension. Can we remove the example now? If the example is important because it shows how to build a custom extension, then perhaps we need to rework it into something else?

@dhmlau @strongloop/loopback-next Thoughts?

raymondfeng:
> If the example is important because it shows how to build a custom extension, then perhaps we need to rework it into something else?

Yes, that's the case. We can probably rename the example and make it clear in the README.

raymondfeng:
@bajtos 
> To wrap it up: it's great that you moved Fluentd tests into acceptance folder 👍 Now let's extract the code starting (and stopping?) the docker container into a standalone script that we can run only once.

For CI testing purpose, automatic start/stop a fluentd container makes it easy and it also allows us to run `docker exec` to read the logs to assert our integration happens as expected. I have to find out how to do if the fluentd container is started by a script.

Let's defer the optimization for local testing after we find out how to best verify the logs are sent to the fluentd instance.

agnes512:
Is it possible to make it take less time to run the acceptance test?

raymondfeng:
@agnes512 What's your concern? From https://travis-ci.com/strongloop/loopback-next/jobs/258791183m the tests are run in 3 seconds.

agnes512:
> What's your concern? 

I meant the time it runs the script and tests. I thought the one I checked was almost 3 min ( I was wrong). It seems that the avg is 2:30, which is fine. It'd be good if we can speed up the script though.

raymondfeng:
> I meant the time it runs the script and tests. I thought the one I checked was almost 3 min ( I was wrong). It seems that the avg is 2:30, which is fine. It'd be good if we can speed up the script though.

Travis takes 2-3 mins to run the job, which includes checking out the repo, run `npm i` and `npm run build`.

bajtos:
> Let's defer the optimization for local testing after we find out how to best verify the logs are sent to the fluentd instance.

Do you have any estimate on when will that happen? I am concerned that we will get stuck with slow tests for long time :(

extensions/logging/src/decorators/log.decorator.ts
```diff
@@ -0,0 +1,28 @@
...
+ *   hello(name: string) {
+ *     return `Hello, ${name}`;
+ *   }
+ * }
+ * ```
+ */
+export function log() {
+  // A shortcut to `@intercept` that invokes the winston interceptor that logs
+  // method invocations
+  return intercept(LoggingBindings.WINSTON_INTERCEPTOR);
```
bajtos: Please find a name that better communicates what is this interceptor for. For example: `WINSTON_INVOCATION_ARGS_LOGGING_INTERCEPTOR`.

extensions/logging/src/interceptors/logging.interceptor.ts
```diff
@@ -0,0 +1,123 @@
...
+import {LoggingBindings} from '../keys';
+
+/**
+ * A local interceptor that provides logging for method invocations.
+ */
+@bind({
+  tags: {[ContextTags.KEY]: LoggingBindings.WINSTON_INTERCEPTOR},
+  scope: BindingScope.SINGLETON,
+})
+export class LoggingInterceptor implements Provider<Interceptor> {
```
bajtos: Please find a name that better communicates what this interceptor logs. For example: `LogInvocationArgumentsInterceptor`.

extensions/logging/src/keys.ts
```diff
@@ -0,0 +1,40 @@
...
+  );
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export const FLUENT_SENDER = BindingKey.create<FluentSender<any>>(
+    'logging.fluent.sender',
+  );
+
+  export const WINSTON_LOGGER = BindingKey.create<Logger>(
+    'logging.winston.logger',
+  );
```
bajtos: IIUC, we agreed to build the logging extension around Winston module, making the logger interface hard-coded. I find it redundant to include `WINSTON_` prefix in the keys, because there aren't going to be non-winston keys. We already have `LoggingBindings` as a namespace.

Proposed constant: `LOGGER` and key value: `logging.logger`.

(Same comment applies to other keys below).

extensions/logging/src/keys.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  export const WINSTON_TRANSPORT_FLUENT = BindingKey.create<Transport>(
+    'logging.winston.transports.fluent',
```
bajtos: I find this weird, why do we need the key of the Fluent transport to be public? The transport should be picked by the extension point architecture independently of the key.

extensions/logging/src/interceptors/logging.interceptor.ts
```diff
@@ -0,0 +1,123 @@
...
+    return this.intercept.bind(this);
+  }
+
+  async intercept<T>(
+    invocationCtx: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ) {
+    const reqCtx = await invocationCtx.get<RequestContext>(
+      RestBindings.Http.CONTEXT,
+    );
```
bajtos: What happens if the interceptor is invoked for a proxied method that's not a HTTP handler (controller method/route)?

extensions/logging/src/keys.ts
```diff
@@ -0,0 +1,40 @@
...
+  );
+
+  export const WINSTON_INTERCEPTOR = BindingKey.create<GenericInterceptor>(
+    'logging.winston.interceptor',
+  );
+
+  export const WINSTON_ACCESS_LOGGER = BindingKey.create<GenericInterceptor>(
```
bajtos: Since this is essentially logging HTTP requests, I'd like to include that information in the name. For example:

```suggestion
  export const HTTP_ACCESS_LOGGER = BindingKey.create<GenericInterceptor>(
```

extensions/logging/README.md
```diff
@@ -0,0 +1,158 @@
...
+- FLUENT_SENDER - A fluent sender
+- WINSTON_LOGGER - A winston logger
+- WINSTON_TRANSPORT_FLUENT - A fluent transport for winston
+- WINSTON_INTERCEPTOR - A local interceptor set by `@log` to log method
+  invocations
+- WINSTON_ACCESS_LOGGER - A global interceptor that logs http access with
+  [Morgan](https://github.com/expressjs/morgan) format
+
+The fluent sender and transport for winston can be configured against
+`FLUENT_SENDER`:
```
bajtos: How can the user disable Fluent transport & sender completely, e.g. to use their own transport instead?

extensions/logging/README.md
```diff
@@ -0,0 +1,158 @@
...
+  .to(format.colorize())
+  .apply(extensionFor(WINSTON_FORMAT));
+```
+
+The access log interceptor can also be configured to customize
+[Morgan format and options](https://github.com/expressjs/morgan#morganformat-options):
+
+```
+ctx.configure(LoggingBindings.WINSTON_ACCESS_LOGGER).to({format: 'combined'});
+```
```
bajtos: How is this interceptor applied? IIUC, it's defined as a global interceptor, but IIRC our boot does not load global interceptors from the extensions. If it's automatically registered by other means, then how can the application developer disable this logger?

extensions/logging/README.md
```diff
@@ -0,0 +1,158 @@
...
+facilities:
+
+```ts
+import {inject} from '@loopback/context';
+import {Logger} from '@loopback/extension-logging';
+
+class MyController {
+  // Inject a winston logger
+  @inject(LoggingBindings.WINSTON_LOGGER)
+  private logger: Logger;
```
bajtos: Maybe we should rename the interface to `WinstonLogger` if we want to make it more clear that our Logger has exactly the same interface as provided `Winston`?

extensions/logging/README.md
```diff
@@ -0,0 +1,158 @@
...
+  defaultMeta: {framework: 'LoopBack'},
+});
+```
+
+The winston logger accepts two types of extensions to the following extension
+points:
+
+- WINSTON_TRANSPORT = 'logging.winston.transport'
+- WINSTON_FORMAT = 'logging.winston.format'
+
```
bajtos: I am missing a high-level summary of what is extensible and what is hard-coded, what is provided out-of-the-box and what requires explicit opt-in/setup.

IIUC:

- Winston logger API is hard-coded 🤷‍♂
- It's possible to provide a custom Winston logger instance (🤔 why would anybody want to do that?)
- It's possible to register multiple transports 👍 
- It's possible to register multiple formatters 👍 

We should also answer the following questions
- What formatters are registered by default? How can I configure them? How to disable them?
- What transports are registered by default? How can I configure them? How to disable them?
- What logs are produced by default, e.g. via global interceptors? How can I configure them? How to disable them?

raymondfeng:
> Do you have any estimate on when will that happen? I am concerned that we will get stuck with slow tests for long time :(

From what we can tell from Travis, the fluentd acceptance tests only take 3 seconds to run. The job itself uses 2-3 mins to set up the build (git, npm i, bootstrap etc).

extensions/logging/README.md
raymondfeng: Fixed the README - it can be disabled by setting `enableFluent` to `false` for the logging component.

extensions/logging/src/interceptors/logging.interceptor.ts
raymondfeng: Renamed to WinstonInvocationLogger

extensions/logging/src/keys.ts
raymondfeng: The key is for testing.

extensions/logging/src/decorators/log.decorator.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * export class HelloController {
+ *   @log()
+ *   hello(name: string) {
+ *     return `Hello, ${name}`;
+ *   }
+ * }
+ * ```
+ */
+export function log() {
```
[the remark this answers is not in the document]
raymondfeng: The function name is going to used as a decorator - `@log`. Personally, I would prefer to have short names for decorators.

extensions/logging/README.md
```diff
@@ -0,0 +1,158 @@
...
+class MyController {
+  // Inject a winston logger
+  @inject(LoggingBindings.WINSTON_LOGGER)
+  private logger: Logger;
+
+  // http access is logged by a global interceptor
+  @get('/greet/{name}')
+  // log the `greet` method invocations
+  @log()
```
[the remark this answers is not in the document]
raymondfeng: Added imports.

extensions/logging/src/decorators/log.decorator.ts
raymondfeng: `WINSTON_INVOCATION_LOGGER`

extensions/logging/src/keys.ts
raymondfeng: I intentionally have `winston` in the names to remind our users to follow winston interfaces/types.

extensions/logging/README.md
raymondfeng: The component registers the interceptor. It can be disabled by setting `enableHttpAccessLog` to `false` for the component config.

extensions/logging/README.md
raymondfeng: README updated.

emonddr:
@raymondfeng , great stuff. Coverage has dropped though.

ochemerys:
Hi,
I got logging branch latest from loopback-next/extensions/logging/ 
when i build it I have error:

`src/interceptors/logging.interceptor.ts:81:18 - error TS2551: Property 'GLOBAL_INTERCEPTOR_SOURCE' does not exist on type 'typeof ContextTags'. Did you mean 'GLOBAL_INTERCEPTOR_GROUP'?

81     [ContextTags.GLOBAL_INTERCEPTOR_SOURCE]: 'route',
                    ~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@loopback/context/dist/keys.d.ts:36:11
    36     const GLOBAL_INTERCEPTOR_GROUP = "globalInterceptorGroup";
                 ~~~~~~~~~~~~~~~~~~~~~~~~
    'GLOBAL_INTERCEPTOR_GROUP' is declared here.


Found 1 error.`

Did I miss anything?
Thank you

raymondfeng:
@ochemerys You either have to wait for a release as the PR is still under review or use our sandbox.  See https://github.com/strongloop/loopback-next/tree/master/sandbox to link your application with a local git checkout of `loopback-next`.

ochemerys:
Thank you,
Unfortunately after running
`git clone git@github.com:strongloop/loopback-next.git`
output is 
```
Cloning into 'loopback-next'...
Warning: Permanently added the RSA host key for IP address '140.82.113.3' to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

Probably I have to wait

raymondfeng:
@ochemerys Try the following:

```sh
git clone https://github.com/strongloop/loopback-next.git
```

bajtos:
> From what we can tell from Travis, the fluentd acceptance tests only take 3 seconds to run. The job itself uses 2-3 mins to set up the build (git, npm i, bootstrap etc).

I am more concerned about the developer experience while working locally, see https://github.com/strongloop/loopback-next/pull/4117#discussion_r347971978

extensions/logging/src/decorators/log.decorator.ts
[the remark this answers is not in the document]
bajtos: I agree a short name would be better. We need to find a right balance between a name that's too generic/not enough descriptive (`log`) and a name that's descriptive but too long (`logInvocationArguments`).

How about `logArgs`?

.travis.yml
```diff
@@ -24,9 +24,9 @@ matrix:
   include:
     - node_js: "12"
       os: osx
-      # using the default script to build & run the tests
 
     - node_js: "12"
+      # using the default script to build & run the tests
```
bajtos: Please move this comment back to L27. It's saying that the osx job is using the default build & test script.

I find the comment confusing in the context of code-lint job which uses a custom script instead.

extensions/logging/README.md
```diff
@@ -0,0 +1,190 @@
...
+## Configure the logging component
+
+The logging component can be configured as follows:
+
+```ts
+app.configure(LoggingBindings.COMPONENT).to({
+  enableFluent: false, // default to true
+  enableHttpAccessLog: true, // default to true
+});
+```
```
bajtos: I have mixed feelings about the design where we call `this.component()` in one place and then `app.configure` in another. Personally, I would expect to find the configuration in the same place where the component is mounted, i.e.

```ts
this.component(LoggingComponent, {
   enableFluent: false, // default to true
   enableHttpAccessLog: true, // default to true
});
```

Implementation wise, I can imagine `this.component` calling `app.configure` under the hood.

The biggest issue I see is how to add type safety, to let the compiler understand what configuration options the component allows. IIUC, this problem is not solved for `app.configure` either, therefore it's out of scope of this pull request. I just want to point it out.

On the second thought, I think there is even more important aspect we should consider: which part of the configuration are specific to deployment environment (dev/staging/production) and which are wiring different code bits together (thing of DI)? The former is typically provided from outside of the application, for example consider the port where the REST server will listen on.

In this case, I think both `enableFluent` and `enableHttpAccessLog` option are affecting which code bits are mixed into the application, therefore they should be hard-coded by the app, similarly to how we are hard-coding dependency wirings.

On the other hand, I feel the configuration of the transport and possibly logger is something to configure from outside of the app, depending on the deployment environment.

Anyhow, I guess this can be left out of scope of this initial pull request, because the logging extension is clearly labelled as experimental.

extensions/logging/src/decorators/log.decorator.ts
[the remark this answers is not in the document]
bajtos: I agree a short name would be better. We need to find a right balance between a name that's too generic/not enough descriptive (`log`) and a name that's descriptive but too long (`logInvocationArguments`).

How about `logArgs`?

extensions/logging/src/fluent.ts
```diff
@@ -0,0 +1,59 @@
...
+  private binding: Binding<unknown>;
+
+  async value() {
+    const options = await this.getFluentConfig();
+    if (options == null) {
+      throw new Error(
+        `Fluent is not configured. Please configure ${this.binding.key}.`,
+      );
+    }
+    const cls = fluent.support.winstonTransport();
```
bajtos: The acronym `cls` often means continuation-local-storage. Can you please use a different variable name?

extensions/logging/src/interceptors/logging.interceptor.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  async intercept<T>(
+    invocationCtx: InvocationContext,
+    next: () => ValueOrPromise<T>,
+  ) {
+    const reqCtx = await invocationCtx.get<RequestContext>(
+      RestBindings.Http.CONTEXT,
+      {optional: true},
+    );
+    if (reqCtx == null) return next();
```
bajtos: Is there a test to cover this branch? I could not find it.

raymondfeng:
@bajtos I renamed `@log` to `@logInvocation` - to log invocation for the decorated method, including args, return value, and errors.

extensions/logging/src/interceptors/logging.interceptor.ts
raymondfeng: I reverted this check as now the interceptor is only applied to `route` source.

raymondfeng:
@bajtos PTAL

extensions/logging/README.md
bajtos: I have troubles finding answers to the 3 questions about the default behavior. Can you please point me to relevant place in README?

> - What formatters are registered by default? How can I configure them? How to disable them?
> - What transports are registered by default? How can I configure them? How to disable them?
> - What logs are produced by default, e.g. via global interceptors? How can I configure them? How to disable them?

raymondfeng: See https://github.com/strongloop/loopback-next/blob/logging/extensions/logging/README.md#configure-the-logging-component

emonddr:
Cool stuff, @raymondfeng .

Please fix typo in test folder name:

image

Issue #3770, opened 2019-09-19
[Spike] Investigate how to allow complex object in filter object in API Explorer


dhmlau:
Originated from https://github.com/strongloop/loopback-next/issues/2208.

## Description
swagger-ui, the library powering our API Explorer, unfortunately does not support complex objects in query strings. This spike is to investigate if/how we can allow it. 

_Extracting from https://github.com/strongloop/loopback-next/issues/2208#issuecomment-533008671 from @bajtos_ 

Based on OAI/OpenAPI-Specification#1706 (comment) (cross-posted below), maybe we can switch filter encoding from deepObject style into JSON?

Here is an experiment we can try:

1. Let's keep @param.query.object with no changes for backwards compatibility.
2. Introduce a new helper @param.query.jsonObject that will use JSON encoding instead of deepObject style.
3. Modify the function parseJsonIfNeeded to detect JSON encoding and parse the string value - see packages/rest/src/coercion/coerce-parameter.ts
4. Update our example Todo controller to use @param.query.jsonObject - see examples/todo/src/controllers/todo.controller.ts
5. Start the Todo app, open API Explorer and check how is the filter parameter handled. 

## Acceptance Criteria
<del>- [ ] Review the workaround posted by various users in https://github.com/strongloop/loopback-next/issues/2208 </del>
- [x] Try out the experiment from @bajtos (listed above)

bajtos:
An important aspect to consider: if we switch the spec of `format` argument from deepObject-style to json-content, how is it going to affect the actual REST API? Will it mean that clients will be no longer able to send queries like `find?filter[limit]=1` and will have to always use JSON encoding only (`find?filter={"limit":1}`)?

deepakrkris:
For having the "include" parameter in filter condition `filter["include"]=` does not work whereas passing filter as deep object works.
example:
`http://localhost:3000/todos/1?filter=%7B%22include%22%3A%5B%7B%22relation%22%3A%22todoList%22%7D%5D%7D`

deepakrkris:
url encoding of json can be done with javascript built in function `encodeURIComponent(stringified-json)`

deepakrkris:
In the open api spec (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#encoding-object) , explanation for explode setting is, 
```
When this is true, property values of type array or object generate separate parameters for each value of the array, or key-value-pair of the map. For other types of properties this property has no effect.
```
[...]

Issue #3493, opened 2019-08-01
[Spike/Discussion] The proposal of the new requestBody decorator


jannyHou:
### Description

A follow up story of https://github.com/strongloop/loopback-next/issues/2654
In the spike story https://github.com/strongloop/loopback-next/issues/2654, @jannyHou proposed a new signature `@requestBody(spec, model, schemaOptions)` to simplify the way of providing schema configs, the design is explained in [file](https://github.com/strongloop/loopback-next/blob/5877f0cc0201af59f2a9636a09753207c599ab26/packages/openapi-v3/_spike-request-body.md).

This story is created for further discussion: 

The conversations are mainly tracked in: https://github.com/strongloop/loopback-next/issues/2654, https://github.com/strongloop/loopback-next/pull/3398 and https://github.com/strongloop/loopback-next/pull/3466#discussion_r308650887

There are different proposals from @bajtos (I will summarize Miroslav's opinions later)
    - `@requestBody(specWithOptions, model)`
    - or `@requestBody(spec, getModelSchemaRef(model, options))`
    - or any better signatures

## Acceptance criteria

- [ ] Let's agree on the proposal first
- [ ] If we agree on the original proposal mentioned in [file](https://github.com/strongloop/loopback-next/blob/5877f0cc0201af59f2a9636a09753207c599ab26/packages/openapi-v3/_spike-request-body.md), decide which signatures should be disabled
  - Related discussion: follow the chat in https://github.com/strongloop/loopback-next/pull/3466/files#r308655533

emonddr:
Let's have a meeting to discuss this first.

PR #4141, opened 2019-11-18
feat: spike on API Explorer supporting complex objects in query param

deepakrkris:
https://github.com/strongloop/loopback-next/issues/3770

## Scope:  
Currently loopback supports both exploded and uri encoded complex objects in query params, even without the changes suggested in this PR. The purpose of this PR is to ONLY fix API Explorer to support complex objects in query params.

To test this spike, 
1. Install an example app (eg: todoList)
2. change the param decorators for the filter query in a controller method to @param.query.jsonObject
3. run npm build and copy the installed example app into this PR branch -> loopback-next/sandbox directory
4. run npm i under loopback-next , to create symlinks 
5. cd sandbox/loopback4-example-todoList && npm run start
6. goto API Explorer http://localhost:3000/explorer
7. check invoking controller method that has @param.query.jsonObject works when a filter object is passed

### DIFF
--- a/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts
+++ b/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts
@@ -204,6 +204,17 @@ describe('TodoListApplication', () => {
     });
   });
 
+  it('exploded filter conditions work', async () => {
+    const list = await givenTodoListInstance(todoListRepo);
+    await givenTodoInstance(todoRepo, {title: 'todo1', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo2', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo3', todoListId: list.id});
+
+    const response = await client.get('/todos').query('filter[limit]=2');
+
+    expect(response.body).to.have.length(2);
+  });
+
   /*
    ============================================================================
    TEST HELPERS
--- a/packages/openapi-v3/src/decorators/parameter.decorator.ts
+++ b/packages/openapi-v3/src/decorators/parameter.decorator.ts
@@ -214,6 +214,11 @@ export namespace param {
         in: 'query',
         style: 'deepObject',
         explode: true,
+        content: {
+          'application/json': {
+            schema,
+          },
+        },
         schema,
         ...spec,
       });
--- a/packages/openapi-v3/src/decorators/parameter.decorator.ts
+++ b/packages/openapi-v3/src/decorators/parameter.decorator.ts
@@ -50,8 +50,11 @@ export function param(paramSpec: ParameterObject) {
         // generate schema if `paramSpec` has `schema` but without `type`
         (isSchemaObject(paramSpec.schema) && !paramSpec.schema.type)
       ) {
-        // please note `resolveSchema` only adds `type` and `format` for `schema`
-        paramSpec.schema = resolveSchema(paramType, paramSpec.schema);
+        // If content explicitly mentioned do not resolve schema
+        if (!paramSpec.content) {
+          // please note `resolveSchema` only adds `type` and `format` for `schema`
+          paramSpec.schema = resolveSchema(paramType, paramSpec.schema);
+        }
       }
     }
 
@@ -212,14 +215,11 @@ export namespace param {
       return param({
         name,
         in: 'query',
-        style: 'deepObject',
-        explode: true,
         content: {
           'application/json': {
             schema,
           },
         },
-        schema,
         ...spec,
       });
     },
--- a/packages/rest/src/coercion/coerce-parameter.ts
+++ b/packages/rest/src/coercion/coerce-parameter.ts
@@ -32,7 +32,14 @@ export function coerceParameter(
   data: string | undefined | object,
   spec: ParameterObject,
 ) {
+  if (!spec.schema && spec.content) {
+    const content = spec.content;
+    const jsonSpec = content['application/json'];
+    spec.schema = jsonSpec.schema;
+  }
+
   const schema = spec.schema;
+
   if (!schema || isReferenceObject(schema)) {
     debug(
       'The parameter with schema %s is not coerced since schema' +
@@ -172,7 +179,7 @@ function parseJsonIfNeeded(
 ): string | object | undefined {
   if (typeof data !== 'string') return data;
 
-  if (spec.in !== 'query' || spec.style !== 'deepObject') {
+  if (spec.in !== 'query' || (spec.style !== 'deepObject' && !spec.content)) {
     debug(
       'Skipping JSON.parse, argument %s is not in:query style:deepObject',
       spec.name,
--- a/packages/openapi-v3/src/__tests__/unit/decorators/param/param-query.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/param/param-query.decorator.unit.ts
@@ -235,6 +235,11 @@ describe('Routing metadata for parameters', () => {
           type: 'object',
           additionalProperties: true,
         },
+        content: {
+          'application/json': {
+            schema: {type: 'object', additionalProperties: true},
+          },
+        },
       };
       expectSpecToBeEqual(MyController, expectedParamSpec);
     });
@@ -258,6 +263,17 @@ describe('Routing metadata for parameters', () => {
         in: 'query',
         style: 'deepObject',
         explode: true,
+        content: {
+          'application/json': {
+            schema: {
+              type: 'object',
+              properties: {
+                where: {type: 'object', additionalProperties: true},
+                limit: {type: 'number'},
+              },
+            },
+          },
+        },
         schema: {
           type: 'object',
           properties: {
@@ -305,6 +321,11 @@ describe('Routing metadata for parameters', () => {
       schema: {
         type: 'object',
         additionalProperties: true,
+      },
+      content: {
+        'application/json': {
+          schema: {type: 'object', additionalProperties: true},
+        },
       },
     };
     expectSpecToBeEqual(MyController, expectedParamSpec);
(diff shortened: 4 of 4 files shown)


deepakrkris:
Screen Shot 2019-11-18 at 9 39 21 PM

bajtos:
Thank you for providing test instruction and a screenshot of the outcome. Just FYI, the instructions could be simplified by taking one of our example applications (e.g. `examples/todo`) and changing the param decorators for the filter query in that app.

bajtos:
It's great to see that swagger-ui can handle json-encoded query parameters. Now we need to carefully consider backwards compatibility, see https://github.com/strongloop/loopback-next/issues/3770#issuecomment-533462367.

> An important aspect to consider: if we switch the spec of `format` argument from deepObject-style to json-content, how is it going to affect the actual REST API? Will it mean that clients will be no longer able to send queries like `find?filter[limit]=1` and will have to always use JSON encoding only (`find?filter={"limit":1}`)? Can we find a way how to preserve support for both flavors?

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -189,7 +189,7 @@ describe('TodoApplication', () => {
 
     await client
       .get('/todos')
-      .query({filter: {where: {isComplete: false}}})
+      .query({filter: encodeURIComponent(JSON.stringify({where: {isComplete: false}})) })
```
bajtos: Ideally, both flavors should be supported:
 - `filter: {where: {isComplete: false}}` 
 - `filter: encodeURIComponent(JSON.stringify(...))`

Instead of changing this existing test, we should add a new test to verify the other flavor. I believe these two flavors are already supported now, in which case it would be great to have tests to verify all different ways how to provide the `filter` argument.

bajtos: (This code change & discussion thread is directly related to my previous comment about backwards compatibility.)

bajtos:
Last but not least, please check why Travis CI is failing and try to make all tests pass.

I see the following compilation error, there is also a linting error.

```
packages/openapi-v3/src/decorators/parameter.decorator.ts(252,9): error TS1005: ',' expected.
```

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
bajtos: An idea to consider & try: can we mix `deepObject` style with `json` style in a single parameter? Something along the following lines:

```ts
    object: function(
      name: string,
      schema: SchemaObject | ReferenceObject = {
        type: 'object',
        additionalProperties: true,
      },
      spec?: Partial<ParameterObject>,
    ) {
      schema = {
        type: 'object',
        ...schema,
      };
      return param({
        name,
        in: 'query',

        // deepObject
        style: 'deepObject',
        explode: true,

        // new addition - json encoding
        content: {
           "application/json": {
             schema
           }
         },

         // the rest
        schema,
         ...spec,
       });
     },
```

packages/openapi-v3/src/decorators/parameter.decorator.ts
```diff
@@ -218,8 +218,45 @@ export namespace param {
...
+
+    /**
+     * Define a parameter accepting a url encoded string of a JSON object,
+     *
+     * parameter is created with explode: false
+     * 
+     * @param name - Parameter name
+     * @param schema - Optional OpenAPI Schema describing the JSON object.
+     */
+    jsonObject: function(
```
jannyHou: The solution is cool! 
Implementation details...I think we don't need a new decorator to support the multi contents, @bajtos 's comment is sth we could try https://github.com/strongloop/loopback-next/pull/4141/files#r347767523.

Or provide the `content` object as the 3rd parameter `spec?: Partial<ParameterObject>`

@deepakrkris You may want to check is it a valid spec to have `schema` and `content` both defined, like:
```
 // new addition - json encoding
content: {
    "application/json": {
         schema
     }
},

// the rest
// @jannyHou: Is it valid to have both `schema` and `content` defined?
 schema,
```

jannyHou: And I feel the new signature of `param.query.object()` would be very similar to `@requestBody`, which reminds me to have a chat with @bajtos on https://github.com/strongloop/loopback-next/issues/3493, I will schedule a meeting asap and update our conclusion here. Anyway the spike is good enough to me :) We can determine the signature in the follow-up stories.

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
jannyHou: +1 Let's make sure the default signature `filter: {where: {isComplete: false}}` still works.

deepakrkris: > Ideally, both flavors should be supported:
> 
> * `filter: {where: {isComplete: false}}`
> * `filter: encodeURIComponent(JSON.stringify(...))`
>

@bajtos while using super-test both the above are the same , ie, using `encodeURIComponent(JSON.stringify(` for super test is redundant, while for other http client invocations it is required. sending just `filter: {where: {isComplete: false}}` in url will not work.
I suppose you are asking to support exploded query params like `filter[limit]=1` 

> Instead of changing this existing test, we should add a new test to verify the other flavor. I believe these two flavors are already supported now, in which case it would be great to have tests to verify all different ways how to provide the `filter` argument.

yes both the exploded and complex objects are already supported in loopback, even without the changes in this PR. The suggested changes in this PR is only for API Explorer to support complex objects.

deepakrkris: @bajtos @jannyHou I have now implemented your suggestions to modify existing decorator and also have backward compatibility, ie, support both exploded query params and complex objects in query params.

packages/openapi-v3/src/decorators/parameter.decorator.ts
deepakrkris: @jannyHou  I have modified existing decorator with backward compatibility and also fix API Explorer for complex objects in query params.

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
deepakrkris: @bajtos @jannyHou , to be clear , currently loopback supports both exploded and uri encoded complex objects in query params, even without the changes suggested in this PR. The purpose of this PR is to ONLY fix API Explorer to support complex objects in query params.

deepakrkris: @bajtos @jannyHou , I have tested , for example apps API Explorer  is working fine now for complex objects in query params.

examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts
```diff
@@ -204,6 +204,17 @@ describe('TodoListApplication', () => {
     });
   });
 
+  it('exploded filter conditions work', async () => {
+    const list = await givenTodoListInstance(todoListRepo);
+    await givenTodoInstance(todoRepo, {title: 'todo1', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo2', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo3', todoListId: list.id});
+
+    const response = await client.get('/todos').query('filter[limit]=2');
```
bajtos: This is a bit redundant, because `query({filter: {limit: 2}})` is automatically converted by supertest to `?filter[limit]=2`. Having said that, I think it's a good idea to have tests that are explicitly verifying different encoding styles, as opposed to the current status, where such encodings are tested implicitly, see:

https://github.com/strongloop/loopback-next/blob/303d3ff6ebe751b238ab16e021f6a25f09f4a1be/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts#L110

https://github.com/strongloop/loopback-next/blob/303d3ff6ebe751b238ab16e021f6a25f09f4a1be/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts#L196-L198

Writing such tests is out of scope of this spike, but please do include a task/todo-item in the acceptance criteria of the story that will track the real implementation.

bajtos:
Considering that runtime supports both `deepObject` and `json` encodings (no changes are needed to fix API Explorer experience), and many of our users (hopefully?) don't care that much how object parameters are described in the spec, as long as it's a valid spec that's supported by API Explorer, I am leaning towards removing `deepObject` style from the generating spec and providing `content` only.

This can break existing applications (for example when using clients code-generated by a tool that may not support json encoding), therefore this change needs to be published as semver-major of `@loopback/openapi-v3` and `@loopback/rest`.

Thoughts?

/cc @strongloop/loopback-next

jannyHou:
@bajtos I am glad you verified manually that `schema` and `content` has conflict and found the related doc.

> I am leaning towards removing deepObject style from the generating spec and providing content only.

Sounds good to me.

> This can break existing applications (for example when using clients code-generated by a tool that may not support json encoding), therefore this change needs to be published as semver-major of @loopback/openapi-v3 and @loopback/rest.

Fair enough to have a semver-major release to support querying nested filter.

deepakrkris:
@bajtos @jannyHou  , I have fixed the parameter spec. removed style , explode and schema fields. Now schema passes https://editor.swagger.io/ . But there are additional changes required in loopback to support extracting param schema from `content` field. Tested with API Explorer and browser, works fine.

deepakrkris:
@bajtos I have also removed the `explode` and `style` properties because `https://editor.swagger.io/` complains for them as well (when content field is included in the schema)

deepakrkris:
All the suggestions are implemented and tested and the CI passes, so I am closing this PR. findings of this spike are updated in the spike PR as well as the linked original issue.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -32,7 +32,14 @@ export function coerceParameter(
   data: string | undefined | object,
   spec: ParameterObject,
 ) {
+  if (!spec.schema && spec.content) {
+    const content = spec.content;
+    const jsonSpec = content['application/json'];
+    spec.schema = jsonSpec.schema;
```
bajtos: This is ok in a spike, but please note that the real implementation must not mutate (modify) the `spec` object given to us by the caller.

Issue #3806, opened 2019-09-25
String type in the generated model class from lb4 discover not correct


dhmlau:
Related to https://github.com/strongloop/loopback-next/issues/2953

It seems like some of the items mentioned in https://github.com/strongloop/loopback-next/issues/2953 wasn't fixed completely.

Runs the `lb4 discover` command and the generated code isn't exactly what we're looking for. 

## Steps to reproduce

1. Create a very simple table, e.g. 
```
  Table "public.customer"
  Column  | Type | Collation | Nullable | Default 
----------+------+-----------+----------+---------
 custid   | text |           | not null | 
 custname | text |           |          | 
```

This is my command on postgres: 
```
CREATE TABLE customer (custid text PRIMARY KEY, custname text);
```

2. Scaffold an app and create a datasource that connects to a database

3. Run:
```
npm run build
npm run migrate
```

## Current Behavior

```
@property({
    type: String,
    required: true,
    id: 1,
    postgresql: {"columnName":"custid","dataType":"text","dataLength":null,"dataPrecision":null,"dataScale":null,"nullable":"NO"},
  })
  custid: String;
```

## Expected Behavior

```
@property({
    type: string, <----- problem 1: lower case "string"
    required: true,
    id: 1,
    postgresql: {"columnName":"custid","dataType":"text","dataLength":null,"dataPrecision":null,"dataScale":null,"nullable":"NO"}, <--- problem 2: ideally it is formatted properly
  })
  custid: String;  <---- problem 3: lower case "string" 
```

See details in https://github.com/strongloop/loopback-next/issues/2953#issuecomment-495578669

## Acceptance Criteria
There are more details in https://github.com/strongloop/loopback-next/issues/3711 from @bajtos. 
- [x]  inside `@property`, change the type to be `string` instead of `String` (listed as problem 1 above) --> see #4143
- [x] as the type of the property, change the type to be `string` instead of `String` (listed as problem 3 above) --> see #4143
- [ ] Stretch goal: in the connector specific json inside `@property`, represent it as a TypeScript object --> see #4180

bajtos:
See also #3711 that describes another aspect of this problem, one related to controllers & repositories.

emonddr:
Double-check for other types as well.

jannyHou:
My finding so far:

[...]

PR #4143, opened 2019-11-18
fix(cli): emit correct property definitions for built-in types in discovered models

bajtos:
This pull request fixes model discovery to correctly handle built-in types and convert e.g. `type: String` to `type: 'string'`, see https://github.com/strongloop/loopback-next/issues/3806.

I split the change into several smaller commits:

1. Add snapshots to CLI tests to give us better understanding of what code we are actually scaffolding. This is also needed to have a test that demonstrates the problem!

2. Refactor the current implementation to make the fix easier: extract helper function converting discovered model to template data, extract `findBuiltintType` helper.

These first steps are not changing observable behavior. With the preparations done, we can make the fixes now.

3. Emit correct property definitions for built-in types from `lb4 discover`.

4. Disconnect the datasource after the model was discovered by `lb4 model`.

5. Emit correct property definitions for built-in types from `lb4 model`. 

## Checklist

### DIFF
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -161,6 +161,7 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
         views: true,
       },
     );
+    await this.artifactInfo.dataSource.disconnect();
 
     if (!schemaDef) {
       this.exit(
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -14,7 +14,10 @@ const utils = require('../../lib/utils');
 const chalk = require('chalk');
 const path = require('path');
 
-const {createPropertyTemplateData} = require('./property-definition');
+const {
+  createPropertyTemplateData,
+  findBuiltinType,
+} = require('./property-definition');
 
 const PROMPT_BASE_MODEL_CLASS = 'Please select the model base class';
 const ERROR_NO_MODELS_FOUND = 'Model was not found in';
@@ -466,9 +469,11 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
 
     debug('scaffolding');
 
-    Object.entries(this.artifactInfo.properties).forEach(([k, v]) =>
-      modelDiscoverer.sanitizeProperty(v),
-    );
+    Object.entries(this.artifactInfo.properties).forEach(([k, v]) => {
+      const builtinType = findBuiltinType(v.type);
+      if (builtinType) v.type = builtinType;
+      modelDiscoverer.sanitizeProperty(v);
+    });
 
     // Data for templates
     this.artifactInfo.outFile = utils.getModelFileName(this.artifactInfo.name);
--- a/packages/cli/generators/model/property-definition.js
+++ b/packages/cli/generators/model/property-definition.js
@@ -11,6 +11,7 @@ const BUILTIN_TYPES = [...TS_TYPES, ...NON_TS_TYPES];
 
 module.exports = {
   createPropertyTemplateData,
+  findBuiltinType,
   BUILTIN_TYPES,
 };
 
@@ -70,4 +71,15 @@ function createPropertyTemplateData(val) {
   }
 
   return val;
+}
+
+/**
+ * Check if the type is a built-in type, return the canonical type name in
+ * such case (e.g. convert 'String' to 'string').
+ *
+ * @param {string} typeName Property type name, e.g. 'String' or 'Address'
+ * @returns {string|undefined}  Built-in type name (e.g. 'string') or undefined
+ */
+function findBuiltinType(typeName) {
+  return BUILTIN_TYPES.find(t => t === typeName.toLowerCase());
 }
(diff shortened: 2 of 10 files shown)


packages/cli/test/unit/discover/import-discovered-model.test.js
```diff
@@ -0,0 +1,162 @@
...
+        },
+      },
+    };
+    const modelData = importDiscoveredModel(discoveredModel);
+    expect(modelData.properties)
+      .to.have.property('createdAt')
+      .deepEqual({
+        type: `'date'`,
+        tsType: 'string',
+        // length: null,
```
jannyHou: hmm, why some config properties are commented?

packages/cli/generators/model/property-definition.js
```diff
@@ -71,3 +72,14 @@ function createPropertyTemplateData(val) {
...
+
+/**
+ * Check if the type is a built-in type, return the canonical type name in
+ * such case (e.g. convert 'String' to 'string').
+ *
+ * @param {string} typeName Property type name, e.g. 'String' or 'Address'
+ * @returns {string|undefined}  Built-in type name (e.g. 'string') or undefined
+ */
+function findBuiltinType(typeName) {
+  return BUILTIN_TYPES.find(t => t === typeName.toLowerCase());
```
jannyHou: 👍

packages/cli/test/unit/discover/import-discovered-model.test.js
```diff
@@ -48,9 +48,8 @@ describe('importDiscoveredModel', () => {
     expect(modelData.properties)
       .to.have.property('createdAt')
       .deepEqual({
-        type: 'Date',
-        tsType: 'Date',
-        required: false,
```
nabdelgadir: why is this property removed?

nabdelgadir:
LGTM so far 👍

packages/cli/test/unit/discover/import-discovered-model.test.js
bajtos: I am not sure what to do here. Metadata set to `null` is excluded from the emitted code, I think the intention was to leave out metadata set to default values (`undefined`, `null`, `false`), which looks reasonable to me.

Now I am not sure if we should make this fact explicit in the tests by keeping those commented-out lines around (perhaps adding another comment with an explanation?), or whether it's better to remove these lines completely.

WDYT?

packages/cli/test/unit/discover/import-discovered-model.test.js
bajtos: Good catch!

Model properties are optional by default. Specifying `required: false` is a bit redundant, therefore we are omitting that metadata from the scaffolded code. This behavior is already implemented by `lb4 model` and `lb4 import-lb3-models`, my change is bringing it to `lb4 discover` too, which is good for consistency.

bajtos:
Added one more commit to fix `lb4 model --dataSource {db} --table {name}`. 

The implementation is not as clean as I would like it to be - we have too many different code paths building model template data - but I guess it's good enough for now.

The pull request is ready for final review as far as I am concerned.

packages/cli/generators/model/index.js
```diff
@@ -466,9 +469,11 @@ module.exports = class ModelGenerator extends ArtifactGenerator {
 
     debug('scaffolding');
 
-    Object.entries(this.artifactInfo.properties).forEach(([k, v]) =>
-      modelDiscoverer.sanitizeProperty(v),
-    );
+    Object.entries(this.artifactInfo.properties).forEach(([k, v]) => {
+      const builtinType = findBuiltinType(v.type);
+      if (builtinType) v.type = builtinType;
```
nabdelgadir: Will `v.type` always be either a built in type by default or be found by `findBuildinType`? If it's not, can we put a warning?

packages/cli/test/unit/discover/import-discovered-model.test.js
jannyHou: Thank you for the explanation.
So IIUC the generated property def for `createdAt` (which should be an object) doesn't have `length`, `precision` and `scale` at all, instead of having them as `null` value? 
If that's the case I would go for deleting these commented code to reflect the generated results :)

packages/cli/generators/model/index.js
jannyHou: I think if `type` is not a built-in type then `findBuiltinType(v.type)` will returns undefined.

bajtos: `v.type` can be either a built-in type (e.g. `String` or `string`) or a user-defined type (e.g. `Address`).

In the case of built-in types, we want to convert the type name to all-lower-case (e.g. `string`), that's what these two lines L473-L474 are doing.

In the case of user-defined types, we keep the type name as-is.

packages/cli/test/unit/discover/import-discovered-model.test.js
bajtos: > So IIUC the generated property def for `createdAt` (which should be an object) doesn't have `length`, `precision` and `scale` at all, instead of having them as `null` value? 

Yes, that's what the connector is giving us back from the discovery.

> If that's the case I would go for deleting these commented code to reflect the generated results :) 

+1, will do.

Issue #1998, opened 2018-11-08
The First Scenario: Authenticated orders (a minimal authentication)


nabdelgadir:
## Description
Step 3 from https://github.com/strongloop/loopback-next/issues/1035#issuecomment-429774436.

Rework `UserOrderController` to obtain the customer id from the request (the access token provided by the client) instead of a URL-path parameter.

For example, we can remove `/users/{userId}` prefix from all endpoints to end up with the following API:

- `POST /orders` creates a new order
- `GET /orders` returns all orders of the current user
- `PATCH /orders?where=` to update some of the orders of the current user. I think this endpoint does not make sense in a Shopping app and should be eventually removed.
- `DELETE /orders?where=` to delete some of the orders of the current user. I think this endpoint should be eventually removed, because orders are never deleted, they can be only closed (e.g. as cancelled).

Existing REST API exposed by `UserController` should remain unauthenticated (allowing anonymous access).

The goal of this story is to leverage `@loopback/authentication` to implement this feature and improve `@loopback/authentication` along the way as needed.

Previous step: https://github.com/strongloop/loopback-next/issues/1997
Next step: https://github.com/strongloop/loopback-next/issues/1999

dhmlau:
Originated from https://github.com/strongloop/loopback-next/issues/1035#issuecomment-438387429

## Acceptance Criteria
- [ ] Implement the "verify" api in the authentication provider. i.e. accept access token. and tell you whether the token is valid plus user info
   1. custom sequence action - authenticate.  This will verify the access token and bind the security context in the context. 
   2. well-known binding key for DI for the security context.

After this task is done, we should be able to support these APIs:
`POST /orders` creates a new order
`GET /orders` returns all orders of the current user

See @loopback/authentication as inspiration.

jannyHou:
Hi @raymondfeng @bajtos I plan to work on this story while realize PR https://github.com/strongloop/loopback4-example-shopping/pull/26 already covers the acceptance criteria as [the created custom authentication action](https://github.com/strongloop/loopback4-example-shopping/blob/master/src/providers/custom.authentication.provider.ts)

Are we good to close this task?

> After this task is done, we should be able to support these APIs:
POST /orders creates a new order
GET /orders returns all orders of the current user

And then create another story for ^

bajtos:
At the moment, `UserOrderController` allow clients to impersonate any user, see e.g.

[...]

Issue #4135, opened 2019-11-15
Missing `models/index.ts` after running `lb4 openapi` command


dhmlau:
After running the `lb4 openapi` command, models are created.  However, there is no `index.ts` file in the `models` folder.  If I generate the repository for the model, it will get an error during build:

```

src/repositories/pet.repository.ts:2:33 - error TS2307: Cannot find module '../models'.

2 import {Pet, PetRelations} from '../models';
                                  ~~~~~~~~~~~
```

It's not a big issue because I can create the index.ts, but generating the index.ts when generating the models would be helpful. 

## Steps to reproduce

1. In a LB application, run 
```
lb4 openapi https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore-expanded.yaml
```

2. It creates the Pet and NewPet and Error model. 
3. Create datasource and Repository
4. Run `npm run build`. 
It gives the error I've shown above because of the missing index.ts to export the models. 
UPDATED: it also complained about `PetRelations` not imported in the repository class but it actually doesn't exist. 

### Acceptance Criteria
- [ ] Generate the index.ts model file during `lb4 openapi`
- [ ] the generated code should not have any build error. (about the missing `{ModelName}Relations`)

raymondfeng:
Interesting. Maybe it's a regression? The index files should be updated by calling https://github.com/strongloop/loopback-next/blob/master/packages/cli/lib/artifact-generator.js#L127

dhmlau:
I'm using the latest version of @loopback/cli.  Just tried it again to make sure, it doesn't generate the index.ts.

dhmlau:
After I added the index.ts in `models` folder, I realized that `PetRelations` does not exist but the repository class is referencing it. 

I've put my project in https://github.com/dhmlau/openapi-test.

Issue #4042, opened 2019-11-01
Model strict: false is ignored with Postgre SQL repository


shreking-around:
Hello,

I have a problem with class inheritance and repository using Postgre SQL.

Let's say we have Animal parent class and then Dog and Chicken child classes. I store them in one table (Table Per Hierarchy Inheritance) called animal.

In LB4 App I have Animal base model (entityType is a discriminator): 

```
@model({
    settings: {
        postgresql: {
            table: 'animal',
        },
        idInjection: false,
        strict: false,
    },
})
export class Animal extends Entity {
    @property({
        type: 'string',
        id: true,
        postgresql: {
            dataType: 'uuid',
        },
    })
    id: string;

    @property({
        type: 'string',
        required: true,
    })
    entityType: string;
 
    @property({
        type: 'number',
        required: true,
    })
    weight: number;

    @hasMany(() => Zookeeper)
    zookepers?: Zookeeper[];
 

    constructor(data?: Partial<Animal>) {
        super(data);
    }
}
```
and child class dog

```
@model()
export class Dog extends Animal {
 
    @property({
        type: 'number',
        required: true

    })
    barkingLevel: number;

    constructor(data?: Partial<Dog>) {
        super(data);
    }
}
```
When I query a dog from the repository barkingLevel is ommited on output though the strict is set to false.

This will return animal (not a dog) without barkingLevel .

```
 const dog : Dog = await this.animalRepository.findById(id, {
            include: [
                { relation: 'zookepers' }
            ],
        }) as Dog;

```

Repository:

```
export class AnimalRepository extends DefaultCrudRepository<
    Animal,
    typeof Animal.prototype.id
> {
    //declare has-many repositories for related zookeeper entities
    public readonly zookeepers: HasManyRepositoryFactory<
        Zookeeper,
        typeof Zookeeper.prototype.id
    >;

....
```

Expected result:  Returned animal from repository should include extra properties from dog since the strict setting is set to false.

Current result: Only properties belonging to parent class are returned. 

NOTE: If I do the same thing in Mongo DB, everything works as expected.

Thank you.

## Acceptance criteria

See https://github.com/strongloop/loopback-next/issues/4042#issuecomment-549801710
> At the moment, the data is discarded silently, the framework does not print any warnings about unsupported `strict` value used with a SQL connector. Personally, I'd like the framework to detect this scenario and print a warning (once for each model), so that developers have easier time troubleshooting the unexpected behavior.

[...]

Issue #3957, opened 2019-10-17
How to migrate boot scripts


bajtos:
_This is a follow-up for #3718 and #3922._

Write content for `docs/site/migration/boot-scripts.md`, describe how to migrate LB3 boot scripts in `server/boot/*.js` to LB4 life cycle observers.

## Acceptance criteria

- [ ] Migration guide describing manual steps

jannyHou:
closed since PR merged

PR #4144, opened 2019-11-18
[RFC] feat(docs): dec milestone

dhmlau:
**Do not merge this PR** 

This PR is to propose the Dec milestone.  Once we agree, a github issue will be created based on the content. This PR is not meant to be merged. 

Currently we've committed in *37* points.  It's lower than our team velocity, but it might be more than enough because of the holidays.  

## Checklist

### DIFF
--- a/docs/_milestone_.md
+++ b/docs/_milestone_.md
@@ -1,0 +1,79 @@
+# December Milestone
+
+## Overview
+
+- Complete the epics:
+  - `Add Support for Partitioned Database`
+  - `Inclusion of related models`
+  - `Production deployment`
+- Complete the spikes for the migration guide
+
+## Goals
+
+- [ ] Inclusion of related models [MVP]
+      https://github.com/strongloop/loopback-next/issues/1352
+
+  - [ ] [5]Reject create/update requests when data contains navigational
+        properties (WIP) https://github.com/strongloop/loopback-next/issues/3439
+
+- [ ] Authentication & Authorization
+      https://github.com/strongloop/loopback-next/issues/3902
+
+  - [ ] [5]Contribute OpenAPI spec pieces from extensions
+        https://github.com/strongloop/loopback-next/issues/3854
+  - [ ] [3]Add user profile factory for authentication modules
+        https://github.com/strongloop/loopback-next/issues/3846
+
+- [ ] [Epic] Add Support for Partitioned Database
+      https://github.com/strongloop/loopback-connector-cloudant/issues/219
+
+  - [ ] [3]Optimize query by reading partition key from payload
+        https://github.com/strongloop/loopback-connector-cloudant/issues/217
+
+- [ ] From model definition to REST API with no custom repository/controller
+      classes https://github.com/strongloop/loopback-next/issues/2036
+
+  - [ ] [5]Model API booter & builder(WIP)
+        https://github.com/strongloop/loopback-next/issues/3736
+
+- [ ] Migration guide between LB3 to LB4 [MVP]
+      https://github.com/strongloop/loopback-next/issues/453
+
+  - [ ] [3]Spike: Migration guide from LB3 - Authentication & authorization
+        https://github.com/strongloop/loopback-next/issues/3719
+
+- [ ] Miscellaneous
+  - [ ] [8]Spike: Discover and define models at runtime
+        https://github.com/strongloop/loopback-next/issues/2484
+  - [ ] [5] API Explorer ignores filter options where, fields, and order #2208
+
+## Stretch Goals
+
+- [ ] Bug fixes
+
+  - [ ] Id deleted on update on MongoDB #3267
+  - [ ] TodoList tutorial - GET ​/todos​/{id}​/todo-list gives unexpected
+        response #4147
+
+- [ ] Inclusion of Related Models (Post MVP)
+
+  - [ ] Include related models with a custom scope #3453
+
+- [ ] From model definition to REST API with no custom repository/controller
+      classes https://github.com/strongloop/loopback-next/issues/2036
+
+  - [ ] [3]Add CrudRestApiBuilder to `@loopback/rest-crud`,
+        https://github.com/strongloop/loopback-next/issues/3737
+  - [ ] [3]Example app showing CrudRestApiBuilder
+        https://github.com/strongloop/loopback-next/issues/3738
+
+- [ ] Migration guide between LB3 to LB4 [MVP]
+      https://github.com/strongloop/loopback-next/issues/453
+
+  - [ ] [3]How to migrate datasources from LB3 to LB4 #3946
+  - [ ] [3]How to migrate boot scripts #3957
+  - [ ] [3]How to migrate user-defined model methods #3949
+
+- Micellaneous
+  - [ ] [5]Column is always integer when running `npm run migrate`
+        https://github.com/strongloop/loopback-next/issues/2398
--- a/docs/_milestone_.md
+++ b/docs/_milestone_.md
@@ -8,7 +8,8 @@
   - `Production deployment`
 - Complete the spikes for the migration guide
 - Community Support: focus on triaging and PR reviews in
-  `loopback-connector-mongodb` and/or `loopback-connector-postgresql`
+  `loopback-datasource-juggler` and/or `loopback-connector-postgresql`
+- Time off for Christmas and New Year!
 
 ## Goals
 
--- a/docs/_milestone_.md
+++ b/docs/_milestone_.md
@@ -7,6 +7,8 @@
   - `Inclusion of related models`
   - `Production deployment`
 - Complete the spikes for the migration guide
+- Community Support: focus on triaging and PR reviews in
+  `loopback-connector-mongodb` and/or `loopback-connector-postgresql`
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


docs/_milestone_.md
```diff
@@ -0,0 +1,68 @@
...
+      classes https://github.com/strongloop/loopback-next/issues/2036
+
+  - [ ] [3]Add CrudRestApiBuilder to `@loopback/rest-crud`,
+        https://github.com/strongloop/loopback-next/issues/3737
+  - [ ] [3]Example app showing CrudRestApiBuilder
+        https://github.com/strongloop/loopback-next/issues/3738
+
+- [ ] Authentication & Authorization
+      https://github.com/strongloop/loopback-next/issues/3902
+  - [ ] [5]The First Scenario: Authenticated orders (a minimal authentication)
```
dhmlau: @raymondfeng @bajtos , we might've talked about this before, I couldn't recall what's left for this task to do.

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
+# December Milestone
+
+## Overview
+
+- Complete the epics:
+  - `Add Support for Partitioned Database`
+  - `Inclusion of related models`
+  - `Production deployment`
+
```
bajtos: I'd also like use to complete spike on migrating auth & auth from LB3 to LB4, so that we can start planning work on the migration guide and/or tooling for 2020Q1

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
...
+## Goals
+
+- [ ] Inclusion of related models [MVP]
+      https://github.com/strongloop/loopback-next/issues/1352
+
+  - [ ] [5]Reject create/update requests when data contains navigational
+        properties https://github.com/strongloop/loopback-next/issues/3439
+
```
bajtos: If all goes well, the MVP scope will be done in November 🤞We may want to consider the following post-MVP story because it already has 5 upvotes:

- [NEEDS ESTIMATE] Include related models with a custom scope https://github.com/strongloop/loopback-next/issues/3453

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
...
+  - [ ] [5]Model API booter & builder
+        https://github.com/strongloop/loopback-next/issues/3736
+
+- [ ] Migration guide between LB3 to LB4 [MVP]
+      https://github.com/strongloop/loopback-next/issues/453
+
+  - [ ] [3]Spike: Migration guide from LB3 - Authentication & authorization
+        https://github.com/strongloop/loopback-next/issues/3719
+  - [ ] [3]How to migrate user-defined model methods #3949
```
bajtos: Personally, I would choose the following order of tasks, but I guess the order does not really matter.

- `[3]` How to migrate datasources from LB3 to LB4 https://github.com/strongloop/loopback-next/issues/3946
- `[3]` How to migrate boot scripts https://github.com/strongloop/loopback-next/issues/3957
- `[3]` How to migrate user-defined model methods https://github.com/strongloop/loopback-next/issues/3949
- `[5]` How to migrate remoting hooks https://github.com/strongloop/loopback-next/issues/3950

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [ ] Miscellaneous
+  - [ ] [8]Spike: Discover and define models at runtime
+        https://github.com/strongloop/loopback-next/issues/2484
+  - [ ] [5]Column is always integer when running `npm run migrate`
+        https://github.com/strongloop/loopback-next/issues/2398
+
```
bajtos: Do we have any room for bug fixes and DX improvements? Here are my top picks.

**Bug fixes**

- Id deleted on update on MongoDB https://github.com/strongloop/loopback-next/issues/3267
- TodoList tutorial - GET ​/todos​/{id}​/todo-list gives unexpected response https://github.com/strongloop/loopback-next/issues/4147

**Developer experience**
- `[NEEDS ESTIMATE]` API Explorer ignores filter options where, fields, and order https://github.com/strongloop/loopback-next/issues/2208
_Continue the work started by the spike task in November to keep the ball rolling and don't loose hard-learned knowledge._
- `[NEEDS ESTIMATE]` Expose `isActive()` method in Transaction interface https://github.com/strongloop/loopback-next/issues/3471
- `[NEEDS ESTIMATE]` Missing `model.ts` after running `lb4 openapi` command https://github.com/strongloop/loopback-next/issues/4135
- `[NEEDS ESTIMATE]` Model strict: false is ignored with Postgre SQL repository https://github.com/strongloop/loopback-next/issues/4042
- `[NEEDS ESTIMATE]` Column names in lowercase using lb4 discover method - SQL connector https://github.com/strongloop/loopback-next/issues/3343

docs/_milestone_.md
bajtos: This story was created about a year ago as a kind of a workaround allowing us to show authentication in practice, without any authorization layer (that was not available at that time). I think the proposal still makes sense from the point of DX offered by shopping example REST API, but then I don't know what's the current status of the example app and what changes were implemented since the issue was created & discussed for the last time.

To be honest, I no longer care about this particular story. If the rest of the team thinks the proposed changes are not worth the effort, then I am fine to close the story as being no longer relevant.

dhmlau: @jannyHou @emonddr, what's your thoughts?

emonddr: I agree with @bajtos . We can drop the `authenticated orders` item, and simply ensure whatever remaining authentication and authorization we want in the shopping example is completed. (right now only a few endpoints have authentication or authorization, and we should discuss if we want more...probably when we discuss what other bells and whistles we want to see in shopping cart example)

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
...
+  - `Inclusion of related models`
+  - `Production deployment`
+
+## Goals
+
+- [ ] Inclusion of related models [MVP]
+      https://github.com/strongloop/loopback-next/issues/1352
+
+  - [ ] [5]Reject create/update requests when data contains navigational
+        properties https://github.com/strongloop/loopback-next/issues/3439
```
jannyHou: @dhmlau hopefully story https://github.com/strongloop/loopback-next/issues/3439 will be completed in this month. It's a committed story and Agnes is actively working on it, I can take it over if we need one more week(still within Nov.).

docs/_milestone_.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [ ] Authentication & Authorization
+      https://github.com/strongloop/loopback-next/issues/3902
+
+  - [ ] [5]Contribute OpenAPI spec pieces from extensions
+        https://github.com/strongloop/loopback-next/issues/3854
+  - [ ] [3]Add user profile factory for authentication modules
```
jannyHou: I would say https://github.com/strongloop/loopback-next/issues/3846 is not a high priority.
"support token based authentication in API Explorer in shopping example" from the epic story is much more important, I don't remember if we have a story for it(probably not), I remember Raymond wants to have a default auth dialog for token based authentication, and I agree with him.

docs/_milestone_.md
jannyHou: After Deepak landed his authorization PR, the user/{userId}/orders endpoints are secured with authorization, which means we now compare the `userId` from path and the one decoded from access token to ensure they match.
If we change the UX to `user/orders`(which retrieves the user id from token directly) then we need to create new examples for authorization scenarios...

I didn't consider the UX change in https://github.com/strongloop/loopback-next/issues/1998 when create the authorization PoC(sorry for that...), but since the authorization is already added in the shopping example, I would rather we close https://github.com/strongloop/loopback-next/issues/1998 since it doesn't fit the current design.

dhmlau: Thanks for your input. I'll close #1998 then. thanks.

docs/_milestone_.md
dhmlau: Thanks. The milestone plan here is tentative, will depend on what's left from Nov.

docs/_milestone_.md
dhmlau: I thought #3854 would allow that. no?

docs/_milestone_.md
dhmlau: Per our discussion, unfortunately we need to commit less for December due to the holidays and vacation time.  But this will be high on the list for Jan for sure. 
cc @mamillar93

dhmlau: will move to stretch goals

dhmlau:
I've updated the milestone based on the review comments. We now target for 37 points because we need to account for the vacation time.  Besides issues we might complete before Dec, if everyone agrees, we'll use this as the Dec milestone. Thanks!

docs/_milestone_.md
mamiller93: Thanks friends!

docs/_milestone_.md
```diff
@@ -0,0 +1,84 @@
...
+- Community Support: focus on triaging and PR reviews in
+  `loopback-datasource-juggler` and/or `loopback-connector-postgresql`
+- Time off for Christmas and New Year!
+
+## Goals
+
+- [ ] Inclusion of related models [MVP]
+      https://github.com/strongloop/loopback-next/issues/1352
+
+  - [ ] [5]Reject create/update requests when data contains navigational
```
dhmlau: If this task is done in Nov, please replace this with 
```
Include related models with a custom scope #3453
```

docs/_milestone_.md
jannyHou: @dhmlau #3854 is "allow extension contributes spec", "support token based authentication in API Explorer in shopping example" means our framework provide built-in spec

docs/_milestone_.md
```diff
@@ -0,0 +1,84 @@
...
+
+- [ ] Authentication & Authorization
+      https://github.com/strongloop/loopback-next/issues/3902
+
+  - [ ] [5]Contribute OpenAPI spec pieces from extensions
+        https://github.com/strongloop/loopback-next/issues/3854
+  - [ ] [3]Add user profile factory for authentication modules
+        https://github.com/strongloop/loopback-next/issues/3846
+
+- [ ] Add Support for Partitioned Database
```
jannyHou: this is already done :)

docs/_milestone_.md
dhmlau: yes

docs/_milestone_.md
dhmlau: the blog post?

dhmlau:
Thanks @agnes512 for creating the milestone issue. See https://github.com/strongloop/loopback-next/issues/4236.

PR #4145, opened 2019-11-18
feat(core): improve states and add shutdown hooks for applications

raymondfeng:
Inspired by https://github.com/godaddy/terminus

The PR improves the application lifecycle as follows:

1. Allow graceful shutdown when the application is running inside a managed container, such as Kubernetes Pods. The container sends signals such as `SIGTERM` to notify the application so that it has a chance to close resources before it's terminated after a grace period. See https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-terminating-with-grace.

2. Tidy up the check for application states to leave no surprise to `app.start` and `app.stop` calls if they are invoked out of sequence.

Docs: https://github.com/strongloop/loopback-next/blob/add-shutdown-hooks/docs/site/Life-cycle.md

## Checklist


raymondfeng:
@bajtos PR description updated. I'll add some docs too.

raymondfeng:
Docs updated. See https://github.com/strongloop/loopback-next/blob/add-shutdown-hooks/docs/site/Life-cycle.md.

packages/core/src/__tests__/acceptance/application.shutdown.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+  skipIf(isWindows, it, 'traps registered signals - SIGINT', () => {
+    return testSignal('SIGINT');
+  });
+
+  skipIf(isWindows, it, 'does not trap unregistered signals - SIGHUP', () => {
+    return testSignal('SIGHUP', undefined, ['start\n']);
+  });
+
+  function normalizeStdoutData(output: string[]) {
+    // The received events can be `['start\n', 'stop\nstopped\n']` instead
```
emonddr: What is the difference in these two lists in the comments?

packages/core/src/__tests__/acceptance/application.shutdown.acceptance.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+    // Wait until the child process exits
+    const childExit = new Promise((resolve, reject) => {
+      child.on('exit', (code, sig) => {
+        if (typeof sig === 'string') {
+          // FIXME(rfeng): For some reason, the sig can be null
+          expect(sig).to.eql(expectedSignal);
+        }
+        // The received events can be `['start\n', 'stop\nstopped\n']` instead
+        // of [ 'start\n', 'stop\n', 'stopped\n' ]
```
emonddr: What is the difference in these two lists in the comments?

packages/core/src/__tests__/acceptance/application.shutdown.acceptance.ts
raymondfeng: The array has different number of items.

docs/site/Life-cycle.md
```diff
@@ -14,6 +14,75 @@ to control the transition of states of `Application`.
...
+state can transition as follows by `start` and `stop`:
+
+1. start()
+
+- created -> starting -> started
+- stopped -> starting -> started
+- started -> started (no-op)
+
+The `start` throws an error if it's called with an invalid application state,
+such as `starting` and `stopping`.
```
bajtos: Should we allow `start` as a no-op while the application is in a `starting` state?

docs/site/Life-cycle.md
```diff
@@ -14,6 +14,75 @@ to control the transition of states of `Application`.
...
+
+2. stop()
+
+- started -> stopping -> stopped
+- created -> created (no-op)
+- stopped -> stopped (no-op)
+
+The `stop` throws an error if it's called with an invalid application state,
+such as `starting` and `stopping`.
```
bajtos: Similarly here, I think `stop` should be allowed while `stopping`.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
```diff
@@ -36,6 +36,7 @@ describe('application metadata booter acceptance tests', () => {
     app = new MyApp({
       rest: givenHttpServerConfig(),
     });
+    await app.start();
```
bajtos: I am confused. Why do we need to start the app? I don't see any change in the tests above. Is this need for a change a signal that you are introducing a breaking change, i.e. existing LB4 projects can break after upgrading to a newer version of LB4 runtime?

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -22,6 +22,46 @@ import {
...
+      const stop = app.stop();
+      expect(app.state).to.equal('stopping');
+      await stop;
+      expect(app.state).to.equal('stopped');
+    });
+
+    it('allows application.stop when it is created', async () => {
+      const app = new Application();
+      await app.stop(); // no-op
+      expect(app.state).to.equal('created');
```
bajtos: I find this confusing. How comes that after calling `app.stop`, the app can be either in `created` or `stopped` state? I would expect it to be always in `stopped` state. By allowing two different states, we are pushing more complexity on our users.

Maybe this is a hint that instead of having two states `created` and `stopped`, we should have a single state only, e.g. `not-running`?

packages/core/src/application.ts
```diff
@@ -174,20 +201,44 @@ export class Application extends Context implements LifeCycleObserver {
...
+
   /**
-   * Start the application, and all of its registered observers.
+   * Start the application, and all of its registered observers. The application
+   * state is checked to ensure the integrity of `start`.
+   *
+   * If the application is already started, no operation is performed.
    */
   public async start(): Promise<void> {
+    this.assertValidStates('start', 'created', 'stopped', 'started');
```
bajtos: Should we define named constants for these magic strings? Preferably as a string enum, see https://www.typescriptlang.org/docs/handbook/enums.html#string-enums.

packages/core/src/application.ts
```diff
@@ -315,12 +366,50 @@ export class Application extends Context implements LifeCycleObserver {
...
+   * Set up signals that are captured to shutdown the application
+   * @param signals - An array of signals to be trapped
+   * @param gracePeriod - A grace period in ms before forced exit
+   */
+  protected setupShutdown(signals: NodeJS.Signals[], gracePeriod?: number) {
+    const cleanup = async (signal: string) => {
+      const kill = () => {
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        signals.forEach(sig => process.removeListener(sig, cleanup));
+        process.kill(process.pid, signal);
```
bajtos: IIUC, this code is trying to kill the process by sending it the same signal that we have already intercepted. Is this the right usage of `process.on` feature? I mean aren't we kind of creating an infinite loop, where the signal handler send the same signal again?

@sam-github would you mind lending us our expertise? We are trying to intercept signals like KILL so that we can perform application-specific cleanup before exiting the process.

docs/site/Life-cycle.md
raymondfeng: IMO, no. Consider the following code:

```ts
await app.start(); // The promise should be not resolved until the application is started
// We assume the app is now in `started` state
```

The other option is to have `app.start()` wait for `started` event while it's starting.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
raymondfeng: We can revert this.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
raymondfeng: Good question. We may have a few choices:

1. Do not allow `stop`
2. Transition from `created` to `stopped`
3. Merge the state

packages/core/src/application.ts
raymondfeng: I'm considering so. On the other hand, do we allow other states, such as `booting` and `booted`?

packages/core/src/application.ts
raymondfeng: No, it's intentional. The first `kill` is trapped to run `stop` as the `_isShuttingDown` flag is `false`. The 2nd `kill` proceeds with routine Node.js default logic to shut down the process.

packages/core/src/__tests__/acceptance/application.shutdown.acceptance.ts
emonddr: > The array has different number of items.

array 1: ['start\n', 'stop\nstopped\n']        has two items: 'start\n' and 'stop\nstopped'
array2:   [ 'start\n', 'stop\n', 'stopped\n' ] has three items: 'start'\n, 'stop\n', 'stopped\n'

why does the second item in the first array contain 2 items separated by \n ?  Is this a typo?

raymondfeng: No, that's how Node.js 8.x writes data to the stream - it ends up with one `data` event instead of two (in Node 10.x and above).

docs/site/Life-cycle.md
bajtos: > The other option is to have `app.start()` wait for `started` event while it's starting.

That's what I would expect from `app.start()` method. As a developer calling `app.start`, I want to ensure the app is running. I don't necessarily care if it's stopped or started, I want to get it to the started state.


In my experience, APIs that offer [idempotence](https://en.wikipedia.org/wiki/Idempotence) are easier to use.

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
bajtos: Originally, I wanted to propose to merge the two states and use `stopped` instead.

However, now that you have introduced `booting` & `booted` states, this is becoming even more confusing. Can we call `app.start()` before the application was booted? Or does it require the app to be in the state `booted`? Can we "un-boot" the application, i.e. go from `booted` back to `created`? I think that's not possible and not feasible to implement either.

So IIUC, we have the following states with the following allowed transitions:

1. `created` - can call `app.boot()` to go to `booting`
2. `booting` - no transition allowed, boot in progress, will go to `booted`
3. `booted` - app is booted but not started yet. Call `app.start()` to go to `starting` state
4. `starting` - no transition allowed, will go to `started`
5. `started` - can call `app.stop()` to go to `stopping` state
6. `stopping` - no transition allowed, will go to `stopped`
7. `stopped` - can call `app.start()` to go to `starting` state

If the list above is correct, then I am proposing to remove `stopped` state and move the app to `booted` state after it has stopped.

Thoughts?

/cc @strongloop/loopback-next

packages/core/src/__tests__/unit/application-lifecycle.unit.ts
```diff
@@ -22,6 +22,71 @@ import {
...
+      const start = app.start();
+      expect(app.state).to.equal('starting');
+      await start;
+      expect(app.state).to.equal('started');
+      const stop = app.stop();
+      expect(app.state).to.equal('stopping');
+      await stop;
+      expect(app.state).to.equal('stopped');
+    });
```
bajtos: This test ignores `booting` and `booted` states. Please verify that boot + start work together as intended & expected.

raymondfeng:
@bajtos PTAL.

raymondfeng:
application states

PR #4146, opened 2019-11-18
feat(http-server): handle http keepalive connections upon application.stop()

raymondfeng:
Inspired by https://github.com/godaddy/terminus and https://github.com/hunterloftis/stoppable.

## Checklist


packages/testlab/src/request.ts
```diff
@@ -3,29 +3,39 @@
...
 
 /**
  * Async wrapper for making HTTP GET requests
  * @param urlString
  */
-export function httpGetAsync(urlString: string): Promise<IncomingMessage> {
+export function httpGetAsync(
+  urlString: string,
+  agent?: http.Agent,
+): Promise<IncomingMessage> {
```
bajtos: These testlab changes would be best to make in a standalone commit, so that we get a meaningful changelog entry.

packages/http-server/src/__tests__/integration/http-server.integration.ts
```diff
@@ -49,6 +52,42 @@ describe('HttpServer (integration)', () => {
...
+  it('stops server with grace period and inflight request', async () => {
+    const serverOptions = givenHttpServerConfig() as HttpServerOptions;
+    serverOptions.gracePeriodForClose = 1000;
+    server = new HttpServer(createRequestHandler(100), serverOptions);
+    await server.start();
+    const agent = new Agent({keepAlive: true});
+    // Send a request with keep-alive
+    const req = httpGetAsync(server.url, agent);
+    // Sleep about 10ms so that the request is accepted by the server
+    await sleep(10);
```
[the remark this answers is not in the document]
bajtos: I think I have a bug in the dummy request handler above. Here is a version that should be more correct:

```js
function dummyRequestHandlerWithDelay(
  req: IncomingMessage,
  res: ServerResponse,
): void {
  requestAccepted.resolve();
  resumeHandler
    .then(() => res.end());
}
```

In hindsight, the variable name `resumeHandler` is a bit misleading, something like `shouldResumeHandler` would probably work better.

packages/http-server/src/__tests__/integration/http-server.integration.ts
```diff
@@ -12,10 +12,13 @@ import {
...
 } from '@loopback/testlab';
 import * as fs from 'fs';
-import {IncomingMessage, Server, ServerResponse} from 'http';
+import {Agent, IncomingMessage, Server, ServerResponse} from 'http';
 import * as os from 'os';
 import * as path from 'path';
+import {promisify} from 'util';
 import {HttpOptions, HttpServer, HttpsOptions} from '../../';
+import {HttpServerOptions} from '../../http-server';
+const sleep = promisify(setTimeout);
```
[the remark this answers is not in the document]
raymondfeng: See https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_args

packages/testlab/src/request.ts
raymondfeng: The change to `testlab` is already in its own commit.

packages/http-server/src/__tests__/integration/http-server.integration.ts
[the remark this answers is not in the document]
raymondfeng: I found a simpler solution:

```ts
// Wait until the request is accepted by the server
await pEvent(server.server, 'request');
```

packages/http-server/src/__tests__/integration/http-server.integration.ts
```diff
@@ -264,6 +302,17 @@ describe('HttpServer (integration)', () => {
...
   }
 
+  function createRequestHandler(delay: number) {
+    return function dummyRequestHandlerWithDelay(
+      req: IncomingMessage,
+      res: ServerResponse,
+    ): void {
+      setTimeout(() => {
+        res.end();
+      }, delay);
```
bajtos: Hard-coded timeouts are brittle and can lead to intermittent CI failures. Can you find a more robust solution please? (See https://github.com/strongloop/loopback-next/pull/4146#discussion_r347771308).

I like the solution you found for waiting until a request is accepted.

Let's please find a solution that will allow the request handler to avoid timeouts too.

packages/cli/generators/app/templates/index.js.ejs
```diff
@@ -8,6 +8,9 @@ if (require.main === module) {
     rest: {
       port: +(process.env.PORT || 3000),
       host: process.env.HOST,
+      // Provide a graceful close for http/https servers
+      // See https://www.npmjs.com/package/stoppable
```
bajtos: Please include the information from code comment:

> Defaults to Infinity (don't force-close). If you want to immediately kill all sockets you can use a grace of 0.

Alternatively, add a link to our docs where this configuration option is described (_`rest` options_ section in `Server.md`).

docs/site/Server.md
```diff
@@ -290,21 +290,22 @@ for more details.
...
+| Property            | Type                      | Purpose                                                                                                                                |
+| ------------------- | ------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| host                | string                    | Specify the hostname or ip address on which the RestServer will listen for traffic.                                                    |
+| port                | number                    | Specify the port on which the RestServer listens for traffic.                                                                          |
+| protocol            | string (http/https)       | Specify the protocol on which the RestServer listens for traffic.                                                                      |
+| gracePeriodForClose | number                    | Specify the grace period in milliseconds to allow the RestServer to finish processing in-flight requests before closing the connection |
```
bajtos: Please capture this information from code comments too:

> Defaults to Infinity (don't force-close). If you want to immediately kill all sockets you can use a grace of 0.

and 
> See https://www.npmjs.com/package/stoppable

I am also not sure if the current description conveys the important information that `gracePeriodForClose` applies only to server shutdown?

raymondfeng:

Issue #3439, opened 2019-07-26
Reject create/update requests when data contains navigational properties


bajtos:
## Suggestion

When the client attempts to create or update a model and related models in a single call, then we should reject such requests with a helpful error message explaining the problem.

## Acceptance criteria

Step 1:

- [ ] Introduce a new protected method `DefaultCrudRepository.fromEntity`.
- [ ] Update repository methods to call `fromEntity` whenever we need to convert LB4 Entity into data to be passed to juggler's PersistedModel.
- [ ] This new method is an extension point for app & extension developers, it complements already existing `toEntity` method and provides functionality similar to LB3 Operation Hook `persist`.

Step 2:

- [ ] Modify `fromEntity` to detect navigational properties in data and throw a helpful error.

See #3387 and https://github.com/strongloop/loopback-next/commit/5beb7b93a3d1ce1077538bed39abfa31c309eba0 in particular.

**IMPORTANT** In the spike, `fromEntity` is accepting options allowing the caller to decide how to treat navigational properties (`relations?: true | false | 'throw'`). Based on the discussion we had, this is not what we want to implement. Instead, we want `fromEntity` to ALWAYS REJECT navigational properties.

jannyHou:
estimation meeting discussion:
- `toEntity` converts a JSON object(from the database) to an entity class instance 
- maybe we should think of a better name for the function `fromEntity`, it should do the reverse vs `toEntity`, then we should put the rejection logic outside the function.
- some proposal: name it as `ensurePersistable`, the function makes sure the data doesn't contain non-persistent data.

agnes512:
closing as done

Issue #2095, opened 2018-11-29
Question: Best practice for custom attributes


TomerSalton:
##  Description / Steps to reproduce / Feature proposal
In my application, I have models with properties that need to be encrypted before saved in the database.
I'm trying to decide what's the best practice to implement that use-case.

## Current implementation
What I'm currently doing is set an attribute in the property `itself:
`encrypted: true`

In my repository, I override the relevant methods. e.g. create:

```
async create(entity: DataObject<MyModel>, options?: AnyObject): Promise<MyModel> {
  for (let propertyName in this.entityClass.definition.properties) {
    if (this.entityClass.definition.properties[propertyName]['encrypted']) {
      entity[propertyName] = encrypted(entity, propertyName)    
    }
  }
  return (await super.create(entity, options));
}
```

There are couple of problems with this implementation.
- This method is not generic, as if tomorrow I would like to implement another logic?
- It requires to override a lot of methods in the repository.
- It requires editing per repository meaning that the CLI only covers small part of the creation process.

Is there any better implementation for this issue?

TomerSalton:
Hey @bajtos,
Could you help me with this question?

bajtos:
Loosely related: https://github.com/strongloop/loopback-next/issues/1919

I think this feature would be much easier to implement if we had load/persist hooks in LB4. The "load" hook could be implemented by overriding `toEntity` method in `DefaultCrudRepository`, see https://github.com/strongloop/loopback-next/blob/377fa084c3a59447176cf59a86c8159192b9a097/packages/repository/src/repositories/legacy-juggler-bridge.ts#L391-L393

I think we should add `fromEntity` method to `DefaultCrudRepository` and use it in all places that are passing `entity` instance as instance data.

Hopefully, such hooks will address your concern _It requires to override a lot of methods in the repository._

As for the third concern, _It requires editing per repository meaning that the CLI only covers a small part of the creation process_:

1. Create your own Repository base class.
2. Inherit from `DefaultCrudRepository` and override methods as needed.
3. Change your per-model repositories to inherit from your new base Repository class instead of DefaultCrudRepository. 

I am not sure if our CLI allows developers to pick a custom Repository class as the base. If it does not, then I think we should enhance it to do so.

> This method is not generic, as if tomorrow I would like to implement another logic?

I am afraid I don't understand, could you please explain in more details?

TomerSalton:
First of all, I agree with you.
[...]

PR #4148, opened 2019-11-19
feat(repository): rejects create/update operations for navigational  properties

agnes512:
implements https://github.com/strongloop/loopback-next/issues/3439.

### **_please review it with "Hide whitespace changes"_**

Add two new protected methods:
`DefaultCrudRepository.ensurePersistable` and `DefaultCrudRepository.entityToData`.

This new function checks `create*(), update*(), save(), delete*(), and replaceById()` methods. Throws when the target data ( the one we are going to create, update) contains navigational properties: 
```
target data: {id: 1, name: 'cusotmer', orders: [{id: 1...}]}
```

- in the `ensurePersistable`, the `//FIXME` from @bajtos says mongoDB would break `replaceById` with ObjectID issue if we apply `ensurePersistable` to `replaceIdBy`. 
By debugging, I think it's caused by replacing the **id project** (`ObjectID` type) to a **string type id**. ~So in my proposal, I remove the id property from the target data for `replaceById` method in `ensurePersistable` since we don't need it anyway~.
I decide to open a follow up issue to handle the issue cause it relates to `dao.js` and `mongo.js` files, which is not a easy fix for me according to my knowledge of the code base :no_mouth:

- tests for create/update with nav properties are added to 
[`legacy-juggler-bridge.unit.ts`](https://github.com/strongloop/loopback-next/pull/4148/files?utf8=%E2%9C%93&diff=unified&w=1#diff-73f3835db4f07793d68c31ec97c55d19R507) (for basic functionality) and
 [`has-many.relation.acceptance.ts `](https://github.com/strongloop/loopback-next/pull/4148/files#diff-11ce87c51444cd7183f671fc75716321R177) ( against real DBs).
- tests for replacement(`save`, `replaceById`) are added to 
[`has-many-inclusion-resolver.relation.acceptance.ts`](https://github.com/strongloop/loopback-next/pull/4148/files#diff-cfaa21cba5e38026c57f793615b06846R190) to check if the new instance can be found with inclusion ( these tests are skip for **Cloudant** as it needs `_rev` token).

## Checklist

### Followups 
Commented the `objectId` issue in #3456 
Simplify fixtures and tests #4249
Enable replaceById + inclusion tests run on Cloudant https://github.com/strongloop/loopback-next/issues/4251

### DIFF
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
@@ -187,9 +187,142 @@ export function hasManyInclusionResolverAcceptance(
       expect(toJSON(result)).to.deepEqual(toJSON(expected));
     });
 
-    it('throws when navigational properties are present when updating model instance', async () => {
-      const created = await customerRepo.create({name: 'customer'});
-      const customerId = created.id;
+    skipIf(
+      features.hasRevisionToken,
+      it,
+      'returns inclusions after running save() operation',
+      async () => {
+        // this shows save() works well with func ensurePersistable and ObjectId
+        // the test skips for Cloudant as it needs the _rev property for replacement.
+        // see replace-by-id.suite.ts
+        const thor = await customerRepo.create({name: 'Thor'});
+        const odin = await customerRepo.create({name: 'Odin'});
+
+        const thorOrder = await orderRepo.create({
+          customerId: thor.id,
+          description: 'Pizza',
+        });
+
+        const pizza = await orderRepo.findById(thorOrder.id);
+        pizza.customerId = odin.id;
+
+        await orderRepo.save(pizza);
+        const odinPizza = await orderRepo.findById(thorOrder.id);
+
+        const result = await customerRepo.findById(odin.id, {
+          include: [{relation: 'orders'}],
+        });
+        const expected = {
+          ...odin,
+          parentId: features.emptyValue,
+          orders: [
+            {
+              ...odinPizza,
+              isShipped: features.emptyValue,
+              // eslint-disable-next-line @typescript-eslint/camelcase
+              shipment_id: features.emptyValue,
+            },
+          ],
+        };
+        expect(toJSON(result)).to.containEql(toJSON(expected));
+      },
+    );
+
+    skipIf(
+      features.hasRevisionToken,
+      it,
+      'returns inclusions after running replaceById() operation',
+      async () => {
+        // this shows replaceById() works well with func ensurePersistable and ObjectId
+        // the test skips for Cloudant as it needs the _rev property for replacement.
+        // see replace-by-id.suite.ts
+        const thor = await customerRepo.create({name: 'Thor'});
+        const odin = await customerRepo.create({name: 'Odin'});
+
+        const thorOrder = await orderRepo.create({
+          customerId: thor.id,
+          description: 'Pizza',
+        });
+
+        const pizza = await orderRepo.findById(thorOrder.id.toString());
+        pizza.customerId = odin.id;
+        // FIXME: [mongo] if pizza obj is converted to JSON obj, it would get an error
+        // because it tries to convert ObjectId to string type.
+        // should test with JSON obj once it's fixed.
+
+        await orderRepo.replaceById(pizza.id, pizza);
+        const odinPizza = await orderRepo.findById(thorOrder.id);
+
+        const result = await customerRepo.find({
+          include: [{relation: 'orders'}],
+        });
+        const expected = [
+          {
+            ...thor,
+            parentId: features.emptyValue,
+          },
+          {
+            ...odin,
+            parentId: features.emptyValue,
+            orders: [
+              {
+                ...odinPizza,
+                isShipped: features.emptyValue,
+                // eslint-disable-next-line @typescript-eslint/camelcase
+                shipment_id: features.emptyValue,
+              },
+            ],
+          },
+        ];
+        expect(toJSON(result)).to.deepEqual(toJSON(expected));
+      },
+    );
+
+    it('returns inclusions after running updateById() operation', async () => {
+      const thor = await customerRepo.create({name: 'Thor'});
+      const odin = await customerRepo.create({name: 'Odin'});
+
+      const thorOrder = await orderRepo.create({
+        customerId: thor.id,
+        description: 'Pizza',
+      });
+
+      const pizza = await orderRepo.findById(thorOrder.id.toString());
+      pizza.customerId = odin.id;
+      const reheatedPizza = toJSON(pizza);
+
+      await orderRepo.updateById(pizza.id, reheatedPizza);
+      const odinPizza = await orderRepo.findById(thorOrder.id);
+
+      const result = await customerRepo.find({
+        include: [{relation: 'orders'}],
+      });
+      const expected = [
+        {
+          ...thor,
+          parentId: features.emptyValue,
+        },
+        {
+          ...odin,
+          parentId: features.emptyValue,
+          orders: [
+            {
+              ...odinPizza,
+              isShipped: features.emptyValue,
+              // eslint-disable-next-line @typescript-eslint/camelcase
+              shipment_id: features.emptyValue,
+            },
+          ],
+        },
+      ];
+      expect(toJSON(result)).to.deepEqual(toJSON(expected));
+    });
+
+    it('throws when navigational properties are present when updating model instance with save()', async () => {
+      // save() calls replaceById so the result will be the same for replaceById
+      const customer = await customerRepo.create({name: 'customer'});
+      const anotherCus = await customerRepo.create({name: 'another customer'});
+      const customerId = customer.id;
 
       await orderRepo.create({
         description: 'pizza',
@@ -201,11 +334,19 @@ export function hasManyInclusionResolverAcceptance(
       });
       expect(found.orders).to.have.lengthOf(1);
 
+      const wrongOrder = await orderRepo.create({
+        description: 'wrong order',
+        customerId: anotherCus.id,
+      });
+
       found.name = 'updated name';
+      found.orders.push(wrongOrder);
+
       await expect(customerRepo.save(found)).to.be.rejectedWith(
-        'The `Customer` instance is not valid. Details: `orders` is not defined in the model (value: undefined).',
+        /Navigational properties are not allowed.*"orders"/,
       );
     });
+
     // scope for inclusion is not supported yet
     it('throws error if the inclusion query contains a non-empty scope', async () => {
       const customer = await customerRepo.create({name: 'customer'});
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -339,14 +339,18 @@ export class DefaultCrudRepository<
   }
 
   async create(entity: DataObject<T>, options?: Options): Promise<T> {
-    const model = await ensurePromise(this.modelClass.create(entity, options));
+    // perform persist hook
+    const data = await this.entityToData(entity, options);
+    const model = await ensurePromise(this.modelClass.create(data, options));
     return this.toEntity(model);
   }
 
   async createAll(entities: DataObject<T>[], options?: Options): Promise<T[]> {
-    const models = await ensurePromise(
-      this.modelClass.create(entities, options),
+    // perform persist hook
+    const data = await Promise.all(
+      entities.map(e => this.entityToData(e, options)),
     );
+    const models = await ensurePromise(this.modelClass.create(data, options));
     return this.toEntities(models);
   }
 
@@ -415,7 +419,9 @@ export class DefaultCrudRepository<
     return this.updateById(entity.getId(), entity, options);
   }
 
-  delete(entity: T, options?: Options): Promise<void> {
+  async delete(entity: T, options?: Options): Promise<void> {
+    // perform persist hook
+    await this.entityToData(entity, options);
     return this.deleteById(entity.getId(), options);
   }
 
@@ -425,8 +431,9 @@ export class DefaultCrudRepository<
     options?: Options,
   ): Promise<Count> {
     where = where ?? {};
+    const persistedData = await this.entityToData(data, options);
     const result = await ensurePromise(
-      this.modelClass.updateAll(where, data, options),
+      this.modelClass.updateAll(where, persistedData, options),
     );
     return {count: result.count};
   }
@@ -451,7 +458,8 @@ export class DefaultCrudRepository<
     options?: Options,
   ): Promise<void> {
     try {
-      await ensurePromise(this.modelClass.replaceById(id, data, options));
+      const payload = await this.entityToData(data, options);
+      await ensurePromise(this.modelClass.replaceById(id, payload, options));
     } catch (err) {
       if (err.statusCode === 404) {
         throw new EntityNotFoundError(this.entityClass, id);
@@ -526,6 +534,60 @@ export class DefaultCrudRepository<
     options?: Options,
   ): Promise<(T & Relations)[]> {
     return includeRelatedModels<T, Relations>(this, entities, include, options);
+  }
+
+  /**
+   * This function works as a persist hook.
+   * It converts an entity from the CRUD operations' caller
+   * to a persistable data that can will be stored in the
+   * back-end database.
+   *
+   * User can extend `DefaultCrudRepository` then override this
+   * function to execute custom persist hook.
+   * @param entity The entity passed from CRUD operations' caller.
+   * @param options
+   */
+  protected async entityToData<R extends T>(
+    entity: R | DataObject<R>,
+    options = {},
+  ): Promise<legacy.ModelData<legacy.PersistedModel>> {
+    return this.ensurePersistable(entity, options);
+  }
+
+  /** Converts an entity object to a JSON object to check if it contains navigational property.
+   * Throws an error if `entity` contains navigational property.
+   *
+   * @param entity The entity passed from CRUD operations' caller.
+   * @param options
+   */
+  protected ensurePersistable<R extends T>(
+    entity: R | DataObject<R>,
+    options = {},
+  ): legacy.ModelData<legacy.PersistedModel> {
+    // FIXME(bajtos) Ideally, we should call toJSON() to convert R to data object
+    // Unfortunately that breaks replaceById for MongoDB connector, where we
+    // would call replaceId with id *argument* set to ObjectID value but
+    // id *property* set to string value.
+    /*
+    const data: AnyObject =
+      typeof entity.toJSON === 'function' ? entity.toJSON() : {...entity};
+    */
+
+    // proposal solution from agnes: delete the id property in the
+    // target data when runs replaceById
+
+    const data: AnyObject = new this.entityClass(entity);
+
+    const def = this.entityClass.definition;
+    for (const r in def.relations) {
+      const relName = def.relations[r].name;
+      if (relName in data) {
+        throw new Error(
+          `Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}")`,
+        );
+      }
+    }
+    return data;
   }
 
   /**
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
@@ -174,7 +174,7 @@ export function hasManyRelationAcceptance(
       );
     });
 
-    it('does not create an array of the related model', async () => {
+    it('throws when tries to create() an instance with navigational property', async () => {
       await expect(
         customerRepo.create({
           name: 'a customer',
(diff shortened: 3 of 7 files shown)


packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +534,50 @@ export class DefaultCrudRepository<
...
+      typeof entity.toJSON === 'function' ? entity.toJSON() : {...entity};
+    */
+
+    // proposal solution from agnes: delete the id property in the target dtat
+    // when runs replaceById
+
+    const data: AnyObject = new this.entityClass(entity);
+
+    const def = this.entityClass.definition;
+    for (const r in def.relations) {
```
jannyHou: The way how we detect the navigational property is reasonable to me 👍

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
```diff
@@ -187,9 +187,90 @@ export function hasManyInclusionResolverAcceptance(
       expect(toJSON(result)).to.deepEqual(toJSON(expected));
     });
 
-    it('throws when navigational properties are present when updating model instance', async () => {
-      const created = await customerRepo.create({name: 'customer'});
-      const customerId = created.id;
+    describe('checks save() and replaceById()', () => {
+      /* istanbul ignore next */
+      (features.hasRevisionToken ? it.skip : it)(
```
bajtos: Please use `itSkipIf` helper from `@loopback/testlab`.

bajtos:
> Add a new protected method `DefaultCrudRepository.ensurePersistable`.
( I didn't use name `fromEntity` cause I get confused easily with the name while doing the task..)


+1 to find a better name than `fromEntity`.

Originally, I wanted to introduce `fromEntity` as something that's useful beyond validation of navigational properties, it has the potential to serve as a solution for [LB3 hook `persist`](https://loopback.io/doc/en/lb3/Operation-hooks.html#persist).

The idea is to have a protected Repository method that's called whenever a CRUD method needs to convert data supplied by the user to data passed down to juggler. This new method can modify the data (e.g. by returning a new data object), but also perform validations (e.g. reject invalid data by throwing an error).

It makes me wonder, is this something we should aim for as part of this work?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +534,50 @@ export class DefaultCrudRepository<
...
+      const relName = def.relations[r].name;
+      if (relName in data) {
+        throw new Error(
+          `Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}")`,
+        );
+      }
+    }
+    if (options.replacement === true) {
```
jannyHou: Hmm, why do we need to deal with the `idProperty` in `ensurePersistable`? I don't think for replacement people can include the id from input, we should prevent the id field by excluding it from the request body schema defined in controller. 

See [the type used for request body](https://github.com/strongloop/loopback-next/blob/8a7f78dccdb384d19f16d382aaf0ac74be874e84/examples/todo/src/controllers/todo.controller.ts#L102), I think the solution should be changing it to sth like
```ts
@requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(Todo, {title: 'NewTodo', exclude: ['id']}),
        },
      },
    })
    todo: Omit<Todo, 'id'>,
```

packages/repository/src/repositories/legacy-juggler-bridge.ts
jannyHou: Improvement to consider: not the scope of this PR, if the same check applies to all CRUD functions, maybe we should define it as an interceptor?

jannyHou:
@agnes512 The added tests LGTM in general(I will do a line by line review later), left two comments about the design of `ensurePersistable`:

https://github.com/strongloop/loopback-next/pull/4148#discussion_r349634136 feel free to improve it in a separate story or at least a separate PR

https://github.com/strongloop/loopback-next/pull/4148#discussion_r349633521 can be done in a separate story.

jannyHou:
> Originally, I wanted to introduce fromEntity as something that's useful beyond validation of navigational properties, it has the potential to serve as a solution for LB3 hook persist.

@bajtos Could you elaborate more about ^? are you looking for something like an `ensurePersistant` interceptor? my related comment https://github.com/strongloop/loopback-next/pull/4148#discussion_r349634136

packages/repository/src/repositories/legacy-juggler-bridge.ts
agnes512: I just tested out the code with @bajtos 's fix of `ObjectId`: https://github.com/strongloop/loopback-connector-mongodb/pull/553. I think I can remove the `options.replacement` part cause it's fixed 👏 

And yes I agree that `id` should be excluded from `update/replace` methods at **both** controller and repository level. Currently `update/replacement` methods take `id` property and ignore it in the operation. The behavior is **correct** but I think it should throw errors when `id` is included. 

So I am thinking, like Miroslav suggested above, the new function can perform validations such as checking nav properties, and checking if id is included for `update/replace`.

bajtos:
> > Originally, I wanted to introduce fromEntity as something that's useful beyond validation of navigational properties, it has the potential to serve as a solution for LB3 hook persist.
> 
> Could you elaborate more about ^? are you looking for something like an `ensurePersistant` interceptor? my related comment https://github.com/strongloop/loopback-next/pull/4148#discussion_r349634136

The use case I have in mind is storing certain properties as encrypted in the database, see https://github.com/strongloop/loopback-next/issues/2095.

- Inside `toEntity` method, we can decrypt the value fetched from the database and store plain-text value in the Entity property.
- Inside `fromEntity` method, we can encrypt the plain-text value stored in the Entity property and send encrypted value in the "raw data" sent to to underlying DAO/connector methods.

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
```diff
@@ -187,9 +187,92 @@ export function hasManyInclusionResolverAcceptance(
       expect(toJSON(result)).to.deepEqual(toJSON(expected));
     });
 
-    it('throws when navigational properties are present when updating model instance', async () => {
-      const created = await customerRepo.create({name: 'customer'});
-      const customerId = created.id;
+    skipIf(
+      !features.hasRevisionToken,
+      it,
+      'returns inclusions after running save() operation',
```
bajtos: Could you please provide more details on why are these new test cases skipped? IIUC, you are running these tests only for CouchDB/Cloudant and skipping them for all other databases (memory, PostgreSQL, MongoDB). The code in the test looks legit to me, I find it suspicious that it does not work on so many database types.

Ideally, I'd like to find a way how to enable these tests for all databases we support.

If that's not possible or if it requires too much effort, then I would like you to add code comments near `!features.hasRevisionToken` line to explain why each test is skipped.

jannyHou:
@bajtos Thank you for the explanation. I am looking at your original proposal as
```ts
class DefaultCrudRepository<T, ID> /*...*/{
  protected async entityToData(entity: DataObject<T>, options?: Options): DataObject<T> {
    // "persist hook" - no-op by default
    return entity;
  }

  protected async dataToEntity(data: DataObject<T>, options?: Options): DataObject<T> {
    // "load hook" - no-op by default
    return this.toEntity(data);
  }

  async create(entity: DataObject<T>, options?: Options): Promise<T> {
    const data = await this.entityToData(entity, options);
    const model = await ensurePromise(this.modelClass.create(data, options));
    const result = await this.dataToEntity(model);
    return result;
  }

  // etc.
}
```
Do you mean we implement no-op `entityToData` and `dataToEntity` in `DefaultCrudRepository` then add Agnes's navigation property check in generated Repositories by CLI?

```ts
// Check navigation properties
// This check will be added in EACH created Repository that extends `DefaultCRUDRepository`
// as the built-in implementation for `entityToData`
// If developer wants to custom the logic how would they do it? Create a custom
// `BaseCRUDRepository` with their implementations of `entityToData` and `dataToEntity`, then
// make the created Repositories extend the `BaseCRUDRepository`?

protected async entityToData(entity: DataObject<T>, options?: Options): DataObject<T> {
    const data: AnyObject = new this.entityClass(entity);

    const def = this.entityClass.definition;
    for (const r in def.relations) {
      const relName = def.relations[r].name;
      if (relName in data) {
        throw new Error(
          `Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}")`,
        );
      }
    }
}
```

bajtos:
> Do you mean we implement no-op entityToData and dataToEntity in DefaultCrudRepository then add Agnes's navigation property check in generated Repositories by CLI?

I don't think the navigation property check belongs to generated repositories, IMO this check should be executed by `DefaultCrudRepository`.

When I was writing those comments for `entityToData` and `dataToEntity`, I was not aware that these methods can be useful for checking the presence of navigational properties.

I think we should change the default implementation of `entityToData` from no-op to a check rejecting navigational properties. Something along the following lines:

```ts
class DefaultCrudRepository<T, ID> /*...*/{
  // "persist hook"
  protected async entityToData(entity: DataObject<T>, options?: Options): DataObject<T> {
    ensurePersistable(entity, options);
    return entity;
  }

  // ...
}
```

Please note the name `entityToData` may not be the best one, especially considering that we already have `toEntity` method.

jannyHou:
```ts
class DefaultCrudRepository<T, ID> /*...*/{
  // "persist hook"
  protected async entityToData(entity: DataObject<T>, options?: Options): DataObject<T> {
    ensurePersistable(entity, options);
    return entity;
  }

  // ...
}
```

Cool thank you 👍 now I understand your plan

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
agnes512: The test should run on all databases except `Cloudant` cause it needs the `_rev` property. I will add comments to clarify.

jannyHou: @agnes512 thank you for replying in time, I am running these 2 tests and I think I understand why they fail for Cloudant and mongodb, fixing it.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +536,56 @@ export class DefaultCrudRepository<
...
+  protected async entityToData<R extends T>(
+    entity: R | DataObject<R>,
+    options = {},
+  ): Promise<legacy.ModelData<legacy.PersistedModel>> {
+    this.ensurePersistable(entity, options);
+    return entity;
+  }
+
+  /** Converts an entity object to a JSON object to check if it contains navigational property.
+   * Throws an error if `entity` contains navigational property.
```
bajtos: A nitpick:

```ts
/**
 * Converts an entity object to a JSON object to check if it contains navigational property.
 * Throws an error if `entity` contains navigational property.
```

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +536,56 @@ export class DefaultCrudRepository<
     return includeRelatedModels<T, Relations>(this, entities, include, options);
   }
 
+  protected async entityToData<R extends T>(
```
bajtos: I like this new API 👏 

Can you please add a tsdoc comment for our users?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +536,56 @@ export class DefaultCrudRepository<
...
+    entity: R | DataObject<R>,
+    options = {},
+  ): Promise<legacy.ModelData<legacy.PersistedModel>> {
+    this.ensurePersistable(entity, options);
+    return entity;
```
bajtos: `this.ensurePersistable` is returning a new data object. Should we use it instead of returning `entity`?

packages/repository/src/repositories/legacy-juggler-bridge.ts
jannyHou: added :)

packages/repository/src/repositories/legacy-juggler-bridge.ts
jannyHou: 👍 fixed

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
jannyHou: @bajtos The test case here skips Cloudant connector because the [Customer model](https://github.com/strongloop/loopback-next/blob/master/packages/repository-tests/src/crud/relations/fixtures/models/customer.model.ts#L23) in fixtures doesn't have `_rev` property defined. 

`skipIf(features.hasRevisionToken, testcase){//tests}`means when `features.hasRevisionToken` is true(which only applies for the cloudant connector), the test will be skipped.

I am not very familiar with the shared test case, so agreed with Agnes's solution, 
if you want to enable same tests for cloudant, I can add them in a separate PR (within same story).

jannyHou: @bajtos The test fails for mongodb because the id type is `object` not string, therefore fails at the check in https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2621, hmm...any suggestion on how to fix it?

bajtos:
This is tricky, I'll need to find more time to look into the issue in more depth.

My initial feeling is that we are trying to work around a flaw in the way how repository code and the shared test suite is designed. Instead of adding more and more workarounds, I would prefer to find a better design that will not require these hacks.

> in the ensurePersistable, the //FIXME from @bajtos says mongoDB would break replaceById with ObjectID issue if we apply ensurePersistable to replaceIdBy.
> By debugging, I think it's caused by replacing the id project (ObjectID type) to a string type id. So in my proposal, I remove the id property from the target data for replaceById method in ensurePersistable since we don't need it anyway.

Ideally, when we run the test suite against MongoDB, all PK and FK values should be always strings. This is typically achieved by defining the properties as `{type: 'string', mongodb: {dataType: 'ObjectID'}}` (plus `id`, `required` and `generated` flags). Unfortunately, when a PK property is generated, LoopBack ignores the user-provided type and uses the type defined by the connector. We have recently added a flag to disable that behavior, see https://github.com/strongloop/loopback-datasource-juggler/pull/1783. What happens when we enable that flag in `repository-tests` models - will it solve the issues we are observing here?

> The test case here skips Cloudant connector because the Customer model in fixtures doesn't have _rev property defined.

This is similar to the discussion we had in https://github.com/strongloop/loopback-next/pull/3968, see https://github.com/strongloop/loopback-next/pull/3968#discussion_r347239532 IMO, we should find a way how to modify Customer model to include `_rev` property when testing Cloudant.

---

IMO, the current way of setting up test models in `repository-tests` is no longer supporting the needs of different connectors.

- MongoDB connector wants to set `mongodb: {dataType: 'ObjectID'}`.
- Cloudant wants to add `_rev` property
- SQL connectors want to use `number` as the type for primary & foreign keys. NoSQL connectors want to use `string` type.

Ideally, `CrudFeatures` should provide generic extension points allowing connectors to supply additional model/property configuration that will be applied by the shared test suite.

One of the problems I see in the current setup is that models used by relations are hard-coded in TypeScript source files (see e.g. [`packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts`](https://github.com/strongloop/loopback-next/blob/1de6e122da3b27190dc8a08b739f0e484470d524/packages/repository-tests/src/crud/relations/fixtures/models/address.model.ts)). This makes it difficult to create a model that's tailored to the connector being used. See for example how we are using `MixedIdType` and modifying PK property metadata in the tests.

IMO, our life would be much easier if model classes were created dynamically by a factory function, as can be see in the non-relation tests - e.g. here:

https://github.com/strongloop/loopback-next/blob/1de6e122da3b27190dc8a08b739f0e484470d524/packages/repository-tests/src/crud/create-retrieve.suite.ts#L29-L53

Anyhow, I understand such change may require more effort that we want to invest right now. Even if we decide to do it, it belongs to a new pull request, not here.

In that light, let's try to find a way how to patch the current test setup so that we don't have to implement workarounds in the production code, and instead have the test models use the right model settings & property definitions, ones that our users should apply too.

I think that means we need to add `_rev` property to the Customer model in the fixture. I am not happy to use `_rev` when testing SQL databases, but I am ok to accept it as a temporary workaround for the limitations of current test fixture setup. We also need to fix definition of PK/FK properties and/or enable `useDefaultIdType`, so that PK/FK properties are always converted to `string` when testing MongoDB. The last step is to open a follow-up issue to refactor the shared test suite to better support needs of different connectors.

Thoughts?

packages/repository/src/repositories/legacy-juggler-bridge.ts
bajtos: > I just tested out the code with @bajtos 's fix of ObjectId: strongloop/loopback-connector-mongodb#553. I think I can remove the options.replacement part cause it's fixed 👏

Awesome!

> And yes I agree that id should be excluded from update/replace methods at both controller and repository level. Currently, update/replacement methods take id property and ignore it in the operation. The result is correct but I think it should throw errors when id is included.

I believe it was an intentional design to require `id` property to be present in the payload of `replaceById` request. This operation does exactly what is says - replaces the entire model instance in the database with the supplied data. This is important when free-form properties come to play, because it's not possible to remove a property via `patch` calls. The call `patchById(id, {someProp: undefined})` is a no-op, because `undefined` cannot be represented in JSON.

I feel it's also better UX when the `id` is allowed in the model data sent to both `patch` and `replace` methods. Clients often store the model instance in a variable, then make some change to some properties (e.g. via UI) and then want to persist that changes. If the PK property is forbidden in the payload, then the clients must do extra work to remove it.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +536,60 @@ export class DefaultCrudRepository<
...
+   * It converts an entity from the CRUD operations' caller
+   * to a persistable data that can will be stored in the
+   * back-end database.
+   *
+   * User can extend `DefaultCrudRepository` then override this
+   * function to execute custom persist hook.
+   * @param entity The entity passed from CRUD operations' caller.
+   * @param options
+   */
+  protected async entityToData<R extends T>(
```
bajtos: Since `entityToData` is a new public API, we should have tests to describe and verify the intended behavior:
 - `entityToData` is called for create/patch/replace operations (probably for `delete` too)
 -  the value returned by `entityToData` is sent to persisted model, i.e. modifications made by this function are applied

Thoughts?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -528,6 +536,60 @@ export class DefaultCrudRepository<
...
+    return this.ensurePersistable(entity, options);
+  }
+
+  /** Converts an entity object to a JSON object to check if it contains navigational property.
+   * Throws an error if `entity` contains navigational property.
+   *
+   * @param entity The entity passed from CRUD operations' caller.
+   * @param options
+   */
+  protected ensurePersistable<R extends T>(
```
bajtos: Is it a good idea to expose this method as `protected`? We already have `entityToData`, I think that's good enough, and thus `ensurePersistable` can be `private`.

Thoughts?

jannyHou:
@bajtos Thanks for the review and very detailed suggestion!

> We have recently added a flag to disable that behavior, see strongloop/loopback-datasource-juggler#1783. What happens when we enable that flag in repository-tests models - will it solve the issues we are observing here?

I tried `useDefaultIdType` but not working, maybe I didn't update the juggler to include that fix, let me try again, and that's why I used the solution in the last commit as a workaround(while it turns out to be a wrong solution).


> I think that means we need to add _rev property to the Customer model in the fixture.

👍 I updated in the 2nd last commit, and all tests pass EXCEPT the one failure in the mongodb test suite. The changes in the last commit are actually quite broken.

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
```diff
@@ -187,9 +187,92 @@ export function hasManyInclusionResolverAcceptance(
...
+      });
+      const expected = {
+        ...odin,
+        parentId: features.emptyValue,
+        orders: [
+          {
+            ...odinPizza,
+            isShipped: features.emptyValue,
+            // eslint-disable-next-line @typescript-eslint/camelcase
+            shipment_id: features.emptyValue,
```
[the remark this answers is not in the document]
agnes512: I agree. But we are using almost the same group of fixtures for all repository-tests. We need to include`shipment_id`, `parentId`, etc, for MySQL to pass.

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
```diff
@@ -187,9 +187,92 @@ export function hasManyInclusionResolverAcceptance(
...
+      });
+
+      const pizza = await orderRepo.findById(thorOrder.id);
+      pizza.customerId = odin.id;
+
+      await orderRepo.save(pizza);
+      const odinPizza = await orderRepo.findById(thorOrder.id);
+
+      const result = await customerRepo.findById(odin.id, {
+        include: [{relation: 'orders'}],
```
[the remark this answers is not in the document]
agnes512: > I don't understand why would this test fail before the changes made in this pull request. Could you > please verify that it does fail against the current master?

yes I think `replaceById` fails if we set `dataType: mongodb` and pass in a JSON object as the target data. 

> the purpose of this new test and how it relates to rejecting navigational properties in create/patch/replace operations. Can you please explain?

I would like to check the new function would through error and also work for inclusion.

packages/repository-tests/src/crud/relations/acceptance/has-many-inclusion-resolver.relation.acceptance.ts
```diff
@@ -187,9 +187,92 @@ export function hasManyInclusionResolverAcceptance(
...
+        customerId: thor.id,
+        description: 'Pizza',
+      });
+
+      const pizza = await orderRepo.findById(thorOrder.id.toString());
+      pizza.description = 'Reheated pizza';
+      const reheatedPizza = toJSON(pizza);
+
+      // coerce the id for mongodb connector to pass the id equality check
+      // in juggler's replaceById function
```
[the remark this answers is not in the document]
agnes512: removed `const reheatedPizza = toJSON(pizza);`

packages/repository/src/repositories/legacy-juggler-bridge.ts
agnes512: unit test for `entityToData` is added. It is invoked by `create()` method cause it's protected.

CedricBad:
Hi there, I am trying to perform a crud operation (post and patch) without any navigational property, but the error shows off, I can't understand why...

To make it short, I have a simple model :
```typescript
class Bag extends Entity {
@property() id;
@property() name;
@hasMany(() => Stuff)
  stuffs?: Array<Stuff> = [];
}
```

In my repository I do something like:
```typescript
this.updateById(1, {name: 'new bag');
```

And I got this error : Error: Navigational properties are not allowed in model data (model "Bag" property "stuffs");

Am I missing something ?

agnes512:
@CedricBad with the hasMany decorator,
```
@hasMany(() => Stuff)
  stuffs?: Array<Stuff>
```
if a `Bag` instance doesn't have any `Stuff`, `stuffs` would be empty and won't be included in the result if you travers `bagRepository.find({include:[{relation: 'stuffs'}]})`.

Please let me know if you have any questions.

CedricBad:
@agnes512 

I was looking for a way to have my stuffs always included, even if there is none.  That is why I assigned an empty array to the stuffs. But then I have the error mentionned aboved, so I guess there is no way to have a default value assigned to a relationnal item.
Sorry if I'm not clear, I'm giving you my best english :dancer:

PR #4168, opened 2019-11-20
feat(context): make it possible to set source information for interceptions

raymondfeng:
Interceptors can be invoked in the following cases:

1. A route to a controller method
2. A controller to a repository/service with injected proxy
3. A repository/service method invoked explicitly with `invokeMethod/invokeMethodWithInterceptors`

Some interceptors want to check invocationContext.source to decide if its logic should be applied. For example, an http access logger only cares about invocations from the rest layer to the first controller.

This is also useful for metrics, tracing and logging.

This PR sets `source` as follows:
- the ResolutionSession in `ProxySource` for injected proxies. 
- the Rest Route in `RouteSource` for REST

## Checklist

### DIFF
--- a/docs/site/Interceptors.md
+++ b/docs/site/Interceptors.md
@@ -514,6 +514,7 @@ bindings. It extends `Context` with additional properties as follows:
   (for instance methods)
 - `methodName` (`string`): Method name
 - `args` (`InvocationArgs`, i.e., `any[]`): An array of arguments
+- `source`: Source information about the invoker of the invocation
 
 ```ts
 /**
@@ -533,6 +534,7 @@ export class InvocationContext extends Context {
     public readonly target: object,
     public readonly methodName: string,
     public readonly args: InvocationArgs, // any[]
+    public readonly source?: InvocationSource,
   ) {
     super(parent);
   }
@@ -541,6 +543,48 @@ export class InvocationContext extends Context {
 
 It's possible for an interceptor to mutate items in the `args` array to pass in
 transformed input to downstream interceptors and the target method.
+
+### Source for an invocation
+
+The `source` property of `InvocationContext` is defined as `InvocationSource`:
+
+```ts
+/**
+ * An interface to represent the caller of the invocation
+ */
+export interface InvocationSource<T = unknown> {
+  /**
+   * Type of the invoker, such as `proxy` and `route`
+   */
+  readonly type: string;
+  /**
+   * Metadata for the source, such as `ResolutionSession`
+   */
+  readonly value: T;
+}
+```
+
+The `source` describes the caller that invokes a method with interceptors.
+Interceptors can be invoked in the following cases:
+
+1. A route to a controller method
+
+   - The source describes the REST Route
+
+2. A controller to a repository/service with injected proxy
+
+   - The source describes a ResolutionSession that tracks a stack of bindings
+     and injections
+
+3. A controller/repository/service method invoked explicitly with
+   `invokeMethodWithInterceptors()` or `invokeMethod`
+
+   - The source can be set by the caller of `invokeMethodWithInterceptors()` or
+     `invokeMethod`
+
+The implementation of an interceptor can check `source` to decide if its logic
+should apply. For example, a global interceptor that provides caching for REST
+APIs should only run if the source is from a REST Route.
 
 ### Logic around `next`
 
--- a/docs/site/Interceptors.md
+++ b/docs/site/Interceptors.md
@@ -586,6 +586,32 @@ The implementation of an interceptor can check `source` to decide if its logic
 should apply. For example, a global interceptor that provides caching for REST
 APIs should only run if the source is from a REST Route.
 
+A global interceptor can also be tagged with
+`ContextTags.GLOBAL_INTERCEPTOR_SOURCE` using a value of string or string array
+to indicate if it should be applied to source types of invocations. If the tag
+is not present, the interceptor applies to invocations of any source type. For
+example:
+
+```ts
+ctx
+  .bind('globalInterceptors.authInterceptor')
+  .to(authInterceptor)
+  .apply(asGlobalInterceptor('auth'))
+  // Do not apply for `proxy` source type
+  .tag({[ContextTags.GLOBAL_INTERCEPTOR_SOURCE]: 'route'});
+```
+
+The tag can also be declared for the provider class.
+
+```ts
+@globalInterceptor('log', {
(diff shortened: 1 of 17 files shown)


packages/context/src/invocation.ts
```diff
@@ -47,6 +47,8 @@ export class InvocationContext extends Context {
     public readonly target: object,
     public readonly methodName: string,
     public readonly args: InvocationArgs,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    public readonly source?: any,
```
bajtos: Can we avoid `any` please? (See my other comment below.)

packages/context/src/invocation.ts
```diff
@@ -129,6 +131,12 @@ export type InvocationOptions = {
    * Skip invocation of interceptors
    */
   skipInterceptors?: boolean;
+  /**
+   * Information about the source object that makes the invocation. For REST,
+   * it's a `Route`. For injected proxies, it's a `Binding`.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  source?: any;
```
bajtos: Can we avoid `any` please? It turns off compiler checks for any code that's reading `source`.

At minimum, I would prefer to use `unknown` to force consumers of `source` to deal with the fact that the actual source type is not known at compile time.

If we are sure that `Route` and `Binding` are the only possible types, then we can also replace `any` with `Route | Binding`, but I am not sure if that's feasible because here in `@loopback/context` we don't have access to `Route` from `@loopback/rest`.

On the second thought, this looks like a sign that our current design is no longer good enough and we need to improve it. Maybe the source should be always `Binding` and we should find a way how to access REST Route from its Binding?

I think the most important aspect is how are developers accessing `source` going to know what type to expect? For example, when writing a generic interceptor that can be applied to both controller methods and proxied methods, how is the author going to learn about different types they may receive in `source`?

I find the current design, where the type of  `source` is pretty much unknown or can be anything, as difficult to understand.

raymondfeng:
@bajtos I have improved the design with `InvocationSource`. See docs at https://github.com/strongloop/loopback-next/blob/allow-source-information-for-interception/docs/site/Interceptors.md#source-for-an-invocation.

packages/context/src/interception-proxy.ts
```diff
@@ -77,6 +94,7 @@ export class InterceptionHandler<T extends object> implements ProxyHandler<T> {
           target,
           propertyName,
           args,
+          {source: this.session && new ProxySource(this.session)},
```
bajtos: Should we use `??` operator here?

```suggestion
          {source: this.session ?? new ProxySource(this.session)},
```

packages/context/src/interceptor.ts
```diff
@@ -48,14 +48,38 @@ export class InterceptedInvocationContext extends InvocationContext {
...
     return keys;
   }
 
+  /**
+   * Check if the binding for a global interceptor matches the source type
+   * of the invocation
+   * @param binding - Binding
+   */
+  private applicableTo(binding: Readonly<Binding<unknown>>) {
+    const source = this.source && this.source.type;
```
bajtos: Ditto, should we use `??`

docs/site/Interceptors.md
```diff
@@ -542,6 +544,75 @@ export class InvocationContext extends Context {
...
+### Source for an invocation
+
+The `source` property of `InvocationContext` is defined as `InvocationSource`:
+
+```ts
+/**
+ * An interface to represent the caller of the invocation
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface InvocationSource<T = any> {
```
bajtos: Can we use `unknown` instead of `any` please? Or perhaps `object`?

packages/context/src/interception-proxy.ts
raymondfeng: Will switch after #4213 is landed.

raymondfeng:
@bajtos PTAL

packages/context/src/interception-proxy.ts
raymondfeng: Hmm, I see this error:
```
src/interception-proxy.ts:97:52 - error TS2345: Argument of type 'undefined'
is not assignable to parameter of type 'ResolutionSession'.

97           {source: this.session ?? new ProxySource(this.session)},
```

It seems that `tsc` does not use `??` as an assertion.

packages/context/src/interceptor.ts
raymondfeng:
```
src/interceptor.ts:68:35 - error TS2532: Object is possibly 'undefined'.

68     const source = this.source ?? this.source.type;
                                     ~~~~~~~~~~~
```

packages/context/src/interception-proxy.ts
bajtos: > `? this.session ?? new ProxySource(this.session)`

My fault. The code above says "use `this.session` if it is set, otherwise create new ProxySource".  (It's a replacement for `||`, not `&&`).

packages/context/src/interceptor.ts
bajtos: Sorry, I suggested a wrong operator. I am still learning the ropes here...

I think the following is the new style to use?

```suggestion
    const source = this.source?.type;
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

docs/site/Interceptors.md
bajtos: > Can we use `unknown` instead of `any` please? Or perhaps `object`?

I feel this thread is not resolved yet.

See also https://github.com/strongloop/loopback-next/pull/4168#discussion_r349093431.

My main concerns is that we make it easy to introduce `any` value into code written by consumers of this interface. Once the `any` type is introduced, it pretty much disables all type checks going forward. I am ok to allow to use our users `any`, but I want to make that fact obvious in their code.

Consider the following code:

```ts
const foo: Interceptor = async (invocationCtx, next) => {
  const source = invocationCtx.source;
  const bindingPath = source.getBindingPath();
  // etc.
};
```

The code looks legitimate, it will compile (because `source` is `any`), it will even work as long as the interceptor is invoked only for proxies. The problem will manifest when the interceptor is called for a controller/handler route. Notice that `bindingPath` has also type `any`, which is super easy to miss while reading the code.

bajtos: Oh. I see that in the code, you have already changed the default `T` value to `unknown`. 

Please fix the documentation too and then this thread will be resolved 👍

raymondfeng:
@bajtos Doc fixed.

PR #4180, opened 2019-11-21
Use stringifyObject instead of JSON.stringify for connector metadata

### LINKED ISSUE #3806: String type in the generated model class from lb4 discover not correct

bajtos:
Rework the code handling object-like property metadata (typically connector-specific conciguration) to use `stringifyObject` instead of `JSON.parse`. Resolve #3806.

Before this change, a model property discovered from a database could
look as follows:

```ts
  @property({
    type: String,
    postgresql: {"columnName":"title","dataType":"text"},
  })
  title: String;
```

With this commit in place, the property definition is generated
in TypeScript style:

```ts
  @property({
    type: String,
    postgresql: {columnName: 'title', dataType: 'text'},
  })
  title: String;
```

Please note that in a typical case, connector metadata won't fit into a single line. Ideally, we would like to split the object into multiple lines. Unfortunately, that requires more complex changes in the way how we are generating property definitions, which is something I don't have bandwidth for right now. Users can use tools like Prettier and/or `eslint --fix .` to reformat the code as they like.

## Checklist

### DIFF
--- a/packages/cli/lib/model-discoverer.js
+++ b/packages/cli/lib/model-discoverer.js
@@ -1,6 +1,7 @@
 const debug = require('./debug')('model-discoverer');
 const fs = require('fs');
 const path = require('path');
+const {stringifyObject} = require('../lib/utils');
 
 /**
  * Given a datasource and discovery options,
@@ -85,8 +86,11 @@ const sanitizeProperty = function(o) {
     }
 
     // If you are an object or array, stringify so you don't appear as [object [object]
-    if (v === Object(v)) {
-      o[k] = JSON.stringify(o[k]);
+    if (typeof v === 'object' && v !== null) {
+      o[k] = stringifyObject(o[k], {
+        // don't break lines to avoid formatting problems in the model template
+        inlineCharacterLimit: Infinity,
+      });
     }
   });
 
--- a/packages/cli/lib/utils.js
+++ b/packages/cli/lib/utils.js
@@ -565,16 +565,18 @@ exports.dataSourceToJSONFileName = function(dataSourceClass) {
   );
 };
 
+exports.stringifyObject = function(data, options = {}) {
+  return stringifyObject(data, {
+    indent: '  ', // two spaces
+    singleQuotes: true,
+    inlineCharacterLimit: 80,
+    ...options,
+  });
+};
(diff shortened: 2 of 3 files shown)


raymondfeng:
@bajtos `--format` option is available for all `lb4` commands to reformat the code as part of a command.

PR #4201, opened 2019-11-25
Update Authentication Tutorial with UserCredentials 

### LINKED ISSUE #1996: Refactoring: Customer credentials
## Description
Step 1 from https://github.com/strongloop/loopback-next/issues/1035#issuecomment-429774436.

Let's keep things simple for now and use a password as the only credential. (No 2-factor-auth, no OAuth/OpenID/SAML for now.)

In LoopBack 3.x, we are storing storing user's password together with user data in the same model (table), which opens a lot of security vulnerabilities to deal with. For example:

- When returning user data in HTTP response, the password property must be filtered out.
- When searching for users, the password must be excluded from the query.
- It is not possible to do a full replace of user data from a REST client (e.g. "save()") because a) the client usually does not know the password b) password must be changed using a different flow.
- A policy preventing users from reusing the same password is difficult to implement because old password are not preserved.

For LB4, I would like us to explore the design where passwords are stored in their own table and a relation "user has one password" is configured (#1422 is tracking HasOne relation). We can also use "user has many password" and include a flag (a Password model property) to distinguish between the current active password and the passwords used in the past.

The current version of the Shopping Server is already storing the password in the User model. Let's move it out to a dedicated model as part of this preparation work.

Ideally, there should be a documentation and/or a blog-post and/or a reference implementation to make it easier for LB4 users to implement similar functionality in their project.

Next step: https://github.com/strongloop/loopback-next/issues/1997

## Acceptance criteria
_Based on https://github.com/strongloop/loopback-next/issues/1996#issuecomment-500502620_

- [x] in the example-shopping repo, divide the UserRepository into user credential and user profile information - to separate sensitive information from insensitive information as best practices.
- [ ] documentation and/or a blog-post and/or a reference implementation

bajtos:
In https://github.com/strongloop/loopback4-example-shopping/pull/385, I am extracting local user credentials to a new model `UserCredentials`.

This pull requests updates our documentation to be in sync with the new version of our shopping example app.

Resolve #1996 

## Checklist


docs/site/tutorials/authentication/Authentication-Tutorial.md
```diff
@@ -209,6 +207,9 @@ export class ShoppingApplication extends BootMixin(
     this.component(AuthenticationComponent);
 
     // ...
+  }
+  // ...
+}
```
bajtos: Before this change, GitHub's syntax highlighter was confused - notice the text in "Securing and Enpoint..." is rendered in blue:

Screen Shot 2019-11-25 at 09 36 02

Also Prettier was not able to automatically format the code. Once I added closing curly braces, Prettier shortened the `import` line and GitHub started to render the text in the next section as regular text (in black color) again.

docs/site/tutorials/authentication/Authentication-Tutorial.md
```diff
@@ -88,11 +88,11 @@ application, follow these steps:
 
    ```json
    {
-   "id": "1",
-   "email": "user1@example.com",
-   "password": "thel0ngp@55w0rd",
-   "firstName": "User",
-   "lastName": "One"
+     "id": "5dd6acee242760334f6aef65",
```
emonddr: @bajtos , let's avoid passing in an 'id' field with an ObjectID value to make it easy for the user. Mongo db will create the Object ID , and we will show later when `/users/me` is called.

emonddr: Just like the old example showed, it is also possible to pass in ids with values like "1" and "2", and this is easier for the user to do when creating new users.

bajtos:
>  let's avoid passing in an 'id' field with an ObjectID value to make it easy for the user. 

Fixed in 748a4d6. @emonddr LGTY now?

PR #4213, opened 2019-11-26
[RFC] enable TypeScript 3.7 nullish coalescing

raymondfeng:
See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

## Checklist

### DIFF
--- a/acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
+++ b/acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
@@ -9,12 +9,12 @@ const connector = require('loopback-connector-cloudant');
 
 export const CLOUDANT_CONFIG: DataSourceOptions = {
   connector,
-  host: process.env.CLOUDANT_HOST || 'localhost',
-  port: process.env.CLOUDANT_PORT || 5984,
-  username: process.env.CLOUDANT_USER || 'admin',
-  password: process.env.CLOUDANT_PASSWORD || 'pass',
-  database: process.env.CLOUDANT_DATABASE || 'repository_tests',
-  url: process.env.CLOUDANT_URL || 'http://localhost:5984',
+  host: process.env.CLOUDANT_HOST ?? 'localhost',
+  port: process.env.CLOUDANT_PORT ?? 5984,
+  username: process.env.CLOUDANT_USER ?? 'admin',
+  password: process.env.CLOUDANT_PASSWORD ?? 'pass',
+  database: process.env.CLOUDANT_DATABASE ?? 'repository_tests',
+  url: process.env.CLOUDANT_URL ?? 'http://localhost:5984',
 };
 
 export const CLOUDANT_FEATURES: Partial<CrudFeatures> = {
--- a/extensions/metrics/src/observers/metrics.push.observer.ts
+++ b/extensions/metrics/src/observers/metrics.push.observer.ts
@@ -26,7 +26,7 @@ export class MetricsPushObserver implements LifeCycleObserver {
     this.gateway = new Pushgateway(gwConfig.url);
     this.interval = setInterval(() => {
       this.gateway.pushAdd({jobName: 'loopback'}, () => {});
-    }, gwConfig.interval || 5000);
+    }, gwConfig.interval ?? 5000);
   }
 
   stop() {
--- a/packages/repository/src/mixins/repository.mixin.ts
+++ b/packages/repository/src/mixins/repository.mixin.ts
@@ -110,13 +110,17 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
     ): Binding<D> {
       // We have an instance of
       if (dataSource instanceof juggler.DataSource) {
-        const key = `datasources.${name || dataSource.name}`;
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+        name = name || dataSource.name;
+        const key = `datasources.${name}`;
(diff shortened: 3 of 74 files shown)


acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
```diff
@@ -9,12 +9,12 @@ const connector = require('loopback-connector-cloudant');
...
 export const CLOUDANT_CONFIG: DataSourceOptions = {
   connector,
-  host: process.env.CLOUDANT_HOST || 'localhost',
-  port: process.env.CLOUDANT_PORT || 5984,
-  username: process.env.CLOUDANT_USER || 'admin',
-  password: process.env.CLOUDANT_PASSWORD || 'pass',
-  database: process.env.CLOUDANT_DATABASE || 'repository_tests',
-  url: process.env.CLOUDANT_URL || 'http://localhost:5984',
+  host: process.env.CLOUDANT_HOST ?? 'localhost',
+  port: process.env.CLOUDANT_PORT ?? 5984,
```
bajtos: IIUC, with this change in place, we will convert `CLOUDANT_PORT=0` to `port: 0`. Before this change, we would set `port: 5984`.

Do we care about this change?

The same comment applies to other similar places.

extensions/metrics/src/observers/metrics.push.observer.ts
```diff
@@ -26,7 +26,7 @@ export class MetricsPushObserver implements LifeCycleObserver {
     this.gateway = new Pushgateway(gwConfig.url);
     this.interval = setInterval(() => {
       this.gateway.pushAdd({jobName: 'loopback'}, () => {});
-    }, gwConfig.interval || 5000);
+    }, gwConfig.interval ?? 5000);
```
bajtos: This change will allow `interval` to be set to value `0`. What is the expected behavior when the interval is `0`? Should we release this a semver-major?

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -110,13 +110,13 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
     ): Binding<D> {
       // We have an instance of
       if (dataSource instanceof juggler.DataSource) {
-        const key = `datasources.${name || dataSource.name}`;
+        const key = `datasources.${name ?? dataSource.name}`;
```
bajtos: Before this change, an empty name `''` will be converted to `dataSource.name`. With the change in place, we will preserve the empty string and create key value `datasource.`. I think that's not correct. Thoughts?

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -110,13 +110,13 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
...
         return this.bind(key)
           .to(dataSource)
           .tag('datasource');
       } else if (typeof dataSource === 'function') {
         const binding = createBindingFromClass(dataSource, {
-          name: name || dataSource.dataSourceName,
+          name: name ?? dataSource.dataSourceName,
```
bajtos: Ditto.

Before this change, an empty name `''` will be converted to `dataSource.name`. With the change in place, we will preserve the empty string. I think that's not correct. Thoughts?

acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
raymondfeng: I think the new behavior better reflects our intention. `??` is stricter (==null) than `||` (falsy).

bajtos: I understand the difference between `??` and `||`.

What I am trying to say is that port `0` is not a valid value for clients. At the moment, we are replacing this invalid value with the default port 5984, which seems like a good solution to me. With your change in place, we will use the invalid port. To me, that seems like a change for worse.

@strongloop/sq-lb-apex thoughts?

packages/repository/src/mixins/repository.mixin.ts
bajtos: AFAICT, this is not resolved. Empty name `''` will be passed to `createBindingFromClass`. Does that helper correctly handle empty string in `name` option? Shouldn't we use `dataSource.dataSourceName` instead?

acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
raymondfeng: IMO, the purpose of using `||` or `??` is not to exclude invalid values. For example, `0`, `1`, `2` are invalid values set explicitly by the client but we cannot tell. It's better to expose such errors to the client instead of hiding them.

The intention here should be:

1. If the env var is set (!=null), we use it as is.
2. Otherwise use 5984.

extensions/metrics/src/observers/metrics.push.observer.ts
raymondfeng: See https://nodejs.org/api/timers.html#timers_setinterval_callback_delay_args
> When delay is larger than 2147483647 or less than 1, the delay will be set to 1. Non-integer delays are truncated to an integer.

acceptance/repository-cloudant/src/__tests__/cloudant.datasource.ts
bajtos: Fair enough 👍

PR #4221, opened 2019-11-28
feat(eslint-config): enable no-extra-non-null-assertion

bajtos:
Cross-posting from https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-extra-non-null-assertion.md:

Examples of **incorrect** code for this rule:

 ```ts
 const foo: { bar: number } | null = null;
 const bar = foo!!!.bar;
 ```

 ```ts
 function foo(bar: number | undefined) {
   const bar: number = bar!!!;
 }
 ```

 Examples of **correct** code for this rule:

 ```ts
 const foo: { bar: number } | null = null;
 const bar = foo!.bar;
 ```

 ```ts
 function foo(bar: number | undefined) {
   const bar: number = bar!;
 }
 ```

## Checklist

### DIFF
--- a/packages/cli/generators/example/index.js
+++ b/packages/cli/generators/example/index.js
@@ -104,7 +104,7 @@ module.exports = class extends BaseGenerator {
     const cwd = process.cwd();
     const absOutDir = await downloadAndExtractExample(this.exampleName, cwd);
     this.outDir = path.relative(cwd, absOutDir);
-    fs.rename(
+    return fs.rename(
       `${this.outDir}/tsconfig.build.json`,
       `${this.outDir}/tsconfig.json`,
     );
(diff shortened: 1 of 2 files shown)


raymondfeng:
@bajtos Do you see any code in our repo that violates the rule? I don't mind add this rule. But if it's rarely used, we probably won't take the burden to enforce it.

bajtos:
> Do you see any code in our repo that violates the rule? I don't mind add this rule. But if it's rarely used, we probably won't take the burden to enforce it.

I don't see any such code now. I consider this rule as an prevention, to avoid accidental introduction of such code later.

Personally, I prefer this check to be automated by eslint, instead of having to point violations in pull requests.

PR #4226, opened 2019-11-28
RFC: enable no-untyped-public-signature

bajtos:
eslint-typescript recently introduced a new rule [no-untyped-public-signature](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-untyped-public-signature.md).

I'd like to discuss whether we want to enable it. We have more than 200 violations of this rule in our current code base and unfortunately there is not automatic fixer available. Before I go ahead and manually fix all violations, I'd like to get consensus that we want to enable the rule.

Why would we want to enable it? In my experience, code with missing type information is difficult to read. When LB users are trying to learn about our public API from our code, they have difficult times figuring out what are the methods and functions returning when the return type is not explicitly described.

## Checklist

### DIFF
--- a/packages/eslint-config/eslintrc.js
+++ b/packages/eslint-config/eslintrc.js
@@ -150,6 +150,7 @@ module.exports = {
     '@typescript-eslint/no-misused-promises': 'error',
 
     '@typescript-eslint/prefer-optional-chain': 'error',
+    '@typescript-eslint/no-untyped-public-signature': 'error',
   },
 
   overrides: [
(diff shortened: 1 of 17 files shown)


agnes512:
+1 to enable the new rule. I noticed that not all of functions have API docs. The type would be helpful to learn what a function actually does.

raymondfeng:
I'm neutral on this:

1. Adding explicit return type for public methods make the code easier to understand.
2. TypeScript can infer the return type - I assume tsdocs use the inferred type.
3. The method signature is a bit longer and verbose.

jannyHou:
I also agree on enabling this rule to ensure the consistency within a function's callers.

bajtos:
I have revisited my priorities and feel this pull request requires more effort than I am willing to invest in the next few weeks. Feel free to pick up the work where I left.

Issue #2484, opened 2019-02-26
Spike: Discover and define models at runtime


bajtos:
Certain applications don't use static JSON files to define their models, they create model definitions at runtime, often by running model discovery to obtain the model schema. Once the model is defined, the application usually exposed it via REST API.

This can be easily done in LB 3.x as follows:

```ts
const MyModel = app.registry.createModel(name, properties, settings);
app.model(MyModel, {dataSource: 'some-datasource', public: true});
```

Let's find out how to allow LB 4.x users to do the same.

The scope of this spike is to research solutions for discovering and defining models in such way, that the infrastructure implemented in #2036 can pick up new definitions and build REST APIs.

## Acceptance criteria

- [ ] A proof-of-concept implementation of the necessary run-time parts.
- [ ] A demo application providing a new REST API endpoint that will:
    1. Accept a MySQL connection string (host+port+database+credentials) and a list of table names.
    2. Connect to the specified MySQL database.
    3. Discover model schemas from the given tables
    4. Define LB4 models for the given tables
    5. Expose these models via REST API using #2036 
    6. Return a list of URL paths of the newly added models (e.g. `/Products`, `/CoffeeShops`)

    We can use the public database from [loopback-getting-started](https://github.com/strongloop/loopback-getting-started)) when testing this new endpoint manually.
- [ ] List of GitHub issues - follow-up user stories covering the implementation. Documentation and blog post should be included in the implementation plan.
- [ ] Decide what to do about relations. The story #2483 is tracking REST API for relations, but I don't think it's a good idea to include relations in the first iteration. Create a new GH issue to add support for discovery & definition of relations, this new issue can be a spike or the real implementation, depending on whether the implementation would be clear enough.

jannyHou:
Things to consider when do the spike:

- We probably need an open model class(weakly-typed model class) which allows any property
  - Model metadata describes the data shape of backend
  - This story is particular for relational databases(at least the first iteration let's focus on relational dbs)
  - Model instance is used in CRUD operations
- What is the run time representation of the discovered model
- Take the validation into consideration: using model definition for validation

bajtos:
Here is a code snippet/mock-up that I wrote while discussing this feature recently:

```ts
const ds = new DataSource({/*config*/})
const discovered = await ds.connector.discover();
/* 
{
  name: 'Product',
  properties: {
    'name': {
      type: 'string'
    }
  }
}
*/

// we need to transform modelDefinitionData slightly
[...]

Issue #2740, opened 2019-04-12
Docs for exposing REST API of a Model with no custom classes


bajtos:
Documentation for the feature From model definition to REST API with no custom repository/controller classes #2036

## Acceptance criteria

 - [ ] Docs for application developers consuming this feature. Describe the new CLI command, how to use controller/repositories contributed by extensions, what model config options are available.

 - [ ] Docs for  extension developers building custom controller/repository classes. Explain what API contract the custom controller/repository class must implement, show how the component must export the classes so that they can be picked by the Booter.

dhmlau:
Cross posting from https://github.com/strongloop/loopback-next/issues/4296#issuecomment-603361853 from @raymondfeng : 

> Existing docs:
> 
> - https://github.com/strongloop/loopback-next/blob/master/docs/site/Creating-CRUD-REST-apis.md (README)
> - https://github.com/strongloop/loopback-next/blob/master/docs/site/Rest-Crud-generator.md
> - https://github.com/strongloop/loopback-next/blob/master/examples/rest-crud/README.md (README)
> I think we can convert https://github.com/strongloop/loopback-next/blob/master/docs/site/Creating-CRUD-REST-apis.md into a regular page that glues the runtime/cli/example today to tell a story.

PR #4235, opened 2019-12-02
Spike: Discover and define models at runtime

bajtos:
This pull request implements a controller allowing clients to define models & expose their REST APIs at runtime, with no model/repository/controller source files required.

Cross-posting from https://github.com/strongloop/loopback-next/issues/2484:

1. Accept a MySQL connection string (host+port+database+credentials) and a list of table names.
2. Connect to the specified MySQL database.
3. Discover model schemas from the given tables
4. Define LB4 models for the given tables
5. Expose these models via REST API
6. Return a list of URL paths of the newly added models (e.g. `/Products`, `/CoffeeShops`)

> We can use the public database from [loopback-getting-started](https://github.com/strongloop/loopback-getting-started)) when testing this new endpoint manually.

Please note that public database is no longer available. What I did instead:
 - Setup a local MySQL server (e.g. via Docker)
 - Run the following SQL script to create `Coffeeshop` table and seed it with some records: https://github.com/strongloop/loopback-getting-started/blob/master/bin/demo.sql

## Try it out

1. Setup the MySQL server, create some tables
1. Checkout the feature branch `spike/import-model-at-runtime`
2. `npm install && npm build`
3. Open API Explorer - http://127.0.0.1:3000/explorer/#/ModelAdminController/ModelAdminController.discoverAndPublishModels
4. Post a valid MySQL connection string and a list of existing tables, for example:

    ```json
    {
      "connectionString": "mysql://root@localhost/test",
      "tableNames": [
        "Coffeeshop"
      ]
    }
    ```

5. Reload the API Explorer to see the new endpoints for your discovered model(s)

## Status

- [x] Proof of concept (implementation)
- [x] Spike document, including a list of GitHub issues - follow-up user stories covering the implementation.

### DIFF
--- a/examples/todo/src/controllers/model-admin.controller.ts
+++ b/examples/todo/src/controllers/model-admin.controller.ts
@@ -1,0 +1,242 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/example-todo
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {CoreBindings, inject} from '@loopback/core';
+import {
+  AnyObject,
+  Entity,
+  juggler,
+  model,
+  Model,
+  ModelDefinition,
+  property,
+} from '@loopback/repository';
+import {getModelSchemaRef, post, requestBody} from '@loopback/rest';
+import {
+  CrudRestControllerOptions,
+  defineCrudRepositoryClass,
+  defineCrudRestController,
+} from '@loopback/rest-crud';
+import * as assert from 'assert';
+import {TodoListApplication} from '..';
+
+@model()
+class DiscoverRequest extends Model {
+  @property({
+    required: true,
+    description: 'Database connection string (url), only MySQL is supported.',
+  })
+  connectionString: string;
+
+  @property({
+    type: 'array',
+    itemType: 'string',
+    description: 'List of table models to discover & load',
+  })
+  tableNames: string[];
+
+  constructor(data?: Partial<DiscoverRequest>) {
+    super(data);
+  }
+}
+
+@model({settings: {strict: false}})
+class DiscoverResponse extends Model {
+  @property({
+    type: 'object',
+    required: true,
+    description:
+      'A map from table names to endpoint paths, e.g. PRODUCT -> /products',
+  })
+  endpoints: {
+    [tableName: string]: string;
+  };
+
+  constructor(data?: Partial<DiscoverResponse>) {
+    super(data);
+    if (!this.endpoints) this.endpoints = {};
+  }
+}
+
+export class ModelAdminController {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE)
+    private app: TodoListApplication,
+  ) {}
+
+  @post('/discover', {
+    responses: {
+      200: {
+        description: 'Information about discovered models',
+        content: {
+          'application/json': {schema: getModelSchemaRef(DiscoverResponse)},
+        },
+      },
+    },
+  })
+  async discoverAndPublishModels(
+    @requestBody() {connectionString, tableNames}: DiscoverRequest,
+  ): Promise<DiscoverResponse> {
+    const result = new DiscoverResponse();
+
+    const ds = await getDataSourceForConnectionString(
+      this.app,
+      connectionString,
+    );
+
+    for (const table of tableNames) {
+      console.log('Discovering table %j', table);
+
+      const modelDef = await discoverModelDefinition(ds, table);
+
+      const modelClass = defineModelClass(Entity, modelDef);
+      const basePath = '/' + modelDef.name;
+      bootModelApi(this.app, ds.name, modelClass, {basePath});
+
+      result.endpoints[table] = basePath;
+    }
+
+    return result;
+  }
+}
+
+let dbCounter = 0;
+
+async function getDataSourceForConnectionString(
+  app: TodoListApplication,
+  connectionString: string,
+) {
+  // To keep our proof-of-concept simple, we create a new datasource instance
+  // for each new discovery request.
+  // In a real application, we should probably re-use datasource instances
+  // sharing the same connection string.
+  const dsName = `db-${++dbCounter}`;
+  console.log('Connecting to %j - datasource %j', connectionString, dsName);
+  const ds = new juggler.DataSource({
+    name: dsName,
+    connector: require('loopback-connector-mysql'),
+    url: connectionString,
+  });
+  await ds.connect();
+  app.dataSource(ds, dsName);
+  assert.equal(ds.name, dsName);
+  return ds;
+}
+
+// Ideally, `@loopback/repository` should provide a function to discover
+// model definition in LB4 format. For example, it could be responsibility
+// of a DataSource, in which case we need to implement custom DataSource
+// class backed by juggler DataSource but performing any necessary conversions.
+
+async function discoverModelDefinition(
+  ds: juggler.DataSource,
+  table: string,
+): Promise<ModelDefinition> {
+  // FIXME(bajtos) fix discoverSchemas, it does not return PromiseOrVoid
+  const discoveredModels = (await ds.discoverSchemas(table)) as AnyObject;
+  /* Example output from MySQL:
+     {
+       'test.CoffeeShop': {
+         name: 'Coffeeshop',
+         options: {
+           idInjection: false,
+           mysql: {schema: 'test', table: 'CoffeeShop'},
+         },
+         properties: {
+           id: {
+             type: 'Number',
+             required: true,
+             length: null,
+             precision: 10,
+             scale: 0,
+             id: 1,
+             mysql: {
+               columnName: 'id',
+               dataType: 'int',
+               dataLength: null,
+               dataPrecision: 10,
+               dataScale: 0,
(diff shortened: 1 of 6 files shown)


raymondfeng:
@bajtos Thank you for kicking off the topic.

I have been thinking about leveraging Kubernetes's `Operator` and `Custom Resource Definition` pattern/formats to achieve similar goals recently. 

- Each artifact will be described as CRD
- Allow extensions for artifact controllers/operators

See:
- https://kubernetes.io/docs/concepts/extend-kubernetes/operator/
- https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/
- https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/

packages/rest/src/rest.application.ts
```diff
@@ -133,6 +133,11 @@ export class RestApplication extends Application implements HttpServerLike {
     this.restServer.basePath(path);
   }
 
+  controller<T>(controllerCtor: ControllerClass<T>, name?: string): Binding {
+    this.restServer.invalidateRoutingCache();
```
raymondfeng: I would rather to use a `ContextView` to invalidate the routing cache.

bajtos: +1, I would like to see a ContextView-based solution too. In this spike, I used the fastest path to get done what I need.

@raymondfeng Do you have bandwidth to implement ContextView-based version yourself? Here is the GitHub issue tracking this user story: https://github.com/strongloop/loopback-next/issues/433

examples/todo/src/controllers/model-admin.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+// model definition in LB4 format. For example, it could be responsibility
+// of a DataSource, in which case we need to implement custom DataSource
+// class backed by juggler DataSource but performing any necessary conversions.
+
+async function discoverModelDefinition(
+  ds: juggler.DataSource,
+  table: string,
+): Promise<ModelDefinition> {
+  // FIXME(bajtos) fix discoverSchemas, it does not return PromiseOrVoid
+  const discoveredModels = (await ds.discoverSchemas(table)) as AnyObject;
```
bajtos: See https://github.com/strongloop/loopback-datasource-juggler/pull/1807

examples/todo/src/controllers/model-admin.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+  dsName: string,
+  modelClass: typeof Entity,
+  options: CrudRestControllerOptions,
+) {
+  const RepositoryClass = defineCrudRepositoryClass(modelClass);
+  inject(`datasources.${dsName}`)(RepositoryClass, undefined, 0);
+  const repoBinding = app.repository(RepositoryClass);
+  console.log('Defined repository %s', RepositoryClass.name);
+
+  const ControllerClass = defineCrudRestController(modelClass, options);
```
bajtos: Cross-posting from a Slack chat:

> In LB3, we have a feature to expose the models over REST or not via `public:true/false` parameter and we need this feature as well as we are not exposing the models via REST. We are accessing them in nodejs and don't want to expose those routes

Good point! This is certainly possible. If you look at how `bootModelApi` is implemented, it does two things: define a repository class to connect the model with a datasource, and then define a controller class to build REST API. You can skip controller creation to get a model that's accessible inside the server code only.

examples/todo/src/controllers/model-admin.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+// This will be implemented by
+// https://github.com/strongloop/loopback-next/issues/2036
+// In particular https://github.com/strongloop/loopback-next/issues/3737
+function bootModelApi(
+  app: TodoListApplication,
+  dsName: string,
+  modelClass: typeof Entity,
+  options: CrudRestControllerOptions,
+) {
+  const RepositoryClass = defineCrudRepositoryClass(modelClass);
```
bajtos: I am thinking about moving `defineCrudRepositoryClass` to `@loopback/repository`. I see the function as very coupled to `DefaultCrudRepository` class, not adding anything opinionated or controversial on top of the named constructor.

examples/todo/src/controllers/model-admin.controller.ts
```diff
@@ -0,0 +1,242 @@
...
+  return new ModelDefinition({
+    name: jugglerDef.name,
+    // TODO: convert from juggler/LB3 style to LB4
+    // For example, we need to transform array-type definitions.
+    properties: jugglerDef.properties,
+    settings: jugglerDef.options,
+  });
+}
+
+// TODO: move this function to '@loopback/repository'
```
bajtos: See https://github.com/strongloop/loopback-next/pull/4266

bajtos:
In a878d8301, I have simplified the demo controller implementation using the recently released features strongloop/loopback-datasource-juggler#1807 and #4266. I have also reorganized the code to make it more clear how a model can stay private (not exposed via REST API).

I consider the demo app as done and proposing the following follow-up stories:

1. Dynamic binding/rebinding of controllers after app start #433 _(already exists)_

2. How to define models, repositories and controllers at runtime

    The idea is create a new documentation page that will explain how to:
  
    1. Define a model class from a `ModelDefinition` object
    2. Define a repository class for the given model, how to configure dependencies (the datasource to use)
    3. Define a controller class for the given model & repository, how to configure dependencies (the repository to use)
    4. Discover model definition from a database, build `ModelDefinition` object from the discovered schema.

    The page can use code snippets from my demo controller to illustrate the approach in practice.

In my demo, I have a comment mentioning that it would be nice to have a LB4-native method for discovering model schemas, a method that return LB4 `ModelDefinition` object instead of LB3-like data. Considering that we don't know how many people will actually implement model discovery at runtime, I am proposing to leave this feature out of scope of the initial work.

@strongloop/sq-lb-apex Please take a look and let me know if you have any comments or suggestions. I'd like to get at least one approval and then I'll move on to create the follow-up story/stories.

packages/rest/src/rest.server.ts
```diff
@@ -703,6 +713,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
     let spec = this.getSync<OpenApiSpec>(RestBindings.API_SPEC);
     const defs = this.httpHandler.getApiDefinitions();
 
+    // Apply shallow-clone to prevent modification of user-provided API_SPEC
```
jannyHou: @bajtos could you explain more about the shallow clone here?
Wouldn't it ALLOW modification of user-provided API_SPEC?

The line below it(L#706) applies deep clone which prevents the modification of the original spec.

bajtos: This is just a temporary workaround to get my demo working. The actual implementation is likely to be different, see the discussion in _Dynamic binding/rebinding of controllers after app start_ #433.

examples/todo/src/controllers/model-admin.controller.ts
```diff
@@ -0,0 +1,221 @@
...
+    console.log('Defined model %s', ModelClass.name);
+
+    // Step 3: define a repository class adding CRUD behavior to our model
+    const RepositoryClass = defineCrudRepositoryClass(ModelClass);
+    inject(`datasources.${ds.name}`)(RepositoryClass, undefined, 0);
+    const repoBinding = this.app.repository(RepositoryClass);
+    console.log('Defined repository %s', RepositoryClass.name);
+
+    // Step 4: Optionally, expose the new model via REST API
+    const basePath = '/' + modelDef.name;
```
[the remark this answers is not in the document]
bajtos: Sure, plural names would be better. There are different ways how to build plurals, I am leaving it up to our users to decide.

bajtos:
Thank you @jannyHou and @agnes512 for thoughtful comments!

> I only have one suggestion. If I were a user who uses this feature a lot, I'd like to know how I can modify/config the controller to match my requirements. It'd be good if we can make this part of documentation clear/detailed in the follow-up story.

Good idea :+1:

In my mind, this is related to the epic _From model definition to REST API_ https://github.com/strongloop/loopback-next/issues/2036, where we already have a story to document how can extension authors build custom controller factories - see https://github.com/strongloop/loopback-next/issues/2740. 

Let's include #2740 in the follow-up stories for this spike.

> (maybe another story)From UX's perspective...shall we create another endpoint that prints out the generated model def, like `GET /showModelDefByName/{modelname}`.
When testing the dynamically generated APIs, people can  learn the signature of endpoints from /Todos, but It's not very easy to figure out the model properties by only having the path of the model. 

I see your point. At the moment, you can find model properties in OpenAPI schema. 

It is possible to build an endpoint you have described, or perhaps add the model definition to the response returned by `POST /discover` endpoint. I feel those changes are out of scope of this spike/demo app though. The purpose of this work is to discover missing building blocks and ensure everything fits together.

IMO, real world applications will never follow exactly our demo version, they will always combine these blocks in unique ways.

bajtos:
> How to define models, repositories and controllers at runtime

Created https://github.com/strongloop/loopback-next/issues/4296

bajtos:
Follow-up stories:

- Dynamic binding/rebinding of controllers after app start #433.
- How to build models, repositories and controllers dynamically at runtime #4296
- Docs for exposing REST API of a Model with no custom classes #2740

I am closing this spike as done.

PR #4240, opened 2019-12-02
feat(repository): add title property to Count schema definition

dougal83:
Add title property to Count schema definition under common types in order to enable better model
generation by third party tools such as https://github.com/OpenAPITools/openapi-generator

Without a title property the generator would simply name the count response as InlineObject<n>.  By adding the title property, the tooling can produce a model named after the object title.

## Checklist

### DIFF
--- a/packages/repository/src/common-types.ts
+++ b/packages/repository/src/common-types.ts
@@ -98,5 +98,6 @@ export interface Count {
  */
 export const CountSchema = {
   type: 'object',
+  title: 'loopback.count',
   properties: {count: {type: 'number'}},
 };


packages/repository/src/common-types.ts
```diff
@@ -98,5 +98,6 @@ export interface Count {
  */
 export const CountSchema = {
   type: 'object',
+  title: 'Count',
```
raymondfeng: Do we want to make the title qualified such as `loopback.Count`? Would it confuse code generation tools?

dougal83: You are right, I was concerned by naming collisions that tend to cause issues with some generation tools in particular circumstances.  Qualifying this property will result in a string conversion such as camelCase/PascalCase, so no problem(on aforementioned tool).

Updated to ``loopback.count`` as per feedback.

dhmlau:
In case you're interested, there is a LoopBack Developer badge for those who made contributions to LoopBack. Please check this out:  https://www.ibm.com/services/learning/ites.wss/zz-en?pageType=badges&id=0de05804-8037-4791-ac83-09829b566d51. Thanks!

PR #4246, opened 2019-12-03
chore: upgrade Node to 12.x for Travis jobs

bajtos:
With Node.js version 12 already in LTS mode, it's time to upgrade our Travis jobs to use the latest & greatest & fastest! version of Node.

Also note that the version 8 is going end of life soon, it's time to upgrade to a newer one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       os: osx
       # using the default script to build & run the tests
 
-    - node_js: "8"
+    - node_js: "12"
       os: linux
       env: TASK=code-lint
       # Running Code Linter -- Requires @loopback/build so it's bootstrapped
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -51,7 +51,7 @@ matrix:
         - npm run postinstall -- --scope "@loopback/test-repository-mongodb" --include-dependencies
         - npm run build -- --scope "@loopback/test-repository-mongodb" --include-dependencies
         - cd acceptance/repository-mongodb && npm run mocha
-    - node_js: "10"
+    - node_js: "12"
```
emonddr: why are we changing "10" to "12" then?
The title says we need to change "8" to "12"

bajtos: We don't need to change anything yet. But I _want_ to upgrade to the latest LTS version.

bajtos: I reworked the commit message as follows:

```
chore: upgrade Node to 12.x for Travis jobs

With Node.js version 12 already in LTS mode, it's time to upgrade
our Travis jobs to use the latest & greatest & fastest! version of Node.

Also note that then version 8 is going end of life soon, it's time
to upgrade to a newer one.
```

LGTY now?

PR #4258, opened 2019-12-04
feat: openapi spec contributor extension point

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/3854

- Created a service as extension point for OpenAPI spec enhancers under folder `src/enhancers`.
  - the extension point is created in file `spec-enhancer.service.ts`
  - `types.ts` describes the extension interface
  - `keys.ts` gives the extensionpoint a key

- Added two extensions in test fixtures, they are also the demo example for extension contributors
  - add info spec
  - add security spec

- Added unit test for service utility functions like `findEnhancerByName`, `applyEnhancerByName`, `generateSpec`

- Feedback addressed: 
  - renamed the extension point to be enhancer instead of contributor(the name is still open for discussion, opinions are welcomed.)
  - based on discussion in [comment](https://github.com/strongloop/loopback-next/pull/4258#discussion_r355541762), I added a default spec merge function leveraging `json-merge-patch`, it's also used by module `ajv`.
  - added function `findEnhancerByName` and `applyEnhancerByName` to allow people do the merge with a single enhancer, I will enable loading the enhancers by group names in a separate PR.

- TBD: apply the extensionpoint to rest and other related modules, create corresponding extensions like controller spec builder, security spec builder, etc...in follow-up stories.

- TBD: load the enhancers by group names in a separate PR.

## Checklist

### DIFF
--- a/docs/site/Extending-OpenAPI-specification.md
+++ b/docs/site/Extending-OpenAPI-specification.md
@@ -1,0 +1,153 @@
+---
+lang: en
+title: 'Extending OpenAPI Specification'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Extending-OpenAPI-specification.html
+---
+
+## OpenAPI Specification Enhancer
+
+The APIs in a LoopBack `RestApplication` are described by the
+[OpenAPI Specification (short for OAS)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md).
+An application's OAS is mainly generated from
+[controllers](https://loopback.io/doc/en/lb4/Controllers.html) and their
+members' metadata. Besides this, we would also like to contribute specifications
+from other places. Therefore, an extension point `OASEnhancerService` is created
+to allow registered extensions to provide their OAS fragments and modify a rest
+application's specification.
+
+_Read about the extension point/extension pattern in
+[documentation](Extension-point-and-extensions.md)_
+
+## Adding a New OAS Enhancer
+
+Interface `OASEnhancer` is created in `@loopback/openapi-v3` to describe the
+specification enhancers. A typical OAS enhancer class should have a string type
+`name` field and a function `modifySpec()` to modify the current specification.
+
+For example, to modify the `info` field of an OAS, you can create an
+`InfoSpecEnhancer` that implements interface `OASEnhancer` as follows:
+
+```ts
+import {bind} from '@loopback/core';
+import {
+  mergeOpenAPISpec,
+  asSpecEnhancer,
+  OASEnhancer,
+  OpenApiSpec,
+} from '@loopback/openapi-v3';
+
+/**
+ * A spec enhancer to add OpenAPI info spec
+ */
+@bind(asSpecEnhancer)
+export class InfoSpecEnhancer implements OASEnhancer {
+  // give your enhancer a proper name
+  name = 'info';
+
+  // takes in the current spec, modifies it, and returns a new one
+  modifySpec(spec: OpenApiSpec): OpenApiSpec {
+    const InfoPatchSpec = {
+      info: {title: 'LoopBack Test Application', version: '1.0.1'},
+    };
+    // the example calls a default helper function to merge the fragment spec.
+    const mergedSpec = mergeOpenAPISpec(spec, InfoPatchSpec);
+    return mergedSpec;
+  }
+}
+```
+
+- The class is decorated with a binding template `asSpecEnhancer`.
+- The enhancer has a name as `info`. Name can be used to retrieve a certain
+  enhancer (explained in the
+  [extension point section](#oas-enhancer-service-as-extension-point)).
+- The enhancer changes the current specification's `info` object in function
+  `modifySpec`.
+- It calls [`mergeOpenAPISpec`](#default-merge-function) to merge the
+  specification fragment into the current spec.
+
+### Default Merge Function
+
+Since `modifySpec` has full access to the current spec, it can perform any
+operation: merge, delete, or more complicated changes. This is totally
+determined by the extension contributor.
+
+To apply the basic merging, we provide a default helper function called
+`mergeOpenAPISpec` that leverages
+[`json-merge-patch`](https://github.com/pierreinglebert/json-merge-patch) to
+merge two json objects. You can find its usage in the
+[previous section](#adding-a-new-oas-enhancer)
+
+### Registering an Enhancer
+
+After decorating your enhancer properly with `@bind(asSpecEnhancer)`, you can
+bind it to your application as follows:
+
+```ts
+import {createBindingFromClass} from '@loopback/core';
+import {InfoSpecEnhancer} from './enhancers/infoSpecEnhancer';
+
+class MyApplication extends RestApplication {
+  constructor() {
+    super();
+    this.add(createBindingFromClass(InfoSpecEnhancer));
+  }
+}
+```
+
+## OAS Enhancer Service as Extension Point
+
+The OAS enhancer extension point is created in package `@loopback/openapi-v3`.
+It organizes the registered OAS enhancers, and provides APIs to either apply one
+enhancer by name, or apply all enhancers automatically.
+
+### Registering an Enhancer Service
+
+You can bind the OAS enhancer extension point to your app via key
+`OAS_ENHANCER_SERVICE`:
+
+```ts
+import {RestApplication} from '@loopback/rest';
+import {OASEnhancerService, OAS_ENHANCER_SERVICE} from '@loopback/openapi-v3';
+
+class MyApplication extends RestApplication {
+  constructor() {
+    super();
+    this.add(
+      createBindingFromClass(OASEnhancerService, {
+        key: OAS_ENHANCER_SERVICE,
+      }),
+    );
+  }
+
+  // define a function to return a spec service by the same key
+  getSpecService() {
+    return this.get(OAS_ENHANCER_SERVICE);
+  }
+}
+```
+
+### Applying Registered Enhancers
+
+To automatically apply all the registered enhancers, call `applyAllEnhancers`:
+
+```ts
+await app.getSpecService.applyAllEnhancers();
+```
+
+_In the future we will support applying enhancers by a custom sequence. The
+sequence will be determined by a combination of group names and the alphabetical
+order._
+
+To retrieve an enhancer by name, call `getEnhancerByName`:
+
+```ts
+await app.getSpecService.getEnhancerByName('info');
+```
+
+To apply an enhancer by name, call `applyEnhancerByName`:
+
+```ts
+await app.getSpecService.applyEnhancerByName('info');
+```
--- a/packages/openapi-v3/package.json
+++ b/packages/openapi-v3/package.json
@@ -6,11 +6,12 @@
     "node": ">=8.9"
   },
   "dependencies": {
-    "@loopback/context": "^1.25.0",
     "@loopback/repository-json-schema": "^1.11.3",
+    "@loopback/core": "^1.12.0",
     "debug": "^4.1.1",
     "lodash": "^4.17.15",
-    "openapi3-ts": "^1.3.0"
+    "openapi3-ts": "^1.3.0",
+    "json-merge-patch": "^0.2.3"
   },
   "devDependencies": {
     "@loopback/build": "^3.0.0",
--- a/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/application.ts
+++ b/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/application.ts
@@ -1,0 +1,28 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Application, createBindingFromClass} from '@loopback/core';
+import {OASEnhancerService, OAS_ENHANCER_SERVICE} from '../../../..';
+import {InfoSpecEnhancer} from './info.spec.extension';
+import {SecuritySpecEnhancer} from './security.spec.extension';
+
+export class SpecServiceApplication extends Application {
+  constructor() {
+    super();
+    this.add(
+      createBindingFromClass(OASEnhancerService, {
+        key: OAS_ENHANCER_SERVICE,
+      }),
+    );
+    this.add(createBindingFromClass(SecuritySpecEnhancer));
+    this.add(createBindingFromClass(InfoSpecEnhancer));
+  }
+
+  async main() {}
+
+  getSpecService() {
+    return this.get(OAS_ENHANCER_SERVICE);
+  }
+}
--- a/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/security.spec.extension.ts
+++ b/packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/security.spec.extension.ts
@@ -1,0 +1,44 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {bind} from '@loopback/core';
+import debugModule from 'debug';
+import {inspect} from 'util';
+import {
+  mergeOpenAPISpec,
+  ReferenceObject,
+  SecuritySchemeObject,
+} from '../../../..';
+import {asSpecEnhancer, OASEnhancer} from '../../../../enhancers/types';
+import {OpenApiSpec} from '../../../../types';
+const debug = debugModule('loopback:openapi:spec-enhancer');
+
+export type SecuritySchemeObjects = {
+  [securityScheme: string]: SecuritySchemeObject | ReferenceObject;
+};
+
+export const SECURITY_SCHEME_SPEC: SecuritySchemeObjects = {
+  bearerAuth: {
+    type: 'http',
+    scheme: 'bearer',
+    bearerFormat: 'JWT',
+  },
+};
+
+/**
+ * A spec enhancer to add bearer token OpenAPI security entry to
+ * `spec.component.securitySchemes`
+ */
+@bind(asSpecEnhancer)
+export class SecuritySpecEnhancer implements OASEnhancer {
+  name = 'security';
+
+  modifySpec(spec: OpenApiSpec): OpenApiSpec {
+    const patchSpec = {components: {securitySchemes: SECURITY_SCHEME_SPEC}};
+    const mergedSpec = mergeOpenAPISpec(spec, patchSpec);
+    debug(`security spec extension, merged spec: ${inspect(mergedSpec)}`);
+    return mergedSpec;
+  }
+}
--- a/packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
@@ -1,0 +1,96 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {OASEnhancerService} from '../../..';
+import {SpecServiceApplication} from './fixtures/application';
+
+describe('spec-enhancer-extension-point', () => {
+  let app: SpecServiceApplication;
+  let specService: OASEnhancerService;
+
+  beforeEach(givenAppWithSpecService);
+
+  it('setter - can set spec', async () => {
+    const EXPECTED_SPEC = {
+      openapi: '3.0.0',
+      info: {title: 'LoopBack Application', version: '1.0.0'},
+      // setter adds paths spec based on the default spec
+      paths: getSamplePathSpec(),
+    };
+    const PATHS_SPEC = getSamplePathSpec();
+    specService.spec = {...specService.spec, paths: PATHS_SPEC};
+    expect(specService.spec).to.eql(EXPECTED_SPEC);
+  });
+
+  it('generateSpec - loads and creates spec for ALL registered extensions', async () => {
+    const EXPECTED_SPEC = {
+      openapi: '3.0.0',
+      // info object is updated by the info enhancer
+      info: {title: 'LoopBack Test Application', version: '1.0.1'},
+      paths: {},
+      // the security scheme entry is added by the security enhancer
+      components: {
+        securitySchemes: {
+          bearerAuth: {
+            type: 'http',
+            scheme: 'bearer',
+            bearerFormat: 'JWT',
+          },
+        },
+      },
+    };
+    const specFromService = await specService.applyAllEnhancers();
+    expect(specFromService).to.eql(EXPECTED_SPEC);
+  });
+
+  it('getEnhancerByName - returns the enhancer with a given name', async () => {
+    const enhancer = await specService.getEnhancerByName('info');
+    expect(enhancer).to.not.be.undefined();
+    expect(enhancer?.name).to.equal('info');
+  });
+
+  it('applyEnhancerByName - returns the merged spec after applying a given enhancer', async () => {
+    const EXPECTED_SPEC = {
+      openapi: '3.0.0',
+      // info object is updated by the info enhancer
+      info: {title: 'LoopBack Test Application', version: '1.0.1'},
+      paths: {},
+    };
+    const mergedSpec = await specService.applyEnhancerByName('info');
+    expect(mergedSpec).to.eql(EXPECTED_SPEC);
+  });
+
+  async function givenAppWithSpecService() {
+    app = new SpecServiceApplication();
+    specService = await app.getSpecService();
+  }
+
+  function getSamplePathSpec() {
+    return {
+      '/pets': {
+        get: {
+          description:
+            'Returns all pets from the system that the user has access to',
+          responses: {
+            '200': {
+              description: 'A list of pets.',
+              content: {
+                'application/json': {
+                  schema: {
+                    type: 'array',
+                    items: {
+                      $ref: '#/components/schemas/pet',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    };
+  }
+});
--- a/packages/openapi-v3/src/enhancers/index.ts
+++ b/packages/openapi-v3/src/enhancers/index.ts
@@ -1,0 +1,8 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+export * from './keys';
+export * from './spec-enhancer.service';
+export * from './types';
--- a/packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
+++ b/packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
@@ -1,0 +1,121 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {config, extensionPoint, extensions, Getter} from '@loopback/core';
+import debugModule from 'debug';
+import * as _ from 'lodash';
+import {inspect} from 'util';
+import {OpenApiSpec} from '../types';
+import {OASEnhancer, OAS_ENHANCER_EXTENSION_POINT_NAME} from './types';
+const jsonmergepatch = require('json-merge-patch');
+
+const debug = debugModule('loopback:openapi:spec-enhancer');
+
+/**
+ * Options for the OpenAPI Spec enhancer extension point
+ */
+export interface OASEnhancerServiceOptions {
+  // no-op
+}
+
+/**
+ * An extension point for OpenAPI Spec enhancement
+ * This service is used for enhancing an OpenAPI spec by loading and applying one or more
+ * registered enhancers.
+ *
+ * A typical use of it would be generating the OpenAPI spec for the endpoints on a server
+ * in the `@loopback/rest` module.
+ */
+@extensionPoint(OAS_ENHANCER_EXTENSION_POINT_NAME)
+export class OASEnhancerService {
+  constructor(
+    /**
+     * Inject a getter function to fetch spec enhancers
+     */
+    @extensions()
+    private getEnhancers: Getter<OASEnhancer[]>,
+    /**
+     * An extension point should be able to receive its options via dependency
+     * injection.
+     */
+    @config()
+    public readonly options?: OASEnhancerServiceOptions,
+  ) {}
+
+  private _spec: OpenApiSpec = {
+    openapi: '3.0.0',
+    info: {
+      title: 'LoopBack Application',
+      version: '1.0.0',
+    },
+    paths: {},
+  };
+
+  /**
+   * Getter for `_spec`
+   */
+  get spec(): OpenApiSpec {
+    return this._spec;
+  }
+  /**
+   * Setter for `_spec`
+   */
+  set spec(value: OpenApiSpec) {
+    this._spec = value;
+  }
+
+  /**
+   * Find an enhancer by its name
+   * @param name The name of the enhancer you want to find
+   */
+  async getEnhancerByName(name: string): Promise<OASEnhancer | undefined> {
+    // Get the latest list of enhancers
+    const enhancers = await this.getEnhancers();
+    return enhancers.find(e => e.name === name);
+  }
+
+  /**
+   * Apply a given enhancer's merge function. Return the latest _spec.
+   * @param name The name of the enhancer you want to apply
+   */
+  async applyEnhancerByName(name: string): Promise<OpenApiSpec> {
+    const enhancer = await this.getEnhancerByName(name);
+    if (enhancer) this._spec = enhancer.modifySpec(this._spec);
+    return this._spec;
+  }
+
+  /**
+   * Generate OpenAPI spec by applying ALL registered enhancers
+   * TBD: load enhancers by group names
+   */
+  async applyAllEnhancers(options = {}): Promise<OpenApiSpec> {
+    const enhancers = await this.getEnhancers();
+    if (_.isEmpty(enhancers)) return this._spec;
+    for (const e of enhancers) {
+      this._spec = e.modifySpec(this._spec);
+    }
+    debug(`Spec enhancer service, generated spec: ${inspect(this._spec)}`);
+    return this._spec;
+  }
+}
+
+/**
+ * The default merge function to patch the current OpenAPI spec.
+ * It leverages module `json-merge-patch`'s merge API to merge two json objects.
+ * It returns a new merged object without modifying the original one.
+ *
+ * A list of merging rules can be found in test file:
+ * https://github.com/pierreinglebert/json-merge-patch/blob/master/test/lib/merge.js
+ *
+ * @param currentSpec The original spec
+ * @param patchSpec The patch spec to be merged into the original spec
+ */
+export function mergeOpenAPISpec(
+  currentSpec: Partial<OpenApiSpec>,
+  patchSpec: Partial<OpenApiSpec>,
+) {
+  const mergedSpec = jsonmergepatch.merge(currentSpec, patchSpec);
+  return mergedSpec;
+}
--- a/packages/openapi-v3/src/enhancers/types.ts
+++ b/packages/openapi-v3/src/enhancers/types.ts
@@ -1,0 +1,30 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {BindingTemplate, extensionFor} from '@loopback/core';
+import {OpenApiSpec} from '../types';
+
+/**
+ * Typically an extension point defines an interface as the contract for
+ * extensions to implement
+ */
+export interface OASEnhancer {
+  name: string;
+  modifySpec(spec: OpenApiSpec): OpenApiSpec;
+}
+
+/**
+ * Name/id of the OAS enhancer extension point
+ */
+export const OAS_ENHANCER_EXTENSION_POINT_NAME = 'oas-enhancer';
+
+/**
+ * A binding template for spec contributor extensions
+ */
+export const asSpecEnhancer: BindingTemplate = binding => {
+  extensionFor(OAS_ENHANCER_EXTENSION_POINT_NAME)(binding);
+  // is it ok to have a different namespace than the extension point name?
+  binding.tag({namespace: 'oas-enhancer'});
+};
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -288,6 +288,10 @@ children:
     url: Extension-life-cycle.html
     output: 'web, pdf'
 
+  - title: 'Extending OpenAPI specification'
+    url: Extending-OpenAPI-specification.html
+    output: 'web, pdf'
+
   - title: 'Testing your extension'
     url: Testing-your-extension.html
     output: 'web, pdf'
--- a/packages/openapi-v3/package-lock.json
+++ b/packages/openapi-v3/package-lock.json
@@ -30,6 +30,112 @@
 				"ms": "^2.1.1"
 			}
 		},
+		"deep-equal": {
+			"version": "1.1.1",
+			"resolved": "https://registry.npmjs.org/deep-equal/-/deep-equal-1.1.1.tgz",
+			"integrity": "sha512-yd9c5AdiqVcR+JjcwUQb9DkhJc8ngNr0MahEBGvDiJw8puWab2yZlh+nkasOnZP+EGTAP6rRp2JzJhJZzvNF8g==",
+			"requires": {
+				"is-arguments": "^1.0.4",
+				"is-date-object": "^1.0.1",
+				"is-regex": "^1.0.4",
+				"object-is": "^1.0.1",
+				"object-keys": "^1.1.1",
+				"regexp.prototype.flags": "^1.2.0"
+			}
+		},
+		"define-properties": {
+			"version": "1.1.3",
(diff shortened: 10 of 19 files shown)


packages/openapi-v3/src/extension-point/types.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface OAISpecContributor {
+  /**
+   * The name of root level fields in an OpenAPI specification,
+   * e.g. ROOT(the entire spec), info, paths, security, tags, etc...
+   *
+   *
+   * // question: should we use a union type to enum the valid fields
+   * // instead of using string?
+   */
+  fieldName: string;
```
raymondfeng: I don't think we need to have `fieldName` for filtering. Each contributor will have the opportunity to mutate the spec.

packages/openapi-v3/src/extension-point/types.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import {BindingTemplate} from '@loopback/context';
+import {extensionFor} from '@loopback/core';
+import {OpenApiSpec} from '../types';
+
+/**
+ * Typically an extension point defines an interface as the contract for
+ * extensions to implement
+ */
+export interface OAISpecContributor {
```
raymondfeng: Maybe `OpenApiSpecEnhancer`?

packages/openapi-v3/src/extension-point/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * The name of root level fields in an OpenAPI specification,
+   * e.g. ROOT(the entire spec), info, paths, security, tags, etc...
+   *
+   *
+   * // question: should we use a union type to enum the valid fields
+   * // instead of using string?
+   */
+  fieldName: string;
+  addSpec(spec: Partial<OpenApiSpec>): Partial<OpenApiSpec>;
```
raymondfeng: The method name can be `update`

packages/openapi-v3/src/extension-point/types.ts
jannyHou: then the implementation should be simpler, I don't have a very strong opinion on it and am glad to go with the simpler solution, it also gives a better flexibility.

packages/openapi-v3/src/extension-point/types.ts
jannyHou: hmm....I would expect the name implies the extension will modify the spec. I kept the original name just for demo, when we agree on the proper name, I will rename all the interfaces.

raymondfeng: Enhancer is a term that augments an existing artifact. See https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cglib/proxy/Enhancer.html.

I don't expect an extension to remove things from the spec, but it enhances the spec by adding additional information such as `security`.

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/application.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor() {
+    super();
+    this.component(SpecComponent);
+    this.add(createBindingFromClass(SecuritySpecContributor));
+    this.add(createBindingFromClass(InfoSpecContributor));
+  }
+
+  async main() {}
+
+  async getSpecService() {
```
raymondfeng: The `async` keyword is not needed here.

packages/openapi-v3/src/extension-point/component.ts
```diff
@@ -0,0 +1,12 @@
+import {createBindingFromClass} from '@loopback/context';
+import {Component} from '@loopback/core';
+import {SPEC_SERVICE} from './keys';
+import {SpecService} from './spec-contributor.service';
+
+export class SpecComponent implements Component {
```
raymondfeng: Maybe it's overkill to have a `SpecComponent`. We can register the `SpecService` as part of `RestComponent`.

packages/openapi-v3/src/extension-point/component.ts
```diff
@@ -0,0 +1,12 @@
+import {createBindingFromClass} from '@loopback/context';
```
raymondfeng: Let's use a different directory name for `extension-point`, which is generic.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,72 @@
...
+      version: '1.0.0',
+    },
+    paths: {},
+  };
+
+  /**
+   * Find contributors for a given field
+   * @param fieldName - The field name
+   */
+  async findContributors(): Promise<OAISpecContributor[] | undefined> {
```
raymondfeng: This method can be removed to use `this.getContributors()` directly.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.getContributors();
+  }
+
+  /**
+   * Generate info spec from contributors
+   */
+  async generateSpec(options = {}): Promise<OpenApiSpec> {
+    const contributors = await this.findContributors();
+    if (!contributors) return this._spec;
```
raymondfeng: We'll get `[]` if no contributors are found.

packages/openapi-v3/src/extension-point/component.ts
jannyHou: Good point agreed.

packages/openapi-v3/src/extension-point/component.ts
jannyHou: hmm, any suggestion? cuz I feel `extension-point` is generic and `spec-enhancer` is more specific

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
jannyHou: right. created it for filtering by field, not needed any more.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,69 @@
...
+      title: 'LoopBack Application',
+      version: '1.0.0',
+    },
+    paths: {},
+  };
+
+  get spec(): OpenApiSpec {
+    return this._spec;
+  }
+  set spec(value: OpenApiSpec) {
```
jannyHou: Copied from chat with @raymondfeng 

> when openapi spec is generated, we first build it from controller inspections, then call enhancers to update the spec

I add a setter for the `_spec` here for ^

packages/openapi-v3/src/extension-point/component.ts
raymondfeng: Maybe `enhancers`

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/application.ts
```diff
@@ -0,0 +1,23 @@
+import {Application, createBindingFromClass} from '@loopback/core';
+import {OAISpecEnhancerService, OAISPEC_ENHANCER_SERVICE} from '../../../../';
```
emonddr: Why do we use `OAISpec` instead of `OpenAPISpec` or `OASpec`?  What would the `I` mean in the acronym? Isn't the `I` part of the other acronym `API`?

packages/openapi-v3/src/__tests__/unit/spec-contributor/spec-contributor.extensions.unit.ts
```diff
@@ -0,0 +1,43 @@
...
+import {SpecServiceApplication} from './fixtures/application';
+
+describe('spec-enhancer-extension-point', () => {
+  let app: SpecServiceApplication;
+  let specService: OAISpecEnhancerService;
+
+  beforeEach(givenAppWithSpecComponent);
+  beforeEach(findSpecService);
+
+  it('greets by language', async () => {
```
emonddr: > greets by language 

Need to change

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/example-greeter-extension
```
emonddr: > @loopback/example-greeter-extension

Need to change

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    public readonly options?: OAISpecContributorServiceOptions,
+  ) {}
+
+  private _spec: OpenApiSpec = {
+    openapi: '3.0.0',
+    info: {
+      title: 'LoopBack Application',
+      version: '1.0.0',
+    },
+    paths: {},
```
emonddr: We need to set 

```
servers: [{url: '/'}],
```

as well. 

Please see : https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/application.ts#L68 

When we added the code to add security spec to the open api spec for this shopping example, we also had to set the servers `url` to `'/'`

emonddr:
@jannyHou , great start :)

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/application.ts
jannyHou: Agreed, double checked with https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#introduction, I think a more accurate abbreviation should be `OAS`, which stands for OpenAPI Specification. 

I used `OAI` because it's their github organization name.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
jannyHou: Please note not all root level fields are required in a complete OpenAPI spec, see [its interface](https://github.com/metadevpro/openapi3-ts/blob/master/src/model/OpenApi.ts#L18)

Like `servers: [{url: '/'}],` should be filled by our rest server component.

packages/openapi-v3/package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "@loopback/context": "^1.24.0",
     "@loopback/repository-json-schema": "^1.11.2",
+    "@loopback/core": "^1.11.0",
```
bajtos: If you need things from `core` that are not part of `context`, then please remove the dependency on `@loopback/context` on L9 and rework imports to use `@loopback/core` instead.

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/security.spec.extension.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * A spec contributor to add bearer token
+ * OpenAPI security entry
+ */
+@bind(asSpecEnhancer)
+export class SecuritySpecEnhancer implements OAISpecEnhancer {
+  modifySpec(spec: OpenApiSpec) {
+    spec.components = spec.components ?? {};
+    spec.components.securitySchemes = SECURITY_SCHEME_SPEC;
+    return spec;
```
bajtos: I am little bit concerned about the design where the spec-enhancer is modifying the spec directly. Every implementation has to deal with detecting whether optional objects like `spec.compontents` are set and be prepared to handle the case when they are not present. 

Have you considered a different approach, where  the enhancer provides new data to be added to the spec, and LB runtime ensures the data is cleverly merged together? 

Take the `InfoSpecEnhancer` for an example. IIUC, it will discard any `info` fields provided by the app developer and replace them with `info` object containing only `title` and `version`. This problem is easy to miss when reviewing the code. I would prefer to design spec-enhancing infrastructure in a way that will make errors like these more difficult to make.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+  set spec(value: OpenApiSpec) {
+    this._spec = value;
+  }
+
+  /**
+   * Generate OpenAPI spec from contributors
+   */
+  async generateSpec(options = {}): Promise<OpenApiSpec> {
+    const contributors = await this.getContributors();
```
bajtos: How are we going to decide in which order the contributors are applied? What if we have two contributors adding conflicting values for the same field in the spec - how is the user going to determine which contributor wins?

bajtos:
> TBD: rename the extensionpoint and related types/keys
>
> - OpenApiSpecEnhancer?
> - OpenApiSpecContributor?
> - other better name?

How about using the concept of a fragment from GraphQL? Each contributor can provide a fragment of the spec to be merged with other fragments. I can imagine a name like `OpenApiFragmentProvider` if we want to follow `Provider<T>` API, otherwise `OpenApiFragmentContributor` would be my preferred name.

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/security.spec.extension.ts
jannyHou: > Have you considered a different approach, where the enhancer provides new data to be added to the spec, and LB runtime ensures the data is cleverly merged together?

Good question, I understand your concern and the risk to give enhancers the right to modify the entire spec, and actually for the original issue(add endpoint level security spec for auth automatically) we are addressing, WE CONTROL THE MERGING LOGIC is easier than my current approach in this PR.

While my concern is...for a fragment contributed by an arbitrary extension, how do we decide the merging logic?

jannyHou: Add endpoint level security spec as follow:
```
paths:
  /something:
    get:
      security:
        - basicAuth:[]
```
- Developer decorate endpoint as `@authenticate('basicAuth')`
- Basic auth strategy component provide the security spec fragment and register the it as an OASEnhancer extension.(here the extension only provides the spec, but don't have merge logic)
- In the rest module, we first generate the entire `paths` spec for the controller methods with auth metadata included, e.g. `x-auth-strategy-name: 'basicAuth'`. Then iterate through all the endpoints, retrieve the security spec fragment based on their auth metadata by calling `OASenhancer.getSpecByName('basicAuth')`, merge fragment into the path spec.

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
jannyHou: Good question * 2, I share your concern when implementing the extension point but haven't figured out a concrete plan, maybe use a combination of  **group** and **alphabet order**? like how we load observers group by group in https://loopback.io/doc/en/lb4/Life-cycle.html#observer-groups

jannyHou: Moreover, if extension only contributes spec, and the app decides ALL the merge logic, then the order won't be a problem...

So the design here is related to our discussion in https://github.com/strongloop/loopback-next/pull/4258#discussion_r355691708.

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/security.spec.extension.ts
bajtos: > While my concern is...for a fragment contributed by an arbitrary extension, how do we decide the merging logic?

I am proposing to start with the simplest option that works for security schemas. We can look into more flexible & powerful merging strategies later, as we learn about new different use cases.

I think the initial merging strategy can do the following:
- If the value is scalar (not object, not an array), then set it as-is.
- If the value is an object, then recursively merge all properties.
- If the value is an array, then append items provided by OASenhancer to the existing array.

A mock-up implementation:

```js
function merge(current, enhancement) {
  const result = _.cloneDeep(current);
  for (key in enhancement) {
    const val = enhancement[key];
    if (Array.isArray(val)) {
      // TODO: verify that current[key] is either `undefined` or an array
      result[key] = [...current[key], ...val];
    } else if (typeof val === 'object') {
      // TODO: verify that current[key] is either `undefined` or an object
      result[key] = merge(current[key] ?? {}, val);
    } else {
      result[key] = val;
    }
  }
  return result;
}
```

Thoughts?

packages/openapi-v3/src/extension-point/spec-contributor.service.ts
bajtos: > maybe use a combination of group and alphabet order? like how we load observers group by group in https://loopback.io/doc/en/lb4/Life-cycle.html#observer-groups

Sounds good, I like the idea of using a mechanism we already have in place elsewhere in the framework. This way our users have less to learn.

I think it's ok to leave custom order out of scope of the initial implementation, as long as the design makes it possible to introduce custom order later.

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/info.spec.extension.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+@bind(asSpecEnhancer)
+export class InfoSpecEnhancer implements OAISpecEnhancer {
+  modifySpec(spec: OpenApiSpec) {
+    spec.info = {
+      title: 'LoopBack Test Application',
+      version: '1.0.1',
+    };
+    return spec;
+  }
```
bajtos: I would prefer the framework to provide a tool for merging OpenAPI spec fragments for us, so that enhancers don't have to (re)implement the logic.

```ts
@bind(asSpecEnhancer)
export class InfoSpecEnhancer implements OAISpecEnhancer {
  modifySpec(spec: OpenApiSpec) {
    return assignSpecFragment(spec, {
      info: {
        title: 'LoopBack Test Application',
        version: '1.0.1',
      },
    });
  }
}
```

packages/openapi-v3/src/__tests__/unit/spec-contributor/fixtures/security.spec.extension.ts
bajtos: Ah, based on our chat, I see that I misunderstood this thread. My comment above is proposing a merging rule for [security schemes](https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#componentsSecuritySchemes) in `/components/security`, not for endpoint-level [security requirements](https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#security-requirement-object).

**When it comes to security requirements at endpoint level, I feel it's important to have a somewhat generic tool provided by `@loopback/rest`, a tool that any custom authentication solution can leverage.**

This will make it easy for any extensions to contribute endpoint-level security requirements. If somebody decides to build an alternative to `@loopback/authentication` extension, then they should have all tools available and should not need to re-implement contributions of endpoint security requirements.

The rest is just implementation details in a sense.

---

One option is to use the spec enhancer as proposed so far.

In https://github.com/strongloop/loopback-next/pull/4258/files#r357211992, I proposed to implement a helper function for merging OpenAPI spec fragments. If we decide to use the concept of OpenAPI spec enhancer as the tool for contributing endpoint security requirements, then I would like `@loopback/rest` to expose a function for setting (or adding) a new security requirement for the given endpoints. 

For example (assuming a hard-coded security spec to keep the example simpler):

```ts
import {assignSecurityRequirement, /*...*/} from '@loopback/rest';

@bind(asSpecEnhancer)
export class BasicAuthSpecEnhancer implements OAISpecEnhancer {
  modifySpec(spec: OpenApiSpec) {
    return assignSecurityRequirement({
      spec, // the original spec
      verb: 'get', // http verb
      path: '/greet',  // http path
      name: 'petstore_auth', // name of the security schema
      scopes: ['read:pets'], // optional, defaults to []
   });
  }
}
```

On the second thought, maybe we don't need any new function and `assignSpecFragment` is actually all we need:

```ts
assignSpecFragment(spec, {
  paths: {
    [endpointPath]: {
      [endpointVerb]: {
        security: {
          petstore_auth: [
            'read:pets'
          ],
         }
       }
     }
  }
});
```

---

Another option is to introduce a new decorator. If you think about how we are defining endpoint metadata now, we have several decorators like `@param` and `@requestBody` that are contributing spec fragments that are later combined into a single operation object. I have an idea to implement a new decorator for defining security requirements.

Example usage:

```ts
import {get, security} from '@loopback/rest'

class GreetController {
  @security('petstore_auth', [
    "write:pets",
    "read:pets",
  ])
  @get('/hello')
   greet() {
     return 'hello world';
  }
}
```

The `@authenticate` decorator can then call `@security` under the hood to contribute the correct security requirement to the OpenAPI spec.

```ts
import {get} from '@loopback/rest';
import {authenticate} from '@loopbakc/authenticate';

class GreetController {
  @authenticate('basicAuth')
  // ^-- will call security('basicAuth') under the hood
  @get('/hello')
   greet() {
     return 'hello world';
  }
}
```

Anyhow. I don't know if the `@security` decorator is a viable solution that can support our current auth/auth design.


I am posting these ideas just for inspiration, please don't hold to them.

jannyHou: @bajtos Good summary, based on all our discussions, I changed the implementation and design to be:

- I still kept the `modifySpec` function in the interface to give people the max flexibility to modify the original spec:
- As we chatted in the meeting, I introduced a default spec merge function(merge a patch spec into the original spec), and exported it for extension developers to use in the `modifySpec` function.
  - It leverages an existing `json-merge-patch` module to merge the spec. I chose it because [`ajv` module](https://github.com/epoberezkin/ajv-merge-patch) also uses it. More details can be found in the API documentation.
- I added two more util functions to the service `getEnhancerByName` and `applyEnhancerByName`, so people can apply a single enhancer in case they need, or doesn't  want to **automatically** apply all the enhancers in `generateSpec`.
- For the endpoint level security spec, I tend to agree with adding a new decorator `@security()`. Reasons are same as what you explained:
  - "When it comes to security requirements at endpoint level, I feel it's important to have a somewhat generic tool provided by @loopback/rest, a tool that any custom authentication solution can leverage."

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it('applyEnhancerByName - returns the merged spec after applying a given enhancer', async () => {
+    const EXPECTED_SPEC = {
+      openapi: '3.0.0',
+      // info object is updated by the info enhancer
+      info: {title: 'LoopBack Test Application', version: '1.0.1'},
+      paths: {},
+    };
+    const mergedSpec = await specService.applyEnhancerByName('info');
```
agnes512: Why `SECURITY_SCHEME_SPEC` is not part of the expected result? Does the enhancer only update the `info` part? Cause I thought (this line)[https://github.com/strongloop/loopback-next/pull/4258/files#diff-dd8c9f9bac61f1da3ef811d81198fb9aR40] merges the security spec to it.

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
```diff
@@ -0,0 +1,121 @@
...
+   * @param name The name of the enhancer you want to apply
+   */
+  async applyEnhancerByName(name: string): Promise<OpenApiSpec> {
+    const enhancer = await this.getEnhancerByName(name);
+    if (enhancer) this._spec = enhancer.modifySpec(this._spec);
+    return this._spec;
+  }
+
+  /**
+   * Generate OpenAPI spec by apply ALL registered enhancers
```
emonddr: by `applying` ALL

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
```diff
@@ -0,0 +1,121 @@
...
+   * TBD: load enhancers by group names
+   */
+  async generateSpec(options = {}): Promise<OpenApiSpec> {
+    const enhancers = await this.getEnhancers();
+    if (_.isEmpty(enhancers)) return this._spec;
+    for (const e of enhancers) {
+      this._spec = e.modifySpec(this._spec);
```
emonddr: Why don't we call `applyEnhancerByName` here?

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+    debug(`Spec enhancer service, generated spec: ${inspect(this._spec)}`);
+    return this._spec;
+  }
+}
+
+/**
+ * The default merge function to patch the current OpenAPI spec.
+ * It leverages module `json-merge-patch`'s merge API to merge two json object.
```
emonddr: two json objects.

packages/openapi-v3/src/enhancers/types.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Name/id of the OAS enhancer extension point
+ */
+export const OAS_ENHANCER_EXTENSION_POINT_NAME = 'oas-enhancer';
+
+/**
+ * A binding template for spec contributor extensions
+ */
+export const asSpecEnhancer: BindingTemplate = binding => {
+  extensionFor(OAS_ENHANCER_EXTENSION_POINT_NAME)(binding);
+  // is it ok to have a different namespace than the extension point name?
```
emonddr: Good question. Raymond originally used the same value in both places when he created
this function to register authentication strategies.  https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/types.ts#L91 .

packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/security.spec.extension.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A spec enhancer to add bearer token OpenAPI security entry to
+ * `spec.component.securitySchemes`
+ */
+@bind(asSpecEnhancer)
+export class SecuritySpecEnhancer implements OASEnhancer {
+  name = 'security';
+
+  modifySpec(spec: OpenApiSpec): OpenApiSpec {
```
emonddr: Security stuff can be added at the top level, or for operation level. Does this method take this into account?

emonddr: e.g. described here : https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#specifying-the-security-settings-in-the-openapi-specification

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(mergedSpec).to.eql(EXPECTED_SPEC);
+  });
+
+  function givenAppWithSpecComponent() {
```
emonddr: by component here, do we mean a LB4 component, or a openApiSpec document with a 'component' section?

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+      info: {title: 'LoopBack Application', version: '1.0.0'},
+      // setter adds paths spec based on the default spec
+      paths: getSamplePathSpec(),
+    };
+    const PATHS_SPEC = getSamplePathSpec();
+    specService.spec = {...specService.spec, paths: PATHS_SPEC};
+    expect(specService.spec).to.eql(EXPECTED_SPEC);
+  });
+
+  it('generateSpec - loads and create spec for ALL registered extensions', async () => {
```
emonddr: loads and creates spec

OR

loads and applies spec  

Do we want to say create or apply? You have methods like ApplyEnhancer etc

emonddr:
@jannyHou, please update the Checklist section by checking the relevant checkboxes

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
emonddr: > Why `SECURITY_SCHEME_SPEC` is not part of the expected result? Does the enhancer only update the `info` part? Cause I thought (this line)[https://github.com/[/pull/4258](https://github.com/strongloop/loopback-next/pull/4258)/files#diff-dd8c9f9bac61f1da3ef811d81198fb9aR40] merges the security spec to it.

The service calls all contributors...including SecuritySpecEnhancer which applies the spec you are refering to.

dhmlau:
@jannyHou, are you planning to add docs as a follow up PR of the task? Thanks.

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
jannyHou: @agnes512 @emonddr By calling `applyEnhancerByName` you ONLY APPLY A SINGLE ENHANCER, in this case is `info` enhancer, that's why the security spec won't be included.

Let me rename the `generateSpec` function to sth like `applyAllEnhancers` to be more clear.

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
jannyHou: `applyEnhancerByName()` calls `getEnhancerByName()`, which loads all enhancers again by calling `getEnhancers()`, see its implementation:

```ts
/**
   * Find an enhancer by its name
   * @param name The name of the enhancer you want to find
   */
  async getEnhancerByName(name: string): Promise<OASEnhancer | undefined> {
    // Get the latest list of enhancers
    const enhancers = await this.getEnhancers();
    return enhancers.find(e => e.name === name);
  }
```

And `applyEnhancerByName` essentially calls enhancer's `modifySpec` function.

packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/security.spec.extension.ts
jannyHou: The operation level spec will be addressed in a new story, see all the discussion in https://github.com/strongloop/loopback-next/pull/4258#discussion_r358465354, me and @bajtos tend to introduce `@security` to provide the spec. So the extension in this PR only focus on merging the security scheme into components.

packages/openapi-v3/src/__tests__/unit/enhancers/spec-enhancer.extensions.unit.ts
jannyHou: good catch, no component any more.

raymondfeng:
@jannyHou I like the idea to separate `modifySpec` and `mergeSpec`. The `mergeSpec` is a helper but we should allow an enhancer to modify the spec including removing some part of the spec, which will be very difficult to achieve via merging. For the 1st version, `modifySpec` is the simplest form that doesn't require the extension point to be magic in merging.

jannyHou:
> are you planning to add docs as a follow up PR of the task? Thanks.

Yep adding the doc :)

>  please update the Checklist section by checking the relevant checkboxes

Will update after adding doc :)

> I like the idea to separate modifySpec and mergeSpec.
> For the 1st version, modifySpec is the simplest form that doesn't require the extension point to be magic in merging.

Great 👍 I believe now team agrees on the plan.

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
emonddr: @jannyHou , please make the change I recommend above to
the existing comment:
```
* Generate OpenAPI spec by apply ALL registered enhancers
```

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
emonddr: I realize that. I just though we could make use of applyEnhancerByName from within the loop, but just noticed that the latter method performs a get on the enhancers and we won't want to be doing that twice. :) Looks good. disregard my question. :)

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+permalink: /doc/en/lb4/Extending-OpenAPI-specification.html
+---
+
+## OpenAPI Specification Enhancer
+
+The APIs in a LoopBack `RestApplication` are described by the
+[OpenAPI Specification(short for OAS)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md).
+An application's OAS are mainly generated from
+[controllers](https://loopback.io/doc/en/lb4/Controllers.html) and their
+members' metadata, but we would also like to contribute specifications from
```
agnes512: maybe `, but` -> `Besides this, we ..`
or
`Besides generating app's OAS from..., we would also`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+other places. Therefore an extension point `OASEnhancerService` is created to
```
agnes512: I think  `Therefore` can be removed or `Therefore, an extension..`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+allow registered extensions to provide their OAS fragments and modify a rest
+application's specification.
+
+_Read about the extension point/extension patten in
+[documentation](https://loopback.io/doc/en/lb4/Extension-point-and-extensions.html)_
+
+## Adding a New OAS Enhancer
+
+Interface `OASEnhancer` is created in `@loopback/openapi-v3` to describe the
+specification enhancers. A typical OAS enhancer class should have a `name` as
```
agnes512: Suggestion:
`A typical OAS enhancer class should have a 'name' as string, and a ...` ->
`A typical OAS enhancer class should have a string type 'name' field ( no comma) and a function 'modifySpec()' that can modify the current specification`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+To apply the basic merging, we provide a default helper function called
+`defaultMergeFn` that leverages
+[`json-merge-patch`](https://github.com/pierreinglebert/json-merge-patch) to
+merge two json objects. You can find its usage in the
+[previous section](#adding-a-new-oas-enhancer)
+
+### Registering an Enhancer
+
+After decorating your enhancer properly with `@bind(asSpecEnhancer)`, you can
+bind it to your application as follow:
```
agnes512: `as follows`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+    super();
+    this.add(createBindingFromClass(InfoSpecEnhancer));
+  }
+}
+```
+
+## OAS Enhancer Service as Extension Point
+
+The OAS enhancer extension point is created in package `@loopback/openapi-v3`.
+It organizes the registered OAS enhancers, and as the first implementation, it
```
agnes512: What does `as the first implementation` mean?

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  // define a function to return a spec service by the same key
+  getSpecService() {
+    return this.get(OAS_ENHANCER_SERVICE);
+  }
+}
+```
+
+### Apply Registered Enhancers
```
agnes512: `Applying` to match with other titles?

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+title: 'Extending OpenAPI Specification'
+keywords: LoopBack 4.0, LoopBack 4
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Extending-OpenAPI-specification.html
+---
+
+## OpenAPI Specification Enhancer
+
+The APIs in a LoopBack `RestApplication` are described by the
+[OpenAPI Specification(short for OAS)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md).
```
nabdelgadir: nitpick: missing space before opening bracket `Specification (`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+An application's OAS are mainly generated from
```
nabdelgadir: nitpick: `are` -> `is`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+[controllers](https://loopback.io/doc/en/lb4/Controllers.html) and their
+members' metadata, but we would also like to contribute specifications from
+other places. Therefore an extension point `OASEnhancerService` is created to
+allow registered extensions to provide their OAS fragments and modify a rest
+application's specification.
+
+_Read about the extension point/extension patten in
```
nabdelgadir: typo: `patten` -> `pattern`
nitpick: add `the` after `in`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+[documentation](https://loopback.io/doc/en/lb4/Extension-point-and-extensions.html)_
```
nabdelgadir: suggestion: https://loopback.io/doc/en/lb4/Extension-point-and-extensions.html -> `Extension-point-and-extensions.md`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Adding a New OAS Enhancer
+
+Interface `OASEnhancer` is created in `@loopback/openapi-v3` to describe the
+specification enhancers. A typical OAS enhancer class should have a `name` as
+string, and a function `modifySpec()` to modify the current specification.
+
+For example, to modify the `info` field of an OAS, you can create an
+`InfoSpecEnhancer` that implements interface `OASEnhancer` as follow:
```
nabdelgadir: nitpick: `as follow` -> `as follows`

docs/site/Extending-OpenAPI-specification.md
nabdelgadir: nitpick: `as follow` -> `as follows`

docs/site/Extending-OpenAPI-specification.md
nabdelgadir: Question, what do you mean by `as the first implementation`?

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Apply Registered Enhancers
+
+To automatically apply all the registered enhancers, call `applyAllEnhancers`:
+
+```ts
+await app.getSpecService.applyAllEnhancers();
+```
+
+_In the future we will support apply enhancers by a custom sequence. The
```
nabdelgadir: nitpick: `apply` -> `applying`

docs/site/Extending-OpenAPI-specification.md
```diff
@@ -0,0 +1,154 @@
...
+sequence will be determined by a combination of group names and the alphabet
```
nabdelgadir: nitpick: `and the alphabet order` -> `and alphabetical order`
Can you clarify what you mean by `a combination of group names`?

packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/application.ts
```diff
@@ -0,0 +1,23 @@
+import {Application, createBindingFromClass} from '@loopback/core';
```
nabdelgadir: missing copyright header

packages/openapi-v3/src/enhancers/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './keys';
```
nabdelgadir: missing copyright header

docs/site/Extending-OpenAPI-specification.md
jannyHou: changing it to be `Therefore, `

docs/site/Extending-OpenAPI-specification.md
jannyHou: we will have follow-up stories to improve this feature.

docs/site/Extending-OpenAPI-specification.md
jannyHou: good catch

docs/site/Extending-OpenAPI-specification.md
jannyHou: @nabdelgadir See our discussion in https://github.com/strongloop/loopback-next/pull/4258#discussion_r355542472, same approach as the applying the observers :)

docs/site/Extending-OpenAPI-specification.md
jannyHou: see https://github.com/strongloop/loopback-next/pull/4258#discussion_r359922317

seems people are confused here, removed it.

packages/openapi-v3/src/enhancers/spec-enhancer.service.ts
```diff
@@ -0,0 +1,121 @@
...
+ * It leverages module `json-merge-patch`'s merge API to merge two json objects.
+ * It returns a new merged object without modifying the original one.
+ *
+ * A list of merging rules can be found in test file:
+ * https://github.com/pierreinglebert/json-merge-patch/blob/master/test/lib/merge.js
+ *
+ * @param currentSpec The original spec
+ * @param patchSpec The patch spec to be merged into the original spec
+ */
+export function defaultMergeFn(
```
raymondfeng: We should rename it to something like `patchOpenApiSpec` or `mergeOpenApiSpec`. For a consumer, it is a bit strange to call `defaultMergeFn(...)`.

jannyHou: agreed, `mergeOpenApiSpec` sounds better to me, will rename it.

packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/security.spec.extension.ts
```diff
@@ -0,0 +1,44 @@
...
+import {asSpecEnhancer, OASEnhancer} from '../../../../enhancers/types';
+import {OpenApiSpec} from '../../../../types';
+const debug = debugModule('loopback:openapi:spec-enhancer');
+
+export type SecuritySchemeObjects = {
+  [securityScheme: string]: SecuritySchemeObject | ReferenceObject;
+};
+
+export const SECURITY_SCHEME_SPEC: SecuritySchemeObjects = {
+  bearerAuth: {
```
dougal83: @emonddr Nitpick: Would it make sense to use `jwt` specifically rather than `bearerAuth` to be more descriptive?
````ts
      "jwt": {
        "type": "http",
        "scheme": "bearer",
        "bearerFormat": "jwt"
      }
````
It would make the openapi security object more informative IMO as JWT's are nearly always used as bearer tokens where are the former can vary.
````
  "security": [
    {
      "jwt": []
    }
  ],
````

Issue #3453, opened 2019-07-26
Include related models with a custom scope


bajtos:
Besides specifying the relation name to include, it's also possible to specify additional `scope` constraints:

- custom `order`, `limit`, `skip` and `fields`
- additional `scope.where` constraint

For example, the following filter will include only the first active product:

```js
filter.include = [{
  relation: 'products',
  scope: {
    where: {active: true},
    limit: 1
  }
}
```

The initial version of Inclusion does not support custom scopes. In this story, we should improve `findByForeignKeys` function to support additional scope constraints.

LB3 test suite: [loopback-datasource-juggler/test/include.test.js#L247-L253)](https://github.com/strongloop/loopback-datasource-juggler/blob/f0a6bd146b7ef2f987fd974ffdb5906cf6a584db/test/include.test.js#L247-L253)

See also https://github.com/strongloop/loopback-next/pull/3387

## Acceptance criteria
- [ ] support nested include relations.
- [ ] add new mocha test cases to exercise these scenarios
- [ ] update docs (add nexted query examples)

luncht1me:
Was trying to do a 
```ts
include: [
{ relation: 'relatedModel', scope: { include[{ relation: 'relatedRelatedModel' }] } },
]
```
but got shut down by 'scope not supported' error. I see it's written in the type definitions though, is this just a bad package version on my part perhaps? Or is scope within a relation actually not hashed out completely yet?

kundk-ckret:
I have the same issue and need to query nested relations...
Is there no way atm to do this automatically? I have a tree structure and don't want to query for each node seperatly :(

st119848:
i need inclusion in inclusion toooo. @luncht1me

theophane-girard:
Me too ! :)

tipsypastels:
Would really like to see this added.

ericalves:
Is too hard use lb4 without nested include relations.. mostly if you came from lb3..

ericalves:
> Was trying to do a filter query
> 
> ```ts
> include: [
> { relation: 'relatedModel', scope: { include:[{ relation: 'relatedRelatedModel' }] } },
> ]
> ```
> 
> but got shut down by 'scope not supported' error. I see it's written in the type definitions though, is this just a bad package version on my part perhaps? Or is scope within a relation actually not hashed out completely yet?
> 
[...]

Issue #3205, opened 2019-06-21
Simpler Syntax for Inclusion


nabdelgadir:
## Description / Steps to reproduce / Feature proposal

See https://github.com/strongloop/loopback-next/pull/3171/files#r296128644.

Currently to specify inclusion, the following filter syntax is needed: `{include: [{relation: 'todos'}]}` e.g. http://127.0.0.1:3000/todo-lists?filter[include][][relation]=todos

If we try the following call:

```ts
await todoListRepo.findById(todoList.id, {include: ['todos']});
```

The complier complains: `Argument of type '{ include: string[]; }' is not assignable to parameter of type 'Filter<TodoList>'` because the type is currently: `include?: Inclusion<MT>[];` 

As LoopBack 3 supports `{include: ['todos']}`, LoopBack 4 should support this shortcut as well.

## Acceptance Criteria
- [ ] Update [include](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/query.ts#L213) to accept an array of model relation strings so that something like `{include: ['todos']}` is accepted.
- [ ] Test coverage at both acceptance and integration level.
- [ ] Update examples, tutorials and documentation to leverage the new simpler syntax wherever possible

Important: we need to preserve backwards compatiblity, i.e. the current verbose syntax must stay supported too.

bajtos:
Few examples of the syntax supported by LB3:

```ts
userRepo.find({include: ['posts', 'passports']});

userRepo.find({
  include: [
    {relation: 'posts', scope: {where: {title: 'Post A'}}},
    'passports',
  ],
});
```

Issue #2208, opened 2019-01-03
API Explorer ignores filter options where, fields, and order


ewrayjohnson:
I am using the example Todo model with the memory connector.  The GET/ operations show the following example filter object, but "where", "fields", and "order "o are stripped before getting to the controller's find method.   
```
{
  "where": {},
  "fields": {},
  "offset": 0,
  "limit": 0,
  "skip": 0,
  "order": [
    "string"
  ]
}
```
From other posts, this may be a coercion issue, but I would expect this functionality from LB3 to work flawlessly in LB4.  I strongly disagree with the announcement that "...LoopBack 4 GA is Now Ready for Production Use!"  Also, I cannot find ANY documentation on filters in the LB4 documentation.  My test filter is as simple as:
```
{
  "where": {
        "id" : 1
   } 
}
```

## Acceptance Criteria:

- [x] - this is a breaking fix, needs semver major release
- [x] -  change existing param.query.object decorator to support json objects in query params.
including content field in param spec, helps to support both exploded filter conditions like filter[limit] = 0 and json objects as per spike PR https://github.com/strongloop/loopback-next/pull/4141
- [x] - parameter schema cannot have `content` field along with `schema`, `explode`, `type` fields. swagger validation fails (check with https://editor.swagger.io/) if they appear together.
- [x] - loopback currently does not support extracting parameter schema from the `content` field. if we are including schema under `content` and not `schema` field, then it means additional changes in tests and source.
- [x] - add additional unit, acceptance and regression tests
- [x] - add documentation for filter conditions and above changes
- [x] - Close related stories

bajtos:
> "where", "fields", and "order "o are stripped before getting to the controller's find method.

I assume this is happening when making the request from the API Explorer, is that correct?

I am afraid this is a known bug/limitation of swagger-ui and swagger-js, see the discussion in https://github.com/strongloop/loopback-next/pull/1679 and the tracking issue https://github.com/swagger-api/swagger-js/issues/1385.

You can see this yourself in API Explorer UI. When you send the request, the UI shows the full URL, request headers and response headers. There you can see  that the request URL created by API Explorer (swagger-js) does not contain the filter properties.

Unfortunately swagger-ui does not warn the user about this problem, which makes the entire experience less than ideal.

Having said that, the actual implementation in LoopBack4 does support where/fields/order, you need to create the request URL manually or use a different OpenAPI client.

[...]

PR #4263, opened 2019-12-05
feat(repository): enable inclusion with custom scope

agnes512:
Implements https://github.com/strongloop/loopback-next/issues/3453

This will allow using custom scope in inclusion. Example:
```
Order: [{id: 1, customerId: 1, name: "to be excluded"}, {id: 2, customerId: 1, name: "to be included"}]
result = await customerRepo.find({
     include: [
       {
         relation: 'orders',
         scope: {where: {id: 2}, include: [{relation: 'customer'}]},
       },
     ],
   });
```
the result should be:
```
[{ id: 1, 
   name: 'IBM', 
   orders:[
      {id: 2, 
       customerId: 1, 
       name: "to be included",
       customer: {id: 1, name: 'IBM'}
      }
   ]
 }]
```

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -43,6 +43,8 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+### Relation Metadata
+
 The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
 decorator. The decorator takes in a function resolving the target model class
 constructor and designates the relation type. It also calls `property()` to
@@ -356,7 +358,7 @@ or use APIs with controllers:
 GET http://localhost:3000/orders?filter[include][][relation]=customer
 ```
 
-### Enable/disable the inclusion resolvers:
+### Enable/disable the inclusion resolvers
 
 - Base repository classes have a public property `inclusionResolvers`, which
   maintains a map containing inclusion resolvers for each relation.
@@ -442,17 +444,22 @@ export class OrderRepository extends DefaultCrudRepository {
   ];
   ```
 
-  Here is a diagram to make this more intuitive:
+{% include note.html content="The query syntax is a slightly different from LB3. We are also thinking about simplifying the query syntax. Check our GitHub issue for more information: [Simpler Syntax for Inclusion](https://github.com/strongloop/loopback-next/issues/3205)" %}
 
-  ![Graph](./imgs/belongsTo-relation-graph.png)
+Here is a diagram to make this more intuitive:
+
+![Graph](./imgs/belongsTo-relation-graph.png)
 
 - You can delete a relation from `inclusionResolvers` to disable the inclusion
   for a certain relation. e.g
   `orderRepository.inclusionResolvers.delete('customer')`
 
-{% include note.html content="
-Inclusion with custom scope:
-Besides specifying the relation name to include, it's also possible to specify additional scope constraints.
-However, this feature is not supported yet. Check our GitHub issue for more information:
-[Include related models with a custom scope](https://github.com/strongloop/loopback-next/issues/3453).
+### Query multiple relations
+
+It is possible to query several relations or nested include relations with
+custom scope once you have the inclusion resolver of each relation set up.
+Check[HasMany - Query multiple relations](HasMany-relation.md#query-multiple-relations)
+for the usage and examples.
+
+{% include important.html content="There are some limitations of inclusion:. <br/><br/> We don’t support recursive inclusion of related models. Related GH issue: [Recursive inclusion of related models](https://github.com/strongloop/loopback-next/issues/3454). <br/><br/> It doesn’t split numbers of queries. Related GH issue: [Support inq splitting](https://github.com/strongloop/loopback-next/issues/3444). <br/><br/> It might not work well with ObjectId of MongoDB. Related GH issue: [Spike: robust handling of ObjectID type for MongoDB](https://github.com/strongloop/loopback-next/issues/3456).
 " %}
--- a/docs/site/HasMany-relation.md
+++ b/docs/site/HasMany-relation.md
@@ -449,7 +449,7 @@ or use APIs with controllers:
 GET http://localhost:3000/customers?filter[include][][relation]=orders
 ```
 
-### Enable/disable the inclusion resolvers:
+### Enable/disable the inclusion resolvers
 
 - Base repository classes have a public property `inclusionResolvers`, which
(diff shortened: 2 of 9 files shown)


packages/repository/src/relations/relation.helpers.ts
```diff
@@ -60,9 +54,26 @@ export async function findByForeignKeys<
...
+/** A helper that combines where clauses for the scope filter
+ *
+ * @param whereClause a new added where clause
+ * @param scope the passed in scope filter
+ */
+async function combineFilters<Target extends Entity>(
+  whereClause: Where<Target>,
+  scope: Filter<Target>,
+): Promise<Filter<Target>> {
+  scope.where = Object.assign(whereClause, scope.where);
```
bajtos: I believe we have utility functions to combine multiple where clauses safely, to ensure operators like `and` and `or` are handled correctly. @raymondfeng do you remember where to find these helpers? I think it's something like `FilterBuilder` and `WhereBuilder`.

docs/site/HasMany-relation.md
```diff
@@ -529,9 +531,107 @@ export class CustomerRepository extends DefaultCrudRepository {
...
+    },
+  ],
+});
+```
+
+Equivalently, with url, you can do:
+
+```
+// need to fix this
+ GET http://localhost:3000/customers?filter[include][0][relation]=orders&filter[include][0][scope][0]filter[where][name]=ToysRUs
```
agnes512: This url filters`[name]=ToysRUs` on `orders`. But I don't know what the syntax to query nested inclusion is. 
I tried
```
http://localhost:3000/customers?filter[include][0][relation]=orders&filter[include][0][scope]filter[include][0][relation]= manufacturers
```
but it doesn't work. I need some help here, thanks!

packages/repository/src/relations/relation.helpers.ts
agnes512: Applied

docs/site/HasMany-relation.md
bajtos: My recommendation is to use [`qs.stringify`](https://www.npmjs.com/package/qs#stringifying) to convert the filter object to a query string.

Inside `packages/rest`:
```
$ node
> require('qs').stringify({filter: {include:[{relation: 'orders', scope: {include:[{relation: 'manufacturers'}]}}]}})
'filter%5Binclude%5D%5B0%5D%5Brelation%5D=orders&filter%5Binclude%5D%5B0%5D%5Bscope%5D%5Binclude%5D%5B0%5D%5Brelation%5D=manufacturers'
```

url-decoded:

```
$ node
> decodeURI(require('qs').stringify({filter: {include:[{relation: 'orders', scope: {include:[{relation: 'manufacturers'}]}}]}}))
'filter[include][0][relation]=orders&filter[include][0][scope][include][0][relation]=manufacturers'
```

docs/site/HasMany-relation.md
```diff
@@ -521,6 +521,8 @@ export class CustomerRepository extends DefaultCrudRepository {
   ];
   ```
 
+  (Notice that the query syntax is a slightly different from LB3)
```
bajtos: Have you considered using an alert -Note or Tip? See https://loopback.io/doc/en/contrib/alerts.html

Personally, I would also include a link to https://github.com/strongloop/loopback-next/issues/3205 to let people know that we are thinking about simplifying the syntax.

docs/site/HasMany-relation.md
```diff
@@ -529,9 +531,107 @@ export class CustomerRepository extends DefaultCrudRepository {
...
-" %}
+### More use cases - query multiple relations:
+
+It is possible to query several relations or nested include relations with
+custom scope. Once you have the inclusion resolver of each relation set up, the
+following queries would allow you traverse data differently:
+
+In our example, we have relations:
+
+- `Customer` _belongsTo_ an `Address` - denoted as `address`.
```
bajtos: I find this relation weird, it's usually an address that belongs to a customer. Have you considered using `HasOne` or `HasMany` relation instead?

docs/site/HasMany-relation.md
bajtos: Personally, I would show users how to encode the filter via JSON:

```js
encodeURIComponent(JSON.stringify(filter))
```

There is a subtle difference in the output of `encodeURI` and `encodeURIComponent`, one of them is not correctly encoding special characters that must be encoded in query string values. I hope I remember it correctly that `encodeURIComponent` is the function to use.

agnes512: I still couldn't find a good way to encode it.
- `qs.stringify` and `encodeURIComponent` give different result.

- I found something weird, for filter
```
{filter: {include:[{relation: 'orders', scope: {include:[{relation: 'manufacturers'}]}}]}}
```
The outer relation is encoded to `{relation:"outer"}` as expected. But the inner one is not correct: `{[relation: inner]}`, which is causing invalid filter problems. I wonder why...

Even I would like to leave the encode part to users, I still think it's better to have at least one simple and working nested inclusion syntax as an example in docs.

docs/site/BelongsTo-relation.md
```diff
@@ -442,17 +444,144 @@ export class OrderRepository extends DefaultCrudRepository {
...
 - You can delete a relation from `inclusionResolvers` to disable the inclusion
   for a certain relation. e.g
   `orderRepository.inclusionResolvers.delete('customer')`
 
-{% include note.html content="
-Inclusion with custom scope:
-Besides specifying the relation name to include, it's also possible to specify additional scope constraints.
-However, this feature is not supported yet. Check our GitHub issue for more information:
-[Include related models with a custom scope](https://github.com/strongloop/loopback-next/issues/3453).
+### More use cases - query multiple relations:
```
dhmlau: Since we're duplicating the contents for the 3 relations pages, should we put it at the top level to avoid duplication?

agnes512: It doesn't make sense to me if we put it at `Relation.md`. How about I link this part from `belongsTo` and `hasOne` to `hasMany`?

docs/site/HasMany-relation.md
```diff
@@ -521,17 +521,145 @@ export class CustomerRepository extends DefaultCrudRepository {
...
+      relation: 'orders',
+      scope: {
+        include: [{relation: 'manufacturers'}],
+      },
+    },
+  ],
+});
+```
+
+<!-- FIXME: the url isn't being converted to JSON correctly. Add an example url once it's fixed
```
jannyHou: Maybe we can wait until https://github.com/strongloop/loopback-next/issues/2208 finished and see what's the generated url that explorer gives us, then update the example here.

agnes512: Sounds good

PR #4268, opened 2019-12-06
chore: update package-lock.json files

hacksparrow:
CI is complaining about invalid package-lock.json files.

### DIFF
--- a/packages/repository/package-lock.json
+++ b/packages/repository/package-lock.json
@@ -48,12 +48,9 @@
 			}
 		},
 		"async": {
-			"version": "2.6.3",
-			"resolved": "https://registry.npmjs.org/async/-/async-2.6.3.tgz",
-			"integrity": "sha512-zflvls11DCy+dQWzTW2dzuilv8Z5X/pjfmZOWba6TNIVDm+2UDaJmXSOXlasHKfNBs8oo3M0aT50fDEWfKZjXg==",
-			"requires": {
-				"lodash": "^4.17.14"
-			}
+			"version": "3.1.0",
+			"resolved": "https://registry.npmjs.org/async/-/async-3.1.0.tgz",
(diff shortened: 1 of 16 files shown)


packages/repository/package-lock.json
```diff
@@ -300,36 +297,14 @@
...
-						"yamljs": "^0.3.0"
-					}
-				}
 			}
 		},
 		"loopback-datasource-juggler": {
-			"version": "4.16.0",
-			"resolved": "https://registry.npmjs.org/loopback-datasource-juggler/-/loopback-datasource-juggler-4.16.0.tgz",
-			"integrity": "sha512-VV0GVNzsQFTD+iKtQe39Ojzjph3onbxRMXl7qq7IP2NJN4tgJb/WsB586Sg6OwUdcw66wvwqo34V+dNIQRw0dw==",
+			"version": "4.17.0",
```
bajtos: Upgrade to juggler v17 is handled by https://github.com/strongloop/loopback-next/pull/4267. Let's land that PR first and then re-run the update of package locks.

bajtos:
FYI: RenovateBot is configured to update deep dependencies in our lock files every Monday, see https://github.com/strongloop/loopback-next/blob/fdd75ecd1a39e9c222b3929d69fda0dfc75c53ee/.renovaterc.json#L22-L25 

We can configure RenovateBot to update them more often, but that will mean more `chore` PRs.

jannyHou:
Personal opinion: I feel a weekly update of dependencies is reasonable, unless we have security alert, which doesn't occur frequently and could be fixed manually, like the PR I merged last time.

(That's also why I didn't create another to update the schedule, I think it's good enough.)

hacksparrow:
Closing in favor of RenovateBot.

PR #4269, opened 2019-12-06
[Very Small] Update API Documentation URL

erikburt:
This is such an insignificant change I figure I wouldn't need to jump through the hoops needed for submitting an actual PR.

Updated a link in the Readme because when I tried to access it would bring me to a page giving a 404 and saying `Cannot GET /@loopback%2fdocs/apidocs.html`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ by using it for your next project. Start by having a look at
 [Getting Started](https://loopback.io/doc/en/lb4/Getting-started.html).
 
 Check the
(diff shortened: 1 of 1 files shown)


raymondfeng:
@erikburt Thank you for the patch. Please revise your PR to fix build errors - https://github.com/strongloop/loopback-next/pull/4269/checks?check_run_id=336946508. 

See https://loopback.io/doc/en/lb4/submitting_a_pr.html for instructions. 

BTW, please don't erase the `Checklist` from PR template. It should have a few checkboxes similar as https://github.com/strongloop/loopback-next/pull/4091

Issue #3602, opened 2019-08-25
Preserve custom type of auto-generated id property


frbuceta:
## Steps to reproduce

1. Create a model with an id property with the option `generated: true`
![Captura de pantalla 2019-08-25 a las 16 25 52](https://user-images.githubusercontent.com/13822438/63651489-89d38880-c755-11e9-918e-2788a95dc5c6.png)
2. You make a find and the answer is ... uuid: NaN or id: NaN
![Captura de pantalla 2019-08-25 a las 16 26 03](https://user-images.githubusercontent.com/13822438/63651494-8e983c80-c755-11e9-8e5e-a3ba08d7218b.png)

## Current Behavior

As I describe above, when the model is in `generated: true` it has to return the normal id, not NaN

## Expected Behavior

![Captura de pantalla 2019-08-25 a las 16 26 47](https://user-images.githubusercontent.com/13822438/63651526-dd45d680-c755-11e9-8e2e-30d0239d4a5b.png)

## Link to reproduction sandbox

When I put generated: false ...
![Captura de pantalla 2019-08-25 a las 16 26 37](https://user-images.githubusercontent.com/13822438/63651536-f77fb480-c755-11e9-8847-267abc0ae9ef.png)
![Captura de pantalla 2019-08-25 a las 16 26 47](https://user-images.githubusercontent.com/13822438/63651526-dd45d680-c755-11e9-8e2e-30d0239d4a5b.png)

## Additional information

OS: MacOS 10.14.6
Node: v12.9.0
Npm: v6.11.2
Loopback: latest -> 
```
@loopback/cli version: 1.21.4

@loopback/* dependencies:
  - @loopback/authentication: ^2.1.11
  - @loopback/boot: ^1.5.3
  - @loopback/build: ^2.0.8
  - @loopback/context: ^1.21.4
  - @loopback/core: ^1.9.3
  - @loopback/metadata: ^1.2.10
  - @loopback/openapi-spec-builder: ^1.2.10
  - @loopback/openapi-v3: ^1.9.4
  - @loopback/repository-json-schema: ^1.9.5
  - @loopback/repository: ^1.12.0
  - @loopback/rest: ^1.17.0
  - @loopback/testlab: ^1.7.4
  - @loopback/docs: ^1.29.3
  - @loopback/example-hello-world: ^1.2.11
  - @loopback/example-log-extension: ^1.2.11
  - @loopback/example-rpc-server: ^1.2.11
[...]

Issue #2398, opened 2019-02-14
Column is always integer when running `npm run migrate`


dhmlau:
## Description / Steps to reproduce / Feature proposal
I have a model property of type `string`.  After running `npm run migrate`, the table gets created, and the type shows `integer`.   
- I'm using postgresql (not sure if it makes a difference)
- my model ts looks like:
```
import {Entity, model, property} from '@loopback/repository';

@model()
export class Customer extends Entity {
  @property({
    type: 'string',
    id: true,
    generated: true,
  })
  custid: string;

  @property({
    type: 'string',
    required: true,
  })
  custname: string;

  constructor(data?: Partial<Customer>) {
    super(data);
  }
}

```

Running `\d customer` on postgresql, it shows:
```
testdb=# \d customer
                               Table "public.customer"
  Column  |  Type   | Collation | Nullable |                 Default                  
----------+---------+-----------+----------+------------------------------------------
 custid   | integer |           | not null | nextval('customer_custid_seq'::regclass)
 custname | text    |           | not null | 
Indexes:
    "customer_pkey" PRIMARY KEY, btree (custid)

```

## Current Behavior
`custid` is of type integer on the database

## Expected Behavior
`custid` should be of type `text` instead.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

dhmlau:
FYI - my sample repo is here: https://github.com/dhmlau/loopback4-example-todo-customized

bajtos:
I think the PostgreSQL connector is enforcing the type of the primary key to be a number. You can see that the `custid` column is configured to use a default value computed by `nextval('customer_custid_seq'::regclass)`.

I think the same problem may apply to other SQL connectors too.

While this behavior is intended, I can see how it can be confusing for people building LB4 applications.

agnes512:
Debug note: for `MySQL`, id property can be type `String` by setting 
```
  @property({
    id: true,
    type: "String",  //not case sensitive
    //generated: true,
  })
  id: string;
```
The key is in this case id cannot be generated automatically.

bajtos:
> Need to check how to migrate `uuid` as property type. And do we want the migrated to have `uuid` as type + auto generate

FYI, I think it should be possible to replace `generated: true` with `defaultFn` option set to `uuidv4`, see https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#general-property-properties.

[...]

PR #4270, opened 2019-12-06
Preserve custom type of auto-generated id property

frbuceta:
Implements #3602 

## Checklist

- [x] New tests added or **existing tests modified to cover all changes**

### DIFF
--- a/packages/repository-tests/src/crud/relations/acceptance/multi-relations-inclusion-resolver.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/multi-relations-inclusion-resolver.relation.acceptance.ts
@@ -16,6 +16,8 @@ import {
   withCrudCtx,
 } from '../../../helpers.repository-tests';
 import {
+  Address,
+  AddressRepository,
   Customer,
   CustomerRepository,
   Order,
@@ -36,6 +38,7 @@ export function hasManyInclusionResolverAcceptance(
   );
   function suite() {
     before(deleteAllModelsInDefaultDataSource);
+    let addressRepo: AddressRepository;
     let customerRepo: CustomerRepository;
     let orderRepo: OrderRepository;
 
@@ -43,14 +46,18 @@ export function hasManyInclusionResolverAcceptance(
       withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
         // this helper should create the inclusion resolvers and also
         // register inclusion resolvers for us
-        ({customerRepo, orderRepo} = givenBoundCrudRepositories(
+        ({customerRepo, orderRepo, addressRepo} = givenBoundCrudRepositories(
           ctx.dataSource,
           repositoryClass,
           features,
         ));
         expect(customerRepo.orders.inclusionResolver).to.be.Function();
 
-        await ctx.dataSource.automigrate([Customer.name, Order.name]);
+        await ctx.dataSource.automigrate([
+          Customer.name,
+          Order.name,
+          Address.name,
+        ]);
       }),
     );
 
@@ -63,7 +70,14 @@ export function hasManyInclusionResolverAcceptance(
       const parent = await customerRepo.create({name: 'parent'});
       const customer = await customerRepo.create({
         name: 'customer',
-        parentId: parent.id,
+        //parentId: parent.id,
+      });
+      const address = await addressRepo.create({
+        street: '8200 Warden',
+        city: 'Markham',
+        province: 'On',
+        zipcode: '8200',
+        customerId: parent.id,
       });
       const order = await orderRepo.create({
         description: 'an order',
@@ -71,7 +85,7 @@ export function hasManyInclusionResolverAcceptance(
       });
 
       const result = await customerRepo.find({
-        include: [{relation: 'orders'}, {relation: 'customers'}],
+        include: [{relation: 'orders'}, {relation: 'address'}],
       });
       const expected = [
         {
@@ -84,16 +98,11 @@ export function hasManyInclusionResolverAcceptance(
               shipmentInfo: features.emptyValue,
             },
           ],
-          customers: [
-            {
-              ...customer,
-              parentId: parent.id,
-            },
-          ],
+          address: address,
         },
         {
           ...customer,
-          parentId: parent.id, // doesn't have any related models
+          parentId: features.emptyValue, //parent.id, // doesn't have any related models
         },
       ];
       expect(toJSON(result)).to.deepEqual(toJSON(expected));
--- a/packages/repository/src/model.ts
+++ b/packages/repository/src/model.ts
@@ -106,6 +106,16 @@ export class ModelDefinition {
     const definition = (definitionOrType as PropertyDefinition).type
       ? (definitionOrType as PropertyDefinition)
       : {type: definitionOrType};
+
+    if (
+      definition.id === true &&
+      definition.generated === true &&
+      definition.type !== undefined &&
+      definition.useDefaultIdType === undefined
+    ) {
+      definition.useDefaultIdType = false;
+    }
+
     this.properties[name] = definition;
     return this;
   }
--- a/packages/repository-tests/src/crud/create-retrieve.suite.ts
+++ b/packages/repository-tests/src/crud/create-retrieve.suite.ts
@@ -12,9 +12,9 @@ import {
   property,
 } from '@loopback/repository';
 import {expect, toJSON} from '@loopback/testlab';
-import {MixedIdType} from '../helpers.repository-tests';
 import {
   deleteAllModelsInDefaultDataSource,
+  MixedIdType,
   withCrudCtx,
 } from '../helpers.repository-tests';
 import {
@@ -89,7 +89,7 @@ export function createRetrieveSuite(
       const pens = await repo.create({name: 'Pens', categoryId: 1});
       const pencils = await repo.create({name: 'Pencils', categoryId: 2});
       const products = await findByForeignKeys(repo, 'categoryId', [1]);
-      expect(products).deepEqual([pens]);
+      expect(toJSON(products)).deepEqual(toJSON([pens]));
       expect(products).to.not.containDeep(pencils);
     });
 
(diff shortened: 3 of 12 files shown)


frbuceta:
Can someone help me with the test? It seems not to be happening correctly

derdeka:
> Can someone help me with the test? It seems not to be happening correctly

The changes looks reasonable to me and local unit test works successfully.

TASK=test-repository-mongodb and TASK=test-repository-mysql fail in travis ci.
The `useDefaultIdType=false` seems not to be used by the juggler.
For mysql the following ddl is still created:
```
CREATE TABLE `User` (
`id` TEXT NOT NULL AUTO_INCREMENT PRIMARY KEY,
`name` VARCHAR(512) NULL,
`roles` TEXT NULL,
`address` TEXT NULL
)
```
Type `TEXT` ad `AUTO_INCREMENT` are not a valid combination.

frbuceta:
> > Can someone help me with the test? It seems not to be happening correctly
> 
> The changes looks reasonable to me and local unit test works successfully.
> 
> TASK=test-repository-mongodb and TASK=test-repository-mysql fail in travis ci.
> The `useDefaultIdType=false` seems not to be used by the juggler.
> For mysql the following ddl is still created:
> 
> ```
> CREATE TABLE `User` (
> `id` TEXT NOT NULL AUTO_INCREMENT PRIMARY KEY,
> `name` VARCHAR(512) NULL,
> `roles` TEXT NULL,
> `address` TEXT NULL
> )
> ```
> 
> Type `TEXT` ad `AUTO_INCREMENT` are not a valid combination.

I certainly did not know that I took code from outside the repository to test, thanks for your help.
As always here I am constantly learning

packages/repository/src/model.ts
```diff
@@ -106,6 +106,11 @@ export class ModelDefinition {
     const definition = (definitionOrType as PropertyDefinition).type
       ? (definitionOrType as PropertyDefinition)
       : {type: definitionOrType};
+
+    if (definition.id === true && definition.generated === true) {
```
bajtos: To be safe, I am proposing to extend this check to apply `useDefaultIdType: false` only when an explicit `type` is provided.

This property should set `useDefaultIdType: false`:

```ts
@property({
  type: 'string',
  id: true,
  generated: true
})
id: string;
```

This property should NOT set `useDefaultIdType: false`:

```ts
@property({
  id: true,
  generated: true
})
id: string | number;
```

Thoughts?

bajtos:
Thank you @frbuceta for the pull request, this is great!

> Type `TEXT` ad `AUTO_INCREMENT` are not a valid combination.

Yeah, this is a problem we have recently discovered too, see https://github.com/strongloop/loopback-next/issues/2398. I think we need to fix our connectors to correctly support auto-generated string ids (primary keys) before we can move forward with this pull request.

@agnes512 you were looking in to #2398, what do you think?

bajtos:
> TASK=test-repository-mongodb and TASK=test-repository-mysql fail in travis ci.
The `useDefaultIdType=false` seems not to be used by the juggler.

On the second thought, I think this is a bug in the way how our shared test suite is creating models. 

The way how this test suite is supposed to work, is that the connector specifies what type is used for primary keys (number vs. string) and the test suite honors that type when defining new models.

I suspect the code building `User` model is not taking into account database-specific type and always uses `string`.

AFAICT, `User.id` property is defined here:

https://github.com/strongloop/loopback-next/blob/4995057ad4c938d6751c329632eb44130ac1567a/packages/repository-tests/src/crud/nested-model-properties.suite.ts#L91-L95

Here is the correct definition of the PK property in a different model:

https://github.com/strongloop/loopback-next/blob/4995057ad4c938d6751c329632eb44130ac1567a/packages/repository-tests/src/crud/create-retrieve.suite.ts#L36-L42

I think we need to add `type: features.idType` to `User.id` definition, that should hopefully fix the problem you are observing.

agnes512:
`AUTO_INCREMENT` can only apply to type `INT`. And `generated` along with `useDefaultIdType` are not enough for auto-generate string type attribute. We need to rework at **connector level** to allow so. 

I am working on [Postgres: allow string type attribute to be auto-generated ](https://github.com/strongloop/loopback-connector-postgresql/pull/404). That PR is proposing to use `uuid_generate_v4()` as a default function once the model is set to:
```
  @property({
    id: true, //optional
    type: 'string'
    generated: true,
    useDefaultIdType: false,
    postgresql: {
      dataType: 'uuid',
    },
  })
  id: String;
```
I've tested it locally, this solution works in `repository-tests` with postgresql.

Once it's done, we can rework the MySQL connector with a similar solution. Please share your idea there if you have any :smile:

derdeka:
> That PR is proposing to use `uuid_generate_v4()` as a default function

What if someone wants to have a different generator function? (e.g. different UUID version or database extension, different function name)
I think the ddl generation should respect the defaultFn if present.

agnes512:
> What if someone wants to have a different generator function? (e.g. different UUID version or database extension, different function name)

That's doable. It will be user's responsibility to provide valid extension and defaultFn.
And if the extension and defaultFn are not provided at the same time, it'd use `uuid_generate_v4()`. WDYT?

packages/repository/src/model.ts
frbuceta: Yes, it seems right to me. I just implemented it

frbuceta:
> > TASK=test-repository-mongodb and TASK=test-repository-mysql fail in travis ci.
> > The `useDefaultIdType=false` seems not to be used by the juggler.
> 
> On the second thought, I think this is a bug in the way how our shared test suite is creating models.
> 
> The way how this test suite is supposed to work, is that the connector specifies what type is used for primary keys (number vs. string) and the test suite honors that type when defining new models.
> 
> I suspect the code building `User` model is not taking into account database-specific type and always uses `string`.
> 
> AFAICT, `User.id` property is defined here:
> 
> https://github.com/strongloop/loopback-next/blob/4995057ad4c938d6751c329632eb44130ac1567a/packages/repository-tests/src/crud/nested-model-properties.suite.ts#L91-L95
> 
> Here is the correct definition of the PK property in a different model:
> 
> https://github.com/strongloop/loopback-next/blob/4995057ad4c938d6751c329632eb44130ac1567a/packages/repository-tests/src/crud/create-retrieve.suite.ts#L36-L42
> 
> I think we need to add `type: features.idType` to `User.id` definition, that should hopefully fix the problem you are observing.

I've tried this but it doesn't work. Right now with the only one I have problems with is mongo.

packages/repository/src/model.ts
```diff
@@ -106,6 +106,17 @@ export class ModelDefinition {
     const definition = (definitionOrType as PropertyDefinition).type
       ? (definitionOrType as PropertyDefinition)
       : {type: definitionOrType};
+
+    if (
+      definition.id === true &&
+      definition.generated === true &&
+      definition.type === 'string'
```
bajtos: I am afraid we misunderstood each other. Here is what I meant:

```suggestion
      definition.type !== undefined
```

Either way, please add a new test to verify that `useDefaultIdType` is not set when `definition.type` was not provided (and was not inferred from TypeScript metadata either).

packages/repository/src/model.ts
```diff
@@ -106,6 +106,17 @@ export class ModelDefinition {
...
+    ) {
+      Object.assign(definition, {
```
bajtos: Nitpick (performance micro-optimization):

```ts
definition.useDefaultIdType = false;
```

bajtos:
> > > TASK=test-repository-mongodb and TASK=test-repository-mysql fail in travis ci.
> > 
> > On the second thought, I think this is a bug in the way how our shared test suite is creating models.
> > I think we need to add `type: features.idType` to `User.id` definition, that should hopefully fix the problem you are observing.
>
> I've tried this but it doesn't work. Right now with the only one I have problems with is mongo.

Hmm...

@agnes512 IIRC, you were dealing with similar issues recently, while working on inclusion resolvers. Can you please help @frbuceta to find a way how to make repository tests pass against MongoDB?

packages/repository/src/model.ts
frbuceta: Take note

frbuceta:
Thanks for your help @agnes512

packages/repository/src/model.ts
frbuceta: There is no model in the tests of the type as undefined, how should I generate a model with these features?

packages/repository/src/model.ts
```diff
@@ -106,6 +106,15 @@ export class ModelDefinition {
     const definition = (definitionOrType as PropertyDefinition).type
       ? (definitionOrType as PropertyDefinition)
       : {type: definitionOrType};
+
+    if (
```
agnes512: what if I have:
```ts
@property({
  id: true,
  generated: true,
  useDefaultIdType: true
})
id: string;
```
Would it overwrite the value of `useDefaultIdType` above?

agnes512: I added a check `&&
      definition.useDefaultIdType === undefined`

packages/repository-tests/src/crud/relations/acceptance/multi-relations-inclusion-resolver.relation.acceptance.ts
```diff
@@ -63,15 +70,22 @@ export function hasManyInclusionResolverAcceptance(
...
+        zipcode: '8200',
+        customerId: parent.id,
       });
       const order = await orderRepo.create({
         description: 'an order',
         customerId: parent.id,
       });
 
       const result = await customerRepo.find({
-        include: [{relation: 'orders'}, {relation: 'customers'}],
```
agnes512: While I was fixing the CI failure, I think the flag might somehow break  how we deal with `ObjectId` in MongoDB. Before we were able to traverse related instances with `ObjectId`.  Not sure it has something to do with the type.

packages/repository/src/model.ts
frbuceta: Good approach, I have seen that you have already changed it so I will give this revision as resolved

bajtos:
@agnes512 @frbuceta what's the status of this pull request? I see that the CI is green and the code looks reasonable, even if not all edge cases may be covered by tests. Should we perhaps proceed to land the changes as they are now?

PR #4288, opened 2019-12-11
feat: add title property to ping response schema definition

derdeka:
Add title property to ping response schema definition in order to enable better model
generation by third party tools such as https://github.com/OpenAPITools/openapi-generator .

Without a title property the generator would simply name the ping response as `InlineResponse200`. By adding the title property, the tooling can produce a model named after the object title.

Similar to #4240

## Checklist

### DIFF
--- a/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
+++ b/packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
@@ -10,6 +10,7 @@ const PING_RESPONSE: ResponseObject = {
     'application/json': {
       schema: {
         type: 'object',
+        title: 'PingResponse',
         properties: {
           greeting: {type: 'string'},
           date: {type: 'string'},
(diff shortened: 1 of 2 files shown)


packages/cli/generators/app/templates/src/controllers/ping.controller.ts.ejs
```diff
@@ -10,6 +10,7 @@ const PING_RESPONSE: ResponseObject = {
     'application/json': {
       schema: {
         type: 'object',
+        title: 'loopback.PingResponse',
```
raymondfeng: I guess `PingResponse` should be good enough as it will be part of the application code - not a built-in schema from LoopBack.

derdeka: Ok, now it's `PingResponse`.

dhmlau:
In case you're interested, there is a LoopBack Developer badge for those who made contributions to LoopBack. Please check this out:  https://www.ibm.com/services/learning/ites.wss/zz-en?pageType=badges&id=0de05804-8037-4791-ac83-09829b566d51. Thanks!

PR #4247, opened 2019-12-03
feat(repository): add interface for hasManyThrough

KalleV:
Adds interface for the new relation type "hasManyThrough" to support many-to-many model relations.

This PR is a continuation of #2359 and implements the first change described in https://github.com/strongloop/loopback-next/pull/2359#issuecomment-559719080.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4312, opened 2019-12-16
feat(repository): add interface for hasManyThrough repository

KalleV:
The hasManyThrough repository interface defines the actions that can be performed on a many-to-many model relation.

This PR is a continuation of https://github.com/strongloop/loopback-next/pull/2359 and implements a step from https://github.com/strongloop/loopback-next/pull/2359#issuecomment-559719080.

Relates to: https://github.com/strongloop/loopback-next/pull/4247

## Checklist

### DIFF
--- a/packages/repository/src/relations/has-many/has-many-through.repository.ts
+++ b/packages/repository/src/relations/has-many/has-many-through.repository.ts
@@ -1,0 +1,100 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Count, DataObject, Entity, Filter, Options, Where} from '../..';
+
+/**
+ * CRUD operations for a target repository of a HasManyThrough relation
+ *
+ * EXPERIMENTAL: This interface is not stable and may change in the near future.
+ * Backwards-incompatible changes may be introduced in semver-minor versions.
+ */
+export interface HasManyThroughRepository<
+  Target extends Entity,
+  TargetID,
+  Through extends Entity
+> {
+  /**
+   * Create a target model instance
+   * @param targetModelData - The target model data
+   * @param options - Options for the operation
+   * @returns A promise which resolves to the newly created target model instance
+   */
+  create(
+    targetModelData: DataObject<Target>,
+    options?: Options & {
+      throughData?: DataObject<Through>;
+      throughOptions?: Options;
+    },
+  ): Promise<Target>;
+
+  /**
+   * Find target model instance(s)
+   * @param filter - A filter object for where, order, limit, etc.
+   * @param options - Options for the operation
+   * @returns A promise which resolves with the found target instance(s)
+   */
+  find(
+    filter?: Filter<Target>,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<Target[]>;
+
+  /**
+   * Delete multiple target model instances
+   * @param where - Instances within the where scope are deleted
+   * @param options
+   * @returns A promise which resolves the deleted target model instances
+   */
+  delete(
+    where?: Where<Target>,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<Count>;
+
+  /**
+   * Patch multiple target model instances
+   * @param dataObject - The fields and their new values to patch
+   * @param where - Instances within the where scope are patched
+   * @param options
+   * @returns A promise which resolves the patched target model instances
+   */
+  patch(
+    dataObject: DataObject<Target>,
+    where?: Where<Target>,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<Count>;
+
+  /**
+   * Creates a new many-to-many association to an existing target model instance
+   * @param targetModelId - The target model ID to link
+   * @param options
+   * @returns A promise which resolves to the linked target model instance
+   */
+  link(
+    targetModelId: TargetID,
+    options?: Options & {
+      throughData?: DataObject<Through>;
+      throughOptions?: Options;
+    },
+  ): Promise<Target>;
+
+  /**
+   * Removes an association to an existing target model instance
+   * @param targetModelId - The target model to unlink
+   * @param options
+   * @returns A promise which resolves to null
+   */
+  unlink(
+    targetModelId: TargetID,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<void>;
+}


packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Create a target model instance
+   * @param targetModelData - The target model data
+   * @param throughModelData - The through model data
+   * @param options - Options for the operation
+   * @param throughOptions - Options passed to create through
+   * @returns A promise which resolves to the newly created target model instance
+   */
+  create(
+    targetModelData: DataObject<Target>,
```
derdeka: How to distingush between an existing or non-existing target entity? I think we need more operations in a `HasManyThroughRepository`. In `lb3` there was a link/unlink to create/delete a relation with an existing target entity.

KalleV: Thank you for the review. Would you like to see a method definition for “exists” as well as “link” and “unlink” (for example, https://apidocs.strongloop.com/loopback/#persistedmodel-exists)?

KalleV: @derdeka I added "add", "remove" and "exists" based on https://loopback.io/doc/en/lb3/HasManyThrough-relations.html and https://github.com/strongloop/loopback-datasource-juggler/blob/2f867576b66fe535aa1f82bceaaa680ecd4eec6c/lib/relation-definition.js.

KalleV: Example implementations of "add" and "remove" using the interface definition:

```
async add(
    targetModel: TargetEntity,
    throughModelData: DataObject<ThroughEntity> = {},
    throughOptions?: Options
  ): Promise<TargetEntity> {
    const throughRepository = await this.getThroughRepository();
    const throughConstraint = this.getThroughConstraint(targetModel);

    await throughRepository.create(
      constrainDataObject(
        throughModelData,
        throughConstraint as DataObject<ThroughEntity>
      ),
      throughOptions
    );

    return targetModel;
  }

  async remove(
    targetModel: TargetEntity,
    options?: Options,  // Note: maybe we don't need both options parameters?
    throughOptions?: Options
  ): Promise<void> {
    const throughRepository = await this.getThroughRepository();
    const throughConstraint = this.getThroughConstraint(targetModel);

    await throughRepository.deleteAll(
      constrainWhere(undefined, throughConstraint),
      throughOptions
    );
  }
```

KalleV:
Hi. Are there any changes you would like to see to the interface?

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,102 @@
...
+    throughModelData?: DataObject<Through>,
+    options?: Options,
+    throughOptions?: Options,
+  ): Promise<Target>;
```
bajtos: I find this signature as too easy to misuse. Consider the following call:

```ts
await repo.create({name: 'Miroslav'}, options);
```

The pattern look very similar to what we do with other repository classes (e.g. `HasManyRepository`), but it is actually an error because the second argument is "through data", not "options".

I am proposing to move `throughModelData` and `throughOptions` into `options` argument.

```ts
export interface HasManyThroughRepository<
   Target extends Entity,
   TargetID,
   Through extends Entity
 > {

  create(
    targetModelData: DataObject<Target>,
    options: Options & {throughData?: DataObject<Through>; throughOptions?: Options;},
  ): Promise<Target>;
}
```

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Creates a new many-to-many association to an existing target model instance
+   * @param targetModel - The target model to link
+   * @param throughModelData - Optional through model data
+   * @param options
+   * @param throughOptions
+   * @returns A promise which resolves to the linked target model instance
+   */
+  add(
+    targetModel: Target,
```
bajtos: I think this method should accept the the target id only.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Removes an association to an existing target model instance
+   * @param targetModel - The target model to unlink
+   * @param options
+   * @param throughOptions
+   * @returns A promise which resolves to null
+   */
+  remove(
+    targetModel: Target,
```
bajtos: I think this method should accept the the target id only.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Checks if the given target model instance is associated with the source model.
+   * @param targetModel - The target model to check
+   * @param options
+   * @param throughOptions
+   * @returns A promise which resolves to true when the association exists
+   * and false otherwise.
+   */
+  exists(
+    targetModel: Target,
```
bajtos: I think this method should accept the the target id only.

packages/repository/src/relations/has-many/has-many-through.repository.ts
KalleV: Sure, I added findById and destroyById and renamed the add/remove methods to link/unlink.

KalleV:
@bajtos Thank you for the review. I addressed the comments. Let me know if you would like any further changes.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,136 @@
...
+   * Creates a new many-to-many association to an existing target model instance
+   * @param targetModelId - The target model ID to link
+   * @param options
+   * @returns A promise which resolves to the linked target model instance
+   */
+  link(
+    targetModelId: TargetID,
+    options?: Options & {
+      throughOptions?: Options;
+    },
```
bajtos: IMO, `link` should accept `throughData` too, so that users can provide additional fields for the record linking the two models together.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,136 @@
...
+  ): Promise<Target[]>;
+
+  /**
+   * Find a related model by ID
+   * @param targetModelId - The target model ID to find
+   * @param options - Options for the operation
+   * @returns A promise which resolves with the found target instance
+   */
+  findById(
```
bajtos: In a nutshell, we are defining the following methods:

```
create(targetModelData, options);
find(filter, options);
findById(targetModelId, options);
delete(where, options);
deleteById(targetModelId: TargetID, options);
patch(dataObject, where options);
link(targetModelId, options);
unlink(targetModelId, options);
exists(targetModelId, options);
```

As we have discussed when implementing other relations (most notably HasMany), `*ById` methods are not very useful - if you already have the id of the target model, then you can load it directly via the target repository, there is no need to make 2 more queries.

I am proposing to remove `findById` and `deleteById`, at least from the initial implementation. Note that `HasMany` repository does not provide these two methods either.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,136 @@
...
+  ): Promise<void>;
+
+  /**
+   * Checks if the given target model instance is associated with the source model.
+   * @param targetModelId - The target model ID to check
+   * @param options
+   * @returns A promise which resolves to true when the association exists
+   * and false otherwise.
+   */
+  exists(
```
bajtos: As I mentioned above, if you already have the id of the target model, then you can check it's existence directly by calling the target repository, there is no need to make an extra query to the through table.

There are two things we may want to check for:

1. Is there an target model instance with the given id?
2. Is the target model instance linked (related) to the source model instance?

I am arguing these two checks can be performed by calling `find`, e.g. `repo.find({where:{id: targetModelId}})`.

If we are interested only in the existence of the target model instance, regardless of whether it's related to the source or not, then we can call `find` on the target repository directly.

I am proposing to remove `exists`, at least from the initial implementation. Note that `HasMany` repository does not provide this method either.

KalleV:
@bajtos Thank you for the 2nd review. I simplified the interface as suggested and added the throughData property to the `link` method options.

KalleV:
> Great, looks good to me.
> 
> I really like the idea of providing `throughData` in `link`. But i think there is also the usecase of updating this `throughData`. How to handle this?

Perhaps the patch method could also pass `throughData` which would optionally update the through model in addition to the target?

derdeka:
> @derdeka Perhaps the patch method could also pass `throughData` which would optionally update the through model in addition to the target?

Let‘s do that in later iterations.

@bajtos as assignee PTAL. Ready to be get merged?

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,97 @@
...
+> {
+  /**
+   * Create a target model instance
+   * @param targetModelData - The target model data
+   * @param options - Options for the operation
+   * @returns A promise which resolves to the newly created target model instance
+   */
+  create(
+    targetModelData: DataObject<Target>,
+    options: Options & {
```
derdeka:
```suggestion
    options?: Options & {
```
After some experiments with this API, i think options should be optional here too.

hacksparrow:

We need the corresponding tests.

bajtos:
> We need the corresponding tests.

How do you propose to test this new interface? AFAIK, we don't have any tests for other relation-repository interface either.

From the perspective of testing, it would be better to land this interface together with an implementation, so that we have some coverage. On the other hand, since the new interface isn't used anywhere yet, I think it's ok to tweak or change it if we discover any problems in future pull requests.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Count, DataObject, Entity, Filter, Options, Where} from '../..';
+
+/**
+ * CRUD operations for a target repository of a HasManyThrough relation
+ */
```
bajtos: Let's add a warning that this interface is experimental. For example (based on [the spike pull request](https://github.com/strongloop/loopback-next/pull/2359/files#diff-177253a1afbdd8eb83b99efef86dd415R269-R271)):

```
 * EXPERIMENTAL: This interface is not stable and may change in the near future. 
 * Backwards-incompatible changes may be introduced in semver-minor versions.
```

Issue #4147, opened 2019-11-18
TodoList tutorial - GET  ​/todos​/{id}​/todo-list gives unexpected response


emonddr:
TodoList tutorial - GET ​/todos​/{id}​/todo-list gives unexpected response
for a todo item that is not part of a todo list.

## Current Behavior
In the `Todo tutorial`, a few todo items are already inside db.json.

```
[
  {
    "id": 1,
    "title": "Take over the galaxy",
    "desc": "MWAHAHAHAHAHAHAHAHAHAHAHAHAMWAHAHAHAHAHAHAHAHAHAHAHAHA"
  },
  {
    "id": 2,
    "title": "destroy alderaan",
    "desc": "Make sure there are no survivors left!"
  },
  {
    "id": 3,
    "title": "play space invaders",
    "desc": "Become the very best!"
  },
  {
    "id": 4,
    "title": "crush rebel scum",
    "desc": "Every.Last.One."
  },
]
```
They are not associated with any list.
No problem.

I reach the end of the `TodoList tutorial`
I create a TodoList #1 
I add two new todo items to it.
No problem.

```
[
  {
    "id": 1,
    "title": "Take over the galaxy",
    "desc": "MWAHAHAHAHAHAHAHAHAHAHAHAHAMWAHAHAHAHAHAHAHAHAHAHAHAHA"
  },
  {
    "id": 2,
    "title": "destroy alderaan",
    "desc": "Make sure there are no survivors left!"
  },
  {
    "id": 3,
    "title": "play space invaders",
    "desc": "Become the very best!"
  },
  {
    "id": 4,
    "title": "crush rebel scum",
    "desc": "Every.Last.One."
  },
  {
    "id": 5,
    "title": "carrots",
    "desc": "White Carrots for soup",
    "isComplete": false,
    "todoListId": 1
  },
  {
    "id": 6,
    "title": "bananas",
    "desc": "Yellow bananas for banana bread",
    "isComplete": false,
    "todoListId": 1
  },
]
```

When I use
`GET /todos​/{todo id}/todo-list` for the two new todo items (id=5 or 6) above, I get the 
proper response of

```
{
  "id": 1,
  "title": "grocery",
  "color": "green"
}
```
No problem.

But when I use 
`GET /todos​/{todo id}/todo-list` for the any item items (id=1 through 4) above (not associated with a list), I get the `unexpected response` of:

```
{
  "id": 1,
  "title": "grocery",
  "color": "green"
}
```

[...]

PR #4325, opened 2019-12-20
fix(repository): belongsTo accessor should return undefined if foreign key is not included

agnes512:
fixes https://github.com/strongloop/loopback-next/issues/4147

Relates to https://github.com/strongloop/loopback-next/issues/2333

## Current behavior
For instances:
```
Customer {
  id: 1,
  name: 'exists'
}

Order {
  id: 1,
  name: 'has fk',
  customerId: 1
}
Order {
  id: 2,
  name: 'customer doesn't exist',
  customerId: 999
}
Order {
  id: 3,
  name: 'no fk',
}
```
The result of await `orderRepo.customer(order.id = 3);` returns the first customer in the database even it doesn't have the foreign [key](url)

## Proposal

**After reading through #2333, this is only a workaround.**
~Should check the foreign key value of the source instance.
Since different databases use different empty value, we check both `undefined` and `null`.~

I think this issue can be closed as duplicate.. 
- when user uses lb4, the memory connector is mostly for tutorials. The priority is relatively low.
- the goal of #2333 is to set referential integrity instead of returning `[]` in this case. 
- this workaround changes the result of other connectors. They are able to throw error `with id "constraint {\"id\":null}"`. With the change, they return `[]`.

## Checklist

### DIFF
--- a/packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
+++ b/packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
@@ -53,6 +53,12 @@ export function createBelongsToAccessor<
     const primaryKey = meta.keyTo;
     const sourceModel = await sourceRepository.findById(sourceId);
     const foreignKeyValue = sourceModel[foreignKey as keyof Source];
+    // workaround to check referential integrity.
+    // should be removed once the memory connector ref integrity is done
+    // GH issue: https://github.com/strongloop/loopback-next/issues/2333
+    if (!foreignKeyValue) {
+      return (undefined as unknown) as Target;
+    }
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const constraint: any = {[primaryKey]: foreignKeyValue};
     const constrainedRepo = new DefaultBelongsToRepository(
--- a/packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
@@ -93,6 +93,18 @@ export function belongsToRelationAcceptance(
       expect(result).to.deepEqual(shipment);
     });
 
+    it('returns undefined if the source instance does not have the foreign key', async () => {
+      await shipmentRepo.create({
+        name: 'Tuesday morning shipment',
+      });
+      const order = await orderRepo.create({
+        // doesn't have the foreign key
+        description: 'Order that is shipped Tuesday morning',
+      });
+      const result = await orderRepo.shipment(order.id);
+      expect(result).to.be.undefined();
+    });
+
     it('throws EntityNotFound error when the related model does not exist', async () => {
       const deletedCustomer = await customerRepo.create({
         name: 'Order McForder',


packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
```diff
@@ -53,6 +53,10 @@ export function createBelongsToAccessor<
     const primaryKey = meta.keyTo;
     const sourceModel = await sourceRepository.findById(sourceId);
     const foreignKeyValue = sourceModel[foreignKey as keyof Source];
+    // different dbs use different empty values
+    if (foreignKeyValue === undefined || foreignKeyValue === null) {
```
jannyHou: maybe use a falsy check like `if(!foreignKeyValue) {}`?

agnes512: 👍 

@jannyHou  Another thing is, we didn't notice that we already have #2333 opened. #4147 is duplicate. I don't know if I should continue on this PR. Because this 'fix' will change some behavior of some connectors ( please the proposal above). 
So I am thinking to close this PR without merging it. Thoughts?

jannyHou: @agnes512 I think #2333 and #4147 are talking about different issues:

2333: means when CREATE related item, if you provide a non-existing fk(like you have users with id 1 through 10, but you create an order for user with id=11), the record will still be created, which is wrong. The solution would be checking a fk's corresponding source model instance exists in db first, then perform the creation. The story is more like an improvement of our design.

4147: looks like a bug, when GET related item, the included items are wrong. like a todo item with id=1 doesn't belong to any todoList, which means it has an empty fk, but GET todos/1/todoList returns a todo list...and please note the WRONGLY returned todo list does exist in the db, the problem is it doesn't own the todo specified in the request path...

jannyHou: Back to this PR, since the `belongsTo` relation only has a `get` method(because child only has the get access to its parent, but cannot modify it), I am good with your approach here :) just return an empty result when fk is missing. 

Something you may want to double check: should it return an empty ARRAY or just null/undefined?

agnes512: You're right. I think the belongsTo accessor only checks the existence. And hasMany and hasOne do creation and traverse.

As for the return, I think this workaround should match with other connectors to throw an error: `with id "constraint {\"id\":null}"`

Once we done the memory connector referential integrity, this workaround should be removed, and let each connector to handle it.

packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
```diff
@@ -53,6 +53,11 @@ export function createBelongsToAccessor<
     const primaryKey = meta.keyTo;
     const sourceModel = await sourceRepository.findById(sourceId);
     const foreignKeyValue = sourceModel[foreignKey as keyof Source];
+    // workaround to check referential integrity.
+    // should be removed once the memory connector ref integrity is done
```
nabdelgadir: Can you add a link to the issue?

packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
nabdelgadir: > As for the return, I think `undefined` is better than `[]`.

In the PR description you said the connectors return `[]`, so should we try to be consistent with that?

agnes512: @nabdelgadir that was my proposal before. But I think Janny's suggestion makes more sense. Will update the description and the title later

packages/repository/src/relations/belongs-to/belongs-to-accessor.ts
agnes512: Good suggestion

PR #4347, opened 2019-12-30
fix: suport complex objects for query params in api explorer

deepakrkris:
issue: https://github.com/strongloop/loopback-next/issues/2208

BREAKING CHANGE

Based on spike https://github.com/strongloop/loopback-next/pull/4141

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/parser.unit.ts
+++ b/packages/rest/src/__tests__/unit/parser.unit.ts
@@ -300,7 +300,7 @@ describe('operationArgsParser', () => {
     expect(args).to.eql(['<html><body><h1>Hello</h1></body></html>']);
   });
 
-  context('in:query style:deepObject', () => {
+  context('in:query content:application/json', () => {
     it('parses JSON-encoded string value', async () => {
       const req = givenRequest({
         url: '/?value={"key":"value"}',
@@ -346,6 +346,32 @@ describe('operationArgsParser', () => {
       );
     });
 
+    it('parses complex json object', async () => {
+      const req = givenRequest({
+        url: '/?filter={"include": [{"relation": "todoList"}]}',
+      });
+
+      const spec = givenOperationWithObjectParameter('filter');
+      const route = givenResolvedRoute(spec);
+
+      const args = await parseOperationArgs(req, route, requestBodyParser);
+
+      expect(args).to.eql([{include: [{relation: 'todoList'}]}]);
+    });
+
+    it('parses url-encoded complex json object', async () => {
+      const req = givenRequest({
+        url: '/?filter=%7B"include"%3A%5B%7B"relation"%3A"todoList"%7D%5D%7D',
+      });
+
+      const spec = givenOperationWithObjectParameter('filter');
+      const route = givenResolvedRoute(spec);
+
+      const args = await parseOperationArgs(req, route, requestBodyParser);
+
+      expect(args).to.eql([{include: [{relation: 'todoList'}]}]);
+    });
+
     it('rejects array values encoded as JSON', async () => {
       const req = givenRequest({
         url: '/?value=[1,2]',
@@ -381,9 +407,11 @@ describe('operationArgsParser', () => {
         {
           name,
           in: 'query',
-          style: 'deepObject',
-          explode: true,
-          schema,
+          content: {
+            'application/json': {
+              schema,
+            },
+          },
         },
       ]);
     }
--- a/packages/rest/src/coercion/coerce-parameter.ts
+++ b/packages/rest/src/coercion/coerce-parameter.ts
@@ -32,7 +32,14 @@ export function coerceParameter(
   data: string | undefined | object,
   spec: ParameterObject,
 ) {
-  const schema = spec.schema;
+  let schema = spec.schema;
+
+  // If a query parameter is a url encoded Json object, the schema is defined under content['application/json']
+  if (!schema && spec.in === 'query' && spec.content) {
+    const jsonSpec = spec.content['application/json'];
+    schema = jsonSpec.schema;
+  }
+
   if (!schema || isReferenceObject(schema)) {
     debug(
       'The parameter with schema %s is not coerced since schema' +
@@ -172,9 +179,9 @@ function parseJsonIfNeeded(
 ): string | object | undefined {
   if (typeof data !== 'string') return data;
 
-  if (spec.in !== 'query' || spec.style !== 'deepObject') {
+  if (spec.in !== 'query' || (spec.in === 'query' && !spec.content)) {
     debug(
-      'Skipping JSON.parse, argument %s is not in:query style:deepObject',
+      'Skipping JSON.parse, argument %s is not a url encoded json object query parameter (since content field is missing in parameter schema)',
       spec.name,
     );
     return data;
--- a/packages/rest/src/coercion/validator.ts
+++ b/packages/rest/src/coercion/validator.ts
@@ -65,9 +65,34 @@ export class Validator {
   isAbsent(value: any) {
     if (value === '' || value === undefined) return true;
 
+    const spec: ParameterObject = this.ctx.parameterSpec;
     const schema: SchemaObject = this.ctx.parameterSpec.schema ?? {};
-    if (schema.type === 'object' && value === 'null') return true;
+    const valueIsNull = value === 'null' || value === null;
 
+    if (this.isUrlEncodedJsonParam()) {
+      // is this an url encoded Json object query parameter?
+      // check for NULL values
+      if (valueIsNull) return true;
+    } else if (spec.schema) {
+      // if parameter spec contains schema object, check if supplied value is NULL
+      if (schema.type === 'object' && valueIsNull) return true;
+    }
+
+    return false;
+  }
+
+  /**
+   * Return `true` if defined specification is for a url encoded Json object query parameter
+   *
+   * for url encoded Json object query parameters,
+   * schema is defined under content['application/json']
+   */
+  isUrlEncodedJsonParam() {
+    const spec: ParameterObject = this.ctx.parameterSpec;
+
+    if (spec.in === 'query' && spec.content?.['application/json']?.schema) {
+      return true;
+    }
     return false;
   }
 }
--- a/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts
+++ b/examples/todo-list/src/__tests__/acceptance/todo-list.acceptance.ts
@@ -204,6 +204,16 @@ describe('TodoListApplication', () => {
     });
   });
 
+  it('exploded filter conditions work', async () => {
+    const list = await givenTodoListInstance(todoListRepo);
+    await givenTodoInstance(todoRepo, {title: 'todo1', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo2', todoListId: list.id});
+    await givenTodoInstance(todoRepo, {title: 'todo3', todoListId: list.id});
+
+    const response = await client.get('/todos').query('filter[limit]=2');
+    expect(response.body).to.have.length(2);
+  });
+
   /*
    ============================================================================
    TEST HELPERS
--- a/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
+++ b/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
@@ -193,6 +193,17 @@ describe('TodoApplication', () => {
       .expect(200, [toJSON(todoInProgress)]);
   });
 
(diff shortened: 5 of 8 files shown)


packages/rest/src/coercion/validator.ts
```diff
@@ -65,9 +65,19 @@ export class Validator {
   isAbsent(value: any) {
     if (value === '' || value === undefined) return true;
 
+    const spec: ParameterObject = this.ctx.parameterSpec;
```
jannyHou: Maybe refactor the code here to be:

```ts
const spec: ParameterObject = this.ctx.parameterSpec;
const valueIsNull = value === 'null' || value === null;

if (spec.schema) {
  if (schema.type === 'object' && valueIsNull) return true;
} else {
  if (spec.in === 'query' && spec.content && valueIsNull)
    return true;
}
```

if I understand the logic here correctly.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -32,7 +32,14 @@ export function coerceParameter(
   data: string | undefined | object,
   spec: ParameterObject,
 ) {
-  const schema = spec.schema;
+  let schema = spec.schema;
+
+  if (!schema && spec.content) {
+    const content = spec.content;
+    const jsonSpec = content['application/json'];
```
jannyHou: @deepakrkris The encoding type should be inferred from the request context, `coerceParameter` is called [here](https://github.com/strongloop/loopback-next/blob/4406bd3b9de0fa023896f7c97cb2c98157ca9b71/packages/rest/src/parser.ts#L86), which has access to the `request`.

So I would suggest add a 3rd parameter `options` for `coerceParameter`, and provide the `request` as `options.reqCtx`(or some better name), then retrieve the encoding type from `request`.

If `schema` found then use it, if no `schema` field, search for `spec.content[encodingType]`.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -172,7 +179,7 @@ function parseJsonIfNeeded(
 ): string | object | undefined {
   if (typeof data !== 'string') return data;
 
-  if (spec.in !== 'query' || spec.style !== 'deepObject') {
+  if (spec.in !== 'query' || (spec.style !== 'deepObject' && !spec.content)) {
     debug(
       'Skipping JSON.parse, argument %s is not in:query style:deepObject',
```
bajtos: Please update this debug log to match the new condition.

packages/rest/src/coercion/validator.ts
deepakrkris: @jannyHou that is perfect, I will change as per your suggestion.

deepakrkris: changes done as suggested

1
```diff
@@ -0,0 +1,15 @@
+fix: code review comments
```
raymondfeng: What is this file named as `1`? Please remove it.

deepakrkris: agreed

packages/rest/src/coercion/validator.ts
```diff
@@ -65,8 +65,25 @@ export class Validator {
...
+    // if parameter spec contains schema object, check if supplied value is NULL
+    if (spec.schema) {
+      if (schema.type === 'object' && valueIsNull) return true;
+    } else {
+      // for url encoded Json object query parameters, schema is defined under content['application/json']
+      if (
+        spec.in === 'query' &&
+        spec.content &&
+        spec.content['application/json'] &&
+        spec.content['application/json'].schema
```
raymondfeng: Is it possible to use https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining?

Can we have a helper method to check this case?

packages/rest/src/coercion/coerce-parameter.ts
deepakrkris: agreed

packages/rest/src/coercion/coerce-parameter.ts
deepakrkris: @jannyHou , 'application/json' here does not refer to the incoming request's encoding type. It tells what type of content a particular `parameter` carries.  As per Open API spec 3 , `in:query` and `content: application/json: schema` together means the query parameter is a json encoded string.

deepakrkris: @jannyHou we can safely look for only `application/json` in the content field for condition `in: query` , because Open API v3 supports only this encoding type inside content field for a query parameter.

deepakrkris:
> According to the spike, the implementation part LGTM. Can we have some tests to check the changes and different encoding styles?

I have added various test cases. Could you please verify and comment.

packages/rest/src/__tests__/unit/parser.unit.ts
```diff
@@ -346,6 +346,19 @@ describe('operationArgsParser', () => {
       );
     });
 
+    it('parses url-encoded complex json object', async () => {
+      const req = givenRequest({
```
jannyHou: nitpick: do we have to use an encoded url with `%7B`, %3A, etc..?

deepakrkris:
> Those tests LGTM. I have two concerns:
> 
> * you might want check if the description of the tests that you added/modified can be more descriptive
> * will the docs be added in the short future?

agreed, I will modify the test descriptions to be more explicit. 
I will add the docs after this PR is merged. It is covered as a task in issue https://github.com/strongloop/loopback-next/issues/2208

packages/rest/src/coercion/validator.ts
jannyHou: +1 for Raymond's comment, try
```ts
if (spec.content?.['application/json']?.schema) {
//
}
```
?

packages/rest/src/__tests__/unit/parser.unit.ts
deepakrkris: @jannyHou using the encoded url, is an explicit way of testing if the same query request will work from api explorer. otherwise supertest implicitly converts json query objects to url encoded request.

packages/rest/src/coercion/validator.ts
deepakrkris: @raymondfeng @jannyHou I have implemented as per your suggestions with optional chaining. please verify.

raymondfeng:
@deepakrkris In the commit message for https://github.com/strongloop/loopback-next/pull/4347/commits/cd3b772c8d94d57c00b091ce1acfc47a17115eb9, we should add more details about the `BREAKING CHANGE` to explain:

1. What's breaking?
2. What do current users have to adjust?

deepakrkris:
> @deepakrkris In the commit message for [cd3b772](https://github.com/strongloop/loopback-next/commit/cd3b772c8d94d57c00b091ce1acfc47a17115eb9), we should add more details about the `BREAKING CHANGE` to explain:
> 
> 1. What's breaking?
> 2. What do current users have to adjust?

@raymondfeng @bajtos updated the commit message with the `BREAKING CHANGE` description

```
BREAKING CHANGE: This fix changes the schema definition of query parameters with `in: query`
and `style: deepObject` by moving the `schema` field under `content[application/json]`,
i.e, json query params are changed to accomodate url encoding as per open api specifications.
The `style` and `explode` fields are also removed for such query parameters`.
The signature of the `param.query.object` decorator has not changed due to this fix.
There is no code changes required after upgrading this fix. No method signatures or
data structures are impacted. The impact is only on the open api specification generated
from the LoopBack APIs with json query params.
All clients and consumers of such APIs may have to regenerate the open api specifications.
```

raymondfeng:
@deepakrkris Can you squash the commits into one?

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -172,9 +179,9 @@ function parseJsonIfNeeded(
 ): string | object | undefined {
   if (typeof data !== 'string') return data;
 
-  if (spec.in !== 'query' || spec.style !== 'deepObject') {
+  if (spec.in !== 'query' || (spec.in === 'query' && !spec.content)) {
```
bajtos: IIUC, this is introducing another breaking change. If I am reading the code right:

- Before this change, when a parameter was described with `style: deepObject`, we attempted to parse it as JSON.
- After this change, when a parameter is described with `style: deepObject`, we never parse the value as JSON and require the value to be encoded using deep query string.

While the new behavior is technically more correct, I am little bit concerned about the impact on the existing applications. I understand this does not apply to apps using `@param.query.object` decorator, which is probably the majority of the apps out there. However, it is still possible to specify `style: deepObject` encoding for parameters, for example using generic `@param` decorator or even using a design-first approach where the entire OpenAPI spec is provided from outside via `app.api()` method, see [Defining the API using design-first approach](https://github.com/strongloop/loopback-next/blob/4406bd3b9de0fa023896f7c97cb2c98157ca9b71/docs/site/Defining-the-API-using-design-first-approach.shelved.md).

If don't have strong opinions on whether preserve old (more lenient) behavior or introduce the new (more strict) handling. If we decide to be more strict, then please make sure this change is spelled out in the release notes (the commit message).

bajtos: On the second thought, I think now is a good time to introduce this breaking change, since we are already changing other things. It should be easy (and backwards compatible) to reintroduce support for JSON-encoded deep-object parameters in the future, assuming there is anybody asking for that feature.

So the only remaining step is to update the commit message and explain the breaking change.

bajtos:
BTW, it makes me wonder if there are any changes needed in these areas, to ensure they stay in sync with the actual runtime behavior?

For example, if we have any documentation page showing an example of an OpenAPI spec document (or a fragment) generated by LoopBack, then we need to update it to reflect the changes in the runtime.

A quick string search for `deepObject` found the following place that I think needs an update: https://github.com/strongloop/loopback-next/blob/75ba77d5bdfe24f612c418c9100fa7e008211709/docs/site/Parsing-requests.md#object-values, see the full content below the line.

---

OpenAPI specification describes several ways how to encode object values into a
string, see [Style Values](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-values) and [Style Examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-examples).

At the moment, LoopBack supports object values for parameters in query strings with `style: "deepObject"` only. Please note that this style does not preserve encoding of primitive types, numbers and booleans are always parsed as strings.

For example:

```
GET /todos?filter[where][completed]=false
// filter={where: {completed: 'false'}}
```

As an extension to the deep-object encoding described by OpenAPI, when the parameter is specified with `style: "deepObject"`, we allow clients to provide the object value as a JSON-encoded string too.

For example:

```
GET /todos?filter={"where":{"completed":false}}
// filter={where: {completed: false}}
```

packages/rest/src/coercion/coerce-parameter.ts
deepakrkris: > * After this change, when a parameter is described with `style: deepObject`, we never parse the value as JSON and require the value to be encoded using deep query string.
@bajtos share your concern on the above , but as of now there is no decorator to annotate a parameter with `style:deepObject` so there is no way a customer can have a case of trying to send a complex json for a parameter marked with spec `deepObject`.

deepakrkris: > * After this change, when a parameter is described with `style: deepObject`, we never parse the value as JSON and require the value to be encoded using deep query string.

for the change in this behavior I have added the following in the commit footer:
``` 
This fix has also modified the coercion behaviour for `in:query`, `deep-object` query
parameters by removing json parsing for them and the expectation is to pass in only
string values. But since the `exploded` `deep-object` style definition for query params
has been removed, this behaviour remains just an internal source code aspect as of now.
```

deepakrkris:
> A quick string search for `deepObject` found the following place that I think needs an update: https://github.com/strongloop/loopback-next/blob/75ba77d5bdfe24f612c418c9100fa7e008211709/docs/site/Parsing-requests.md#object-values, see the full content below the line.
> 
> OpenAPI specification describes several ways how to encode object values into a
> string, see [Style Values](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-values) and [Style Examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-examples).
> 
> At the moment, LoopBack supports object values for parameters in query strings with `style: "deepObject"` only. Please note that this style does not preserve encoding of primitive types, numbers and booleans are always parsed as strings.

@bajtos I will create a separate PR to update for this. Thanks for bringing this to my notice.

dhmlau:
@slnode test please

Issue #4290, opened 2019-12-11
feat(repository-json-schema): add title to filter


derdeka:
## Suggestion

Add `title` property to filter schema definition in order to enable better code generation by third party tools such as https://github.com/OpenAPITools/openapi-generator .

Without a title property the generator simply name the filter as `Filter1`, `Filter2`, etc. By adding the title property, the tooling can produce a filter interface named after `title`.

## Use Cases

Parameters of controller methods are decorated with `@param.query.object` where `getFilterSchemaFor(User)` generates a `JsonSchema` for the filter.
```
@param.query.object('filter', getFilterSchemaFor(User)) filter?: Filter<User>,
```

When using openapi-generator, this filter gets just named `Filter1`. With a large number of filters in a project and randomized order of generation, the git history of the generated files gets trashed.

## Discussion points

One filter `getFilterSchemaFor(ModelX)` can be used multiple times in a project. Currently the filters are created inline. Simply adding a `title` like `ModelXFilter` creates name collisions when generating the client sdk. Is it helpful to move Filters to `/components/schemas` ?

## Acceptance criteria

TBD - will be filled by the team.

dougal83:
+1 for refactoring the filter object.  Thank you for looking further into this!

bajtos:
@derdeka thank you for a great suggestion! I agree we should add `title` to the schema generated for model filters.

> One filter `getFilterSchemaFor(ModelX)` can be used multiple times in a project. Currently the filters are created inline. Simply adding a `title` like `ModelXFilter` creates name collisions when generating the client sdk. Is it helpful to move Filters to `/components/schemas` ?

I see two options (but there may be more):

Short-term, we can use the approach we have in place for model schemas - instead of emitting an "inline" schema, we generate a schema reference and include the target schema in definitions property. Take a look at `getModelSchemaRef` works under the hood, we can build `getFilterSchemaRefFor` on a similar way. I find this approach as not very pretty, but it makes the implementation of the requested feature quick & easy.

For longer term, I would like our REST layer to be smarter and able to automatically extract schemas used in multiple places into `/components/schemas` and replace them with `$ref` at usage points.

This logic can work as follows:
1. When building OpenAPI schema, we need to scan the entire OpenAPI spec tree and look for all known places where a schema object can be provided. Those places include endpoint parameters, request bodies, response schemas, and so one. I think it's ok to start with few commonly-used places first and add more of them later, as we learn more.
[...]

PR #4355, opened 2020-01-01
Add title to filter schemas

dougal83:
Adding titles to json filter schemas as groundwork toward openapi spec ref consolidation based on title property. See #4290

## Checklist

### DIFF
--- a/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
@@ -21,13 +21,54 @@ describe('filterSchema', () => {
     const schema = getFilterSchemaFor(MyUserModel);
     expect(MyUserModel.definition.name).to.eql('my-user-model');
     expect(schema).to.eql({
+      title: 'my-user-model.Filter',
       properties: {
         where: {
           type: 'object',
+          title: 'my-user-model.WhereFilter',
           additionalProperties: true,
         },
         fields: {
           type: 'object',
+          title: 'my-user-model.Fields',
+          properties: {
+            id: {type: 'boolean'},
+            age: {type: 'boolean'},
+          },
+          additionalProperties: false,
+        },
+        offset: {type: 'integer', minimum: 0},
+        limit: {type: 'integer', minimum: 1, example: 100},
+        skip: {type: 'integer', minimum: 0},
+        order: {type: 'array', items: {type: 'string'}},
+      },
+      additionalProperties: false,
+    });
+  });
+
+  @model({
+    name: 'CustomUserModel',
+  })
+  class CustomUserModel extends Entity {
+    @property() id: string;
+
+    @property() age: number;
+  }
+
+  it('generates filter schema with custom name', () => {
+    const schema = getFilterSchemaFor(CustomUserModel);
+    expect(CustomUserModel.definition.name).to.eql('CustomUserModel');
+    expect(schema).to.eql({
+      title: 'CustomUserModel.Filter',
+      properties: {
+        where: {
+          type: 'object',
+          title: 'CustomUserModel.WhereFilter',
+          additionalProperties: true,
+        },
+        fields: {
+          type: 'object',
+          title: 'CustomUserModel.Fields',
           properties: {
             id: {type: 'boolean'},
             age: {type: 'boolean'},
--- a/packages/repository-json-schema/src/__tests__/unit/filter-json-schema.unit.ts
+++ b/packages/repository-json-schema/src/__tests__/unit/filter-json-schema.unit.ts
@@ -8,6 +8,7 @@ import {expect} from '@loopback/testlab';
 import Ajv from 'ajv';
 import {JsonSchema} from '../..';
 import {
+  getFieldsJsonSchemaFor,
   getFilterJsonSchemaFor,
   getWhereJsonSchemaFor,
 } from '../../filter-json-schema';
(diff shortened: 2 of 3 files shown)


packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -22,6 +25,10 @@ const scopeFilter = getFilterJsonSchemaFor(EmptyModel);
  */
 export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
   const schema: JsonSchema = {
+    title:
+      modelCtor.modelName === 'EmptyModel'
```
bajtos: Checking the model name is not safe enough, because users can have their own model that's called `EmptyModel`. We should compare the model ctor function directly.

```suggestion
      modelCtor === EmptyModel
```

To be honest, I am not very happy that we have to pollute schema-generating functions with checks for EmptyModel. I am concerned that it's too easy for future contributors to forget to implement this check and accidentally introduce `EmptyModel` title into the result.

Since `EmptyModel` was introduced only as a tool to simplify implementation of `scopeFilter`, I prefer to keep all EmptyModel-related handling inside the code building `scopeFilter` value.

How about introducing an optional `options` argument to all `get*SchemaFor` methods, one that will allow the caller to request a schema with no title set?

```ts
const scopeFilter = {
  ...getFilterJsonSchemaFor(EmptyModel, {setTitle: false}),
  title: 'ScopeFilter', // TODO(dougal83) base title on model
};
```

Implementation:

```ts
export function getFilterJsonSchemaFor(
  modelCtor: typeof Model,
  options: FilterSchemaOptions
): JsonSchema {
  const schema: JsonSchema = {
    title: /*...*/,
    properties: {
      // Forward the options. If the caller does not want title for the top-level schema,
      // then we should not set titles in the nested schemas either
      where: getWhereJsonSchemaFor(modelCtor, options),

      fields: getFieldsJsonSchemaFor(modelCtor, options),
   
      // ...
    },
    // ...
  };
  // ...
}
```

Don't forget to write new tests to verify correctness of the implementation handling the new option.

packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
```diff
@@ -21,13 +21,16 @@ describe('filterSchema', () => {
     const schema = getFilterSchemaFor(MyUserModel);
     expect(MyUserModel.definition.name).to.eql('my-user-model');
     expect(schema).to.eql({
+      title: 'my-user-modelFilter',
```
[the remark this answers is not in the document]
dougal83: I'll alter model name to be PascalCase (wasn't sure if it was [set for a reason](https://github.com/strongloop/loopback-next/blob/d4e42abc926af27f968765dee9f41017023fc369/packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts#L12)).  Going with `MyUserModel.Filter` as it is Azure friendly and consistent with the qualified count `loopback.count`.

packages/repository-json-schema/src/filter-json-schema.ts
dougal83: Using options is a much better idea!  Will update.

dougal83:
I've done a `git rebase -i` to fix history but that is probably a bad idea... not having a good evening.

bajtos:
> I've done a `git rebase -i` to fix history but that is probably a bad idea... not having a good evening.

No worries.

When working on a pull request, I usually run `git rebase master` to bring changes from the master branch, and then commit additional changes in new commits. This allows reviewers to look at incremental changes only.

Fortunately, your pull request is reasonably small, so it's not that difficult to review it in whole.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -20,12 +32,16 @@ const scopeFilter = getFilterJsonSchemaFor(EmptyModel);
  *
  * @param modelCtor - The model constructor to build the filter schema for.
  */
-export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
+export function getFilterJsonSchemaFor(
+  modelCtor: typeof Model,
+  options: FilterSchemaOptions,
```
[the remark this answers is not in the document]
dougal83: Thanks.  I couldn't put my finger on it the other night!

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -83,13 +102,19 @@ export function getFilterJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
  *
  * @param modelCtor - The model constructor to build the filter schema for.
  */
-export function getWhereJsonSchemaFor(modelCtor: typeof Model): JsonSchema {
+export function getWhereJsonSchemaFor(
+  modelCtor: typeof Model,
+  options: FilterSchemaOptions,
+): JsonSchema {
   const schema: JsonSchema = {
+    title: `${modelCtor.modelName}.Where`,
```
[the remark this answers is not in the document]
dougal83: Currently going with `.Filter`, `.WhereFilter`, `.IncludeFilter`, and `.ScopeFilter`.  Plus `.Fields`.

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -6,10 +6,22 @@
...
+  setTitle?: boolean;
+}
+
 @model({settings: {strict: false}})
 class EmptyModel extends Model {}
 
-const scopeFilter = getFilterJsonSchemaFor(EmptyModel);
+const scopeFilter = {
+  ...getFilterJsonSchemaFor(EmptyModel, {setTitle: false}),
+  title: 'ScopeFilter', // TODO(dougal83) base title on model
```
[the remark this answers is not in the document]
dougal83: I think I'll do something like this... currently rendering the new `setTitle` property as useful as a chocolate teapot?

````ts
export function getScopeFilterJsonSchemaFor(
  modelCtor: typeof Model,
  options: FilterSchemaOptions = {},
): JsonSchema {
  @model({settings: {strict: false}})
  class EmptyModel extends Model {}

  const schema: JsonSchema = {
    ...getFilterJsonSchemaFor(EmptyModel, {setTitle: false}),
    title:
      options.setTitle !== false
        ? `${modelCtor.modelName}.ScopeFilter`
        : undefined,
  };

  return schema;
}
````

File: packages/repository-json-schema/src/__tests__/unit/filter-json-schema.unit.ts [hunk 6]
```diff
@@ -157,15 +171,52 @@ describe('getFilterJsonSchemaFor', () => {
...
+      .include as JsonSchema;
+    expect(filterInclude.title).to.equal('Customer.IncludeFilter');
+  });
+
+  it('returns "scope.title" when a single option "setTitle" is set', () => {
+    const filterInclude = customerFilterSchema.properties!.include;
+    const filterScope = ((filterInclude as JsonSchema).items as JsonSchema)
+      .properties!.scope as JsonSchema;
+    expect(filterScope.title).to.equal('Customer.ScopeFilter');
+  });
```
dougal83: Is there a better way than as ... as ... ?

packages/openapi-v3/src/__tests__/unit/filter-schema.unit.ts
```diff
@@ -21,13 +21,54 @@ describe('filterSchema', () => {
...
+  @model({
+    name: 'CustomUserModel',
+  })
+  class CustomUserModel extends Entity {
+    @property() id: string;
+
+    @property() age: number;
+  }
+
+  it('generate filter schema with custom name', () => {
```
bajtos: Nitpick:

```suggestion
  it('generates filter schema with custom name', () => {
```

As in an English sentence: _it generates filter schema etc._

packages/repository-json-schema/src/__tests__/unit/filter-json-schema.unit.ts
bajtos: I think the problem is in the way how `@types/json-schema` describes the schema object. Each property definition can be either `JSONSchema6` (what we use) or `boolean`. TypeScript cannot tell that we are always using `JSONSchema6` and never `boolean`, that's why you have to use explicit `as` casts.

It may be possible to work around this problem by explaining TypeScript that `getFilterJsonSchemaFor` is returning not just some random `JsonSchema`, but a schema describing an object with few well-known properties (`where`, `fields`, etc.) that are defined using a `JSONSchema6` object, not as a `boolean` value. There may be a way how to do that in a generic way, something along the way of "for every key of `Filter` type, tell TypeScript that the schema describes a property of that name.

I am not sure if it's worth the effort though.

To simplify your tests, maybe you can use deep property checks, e.g. using [propertyByPath](http://shouldjs.github.io/#assertion-propertybypath) assertion. An mock-up to illustrate what I mean:

```ts
expect(customerFilterSchema.properties)
  .to.have.propertyByPath(
    ['include', 'items', 'properties', 'scope', 'title']
  )
  .equal('Customer.ScopeFilter');
```

dougal83:
@bajtos Thanks for the feedback.  How do I resolve the commit lint failure?

bajtos:
@dougal83 

>  How do I resolve the commit lint failure

Huh, it looks like the linter does not understand that your pull request is adding a single commit only and as a result, it's linting existing commits on master. I vaguely remember that we have encountered this problem once or twice in the past, but that was years ago and I don't remember much details.

Let's try to rebase the pull request on top of the latest master to see if it helps.

bajtos:
Rebase fixed the problem.

The pull request is landed, thank you @dougal83 for the contribution! ❤️

dougal83:
@bajtos Thank you for sorting it out for me!

PR #4362, opened 2020-01-03
docs: boot script migration

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/3957

This PR contains the migration guide for LB3 boot script, and two observers migrated from 2 typical LB3 scripts as demo. After the doc gets approved, I will delete the code part.

An outline of the covered topics:

- [x] sync boot script
- [x] async boot script
- [x] migrate model/datasource/other artifacts used in the script first
- [x] modify the script accordingly, server-->app
- [x] ignore the two default scripts, auth and root

## Checklist

### DIFF
--- a/docs/site/migration/boot-scripts.md
+++ b/docs/site/migration/boot-scripts.md
@@ -6,8 +6,287 @@ sidebar: lb4_sidebar
 permalink: /doc/en/lb4/migration-boot-scripts.html
 ---
 
-{% include note.html content="
-This is a placeholder page, the task of adding content is tracked by the
-following GitHub issue:
-[loopback-next#3957](https://github.com/strongloop/loopback-next/issues/3957)
-" %}
+In LoopBack 3, predefined boot scripts are organized in the `/server/boot`
+directory, and are executed right before the server starts to perform some
+custom application initialization. The same functionality can also be achieved
+in LoopBack 4 application by adding observers.
+
+The LoopBack 3 boot script is mapped to LoopBack 4 observer's `start` function.
+And moreover, besides participating in an application's start phase, an observer
+also allows you to perform logic in the stop phase.
+
+## LoopBack 3 Boot Script
+
+LoopBack 3 supports three styles of boot scripts:
+
+- synchronous boot script
+- asynchronous boot script that takes in a callback
+- asynchronous boot script that returns a promise
+
+They all have the first parameter as `server` (some people are used to name it
+as `app`). The callback based asynchronous script has a 2nd parameter `done` as
+the callback function.
+
+A typical synchronous boot script that prints out a datasource's name before
+application runs:
+
+{% include code-caption.html content="server/boot/print-info.js" %}
+
+```js
+'use strict';
+
+module.exports = function printInfo(server) {
+  // SUPPOSE your application has a datasource called `db`, which is usually the default
+  // memory datasource in a sample LoopBack application created by CLI, regardless if it's
+  // an LB3 app or LB4 app
+  const db = server.datasources.db;
+  console.log('This is a synchronous script.');
+  console.log('Your app has a datasource called: ', db.name);
+};
+```
+
+A typical asynchronous boot script that takes in a callback:
+
+{% include code-caption.html content="server/boot/create-sample-cb.js" %}
+
+```js
+// in file `server/boot/create-sample-cb.js`
+'use strict';
+
+module.exports = function createSample(server, done) {
+  // SUPPOSE your application has a Todo model
+  const Todo = server.models.Todo;
+  const sample = {
+    title: 'a todo sample - cb',
+    desc: 'created by callback based script',
+  };
+  Todo.create(sample, (err, result) => {
+    if (err) return done(err);
+    console.log('Sample created as: ', result);
+    done();
+  });
+};
+```
+
+A typical asynchronous boot script that returns a promise:
+
+{% include code-caption.html content="server/boot/create-sample-promise.js" %}
+
+```js
+'use strict';
+
+module.exports = function createSample(server) {
+  // SUPPOSE your application has a Todo model
+  const Todo = server.models.Note;
+  const sample = {
+    title: 'a todo sample - promise',
+    desc: 'created by promise based script',
+  };
+  return Todo.create(sample).then(result => {
+    console.log('Sample created as: ', result);
+  });
+};
+```
+
+Since a LoopBack 4 application is created in TypeScript, which supports the
+latest ECMAScript features including
+[`async/await`](https://javascript.info/async-await), there is no such signature
+difference any more, the `start` function in the observer is always an `async`
+function.
+
+## Migrating to LoopBack 4 Observer
+
+To perform the same logic defined in a LoopBack 3 boot script, you should create
+a corresponding observer in your LoopBack 4 application.
+
+LoopBack 4 has its own life cycles at runtime and has `boot`, `start`, `stop` as
+its transition states. Observer allows you to participate in the `start` state,
+which is equivalent to the LoopBack 3 custom boot script.
+
+{% include note.html content="We highly recommend you learn about
+[Life cycle events and observers](https://loopback.io/doc/en/lb4/Life-cycle.html)
+first before reading the following guide." %}
+
+Taking the two scripts(`server/boot/print-info.js` and
+`server/boot/create-sample-promise.js`) above as examples, let's first migrate
+`server/boot/print-info.js` to your LoopBack 4 application.
+
+### Creating Observer
+
+Run command `lb4 observer` to create an observer called `printInfo`:
+
+```sh
+lb4 observer
+? Observer name: printInfo
+? Observer group:
+   create src/observers/print-info.observer.ts
+   update src/observers/index.ts
+
+Observer PrintInfo was created in src/observers/
+```
+
+You will find a file called `PrintInfo.observer.ts` created in folder
+`src/observers/`. Open that file, there is a class called `PrintInfoObserver`,
+which implements `LifeCycleObserver` and has `start` and `stop` functions. The
+`start` function is where you should copy over the content of LB3's `printInfo`
+function.
+
+### Injecting Application
+
+The LoopBack 3 boot script takes in `server` as its first parameter, which is
+essentially the `application` injected using key
+`CoreBindings.APPLICATION_INSTANCE`:
+
+```ts
+// ... imports
+export class PrintInfoObserver implements LifeCycleObserver {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
+  ) {
+    // ... other class members
+  }
+}
+```
+
+Now within the class `PrintInfoObserver` you have access to the application.
+
+### Migrating Artifacts
+
+A boot script usually accesses or manipulates an application's artifacts like
+models, datasources, etc... So when moving its content to the `start` function,
+make sure you edit the code to retrieve those artifacts in LoopBack 4's way.
+
+For the print info script, since it retrieves a datasource's info, you need to
+**MIGRATE** the LoopBack datasource `db` to your LoopBack 4 application. The
+steps are well documented on page [migration/datasources.md](./datasources.md).
+And in the `start` function, **EDIT CODE** to inject the datasource using
+`@inject('datasources.db') private ds: juggler.DataSource`:
+
+_The code snippet is the complete observer file, the LB3 boot script logic is
+only in the `start` function_
+
+```ts
+import {
+  /* inject, Application, CoreBindings, */
+  lifeCycleObserver, // The decorator
+  LifeCycleObserver,
+  inject,
+  CoreBindings,
+  Application, // The interface
+} from '@loopback/core';
+import {juggler} from '@loopback/repository';
+
+/**
+ * This class will be bound to the application as a `LifeCycleObserver` during
+ * `boot`
+ */
+@lifeCycleObserver()
+export class PrintInfoObserver implements LifeCycleObserver {
+  constructor(
+    // inject `app` if you need access to other artifacts by `await this.app.get()`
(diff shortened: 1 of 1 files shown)


docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+
+The LoopBack 3 boot script is mapped to LoopBack 4 observer's `start` function.
+And moreover, besides participating in an application's start phase, an observer
+also allows you to perform logic in the stop phase.
+
+## Migrating Boot Scripts
+
+### LoopBack 3 Boot Script
+
+LoopBack 3 has different signatures for synchronized and asynchronized boot
```
bajtos: AFAIK, we use the terms "synchronous" and "asynchronous", not synchronized and asynchronized. See e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function.

```suggestion
LoopBack 3 has different signatures for synchronous and asynchronous boot
```

Please update all new content accordingly.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+  const Todo = server.models.Todo;
+  const sample = {title: 'a todo sample', desc: 'Something to do.'};
+  Todo.create(sample)
+    .then(result => {
+      console.log('Sample created as: ', result);
+    })
+    .then(done);
+};
+```
+
```
bajtos: I believe that boot scripts can use `async` functions too (or in fact any function that returns a promise). Can you please mention this fact and perhaps show a code snippet too?

Also in the callback-based code snippet above, please don't mix promises with callbacks, the conversion from a promise to a callback is tricky. For example, your code above does not handle errors (promise rejections) and will trigger unhandled promise rejection warning.

A better example of a callback-based boot script:

```ts
module.exports = function createSample(server, done) {
  // SUPPOSE your application has a Todo model
  const Todo = server.models.Todo;
  const sample = {title: 'a todo sample', desc: 'Something to do.'};
  Todo.create(sample, (err, result) => {
    if (err) return done(err);
    console.log('Sample created as: ', result);
    done();
  });
 };
 ```

An example of a promise-based boot script:

```ts
module.exports = function createSample(server) {
  // SUPPOSE your application has a Todo model
  const Todo = server.models.Todo;
  const sample = {title: 'a todo sample', desc: 'Something to do.'};
  return Todo.create(sample)
    .then(result => {
      console.log('Sample created as: ', result);
    });
 };
 ```

File: docs/site/migration/boot-scripts.md [hunk 3]
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+In LoopBack 3, predefined boot scripts are organized in the `/server/boot`
+directory, and are executed right before the server starts to perform some
+custom application initialization. The same functionality can also be achieved
+in LoopBack 4 application by adding observers.
+
+The LoopBack 3 boot script is mapped to LoopBack 4 observer's `start` function.
+And moreover, besides participating in an application's start phase, an observer
+also allows you to perform logic in the stop phase.
+
+## Migrating Boot Scripts
```
bajtos: Since this whole page is titled "Migrating boot scripts" (the level 1 heading), I feel it's redundant to introduce a level 2 heading with the same name.

I am proposing to remove this line and reduce level of the headings below, e.g. `## LoopBack 3 Boot Script`.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+
+```ts
+// ... imports
+export class PrintInfoObserver implements LifeCycleObserver {
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE) private app: Application,
+  ) {
+    // ... other class members
+  }
+```
```
bajtos: Please add `}` to close the class definition.

The markdown & typescript parser & highlighter used by GitHub is confused now and shows the text below in blue-ish color, while regular text should be rendered in black (see the text before this code snippet).

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+
+A boot script usually accesses or manipulates an application's artifacts like
+models, datasources, etc... So when moving its content to the `start` function,
+make sure you edit the code to retrieve those artifacts in LoopBack 4's way.
+
+For the print info script, since it retrieves a datasource's info, you need to
+**MIGRATE** the LoopBack datasource `db` to your LoopBack 4 application. The
+steps are well documented on page [migration/datasources.md](./datasources.md).
+And in the `start` function, **EDIT CODE** to retrieve the datasource using
+`this.app.getSync('datasources.db')`:
```
bajtos: I have two comments:

1. In general, we should teach users to call `await app.get()`, because we cannot guarantee that all datasources will be always bound in a way that allows synchronous resolution.

2. Resolving dependencies manually via `get`/`getSync` is an anti-pattern. Users should use `@inject` instead. If the dependency is not bound at the time when constructor dependencies are resolved, then users should replace `@inject` with `@inject.getter`.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+
+  /**
+   * This method will be invoked when the application starts
+   */
+  async start(): Promise<void> {
+    // Add your logic for start
+    console.log('This is a migrated asynchronized boot script');
+
+    // Retrieve the repository by key 'repositories.TodoRepository'
+    const TodoRepo: TodoRepository = this.app.getSync(
```
bajtos: Please use `@inject` or `@inject.getter` to obtain the repository.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,249 @@ sidebar: lb4_sidebar
...
+  CoreBindings,
+  Application, // The interface
+} from '@loopback/core';
+import {juggler} from '@loopback/repository';
+
+/**
+ * This class will be bound to the application as a `LifeCycleObserver` during
+ * `boot`
+ */
+@lifeCycleObserver('')
```
hacksparrow: Can't the `''` be omitted?

docs/site/migration/boot-scripts.md
jannyHou: Good point, refactored this part to be:

- sync script
- async script with cb
- async script returning a promise

docs/site/migration/boot-scripts.md
jannyHou: oops

docs/site/migration/boot-scripts.md
jannyHou: Agreed changed to `@inject`
> because we cannot guarantee that all datasources will be always bound in a way that allows synchronous resolution.

Hmm...why it is not guaranteed? I think datasources are loaded first then the observers.

docs/site/migration/boot-scripts.md
jannyHou: yep, removed and added one more section for setting the scripts order.

docs/site/migration/boot-scripts.md
bajtos: > > because we cannot guarantee that all datasources will be always bound in a way that allows synchronous resolution.
>
> Hmm...why it is not guaranteed? I think datasources are loaded first then the observers.

Consider the following code:

```ts
app
  .bind('datasources.db')
  .toDynamicValue(async function createDb() {
    const config = // do some async work
    return new juggler.DataSource(config);
  })
  .inScope(BindingScope.SINGLETON);
```

It binds the datasource in such way that the first call of `app.get` or `app.getSync` is required to perform async computation.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,278 @@ sidebar: lb4_sidebar
...
+
+The LoopBack 3 boot script is mapped to LoopBack 4 observer's `start` function.
+And moreover, besides participating in an application's start phase, an observer
+also allows you to perform logic in the stop phase.
+
+## LoopBack 3 Boot Script
+
+LoopBack 3 supports synchronous and asynchronous boot scripts. And you can
+choose either callback or promise based implementation for asynchronous scripts.
+They all have the first parameter as `server`(some people are used to name it as
```
bajtos: Typo.

```suggestion
They all have the first parameter as `server` (some people are used to name it as
```

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,278 @@ sidebar: lb4_sidebar
...
+    // Add your logic for stop
+    console.log('create sample observer has stopped.');
+  }
+}
+```
+
+### Setting Order for Migrated Boot Scripts
+
+In LoopBack 3, boot scripts are loaded according to the script file'
+name(alphabetically). LoopBack 4 observer provides an advanced way to control
```
bajtos:
```suggestion
name (alphabetically). LoopBack 4 observer provides an advanced way to control
```

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,278 @@ sidebar: lb4_sidebar
...
+the order of scripts by specifying and setting group names. The tutorial is well
+documented in the section
+[Notify life cycle observers of start/stop related events by order](../Life-cycle.md#notify-life-cycle-observers-of-startstop-related-events-by-order).
+
+## Predefined LoopBack 3 Boot Scripts
+
+LoopBack 3 application has two predefined boot scripts: `/server/boot/root.js`
+and `/server/boot/authentication.js`. Please do not create corresponding
+observers for them because:
```
bajtos: I feel the word "because" is redundant here.

```suggestion
observers for them.
```

examples/todo/data/db.json
```diff
@@ -1,13 +1,18 @@
...
       "1": "{\"title\":\"Take over the galaxy\",\"desc\":\"MWAHAHAHAHAHAHAHAHAHAHAHAHAMWAHAHAHAHAHAHAHAHAHAHAHAHA\",\"id\":1}",
       "2": "{\"title\":\"destroy alderaan\",\"desc\":\"Make sure there are no survivors left!\",\"id\":2}",
       "3": "{\"title\":\"play space invaders\",\"desc\":\"Become the very best!\",\"id\":3}",
-      "4": "{\"title\":\"crush rebel scum\",\"desc\":\"Every.Last.One.\",\"id\":4}"
+      "4": "{\"title\":\"crush rebel scum\",\"desc\":\"Every.Last.One.\",\"id\":4}",
+      "5": "{\"title\":\"a todo sample\",\"desc\":\"Something to do.\",\"id\":5}",
+      "6": "{\"title\":\"a todo sample\",\"desc\":\"Something to do.\",\"id\":6}",
+      "7": "{\"title\":\"a todo sample\",\"desc\":\"Something to do.\",\"id\":7}",
+      "8": "{\"title\":\"a todo sample\",\"desc\":\"Something to do.\",\"id\":8}",
+      "9": "{\"title\":\"a todo sample\",\"desc\":\"Something to do.\",\"id\":9}"
```
bajtos: Please don't forget to revert changes in the sample data too when removing sample observers.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,278 @@ sidebar: lb4_sidebar
...
+   * This method will be invoked when the application starts
+   */
+  async start(): Promise<void> {
+    // Add your logic for start
+    console.log('This is a migrated asynchronous boot script');
+
+    const sample = {title: 'a todo sample', desc: 'Something to do.'};
+    // Create the sample by calling TodoRepo.create()
+    const result = await TodoRepo.create(sample);
+    console.log(`Sample created as: ${result}`);
```
bajtos: AFAICT from the CI logs, this line prints the following output:

```
Sample created as: [object Object]
```

`[object Object]` is not very useful. I am proposing to rewrite this line as follows:

```suggestion
    console.log('Sample created as', result);
```

examples/todo/data/db.json
jannyHou: 🙈 sure, need to clean up them every time run tests on local

docs/site/migration/boot-scripts.md
jannyHou: I see thank you~

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,279 @@ sidebar: lb4_sidebar
...
+
+## LoopBack 3 Boot Script
+
+LoopBack 3 supports synchronous and asynchronous boot scripts. And you can
+choose either callback or promise based implementation for asynchronous scripts.
+They all have the first parameter as `server` (some people are used to name it
+as `app`). The callback based asynchronous script has a 2nd parameter `done` as
+the callback function.
+
+A typical synchronous boot script that prints out a datasource's name before
```
raymondfeng: Maybe say `LoopBack 3 supports three styles of boot scripts` and make each of the examples below a bullet item.

docs/site/migration/boot-scripts.md
```diff
@@ -6,8 +6,279 @@ sidebar: lb4_sidebar
...
+  async stop(): Promise<void> {
+    // Add your logic for stop
+    console.log('create sample observer has stopped.');
+  }
+}
+```
+
+### Setting Order for Migrated Boot Scripts
+
+In LoopBack 3, boot scripts are loaded according to the script file' name
```
nabdelgadir: `file'` -> `file's`?

jannyHou:
@nabdelgadir 

> but just wondering, should there be tests for the new observers added to the Todo example?


This PR will only contain the documentation :) The code changes will be reverted when merge, they are just for demo. See my explanation in the description https://github.com/strongloop/loopback-next/pull/4362#issue-359153879

docs/site/migration/boot-scripts.md
jannyHou: good catch

Issue #4328, opened 2019-12-22
feat(rest): refactoring common filter schemas


achrinza:
Related: https://github.com/strongloop/loopback-next/issues/4290#issuecomment-565000888

Currently, the filter schemas are generated inline in the OAS `openapi.json` document. This may lead to the filter schemas being repeated within the same OAS docs.

## Suggestion

Instead, these schemas should be refactored into `/components/schemas`.

## Use Cases

The OAS document may be more compact and have increased readability as the filter schemas will not be repeated. Furthermore, the generated client-side code may also benefit similarly.

## Examples

(Taken from: https://github.com/strongloop/loopback-next/issues/4290#issuecomment-566683037)
```
{
  "TodoFields": {
    "title": "TodoFields",
    "type": "object",
    "properties": {
      "id": {
        "type": "boolean"
      }
     // ...
    }
  },
  "TodoFilter": {
    "title": "TodoFilter",
    "type": "object",
    "properties": {
      "where": {
        "type": "object"
      },
      "fields": {
        "$ref": "#/components/schemas/TodoFields"
      },
      "offset": {
        "minimum": 0,
        "type": "integer",
        "format": "int32"
      },
      "limit": {
        "minimum": 1,
        "type": "integer",
        "format": "int32",
        "example": 100
      },
      "skip": {
        "minimum": 0,
        "type": "integer",
        "format": "int32"
      },
      "order": {
        "type": "array",
        "items": {
          "type": "string"
        },
        "description": ""
      },
      "include": {
        "type": "array",
        "items": {
          "$ref": "#/components/schemas/Include"
        },
        "description": ""
      }
    }
  }
```

## Acceptance criteria

TBD - will be filled by the team.

dougal83:
@bajtos Here is a [draft gist](https://gist.github.com/dougal83/92b1c5bcc05bd4835861246bc7196e4f).  I need to run code against samples and write tests(*sigh).

Where would be the best to place this code and where to inject it's functionality?  I'll put into PR for review pending your feedback.

jannyHou:
@dougal83 commented in https://github.com/strongloop/loopback-next/issues/4290#issuecomment-569807646 :)

bajtos:
After we land _feat(rest): add openapi schema consolidation_ #4365, I'd like to start the process of deprecating `getModelSchemaRef` and reworking both our existing code and CLI templates to leverage the new schema consolidation:

[...]

Issue #4119, opened 2019-11-13
"error TS5056 overwritten by multiple input files" found after upgraded loopback core to 1.10.6 and typescript to 3.7.2


wayahead:
1. upgraded loopback as below:
```
   "dependencies": {
-    "@loopback/boot": "^1.5.10",
-    "@loopback/context": "^1.23.4",
-    "@loopback/core": "^1.10.6",
-    "@loopback/openapi-v3": "^1.10.0",
-    "@loopback/repository": "^1.15.3",
-    "@loopback/rest": "^1.22.0",
-    "@loopback/rest-explorer": "^1.4.3",
-    "@loopback/service-proxy": "^1.3.10",
+    "@loopback/boot": "^1.5.11",
+    "@loopback/context": "^1.23.5",
+    "@loopback/core": "^1.10.7",
+    "@loopback/openapi-v3": "^1.10.1",
+    "@loopback/repository": "^1.15.4",
+    "@loopback/rest": "^1.23.0",
+    "@loopback/rest-explorer": "^1.4.4",
+    "@loopback/service-proxy": "^1.3.11",
   },
   "devDependencies": {
-    "@loopback/build": "^2.0.15",
-    "@loopback/eslint-config": "^4.1.3",
-    "@loopback/testlab": "^1.9.3",
+    "@loopback/build": "^2.0.16",
+    "@loopback/eslint-config": "^4.1.4",
+    "@loopback/testlab": "^1.9.4",
-    "typescript": "~3.6.4"
+    "typescript": "~3.7.2"
   }
```
I clean and rebuild the project, got the below error:

error TS5056: Cannot write file '/home/wayahead/workspace/loopback/pinwheel/besim/dist/datasources/client.datasource.d.ts' because it would be overwritten by multiple input files.

2. if I roll typescript back 3.6.4, then there is no error. but if checked into @loopback/build/package.json, typescript 3.7.2 is found.

The project and client.datasource.ts was generated by previous version of loopback.

So how to solve this issue caused by loopback upgrading?

raymondfeng:
Either stick with `typescript@~3.6.4` or rename `<name>.datasource.json` to `<name>.datasource.config.json`. The latest version of `@loopback/cli` (1.25.0) already generates datasources with the new convention to deal with https://github.com/microsoft/TypeScript/issues/34761.

PR #4365, opened 2020-01-04
feat(rest): add openapi schema consolidation

dougal83:
Add openapi schema consolidation to rest server as core OAS Enhancer

Implements #4328
See also #4290 (as follows)

For longer term, I would like our REST layer to be smarter and able to automatically extract schemas used in multiple places into `/components/schemas` and replace them with `$ref` at usage points.

This logic can work as follows:
1. When building OpenAPI schema, we need to scan the entire OpenAPI spec tree and look for all known places where a schema object can be provided. Those places include endpoint parameters, request bodies, response schemas, and so one. I think it's ok to start with few commonly-used places first and add more of them later, as we learn more.
2. For each schema object encountered, if the object has a title, then add it to a look-up table (title -> schema) and increase the reference counter.
3. After we scanned the spec tree, iterate over all entries in the look-up table, pick entries with two or more references and move them to `/components/schemas`.
4. Scan the spec tree again, replace all schemas that were moved to `/components/schemas` in step 3 with `$ref`. (I think this could be optimized to avoid the second spec tree scan, but that could be done later.)

_Originally posted by @bajtos in https://github.com/strongloop/loopback-next/issues/4290#issuecomment-565000888_

## Checklist


packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
```diff
@@ -0,0 +1,195 @@
...
+        version: '',
+      },
+      paths: {
+        schema: {
+          $ref: '#/components/schemas/LoopbackExample',
+        },
+      },
+      components: {
+        schemas: {
+          LoopbackExample: {
```
bajtos: What is the reasoning for converting schema titles into different key values (`loopback.example` to `LoopbackExample`)? So far, we are always using the title as the key.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
```diff
@@ -0,0 +1,195 @@
...
+            title: 'Different LoopbackExample exists',
+            properties: {
+              testDiff: {
+                type: 'string',
+              },
+            },
+          },
+          LoopbackExample2: {
+            title: 'loopback.example',
```
bajtos: This a an interesting edge case! I am glad you included it in your pull request.

I am not sure what's the best or the right way how to handle the situation, I guess your proposal is a good start.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
```diff
@@ -0,0 +1,195 @@
...
+            items: {
+              $ref: '#/components/schemas/MyArray',
+            },
+          },
+        },
+      },
+      components: {
+        schemas: {
+          MyArray: {
+            title: 'my.array',
```
bajtos: Let's discuss.

Should we extract the entire array schema in this case? I am guessing that your intention is to support an array of model instances (e.g. the return value of `find` method) by keeping the array definition inside the endpoint spec and moving the item schema to shared components. I see how that can be a better solution. 

When moving the array item schema only, I feel it's confusing to see `MyArray` (or `my.array`) name when the schema describes array items (not the array itself). I am proposing to append `.items` suffix to avoid such confusion.

```suggestion
            title: 'my.array.items',
```

As mentioned in https://github.com/strongloop/loopback-next/pull/4355#discussion_r363785139, when building custom titles, there are few constraints to keep in mind:

- If we don't use any separator, then we can create a naming conflict when the application provides a model with the same name as we have created.
- IIRC, `.` and `-` are not allowed in LoopBack model names, therefore it should be safe to use them as a separator.
- Some OpenAPI consumers (e.g. Microsoft Azure) restrict the subset of characters allowed in schema titles. In particular, Azure requires titles to match the regular expression `^[a-zA-Z0-9.-_]+$` (see https://github.com/strongloop/loopback-next/pull/3504 and https://github.com/strongloop/loopback-next/issues/1466).

packages/openapi-v3/src/consolidate-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+  [schema: string]: SchemaObject | ReferenceObject;
+};
+let $paths: PathsObject;
+
+/*
+ *  Recursively search OpenApiSpec PathsObject for SchemaObjects with title property.
+ *  Move reusable schema bodies to #/components/schemas and replace with json pointer.
+ *  Handles collisions /w title, schema body pair comparision.
+ */
+export function ConsolidateSchemaObjects(spec: OpenApiSpec): OpenApiSpec {
```
bajtos: We use camelCase names for functions.

```suggestion
export function consolidateSchemaObjects(spec: OpenApiSpec): OpenApiSpec {
```

packages/openapi-v3/src/consolidate-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+  $schemaRefs =
```
bajtos: We are avoiding global variables (shared state) in our code base. While they are safe to use as long as the function is synchronous, it makes it easy to accidentally introduce a race condition when a sync function changes to async and the developer forget about the shared state.

Our usual approach is to create some sort of a "context" object that can hold `$schemaRefs`, `$paths` and any other data. This context object is passed as a function argument to lower-level helpers.

Alternatively, you can also create a new class and move schemaRefs & paths to member (instance) properties, treating `this` as the context.

packages/openapi-v3/src/consolidate-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+      : {};
+  $paths = _.cloneDeep(spec.paths);
+
+  recursiveWalk($paths);
+
+  const updatedSpec = {
+    ...spec,
+    ...{
+      paths: $paths,
+      components: {...spec.components, ...{schemas: $schemaRefs}},
```
bajtos: IIUC, this code replaces `spec.components.schemas` with `$schemaRefs`. Is that a safe operation? The name `schemaRefs` implies to me that this object contains only references (schemas extracted by our algorithm?), but what about the original component schemas provided by the user?

packages/openapi-v3/src/consolidate-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (Object.keys(updatedSpec.components).length === 0) {
+    delete updatedSpec.components;
+  }
+
+  return updatedSpec;
+}
+
+function recursiveWalk(rootSchema: ISpecificationExtension) {
+  if (rootSchema !== null && typeof rootSchema == 'object') {
+    Object.entries(rootSchema).map(([key, subSchema]) => {
```
bajtos: `map` is typically used to transform values (`['a', 'b'] -> ['c', 'd']`). Since your callback is not returning any value, you should use `forEach` instead.

```suggestion
    Object.entries(rootSchema).forEach(([key, subSchema]) => {
```

packages/rest/src/rest.server.ts
```diff
@@ -425,7 +426,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
     );
 
     specForm = specForm ?? {version: '3.0.0', format: 'json'};
-    const specObj = this.getApiSpec(requestContext);
+    const specObj = ConsolidateSchemaObjects(this.getApiSpec(requestContext));
```
bajtos: Please perform the consolidation from inside `this.getApiSpec()`. We should provide the same spec regardless of whether it was obtained via REST API or by calling `server.getApiSpec()`.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
dougal83: I was thinking along the lines of the Openapi spec being consumed by a code generator.  Model names generated by picking up the title and PascalCase was prettier.  It's not essential, can drop.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
dougal83: Good idea, will append `.items` as next step.

packages/openapi-v3/src/consolidate-schema.ts
dougal83: > but what about the original component schemas provided by the user?

Original schemas in `spec.components.schemas` cloned to `$schemaRefs` at start.  `$schemaRefs` used as working object then replaces original.  Would it help to rename anything?

packages/openapi-v3/src/consolidate-schema.ts
dougal83: I'll see if I can figure out your usual approach and implement.

jannyHou: FYI - I created a story for adding the oaiEnhancerService in the rest module:
https://github.com/strongloop/loopback-next/issues/4380

Even before #4380 happen, I think you can still create the consolidation as a class that implements the enhancer. And inside `getApiSpec()`, temporarily apply it as:

```ts
// ...other code
const consolidationEnhancer = new ConsolidationEnhancer();
spec = consolidationEnhancer.modifySpec(spec);
// ...
```

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
dougal83: > Should we extract the entire array schema in this case? I am guessing that your intention is to support an array of model instances

It seems like the norm to use reference for items from what I can see with schema transformers.  A particularly great one is https://www.apimatic.io/transformer/, where you can import and export from/to Openapi v3 json and there are some nice things like specifically adding depreciation=false on endpoints.  In an ideal world they'd contribute to loopback-next! :)

packages/openapi-v3/src/consolidate-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+    while (
+      refExists(title) &&
+      !compare(schema as ISpecificationExtension, getRefValue(title), {
+        ignore: ['description'],
+      })
+    ) {
+      i++;
+      title = `${titlePrefix}${i}`;
+    }
+    if (!refExists(title)) {
```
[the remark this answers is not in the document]
dougal83: Discussion time! :) If my logic is sound, I was thinking that in terms of processing what is more expensive considering that the where clause includes a schema comparison(duplicate check). Should we at (L90):

- Check ref key exists, no write if true as is duplicate (current)
- Don't check ref key exists, just (over)write
- (!) Rely entirely on title property being unique, no duplicate/name collision check(preceding code) and write if key does not exist / or just (over)write.

I wasn't sure which path to take.

On another note, would you recommend an alternative object comparison package in place of the json-schema-compare used here?  I thought the ignore feature was neat at the time but not checked alternatives.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
dougal83: Dropped as unable to justify here.

packages/openapi-v3/src/__tests__/unit/consolidate-schema.unit.ts
dougal83: Thought is would be prudent with new enhancers allowing developers to 'get creative'.  Might need to be more edgy in future!

dougal83:
I'm going to go lie down before I hurt myself. 🤦‍♂ Not sure I I rebased correctly?

Let me know how it's looking.

dougal83:
I've moved the code to the enhancers folder, sorry that it kills the file change tracking!  In future will leave file name changes till later.

Please excuse the force push-fest, I was experimenting with git and passing to another machine to run tests.

packages/openapi-v3/src/enhancers/consolidate-schema.enhancer.ts
```diff
@@ -0,0 +1,175 @@
...
+    return updatedSpec;
+  }
+
+  private recursiveWalk(
+    rootSchema: ISpecificationExtension,
+    ctx: ConsolidateContext,
+  ) {
+    if (rootSchema && typeof rootSchema === 'object') {
+      Object.entries(rootSchema).forEach(([key, subSchema]) => {
+        if (subSchema) {
```
dougal83: I might want to check for object type here?  
```ts
if (subSchema && typeof subSchema === 'object') {
```

dougal83:
Hey @jannyHou 

Is the enhancer setup correctly and in the expected file structure?

packages/openapi-v3/src/__tests__/unit/enhancers/consolidate-schema.enhancer.unit.ts
```diff
@@ -0,0 +1,197 @@
...
+const consolidationEnhancer = new ConsolidationEnhancer();
+
+describe('consolidateSchemaObjects', () => {
+  it('moves schema with title to component.schemas, replace with reference', () => {
+    const inputSpec: OpenAPIObject = {
+      openapi: '',
+      info: {
+        title: '',
+        version: '',
+      },
```
bajtos: Please use `OpenApiSpec` builder to avoid dealing with boilerplate required to create a valid spec. See https://github.com/strongloop/loopback-next/tree/master/packages/openapi-spec-builder.

The same comment applies to other similar places in this pull request too.

packages/openapi-v3/src/__tests__/unit/enhancers/consolidate-schema.enhancer.unit.ts
```diff
@@ -0,0 +1,197 @@
...
+  it('If array items has no title, copy parent title if exists', () => {
+    const inputSpec: OpenAPIObject = {
+      openapi: '',
+      info: {
+        title: '',
+        version: '',
+      },
+      paths: {
+        schema: {
+          myarray: {
```
bajtos: IIUC, you are placing the schema directly under the `paths` object. I don't think that's a valid OpenAPI spec:

- According to https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#openapi-object, the top-level object contains `paths` property with `Paths Object` value.
- Paths Object is defined here: https://github.com/OAI/OpenAPI-Specification/blob/3.0.1/versions/3.0.1.md#pathsObject, it is supposed to contain path entries
- Path Item Object can contain HTTP verbs and shared metadata like parameters

Can you please update all tests to use a valid OpenAPI spec document? For example, you can describe a dummy `GET /` endpoint and put the schema to `responses` section.

```ts
const spec = new OpenApiSpecBuilder()
  .withOperation(
    'get', 
    '/', 
    new OperationSpecBuilder().withResponse(200, {
      content: {
        'text/json': {
          schema: {
            // put your schema here
          }
        }
    })
  )
  .build();
```

(Feel free to add new builder methods to further simplify your test code, just make sure to do that in a new commit and/or a new pull request.)

packages/rest/src/rest.server.ts
```diff
@@ -716,6 +717,8 @@ export class RestServer extends Context implements Server, HttpServerLike {
     if (requestContext) {
       spec = this.updateSpecFromRequest(spec, requestContext);
     }
+    const consolidationEnhancer = new ConsolidationEnhancer();
+    spec = consolidationEnhancer.modifySpec(spec);
```
bajtos: Implementing the tricky details in a new class & new source file is the right approach, I love how small the changes in RestServer are now 👏

packages/openapi-v3/src/__tests__/unit/enhancers/consolidate-schema.enhancer.unit.ts
dougal83: Yes, I was shortcutting /w invalid spec based on knowledge of how the crawler operates.  Updated with OpenApiSpecBuilder but [initially patched in `components.schemas`](https://github.com/strongloop/loopback-next/pull/4365/files#diff-5f5f952350cff2452a82b82c113bca9dR57-R72).

dougal83:
Created [alternate commit to merge instead](https://github.com/dougal83/loopback-next/commit/624ae3d6d547ff18702c4555826daf2b7e94b549).  I think the approach may be tidier than current iteration.  Can pull in if preferred to current.

dougal83:
I've simplified the implementation to a core process to work directly on spec.  The preprocessing has been dropped in favour of template change #4119

raymondfeng:
@dougal83 Please rebase against latest master to remove the `merge` commits.

dougal83:
> @dougal83 Please rebase against latest master to remove the `merge` commits.

Reflog is now my favourite feature of git. 👍   The penny finally dropped... `git rebase upstream/master`, that works. 🚀

dougal83:
For reference, test [openapi.json](https://gist.github.com/dougal83/a8dc836de2718b59ca2bb8142e7ad4ed) output.

dougal83:
@bajtos @jannyHou  Thinking about the `title` property... is it used for anything else?  Maybe we can remove it from the resultant spec as it has served its purpose.  Maybe a candidate for an enhancer to remove surplus output... `title` among others like `"additionalProperties": true`, `"x-controller-name": "UserController"` and `"x-operation-name": "count"`.

I might put together a raft of opinionated enhancers for evaluation.

bajtos:
> Thinking about the `title` property... is it used for anything else? Maybe we can remove it from the resultant spec as it has served its purpose.

AFAIK, the `title` property is used by consumers of the spec, for example our own REST API Explorer but also client-code generators like [openapi-codegen](https://github.com/Mermade/openapi-codegen). We have recently landed a pull request to add `title` property to Filter/Where schemas - see #4355

I feel it would be a big step backwards if we removed it.

> Maybe a candidate for an enhancer to remove surplus output... `title` among others like `"x-controller-name": "UserController"` and `"x-operation-name": "count"`.

Please preserve those `x-` fields too, we are adding them intentionally.

> Also, shall I add a fallback in case of catastrophic failure?:

I am not sure what's the best alternative. Either log an error (or a warning) & return the original spec, or abort the app startup by throwing an error. I guess this depends on whether `modifySpec` is called early enough, so that unhandled errors are going to stop the app.

@raymondfeng I'd like to get your thoughts on this feature too, PTAL.

bajtos:
> AFAIK, the `title` property is used by consumers of the spec, for example our own REST API Explorer but also client-code generators like [openapi-codegen](https://github.com/Mermade/openapi-codegen). We have recently landed a pull request to add `title` property to Filter/Where schemas - see #4355

See https://github.com/strongloop/loopback-next/issues/4290:

> Add `title` property to filter schema definition in order to enable better code generation by third party tools such as https://github.com/OpenAPITools/openapi-generator .
>
> Without a title property the generator simply name the filter as `Filter1`, `Filter2`, etc. By adding the title property, the tooling can produce a filter interface named after `title`.

jannyHou:
> Thinking about the title property... is it used for anything else? 

+1 for https://github.com/strongloop/loopback-next/pull/4365#issuecomment-587007019 and `title` is a property inherited from JSON schema, considering that we [convert a model class to JSON schema first](https://github.com/strongloop/loopback-next/tree/master/packages/repository-json-schema) then generate the oas schema based on it, let's keep that property.

packages/openapi-v3/src/enhancers/extensions/consolidate.spec.extension.ts
```diff
@@ -0,0 +1,130 @@
...
+  isSchemaObject,
+  OpenApiSpec,
+  ReferenceObject,
+  SchemaObject,
+} from '../../types';
+import {asSpecEnhancer, OASEnhancer} from '../types';
+
+/**
+ * A spec enhancer to consolidate OpenAPI specs
+ *
```
jannyHou: Could you add more description about the rules? So that users can understand this awesome feature quickly by reading the API doc :) Here is my draft summary(IIUC):

This enhancer consolidates schemas used in multiple places into `/components/schemas` and replaces them with $ref at usage points.

For example, if a schema called  `user` exists in more than one path, it will be consolidated into `/components/schemas/user` and the path objects reference to it as `schema: { $ref: '#/components/schemas/user'}`.

When comparing two schemas, the `description` field is ignored.

packages/openapi-v3/src/__tests__/unit/enhancers/consolidate.spec.extension.unit.ts
```diff
@@ -0,0 +1,166 @@
...
+              test: {
+                type: 'string',
+              },
+            },
+          }),
+      )
+      .build();
+
+    expect(consolidationEnhancer.modifySpec(inputSpec)).to.eql(expectedSpec);
+  });
```
jannyHou: (Feel free to leave it out of the scope of this PR, I know it's exhausting to be blocked by such minor changes :p and appreciate your patience again!)

Maybe add one test for consolidating the same schema that exists multiple times into one `component/schema/<schema_name>`?

```ts
// The input spec
const inputSpec = new OpenApiSpecBuilder()
      .withOperation(
        'get',
        // first time has 'loopback.example' 
        '/path1',
        new OperationSpecBuilder().withResponse(200, {
          description: 'Example',
          content: {
            'application/json': {
              schema: {
                title: 'loopback.example',
                properties: {
                  test: {
                    type: 'string',
                  },
                },
              },
            },
          },
        }),
      )
.withOperation(
        'get',
        // second time has 'loopback.example' 
        '/path2',
        new OperationSpecBuilder().withResponse(200, {
          description: 'Example',
          content: {
            'application/json': {
              schema: {
                title: 'loopback.example',
                properties: {
                  test: {
                    type: 'string',
                  },
                },
              },
            },
          },
        }),
      )
.build()
```

and

```ts
const expectedSpec = new OpenApiSpecBuilder()
      .withOperation(
        'get',
        '/path1',
        new OperationSpecBuilder().withResponse(200, {
          description: 'Example',
          content: {
            'application/json': {
              schema: {
                $ref: '#/components/schemas/loopback.example',
              },
            },
          },
        }),
      )
   .withOperation(
        'get',
        '/path2',
        new OperationSpecBuilder().withResponse(200, {
          description: 'Example',
          content: {
            'application/json': {
              schema: {
                $ref: '#/components/schemas/loopback.example',
              },
            },
          },
        }),
      )
      .withComponents(
        new ComponentsSpecBuilder()
          .withSchema('loopback.example', {
            title: 'loopback.example',
            properties: {
              test: {
                type: 'string',
              },
            },
          }),
      )
      .build();
```

dougal83: Added.  Thank you!

raymondfeng:
@dougal83 Is it possible to introduce a flag for `RestServerConfig` so that applications have a way to opt out consolidation?

dougal83:
@raymondfeng I think it would be a good idea to have a way to disable consolidation.  Omitted for simplicity, but I'd like to add it.

I'll add the feature as it can always be removed on review.

packages/openapi-v3/src/enhancers/extensions/consolidate.spec.extension.ts
```diff
@@ -0,0 +1,157 @@
...
+ *   $ref: '#/components/schemas/loopback.example',
+ * }
+ * ```
+ *
+ * When comparing schemas to avoid naming collisions, the description field
+ * is ignored.
+ */
+@bind(asSpecEnhancer)
+export class ConsolidationEnhancer implements OASEnhancer {
+  name = 'loopback.consolidate.schemas';
```
emonddr: Perhaps make this a constant. Then you can use this constant in your mocha test as well?

packages/rest/src/rest.server.ts
```diff
@@ -1118,6 +1129,12 @@ export interface RestServerResolvedOptions {
    * facade Express application. If not set, the value is default to `true`.
    */
   listenOnStart?: boolean;
+
+  /**
+   * Set this flag to `true` to disable OAS schema consolidation. If not set,
+   * the value defaults to `false`.
+   */
+  disableOASConsolidator?: boolean;
```
raymondfeng: Would it be better to make it part of `OpenApiSpecOptions`?

```ts
export interface OpenApiSpecOptions {
  // ...
  disableConsolidation?: boolean;
}
```

packages/openapi-v3/src/enhancers/extensions/consolidate.spec.extension.ts
dougal83: Currently, I have left as is for consistency with other enhancers.

dougal83:
@raymondfeng I've altered location/binding to be more in line with your info enhancer.  I moved the option as per your instruction.  Happy if you wish to alter/finish/redo implementation if there are glaring mistakes. :+1:

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -375,6 +386,26 @@ describe('RestServer.getApiSpec()', () => {
     expect(spec.info).to.eql(EXPECTED_SPEC_INFO);
   });
 
+  context('OpenApiSpecOptions', () => {
+    it('disables consolidator if disableOASConsolidator is set to true', async () => {
+      const options: {rest: RestServerConfig} = {
+        rest: {openApiSpec: {disableOASConsolidator: true}},
```
bajtos, line 392: Have you considered using a shorter flag name? In my experience, the best names are using the point of view of the consumer (what are they trying to accomplish using the flag), not the particular implementation details (how is the flag implemented). In that light, I am proposing to use the name `consolidate` with the default value `true`.

```suggestion
        rest: {openApiSpec: {consolidate: false}},
```

Thoughts?

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -375,6 +386,26 @@ describe('RestServer.getApiSpec()', () => {
...
+      };
+      options.rest = givenHttpServerConfig(options.rest);
+      app = new Application(options);
+      app.component(RestComponent);
+      server = await app.getServer(RestServer);
+      await server.start();
+
+      const enhancer = (await server.OASEnhancer.getEnhancerByName(
+        'consolidate',
+      )) as ConsolidationEnhancer;
```
bajtos, line 402: I consider explicit code cast as a code smell, can we find a way how to avoid it here?

For example, we can improve `getEnhancerByName` to accept a generic parameter.

```suggestion
    const enhancer = await server.OASEnhancer.getEnhancerByName<
      ConsolidationEnhancer
    >('consolidate');
```

Implementation in `spec-enhancer.service.ts`:

```ts
export class OASEnhancerService {
  // ...

  async getEnhancerByName<T extends OASEnhancer>(
    name: string,
  ): Promise<T | undefined> {
    // Get the latest list of enhancers
    const enhancers = await this.getEnhancers();
    return enhancers.find(e => e.name === name) as T | undefined;
  }
}
```

Thoughts?

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -375,6 +386,26 @@ describe('RestServer.getApiSpec()', () => {
...
+      expect(enhancer.isDisabled).to.be.true();
```
bajtos, line 403: Let's use `disabled` for consistency with other places that are already using this name. See e.g. here:

https://github.com/strongloop/loopback-next/blob/42e3f977181cae288c86ad1b227d6eba5c7e10df/packages/rest/src/rest.server.ts#L1068

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
bajtos: > what are they trying to accomplish using the flag

I.e. the user wants to consolidate the spec, or disable consolidation.

> how is the flag implemented

I.e. OASConsolidator is invoked or skipped

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
dougal83: As always much better with your direction. 👍 I've made the change to getEnhancerByName but couldn't get test to work so fell back to running schema through consolidator.

dougal83:
> Sorry for taking so long to review your updates.

Not a problem, we all have finite bandwidth.

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
bajtos: Maybe the problem was caused by missing default value for the new generic argument?

```ts
export class OASEnhancerService {
  // ...

  async getEnhancerByName<T extends OASEnhancer = OASEnhancer >(
  // ...
```

dougal83: I think the issue is that I was grabbing the class when I wanted an instance, so it wouldn't have worked.  I have left the change (excluding default) to getEnhancerByName in as a commit.  Squashed commits into one ready for merge.

packages/rest/src/spec-enhancers/consolidate.spec-enhancer.ts
```diff
@@ -0,0 +1,182 @@
...
+@bind(asSpecEnhancer, {scope: BindingScope.SINGLETON})
+export class ConsolidationEnhancer implements OASEnhancer {
+  name = 'consolidate';
+  disabled: boolean;
+
+  constructor(
+    @inject(CoreBindings.APPLICATION_CONFIG, {optional: true})
+    readonly config?: ApplicationConfig,
+  ) {
+    this.disabled = !(this.config?.rest?.openApiSpec?.consolidate || true);
```
bajtos, line 70: I think this won't work as expected. When `consolidate` is set to `false, the condition `consolidate || true` will be resolved as `true`. I am surprised this was not caught by the test suite 🤔 

Did you perhaps mean this?

```suggestion
    this.disabled = !(this.config?.rest?.openApiSpec?.consolidate ?? true);
```

bajtos:
@emonddr can you PTAL and check if your comments have been addressed by now?

packages/openapi-v3/src/enhancers/extensions/consolidate.spec.extension.ts
emonddr: ok

dhmlau:
@dougal83, good to merge? I'll leave the honor to you :)

dougal83:
@dhmlau 👍 I just need to address the following (and ensure test completes as expected): https://github.com/strongloop/loopback-next/pull/4365#discussion_r408183778

packages/rest/src/spec-enhancers/consolidate.spec-enhancer.ts
dougal83: Hey @bajtos, that was a magnificent catch!  I did not create adequate tests.  I think I need a bit of help with them.  See [last commit](https://github.com/strongloop/loopback-next/pull/4365/commits/3fe2df25b8b900a6aa7c68dae8bf8b749cc2d677#diff-ed9321f709f68e00fe664ccf139a3ca6R422).

bajtos: Uh oh, I am afraid I don't have bandwidth to look into this in detail. Perhaps one of the OpenAPI owners can help?

Ping @jannyHou (primary owner) and @deepakrkris @emonddr (standby owners).

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -393,6 +408,137 @@ describe('RestServer.getApiSpec()', () => {
...
+        )
+        .build();
+
+      const spec = await server.getApiSpec();
+      expect(spec).to.eql(EXPECTED_SPEC);
+
+      await server.stop();
+    });
+  });
+
```
jannyHou, line 541: I think you can add another test for null/undefined to make sure the default value is honored.

dougal83: Hey, Thanks for taking a look. I couldn't figure out how to get the rest config to apply.  I'm not confident if it is either that that the service is bound incorrectly to the rest server or I just cannot form the test correctly.

packages/rest/src/spec-enhancers/consolidate.spec-enhancer.ts
jannyHou: @dougal83 the nullish coalescing looks correct to me. I see you also added two tests for the flag in https://github.com/strongloop/loopback-next/pull/4365#discussion_r410273651, better to have a 3rd check for the default value `true`.

dougal83: I couldn't figure out why the path is returned as an empty object. Please see error : https://travis-ci.com/github/strongloop/loopback-next/jobs/320173171#L468

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -393,6 +408,137 @@ describe('RestServer.getApiSpec()', () => {
...
+              },
+            },
+          }),
+        )
+        .build();
+
+      const spec = await server.getApiSpec();
+      expect(spec).to.eql(EXPECTED_SPEC);
+
+      await server.stop();
```
jannyHou, line 471: We usually put `server.stop()` in the `afterEach` or `after` hook, thanks for adding it in the test, could you refactor it into the `afterEach` hook? will benefit other tests in this file.

packages/rest/src/spec-enhancers/consolidate.spec-enhancer.ts
```diff
@@ -0,0 +1,182 @@
...
+@bind(asSpecEnhancer, {scope: BindingScope.SINGLETON})
+export class ConsolidationEnhancer implements OASEnhancer {
+  name = 'consolidate';
+  disabled: boolean;
+
+  constructor(
+    @inject(CoreBindings.APPLICATION_CONFIG, {optional: true})
+    readonly config?: ApplicationConfig,
+  ) {
+    this.disabled = !(this.config?.rest?.openApiSpec?.consolidate ?? true);
```
raymondfeng, line 70: I prefer to use `this.disabled = this.config?.rest?.openApiSpec?.consolidate === false`. It is a bit easier to read.

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
```diff
@@ -393,6 +408,137 @@ describe('RestServer.getApiSpec()', () => {
...
+                    },
+                  },
+                },
+              },
+            },
+          }),
+        )
+        .build();
+
+      const spec = await server.getApiSpec();
```
raymondfeng, line 468: @dougal83 This will remove `spec.paths` as we try to build openapi spec from the handler. See https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L800. 

You probably have to add a route with the expected api spec.

raymondfeng: @dougal83 The following code works for me.

1. We need to call server.route() to set up a real route
2. server.route() needs to be called before `start`.

```ts
  context('options', () => {
    it('disables consolidator if consolidate is set to false', async () => {
      const options: {rest: RestServerConfig} = {
        rest: {openApiSpec: {consolidate: false}},
      };
      options.rest = givenHttpServerConfig(options.rest);
      app = new Application(options);
      app.component(RestComponent);
      server = await app.getServer(RestServer);

      const EXPECTED_SPEC = anOpenApiSpec()
        .withOperation(
          'get',
          '/',
          anOperationSpec().withResponse(200, {
            description: 'Example',
            content: {
              'application/json': {
                schema: {
                  title: 'loopback.example',
                  properties: {
                    test: {
                      type: 'string',
                    },
                  },
                },
              },
            },
          }),
        )
        .build();

      server.route('get', '/', EXPECTED_SPEC.paths['/'].get, () => {});

      await server.start();
      const spec = await server.getApiSpec();
      expect(spec).to.eql(EXPECTED_SPEC);

      await server.stop();
    });

    it('runs consolidator if consolidate is set to true', async () => {
      const options: {rest: RestServerConfig} = {
        rest: {openApiSpec: {consolidate: true}},
      };
      options.rest = givenHttpServerConfig(options.rest);
      app = new Application(options);
      app.component(RestComponent);
      server = await app.getServer(RestServer);

      const EXPECTED_SPEC = anOpenApiSpec()
        .withOperation(
          'get',
          '/',
          anOperationSpec().withResponse(200, {
            description: 'Example',
            content: {
              'application/json': {
                schema: {
                  $ref: '#/components/schemas/loopback.example',
                },
              },
            },
          }),
        )
        .withComponents(
          aComponentsSpec().withSchema('loopback.example', {
            title: 'loopback.example',
            properties: {
              test: {
                type: 'string',
              },
            },
          }),
        )
        .build();

      server.route(
        'get',
        '/',
        anOperationSpec()
          .withResponse(200, {
            description: 'Example',
            content: {
              'application/json': {
                schema: {
                  title: 'loopback.example',
                  properties: {
                    test: {
                      type: 'string',
                    },
                  },
                },
              },
            },
          })
          .build(),
        () => {},
      );

      await app.start();
      const spec = await server.getApiSpec();
      expect(spec).to.eql(EXPECTED_SPEC);

      await app.stop();
    });
  });
```

raymondfeng: BTW, this is not really a unit test anymore. We probably should move it to `integration`.

jannyHou: thank you @raymondfeng . In `getApiSpec`, [`spec.paths`](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L800) and [`spec.components.schemas`](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L803) are not preserved, is it intended? 
I remember we discussed this before, if the paths and schemas should be kept I can submit a PR to fix it. 
(the discussion here is not related to PR I can open a new issue if we need more opinions)

jannyHou: For the rest configurations:
the unit tests are added in [__tests__/unit/rest.server/rest.server.unit.ts](https://github.com/strongloop/loopback-next/blob/a81ce7e1193f7408d30d984d0c3ddcec74f7c316/packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts#L78-L94)
and integration tests are in [__tests__/integration/rest.server.integration.ts](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/__tests__/integration/rest.server.integration.ts#L851)
links point to the existing options tests.

dougal83: @raymondfeng Ah OK, I didn't understand the getApiSpec() implementation.  Could you please take a look at the code below? I want to test the rest config is being applied.

```ts
     server.api(INPUT_SPEC);

      server.OASEnhancer.spec = await server.get<OpenApiSpec>(
        RestBindings.API_SPEC,
      );
      await server.OASEnhancer.applyEnhancerByName('consolidate');
      const spec = await server.get<OpenApiSpec>(RestBindings.API_SPEC);
      expect(spec).to.eql(INPUT_SPEC);
```

dougal83:
@jannyHou Thanks for pointers.  Could you please check changes to `packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts` and if it is OK then I think we are good to merge pending tests passing. 👍

dougal83:
I've just noticed comments I've missed via notifications. Will revisit pr tomorrow to double check. (I've not ignored them)

packages/rest/src/__tests__/unit/rest.server/rest.server.open-api-spec.unit.ts
dougal83: @raymondfeng Thank you very much for your guidance! 👍 

@jannyHou Thank you for your help!  Could you please check that the tests are located in the correct place now?  I've no formal training on tests so I'm really just hacking.

dougal83:
> Great effort and patience. Please squash all commits into one before merge.

@raymondfeng @bajtos Quick query.  Should I drop the following commit as it was not used? a0a6c22bb25e2f81e0fcf17e601e00d2a005008a

raymondfeng:
I would keep it.

Issue #4209, opened 2019-11-26
bug(cli): cli does not respect custom relation getter name


derdeka:
## Steps to reproduce

Generating `hasMany` relation between `News` and `NewsFile` object. Naming it `files` instead of the default `newsFiles`:

![image](https://user-images.githubusercontent.com/13640166/69612163-76bfa580-102f-11ea-9107-f3d26e57ee83.png)

@loopback/cli@1.26.0

## Current Behavior

The custom name `files` is used in the model, but not in the repository:

![image](https://user-images.githubusercontent.com/13640166/69612343-c900c680-102f-11ea-930a-fc1c129d6a11.png)

![image](https://user-images.githubusercontent.com/13640166/69612417-ea61b280-102f-11ea-8198-809d9099bfa1.png)

## Expected Behavior

Expecting consistent naming across model and repository.

achrinza:
Correct me if I'm wrong but I believe this is expected behaviour. The CLI prompts for setting the `property` name (not the relation name), which in this case works fine.

derdeka:
@achrinza if this is the expected behavior, maybe there should be another prompt for customizing the naming in the repository?

achrinza:
Just curious, what’s the use case for custom relation names?

________________________________

derdeka:
> Just curious, what’s the use case for custom relation names?

I have a model inheritance like `NewsFile extends BaseFile` and `ContractFile extends BaseFile` to have shared properties and some additional custom properties per subclass - each subclass is storing in it's own relational database table. On the api it's just named `files`.

achrinza:
@derdeka Thanks for providing extra info. I think this feature is a possibility, though there are [limitations on using class inheritance](https://github.com/strongloop/loopback-next/issues/4042#issuecomment-549801710).

achrinza:
I've dug into the code and docs and found some inconsistencies:

1. `--relationName` option is used for setting the property ([see here](https://loopback.io/doc/en/lb4/Relation-generator.html))
2. Internally, the function responsible referred to as a "promptRelationName" ([see here](https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/relation/index.js#L339-L382))
3. Internally, the constant storing the prompt message is called ([see here](https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/relation/index.js#L30-L31))
4. The end result is the foreign key property name being affected and not the relations name.

achrinza:
[...]

Issue #3287, opened 2019-07-01
lb4 relation should discard changes when something failed


ErwSim:
## Suggestion

When something failed (ie: no pk or property already exists), changes that have been applied should be discarded

## Use Cases

It would permit to avoid reverting manually changes since the command has failed

## Examples

lb4 relation
  > fill
  > an error has occured
  > should be revert back

## Acceptance criteria

- [x] Do not generate any code if the the generation fails or gets aborted.
- [x] stretch goal: warn users earlier about missing Repository class (perhaps at the point right after asking for target and source model). 
- [ ] stretch goal: warn users when they are using customized FK name. 
     > - use customized foreign key name for belongsTo relation ( myCustomer in this case).
     > - use customized source property name for hasMany relation (e.g myOrder instead of orders).
because our the CLI doesn't support customizing names atm.
- [ ] Update documentation if there is behaviour change for prompts.

dhmlau:
I've opened #4122 without knowing an existing issue. Cross posting here.

---

When running `lb4 relation` (for example following the [TodoList tutorial](https://loopback.io/doc/en/lb4/todo-list-tutorial.html)), it complained that there is no Repository class. However, in the `@hasMany` decorator has been added to todo-list.model.ts, possibly some other code had changed too.  


## Steps to reproduce
1. Run `lb4 example todo` to get the todo example 
2. Create the `TodoList` model according to https://loopback.io/doc/en/lb4/todo-list-tutorial-model.html#building-a-checklist-for-your-todo-models
3. Run `lb4 relation` command. It complained because I don't have `todo-list.repository.ts`. 
```
$ lb4 relation
? Please select the relation type hasMany
? Please select source model TodoList
? Please select target model Todo
? Foreign key name to define on the target model todoListId
? Source property name for the relation getter todos
? Allow TodoList queries to include data from related Todo instances? Yes
   create src/controllers/todo-list-todo.controller.ts
Generation is aborted: Error: File not found: /Users/dhmlau/loopback4-playground/loopback4-example-todo/src/repositories/todo-list.repository.ts
```
 
However, I noticed that in `todo-list.model.ts`, the `@hasMany` decorator has been added.
```
  @hasMany(() => Todo)
  todos: Todo[];
```




## Current Behavior



There were code changed even the command says it failed. 

## Expected Behavior


There should be no code change when the command failed to execute. 
[...]

PR #4373, opened 2020-01-06
fix(cli): discard changes if lb4 `relation` fails. Allow customized relation name for belongsTo

agnes512:
Fixes https://github.com/strongloop/loopback-next/issues/3287 and https://github.com/strongloop/loopback-next/issues/4209

1st commit: code changes
2nd commit: tests

#### Example: 

- **hasMany**
? SourceModel: `Customer`
? DestModel: `Order`,
? Foreign key to define on the target model: `customerId`
? Source property name for the relation getter: `orders
? Allow Order queries to include data from related Customer instances? Yes
==> issues:
  - if generation gets aborted, part of code still gets generated in Model and Repository 
  - the default value doesn't match with the code in `/repository/relation`

- **belongTo**
? SourceModel: `Order`
? DestModel: `Customer`,
? Source property name for the relation getter: `customerId`
? Allow ..inclusion? Yes
==> issues:
  - if generation gets aborted, part of code still gets generated in Model and Repository
  - the default value doesn't match with the code in `/repository/relation` 
  - generated code failed if the foreign key name is not the default value
  - generator doesn't respect custom relation name, the only relation it generates is `targetModelName` (e.g `customer`)

## New changes

- checks if corresponding repository exists
- checks if pk exists and validates the name
- checks if fk exists and validates the name
- asks fk for both hasMany and belongsTo
- asks relation name for both hasMany and belongsTo
- checks if the relation name exists
- allow custom relation name for belongsTo
  - This fixes https://github.com/strongloop/loopback-next/issues/4209
  - customized relation name will be specified in the relation metadata

#### Examples of new prompt:

- **hasMany**
? Please select the relation type `hasMany`
? Please select source model `Customer`
? Please select target model `Order`
? Foreign key name to define on the _target_ model `c_id` // not default value
? Source property name for the relation getter (will be the relation name) `orders`
? Allow Customer queries to include data from related Review instances? Yes

=>
```
// Customer property:
  @hasMany(() => Order, {keyTo: 'c_id'})  // custom fk
  orders: Order[];

// Order fk:
  @property({
    type: 'number',
  })
  c_id?: number;  // custom fk
```

- **belongsTo**
? Please select the relation type `belongsTo`
? Please select source model `Order`
? Please select target model `Customer`
? Foreign key name to define on the _source_ model `c_id`
? Relation name my_customer   // not default name
? Allow Customer queries to include data from related Review instances? Yes

=> 
```
//Order fk:
  @belongsTo(() => Customer, {name: 'my_customer'}) . // allow custom relation name
  c_id: number;    // allow custom fk name
```

For belongsTo, if the fk name is the same as the relation name, the prompt would ask the user to re-enter a new name to avoid a known issue: [Relations with custom names not working ](https://github.com/strongloop/loopback-next/issues/4392)

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -43,6 +43,11 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+LB4 also provides an CLI tool `lb4 relation` to generate `belongsTo` relation
+for you. Before you check out the
+[`Relation Generator`](https://loopback.io/doc/en/lb4/Relation-generator.html)
+page, read on to learn how you can define relations to meet your requirements.
+
 ### Relation Metadata
 
 LB4 uses three `keyFrom`, `keyTo` and `name` fields in the `belongsTo` relation
--- a/docs/site/Relation-generator.md
+++ b/docs/site/Relation-generator.md
@@ -32,7 +32,8 @@ lb4 relation [options]
 - `--relationType`: Relation type.
 - `--sourceModel`: Source model.
 - `--destinationModel`: Destination model.
-- `--foreignKeyName`: Destination model foreign key name.
+- `--foreignKeyName`: Destination/Source model foreign key name for
+  HasMany/BelongsTo relation, respectively.
 - `--relationName`: Relation name.
 - `-c`, `--config`: JSON file name or value to configure options.
 - `-y`, `--yes`: Skip all confirmation prompts with default or provided value.
@@ -88,18 +89,31 @@ The tool will prompt you for:
 - **Name of the `target` model.** _(targetModel)_ Prompts a list of available
   models to choose from as the target model of the relation.
 
-- **Name of the `Source property`.** _(relationName)_ Prompts for the Source
-  property name. Note: Leave blank to use the default.
+- **Name of the `foreign key`.** _(relationName)_ Prompts a property name that
+  references the primary key property of the another model. Note: Leave blank to
+  use the default.
+
+  Default values: `<foreignKeyName>` + `Id` in camelCase, e.g `categoryId`
+
+- **Name of the `relation`.** _(relationName)_ Prompts for the Source property
+  name. Note: Leave blank to use the default.
 
   Default values:
 
-  - `<targetModel><targetModelPrimaryKey>` for `belongsTo` relations, e.g.
-    `categoryId`
   - plural form of `<targetModel>` for `hasMany` relations, e.g. `products`
+  - Based on the foreign key for `belongsTo` relations, e.g. when the foreign
+    key is the default name, e.g `categoryId`, the default relation name is
+    `category`.
 
-- **Name of Foreign key** _(foreignKeyName)_ to be created in target model. For
-  hasMany relation type only, default: `<sourceModel><sourceModelPrimaryKey>`.
-  Note: Leave blank to use the default.
+{% include warning.html content="Based on your input, the default foreign key name might be the same as the default relation name, especially for belongsTo relation. Please name them differently to avoid a known issue [Navigational Property Error](https://github.com/strongloop/loopback-next/issues/4392)
+" lang=page.lang %}
+
+The generator has some limitations. It only asks the most basic factors for
+generating relations. For example, it uses the id property as the source key.
+LB4 allows you customize the source key name, the foreign key name, the relation
+name, and even the DB column name. Check the
+[Relation Metadata](HasMany-relation.md#relation-metadata) section in each
+relation to learn how you can define relations.
 
 ### Output
 
--- a/packages/cli/generators/relation/base-relation.generator.js
+++ b/packages/cli/generators/relation/base-relation.generator.js
@@ -40,11 +40,11 @@ module.exports = class BaseRelationGenerator extends ArtifactGenerator {
 
   async generateAll(options) {
     this._setupGenerator();
-    await this.generateControllers(options);
-    this._setupGenerator();
     await this.generateModels(options);
     this._setupGenerator();
     await this.generateRepositories(options);
+    this._setupGenerator();
+    await this.generateControllers(options);
   }
 
   async generateControllers(options) {
@@ -98,13 +98,8 @@ module.exports = class BaseRelationGenerator extends ArtifactGenerator {
       type: this._getRepositoryRelationPropertyType(),
     };
 
-    if (relationUtils.doesPropertyExist(classDeclaration, property.name)) {
-      throw new Error(
-        'property ' + property.name + ' already exist in the repository.',
-      );
-    } else {
-      relationUtils.addProperty(classDeclaration, property);
-    }
+    // already checked the existence of property before
+    relationUtils.addProperty(classDeclaration, property);
   }
 
   _addParametersToRepositoryConstructor(classConstructor) {
@@ -112,9 +107,8 @@ module.exports = class BaseRelationGenerator extends ArtifactGenerator {
       utils.camelCase(this.artifactInfo.dstRepositoryClassName) + 'Getter';
 
     if (relationUtils.doesParameterExist(classConstructor, parameterName)) {
-      throw new Error(
-        'Parameter ' + parameterName + ' already exist in the constructor.',
-      );
+      // no need to check if the getter already exists
+      return;
     }
 
     classConstructor.addParameter({
--- a/packages/cli/generators/relation/belongs-to-relation.generator.js
+++ b/packages/cli/generators/relation/belongs-to-relation.generator.js
@@ -73,12 +73,16 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator 
   }
 
   async generateModels(options) {
+    // for repo to generate relation name
+    this.artifactInfo.relationName = options.relationName;
     const modelDir = this.artifactInfo.modelDir;
     const sourceModel = options.sourceModel;
 
     const targetModel = options.destinationModel;
     const relationType = options.relationType;
     const relationName = options.relationName;
+    const defaultRelationName = options.defaultRelationName;
+    const foreignKeyName = options.foreignKeyName;
     const fktype = options.destinationModelPrimaryKeyType;
 
     const project = new relationUtils.AstLoopBackProject();
@@ -88,10 +92,16 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator 
       sourceModel,
     );
     const sourceClass = relationUtils.getClassObj(sourceFile, sourceModel);
+    // this checks if the foreign key already exists, so the 2nd param should be foreignKeyName
+    relationUtils.doesRelationExist(sourceClass, foreignKeyName);
 
-    relationUtils.doesRelationExist(sourceClass, relationName);
-
-    const modelProperty = this.getBelongsTo(targetModel, relationName, fktype);
+    const modelProperty = this.getBelongsTo(
+      targetModel,
+      relationName,
+      defaultRelationName,
+      foreignKeyName,
+      fktype,
+    );
 
     relationUtils.addProperty(sourceClass, modelProperty);
     const imports = relationUtils.getRequiredImports(targetModel, relationType);
@@ -101,10 +111,32 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator 
     await sourceFile.save();
   }
 
-  getBelongsTo(className, relationName, fktype) {
+  getBelongsTo(
+    className,
+    relationName,
+    defaultRelationName,
+    foreignKeyName,
+    fktype,
+  ) {
+    // checks if relation name is customized
+    let relationDecorator = [
+      {
+        name: 'belongsTo',
+        arguments: [`() =>  ${className}`],
+      },
+    ];
+    // already checked if the relation name is the same as the source key before
+    if (defaultRelationName !== relationName) {
+      relationDecorator = [
+        {
+          name: 'belongsTo',
+          arguments: [`() =>  ${className}, {name: '${relationName}'}`],
+        },
+      ];
+    }
     return {
-      decorators: [{name: 'belongsTo', arguments: [`() =>  ${className}`]}],
-      name: relationName,
+      decorators: relationDecorator,
+      name: foreignKeyName,
       type: fktype,
     };
   }
@@ -122,7 +154,7 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator 
   }
 
   _getRepositoryRelationPropertyName() {
-    return utils.camelCase(this.artifactInfo.dstModelClass);
+    return this.artifactInfo.relationName;
   }
 
   _initializeProperties(options) {
@@ -140,22 +172,22 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator 
   }
 
   _addCreatorToRepositoryConstructor(classConstructor) {
-    const relationPropertyName = this._getRepositoryRelationPropertyName();
+    const relationName = this.artifactInfo.relationName;
     const statement =
-      `this.${relationPropertyName} = ` +
+      `this.${relationName} = ` +
       `this.createBelongsToAccessorFor('` +
-      `${this.artifactInfo.relationName.replace(/Id$/, '')}',` +
+      `${relationName}',` +
       ` ${utils.camelCase(this.artifactInfo.dstRepositoryClassName)}` +
       `Getter,);`;
     classConstructor.insertStatements(1, statement);
   }
 
   _registerInclusionResolverForRelation(classConstructor, options) {
-    const relationPropertyName = this._getRepositoryRelationPropertyName();
+    const relationName = this.artifactInfo.relationName;
     if (options.registerInclusionResolver) {
       const statement =
         `this.registerInclusionResolver(` +
-        `'${relationPropertyName}', this.${relationPropertyName}.inclusionResolver);`;
+        `'${relationName}', this.${relationName}.inclusionResolver);`;
       classConstructor.insertStatements(2, statement);
     }
   }
--- a/packages/cli/generators/relation/has-many-relation.generator.js
+++ b/packages/cli/generators/relation/has-many-relation.generator.js
@@ -74,6 +74,8 @@ module.exports = class HasManyRelationGenerator extends BaseRelationGenerator {
   }
 
   async generateModels(options) {
+    // for repo to generate relation name
+    this.artifactInfo.relationName = options.relationName;
     const modelDir = this.artifactInfo.modelDir;
     const sourceModel = options.sourceModel;
 
@@ -166,7 +168,7 @@ module.exports = class HasManyRelationGenerator extends BaseRelationGenerator {
   }
 
   _getRepositoryRelationPropertyName() {
-    return utils.pluralize(utils.camelCase(this.artifactInfo.dstModelClass));
+    return this.artifactInfo.relationName;
   }
 
   _getRepositoryRelationPropertyType() {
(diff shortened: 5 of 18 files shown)


packages/cli/generators/relation/belongs-to-relation.generator.js
```diff
@@ -101,10 +111,36 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator
...
+  ) {
+    // checks if relation name is customized
+    let relationDecorator = [
+      {
+        name: 'belongsTo',
+        arguments: [`() =>  ${className}`],
+      },
+    ];
+    // except the customized relation name, also need to check if the property name is the same as relation name
+    // or it will get 'navigational property' error when creating instances
```
agnes512: related [stackoverflow question](https://stackoverflow.com/questions/59468329/unhandled-error-in-post-order-products-500-error-navigational-properties-are).

packages/cli/generators/relation/index.js
```diff
@@ -267,24 +314,69 @@ module.exports = class RelationGenerator extends ArtifactGenerator {
...
+      // should check the existance of fk here.
+      // default source key for belongsTo
+      this.artifactInfo.defaultSourceKeyName =
+        utils.camelCase(this.artifactInfo.destinationModel) +
+        utils.toClassName(this.artifactInfo.destinationModelPrimaryKey);
+
+      return this.prompt([
+        {
+          type: 'string',
+          name: 'sourceKeyName',
```
agnes512: use `sourceKeyName`, `relationName`, and `defaultRelationName` to check if the source key of belongsTo is customized. These three properties also check if the `name` field of the relation metadata needs to be specified.

packages/cli/generators/relation/index.js
```diff
@@ -338,50 +430,64 @@ module.exports = class RelationGenerator extends ArtifactGenerator {
...
     if (this.options.relationName) {
       debug(
         `Relation name received from command line: ${this.options.relationName}`,
       );
       this.artifactInfo.relationName = this.options.relationName;
     }
+    this.artifactInfo.defaultRelationName = this._getDefaultRelationName();
+    // for hasMany && hasOne, the source key is the same as the relation name
+    const msg =
+      this.artifactInfo.relationType === 'belongsTo'
```
agnes512: This property name will be the source key and the relation name for hasMany and hasOne. So I separate it from belongsTo here.

agnes512:
@derdeka @achrinza could you take a look? part of code fixes https://github.com/strongloop/loopback-next/issues/4209. Thanks.

packages/cli/generators/relation/base-relation.generator.js
```diff
@@ -98,23 +98,17 @@ module.exports = class BaseRelationGenerator extends ArtifactGenerator {
...
     };
 
-    if (relationUtils.doesPropertyExist(classDeclaration, property.name)) {
-      throw new Error(
-        'property ' + property.name + ' already exist in the repository.',
-      );
-    } else {
-      relationUtils.addProperty(classDeclaration, property);
-    }
+    // already chcked the existence of property before
```
bajtos: Typo.

```suggestion
    // already checked the existence of property before
```

packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
```diff
@@ -255,7 +259,7 @@ export class Order extends Entity {
   name?: string;
 
   @belongsTo(() => Customer)
-  myRelation: number;
+  customerId: number;
```
bajtos: Is this change intentional and legitimate? I have no idea, just checking to ensure we don't accidentally overlook a regression.

agnes512: This test aims to check if the pre-exist property `@property{} something;` will be overwritten by `@belongsTo` decorator. That's why I use the default relation name to make the intention more clear.

packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
```diff
@@ -46,12 +50,12 @@ export class OrderClassRepository extends DefaultCrudRepository<
   typeof OrderClass.prototype.orderNumber
 > {
 
-  public readonly customerClass: BelongsToAccessor<CustomerClass, typeof OrderClass.prototype.orderNumber>;
+  public readonly customerClassCustNumber: BelongsToAccessor<CustomerClass, typeof OrderClass.prototype.orderNumber>;
```
jannyHou: hmm what does `Cust` mean 🤔 ?

packages/cli/test/integration/generators/relation.has-many.integration.js
```diff
@@ -96,6 +98,38 @@ describe('lb4 relation HasMany', function() {
     );
   });
 
+  context('Execute relation with exist relation name', () => {
```
jannyHou: nitpick: `existing`

packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
agnes512: `CustNumber` is the name of the id property of `customerClass` model.

packages/cli/generators/relation/index.js
```diff
@@ -338,50 +430,64 @@ module.exports = class RelationGenerator extends ArtifactGenerator {
...
+      this.artifactInfo.srcRepositoryFileObj = new relationUtils.AstLoopBackProject().addExistingSourceFile(
+        this.artifactInfo.srcRepositoryFile,
+      );
+
+      const repoClassDeclaration = this.artifactInfo.srcRepositoryFileObj.getClassOrThrow(
+        this.artifactInfo.srcRepositoryClassName,
+      );
+      // checks if the relation name already exists in repo
+      if (
+        relationUtils.doesPropertyExist(
```
jannyHou: Just curious: I think the relation definition is in the model file, why do we have to use more steps to get the corresponding repository file and check the existence?

packages/cli/generators/relation/index.js
```diff
@@ -267,24 +314,69 @@ module.exports = class RelationGenerator extends ArtifactGenerator {
...
+      this.artifactInfo.defaultSourceKeyName =
+        utils.camelCase(this.artifactInfo.destinationModel) +
+        utils.toClassName(this.artifactInfo.destinationModelPrimaryKey);
+
+      return this.prompt([
+        {
+          type: 'string',
+          name: 'sourceKeyName',
+          message: PROMPT_MESSAGE_SOURCE_KEY_NAME,
+          when: this.artifactInfo.relationName === undefined,
```
jannyHou: nitpick: `when: !this.artifactInfo.relationName`

packages/cli/test/integration/generators/relation.has-many.integration.js
emonddr: with `existing` relation

packages/cli/test/integration/generators/relation.has-many.integration.js
```diff
@@ -96,6 +98,38 @@ describe('lb4 relation HasMany', function() {
...
+                SourceEntries.CustomerModel,
+                SourceEntries.OrderModel,
+                SourceEntries.CustomerRepository,
+                SourceEntries.OrderRepository,
+              ],
+            }),
+          )
+          .withPrompts(prompt),
+      ).to.be.rejectedWith(
+        `relation myCustomer already exist in the repository OrderRepository.`,
```
emonddr: already `exists`

emonddr:
The relation-based test cases pass, so that's great.
Some coverage fell though :S

packages/cli/generators/relation/index.js
agnes512: Good question! From my understanding the model has _the property_ (e.g `orders` or `customerId`). And the repository has _the relation name_ (e.g `orders` or `customer`). They might have the same name but most of time they don't. For example, for belongsTo relation, the relation name is `customer`.

The generator throws errors if the repository already has the relation name defined. And the generator didn't have such checks before .

packages/cli/generators/relation/belongs-to-relation.generator.js
```diff
@@ -140,22 +176,22 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator
...
 
   _addCreatorToRepositoryConstructor(classConstructor) {
-    const relationPropertyName = this._getRepositoryRelationPropertyName();
+    const relationName = this.artifactInfo.relationName;
     const statement =
-      `this.${relationPropertyName} = ` +
+      `this.${relationName} = ` +
       `this.createBelongsToAccessorFor('` +
-      `${this.artifactInfo.relationName.replace(/Id$/, '')}',` +
+      `${relationName}',` +
```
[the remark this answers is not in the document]
agnes512: I have `relationName` defined above so we don't need it here anymore.

packages/cli/generators/relation/has-many-relation.generator.js
```diff
@@ -166,7 +168,7 @@ module.exports = class HasManyRelationGenerator extends BaseRelationGenerator {
   }
 
   _getRepositoryRelationPropertyName() {
-    return utils.pluralize(utils.camelCase(this.artifactInfo.dstModelClass));
+    return this.artifactInfo.relationName;
```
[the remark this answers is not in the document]
agnes512: I have relationName defined above so we can just use the value here.

packages/cli/generators/relation/index.js
```diff
@@ -134,6 +156,16 @@ module.exports = class RelationGenerator extends ArtifactGenerator {
...
     }
+    let repoList;
+    try {
+      debug(`repository list dir ${this.artifactInfo.repoDir}`);
+      repoList = await utils.getArtifactList(
+        this.artifactInfo.repoDir,
+        'repository',
+      );
+    } catch (err) {
+      return this.exit(err);
```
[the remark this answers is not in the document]
agnes512: We use exit for generators

agnes512:
> The relation-based test cases pass (i ran them on my laptop), so that's great. Doesn't appear to be any regressions.
> Some coverage fell though :S

If you check the tests It is caused by the generator every time it does `exit(err)`, not the actual code itself. I don't any way to skip it. If anyone knows please let me know :smile:

agnes512:
> Just curious; do we also need to update the docs that reference the CLI prompt procedure?
> 
> See:
> 
> * https://loopback.io/doc/en/lb4/Relation-generator.html
> * https://loopback.io/doc/en/lb4/todo-list-tutorial-relations.html

Yes, I will add them once the proposal looks good to most of you :smile:

packages/cli/generators/relation/index.js
jannyHou: >  And the repository has the relation name

Oh that's smart! 👍 makes perfect sense then

packages/cli/generators/relation/base-relation.generator.js
jannyHou: (sorry I am not very familiar with the relation cli which results in many questions ❔ ❔ ) Hmm, where is it checked? did a quick search of code before it but didn't find any clue 🙈

packages/cli/generators/relation/base-relation.generator.js
```diff
@@ -98,23 +98,17 @@ module.exports = class BaseRelationGenerator extends ArtifactGenerator {
...
 
   _addParametersToRepositoryConstructor(classConstructor) {
     const parameterName =
       utils.camelCase(this.artifactInfo.dstRepositoryClassName) + 'Getter';
 
     if (relationUtils.doesParameterExist(classConstructor, parameterName)) {
-      throw new Error(
-        'Parameter ' + parameterName + ' already exist in the constructor.',
-      );
+      // no need to check if the getter already exists
```
jannyHou: Doesn't parameter exist imply the relation already exists? And therefore an error should be thrown like the deleted code?

packages/cli/generators/relation/belongs-to-relation.generator.js
```diff
@@ -101,10 +111,32 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator
...
+        name: 'belongsTo',
+        arguments: [`() =>  ${className}`],
+      },
+    ];
+    // already checked if the relation name is the same as the source key before
+    if (defaultRelationName !== relationName) {
+      relationDecorator = [
+        {
+          name: 'belongsTo',
+          arguments: [`() =>  ${className}, {name: '${relationName}'}`],
```
jannyHou: I am a bit confused about `defaultRelationName` and `sourceKeyName`, wouldn't they be the same? 
I find it hard to think of a user scenario that `defaultRelationName`, `sourceKeyName` and `relationName` are all different...

packages/cli/generators/relation/belongs-to-relation.generator.js
```diff
@@ -122,7 +154,7 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator
   }
 
   _getRepositoryRelationPropertyName() {
-    return utils.camelCase(this.artifactInfo.dstModelClass);
+    return this.artifactInfo.relationName;
```
jannyHou: I guess this line fixes https://github.com/strongloop/loopback-next/issues/4209?

packages/cli/generators/relation/index.js
```diff
@@ -23,14 +23,19 @@ const ERROR_SOURCE_MODEL_PRIMARY_KEY_DOES_NOT_EXIST =
   'Source model primary key does not exist.';
 const ERROR_DESTINATION_MODEL_PRIMARY_KEY_DOES_NOT_EXIST =
   'Target model primary key does not exist.';
+const ERROR_REPOSITORY_DOES_NOT_EXIST =
+  'class does not exist. Please create repositories first with "lb4 repository" command.';
```
jannyHou: nitpick: `create **repository**`, instead of `create **repositories**`

packages/cli/generators/relation/base-relation.generator.js
agnes512: Here the property is the relation name, which has been checked in [`index file`](https://github.com/strongloop/loopback-next/pull/4373/files#diff-d9d0cb4d78404ced50ce6c783a09ccbaR488). Sorry the PR is big, I should've separated them.

packages/cli/generators/relation/base-relation.generator.js
agnes512: The param is getter, such as `@repository.getter('OrderRepository') protected orderRepositoryGetter: Getter<OrderRepository>`. And I don't think this implies a certain _relation name_ already exists. That's why I removed it. 

And I was also thinking about some edge cases. For example, since we allow customized `keyFrom`, user could have two _same_ relations with 2 models with two different names. In this case, they will share the same getter.

packages/cli/generators/relation/belongs-to-relation.generator.js
agnes512: Good point. I will list out the differences:

Before, the generator asked `foreign key`, `relationName` for **hasMany**.  
 and it only asked `relationName` for **belongTo** if `defaultRelationName` doesn't exist. It used `relationName` as the foreign key name (e.g `customerId`). Then it use the target model as the _real relation name_ (e.g `customer).

This is why how it dealt with **belongsTo** is problematic:
1. if the default foreign key already exists, it didn't prompt
2. if the default foreign key doesn't exists, it takes whatever it is, but the _real relation name_ is still the target model, and this is different from how the `belongsToAccessor` generates the default name. For example: if you type in `customer_id` as your foreign key, the generator would have `customer` as the default relation name. But for the `belongsToAccessor`, the default relation name is `customer_id` instead of `customer`. That's why users are getting `Cannot find type of undefined` errors. 

In order to allow custom fk and relation name for belongsTo, I added a new field `sourceKeyName` to represent the foreign key of belongsTo. And `relationName` to represent the _real relation name_. 

The line `defaultRelationName !== relationName` here is to check if the relationName is customized. If so, it should be specified in the `name` field, just like how we documented in our docs [Relation Metadata](https://loopback.io/doc/en/lb4/BelongsTo-relation.html#relation-metadata).

Therefore, it can have correct property:
```
  @belongsTo(() => Customer, {name: 'my_customer'}) // customized relation name
  customer_id: number; // customized fk
```

packages/cli/generators/relation/belongs-to-relation.generator.js
agnes512: See comment https://github.com/strongloop/loopback-next/pull/4373/files#r365448472. This line is part of fix.

packages/cli/lib/utils.js
```diff
@@ -140,6 +140,84 @@ exports.validateNotExisting = function(projDir) {
   return true;
 };
 
+/**
+ * validate source key or foreign key for relations
+ */
```
nabdelgadir: Call `/* istanbul ignore next */` here instead?

packages/cli/test/fixtures/relation/index.js
```diff
@@ -83,12 +83,24 @@ const SourceEntries = {
     content: readSourceFile('./repositories/order-class-type.repository.ts'),
   },
 
-  NoKeyModel: {
+  NokeyModel: {
```
nabdelgadir: I think this should stay as `NoKeyModel`

packages/cli/generators/relation/belongs-to-relation.generator.js
```diff
@@ -140,22 +172,22 @@ module.exports = class BelongsToRelationGenerator extends BaseRelationGenerator
   }
 
   _addCreatorToRepositoryConstructor(classConstructor) {
-    const relationPropertyName = this._getRepositoryRelationPropertyName();
+    const relationName = this.artifactInfo.relationName;
```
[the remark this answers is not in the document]
agnes512: Yes, just personal preference as I have it defined above.

packages/cli/test/fixtures/relation/index.js
agnes512: I want to make it consistent as the model name is `Nokey` in `nokey.model.ts`.

packages/cli/generators/relation/base-relation.generator.js
jannyHou: > since we allow customized keyFrom, user could have two same relations with 2 models with two different names. In this case, they will share the same getter.

👍 Thank you I agree with you.
The property should be checked instead of parameter.

packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
agnes512: I change the generator to generate `customerClassId` instead to match with our code in `/repository/relation`

packages/cli/lib/utils.js
agnes512: Right, it's so much better, thanks!

packages/cli/test/fixtures/relation/index.js
nabdelgadir: Should we update the one in `nokey.model.ts` then?

docs/site/BelongsTo-relation.md
```diff
@@ -43,6 +43,11 @@ related routes, you need to perform the following steps:
 This section describes how to define a `belongsTo` relation at the model level
 using the `@belongsTo` decorator to define the constraining foreign key.
 
+LB4 also provides an CLI tool `lb4 relation` to generate `belongsTo` relation
```
nabdelgadir: So that the reader doesn't have to come back to this section, can you move it to after the `Relation Metadata` section?

docs/site/Relation-generator.md
```diff
@@ -32,7 +32,8 @@ lb4 relation [options]
 - `--relationType`: Relation type.
 - `--sourceModel`: Source model.
 - `--destinationModel`: Destination model.
-- `--foreignKeyName`: Destination model foreign key name.
+- `--foreignKeyName`: Destination (Source) model foreign key name for HasMany
+  (BelongsTo) relation.
```
nabdelgadir: suggestion: `Destination/Source model foreign key name for HasMany/BelongsTo relation, respectively.` (feel free to ignore)

packages/cli/test/fixtures/relation/index.js
agnes512: either is fine..

docs/site/BelongsTo-relation.md
agnes512: I was thinking where is the best place to put it. I don't think after relation metadata is a good idea because we introduce how users can customize keyFrom, keyTo, relation name, and even column name, and most of them are not supported by the CLI. That's why I put it here, right after we introduce the basic idea of each relation.

Issue #4363, opened 2020-01-04
MaxListenersExceededWarning: Possible EventEmitter memory leak detected


gohmc:
When I ran load test against lb4 backed application, i got flooded with the above warnings and observed the lb4 application slow down significantly (4x slower than normal with very poor throughput). However, the application didn't crash.

It looks like too many interceptors need to watch the event loop? It can be reproduce with one of the lb4 example: 

lb4 example greeting-app
cd loopback4-example-greeting-app
npm start     // with --trace-warnings
Use a load test application against http://127.0.0.1:3000/greet/world. Example:
autocannon -c 10 -d 40 -p 10 localhost:3000/greet/world

(node:30929) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 bind listeners added to [RestServer]. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:268:17)
    at RestServer.addListener (events.js:284:10)
    at RequestContext.addParentEventListener (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/context.js:152:22)
    at RequestContext.setupEventHandlersIfNeeded (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/context.js:113:14)
    at RequestContext.subscribe (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/context.js:414:14)
    at ContextView.open (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/context-view.js:46:43)
    at Object.createViewGetter (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/context-view.js:171:10)
    at Object.resolve (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/core/dist/extension-point.js:58:26)
    at /home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/resolver.js:102:30
    at /home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/resolution-session.js:92:53
    at Object.tryWithFinally (/home/ec2-user/projects/loopback/loopback4-example-greeting-app/node_modules/@loopback/context/dist/value-promise.js:157:18)
[...]

PR #4377, opened 2020-01-07
feat: improve context.find/findByTag() and interceptor perf

### LINKED ISSUE #4356: Performance degradation
##  Steps to reproduce

Run performance tests, when the performance criteria are requests per second and latency. 

## Current Behavior

I am using  `"@loopback/context": "^1.25.0"`, and I noticed a degradation in performance.
After analyzing the results using a profiler, I found out that interceptors loading is one of the main causes of this degradation.

Specifically, the following line (in `getGlobalInterceptorBindingKeys` function of `InterceptedInvocationContext` class)  is causing a performance issue :

```ts
  const bindings: Readonly<Binding<Interceptor>>[] = this.find(
      binding =>
        filterByTag(ContextTags.GLOBAL_INTERCEPTOR)(binding) &&
        // Only include interceptors that match the source type of the invocation
        this.applicableTo(binding),
    );
```

After forcing the `controllerRoute` to use `invokeTargetMethodWithInjection` instead of `invokeMethodWithInterceptors`, the performance (Requests per seconds) improved by 25%.
Note also that I don't use interceptors in my application.

## Expected Behavior

No degradation in performance after upgrading loopback

raymondfeng:
Fixes #4356 and #4363 

Matching all bindings by a filter function can be expensive. This PR improves performance for one of the primary usage - find bindings by tags.

Main changes in this PR:

1. Make `Binding` to be `EventEmitter` - emitting events when binding scope/tags/value are changed
2. Set up listeners in `Context` to react to binding events to maintain an index of bindings by tag
3. Optimize `Context.findByTag` to leverage binding index if possible
4. Change interceptor to find matching global interceptors by tag

## Checklist

### DIFF
--- a/packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
@@ -186,8 +186,8 @@ describe('Interceptor', () => {
       }
     }
 
-    // No listeners yet
-    expect(ctx.listenerCount('bind')).to.eql(0);
+    // No invocation context related listeners yet
+    const listenerCount = ctx.listenerCount('bind');
     const controller = new MyController();
 
     // Run the invocation 5 times
@@ -199,7 +199,7 @@ describe('Interceptor', () => {
         'greet',
         ['John'],
       );
-      // New listeners are added to `ctx`
+      // New listeners are added to `ctx` by the invocation context
       expect(ctx.listenerCount('bind')).to.be.greaterThan(count);
 
       // Wait until the invocation finishes
@@ -208,7 +208,7 @@ describe('Interceptor', () => {
 
     // Listeners added by invocation context are gone now
     // There is one left for ctx.observers
-    expect(ctx.listenerCount('bind')).to.eql(1);
+    expect(ctx.listenerCount('bind')).to.eql(listenerCount + 1);
   });
 
   it('invokes static interceptors', async () => {
--- a/packages/context/src/binding-filter.ts
+++ b/packages/context/src/binding-filter.ts
@@ -5,6 +5,7 @@
 
 import {Binding, BindingTag} from './binding';
 import {BindingAddress} from './binding-key';
+import {MapObject} from './value-promise';
 
 /**
  * A function that filters bindings. It returns `true` to select a given
@@ -60,26 +61,77 @@ export function isBindingAddress(
 }
 
 /**
+ * Binding filter function that holds a binding tag pattern. `Context.find()`
+ * uses the `bindingTagPattern` to optimize the matching of bindings by tag to
+ * avoid expensive check for all bindings.
+ */
+export interface BindingTagFilter extends BindingFilter<unknown> {
+  /**
+   * A special property on the filter function to provide access to the binding
+   * tag pattern which can be utilized to optimize the matching of bindings by
+   * tag in a context.
+   */
+  bindingTagPattern: BindingTag | RegExp;
+}
+
+/**
+ * Type guard for BindingTagFilter
+ * @param filter - A BindingFilter function
+ */
+export function isBindingTagFilter(
+  filter?: BindingFilter,
+): filter is BindingTagFilter {
+  if (filter == null || !('bindingTagPattern' in filter)) return false;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const tagPattern = (filter as any).bindingTagPattern;
+  return (
+    tagPattern instanceof RegExp ||
+    typeof tagPattern === 'string' ||
+    typeof tagPattern === 'object'
+  );
+}
+
+/**
  * Create a binding filter for the tag pattern
  * @param tagPattern - Binding tag name, regexp, or object
  */
-export function filterByTag(tagPattern: BindingTag | RegExp): BindingFilter {
-  if (typeof tagPattern === 'string' || tagPattern instanceof RegExp) {
-    const regexp =
-      typeof tagPattern === 'string'
-        ? wildcardToRegExp(tagPattern)
-        : tagPattern;
-    return b => Array.from(b.tagNames).some(t => regexp!.test(t));
+export function filterByTag(tagPattern: BindingTag | RegExp): BindingTagFilter {
+  let filter: BindingFilter;
+  let regex: RegExp | undefined = undefined;
+  if (tagPattern instanceof RegExp) {
+    // RegExp for tag names
+    regex = tagPattern;
+  }
+  if (
+    typeof tagPattern === 'string' &&
+    (tagPattern.includes('*') || tagPattern.includes('?'))
+  ) {
+    // Wildcard tag name
+    regex = wildcardToRegExp(tagPattern);
+  }
+
+  if (regex != null) {
+    // RegExp or wildcard match
+    filter = b => b.tagNames.some(t => regex!.test(t));
+  } else if (typeof tagPattern === 'string') {
+    // Plain tag string match
+    filter = b => b.tagNames.includes(tagPattern);
   } else {
-    return b => {
+    // Match tag name/value pairs
+    const tagMap = tagPattern as MapObject<unknown>;
+    filter = b => {
       for (const t in tagPattern) {
         // One tag name/value does not match
-        if (b.tagMap[t] !== tagPattern[t]) return false;
+        if (b.tagMap[t] !== tagMap[t]) return false;
       }
       // All tag name/value pairs match
       return true;
     };
   }
+  // Set up binding tag for the filter
+  const tagFilter = filter as BindingTagFilter;
+  tagFilter.bindingTagPattern = regex ?? tagPattern;
+  return tagFilter;
 }
 
 /**
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -11,9 +11,15 @@ import {
   ConfigurationResolver,
   DefaultConfigurationResolver,
 } from './binding-config';
-import {BindingFilter, filterByKey, filterByTag} from './binding-filter';
+import {
+  BindingFilter,
+  filterByKey,
+  filterByTag,
+  isBindingTagFilter,
+} from './binding-filter';
 import {BindingAddress, BindingKey} from './binding-key';
 import {BindingComparator} from './binding-sorter';
+import {ContextEvent, ContextEventListener} from './context-event';
 import {
   ContextEventObserver,
   ContextEventType,
@@ -21,6 +27,7 @@ import {
   Notification,
   Subscription,
 } from './context-observer';
+import {ContextTagIndexer} from './context-tag-indexer';
 import {ContextView} from './context-view';
 import {ContextBindings} from './keys';
 import {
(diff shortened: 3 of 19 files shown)


packages/context/src/binding.ts
```diff
@@ -148,7 +157,7 @@ type ValueGetter<T> = (
  * Binding represents an entry in the `Context`. Each binding has a key and a
  * corresponding value getter.
  */
-export class Binding<T = BoundValue> {
+export class Binding<T = BoundValue> extends EventEmitter {
```
bajtos: I am concerned about the performance implications of turning every binding into an event emitter. I am expecting that consumers of these events will want to listen on Context instances, not on individual bindings.

Have you considered changing this implementation to emit the events on the Context object the binding is added to? If the binding is not added to any Context object yet, then I am arguing that the events can be silently discarded.

I am also surprised that we did not these Binding-level events when implementing [Context Observer](https://loopback.io/doc/en/lb4/Context.html#context-observers). Does it mean that Context observers were not able to be notified about these kinds of changes before? Did we perhaps use a different mean to achieve this functionality for context observers?

packages/context/src/context.ts
```diff
@@ -53,6 +53,12 @@ import {iterator, multiple} from 'p-event';
 
 const debug = debugFactory('loopback:context');
 
+export type ContextEventListener = (
+  binding: Readonly<Binding<unknown>>,
+  context: Context,
+  event: string,
+) => void;
```
bajtos: Can you please explain what's the difference between `ContextObserver` and `ContextEventListener`? When should our users use which?

Please add a tsdoc comment for this interface. I think we should update the documentation to mention this new way of observing context changes.

packages/context/src/context.ts
```diff
@@ -357,9 +375,12 @@ export class Context extends EventEmitter {
     this.registry.set(key, binding);
     if (existingBinding !== binding) {
       if (existingBinding != null) {
-        this.emit('unbind', existingBinding, this);
+        this.emit('unbind', existingBinding, this, 'unbind');
+        this.removeTagIndexForBinding(existingBinding);
```
bajtos: I am quite unhappy that you are modifying so many places to deal with tag indexing. I feel you are mixing implementation of different concerns in a single method/function, which make the codebase difficult to understand and modify.

Please find a way how to implement tag indexing as a de-coupled piece using events and/or ContextObserver to be notified about the relevant changes.

I am ok to include tag index inside the Context class, I just want the low-level implementation bits to live outside of Context.

packages/context/src/context.ts
```diff
@@ -686,7 +734,36 @@ export class Context extends EventEmitter {
...
   ): Readonly<Binding<ValueType>>[] {
-    return this.find(filterByTag(tagFilter));
+    if (
+      tagFilter instanceof RegExp ||
+      (typeof tagFilter === 'string' &&
+        (tagFilter.includes('*') || tagFilter.includes('?')))
+    ) {
+      return this.find(filterByTag(tagFilter));
+    }
+    return this._findByTagIndex(tagFilter);
```
bajtos: While this solution works for the particular case of improving performance of controller methods when there are no interceptors configured, I am concerned it's too specialized and won't work for other scenarios. For example, if the application has 100 different interceptors bound in the context, to execute a controller method with no interceptors, we will still have to scan that array of 100 interceptors.

Also IIRC our previous conversations, we wanted to move away from `findByTag` in favor of more generic filtering. In this pull request, you are reversing that direction.

Have you considered implementing a more generic solution, one that will allow consumers outside of `@loopback/context` to implement their own cache and have an easy solution for invalidating it?

I think ideally, we want to:
- for each controller method, cache the actual list of interceptors to invoke 
- invalidate cache entries when the controller method metadata changes or when an interceptors is bound or unbound

Maybe we can treat this idea as a long-term goal and implement tag indexing as a short-term performance improvement.

Thoughts?

packages/context/src/binding.ts
raymondfeng: At this moment, the same binding instance can be added to multiple context objects. If we decide to treat each binding to be optionally owned by one context, we can emit such events on the owning context.

When we implemented ContextObserver - which is async, we choose to process `bind` and `unbind` events at  a late cycle to work around the issue that bindings can be changed. Now with binding events, we can probably revisit the implementation to track `binding.changed` events.

packages/context/src/context.ts
raymondfeng: `ContextEventListener` is a sync function as typical Node.js event listeners. The `ContextObserver` is async handler.

packages/context/src/context.ts
raymondfeng: I have added a few more commits to fully leverage `BindingTagFilter` is possible - even for `ContextView`.

packages/context/src/binding.ts
raymondfeng: IMO, setting `Context` ref on `Binding` is similar as `Context` registering itself as a listener on `Binding` from performance overhead perspective.

raymondfeng:
@bajtos PTAL.

packages/context/src/binding.ts
bajtos: > At this moment, the same binding instance can be added to multiple context objects.

Somehow I thought that a Binding is already holding a reference to the single owning Context. I check the source code and see that it does not. As you wrote, it's possible to add the same binding to multiple contexts, therefore I agree with your originally proposed design, where each Binding is an EventEmitter 👍 

> When we implemented ContextObserver - which is async, we choose to process `bind` and `unbind` events at a late cycle to work around the issue that bindings can be changed. Now with binding events, we can probably revisit the implementation to track `binding.changed` events.

Make sense. I'd like to revisit this part sooner rather than later, to ensure consistency.

packages/context/src/context.ts
bajtos: > `ContextEventListener` is a sync function as typical Node.js event listeners. The `ContextObserver` is async handler.

Sure, I can tell this part from the source code. I am asking for user-centric documentation - when should our users choose ContextObserver and when to choose ContextEventListener? As framework authors, we should set clear guidance here.

packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
```diff
@@ -187,7 +187,7 @@ describe('Interceptor', () => {
     }
 
     // No listeners yet
-    expect(ctx.listenerCount('bind')).to.eql(0);
+    expect(ctx.listenerCount('bind')).to.eql(1);
```
bajtos: Why is the listener count 1 considered as "no listeners yet"? Please update the code comment to explain.

packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
```diff
@@ -208,7 +208,7 @@ describe('Interceptor', () => {
 
     // Listeners added by invocation context are gone now
     // There is one left for ctx.observers
-    expect(ctx.listenerCount('bind')).to.eql(1);
+    expect(ctx.listenerCount('bind')).to.eql(2);
```
bajtos: Ditto.

raymondfeng:
It now depends on https://github.com/strongloop/loopback-next/pull/4430

raymondfeng:
@bajtos I landed #4430 and refactored this PR into 3 commits. PTAL.

bajtos:
> It now depends on #4430

Awesome!

> refactored this PR into 3 commits

That's helpful too.

> chore(context): tidy up context listeners for bind/unbind events

As far as I can see, this is not a chore, you are introducing a new feature `ContextEventListener`. Let's move this commit into a new PR please and make sure to include documentation.

See the thread in https://github.com/strongloop/loopback-next/pull/4377#discussion_r364745654:

> > `ContextEventListener` is a sync function as typical Node.js event listeners. The `ContextObserver` is async handler.

> Sure, I can tell this part from the source code. I am asking for user-centric documentation - when should our users choose ContextObserver and when to choose ContextEventListener? As framework authors, we should set clear guidance here.

packages/context/src/binding-filter.ts
```diff
@@ -59,27 +60,63 @@ export function isBindingAddress(
...
+
+/**
+ * Type guard for BindingTagFilter
+ * @param filter - A BindingFilter function
+ */
+export function isBindingTagFilter(
+  filter?: BindingFilter,
+): filter is BindingTagFilter {
+  return filter != null && 'bindingTag' in filter;
+}
```
bajtos: Can you please add some tests for this new method?

packages/context/src/binding-filter.ts
```diff
@@ -59,27 +60,63 @@ export function isBindingAddress(
   return typeof bindingSelector !== 'function';
 }
 
+/**
+ * Binding filter function that uses tags
+ */
+export interface BindingTagFilter extends BindingFilter<unknown> {
```
bajtos: As far as this commit goes, it's not clear to me how is `BindingTagFilter` supposed to be used. Can you please enhance the docs and perhaps add some tests to verify the scenarios where this interface is used.

For example, you can call `filterByTag` with different arguments (a tag name, a regexp, and so on; one call per test please!) and verify that it returns `BindingTagFilter` instance with the correct value set inside `bindingTag`.

raymondfeng:
It now depends #4451

packages/context/src/__tests__/acceptance/interceptor.acceptance.ts
```diff
@@ -208,7 +208,7 @@ describe('Interceptor', () => {
 
     // Listeners added by invocation context are gone now
     // There is one left for ctx.observers
-    expect(ctx.listenerCount('bind')).to.eql(1);
+    expect(ctx.listenerCount('bind')).to.eql(listenerCount + 1);
```
agnes512: Does the number of `BindingEventListener` change alone with the number of the tags of a certain binding and tags only?

raymondfeng: This test is for number of listeners on a `Context`. Please note that `BindingEventListener` is registered on bindings. When a binding is added to a context, there will be one binding event listener added to the binding by the context.

packages/context/src/context.ts
```diff
@@ -637,6 +686,32 @@ export class Context extends EventEmitter {
...
+    this.removeTagIndexForBinding(binding);
+    for (const tag of binding.tagNames) {
+      let bindings = this.bindingsIndexedByTag.get(tag);
+      if (bindings == null) {
+        bindings = new Set();
+        this.bindingsIndexedByTag.set(tag, bindings);
+      }
+      bindings.add(binding);
+    }
+  }
```
bajtos: I am concerned about the ever-growing size of `packages/context/src/context.ts`. Is it possible to move these new functions to a standalone class, e.g. `ContextTagIndexer` and then use methods of that class for event listeners and when looking up bindings by tags?

packages/context/src/context.ts
```diff
@@ -719,6 +799,65 @@ export class Context extends EventEmitter {
     return this.find(filterByTag(tagFilter));
   }
 
+  /**
+   * Find bindings by tag leveraging indexes
+   * @param tag - Tag name pattern or name/value pairs
+   */
+  protected _findByTagIndex<ValueType = BoundValue>(
```
bajtos: Ideally, this method should be implemented by `ContextTagIndexer` too.

packages/context/src/context.ts
raymondfeng: Extracted into `ContextTagIndexer`.

raymondfeng:
@bajtos I also added a commit to extract context observer subscription logic into `ContextSubscriptionManager` class.

bajtos:
@raymondfeng thank you for extracting smaller pieces out of the context class/file, the code structure looks much better now!

I am afraid I don't have enough energy to review the pull request in detail this week. I quickly skimmed through the changes and don't see any obvious major issues. Let's ask other owners of this functional area to review & approve the changes.

@deepakrkris @emonddr @jannyHou PTAL

Issue #4392, opened 2020-01-08
Relations with custom names not working


LonguCodes:
## Steps to reproduce

Create a new Model with a property with belongsTo decorator and add name to definiton. Add relation to repository with the same name.

## Current Behavior

The relation is not created with the given name

## Expected Behavior

There would be a relation to the given name. If done using the default convention `relation name` + Id works as expected.

## Additional information

  linux x64 10.16.3
 ├── @loopback/authentication@3.2.4
├── @loopback/authentication-passport@1.0.6
├── @loopback/boot@1.6.0
├── @loopback/context@1.25.0
├── @loopback/core@1.12.0
├── @loopback/openapi-v3@1.10.3
├── @loopback/repository@1.16.0
├── @loopback/rest@1.25.0
├── @loopback/rest-explorer@1.4.6
├── @loopback/service-proxy@1.3.13
├── loopback-connector-mysql@5.4.2

dhmlau:
@agnes512, could you please help? Thanks.

agnes512:
@LonguCodes Hi, this is the limitation of our current design. We reject CRUD operations if the request body contains navigational properties. In your case, your source key and the relation name have the same name. It gets rejected when your request contains _the property name_. For example, when you create an instance with `{id: 1, name: 'my order', customer: 1}`, and your relation also named `customer`, it would recognize `customer` as the navigational property(i.e a `Customer` instance) instead of a property of model itself. So the current workaround is **not to use the same name for belongsTo source key and the relation name**.

Ref: [BelongsTo relation metadata](https://loopback.io/doc/en/lb4/BelongsTo-relation.html#relation-metadata), to customize your source key name in the source model for the belongsTo relation, this is what you need to do:

For example,  an`Order` belongsTo a `Customer`. The source key is `Order.customerId` and the relation name is `customer` by default.
I'd like to have the source key named `c_id` and the relation name as `myCustomer`. Since the source key defaults to the one decorated by the `@belongsTo` decorator, which is `c_id` in our case, we'd still need to specify the customized relation name:

```ts
export class Order extends Entity {
  @property({
    id: true,
    generated: true,
    type: 'number'
  })
  id: number;

  @belongsTo(() => Customer, {name: 'myCustomer'})
  c_id: number;
```

LonguCodes:
Thanks for the help. It cleared up why i had problems before.

PR #4400, opened 2020-01-09
fix(repository): make the navigational property err msg more clear

agnes512:
related https://github.com/strongloop/loopback-next/issues/4392,  [stackoverflow](https://stackoverflow.com/questions/59468329/unhandled-error-in-post-order-products-500-error-navigational-properties-are)

The `belongsTo` generator generates the default relation name by taking the `@belongsTo` decorated name then removing `Id` from it. **Idealy**, the default fk name is expected to be `customerId`, and the ideal relation name will be `customer` ([code]( https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L45)).

However, lots of users would come up with their own customized name, `customer_id` for example.  And the default relation name generated by the belongsTo will be `customer_id` according to the code. This is problematic because in CRUD operations, it recognize `customer_id` as _a navigational property_ instead of a model property. As a result, operations like
```
orderRepo.create({id: 1, name: 'LoopBack', customer_id: 1})
```
will be rejected with error `500 Error: Navigational properties are not allowed in model data`.

Currently I couldn't find a good way to improve the way it generates the default relation name. So I am proposing to make the error message more clear and also update the site before we come up a better way for generating the default relation name.

Example new error message.
`Navigational properties are not allowed in model data (model "Order" property "customer_id"). Please remove it or make sure the relation name is not the same as the property name in belongsTo relation.`

## Checklist

### DIFF
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
@@ -221,7 +221,7 @@ export function hasManyRelationAcceptance(
           ],
         }),
       ).to.be.rejectedWith(
-        'Navigational properties are not allowed in model data (model "Customer" property "orders")',
+        'Navigational properties are not allowed in model data (model "Customer" property "orders"), please remove it.',
       );
     });
 
@@ -238,7 +238,7 @@ export function hasManyRelationAcceptance(
           },
         ]),
       ).to.be.rejectedWith(
-        'Navigational properties are not allowed in model data (model "Customer" property "orders")',
+        'Navigational properties are not allowed in model data (model "Customer" property "orders"), please remove it.',
       );
     });
 
@@ -296,7 +296,7 @@ export function hasManyRelationAcceptance(
       });
 
       await expect(customerRepo.delete(found)).to.be.rejectedWith(
-        'Navigational properties are not allowed in model data (model "Customer" property "orders")',
+        'Navigational properties are not allowed in model data (model "Customer" property "orders"), please remove it.',
       );
     });
 
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -573,17 +573,23 @@ export class DefaultCrudRepository<
       typeof entity.toJSON === 'function' ? entity.toJSON() : {...entity};
     */
 
-    // proposal solution from agnes: delete the id property in the
-    // target data when runs replaceById
-
     const data: AnyObject = new this.entityClass(entity);
 
     const def = this.entityClass.definition;
+    const props = def.properties;
     for (const r in def.relations) {
       const relName = def.relations[r].name;
       if (relName in data) {
+        let invalidNameMsg = '';
+        if (relName in props) {
+          invalidNameMsg =
+            ` The error might be invoked by belongsTo relations, please make sure the relation name is not the same as` +
+            ` the property name.`;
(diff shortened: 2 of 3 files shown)


packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts
```diff
@@ -221,7 +221,7 @@ export function hasManyRelationAcceptance(
           ],
         }),
       ).to.be.rejectedWith(
-        'Navigational properties are not allowed in model data (model "Customer" property "orders")',
+        'Navigational properties are not allowed in model data (model "Customer" property "orders"). Please remove it or make sure the relation name is not the same as the property name in belongsTo relation.',
```
bajtos: This line is too long and difficult to review on GitHub. Please split it into multiple lines, e.g.

```ts
.to.be.rejectedWith(
  'Navigational properties are not allowed in model data (model "Customer" property "orders"). ' + 
    'Please remove it or make sure the relation name is not the same as the property name in belongsTo relation.',
);
```

(It's just an example, I am not sure if that's enough to keep the line length under 80 chars.)

The same comment applies to most other changed lines too.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -583,7 +583,7 @@ export class DefaultCrudRepository<
       const relName = def.relations[r].name;
       if (relName in data) {
         throw new Error(
-          `Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}")`,
+          `Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}"). Please remove it or make sure the relation name is not the same as the property name in belongsTo relation.`,
```
jannyHou: Should it be `belongsTo` or `hasMany`? 
Like `Customer` hasMany `orders`, IIUC the navigational property here is `orders`, which is the relation name for hasMany

agnes512: Yes, 
For hasMany, `orders` would only be the navigational property. When you create a customer, you would do `customerRepo.create({id: 1, name: 'customer'});` 

But for belongsTo, by how we defining the default relation name, the fk `customer_id` would have the default relation name as `customer_id`. When you do `orderRepo.create({id: 1, des: 'order', customer_id: 1})`, it recognizes the `customer_id` in the data as the **_relation name_** instead of the foreign key. That's why I state it clear that that the error might be caused by the nav property OR the naming issue here. 

Please let me know if you have any suggestions to make it more intuitive.

jannyHou: @agnes512 👍 thank you got it. How about the following warning:

Navigational properties are not allowed in model data (model "${this.entityClass.modelName}" property "${relName}"), please remove it. 
For relation belongsTo, please make sure the relation name is not the same as the property name.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -579,11 +579,20 @@ export class DefaultCrudRepository<
     const data: AnyObject = new this.entityClass(entity);
 
     const def = this.entityClass.definition;
+    const props = def.properties;
     for (const r in def.relations) {
       const relName = def.relations[r].name;
       if (relName in data) {
+        let invalidName = '';
```
jannyHou: nitpick: `invalidName` --> `invalidNameMsg` :-p

PR #4401, opened 2020-01-09
docs: update belongsTo relation metadata

agnes512:
Related https://github.com/strongloop/loopback-next/issues/4392,  [stackoverflow](https://stackoverflow.com/questions/59468329/unhandled-error-in-post-order-products-500-error-navigational-properties-are)

The `belongsTo` generator generates the default relation name by taking the `@belongsTo` decorated name then removing `Id` from it. **Idealy**, the default fk name is expected to be `customerId`, and the ideal relation name will be `customer` ([code]( https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L45)).

However, lots of users would come up with their own customized name, `customer_id` for example.  And the default relation name generated by the belongsTo will be `customer_id` according to the code. This is problematic because in CRUD operations, it recognize `customer_id` as _a navigational property_ instead of a model property. As a result, operations like
```
orderRepo.create({id: 1, name: 'LoopBack', customer_id: 1})
```
will be rejected with error `500 Error: Navigational properties are not allowed in model data`.

Currently I couldn't find a good way to improve the way it generates the default relation name. Updating the site before we come up a better way for generating the default relation name.

## Checklist

### DIFF
--- a/docs/site/BelongsTo-relation.md
+++ b/docs/site/BelongsTo-relation.md
@@ -45,30 +45,48 @@ using the `@belongsTo` decorator to define the constraining foreign key.
 
 ### Relation Metadata
 
-The definition of the `belongsTo` relation is inferred by using the `@belongsTo`
-decorator. The decorator takes in a function resolving the target model class
-constructor and designates the relation type. It also calls `property()` to
-ensure that the decorated property is correctly defined.
+LB4 uses three `keyFrom`, `keyTo` and `name` fields in the `belongsTo` relation
+metadata to configure relations. The relation metadata has its own default
+values for these three fields:
 
-The `@belongsTo` decorator takes three parameters:
+<table>
+  <thead>
+    <tr>
+      <th width="95">Field Name</th>
+      <th width="260">Description</th>
+      <th width="260">Default Value</th>
+      <th>Example</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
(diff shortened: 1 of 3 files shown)


docs/site/BelongsTo-relation.md
```diff
@@ -45,30 +45,48 @@ using the `@belongsTo` decorator to define the constraining foreign key.
...
+    <tr>
+      <td><code>keyTo</code></td>
+      <td>the source key of the target model</td>
+      <td>the primary key in the target model</td>
+      <td><code>Customer.id</code></td>
+    </tr>
+    <tr>
+      <td><code>name</code></td>
+      <td>the name of the relation</td>
+      <td>@belongTo decorated property name without <code>Id</code></td>
```
nabdelgadir: `belongTo` -> `belongsTo`

docs/site/BelongsTo-relation.md
```diff
@@ -102,34 +120,65 @@ export interface OrderRelations {
...
+// import statements
+@model()
+export class Order extends Entity {
+  @property({
+    type: 'number',
+    id: true,
+  })
+  id: number;
+
+  @belongsTo(() => Customer, {name: 'customer'})
```
nabdelgadir: Maybe put a comment here saying this is the correct way since from above you're warning what not to do

PR #4404, opened 2020-01-10
feat(eslint-config): enable no-throw-literal rule

bajtos:
~~IMPORTANT: landing of this pull request is blocked by https://github.com/typescript-eslint/typescript-eslint/issues/1427~~

Quoting from eslint docs for [`no-throw-literal`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-throw-literal.md):

> It is considered good practice to only throw the `Error` object itself or an object using the `Error` object as base objects for user-defined exceptions. The fundamental benefit of `Error` objects is that they automatically keep track of where they were built and originated.
>
> This rule restricts what can be thrown as an exception. When it was first created, it only prevented literals from being thrown (hence the name), but it has now been expanded to only allow expressions which have a possibility of being an `Error` object.

## Checklist

### DIFF
--- a/packages/eslint-config/eslintrc.js
+++ b/packages/eslint-config/eslintrc.js
@@ -154,6 +154,7 @@ module.exports = {
     '@typescript-eslint/prefer-nullish-coalescing': 'error',
     '@typescript-eslint/no-extra-non-null-assertion': 'error',
     '@typescript-eslint/return-await': 'error',
+    '@typescript-eslint/no-throw-literal': 'error',
   },
 
   overrides: [


raymondfeng:
Rebased to master.

bajtos:
Status update: there are still few false positives reported by the new rule.

Few examples:

https://github.com/strongloop/loopback-next/blob/d8dfc9ae62729db1341bea3efdafb0c958a040c3/packages/rest/src/body-parsers/body-parser.ts#L76

https://github.com/strongloop/loopback-next/blob/d8dfc9ae62729db1341bea3efdafb0c958a040c3/packages/rest/src/rest.server.ts#L577-L579

We need to report these false-positives to typescript-eslint and either help the get the problem fixed or wait until somebody contributes such fix(es).

Unfortunately, I don't have bandwidth to submit those issues myself right now.

bajtos:
I don't have bandwidth to further pursue this work.

Issue #4138, opened 2019-11-17
I want to disable API explorer how to do it ? 


niti2539:
How to disable API explorer page ? 
I do as the document say but it not work 
here is my config 

const config = {
    rest: {
      port: +(process.env.PORT || 3000),
      host: process.env.HOST,
      openApiSpec: {
        setServersFromRequest: true,
        disabled: true
      },
      apiExplorer: {
        disabled: true
      }
    }
  };

openApiSpec can't access but  API explorer can access.

achrinza:
`rest.apiExplorer.disabled` [only disables redirect to an externally-hosted Swagger UI instance](https://loopback.io/doc/en/lb4/Server.html#disable-redirect-to-api-explorer).

To disable the self-hosted instance, reverse the [steps to add self-hosted `apiExplorer`](https://loopback.io/doc/en/lb4/Self-hosted-rest-api-explorer.html)

shadyanwar:
In application.ts, add `useSelfHostedSpec: false` to `RestExplorerBindings.CONFIG`. For example:
```ts
  this.bind(RestExplorerBindings.CONFIG).to({
    path: '/explorer',
    useSelfHostedSpec: false
  });
```

achrinza:
Thanks @shadyanwar for the alternate solution. To avoid confusion, as [per-docs](https://loopback.io/doc/en/lb4/Self-hosted-rest-api-explorer.html#advanced-configuration-and-reverse-proxies), `useSelfHostedSpec` does remove the explorer's self-hosting of `openapi.json` for it's own use. This however does not fully remove the explorer.

shadyanwar:
@achrinza I assumed the aim of the question was to disable access to the explorer. Is it still accessible in any way after setting `useSelfHostedSpec` to `false`?

achrinza:
@shadyanwar I've just tested this to be sure, the explorer is still accessible at `/explorer`. `useSelfHostedSpec: false` removed `/explorer/openapi.json` but not the explorer itself.

dougal83:
For completeness, to entirely disable explorer make **both** changes below:
`./index.js` - Add the following setting to options. [disable remote fallback]
````js
  const config = {
    rest: {
      // ...
      apiExplorer: {
        disabled: true,
      },
    },
  };
````
`./src/application.ts` - Remove/comment the following code [remove self hosted component]
````ts
    // Customize @loopback/rest-explorer configuration here
[...]

PR #4408, opened 2020-01-11
docs: add instruction to remove api explorer

dougal83:
Add instructions to completely disable API explorer for both remote and self-hosted.

See also #4138

## Checklist

### DIFF
--- a/docs/site/Server.md
+++ b/docs/site/Server.md
@@ -141,6 +141,9 @@ const app = new RestApplication({
 });
 ```
 
+{% include note.html content="To completely disable API Explorer, we also need
+to [disable the self-hosted REST API Explorer extension](./Self-hosted-REST-API-Explorer.md#disable-self-hosted-api-explorer)." %}
+
 ### Use a self-hosted API Explorer
 
 Hosting the API Explorer at an external URL has a few downsides, for example a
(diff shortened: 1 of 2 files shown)


docs/site/Server.md
```diff
@@ -141,6 +141,9 @@ const app = new RestApplication({
 });
 ```
 
+{% include note.html content="To completely disable API Explorer then we also
```
dhmlau: Do you mean the following? 
```
To completely disable API Explorer, we also need to disable the self-hosted REST API Explorer extension
```

packages/rest-explorer/README.md
```diff
@@ -90,6 +90,22 @@ Note also that you cannot use a url-relative path for the `servers` entry, as
...
+located in `./src/application.ts` and consist of two items, for example:
+
+```ts
+this.bind(RestExplorerBindings.CONFIG).to({
+  path: '/openapi/ui',
+});
+this.component(RestExplorerComponent);
+```
+
+{% include note.html content="To completely disable API Explorer then we also
```
dhmlau: Similar to the above comment, do you mean:
```
To completely disable API Explore, we also need to disable the redirect to the externally hosted API Explorer
```

docs/site/Server.md
dougal83: Much better.  Thanks!

packages/rest-explorer/README.md
dougal83: Much better.  Thanks!

PR #4406, opened 2020-01-10
feat: adds several convenience decorators

mschnee:
## Warning
This PR is incomplete- I am issuing it to request a preliminary round of feedback.
There are a bunch of concepts involved here that I don't entirely grok- such as the specific implementations of the `DecoratorFactories`, and a couple logical decisions that may be incorrect due to my amateur understanding of the OpenAPI specification.

### This branch adds the following openapi convenience decorators
- `@deprecated()` to a class or method
- `@tags(tagNames: string [])` to a class or method
- `@response(code: number, spec: ModelOrSpec, options?: {contentType: string, description: string})` to methods only

### This branch adds the following features:
- creates a `MethodMultiDecoratorFactory`, allowing multiple uses of the
same decorator per method, see the `@response()` decorator.
- adds support for the schema keywords 'allOf', 'anyOf', 'oneOf', and
'not' in resolving schemas with the 'x-ts-type' extension
- adds a `get-dist-file` script to `packages/build`  as a utility to make it easier to
debug a unit test file directly in vscode, see `Debug Current Test File" in `.vscode/launch.json`

## Examples
The following TypeScript code:
```ts
@model()
class SuccessModel extends Model {
  constructor(err: Partial<SuccessModel>) {
    super(err);
  }
  @property({default: 'ok'})
  message: string;
}

@model()
class FooError extends Model {
  constructor(err: Partial<FooError>) {
    super(err);
  }
  @property({default: 'foo'})
  foo: string;
}

@model()
class BarError extends Model {
  constructor(err: Partial<BarError>) {
    super(err);
  }
  @property({default: 'bar'})
  bar: string;
}

class MyController {
  @get('/greet', {
    responses: {
      200: {
        description: 'Unknown',
        content: {
          'application/json': {schema: FIRST_SCHEMA},
        },
      },
    },
  })
  @response(200, {
        type: 'string',
        format: 'base64',
      }, {contentType: 'application/pdf'})
  @response(200, SuccessModel, {contentType: 'application/jsonc'})
  @response(404, [FooError, BarError], {contentType: 'application/jsonc'})
  @tags(['Foo', 'Bar'])
  @deprecated()
  greet() {
    return new SuccessModel({message: 'Hello, world!'});
  }
}
```
The following typec
Will result in the following json:
```json
{
  "paths": {
    "/greet": {
      "get": {
        "responses": {
          "200": {
            "description": "Standard response for successful HTTP requests.",
            "content": {
              "application/json": {
                "schema": {
[...]

Issue #4300, opened 2019-12-12
Add more convenience openapi method decorators 


mschnee:
Convenience Decorators have been added to `@loopback/openapi-v3` under the `oas` namespace.

**Tags**
[[PR](https://github.com/strongloop/loopback-next/pull/4550)]
`@oas.tags(...tagNames: string[])`
Adds arbitrary tag strings to the path spec for a method or a class. 

**Deprecated**
[[(PR](https://github.com/strongloop/loopback-next/pull/4415)]
`@oas.deprecated(isDeprecated: boolean = true)`
When applied to a class, marks all paths in that controller as deprecated.  This can be overridden at the method level with `@oas.deprecated(false)`.

When applied to a method, marks that path as deprecated.

**Response**
[[PR](https://github.com/strongloop/loopback-next/pull/4550)]
`@oas.response(responseCode: number, ...modelOrSpecOrRefOrContent:  msrc[])`
Convenience decorator that lets you use `Model` objects to easily define responses.

## Example
```ts
// my.controller.ts
import {oas} from '@loopback/openapi-v3';
import {model, property, Model} from '@loopback/repository';

@model()
class SuccessModel extends Model {
  @property()
  message: string;
}

@model()
class MessageNotFoundModel extends Model {}

@model()
class SomethingNotFoundModel extends Model {}

@oas.deprecated()
export class MyController {
    @oas.get('/greet')
    @oas.tags('v1', 'MyController')
    @oas.response(200, SuccessModel)
    @oas.response(404, MessageNotFoundModel, SomethingNotFoundModel)
    @oas.deprecated(false)
    public async getGreeting() {
        return new SuccessModel({message: 'Hello, world'});
    }

    // This path will appear deprecated in the generated OpenApi spec
    @oas.get('/old-greet')
    public async notDeprecatedGreeting() {
      return 'Hello!'
    }
}
```
## Original Suggestion
Loopback is already generating a fully compliant specification- but right now it's an all or nothing approach with the method convenience decorators.  Currently, if you want to add any additional openapi specification data to an api, you are required to provide a nearly complete Operation Object to do so:
```ts
@get('/test', {
  tags: ['MyTag']
  responses: {
    '200': {
      description: 'This is required',
      content: {
        'application/json' : {}
      }
    }
  }
)
public async iJustWantToBeTagged() {
}
```

Instead, we could have done this:
```ts
@get('/test')
@tags(['MyTag'])
public async iJustWantToBeTagged() {
}
```

[...]

PR #4415, opened 2020-01-13
feat: adds @deprecated convenience decorator

mschnee:
Partially implements: #4300 
See: #4406

## Added
This PR adds `@deprecated` to Controller classes and methods, and marks all paths as `deprecated: true` in accordance with the [OpenAPI Operation Object specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object).

As the spec defines `deprecated` as an _optional property_, this feature will only add `deprecated: true` and never `deprecated: false`.

## Not (yet) supported
This PR does _not_ add support for marking parameters as `deprecated`

## Examples
```ts
@deprecated()
class MyController {
   @get('/greet')
   public async function greet() {
    return 'Hello, World!'
  }

  @get('/greet-v2')
  @deprecated(false)
  public async function greetV2() {
    return 'Hello, World!'
  }
}

class MyOtherController {
  @get('/echo')
  @deprecated()
  public async function echo() {
    return 'Echo!'
  }
}
```
## Checklist

### DIFF
--- a/docs/site/decorators/Decorators_openapi.md
+++ b/docs/site/decorators/Decorators_openapi.md
@@ -458,6 +458,44 @@ This namespace contains decorators that are specific to the OpenAPI
 specification, but are also similar to other well-known decorators available,
 such as `@deprecated()`
 
+### @oas.deprecated
+
+[API document](https://loopback.io/doc/en/lb4/apidocs.openapi-v3.deprecated.html),
+[OpenAPI Operation Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object)
+
+This decorator can currently be applied to class and a class method. It will set
+the `deprecated` boolean property of the Operation Object. When applied to a
+class, it will mark all operation methods of that class as deprecated, unless a
+method overloads with `@oas.deprecated(false)`.
+
+This decorator does not currently support marking
+(parameters)[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object]
+as deprecated.
+
+```ts
+@oas.deprecated()
+class MyController {
+   @oas.get('/greet')
+   public async function greet() {
+    return 'Hello, World!'
+  }
+
+  @oas.get('/greet-v2')
+  @oas.deprecated(false)
+  public async function greetV2() {
+    return 'Hello, World!'
+  }
+}
+
+class MyOtherController {
+  @oas.get('/echo')
+  @oas.deprecated()
+  public async function echo() {
+    return 'Echo!'
+  }
+}
+```
+
 ### @oas.tags
 
 [API document](https://loopback.io/doc/en/lb4/apidocs.openapi-v3.tags.html),
--- a/packages/openapi-v3/src/__tests__/unit/decorators/deprecated.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/deprecated.decorator.unit.ts
@@ -1,0 +1,112 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {expect} from '@loopback/testlab';
+import {api, get, getControllerSpec, oas} from '../../..';
+
+describe('deprecation decorator', () => {
+  it('Returns a spec with all the items decorated from the class level', () => {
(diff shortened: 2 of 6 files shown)


mschnee:
Note: The Node 8 build failed due to a 404 from `http://calculator-webservice.mybluemix.net/calculator?wsdl`, which at the moment (2 hours later) is not returning a 404.  Will trigger a new build.

raymondfeng:
My only concern is that `@deprecated` may confuse our developers as it's not obvious for OpenAPI spec only.

examples/todo/src/controllers/todo.controller.ts
```diff
@@ -70,6 +71,22 @@ export class TodoController {
     return this.todoRepo.findById(id);
   }
 
+  @deprecated()
```
emonddr: @mschnee , same question as in #4416. Not sure why the todo controller in our examples needs to change.
I think the code snippet examples you provide in your docs are sufficient.

mschnee: I will make this modification, however I am waiting for the full scope of changes to #4416 to be accepted first, so that I can apply it all here as well.

packages/openapi-v3/src/controller-spec.ts
```diff
@@ -77,6 +77,25 @@ function resolveControllerSpec(constructor: Function): ControllerSpec {
...
+  );
+
+  if (isClassDeprecated) {
+    debug('  using class-level @deprecated()');
+  }
+
+  if (isClassDeprecated) {
+    for (const path of Object.keys(spec.paths)) {
+      for (const method of Object.keys(spec.paths[path])) {
+        if (isClassDeprecated) {
```
jannyHou: I think the 2nd `isClassDeprecated` check is a duplicate.

dougal83:
> As the spec defines `deprecated` as an _optional property_, this feature will only add `deprecated: true` and never `deprecated: false`.

Thinking about it, it wouldn't hurt to explicitly default to setting `depreciated: false` as the last property of a `PathsObject`.  Considering the default is `false` it would be informational.

docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,49 @@ export class SomeController {
...
+
+## Convenience Decorators
+
+While you can supply a fully valid OpenAPI specification for the class-level
+`@api` decorator, and full operation OpenAPI specification for `@op` and the
+other convenience decorators, there are also a number of utility decorators that
+allow you to supply specific OpenAPI information without requiring you to use
+verbose JSON.
+
+### @deprecated
```
raymondfeng: Let's use `@oas.deprecated`

mschnee: Done!

packages/openapi-v3/src/__tests__/unit/decorators/deprecated.decorator.unit.ts
```diff
@@ -0,0 +1,113 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {expect} from '@loopback/testlab';
+import {getControllerSpec} from '../../..';
+import {api, get, oas} from '../../../decorators';
```
raymondfeng: Can we merge the import statement with the one above?

packages/openapi-v3/src/decorators/deprecated.decorator.ts
```diff
@@ -0,0 +1,88 @@
...
+ * class MyOtherController {
+ *   @get('/echo')
+ *   @deprecated()
+ *   public async function echo() {
+ *     return 'Echo!'
+ *   }
+ * }
+ * ```
+ */
+export function deprecated(isDeprecated = true) {
```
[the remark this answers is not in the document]
mschnee: Agreed!  The namespace export is being done in `index.ts`

packages/openapi-v3/src/controller-spec.ts
mschnee: Changed!  The top if block is for `if (isClassDeprecated || classHasTags) {}`.  Preferring to iterate over existing paths once, and not have two `for (path of spec.paths) {}` loops.

packages/openapi-v3/src/__tests__/unit/decorators/deprecated.decorator.unit.ts
mschnee: Done!

mschnee:
Squash will happen after all the rests are clean :)

PR #4416, opened 2020-01-13
feat: adds @tags convenience decorator

mschnee:
Partially implements: #4300 
See: #4406

## Added
This PR adds `@oas.tags` to Controller classes and methods, and adds an array of tag strings in accordance with the [OpenAPI Operation Object specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object).

As the spec defines `tags` as an _optional property_, this feature will only add `tags: string[]` when tags are set, and never an empty array of tags.

Using the `@oas.tags` decorator will set the tags on a path, and will be appended to any `@tags` set on a class.

This decorator is added to the `oas` namespace export in order to prevent naming conflicts with other popular decorators.  The existing decorators in `@loopback/openapi-v3` have also been added to the `oas` namespace (but not removed from the main export).

## Not supported
This PR has no bearing on the top-level `tags` entity in the [OpeAPI Tag Object specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#tag-object).  Operation Tags in the spec are themselves convenience for easily grouping and sorting paths and operations, and don't require an entry in the top level.  

## Examples
```ts
@oas.tags('Foo', 'Bar')
class MyController {
  @oas.tags('Baz')
  @oas.get('/greet')
  public async greet() {
    // tags will be [Foo, Bar, Baz]
  }
}
```
## Checklist

### DIFF
--- a/docs/site/decorators/Decorators_openapi.md
+++ b/docs/site/decorators/Decorators_openapi.md
@@ -430,3 +430,61 @@ export class SomeController {
 When the OpenAPI spec is generated, the `xs-ts-type` is mapped to
 `{$ref: '#/components/schemas/MyModel'}` and a corresponding schema is added to
 `components.schemas.MyModel` of the spec.
+
+## Convenience Decorators
+
+While you can supply a fully valid OpenAPI specification for the class-level
+`@api` decorator, and full operation OpenAPI specification for `@operation` and
+the other convenience decorators, there are also a number of utility decorators
+that allow you to supply specific OpenAPI information without requiring you to
+use verbose JSON.
+
+## Shortcuts for the OpenAPI Spec (OAS) Objects
+
+All of the above are direct exports of `@loopback/openapi-v3`, but they are also
+available under the `oas` namespace:
+
+```ts
+import {oas} from '@loopback/openapi-v3';
+
+@oas.api({})
+class MyController {
+  @oas.get('/greet/{id}')
+  public greet(@oas.param('id') id: string) {}
+}
+```
+
+This namespace contains decorators that are specific to the OpenAPI
+specification, but are also similar to other well-known decorators available,
+such as `@deprecated()`
+
+### @oas.tags
+
+[API document](https://loopback.io/doc/en/lb4/apidocs.openapi-v3.tags.html),
+[OpenAPI Operation Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object)
+
+This decorator can be applied to a controller class and to controller class
+methods. It will set the `tags` array string property of the Operation Object.
+When applied to a class, it will mark all operation methods of that class with
+those tags. Usage on both the class and method will combine the tags.
+
+```ts
+@oas.tags('Foo', 'Bar')
+class MyController {
+  @oas.get('/greet')
+  public async greet() {
+    // tags will be [Foo, Bar]
+  }
+
+  @oas.tags('Baz')
+  @oas.get('/echo')
+  public async echo() {
+    // tags will be [Foo, Bar, Baz]
+  }
+}
+```
+
+This decorator does not affect the top-level `tags` section defined in the
+[OpenAPI Tag Object specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#tag-object).
+This decorator only affects the spec partial generated at the class level. You
+may find that your final tags also include a tag for the controller name.
--- a/packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
+++ b/packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
@@ -1,0 +1,140 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {expect} from '@loopback/testlab';
+import {api, get, getControllerSpec, oas} from '../../..';
+
+describe('@oas.tags decorator', () => {
+  context('Without a top-level @api definition', () => {
+    it('Allows a class to not be decorated with @oas.tags at all', () => {
+      class MyController {
+        @get('/greet')
+        greet() {
+          return 'Hello world!';
+        }
+
+        @get('/echo')
+        echo() {
+          return 'Hello world!';
+        }
+      }
+
+      const actualSpec = getControllerSpec(MyController);
+      expect(actualSpec.paths['/greet'].get.tags).to.be.undefined();
+    });
+
+    it('Allows a class to decorate methods with @oas.tags', () => {
+      @oas.tags('Foo', 'Bar')
+      class MyController {
+        @get('/greet')
+        greet() {
+          return 'Hello world!';
+        }
+
+        @get('/echo')
+        echo() {
+          return 'Hello world!';
+        }
+      }
+
+      const actualSpec = getControllerSpec(MyController);
+      expect(actualSpec.paths['/greet'].get.tags).to.eql(['Foo', 'Bar']);
+      expect(actualSpec.paths['/echo'].get.tags).to.eql(['Foo', 'Bar']);
+    });
+
+    it('Allows @oas.tags with options to append', () => {
+      @oas.tags('Foo')
+      class MyController {
+        @get('/greet')
+        greet() {
+          return 'Hello world!';
+        }
+
+        @get('/echo')
+        @oas.tags('Bar')
+        echo() {
+          return 'Hello world!';
+        }
+      }
+
+      const actualSpec = getControllerSpec(MyController);
+      expect(actualSpec.paths['/greet'].get.tags).to.eql(['Foo']);
+      expect(actualSpec.paths['/echo'].get.tags).to.eql(['Foo', 'Bar']);
+    });
+
+    it('Does not allow a member variable to be decorated', () => {
+      const shouldThrow = () => {
+        class MyController {
+          @oas.tags('foo', 'bar')
+          public foo: string;
+
+          @get('/greet')
+          greet() {}
+        }
+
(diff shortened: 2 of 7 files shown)


docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,61 @@ export class SomeController {
...
+
+[API document](https://loopback.io/doc/en/lb4/apidocs.openapi-v3.tags.html),
+[OpenAPI Operation Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object)
+
+This decorator can currently be applied to class and a class method. It will set
+the `tags` array string property of the Operation Object. When applied to a
+class, it will mark all operation methods of that class with those tags.
+
+Putting `@tags` on a method in a class that already has `@tags` will overwrite
+the tags for that path entry, unless you pass the options `{append: true}`
```
raymondfeng: Let's remove the `append` option. If a method decides to append, the `@tag` can list all names including the ones from class level.

docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,61 @@ export class SomeController {
...
+
+```ts
+@tags(['Foo', 'Bar'])
```
raymondfeng: I prefer to have `...names: string[]` so that we can allow `@tags('Foo', 'Bar')`.

mschnee: Looks good!

docs/site/decorators/Decorators_openapi.md
mschnee: Done!

raymondfeng:
I wonder if we should expose such decorators under a namespace (for example, `@oas.tags`) to avoid confusion against `@bind({tags: ...}`.

mschnee:
That's a question I don't myself have an opinion on as a developer.  I do know that sometimes, vscode intellisense will auto-import in a way I wasn't expecting: for example, typing `@get` might cause it to `import { get } from 'http';`, 

But at the same time, I could `import as` instead:
```ts
import * as oas from '@loopback/openapi-v3';
@oas.api({})
@oas.deprecated()
class MyController {
    @oas.get('/greet')
    public async greet() {}
}
```

My only opinion as a developer is to be able to use some convenience decorators so that I don't have to write and maintain such verbose json blocks.

docs/site/decorators/Decorators_openapi.md
raymondfeng: Please remove the `append` option.

docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,61 @@ export class SomeController {
...
+class MyController {
+  @tags(['Foo', 'Baz'])
+  @get('/greet')
+  public async greet() {
+    // tags will be [Foo, Baz]
+  }
+
+  @tags(['Baz'], {append: true})
```
raymondfeng: Ditto

docs/site/decorators/Decorators_openapi.md
raymondfeng: Let's use `@tags('Foo', 'Bar')` style consistently.

raymondfeng:
@bajtos Please review.

examples/todo/src/controllers/todo.controller.ts
```diff
@@ -15,11 +15,13 @@ import {
   post,
   put,
   requestBody,
+  tags,
 } from '@loopback/rest';
 import {Todo} from '../models';
 import {TodoRepository} from '../repositories';
 import {GeocoderService} from '../services';
 
+@tags(['TodoController'])
```
emonddr: @mschnee , why is the todo.controller.ts file in `examples` being changed? I think your controller classes in your mocha tests are sufficient to test the decorator.

mschnee: I was looking through the documentation and some other PRs to determing best practices for updating examples, but I agree, the documentation and tests are quite clear.  I'll clean this up.

What's the best way forward here on namespacing the decorators?  My first thought is to maintain the current exports (for maintenance), but also re-export them all under the new namespace as well, as they're related (for consistency):
```ts
export * from './api.decorator';
export * from './operation.decorator';
export * from './parameter.decorator';
export * from './request-body.decorator';
export * from './tags.decorator';

import {api} from './api.decorator';
import {del, get, operation, patch, post, put} from './operation.decorator';
import {param} from './parameter.decorator';
import {requestBody} from './request-body.decorator';
import {tags} from './tags.decorator';

export const oas = {
  api,
  operation,
  method: {
    get,
    post,
    del,
    patch,
    put,
  },
  param,
  requestBody,
  tags,
};
```
```

Comment (raymondfeng) [reply examples/todo/src/controllers/todo.controller.ts]: +1 to keep direct exports for most of the decorators.

```ts
export const oas = {
  tags,
  // The rest names are optional
  api,
  operation,
  get,
  post,
  del,
  patch,
  put,
  param,
  requestBody,
};
```

mschnee: Awesome, I'll do that :+1:

mschnee:
:point_up: I am attempting to resolve build issues after the most recent rebase.

packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
```diff
@@ -0,0 +1,141 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/openapi-v3
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
+import {expect} from '@loopback/testlab';
+import {getControllerSpec} from '../../..';
+import {oas} from '../../../decorators';
```
raymondfeng: This should be merged into the `import` as follows:
```ts
import {getControllerSpec, oas} from '../../..';
```

packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
```diff
@@ -0,0 +1,141 @@
...
+
+describe('@oas.tags decorator', () => {
+  context('Without a top-level @oas.api definition', () => {
+    it('Allows a class to not be decorated with @oas.tags at all', () => {
+      class MyController {
+        @oas.get('/greet')
```
raymondfeng: Should we keep using `@get` instead?

packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
mschnee: :+1:, guess auto-import didn't do it's thing correctly.

packages/openapi-v3/src/__tests__/unit/decorators/tags.decorator.unit.ts
mschnee: For the purposes of the unit test, it doesn't matter- though I intend to write code with the namespace consistently.

docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,61 @@ export class SomeController {
 When the OpenAPI spec is generated, the `xs-ts-type` is mapped to
 `{$ref: '#/components/schemas/MyModel'}` and a corresponding schema is added to
 `components.schemas.MyModel` of the spec.
+
+## Convenience Decorators
+
+While you can supply a fully valid OpenAPI specification for the class-level
+`@api` decorator, and full operation OpenAPI specification for `@op` and the
```
raymondfeng: `@op` -> `@operation`?

docs/site/decorators/Decorators_openapi.md
```diff
@@ -430,3 +430,61 @@ export class SomeController {
...
+other convenience decorators, there are also a number of utility decorators that
+allow you to supply specific OpenAPI information without requiring you to use
+verbose JSON.
+
+## OAS
```
raymondfeng: Maybe `## Shortcuts for OpenAPI Spec (OAS) Objects`?

dougal83:
Thinking about OAS enhancers, would it be a good/practical idea to add an optional description for each tag that would populate a sibling extension property?  This information can be used (and cleaned up) by enhancer that would [populate `spec.tags`](https://gist.github.com/dougal83/e6f0a96f878bf98cf530cf93fe309250#file-openapi-json-L1144-L1165).  Draft [tag enhancer](https://github.com/strongloop/loopback-next/pull/4539/commits/933b0b2509a001370ae1ec2e1a3baefb7fa9ae48#diff-64bb6d25b1cfcd9e40529109be544021R38-R47).

```json
  "tags": [
    "PingController"
  ],
  "x-tags": [
    {"description": "Api Ping response" }
  ],
```

raymondfeng:
@mschnee Please address my last two comments so that we can land this PR soon.

raymondfeng:
@dougal83 Please note there is a top level `tags` as an array of [TagObject](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#tagObject). The operation level `tags` is `string[]` containing references to top-level tags by name.

mschnee:
Busy week, sorry for the delay!  I am making the requested updated now.

@dougal83 I love the idea in general- the OAS spec has a top level `tags` block that allows for more extensive description.  You sen [peruse the spec documents here)[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#tag-object].  It's not required to reference a `tag` in the operation spec, though.  They can be the same, but the strings in a path tags are completely arbitrary.

For the first iteration of these convenience decorators, I'm trying to limit the scope to the direct developer experience of writing controllers.

mschnee:
Documentation updated.  

Thanks for your patience and quick response @raymondfeng !  I'm going to take all of your feedback here and apply it to my PRs for the `@deprecated` and `@response` decorators.

raymondfeng:
@mschnee PR landed. Great contribution! 🥇

PR #4417, opened 2020-01-13
feat: adds MultiMethodDecoratorFactory

mschnee:
Partially implements: #4300 
See: #4406

## Added
- **MultiMethodDecoratorFactory**.  This factory supports using multiple decorators on a method, storing and returning the metadata as an array.  This is in support of the `@response` decorator, which will be split from #4406 into a separate pull request.

## Examples
```ts
import {MethodMultiDecoratorFactory} from '@loopback/metadata';

interface Point {
  x?: number;
  y?: number;
  z?: number;
}

function myMultiMethodDecorator(spec: Point): MethodDecorator {
  return MethodMultiDecoratorFactory.createDecorator<Point>(
    'metadata-key-for-my-method-multi-decorator',
    spec,
  );
}

class MyController {
  @myMultiMethodDecorator({x: 1})
  @myMultiMethodDecorator({y: 2})
  @myMultiMethodDecorator({z: 3})
  public point() {}
}

class MyOtherController {
  @myMultiMethodDecorator([{x: 1}, {y: 2}, {z: 3}])
  public point() {}
}

const arrayOfPoints = MetadataInspector.getMethodMetadata<Point>(
  'metadata-key-for-my-method-multi-decorator',
  MyController.prototype,
  'point',
); // [{x:1}, {y:2}, {z: 3}]

const anotherArrayOfPoints = MetadataInspector.getMethodMetadata<Point>(
  'metadata-key-for-my-method-multi-decorator',
  MyOtherController.prototype,
  'point',
); // [{x:1}, {y:2}, {z: 3}]
```
## Checklist


.vscode/launch.json
```diff
@@ -18,6 +18,22 @@
         "0"
       ]
     },
+    {
```
raymondfeng: Let's make the `lauch.json` changes out of this PR. I assume it's not required to support `MultiMethodDecoratorFactory`.

mschnee: Will do, another PR incoming :)

packages/metadata/README.md
```diff
@@ -87,6 +87,92 @@ class MyController {
...
+  'metadata-key-for-my-method-multi-decorator',
+  constructor.prototype,
+  op,
+);
+
+// [{x:1}, {y:2}, {z: 3}]
+```
+
+Note that the order of items is **not** guaranteed and should be treated as
+unsorted.
```
raymondfeng: TypeScript applies decorators from bottom to up. We should ensure the order of metadata in the array follows that order.

packages/metadata/README.md
```diff
@@ -87,6 +87,92 @@ class MyController {
...
+```ts
+interface Point {
+  x?: number;
+  y?: number;
+  z?: number;
+}
+interface GeometryMetadata {
+  points: Point[];
+}
+function geometry(points: Point | Point[]): MethodDecorator {
```
raymondfeng: A better way is to allow `...points: Point[]`.

packages/metadata/README.md
mschnee: Will do, I'll update the tests and ensure that the order is correct in the inheritance cases as well.

packages/metadata/README.md
mschnee: Done!

mschnee:
Updates made- thanks for the quick feedback @raymondfeng !

packages/metadata/README.md
```diff
@@ -87,6 +87,113 @@ class MyController {
...
+function geometry(...points: Point[]): MethodDecorator {
+  return MethodMultiDecoratorFactory.createDecorator<GeometryMetadata>(
+    'metadata-key-for-my-method-multi-decorator',
+    points,
+  );
+}
+
+class MyGeoController {
+  @geometry({x: 1})
+  @geometry([{x: 2}, {y: 3}])
```
raymondfeng: It should be fixed as:
```ts
@geometry({x: 2}, {y: 3})
``

File: packages/metadata/README.md [hunk 5]
```diff
@@ -87,6 +87,113 @@ class MyController {
...
+  constructor.prototype,
+  op,
+);
+
+// [{z: 3}, {y: 2}, {x: 1}]
+```
+
+Typescript applies decorators in reverse order per class, from the parent down.
+In cases when an array is passed as metadata, it will be reversed to respect
+this order.
```
raymondfeng: I would rather not reverse the array - this magic may trip our users. Without the `reverse`, we can point them to TypeScript docs - https://www.typescriptlang.org/docs/handbook/decorators.html to understand how decorators are applied.

PR #4419, opened 2020-01-14
fix(cli): reject datasources with no name property for service generator

nabdelgadir:
When generating a service using `lb4 service`, the generator would look for a `name` property in the config file of the datasource and if it didn't include it, it would still generate the service with errors.

For example:

```ts
import {getService} from '@loopback/service-proxy';
import {inject, Provider} from '@loopback/core';
import {Error: bad inputDataSource} from '../datasources'; // notice this import statement

export interface Geocoder {
  // this is where you define the Node.js methods that will be
  // mapped to REST/SOAP/gRPC operations as stated in the datasource
  // json file.
}

export class GeocoderProvider implements Provider<Geocoder> {
  constructor(
    //  must match the name property in the datasource json file

   // **SEE THE FOLLOWING TWO LINES**
    @inject('datasources.')
    protected dataSource: Error: bad inputDataSource = new Error: bad inputDataSource(),
  ) {}

  value(): Promise<Geocoder> {
    return getService(this.dataSource);
  }
}

```

## Checklist

### DIFF
--- a/packages/cli/generators/service/index.js
+++ b/packages/cli/generators/service/index.js
@@ -233,6 +233,10 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
           this.artifactInfo.dataSourceClass,
         );
 
+        if (!this.artifactInfo.dataSourceName) {
+          throw new Error('Datasource config does not have `name` property');
+        }
+
         this.artifactInfo.dataSourceClassName =
           utils.toClassName(this.artifactInfo.dataSourceName) + 'DataSource';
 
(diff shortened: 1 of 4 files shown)


packages/cli/generators/service/index.js
```diff
@@ -233,6 +233,12 @@ module.exports = class ServiceGenerator extends ArtifactGenerator {
           this.artifactInfo.dataSourceClass,
         );
 
+        if (!this.artifactInfo.dataSourceName) {
+          return Promise.reject(
+            'Error: datasource config does not have `name` property',
+          );
+        }
```
raymondfeng: It's probably simpler to just `throw new Error('Datasource config does not have `name` property'');`.

Issue #3950, opened 2019-10-17
How to migrate remoting hooks


bajtos:
_This is a follow-up for #3718 and #3922._

Write content for `docs/site/migration/models/remoting-hooks.md`, explain how to migrate LB3 remoting hooks to LB4 interceptors.

- Global hooks, e.g. `app.beforeRemote('**', handlerFn)`
- Model-level hooks, e.g. `app.beforeRemote('User.**', handlerFn)`
- Method-level hooks, e.g. `User.beforeRemote('create', handlerFn)`

Explain how to map properties provided by LB3 remoting context to LB4 concepts
and how to access that data via Dependency Injection.

## Acceptance criteria

Migration guide describing manual steps and covering the following topics:

- [ ] Global hooks
- [ ] Model-level hooks
- [ ] Method-level hooks
- [ ] How to map LB3 context properties to LB4 concepts

Use (some of) the examples in LB3 doc page [Remote hooks](https://loopback.io/doc/en/lb3/Remote-hooks.html) to illustrate the migration process on a real code.

PR #4424, opened 2020-01-14
feat(cli): add static BINDING_KEY prop to non-global interceptors 

bajtos:
While working on #3950, I discovered that non-global interceptors are difficult to use via `@intercept` when they are implemented as provider classes. In this pull request, I am improving the CLI templates to add a static property `BINDING_KEY` when generating a new non-global interceptor.

At the moment, we have to use a magic string to reference an interceptor:

```ts
@inject('interceptors.FooBarInterceptor')
```

With my change in place, we can use the new static variable instead:

```ts
@inject(FooBarInterceptor.BINDING_KEY)
```

The pull request has two commits:
1. The first commit reworks interceptor tests to use snapshot-based assertions, no changes in the actual functionality as made.
2. The second commit modifies the template per the feature description above.

## Checklist

### DIFF
--- a/docs/site/Interceptor-generator.md
+++ b/docs/site/Interceptor-generator.md
@@ -121,8 +121,10 @@ import {
  * This class will be bound to the application as a global `Interceptor` during
  * `boot`
  */
-@bind({tags: {namespace: 'interceptors', name: 'test'}})
+@bind({tags: {key: TestInterceptor.BINDING_KEY}})
 export class TestInterceptor implements Provider<Interceptor> {
+  static readonly BINDING_KEY = `interceptors.${TestInterceptor.name}`;
+
   /*
   constructor() {}
   */
--- a/packages/cli/generators/interceptor/templates/interceptor-template.ts.ejs
+++ b/packages/cli/generators/interceptor/templates/interceptor-template.ts.ejs
@@ -19,9 +19,13 @@ import {
 <% if (isGlobal) { -%>
 @globalInterceptor('<%= group %>', {tags: {name: '<%= name %>'}})
 <% } else { -%>
-@bind({tags: {namespace: 'interceptors', name: '<%= name %>'}})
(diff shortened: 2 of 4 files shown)


packages/cli/generators/interceptor/templates/interceptor-template.ts.ejs
```diff
@@ -19,9 +19,13 @@ import {
 <% if (isGlobal) { -%>
 @globalInterceptor('<%= group %>', {tags: {name: '<%= name %>'}})
 <% } else { -%>
-@bind({tags: {namespace: 'interceptors', name: '<%= name %>'}})
+@bind({tags: {key: <%= className %>.BINDING_KEY}})
 <% } -%>
 export class <%= className %> implements Provider<Interceptor> {
+<% if (!isGlobal) { -%>
+  static readonly BINDING_KEY = `interceptors.${<%= className %>.name}`;
```
bajtos: Discussion point: Should we emit `BINDING_KEY` property for global interceptors too? Such interceptors are typically picked up by the booter and applied automatically on all requests, therefore I feel there is no need to access the actual binding key.

Thoughts? /cc @strongloop/loopback-next

docs/site/Interceptor-generator.md
```diff
@@ -121,8 +121,10 @@ import {
  * This class will be bound to the application as a global `Interceptor` during
  * `boot`
  */
-@bind({tags: {namespace: 'interceptors', name: 'test'}})
+@bind({tags: {key: TestInterceptor.BINDING_KEY}})
```
bajtos: Discussion point: 
- Before my change, we would scaffold the interceptor in such way that it's bound to `interceptors.test` key. 
- With my change in place, it will be bound to `interceptors.TestInterceptor` key.

My reasoning: the binding name should match the class name. When the class is renamed by the developer, the binding name should be updated accordingly.

Thoughts?

/cc @strongloop/loopback-next

PR #4432, opened 2020-01-14
feat: adds utility launch for debugging a single test file

mschnee:
Partially implements: #4300
See: #4406

This update adds a developer convenience Launch task to allow debugging a single unit test file in Visual Studio Code.  

## Adds
- **`bin/mocha-current-file.js`** - this is a utility script that attempts to determine the dist output of the given test file, and passes it off to mocha in-process.
- **VSCode: Debug Current Test File** launch configuration.  This launch configuration allows you to debug the current unit test file currently focused.  It uses the above helper to determine the correct name and path of the compiled JavaScript file to run.

## Modifies
- The launch for "Run Mocha Tests" so that it also runs on Windows.

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,6 +9,7 @@
       "request": "launch",
       "name": "Run Mocha tests",
       "program": "${workspaceRoot}/packages/build/node_modules/mocha/bin/_mocha",
+      "runtimeArgs": ["-r", "${workspaceRoot}/packages/build/node_modules/source-map-support/register"],
       "cwd": "${workspaceRoot}",
       "args": [
         "--config",
@@ -17,6 +18,25 @@
         "-t",
         "0"
       ]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Current Test File",
+      "program": "${workspaceRoot}/bin/mocha-current-file",
+      "runtimeArgs": ["-r", "${workspaceRoot}/packages/build/node_modules/source-map-support/register"],
+      "cwd": "${workspaceRoot}",
+      "args": [
+        "--config",
+        "${workspaceRoot}/packages/build/config/.mocharc.json",
+        "-t",
+        "0",
+        "${file}"
+      ],
+      // This ensures vscode won't set breakpoints until after sourcemaps
+      // have been loaded for the file.  Important when your tests are compiled
+      // from typescript.
+      "disableOptimisticBPs": true,
     },
     {
       "type": "node",
--- a/bin/mocha-current-file.js
+++ b/bin/mocha-current-file.js
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback-next
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/*
+================================================================================
+This is used in the launch.json to enable you to debug a test file written in
+typescript.  This function attempts to convert the passed typescript file to
+the best-guess output javascript file.
+
+It walks up the filesystem from the current file, stops at package.json, and
+looks in `dist`
+
+Ideally, we could somehow use the typescript compiler and tsconfig.json to get
+the explicit output file instead of trying to guess it, but there might be
+overhead.
+
+Ex:
+```jsonc
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Current Test File",
+      "program": "${workspaceRoot}/bin/mocha-current-file",
+      "runtimeArgs": ["-r", "${workspaceRoot}/packages/build/node_modules/source-map-support/register"],
+      "cwd": "${workspaceRoot}",
+      "autoAttachChildProcesses": true,
+      "args": [
+        "--config",
+        "${workspaceRoot}/packages/build/config/.mocharc.json",
+        "-t",
+        "0",
+        "${file}"
+      ],
+      "disableOptimisticBPs": true
+    }
+  ]
+}
+```
+================================================================================
+*/
+
+'use strict';
+const path = require('path');
+const fs = require('fs');
+
+function findDistFile(filename) {
+  const absolutePath = path.resolve(filename);
+  const systemRoot = path.parse(absolutePath).root;
+  let currentDir = path.dirname(absolutePath);
+
+  let isPackageRoot = fs.existsSync(path.resolve(currentDir, 'package.json'));
+  while (!isPackageRoot) {
+    if (path.dirname(currentDir) === systemRoot) {
+      throw new Error(
+        `Could not find a package.json file in the path heirarchy of ${absolutePath}`,
+      );
+    }
+    currentDir = path.join(currentDir, '..');
+    isPackageRoot = fs.existsSync(path.resolve(currentDir, 'package.json'));
+  }
+  const base = path.resolve(currentDir);
+  const relative = path.relative(currentDir, absolutePath);
+  const resultPath = relative.replace(/^src/, 'dist').replace(/\.ts$/, '.js');
+  return path.resolve(base, resultPath);
+}
+
+const newFile = findDistFile(process.argv.splice(-1)[0]);
+
+if (newFile) {
+  require('../packages/build/node_modules/mocha/lib/cli').main(
+    [...process.argv, newFile].slice(2),
+  );
+}


mschnee:
There are a couple of instances (e.g. debugging `packages/metadata/src/__tests__/decorator-factory.unit.ts`) where breakpoints don't get picked up as expected.  This could be the `"autoAttachChildProcesses": true,` in the launch configuration.

.vscode/launch.json
```diff
@@ -18,6 +18,22 @@
...
+      "name": "Debug Current Test File",
+      "program": "${workspaceRoot}/packages/build/node_modules/.bin/_mocha",
+      "cwd": "${workspaceRoot}",
+      "autoAttachChildProcesses": true,
+      "args": [
+        "--config",
+        "${workspaceRoot}/packages/build/config/.mocharc.json",
+        "-t",
+        "0",
+        "$(node ${workspaceRoot}/packages/build/bin/get-dist-file ${file})"
```
raymondfeng: I assume this won't work on Windows, right?

mschnee: This has been bugging me, so I made a few changes:
- "Run Mocha tests" now has the correct windows-specific configuration.
- replaced the "Debug Current Test File" subshell by writing a program that invokes the mocha cli directly.
- removed the `autoAttachChildProcess` setting.
- confirmed I could set a breakpoint in `packages/metadata/src/decorator-factory.ts` and vscode would stop on it when "Debug Current Test File" is run on `packages/metadata/src/__tests__/unit/decorator-factory.unit.ts`

mschnee:
NOTE: I have rebased from upstream and squashed all commits, which should address the travis commitlint and doc errors.

packages/build/bin/mocha-current-file.js
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env node
```
raymondfeng: If this file is only used for `loopback-next` repo, please move it to `loopback-next/bin`.

mschnee: Done!

bin/mocha-current-file.js
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env node
+
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: @loopback/build
```
raymondfeng: Let's fix the headers.

bin/mocha-current-file.js
```diff
@@ -0,0 +1,76 @@
...
+        "0",
+        "${file}"
+      ],
+      "disableOptimisticBPs": true
+    }
+  ]
+}
+```
+
+For your personal projects, you can invoke directly from @loopback/build:
```
raymondfeng: This is no longer true.

bin/mocha-current-file.js
mschnee: ...done?  It's not consistent in the files in this directory.

bin/mocha-current-file.js
mschnee: Removed.

.vscode/launch.json
```diff
@@ -9,14 +9,34 @@
...
       "args": [
         "--config",
         "${workspaceRoot}/packages/build/config/.mocharc.json",
         "packages/*/dist/__tests__/**/*.js",
         "-t",
         "0"
-      ]
+      ],
+      "windows": {
+        "program": "${workspaceRoot}/packages/build/node_modules/mocha/bin/mocha",
```
bajtos: I think this part is no longer needed, see https://github.com/strongloop/loopback-next/pull/4643/commits/992dfe98c4c10087ed30125e5e4faed734f8a890

.vscode/launch.json
```diff
@@ -9,14 +9,34 @@
       "request": "launch",
       "name": "Run Mocha tests",
       "program": "${workspaceRoot}/packages/build/node_modules/.bin/_mocha",
+      "runtimeArgs": ["-r", "${workspaceRoot}/packages/build/node_modules/source-map-support/register"],
```
bajtos: Nice 👍

.vscode/launch.json
```diff
@@ -9,14 +9,34 @@
...
+      "cwd": "${workspaceRoot}",
+      "args": [
+        "--config",
+        "${workspaceRoot}/packages/build/config/.mocharc.json",
+        "-t",
+        "0",
+        "${file}"
+      ],
+      "disableOptimisticBPs": true,
```
bajtos: Can you please add a comment to explain why is this flag needed?

mschnee:
Sorry for the delay in getting back to this PR!  I had a brief vacation.  Will update and address things :)

.vscode/launch.json
mschnee: Done!

.vscode/launch.json
mschnee: Done!

mschnee:
With approvals from @raymondfeng and @bajtos I'd be happy to merge this in :)

bin/mocha-current-file.js
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env node
+
+// Copyright IBM Corp. 2018. All Rights Reserved.
```
raymondfeng: 2018 -> 2020.

bin/mocha-current-file.js
```diff
@@ -0,0 +1,70 @@
...
+```jsonc
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Current Test File",
+      "program": "${workspaceRoot}/bin/mocha-current-file",
+      "cwd": "${workspaceRoot}",
```
derdeka: Should this block be identical to launch.json? runtimeArgs is missing.

.vscode/launch.json
bajtos: I think we misunderstood each other. I was trying to say that the `windows` section is no longer needed, because the main `program` setting was already updated to work on all platforms (Windows, Linux, MacOS).

I _think_ that your current config will actually break debugging Mocha tests on Windows, because `mocha` is just a thin wrapper that spawns another child process (running `_mocha` script).

Anyhow, let's remove the `windows` section please, it shouldn't be needed anymore.

bin/mocha-current-file.js
bajtos: Yes please, let's include `runtimeArgs` too.

bin/mocha-current-file.js
mschnee: Done!

bin/mocha-current-file.js
mschnee: Added!

.vscode/launch.json
mschnee: Fixed!

mschnee:
Addressing the commitlint changes :)

mschnee:
@derdeka What's the correct scope?  `core`?

bajtos:
> What's the correct scope? core?

`core` is for changing affecting `@loopback/core` package.

In this case, I think it's best to leave the scope out. Also please note that we are using imperative form ("add something"), not present tense ("adds something"). See [subject](https://loopback.io/doc/en/lb4/code-contrib-lb4.html#subject) in our Commit Message Guidelines and also [A Note About Git Commit Messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

I am proposing the following commit message:

```
feat: add a launch task for debugging a single test file
```

(Personally, I would choose `chore` instead of `feat` because this change is not visible to LoopBack users, but that does not really matter.)

bin/mocha-current-file.js
```diff
@@ -0,0 +1,71 @@
...
+  let isPackageRoot = fs.existsSync(path.resolve(currentDir, 'package.json'));
+  while (!isPackageRoot) {
+    currentDir = path.join(currentDir, '..');
+    isPackageRoot = fs.existsSync(path.resolve(currentDir, 'package.json'));
+  }
+  const base = path.resolve(currentDir);
+  const relative = path.relative(currentDir, absolutePath);
+  const resultPath = relative.replace(/^src/, 'dist').replace(/\.ts$/, '.js');
+  return path.resolve(base, resultPath);
+}
```
raymondfeng: Nitpick: a more robust implementation can be:
```js
// Find root directory of the file system
const root = path.parse(process.cwd()).root;

let pkgFile = path.resolve(filename, '../package.json');
while (!fs.existsSync(pkgFile)) {
  if (path.dirname(pkgFile) === root) {
    // We reach the root of file system
    pkgFile = undefined;
    break;
  }
  // Try the parent directory level package.json
  pkgFile = path.resolve(pkgFile, '../../package.json');
}

if (!pkgFile) return filename; // Or throw an error
const projectRoot = path.dirname(pkgFile);
const relative = path.relative(projectRoot, path.resolve(filename));
const jsFilePath = relative.replace(/^src/, 'dist').replace(/\.ts$/, '.js');
return path.resolve(projectRoot, jsFilePath);
```

mschnee: I made a small change to the current function, and tested it by invoking this cli against both a test file inside of the `loopback-next` workspace, and also a `fake.test.js` file that has no `package.json` file in it's filesystem heirarchy

bin/mocha-current-file.js
```diff
@@ -0,0 +1,80 @@
...
+
+function findDistFile(filename) {
+  const absolutePath = path.resolve(filename);
+  const systemRoot = path.parse(absolutePath).root;
+  let currentDir = path.dirname(absolutePath);
+
+  let isPackageRoot = fs.existsSync(path.resolve(currentDir, 'package.json'));
+  while (!isPackageRoot) {
+    currentDir = path.join(currentDir, '..');
+    if (path.dirname(currentDir) === systemRoot) {
```
raymondfeng: This test should happen before change `currentDir` to its parent.

mschnee: Good catch, updated :)

raymondfeng: Nitpick again :-). I think you code will throw if `/myproject` is the current dir and no `package.json` is found there. It's really an edge case. So feel free to move forward.

mschnee: Very edge case!  Since the idea is to run a single unit test inside one of the repo package directories, it's likely an expected error if you try to debug a file at the system room.

PR #4438, opened 2020-01-15
[WIP] feat(repository): hasManyThrough

derdeka:
Just another try / draft / poc to implement hasManyThrough
as the original implementation has birthday in a few days.

Early feedback welcome.

This PR is a continuation of #2359.

## Checklist

### DIFF
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
@@ -1,0 +1,99 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository-tests
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect, toJSON} from '@loopback/testlab';
+import {
+  CrudFeatures,
+  CrudRepositoryCtor,
+  CrudTestContext,
+  DataSourceOptions,
+} from '../../..';
+import {
+  deleteAllModelsInDefaultDataSource,
+  MixedIdType,
+  withCrudCtx,
+} from '../../../helpers.repository-tests';
+import {
+  Customer,
+  CustomerRepository,
+  Order,
+  OrderRepository,
+  Seller,
+  SellerRepository,
+} from '../fixtures/models';
+import {givenBoundCrudRepositories} from '../helpers';
+
+export function hasManyRelationAcceptance(
+  dataSourceOptions: DataSourceOptions,
+  repositoryClass: CrudRepositoryCtor,
+  features: CrudFeatures,
+) {
+  describe('HasManyThrough relation (acceptance)', () => {
+    before(deleteAllModelsInDefaultDataSource);
+    let customerRepo: CustomerRepository;
+    let orderRepo: OrderRepository;
+    let sellerRepo: SellerRepository;
+    let existingCustomerId: MixedIdType;
+
+    before(
+      withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+        ({customerRepo, orderRepo, sellerRepo} = givenBoundCrudRepositories(
+          ctx.dataSource,
+          repositoryClass,
+          features,
+        ));
+        await ctx.dataSource.automigrate([
+          Customer.name,
+          Order.name,
+          Seller.name,
+        ]);
+      }),
+    );
+
+    beforeEach(async () => {
+      await customerRepo.deleteAll();
+      await orderRepo.deleteAll();
+      await sellerRepo.deleteAll();
+    });
+
+    beforeEach(async () => {
+      existingCustomerId = (await givenPersistedCustomerInstance()).id;
+    });
+
+    it('can create related models', async () => {
+      // TODO(derdeka): creating seller though order is not the best example usecase - find alternative
+      const sellerData: Partial<Seller> = {
+        name: 'Domino’s Pizza',
+      };
+      const orderData: Partial<Order> = {
+        description: 'pizza',
+      };
+      const seller = await customerRepo
+        .sellers(existingCustomerId)
+        .create(sellerData, {
+          throughData: orderData,
+        });
+      expect(toJSON(seller)).containDeep(toJSON(sellerData));
+
+      // check target object
+      const persistedSeller = await sellerRepo.findById(seller.id);
+      expect(toJSON(persistedSeller)).containDeep(toJSON(sellerData));
+
+      // check through object
+      const persistedOrders: Order[] = await orderRepo.find({
+        where: {
+          sellerId: seller.id,
+          customerId: existingCustomerId,
+        },
+      });
+      expect(persistedOrders.length).to.eql(1);
+      expect(persistedOrders[0]).containDeep(toJSON(orderData));
+    });
+
+    async function givenPersistedCustomerInstance() {
+      return customerRepo.create({name: 'a customer'});
+    }
+  });
+}
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
@@ -25,7 +25,7 @@ import {
 } from '../fixtures/models';
 import {givenBoundCrudRepositories} from '../helpers';
 
-export function hasManyRelationAcceptance(
+export function hasManyThroughRelationAcceptance(
   dataSourceOptions: DataSourceOptions,
   repositoryClass: CrudRepositoryCtor,
   features: CrudFeatures,
--- a/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
+++ b/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
@@ -1,24 +1,68 @@
-import {Entity, EntityCrudRepository, Getter, HasManyThroughDefinition} from "../..";
-import {DefaultHasManyThroughRepository, HasManyThroughRepository} from "./has-many-through.repository";
+import {
+  DataObject,
+  Entity,
+  EntityCrudRepository,
+  Getter,
+  HasManyThroughDefinition,
+} from '../..';
+import {
+  createTargetConstraint,
+  createThroughConstraint,
+  resolveHasManyThroughMetadata,
+} from './has-many-through.helpers';
+import {
+  DefaultHasManyThroughRepository,
+  HasManyThroughRepository,
+} from './has-many-through.repository';
 
 export type HasManyThroughRepositoryFactory<
   TargetEntity extends Entity,
   TargetID,
   ThroughEntity extends Entity,
-  ThroughID,
-> = (fkValue: ThroughID) => HasManyThroughRepository<TargetEntity, TargetID, ThroughEntity>;
+  ForeignKeyType
+> = (
+  fkValue: ForeignKeyType,
+) => HasManyThroughRepository<TargetEntity, TargetID, ThroughEntity>;
 
 export function createHasManyThroughRepositoryFactory<
   TargetEntity extends Entity,
   TargetID,
   ThroughEntity extends Entity,
   ThroughID,
+  ForeignKeyType
 >(
   relationMetadata: HasManyThroughDefinition,
   targetRepositoryGetter: Getter<EntityCrudRepository<TargetEntity, TargetID>>,
-  throughRepositoryGetter: Getter<EntityCrudRepository<ThroughEntity, ThroughID>>,
-): HasManyThroughRepositoryFactory<TargetEntity, TargetID, ThroughEntity, ThroughID> {
-  return (fkValue?: ThroughID) => {
+  throughRepositoryGetter: Getter<
+    EntityCrudRepository<ThroughEntity, ThroughID>
+  >,
+): HasManyThroughRepositoryFactory<
+  TargetEntity,
+  TargetID,
+  ThroughEntity,
+  ForeignKeyType
+> {
+  const meta = resolveHasManyThroughMetadata(relationMetadata);
+  return (fkValue?: ForeignKeyType) => {
+    const getTargetContraint = (
+      throughInstances: ThroughEntity[],
+    ): DataObject<TargetEntity> => {
+      return createTargetConstraint<TargetEntity, ThroughEntity>(
+        meta,
+        throughInstances,
+      );
+    };
+    const getThroughConstraint = (
+      targetInstance?: TargetEntity,
+    ): DataObject<ThroughEntity> => {
+      const constriant: DataObject<ThroughEntity> = createThroughConstraint<
+        TargetEntity,
+        ThroughEntity,
+        ForeignKeyType
+      >(meta, fkValue, targetInstance);
+      return constriant;
+    };
+
     return new DefaultHasManyThroughRepository<
       TargetEntity,
       TargetID,
@@ -29,6 +73,8 @@ export function createHasManyThroughRepositoryFactory<
     >(
       targetRepositoryGetter,
       throughRepositoryGetter,
+      getTargetContraint,
+      getThroughConstraint,
     );
-  }
+  };
 }
--- a/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
+++ b/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
@@ -1,0 +1,34 @@
+import {Entity, EntityCrudRepository, Getter, HasManyThroughDefinition} from "../..";
+import {DefaultHasManyThroughRepository, HasManyThroughRepository} from "./has-many-through.repository";
+
+export type HasManyThroughRepositoryFactory<
+  TargetEntity extends Entity,
+  TargetID,
+  ThroughEntity extends Entity,
+  ThroughID,
+> = (fkValue: ThroughID) => HasManyThroughRepository<TargetEntity, TargetID, ThroughEntity>;
+
+export function createHasManyThroughRepositoryFactory<
+  TargetEntity extends Entity,
+  TargetID,
+  ThroughEntity extends Entity,
+  ThroughID,
+>(
+  relationMetadata: HasManyThroughDefinition,
+  targetRepositoryGetter: Getter<EntityCrudRepository<TargetEntity, TargetID>>,
+  throughRepositoryGetter: Getter<EntityCrudRepository<ThroughEntity, ThroughID>>,
+): HasManyThroughRepositoryFactory<TargetEntity, TargetID, ThroughEntity, ThroughID> {
+  return (fkValue?: ThroughID) => {
+    return new DefaultHasManyThroughRepository<
+      TargetEntity,
+      TargetID,
+      EntityCrudRepository<TargetEntity, TargetID>,
+      ThroughEntity,
+      ThroughID,
+      EntityCrudRepository<ThroughEntity, ThroughID>
+    >(
+      targetRepositoryGetter,
+      throughRepositoryGetter,
+    );
+  }
+}
--- a/packages/repository/src/relations/has-many/has-many-through.helpers.ts
+++ b/packages/repository/src/relations/has-many/has-many-through.helpers.ts
@@ -1,0 +1,51 @@
+import {DataObject, Entity, HasManyThroughDefinition} from '../..';
+
+export type HasManyThroughResolvedDefinition = HasManyThroughDefinition;
+
+/**
+ * Creates constraint used to query target
+ * @param relationMeta - hasManyThrough metadata to resolve
+ * @param throughInstances - Instances of through entities used to constrain the target
+ * @internal
+ */
+export function createTargetConstraint<
+  Target extends Entity,
+  Through extends Entity
+>(
+  relationMeta: HasManyThroughResolvedDefinition,
+  throughInstances: Through[],
+): DataObject<Target> {
+  throw new Error('Not implemented');
+}
+
+/**
+ * Creates constraint used to query through
+ * @param relationMeta - hasManyThrough metadata to resolve
+ * @param fkValue - Value of the foreign key used to constrain through
+ * @param targetInstance - Instance of target entity used to constrain through
+ * @internal
+ */
+export function createThroughConstraint<
+  Target extends Entity,
+  Through extends Entity,
+  ForeignKeyType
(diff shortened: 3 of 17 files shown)


packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -98,3 +100,75 @@ export interface HasManyThroughRepository<
...
+  TargetEntity extends Entity,
+  TargetID,
+  TargetRepository extends EntityCrudRepository<TargetEntity, TargetID>,
+  ThroughEntity extends Entity,
+  ThroughID,
+  ThroughRepository extends EntityCrudRepository<ThroughEntity, ThroughID>,
+  > implements HasManyThroughRepository<TargetEntity, TargetID, ThroughEntity> {
+
+  constructor(
+    public targetRepositoryGetter: Getter<TargetRepository>,
```
KalleV: Should the constructor pass in a target/through constraint like the original reference PR (for example: https://github.com/strongloop/loopback-next/pull/2359/files#diff-a35f109a989bf5b2ac1206db18756a0cR63) or is it intentionally omitted to avoid implementation details in the initial PR?

derdeka: Thanks for reviewing. I tried different solutions if there is another way to get the constraint. But giving up for now and copied from the referenced PR.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -262,6 +265,38 @@ export class DefaultCrudRepository<
...
+    relationName: string,
+    targetRepositoryGetter: Getter<EntityCrudRepository<TargetEntiy, TargetID>>,
+    throughRepositoryGetter: Getter<
+      EntityCrudRepository<ThroughEntity, ThroughID>
+    >,
+  ): HasManyThroughRepositoryFactory<
+    TargetEntiy,
+    TargetID,
+    ThroughEntity,
+    ForeignKeyType
```
derdeka: Not sure about `ForeignKeyType`. Do we really need `ThroughID` and `ForeignKeyType`?
@KalleV PTAL

KalleV: @derdeka I think the ForeignKeyType is needed. It is consistent with the HasManyRepositoryFactory typing. For example:

```
export class UserRepository<
  User,
  typeof User.prototype.id
> {
  public readonly groups: HasManyThroughRepositoryFactory<
    Group,
    UserGroup,
    typeof Group.prototype.id
  >;
}
```

bajtos:
Thank you @derdeka for the pull request. 

@agnes512 & @hacksparrow can you please review?

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -98,3 +105,129 @@ export interface HasManyThroughRepository<
     },
   ): Promise<void>;
 }
+
+export class DefaultHasManyThroughRepository<
```
agnes512: Let's add API docs for it.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -98,3 +105,129 @@ export interface HasManyThroughRepository<
...
+  async delete(
+    where?: Where<TargetEntity>,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<Count> {
+    const targetRepository = await this.getTargetRepository();
+    const throughRepository = await this.getThroughRepository();
+    const throughConstraint = this.getThroughConstraint();
+    // TODO(derdeka): How to delete throughInstances?
```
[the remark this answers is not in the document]
KalleV: I think it should delete the through instance too. When the target is deleted, the FK in the through model will no longer reference a valid target model.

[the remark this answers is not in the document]
derdeka: On a relational database this can be easily done by a fk trigger `on delete cascade`. But what should happen on a non-relational database or if there is no trigger?

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -0,0 +1,99 @@
...
+      };
+      const orderData: Partial<Order> = {
+        description: 'pizza',
+      };
+      const seller = await customerRepo
+        .sellers(existingCustomerId)
+        .create(sellerData, {
+          throughData: orderData,
+        });
+      expect(toJSON(seller)).containDeep(toJSON(sellerData));
```
derdeka: Creating a `seller` though `order` is not the best example / test usecase for `hasManyThrough`.
Can anyone give a hint, which models / usecase we could use?

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -0,0 +1,99 @@
...
+  Customer,
+  CustomerRepository,
+  Order,
+  OrderRepository,
+  Seller,
+  SellerRepository,
+} from '../fixtures/models';
+import {givenBoundCrudRepositories} from '../helpers';
+
+export function hasManyRelationAcceptance(
```
hacksparrow: Why do need to export this function? Also, I don't see any place where it is called.

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -0,0 +1,99 @@
...
+  describe('HasManyThrough relation (acceptance)', () => {
+    before(deleteAllModelsInDefaultDataSource);
+    let customerRepo: CustomerRepository;
+    let orderRepo: OrderRepository;
+    let sellerRepo: SellerRepository;
+    let existingCustomerId: MixedIdType;
+
+    before(
+      withCrudCtx(async function setupRepository(ctx: CrudTestContext) {
+        ({customerRepo, orderRepo, sellerRepo} = givenBoundCrudRepositories(
```
hacksparrow: Why enclose within `()`?

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
derdeka: Same as other relation tests, which are picked up by:
https://github.com/strongloop/loopback-next/blob/098dc971f1fcb00321ca9226a2475c62aa5d00d0/packages/repository-tests/src/crud-test-suite.ts#L57-L58

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
derdeka: Short notation of
```ts
const {customerRepo, orderRepo, sellerRepo} = givenBoundCrudRepositories(
  ctx.dataSource,
  repositoryClass,
  features,
);
```
same as here:
https://github.com/strongloop/loopback-next/blob/098dc971f1fcb00321ca9226a2475c62aa5d00d0/packages/repository-tests/src/crud/relations/acceptance/has-many.relation.acceptance.ts#L42

packages/repository/src/relations/relation.types.ts
```diff
@@ -87,12 +87,12 @@ export interface HasManyThroughDefinition extends RelationDefinitionBase {
   /**
    * The foreign key in the source model, e.g. Customer#id.
    */
-  keyFrom: string;
+  keyFrom?: string;
```
hacksparrow: Why make these keys optional? If justified, this change should be PR of its own to minimize the size of the PR.

derdeka: I'll happy to extract this into a PR if there is agreement in overall structure.

amitgiri0001:
Do we have any ETA for this?

dhmlau:
@derdeka, if your PR is ready for review, could you please make it a non-draft PR? Thanks!

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+  if (!isTypeResolver(relationMeta.target)) {
+    const reason = 'target must be a type resolver';
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+  if (!relationMeta.through) {
+    const reason = 'through must be specified';
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+  if (!isTypeResolver(relationMeta.through?.model)) {
```
bajtos, line 97: `through` is always set by this point, I think it's ok to use regular property access here.

```suggestion
  if (!isTypeResolver(relationMeta.through.model)) {
```

Unless you are trying to be extra defensive in case `through` is a truthy but non-object value? In that case it's better to make the check on line 93 more strict, e.g.

```ts
if (typeof relationMeta.through !== 'object' || relationMeta.through == null) {
  const reason = 'through must be a non-null object';
  throw new InvalidRelationError(reason, relationMeta);
}

File: packages/repository/src/relations/has-many/has-many-through.helpers.ts [hunk 10]
```diff
@@ -0,0 +1,176 @@
...
+    const reason = 'through.model must be a type resolver';
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  const throughModel = relationMeta.through.model();
+  const throughModelProperties = throughModel.definition?.properties;
```
bajtos, line 103: According to our types, model's `definition` property is always set. Under which condition do you expect `throughModel.definition` to not be an object?

Can we simplify this line as follows?

```suggestion
  const throughModelProperties = throughModel.definition.properties;
```

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  const targetModel = relationMeta.target();
+  const targetModelProperties = targetModel.definition?.properties;
```
bajtos, line 106: Ditto, can we simplify?

```suggestion
  const targetModelProperties = targetModel.definition.properties;
```

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  // check if metadata is already complete
+  if (
+    relationMeta.through?.keyTo &&
+    throughModelProperties[relationMeta.through.keyTo] &&
+    relationMeta.through?.keyFrom &&
```
bajtos, line 112:
```suggestion
    relationMeta.through.keyTo &&
    throughModelProperties[relationMeta.through.keyTo] &&
    relationMeta.through.keyFrom &&
```

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+    const reason = `through model ${throughModel.name} is missing definition of source foreign key ${sourceFkName}`;
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  const targetFkName =
+    relationMeta.through?.keyTo ?? camelCase(targetModel.modelName + '_id');
+  if (!throughModelProperties[targetFkName]) {
+    const reason = `through model ${throughModel.name} is missing definition of target foreign key ${targetFkName}`;
+    throw new InvalidRelationError(reason, relationMeta);
+  }
```
bajtos, line 151: Are the default cases covered by tests? I checked `packages/repository/src/__tests__/unit/repositories/has-many-through-repository-factory.unit.ts` and didn't see any test verifying what `through.keyFrom` and `through.keyTo` are being created by default.

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+  );
+
+  debug(
+    'Resolved model %s from given metadata: %o',
+    throughModel.modelName,
+    throughModel,
+  );
+
+  const sourceFkName =
+    relationMeta.through?.keyFrom ?? camelCase(sourceModel.modelName + '_id');
```
bajtos, line 140:
```suggestion
    relationMeta.through.keyFrom ?? camelCase(sourceModel.modelName + '_id');
```

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (!throughModelProperties[sourceFkName]) {
+    const reason = `through model ${throughModel.name} is missing definition of source foreign key ${sourceFkName}`;
+    throw new InvalidRelationError(reason, relationMeta);
+  }
+
+  const targetFkName =
+    relationMeta.through?.keyTo ?? camelCase(targetModel.modelName + '_id');
```
bajtos, line 147:
```suggestion
    relationMeta.through.keyTo ?? camelCase(targetModel.modelName + '_id');
```

packages/repository/src/relations/has-many/has-many-through.repository.ts
bajtos: +1, also include "experimental feature" warning.

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
bajtos: Alternatively, we can reverse the relation to "Seller has many Customer through Order". I think in such case it's more reasonable to create a new customer via the seller?

packages/repository/src/relations/has-many/has-many-through.repository.ts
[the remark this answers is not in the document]
bajtos: IMO, this should delete through instances first and then delete the target instances second. That way:
1. We avoid violating referential integrity enforced by SQL databases by deleting target models that are still references by through models.
2. We avoid leaving orphaned "through" model instances in the databases. By an orphaned instance I mean an instance that's referencing a target model that no longer exists.

Please make sure to include acceptance-level test for this feature!

Here are two scenario to consider & cover by tests, I'll use "Source", "Target" and "Through" as model names instead of Customer/Seller/Order.

1. Given through links: `{sourceId: 1, targetId: 1}` and `{sourceId: 2, targetId: 2}`.
   When I delete all models belonging to source id 1,
   then the target model id 1 and the first link is deleted.

2. Given trough links `{sourceId: 1, targetId: 1}`, `{sourceId: 2, targetId: 1}` and `{sourceId: 2, targetId: 2}`. 
  When I delete all models belonging to source id 1, 
  then the target model id 1 and the first two links are deleted. Only `{sourceId: 2, targetId: 2}` link is preserved.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -98,3 +105,129 @@ export interface HasManyThroughRepository<
...
+      constrainWhere(where, targetConstraint as Where<TargetEntity>),
+      options,
+    );
+  }
+
+  async patch(
+    dataObject: DataObject<TargetEntity>,
+    where?: Where<TargetEntity>,
+    options?: Options & {
+      throughOptions?: Options;
```
bajtos, line 197: Do we want to allow partial `throughData` too, so that the patch operation can update both the target and the through model instances?

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -98,3 +105,129 @@ export interface HasManyThroughRepository<
...
+  }
+
+  async unlink(
+    targetModelId: TargetID,
+    options?: Options & {
+      throughOptions?: Options;
+    },
+  ): Promise<void> {
+    throw new Error('Method not implemented.');
+  }
```
bajtos, line 232: What are your plans for these two not-implemented methods?

I slightly prefer to include their implementation in the initial pull request to verify that the current design of `DefaultHasManyThroughRepository` has all necessary pieces required to implement link/unlink functionality.

However, if you prefer to leave this out, then I am fine with that too, as long as this limitation is clearly documented somewhere and we have a follow-up task (or a new issue) tracked on GitHub.

packages/repository/src/relations/relation.types.ts
bajtos: I am ok with these changes, I see they are consistent with what we already have in place for `HasManyDefinition` (see above in the source code file).

I agree a new PR would allow us to land these fixes sooner, on the other hand such PR would be a bit anemic - since no code is using `HasManyThroughDefinition` yet, there will be no way how to verify correctness of such small PR.

In that light, I think it may be better better to keep these changes inside this bigger PR.

packages/repository/src/relations/relation.types.ts
```diff
@@ -87,12 +87,12 @@ export interface HasManyThroughDefinition extends RelationDefinitionBase {
...
 
   /**
    * The primary key of the target model, e.g Seller#id.
    */
-  keyTo: string;
+  keyTo?: string;
 
   through: {
```
bajtos, line 97: A side note: in the future, I would like to add support for convenience shortcuts, e.g.

```ts
@hasMany(() => Seller, {through: () => Order})
```

See https://github.com/strongloop/loopback-next/pull/2359#discussion_r317012754

This is out of scope of this pull request though.

packages/repository/src/relations/has-many/has-many.decorator.ts
```diff
@@ -17,7 +21,7 @@ import {HasManyDefinition, RelationType} from '../relation.types';
  */
 export function hasMany<T extends Entity>(
   targetResolver: EntityResolver<T>,
-  definition?: Partial<HasManyDefinition>,
+  definition?: Partial<HasManyDefinition | HasManyThroughDefinition>,
 ) {
   return function(decoratedTarget: object, key: string) {
     const meta: HasManyDefinition = Object.assign(
```
bajtos, line 27: This is not correct for "has-many-through" case, is it?

See the discussion in for https://github.com/strongloop/loopback-next/pull/2359#discussion_r317011708 for more context.

packages/repository/src/relations/relation.types.ts
```diff
@@ -87,12 +87,12 @@ export interface HasManyThroughDefinition extends RelationDefinitionBase {
...
    * The foreign key in the source model, e.g. Customer#id.
    */
-  keyFrom: string;
+  keyFrom?: string;
 
   /**
    * The primary key of the target model, e.g Seller#id.
    */
-  keyTo: string;
+  keyTo?: string;
```
bajtos, line 95: While we are discussing `HasManyThroughDefinition`. I am not sure if it's actually a good idea to have such interface, see the discussion in https://github.com/strongloop/loopback-next/pull/2359#discussion_r317011708 for more context.

Having two different interfaces for hasMany and hasManyThrough relations is already creating extra complexity as can be seen in my comment above related to `@hasMany()` decorator.

Cross-posting from https://github.com/strongloop/loopback-next/pull/2359#discussion_r317011708:

> So far, we are using `type` to discriminate between different interfaces inheriting from `RelationDefinitionBase`. See https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions
>
> If we want to keep using `hasMany` value for has-many-through relations, then I am fine with that, but let's capture that in the type system too.
>
> We can either get rid of `HasManyThroughDefinition` completely and move all members to `HasManyDefinition`, or else modify `HasManyThroughDefinition` to inherit from `HasManyDefinition`.

Personally, I like most the option where there is a single `HasManyDefinition` interface with an optional `through` field that's used to distinguish between hasMany and hasManyThrough cases. Similarly, `HasManyResolvedDefinition` can contain an optional `through` field that's used to distinguish hasManyThrough case, this field can use a type like `ThroughResolvedDefinition` to tell the compiler that optional properties were resolved and will be always present.

I think it would be best to extract the changes in relation definitions, the implementation bits of resolving through relations and the accompanying unit tests into a new pull request, to get a smaller patch that's easier to review & iterate on.

packages/repository/src/relations/relation.types.ts
bajtos: Cross-posting from https://github.com/strongloop/loopback-next/pull/4438#discussion_r402117145:

> I think it would be best to extract the changes in relation definitions, the implementation bits of resolving through relations and the accompanying unit tests into a new pull request, to get a smaller patch that's easier to review & iterate on.

I think this is the best way how to get a PR changing relation definition that's smaller but still meaningful.

bajtos:
@codejamninja Ping 👋 Since you are the author of https://github.com/strongloop/loopback-next/pull/2359, which kick-started the implementation of hasManyThrough, would you like to join the discussion and/or review the proposed implementation too?

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,80 @@
...
+  createTargetConstraint,
+  createThroughConstraint,
+  resolveHasManyThroughMetadata,
+} from './has-many-through.helpers';
+import {
+  DefaultHasManyThroughRepository,
+  HasManyThroughRepository,
+} from './has-many-through.repository';
+
+export type HasManyThroughRepositoryFactory<
```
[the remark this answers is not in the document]
achrinza: @bajtos We might want to also utilize the tsdoc [@beta modifier](https://api-extractor.com/pages/tsdoc/tag_beta/).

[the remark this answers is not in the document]
bajtos: I think `@alpha` would be more appropriate, see https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags

>  Indicates that an API item is eventually intended to be public, but currently is in an early stage of development. Third parties should not use “alpha” APIs.

AFAIK, we are not using these modifiers yet, I have no idea how are they rendered in our API docs (if they are rendered at all).

MacGyver26:
Hello guys !

Do you know when this feature is going to be available in the framework ? thanks in advance.

atilerturk:
> Hello guys !
> 
> Do you know when this feature is going to be available in the framework ? thanks in advance.

Another 10 years

agnes512:
hasManyThrough helpers and definitions have been modified/implemented in #5354 . Will work on factory next.

agnes512:
The code this PR covered are being landed, though the HasManyThrough relation is still in an experimental stage (a bug found in https://github.com/strongloop/loopback-next/issues/5852). The documentation will be added in the near future.

PR #4430, opened 2020-01-14
feat(context): make bindings as event emitters to report changes

raymondfeng:
Spin-off from #4377 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4451, opened 2020-01-17
feat(context): improve context listener and events

raymondfeng:
Spin-off from #4377 

1. Add `ContextEventListener` interface
2. Tidy up parent context event handling
3. Set `maxListeners` to `Infinity` by default

## Checklist

### DIFF
--- a/docs/site/Context.md
+++ b/docs/site/Context.md
@@ -240,7 +240,7 @@ the form of metadata) to your intent.
 
 ## Context events
 
-The `Context` emits the following events:
+An instance of `Context` can emit the following events:
 
 - `bind`: Emitted when a new binding is added to the context.
   - binding: the newly added binding object
@@ -252,11 +252,50 @@ The `Context` emits the following events:
   process
   - err: the error object thrown
 
+The bind/unbind events are represented as the following type:
+
+```ts
+/**
+ * Events emitted by a context
+ */
+export type ContextEvent = {
+  /**
+   * Source context that emits the event
+   */
+  context: Context;
+  /**
+   * Binding that is being added/removed/updated
+   */
+  binding: Readonly<Binding<unknown>>;
+  /**
+   * Event type
+   */
+  type: string; // 'bind' or 'unbind'
+};
+```
+
 When an existing binding key is replaced with a new one, an `unbind` event is
 emitted for the existing binding followed by a `bind` event for the new binding.
 
 If a context has a parent, binding events from the parent are re-emitted on the
 context when the binding key does not exist within the current context.
+
+A context event listener should conform to the following signature:
+
+```ts
+/**
+ * Synchronous event listener for the `Context` as en event emitter
+ */
+export type ContextEventListener = (event: ContextEvent) => void;
+```
+
+By default, `maxListeners` is set to `Infinity` for context objects to avoid
+[memory leak warnings](https://github.com/strongloop/loopback-next/issues/4363).
+The value can be reset as follows:
+
+```ts
+ctx.setMaxListeners(128);
+```
 
 ## Context observers
 
@@ -275,7 +314,7 @@ come and go. There are a few caveats associated with that:
      .bind('foo')
      .to('foo-value')
      .tag('foo-tag');
-   ctx.on('bind', binding => {
+   ctx.on('bind', {binding} => {
      console.log(binding.tagNames); // returns an empty array `[]`
    });
    ```
@@ -293,7 +332,7 @@ come and go. There are a few caveats associated with that:
      .to('foo-value')
      .tag('foo-tag');
    ctx.add(binding);
-   ctx.on('bind', binding => {
+   ctx.on('bind', {binding} => {
      console.log(binding.tagMap); // returns `['foo-tag']`
    });
    ```
@@ -343,6 +382,15 @@ export type ContextEventObserver = ContextObserver | ContextObserverFn;
 ```
 
 If `filter` is not required, we can simply use `ContextObserverFn`.
+
+Please note that `ContextEventObserver` is different from
+`ContextEventListener`:
+
+- A `ContextEventListener` is synchronous and it's invoked when the event is
+  emitted (before `emit` returns).
+
+- A `ContextEventObserver` is asynchronous and it's invoked by the notification
+  queue after the event is emitted (after `emit` returns).
 
 2. Context APIs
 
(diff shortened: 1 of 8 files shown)


packages/context/src/context.ts
```diff
@@ -548,11 +554,10 @@ export class Context extends EventEmitter {
       });
       this.notificationQueue = undefined;
     }
-    if (this._parent && this._parentEventListeners) {
-      for (const [event, listener] of this._parentEventListeners) {
-        this._parent.removeListener(event, listener);
-      }
-      this._parentEventListeners = undefined;
+    if (this._parent && this.parentEventListener) {
+      this._parent.removeListener('bind', this.parentEventListener);
```
jannyHou: The old code removes all the listeners, any reason why only moves the bind and unbind now?

raymondfeng: The old code uses a map for registered listeners. With this PR, only one function is used on bind/unbind events.

packages/context/src/context.ts
```diff
@@ -161,8 +167,23 @@ export class Context extends EventEmitter {
...
+        binding: Readonly<Binding<unknown>>,
+        context: Context,
+        event: string,
+      ) => {
+        this.handleParentEvent(binding, context, event);
+      };
+
+      // Listen on the parent context events
+      this._parent.on('bind', this.parentEventListener!);
+      this._parent.on('unbind', this.parentEventListener!);
```
bajtos: Should we also listen on the `changed` event introduced by https://github.com/strongloop/loopback-next/pull/4430 and propagate it via `parentEventListener`?

docs/site/Context.md
```diff
@@ -258,6 +258,27 @@ emitted for the existing binding followed by a `bind` event for the new binding.
 If a context has a parent, binding events from the parent are re-emitted on the
```
bajtos: The list above says that the `Context` emits `bind`, `unbind` and `error` events. I think we need to add `changed` event to that list?

docs/site/Context.md
```diff
@@ -258,6 +258,27 @@ emitted for the existing binding followed by a `bind` event for the new binding.
...
 context when the binding key does not exist within the current context.
 
+A context event listener should conform to the following signature:
```
bajtos: I find this new content a bit out of context. What is a "context event listener"? This concept hasn't been introduced and explained to readers yet.

Also as I commented https://github.com/strongloop/loopback-next/pull/4377#discussion_r364745654, we should explain our users what is the difference between  `ContextObserver` and `ContextEventListener` & when to use which one.

packages/context/src/context.ts
```diff
@@ -53,6 +53,15 @@ import {iterator, multiple} from 'p-event';
...
 const debug = debugFactory('loopback:context');
 
+/**
+ * Synchronous event listener for the `Context` as en event emitter
+ */
+export type ContextEventListener = (
+  binding: Readonly<Binding<unknown>>, // Binding to be added or removed
+  context: Context, // The source context
+  event: string, // 'bind' or 'unbind'
+) => void;
```
bajtos: Positional arguments are not very easy to use when the number of arguments is more than 1-2, plus they are difficult to extend with optional arguments in the future. Can we rework the signature to use an object with named properties please?

```ts
export interface ContextEvent {
  binding: Readonly<Binding<unknown>>; // Binding to be added or removed
 context: Context; // The source context
 event: string; // 'bind' or 'unbind'
}

export type ContextEventListener = (event: ContextEvent) => void;
```

We can use object de-structuring to keep usage easy:

```ts
function myListener({binding, context, event}: ContextEvent) {
  // ...
}
```

packages/context/src/context.ts
```diff
@@ -357,9 +363,9 @@ export class Context extends EventEmitter {
     this.registry.set(key, binding);
     if (existingBinding !== binding) {
       if (existingBinding != null) {
-        this.emit('unbind', existingBinding, this);
+        this.emit('unbind', existingBinding, this, 'unbind');
```
bajtos: Why do we have to repeat the event name in two places (the first and the last argument)? I find that cumbersome and making it easy to accidentally introduce mistakes in the future.

packages/context/src/__tests__/unit/context-observer.unit.ts
```diff
@@ -21,8 +21,8 @@ const setImmediateAsync = promisify(setImmediate);
  * for assertions
  */
 class TestContext extends Context {
-  get parentEventListeners() {
-    return this._parentEventListeners;
+  get eventListener() {
```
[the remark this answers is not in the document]
raymondfeng: Fixed by redeclaring the property as public in the subclass.

packages/context/src/context.ts
raymondfeng: Please note `changed` events are only emitted by bindings at the moment. We're yet to introduce `bindingChanged` event at context level.

packages/context/src/context.ts
raymondfeng: Making the event type available in the `ContextEvent` object allows one function to listen on multiple event types to minimize overhead. In a strongly-typed language such as Java, the event class can be used as the discriminator.

raymondfeng:
@bajtos Your comments have been addressed. PTAL.

packages/context/src/context.ts
bajtos: I see the point. Although JavaScript is not a strongly-typed language, right?

Have you considered adding a helper function?

```ts
// usage
this.emitBindingEvent('unbind', existingBinding);

// impl
emitBindingEvent(name: string, binding: Binding) {
  this.emit(name, binding, this, name);
}
```

packages/context/src/context.ts
```diff
@@ -357,9 +387,13 @@ export class Context extends EventEmitter {
...
       if (existingBinding != null) {
-        this.emit('unbind', existingBinding, this);
+        this.emitEvent('unbind', {
+          binding: existingBinding,
+          context: this,
+          type: 'unbind',
+        });
       }
-      this.emit('bind', binding, this);
+      this.emitEvent('bind', {binding, context: this, type: 'bind'});
```
bajtos: Can `emitEvent` read the event name from `event.type`?

docs/site/Context.md
```diff
@@ -293,7 +332,7 @@ come and go. There are a few caveats associated with that:
      .to('foo-value')
      .tag('foo-tag');
    ctx.add(binding);
-   ctx.on('bind', binding => {
+   ctx.on('bind', {binding} => {
```
bajtos: Hmm, I didn't realize we already have context-event-listener contract with positional arguments. I think the switch from positional arguments to ContextEvent object can break existing code. Do we want to release this as a semver-major version?

raymondfeng: Marked it as breaking change.

PR #4514, opened 2020-01-27
refactor(repository): alter CountSchema title case

dougal83:
alter CountSchema title to be consistent with repository-json-schema titles

See also #4240

## Checklist

### DIFF
--- a/packages/repository/src/common-types.ts
+++ b/packages/repository/src/common-types.ts
@@ -100,7 +100,7 @@ export interface Count {
  */
 export const CountSchema = {
   type: 'object',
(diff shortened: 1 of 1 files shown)


packages/repository/src/common-types.ts
```diff
@@ -100,7 +100,7 @@ export interface Count {
  */
 export const CountSchema = {
   type: 'object',
-  title: 'loopback.count',
+  title: 'Loopback.Count',
```
raymondfeng: The namespace is preferred to be lower case. In this case, I would like to see `loopback.Count`.

PR #4515, opened 2020-01-27
fix(cli): fix hasMany relation controller artifacts

agnes512:
Fix relation controller. Currently it's using the **default relation name** in the template. It's supposed to use the relation name defined by CLI. 
A test is modified to verify the new change.
Also delete a redundant test in this PR.

BelongsTo controller has the correct setup : https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/relation/belongs-to-relation.generator.js#L49
## Checklist

### DIFF
--- a/packages/cli/generators/relation/has-many-relation.generator.js
+++ b/packages/cli/generators/relation/has-many-relation.generator.js
@@ -44,9 +44,8 @@ module.exports = class HasManyRelationGenerator extends BaseRelationGenerator {
     this.artifactInfo.targetModelRequestBody = utils.camelCase(
       this.artifactInfo.targetModelName,
     );
-    this.artifactInfo.relationPropertyName = utils.pluralize(
-      utils.camelCase(options.destinationModel),
-    );
+    this.artifactInfo.relationPropertyName = options.relationName;
+
     this.artifactInfo.sourceModelPrimaryKey = options.sourceModelPrimaryKey;
     this.artifactInfo.sourceModelPrimaryKeyType =
       options.sourceModelPrimaryKeyType;
--- a/packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
+++ b/packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
@@ -198,7 +198,7 @@ export class CustomerOrderController {
 `;
 
 exports[
-  `lb4 relation HasMany checks if the controller file created  answers {"relationType":"hasMany","sourceModel":"CustomerClass","destinationModel":"OrderClass"} checks controller content with hasMany relation 1`
+  `lb4 relation HasMany checks if the controller file created  answers {"relationType":"hasMany","sourceModel":"CustomerClass","destinationModel":"OrderClass","relationName":"myOrders"} checks controller content with hasMany relation 1`
 ] = `
 import {
   Count,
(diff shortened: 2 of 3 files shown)


packages/cli/generators/relation/has-many-relation.generator.js
```diff
@@ -44,9 +44,8 @@ module.exports = class HasManyRelationGenerator extends BaseRelationGenerator {
     this.artifactInfo.targetModelRequestBody = utils.camelCase(
       this.artifactInfo.targetModelName,
     );
-    this.artifactInfo.relationPropertyName = utils.pluralize(
-      utils.camelCase(options.destinationModel),
-    );
+    this.artifactInfo.relationPropertyName = options.relationName;
```
jannyHou: is `utils.pluralize` and `utils.camelCase` still needed?

packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
```diff
@@ -307,123 +307,7 @@ export class CustomerClassOrderClassController {
     @param.path.number('id') id: number,
     @param.query.object('where', getWhereSchemaFor(OrderClass)) where?: Where<OrderClass>,
   ): Promise<Count> {
-    return this.customerClassRepository.orderClasses(id).delete(where);
-  }
-}
-
-`;
-
-exports[
```
jannyHou: Hmm any reason removing this exports?

packages/cli/test/integration/generators/relation.has-many.integration.js
```diff
@@ -313,11 +313,7 @@ describe('lb4 relation HasMany', function() {
         relationType: 'hasMany',
         sourceModel: 'CustomerClass',
         destinationModel: 'OrderClass',
-      },
-      {
-        relationType: 'hasMany',
```
jannyHou: same here, any reason remove it?

packages/cli/generators/relation/has-many-relation.generator.js
agnes512: Yes, we need them for path names and model/repo names.

packages/cli/test/integration/generators/relation.has-many.integration.js
agnes512: This test case is the same as 
```
      {
        relationType: 'hasMany',
        sourceModel: 'Customer',
        destinationModel: 'Order',
      },
```
as a default value case. They are the same except the class names.

packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
agnes512: A snapshot matches one test case. This one is the test case that I removed.

packages/cli/generators/relation/has-many-relation.generator.js
jannyHou: oh 🙈 then why they are removed in the new code?

packages/cli/test/integration/generators/relation.has-many.integration.js
jannyHou: ah I see, fair enough.

packages/cli/generators/relation/has-many-relation.generator.js
agnes512: Sorry I thought you meant the import declaration. 

This is how we defined **default relation name** before. But it didn't respect the relation name we got from the prompt:
```
    this.artifactInfo.relationPropertyName = utils.pluralize(
      utils.camelCase(options.destinationModel),
    );
```
And the new change would allow the customized relation name.

jannyHou: oh I see. 👍

Issue #4530, opened 2020-01-28
CORS origin custom function throw and crash the app


ludohenin:
## Steps to reproduce

Setting the origin to a custom function make the app crash if an error is passed to the callback function.

```js
corsConfig.origin = function customOrigin(origin, callback) {
  if (whitelist.indexOf(origin) !== -1) {
    callback(null, true);
  } else {
    callback(new HttpErrors.Forbidden('Not allowed by CORS'));
  }
};
```

## Current Behavior

```bash
/home/ludohen/github/ta-api-core-2/node_modules/@loopback/rest/dist/rest.server.js:616
            throw err;
```

## Expected Behavior

send a response with the correct error code.

PR #4532, opened 2020-01-28
fix(rest): improves error handling for express middleware

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/4530

## Checklist

### DIFF
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -23,12 +23,13 @@ import {
 import {AssertionError} from 'assert';
 import cors from 'cors';
 import debugFactory from 'debug';
-import express from 'express';
+import express, {ErrorRequestHandler} from 'express';
 import {PathParams} from 'express-serve-static-core';
 import {IncomingMessage, ServerResponse} from 'http';
 import {ServerOptions} from 'https';
 import {safeDump} from 'js-yaml';
 import {ServeStaticOptions} from 'serve-static';
+import {writeErrorToResponse} from 'strong-error-handler';
 import {BodyParser, REQUEST_BODY_PARSER_TAG} from './body-parsers';
 import {HttpHandler} from './http-handler';
 import {RestBindings} from './keys';
@@ -237,11 +238,34 @@ export class RestServer extends Context implements Server, HttpServerLike {
     });
 
     // Mount our error handler
-    this._expressApp.use(
-      (err: Error, req: Request, res: Response, next: Function) => {
-        this._onUnhandledError(req, res, err);
-      },
-    );
+    this._expressApp.use(this._unexpectedErrorHandler());
+  }
+
+  /**
+   * Get an Express handler for unexpected errors
+   */
+  protected _unexpectedErrorHandler(): ErrorRequestHandler {
+    const handleUnExpectedError: ErrorRequestHandler = (
+      err,
+      req,
+      res,
+      next,
+    ) => {
+      // Handle errors reported by Express middleware such as CORS
+      // First try to use the `REJECT` action
+      this.get(SequenceActions.REJECT, {optional: true})
+        .then(reject => {
+          if (reject) {
+            // TODO(rfeng): There is a possibility that the error is thrown
+            // from the `REJECT` action in the sequence
+            return reject({request: req, response: res}, err);
+          }
+          // Use strong-error handler directly
+          writeErrorToResponse(err, req, res);
+        })
+        .catch(unexpectedErr => next(unexpectedErr));
+    };
+    return handleUnExpectedError;
   }
 
   /**
@@ -734,7 +758,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
   }
 
   /**
-   * Update or rebuild OpenAPI Spec object to be appropriate for the context of a specific request for the spec, leveraging both app config and request path information.
+   * Update or rebuild OpenAPI Spec object to be appropriate for the context of
+   * a specific request for the spec, leveraging both app config and request
+   * path information.
    *
    * @param spec base spec object from which to start
    * @param requestContext request to use to infer path information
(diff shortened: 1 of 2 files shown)


packages/rest/src/rest.server.ts
```diff
@@ -884,18 +884,25 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+      // The error can be thrown from cors middleware
+      // See https://github.com/strongloop/loopback-next/issues/4530
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      res.status((err as any).statusCode || 500).send({error: err});
+      return;
     }
 
-    // It's the responsibility of the Sequence to handle any errors.
-    // If an unhandled error escaped, then something very wrong happened
-    // and it's best to crash the process immediately.
```
bajtos: This removed code comment is explaining how was the original HTTP server designed. IMO, the problem described by #4530 is caused by the way how we hacked CORS support into our HTTP server, ignoring the Sequence-based design.

My main objection against the proposed changes is that it creates two kinds of error responses: errors handled inside a Sequence will return different HTTP response body than errors reported by additional middleware.

For longer term, I would like us to execute all middleware (including CORS) as a new sequence action, see https://github.com/strongloop/loopback-next/issues/1293

To fix the problem described in https://github.com/strongloop/loopback-next/issues/4530, I am proposing to mount a new Express error handler middleware just after the CORS middleware.

```ts
    // Allow CORS support for all endpoints so that users
    // can test with online SwaggerUI instance
    this._expressApp.use(cors(this.config.cors));

    this._expressApp.use(function handleCorsErrors(err, request, response, next) {
      // Handle CORS-related errors.
      // First try to obtain the error handler from the application.
      // This works well if application's sequence uses SequenceActions.REJECT
      // to resolve the error handler, as applications scaffolded by lb4 do.
      // When the app is using a different approach to Sequence and/or error handling,
      // then we will probably produce a different response body, but I guess that's inevitable
      // right now.
      ctx.get(SequenceActions.REJECT, {optional: true})
        .then(reject => {
          if (reject) {
            return reject({request, response}, err)
          }
          // App does not provide any standard error handler. What to do now?
          // Maybe it's ok to crash the app?
          next(err);
          // Alternatively, we can use strong-error handler directly
          writeErrorToResponse(err, request, response);
        })
        .catch(unexpectedErr => next(unexpectedErr));
    });

    // Set up endpoints for OpenAPI spec/ui
    this._setupOpenApiSpecEndpoints();

    // etc.
```

packages/rest/src/rest.server.ts
```diff
@@ -884,18 +884,25 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
-    process.nextTick(() => {
-      throw err;
-    });
+    // https://expressjs.com/en/guide/error-handling.html
+    // So when you add a custom error handler, you must delegate to the
+    // default Express error handler, when the headers have already been sent
+    // to the client:
```
bajtos: Please note that our custom error handler is specifically designed to detect the situation when the headers have been already sent, therefore this comment does not apply to a typical LB4 application.

Error handling in Express is a bit messy and difficult to get right in the sense that the server should return the same HTTP response body regardless of where the error originates from, don't leak any sensitive information and yet provide helpful information in debug mode. That's why we have carefully designed our own error handling schema, both in LB3 and LB4.

Let's make sure we preserve the benefits of the current design.

packages/rest/src/rest.server.ts
raymondfeng: I generalized your idea into an Express error handler for all middleware.

packages/rest/src/rest.server.ts
raymondfeng: Now the unexpected error is handled by the same Express error handler for CORS.

raymondfeng:
@bajtos PTAL.

packages/rest/src/rest.server.ts
```diff
@@ -236,11 +237,20 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
     // Mount our error handler
-    this._expressApp.use(
-      (err: Error, req: Request, res: Response, next: Function) => {
-        this._onUnhandledError(req, res, err);
-      },
-    );
+    const errorHandler: ErrorRequestHandler = (err, req, res, next) => {
+      // Handle errors reported by Express middleware such as CORS
+      // First try to use the `REJECT` action
+      this.get(SequenceActions.REJECT, {optional: true})
```
bajtos: Can you please move this code to `_onUnhandledError` function (see below)? The function `_setupRequestHandlerIfNeeded` is intended to register different middleware, not to implement middleware handlers.

raymondfeng:
> I am also a bit concerned that by invoking the new error handler also for errors leaked by the sequence, we may end up with calling reject two times:
First, we call reject from the sequence. Something unexpected happens and an error is thrown.
Then we call reject again from the last-resort error handler.

Good question. I think it's a rare edge case and probably not too bad to call it twice.

packages/rest/src/rest.server.ts
```diff
@@ -236,11 +237,28 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
-      },
-    );
+    this._expressApp.use(this._unexpectedErrorHandler());
+  }
+
+  /**
+   * Get an Express handler for unexpected errors
+   */
+  protected _unexpectedErrorHandler(): ErrorRequestHandler {
+    return (err, req, res, next) => {
```
bajtos: Please use a named function, so that if any code inside this handler throws an error, we have a useful stack frame in the error stack trace. That's why the original code uses `=>` function when calling `app.use` and then deferring to `this._onUnhandledError`.

```suggestion
    return function handleUnexpectedError(err, req, res, next) {
```

raymondfeng: `const functionName = (arg) => {...}` is a better way to name an arrow function. See https://stackoverflow.com/questions/27977525/how-do-i-write-a-named-arrow-function-in-es2015/37488652.

Issue #3343, opened 2019-07-12
Column names in lowercase using lb4 discover method - SQL connector


sureshkodur:
## Describe

Hi Team. I'm generating models from SQL db using **`lb4 discover`** method. But, the column names generating in **lowercase**. While in my SQL db, they are in camelCase. See, below code:
```code
@property({
    type: Number,
    required: true,
    precision: 10,
    scale: 0,
    id: 1,
    mssql: 
         {"columnName":"ErrorID",
          "dataType":"int",
          "dataLength":null,
          "dataPrecision":10,
          "dataScale":0,"
          nullable":"NO"
         },
  })
  errorid: Number;
```

## Current behavior
 The column name is generated in lowercase as `errorid` instead of `ErrorID`.

## Expected Behavior
- Would like to get the column name `ErrorID` as it is. 

## Acceptance Criteria
- [ ] column names should default to camelCase instead of lowercase
- [ ] add a prompt of 2 options when doing `lb4 discover` 
  - column names to property names:
    - `camelCase ( LB4 default, recommended)`
    - `keep it the same as the column names ( warning: might cause unexpected behavior. please read the ....file)`
- [ ] enhance the documentation. List out options and their limitations.

_Out of scope_:
- [ ]  an extension point allowing users to provide custom function to transform database column names to LB4 property names

sureshkodur:
@bajtos , @emonddr , @raymondfeng, @ritch , Can anyone, any update on this. Help me

dhmlau:
@agnes512, could you please take a look? Thanks.

agnes512:
I reproduced the above issue by using `lb4 discover --schema testdb` with datasource config:
 ```
{
  "name": "db",
  "connector": "mysql",
   ...
  "database": "testdb"
}
```
I can access database with the generated models. I figured that this issue is more about developer-experience instead of a bug.
As for the naming convention feature, we need to discuss about it. 
Will link this issue to other similar issues.

sureshkodur:
@agnes512 , I didn't understand... can you explain in details?

agnes512:
@sureshkodur Hi! I was testing it out yesterday. And it does generate `errorid` instead of `ErrorID`. If it has errors because the naming convention, we would label it as a bug. But it works fine for me. Are you getting errors causing by the names?

sureshkodur:
[...]

PR #4540, opened 2020-01-29
feat(cli): allow different naming convention for discover

agnes512:
connects to #3343 

Add 2 prompts:
- asking naming convention for db column names: **UPDATED**
```
? Select a convention to convert db column names(EXAMPLE_COLUMN) to model property names:
    ❯ Camel case (exampleColumn) (Recommend) 
    No conversion (EXAMPLE_COLUMN) 
```
If the user pick `keep the same as db`, prompt:
```
? You might need to specify these customized names in relation 
definition. Would you like to continue?
  ❯ Yes 
  No (back to the previous step) 
```

## Checklist

### DIFF
--- a/docs/site/Discovering-models.md
+++ b/docs/site/Discovering-models.md
@@ -43,3 +43,19 @@ placed. Default is `src/models`
 
 `--schema`: Specify the schema which the datasource will find the models to
 discover
+
+### Interactive Prompts
+
+Based on the option, the tool may prompt you for:
+
+- **Name of the connector to discover**: Prompts a list of available
+  connectors(datasources) to choose.
+- **Name of the models to discover**: Prompts choices of available models. The
+  answer can be multiple.
+- **Database column naming convention**: By default, LoopBack converts
+  discovered model properties to `camelCase`. This is recommended. You can
+  choose to keep them the same as the database column names. However, we
+  recommend to use LoopBack default convention. You might need to specify the
+  discovered property names in relation definition later. Check the
+  [Relation Metadata](HasMany-relation.md#relation-metadata) section in each
+  relation for details of customizing names.
--- a/packages/cli/generators/discover/index.js
+++ b/packages/cli/generators/discover/index.js
@@ -60,6 +60,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    * @returns {*}
    */
   setOptions() {
+    /* istanbul ignore next */
     if (this.options.dataSource) {
       debug(`Data source specified: ${this.options.dataSource}`);
       this.artifactInfo.dataSource = modelMaker.loadDataSourceByName(
@@ -74,6 +75,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    * Ensure CLI is being run in a LoopBack 4 project.
    */
   checkLoopBackProject() {
+    /* istanbul ignore next */
     if (this.shouldExit()) return;
     return super.checkLoopBackProject();
   }
@@ -83,6 +85,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    */
   async loadAllDatasources() {
     // If we have a dataSourcePath then it is already loaded for us, we don't need load any
+    /* istanbul ignore next */
     if (this.artifactInfo.dataSource) {
       return;
     }
@@ -106,6 +109,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    * Ask the user to select the data source from which to discover
    */
   promptDataSource() {
+    /* istanbul ignore next */
     if (this.shouldExit()) return;
     const prompts = [
       {
@@ -120,6 +124,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
     ];
 
     return this.prompt(prompts).then(answer => {
+      /* istanbul ignore next */
       if (!answer.dataSource) return;
       debug(`Datasource answer: ${JSON.stringify(answer)}`);
 
@@ -133,12 +138,16 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    * Puts all discoverable models in this.modelChoices
    */
   async discoverModelInfos() {
+    /* istanbul ignore if */
     if (this.artifactInfo.modelDefinitions) return;
     debug(`Getting all models from ${this.artifactInfo.dataSource.name}`);
 
     this.modelChoices = await modelMaker.discoverModelNames(
       this.artifactInfo.dataSource,
-      {views: this.options.views, schema: this.options.schema},
+      {
+        views: this.options.views,
+        schema: this.options.schema,
+      },
     );
     debug(
       `Got ${this.modelChoices.length} models from ${this.artifactInfo.dataSource.name}`,
@@ -151,6 +160,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
    */
   promptModelChoices() {
     // If we are discovering all we don't need to prompt
+    /* istanbul ignore next */
     if (this.options.all) {
       this.discoveringModels = this.modelChoices;
     }
@@ -172,6 +182,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
     ];
 
     return this.prompt(prompts).then(answers => {
+      /* istanbul ignore next */
       if (!answers.discoveringModels) return;
       debug(`Models chosen: ${JSON.stringify(answers)}`);
       this.discoveringModels = [];
@@ -204,6 +215,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
         default: false,
       },
     ]).then(props => {
+      /* istanbul ignore next */
       if (!props.disableCamelCase) return;
       props.disableCamelCase = props.disableCamelCase !== 'camelCase';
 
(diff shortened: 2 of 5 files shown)


packages/cli/generators/discover/index.js
```diff
@@ -180,6 +191,69 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
...
+
+      Object.assign(this.artifactInfo, props);
+      debug(`props after naming convention prompt: ${props.disableCamelCase}`);
+      return props;
+    });
+  }
+
+  /**
+   * Warns about potential issues for non-LB-default property naming convention.
+   * The prcess can be aborted here.
```
jannyHou: nitpick: process

packages/cli/generators/discover/index.js
```diff
@@ -180,6 +191,69 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
...
+   */
+  promptConfirmDisable() {
```
jannyHou: a suggestion: maybe display the warning in the previous prompt? Instead of having two prompts?

packages/cli/generators/discover/index.js
```diff
@@ -171,6 +181,7 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
     ];
 
     return this.prompt(prompts).then(answers => {
+      /* istanbul ignore next */
```
jannyHou: I found multiple ignores added, are they necessary?

packages/cli/snapshots/integration/generators/discover.integration.snapshots.js
```diff
@@ -57,6 +57,53 @@ export type ViewWithRelations = View & ViewRelations;
...
+    type: 'number',
+    id: 1,
+    required: true,
+  })
+  ID: number;
+
+  @property({
+    type: 'number',
+  })
+  snake_case?: number;
```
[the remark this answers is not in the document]
agnes512: The property type? This test case focuses on the property name so I just use type number for all :p

packages/cli/generators/discover/index.js
agnes512: I was thinking that too. But I don't know how to do it :sweat_smile:. Let me try to put them in `validation`.

packages/cli/snapshots/integration/generators/discover.integration.snapshots.js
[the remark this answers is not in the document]
jannyHou: I see fair enough~ (the lowercase is a boolean 🙈)

packages/cli/generators/discover/index.js
agnes512: CLI files are a bit different from others. We don't really test the CLI themselves. And the coverage would complain about the not covered if statements in these files. 
I am okay to remove them just that the coverage test would fail.

packages/cli/generators/discover/index.js
```diff
@@ -180,6 +191,69 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
     });
   }
 
+  /**
+   * Prompts what naming convention they would like to have for column names.
+   */
+  promptColNamingConvention() {
+    this.namingConvention = {
+      'LoopBack default camelCase convention (Recommend)': false,
```
raymondfeng: `Recommend` -> `Recommended`?

packages/cli/generators/discover/index.js
agnes512: I am still having 2 prompts. And I made some changes, now users could go back to the previous step to choose the column naming convention instead of getting aborted. WDYT?

packages/cli/generators/discover/index.js
```diff
@@ -180,6 +191,66 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
     });
   }
 
+  /**
+   * Prompts what naming convention they would like to have for column names.
+   */
+  promptColNamingConvention() {
+    this.namingConvention = {
+      'LoopBack default camelCase convention (Recommended)': false,
+      'keep property names the same as database column names': true,
```
raymondfeng: Using boolean value for different choices is a bit confusing. We can use `camelCase` and `noCase`. In the future, maybe we should prompt a list of case conversions, such as:

```
> Camel case (exampleColumn) 
> Pascal case (...)
> No conversion
> Lower case
> Upper case
```

docs/site/Discovering-models.md
```diff
@@ -43,3 +43,20 @@ placed. Default is `src/models`
 
 `--schema`: Specify the schema which the datasource will find the models to
 discover
+
+### Interactive Prompts
+
+After typing `lb4 discover --<your option> [--<arg>]`, based on the option, the
```
dhmlau: Looking at other CLI docs page, it uses: 
>The tool will prompt you for:

_extracted from https://loopback.io/doc/en/lb4/Model-generator.html#interactive-prompts_ 

Do you wanna follow that?

docs/site/Discovering-models.md
```diff
@@ -43,3 +43,20 @@ placed. Default is `src/models`
...
+tool may prompt you for:
+
+- **Name of the connector to discover**: Prompts a list of available
+  connectors(datasources) to choose.
+- **Name of the models to discover**: prompt choices of available models. The
```
dhmlau: `prompt` -> `Prompts`?

docs/site/Discovering-models.md
```diff
@@ -43,3 +43,20 @@ placed. Default is `src/models`
...
+  answer can be multiple.
+- **Database column naming convention**: By default, LoopBack converts
+  discovered model properties to `camelCase`. You can choose to keep them the
```
dhmlau: Suggestion:  
By default, LoopBack converts discovered model properties to `camelCase`. This is recommended. You can choose to...

Reading [Raymond's comment](https://github.com/strongloop/loopback-next/pull/4540/files#r373672012), this might need to change too.

packages/cli/generators/discover/index.js
agnes512: I use boolean to represent `disableCamelCase`. When it prompts, it only shows the messages as options 
```
    ❯ LoopBack default camelCase convention (Recommend)  // maps to disableCamelCase = false
    keep property names the same as database column names // maps to disableCamelCase = true
```

I agree that ideally we can prompt a list of different conventions. I think boolean is probably enough for now. How about improving the prompt messages to:
```
❯  camelCase ( Recommended)
No conversion ( same as database column names)
```

docs/site/Discovering-models.md
agnes512: Changed to `Based on the option, the tool may prompt you for:` 
Because for option `--all`, it doesn't ask the name of models to discover. WDYT?

packages/cli/generators/discover/index.js
raymondfeng: Do you want to give an example for each choice? Such as `first-name (column)` to `firstName (property)`?

I know your intention to directly map the two choices to `false` and `true`. It's just not very intuitive. Personally, I would rather use `disableCamelCase = (answer !== 'camelCase')`.

agnes512:
Is there anything I need to add according to the PR https://github.com/strongloop/loopback-next/pull/3479? I am thinking to close that PR and add anything needed in this PR.

packages/cli/generators/discover/index.js
```diff
@@ -180,6 +191,66 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
...
   }
 
+  /**
+   * Prompts what naming convention they would like to have for column names.
+   */
+  promptColNamingConvention() {
+    this.namingConvention = [
+      'Camel case (exampleColumn) (Recommended)',
+      'No conversion (EXAMPLE_COLUMN)',
+    ];
```
raymondfeng: Nitpick, it would be even better to have:
```ts
this.namingConvention = [
      {
        name: 'Camel case (exampleColumn) (Recommended)', 
        value: 'camelCase',
      },
      {
        name: 'No conversion (EXAMPLE_COLUMN)', 
        value: 'noCase',
      },
    ];
```

packages/cli/generators/discover/index.js
raymondfeng: I think it's a bit overkill to have a prompt to confirm the 2nd choice. We can make the warning part of the choice.

jannyHou: Yeah I tend to merge the warning to the first prompt. The warning message could be:

> By disabling Camel case, you might need to specify these customized names in relation definition. 

Usually the warning is part of the prompt message, you can use a different color (use `chalk` module) to emphasize and distinguish the warning from the prompt message.

See `chalk`'s example usage in https://github.com/strongloop/generator-loopback/blob/4c9c2068a65aa8a2ee3d47c00efff0b80762fd6c/model/index.js#L107

packages/cli/generators/discover/index.js
```diff
@@ -180,6 +192,47 @@ module.exports = class DiscoveryGenerator extends ArtifactGenerator {
...
+      },
+    ]).then(props => {
+      /* istanbul ignore next */
+      if (!props.disableCamelCase) return;
+      props.disableCamelCase = props.disableCamelCase !== 'camelCase';
+
+      Object.assign(this.artifactInfo, props);
+      /* istanbul ignore next */
+      if (props.disableCamelCase) {
+        console.warn(
```
raymondfeng: Let's use `this.log`

Issue #4386, opened 2020-01-08
Add bearer auth scheme as the default security scheme


jannyHou:
## Suggestion

After story https://github.com/strongloop/loopback-next/issues/4380 finished, we can add a security spec enhancer to have the bearer auth scheme as the default(or built-in) security scheme. So that explorer has the authorization dialog for people to inject the token per request.

## Use Cases

Add an OAI enhancer that add the following spec into OpenAPI spec generated in the rest server:

```ts
"components": {
    "securitySchemes": {
      "jwt": {
        "type": "http",
        "scheme": "bearer",
        "bearerFormat": "JWT"
      }
    },
```

## Examples

See the screenshot in https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#specifying-the-security-settings-in-the-openapi-specification

## Acceptance criteria

- [ ] add a security spec enhancer to have the bearer auth scheme as the default(or built-in) security scheme

dougal83:
Good idea to add to develop the schema.  Personally I'd be more specific and name it `jwt` rather than `bearerAuth`.

````json
      "jwt": {
        "type": "http",
        "scheme": "bearer",
        "bearerFormat": "JWT"
      }
````

jannyHou:
@dougal83 ah, true `jwt` is more accurate 👍

emonddr:
@dougal83 , @jannyHou ,

Regarding the comment

> Good idea to add to develop the schema. Personally I'd be more specific and name it jwt rather than bearerAuth.  

Exactly what are we talking about here?

The `title` of the github issue : 
`Add bearer auth scheme as the default security scheme` ?

Or

![image](https://user-images.githubusercontent.com/6864736/72074807-94d02500-32c0-11ea-99d7-e19e839e137a.png)

in https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#specifying-the-security-settings-in-the-openapi-specification ?

Because the JSON examples above do not have 
`bearerAuth`

Please clarify.

Thanks

:)

dougal83:
Hey @emonddr 

This issue is to enhance the openApi spec by adding to components.

I've just jumped on it to suggest a name change from `bearerAuth` to `jwt`.  I'm aware that `bearerAuth` is currently in use so if consensus for change is found then all instances would need to be updated.  The `bearerFormat` property is just arbitrary and so using `jwt` as security scheme name would be better IMO.

[...]

Issue #4380, opened 2020-01-07
Add OpenAPI enhancer service in @loopback/rest


jannyHou:
## Suggestion

This is a follow-up story for https://github.com/strongloop/loopback-next/pull/4258, after creating the oai enhancer service as extension point, we should add the service in `@loopback/rest` module so that other enhancers could register and contribute OpenAPI specs into the application(server).

## Use Cases

Inside function `getApiSpec()`, you can apply an enhancer like:

```ts
getApiSpec (requestContext?: RequestContext): OpenApiSpec {
  // ...other code
  spec = this.OAIEnhancer.applyEnhancerByName('authorization-bearer');
}
```

## Examples

Same as use cases

## Acceptance criteria

- [ ] add an `OASEnhancerService` in the [rest server](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L110), as how it has other properties like `_requestHandler`.

- [ ] add tests for registering and invoking enhancers

- [ ] (**Stretch Goal**, feel free to open new stories)if any existing spec contribution code in https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest.server.ts#L702 should be organized into enhancers, then refactor them into enhancers

dhmlau:
@strongloop/loopback-next @strongloop/loopback-maintainers @mschnee 

Call for contribution: 
This task is part of the [epic "Allow out-of-box token based authentication in API Explorer" ](https://github.com/strongloop/loopback-next/issues/4391), that we wish to get it done in 2020Q1. If you're interested in working on it, please leave a message here and we'll assign it to you.  We'll take the first person who responds. 😬

Happy contributing!

dougal83:
#4539 related pull.

dougal83:
@jannyHou I've had a cursory look over the code and I do have a query on how to proceed with implementation.  How would I handle the sync call to an async function(is there an example to study?)

The sync example above:
```ts
getApiSpec (requestContext?: RequestContext): OpenApiSpec {
  // ...other code
  spec = this.OASEnhancer.applyEnhancerByName('authorization-bearer');
}
```
The code in `spec-enhancer.service.ts`:
```ts
  async applyEnhancerByName(name: string): Promise<OpenApiSpec> {
    const enhancer = await this.getEnhancerByName(name);
    if (enhancer) this._spec = enhancer.modifySpec(this._spec);
    return this._spec;
  }
```
[...]

PR #4554, opened 2020-02-03
feat(rest): add OAS Enhancer service

dougal83:
Add OpenAPI enhancer service in @loopback/rest

Implements #4365

## Checklist

### DIFF
--- a/examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
+++ b/examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
@@ -138,7 +138,7 @@ describe('CoffeeShopApplication', () => {
     let apiSpec: OpenApiSpec;
 
     before(async () => {
-      apiSpec = app.lbApp.restServer.getApiSpec();
+      apiSpec = await app.lbApp.restServer.getApiSpec();
     });
 
     it('has the same properties in both the LB3 and LB4 specs', () => {
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -10,6 +10,7 @@ import {
   Constructor,
   Context,
   ContextObserver,
+  createBindingFromClass,
   filterByKey,
   filterByTag,
   inject,
@@ -19,6 +20,8 @@ import {Application, CoreBindings, Server} from '@loopback/core';
 import {HttpServer, HttpServerOptions} from '@loopback/http-server';
 import {
   getControllerSpec,
+  OASEnhancerService,
+  OAS_ENHANCER_SERVICE,
   OpenAPIObject,
   OpenApiSpec,
   OperationObject,
@@ -134,6 +137,12 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param res - The response.
    */
 
+  protected _OASEnhancer: OASEnhancerService;
+  public get OASEnhancer(): OASEnhancerService {
+    this._setupOASEnhancerIfNeeded();
+    return this._OASEnhancer;
+  }
+
   protected _requestHandler: HttpRequestListener;
   public get requestHandler(): HttpRequestListener {
     if (this._requestHandler == null) {
@@ -227,6 +236,16 @@ export class RestServer extends Context implements Server, HttpServerLike {
 
     this.bind(RestBindings.BASE_PATH).toDynamicValue(() => this._basePath);
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
+  }
+
+  protected _setupOASEnhancerIfNeeded() {
+    if (this._OASEnhancer != null) return;
+    this.add(
+      createBindingFromClass(OASEnhancerService, {
+        key: OAS_ENHANCER_SERVICE,
+      }),
+    );
+    this._OASEnhancer = this.getSync(OAS_ENHANCER_SERVICE);
   }
 
   protected _setupRequestHandlerIfNeeded() {
@@ -794,6 +813,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
     if (requestContext) {
       spec = this.updateSpecFromRequest(spec, requestContext);
     }
+
+    // Apply OAS enhancers to the OpenAPI specification
+    this.OASEnhancer.spec = spec;
+    spec = await this.OASEnhancer.applyAllEnhancers();
 
     return spec;
   }
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -497,7 +497,7 @@ export class RestServer extends Context implements Server, HttpServerLike {
     );
 
     specForm = specForm ?? {version: '3.0.0', format: 'json'};
-    const specObj = this.getApiSpec(requestContext);
+    const specObj = await this.getApiSpec(requestContext);
 
     if (specForm.format === 'json') {
       const spec = JSON.stringify(specObj, null, 2);
@@ -777,8 +777,8 @@ export class RestServer extends Context implements Server, HttpServerLike {
    * @param requestContext - Optional context to update the `servers` list
    * in the returned spec
    */
-  getApiSpec(requestContext?: RequestContext): OpenApiSpec {
-    let spec = this.getSync<OpenApiSpec>(RestBindings.API_SPEC);
+  async getApiSpec(requestContext?: RequestContext): Promise<OpenApiSpec> {
+    let spec = await this.get<OpenApiSpec>(RestBindings.API_SPEC);
     const defs = this.httpHandler.getApiDefinitions();
 
     // Apply deep clone to prevent getApiSpec() callers from
--- a/packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
+++ b/packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
@@ -41,8 +41,8 @@ describe('booter-lb3app', () => {
   });
 
   context('generated OpenAPI spec', () => {
-    it('uses different request-body schema for "create" operation', () => {
-      const spec = app.restServer.getApiSpec();
+    it('uses different request-body schema for "create" operation', async () => {
+      const spec = await app.restServer.getApiSpec();
       const createOp: OperationObject = spec.paths['/api/CoffeeShops'].post;
       expect(createOp.requestBody).to.containDeep({
         content: {
@@ -66,8 +66,8 @@ describe('booter-lb3app', () => {
         });
     });
 
-    it('includes the target model as a property of the source model in a relation', () => {
-      const spec = app.restServer.getApiSpec();
+    it('includes the target model as a property of the source model in a relation', async () => {
+      const spec = await app.restServer.getApiSpec();
       const schemas = (spec.components ?? {}).schemas ?? {};
 
       expect(schemas.CoffeeShop)
@@ -118,8 +118,8 @@ describe('booter-lb3app', () => {
       }
     });
 
-    it('includes LoopBack 3 endpoints with `/api` base in OpenApiSpec', () => {
-      const apiSpec = app.restServer.getApiSpec();
+    it('includes LoopBack 3 endpoints with `/api` base in OpenApiSpec', async () => {
+      const apiSpec = await app.restServer.getApiSpec();
       const paths = Object.keys(apiSpec.paths);
       expect(paths).to.containDeep([
         '/api/CoffeeShops/{id}',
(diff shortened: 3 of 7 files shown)


packages/openapi-v3/src/__tests__/unit/enhancers/fixtures/application.ts
```diff
@@ -5,8 +5,8 @@
 
 import {Application, createBindingFromClass} from '@loopback/core';
 import {OASEnhancerService, OAS_ENHANCER_SERVICE} from '../../../..';
+import {SecuritySpecEnhancer} from '../../../../enhancers/extensions/security.spec.extension';
```
raymondfeng: Please merge this import with the one above.

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
```diff
@@ -138,7 +138,7 @@ describe('CoffeeShopApplication', () => {
     let apiSpec: OpenApiSpec;
 
     before(async () => {
-      apiSpec = app.lbApp.restServer.getApiSpec();
+      apiSpec = await app.lbApp.restServer.getApiSpec();
```
raymondfeng: This is a breaking change. Please update the commit message per https://loopback.io/doc/en/lb4/code-contrib-lb4.html#footer-optional. See an example at https://github.com/strongloop/loopback-next/pull/4451/commits/98a3f422568877a3b813160186eec9630208319f and https://github.com/strongloop/loopback-next/blob/master/packages/context/CHANGELOG.md#breaking-changes.

packages/openapi-v3/src/enhancers/extensions/security.spec.extension.ts
```diff
@@ -6,37 +6,36 @@
...
+ * A spec enhancer to add jwt bearer token OpenAPI security entry to
  * `spec.component.securitySchemes`
  */
 @bind(asSpecEnhancer)
 export class SecuritySpecEnhancer implements OASEnhancer {
   name = 'security';
 
   modifySpec(spec: OpenApiSpec): OpenApiSpec {
-    const patchSpec = {components: {securitySchemes: SECURITY_SCHEME_SPEC}};
+    if (spec?.components?.securitySchemes?.jwt) return spec;
```
raymondfeng: I would expect the security spec enhancers to be contributed by authentication strategies instead of being hard-coded in `@loopback/rest` module. This code was added as part of the test fixture intentionally.

dougal83: Subject to the story(#4386) it is a default and I've opted to not override an existing item. @jannyHou What to do?

examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts
dougal83: Please check the commit message is acceptable.

packages/openapi-v3/src/enhancers/extensions/security.spec.extension.ts
jannyHou: @raymondfeng I think @dougal83 is adding the default authorizor button in explorer.

Screen Shot 2020-02-07 at 2 04 03 PM
Screen Shot 2020-02-07 at 2 03 57 PM

jannyHou: @raymondfeng I think we agree to not bind a built-in jwt strategy into LB4 app, so the only thing we can provide the the security spec, as what @dougal83 adds here.

How about introduce an option `useDefaultSecuritySpec`, sets to true by default so that our explorer shows the button. If people provide their own auth strategy & security spec enhancer , they can turn off the option.

packages/rest/src/rest.server.ts
```diff
@@ -717,6 +742,10 @@ export class RestServer extends Context implements Server, HttpServerLike {
       spec = this.updateSpecFromRequest(spec, requestContext);
     }
 
+    // Apply OAS enhancers to the OpenAPI specification
+    this.OASEnhancer.spec = spec;
+    spec = await this.OASEnhancer.applyEnhancerByName('security');
```
jannyHou: I would suggest rename it to be `lb-default-security`

packages/openapi-v3/src/enhancers/extensions/security.spec.extension.ts
```diff
@@ -6,37 +6,36 @@
...
 };
 
 /**
- * A spec enhancer to add bearer token OpenAPI security entry to
+ * A spec enhancer to add jwt bearer token OpenAPI security entry to
  * `spec.component.securitySchemes`
  */
 @bind(asSpecEnhancer)
 export class SecuritySpecEnhancer implements OASEnhancer {
   name = 'security';
```
jannyHou: same as https://github.com/strongloop/loopback-next/pull/4554/files#r376586868, I would rename it to `lb-default-security`

packages/rest/src/rest.server.ts
bajtos: I don't think the REST server should be invoking specific enhancers defined by a name, because that makes it impossible for other extensions to contribute new enhancers.

As I understand the proposed design:
1. The REST API invokes all registered API enhancers via `applyAllEnhancers()`.
2. Extensions like `@loopback/authentication` contribute custom API enhancers to provide additional OpenAPI metadata like the authentication scheme.

The REST component (`@loopback/rest`) should not include any security schemas in the OpenAPI spec out of the box, simply because it does not implement any authentication out of the box.

packages/openapi-v3/src/enhancers/extensions/security.spec.extension.ts
bajtos: > I would expect the security spec enhancers to be contributed by authentication strategies instead of being hard-coded in `@loopback/rest` module.

+1

> How about introduce an option useDefaultSecuritySpec, sets to true by default so that our explorer shows the button. If people provide their own auth strategy & security spec enhancer , they can turn off the option.

As I mentioned in my other comment, it's my understanding that the REST component does not provide any authentication mechanism out of the box, there is no "default security". Therefore the option `useDefaultSecuritySpec` does not make sense to me.

IMO, the security spec should be contributed by `@loopback/authentication`, because that's the extension adding support for bearer tokens. The enhancer should be clever enough to handle the case when there are multiple authentication extensions and/or security specs involved (e.g. Bearer token, HTTP Basic/Digest auth, etc.) and preserve those additional security specs.

packages/rest/src/rest.server.ts
dougal83: Totally agree, I was being a slave to example in story #4380  Could you provide direction on following:

- Add option to disable all enhancers on rest server?  Alternatively option to override by providing array list of enhancers to apply?
- Would I follow similar pattern to [authentication strategies](https://github.com/strongloop/loopback4-example-shopping/tree/master/packages/shopping/src) say `spec-enhancers` folder alongside or would I seek to embed the enhancers within strategies.  If you have the time, perhaps a top level stepwise implementation path would be useful to follow.

bajtos: > Add option to disable all enhancers on rest server? Alternatively option to override by providing array list of enhancers to apply?

I am proposing to leave this part out of the initial implementation and wait until there is user demand for that.

> Would I follow similar pattern to [authentication strategies](https://github.com/strongloop/loopback4-example-shopping/tree/master/packages/shopping/src) say `spec-enhancers` folder alongside or would I seek to embed the enhancers within strategies.  If you have the time, perhaps a top level stepwise implementation path would be useful to follow.

IMO, it's not a concern of this pull request how the spec enhancer classes are registered.

I am expecting the `@loopback/authentication` extension to contribute a binding with `@bind(asSpecEnhancer)` decorator. As I understand the current implementation of `applyAllEnhancers`, it will automatically pick up such class.

See my ModelApiBooter spike in #3617 for an example. I really wish we had better documentation about the ExtensionPoint/Extension pattern :-/

Here the app decides to register a component (it's CrudRestComponent in the spike, it would be `@loopback/authentication` in your case):

https://github.com/strongloop/loopback-next/blob/d8fe4736fd07fc9da34ed345cbb77bdbd761b752/examples/todo/src/application.ts#L29

Here the component implements an Extension:

https://github.com/strongloop/loopback-next/blob/d8fe4736fd07fc9da34ed345cbb77bdbd761b752/packages/rest-crud/src/crud-rest-builder.plugin.ts#L30-L31

And here is the Extension implementation contributed by the Component to the Application context:

https://github.com/strongloop/loopback-next/blob/d8fe4736fd07fc9da34ed345cbb77bdbd761b752/packages/rest-crud/src/crud-rest.component.ts#L9-L11

dougal83:
@bajtos I've cut this PR down down to a singular purpose.  Please let me know if it is back on track.

dougal83:
I'm thinking that the scope should be TRANSIENT, should I drop the scope declaration to return to default?

packages/openapi-v3/src/enhancers/extensions/security.spec.extension.ts
dougal83: > @raymondfeng I think @dougal83 is adding the default authorizor button in explorer.
> 
> Screen Shot 2020-02-07 at 2 04 03 PM
> 
> Screen Shot 2020-02-07 at 2 03 57 PM

@jannyHou #4386 Having read the issue/story properly just now... I had made a big mistake on my understanding... I thought that is was generally adding a default to the rest server! 🤦‍♂ To make amends I guess I'll have to look into explorer and making the change as per the story!

jannyHou:
@dougal83 @bajtos got distracted by the auth migration last week, sorry for replying late.

Let me clarify the goal first:

- For https://github.com/strongloop/loopback-next/issues/4386, the ultimate goal is to enable the "authorize" button in the explorer out-of-box in a LB4 app, and the default scheme(bearer auth scheme) should be the spec snippet I post in story description. Do we still agree on it? If yes, we can discuss how to do it.

- For https://github.com/strongloop/loopback-next/issues/4380, the ultimate goal is to add an OAI enhancer service in the rest server, regardless of what enhancers it picks up by `OAIEnhancer.applyEnhancerByName()` or `OAIEnhancer.applyAllEnhancers()`, and where those enhancers are applied(inside `getApiSpec` or in an app's constructor).

And if we still agree on ^, let's discuss each module's responsibility. I think I somehow misunderstand the expected approaches and my understanding has been "@loopback/rest provides the bearer auth scheme enhancer" and "it should be applied inside the rest server class, like the consolidation enhancer added in https://github.com/strongloop/loopback-next/pull/4365" all the time. If this is not the expected behaviour anymore, first sorry for misleading @dougal83 on the implementation... then let's agree on which module provides the enhancer and where to apply it.

> I am expecting the @loopback/authentication extension to contribute a binding with @bind(asSpecEnhancer) decorator. As I understand the current implementation of applyAllEnhancers, it will automatically pick up such class.

`@loopback/authentication` is good to me. And yes `applyAllEnhancers` would automatically pick up it. Do we want to call `applyAllEnhancers` inside the rest server class? inside `getApiSpec` or `_serveOpenApiSpec`? Or in the app constructor?

> When created the story, I didn't realize applying the enhancer would turn the getApiSpec into an async function. I tend to keep it sync and apply the default security enhancer somewhere else, like _serveOpenApiSpec

If we modify the app template and apply the enhancers in app constructor, then we don't have the concern above. WDYT?

dougal83:
@jannyHou Thank you for the reply.  Don't worry about misleading, it's more likely my misunderstanding. :)

From looking at the stories and attempting PRs, it does make sense to have the rest server apply the extensions by default via `applyAllEnhancers` but you make a good alternative suggestion. 👍 

From comments made I also [attempted to extend the AuthStrategies provider](https://github.com/strongloop/loopback-next/pull/4693) to contribute the OASEhancer.  Love to hear your thoughts on the idea!

raymondfeng:
@dougal83 The PR looks good to me now. Would you please add a test to verify that an enhancer is invoked as part of `getApiSpec()` call?

dougal83:
@bajtos @raymondfeng Please let me know if further changes are required.

packages/rest/src/rest.server.ts
```diff
@@ -229,6 +238,16 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this.bind(RestBindings.HANDLER).toDynamicValue(() => this.httpHandler);
   }
 
+  protected _setupOASEnhancerIfNeeded() {
+    if (this._OASEnhancer != null) return;
```
raymondfeng: Nitpick: we could use `this._OASEnhancer = this.getSync(OAS_ENHANCER_SERVICE, {optional: true});` to check so that the service can be bound to the server.

Issue #4520, opened 2020-01-27
[migration/auth] Migrate https://github.com/strongloop/loopback-example-access-control to LB4


raymondfeng:
## Suggestion

Create an example in LoopBack 4 to illustrate similar access control features as https://github.com/strongloop/loopback-example-access-control. The other option is to add similar features to an existing example such as https://github.com/strongloop/loopback4-example-shopping.

## Acceptance criteria

- [x] The example should have ability to authenticate 4 users
  - Use user/user credential DB tables
- [x] Support 4 different roles - guest, owner, team member, administrator
- [x] Authorize different routes against different users/roles
- [x] Document the migration at https://github.com/strongloop/loopback-next/blob/auth-migration/docs/site/migration/auth/example.md

jannyHou:
PR created in https://github.com/strongloop/loopback-next/pull/4571, in progress

jannyHou:
close after merging https://github.com/strongloop/loopback-next/pull/4706
follow up stories: 
- https://github.com/strongloop/loopback-next/issues/4753
- 2nd phase improvement for casbin persistence(TBD)

PR #4571, opened 2020-02-05
 feat: add access control migration app

jannyHou:
Implements https://github.com/strongloop/loopback-next/issues/4520

This example migrates the lb3 acl example to lb4 https://github.com/strongloop/loopback-example-access-control

### App Description

This app is a simple management system with RBAC for authorization. 
3 models: User, Team, Project
- One project belongs to one user. 
- One user can create a team with other users as members.
- Project has 5 operations, users with different roles can perform different operations.

Here is the rules for the 5 corresponding endpoints: 

Screen Shot 2020-02-05 at 6 01 08 PM

Where:

- `class level` means :  /projects 
- `instance level` means: /project{id} 

### Integrate with Casbin

The current design uses casbin system to implement the ROLE MAPPING.
It collects **subject**, **object**, **operation** from request, then the authorizer calls casbin enforcers to make decision. 

For example: Operation "show-balance" only allows a project's `owner` or `team`  to click. When a request comes in, the authorizer will examine the owner policy(stored in rbac_policy.admin.csv) and team policy(stored in rbac_policy.owner.csv) to make decision: if user X is the `owner` OR `team` member  of project Y, then it can see the balance.

### DESIGN QUESTION:

Storing instance level policies in casbin means we need EXTRA effort to update the policies when creating data. For example, when a user(id=1) creates a new project(id=3):
- the owner policy need to update
  ```
  p, p3_owner, project3, show-balance
  p, p3_owner, project3, donate
  p, p3_owner, project3, withdraw
  g, u1, p3_owner
  ```
- the team policy need to update
  ```
  p, p3_team, project3, show-balance
  p, p3_team, project3, donate
  g, u1, p3_team
  g, u2, p3_team
  ```

**Is this the expected design?** 
cc: I would like to hear more feedback from @strongloop/loopback-maintainers 

### For Reviewers

There are amount of app setup codes(models, repos, authentication setup) which are not related to core logic of authorization and access control. The following files are what reviewers should focus on:

- `controllers/project.controller.ts`: especially the ACL rules defined at the beginning
- `services/casbin.authorizer.ts`: the core logic of authorizer
- `services/casbin.enforcers.ts`: the core logic of casbin enforcers
- `services/assign-project-instance-id.voter.ts`: 
- model and policy files in `fixture/casbin`
- acceptance test in `__tests__/acceptance/project.acceptance.ts`

## Checklist


jannyHou:
If the current design is on the right track, here are the next steps(You will find corresponding comments in the code as well):

- Support instance level authorization context ⭐️ ⭐️ 
- The password is exposed (to simplify authentication :p), should create a credential model like what we have in the shopping example ⭐️ 
- Enhance relations between User and Team, now the team members are stored in an array ⭐️ ⭐️ 
- Persistency of the casbin model/policies ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ 
- Sync with casbin model/policies when data changes ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ 
- Add tests ⭐️ ⭐️ ⭐️ 
- Code clean up ⭐️ 

(⭐️  means difficulty)

emonddr:
@jannyHou , perhaps suggest necessary reading before reviewing this PR.

e.g.

- Learn https://github.com/strongloop/loopback-example-access-control inside and out.
- Read up on casbin and do a casbin tutorial (provide your favourite links)
- Read up on LB4 Authorization and tutorial
- and then review this PR.

examples/access-control-migration/src/services/casbin.enforcers.ts
```diff
@@ -0,0 +1,36 @@
+import * as casbin from 'casbin';
+import path from 'path';
+
+export async function getCasbinEnforcerByName(name: string): Promise<casbin.Enforcer | undefined> {
+  if (name == 'admin') return createAdminEnforcer();
+  if (name == 'owner') return createOwnerEnforcer();
+  if (name == 'team') return createTeamEnforcer();
```
derdeka: As the implementation for `createAdminEnforcer()`, `createOwnerEnforcer()` and `createTeamEnforcer()` is highly the same, maybe you want to pass the configuration file as an parameter to an `createEnforcer()` function?

examples/access-control-migration/src/services/casbin.authorizer.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!allowedRoles) return AuthorizationDecision.ALLOW;
+    if (allowedRoles.length < 1) return AuthorizationDecision.DENY;
+
+    let allow: boolean = false;
+
+    for (const role of allowedRoles) {
+      const enforcer = await this.enforcerFactory(role);
+
+      const allowed_by_role = await enforcer.enforce(
```
agnes512: IIUC, is this where the authorizer calls casbin with enforcers to make decision?

jannyHou: @agnes512 yep

jannyHou:
About instance level authorization context:

Look at the current `AuthorizationContext` interface:
https://github.com/strongloop/loopback-next/blob/3c4207730cd87e9399b18969c71515d44991e547/packages/authorization/src/types.ts#L70-L94

It describes the fields for class level information(e.g. resource: 'project'), but lacks instance level information(e.g. resource: 'project', instanceId: '1')

This PR needs to generate instance level resource name like `project${id}`, so I
- add a voter for instance level endpoints `apis/projects/{id}/*` to re-write the resource name. 
    - please note, this voter violates a voter's responsibility, ideally the interceptor should do something to provide the instance level info. so I call this solution as a workaround
- and to let the voter retrieve the instance id from authorization context, I allowed arbitrary fields in the authorization context as:

https://github.com/strongloop/loopback-next/blob/251abf710f6e98490506571dcb59cb8a1a9b96d9/packages/authorization/src/types.ts#L96-L100

**Conclusion: I think we should implement something to allow authorization context contribute instance level info.**

jannyHou:
2nd Phase implementation:

Casbin persistency and synchronize:

- create a new **project**
  - create a set of `p, project${id}_owner, operation` policies
  - create a set of `p, project${id}_team, operation` policies
- add a new **user** to a **team**
  - find the projects owned by the team owner, then create role inherit rules `g, u${id}, project${id}_team`
- add a new endpoint(operation)
    - for each of its allowed roles, add `p, ${role}, operation` policy

jannyHou:
@emonddr I am going to summarize your suggestions in https://github.com/strongloop/loopback-next/pull/4571#issuecomment-582996309 and upload the meeting slides.

examples/access-control-migration/src/application.ts
```diff
@@ -0,0 +1,103 @@
...
+      },
+    };
+  }
+
+  setUpBindings(): void {
+    this.bind(TokenServiceBindings.TOKEN_SECRET).to(
+      TokenServiceConstants.TOKEN_SECRET_VALUE,
+    );
+    this.bind(TokenServiceBindings.TOKEN_EXPIRES_IN).to(
+      TokenServiceConstants.TOKEN_EXPIRES_IN_VALUE,
```
emonddr: set this to 21600 instead of 600. So when people `try it out` or debug it...the token doesn't expire in 10 minutes.

examples/access-control-migration/src/migrate.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
```
emonddr: why do we need a migrate.ts file for a sample?

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
```diff
@@ -0,0 +1,35 @@
...
+  AuthorizationContext,
+  AuthorizationMetadata,
+  AuthorizationDecision,
+} from '@loopback/authorization';
+import _ from 'lodash';
+
+// Instance level authorizer
+// This is a WORKAROUND to modify the authorization context
+// It is not used for making a decision, so just returns ABSTAIN
+export async function assignProjectInstanceId(
```
emonddr: We'll need a function to compute the resourceId value in authorization context, but in the end, it shouldn't be a `voter` function. Those should return a decision. (but I understand why you put it here in the meantime...it sets up the data that is used by casbinproviderauthorizer.)

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
```diff
@@ -0,0 +1,35 @@
...
+  authorizationCtx: AuthorizationContext,
+  metadata: AuthorizationMetadata,
+) {
+  // hard-coded for a collection of known endpoints
+  // TBD(FEAT): We should support instance level context for object,
+  // like project1, project2, project3
+  // the current AuthrozationCtx is designed for class level object,
+  // like projects
+  const projectId = authorizationCtx.invocationContext.args[0];
+  const resourceName = getResourceName(metadata.resource ?? authorizationCtx.resource, projectId);
```
emonddr: at the moment `authorizationCtx.resource` is not really used  . wondering if we should remove it.

examples/access-control-migration/src/services/casbin.authorizer.ts
```diff
@@ -0,0 +1,72 @@
...
+  value(): Authorizer {
+    return this.authorize.bind(this);
+  }
+
+  async authorize(
+    authorizationCtx: AuthorizationContext,
+    metadata: AuthorizationMetadata,
+  ): Promise<AuthorizationDecision> {
+    const subject = this.getUserName(authorizationCtx.principals[0].id);
+    const object = authorizationCtx.resourceId ?? metadata.resource ?? authorizationCtx.resource;
```
emonddr: `authorizationCtx.resource` not really used at all.

examples/access-control-migration/src/services/casbin.authorizer.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const request: AuthorizationRequest = {
+      subject,
+      object,
+      action: (metadata.scopes && metadata.scopes[0]) || 'execute',
```
emonddr: is `execute` defined in the casbin policy or csv files?

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
```diff
@@ -0,0 +1,35 @@
...
+  // resourceId will override the resource name from metadata
+  authorizationCtx.resourceId = resourceName;
+  return AuthorizationDecision.ABSTAIN;
+}
+
+// TBD: REFACTOR to a casbin util file
+// Generate the resource name according to the naming convention
+// in casbin policy
+function getResourceName(resource: string, id?: number): string {
```
emonddr: if we plan on having `u1` in the casbin files, then we should have a table that maps `u1` to a unique `email` or `userid` instead of a database `index`.  [securityId] and id in principal are usually the same and are database uuid or indexes. I think for casbin, this will cause problems .

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: @emonddr Yeah I also had that concern, but after the meeting I think we agree that it's ok to contain per-process or post-process logic in the voters, see my comment in https://github.com/strongloop/loopback-next/pull/4571#issuecomment-584296913, especially the update at the bottom.

examples/access-control-migration/src/migrate.ts
jannyHou: copied from todo example :p, removed.

examples/access-control-migration/src/services/casbin.enforcers.ts
jannyHou: refactored in commit https://github.com/strongloop/loopback-next/pull/4571/commits/4ed9b73865c1bb8143a88bc022fcc89f506ab519

dougal83:
@emonddr Thanks, I think I need to read up.  Very helpful.
> * Learn https://github.com/strongloop/loopback-example-access-control inside and out.
> * Read up on casbin and do a casbin tutorial (provide your favourite links)
> * Read up on LB4 Authorization and tutorial
> * and then review this PR.

Not sure I like casbin personally.  What were the other options when it was chosen?

jannyHou:
@emonddr @dougal83 
> Learn https://github.com/strongloop/loopback-example-access-control inside and out.

To save reviewer's time, I already summarized  the user scenario or the original app at the very beginning, see the #App-description part in https://github.com/strongloop/loopback-next/pull/4571#issue-371638387

And I created another diagram with precreated data in 
Screen Shot 2020-02-14 at 10 38 40 AM

> Read up on casbin and do a casbin tutorial (provide your favourite links)

The app has very basic usage of Casbin:

- uses a model file to describe the shape of request, policy, role mapping, and the decision rules.
  - syntax doc for [casbin models](https://casbin.org/docs/en/syntax-for-models)
  - syntax doc for [casbin role inheritance](https://casbin.org/docs/en/rbac)
  - what is request definition: a set of info collected from the client request. In our case: subject, object, operation
  - what is policy definition: who has permission for what. In our case the info are same as request: subject, object, operation
  - when make a decision, the casbin enforcer compares the info from request to the policies
- uses policy files to store the policies.
  - the casbin docs I post above also contain the policy examples
  - divide policy files into multiple smaller ones based on roles
  
The following diagram shows how this app defines casbin model and policy, different colors maps to different concepts.

Screen Shot 2020-02-14 at 10 38 23 AM

And one more diagram for the design of authorization:

Screen Shot 2020-02-14 at 10 39 13 AM

.vscode/launch.json
```diff
@@ -13,7 +13,7 @@
       "args": [
         "--config",
         "${workspaceRoot}/packages/build/config/.mocharc.json",
-        "packages/*/dist/__tests__/**/*.js",
+        "examples/*/dist/__tests__/**/*.js",
```
raymondfeng: Please revert this or include both `packages` and `examples`.

examples/access-control-migration/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2018,2019. All Rights Reserved.
+Node module: @loopback/example-todo
```
raymondfeng: Please fix module names.

examples/access-control-migration/README.md
```diff
@@ -0,0 +1,76 @@
+# @loopback/example-todo
```
raymondfeng: Wrong module name.

examples/access-control-migration/src/__tests__/acceptance/project.acceptance.ts
```diff
@@ -0,0 +1,191 @@
...
+   - operate on a "clean" environment each time (a fresh in-memory database)
+   - avoid polluting the test with large quantities of setup logic to keep
+   them clear and easy to read
+   - keep them DRY (who wants to write the same stuff over and over?)
+   ============================================================================
+   */
+
+  async function givenRunningApplication() {
+    process.env.SEED_DATA = '1';
+    app = new AccessControlApplication();
```
raymondfeng: Please use:
```ts
app = new AccessControlApplication({
  rest: givenHttpServerConfig(),
});
```
```

File: examples/access-control-migration/package.json [hunk 14]
```diff
@@ -0,0 +1,80 @@
...
+    "lint": "npm run prettier:check && npm run eslint",
+    "lint:fix": "npm run eslint:fix && npm run prettier:fix",
+    "prettier:cli": "lb-prettier \"**/*.ts\"",
+    "prettier:check": "npm run prettier:cli -- -l",
+    "prettier:fix": "npm run prettier:cli -- --write",
+    "eslint": "lb-eslint --report-unused-disable-directives .",
+    "eslint:fix": "npm run eslint -- --fix",
+    "pretest": "npm run build",
+    "test": "lb-mocha \"dist/__tests__/**/*.js\"",
+    "test:dev": "lb-mocha --allow-console-logs dist/__tests__/**/*.js && npm run posttest",
```
fidgi: it looks like posttest is not defined

jannyHou: @fidgi good catch, removed.

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: in case metadata doesn't have `resource`, I think we should still keep it.

examples/access-control-migration/src/services/casbin.authorizer.ts
jannyHou: `execute` is the default scope. I define `scopes` for my endpoints, but we cannot make assumption that all endpoints will be decorated with that metadata.

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: not sure I understand:
>  [securityId] and id in principal are usually the same and are database uuid or indexes.
That's why I create the user field in casbin with conversion `u${id}`, to ensure it's unique.
So what is the problem here?

emonddr: What I am saying is that right now, the database id for a user is : 1,2,3, etc. But if I store stuff in Mongodb it will be a long uuid or guid, so you would need `u0209b175-b39c-4dd4-a158-4dc81547e77e` in your casbin file. And if ever users are deleted and added to a database again, these guid values will be different, and so you will need to change your casbin files accordingly.

So having users in memory database:

```
   "User": {
      "1": "{\"id\":1,\"username\":\"John\",\"email\":\"john@doe.com\"}",
      "2": "{\"id\":2,\"username\":\"Jane\",\"email\":\"jane@doe.com\"}",
      "3": "{\"id\":3,\"username\":\"Bob\",\"email\":\"bob@projects.com\"}"
    },
```

and casbin file using  `u1`, `u2` etc  works for this example, but perhaps a more real world example would place
unique user names in the casbin file like: `jannyhou88` and `dremond71` , and in real database these users would have ID values like a guid or long int.

So what I am saying is that we should have a function that takes the securityId and queries the database for the unique userid, and this userid is used by cabin policy decisions.

examples/access-control-migration/src/services/casbin.authorizer.ts
emonddr: Perhaps create a constant named DEFAULT_SCOPE-'execute', and use this constant instead.

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: @emonddr Hmm, this is a design choice that I don't quite agree:
> the database id for a user is : 1,2,3, etc. But if I store stuff in Mongodb it will be a long uuid or guid, so you would need u0209b175-b39c-4dd4-a158-4dc81547e77e in your casbin file.

First, my app as a prototype demo for RBAC, wants to avoid advise on those noisy implementation details not related to the core authorization process.
The design principle I aim to emphasize is [what to do](https://github.com/strongloop/loopback-next/blob/363251f2069b4586afc8668d3eb5f315aed6be46/examples/access-control-migration/imgs/authorization-casbin.png): what authorization system does, what 3rd-party lib does. The diagram explaining that concepts is added in the doc PR, NOT "how to do".

So, things like "how principal(user) id maps to the 3rd-party lib's record"(what we are discussing here), "how to define your casbin policy models", "how to optimize the model relations", etc...really depends on each project's business requirements. IMO An example app shouldn't be over engineered to advise on such decisions.

Back to your question, "But if I store stuff in Mongodb it will be a long uuid or guid...", since my app is created with memory db, I don't have this concern and I don't see any value of designing an example in a way that it is able to support all popular databases. Like I use `opensesame` as the password for everyone to avoid dealing with designing a robust security system.

And in a real project, I don't know is it a good idea to have the back-end data coupled with the 3rd-party library. For example, what if I decide to not use casbin but use oauth0 instead? What if I change the casbin model and the subject uses a combination of 2 fields in the user model? This is also a reason why I'd like to make the back-end data agnostic of the 3rd-party lib.

WDYT?

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: Verified again, it has to be there, because according to the interface [AuthorizationContext](https://github.com/strongloop/loopback-next/blob/master/packages/authorization/src/types.ts#L89) and [AuthorizationMetadata](https://github.com/strongloop/loopback-next/blob/master/packages/authorization/src/types.ts#L56), metadata has `resource` as optional while context has it as required, which means the authorization context guarantees the `resource` is provided.

If I remove it, `getResourceName` has to make the 1st parameter optional:

```ts
function getResourceName(resource: string, id?: number): string {
  // instance level name
  if (id) return `${resource}${id}`;
  // class level name
  return `${resource}*`;
}
```
having multiple optional parameters is annoying, and also doesn't reflect the truth, the resource name should be guaranteed to exist when call the function.

examples/access-control-migration/src/services/casbin.authorizer.ts
jannyHou: see https://github.com/strongloop/loopback-next/pull/4571#discussion_r382813314 ^

dhmlau:
I just realized there's a doc PR covering the README and more: https://github.com/strongloop/loopback-next/pull/4706. Please disregard my comments about README. Thanks!

jannyHou:
@dhmlau 
> I just realized there's a doc PR covering the README and more: #4706

Yes, that PR has `example/acl-migration` as the base branch, if this PR lands first I will have a placeholder for the tutorial in this readme file

examples/access-control-migration/src/services/user.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const credentialsFound = await this.userRepository.findCredentials(
+      foundUser.id,
+    );
+    if (!credentialsFound) {
+      throw new HttpErrors.Unauthorized(invalidCredentialsError);
+    }
+    // TBD: Hash password
+    const passwordMatched = credentialsFound.password === credentials.password;
```
bajtos: Uh oh, storing plaintext passwords looks like a big red flag to me. Please ensure that hashing is implemented before this pull request lands.

jannyHou: oops one more refactor, almost forgot it.

examples/access-control-migration/src/services/jwt.service.ts
```diff
@@ -0,0 +1,77 @@
...
+import {promisify} from 'util';
+import {TokenService} from '@loopback/authentication';
+import {UserProfile, securityId} from '@loopback/security';
+import {TokenServiceBindings} from '../keys';
+
+const jwt = require('jsonwebtoken');
+const signAsync = promisify(jwt.sign);
+const verifyAsync = promisify(jwt.verify);
+
+export class JWTService implements TokenService {
```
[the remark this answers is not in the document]
jannyHou: Yeah was talking with @dhmlau about it as well. She points out the migration doc suggests users build the authentication system from scratch, which is not very friendly, especially for those from lb3, with a built-in middleware to use. 
We can extract the jwt strategy, jwt service and user service into a component(or extension). Then people can just mount the component.

examples/access-control-migration/src/observers/sample.observer.ts
```diff
@@ -0,0 +1,92 @@
...
+      {
+        id: 3,
+        username: 'Bob',
+        email: 'bob@projects.com',
+        password: 'opensesame',
+      },
+    ];
+
+    for (const u of users) {
+      await this.userRepo.create(_.pick(u, ['id', 'email', 'username']));
```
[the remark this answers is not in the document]
jannyHou: The data is pre-generated in `db.json` by executing file, then this observer file is set to skip when start the app. I keep it because the original app has data creation script. 
see the explanation in doc https://github.com/strongloop/loopback-next/pull/4706/files#r383975344

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
emonddr: Ok. We can leave it as-is for the first version. :)

examples/access-control-migration/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@loopback/example-access-control-migration",
+  "version": "1.9.5",
+  "description": "Tutorial example on how to migrate the access control example with LoopBack 4.",
+  "main": "index.js",
+  "engines": {
+    "node": ">=8.9"
```
raymondfeng: `>=10`

examples/access-control-migration/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@loopback/example-access-control-migration",
+  "version": "1.9.5",
```
raymondfeng: 1.0.0?

examples/access-control-migration/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "@loopback/openapi-v3": "^2.0.0",
+    "@loopback/repository": "^1.19.1",
+    "@loopback/rest": "^2.0.0",
+    "@loopback/rest-explorer": "^1.4.10",
+    "@loopback/security": "^0.1.13",
+    "@loopback/service-proxy": "^1.3.17",
+    "casbin": "^3.1.0",
+    "jsonwebtoken": "^8.5.1",
+    "loopback-connector-rest": "^3.6.0",
+    "path": "^0.12.7"
```
raymondfeng: Why `path`? We should use the one from node core.

examples/access-control-migration/public/index.html
```diff
@@ -0,0 +1,73 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>@loopback/example-todo</title>
```
raymondfeng: Pls fix the title.

examples/access-control-migration/public/index.html
```diff
@@ -0,0 +1,73 @@
...
+      text-align: center;
+      margin-bottom: 3em;
+      margin-top: 1em;
+    }
+  </style>
+</head>
+
+<body>
+  <div class="info">
+    <h1>@loopback/example-todo</h1>
```
raymondfeng: Same here.

examples/access-control-migration/src/services/user.service.ts
jannyHou: added in the new commit

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: thanks there will be a few follow-up stories after this PR, will definitely include our discussion here. 🤝

examples/access-control-migration/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2018,2019. All Rights Reserved.
```
raymondfeng: Pls fix the header to be `2020`.

examples/access-control-migration/README.md
```diff
@@ -0,0 +1,73 @@
+# @loopback/example-access-control-migration
+
+This example is migrated from
+[loopback-example-access-control](https://github.com/strongloop/loopback-example-access-control),
+and uses the authentication and authorization system in LoopBack 4 to implement
+the access control.
+
+## Overview
+
+This tutorial demonstrates how to implement a RBAC(role based access control)
```
raymondfeng: `RBAC (Role Based Access Control)`?

examples/access-control-migration/README.md
```diff
@@ -0,0 +1,73 @@
...
+system and provides 5 endpoints to test different role's permissions. The
```
raymondfeng: `different roles' permissions`?

examples/access-control-migration/README.md
```diff
@@ -0,0 +1,73 @@
...
+tutorial of building it from a dummy application is documented in
```
raymondfeng: `is documented` -> 'documented`

examples/access-control-migration/README.md
```diff
@@ -0,0 +1,73 @@
...
+![auth-example-migration-tutorial](https://github.com/strongloop/loopback-next/blob/auth-migration/docs/site/migration/auth/example.md)
+
+## Setup
+
+First, you'll need to install a supported version of Node:
+
+- [Node.js](https://nodejs.org/en/) at v10.19.0 or greater
```
raymondfeng: `v10`

examples/access-control-migration/src/application.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // authorization
+    this.bind('casbin.enforcer.factory').to(getCasbinEnforcerByName);
+    this.bind('authorizationProviders.casbin-provider')
+      .toProvider(CasbinAuthorizationProvider)
+      .tag(AuthorizationTags.AUTHORIZER);
+
+    // authentication
+    registerAuthenticationStrategy(this, JWTAuthenticationStrategy);
+  }
```
raymondfeng: Maybe we should move `setupBindings` to  a local component inside this app. This will show the possibility to have ACLs from an extension package in the future.

examples/access-control-migration/src/keys.ts
```diff
@@ -0,0 +1,32 @@
...
+  export const TOKEN_SERVICE = BindingKey.create<TokenService>(
+    'services.authentication.jwt.tokenservice',
+  );
+}
+
+export namespace UserServiceBindings {
+  export const USER_SERVICE = BindingKey.create<UserService<User, Credentials>>(
+    'services.user.service',
+  );
+}
```
raymondfeng: A good candidate to be moved into the local component too.

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
```diff
@@ -0,0 +1,47 @@
...
+  authorizationCtx: AuthorizationContext,
+  metadata: AuthorizationMetadata,
+) {
+  const projectId = authorizationCtx.invocationContext.args[0];
+  const resourceName = getResourceName(
+    metadata.resource ?? authorizationCtx.resource,
+    projectId,
+  );
+  // resourceId will override the resource name from metadata
+  authorizationCtx.resourceId = resourceName;
```
raymondfeng: Nitpick: would it be better to bind to the invocationContext instead of a property?

examples/access-control-migration/src/application.ts
jannyHou: @raymondfeng yeah moving the authentication setup to a component/extension is tracked in discussion https://github.com/strongloop/loopback-next/pull/4571#discussion_r383710046, I will create a follow-up story to extract the component.

jannyHou:
FYI - Looking at the coverage complain, all of them are for services. Since I will extract the authentication related services into a component, will add unit tests for those services there.
https://coveralls.io/builds/28956490

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
jannyHou: @raymondfeng Hmm not sure what's the difference between "to the invocationContext" and "to a property", the property `resourceId` is a key(property) of `invocationContext` already..

raymondfeng: It’s a property on authorization context. I was referring to bind to invocation context.

examples/access-control-migration/src/services/assign-project-instance-id.voter.ts
```diff
@@ -24,12 +24,12 @@ export async function assignProjectInstanceId(
...
 ) {
   const projectId = authorizationCtx.invocationContext.args[0];
-  const resourceName = getResourceName(
+  const resourceId = getResourceName(
     metadata.resource ?? authorizationCtx.resource,
     projectId,
   );
   // resourceId will override the resource name from metadata
-  authorizationCtx.resourceId = resourceName;
+  authorizationCtx.invocationContext.bind('resourceId').to(resourceId);
```
raymondfeng: Nitpick: do you want to define a constant for the key?

examples/access-control-migration/src/services/jwt.service.ts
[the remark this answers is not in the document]
jannyHou: cc @bajtos @dhmlau I created the follow-up refactor story in https://github.com/strongloop/loopback-next/issues/4753

Issue #4576, opened 2020-02-06
fix code snippet in registering authorization component with options documentation 


alyhegazy:
## Steps to reproduce
The code snippet in the documentation for how to register authorization component with options doesn't add the options.
https://loopback.io/doc/en/lb4/Loopback-component-authorization.html#registering-the-authorization-component

## Current Behavior
Options are not added to the authorization component.

```
const options: AuthorizationOptions = {
  precedence: AuthorizationDecision.DENY,
  defaultDecision: AuthorizationDecision.DENY,
};

const binding = app.component(AuthorizationComponent);
app.configure(binding.key).to(options);
```

`app.component()` has a void return type so `binding` will be `undefined`.

## Expected Behavior
Options should be added to the authorization component.

The following worked for me:
```
import {
  AuthorizationComponent,
  AuthorizationOptions,
  AuthorizationBindings,
} from '@loopback/authorization';

const options: AuthorizationOptions = {
      precedence: AuthorizationDecision.ALLOW,
      defaultDecision: AuthorizationDecision.DENY,
    };

app.component(AuthorizationComponent);
app.configure(AuthorizationBindings.COMPONENT).to(options);
```

## Additional information
Node: v10.14.0
@loopback/cli version: 1.29.0
@loopback/* dependencies:
  - @loopback/authorization: ^0.4.8

## Related Issues

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

jannyHou:
@alyhegazy Good catch.
I verified with an app in PR https://github.com/strongloop/loopback-next/pull/4571

```ts
const auth_options: AuthorizationOptions = {
      defaultDecision: AuthorizationDecision.ALLOW,
      precedence: AuthorizationDecision.ALLOW,
    }
    
this.configure(AuthorizationBindings.COMPONENT).to(auth_options);
```
works.
The debug log:
```
loopback:authorization:interceptor Authorization options { defaultDecision: 'Allow', precedence: 'Allow' } +272ms
```
And according to the code in
https://github.com/strongloop/loopback-next/blob/6eea5e428b145cafb84a998bd53979da8c8fba07/packages/authorization/src/authorize-interceptor.ts#L38

Your solution is the correct fix. Submitting a PR to update the doc

PR #4580, opened 2020-02-06
docs: update auth options

jannyHou:
Fixes: https://github.com/strongloop/loopback-next/issues/4576

Copy from ^

I verified with an app in PR https://github.com/strongloop/loopback-next/pull/4571

```ts
const auth_options: AuthorizationOptions = {
      defaultDecision: AuthorizationDecision.ALLOW,
      precedence: AuthorizationDecision.ALLOW,
    }
    
this.configure(AuthorizationBindings.COMPONENT).to(auth_options);
```
works.
The debug log:
```
loopback:authorization:interceptor Authorization options { defaultDecision: 'Allow', precedence: 'Allow' } +272ms
```
And according to the code in
https://github.com/strongloop/loopback-next/blob/6eea5e428b145cafb84a998bd53979da8c8fba07/packages/authorization/src/authorize-interceptor.ts#L38

Your solution is the correct fix. Submitting a PR to update the doc

## Checklist

### DIFF
--- a/docs/site/Loopback-component-authorization.md
+++ b/docs/site/Loopback-component-authorization.md
@@ -99,8 +99,8 @@ class.
     defaultDecision: AuthorizationDecision.DENY,
   };
 
-  const binding = app.component(AuthorizationComponent);
-  app.configure(binding.key).to(options);
+  app.configure(AuthorizationBindings.COMPONENT).to(options);
+  app.component(AuthorizationComponent);
   ```
 
 - The authorization `options` are provided specifically for enforcing the


docs/site/Loopback-component-authorization.md
```diff
@@ -99,8 +99,8 @@ class.
     defaultDecision: AuthorizationDecision.DENY,
   };
 
-  const binding = app.component(AuthorizationComponent);
-  app.configure(binding.key).to(options);
+  app.component(AuthorizationComponent);
+  app.configure(AuthorizationBindings.COMPONENT).to(options);
```
derdeka: Are you sure that `configure` after `component` works? On other places we have to do `configure` before `component` to make it working?

jannyHou: @derdeka I wasn't sure, so I verified with app in https://github.com/strongloop/loopback-next/pull/4571, you can read the PR description to see my testing result.

`this.configure()` is called in the app constructor, and whenever the authorization interceptor is invoked, the options will be injected. So I believe after app runs, making API calls will inject the configured options.

derdeka: I would like to suggest to switch the order to have it consistent with other docs.

jannyHou: @derdeka 👌 switched the order to have the configure before mounting component.

Issue #3737, opened 2019-09-13
Add CrudRestApiBuilder to `@loopback/rest-crud`


bajtos:
_This is a follow-up for the spike https://github.com/strongloop/loopback-next/issues/2738, see  #3617._

Implement a model-api builder for CRUD REST API pattern - add `CrudRestApiBuilder` to `@loopback/rest-crud`. Modify `README`, rework "Basic use" to show how to use the package together with `ModelApiBooter` to go from a model to REST API. Move the current content of "basic use" into a new section, e.g. "Advanced use".

See the PoC implementation in the spike for inspiration.

**Prerequisites:**
- Implement TestSandbox.writeTextFile #3731
- Improve `defineCrudRestController` to create a named controller class #3732
- Add `defineCrudRepositoryClass` - a helper to create a named repository class #3733
- Model API booter & builder #3736

## Acceptance criteria

- [ ] Implementation
- [ ] API docs
- [ ] Automated tests
- [ ] Updated README

PR #4589, opened 2020-02-07
feat(rest-crud): add CrudRestApiBuilder

nabdelgadir:
Closes https://github.com/strongloop/loopback-next/issues/3737

`CrudRestApiBuilder` is a model API builder that builds the default repository and controller class for a given Entity class.

## Checklist

### DIFF
--- a/packages/rest-crud/README.md
+++ b/packages/rest-crud/README.md
@@ -15,11 +15,60 @@ npm install --save @loopback/rest-crud
 
 ## Basic use
 
-`@loopback/rest-crud` exposes two helper methods (`defineCrudRestController` and
-`defineCrudRepositoryClass`) for creating controllers and respositories using
-code.
+`@loopback/rest-crud` can be used along with the built-in `ModelApiBooter` to
+easily create a repository class and a controller class for your model. The
+following use is a simple approach for this creation, however, you can look at
+the "Advanced use" section instead for a more flexible approach.
 
 For the examples in the following sections, we are assuming a model named
+`Product` and a datasource named `db` have already been created.
+
+In your `src/application.ts` file:
+
+```ts
+// add the following import
+import {CrudRestComponent} from '@loopback/rest-crud';
+
+export class TryApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options: ApplicationConfig = {}) {
+    // other code
+
+    // add the following line
+    this.component(CrudRestComponent);
+  }
+}
+```
+
+Create a new file for the configuration, e.g.
+`src/model-endpoints/product.rest-config.ts` that defines the `model`,
+`pattern`, `dataSource`, and `basePath` properties:
+
+```ts
+import {ModelCrudRestApiConfig} from '@loopback/rest-crud';
+import {Product} from '../models';
+
+module.exports = <ModelCrudRestApiConfig>{
+  model: Product,
+  pattern: 'CrudRest', // make sure to use this pattern
+  dataSource: 'db',
+  basePath: '/products',
+};
+```
+
+Now your `Product` model will have a default repository and default controller
+class defined without the need for a repository or controller class file.
+
+## Advanced use
+
+If you would like more flexibility, e.g. if you would only like to define a
+default `CrudRest` controller or repository, you can use the two helper methods
+(`defineCrudRestController` and `defineCrudRepositoryClass`) exposed from
+`@loopback/rest-crud`. These functions will help you create controllers and
+respositories using code.
+
+For the examples in the following sections, we are also assuming a model named
 `Product`, and a datasource named `db` have already been created.
 
 ### Creating a CRUD Controller
@@ -37,7 +86,7 @@ endpoints of an existing model with a respository.
    >(Product, {basePath: '/products'});
    ```
 
-2. Set up dependency injection for the ProductController.
+2. Set up dependency injection for the `ProductController`.
 
    ```ts
    inject('repositories.ProductRepository')(ProductController, undefined, 0);
@@ -73,10 +122,10 @@ export class TryApplication extends BootMixin(
   ServiceMixin(RepositoryMixin(RestApplication)),
 ) {
   constructor(options: ApplicationConfig = {}) {
-    ...
+    // ...
   }
 
-  async boot():Promise<void> {
+  async boot(): Promise<void> {
     await super.boot();
 
     const ProductRepository = defineCrudRepositoryClass(Product);
@@ -85,9 +134,9 @@ export class TryApplication extends BootMixin(
     inject('datasources.db')(ProductRepository, undefined, 0);
 
     const ProductController = defineCrudRestController<
-    Product,
-    typeof Product.prototype.id,
-    'id'
+      Product,
+      typeof Product.prototype.id,
+      'id'
     >(Product, {basePath: '/products'});
 
     inject(repoBinding.key)(ProductController, undefined, 0);
--- a/packages/rest-crud/src/crud-rest.api-builder.ts
+++ b/packages/rest-crud/src/crud-rest.api-builder.ts
@@ -1,0 +1,140 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/rest-crud
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  bind,
+  BindingSelector,
+  Constructor,
+  ControllerClass,
+  inject,
+} from '@loopback/core';
+import {
+  asModelApiBuilder,
+  ModelApiBuilder,
+  ModelApiConfig,
+} from '@loopback/model-api-builder';
+import {
+  ApplicationWithRepositories,
+  Class,
+  Entity,
+  EntityCrudRepository,
+} from '@loopback/repository';
+import {Model} from '@loopback/rest';
+import debugFactory from 'debug';
+import {defineCrudRepositoryClass, defineCrudRestController} from '.';
+
+const debug = debugFactory('loopback:boot:crud-rest');
+
+export interface ModelCrudRestApiConfig extends ModelApiConfig {
+  // E.g. '/products'
+  basePath: string;
+}
+
+@bind(asModelApiBuilder)
+export class CrudRestApiBuilder implements ModelApiBuilder {
+  readonly pattern: string = 'CrudRest';
+
+  build(
+    application: ApplicationWithRepositories,
+    modelClass: typeof Model & {prototype: Model},
+    cfg: ModelApiConfig,
+  ): Promise<void> {
+    const modelName = modelClass.name;
+    const config = cfg as ModelCrudRestApiConfig;
+    if (!config.basePath) {
+      throw new Error(
+        `Missing required field "basePath" in configuration for model ${modelName}.`,
+      );
+    }
+
+    if (!(modelClass.prototype instanceof Entity)) {
+      throw new Error(
+        `CrudRestController requires a model that extends 'Entity'. (Model name ${modelName} does not extend 'Entity')`,
+      );
+    }
+    const entityClass = modelClass as typeof Entity & {prototype: Entity};
+
+    let repoBindingName = `repositories.${entityClass.name}Repository`;
+
+    if (application.isBound(repoBindingName)) {
+      debug('Using the existing Repository binding %j', repoBindingName);
+    } else {
+      // repository class does not exist
+      const repositoryClass = setupCrudRepository(entityClass, config);
+      application.repository(repositoryClass);
+      repoBindingName = repositoryClass.name;
+      debug('Registered repository class', repoBindingName);
+    }
+
+    const controllerClass = setupCrudRestController(entityClass, config);
+    application.controller(controllerClass);
+    debug('Registered controller class', controllerClass.name);
+
+    return Promise.resolve();
+  }
+}
+
+/**
+ * Set up a CRUD Repository class for the given Entity class.
+ *
+ * @param entityClass - the Entity class the repository is built for
+ * @param config - configuration of the Entity class
+ */
+function setupCrudRepository(
+  entityClass: typeof Entity & {prototype: Entity},
+  config: ModelCrudRestApiConfig,
+): Class<EntityCrudRepository<Entity, unknown>> {
+  const repositoryClass = defineCrudRepositoryClass(entityClass);
+
+  injectFirstConstructorArg(
+    repositoryClass,
+    `datasources.${config.dataSource}`,
+  );
+
+  return repositoryClass;
+}
+
+/**
+ * Set up a CRUD Controller class for the given Entity class.
+ *
+ * @param entityClass - the Entity class the controller is built for
+ * @param config - configuration of the Entity class
+ */
+function setupCrudRestController(
+  entityClass: typeof Entity & {prototype: Entity},
+  config: ModelCrudRestApiConfig,
+): ControllerClass {
+  const controllerClass = defineCrudRestController(
+    entityClass,
+    // important - forward the entire config object to allow controller
+    // factories to accept additional (custom) config options
+    config,
+  );
+
+  injectFirstConstructorArg(
+    controllerClass,
+    `repositories.${entityClass.name}Repository`,
+  );
+
+  return controllerClass;
+}
+
+/**
+ * Inject given key into a given class constructor
+ *
+ * @param ctor - constructor for a class (e.g. a controller class)
+ * @param key - binding to use in order to resolve the value of the decorated
+ * constructor parameter or property
+ */
+function injectFirstConstructorArg<T>(
+  ctor: Constructor<T>,
+  key: BindingSelector,
+) {
+  inject(key)(
+    ctor,
+    undefined, // constructor member
+    0, // the first argument
+  );
+}
--- a/packages/boot/package.json
+++ b/packages/boot/package.json
@@ -37,6 +37,7 @@
     "@loopback/eslint-config": "^5.0.3",
     "@loopback/openapi-v3": "^2.0.0",
     "@loopback/rest": "^2.0.0",
+    "@loopback/rest-crud": "^0.6.6",
     "@loopback/testlab": "^1.10.3",
     "@types/node": "^10.17.16"
   },
--- a/packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
@@ -1,0 +1,172 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/boot
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {ApplicationConfig} from '@loopback/core';
+import {juggler, RepositoryMixin} from '@loopback/repository';
+import {RestApplication} from '@loopback/rest';
+import {CrudRestComponent} from '@loopback/rest-crud';
+import {expect, givenHttpServerConfig, TestSandbox} from '@loopback/testlab';
+import {resolve} from 'path';
+import {BootMixin, ModelApiBooter} from '../..';
+import {ProductRepository} from '../fixtures/product.repository';
+
+describe('CRUD rest builder acceptance tests', () => {
+  let app: BooterApp;
+  const SANDBOX_PATH = resolve(__dirname, '../../.sandbox');
+  const sandbox = new TestSandbox(SANDBOX_PATH);
+
+  beforeEach('reset sandbox', () => sandbox.reset());
+  beforeEach(givenAppWithDataSource);
+
+  afterEach(stopApp);
+
+  it('binds the controller and repository to the application', async () => {
+    await sandbox.copyFile(
+      resolve(__dirname, '../fixtures/product.model.js'),
+      'models/product.model.js',
+    );
+
+    // when creating the config file in a real app, make sure to use
+    // module.exports = <ModelCrudRestApiConfig>{...}
+    // it's not used here because this is a .js file
+    await sandbox.writeTextFile(
+      'model-endpoints/product.rest-config.js',
+      `
+const {Product} = require('../models/product.model');
+module.exports = {
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  basePath: '/products',
+};
+      `,
+    );
+
+    // Boot & start the application
+    await app.boot();
+    await app.start();
+
+    expect(app.getBinding('repositories.ProductRepository').key).to.eql(
+      'repositories.ProductRepository',
+    );
+
+    expect(app.getBinding('controllers.ProductController').key).to.eql(
+      'controllers.ProductController',
+    );
+  });
+
+  it('uses bound repository class if it exists', async () => {
+    await sandbox.copyFile(
+      resolve(__dirname, '../fixtures/product.model.js'),
+      'models/product.model.js',
+    );
+
+    await sandbox.writeTextFile(
+      'model-endpoints/product.rest-config.js',
+      `
+const {Product} = require('../models/product.model');
+module.exports = {
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  basePath: '/products',
+};
+      `,
+    );
+
+    app.repository(ProductRepository);
+
+    const bindingName = 'repositories.ProductRepository';
+
+    const binding = app.getBinding(bindingName);
+    expect(binding.valueConstructor).to.eql(ProductRepository);
+
+    // Boot & start the application
+    await app.boot();
+    await app.start();
+
+    // Make sure it is still equal to the defined ProductRepository after
+    // booting
+    expect(app.getBinding(bindingName).valueConstructor).to.eql(
+      ProductRepository,
+    );
+
+    expect(app.getBinding('controllers.ProductController').key).to.eql(
+      'controllers.ProductController',
+    );
+  });
+
+  it('throws if there is no base path in the config', async () => {
+    await sandbox.copyFile(
+      resolve(__dirname, '../fixtures/product.model.js'),
+      'models/product.model.js',
+    );
+
+    await sandbox.writeTextFile(
+      'model-endpoints/product.rest-config.js',
+      `
+const {Product} = require('../models/product.model');
+module.exports = {
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  // basePath not specified
+};
+      `,
+    );
+
+    // Boot the application
+    await expect(app.boot()).to.be.rejectedWith(
+      /Missing required field "basePath" in configuration for model Product./,
+    );
+  });
+
+  it('throws if a Model is used instead of an Entity', async () => {
+    await sandbox.copyFile(
+      resolve(__dirname, '../fixtures/no-entity.model.js'),
+      'models/no-entity.model.js',
+    );
+
+    await sandbox.writeTextFile(
+      'model-endpoints/no-entity.rest-config.js',
+      `
+const {NoEntity} = require('../models/no-entity.model');
+module.exports = {
+  // this model extends Model, not Entity
+  model: NoEntity,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  basePath: '/no-entities',
+};
+      `,
+    );
+
+    // Boot the application
+    await expect(app.boot()).to.be.rejectedWith(
+      /CrudRestController requires a model that extends 'Entity'./,
+    );
+  });
+
+  class BooterApp extends BootMixin(RepositoryMixin(RestApplication)) {
+    constructor(options?: ApplicationConfig) {
+      super(options);
+      this.projectRoot = sandbox.path;
+      this.booters(ModelApiBooter);
+      this.component(CrudRestComponent);
+    }
+  }
+
+  async function givenAppWithDataSource() {
+    app = new BooterApp({
+      rest: givenHttpServerConfig(),
+    });
+    app.dataSource(new juggler.DataSource({connector: 'memory'}), 'db');
+  }
+
+  async function stopApp() {
+    if (app.state !== 'started') return;
+    await app.stop();
+  }
+});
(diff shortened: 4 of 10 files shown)


packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  async function stopApp() {
+    try {
+      await app.stop();
+    } catch (err) {
+      // application is booting
+    }
+  }
+});
```
agnes512: The tests look legit to me 👍 I don't know if we support configurations yet. It'd be good if we can have a test for it.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,168 @@
...
+      rest: givenHttpServerConfig(),
+    });
+    app.dataSource(new juggler.DataSource({connector: 'memory'}), 'db');
+  }
+
+  async function stopApp() {
+    try {
+      await app.stop();
+    } catch (err) {
+      // application is booting
```
raymondfeng: Why do we want to hide the error? We should let `mocha` to capture and report it.

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    try {
+      application.getBinding('repositories.' + repoClassName);
+      debug(
+        'Using repository class',
+        repoClassName,
+        ', as it is already bound to application',
+      );
+    } catch {
+      // repository class does not exist
```
raymondfeng: We should use `application.isBound('repositories.' + repoClassName)` to test instead of `try-catch`.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
nabdelgadir: Some of the tests are meant to throw an error during `app.boot()` so the application never starts and so if we let `mocha` capture it, it'll always throw errors every time the tests are run.

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Set up a CRUD Repository class for the given Entity class.
+ *
+ * @param entityClass - the Entity class the repository is built for
+ * @param modelConfig - configuration of the Entity class
+ */
+function setupCrudRepository(
+  entityClass: typeof Entity & {prototype: Entity},
+  modelConfig: ModelCrudRestApiConfig,
+): Class<EntityCrudRepository<Entity, unknown>> {
+  const repositoryClass = defineCrudRepositoryClass(entityClass);
```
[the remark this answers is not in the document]
nabdelgadir: I don't think it's configurable right now based on the definition of [`defineCrudRepositoryClass`](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/rest-crud/src/repository-builder.ts#L25), but what kind of config options were you thinking?

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Set up a CRUD Repository class for the given Entity class.
+ *
+ * @param entityClass - the Entity class the repository is built for
+ * @param modelConfig - configuration of the Entity class
+ */
+function setupCrudRepository(
+  entityClass: typeof Entity & {prototype: Entity},
+  modelConfig: ModelCrudRestApiConfig,
```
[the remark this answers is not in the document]
nabdelgadir: `modelConfig` is supposed to be configuration for the model class. The type is `ModelCrudRestApiConfig` which extends [`ModelApiConfig`](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/model-api-builder/src/model-api-config.ts#L12). For the repository, we're only using this config to get the model's datasource and for the controller, it's for the basepath, but I'll rename to `config` so it's more generic. 👍

packages/rest-crud/README.md
```diff
@@ -15,11 +15,55 @@ npm install --save @loopback/rest-crud
...
+  constructor(options: ApplicationConfig = {}) {
+    // other code
+
+    // add the following line
+    this.component(CrudRestComponent);
+  }
+}
+```
+
+Create a new file for the configuration, e.g.
```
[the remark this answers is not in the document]
nabdelgadir: > What kind of configurations we are talking about? Can we explain it a bit?

I'll make it more clear, but it's only the ones mentioned below:
```ts
model: Product,
pattern: 'CrudRest',
dataSource: 'db',
basePath: '/products',
```

> For example, what is needed in the file for it to configure repository and controller? 

We don't have options for specifically configuring the repository or controller. ATM, the `basePath` is for the controller and the `dataSource` is for the repository.

> Besides, what patterns are available? Is `CrudRest` the only one that being supported?

Users can make their own with the `@loopback/model-api-builder` package, but yes LoopBack only has the `CrudRest` one so far.

packages/rest-crud/src/crud-rest-builder.plugin.ts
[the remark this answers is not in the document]
agnes512: I think for repositories, the only one I can come up is the inclusion resolver. But I am not sure if we allow it to be configured.

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,123 @@
...
+): Class<EntityCrudRepository<Entity, unknown>> {
+  const repositoryClass = defineCrudRepositoryClass(entityClass);
+
+  inject(`datasources.${modelConfig.dataSource}`)(
+    repositoryClass,
+    undefined,
+    0,
```
jannyHou: @nabdelgadir might be a silly question: I am looking at the [`inject`](https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L127) function, how do we know the index is 0 and why the `member` name is undefined?

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,168 @@
...
+    await sandbox.copyFile(
+      resolve(__dirname, '../fixtures/product.model.js'),
+      'models/product.model.js',
+    );
+
+    await sandbox.writeTextFile(
+      'model-endpoints/product.rest-config.js',
+      `
+const {Product} = require('../models/product.model');
+module.exports = {
```
jannyHou: we don't need to specify the type like documented [in the doc](https://github.com/strongloop/loopback-next/blob/bcc3000d55f457c15463cf43ee18091dcc0dd09f/packages/rest-crud/README.md#basic-use)?
```ts
module.exports = <ModelCrudRestApiConfig>{
  model: Product,
  pattern: 'CrudRest',
  dataSource: 'db',
  basePath: '/products',
};
```

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,168 @@
...
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  basePath: '/products',
+};
+      `,
+    );
+
+    const ProductRepository = defineCrudRepositoryClass(Product);
```
jannyHou: If the repository will be created and bound when the config exists, what's the point of creating it here again?
Or do we want to test that if a custom product repository is created and bound to the app, the default one won't be created? Then I think the test needs to make the ProductRepository something different than the default one.

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,112 @@
...
+    const config = cfg as ModelCrudRestApiConfig;
+    if (!config.basePath) {
+      throw new Error(
+        `Missing required field "basePath" in configuration for model ${modelName}.`,
+      );
+    }
+
+    if (!(modelClass.prototype instanceof Entity)) {
+      throw new Error(
+        `CrudRestController requires an Entity, Models are not supported. (Model name: ${modelName})`,
```
emonddr: > `CrudRestController requires an Entity, Models are not supported. (Model name: ${modelName})`,

What are we trying to say here? That the model needs to extend Entity?

Perhaps

`CrudRestController requires a model that extends 'Entity'. (Model name: ${modelName} does not extend 'Entity')`

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,112 @@
...
+      );
+    }
+    const entityClass = modelClass as typeof Entity & {prototype: Entity};
```
emonddr: Why is `& {prototype: Entity}` needed here? Isn't first part sufficient?

@raymondfeng @nabdelgadir   ^

packages/rest-crud/src/crud-rest-builder.plugin.ts
emonddr: Why are we calling the file 'plugin'?

> packages/rest-crud/src/crud-rest-builder.plugin.ts

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,112 @@
...
+    entityClass,
+    // important - forward the entire config object to allow controller
+    // factories to accept additional (custom) config options
+    modelConfig,
+  );
+
+  inject(`repositories.${entityClass.name}Repository`)(
+    controllerClass,
+    undefined,
+    0,
```
emonddr: @nabdelgadir  Can you explain how the `0` is used? It is not clear in the function that receives this value. thx.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
nabdelgadir: Since we're using a js file (model-endpoints/product.rest-config.js) instead of a ts file, the type can't be specified like it is in the docs (or else it'll have compilation errors).

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
nabdelgadir: > Or do we want to test that if a custom product repository is created and bound to the app, the default one won't be created? Then I think the test needs to make the ProductRepository something different than the default one.

This was the intention. I also tried creating a separate repository class and using that one but thinking about it now, I don't think there's a way to differentiate between them since their bindings are the same, so I might remove this test.

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: > Why are we calling the file 'plugin'?
> 
> > packages/rest-crud/src/crud-rest-builder.plugin.ts

Following the terminology in the [spike](https://github.com/strongloop/loopback-next/pull/3617/files#diff-1ac72022ee3c92074711610d69356271R49-R59)

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: I remember seeing in [legacy-juggler-bridge](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/repository/src/repositories/legacy-juggler-bridge.ts#L116), it's mentioned that "entityClass should have type "typeof T", but that's not supported by TSC" (where T extends Entity) and also it matches the type that's passed into [`defineCrudRepositoryClass`](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/rest-crud/src/repository-builder.ts#L30)

packages/rest-crud/src/crud-rest-builder.plugin.ts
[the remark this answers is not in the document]
nabdelgadir: I don't think relations are part of the default crud repo that's generated with this plugin

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: From my understanding, the only reason we're using a number (that happens to be 0 in this case) is so that it goes into [this `if` statement](https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L134), but I don't think the number actually matters. And we're only passing in `member` since it's required and so we're able to pass in the number as the third argument, so leaving it as undefined is the easiest way to go about it (but I don't actually know the role of `member` in the injection, maybe @bajtos could comment on this?).

packages/rest-crud/src/crud-rest-builder.plugin.ts
emonddr: @nabdelgadir ,

Hmm. Let's not use the word plugin. We don't mention it in the overall LB4 docs.

As an example,  AuthenticationStrategyProvider is a class that defines an extension point where developers can create extensions (custom authentication strategies) and register them against the extension point... but we don't call any of the files  `.plugin.ts` . 

Same for the extensions for greeter extension point : https://github.com/strongloop/loopback-next/blob/master/examples/greeter-extension/src/greeters/greeter-en.ts#L6  , we don't use `.plugin.ts` .

https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-strategy.provider.ts#L28

packages/rest-crud/src/crud-rest-builder.plugin.ts
emonddr: > I remember seeing in [legacy-juggler-bridge](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/repository/src/repositories/legacy-juggler-bridge.ts#L116), it's mentioned that "entityClass should have type "typeof T", but that's not supported by TSC" (where T extends Entity) and also it matches the type that's passed into [`defineCrudRepositoryClass`](https://github.com/strongloop/loopback-next/blob/crud-rest/packages/rest-crud/src/repository-builder.ts#L30)

Good link. It answers my question.

packages/rest-crud/src/crud-rest-builder.plugin.ts
emonddr: @bajtos  ^ (I have the same question)

bajtos: When you set the member to `undefined`, it means the constructor function. (The constructor does not have any name.

The index `0` is identifying the first argument of the target function (the class constructor in this case).

In this particular invocation of `inject`, we are decorating the first argument of the repository constructor to inject the datasource instance. It's an equivalent of the following TypeScript code:

```ts
class MyRepository /*extends ...*/ {
  constructor(
    @inject(`datasources.${modelConfig.dataSource}`)
    public dataSource: juggler.DataSource
  ) {}
}
```

Because decorators are converted to JavaScript calls of TypeScript API at compile time, we cannot use decorator syntax when building the repository class dynamically at runtime, we have to invoke the decorator function manually.

packages/rest-crud/src/crud-rest-builder.plugin.ts
bajtos: > Let's not use the word plugin. We don't mention it in the overall LB4 docs.

Considering that we are using ExtensionPoint/Extension design pattern, maybe we can use "extension" instead of "plugin"?

packages/rest-crud/src/crud-rest-builder.plugin.ts
emonddr: Can we create a constant and use that instead of 0? Perhaps also one for the undefined value? So both constants can provide some clarity to the method call?

```
inject(`datasources.${modelConfig.dataSource}`)(
    repositoryClass,
    UNDEFINED_CLASS_CONSTRUCTOR_FUNCTION,
    INDEX_OF_INJECTED_DATASOURCE_PARAMETER_INSIDE_REPOSITORY_CONSTRUCTOR,
```

I know the names are silly, but you get the point.

packages/rest-crud/README.md
```diff
@@ -15,11 +15,56 @@ npm install --save @loopback/rest-crud
...
+
+module.exports = <ModelCrudRestApiConfig>{
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+  basePath: '/products',
+};
+```
+
+Now your `Product` model will have a default repository and default controller
```
emonddr: So there are 2 ways that a developer can create the `default` controller and `default `repository for Product?
  1. Performing Basic steps  and **not** performing Advanced steps
  1. Performing Advanced steps and **not** performing Basic steps

So they are mutually exclusive then? If so let's specify that Basic and Advanced are mutually exclusive. Another question: why would someone want to do Advanced steps if it is shorter to perform Basic steps?

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    await sandbox.writeTextFile(
+      'model-endpoints/product.rest-config.js',
+      `
+const {Product} = require('../models/product.model');
+module.exports = {
+  model: Product,
+  pattern: 'CrudRest',
+  dataSource: 'db',
+};
```
bajtos: Perhaps add a code comment to make it more clear that `basePath` was omitted here?

```js
 const {Product} = require('../models/product.model');
 module.exports = {
   model: Product,
   pattern: 'CrudRest',
   dataSource: 'db',
   // basePath not specified
 };
```

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
bajtos: Does it mean that `app.stop()` throws an error if the app was not started? I find that surprising, because the implementation of `stop` does seem to short-circuit to a no-op when not running. 

https://github.com/strongloop/loopback-next/blob/4192217263f5f4fc1dfe5c0a0a4deac6bcf9aad1/packages/core/src/application.ts#L285-L294

I guess fixing this problem is out of scope of this pull request? 

Have you considered checking if the app is running and don't call stop if it is not? Something along the following lines:

```ts
async function stopApp() {
  if (app.state !== 'started') return;
  await app.stop();
}
```

Maybe the `boot` method is adding more states (e.g. `booting`, `booted`) and we need to modify the checks in `start` and `stop` to account for them? IDK.

If we decide to keep ignoring the error:

Silently swallowing all errors is almost always a bad idea, because if an unexpected error happens and the rest of the test suite starts to behave differently as a result, then we have no way how to identify the problem. As a general rule:
- Try to narrow down the kinds of ignored errors as much as possible. Ideally, the error thrown by `app.stop()` should have a unique `code` property we can check.
- Print the ignored error via a `debug` log, so that when we need to troubleshoot this part, we have a way how to see all details about the ignored error.

packages/rest-crud/README.md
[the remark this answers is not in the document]
bajtos: FYI, we have a follow-up task to write proper documentation both for application developers consuming model-api-booter extensions and for extension developers building such extensions, see https://github.com/strongloop/loopback-next/issues/2740

Having said that, I can see how the current README content could be confusing. How about approaching this text from the point of view of what the app developers wants to accomplish using this feature?

For example:

> To expose CRUD REST API operations for a model, you should create a new per-model configuration files inside `src/model-endpoints` directory. In this file, you should provide configure what model you are exposing (e.g. `model: Product`), what API pattern to use (`pattern: 'CrudRest'`), what dataSource to use to access model data (e.g. `dataSource: 'db'`) and finally the base path at which the REST API operations should be exposed (e.g. `basePath: '/products'`).

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (!(modelClass.prototype instanceof Entity)) {
+      throw new Error(
+        `CrudRestController requires a model that extends 'Entity'. (Model name ${modelName} does not extend 'Entity')`,
+      );
+    }
+    const entityClass = modelClass as typeof Entity & {prototype: Entity};
+
+    let repoClassName = entityClass.name + 'Repository';
+
+    if (application.isBound('repositories.' + repoClassName)) {
```
bajtos: Please add a test to verify this scenario. AFAICT, it's not covered now.

https://coveralls.io/builds/28705524/source?filename=packages/rest-crud/src/crud-rest-builder.plugin.ts#L56

packages/rest-crud/src/crud-rest-builder.plugin.ts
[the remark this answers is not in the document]
bajtos: Relations are not supported yet, there is an entire new epic to implement that - see https://github.com/strongloop/loopback-next/issues/2483

packages/rest-crud/src/crud-rest-builder.plugin.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param entityClass - the Entity class the repository is built for
+ * @param config - configuration of the Entity class
+ */
+function setupCrudRepository(
+  entityClass: typeof Entity & {prototype: Entity},
+  config: ModelCrudRestApiConfig,
+): Class<EntityCrudRepository<Entity, unknown>> {
+  const repositoryClass = defineCrudRepositoryClass(entityClass);
+
+  inject(`datasources.${config.dataSource}`)(repositoryClass, undefined, 99);
```
bajtos: Where is the number `99` coming from?

packages/rest-crud/src/crud-rest-builder.plugin.ts
bajtos: Somehow I am not able to comment in https://github.com/strongloop/loopback-next/pull/4589#discussion_r377742892, so I am continuing the discussion here.

> Can we create a constant and use that instead of 0? Perhaps also one for the undefined value? So both constants can provide some clarity to the method call?

```
inject(`datasources.${modelConfig.dataSource}`)(
    repositoryClass,
    UNDEFINED_CLASS_CONSTRUCTOR_FUNCTION,
    INDEX_OF_INJECTED_DATASOURCE_PARAMETER_INSIDE_REPOSITORY_CONSTRUCTOR,
```

Personally, I prefer to create a named function for setting up injection of the first constructor argument.

```ts
// usage
injectFirstConstructorArg(controllerClass, `repositories.${entityClass.name}Repository`);
injectFirstConstructorArg(repositoryClass, `datasources.${config.dataSource}`);

// implementation
function injectFirstConstructorArg(ctor, key) {
  inject(key)(ctor, undefined /* constructor member */, 0 /* the first argument */);
}
```

That way the code inside `setupCrudRepository` and `setupCrudRestController` can stay high-level and still communicate the intent well. People interested in implementation details of `injectFirstConstructorArgument` can look into function body, at which point they should be prepared to encounter code assuming knowledge of low-level decorator details.

packages/rest-crud/src/crud-rest-builder.plugin.ts
bajtos: On the other hand, since this file contains `CrudRestApiBuilder` class which is bound `asModelApiBuilder`, maybe `crud-rest.model-api-builder.ts` would be a good name, using `model-api-builder` as the type suffix (similar to `controller` and `repository` we use elsewhere). If `model-api-builder` is too long, the perhaps `api-builder` is good enough too? (I.e. `crud-rest.api-builder.ts`).

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: Sorry I didn't mean to push this, I'll change it back to 0.

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: I couldn't find a good way to test this (https://github.com/strongloop/loopback-next/pull/4589#discussion_r378684011) since the binding for both repositories would be the same since they would share the same name so I don't know if there's a way to verify it's using the repository that was bound before. I have an attempt in [this commit](https://github.com/strongloop/loopback-next/pull/4589/commits/324606155a1ae8ee8e8f1160a1782dbdffa41ec0#diff-858d1af4393d324a034e5a75fbdd5b64R60).

packages/rest-crud/README.md
nabdelgadir: > So they are mutually exclusive then? If so let's specify that Basic and Advanced are mutually exclusive.

Yeah, I'll specify that, thanks!

> Another question: why would someone want to do Advanced steps if it is shorter to perform Basic steps?

The advanced one helps if you only want to create the default controller or default repository but not both.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
nabdelgadir: > Does it mean that app.stop() throws an error if the app was not started? I find that surprising, because the implementation of stop does seem to short-circuit to a no-op when not running.

If the application hasn't started, it'll throw an error from this function:

https://github.com/strongloop/loopback-next/blob/crud-rest/packages/core/src/application.ts#L225-L230

> ```ts
> async function stopApp() {
>   if (app.state !== 'started') return;
>   await app.stop();
> }
> ```

I think this is a good way to check (since `app.state` is otherwise `booting` and in that case we don't want to stop the application), so I'll use it.

packages/rest-crud/src/crud-rest-builder.plugin.ts
nabdelgadir: I like `crud-rest.api-builder.ts` because it's shorter (and the builder is `CrudRestApiBuilder` instead of `CrudRestModelApiBuilder`, so the `Model` part is implied).

packages/rest-crud/README.md
```diff
@@ -15,11 +15,60 @@ npm install --save @loopback/rest-crud
 
 ## Basic use
 
-`@loopback/rest-crud` exposes two helper methods (`defineCrudRestController` and
-`defineCrudRepositoryClass`) for creating controllers and respositories using
-code.
+`@loopback/rest-crud` can be used along with `@loopback/model-api-builder` to
```
bajtos: I think this is not entirely correct. `@loopback/model-api-builder` provides only _Types and helpers for packages contributing Model API builders_ (quoting from README).

Did you perhaps mean the built-in [ModelApiBooter](https://github.com/strongloop/loopback-next/blob/master/packages/boot/src/booters/model-api.booter.ts)?

```suggestion
`@loopback/rest-crud` can be used along with the built-in `ModelApiBooter` to
```

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
bajtos: >  I also tried creating a separate repository class and using that one but thinking about it now, I don't think there's a way to differentiate between them since their bindings are the same, so I might remove this test.

IMO, we should add the test back and modify the assertions to verify that the repository binding is using the user-provided repository class, not a repository class created by the model-api-builder. You can use `valueConstructor` property to access the class passed to `.toClass()` method when the `Binding` was configured.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
```diff
@@ -0,0 +1,130 @@
...
+  async function givenAppWithDataSource() {
+    app = new BooterApp({
+      rest: givenHttpServerConfig(),
+    });
+    app.dataSource(new juggler.DataSource({connector: 'memory'}), 'db');
+  }
+
+  async function stopApp() {
+    if (app.state !== 'started') return;
+    await app.stop();
```
bajtos: 👍

packages/rest-crud/src/crud-rest.api-builder.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @param key - binding to use in order to resolve the value of the decorated
+ * constructor parameter or property
+ */
+function injectFirstConstructorArg<T>(
+  ctor: Constructor<T>,
+  key: BindingSelector,
+) {
+  inject(key)(
+    ctor,
+    undefined /* constructor member */,
```
bajtos: Nitpick:

```suggestion
    undefined, // constructor member
```

packages/rest-crud/src/crud-rest.api-builder.ts
```diff
@@ -0,0 +1,144 @@
...
+    0 /* the first argument */,
```
bajtos: Nitpick:

```suggestion
    0, // the first argument
```

packages/rest-crud/src/crud-rest.api-builder.ts
```diff
@@ -0,0 +1,144 @@
...
+    const entityClass = modelClass as typeof Entity & {prototype: Entity};
+
+    let repoClassName = entityClass.name + 'Repository';
+
+    if (application.isBound('repositories.' + repoClassName)) {
+      debug(
+        'Using repository class',
+        repoClassName,
+        ', as it is already bound to application',
+      );
```
bajtos: _Nitpick, feel free to ignore._

We usually use `%s` to format debug logs.


```ts
debug(
  'Using repository class %s, as it is already bound to application.', 
  repoClassName
);
```

bajtos: Please note the actual binding can be bound to a class with a different name, or even to something else - a provider, a static value (class instance).

I think a debug message along the following line would be more precise:

```js
debug(
  'Using the existing Repository binding %j',
  'repositories.' + repoClassName,
);
```

(By using `%j`, we convert the binding key to JSON, which will wrap the string in double quotes and escape any double quotes inside the key.)

I would also consider extracting the snippet `'repositories.' + repoClassName` into a new variable.

packages/boot/src/__tests__/acceptance/crud-rest-builder.acceptance.ts
bajtos: > I think this is a good way to check (since app.state is otherwise booting and in that case we don't want to stop the application), so I'll use it.

Sounds good :+1:

packages/rest-crud/README.md
nabdelgadir: In this case, `CrudRestApiBuilder` implements `ModelApiBuilder` (not the booter). Quoting from the README: it also "provides a contract for extensions that contribute builders for repositories and controllers".

packages/rest-crud/src/crud-rest-builder.plugin.ts
jannyHou: > Because decorators are converted to JavaScript calls of TypeScript API at compile time, we cannot use decorator syntax when building the repository class dynamically at runtime, we have to invoke the decorator function manually.

Thanks everyone for the reply. This is something I didn't know before 👍

PR #4628, opened 2020-02-11
chore(benchmark): replace request with axios

raymondfeng:
Replacing `request` with `axios`

## Checklist

### DIFF
--- a/benchmark/package-lock.json
+++ b/benchmark/package-lock.json
@@ -22,9 +22,9 @@
 			},
 			"dependencies": {
 				"@types/node": {
-					"version": "13.7.0",
-					"resolved": "https://registry.npmjs.org/@types/node/-/node-13.7.0.tgz",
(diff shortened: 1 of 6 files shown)


bajtos:
In https://github.com/strongloop/loopback-next/issues/2672, I raised the following concern:

Many `request` alternatives are replacing errors reported by Node.js core with its own error classes, which makes troubleshooting extremely difficult.

That was in May 2019. What's the current status of Axios implementation, does it preserve low-level error details?

Ideally, I'd like us to use the same library across the entire loopback-next codebase, including connectors like REST and SOAP.

raymondfeng:
> Many request alternatives are replacing errors reported by Node.js core with its own error classes, which makes troubleshooting extremely difficult.
That was in May 2019. What's the current status of Axios implementation, does it preserve low-level error details?

I found a way to allow lower-level errors to be reported by providing a `validateStatus` function in the config. But this feature is not needed for `benchmark` testing.

> Ideally, I'd like us to use the same library across the entire loopback-next codebase, including connectors like REST and SOAP.

Since this PR focuses on `benchmark` testing which is an internal module. I think `axios` does it job perfectly here. For connecting to other http services, I agree we can have a common `http-client` package.

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/4637 as a follow-up PR.

PR #4637, opened 2020-02-12
feat(http-caching-proxy): replace request package with axios

raymondfeng:
This is part 2 to fix https://github.com/strongloop/loopback-next/issues/2672

## Checklist

### DIFF
--- a/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
+++ b/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
@@ -4,13 +4,19 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {expect} from '@loopback/testlab';
+import axios, {
+  AxiosProxyConfig,
+  AxiosRequestConfig,
+  AxiosResponse,
+} from 'axios';
 import delay from 'delay';
 import http from 'http';
 import {AddressInfo} from 'net';
 import pEvent from 'p-event';
 import path from 'path';
-import makeRequest from 'request-promise-native';
 import rimrafCb from 'rimraf';
+import tunnel, {ProxyOptions as TunnelProxyOptions} from 'tunnel';
+import {URL} from 'url';
 import util from 'util';
 import {HttpCachingProxy, ProxyOptions} from '../../http-caching-proxy';
 
@@ -45,9 +51,7 @@ describe('HttpCachingProxy', () => {
 
     await givenRunningProxy();
     const result = await makeRequest({
-      uri: 'http://example.com',
-      proxy: proxy.url,
-      resolveWithFullResponse: true,
+      url: 'http://example.com',
     });
 
     expect(result.statusCode).to.equal(200);
@@ -62,15 +66,13 @@ describe('HttpCachingProxy', () => {
     await givenRunningProxy({logError: false});
     await expect(
       makeRequest({
-        uri: 'http://does-not-exist.example.com',
-        proxy: proxy.url,
-        resolveWithFullResponse: true,
+        url: 'http://does-not-exist.example.com',
       }),
     ).to.be.rejectedWith(
       // The error can be
       // '502 - "Error: getaddrinfo EAI_AGAIN does-not-exist.example.com:80"'
       // '502 - "Error: getaddrinfo ENOTFOUND does-not-exist.example.com'
-      /502 - "Error\: getaddrinfo/,
+      /502 - "Error: getaddrinfo/,
     );
   });
 
@@ -78,12 +80,10 @@ describe('HttpCachingProxy', () => {
     await givenRunningProxy({logError: false, timeout: 1});
     await expect(
       makeRequest({
-        uri:
+        url:
           'http://www.mocky.io/v2/5dade5e72d0000a542e4bd9c?mocky-delay=1000ms',
-        proxy: proxy.url,
-        resolveWithFullResponse: true,
       }),
-    ).to.be.rejectedWith(/502 - "Error: ETIMEDOUT"/);
+    ).to.be.rejectedWith(/502 - "Error: timeout of 1ms exceeded/);
   });
 
   it('proxies HTTPs requests (no tunneling)', async function() {
@@ -93,10 +93,7 @@ describe('HttpCachingProxy', () => {
 
     await givenRunningProxy();
     const result = await makeRequest({
-      uri: 'https://example.com',
-      proxy: proxy.url,
-      tunnel: false,
-      resolveWithFullResponse: true,
+      url: 'https://example.com',
     });
 
     expect(result.statusCode).to.equal(200);
@@ -105,16 +102,17 @@ describe('HttpCachingProxy', () => {
 
   it('rejects CONNECT requests (HTTPS tunneling)', async () => {
     await givenRunningProxy();
+    const agent = tunnel.httpsOverHttp({
+      proxy: getTunnelProxyConfig(proxy.url),
+    });
     const resultPromise = makeRequest({
-      uri: 'https://example.com',
-      proxy: proxy.url,
-      tunnel: true,
-      simple: false,
-      resolveWithFullResponse: true,
+      url: 'https://example.com',
+      httpsAgent: agent,
+      proxy: false,
     });
 
     await expect(resultPromise).to.be.rejectedWith(
-      /tunneling socket.*statusCode=501/,
+      /tunneling socket could not be established, statusCode=501/,
     );
   });
 
@@ -123,11 +121,9 @@ describe('HttpCachingProxy', () => {
     givenServerDumpsRequests();
 
     const result = await makeRequest({
-      uri: stubServerUrl,
-      json: true,
+      url: stubServerUrl,
+      responseType: 'json',
       headers: {'x-client': 'test'},
-      proxy: proxy.url,
-      resolveWithFullResponse: true,
     });
 
     expect(result.headers).to.containEql({
@@ -144,12 +140,11 @@ describe('HttpCachingProxy', () => {
 
     const result = await makeRequest({
       method: 'POST',
-      uri: stubServerUrl,
-      body: 'a text body',
-      proxy: proxy.url,
+      url: stubServerUrl,
+      data: 'a text body',
     });
 
-    expect(result).to.equal('a text body');
+    expect(result.body).to.equal('a text body');
   });
 
   it('caches responses', async () => {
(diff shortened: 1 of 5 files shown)


packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -62,28 +82,26 @@ describe('HttpCachingProxy', () => {
...
-        resolveWithFullResponse: true,
+        url: 'http://does-not-exist.example.com',
+        proxy: getProxyConfig(),
       }),
     ).to.be.rejectedWith(
       // The error can be
       // '502 - "Error: getaddrinfo EAI_AGAIN does-not-exist.example.com:80"'
       // '502 - "Error: getaddrinfo ENOTFOUND does-not-exist.example.com'
-      /502 - "Error\: getaddrinfo/,
+      /502 - getaddrinfo/,
```
bajtos: Please update the example errors in the code comment above to match the real error messages.

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -4,20 +4,41 @@
...
+const makeRequest = async (config: AxiosRequestConfig) => {
+  const res = await axiosInstance(config);
+  // Throw an error with message from the original error
+  if (res.status >= 300) {
+    const err = new Error(`${res.status} - ${res.data}`);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (err as any).status = res.status;
+    throw err;
+  }
+  return res;
```
bajtos: Have you considered enhancing `res` object with `body` and `statusCode` properties, so that we don't have to modify all tests below?

```suggestion
  return Object.assign(res, {statusCode: res.statusCode, body: res.data});
```

packages/http-caching-proxy/src/http-caching-proxy.ts
```diff
@@ -225,18 +250,19 @@ export class HttpCachingProxy {
     // Since this proxy is for testing only, buffering the entire
     // response body is acceptable.
 
+    let data = backendResponse.data;
+    if (typeof data !== 'string') {
+      data = JSON.stringify(data);
```
bajtos: ❗️ I find this problematic because we are reformatting JSON responses here. If the server responds with formatted JSON data (think of `JSON.stringify(value, null, 2)`), then our proxy will return JSON body with no formatting.

Even worse, if the original server returns non-JSON payload that's parsed by Axios (e.g. XML, see https://github.com/axios/axios/issues/667), then our proxy will change the response type! 

Please find a way how to forward the client request body and the backend response exactly as provided by the client and the backend, with no changes made by the proxy.

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -4,20 +4,41 @@
...
 
 const CACHE_DIR = path.join(__dirname, '.cache');
 
 const rimraf = util.promisify(rimrafCb);
 
+const axiosInstance = axios.create({
+  // Provide a custom function to control when Axios throws errors based on
+  // http status code. Please note that Axios creates a new error in such
+  // condition and the original low-level error is lost
+  validateStatus: () => true,
```
bajtos: I am not very happy about this part, it looks rather hacky to me. I also dislike the fact that when using Axios, we must be remember to modify the default configuration if we want to get decent error handling. I am pretty sure we won't be consistent and in some places we will forget to do the necessary tweaks, thus making it difficult for our users to troubleshoot request errors they may encounter in production. What's worse, when they fill an issue about the problem, it's unlikely that we will remember by that time what's the culprint and how to fix Axios configuration.

bajtos: I am taking back the comment above. I checked Axios' source code and run our tests with DEBUG statements enabled, and both checks confirmed that Axios is preserving original error details & stack traces when an error is reported by Node.js core.

Here is the relevant code: https://github.com/axios/axios/blob/03e6f4bf4c1eced613cf60d59ef50b0e18b31907/lib/adapters/http.js#L224-L245

As I understand `validateStatus`, it determines whether a successfully received HTTP response is converted into a resolved or a rejected promise, depending on the status code. (I.e. whether `await axios(/*...*/)` should throw an error when the server responds with `500 Internal Server Error`.)

bajtos: Quoting from Axios docs:

```
  // `validateStatus` defines whether to resolve or reject the promise for a given
  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
  // or `undefined`), the promise will be resolved; otherwise, the promise will be
  // rejected.
```

Have you considered using the simpler `null` variant?

```suggestion
  validateStatus: null,
```

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -4,20 +4,41 @@
...
+});
+
+const makeRequest = async (config: AxiosRequestConfig) => {
```
bajtos: It's our best practice to put test helper at the end of the file, after the tests - see https://loopback.io/doc/en/contrib/style-guide-es6.html#layout-of-test-files

Please move `makeRequest` and `axiosInstance` to the bottom.

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -216,6 +230,31 @@ describe('HttpCachingProxy', () => {
     await proxy.stop();
   }
 
+  function getProxyConfig(): AxiosProxyConfig {
```
bajtos: Have you considered calling `getProxyConfig` from inside `makeRequest`, so that `makeRequest` can keep accepting `proxy` configuration as a string URL? That way there is less things to change in our tests.

In longer term, we can contribute an enhancement to Axios to let is support `proxy` specified as a string.

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
```diff
@@ -62,28 +82,26 @@ describe('HttpCachingProxy', () => {
...
 
   it('reports timeout error for HTTP requests', async function() {
     await givenRunningProxy({logError: false, timeout: 1});
     await expect(
       makeRequest({
-        uri:
+        url:
           'http://www.mocky.io/v2/5dade5e72d0000a542e4bd9c?mocky-delay=1000ms',
-        proxy: proxy.url,
-        resolveWithFullResponse: true,
```
bajtos: `resolveWithFullResponse` is request's counter-part of Axios' `validateStatus`. Have you considered configuring `validateStatus` on a per-test basis, as we are doing with `resolveWithFullResponse` now?

packages/http-caching-proxy/src/http-caching-proxy.ts
```diff
@@ -124,7 +132,8 @@ export class HttpCachingProxy {
   private _handle(request: IncomingMessage, response: ServerResponse) {
     const onerror = (error: Error) => {
       this.logError(request, error);
-      response.statusCode = error.name === 'RequestError' ? 502 : 500;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      response.statusCode = (error as any).statusCode || 502;
       response.end(error.message);
```
bajtos: Should we preserve the error name (`Error`, `TypeError`, etc.) in the error response body returned by our proxy?

```suggestion
      response.end(error.name + ': ' + error.message);
```

bajtos:
Based on the investigation I did today, I am happy to choose Axios as our replacement for request 🎉 

Let's improve implementation details now.

Also, I am not confident that the switch from request to axios is fully backwards compatible in all edge cases, let's make this a semver-major change please.

bajtos:
> coverage/coveralls — Coverage decreased (-0.01%) to 92.492%

I see quite few places in `http-caching-proxy.ts` not covered by our test suite, can you please check? I am mostly concerned about code supposed to handle error situations but not being verified to work as intended (see [here](https://coveralls.io/builds/28714794/source?filename=packages/http-caching-proxy/src/http-caching-proxy.ts#L228) and [here](https://coveralls.io/builds/28714794/source?filename=packages/http-caching-proxy/src/http-caching-proxy.ts#L143)).

packages/http-caching-proxy/src/http-caching-proxy.ts
```diff
@@ -187,27 +196,43 @@ export class HttpCachingProxy {
...
+        'Got response for %s %s -> %s',
+        clientRequest.method,
+        clientRequest.url,
+        backendResponse.status,
+        backendResponse.headers,
+        backendResponse.data,
+      );
+    } catch (err) {
+      debug(
+        'Error sending request %s %s',
```
bajtos: Is this debug statement needed? I believe the error will be logged by `onerror` handler, no need to log it again here.

packages/http-caching-proxy/src/http-caching-proxy.ts
```diff
@@ -187,27 +196,43 @@ export class HttpCachingProxy {
...
+        clientRequest.method,
+        clientRequest.url,
+        err,
+      );
+      throw err;
+    }
+
+    if (backendResponse.status >= 300) {
```
bajtos: Would it be better to use `validateStatus` and let Axios to throw errors when the server responds with an error status code?

bajtos:
FYI: I opened a feature request in Axios to make their API closer to `request`, see https://github.com/axios/axios/issues/2741

packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
raymondfeng: I prefer to make it explicit. Personally I find it confusing if `null` and `undefined` have different meanings in the config.

packages/http-caching-proxy/src/http-caching-proxy.ts
raymondfeng: I removed the code so the proxy simply passes through.

packages/http-caching-proxy/src/http-caching-proxy.ts
raymondfeng: Good catch. I forced axios to use `arraybuffer` now.

raymondfeng:
@bajtos PTAL.

bajtos:
@raymondfeng As I commented elsewhere of before, I'd like us to release this change as a semver-major release. `axios` may behave differently than `request` in edge cases, I'd rather make this change explicit in the release notes. Can you please add `BREAKING CHANGE` footer to the commit message?

Issue #4425, opened 2020-01-14
npm test hangs infinitly on windows


derdeka:
## Steps to reproduce

- check out loopback-next repo
- `npm install`
- `npm test` or `DEBUG=mocha:* npm test`

## Current Behavior

Many tests are executed - one test does not end. `npm test` never ends and runs infinitely. 
Tried with `cmd`, `powershell` and git `bash`.

> mocha:runner run suite app-generator (SLOW) +17ms

^^^^^^is the last logged console output.

## Expected Behavior

`npm test` should end successfully or with some error message.

## Additional information

Windows 10 Pro ver 1909
node v10.15.3
Docker version 19.03.4, build 9013bf5
@loopback/cli version: 1.28.1

when executing `DEBUG=mocha:* npm test` these are the last few log messages:
```bash
..  mocha:runner run suite TodoApplication +60ms
  mocha:runnable timeout 30000 +11s
  mocha:runnable timeout 30000 +2s
...  mocha:runnable timeout 30000 +25ms
..  mocha:runner run suite TodoApplication when dealing with a single persisted todo +3s
........  mocha:runner run suite GeoLookupService +71ms
.  mocha:runner run suite TodoController +18ms
  mocha:runner run suite TodoController createTodo +0ms
..  mocha:runner run suite TodoController findTodoById +37ms
.  mocha:runner run suite TodoController findTodos +14ms
...  mocha:runner run suite TodoController replaceTodo +36ms
.  mocha:runner run suite TodoController updateTodo +12ms
.  mocha:runner run suite TodoController deleteTodo +11ms
.  mocha:runner run suite app-generator (SLOW) +17ms
  mocha:runnable timeout 60000 +235ms
```

dougal83:
I found it runs horribly on Win10 ([other issues, /w some tests](https://github.com/strongloop/loopback4-example-shopping/issues/452)).  The best solution for me was to use a virtual machine with linux.

If you have HyperV capable PC then it's extremely quick to set up a VM.

bajtos:
This is a known issue, but also difficult to reproduce locally. See the discussion in #1350 and nodejs/node#21210 for more details.

To avoid this issue on CI, we are using Node.js 8.x on Windows:

https://github.com/strongloop/loopback-next/blob/c1cff12ba579975e17a9f0d03e8b8bf0d3e3c09a/appveyor.yml#L3

[...]

PR #4652, opened 2020-02-14
fix(cli): correct root directory in test

dougal83:
Fixes error on windows tests.  Removes instances of sandbox folder
erroniously placed in packages folder.

See #4425

## Checklist

### DIFF
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -217,7 +217,7 @@ process.env.CI && !process.env.DEBUG
 
 /** For testing if the generator handles default values properly */
 describe('app-generator with default values', () => {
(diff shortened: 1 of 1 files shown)


bajtos:
Thank you for the pull request. I think it may be an intentional decision to place the sandbox directory into `packages`, because that way `lerna bootstrap` will install dependencies using monorepo-local links to resolve LB4 packages. I don't know if this is applicable to the particular tests you are modifying.

AFAICT, one of the problematic tests was introduced by https://github.com/strongloop/loopback-next/commit/630210117365e8d89f8dd3b797f0b0096936b79e by @agnes512.

Agnes, do you happen to remember what was your intention? What is the correct `rootDir` value to use?

agnes512:
I think @dougal83 is right. The test was supposed to be tested in sandbox. I made a mistake, my bad D:

PR #4655, opened 2020-02-14
feat(context): improve ctx.inspect() to allow classes with colliding names

raymondfeng:
Classes with the same name can be declared and bound from different scopes
or modules. For inspection, we need to have unique names to treat them
differently.

## Checklist

### DIFF
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -34,6 +34,7 @@ import {
 } from './resolution-session';
 import {
   BoundValue,
+  Constructor,
   getDeepProperty,
   isPromiseLike,
   ValueOrPromise,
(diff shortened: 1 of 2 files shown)


packages/context/src/context.ts
```diff
@@ -797,27 +798,81 @@ export class Context extends EventEmitter {
...
+
+  visit(ctor: Constructor<unknown>) {
+    let name = this.classes.get(ctor);
+    if (name == null) {
+      name = ctor.name;
+      // Now check if the name collides with another class
+      let index = this.nameIndex.get(name);
+      if (typeof index === 'number') {
+        // A conflict is found, mangle the name
+        this.nameIndex.set(name, ++index);
```
emonddr: The resulting names look a lot like the names of nested java classes.  `Myprovider$1` .

For bindings, you gave them an unique id: `Binding_2_3`  for example.
Can we do the same?

raymondfeng: Binding ids are not generated by `inspect()`.

I was trying to use `_1` but there is a possibility that a class named `MyProvider_1` may exist. I don't want to make the name mangling too complex.

emonddr: I see ok.

packages/context/src/context.ts
```diff
@@ -797,27 +798,81 @@ export class Context extends EventEmitter {
...
+        name = `${name}$${index}`;
```
bajtos: Can we use a space character as the separator? I believe spaces are not allowed in JavaScript class/function names.

```suggestion
        name = `${name} #${index}`;
```

(`Myprovider #1`, `Binding #2`, etc.)

PR #4657, opened 2020-02-15
fix: extend test timeout for windows

### LINKED ISSUE #4425: npm test hangs infinitly on windows

dougal83:
Give windows extra time to complete tests
Fixes #4425

fix: cleanup tilde-path-app post test …
ensure test app is remove afterward
Fixes #4652

## Checklist

### DIFF
--- a/packages/cli/test/integration/generators/app.integration.js
+++ b/packages/cli/test/integration/generators/app.integration.js
@@ -263,7 +263,11 @@ describe('app-generator with tilde project path', () => {
   const cwd = process.cwd();
 
   // If the test runs outside $home directory
-  if (process.env.CI && !process.env.DEBUG && tildify(sandbox) === sandbox) {
+  const runsOutsideRoot =
+    process.env.CI && !process.env.DEBUG && tildify(sandbox) === sandbox
+      ? true
+      : false;
+  if (runsOutsideRoot) {
     sandbox = path.join(os.homedir(), '.lb4sandbox/tilde-path-app');
     pathWithTilde = '~/.lb4sandbox/tilde-path-app';
   }
@@ -295,7 +299,12 @@ describe('app-generator with tilde project path', () => {
     // eslint-disable-next-line no-invalid-this
     this.timeout(30 * 1000);
 
-    process.chdir(sandbox);
+    // Handle special case - Skipping... not inside the project root directory.
+    if (runsOutsideRoot) {
+      process.chdir(sandbox);
+    } else {
+      process.chdir(rootDir);
+    }
     build.clean(['node', 'run-clean', sandbox]);
     process.chdir(cwd);
   });
--- a/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
+++ b/packages/http-caching-proxy/src/__tests__/integration/http-caching-proxy.integration.ts
@@ -196,7 +196,10 @@ describe('HttpCachingProxy', () => {
     expect(result2).to.equal('2');
   });
 
-  it('handles the case where backend service is not running', async () => {
+  it('handles the case where backend service is not running', async function() {
+    // This test takes a bit longer to finish on windows.
+    // eslint-disable-next-line no-invalid-this
+    this.timeout(3000);
     await givenRunningProxy({logError: false});
 
     await expect(
--- a/packages/http-server/src/__tests__/integration/http-server.integration.ts
+++ b/packages/http-server/src/__tests__/integration/http-server.integration.ts
@@ -44,7 +44,10 @@ describe('HttpServer (integration)', () => {
       .expect(200);
   });
 
-  it('stops server', async () => {
+  it('stops server', async function() {
+    // This test takes a bit longer to finish on windows.
+    // eslint-disable-next-line no-invalid-this
+    this.timeout(3000);
     const serverOptions = givenHttpServerConfig();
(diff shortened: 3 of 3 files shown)


dougal83:
@derdeka If you have time, could you please run tests against this PR?

packages/cli/test/integration/generators/app.integration.js
```diff
@@ -295,7 +295,12 @@ describe('app-generator with tilde project path', () => {
     // eslint-disable-next-line no-invalid-this
     this.timeout(30 * 1000);
 
-    process.chdir(sandbox);
+    // If the test runs outside $home directory
+    if (process.env.CI && !process.env.DEBUG) {
+      process.chdir(sandbox);
+    } else {
+      process.chdir(rootDir);
+    }
```
dougal83: @agnes512 Does this look reasonable?  I had to do this for CI test but I'm not sure how that works.  I've also dropped a clause to compare the sandbox path from L266.

packages/cli/test/integration/generators/app.integration.js
```diff
@@ -295,7 +295,12 @@ describe('app-generator with tilde project path', () => {
     // eslint-disable-next-line no-invalid-this
     this.timeout(30 * 1000);
 
-    process.chdir(sandbox);
+    // If the test runs outside $home directory
+    if (process.env.CI && !process.env.DEBUG) {
```
bajtos: I am little bit concerned about duplicating the logic deciding whether to use `sandbox` or `rootDir`, because if we change the original condition [on line 266](https://github.com/strongloop/loopback-next/blob/9aba52e59c80830dd0871c4a087b90ec2aa21db2/packages/cli/test/integration/generators/app.integration.js#L266), then how are we going to ensure this copy is updated as well?

If we want to change to a different directory, then let's introduce a new variable set to true/false depending on whether the condition on line 266 was triggered or not.

Having said that, I am not sure if changing the directory to `rootDir` is the right thing to do, see my comment below.

packages/cli/test/integration/generators/app.integration.js
bajtos: Hmm, I find this suspicious. On the next line, we are running a cleanup script in the `sandbox` directory:

https://github.com/strongloop/loopback-next/blob/9aba52e59c80830dd0871c4a087b90ec2aa21db2/packages/cli/test/integration/generators/app.integration.js#L304

Does the behavior of `run-clean` depend on the current working directory? If it does not, then perhaps we can get rid of this first `process.chdir` entirely, and keep the last `chdir(cwd)` only?

```ts
   after(function() {
    // Increase the timeout to accommodate slow CI build machines
    // eslint-disable-next-line no-invalid-this
    this.timeout(30 * 1000);

    build.clean(['node', 'run-clean', sandbox]);
    process.chdir(cwd);
  });
```

packages/cli/test/integration/generators/app.integration.js
dougal83: > If we want to change to a different directory, then let's introduce a new variable set to true/false depending on whether the condition on line 266 was triggered or not.

New variable utilised.

packages/cli/test/integration/generators/app.integration.js
dougal83: This appears to be a special case for tilde check. We are using sandbox that is outside the code's project root. Not setting the `process.chdir` gives for example: `Skipping /home/dev/workspace/loopback-next/sandbox/tilde-path-app as it is not inside the project root directory.`

dougal83:
`npm ERR! 429 Too Many Requests - GET https://registry.npmjs.org/@commitlint/cli/-/cli-8.3.5.tgz`

Usually Cloudflare?  Will manually restart test later.

dougal83:
> The `chdir` fix is not pretty, but as long as it works, I think it's best to land it now and then clean it up later (if at all).

Amen. 🙏

PR #4658, opened 2020-02-15
Custom ajv formats & keywords

mastermunj:
Allow custom formats & keywords for AJV validations. Custom formats are useful in cases when certain formats are common across models. Custom keywords on other cases are useful for validating cush formats against the database if needed.

## Checklist

### DIFF
--- a/packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
+++ b/packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
@@ -7,9 +7,7 @@
 
 'use strict';
 
-exports[
-  `lb4 relation checks generated source class repository  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} generates Order repository file with different inputs 1`
-] = `
+exports[`lb4 relation checks generated source class repository answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} generates Order repository file with different inputs 1`] = `
 import {DefaultCrudRepository, BelongsToAccessor, repository} from '@loopback/repository';
 import {Order, Customer} from '../models';
 import {DbDataSource} from '../datasources';
@@ -36,9 +34,8 @@ export class OrderRepository extends DefaultCrudRepository<
 
 `;
 
-exports[
-  `lb4 relation checks generated source class repository  answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"customer","registerInclusionResolver":true} generates OrderClass repository file with different inputs 1`
-] = `
+
+exports[`lb4 relation checks generated source class repository answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"customer","registerInclusionResolver":true} generates OrderClass repository file with different inputs 1`] = `
 import {DefaultCrudRepository, repository, BelongsToAccessor} from '@loopback/repository';
 import {OrderClass, CustomerClass} from '../models';
 import {MyDBDataSource} from '../datasources';
@@ -61,9 +58,8 @@ export class OrderClassRepository extends DefaultCrudRepository<
 
 `;
 
-exports[
-  `lb4 relation checks generated source class repository  answers {"relationType":"belongsTo","sourceModel":"OrderClassType","destinationModel":"CustomerClassType","relationName":"customer","registerInclusionResolver":false} generates OrderClassType repository file with different inputs 1`
-] = `
+
+exports[`lb4 relation checks generated source class repository answers {"relationType":"belongsTo","sourceModel":"OrderClassType","destinationModel":"CustomerClassType","relationName":"customer","registerInclusionResolver":false} generates OrderClassType repository file with different inputs 1`] = `
 import {DefaultCrudRepository, repository, BelongsToAccessor} from '@loopback/repository';
 import {OrderClassType, CustomerClassType} from '../models';
 import {MyDBDataSource} from '../datasources';
@@ -85,9 +81,8 @@ export class OrderClassTypeRepository extends DefaultCrudRepository<
 
 `;
 
-exports[
-  `lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} checks controller content with belongsTo relation 1`
-] = `
+
+exports[`lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} checks controller content with belongsTo relation 1`] = `
 import {
   repository,
 } from '@loopback/repository';
@@ -129,16 +124,14 @@ export class OrderCustomerController {
 
 `;
 
-exports[
-  `lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} the new controller file added to index.ts file 1`
-] = `
+
+exports[`lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} the new controller file added to index.ts file 1`] = `
 export * from './order-customer.controller';
 
 `;
 
-exports[
-  `lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} checks controller content with belongsTo relation 1`
-] = `
+
+exports[`lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} checks controller content with belongsTo relation 1`] = `
 import {
   repository,
 } from '@loopback/repository';
@@ -180,16 +173,14 @@ export class OrderClassCustomerClassController {
 
 `;
 
-exports[
-  `lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} the new controller file added to index.ts file 1`
-] = `
+
+exports[`lb4 relation checks if the controller file created  answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} the new controller file added to index.ts file 1`] = `
 export * from './order-class-customer-class.controller';
 
 `;
 
-exports[
-  `lb4 relation generates model relation for existing property name verifies that a preexisting property will be overwritten 1`
-] = `
+
+exports[`lb4 relation generates model relation for existing property name verifies that a preexisting property will be overwritten 1`] = `
 import {Entity, model, property, belongsTo} from '@loopback/repository';
 import {Customer} from './customer.model';
 
@@ -217,9 +208,8 @@ export class Order extends Entity {
 
 `;
 
-exports[
-  `lb4 relation generates model relation with custom relation name answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer","foreignKeyName":"customerId","relationName":"my_customer"} relation name should be my_customer 1`
-] = `
+
+exports[`lb4 relation generates model relation with custom relation name answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer","foreignKeyName":"customerId","relationName":"my_customer"} relation name should be my_customer 1`] = `
 import {Entity, model, property, belongsTo} from '@loopback/repository';
 import {Customer} from './customer.model';
 
@@ -247,9 +237,8 @@ export class Order extends Entity {
 
 `;
 
-exports[
-  `lb4 relation generates model relation with custom relation name answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} relation name should be my_customer 1`
-] = `
+
+exports[`lb4 relation generates model relation with custom relation name answers {"relationType":"belongsTo","sourceModel":"OrderClass","destinationModel":"CustomerClass","relationName":"my_customer"} relation name should be my_customer 1`] = `
 import {Entity, model, property, belongsTo} from '@loopback/repository';
 import {CustomerClass} from './customer-class.model';
 
@@ -276,9 +265,8 @@ export class OrderClass extends Entity {
 
 `;
 
-exports[
-  `lb4 relation generates model relation with default values answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} has correct default imports 1`
-] = `
+
+exports[`lb4 relation generates model relation with default values answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} has correct default imports 1`] = `
 import {Entity, model, property, belongsTo} from '@loopback/repository';
 import {Customer} from './customer.model';
 
--- a/packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
+++ b/packages/cli/snapshots/integration/generators/relation.has-many.integration.snapshots.js
@@ -7,9 +7,7 @@
 
 'use strict';
 
-exports[
-  `lb4 relation HasMany checks generated source class repository answers {"relationType":"hasMany","sourceModel":"Customer","destinationModel":"Order"} generates Customer repository file with different inputs 1`
-] = `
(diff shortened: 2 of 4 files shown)


mastermunj:
Not sure how review request went to so many of you, maybe I selected a wrong option while raising PR or it's the default behavior.

dougal83:
> Not sure how review request went to so many of you, maybe I selected a wrong option while raising PR or it's the default behavior.

Don't worry, it's automatic based on files changed.

mastermunj:
Is it possible to run Travis manually? On my system, all tests run successfully. Also, the test which has failed is related to timeout.

https://travis-ci.com/strongloop/loopback-next/jobs/287666240#L243

Also, any inputs for code coverage decrease?

docs/site/Parsing-requests.md
```diff
@@ -234,6 +234,76 @@ class Product extends Entity {
...
+        async: true,
+        validate: async (code: string) => {
+          return /^[a-zA-Z]{3}-\d{3}$/.test(code)
+            ? Promise.resolve(true)
+            : Promise.resolve(false);
+        },
+      },
+    },
+  },
+});
```
achrinza: There's already a `pattern` key under the  `@property`. This  would, AFAIK, reflect under the OpenAPI 3 schema, while AJV custom formats won't as it's a function. The pattern can also be stored inside a variable for re-use, similar to what AJV custom formats is doing.

Is there a benefit to using AJV's Custom Formats?

docs/site/Parsing-requests.md
```diff
@@ -234,6 +234,76 @@ class Product extends Entity {
...
+AJV Custom Formats can be configured by binding the value
+`validation.ajvCustomFormats` to `RestBindings.REQUEST_BODY_PARSER_OPTIONS`
+('rest.requestBodyParserOptions').
+
+For example,
+
+```ts
+server.bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS).to({
+  validation: {
+    ajvCustomFormats: {
```
achrinza: I think it should be named something else other than `ajvCustomFormats` to provide abstraction on the `AJV` module (The `pattern` key in `@properties` does not have the `ajv` prefix)

docs/site/Parsing-requests.md
mastermunj: The example I have used is a simpler one, however, use cases are there.

A `format` for ajv can do more than just pattern match.

Example from ajv library that does the async process for a format.

https://github.com/epoberezkin/ajv/blob/0020556493f3d443002596697768f3f047782198/spec/async_validate.spec.js#L379-L394

One more: https://runkit.com/esp/ajv-asynchronous-validation

docs/site/Parsing-requests.md
mastermunj: The key is meant to provide details for creating a new custom format for ajv. As of now, they are meant to be for ajv thus name `ajvCustomFormats`.

docs/site/Parsing-requests.md
```diff
@@ -234,6 +234,76 @@ class Product extends Entity {
...
+      'only-alphanumeric': /^[a-zA-Z0-9 ]+$/.source,
+      'product-code': {
+        async: true,
```
achrinza: I think asynchronous detection should be automated (Possible solution: https://stackoverflow.com/questions/38508420/how-to-know-if-a-function-is-async). Having a manual flag is room for human error.

docs/site/Parsing-requests.md
achrinza: @mastermunj I understand that they're meant for AJV, I just think we need to create an abstraction layer from AJV so that we have more control over the APIs exposed to the developers. Postponing the change would mean a breaking change (major semver).

docs/site/Parsing-requests.md
```diff
@@ -234,6 +234,76 @@ class Product extends Entity {
...
+
+AJV Custom Keywords can be configured by binding the value
+`validation.ajvCustomKeywords` to `RestBindings.REQUEST_BODY_PARSER_OPTIONS`
+('rest.requestBodyParserOptions').
+
+```ts
+server.bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS).to({
+  validation: {
+    ajvCustomFormats: {
+      'only-alphanumeric': /^[a-zA-Z0-9 ]+$/.source,
```
achrinza: As per-your earlier comment, AJV custom formats can be useful for using functions for validation. I think we should prevent strings or regexp from being used as that'd overlap with the use of the `pattern` key.

docs/site/Parsing-requests.md
achrinza: Ah noted;

docs/site/Parsing-requests.md
mastermunj: I understand your point, however, ajv is in a way tightly integrated into LB4 schema validation as of now. `ajvKeywords` & `ajvErrors` are examples of it.

I believe, we will need to look at the inclusion of ajv as a whole for achieving the abstraction layer.

achrinza: I agree that AJV is already tighly-integrated as per-your examples; However, the APIs exposed aren't prefixed with `ajv`. (e.g. `ajvErrors` is exposed as `jsonSchema.errorMessage`). I just think it should be named something else.

dougal83:
> Is it possible to run Travis manually?

@mastermunj Yes, maintainers can restart manually but others are restricted to submissions.

docs/site/Parsing-requests.md
mastermunj: An input to custom format could be string, regexp, function or an object altogether as explained at https://ajv.js.org/#api-addformat.

Although the async function can be detected, I believe doing it all will add unnecessary logic on framework side. ajv anyway does all these validations.

Plus, at times, too much automation might not be a good choice.

docs/site/Parsing-requests.md
mastermunj: I do not see the reason to restrict string or regexp. Let the programmer trying to implement custom format decide on best use either via string/regexp in custom format or creating own logic for reusing string/regexp.

achrinza: _If_ we were to go forward with the duplicate RegExp validation feature, I see 2 concerns:

1. Will `ajvCustomFormats` be reflected on the [OAS 3 Schema Object Property, `pattern`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#properties)?
2. If both a RegExp `format` and `pattern` are defined, which takes priority (for validation and OAS 3 `pattern` property)?

IMO while it's good to give developers options, duplicate features are not beneficial.

I think we need to more carefully-consider how we are exposing this API to the developers.

docs/site/Parsing-requests.md
achrinza: IMO it there isn't much code added to detect async, and it's a better tradeoff than unexpected behaviour due to a separate flag that was set incorrectly.

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -497,6 +529,212 @@ describe('Validation at REST level', () => {
...
+          $data: true,
+          ajvKeywords: true,
+          ajvErrors: {
+            singleError: true,
+          },
+          ajvCustomFormats: {
+            'only-alphanumeric': /^[a-zA-Z0-9 ]+$/.source,
+            'product-code': {
+              async: true,
+              validate: async (code: string) => {
```
deepakrkris: Is there a way to restrict asynchronous validation for now ? Any benefits or necessity of having to support both ajv custom formats and asynchronous validation at the same time ?

docs/site/Parsing-requests.md
mastermunj: Valid points.
1. Nope, `ajvCustomFormats` won't reflect on OAS 3 Schema property `pattern`.
2. I searched, however, I do not have an answer to which one takes priority between `format` and `pattern`. I will try running them in parallel with different regex to check which one is validated first.

`format` is a valid attribute of json schema validation draft though. http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7

packages/rest/src/types.ts
```diff
@@ -100,6 +100,14 @@ export type AjvErrorOptions = {
   singleError?: boolean;
 };
 
+export type AjvCustomFormat = {
+  [name: string]: ajv.FormatValidator | ajv.FormatDefinition;
```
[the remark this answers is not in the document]
mastermunj: The purpose is to support custom ajv formats, however, since ajv supports async validators, I made necessary changes to make it work.

Now I understand that it comes up as a breaking change.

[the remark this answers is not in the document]
mastermunj: During a more detailed search, I found an option `formats` that allows providing custom formats in a similar way I have implemented with `addCustomFormats`. Silly me to not notice it earlier.

Now the point remains on async nature. Since ajv allows custom formats with default options, and they can be async in nature as well, should those be allowed or not?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -142,13 +143,24 @@ function validateValueAgainstSchema(
     cache.set(schema, validatorMap);
   }
 
-  if (validate(body)) {
+  let validationResult = validate(body);
+  let validationErrors: AJV.ErrorObject[] = [];
+  if (isPromiseLike(validationResult)) {
```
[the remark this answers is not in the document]
mastermunj: With [new information](https://github.com/strongloop/loopback-next/pull/4658#discussion_r380657228) at hand, the discussion changes to two points:

1. To allow async validations or not?
2. To allow custom keywords to be created or not? <= This because the PR has provision for both formats & keywords since ajv has provision to allow custom formats at options level.

docs/site/Parsing-requests.md
achrinza: 1. If we were to allow RegExp `ajvCustomFormats`, it should update OAS 3 `pattern`.
2. As per-my previous comment, if we're going to implement this, I think we should restrict to function-based validation only to prevent this overlap. Taking a shot in the dark, both Regexp `ajvCustomFormats` and `pattern` will be used for validation.

`format` is valid and is being used by lb4 for types such as `Date` (`date-time`). In this case, there's the same concern of priority (of which should be documented).

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -55,6 +55,38 @@ describe('Validation at REST level', () => {
...
+  class ProductWithCode {
+    @property()
+    id: number;
+
+    @property({
+      required: true,
+      jsonSchema: {
+        format: 'only-alphanumeric',
+        errorMessage: {format: 'Name must not have special characters'},
+      },
```
achrinza: Internally, I understand why it's placed under `jsonSchema`. However, this seems a bit misleading as it's not representative of the JSON schemas defined in OAS 3. Though this may be out-of-scope of this PR.

docs/site/Parsing-requests.md
hacksparrow: Not very convinced by `async: true`. It feels redundant when we are already specifying the validation function is async.

packages/rest/src/types.ts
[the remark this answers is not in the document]
hacksparrow: Agreed.

docs/site/Parsing-requests.md
hacksparrow: > I think it should be named something else other than `ajvCustomFormats`

I share the same opinion. The interface of AJV should be abstracted as much as possible. The key should have a generic name, independent of the underlying library; there is no guarantee we will never switch the validation library.

bajtos:
Regarding naming & coupling with AJV - can we simply use `customFormats` and `customKeywords` as the config keys?

I have one high-level concern to consider: IMO, specifying all custom formats & keywords inside app config is cumbersome. In most cases, I expect LB users to extract common formats and/or common keywords into an extension that can be shared among multiple projects. We should find a way how to allow extensions to contribute such custom formats and keywords and provide necessary documentation for extension developers & app developers.

A possible approach that may not require any extra work at framework side:

```ts
// inside an extension
export const MyCustomValidators {
  formats: {
    'only-alphanumeric': /^[a-zA-Z0-9 ]+$/.source,
    'product-code': {
      async: true,
      validate: validateProductCode,
    },
  },
  keywords: {
    'product-blacklist': {
      async: true,
      compile: (schema, parentSchema, it) => {
        return async (data: string) => {
          const blacklistedSeries = ['ABC', 'PQR', 'XYZ'];
          const series = (data as string).substring(0, 3).toUpperCase();
          return Promise.resolve(blacklistedSeries.indexOf(series) < 0);
        };
      },
      errors: false,
    },
  },
};

// in an app
import {MyCustomValidators} from 'my-extension';

app
  .bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS)
  .to({
     validation: {
        customFormats: {
          ...MyCustomValidators.formats,
          // add your app-specific formats here
        },
        customKeywords: {
          ...MyCustomValidators.keywords,
          // add your app-specific formats here
        },
     },
  });  
```
  
const

bajtos:
An alternative approach is to use ExtensionPoint/Extension pattern:
- The validation implementation in the framework defines two extension points - ValidationFormat and ValidationKeyword.
- When configuring AJV, we need to find all formats/keywords from the context.
- Extensions can use `@bind(asValidationFormat)` and `@bind(asValidationKeyword)` to contribute new formats & keywords.

See https://loopback.io/doc/en/lb4/core-tutorial-part5.html and https://github.com/strongloop/loopback-next/tree/master/examples/log-extension for more details.

@raymondfeng do you have any opinions on the high-level architecture of this feature?

packages/rest/src/validation/request-body.validator.ts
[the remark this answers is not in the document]
bajtos: My opinions:

> To allow async validations or not?

I think allowing async validation is inevitable, let's make it happen now. We are going to publish a semver-major version of all packages in the next week or two because we have dropped support for Node.js 8.x, let's try to include the switch to async validation in that release if possible.

Either way, I think the switch to async validations deserves a standalone pull request. That way we can focus on the implications of switching to async validations and don't get distracted about formats/keywords.

> To allow custom keywords to be created or not? <= This because the PR has provision for both formats & keywords since ajv has provision to allow custom formats at options level.

I think we should allow app developers & extension developers to contribute both custom keywords and custom formats. I don't have a strong opinion on how that should be implemented, whether via standard AJV options we already support, or via additional validation config (as presented in this PR, IIUC) or via new extension-points (as I commented in https://github.com/strongloop/loopback-next/pull/4658#issuecomment-589672199).

mastermunj:
With [Release 6.12.0](https://github.com/epoberezkin/ajv/releases/tag/v6.12.0), AJV has allowed provision for adding `keywords` with `options` removing the need of using `addKeyword` method.

With that change, this PR is not needed anymore, however, I'd request if we can make validations async before the next major release.

raymondfeng:
@mastermunj Would you like to clean up the PR to add `async` validation support? It's also great to allow keywords/formats to be contributed by extensions as proposed by https://github.com/strongloop/loopback-next/pull/4658#issuecomment-589672199.

mastermunj:
@raymondfeng I have cleaned up the PR.

Request you to check [this](https://github.com/strongloop/loopback-next/pull/4658/files#diff-5b2d24820db7f9cff0ed5e93d171c73cR230) line of code in particular if it's the right approach or it can be done in a better way.

achrinza:
It seems to be a borked rebase against master which caused the upstream and feature branch commits to become intertwined:

image

At this stage, it might be best to overwrite all the commits with 1 commit:

(This is my suggestion, there may be other ways to achieve the same result)

1. Add remote `upstream`
```
git remote add upstream https://github.com/strongloop/loopback-next.git
```
2. Checkout `master`
```
git checkout master
```
3. Rebase against upstream
```
git pull --rebase upstream master
```
4. Force-push **with lease**
```
git push --force-with-lease
```
5. Checkout feature branch
```
git checkout custom-ajv-formats-keywords
```
6. Fetch all and hard reset to `origin/master`
```
git fetch --all
git reset --hard origin/master
```
7. Reapply **all** final changes manually
8. Add commit
```
git add --all
git commit -sm "feat(rest): add async validation support"
```
9. Force-push **with lease**
    **This will permanently overwrite git history. Any previous commits will be lost. See step 3 if the changes were not commited yet.**
```
git push --force-with-lease
```

Full rebase instructions can be found at [Submitting a PR#Rebasing](https://loopback.io/doc/en/lb4/submitting_a_pr.html).

mastermunj:
Thanks a lot, @achrinza. I will raise another PR in a while.

dougal83:
@achrinza Would the following not have done the trick on the branch?  Just curious how borked it was as cannot see now.

```
git fetch --all
git checkout custom-ajv-formats-keywords
git rebase upstream/master
git push --force-with-lease
```

Issue #4584, opened 2020-02-07
vscode source code formatting vs. linting rules?


derdeka:
**Questions**

When saving a typescript file in vscode it's getting automatically formated by some linting rules:

https://github.com/strongloop/loopback-next/blob/34ee060fee8757a1a3ff675355d6b899953c21cf/.vscode/settings.json#L10-L13

When saving it automatically gets formated, e.g. some spaces after brackets are inserted:

![image](https://user-images.githubusercontent.com/13640166/74026272-1e4d4280-49a6-11ea-83ac-0e7b1b98eaa8.png)
![image](https://user-images.githubusercontent.com/13640166/74026296-2c9b5e80-49a6-11ea-8845-37b7d73289af.png)

When running `npm run lint:fix` theese spaces are removed again to conform the .prettierrc or .eslintrc.js linting rules.

**Is there any way to let vscode and prettier apply the same linting rules?** Is this the expected behavior or is there something wrong with my local setup?

Windows 10 Version 1909 Enterprise
vscode 1.41.1
vscode ESLint Plugin 2.0.15

dougal83:
Add to settings:
`"typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,`

derdeka:
> Add to settings:
> `"typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,`

This setting works for me. Should we add this to .vscode/settings.json ?

> The above works as standard.

Do you mean this would be the standard if i disable all plugins?

dougal83:
> This setting works for me. Should we add this to .vscode/settings.json?

Wouldn't be a bad idea if you ask me.  Would you like to do the honours with PR?

> Do you mean this would be the standard if i disable all plugins?
I believe it would apply with vanilla VSCode install if you have a suitably recent version of Typescript in use.  Can't remember of the top of my head.

bajtos:
FYI: here is our recommended VSCode setup: https://github.com/strongloop/loopback-next/blob/master/docs/site/VSCODE.md

derdeka:
> FYI: here is our recommended VSCode setup: https://github.com/strongloop/loopback-next/blob/master/docs/site/VSCODE.md

Thanks @bajtos @dougal83, i forgot to install prettier plugin on my new laptop. Anyway, i think the .vscode setting might be helpful.

PR #4660, opened 2020-02-16
feat: .vscode add no spaces in braces + single quotes typescript rules

dougal83:
useful rule for when Prettier not enabled.

Implements #4584

## Checklist

### DIFF
--- a/packages/cli/generators/project/templates/.vscode/settings.json
+++ b/packages/cli/generators/project/templates/.vscode/settings.json
@@ -4,6 +4,10 @@
   "editor.tabSize": 2,
   "editor.trimAutoWhitespace": true,
   "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": true,
+    "source.fixAll.eslint": true
+  },
 
   "files.exclude": {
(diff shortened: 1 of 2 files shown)


dougal83:
> BTW how about single quotes (`'`) vs double quotes (`"`)? I vaguely remember that TypeScript may be using double quotes by default, should we change that config too?

Good shout.  Will add.

raymondfeng:
@dougal83 Can you apply the same change to https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/project/templates/.vscode/settings.json#L19 as well? This way, newly created LB4 projects will have the good default settings too.

packages/cli/generators/project/templates/.vscode/settings.json
```diff
@@ -4,6 +4,10 @@
   "editor.tabSize": 2,
   "editor.trimAutoWhitespace": true,
   "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": true,
```
dougal83: Also added `editor.codeActionsOnSave.source.organizeImports: true,` as nice to have.

PR #4679, opened 2020-02-18
feat(rest): allow controllers to be added or removed after server start

raymondfeng:
Implements #433 

## Checklist

### DIFF
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -378,7 +378,6 @@ export class RestServer extends Context implements Server, HttpServerLike {
       );
       for (const route of routes) {
         this.bindRoute(route);
-        this._httpHandler.registerRoute(route);
       }
     }
 
@@ -414,7 +413,9 @@ export class RestServer extends Context implements Server, HttpServerLike {
 
     const controllerName = spec['x-controller-name'];
     if (typeof controllerName === 'string') {
-      const b = this.find(`controllers.${controllerName}`)[0];
+      const b = this.getBinding(`controllers.${controllerName}`, {
+        optional: true,
+      });
       if (!b) {
         throw new Error(
           `Unknown controller ${controllerName} used by "${verb} ${path}"`,
--- a/packages/rest/src/rest.server.ts
+++ b/packages/rest/src/rest.server.ts
@@ -9,7 +9,11 @@ import {
   BindingScope,
   Constructor,
   Context,
+  ContextObserver,
+  filterByKey,
+  filterByTag,
   inject,
+  Subscription,
 } from '@loopback/context';
 import {Application, CoreBindings, Server} from '@loopback/core';
 import {HttpServer, HttpServerOptions} from '@loopback/http-server';
@@ -32,7 +36,7 @@ import {ServeStaticOptions} from 'serve-static';
 import {writeErrorToResponse} from 'strong-error-handler';
 import {BodyParser, REQUEST_BODY_PARSER_TAG} from './body-parsers';
 import {HttpHandler} from './http-handler';
-import {RestBindings} from './keys';
+import {RestBindings, RestTags} from './keys';
 import {RequestContext} from './request-context';
 import {
   ControllerClass,
@@ -146,6 +150,12 @@ export class RestServer extends Context implements Server, HttpServerLike {
     this._setupHandlerIfNeeded();
     return this._httpHandler;
   }
+
+  /**
+   * Context event subscriptions for route related changes
+   */
+  private _routesEventSubscription: Subscription;
+
   protected _httpServer: HttpServer | undefined;
 
   protected _expressApp: express.Application;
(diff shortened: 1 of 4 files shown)


packages/rest/src/rest.server.ts
```diff
@@ -338,20 +343,42 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
+          filterByKey(RestBindings.API_SPEC.key)(binding) ||
+          (filterByKey(/^(controllers|routes)\..+/)(binding) &&
+            // Exclude controller routes to avoid circular events
+            !filterByTag(RestTags.CONTROLLER_ROUTE)(binding)),
+      );
+    }
+    this._createHttpHandler();
+
+    // See https://github.com/strongloop/loopback-next/issues/433
+    this._controllersAndRoutesView.on('refresh', () => {
```
bajtos: From what I remember, ContextView is rather heavy-weight and may involve async operations. Have you considered using ContextListener instead? 

I am also not sure about the behavior of the `refresh` event, is it fired on the same tick of the event loop when the bindings were added/removed? I am concerned that if `ContextView` was delaying those events because to the next tick of event loop or Promise microqueue, then it would create a race condition where a request may arrive after we have registered the new controller but before the changes were picked up by the routing layer.

packages/rest/src/rest.server.ts
```diff
@@ -338,20 +343,42 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
      */
     const router = this.getSync(RestBindings.ROUTER, {optional: true});
     const routingTable = new RoutingTable(router, this._externalRoutes);
 
     this._httpHandler = new HttpHandler(this, this.config, routingTable);
-    for (const b of this.find('controllers.*')) {
+
+    // Remove controller routes
+    for (const b of this.findByTag(RestTags.CONTROLLER_ROUTE)) {
+      this.unbind(b.key);
```
derdeka: Does this mean all routes of all controllers are unregisterd and registerd again when a new controller gets added? What happens if requests arrive in between?

raymondfeng: Bindings for controller based routes have to be refreshed to reflect currently registered controllers. The recycling is synchronous. No request is handled by the event loop before it finishes.

packages/rest/src/rest.server.ts
raymondfeng: `ContextView` is notified to refresh after the binding is fully populated with `ctx.bind(...).tag(...)`. It will be executed in another tick.

When a new controller is added by `app.controller()`, the following steps will happen:

1. A new binding for the controller is added (sync)
2. A context `bind` event is emitted (sync) and a `notification` event is added to a queue
3. In another tick, the notification is processed to invoke ContextView's `observe()`, which calls `refresh()` and emits `refresh` event.

I think the race condition is acceptable as `app.controller` call does not expect the controller will be available in the same tick.

packages/rest/src/rest.server.ts
derdeka: Thanks for explanation.

packages/rest/src/rest.server.ts
```diff
@@ -650,11 +681,13 @@ export class RestServer extends Context implements Server, HttpServerLike {
     );
   }
 
-  private bindRoute(r: RouteEntry, namespace = 'routes') {
+  private bindRoute(r: RouteEntry) {
+    const namespace = RestBindings.ROUTES;
```
jannyHou: Does the change here mean we don't allow custom namespace anymore?

packages/rest/src/rest.server.ts
```diff
@@ -338,20 +343,42 @@ export class RestServer extends Context implements Server, HttpServerLike {
...
-    // instance whenever a controller was added/deleted.
-    // See https://github.com/strongloop/loopback-next/issues/433
     if (this._httpHandler) return;
+    if (this._controllersAndRoutesView == null) {
+      // Watch for binding events
+      this._controllersAndRoutesView = this.createView(
+        binding =>
+          filterByKey(RestBindings.API_SPEC.key)(binding) ||
+          (filterByKey(/^(controllers|routes)\..+/)(binding) &&
+            // Exclude controller routes to avoid circular events
```
jannyHou: Could you explain more about the `CONTROLLER_ROUTE`? I also feel confused for the test related to it. 
For example, when do we tag a binding with `REST_ROUTE` and when tag `CONTROLLER_ROUTE`? Do they overlap? And why `CONTROLLER_ROUTE` need to be excluded here?

packages/rest/src/rest.server.ts
raymondfeng: Oh, the PR that introduces `bindRoute` has not been released yet. So there is no breaking change.

packages/rest/src/rest.server.ts
raymondfeng: We discover controller bindings and add routes for each controller as a binding tagged with `CONTROLLER_ROUTE`. The `routes.*` also contain routes from non-controllers. This tag allows us to rebuild controller routes when a controller is added/removed.

packages/rest/src/rest.server.ts
bajtos: I believe your work on `ContextListener` included a way how to be notified about changes in the binding metadata?

Setting timing issues aside, I feel `ContextView` solution is too heavy weight. IIRC, `ContextView` is maintaining a list of all matched bindings, which unnecessarily increases memory consumption, considering that we are not using that list at all.

Anyhow, as long as ContextView vs. ContextListener is just an implementation detail that can be changed later, I am fine with landing the current approach.

raymondfeng: I switched to `ContextObserver`.

packages/rest/src/rest.server.ts
jannyHou: I see thanks for the explanation!

PR #4693, opened 2020-02-20
feat(Authentication) enable authentication strategies to contribute OASEnhancer

dougal83:
Extend AuthenticationStrategy to implement OASEnhancer.

Example usage:
App > jwt-strategy.ts:
```ts
// ...
@bind(asAuthStrategy, asSpecEnhancer)
export class JWTAuthenticationStrategy
  implements AuthenticationStrategy, OASEnhancer {
  // ...
  modifySpec(spec: OpenApiSpec): OpenApiSpec {
    return mergeSecuritySchemeToSpec(spec, this.name, {
      type: 'http',
      scheme: 'bearer',
      bearerFormat: 'JWT',
    });
  }
}
```
App > application.ts:
```ts
// ...
// authentication
this.add(createBindingFromClass(JWTAuthenticationStrategy));
```

See also #4554
Impl. #3669

## Checklist

### DIFF
--- a/packages/authentication/src/types.ts
+++ b/packages/authentication/src/types.ts
@@ -1,9 +1,15 @@
-// Copyright IBM Corp. 2018,2019. All Rights Reserved.
+// Copyright IBM Corp. 2018,2020. All Rights Reserved.
 // Node module: @loopback/authentication
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {addExtension, Constructor, Context} from '@loopback/core';
+import {
+  addExtension,
+  BindingTemplate,
+  Constructor,
+  Context,
+  extensionFor,
+} from '@loopback/core';
 import {Request} from '@loopback/rest';
 import {UserProfile} from '@loopback/security';
 import {AuthenticationBindings} from './keys';
@@ -93,6 +99,7 @@ export const USER_PROFILE_NOT_FOUND = 'USER_PROFILE_NOT_FOUND';
  * Registers an authentication strategy as an extension of the
  * AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME extension
  * point.
+ *
  * @param context - Context object
  * @param strategyClass - Class for the authentication strategy
  */
@@ -110,3 +117,16 @@ export function registerAuthenticationStrategy(
     },
   );
 }
+
+/**
+ * A binding template for auth strategy contributor extensions
+ */
+export const asAuthStrategy: BindingTemplate = binding => {
+  extensionFor(
+    AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  )(binding);
+  binding.tag({
+    namespace:
+      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
+  });
+};
--- a/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/basic-auth-extension.acceptance.ts
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2019. All Rights Reserved.
+// Copyright IBM Corp. 2019,2020. All Rights Reserved.
 // Node module: @loopback/authentication
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
@@ -9,7 +9,7 @@ import {anOpenApiSpec} from '@loopback/openapi-spec-builder';
 import {api, get} from '@loopback/openapi-v3';
 import {Request, RestServer} from '@loopback/rest';
 import {SecurityBindings, securityId, UserProfile} from '@loopback/security';
-import {Client, createClientForHandler} from '@loopback/testlab';
+import {Client, createClientForHandler, expect} from '@loopback/testlab';
 import {
   authenticate,
   AuthenticationBindings,
@@ -164,6 +164,7 @@ describe('Basic Authentication', () => {
         },
       });
   });
+
   it('returns error when undefined user profile returned from authentication strategy', async () => {
     class BadBasicStrategy implements AuthenticationStrategy {
       name = 'badbasic';
@@ -193,6 +194,22 @@ describe('Basic Authentication', () => {
         },
       });
   });
+
+  it('adds security scheme component to apiSpec', async () => {
+    const EXPECTED_SPEC = {
+      components: {
+        securitySchemes: {
+          basic: {
+            type: 'http',
+            scheme: 'basic',
+          },
+        },
+      },
+    };
+    const spec = await server.getApiSpec();
+    expect(spec).to.containDeep(EXPECTED_SPEC);
+  });
+
   async function givenAServer() {
     app = getApp();
     server = await app.getServer(RestServer);
--- a/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
+++ b/packages/authentication/src/__tests__/acceptance/jwt-auth-extension.acceptance.ts
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2019. All Rights Reserved.
+// Copyright IBM Corp. 2019,2020. All Rights Reserved.
 // Node module: @loopback/authentication
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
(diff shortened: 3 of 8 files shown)


dougal83:
@derdeka Would this satisfy #3669? PTAL

packages/authentication/src/types.ts
```diff
@@ -110,3 +139,28 @@ export function registerAuthenticationStrategy(
...
+
+/**
+ * Registers a security scheme spec as an extension of the @loopback/openapi-v3
+ * OAS_ENHANCER_EXTENSION_POINT_NAME extension point.
+ *
+ * @param context - Context object
+ * @param name - Prefix for security scheme spec enhancer name
+ * @param securitySchemes - OAS security scheme objects
+ */
+export function registerAuthSecuritySpec(
```
jannyHou: This register function looks reasonable to me 👍 
Next we could add a few acceptance test to verify if a security spec get merged when call `registerAuthenticationStrategy(this, strategyName);`

jannyHou: And I need to double check if we register strategy in the app constructor, does it happen before or after the rest server gets instantiated.

packages/authentication/src/types.ts
```diff
@@ -70,6 +84,12 @@ export interface AuthenticationStrategy {
    */
   name: string;
 
+  /**
+   * The 'spec' property allows a security schemes object to be provided for the
+   * strategy that will be registered as an OAS spec enhancer.
+   */
+  spec?: SecuritySchemeObjects;
```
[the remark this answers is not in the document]
dougal83: Updated.  Used `const strategy: AuthenticationStrategy = strategyClass.prototype;` to avoid instantiation.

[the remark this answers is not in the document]
dougal83: @jannyHou How would I go about adding a static function?  Using instantiation for the meantime.

dougal83:
Restricted to a single `SecuritySchemeObject` as I didn't see eventuality of adding multiple.

packages/authentication/src/types.ts
```diff
@@ -70,6 +83,12 @@ export interface AuthenticationStrategy {
    */
   name: string;
 
+  /**
+   * The 'securitySpec' property allows a security scheme to be provided for the
+   * strategy that will be registered as an OAS spec enhancer.
+   */
+  securitySpec?: () => SecurityScheme;
```
bajtos: I am a bit confused. The `name` property is defined and implemented as a prototype property (see e.g. `JWTAuthenticationStrategy#name`), but `securitySpec` is defined as a prototype property but implemented as a static property in your examples. As a result, the definition here is disconnected from the actual security classes.

Based on the implementation below, it seems that we need to access the security spec before the strategy class is instantiated. This makes sense to me, in that case I think we need to rework this property to be a static one.

Also, I don't understand why is this a getter function (`() => SecurityScheme`) and not a value property?

```suggestion
  securitySpec?: SecurityScheme;
```

packages/authentication/src/types.ts
```diff
@@ -92,14 +116,18 @@ export const USER_PROFILE_NOT_FOUND = 'USER_PROFILE_NOT_FOUND';
...
+ * point. Initiates regististration of security spec if provided.
+ *
  * @param context - Context object
  * @param strategyClass - Class for the authentication strategy
  */
 export function registerAuthenticationStrategy(
   context: Context,
   strategyClass: Constructor<AuthenticationStrategy>,
 ) {
+  const spec = strategyClass.prototype.constructor.securitySpec;
```
bajtos: Please note that `strategyClass.prototype` has type any, therefore everything after that is not type-checked by the compiler. Let's find a way how to preserve compiler checks for `securitySpec`.

First of all, I think we should access `securitySpec` directly via the class constructor:

```suggestion
  const spec = strategyClass.securitySpec;
```

Now this is obviously rejected by the compiler because `Constructor<>` does not have any `securitySpec` property.

```
ts(2339)
Property 'securitySpec' does not exist on type 'Constructor<AuthenticationStrategy>' 
```

A possible solution is to change the type of `strategyClass` to describe `securitySpec` argument.

```diff
-   strategyClass: Constructor<AuthenticationStrategy>,
+   strategyClass: Constructor<AuthenticationStrategy> & {securitySpec?: () => SecurityScheme; }
```

(This addresses the "static property" part of my previous comment.)

To make the "strategy class" type easier to use in other places, I suggest to create a type alias:

```ts
export type AuthenticationStrategyClass = Constructor<AuthenticationStrategy> & {
  securitySpec?: () => SecurityScheme; 
}
```

packages/authentication/src/types.ts
```diff
@@ -110,3 +138,30 @@ export function registerAuthenticationStrategy(
...
+ *
+ * @param context - Context object
+ * @param securityScheme - security scheme object
+ */
+export function registerAuthSecuritySpec(
+  context: Context,
+  securityScheme: SecurityScheme,
+) {
+  @bind(asSpecEnhancer)
+  class SecuritySpecEnhancer implements OASEnhancer {
```
bajtos: I have mixed feelings about creating a new `SecuritySpecEnhancer` class for each authentication strategy registered. Have you considered an alternative approach where each `AuthenticationStrategy` class can decide to implement `OASEnhancer` and register itself `asSpecEnhancer`?

IMO, such solution will require less code in the framework and offer more flexibility to strategy authors.

A mock-up:

```ts
import {
   asSpecEnhancer,
   // new helper API to be implemented
   mergeSecuritySchemeToSpec,
   OASEnhancer,
 } from '@loopback/openapi-v3';

@bind(
  asSpecEnhancer,
 // should we also add something like asAuthStrategy
 // to correctly set the binding tags needed to annotate
 // this class as an authentication strategy?
)
export class JWTAuthenticationStrategy implements AuthenticationStrategy, OASEnhancer {
  // ...

  modifySpec(spec: OpenApiSpec): OpenApiSpec {
    return mergeSecuritySchemeToSpec(spec, 'jwt', {
      type: 'http',
      scheme: 'bearer',
      bearerFormat: 'JWT',
    });
  }
}

// inside @loopback/openapi-v3'

export function mergeSecuritySchemeToSpec(spec: OpenApiSpec, schemeName: string, schemeSpec: SecurityScheme) {
  const patchSpec = {
    components: {
      securitySchemes: {[schemeName]: schemeSpec},
    },
  };

  const mergedSpec = mergeOpenAPISpec(spec, patchSpec);
  return mergedSpec;
}
```

Regardless of the approach we choose for integrating AuthenticationSchemes and OASEnhancer, I think we should always extract the helper `mergeSecuritySchemeToSpec` to `@loopback/openapi-v3`, because it's something that can be used in different places too. (Feel free to pick a better function name.)

packages/authentication/src/types.ts
dougal83: I was a bit confused but tried to apply some feedback rightly or wrongly.  Happy to revert to value property. 👍

packages/authentication/src/types.ts
dougal83: I had no idea you could simply append an argument description.  Something new learned, thank you! 🙇

packages/authentication/src/types.ts
dougal83: @bajtos Much better your way! 👌 I've created `asAuthStrategy` but couldn't figure out how to do something like `@bind(asSpecEnhancer, asAuthStrategy)`.  The documents seem to cover a single binding but not multiple.

Thank you for your guidance, it is most appreciated!

raymondfeng: The signature of `@bind` is `export function bind(...specs: BindingSpec[])`. It can take any number of binding template functions, such as `@bind(asSpecEnhancer, asAuthStrategy)`.

dougal83: I did try `@bind(asSpecEnhancer, asAuthStrategy)` but are there any considerations when adding to context?  I used only the following: `context.add(createBindingFromClass(strategyClass));` and found that only one of the bindings was happy depending on the order in the `@bind` decorator.  I ran out of time to poke at it, is it something simple I need to do?  Thank you for your help.

raymondfeng: See https://github.com/strongloop/loopback-next/blob/master/packages/context/src/__tests__/unit/binding-decorator.unit.ts#L96

raymondfeng: It seems that you hit the limitation of https://github.com/strongloop/loopback-next/blob/master/packages/core/src/extension-point.ts#L127. The `extensionFor` uses the same tag to point to an `extensionPoint` and that's why two templates override each other.

raymondfeng: What about https://github.com/strongloop/loopback-next/pull/4816?

packages/authentication/src/types.ts
jannyHou: I suggested a function or a static property in https://github.com/strongloop/loopback-next/pull/4693#discussion_r387795232 so @dougal83 tried it, but the new approach looks much better than my original proposal, which creates a separate class for enhancer.
Thanks @bajtos and @dougal83

packages/authentication/src/types.ts
raymondfeng: #4816 is now merged into master. Please rebase to pick it up.

dougal83:
The simplicity of the PR is down to great architectural design by the Loopback Team.  It was surprising how little work was required in the end!  Keep up the great work! ❤️

jannyHou:
🎉 🎉 Merged! Thank you for the effort and patience, next target is https://github.com/strongloop/loopback-next/pull/4365. (design kudos goes to Miroslav and Raymond :p)

Issue #3700, opened 2019-09-10
error 500 when using an array in post method.


rsa408:
Hi guys
Back to the issue [why Loopback + MongoDB is not working?](https://github.com/strongloop/loopback-next/issues/3645) another problem I found:

I removed exclude from schema:
```
async create(
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(Use_r, { exclude: ['id'] }_),
        },
      },
    })
    user: Omit<User, 'id'>,
  ): Promise<User> {
    return this.userRepository.create(user);
  }

```

then  I add `generated: true` and removed `required: true` from the user model . 
```
 @property({
    type: 'number',
    id: true,
    generated: true
  })
  id: number;
```

it is looking fine when I used POST method, for one record/document when in add several records it ( array of records )I get:
```
[{
		"email": "nunc.sed@semelit.net",
		"password": "571087741"
	},
	{
		"email": "lacus@dui.org",
		"password": "346470131"
	},
	{
		"email": "ipsum@aaliquetvel.org",
		"password": "243339637"
	}]
```
```
{
  "error": {
    "statusCode": 500,
    "message": "Internal Server Error"
  }
}
```
```
Unhandled error in POST /users: 500 TypeError: model.toObject is not a function
    at UserRepository.toEntity (D:\apps\test\node_modules\@loopback\repository\src\repositories\legacy-juggler-bridge.ts:471:39)
    at UserRepository.create (D:\apps\test\node_modules\@loopback\repository\src\repositories\legacy-juggler-bridge.ts:338:17)
```

But regardless of the error message, it was POSTed all records to MongoDB correctly data with autogenerated id.

dougal83:
Can't help without seeing the code, post an example repo.  This is not clear at all.

rsa408:
> Can't help without seeing the code, post an example repo. This is not clear at all.

please find the project here: [https://github.com/rsa408/lb4-test](https://github.com/rsa408/lb4-test)
also, [test data.json](https://github.com/rsa408/lb4-test/blob/master/test%20data.json) is there.

dougal83:
First thing, if you are generating id then you should exclude the id on creates request body.  What you have currently didn't make sense by allowing users to be submitted with id while generating is enabled. Also can you confirm that you've used ``npm run migrate`` to set up the db correctly?

[...]

PR #4701, opened 2020-02-21
DO NOT MERGE: feat(rest): reject array values in JSON requests

achrinza:
see #3700

Signed-off-by: Rifa Achrinza <25147899+achrinza@users.noreply.github.com>

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/body-parser.unit.ts
+++ b/packages/rest/src/__tests__/unit/body-parser.unit.ts
@@ -90,6 +90,25 @@ describe('body parser', () => {
     });
   });
 
+  it('rejects top-level arrays in application/json', async () => {
+    const req = givenRequest({
+      url: '/',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      payload: [],
+    });
+    const spec = givenOperationWithRequestBody({
+      description: 'data',
+      content: {},
+    });
+    return expect(
+      requestBodyParser.loadRequestBodyIfNeeded(spec, req),
+    ).to.be.rejectedWith(
+      'The request body is invalid. See error object `details` property for more info.',
+    );
+  });
+
   it('allows text/json to be parsed', async () => {
     const req = givenRequest({
       url: '/',
--- a/packages/rest/src/body-parsers/body-parser.json.ts
+++ b/packages/rest/src/body-parsers/body-parser.json.ts
@@ -5,17 +5,19 @@
 
 import {inject} from '@loopback/context';
 import {json} from 'body-parser';
+import _ from 'lodash';
 import {is} from 'type-is';
 import {RestBindings} from '../keys';
+import {sanitizeJsonParse} from '../parse-json';
+import {RestHttpErrors} from '../rest-http-error';
 import {Request, RequestBodyParserOptions} from '../types';
 import {
   BodyParserMiddleware,
+  builtinParsers,
   getParserOptions,
   invokeBodyParserMiddleware,
-  builtinParsers,
 } from './body-parser.helpers';
 import {BodyParser, RequestBody} from './types';
-import {sanitizeJsonParse} from '../parse-json';
 
 export class JsonBodyParser implements BodyParser {
   name = builtinParsers.json;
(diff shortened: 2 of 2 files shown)


packages/rest/src/body-parsers/body-parser.json.ts
```diff
@@ -36,6 +38,9 @@ export class JsonBodyParser implements BodyParser {
 
   async parse(request: Request): Promise<RequestBody> {
     let body = await invokeBodyParserMiddleware(this.jsonParser, request);
+
+    if (!_.isPlainObject(body)) throw RestHttpErrors.invalidRequestBody();
```
raymondfeng: Rejecting non-object body unconditionally is not desired. It should be perfect fine to expose an API to accept json array in the body, for example, an array of deals to be added.

raymondfeng:
@achrinza To fix #3700, we probably need to fix the root cause (toObject() is invoked on an array).

achrinza:
@raymondfeng Ah noted; `toObject()` is in `@loopback/repository`. In that case, how should we throw an exception?

bajtos:
Thank you @achrinza for the pull request!

> To fix #3700, we probably need to fix the root cause (toObject() is invoked on an array).

I disagree. IMO, if I decorate my controller method with OpenAPI spec saying that `data` argument allows only objects, then LoopBack's REST layer must ensure my controller method never receives array instances.

In https://github.com/strongloop/loopback-next/issues/3700, I proposed:

> Improve our REST validation to reject Array values for parameters accepting an object.

Similarly, we can reject object (non-array) values for parameters accepting an array.

> Ah noted; toObject() is in @loopback/repository. In that case, how should we throw an exception?

Yes, we should throw an exception with 4xx status code - the client sent us a bad request.

packages/rest/src/__tests__/unit/body-parser.unit.ts
```diff
@@ -90,6 +90,25 @@ describe('body parser', () => {
...
+    const req = givenRequest({
+      url: '/',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      payload: [],
+    });
+    const spec = givenOperationWithRequestBody({
+      description: 'data',
+      content: {},
```
bajtos: IMO it's crucial to provide a schema for the request body object, this schema must describe `type: object`.

As I interpret the currently proposed spec, it allows any values for the request body, therefore LoopBack's REST layer must not reject arrays.

packages/rest/src/body-parsers/body-parser.json.ts
bajtos: I think this is not the right place for rejecting array values for object-only bodies. What if the request body was sent as urlencoded formdata (or XML in the future when we add support for parsing XML request)? We also don't know what's the allowed body schema at this point.

I would expect this additional validation to happen inside [request-body.validator.ts](https://github.com/strongloop/loopback-next/blob/cf97cd9a9b8ea1ff7334de7ffb7bb870580486e2/packages/rest/src/validation/request-body.validator.ts).

Note that we already have logic to validate that non-body parameters are valid objects, see here:

https://github.com/strongloop/loopback-next/blob/cf97cd9a9b8ea1ff7334de7ffb7bb870580486e2/packages/rest/src/coercion/coerce-parameter.ts#L161-L174

Can we perhaps reuse this logic for request bodies too?

TBH, I am surprised that AJV is not handling this edge case for us and allows array values when the schema describes an object 🤔

raymondfeng:
> I disagree. IMO, if I decorate my controller method with OpenAPI spec saying that data argument allows only objects, then LoopBack's REST layer must ensure my controller method never receives array instances.

I was confused by JSON object/array vs. JavaScript Object/Array. For JSON schema validation, `object` type should not allow JSON array.

emonddr:
@achrinza , any update on this PR? Thank you :)

achrinza:
@emonddr Sorry for the delay; just a bit busy lately with the current events. I'm still working on it, though I'm fine if someone else wants to take over.

emonddr:
> @emonddr Sorry for the delay; just a bit busy lately with the current events. I'm still working on it, though I'm fine if someone else wants to take over.

Understandable, just trying to keep the discussion active :)

dhmlau:

```
git commit --amend -s
git push --force-with-lease
```

PR #4702, opened 2020-02-21
feat(repository): make `targetsMany` property required

bajtos:
Modify `RelationDefinitionBase` to require all relation definitions to provide value of `targetsMany` field.

**BREAKING CHANGE**

If you are building a custom relation type with its own definition interface, make sure the interface includes `targetsMany` property. Typically, the type of this property is hard-coded as `true` or `false`, depending on the relation type.

```ts
interface HasManyDefinition extends RelationDefinitionBase {
  type: RelationType.hasMany;
  targetsMany: true;
  // etc.
}

export interface BelongsToDefinition extends RelationDefinitionBase {
  type: RelationType.belongsTo;
  targetsMany: false;
  // etc.
}
```

When creating an instance of a relation definition, make sure to include a value for `targetsMany` property.

```ts
new ModelDefinition('Order')
  .addRelation({
    name: 'customer',
    type: RelationType.belongsTo,
    targetsMany: false,
    source: Order,
    target: () => Customer,
    keyFrom: 'customerId',
    keyTo: 'id',
  });
```

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
+++ b/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
@@ -217,6 +217,7 @@ class Order extends Entity {
     .addRelation({
       name: 'customer',
       type: RelationType.belongsTo,
+      targetsMany: false,
       source: Order,
       target: () => Customer,
       keyFrom: 'customerId',
--- a/packages/repository/src/__tests__/unit/errors/invalid-relation-error.test.ts
+++ b/packages/repository/src/__tests__/unit/errors/invalid-relation-error.test.ts
@@ -48,6 +48,7 @@ function givenAnErrorInstance() {
   return new InvalidRelationError('a reason', {
     name: 'products',
     type: RelationType.hasMany,
+    targetsMany: true,
     source: Category,
     target: () => Product,
   });
--- a/packages/repository/src/relations/relation.types.ts
+++ b/packages/repository/src/relations/relation.types.ts
@@ -24,14 +24,12 @@ export interface RelationDefinitionBase {
    */
   type: RelationType;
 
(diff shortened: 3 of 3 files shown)


dhmlau:
Just curious.. why do we need to make `targetsMany` mandatory? Would we consider that if `targetsMany` is not specified, the default will be true (or false)?

bajtos:
> Is `targetsMany` only for OpenAPI/JSON Schema generator? 

Yes, we use it to determine how to describe navigational property in the response of "find + include related models'. 

See here:

https://github.com/strongloop/loopback-next/blob/d30825803678a08df076f654e90a4fa40ba67f57/packages/repository-json-schema/src/build-schema.ts#L264-L282

> Will the related docs be updated? Sorry this is new to me :see_no_evil:

The flag `targetsMany` is for developers building new relation types (e.g. `ReferencesMany`), it's not used by LB4 app developers. 

I did a text search in `docs` and did not find any documentation file mentioning that string. IMO, there is no documentation to update.

> Just curious.. why do we need to make `targetsMany` mandatory? Would we consider that if `targetsMany` is not specified, the default will be true (or false)?

The problem is that this flag does not have any good default value. Some relations are targeting multiple instances (`hasMany`, in the future `referencesMany`, etc.), while other relations are targeting a single instance (`belongsTo`, `hasOne`, etc.).

By making the flag required, when somebody is implementing a new relation type, we are forcing them to think about this aspect and tell the framework what kind of a relation it is.

If the flag provided a default value, then it would be too easy to leave the default value and then be surprised about wrong JSON/OpenAPI schema emitted for `find` API. 

At the moment, the flag is optional, so that we can throw an error at runtime if the relation developer forgets to set it.

With my change in place, the problem can be flagged at compile time.

PR #4382, opened 2020-01-07
feat: implement lint-staged

frbuceta:
> lint-staged - Run linters on git staged files

Implementation of lint-staged throughout the project to upload the code without errors

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- `npm test` passes on your machine
- New tests added or existing tests modified to cover all changes
- Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- API Documentation in code was updated
- Documentation in [/docs/site](../tree/master/docs/site) was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4707, opened 2020-02-21
Remove dist files from top-level tsconfig to speed up eslint checks 🏎🚀

bajtos:
This speeds up `npm run eslint` from about `4m10s` down to `2m50s`.

More importantly, it enables proper caching behavior, so that subsequent runs of `npm run eslint` are super quick, even after `npm run build` modified `dist` files.

```
$ npm run build
(...)
$ time npm run eslint

> loopback-next@0.1.0 eslint /Users/bajtos/src/loopback/next
> node packages/build/bin/run-eslint --report-unused-disable-directives --cache .


real	0m1.626s
user	0m1.143s
sys	0m0.393s
```

(That's less than 2 seconds for subsequent eslint runs leveraging the cached data!)

See the discussion in https://github.com/strongloop/loopback-next/pull/4382 for more context.

## Checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,7 @@
     "examples/*/node_modules/**",
     "packages/*/node_modules/**",
     "extensions/*/node_modules/**",
+    "**/dist/**",
     "**/*.d.ts"
   ]
 }


tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "examples/*/node_modules/**",
     "packages/*/node_modules/**",
     "extensions/*/node_modules/**",
```
derdeka:
```suggestion
    "**/node_modules/**",
```
BTW, any idea why this is not `**/node_modules/**,` ?

bajtos: IDK, to be honest. I was thinking about simplifying the pattern too, but then decided to keep my changes minimal to get them landed faster.

Would you like to open a follow-up pull request to simplify this part of `exclude` configuration?

raymondfeng:
@bajtos Interesting. Does it hint that the patterns in https://github.com/strongloop/loopback-next/blob/master/.eslintignore have some issues?

bajtos:
> Does it hint that the patterns in https://github.com/strongloop/loopback-next/blob/master/.eslintignore have some issues?

As far as I understand from reading the discussion in https://github.com/typescript-eslint/typescript-eslint/issues/389, `.eslintignore` is applied by `eslint` only. `typescript-eslint` adds TypeScript to the mix, at which point `tsconfig.json` is consulted independently of `.eslintignore`. So it's important to exclude build artifacts in both files.

Quoting from https://github.com/typescript-eslint/typescript-eslint/issues/389#issuecomment-545849837:

> if you came here because your tslint/typescript setup is slow, I recommend you to try the following:
> 
> 1. Setup an `.eslintignore` file to ignore node_modules as well as build / non-typescript files
> 2. Do the same thing for `tsconfig.json` via the "exclude" property to exclude node_modules or non typescript files.

raymondfeng:
Should we fix https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/project/templates/tsconfig.json.ejs#L33-L36 too?

bajtos:
> @bajtos Interesting. Does it hint that the patterns in https://github.com/strongloop/loopback-next/blob/master/.eslintignore have some issues?

To actually answer your question, I believe the patterns in `.eslintignore` are fine, nothing to fix there.

raymondfeng:
Did you remove `.eslintcache` file before `npm run eslint` for the benchmark?

bajtos:
> Did you remove `.eslintcache` file before `npm run eslint` for the benchmark?

No. First I run `npm run lint` to update the cache, that took several minutes (more or less the same time as before my changes).  Then I run `npm run build` and measured how long it takes to run `npm run lint`.

bajtos:
I am going to land this pull request as it is now, since it has been approved by many reviewers and there are no objections ("request changes" reviews).

Things we can fix in follow-up pull requests:
-  fix https://github.com/strongloop/loopback-next/blob/master/packages/cli/generators/project/templates/tsconfig.json.ejs#L33-L36 too?
- use pattern `**/node_modules/**`  - see https://github.com/strongloop/loopback-next/pull/4707#discussion_r382656257

Issue #3951, opened 2019-10-17
How to migrate model mixins


bajtos:
_This is a follow-up for #3718 and #3922._

Write content for `docs/site/migration/models/mixins.md`, explain how to migrate mixins from LB3 style (see [Mixins docs](https://loopback.io/doc/en/lb3/Defining-mixins.html)) to LB4 style (ES2015 mixin classes).

Important: In LB3, a model class provided both Schema definition, Persistence behavior and Public API. As a result, a single LB3 mixin can contribute to all three layers. In LB4, each of these layers map to a different concept (Model class, Repository class, Controller class). 

## Acceptance criteria

Migration guide describing how to:

- [ ] Convert a LB3 "mixin function" to LB4 mixin class factory
- [ ] Migrate mixins contributing to model schema (e.g. new property definitions)
- [ ] Migrate mixins modifying persistence behavior (e.g. adding new methods, installing operation hooks, etc.)
- [ ] Migrate mixins modifying API (e.g. adding new public methods, modifying remoting metadata of built-in LB3 methods).
- [ ] Migrate mixin configuration in LB3 model definition to LB4 Model and/or Repository.

UPDATED 2020-02-27 by @bajtos: Based on discussion with @emonddr, we decided to removed the following entry, because "mixin model" approach is not documented in LB3:
- Convert a LB3 "mixin model" to LB4 mixin class factory

bajtos:
I think this story may need a spike to explore different LB3 mixin kinds and what's the best way migration path for them.

You can use the following npm search to find existing LB3 mixins as examples to help us better understand different usage patterns we need to cover in our migration guide:

https://www.npmjs.com/search?q=loopback%20mixin

emonddr:
As suggested by @bajtos , we should have a LoopBack 4 topic in our documentation on how to use mixins. https://github.com/strongloop/loopback-next/issues/4759

PR #4723, opened 2020-02-24
docs: migrating lb3 model property mixins, and custom model method/remote method mixins to lb4

emonddr:
Related to : https://github.com/strongloop/loopback-next/issues/3951

## Checklist

### DIFF
--- a/docs/site/migration/models/mixins.md
+++ b/docs/site/migration/models/mixins.md
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
 permalink: /doc/en/lb4/migration-models-mixins.html
 ---
 
-{% include note.html content="
-This is a placeholder page, the task of adding content is tracked by the
-following GitHub issue:
-[loopback-next#3951](https://github.com/strongloop/loopback-next/issues/3951)
-" %}
+## Introduction
+
+This document will guide you in migrating custom model mixins, and custom
+method/remote method mixins in LoopBack 3 to their equivalent implementations in
+LoopBack 4.
+
+For an understanding of how models in LoopBack 3 are now architectually
+decoupled into 3 classes (model, repository, and controller) please read
+[Migrating custom model methods](./methods.md).
+
+In LoopBack 3, it was easy to add property mixins and method
+[mixins](https://loopback.io/doc/en/lb3/Defining-mixins.html).
+
+In LoopBack 4, it is also easy and is accomplished by using a mixin class
+factory function.
+
+## Creating a Property Mixin
+
+This section covers the approach that LoopBack 3 and LoopBack 4 can use to add a
+property to a model via a mixin.
+
+### LoopBack 3 Approach
+
+In LoopBack 3, a developer is able to create a model property mixin by:
+
+- placing the mixin logic in a file in a mixins directory
+- updating the server/model-config.json file with the mixin directory location
+- updating the model's json file to include the mixin's name and a boolean
+
+As an example, we will create a mixin that adds a **category** property to a
+model.
+
+#### Defining The Model Property Mixin category.js
+
+The developer defines a model property mixin in **common/mixins/category.js**
+which adds a required property named `category` to any model.
+
+{% include code-caption.html content="common/mixins/category.js" %}
+
+```js
+module.exports = function(Model, options) {
+  Model.defineProperty('category', {type: 'string', required: true});
+};
+```
+
+#### Updating model-config.json
+
+The **server/model-config.json** needs to contain:
+
+- the locations of all **models**
+- the location of all **mixins**
+- the entry of the model that receives the mixin content (for this example
+  `Note`)
+
+{% include code-caption.html content="server/model-config.json" %}
+
+```
+{
+  "_meta": {
+    "sources": [
+      "loopback/common/models",
+      "loopback/server/models",
+      "../common/models",
+      "./models"
+    ],
+    "mixins": [
+      "loopback/common/mixins",
+      "loopback/server/mixins",
+      "../common/mixins",
+      "./mixins"
+    ]
+  },
+
+ // ... other entries
+
+  "Note": {
+    "dataSource": "db"
+  }
+}
+
+```
+
+Please see
+[Reference mixins in model-config.js](https://loopback.io/doc/en/lb3/Defining-mixins.html#reference-mixins-in-model-configjs)
+for a short explanation of this file.
+
+#### Applying The category.js Mixin To A Model
+
+To extend the model `Note` with the **category.js** mixin, we need to add a
+**mixins** section in **common/models/note.json** to indicate which mixins
+should be applied to it.
+
+{% include code-caption.html content="common/models/note.json" %}
+
+```json
+{
+  "name": "Note",
+  "properties": {
+    "title": {
+      "type": "string",
+      "required": true
+    },
+    "content": {
+      "type": "string"
+    }
+  },
+  "mixins": {
+    "Category": true
+  }
+}
+```
+
+Specifying a value of **true** for `Category` will apply the **category.js**
+property model mixin to the `Note` model. A value of **false** will not apply
+the mixin.
+
+### LoopBack 4 Approach
+
+In LoopBack 4, a developer is able to create a model property mixin by:
+
+- creating a base model class which extends `Entity`
+- placing the mixin class factory function in a separate file
+- generating a model using the CLI as usual
+- adjusting the model file to make use of the mixin class factory function
+
+#### Defining A BaseEntity Class Which Extends Entity
+
+Let's define a base model class `BaseEntity` in **src/models/base-entity.ts**.
+It will be used as input to the mixin later.
+
+{% include code-caption.html content="src/models/base-entity.ts" %}
+
+```ts
+import {Entity} from '@loopback/repository';
+export class BaseEntity extends Entity {}
+```
+
+This is necessary because the
+[Entity](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L276)
+class is abstract and doesn't have a constructor.
+
+#### Defining The Model Property Mixin Class Factory Function
+
+This mixin class factory function `AddCategoryPropertyMixin` in
+**src/mixins/category-property-mixin.ts** adds the required property
+**category** to any model.
+
+{% include code-caption.html content="src/mixins/category-property-mixin.ts" %}
+
+```ts
+import {Constructor} from '@loopback/context';
+import {property, Model} from '@loopback/repository';
+
+/**
+ * A mixin factory to add `category` property
+ *
+ * @param superClass - Base Class
+ * @typeParam T - Model class
+ */
+export function AddCategoryPropertyMixin<T extends Constructor<Model>>(
+  superClass: T,
+) {
+  class MixedModel extends superClass {
+    @property({
+      type: 'string',
+      required: true,
+    })
+    category: string;
+  }
+  return MixedModel;
+}
+```
+
+{% include note.html content="At the moment, [TypeScript does not allow decorators in class expressions](https://github.com/microsoft/TypeScript/issues/7342). This is why we need to declare the class with a name, and then return it." %}
+
+#### Generating A Model Via The CLI
+
+A CLI-generated model named `Note` with 3 properties: **id**, **title**, and
+**content** would look like this:
+
+{% include code-caption.html content="src/models/note.model.ts" %}
+
+```ts
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class Note extends Entity {
+  @property({
+    type: 'number',
+    id: true,
+    generated: true,
+  })
+  id?: number;
+
+  @property({
+    type: 'string',
+    required: true,
+  })
+  title: string;
+
+  @property({
+    type: 'string',
+  })
+  content?: string;
+
+  constructor(data?: Partial<Note>) {
+    super(data);
+  }
+}
+
+export interface NoteRelations {
+  // describe navigational properties here
+}
+
+export type NoteWithRelations = Note & NoteRelations;
+```
+
+#### Adjusting The Model File To Use AddCategoryPropertyMixin
+
+The model file only requires a few adjustments:
+
+- import the `BaseEntity` class
+- import the `AddCategoryPropertyMixin` mixin
+- Change the class declaration of `Note` so that it extends the class returned
+  from the mixin function which takes in the `BaseEntity` superclass as input
+
+{% include code-caption.html content="src/models/note.model.ts" %}
+
+```ts
+import {model, property} from '@loopback/repository';
+import {AddCategoryPropertyMixin} from '../mixins/category-property-mixin';
+import {BaseEntity} from './base-entity';
+
+@model()
+export class Note extends AddCategoryPropertyMixin(BaseEntity) {
+  @property({
+    type: 'number',
+    id: true,
+    generated: true,
+  })
+  id?: number;
+
+  @property({
+    type: 'string',
+    required: true,
+  })
+  title: string;
+
+  @property({
+    type: 'string',
+  })
+  content?: string;
+
+  constructor(data?: Partial<Note>) {
+    super(data);
+  }
+}
+
+export interface NoteRelations {
+  // describe navigational properties here
+}
+
+export type NoteWithRelations = Note & NoteRelations;
+```
+
+The required property `category` has now been added to the `Note` model via a
+mixin class factory function.
+
+## Creating A Custom Model Method And Remote Model Method Mixin
+
+This section covers the approach that LoopBack 3 can use to add a custom method
+/remote method to a model via a mixin, and similarly how LoopBack 4 can add a
+custom method to a repository and controller via a mixin.
+
+### LoopBack 3 Approach
+
+The
+[Add a New Model Method And a New Endpoint](./methods.md#add-a-new-model-method-and-a-new-endpoint)
+section of the [Migrating custom model methods](./methods.md) document explains
+how a LoopBack 3 developer can define a custom model method named `findByTitle`
+on the `Note` model, and define a remote method to make it available as a new
+endpoint.
+
+In this section, we will show how a LoopBack 3 developer can define a mixin to
+accomplish this.
+
+In LoopBack 3, a developer is able to create a custom model method/remote method
+mixin by:
+
+- placing the mixin logic in a file in a mixins directory
+- updating the server/model-config.json file with the mixin directory location
+- updating the model's json file to include the mixin's name (and options object
+  or boolean)
+
+#### Defining The Model Method Mixin findByTitle.js
+
+The developer defines a custom model method/remote method mixin in
+**common/mixins/findByTitle.js** which adds a custom method `findByTitle` to any
+model, and adds a corresponding remote method definition with path
+`/findByTitle` as well. An options property `returnArgumentName` makes it
+possible to customize the name of the return argument. If it is not specified,
+the return argument of 'items' is used as a default.
+
+{% include code-caption.html content="common/mixins/findByTitle.js" %}
+
+```js
+module.exports = function(Model, options) {
+  const returnArgumentName = options.returnArgumentName
+    ? options.returnArgumentName
+    : 'items';
+
+  Model.remoteMethod('findByTitle', {
+    http: {
+      path: '/findByTitle',
+      verb: 'get',
+    },
+    accepts: {arg: 'title', type: 'string'},
+    returns: {arg: returnArgumentName, type: [Model], root: true},
+  });
+
+  Model.findByTitle = function(title, cb) {
+    var titleFilter = {
+      where: {
+        title: title,
+      },
+    };
+    Model.find(titleFilter, cb);
+  };
+};
+```
+
+For a model named `Note`, this will expose an endpoint of `/Notes/findByTitle`.
+
+Ensure **model-config.json** is set up properly as specified earlier in
+[Updating model-config.json](#updating-model-config.json)
+
+#### Applying The findByTitle.js Mixin To A Model
+
+To extend the model `Note` with the **findByTitle.js** mixin, we need to add a
+**mixins** section in **common/models/note.json** to indicate which mixins
+should be applied to it.
+
+{% include code-caption.html content="common/models/note.json" %}
+
+```json
+{
+  "name": "Note",
+  "properties": {
+    "title": {
+      "type": "string",
+      "required": true
+    },
+    "content": {
+      "type": "string"
+    }
+  },
+
+  "mixins": {
+    "FindByTitle": {
+      "returnArgumentName": "notes"
+    },
+
+    "Category": true
+  }
+}
+```
+
+Specifying an options object for `FindByTitle` is the same as specifying a value
+of **true** as it will apply the **findByTitle.js** custom model method/remote
+method mixin to the `Note` model. A value of **false** will not apply the mixin.
+
+### LoopBack 4 Approach
+
+As mentioned in the previous section, the
+[Add a New Model Method And a New Endpoint](./methods.md#add-a-new-model-method-and-a-new-endpoint)
+section of the [Migrating custom model methods](./methods.md) document explains
+how a LoopBack 3 developer can define a custom model method named `findByTitle`
+on the `Note` model, and define a remote method to make it available as a new
+endpoint. It then shows how a LoopBack 4 developer can implement a `findByTitle`
+method on the `NoteRepository` and on the `NoteController` to accomplish the
+same thing.
+
+In this section, we will show how a LoopBack 4 developer can define two mixins (
+a repository mixin and a controller mixin) to add a `findByTitle` method to
+`NoteRepository` and `NoteController` respectively.
+
+In LoopBack 4, a developer is able to create a repository and controller method
+mixin by:
+
+- defining a common interface for both mixin class factory functions
+- placing the mixin class factory functions in separate files
+- generating a repository and controller using the CLI as usual
+- adjusting the repository and controller files to make use of its respective
+  mixin class factory function
+
+#### Defining A Common Interface For The findByTitle Method
+
+Let's define a common interface `FindByTitle` in
+**src/mixins/find-by-title-interface.ts**.
+
+{% include code-caption.html content="src/mixins/find-by-title-interface.ts" %}
+
+```ts
+import {Model} from '@loopback/repository';
+
+/**
+ * An interface to allow finding notes by title
+ */
+export interface FindByTitle<M extends Model> {
+  findByTitle(title: string): Promise<M[]>;
+}
+```
+
+#### Defining A Repository Mixin Class Factory Function
+
+In **src/mixins/find-by-title-repository-mixin.ts**, let's define the mixin
+class factory function `FindByTitleRepositoryMixin` which adds the `findByTitle`
+method to any repository.
+
+{% include code-caption.html content="src/mixins/find-by-title-repository-mixin.ts" %}
+
+```ts
+import {Constructor} from '@loopback/context';
+import {Model, CrudRepository, Where} from '@loopback/repository';
+import {FindByTitle} from './find-by-title-interface';
+
+/*
+ * This function adds a new method 'findByTitle' to a repository class
+ * where 'M' is a model which extends Model
+ *
+ * @param superClass - Base class
+ *
+ * @typeParam M - Model class which extends Model
+ * @typeParam R - Repository class
+ */
+export function FindByTitleRepositoryMixin<
+  M extends Model & {title: string},
+  R extends Constructor<CrudRepository<M>>
+>(superClass: R) {
+  class MixedRepository extends superClass implements FindByTitle<M> {
+    async findByTitle(title: string): Promise<M[]> {
+      const where = {title} as Where<M>;
+      const titleFilter = {where};
+      return this.find(titleFilter);
+    }
+  }
+  return MixedRepository;
+}
+```
+
+#### Generating A Repository Via The CLI
+
+A CLI-generated repository for a model `Note` would look like this:
+
+{% include code-caption.html content="src/repositories/note.repository.ts" %}
+
+```ts
+export class NoteRepository extends DefaultCrudRepository<
+  Note,
+  typeof Note.prototype.id,
+  NoteRelations
+> {
+  constructor(@inject('datasources.db') dataSource: DbDataSource) {
+    super(Note, dataSource);
+  }
+}
+```
+
+#### Adjusting NoteRepository To Use FindByTitleRepositoryMixin
+
+The repository file only requires a few adjustments:
+
+- import the `FindByTitleRepositoryMixin` mixin class factory function
+- adjust the declaration of the `NoteRepository` class to extend the class
+  returned from the mixin function which takes in the `DefaultCrudRepository`
+  superclass as input.
+
+{% include code-caption.html content="src/repositories/note.repository.ts" %}
+
+```ts
+import {FindByTitleRepositoryMixin} from '../mixins/find-by-title-repository-mixin';
+import {DefaultCrudRepository} from '@loopback/repository';
+import {Note, NoteRelations} from '../models';
+import {DbDataSource} from '../datasources';
+import {inject, Constructor} from '@loopback/core';
+
+/**
+ * A repository for `Note` with `findByTitle`
+ */
+export class NoteRepository extends FindByTitleRepositoryMixin<
+  Note,
+  Constructor<
+    DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>
+  >
+>(DefaultCrudRepository) {
+  constructor(@inject('datasources.db') dataSource: DbDataSource) {
+    super(Note, dataSource);
+  }
+}
+```
+
+We have now added the `findByTitle` method to a repository via a mixin class
+factory function.
+
+#### Defining A Controller Mixin Class Factory Function
+
+In **src/mixins/find-by-title-controller-mixin.ts**, let's define the mixin
+class factory function `FindByTitleControllerMixin` which adds the `findByTitle`
+method to any controller.
+
+{% include code-caption.html content="src/mixins/src/mixins/find-by-title-controller-mixin.ts" %}
+
+```ts
+import {Constructor} from '@loopback/context';
+import {Model} from '@loopback/repository';
+import {FindByTitle} from './find-by-title-interface';
+import {param, get, getModelSchemaRef} from '@loopback/rest';
+
+/**
+ * Options to mix in findByTitle
+ */
+export interface FindByTitleControllerMixinOptions {
+  /**
+   * Base path for the controller
+   */
+  basePath: string;
+  /**
+   * Model class for CRUD
+   */
+  modelClass: typeof Model;
+}
+
+/**
+ * A mixin factory for controllers to be extended by `FindByTitle`
+ * @param superClass - Base class
+ * @param options - Options for the controller
+ *
+ * @typeParam M - Model class
+ * @typeParam T - Base class
+ */
+export function FindByTitleControllerMixin<
+  M extends Model,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends Constructor<any> = Constructor<object>
+>(superClass: T, options: FindByTitleControllerMixinOptions) {
+  class MixedController extends superClass implements FindByTitle<M> {
+    @get(`${options.basePath}/findByTitle/{title}`, {
+      responses: {
+        '200': {
+          description: `Array of ${options.modelClass.modelName} model instances`,
+          content: {
+            'application/json': {
+              schema: {
+                type: 'array',
+                items: getModelSchemaRef(options.modelClass, {
+                  includeRelations: true,
+                }),
+              },
+            },
+          },
+        },
+      },
+    })
+    async findByTitle(@param.path.string('title') title: string): Promise<M[]> {
+      return this.repository.findByTitle(title);
+    }
+  }
+
+  return MixedController;
+}
+```
+
+To customize certain portions of the OpenAPI description of the endpoint, the
+mixin class factory function needs to accept some options. We defined an
+interface `FindByTitleControllerMixinOptions` to allow for this.
+
+It is also a good idea to give the injected repository (in the controller super
+class) a generic name like `this.respository` to keep things simple in the mixin
+class factory function.
+
+#### Generating A Controller Via The CLI
+
+A CLI-generated **controller** for the model `Note` would look like this: (To
+save space, only a **partial** implementation is shown)
+
+{% include code-caption.html content="src/controllers/note.controller.ts" %}
+
+```ts
+export class NoteController {
+  constructor(
+    @repository(NoteRepository)
+    public noteRepository: NoteRepository,
+  ) {}
+
+  @post('/notes', {
+    responses: {
+      '200': {
+        description: 'Note model instance',
+        content: {'application/json': {schema: getModelSchemaRef(Note)}},
+      },
+    },
+  })
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(Note, {
+            title: 'NewNote',
+            exclude: ['id'],
+          }),
+        },
+      },
+    })
+    note: Omit<Note, 'id'>,
+  ): Promise<Note> {
+    return this.noteRepository.create(note);
+  }
+
+  // ...
+  // remaining CRUD endpoints
+  // ...
+}
+```
+
+For a full example of a CLI-generated controller for a model `Todo`, see
+[TodoController ](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/controllers/todo.controller.ts).
+
+#### Adjusting NoteController To Use FindByTitleControllerMixin
+
+The controller file only requires a few adjustments:
+
+- import the `FindByTitleControllerMixinOptions` interface
+- import the `FindByTitleControllerMixin` mixin class factory function
+- prepare the options for the mixin
+- adjust the declaration of the `NoteController` class to extend the class
+  returned from the mixin function which takes in the `Object` superclass as
+  input.
+- pass the input options into the mixin
+- change the name of the injected repository from `noteRepository` to
+  `repository` to keep things simple for the mixin class factory function
+
+{% include code-caption.html content="src/controllers/note.controller.ts" %}
+
+```ts
+import {Note} from '../models';
+import {
+  FindByTitleControllerMixin,
+  FindByTitleControllerMixinOptions,
+} from '../mixins/find-by-title-controller-mixin';
+import {Constructor} from '@loopback/core';
+
+import {
+  Count,
+  CountSchema,
+  Filter,
+  repository,
+  Where,
+} from '@loopback/repository';
+import {
+  post,
+  param,
+  get,
+  getFilterSchemaFor,
+  getModelSchemaRef,
+  getWhereSchemaFor,
+  patch,
+  put,
+  del,
+  requestBody,
+} from '@loopback/rest';
+import {NoteRepository} from '../repositories';
+
+const options: FindByTitleControllerMixinOptions = {
+  basePath: '/notes',
+  modelClass: Note,
+};
+
+export class NoteController extends FindByTitleControllerMixin<
+  Note,
+  Constructor<Object>
+>(Object, options) {
+  constructor(
+    @repository(NoteRepository)
+    public repository: NoteRepository,
+  ) {
+    super();
+  }
+
+  @post('/notes', {
+    responses: {
+      '200': {
+        description: 'Note model instance',
+        content: {'application/json': {schema: getModelSchemaRef(Note)}},
+      },
+    },
+  })
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(Note, {
+            title: 'NewNote',
+            exclude: ['id'],
+          }),
+        },
+      },
+    })
+    note: Omit<Note, 'id'>,
+  ): Promise<Note> {
+    return this.repository.create(note);
+  }
+
+  // ...
+  // remaining CRUD endpoints
+  // ...
+}
+```
+
+We have now added the `findByTitle` method to a controller via a mixin class
+factory function.
+
+This will also expose an endpoint of `/notes/findByTitle/{title}`.
+
+## Summary
+
+As the examples above show, migrating mixins from LoopBack 3 to LoopBack 4 is
+relatively straightforward using class factory functions.
--- a/packages/repository/src/__tests__/fixtures/mixins/category-property-mixin.ts
+++ b/packages/repository/src/__tests__/fixtures/mixins/category-property-mixin.ts
@@ -1,0 +1,26 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {Model, property} from '../../..';
+
+/**
+ * A mixin factory to add `category` property
+ *
+ * @param superClass - Base Class
+ * @typeParam T - Model class
+ */
+export function AddCategoryPropertyMixin<T extends Constructor<Model>>(
+  superClass: T,
+) {
+  class MixedModel extends superClass {
+    @property({
+      type: 'string',
+      required: true,
+    })
+    category: string;
+  }
+  return MixedModel;
+}
--- a/packages/repository/src/__tests__/fixtures/mixins/find-by-title-repo-mixin.ts
+++ b/packages/repository/src/__tests__/fixtures/mixins/find-by-title-repo-mixin.ts
@@ -1,0 +1,37 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {CrudRepository, Model, Where} from '../../..';
+
+/**
+ * An interface to allow finding notes by title
+ */
+export interface FindByTitle<M extends Model> {
+  findByTitle(title: string): Promise<M[]>;
+}
+
+/*
+ * This function adds a new method 'findByTitle' to a repository class
+ * where 'M' is a model which extends Model
+ *
+ * @param superClass - Base class
+ *
+ * @typeParam M - Model class which extends Model
+ * @typeParam R - Repository class
+ */
+export function FindByTitleRepositoryMixin<
+  M extends Model & {title: string},
+  R extends Constructor<CrudRepository<M>>
+>(superClass: R) {
+  class MixedRepository extends superClass implements FindByTitle<M> {
+    async findByTitle(title: string): Promise<M[]> {
+      const where = {title} as Where<M>;
+      const titleFilter = {where};
+      return this.find(titleFilter);
+    }
+  }
+  return MixedRepository;
+}
--- a/packages/repository/src/__tests__/fixtures/models/note.model.ts
+++ b/packages/repository/src/__tests__/fixtures/models/note.model.ts
@@ -1,0 +1,39 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {model, property} from '../../../';
+import {BaseEntity} from '../mixins/base-entity';
+import {AddCategoryPropertyMixin} from '../mixins/category-property-mixin';
+
+@model()
+export class Note extends AddCategoryPropertyMixin(BaseEntity) {
+  @property({
+    type: 'number',
+    id: true,
+    generated: true,
+  })
+  id?: number;
+
+  @property({
+    type: 'string',
+    required: true,
+  })
+  title: string;
+
+  @property({
+    type: 'string',
+  })
+  content?: string;
+
+  constructor(data?: Partial<Note>) {
+    super(data);
+  }
+}
+
+export interface NoteRelations {
+  // describe navigational properties here
+}
+
+export type NoteWithRelations = Note & NoteRelations;
--- a/packages/repository/src/__tests__/fixtures/repositories/note.repository.ts
+++ b/packages/repository/src/__tests__/fixtures/repositories/note.repository.ts
@@ -1,0 +1,27 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/core';
+import {DefaultCrudRepository, juggler} from '../../..';
+import {FindByTitleRepositoryMixin} from '../mixins/find-by-title-repo-mixin';
+import {Note, NoteRelations} from '../models/note.model';
+
+/**
+ * A repository for `Note` with `findByTitle`
+ */
+export class NoteRepository extends FindByTitleRepositoryMixin<
+  Note,
+  Constructor<
+    DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>
+  >
+>(DefaultCrudRepository) {
+  constructor(
+    dataSource: juggler.DataSource = new juggler.DataSource({
+      connector: 'memory',
+    }),
+  ) {
+    super(Note, dataSource);
+  }
+}
--- a/packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
+++ b/packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
@@ -1,0 +1,44 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {Note} from '../../fixtures/models/note.model';
+import {NoteRepository} from '../../fixtures/repositories/note.repository';
+
+describe('add method to crud repository via mixin', () => {
+  let repo: NoteRepository;
+  const noteData = {
+    title: 'groceries',
+    content: 'eggs,bacon',
+    category: 'keto',
+  };
+  let note: Note;
+
+  beforeEach(async () => {
+    repo = new NoteRepository();
+  });
+
+  // method from CrudRepository
+  it(`non-mixin method 'create' exists`, async () => {
+    note = await repo.create(new Note(noteData));
+    expect(note.toJSON()).to.deepEqual({id: 1, ...noteData});
+  });
+
+  // method from EntityCrudRepository
+  it(`non-mixin method 'findById' exists`, async () => {
+    note = await repo.create(new Note(noteData));
+    const foundNote: Note = await repo.findById(note.id);
+    expect(foundNote).to.deepEqual(note);
+  });
+
+  // method from mixin
+  it(`mixin method 'findByTitle' exists`, async () => {
+    note = await repo.create(
+      new Note({title: 'groceries', content: 'eggs,bacon', category: 'keto'}),
+    );
+    const foundNotes: Note[] = await repo.findByTitle('groceries');
+    expect(foundNotes).to.deepEqual([note]);
+  });
+});
--- a/packages/repository/src/__tests__/unit/mixins/model.mixin.unit.ts
+++ b/packages/repository/src/__tests__/unit/mixins/model.mixin.unit.ts
@@ -1,0 +1,47 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {MetadataInspector, MetadataMap} from '@loopback/context';
+import {expect} from '@loopback/testlab';
+import {MODEL_PROPERTIES_KEY} from '../../../';
+import {Note} from '../../fixtures/models/note.model';
+
+describe('add property to model via mixin', () => {
+  let meta: MetadataMap<unknown>;
+
+  before(async () => {
+    meta =
+      MetadataInspector.getAllPropertyMetadata(
+        MODEL_PROPERTIES_KEY,
+        Note.prototype,
+      ) ?? /* istanbul ignore next */ {};
+  });
+
+  it('metadata for non-mixin properties exist', () => {
+    expect(!!meta).to.be.True();
+    expect(meta.id).to.eql({
+      type: 'number',
+      id: true,
+      generated: true,
+      useDefaultIdType: false,
+    });
+
+    expect(meta.title).to.eql({
+      type: 'string',
+      required: true,
+    });
+
+    expect(meta.content).to.eql({
+      type: 'string',
+    });
+  });
+
+  it(`metadata for mixin property 'category' exists`, () => {
+    expect(meta.category).to.eql({
+      type: 'string',
+      required: true,
+    });
+  });
+});
--- a/packages/rest/src/__tests__/fixtures/controllers/note.controller.ts
+++ b/packages/rest/src/__tests__/fixtures/controllers/note.controller.ts
@@ -1,0 +1,193 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor, inject} from '@loopback/core';
+import {Count, CountSchema, Filter, Where} from '@loopback/repository';
+import {
+  del,
+  get,
+  getFilterSchemaFor,
+  getModelSchemaRef,
+  getWhereSchemaFor,
+  param,
+  patch,
+  post,
+  put,
+  requestBody,
+} from '../../../';
+import {
+  FindByTitleControllerMixin,
+  FindByTitleControllerMixinOptions,
+} from '../mixins/find-by-title-controller-mixin';
+import {Note} from '../models/note.model';
+import {
+  NoteRepository,
+  NOTE_REPO_BINDING_KEY,
+} from '../repositories/note.repository';
+
+const options: FindByTitleControllerMixinOptions = {
+  basePath: '/notes',
+  modelClass: Note,
+};
+
+export class NoteController extends FindByTitleControllerMixin<
+  Note,
+  Constructor<Object>
+>(Object, options) {
+  constructor(
+    @inject(NOTE_REPO_BINDING_KEY)
+    public repository: NoteRepository = new NoteRepository(),
+  ) {
+    super();
+  }
+
+  @post('/notes', {
+    responses: {
+      '200': {
+        description: 'Note model instance',
+        content: {'application/json': {schema: getModelSchemaRef(Note)}},
+      },
+    },
+  })
+  async create(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(Note, {
+            title: 'NewNote',
+            exclude: ['id'],
+          }),
+        },
+      },
+    })
+    note: Omit<Note, 'id'>,
+  ): Promise<Note> {
+    return this.repository.create(note);
+  }
+
+  @get('/notes/count', {
+    responses: {
+      '200': {
+        description: 'Note model count',
+        content: {'application/json': {schema: CountSchema}},
+      },
+    },
+  })
+  async count(
+    @param.query.object('where', getWhereSchemaFor(Note)) where?: Where<Note>,
+  ): Promise<Count> {
+    return this.repository.count(where);
+  }
+
+  @get('/notes', {
+    responses: {
+      '200': {
+        description: 'Array of Note model instances',
+        content: {
+          'application/json': {
+            schema: {
+              type: 'array',
+              items: getModelSchemaRef(Note, {includeRelations: true}),
+            },
+          },
+        },
+      },
+    },
+  })
+  async find(
+    @param.query.object('filter', getFilterSchemaFor(Note))
+    filter?: Filter<Note>,
+  ): Promise<Note[]> {
+    return this.repository.find(filter);
+  }
+
+  @patch('/notes', {
+    responses: {
+      '200': {
+        description: 'Note PATCH success count',
+        content: {'application/json': {schema: CountSchema}},
+      },
+    },
+  })
+  async updateAll(
+    @requestBody({
+      content: {
+        'application/json': {
+          schema: getModelSchemaRef(Note, {partial: true}),
+        },
+      },
+    })
+    note: Note,
+    @param.query.object('where', getWhereSchemaFor(Note)) where?: Where<Note>,
+  ): Promise<Count> {
+    return this.repository.updateAll(note, where);
+  }
+
+  @get('/notes/{id}', {
+    responses: {
+      '200': {
+        description: 'Note model instance',
+        content: {
+          'application/json': {
+            schema: getModelSchemaRef(Note, {includeRelations: true}),
+          },
+        },
+      },
+    },
+  })
+  async findById(
+    @param.path.number('id') id: number,
+    @param.query.object('filter', getFilterSchemaFor(Note))
+    filter?: Filter<Note>,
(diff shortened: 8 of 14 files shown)


raymondfeng:
@emonddr Maybe we should create an example under `examples/migration-lb3-mixin` to host your code?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+  })
+  content?: string;
+
+  constructor(data?: Partial<TempNote>) {
+    super(data);
+  }
+}
+
+@model()
+export class Note extends addCategoryPropertyMixin(TempNote) {
```
derdeka:
```suggestion
export class Note extends addCategoryPropertyMixin(Entity) {
```
Can you shortly explain why there is a need to rename and use `TempNote` instead of using a fully implemented `Note` extending `Entity` which was exhanced by the mixin?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+}
+
+export type NoteWithRelations = Note & NoteRelations;
+```
+
+### Adjusting The Model File To Use addCategoryPropertyMixin
+
+Before we use the mixin class factory function:
+
+- rename this class to `TempNote`
```
bajtos: The name `TempNote` suggests to me that this is just a temporary thing that will be removed later.  That's not true, is it?

Let's use a more descriptive class name than `TempNote` please, for example `BaseNote`.

docs/site/migration/models/mixins.md
bajtos: +1, I wanted to suggest the same. Take a look at how we are using mixins for Application classes:

https://github.com/strongloop/loopback-next/blob/28ebb8b7276577494d398423104d124b80241d43/examples/todo/src/application.ts#L15-L17

Ideally, I'd like use to describe the same approach for model mixins too. Here is an illustration of what I mean:

```ts
@model()
export class Note extends addCategoryPropertyMixin(Entity) {
  @property({
    type: 'number',
    id: true,
    generated: true,
  })
  id?: number;

  // etc.
}
```

Unless there are reasons why that does not work and the mixin must be applied on the class containing other model properties?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+to any model.
+
+{% include code-caption.html content="src/mixins/categoryPropertyMixin.ts" %}
+
+```ts
+import {Constructor} from '@loopback/context';
+
+import {property} from '@loopback/repository';
+
+export function addCategoryPropertyMixin<T extends Constructor<any>>(
```
bajtos: We usually use PascalCase naming for mixins applied on class constructors, see e.g. https://github.com/strongloop/loopback-next/blob/28ebb8b7276577494d398423104d124b80241d43/examples/todo/src/application.ts#L15-L17

```suggestion
export function AddCategoryPropertyMixin<T extends Constructor<any>>(
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+
+In LoopBack 3, a developer is able to create a custom model method/remote method
+mixin by:
+
+- placing the mixin logic in a file in a mixins directory
+- updating the server/model-config.json file with the mixin directory location
+- updating the model's json file to include the mixin's name (and options object
+  or boolean)
+
+### Defining The Model Property Mixin findByTitle.js
```
bajtos: I think this should say "Model Method Mixin"?

```suggestion
### Defining The Model Method Mixin findByTitle.js
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+    returns: {arg: returnArgumentName, type: [Model], root: true},
+  });
+
+  Model.findByTitle = function(title, cb) {
+    var titleFilter = {
+      where: {
+        title: title,
+      },
+    };
+    Model.find(titleFilter, function(err, foundItems) {
```
bajtos: Since your callback is just forwarding the results to `cb`, there is no need to write it and you can pass `cb` directly to `find`.

```suggestion
    Model.find(titleFilter, cb);
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+For a model named `Note`, this will expose an endpoint of `/Notes/findByTitle`.
+
+### Updating model-config.json
+
+The **server/model-config.json** needs to contain:
+
+- the locations of all **models**
+- the location of all **mixins**
+- the entry of the model that receives the mixin content (for this example Note)
+
```
bajtos: I think these instructions are the same as for Model Property mixins, can we remove them to avoid duplication and point readers to the content for Model Property mixins?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+In LoopBack 4, it is also easy and is accomplished by using a mixin class
+factory function.
+
+## LoopBack 3 Approach To Creating A Model Property Mixin
+
+In LoopBack 3, a developer is able to create a model property mixin by:
+
+- placing the mixin logic in a file in a mixins directory
+- updating the server/model-config.json file with the mixin directory location
+- updating the model's json file to include the mixin's name and a boolean
```
bajtos: I love ❤️ these short list of steps that you are showing in different places of the document, they make it easier to grasp the overal structure of this guide 👍

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+
+{% include code-caption.html content="src/mixins/findByTitleRepositoryMixin.ts" %}
+
+```ts
+import {Constructor} from '@loopback/context';
+
+import {Model} from '@loopback/repository';
+import {FindByTitleInterface} from './findByTitleInterface';
+
+export function FindByTitleRepositoryMixin<
```
bajtos: Kudos for using PascalCase for the mixin name :+1:

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+  M extends Model,
+  T extends Constructor<any>
```
bajtos: Is it possible to clarify these generic arguments to make it more clear that T is expected to be a class implementing `CrudRepository<M>`? 

Ideally, this should be expressed in a way that allows the TypeScript compiler to verify the contract at compile time. If that's not possible or too difficult, then at least add a short comment to explain what is `T` for. 

I also think that `R` would be a better name (`R` as a shortcut for `Repository`).

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+{% include code-caption.html content="src/repositories/note.repository.ts" %}
+
+```ts
+import {FindByTitleRepositoryMixin} from '../mixins/findByTitleRepositoryMixin';
+import {DefaultCrudRepository} from '@loopback/repository';
+import {Note, NoteRelations} from '../models';
+import {DbDataSource} from '../datasources';
+import {inject, Constructor} from '@loopback/core';
+
+export class NoteRepository extends FindByTitleRepositoryMixin<
```
bajtos: Kudos for using the pattern `class MyClass extends MyMixin(BaseClass)` 👍

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+export interface FindByTitleControllerMixinParms {
+  basePath: string;
+  modelClass: Constructor<{}>;
+  modelClassName: string;
+}
+
+export function FindByTitleControllerMixin<
+  M extends Model,
+  T extends Constructor<any>
+>(superClass: T, parms: FindByTitleControllerMixinParms) {
```
bajtos: I believe we usually abbreviate `parameters` as `params` (i.e. `params` argument, `FindByTitleControllerMixinParams` interface).

Having said that, we tend to use the name `options` instead of `params`.

```suggestion
>(superClass: T, options: FindByTitleControllerMixinOptions) {
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+  class MixedController extends superClass implements FindByTitleInterface<M> {
+    constructor(...args: any[]) {
+      super(...args);
+    }
+
+    @get(parms.basePath + '/findByTitle/{title}', {
+      responses: {
+        '200': {
+          description: `Array of ${parms.modelClassName} model instances`,
```
bajtos: I believe `modelClassName` argument is redundant, you can get the model name via `parms.modelClass.modelName`.

https://github.com/strongloop/loopback-next/blob/abc4c07863060921abf965af3f29718c0caab0b8/packages/repository/src/model.ts#L203-L206

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+              },
+            },
+          },
+        },
+      },
+    })
+    async findByTitle(@param.path.string('title') title: string): Promise<[M]> {
+      let foundItems = await this.repository.findByTitle(title);
+
+      return foundItems;
```
bajtos: Nitpick: the method body can be simplified as follows:

```ts
class MixedController  /*...*/ {
  
  @get(parms.basePath + '/findByTitle/{title}', {
    // ...
  })
  findByTitle(@param.path.string('title') title: string): Promise<[M]> {
    return this.repository.findByTitle(title);
  }
}
```

This is not only less lines of code, but also slightly more performant, because it creates one less Promise object at runtime.

See https://bajtos.net/2018-AsyncAwait/#/9/6 and [promise rewrapping](https://exploringjs.com/es2016-es2017/ch_async-functions.html#_returned-promises-are-not-wrapped)

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+
+- remove the class' `export` directive
+- rename this class to `TempNoteController`
+- change the name of the injected repository from `noteRepository` to
+  `repository` to keep things simple for the mixin class factory function
+
+{% include code-caption.html content="src/controllers/note.controller.ts" %}
+
+```ts
+class TempNoteController {
```
bajtos: Same comments as above:

1. Don't use `Temp` in the name, find a better prefix/postfix please. E.g. `NoteControllerBase` or `BaseNoteController`.
2. Even better, don't create any temporary classes, apply the mixin on the base class of the controller. In this particular case, I think you should use `Object` as the base.

```ts
class NoteController extends extends FindByTitleControllerMixin<
   Note,
   Constructor<TempNoteController>
 >(Object, mixinParms) {}
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+  T extends Constructor<any>
+>(superClass: T) {
+  return class extends superClass implements FindByTitleInterface<M> {
+    async findByTitle(title: string): Promise<M[]> {
+      const titleFilter = {
+        where: {
+          title: title,
+        },
+      };
+      const foundItems = await this.find(titleFilter);
```
jannyHou: a question...if the mixin doesn't have a function `find`, how could it call `find` here?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+    super(Note, dataSource);
+  }
+}
+```
+
+### Adjust NoteRepository To Use FindByTitleRepositoryMixin
+
+Import the `FindByTitleRepositoryMixin` mixin class factory function, and adjust
+the declaration of the `NoteRepository` class to extend the class returned from
+the mixin function that takes in the `DefaultCrudRepository` superclass an
```
jannyHou: nitpick: an --> as?

docs/site/migration/models/mixins.md
emonddr: The mixin requires something with a constructor. And `Entity` is an abstract class.

```
export abstract class Entity extends Model implements Persistable 
```
https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L276

emonddr: hmm... the only way I was able to get it working was by doing this:
```
import {addCategoryPropertyMixin} from '../mixins/categoryPropertyMixin';

class BaseClass extends Entity {};

@model()
export class Note extends addCategoryPropertyMixin(BaseClass) {
  constructor(data?: Partial<Note>) {
    super(data);
  }

  @property({
    type: 'number',
    id: true,
    generated: true,
  })
  id?: number;

  @property({
    type: 'string',
    required: true,
  })
  title: string;

  @property({
    type: 'string',
  })
  content?: string;

}


export interface NoteRelations {
}

export type NoteWithRelations = Note & NoteRelations;

```

emonddr: @raymondfeng suggested to me that we can make Entity a normal class (non-abstract)

docs/site/migration/models/mixins.md
emonddr: > The name TempNote suggests to me that this is just a temporary thing that will be removed  later. That's not true, is it? Let's use a more descriptive class name than TempNote please, for example BaseNote. 

No longer using TempNote

docs/site/migration/models/mixins.md
emonddr: Hmmm. It doesn't work

image

emonddr: @bajtos

docs/site/migration/models/mixins.md
emonddr: @jannyHou `find` is a function available on all models in lb3.

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+    "title": {
+      "type": "string",
+      "required": true
+    },
+    "content": {
+      "type": "string"
+    }
+  },
+  "mixins": {
+    "Category": true
```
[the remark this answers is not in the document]
emonddr: Not sure why, but when entering mixin information, things are in uppercase on first letter.
As in 👍 
https://github.com/strongloop/loopback-example-mixins/blob/master/common/models/dog.json#L26

```
  "mixins": {
    "TimeStamp": true,
    "Tag": true,
    "Squirrel": true
  }
```
all the mixin files are in lower case, but when specifying the info in  dog.json, they must be with first letter uppercase. 

@bajtos , @raymondfeng  , any ideas why this is?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+The **server/model-config.json** needs to contain:
+
+- the locations of all **models**
+- the location of all **mixins**
+- the entry of the model that receives the mixin content (for this example
+  `Note`)
+
+{% include code-caption.html content="server/model-config.json" %}
+
+```
```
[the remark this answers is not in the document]
emonddr: I can't because when I do that, it makes my comment

```
// ... other entries
```
all red and weird

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,805 @@ sidebar: lb4_sidebar
...
+{
+  "_meta": {
+    "sources": [
```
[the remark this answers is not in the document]
emonddr: This is a typical setup for mode-config.json as explained in 👍 
https://loopback.io/doc/en/lb3/Defining-mixins.html#reference-mixins-in-model-configjs

I will make a reference to this for an explanation.

emonddr:
I still need to refactor the note, and controller classes as suggested above, but made most suggested changes in 2nd commit.

docs/site/migration/models/mixins.md
derdeka: > @raymondfeng suggested to me that we can make Entity a normal class (non-abstract). Then I don't need to create BaseClass at all. Just pass in Entity.

+1, but I guess this will introduce some side effects.

I just remembered that i stumbled on this a year ago when using model mixins in my applications. I also introduced a `BaseModel` but with common attributes for all my models.

There is a similar issue when applying mixins on `ValueObject`.

emonddr: @derdeka , so you are ok with 

```
class BaseModel extends Entity {};

...

export class Note extends addCategoryPropertyMixin(BaseModel)

```

approach

and no longer ok with 

```
export class Note extends addCategoryPropertyMixin(Entity)
```
if it means making `Entity` non-abstract?

What kind of side-effects on Entity did this cause? I will run regression tests surrounding it to make sure there are no regressions (or side effects), but if you could elaborate, that would be helpful. thx.

docs/site/migration/models/mixins.md
emonddr: @bajtos , I did look at how Mixins were used in lb4, and at the moment, they only center around the main application. :)

docs/site/migration/models/mixins.md
derdeka: Both approaches are fine to me.

docs/site/migration/models/mixins.md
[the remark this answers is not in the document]
jannyHou: Oh you make a good point, JSON doesn't support comment. 👍 never mind, I am ok with either, w/ or w/o the type
https://stackoverflow.com/questions/244777/can-comments-be-used-in-json

docs/site/migration/models/mixins.md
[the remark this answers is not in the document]
jannyHou: I see then I guess the loopback-boot has some name mapping happen behind the scene.

docs/site/migration/models/mixins.md
jannyHou: oh I didn't mean lb3, I mean, typescript doesn't complain sth like `find` not found?

emonddr: no transpilation errors.

emonddr: @raymondfeng , any idea why there is no complaint of `this.find()`?

I know this code works:

```
export function FindByTitleRepositoryMixin<
  M extends Model,
  R extends Constructor<any>
>(superClass: R) {
  return class extends superClass implements FindByTitleInterface<M> {
    async findByTitle(title: string): Promise<M[]> {
      const titleFilter = {
        where: {
          title: title,
        },
      };
      return await this.find(titleFilter);
    }
  };
}
```
But just curious why ts doesn't complain.

emonddr:
Running test suite to ensure no regressions by making Entity class non-abstract....

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,699 @@ sidebar: lb4_sidebar
...
+import {FindByTitleRepositoryMixin} from '../mixins/findByTitleRepositoryMixin';
+import {DefaultCrudRepository} from '@loopback/repository';
+import {Note, NoteRelations} from '../models';
+import {DbDataSource} from '../datasources';
+import {inject, Constructor} from '@loopback/core';
+
+export class NoteRepository extends FindByTitleRepositoryMixin<
+  Note,
+  Constructor<
+    DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>
```
raymondfeng: Can it be as simple as `Constructor<Note>` or `typeof Note`?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,699 @@ sidebar: lb4_sidebar
...
+
+```ts
+import {Constructor} from '@loopback/context';
+import {Model} from '@loopback/repository';
+import {FindByTitleInterface} from './findByTitleInterface';
+import {param, get, getModelSchemaRef} from '@loopback/rest';
+
+export interface FindByTitleControllerMixinOptions {
+  basePath: string;
+  modelClass: Constructor<{}>;
```
raymondfeng: `Constructor<object>`?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,699 @@ sidebar: lb4_sidebar
...
+export function FindByTitleControllerMixin<
+  M extends Model,
+  T extends Constructor<any>
+>(superClass: T, options: FindByTitleControllerMixinOptions) {
+  class MixedController extends superClass implements FindByTitleInterface<M> {
+    constructor(...args: any[]) {
+      super(...args);
+    }
+
+    @get(options.basePath + '/findByTitle/{title}', {
```
raymondfeng: It's better to use:
```ts
@get(`${options.basePath}/findByTitle/{title}`, {
```

packages/repository/src/model.ts
```diff
@@ -273,7 +273,7 @@ export abstract class ValueObject extends Model implements Persistable {}
 /**
  * Base class for entities which have unique ids
  */
-export abstract class Entity extends Model implements Persistable {
+export class Entity extends Model implements Persistable {
```
raymondfeng: Can you have a separate PR for this? Is it still required with `BaseModel`?

docs/site/migration/models/mixins.md
emonddr: > Can it be as simple as `Constructor<Note>` or `typeof Note`?

Can you give an example of what you mean? thx. :)

packages/repository/src/model.ts
emonddr: > Can you have a separate PR for this? Is it still required with `BaseModel`?

I just snuck the commit in there so I can run the CI. I ran the tests locally but they keep dying on http proxy and example.com .  This change is not currently part of my proposed solution. I use BaseModel extends Entity {} right now.

docs/site/migration/models/mixins.md
raymondfeng: To replace `DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>`

packages/repository/src/model.ts
raymondfeng: If so, please revert the change.

emonddr: Reverted the change. No longer exploring the idea of making `Entity` non-abstract.

docs/site/migration/models/mixins.md
emonddr: @raymondfeng , I cannot get this to work in the suggested manner.

docs/site/migration/models/mixins.md
bajtos: I wish there was a runnable code I could use to try this out myself and tinker with the types.

I think the problem is caused by the fact your `modelClass` argument is specified as `Constructor<object>`, which means any class can be provided as the value and therefore we cannot guarantee that `modelName` static property is available.

IMO, this is a bug, `FindByTitleControllerMixin` should allow only Model constructors. Here is the type definition that should (hopefully) work:

```ts
export interface FindByTitleControllerMixinOptions {
  basePath: string;
  modelClass: typeof Model;
}
```

docs/site/migration/models/mixins.md
bajtos: Because `R` is defined as `Constructor<any>`, your super class is effectively extending from `any`. If you hover your mouse over `this.find` in VS Code, you will see that it has `any` type.

Screen Shot 2020-02-27 at 14 34 46

Here is a more correct definition of `FindByTitleRepositoryMixin`:

```ts
export function FindByTitleRepositoryMixin<
  M extends Model,
  R extends Constructor<CrudRepository<M>>
>(superClass: R) {
  return class extends superClass implements FindByTitleInterface<M> {
    // ...
  };
}
```

I am choosing `CrudRepository` because that's the interface where `find` is defined. By using that interface instead of e.g. `DefaultCrudRepository`, we make it possible to use this mixin with other repository implementations too.

bajtos: The problem with the code above is that TS now complains that `title` cannot be used in a filter, because a generic `Model` does not necessarily have a filter property. The solution is to limit `M` to `extends Model & {title: string}`.

```ts
export function FindByTitleRepositoryMixin<
  M extends Model & {title: string},
  R extends Constructor<CrudRepository<M>>
>(superClass: R) {
```

Unfortunately that's still not enough, because our `Condition<MT>` type cannot correctly handle the case where `MT` is a generic argument extending `Model & {title: string}`. I am proposing to use the following implementation:

```ts
    async findByTitle(title: string): Promise<M[]> {
      const where = {title} as Where<M>;
      const titleFilter = {where};
      return this.find(titleFilter);
    }
```

/cc @raymondfeng would you like to take a look on why `{title}` cannot be assigned to `Condition<M>` where `M` is a generic argument defined as `M extends Model & {title: string}`?

Example code to reproduce:

```ts
function test<M extends Model & {title: string}>(): Condition<M> {
  return {title: 'foo'};
}
```

Compiler error:

```
Type '{ title: "foo"; }' is not assignable to type 'Condition<M>'.ts(2322)
```

docs/site/migration/models/mixins.md
```diff
@@ -116,13 +121,30 @@ the mixin.
...
+### Defining A BaseModel Class Which Extends Entity
+
+Let's define a base model class `BaseModel` in **src/models/baseModel.ts**. It
+will be used an input to the mixin later.
+
+{% include code-caption.html content="src/models/baseModel.ts" %}
+
+```ts
+import {Entity} from '@loopback/repository';
+export class BaseModel extends Entity {}
```
bajtos: I find it confusing that the class is called `BaseModel` when it's extending from `Entity`. What if the user wants to implement a mixin that can be applied to `Model` subclasses too, not only to `Entity` descendants? 

I am proposing to use the name `BaseEntity` instead.

docs/site/migration/models/mixins.md
```diff
@@ -566,30 +505,30 @@ import {Model} from '@loopback/repository';
 import {FindByTitleInterface} from './findByTitleInterface';
 import {param, get, getModelSchemaRef} from '@loopback/rest';
 
-export interface FindByTitleControllerMixinParms {
+export interface FindByTitleControllerMixinOptions {
   basePath: string;
-  modelClass: Constructor<{}>;
+  modelClass: Constructor<object>;
```
bajtos: As mentioned in https://github.com/strongloop/loopback-next/pull/4723#discussion_r383925889, I think this should be:

```suggestion
  modelClass: typeof Model;
```

docs/site/migration/models/mixins.md
```diff
@@ -566,30 +505,30 @@ import {Model} from '@loopback/repository';
...
   modelClassName: string;
```
bajtos: Let's find a way how to avoid the requirement to specify both model class & model name, see the discussion in https://github.com/strongloop/loopback-next/pull/4723#discussion_r383925889

bajtos:
> I think we should introduce an automated test to put the proposed mixin code snippets together and verify that the code can be compiled and works as intended at runtime. See the #4723 for inspiration. That way when we need to tweak the migration instructions later in the future, we have a place where to try the proposed modifications and verify that they work as intended.

Just to make it clear, I am asking for tests verifying the mechanics of implementing a LB4 mixin class and applying it to a model, a repository and a controller. These tests SHOULD NOT reproduce an entire end-to-end LB4 application with its different source code files and so on. The migration guide tells users to put different pieces of code to different source-code files. In a test, all pieces can be together in the test file (or a fixture file, if that's more appropriate).

See [Define your testing strategy](https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html#define-your-testing-strategy) and [How to build a great test suite](https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html#how-to-build-a-great-test-suite) in our docs.

emonddr:
@bajtos 

Using 
```
export function FindByTitleRepositoryMixin<
  M extends Model & {title: string},
  R extends Constructor<CrudRepository<M>>
>(superClass: R) {
```

and

```
   async findByTitle(title: string): Promise<M[]> {
      const where = {title} as Where<M>;
      const titleFilter = {where};
      return this.find(titleFilter);
    }
```

works

thx

emonddr:
Just had a short meeting with @bajtos , and we agreed that I should create some model and repo mixin test cases in the package/repository  and a controller mixin in package/rest .

An examples/mixins can be a separate PR and task.

emonddr:
> > I think we should introduce an automated test to put the proposed mixin code snippets together and verify that the code can be compiled and works as intended at runtime. See the #4723 for inspiration. That way when we need to tweak the migration instructions later in the future, we have a place where to try the proposed modifications and verify that they work as intended.
> 
> Just to make it clear, I am asking for tests verifying the mechanics of implementing a LB4 mixin class and applying it to a model, a repository and a controller. These tests SHOULD NOT reproduce an entire end-to-end LB4 application with its different source code files and so on. The migration guide tells users to put different pieces of code to different source-code files. In a test, all pieces can be together in the test file (or a fixture file, if that's more appropriate).
> 
> See [Define your testing strategy](https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html#define-your-testing-strategy) and [How to build a great test suite](https://loopback.io/doc/en/lb4/Defining-your-testing-strategy.html#how-to-build-a-great-test-suite) in our docs.

I will work on these tests today.

emonddr:
I also created this followup doc task suggested by @bajtos : https://github.com/strongloop/loopback-next/issues/4759

emonddr:
@bajtos , @raymondfeng , I have added tests to prove that mixins in docs work.

dhmlau:
A suggestion to consider: Since it's a migration guide, instead of we have:
```
LoopBack 3 approach to .....
LoopBack 4 approach to .....
```

would it be better to have:
```
Creating a Property Mixin
   LoopBack 3 approach
   LoopBack 4 approach
Creating A Custom Model Method And Remote Model Method Mixin
   LoopBack 3 approach
   LoopBack 4 approach
```

packages/repository/src/__tests__/fixtures/mixins/categoryPropertyMixin.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
bajtos: We use kebab-case in file names. Please rename this file to `category-property-mixin.ts`.

packages/repository/src/__tests__/fixtures/mixins/categoryPropertyMixin.ts
```diff
@@ -0,0 +1,26 @@
...
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {property} from '../../../index';
```
bajtos: Please either remove `index` (import from `'../../..`), or change the import to reference the specific file implementing the decorator (I think that would be `'../../../src/decorators/model.decorator`).

packages/repository/src/__tests__/fixtures/mixins/categoryPropertyMixin.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function AddCategoryPropertyMixin<T extends Constructor<any>>(
```
bajtos: Can we use `Constructor<Entity>` to avoid no-explicit-any problem? I think such type is more correct than `Constructor<any>`.

packages/repository/src/__tests__/fixtures/mixins/findByTitleInterface.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
bajtos: Please rename this file to `find-by-title-interface.ts`.

Personally, I find it too verbose to have a dedicated file for this interface, have you considering moving the interface to the same file where `FindByTitleRepositoryMixin` is defined?

packages/repository/src/__tests__/fixtures/mixins/findByTitleInterface.ts
```diff
@@ -0,0 +1,10 @@
...
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Model} from '../../../index';
```
bajtos: Please remove `index` from the import path.

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {CrudRepository, Model, Where} from '../../../index';
```
bajtos: Please remove `index` from the import path.

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
bajtos: Please rename the file to `find-by-title-repo-mixin.ts`.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,48 @@
...
+    const methodsFound = crudMethods.filter(methodName => {
+      return methodName in repo;
+    });
+    expect(crudMethods.length).to.be.equal(methodsFound.length);
+  });
+
+  it(`mixin method 'findByTitle' exists`, () => {
+    expect('findByTitle' in repo).to.be.True();
+    expect(typeof repo.findByTitle === 'function').to.be.True();
+  });
```
bajtos: Can you please add a code calling `findByTitle`, to verify that TypeScript is seeing a correct signature? Something along the lines of:

```ts
const note: Note = await repo.findByTitle('a-title');
```

packages/repository/src/__tests__/unit/mixins/model.mixin.unit.ts
```diff
@@ -0,0 +1,48 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {MetadataInspector, MetadataMap} from '@loopback/context';
+import {expect} from '@loopback/testlab';
+import {MODEL_PROPERTIES_KEY} from '../../../';
+import {Note} from '../../fixtures/models/note.model';
+
+describe('add property to model via mixin', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let meta: MetadataMap<any>;
```
bajtos: Can we replace `any` with `unknown`?

packages/rest/src/__tests__/fixtures/controllers/note.controller.ts
```diff
@@ -0,0 +1,200 @@
...
+  @del('/notes/{id}', {
+    responses: {
+      '204': {
+        description: 'Note DELETE success',
+      },
+    },
+  })
+  async deleteById(@param.path.number('id') id: number): Promise<void> {
+    return Promise.resolve();
+  }
```
bajtos: Do we really need all these CRUD controller methods for the test? Please keep only those methods that are relevant. I think that means either one method (to verify that both mixed-in and user-provided methods are recognized) or no methods at all (because we want to verify that mixed-in methods are added, we don't really care about user-provided methods).

packages/rest/src/__tests__/fixtures/mixins/findByTitleControllerMixin.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export interface FindByTitleControllerMixinOptions {
+  basePath: string;
+  modelClass: typeof Model;
+}
+
+export function FindByTitleControllerMixin<
+  M extends Model,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends Constructor<any>
```
bajtos: Can we use `Constructor` or `Constructor<Object>` to avoid `any` problem?

packages/rest/src/__tests__/fixtures/mixins/findByTitleControllerMixin.ts
```diff
@@ -0,0 +1,50 @@
...
+>(superClass: T, options: FindByTitleControllerMixinOptions) {
+  class MixedController extends superClass implements FindByTitleInterface<M> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
```
bajtos: It makes me wonder if it's possible to use `unknown` instead of `any` here.

packages/rest/src/__tests__/fixtures/mixins/findByTitleControllerMixin.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
bajtos: Please use kebab-case in the file name.

packages/rest/src/__tests__/fixtures/mixins/findByTitleInterface.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Model} from '@loopback/repository';
+
+export interface FindByTitleInterface<M extends Model> {
+  findByTitle(title: string): Promise<M[]>;
+}
```
bajtos: Please rename this file to `find-by-title-interface.ts`.

Personally, I find it too verbose to have a dedicated file for this interface, have you considering moving the interface to the same file where `FindByTitleControllerMixin` is defined?

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,38 @@
...
+      return methodName in controller;
+    });
+    expect(methods.length).to.be.equal(methodsFound.length);
+  });
+
+  it(`mixin method 'findByTitle' exists`, () => {
+    expect('findByTitle' in controller).to.be.True();
+    expect(typeof controller.findByTitle === 'function').to.be.True();
+  });
+});
```
bajtos: It would be nice to have a test that's invoking the mixed-in endpoint via HTTP, to ensure it's correctly picked up by our REST layer.

packages/repository/src/__tests__/fixtures/mixins/baseEntity.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity} from '../../../index';
```
bajtos: Please either remove `index` (import from `'../../..`), or change the import to reference the specific file implementing the imported artifacts.

raymondfeng: It should be `import {Entity} from '../../..`

packages/repository/src/__tests__/fixtures/mixins/baseEntity.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
raymondfeng: The file name uses camel case, which is not consistent with our code base. I guess you rename it to `base-entity.ts`.

emonddr:
> Can we use Constructor<Entity> to avoid no-explicit-any problem? I think such type is more correct than Constructor<any>. 

I spoke with @bajtos  and we agreed I can use `Model` instead.

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
```diff
@@ -0,0 +1,25 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {CrudRepository, Model, Where} from '../../../index';
+import {FindByTitleInterface} from './findByTitleInterface';
+
+/*
+ * This function adds a new method 'findByTitle' to a repository class
+ * where 'M' is a model and 'R' is DefaultCrudRepository
```
raymondfeng: Please use `@typeParam` tsdoc comments.

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/context';
+import {CrudRepository, Model, Where} from '../../../index';
+import {FindByTitleInterface} from './findByTitleInterface';
```
raymondfeng: I think we can simply move `FindByTitleInterface` into this file. The interface name can be just `FindByTitle`.

packages/repository/src/__tests__/fixtures/models/note.model.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {model, property} from '../../../';
+import {BaseEntity} from '../../fixtures/mixins/baseEntity';
+import {AddCategoryPropertyMixin} from '../../fixtures/mixins/categoryPropertyMixin';
```
raymondfeng: It should be:
```ts
import {BaseEntity} from '../mixins/baseEntity';
import {AddCategoryPropertyMixin} from '../mixins/categoryPropertyMixin';
```

packages/repository/src/__tests__/fixtures/repositories/note.repository.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {FindByTitleRepositoryMixin} from '../mixins/findByTitleRepoMixin';
+import {DefaultCrudRepository, juggler} from '../../..';
+import {Note, NoteRelations} from '../../fixtures/models/note.model';
+import {Constructor} from '@loopback/core';
```
raymondfeng: It should be:

```ts
import {Constructor} from '@loopback/core';
import {DefaultCrudRepository, juggler} from '../../..';
import {FindByTitleRepositoryMixin} from '../mixins/findByTitleRepoMixin';
import {Note, NoteRelations} from '../models/note.model';
```

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {NoteRepository} from '../../fixtures/repositories/note.repository';
```
raymondfeng: Fix this!

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/*
+ * This function adds a new method 'findByTitle' to a repository class
+ * where 'M' is a model and 'R' is DefaultCrudRepository
+ */
+export function FindByTitleRepositoryMixin<
+  M extends Model & {title: string},
+  R extends Constructor<CrudRepository<M>>
+>(superClass: R) {
```
raymondfeng: It can be simplified as:
```ts
export function FindByTitleRepositoryMixin<M extends Model & {title: string}>(
  superClass: Constructor<CrudRepository<M>>,
) {
```

emonddr:
@bajtos , @raymondfeng 

regarding

> constructor(...args: any[])

Changing this 

`constructor(...args: unknown[])`

gives the error

```
A mixin class must have a constructor with a single rest parameter of type 'any[]'.ts(2545)
```

emonddr:
Not sure what is wrong with this : https://github.com/strongloop/loopback-next/pull/4723#discussion_r385859394

raymondfeng:
> constructor(...args: any[])

We have to keep it.

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(groceryNote).to.deepEqual(foundNote[0]);
+  });
+
+  it('mixin endpoint /notes/findByTitle reachable', async () => {
+    const app = givenAnApplication();
+    const server = await givenAServer(app);
+    return whenIMakeRequestTo(server)
+      .get('/notes/findByTitle/groceries')
+      .expect(200)
+      .expect([groceryNote.toJSON()]);
```
raymondfeng: Nitpick: it can be merged as `.expect(200, [groceryNote.toJSON()]);`

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,58 @@
...
+describe('add method to controller via mixin', () => {
+  let controller: NoteController;
+
+  before(async () => {
+    controller = new NoteController();
+  });
+
+  it(`non mixin method 'create' exists`, () => {
+    expect('create' in controller).to.be.True();
+    expect(typeof controller.findByTitle === 'function').to.be.True();
```
raymondfeng: It should be `expect(controller.findByTitle).to.be.a.function();`

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,58 @@
...
+
+describe('add method to controller via mixin', () => {
+  let controller: NoteController;
+
+  before(async () => {
+    controller = new NoteController();
+  });
+
+  it(`non mixin method 'create' exists`, () => {
+    expect('create' in controller).to.be.True();
```
raymondfeng: It should be refined as `expect(controller).to.have.property('create');`

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,58 @@
...
+    expect(typeof controller.findByTitle === 'function').to.be.True();
+  });
+
+  it(`mixin method 'findByTitle' exists`, async () => {
+    expect('findByTitle' in controller).to.be.True();
+    expect(typeof controller.findByTitle === 'function').to.be.True();
```
raymondfeng: Ditto.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,55 @@
...
+      'delete',
+      'findById',
+      'updateById',
+      'replaceById',
+      'deleteById',
+      'exists',
+    ];
+    const methodsFound = crudMethods.filter(methodName => {
+      return methodName in repo;
+    });
```
raymondfeng:
```ts
const methodsFound = crudMethods.filter(methodName => typeof repo[methodName] === 'function');
```

emonddr: > ```ts
> const methodsFound = crudMethods.filter(methodName => typeof repo[methodName] === 'function');
> ```

Cannot seem to use this...it gives an error:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'NoteRepository'.
  No index signature with a parameter of type 'string' was found on type 'NoteRepository'.ts(7053)
```

bajtos: Re-posting from our chat:

Personally, I am not a big fan of checking exactly what methods an interface or a class has. Such tests are brittle, because when we add a new method to `DefaultCrudRepository`, then the test will start to fail. Surely the test is not interested in exact method names provided by `DefaultCrudRepository`? IIUC, you want to say "repo has all DefaultCrudRepository members plus `findByTitle`". So the correct solution is to list all members of DefaultCrudRepository, add `findByTitle` to that array, and use that as the expected list of members in `repo`.

---

Let's take a look at your code:

```ts
    const methodsFound = crudMethods.filter(methodName => {
      return methodName in repo && typeof repo[methodName] === 'function';
    });
```

It takes an array of strings (crudMethods) and for each string value, it tries to access a member of repo with that name.

The problem is that from TypeScript's point of view, those strings can be arbitrary - you can sneak in `explode` or make a typo and use `craete` instead of `create`.

When analyzing the expression `repo[methodName]`, TypeScript is not able to figure out what type does it resolves to. It's a valid statement in JavaScript and because TypeScript cannot figure out what's the type, it uses `any`.

The problem with `any` is that once you get `any`, all further compile checks are pretty much disabled. This is often difficult to spot, therefore we have configured the compiler to tell us about these situations and reject them with errors.

I am proposing to build the list of expected methods as follow (please add the typeof function check yourself):

```ts
    const crudMethods: (keyof DefaultCrudRepository<
      Note,
      typeof Note.prototype.id
    >)[] = Object.keys(DefaultCrudRepository);
```

With that change in place, the compiler will discover the problem described in my comment https://github.com/strongloop/loopback-next/pull/4723#pullrequestreview-367270278 below.

bajtos: The code snippet above does not work, here is the corrected version:

```ts
    const crudProto = DefaultCrudRepository.prototype;
    const crudMembers = Object.keys(crudProto) as (keyof DefaultCrudRepository<
      Note,
      typeof Note.prototype.id
    >)[];
    const crudMethods = crudMembers.filter(
      key => typeof crudProto[key] === 'function',
    );
```

raymondfeng: Good catch!

packages/repository/src/__tests__/fixtures/mixins/findByTitleRepoMixin.ts
emonddr: > It can be simplified as:
> 
> ```ts
> export function FindByTitleRepositoryMixin<M extends Model & {title: string}>(
>   superClass: Constructor<CrudRepository<M>>,
> ) {
> ```

It turns out your new proposal works at runtime, but the compiler is not aware of certain methods like repo.exists or repo.findById. 

This was discovered when @bajtos suggested tests for these methods too.

Then we re-worked the repo mixin to your original suggestion last week:

```
export function FindByTitleRepositoryMixin<
  E extends Entity & {title: string},
  ID,
  Relations extends object = {}
>(superClass: Constructor<EntityCrudRepository<E, ID, Relations>>) {
  return class extends superClass implements FindByTitle<E> {
    async findByTitle(title: string): Promise<E[]> {
      const where = {title} as Where<E>;
      const titleFilter = {where};
      return this.find(titleFilter);
    }
  };
}
```

emonddr:
> Re-posting from our chat:
> 
> > The compiler sees only `CrudRepository` members, it does not see members contributed by `EntityCrudRepository` on top of `CrudRepository` (e.g. `exists` and `findById`)
> 
> I think it is very important to capture this scenario in a test and make sure it works both at compile time and runtime too.
> 
> An example test case:
> 
> ```ts
> it('has access to EntityCrudRepository methods', async () => {
>   const note: Note = await repo.create(
>     new Note({title: 'groceries', content: 'eggs,bacon', category: 'keto'}),
>   );
> 
>   const exists = await repo.exists(note.id);
>   expect(exists).to.equal(true);
> });
> ```

In our slack conversation you mentioned `exists` or `findById` since they come from `EntityCrudRepository`. `create` comes from `CrudRepository`, and `findByTitle` from mixin. So I added a test case for : create, findById, and findByTitle. No need for `exists` if we have `findById`. :)

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
emonddr: @bajtos thanks for the workaround. But I decided only to test `create`, `findById`, and `findByTitle` ; instead of all the crud functions.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,55 @@
...
+    ds = new DataSource({
+      connector: 'memory',
+    });
+    repo = new NoteRepository(ds);
+  });
+
+  // method from CrudRepository
+  it(`non-mixin method 'create' exists`, async () => {
+    expect(repo).to.have.property('create');
+    expect(repo.create).to.be.a.Function();
```
raymondfeng: Nitpick: this assertion already covers the one above.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,55 @@
...
+    note = await repo.create(new Note(noteData));
+    expect(note.toJSON()).to.deepEqual({id: 1, ...noteData});
+  });
+
+  // method from EntityCrudRepository
+  it(`non-mixin method 'findById' exists`, async () => {
+    expect(repo).to.have.property('findById');
+    expect(repo.findById).to.be.a.Function();
```
raymondfeng: Ditto.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,55 @@
...
+    note = await repo.create(new Note(noteData));
+    const foundNote = await repo.findById(note.id);
+    expect(foundNote).to.deepEqual(note);
+  });
+
+  // method from mixin
+  it(`mixin method 'findByTitle' exists`, async () => {
+    expect(repo).to.have.property('findByTitle');
```
raymondfeng: Ditto.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
emonddr: @bajtos , actually after having a discussion with @raymondfeng , we decided to keep the using of `Model` instead of `Entity` for all the mixin functions, and use `CrudRepository` instead of `EntityCrudRepository` for the sole reason that the mixin method in the documentation example add a `findByTitle` method which only needs to use `repo.find`...it doesn't need `create` or `findById` nor `exists`. So our mocha tests will reflect this as well. I have reverted the repomixin changes, and will only test `create` and `findByTitle`.  :)

emonddr:
hmm. A repo and controller generated from CLI expects EntityCrudRepository methods for the repo. So I need to change the repo mixin again. Will update this tonight.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
emonddr:
```
    const crudProto = DefaultCrudRepository.prototype;
    const crudMembers = Object.keys(crudProto) as (keyof DefaultCrudRepository<
      Note,
      typeof Note.prototype.id
    >)[];
    const crudMethods = crudMembers.filter(
      key => typeof crudProto[key] === 'function',
    );
```
This just produces empty arrays for me.

emonddr:
1) Working

If we have

**repo mixin:**

```
export function FindByTitleRepositoryMixin<
  E extends Entity & {title: string},
  R extends Constructor<CrudRepository<E>>
>(superClass: R) {
  class MixedRepository extends superClass implements FindByTitle<E> {
    async findByTitle(title: string): Promise<E[]> {
      const where = {title} as Where<E>;
      const titleFilter = {where};
      return this.find(titleFilter);
    }
  }
  return MixedRepository;
}
```

**note repo:**

```
export class NoteRepository extends FindByTitleRepositoryMixin<
  Note,
  Constructor<
    DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>
  >
>(DefaultCrudRepository) {
  constructor(dataSource: juggler.DataSource) {
    super(Note, dataSource);
  }
}
```

Then repo has access to all CrudRepo and EntityCrudRep methods.

2) Not Working

repo mixin:

```
export function FindByTitleRepositoryMixin<E extends Entity & {title: string}>(
  superClass: Constructor<CrudRepository<E>>,
) {
  return class extends superClass implements FindByTitle<E> {
    async findByTitle(title: string): Promise<E[]> {
      const where = {title} as Where<E>;
      const titleFilter = {where};
      return this.find(titleFilter);
    }
  };
}
```

note repo :  

```
export class NoteRepository extends FindByTitleRepositoryMixin<Note>(
  DefaultCrudRepository,
) {
  constructor(dataSource: juggler.DataSource) {
    super(Note, dataSource);
  }
}
```

Then repo doesn't have access to EntityCrudRepo methods.

emonddr:
@bajtos , @raymondfeng  , I got the mixins using Entity instead of Model, and I reverted the repo mixin definition into one where repo can see the CrudRep methods and EntityCrudRepo methods. So if our mocha test for repo mixin can identify the 'create' (CrudRepo), 'findById' (EntityCrudRepo), and findByTitle (mixin), we are good. This also means a CLI generated controller which uses a CLI generated repo (using DefaultCrudRepository) can make use of all the repo methods. Late last week we moved from this repo mixin definition which was working, to a slimmer one we thought worked, but it actually made the entitycrudrepo methods unavailable.

Please take a look at the mixin definitions in the mocha tests directory, and if everyone's happy...I will update the mixin.md file accordingly. thx.

raymondfeng:
@emonddr I have suggested changes in https://github.com/strongloop/loopback-next/pull/4778. If you agree, please merge #4778 to your feature branch directly (you don't have to manually update your branch).

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
bajtos: > we decided to keep the using of `Model` instead of `Entity` for all the mixin functions, and use `CrudRepository` instead of `EntityCrudRepository` for the sole reason that the mixin method in the documentation example add a `findByTitle` method which only needs to use `repo.find`...it doesn't need `create` or `findById` nor `exists`.

Sure, that makes a lot of sense. I my view, we are designing the mixing to require only those methods that we actually need in the mixin implementation. If `findByTitle` needs only the `find` method, then we should require `CrudRepository` only and don't ask for `EntityCrudRepository` methods that we don't need.

> will only test `create` and `findByTitle`. 

I am afraid I strongly disagree with the second part. When the user calls `FindByTitleRepositoryMixin` and provides a subclass of `DefaultCrudRepository` (which implements `EntityCrudRepository` interface), then the class created by the mixin should have all `EntityCrudRepository` methods. This will be always true at runtime, we need to find a way how to make the TypeScript compiler understand that at compile-time too.

A test calling `repo.findById` verifies both compile-time and run-time behavior, I think it's crucial to keep such test included.

Based on the newer comments and the the latest version presented in cbcc296, I think my concerns are actually addressed because the test for `findById` is present :+1:

packages/repository/src/__tests__/fixtures/mixins/find-by-title-repo-mixin.ts
```diff
@@ -0,0 +1,32 @@
...
+/*
+ * This function adds a new method 'findByTitle' to a repository class
+ * where 'E' is a model which extends Entity
+ *
+ * @typeParam E - Model class which extends Entity
+ */
+
+export function FindByTitleRepositoryMixin<
+  E extends Entity & {title: string},
+  R extends Constructor<CrudRepository<E>>
```
bajtos: It seems to me that you are mixing two clean approaches here and arriving at a mixed and a kind of "dirty" solution.

- Approach 1: Require only Model + CRUD methods: use `M extends Model & {title: string}` and `R extends Constructor<CrudRepository<M>>`
- Approach 2: Require `Entity` + CRUD + entity-specific methods, to make the mixin easier to use: use `E extends Entity & {title: string}` and `R extends Constructor<EntityCrudRepository<E>>`.

IIRC, the idea behind the second approach was to simplify definition of `NoteRepository` so that it does not have to specify full repository type as you have now:

```ts
export class NoteRepository extends FindByTitleRepositoryMixin<
   Note,
   Constructor<
     DefaultCrudRepository<Note, typeof Note.prototype.id, NoteRelations>
   >
 >(DefaultCrudRepository)  {
  // ...
}
```

IIRC, we were aiming for the following usage in Approach 2:

```ts
export class NoteRepository extends FindByTitleRepositoryMixin<
   Note,
   typeof Note.prototype.id,
   NoteRelations,
 >(DefaultCrudRepository) {
  // ...
}
```

Based on @raymondfeng proposal in https://github.com/strongloop/loopback-next/pull/4778, I think we should simply revert back to Approach 1 and use `Model` + `CrudRepository` in the generic arguments of the mixin.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,52 @@
...
+  beforeEach(async () => {
+    ds = new DataSource({
+      connector: 'memory',
+    });
+    repo = new NoteRepository(ds);
+  });
+
+  // method from CrudRepository
+  it(`non-mixin method 'create' exists`, async () => {
+    expect(repo.create).to.be.a.Function();
```
bajtos: In fact, calling `repo.create` on the next line is already covering the assertion `expect(repo.create).to.be.a.Function()`. If `create` is not a method (a function), then the JavaScript runtime will throw an error that will fail the test:

```
Uncaught TypeError: repo.create is not a function
```

IMO, the error message reported by the runtime will be actually more helpful than the assertion message, because it can describe both the variable (`repo`) and the member name (`create`). The assertion message will just say something like `expected undefined to be a function`.

I am proposing to remove these `.to.be.a.Function()` assertions from all three tests in this file.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,52 @@
...
+    note = await repo.create(new Note(noteData));
+    expect(note.toJSON()).to.deepEqual({id: 1, ...noteData});
+  });
+
+  // method from EntityCrudRepository
+  it(`non-mixin method 'findById' exists`, async () => {
+    expect(repo.findById).to.be.a.Function();
```
bajtos: Please remove per my comment above.

packages/repository/src/__tests__/unit/mixins/add-method-to-repository.mixin.unit.ts
```diff
@@ -0,0 +1,52 @@
...
+    note = await repo.create(new Note(noteData));
+    const foundNote: Note = await repo.findById(note.id);
+    expect(foundNote).to.deepEqual(note);
+  });
+
+  // method from mixin
+  it(`mixin method 'findByTitle' exists`, async () => {
+    expect(repo.findByTitle).to.be.a.Function();
```
bajtos: Please remove per my comment above.

packages/rest/src/__tests__/fixtures/controllers/note.controller.ts
```diff
@@ -0,0 +1,69 @@
+import {Constructor} from '@loopback/core';
```
bajtos: Missing license header, please add (see #4778)

packages/rest/src/__tests__/unit/controller-mixin/controller.mixin.unit.ts
```diff
@@ -0,0 +1,54 @@
...
+  before(async () => {
+    controller = new NoteController();
+  });
+
+  it(`non-mixin method 'create' exists`, () => {
+    expect(controller.create).to.be.a.Function();
+  });
+
+  it(`mixin method 'findByTitle' exists`, async () => {
+    expect(controller.findByTitle).to.be.a.Function();
```
bajtos: Similarly here, this check is redundant and produces less helpful failure message than the next line calling `controller.findByTitle`. Please remove.

emonddr:
@raymondfeng 

> @emonddr I have suggested changes in #4778. If you agree, please merge #4778 to your feature branch directly (you don't have to manually update your branch).

Actually there are a few spots that `Entity` got changed back to `Model`, so I won't be applying your PR to my feature branch. I've looked at the PR, and have manually copied over the changes. But kept `Entity` instead of changing it to Model. Thank you for the enhancements, though. :)

packages/repository/src/__tests__/fixtures/mixins/find-by-title-repo-mixin.ts
emonddr: @bajtos, @raymondfeng 

I cannot seem to make this work:

> R extends Constructor<EntityCrudRepository<E, ID, Relations>>

VSCode complains about ID and Relations.

emonddr: > It seems to me that you are mixing two clean approaches here and arriving at a mixed and a kind of "dirty" solution. 

Not sure why using Entity instead of Model is a dirty approach. Can you clarify?

raymondfeng: With the introduction of `R extends Constructor<CrudRepository<M>>` in your PR, I was able to downgrade `Entity` to `Model` in #4778 and all tests are passing. I think we now have a good solution for the repo mixin to use `Model` directly.

packages/repository/src/__tests__/fixtures/repositories/note.repository.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Constructor} from '@loopback/core';
+import {DataSource} from 'loopback-datasource-juggler';
```
raymondfeng: We should avoid direct import to `loopback-datasource-juggler`.  Use `juggler.DataSource` instead.

```
import {juggler, DefaultCrudRepository} from '../../..';
```

packages/rest/package.json
```diff
@@ -63,7 +63,8 @@
     "@types/node": "^10.17.17",
     "@types/on-finished": "^2.3.1",
     "@types/qs": "^6.9.1",
-    "multer": "^1.4.2"
+    "multer": "^1.4.2",
+    "loopback-datasource-juggler": "^4.18.1"
```
raymondfeng: Please revert this.

packages/rest/src/__tests__/fixtures/repositories/note.repository.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {DefaultCrudRepository, Where} from '@loopback/repository';
+import {DataSource} from 'loopback-datasource-juggler';
```
raymondfeng: Ditto. Use `juggler.DataSource`

packages/repository/src/__tests__/fixtures/mixins/category-property-mixin.ts
```diff
@@ -0,0 +1,25 @@
...
+) {
+  class MixedModel extends superClass {
+    @property({
+      type: 'string',
+      required: true,
+    })
+    category: string;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
```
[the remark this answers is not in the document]
emonddr: I removed the constructors from all the mixins as they are not needed. tested this. They still work.

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
...
+
+- creating a base model class which extends `Entity`
+- placing the mixin class factory function in a separate file
+- generating a model using the CLI as usual
+- adjusting the model file to make use of the mixin class factory function
+
+#### Defining A BaseEntity Class Which Extends Entity
+
+Let's define a base model class `BaseEntity` in **src/models/base-entity.ts**.
+It will be used an input to the mixin later.
```
jannyHou:
```suggestion
It will be used as input to the mixin later.
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
...
+
+{% include code-caption.html content="src/models/base-entity.ts" %}
+
+```ts
+import {Entity} from '@loopback/repository';
```
jannyHou: (I know it's a rare case and outside of the scope of this PR :p) If we apply the mixin to a `Model` instead of `Entity`, also need to define a BaseModel class to extend the abstract [`Model`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L203) right?

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
...
+export type NoteWithRelations = Note & NoteRelations;
+```
+
+The required property `category` has now been added to the `Note` model via a
+mixin class factory function.
+
+## Creating A Custom Model Method And Remote Model Method Mixin
+
+This section covers the approach that LoopBack 3 can use to add a custom method
+/ remote method to a model via a mixin, and similarly how LoopBack 4 can add a
```
jannyHou:
```suggestion
/remote method to a model via a mixin, and similarly how LoopBack 4 can add a
```

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
...
+- updating the server/model-config.json file with the mixin directory location
+- updating the model's json file to include the mixin's name (and options object
+  or boolean)
+
+#### Defining The Model Method Mixin findByTitle.js
+
+The developer defines a custom model method/remote method mixin in
+**common/mixins/findByTitle.js** which adds a custom method `findByTitle` to any
+model, and adds a corresponding remote method definition with path
+`/findByTitle` as well. An options property `returnArgumentName` makes it
```
jannyHou:
```suggestion
`/findByTitle` as well. An option property `returnArgumentName` makes it
```
ignore it if I am wrong :p

docs/site/migration/models/mixins.md
```diff
@@ -6,8 +6,730 @@ sidebar: lb4_sidebar
...
+  // ...
+  // remaining CRUD endpoints
+  // ...
+}
+```
+
+For a full example of a CLI-generated controller for a model `Todo`, see
+[TodoController ](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/controllers/todo.controller.ts).
+
+#### Adjust NoteController To Use FindByTitleControllerMixin
```
jannyHou:
```suggestion
#### Adjusting NoteController To Use FindByTitleControllerMixin
```

emonddr:
> looks like mixins are not automatically migrated. We ask developers to create the necessary classes, is it ? should we mention that some where in the notes ?

@deepakrkris , automated commands for migration will possibly come later; depending on our roadmap. At the moment, we are providing manual instructions for migration.

docs/site/migration/models/mixins.md
emonddr: > (I know it's a rare case and outside of the scope of this PR :p) If we apply the mixin to a `Model` instead of `Entity`, also need to define a BaseModel class to extend the abstract [`Model`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/model.ts#L203) right?

Out of scope for this PR. Most people will use Entity and EntityCrudRepository. This guide describes HOW to create a mixin in LB4.  With that knowledge, they can create whatever mixins they want, and targeting whatever model objects they want. :)

docs/site/migration/models/mixins.md
emonddr: @jannyHou  : `options` is the name of the parameter. And `returnArgumentName` is a property inside it.

docs/site/migration/models/mixins.md
emonddr: Good catch!

Issue #3952, opened 2019-10-17
How to migrate CRUD Operation Hooks


bajtos:
_This is a follow-up for #3718 and #3922._

Write content for `docs/site/migration/models/operation-hooks.md`, explain how to migrate operation hooks (see LB3 docs: [Operation Hooks](https://loopback.io/doc/en/lb3/Operation-hooks.html)).

Explain that we don't have first-class support for Operation Hooks in LB4 yet, refer to https://github.com/strongloop/loopback-next/issues/1919

As a temporary solution, describe how to leverage `PersistedModel` used by our legacy juggler bridge. Example implementation:

```ts
export class TodoRepository extends DefaultCrudRepository<
  Todo,
  typeof Todo.prototype.id,
  TodoRelations
> {
  constructor(@inject('datasources.db') dataSource: juggler.DataSource) {
    super(Todo, dataSource);

    this.modelClass.observe('save', async ctx => {
      console.log(
        'Going to write to an instance of model %s',
        ctx.Model.modelName,
      );
    });
  }
}
```

Caveats:

(1)
Repository constructor is executed once for each incoming request. We don't want to install another copy of the same operation hook on each request, we need to find a way how to install hooks only once. Ideally, `DefaultCrudRepository` should provide a protected method that subclasses can use to supply one-time model setup.

For example:

```ts
export class TodoRepository extends DefaultCrudRepository<
  Todo,
  typeof Todo.prototype.id,
  TodoRelations
> {
  constructor(@inject('datasources.db') dataSource: juggler.DataSource) {
    super(Todo, dataSource);
  }
  
  protected setupPersistedModel() {
    super.setupPersistedModel();

    this.modelClass.observe('save', async ctx => {
      console.log(
        'Going to write to an instance of model %s',
        ctx.Model.modelName,
      );
    });
  }
}
```

(2)
Current TypeScript typings in loopback-datasource-juggler do not contain operation hooks and `.observe()` method. We need to fix this.

## Acceptance criteria

- [x] Add typings for Operation Hooks to loopback-datasource-juggler
- [x] Add a protected method `definePersistedModel` to `DefaultCrudRepository`, this method should overriden by user's repository class to access the model class where the operation hooks can be applied. 
- [x] Migration guide describing how to move LB3 operation hooks to LB4 repositories.

marioestradarosa:
@bajtos 

How about decorators inside the repository to allow a particular method to become part of the event (before/after etc)

```ts
[...]

PR #4730, opened 2020-02-24
feat: support operation hooks

hacksparrow:
Added support for operation hooks. Addresses https://github.com/strongloop/loopback-next/issues/3952.

Signed-off-by: Hage Yaapa <hage.yaapa@in.ibm.com>

## Checklist

### DIFF
--- a/docs/site/migration/models/operation-hooks.md
+++ b/docs/site/migration/models/operation-hooks.md
@@ -6,8 +6,46 @@ sidebar: lb4_sidebar
 permalink: /doc/en/lb4/migration-models-operation-hooks.html
 ---
 
-{% include note.html content="
-This is a placeholder page, the task of adding content is tracked by the
-following GitHub issue:
-[loopback-next#3952](https://github.com/strongloop/loopback-next/issues/3952)
-" %}
+Operation hooks are not supported in LoopBack 4 yet. See the
+[Operation hooks for models/repositories spike](https://github.com/strongloop/loopback-next/issues/1919)
+to follow the progress made on this subject.
+
+In the meantime, we are providing a temporary API for enabling operation hooks
+in LoopBack 4: override `DefaultCrudRepository`'s `definePersistedModel` method
+in the model's repository.
+
+The `definePersistedModel` method of `DefaultCrudRepository` returns a model
+class on which you can apply the
+[LoopBack 3 operation hooks](https://loopback.io/doc/en/lb3/Operation-hooks.html).
+Make sure to return the model class from your repository's
+`definePersistedModel` method.
+
+Here is an example of a repository implementing `definePersistedModel` and
+applying an operation hook on a model:
+
+```ts
+class ProductRepository extends DefaultCrudRepository<
+  Product,
+  typeof Product.prototype.id,
+  ProductRelations
+> {
+  constructor(dataSource: juggler.DataSource) {
+    super(Product, dataSource);
+  }
+
+  definePersistedModel(entityClass: typeof Product) {
+    const modelClass = super.definePersistedModel(entityClass);
+    modelClass.observe('before save', async ctx => {
+      console.log(`going to save ${ctx.Model.modelName}`);
+    });
+    return modelClass;
+  }
+}
+```
+
+Although possible, we are not providing an API which directly exposes the
+`observe` method of the model class. The current API makes the registration of
+operation hooks a process that is possible only at the time when the model class
+is attached to the repository and accidental registration of the same operation
+hook multiple times becomes obvious. With an API which directly exposes the
+`observe` method of the model class, this would not have been possible.
--- a/packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
+++ b/packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
@@ -1,0 +1,57 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {expect} from '@loopback/testlab';
+import {DataSource} from 'loopback-datasource-juggler';
+import {DefaultCrudRepository, juggler} from '../..';
+import {Product, ProductRelations} from '../fixtures/models/product.model';
+
+// This test shows the recommended way how to use @loopback/repository
+// together with existing connectors when building LoopBack applications
+describe('Operation hooks', () => {
+  let repo: ProductRepository;
+  beforeEach(givenProductRepository);
+
+  const beforeSave = 'before save';
+  const afterSave = 'after save';
+  const expectedArray = [beforeSave, afterSave];
+
+  it('supports operation hooks', async () => {
+    await repo.create({slug: 'pencil'});
+    expect(repo.hooksCalled).to.eql(expectedArray);
+  });
+
+  function givenProductRepository() {
+    const db = new DataSource({
+      connector: 'memory',
+    });
+
+    repo = new ProductRepository(db);
+  }
+
+  class ProductRepository extends DefaultCrudRepository<
+    Product,
+    typeof Product.prototype.id,
+    ProductRelations
+  > {
+    constructor(dataSource: juggler.DataSource) {
+      super(Product, dataSource);
+    }
+
+    hooksCalled: string[] = [];
+
+    definePersistedModel(entityClass: typeof Product) {
+      const modelClass = super.definePersistedModel(entityClass);
+      modelClass.observe(beforeSave, async ctx => {
+        this.hooksCalled.push(beforeSave);
+      });
+
+      modelClass.observe(afterSave, async ctx => {
+        this.hooksCalled.push(afterSave);
+      });
+      return modelClass;
+    }
+  }
+});
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -127,11 +127,11 @@ export class DefaultCrudRepository<
       `Entity ${entityClass.name} must have at least one id/pk property.`,
     );
 
-    this.modelClass = this.definePersistedModel(entityClass);
+    this.modelClass = this.ensurePersistedModel(entityClass);
   }
 
   // Create an internal legacy Model attached to the datasource
-  private definePersistedModel(
+  private ensurePersistedModel(
     entityClass: typeof Model,
   ): typeof juggler.PersistedModel {
     const definition = entityClass.definition;
@@ -147,6 +147,21 @@ export class DefaultCrudRepository<
       // The backing persisted model has been already defined.
       return model as typeof juggler.PersistedModel;
     }
+
+    return this.definePersistedModel(entityClass);
+  }
+
+  /**
+   * Creates a legacy persisted model class, attaches it to the datasource and
+   * returns it. This method can be overriden in sub-classes to acess methods
+   * and properties in the generated model class.
+   * @param entityClass - LB4 Entity constructor
+   */
+  protected definePersistedModel(
+    entityClass: typeof Model,
+  ): typeof juggler.PersistedModel {
+    const dataSource = this.dataSource;
+    const definition = entityClass.definition;
 
     // To handle circular reference back to the same model,
     // we create a placeholder model that will be replaced by real one later
@@ -192,7 +207,7 @@ export class DefaultCrudRepository<
   private resolvePropertyType(type: PropertyType): PropertyType {
     const resolved = resolveType(type);
     return isModelClass(resolved)
-      ? this.definePersistedModel(resolved)
+      ? this.ensurePersistedModel(resolved)
       : resolved;
   }
 


docs/site/migration/models/operation-hooks.md
```diff
@@ -6,8 +6,38 @@ sidebar: lb4_sidebar
...
+  typeof Product.prototype.id,
+  ProductRelations
+> {
+  constructor(dataSource: juggler.DataSource) {
+    super(Product, dataSource);
+  }
+
+  definePersistedModel(entityClass: typeof Product) {
+    const model = super.definePersistedModel(entityClass);
+    model.observe('before save', async ctx => {
```
raymondfeng: We can access the model as `this.modelClass`. Why not leveraging that to set up observers instead of overriding `definedPersistedModel`?

hacksparrow:
Needs https://github.com/strongloop/loopback-datasource-juggler/pull/1820 to land for tests to pass.

docs/site/migration/models/operation-hooks.md
hacksparrow: Initially we were considering having an `observe` method on `DefaultCrudRepository`, which would call `this.modelClass.observe`. After some brainstorming, @bajtos suggested the possible problems with `observe` is not worth it. Eg: calling `observe()` multiple times and how to handle it. He had suggested another approach but it wasn't very DX-friendly, so we settled on the current one.

"You want to directly access the model? Then you will have to override the `definePersistedModel ` method."

bajtos: @raymondfeng please check the description of the user story and the discussion in that GH issue for more context.

@hacksparrow please update PR description with a reference to the GH issue this is related to.

raymondfeng: @hacksparrow We do support multiple observers for a given model class/operation - see https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/observer.js#L51

hacksparrow: Yes, that's understood. We were wondering how should LB4 respond when an operation hook is installed multiple times: a. Throw an error? b. Accept silently - how does the user know what happened?

Also, it is not impossible to call the `observe` method before the `modelClass` is defined. What happens in that case?

Since this is a temporary workaround, we didn't want to invest too much time handling various possible scenarios, edge cases etc. We can do that when we investigate the use of interceptor for the "real" implementation.

docs/site/migration/models/operation-hooks.md
```diff
@@ -6,8 +6,38 @@ sidebar: lb4_sidebar
 permalink: /doc/en/lb4/migration-models-operation-hooks.html
 ---
 
-{% include note.html content="
-This is a placeholder page, the task of adding content is tracked by the
-following GitHub issue:
-[loopback-next#3952](https://github.com/strongloop/loopback-next/issues/3952)
-" %}
+Operation hooks are not supported in LoopBack 4 yet. It will be, eventually; you
```
dhmlau: Are we really going to have operation hooks in LB4? Maybe reword it as:
`See the [Operation hooks for models/repositories spike](https://github.com/strongloop/loopback-next/issues/1919) for the progress`

hacksparrow: Updated as per suggestion.

docs/site/migration/models/operation-hooks.md
```diff
@@ -6,8 +6,38 @@ sidebar: lb4_sidebar
...
+in LoopBack 4: override `DefaultCrudRepository`'s `definePersistedModel` method
+in the model's repository.
+
+The `definePersistedModel` method of `DefaultCrudRepository` returns a model
+instance on which you can apply the
+[LoopBack 3 operation hooks](https://loopback.io/doc/en/lb3/Operation-hooks.html).
+Make sure to return the model instance from the derived class.
+
+Here is an example of a repository implementing `definePersistedModel` and
+applying an operation hook on a model instance:
```
jannyHou: nitpick: I remember the `observe` method is called by a model class, not instance, not sure is if it's still the case in the legacy juggler bridge.

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
```diff
@@ -0,0 +1,56 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
```
agnes512: // Copyright IBM Corp. 2020

docs/site/migration/models/operation-hooks.md
hacksparrow: Thanks! Corrected.

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
```diff
@@ -0,0 +1,56 @@
...
+
+// This test shows the recommended way how to use @loopback/repository
+// together with existing connectors when building LoopBack applications
+describe('Operation hooks', () => {
+  let repo: ProductRepository;
+  beforeEach(givenProductRepository);
+
+  it('supports operation hooks', async () => {
+    const p = await repo.create({slug: 'pencil'});
+    expect(repo.callCounter).to.equal(2);
```
bajtos: When a check like this fails, it's hard to tell what exactly went wrong, because we don't know which hooks were called and which were not.

Also if somehow the hook registration logic misbehaves and registers the callback for wrong hooks (e.g. `access` instead of `before save`), then we won't catch such problem.

In the past, I have good experience with replacing a counter with an array of hook names called.

```suggestion
    expect(repo.hooksCalled).to.eql(['before save', 'after save']);
```

docs/site/migration/models/operation-hooks.md
```diff
@@ -6,8 +6,39 @@ sidebar: lb4_sidebar
...
+
+  definePersistedModel(entityClass: typeof Product) {
+    const modelClass = super.definePersistedModel(entityClass);
+    modelClass.observe('before save', async ctx => {
+      console.log(`going to save ${ctx.Model.modelName}`);
+    });
+    return modelClass;
+  }
+}
+```
```
bajtos: Please explain the problem of multiple registration of the same hook and how we are avoiding it by registering the hooks inside `definePersistedModel` method.

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
hacksparrow: 👍

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
hacksparrow: Updated

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
hacksparrow: Done

docs/site/migration/models/operation-hooks.md
hacksparrow: Done

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -5,7 +5,7 @@
 
 import {Getter} from '@loopback/context';
 import assert from 'assert';
-import legacy from 'loopback-datasource-juggler';
+import legacy, {ObserverMixin} from 'loopback-datasource-juggler';
```
bajtos: I think we are intentionally keeping juggler types very separated from LB4 types. Take a look at lines 41-53, where we are re-exporting a subset of legacy types in `juggler` namespace.

Please make the following changes:

- Revert this line back.
- Re-export `legacy.ObserverMixin` in `juggler` namespace 
- Modify all places below using `ObserverMixin` to use `juggler.Mixin`

Please add a comment to all those modified places to mention that it's a temporary workaround and add a link to the GitHub issue tracking the follow-up task of improving juggler typings to correctly include ObserverMixin in `PersistedModelClass` API.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -127,13 +127,12 @@ export class DefaultCrudRepository<
       `Entity ${entityClass.name} must have at least one id/pk property.`,
     );
 
-    this.modelClass = this.definePersistedModel(entityClass);
+    this.modelClass = this.ensurePersistedModel(entityClass);
   }
 
-  // Create an internal legacy Model attached to the datasource
```
bajtos: Please preserve this comment.

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -145,9 +144,19 @@ export class DefaultCrudRepository<
...
     if (model) {
       // The backing persisted model has been already defined.
-      return model as typeof juggler.PersistedModel;
+      return model as typeof juggler.PersistedModel & ObserverMixin;
     }
 
+    return this.definePersistedModel(entityClass);
+  }
+
+  // Create an internal legacy Model attached to the datasource
```
bajtos: Please replace this comment with a proper tsdoc-style one, explain LB4 app developers that they can override this method in their custom repository classes to tweak the legacy Model (e.g. register operation hooks).

packages/repository/src/repositories/legacy-juggler-bridge.ts
hacksparrow: 👍

packages/repository/src/repositories/legacy-juggler-bridge.ts
hacksparrow: 👍

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -144,16 +148,23 @@ export class DefaultCrudRepository<
...
 
     return this.definePersistedModel(entityClass);
   }
 
-  // Create an internal legacy Model attached to the datasource
+  /**
+   * Create a model class and attach it to the datasource. This method can be
+   * overriden in sub-classes to acess methods and properties in the generated
+   * model class.
+   * @param entityClass - Legacy entity class
```
bajtos: This is not correct. `entityClass` is the LB4 `Entity` constructor, e.g. `Product` or `User` inheriting from LB4 `Entity`.

The function returns the legacy persisted model class (e.g. `Product` inheriting from `juggler.PersistedModel`).

Please update this doc accordingly.

hacksparrow: Updated to:
```ts
  /**
   * Creates a legacy persisted model class, attaches it to the datasource and
   * returns it. This method can be overriden in sub-classes to acess methods
   * and properties in the generated model class.
   * @param entityClass - LB4 Entity constructor
   */
```

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('Operation hooks', () => {
+  let repo: ProductRepository;
+  beforeEach(givenProductRepository);
+
+  const beforeSave = 'before save';
+  const afterSave = 'after save';
+  const expectedArray = [beforeSave, afterSave];
+
+  it('supports operation hooks', async () => {
+    const p = await repo.create({slug: 'pencil'});
```
dhmlau: code linter complaining about `p` is not being used. 
Should change this to:
```
await repo.create({slug: 'pencil'});
```

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
```diff
@@ -0,0 +1,57 @@
...
+  > {
+    constructor(dataSource: juggler.DataSource) {
+      super(Product, dataSource);
+    }
+
+    hooksCalled: string[] = [];
+
+    definePersistedModel(entityClass: typeof Product) {
+      const modelClass = super.definePersistedModel(entityClass);
+      modelClass.observe(beforeSave, async ctx => {
```
dhmlau: code linter is complaining about `Promise returned in function argument where a void return was expected`. It might be because we're waiting for your changes in juggler to be released: https://github.com/strongloop/loopback-datasource-juggler/pull/1820.

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
```diff
@@ -0,0 +1,57 @@
...
+        this.hooksCalled.push(beforeSave);
+      });
+
+      modelClass.observe(afterSave, async ctx => {
```
dhmlau: code linter is complaining about `Promise returned in function argument where a void return was expected`. It might be because we're waiting for your changes in juggler to be released: https://github.com/strongloop/loopback-datasource-juggler/pull/1820.

packages/repository/src/__tests__/acceptance/operation-hooks.acceptance.ts
hacksparrow: 👍

hacksparrow:
@slnode test please

hacksparrow:
@slnode test please

docs/site/migration/models/operation-hooks.md
```diff
@@ -6,8 +6,46 @@ sidebar: lb4_sidebar
...
+
+Although possible, we are not providing an API which directly exposes the
+`observe` method of the model class. The current API makes the registration of
+operation hooks a process that is possible only after the model class is
```
bajtos:
```suggestion
operation hooks a process that is possible only at the time when the model class is
```

hacksparrow: 👍

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -148,6 +148,21 @@ export class DefaultCrudRepository<
       return model as typeof juggler.PersistedModel;
     }
 
+    return this.definePersistedModel(entityClass);
```
raymondfeng: Why don’t we emit an event when definePersistedModel is done so that listeners can be registered to set up operation hooks? Overriding a method is always more advanced for a lot of developers.

raymondfeng: For the record, we briefly discussed if it's feasible to use events to set up operation hooks and decided to defer that.

PR #4737, opened 2020-02-25
feat(cli): import LB3 models with a custom base class

bajtos:
The initial implementation of `lb4 import-lb3-models` was able to import only models inheriting from models that have a built-in counter-part in LB4: `Model`, `PersistedModel`, `KeyValueModel`. 

In this pull request, I am implementing import of models inheriting from non-LB4-builtins. There are two major use cases:
1. Import a model inheriting from an LB3 built-in model like `User`.
2. Import a model inheriting from an application-specific model.

The changes are split into multiple smaller commits to make the patch easier to review.

## Checklist


bajtos:
@emonddr  @deepakrkris @nabdelgadir @agnes512 please review

packages/cli/test/unit/import-lb3-model/migrate-model.test.js
```diff
@@ -33,7 +33,6 @@ describe('importLb3ModelDefinition', () => {
       expect(toJSON(modelData.settings)).to.deepEqual({
         // By default, LB3 models are not strict
         strict: false,
-        replaceOnPUT: true,
```
bajtos: FYI: `replaceOnPUT` is a flag set by juggler, it's used for backwards compatibility with LoopBack 2.x. 

Quoting from https://loopback.io/doc/en/lb4/Model.html:

> `replaceOnPUT` - This entry is no longer supported as LB4 controllers scaffolded by LB4 controller, PUT is always calling replaceById. Users are free to change the generated code to call patchById if needed.

packages/cli/generators/import-lb3-models/index.js
```diff
@@ -136,6 +148,20 @@ Learn more at https://loopback.io/doc/en/lb4/Importing-LB3-models.html
...
+      const baseName = templateData.modelBaseClass;
+      if (
+        !this.existingLb4ModelNames.includes(baseName) &&
+        !this.modelNames.includes(baseName)
+      ) {
+        this.log(
+          'Adding %s (base of %s) to the list of imported models.',
+          chalk.yellow(baseName),
+          chalk.yellow(name),
+        );
```
bajtos: Here is an example console output.

```
$ lb4 import-lb3-models ~/src/loopback/next/packages/cli/test/fixtures/import-lb3-models/app-using-model-inheritance.js

WARNING: This command is experimental and not feature-complete yet.
Learn more at https://loopback.io/doc/en/lb4/Importing-LB3-models.html

? Select models to import: Customer
Model Customer will be created in src/models/customer.model.ts

Adding UserBase (base of Customer) to the list of imported models.
Model UserBase will be created in src/models/user-base.model.ts

Adding User (base of UserBase) to the list of imported models.
Model User will be created in src/models/user.model.ts

Import of model relations is not supported yet. Skipping the following relations: accessTokens
Ignoring the following unsupported settings: acls
   create src/models/customer.model.ts
   create src/models/user-base.model.ts
   create src/models/user.model.ts
   update src/models/index.ts
   update src/models/index.ts
   update src/models/index.ts
```

packages/cli/snapshots/integration/generators/import-lb3-models.integration.snapshots.js
```diff
@@ -11,13 +11,7 @@ exports[`lb4 import-lb3-models imports CoffeeShop model from lb3-example app 1`]
...
 
 @model({
-  settings: {
-    strict: false,
-    forceId: false,
-    replaceOnPUT: true,
-    validateUpsert: true,
-    idInjection: true
-  }
+  settings: {strict: false, forceId: false, validateUpsert: true, idInjection: true}
```
bajtos: The following settings are no longer imported because they are inherited:

- `replaceOnPUT`

Please note that `forceId`, `idInjection` and `validateUpsert` are explicitly configured in coffee-shop's model JSON file, therefore they are preserved in the generated output. You can compare this line with `User` settings below https://github.com/strongloop/loopback-next/pull/4737/files#diff-e9be610fe11bc2c954313d9283d0eda3R128-R137, where only `strict` + custom settings are generated.

packages/cli/generators/import-lb3-models/index.js
```diff
@@ -136,6 +148,20 @@ Learn more at https://loopback.io/doc/en/lb4/Importing-LB3-models.html
     );
     debug('LB4 model data', templateData);
 
+    if (!templateData.isModelBaseBuiltin) {
+      const baseName = templateData.modelBaseClass;
+      if (
+        !this.existingLb4ModelNames.includes(baseName) &&
```
agnes512: Just want to make sure that I understand it correctly, the reason we check the existence of the base model is because we need to import the custom base class and also it's parent class if that doesn't exist?

packages/cli/generators/import-lb3-models/migrate-model.js
```diff
@@ -43,12 +44,29 @@ function importLb3ModelDefinition(modelCtor, log) {
 
   logNamingIssues(modelName, log);
 
+  const baseDefinition = modelCtor.base.definition;
+  const baseProps = {...baseDefinition.properties};
+
+  // Core LB3 models like PersistedModel come with an id property that's
+  // injected by juggler. We don't want to inherit that property, because
+  // in LB4, we want models to define the id property explicitly.
+  if (isCoreModel(modelCtor.base)) {
```
agnes512: Sorry I am a bit confused. Does this mean if a custom model inherits a base model, it won't have the id property after importing to LB4 app?

packages/cli/generators/import-lb3-models/index.js
nabdelgadir: So just making sure I understand correctly, we're importing `User` because `UserBase` inherits from it? And do we always eventually want one model to extend from a built in model (e.g. in this case `User` inheriting from `Entity`)?

bajtos: > We're importing `User` because `UserBase` inherits from it? 

Yes, exactly :+1:

> And do we always eventually want one model to extend from a built in model (e.g. in this case `User` inheriting from `Entity`)?

Yes. Every LoopBack 3 model has the LB3 built-in `Model` at the top of the inheritance chain. Some models are inheriting from `PersistedModel` (which inherits from `Model` and is mapped to `Entity` in LB4), some models are inheriting from `KeyValueModel` (which inherits from `Model` and is probably mapped to `KeyValueModel` in LB4, I am not sure).

packages/cli/generators/import-lb3-models/index.js
bajtos: Yes.

Let's say the application has:
- `AuditedPersistedModel` which is adding auditing functionality for all CRUD operations provided by `PersistedModel`,
- several models like `Product`, `Category` and `Store`, where all models are based on `AuditedPersistedModel`.

When we run import for the first time and choose only `Store` to import, we also need to import `AuditedPersistedModel` to allow the new LB4 model to inherit from the correct base class (model).

When we run import for the second time and choose to import `Product` and `Category`, we should not import `AuditedPersistedModel` again, because that could overwrite bits imported manually in the previous step. Instead, LB4 CLI will skip importing this base model.

packages/cli/generators/import-lb3-models/migrate-model.js
bajtos: When scaffolding code for an imported model, we want to skip all properties inherited from the base model. For example, if `User` model has `email` property and `Customer` model inherits from `User`, then the code for `Customer` SHOULD NOT re-define `email` property again. The property should be inherited instead, both at TypeScript and juggler-metadata level.

Now juggler's `PersistedModel` always comes with `id` property (in fact, the property is added to every model created by juggler, including the base `Model` class), see [ModelBuilder.prototype.define()](https://github.com/strongloop/loopback-datasource-juggler/blob/36303b3b63932478cb429b5c5af6fec7f09b602c/lib/model-builder.js#L311-L339).

As a result, when deciding which properties belong to the model being imported (e.g. `Product`) and which are inherited (e.g. from `PersistedModel`), the primary key property `id` was always skipped before I introduced this check, simply because it was considered as coming from the base model (`PersistedModel`).

Issue #1749, opened 2018-09-24
Sugar decorators for Filter and Where parameters


bajtos:
See https://github.com/strongloop/loopback-next/pull/1679#discussion_r219726035

Let's add few more `@param` flavors to simplify declaration of Filter and Where arguments.

```ts
@param.query.filter(Todo)
// alias for @param.query.object('filter', getFilterSchemaFor(Todo))

@param.query.filter('filterObj', Todo)
// alias for @param.query.object('filterObj', getFilterSchemaFor(Todo))

@param.query.where(Todo)
// alias for @param.query.object('where', getWhereSchemaFor(Todo))

@param.query.where('whereObj', Todo)
// alias for @param.query.object('whereObj', getWhereSchemaFor(Todo))
```

## Acceptance criteria
- [ ] Implement `@param.query.filter` and `@param.query.where` as outlined above. Test must be included, try to work test first. (Follow TestDrivenDevelopment - start by adding a failing test, and so on.)
- [ ] Update REST Controller template to use these new shortcuts
- [ ] Update all example controllers to use these new shortcuts
- [ ] Update code examples in the tutorial documentation pages to match the updated source code of controllers

dougal83:
``@param.query.object('whereObj', getWhereSchemaFor(Todo))``

I have a concern with there being a query parameter being described externally as an object.  If we have the open api spec describing the query parameter as an object, it raises a question of how we serialise into a string and satisfy the type externally based on the spec?

dhmlau:
@bajtos @dougal83, what's your suggestion on how to handle this issue? (part of the Q1 planning). Thanks.

raymondfeng:
Implemented by https://github.com/strongloop/loopback-next/pull/4745

PR #4745, opened 2020-02-25
feat: simplify filter and where usage for constraint, schema, and OpenAPI mapping

### LINKED ISSUE #1721: Add support for "include" and "fields" to findById (REST API)

raymondfeng:
Fixes #1721, #3336, and #1749 

- #1721: add type and schema for filter excluding where for `findById`
- #3336: add a test for findById with `filter.fields`
- #1749: add sugar decorators for `filter` and `where`

## Checklist

### DIFF
--- a/packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
+++ b/packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
@@ -12,6 +12,7 @@ import {
   Count,
   CountSchema,
   Filter,
+  FilterExcludingWhere,
   repository,
   Where,
 } from '@loopback/repository';
@@ -69,7 +70,7 @@ export class ProductReviewController {
     },
   })
   async count(
-    @param.query.object('where', getWhereSchemaFor(ProductReview)) where?: Where<ProductReview>,
+    @param.where(ProductReview) where?: Where<ProductReview>,
   ): Promise<Count> {
     return this.barRepository.count(where);
   }
@@ -90,7 +91,7 @@ export class ProductReviewController {
     },
   })
   async find(
-    @param.query.object('filter', getFilterSchemaFor(ProductReview)) filter?: Filter<ProductReview>,
+    @param.filter(ProductReview) filter?: Filter<ProductReview>,
   ): Promise<ProductReview[]> {
     return this.barRepository.find(filter);
   }
@@ -112,7 +113,7 @@ export class ProductReviewController {
       },
     })
     productReview: ProductReview,
-    @param.query.object('where', getWhereSchemaFor(ProductReview)) where?: Where<ProductReview>,
+    @param.where(ProductReview) where?: Where<ProductReview>,
   ): Promise<Count> {
     return this.barRepository.updateAll(productReview, where);
   }
@@ -131,7 +132,7 @@ export class ProductReviewController {
   })
   async findById(
     @param.path.number('id') id: number,
-    @param.query.object('filter', getFilterSchemaFor(ProductReview)) filter?: Filter<ProductReview>
+    @param.filter(ProductReview, {exclude: 'where'}) filter?: FilterExcludingWhere<ProductReview>
   ): Promise<ProductReview> {
     return this.barRepository.findById(id, filter);
   }
@@ -191,6 +192,7 @@ import {
   Count,
   CountSchema,
   Filter,
+  FilterExcludingWhere,
   repository,
   Where,
 } from '@loopback/repository';
@@ -248,7 +250,7 @@ export class ProductReviewController {
     },
   })
   async count(
-    @param.query.object('where', getWhereSchemaFor(ProductReview)) where?: Where<ProductReview>,
+    @param.where(ProductReview) where?: Where<ProductReview>,
   ): Promise<Count> {
     return this.barRepository.count(where);
   }
@@ -269,7 +271,7 @@ export class ProductReviewController {
     },
   })
   async find(
-    @param.query.object('filter', getFilterSchemaFor(ProductReview)) filter?: Filter<ProductReview>,
+    @param.filter(ProductReview) filter?: Filter<ProductReview>,
   ): Promise<ProductReview[]> {
     return this.barRepository.find(filter);
   }
@@ -291,7 +293,7 @@ export class ProductReviewController {
       },
     })
     productReview: ProductReview,
-    @param.query.object('where', getWhereSchemaFor(ProductReview)) where?: Where<ProductReview>,
+    @param.where(ProductReview) where?: Where<ProductReview>,
   ): Promise<Count> {
     return this.barRepository.updateAll(productReview, where);
   }
@@ -310,7 +312,7 @@ export class ProductReviewController {
   })
   async findById(
     @param.path.number('id') id: number,
-    @param.query.object('filter', getFilterSchemaFor(ProductReview)) filter?: Filter<ProductReview>
+    @param.filter(ProductReview, {exclude: 'where'}) filter?: FilterExcludingWhere<ProductReview>
   ): Promise<ProductReview> {
     return this.barRepository.findById(id, filter);
   }
--- a/packages/openapi-v3/src/decorators/parameter.decorator.ts
+++ b/packages/openapi-v3/src/decorators/parameter.decorator.ts
@@ -4,6 +4,8 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {MetadataInspector, ParameterDecoratorFactory} from '@loopback/core';
+import {FilterSchemaOptions, Model} from '@loopback/repository-json-schema';
+import {getFilterSchemaFor, getWhereSchemaFor} from '../filter-schema';
 import {resolveSchema} from '../generate-schema';
 import {OAI3Keys} from '../keys';
 import {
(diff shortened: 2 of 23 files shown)


raymondfeng:
@dougal83 I added one more feature. PTAL.

packages/openapi-v3/src/decorators/parameter.decorator.ts
```diff
@@ -437,6 +443,70 @@ export namespace param {
...
+    name = 'filter',
+  ) {
+    return param.query.object(
+      name,
+      getFilterExcludingWhereSchemaFor(modelCtor),
+    );
+  }
+
+  /**
+   * Sugar decorator for `filter` query parameter excluding `where`
```
nabdelgadir: This decorator isn't excluding `where`?

raymondfeng: good catch - copy/paste error :-)

nabdelgadir:
@raymondfeng I'm a bit confused about why we're excluding `where` in `findById` calls. I understand from https://github.com/strongloop/loopback-next/issues/3336 that it's being ignored, but do we want to ignore it rather than allow restrictions through it on `findById` calls (e.g. returning `undefined` if it doesn't match the restriction)?

(also there's a typo in the second commit message, `excldue` -> `exclude`)

raymondfeng:
> I'm a bit confused about why we're excluding where in findById calls. I understand from #3336 that it's being ignored, but do we want to ignore it rather than allow restrictions through it on findById calls (e.g. returning undefined if it doesn't match the restriction)?

The rationale behind `findById` is to look up a record by its primary key (id). No other search criteria should be used. If a client wants to search beyond id, use `find` or `findOne` instead.

derdeka:
> The rationale behind `findById` is to look up a record by its primary key (id). No other search criteria should be used. If a client wants to search beyond id, use `find` or `findOne` instead.

@raymondfeng
Thanks for explanation. I did not know this before.
To clarify this for developers, maybe we can add this rationale here?:
https://github.com/strongloop/loopback-next/blob/2e52f355f95cee82b5401bfa2bcbbf6ccc5a91d2/packages/repository/src/repositories/repository.ts#L173-L181

packages/repository/src/repositories/repository.ts
```diff
@@ -303,7 +303,11 @@ export class CrudRepositoryImpl<T extends Entity, ID>
     );
   }
 
-  async findById(id: ID, filter?: Filter<T>, options?: Options): Promise<T> {
+  async findById(
+    id: ID,
+    filter?: FilterExcludingWhere<T>,
+    options?: Options,
+  ): Promise<T> {
```
achrinza: >
> To clarify this for developers, maybe we can add this rationale here?:
>

Should be doable as such:

```suggestion
  /**
   * {@inheritDoc EntityCrudRepository.findById}
   */
  async findById(
    id: ID,
    filter?: FilterExcludingWhere<T>,
    options?: Options,
  ): Promise<T> {
```

See: https://api-extractor.com/pages/tsdoc/tag_inheritdoc/

packages/cli/snapshots/integration/generators/controller.integration.snapshots.js
```diff
@@ -131,7 +133,7 @@ export class ProductReviewController {
   })
   async findById(
     @param.path.number('id') id: number,
-    @param.query.object('filter', getFilterSchemaFor(ProductReview)) filter?: Filter<ProductReview>
+    @param.query.object('filter', getFilterExcludingWhereSchemaFor(ProductReview)) filter?: FilterExcludingWhere<ProductReview>
```
bajtos: Have you considered using the `options` argument of `getFilterSchemaFor` and `@param.filter`, instead of introducing new functions?

I am concerned that if we need to add another modifier in the future (e.g. "exclude inclusion of related models"), then we will have to introduce four new APIs:
 - `getFilterSchemaExcludingInclusionFor`
 - `getFilterSchemaExcludingWhereExcludingInclusionFor`
 - `@param.filterExcludingInclusion`
 - `@param.filterExcludingWhereExcludingInclusion`

If we add yet another modifier, then we will have to add 8 more APIs, and so on.

I am thinking about the following use:

```suggestion
    @param.query.object('filter', getFilterFor(ProductReview, {excludeWhere: true})) filter?: FilterExcludingWhere<ProductReview>
```

packages/openapi-v3/src/decorators/parameter.decorator.ts
```diff
@@ -437,6 +443,70 @@ export namespace param {
...
+   * }
+   * ```
+   * @param modelCtor - Model class
+   * @param name - Custom name for the parameter, default to `filter`
+   *
+   */
+  export function filterExcludingWhere(
+    modelCtor: typeof Model,
+    name = 'filter',
+  ) {
```
bajtos: Similarly here, have you considered adding `options` argument to `@param.filter`, instead of introducing another decorator?

packages/repository/src/query.ts
```diff
@@ -221,6 +221,14 @@ export interface Filter<MT extends object = AnyObject> {
   include?: Inclusion[];
 }
 
+/**
+ * Filter without `where` property
+ */
+export type FilterExcludingWhere<MT extends object = AnyObject> = Omit<
+  Filter<MT>,
+  'where'
+>;
```
bajtos: I am not sure if it's worth introducing this type alias. IMO, `Omit` is a regular TypeScript tool that we are already using in other places (e.g. `Omit<Todo, 'id'>` when creating new model instances), what are you arguments against using `Omit<Filter<MyModel>, 'where'>`  instead of this new alias?

raymondfeng: Since the type is used for all generated CRUD controllers, I prefer to have a type alias to make it easier to understand.

packages/repository/src/repositories/repository.ts
raymondfeng: I added tsdocs to the base interface.

dougal83:
> Personally, I think it's a better design to allow callers of `getFilterSchemaFor` and `@param.filter` to provide `{excludeWhere: true}` when they want to remove `where` from the schema, I find it more extensible and future-proof.

@bajtos +1 If given a preference I would prefer to use this approach that you've given.  While either way will work it is neater to use an option.

raymondfeng:
@dougal83 The PR has been refined to address the review comments. PTAL.

packages/cli/snapshots/integration/generators/relation.belongs-to.integration.snapshots.js
```diff
@@ -7,9 +7,7 @@
 
 'use strict';
 
-exports[
-  `lb4 relation checks generated source class repository  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} generates Order repository file with different inputs 1`
-] = `
+exports[`lb4 relation checks generated source class repository  answers {"relationType":"belongsTo","sourceModel":"Order","destinationModel":"Customer"} generates Order repository file with different inputs 1`] = `
```
dougal83: nitpick: double space after `repository`?

raymondfeng: Sorry, I'll move this commit to a separate PR. The snapshot is generated. If we need to fix the prompt, we should fix `cli`.

raymondfeng: See https://github.com/strongloop/loopback-next/pull/4761

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -14,6 +14,11 @@ export interface FilterSchemaOptions {
    *
    */
   setTitle?: boolean;
+
+  /**
+   * To exclude one or more property from `filter`
+   */
+  exclude?: string[] | string;
```
bajtos: :+1:

I see that both `string[]` and `string` flavors are supported, that's great ergonomics! Could you please double check that we have test coverage for both of them? When I was skimming through the tests, I saw `string[]` variant only.

This is important because the CLI templates are using `string` flavor, we should verify that such flavor works as intended!

packages/repository-json-schema/src/filter-json-schema.ts
```diff
@@ -50,43 +55,61 @@ export function getScopeFilterJsonSchemaFor(
...
-    }),
-    properties: {
-      where: getWhereJsonSchemaFor(modelCtor, options),
+  let excluded: string[];
+  if (typeof options.exclude === 'string') {
+    excluded = [options.exclude];
+  } else {
+    excluded = options.exclude ?? [];
+  }
+  if (typeof excluded === 'string') excluded = [excluded];
```
bajtos: I think this last check is redundant, because `excluded` can be a string only when `options.exclude` was a string, which was already handled.

```suggestion
  let excluded: string[];
  if (typeof options.exclude === 'string') {
    excluded = [options.exclude];
  } else {
    excluded = options.exclude ?? [];
  }
```

packages/repository/src/query.ts
bajtos: Fair enough.

bajtos:
@raymondfeng Feel free to merge the PR after my comments are addressed, no need to wait for another review from me (unless you want to).

Issue #4573, opened 2020-02-06
feat(authentication): refreshing/replacing the token


derdeka:
## Suggestion

TTL of AccessToken (`JWT`, `LB3 legacy token` or any other token system) should be extendable or token should be replaced after some time of beeing used. AFAIK the `@loopback/authentication` component does not support this yet.

## Use Cases

Cross-Posting @sformisano 's comment from https://github.com/strongloop/loopback-next/issues/3673#issuecomment-560954366 as discussion point:
> Most JWT based auth systems do actually carry a refresh mechanism for the access token, either through a secondary refresh token that lasts longer than the access token or with a long-living access token that also works as a refresh token.
> 
> Based on how this seems to work at the moment in LB4, if I want to set a low TTL for the access token, which would be good for security, once the token expires, I would have to login again, even if I was using the app right when the access token expired.
> 
> I'm sure users having to log in every 10 minutes is not what the LB team is shooting for so why don't we review a few options to avoid that:
> 
> 1) The simplest approach: very long-lived access token with no refresh mechanism.
> This means creating access tokens with TTL set to weeks or months. As a consequence, the problem described above is vastly mitigated, but still not resolved. The user will still, at some point, be logged out, even if it uses the app every single day. The user could even be logged out while it is using the app because when the token does expire, the only way to get a new one is to log in again.
> 
> This solution is also not great from a security standpoint: by definition, an access token is a token that provides access to resources, and in a stateless authentication environment a token of this kind that lasts for months can be problematic. There are a number of ways to mitigate this type of issue as well, but they are not very efficient.
> 
> One example would be having an access tokens blacklist. Every time there's a request, the token would be verified against the blacklist. Running this type of check for every requested is not the most efficient way to solve this problem.
> 
> A user-level ban mechanism would have the same exact efficiency limitation.
> 
> As far as I can see, this is the only solution that seems to be available out of the box within @loopback/authentication, which is why I started looking for this kind of discussion.
> 
> 2) A better approach: long-lived access token with refresh mechanism.
> This solution would have a single access token with a considerable TTL, but not as long as for solution (1), perhaps a few days. The improvement over solution (1) is that the API would be responsible for refreshing the access token and sending them back to the client regularly. It would work like this:
> 
> Every time there's a request, access token TTL is checked by the API.
> 
> If the access token TTL falls below a certain threshold (e.g. 50% of original TTL), a new access token would be created and automatically sent to the client.
> 
[...]

PR #4746, opened 2020-02-25
feat(authentication): extend TokenService for revokeable tokens

derdeka:
When implementing token systems other then `jwt` there is a need to **revoke a token**. I also discovered that there is a need to access the `UserProfile` of the current user to validate tokens in some token systems.

Example usecases:
 - `RefreshToken`, e.g https://github.com/strongloop/loopback4-example-shopping/pull/537
 - Reimplementing `lb3` `AccessToken` in `lb4`

Related to #4573.

With this PR I'm suggesting to change the `TokenService` interface.
~~This is a BREAKING CHANGE.~~

WDYT?

## Checklist

### DIFF
--- a/packages/authentication/src/services/token.service.ts
+++ b/packages/authentication/src/services/token.service.ts
@@ -11,10 +11,28 @@ import {UserProfile} from '@loopback/security';
 export interface TokenService {
   /**
    * Verifies the validity of a token string and returns a user profile
+   *
+   * @param token The token/secret which should be validated/verified.
+   *
+   * @returns The UserProfile which belongs to the given token.
    */
   verifyToken(token: string): Promise<UserProfile>;
+
   /**
    * Generates a token string based on a user profile
+   *
+   * @param userProfile A UserProfile for which a token should be generated.
+   *
+   * @returns a generated token/secret for a given UserProfile.
    */
   generateToken(userProfile: UserProfile): Promise<string>;
+
+  /**
+   * Revokes a given token (if supported by token system)
+   *
+   * @param token The token/secret which should be revoked/invalidated.
+   *
+   * @returns true, if the given token was invalidated.
+   */
+  revokeToken?(token: string): Promise<boolean>;
 }


packages/authentication/src/services/token.service.ts
```diff
@@ -11,10 +11,32 @@ import {UserProfile} from '@loopback/security';
...
+  /**
+   * Revokes a given token (if supported by token system)
+   *
+   * @param token The token/secret which should be revoked/invalidated.
+   * @param userProfile An optional UserProfile to check if the token
+   *   is valid for a given User.
+   *
+   * @returns true, if the given token was invalidated.
+   */
+  revokeToken(token: string, userProfile?: UserProfile): Promise<boolean>;
```
raymondfeng: What about make it optional? `revokeToken?(...)`?

derdeka: Thanks for your feedback. I didn't think of this simple solution - this makes the change non-breaking, but more complex in using it. How you would check for existance of this method?

I'm also not sure if it's better to return an boolean or just throw an exception if a token can not be invalidated. In the light of making revokeToken optional, do you suggest to just throw?

raymondfeng: I was wondering about that too. I guess `Promise<boolean>` is fine for now. If `revokeToken` is not present, we can default to `async() => false`.

derdeka: Changed to be optional.

packages/authentication/src/services/token.service.ts
```diff
@@ -11,10 +11,32 @@ import {UserProfile} from '@loopback/security';
...
    * Verifies the validity of a token string and returns a user profile
+   *
+   * @param token The token/secret which should be validated/verified.
+   * @param userProfile Depending on the token system this optional parameter
+   *   can be used to check if the token is valid for the given user.
+   *
+   * @returns The UserProfile which belongs to the given token.
    */
-  verifyToken(token: string): Promise<UserProfile>;
+  verifyToken(token: string, userProfile?: UserProfile): Promise<UserProfile>;
```
raymondfeng: In what case the user profile is available for the token is verified?

derdeka: For example for `refreshToken` use case, its required to verify if the `refreshToken` if it belongs to the current user (which was verified by the `accessToken` before).

Here is an example:
https://github.com/strongloop/loopback4-example-shopping/blob/980a065224a56fd7dbd6b512519c944bd0415991/packages/shopping/src/controllers/user.controller.ts#L301-L309

nflaig: @derdeka why are you validating the jwt when you are doing a refresh? The only reason to do a refresh of the access token is if it is expired in which case then `@authenticate(jwt)` throws an error anways

derdeka: @nflaig
Thanks for your feedback. I assumed it would be a good idea to extend the `accessToken` using the `refreshToken` as long as it is still valid with the benefit of improved security.

I would like to ask you for a review of https://github.com/strongloop/loopback4-example-shopping/pull/537 to have a look at the whole refresh process.

raymondfeng: Refresh token itself is subject to verification. I don't expect the `refresh` to require an valid access token.

nflaig: @derdeka the current implementation would mean the user has a session timeout if he is inactive for 10mins becasue of the access token expiration which is probably not the expected behaviour.

I would create a relation between a user which can have many refresh tokens e.g. one for each device he is logged in. The refresh token value should probably not be the database Id but rather be generated in a [secure way](https://gist.github.com/joepie91/7105003c3b26e65efcea63f3db82dfba). This means if a request to the refresh token endpoint happens you just look up the refresh token and if it exists and has not expired you can issue a new access token for the user to whom the refresh token belongs to. The refresh token is similiar to traditional session tokens which are basically just a secure random and unguessable string of characters.

derdeka: Feedback applied.

derdeka:
@strongloop/loopback-maintainers PTAL

derdeka:
@raymondfeng May i ask for your approvement again to get this landed?

agnes512:
👏

PR #4762, opened 2020-02-28
add async validation support

mastermunj:
Add support for async validations. The need arises from allowing custom formats & keywords for ajv, which can also be async in nature.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/request-body.validator.test.ts
+++ b/packages/rest/src/__tests__/unit/request-body.validator.test.ts
@@ -51,17 +51,17 @@ const INVALID_ACCOUNT_SCHEMA = {
 };
 
 describe('validateRequestBody', () => {
-  it('accepts valid data omitting optional properties', () => {
-    validateRequestBody(
+  it('accepts valid data omitting optional properties', async () => {
+    await validateRequestBody(
       {value: {title: 'work'}, schema: TODO_SCHEMA},
       aBodySpec(TODO_SCHEMA),
     );
   });
 
   // Test for https://github.com/strongloop/loopback-next/issues/3234
-  it('honors options for AJV validator caching', () => {
+  it('honors options for AJV validator caching', async () => {
     // 1. Trigger a validation with `{coerceTypes: false}`
-    validateRequestBody(
+    await validateRequestBody(
       {
         value: {city: 'San Jose', unit: 123, isOwner: true},
         schema: ADDRESS_SCHEMA,
@@ -72,7 +72,7 @@ describe('validateRequestBody', () => {
     );
 
     // 2. Trigger a validation with `{coerceTypes: true}`
-    validateRequestBody(
+    await validateRequestBody(
       {
         value: {city: 'San Jose', unit: '123', isOwner: 'true'},
         schema: ADDRESS_SCHEMA,
@@ -83,7 +83,7 @@ describe('validateRequestBody', () => {
     );
 
     // 3. Trigger a validation with `{coerceTypes: false}` with invalid data
-    expect(() =>
+    await expect(
       validateRequestBody(
         {
           value: {city: 'San Jose', unit: '123', isOwner: true},
@@ -93,10 +93,10 @@ describe('validateRequestBody', () => {
         {},
         {coerceTypes: false},
       ),
-    ).to.throw(/The request body is invalid/);
+    ).to.be.rejectedWith(/The request body is invalid/);
   });
 
-  it('rejects data missing a required property', () => {
+  it('rejects data missing a required property', async () => {
     const details: RestHttpErrors.ValidationErrorDetails[] = [
       {
         path: '',
@@ -105,7 +105,7 @@ describe('validateRequestBody', () => {
         info: {missingProperty: 'title'},
       },
     ];
-    verifyValidationRejectsInputWithError(
+    await verifyValidationRejectsInputWithError(
       INVALID_MSG,
       'VALIDATION_FAILED',
       details,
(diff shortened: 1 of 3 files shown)


packages/rest/src/__tests__/unit/request-body.validator.test.ts
```diff
@@ -83,20 +83,23 @@ describe('validateRequestBody', () => {
...
         aBodySpec(ADDRESS_SCHEMA),
         {},
         {coerceTypes: false},
-      ),
-    ).to.throw(/The request body is invalid/);
+      );
+    } catch (error) {
+      expect(error.message).to.equal(INVALID_MSG);
+      expect(error.code).to.equal('VALIDATION_FAILED');
+    }
```
raymondfeng: We use `await expect(aPromise).to.be.rejectedWith(...)` style.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -142,13 +142,22 @@ function validateValueAgainstSchema(
     cache.set(schema, validatorMap);
   }
 
-  if (validate(body)) {
+  let validationResult = validate(body);
+  let validationErrors: AJV.ErrorObject[] = [];
+  try {
+    validationResult = await validationResult;
```
raymondfeng: It's not very friendly to change the type of `validationResult` in the same function.

packages/rest/src/__tests__/unit/request-body.validator.test.ts
mastermunj: Done!

packages/rest/src/validation/request-body.validator.ts
mastermunj: Thanks for pointing it out. I've simplified the logic. Is new one better & clear?

mastermunj:
Uh oh! I messed it up it seems.

dougal83:
@mastermunj Don't worry if you make mistakes with Git, we've all been there.  There are plenty of very helpful people who can help you fix what seems like a disaster.  It's source control... most problems can be fixed.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -142,13 +142,18 @@ function validateValueAgainstSchema(
...
-    debug('Request body passed AJV validation.');
-    return;
+  let validationErrors: AJV.ErrorObject[] = [];
+  try {
+    const validationResult = await validate(body);
+    // When schema is not available, ajv returns null
+    if (validationResult || validationResult === null) {
+      debug('Request body passed AJV validation.');
+      return;
+    }
```
achrinza: Seems like this branch isn't fully covered with tests yet (hence the reduced test coverage).

Is there a case where the schema isn't available when AJV is called?

It'll be great to add a test case to prevent regression.

Here's the [Coveralls report](https://coveralls.io/builds/29048569/source?filename=packages%2Frest%2Fsrc%2Fvalidation%2Frequest-body.validator.ts#L149)

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -33,7 +33,7 @@ const ajvErrors = require('ajv-errors');
  * @param globalSchemas - The referenced schemas generated from `OpenAPISpec.components.schemas`.
  * @param options - Request body validation options for AJV
  */
-export function validateRequestBody(
+export async function validateRequestBody(
```
achrinza: Should this be considered a breaking change?

packages/rest/src/validation/request-body.validator.ts
mastermunj: Hi @achrinza, [this](https://github.com/strongloop/loopback-next/pull/4762/files#diff-abbe6f4cfbede11c53ba54c636a7b24dR209-R214) test case doesn't pass without `validationResult === null` as promise returned by `validate` returns null.

My comment there seems misleading with `schema is not available`. Schema is available but body is optional and value is empty.

packages/rest/src/validation/request-body.validator.ts
mastermunj: Yes, it should be considered a breaking change.

Reference: https://github.com/strongloop/loopback-next/pull/4658#discussion_r382607886

achrinza: @mastermunj Could we amend the commit message to include the `BREAKING CHANGE` flag? This can be done with `git commit --amend`

Suggested commit message:
```
feat(rest): add async validation support

BREAKING CHANGE: `validateRequestBody` is now async

feat(rest): simplified ajv validation promise use

feat(rest): one more improvement
```

This'll ensure a semver major bump.

mastermunj: @achrinza, I have amended the commit message as you suggested.

raymondfeng:
@mastermunj I suggest you further refine the commit message as follows:

feat(rest): add async validation support

BREAKING CHANGE: `validateRequestBody` is now an async function to allow asynchronous validations by custom Ajv keywords and formats. See https://ajv.js.org/#asynchronous-validation
for more details.

mastermunj:
@raymondfeng modified commit message as per your suggestion.

packages/rest/src/validation/request-body.validator.ts
mastermunj: @achrinza did you get a chance to look into this?

achrinza: @mastermunj I would suggest editing the comment to be more accurate to the purpose of the conditional statement.

Though @raymondfeng has approved the PR and this issue is a non-blocker for me.

bajtos: > My comment there seems misleading with `schema is not available`. Schema is available but body is optional and value is empty.

I would really like the commit to be updated before this pull request is landed, to avoid confusion in the future.

dhmlau:
@slnode test please

PR #4773, opened 2020-03-02
docs: update relations' api docs

agnes512:
Updates  API docs to match with currently design.

## Checklist

### DIFF
--- a/docs/site/HasOne-relation.md
+++ b/docs/site/HasOne-relation.md
@@ -407,8 +407,8 @@ factory `account` for instances of `SupplierRepository`:
 - `create` for creating a target model instance belonging to `Supplier` model
   instance
   ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.create.html))
-- `find` finding target model instance belonging to Supplier model instance
-  ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.find.html))
+- `get` gets target model instance belonging to Supplier model instance
+  ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.get.html))
 - `delete` for deleting target model instance belonging to Supplier model
   instance
   ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.delete.html))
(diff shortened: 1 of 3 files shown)


docs/site/HasOne-relation.md
```diff
@@ -407,8 +407,8 @@ factory `account` for instances of `SupplierRepository`:
 - `create` for creating a target model instance belonging to `Supplier` model
   instance
   ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.create.html))
-- `find` finding target model instance belonging to Supplier model instance
-  ([API Docs](https://loopback.io/doc/en/lb4/apidocs.repository.hasonerepository.find.html))
+- `get` finding target model instance belonging to Supplier model instance
```
derdeka:
```suggestion
- `get` gets target model instance belonging to Supplier model instance
```

packages/repository/src/relations/belongs-to/belongs-to.repository.ts
```diff
@@ -17,6 +17,8 @@ export interface BelongsToRepository<Target extends Entity> {
   /**
    * Gets the target model instance
    * @param options
+   * @returns A promise of the target object, or throws an EntityNotFound error
+   * when target model instance not found.
```
raymondfeng: Does it throw an error or return a Promise that rejects with the error?

packages/repository/src/relations/has-one/has-one.repository.ts
```diff
@@ -34,7 +34,8 @@ export interface HasOneRepository<Target extends Entity> {
    * Find the only target model instance that belongs to the declaring model.
    * @param filter - Query filter without a Where condition
    * @param options - Options for the operations
-   * @returns A promise of the target object or null if not found.
+   * @returns A promise of the target object, or throws an EntityNotFound error
+   * when target model instance not found.
```
raymondfeng: Ditto.

packages/repository/src/relations/belongs-to/belongs-to.repository.ts
agnes512: Returns 
- A promise of the target object, ( if found)
-  or throws an EntityNotFound error when target model instance not found ( not found)

How about :
`returns A promise of the target object, otherwise throws...`?

bajtos: I believe this API is implemented as an `async` function, therefore the `EntitNotFound` error thrown by the code is automatically converted to a rejected promise by JS runtime.

bajtos: How about the following:

> A promise resolved with the target object or rejected with an `EntityNotFoundError`
when target model instance was not found.

PR #1058, opened 2018-02-27
 fix(rest): log unexpected errors to console

bajtos:
This is a follow-up and a partial revert of #939.

The first commit adds a check to verify that no console logs are produced by our tests, beyond what Mocha prints. This is needed to ensure clean test output in the future.

The second commit reverts the default implementation of `LogError` sequence action to use `console.error` again (instead of `debug`) and fixes tests to suppress error messages for requests that fail with the expected error.

## Checklist

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- Related API Documentation was updated
- Affected artifact templates in `packages/cli` were updated
- Affected example projects in `packages/example-*` were updated

PR #4780, opened 2020-03-03
Add more details about rejecting console logs in tests 

bajtos:
As a follow-up to the conversation we had on Slack, I am proposing to add some documentation for `ATTENTION - INVALID USAGE OF CONSOLE LOGS DETECTED` and make this docs easier to discover by modifying `lb-mocha` to print a URL for that content.

See also https://github.com/strongloop/loopback-next/pull/1058 which introduced log detection to `lb-mocha`.

## Checklist

### DIFF
--- a/packages/build/README.md
+++ b/packages/build/README.md
@@ -132,6 +132,36 @@ npm run build
 
     `converage:ci` sets up integration with [Coveralls](https://coveralls.io/).
 
+## A note on console logs printed by tests
+
+We consider (console) logging from tests as a bad practice, because such logs
+usually clutter the test output and make it difficult to distinguish legitimate
+error messages from the noise.
+
+By default, `lb-mocha` detects when the tests and/or the application tested have
+printed console logs and fails the test run with the following message:
+
+```
+=== ATTENTION - INVALID USAGE OF CONSOLE LOGS DETECTED ===
+```
+
+If you need more information about behavior in the test, then the first choice
+should be to use a better or more descriptive error assertion. If that's not
+possible, then use debug statements to print additional information when
+explicitly requested.
+
+A typical situation is that a test is sending an HTTP request and the server
+responds with an error code as expected. However, because the server is
+configured to log failed requests, it will print a log also for requests where
+the failure was expected and intentional. The solution is to configure your REST
+server to suppress error messages for that specific error code only. Our
+`@loopback/testlab` module is providing a helper
+[`createUnexpectedHttpErrorLogger`](https://github.com/strongloop/loopback-next/tree/master/packages/testlab#createUnexpectedHttpErrorLogger)
+that makes this task super easy.
+
+Alternatively, it's also possible to disable detection of console logs by
+calling `lb-mocha` with `--allow-console-logs` argument.
+
 ## Contributions
 
 - [Guidelines](https://github.com/strongloop/loopback-next/blob/master/docs/CONTRIBUTING.md)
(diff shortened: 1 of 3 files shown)


packages/build/src/fail-on-console-logs.js
```diff
@@ -54,7 +54,8 @@ process.on('exit', code => {
   if (!problems.length) return;
   const log = originalConsole.log;
 
-  log('\n=== ATTENTION - INVALID USAGE OF CONSOLE LOGS DETECTED ===\n');
+  log('\n=== ATTENTION - INVALID USAGE OF CONSOLE LOGS DETECTED ===');
+  log('Learn more at http://ibm.biz/lb-mocha-console-logs\n');
```
achrinza: I think a GitHub permalink be better:

1. Docs are accurate for the specific version of `@loopback/build`

    The docs may change over time to reflect functional changes. This means that the latest version of the docs may not be accurate for the version of `@loopback/build` that is used by the user.

2. The link can be updated when the docs are updated

    This ties in with the first point - when the docs are updated, the link can be updated to the new permalink. This would ensure that the users are looking at the most accurate version of the docs.

raymondfeng: I prefer a `github` or `loopback.io` link too.

packages/build/README.md
```diff
@@ -132,6 +132,36 @@ npm run build
 
     `converage:ci` sets up integration with [Coveralls](https://coveralls.io/).
 
+## A note on console logs printed by tests
+
+We consider (console) logging from tests as a bad practice, because such logs
+usually clutter the test output and make it difficult to distinguish legitimate
+error messages from the noise.
+
+By default, `lb-mocha` detects when the tests and/or the application tested is
```
nabdelgadir: `is` -> `has`

packages/build/README.md
```diff
@@ -132,6 +132,36 @@ npm run build
...
+printed console logs and fails the test run with the following message:
+
+```
+=== ATTENTION - INVALID USAGE OF CONSOLE LOGS DETECTED ===
+```
+
+If your need more information about behavior in the test, then the first choice
```
nabdelgadir: `your` -> `you`

packages/build/src/fail-on-console-logs.js
nabdelgadir: +1 to a GitHub link. (Also, the current link adds a single quote at the end for me: https://github.com/strongloop/loopback-next/blob/master/packages/build/README.md#a-note-on-console-logs-printed-by-tests')

packages/build/README.md
bajtos: Good catch!

I am going to use `have`, because there are multiple subjects involved (`the tests and/or the application`).

packages/build/src/fail-on-console-logs.js
bajtos: Thank you for the feedback, I agree with your arguments and will change the link to point to GitHub `master` branch.

bajtos:
Thank you @achrinza @raymondfeng @agnes512 for valuable feedback ❤️ 

I added a commit to fix the issues, does the patch LGTY now?

raymondfeng:
@bajtos Commit lint is failing - https://travis-ci.com/strongloop/loopback-next/jobs/294887891

nabdelgadir:
> @bajtos Commit lint is failing - https://travis-ci.com/strongloop/loopback-next/jobs/294887891

It errored out so I restarted it and it's passing now 👍 

It says the coverage dropped, but this PR isn't adding new code (besides one line of text). 😕

Issue #4296, opened 2019-12-12
How to build models, repositories and controllers dynamically at runtime


bajtos:
_This is a follow-up for the spike https://github.com/strongloop/loopback-next/pull/4235_

So far, our documentation and tooling was focused on the use case where most of application artifacts are static, defined in source files. In the Epic _From model definition to REST API #2036_, we are working on a more declarative approach, where we have a model source file and a configuration file, and the bootstrapper creates the repository & controller classes during the boot process.

In this story, I would like to show our users even more flexible and dynamic approach, where they have full control of all aspects at runtime - from defining the model class (e.g. based on schema discovery) to exposing the REST API (or decide not to).

## Examples

See the demo application presented in https://github.com/strongloop/loopback-next/pull/4235:

> 1. Accept a MySQL connection string (host+port+database+credentials) and a list of table names.
> 2. Connect to the specified MySQL database.
> 3. Discover model schemas from the given tables
> 4. Define LB4 models for the given tables
> 5. Expose these models via REST API
> 6. Return a list of URL paths of the newly added models (e.g. `/Products`, `/CoffeeShops`)

## Acceptance criteria

A new documentation page with the following sections & content:

**MUST HAVE**

- [x] Define a model class from a `ModelDefinition` object
- [x] Define a repository class for the given model, how to configure dependencies (the datasource to use)
- [x] Define a controller class for the given model & repository, how to configure dependencies (the repository to use) and how to build the REST API path (e.g. `/products` for `Product` model).
- [x] Discover model definition from a database, build `ModelDefinition` object from the discovered schema in juggler/LB3 format.

The page can use code snippets from the demo controller (see #4235) to illustrate the approach in practice.

**SHOULD HAVE**

_These items are not strictly required for the initial version. If you decide to leave them out, then create follow-up stories and/or move them to acceptance criteria of and existing story like #2740._

- [x] How to use a custom Model base class (e.g. `UserBase` provided by the app) instead of `Model` or `Entity`.
[...]

PR #4792, opened 2020-03-04
feat: move functions to define repository classes from @loopback/rest-crud to @loopback/repository

raymondfeng:
The PR is inspired by:

- https://github.com/strongloop/loopback-next/pull/4235#discussion_r354206659
- We already have `defineModelClass` in `@loopback/repository`

~~For `@loopback/rest-crud`:~~

~~BREAKING CHANGE: `defineCrudRepositoryClass` is now removed from
`@loopback/rest-crud`. Please use `defineEntityCrudRepositoryClass` from
`@loopback/repository` instead.~~

See some context at https://github.com/strongloop/loopback-next/issues/4296

## Checklist


packages/repository/src/repository-builder.ts
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @param entityClass - An entity class such as `Product`.
+ * @param baseRepositoryClass - Base repository class. Defaults to `DefaultCrudRepository`
+ *
+ * @typeParam E - An entity class
+ * @typeParam IdType - ID type for the entity
+ * @typeParam Relations - Relations for the entity
+ */
+export function defineEntityCrudRepositoryClass<
+  T extends Entity,
```
nabdelgadir: Can you change `T` to `E` for consistency with the docs and the interface below?

packages/repository/src/repository-builder.ts
```diff
@@ -0,0 +1,163 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest-crud
```
nabdelgadir: // Copyright IBM Corp. 2020. All Rights Reserved.
// Node module: @loopback/repository

packages/repository/src/__tests__/unit/repositories/define-crud-repository-class.unit.ts
```diff
@@ -0,0 +1,121 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest-crud
```
nabdelgadir: // Copyright IBM Corp. 2020. All Rights Reserved.
// Node module: @loopback/repository

bajtos:
> BREAKING CHANGE: `defineCrudRepositoryClass` is now removed from
`@loopback/rest-crud`. Please use `defineEntityCrudRepositoryClass` from
`@loopback/repository` instead.

Have you considered preserving backwards compatibility by adding re-export of `defineCrudRepositoryClass`to `@loopback/rest-crud`?

> I'm confused about the context of the move (I read #4296). Can you explain it a bit more?

+1, can you please @raymondfeng shed more light on your motivations?

Personally, I did consider to move `defineEntityCrudRepositoryClass` from `@loopback/rest-crud` to  `@loopback/repository` in the past too. However, I think we will need to make this function more generic in the (near?) future and allow the caller to provide their own repository class to inherit from. I am not sure if such generic function belongs to the repository package?

Let's discuss.

packages/repository/src/repository-builder.ts
```diff
@@ -0,0 +1,163 @@
...
+ * ```
+ *
+ * @param modelClass - A model class such as `Address`.
+ *
+ * @typeParam M - Model class
+ */
+export function defineCrudRepositoryClass<M extends Model>(
+  modelClass: typeof Model & {prototype: M},
+  baseRepositoryClass: Constructor<CrudRepository<M>>,
+): CrudRepositoryClass<M> {
```
bajtos: IMO, we need to make the interface of the repository class configurable. Ideally, `defineEntityCrudRepositoryClass` should delegate most of the implementation to `defineCrudRepositoryClass`.

A mock-up, I am not sure if it will work:

```ts

export function defineCrudRepositoryClass<
  M extends Model, 
  R extends Constructor<CrudRepository<M>>
>(
   modelClass: typeof Model & {prototype: M},
   baseRepositoryClass: R,
 ): R {
  // ...
}

export function defineEntityCrudRepositoryClass<
   T extends Entity,
   IdType,
   Relations extends object = {},
   R extends Constructor<EntityCrudRepository<T, IdType, Relations>>
    = Constructor<EntityCrudRepository<T, IdType, Relations>
 >(
   entityClass: typeof Entity & {prototype: T},
   baseRepositoryClass: R = DefaultCrudRepository as Constructor<
     EntityCrudRepository<T, IdType, Relations>
   >,
 ): R {
  return defineCrudRepositoryClass<M, R>(entityClass, baseRepositoryClass);
}
```

A test case to verify that extra API added by custom base classes are correctly propagated at compile time:

```ts
class TitledEntityCrudRepository<
   T extends Entity & {title: string},
   IdType,
   Relations extends object = {},
> extends DefaultCrudRepository<T, IdType, Relations> {
  findByTitle(title: string) {
    return this.find({where: {title}});
  }
}

@model()
class Note extends Entity {
  @property({id: true})
  id: number;

  @property()
  title: string;
}

const NoteRepository = defineEntityCrudRepositoryClass(Note, TitledEntityCrudRepository);
const db = new juggler.DataSource({connector: 'memory'});
const repo = new NoteRepository(db);

const created = await repo.create({title: 'welcome'});
const found = await repo.findByTitle('welcome');
expect(found).to.deepEqual(created);
```

packages/rest-crud/src/repository-builder.ts
```diff
@@ -1,55 +0,0 @@
...
-    'EntityCtor',
-    'BaseRepository',
-    `return class ${repoName} extends BaseRepository {
-      constructor(dataSource) {
-        super(EntityCtor, dataSource);
-      }
-    };`,
-  );
-
-  // TODO(bajtos) make DefaultCrudRepository configurable (?)
```
bajtos: The idea is to allow the caller of `defineCrudRepositoryClass` to provide the base class to inherit from. For example, if they want to use their own class `AuditedEntityCrudRepository` as the repository class to be used by all models.

raymondfeng: Yes.

raymondfeng:
@bajtos @nabdelgadir PTAL

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -0,0 +1,121 @@
...
+    street: string;
+
+    @property()
+    city: string;
+
+    @property()
+    state: string;
+  }
+
+  class MockCrudRepository<M extends Model> implements CrudRepository<M> {
```
bajtos: There is a subtle difference between mocks and stubs.

Typically, mocks are used when the class implementation is provided entirely by the mocking framework and we want to assert on what mocked methods were called & with what arguments.

```ts
const mock = sinon.mock(myApi);

// run the test invoking mocked methods

sinon.assert.calledOnce(mock.create);
```

Stubs are classes providing a dummy hand-written implementations of the given interface, which is the case of your test file.

Please rename `MockCrudRepository` to `CrudRepositoryStub` or even `DummyCrudRepository`.

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Signature for CrudRepository classes
+ *
+ * @typeParam M - Model class
+ * @typeParam R - Repository class/interface
+ */
+export interface CrudRepositoryClass<
+  M extends Model,
+  R extends CrudRepository<M>
+> {
+  new (ds: juggler.DataSource): R;
```
bajtos: A typical "base" repository class requires two arguments:
- The datasource to use
- The model class to persist

Custom base repository implementations may require more arguments, e.g. a third `options` argument.

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @typeParam M - Model class
+ * @typeParam R - Repository class/interface
+ */
+export function defineCrudRepositoryClass<
+  M extends Model,
+  R extends CrudRepository<M>
+>(
+  modelClass: typeof Model & {prototype: M},
+  baseRepositoryClass: Constructor<R>,
```
bajtos: I find this construct confusing.

First, you say that `R extends CrudRepository<M>`, meaning that `R` must provide a constructor accepting a single argument `ds: juggler.DataSource`.

Next, you say that `baseRepositoryClass` can be a class extending `R` and having arbitrary constructor.

Why are we pinning down the allowed constructor arguments in `R` then? AFAICT, we never rely on that list of known arguments because it's always replaced with the arbitrary list of args described by `Constructor`.

Can you please take a look at the approach we described for repository mixins in https://github.com/strongloop/loopback-next/pull/4723? I think it would be good to find a single solution that we can use consistently in all places.

Have you considered an API like shown in the following snippet?

```ts
export function defineCrudRepositoryClass<<
   M extends Model,
   R extends Constructor<CrudRepository<M>>
 >(
  modelClass: typeof Model & {prototype: M},
  baseRepositoryClass,
) {
  // ...
}
```

Considering how this factory function works, I think we should strongly consider providing strong type checks for the repository constructor arguments supported by the factory function and also for the class returned by the factory:

1. The base repository must accept `(ModelCtor, dataSource)` (and exactly in this order).
2. The class returned by the factory accepts a single `(ModelCtor)` argument.

In the future, we can allow additional constructor arguments (`(ModelCtor, dataSource, ...args)` on input and `(ModelCtor, ...args)` in output), but let's leave that out of scope of this PR - it's already too complex.

packages/rest-crud/README.md
```diff
@@ -64,7 +64,7 @@ class defined without the need for a repository or controller class file.
 
 If you would like more flexibility, e.g. if you would only like to define a
 default `CrudRest` controller or repository, you can use the two helper methods
-(`defineCrudRestController` and `defineCrudRepositoryClass`) exposed from
+(`defineCrudRestController` and `defineEntityCrudRepositoryClass`) exposed from
 `@loopback/rest-crud`. These functions will help you create controllers and
```
bajtos: IIUC the scope of the change, then this information is no longer correct, because `defineEntityCrudRepositoryClass` is no longer exported by `@loopback/rest-crud`.

BTW are you sure the benefits of removing `defineCrudRepositoryClass` from `@loopback/rest-crud` are worth the costs of publishing a semver-major version so shortly after the previous one?

Have you considered adding the following line to `packages/rest-crud/src/index.ts`?

```ts
// Perserve backwards compatibility
export {defineEntityCrudRepositoryClass as defineCrudRepositoryClass} from '@loopback/repository';
```

packages/repository/src/define-repository-class.ts
bajtos: OK, now I see that you moved this interface from my original implementation.

It's important to keep in mind that in the original version, this interface was used to describe the type returned by the factory function, it's a repository bound to a given model.

When reviewing the code below, I got confused because this distinction was not clear. In other places, we use `FooClass` as an alias for `typeof Foo`, but here `CrudRepositoryClass` is not necessarily the same thing as `typeof CrudRepository`.

Can we please find a name for this interface which will better communicate the fact that it's describing a class returned by one of the factory functions below?

packages/repository/src/define-repository-class.ts
bajtos: Uh oh, after reading the second commit, I realized that I got confused and the following part of my comment is not true:

> First, you say that `R extends CrudRepository<M>`, meaning that R must provide a constructor accepting a single argument `ds: juggler.DataSource`.

Your code is saying that `R` can have arbitrary constructor.

Which is still a problem, because we need `R` to accept `(ModelCtor, dataSource)` (and exactly in this order).

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,156 @@
...
+): Constructor<R> {
```
bajtos: I think this should be `CrudRepositoryClass<M>` to communicate the fact that the class constructor accepts only a single `dataSource` argument?

packages/rest-crud/README.md
bajtos: Never mind, I see that `@loopback/rest-crud` has version `0.7.0` now, so breaking changes are not a problem. 

Let's make sure the documentation is correct, no need to preserve the old API.

raymondfeng: Since `@loopback/rest-crud` is an experimental package (0.x.y), I chose to make it a breaking change. But I'm not strongly against the backward compatibility.

raymondfeng:
@bajtos I believe the latest PR has addressed your comments. PTAL.

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -0,0 +1,121 @@
...
+    it('should generate CRUD repository class', async () => {
+      const AddressRepository = defineCrudRepositoryClass(
+        Address,
+        MockCrudRepository,
+      );
+
+      expect(AddressRepository.name).to.equal('AddressRepository');
+      expect(Object.getPrototypeOf(AddressRepository)).to.equal(
+        MockCrudRepository,
+      );
```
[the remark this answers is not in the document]
bajtos: I am afraid my comment above does not seem addressed. As I understand the checks above, they are all performed at runtime. We don't know if the TypeScript compiler will allow us to call custom methods provided by the repository base class passed to `defineCrudRepositoryClass` function. Please note this is not just a theoretical concern, we have run into compile-time errors when working on the migration guide for Operation Hooks together with @emonddr.

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -0,0 +1,130 @@
...
+} from '../../..';
+
+describe('RepositoryClass builder', () => {
+  describe('defineCrudRepositoryClass', () => {
+    it('should generate CRUD repository class', async () => {
+      const AddressRepository = defineCrudRepositoryClass(
+        Address,
+        DummyCrudRepository,
+      );
+      expect(AddressRepository.prototype.find).to.be.a.Function();
```
bajtos: Nice, I see that `AddressRepository.prototype` is typed by TypeScript as `CrudRepositoryClass<Address, DummyCrudRepository<any>>.prototype: DummyCrudRepository<any>` and the `find` method has the right type too: `DummyCrudRepository<any>.find(filter?: Filter<any> | undefined, options?: AnyObject | undefined): Promise<any[]>`.

I guess that's a good alternative how to verify that DummyRepository's method can be called.

Do you think the check is enough to verify that custom repository methods (e.g. `DummyCrudRepository#findByTitle`) will be included in `AddressRepository` prototype too? I would prefer to verify this explicitly: Add `findByTitle` method to `DummyCrudRepository`, verify `AddressRepository.prototype.findById` is a function.

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,192 @@
...
+>(
+  entityClass: typeof Entity & {prototype: E},
+  baseRepositoryClass: BaseRepositoryClass<
+    E,
+    EntityCrudRepository<E, IdType, Relations>
+  > = (DefaultCrudRepository as unknown) as BaseRepositoryClass<
+    E,
+    EntityCrudRepository<E, IdType, Relations>
+  >,
+): EntityCrudRepositoryClass<E, IdType, Relations> {
```
bajtos: I think this is not going to support additional methods that could be provided by `baseRepositoryClass` on top of `EntityCrudRepository` interface (e.g. `findByTitle`). A possible solution is to introduce another generic argument `R`, similarly to what `defineCrudRepositoryClass` does.

Please make sure to add a test to verify this use case.

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * @param entityClass - An entity class such as `Product`.
+ * @param baseRepositoryClass - Base KeyValue repository class.
+ * Defaults to `DefaultKeyValueRepository`
+ *
+ * @typeParam E - Entity class
+ * @typeParam R - KeyValueRepository class/interface
+ */
+export function defineKeyValueRepositoryClass<
+  E extends Entity,
```
bajtos: Please not that KeyValue repository does not require Entities, it can be used with any model.

https://github.com/strongloop/loopback-next/blob/c016550f51f2f0db74527b96b1ad06d056496448/packages/repository/src/repositories/kv.repository.ts#L25

This was an intentional decision on my side when implementing LB3 key-value APIs, to allow the `key` value to be provided outside of the data stored, thus enabling use cases like storing arbitrary data (including binary blobs) via the key-value API.

packages/rest-crud/README.md
bajtos: I think the docs here does not match the actual implementation below.

The implementation:

```ts
// Reexport `defineEntityCrudRepositoryClass` from `@loopback/repository` as
 // `defineCrudRepositoryClass` for backward-compatibility
 export {defineEntityCrudRepositoryClass as defineCrudRepositoryClass} from '@loopback/repository';
```

The docs say:

```markdown
(`defineCrudRestController` and `defineEntityCrudRepositoryClass`) exposed from
 `@loopback/rest-crud`
```

raymondfeng:
@bajtos PTAL.

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -0,0 +1,166 @@
...
+  property,
+} from '../../..';
+
+describe('RepositoryClass builder', () => {
+  describe('defineCrudRepositoryClass', () => {
+    it('should generate CRUD repository class', async () => {
+      const AddressRepository = defineCrudRepositoryClass(
+        Address,
+        DummyCrudRepository,
+      );
```
bajtos: I find this usage problematic, here is what VSCode shows to me in a type hint:

```ts
const AddressRepository: CrudRepositoryClass<Address, DummyCrudRepository<any>>
```

Notice that `DummyCrudRepository` is using type `any` instead of `Address`. This is going to disable type checks in code using address repository instances.

```ts
const repo = new AddressRepository(db);
const address = await repo.create({/*...*/});
// boom, address has type `any` and the compiler don't warn us at all
```

bajtos: Unfortunately, I was not able to find a reliable way how to force TypeScript to reject such usage and always force the user to provide all generic arguments. It seems to me that TypeScript thinks it's able to infer those generic args, but the inference does not work as intended :(

emonddr:
@raymondfeng , @bajtos 
If this is applied and/or #4949 , we may need to revisit the LB4 mixin class factory functions in : https://loopback.io/doc/en/lb4/migration-models-mixins.html .

bajtos:
I rebase the PR on top of the latest master and merged in changes from https://github.com/strongloop/loopback-next/pull/4949. The rebase was a bit tricky, I ended up re-creating both commits from scratch.

packages/repository/src/define-repository-class.ts
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Signature for a Repository class bound to a given model. The constructor
+ * accepts only the dataSource to use for persistence.
+ *
+ * `define*` functions return a class implementing this interface.
+ *
+ * @typeParam M - Model class
+ * @typeParam R - Repository class/interface
+ */
+export interface NamedRepositoryClass<
```
bajtos, line 25: The name of this class is a discussion point we haven't resolved yet. See https://github.com/strongloop/loopback-next/pull/4949#discussion_r400189603.

Cross-posting my latest comment:

> Maybe we should take another angle. My intention is to distinguish between a generic repository class that works for any Model/Entity and a repository class that's already specialized for (bound to) a given Model/Entity.
> 
> In that light, perhaps `BoundRepositoryClass` or `SpecializedRepositoryClass` would be better names?

packages/rest-crud/README.md
```diff
@@ -64,9 +64,9 @@ class defined without the need for a repository or controller class file.
 
 If you would like more flexibility, e.g. if you would only like to define a
 default `CrudRest` controller or repository, you can use the two helper methods
-(`defineCrudRestController` and `defineCrudRepositoryClass`) exposed from
-`@loopback/rest-crud`. These functions will help you create controllers and
-respositories using code.
+(`defineCrudRestController` from `@loopback/rest-crud` and
+`defineCrudRepositoryClass` from `@loopback/repository`). These functions will
+help you create controllers and repositories using code.
```
bajtos, line 69: Should we revert this change since `defineCrudRepositoryClass` is still exported by `@loopback/rest-crud`? (Same comment applies to other similar doc changes below.)

raymondfeng: I'm fine either way.

packages/repository/src/define-repository-class.ts
raymondfeng: Maybe `TypedRepositoryClass`?

raymondfeng: I renamed it to `ModelRepositoryClass`. Let's just land it - no need to be perfect.

Arathy-sivan:
In loopback4 have any distribution folder that we can use to publish in production????

achrinza:
@Arathy-sivan Please either open a new GitHub question issue or [ask on Slack](https://github.com/strongloop/loopback-next/issues/5048). Let's keep to the discussion related to the Pull Request.

Issue #4372, opened 2020-01-06
Belongs-To relation tries to resolve for foreign key NULL


deepakrkris:
## Steps to reproduce
Detailed steps are in issue https://github.com/strongloop/loopback-next/issues/4332

## Current Behavior
Belongs-To resolver tries to fetch objects for foreign keys by constructing an IN clause (the juggler does this on behalf of the resolver). If a foreign key has NULL value the constructed IN clause has a `NULL` value in it. ~This may result in errors in certain conditions~ Edit: This is causing failures in **Postgres** (issue https://github.com/strongloop/loopback-next/issues/4332). When Postgres queries with `null` value id, it throws: `error: DEFAULT is not allowed in this context`.

## Expected Behavior

NULL values in foreign keys need not be queried (edit) or should be handled well as MySQL does, unless we are allowing for logical relations to be defined where the foreign key is just a string value and may not refer to a primary key of another table.

## Link to reproduction sandbox

Provided in issue https://github.com/strongloop/loopback-next/issues/4332

## Acceptance criteria

- [x] Fix Postgres connector to handle the case when `null` values are passed in 
  - example: `Order {id: 1, customerId: null}, {id: 2, customerId: 1}`. When traverse `Order` with inclusion, Postgres should handle the null fk properly.
- [ ]  fix querying related entities with null keys and add related tests to `repository-test`
- [x] check if `Mongo` and `MySQL` connectors handles this case.

agnes512:
So far the issue is only found in Postgres. MySQL and in-memory db work as expected. 
Might need to check Postgres' query syntax to see how it handles `null` value. MySQL repository might be a good reference.

dhmlau:
PR in progress: https://github.com/strongloop/loopback-connector-postgresql/pull/420.

deepakrkris:
PR strongloop/loopback-connector-postgresql#420 is merged.

PR #4795, opened 2020-03-04
fix: filter null keys when including belongs-to relations in queries 

deepakrkris:
fixes https://github.com/strongloop/loopback-next/issues/4372

 fix querying related entities with null keys and add tests to repository-test

## Checklist

### DIFF
--- a/packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
+++ b/packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
@@ -52,12 +52,13 @@ export function createBelongsToInclusionResolver<
     const sourceKey = relationMeta.keyFrom;
     const sourceIds = entities.map(e => (e as AnyObject)[sourceKey]);
     const targetKey = relationMeta.keyTo as StringKeyOf<Target>;
+    const dedupedSourceIds = deduplicate(sourceIds);
 
     const targetRepo = await getTargetRepo();
     const targetsFound = await findByForeignKeys(
       targetRepo,
       targetKey,
-      deduplicate(sourceIds),
+      dedupedSourceIds.filter(e => e),
       inclusion.scope as Filter<Target>,
       options,
     );
--- a/packages/repository-tests/src/crud/relations/acceptance/belongs-to.inclusion-resolver.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/belongs-to.inclusion-resolver.relation.acceptance.ts
@@ -160,6 +160,44 @@ export function belongsToInclusionResolverAcceptance(
       expect(toJSON(result)).to.deepEqual(toJSON(expected));
     });
 
+    it('queries entities with null foreign key', async () => {
+      const customer = await customerRepo.create({
+        name: 'Thor',
+      });
+
+      // order with customer relation
+      const order1 = await orderRepo.create({
+        customerId: customer.id,
+        description: 'Take Out',
+      });
+
+      // order without customer relation
+      const order2 = await orderRepo.create({
+        description: 'Dine in',
+      });
+
+      const expected = [
+        {
+          ...order1,
+          isShipped: features.emptyValue,
+          shipmentInfo: features.emptyValue,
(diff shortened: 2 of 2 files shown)


agnes512:
What does this PR fix? I didn't see any code change.

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
```diff
@@ -106,6 +106,26 @@ export function belongsToRelationAcceptance(
...
+        description: 'Take Out',
+      });
+
+      // order without customer relation
+      await orderRepo.create({
+        description: 'Dine in',
+      });
+
+      let results = await orderRepo.find({include: [{relation: 'customer'}]});
+      expect(results.length).to.equal(2);
```
agnes512: What are we trying to fix in the PR? I didn't see the code changes, and new added test doesn't seem to test the `null` fk.

deepakrkris: Hi @agnes512, I moved in with only the tests first to check all the datasources that fail for this condition. I am adding the code changes now.

deepakrkris: I should have opened a draft PR, sorry about the false alarm.

deepakrkris: @agnes512 the new tests do check for null fk , because the order 'Dine in' has no customer Id.

hacksparrow:
Does this PR make the behavior of all the connectors uniform now? Is there any difference between Postgres and others? If yes, we should add the details on the website docs.

packages/repository-tests/src/crud/relations/acceptance/belongs-to.relation.acceptance.ts
agnes512: Oh haha I see. Sorry that I got the notification and just reviewed it. 

And I have a few questions. I checked other tests in this file, the test in [#L 97](https://github.com/strongloop/loopback-next/pull/4795/files#diff-c77e7e4e7301ea16d5acce2d1252dcd1R97) checks the case that source instance does not have the foreign key. Why is that one passing before the changes? Is the bug caused by the inclusion? If so, I think file `belongs-to.inclusion-resolver.relation.acceptance.ts` fits this test more. From my understanding, `order.customer(order.id)` is different from `order.find({include customer}`, because the later one uses the inclusion resolver, where the problem lies in. It'd also be better if we can check content of `result` to see if there is only one instance has the related model.

deepakrkris: @agnes512 the other test passes because it directly queries the target relation. You are right, I will move this test to the inclusion resolver tests. it suites better there.

deepakrkris:
> Does this PR make the behavior of all the connectors uniform now? Is there any difference between Postgres and others? If yes, we should add the details on the website docs.

@hacksparrow this is not a behavioral change, users will not see any difference in querying. It only removes null keys before querying a relation.

packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
```diff
@@ -52,12 +52,13 @@ export function createBelongsToInclusionResolver<
...
     const sourceIds = entities.map(e => (e as AnyObject)[sourceKey]);
     const targetKey = relationMeta.keyTo as StringKeyOf<Target>;
+    const dedupedSourceIds = deduplicate(sourceIds);
 
     const targetRepo = await getTargetRepo();
     const targetsFound = await findByForeignKeys(
       targetRepo,
       targetKey,
-      deduplicate(sourceIds),
+      dedupedSourceIds.filter(e => e),
```
bajtos: Is it possible to have a valid id value that's not truthy? For example, I can imagine `0` being a valid primary key. Your condition will filter such value out.

PR #4801, opened 2020-03-05
feat(example-file-upload): add an example for file upload with LoopBack 4

raymondfeng:
An example app for file upload.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [x] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4808, opened 2020-03-06
refactor(rest): extract AJV to a separate service provider

raymondfeng:
Inspired by #4801 

- Extract AJV to a separate service provider (use as a factory function for now without DI)
- Allow to ignore `binary` format by default

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/unit/ajv.service.unit.ts
+++ b/packages/rest/src/__tests__/unit/ajv.service.unit.ts
@@ -1,0 +1,53 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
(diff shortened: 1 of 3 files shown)


deepakrkris:
@raymondfeng you could also add some integration tests to check if requests with binary formats are ignored, similar to https://github.com/strongloop/loopback-next/pull/4801

raymondfeng:
> @raymondfeng you could also add some integration tests to check if requests with binary formats are ignored, similar to #4801

That's already covered by the newly added unit test.

emonddr:
@raymondfeng 
regarding 
> // FIXME(rfeng): We would prefer to inject the Ajv service

why can't you inject it?

raymondfeng:
> why can't you inject it?

At this point, `validateRequestBody` is a plain function. DI can only happen with a class with decorations. We need to do some refactoring to allow DI for AJV.

PR #4816, opened 2020-03-08
feat(core): allow an extension to be used by multiple extension points

raymondfeng:
The PR is to address the need by https://github.com/strongloop/loopback-next/pull/4693#discussion_r389321928.

1. Introduce `TagValueMatcher` function type for more flexible tag value matching used by `filterByTag`
2. Use an array inclusion based TagValueMatcher to discover extensions tagged for multiple extension points.

## Checklist

### DIFF
--- a/packages/context/src/binding-filter.ts
+++ b/packages/context/src/binding-filter.ts
@@ -112,6 +112,19 @@ export function isBindingTagFilter(
 }
 
 /**
+ * A function to check if a given tag value is matched for `filterByTag`
+ */
+export interface TagValueMatcher {
+  /**
+   * Check if the given tag value matches the search criteria
+   * @param tagValue - Tag value from the binding
+   * @param tagName - Tag name
+   * @param tagMap - Tag map from the binding
+   */
+  (tagValue: unknown, tagName: string, tagMap: MapObject<unknown>): boolean;
+}
+
+/**
  * A symbol that can be used to match binding tags by name regardless of the
  * value.
  *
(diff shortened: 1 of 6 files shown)


raymondfeng:
@dougal83 Thank you for the review. I further improved the PR to make filterByTag even more flexible. PTAL.

packages/context/src/binding-filter.ts
```diff
@@ -125,7 +133,26 @@ export function isBindingTagFilter(
...
-export const ANY_TAG_VALUE = Symbol.for('loopback.AnyTagValue');
+export const ANY_TAG_VALUE: TagValueMatcher = (tagMap, tagName) =>
+  tagName in tagMap;
+
+/**
+ * Create a tag value matcher function that returns `true` if the target tag
+ * value is an array that includes the given item value.
+ * @param itemValue - Tag item value
+ */
+export function includeTagValueMatcher(itemValue: unknown): TagValueMatcher {
```
bajtos: I find this function name difficult to read and understand what the function is doing. Take a look at an example use in your tests:

```ts
      const filter = filterByTag({
         controller: includeTagValueMatcher('MyController'),
         name: 'my-name',
       });
```

Let's find a better name please. How about `containsTagValue` or `includesTagValue`?

```ts
      const filter = filterByTag({
         controller: containsTagValue('MyController'),
         name: 'my-name',
       });
```

In general, I find the overall design of this feature difficult to comprehend too. `TagValueMatcher` is accepting a tag map and a tag name, where the map is coming from the binding (which is expected), but the name is supplied by some intermediate parts (code inside `filterByTag` in this case). When reading the example above, we need to understand a lot about the way how LoopBack works with binding tags.

I understand the current `TagValueMatcher` offers a lot of flexibility, but the cost is too high in my opinion.

Have you considered a simpler design where the matcher receives just the tag value to match?

```ts
export type TagValueMatcher = (value: unknown) => boolean;
```

Then the "includes" matcher can be greatly simplified or even not needed at all:

```ts
      const filter = filterByTag({
         controller: value => 
           value === 'MyController' || 
           (Array.isArray(value) && value.includes('MyController')),
         name: 'my-name',
       });
```

bajtos:
> The first commit looks mostly good to me. I have reservations about the complexity of the proposed design in the second commit, let's discuss.

Hmm, it looks like GitHub is listing the commits in the opposite order.

By "first commit" I meant 0275dc5c40b6fecf27c01aab579a450f425d54f6 (allow extensions to be used by multiple extension points), "second commit" meant dd6da12ba1d3b1bf3487ee1c9e64ea226b975a9f

packages/context/src/binding-filter.ts
raymondfeng: Sounds good. I now use a similar syntax as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach to pass in the following params:

`(tagValue, tagName, tagMap)`

bajtos:
> - [ ] Documentation in /docs/site was updated

I think these additions deserve documentation too. I think both docs for Context and for ExtensionPoint/Extension pattern should be updated.

PR #4823, opened 2020-03-09
chore: fix setup of docs preview site

bajtos:
Rework the script setting up docs preview to create hard-links instead of symbolic links, because Jekyll no longer supports symbolic links pointing outside of the site directory.

The new solution is based on `pax` tool, which is a part of POSIX, included in MacOS but often not included in GNU/Linux distributions by default. Developers working on loopback-next in Linux may need to install the tool via the package manager provided by their distro. Learn more here: https://en.wikipedia.org/wiki/Pax_(Unix)

## Checklist

### DIFF
--- a/docs/bin/build-preview-site.sh
+++ b/docs/bin/build-preview-site.sh
@@ -51,11 +51,17 @@ cp -r $SOURCE_DIR/doc/index.md $JEKYLL_DIR/doc
 
 echo "Setting up LB4 doc pages",
 rm -rf $JEKYLL_DIR/pages
-ln -s $PWD/site $JEKYLL_DIR/pages
+# Create hardlinks because Jekyll does not support symbolic links any more.
+# Use `pax` because `ln` does not support directory recursion.
+mkdir $JEKYLL_DIR/pages
+(TARGET="$PWD/$JEKYLL_DIR/pages" && cd "$PWD/site" && pax -rwlpe . $TARGET)
 
 echo "Setting up sidebar(s)"
 rm -rf $JEKYLL_DIR/_data/sidebars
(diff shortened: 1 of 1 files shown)


docs/bin/build-preview-site.sh
```diff
@@ -51,11 +51,17 @@ cp -r $SOURCE_DIR/doc/index.md $JEKYLL_DIR/doc
 
 echo "Setting up LB4 doc pages",
 rm -rf $JEKYLL_DIR/pages
-ln -s $PWD/site $JEKYLL_DIR/pages
+# Create hardlinks because Jekyll does not support symbolic links any more.
```
emonddr: There's a need to add 
` "realpath": "^1.1.0",`
to devDependencies

bajtos: Weird. I believe `realpath` is a Unix tool, this is what I see on my machine:

```
$ ls -l `which realpath`
lrwxr-xr-x  1 bajtos  admin  37 Sep 26 15:26 /usr/local/bin/realpath -> ../Cellar/coreutils/8.31/bin/realpath
```

See https://www.gnu.org/software/coreutils/manual/html_node/realpath-invocation.html

Anyhow, I guess it's better to explicitly install realpath as a Node.js dependency, rather than rely on `coreutils` being installed.

bajtos:
I found a way how to avoid dependency on `realpath`, see 3ce643a.

LGTY now, @emonddr?

docs/bin/build-preview-site.sh
dhmlau: FYI - I was able to run this script without installing `realpath`.

Issue #3058, opened 2019-06-06
Required model property with default value throws error when property not provided in create


aharbis:
## Description / Steps to reproduce / Feature proposal

Given a model property that is defined as `required` with a `default` value, trying to create a new model instance and omitting that property from the request yields a 422 error stating the property is missing.

For example, given the following model property definition:

```ts
  @property({
    type: 'string',
    required: true,
    default: 'admin',
  })
  username: string;
```

Trying to create a model instance and omitting the `username` property yields:

```json
{
    "error": {
        "statusCode": 422,
        "name": "UnprocessableEntityError",
        "message": "The request body is invalid. See error object `details` property for more info.",
        "code": "VALIDATION_FAILED",
        "details": [
            {
                "path": "",
                "code": "required",
                "message": "should have required property 'username'",
                "info": {
                    "missingProperty": "username"
                }
            }
        ]
    }
}
```

I've recreated the issue using the `todo` application, and you can view that on my forked feature branch:

https://github.com/aharbis/loopback-next/tree/default-property

I am using Node `v10.15.3`, and based the recreate on an up-to-date fork of `loopback-next`.

## Current Behavior

The `default` value defined for the model property is not utilized, and instead a 422 error is thrown stating the required property is missing.

## Expected Behavior

If a `required` property has a `default` value, and the given property is not provided in the request to create the model instance, the `default` value should be used.

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

aharbis:
It is worth noting that adding a `default` value to a non-required property *does* populate the default value if the request does not contain that property. I also demonstrated this behavior in the recreate - just see the added default value for `isComplete`, and added test validation for the default:

Non-required default property:

```ts
  @property({
    type: 'boolean',
    default: false
  })
  isComplete?: boolean;
```

Test case passes:

```ts
  it('default isComplete set if none provided during create', async () => {
    const todo = givenTodo();
    delete todo.isComplete;
    const response = await client
      .post('/todos')
      .send(todo)
      .expect(200);
    expect(response.body.isComplete).to.equal(false);
  });
```

dougal83:
I could be wrong but I think if you set the field as required, it needs to be provided regardless of there being a default.  If I recall correctly, this is not the same as 'not null' in a db.  That being said if I'm mistaken and if you are using a data connector that supports schema then make sure you use the migrate tools so the defaults are in place to avoid not null.

[...]

PR #4829, opened 2020-03-09
docs: update Models.md

agnes512:
clean up old content and make it less relate to lb3
some content are copied from lb3 and rewrite it with lb4-way

closes https://github.com/strongloop/loopback.io/issues/907

While doing ^907, I realized that we link to LB3 docs a lot. Considering there are several differences between them and not all users are familiar with LB3, I think it's time to separate them. ( And ofc it is also annoying to jump back and forth between LB3 and LB4 docs)

So I edited  the `Model Decorator` and `Property Decorator` sections that:

- copy content form LB3 if we linked it to LB3 before.
- Remove the unsupported part and rewrite it with Ts-way to LB4.
- Edit them as a whole to make it more consistent

I tested them briefly before I removing some features/entries. If you happen to know that they are still available but got deleted, please kindly let me know, thanks!!

## Checklist

### DIFF
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -139,9 +139,10 @@ PersistedModel classes to enforce `strict` mode.
 LB4 supports creating a model that allows both well-defined but also arbitrary
 extra properties for **NoSQL** databases such as MongoDB. You need to disable
 `strict` mode in model settings. Besides modifying model settings directly, it
-can also be done through the CLI by setting `allowed free-from properties` to
-true by saying yes when given this prompt and telling TypeScript to allow
-arbitrary additional properties to be set on model instances.
+can also be done through the CLI by setting
+`Allow additional (free-form) properties` to `true` by saying yes when given
+this prompt and telling TypeScript to allow arbitrary additional properties to
+be set on model instances.
 
 ```ts
 @model({settings: {strict: false}})
@@ -156,7 +157,7 @@ class MyFlexibleModel extends Entity {
 }
 ```
 
-### Model Decorator
+## Model Decorator
 
 The model decorator can be used without any additional parameters, or can be
 passed in a ModelDefinitionSyntax:
@@ -190,29 +191,15 @@ class Product extends Entity {
 }
 ```
 
-However, the model decorator in LoopBack 4 is not exactly the same as what it is
-in LoopBack 3. For example, in
-[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
-pass in a model definition syntax in the model decorator, such as properties,
-options, relation etc. But not all these entries are available in lb4 model
-decorator:
+{% include note.html content="If you used LoopBack 3 before, the model decorator in LoopBack 4 is not exactly the same as what it is
+in LB3. For example, properties and relations cannot be defined through the model decorator. Please check the following section for available entries." %}
 
 <!-- Please modify this part when options is available -->
-
-NOTICE: in LoopBack 4 we only support `settings` in the ModelDefinitionSyntax
-for now. Those `top-level properties` in lb3 now are passed in `settings`.
-
-- `properties` now are defined in `@property` decorator (see below for more
-  information).
-- [`options`](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options)
-  in lb3 doesn't have the mapping feature in LB4 yet. (see
-  [issue #2142](https://github.com/strongloop/loopback-next/issues/2142) for
-  further discussion.)
 
 As for entries in `settings`, LoopBack 4 supports these built-in entries for
 now:
 
-#### Supported Entries of Settings
+### Supported Entries of Model Definition
 
 <!-- These entries might need to update once we've made some changes:
   - description [in lb3 we support array or string, but here we documented it as string only]
@@ -230,9 +217,14 @@ now:
   </thead>
 
   <tbody>
-
   <tr>
-  <td><code>description</code></td>
+    <td><code>name</code></td>
+    <td>String</td>
+    <td>None</td>
+    <td>Name of the model.</td>
+  </tr>
+  <tr>
+  <td><code>settings.description</code></td>
   <td>String</td>
   <td>None</td>
   <td>
@@ -240,16 +232,33 @@ now:
   </td></tr>
 
   <tr>
-  <td><code>forceId</code></td>
-  <td>Boolean</td>
-  <td><code>true</code></td>
-  <td>
-    If true, prevents clients from setting the auto-generated ID value manually.
+    <td><code>settings.forceId</code></td>
+    <td>Boolean</td>
+    <td><code>true</code></td>
+    <td>
+      Set it to <code>true</code> to prevent clients from setting the auto-generated ID value manually.
+    </td>
+  </tr>
+
+  <tr>
+    <td><code>settings.hiddenProperties</code></td>
+    <td>Array of String</td>
+    <td><code>None</code></td>
+    <td>
+      The properties can be hidden from response bodies
+      (<code>.toJSON()</code> output). See <a href="#hidden-properties">Hidden properties</a> section below for details.
   </td>
   </tr>
 
   <tr>
-    <td><code>strict</code></td>
+    <td><code>settings.scope</code></td>
+    <td>Object</td>
+    <td>N/A</td>
+    <td>Scope enables you to set a scope that will apply to every query made by the model's repository. See <a href="#scope">Scope</a> below for more details and examples.</td>
+  </tr>
+
+  <tr>
+    <td><code>settings.strict</code></td>
     <td>Boolean or String</td>
     <td><code>true</code>.<br/></td>
     <td>
@@ -269,39 +278,14 @@ now:
     </td>
   </tr>
 
-  <tr>
-    <td><code>idInjection</code></td>
-    <td>Boolean</td>
-    <td><code>true</code></td>
-    <td>
-      Whether to automatically add an <code>id</code> property to the model:
-    <ul>
-    <li><code>true</code>: <code>id</code> property is added to the model automatically. This is the default.</li>
-    <li><code>false</code>: <code>id</code> property is not added to the model</li>
-    </ul>
-    See <a href="#id-properties">ID properties</a> for more information.  The <code>idInjection</code> property in <code>options</code> (if present) takes precedence.
-    </td>
-  </tr>
-
-  <tr>
-    <td><code>name</code></td>
-    <td>String</td>
-    <td>None</td>
-    <td>Name of the model.</td>
-  </tr>
-
-  <tr>
-    <td><code>scopes</code></td>
-    <td>Object</td>
-    <td>N/A</td>
-    <td>See <a href="https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#scopes">Scopes</a> in lb3 docs.</td>
-  </tr>
-
   </tbody>
 
   </table>
 
-#### Unsupported Entries of Settings
+### Unsupported Entries
+
+If you're a LB3 user, there are several entries that are no longer available in
+LB4:
 
 <!-- Please update the OPTIONS, ACLS field when they are available -->
 
@@ -382,7 +366,7 @@ To discover more about `Model Decorator` in LoopBack 4, please check
 and
 [model-builder file](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js).
 
-#### Hidden properties
+### Hidden Properties
 
 The properties are stored in the database, available in JS/TS code, can be set
 via POST/PUT/PATCH requests, but they are removed from response bodies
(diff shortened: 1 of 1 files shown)


docs/site/Model.md
```diff
@@ -230,26 +216,48 @@ now:
   </thead>
 
   <tbody>
-
   <tr>
-  <td><code>description</code></td>
+    <td><code>name</code></td>
+    <td>String</td>
+    <td>None</td>
```
jannyHou: nitpick: shall we keep the default value consistent with other fields as `N/A` instead of `None`?

docs/site/Model.md
```diff
@@ -230,26 +216,48 @@ now:
...
+    <td><code>N/A</code></td>
+    <td>
+      The properties can be hidden from response bodies
+      (<code>.toJSON()</code> output). See <a href="#hidden-properties">Hidden properties</a> section below for details.
   </td>
   </tr>
 
   <tr>
-    <td><code>strict</code></td>
+    <td><code>setting.scope</code></td>
```
jannyHou:
```suggestion
    <td><code>settings.scopes</code></td>
```

docs/site/Model.md
```diff
@@ -230,26 +216,48 @@ now:
...
+    <td>Object</td>
+    <td>N/A</td>
+    <td>Scope enable you to specify commonly-used queries that you can reference as method calls on a model. See <a href="#scope">Scope</a> below for more details and examples.</td>
```
jannyHou:
```suggestion
    <td>Scope enables you to specify commonly-used queries that you can reference as method calls on a model. See <a href="#scope">Scope</a> below for more details and examples.</td>
```

docs/site/Model.md
```diff
@@ -407,12 +389,42 @@ class MyUserModel extends Entity {
 }
 ```
 
-### Property Decorator
+### Scope
```
jannyHou:
```suggestion
### Scopes
```

jannyHou: The field name is in plural :p https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#scopes

docs/site/Model.md
```diff
@@ -407,12 +389,42 @@ class MyUserModel extends Entity {
...
+
+_Scope_ enable you to specify commonly-used queries that you can reference as method calls on a model.
+
+If you wish for a scope to be applied across all queries to the model, you can set, for example:
+
+```ts
+{
+  name: "Product",
+  scope: {
```
jannyHou:
```suggestion
  scopes: {
```

jannyHou: And I think you need to give each scope a name?
Like 

```js
"scopes": {
	"vips": {"where": {"vip": true}}, 
	"top5": {"limit": 5, "order": "age"}
}

```

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>   
+    <tr>
+      <td>generated</td>
+      <td>No</td>
+      <td>Boolean</td>
+      <td><strong>For auto-migrate usage</strong>. The <code>generated</code> property indicates the ID will be automatically generated by the database. When it sets to true, the value of the id property will be generated by the database automatically with its default type (e.g integer for MySQL and string for MongoDB).</td>
```
jannyHou:
```suggestion
      <td><strong>For auto-migrate usage</strong>. The <code>generated</code> property indicates the ID will be automatically generated by the database. When it sets to `true`, the value of the id property will be generated by the database automatically with its default type (e.g integer for MySQL and string for MongoDB).</td>
```

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
+    </tr>
+    <tr>
+      <td>useDefaultIdType</td>
+      <td>No</td>
+      <td>Boolean</td>
+      <td><strong>For auto-migrate usage</strong>. Set it to false when it's needed to auto-generate non-default type property values. For example, for PostgreSQL, to use uuid as the id property, the id type should set to string, <code>generated</code> should set to true, and this field should be set to false. Please check each connector's README file for more information about auto-migration/auto-update.</td>
```
jannyHou:
```suggestion
      <td><strong>For auto-migrate usage</strong>. Set it to `false` when it's needed to auto-generate non-default type property values. For example, for PostgreSQL, to use uuid as the id property, the id type should set to string, <code>generated</code> should set to `true`, and this field should be set to `false`. Please check each connector's README file for more information about auto-migration/auto-update.</td>
```

docs/site/Model.md
agnes512: Yes, but in LB4 it only works with settings `scope`:
```ts
@model({
  settings: {
    scope: {
      'limit': 2,
    },
    idInjection: true
  }
})
```
:( that's why I made them `scope` instead of `scopes`

docs/site/Model.md
agnes512: I don't think this part works in LB4, at least from what I tried.

docs/site/Model.md
jannyHou: I see! remembered, LB3 supports both `scopes` and `scope`:
- scopes: a collection of predefined scopes
- scope: the default scope

So your change makes perfect sense 💯

docs/site/Model.md
jannyHou: 👍 take my comment back :p

docs/site/Model.md
jannyHou: ⬆️ take my comment back, it should be `scope`

docs/site/Model.md
agnes512: Just want to double check this part: in LB3 we allow:
```ts
"scopes": {
	"vips": {"where": {"vip": true}}, 
	"top5": {"limit": 5, "order": "age"}
}
```
then the scopes can be used as 
```ts
User.scope("vips", { where: { vip: true } });
User.scope("top5", { limit: 5, order: "age" });
```
I don't think this is valid in LB4 because:
- `User` model, or any lb4 model, doesn't have such method `scope`
- LB4 uses repositories to query data. So `User.scope` won't work.

Is this correct?

docs/site/Model.md
```diff
@@ -140,7 +140,7 @@ LB4 supports creating a model that allows both well-defined but also arbitrary
 extra properties for **NoSQL** databases such as MongoDB. You need to disable
 `strict` mode in model settings. Besides modifying model settings directly, it
 can also be done through the CLI by setting `allowed free-from properties` to
```
nabdelgadir: I know this isn't part of your PR but can you fix `allowed free-from properties` to `allow free-form properties`?

docs/site/Model.md
```diff
@@ -190,29 +190,15 @@ class Product extends Entity {
...
 ```
 
-However, the model decorator in LoopBack 4 is not exactly the same as what it is
-in LoopBack 3. For example, in
-[lb3](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) we can
-pass in a model definition syntax in the model decorator, such as properties,
-options, relation etc. But not all these entries are available in lb4 model
-decorator:
+{% include note.html content="If you used LB3 before, the model decorator in LoopBack 4 is not exactly the same as what it is
+in LoopBack 3. For example, properties and relation cannot be defined through the model decorator. Please check the following section for available entires." %}
```
nabdelgadir: Can you change LB3 in the line before to "LoopBack 3" and change the second one to LB3?
`relation` -> `relations`
`entires` -> `entries`

docs/site/Model.md
```diff
@@ -230,26 +216,48 @@ now:
...
+      (<code>.toJSON()</code> output). See <a href="#hidden-properties">Hidden properties</a> section below for details.
   </td>
   </tr>
 
   <tr>
-    <td><code>strict</code></td>
+    <td><code>settings.scope</code></td>
+    <td>Object</td>
+    <td>N/A</td>
+    <td>Scope enable you to specify commonly-used queries that you can reference as method calls on a model. See <a href="#scope">Scope</a> below for more details and examples.</td>
```
nabdelgadir: `enable` -> `enables`
Just wondering, what does "method calls" mean?

docs/site/Model.md
```diff
@@ -407,12 +389,42 @@ class MyUserModel extends Entity {
 }
 ```
 
-### Property Decorator
+### Scope
+
+_Scope_ enables you to specify commonly-used queries that you can reference as method calls on a model.
+
+If you wish for a scope to be applied across all queries to the model, you can set, for example:
```
nabdelgadir: `set, for example` -> `set the scope to do so. For example`?

docs/site/Model.md
```diff
@@ -407,12 +389,42 @@ class MyUserModel extends Entity {
...
+Now, any CRUD operation with a query parameter runs in the default scope will be applied; for example, assuming the above scope, a find operation such as
+
+```ts
+await Product.find({offset: 0});
+```
+
+Becomes the equivalent of this:
+
+```ts
+await Product.find({order: "name", offset: 0, limit: 100, where: {deleted: false}})
```
nabdelgadir: Is 100 the default limit?

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
+      <td>
+        Property type. Can be any type described in <a href="LoopBack-types.html">LoopBack types</a>.
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+### ID Properties
+
+LoopBack 4 expects a model to have one _ID properties_ that uniquely identify the model instance.
```
nabdelgadir: `properties` -> `property`
`identify` -> `identifies`

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
+
+{% include important.html content="LB4 doesn't support composite keys for now. e.g joining two tables with more than one source key. Related GitHub issue: [Composite primary/foreign keys](https://github.com/strongloop/loopback-next/issues/1830)" %}
```
nabdelgadir: `for now. e.g.` -> `for now, e.g.`

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>   
+    <tr>
+      <td>generated</td>
+      <td>No</td>
+      <td>Boolean</td>
+      <td><strong>For auto-migrate usage</strong>. The <code>generated</code> property indicates the ID will be automatically generated by the database. When it sets to <code>true</code>, the value of the id property will be generated by the database automatically with its default type (e.g integer for MySQL and string for MongoDB).</td>
```
nabdelgadir: `it sets to` -> `it is set to`

docs/site/Model.md
```diff
@@ -432,78 +444,261 @@ class Product extends Entity {
...
-property decorator. Aliased types or types that extracted from a class or
-interface (e.g. `public name: OtherClass['otherProperty']`) will not work
-properly and will result in the property type being resolved as an empty object
-rather than the intended type in the generated OpenAPI specifcation. This is due
-to a limitation and flaw in the way TypeScript currently generates the metadata
-that is used to generate the OpenAPI specification for the application.
+1.  LoopBack CRUD methods expect the model to have an "id" property if the model is backed by a database.
+2.  A model without any "id" properties can only be used without attaching to a database.
+3.  If an ID property has `generated` set to `true`, the connector decides what type to use for the auto-generated key. For example for SQL Server, it defaults to `number`. This can be overwritten by setting `useDefaultIdType` to `false`.
+4.  Checkout [Database Migration](Database-migrations.md) if you'd like to have LoopBack 4 to create relational database schemas for you based on model definitions. Always check [Database Connectors](Database-connectors.md) for details and examples for database migration / discover.
```
nabdelgadir: `Checkout` -> `Check out`
`to create` -> `create`
`database` -> `database's`

docs/site/Model.md
agnes512: Good catch! Changed it to `Allow additional (free-form) properties` to match with the current CLI.

docs/site/Model.md
agnes512: I just copied it from LB3 files :p I guess it's not the case if https://github.com/strongloop/loopback-next/pull/4829#discussion_r389977511  is not available in LB4.


How about: `Scope enables you to set a scope that will be applying to every query made by the repository of this model`?

And I will update examples as:

If you wish for a scope to be applied across all queries to **the repository of the model**, you can set, for example:

```
@model({
  settings: {
    scope: {
      'limit': 2,
    },
    idInjection: true
  }
})
```

Now, any CRUD operation with a query parameter runs in the default scope will be applied; for example, assuming the above scope, a find operation such as

```ts
await ProductRepository.find({offset: 0});
```

Becomes the equivalent of this:

```ts
await ProductRepository.find({order: "name", offset: 0, limit: 100, where: {deleted: false}})
```

docs/site/Model.md
agnes512: updated

agnes512:
Enhanced the docs for `required` and `default` fields that we discussed in https://github.com/strongloop/loopback-next/issues/3058 in commit https://github.com/strongloop/loopback-next/pull/4829/commits/685361a67ab2589f3053f356d73d2e503d9d981f

docs/site/Model.md
jannyHou: Defining 
```ts
"scopes": {
	"vips": {"where": {"vip": true}}, 
	"top5": {"limit": 5, "order": "age"}
}
```
is equivalent to 
```ts
// please note User.scope defines scope, not apply scope
User.scope("vips", { where: { vip: true } });
User.scope("top5", { limit: 5, order: "age" });
```
After defining them, you can call APIs like `User.vips()` to query with the predefined filter.

But your conclusion is correct 👌 we don't have such APIs in User repository. Therefore `scopes` won't be supported. Only specifying the default `scope` works.

docs/site/Model.md
nabdelgadir: Looks good to me! I would just reword a bit: `Scope enables you to set a scope that will apply to every query made by the model's repository`

agnes512:
Even the settings can be imported from LB3, see  https://github.com/strongloop/loopback-next/pull/4737/files#r383816292, I've decided to remove `idInjection`. Because even the DB would have an id property added to the model automatically with `idInjection: true` set, the id property is still needed in the model definition so that the repository/controller  can be set afterwards. 

We can revert the change if ^ the conclusion is incorrect.

PR #4706, opened 2020-02-21
docs: access control migration

jannyHou:
This is the doc for https://github.com/strongloop/loopback-next/issues/4520, it helps review feat PR #4571

- The content is temporarily written in the readme file. It should be moved to https://github.com/strongloop/loopback-next/blob/auth-migration/docs/site/migration/auth/example.md

- To make the tutorial more readable and consistent, I am going to put all the code snippet, cmd prompts into references. Now I only have the tutorial part, those references are in progress

- Since the feat PR is still in change, I just write an outline for the authorizer and enforcers.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

Issue #4521, opened 2020-01-27
[migration/auth] Create a guide to describe how to migrate LB3 User based authentication to LB4


raymondfeng:
Create a guide to describe how to migrate LB3 built-in User/AccessToken model based authentication

## Acceptance criteria

- [x] Describe usage of LB4 datasource and repository for the user/access token data access
- [x] Migrate model definitions from LB3 User/AccessToken to LB4
- [x] Create a login controller to perform authentication
- [x] Create a token service to generate access tokens
- [x] Create an authentication strategy to validate access tokens
- [x] Update https://github.com/strongloop/loopback-next/blob/auth-migration/docs/site/migration/auth/built-in.md

dhmlau:
@bajtos @jannyHou, I'd like to continue the discussion from https://github.com/strongloop/loopback-next/pull/4571#discussion_r383710046 of having a reusable extension/component for the JWT authentication. If that's what we're going to create, should we get that component created first before working on this docs? 

cc @raymondfeng

jannyHou:
@dhmlau Yes. Moving here after close https://github.com/strongloop/loopback-next/issues/4753

jannyHou:
Will address the comment in https://github.com/strongloop/loopback-next/pull/4706#discussion_r383625928

jannyHou:
closed as PRs merged. 🎉

PR #4855, opened 2020-03-10
refactor: migrate models from cli

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/4521, prepare for User model migration

address comment https://github.com/strongloop/loopback-next/pull/4706#discussion_r383625928

## Checklist

- [ ] ~New tests added or existing tests modified to cover all changes~
- [ ] ~API Documentation in code was updated~
- [ ] ~Affected artifact templates in `packages/cli` were updated~

### DIFF
--- a/docs/site/migration/auth/example.md
+++ b/docs/site/migration/auth/example.md
@@ -297,93 +297,95 @@ To make the tutorial concise, the code details is omitted here. You can find the
 ACLs and how the endpoints are decorated in file
 [src/controller/project.controller.ts](https://github.com/strongloop/loopback-next/tree/master/examples/access-control-migration/src/controllers/project.controller.ts)
 
-Next, we write the authorizer that calls casbin enforcers to make the decision
+Next, we write the authorizer that calls casbin enforcers to make the decision.
 
 1. Create an authorizer that retrieves the authorization metadata from context,
    then execute casbin enforcers to make decision.
 
-_If you are not familiar with the concept authorizer, you can learn it in the
-document
-[Programming Access Policies](../../Loopback-component-authorization#programming-access-policies)_
+   _If you are not familiar with the concept authorizer, you can learn it in the
+   document
+   [Programming Access Policies](../../Loopback-component-authorization#programming-access-policies)_
 
-The complete authorizer file can be found in
-[components/casbin-authorization/services/casbin.authorizer.ts](https://github.com/strongloop/loopback-next/blob/master/examples/access-control-migration/src/components/casbin-authorization/services/casbin.authorizer.ts).
-It retrieves the three required fields from the authorization context: subject
-from `principal`, `resource` as object, and action, to invoke casbin enforcers
-and make decision.
+   The complete authorizer file can be found in
+   [components/casbin-authorization/services/casbin.authorizer.ts](https://github.com/strongloop/loopback-next/blob/master/examples/access-control-migration/src/components/casbin-authorization/services/casbin.authorizer.ts).
+   It retrieves the three required fields from the authorization context:
+   subject from `principal`, `resource` as object, and action, to invoke casbin
+   enforcers and make decision.
 
 2. Create a voter for instance level endpoints to append the project id to the
    resource name
 
-Class level means operations applied to all projects like `/projects/*`, whereas
-instance level operation applies to a certain project like `/project{id}/*`. For
-instance level endpoints, the resource name should include the resource's `id`.
-Since the `id` comes from the request, the endpoint metadata cannot provide it
-and therefore we define the pre-process logic in a voter to generate the
-resource name as `project${id}`, then passes it to the authorizer.
+   Class level means operations applied to all projects like `/projects/*`,
+   whereas instance level operation applies to a certain project like
+   `/project{id}/*`. For instance level endpoints, the resource name should
+   include the resource's `id`. Since the `id` comes from the request, the
+   endpoint metadata cannot provide it and therefore we define the pre-process
+   logic in a voter to generate the resource name as `project${id}`, then passes
+   it to the authorizer.
 
-The complete voter file can be found in file
-[components/casbin-authorization/services/assign-project-instance-id.voter.ts](https://github.com/strongloop/loopback-next/blob/master/examples/access-control-migration/src/components/casbin-authorization/services/assign-project-instance-id.voter.ts)
+   The complete voter file can be found in file
+   [components/casbin-authorization/services/assign-project-instance-id.voter.ts](https://github.com/strongloop/loopback-next/blob/master/examples/access-control-migration/src/components/casbin-authorization/services/assign-project-instance-id.voter.ts)
 
 3. Create casbin enforcers
 
-A casbin enforcer is configured with a policy file. It compares the given data
-with the policy file and returns a decision when invoked. To optimize the scope
-and speed of the access check, this example splits the policies into separate
-files per role. The authorizer will only invoke the enforcers for allowed
-role(s).
+   A casbin enforcer is configured with a policy file. It compares the given
+   data with the policy file and returns a decision when invoked. To optimize
+   the scope and speed of the access check, this example splits the policies
+   into separate files per role. The authorizer will only invoke the enforcers
+   for allowed role(s).
 
-The complete enforcer file can be found in file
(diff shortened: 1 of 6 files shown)


docs/site/migration/auth/example.md
```diff
@@ -446,143 +446,64 @@ Try the 'team-member' role:
...
 
-Create UserCredentials
+Choose model `User`, `Project`, `Team` for the prompt.
+
+Considering the difference between the original application and the migrated
+one, you need to adjust the properties and add the relations.
+
+- For `Project`, open file `src/models/project.model.ts`
+
+  - decorate `ownerId` with `@belongsTo(() => User)` and remove the generated
```
dhmlau: In our migration guide https://loopback.io/doc/en/lb4/migration-models-relations.html#migration-path, we recommend to run `lb4 relation` for adding relations. Could we follow that instructions too? Thanks.

jannyHou: @dhmlau We don't expose endpoints for relations so the generated controllers are deleted, that's why I didn't user the cmd to generate relation. But refactored and added.

dhmlau: Thanks. I didn't realize that.

docs/site/migration/auth/example.md
```diff
@@ -446,143 +448,64 @@ Try the 'team-member' role:
...
+    // change it from `true` to `false`
+    generated: false,
+    updateOnly: true,
+  })
+  // change it from optional `id?` to required `id`
+  id: number;
+  ```
 
-Create UserCredentials
+For model `UserCredentials`, you need to create it by `lb4 model`:
```
dhmlau: I'd suggest to move the model creation right after importing other models, so that we'll have all the models created by that time. Then the next few steps are about changing the model.ts and adding relations.

docs/site/migration/auth/example.md
```diff
@@ -636,6 +559,56 @@ $ lb4 repository
 ? Please select the repository base class DefaultCrudRepository (Legacy juggler bridge)
 ```
 
+Create relations:
+
+- `Project` belongsTo `User`
```
dhmlau: in the instruction, it changes the project.model.ts to add the `belongsTo` relation with `User`.  Is it a duplicate in either place?

docs/site/migration/auth/example.md
```diff
@@ -636,6 +559,56 @@ $ lb4 repository
...
+? Foreign key name to define on the source model ownerId
+? Relation name owner
+? Allow Project queries to include data from related User instances? Yes
+```
+
+{% include tip.html content="
+Please delete the generated controller file `src/controllers/project-user.controller.ts` to keep the exposed endpoints clean.
+" %}
+
+- `User` hasOne `UserCredentials`
```
dhmlau: Similar comment ... We've defined the relation when changing the user.model.ts

docs/site/migration/auth/example.md
```diff
@@ -297,93 +297,95 @@ To make the tutorial concise, the code details is omitted here. You can find the
...
-It retrieves the three required fields from the authorization context: subject
-from `principal`, `resource` as object, and action, to invoke casbin enforcers
-and make decision.
+   The complete authorizer file can be found in
+   [components/casbin-authorization/services/casbin.authorizer.ts](https://github.com/strongloop/loopback-next/blob/master/examples/access-control-migration/src/components/casbin-authorization/services/casbin.authorizer.ts).
+   It retrieves the three required fields from the authorization context:
+   subject from `principal`, `resource` as object, and action, to invoke casbin
+   enforcers and make decision.
 
-2. Create a voter for instance level endpoints to append the project id to the
```
agnes512: Why do we change them all to`1.`?

docs/site/migration/auth/example.md
```diff
@@ -446,143 +448,64 @@ Try the 'team-member' role:
 
 ### Model Creation
 
-Create User Model
+Import model `User`, `Project`, `Team` by the migration CLI:
```
agnes512: 'Import model `User`, `Project`, `Team` from the LB3 app by the migration CLI:' ?

examples/access-control-migration/src/models/project.model.ts
```diff
@@ -3,32 +3,32 @@
...
 
-@model({settings: {strict: false}})
+@model({settings: {strict: false, idInjection: true}})
 export class Project extends Entity {
   @property({
     type: 'number',
-    id: true,
+    id: 1,
     generated: false,
+    updateOnly: true,
```
agnes512: I have a few concerns here but they might not be related to the PR itself:
- Is `settings. idInjection` from model migration? While I was doing task https://github.com/strongloop/loopback-next/pull/4829, I removed `idInjection` from the LB4 docs because it doesn't work when I tested it ( I might test it incorrectly)
- LB3 uses  `id: 1` cause it allows composite ids, and 1 indicates true and also its index. Ideally we should just use `id: true` for imported models.

examples/access-control-migration/src/repositories/project.repository.ts
```diff
@@ -20,13 +20,15 @@ export class ProjectRepository extends DefaultCrudRepository<
 > {
   public readonly user: BelongsToAccessor<User, typeof Project.prototype.id>;
 
+  public readonly owner: BelongsToAccessor<User, typeof Project.prototype.id>;
+
```
agnes512: There are several relations. If I read it correctly, `Project` only has a belongsTo relation `owner`. 
Then I think `public readonly user: BelongsToAccessor<User, typeof Project.prototype.id>;` can be deleted.

docs/site/migration/auth/example.md
jannyHou: deleted the manual creation ones :p

docs/site/migration/auth/example.md
jannyHou: I think either incremental number or all as 1. works.
See example in https://loopback.io/doc/en/lb4/Authentication-Tutorial.html#try-it-out

The real fix for the listing is the indence, so the number doesn't quite matter. But since the number causes confusion I revert it.

examples/access-control-migration/src/models/project.model.ts
jannyHou: @agnes512 I didn't have it before, it's generated by migration CLI. If it's not correct, I think we'd better fix it from CLI side.

agnes512: True, but I don't know how much effort it needs

jannyHou: ah got it after reading the story #4829 , let me remove them.

docs/site/migration/auth/example.md
jannyHou: removed as https://github.com/strongloop/loopback-next/pull/4855#discussion_r391254286

docs/site/migration/auth/example.md
```diff
@@ -510,132 +476,131 @@ Enter an empty property name when done
...
-   create src/models/project.model.ts
-   update src/models/index.ts
+  - remove `password` because we have a `UserCredential` model created to
+    separate it from `User`
 
-Model Project was created in src/models/
-```
+- For all the three models above, allow generating their `id` field:
+  ```ts
+    @property({
```
dhmlau: nitpick: there's an extra spacing.

examples/access-control-migration/src/models/project.model.ts
agnes512: I check what @bajtos did in his PR: https://github.com/strongloop/loopback-next/pull/4737/files#r383816292

And this is what I tried:
1. Define a LB4 model with `idInjection: true`
```@model({
  settings: {
    idInjection: true,
  },
})
export class User extends Entity {
  // @property({   // no id property defined
  //   type: 'number',
  //   id: true,
  //   generated: true,
  // })
  // id?: number;

  @property({
    type: 'string',
  })
  name: string;
...
}
```

2. then run migration-> failed.

I am not hundred percent sure if `idInjection` works in LB4. I am okay to keep it for sake of consistency.

jannyHou: @agnes512 many thanks for double checking and verifying with code! 
> I am okay to keep it for sake of consistency.

If it causes failure I'd like to remove in the example, I can investigate a bit (after having PR ready for https://github.com/strongloop/loopback-next/issues/4521) and create issue/PR to fix if needed.

PR #4896, opened 2020-03-16
[RFC] feat(cron): add an extension to support cron jobs for LoopBack 4

raymondfeng:
This PR adds an extension for LoopBack 4 to provide `cron` capability with https://github.com/kelektiv/node-cron.

## Checklist


emonddr:
Great. Some folks were asking for something like this: https://github.com/strongloop/loopback-next/pull/4896

extensions/cron/README.md
```diff
@@ -0,0 +1,112 @@
...
+### Register Cron jobs
+
+We can create an instance of `CronJob` and bind it to the application context so
+it can be managed by the `CronComponent` life cycle.
+
+```ts
+import {CronJob, asCronJob} from '@loopback/cron';
+
+const job = new CronJob({
+  cronTime: '*/1 * * * * *', // Every one second
```
emonddr: > // Every one second 

// Every second

extensions/cron/README.md
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+const jobBinding = createBindingFromClass(CronJobProvider);
+app.add(jobBinding);
+const now = new Date();
+now.setMilliseconds(now.getMilliseconds() + 10);
+const jobConfig: CronJobConfig = {
+  name: 'job-B', // Name the job
+  cronTime: sendAt(now), // Start the job in 10 ms
```
emonddr: I love the concept of running cron jobs from a loopback application.
Can we have another example besides one that executes every 10 seconds. Perhaps, register a job to occur at midnight everyday with a name like 'produce report' or something along those lines? Just a thought.

extensions/cron/README.md
```diff
@@ -0,0 +1,112 @@
...
+  runOnInit: false,
+};
+app.configure(jobBinding.key).to(jobConfig);
+```
+
+By default, all cron jobs will be started and stopped by `CronComponent` when
+the application is started or stopped.
+
+To start a cron job after application starts, we can either set `runOnInit` to
```
emonddr: Are we really `starting` the cron job as soon as application starts? e.g. running the job. Or do we mean `register a cron job with the operating system` as soon as the app starts?

emonddr: Is the module we are using actually going to enter a cron job into linux OS? Or does it `mimic` the idea of `cron` on linux?

raymondfeng: It mimics the `cron` scheduler.

extensions/cron/README.md
raymondfeng: The job is for testing purpose and we can not wait for midnight for a CI run. :-).

Issue #3958, opened 2019-10-17
How to migrate apps using loopback-component-passport


bajtos:
_This is a follow-up for #3718 and #3922._

Write content for `docs/site/migration/auth/passport.md`, explain how to migrate LB3 applications using [loopback-component-passport](https://github.com/strongloop/loopback-component-passport) to LB4.

I think the long term goal should be to move our users to [`@loopback/authentication-passport`](https://github.com/strongloop/loopback-next/tree/master/extensions/authentication-passport).

~~For short to medium term, while they are running both LB3 and LB4 app at the same time, we can show them how to use Passport as Express middleware, see https://github.com/strongloop/loopback-next/issues/3947.~~ 
scoped to #5005

## Acceptance criteria

Migration guide describing manual steps for the following scenarios:

- [x] LB4-only, this should show how to migrate to our recommended setup. (PR : https://github.com/strongloop/loopback-next/pull/4899)
     - [x]   add support for oauth2 redirection in loopback strategy (https://github.com/strongloop/loopback-next/issues/4902)
     - [x]   create example app for login with facebook and google, similar to https://github.com/strongloop/loopback-example-passport (https://github.com/strongloop/loopback-next/issues/2624)
     - [x]  enabling client sessions in the example app to save user profiles on login and delete session on logout

- [x] ~~Both LB3 and LB4 applications running in the same process, we want to use a single passport configuration to secure both LB3 and LB4 routes.~~
scoped into a separate issue: https://github.com/strongloop/loopback-next/issues/5005

deepakrkris:
closing this epic as the scoped out issue https://github.com/strongloop/loopback-next/issues/5005 is not closely related to the original story.

PR #4899, opened 2020-03-16
docs: migrate loopback-example-passport repo as lb4 example

deepakrkris:
satisfies: https://github.com/strongloop/loopback-next/issues/3958

## Test steps

** This example repo would need manual test

* preconditions: create a test app and test user in facebook, and make sure to give `email` permission to test user. (emonddr : visit https://developers.facebook.com to do this.) 
* preconditions: Copy test app's client id and secret and update those in the examples/passport-poauth2-login/authentication-strategies/oauth2-strategy.ts

* Run `npm i` from loopback-next root folder, to create symlinks with master packages
* cd examples/passport-oauth2-login & npm start
* open browser -> http://localhost:3000  (emonddr : Open a new window in Incognito mode, otherwise may not work)
* click login -> login with facebook 
* page redirects to facebook, enter test user credentials
* if authentication was successful, page redirects to login profiles page and shows linked facebook account details along with access token

## Checklist


examples/passport-oauth2-login/src/authentication-strategies/redirect-route.ts
```diff
@@ -0,0 +1,68 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/rest
```
emonddr: I've noticed in a few files the `Node module` comment seems to have different values:

```
// Node module: @loopback/authentication or @loopback/rest
```

Please make them all the same value appropriate for this example. See other examples. thx.

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Adapter class to invoke passport-strategy
+ *   1. provides express dependencies to the passport strategies
+ *   2. provides shimming of requests for passport authentication
+ *   3. provides life-cycle similar to express to the passport-strategy
+ *   4. provides state methods to the strategy instance
+ * see: https://github.com/jaredhanson/passport
+ */
+export class StrategyAdapter<U> {
```
emonddr: Why is another strategy adapter being defined here? you should use: https://github.com/strongloop/loopback-next/blob/master/extensions/authentication-passport/src/strategy-adapter.ts .

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+import {StrategyAdapter} from './strategy-adapter';
+import {Strategy, StrategyOption, Profile} from 'passport-facebook';
+import {bind, inject} from '@loopback/context';
+import {Request} from '@loopback/rest';
+import {UserProfile, securityId} from '@loopback/security';
+import {UserService, UserServiceBindings} from '../services';
+import {User} from '../models';
+import {asSpecEnhancer, OASEnhancer} from '@loopback/openapi-v3';
+
+const options: StrategyOption = {
```
emonddr: Are these real ids? Perhaps remove them?

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    this.facebookStrategy = new Strategy(options, this.verify.bind(this));
+    this.facebookAuthentication = new StrategyAdapter(
+      this.facebookStrategy,
+      'facebook',
+      mapFaceBookProfile,
+    );
+  }
+  modifySpec(
+    spec: import('@loopback/rest').OpenAPIObject,
```
emonddr: What is being done here exactly? What is being added to the app's open api spec? thx.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,104 @@
...
+  verify(
+    accessToken: string,
+    refreshToken: string,
+    profile: Profile,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    done: (error: any, user?: any, info?: any) => void,
+  ) {
+    const userProfile: UserProfile = mapFaceBookProfile(profile);
+    this.userService
+      .findOrCreateExternalUser(userProfile)
```
emonddr: What is meant by external user? If you could clarify. thx.

examples/passport-oauth2-login/src/authentication-strategies/redirect-route.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  async invokeHandler(
+    {response}: RequestContext,
+    args: OperationArgs,
+  ): Promise<OperationRetval> {
+    response.redirect(this.statusCode, this.targetLocation);
+  }
+
+  updateBindings(requestContext: RequestContext) {
```
emonddr: Do we need this as an empty method? Please clarify. thx.

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
emonddr: Were you waiting for https://github.com/strongloop/loopback-next/pull/4919 to land?

examples/passport-oauth2-login/src/models/access-token.model.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/example-passport-oauth2-login
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+@model({settings: {strict: false}})
```
emonddr: Why are settings required in the @model decorator?

emonddr: This link https://loopback.io/doc/en/lb4/Model.html#using-the-juggler-bridge shows 2 examples. One where no settings are placed in @model decorator, and one with . If you could elaborate. thx.

examples/passport-oauth2-login/src/authentication-strategies/auth-action.provider.ts
```diff
@@ -0,0 +1,84 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Getter, inject, Provider, Setter} from '@loopback/context';
```
jannyHou: Hmm, wouldn't this action be provided by the `@loopback/authentication` module?

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
```diff
@@ -0,0 +1,86 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication-passport
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {UserProfileFactory} from '@loopback/authentication';
```
jannyHou: I think we should leverage the adapter from https://github.com/strongloop/loopback-next/tree/master/extensions/authentication-passport

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
deepakrkris: hey @emonddr yes exactly, I was waiting for 4919 to be landed, the changes will have to be published as well.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: @emonddr , yep I created a test app in facebook, it is not a security risk as of now. but I am going to remove it shortl

examples/passport-oauth2-login/src/authentication-strategies/redirect-route.ts
deepakrkris: agreed

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
deepakrkris: symlinks with master package is enabled now, please see PR description for manual test steps

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: this is not relevant, removed now.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: we are creating a local profile on behalf of an external provider, added additional comments to clarify this now.

examples/passport-oauth2-login/src/authentication-strategies/redirect-route.ts
deepakrkris: agreed, removed

examples/passport-oauth2-login/src/models/access-token.model.ts
deepakrkris: I am still looking at this

examples/passport-oauth2-login/src/authentication-strategies/auth-action.provider.ts
deepakrkris: agreed

examples/passport-oauth2-login/src/authentication-strategies/strategy-adapter.ts
deepakrkris: agreed, symlinks to master packages enabled now.

examples/passport-oauth2-login/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "passport-oauth2-login",
```
raymondfeng: The name should be something like `@loopback/example-passport-oauth2`.

examples/passport-oauth2-login/index.js
```diff
@@ -0,0 +1,29 @@
+const application = require('./dist');
```
raymondfeng: Copyright/license header is missing.

examples/passport-oauth2-login/client/web-app.js
```diff
@@ -0,0 +1,93 @@
+// Copyright IBM Corp. 2014,2015. All Rights Reserved.
+// Node module: loopback-example-passport
```
raymondfeng: The module name does not match package.json.

examples/passport-oauth2-login/src/application.ts
```diff
@@ -0,0 +1,48 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {BootMixin} from '@loopback/boot';
+import {RepositoryMixin} from '@loopback/repository';
+import {RestApplication} from '@loopback/rest';
+import {ServiceMixin} from '@loopback/service-proxy';
+import {MySequence} from './sequence';
+import {AuthenticationComponent} from '@loopback/authentication';
+import {Oauth2Controller} from './controllers';
```
raymondfeng: Should we use `OAuth2` instead of `Oauth2`?

examples/passport-oauth2-login/src/datasources/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/example-access-control-migration
```
raymondfeng: Wrong module name

examples/passport-oauth2-login/key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
emonddr, line 1: What are these .pem files for? thx

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -8,51 +8,89 @@ import {
...
+   * @param request
+   */
   async authenticate(request: Request): Promise<UserProfile | undefined> {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const response: any = await this.facebookAuthentication.authenticate(
-      request,
-    );
-    return response;
+
+    let poviderName = request.route.provider;
```
emonddr, line 74: > poviderName

providerName

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -71,34 +109,38 @@ export class Oauth2Authorization
...
+  facebookProfile: Profile,
 ): UserProfile {
+  let user = facebookProfile as UserWithToken;
   const userProfile: UserProfile = {
     [securityId]: user.id,
     profile: {
       ...user,
-      source: 'facebook',
-    },
+      _json: null,
```
emonddr, line 141: whats _json and _raw? Perhaps provide a comment. thx.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,146 @@
...
+    public userService: UserService,
+  ) {
+    this.facebookStrategy = new Strategy(facebookOptions, this.verify.bind(this));
+    this.facebookAuthentication = new StrategyAdapter(
+      this.facebookStrategy,
+      'facebook',
+      mapFaceBookProfile,
+    );
+    this.googleStrategy = new GoogleStrategy(googleOptions, this.verify.bind(this));
+    this.googleAuthentication = new StrategyAdapter(this.googleStrategy, 'google', mapFaceBookProfile);
```
raymondfeng, line 65: We should consider to use `@extensions` to inject different strategies.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,146 @@
...
+  callbackURL: oauth2Providers['facebook-login'].callbackURL,
+  profileFields: oauth2Providers['facebook-login'].profileFields
+};
+
+const googleOptions: GoogleOptions = {
+  clientID: oauth2Providers['google-login'].clientID,
+  clientSecret: oauth2Providers['google-login'].clientSecret,
+  callbackURL: oauth2Providers['google-login'].callbackURL,
+  scope: ['email']
+};
```
raymondfeng, line 34: These options can be possibly refactored into injectable configurations.

examples/passport-login/src/authentication-strategies/provider.ts
```diff
@@ -5,31 +5,62 @@
...
 
-    if (!strategy) throw new HttpErrors.Unauthorized('unknown provider');
+  async value(): Promise<AuthenticationStrategy | undefined> {
+    if (!this.metadata) return;
+    const name = this.metadata.strategy;
+    const strategy = await this.findAuthenticationStrategy(name);
     return strategy;
   }
+
+  async findAuthenticationStrategy(name: string) {
```
emonddr, line 61: i am a little confused here. We already have an authentication strategy provider. https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-strategy.provider.ts#L41 .

examples/passport-oauth2-login/src/models/access-token.model.ts
deepakrkris: I have removed the strict: false setting now

examples/passport-oauth2-login/package.json
deepakrkris: agreed

examples/passport-oauth2-login/index.js
deepakrkris: agreed

examples/passport-oauth2-login/src/application.ts
deepakrkris: agreed

examples/passport-oauth2-login/client/web-app.js
deepakrkris: agreed

examples/passport-oauth2-login/src/datasources/index.ts
deepakrkris: agreed

examples/passport-oauth2-login/key.pem
deepakrkris: agreed, removing them now

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: agreed

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: agreed

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: agreed, @raymondfeng I have added extensions for provider specific implementation now

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
deepakrkris: agreed, I am injecting into the request context in sequence.ts

examples/passport-oauth2-login/src/sequence.ts
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      const {request, response} = context;
+      const route = this.findRoute(request);
+
+      // usually authentication is done before proceeding to parse params and invoke
+      // but in our case we need the path params to know the provider name
+
+      // parse request to get path params
+      const args = await this.parseParams(request, route);
+      request.route = route.pathParams;
```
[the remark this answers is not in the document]
deepakrkris: agreed, removing this now

examples/passport-oauth2-login/src/controllers/oauth2.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    const userProfile: SessionProfile = {
+      ...user.profile
+    };
+    if (!request.session.user) {
+      request.session.user = {
+        profiles: [userProfile],
+      };
+    }
+    if (request.session.user) {
```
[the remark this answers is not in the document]
deepakrkris: agreed

examples/passport-oauth2-login/src/application.ts
```diff
@@ -0,0 +1,48 @@
...
+    super(options);
+
+    this.setUpBindings();
+
+    this.add(createBindingFromClass(Oauth2Authorization));
+
+    // Set up the custom sequence
+    this.sequence(MySequence);
+
+    this.controller(Oauth2Controller);
```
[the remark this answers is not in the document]
deepakrkris: is it ? it does not seem to work for me, let me check.

examples/passport-oauth2-login/src/authentication-strategies/oauth2-strategy.ts
```diff
@@ -0,0 +1,146 @@
...
+   */
+  async authenticate(request: Request): Promise<UserProfile | undefined> {
+
+    let poviderName = request.route.provider;
+
+    // for now we authenticate only with facebook and google
+    if (poviderName === 'facebook') {
+      // using request params we can redirect to multiple providers
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const response: any = await this.facebookAuthentication.authenticate(
```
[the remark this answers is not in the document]
deepakrkris: agreed, corrected this now.

examples/passport-oauth2-login/src/services/user.service.ts
```diff
@@ -0,0 +1,54 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {repository} from '@loopback/repository';
+import {Profile as UserProfile} from 'passport';
+import {UserRepository} from '../repositories';
+import {User} from '../models';
+import * as _ from 'lodash';
+
+export interface UserService {
```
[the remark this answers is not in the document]
deepakrkris: agreed, added a new interface 'UserIdentityService' in authentication package

examples/passport-login/src/authentication-strategies/provider.ts
deepakrkris: agreed, removing this now.

examples/passport-oauth2-login/src/application.ts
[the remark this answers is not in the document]
emonddr: yeah @jannyHou  is right. For example, the Todo app doesn't do this: https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/application.ts#L27 .

examples/passport-login/index.ts
```diff
@@ -0,0 +1 @@
+export * from './src';
```
raymondfeng, line 1: Copy right header is missing. Please run `../../packages/cli/bin/cli-main.js copyright` in `examples/passport-login` to fix it.

examples/passport-login/oauth2-providers.json
```diff
@@ -0,0 +1,47 @@
+{
+    "facebook-login": {
+        "provider": "facebook",
+        "module": "passport-facebook",
+        "profileFields": ["gender", "link", "locale", "name", "timezone",
+        "verified", "email", "updated_time", "displayName", "id"],
+        "clientID": "{facebook-client-id}",
+        "clientSecret": "ad4315e3453b012f6e4b8dd4bc1c0ae6",
```
raymondfeng, line 8: Please remove the `clientSecret`.

examples/passport-login/oauth2-providers.json
```diff
@@ -0,0 +1,47 @@
...
+        "failureRedirect": "/login",
+        "scope": ["email"],
+        "failureFlash": true
+    },
+    "google-login": {
+        "provider": "google",
+        "module": "passport-google-oauth2",
+        "strategy": "OAuth2Strategy",
+        "clientID": "{google-client-id}",
+        "clientSecret": "FcldSqY5-9usqlRmQhZawN__",
```
raymondfeng, line 22: Ditto.

examples/passport-login/oauth2-providers.json
```diff
@@ -0,0 +1,47 @@
...
+        "failureRedirect": "/login",
+        "scope": ["email", "profile"],
+        "failureFlash": true
+    },
+    "oauth2": {
+        "provider": "oauth2",
+        "module": "passport-google-oauth2",
+        "strategy": "OAuth2Strategy",
+        "clientID": "{google-client-id}",
+        "clientSecret": "FcldSqY5-9usqlRmQhZawN__",
```
raymondfeng, line 36: Ditto.

examples/passport-login/package.json
```diff
@@ -0,0 +1,91 @@
+{
+  "name": "@loopback/passport-login",
+  "version": "1.0.0",
```
raymondfeng, line 3: Let's name it as `@loopback/example-passport-login` to be consistent with other example projects.

examples/passport-login/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "url": "https://github.com/strongloop/loopback-next.git",
+    "directory": "examples/passport-login"
+  },
+  "author": "IBM Corp.",
+  "license": "MIT",
+  "dependencies": {
+    "@loopback/authentication": "^4.1.1",
+    "@loopback/authentication-passport": "^2.0.0",
+    "@loopback/boot": "2.0.2",
+    "@loopback/context": "3.2.0",
```
raymondfeng, line 40: Please use `^` for all `@loopback/*` dependencies.

examples/passport-login/src/authentication-strategies/facebook.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: example-passport-oauth2-login
```
raymondfeng, line 2: Incorrect module name. The `lb4 copyright` command will fix it.

examples/passport-login/src/authentication-strategies/google.ts
```diff
@@ -0,0 +1,96 @@
...
+import {Request, RedirectRoute} from '@loopback/rest';
+import {PassportAuthenticationBindings} from './oauth2';
+
+@bind(
+  asAuthStrategy,
+  extensionFor(PassportAuthenticationBindings.OAUTH2_STRATEGY),
+)
+export class GoogleOauth2Authorization implements AuthenticationStrategy {
+  name = 'oauth2-google';
+  passportstrategy: Strategy;
```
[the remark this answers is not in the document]
emonddr: Also, to user `extensionFor` with a unique string value, aren't you required to have a class that defines `@extensionPoint` which is responsible for finding all the extensions? For example: https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-strategy.provider.ts#L24 .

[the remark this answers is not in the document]
emonddr: @raymondfeng, is my assumption above correct?

examples/passport-login/src/authentication-strategies/oauth2.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * create an oauth2 strategy
+   */
+  constructor(
+    /**
+     * enable extensions for provider specific oauth2 implementations
+     * reroute to the specific extension based on given provider name
+     */
+    @extensions(PassportAuthenticationBindings.OAUTH2_STRATEGY)
```
emonddr, line 39: Google and Facebook are registered as separate authentication strategies. Why does this  OAuth strategy have a getter to find facebook and google strategies?

emonddr: How do you intend to decorate a controller endpoint?

So far it is possible to say:

@authenticate('local')
or 
@authentication('facebook')
or
@authenticate('google')
or
@authenticate('oauth2') 

what will happen if it is 'oauth2'? Is it meant to handle many passports authentications that adhere to oauth2 settings or something?

examples/passport-login/src/authentication-strategies/oauth2.ts
```diff
@@ -0,0 +1,123 @@
...
+import {User} from '../models';
+import {UserServiceBindings} from '../services';
+import {inject, bind, extensions, Getter} from '@loopback/core';
+
+export namespace PassportAuthenticationBindings {
+  export const OAUTH2_STRATEGY = 'passport.authentication.oauth2.strategy';
+}
+
+@bind(asAuthStrategy)
+export class Oauth2AuthStrategy implements AuthenticationStrategy {
```
emonddr, line 26: I think this class needs to declare the extension point

@extensionPoint(PassportAuthenticationBindings.OAUTH2_STRATEGY)  . see my comment above.

examples/passport-login/src/authentication-strategies/oauth2.ts
deepakrkris: to avoid these many authenticate decorators and separate controllers for each provider, say google, facebook, twitter, etc etc, we have created a single controller with path param 'provider'

deepakrkris: @emonddr we use the extensions in the oauth2 strategy to handle specific providers

examples/passport-login/src/authentication-strategies/google.ts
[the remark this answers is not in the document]
deepakrkris: @emonddr the extension point in the OAuth strategy , https://github.com/strongloop/loopback-next/blob/fbd9b3c191962db6b2d2517a0533486382337509/examples/passport-login/src/authentication-strategies/oauth2.ts#L39

examples/passport-login/src/authentication-strategies/google.ts
```diff
@@ -0,0 +1,96 @@
...
+import {Strategy, StrategyOptions} from 'passport-google-oauth2';
+import {bind, inject} from '@loopback/context';
+import {UserServiceBindings} from '../services';
+import {extensionFor} from '@loopback/core';
+import {securityId, UserProfile} from '@loopback/security';
+import {User} from '../models';
+import {Request, RedirectRoute} from '@loopback/rest';
+import {PassportAuthenticationBindings} from './oauth2';
+
+@bind(
```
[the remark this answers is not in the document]
deepakrkris: @emonddr answers below
1. 
    A. registerAuthenticationStrategy is a short form for "`addExtension()` with tag AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME"
    B. addExtension creates a `New` binding from a given class which is decorated with @bind , and adds additional tags to it as extension keys
    C. extensionFor adds the given tags as extension keys to a given binding
    As you can see each of these are distinct short forms ie, sugar for larger code samples.

2. @bind is a decorator that tells this a class with tags x,y,z

3. In this PR we are using @bind(asAuthStrategy) to decorate the strategy class with authentication tags. Here we add tags from extensionFor with "authentication extension points". Then in the application.ts we create a new binding using createBindingFromClass for each of the auth strategy classes. The tags readily available from the @bind decorator covering those classes.

4. Shopping example seems to do the same as above point 3, except for adding the authentication extensions tag.

examples/passport-login/src/authentication-strategies/local.ts
```diff
@@ -0,0 +1,119 @@
...
+import {User} from '../models';
+import {bind} from '@loopback/context';
+import {Strategy, IVerifyOptions} from 'passport-local';
+import {repository} from '@loopback/repository';
+import {UserRepository} from '../repositories';
+import {extensionFor} from '@loopback/core';
+
+@bind(
+  asAuthStrategy,
+  extensionFor(
```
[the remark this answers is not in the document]
deepakrkris: @emonddr I have explained in comment https://github.com/strongloop/loopback-next/pull/4899#discussion_r405138452 . This is valid for the given usecase and works perfectly fine.

examples/passport-login/src/authentication-strategies/oauth2.ts
deepakrkris: @emonddr @extensionPoint is just an addition to @extensions() decorator. It is optional. If you test the given application , you will see that the app works fine without it.

emonddr: https://loopback.io/doc/en/lb4/core-tutorial-part5.html#access-extensions-for-a-given-extension-point   uses @extensionPoint on the class that searches for @extensions. 

I don't see it mention that @extensionPoint is optional. But I do see this sentence:
```
By default, the name of the extension point is the same as the class name. To customize metadata such as name for the extension point, we can use @extensionPoint to decorate the class
```
So does this mean that any class bound to the context (class doesn't use @extensionPoint) is considered an extension point. So if I have N different classes that I am binding to a context (that have nothing to do with the extensionPoint/extension pattern, they are automatically considered an extensionPoint? I thought that having the @extensionPoint decorator whas to give a particular class the unique status of being an actual extension point to which someone can register extensions. @raymond, is @extensionPoint really optional?

examples/passport-login/src/authentication-strategies/local.ts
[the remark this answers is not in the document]
emonddr: My point in this particular line, is that asAuthStrategy is defined as

```
export const asAuthStrategy: BindingTemplate = binding => {
  extensionFor(
    AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
  )(binding);
  binding.tag({
    namespace:
      AuthenticationBindings.AUTHENTICATION_STRATEGY_EXTENSION_POINT_NAME,
  });
};
```
which already calls extensionFor against the same name. So your 2nd parameter is duplicating the work that your 1st parameter is doing.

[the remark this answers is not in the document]
deepakrkris: @emonddr , you are right, in this class alone the binding key is duplicated, I will correct this.

examples/passport-login/src/models/user-identity.model.ts
```diff
@@ -0,0 +1,54 @@
...
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class UserIdentity extends Entity {
+  @property({
+    type: 'string',
+    id: 1,
```
emonddr, line 12: why are we giving this id a value of 1?

examples/passport-login/src/models/user.model.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import {Entity, model, property, hasOne, hasMany} from '@loopback/repository';
+import {UserCredentials} from './user-credentials.model';
+import {UserIdentity} from './user-identity.model';
+
+@model()
+export class User extends Entity {
+  @property({
+    type: 'number',
+    id: 1,
```
emonddr, line 14: why are we giving this id a value of 1?

examples/passport-login/src/services/user.service.ts
```diff
@@ -0,0 +1,112 @@
...
+   * @param userId
+   * @param userIdentity
+   */
+  async linkExternalProfile(
+    userId: string,
+    userIdentity: PassportProfile,
+  ): Promise<User> {
+    let profile;
+    try {
+      profile = await this.userIdentityRepository.findById(userIdentity.id);
```
emonddr, line 68: I think in UserIdentityRepo and UserCredentialsRepo, you would need to override the findById to locate using the `userId` field (instead of id field), wouldn't you? The value of User.id is stored in UserIdentity.userId and UserCredential.userId since this is the foreignKey pointing back to user right? I think it is working right now because everything has id 1 and it works. Or perhaps you can use  .find and pass a where clause where userId = something.

emonddr:
In addition to the controllers you have, which seem to involve logging in (local) and oauth callbacks (oauth2), I would suggest  the default ping controller that is automatically built for an lb4 app (and requires no authentication), and create another controller with an endpoint that prints out currently logged in user as a response. Decorate this method  with @authenticate('oauth2'). This would show the we can access the /ping endpoint with or without being logged in, and the /printme endpoint only if we are properly logged in.

examples/passport-login/src/authentication-strategies/google.ts
```diff
@@ -0,0 +1,96 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: example-passport-oauth2-login
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  asAuthStrategy,
+  AuthenticationStrategy,
+  UserIdentityService,
+  AuthenticationBindings,
```
emonddr, line 10: AuthenticationBindings not used

examples/passport-login/src/authentication-strategies/google.ts
```diff
@@ -0,0 +1,96 @@
...
+   * create an oauth2 strategy for google
+   */
+  constructor(
+    @inject(UserServiceBindings.PASSPORT_USER_IDENTITY_SERVICE)
+    public userService: UserIdentityService<Profile, User>,
+    @inject('googleOAuth2Options')
+    public googleOptions: StrategyOptions,
+  ) {
+    this.passportstrategy = new Strategy(googleOptions, this.verify.bind(this));
+    this.strategy = new StrategyAdapter(
```
emonddr, line 42: This setup for google doesn't match what you have for FaceBook or Local
where you use .bind(this)

```
    this.passportstrategy = new Strategy(
      facebookOptions,
      this.verify.bind(this),
    );
    this.strategy = new StrategyAdapter(
      this.passportstrategy,
      this.name,
      this.mapProfile.bind(this),
    );
```

emonddr:
@deepakrkris , please provide a readme.md file with instructions on how to run this example.
I automatically ran `npm start` and the server says it is running on 3000 but I couldn't visit the url.
Also `npm run test` fails to run
```
․․․!!!!

  3 passing (436ms)
  4 failing

  1) example-passport-login acceptance test
       login options
         sign up via social app
           call to authorization url is redirected to oauth providers login page:
     Error: ECONNREFUSED: Connection refused
      at Test.assert (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:165:15)
      at localAssert (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:131:12)
      at /Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:128:5
      at Test.Request.callback (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:728:3)
      at ClientRequest.<anonymous> (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:647:10)
      at Socket.socketErrorListener (_http_client.js:415:9)
      at emitErrorNT (internal/streams/destroy.js:92:8)
      at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
      at processTicksAndRejections (internal/process/task_queues.js:81:21)

  2) example-passport-login acceptance test
       login options
         sign up via social app
           login page redirects to authorization app callback endpoint:
     Error: ECONNREFUSED: Connection refused
      at Test.assert (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:165:15)
      at localAssert (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:131:12)
      at /Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/supertest/lib/test.js:128:5
      at Test.Request.callback (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:728:3)
      at ClientRequest.<anonymous> (/Users/dremond/Documents/PRs_LoopBack/deepak_oauth2/loopback-next/packages/testlab/node_modules/superagent/lib/node/index.js:647:10)
      at Socket.socketErrorListener (_http_client.js:415:9)
      at emitErrorNT (internal/streams/destroy.js:92:8)
      at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
      at processTicksAndRejections (internal/process/task_queues.js:81:21)

  3) example-passport-login acceptance test
       login options
         sign up via social app
           callback url contains access code:
     TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received type undefined
      at validateString (internal/validators.js:112:11)
      at Url.parse (url.js:155:3)
      at Object.urlParse [as parse] (url.js:150:13)
      at Context.<anonymous> (src/__tests__/acceptance/passport-login.acceptance.ts:108:20)
      at processImmediate (internal/timers.js:439:21)

  4) example-passport-login acceptance test
       login options
         sign up via social app
           access code can be exchanged for token:
     TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received type undefined
      at validateString (internal/validators.js:112:11)
      at Url.parse (url.js:155:3)
      at Object.urlParse [as parse] (url.js:150:13)
      at Context.<anonymous> (src/__tests__/acceptance/passport-login.acceptance.ts:112:34)
      at processImmediate (internal/timers.js:439:21)


```

emonddr:
Regarding

> preconditions: create a test app and test user in facebook, and make sure to give email permission to test user 

This is actually quite confusing for someone who has never gone here. I have, but never set up a user. So a detailed step by step description of what to do including links would be helpful.

emonddr:
> open browser -> http://localhost:3000 

Didn't work for me. I got

image


I had to open a new window in Incognito mode for Chrome.

emonddr:
I signed up by going to `Sign Up Options`, and was able to see data on `View Account`.
Then I logged out.
Then I went to `Log In`.
The browser goes to `View Account` and I see:

image

emonddr:
Clicking on the [Server Status] link on the website,
I get an error 

image

emonddr:
If I click on `Sign Up`,
and specify my data (not associated with social media)
dom emond
dom@example.com
etc

I get redirected to /login
and I am able to login

Like above [View Account] has no info about me.

When I try to Log out

I get an error 

image

emonddr:
I like the UI, by the way

image

I was thinking we should only show certain menu options when they are applicable.
e.g. Let's not show Logout until someone is actually logged in 
       Let's not show Sign Up, Sign Up Options, Login when someone is logged in

emonddr:
And for some reason, my db.json ends up with many test users



```
  "models": {
    "User": {
      "3": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":3}",
      "4": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":4}",
      "5": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":5}",
      "6": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":6}",
      "7": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":7}",

```

I have restarted several times. I this done on startup?

emonddr:
@deepakrkris , please see my comments from 4 to 5 days ago. Thanks.

examples/passport-login/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,89 @@
...
+  @post('/login')
+  async login(
+    @requestBody({
+      description: 'login to create a user session',
+      required: true,
+      content: {
+        'application/x-www-form-urlencoded': {schema: CredentialsSchema},
+      },
+    })
+    credentials: Credentials,
```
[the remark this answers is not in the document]
deepakrkris: @emonddr credentials is an argument injected from requestBody, so the requestBody schema can be tied to the method call and the authentication strategy can access the requestbody.

examples/passport-login/src/models/user-identity.model.ts
deepakrkris: @emonddr I used it in place of true , they both seem to work alike. I will change it to true.

deepakrkris: agreed

examples/passport-login/src/models/user.model.ts
deepakrkris: agreed, will change it to true,

examples/passport-login/src/services/user.service.ts
deepakrkris: @emonddr the UserIdentity.id is the id generated and available in the external oauth profile. 
the UserCredential.id is the username , that is already available in the User model. These are ideally the identifiers of these models. The uniqueness of the models are also guaranteed.

examples/passport-login/src/authentication-strategies/oauth2.ts
deepakrkris: @emonddr I dont have much knowledge on how the metadata is organized differently, by using `@extensionpoint` and the docs dont seem to clarify that as well. @raymondfeng should be able to answer that. But the code works for this example.

examples/passport-login/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,89 @@
...
+      name: credentials.name,
+    });
+    response.redirect('/login');
+    return response;
+  }
+
+  @authenticate('local')
+  @post('/login')
+  async login(
+    @requestBody({
```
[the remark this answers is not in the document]
deepakrkris: @emonddr , the requestBody schema is the reason we have injected credentials as an argument.

examples/passport-login/src/authentication-strategies/google.ts
deepakrkris: agreed, I will fix this.

examples/passport-login/src/authentication-strategies/local.ts
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @param username
+   * @param password
+   * @param done
+   *
+   * @returns User model
+   */
+  verify(
+    username: string,
+    password: string,
```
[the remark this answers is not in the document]
deepakrkris: @emonddr agreed, password was not used to compare user credentials. Only username was used for the sake of local testing. I will fix this.

examples/passport-login/package.json
deepakrkris: @raymondfeng , agreed, fixed now

examples/passport-login/src/authentication-strategies/facebook.ts
deepakrkris: agreed, fixed now

deepakrkris:
> In addition to the controllers you have, which seem to involve logging in (local) and oauth callbacks (oauth2), I would suggest the default ping controller that is automatically built for an lb4 app (and requires no authentication), and create another controller with an endpoint that prints out currently logged in user as a response. Decorate this method with @authenticate('oauth2'). This would show the we can access the /ping endpoint with or without being logged in, and the /printme endpoint only if we are properly logged in.

@emonddr the authentication endpoints are used in this example to create an authenticated user session for the express web application. This is similar to https://github.com/strongloop/loopback-example-passport . To authenticate the api endpoints we would need to add a different authentication strategy like 'jwt'.

deepakrkris:
> And for some reason, my db.json ends up with many test users
> 
> ```
>   "models": {
>     "User": {
>       "3": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":3}",
>       "4": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":4}",
>       "5": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":5}",
>       "6": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":6}",
>       "7": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":7}",
> ```
> 
> I have restarted several times. I this done on startup?
> I've never used this user.

@emonddr thanks for testing this. I have fixed this now.

deepakrkris:
> I like the UI, by the way
> 
> image
> 
> I was thinking we should only show certain menu options when they are applicable.
> e.g. Let's not show Logout until someone is actually logged in
> Let's not show Sign Up, Sign Up Options, Login when someone is logged in

agree that the menu is very naive, but I just copied most of the UI templates from https://github.com/strongloop/loopback-example-passport

deepakrkris:
> If I click on `Sign Up`,
> and specify my data (not associated with social media)
> dom emond
> [dom@example.com](mailto:dom@example.com)
> etc
> 
> I get redirected to /login
> and I am able to login
> 
> Like above [View Account] has no info about me.
> 
> When I try to Log out
> 
> I get an error
> 
> image

@emonddr the logout function is fixed now . I have also added acceptance test cases for these scenarios.

deepakrkris:
> Clicking on the [Server Status] link on the website,
> 
> image
> 
> I get an error
> 
> image
> 
> Clicking `Logout` I get this:
> 
> image

The `Server status` was a OOTB functionality in lb3. I have removed that link now. Logout had an error, I have fixed that now.

deepakrkris:
> Regarding
> 
> > preconditions: create a test app and test user in facebook, and make sure to give email permission to test user
> 
> This is actually quite confusing for someone who has never gone here. I have, but never set up a user. So a detailed step by step description of what to do including links would be helpful.

@emonddr I have added details on facebook login in the README.md now.

emonddr:
> @emonddr the authentication endpoints are used in this example to create an authenticated user session for the express web application. This is similar to https://github.com/strongloop/loopback-example-passport . To authenticate the api endpoints we would need to add a different authentication strategy like 'jwt'. 

@deepakrkris , apologies. I don't follow. Isn't the whole purpose of this example to see how someone can use facebook or google login to get authenticated, and then get access to LB4 APIs?

examples/passport-login/src/controllers/user.controller.ts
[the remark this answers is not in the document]
emonddr: > @emonddr credentials is an argument injected from requestBody, so the requestBody schema can be tied to the method call and the authentication strategy can access the requestbody.

Perhaps add a comment explaining the reasoning to folks downloading the example.

examples/passport-login/src/models/user-identity.model.ts
emonddr: > @emonddr I used it in place of true , they both seem to work alike. I will change it to true.

Never seen this usage.

deepakrkris:
> > @emonddr the authentication endpoints are used in this example to create an authenticated user session for the express web application. This is similar to https://github.com/strongloop/loopback-example-passport . To authenticate the api endpoints we would need to add a different authentication strategy like 'jwt'.
> 
> @deepakrkris , apologies. I don't follow. Isn't the whole purpose of this example to see how someone can use facebook or google login to get authenticated, and then get access to LB4 APIs?

@emonddr the purpose is to authenticate with facebook/google, YES. But not to validate user access to other LB4 APIs, but to validate access to the express web application. In this case the path `/auth/account` is the only page that gets validated for a user session.

examples/passport-login/src/services/user.service.ts
emonddr: can you produce a diagram displaying a User model and all its field, and  the id field and userId field for the UserIdentity and UserCredential models. I am having a hard time see that the relations are going to work. thx.

emonddr:
I don't see belongsTo decorator in UserIdentity or UserCredential but User has @hasOne and @hasMany respectively. Also in latest commit, if you are using `rest-crud` to create user endpoints, I don't think it handles relations at the moment. @nabdelgadir , I saw this in your blog. Can you confirm? thx.

examples/passport-login/data/db.json
```diff
@@ -1,15 +1,22 @@
 {
   "ids": {
-    "User": 6,
-    "UserIdentity": 104566694532673,
-    "UserCredentials": 3
+    "User": 9,
+    "UserIdentity": 104566694532676,
+    "UserCredentials": 4
```
emonddr, line 5: Is it possible for the this file to zero entries in User, UserIdentity, and UserCredentials, and have the "ids" have entries with value 1 or something. Right now the "ids".UserCredentials is 4, but looking in UserCredentials, the userId is 6. So it looks weird.

examples/passport-login/src/controllers/ping.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+};
+
+/**
+ * A simple controller to bounce back http requests
+ */
+export class PingController {
+  constructor(@inject(RestBindings.Http.REQUEST) private req: Request) {}
+
+  @authenticate('session')
```
emonddr, line 40: User should always be able to visit /ping without authentication (even if logged in with an auth strategy)

Let's leave the /ping without  the `@authenticate` decorator

but have another endpoint in this controller like 

@authenticate('session')
@get(/whoAmi )     
which returns the currently logged in user.

What will the request headers show?

emonddr:
@deepakrkris , thanks for addressing my comments yesterday. We're almost there.

emonddr:
> @deepakrkris , I don't see belongsTo decorator in UserIdentity or UserCredential but User has @hasone and @hasmany respectively. Also in latest commit, if you are using `rest-crud` to create user endpoints, I don't think it handles relations at the moment. @nabdelgadir , I saw this in your blog. Can you confirm? thx. @agnes512 , can you chime in on this as well? In a hasMany or hasOne, there must be a belongsTo on the other side of the relation right? @belongsTo is not optional, correct?
> 
> Also, perhaps rename : `examples/passport-login/src/controllers/user.controller.ts` to `user-signup-login.controller.ts` since user.controller.ts is usually created for a user model for all its endpoints.

Hi @deepakrkris , so I spoke with @agnes512 . So it turns out you were correct. You don't need the belongsTo in UserCredential and UserIdentity if you don't plan on navigating from any of these back to the original User. If you only need to navigate from User to UserIdentity or User to UserCredentials, then you only need the hasMany and hasOne. If you need to navigate from parent to child and child to parent, then you need to use hasMany and belongsTo, or hasOne and belongsTo.

packages/authentication/src/services/user-identity.service.ts
```diff
@@ -0,0 +1,61 @@
...
+ *      public userRepository: UserRepository,
+ *      @repository(UserIdentityRepository)
+ *      public userIdentityRepository: UserIdentityRepository,
+ *    ) {}
+ *  }
+ */
+export interface UserIdentityService<I, U> {
+  /**
+   * find or create a local user using a profile from an external source
+   * @param userIdentity
```
jannyHou, line 24: maybe write it more clear that `userIdentity` here refers to the external identity.

examples/passport-login/src/services/user.service.ts
jannyHou: Followed the conversation here, a suggestion for `UserIdentity`: maybe rename it to something that clearly say it's an EXTERNAL identity?

Like `ExternalIdentity`, `ThirdPartyIdentity`? just to distinguish it with the local `UserPorfile`.

Same for the newly added interface `UserIdentity` in `@loopback/authentication`.
WDYT? cc @deepakrkris @emonddr 

Feel free to ignore it if it's only my own concern :) I understand that renaming takes huge refactor effort.

examples/passport-login/src/server.ts
```diff
@@ -0,0 +1,78 @@
...
+ * An express server with multiple apps
+ *
+ *  1. WEB App
+ *       a. An express web app which requires various user sign up provisions
+ *       b. The express router is mounted in the root '/' path
+ *  2. LB4 API server
+ *       a. LB4 application provides passport login services for the express app
+ *       b. The LB4 login apis are wrapped with session middleware to allow client sessions with user profiles
+ */
+export class ExpressServer {
```
jannyHou, line 23: I am confused about the server+app structure here...so you setup an express server, and use the `webApp` and `lbApp` as two routers?

examples/passport-login/data/db.json
```diff
@@ -0,0 +1,22 @@
...
+      "6": "{\"name\":\"Test User\",\"username\":\"test@example.com\",\"email\":\"test@example.com\",\"id\":6}",
+      "7": "{\"name\":\"\\\"tinkerbell\\\"\",\"username\":\"usr1@lb.com\",\"email\":\"usr1@lb.com\",\"id\":7}",
+      "8": "{\"name\":\"{\\\"familyName\\\":\\\"User\\\",\\\"givenName\\\":\\\"Open\\\",\\\"middleName\\\":\\\"Graph Test\\\"}\",\"username\":\"dplojbdiuc_1585460374@tfbnw.net\",\"email\":\"dplojbdiuc_1585460374@tfbnw.net\",\"id\":8}"
+    },
+    "UserIdentity": {
+      "1001": "{\"id\":\"1001\",\"provider\":\"custom-oauth2\",\"profile\":{\"emails\":[{\"value\":\"usr1@lb.com\"}]},\"authScheme\":\"custom-oauth2\",\"created\":\"2020-04-13T22:01:23.670Z\",\"userId\":7}",
+      "1003": "{\"id\":\"1003\",\"provider\":\"custom-oauth2\",\"profile\":{\"emails\":[{\"value\":\"test@example.com\"}]},\"authScheme\":\"custom-oauth2\",\"created\":\"2020-04-13T22:01:23.556Z\",\"userId\":6}",
+      "104566694532668": "{\"id\":\"104566694532668\",\"provider\":\"facebook\",\"profile\":{\"emails\":[{\"value\":\"dplojbdiuc_1585460374@tfbnw.net\"}]},\"authScheme\":\"facebook\",\"created\":\"2020-04-13T23:18:46.875Z\",\"userId\":8}"
+    },
+    "UserCredentials": {
```
jannyHou, line 18: one `User` should have one `UserCredentials`, any reason the db here has 3 users but only 1 entry of credentials..?

examples/passport-login/src/controllers/ping.controller.ts
jannyHou: +1, `/whoAmI` or `/me` prints the user info, `/ping` is usually public.

deepakrkris: agreed, this is done now

examples/passport-login/data/db.json
deepakrkris: @emonddr I have enabled clearing test data after every test run

deepakrkris:
> @deepakrkris Great effort!!! a few questions & suggestions(commented with details):
> 
> * the method in the ping controller looks good in general, but it's better to have it in the existing user controller and transfer the method to `/whoAmI` or `/me`
> * renaming `UserIdentity`?
> * maybe more explanation about the the app/server structure
> * data in db.json
> 
> I need more time to review each file in details but other than questions above I don't see obvious problem.

Hey @jannyHou thanks for the review,

1. I have added endpoint '/whoAmI' to print the current user , as per your suggestion
2. UserIdentity is a term borrowed from the original loopback-component-passport , since this is a migration story, may be better to have it for now.
3. The server structure is `Express Composition` , we already have examples in detail for it
4. The user credentials are created only for locally signed up users. This is also borrowed from the loopback-component-passport. We can change that after completing this migration story.

examples/passport-login/oauth2-providers.json
deepakrkris: @raymondfeng agreed , modified the configurations now

examples/passport-login/oauth2-providers.json
deepakrkris: agreed

examples/passport-login/oauth2-providers.json
deepakrkris: agreed

examples/passport-login/package.json
deepakrkris: agreed, modified now as suggested

examples/passport-login/README.md
```diff
@@ -0,0 +1,112 @@
...
+By default the user data is stored using a memory connector and saved locally to
+`data/db.json`
+
+Start the application with
+
+```
+$ npm start
+```
+
+- Open your browser to the example app with, `http://localhost:3000`
```
emonddr, line 106: @deepakrkris , thanks for making the UI changes and addressing my earlier comments.

Let's update this section with the latest UI changes in mind (some of the instructions are out of date.)

But also it is important to register a local user BEFORE any social media login, otherwise there will be issues.

1. Open browser to localhost:3000
2. Sign up as a local user with the email of a facebook test user you created above, and provide a password for local purposes. This registers a local user, and it will eventually be linked to social media credentials in the following steps.
3. Login as this local user. You will be brought to the "View Account" tab ( and describe what they should see)
4. Click on Sign Up or Login, and the pages indicate you've already logged in.
5. Click on [Current User] button, and it calls https://localhost:3000/api/whoAmI endpoint and displays currently authenticated user.
6. Click on Logout.
7. Click on View Account, and you are not permitted since you are logged out.
8. Look at the db.json and mention what they should see in there. User with specific email, and in UserCredentials, the password you enterered when you signed up.
9. Now click on Login, and choose `Login with Facebook`. Enter the email of your facebook test user (same email you entered above), but specify the test user's facebook password (not local password).
10. Upon successful FaceBook login, you are brought to `View Account`. (Explain the data they should see) 
11. Click on Sign Up or Login, and the pages indicate you are already logged in.
12. Click on [Current User] to view the currently authenticated user.
13. Click on Logout 
14. Click on `View Account`, and you are not allowed.
15. Look at the db.json, there is still one user. It still has the UserCredentials entry as earlier, But now there is an entry in the UserIdentity section which contains the facebook profile info.
At any time (authenticated or not), open http://localhost:3000/api/ping in the browser and you should see the ping info returned. This endpoint doesn't require authentication.

Please re-word my list above to your liking. But we need to hold the user's hand to explain our UI to have a sense of what the ui can do based on the endpoints we've secured. As well as the database info that helps tie the local and social stuff together.

emonddr: Also

image

examples/passport-login/src/services/user.service.ts
jannyHou: > UserIdentity is a term borrowed from the original loopback-component-passport , since this is a migration story, may be better to have it for now.

Fair enough 👍

examples/passport-login/src/__tests__/acceptance/passport-login.acceptance.ts
```diff
@@ -0,0 +1,406 @@
...
+        expect(url.parse(callbackToLbApp).pathname).to.equal(
+          '/api/auth/thirdparty/oauth2/callback',
+        );
+      });
+
+      it('callback url contains access code', async () => {
+        expect(url.parse(callbackToLbApp).query).to.containEql('code');
+      });
+
+      it('access code can be exchanged for token', async () => {
```
jannyHou, line 230: hmm how does the code and token exchange get tested in it?

examples/passport-login/src/authentication-strategies/google.ts
```diff
@@ -0,0 +1,73 @@
...
+   */
+  async authenticate(request: Request): Promise<UserProfile | RedirectRoute> {
+    return this.strategy.authenticate(request);
+  }
+
+  /**
+   * map passport profile to user profile
+   * @param user
+   */
+  mapProfile(user: User): UserProfile {
```
jannyHou, line 64: If the 4 `mapProfile` functions are same in the 3 strategies, you can extract them into utils.

examples/passport-login/src/authentication-strategies/oauth2.ts
jannyHou: @emonddr The auth provider should be decorated with `@extensionPoint`, see example in https://loopback.io/doc/en/lb4/Loopback-component-authentication.html#registering-a-custom-authentication-strategy

The class here is just an auth strategy.

jannyHou: @emonddr @deepakrkris See the comment in https://github.com/strongloop/loopback-next/blob/master/packages/core/src/extension-point.ts#L67-L71

The extension name can be provided directly by the 1st param in `@extensions()`, `@extensionPoint()` is another way to provide the name.
That's why Deepak's code works.

examples/passport-login/src/authentication-strategies/session.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) {}
+
+  /**
+   * authenticate a request
+   * @param request
+   */
+  async authenticate(
+    request: RequestWithSession,
+  ): Promise<UserProfile | RedirectRoute | undefined> {
+    if (!request.user) {
```
jannyHou, line 30: hmm why the session strategy doesn't retrieve the user from request's session but from request directly?

examples/passport-login/src/authentication-strategies/types.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export type VerifyFunction = (
+  accessToken: string,
+  refreshToken: string,
+  profile: Profile,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  done: (error: any, user?: any, info?: any) => void,
+) => void;
+
+export namespace PassportAuthenticationBindings {
```
jannyHou, line 25: please move the bindings into  `keys.ts` file.

examples/passport-login/src/controllers/oauth2.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  @authenticate('oauth2')
+  @get('/auth/thirdparty/{provider}')
+  /**
+   * Endpoint: '/auth/thirdparty/{provider}'
+   *          an endpoint for api clients to login via a third party app, redirects to third party app
+   */
+  loginToThirdParty(
+    @param.path.string('provider') provider: string,
+    @inject('authentication.redirect.url')
```
jannyHou, line 31: a reminder: doesn't have to apply in this PR, please create binding keys for the redirect url and status.

deepakrkris: agreed

examples/passport-login/src/authentication-strategies/types.ts
deepakrkris: agreed

examples/passport-login/src/authentication-strategies/session.ts
deepakrkris: agreed

examples/passport-login/src/authentication-strategies/google.ts
deepakrkris: agreed

examples/passport-login/src/__tests__/acceptance/passport-login.acceptance.ts
deepakrkris: authorization code exchange test is also done in acceptance tests of extension-passport

examples/passport-login/README.md
deepakrkris: @emonddr agreed, this is fixed now

deepakrkris: > @deepakrkris , thanks for making the UI changes and addressing my earlier comments.
> 
> Let's update this section with the latest UI changes in mind (some of the instructions are out of date.)
> 
> But also it is important to register a local user BEFORE any social media login, otherwise there will be issues.
> 
> 1. Open browser to localhost:3000
> 2. Sign up as a local user with the email of a facebook test user you created above, and provide a password for local purposes. This registers a local user, and it will eventually be linked to social media credentials in the following steps.
> 3. Login as this local user. You will be brought to the "View Account" tab ( and describe what they should see)
> 4. Click on Sign Up or Login, and the pages indicate you've already logged in.
> 5. Click on [Current User] button, and it calls https://localhost:3000/api/whoAmI endpoint and displays currently authenticated user.
> 6. Click on Logout.
> 7. Click on View Account, and you are not permitted since you are logged out.
> 8. Look at the db.json and mention what they should see in there. User with specific email, and in UserCredentials, the password you enterered when you signed up.
> 9. Now click on Login, and choose `Login with Facebook`. Enter the email of your facebook test user (same email you entered above), but specify the test user's facebook password (not local password).
> 10. Upon successful FaceBook login, you are brought to `View Account`. (Explain the data they should see)
> 11. Click on Sign Up or Login, and the pages indicate you are already logged in.
> 12. Click on [Current User] to view the currently authenticated user.
> 13. Click on Logout
> 14. Click on `View Account`, and you are not allowed.
> 15. Look at the db.json, there is still one user. It still has the UserCredentials entry as earlier, But now there is an entry in the UserIdentity section which contains the facebook profile info.
>     At any time (authenticated or not), open http://localhost:3000/api/ping in the browser and you should see the ping info returned. This endpoint doesn't require authentication.
> 
> Please re-word my list above to your liking. But we need to hold the user's hand to explain our UI to have a sense of what the ui can do based on the endpoints we've secured. As well as the database info that helps tie the local and social stuff together.

@emonddr I have incorporated these steps as suitable for user readability. please review.

examples/passport-login/index.ts
deepakrkris: agreed, the command `../../packages/cli/bin/cli-main.js copyright` is run and all files have headers now. thank you.

PR #4917, opened 2020-03-18
feat(core): add more flavors of @extensions decorator

raymondfeng:
This allows a getter, a view, or a list of extensions to be injected.

## Checklist

### DIFF
--- a/docs/site/Extension-point-and-extensions.md
+++ b/docs/site/Extension-point-and-extensions.md
@@ -50,6 +50,11 @@ decorators and functions are provided to ensure consistency and convention.
   custom name
 - `@extensions`: injects a getter function to access extensions to the target
   extension point
+- `@extensions.view`: injects a context view to access extensions to the target
+  extension point. The view can be listened for context events.
+- `@extensions.list`: injects an array of extensions to the target extension
+  point. The list is fixed when the injection is done and it does not add or
+  remove extensions afterward.
 - `extensionFilter`: creates a binding filter function to find extensions for
   the named extension point
 - `extensionFor`: creates a binding template function to set the binding to be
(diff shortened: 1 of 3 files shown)


docs/site/Extension-point-and-extensions.md
```diff
@@ -58,7 +63,47 @@ decorators and functions are provided to ensure consistency and convention.
...
+1. Inject a getter function for extensions
+
+   ```ts
+   import {Getter} from '@loopback/context';
+   import {extensionPoint, extensions} from '@loopback/core';
+
+   @extensionPoint('greeters')
+   class GreetingService {
+     @extensions()
+     public getGreeters: Getter<Greeter[]>;
```
emonddr: If I have the traditional way of obtaining all the extensions using getGreeters(), why do we need
@extensions.view() or @extensions.list()? What is the advantage of having a context view compared to not having a context view? Or the list instead of the view.

The developer can call:

const greeters = await greeterService.getGreeters();
OR
const greeters = await greeterService.greetersView.values();
OR
const greeters = greeterService.greeters;

Sometimes too many choices is confusing, so a clear explanation for the purpose of each is important.

raymondfeng: The difference is not at fetching the list of greeters.

- getter: picks up latest list of greeters
- view: gives more control to handle fine-grained update of greeters (in addition to getter). I'm adding this feature to support a use case in extension point/extension pattern to allow fine-grained handling of an extension being removed. For example, if a cron job is removed after the app starts, we can stop the resolved instance if it exists.
- list: only contains a snapshot of greeters (no updates will be picked up. some extension point might want to have this behavior).

emonddr: @raymondfeng , thanks for the clarifications. This makes sense now. :)

emonddr: @raymondfeng , perhaps `beef up` the docs with these examples to help clarify the choices. thx.

Issue #4902, opened 2020-03-17
Support for oauth2 passport strategies, missing provisions for redirect and callback


deepakrkris:
## Suggestion

The StrategyAdapter is not adding the redirect() method in passport strategies, required for Oauth2 authorization flows.
This would also need some handling on the invocation side, i.e the Sequence class and others might need to participate according to the oauth flow.

## Use Cases

login thru third party apps like facebook, google
https://github.com/strongloop/loopback-next/issues/3958

## Acceptance criteria

- [ ] - strategy adapter must add all necessary calls for oauth2 in passport strategies
- [ ] - unit tests with mock redirect and mock third party app
- [ ] - must satisfy https://github.com/strongloop/loopback-next/issues/3958 for login with third party apps

dhmlau:
related to https://github.com/strongloop/loopback-next/issues/2624

deepakrkris:
PR submitted for review https://github.com/strongloop/loopback-next/pull/4919

Issue #2624, opened 2019-03-21
[Spike] Explore/research the oauth2+session based authentication strategy


jannyHou:
## Description / Steps to reproduce / Feature proposal

A follow-up story for PR https://github.com/strongloop/loopback-next/pull/2576

Add new authentication strategies in `loopback4-example-shopping` to investigate how to verify a 3rd-party user using oauth2, and track the user using session.

## Acceptance Criteria:

*The acceptance criteria is written with oauth2+google, the story owner can choose any 3rd party auth provider: oauth2+fb, oauth2+google, etc...*

- [x] Create a static login page that allows user to choose from local login or 3rd-party login. Keep the UI as simple as possible
- [x]  ~Create an oauth2 strategy to authenticate the users by interacting with 3rd-party provider's auth server~ completed with https://github.com/strongloop/loopback-next/pull/4919 
  - [x] Please note that usually 3rd-party login requires a callback URL, which implies 2 endpoints are involved in the login flow. E.g. `login/facebook` and `login/facebook/callback`. And this is the most important user experience we need to figure out from design's perspective.

- [x] Track the user profile using session. Keep the session related functionalities as simple as possible, we can create sub-stories to enhance.
- [x] Create an example endpoint that verifies user using session.
~- [ ] Update the md file `docs/strategies/oauth2.md`~

emonddr:
After talking with @raymondfeng  yesterday, he suggested we investigate the two main flows of OAuth 

 - Implicit Flow 
 - Authorization Code Flow 

`Implicit Flow` is when `all` of the logic for authentication is in the UI flow, and the goal is to get the access token, and then pass this token to the backend when accessing endpoints. 

https://auth0.com/docs/flows/concepts/implicit 

`Authorization Code Flow` is when the `backend` server has to direct the user to login, and the oauth server sends an  authorization code to the backend. The backend uses this temporary code to get an access token and passes it to the client. The client then has to use it to access all the backend's protected endpoints. 

https://auth0.com/docs/flows/concepts/auth-code 


With regards to https://www.npmjs.com/package/@loopback/authentication-passport , we know the adapter works well with passport-jwt and passport-http.

We need to adjust the design (if necessary) to ensure that it can work well with OAuth2.0  passport-* modules like passport-facebook, passport-google, etc.

emonddr:
Most people uses Facebook authentication to get a token to then invoke Facebook API.
This is not our goal per se.
[...]

Issue #1881, opened 2018-10-19
Docs: How to receive context in controller


diegolacarta:
Is there any way to receive the context inside a controller?

Goal is to access request data (headers, ip, etc.)

Thanks

## Acceptance criteria
- [ ] a section in [Controllers.html](https://loopback.io/doc/en/lb4/Controllers.html) explaining how to access request and response headers or body by using decorators like ` @inject(RestBindings.Http.REQUEST) public request: Request` for different cases:
  - [ ] A request header shared by most (or all) endpoints and processed by an express middleware/LB4 sequence action. Example: Authorization or X-API-Key.
  - [ ] A request header shared by many endpoints but used internally in the endpoint implementation (an express route handler/LB4 controller method). Maybe come up with a somewhat realistic example of such header.
  - [ ] A response header set by an express middleware/LB4 sequence action. For example X-Rate-Limit information provided by a rate-limiting extension in all responses.
  - [ ] A response header set by the endpoint (an express route handler/LB4 controller method). An example: Location header included ind 201 Created responses.

bajtos:
> Goal is to access request data (headers, ip, etc.)

Ideally, you should decouple the code building request data from your controllers to enable you to test the controllers in isolation (unit-test style).

Typically, request headers are mapped to regular method parameters, OpenAPI spec supports `in: header` source.

I think it makes sense to expose the client IP via a custom binding that will allow you to receive the IP address in your controller methods via dependency injection (e.g. `@inject(RestBindings.CLIENT_IP)`). This is an addition to be implemented in `@loopback/rest`, would you like to contribute it yourself? We can help you along the way.


> Is there any way to receive the context inside a controller?

Since you are interested in the Request object only, you can configure your controller to receive the request as follows:

```ts
class MyController {
  constructor(
    @inject(RestBindings.Http.REQUEST) public request: Request,
    // ...
  ) {}

  @get('/foo')
  foo() {
    console.log(this.request.headers);
  }
}
```

Accessing the context directly should be the last resort solution. You can receive the current context using [`@inject.context()`](http://apidocs.loopback.io/@loopback%2fdocs/context.html#inject.context).

```ts
class MyController {
  constructor(
    @inject.context() public context: RequestContext,
    // ...
  ) {}

  @get('/foo')
  foo() {
    console.log(this.context.request.headers);
  }
}
```

bajtos:
[...]

PR #4919, opened 2020-03-18
fix: passport strategy adapter must support oauth2 flows

deepakrkris:
fixes: https://github.com/strongloop/loopback-next/issues/4902

**highlights of changes** :

* `packages/rest/src/writer.ts` - modified to handle redirections, controller can choose to redirect an api call
* `extensions/authentication-passport/src/strategy-adapter.ts ` - adapter is modified to handle redirection calls in passport strategy
* `extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts`
        - Mock oauth2 provider to mock the whole authorization flow
* `extensions/authentication-passport/src/_tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts`
        - test cases for the whole  authorization flow

## Checklist

### DIFF
--- a/extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
+++ b/extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
@@ -1,0 +1,331 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/authentication-passport
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {UserProfileFactory, authenticate} from '@loopback/authentication';
+import {
+  Strategy as Oauth2Strategy,
+  StrategyOptions,
+  VerifyFunction,
+  VerifyCallback,
+} from 'passport-oauth2';
+import {MyUser, userRepository} from './fixtures/user-repository';
+import {
+  simpleRestApplication,
+  configureApplication,
+} from './fixtures/simple-rest-app';
+import {securityId, UserProfile, SecurityBindings} from '@loopback/security';
+import {StrategyAdapter} from '../../strategy-adapter';
+import {get} from '@loopback/openapi-v3';
+import {
+  Client,
+  createClientForHandler,
+  expect,
+  supertest,
+} from '@loopback/testlab';
+import {RestApplication, RestBindings, Response} from '@loopback/rest';
+import {
+  startApp as startMockOauth2Server,
+  stopApp as stopMockOauth2Server,
+} from './fixtures/mock-oauth2-social-app';
+import * as url from 'url';
+import {inject} from '@loopback/core';
+import axios from 'axios';
+import qs from 'qs';
+
+/**
+ * This test consists of three main components -> the supertest client, the LoopBack app (simple-rest-app.ts)
+ * and the Mock Authorization Server (mock-oauth2-social-app.ts)
+ *
+ *  Mock Authorization Server (mock-oauth2-social-app.ts) :
+ *           mocks the authorization flow login with a social app like facebook, google, etc
+ *  LoopBack app (simple-rest-app.ts) :
+ *           has a simple app with no controller, Oauth2Controller is added in this test
+ *
+ * Test steps:
+ *
+ * A. an Oauth2Controller is added to LB App : defines two endpoints `/auth/thirdparty` and `/auth/thirdparty/callback`
+ *
+ * B. start LB app and Mock Authorization Server
+ *
+ * C. Test stage 1 : Authorization code grant - Get access code
+ *                   from below diagram check flows - [1, 2, 3, 4, 5]
+ *                   [1] - test end point `/auth/thirdparty`
+ *                   [2, 3] - test redirection to mock auth server login page
+ *                   [4, 5] - test login with mock server redirects to callback url
+ *
+ * D. Test stage 2 : Authentication - Exchange access code for access token
+ *                   from below diagram check flows - [6, 7, 8]
+ *                   [5, 6] - test callback endpoint `/auth/thirdparty/callback`
+ *                   [7, 8] - check if valid access token was fetched
+ */
+
+//  +------------+                                                              +--------------+
+//  |            |                                                              | LoopBack App |
+//  | web client |     -------------------[1]--------------------->             | (simple-rest |
+//  | (supertest)|     auth request to LB App on behalf of a user,              |  -app.ts)    |
+//  |            |     payload: {'client_id': , 'secret': }                     | ****         |
+//  |            |                                                              |  ^           |
+//  |            |              +---------------+                               |  |           |
+//  |            |              |               | <---------[2]-------------    |  |           |
+//  |            |              | Mock          | LB App redirects browser      |  |           |
+//  |            |              | Authorization | to auth server,payload:       |  |           |
+//  |            |              | Server        | {'client_id':,                |  |           |
+//  |            |              | (mock-oauth2- |     'callback_url': app url } |  Stage 1     |
+//  |            |              | social-app.ts)|                               |  |           |
+//  |            |              |               |----+ auth server redirects    |  |           |
+//  |            |              |               |    | browser to login page,   |  |           |
+//  |            |              |               |  [3] client_id and            |  |           |
+//  |            |              |               |    | callback_url are         |  |           |
+//  |            |              |               |<---+ passed as hidden params  |  |           |
+//  |            |              |               |                               |  |           |
+//  |            | -----[4]---> |               |                               |  v           |
+//  |            |   login with |               | -------[5]------------->      | ****         |
+//  |            |    user name |               | login success, auth server    |  ^           |
+//  |            |    /password |               | redirects browser to callback |  |           |
+//  |            |              |               | url with access_code          |  |           |
+//  |            |              |               |                               |  |           |
+//  |            |              |               | <-------------[6]---------    |  |           |
+//  |            |              |               |  LB app requests access token |  Stage 2     |
+//  |            |              |               |  with access_code             |  |           |
+//  |            |              |               |                               |  |           |
+//  |            |              |               | ---------------[7]--------->  |  |           |
+//  |            |              +---------------+       returns access token    |  |           |
+//  |            |                                                              |  v           |
+//  |            |      <------------------------[8]-----------------------     | ****         |
+//  +------------+        LB App returns to browser the access token            +--------------+
+
+/**
+ * options to pass to the Passport Strategy
+ */
+const oauth2Options: StrategyOptions = {
+  clientID: '1111',
+  clientSecret: '1917e2b73a87fd0c9a92afab64f6c8d4',
+  // `redirect_uri`:
+  // 'passport-oauth2' takes care of sending the configured `callBackURL` setting as `redirect_uri`
+  // to the authorization server. This behaviour is inherited by all other oauth2 modules like facebook, google, etc
+  callbackURL: 'http://localhost:8080/auth/thirdparty/callback',
+  // 'authorizationURL' is used by 'passport-oauth2' to begin the authorization grant flow
+  authorizationURL: 'http://localhost:9000/oauth/dialog',
+  // `tokenURL` is used by 'passport-oauth2' to exchange the access code for an access token
+  // this exchange is taken care internally by 'passport-oauth2'
+  // when the `callbackURL` is invoked by the authorization server
+  tokenURL: 'http://localhost:9000/oauth/token',
+};
+
+/**
+ * verify function for the oauth2 strategy
+ * This function mocks a lookup against a user profile datastore
+ *
+ * @param accessToken
+ * @param refreshToken
+ * @param profile
+ * @param done
+ */
+const verify: VerifyFunction = function (
+  accessToken: string,
+  refreshToken: string,
+  userProfile: MyUser,
+  done: VerifyCallback,
+) {
+  userProfile.token = accessToken;
+  if (!userRepository.findById(userProfile.id)) {
+    userRepository.createExternalUser(userProfile);
+  }
+  return done(null, userProfile);
+};
+
+/**
+ * convert user info to user profile
+ * @param user
+ */
+const myUserProfileFactory: UserProfileFactory<MyUser> = function (
+  user: MyUser,
+): UserProfile {
+  const userProfile = {[securityId]: user.id, ...user};
+  return userProfile;
+};
+
+/**
+ * Login controller for third party oauth provider
+ *
+ * This creates an authentication endpoint for the third party oauth provider
+ *
+ * Two methods are expected
+ *
+ * 1. loginToThirdParty
+ *           i. an endpoint for api clients to login via a third party app
+ *          ii. the passport strategy identifies this call as a redirection to third party
+ *         iii. this endpoint redirects to the third party authorization url
+ *
+ * 2. thirdPartyCallBack
+ *           i. this is the callback for the thirdparty app
+ *          ii. on successful user login the third party calls this endpoint with an access code
+ *         iii. the passport oauth2 strategy exchanges the code for an access token
+ *          iv. the passport oauth2 strategy then calls the provided `verify()` function with the access token
+ */
+export class Oauth2Controller {
+  constructor() {}
+
+  // this configures the oauth2 strategy
+  @authenticate('oauth2')
+  // we have modeled this as a GET endpoint
+  @get('/auth/thirdparty')
+  // loginToThirdParty() is the handler for '/auth/thirdparty'
+  // this method is injected with redirect url and status
+  // the value for 'authentication.redirect.url' is set by the authentication action provider
+  loginToThirdParty(
+    @inject('authentication.redirect.url')
+    redirectUrl: string,
+    @inject('authentication.redirect.status')
+    status: number,
+    @inject(RestBindings.Http.RESPONSE)
+    response: Response,
+  ) {
+    // controller handles redirect
+    // and returns response object to indicate response is already handled
+    response.statusCode = status || 302;
+    response.setHeader('Location', redirectUrl);
+    response.end();
+    return response;
+  }
+
+  // we configure the callback url also with the same oauth2 strategy
+  @authenticate('oauth2')
+  // this SHOULD be a GET call so that the third party can ask the browser to redirect
+  @get('/auth/thirdparty/callback')
+  // thirdPartyCallBack() is the handler for '/auth/thirdparty/callback'
+  // the oauth2 strategy identifies this as a callback with the request.query.code sent by the third party app
+  // the oauth2 strategy exchanges the access code for a access token and then calls the provided verify() function
+  // the verify function creates a user profile after verifying the access token
+  thirdPartyCallBack(@inject(SecurityBindings.USER) user: UserProfile) {
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    return {access_token: user.token};
+  }
+}
+
+describe('Oauth2 authorization flow', () => {
+  let app: RestApplication;
+  let oauth2Strategy: StrategyAdapter<MyUser>;
+  let client: Client;
+
+  before(startMockOauth2Server);
+  after(stopMockOauth2Server);
+
+  before(givenLoopBackApp);
+  before(givenOauth2Strategy);
+  before(setupAuthentication);
+  before(givenControllerInApp);
+  before(givenClient);
+
+  let oauthProviderUrl: string;
+  let providerLoginUrl: string;
+  let callbackToLbApp: string;
+  let loginPageParams: string;
+
+  context('Stage 1 - Authorization code grant', () => {
+    describe('when client invokes oauth flow', () => {
+      it('call is redirected to third party authorization url', async () => {
+        // HTTP status code 303 means see other,
+        // on seeing which the browser would redirect to the other location
+        const response = await client.get('/auth/thirdparty').expect(303);
+        oauthProviderUrl = response.get('Location');
+        expect(url.parse(response.get('Location')).pathname).to.equal(
+          url.parse(oauth2Options.authorizationURL).pathname,
+        );
+      });
+
+      it('call to authorization url is redirected to oauth providers login page', async () => {
+        // HTTP status code 302 means redirect to new uri,
+        // on seeing which the browser would redirect to the new uri
+        const response = await supertest('').get(oauthProviderUrl).expect(302);
+        providerLoginUrl = response.get('Location');
+        loginPageParams = url.parse(providerLoginUrl).query ?? '';
+        expect(url.parse(response.get('Location')).pathname).to.equal('/login');
+      });
+
+      it('login page redirects to authorization app callback endpoint', async () => {
+        const loginPageHiddenParams = qs.parse(loginPageParams);
+        const params = {
+          username: 'user1',
+          password: 'abc',
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          client_id: loginPageHiddenParams.client_id,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          redirect_uri: loginPageHiddenParams.redirect_uri,
+          scope: loginPageHiddenParams.scope,
+        };
+        // On successful login, the authorizing app redirects to the callback url
+        // HTTP status code 302 is returned to the browser
+        const response = await supertest('')
+          .post('http://localhost:9000/login_submit')
+          .send(qs.stringify(params))
+          .expect(302);
+        callbackToLbApp = response.get('Location');
+        expect(url.parse(callbackToLbApp).pathname).to.equal(
+          '/auth/thirdparty/callback',
+        );
+      });
+
+      it('callback url contains access code', async () => {
+        expect(url.parse(callbackToLbApp).query).to.containEql('code');
+      });
+    });
+  });
+
+  context('Stage 2: Authentication', () => {
+    describe('Invoking call back url returns access token', () => {
+      it('access code can be exchanged for token', async () => {
+        expect(url.parse(callbackToLbApp).path).to.containEql(
+          '/auth/thirdparty/callback',
+        );
+        const path: string = url.parse(callbackToLbApp).path ?? '';
+        const response = await client.get(path).expect(200);
+        expect(response.body).property('access_token');
+      });
+    });
+  });
+
+  function givenLoopBackApp() {
+    app = simpleRestApplication();
+  }
+
+  function givenOauth2Strategy() {
+    const passport = new Oauth2Strategy(oauth2Options, verify);
+
+    // passport-oauth2 base class leaves user profile creation to subclass implementations
+    passport.userProfile = (accessToken, done) => {
+      // call the profile url in the mock authorization app with the accessToken
+      axios
+        .get('http://localhost:9000/verify?access_token=' + accessToken, {
+          headers: {Authorization: accessToken},
+        })
+        .then(response => {
+          done(null, response.data);
+        })
+        .catch(err => {
+          done(err);
+        });
+    };
+
+    // use strategy adapter to fit passport strategy to LoopBack app
+    oauth2Strategy = new StrategyAdapter(
+      passport,
+      'oauth2',
+      myUserProfileFactory,
+    );
+  }
+
+  async function setupAuthentication() {
+    await configureApplication(oauth2Strategy, 'oauth2');
+  }
+
+  function givenControllerInApp() {
+    return app.controller(Oauth2Controller);
+  }
+
+  function givenClient() {
+    client = createClientForHandler(app.requestHandler);
+  }
+});
--- a/extensions/authentication-passport/src/__tests__/acceptance/fixtures/mock-oauth2-social-app.ts
+++ b/extensions/authentication-passport/src/__tests__/acceptance/fixtures/mock-oauth2-social-app.ts
@@ -1,0 +1,334 @@
+// Copyright IBM Corp. 2018,2019. All Rights Reserved.
+// Node module: @loopback/authentication-passport
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ *  Mock Authorization Server:
+ *           mocks the authorization flow with a social app login like facebook, google, etc
+ *
+ *  Endpoints :
+ *        `/oauth/dialog` -  opens the oauth2 flow, redirects to login page
+ *        `/login` - loads the login page
+ *        `/login_submit` - submit username , password
+ *        `/oauth/token` - returns a token in exchange for a valid authorization code
+ *        `/verify` - verifies token
+ */
+
+'use strict';
+
+import express from 'express';
+import {Server} from 'http';
+const jwt = require('jsonwebtoken');
+const bodyParser = require('body-parser');
+const _ = require('lodash');
+import {MyUser} from './user-repository';
+
+const app = express();
+let server: Server;
+
+// to support json payload in body
+app.use('parse', bodyParser.json());
+// to support html form bodies
+app.use(bodyParser.text({type: 'text/html'}));
+// create application/x-www-form-urlencoded parser
+const urlencodedParser = bodyParser.urlencoded({extended: false});
+
+interface JWT {
+  payload: {
+    jti: string;
+    client_id: string;
+  };
+}
+
+/**
+ * datastructure for an app registration, also holds issued tokens for an app
+ */
+interface App {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
+  client_secret: string;
+  tokens: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [key: string]: any;
+  };
+}
+
+/**
+ * list of registered apps for this oauth2 provider identified by their client ids
+ */
+interface AppRegistry {
+  [clientId: string]: App;
+}
+
+/**
+ * apps registered with this provider
+ * format:
+ *   { clientId: {client_secret, list_of_tokens} }
+ */
+const registeredApps: AppRegistry = {
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  '1111': {client_secret: 'app1_secret', tokens: {}},
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  '2222': {client_secret: 'app2_secret', tokens: {}},
+};
+
+/**
+ * user registry
+ */
+const users = [
+  {
+    id: 1001,
+    username: 'user1',
+    password: 'abc',
+    email: 'usr1@lb.com',
+    signingKey: 'AZeb==',
+  },
+  {
+    id: 1002,
+    username: 'user2',
+    password: 'xyz',
+    email: 'usr2@lb2.com',
+    signingKey: 'BuIx=+',
+  },
+];
+
+/**
+ * find a user by a name and password
+ * @param {*} username
+ * @param {*} password
+ */
+function findUser(username: string, password: string) {
+  const usr = _.filter(
+    users,
+    (user: MyUser) => user.username === username && user.password === password,
+  );
+  if (usr.length > 0) return usr[0];
+  return null;
+}
+
+/**
+ * create a jwt token
+ * @param {*} user
+ * @param {*} scopes
+ * @param {*} signingKey
+ */
+async function createJwt(
+  user: MyUser,
+  scopes: string,
+  signingKey: string,
+  clientId: string,
+) {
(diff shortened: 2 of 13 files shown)


extensions/authentication-passport/src/__tests__/acceptance/fixtures/simple-rest-app.ts
```diff
@@ -0,0 +1,82 @@
...
+  app.component(AuthenticationComponent);
+  return app;
+}
+
+export async function configureApplication(authStrategy: StrategyAdapter<MyUser>, authKey: string) {
+  server = await app.getServer(RestServer);
+  app
+    .bind(authKey)
+    .to(authStrategy)
+    .tag({
```
raymondfeng: Please use `apply(asAuthStrategy)` instead.

packages/rest/src/types.ts
```diff
@@ -175,7 +175,7 @@ export type OperationArgs = any[];
  * This is a type alias for "any", used to distinguish
  * operation results from other "any" typed values.
  */
-export type OperationRetval = any;
+export type OperationRetval = any | RedirectRoute;
```
raymondfeng: `any` already allows `RedirectRoute`.

packages/rest/src/writer.ts
```diff
@@ -30,6 +31,16 @@ export function writeResultToResponse(
     return;
   }
 
+  // result is of type RedirectRoute
+  const isRedirect = result instanceof RedirectRoute;
```
raymondfeng: `instanceof` is not always reliable as the `RedirectRoute` can be loaded from different places. Let's use a TypeScript type guard function instead.

packages/rest/src/writer.ts
```diff
@@ -30,6 +31,16 @@ export function writeResultToResponse(
...
+  // Controller methods can return a redirect handler to handle redirections.
+  // This will be useful in certain oauth flows like `authorization code`
+  // , where the authenticating endpoint needs to redirect to a third-party oauth provider
+  if (isRedirect) {
```
raymondfeng: We don't have to handle `RedirectRoute` operation result in the `writer.ts`.  The controller can handle redirect and simple return `response` as is to skip response serialization.

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,194 @@
...
+  constructor() {}
+
+  // this configures the oauth2 strategy
+  @authenticate('oauth2')
+  // we have modeled this as a GET endpoint
+  @get('/auth/thirdparty')
+  // loginToThirdParty() is the handler for '/auth/thirdparty'
+  // this method is injected with 'x-loopback-authentication-redirect-url'
+  // the value for 'x-loopback-authentication-redirect-url' is set by the passport strategy adapter
+  loginToThirdParty(@param.query.string('x-loopback-authentication-redirect-url') redirectUrl: string,
```
raymondfeng: I wonder if we can inject such information from `RequestContext` instead of query string.

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
```diff
@@ -0,0 +1,253 @@
+// Copyright IBM Corp. 2018,2019. All Rights Reserved.
```
emonddr: Perhaps use a different term than provider? Provider has a particular meaning and usage in LB4. Just a thought.

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
```diff
@@ -0,0 +1,253 @@
...
+// Node module: @loopback/cli
```
emonddr: > Node module: @loopback/cli 

Is this really in cli ?

extensions/authentication-passport/src/strategy-adapter.ts
```diff
@@ -73,6 +73,18 @@ export class StrategyAdapter<U> implements AuthenticationStrategy {
...
+      // handle redirection for oauth2 authorization flows
+      strategy.redirect = function(url: string, status: number) {
+        // set redirect options in query
+        // the controller configured with the oauth2 strategy will have to handle actual redirect
+        // the controller can access these values from the request
+        request.query['x-loopback-authentication-redirect-url'] = url;
+        request.query['x-loopback-authentication-redirect-status'] = status;
+        request.query['x-loopback-authentication-redirect-context'] = this.name;
+        // resolve with dummy user profile
+        resolve({[securityId]: ''});
```
emonddr: What does 
> resolve({[securityId]: ''}); 

do exactly; in the part it plays in the auth flow? thx

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -4,7 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {Getter, inject, Provider, Setter} from '@loopback/context';
-import {Request} from '@loopback/rest';
+import {Request, RequestContext} from '@loopback/rest';
```
emonddr: Why is this needed?

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
```diff
@@ -0,0 +1,253 @@
...
+        res.send("invalid app");
+    }
+});
+
+/**
+ * login page
+ * 
+ * handles login part of the authorization call
+ */
+app.get('/login', function(req, response) {
```
emonddr: Is this code attempting to simulate an oauth2 provider like google and facebook, and you are just using it as a stub for your mocha test cases? Please add some comments to explain the purpose of this file and its methods. thank you.

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,194 @@
...
+    it('call is redirected to third party authorization url', async () => {
+      const response = await client.get('/auth/thirdparty').expect(303);
+      oauthProviderUrl = response.get('Location');
+      expect(url.parse(response.get('Location')).pathname).to.equal(url.parse(oauth2Options.authorizationURL).pathname);
+    });
+
+    it('call to authorization url is redirected to oauth providers login page', async () => {
+      const response = await supertest('').get(oauthProviderUrl).expect(302);
+      providerLoginUrl = response.get('Location');
+      expect(url.parse(response.get('Location')).pathname).to.equal('/login');
```
emonddr: So to be clear, you are running an http server that is capable of hosting these urls, but we are only testing the calling of these endpoints, but obviously not interacting with the html fields (user and password and submit).

emonddr:
@deepakrkris , thanks for tackling this :)
Great job.
How does the express app interact with the LB4 app exactly? I know the mocha tests make sure that these express endpoints are called. But I am having a hard time seeing how the mocha tests test the entire interaction of the UI and server (entering a user name and password. Is the oauth-provider express app a UI that user will have to code, and the only LB4 (backend) code that needs to be done is OAuth2 strategy adapter with redirect method, and a few tiny changes to the LB4 rest layer?

packages/rest/src/types.ts
deepakrkris: agreed

extensions/authentication-passport/src/__tests__/acceptance/fixtures/simple-rest-app.ts
deepakrkris: agreed

packages/rest/src/writer.ts
deepakrkris: agreed

packages/rest/src/writer.ts
deepakrkris: agreed

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,194 @@
...
+  // the oauth2 strategy identifies this as a callback with the request.query.code sent by the third party app
+  // the oauth2 strategy exchanges the access code for a access token and then calls the provided verify() function
+  // the verify function creates a user profile after verifying the access token
+  thirdPartyCallBack(@inject(SecurityBindings.USER) user: UserProfile) {
+    console.log('thirdPartyCallBack', user);
+    return user.token;
+  }
+}
+
+describe.only('Oauth2 authorization flow', () => {
```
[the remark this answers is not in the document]
deepakrkris: agreed

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,194 @@
...
+  before(givenClient);
+
+  let oauthProviderUrl: string;
+  let providerLoginUrl: string;
+  let callbackToLbApp: string;
+
+  context('when client invokes oauth flow', () => {
+
+    it('call is redirected to third party authorization url', async () => {
+      const response = await client.get('/auth/thirdparty').expect(303);
```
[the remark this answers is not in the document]
deepakrkris: agreed

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
deepakrkris: agreed

deepakrkris: I am thinking of `MockOauth2LoginApp` or `MockSocialMediaApp`

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
deepakrkris: agreed, copy paste error

extensions/authentication-passport/src/strategy-adapter.ts
deepakrkris: the redirect() function cannot return a user profile ,  but the authenticate function's contract does not allow an empty resolve, I can resolve with a null value like `resolve(null)`

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
deepakrkris: yes exactly this is a stub for a social media app, in fact it works with your express facebook example, https://github.com/emonddr/express-4.x-facebook-example , when I replaced all urls. You can start both apps and login redirects to the mock oauth2 provider, enter user id , password and then see it redirected to your app with an oauth token.

deepakrkris: I will add necessary comments to make this clear

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
deepakrkris: the test for login is also done , please see 
https://github.com/strongloop/loopback-next/blob/75a02cabe61be1414a22084984ec87e1982ee17c/extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts#L148-L152

deepakrkris: the entire oauth2 auth flow is tested, we only test deliberately for redirected urls, because browsers handle that redirection when seeing a 302 or 303.

extensions/authentication-passport/src/strategy-adapter.ts
deepakrkris: I have now changes the authentication strategy to resolve either UserProfile or RedirectRoute

deepakrkris:
all tests passed , except for a minor code lint error https://github.com/strongloop/loopback-next/runs/531315573 , I have added a second commit to fix the code lint error.

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: agreed

packages/rest/src/writer.ts
deepakrkris: @raymondfeng on second thought, I think having the controller return a simple redirect object makes the design very neat and clear.

deepakrkris: this would also make user code more unit testable.

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
emonddr: > yes exactly this is a stub for a social media app, in fact it works with your express facebook example, https://github.com/emonddr/express-4.x-facebook-example , when I replaced all urls. You can start both apps and login redirects to the mock oauth2 provider, enter user id , password and then see it redirected to your app with an oauth token.

haha cool. How did you find my app?

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
emonddr: I see you are mocking that the form was filled by user by providing the parameters in the url. But there is no UI robot clicking and typing...is what I meant.

packages/rest/src/writer.ts
```diff
@@ -30,6 +31,16 @@ export function writeResultToResponse(
     return;
   }
 
+  // result is of type RedirectRoute
+  const isRedirect = RedirectRoute.isRedirectRoute(result);
+  // Controller methods can return a redirect handler to handle redirections.
+  // This will be useful in certain oauth flows like `authorization code`
+  // , where the authenticating endpoint needs to redirect to a third-party oauth provider
+  if (isRedirect) {
+    const redirectHandler = result as RedirectRoute;
```
raymondfeng: The type guard should already tell TS compiler that `result` is an instance of `RedirectRoute`. We should be able to use `result` without casting.

packages/rest/src/writer.ts
raymondfeng: I'm not a big fan of hard-coding a lot of flavors in the writer unless we introduce extensibility of response handlers. Meanwhile, I won't block the PR either.

packages/rest/src/router/redirect-route.ts
```diff
@@ -37,11 +42,54 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
...
+   * get target location
+   */
+  getTargetLocation() {
+    return this.targetLocation;
+  }
+
+  /**
+   * get status code
+   */
+  getStatusCode() {
```
raymondfeng: Don't introduce java-like getter here. Simply change the property to be `public readonly statusCode`.

packages/rest/src/router/redirect-route.ts
```diff
@@ -37,11 +42,54 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
...
   describe(): string {
     return `RedirectRoute from "${this.sourcePath}" to "${this.targetLocation}"`;
   }
+
+  /**
+   * get target location
+   */
+  getTargetLocation() {
+    return this.targetLocation;
+  }
```
raymondfeng: Don't introduce java-like getter here. Simply change the property to be `public readonly targetLocation`.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -49,8 +49,28 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
...
+      // set the redirect options in the request query
+      // set redirection url as 'x-loopback-authentication-redirect-url'
+      // set redirection status as 'x-loopback-authentication-redirect-status'
+      // controller will handle actual redirection
+      request.query[
+        'x-loopback-authentication-redirect-url'
+      ] = redirectOptions.getTargetLocation();
+      request.query[
+        'x-loopback-authentication-redirect-status'
+      ] = redirectOptions.getStatusCode();
```
raymondfeng: We should avoid mutating `request` as much as possible.

extensions/authentication-passport/src/__tests__/acceptance/fixtures/oauth2-provider.ts
deepakrkris: you had commented in issue https://github.com/strongloop/loopback-next/issues/2624 about your poc.

packages/authentication/src/providers/auth-action.provider.ts
raymondfeng: What about the following?

```ts
requestCtx.bind('authentication.oauth2.redirectUrl').to(...);
```

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -49,8 +49,22 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
...
+
+    if (RedirectRoute.isRedirectRoute(authResponse)) {
+      const redirectOptions = authResponse as RedirectRoute;
+      request.query[
+        'x-loopback-authentication-redirect-url'
+      ] = redirectOptions.getTargetLocation();
+      request.query[
+        'x-loopback-authentication-redirect-status'
+      ] = redirectOptions.getStatusCode();
+    } else if (authResponse) {
```
emonddr: should we have a type-check here to make sure it is `UserProfile` ?

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: this means the request context will have to be be injected into the controller, I saw Miroslav mentioning that as an anti pattern in one of the other issues , https://github.com/strongloop/loopback-next/issues/1881#issuecomment-534596582

packages/rest/src/router/redirect-route.ts
deepakrkris: agreed

packages/rest/src/router/redirect-route.ts
deepakrkris: agreed

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: ah, we dont have to inject the whole request context, we can just inject the url value alone. you are right. I will make this change.

deepakrkris: agreed

packages/rest/src/writer.ts
deepakrkris: @raymondfeng to an extent I do agree with your view. I will make the change.
I am also going to see what other reviewers comment.

packages/rest/src/writer.ts
deepakrkris: agreed

extensions/authentication-passport/src/strategy-adapter.ts
```diff
@@ -42,9 +42,9 @@ export class StrategyAdapter<U> implements AuthenticationStrategy {
    *     3. authenticate using the strategy
    * @param request The incoming request.
    */
-  authenticate(request: Request): Promise<UserProfile> {
+  authenticate(request: Request): Promise<UserProfile | RedirectRoute> {
```
jannyHou: @deepakrkris could you explain more about why we have to allow returning a `RedirectRoute` here?

According to the [passport-oauth2](https://github.com/jaredhanson/passport-oauth2#configure-strategy) module, it returns a user (which conceptually same as lb's `userProfile`) after the authentication done, wondering why we couldn't follow the same pattern here?

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,322 @@
...
+ *           i. this is the callback for the thirdparty app
+ *          ii. on successful user login the third party calls this endpoint with an access code
+ *         iii. the passport oauth2 strategy exchanges the code for an access token
+ *          iv. the passport oauth2 strategy then calls the provided `verify()` function with the access token
+ */
+export class Oauth2Controller {
+  constructor() {}
+
+  // this configures the oauth2 strategy
+  @authenticate('oauth2')
```
jannyHou: @deepakrkris related to my comment in []()
I think the scenario here is different than what's expected from our authentication system...user profile injection should happen here, which means after the oauth2 strategy is invoked, a user profile should be returned as the result of the authentication action. 
My confusion point is, why we have to handle the redirect then get the user profile?

My expectation would be, the user profile is injected at controller level instead of the endpoint level. So that when the controller functions are invoked, the user profile is ready to use, both `/thirdparty` and `/callback` can print it.

I am not very familiar with the newly created `redirectRoute` therefore having a hard time to illustrate which part should handle the callback redirection, the strategy? or the controller?

extensions/authentication-passport/src/strategy-adapter.ts
deepakrkris: The `configure-strategy` deep-link you have commented above, is only to set the list of strategies in the base 'passport' module (which is heavily dependent on express).
The `passport.authenticate()` method that is handling the [express routes] (https://github.com/jaredhanson/passport-oauth2#authenticate-requests) interprets the incoming request and does the redirection. The base 'passport' module implements this as an express middleware without any code written by the user. For LoopBack, the strategy adapter will have to handle the redirect callback from the passport-strategy (done by the base `passport` module in express)  + we also need the controller to configure the routes (done by the `passport.authenticate()` method) + the injected send() method in the sequence must be aware of this redirection (done by the base passport module in express).

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
deepakrkris: To explain that quickly, business`Usecase` endpoints wont be configured with an`oauth2` strategy. The `OAuth2Controller` is configured with an oauth2 strategy so that it can begin the `Authorization Code Grant` dialog. Once that flow is completed, the web client would switch to other strategies like `session`  or `jwt`.

deepakrkris: The longer version :)  : 
`/thirdparty` route acts as an interface to begin the oauth2 dialog with the thirdparty and `/callback` is for the third-party login to redirect back to LoopBack app. Once `/callback` endpoint has the oauth token + the user profile, it can proceed in three different ways. 

1. It can create a browser session. This is the most popular one we use everyday to login to an app using facebook/google credentials. This means the client from there on would use the `passport-session` strategy to access other `Usecase` endpoints in the LoopBack App.

2. Or it can return the original oauth token from the third-party to the web client which can then call other `Usecase` endpoints using the token as a `Bearer`(`jwt` strategy)

3. Or it can create a new token with the same expiration time of the original thirdparty token (the `exp` field in that token says it) and send that to the web client. (`jwt` strategy)

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  function givenLoopBackApp() {
+    app = simpleRestApplication();
+  }
+
+  function givenOauth2Strategy() {
+    const passport = new Oauth2Strategy(oauth2Options, verify);
+
+    // passport-oauth2 base class leaves user profile creation to subclass implementations
+    passport.userProfile = (accessToken, done) => {
```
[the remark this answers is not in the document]
emonddr: Good question

extensions/authentication-passport/src/strategy-adapter.ts
emonddr: This is the part I found tricky (due to my lack of knowledge in authentication when I started in Jan 2019), the authentication strategy code needs to execute before it reaches a controller method. If the user is not allowed to call this controller method, then they should get a `not authenticated` http error.  So in the flow you're following in all the code updates you've made @deepakrkris , is a user able to access and get into controller method code `before` this oauth authentication flow is finished? Perhaps we need a chart explaining each step of the code path through our code and through the oauth provider backend?

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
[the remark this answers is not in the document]
deepakrkris: @jannyHou userprofile is a no-op function in 'Passport-oauth2' Strategy since it is a base class for all oauth2 implementations like 'passport-facebook', etc. we are just overriding it here for our oauth2 impl.

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,313 @@
...
+      });
+
+      it('login page redirects to authorization app callback endpoint', async () => {
+        let params = url.parse(providerLoginUrl).query;
+        params = params + '&&username=user1&&password=abc';
+        // On successful login, the authorizing app redirects to the callback url
+        // HTTP status code 302 is returned to the browser
+        const response = await supertest('')
+          .post('http://localhost:9000/login_submit?' + params)
+          .send({username: 'user', password: 'abc'})
```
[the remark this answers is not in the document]
deepakrkris: I was having a hard time to submit the form with variables in the beginning, so tried all kinds of approach. I will remove the query parameters since form submission is working now.

extensions/authentication-passport/src/__tests__/acceptance/fixtures/mock-oauth2-social-app.ts
```diff
@@ -0,0 +1,334 @@
...
+ * handles callback part of the authorization call
+ *
+ * 1. creates access code
+ * 2. generates token
+ * 3. stores token
+ * 4. redirects to callback url with access code
+ */
+app.post('/login_submit', urlencodedParser, async function (req, res) {
+  const user = findUser(
+    req.body.username || 'user1',
```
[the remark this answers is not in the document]
deepakrkris: I will remove the default values, added them when I was trying to make the login page submit work.

extensions/authentication-passport/src/strategy-adapter.ts
deepakrkris: @jannyHou @emonddr users are able to access the oauth2controller methods before authentication completes because they participate to get the user validated with an external system. Those methods are essentially for authenticating the user and are small parts of the larger oauth2 dialog.

deepakrkris: the Oauth2Controller methods that we have defined are similar to the endpoints defined in the [authenticate requests](https://github.com/jaredhanson/passport-oauth2#authenticate-requests) section of the passport oauth2 readme.

packages/authentication/src/providers/auth-action.provider.ts
emonddr: I see, ok

emonddr:
Thanks for adding more documentation; including that ASCII diagram in the comment ;)

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,333 @@
...
+//  |            |              |               | <-------------[6]---------    |  |           |
+//  |            |              |               |  LB app requests access token |  Stage 2     |
+//  |            |              |               |  with access_code             |  |           |
+//  |            |              |               |                               |  |           |
+//  |            |              |               | ---------------[7]--------->  |  |           |
+//  |            |              +---------------+       returns access token    |  |           |
+//  |            |                                                              |  v           |
+//  |            |      <------------------------[8]-----------------------     | ****         |
+//  +------------+        LB App returns to browser the access token            +--------------+
+
```
raymondfeng: Thank you for the detailed explanation. It might be better to have a separate `README.md` though.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -49,8 +50,26 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
...
+    const authResponse = await strategy.authenticate(request);
+    let userProfile: UserProfile;
+
+    // response from `strategy.authenticate()` could return an object of type UserProfile or RedirectRoute
+    if (RedirectRoute.isRedirectRoute(authResponse)) {
+      const redirectOptions = authResponse;
+      // bind redirection url and status as 'authentication.oauth2.redirectUrl' to context
+      // controller should handle actual redirection
+      this.ctx
+        .bind('authentication.redirect.url')
```
raymondfeng: Please define constants for such keys in `keys.ts`.

packages/rest/src/router/redirect-route.ts
```diff
@@ -37,11 +42,40 @@ export class RedirectRoute implements RouteEntry, ResolvedRoute {
...
 
+  /**
+   * handle redirection to a target location
+   *
+   * @param response
+   */
+  redirect(response: Response): void {
+    response.statusCode = this.statusCode || 302;
+    response.setHeader('Location', this.targetLocation);
+    response.setHeader('Content-Length', '0');
```
raymondfeng: This is probably not needed.

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
```diff
@@ -0,0 +1,333 @@
...
+    @inject('authentication.redirect.url')
+    redirectUrl: string,
+    @inject('authentication.redirect.status')
+    status: number,
+    @inject(RestBindings.Http.RESPONSE)
+    response: Response,
+  ) {
+    const redirectHandler = new RedirectRoute('/', redirectUrl, status);
+    redirectHandler.redirect(response);
+    return redirectHandler;
```
raymondfeng: Change it to `return response` please.

packages/rest/src/writer.ts
raymondfeng: We should remove this condition now as `redirect` is already handled in the controller.

extensions/authentication-passport/src/__tests__/acceptance/authentication-with-passport-strategy-oauth2-adapter.acceptance.ts
deepakrkris: @raymondfeng , if RedirectRoute is not returned from the controller, the writer.ts (send function) cannot do a type check for RedirectRoute.

deepakrkris: If we should return the Response object, in the writer.ts should we do a lookup against the request context for a redirect url value instead of the RedirectRoute type check ?

deepakrkris: Raymond I see your point, writer.ts returns immediately on a response result. Thank you. I will make the change.

extensions/authentication-passport/src/strategy-adapter.ts
jannyHou: @deepakrkris Thank you for the explanation 👍 the redirection passes the authenticated user's info to the callback endpoint, which results in the controller function maps to `/thirdparty`  won't have a user profile.

packages/authentication/src/providers/auth-action.provider.ts
jannyHou: +1

packages/authentication/src/types.ts
```diff
@@ -87,7 +87,9 @@ export interface AuthenticationStrategy {
    *
    * @param request - Express request object
    */
-  authenticate(request: Request): Promise<UserProfile | undefined>;
+  authenticate(
```
jannyHou: Had a chat with @deepakrkris yesterday and I agree it's reasonable to allow returning a redirect route. Might be a breaking change.

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: agreed

packages/rest/src/router/redirect-route.ts
deepakrkris: agreed

packages/rest/src/router/redirect-route.ts
```diff
@@ -6,11 +6,7 @@
 import {OperationObject, SchemasObject} from '@loopback/openapi-v3';
 import {ResolvedRoute, RouteEntry} from '.';
 import {RequestContext} from '../request-context';
-import {
-  OperationArgs,
-  OperationRetval,
-  PathParameterValues,
-} from '../types';
+import {OperationArgs, OperationRetval, PathParameterValues} from '../types';
```
deepakrkris: effectively no change here, just undid the change for last commit

deepakrkris:
@raymondfeng could you please review/approve, I have fixed your review comments.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -29,6 +29,7 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
     readonly getStrategy: Getter<AuthenticationStrategy>,
     @inject.setter(SecurityBindings.USER)
     readonly setCurrentUser: Setter<UserProfile>,
+    @inject.context() private ctx: Context,
```
raymondfeng: `inject.setter` is a better option. See https://loopback.io/doc/en/lb4/Decorators_inject.html for examples.

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -49,8 +50,26 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
...
+    if (RedirectRoute.isRedirectRoute(authResponse)) {
+      const redirectOptions = authResponse;
+      // bind redirection url and status as 'authentication.oauth2.redirectUrl' to context
+      // controller should handle actual redirection
+      this.ctx
+        .bind(AuthenticationBindings.AUTHENTICATION_REDIRECT_URL)
+        .to(redirectOptions.targetLocation);
+      this.ctx
+        .bind(AuthenticationBindings.AUTHENTICATION_REDIRECT_STATUS)
+        .to(redirectOptions.statusCode);
```
raymondfeng: These can be replaced with setters (https://loopback.io/doc/en/lb4/Decorators_inject.html).

packages/authentication/src/providers/auth-action.provider.ts
raymondfeng: We already use `@inject.setter` for user:

```ts
@inject.setter(SecurityBindings.USER)
    readonly setCurrentUser: Setter<UserProfile>,
```

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: agreed

packages/authentication/src/providers/auth-action.provider.ts
deepakrkris: agreed

deepakrkris:
@raymondfeng ready for review again, thank you for the perfect comments

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -29,6 +29,10 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
     readonly getStrategy: Getter<AuthenticationStrategy>,
     @inject.setter(SecurityBindings.USER)
     readonly setCurrentUser: Setter<UserProfile>,
+    @inject.setter(AuthenticationBindings.AUTHENTICATION_REDIRECT_URL)
+    readonly setRedirectUrl: Setter<string>,
+    @inject.setter(AuthenticationBindings.AUTHENTICATION_REDIRECT_STATUS)
+    readonly setRedirectStatus: Setter<number>,
```
raymondfeng: 👍

Issue #4255, opened 2019-12-04
How to define a model @property that allows any scalar value?


freakpol:
I'm want to store any type on a property, for example:

```
@model()
class MyModel extends Entity{
    @property({id: true})
    id: number;

    @property({
        type: 'I DONT KNOW WHT TO PUT HERE TO ALLOW ANY VALUE'
    })
    value: any;
}
```

The idea, is that, when I POST to that entity I can send the following examples, and all work the same

```
POST /mymodels 
{
  "id": 1,
  "value": 1
}
---
POST /mymodels 
{
  "id": 2,
  "value": "1"
}
---
POST /mymodels 
{
  "id": 3,
  "value": true
}
---
POST /mymodels 
{
  "id": 4,
  "value": null
}
---
POST /mymodels 
{
  "id": 5,
  "value": {"a":"b" }
}
```

Note: Database is a mongo db, so it shouldn't be a problem.

## Acceptance Criteria
-  [ ] Specify type `any` in `lb4 model` cli should not result in "Unsupported type" error. 
-  [ ] Add test to make sure the type can be converted to OpenAPI spec correctly

References: 
- Code that gives out the "unsupported type" error: https://github.com/strongloop/loopback-next/blob/master/packages/repository-json-schema/src/build-schema.ts#L167
- Supported types: https://loopback.io/doc/en/lb3/LoopBack-types.html.  This is a LB3 docs, I assume it would be the same for LB4 (?).

dougal83:
Try `type: any`

Ref: https://loopback.io/doc/en/lb3/LoopBack-types.html

freakpol:
Nope, if I use any, loopback says that is not a valid type for the openapi spec:

```
Cannot start the application. Error: Unsupported type: any
```

dougal83:
Perhaps it needs to be as following:

````
@property({
    type: 'any',
    required: true,
  })
  test: any;
````

This is what `lb4 model` produces.

freakpol:
Yep, thats what I've tried.  And get that error.

dougal83:
I see sorry, I had assumed it the `lb4 model` command would not give unusable results. I'd guess that an option would be to use `type string` and store as json.

freakpol:
Yeah, kinda odd.

dhmlau:
@freakpol @dougal83, thanks for reporting this issue and analyzing it.  
[...]

PR #4930, opened 2020-03-19
feat: support any type

jannyHou:
Implements https://github.com/strongloop/loopback-next/issues/4255, which supports the `any` as the type for arbitrary model property.

*This is the first PR created to support `any` type. We need another PR to append the `any` type's openapi schema in `components.schemas`*

### Description 

Define the property in model as:
  
```ts
import {AnyType} from '@loopback/repository';
@property({
    // `@loopback/repository-json-schema` recognizes it
    type: 'any'
  })
  // specify `AnyType` for it
  arbitraryProp: AnyType
```

Generated json schema would be

```ts
arbitraryProp: {
    $ref: '#/definitions/AnyType'
}
```

Generated openapi spec would be

```ts
"components": {
    "schemas": {
      "Todo": {
        "title": "Todo",
        "properties": {
          "id": {
            "type": "number"
          },
          // ...other properties
          // The schema will be generated as a reference
          // TODO: to completely support this feature, we can provide swagger AnyType schema
          // out-of-the-box
          // see link https://swagger.io/docs/specification/data-models/data-types/#any
          "arbitraryProp": {
            "$ref": "#/components/schemas/AnyType"
          }
        },
        "required": [
          "title"
        ],
        "additionalProperties": false
      },
  },
}
```

-->

## Checklist

### DIFF
--- a/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
+++ b/examples/todo/src/__tests__/acceptance/todo.acceptance.ts
@@ -65,6 +65,14 @@ describe('TodoApplication', () => {
     expect(result).to.containDeep(todo);
   });
 
+  it('creates a todo with arbitrary property', async function () {
+    const todo = givenTodo({tag: {random: 'random'}});
+    const response = await client.post('/todos').send(todo).expect(200);
+    expect(response.body).to.containDeep(todo);
+    const result = await todoRepo.findById(response.body.id);
+    expect(result).to.containDeep(todo);
+  });
+
   it('rejects requests to create a todo with no title', async () => {
     const todo = givenTodo();
     delete todo.title;
--- a/examples/todo/src/models/todo.model.ts
+++ b/examples/todo/src/models/todo.model.ts
@@ -41,6 +41,12 @@ export class Todo extends Entity {
   })
   remindAtGeo?: string; // latitude,longitude
 
+  @property({
+    type: 'any',
+  })
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  tag?: any;
+
   constructor(data?: Partial<Todo>) {
     super(data);
   }
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -179,6 +179,24 @@ describe('build-schema', () => {
         expectValidJsonSchema(jsonSchema);
       });
 
+      it('properly converts any properties', () => {
+        @model()
+        class TestModel {
+          @property({
+            type: 'any',
+          })
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          anyProp: any;
+        }
+
+        const jsonSchema = modelToJsonSchema(TestModel);
+        expect(jsonSchema.properties).to.deepEqual({
+          anyProp: {},
+        });
+
+        expectValidJsonSchema(jsonSchema);
+      });
+
       it('properly converts primitive array properties', () => {
         @model()
         class TestModel {
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -181,7 +181,8 @@ export function stringTypeToWrapper(type: string | Function): Function {
       wrapper = Array;
       break;
     }
-    case 'object': {
+    case 'object':
+    case 'any': {
       wrapper = Object;
       break;
     }
@@ -235,6 +236,8 @@ export function metaToJsonProperty(meta: PropertyDefinition): JSONSchema {
       type: 'string',
       format: 'date-time',
     });
+  } else if (propertyType === 'any') {
+    // no-op, the json schema for any type is {}
   } else if (isBuiltinType(resolvedType)) {
     Object.assign(propDef, {
       type: resolvedType.name.toLowerCase(),
--- a/packages/rest/src/__tests__/unit/ajv.service.unit.ts
+++ b/packages/rest/src/__tests__/unit/ajv.service.unit.ts
@@ -39,6 +39,93 @@ describe('Ajv service', () => {
     expect(result).to.be.true();
   });
 
+  // possible values for type any
+  const TEST_VALUES = {
+    string: 'abc',
+    number: 123,
+    object: {random: 'random'},
+    array: [1, 2, 3],
+    null: null,
+  };
+
+  context('accepts any type with schema {}', () => {
+    for (const v in TEST_VALUES) {
+      testAnyTypeWith(v);
+    }
+
+    function testAnyTypeWith(value: string) {
+      it(`with value ${value}`, async () => {
+        const ajv = new AjvProvider().value();
+        const validator = ajv.compile({});
+        const result = await validator(value);
+        expect(result).to.be.true();
+      });
+    }
+  });
+
+  context('accepts any type with schema {} - property', () => {
+    for (const v in TEST_VALUES) {
+      testAnyTypeWith(v);
+    }
+
+    function testAnyTypeWith(value: string) {
+      it(`with value ${value}`, async () => {
+        const ajv = new AjvProvider().value();
+        const validator = ajv.compile({
+          type: 'object',
+          properties: {
+            name: {type: 'string'},
+            arbitraryProp: {},
+          },
+        });
+        const result = await validator({
+          name: 'Zoe',
+          arbitraryProp: value,
+        });
+        expect(result).to.be.true();
+      });
+    }
+  });
+
+  context('accepts any type with schema true', () => {
+    for (const v in TEST_VALUES) {
+      testAnyTypeWith(v);
+    }
+
+    function testAnyTypeWith(value: string) {
+      it(`with value ${value}`, async () => {
+        const ajv = new AjvProvider().value();
+        const validator = ajv.compile(true);
+        const result = await validator(value);
+        expect(result).to.be.true();
+      });
+    }
+  });
+
+  context('accepts any type with schema true - property', () => {
+    for (const v in TEST_VALUES) {
+      testAnyTypeWith(v);
+    }
+
+    function testAnyTypeWith(value: string) {
+      it(`with value ${value}`, async () => {
+        const ajv = new AjvProvider().value();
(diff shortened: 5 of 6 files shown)


jannyHou:
---

DISCUSSION POINT (the todo from last comment)

I plan to create anther PR adding the openapi shema for `any`.
According to https://swagger.io/docs/specification/data-models/data-types/#any, the openapi schema spec for any type will be:

```ts
components: {
   schemas: {
      AnyValue: {}
   }
}
```

The reason why I plan to integrate this schema into the rest server's spec is:

When a model has a property in type of another model, we usually expect user to define that model properly so that its schema is appended in the server OpenAPI spec's `components.schemas`. But the situation for property in `AnyType` is tricky, because it's a LoopBack type we defined in [@loopback/repository](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/types/any.ts#L13).
Therefore users don't have a model for it in their project's `/models` folder, like how they have `/models/address.model.ts` defined for a `Customer` model that contains an address property.

This PR supports the any type at json schema level. And due to the above reason, we need another PR which adds schema to the rest server's spec. And also update the todo example to have an arbitrary property.

WDYT?

agnes512:
Thanks Janny for bring up the discussion.  I have two questions/concerns as I am not that familiar with this area:
1. Since LoopBack need the defined `AnyType` in the first place, what would it affect if we change it to an arbitrary type? 
2. What are we going to do with the `AnyType`?

raymondfeng:
I'm not sure if OpenAPI spec json schema supports `any` type.

raymondfeng:
Maybe `any` in json schema is `oneOf` the following:

- string | number | boolean | null 
- json array 
- json object

raymondfeng:
You can use the following schema definition for any:

```js
properties:
{
  anyProp: true, // or anyProp: {}
}
```

.vscode/launch.json
```diff
@@ -43,6 +49,15 @@
...
       "port": 5858
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Example",
+      "program": "${workspaceFolder}/examples/todo/index.js",
+      "outFiles": [
+        "${workspaceFolder}/examples/todo/**/*.js"
+      ]
```
raymondfeng: Please revert this change.

examples/todo/src/models/todo.model.ts
```diff
@@ -41,6 +41,12 @@ export class Todo extends Entity {
   })
   remindAtGeo?: string; // latitude,longitude
 
+  @property({
+    type: 'any',
+  })
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  arbitraryProp?: any;
```
raymondfeng: Let's have a more meaningful name such as `tag`?

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -179,6 +180,23 @@ describe('build-schema', () => {
         expectValidJsonSchema(jsonSchema);
       });
 
+      it('properly converts any properties', () => {
+        @model()
+        class TestModel {
+          @property({
+            type: 'any',
+          })
+          anyProp: AnyType;
```
raymondfeng: The TypeScript typing should not be `AnyType`. Use `any` or `unknown` instead.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -179,6 +180,23 @@ describe('build-schema', () => {
...
+        }
+
+        const jsonSchema = modelToJsonSchema(TestModel);
+        expect(jsonSchema.properties).to.deepEqual({
+          anyProp: {},
```
raymondfeng: Do we prefer `{}` or `true`?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -193,6 +194,10 @@ export function stringTypeToWrapper(type: string | Function): Function {
       wrapper = Buffer;
       break;
     }
+    case 'any': {
+      wrapper = AnyType;
```
raymondfeng: The wrapper is probably `Object`. `AnyType` is used internally for conversions.

packages/rest/src/__tests__/unit/ajv.service.unit.ts
```diff
@@ -39,6 +39,53 @@ describe('Ajv service', () => {
...
+    }
+
+    function testAnyTypeWith(value: string) {
+      it(`with value ${value}`, async () => {
+        const ajv = new AjvProvider().value();
+        const validator = ajv.compile({
+          type: 'object',
+          properties: {
+            name: {type: 'string'},
+            arbitraryProp: {},
```
raymondfeng: Can we also test `arbitraryProp: true`?

packages/rest/src/__tests__/unit/ajv.service.unit.ts
```diff
@@ -39,6 +39,53 @@ describe('Ajv service', () => {
     expect(result).to.be.true();
   });
 
+  // possible values for type any
+  const TEST_VALUES = {
+    string: 'abc',
+    number: 123,
+    object: {random: 'random'},
+    array: [1, 2, 3],
```
raymondfeng: Add `null`?

examples/todo/src/models/todo.model.ts
jannyHou: `tag` sounds good

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: good catch, updated to `any`

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -68,6 +68,21 @@ describe('TodoApplication', () => {
     expect(result).to.containDeep(todo);
   });
 
+  it('creates a todo with arbitrary property', async function() {
+    // Set timeout to 30 seconds as `post /todos` triggers geocode look up
+    // over the internet and it takes more than 2 seconds
+    // eslint-disable-next-line no-invalid-this
+    this.timeout(30000);
```
nabdelgadir: Since there's no geocode in this test, you don't need this part

.vscode/launch.json
```diff
@@ -24,7 +27,10 @@
       "request": "launch",
       "name": "Debug Current Test File",
       "program": "${workspaceRoot}/bin/mocha-current-file",
-      "runtimeArgs": ["-r", "${workspaceRoot}/packages/build/node_modules/source-map-support/register"],
+      "runtimeArgs": [
+        "-r",
+        "${workspaceRoot}/packages/build/node_modules/source-map-support/register"
+      ],
```
raymondfeng: I would rather revert this file as it's not related to this feature.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
raymondfeng: Not yet. I'm still seeing `AnyType`.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: @raymondfeng The reason I eventually chose `{}` is because the object can take in other properties like `description`, see [code](https://github.com/strongloop/loopback-next/blob/master/packages/repository-json-schema/src/build-schema.ts#L246-L254)
I verified `{description: 'some desc'}` is still a valid syntax for any type.
WDYT? I can switch to `true` if we want to align with the latest standard.

packages/rest/src/__tests__/unit/ajv.service.unit.ts
jannyHou: cc @raymondfeng see the comment in https://github.com/strongloop/loopback-next/pull/4930#discussion_r398970633

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
raymondfeng: As long as we test `true`, I'm good either way.

raymondfeng: See https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.3.2

raymondfeng:
Pls fix code lint error - https://travis-ci.com/github/strongloop/loopback-next/jobs/305196649

jannyHou:
@agnes512 Thank you for reviewing and good questions:

> Since LoopBack need the defined AnyType in the first place, what would it affect if we change it to an arbitrary type?

`AnyType` is only for internal use(didn't realize that in the original implementation), it's replaced by type `any` now. For "what would it affect if we change it to an arbitrary type?" I am not sure what do you mean by "arbitrary type" here...since `any` already allows the arbitrary type :p

e.g. given a property `tag` that allows any value:

```ts
@property({type: 'any'})
tag?: any
```
Which part do you want to change and see how it affects?

> What are we going to do with the AnyType?

Since `AnyType` is not used anymore, I assume your question is about `any` now. 

- `any` is the type of your model property, like the `tag` property above
- the {type: 'any'} as the meta data of a model property, is consumed by the `@loopback/repository-json-schema` module, to generate the correct json schema (`{}` or `true`, still discuss with @raymondfeng ) for the `tag` field.
- eventually a property with type `any` will have `{}` (or `true`) as its openapi schema

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: sounds good, I added another 2 set for `true`.

examples/todo/src/__tests__/acceptance/todo.acceptance.ts
```diff
@@ -65,6 +65,14 @@ describe('TodoApplication', () => {
     expect(result).to.containDeep(todo);
   });
 
+  it('creates a todo with arbitrary property', async function () {
```
raymondfeng: Nitpick: Let's use `async () =>` to be consistent.

Issue #4398, opened 2020-01-09
Is it time for a Long Term Support (LTS) version?


bajtos:
We are seeing growing adoption of LoopBack 4, with many projects already deploying to production. It makes me wonder what kind of stability guarantees are our users expecting from the framework?

At the moment, there is a single release line "Current" where we make all changes (features small and large, bug fixes, security fixes, etc.). Upgrading a LB4-based project to a newer version may require some work on the user side, e.g. when a new TypeScript version is released. This could become a problem when a security vulnerability is fixed and the upgrade from a vulnerable to the fixed version is not trivial.

Is it perhaps time to introduce an LTS release line, where only bug fixes and security patches will be landed? The current LTS policy used for LoopBack 3.x can be found [here](https://loopback.io/doc/en/contrib/Long-term-support.html):

> **Active LTS**
> 
> A major LoopBack version (for example, 3.x) enters Active LTS when the next major version is released (for example, 4.0) and stays in Active LTS mode for at least six months.
> 
> Once a release enters LTS, no new features may be added to that release. Changes are limited to:
> 
> - Bug fixes;
> - Security updates;
> - Relevant documentation updates;
> - Certain performance improvements where the risk of breaking existing applications is minimal;
> - Changes that introduce large amount of code churn where the risk of breaking existing applications is low and where the change in question may significantly ease the ability to backport future changes due to the reduction in diff noise. Semver-minor changes are only permitted if required for bug fixes. Semver-major changes are only permitted if required for critical security and bug fixes.
> 
> Support for new major Node.js versions may be added if the required changes have a low risk of breaking existing applications.
>
>
> **Maintenance LTS**
>
> When a new major version (for example, 4.0) is released, the oldest Active LTS version (for example, 2.x) enters Maintenance LTS mode, where it will stay for as long as the Node.js LTS versions available at release time are maintained by the Node.js project.
> 
> Once a release moves into Maintenance LTS mode, only critical bugs, critical security fixes, and documentation updates will be permitted.
> 
> Specifically, adding support for new major Node.js versions is not permitted.

I'd like to use this issue to kick-off the initial discussion, where we can learn what our users actually need from the framework. Pinging @strongloop/loopback-maintainers @strongloop/loopback-next @raymondfeng @dhmlau

ibmjm:
[...]

PR #4937, opened 2020-03-20
fix: update peerDependencies

### LINKED ISSUE #4889: (model-api-builder) Peer dependencies out of date?
Should these dependencies be `^.2.0.0`?

https://github.com/strongloop/loopback-next/blob/4a25fa25ce33129f52b0e6eb196d2100c55ccb6e/packages/model-api-builder/package.json#L22-L25

derdeka:
Manually fixes peerDependency, maybe there will be a script to fix this automated in future.

Fixes #4889

## Checklist

### DIFF
--- a/packages/booter-lb3app/package.json
+++ b/packages/booter-lb3app/package.json
@@ -25,9 +25,9 @@
     "tslib": "^1.11.1"
   },
   "peerDependencies": {
-    "@loopback/boot": "^1.2.2",
-    "@loopback/core": "^1.5.0",
-    "@loopback/rest": "^1.10.3"
+    "@loopback/boot": "^1.2.2 || ^2.0.0",
+    "@loopback/core": "^1.5.0 || ^2.0.0",
+    "@loopback/rest": "^1.10.3 || ^2.0.0  || ^3.0.0"
   },
   "devDependencies": {
     "@loopback/boot": "^2.0.1",
--- a/packages/model-api-builder/package.json
+++ b/packages/model-api-builder/package.json
@@ -20,8 +20,8 @@
   "copyright.owner": "IBM Corp.",
   "license": "MIT",
   "peerDependencies": {
-    "@loopback/core": "^1.12.0",
-    "@loopback/repository": "^1.16.0"
+    "@loopback/core": "^1.12.0 || ^2.0.0",
+    "@loopback/repository": "^1.16.0 || ^2.0.0"
   },
   "devDependencies": {
     "@loopback/build": "^1.7.1",
--- a/packages/repository-tests/package.json
+++ b/packages/repository-tests/package.json
@@ -35,7 +35,7 @@
     "tslib": "^1.11.1"
   },
   "peerDependencies": {
-    "@loopback/repository": "^1.6.1"
+    "@loopback/repository": "^1.6.1 || ^2.0.0"
   },
   "files": [
     "README.md",
--- a/packages/rest-crud/package.json
+++ b/packages/rest-crud/package.json
@@ -34,9 +34,9 @@
     "@types/node": "^10.17.17"
   },
   "peerDependencies": {
-    "@loopback/core": "^1.12.4",
-    "@loopback/repository": "^1.19.1",
-    "@loopback/rest": "^2.0.0"
+    "@loopback/core": "^1.12.4 || ^2.0.0",
+    "@loopback/repository": "^1.19.1 || ^2.0.0",
+    "@loopback/rest": "^2.0.0 || ^3.0.0"
   },
   "files": [
     "README.md",


packages/booter-lb3app/package.json
```diff
@@ -25,9 +25,9 @@
     "tslib": "^1.11.1"
   },
   "peerDependencies": {
-    "@loopback/boot": "^1.2.2",
-    "@loopback/core": "^1.5.0",
-    "@loopback/rest": "^1.10.3"
+    "@loopback/boot": "^1.2.2 || ^2.0.0",
+    "@loopback/core": "^1.5.0 || ^2.0.0",
+    "@loopback/rest": "^1.10.3 || ^2.0.0  || ^3.0.0"
```
raymondfeng: I'm also not very confident with the wide range as the CI only tests the module against latest versions of such dependencies.

The most reliable collection of `@loopback/*` modules that are guaranteed to work together are these being released together :-).

packages/booter-lb3app/package.json
```diff
@@ -25,9 +25,9 @@
     "tslib": "^1.11.1"
   },
   "peerDependencies": {
-    "@loopback/boot": "^1.2.2",
-    "@loopback/core": "^1.5.0",
-    "@loopback/rest": "^1.10.3"
+    "@loopback/boot": "^1.2.2 || ^2.0.0",
```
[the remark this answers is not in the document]
derdeka: @bajtos WDYT?

packages/booter-lb3app/package.json
achrinza: @raymondfeng I think users will expect peer deps that meets the folowing criteria would work:

- Same major semver
- Equal or higher that min. version within semver (e.g. `@loopback/boot: ^1.2.2`)

If there's incompatibilities between 2 versions that meet the aforementioned criteria then we are breaking semver conventions.

So it may be a good idea to keep this range instead of pinning it to an exact version.

raymondfeng: @achrinza I meant to suggest we use `"@loopback/rest": "^3.0.0"`. It's hard to ensure that this module is compatible with 1.x, 2.x, and 3.x without meaningful CI test coverage.

raymondfeng:
@derdeka See https://github.com/strongloop/loopback-next/pull/4972

packages/booter-lb3app/package.json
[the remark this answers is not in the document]
bajtos: Let's discuss.

I introduced peer dependencies as a solution for two problems:

(1)
As an extension (e.g. `@loopback/booter-lb3app`), I want to use the core packages provided by the target application (LoopBack project). I do not want to bundle my own copy of core packages. 

Why is this important? It gives the target application full control of which version of core packages are used. I am not talking about version numbers - I want to allow users to use their own fork of a core package if they have a bug fix that is not a part of an official release (yet).

(2)
As an extension, I want to specify which versions of core packages I am compatible with. This way LoopBack users can pick the right versions and if they don't, then `npm install` will warn them about a possible problem.

I am ok to use a different solution than peer dependencies as long as it preserves the properties of the current setup.

I see two discussion points:

1. Do we want to continue using `peerDependencies` or not?
2. If yes, then what version ranges to use in peer dependencies? (Support only a single semver-major version or support as many versions as feasible?)

Let's keep this thread for discussing the first point and discuss the second point in https://github.com/strongloop/loopback-next/pull/4937/files#r395738578 below.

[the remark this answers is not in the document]
bajtos: > peerDependencies as it's really half-baked and not bullet-proof with npm.

From what I remember, peer dependencies used to be considered as deprecated in the past, but in the last few years the situation improved and they are no longer frowned upon. Popular tools like `eslint` and `babel` are using peer dependencies for their plugins.

```
$ npm info eslint-config-airbnb peerDependencies
{
  eslint: '^5.16.0 || ^6.8.0',
  'eslint-plugin-import': '^2.20.1',
  'eslint-plugin-jsx-a11y': '^6.2.3',
  'eslint-plugin-react': '^7.19.0',
  'eslint-plugin-react-hooks': '^2.5.0 || ^1.7.0'
}

$ npm info @babel/preset-env peerDependencies
{
  '@babel/core': '^7.0.0-0' 
}
```

packages/booter-lb3app/package.json
bajtos: Take our connectors as an example. They support both LB3 and LB4, therefore we have only one version to maintain for both LB3 and LB4 users.

I think it makes a lot of sense to adopt the same approach for our extensions once we start maintaining LB4 LTS versions (see https://github.com/strongloop/loopback-next/issues/4398). This may include writing acceptance tests to verify that the current (`master`) version of an extension works with older (LTS) versions of a core component, as @raymondfeng has mentioned in his comment.

However, we don't provide LTS for LB4 yet, therefore I think it's ok to use a narrow range for the peer dependencies for now. Especially considering that we don't have CI test coverage to verify compatibility with older versions. I don't think it would be a good use of our time to work on such CI setup now.

If we decide to limit the supported peer versions to the latest major version only, then can we please find a way how to automate the update process - ideally as part of our version release setup? I.e. when we publish a new version, then the peer dependencies should be updated to match the new versions created by the publishing.

emonddr:
I assume we're now switching to the #4972 solution?

raymondfeng:
Close as #4972 is landed. We'll fix peer deps in next release (tentatively next week).

PR #4949, opened 2020-03-23
Repo class factory improvements

bajtos:
A proposal how to improve several aspects of https://github.com/strongloop/loopback-next/pull/4792.

@raymondfeng if you accept my suggestions, then please preserve my co-authorship in the git commit message (see https://github.blog/2018-01-29-commit-together-with-co-authors/)


emonddr:
@bajtos @raymondfeng ,
If this is applied, we may need to revisit the LB4 mixin class factory functions in : https://loopback.io/doc/en/lb4/migration-models-mixins.html .

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -29,10 +28,10 @@ import {
 describe('RepositoryClass builder', () => {
   describe('defineCrudRepositoryClass', () => {
     it('should generate CRUD repository class', async () => {
-      const AddressRepository = defineCrudRepositoryClass(
-        Address,
-        DummyCrudRepository,
-      );
+      const AddressRepository = defineCrudRepositoryClass<
+        typeof Address,
+        DummyCrudRepository<Address>
```
raymondfeng: So TypeScript cannot infer the generic types from the parameters in this case?

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
```diff
@@ -92,6 +92,26 @@ describe('RepositoryClass builder', () => {
...
+        M
+      > {
+        async getByName(name: string): Promise<Product[]> {
+          throw new Error('not implemented');
+        }
+      }
+
+      const ProductRepository = defineKeyValueRepositoryClass<
+        typeof Product,
+        MyKeyValueRepo<Product>
```
raymondfeng: Same question here.

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
raymondfeng: To make inference happy, I have to do the following:

```ts
function castDummyRepository<M extends Model>() {
  return class extends DummyCrudRepository<M> {};
}

const AddressRepository = defineCrudRepositoryClass(
  Address,
  castDummyRepository<Address>(),
);
```

raymondfeng: TypeScript does not allow `typeof DummyCrudRepository<M>`.

bajtos:
> If this is applied, we may need to revisit the LB4 mixin class factory functions in : https://loopback.io/doc/en/lb4/migration-models-mixins.html .

@emonddr I am little bit confused, how is this new code related to `define*Repository` functions related to the documentation on mixin migration? AFAIK, the migration guide is not affected by these changes at all. What am I missing?

packages/repository/src/__tests__/unit/repositories/define-repository-class.unit.ts
bajtos: > So TypeScript cannot infer the generic types from the parameters in this case?

Unfortunately not. I think we should report a bug to TypeScript (if it wasn't already reported).

> To make inference happy, I have to do the following (...)

IMO, `castDummyRepository` does not solve the problem of people forgetting to specify the repository type and relying on the compiler to infer it. Once they realize some extra effort is needed, then I think it's best to provide generic type parameters and there is no need to introduce `cast*Repository` functions.

packages/repository/src/define-repository-class.ts
```diff
@@ -3,28 +3,28 @@
...
+ * Signature for a Repository class bound to a given model. The constructor
+ * accepts only the dataSource to use for persistence.
+ *
+ * `define*` functions return a class implementing this interface.
  *
  * @typeParam M - Model class
  * @typeParam R - Repository class/interface
  */
-export interface CrudRepositoryClass<
+export interface NamedRepositoryClass<
```
bajtos: There is no need to have specialized types for KeyValue and CRUD repositories, we can use a generic Repository type as the starting point.

I am not comfortable using `RepositoryClass` name, because it's too ambiguous and possibly confusing, since it uses different constructor argument than `PersistedModelClass` and friends.

I was looking for a name that would describe the fact that the repository class is already bound to a given model. `ModelRepositoryClass` did not feel right either, so I settled on `NamedRepositoryClass`. Feel free to propose a better name.

packages/repository/src/define-repository-class.ts
```diff
@@ -34,62 +34,73 @@ export interface CrudRepositoryClass<
...
+  R extends Repository<PrototypeOf<M>>
+> {
   /**
    * The constructor for the generated repository class
    * @param modelClass - Model class
    * @param dataSource - DataSource object
    */
-  new (
-    // Make model class conditional of Model or Entity
-    modelClass: (M extends Entity ? typeof Entity : typeof Model) & {
```
bajtos: While this works for Entities, it won't work for other subclasses that people may use for their base repository classes.

Consider the following:

```ts
class MyAuditedEntityRepository<E extends AuditedEntity, /*...*/> extends DefaultCrudRepository<E, /*...*/> {
  constructor(modelClass: typeof AuditedEntity, dataSource: juggler.dataSource) {
    // ...
  }
  // ...
}
```

To support such use case, I think it's best to use `M extends typeof Model` as the generic type parameter.

packages/repository/src/define-repository-class.ts
```diff
@@ -34,62 +34,73 @@ export interface CrudRepositoryClass<
...
- * @typeParam M - Model class
- * @typeParam R - CRUD Repository class/interface
+ * @typeParam M - Model class constructor (e.g. `typeof Address`)
+ * @typeParam R - Repository class (e.g. `DefaultCrudRepository<Address, number>`)
  */
-export function defineCrudRepositoryClass<
-  M extends Model,
-  R extends CrudRepository<M>
+export function defineRepositoryClass<
+  M extends typeof Model,
```
bajtos: Similarly here, we need `M` to represent the class constructor, not the class instance.

packages/repository/src/define-repository-class.ts
```diff
@@ -125,75 +141,30 @@ export function defineCrudRepositoryClass<
 export function defineEntityCrudRepositoryClass<
```
bajtos: I'd like to discuss the name of this function.

The actual base repository class is called `DefaultCrudRepository` (there is no `Entity` in the name). From what I remember, it was an intentional decision - we don't provide non-entity Crud repository implementation, the only Crud repository we have requires an Entity.

I think we should rename the class factory function to `defineCrudRepositoryClass` for consistence.

Note that `@loopback/rest-crud` is already using the name `defineCrudRepositoryClass`.

Thoughts?

packages/repository/src/define-repository-class.ts
raymondfeng: If it's `NamedRepositoryClass`, should we add `name: string` explicitly?

What about `GeneratedRepositoryClass`?

packages/repository/src/define-repository-class.ts
raymondfeng: That's fine with me.

raymondfeng:
@bajtos I have rebased this branch against latest master.

packages/repository/src/define-repository-class.ts
bajtos: Maybe we should take another angle. My intention is to distinguish between a generic repository class that works for any Model/Entity and a repository class that's already specialized for (bound to) a given Model/Entity.

In that light, perhaps `BoundRepositoryClass` or `SpecializedRepositoryClass` would be better names?

bajtos:
Merged manually.

Issue #4938, opened 2020-03-19
Partial patch not working with required true of nested objects


bluewayll:
Hello guys,

Dependencies version i'm using :

`"@loopback/boot": "^2.0.0",
    "@loopback/context": "^3.0.0",
    "@loopback/core": "^2.0.0",
    "@loopback/openapi-v3": "^3.0.0",
    "@loopback/repository": "^2.0.0",
    "@loopback/rest": "^3.0.0",
    "@loopback/rest-explorer": "^2.0.0",
    "@loopback/service-proxy": "^2.0.0",`

What i want to perform is a partial patch by id, my code : 

`@patch('/products/{id}', {
    responses: {
      '204': {
        description: 'Product PATCH success',
      },
    },
  })
  async updateById(
    @param.path.string('id') id: string,
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(Product, {partial: true}),
        },
      },
    })
    product: Partial<Product>,
  ): Promise<void> {
    await this.productRepository.updateById(id, product);
  }`

My models is : 
> Image => link (required: true), description (required; false)
> Product => id (id: true, generated: true), name (required: true), image (Model Image)(required: true)

What i get from the patch is : 

`{
  "error": {
    "statusCode": 422,
    "name": "UnprocessableEntityError",
    "message": "The request body is invalid. See error object `details` property for more info.",
    "code": "VALIDATION_FAILED",
    "details": [
      {
        "path": ".image",
        "code": "required",
        "message": "should have required property 'link'",
        "info": {
          "missingProperty": "link"
        }
      }
    ]
  }
}`

Thank you.

# Acceptance Criteria
- [] PATCH /Products/1 should work without error when only only the `image` property  object is specified, and only `description` property is specified for the image object. The `link` property shouldn't be required on an update, only on a create.

emonddr:
@mastermunj  , can you take a look at this issue? It is related to your commit above. Thank you. :)

emonddr:
`name` is required on Product model, and in a PATCH /products/1 , if we don't provide `name` property, the ajv doesn't mind. However, if in a PATCH /products/1 , we don't provide any product properties except `image`, and its contents does not include `link`, then ajv complains that this field is required. Seems inconsistent. How is the ajv validator informed that we are updating versus creating a Product model?

bluewayll:
Hello @emonddr ,

Just for information if i put :

`schema: {type: 'object'},`

in place of : 

[...]

PR #4968, opened 2020-03-25
fix(repository-json-schema): honor options for referenced types

raymondfeng:
Fixes https://github.com/strongloop/loopback-next/issues/4938

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -38,8 +38,40 @@ export interface JsonSchemaOptions<T extends object> {
    * Set this flag to mark all model properties as optional. This is typically
    * used to describe request body of PATCH endpoints. This option will be
    * overridden by the "optional" option if it is set and non-empty.
+   *
+   * The flag also applies to nested model instances if its value is set to
+   * 'deep', such as:
+   *
+   * @example
+   * ```ts
+   * @model()
+   * class Address {
+   *  @property()
+   *  street: string;
+   *  @property()
+   *  city: string;
+   *  @property()
+   *  state: string;
+   *  @property()
+   *  zipCode: string;
+   * }
+   *
+   * @model()
+   * class Customer {
+   *   @property()
+   *   address: Address;
+   * }
+   *
+   * // The following schema allows properties of `customer` optional, but not
+   * // `customer.address`
+   * const schemaRef1 = getModelSchemaRef(Customer, {partial: true});
+   *
+   * // The following schema allows properties of `customer` and
+   * // `customer.address` optional
+   * const schemaRef2 = getModelSchemaRef(Customer, {partial: 'deep'});
+   * ```
    */
-  partial?: boolean;
+  partial?: boolean | 'deep';
 
   /**
    * List of model properties to exclude from the schema.
@@ -78,7 +110,7 @@ export function buildModelCacheKey<T extends object>(
 /**
  * Gets the JSON Schema of a TypeScript model/class by seeing if one exists
  * in a cache. If not, one is generated and then cached.
- * @param ctor - Contructor of class to get JSON Schema from
+ * @param ctor - Constructor of class to get JSON Schema from
  */
 export function getJsonSchema<T extends object>(
   ctor: Function & {prototype: T},
@@ -437,9 +469,28 @@ export function modelToJsonSchema<T extends object>(
       continue;
     }
 
-    const propSchema = getJsonSchema(referenceType, options);
+    const propOptions = {...options};
+    if (propOptions.partial !== 'deep') {
+      // Do not cascade `partial` to nested properties
+      delete propOptions.partial;
+    }
 
-    includeReferencedSchema(referenceType.name, propSchema);
+    const propSchema = getJsonSchema(referenceType, propOptions);
+
+    // JSONSchema6Definition allows both boolean and JSONSchema6 types
+    if (typeof result.properties[p] !== 'boolean') {
+      const prop = result.properties[p] as JSONSchema;
+      const propTitle = propSchema.title ?? referenceType.name;
+      const targetRef = {$ref: `#/definitions/${propTitle}`};
+
+      if (prop.type === 'array' && prop.items) {
+        // Update $ref for array type
+        prop.items = targetRef;
+      } else {
+        result.properties[p] = targetRef;
+      }
+      includeReferencedSchema(propTitle, propSchema);
+    }
   }
 
   result.additionalProperties = meta.settings.strict === false;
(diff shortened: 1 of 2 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -436,7 +436,19 @@ export function modelToJsonSchema<T extends object>(
 
     const propSchema = getJsonSchema(referenceType, options);
 
-    includeReferencedSchema(referenceType.name, propSchema);
+    if (typeof result.properties[p] !== 'boolean') {
```
emonddr: why are we testing if it's not equal to `boolean` specifically?

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -436,7 +436,19 @@ export function modelToJsonSchema<T extends object>(
...
+      const prop = result.properties[p] as JSONSchema;
+      const propTitle = propSchema.title ?? referenceType.name;
+      const targetRef = {$ref: `#/definitions/${propTitle}`};
```
emonddr: I've observed that sometime people put values in title that don't align with a type name:

```
 async create(
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(NewUserRequest, {
            title: 'NewUser',
            exclude: ['id'],
          }),
        },
      },
    })
```
so could something errorneous happen here using `propTitle`?

packages/repository-json-schema/src/build-schema.ts
raymondfeng: JSON schema draft 6 allows a boolean for the property schema. This test skips that case so that we can safely cast the schema to `JSONSchema`.

packages/repository-json-schema/src/build-schema.ts
raymondfeng: That's fine as long as the title is unique. Please note we add `Partial` suffix to schemas built with `{partial: true}`.

packages/repository-json-schema/src/build-schema.ts
emonddr: @raymondfeng , I see. Please add a comment indicating the reason to help maintainers understand that particular check.

raymondfeng:
I marked it as a breaking change too.

emonddr:
Hmm noticed an issue.

If I have this product in memory database

```
{
    "id": "3",
    "enabled": true,
    "name": "prod1 a",
    "image": {
      "link": "link prod1 a",
      "description": "dom prod1 a"
    }
  }
```

and I perform a PATCH /products/3 with 

```
{
   
    "image": {
          "description": "dom prod1 a"
    }
  }
```

it now succeeds.

But if I do : GET /products  ,  product with id 3 is missing the link field.

```
[
 
  {
    "id": "3",
    "enabled": true,
    "name": "prod1 a",
    "image": {
      "description": "dom prod1 a"
    }
  }
]
```

emonddr:
Also the app contains these models,

```
@model({settings: {}})
export class Image extends Model {

  @property({
    type: 'string',
    required: true,
  })
  link: string;

  @property({
    type: 'string',
    required: false,
  })
  description: string;

  constructor(data?: Partial<Image>) {
    super(data);
  }
}
```
and 

```
  @property({
    type: 'string',
    id: true,
    generated: true
  })
  id: string;

  @property({
    type: 'boolean',
    default: true,
    required: false,
  })
  enabled: boolean;

  @property({
    type: 'string',
    required: true,
  })
  name: string;



  
  @property({
    type: Image,
    required: true,
  })
  image: Image;

  constructor(data?: Partial<Product>) {
    super(data);
  }
}
```

where property of type Image is defined differently than in the mocha test 

```
 class TestModel {
            @property(CustomType)
            cusType: CustomType;
          }
```

Is are they essentially the same, just using different syntax in @property decorator?

raymondfeng:
It depends on how `patch` is implemented (deep vs shallow):
- case 1: replace `image` with the supplied partial value (now the product loses data integrity as required image.link is missing)
- case 2: patch `image` with the supplied partial value (only replaces `description`)

emonddr:
I confirmed the observation https://github.com/strongloop/loopback-next/pull/4968#issuecomment-604038891 occurs without this PR.

emonddr:
Regarding

```
It depends on how patch is implemented (deep vs shallow):

case 1: replace image with the supplied partial value (now the product loses data integrity as required image.link is missing)
case 2: patch image with the supplied partial value (only replaces description)
```
Is case 1) a put ? and case 2) a patch?

I believe I tried 2) only patch the description of image, and the link disappears from image after I go a GET /products.

raymondfeng:
> Is case 1) a put ? and case 2) a patch?

Both case 1 and 2 are `patch` but 1 is shallow while 2 is deep.

> I believe I tried 2) only patch the description of image, and the link disappears from image after I do a GET /products.

The default repository implementation does not perform deep patch.

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -436,7 +463,20 @@ export function modelToJsonSchema<T extends object>(
 
     const propSchema = getJsonSchema(referenceType, options);
 
-    includeReferencedSchema(referenceType.name, propSchema);
+    // JSONSchema6Definition allows both boolean and JSONSchema6 types
+    if (typeof result.properties[p] !== 'boolean') {
+      const prop = result.properties[p] as JSONSchema;
+      const propTitle = propSchema.title ?? referenceType.name;
+      const targetRef = {$ref: `#/definitions/${propTitle}`};
```
jannyHou: @raymondfeng We have a function to generate the title suffix in [here](https://github.com/strongloop/loopback-next/blob/6a41d752f7f7dbc9f85ad8d74eb38db71e65e7c7/packages/repository-json-schema/src/build-schema.ts#L331), and it handles the `partial` option, why do we check again here?

raymondfeng: The purpose is to make sure `ParentPartial` uses `ChildPartial` instead of `Child`.

emonddr:
> I marked it as a breaking change too.

What makes it a breaking change exactly?

packages/repository-json-schema/src/build-schema.ts
emonddr: @jannyHou , if you're ok with @raymondfeng 's answer, please approve. thx.

jannyHou: > The purpose is to make sure ParentPartial uses ChildPartial instead of Child.

Fair enough 👍

raymondfeng:
> What makes it a breaking change exactly?

It used to be `Partial` but now it behaves as `DeepPartial`.

raymondfeng:
Instead of changing the default behavior of `{partial: true}`, I add a new option `{partial: 'deep'}` to allow deep-partial use cases. This make the PR backward compatible.

Please note that our repository implementation probably support shallow patches for `PATCH`.

PR #4874, opened 2020-03-12
feat(docs): add validation docs

dhmlau:
Add more documentation on how to add validation to a LoopBack application. 

Propose to add the validation docs under "Usage scenarios"
Screen Shot 2020-03-13 at 11 25 40 AM

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #4969, opened 2020-03-25
feat(example-validation-app): customize error

dhmlau:
Add code for customizing error to `validation-app` example.

Will add docs after https://github.com/strongloop/loopback-next/pull/4874 lands.

## Checklist

### DIFF
--- a/docs/site/Validation-REST-layer.md
+++ b/docs/site/Validation-REST-layer.md
@@ -175,3 +175,84 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
 ```
 
 {% include tip.html content="RegExp can be converted into a string with `.source` to avoid escaping backslashes" %}
+
+## Customize validation errors
+
+Since the error is being caught at the REST layer, the simplest way to customize
+the errors is to customize the
+[sequnce](https://loopback.io/doc/en/lb4/Sequence.html). It exists in all
+LoopBack applications scaffolded by using the `lb4` command and can be found in
+`src/sequence.ts`.
+
+Let's take a closer look at how to customize the error. A few things to note in
+the below code snippet:
+
+1. inject RestBindings.SequenceActions.LOG_ERROR for logging error and
+   RestBindings.ERROR_WRITER_OPTIONS for options
+2. customize error for particular endpoints
+3. create a new error with customized properties
+4. log the error using RestBindings.SequenceActions.LOG_ERROR
+
+{% include code-caption.html content="/src/sequence.ts" %}
+
+```ts
+export class MySequence implements SequenceHandler {
+  // 1. inject RestBindings.SequenceActions.LOG_ERROR for logging error
+  // and RestBindings.ERROR_WRITER_OPTIONS for options
+  constructor(
+    /*..*/
+    @inject(RestBindings.SequenceActions.LOG_ERROR)
+    protected logError: LogError,
+    @inject(RestBindings.ERROR_WRITER_OPTIONS, {optional: true})
+    protected errorWriterOptions?: ErrorWriterOptions,
+  ) {}
+
+  async handle(context: RequestContext) {
+    try {
+      // ...
+    } catch (err) {
+      this.handleError(context, err as HttpErrors.HttpError);
+    }
+  }
+
+  /**
+   * Handle errors
+   * If the request url is `/coffee-shops`, customize the error message.
+   */
+  handleError(context: RequestContext, err: HttpErrors.HttpError) {
+    // 2. customize error for particular endpoint
+    if (context.request.url === '/coffee-shops') {
+      // if this is a validation error
+      if (err.statusCode === 422) {
+        const customizedMessage = 'My customized validation error message';
+
+        let customizedProps = {};
+        if (this.errorWriterOptions?.debug) {
+          customizedProps = {stack: err.stack};
+        }
+
+        // 3. Create a new error with customized properties
+        // you can change the status code here too
+        const errorData = {
+          statusCode: 422,
+          message: customizedMessage,
+          resolution: 'Contact your admin for troubleshooting.',
+          code: 'VALIDATION_FAILED',
+          ...customizedProps,
+        };
+
+        context.response.status(422).send(errorData);
+
+        // 4. log the error using RestBindings.SequenceActions.LOG_ERROR
+        this.logError(err, err.statusCode, context.request);
+
+        // The error was handled
+        return;
+      }
+    }
+
+    // Otherwise fall back to the default error handler
+    this.reject(context, err);
+  }
+}
+```
--- a/examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
+++ b/examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Client} from '@loopback/testlab';
+import {Client, expect} from '@loopback/testlab';
 import {ValidationApplication} from '../..';
 import {setupApplication} from './test-helper';
 
@@ -31,30 +31,49 @@ describe('validate properties', () => {
     await client.post('/coffee-shops').send(validCoffeeShop).expect(200);
   });
 
-  it('create CoffeeShop failed with city name exceeds length limit', async () => {
+  it('fails when creating a CoffeeShop with the city name exceeds length limit', async () => {
     const invalidCityNameCS = validCoffeeShop;
     invalidCityNameCS.city = 'Tooooooooooronto';
     await client.post('/coffee-shops').send(invalidCityNameCS).expect(422);
   });
 
-  it('create CoffeeShop failed with invalid phone number', async () => {
+  it('fails when creating a CoffeeShop with an invalid phone number', async () => {
     const invalidPhoneNumCS = validCoffeeShop;
     invalidPhoneNumCS.phoneNum = '1111111111';
     await client.post('/coffee-shops').send(invalidPhoneNumCS).expect(422);
   });
 
-  it('create CoffeeShop failed with capacity exceeds limit', async () => {
+  it('fails when creating a CoffeeShop with the capacity exceeds limit', async () => {
     const invalidCapacityCS = validCoffeeShop;
     invalidCapacityCS.capacity = 10000;
     await client.post('/coffee-shops').send(invalidCapacityCS).expect(422);
   });
 
-  it('create CoffeeShop failed with invalid area code', async () => {
+  it('fails when creating a CoffeeShop with an invalid area code', async () => {
     const invalidAreaCodeCS = {
       city: 'Toronto',
       phoneNum: '999-111-1111',
       capacity: 10,
     };
-    await client.post('/coffee-shops').send(invalidAreaCodeCS).expect(422);
+    const response = await client.post('/coffee-shops').send(invalidAreaCodeCS);
+    expect(response.status).to.eql(400);
+    expect(response.body.error.message).to.eql(
+      'Area code and city do not match',
+    );
+  });
+
+  it('fails when creating a CoffeeShop with an invalid type', async () => {
+    const invalidCapacityTypeCS = {
+      city: 'Toronto',
+      phoneNum: '416-111-1111',
+      capacity: '10',
+    };
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidCapacityTypeCS)
+      .expect(422);
+
+    // with customized response
+    expect(response.body).to.have.property('resolution');
   });
 });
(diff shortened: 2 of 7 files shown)


raymondfeng:
CI fails due to code formatting issues. Run `npm run prettier:fix` to fix it.

raymondfeng:
@dhmlau Please fix https://travis-ci.com/github/strongloop/loopback-next/jobs/303320121 as follows:

```sh
cd examples/validation-app
rm -rf node_modules package-lock.json
cd ../..
npm i
```

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,48 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+          code: 'VALIDATION_FAILED',
+          resolution: 'Contact your admin for troubleshooting.',
+        };
+
+        // you can change the status code here too
+        response.status(422).send(JSON.stringify(newError));
+      }
+    }
+
+    response.status(httpError.statusCode).end(JSON.stringify(httpError));
```
bajtos: This isn't the best way how to present an error in the response. I believe this is going to drop the stack trace, which will make debugging difficult? At the same time, we don't want to print stack traces in production. All of this is handled by our default reject action. Can you please update this example to show how the custom error handler can fall back to the default implementation?

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,48 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+Since the error is being caught at the REST layer, the best way to customize the
+errors is to
+[customize the sequence of actions](https://loopback.io/doc/en/lb4/Sequence.html#customizing-sequence-actions)
+using a
+[Provider](https://loopback.io/doc/en/lb4/apidocs.context.provider.html).
+
+To do so, create a Provider class in `src/providers` folder that implements
+`Provider<Reject>`. Using the
+[validation-app](https://github.com/strongloop/loopback-next/tree/master/examples/validation-app)
+as an example, here is how the Provider class look like:
```
emonddr: > here is how the Provider class look like

here is how the Provider class looks like

docs/site/Validation-REST-layer.md
emonddr: @dhmlau , in our shopping example sequence, we catch an error like this : https://github.com/strongloop/loopback4-example-shopping/blob/master/packages/shopping/src/sequence.ts#L72  inside the sequence, and pass them to default reject action .

dhmlau: @emonddr, thanks for pointing to the code snippet!

dhmlau:
@emonddr @bajtos, for falling back to the default reject action, I used `RejectProvider`. Please see my latest commit https://github.com/strongloop/loopback-next/pull/4969/commits/76594bdb03fd5f49013c71d6d962dec432fb0250. Thanks.

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,47 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
 
 {% include tip.html content="RegExp can be converted into a string with `.source` to avoid escaping backslashes" %}
+
+## Customize validation errors
+
+Since the error is being caught at the REST layer, the best way to customize the
+errors is to
+[customize the sequence of actions](https://loopback.io/doc/en/lb4/Sequence.html#customizing-sequence-actions)
+using a
+[Provider](https://loopback.io/doc/en/lb4/apidocs.context.provider.html).
```
nabdelgadir, line 185: Would it make sense to link to https://loopback.io/doc/en/lb4/Creating-components.html#providers instead?

docs/site/Validation-REST-layer.md
nabdelgadir: I don't think this part was updated

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
```diff
@@ -55,6 +55,21 @@ describe('validate properties', () => {
       phoneNum: '999-111-1111',
       capacity: 10,
     };
-    await client.post('/coffee-shops').send(invalidAreaCodeCS).expect(422);
+    await client.post('/coffee-shops').send(invalidAreaCodeCS).expect(400);
+  });
+
+  it('create CoffeeShop failed with invalid type', async () => {
```
nabdelgadir, line 61: `fails when creating a CoffeeShop with an invalid type`?

examples/validation-app/src/interceptors/validate-phone-num.interceptor.ts
```diff
@@ -60,7 +60,8 @@ export class ValidatePhoneNumInterceptor implements Provider<Interceptor> {
       const err: ValidationError = new ValidationError(
         'Area code and city do not match',
       );
-      err.statusCode = 422;
+      // change the error status code to 400 - "Bad Request"
+      err.statusCode = 400;
```
nabdelgadir, line 64: Just wondering, why the update here?

examples/validation-app/src/interceptors/validate-phone-num.interceptor.ts
```diff
@@ -72,14 +73,11 @@ export class ValidatePhoneNumInterceptor implements Provider<Interceptor> {
...
-    if (
-      !(
-        city.toLowerCase() === 'toronto' &&
-        (areaCode === '416' || areaCode === '647')
-      )
-    )
-      return false;
 
+    if (city.toLowerCase() === 'toronto') {
+      if (areaCode === '416' || areaCode === '647') return true;
```
nabdelgadir, line 78: I think this can be simplified to 
```ts
return areaCode === '416' || areaCode === '647';
```

examples/validation-app/src/providers/my-validation-error-provider.ts
```diff
@@ -0,0 +1,31 @@
+import {HandlerContext, HttpErrors, RejectProvider} from '@loopback/rest';
```
nabdelgadir, line 1: Please add copyright to the top of the file

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,47 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+      if (httpError.statusCode === 422) {
+        const newError = {
+          message: 'My customized validation error message',
+          code: 'VALIDATION_FAILED',
+          resolution: 'Contact your admin for troubleshooting.',
+        };
+
+        // you can change the status code here too
+        response.status(422).send(JSON.stringify(newError));
+      }
```
bajtos, line 216: Missing `return`?

At the moment, you send the custom response and also call the default reject provider. Is that intended?

Fortunately, our error handler detects the situation when the response has been already sent and becomes no-op, so this isn't going to crash the app, but still.

An important side effect to be aware of: `RejectProvider` is handling error logging too. Depending on how you write your Reject provider, the error may or may not be logged for custom error path. (Your current solution will log the error because it always call `super.action`).

Since your example is interested only in changing the HTTP response,  maybe it would be better to change `RejectProvider` to receive the implementation of `writeErrorToResponse` via dependency injection too, so that code like yours can tweak only the part creating the HTTP response?

https://github.com/strongloop/loopback-next/blob/29e2076bff3cce7ca4445bde720f4cf9f1dae0e9/packages/rest/src/providers/reject.provider.ts#L41

bajtos:
I was on a call with few IBMers interested in custom error handling and it occurred to me that the solution based on a custom Reject implementation may be too complex. Unless the developer want to share the custom error handling, then I think it's much easier to implement the custom path inside the sequence (`src/sequence.ts` in projects scaffolded via `lb4`).

```ts
  // ...
    try {
      // ...
    } catch (error) {
      const httpError = <HttpErrors.HttpError>error;
      if (request.url === '/coffee-shops') {
        // if this is a validation error
        if (httpError.statusCode === 422) {
          const newError = {
            message: 'My customized validation error message',
            code: 'VALIDATION_FAILED',
            resolution: 'Contact your admin for troubleshooting.',
          };

          // you can change the status code here too
          response.status(422).send(JSON.stringify(newError));

          // TODO: show how to log the error using RestBindings.SequenceActions.LOG_ERROR

          // The error was handled
          return;
        }
      }

      // Otherwise fall back to the default error handler
      this.reject(context, error);
    }
  }
}
```

Just an idea to consider.

bajtos:
For longer term, I think it would be best to implement support for a custom JSON serializer. This will allow app developers to simply change `RestBindings.ERROR_WRITER_OPTIONS` to provide a custom function that will be called by LB to convert error objects into HTTP response bodies. See https://github.com/strongloop/strong-error-handler/issues/8

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
dhmlau: will update for the rest of the test case names too.

examples/validation-app/src/providers/my-validation-error-provider.ts
dhmlau: ran `slt copyright` in this example.

examples/validation-app/src/interceptors/validate-phone-num.interceptor.ts
dhmlau: let me change it back. :)

dhmlau: UPDATED: I recalled why I set it to `400`. Since I've handled the error coming from this interceptor, I want to set it to be 400 (bad request) rather than 422 (Unprocessable Entity) error. Otherwise, the error in here will be processed again in the `sequence.ts`.  Hope it makes sense.

dhmlau:
Thanks @bajtos for the code snippet. I was trying to figure out how to fall back to the default implementation. :). 
I've updated accordingly to your review comments in https://github.com/strongloop/loopback-next/pull/4969/commits/312e496a8143f0476b27815b2dfcea0a60d1b3b1, PTAL again, @nabdelgadir @bajtos. Thanks!

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,76 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+        if (httpError.statusCode === 422) {
+          const customizedMessage = 'My customized validation error message';
+          const customizedProps = {
+            resolution: 'Contact your admin for troubleshooting.',
+            code: 'VALIDATION_FAILED',
+          };
+
+          // 3. Create a new error with customized properties
+          // you can change the status code here too
+          const newError: HttpErrors.HttpError = createHttpErrors(
```
bajtos, line 226: Do we really need a new error object? Please note that this new error will have a stack trace pointing to this line, not to the code that triggered the original error.

If we want to send this directly to HTTP response, then a plain data object should be good enough.

```ts
const errorData = {
  statusCode: 422,
  message: customizedMessage;
  ...customizedProps,
};

if (errorWriterOptions?.debug) {
  errorData.stack = httpError.stack;
}

context.response.status(422).send(newError);
```

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,76 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+
+1. inject RestBindings.ERROR_WRITER_OPTIONS for logging error
+2. customize error for particular endpoints
+3. create a new error with customized properties
+4. log the error using RestBindings.SequenceActions.LOG_ERROR
+
+{% include code-caption.html content="/src/sequence.ts" %}
+
+```ts
+// 1. inject RestBindings.ERROR_WRITER_OPTIONS for logging error
```
bajtos, line 198: Can you please include the line defining the sequence class, to make it more clear that both `constructor` and `handle` belong to the sequence class? This may be also needed to get property syntax highlighting.

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,76 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+    @inject(SequenceActions.REJECT) public reject: Reject,
+    @inject(RestBindings.ERROR_WRITER_OPTIONS, {optional: true})
+    protected errorWriterOptions?: ErrorWriterOptions,
+  ) {}
+
+async handle(context: RequestContext) {
+    try {
+      //...
+    } catch (err) {
+      const httpError = <HttpErrors.HttpError>err;
```
bajtos, line 213: The catch block is very long, please extract it into a new method.

```ts
class MySequence /*...*/ {
  // ...
     try {
       //...
     } catch (err) {
        this.handleError(err);
     }
  }

  handleError(err: Error) {
    const httpError = <HttpErrors.HttpError>err;
    // ...

    // Otherwise fall back to the default error handler
    this.reject(context, err); 
  }
}
```

docs/site/Validation-REST-layer.md
```diff
@@ -175,3 +175,84 @@ of 10 digits separated by `-` after the 3rd and 6th digits.
...
+    protected logError: LogError,
+    @inject(RestBindings.ERROR_WRITER_OPTIONS, {optional: true})
+    protected errorWriterOptions?: ErrorWriterOptions,
+  ) {}
+
+  async handle(context: RequestContext) {
+    try {
+      // ...
+    } catch (err) {
+      this.handleError(context, <HttpErrors.HttpError>err);
```
bajtos, line 214: I think we prefer to use `as Type` instead of `<Type>` syntax for explicit type-casts.

```suggestion
      this.handleError(context, err as HttpErrors.HttpError);
```

PR #4972, opened 2020-03-25
chore: improve npm release scripts

raymondfeng:
Improve npm build scripts:
- Add `update-peer-deps` to maintain peer dependencies during the release process
- Move update of package-lock.json out of `version` phase to ensure testing against latest versions and remove duplicate update.
- Improve rebuild-package-locks to accept optional lerna scope to rebuild matched packages only.

## Checklist


bin/rebuild-package-locks.js
```diff
@@ -13,47 +13,69 @@
...
 /**
  * Rebuild package-lock.json files:
  *
- * 1. Remove node_modules and package-lock.json for all packages, including the
- * root one
- * 2. Run `npm install` to regenerate package-lock.json files
+ * 1. Remove node_modules and package-lock.json for matching packages,
+ * including the root one if no scopes is specified
+ * 2. Run `npx lerna bootstrap or npm install` to regenerate package-lock.json
+ * files
```
bajtos: I am not a big fan of updating package lock files during our release process.
- I don't want the release commit to be cluttered with changes in package-lock files. 
- Also an update of a (deep) dependency may introduce a bug. This bug could be platform-specific, local `npm test` may not catch it. IMO, we should always test package-lock updates via our CI. 
- If the bug does not manifest in our CI env but only later in user environments, then the task of reverting the package-lock update becomes tricky, as the package-lock changes are coupled with the new version release. 

IMO, update of package-lock files should be a standalone commit.

We are using RenovateBot to update the lock files. As long as we merge all outstanding RenovateBot PRs before starting the release process, then I believe there is no need to touch package locks in our release process.

If we think it's important to touch package locks as part of the release process, then please make sure those changes are committed separately from version-specific changes.

bin/update-peer-deps.js
```diff
@@ -0,0 +1,65 @@
...
+    const pkgJsonFile = p.manifestLocation;
+    const pkgJson = require(pkgJsonFile);
+    let updated = false;
+    if (pkgJson.peerDependencies) {
+      // Check all entries in peerDependencies
+      for (const d in pkgJson.peerDependencies) {
+        const pkgVersion = lbModuleVersions[d];
+        if (pkgVersion) {
+          // Update the version range to be `^<major>.0.0`
+          pkgJson.peerDependencies[d] = `^${semver.major(pkgVersion)}.0.0`;
```
bajtos: Should we use the actual `pkgVersion`, since that's the version we are testing in our CI? If we use `MAJOR.0.0`, then it's possible to create a situation where the peer dep spec is too wide, because the extension depends on features introduced in semver-minor version after `MAJOR.0.0`.

Another aspect to consider: removing a version from `peerDependencies` range is a breaking change, similarly to how dropping support for a Node.js major version is a breaking change too. 

It would be great if `lerna` was handling these complexities automatically for us, see the discussion in https://github.com/lerna/lerna/issues/955 and https://github.com/lerna/lerna/issues/1018

bin/rebuild-package-locks.js
raymondfeng: I agree. The `build:full` script is now changed to reflect CI behavior:

```
"build:full": "npm ci --ignore-scripts && lerna bootstrap --ci && npm run clean && npm run build",
```

It basically run `npm ci` for all packages to establish a clean and frozen state for all dependencies. Then a clean TSC build is performed.

bin/update-peer-deps.js
raymondfeng: Let's be conservative to use the `^<pkgVersion>` for peer dependencies as a release is cut.

raymondfeng:
@bajtos PTAL.

bin/rebuild-package-locks.js
bajtos: Sounds good. Can you please update this comment to match the actual implementation?

raymondfeng: The comment matches the implementation of `rebuild-package-locks`. The release script does not use it to have a clean build.

PR #4984, opened 2020-03-28
feat: add a script to scaffold and bootstrap a new package

raymondfeng:
We often need to add new packages to `loopback-next`. Copy/paste is error prone. This script scaffolds a new project under `packages` or `extensions` depending on the current directory. Then it uses `lerna bootstrap` to set up dependencies.

Depends on https://github.com/strongloop/loopback-next/pull/4994

## Checklist

### DIFF
--- a/docs/site/DEVELOPING.md
+++ b/docs/site/DEVELOPING.md
@@ -549,6 +549,40 @@ repository.
 
 ### Create a new package
 
+Please run the following command:
+
+```sh
+cd loopback-next
+node bin/create-package.js
+```
+
+The script does the following steps:
+
+1.  Determine the location and package name.
+
+    The first argument of the command can be one of the following:
+
+    - package-name
+    - @loopback/package-name
+    - extensions/package-name
+    - packages/package-name
+
+    If the location is not specified, it tries to guess by the current directory
+    and falls back to `extensions`.
+
+2.  Run `lb4 extension` to scaffold the project without `npm install`. If
+    `--yes` or `-y` is provide by the command, interactive prompts are skipped.
+
+3.  Tidy up the project
+
+    - Remove unused files
+    - Rename `tsconfig.json` to `tsconfig.build.json`
+    - Improve `package.json`
+
+4.  Run `lerna boostrap` on the newly added package to set up dependencies
+
+If you would like to do it manually, follow steps below:
+
 To add a new package, create a folder in
 [`packages`](https://github.com/strongloop/loopback-next/tree/master/packages)
 as the root directory of your module. For example,
--- a/bin/create-package.js
+++ b/bin/create-package.js
@@ -1,0 +1,214 @@
+#!/usr/bin/env node
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback-next
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * This is an internal script to add a new package to `packages` or `extensions`
+ * of the `loopback-next` monorepo.
+ *
+ * The script does the following steps:
+ *
+ * 1. Determine the parentDir and package name.
+ *
+ * The first argument can be one of the following:
+ * - package-name
+ * - @loopback/package-name
+ * - extensions/package-name
+ * - packages/package-name
+ *
+ * If the parentDir is not specified, it tries to guess by the current directory
+ * and falls back to `extensions`.
+ *
+ * 2. Run `lb4 extension` to scaffold the project without `npm install`. If
+ * `--yes` or `-y` is provide by the command, interactive prompts are skipped.
+ *
+ * 3. Tidy up the project
+ * - Remove unused files
+ * - Rename `tsconfig.json` to `tsconfig.build.json`
+ * - Improve `package.json`
+ *
+ * 4. Run `lerna boostrap` on the newly added package to set up dependencies
+ */
+'use strict';
+
+const build = require('../packages/build');
+const path = require('path');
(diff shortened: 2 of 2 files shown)


bin/add-extension-package.js
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env node
```
bajtos: Since this script can be used to add all kinds of packages (`extensions`, `packages`, `acceptance`, etc.), I find the name `add-extension-package.js` as misleading. Let's find a better one, for example:

- `bin/create-package.js`
- `bin/add-new-package.js`

bin/add-extension-package.js
```diff
@@ -0,0 +1,59 @@
...
+'use strict';
+
+const build = require('../packages/build');
+const path = require('path');
+const cwd = process.cwd();
+
+let name = process.argv[2];
+if (name == null) {
+  console.error(
+    'Usage: %s <package-name>',
```
bajtos: Please describe how is the script going to determine where to put the new package (`packages`, `extensions`, etc.).

Personally, I would like the script to support the following usage:

```
$ node bin/create-package extensions/my-cool-pkg
```

bin/add-extension-package.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+if (name.startsWith('@loopback/')) {
+  name = name.substring('@loopback/'.length);
+}
+
+const rootPath = path.resolve(__dirname, '..');
+let location = path.relative(rootPath, cwd);
+if (location !== 'packages' && location !== 'extensions') {
+  location = 'extensions';
```
bajtos: Please print a message to tell the user about the location picked by the tool.

We may want to ask the user to confirm the final location before proceeeding to create the new project.

We also may want to ask few questions about the new package, e.g. what `description` to put into `package.json`.

bin/add-extension-package.js
```diff
@@ -0,0 +1,59 @@
...
+// Run `lb4 extension`
+const child = build.runCLI(path.join(rootPath, 'packages/cli/bin/cli-main'), [
+  'extension',
+  `@loopback/${name}`,
+  '--outdir',
+  name,
+  '--yes',
+  '--skip-install',
+]);
+
```
bajtos: How about the remaining steps described in https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#adding-a-new-package?

Ideally, the script should run them all.

If that's too much work, then please print a descriptive message at the end to tell the user what are the next required steps in the package registration process, and include a link to our DEVELOPING docs.

---

Our docs are missing instructions for setting `repository.directory` field based on the location in the monorepo, see e.g. here:

https://github.com/strongloop/loopback-next/blob/1daa2b83bbd2474ab42d5227f4b9b97ef8b2e443/packages/core/package.json#L42-L46

Please make sure your tool is setting correctly all `repository` metadata.

bin/add-extension-package.js
```diff
@@ -0,0 +1,59 @@
...
+const rootPath = path.resolve(__dirname, '..');
+let location = path.relative(rootPath, cwd);
+if (location !== 'packages' && location !== 'extensions') {
+  location = 'extensions';
+  process.chdir(path.join(rootPath, 'extensions'));
+}
+
+console.log('Adding project %s/%s...', location, name);
+
+// Run `lb4 extension`
```
bajtos: What if we want to add a regular package (`packages/{name}`) or even an acceptance test (`acceptance/{name}`) - is `lb4 extension` a good command to use?

raymondfeng: For regular package, yes. But not for acceptance tests.

bin/add-extension-package.js
raymondfeng: By default, the interactive prompts will come up unless `--yes` or `-y` is specified.

bin/add-extension-package.js
raymondfeng: Fixed.

raymondfeng:
> Without that, I don't see much value in this script, when compared to running lb4 extension && lerna bootstrap manually.

Please see updated `DEVELOPING.md`. The script automates most steps except updating CODEOWNER and MONOREPO.md. But we do print out a message after the command finishes.

raymondfeng:
I'll add the copyright step once https://github.com/strongloop/loopback-next/pull/4994 is landed.

raymondfeng:
> I feel it's important to write some tests to verify what content the new package contains and even more importantly to verify that it's correctly registered in our monorepo (see https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#adding-a-new-package).

Not sure we want to write test for internal scripts.

bajtos:
Thank you for the updates, I'll take another look next week.

raymondfeng:
@bajtos Since it's an internal script. I went ahead to merge it. Please continue to provide feedback so that we can improve it if necessary.

docs/site/DEVELOPING.md
```diff
@@ -549,6 +549,40 @@ repository.
...
+    If the location is not specified, it tries to guess by the current directory
+    and falls back to `extensions`.
+
+2.  Run `lb4 extension` to scaffold the project without `npm install`. If
+    `--yes` or `-y` is provide by the command, interactive prompts are skipped.
+
+3.  Tidy up the project
+
+    - Remove unused files
+    - Rename `tsconfig.json` to `tsconfig.build.json`
```
bajtos, line 579: @raymondfeng I think this is no longer accurate, now that we are using project references (via https://github.com/strongloop/loopback-next/pull/5155)?

PR #4994, opened 2020-03-30
feat(cli): add `lb4 copyright` command to generate/update file headers

raymondfeng:
Add `lb4 copyright` command to update JS/TS file headers with copyright and license. Ported from [`strong-tools`](https://www.npmjs.com/package/strong-tools#slt-copyright) with enhancement to support lerna managed monorepo.

Todos:

- [x] Add tests
- [x] Add docs

## Checklist


packages/cli/generators/copyright/header.js
```diff
@@ -0,0 +1,287 @@
...
+
+// Components of the copyright header.
+const COPYRIGHT = [
+  'Copyright <%= owner %> <%= years %>. All Rights Reserved.',
+  'Node module: <%= name %>',
+];
+const LICENSE = [
+  'This file is licensed under the <%= license %>.',
+  'License text available <%= ref %>',
+];
```
bajtos: Have you considered using ES6 template literals instead of lodash templates?

```ts
function license({license, ref}) {
  return [
    `This file is licensed under the ${license}.`,
    `License text available ${ref}`,
  ];
}
```

raymondfeng: It's a bit involved. We can defer it for future improvement.

docs/site/Copyright-generator.md
```diff
@@ -0,0 +1,76 @@
...
+
+The command prompts you for:
+
+1. The copyright owner. The default value is read from `copyright.owner` or
+   `author` in the `package.json`.
+
+2. The license name. The default value is read from `license` in the
+   `package.json`.
+
+The default owner is `IBM Corp.` and license is `MIT` with the following
```
dhmlau: IIRC if it behaves the same way as strong-tools, if there's no value set for `copyright.owner` and `author`, the default owner is `Owner`.

raymondfeng:
Updated `docs/site/tables/lb4-project-commands.html`

docs/site/Copyright-generator.md
raymondfeng: It says `with the following package.json` as we read values from package.json as defaults for the prompt.

achrinza:
It's probably out-of-scope, but we could create a future improvement PR to...

1. Leverage [spdx-license-list](https://www.npmjs.com/package/spdx-license-list)

2. Utilize `SPDX-License-Identifier`

raymondfeng:
> Leverage spdx-license-list
> Utilize SPDX-License-Identifier

Thank you for the suggestion. I would like to integrate with the module but it will add at least 7MB to the dependency tree :-(.

> I'm wondering if it should be ported into the LoopBack CLI. A copyright header management tool may be better-suited as a separate package that can be depended upon and called from the LoopBack CLI. This would allow users to utilize the tool a-la carte without needing the entire LoopBack CLI.

The utility is ported from https://www.npmjs.com/package/strong-tools and improved with lerna support. We can possibly break down `cli` module into a few packages in the future and introduce extensibility via `@loopback/context`.

raymondfeng:
@achrinza I have improved the PR to use spdx list and https://www.npmjs.com/package/inquirer-autocomplete-prompt. But it seems to be causing test issues.

Issue #1464, opened 2018-06-25
[Spike] Allow extensions to contribute custom convention for environment-specific operational configuration


bajtos:
**Timeboxed to 2 weeks**; don't spend extra time if done earlier

Different platforms use different ways for configuring operational aspects of application in test/dev/production. LB4 should make it easy to write extensions to support these different platforms.

The goal of this spike is to find out and document what's possible today, identify gaps, propose solutions and create a list of follow-up tasks.

Few examples of operational config:
 - configure datasources from environment variables
 - use a faster but less-secure hashing algorithm in dev/test (a hashing algorithm is used e.g. to store user passwords)
 - apply datasource configuration provided by `VCAP_SERVICES` env variable in CloudFoundry/IBM Cloud - see https://github.com/strongloop/loopback-next/pull/1574

Possibly related: https://github.com/strongloop/loopback-next/pull/983 and https://github.com/strongloop/loopback-next/issues/1396

Edo78:
Is there any news about this issue?
Right now the only way I have found to configure a datasource with environment variables is to instantiate the related class (in application.ts) passing a custom configuration
```Javascript
let dbDataSource = new DbDataSource({
  name: 'db',    
  connector: 'mysql',
  hostname: process.env.DB_HOST,
  port: process.env.DB_PORT,
  user: process.env.DB_USER,
  password: process.env.DB_PASSWORD,
  database: process.env.DB_DATABASE
});
```
but I'm not sure this could be a best practice.
Am I missing something?

bajtos:
@Edo78 see the following resources to learn about other options available:
 - https://loopback.io/doc/en/lb4/Deploying-to-IBM-Cloud.html#updating-application
 - https://itnext.io/loopback-4-database-configuration-8f085399268

TL;DR: if you are using `@loopback/boot` to load your datasources (as is the default in LB4 applications scaffolded using `lb4` CLI tool), then you can bind just the custom datasource configuration.

```ts
this.bind('datasources.config.db').to({
  name: 'db',    
  connector: 'mysql',
  hostname: process.env.DB_HOST,
  port: process.env.DB_PORT,
  user: process.env.DB_USER,
  password: process.env.DB_PASSWORD,
  database: process.env.DB_DATABASE
});
```

bajtos:
Cross-posting from https://github.com/strongloop/loopback-next/issues/1609

https://github.com/strongloop/loopback-next/issues/441#issuecomment-392522724 #4


[...]

PR #4030, opened 2019-10-31
chore: improve issue templates

bajtos:
Leverage the recently introduced GitHub feature and rework our issue picker:

- Disallow free-form issue creation, force users to pick Feature or Bug
- Replace Security item with a link explaining how to disclose
  vulnerabilities via email
- Replace Question item with a link to StackOverflow and our mailing
  list

See https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

You can view the proposed picker here:
https://github.com/bajtos/test-issue-templates/issues/new/choose

Screen Shot 2019-10-31 at 16 36 34

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

Issue #844, opened 2018-01-05
Get rid of Yeoman


bajtos:
Refactor the CLI package to use [inquirer](https://www.npmjs.com/package/inquirer) and [ejs](https://www.npmjs.com/package/ejs) directly, get rid of the dependency on Yeoman.

This will allow us to use TypeScript too, we should migrate the CLI codebase to the standard TypeScript project as part of this task.

Related:

 - https://github.com/strongloop/generator-loopback/issues/139
 - https://github.com/strongloop/loopback-next/pull/825/files#r159253215

privateOmega:
@bajtos I would like to help however I can on this issue.

bajtos:
@raymondfeng Here are the issues I was experiencing in the old generator-loopback frequently:
 - Our "model" generator is invoking "property" generator in a loop. This is not supported by Yeoman anymore, we have a hacky workaround in place. As a result, upgrades to newer versions of Yeoman were usually painful, because the hacky workaround had to be updated to match the new internals of Yeoman.
 - Invoking a generator directly from CLI is a bit of PITA, there are more hacks needed to make this happen - see e.g. https://github.com/strongloop/loopback-next/blob/6cc83b666e4b2ce2f979eab44dec78b7c01c41be/packages/cli/bin/cli.js#L66-L68
 - I find individual generators difficult to test in isolation. Verifying how CLI arguments are handled are even more difficult. 
 - AFAICT, there is no way how to test non-trivial prompting logic - for example the loop asking for multiple properties. A mocked prompt can define only a single set of answers. Running the generator/CLI end-to-end and parsing stdout is brittle because Inquirer uses lots of ANSI control codes to control what's displayed on the screen.
 - Subjectively, I dislike the conventions forced by Yeoman: 
    - All generators live in different `index.js` files, which makes navigation difficult to me. Instead of `app/index.js`, I'd like to have `app.js` file.
    - Implementing individual steps as named methods on a generator object made sense back in ES5 days, because it simplified async flow control. Today with ES2017's `async`/`await`, I prefer to write idiomatic JavaScript instead.

@privateOmega thank you for chiming in! Right now, we are debating whether we actually want to drop Yeoman or whether it's perhaps not that bad. Once we reach consensus, I can show you pointers where to start.

bajtos:
[...]

PR #5000, opened 2020-03-31
Move datasource config from JSON to TS files

bajtos:
For quite some time I was thinking about our current project layout for datasources and the issues it brings:
1. When using connectors from npm in example repositories, we need to modify the static config from the JSON file to replace the connector name with the real connector module.
2. Dynamic configuration of datasources requires changes in the main application (see https://github.com/strongloop/loopback-next/issues/1464#issuecomment-436260086), which is difficult to reason about.
3. DataSource JSON file is a kind of an oddball, all other artifacts (models, repositories, etc.) are using TypeScript files only.

IIRC, we introduced DataSource JSON files because it was the design already used by LB3 and loopback-workspace. We were anticipating that in early future (from the point of view of 2017), we will write tooling to edit DataSource configuration programmatically - somethings that's easier to implement for JSON files rather than TS files.

Clearly, programmatic editing of datasource config didn't become a priority in the last several years and most likely won't become important anytime soon.

Additionally, we already have pretty good tooling for editing TypeScript source code programmatically, we use it to edit model and repository files when scaffolding new relations.

In that light, I feel there is no need to use DataSource JSON files any more and it's better to move the definition of the (default) configuration object to the "main" TypeScript file.

And that's exactly what this pull request is about. I split the changes into three parts to make them easier to review.

1. f591ab4 updates example applications,
2. fdf15c2 updates CLI templates,
3. ff8db3a updates the docs.
 
~~**Open questions**~~

~~What's the best way for communicating these change to our users? Should we treat the CLI change as a breaking change? The new version is not going to break existing projects, but it will start emitting different source code for new datasources. Should we provide a migration guide?~~

**UPDATE 2020-04-06**

Based on the discussion, I added a migration guide to [Datasource-generator.md]( https://loopback.io/doc/en/lb4/DataSource-generator.html) page and implemented backwards compatibility layer to support both JSON-based and TypeScript-based configuration of datasources.

My intention is to label this change as a backwards-compatible feature triggering a semver-minor release.

**Out of scope**

While reviewing all things related to DataSources, I found few more opportunities for cleanup. I prefer to leave them out of this PR, but would like to implement them soon after this patch is landed.

1. Update https://github.com/strongloop/loopback4-example-shopping and https://github.com/strongloop/loopback-next/blob/master/docs/site/Inside-Loopback-Application.md to follow the new style
1. Make the static DataSource property `name` read-only, set it from `config.name` to avoid duplication.
1. Rework datasources to receive the configuration via `@config`, so that they can be configured via `app.configure()`. Update the docs (especially Cloudant & Bluemix guides), make sure to preserve content for the old configuration style (think of existing projects created before the change).
1. Fix datasource files in `examples` to use the same style as scaffolded by `lb4 datasource`. Most notably, change them to participate in app life-cycle and disconnect when the application has stopped. Update the docs accordingly. 
1. Verify that integration tests instantiating datasource classes directly are stopping them after the tests are over + update the relevant docs (e.g. testing best practices).

I'd like to get feedback especially from @raymondfeng as our architect and @nabdelgadir as the FA owner of LB4 project conventions.

/cc @strongloop/loopback-next because this change is going to affect most of our users.

## Checklist

**TODO**

Fix the tests (and possibly the implementation) of the following CLI commands to support & test both the new style but also the legacy one.

- [x] `lb4 relation`
- [x] `lb4 repository`
- [x] `lb4 rest-crud`
- [x] `lb4 service`
- [x] `lb4 openapi`


achrinza:
> What's the best way for communicating these change to our users?

Other than through the changelogs, we could...

1. Create an "info" alert in the [CLI References docs](https://loopback.io/doc/en/lb4/Command-line-interface.html)

2. (long-term proposal) Create a page similar to [update.angular.io](https://update.angular.io/)

3. (long-term proposal) Create a read-only `announcement` channel on Slack for users who want a low-traffic channel (loosely related to https://github.com/strongloop/loopback-next/pull/4030#issuecomment-549057758)

> Should we treat the CLI change as a breaking change?

As there isn't a clear scope on which parts of the CLI are expected to be consistent, it should be considered a breaking change.

Examples of scopes:

- JSON-based config ingestion using `--config`
- Files generated by the CLI
- CLI commands (e.g. `lb4 app`)
- Module API (importing `@loopback/cli` as a Typesscript module)

bajtos:
Thank you @achrinza for chiming in!

> Create an "info" alert in the CLI References docs

That's what I am leaning towards too. I am thinking about adding a sub-section to https://loopback.io/doc/en/lb4/DataSource-generator.html, where I can explain the difference in the output of `lb4 datasource` in versions 2.2.1 (or older) and the to-be-current (or newer) versions, and also provide a short migration guide.

> As there isn't a clear scope on which parts of the CLI are expected to be consistent, it should be considered a breaking change.

IIUC, you are saying that if we are not sure, then we should treat the change as breaking? Such strategy seems a bit too conservative to me.

My changes are not going to break any existing projects or applications. After the CLI is upgraded, `lb4 datasource` will emit different files than before, but that's going to affect only newly scaffolded files. All existing datasources will keep working as before.

The question is whether `lb4 repository` and `lb4 service` will support both JSON and TS datasources. At the moment, only JSON datasources are supported - that's something I need to fix before this pull request can be landed. I think the fix should preserve support for both (legacy) JSON and (new) TS style, so that users are not forced to migrate their existing datasources at once.

https://github.com/strongloop/loopback-next/blob/e5a48fc82315292878883e53f63ccf197edbc373/packages/cli/lib/utils.js#L571-L615

Do you still think that a semver-major release is needed when the only change is in the newly scaffolded files?

docs/site/Repositories.md
```diff
@@ -307,50 +307,41 @@ We can now access key-value stores such as [Redis](https://redis.io/) using the
...
+  connector: 'kv-redis',
+  host: '127.0.0.1',
+  port: 6379,
+  password: '',
+  db: 0,
+};
+
+export class RedisDataSource extends juggler.DataSource {
+  static dataSourceName = 'redis';
+
```
raymondfeng: Missing `static defaultConfig = config;`?

raymondfeng:
Does discovery of datasources continue to work for related `lb4` commands?

nabdelgadir:
I think the proposal at a high level is good, but +1 to @raymondfeng's comment. It's important to make sure that discovery related CLI commands are compatible. 

> I am thinking about adding a sub-section to https://loopback.io/doc/en/lb4/DataSource-generator.html, where I can explain the difference in the output of lb4 datasource in versions 2.2.1 (or older) and the to-be-current (or newer) versions, and also provide a short migration guide.

👍 This is what I would've proposed as well. 

> Should we treat the CLI change as a breaking change?

We can see how the other CLI commands are affected, then decide this.

docs/site/Repositories.md
bajtos: Fixed in 8c73f9d

bajtos:
Hello 👋 I updated the pull request with the following improvements:

- Rebased on top of the latest master
- Fixed @raymondfeng's [review comment](https://github.com/strongloop/loopback-next/pull/5000#discussion_r401175423) in 8c73f9d
- Added migration guide to 203a1b4
- In eb5774b, I fixed `lb4 repository` tests to use the new style, added one test to verify support for the legacy style, fixed the implementation to support both styles.

I'd like to get your feedback on this increment, @raymondfeng @nabdelgadir @achrinza 

What's remaining: fix the tests of the following CLI commands:
  - `lb4 service`
  - `lb4 rest-crud`
  - `lb4 relation`

raymondfeng:
I think it's time to break down `@loopback/cli` into one core and a few plugins. Maybe convert it to TypeScript and start to use `@loopback/core` for extensibilities.

docs/site/DataSource-generator.md
```diff
@@ -50,10 +50,93 @@ Once all the prompts have been answered, the CLI will do the following:
...
+   ```
+
+4. Modify the tests importing the default datasource configuration from the JSON
+   file, get the default configuration via the new static property instead.
+
+   For example:
+
+   ```diff
+   -import GEO_CODER_CONFIG from '../datasources/geocoder.datasource.config.json';
+   +import {GeocoderDataSource} from '../datasources/geocoder.datasource';
```
nabdelgadir, line 132: Maybe we could use the same datasource throughout instead of using all of `db`, `ds`, and the geocoder one

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -517,6 +517,62 @@ describe('lb4 repository', function () {
       );
     });
   });
+
+  describe('legacy JSON-based configuration', () => {
```
nabdelgadir, line 521: 👍

packages/cli/test/integration/generators/repository.integration.js
```diff
@@ -517,6 +517,62 @@ describe('lb4 repository', function () {
...
+    it('loads config from `{name}.datasource.config.json', async () => {
```
nabdelgadir, line 522: missing ending `

packages/cli/lib/ast-helper.js
```diff
@@ -141,3 +141,74 @@ function isPrimaryKeyFlag(idInitializer) {
...
+
+exports.getDataSourceConfig = function (fileContent) {
+  const ast = tsquery.ast(fileContent);
+  // Find a top-level declaration of `config` variable
+  const configVarDecl = tsquery(
+    ast,
+    // It must be a top-level declaration, not inside a function
+    'SourceFile>VariableStatement>VariableDeclarationList>' +
+      // The declaration should be for property called `config`
+      'VariableDeclaration:has([name="config"])',
```
nabdelgadir, line 153: I'm a bit confused; does it have to be called `config` or does it have to include the word `config`?

bajtos:
> I think it's time to break down `@loopback/cli` into one core and a few plugins. Maybe convert it to TypeScript and start to use `@loopback/core` for extensibilities. 

Sure, but that's out of scope of this pull request.

Personally, I think we should start with extracting code for working with project artifacts into a new package, something similar to https://github.com/strongloop/loopback-workspace.

See also https://github.com/strongloop/loopback-next/issues/844_[CLI] Get rid of Yeoman_.

docs/site/DataSource-generator.md
bajtos: Good catch. I'll unify `db` and `ds` into one name.

Geocoder is a bit special, it's a datasource connected to a webservice (not a database), our users may be familiar with this datasource from our Todo tutorial - see https://loopback.io/doc/en/lb4/todo-tutorial-geocoding-service.html I prefer to keep geocoder as is now.

packages/cli/test/integration/generators/repository.integration.js
bajtos: Good catch!

packages/cli/lib/ast-helper.js
bajtos: I see now that the comment on line 152 is confusing.

The variable must be called exactly `config`, as in:

```ts
const config = // ...
```

bajtos: Let me explain in more detail.

`VariableDeclaration` node has two children:
 - `Identifier` describing the variable name,
 - `ObjectLiteralExpression` describing the variable value when it's an object.

The query `VariableDeclaration:has([name="config"])` is matching `VariableDeclaration` that has a child node with `name="config"`.

You can try https://tsquery-playground.firebaseapp.com to build a better understanding. The playground provides visualization of TypeScript AST tree for any code snippet you provide and can run queries too.

docs/site/DataSource-generator.md
bajtos: Fixed in 23c03c99.

packages/cli/lib/ast-helper.js
nabdelgadir: Ah okay, I see. Thank you for the explanation and link!

achrinza:
Since we can now effectively do variable substitution, this may be a good opportunity to port over [Specifying database credentials with environment variables](https://loopback.io/doc/en/lb3/Defining-data-sources.html#specifying-database-credentials-with-environment-variables) into the LoopBack 4 docs.

bajtos:
> Since we can now effectively do variable substitution, this may be a good opportunity to port over [Specifying database credentials with environment variables](https://loopback.io/doc/en/lb3/Defining-data-sources.html#specifying-database-credentials-with-environment-variables) into the LoopBack 4 docs.

Thank you @achrinza for raising this point!

It was indeed one of my goals to make it easier to provide datasource configuration from environment variables.

However, we haven't decided yet what should be the recommended way for handling environment-specific configuration in LB4. There was some discussion in https://github.com/strongloop/loopback-next/issues/1464 plus we already have some LB4 documentation in our deployment guides:
- CloudFoundry (legacy IBM Cloud from pre-Kubernetes days): https://loopback.io/doc/en/lb4/Deploying-to-IBM-Cloud.html#step-4-updating-the-application

We don't have any Kubernetes-specific advice, I am not sure if it's because Kubernetes deployments don't need to change datasource configuration or because we didn't consider that aspect when writing the guide.

Either way, I feel that documenting how to use environment variables in datasource config is out of scope of this pull request, the PR is already too large to my taste. Would you mind creating a follow-up issue for the docs you have in mind?

bajtos:
I pushed few new commits today, I consider the patch ready for final review. I'll clean up the commit history after the changes are approved.

List of recent changes:

- 3283629 address review comment for docs
- 16ed3dbmigration guide clean-up per review comments
- 41dc940 address review comments in tests 
- ffa7c01 update after rebase on master
- 2f2c023 fix `lb4 openapi`
- 86fff28  update after rebase on master (again)
- 7920ed8 update `lb4 relation` test fixtures
- 891ab1d preserve backwards compat for `lb4 rest-crud` 
- b9b3d15 preserve backwards compat for `lb4 service`

The code in tests is not as clean & consistent as I it would like to be, but I have only so much time available for refactoring existing cruft 🤷 

@achrinza @nabdelgadir @raymondfeng Do you have any more comments? Does the pull request LGTY now?

bajtos:
(Investigating test failures now. It looks like the test is not picking `positional: true` field from the datasource config.)

bajtos:
ad7732c should fix the build 🤞

bajtos:
Oh well, looks like snapshots are not the best tool when testing file paths :(

```diff
 const config = {
   name: 'petStore',
   connector: 'openapi',
-  spec: 'C:\projects\loopback-next\packages\cli\test\fixtures\openapi\3.0\petstore-expanded.yaml',
+  spec: '../../../fixtures/openapi/3.0/petstore-expanded.yaml',
   validate: false,
   positional: true
 };
```

bajtos:
892d733 is fixing the failing snapshot tests on Windows in a way that (I hope) will improve lives of all LB users on Windows

bajtos:
@achrinza @nabdelgadir @raymondfeng PTAL, does the pull request LGTY now?

packages/cli/lib/ast-helper.js
```diff
@@ -141,3 +141,74 @@ function isPrimaryKeyFlag(idInitializer) {
...
+  if (!configVarDecl || configVarDecl.length < 1) {
+    debug('No top-level declaration of variable "config" was not found.');
+    return undefined;
+  }
+
+  const initializer = configVarDecl[0].initializer;
+  const initializerKindName = syntaxKindName(initializer.kind);
+  if (initializerKindName !== 'ObjectLiteralExpression') {
+    debug(
+      'Variable "config" is declare with an unsupported initializer kind %s.',
```
nabdelgadir, line 164: `declare` -> `declared`

bajtos:
I have rebased the patch on top of the latest master and squashed the commits. Unless there are any more comments and/or objections, my plan is to go ahead and merge this change as a backwards-compatible feature on Friday April 24th.

PR #5011, opened 2020-04-02
feat: enable parallel mocha testing

raymondfeng:
Enable parallel mocha testing - see https://github.com/strongloop/loopback-next/issues/3159#issuecomment-607498466

I'm starting to break down this PR into a few smaller ones:

- [x] refactor: use sandbox.path instead of SANDBOX to access the sandbox directory #5014
- [x] fix(example-express-composition): use an assinged port number for testing #5022
- [x] feat(core): improve application signal handler registration #5020 
- [x] feat: improve test sandbox #5016
- [x] fix(http-server): ensure rejection handler is registered before the error event is emitted #5031
- [x] fix(context): fix context subscription manager to not emit errors to its context if there is an error listener is in place #5021
- [ ] Rework snapshot matcher to support parallel mocha #5813

## Checklist

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,16 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: loopback-next
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// Reuse the mocha config from `@loopback/cli`
+const config = require('./packages/cli/.mocharc.js');
+
+// Set max listeners to 16 for testing to avoid the following warning:
+// (node:11220) MaxListenersExceededWarning: Possible EventEmitter
+// memory leak detected. 11 SIGTERM listeners added to [process].
+// Use emitter.setMaxListeners() to increase limit
+// It only happens when multiple app instances are started but not stopped
+process.setMaxListeners(16);
+
+module.exports = config;
--- a/packages/build/bin/run-mocha.js
+++ b/packages/build/bin/run-mocha.js
@@ -15,6 +15,9 @@ Usage:
 
 'use strict';
 
+const path = require('path');
+const fs = require('fs-extra');
+
 function run(argv, options) {
   const utils = require('./utils');
 
@@ -56,6 +59,17 @@ function run(argv, options) {
   if (lang !== -1) {
     process.env.LANG = mochaOpts[lang + 1];
     mochaOpts.splice(lang, 2);
+  }
+
+  // Set `--parallel` for `@loopback/*` packages
+  if (!mochaOpts.includes('--parallel')) {
+    const pkgFile = path.join(utils.getPackageDir(), 'package.json');
+    if (fs.existsSync(pkgFile)) {
+      const pkg = fs.readJsonSync(pkgFile);
+      if (pkg.name.startsWith('@loopback/')) {
+        mochaOpts.push('--parallel');
+      }
+    }
   }
 
   const args = [...mochaOpts];
--- a/packages/build/bin/run-mocha.js
+++ b/packages/build/bin/run-mocha.js
@@ -24,7 +24,6 @@ function run(argv, options) {
     !utils.isOptionSet(
       mochaOpts,
       '--config', // mocha 6.x
-      '--opts', // legacy
       '--package', // mocha 6.x
       '--no-config', // mocha 6.x
     ) && !utils.mochaConfiguredForProject();
--- a/packages/build/config/.mocharc.json
+++ b/packages/build/config/.mocharc.json
@@ -1,5 +1,5 @@
 {
-  "require": "source-map-support/register",
+  "require": ["source-map-support/register"],
   "recursive": true,
   "exit": true,
   "reporter": "dot"
--- a/packages/cli/.mocharc.js
+++ b/packages/cli/.mocharc.js
@@ -1,0 +1,53 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/cli
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// A workaround to use `--require ./test/snapshot-matcher.js` so that root
+// hooks are executed by mocha parallel testing for each job
+
+const debug = require('debug')('loopback:cli:test');
+
+/**
+ * Build mocha config for `@loopback/cli`
+ */
+function buildConfig() {
+  // Use the default config from `@loopback/build`
+  const mochaConfig = require('@loopback/build/config/.mocharc.json');
+  debug('Default mocha config:', mochaConfig);
+  // Resolve `./test/snapshot-matcher.js` to get the absolute path
+  const mochaHooksFile = require.resolve('./test/snapshot-matcher.js');
+  debug('Root hooks for --require %s', mochaHooksFile);
+  const config = {...mochaConfig, timeout: 5000};
+
+  // Allow env var `MOCHA_JOBS` to override parallel testing parameters
+  const jobs = +process.env.MOCHA_JOBS;
+  if (jobs === 0) {
+    // Disable parallel testing
+    config.parallel = false;
+  } else if (jobs > 0) {
+    // Override the default number of concurrent jobs
+    config.parallel = true;
+    config.jobs = jobs;
+  }
+  addRequire(config, mochaHooksFile);
+  debug('Final mocha config:', config);
+  return config;
+}
+
+/**
+ * Add a new entry to the mocha config.require
+ * @param {object} config - Mocha config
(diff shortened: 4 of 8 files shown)


raymondfeng:
The snapshot-related cli tests are still failing. I think I have a theory as follows:

1. The snapshots.js creates a global beforeEach hook (which is done once per process with node.js module require)
2. Parallel mocha uses a pool of worker processes
3. When the worker process tests the 1st file, beforeEach is correctly called
4. When the worker process is reused to test a 2nd file, beforeEach is not available to mocha as the node require cache is not flushed. As a result, beforeEach is not invoked to refresh currentTest
5. As a result, the last currentTest from the 1st test file is accidentally used for other tests and it fails

raymondfeng:
@boneskull Is it possible to have something as follows to be picked up by `mocha` without `---require` or `--file`?

```js
'use strict';
exports.mochaHooks = {
  beforeAll() {
    console.log('beforeAll');
  },
  beforeEach() {
    console.log('beforeEach');
  },
  afterAll() {
    console.log('afterAll');
  },
  afterEach() {
    console.log('afterEach');
  }
};
```

bajtos:
@raymondfeng what's the status of this PR? I see that all smaller PRs were already landed. Are the snapshot failures the last missing piece? BTW the snapshot matcher is implemented in such way that it should be relatively easy to install the `beforeEach` hook inside every test file, we may not need to use Mocha hooks. Would you like me to help you with this part?

.mocharc.js
```diff
@@ -0,0 +1,16 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+// Reuse the mocha config from `@loopback/cli`
+const config = require('./packages/cli/.mocharc.js');
+
+// Set max listeners to 16 for testing to avoid the following warning:
+// (node:11220) MaxListenersExceededWarning: Possible EventEmitter
+// memory leak detected. 11 SIGTERM listeners added to [process].
+// Use emitter.setMaxListeners() to increase limit
+// It only happens when multiple app instances are started but not stopped
```
bajtos, line 13: I think it's a problem in our test suite when we start apps but don't stop them. Instead of suppressing the helpful warning, can we please fix our tests to always stop the applications started?

raymondfeng:
> @raymondfeng what's the status of this PR? I see that all smaller PRs were already landed. Are the snapshot failures the last missing piece? BTW the snapshot matcher is implemented in such way that it should be relatively easy to install the beforeEach hook inside every test file, we may not need to use Mocha hooks. Would you like me to help you with this part?

The PR is fully functional. We're waiting for a release from Mocha to land it.

If you would like to improve the `beforeEach` hook, please do. I tried but there was no success.

boneskull:
yes, I have some work to do to split the huge PR into parts. but it’s reasonably stable atm. 

once it lands I can publish a “beta” version before the next official release if you want to adopt early.

packages/build/config/.mocharc.json
```diff
@@ -1,7 +1,9 @@
 {
-  "require": "source-map-support/register",
+  "require": ["source-map-support/register"],
   "recursive": true,
   "exit": true,
-  "reporter": "dot"
+  "reporter": "dot",
+  "timeout": 5000,
+  "parallel": true
```
bajtos, line 7: I am against forcing all LB users to adopt parallel testing mode. As you know from our own experience, adapting a test suite to support parallel execution is a non-trivial amount of work. The situation is even more complicated when stateful 3rd-party services are involved, typically when the test suite works with data stored in a real database. (We are not affected much, because most of our tests are using in-memory datasources.)

In smaller projects, the cost of enabling parallel testing is often not worth the benefits.

I am concerned that by making parallel testing the default in new projects, we will make the framework more difficult to adopt for less-experienced developers.

bajtos:
Let's not mix the upgrade to Mocha v8.x with the transition to parallel tests please. I'd like to see a standalone PR to upgrade to Mocha 8.x and make the necessary updates in our config files. I think commits like cbba724 and f9af11e can be landed independently too.

Mocha 8.x dropped support for `mocha.opts` file, which we still support in `run-mocha` script. This should be fixed as part of the upgrade to Mocha 8.x.

boneskull:
agree that parallel mode shouldn’t be the default for _users_.

bajtos:
@raymondfeng can you please extract 7334f20 in to a standalone pull request to get it landed sooner, and make it easier for others (me in particular) to experiment with different approaches to parallel testing?

.mocharc.js
```diff
@@ -0,0 +1,16 @@
...
+process.setMaxListeners(16);
```
bajtos, line 14: How is the change in `stMaxListeners(16)` relevant to snapshot matching? I think this belongs to a different commit.

packages/cli/.mocharc.js
```diff
@@ -0,0 +1,53 @@
...
+ * Build mocha config for `@loopback/cli`
+ */
+function buildConfig() {
+  // Use the default config from `@loopback/build`
+  const mochaConfig = require('@loopback/build/config/.mocharc.json');
+  debug('Default mocha config:', mochaConfig);
+  // Resolve `./test/snapshot-matcher.js` to get the absolute path
+  const mochaHooksFile = require.resolve('./test/snapshot-matcher.js');
+  debug('Root hooks for --require %s', mochaHooksFile);
+  const config = {...mochaConfig, timeout: 5000};
```
bajtos, line 21: I find this approach brittle because it requires us to deal with root hooks both inside the package test setup and in monorepo test setup.

packages/build/bin/run-mocha.js
```diff
@@ -58,6 +61,17 @@ function run(argv, options) {
     mochaOpts.splice(lang, 2);
   }
 
+  // Set `--parallel` for `@loopback/*` packages
```
bajtos, line 64: I am not sure if this is a good idea, because it affects code outside our monorepo too. Two examples:

- Example projects created via `lb4 example <name>`. Besides the extra complexity related to parallel testing, I am concerned about the impact on performance.

- External repositories like https://github.com/strongloop/loopback4-example-shopping. I am not sure how much work is required to support parallel testing in the shopping example and I would rather not block upgrading to the latest `@loopback/build` version until we can figure out parallel testing.

I am proposing to enable parallel testing by changing `npm run mocha` script to add `--parallel` flag to mocha CLI options.

bajtos: See e.g. b44c7ee398e2aecdf89a5

```diff
-    "mocha": "node packages/build/bin/run-mocha --lang en_US.UTF-8 --timeout 5000 \"packages/*/dist/__tests__/**/*.js\" \"extensions/*/dist/__tests__/**/*.js\" \"examples/*/dist/__tests__/**/*.js\" \"packages/cli/test/**/*.js\" \"packages/build/test/*/*.js\"",
+     "mocha": "node packages/build/bin/run-mocha --lang en_US.UTF-8 --timeout 15000 --parallel \"packages/*/dist/__tests__/**/*.js\" \"extensions/*/dist/__tests__/**/*.js\" \"examples/*/dist/__tests__/**/*.js\" \"packages/cli/test/**/*.js\" \"packages/build/test/*/*.js\"",
```

Having said that, I think it's time to move flags from `run-mocha` CLI args into the monorepo-level `.mocharc.js` file. That way they are always applied, for example when running a subset of tests manually via `node packages/build/bin/run-mocha path/to/some/files`.

packages/cli/.mocharc.js
bajtos: In #5724, I have proposed a fix that does not leak the knowledge about snapshot hooks outside of CLI tests.

bajtos:
Opened https://github.com/strongloop/loopback-next/pull/5813 to rework the snapshot matcher to use Mocha global hooks, without actually enabling the parallel mode yet.

packages/cli/.mocharc.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // Allow env var `MOCHA_JOBS` to override parallel testing parameters
+  const jobs = +process.env.MOCHA_JOBS;
```
bajtos, line 24: I feel we don't need to support `process.env.MOCHA_JOBS` as it's already possible to change the number of jobs (worker processes) by using `-j` command line argument and disable parallel execution by setting the number of jobs to 1.

```
$ npm run mocha -- -j 1
```

.mocharc.js
bajtos: I am running Node.js 14.4.0 and haven't encountered `MaxListenersExceededWarning`. Could you please double check that this is still a problem? Is it perhaps specific to older versions of Node or a timing conditions on CI build servers?

bajtos:
I am proposing to close this pull request in favor of https://github.com/strongloop/loopback-next/pull/5831.

PR #5015, opened 2020-04-02
chore(core): set max listeners to 256 for `process` to avoid warnings

raymondfeng:
Step 2 for #5011 

Increase number of process listeners to avoid warnings.

## Checklist

### DIFF
--- a/packages/core/src/application.ts
+++ b/packages/core/src/application.ts
@@ -99,10 +99,7 @@ export class Application extends Context implements LifeCycleObserver {
     this.bind(CoreBindings.APPLICATION_CONFIG).to(this.options);
 
     const shutdownConfig = this.options.shutdown ?? {};
-    this.setupShutdown(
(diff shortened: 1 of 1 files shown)


bajtos:
When we disable max-listener warnings in our tests, then we won't notice if our code is leaking event listeners.

IMO, we should avoid changing the global max-listener configuration at all costs and look for ways how to change max-listeners config in specific places (on specific event emitter instances) only.

boneskull:
Which listeners do you mean?

Mocha isn’t listening for SIGINT more than once or twice. I was thinking those came out of LB (or a dep)

boneskull:
but fwiw I wouldn’t recommend bumping it either, unless we can show that it’s _not_ a leak and they won’t grow further

raymondfeng:
Close to favor https://github.com/strongloop/loopback-next/pull/5020

PR #5014, opened 2020-04-02
refactor: use sandbox.path instead of SANDBOX to access the sandbox directory

raymondfeng:
Step 1 for #5011 

Refactor test sandbox usage to use `sandbox.path` to access the sandbox directory as we'll create unique subdirs for different sandbox instances.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5016, opened 2020-04-02
feat: improve test sandbox

raymondfeng:
Step 3 of #5011 
Depends on #5014 (included here as the 1st commit so that CI can pass).

## Checklist

### DIFF
--- a/examples/file-transfer/src/__tests__/acceptance/test-helper.ts
+++ b/examples/file-transfer/src/__tests__/acceptance/test-helper.ts
@@ -42,6 +42,7 @@ export interface AppWithClient {
 }
 
 export function getSandbox() {
-  const sandbox = new TestSandbox(path.resolve(__dirname, '../../../.sandbox'));
+  // dist/.sandbox/<a unique temporary subdir>
+  const sandbox = new TestSandbox(path.resolve(__dirname, '../../.sandbox'));
   return sandbox;
 }
--- a/packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
@@ -10,7 +10,11 @@ import {BooterApp} from '../fixtures/application';
 
 describe('application metadata booter acceptance tests', () => {
   let app: BooterApp;
-  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'));
+  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'), {
+    // We intentionally use this flag so that `dist/application.js` can keep
+    // its relative path to satisfy import statements
+    subdir: false,
+  });
 
   beforeEach('reset sandbox', () => sandbox.reset());
   beforeEach(getApp);
--- a/packages/testlab/README.md
+++ b/packages/testlab/README.md
@@ -16,6 +16,7 @@ Test utilities to help writing LoopBack 4 tests:
 - Helpers for creating `supertest` clients for LoopBack applications
 - HTTP request/response stubs for writing tests without a listening HTTP server
 - Swagger/OpenAPI spec validation
+- Test sandbox
 
 ## Installation
 
@@ -55,8 +56,9 @@ Table of contents:
 - [httpGetAsync](#httpgetasync) - Async wrapper for HTTP GET requests.
 - [httpsGetAsync](#httpsgetasync) - Async wrapper for HTTPS GET requests.
 - [toJSON](#toJSON) - A helper to obtain JSON data representing a given object.
-- [createUnexpectedHttpErrorLogger](#createUnexpectedHttpErrorLogger) - An error
+- [createUnexpectedHttpErrorLogger](#createunexpectedhttprrrorlogger) - An error
   logger that only logs errors for unexpected HTTP statuses.
+- [TestSandbox](#testsandbox) - A sandbox directory for tests
 
 ### `expect`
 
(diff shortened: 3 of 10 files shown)


examples/file-transfer/src/__tests__/acceptance/test-helper.ts
```diff
@@ -40,6 +40,6 @@ export interface AppWithClient {
 
 const SANDBOX = path.resolve(__dirname, '../../../.sandbox');
 export function getSandbox() {
-  const sandbox = new TestSandbox(SANDBOX);
+  const sandbox = new TestSandbox(SANDBOX, '');
```
bajtos, line 43: What is the reasoning for using an empty subdir string, i.e. using the same sandbox dir from all (child) processes? Did you mean to use the default auto-generated subdir instead?

```suggestion
  const sandbox = new TestSandbox(SANDBOX);
```

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
```diff
@@ -11,7 +11,7 @@ import {BooterApp} from '../fixtures/application';
 describe('application metadata booter acceptance tests', () => {
   let app: BooterApp;
   const SANDBOX_PATH = resolve(__dirname, '../../.sandbox');
-  const sandbox = new TestSandbox(SANDBOX_PATH);
+  const sandbox = new TestSandbox(SANDBOX_PATH, '');
```
bajtos, line 14: Ditto.

packages/testlab/src/test-sandbox.ts
```diff
@@ -37,11 +38,16 @@ export class TestSandbox {
...
-   * @param path - Path of the TestSandbox. If relative (it will be resolved relative to cwd()).
+   * @param rootPath - Path of the TestSandbox. If relative it will be resolved
+   * against the current directory.
+   * @param subdir - Subdirectory for the sandbox. If not provided, the sandbox
+   * will automatically generate a uuid as the name of the subdirectory.
    */
-  constructor(path: string) {
+  constructor(rootPath: string, subdir?: string) {
+    // Create a unique subdirectory so that sandboxes can be used in parallel
+    subdir = subdir ?? v1();
```
bajtos, line 48: I have two comments:

1. I believe the `v1` algorithm is considered as less reliable than `v4`. UUID generation is being discussed as a new JavaScript features and they proposal is to use `v4` algorithm only (or at least as the default). Are there any reasons for preferring v1 over v4 for our sandbox?

2. Do we actually need full UUID semantics? If we want to achieve isolation between (child) processes, then `process.pid` is a unique value that's much shorter than a uuid and does not require any new dependencies.

packages/testlab/src/test-sandbox.ts
```diff
@@ -37,11 +38,16 @@ export class TestSandbox {
...
     // resolve ensures path is absolute / makes it absolute (relative to cwd())
-    this._path = resolve(path);
+    this._path = resolve(join(rootPath, subdir));
```
bajtos, line 50: Please note that `path.resolves` accepts an array of path segments, there is no need to call `path.join`. See https://nodejs.org/api/path.html#path_path_resolve_paths:

> The `path.resolve()` method resolves a sequence of paths or path segments into an absolute path.

```suggestion
    this._path = resolve(rootPath, subdir);
```

packages/testlab/src/test-sandbox.ts
```diff
@@ -37,11 +38,16 @@ export class TestSandbox {
    * Will create a directory if it doesn't already exist. If it exists, you
    * still get an instance of the TestSandbox.
    *
-   * @param path - Path of the TestSandbox. If relative (it will be resolved relative to cwd()).
+   * @param rootPath - Path of the TestSandbox. If relative it will be resolved
+   * against the current directory.
+   * @param subdir - Subdirectory for the sandbox. If not provided, the sandbox
+   * will automatically generate a uuid as the name of the subdirectory.
```
bajtos, line 44: Please mention that it's possible to set `subdir` to an empty string to disable the new subdir behavior.

Maybe add the examples from the commit message?

```ts
// Create a sandbox as a subdirectory named with a uuid under the rootPath
const sandbox = new TestSandbox(rootPath);

// Create a sandbox in the root path directly
// This is same as the old behavior
const sandbox = new TestSandbox(rootPath, '');
const sandbox = new TestSandbox(rootPath, '.');

// Create a sandbox in the `test1` subdirectory of the root path
const sandbox = new TestSandbox(rootPath, 'test1');
```

packages/testlab/src/test-sandbox.ts
raymondfeng: I'll switch to `v4`.

If mocha decides to use worker threads for parallel testing, process.pid won't be unique.

raymondfeng: I switched to https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options and removed uuid dependency.

raymondfeng:
> As usual, the PR would be easier to review if the testlab change was made in a standalone PR.

Please ignore the 1st commit. As I pointed out in the PR, it depends on #5014. After 5014 is landed, I'll rebase to remove the commit.

packages/testlab/src/test-sandbox.ts
bajtos: 👏 I love that we can avoid adding a new dependency and create short subdirectory names - this will make it easier to inspect sandbox content when troubleshooting tests.

bajtos:
> Please ignore the 1st commit. As I pointed out in the PR, it depends on #5014. After 5014 is landed, I'll rebase to remove the commit.

👍 

I'll defer the final review until #5014 is landed and this PR rebased.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
```diff
@@ -10,7 +10,7 @@ import {BooterApp} from '../fixtures/application';
 
 describe('application metadata booter acceptance tests', () => {
   let app: BooterApp;
-  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'));
+  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'), '.');
```
bajtos, line 13: Can we please find a way how to use a unique subdirectory in this test too? What is preventing us from doing that now?

packages/testlab/src/test-sandbox.ts
```diff
@@ -37,19 +38,36 @@ export class TestSandbox {
...
+   * // Create a sandbox as a unique temporary subdirectory under the rootPath
+   * const sandbox = new TestSandbox(rootPath);
+   *
+   * // Create a sandbox in the root path directly
+   * // This is same as the old behavior
+   * const sandbox = new TestSandbox(rootPath, '');
+   * const sandbox = new TestSandbox(rootPath, '.');
+   *
+   * // Create a sandbox in the `test1` subdirectory of the root path
+   * const sandbox = new TestSandbox(rootPath, 'test1');
```
bajtos, line 52: I have mixed feeling about this API, I find it a bit difficult to understand the intent when reading code creating new sandboxes. I feel the reader needs a detailed knowledge of TestSandbox behavior to be able to understand what the second parameter means and how it changes sandbox behavior.

How about introducing a new `options` argument?

```ts
// Create a sandbox as a unique temporary subdirectory under the rootPath
const sandbox = new TestSandbox(rootPath);

const sandbox = new TestSandbox(rootPath, {
  unique: true,
});

// Create a sandbox in the root path directly
// This is same as the old behavior
const sandbox = new TestSandbox(rootPath, {
  unique: false,
});

const sandbox = new TestSandbox(rootPath, {
  subdir: '.', // implies unique: false
});

// Create a sandbox in the `test1` subdirectory of the root path
const sandbox = new TestSandbox(rootPath, {
  subdir: 'test1', // implies unique: false
});
```

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
raymondfeng: This one is tricky as we need to make sure `application.js` can still resolve its imports after being copied to the sandbox.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
raymondfeng: I explained it in the code comment.

packages/testlab/src/test-sandbox.ts
raymondfeng: Improved with `TestSandboxOptions` as `{subdir: boolean | string}`.

examples/file-transfer/src/__tests__/acceptance/test-helper.ts
```diff
@@ -42,6 +42,7 @@ export interface AppWithClient {
 }
 
 export function getSandbox() {
-  const sandbox = new TestSandbox(path.resolve(__dirname, '../../../.sandbox'));
+  // dist/.sandbox
```
bajtos, line 45: Can you please improve the comment and explain why it's ok to use the same directory `dist/.sandbox` when running the tests in parallel?

packages/testlab/README.md
```diff
@@ -372,6 +374,82 @@ describe('MyApp', () => {
 });
 ```
 
+### TestSandbox
+
+Many tests need use a temporary directory as the sandbox to mimic a tree of
+files. The `TestSandbox` class provides such facilities to create and manage a
+sandbox on the file system.
```
bajtos, line 381: +1 for adding docs for TestSandbox 👏

examples/file-transfer/src/__tests__/acceptance/test-helper.ts
raymondfeng: Sorry for the confusion. We do use a unique subdir now. I fixed the comment.

PR #5020, opened 2020-04-02
feat(core): improve application signal handler registration

raymondfeng:
Key changes:
- only register shutdown listeners when the app starts
  - it does not make sense to register shutdown listener before the app starts
  - starts to use system resources in the constructor is not a good idea
- unregister process listeners upon explicit stop
- fix the test case to clean up process listeners
- improve debugging with context name prefix

Replace #5015 

## Checklist

### DIFF
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -3,9 +3,9 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import debugFactory from 'debug';
+import debugFactory, {Debugger} from 'debug';
 import {EventEmitter} from 'events';
-import {v1 as uuidv1} from 'uuid';
+import {v4 as uuidv4} from 'uuid';
 import {Binding, BindingInspectOptions, BindingTag} from './binding';
 import {
   ConfigurationResolver,
@@ -40,8 +40,6 @@ import {
   ValueOrPromise,
 } from './value-promise';
 
-const debug = debugFactory('loopback:context');
-
 /**
  * Context provides an implementation of Inversion of Control (IoC) container
  */
@@ -75,6 +73,21 @@ export class Context extends EventEmitter {
    * Configuration resolver
    */
   protected configResolver: ConfigurationResolver;
+
+  /**
+   * A debug function which can be overridden by subclasses.
+   *
+   * @example
+   * ```ts
+   * import debugFactory from 'debug';
+   * const debug = debugFactory('loopback:context:application');
+   * export class Application extends Context {
+   *   super('application');
+   *   this._debug = debug;
+   * }
+   * ```
+   */
+  protected _debug: Debugger;
 
   /**
    * Create a new context.
@@ -113,13 +126,34 @@ export class Context extends EventEmitter {
     this.name = name ?? this.generateName();
     this.tagIndexer = new ContextTagIndexer(this);
     this.subscriptionManager = new ContextSubscriptionManager(this);
+    this._debug = debugFactory(this.getDebugNamespace());
+  }
+
+  /**
+   * Get the debug namespace for the context class. Subclasses can override
+   * this method to supply its own namespace.
+   *
+   * @example
+   * ```ts
+   * export class Application extends Context {
+   *   super('application');
+   * }
+   *
+   * protected getDebugNamespace() {
+   *   return 'loopback:context:application';
+   * }
+   * ```
+   */
+  protected getDebugNamespace() {
+    if (this.constructor === Context) return 'loopback:context';
+    const name = this.constructor.name.toLowerCase();
+    return `loopback:context:${name}`;
   }
 
   private generateName() {
-    const id = uuidv1();
-    let prefix = `${this.constructor.name}-`;
-    if (prefix === 'Context-') prefix = '';
-    return `${prefix}${id}`;
+    const id = uuidv4();
+    if (this.constructor === Context) return id;
+    return `${this.constructor.name}-${id}`;
   }
 
   /**
@@ -135,14 +169,14 @@ export class Context extends EventEmitter {
    * as the prefix
    * @param args - Arguments for the debug
    */
-  private _debug(...args: unknown[]) {
+  protected debug(...args: unknown[]) {
     /* istanbul ignore if */
-    if (!debug.enabled) return;
+    if (!this._debug.enabled) return;
     const formatter = args.shift();
     if (typeof formatter === 'string') {
-      debug(`[%s] ${formatter}`, this.name, ...args);
+      this._debug(`[%s] ${formatter}`, this.name, ...args);
     } else {
-      debug('[%s] ', this.name, formatter, ...args);
+      this._debug('[%s] ', this.name, formatter, ...args);
     }
   }
 
@@ -184,7 +218,7 @@ export class Context extends EventEmitter {
    */
   add(binding: Binding<unknown>): this {
     const key = binding.key;
-    this._debug('[%s] Adding binding: %s', key);
+    this.debug('[%s] Adding binding: %s', key);
     let existingBinding: Binding | undefined;
     const keyExists = this.registry.has(key);
     if (keyExists) {
@@ -261,14 +295,14 @@ export class Context extends EventEmitter {
         },
       );
       if (configResolver) {
-        debug(
+        this.debug(
           'Custom ConfigurationResolver is loaded from %s.',
           ContextBindings.CONFIGURATION_RESOLVER.toString(),
         );
         this.configResolver = configResolver;
       } else {
         // Fallback to DefaultConfigurationResolver
-        debug('DefaultConfigurationResolver is used.');
+        this.debug('DefaultConfigurationResolver is used.');
         this.configResolver = new DefaultConfigurationResolver(this);
       }
     }
@@ -340,7 +374,7 @@ export class Context extends EventEmitter {
    * @returns true if the binding key is found and removed from this context
    */
   unbind(key: BindingAddress): boolean {
-    this._debug('Unbind %s', key);
+    this.debug('Unbind %s', key);
     key = BindingKey.validate(key);
     const binding = this.registry.get(key);
     // If not found, return `false`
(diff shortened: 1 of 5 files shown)


packages/core/src/__tests__/unit/application.unit.ts
```diff
@@ -16,12 +16,14 @@ import {expect} from '@loopback/testlab';
 import {Application, Component, CoreBindings, CoreTags, Server} from '../..';
 
 describe('Application', () => {
+  let app: Application;
+
+  beforeEach(givenApp);
```
bajtos, line 21: I am not sure if it's a good idea to always create the same application for all tests. What if we need to call the app constructor with different arguments in a test we will write in a future?

A more robust & future-proof solution is to keep the `app` variable and `afterEach` hooks here (shared by all tests), but keep `beforeEach(givenApp)` inside every `describe` block as it is now.

packages/core/src/__tests__/unit/application.unit.ts
```diff
@@ -305,15 +285,63 @@ describe('Application', () => {
...
+    it('registers/removes a SIGTERM listener by start/stop', async () => {
+      await app.start();
+      const count = getListeners().length;
+      await app.stop();
+      expect(getListeners().length).to.eql(count - 1);
+      // Restart
+      await app.start();
+      expect(getListeners().length).to.eql(count);
+    });
+
```
bajtos, line 312: Please add a test to verify that `new Application()` does not register any new listeners.

packages/core/src/application.ts
```diff
@@ -83,7 +87,8 @@ export class Application extends Context implements LifeCycleObserver {
   constructor(configOrParent?: ApplicationConfig | Context, parent?: Context) {
     super(
       configOrParent instanceof Context ? configOrParent : parent,
-      'application',
+      // Default the context name to application name
+      configOrParent instanceof Context ? undefined : configOrParent?.name,
```
bajtos, line 91: What if the application name is not provided? IIUC, before your change, we would use `application` as the context name. With your change in place, we will use an auto-generated UUID instead.

Did you perhaps mean this?

```suggestion
      configOrParent instanceof Context ? 'application' : configOrParent?.name,
```

Also what if `name` is not set? I think we should use `application` in that case too.

TBH, I find these repeated `instanceof` checks a bit difficult to comprehend, let's normalize the arguments first before we start processing them.

```ts
export class Application extends Context implements LifeCycleObserver {
  //...

  constructor(configOrParent?: ApplicationConfig | Context, parent?: Context) {
    let config: ApplicationConfig;
    if (configOrParent instanceof Context) {
      parent = configOrParent;
      config = undefined;
    }

    config = config ?? {};

    super(parent, config.name ?? 'application');
    this.options = config;
  
    //...
  }

  //...
}
```

packages/core/src/application.ts
```diff
@@ -98,11 +103,23 @@ export class Application extends Context implements LifeCycleObserver {
...
-    );
+    this._shutdownOptions = {signals: ['SIGTERM'], ...this.options.shutdown};
+  }
+
+  /**
+   * Wrap the debug statement so that it always print out the context name
+   * as the prefix
+   * @param args - Arguments for the debug
+   */
+  private debugWithName(...args: unknown[]) {
```
bajtos, line 114: Should the Context class prefix debug logs with the context name too?

packages/core/src/__tests__/unit/application.unit.ts
```diff
@@ -305,15 +285,63 @@ describe('Application', () => {
       expect(binding.key).to.equal('services.my-service');
       expect(findKeysByTag(app, 'service')).to.containEql(binding.key);
     });
+  });
 
-    function givenApp() {
-      app = new Application();
+  describe('shutdown signal listener', () => {
+    it('registers a SIGTERM listener when app starts', async () => {
+      const count = getListeners().length;
```
emonddr, line 292: Any reason we aren't checking with actual values like 0 or 1?  Is the getListeners() going to return a variable amount depending on a system option or something?

packages/core/src/application.ts
raymondfeng: Without a name, it will become `Application-<uuid>`.

packages/core/src/__tests__/unit/application.unit.ts
raymondfeng: `process` is a global. We cannot assert the listener count is `0` or `1`.

packages/core/src/application.ts
raymondfeng: I made the context.debugWithName() protected.

packages/context/src/context.ts
```diff
@@ -135,14 +140,15 @@ export class Context extends EventEmitter {
    * as the prefix
    * @param args - Arguments for the debug
    */
-  private _debug(...args: unknown[]) {
+  protected debugWithName(...args: unknown[]) {
```
bajtos, line 143: It makes me wonder, is it necessary to mention the fact that this method is prepending the context name in debug messages? Can we simply call it `debug`?

packages/context/src/context.ts
```diff
@@ -135,14 +140,15 @@ export class Context extends EventEmitter {
    * as the prefix
    * @param args - Arguments for the debug
```
bajtos, line 141: What happens when the child class calls this method but does not provide a custom `_debug` logger? I believe your implementation will use the `debug` function initialized with the scope `loopback:context` (see line 43 above). As a result, debug logs printed by a custom context subclass may use `loopback:context` scope, which may be surprising to users.

At minimum, I'd like this aspect to be clearly described in API docs.

I think a better solution is to find a design that will make it impossible to call debug with the default scope. For example, the base `Context` class can provide a helper function accepting a context instance and a debug logger and returning a debug function that includes the context name in all messages. Then each Context subclass can use that helper to create its own private debug logger.

```ts
export function createDebugWithContextDetails(ctx: Context, debug: Debugger) {
  return function debugWithName(...args: unknown[]) {
    // put your code here, replace `this.name` with `ctx.name`
    // in the future, we can add more details in addition to the name
  }
 }

// inside Context
class Context {
  private _debug = createDebugWithContextDetails(this, debug);
  // ...
}

// in the core
import {createDebugWithContextDetails} from '@loopback/context';

class Application {
  private _debug = createDebugWithContextDetails(this, debug);
  // ...
}
```

If you prefer, the helper can be a static method on the Context class. I am sure we can find a better name too.

packages/core/src/__tests__/unit/application.unit.ts
```diff
@@ -16,12 +16,15 @@ import {expect} from '@loopback/testlab';
 import {Application, Component, CoreBindings, CoreTags, Server} from '../..';
 
 describe('Application', () => {
-  describe('controller binding', () => {
-    let app: Application;
-    class MyController {}
+  let app: Application;
+
+  afterEach('clean up application', () => app.stop());
```
bajtos, line 21: 👍

packages/core/src/application.ts
```diff
@@ -446,8 +483,51 @@ export class Application extends Context implements LifeCycleObserver {
...
+            this.name,
+            process.listenerCount(sig),
+            sig,
+            process.pid,
+            new Error('MaxListenersExceededWarning'),
+          );
+        }
+      }
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      process.removeListener(sig, this._signalListener);
```
bajtos, line 512: I am confused, why are we removing the listener and then adding it again on the next line? Isn't it enough to addd it?

packages/core/src/application.ts
```diff
@@ -470,6 +550,10 @@ export type ShutdownOptions = {
  * Configuration for application
  */
 export interface ApplicationConfig {
+  /**
+   * Name of the application context
+   */
+  name?: string;
```
bajtos, line 556: Is this really needed? Can we use the application class name so that no configuration is necessary?

packages/context/src/context.ts
raymondfeng: Renamed to `debug`.

packages/context/src/context.ts
raymondfeng: We now use the subclass name to build the debug namespace by default. It can be also overridden with `_debug` or `getDebugNamespace`.

packages/core/src/application.ts
raymondfeng: It's an optional property. By default, we use the application class name.

raymondfeng:
@bajtos PTAL.

packages/core/src/application.ts
raymondfeng: The initial intention was to make sure there is no duplicate registration of the same listener. I removed it now.

packages/context/src/context.ts
```diff
@@ -76,6 +74,11 @@ export class Context extends EventEmitter {
    */
   protected configResolver: ConfigurationResolver;
 
+  /**
+   * A debug function which can be overridden by subclasses
+   */
```
bajtos, line 79: Can you please expand the tsdoc comment with an example to show how to use this new property in practice? E.g.

```ts
import debugFactory from 'debug';
const debug = debugFactory('loopback:core:application');

class Application extends Context {
  constructor(parent: Context) {
    super(parent);
    this._debug = debug;
  }
}
```

packages/context/src/context.ts
```diff
@@ -3,9 +3,9 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import debugFactory from 'debug';
+import debugFactory, {Debugger} from 'debug';
```
emonddr, line 6: > import debugFactory, {Debugger} from 'debug';

why not

import {debugFactory, Debugger} from 'debug';

raymondfeng: Please note `debug` module exports a function such as:

```js
module.exports = function debugFactory(...) {}
```

And `Debugger` is a member of the exports.

The TS imports reflect plain JS usage as follows:

```js
const debugFactory = require('debug'); // The module.exports is a function 
require('debug').Debugger; // debug is used as a namespace
```

PR #5021, opened 2020-04-02
fix(context): fix context subscription manager to not emit errors to its context if there is an error listener is in place

raymondfeng:
The unhandled error is reported when context-observer.unit.ts is run by itself. The root cause is that ContextSubscriptionManager always emits errors to the context chain without checking if there is an error listener on itself.

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/context-observer.unit.ts
+++ b/packages/context/src/__tests__/unit/context-observer.unit.ts
@@ -173,7 +173,7 @@ describe('Context', () => {
     it('reports error if an observer fails', () => {
       ctx.bind('bar').to('bar-value');
       return expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
-        'something wrong',
+        'something wrong - async',
       );
     });
 
@@ -239,7 +239,7 @@ describe('Context', () => {
         filter: binding => binding.key === 'bar',
         observe: async () => {
           await setImmediateAsync();
-          throw new Error('something wrong');
+          throw new Error('something wrong - async');
         },
       };
       ctx.subscribe(nonMatchingObserver);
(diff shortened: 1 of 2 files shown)


raymondfeng:
FYI: The PR fixed an unhandled error in the test. As a result, the test coverage goes down a bit, which is expected as the test does not crash or warn unhandled promise rejections.

packages/context/src/__tests__/unit/context-observer.unit.ts
```diff
@@ -170,11 +170,14 @@ describe('Context', () => {
...
-    it('reports error if an observer fails', () => {
+    it('reports error if an observer fails', async () => {
       ctx.bind('bar').to('bar-value');
-      return expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
+      const errorPromise = pEvent(ctx, 'error', {rejectionEvents: []});
+      await expect(ctx.waitUntilObserversNotified()).to.be.rejectedWith(
         'something wrong',
       );
+      const error = await errorPromise;
+      expect(error.message).to.eql('something wrong');
```
bajtos, line 180: I find it difficult to understand what's going on here. Are you saying that the scenario executed by this test is reporting the error twice, once via `waitUntilObserversNotified` and second time via `ctx.on('error')`?

Shouldn't it be the responsibility of `waitUntilObserversNotified` to ensure `ctx.on('error')` does not trigger unhandled error, e.g. by installing a one-time event listener for the error event?

As I understand the issue, the current implementation of `waitUntilObserversNotified` is difficult to use correctly and as a result, the test that looks correct on the first sight has a subtle problem that you are fixing by this patch.

I am concerned that this may not be the last place using `waitUntilObserversNotified` incorrectly and prefer to improve `waitUntilObserversNotified` so that it's easier to use this feature correctly. Think of [Falling into the pit of succcess](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

PR #5031, opened 2020-04-03
fix(http-server): ensure rejection handler is registered before the error event is emitted

raymondfeng:
## Checklist


packages/http-server/src/__tests__/integration/http-server.integration.ts
```diff
@@ -95,13 +95,15 @@ describe('HttpServer (integration)', () => {
     const req = httpGetAsync(server.url, agent);
     // Wait until the request is accepted by the server
     await pEvent(server.server, 'request');
+    // Set up error handler for expected rejection before the event is emitted
```
raymondfeng, line 98: Please note `pEvent` uses `server.server.on('error', ...)` behind the scene. We need to call it first so that an error handler is registered before the error event is emitted.

PR #5041, opened 2020-04-06
feat: upgrade to TypeScript 3.9.x (typescript@3.9.2)

raymondfeng:
This PR fixes our code to be compatible with upcoming TypeScript 3.9.x (using typescript@~3.9.0-beta).

## Checklist


packages/rest/src/rest.server.ts
```diff
@@ -297,10 +297,19 @@ export class RestServer extends Context implements Server, HttpServerLike {
     return handleUnExpectedError;
   }
 
+  /**
+   * Non-null assertion type guide
+   * @param val - A value can be `undefined` or `null`
+   */
+  private assert<T>(val: T | null | undefined): asserts val is T {
```
bajtos, line 304: Is there any particular reason for making this function a class member? AFAICT, it does not need anything from the RestServer class. Also the name `assert` is not very intent-revealing, have you considered something like `assertExists`?

```ts
function assertExists<T>(val: T | null | undefined): asserts val is T {
  assert(val);
}
```

packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
```diff
@@ -171,7 +171,7 @@ describe('RestServer', () => {
         }
 
         get expressApp() {
-          return this._expressApp;
+          return this._expressApp!;
```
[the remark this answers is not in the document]
bajtos: Alternatively, let's use the new assertion you introduced in the server class at https://github.com/strongloop/loopback-next/pull/5041/files#diff-576584d72083bead0929c10dfc101b01R304

packages/testlab/src/http-server-config.ts
```diff
@@ -38,19 +33,19 @@ export type HostPort = {
...
   };
-  const config = Object.assign({}, defaults, customConfig);
+  const config = {...defaults, ...customConfig};
   if (config.host === undefined) config.host = defaults.host;
   if (config.port === undefined) config.port = defaults.port;
   if (isHttpsConfig(config)) {
     setupTlsConfig(config);
   }
-  return config;
+  return config as T;
```
bajtos, line 48: At the first sight, your change seems to be towards worse. 
- At the moment, the return type is described as having host, port and any custom config. No explicit cast is needed when returning the merged config.
- With your changes in place, `HostPort` part is gone - are you sure `T` will always include those two fields? Also an explicit cast is needed here.

Let's try to find a cleaner solution that will not require explicit casts.

examples/todo-list/src/__tests__/acceptance/todo-list-image.acceptance.ts
```diff
@@ -55,7 +55,8 @@ describe('TodoListApplication', () => {
     expect(response.body).to.containEql(expected);
 
     const created = await todoListImageRepo.findById(response.body.id);
-    expect(toJSON(created)).to.deepEqual({id: response.body.id, ...expected});
+    expected.id = response.body.id;
+    expect(toJSON(created)).to.deepEqual(expected);
```
bajtos, line 59: Can you please explain what error is TypeScript 3.9 reporting here? I'd like to better understand what's going on and preferably find a way how to avoid mutation of `expected` object.

raymondfeng: TypeScript reports an error for `expect(toJSON(created)).to.deepEqual({id: response.body.id, ...expected});` with the following reasoning:

1. The inferred type of `expected` has an `id` (non-optional) property.
2. The spread assignment has overriding property `id` from the 1st and 2nd argument. TypeScript think the 1st `id` will be ALWAYS overridden and it is probably a bug in the code.
3. In our case, `expected.id` is not defined when it's instantiated from a model class without `create`.

raymondfeng:
@bajtos PTAL.

examples/todo-list/src/__tests__/acceptance/todo-list-image.acceptance.ts
```diff
@@ -51,10 +51,27 @@ describe('TodoListApplication', () => {
...
+     * with the following reasoning:
+     * 1. The inferred type of `expected` has an `id` (non-optional) property.
+     * 2. The spread assignment has overriding property `id` from the 1st and 2nd
+     * argument.
+     * 3. The TS compiler think the 1st `id` will be ALWAYS overridden and
+     * it is probably a bug in the code.
+     * 4. In our case, `expected.id` is not defined when it's instantiated from
+     * `TodoListImage` class without `create`. Please note the
+     * `TodoListImage.prototype.id` is NOT marked as optional.
+     */
```
bajtos, line 66: Have you considered modifying `givenTodoListImage` (and other similar helpers) to correctly describe the type of the return value?

- `Omit<TodoListImage, 'id'>` - I think that's the most correct one
- `Partial<TodoListImage>` - an easy solution that may be good enough

packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
```diff
@@ -171,10 +171,15 @@ describe('RestServer', () => {
         }
 
         get expressApp() {
+          assertExists(this._expressApp);
           return this._expressApp;
         }
       }
 
+      function assertExists<T>(val: T): asserts val is NonNullable<T> {
```
bajtos, line 179: :+1:

packages/rest/src/__tests__/unit/rest.server/rest.server.unit.ts
```diff
@@ -171,10 +171,15 @@ describe('RestServer', () => {
...
+        expect(val != null).to.be.true();
```
bajtos, line 180: I am concerned that when this assertion fail, it will print a message that's not very helpful. Have you considered using this?

```ts
expect.exists(val);
```

See http://shouldjs.github.io/#should-exist

```sh
$ node
> expect.exists(null)
Uncaught [Error [AssertionError]: expected null to exist] {
  stackStartFunction: [Function (anonymous)]
}
> expect.exists(undefined)
Uncaught [Error [AssertionError]: expected undefined to exist] {
  stackStartFunction: [Function (anonymous)]
}
```

PR #5046, opened 2020-04-06
feat: add jwt auth extension

jannyHou:
Implements https://github.com/strongloop/loopback-next/issues/4903
Extract the jwt authentication component into a standalone extension.

_Considering the user service relies on user + userCredential model and their presistency layer(repository+datasource), I decide to leave them out of the component._

## Checklist


extensions/authentication-jwt/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2019. All Rights Reserved.
+Node module: @loopback/extension-health
```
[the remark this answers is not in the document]
raymondfeng: Next time, try to use https://github.com/strongloop/loopback-next/pull/4984. It will take care of all the project configuration/file headers.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,59 @@
...
+// merge the security spec into the application.
+// This will be improved with a coming enhancer.
+// See section [To Be Done](#to-be-done)
+```
+
+_All the jwt authentication related code are marked with comment "- enable jwt
+auth -", you can search for it to find all the related code you need to enable
+the entire jwt authentication in a LoopBack 4 application._
+
+## To Be Done
```
dhmlau, line 37: `Future work`?

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,59 @@
+# @loopback/extension-authentication-jwt
+
+This module exports the jwt authentication strategy and its corresponding token
+service as a component. You can mount the component to get a prototype token
+based authentication system in your LoopBack 4 application.
+
+## Usage
+
+This module contains an example LoopBack 4 application in folder `fixtures`. It
```
dhmlau, line 9: As a user, I think I would be more interested to see how I can use this extension from ground zero, e.g. I have a todo application, how this extension can be used.  If User and UserCredentials are the pre-requisite for using this extension, it might be good to include there. 

I think it's better to include the models, so that users only need to have minimal setup in order to use this extension. But at the same time, we have the flexibility to extend the models for customization. Is it possible?

jannyHou: @dhmlau Good question. The user service requires **User and UserCredential's persistence**. Therefore including the user service in the component means the corresponding repositories and datasource are also needed:
- User model
- UserCredential model
- User repository
- UserCredential repository
- DbDatasource

But usually a LoopBack app has its own datasource, that's why I find it hard to include the user service.

> But at the same time, we have the flexibility to extend the models for customization

It's not only about the models, but repositories and datasource are also involved...WTDY?

dhmlau: Can we abstract out the implementation of the UserService and TokenService? For example, if the UserService is actually calling out other service to verify the credentials rather than validating against a database entry.

jannyHou: @dhmlau I extracted the repository and model into component.
Now the app can mount the component then inject datasource and call its services in controller 👏

mschnee:
From an application security perspective, a JWT Bearer token is usually (but not always) just one half of the Bearer-token authentication strategy.  The other half is the refresh token.  Generally, access tokens have a very short life, and refresh tokens have a longer one.

Will this extension account for a refresh token as well, or is it specific to the bearer authorization token?

jannyHou:
@mschnee 
>  The other half is the refresh token. Generally, access tokens have a very short life, and refresh tokens have a longer one.

Very helpful suggestion. This PR is a start of the extension, in the future we can support the refresh functionality. There is a demo PR for it in https://github.com/strongloop/loopback4-example-shopping/pull/537, and the token service interface has `revokeToken` function added in https://github.com/strongloop/loopback-next/pull/4746/files.

Would you like to create a new issue to track this feature(add token refresh to the jwt extension)?

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+    this.sequence(MySequence);
+
+    // Set up default home page
+    this.static('/', path.join(__dirname, '../public'));
+
+    // - enable jwt auth -
+    // Add security spec (Future work: refactor it to an enhancer)
+    this.addSecuritySpec();
+    // Mount authentication system
+    this.component(AuthenticationComponent);
```
hacksparrow, line 90: Should show where `AuthenticationComponent ` comes from.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+    // Mount jwt component
+    this.component(JWTAuthenticationComponent);
```
hacksparrow, line 92: Should show where `JWTAuthenticationComponent` comes from.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+  // See section [Future Work](#future-work)
+  addSecuritySpec(): void {
+    this.api({
+      openapi: '3.0.0',
+      info: {
+        title: 'test application',
+        version: '1.0.0',
+      },
+      paths: {},
+      components: {securitySchemes: SECURITY_SCHEME_SPEC},
```
hacksparrow, line 115: Should show  the definition of `SECURITY_SCHEME_SPEC`.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+</details>
+
+_All the jwt authentication related code are marked with comment "- enable jwt
+auth -", you can search for it to find all the related code you need to enable
+the entire jwt authentication in a LoopBack 4 application._
+
+## Add Endpoint in Controller
+
+After mounting the component, you can call token and user services to perform
+login then decorate endpoints with `@authentication('jwt')` to inject the logged
```
hacksparrow, line 138: `login then` -> `login, then`?

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+in user's profile.
+
+This module contains an example LoopBack 4 application in folder `fixtures`. It
```
hacksparrow, line 141: "This module contains an example LoopBack 4 application in folder `fixtures`." -> "This module contains an example application in the `fixtures` folder."?

The term "LoopBack 4 " is redundant in the context of the extension.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+has a controller with endpoints `/login` and `/whoAmI` for demo's purpose.
```
hacksparrow, line 142: "has a controller with endpoints `/login` and `/whoAmI` for demo's purpose." -> "has a controller with endpoints `/login` and `/whoAmI`."?

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+The code snippet for login function:
+
+<details>
```
hacksparrow, line 146: Sometimes it's better to show the code by default, especially if its not a lot. I'd suggest to remove `<details>`.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,209 @@
...
+    return {token};
+  }
+```
+
+</p>
+</details>
+
+The code snippet for whoAmI function:
+
+<details>
```
hacksparrow, line 172: Sometimes it's better to show the code by default, especially if its not a lot. I'd suggest to remove `<details>`.

hacksparrow:
@jannyHou this package is coming with a model definition of its own. How would users use the functionality but still use their own user model (which could be drastically different from the one that's included)?

In my opinion, the package should not contain any model or repositories, they should be provided to the package via dependency injection. For testing purposes they could be under the `fixtures` dir, though.

extensions/authentication-jwt/README.md
jannyHou: true, removed

jannyHou:
@hacksparrow Thank you for the review and bringing this question again, interesting:

> How would users use the functionality but still use their own user model (which could be drastically different from the one that's included)?

The user service file is coupled with user model/repository, therefore moving the user model out of the module would also remove the user services, which is not good.

My original implementation in [the 1st commit](https://github.com/strongloop/loopback-next/pull/5046/commits/3a9fdb72d605907888c0e606c349dd0a4da532ae) EXCLUDED the user service+model+repository:
- its component: [see component](https://github.com/strongloop/loopback-next/tree/3a9fdb72d605907888c0e606c349dd0a4da532ae/extensions/authentication-jwt/src), please not user service is not included in the services folder
- its fixture app: [see app](https://github.com/strongloop/loopback-next/tree/3a9fdb72d605907888c0e606c349dd0a4da532ae/extensions/authentication-jwt/src/__tests__/fixtures), it contains all the model related artifacts.

While after discussing with @dhmlau in https://github.com/strongloop/loopback-next/pull/5046#discussion_r404494235, and seeing a similar authorization extension created in https://github.com/ckoliber/loopback-acl-extension, we think it's better to include the user service+model+repository as prototype implementation, and only inject the datasource from application. It's also easy for people to quickly try the extension.

Developers can still provide their own User model, just like extending the default User model in lb3. And given the original version, it's not hard to replace the default one, I can write additional doc to show how to replace the User model. WDYT?

dhmlau:
@jannyHou, IIUC, I think @hacksparrow and I are on the same page. 

My views in https://github.com/strongloop/loopback-next/pull/5046#discussion_r404828047 and https://github.com/strongloop/loopback-next/pull/5046#discussion_r404494235 were to have minimum implementation of the TokenService and UserService as possible. But since i'm not familiar with the implementation, I wasn't sure what's possible and what's not. 

The bottomline is that we should be clear in README that what's required to use the extension and how to customize it. We could possibly provide the default implementation, and let the users customize.

hacksparrow:
@jannyHou please show users can use the extension with their own models and repositories, that'll give us a better insight into the DX.

jannyHou:
> The bottomline is that we should be clear in README that what's required to use the extension and how to customize it.

> please show how users can use the extension with their own models and repositories, that'll give us a better insight into the DX.

"what's required to use the extension" is already provided in the README.md file, I didn't get this part: "and how to customize it". Ok now I understand what to add. Thx.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,317 @@
...
+
+The complete file is in
+[user.controller.ts](https://github.com/strongloop/loopback-next/tree/master/extensions/authentication-jwt/src/__tests__/fixtures/controllers/user.controller.ts)
+
+## Customize Model
+
+1. Create your own user model and repository by `lb4 model` and
+   `lb4 repository`.
+
+2. The user service requires the user model, to provide your own one, you need
```
hacksparrow, line 186: It makes it sound like copying that file is the only way to do it. The instruction should be more generic.

Suggestion: _Create a corresponding custom `UserService` and bind it to `UserServiceBindings.USER_SERVICE`. Take a look at https://github.com/strongloop/loopback-next/blob/master/extensions/authentication-jwt/src/services/user.service.ts for an example of `UserService` implementation._

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,317 @@
...
+  }
+```
+
+The complete file is in
+[user.controller.ts](https://github.com/strongloop/loopback-next/tree/master/extensions/authentication-jwt/src/__tests__/fixtures/controllers/user.controller.ts)
+
+## Customize Model
+
+1. Create your own user model and repository by `lb4 model` and
+   `lb4 repository`.
```
hacksparrow, line 184: -> by running the `lb4 model` and `lb4 repository` commands.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,317 @@
...
+@authenticate('jwt')
+  async whoAmI(): Promise<string> {
+    return this.user[securityId];
+  }
+```
+
+The complete file is in
+[user.controller.ts](https://github.com/strongloop/loopback-next/tree/master/extensions/authentication-jwt/src/__tests__/fixtures/controllers/user.controller.ts)
+
+## Customize Model
```
hacksparrow, line 181: -> Customization

Since we are dealing with customization beyond just the model.

extensions/authentication-jwt/README.md
```diff
@@ -0,0 +1,317 @@
...
+       this.component(JWTAuthenticationComponent);
+       // Bind datasource
+       this.dataSource(DbDataSource, UserServiceBindings.DATASOURCE_NAME);
+       // Bind user service
+       this.bind(UserServiceBindings.USER_SERVICE).toClass(CustomUserService),
+       // Bind user and credentials repository
+       this.bind(UserServiceBindings.USER_REPOSITORY).toClass(
+         UserRepository,
+       ),
+       this.bind(UserServiceBindings.USER_CREDENTIALS_REPOSITORY).toClass(
```
hacksparrow, line 288: Add a code comment that the value of `TokenServiceBindings` can be changed too if required. This is something many will definitely want to change.

jannyHou: @hacksparrow I added a new section called "customization" and it has "customizing model" as the sub section, see https://github.com/strongloop/loopback-next/pull/5046/files#diff-b7edcb305a3224321ad8aea0c9573fb8R181

jannyHou:
@raymondfeng I added a section explaining this is a prototype extension, developers can customize any element, see https://github.com/strongloop/loopback-next/blob/77f730f15843c487920fd2e2e09be2321615804d/extensions/authentication-jwt/README.md#customization, and "customizing user" is the sub-section of it.

raymondfeng:
> @raymondfeng I added a section explaining this is a prototype extension, developers can customize any element, see https://github.com/strongloop/loopback-next/blob/77f730f15843c487920fd2e2e09be2321615804d/extensions/authentication-jwt/README.md#customization, and "customizing user" is the sub-section of it.

That's good. I'm wondering if we can further improve it as follows.

- Add a simple diagram to show the dependencies of the component
- Make a disclaimer at the top to say the user service is a reference implementation - not recommended for production

jannyHou:
> Add a simple diagram to show the dependencies of the component
    Make a disclaimer at the top to say the user service is a reference implementation - not recommended for production

Sure, PR coming.

Issue #5056, opened 2020-04-07
Multi-tenancy and dynamic schema selection


bajtos:
Recently, several people asked about implementing dynamic schema selection to enable schema-base multi-tenancy, where tenant isolation is achieved via DDL Schemas. (If you are not familiar with DDL schemas then you can learn the basics e.g. in [PostgreSQL docs](https://www.postgresql.org/docs/current/ddl-schemas.html)).

I am opening this Epic to do discuss possible solutions, implement necessary improvements and document how to implement multi-tenancy, possibly including an example app.

bajtos:
I don't have bandwidth to look into this topic in full, but would like to dump few ideas & pointers to make it easier for others to do the research.

Conceptually, querying a database table involves the following actors:

1. Model describes the shape of the data (table columns) and metadata (table name, DDL schema name, etc.)
2. Repository binds a model with a datasource, registers the model with the datasource. At the moment, this involves converting LB4 model definition to LB3/juggler style.
3. Datasource represents a database client and maintains a pool of open database connections. The datasource is configured with a connector to use (PostgreSQL, MongoDB, etc.), connection settings (host/port, credentials) and most importantly the database name to use. Most (if not all) LoopBack connectors require each datasource to use only one database name, it's not possible to switch between databases at runtime.
4. The database server.

In a typical LB4 application, each model is associated with the same Repository class and the same DataSource instance - the wiring is static.

To enable multi-tenancy, we want to make this wiring dynamic. Depending on the current user, we want to use a different repository and/or datasource configuration.

## Lightweight tenant isolation using schemas

In this setup, the authentication layer and all tenants share the same database name and use the same credentials (database user) to access the data. We have 1+N schemas defined in the database: the first schema is used by the authentication layer, plus we have one schema for each tenant. All database queries will use the same LB datasource and thus share the same connection pool.

Implementation wise, we need to tweak the way how a LB4 model is registered with a datasource. Instead of creating the same backing juggler model for all users, we want to create tenant-specific juggler models.

Conceptually, this can be accomplished by tweaking the Repository constructor.

```ts
export class ProductRepository extends DefaultCrudRepository<
  Product,
  typeof Product.prototype.id
> {
  constructor(
    @inject('datasources.db') dataSource: juggler.DataSource,
    @inject(SecurityBindings.USER) currentUser: UserProfile,
  ) {
    super(
      // model constructor
      Product, 
      // datasource to use
      dataSource, 
[...]

PR #5087, opened 2020-04-10
feat: add an example for multi-tenancy

raymondfeng:
See #5056 

## Checklist


hacksparrow:
> We simply rebind `datasources.db` to a tenant specific datasource to select the
right datasource for `UserRepository`.

@raymondfeng  So, if I understand correctly, multitenancy is achieved by switching the datasource to the configured for the tenant.

At what point is the switch done? As soon as a request arrives? What if there is a long running query for a tenant and a new request arrives for another one? What if there was some code to be executed after query was a successful; is it guaranteed that this code will not be executed for another tenant?

raymondfeng:
> @raymondfeng So, if I understand correctly, multitenancy is achieved by switching the datasource to the one configured for the tenant.

> At what point is the switch done? As soon as a request arrives? What if there is a long running query for a tenant and a new request arrives for another one? What if there was some code to be executed after query was a successful; is it guaranteed that this code will not be executed for another tenant?

Please note that we create a new `RequestContext` per request. The switch happens at per request level. The req/res processing is fully isolated from each other with our dependency injections for stateful artifacts.

hacksparrow:
Yes, there's a new `RequestContext` per request, however, `datasources.db` is shared by all, isn't it?

@bajtos thoughts?

raymondfeng:
> Yes, there's a new RequestContext per request, however, datasources.db is shared by all, isn't it?

`datasources.db` by default a singleton with connection pool to the backend DB. It's bound at application level by `app.boot`. We are rebinding it to a new alias at `RequestContext` level so that there is a tenant specific DS for the given request.

```
request 1 with tenant `abc`:
abc: Request context (datasources.db -> datasources.db.abc: DB for abc)
         Application (datasources.db -> default DB)

request 2 with tenant `xyz`:
xyz: Request context (datasources.db -> datasources.db.xyz: DB for xyz)
         Application (datasources.db -> default DB)
```

Dainiu5:
Hi, great stuff! :+1:
From my understanding, in this current implementation one can switch between pre-defined data sources based on the tenant id. However, what could be done to switch between database schemas using a specific postgres datasource?
Any hint or suggestions how it could be done would be highly appreciated. 
Thanks!

Issue #5088, opened 2020-04-10
'title'-option of JsonSchemaOptions<T> causes wrong related class


iLem0n:
## Steps to reproduce

1. download minimum error example project: https://github.com/iLem0n/loopback-relation-error
2. `npm install && npm run start`
3. browse to http://localhost:3000/openapi.json or check the generated openapi-spec 
4. Check 'components -> schema -> ParentWithChildren -> Properties -> children -> items -> $ref'

## Current Behavior

the generated openapi3 spec has the parent class itself as the related class.
see: 'components -> schema -> ParentWithChildren -> Properties -> children -> items -> $ref'

```
"components": {
    "schemas": {
      "ParentWithChildren": {
        "title": "ParentWithChildren",
        "description": "(Schema options: { title: 'ParentWithChildren', includeRelations: true })",
        "properties": {
          "id": {
            "type": "string"
          },
          "name": {
            "type": "string"
          },
          "children": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/ParentWithChildren" // <--- expect "#/components/schemas/ParentWithChildren"
            }
          }
        },
        "required": [
          "name"
        ],
        "additionalProperties": false
      }
    }
  }
```

## Expected Behavior

the related class should be the class 'Children'
```
"components": {
    "schemas": {
      "ParentWithChildren": {
        "title": "ParentWithChildren",
        "description": "(Schema options: { title: 'ParentWithChildren', includeRelations: true })",
        "properties": {
          "id": {
            "type": "string"
          },
          "name": {
            "type": "string"
          },
          "children": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/ParentWithChildren" // <--- expect "#/components/schemas/ParentWithChildren"
            }
          }
        },
        "required": [
          "name"
        ],
        "additionalProperties": false
      }
    }
  }
```

## Link to reproduction sandbox

https://github.com/iLem0n/loopback-relation-error

## Additional information

  node -e 'console.log(process.platform, process.arch, process.versions.node)' --->
```
darwin x64 12.10.0
```

  npm ls --prod --depth 0 | grep loopback --->

```
├── @loopback/boot@2.0.3
├── @loopback/context@3.3.0
├── @loopback/core@2.2.1
├── @loopback/openapi-v3@3.1.2
├── @loopback/repository@2.1.0
├── @loopback/rest@3.2.0
├── @loopback/rest-explorer@2.0.3
[...]

PR #5108, opened 2020-04-13
fix: relation schemas should not inherit `title` from the source model's schema

jannyHou:
Fix https://github.com/strongloop/loopback-next/issues/5088

When building the relation schema, the `title` field should be omitted.
See details in comment https://github.com/strongloop/loopback-next/issues/5088#issuecomment-613069334

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
+++ b/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
@@ -4,6 +4,8 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {
+  Entity,
+  hasMany,
   Model,
   model,
   property,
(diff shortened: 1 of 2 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -501,7 +502,11 @@ export function modelToJsonSchema<T extends object>(
       result.properties = result.properties ?? {};
       const relMeta = meta.relations[r];
       const targetType = resolveType(relMeta.target);
-      const targetSchema = getJsonSchema(targetType, options);
+      // omit the title for related schema
+      const targetSchema = getJsonSchema(
+        targetType,
+        _.omit(options, ['title']),
```
raymondfeng, line 508: The `omit` function can be implemented as:

```ts
function omit(options: object, ...props: string[]} {
  const copy = {...options};
  for (const p of props) {
    delete copy[p];
  }
  return copy;
}

Then just call `omit(options, 'title')`

Comment (raymondfeng) [reply packages/repository-json-schema/src/build-schema.ts:508]: Let's not add `lodash` as a dependency unless necessary.

Comment (bajtos) [reply packages/repository-json-schema/src/build-schema.ts:508]: +1 to avoiding lodash dependency.

I think this line can be also rewritten as follows:

```suggestion
        {...options, title: undefined},
```

packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
```diff
@@ -246,6 +248,56 @@ describe('build-schema', () => {
...
+        color: {type: 'string'},
+      });
+      // No circular references in definitions
+      expect(schema.definitions).to.containEql({
+        ChildWithRelations: {
+          title: 'ChildWithRelations',
+          description: '(Schema options: { includeRelations: true })',
+          properties: {name: {type: 'string'}},
+          additionalProperties: false,
+        },
```
agnes512, line 297: LGTM. Just have a question. If `Parent` has many `Parent`, would the title here be `PatentWIthRelations`?

jannyHou: @agnes512 For self relations, the same schema is cached so that [line 419](https://github.com/strongloop/loopback-next/blob/344784be9d9b7624952db1f17473d5308ba2ae70/packages/repository-json-schema/src/build-schema.ts#L419) would return.

We also have a test case for it just above the new ones I added in this PR :)

jannyHou:
@raymondfeng @bajtos @agnes512 thank you for review, removed lodash 👍 

> I find the PR title misleading.
True. I changed it to "omit title in options when generating relation schemas"

packages/repository-json-schema/src/build-schema.ts
```diff
@@ -501,7 +501,14 @@ export function modelToJsonSchema<T extends object>(
       result.properties = result.properties ?? {};
       const relMeta = meta.relations[r];
       const targetType = resolveType(relMeta.target);
-      const targetSchema = getJsonSchema(targetType, options);
+
+      const targetOptions = {...options};
+      if (targetOptions.title) {
+        // Omit the title for related schema
```
bajtos, line 507: Perhaps add a code comment to explain why is this needed?

jannyHou:
> The idea is to describe the effect observed by LB users, not the implementation details (those should be explained in the commit/PR details).

Agreed 👍 renamed it to "relation schemas should not inherit `title` from the source model's schema", and also added comments to the code+tests.

Issue #2035, opened 2018-11-16
Register Express middleware per route, controller class or controller method


bajtos:
As a user migrating from Express, or simply wishing to leverage an existing Express middleware in my application, I would like to enable this middleware for certain routes, controller methods or entire controller classes.

Consider the following definition in an Express app:

```js
app.get('/api/users', 
  requireAuthenticated(),
  requireScope('read:users:all'), 
  withCacheControl({maxAge: '2 hours', private: true}),
  function getUsers(req, res, next) {
    // the route
  });
```

The code snippets below illustrate different ways of applying the middleware. Please note they are just examples, not a final proposal.

**Per route**

A possible LB4 solution (not very ergonomic, a better solution is needed):

```ts
app.route(new ExpressRoute(
  'get',
  '/api/users',
  {
    parameters: [
      {name: 'filter', in: 'query', schema: {/*...*/}}
    ],
    responses: { /*...*/ }
  },
  [
    requireAuthenticated(),
    requireScope('read:users:all'), 
    withCacheControl({maxAge: '2 hours', private: true}),
  ],
  function findUsers(filter: Filter) {
    return userRepo.find(filter);
  }
));
```

**Per controller method**

```ts
class UserController {
  @get('/api/users')
  @use(
    requireAuthenticated(),
    requireScope('read:users:all'), 
    withCacheControl({maxAge: '2 hours', private: true}),
  )
  find(
    @param.query.object('filter', {/*schema*/})
    filter: Filter
  ) {
    return userRepo.find(filter)
  )
}
```

**Per controller class**

This allows developers to apply the same set of middleware for multiple routes.

```ts
@use(requireAuthenticated())
class UserController {
  @get('/api/users')
  @use(requireScope('read:users:all'))
  find() {
    // ...
  )

  @del('/api/users')
  @use(requireScope('write:users:all'))
  delete() {
    // ...
  )
}
```

## Acceptance criteria

TBD

## Out of scope

- application-wide middleware, e.g. `app.use(passport.initialize())` - see #1293 and #1982
- error-handling middleware, e.g. `app.use((err, req, res, next) => {/*...*/})` - see #1293
- middleware serving content, e.g. `serve-static` - developers should use `app.static()` API.

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/1671, which can be possibly used as the plumbing.

sanadHaj:
@bajtos  @raymondfeng  @dhmlau 
[...]

PR #5118, opened 2020-04-16
feat(rest): enable Express middleware to be used as interceptors with LoopBack

raymondfeng:
Depends on https://github.com/strongloop/loopback-next/pull/5317
Implements #2035 
Supersedes #2434  

This PR adds ability to use Express middleware with LoopBack 4 as follows:

1. The default sequence now has `Middleware` step. It creates an invocation chain to call registered middleware handlers via extension point/extension pattern.
2. The sequence can be customized to have more than one `Middleware` steps.
3. Express middleware can also be wrapped as LB4 interceptors, which can in turn be added to global/class/method level.
4. Move built-in cors and openapi endpoints to middleware action.

restServer -> sequence (middleware, findRoute, ..., invoke, send) --> global middleware interceptor --> class/method level interceptor.

Some of the ideas are inspired by https://github.com/codejamninja/lb4-middleware. Thanks @codejamninja and @victor-enogwe

See docs at:

- https://github.com/strongloop/loopback-next/blob/middleware-interceptor/docs/site/Middleware.md.
- https://github.com/strongloop/loopback-next/blob/middleware-interceptor/docs/site/Express-middleware.md

middleware

## Checklist


raymondfeng:
See tests at https://github.com/strongloop/loopback-next/tree/middleware-interceptor/packages/rest/src/__tests__/acceptance/middleware.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+      this.reject(context, error);
+    }
+  }
+}
+```
+
+Now we can set up applications to leverage the new sequence:
+
+```ts
+// Create another middleware phase
```
hacksparrow, line 213: Which file is this? `src/index.ts`?

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+app
+  .bind('middleware.afterInvoke')
+  .toProvider(MiddlewareProvider)
+  .tag({[CoreTags.EXTENSION_POINT]: POST_INVOCATION_MIDDLEWARE});
+helper.app.sequence(SequenceWithMiddleware);
```
hacksparrow, line 218: What is `helper`? Where does it come from? Can't we just use `app.sequence(SequenceWithMiddleware)`?

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+   middleware: string | MiddlewareFactory<CFG>,
+   middlewareConfig?: CFG,
+   options: MiddlewareActionBindingOptions = {},
+ ): Binding<RequestInterceptor> {
+   // ...
+ }
+````
+
+```ts
+const binding = app.middleware(spy, spyConfig, {
```
hacksparrow, line 135: Would `app.middleware(mw, mwConfig, {` or `app.middleware(middleware, middleware Config, {` be more descriptive?

hacksparrow: Or include the definition of `spy` and `spyConfig` in the code.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+It's often desirable to reuse Express middleware in the sequence to handle API
+requests/responses without reinventing the world. We now add an `Middleware`
+action as the first step in the default sequence. The action itself is an
+interceptor chain of `RequestContext`. It uses the powerful
+[extension point/extension pattern](Extension-point-and-extensions.md) to
+contribute and discover registered Express middleware.
+
+#### Register an Express middleware for the sequence
+
+````ts
```
hacksparrow, line 111: Which file is this? Will users have to code this? If no, then we should have add some context to it.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+Alternatively, we can create a subclass of `MiddlewareInterceptorProvider`.
+
+### Use middleware as invocation interceptors
+
+With the ability to wrap Express middleware as LoopBack 4 interceptors, we can
+use the same programming model to register middleware as global interceptors or
+local interceptors denoted by `@intercept` decorators at class and method
+levels.
```
hacksparrow, line 95: Need an example showing how to do it.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Introduce middleware to REST sequence of actions
+
+There are a few steps involved in the default sequence of actions. See
+`Sequence.md` for more details.
```
hacksparrow, line 100: Will a html link be generated from `Sequence.md`?

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+programming model. We believe that we now have a good middle-ground to combine
+the best of breeds of both projects.
+
+The integration is built on top of the [interceptor](Interceptors.md) support as
+follows:
+
+We create a few helper functions that wraps an Express middleware module into an
+LoopBack 4 interceptor.
+
+### Create an interceptor from Express middleware module name or factory function
```
hacksparrow, line 31: Since this file will be a reference doc, we should prioritize on user API documentation. Including implementation details should be an added bonus, we could even omit it, IMO.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,229 @@
...
+   * Invokes registered middleware (injected via SequenceActions.MIDDLEWARE).
+   */
+  @inject(SequenceActions.MIDDLEWARE, {optional: true})
+  protected middleware?: Middleware;
+
+  // ...
+  async handle(context: RequestContext): Promise<void> {
+    try {
+      const {request, response} = context;
+      if (this.middleware != null) {
```
bajtos, line 159: Instead of using an `if` block here, can you set `middleware` to a no-op function in the constructor when no value is provided by dependency injection?

packages/rest/package.json
```diff
@@ -44,6 +44,7 @@
     "lodash": "^4.17.15",
     "on-finished": "^2.3.0",
     "path-to-regexp": "^6.1.0",
+    "p-event": "^4.1.0",
```
bajtos, line 47: Do we need to add `@types/p-event` to our dev dependencies?

packages/rest/src/middleware/middleware-interceptor.ts
```diff
@@ -0,0 +1,221 @@
...
+ * @typeParam CFG - Configuration type
+ * @typeParam CTX - Context type
+ */
+export function createInterceptor<CFG, CTX extends Context = InvocationContext>(
+  middleware: string | MiddlewareFactory<CFG>,
+  middlewareConfig?: CFG,
+): GenericInterceptor<CTX> {
+  let factory: MiddlewareFactory<CFG>;
+  if (typeof middleware === 'string') {
+    factory = require(middleware);
```
bajtos, line 43: Please note this `require` call will resolve the path relatively to the location of `@loopback/rest` module, not relatively to the code calling `createInterceptor`. This creates confusing behavior when `createInterceptor` is called e.g. from a 3rd-party package that's symlinked into `node_modules` tree.

We have been bitten by this problem multiple times in the past. While there are ways how to work around it, I feel the complexity is not worth the benefits.

Can you please remove the string-based variant and always require the factory function instead?

Also what if the middleware does not provide any factory function and all the users have is a regular function `(req, res, next) => void`? Are they supposed to wrap that function into a factory just for the sake of satisfying this API?

Last but not least, how do you envision caching the instances of middleware between requests? It's important to call middleware factory function only once at app startup, we must not call it for every incoming request. I don't yet fully comprehend how is `createInterceptor` going to be used, so maybe this is not a problem? IDK.

I also feel this function is violating Single Responsibility Principle, because it has two responsibilities:
- Obtain an instance of middleware handler.
- Convert a middleware handler to an interceptor.

Have you considering removing the first responsibility and modifying the function to accept the middleware handler only?

bajtos:
@raymondfeng this is a sizeable change with major design implications, can you please work with @dhmlau to create space in our plan to give us all enough time to review the proposal?

docs/site/Express-middleware.md
raymondfeng: yes.

packages/rest/src/middleware/middleware-interceptor.ts
raymondfeng: - Support for middleware module as `string` is removed.
- Default `InvokeMiddleware` is set to no-op.
- Caching is handled at binding scope level.

packages/rest/package.json
raymondfeng: No, `p-event@4.1.0` has self-contained `d.ts`.

docs/site/Express-middleware.md
raymondfeng: We can improve it once the apidocs is live.

raymondfeng:
FYI: I did further refactoring:

1. Added `@loopback/middleware` as the infrastructure to allow Koa style middleware and make it easy to plug in existing Express middleware as middleware or interceptors so that they can be used at sequence/action, global interceptor, and local interceptor tiers. I also see this module to be an entry point to appeal to Express users that need more extensibility and composability. The `ExpressApplication` and `ExpressServer` shows some of the possibilities.

2. Introduced a `InvokeMiddleware` action to the sequence as the extension point to discover registered middleware and invoke them as an interceptor chain.

3. Moved existing `cors` and `apiSpec` Express middleware into the `InvokeMiddleware` chain. Aligned `mountExpressRouter` with the middleware plumbing.

4. Updated todo example and code generation templates to use the new features.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,490 @@
...
+middleware logic to specific points of controller invocations, such as global,
+class, or method levels.
+
+## Bring Express middleware to LoopBack
+
+Express middleware can now be plugged into LoopBack as follows:
+
+- As a `Middleware` to be executed by `InvokeMiddleware` actions within the
+  sequence. Middleware are registered by extension points to form chains
+  (`MiddlewareChain').
```
bajtos, line 96:
```suggestion
  (`MiddlewareChain`).
```

bajtos:
In my view, this proposal is introducing a lot of complexity and many new moving parts. I am not convinced at all that we need this.

IMO, the following tool would be good enough for 80% of LoopBack users.

```ts
// usage
const compression = require('compression');
const helmet = require('helmet');

const middleware = [
  compression(),
  helmet.hidePoweredBy(),
  helmet.ieNoOpen(),
  helmet.noSniff(),
];

  constructor(
    // inject dependencies (skipped for brevity)
  ) {}

    try {
      if (await executeMiddlewareChain(context, middleware)) {
        // one of the middleware handled the request
        return;
      }
      this.send(response, result);
    } catch (error) {
      this.reject(context, error);
    }
  }
}

// implementation (mock-up)
async function executeMiddlewareChain({request, response}, chain) {
  const responseWritten = onFinishedAsync(response).then(() => true);
  const handlerFinished = new Promise<boolean>((resolve, reject) => {
    let ix = 0;
    next();

    function next(err) {
      if (err) return reject(err);
      if (ix <= chain.length) {
        // Express router/middleware called next, which means the request was not handled (yet)
        return resolve(false);
      }
      const hander = chain[ix++];
      handler(request, response, next);
    }
  });
  return Promise.race([handlerFinished, responseWritten]);
}
```

I see the following benefits of my solution:
1. The API is very easy to learn for people that are already familiar with Express.
2. Definition of which middleware to execute stays in the same place where we define other parts of request handling workflow (i.e. the sequence). As I see it, a middleware is just another sequence step.
3. It's just few lines of code, easy to debug when something goes wrong, easy to maintain for us.

>  Express middleware are one-way handlers that mostly process requests and there is no first-class facility to handle responses. We would love to support [Koa style middleware](https://github.com/koajs/koa/blob/master/docs/guide.md#writing-middleware) that use `async/await` to allow cascading behaviors.

I am not sure if we really want that. I believe we have decided very early on that we don't want to use the concept of a middleware and want to use Sequence and Sequence Actions instead. Unfortunately, we are not keeping documentation on the architecture decisions we made. There may be some reasoning captured in discussions in the early pull request and/or issues, IDK.

> Express middleware are added by the order of `app.use()` and it's hard to contribute middleware from other modules.

Based on my experience with LoopBack 2/3 and middleware phases, I don't think this is really an issue. I am not aware of any extension using the middleware phases feature, which was specifically introduced in LB2/3 to allow extensions to do this. On the opposite, I heard from users that they find middleware phases unnecessarily complicated and don't see any benefits.

I see two problems to solve:
1. How to leverage the vast ecosystem of existing Express middleware when building LB4 apps.
2. How to allow LB4 extensions to contribute middleware-like functionality.

I don't think it's a good idea to aim for a single solution to both of them.

1. As you have correctly pointed out, the design of Express middleware has its limitation. At the same time, it's powerful enough to serve millions of Express users. Instead of trying to shoehorn Express middleware onto LB4 programming model, I think it's best to accept the limitations, provide a thin bridge allowing Express middleware to be used in LB4 apps (following the Express way) and advise the users to use (or write) LB4-native extensions once they outgrow the design of Express middleware.

2. Once we get to the point of writing an LB4 extension, we should use LB4 building blocks and design language. An extension wishing to contribute a new sequence action should do exactly that - contribute a sequence action, not a new middleware. I believe we have already validated this approach in several extensions we provide by now (authentication, authorization, logging, etc.)

I don't see a place for a generic Koa-like middleware here.


> Express does not allow dependency injection. We would like to enable injection of middleware configurations.

I don't understand what is the benefit of using DI for middleware configuration. Can you please explain what use cases you have in mind?

To me, this feels like _"if you have a hammer, everything looks like a nail"_.

bajtos:
Setting my reservations aside, here are few comments for the current implementation.

(0) Most importantly: let's make sure we make it easy to solve easy problems and not too much difficult to solve more complex ones. Adding a global middleware to a LB4 app should be as close to the simplicity of Express' `app.use(compression())` as possible.

(1) I see this PR is adding new packages, did you follow all steps from [Adding a new package](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#adding-a-new-package) checklist?

(2) How do you handle the case when Express middleware does not call `next`? IMO, we must return control back to the sequence class, I don't think your implementation does that. Consider the example from https://github.com/strongloop/loopback-next/pull/2434

```ts
  // constructor with @inject decorators

    const start = process.hrtime();
    try {
      await this.invokeMiddleware(); 
      this.send(response, result);
    } catch (error) {
      this.reject(context, error);
    } finally {
      const delta = process.hrtime(start);
      console.log(`Handled "${request.method} ${request.url}" in ${delta[0]} seconds.`);
    }
  }
}
```

Let's say we have `cors` or `favicon` middleware configured and the middleware handle the request (e.g. returned a response for `OPTIONS` request, returned the favicon image). As a LB4 developer, I would expect a timing log printed for these requests too. I don't think that will happen with your current design?

(3) I find the proposed documentation as too involved for new LB developer. Consider the following user story:

- As a LB4 app developer, I want to add few Express middleware to improve my LB4 application.
- I want `favicon` middleware to be invoked as the first one, _before_ CORS is executed.
- I want to invoke various `helmet` middleware  _after_ CORS is executed, e.g. `helmet.hidePoweredBy()`.
- Middleware ordering may look somewhat arbitrary, but I have intentionally crafted it to force you to consider ordering of middleware in relation to built-in middleware like CORS.

To learn how to do that, I have to read 300+ lines of the docs until I find the line https://github.com/strongloop/loopback-next/pull/5118/files#diff-a429ccda8ffe42b31d5816e9d22a27b5R325.

I think it would be better to structure the docs as follows:

1. Start with a step-by-step guide showing how to solve the most common use case (add a new middleware to the sequence at a given position relative to other middleware). Keep this high-level - describe how to achieve the outcome, don't explain how the sausage is made. Most LB4 users should not need to learn more than this.

2. Then provide content for users that need more than just the basic happy path. Here you can explain the concept of Express, Koa and LoopBack middleware, the types and implementation patterns involved, etc. I think your current docs content is pretty much covering this part :+1:

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,490 @@
...
+import {config, globalInterceptor} from '@loopback/core';
+import helmet, {IHelmetConfiguration} from 'helmet';
+import {ExpressMiddlewareInterceptorProvider} from '@loopback/express';
+
+@globalInterceptor('middleware', {tags: {name: 'Helmet'}})
+export class MorganInterceptor extends ExpressMiddlewareInterceptorProvider<
+  IHelmetConfiguration
+> {
+  constructor(
+    @config()
```
bajtos, line 274: +1 for using existing `@config` infrastructure 👏

packages/express/src/__tests__/acceptance/test-helpers.ts
```diff
@@ -0,0 +1,117 @@
...
+      @intercept(...interceptors)
+      hello(msg: string) {
+        return `Hello, ${msg}`;
+      }
+    }
+    const binding = this.app.controller(MyController);
+    this.app.expressServer.expressApp.post('/hello', async (req, res, next) => {
+      try {
+        const controller = await this.app.get<MyController>(binding.key);
+        const proxy = createProxyWithInterceptors(
```
jannyHou, line 64: hmm why `proxy` is needed here?

victor-enogwe:
@raymondfeng nice work on your pull request, I recently started using loopback, great framework, thanks to all the contributors. I do hope this pull-request solves most of the middleware issues. 

Here is my experience so far with adding an express middleware.

I wanted to run a session middleware before the decorated method using an interceptor. That wasn't happening as according to the [docs](https://loopback.io/doc/en/lb4/Interceptors.html#apply-interceptors), the `@interceptor` decorator only adds metadata to the method and does not wrap the method. 

This metadata is then used to determine when the interceptor will be called based on the grouping. I did not find in the documentation how to make these interceptors run before my `@authenticate` decorated route method. 

To solve this I resulted in either calling the middleware in the sequence handler or adding them directly to my strategy classes or to my controller classes. That sort of introduces an inconsistent way of injecting middleware into the app, which would have been really elegant using only `interceptors`.

I think interceptors already do a great job with adding middleware, and probably just needs a way to determine where and when to call the middleware method, plus loopback does a great job at making almost any app context globally available.

What do you think of the `Hooks and Anchors pattern`(I think that's the name), with priority/grouping as implemented already to add middleware and also determine where the middleware will be called? 

An example would be WordPress's use of hooks like `add_action` and `do_action` to specify where and when to run a function.

raymondfeng:
@bajtos Thank you for the feedback and opinions. I have incorporated some of them into the latest code, such as:

1. Allow `InvokeMiddleware` to run an explicit list of Express middleware handler functions
2. Add more flavors of `expressMiddleware()` to register Express middleware handler/factory functions

I also refactored the `Express-middleware.md` into two files:

- Middleware.md -Introduction of Middleware and the architectural rationale   
- Express-middleware.md - Serving as a user guide to plug in Express middleware into LoopBack as part of sequence actions and interceptors.

I would like to justify why I want to open up `Sequence` to allow pluggable middleware.

1. The `Middleware` and `MiddlewareChain` are not new concepts. They are just specialization of `GenericInterceptor` and `GenericInterceptorChain`.  We already use the same concept for method invocations. Being able to use the same programming model concepts for different layers provide the consistency and ease of learning. I also envision to have similar chains to intercept outbound interactions for datasources/connectors.
 
2. The core of LoopBack is built upon the ideas of IoC, DI, and Extension Point/Extension. We can achieve IoC for most parts of LoopBack except `Sequence`, which is hard-coded actions generated by CLI. It's not very friendly to contribute actions from extension modules using `Component` today. Let me show a few examples:

    - `@loopback/authentication` cannot be plugged into an application by `this.component(AuthenticationComponent)`. The manual step to add `Authenticate` action violates the design of `Component` to some extent. 
    - In contrast, `@loopback/authorization` contributes a global interceptor without requiring extra steps to set up the application beyond mounting the component.   
   - To contribute a `Helmet` component for LoopBack, I would expect it should be possible to contribute middleware without requiring manual update of `Sequence`. The prospect component should be able to leverage DI to customize the configuration.

3. I would like to allow metadata driven composition of the sequence. Imagine a LoopBack `starter` can simply bundle itself as a component that sets up bindings as its template to customize what features to be used and how they are configured.  I don't understand why `Sequence` has to be hard-coded.  For debugging, we have good trace of interceptor chains. It's not different from contributing interceptors.

4. I would like to allow configuration to be updated at runtime by simply rebinding the configuration value and make it possible for the middleware to pick up. For example, adjust logging level from `info` to `debug` if necessary.

5. I would to follow a simple principle: Make it easy to do simple things while make it possible to do advanced things. I agree the PR can solve two of the related problems but it's shooting two birds in one stone, IMO. :-).
    - How to leverage the vast ecosystem of existing Express middleware when building LB4 apps.
    - How to allow LB4 extensions to contribute middleware-like functionality.
We're not reinventing the wheel as we are just extending the interceptor idea to new use cases. We also want to allow adapting and applying Express middleware as invocation interceptors anyway.

Maybe we can find a good middle ground to move forward quickly as follows:

1. I'll try to make APIs as simple as possible for your use cases from whom just wants to drop in Express middleware - similar as `app.use`, for example `RestServer.use`?

2. I have added acceptance tests in the PR to demonstrate some of the advanced use cases I have in mind. You can propose new ideas.

3. We don't have to be perfect for the first iteration. We can always improve the design and refine it over time.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+title: 'Using Express middleware'
+keywords: LoopBack 4.0, LoopBack 4, Express, Middleware
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Express-middleware.html
+---
+
+## Overview
+
+Express is the most popular web framework for Node.js developers. As quoted
+below from
```
emonddr, line 12: > below from

below from the

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+   [`Helmet`](https://helmetjs.github.io/) to provide better protection for
+   LoopBack.
+2. Allow actions to be invoked in a configurable order.
+3. Allow extensions to leverage LoopBack's dependency injection and extension
+   point/extension capabilities.
+
+## Bring Express middleware to LoopBack
+
+Let's start with a few examples to illustrate how we can bring Express
+middleware to LoopBack applications with minimal efforts.
```
emonddr, line 59: > efforts

effort

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+
+### Use Express middleware within the sequence of actions
+
+Express middleware can now be plugged into the REST sequence with an
+`InvokeMiddleware` action being injected to the default sequence.
```
emonddr, line 64: > Express middleware can now be plugged into the REST sequence with an `InvokeMiddleware` action being injected to the default sequence.

Express middleware can now be plugged into the default REST sequence via the `InvokeMiddleware` action.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+  async handle(context: RequestContext): Promise<void> {
+    try {
+      const {request, response} = context;
+      // `this.invokeMiddleware` is an injected function to invoke a list of
+      // Express middleware handler functions
+      const finished = await this.invokeMiddleware(context, [
+        helmet({}), // options for helmet is fixed and cannot be changed at runtime
+        morgan('combined', {}), // options for morgan is fixed and cannot be changed at runtime
+      ]);
+      if (finished) {
```
emonddr, line 89: What is `finished`'s value exactly? What if I want to perform some middleware, and then continue with the sequence actions below?

docs/site/Express-middleware.md
emonddr: Is this section the same as `#### Register Express middleware to be executed by `InvokeMiddleware` actions`?

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+        key: 'middleware.morgan',
+      },
+    );
+  }
+}
+```
+
+### Use Express middleware as interceptors for controllers
+
+The LoopBack global and local interceptors now also serve as a revenue to attach
```
emonddr, line 147: what you mean by `revenue ` ?

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+```ts
+import {createInterceptor} from '@loopback/express';
+import helmet, {IHelmetConfiguration} from 'helmet';
+
+const helmetConfig: IHelmetConfiguration = {};
+const helmetInterceptor = createInterceptor(helmet, helmetConfig);
+```
+
+If the Express middleware module does not expose a factory function conforming
+to `ExpressMiddlewareFactory` signature, a wrapper can be created. For example:
```
emonddr, line 174: > to `ExpressMiddlewareFactory` signature

to the `ExpressMiddlewareFactory` signature

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+
+```ts
+import morgan from 'morgan';
+
+// Register `morgan` express middleware
+// Create a middleware factory wrapper for `morgan(format, options)`
+const morganFactory = (config?: morgan.Options) => morgan('combined', config);
+```
+
+### Define a provider class for middleware based interceptor
```
emonddr, line 184: > middleware based 

middleware-based

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+
+It's often desirable to allow dependency injection of middleware configuration
+for the middleware. We can use `defineInterceptorProvider` to simplify
+definition of such provider classes.
```
emonddr, line 188: > definition of

the definition of

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,317 @@
...
+class MyController {
+  hello(msg: string) {
+    return `Hello, ${msg}`;
+  }
+}
+```
+
+### Use `lb4 interceptor` command to create interceptors for Express middleware
+
+The `lb4 interceptor` can be used to generate skeleton implementation of global
```
emonddr, line 272: > generate skeleton implementation

generate a skeleton implementation for global or local

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+   injection of middleware configurations.
+
+To harvest the best of breeds of both frameworks, we build the integration on
+top of the [interceptor](Interceptors.md) with community input as illustrated in
+the diagram below:
+
+![middleware](imgs/middleware.png)
+
+In the diagram, a specialized type of interceptors is introduced as `Middleware`
+for LoopBack. Our middleware is basically a function with the signature:
```
emonddr, line 89: the signature is missing here.

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+It's often desirable to reuse Express middleware in the sequence to handle API
+requests/responses without reinventing the wheel. We now add an
+`InvokeMiddleware` action as the first step in the default sequence. The action
+itself is an interceptor chain of `MiddlewareContext`. It uses the powerful
+[extension point/extension pattern](Extension-point-and-extensions.md) to
+discover registered Express middleware and invoke them as a chain.
+
+#### Default sequence
+
+The generated `src/sequence.ts` now has a new `middleware` property to
```
emonddr, line 115: > to `DefaultSequence`

in `DefaultSequence`

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+`DefaultSequence` to allow injection of the middleware chain while keeping
```
emonddr, line 116: > while keeping

while keeping the

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+
+{% include code-caption.html content="src/sequence.ts" %}
+
+```ts
+export class SequenceWithMiddleware extends DefaultSequence {
+  async handle(context: RequestContext): Promise<void> {
+    try {
+      const {request, response} = context;
+      // The default middleware chain
+      await this.invokeMiddleware(context);
```
emonddr, line 164: no check for `finished` anymore on top and bottom?

emonddr: why not specifiy :  PRE_INVOCATION_MIDDLEWARE ?

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+      this.send(response, result);
+    } catch (error) {
+      this.reject(context, error);
+    }
+  }
+}
+```
+
+It's also possible to inject multiple instances of `InvokeMiddleware`, each of
+which has its own extension point name.
```
emonddr, line 185: What is the `different` extension point name being used below? It is not obvious.

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+Now we can set up applications to leverage the new sequence:
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+export class MyApplication extends RestApplication {
+  constructor(config: ApplicationConfig) {
+    // Create another middleware phase. This is NOT needed if we use the same
```
emonddr, line 225: Not sure what the comment means. Can you clarify?  we seem to be using invokeMiddleware WITH a different extension, so the comment is confusing.

docs/site/Middleware.md
```diff
@@ -0,0 +1,282 @@
...
+    // `invokeMiddleware` with different extension point names.
+    this.bind('middleware.postInvoke')
+      .toProvider(InvokeMiddlewareProvider)
+      .tag({[CoreTags.EXTENSION_POINT]: POST_INVOCATION_MIDDLEWARE});
+    this.sequence(SequenceWithMiddleware);
+
+    // Register a middleware
+    // `spy` is the factory function of a middleware
+    //
+    const spyBinding = this.expressMiddleware(spy, undefined, {
```
emonddr, line 235: what is the undefined for (spy,undefined) ?

docs/site/Express-middleware.md
raymondfeng: Google search shows `minimal efforts` is more commonly used.

docs/site/Express-middleware.md
raymondfeng: `finished` indicates that http response is produced by one of the middleware and we should not proceed with other actions.

docs/site/Express-middleware.md
raymondfeng: Ha, it should be `avenue`.

docs/site/Middleware.md
raymondfeng: That's the default.

bajtos:
> I would like to justify why I want to open up `Sequence` to allow pluggable middleware.

Thank you for providing more details behind your decision. Can we capture this somewhere in our docs please for future readers? (See [Lightweight Architecture Decision Records](https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records).) Maybe in `Middleware.md`?

>  I would like to allow metadata driven composition of the sequence. Imagine a LoopBack `starter` can simply bundle itself as a component that sets up bindings as its template to customize what features to be used and how they are configured.  I don't understand why `Sequence` has to be hard-coded.  For debugging, we have good trace of interceptor chains. It's not different from contributing interceptors.

Fair enough. Maybe we have outgrown the ideas behind Sequence design and it's time to abandon Sequence in favor of something else?

From what I remember, @ritch introduced Sequence as a way how to avoid all the different problems we observed with Express middleware in LoopBack 3. I hope that by moving from Sequence to Middleware/Interceptor chain, we won't reintroduce those problem back again.


> Maybe we can find a good middle ground to move forward quickly as follows:
> 
> 1. I'll try to make APIs as simple as possible for your use cases from whom just wants to drop in Express middleware - similar as `app.use`, for example `RestServer.use`?

Sure. I am not sure if `use` is the best name as it's not clear that it's for middleware, but that's a detail.

> 2. I have added acceptance tests in the PR to demonstrate some of the advanced use cases I have in mind. You can propose new ideas.

Cool, I'll take a look.

> 3. We don't have to be perfect for the first iteration. We can always improve the design and refine it over time.

In my experience, some design aspects are difficult to change in a backwards compatible way. The difficult part is to tell which aspects are easy to refine and which not 🤔

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,319 @@
...
+
+export class MySequence extends DefaultSequence {
+  async handle(context: RequestContext): Promise<void> {
+    try {
+      const {request, response} = context;
+      // `this.invokeMiddleware` is an injected function to invoke a list of
+      // Express middleware handler functions
+      const finished = await this.invokeMiddleware(context, [
+        helmet({}), // options for helmet is fixed and cannot be changed at runtime
+        morgan('combined', {}), // options for morgan is fixed and cannot be changed at runtime
```
bajtos, line 87: Please note we must not call middleware factories for every incoming request, that would be very bad for performance.

Proposed change:

```ts
const middleware = [
  helmet({}), // options for helmet is fixed and cannot be changed at runtime
  morgan('combined', {}), // options for morgan is fixed and cannot be changed at runtime
];

export class MySequence extends DefaultSequence {
   async handle(context: RequestContext): Promise<void> {
      // ...
      const finished = await this.invokeMiddleware(context, middleware);
      // ...
  }
}
```

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,319 @@
...
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import morgan from 'morgan';
+import {ApplicationConfig} from '@loopback/core';
+import {RestApplication} from '@loopback/rest';
+
+export class MyApplication extends RestApplication {
+  constructor(config: ApplicationConfig) {
+    this.expressMiddleware(
```
bajtos, line 132: Maybe `useExpressMiddleware` or `useMiddleware` would be a better name? The usual convention for method names is to start with a verb.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,319 @@
...
+}
+```
+
+It's also possible to bind the middleware to a context as a local or global
+interceptor.
+
+```ts
+import helmet, {IHelmetConfiguration} from 'helmet';
+const binding = registerExpressMiddlewareInterceptor(
+  app,
```
bajtos, line 249: Where is `app` coming from? Where should be this code snippet placed in a LB4 application?

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -103,6 +103,14 @@ children:
       url: Parsing-requests.html
       output: 'web, pdf'
 
+  - title: 'Middleware'
+    url: Middleware.html
+    output: 'web, pdf'
```
bajtos, line 108: I feel this page does not belong to [Key Concepts](https://loopback.io/doc/en/lb4/Concepts.html) and should be moved to [Behind the scene](https://loopback.io/doc/en/lb4/Behind-the-scene.html).

I am expecting that new LB users won't need to understand the information from this page until they get into really advanced use cases.

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -103,6 +103,14 @@ children:
...
+
+  - title: 'Using Express Middleware'
+    url: Express-middleware.html
+    output: 'web, pdf'
```
bajtos, line 112: I think this page should be nested inside [Sequence](https://loopback.io/doc/en/lb4/Sequence.html). Please note that the Sequence page already provides some content related to Express-middleware, see [Working with Express middleware](https://loopback.io/doc/en/lb4/Sequence.html#working-with-express-middleware).

Please update that existing content to work well with your new pages and to avoid duplicative and/or contradicting information.

_I have also briefly considered the option to put this page into [Usage scenarios](https://loopback.io/doc/en/lb4/Usage-scenarios.html), but since we already have some middleware-related content in Sequence, it may be best to put the new content there too._

packages/express/README.md
```diff
@@ -0,0 +1,153 @@
+# @loopback/express
+
+This package adds middleware support for LoopBack 4 and allows Express
+middleware to be plugged into LoopBack seamlessly. It's used by `@loopback/rest`
+to support Express middleware with `InvokeMiddleware` action within the REST
+sequence.
```
bajtos, line 6: Is it worth the effort to maintain this feature as a standalone package, when it's not an optional extension and it's always used by (included in) `@loopback/rest`?

docs/site/sidebars/lb4_sidebar.yml
bajtos: Also, do you have any plans for https://loopback.io/doc/en/lb4/express-with-lb4-rest-tutorial.html? Should we update it to mention the newer (and better) way for mounting Express middleware? Perhaps we should deprecate and eventually remove that page?

packages/express/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "README.md",
+    "index.js",
+    "index.d.ts",
+    "dist",
+    "src",
+    "!*/__tests__"
+  ],
+  "dependencies": {
+    "@loopback/context": "^3.5.1",
+    "@loopback/core": "^2.4.1",
```
bajtos, line 38: Please make sure this package is using the version of `@loopback/core` and `@loopback/context` as provided by the target LB4 application, so that app users are in full control.

Proposed changes:
- Move `@loopback/core` to peer dependencies
- Remove `@loopback/context` entirely, import context-related APIs from `@loopback/context`.

packages/rest/src/rest.server.ts
```diff
@@ -251,7 +242,11 @@ export class RestServer extends Context implements Server, HttpServerLike {
 
     // Allow CORS support for all endpoints so that users
     // can test with online SwaggerUI instance
-    this._expressApp.use(cors(this.config.cors));
+    this.expressMiddleware(cors, this.config.cors, {
+      injectConfiguration: false,
+      key: 'middleware.cors',
+      group: 'cors',
+    });
```
bajtos, line 249: I think we should eventually not rely on the built-in CORS middleware and modify our application template to include explicit setup of CORS.

docs/site/Express-middleware.md
emonddr: hmm. doesn't sound right. See https://dictionary.cambridge.org/example/english/minimal-effort

docs/site/Middleware.md
```diff
@@ -0,0 +1,335 @@
...
+    try {
+      const {request, response} = context;
+      // The default middleware chain
+      await this.invokeMiddleware(context);
+      const route = this.findRoute(request);
+      const args = await this.parseParams(request, route);
+      const result = await this.invoke(route, args);
+
+      // The second middleware chain for post-invocation processing
+      context.bind('invocation.result').to(result);
```
emonddr, line 257: What does this do? Is it needed?

docs/site/Middleware.md
```diff
@@ -0,0 +1,335 @@
...
+      const {request, response} = context;
+      // The default middleware chain
+      let finished = await this.invokeMiddleware(context);
+      if (finished) return;
+      const route = this.findRoute(request);
+      const args = await this.parseParams(request, route);
+      const result = await this.invoke(route, args);
+
+      // The second middleware chain for post-invocation processing
+      context.bind('invocation.result').to(result);
```
emonddr, line 218: Why is this needed?

packages/rest/src/__tests__/acceptance/middleware/middleware-sequence.acceptance.ts
```diff
@@ -0,0 +1,177 @@
...
+        const {request, response} = context;
+        // The default middleware chain
+        let finished = await this.invokeMiddleware(context);
+        if (finished) return;
+        const route = this.findRoute(request);
+        const args = await this.parseParams(request, route);
+        const result = await this.invoke(route, args);
+
+        // The second middleware chain for post-invocation processing
+        context.bind('invocation.result').to(result);
```
emonddr, line 81: Why is this needed?

packages/rest/src/__tests__/acceptance/middleware/middleware-sequence.acceptance.ts
```diff
@@ -0,0 +1,177 @@
...
+        // The default middleware chain
+        let finished = await this.invokeMiddleware(context);
+        if (finished) return;
+
+        const route = this.findRoute(request);
+        const args = await this.parseParams(request, route);
+        const result = await this.invoke(route, args);
+
+        // The second middleware chain for post-invocation processing
+        context.bind('invocation.result').to(result);
```
emonddr, line 118: Why is this needed?

docs/site/Middleware.md
```diff
@@ -0,0 +1,335 @@
...
+
+To use existing Express middleware, please check out
+[Using Express Middleware](./Express-middleware.md).
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import {ApplicationConfig} from '@loopback/core';
+import {RestApplication} from '@loopback/rest';
+import {log} from './log-middleware';
```
emonddr, line 310: Where is `log-middleware` defined?

docs/site/Express-middleware.md
raymondfeng: I accept `minimal effort` 👍

docs/site/Express-middleware.md
raymondfeng: I choose to use `expressMiddleware` to be consistent with other registration related apis, such as:

- app.controller()
- app.component()
- app.dataSource()
- app.service()

docs/site/sidebars/lb4_sidebar.yml
raymondfeng: I put it under `Key Concepts` for two reasons:

- It's next to `Interceptors`
- Application developers need to add middleware

docs/site/sidebars/lb4_sidebar.yml
raymondfeng: > I think this page should be nested inside Sequence. Please note that the Sequence page already provides some content related to Express-middleware, see Working with Express middleware.

We might have to use this page for two cases:

- Apply Express middleware in the sequence
- Apply Express middleware as interceptors

packages/express/README.md
raymondfeng: I think so. My plan is to allow the usage of an existing Express application without dragging in `@loopback/rest`.

packages/express/package.json
raymondfeng: We can remove `@loopback/context` and use `@loopback/core` only. But we need to apply the same rule for all other packages.

I'm also reluctant to use use `peerDependencies`. IMO, the only reliable way is to use the same set of versions released together and upgrade using `lb4 update`.

Let's defer this to a later time that we can visit all packages in `loopback-next`.

docs/site/Middleware.md
raymondfeng: This shows the possibility to access results from other actions inside a middleware.

docs/site/Middleware.md
raymondfeng: I added a caption for 1st reference.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,327 @@
...
+import {toInterceptor} from '@loopback/express';
+import morgan from 'morgan';
+
+const morganInterceptor = toInterceptor(morgan('combined'));
+```
+
+#### Create an interceptor from Express middleware factory function
+
+```ts
+import {createInterceptor} from '@loopback/express';
```
jannyHou, line 169: could you explain more about when use `createInterceptor` and when use `toInterceptor`?
e.g. for this example, why not using:

`const helmetInterceptor = toInterceptor(helmet(helmetConfig));`

packages/express/src/__tests__/acceptance/test-helpers.ts
```diff
@@ -0,0 +1,129 @@
...
+    const interceptors: InterceptorOrKey[] = [];
+    if (interceptor) interceptors.push(interceptor);
+    class MyController {
+      @intercept(...interceptors)
+      hello(msg: string) {
+        return `Hello, ${msg}`;
+      }
+    }
+    const binding = this.app.controller(MyController);
+    this.app.expressServer.expressApp.post('/hello', async (req, res, next) => {
```
jannyHou, line 61: a bit confused: the testing endpoint is set up by express server, why not use the REST decorators? Is there a difference?

raymondfeng: This module does not have dependency on `@loopback/rest` by design. It should possible to use it with plain Express.

docs/site/Express-middleware.md
raymondfeng: `toInterceptor()` takes the Express handler function directly while `createInterceptor` takes the factory function and config object.

docs/site/Express-middleware.md
```diff
@@ -0,0 +1,321 @@
...
+import helmet, {IHelmetConfiguration} from 'helmet';
+
+const helmetConfig: IHelmetConfiguration = {};
+const helmetInterceptor = createInterceptor(helmet, helmetConfig);
+```
+
+If the Express middleware module does not expose a factory function conforming
+to the `ExpressMiddlewareFactory` signature, a wrapper can be created. For
```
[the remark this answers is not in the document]
raymondfeng: Yes

docs/site/Middleware.md
raymondfeng: Added more comments.

jannyHou:
> @loopback/authentication cannot be plugged into an application by this.component(AuthenticationComponent). The manual step to add Authenticate action violates the design of Component to some extent.
    In contrast, @loopback/authorization contributes a global interceptor without requiring extra steps to set up the application beyond mounting the component.
    To contribute a Helmet component for LoopBack, I would expect it should be possible to contribute middleware without requiring manual update of Sequence. The prospect component should be able to leverage DI to customize the configuration.

👍 👍 After this PR we can register the anthenticate action into a middleware, then no need to manually modify the sequence.

Nice docs and feature! A few questions.

packages/express/src/__tests__/acceptance/test-helpers.ts
deepakrkris: Any advantage we have, by this being an independent package ?

packages/express/src/express.server.ts
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Express settings
+   */
+  settings?: Record<string, unknown>;
+};
+
+/**
+ * An Express server that provides middleware composition and injection
+ */
+export class ExpressServer extends MiddlewareRegistry implements Server {
```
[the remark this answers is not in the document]
raymondfeng: We could make `MiddlewareRegistry` a mixin if needed.

packages/rest/src/sequence.ts
```diff
@@ -95,6 +103,8 @@ export class DefaultSequence implements SequenceHandler {
   async handle(context: RequestContext): Promise<void> {
     try {
       const {request, response} = context;
+      // Invoke registered Express middleware
+      await this.invokeMiddleware(context);
```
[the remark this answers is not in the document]
raymondfeng: Yes.

PR #4794, opened 2020-03-04
[RFC WIP] feat(typeorm): add a component for TypeORM integration

raymondfeng:
A PoC to illustrate possible shadow integration with TypeORM:

- TypeOrmComponent to maintain connections
- `@typeorm.connection` to provide injection of TypeORM connections - can be extended to allow repository...

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5152, opened 2020-04-19
[RFC WIP] feat(typegoose): add a component for typegoose/mongoose integration

mschnee:
Inspired by #4794 and by a project I am working with that has a large number of existing Typegoose and Mongoose schemas/models already defined.

Unlike TypeORM, which provides connections that have typed entity managers, mongoose and Typegoose put the Model above the connection, and developers interact with Model objects directly to fetch/query/update documents through the underlying connection.

This is a very rough WIP, and more of a weekend project to start a discussion. 

## Goals
- Easily take your existing mongoose or typegoose schemas, and use them in loopback without creating a specialized connection provider of your own.
- `@inject()` the connection-models you need when you need them, instead of `require()`ing them in the global scope, which is common in lots of mongoose/typegoose examples.
- Keep it simple.

## Example?
```ts
/**
 * schemas/User.schema.ts
 */
export class User {
  @prop()
  firstName: string;
}
export const UserBindingKey = BindingKey.create<ReturnModelType<typeof User>>(
  'key.for.this.connection.Model',
);

/**
 * schemas/Event.schema.ts
 */
@modelOptions({schemaOptions: {discriminatorKey: 'kind'}})
@pre<Event>('save', function () {
  this.createdAt = new Date();
})
export class Event {
  @prop()
  createdAt: Date;
}
export const EventBindingKey = BindingKey.create<ReturnModelType<typeof Event>>(
  'key.for.this.connection.Event',
);

/**
 * schemas/LoginEvent.schema.ts
 */
export class LoginEvent extends Event {
  @prop({ref: 'User'})
  _user: Ref<User>;
}
const LoginEventBindingKey = BindingKey.create<
  ReturnModelType<typeof LoginEvent>
>('key.for.this.connection.LoginEvent');

/**
 * application.ts
 */
import {Event, LoginEvent, User } from './schemas';
import {EventBindingKey, LoginEventBindingKey, UserEventBindingKey} from './schemas';
import {TypegooseBindings, TypegooseCompeonent} from '@loopback/typegoose';

class Application extends RestApplication {
  constructor(options?: ApplicationOptions) {
    super(options);
  
    this.configure(TypegooseBindings.COMPONENT).to({
      uri: options.MONGO_URL || 'mongodb://localhost:27017',
      schemas: [
        {bindingKey: UserBindingKey, schema: User},
        {bindingKey: EventBindingKey, schema: Event},
      ],
      discriminators: [
        {
          bindingKey: LoginEventBindingKey,
          modelKey: EventBindingKey,
          schema: LoginEvent,
        },
      ],
    });
    this.component(TypegooseComponent);
  }
}

/**
 * controllers/user.controller.ts
 */
export class UserController {
  constructor(
    @inject(UserBindingKey) UserModel: ReturnModelType<typeof User>
  )

  @get('/first-user')
  public async function() {
    const user = await this.UserModel.findOne();
    return user?._id ?? null;
  }
}
```

## TODO
- Tests need to be greatly expanded beyond the simple integration test
- The `TestTypegooseComponent` is messy, and I'm not sure if it's a good pattern.
- Requiring `BindingKey` may be too strict, and too inflexible for some cases where applications may have dynamic connections.
- What's the best way to pass configuration when an application has hundreds of schemas?  `application.ts` seems like the wrong place.
- Also provide a similar component for mongoose

## Background
The Hello World example for mongoose looks like this:
```js
const mongoose = require('mongoose');
await mongoose.connect('mongodb://127.0.0.1:27017');

const userSchema = new mongoose.Schema({
  firstName: { type: String },
})
const UserModel = mongoose.model(userSchema, 'User');

const newUser = await UserModel.create({ firstName: 'Matt' });
```

What's hidden under the hood are the connections.  You can connect to multiple database hosts.

```js
const mongoose = require('mongoose');
const connection1 = await mongoose.createConnection('mongodb://127.0.0.1:27017');
const connection2 = await mongoose.createConnection('mongodb://localhost:27017');

const userSchema = new mongoose.Schema({
  firstName: { type: String },
})

const UserModel1 = connection1.model(userSchema, 'User');
const UserModel2 = connection2.model(userSchema, 'User');
```
_Even though these two connections refer to the same database, mongoose treats references separately_.  

Typegoose adds stricter type checking with Typescript to mongoose:
```ts
class User  {
  @prop()
  firstName: string;
}

const UserModel = getModelForClass(User);
```

```ts
class User  {
  @prop()
  firstName: string;
}

const UserModel = getModelForClass(User, {
  existingConnection, // using an already created connection from mongoose.createConnection()
  existingMongoose, // using an already connected mongoose singleton
});
```
## Checklist


bajtos:
Thank you for starting the discussion about adding support for Mongoose-based ORM framework for working with MongoDB! ❤️ 👏 

Few high-level comments:
- I see this PR is adding new packages, did you follow all steps from [Adding a new package](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#adding-a-new-package) checklist?

- I think it's best to work on this feature iteratively in a series of small(er) pull requests. Let's add a warning to README about the experimental status of the new package, to set the expectations of our users right. Just copy the warning from other experimental packages, e.g. https://github.com/strongloop/loopback-next/blob/66239b87c010c4c3b106dd0bf3fdc58a3dbc0170/extensions/logging/README.md#L7-L12

@raymondfeng can you please review? You are the author of #4794 which inspired this work.
@agnes512 @hacksparrow can you please review too? You are standby owners of MongoDB and Model/Repository areas.

mschnee:
In fact, I did not.  I will do so, and commit changes.

extensions/mongoose/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@loopback/mongoose",
+  "version": "0.0.1",
+  "description": "@loopback/mongoose",
+  "keywords": [
+    "loopback-extension",
+    "loopback"
```
raymondfeng, line 7: Add `mongodb` and `mongoose` into `keywords`?

extensions/mongoose/src/__tests__/integration/schemas/Event.schema.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
raymondfeng, line 1: Let's use lowercase for file names consistently - `event.schema.ts`.

extensions/mongoose/src/__tests__/integration/schemas/Event.schema.ts
```diff
@@ -0,0 +1,26 @@
...
+// Node module: @loopback/typeorm
```
raymondfeng, line 2: Please use `lb4 copyright` to update file headers.

raymondfeng:
@mschnee Great initiative. In addition to my comments inline, I would suggest:

1. Split the PR into two - one for mongoose and the other for typegoose
2. Have more information in README

mschnee:
Done!

I'm not entirely sure how I feel about having to pass all the schemas in via configuration.  It's not the worst possible pattern- for both the `typegoose` and `mongoose` extensions I'm using the `index.ts` file and an exported function for this purpose.

raymondfeng:
> I'm not entirely sure how I feel about having to pass all the schemas in via configuration. It's not the worst possible pattern- for both the typegoose and mongoose extensions I'm using the index.ts file and an exported function for this purpose.

We can possibly define an extension point for Mongoose schemas. Then there can be a booter to load such schemas or have app code to explicitly contribute to the extension point.

extensions/mongoose/package.json
mschnee: Done!

extensions/mongoose/src/__tests__/integration/schemas/Event.schema.ts
mschnee: Done!

bajtos:
> I'm not entirely sure how I feel about having to pass all the schemas in via configuration.  It's not the worst possible pattern- for both the `typegoose` and `mongoose` extensions I'm using the `index.ts` file and an exported function for this purpose.

Will this design allow applications to register additional schemas dynamically at runtime, after the application started?

hacksparrow:
@mschnee thanks for the PR. Here's my feedback.

1. The DX should be in tune with the LB4 way - declarative, bootable, and dynamic (usable after the app starts).
2. We should expose a friendlier API than directly exposing that of Mongoose when it comes to configuration. Bootable artifacts provide that opportunity.
3. The [TypeORM extension](https://github.com/strongloop/loopback-next/tree/master/extensions/typeorm) is recently added extension, new extensions could use it as a template.

dhmlau:

```
git commit --amend -s
git push --force-with-lease
```

PR #5155, opened 2020-04-19
feat: enable TypeScript project references

raymondfeng:
See https://www.typescriptlang.org/docs/handbook/project-references.html

Supersedes https://github.com/strongloop/loopback-next/pull/1636
Implements #2609 

TypeScript project references seem to be working with latest VSCode now. Please help verify.

## Checklist


bin/update-ts-project-refs.js
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env node
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: loopback-next
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * This is an internal script to update TypeScript project references based on
+ * lerna's local package dependencies.
+ *
```
bajtos, line 10: +1 to automate this process 👏 

I think we need to run this script every time we add a new inter-repo dependency, is that correct? At minimum, please describe this step in our dev docs at https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md.

I think docs is not enough, we will often forget to run the script. IMO, we should check for consistency between package dependencies and ts project dependencies as part of our Travis CI build, similarly to how we check that package locks don't refer to monorepo packages, etc.

packages/build/bin/compile-package.js
```diff
@@ -128,7 +133,76 @@ function run(argv, options) {
...
+    '--incremental',
+    '--assumeChangesOnlyAffectDirectDependencies',
+    '-v',
+    '--verbose',
+    '-d',
+    '--dry',
+    '-f',
+    '--force',
+    '--clean',
+  ];
```
bajtos, line 169: Is there a way how to get this list from `tsc`, so that we don't have to maintain it manually?

packages/build/config/tsconfig.common.json
```diff
@@ -19,6 +19,7 @@
     "target": "es2017",
     "sourceMap": true,
     "declaration": true,
+    "declarationMap": true,
```
bajtos, line 22: Is this necessary? I thought that TypeScript automatically enables declaration maps when project references are used.

Also I think declaration maps are no longer needed for IDEs, see https://github.com/microsoft/TypeScript/pull/32028

packages/build/bin/compile-package.js
raymondfeng: No, I digged into `tsc.js`. The list of options is not exported. :-(

raymondfeng:
Some smoke test results for performance:

1. With this PR
- First-time (clean build)
```
real	0m30.565s
user	0m37.962s
sys	0m3.128s
```
- Second-time (incremental build)
```
real	0m1.476s
user	0m0.960s
sys	0m0.420s
```

2. Without this PR
- First-time
```
real	2m1.125s
user	7m11.410s
sys	0m24.406s
```
- Second-time
```
real	1m12.558s
user	3m45.993s
sys	0m20.833s
```

The improvement is significant.

bin/update-ts-project-refs.js
raymondfeng: It's being run as part of `postinstall` after `lerna bootstrap`.

packages/build/config/tsconfig.common.json
raymondfeng: Reverted.

raymondfeng: I'm not 100% sure if it's fine. See https://www.typescriptlang.org/docs/handbook/project-references.html#declarationmaps. I'll test.

raymondfeng:
> With project references in place, we should get rid of top-level manually-written index.js, index.d.ts and index.ts files. Their purpose is to trick IDE into treating the entire monorepo as a single TypeScript project. That's no longer desirable, we want each package to be treated as its own project. Those index files are also causing slow build times. As I shown in #2613, we can speed up our build by getting rid of them.

I would like to defer this change to a follow-up PR. One of the caveats listed in the project references is that we have to run `tsc -b` first. I'm not sure if it's still true. See https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references.

raymondfeng:
> Did you run the steps listed in How to test infrastructure changes? Are there any issues to be aware of?

Yes. I did.

bajtos:
> Some smoke test results for performance:
> (...)
> The improvement is significant.

This is awesome, even better than I expected.

> I would like to defer this change to a follow-up PR. One of the caveats listed in the project references is that we have to run tsc -b first. I'm not sure if it's still true. See https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references.

Yeah, I was thinking about deferring this part too. The current PR is large enough, it's better to work incrementally. I am happy to take a look at index changes myself after this PR is landed.

> > Did you run the steps listed in How to test infrastructure changes? Are there any issues to be aware of?
>
> Yes. I did.

Cool 👏 Did you find any issues? Does everything work as before?

packages/build/bin/compile-package.js
```diff
@@ -128,7 +134,69 @@ function run(argv, options) {
...
 
-  return utils.runCLI('typescript/lib/tsc', args, {cwd, ...options});
+  const validArgs = validArgsForBuild(args);
+
+  return utils.runCLI('typescript/lib/tsc', validArgs, {cwd, ...options});
+}
+
+/**
+ * `tsc -b` only accepts valid arguments. `npm run build -- --<other-arg>` may
+ * pass in extra arguments. We need to remove such arguments.
```
bajtos, line 144: What kind of arguments are getting in via `npm run build -- --<other-arg>`? If they are provided by the user (either on command line or via their package.json) then we must let them know about the problem and especially which parameters are removed by `lb-tsc`.

Personally, I don't think it's a good idea to add such parameter-modifications to `lb-tsc`, I am concerned about support and maintenance costs. IMO, we should let `tsc` fail the build and expect the user to fix the parameters passed to`lb-tsc`/`tsc`.

bajtos:
BTW, `lb-tsc` provides `--copy-resources` option. I don't think that's going to work with project references. IIUC, `lb-tsc` is invoked for the top-level monorepo-root "project" only, that calls `tsc` to run the build and after that individual packages are built using `tsc`. It's not a problem in this monorepo, because we have removed usage of that flag long time ago, but it's something to be aware of. I am not sure if there is anything we can do about that, because when we run `lb-tsc` for the top-level project, then it's difficult to know what `lb-tsc` arguments are configured by individual packages. Maybe we can put a warning to our current `lb-tsc` docs? (I am not sure where the docs is.)

raymondfeng:
The removal of invalid options is for npm script composition as other npm runs can trigger npm run build but npm always pass extra args to all commands.

bajtos:
> The removal of invalid options is for npm script composition as other npm runs can trigger npm run build but npm always pass extra args to all commands.

I see, do you have any example?

I am concerned that some day, a developer will add a new option to `npm run build` script, `lb-tsc` will silently discard it, and the poor developer will have hard times figuring out what's going on and why their new switch is ignored. Can you please find a way how to let developers know about flags removed by the tool? E.g. use `console.warn` or `debug` logs.

raymondfeng:
> I see, do you have any example?

> I am concerned that some day, a developer will add a new option to npm run build script, lb-tsc will silently discard it, and the poor developer will have hard times figuring out what's going on and why their new switch is ignored. Can you please find a way how to let developers know about flags removed by the tool? E.g. use console.warn or debug logs.

It can reproduced by `cd packages/context; npm test -- --timeout 5000`.

There are `debug` statements to show how `loopback:build` is working.

raymondfeng:
I'm not sure why Travis build fails for some of the acceptance tests - https://github.com/strongloop/loopback-next/pull/5155/checks?check_run_id=603795936.

bajtos:
> It can reproduced by `cd packages/context; npm test -- --timeout 5000`.

Interesting! I always thought that npm does not pass the extra arguments to `pre` and `post` scripts (`pretest` in this case). I guess I was wrong 🤷 

> There are `debug` statements to show how `loopback:build` is working.

Are you referring to the following line?

https://github.com/strongloop/loopback-next/blob/04728a07e4e22a248f09535d06ec1d3f7323e099/packages/build/bin/compile-package.js#L175

I find the message rather cryptic, how about something like this:

```ts
debug('Using the following args for tsc -b', validArgs);
```

> I'm not sure why Travis build fails for some of the acceptance tests - https://github.com/strongloop/loopback-next/pull/5155/checks?check_run_id=603795936.

I quickly checked the error log and don't see any obvious problem 🤔

Issue #4838, opened 2020-03-10
Migration Guide: Boot up


hacksparrow:
Document the differences in the boot up process in LB3 and LB4.  Created out of https://github.com/strongloop/loopback-next/issues/3935.

**Acceptance Criteria**

- [ ] Document the difference the boot up process between LB4 & LB3.

Related https://github.com/strongloop/loopback-next/issues/3384.

agnes512:
https://loopback.io/doc/en/lb4/Life-cycle.html and https://loopback.io/doc/en/lb4/Booting-an-Application.html might help.

hacksparrow:
Closed via https://github.com/strongloop/loopback-next/pull/5178

PR #5178, opened 2020-04-21
docs: booting in LB3 vs LB4

hacksparrow:
Differences in LB3 and LB4 booting. Addresses https://github.com/strongloop/loopback-next/issues/4838.

### DIFF
--- a/docs/site/migration/LB3-vs-LB4-booting.md
+++ b/docs/site/migration/LB3-vs-LB4-booting.md
@@ -1,0 +1,153 @@
+---
+lang: en
+title: 'Differences in LoopBack 3 and LoopBack 4 booting process'
+keywords: LoopBack 4.0, LoopBack 4, LoopBack 3.0, LoopBack 3
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/LB3-vs-LB4-booting.html
+---
+
+## Differences between LoopBack 3 and LoopBack 4 booting process
+
+Because of the architectural differences, the booting process is very different
+in LoopBack 3 and LoopBack 4.
+
+### In LoopBack 3
+
+The booting process in LoopBack 3 is handled by
+[loopback-boot](https://github.com/strongloop/loopback-boot), which performs the
+following tasks:
+
+#### 1. Configuration of application settings
+
+It determines the execution environment, the application host, the application
+port, the application root directory and other artifact directories.
+
+Then it configures the LoopBack Application object from the `config.json` file
+in the application root directory.
+
+#### 2. Configuration of datasources
+
+Datasources are determined from application options or datasource files. They
+are then attached to the LoopBack Application object.
+
+#### 3. Definition of models
+
+Models are determined from application options or model files,
+[mixins](https://loopback.io/doc/en/lb3/Defining-mixins.html) are applied, and
+they are attached to their respective datasources. They are then attached to the
+LoopBack Application object.
+
+#### 4. Setting up middleware
+
+Middleware from the middleware configuration file are resolved and added to the
+LoopBack 3 middleware chain.
+
+#### 5. Setting up components
+
+Components from the component configuration file are resolved and are added to
+the LoopBack 3 middleware chain.
+
+#### 6. Running boot scripts
+
+Boot scripts are resolved from the `server/boot` directory and run.
+
+### In LoopBack 4
+
+In LoopBack 4, the booting tasks are shared between the `@loopback/core`'s
+[Application](../apidocs/apidocs.core.application.md) class and the
+[@loopback/boot](../apidocs/apidocs.boot.html) package. Since your application
+will extend `BootMixin` and `Application`, their separation will not be apparent
+to you, and understanding of how they work together as one is not mandatory for
+developing LoopBack 4 projects.
+
+The following are the list of tasks that are performed in the LoopBack 3 booting
+process, and their equivalent in LoopBack 4.
+
+{% include tip.html content="For all the details about booting in LoopBack 4,
+refer to the
+[booting documentation](../Booting-an-Application.md)." %}
+
+#### 1. Configuration of application settings
+
+There is no `config.json` in LoopBack 4. The application options are passed in
+the constructor of the [Application](../apidocs/apidocs.core.application.md)
+class. Read more about application configuration
+[here](../Application.md#configuring-your-application).
+
+#### 2. Configuration of datasources
+
+Datasources in LoopBack 3 are defined in a single `datasources.json` file. In
+LoopBack 4, datasources are defined as TypeScript classes. They all reside in
+the `src/datasources` directory.
+
+{% include note.html content="Before
+[loopback-next#5000](https://github.com/strongloop/loopback-next/pull/5000)
+landed, datasource classes required a configuration JSON file, it is not
+required anymore." %}
+
+`BootMixin` resolves the datasources, and attaches them to the application. Read
+more about datasource booter
+[here](../Booting-an-Application.md#controller-booter).
+
+Read more about LoopBack 4 datasources [here](../DataSources.md). Read more
+about migrating datasources [here](./datasources.md).
+
+#### 3. Definition of models
+
+Although REST APIs are built around Models, they are not a part of the booting
+process in LoopBack 4; Controllers and Repositories are.
+
+Controllers and Repositories import models as regular TypeScript classes since
+they are required at compile time.
+
+{% include note.html content="Booters are LoopBack 4 artifacts that are used
+for modularly adding functionality to the framework. Controllers, Repositories,
+Datasources, Interceptors, etc., are added using booters, custom booters can
+be written for adding new functionality to the framework." %}
+
+Read more about migrating models [here](./models/overview.md).
+
+#### 4. Setting up middleware
+
+Middleware are not supported at LoopBack level yet. However, you can
+[mount a LoopBack 4 application](../express-with-lb4-rest-tutorial.md) on an
+Express application, which would allow you to use the familiar Express routing
+methods.
+
+{% include tip.html content="Follow our GitHub issues
+[#1293](https://github.com/strongloop/loopback-next/issues/1293)
+and
+[#2035](https://github.com/strongloop/loopback-next/issues/2035)
+to track the progress on supporting Express middlweware in LoopBack 4." %}
+
+If you want to mount an Express router in a LoopBack 4 application, you can use
+the
+[RestApplication.mountExpressRouter()](../Routes.html#mounting-an-express-router)
+API.
+
+#### 5. Setting up components
+
+In LoopBack 3, a
+[component](https://loopback.io/doc/en/lb3/LoopBack-components.html) is a simple
+Node.js module which exports a function with the signature
+`function(app, options)`. In LoopBack 4, a
+[component](../Creating-components.md) is a TypeScript class which can add
+[servers](../Server.md), [observers](../Life-cycle.md),
+[providers](../Creating-components.md#providers), and
+[controllers](../Controllers.md) to the application using dependency injection.
+
+LoopBack 3 components adding routes can be migrated to LoopBack 4 by moving the
+functionality to the controller of a LoopBack 4 component.
+
+#### 6. Running boot scripts
+
+If you used LoopBack 3 boot scripts for adding routes to the application, it
+should now be moved to a standalone controller, a component, or implemented
+using `app.mountExpressRouter()`.
+
+The functionality of non-routing boot scripts can be implemented as life cycle
+observers. Read more about life cycle events and observers
+[here](../Life-cycle.md).
+
+For more information about migrating boot scripts in LoopBack 4, refer to the
+[boot scripts migration guide](./boot-scripts.md).
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -797,6 +797,10 @@ children:
     url: Understanding-the-differences.html
     output: 'web, pdf'
 
+  - title: 'Booting process'
+    url: LB3-vs-LB4-booting.html
+    output: 'web, pdf'
+
   - title: 'Request/response cycle'
     url: LB3-vs-LB4-request-response-cycle.html
     output: 'web, pdf'


docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+the LoopBack 3 middleware chain.
+
+#### 6. Running boot scripts
+
+Boot scripts are resolved from the `server/boot` directory and run.
+
+### In LoopBack 4
+
+In LoopBack 4, the booting tasks are shared between the
+[RestApplication](./apidocs/apidocs.rest.restapplication.html) class
```
raymondfeng, line 54: What part of booting does `RestApplication` handle?

raymondfeng: We should mention it's possible to plug in new types of booters.

raymondfeng: Can we have link to https://loopback.io/doc/en/lb4/Booting-an-Application.html?

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+and the
+[@loppback/boot](./apidocs/apidocs.boot.html) package. Since your application
```
nabdelgadir, line 56:
```suggestion
[@loopback/boot](./apidocs/apidocs.boot.html) package. Since your application
```

docs/site/LB3-vs-LB4-booting.md
hacksparrow: > What part of booting does RestApplication handle?

The config part.

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+
+Datasources are determined from application options or datasource files. They
+are then attached to the LoopBack Application object.
+
+#### 3. Definition of models
+
+Models are determined from application options or model files,
+[mixins](https://loopback.io/doc/en/lb3/Defining-mixins.html) are applied,
+and they are attached to their respective datasources. They are then attached
+to the LoopBack Application object.
```
bajtos, line 35: FYI: In the first step, we _define_ all models. In the second step, we _attach_ the models to datasources and the application.

The list of models defined may be longer than the list of models attached:
1. loopback-boot looks at `server/model-config` to see which files should be added to the app.
2. Then it examines inheritance and mixin configurations, to determine dependencies (which models to add to the list)
3. Models are sorted topologically based on inheritance, so that we define parent models before children.
4. All models from the list above are defined via model registry.
5. Only models specified in `server/model-config` are added to the app and attached to a datasource (unless `dataSource` flag is set to `null`).

loopback-boot also supports universal mode, where the entire loopback application is bundled for browser. In that case, the boot process is split in two parts:
1. The first step finds out which models (and their parents, etc.) are needed by the app. It creates "instructions" for booting an app, these instructions are compiled into the browser bundle together with model source files (`common/models/{model-name}.json`).
2. The second step is performed in the browser, it processes the instructions from the first step.

docs/site/LB3-vs-LB4-booting.md
bajtos: I don't think `RestApplication` is handling the config part, isn't this handled by `Application` from `@loopback/core`? AFAIK, `RestApplication` is a thin wrapper around `Application` that's adding a single REST server instance and convenience methods for accessing that server.

Let's build a better understanding of what you are trying to describe, @hacksparrow, and how is that actually implemented.

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+mandatory for developing LoopBack 4 projects.
+
+The following are the list of tasks that are performed in the LoopBack 3
+booting process, and their equivalent in LoopBack 4.
+
+### 1. Configuration of application settings
+
+There is no `config.json` in LoopBack 4. The application options are passed in
+the constructor of the
+[RestApplication](./apidocs/apidocs.rest.restapplication.html) class.
```
bajtos, line 68: Using `RestApplication` is not mandatory, people can use the base `Application` class too. This will become more and more important as we add support for different API flavors (gRPC, event-stream consumers, etc.)

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+
+### 2. Configuration of datasources
+
+Datasources in LoopBack 3 are defined in a single `datasources.json` file. In
+LoopBack 4, datasources are defined as TypeScript classes that come with an
+associated JSON configuration file. They all reside in the `src/datasources`
+directory.
```
bajtos, line 75: Please note that we are moving away from JSON config files for datasources, see https://github.com/strongloop/loopback-next/pull/5000

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+
+`BootMixin` resolves the datasources, and attaches them to the application.
```
bajtos, line 77: Do we have any documentation on  how is the datasource booter working under the hood? For example, what files it looks for, how it decided which exported member of a datasource TS file it will add to the app, etc. If yes, then please add a link to it.

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+
+Read more about LoopBack 4 datasource [here](./DataSources.md).
+
+### 3. Definition of models
+
+Although REST APIs are built around Models, they are not a part of the booting
+process in LoopBack 4; Controllers and Repositories are.
```
bajtos, line 84: Can you please briefly explain why we are not booting models?

The main reason is that we need Models at compile time, therefore they must be imported directly.

docs/site/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,128 @@
...
+
+Controllers and Repositories import models as regular TypeScript classes.
+
```
bajtos, line 87: Can you please briefly explain why we are not booting models?

The main reason is that we need Models at compile time, because as you have correctly pointed out, controllers and repositories are importing them directly.

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -810,6 +810,10 @@ children:
     url: Understanding-the-differences.html
     output: 'web, pdf'
 
+  - title: 'Booting process'
+    url: LB3-vs-LB4-booting.html
```
bajtos, line 814: Since this new content is related to migration from LB3, please move it to `docs/site/migration` directory.

I see we already have `docs/site/migration/boot-scripts.md`, what is the relation between that document and this new page? Should you perhaps add a link from `6. Running boot scripts` to that migration guide?

I think that's a general rule you should follow - for each section (LB4 boot step), add a link to the relevant migration guide.

docs/site/LB3-vs-LB4-booting.md
hacksparrow: So refer to `Application` of  `@loopback/core` instead? We need to tell where the app configuration is applied as part of the booting process.

docs/site/LB3-vs-LB4-booting.md
hacksparrow: We don't have anything in-depth, AFAIK. The closest is https://loopback.io/doc/en/lb4/Booting-an-Application.html#controller-booter, added the link.

docs/site/sidebars/lb4_sidebar.yml
hacksparrow: > I see we already have docs/site/migration/boot-scripts.md, what is the relation between that document and this new page?

The new page describes everything related to booting, `docs/site/migration/boot-scripts.md` is just about boot scripts. Added link.

docs/site/LB3-vs-LB4-booting.md
hacksparrow: @raymondfeng I have updated accordingly.

docs/site/LB3-vs-LB4-booting.md
hacksparrow: Maybe not worth getting into all those details for something that we are moving away from. Especially since the corresponding step in LB4 is an `import`.

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,148 @@
...
+permalink: /doc/en/lb4/LB3-vs-LB4-booting.html
+---
+
+## Differences between LoopBack 3 and LoopBack 4 booting process
+
+The booting process in LoopBack 3 is handled by
+[loopback-boot](https://github.com/strongloop/loopback-boot),
+which performs the following tasks:
+
+### In LoopBack 3
```
jannyHou, line 15: I feel this paragraph 
> The booting process in LoopBack 3 is handled by
[loopback-boot](https://github.com/strongloop/loopback-boot),
which performs the following tasks:

should be under ### In LoopBack 3, instead of before it.

hacksparrow: Indeed.

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,151 @@
...
+
+{% include tip.html content="For all the details about booting in LoopBack 4,
+refer to the
+[booting documentation](./Booting-an-Application.md)." %}
+
+### 1. Configuration of application settings
+
+There is no `config.json` in LoopBack 4. The application options are passed in
+the constructor of the
+[Application](./apidocs/apidocs.core.application.md) class.
```
jannyHou, line 75: it'd be better to include a sample code, or the link https://loopback.io/doc/en/lb4/Application.html#configuring-your-application

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+
+Datasources in LoopBack 3 are defined in a single `datasources.json` file. In
+LoopBack 4, datasources are defined as TypeScript classes that come with an
+associated JSON configuration file. They all reside in the `src/datasources`
+directory.
+
+{% include note.html content="We are moving away from JSON files for datasource
+configuration. Refer to
+[loopback-next#5000](https://github.com/strongloop/loopback-next/pull/5000)
+for the progress." %}
```
bajtos, line 89: The PR has been already landed, please update this section to reflect the current `master`.

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+
+`BootMixin` resolves the datasources, and attaches them to the application.
+Read more about datasource booter
+[here](./Booting-an-Application.md#controller-booter).
```
bajtos, line 93: Did you update all links after moving the file to `docs/site/migration`? I don't think this link is valid now.

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+
+Read more about LoopBack 4 datasource [here](./DataSources.md).
+
```
bajtos, line 96: Should we add a link to https://loopback.io/doc/en/lb4/migration-datasources.html too?

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+process in LoopBack 4; Controllers and Repositories are.
+
+Controllers and Repositories import models as regular TypeScript classes since
+they are required at compile time.
+
+{% include note.html content="Booters are LoopBack 4 artifacts that are used
+for modularly adding functionality to the framework. Controllers, Repositories,
+Datasources, Interceptors, etc., are added using booters, custom booters can
+be written for adding new functionality to the framework." %}
+
```
bajtos, line 109: Should we add a link to https://loopback.io/doc/en/lb4/migration-models-overview.html?

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+mandatory for developing LoopBack 4 projects.
+
+The following are the list of tasks that are performed in the LoopBack 3
+booting process, and their equivalent in LoopBack 4.
+
+{% include tip.html content="For all the details about booting in LoopBack 4,
+refer to the
+[booting documentation](./Booting-an-Application.md)." %}
+
+### 1. Configuration of application settings
```
nabdelgadir, line 71: Should this and the two below be fourth-level headings?

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+{% include note.html content="We are moving away from JSON files for datasource
+configuration. Refer to
+[loopback-next#5000](https://github.com/strongloop/loopback-next/pull/5000)
+for the progress." %}
+
+`BootMixin` resolves the datasources, and attaches them to the application.
+Read more about datasource booter
+[here](./Booting-an-Application.md#controller-booter).
+
+Read more about LoopBack 4 datasource [here](./DataSources.md).
```
nabdelgadir, line 95: Nitpick: `datasource` -> `datasources`

docs/site/migration/LB3-vs-LB4-booting.md
```diff
@@ -0,0 +1,153 @@
...
+
+{% include tip.html content="Follow our GitHub issues
+[#1293](https://github.com/strongloop/loopback-next/issues/1293)
+and
+[#2035](https://github.com/strongloop/loopback-next/issues/2035)
+to track the progress on supporting Express middlweware in LoopBack 4." %}
+
+If you want to mount an Express router in a LoopBack 4 application, you can use
+the
+[RestApplication.mountExpressRouter()](./apidocs/apidocs.rest.restapplication.mountexpressrouter.md)
```
nabdelgadir, line 125: Can we link to https://loopback.io/doc/en/lb4/Routes.html#mounting-an-express-router instead?

docs/site/migration/LB3-vs-LB4-booting.md
hacksparrow: Indeed. Thanks.

hacksparrow:
Thanks @bajtos, fixed them.

hacksparrow:
@nabdelgadir I have applied your feedback. Can you PTAL?

PR #5197, opened 2020-04-22
fix: remove memory backup file from example-passport-login tests

deepakrkris:
## Checklist

### DIFF
--- a/examples/passport-login/src/index.ts
+++ b/examples/passport-login/src/index.ts
@@ -7,6 +7,7 @@ import {ApplicationConfig} from '@loopback/core';
 import {ExpressServer} from './server';
 import * as path from 'path';
 import {oauth2ProfileFunction} from './authentication-strategies';
+import {RestApplication} from '@loopback/rest';
 
 export {ExpressServer};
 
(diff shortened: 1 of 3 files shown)


examples/passport-login/src/index.ts
```diff
@@ -83,6 +85,9 @@ export async function main() {
   } else {
     oauth2Providers = require('../oauth2-providers');
   }
-  const server: ExpressServer = await startApplication(oauth2Providers);
+  const server: ExpressServer = await startApplication(
+    oauth2Providers,
+    '../data/db.json',
```
raymondfeng, line 90: I think it still creates `examples/passport-login/data/db.json`. What about creating it under `examples/passport-login/dist/data/db.json` or `examples/passport-login/.sandbox/data/db.json`?

raymondfeng: The other option is to add `data/db.json` to `.gitignore`.

deepakrkris: it wouldn't use the file for the tests, because the test script calls startApplication() directly without the second parameter. Here, we have just moved the hardcoded memory file parameter to the main() method that is called only when doing an `npm start` from example root.

deepakrkris: @raymondfeng I have added data/db.json to .gitignore

deepakrkris: @raymondfeng the acceptance test will not change the db.json file anymore, since it doesnt pass in the json file path.
https://github.com/strongloop/loopback-next/blob/05370bcc4cc7707ffd018cd6fbfc925935b05a8d/examples/passport-login/src/__tests__/acceptance/passport-login.acceptance.ts#L43-L47

Issue #2023, opened 2018-11-13
Custom visual style of REST API Explorer


bajtos:
_This is a follow-up for https://github.com/strongloop/loopback-next/pull/2014 which introduced a new extension providing a self-hosted API Explorer._

As a user building a LoopBack application with a self-hosted API Explorer, I would like to change the visual style (colors, etc.) of the Explorer UI.

Implementation wise, this may require the following features:
- Allow the app developer to provide a different `index.html.ejs` template
- Allow the app developer to provide custom CSS/JS/other static assets as an alternative to default assets provided by swagger-ui.

We should review how [loopback-component-explorer](https://github.com/strongloop/loopback-component-explorer) is approaching this problem and use it as an inspiration.

Reference files:

- For LB3, you can replace the [loopback.Styles.css](https://github.com/strongloop/loopback-component-explorer/blob/master/public/index.html#L10) to customize the look, see file https://github.com/strongloop/loopback-component-explorer/blob/master/public/index.html
- The corresponding LB4 file: https://github.com/strongloop/loopback-next/blob/master/packages/rest-explorer/templates/index.html.ejs

Acceptance Criteria:

- [ ] Try modifying the theme by overriding the style files like [`swagger-ui.css`](https://github.com/strongloop/loopback-next/blob/master/packages/rest-explorer/templates/index.html.ejs#L8)
- [ ] Document how to do it.

aabou:
Hello @bajtos,
I'd like to customize the title of my app.
So, i tried to change it in my public/index.html. the change has been applied on my home page and not on my /explorer page!.
The title is-it a custom param (expl one change to all page)? and, have you any proposition to change it in the explorer please?
Thanks in advance.
;-)

bajtos:
@aabou thank you for the comment. At the moment, our API explore does not support customization - that's why this issue is still open!

My recommendation is to create your own fork of `@loopback/rest-explorer` and make any changes there. Just remember to change the package name to something else before you publish it to an npm registry.

The source code can be found here: https://github.com/strongloop/loopback-next/tree/master/packages/rest-explorer It's just few files integrating swagger-ui with LoopBack 4.

dougal83:
https://github.com/ostranme/swagger-ui-themes - Feeling blue looks nice if going for a default lookback feel.

jannyHou:
[...]

PR #5200, opened 2020-04-22
customize theme

jannyHou:
This is a demo PR for https://github.com/strongloop/loopback-next/issues/2023 - customizing visual style of REST API Explorer

It introduces a config property in `@loopback/rest-explorer`:

- swaggerThemeFile: provide the custom path of swagger ui theme file to override the default one

The themes fixtures are added for demo purpose:

- `todoApp/src/public/theme-newspaper.css`
- `todoApp/src/application.ts`
- ~`rest-explorer/src/css`: for demo purpose, and also in case we want to include them as popular theme options, suggested in https://github.com/strongloop/loopback-next/issues/2023#issuecomment-601090982.~

### How customization works

The swagger ui assets are kept inside module `node_modules/swagger-ui-dist`, and mounted to endpoint `/explorer` by

https://github.com/strongloop/loopback-next/blob/3674a99396cf80a64c295354ef4242dfdd2209d5/packages/rest-explorer/src/rest-explorer.component.ts#L39

The theme file is added in template by

https://github.com/strongloop/loopback-next/blob/3674a99396cf80a64c295354ef4242dfdd2209d5/packages/rest-explorer/templates/index.html.ejs#L8

Since the theme files are loaded from  `node_modules/swagger-ui-dist`, which we cannot change, the theme file can be provided in the public folder

### Try it out

To test the theme **provided inside app**, just checkout this branch and start the todo app, and go to explorer. You will see the newspaper theme:

Screen Shot 2020-04-22 at 5 12 17 PM

## Checklist

### DIFF
--- a/packages/rest-explorer/README.md
+++ b/packages/rest-explorer/README.md
@@ -44,14 +44,6 @@ this.bind(RestExplorerBindings.CONFIG).to({
 });
 ```
 
-_NOTE: The Explorer UI's visual style is not customizable yet. Our recommended
-solution is to create a fork of this module, make any style changes in the fork
-and publish the modified module under a different name. The
-[GitHub issue #2023](https://github.com/strongloop/loopback-next/issues/2023) is
-requesting a configuration option for customizing the visual style, please
-up-vote the issue and/or join the discussion if you are interested in this
-feature._
-
 ### Advanced Configuration and Reverse Proxies
 
 By default, the component will add an additional OpenAPI spec endpoint, in the
@@ -138,6 +130,47 @@ standards-based means of getting a working explorer. A custom solution would be
 needed in this situation, such as passing a non-standard header from your
 reverse proxy to tell the app the external path, and custom code in your app to
 make the app and explorer aware of this.
+
+### Customizing Swagger UI Theme
+
+The Explorer UI’s visual style can be customized by configuring the
+`swaggerThemeFile` property. Here is the steps to do it:
+
+First, provide your own Swagger-UI theme file in a public folder. For example,
+in the
+[Todo example](https://github.com/strongloop/loopback-next/tree/master/examples/todo)
+application:
+
+Its `/public` folder is set up as the default home page with url `/`. Copy a
+swagger theme file `theme-newspaper.css` to be under `/public`.
+
+Then configure the `swaggerThemeFile` field to be the relative path to home page
+as `/theme-newspaper.css`:
+
+```ts
+export class TodoListApplication extends BootMixin(
+  ServiceMixin(RepositoryMixin(RestApplication)),
+) {
+  constructor(options: ApplicationConfig = {}) {
+    // ...
+
+    // customize the swagger-ui
+    this.configure(RestExplorerBindings.COMPONENT).to({
+      // Keep the theme file in the `public` dir of the app
+      // If required create a dir and keep the file, just specify the path
+      swaggerThemeFile: '/theme-newspaper.css',
+    });
+
+    // ...
+  }
+}
+```
+
+When the application runs, the explorer template will load the
+`theme-newspaper.css` file as its theme.
+
+Here is a repository that contains popular Swagger-UI themes:
+https://github.com/ostranme/swagger-ui-themes.
 
 ## Contributions
 
--- a/packages/rest-explorer/src/rest-explorer.types.ts
+++ b/packages/rest-explorer/src/rest-explorer.types.ts
@@ -29,4 +29,7 @@ export type RestExplorerConfig = {
    * forwarded to the application behind the proxy.
    */
   useSelfHostedSpec?: false;
+
+  // Relative URL of the theme CSS file
+  swaggerThemeFile?: string;
 };
--- a/packages/rest-explorer/templates/index.html.ejs
+++ b/packages/rest-explorer/templates/index.html.ejs
@@ -5,7 +5,7 @@
 <head>
   <meta charset="UTF-8">
   <title>LoopBack API Explorer</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css">
+  <link rel="stylesheet" type="text/css" href="<%- swaggerThemeFile %>">
   <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
   <style>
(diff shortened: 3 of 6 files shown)


examples/todo/src/application.ts
```diff
@@ -24,6 +27,14 @@ export class TodoListApplication extends BootMixin(
     // Set up default home page
     this.static('/', path.join(__dirname, '../public'));
 
+    // customize the swagger-ui
+    this.bind(RestExplorerBindings.CONFIG).to({
```
bajtos, line 31: Can we use `app.configure()` instead?

/cc @raymondfeng

packages/rest-explorer/templates/index.html.ejs
```diff
@@ -5,7 +5,9 @@
 <head>
   <meta charset="UTF-8">
   <title>LoopBack API Explorer</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css">
+  <!-- customize the swagger-ui theme, the corresponding asset path is ./themes/-->
+  <!-- FIXME: the theme path should be provided as a template variable -->
+  <link rel="stylesheet" type="text/css" href="<%- swaggerThemePath %>">
```
bajtos, line 10: Should we perhaps add a new server endpoint that will serve the CSS file, so that the HTML file can keep using the same static path?

```suggestion
  <link rel="stylesheet" type="text/css" href="./swagger-theme.css">
```

The controller method implementing `swagger-theme.css` can use `swagerThemePath` to locate the file to read & return back in the response.

I guess my main concern is that with your current proposal, how are you going to ensure that `swaggerThemePath` is pointing to a file that is served by the static middleware?

packages/rest-explorer/src/rest-explorer.types.ts
```diff
@@ -29,4 +29,21 @@ export type RestExplorerConfig = {
...
+   * path to swagger-ui themes provided in `@loopback/rest-explorer`:
+   * '../../css'
+   * path to assets provided in app:
+   * '../../../../<asset_folder_name>'
+   */
+  assetDir?: string;
+  /**
+   * The custom swagger-ui theme file's path.
+   */
+  swaggerThemePath?: string;
```
bajtos, line 48: Oh, I see. The clients are expected to provide `assetDir` to be mounted via static middleware, and then configure `swaggerThemePath` to point to a file inside `assetDir`. Can you please clarify this information in the tsdoc comments and/or the documentation?

IIRC, in loopback-component-explorer, we were mounting the asset directory at `/`, thus allowing the users to overwrite any files provided by swagger-ui. (The important part of the trick is to mount swagger-ui files first and `assetDir` second.) Have you considered using this approach too? If we do that, then I think `swaggerThemePath` is no longer needed, because we can ask users to save their CSS as `swaggar-ui.css` inside `assetDir`. WDYT?

Either way, please add some tests to verify how these new config options are handled.

packages/rest-explorer/templates/index.html.ejs
bajtos: (I think my comment above can be discarded, but I am keeping here to show other possible approaches how to solve this feature.)

jannyHou:
> IMO, we should not provide those themes and always use LoopBack-style theme only.

However, I think it is a good idea to add a short section to module's documentation to show users where they can find other popular themes and how to use them with LoopBack.

👍 True on a second thought I also feel it's not proper to add the community themes in our module, but will definitely mention them in the documentation.

Thanks for the detailed feedback! I will add some tests.

hacksparrow:
Do we need to allocate an entire directory for this to work? Instead of introducing `assetDir`, how about we should look for the value defined in `swaggerThemePath` in the default `public` dir.

Lesser the configuration, the better.

jannyHou:
> Do we need to allocate an entire directory for this to work? Instead of introducing assetDir, how about we look for the value defined in swaggerThemePath in the default public dir.

Thank you @hacksparrow good point and that is the strategy I tried at the very beginning, but
the following line implies swagger-ui searches for assets from the running server, instead of local files:
https://github.com/strongloop/loopback-next/blob/3674a99396cf80a64c295354ef4242dfdd2209d5/packages/rest-explorer/src/rest-explorer.component.ts#L39

Is there a way to add the style file from local dir like `public`?

hacksparrow:
@jannyHou here is a demo of an implementation that requires only one new property `swaggerThemeFile` and the theme file can reside in the `public` dir of the app - https://github.com/strongloop/loopback-next/tree/explorer-theme-yaapa.

examples/todo/src/application.ts
jannyHou: 👍 refactored to be 
```ts
    this.configure(RestExplorerBindings.COMPONENT).to({
      swaggerThemeFile: '/theme-newspaper.css',
    });
```

packages/rest-explorer/templates/index.html.ejs
jannyHou: > how are you going to ensure that swaggerThemePath is pointing to a file that is served by the static middleware

It's a valid question :)
The user needs to make sure the `swaggerThemeFile` points to the right css file. 
E.g. todo app has `/public` folder set up as the default home page with url `/`. To load the css file in `/public/theme-newspaper.css`, use

```ts
{swaggerThemeFile: '/theme-newspaper.css'}
```

packages/rest-explorer/src/rest-explorer.types.ts
jannyHou: added test and @hacksparrow helps simplified the code, now we only need one config property :)

jannyHou:
@hacksparrow Thank you for helping simplify the PR 🎉 and appreciate @bajtos 's feedback!
 Tests and doc added, PTAL again.

packages/rest-explorer/README.md
```diff
@@ -139,6 +131,46 @@ needed in this situation, such as passing a non-standard header from your
...
+      swaggerThemeFile: '/theme-newspaper.css',
+    });
+    this.component(RestExplorerComponent);
+
+    // ...
+  }
+}
+```
+
+When the application runs, the explorer template can load the
```
hacksparrow, line 168: "can load" -> "will load"?

jannyHou:
Merged. And Thanks for @dougal83 's suggestion in https://github.com/strongloop/loopback-next/issues/2023#issuecomment-601090982 again 🎉 Added https://github.com/ostranme/swagger-ui-themes as popular community themes in the new guide.

PR #5201, opened 2020-04-23
feat: mongoose extension

mschnee:
Inspired by #4794 and by a project I am working with that has a large number of existing Typegoose Mongoose schemas/models already defined.

This is a very rough WIP, and more of a weekend project to start a discussion. 

## Goals
- Easily take your existing mongoose schemas, and use them in loopback without creating a specialized connection provider of your own.
- `@inject()` the connection-models you need when you need them, instead of `require()`ing them in the global scope, which is common in lots of mongoose/typegoose examples.
- Keep it simple.

## TODO
- Tests need to be greatly expanded beyond the simple integration test
- The `TestMongooseComponent` is messy, and I'm not sure if it's a good pattern.
- What's the best way to pass configuration when an application has hundreds of schemas?  `application.ts` seems like the wrong place.

## Checklist
👉 [Read and sign the CLA (Contributor License Agreement)]

### DIFF
--- a/extensions/mongoose/README.md
+++ b/extensions/mongoose/README.md
@@ -1,0 +1,160 @@
+# @loopback/mongoose
+
+## Stability: ⚠️Experimental⚠️
+
+> Experimental packages provide early access to advanced or experimental
+> functionality to get community feedback. Such modules are published to npm
+> using `0.x.y` versions. Their APIs and functionality may be subject to
+> breaking changes in future releases.
+
+[![LoopBack](<https://github.com/strongloop/loopback-next/raw/master/docs/site/imgs/branding/Powered-by-LoopBack-Badge-(blue)-@2x.png>)](http://loopback.io/)
+
+## ⚠️Warnings⚠️
+
+As of `0.0.1` of this extension, the type definitions of `@types/mongoose` are
+several signifigant minor versions out of date: it supports mongoose `5.7.12`,
+and the latest stable version of mongoose is `5.9.10`
+
+## Using your existing Mongoose Schemas in Loopback
+
+Using mongoose to devine your mongodb schemas is relatively straightforward,
+though at this early stage it's still a number of steps:
+
+- Import the component, and it's binding:
+  `import {LoopbackMongooseComponent, MongooseBindings} from '@loopback/mongoose'`
+
+- Register your configuration (see below)
+  ```ts
+  this.configure(MongooseBindings.COMPONENT).to(configuration);
+  ```
+- Add the component to your application
+  ```ts
+  this.component(LoopbackMongooseComponent);
+  ```
+- Finally, `@inject` your Entity Models for use
+  ```ts
+  public async MyService(@inject('loopback-mongoose-component.model.ModelName') ModelName) {
+    const result = await ModelName.findOne({});
+  }
+  ```
+
+## Configuration
+
+There are three pieces of information that the Mongoose Component needs in order
+to create a connection to your database and set up the mongoose models:
+
+- the `uri` of the connection
+- an array of Schemas
+- optionally, an array of Discriminator Schemas
+
+The configuration looks like the following:
+
+```ts
+ this.configure(MongooseBindings.COMPONENT).to({
+   uri: 'mongodb://localhost:27017',
+   connectionOptions: {} // anything valid to mongoose.createConnection(),
+   schemas: [
+     {
+       bindingKey: 'my-custom.binding.key',
+       schema: myImportedSchema,
+       name: 'MyModel'
+      }
+   ],
+   discriminators: [
+     {
+       bindingKey: 'my-custom.discriminator.binding.key',
+       modelKey: 'my-custom.binding.key',
+       schema: myImportedDiscriminatorSchema,
+       value: 'MyModelDiscriminator'
+     }
+   ]
+ })
+```
+
+For multiple connections, pass an array instead of an object:
+
+```ts
+this.configure(MongooseBindings.COMPONENT).to([
+  {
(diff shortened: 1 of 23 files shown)


extensions/mongoose/README.md
```diff
@@ -0,0 +1,160 @@
...
+- the `uri` of the connection
+- an array of Schemas
+- optionally, an array of Discriminator Schemas
+
+The configuration looks like the following:
+
+```ts
+ this.configure(MongooseBindings.COMPONENT).to({
+   uri: 'mongodb://localhost:27017',
+   connectionOptions: {} // anything valid to mongoose.createConnection(),
```
raymondfeng, line 55: We might consider to use settings from a datasource.

extensions/mongoose/README.md
```diff
@@ -0,0 +1,160 @@
...
+   schemas: [
```
raymondfeng, line 56: Maybe it's better to declare `mongoose.schemas` as an extension point to register Mongoose schemas as extensions.

extensions/mongoose/README.md
```diff
@@ -0,0 +1,160 @@
...
+     {
+       bindingKey: 'my-custom.binding.key',
+       schema: myImportedSchema,
+       name: 'MyModel'
+      }
+   ],
+   discriminators: [
```
raymondfeng, line 63: Each schema/model itself can be an extension point for its discriminators.

extensions/mongoose/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+```ts
+this.configure(MongooseBindings.COMPONENT).to([
+  {
+    uri: 'mongodb://localhost:27017',
+    schemas: [...]
+  },
+  {
+    uri: 'mongodb://127.0.0.12:27017',
+    schemas: [...]
```
raymondfeng, line 84: This is a bit similar as how we make `AJV` configurable and extensible.  In https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/validation/ajv-factory.provider.ts, we use `@inject` (or `@config`) for options, keywords, and formats. Both keywords and formats can be either passed in via options, or their own extension points.

mschnee: The extension system/pattern is going to be new to me: is the linked AVJ code a good example to base off of as a starting point?

The question mark I have is that schemas are orthogonal to connections: a model object represents a schema defined on a specific connection, so schemas can be used across multiple connections.  I've seen systems that export all the schemas from an `schems/index.js` file, and then have a separate connection that only uses a handful of them to create models specific to that connection.

And then there's having both mongoose _and_ typegoose models (two different extensions) use the same mongoose connection.

My first thought is to change the `uri` configuration to be 
```ts
uriOrConnection: string | mongoose.Connection
```
to allow you to pass an existing connection instead.

Second thought: having a connection manager separate from mongoose/typegoose components but used by both.

Third thought: try to use the mongoose/mongodb internals to determine if the given uri already references a connection.  Unlikely since nothing I've seen in the documentation can be used for an exact match- we'd have to make the connection first to determine if we're connecting to the same database.

These patterns are new to me- I'd love to get some direction :)

raymondfeng: @mschnee I have updated the TypeORM extension to leverage the ideas for more flexible configuration. See https://github.com/strongloop/loopback-next/blob/typeorm/extensions/typeorm/src/services/connection-manager.ts.

As we show in https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/validation/ajv-factory.provider.ts as well, the configuration be injected. Extensions can contribute additional pieces to the configuration.

We can build connection option adapter for data source settings such as existing mongodb datasources can be discovered and referenced for `mongoose` usage.

dhmlau:
@mschnee, do you have a chance to take a look at @raymondfeng's comments above? Thanks.

extensions/mongoose/README.md
bajtos: In https://github.com/strongloop/loopback-next/pull/5801, we are moving to a different direction than presented in https://github.com/strongloop/loopback-next/blob/typeorm/extensions/typeorm/src/services/connection-manager.ts. Instead of keeping a list of connection options at LoopBack side, we are introducing a new application API for registering a new connection and then we keep the actual connection object around.

https://github.com/strongloop/loopback-next/blob/d714638984da4deced7c36b4eb25457492759c69/packages/typeorm/src/mixin.ts#L37-L44

In this new design, users can use regular `@inject`, `@inject.getter` etc. decorators to receive connection instances.

> Second thought: having a connection manager separate from mongoose/typegoose components but used by both.

Can we use LoopBack's context as the connection manager, the same way as `@loopback/repository` uses application context to keep track of datasources?

> And then there's having both mongoose and typegoose models (two different extensions) use the same mongoose connection. 

IIUC, `typegoose` is using `mongoose` under the hood, expecting app authors to pick & install a compatible version (`mongoose` is in `peerDependency`). I think that means the LoopBack extension should be using `mongoose` APIs to create connections and then find a way how to use those connections with typegoose models?

> The question mark I have is that schemas are orthogonal to connections: a model object represents a schema defined on a specific connection, so schemas can be used across multiple connections.  I've seen systems that export all the schemas from an `schems/index.js` file, and then have a separate connection that only uses a handful of them to create models specific to that connection.

`@loopback/repository` is already designed this way:

- a datasource represents a MongoDB connection
- a model class describes a schema
- a repository class binds a model with a datasource
- it's possible for a single model to be attached to multiple datasources

Can you perhaps use a similar design for Mongoose/Typegoose too?

dhmlau:

```
git commit --amend -s
git push --force-with-lease
```

Issue #5196, opened 2020-04-22
Create model relations between local and remote models


rivoheinsalu:
## Suggestion

In LB3 strong-remoting let's users easily define model relations over multiple microservices. This gives users convenient way to query data with single query over multiple services without having to manually map relations.

Given the more open nature of LoopBack 4 it is not anymore possible to define remote relations on Entity but I think it is worth porting strong-remoting to support LB4 with more strict structure making user responsible for defining correct interfaces.

https://loopback.io/doc/en/lb3/Strong-Remoting.html

## Examples

Extending HasMany relation example https://loopback.io/doc/en/lb4/HasMany-relation.html
User has two microservices:
1. Customer
2. Order

It is possible to define HasMany relation on Customer Entity with remote connector
`@hasMany(() => Order, {connector: 'remote', model: 'Order'})`

Giving user ability to query customers with orders via single endpoint using
`customerRepo.find({include: [{relation: 'orders'}]});`

bajtos:
Thank you for starting this discussion. I am not sure if adding strong-remoting to LoopBack 4 is the right solution. I would like to reframe the discussion a bit and start with describing the problem to solve.

IIUC, you would like to let your LB4 application to have a regular Repository access to models that are persisted by another LoopBack application, which was historically done via the remote connector.

The first question that comes to my mind - what LoopBack version is used by the remote application? Is LB4-to-LB4 remote good enough, or do you also need LB4-to-LB3 remote access?

rivoheinsalu:
>  Is LB4-to-LB4 remote good enough, or do you also need LB4-to-LB3 remote access?

Initially would be good to have LB4-to-LB3 as migrating more than 10 LB3 apps to LB4 one-by-one would give good migration path. But at the end LB4-to-LB4 would be the way.

bajtos:
@rivoheinsalu I have an idea how to implement a solution allowing LB4 application to talk to both LB3 and LB4 applications when resolving relations. I'd like to write a quick PoC today 🤞 

I see how it could be useful to access LB4 models from a LB3 application (LB3 app having a LB4 app acting as a remote). I think it may be possible to extend the ideas from my PoC to implement a LB3 connector acting as a remote connector for LB4, will see.

raymondfeng:
The newly released `lb4 openapi —client` can generate strongly-typed service proxies for REST APIs that expose swagger/openapi spec.

bajtos:
[...]

PR #5222, opened 2020-04-24
An experimental extension to access models persisted by a remote LoopBack application

bajtos:
Add a new experimental extension providing `RemoteCrudRepository` class implementing CRUD operations on top of our OpenAPI connector, effectively delegating persistence to a remote LoopBack application.

**IMPORTANT: I am intentionally keeping the implementation small and limited in features. My goal is to enable the community to help us implement this extension, I don't have bandwidth to do all the work myself.**

Implements https://github.com/strongloop/loopback-next/issues/5196

## Basic use

See README for information on how to use the module, below is a trimmed outline to give you a high-level idea.

Initial setup

1. Create a new DataSource using [openapi connector](https://github.com/strongloop/loopback-connector-openapi), configure it to connect to your remote LoopBack application.

2. Copy model files from your remote LoopBack application to your local project. We
will need these model files for type information.

3. Create a new file `src/repositories/remote-crud.repository.base.ts` to add `RemoteCrudRepository` to the list of base classes offered by `lb4 repository` command.

Setup a remote model

1. Run `lb4 repository` to create a new repository class for accessing the remote
model. Pick `RemoteCrudRepository` as the base class.

2. Optionally, run `lb4 relation` to establish relations between local and remote models.

3. You can also run `lb4 controller` to expose your remote model via a REST API of your local app, effectively making the local app a (smart) proxy.

## TODO

- [ ] Improve TSDoc comments, especially for protected members
- [ ] Add tests to verify that HasMany relation from a local model to a remote model works as expected (inclusion resolvers work in queries, relation repository APIs can be used to modify related model instances).
- [ ] Add a short note to our migration guide, point loopback-connector-remote users to the new package and issue #5196
- [ ] Add a test to verify how HTTP error responses are handled
- [ ] Add comments to tell code coverage to ignore not implemented methods

## Out of scope

See README for the full list of things that are out of scope of this initial version. Most notably:

- Other CRUD methods than `find` and `create`
- LoopBack 3.x support
- Robust checks to ensure we detect mismatch between repository's expectations vs the actual remote API and report helpful error messages.

## Checklist


bajtos:
@rivoheinsalu here you go! Please let me know if my proposal would solve your use case? If we agree on this direction, then I can give you pointers so that you can continue the implementation with adding support for accessing LB 3.x models from LB4.

As for accessing LB4 models from a LB3 app, I think we will need to create a new connector or perhaps modify one of the existing connectors (perhaps loopback-connector-rest?) to provide CRUD Model API expected by LB3.

extensions/remote-repository/README.md
```diff
@@ -0,0 +1,108 @@
...
+### Initial setup
+
+Create a new DataSource using
+[openapi connector](https://github.com/strongloop/loopback-connector-openapi),
+configure it to connect to your remote LoopBack application.
+
+**IMPORTANT: You must configure the datasource to use named parameters. Answer
+the prompt _Use positional parameters instead of named parameters?_ with _n_.**
+
+Copy model files from your remote LoopBack application to your local project. We
```
agnes512, line 29: Does this mean to copy( or import)  LB3/LB4 model from the remote LoopBack app to the local project in a LB4 form?

extensions/remote-repository/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Nice to have: support KeyValue persistence style.
+
+- Nice to have: document how custom subclasses can overwrite protected methods
+  to customize the way how remote endpoints are invoked. Few examples:
+
+  - `getEndpointName` allows users to change the way how model methods are
+    mapped to remote endpoints.
+
+  - `executeEndpoint` allows users to tweak the request parameters before the
```
agnes512, line 86: 👍

raymondfeng:
@bajtos It's a great start. I have a simpler implementation of RemoteCrudRepository derived as follows:

1. Scaffold a LB app with a Crud controller for `Note`
2. Create a `openapi` datasource pointing to `http://localhost:3000/openapi.json` 
3. Use `lb4 openapi --client` command to generate a service proxy.
4. Replace `Note` with a generic type and use built-in types such as `Where` and `Filter`.

Here is what I get as the simple implementation:
```ts
import {Constructor, inject, Provider} from '@loopback/core';
import {
  Count,
  Entity,
  Filter,
  FilterExcludingWhere,
  juggler,
  Where,
} from '@loopback/repository';
import {getService} from '@loopback/service-proxy';
import createError from 'http-errors';

/**
 * The service interface is generated from OpenAPI spec with operations tagged
 * by NoteController.
 */
export interface RemoteEntityCrudRepository<
  E extends Entity,
  ID,
  Relations extends object = {}
> {
  /**
   * @param where
   * @returns Note model count
   */
  count(where: Where<E>): Promise<Count>;

  /**
   * @param data
   * @param id
   */
  replaceById(data: E, id: string): Promise<unknown>;

  /**
   * @param data
   * @param id
   */
  updateById(data: Partial<E>, id: string): Promise<unknown>;

  /**
   * @param id
   * @param filter
   * @returns Note model instance
   */
  findById(id: string, filter: FilterExcludingWhere<E>): Promise<E & Relations>;

  /**
   * @param id
   */
  deleteById(id: string): Promise<void>;

  /**
   * @param data
   * @returns Note model instance
   */
  create(data: Omit<E, 'id'>): Promise<E>;

  /**
   * @param data
   * @param where
   * @returns Note PATCH success count
   */
  updateAll(data: Partial<E>, where: Where<E>): Promise<Count>;

  /**
   * @param filter
   * @returns Array of Note model instances
   */
  find(filter: Filter<E>): Promise<(E & Relations)[]>;
}

export function defineRemoteNoteRepositoryProviderClass<
  E extends Entity,
  ID,
  Relations extends object = {}
>(
  controllerName: string,
  dataSourceKey: string,
): Constructor<Provider<RemoteEntityCrudRepository<E, ID, Relations>>> {
  class RemoteNoteRepositoryProvider<
    E extends Entity,
    ID,
    Relations extends object = {}
  > implements Provider<RemoteEntityCrudRepository<E, ID, Relations>> {
    constructor(
      @inject(dataSourceKey)
      protected dataSource: juggler.DataSource,
    ) {}

    async value(): Promise<RemoteEntityCrudRepository<E, ID, Relations>> {
      const service = await getService<{
        apis: Record<string, Record<string, Function>>;
      }>(this.dataSource);
      const proxy = service.apis[controllerName];
      return cast(proxy);
    }
  }
  return RemoteNoteRepositoryProvider;
}

/**
 * Type for method names of NoteService
 */
type ServiceMethod = keyof RemoteEntityCrudRepository<Entity, unknown>;

/**
 * Map method names to operation ids for NoteService
 */
const METHOD_OPERATION_MAPPING: Record<ServiceMethod, string> = {
  count: 'count',
  replaceById: 'replaceById',
  updateById: 'updateById',
  findById: 'findById',
  deleteById: 'deleteById',
  create: 'create',
  updateAll: 'updateAll',
  find: 'find',
};

/**
 * Cast the proxy to NoteService
 */
function cast<Note extends Entity, ID, Relations extends object>(
  proxy: Record<string, Function>,
): RemoteEntityCrudRepository<Note, ID, Relations> {
  const methods: Partial<Record<ServiceMethod, Function>> = {};
  let m: ServiceMethod;
  for (m in METHOD_OPERATION_MAPPING) {
    const op = METHOD_OPERATION_MAPPING[m];
    methods[m] = async (...args: unknown[]) => {
      const response = await proxy[op](...args);
      /*
       * {
       *   url,
       *   method,
       *   status,
       *   statusText,
       *   headers, // See note below regarding headers
       *
       *   text,    // The textual content
       *   body,    // The body object
       * }
       */
      if (response.status < 400) {
        return response.body;
      }
      const err = createError(response.status, response.statusText, response);
      throw err;
    };
  }
  return methods as RemoteEntityCrudRepository<Note, ID, Relations>;
}
```

I think we can use the approach with the ability to create an openapi datasource from the url.

bajtos:
> It's a great start. I have a simpler implementation of RemoteCrudRepository derived as follows:
> 
> 1. Scaffold a LB app with a Crud controller for `Note`
> 2. Create a `openapi` datasource pointing to `http://localhost:3000/openapi.json`
> 3. Use `lb4 openapi --client` command to generate a service proxy.
> 4. Replace `Note` with a generic type and use built-in types such as `Where` and `Filter`.

Thank you for showing me the alternative solution you have in mind :+1: Let's discuss the pros and cons of these two approaches to find the best one.

Here are the advantages I see in my proposal:

1. The new RemoteRepository is designed to work with the existing CLI commands. User can keep using the commands they are familiar with, e.g. `lb4 repository` and `lb4 controller`, there is no need to learn new concepts.

2. Because the model files are shared between the local app and the remote service, it's easy to share computed properties (e.g. `get fullName() { return `${this.firstName} {$this.surname}`; }`).

3. There are plenty extension points to allow users to tweak the way how CRUD methods are mapped to remote endpoints. I think this is a critical aspect we must keep in mind, because in my experience from LB3 days, most application DO NOT use the vanilla CRUD REST API provided by the framework. 

    Few examples of what customizations are enabled by my design:

    - Tweak the data before it's sent to the server (e.g. add or remove properties)
    - Tweak the data returned by the server before it's returned from the repository method
    - Change the way how Repository methods (`find`, `create`) are mapped to remote endpoints (e.g. `find` could be implemented as `${modelName}Controller_query`).
   - Change the way how local model names are mapped to remote endpoint prefixes (e.g. `UserRepository.find()` may invoke `UserAdminController_find`).

    What's also important, my design makes it relatively easy to discover and implement these customization points, because they are all based on overriding protected class methods, which is a concept we already use in other places.

4. It is possible (and relatively easy) to adopt `RemoteRepository` to talk to a non-LoopBack remote servers. In some cases, it may be enough to subclass `RemoteRepository` and use some of the customization points to change the way how CRUD operations are mapped to endpoints. In other case users can copy & paste the entire class and then re-implement the methods as they like. What's important, the rest of the consumer workflow remains the same. User will install the custom repository instead of `@loopback/remote-repository` and then continue using `lb4 repository` and other commands.

I am aware of one disadvantage only: it requires a bit of boilerplate code in `RemoteRepository` class that we will have to write and maintain.

@raymondfeng can you please describe how do you envision supporting the above aspects in your proposed design based on `lb4 openapi --client`? What advantages do you see in your proposal when compared to mine?

Could you please clarify the steps 3 and 4 from your proposal?

> 3. Use `lb4 openapi --client` command to generate a service proxy.

How are the return types of the generated service proxy methods described? Will they use local application's `Note` model and `Note & NoteRelations` type, or will they use types generated for OpenAPI schemas (e.g. `Note` and `NoteWithRelations`)?

> 4. Replace `Note` with a generic type and use built-in types such as `Where` and `Filter`.

Who is going to perform this step? Let's say the user wants to add RemoteRepository for 10 models. Are they going to repeat the step 4 from your recipe for each of those models? That would be a poor developer experience IMO. Are you suggesting to write a new CLI command and/or modify `lb4 openapi --client` to make these replacement automatically?

Maybe I am don't fully understand what you are proposing. My impression is that your proposal offers a suboptimal developer experience, both in the terms of what's the process of adding code to access a remote model, but also in the terms of the amount of new concepts our users would have to learn to be able to use (and troubleshoot!) this new feature.

extensions/remote-repository/README.md
bajtos: Great question!

If the remote application is written in LoopBack 4, then the model files can be copied over or perhaps shared via an npm package.

If the remote application is written in LoopBack 3, then there are few options available:
1. Import the model(s) from the remote LB3 app to the local LB4 app via `lb4 import-lb3-models`.
2. Use a tool like `lb4 openapi --client` to generate LB4 models (or interfaces?) describing the shape of model data from the Swagger spec provided by the LB3 app.

In both cases, it's important to consider how to update local models after the remote application has changed them? For example, when the remote application adds a new property `updatedAt`, how to update the local model to include that new property too?

- When LB4 models are shared via a copy, then we need to copy the new version again.
- When LB4 models are shared via a npm package, then we need to update local project's dependencies.
- When LB3 models are imported via `lb4 import-lb3-models`, we need to re-run that command again.
- When LB3 models are imported via `lb4 openapi --client`, we need to re-run the import process.

bajtos:
@raymondfeng few more questions regarding your proposal:

1. What project layout conventions do you envision for your repository providers? (In my proposal, we use `src/repositories/{name}.repository.ts`.) How is `@loopback/boot` and `@loopback/cli` going to recognize them (e.g. `lb4 relation` must be aware of remote repositories).

2. How can users add local helper methods to their repository classes? An example for my approach:

     ```ts
    export class TodoRepository extends RemoteRepository<
    Todo,
    typeof Todo.prototype.id,
    TodoRelations
    > {
      findByTitle(title: string) {
        return this.find({where:{title}});
      }
    }
    ```

3. How can app developers apply a mixin on the remote repository class? See https://loopback.io/doc/en/lb4/migration-models-mixins.html#creating-a-custom-model-method-and-remote-model-method-mixin. An example for my approach:

    ```ts
    export class TodoRepository extends FindByTitleRepositoryMixin<
      Todo,
      Constructor<
        RemoteRepository<Todo, typeof Todo, TodoRelations>
      >
    >(RemoteRepository) {
      // findByTitle was already added by the mixin
    }
    ```

raymondfeng:
@bajtos I don't think we are proposing completely different solutions. Let me clarify:

1. The main use case is to allow one LoopBack 4 application to access REST APIs exposed by another LoopBack 4 application with minimal coding using `openapi` connector behind the scenes. If the `Controller` follows a known pattern such as `CrudRest`, we can expose the interface to the client as a service proxy, such as `RemoteCrudRepository`. Otherwise, the client side can use weakly-typed service proxy, self-defined interface, or leverage `lb4 openapi --client` to generate the strong-typed service proxy.

2. IMO, the idea is very much like the client-side `rest-crud` to define strongly-typed service proxy from a model and url to the OpenAPI spec. For example, if the server-side LB4 app exposes `Note` and `Customer` REST CRUD APIs (using `lb4 controller` or `rest-crud`), the client application should be able to access them as `RemoteCrudRepository<Note>` and `RemoteCrudRepository<Customer>`. The model definitions are shared/reused. It's worth pointing out is that `RemoteCrudRepository` is a service proxy to `CrudController` (which is backed by `EntityCrudRepository`).

3. I was proposing a default implementation of `RemoteCrudRepository` from `lb4 openapi --client` as it generates an interface reflecting the operations of a `CrudController`. It's a one-time thing that lead my code. `lb4 openapi` is no longer needed if we use the default implementation as I proposed.

4. To allow custom Controller, we can use `lb4 openapi --client` to generate the strongly-typed service interface without overriding the models - either press `n` to skip such models or add a prompt to select models. BTW, I have a PR https://github.com/strongloop/loopback-next/pull/5248 to make the client side code generation even closer.

bajtos:
> The main use case is to allow one LoopBack 4 application to access REST APIs exposed by another LoopBack 4 application with minimal coding using `openapi` connector behind the scenes. If the `Controller` follows a known pattern such as `CrudRest`, we can expose the interface to the client as a service proxy, such as `RemoteCrudRepository`. Otherwise, the client side can use weakly-typed service proxy, self-defined interface, or leverage `lb4 openapi --client` to generate the strong-typed service proxy.

I am coming to this feature from a different point of view.

The main use case I have in mind is to allow LoopBack 4 application to access models regardless of where they are persisted, whether they are stored in a local database (e.g. via `loopback-connector-mysql`) or at a remote LoopBack server.

I see `loopback-connector-openapi` just as an implementation detail that makes it easier for `RemoteCrudRepository` to call appropriate remote endpoints. In theory, the repository can use `loopback-connector-rest` instead or even support other transports like `gRPC` and `GraphQL` in the future.

>  It's worth pointing out is that `RemoteCrudRepository` is a service proxy to `CrudController` (which is backed by `EntityCrudRepository`).

I agree that internally, `RemoteCrudRepository` is indeed implemented as a service proxy to `CrudController`.

However, at conceptual level, I want to treat `RemoteCrudRepository` as a different implementation of Repository concept, and keep the service-proxy aspect as an internal implementation details.

- We already have `DefaultCrudRepository` that's using loopback-datasource-juggler and connectors as the implementation.
- We also have an unfinished `CrudRepositoryImpl` that's calling connectors directly.
- Now I am introducing `RemoteCrudRepository` as another repository flavor.
- Other repository flavors are possible, e.g. a repository backed by TypeORM.

From my point of view, it's crucial that all of these implementation provide the same Repository (`CrudRepository`) interface and can be used interchangeably. This is the same design we are using in LB3 and that allowed us to build cool features like offline data access & change replication. While I am not saying we should be building offline data access to LB4, at least not now, I also think we should keep the good parts of LB3.

> I was proposing a default implementation of `RemoteCrudRepository` from `lb4 openapi --client` as it generates an interface reflecting the operations of a `CrudController`. It's a one-time thing that lead my code. `lb4 openapi` is no longer needed if we use the default implementation as I proposed.

I am still slightly confused about what exactly you are proposing. Are you saying that I should replace my code with the code from your comment https://github.com/strongloop/loopback-next/pull/5222#issuecomment-619428352?

I see the following downsides of your version:

1. `interface RemoteEntityCrudRepository` - as I explained in my comments above, I don't want custom repository interfaces, I want the remote repository to implement the standard `CrudRepository` interface as implemented by other repository classes. Secondly, I feel it would be adding too much boiler-plate-like code to client applications.

2. I find the implementation based on `defineRemoteNoteRepositoryProviderClass`, `METHOD_OPERATION_MAPPING` and `cast` as more difficult to understand when compared to the implementation proposed in this pull request.

3. Most importantly, I don't see how can applications add mixins and additional helper methods to remote repositories when using your approach? See https://github.com/strongloop/loopback-next/pull/5222#issuecomment-619929293

raymondfeng:
@bajtos I guess you want to have full implementation of `EntitiyCrudRepository` over REST APIs. In that case, I think it should be easy to implement it with the service proxy interface I have as it's reflected from the CrudController.

bajtos:
> I guess you want to have full implementation of `EntitiyCrudRepository` over REST APIs.

Yes, exactly!

>  In that case, I think it should be easy to implement it with the service proxy interface I have as it's reflected from the CrudController.

I tried to explain in https://github.com/strongloop/loopback-next/pull/5222#issuecomment-620421265 why I believe an implementation based on service proxy is not good enough.

Re-posting the relevant part:

1. `interface RemoteEntityCrudRepository` - as I explained in my comments above, I don't want custom repository interfaces, I want the remote repository to implement the standard `CrudRepository` interface as implemented by other repository classes. Secondly, I feel it would be adding too much boiler-plate-like code to client applications.

2. I find the implementation based on `defineRemoteNoteRepositoryProviderClass`, `METHOD_OPERATION_MAPPING` and `cast` as more difficult to understand when compared to the implementation proposed in this pull request.

3. Most importantly, I don't see how can applications add mixins and additional helper methods to remote repositories when using your approach? Reposting from https://github.com/strongloop/loopback-next/pull/5222#issuecomment-619929293:

    How can users add local helper methods to their repository classes? An example for my approach:

     ```ts
    export class TodoRepository extends RemoteRepository<
    Todo,
    typeof Todo.prototype.id,
    TodoRelations
    > {
      findByTitle(title: string) {
        return this.find({where:{title}});
      }
    }
    ```

    How can app developers apply a mixin on the remote repository class? See https://loopback.io/doc/en/lb4/migration-models-mixins.html#creating-a-custom-model-method-and-remote-model-method-mixin. An example for my approach:

    ```ts
    export class TodoRepository extends FindByTitleRepositoryMixin<
      Todo,
      Constructor<
        RemoteRepository<Todo, typeof Todo, TodoRelations>
      >
    >(RemoteRepository) {
      // findByTitle was already added by the mixin
    }
    ```

@raymondfeng I am afraid it's not clear to me from your comments, what changes are you proposing and how can I move this pull request forward. Are you just trying to offer suggestions on alternatives to consider? I prefer to continue in the original direction for the reasons explained above, are you against that? (I don't mind the framework to support both my approach and generated-service-proxy approach in parallel, if that's what you are looking for.)

raymondfeng:
What I had in mind was to implement RemoteCrudRepository using the strongly-typed service proxy I came up. Quite a few improvements went to `openapi` connector. The method mapping is no longer required. It also allows `transformResponse` custom function at datasource level to extract body and headers. See https://github.com/strongloop/loopback-connector-openapi/blob/master/README.md.

I'm not very keen on https://github.com/strongloop/loopback-next/blob/67c61874758cb5409d43481876b691b85b3a8434/extensions/remote-repository/src/repositories/remote-crud-repository.ts#L215, which seems to be a duplicate effort of wrapping service proxies.

bajtos:
I started this pull request with the intention to provide a place where LoopBack users can collaborate together to build a good remote repository implementation. I though it would be a quick task that could be done even outside of our main milestone priorities. As can be seen from the list of TODO items in the PR description and the discussion above, finishing this pull request is not trivial.

In that light, I have decided to abandon this work.

I am encouraging others to pick the code where I left it and bring it to a usable state. 
- You can open a new pull request to add the extension to loopback-next monorepo. This way it will be maintained by us going forward, but you will need to work with loopback-next maintainers to reach consensus about the proposed design & implementation.
- Alternatively, you can create a new GitHub repository to host the extension. This way you can get done faster, because you won't need to worry about getting consensus with loopback-next maintainers. On the other hand, it will be up to you to maintain the extension going forward.

bajtos:
For anybody discovering this pull request and interested in making this feature happen - please upvote https://github.com/strongloop/loopback-next/issues/5196 and/or join the discussion there.

PR #5231, opened 2020-04-24
fix(repository-json-schema): avoid title inheritance

jannyHou:
link to https://github.com/strongloop/loopback-next/issues/5088#issuecomment-617696920

https://github.com/strongloop/loopback-next/pull/5108 fixed the relation property's schema title, this PR also fixes it for the regular property.

Relation property:
```ts
 @hasMany(() => Child)
  children: Child[];
```

Regular property:
```ts
 @property.array(Child)
  children: Child[];
```

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/build-schema.ts
+++ b/packages/repository-json-schema/src/build-schema.ts
@@ -13,8 +13,8 @@ import {
   resolveType,
 } from '@loopback/repository';
 import debugFactory from 'debug';
-import {JsonSchema} from './index';
 import {inspect} from 'util';
+import {JsonSchema} from './index';
 import {JSON_SCHEMA_KEY} from './keys';
 const debug = debugFactory('loopback:repository-json-schema:build-schema');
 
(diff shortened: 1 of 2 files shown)


packages/repository-json-schema/src/build-schema.ts
```diff
@@ -414,6 +414,10 @@ export function modelToJsonSchema<T extends object>(
   debug('Model settings', meta.settings);
 
   const title = buildSchemaTitle(ctor, meta, options);
+  // `title` is the unique identity of a schema,
+  // it should be removed from the `options`
+  // when generating the relation or property schemas
+  if (options.title) delete options.title;
```
raymondfeng, line 420: Maybe clone `options` first?

```ts
const targetOptions = {...options);
delete targetOptions.title;
```

jannyHou: @raymondfeng hmm any reason to clone `options`? Since the `title` should not be inherited by any other schemas(relation or property in custom type), I intended to remove the `title` from `options` instead of its copy.

raymondfeng: It's usually a good idea not to mutate the input object, in this case, `options` as the caller won't expect to assume the argument is changed after the method call.

bajtos: +1 to what Raymond said

jannyHou: @raymondfeng @bajtos 👍 thx got it and applied feedback.

Issue #5239, opened 2020-04-25
@loopback/extension-authentication-jwt is not built before publish


achrinza:
## Steps to reproduce

1. Run `npm i @loopback/extension-authentication-jwt`
2. Try to import the package into an lb4 app (it doesn't work)
2. Navigate to `node_modules/@loopback/extesnion-authentication-jwt` to find no `dist` directory.

## Current Behavior

The extension is unusable.

## Expected Behavior

The extension should have been built before publish.

## Link to reproduction sandbox

N/A

## Additional information

```
$   node -e 'console.log(process.platform, process.arch, process.versions.node)' && npm ls --prod --depth 0 | grep loopback
win32 x64 12.16.1
+-- @loopback/authentication@4.2.1
+-- @loopback/authorization@0.5.6
+-- @loopback/boot@2.1.1
+-- @loopback/context@3.5.1
+-- @loopback/core@2.4.1
+-- @loopback/extension-authentication-jwt@0.1.1
+-- @loopback/openapi-v3@3.2.1
+-- @loopback/repository@2.2.1
+-- @loopback/rest@3.3.1
+-- @loopback/rest-explorer@2.1.1
+-- @loopback/service-proxy@2.1.1
```

## Related Issues

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

raymondfeng:
There was an issue in `package.json` that fails to remove `tsconfig.tsbuildinfo`.

PR #5240, opened 2020-04-25
fix(authentication-jwt): fix package.json and rename the package

raymondfeng:
It should fix https://github.com/strongloop/loopback-next/issues/5239 after the PR is landed and released.

## Checklist


extensions/authentication-jwt/package.json
```diff
@@ -1,16 +1,16 @@
 {
-  "name": "@loopback/extension-authentication-jwt",
+  "name": "@loopback/authentication-jwt",
```
achrinza, line 2: So the existing `@loopback/extension-authentication-jwt` is no longer going to be updated?

What about other packages with the `extension-` prefix (e.g. health, logging)?

raymondfeng: Other authentication related extensions only use `authentication-` prefix. That's why I rename this package to be consistent.

We'll continue to maintain other `extension-` packages.

Issue #5004, opened 2020-03-31
Spike: how to run LB3 tests from LB4 when LB3 is mounted on the LB4 app


agnes512:
### Description 

The spike is inspired by [How to run LB3 app tests when the app is mounted in a LB4 project](https://github.com/strongloop/loopback-next/issues/3978). And also see the [comment](https://github.com/strongloop/loopback-next/issues/3978#issuecomment-606660036).

There is a use case that we'd like to run LB3 tests from LB4 when the LB3 app is mounted on the LB4 app. How can we run existing LB3 tests as part of LB4 project's `npm test command`?

Instead of migrating all tests to LB4, there should be a way to run tests from original JavaScript files stored in the LB3 project.

There are two kinda of tests:
- Acceptance-level tests making HTTP calls to invoke application logic, e.g. POST /users/login This is easier because the tests need just the port number where the top-level app is listening on, this app can be the LB4 (or Express) app.

- Integration-level tests that are using JS API to call application logic, e.g. `MyModel.create()`. These tests need a way how to access LB3 artifacts of the LB3 application booted by LB4.

Some ideas from the [comment](https://github.com/strongloop/loopback-next/issues/3978#issuecomment-606660036):

- It may be better to bind all LB3 models instead of the LB3 app object, so that we can inject only specific LB3 models to LB4 code
- Maybe we can access LB3 models from LB4 controllers. That way we need to enhance the booter to bind the LB3 app object to LB4 application context, so that we can use regular dependency injection to access that app object from other LB4 code in the same way we are injecting only specific LB4 repositories to LB4 controllers.

### Acceptance Criteria

This spike aims to:

- [ ] Propose solutions for:
  - [ ] How to include tests from LB3 app in LB4 test suite. This may be as easy as adding `lb3app/tests/**/*.js` to `mocha` arguments. LB3 tests are not expected to pass yet, see the steps below.
  - [ ] How to change acceptance-level tests to use the outer LB4 app as the server to talk to, not the LB3 app. With the proposed changes in place, the acceptance-level tests (HTTP) should pass now.

- [ ] Create follow-up stories to document them and make any necessary runtime improvements.
### stretch goal
- [ ] How to change integration-level tests to accommodate for the new LB4+LB3 and pass.
[...]

PR #5251, opened 2020-04-26
Spike: Running Mounted LB3 App Tests

nabdelgadir:
Related to https://github.com/strongloop/loopback-next/issues/5004.

Proposal for running acceptance and integration tests. Details can be found in SPIKE.md.

## Checklist

### DIFF
--- a/SPIKE.md
+++ b/SPIKE.md
@@ -3,8 +3,8 @@
 Since LoopBack 4 offers a way to mount LoopBack 3 applications on a LoopBack 4
 project with the use of
 [`@loopback/booter-lb3app`](https://github.com/strongloop/loopback-next/tree/master/packages/booter-lb3app),
-there should also be a way for users to run their tests as part of LoopBack 4's
-`npm test` command.
+there should also be a way for users to run their LoopBack 3 tests as part of
+LoopBack 4's `npm test` command.
 
 We want the LoopBack 3 tests to use the LoopBack 4 server rather than the
 LoopBack 3 application. This spike aims to test running both acceptance and
@@ -95,8 +95,8 @@ application in order to allow JavaScript API to call application logic such as
 `Model.create()`. This can be seen in
 [`packages/booter-lb3app/src/lb3app.booter.ts`](https://github.com/strongloop/loopback-next/blob/spike/lb3test/packages/booter-lb3app/src/lb3app.booter.ts#L76-L85).
 
-In order to retrieve the model from the application's context,
-`getValueOrPromise()` can be used as follows:
+In order to retrieve the model from the application's context, `getSync()` can
+be used as follows:
 
 ```ts
 describe('LoopBack 3 style integration tests', function () {
@@ -110,8 +110,8 @@ describe('LoopBack 3 style integration tests', function () {
   });
 
   before(() => {
-    // follow the syntax: models.lb3-{ModelName}
-    CoffeeShop = app.lbApp.getValueOrPromise('models.lb3-CoffeeShop');
+    // follow the syntax: lb3-models.{ModelName}
+    CoffeeShop = app.lbApp.getSync('lb3-models.CoffeeShop');
   });
 
   after(async () => {
@@ -119,6 +119,15 @@ describe('LoopBack 3 style integration tests', function () {
   });
 
   // your tests here
+});
+```
+
+Alternatively, `get()` can also be used in to retrieve the model asynchronously:
+
+```ts
+before(async () => {
+  // follow the syntax: lb3-models.{ModelName}
+  CoffeeShop = await app.lbApp.get('lb3-models.CoffeeShop');
 });
 ```
 
--- a/SPIKE.md
+++ b/SPIKE.md
@@ -17,3 +17,68 @@ path to `test` entry in package.json:
 
 In this case, the test folder lies is `/lb3app/test` from the root of the
 LoopBack 4 project.
+
+This will run LoopBack 4 tests first then LoopBack 3 tests.
+
+## Acceptance Tests
+
+First, move any LoopBack 3 test dependencies to package.json's devDependencies
+and run:
+
+```sh
+npm install
+```
+
+In your test file:
+
+- Update to use the LB4 Express server when doing requests:
+
+  ```ts
+  // can use lb4's testlab's supertest as the dependency is already installed
+  const request = require('@loopback/testlab').supertest;
+  const assert = require('assert');
+  const should = require('should');
+  const ExpressServer = require('../../dist/server').ExpressServer;
+
+  let app;
+
+  function json(verb, url) {
+    // use the LB4 express server
+    return request(app.server)
+      [verb](url)
+      .set('Content-Type', 'application/json')
+      .set('Accept', 'application/json')
+      .expect('Content-Type', /json/);
+  }
+  ```
+
+- Boot and start the LB4 app in your before hook, and stop the app in the after
+  hook:
+
+  ```ts
+  describe('LoopBack 3 style tests', function () {
+    before(async function () {
+      app = new ExpressServer();
+      await app.boot();
+      await app.start();
+    });
+
+    after(async () => {
+      await app.stop();
+    });
+
+    // your tests here
+  });
+  ```
+
+- Example of this use can be seen in
+  `examples/lb3-application/lb3app/test/acceptance.js` which has the same tests
+  as `src/__tests__/acceptance/lb3app.acceptance.ts`, but in LB3 style.
+
+Now when you run `npm test` your LoopBack 3 tests should be run along with any
+LoopBack 4 tests you have.
+
+Optional: Another option is to migrate your tests to use LoopBack 4 style of
+testing, similar to `src/__tests__/acceptance/lb3app.acceptance.ts`.
+Documentation for LoopBack testing can be found in
+https://loopback.io/doc/en/lb4/Testing-your-application.html.
--- a/SPIKE.md
+++ b/SPIKE.md
@@ -1,3 +1,5 @@
+# Run LB3 tests from LB4 when LB3 is mounted on the LB4 app
+
 Since LoopBack 4 offers a way to mount LoopBack 3 applications on a LoopBack 4
 project with the use of
 [`@loopback/booter-lb3app`](https://github.com/strongloop/loopback-next/tree/master/packages/booter-lb3app),
@@ -15,8 +17,9 @@ path to `test` entry in package.json:
 
 - `"test": "lb-mocha \"dist/**tests**/\*_/_.js\" \"lb3app/test/\*.js\""`
 
-In this case, the test folder lies is `/lb3app/test` from the root of the
-LoopBack 4 project.
+In this case, the test folder is
+[`/lb3app/test`](https://github.com/strongloop/loopback-next/tree/spike/lb3test/examples/lb3-application/lb3app/test)
+from the root of the LoopBack 4 project.
 
 This will run LoopBack 4 tests first then LoopBack 3 tests.
 
@@ -72,8 +75,10 @@ In your test file:
   ```
 
 - Example of this use can be seen in
-  `examples/lb3-application/lb3app/test/acceptance.js` which has the same tests
-  as `src/__tests__/acceptance/lb3app.acceptance.ts`, but in LB3 style.
+  [`examples/lb3-application/lb3app/test/acceptance.js`](https://github.com/strongloop/loopback-next/blob/spike/lb3test/examples/lb3-application/lb3app/test/acceptance.js)
+  which has the same tests as
+  [`src/__tests__/acceptance/lb3app.acceptance.ts`](https://github.com/strongloop/loopback-next/blob/spike/lb3test/examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts),
+  but in LB3 style.
 
 Now when you run `npm test` your LoopBack 3 tests should be run along with any
 LoopBack 4 tests you have.
@@ -88,7 +93,7 @@ https://loopback.io/doc/en/lb4/Testing-your-application.html.
 For the integration tests, LoopBack 3 models were bound to the LoopBack 4
 application in order to allow JavaScript API to call application logic such as
 `Model.create()`. This can be seen in
-`packages/booter-lb3app/src/lb3app.booter.ts`.
+[`packages/booter-lb3app/src/lb3app.booter.ts`](https://github.com/strongloop/loopback-next/blob/spike/lb3test/packages/booter-lb3app/src/lb3app.booter.ts#L76-L85).
 
 In order to retrieve the model from the application's context,
 `getValueOrPromise()` can be used as follows:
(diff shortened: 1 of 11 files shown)


examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -0,0 +1,136 @@
...
+        User.login(
+          {
+            email: 'new@email.com',
+            password: 'L00pBack!',
+          },
+          function (err2, token) {
+            assert.equal(token.userId, user.id);
+            json(
+              'get',
+              `/api/CoffeeShops/greet?access_token=${token.id}`,
```
nabdelgadir, line 124: The url here uses the correct token and the user is logged in so I'm not sure why it's unauthorized 🤔

bajtos:
Thank you @nabdelgadir for looking into ways how to run tests of a LB3 app from inside LB4! I'll try to review the proposal tomorrow 🤞

SPIKE.md
```diff
@@ -0,0 +1,126 @@
+# Run LB3 tests from LB4 when LB3 is mounted on the LB4 app
+
+Since LoopBack 4 offers a way to mount LoopBack 3 applications on a LoopBack 4
+project with the use of
+[`@loopback/booter-lb3app`](https://github.com/strongloop/loopback-next/tree/master/packages/booter-lb3app),
+there should also be a way for users to run their tests as part of LoopBack 4's
```
bajtos, line 6: Let's please clarify that `their tests` are referring to the LB3 application tests.

SPIKE.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Integration Tests
+
+For the integration tests, LoopBack 3 models were bound to the LoopBack 4
+application in order to allow JavaScript API to call application logic such as
+`Model.create()`. This can be seen in
+[`packages/booter-lb3app/src/lb3app.booter.ts`](https://github.com/strongloop/loopback-next/blob/spike/lb3test/packages/booter-lb3app/src/lb3app.booter.ts#L76-L85).
+
+In order to retrieve the model from the application's context,
+`getValueOrPromise()` can be used as follows:
```
bajtos, line 99: Please note that `getValueOrPromise` is an internal helper. Users should preferably use `get` (which returns a promise) or `getSync` (that throws an error if the value cannot be resolved synchronously).

examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -0,0 +1,132 @@
...
+      done();
+    });
+
+    it('makes an authenticated request', function (done) {
+      User.create({email: 'new@email.com', password: 'L00pBack!'}, function (
+        err,
+        user,
+      ) {
+        user.email.should.be.equal('new@email.com');
+        json('post', '/api/users/login')
```
bajtos, line 113: Please find a way how to allow this test to call `User.login` directly via JavaScript API, so that users don't have to rework all their tests calling model API from JavaScript to make HTTP requests instead.

examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -0,0 +1,132 @@
...
+          .send({
+            email: 'new@email.com',
+            password: 'L00pBack!',
+          })
+          .expect(200, function (err2, token) {
+            assert(typeof token.body === 'object');
```
bajtos, line 119: When this assertion fails, you will have no idea what was the actual type of the body object. A better solution:

```js
assert.equals(typeof token.body, 'object');
```

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -73,6 +73,17 @@ export class Lb3AppBooter implements Booter {
...
     }
 
+    const models = lb3App.models;
+    if (models) {
+      const visited: unknown[] = [];
+      Object.keys(models).forEach(key => {
+        const model = models[key];
+        if (visited.includes(model)) return;
+        visited.push(model);
+        this.app.bind(`models.lb3-${key}`).to(model).tag('model');
```
bajtos, line 83: I am not sure if it's a good idea to use `models` namespace and prefix the name with `lb3-`. I am concerned about possible conflicts this could create once we start binding LB4 models too. Have you considered using a different namespace, e.g. `lb3-models`? Same comment applies on the tag. At the end of the day, LB3 models are not interchangeable with LB4 models, therefore it's better to clearly mark them as different artifacts.

```suggestion
        this.app.bind(`lb3-models.${key}`).to(model).tag('lb3-model');
```

packages/booter-lb3app/src/lb3app.booter.ts
```diff
@@ -149,4 +160,5 @@ export interface Lb3AppBooterOptions {
 interface Lb3Application extends ExpressApplication {
   handler(name: 'rest'): ExpressRequestHandler;
   dataSources?: {[name: string]: unknown};
+  models?: {[name: string]: unknown};
```
bajtos, line 163: The actual type is `typeof juggler.PersistedModel`, but unfortunately we don't have access to that type in `@loopback/booter-lb3app`. 

Few ideas to consider:

(1)
Add typings to `loopback` package. To keep it simple, we can start by re-exporting loopback-datasource-juggler types. Maybe move `Lb3Application` to `loopback` along the way. This will give us strong type checks for code like `lb3app.models.User.create`.

```suggestion
  models?: {[name: string]: typeof juggler.PersistedModel};
```

Unfortunately, this will not help us with custom model methods like `User.login`. 

There is also a bug in juggler typings which prevents code like `lb3app.models.User.create({email: 'test@example.com'})` to compile, because PersistedModel does not have any properties and thus object `{email}` cannot be assigned to interface `ModelData<PersistedModel>`.

Considering those issues, I think we should not pursue this approach.

(2)
Considering that LB3 is pure JavaScript with no types, maybe we can use `any` type here? It will effectively disable compiler checks for any code accessing LB3 models methods, but I think that's ok considering that there were no type checks in the original LB3 application code base either.

```suggestion
  models?: {[name: string]: any};
```

WDYT?

examples/lb3-application/lb3app/test/acceptance.js
bajtos: Is this perhaps trying to workaround the problem that `User` has `unknown` type? (See the discussion below.)

packages/booter-lb3app/src/lb3app.booter.ts
nabdelgadir: I agree; I'll update this. I referenced the [code for binding lb3 datasources](https://github.com/strongloop/loopback-next/blob/master/packages/booter-lb3app/src/lb3app.booter.ts#L65-L74) when I was doing this, so I can open a new PR to change that too.

packages/booter-lb3app/src/lb3app.booter.ts
nabdelgadir: I like the second approach 👍

examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -0,0 +1,130 @@
...
+'use strict';
+const lb3App = require('../server/server');
+const request = require('@loopback/testlab').supertest;
+const assert = require('assert');
+const ExpressServer = require('../../dist/server').ExpressServer;
+require('should');
+
+let app;
+
+function json(verb, url) {
```
jannyHou, line 15: nitpick: the function name is a bit confusing, maybe call it as `setContentType`?
And why the test has to set the content type and Accept as `application/json` then send the payload? Is it also part of the migration guide?
(hmm, "migration" may not be the accurate word haha)

examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -0,0 +1,130 @@
...
+      User = lb3App.models.User;
+    });
+
+    it('creates a User and logs them in and out', function (done) {
+      User.create({email: 'new@email.com', password: 'L00pBack!'}, function (
+        err,
+        user,
+      ) {
+        assert.equal(user.email, 'new@email.com');
+        User.login(
```
jannyHou, line 84: glad to see you got it work 👍 and nice to verify that User authentication works as expected.

SPIKE.md
```diff
@@ -0,0 +1,135 @@
...
+  ```ts
+  // can use lb4's testlab's supertest as the dependency is already installed
+  const request = require('@loopback/testlab').supertest;
+  const assert = require('assert');
+  const should = require('should');
+  const ExpressServer = require('../../dist/server').ExpressServer;
+
+  let app;
+
+  function json(verb, url) {
```
jannyHou, line 48: same comment as https://github.com/strongloop/loopback-next/pull/5251/files#r416996993

SPIKE.md
```diff
@@ -0,0 +1,135 @@
...
+LoopBack 4 tests you have.
+
+Optional: Another option is to migrate your tests to use LoopBack 4 style of
+testing, similar to `src/__tests__/acceptance/lb3app.acceptance.ts`.
+Documentation for LoopBack testing can be found in
+https://loopback.io/doc/en/lb4/Testing-your-application.html.
+
+## Integration Tests
+
+For the integration tests, LoopBack 3 models were bound to the LoopBack 4
```
jannyHou, line 93: Besides the model, another important artifact in LB3 is the datasource, I assume they are already bond to the app, could you also write the ds binding key's naming convention here?

examples/lb3-application/lb3app/test/acceptance.js
jannyHou: IIUC...I think Nora tests the endpoint instead of the API here because it's acceptance tests. The APIs are tested in the integration test file.

SPIKE.md
nabdelgadir: Let's see if https://github.com/strongloop/loopback-next/pull/5287 gets accepted 👍

examples/lb3-application/lb3app/test/acceptance.js
nabdelgadir: I realize now this file is a bit confusing because I was trying to mimic [the original acceptance tests](https://github.com/strongloop/loopback-next/blob/master/examples/lb3-application/src/__tests__/acceptance/lb3app.acceptance.ts) which makes this now acceptance + integration but I think I'll put the auth tests in their own file.

examples/lb3-application/lb3app/test/acceptance.js
nabdelgadir: I thought this was traditional lb3 naming because it's what I've seen in lb3 examples (like [here](https://github.com/strongloop/loopback-example-app-logic/blob/master/test/rest_api_test.js#L12-L17) and [here](https://github.com/strongloop/loopback-example-access-control/blob/master/test/rest_api_test.js#L12-L17)) 🙈 

> And why the test has to set the content type and Accept as application/json then send the payload? Is it also part of the migration guide?

Since we want users to change as little code as possible, I maintained the function as is, but I agree we can rename the function name 😄

nabdelgadir: Renamed to `request` to match what the function does; what do you think?

examples/lb3-application/lb3app/test/acceptance.js
bajtos: Here is my point of view: this acceptance test has two parts. In the first part, we want to setup data (e.g. User instance, a valid AccessToken, etc.). In the second part, we test the actual scenario.

I am arguing that for test setup, we should use the most performant solution. This part is not testing anything, it's just preparing data to enable the scenario being tested. That's why I prefer to call model methods directly via JS API. HTTP calls introduce non-negligible slow down.

SPIKE.md
```diff
@@ -38,16 +38,16 @@ In your test file:
 
   ```ts
   // can use lb4's testlab's supertest as the dependency is already installed
-  const request = require('@loopback/testlab').supertest;
+  const supertest = require('@loopback/testlab').supertest;
```
bajtos, line 41: Nitpick, feel free to ignore.

```suggestion
  const {supertest} = require('@loopback/testlab');
```

examples/lb3-application/lb3app/test/acceptance.js
```diff
@@ -4,17 +4,16 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
-const lb3App = require('../server/server');
-const request = require('@loopback/testlab').supertest;
+const supertest = require('@loopback/testlab').supertest;
```
bajtos, line 7: Ditto.

```suggestion
const {supertest} = require('@loopback/testlab');
```

examples/lb3-application/lb3app/test/authentication.js
```diff
@@ -0,0 +1,94 @@
...
+      User = lb3App.models.User;
+    });
+
+    it('creates a User and logs them in and out', function (done) {
+      User.create({email: 'new@email.com', password: 'L00pBack!'}, function (
+        err,
+        user,
+      ) {
+        assert.equal(user.email, 'new@email.com');
+        User.login(
```
bajtos, line 48: Considering that the test name says "it creates a user and logs them in and out", I think the test should call `User.create`, `User.login` and `User.logout` via REST API?

I apologize if this is contradicting my previous comments, I guess I didn't realize what exactly is this test about 🙈

examples/lb3-application/lb3app/test/authentication.js
```diff
@@ -0,0 +1,94 @@
...
+        user.email.should.be.equal('new@email.com');
+        User.login({email: 'new@email.com', password: 'L00pBack!'}, function (
+          err2,
+          token,
+        ) {
+          assert.equal(typeof token, 'object');
+          assert.equal(token.userId, user.id);
+          request(
+            'get',
+            `/api/CoffeeShops/greet?access_token=${token.id}`,
```
bajtos, line 85: This is perfect:
- Test setup is done via `User.create` and `User.login`
- The behavior under test is invoked via REST API call

:+1:

packages/booter-lb3app/src/__tests__/acceptance/booter-lb3app.acceptance.ts
```diff
@@ -242,4 +242,20 @@ describe('booter-lb3app', () => {
...
+  context('binding LoopBack 3 models', () => {
+    before(async () => {
+      ({app, client} = await setupApplication({
+        lb3app: {path: '../fixtures/app-with-model'},
+      }));
+    });
+
+    it('binds model to the context', async () => {
+      const expected = require('../../../fixtures/app-with-model').models.Color;
+      const modelBindings = app.findByTag('model');
```
bajtos, line 255: Did this test pass CI? I would expect the tag to be `lb3-model`. 🤔 

```suggestion
      const modelBindings = app.findByTag('lb3-model');
```

examples/lb3-application/lb3app/test/acceptance.js
jannyHou: > I am arguing that for test setup, we should use the most performant solution. This part is not testing anything, it's just preparing data to enable the scenario being tested. That's why I prefer to call model methods directly via JS API. HTTP calls introduce non-negligible slow down.

Oh I see, thanks for explanation 👍

examples/lb3-application/lb3app/test/acceptance.js
jannyHou: I see, I didn't realize it's from LB3...then I am ok to keep the name as it is. Since it's migration.

nabdelgadir:
Created followup stories:
- Binding LB3 models to LB4 context https://github.com/strongloop/loopback-next/issues/5298

These two should be completed after the first one:
- Add example tests for LB3 app https://github.com/strongloop/loopback-next/issues/5300
- Documentation https://github.com/strongloop/loopback-next/issues/5298

Closing the spike as done.

PR #5263, opened 2020-04-27
Codeowners

dhmlau:
There are 2 commits in this PR. It updates `CODEOWNERS`:
- remove @nabdelgadir as her work term is finishing
- add entries for the newly added packages

## Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -88,15 +88,15 @@
 # For longer term, we may want to re-organize doc files around functional
 # areas (create a new subdirectory for each functional area), so that
 # we can assign more specific code owners to different doc pages.
-/docs @agnes512 @bajtos @deepakrkris @emonddr @jannyHou @hacksparrow @nabdelgadir @raymondfeng
+/docs @agnes512 @bajtos @deepakrkris @emonddr @jannyHou @hacksparrow @raymondfeng
 
 #
 # CLI (`lb4` infrastructure, commands not covered by functional areas below)
 #
 # - Issue label: CLI
 # - Primary owner(s): @agnes512
-# - Standby owner(s): @emonddr @nabdelgadir
-/packages/cli @agnes512 @emonddr @nabdelgadir
+# - Standby owner(s): @emonddr
+/packages/cli @agnes512 @emonddr
 
 #
 # Dependency Injection, Inversion of Control and related areas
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -243,21 +243,21 @@
...
 # - Issue label: n/a
-# - Primary owner(s): @nabdelgadir
+# - Primary owner(s): @raymondfeng
 # - Standby owner(s): n/a
-/packages/model-api-builder @nabdelgadir
-/packages/rest-crud @nabdelgadir
-/examples/rest-crud @nabdelgadir
+/packages/model-api-builder @hacksparrow
+/packages/rest-crud @raymondfeng
+/examples/rest-crud @raymondfeng
```
bajtos, line 260: I see the first entry is using `hacksparrow` but the remaining two are using `raymondfeng`. We usually want to use the same set of people for all packages belonging to the same function area.

BTW I am happy to take primary ownership of these packages myself, if that helps.

CODEOWNERS
```diff
@@ -280,7 +280,9 @@
 /packages/authentication @emonddr @hacksparrow @jannyHou
 /packages/security @emonddr @hacksparrow @jannyHou
 /extensions/authentication-passport @emonddr @hacksparrow @jannyHou
+/extensions/authentication-jwt @emonddr @hacksparrow @jannyHou
 /docs/site/tutorials/authentication @emonddr @hacksparrow @jannyHou
+/examples/passport-login @deepakrkris @emonddr
```
bajtos, line 285: Similarly here, should this line use the same list of CODEOWNERS as the lines above?

CODEOWNERS
dhmlau: Sure. IIRC, @hacksparrow also worked in this area too. @hacksparrow, would you like to be the standby owner?

CODEOWNERS
dhmlau: Hm. I'm not sure about this one. @deepakrkris created this example, so I assume he should be in the list? or are you suggesting the standby owners should be: 
```Standby owner(s): @emonddr @jannyHou @deepakrkris ```

bajtos: I don't mind either way. I just found it surprising that some lines in the same functional area have different owners than other lines.

Maybe we should use all FA owners (because they are FA owners) and add Deepak as the original author?

jannyHou: I am good to be the standby owner :)

dhmlau:
@slnode test please

Issue #2358, opened 2019-02-08
Bash completion script for lb4 CLI


bajtos:
Quoting from https://iridakos.com/tutorials/2018/03/01/bash-programmable-completion-tutorial.html

> ### What is bash completion?
> Bash completion is a functionality through which bash helps users type their commands faster and easier. It accomplishes that by presenting possible options when users press the tab key while typing a command.
>
>```
> $ git<tab><tab>
> git                 git-receive-pack    git-upload-archive  
> gitk                git-shell           git-upload-pack     
> $ git-s<tab>
> $ git-shell
>```
>
> ### How it works
> The completion script is code that uses the builtin bash command complete to define which completion suggestions can be displayed for a given executable. The nature of the completion options vary from simple static to highly sophisticated.
>
> ### Why bother
> It is good to provide such a functionality to users:
> 
> - to save them from typing text when it can be auto-completed
> - to help them know what are the available continuations to their commands
> - to prevent errors and improve their experience by hiding or showing options based on what they have already typed

I am proposing to implement bash completion for our `lb4` CLI. Let's start simple with a static completion offering list of CLI commands. We can add more advanced completion later in subsequent iterations.

```
$ lb4<tab><tab>
app          datasource   service
extension    model        example
controller   repository   openapi
```

A proposal for installation instructions:

```
source <(lb4 --completion-bash)
```

See https://iridakos.com/tutorials/2018/03/01/bash-programmable-completion-tutorial.html for a tutorial on how to build bash completion and `npx --shell-auto-fallback bash` for inspiration on how to implement it in JavaScript/Node.js.

raymondfeng:
Check out https://github.com/mklabs/tabtab

peanutenthusiast:
I'm interested in submitting a PR for this!

dhmlau:
@peanutenthusiast, please feel free!  Let me label it as `community contribution` then.

noctifer20:
I've created basic completion with [tabtab](https://github.com/mklabs/tabtab)
And for installation, you have to run something like `lb4 install-completion`.
Is it suitable for cli ?

<img width="734" alt="Screen Shot 2019-08-23 at 7 42 09 PM" src="https://user-images.githubusercontent.com/18212378/63605398-e9f3ee80-c5de-11e9-8640-f8e270eaea64.png">

noctifer20:
cc @bajtos

bajtos:
> I've created basic completion with tabtab

Awesome!

[...]

PR #5296, opened 2020-04-30
feat/cli-tab-completion: add bash-completion using tabtab package

spirAde:
This makes it possible to use bash-completion using tabtab package. At the moment, all possible options are included for all existing commands, except for general ones (such as --help, --force installation, etc). To activate bash-completion, a user needs to run the `lb4 install-completion` command, to remove the completion `lb4 uninstall-completion`.

Issue #2358

## Checklist


spirAde:
Hello, guys. I would like to clarify the point with common commands (--help, --force-installation, etc) whether we should use them in bash-completion?

spirAde:
Also, the user will need to restart the .bashrc, .zshrc after lb4 install-completion command, so that bash-completion began to work. What is the best way to inform the user about this and is it worth it?

packages/cli/lib/tab-completion.js
```diff
@@ -0,0 +1,224 @@
...
+const debug = require('./debug')();
+const tabtab = require('tabtab');
+
+const tabCompletionCommands = [
+  'install-completion',
+  'uninstall-completion',
+  'completion',
+];
+
+const commands = [
```
raymondfeng, line 17: I think it's possible to build these lists from generator metadata without hard-coding them. Let's look at how `help` is implemented. `--help` lists available commands and their options.

raymondfeng:
@spirAde Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-next. Thanks.

spirAde:
> @spirAde Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-next. Thanks.

I've already signed it
image

image

packages/cli/lib/tab-completion.js
spirAde: @raymondfeng Thank you for your feedback. I've checked the current solution for getting all possible commands. And I was thinking about it before to take `env.getGeneratorsMeta()` solution. But I see a problem with this solution. Even if I get commands from `env.getGeneratorsMeta()`, we don't have the possibility to take command's options from yeoman. Therefore hardcoded options and not-hardcoded commands should have been supported that is not good from my perspective.

But I investigated `yeoman-environment` and found internal `_options` property. What do you think about this proposal: add `getOptions` method inside `BaseGenerator` class.

```
...
getOptions() {
  return this._options;
}
...
```

this._options view for controller:

```
{ help:
   { name: 'help',
     type: [Function: Boolean],
     alias: 'h',
     description: 'Print the generator\'s options and usage' },
  'skip-cache':
   { name: 'skip-cache',
     type: [Function: Boolean],
     description: 'Do not remember prompt answers',
     default: false },
  'skip-install':
   { name: 'skip-install',
     type: [Function: Boolean],
     description: 'Do not automatically install dependencies',
     default: false },
  'force-install':
   { name: 'force-install',
     type: [Function: Boolean],
     description: 'Fail on install dependencies error',
     default: false },
  'ask-answered':
   { type: [Function: Boolean],
     description: 'Show prompts for already configured options',
     default: false,
     name: 'ask-answered',
     hide: false },
  controllerType:
   { type: [Function: String],
     required: false,
     description: 'Type for the controller',
     name: 'controllerType',
     hide: false },
  config:
   { type: [Function: String],
     alias: 'c',
     description: 'JSON file name or value to configure options',
     name: 'config',
     hide: false },
  yes:
   { type: [Function: Boolean],
     alias: 'y',
     description:
      'Skip all confirmation prompts with default or provided value',
     name: 'yes',
     hide: false },
  format:
   { type: [Function: Boolean],
     description: 'Format generated code using npm run lint:fix',
     name: 'format',
     hide: false } }

```

After getting options, I can filter and convert them to a fit view. It makes possible to use dynamic commands and options.

raymondfeng: I'm adding code to get options and args for each command.

spirAde:
> @spirAde Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-next. Thanks.

I've already signed. 
image
image
image

raymondfeng:
@spirAde I rebased the code against latest master and pushed one commit to use inspected options and arguments. Please continue from that. Please also add some tests.

raymondfeng:
You may need to run the following git commands to pull it down:
- git fetch --all
- git reset --hard origin/feat/cli-tab-completion

packages/cli/lib/tab-completion.js
spirAde: It's better solution) Thank you

raymondfeng:
@spirAde The code can be further simplified as follows:

```
/**
 * Get all available commands or options depends on entered input
 * @param partial - completion env.partial
 */
function generateCompletions(optionsAndArgs, partial) {
  const commands = optionsAndArgs.map(i => i.name);
  const enteredCommand = commands.find(command => partial.includes(command));

  if (!enteredCommand) {
    return commands;
  }

  const options = optionsAndArgs[enteredCommand].map(o => o.name);
  const enteredOptions = options.filter(option => partial.includes(option));

  if (!enteredOptions.length) {
    return optionsAndArgs[enteredCommand].options.map(o => `--${o.name}`);
  }

  return options.filter(option => !enteredOptions.includes(option.name));
}

I tried and it seems that getting options take some time. The responsiveness is good. We need to cache the options to the file system.
```

spirAde:
> @spirAde The code can be further simplified as follows:
> 
> ```js
> /**
>  * Get all available commands or options depends on entered input
>  * @param partial - completion env.partial
>  */
> function generateCompletions(optionsAndArgs, partial) {
>   const commands = optionsAndArgs.map(i => i.name);
>   const enteredCommand = commands.find(command => partial.includes(command));
> 
>   if (!enteredCommand) {
>     return commands;
>   }
> 
>   const options = optionsAndArgs[enteredCommand].map(o => o.name);
>   const enteredOptions = options.filter(option => partial.includes(option));
> 
>   if (!enteredOptions.length) {
>     return optionsAndArgs[enteredCommand].options.map(o => `--${o.name}`);
>   }
> 
>   return options.filter(option => !enteredOptions.includes(option.name));
> }
> ```
> 
> I tried and it seems that getting options takes some time. The responsiveness is NOT good. We need to cache the options to the file system.

Yes, that was another reason why I preferred hardcoded data. Time measurement gave out ~ 200ms.

spirAde:
I will write when I get some results

raymondfeng:
one idea is to save the metadata into a json file at release time so that we don’t have to rebuild when tab is hit.

raymondfeng:
@spirAde I went ahead to add one commit:

1. Allow `lb4 --meta` to dump command options to `.yo-rc.json`
2. `lb4 completion` uses information from `.yo-rc.json` for completion
3. I made `lb4 completion` as fast as possible without loading other things

raymondfeng:
@spirAde The first commit was made under `dspiridonov4xxi <dspiridonov@4xxi.com>`. That's probably why the CLA complains. Can we amend the commit to use your github account?

spirAde:
@raymondfeng Thank you for your help. I have measured responsiveness on my local machines(macOS, Linux) and can say it's more than enough. I added tests to reach 100% coverage for `lib/tab-completion`. But I don't understand why CI process fails. I'd checked all tests on my both machines and didn't get any errors.

raymondfeng:
@spirAde Thank you for the update. Don't worry about the CI failures - they are caused by timeouts. I'll increase the timeout for CI.

I'll squash the commits and merge it.

Again, thank you for the great contribution! We love ideas to improve developer experience for LoopBack! 👍

PR #5304, opened 2020-05-01
feat(boot): add helpers to create a booter for component applications

raymondfeng:
Use case:

- Create a LoopBack application
- Add controllers, services, datasources
- Now use this application as a component with the main applicaiton
- The main application needs to have artifacts from the component application

## Checklist

### DIFF
--- a/docs/site/Booting-an-Application.md
+++ b/docs/site/Booting-an-Application.md
@@ -287,3 +287,44 @@ Used to discover the artifacts supported by the `Booter` based on convention.
 **load**
 
 Used to bind the discovered artifacts to the Application.
+
+### Boot an application using component
+
+For a complex project, we may break it down into multiple LoopBack applications,
+each of which has controllers, datasources, services, repositories, and other
+artifacts. How do we compose these sub applications into the main application?
+The component application booter can be created to support this use case.
+
+1. Create a component for the sub-application:
+
+```ts
+import {createComponentApplicationBooterBinding} from '@loopback/boot';
+import {Component} from '@loopback/core';
+
+export class SubAppComponent implements Component {
+  bindings = [
+    createComponentApplicationBooterBinding(
+      new SubApp(), /* an optional binding filter */,
+    ),
+  ];
+}
+```
+
+2. Mount the sub-application as a component to the main application:
+
+```ts
+const mainApp = new MainApp();
+
+// This can be done in the constructor of `MainApp` too. Make sure the component
+// is registered before calling `app.boot()`.
+mainApp.component(SubAppComponent);
+
+// Boot the main application. It will invoke the component application booter
+// to add artifacts from the `SubApp`.
+await mainApp.boot();
+```
+
+A binding filter function can be provided to select what bindings from the
+component application should be added to the main application. The booter skips
+bindings that exist in the component application before `boot`. It does not
+override locked bindings in the main application.
--- a/packages/boot/src/booters/component-application.booter.ts
+++ b/packages/boot/src/booters/component-application.booter.ts
@@ -1,0 +1,113 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/boot
+// This file is licensed under the MIT License.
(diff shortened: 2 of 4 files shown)


packages/boot/src/booters/component-application.booter.ts
```diff
@@ -0,0 +1,66 @@
...
+      @inject(CoreBindings.APPLICATION_INSTANCE) private mainApp: Application,
+    ) {}
+
+    async load() {
+      await componentApp.boot();
+      /**
+       * Add bindings from the component application to the main application
+       */
+      const bindings = componentApp.find(binding => {
+        return filter(binding) && !this.mainApp.contains(binding.key);
```
jannyHou, line 43: A question for `!this.mainApp.contains(binding.key)`:
what if the sub app wants to override the main app's bindings?

raymondfeng: There are certain bindings from the sub-app should not override the main app. We should carefully review such bindings - for example, `CoreBindings.*`, `RestBindings.*`. Maybe one way to solve it is to mark such bindings locked. If so, we can update this code to only skip locked bindings in the main app.

docs/site/Booting-an-Application.md
```diff
@@ -287,3 +287,39 @@ Used to discover the artifacts supported by the `Booter` based on convention.
...
+```
+
+2. Mount the sub-application as a component to the main application:
+
+```ts
+const mainApp = new MainApp();
+
+// This can be done in the constructor of `MainApp` too. Make sure the component
+// is registered before calling `app.boot()`.
+mainApp.component(BooterAppComponent);
```
agnes512, line 320: `SubAppComponent`?

raymondfeng: Yes

packages/boot/src/booters/component-application.booter.ts
jannyHou: > We should carefully review such bindings - for example, `CoreBindings.*`, `RestBindings.*`

True, agreed. I came up with this question due to the chat happened on public slack this morning, I think that community user tries to contribute custom artifacts(like models, in his case) from the component.

> Maybe one way to solve it is to mark such bindings locked. If so, we can update this code to only skip locked bindings in the main app.

I feel it's a case by case preference, maybe we can provide a default filter to skip those core bindings. And if user really wants to override, they provide custom filters :)

(don't have to address it in this PR)

raymondfeng: I improved the binding filtering as follows:

- Bindings that exist in the component application before `boot` are skipped.
- Locked bindings in the main application will not be overridden.
- Other bindings are selected by the binding filter. By default, all bindings are picked.

bhupesh-sf:
Documentation for sub application binding in main application doesn't contain any exmaple. That will be helpful

docs/site/Booting-an-Application.md
```diff
@@ -287,3 +287,44 @@ Used to discover the artifacts supported by the `Booter` based on convention.
...
+```
+
+2. Mount the sub-application as a component to the main application:
+
+```ts
+const mainApp = new MainApp();
+
+// This can be done in the constructor of `MainApp` too. Make sure the component
+// is registered before calling `app.boot()`.
+mainApp.component(SubAppComponent);
```
bajtos, line 320: @raymondfeng I find these instructions quite involved. Would the following approach work & make sense too?

```ts
class MyApp extends BootMixin(Application) {
  constructor(options: ApplicationOptions) {
    super(options);

    this.add(
      createComponentApplicationBooterBinding(new SubApp()),
    );
  }
}
```

I think it would be great to add a helper method to `BootMixin` to wrap those two calls in a single API, but I find it difficult to find a good method name. How about `addApplicationBooter`?

Intended usage:

```ts
class MyApp extends BootMixin(Application) {
  constructor(options: ApplicationOptions) {
    super(options);

    this.addApplicationBooter(new SubApp());
  }
}
```

WDYT?

PR #5309, opened 2020-05-02
feat(core): allow options for artifact registration on an application

raymondfeng:
The new style allows extra options such as `namespace` to be customized.
One use case is that a sub application might boot artifacts into its own
namespace.

## Checklist

### DIFF
--- a/packages/core/src/application.ts
+++ b/packages/core/src/application.ts
@@ -5,6 +5,7 @@
 
 import {
   Binding,
+  BindingFromClassOptions,
   BindingScope,
   Constructor,
   Context,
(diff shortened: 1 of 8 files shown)


packages/core/src/application.ts
```diff
@@ -532,6 +537,17 @@ export class Application extends Context implements LifeCycleObserver {
   }
 }
 
+/**
+ * Normalize name or options to `BindingFromClassOptions`
+ * @param nameOrOptions - Name or options for binding from class
+ */
+function toOptions(nameOrOptions?: string | BindingFromClassOptions) {
```
jannyHou, line 544: nitpick: I see `toOptions` twice, maybe refactor it into a util function?

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -216,6 +229,17 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
   };
 }
 
+/**
+ * Normalize name or options to `BindingFromClassOptions`
+ * @param nameOrOptions - Name or options for binding from class
+ */
+function toOptions(nameOrOptions?: string | BindingFromClassOptions) {
```
jannyHou, line 236: ditto

raymondfeng: They are in two packages. It's not worth the effort IMO.

PR #5329, opened 2020-05-04
fix(repository): updateById behaves like updateAll when called with a…

### LINKED ISSUE #5319: when undefined is passed to repository.updateById as id value, it executes an update on all rows.
## Steps to reproduce

get a repository instance.
Call updateById with the undefined as id parameter.
it will execute an update on all rows.

`    //create table
    await this.repo.execute('CREATE TABLE mymodel (col1 TEXT)', []);

    //populate data
    await this.repo.execute(
      `insert into mymodel(col1) values ('a'),('b'),('c');`,
      [],
    );

    const all = await this.repo.find();
    const first = all[0];
    first.col1 =
      'meant to change one row, but passed in undefined to updateById as id, and now all rows are updated.';
    await this.repo.updateById(undefined, first);
    const newRows = await this.repo.find();
    console.log(newRows);
`
the output should be something like;
[ {  "rowid": 1,
    "col1": "meant to change one row, but passed in undefined to updateById as id, and now all rows are updated."  },
  {  "rowid": 2,
    "col1": "meant to change one row, but passed in undefined to updateById as id, and now all rows are updated."  },
  {   "rowid": 3,
    "col1": "meant to change one row, but passed in undefined to updateById as id, and now all rows are updated."  }
]

## Current Behavior

As a result, it executes an DefaultCrudRepository.updateAll on the entire table.
## Expected Behavior

Nothing to be updated, as a row with that ID does not exist, or it can throw an error saying that id cannot be undefined.

## Link to reproduction sandbox

https://github.com/hazimdikenli/repobug
## Additional information

darwin x64 12.16.3

repobug@1.0.0 /Users/hazimdikenli/learn/loopback/repobug
├── @loopback/boot@2.1.2
├── @loopback/context@3.6.0
├── @loopback/core@2.4.2
├── @loopback/openapi-v3@3.3.0
├── @loopback/repository@2.3.0
├── @loopback/rest@3.3.2
├── @loopback/rest-explorer@2.1.2
├── @loopback/service-proxy@2.1.2
├── loopback-connector-sqlite3@3.0.0
## Related Issues

hazimdikenli:
…n undefined value for id

DefaultCrudRepository.updateById behaves like updateAll when it is called with undefined as id value , and as a side effect instead of trying to update a single row/model it updates all the rows. undefined is not a valid value for id, hence the function should throw an appropriate error when undefined is passed in.
I added an argument value check for id, in the updateById function, throws an error, and added a test to unit tests to check for this behaviour.

Fixes #5319

## Checklist

### DIFF
--- a/packages/repository/src/repositories/legacy-juggler-bridge.ts
+++ b/packages/repository/src/repositories/legacy-juggler-bridge.ts
@@ -17,7 +17,7 @@ import {
 } from '../common-types';
 import {EntityNotFoundError} from '../errors';
 import {Entity, Model, PropertyType} from '../model';
-import {Filter, Inclusion, Where, FilterExcludingWhere} from '../query';
+import {Filter, FilterExcludingWhere, Inclusion, Where} from '../query';
 import {
   BelongsToAccessor,
   BelongsToDefinition,
@@ -458,6 +458,9 @@ export class DefaultCrudRepository<
     data: DataObject<T>,
     options?: Options,
   ): Promise<void> {
+    if (id === undefined) {
+      throw new Error('Invalid Argument: id cannot be undefined');
+    }
     const idProp = this.modelClass.definition.idName();
     const where = {} as Where<T>;
     (where as AnyObject)[idProp] = id;
--- a/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/legacy-juggler-bridge.unit.ts
@@ -609,6 +609,13 @@ describe('DefaultCrudRepository', () => {
     );
   });
 
+  it('throws error when provided an empty id', async () => {
+    const repo = new DefaultCrudRepository(Note, ds);
+    await expect(repo.updateById(undefined, {title: 't4'})).to.be.rejectedWith(
+      'Invalid Argument: id cannot be undefined',
+    );
+  });
+
   it('implements Repository.updateAll()', async () => {
     const repo = new DefaultCrudRepository(Note, ds);
     await repo.create({title: 't3', content: 'c3'});


hazimdikenli:
I can't sign the CLA due to server problems, something is wrong cla.strongloop.com, I am getting 504 ERROR all the time, I am gonna try that later.

hazimdikenli:
> I can't sign the CLA due to server problems, something is wrong cla.strongloop.com, I am getting 504 ERROR all the time, I am gonna try that later.

Looks like I was able to sign it (7 times) even though the server reported a problem.

dhmlau:
@slnode test please

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -458,6 +458,9 @@ export class DefaultCrudRepository<
     data: DataObject<T>,
     options?: Options,
   ): Promise<void> {
+    if (id === undefined) {
```
raymondfeng, line 461: Can we use `id == null` to test both `null` and `undefined`?

hazimdikenli: That is what I did initially, I was checking for both cases, but later when I tried calling updateById with null as id, it did not execute an updateAll, it simply returned EntityNotFoundError. Here is the output when updatById is called with a null id value;
`
{
  "error": {
    "statusCode": 404,
    "name": "Error",
    "message": "Entity not found: mymodel with id null",
    "code": "ENTITY_NOT_FOUND"
  }
}
`
Well you never know someone might have a nullable id field, with only one distinct null value in the database, and might want to update that one, so I changed my initial fix to check for only undefined values. Undefined is the real issue here, 
`(where as AnyObject)[idProp] = id;`
This causes an empty where statement when id is undefined, so all rows are updated as a result with the next updateAll statement;
`const result = await this.updateAll(data, where, options);`

PR #5332, opened 2020-05-04
[RFC] Add a booter to load configuration files from `configs`

raymondfeng:
The booter serves as an extension point to accept configuration loaders to call `app.configure` for configured bindings.

It comes with a built-in `JsJsonYamlLoader` to discover `configs/*` for js/json/yaml/yml files with overriding order.

The `ConfigurationLoader` receives `app, projectRoot, files` so that it can either return an object of `(bindingKey, bindingConfig)` or a Provider Class. The loader can populate `app` directly if necessary.

This will open doors to Node.js configuration frameworks such as `rc` and `node-config`.

See https://github.com/strongloop/loopback-next/blob/config-booter/docs/site/Booting-an-Application.md#configuration-booter

## Checklist


packages/context/src/binding-inspector.ts
```diff
@@ -57,9 +57,11 @@ export type BindingSpec<T = unknown> = BindingTemplate<T> | BindingScopeAndTags;
  * @typeParam T - Value type
  */
 export function isProviderClass<T>(
-  cls: Constructor<T | Provider<T>>,
+  cls: unknown,
 ): cls is Constructor<Provider<T>> {
-  return typeof cls?.prototype?.value === 'function';
+  return (
+    typeof cls === 'function' && typeof cls.prototype?.value === 'function'
```
bajtos, line 63: Please add some tests to verify this change (a new test should fail on the current master and pass with your fix in place).

bajtos:
> It comes with a built-in JsJsonYamlLoader to discover `configs/*` for `js/json/yaml/yml` files with overriding order.

When you say `configs/*`, are you referring to `configs` directory in source code repository root (`/configs`) or in the `dist` folder (`/dist/configs`)?

So far, artifacts like controllers are stored locally to `dist` folder, e.g. `src/controllers/todo.controller.ts` is compiled to `dist/controllers/todo.controller.js` where it's picked up by the controller booter. This works because `projectRoot` is set to `/dist` directory in the source code repository.

Such approach works great for TypeScript (and possibly JavaScript) files. However, it does not work for JSON/YAML files because those files are not copied from `src` to `dist` by TypeScript compiler. We used to use `lb-tsc --copy-resources` flag in the past, but that does not work with project references.

If we move config files to `configs` in repository root to accommodate JSON/YAML files, then users won't be able to use TypeScript.

Maybe we need to support two locations, both `/configs/*` and `/dist/configs/*`?

raymondfeng:
> Maybe we need to support two locations, both /configs/* and /dist/configs/*?

I had it in my original changes and realized that the current `projectRoot` does not allow so. I'll come up a solution for that.

mschnee:
This looks great.  I'm looking forward to adapting a `ProcessEnvConfigurationLoaded` ;)

packages/context/src/binding-inspector.ts
raymondfeng: See https://github.com/strongloop/loopback-next/pull/5338

raymondfeng:
I'll find some time to document the feature.

raymondfeng:
@bajtos See docs at https://github.com/strongloop/loopback-next/blob/config-booter/docs/site/Booting-an-Application.md#configuration-booter

bajtos:
@raymondfeng Before I review the new version. Are the benefits of JSON/YAML support worth the costs in additional complexity, both for us and our users? Have have recently moved away from JSON in datasource configuration, will we eventually do the same for config files? What if we mandated that all config files must be TypeScript only and see how far we can get with such design?

docs/site/Booting-an-Application.md
```diff
@@ -259,12 +259,117 @@ The options for this are passed in a `services` object on `BootOptions`.
...
+
+### Configuration Booter
+
+#### Description
+
+Discovers configuration files in `configs` and `dist/configs` folders and binds
+them to the application context using `app.configure()`.
+
+The booter also serves as an extension point to plug in configuration loaders
+that handle different form of files. By default, a configuration loader is added
```
emonddr, line 277: > different form of files

different forms of files

docs/site/Booting-an-Application.md
```diff
@@ -259,12 +259,117 @@ The options for this are passed in a `services` object on `BootOptions`.
...
+to recognize `.config.js`, `.config.json`, `.config.yaml`, and `.config.yml`
+files. Files with the same base name are loaded in the precedence of js, json,
+yaml, and yml extensions. For example, `x.config.js` shadows `x.config.json`.
+
+The js/json/yaml/yaml files export an object containing binding
```
emonddr, line 282: > js/json/yaml/yaml files 

js/json/yaml/yml files

docs/site/Booting-an-Application.md
```diff
@@ -259,12 +259,117 @@ The options for this are passed in a `services` object on `BootOptions`.
...
+`app.boot(options: BootOptions)`.
+
+The options for this are passed in a `configurations` object on `BootOptions`.
+
+Available options on the `configurations` object on `BootOptions` are as
+follows:
+
+| Options      | Type                 | Default                                                        | Description                                                                                                           |
+| ------------ | -------------------- | -------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
+| `dirs`       | `string \| string[]` | `['configs', '../configs]`                                     | Paths relative to projectRoot to look in for Service artifacts                                                        |
```
emonddr, line 369: What do you mean when you say `Service artifacts` ? Isn't this capability available to all artifacts that can be loaded and configured?

raymondfeng:
> Before I review the new version. Are the benefits of JSON/YAML support worth the costs in additional complexity, both for us and our users? Have have recently moved away from JSON in datasource configuration, will we eventually do the same for config files? What if we mandated that all config files must be TypeScript only and see how far we can get with such design?

Yes. I think this PR makes it possible to use JSON/YAML in addition to TS/JS for configurations. It will serve as an example how extensions can make `declarative` in various forms a reality.

If it's TypeScript only, we can argue this booter is no longer necessary as we can write code to call `app.configure` in the application constructor.

docs/site/Booting-an-Application.md
```diff
@@ -259,12 +259,118 @@ The options for this are passed in a `services` object on `BootOptions`.
...
+The js/json/yaml/yml files export an object containing binding key/configuration
+value pairs.
+
+The `.ts` file will be transpiled to `.js`.
+
+```ts
+export = {
+  'loggers.Log1': {
+    level: 'info',
+    __source: 'js',
```
bajtos, line 292: What is the purpose of `__source` property and why do users have to specify it explicitly? I would expect the framework to take care of that. What happens when the user makes a mistake and either forgets to add `__source` property or provides a wrong value - e.g. `ts` instead of `js`, because that feels more natural?

packages/boot/src/__tests__/acceptance/configuration.booter.acceptance.ts
```diff
@@ -0,0 +1,173 @@
...
+    subdir: false,
+  });
+
+  beforeEach('reset sandbox', () => sandbox.reset());
+  beforeEach(getApp);
+
+  it('binds configuration objects for js', async () => {
+    await copyJs();
+    await testConfigBooting('js');
+  });
```
bajtos, line 33: I find these tests super difficult to understand. One has to dig deep into several other functions and fixture files to understand what is being tested and asserted. Please find a way how to better show the intent of the tests. 

IMO, these tests are also very rigidly enforcing the same set of config entries (keys and values). When we need to test a different config use case in the future, we will be able to reuse very little and will have to write most of the code from scratch.

An example illustrating what I consider as a better solution:

```ts
it('binds configuration objects for js', async () => {
  await fs.writeFileSync('dist/configs/test.config.js', `
export = {
   'loggers.Log1': {
     level: 'info',
     _source: 'js',
   },
   'loggers.Log2': {
     level: 'debug',
     _source: 'js',
   },
 };
`);

  await app.boot();

  expect(await app.getConfig('loggers.Log1')).to.eql({level: 'info', _source: source});
  expect(await app.getConfig('loggers.Log2')).to.eql({level: 'debug', _source: source});
});
```

packages/boot/src/booters/base-artifact.booter.ts
```diff
@@ -48,7 +52,7 @@ export class BaseArtifactBooter implements Booter {
   /**
    * `glob` pattern to match artifact paths
    */
-  glob: string;
+  globs: string[];
```
bajtos, line 55: This looks like a breaking change to me, can you please confirm?

Should we preserve backwards compatibility e.g. by implementing `glob` getter/setter that will use `globs[0]` entry as the storage?

packages/boot/src/booters/base-artifact.booter.ts
```diff
@@ -37,6 +38,9 @@ export class BaseArtifactBooter implements Booter {
    * Project root relative to which all other paths are resolved
    */
   readonly projectRoot: string;
+
+  private packageRoot: string;
```
bajtos, line 42: Please add a tsdoc comment to describe this new property and update both `projectRoot` and `packageRoot` comments to explain how are these two properties different from each other.

packages/boot/src/types.ts
```diff
@@ -30,7 +30,7 @@ export type ArtifactOptions = {
    * A `glob` string to use when searching for files. This takes
```
bajtos, line 30:
```suggestion
   * A list of `glob` strings to use when searching for files. This takes
```

packages/boot/src/types.ts
```diff
@@ -30,7 +30,7 @@ export type ArtifactOptions = {
...
    * precedence over other options.
    */
-  glob?: string;
+  globs?: string | string[];
```
bajtos, line 33: This looks like a breaking change to me, can you please confirm?

Should we preserve backwards compatibility e.g. by computing the final list of glob patterns by concatenaning `globs` and `glob`?

bajtos:
> If it's TypeScript only, we can argue this booter is no longer necessary as we can write code to call `app.configure` in the application constructor.

If that was the case, then we could argue we don't need any booters because we can register all artifact kinds from application constructor.

Personally, I consider supporting `src/configs/*.ts` (`dist/configs/*.js`) files only as the sweet spot where we make it easy for LoopBack users to structure their codebase into smaller files, while keeping the complexity & maintenance costs low. By complexity I mean not only the complexity to implement support for different styles in our code, but also the complexity of explaining the different styles in our docs and the learning curve for our users.

Anyhow, LoopBack already feels way more complex to me than I would like it to be and this pull request is kind of continuing in that direction, so whatever 🤷

docs/site/Booting-an-Application.md
hacksparrow: > The `.ts` file

Which `.ts` file?

Agreed, `__source ` does not feel intuitive.

packages/boot/src/__tests__/acceptance/configuration.booter.acceptance.ts
hacksparrow: Agreed, this could be simpler and easier to read.

I personally prefer test cases to be verbose and self-contained than being normalized and "optimized". Test cases are read/revisited as individuals, having the whole context in the test case block helps a lot. Infrastructural helper functions are OK, assertion helper functions should be avoided IMO.

packages/boot/src/__tests__/fixtures/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "loggers.Log1": {
+    "level": "info",
+    "_source": "json"
```
hacksparrow, line 4: Why the underscore?

docs/site/Booting-an-Application.md
hacksparrow: Agreed, `__source ` does not feel intuitive. Also, it is a single underscore on other files.

packages/boot/src/__tests__/acceptance/configuration.booter.acceptance.ts
```diff
@@ -0,0 +1,173 @@
...
+import {
+  Application,
+  ContextTags,
+  createBindingFromClass,
+  TagValueMatcher,
+} from '@loopback/core';
+import {expect, TestSandbox} from '@loopback/testlab';
+import {Suite} from 'mocha';
+import {resolve} from 'path';
+import {ConfigurationLoader, configurationLoader} from '../../booters';
```
hacksparrow, line 15: `ConfigurationLoader` repeated.

hacksparrow:
@raymondfeng can you show how to enable/disable static files support, and specify the mount path and the public directory, using the proposed feature?

```ts
this.static('/', path.join(__dirname, '../public'));
```

docs/site/Booting-an-Application.md
raymondfeng: The real purpose of `_source` is for testing as we can assert the value is loaded from the expected source.

packages/boot/src/__tests__/acceptance/configuration.booter.acceptance.ts
raymondfeng: No, they diff in case. `ConfigurationLoader` is the type while `configurationLoader` is a decorator function.

raymondfeng:
> @raymondfeng can you show how to enable/disable static files support, and specify the mount path and the public directory, using the proposed feature?
this.static('/', path.join(__dirname, '../public'));

That's not the purpose for this PR. As @bajtos pointed out in other PRs, we differentiate configuration from wirings. The setup of `static` is wiring, not configuring.

packages/boot/src/__tests__/acceptance/configuration.booter.acceptance.ts
raymondfeng: I moved json/yaml config inline.

mdbetancourt:
> > It comes with a built-in JsJsonYamlLoader to discover `configs/*` for `js/json/yaml/yml` files with overriding order.
> 
> When you say `configs/*`, are you referring to `configs` directory in source code repository root (`/configs`) or in the `dist` folder (`/dist/configs`)?
> 
> So far, artifacts like controllers are stored locally to `dist` folder, e.g. `src/controllers/todo.controller.ts` is compiled to `dist/controllers/todo.controller.js` where it's picked up by the controller booter. This works because `projectRoot` is set to `/dist` directory in the source code repository.
> 
> Such approach works great for TypeScript (and possibly JavaScript) files. However, it does not work for JSON/YAML files because those files are not copied from `src` to `dist` by TypeScript compiler. We used to use `lb-tsc --copy-resources` flag in the past, but that does not work with project references.
> 
> If we move config files to `configs` in repository root to accommodate JSON/YAML files, then users won't be able to use TypeScript.
> 
> Maybe we need to support two locations, both `/configs/*` and `/dist/configs/*`?

I have no problem with it since im using just typescript (ts-node) without compile, then i ask why loopback by default need precompile files?

PR #5333, opened 2020-05-05
feat(context): relax parameter type for isProviderClass()

raymondfeng:
This change makes it possible to test if a unknown value is provider class
without casting to Constructor first.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5338, opened 2020-05-05
test(context): add tests for isProviderClass() check

raymondfeng:
Follow-up for #5333 

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/binding-inspector.unit.ts
+++ b/packages/context/src/__tests__/unit/binding-inspector.unit.ts
@@ -11,10 +11,11 @@ import {
   BindingScopeAndTags,
   Constructor,
   Context,
+  ContextTags,
   createBindingFromClass,
+  isProviderClass,
   Provider,
 } from '../..';
-import {ContextTags} from '../../keys';
 
 describe('createBindingFromClass()', () => {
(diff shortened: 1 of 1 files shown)


packages/context/src/__tests__/unit/binding-inspector.unit.ts
```diff
@@ -256,3 +257,52 @@ describe('createBindingFromClass()', () => {
...
+    MyJsProvider.prototype.value = function () {
+      return 'abc';
+    };
+
+    assertProviderClasses(MyProvider, MyAsyncProvider, MyJsProvider);
+  });
+
+  function assertNotProviderClasses(...values: unknown[]) {
+    for (const v of values) {
+      expect(isProviderClass(v)).to.be.false();
```
bajtos, line 299: So when one of these assertions failed, how are we going to tell which value was considered to be a provider? I think it's better to rewrite this test suite as parameterized tests.

```ts
[undefined, null, 'abc', 1, true, false, {x: 1}].forEach(val => {
  it(`recognizes ${val} as NOT a provider`, () => {
    expect(isProviderClass(val)).to.be.false();
  });
});
```

Same comment applies to other tests too.

Issue #3607, opened 2019-08-26
Model.toObject() should preserve prototypes (e.g. Date and ObjectID values)


bajtos:
## Suggestion

ATM, `toObject()` method is implemented via `toJSON` and thus converts object values to JSON representation. This is problematic e.g. for `Date` and `ObjectID` values.

I am proposing to modify `toObject()` to preserve type of Date, ObjectID and other non-model values. Of course, if a model instance has a property set to a value of another Model, _that_ property must be recursively converted via `toObject` to a plain object.

## Use Cases

Consider the following test demonstrating the problem:

```ts
@model()
class Order extends Entity {
  @property({
      type: 'string',
      id: true,
      generated: true,
    })
    id: string;

    @property()
    createdAt: Date;
}

const orderRepo = // ...;
const DATE = new Date(...);
const created = orderRepo.create({
  createdAt: DATE;
});

expect(created.toObject()).to.deepEqual({
  id: created.id,
  createdAt: DATE,
});
```

The assertion above fails now.

## Examples

The test example above should pass.

## Acceptance criteria

- [ ] `.toObject()` is updated to handle `Date`,`ObjectID`, and other non-model values, and handle recursion into fields that are also `Model` objects. See : https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L422
- [ ] Update or add mocha test cases in https://github.com/strongloop/loopback-datasource-juggler/tree/master/test 
- [ ] Update documentation

hacksparrow:
Agreed!

emonddr:
@bajtos I added the Acceptance Criteria. Please adjust it to your liking; if necessary. thx

Also what is an example of non-model value?

bajtos:
> I added the Acceptance Criteria.

Great, thank you @emonddr!

> Also what is an example of non-model value

A plain data object. For example:

```ts
@model()
class Foo extends Entity {
  @property({id: true})
    id: number;

    @property()
    data: object;
}

const foo = new Foo({id: 1, data: {key: 'value'}});
expect(foo.toObject().data).to.deepEqual({key: 'value'});
```

I think the conversion logic should work as follows:

- If the value is not an object, then keep it as-is
- If the value is an array, then (recursively) convert all items of that array
- If the value has `.toObject` method, then call it to perform the conversion
- Otherwise keep the value as-is

[...]

PR #5348, opened 2020-05-06
feat(repository): to object preserves prototype

jannyHou:
Implements https://github.com/strongloop/loopback-next/issues/3607
Discussion see https://github.com/strongloop/loopback-next/issues/3607#issuecomment-624050020

The new conversion rules are:

> toObject() preserves type of Date, ObjectID and other non-model values. Of course, if a model instance has a property set to a value of another Model, that property must be recursively converted via toObject to a plain object.

## Checklist

### DIFF
--- a/packages/repository/src/model.ts
+++ b/packages/repository/src/model.ts
@@ -189,13 +189,21 @@ function asJSON(value: any): any {
   return value;
 }
 
+/**
+ * Convert a value to a plain object as DTO.
+ *
+ * - The prototype of the value in primitive types are preserved,
+ *   like `Date`, `ObjectId`.
+ * - If the value is an instance of custom model, call `toObject` to convert.
+ * - If the value is an array, convert each element recursively.
+ *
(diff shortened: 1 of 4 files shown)


packages/repository/src/model.ts
```diff
@@ -249,18 +257,36 @@ export abstract class Model {
...
       for (const p in this) {
         const val = (this as AnyObject)[p];
         obj[p] = asObject(val, options);
       }
-    } else {
-      obj = this.toJSON();
+      return obj;
     }
+
+    const def = (<typeof Model>this.constructor).definition;
```
bajtos, line 278: _A nitpick, feel free to ignore._

I believe we prefer `as` casts over `<>`?

```suggestion
    const def = (this.constructor as typeof Model).definition;
```

jannyHou: yep agreed, and I also refactored the one in `toJson`  from `<>` to `as`

PR #5339, opened 2020-05-05
feat: try passport login example with express middleware

deepakrkris:
## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5357, opened 2020-05-07
feat(context): force interceptor functions to return a value

raymondfeng:
The `NonVoid` return type prohibits an interceptor function from having
void return type. This is important that an interceptor should either
return the value from downstream interceptors or a value from its own.

For example, the code will fail to compile.
```ts
 const myInterceptor: Interceptor = async (ctx, next) {
    // preprocessing
    // ...
 
    // subtle bug that the returned value from `next` is not returned
    // back to the upstream interceptors
    next();
 
   // postprocessing
    // ...
    // We must have `return ...` here
 }
```

Related to #5339 

## Checklist

### DIFF
--- a/packages/core/src/__tests__/unit/application.unit.ts
+++ b/packages/core/src/__tests__/unit/application.unit.ts
@@ -441,7 +441,9 @@ describe('Application', () => {
       });
     });
 
-    function logInterceptor(ctx: InvocationContext, next: Next) {}
+    function logInterceptor(ctx: InvocationContext, next: Next) {
+      return undefined;
+    }
 
     @bind(asGlobalInterceptor())
     class LogInterceptorProvider implements Provider<Interceptor> {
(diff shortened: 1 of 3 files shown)


raymondfeng:
Unfortunately a negative case will fail to compile.

bajtos:
@raymondfeng 

> Unfortunately a negative case will fail to compile.

TypeScript 3.9 is adding `ts-expect-error` to support exactly this kind of tests, see https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/#ts-expect-error-comments

I am not sure if we are using 3.9 yet, there is an older (but less strict) alternative `ts-ignore` we can already use.

packages/core/src/__tests__/unit/application.unit.ts
```diff
@@ -441,7 +441,9 @@ describe('Application', () => {
       });
     });
 
-    function logInterceptor(ctx: InvocationContext, next: Next) {}
+    function logInterceptor(ctx: InvocationContext, next: Next) {
+      return undefined;
```
bajtos, line 445: I find this redundant and a bit silly, because in JavaScript, functions return `undefined` when there is no `return` statement. I guess the extra verbosity is worth the stronger type safety your change provides.

PR #5359, opened 2020-05-07
build: fix VSCode problem matcher for the build task 

bajtos:
TypeScript compiler enables `pretty` mode when using the build mode (project references). Unfortunately, pretty mode is not supported by VSCode's problem matcher and therefore VSCode is not able to parse the errors to populate the Problems window.

This change addes `--pretty false` arguments to `tsc` when invoked from VSCode task "Build project".

In order to make this work, I had to also fix `lb-tsc` to correctly recognize and pass boolean arguments with a value. Before this change, `lb-tsc -b --pretty false` would discard `--pretty false` arguments.

## Checklist

### DIFF
--- a/packages/build/bin/compile-package.js
+++ b/packages/build/bin/compile-package.js
@@ -164,8 +164,10 @@ function validArgsForBuild(args) {
   });
   let validArgs = args;
   if (args.includes('-b') || args.includes('--build')) {
-    validArgs = filterArgs(args, arg => {
-      if (validBooleanOptions.includes(arg)) return 1;
+    validArgs = filterArgs(args, (arg, next) => {
+      if (validBooleanOptions.includes(arg)) {
+        return next === 'false' || next === 'true' ? 2 : 1;
+      }
       if (validValueOptions.includes(arg)) return 2;
       return 0;
     });
@@ -185,7 +187,7 @@ function filterArgs(args, filter) {
   const validArgs = [];
   let i = 0;
   while (i < args.length) {
-    const length = filter(args[i]);
+    const length = filter(args[i], args[i + 1]);
     if (length === 0) {
       i++;
     } else if (length === 1) {
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -88,7 +88,10 @@
       "args": [
         "run",
         "-s",
-        "build"
+        "build",
+        "--",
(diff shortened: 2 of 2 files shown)


packages/build/bin/compile-package.js
```diff
@@ -164,8 +164,10 @@ function validArgsForBuild(args) {
   });
   let validArgs = args;
   if (args.includes('-b') || args.includes('--build')) {
-    validArgs = filterArgs(args, arg => {
-      if (validBooleanOptions.includes(arg)) return 1;
+    validArgs = filterArgs(args, (arg, next) => {
+      if (validBooleanOptions.includes(arg)) {
+        return next === 'false' || next === 'true' ? 2 : 1;
+      }
```
raymondfeng, line 170: Does `tsc` allow `no-prettier`?

packages/build/bin/compile-package.js
```diff
@@ -185,7 +187,7 @@ function filterArgs(args, filter) {
   const validArgs = [];
   let i = 0;
   while (i < args.length) {
-    const length = filter(args[i]);
+    const length = filter(args[i], args[i + 1]);
```
jannyHou, line 190: wouldn't `args[i+1]` exceeds the array's index? 
e.g. `args.length = 5`, `i = 4`, `args[5]` doesn't exist

emonddr: you need `args.length -1`

packages/build/bin/compile-package.js
bajtos: No, it does not.

```
$ tsc -b --no-prettier
error TS5072: Unknown build option '--no-prettier'.
```

packages/build/bin/compile-package.js
bajtos: > wouldn't `args[i+1]` exceeds the array's index? 
> e.g. `args.length = 5`, `i = 4`, `args[5]` doesn't exist

Yes, `i+1` can exceed the array index. That's fine, `args[i+1]` will be resolved as `undefined` in such case.

```
$ node
> list = [1,2,3]; list[5]
undefined
```

bajtos:
@raymondfeng @jannyHou @emonddr I responded to your comments, PTAL. Does the patch LGTY now?

PR #5366, opened 2020-05-08
feat: introduce middleware based sequence for REST

raymondfeng:
This PR adds a middleware based sequence and wrap existing actions as middleware.

## Checklist


packages/rest/src/sequence.ts
```diff
@@ -116,3 +116,77 @@ export class DefaultSequence implements SequenceHandler {
...
+   * methods as promises.
+   *
+   * @param invokeMiddleware - invoker for registered middleware in a chain.
+   * To be injected via SequenceActions.INVOKE_MIDDLEWARE.
+   */
+  constructor(
+    @inject(SequenceActions.INVOKE_MIDDLEWARE)
+    readonly invokeMiddleware: InvokeMiddleware,
+    @config()
+    readonly options: InvokeMiddlewareOptions[] = [],
```
deepakrkris, line 135: can you use `.configure() ` to inject middleware options, instead of defaulting it ?

packages/rest/src/sequence.ts
```diff
@@ -116,3 +116,77 @@ export class DefaultSequence implements SequenceHandler {
...
+  ) {
+    if (this.options.length === 0) {
+      this.options = [
+        {
+          chain: 'middlewareChain.rest',
+          orderedGroups: [
```
deepakrkris, line 141: `LogErrorProvider ` is not included here

packages/rest/src/sequence.ts
```diff
@@ -116,3 +116,77 @@ export class DefaultSequence implements SequenceHandler {
...
+   * Runs the default sequence. Given a handler context (request and response),
+   * running the sequence will produce a response or an error.
+   *
+   * Default sequence executes these steps
+   *  - Finds the appropriate controller method, swagger spec
+   *    and args for invocation
+   *  - Parses HTTP request to get API argument list
+   *  - Invokes the API which is defined in the Application Controller
+   *  - Writes the result from API into the HTTP response
+   *  - Error is caught and logged using 'logError' if any of the above steps
```
deepakrkris, line 174: `logError` is not included in the options

deepakrkris: also how is an error redirected the `LogErrorProvider`, is that logic part of the new express package ?

packages/rest/src/providers/invoke-method.provider.ts
```diff
@@ -19,3 +20,23 @@ export class InvokeMethodProvider implements Provider<InvokeMethod> {
...
+    protected invokeMethod: InvokeMethod,
+  ) {}
+
+  value(): Middleware {
+    return async (ctx, next) => {
+      const route: RouteEntry = await ctx.get(RestBindings.Operation.ROUTE);
+      const params: OperationArgs = await ctx.get(
+        RestBindings.Operation.PARAMS,
+      );
+      return this.invokeMethod(route, params);
```
deepakrkris, line 39: why is `next` not called here ? for this to be an independent middleware function, it should have no knowledge that it is the last in the middleware chain. The invocation part should handle that.

raymondfeng: It's the middleware logic's decision not to call `next` as `invokeMethod` returns the result.

packages/rest/src/sequence.ts
raymondfeng: logError/LogErrorProvider are not middleware. They are injected into `SendResponseMiddleware` to log errors.

packages/rest/src/sequence.ts
raymondfeng: See https://github.com/strongloop/loopback-next/pull/5366#discussion_r422257512

packages/rest/src/sequence.ts
raymondfeng: The `@config` is here to allow `app.configure` to supply the config value.

packages/rest/src/providers/invoke-method.provider.ts
deepakrkris: ok I see; so next() is just a customary function for the middleware, it does not really link with the middleware function in the chain ?

deepakrkris: that is why the sequence handle is forced to invoke the middleware functions one by one 
 ``` 
for (const opt of this.options) {
      // Invoke registered Express middleware
      debug(
        'Invoking middleware chain %s with groups %s',
        opt.chain,
        opt.orderedGroups,
      );
      const done = await this.invokeMiddleware(context, opt);
      if (done) break;
    } 
```

Comment (raymondfeng) [reply packages/rest/src/providers/invoke-method.provider.ts:39]: > ok I see; so next() is just a customary function for the middleware, it does not really link with the next middleware function in the chain ?

It does. But it's up the middleware logic to decide if the next one should be invoked. The `invokeMethod` is designed to invoke the controller method directly. 

> why sequence.handle() is forced to invoke the middleware functions one by one?
This is to allow multiple chains to be used one by one. If you need to cascade them, add the chain as a middleware or flatten the chains.

**Comment by raymondfeng** [comment]
@bajtos I know you might reject the PR :-).

Yes, I would like to move to a middleware based pipeline by composition. The sequence can stay as the integration point to the application and it can facilitate the transition too. Here are a few thoughts that drive me to come up this PR:

1. To avoid future issues similar as https://github.com/strongloop/loopback-next/issues/5368. The framework needs to have some control over the sequence. 

2. I never see a need to override the sequence code so far beyond adding additional actions. The extensibility of the sequence has already partially solved by the introduction of `InvokeMiddleware`. At bare minimum, we can either get rid of the generated sequence class or should simply have it extend `DefaultSequence`.

3. The `dependency` based ordering is inspired by https://hapipal.com/best-practices/handling-plugin-dependencies.

4. I intentionally didn't remove action providers. Instead, I created middleware provider to leverage existing action implementations. This is an effort to keep it backward-compatible.

5. Having a few hard-coded actions beyond other middleware is an inconsistency and it makes it hard to manage all steps in a consistent way.

6. The PR meant to propose another option and it can co-exist with the current sequence style - which can be removed later if our users are happy with the new one.

**Comment by raymondfeng** [comment]
Here is another reason to unify all actions as middleware - https://github.com/strongloop/loopback-next/pull/5562#issuecomment-634434376

**Comment by raymondfeng** [comment]
Justification #2:

With the current sequence, the `invokeMiddleware` cannot be around `invokeMethod`. As a result, to add a middleware to measure the response time, I have to tweak the sequence to be:

```ts
import {DefaultSequence, RequestContext} from '@loopback/rest';

export class MySequence extends DefaultSequence {
  async handle(context: RequestContext): Promise<void> {
    // Invoke registered Express middleware
    await this.invokeMiddleware(context, {
      next: async () => {
        try {
          this.send(response, result);
          return result;
        } catch (error) {
          this.reject(context, error);
        }
      },
    });
  }
}

```

raymondfeng:
@bajtos I have a few solid use cases that exposes limitations of the current sequence implementation.

1. At the moment, we cannot add `authentication` as a middleware without changing the generated sequence class. The `authentication` needs to be invoked after `findRoute` - which is hard-coded. We'll have to ask our application developers to add another `invokeMiddleware`  after `findRoute`. See https://github.com/strongloop/loopback-next/pull/5562#issuecomment-634434376.

2. With the current sequence, the invokeMiddleware cannot be around invokeMethod. As a result, to add a middleware to measure the response time, I have to tweak the sequence to be:

```ts
import {DefaultSequence, RequestContext} from '@loopback/rest';

export class MySequence extends DefaultSequence {
  async handle(context: RequestContext): Promise<void> {
    // Invoke registered Express middleware
    await this.invokeMiddleware(context, {
      next: async () => {
        try {
          this.send(response, result);
          return result;
        } catch (error) {
          this.reject(context, error);
        }
      },
    });
  }
}
```

Mixing middleware and hard-coded actions makes it very difficult to control the sequence as the cascading middleware are interrupted by hard-coded actions that don't delegate control to downstream actions/middleware.

bajtos:
> 1. At the moment, we cannot add `authentication` as a middleware without changing the generated sequence class. The `authentication` needs to be invoked after `findRoute` - which is hard-coded. We'll have to ask our application developers to add another `invokeMiddleware`  after `findRoute`. See [#5562 (comment)](https://github.com/strongloop/loopback-next/pull/5562#issuecomment-634434376).

IMO, this is a matter of preference. Personally, I like to understand how is the request handling workflow composed of individual action and have easy time troubleshooting possible issues. The current (Sequence-based) approach is shining here. I don't mind to add the call of an authentication action manually, it's the price I am happy to pay.

Having said that, I also see how other people may prefer more automagic where it's enough to add a component to an application and everything else will be wired up automatically. Personally, I think we are going too far here and will cause a lot of confusing and long debugging sessions for our users.

It looks like I am the only one concerned about these aspects and since I don't have appetite to invest the necessary time & energy to research and argumentation how to keep Sequence style, I feel it would be counter-productive to keep my opposing position. I am going to withdraw my rejection and let the code owners to continue the review process.

> Mixing middleware and hard-coded actions makes it very difficult to control the sequence as the cascading middleware are interrupted by hard-coded actions that don't delegate control to downstream actions/middleware.

I am not surprised, I think we kind of knew from the beginning that middleware and Sequence are very different concepts that may not work together.

Please consider the developer experience LoopBack users and try to find a way how to document the old Sequence style, the new Middleware style and the transition process to confuse our users as little as possible.

Few user stories to consider:

1. As a new LoopBack user building a REST API application, I want to learn how to add new request-preprocessing steps like CORS, helmet or authentication. I'd like the documentation to show me the idiomatic solution (using middleware?), I don't want to be overwhelmed with legacy concepts from the past.

2. As an existing LoopBack user working on a REST API application, I want to understand whether I can keep using the existing sequence-based style, for how long it will be supported, and what is the migration path to the new style.

3. As a maintainer of an existing LoopBack 4 extension, I want to understand how to migrate my component to the new style.

4. As a LB project lead/manager, I'd like to understand what effort is required on our side to write the required documentation, update our examples, etc. We need a list of well-defined small(ish) user stories (GitHub issues) that the team can estimate and we can plan into sprints.

hacksparrow:
> If we want to go all-in on middleware, then it may be better move away from Sequence entirely and use pure middleware-based composition of request handling flow.

+1

Too many ways of doing something leads to bug-prone, hard to debug programs. I feel the API surface of LoopBack is pretty vast already. We should take a good amount of time to seriously consider this direction. If we decide we go the middleware way, we should absolutely get rid of the sequence as a user API.

Let's add this as one of the tasks for July - https://github.com/strongloop/loopback-next/pull/5666.

hacksparrow:
Ok, here's my update after spending some more time on the PR.

I agree with @bajtos's view that it is disruptive. However, I am in favor of the PR with the following points.

1. @raymondfeng has mentioned of a case where middleware functionality depends on the result of  `findRoute`; to make the middleware work, `invokeMiddleware` has to be called again. In case there are middleware which depend on the results of `parseParams` and `invoke` too, it would mean calling `invokeMiddleware` four times in the sequence. It is very obvious something is not right.
2. In the course of a journey, if we discover we haven't taken the right path, we should be open to changing it.
3. LB4 is still relatively new, making changes now would be better than at a later time.
4. The sequence handler is a less frequently customized area. So this is would mean way less disruption compared to say, changing the API of controllers.
5. We do not remove the `DefaultSequence` but keep it "frozen". So no disruption at all.
6. Raymond has updated all (?) the examples to use `MiddlewareSequence`. All our new examples would be based on `MiddlewareSequence` - we contain the disruption to ourselves.
7. We put in the necessary effort on documentation explaining the new way, along with a note about the old way. I think the current doc changes in the PR is not enough. Eg: https://loopback.io/doc/en/lb4/Routing-requests.html and https://loopback.io/doc/en/lb4/Parsing-requests.html also have to be updated/re-organized. There may be other areas too. Maybe we update the examples and docs in a separate PR? Because it may be overwhelming to do everything in this PR.

This looks like a case of short term pain for a longer term gain to me. In my opinion, we should go with it.

docs/site/REST-middleware-sequence.md
```diff
@@ -0,0 +1,846 @@
...
+    }
+  }
+}
+```
+
+The legacy `Sequence` is now deprecated but it should continue to work without
+any changes.
+
+If you have never customized `src/sequence.ts` generated by `lb4 app`, or you
+have only modified the code by adding `authenticate` action per our docs, you
```
jannyHou, line 241: can refer to https://loopback.io/doc/en/lb4/Authentication-component-action.html

docs/site/REST-middleware-sequence.md
```diff
@@ -0,0 +1,846 @@
...
+
+    - upstreamGroups: An array of group names that should be upstream to this
+      middleware
+
+      ```ts
+      @bind(
+        asMiddleware({
+          chain: RestTags.REST_MIDDLEWARE_CHAIN,
+          group: 'authentication',
+          upstreamGroups: ['cors', 'findRoute'],
```
jannyHou, line 301: A question: would `asMiddleware` make sure the group name is really after the upstream groups? same question for downstream groups.

docs/site/REST-middleware-sequence.md
```diff
@@ -0,0 +1,846 @@
...
+      'Invoking middleware chain %s with groups %s',
+      this.options.chain,
+      this.options.orderedGroups,
+    );
+    await this.invokeMiddleware(context, this.options);
+  }
+}
+```
+
+![middleware-sequence](imgs/middleware-sequence.png)
```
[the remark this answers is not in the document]
raymondfeng: good catch.

docs/site/REST-middleware-sequence.md
raymondfeng: `asMiddleware` adds tags to the binding to express its ordering requirement and we will use these partial rules to line up all middleware.

raymondfeng:
@hacksparrow @jannyHou PTAL.

packages/rest/src/sequence.ts
```diff
@@ -121,3 +122,83 @@ export class DefaultSequence implements SequenceHandler {
...
 }
+
+/**
+ * A sequence implementation using middleware chains
+ */
+export class MiddlewareSequence implements SequenceHandler {
+  static defaultOptions: InvokeMiddlewareOptions = {
+    chain: RestTags.REST_MIDDLEWARE_CHAIN,
+    orderedGroups: [
+      // Please note that middleware is cascading. The `SendResponse` is
```
hacksparrow, line 133: `SendResponse` -> `sendResponse`

packages/rest/src/sequence.ts
```diff
@@ -121,3 +122,83 @@ export class DefaultSequence implements SequenceHandler {
...
+   *  - `cors`: Enforces `CORS`
+   *  - `openApiSpec`: Serves OpenAPI specs
+   *  - `findRoute`: Finds the appropriate controller method, swagger spec and
+   *    args for invocation
+   *  - `parseParams`: Parses HTTP request to get API argument list
+   *  - `invokeMethod`: Invokes the API which is defined in the Application
+   *    controller method
+   *
+   * In front of the groups above, we have a special middleware called
+   * `SendResponse`, which first invokes downstream middleware to get a result
```
hacksparrow, line 185: `SendResponse` -> `sendResponse`

packages/rest/src/sequence.ts
```diff
@@ -121,3 +122,83 @@ export class DefaultSequence implements SequenceHandler {
...
   }
 }
+
+/**
+ * A sequence implementation using middleware chains
+ */
+export class MiddlewareSequence implements SequenceHandler {
+  static defaultOptions: InvokeMiddlewareOptions = {
+    chain: RestTags.REST_MIDDLEWARE_CHAIN,
+    orderedGroups: [
```
hacksparrow, line 132: Why is this named `orderedGroups`? Do `sendResponse`, `cors`, `apiSpec` etc., resolve to arrays of middleware? If not, the name is misleading. If yes, the fact has to be clearly documented.

packages/rest/src/sequence.ts
```diff
@@ -121,3 +122,83 @@ export class DefaultSequence implements SequenceHandler {
...
+      // Please note that middleware is cascading. The `SendResponse` is
+      // added first to invoke downstream middleware to get the result or
+      // catch errors so that it can produce the http response.
+      'sendResponse',
```
hacksparrow, line 136: If these are actually groups, can we use constants instead of string names here?

packages/rest/src/providers/parse-params.provider.ts
```diff
@@ -36,3 +37,26 @@ export class ParseParamsProvider implements Provider<ParseParams> {
       });
   }
 }
+
+@bind(
+  asMiddleware({
+    group: 'parseParams',
+    upstreamGroups: 'findRoute',
+    chain: RestTags.REST_MIDDLEWARE_CHAIN,
```
hacksparrow, line 45: `parseParams`, `findRoute ` and other strings are being repeated, I think we should be using constants.

packages/rest/src/providers/invoke-method.provider.ts
```diff
@@ -19,3 +20,29 @@ export class InvokeMethodProvider implements Provider<InvokeMethod> {
     return route.invokeHandler(this.context, args);
   }
 }
+
+@bind(
+  asMiddleware({
+    group: 'invokeMethod',
```
hacksparrow, line 26: So, looks like `group` is to LB4, what middleware phase was to LB3, right? @raymondfeng

packages/rest/src/sequence.ts
```diff
@@ -121,3 +122,83 @@ export class DefaultSequence implements SequenceHandler {
...
+      'apiSpec',
+
+      // default
+      'middleware',
+
+      // rest
+      'findRoute',
+
+      // authentication
+      'authentication',
```
hacksparrow, line 149: For this to work, user has to create a `Provider<Middleware>` and set its `group` value to `authentication`; and many such authentication middleware can be added and anyone of them can cause the authentication to fail, right?

packages/rest/src/sequence.ts
hacksparrow: After going through the codebase, it has become clear to me. A name like `executionOrder` or something like that might make it more obvious.

raymondfeng: We inherit `orderedGroups` from interceptors - middleware is a specialized interceptors that are invoked as part of the sequence.

packages/rest/src/providers/invoke-method.provider.ts
raymondfeng: yes.

packages/rest/src/sequence.ts
raymondfeng: A group does not need to have middleware. It's only for ordering purpose.

raymondfeng:
@hacksparrow Thank you for the feedback. I believe that I have addressed your comments. PTAL.

hacksparrow:
What purpose does this serve:

```
orderedGroups: [
  RestMiddlewareGroups.SEND_RESPONSE,
  RestMiddlewareGroups.CORS,
  RestMiddlewareGroups.API_SPEC,
  RestMiddlewareGroups.MIDDLEWARE,
  RestMiddlewareGroups.FIND_ROUTE,
  RestMiddlewareGroups.AUTHENTICATION,
  RestMiddlewareGroups.PARSE_PARAMS,
  RestMiddlewareGroups.INVOKE_METHOD,
],
```

if we still have to do this?

```
group: RestMiddlewareGroups.INVOKE_METHOD,
upstreamGroups: RestMiddlewareGroups.PARSE_PARAMS,
```

I can understand for middleware not in the ordered group, but for those in the group - shouldn't they "know" it? And what happens if there is a conflict:

```
group: RestMiddlewareGroups.INVOKE_METHOD,
upstreamGroups: RestMiddlewareGroups.SEND_RESPONSE
```
?

I think we should add tests for the expected behavior.

raymondfeng:
@hacksparrow The final order of middleware is settled using both the `orderedGroup` and relative `upstreamGroups/downstreamGroups`.

The `orderedGroups` defines the overall order so that not all middleware are required to spell out `upstreamGroups/downstreamGroups`. Both settings contribute partial orders. We combine all rules and calculate the final order using topological sort.

I'll try to add some tests.

raymondfeng:
@hacksparrow I added a test (https://github.com/strongloop/loopback-next/blob/sequence-action-as-middleware/packages/express/src/__tests__/acceptance/middleware-registeration.acceptance.ts#L86) and more documents (https://github.com/strongloop/loopback-next/blob/sequence-action-as-middleware/docs/site/REST-middleware-sequence.md#sort-middleware-by-groups).

PR #5367, opened 2020-05-08
fix: minor fixes to middleware group and binding template

raymondfeng:
1. Consistently use `DEFAULT_MIDDLEWARE_GROUP`
2. Change return type to `BindingTemplate` for `asMiddleware`.

## Checklist

### DIFF
--- a/packages/express/src/middleware.ts
+++ b/packages/express/src/middleware.ts
@@ -127,7 +127,7 @@ export function registerExpressMiddleware<CFG>(
  */
 export function asMiddleware(
   options: MiddlewareBindingOptions = {},
-): BindingTemplate<Middleware> {
+): BindingTemplate {
   return function middlewareBindingTemplate(binding) {
     binding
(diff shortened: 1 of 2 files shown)


packages/express/src/middleware.ts
```diff
@@ -127,11 +127,11 @@ export function registerExpressMiddleware<CFG>(
  */
 export function asMiddleware(
   options: MiddlewareBindingOptions = {},
-): BindingTemplate<Middleware> {
```
deepakrkris, line 130: is this for a fix or will the ts compiler detect `Middleware` from return type ?

raymondfeng: This is a fix for TS compiler to allow `asMiddleware()` to be used with `@bind`. Before this PR, the following code fails to compile as `BindingTemplate<Middleware>` cannot be assigned to `BindingTemplate<unknown>`.

```ts
@bind(asMiddleware('my-group')(
```

PR #5370, opened 2020-05-08
feat(context): provide resolution context metadata for factory functions with toDynamicValue()

raymondfeng:
A provider class can use dependency injection to receive resolution-related
metadata such as `context` and `binding`. But the overhead to wrap a factory
function is not desired for some use cases. This PR introduces a lightweight
alternative using `toDynamicValue` as follows:

1. Add a parameter to the factory function to receive resolution context
2. Support a specialized class with a static `value` method that allows
parameter injection.

Some benchmark tests are added to `benchmark` to measure performances for
various flavors of the dynamic value resolution for bindings.

See https://github.com/strongloop/loopback-next/tree/pass-ctx-to-dynamic-value/benchmark/src/context-binding for examples.

This change is backward compatible. Existing factory functions continue to work.

## Checklist

### DIFF
--- a/benchmark/src/context-binding/README.md
+++ b/benchmark/src/context-binding/README.md
@@ -1,0 +1,30 @@
+# Context binding benchmark
+
+This directory contains a simple benchmarking to measure the performance of
+different styles of context bindings.
+
+## Basic use
+
+```sh
+npm run -s benchmark:context
+```
+
+For example:
+
+```
+npm run -s benchmark:context
+```
+
+## Base lines
+
+| Test                      | Ops/sec   | Relative margin of error | Runs sampled | Count |
+| ------------------------- | --------- | ------------------------ | ------------ | ----- |
+| factory - getSync         | 1,282,238 | ±1.11%                   | 93           | 68537 |
+| factory - get             | 1,222,587 | ±1.19%                   | 87           | 66558 |
+| asyncFactory - get        | 362,457   | ±1.78%                   | 78           | 23284 |
+| staticProvider - getSync  | 484,494   | ±1.04%                   | 92           | 26260 |
+| staticProvider - get      | 475,130   | ±1.13%                   | 86           | 27435 |
+| asyncStaticProvider - get | 339,359   | ±1.43%                   | 86           | 19046 |
+| provider - getSync        | 368,127   | ±1.14%                   | 87           | 21162 |
+| provider - get            | 366,649   | ±0.86%                   | 86           | 21358 |
+| asyncProvider - get       | 291,054   | ±1.94%                   | 82           | 16557 |
--- a/benchmark/src/context-binding/context-binding.ts
+++ b/benchmark/src/context-binding/context-binding.ts
@@ -1,0 +1,136 @@
+// Copyright IBM Corp. 2018,2020. All Rights Reserved.
+// Node module: @loopback/benchmark
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {Context, inject, Provider, ValueFactory} from '@loopback/context';
+import Benchmark from 'benchmark';
+
+/**
+ * Option 1 - use a sync factory function
+ */
+const factory: ValueFactory = ({context}) => {
+  const user = context.getSync('user');
+  return `Hello, ${user}`;
+};
+
+/**
+ * Option 2 - use an async factory function
+ */
+const asyncFactory: ValueFactory = async ({context}) => {
+  const user = await context.get('user');
+  return `Hello, ${user}`;
+};
+
+/**
+ * Option 3 - use a value factory provider class with sync static value() method
+ * parameter injection
+ */
+class StaticGreetingProvider {
+  static value(@inject('user') user: string) {
+    return `Hello, ${user}`;
+  }
+}
+
+/**
+ * Option 4 - use a value factory provider class with async static value() method
+ * parameter injection
+ */
+class AsyncStaticGreetingProvider {
+  static value(@inject('user') user: string) {
+    return Promise.resolve(`Hello, ${user}`);
+  }
+}
+
+/**
+ * Option 5 - use a regular provider class with sync value()
+ */
+class GreetingProvider implements Provider<string> {
+  @inject('user')
+  private user: string;
+
+  value() {
+    return `Hello, ${this.user}`;
+  }
+}
+
+/**
+ * Option 6 - use a regular provider class with async value()
+ */
+class AsyncGreetingProvider implements Provider<string> {
+  @inject('user')
+  private user: string;
+
+  value() {
+    return Promise.resolve(`Hello, ${this.user}`);
+  }
+}
+
+setupContextBindings();
+
+function setupContextBindings() {
+  const ctx = new Context();
+  ctx.bind('user').to('John');
+  ctx.bind('greeting.syncFactory').toDynamicValue(factory);
+  ctx.bind('greeting.asyncFactory').toDynamicValue(asyncFactory);
+  ctx
+    .bind('greeting.syncStaticProvider')
+    .toDynamicValue(StaticGreetingProvider);
+  ctx
+    .bind('greeting.asyncStaticProvider')
+    .toDynamicValue(AsyncStaticGreetingProvider);
+  ctx.bind('greeting.syncProvider').toProvider(GreetingProvider);
+  ctx.bind('greeting.asyncProvider').toProvider(AsyncGreetingProvider);
+  return ctx;
+}
+
+function runBenchmark(ctx: Context) {
+  const options: Benchmark.Options = {
+    initCount: 1000,
+    onComplete: (e: Benchmark.Event) => {
+      const benchmark = e.target as Benchmark;
+      console.log('%s %d', benchmark, benchmark.count);
+    },
+  };
+  const suite = new Benchmark.Suite('context-bindings');
+  suite
+    .add(
+      'factory - getSync',
+      () => ctx.getSync('greeting.syncFactory'),
+      options,
+    )
+    .add('factory - get', () => ctx.get('greeting.syncFactory'), options)
+    .add('asyncFactory - get', () => ctx.get('greeting.asyncFactory'), options)
+    .add(
+      'staticProvider - getSync',
+      () => ctx.getSync('greeting.syncStaticProvider'),
+      options,
+    )
+    .add(
+      'staticProvider - get',
+      () => ctx.get('greeting.syncStaticProvider'),
+      options,
+    )
+    .add(
+      'asyncStaticProvider - get',
+      () => ctx.get('greeting.asyncStaticProvider'),
+      options,
+    )
+    .add(
+      'provider - getSync',
+      () => ctx.getSync('greeting.syncProvider'),
+      options,
+    )
+    .add('provider - get', () => ctx.get('greeting.syncProvider'), options)
+    .add(
+      'asyncProvider - get',
+      () => ctx.get('greeting.asyncProvider'),
+      options,
+    )
+    .run({async: true});
+}
+
+if (require.main === module) {
+  const ctx = setupContextBindings();
+  runBenchmark(ctx);
+}
--- a/benchmark/tsconfig.json
+++ b/benchmark/tsconfig.json
@@ -14,6 +14,9 @@
       "path": "../examples/todo/tsconfig.json"
     },
     {
+      "path": "../packages/context/tsconfig.json"
+    },
+    {
       "path": "../packages/openapi-spec-builder/tsconfig.json"
     },
     {
--- a/docs/site/Binding.md
+++ b/docs/site/Binding.md
@@ -83,6 +83,47 @@ binding.toDynamicValue(() => new Date());
 binding.toDynamicValue(() => Promise.resolve('my-value'));
 ```
 
+The factory function can receive extra information about the context, binding,
+and resolution options.
+
+```ts
+import {ValueFactory} from '@loopback/context';
+
+// The factory function now have access extra metadata about the resolution
+const factory: ValueFactory<string> = resolutionCtx => {
+  return `Hello, ${resolutionCtx.context.name}#${
+    resolutionCtx.binding.key
+  } ${resolutionCtx.options.session?.getBindingPath()}`;
+};
+const b = ctx.bind('msg').toDynamicValue(factory);
+```
+
+Object destructuring can be used to further simplify a value factory function
+that needs to access `context`, `binding`, or `options`.
+
+```ts
+const factory: ValueFactory<string> = ({context, binding, options}) => {
+  return `Hello, ${context.name}#${
+    binding.key
+  } ${options.session?.getBindingPath()}`;
+};
+```
+
+An advanced form of value factory is a class that has a static `value` method
+that allows parameter injection.
+
+```ts
+import {inject} from '@loopback/context';
+
+class GreetingProvider {
+  static value(@inject('user') user: string) {
+    return `Hello, ${user}`;
+  }
+}
+
+const b = ctx.bind('msg').toDynamicValue(GreetingProvider);
+```
+
 #### A class
 
 The binding can represent an instance of a class, for example, a controller. A
@@ -118,6 +159,9 @@ class MyValueProvider implements Provider<string> {
 
 binding.toProvider(MyValueProvider);
 ```
+
+The provider class serves as the wrapper to declare dependency injections. If
+dependency is not needed, `toDynamicValue` can be used instead.
 
 #### An alias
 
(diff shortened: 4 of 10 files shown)


docs/site/Binding.md
```diff
@@ -83,6 +83,21 @@ binding.toDynamicValue(() => new Date());
...
+
+// The factory function now have access extra metadata about the resolution
+const factory: ValueFactory<string> = (_ctx, _binding, options) => {
+  return `Hello, ${_ctx.name}#${
+    _binding.key
+  } ${options.session?.getBindingPath()}`;
+};
+const b = ctx.bind('msg').toDynamicValue(factory);
+```
+
```
bajtos, line 100: In the past, I was against allowing dynamic value factories to access the context object, because it makes it too easy to slip into using Service Locator design pattern (`ctx.get('user')`) instead of the preferred Dependency Injection pattern.

Considering how little was done for dynamic values in the last three years, I agree it's a good idea to allow them to access the context (and other metadata) directly :+1:

At the same time, I would like to keep the doors open to allow us to implement dependency injection for dynamic values too, e.g.

```ts
ctx.bind('msg').toDynamicValue(SecurityBindings.USER, (ctx, user) => {
  return `Hello ${user.name}`;
});

// Note that `user` is strongly typed as `UserProfile`,
// using the type info provided by USER binding key
```

I feel the current design, where you are adding three arguments (`ctx`, `binding`, `options`), could make it difficult to come up with an ergonomic API for injecting values to dynamic value factories in the future.

I am proposing to convert these arguments into a single object argument.

```ts
const factory: ValueFactory<string> = ({context, binding, options}) => {
   return `Hello, ${context.name}#${
     binding.key
   } ${options.session?.getBindingPath()}`;
 };
```

I am surprised we don't have a type to hold all this fields yet, e.g. `ResolutionContext`, conceptually similar to `InvocationContext` and `RequestContext`.

I also noticed that `ResolverFunction` has similar but slightly different signature:

https://github.com/strongloop/loopback-next/blob/fcace9ea07c2d218835a55f3bc2c249ce0963620/packages/context/src/inject.ts#L46-L51

The good part is that the session can be accessed from dynamic value factories via `ResolutionOptions.session` :+1: 

The bad part is that we have different flavors of resolvers:
- Dynamic value factory receives full resolution options
- Resolvers receive the current session only

Have you considered any alternative designs that would be more consistent? Maybe this inconsistency is just a price we have to pay for evolving design and we should put a TODO item to modify `ResolveFactory` to accept full resolution options when we are working on the next major release of `@loopback/context`?

As I am thinking about this some more, I guess it would be best to:
- Keep `ResolverFunction` as is to preserve backwards compatibility.
- Implement `ValueFactory` using the design we want to use going forward (i.e. with access to full options).
  - Personally, I strongly prefer `ResolutionContext` over positional arguments. Besides the arguments posted above, this approach is also more future proof.
- Add `TODO(semver-major)` to update `ResolverFunction` to the new style in the future.

bajtos:
> In the past, I was against allowing dynamic value factories to access the context object, because it makes it too easy to slip into using Service Locator design pattern (`ctx.get('user')`) instead of the preferred Dependency Injection pattern.

Let me add more context to explain the ideal I'd like us to eventually reach in the future.

Let's take a slightly modified example from the PR description:

```js
ctx.bind('user').to('John');
ctx.bind('greeting').toDynamicValue(getUserGreeting);

function getUserGreeting(_ctx) {
  const user = _ctx.getSync('user');
  return `Hello, ${user}`;
}
```

Using Service Locator has several flaws:

1. From outside, it's difficult to say what are the dependencies of the function `getUserGreeting`. All we see is a function accepting a Context instance.

2. It's difficult to test `getUserGreeting` in isolation. We have to create a full context object to be able to call the function.

    ```js
    describe('getUserGreeting', () => {
      it('greets user with name', () => {
        const ctx = new Context();
        // Code smell: magic constant 'user'! 
        // Where is the binding key coming from?
        ctx.bind('user').to('John');
      
        const greeting = getUserGreeting(ctx);
        expect(greeting).to.equal('Hello, John');
      });
    });
    ```

3. When a new (required) dependency is added to `getUserGreeting`, the compiler cannot point us to places that need to be updated, we will discover the missing dependency only at runtime.

Let's see how Dependency Injection makes everything better:

1. The function signature clearly describes the dependencies, the implementation is much simpler because it does not have to resolve dependencies.

    ```js
    function getUserGreeting(user) {
      return `Hello, ${user}`;
    }

    ctx.bind('user').to('John');
    ctx.bind('greeting').toDynamicValue<string>('user', getUserGreeting);
    ```

2. The function is easy to test in isolation, no context instance is needed.

    ```js
    describe('getUserGreeting', () => {
      it('greets user with name', () => {
        const greeting = getUserGreeting('John');
        expect(greeting).to.equal('Hello, John');
      });
    });
    ```

3. When a new (required) dependency is added to `getUserGreeting`, the compiler will print errors for all places calling `getUserGreeting` with insufficient number of arguments.

Here is a mock-up of a possible implementation that I was playing with few years ago:

```ts
export class Binding<T = BoundValue> extends EventEmitter {
  // ...

   // no dependencies
  toDynamicValue(factoryFn: () => ValueOrPromise<T>): this;
  // 1 dependency
  toDynamicValue<D1>(key1: BindingKey<D1>, factoryFn: (d1: D1) => ValueOrPromise<T>): this;
  // 2 dependencies
  toDynamicValue<D1, D2>(key1: BindingKey<D1>, key2: BindingKey<D2>, factoryFn: (d1: D1, d2: D2) => ValueOrPromise<T>): this;
  // rinse & repeat for 3,4,5,... dependencies

  // the implementation
  toDynamicValue(...args: unknown) {
    const factoryFn = args.pop() as (...deps: unknown[]) => ValueOrPromise<T>;
    const depKeys = args as BindingKey<unknown>[];
    this._setValueGetter(ctx => {
      const depValues = resolveList(depKeys, k => ctx.getPromiseOrValue(k));
      return transformValueOrPromise(depValues, vals => factoryFn(values));
    });
  }
}
```

Now as I wrote before, I am not against the Service Locator approach proposed in this pull request, I am fine to implement it as a short-to-medium-term solution for dynamic value factories. I am just asking to design the APIs in such way that will make it feasible to add Dependency Injection to dynamic value factories later in the future, in a way that won't be disruptive to existing applications using dynamic values based on the Service Locator pattern.

raymondfeng:
@bajtos I think I have a better/lighter DI solution without reinventing the wheels.

```ts
class StaticGreetingProvider {
  static value(@inject('user') user: string) {
    return `Hello, ${user}`;
  }
}

ctx.bind('greeting').toDynamicValue(StaticGreetingProvider);
```

The idea is to have a class with `static` `value` method that supports method parameter injection. The latest PR has implemented the idea.

benchmark/src/context-binding/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+2. Use dynamic value
+
+```ts
+/**
+ * Second option - use a factory function
+ */
+const factory: ValueFactory = _ctx => {
+  const user = _ctx.getSync('user');
```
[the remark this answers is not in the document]
raymondfeng: Added async to benchmark tests.

benchmark/tsconfig.json
```diff
@@ -13,6 +13,9 @@
     {
       "path": "../packages/testlab/tsconfig.json"
     },
+    {
+      "path": "../packages/context/tsconfig.json"
+    },
```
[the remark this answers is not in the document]
raymondfeng: See https://github.com/strongloop/loopback-next/pull/5393. There was a bug in sorting ts refs.

docs/site/Binding.md
raymondfeng: I added some comments to `ResolverFn` and defer the refactoring to the future.

benchmark/src/context-binding/context-binding.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * First option - use a factory function
+ */
+const factory: ValueFactory = ({context}) => {
+  const user = context.getSync('user');
+  return `Hello, ${user}`;
+};
+
+/**
+ * First option - use a factory function
```
bajtos, line 18: Typo?

```suggestion
 * Second option - use an async factory function
```

benchmark/src/context-binding/context-binding.ts
```diff
@@ -0,0 +1,108 @@
...
+
+function setupContextBindings() {
+  const ctx = new Context();
+  ctx.bind('user').to('John');
+  ctx.bind('greeting1').toDynamicValue(factory);
+  ctx.bind('greeting2').toDynamicValue(asyncFactory);
+  ctx.bind('greeting3').toDynamicValue(StaticGreetingProvider);
+  ctx.bind('greeting4').toDynamicValue(AsyncStaticGreetingProvider);
+  ctx.bind('greeting5').toProvider(GreetingProvider);
+  ctx.bind('greeting6').toProvider(AsyncGreetingProvider);
```
bajtos, line 72: Please use more descriptive binding names than `greeting1` to `greeting6`. For example:

```suggestion
  ctx.bind('greeting.AsyncGreetingProvider').toProvider(AsyncGreetingProvider);
```

benchmark/src/context-binding/context-binding.ts
```diff
@@ -0,0 +1,108 @@
...
+  return ctx;
+}
+
+function runBenchmark(ctx: Context) {
+  const options: Benchmark.Options = {
+    initCount: 100,
```
bajtos, line 78: Let's use a higher number to ensure V8 has a chance to optimize functions on the hot path. AFAICT, the slowest case is `326,206 ops/sec`, I think we can afford to run 10k or even 100k runs to warm up.

benchmark/src/context-binding/README.md
```diff
@@ -0,0 +1,30 @@
...
+factoryFunction - getSync x 1,355,496 ops/sec ±0.86% (89 runs sampled) 71729
+factoryFunction x 1,282,558 ops/sec ±1.22% (90 runs sampled) 69158
+asyncFactoryFunction x 386,701 ops/sec ±1.69% (88 runs sampled) 21441
+factoryProvider - getSync x 629,167 ops/sec ±0.91% (93 runs sampled) 33437
+factoryProvider x 600,332 ops/sec ±0.86% (89 runs sampled) 32014
+asyncFactoryProvider x 348,149 ops/sec ±1.50% (80 runs sampled) 21240
+provider - getSync x 433,978 ops/sec ±1.03% (87 runs sampled) 24909
+provider x 399,416 ops/sec ±1.00% (91 runs sampled) 21647
+asyncProvider x 326,206 ops/sec ±1.41% (86 runs sampled) 18164
+```
```
bajtos, line 30: This is great! I wish these numbers were formatted as a Markdown table, but that's a small detail we can improve in the future.

benchmark/src/context-binding/context-binding.ts
```diff
@@ -0,0 +1,108 @@
...
+    return Promise.resolve(`Hello, ${user}`);
+  }
+}
+
+/**
+ * Fourth option - use a regular provider class
+ */
+class GreetingProvider implements Provider<string> {
+  @inject('user')
+  private user: string;
```
bajtos, line 46: Is there any difference in performance when using property injection vs. constructor argument injection?

benchmark/src/context-binding/context-binding.ts
```diff
@@ -0,0 +1,108 @@
...
+    .add('provider - getSync', () => ctx.getSync('greeting5'), options)
+    .add('provider', () => ctx.get('greeting5'), options)
+    .add('asyncProvider', () => ctx.get('greeting6'), options)
+    .run({async: true});
+}
+
+if (require.main === module) {
+  let tests = process.argv.slice(2);
+  if (!tests.length) {
+    tests = ['1000'];
```
bajtos, line 101: How is the number `1000` used? AFAICT, it's completely ignored now.

Did you perhaps mean to use it as the number of iterations to pass to the benchmarking tool? If we make the number of iterations configurable, then we also need to adjust `initCount` accordingly.

As far as I am concerned, there is no need to make the number of iterations configurable in this benchmark implementation.

benchmark/tsconfig.json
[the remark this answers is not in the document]
bajtos: Thanks for fixing the package sorter. I think you need to rebase this PR on top of the latest master and re-run the code to update project references?

docs/site/Binding.md
```diff
@@ -83,6 +83,36 @@ binding.toDynamicValue(() => new Date());
...
 ```
 
+The factory function can receive extra information about the context, binding,
+and resolution options.
+
+```ts
+import {ValueFactory} from '@loopback/context';
+
+// The factory function now have access extra metadata about the resolution
+const factory: ValueFactory<string> = resolutionCtx => {
```
bajtos, line 93: Is there any particular reason against using object destructuring here? I think it's a pretty common pattern that most JS/TS developers should be familiar with.

```suggestion
const factory: ValueFactory<string> = ({context, binding, options}) => {
```

packages/context/src/__tests__/unit/binding.unit.ts
```diff
@@ -172,6 +173,44 @@ describe('Binding', () => {
...
+          _binding.key
+        } ${options.session?.getBindingPath()}`;
+      };
+      const b = ctx.bind('msg').toDynamicValue(factory);
+      const value = await ctx.get<string>('msg');
+      expect(value).to.equal('Hello, test#msg msg');
+      expect(b.type).to.equal(BindingType.DYNAMIC_VALUE);
+    });
+
+    it('supports a factory to use context', () => {
```
bajtos, line 192: This test seems like it's pretty much the same as `it('support a factory to access context/binding/session')` above. Please rework it to make it more clear what are the important differences from the previous one, or remove it completely if it's not testing anything new.

packages/context/src/binding.ts
```diff
@@ -170,11 +172,63 @@ export type BindingEventListener = (
...
+    invokeMethod(provider, 'value', resolutionCtx.context, [], {
+      skipInterceptors: true,
+    });
+}
+
+/**
+ * Check if the factory is a value factory provider class
+ * @param factory - A factory function or a dynamic value provider class
+ */
+function DynamicValueProviderClass(
```
bajtos, line 220: Functions usually start with a verb and have `camelCase` names.

```suggestion
function isDynamicValueProviderClass(
```

packages/context/src/binding.ts
```diff
@@ -170,11 +172,63 @@ export type BindingEventListener = (
...
+  factory: unknown,
+): factory is DynamicValueProviderClass {
+  // Not a class
+  if (typeof factory !== 'function' || !String(factory).startsWith('class ')) {
+    return false;
+  }
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const valueMethod = (factory as any).value;
```
bajtos, line 228: Is it possible to avoid `any` by casting `factory` to `DynamicValueProviderClass`? Something along the following lines:

```ts
const valueMethod = ((factory as unknown) as DynamicValueProviderClass).value;
```

packages/context/src/resolution-session.ts
```diff
@@ -366,3 +367,12 @@ export function asResolutionOptions(
...
   return optionsOrSession ?? {};
 }
+
+/**
+ * Context for resolution
+ */
+export interface ResolutionContext<T = unknown> {
+  context: Context;
+  binding: Readonly<Binding<T>>;
+  options: ResolutionOptions;
```
bajtos, line 377: Should we mark all fields as `readonly` to prevent accidental modification?

```ts
export interface ResolutionContext<T = unknown> {
  readonly context: Context;
  readonly binding: Readonly<Binding<T>>;
  readonly options: Readonly<ResolutionOptions>;
}
```

docs/site/Binding.md
raymondfeng: IMO, it's better to show the original parameter type for documentation purpose.

raymondfeng: I added another example with object destructuring.

benchmark/src/context-binding/README.md
raymondfeng: formatted.

benchmark/src/context-binding/context-binding.ts
raymondfeng: I increased it to 1000.

benchmark/src/context-binding/context-binding.ts
raymondfeng: It should be the same. We can add more benchmark tests in the future.

benchmark/src/context-binding/context-binding.ts
raymondfeng: Fixed.

raymondfeng:

PR #5391, opened 2020-05-11
fix: update default sequence and docs

raymondfeng:
## Checklist

### DIFF
--- a/docs/site/Sequence.md
+++ b/docs/site/Sequence.md
@@ -426,6 +426,11 @@ actions.
 
 ## Working with Express middleware
 
+{% include warning.html content="First-class support for Express middleware has
+been added to LoopBack since v4.0.0 of `@loopback/rest`. Please refer to
+[Using Express Middleware](Express-middleware.md). The following information
+only applies to earlier versions of `@loopback/rest`." %}
+
 Under the hood, LoopBack leverages [Express](https://expressjs.com) framework
 and its concept of middleware. To avoid common pitfalls, it is not possible to
 mount Express middleware directly on a LoopBack application. Instead, LoopBack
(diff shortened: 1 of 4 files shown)


docs/site/Sequence.md
```diff
@@ -15,21 +15,72 @@ The contract of a `Sequence` is simple: it must produce a response to a request.
...
+      if (finished) {
+        // The response been produced by the middleware chain
+        return;
+      }
+      const route = this.findRoute(request);
+      const args = await this.parseParams(request, route);
+      const result = await this.invoke(route, args);
+
+      debug('%s result -', route.describe(), result);
+      this.send(response, result);
```
bajtos, line 81: What happens when `this.send` returns a promise that will be eventually rejected? IMO, we should either keep `await` or return the send result directly.

```suggestion
      return this.send(response, result);
```

docs/site/Sequence.md
```diff
@@ -15,21 +15,72 @@ The contract of a `Sequence` is simple: it must produce a response to a request.
...
     } catch (error) {
-      await this.reject(context, error);
+      this.reject(context, error);
```
bajtos, line 83: Ditto.

```suggestion
      return this.reject(context, error);
```

docs/site/Sequence.md
```diff
@@ -71,6 +129,13 @@ class MySequence extends DefaultSequence {
 }
```
bajtos, line 129: Two lines above, you are calling `await this.send(...)`. Please make sure we are consistently showing the same approach in all places in our docs.

docs/site/Sequence.md
raymondfeng: The signature of `Send` returns `void`:

```ts
export type Send = (response: Response, result: OperationRetval) => void;
```

docs/site/Sequence.md
raymondfeng: `reject` returns `void` too.

```ts
export type Reject = (handlerContext: HandlerContext, err: Error) => void;
```

docs/site/Sequence.md
bajtos: Interesting, I was remembering that both `Send` and `Reject` were returning a promise.

This change is fine then, just make sure to update other places for consistency please.

Issue #4846, opened 2020-03-10
[Docs] Add "error handling" page


dhmlau:
Originated from: https://github.com/strongloop/loopback-next/issues/4844

In LB3 docs, there's Running and debugging apps page. A similar version for LB4 would be useful.

### Acceptance Criteria
- [ ] Review and clean up any error handling related content we have in LB4 docs
- [ ] Document how to configure http response for errors and how to report error from controller. 
    -  Use [strong-error-handler page](https://loopback.io/doc/en/lb3/Using-strong-error-handler.html) as a reference
    - the initial idea is to create a child page under `Running and debugging apps` page in LB4.

Related: https://github.com/strongloop/loopback-next/issues/4845

agnes512:
Closing as done

PR #5392, opened 2020-05-12
docs: add using strong-error-handler page

agnes512:
Implements https://github.com/strongloop/loopback-next/issues/4846

截屏2020-05-13上午11 51 17

截屏2020-05-13上午11 51 28

截屏2020-05-13上午11 51 45

截屏2020-05-13上午11 52 01

截屏2020-05-13上午11 52 11

## Checklist

### DIFF
--- a/docs/site/Using-strong-error-handler.md
+++ b/docs/site/Using-strong-error-handler.md
@@ -1,0 +1,147 @@
+---
+title: 'Using strong-error-handler'
+lang: en
+keywords: LoopBack 4, error handling
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Using-strong-error-handler.html
+---
+
+# strong-error-handler
+
+As a dependency of
+[`@loopback/rest`](https://github.com/strongloop/loopback-next/tree/master/packages/rest),
+package `strong-error-handler` is an error handler for use in both development
+(debug) and production environments.
+
+In LoopBack 4, the request handling process starts with the app's
+[sequence](https://loopback.io/doc/en/lb4/Sequence.html), a simple class with
+five injected helper methods in the constructor. It is the gatekeeper of all
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package to send back an HTTP response describing the error.
+
+LoopBack 4 apps require 3.x versions of `strong-error-handler`.
+
+## Usage
+
+In production mode (default), `reject` omits details from error responses to
+prevent leaking sensitive information.
+
+In debug mode, `reject` also can return full error stack traces and internal
+details of any error objects to the client in the HTTP responses. Once the
+`debug` flag is set, `strong-error-handler` would handle such options.
+
+```ts
+app.bind(RestBindings.ERROR_WRITER_OPTIONS).to({debug: true});
+```
+
+Please check [Sequence - Handling errors](Sequence.md#handling-errors) section
+for usages and examples of these two modes.
+
+In general, `reject` is executed in `try {} catch(err) {}` block.
+`strong-error-handler` is the last utility to be used.
+
+The above configuration will log errors to the server console, but not return
+stack traces in HTTP responses. The behaviors of the error handler are also
+configurable by setting other options to `ERROR_WRITER_OPTIONS`. For details on
+available options, see below.
+
+## Options
+
+| Option               | Type                      | Default  | Description                                                                                                                                                                                                                                                                                                                                               |
+| -------------------- | ------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| debug                | Boolean&nbsp;&nbsp;&nbsp; | `false`  | If `true`, HTTP responses include all error properties, including sensitive data such as file paths, URLs and stack traces. See [Example output](#example) below.                                                                                                                                                                                         |
+| log                  | Boolean                   | `true`   | If `true`, all errors are printed via `console.error`, including an array of fields (custom error properties) that are safe to include in response messages (both 4xx and 5xx). <br/> If `false`, sends only the error back in the response.                                                                                                              |
+| safeFields           | [String]                  | `[]`     | Specifies property names on errors that are allowed to be passed through in 4xx and 5xx responses. See [Safe error fields](#safe-error-fields) below.                                                                                                                                                                                                     |
+| defaultType          | String                    | `"json"` | Specify the default response content type to use when the client does not provide any Accepts header.                                                                                                                                                                                                                                                     |
+| negotiateContentType | Boolean                   | true     | Negotiate the response content type via Accepts request header. When disabled, strong-error-handler will always use the default content type when producing responses. Disabling content type negotiation is useful if you want to see JSON-formatted error responses in browsers, because browsers usually prefer HTML and XML over other content types. |
+
+### Safe error fields
+
+By default, `strong-error-handler` will only pass through the `name`, `message`
+and `details` properties of an error. Additional error properties may be allowed
+through on 4xx and 5xx status code errors using the `safeFields` option to pass
+in an array of safe field names:
+
+```ts
+app.bind(RestBindings.ERROR_WRITER_OPTIONS).to({safeFields: ['errorCode']});
+```
+
+Using the above configuration, an error containing an `errorCode` property will
+produce the following response:
+
+```
+{
+  "error": {
+    "statusCode": 500,
+    "message": "Internal Server Error",
+    "errorCode": "INTERNAL_SERVER_ERROR"
+  }
+}
+```
+
+## Response format and content type
+
+The `strong-error-handler` package supports JSON, HTML and XML responses:
+
+- When the object is a standard Error object, it returns the string provided by
+  the stack property in HTML/text responses.
+- When the object is a non-Error object, it returns the result of `util.inspect`
+  in HTML/text responses.
+- For JSON responses, the result is an object with all enumerable properties
+  from the object in the response.
+
+The content type of the response depends on the request's `Accepts` header.
+
+- For Accepts header `json` or `application/json`, the response content type is
+  JSON.
+- For Accepts header `html` or `text/html`, the response content type is HTML.
+- For Accepts header `xml` or `text/xml`, the response content type is XML.
+
+_There are plans to support other formats such as `text/plain`._
+
+## Migration from LoopBack 3 error handler
+
+Unlike LoopBack 3’s phase-based middleware routing system, LB4 uses a sequence
+handler that sits in front of a routing table and handles errors. To learn the
+differences and how `strong-error-handler` is being used in LB4, see
+[LB3 vs LB4 request response cycle](LB3-vs-LB4-request-response-cycle.md).
+
+## Example
+
+5xx error generated when `debug: false` :
+
+```
+{ error: { statusCode: 500, message: 'Internal Server Error' } }
+```
+
+The same error generated when `debug: true` :
+
+```
+{ error:
+  { statusCode: 500,
+  name: 'Error',
+  message: 'a test error message',
+  stack: 'Error: a test error message
+  at Context.<anonymous> (User/strong-error-handler/test/handler.test.js:220:21)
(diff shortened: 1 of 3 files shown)


docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,343 @@
...
+# strong-error-handler
+
+This package is an error handler for use in both development (debug) and
+production environments.
+
+## Supported versions
+
+| Current | Long Term Support | End-of-Life |
+| :-----: | :---------------: | :---------: |
+|   4.x   |        3.x        |     2.x     |
```
bajtos, line 28: We usually format the LTS table differently: each major version is in a new row and we also show the EOL date for each version, see e.g. here:

https://github.com/strongloop/loopback#module-long-term-support-policy

I think there is no need to show 2.x version. The current 3.x version was published at [2018-06-11](https://github.com/strongloop/strong-error-handler/blob/master/CHANGES.md#2018-06-11-version-300), before LB4 GA. I would be very surprised if there was any LB4 user using 2.x version of strong-error handler.

Last but not least, AFAICT, the latest strong-error-handler version is 3.x. I am not sure where is 4.x coming from? See https://github.com/strongloop/strong-error-handler#supported-versions and https://github.com/strongloop/strong-error-handler/blob/master/package.json#L5

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,343 @@
...
+
+```bash
+$ npm install --save strong-error-handler
+```
+
+## Use
+
+In production mode, the default implementation of `reject`,
+`strong-error-handler` omits details from error responses to prevent leaking
+sensitive information:
```
bajtos, line 43: This section is duplicating existing content in https://loopback.io/doc/en/lb4/Sequence.html#handling-errors. Let's find a way how to avoid that duplication please. For example, the duplicate content on one page can be replaced with a link (a reference) to the other page.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,251 @@
...
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package.
+
+# strong-error-handler
+
+This package is an error handler for use in both development (debug) and
+production environments.
+
+## Supported versions
```
dhmlau, line 24: just a suggestion.. it seems a bit strange to have this supported version table where there's only one entry. If only 3.x works with LB4, perhaps we can mention in the paragraph above that LB4 requires `strong-error-handler` 3.x and up? (though there's no "up" yet).

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,251 @@
...
+        res,
+        errHandlingOptions,
+      );
+    }
+  })
+  .listen(3000);
+```
+
+In general, `strong-error-handler` must be the last middleware function
+registered.
```
dhmlau, line 72: maybe this is more grammatically correct? (you may want to double check) 
```suggestion
to be registered.
```

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,251 @@
...
+  from the object in the response.
+
+The content type of the response depends on the request's `Accepts` header.
+
+- For Accepts header `json` or `application/json`, the response content type is
+  JSON.
+- For Accepts header `html` or `text/html`, the response content type is HTML.
+- For Accepts header `xml` or `text/xml`, the response content type is XML.
+
+_There are plans to support other formats such as Plain-text._
```
dhmlau, line 108: should it be `text/plain`? 
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,251 @@
...
+    "message": "Internal Server Error",
+    "errorCode": "INTERNAL_SERVER_ERROR"
+  }
+}
+```
+
+## Migration from LoopBack 3 error handler
+
+Unlike LoopBack 3’s phase-based middleware routing system, LB4 uses a sequence
+handler that sits infront of a routing table and handles errors. To learn the
```
dhmlau, line 210:
```suggestion
handler that sits in front of a routing table and handles errors. To learn the
```

docs/site/Using-strong-error-handler.md
agnes512: Sounds good! I agree that it's weird to have a table that only have one item :sweat_smile: Will reword it.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,247 @@
...
+  from the object in the response.
+
+The content type of the response depends on the request's `Accepts` header.
+
+- For Accepts header `json` or `application/json`, the response content type is
+  JSON.
+- For Accepts header `html` or `text/html`, the response content type is HTML.
+- For Accepts header `xml` or `text/xml`, the response content type is XML.
+
+_There are plans to support other formats such as text/plain._
```
dhmlau, line 104: nitpick
```suggestion
_There are plans to support other formats such as `text/plain`._
```

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,244 @@
...
+
+This package is an error handler for use in both development (debug) and
+production environments.
+
+In LoopBack 4, the request handling process starts with the app's
+[sequence](https://loopback.io/doc/en/lb4/Sequence.html), a simple class with
+five injected helper methods in the constructor. It is the gatekeeper of all
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package.
```
jannyHou, line 19:
```suggestion
package to send back an HTTP response describing the error.
```
It'd be better to explain calls the module for what purpose.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,244 @@
...
+
+Under the hood, LoopBack leverages [Express](https://expressjs.com) framework
+and its concept of middleware. To avoid common pitfalls, it is not possible to
+mount Express middleware directly on a LoopBack application.
+
+`strong-error-handler` can be bond to the app, which is the same as above. For
+example:
+
+```ts
+app.bind(RestBindings.ERROR_WRITER_OPTIONS).to({debug: true});
```
jannyHou, line 80: > `strong-error-handler` can be bond to the app, which is the same as above

The binding here only adds the configuration, could you add the code showing how to bind the handler itself?

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,244 @@
...
+
+The module also exports `writeErrorToResponse`, a non-middleware flavor of the
+error handler:
+
+```js
+import {Server} from 'http';
+import {writeErrorToResponse} from 'strong-error-handler';
+const errHandlingOptions = {debug: true, log: true};
+
+http
```
jannyHou, line 50: could you elaborate more about how to use/combine this in a lB4 application?

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,244 @@
...
+## Installation
+
+```bash
+$ npm install --save strong-error-handler
+```
+
+## Usage
+
+As mentioned above, errors are handled by the `reject` action of the app's
+[sequence](https://loopback.io/doc/en/lb4/Sequence.html). By default,
```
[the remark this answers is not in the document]
agnes512: I'd like to explain that in LB4, errors are handled by `Sequence.reject` first, then explain the default mode and the debug mode. Do you have any suggestions of rewording it?

docs/site/Using-strong-error-handler.md
agnes512: I decide to remove this part as I think it's the same as the [Customizing log message] section below.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,234 @@
...
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package to send back an HTTP response describing the error.
+
+LB4 apps require 3.x versions of `strong-error-handler`.
+
+## Installation
+
+```bash
+$ npm install --save strong-error-handler
```
raymondfeng, line 26: For LoopBack's `@loopback/rest` package, `strong-error-handler` is already a dependency. I don't think an explicit installation is needed.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,234 @@
...
+responses to prevent leaking sensitive information.
+
+`strong-error-handler` also can return full error stack traces and internal
+details of any error objects to the client in the HTTP responses once the
+`debug` flag is set. Please check
+[Sequence - Handling errors](Sequence.md#handling-errors) section for usages and
+examples.
+
+In general, `strong-error-handler` must be the last middleware function to be
+registered.
```
raymondfeng, line 43: For the sequence we have, the `reject` is executed in `try {} catch(err) {}` block. `strong-error-handler` itself is a utility, not a middleware.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,234 @@
...
+
+The above configuration will log errors to the server console, but not return
+stack traces in HTTP responses. For details on configuration options, see below.
+
+### Working with Express middleware
+
+Under the hood, LoopBack leverages [Express](https://expressjs.com) framework
+and its concept of middleware. To avoid common pitfalls, it is not possible to
+mount Express middleware directly on a LoopBack application.
```
raymondfeng, line 52: This has changed since the introduction of Express middleware support in `@loopback/rest@4.0.0`. But the key is that the error handler is stilled used.

docs/site/Using-strong-error-handler.md
bajtos: +1

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,146 @@
...
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package to send back an HTTP response describing the error.
+
+LB4 apps require 3.x versions of `strong-error-handler`.
+
+## Usage
+
+As mentioned above, errors are handled by the `reject` action of the app's
+[sequence](https://loopback.io/doc/en/lb4/Sequence.html). By default, `reject`
```
jannyHou, line 28: Nitpick: a suggestion for rephrasing the usage example :)

---
You can configure the error handler's behaviors by setting the error writer options.
For example, by default, `reject` omits details from error responses to prevent leaking sensitive information. But `reject` also can return full error stack traces and internal details of any error objects to the client in the HTTP responses. This is controlled by the option property `debug`:

```ts
app.bind(RestBindings.ERROR_WRITER_OPTIONS).to({debug: true});
```

Once the `debug` flag is set, `strong-error-handler` would handle such options.

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,147 @@
...
+title: 'Using strong-error-handler'
+lang: en
+keywords: LoopBack 4, error handling
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Using-strong-error-handler.html
+---
+
+# strong-error-handler
+
+As the part of the dependencies of
```
raymondfeng, line 11: `As a dependency of`?

docs/site/Using-strong-error-handler.md
```diff
@@ -0,0 +1,147 @@
...
+(debug) and production environments.
+
+In LoopBack 4, the request handling process starts with the app's
+[sequence](https://loopback.io/doc/en/lb4/Sequence.html), a simple class with
+five injected helper methods in the constructor. It is the gatekeeper of all
+requests to the app. Errors are handled by one of the Sequence actions,
+[`reject`](Sequence.md#handling-errors), which calls `strong-error-handler`
+package to send back an HTTP response describing the error.
+
+LB4 apps require 3.x versions of `strong-error-handler`.
```
raymondfeng, line 23: `LB4` -> `LoopBack 4`.

PR #5394, opened 2020-05-12
refactor: enable type checks inside mixin classes

bajtos:
Introduce a new type helper `MixinTarget` allowing mixin functions to accept a type that describes public members of the target class only. This is working around the current TypeScript limitations:
- https://github.com/microsoft/TypeScript/issues/17293
- https://github.com/microsoft/TypeScript/issues/17744
- https://github.com/microsoft/TypeScript/issues/36060

Rework all existing mixins and the related documentation to use `MixinTarget<RealClass>` instead of `Constructor<any>`.

Fix any errors discovered by the compiler after enabling type checks.

See also https://github.com/microsoft/TypeScript/issues/38496

~~This PR depends on https://github.com/strongloop/loopback-datasource-juggler/pull/1839, the build will keep failing until a new version of juggler is released & loopback-next package locks are updated.~~

## Checklist

### DIFF
--- a/docs/site/migration/models/mixins.md
+++ b/docs/site/migration/models/mixins.md
@@ -16,7 +16,7 @@ This document will guide you in migrating custom model mixins, and custom
 method/remote method mixins in LoopBack 3 to their equivalent implementations in
 LoopBack 4.
 
-For an understanding of how models in LoopBack 3 are now architectually
+For an understanding of how models in LoopBack 3 are now architecturally
 decoupled into 3 classes (model, repository, and controller) please read
 [Migrating custom model methods](./methods.md).
 
@@ -169,7 +169,7 @@ import {property, Model} from '@loopback/repository';
  * @param superClass - Base Class
  * @typeParam T - Model class
  */
-export function AddCategoryPropertyMixin<T extends Constructor<Model>>(
+export function AddCategoryPropertyMixin<T extends MixinTarget<Model>>(
   superClass: T,
 ) {
   class MixedModel extends superClass {
@@ -447,7 +447,7 @@ import {FindByTitle} from './find-by-title-interface';
  */
 export function FindByTitleRepositoryMixin<
   M extends Model & {title: string},
-  R extends Constructor<CrudRepository<M>>
+  R extends MixinTarget<CrudRepository<M>>
 >(superClass: R) {
   class MixedRepository extends superClass implements FindByTitle<M> {
     async findByTitle(title: string): Promise<M[]> {
@@ -552,10 +552,12 @@ export interface FindByTitleControllerMixinOptions {
  */
 export function FindByTitleControllerMixin<
   M extends Model,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  T extends Constructor<any> = Constructor<object>
+  T extends MixinTarget<object>
 >(superClass: T, options: FindByTitleControllerMixinOptions) {
   class MixedController extends superClass implements FindByTitle<M> {
+    // Value will be provided by the subclassed controller class
+    repository: FindByTitle<M>;
+
     @get(`${options.basePath}/findByTitle/{title}`, {
       responses: {
         '200': {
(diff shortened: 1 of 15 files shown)


packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -57,26 +48,42 @@ export function BootMixin<T extends Constructor<any>>(superClass: T) {
         () => this.projectRoot,
       );
       this.bind(BootBindings.BOOT_OPTIONS).toDynamicValue(
-        () => this.bootOptions,
+        () => this.bootOptions ?? {},
```
bajtos, line 51: This problem was hidden by `any` before.

packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -115,9 +122,14 @@ export function BootMixin<T extends Constructor<any>>(superClass: T) {
...
      */
+    // Unfortunately, TypeScript does not allow overriding methods inherited
+    // from mapped types. https://github.com/microsoft/TypeScript/issues/38496
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
     public component(component: Constructor<{}>) {
-      super.component(component);
+      const binding = super.component(component);
       this.mountComponentBooters(component);
+      return binding;
```
bajtos, line 132: This problem was hidden by `any` before.

packages/repository/src/mixins/repository.mixin.ts
```diff
@@ -161,12 +163,17 @@ export function RepositoryMixin<T extends Class<any>>(superClass: T) {
...
+    // @ts-ignore
+    public component<C extends Component = Component>(
+      componentCtor: Constructor<C>,
       nameOrOptions?: string | BindingFromClassOptions,
     ) {
-      super.component(component, nameOrOptions);
-      this.mountComponentRepositories(component);
+      const binding = super.component(componentCtor, nameOrOptions);
+      this.mountComponentRepositories(componentCtor);
+      return binding;
```
bajtos, line 176: This problem was hidden by `any` before.

docs/site/migration/models/mixins.md
```diff
@@ -552,10 +552,12 @@ export interface FindByTitleControllerMixinOptions {
  */
 export function FindByTitleControllerMixin<
   M extends Model,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  T extends Constructor<any> = Constructor<object>
+  T extends MixinTarget<object>
```
raymondfeng, line 555: Should we use `object` consistently for all `MixinTarget<Object>`?

packages/core/src/core.types.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright IBM Corp. 2017,2020. All Rights Reserved.
```
raymondfeng, line 1: I'm not sure if I like the name `core.types.ts`. Adding `core.` prefix seems to be redundant as the file is already in `core` package. Otherwise, we need to add `core.` to all files?

packages/core/src/core.types.ts
```diff
@@ -0,0 +1,70 @@
...
+ *    }
+ * }
+ * ```
+ */
+export type MixinTarget<T extends object> = Constructor<
+  {
+    // Enumerate only public members to avoid the following compiler error:
+    //   Property '(name)' of exported class expression
+    //   may not be private or protected.ts(4094)
+    [p in keyof T]: T[p];
```
raymondfeng, line 68: For types, TS use `P` instead of `p`.

docs/site/migration/models/mixins.md
bajtos: Good catch, I agree to use `MixinTarget<object>` everywhere :+1:

packages/core/src/core.types.ts
bajtos: Well, if you have a `types.ts` file in several packages, then it's difficult to tell which file you are editing. You have to check the full path to see. The package name is not even the last entry in the path (`packages/core/src`), which makes it even more difficult. That's why I prefer to add package name prefix to common files like `types`, `keys`, `component`, and so on.

Since the code comment is many lines long, maybe it will be best to move the code to a file called something like `packages/core/src/mixin-target.ts`, WDYT?

packages/core/src/core.types.ts
bajtos: Good catch, will fix.

packages/core/src/core.types.ts
raymondfeng: +1 to use `src/mixin-target.ts`.

Issue #4673, opened 2020-02-18
linter: package metadata and registration


bajtos:
When adding a new package, it's often easy to forget about the checklist we have in our docs ([Adding a new package](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#adding-a-new-package)).

Let's implement a linter that will verify that all packages in our monorepo are registered in all required places and have the expected metadata in their `package.json`.

## Acceptance criteria

- [ ] Add a new script to the top-level `bin` directory. This script should obtain a list of all packages known by lerna (see e.g. [here](https://github.com/strongloop/loopback-next/blob/7fe694865b337c99beb25afc581c5429fb758746/bin/sync-dev-deps.js#L19-L22)) and run the following checks for each package:
- [ ] The package has `LICENSE` file with the expected content (MIT license + npm package name)
- [ ] The package has an entry in [MONOREPO.md](https://github.com/strongloop/loopback-next/blob/master/docs/site/MONOREPO.md)
- [ ] The package has an entry in [CODEOWNERS](https://github.com/strongloop/loopback-next/blob/master/CODEOWNERS)
- [ ] Package's `package.json` file has `publishConfig` with `access` set to `public`, see [Make a scoped package public](https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#make-a-scoped-package-public)
- [ ] Package's `package.json` file has `author`, `copyright.owner` and `license` fields set to correct values (let's use `package.json` in monorepo root as the source of truth)
- [ ] Package's `package.json` file has correct `repository` configuration, pointing to loopback-next at GitHub and including `directory` field with a relative path inside our monorepo
- [ ] Modify Travis CI task [`code-lint`](https://github.com/strongloop/loopback-next/blob/7fe694865b337c99beb25afc581c5429fb758746/.travis.yml#L28-L36) to run the new script as part of our CI build.

spirAde:
@bajtos Hello, I started to implement this script. At the moment I'd finished license checker and found that docs folder(`loopback-next/docs`) doesn't contain `LICENSE` file although Lerna includes this folder as a package. Should I exclude the docs folder from checks or rather to add the `LICENSE` file in the docs folder?

spirAde:
[...]

PR #5446, opened 2020-05-15
chore: add a script to lint package metadata

spirAde:
This makes it possible to verify that all packages in the monorepo are registered in all required places and have the expected metadata in their package.json

Issue: #4673

## Checklist


bin/check-package-metadata.js
```diff
@@ -0,0 +1,176 @@
...
+ */
+'use strict';
+
+const path = require('path');
+const fs = require('fs-extra');
+
+const Project = require('@lerna/project');
+
+const MIT_LICENSE_TEXT = `
+MIT license
```
spirAde, line 20: On the one hand, I’m not sure about this type of check, on the other hand, the license file will be the same everywhere.

spirAde:
@bajtos You also mentioned that check with the team who owns the apiconnection package from the comment here: [https://github.com/strongloop/loopback-next/issues/4673#issuecomment-629278284](url). Without this, the script will not work, it will crash at the test `CODEOWNERS.md`

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/5449

raymondfeng:
@spirAde Thank you for helping out here. We use conventional commits. The commit message should be amended to be something like `chore: add a script to lint package metadata`. See https://github.com/strongloop/loopback-next/pull/5449.

bin/check-package-metadata.js
raymondfeng: For our monorepo, we run `lb4 copyright` command once a while to make sure the license headers are up to date.

spirAde:
> @spirAde Thank you for helping out here. We use conventional commits. The commit message should be amended to be something like `chore: add a script to lint package metadata`. See https://loopback.io/doc/en/lb4/submitting_a_pr.html#4-commit-changes-to-the-local-feature-branch.

Oh, sorry. I know about conventional commits. I thought I can use label pointer `feature` from here https://github.com/strongloop/loopback-next/issues/4673 as part of commit message. I'll fix it

bin/check-package-metadata.js
spirAde: Maybe I can take root folder LICENSE file as source of truth, but in this case loopback needs some way how to support this file in actual state? Or I can simplify check for license file

raymondfeng:
@spirAde Another logistic issue - please run `npm run prettier:fix` to fix formatting issue.

spirAde:
@raymondfeng fixed, but one test failed

bin/check-package-metadata.js
spirAde: Or another solution: take package `spdx-license-list` which is using in `packages/cli`. Anyway, guys, I need your decision what is the sufficient way from your perspective? @raymondfeng @bajtos

raymondfeng: Checking the existence of `LICENSE` file is probably good enough as `lb4 copyright` can deal with the consistency.

Issue #5426, opened 2020-05-14
How can extensions introspect application artifacts


bajtos:
This story is extracted from #4099, where we researched existing LoopBack 3 components and various techniques they use.

Now we need to write documentation for extension authors to show how to solve the following use cases in LoopBack 4:

1. Get a list of all models (entity classes) used by the application

2. Get definition of a given model (information about properties, relations, mixins, etc.)

3. Get a list of all remote methods (REST API endpoints) provided by a given model.

4. Obtain a list of all REST endpoints provided by the application, including metadata about request parameters & response schema, and any other information necessary to configure reverse-proxy routing rules (Kong, nginx,
    etc.).

In most cases, the new content will be useful to authors building new LB4 components too, therefore we should structure the content in two parts:
- A guide explaining how to build a component contributing a service, this should go to [Extending LoopBack](https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html)
- A guide explaining how to migrate LB3 components, this guide should go to [Migrating components and extensions](https://loopback.io/doc/en/lb4/migration-extensions-overview.html) and focus on aspects unique to extensions migrating from LB3 and refer to content in [Extending LoopBack](https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html) wherever possible & appropriate.

PR #5456, opened 2020-05-15
feat(binding): use `_source` to track the original value source for a binding

raymondfeng:
We track the source for a binding using multiple properties such as `valueConstructor`, `providerConstructor`, `_alias`. This PR converges it to a `_source` property and have getters to keep backward compatibility. 

This is loosely related to https://github.com/strongloop/loopback-next/issues/5426. To inspect bindings for certain artifacts, being able to access the `source` is useful for extensions to enhance the inspection results.

## Checklist

### DIFF
--- a/packages/context/src/binding.ts
+++ b/packages/context/src/binding.ts
@@ -131,6 +131,56 @@ export enum BindingType {
   ALIAS = 'Alias',
 }
 
+/**
+ * Binding source for `to`
+ */
+export type ConstantBindingSource<T> = {
+  type: BindingType.CONSTANT;
+  value: T;
+};
+
+/**
+ * Binding source for `toDynamicValue`
+ */
+export type DynamicValueBindingSource<T> = {
+  type: BindingType.DYNAMIC_VALUE;
+  value: ValueFactory<T> | DynamicValueProviderClass<T>;
+};
+
+/**
+ * Binding source for `toClass`
+ */
+export type ClassBindingSource<T> = {
+  type: BindingType.CLASS;
+  value: Constructor<T>;
+};
+
+/**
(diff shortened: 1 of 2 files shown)


packages/context/src/binding.ts
```diff
@@ -263,24 +263,31 @@ export class Binding<T = BoundValue> extends EventEmitter {
...
 
-  private _valueConstructor?: Constructor<T>;
-  private _providerConstructor?: Constructor<Provider<T>>;
-  private _alias?: BindingAddress<T>;
+  /**
+   * The original source value received from `to`, `toClass`, `toDynamicValue`,
+   * `toProvider`, or `toAlias`.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private _source: any;
```
bajtos, line 271: Can we please avoid using `any`? At minimum, let's try to use `unknown` instead.

Alternatively, can we use a union type here?

```suggestion
  private _source: 
    | T // .to()
    | Constructor<T> // .toClass()
    | Constructor<Provider<T>> // .toProvider()
    | ValueFactory<T> | DynamicValueProviderClass<T> // .toDynamicValue()
    | BindingAddress<T> // .toAlias()
    ;
```

bajtos: I think it would be best to create a new type `SourceType<T>` and provide proper TSDoc comment to explain all different kinds of values the source can hold.

raymondfeng: I added `BindingSource` type per https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions.

raymondfeng:
@bajtos PTAL.

PR #5460, opened 2020-05-16
chore: use eslint-plugin-prettier to utilise cache

dougal83:
Add eslint-plugin-prettier to enable linting to utilise cache feature of eslint.

See also [comment](https://github.com/strongloop/loopback-next/pull/4330#issuecomment-629225827)

Hopefully implemented correctly in all counts!  Let me know if this is useful.

Would it be fine to just run eslint after tests?

## Checklist

FAO @bajtos @raymondfeng


raymondfeng:
The change LGTM. But I'm curious how much it can buy us. In my experience, `npm run prettier:fix` is pretty fast. The main cost is to `npm run eslint` for the first time when the cache is not available.

raymondfeng:
@dougal83 I have rebased the branch to latest master.

dougal83:
> The change LGTM. But I'm curious how much it can buy us. In my experience, `npm run prettier:fix` is pretty fast. The main cost is to `npm run eslint` for the first time when the cache is not available.

@raymondfeng I'll happily run timing benchmarks out of curiosity.  I'll go look up commands tomorrow...  Thank you for rebase.

packages/eslint-config/package.json
```diff
@@ -14,10 +14,12 @@
...
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-eslint-plugin": "^2.2.1",
-    "eslint-plugin-mocha": "^7.0.0"
+    "eslint-plugin-mocha": "^7.0.0",
+    "eslint-plugin-prettier": "^3.1.3"
   },
   "peerDependencies": {
-    "eslint": "^7.0.0"
+    "eslint": "^7.0.0",
+    "prettier": "^2.0.5"
```
bajtos, line 22: I am not sure if it's a good idea to force all LB projects to check prettier formatting via eslint. It makes me wonder how easy or difficult it would be to enable the new setup only for loopback-next monorepo?

bajtos:
> I'm curious how much it can buy us. In my experience, `npm run prettier:fix` is pretty fast. The main cost is to `npm run eslint` for the first time when the cache is not available.

Here is the problem I see in the current setup (in `master`):

```ts
$ time npm run prettier:check
(...)
All matched files use Prettier code style!

real	0m22.537s
user	0m34.626s
sys	0m1.824s
```

When I run `npm test` to verify that all is good, I have to wait more than 20 seconds just for the prettier check, because prettier does not implement caching.

dougal83:
> Did you check what is the new developer experience in VS Code?
@bajtos With the prettier plugin enabled I have noticed no complaints, although I do have format on save set to run based on vscode settings.  As far as I can tell prettier is executing as normal /w same rules but is now using cache.

I ran some tests on a remote VM(not very well resourced) to simulate benefit to slower machines:

```node
$ time npm run prettier:check
(...)
All matched files use Prettier code style!

real	1m0.297s
user	1m27.764s
sys	0m2.455s
```

```node
$ time npm run eslint (...uncached)

real	9m47.225s
user	10m28.126s
sys	0m58.774s
```

```node
$ time npm run eslint (cached, minor changes)

real	0m2.410s
user	0m2.519s
sys	0m0.251s
```

Working on this machine really is 'cup of tea time when running tests!

packages/eslint-config/package.json
dougal83: I'll look into it. 👍

dougal83:
Some timings on master:

```node
$ time npm run prettier:check (no changes)
(...)
All matched files use Prettier code style!

real    1m14.536s
user    1m43.523s
sys     0m3.443s
```

```node
$ time npm run eslint (cached, minor changes)

real    0m18.628s
user    0m29.016s
sys     0m1.241s
```

```node
$ time npm run eslint (cached, no changes)

real    0m2.892s
user    0m2.900s
sys     0m0.373s
```

Also if not formatting on save, eslint will now highlight prettier rules.

image

If anyone is able to verify time savings it would be appreciated. 👍

raymondfeng:
I tried a few times with your PR:

- first-time lint:
```
real	5m31.220s
user	4m28.761s
sys	1m49.805s
```

- cached lint without changes:
```
real	0m2.155s
user	0m1.822s
sys	0m0.588s

real	0m2.069s
user	0m1.828s
sys	0m0.575s
```
- cached lint after `touch packages/core/src/*.ts`
```
real	0m11.737s
user	0m14.251s
sys	0m2.037s
```

bajtos:
> With the prettier plugin enabled I have noticed no complaints, although I do have format on save set to run based on vscode settings. As far as I can tell prettier is executing as normal /w same rules but is now using cache.

That's great to hear!

Also thank you both @dougal83 and @raymondfeng for checking the speed improvements. I think they are pretty noticeable, aren't they?

dougal83:
> I think they are pretty noticeable, aren't they?

It is a nice little improvement, every little helps. Thank you @raymondfeng for verifying timings.

package.json
```diff
@@ -23,10 +23,12 @@
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-eslint-plugin": "^2.2.1",
     "eslint-plugin-mocha": "^7.0.0",
+    "eslint-plugin-prettier": "^3.1.3",
     "fs-extra": "^9.0.0",
     "husky": "^4.2.5",
     "lerna": "^3.21.0",
     "open-cli": "^6.0.1",
+    "prettier": "^2.0.5",
```
raymondfeng, line 31: Is there a possibility to reuse `prettier` from `@loopback/build`?

raymondfeng: At minimum, please update https://github.com/strongloop/loopback-next/blob/master/bin/sync-dev-deps.js#L42 to include `prettier` so that we can ensure the same versions of `prettier` being used.

raymondfeng: @dougal83 Since you are here, maybe we should use `.prettierrc.js` so that it can reuse https://github.com/strongloop/loopback-next/blob/master/packages/build/config/.prettierrc.

See https://prettier.io/docs/en/configuration.html#basic-configuration

bajtos: +1 to update `sync-dev-deps` to ensure we have the same version of `prettier` in monorepo root and in the build package.

We already have a `.prettierrc` file in our monorepo root, this pull request is not changing that in any way. IMO, reworking the monorepo-level prettier config to inherit from `@loopback/build` is out of scope of this pull request.

raymondfeng: I'm fine to leave `.prettyrc` refactoring out.

dougal83:
Apologies for the delay.  Have I made the change to `sync-dev-deps` correctly?

package.json
```diff
@@ -48,8 +50,8 @@
     "coverage:ci": "node packages/build/bin/run-nyc report --reporter=text-lcov | coveralls",
     "precoverage": "npm test",
     "coverage": "open-cli coverage/index.html",
-    "lint": "npm run prettier:check && npm run eslint && node bin/check-package-locks",
-    "lint:fix": "npm run eslint:fix && npm run prettier:fix",
+    "lint": "npm run eslint && node bin/check-package-locks",
+    "lint:fix": "npm run eslint:fix",
     "eslint": "node packages/build/bin/run-eslint --report-unused-disable-directives --cache .",
     "eslint:fix": "npm run eslint -- --fix",
     "prettier:cli": "node packages/build/bin/run-prettier \"**/*.ts\" \"**/*.js\" \"**/*.md\"",
```
bajtos, line 57: I realized just today that by moving all Prettier linting to eslint, we are no longer going to check markdown files (`**/*.md`).

I am proposing to keep `prettier` as part of our linter checks, but change `prettier:cli` to scan only markdown files.

```suggestion
    "prettier:cli": "node packages/build/bin/run-prettier \"**/*.md\"",
```

raymondfeng: IMO, we should keep the npm scripts to run full prettier check/fix.

dhmlau:
@bajtos, just wondering if this PR is good to land (after resolving the conflict files issue). There are 2 approvals in the PR, but you had a [comment above](https://github.com/strongloop/loopback-next/issues/5460#discussion_r431832141). Thanks.

bajtos:
We need to double-check that markdown files are still linted and update the pull request if needed.

dougal83:
Sorry for the delay, I've been swamped!  I'll check `.md` files with regards to linting as soon as possible.

package.json
dougal83: So, shall we leave as-is for "prettier:cli"?

bajtos: Let's discuss the high-level user experience we are aiming for.

Personally, I'd like the following behavior:

- When I run `npm test` or `npm run lint`, pretty formatting of JavaScript & TypeScript files is handled by `eslint` to leverage the caching mechanism and Markdown files are checked by `prettier`.

In other words, I'd like to use caching mechanism to skip checking files that are known to be valid. Prettier does not support caching yet, but eslint does, that's why I'd like to check pretty formatting using eslint. However, eslint cannot check Markdown files, we have to keep using `prettier` for this step.

dougal83:
One concern of this PR in retrospect is that it could seem that the lint process is frozen as there is no feedback as with direct use of Prettier.  Although time is saved on subsequent runs, do we want people querying the lack of visual feedback as the linter is running for some time?  We could use this PR or not as the benefits of each approach are evaluated.

Have gone with `prettier:docs` scripts to avoid changes to `prettier:cli` as per comments above.

bajtos:
> One concern of this PR in retrospect is that it could seem that the lint process is frozen as there is no feedback as with direct use of Prettier. Although time is saved on subsequent runs, do we want people querying the lack of visual feedback as the linter is running for some time? We could use this PR or not as the benefits of each approach are evaluated.

That's a valid concern.

In my experience so far, `npm run lint` already seems to be frozen in the eslint step for several minutes on the first run, so the problem is already present. Is this pull request making it any worse (extending the time needed to check prettier formatting)?

bajtos:
Maybe squash all commits into a single one?

dougal83:
@raymondfeng Ready for review.  Up to you for merge/close. 👍

PR #5485, opened 2020-05-19
chore: add a script to lint package metadata

spirAde:
This makes it possible to verify that all packages in the monorepo are registered in all required places and have the expected metadata in their package.json

Issue: #4673

## Checklist


spirAde:
@raymondfeng I have to recreate my PR, because I made a mistake with branches in my previous PR. I fixed all mentioned concers

spirAde:
@raymondfeng If you want I can also update all licenses headers in project and create new PR for this

docs/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/strongloop/loopback-next",
+    "url": "https://github.com/strongloop/loopback-next.git",
```
raymondfeng, line 32: My personal preference is `"url": "https://github.com/strongloop/loopback-next" to avoid redirect.

spirAde: Fixed

raymondfeng:
@spirAde Please rebase against latest master to pick up a fix for the CI build.

raymondfeng:
We also use `rebase` instead of `merge`commits.

spirAde:
@raymondfeng Done. Thank you for feedback

bin/check-package-metadata.js
```diff
@@ -0,0 +1,149 @@
...
+
+    if (pkg.license !== rootPkg.license) {
+      wrongFieldValueError(packageName, 'license', rootPkg.license);
+    }
+
+    const isCorrectRepositoryUrl =
+      pkg.repository &&
+      pkg.repository.url &&
+      pkg.repository.url.includes(
+        'https://github.com/strongloop/loopback-next',
```
bajtos, line 112: I think the repository url should always end with `.git` extension. As far as I understand it, https://github.com/strongloop/loopback-next is pointing to project web page on GitHub, while https://github.com/strongloop/loopback-next.git is pointing to the git repository.

Are there any reasons why we would use https://github.com/strongloop/loopback-next in some places and https://github.com/strongloop/loopback-next.git in others?

I am proposing to use (and enforce) https://github.com/strongloop/loopback-next.git everywhere.

Thoughts?

bin/check-package-metadata.js
```diff
@@ -0,0 +1,149 @@
...
+    const isPackageExist =
+      monorepoFile.includes(p.name) || monorepoFile.includes(packagePath);
+
+    if (!isPackageExist) {
+      throw new Error(
+        `MONOREPO.md doesn't contain mention about the ${p.name}`,
+      );
+    }
+  }
+}
```
bajtos, line 42: Can we also assert that the entires in MONOREPO.md are sorted lexicographically? I've encounter this problem recently when reviewing a pull request.

I am fine to leave such check out of scope of this pull request.

bin/check-package-metadata.js
```diff
@@ -0,0 +1,149 @@
...
+
+  for (const p of packages) {
+    if (excludes.includes(p.name)) {
+      continue;
+    }
+
+    const packagePath = path.relative(rootPath, p.location);
+
+    if (!codeOwnersFile.includes(packagePath)) {
+      throw new Error(
```
bajtos, line 57: Reporting violations via `throw new Error` is not very user friendly, because only the first error is reported and users have to repeatedly re-run the linter to discover all problems.

A better solution is to re-design the check functions to return either `undefined` on success or a list of error messages on failure.

bin/check-package-metadata.js
```diff
@@ -0,0 +1,149 @@
...
+
+  await checkLicenseFiles(packages);
+  await checkMonorepoFile(packages, rootPath);
+  await checkCodeOwnersFile(packages, rootPath);
+  await checkPkgsPackageJson(packages, rootPkg);
+}
+
+if (require.main === module) {
+  checkPackagesMetadata().catch(err => {
+    console.error(err);
```
bajtos, line 146: Please explain the users how they can re-run the check locally to verify that all problems have been fixed. 

Example message:

```
Some of the package are not following loopback-next guidelines:

- @loopback/some-name directory doesn't exist in monorepo
- (...more messages...)

Please fix the problems listed above and run the following command to verify:
  npm run check-package-metadata
```

spirAde: @bajtos Thank you for feedback. Do you offer to gather all the information and return list of all problem packages? Because I chose another way, the script throws the error on each step, which will be encountered, while the user will not fix all problem places. If you consider your approach as more appropriate I will fix the script, but I need your final decision.

bin/check-package-metadata.js
raymondfeng: I proposed to remove `.git` for a few reasons:

1. The main usage of repository url is for the npmjs.com to render a link to the github repo. With `.git`, it gets redirected back to https://github.com/strongloop/loopback-next

2. git clone works with both urls (https://github.com/strongloop/loopback-next.git or https://github.com/strongloop/loopback-next).

bin/check-package-metadata.js
spirAde: @bajtos After some thought, I come to the conclusion that you are right. Otherwise the user should run script again and again instead of get the full list one time, it's more user-friendly approach. I'will fix it

spirAde:
@bajtos Thank you. I'm working on fixing all mentioned concerns

spirAde:
@bajtos @raymondfeng fixed concerns, added jsdoc

bin/check-package-metadata.js
spirAde: @bajtos What do you think to generate `MONOREPO.md` dynamically before a new release? For example in the invocation of `npm run version` command

bin/check-package-metadata.js
bajtos: Quoting from https://docs.npmjs.com/files/package.json#repository

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It’s for computers.

The examples in npmjs docs are showing `.git` URLs too.

```
"repository": {
  "type" : "git",
  "url" : "https://github.com/npm/cli.git"
}
```

Regardless of which flavor we decide to use, we should use it consistently in all packages. In the current proposal, some packages are using `loopback-next.git` while others use `loopback-next`.

https://github.com/strongloop/loopback-next/blob/a97266664bb5ec0a379cc131d96e05ea138e3791/packages/cli/package.json#L91-L95

https://github.com/strongloop/loopback-next/blob/a97266664bb5ec0a379cc131d96e05ea138e3791/packages/rest/package.json#L82-L86

bin/check-package-metadata.js
bajtos: I like the idea to generate `MONOREPO.md` automatically. I think we should update it as soon as a new package is added to the project and don't wait until a release happens. See `npm run update-ts-project-refs` for a similar automated step. Even with the automated generator in place, I think it's still a good idea to have a linter check run by CI, because sometimes people forget to commit the generated updates.

examples/file-transfer/package.json
```diff
@@ -62,5 +62,6 @@
     "example",
     "file",
     "upload"
-  ]
+  ],
+  "copyright.owner": "IBM Corp."
```
bajtos, line 66: Can we use somewhat consistent ordering of package.json fields across all packages please? I see that other packages have `copyright.owner` right after the `author` field, can we do the same everywhere you are adding `copyright.owner`?

examples/validation-app/package.json
```diff
@@ -64,5 +64,6 @@
     "LoopBack",
     "example",
     "validation"
-  ]
+  ],
+  "copyright.owner": "IBM Corp."
```
bajtos, line 68: Ditto. Let's put this right after `author` field.

packages/cli/package.json
```diff
@@ -170,5 +171,8 @@
     }
   },
   "copyright.owner": "IBM Corp.",
-  "license": "MIT"
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  }
```
bajtos, line 177: Similarly here, let's move `copyright.owner`, `license` and `publishConfig` closer to the top.

bin/check-package-metadata.js
```diff
@@ -0,0 +1,192 @@
...
+const fs = require('fs-extra');
+
+const Project = require('@lerna/project');
+
+/**
+ * Check existence of LICENSE file in the monorepo packages
+ * @param {Package[]} packages A list of @lerna/project packages
+ */
+async function checkLicenseFiles(packages) {
+  return new Promise(resolve => {
```
bajtos, line 24: `new Promise` is a kind of an anti-pattern, the only place where it's justified is when you need to wrap callback-based APIs.

```ts
async function checkLicenseFiles(packages) {
  const errors = [];
  for (const p of packages) {
    if (!fs.existsSync(path.join(p.location, 'LICENSE'))) {
      errors.push(`${p.name} directory doesn't contain LICENSE file`);
    }
   }
  return errors;
}
```

Your current implementation is not async, so you can simply change `async function` to a regular `function`. 

However, I think it would be better to check the files in parallel to get better performance.

```ts
function checkLicenseFiles(packages) {
  return Promise.all(packages.map(async (p) => {
    const exists = await fs.pathExists(path.join(p.location, 'LICENSE'));
    return exists ? undefined : `${p.name} directory doesn't contain LICENSE file`
  });
}
```

bin/check-package-metadata.js
```diff
@@ -0,0 +1,192 @@
...
+ * @param {string} rootPath A project.rootPath
+ */
+async function checkMonorepoFile(packages, rootPath) {
+  const errors = [];
+  const monorepoFile = await fs.readFile('docs/site/MONOREPO.md', 'utf8');
+
+  for (const p of packages) {
+    const packagePath = path.relative(rootPath, p.location);
+
+    const isPackageExist =
```
bajtos, line 49:
```suggestion
    const packageExists =
```

bin/check-package-metadata.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return errors;
+}
+
+function formatErrorsText(errors) {
+  const errorPackages = errors.map(error => `- ${error}`).join('\n');
+
+  return [
+    'Some of the package are not following loopback-next guidelines:',
```
bajtos, line 158:
```suggestion
    'Some of the packages are not following loopback-next guidelines:',
```

bin/check-package-metadata.js
spirAde: @bajtos Ok, I'm ready to implement it after this PR

bin/check-package-metadata.js
spirAde: Agree

bin/check-package-metadata.js
raymondfeng: @spirAde You decide which flavor to use. I agree that consistency is the only purpose.

bin/check-package-metadata.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  function getErrorText(packageName, field) {
+    return `${packageName} package.json doesn't contain ${field} field or has incorrect value`;
+  }
+
+  for (const p of pkgs) {
+    const pkg = fs.readJsonSync(path.join(p.location, 'package.json'));
+
+    const isPublicAccess =
```
raymondfeng, line 106: Let's also check the following properties:

```json
{
  "main": "dist/index.js", 
  "types": "dist/index.d.ts"
}
```

spirAde: @raymondfeng , I'd checked the field `main` and `types` in the packages, and noticed the packages: `@loopback/docs`, `@loopback/build`, `@loopback/cli`, `@loopback/eslint-config` should be excluded from `checkPkgsPackageJson` method. I found it's ugly to add this kind of checks in the `checkPkgsPackageJson` method. I offer to standardize validation process:

```js
function defaultErrorText(packet, field) {
  return `${packet.name} package.json doesn't contain ${field} field or has incorrect value`;
}

function getPkgField(pkg, field) {
  return (
    pkg[field] ||
    field.split('.').reduce((accumulator, part) => {
      return accumulator && accumulator[part];
    }, pkg)
  );
}

function applyRule(packet, pkg, field, rule) {
  if (typeof rule === 'function') {
    return rule(packet, pkg);
  }

  return rule === getPkgField(pkg, field);
}

function applyRules(packet, pkg, field, rules) {
  if (!Array.isArray(rules)) {
    rules = [rules];
  }

  return rules.reduce(
    (accumulator, {rule, exclude = [], getErrorText = defaultErrorText}) => {
      if (exclude.includes(packet.name)) {
        return accumulator;
      }

      if (!applyRule(packet, pkg, field, rule)) {
        accumulator.push(getErrorText(packet, field));
      }

      return accumulator;
    },
    [],
  );
}

function applySchema(packet, schema) {
  const errors = [];
  const pkg = fs.readJsonSync(path.join(packet.location, 'package.json'));

  for (const field in schema) {
    errors.push(...applyRules(packet, pkg, field, schema[field]));
  }

  return errors;
}

async function checkPkgsPackageJson(packages, rootPkg) {
  const SCHEMA = {
    author: {
      rule: rootPkg.author,
    },
    'copyright.owner': {
      rule: getCopyrightOwner(rootPkg),
    },
    license: {
      rule: rootPkg.license,
    },
    'publishConfig.access': {
      rule: 'public',
    },
    main: {
      rule: 'dist/index.js',
      exclude: [
        '@loopback/docs',
        '@loopback/build',
        '@loopback/cli',
        '@loopback/eslint-config',
      ],
    },
    types: {
      rule: 'dist/index.d.ts',
      exclude: [
        '@loopback/docs',
        '@loopback/build',
        '@loopback/cli',
        '@loopback/eslint-config',
      ],
    },
    'repository.url': {
      rule: 'https://github.com/strongloop/loopback-next.git',
    },
    'repository.directory': [
      {
        rule: (_, pkg) => pkg.repository.directory,
      },
      {
        rule: (_, pkg) => fs.existsSync(pkg.repository.directory),
        getErrorText: p => `${p.name} directory doesn't exist in the monorepo`,
      },
    ],
  };

  const errors = [];
  const pkgs = packages.filter(p => !p.private);

  function getCopyrightOwner(pkg) {
    return pkg['copyright.owner'] || (pkg.copyright && pkg.copyright.owner);
  }

  for (const p of pkgs) {
    errors.push(...applySchema(p, SCHEMA));
  }

  return errors;
}
```

I guess it's easy to support and easy to add new rules. But maybe it's overengineering and doesn't make sense. Anyway, please give your feedback about my suggestion.

raymondfeng: Haha, it is actually simpler. Those packages are not TypeScript modules. We can check the presence of tsconfig.json.

spirAde: OMG) Ok, thank you)

bin/check-package-metadata.js
spirAde: @raymondfeng @bajtos Let's finalize the decision, please)

raymondfeng: I'm fine either way.

spirAde:
@raymondfeng Resolved last conflicts and fixed `example/todo-jwt` package.json missed package.json fields.

raymondfeng:
@spirAde Thank you for the contribution with great patience! 🥇

PR #5494, opened 2020-05-19
fix(security): support type number in securityId

frbuceta:
I have a model with the id of type number and I have to change it to string and then change it back to number. Hence this PR

## Checklist

### DIFF
--- a/examples/passport-login/src/controllers/user.controller.ts
+++ b/examples/passport-login/src/controllers/user.controller.ts
@@ -138,7 +138,7 @@ export class UserLoginController {
     @inject(SecurityBindings.USER) profile: UserProfile,
   ) {
     const user = await this.userRepository.findById(
-      parseInt(profile[securityId]),
+      parseInt(<string>profile[securityId]),
       {
         include: [
           {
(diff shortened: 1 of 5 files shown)


raymondfeng:
@frbuceta The `UserProfile` is an open interface. If you need to use `number` as the user id, can you just have a `getter` for `[securityId]` to return string value?

frbuceta:
> @frbuceta The `UserProfile` is an open interface. If you need to use `number` as the user id, can you just have a `getter` for `[securityId]` to return string value?

How would I do something like that?

raymondfeng:
```ts
const userProfile: UserProfile = {
  get [securityId]() { return String(this.id); },
  id: 123
}
```

frbuceta:
> ```ts
> const userProfile: UserProfile = {
>   get [securityId]() { return String(this.id); },
>   id: 123
> }
> ```

Of course but I have an id of type number that I pass to string and when I get it back I need it in number

```ts
const x: UserProfile = {
  get [securityId]() { return parseInt(this.id); },
  set [securityId](value) { this.id = value },
  id: '123'
}
```
This does not work

frbuceta:
I have deleted a comment. I was doing the same as you but different I already understand

frbuceta:
I close the PR and leave comments for future questions of the users

PR #5507, opened 2020-05-20
feat: boot application with config scripts

deepakrkris:
## Checklist

### DIFF
--- a/examples/passport-login/src/application.ts
+++ b/examples/passport-login/src/application.ts
@@ -5,35 +5,10 @@
 
 import {BootMixin} from '@loopback/boot';
 import {RepositoryMixin} from '@loopback/repository';
-import {RestApplication, toInterceptor} from '@loopback/rest';
+import {RestApplication} from '@loopback/rest';
 import {ServiceMixin} from '@loopback/service-proxy';
 import {MySequence} from './sequence';
-import {AuthenticationComponent} from '@loopback/authentication';
-import {
-  FaceBookOauth2Authorization,
-  GoogleOauth2Authorization,
-  Oauth2AuthStrategy,
-  LocalAuthStrategy,
-  SessionStrategy,
-  BasicStrategy,
-} from './authentication-strategies';
-import {
-  FacebookOauth,
-  GoogleOauth,
-  CustomOauth2,
-  FacebookOauth2ExpressMiddleware,
-  GoogleOauth2ExpressMiddleware,
-  CustomOauth2ExpressMiddleware,
-} from './authentication-strategy-providers';
-import {
-  SessionAuth,
-  FacebookOauthInterceptor,
-  GoogleOauthInterceptor,
-  CustomOauth2Interceptor,
-} from './authentication-interceptors';
-import {PassportUserIdentityService, UserServiceBindings} from './services';
-import {ApplicationConfig, createBindingFromClass} from '@loopback/core';
-import {CrudRestComponent} from '@loopback/rest-crud';
+import {ApplicationConfig} from '@loopback/core';
 import passport from 'passport';
 
 export class OAuth2LoginApplication extends BootMixin(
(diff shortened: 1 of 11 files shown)


raymondfeng:
This seems to be subset of https://github.com/strongloop/loopback-next/pull/5332

deepakrkris:
> This seems to be subset of #5332

@raymondfeng  I created this PR with respect to the blog story for apic integration, for letting users inject config scripts in a `config`  folder for adding the apic spec enhancer. It would be much better option than asking readers to go edit the application.ts

deepakrkris:
@raymondfeng https://github.com/strongloop/loopback-next/pull/5332 is able to support some simple config functions like
export function addApicEnhancer(app: Application) {
  app.addComponent(ApicConnectExtension);
}
then I can close this PR.

examples/passport-login/src/application.ts
```diff
@@ -70,44 +31,4 @@ export class OAuth2LoginApplication extends BootMixin(
...
-    this.add(createBindingFromClass(SessionStrategy));
-    this.add(createBindingFromClass(BasicStrategy));
-    // Express style middleware interceptors
-    this.bind('passport-init-mw').to(toInterceptor(passport.initialize()));
-    this.bind('passport-session-mw').to(toInterceptor(passport.session()));
-    this.bind('passport-facebook').toProvider(FacebookOauthInterceptor);
-    this.bind('passport-google').toProvider(GoogleOauthInterceptor);
-    this.bind('passport-oauth2').toProvider(CustomOauth2Interceptor);
-    this.bind('set-session-user').toProvider(SessionAuth);
-  }
```
bajtos, line 112: As I have explained in the past, we should distinguish between two kinds of configuration:
- how different components and extensions are wired up together to form the application
- environment-specific configuration

Components, service & provider bindings, etc. are wiring. Port configuration, datasource connection strings, etc. are configuration.

I would like to preserve this distinction in the design for`config/*` files too. These files should contribute environment-specific configuration, they should not be used to wire up application components.

At the same time, I do recognize the need for a conventional loading of application artifacts like components and strategy bindings. IMO, we should implement dedicated booters for these building blocks.

By looking at the list of bindings above, I see the following kinds of artifacts to implement booters for:

(1)
```ts
this.add(createBindingFromClass(...));
```

Let's create a generic "binding class booter" that can create a binding from a class (or a provider class) exported by a file. This can be similar to the existing "service booter" - see https://github.com/strongloop/loopback-next/blob/b6edeebcf1c9727fffff03b2270812a7a0c8f403/packages/boot/src/booters/service.booter.ts#L52-L61

Proposed convention: 
- file name `src/bindings/{name}.binding.ts`
- the file should export a class to be bound to the context

As an alternative, we can recommend users to put classes like `PassportUserIdentityService` and `LocalAuthStrategy` to `src/services/*.service.ts` and use the existing service booter to pick them up, so that no changes in `@loopback/boot` are needed.

Personally, I would start with that approach and see how far it can get us and how does the new project layout feels to work with.

(2)
```ts
this.bind('passport-init-mw').to(toInterceptor(passport.initialize()));
```

We already have "interceptor booter", see https://github.com/strongloop/loopback-next/blob/b6edeebcf1c9727fffff03b2270812a7a0c8f403/packages/boot/src/booters/interceptor.booter.ts#L57-L64

At the moment, it supports only interceptor classes. Since `toInterceptor` is returning a different type, I think we should look into ways how to extend the existing interceptor booter to support middleware-based interceptors too.

(3)
```ts
    this.component(AuthenticationComponent);
```

Personally, I think it's ok to keep `app.component()` calls inside the Application constructor because I expect most applications to use only few components.

If you strongly feel that we should move this part somewhere else, then let's open a new issue to discuss how a "component booter" would look like. For example, do we want to provide the component configuration alongside the component class? Conceptually, think of `app.component(MyComponent, {/* component config */})` that we have in LB3.

(4)
```ts
    passport.serializeUser(function (user: any, done) {
       done(null, user);
     });
```

This is a kind of an oddball. I am surprised that we have to provide these functions directly via `passport` API, shouldn't the LB4 passport extension provide a LB4 way for contributing these two functions? Perhaps via `UserSerializer` interface that the extension obtains from a well-defined binding key.

raymondfeng:
I had some discussion with @deepakrkris. We agree that his requirement was to have a bootable set of bindings for `assembling` the application (wirings in your term).

We should be able to add a booter that discovers `src/bindings` or `src/components` and binds them to the application.

PR #5508, opened 2020-05-21
feat(cli): add `lb4 openapi-spec` command to save openapi spec to json or yaml

raymondfeng:
## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [x] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [x] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5521, opened 2020-05-22
feat: generate src/openapi-spec.ts to export openapi spec into a file

raymondfeng:
Alternative option to https://github.com/strongloop/loopback-next/pull/5508

## Checklist

### DIFF
--- a/examples/todo/src/openapi-spec.ts
+++ b/examples/todo/src/openapi-spec.ts
@@ -1,0 +1,28 @@
+// Copyright IBM Corp. 2018,2020. All Rights Reserved.
+// Node module: @loopback/example-todo
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {ApplicationConfig} from '@loopback/core';
+import {TodoListApplication} from './application';
+
+/**
+ * Export the OpenAPI spec from the application
+ */
+async function exportOpenApiSpec(): Promise<void> {
+  const config: ApplicationConfig = {
+    rest: {
+      port: +(process.env.PORT ?? 3000),
+      host: process.env.HOST ?? 'localhost',
+    },
+  };
+  const outFile = process.argv[2] ?? '';
+  const app = new TodoListApplication(config);
+  await app.boot();
+  await app.exportOpenApiSpec(outFile);
+}
+
+exportOpenApiSpec().catch(err => {
+  console.error('Fail to export OpenAPI spec from the application.', err);
+  process.exit(1);
+});
--- a/packages/rest/src/__tests__/integration/export-openapi-spec.integration.ts
+++ b/packages/rest/src/__tests__/integration/export-openapi-spec.integration.ts
@@ -1,0 +1,118 @@
+// Copyright IBM Corp. 2019,2020. All Rights Reserved.
+// Node module: @loopback/rest
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {ApplicationConfig} from '@loopback/core';
+import {expect, givenHttpServerConfig, TestSandbox} from '@loopback/testlab';
+import fs from 'fs';
+import {safeLoad} from 'js-yaml';
+import path from 'path';
+import {format} from 'util';
+import {get, RestApplication} from '../..';
+
+const sandbox = new TestSandbox(path.resolve(__dirname, '../../../.sandbox'));
+
+describe('exportOpenApiSpec', () => {
+  let app: MyApp;
+  let lastLog = '';
+
+  const log = (formatter: unknown, ...args: unknown[]) => {
+    lastLog = format(formatter, ...args);
+  };
+
+  const expectedSpec = {
+    openapi: '3.0.0',
+    info: {
+      title: 'LoopBack Application',
+      version: '1.0.0',
+    },
+    paths: {
+      '/hello': {
+        get: {
+          'x-controller-name': 'MyController',
+          'x-operation-name': 'hello',
+          tags: ['MyController'],
+          responses: {
+            '200': {
+              description: 'Return value of MyController.hello',
+            },
+          },
+          operationId: 'MyController.hello',
+        },
+      },
+    },
+    servers: [
+      {
+        url: '/',
+      },
+    ],
+  };
+
+  beforeEach(async () => {
+    lastLog = '';
+    await givenApp();
+  });
+
+  it('saves the spec to a json file for RestApplication', async () => {
+    const file = path.join(sandbox.path, 'openapi.json');
+    await app.exportOpenApiSpec(file, log);
+    expectJsonSpec(file);
+  });
+
+  it('saves the spec to a json file for RestServer', async () => {
+    const file = path.join(sandbox.path, 'openapi.json');
+    await app.restServer.exportOpenApiSpec(file, log);
+    expectJsonSpec(file);
+  });
+
+  it('writes the spec as a json document to console', async () => {
+    await app.exportOpenApiSpec('-', log);
+    expect(JSON.parse(lastLog)).to.eql(expectedSpec);
+  });
+
+  it('saves the spec to a yaml file', async () => {
+    const file = path.join(sandbox.path, 'openapi.yaml');
+    await app.restServer.exportOpenApiSpec(file, log);
+    expect(lastLog.match(/The OpenAPI spec is saved to .+ openapi\.yaml$/));
+    const content = fs.readFileSync(file, 'utf-8');
+    expect(safeLoad(content)).to.eql(expectedSpec);
+  });
+
+  it('saves the spec to a yml file', async () => {
+    const file = path.join(sandbox.path, 'openapi.yml');
+    await app.restServer.exportOpenApiSpec(file, log);
+    expect(lastLog.match(/The OpenAPI spec is saved to .+ openapi\.yml$/));
+    const content = fs.readFileSync(file, 'utf-8');
+    expect(safeLoad(content)).to.eql(expectedSpec);
+  });
+
+  after(() => sandbox.reset());
+
+  function expectJsonSpec(file: string) {
+    expect(lastLog.match(/The OpenAPI spec is saved to .+ openapi\.json$/));
+    const content = fs.readFileSync(file, 'utf-8');
+    expect(JSON.parse(content)).to.eql(expectedSpec);
+  }
+
+  class MyController {
+    @get('/hello')
+    hello() {
+      return 'Hello';
+    }
+  }
+
+  class MyApp extends RestApplication {
+    constructor(config: ApplicationConfig) {
+      super(config);
+      this.controller(MyController);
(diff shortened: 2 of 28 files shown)


examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import fs from 'fs';
+import {dump} from 'js-yaml';
+import {TodoListApplication} from './application';
+
+/**
+ * Save the OpenAPI spec to the given file
+ * @param outFile - File name for the spec
+ * @param options - Application config
+ */
+export async function dumpOpenApiSpec(
```
bajtos, line 16: Perhaps `saveOpenApiSpec` or `exportOpenApiSpec` would be a better name?

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  outFile: string,
+  options?: ApplicationConfig,
+) {
+  options = options ?? {};
+  options.rest = {...options.rest, listenOnStart: false};
```
bajtos, line 21: Let's not modify the `options` object provided by the caller.

```suggestion
  options = {...options};
  options.rest = {...options.rest, listenOnStart: false};
```

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  const fileName = outFile.toLowerCase();
+  if (fileName.endsWith('.yaml') || fileName.endsWith('.yml')) {
+    const yaml = dump(spec);
+    fs.writeFileSync(outFile, yaml, 'utf-8');
+  } else {
+    const json = JSON.stringify(spec, null, 2);
+    fs.writeFileSync(outFile, json, 'utf-8');
+  }
+  console.log('The OpenAPI spec is saved to %s.', outFile);
+  return spec;
```
bajtos, line 40: I am concerned about the amount of code we are generating here. First, it makes the `open-api.ts` script more difficult to understand. Secondly, and this is more important, if we discover a bug in this code, then we won't be able to distribute the fix to our users.

I am proposing to move the implementation of this function to `RestApplication` and/or `RestServer` class, so that the scaffolded function can be shortened to something like this:

```ts
const app = new TodoListApplication(options);
await app.boot();
await app.exportOpenApiSpec(outFile);
```

packages/rest/src/rest.server.ts
```diff
@@ -987,6 +988,37 @@ export class RestServer extends BaseMiddlewareRegistry
...
+   * - `json` or other: JSON
+   * If the outFile is `''` or `'-'`, the spec is written to the console.
+   * @param log - Log function, default to `console.log`
+   */
+  async exportOpenApiSpec(
+    outFile: string,
+    log = console.log,
+  ): Promise<OpenApiSpec> {
+    const spec = await this.getApiSpec();
+    if (outFile === '-' || outFile === '') {
```
bajtos, line 1006: What if `outFile` is not provided at all (equals to `undefined`)? Can we please treat that case as "print to stdout" too?

packages/rest/src/rest.server.ts
```diff
@@ -987,6 +988,37 @@ export class RestServer extends BaseMiddlewareRegistry
...
+    }
+    const fileName = outFile.toLowerCase();
+    if (fileName.endsWith('.yaml') || fileName.endsWith('.yml')) {
+      const yaml = safeDump(spec);
+      fs.writeFileSync(outFile, yaml, 'utf-8');
+    } else {
+      const json = JSON.stringify(spec, null, 2);
+      fs.writeFileSync(outFile, json, 'utf-8');
+    }
+    log('The OpenAPI spec is saved to %s.', outFile);
```
bajtos, line 1019: English is not my mother tongue, but since the saving operation has completed, I think we should use past tense here?

```suggestion
    log('The OpenAPI spec was saved to %s.', outFile);
```

packages/rest/src/__tests__/integration/export-openapi-spec.integration.ts
```diff
@@ -0,0 +1,125 @@
...
+    expect(JSON.parse(lastLog)).to.eql(expectedSpec);
+  });
+
+  it('saves the spec to a yaml file', async () => {
+    const file = path.join(sandbox.path, 'openapi.yaml');
+    await app.restServer.exportOpenApiSpec(file, log);
+    expect(lastLog.match(/The OpenAPI spec is saved to .+ openapi\.yaml$/));
+    const content = fs.readFileSync(file, 'utf-8');
+    expect(safeLoad(content)).to.eql(expectedSpec);
+    expect(bootCalled).to.be.true();
```
bajtos, line 82: I find `bootCalled` checks redundant. The test is relying on a beforeEach hook to setup the application and boot it up, and the test does not call any boot-related APIs. Why would ever `bootCalled` be false?

packages/rest/src/__tests__/integration/export-openapi-spec.integration.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  class MyApp extends RestApplication {
+    constructor(config: ApplicationConfig) {
+      super(config);
+      this.controller(MyController);
+    }
+    async boot(): Promise<void> {
+      bootCalled = true;
```
bajtos, line 116: Do we need to deal with booting at all? RestApplication does not care how the artifacts are registered, that's why we have `BootMixin` that can be applied to an app (or not).

I am proposing to simplify the tests in this file and remove everything related to booting.

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,34 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+import {ApplicationConfig} from '@loopback/core';
+import {OpenAPIObject} from '@loopback/rest';
+import {TodoListApplication} from './application';
+
+/**
+ * Export the OpenAPI spec from the application
+ */
+async function exportOpenApiSpec(): Promise<OpenAPIObject> {
```
bajtos, line 13: I find this design weird. At one hand, this method is returning `OpenAPIObject` which leaves the impression that perhaps it can be called from other JS/TS modules. On the other hand, the implementation reads `process.argv[2]`, which is unlikely to contain a meaningful value when this function is called from another module.

Personally, I see little value in exporting this function as a public JS/TS API. It's easy to obtain application's OpenAPI spec programmatically via `app.boot()` and `app.getApiSpec()`, plus callers will need to carefully consider what ApplicationConfig to use (`process.env`-based approach may not be the right solution).

I am proposing to simplify `src/openapi-spec.ts` along the lines of `src/migrate.ts`.

```ts
async function exportOpenApiSpec(): Promise<void> {
 // ...
}

exportOpenApiSpec().catch(err => {
  console.error(/*...*/);
  process.exit(1);
});
```

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  const config: ApplicationConfig = {
+    rest: {
+      port: +(process.env.PORT ?? 3000),
+      host: process.env.HOST ?? 'localhost',
+      openApiSpec: {
+        setServersFromRequest: true,
```
bajtos, line 19: AFAIK, `setServersFromRequest` is used only when the OpenAPI spec is requested via HTTP. Since we are calling `getApiSpec` directly, I believe the `servers` field will be always populated using `host` and `port` values from the config, and thus the entire `openApiSpec` section can be removed from the config here.

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  };
+  const outFile = process.argv[2] ?? '';
+  const app = new TodoListApplication(config);
+  await app.boot();
+  return app.exportOpenApiSpec(outFile);
+}
+
+if (require.main === module) {
+  exportOpenApiSpec().catch(err => {
+    console.error('Fail to export OpenAPI spec from the application.', err);
```
bajtos, line 31:
```suggestion
    console.error('Cannot export OpenAPI spec from the application.', err);
```

raymondfeng: I'll stick with `Fail to` as there is a failure in exporting the spec. IMO, `Cannot` is better when we know why it fails.

packages/rest/src/rest.server.ts
raymondfeng: Fixed with a default value.

packages/rest/src/rest.server.ts
raymondfeng: I used `is` to emphasize the state. I switched to `has been saved` now.

packages/rest/src/__tests__/integration/export-openapi-spec.integration.ts
raymondfeng: removed.

examples/todo/src/openapi-spec.ts
raymondfeng: The function is not exported. I fixed the return type.

examples/todo/src/openapi-spec.ts
raymondfeng: Cleaned.

examples/todo/src/openapi-spec.ts
bajtos: I am fine with using "fail" instead of "cannot. However, shouldn't this be in the past tense?

```suggestion
    console.error('Failed to export OpenAPI spec from the application.', err);
```

cc @emonddr

examples/todo/src/openapi-spec.ts
```diff
@@ -0,0 +1,30 @@
...
+      host: process.env.HOST ?? 'localhost',
+    },
+  };
+  const outFile = process.argv[2] ?? '';
+  const app = new TodoListApplication(config);
+  await app.boot();
+  await app.exportOpenApiSpec(outFile);
+}
+
+if (require.main === module) {
```
bajtos, line 25: What's the purpose of this check? When we don't all `exportOpenApiSpec` then this module is a no-op. Can we remove it please, also for consistency with `src/migrate.ts`?

packages/rest/src/rest.application.ts
```diff
@@ -387,4 +387,18 @@ export class RestApplication extends Application implements HttpServerLike {
...
   }
+
+  /**
+   * Export the OpenAPI spec to the given json or yaml file
+   * @param outFile - File name for the spec. The extension of the file
+   * determines the format of the file.
+   * - `yaml` or `yml`: YAML
+   * - `json` or other: JSON
+   * If the outFile is not provided or its value is `''` or `'-'`, the spec is
+   * written to the console.
```
bajtos, line 398:
```suggestion
   * written to the console using the `log` function.
```

examples/todo/src/openapi-spec.ts
raymondfeng: I think most systems use current tense. For example, `Cannot` instead of `Couldn't`.

PR #5545, opened 2020-05-24
feat(graphql): add an extension for type-graphql integration

raymondfeng:
This PR adds an extension package to provide GraphQL integration using [type-graphql](https://github.com/MichalLytek/type-graphql).

https://github.com/strongloop/loopback-next/blob/graphql/extensions/graphql/README.md

type-graphql

## Checklist


bajtos:
4kLOC of new code, that's a big chunk to review. Who else can help with reviewing this new feature? @raymondfeng @dhmlau

extensions/graphql/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+```sh
+npm install --save @loopback/graphql
+```
+
+## Try it out
+
+```sh
+npm run build
+node dist/__tests__/fixtures/graphql-test/src/index.js
```
bajtos, line 67: Can you please add a new `npm run` script to hide this long path? For example:

```suggestion
npm run demo
```

bajtos:
+1 for depending on a 3rd-party library and not inventing our own solution.

I have few high-level questions:

1. How is this proposal avoiding SELECT N+1 problem? Is `type-graphql` providing data loaders out of the box or is it up to the application developer to be aware of this problem and work around it? (See e.g. https://itnext.io/what-is-the-n-1-problem-in-graphql-dd4921cb3c1a)

2. A naive implementation of GraphQL layer has usually worse performance than a typical REST API version, because it spends a lot of time on parsing & traversing GraphQL AST. Fastify uses [graphql-jit](https://www.npmjs.com/package/graphql-jit) to compile queries into javascript functions that can be further optimized by V8. Depending on the scenario, graphql-jit can be 5x to 30x faster than graphql-js. Does your proposed implementation design allow us to add graphql-jit in a reasonably easy way? Do you perhaps have any other solutions in mind for improving the performance?

3. The biggest benefit I see in a GraphQL layer for LoopBack is the opportunity to leverage our knowledge of models and repository operations and thus make it easier for LB app developers to build standard CRUD-like GraphQL APIs. For example, all GraphQL query resolvers can understand LoopBack's `filter` language and offer things like `limit` and `skip` out of the box. As far as I understand this pull request, it's focusing on the transport side only, i.e. on the mechanics of using type-graphql to provide a generic GraphQL API. I am fine with that as the first increment, but I would like to understand what's the bigger picture and the next steps?

extensions/graphql/src/__tests__/fixtures/graphql-test/src/services/recipe.service.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor(@inject('recipes') private readonly items: Recipe[]) {
+    this.autoIncrementValue = this.items.length;
+  }
+
+  async getAll() {
+    return this.items;
+  }
+
+  async getOne(id: string) {
+    return this.items.find(it => it.id === id);
```
bajtos, line 24: `RecipeService` looks like a poor-man's version of `RecipeRepository`. I am confused why you are writing this new service and not using `RecipeRepository` instead?

dhmlau:
> 4kLOC of new code, that's a big chunk to review. Who else can help with reviewing this new feature? @raymondfeng @dhmlau

Talked to @hacksparrow, he will review this as well. Thanks @hacksparrow!

extensions/graphql/src/__tests__/fixtures/graphql-test/src/services/recipe.service.ts
raymondfeng: Yes.

raymondfeng:
@bajtos Thank you for the feedback.

> I have few high-level questions:
> How is this proposal avoiding SELECT N+1 problem? Is type-graphql providing data loaders out of the box or is it up to the application developer to be aware of this problem and work around it? (See e.g. https://itnext.io/what-is-the-n-1-problem-in-graphql-dd4921cb3c1a)

`type-graphql` does not deal with graph data resolution itself. It simply manages `resolver` classes which have graphql related decorations. I'm exploring the extension points from type-graphql to see how LoopBack can add values. 

- Plug in LoopBack as the IoC container (thus leverages other LoopBack features such as repositories and services)
- Discover/load graphql resolvers

I also take this opportunities to find limitation of type-graphql. For example, it uses a global registry for resolver metadata, which is problematic for us with the ability to start/stop applications.

> A naive implementation of GraphQL layer has usually worse performance than a typical REST API version, because it spends a lot of time on parsing & traversing GraphQL AST. Fastify uses graphql-jit to compile queries into javascript functions that can be further optimized by V8. Depending on the scenario, graphql-jit can be 5x to 30x faster than graphql-js. Does your proposed implementation design allow us to add graphql-jit in a reasonably easy way? Do you perhaps have any other solutions in mind for improving the performance?

I'm not there yet. It's my priority to get the right programming model in place 1st.

> The biggest benefit I see in a GraphQL layer for LoopBack is the opportunity to leverage our knowledge of models and repository operations and thus make it easier for LB app developers to build standard CRUD-like GraphQL APIs. For example, all GraphQL query resolvers can understand LoopBack's filter language and offer things like limit and skip out of the box. As far as I understand this pull request, it's focusing on the transport side only, i.e. on the mechanics of using type-graphql to provide a generic GraphQL API. I am fine with that as the first increment, but I would like to understand what's the bigger picture and the next steps?

That seems to be possible. We can follow what we did for REST CRUD controllers. For example, generating resolver classes from models and relations.

bajtos:
> I'm exploring the extension points from type-graphql to see how LoopBack can add values.

Makes sense. Unfortunately, this was not clear to me from reading the pull request description and skimming through the docs.

> > A naive implementation of GraphQL layer has usually worse performance than a typical REST API version, because it spends a lot of time on parsing & traversing GraphQL AST. Fastify uses graphql-jit to compile queries into javascript functions that can be further optimized by V8. Depending on the scenario, graphql-jit can be 5x to 30x faster than graphql-js. Does your proposed implementation design allow us to add graphql-jit in a reasonably easy way? Do you perhaps have any other solutions in mind for improving the performance?
>
> I'm not there yet. It's my priority to get the right programming model in place 1st.

Sure, I am fine to focus on getting right the programming model first. I just want to avoid the situation where we settle on a programming model that will make it very difficult to optimize the underlying implementation for a good performance.

> We can follow what we did for REST CRUD controllers. For example, generating resolver classes from models and relations.

+1

I think we should also investigate how to build GraphQL schemas from our `ModelDefinition` metadata, similarly how we are building JSON/OpenAPI schemas for the REST layer.

BTW, regarding SELECT 1+N problem and DataLoaders, here is a short video explaining both the issue and a possible solution: https://www.youtube.com/watch?v=SSj4BGkIBGo

MichalLytek:
> also take this opportunities to find limitation of type-graphql. For example, it uses a global registry for resolver metadata, which is problematic for us with the ability to start/stop applications.

The `1.0.0-rc.2` has implemented proper schema isolation, so there shouldn't be any problem with restarting app in dev mode (hot reload), apart from a few more MBs allocated due to storing the references to the "new" classes.

Global storage is not a problem, the standard JS `Reflect` is also global. The issue is with storing the metadata in arrays, not in a `WeakMap` that would allow the garbage collector to clean up the memory.

If you have any questions, feel free to `@ping` me 😉

MichalLytek:
BTW, is the decorators renaming/reexporting intentional?

Are you going to maintain your own docs to not confuse the people with redirecting to `typegraphql.com` site while importing `resolver` decorator from `@loopback/graphql` package?

raymondfeng:
@MichalLytek Thank you for chiming in here.

1. I'll try to see the isolation is good now. I ran into an issue where we have two tests that call `buildSchema` with the same resolver class in the sam process and `type-graphql` complain.

2. For decorators, I'm still deciding. On one hand, I would like to use `type-graphql` decorators as is. One the other, I want to see if it's possible to use a shim decorator so that we can not only configure the metadata for GraphQL but also for LoopBack. You can see that two sets of `decorators` are used on https://github.com/strongloop/loopback-next/blob/graphql/extensions/graphql/src/__tests__/fixtures/graphql-test/src/graphql-types/recipe-type.ts.

3. Are you open to having a joint blog for this module once it's ready to be released?

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+> Experimental packages provide early access to advanced or experimental
+> functionality to get community feedback. Such modules are published to npm
+> using `0.x.y` versions. Their APIs and functionality may be subject to
+> breaking changes in future releases.
+
+## Basic Use
+
+```ts
+export class MyApplication extends BootMixin(RestApplication) {
```
hacksparrow, line 18: It is strange to see **GraphQL**Component with **Rest**Application. If `RestApplication` can be used this way, maybe it should not be named `RestApplication`? Maybe just `WebApplication`?

The application type is then determined by whether you do:

1. `this.component(RestExplorerComponent);`
or
2. `this.component(GraphQLComponent);`
or
3. `this.component(GrpcComponent);`
etc.

I think this observation is an indication that `RestApplication` should be made more generic, and the REST-specific parts should be provided by the Component.

@bajtos @raymondfeng what do you think?

MichalLytek:
> Are you open to having a joint blog for this module once it's ready to be released?

@raymondfeng Sure, we can prepare something together 😉 

If you have any question or issue with the integration, feel free to ping me or DM on gitter.

extensions/graphql/README.md
raymondfeng: REST/GraphQL both use `http/https` as the underlying protocol. `gRPC` uses `http/2`. I don't see a urgent need to rename `RestApplication` to `HttpApplication`.

hacksparrow: Agreed it's not urgent. However, it is not just renaming to `HttpApplication`, we will need to extract the REST-specific parts from `RestApplication` and enable REST functionality the way we enable GraphQL or other functionalities. In the current state, `RestComponent` is loaded by default in `RestApplication `. 

Or will this work?

```ts
class GraphQlApplication extends Application implements HttpServerLike
...
MyApplication extends BootMixin(GraphQlApplication) 
```

Nah, it would be repetitive code or create more files. The better approach would be to rename `RestApplication`, make it more generic, and extract the REST bits out of it.

@bajtos thoughts?

extensions/graphql/src/__tests__/acceptance/graphql-middleware.acceptance.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/graphql
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {createRestAppClient, givenHttpServerConfig} from '@loopback/testlab';
```
hacksparrow, line 6: `createRestAppClient` <-- we should rename it, so we can use it with other API types without confusion.

extensions/graphql/README.md
hacksparrow: If GraphQL support is to be implemented as an extension, REST should be no different.

extensions/graphql/src/__tests__/acceptance/graphql.acceptance.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('GraphQL integration', () => {
+  let server: GraphQLServer;
+
+  before(givenServer);
+  after(stopServer);
+
+  runTests(() => supertest(server.httpServer?.url));
+
+  async function givenServer() {
+    server = new GraphQLServer(undefined, {host: '127.0.0.1', port: 0});
```
hacksparrow, line 25: If it can be `undefined`, why not switch the order and just omit it?

extensions/graphql/src/decorators/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  InputType,
+  Mutation,
+  ObjectType,
+  Query,
+  Resolver,
+  Root,
+} from 'type-graphql';
+
+// export namespace graphql {
+export const arg = Arg;
```
hacksparrow, line 23: Why are we exporting lowercased version?

extensions/graphql/src/decorators/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  FieldResolver,
+  InputType,
+  Mutation,
+  ObjectType,
+  Query,
+  Resolver,
+  Root,
+} from 'type-graphql';
+
+// export namespace graphql {
```
hacksparrow, line 22: Remove this comment and the one below?

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+    this.projectRoot = __dirname;
+    this.component(GraphQLComponent);
+    this.configure(GraphQLBindings.GRAPHQL_SERVER).to({asMiddlewareOnly: true});
+  }
+}
+```
+
+## Configure GraphQLServer
+
+This package can be used in two flavors:
```
hacksparrow, line 30: Need to document how to do it.

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+  constructor(config: ApplicationConfig) {
+    super(config);
+    this.projectRoot = __dirname;
+    this.component(GraphQLComponent);
+    this.configure(GraphQLBindings.GRAPHQL_SERVER).to({asMiddlewareOnly: true});
+  }
+}
+```
+
+## Configure GraphQLServer
```
hacksparrow, line 28: How to configure?

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+  constructor(
+    // constructor injection of service
+    @repository('RecipeRepository')
+    private readonly recipeRepo: RecipeRepository,
+    @service(RecipeService) private readonly recipeService: RecipeService,
+  ) {}
+}
+```
+
+## Boot GraphQL resolvers
```
hacksparrow, line 99: Sounds like an instruction, which it is not. How about "Boot GraphQL resolvers"?

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+    @root() recipe: Recipe,
+    @arg('minRate', type => Int, {defaultValue: 0.0})
+    minRate: number,
+  ): number {
+    return this.recipeService.ratingsCount(recipe, minRate);
+  }
+}
+```
+
+## Use LoopBack dependency injections in resolver classes
```
hacksparrow, line 81: How about "Using LoopBack dependency injections in resolver classes"?

extensions/graphql/README.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+## Configure GraphQLServer
+
+This package can be used in two flavors:
+
+- As a server for LoopBack applications
+- As a middleware for LoopBack REST applications
+
+## Add a resolver
```
hacksparrow, line 35: How about "Adding a resolver"?

hacksparrow:
Implementation-wise, I don't see any obvious issues; the documentation can be better. However, the state of the codebase hints we should take out the REST-specific implementation and convert it into an extension as well. 

> A lot of confusion goes away once LoopBack is looked at as a meta framework, not a REST API framework. However, our documentation, tooling, and examples present it primarily as a REST API framework.
> -- https://github.com/strongloop/loopback-next/issues/5551#issuecomment-633665689

We cannot bake REST into the "minor guts" of LoopBack and call it a general-purpose API framework. Let's create a REST extension, which would then be loaded into the app just like any other extension.

A natural side-effect of creating a REST extension will its specific documentation. I had suggested creating a new framework, but a REST extension is actually better. A documentation of this extension is what most of our users are looking for, it will solve many of the documentation problems.

@bajtos @raymondfeng let's do it!

raymondfeng:
> Implementation-wise, I don't see any obvious issues; the documentation can be better. However, the state of the codebase hints we should take out the REST-specific implementation and convert it into an extension as well.

> A lot of confusion goes away once LoopBack is looked at as a meta framework, not a REST API framework. However, our documentation, tooling, and examples present it primarily as a REST API framework.
-- #5551 (comment)

> We cannot bake REST into the "minor guts" of LoopBack and call it a general-purpose API framework. Let's create a REST extension, which would then be loaded into the app just like any other extension.

> A natural side-effect of creating a REST extension will be its specific documentation. I had suggested creating a new framework, but a REST extension is actually better. A documentation of this extension is what most of our users are looking for, it will solve many of the documentation problems.

IIUC, you are proposing to extract the HTTP inbound processing into a separate module so that it can be used to built `GraphQL` and `OpenAPI` capabilities. Here is how I see it:

- http-server (the base wrapper for incoming http/https support)
- (http/express)-application (common http/https sequence/middleware based on Express)
   - https://github.com/strongloop/loopback-next/blob/master/packages/express/src/express.application.ts (can be seen as v0)
  - https://github.com/strongloop/loopback-next/blob/master/packages/express/src/express.server.ts (can be seen as v0)
- REST/OpenAPI application (contributes its specific actions to the http sequence)
- GraphQL application ((contributes its specific actions to the http sequence)

I agree with the general direction. Let's open a new issue for that discussion.

At the same time, I don't think we should block this PR.

extensions/graphql/src/decorators/index.ts
raymondfeng: See https://github.com/strongloop/loopback-next/pull/5545#issuecomment-634220374

extensions/graphql/README.md
raymondfeng: I'm fine with `Using` or `Use`. Just have to be consistent.

raymondfeng:
@hacksparrow Thank you for the review. PTAL.

hacksparrow:
> IIUC, you are proposing to extract the HTTP inbound processing into a separate module so that it can be used to built GraphQL and OpenAPI capabilities.

That's correct.

> At the same time, I don't think we should block this PR.

Approved.

hacksparrow:
Opened issue for implementing REST/OpenAPI functionality as an extension https://github.com/strongloop/loopback-next/issues/5697.

extensions/graphql/README.md
bajtos: I feel we have bigger fish to fry, I am fine to use _GraphQLComponent with RestApplication_ for initial iterations. Let's build a better understanding of what our users need and how they are going to use GraphQL before we jump on refactorings.

extensions/graphql/README.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+## Add a resolver
+
+Please add `recipe-resolver.ts` to `src/graphql-resolvers`.
+
+```ts
+@resolver(of => Recipe)
+export class RecipeResolver implements ResolverInterface<Recipe> {
```
bajtos, line 83: Please add `import` statements to show where are `ResolverInterface`, `@resolver` and `Recipe` coming from.

extensions/graphql/README.md
```diff
@@ -0,0 +1,221 @@
...
+    return index + 1;
+  }
+
+  @fieldResolver()
+  ratingsCount(
+    @root() recipe: Recipe,
+    @arg('minRate', type => Int, {defaultValue: 0.0})
+    minRate: number,
+  ): number {
+    return this.recipeService.ratingsCount(recipe, minRate);
```
bajtos, line 118: IIUC, this example is showing how to resolve relations, where `Recipe` has many `Rating` instances, and demonstrates SELECT N+1 footgun in practice. When querying 10 recipes including ratingCount, we are going to make 11 queries (service calls): the first to build the list of N recipes, and then N calls to obtain `ratingsCount` for each one.

Since the purpose of this PR is to explore the programming model, I'd like us to discuss the SELECT N+1 issue and find such programming model that makes it easy to avoid that issue.

For example, can users write a data loader that is able to resolve field for multiple instances in a single call?

```ts
export class RecipeResolver implements ResolverInterface<Recipe> {
  @fieldResolver()
   ratingsCount(
     @root() recipes: Recipe[], // <-- IMPORTANT: we want a list of recipes
     @arg('minRate', type => Int, {defaultValue: 0.0})
     minRate: number,
   ): number {
     return this.recipeRepository.ratings.inclusionResolver(
        recipes,
        {scope: {where: {rate: {gt: minRate}}},
        {/* no options*/});
    }
}
```

bajtos: At the same time, I think we should not force all field resolvers to accept an array of items, that would be a suboptimal UX too.

The "canonical" way is to leverage [DataLoader](https://github.com/graphql/dataloader) pattern, where the resolver delegates the logic to a dataloader that the user configures elsewhere. I find this approach as not very user friendly, it feels like a band-aid for broken design to me.

fastify-gql allows users to provide "loaders" in addition to "resolvers", see https://github.com/mcollina/fastify-gql#defineLoaders. Based on my limited understanding: resolvers are used to resolve a single link at a time (`recipe => ratingsCount`) and loaders are used to resolve multiple links at a time (`recipe[] => ratingsCount[]`). I presume that the plugin is building resolvers from the provided loaders behind the scenes, so that app developers don't have to deal with this complexity themselves.

I am open minded about the approach to take in LoopBack, as long as it makes it easy for LB users to avoid SELECT N+1 trap in their applications.

raymondfeng:
@bajtos Thank you for the feedback. I agree it would be ideal to figure out how GraphQL is natively and optimally resolved by model with relations. But I would prefer to defer that to future PRs. I see primary objectivea of experimental extensions are the following:

1. Try to bring another framework/module to LoopBack following the established patterns, such as DI, extension point/extension, and conventional configuration.

2. Provide shadow integration to harvest the best breeds of both sides.

3. Invite community members to try out and give us feedback to lead to golden paths for mostly common use cases before going too deep.

In this context, I would like to have 0.x.y releases as soon as we can.

bajtos:
> I agree it would be ideal to figure out how GraphQL is natively and optimally resolved by model with relations. But I would prefer to defer that to future PRs.

I understand your position @raymondfeng  and agree that implementing advanced concepts like loaders and JIT compilation is out of scope of this initial pull request.

However, as you wrote yourself, the goal of this pull request is _to get the right programming model in place 1st_.

To me, it's important to verify that the programming model keeps the door open to implement the necessary improvements we are already aware of (data loaders to avoid SELECT N+1 problem, JIT compilation to avoid performance issues).

You wrote:

> `type-graphql` does not deal with graph data resolution itself. It simply manages `resolver` classes which have graphql related decorations. I'm exploring the extension points from type-graphql to see how LoopBack can add values. 

Based on my understanding, we will need to implement support for (data) `loader` classes in addition to `resolver` classes. What options do we have in your proposed programming model? Is this something that we can implement on our side (perhaps using a new decorator) easily?

@MichalLytek do you have any opinions on how to allow graphql developers to avoid SELECT N+1 issues? (See the discussion in https://github.com/strongloop/loopback-next/pull/5545#discussion_r439396265). 

I see that type-graphql has the concept of [Field Resolvers](https://typegraphql.com/docs/resolvers.html#field-resolvers) to resolve relations (e.g. `Category` has many `products`) once instance at a time (`recipe => ratingsCount`).

Would you be open to implement "Field Loaders" that can resolve multiple instances at once? (`recipe[] => ratingsCount[]`)? (We may be able to contribute such implementation ourselves.) Example use:

```ts
import { DefaultCrudRepository } from '@loopback/repository';

@Resolver(of => Recipe)
class RecipeResolver implements ResolverInterface<Recipe> {
  constructor(
    private userRepository: DefaultCrudRepository<User, number, UserRelations>, // dependency injection
  ) {}

  @FieldLoader()
  async author(@Root() recipes: Recipe[]) {
    const authors = await this.userRepository.find({
      where:{id: {inq: recipes.map(r => r.userId)}}
    });
    if (authors.length !== recipes.length) throw new SomethingWentWrongError();
    return authors;
  }
}
```

@raymondfeng In one of your earlier comments, you mentioned few type-graphql limitations that may block our adoption:

> I also take this opportunities to find limitation of type-graphql. For example, it uses a global registry for resolver metadata, which is problematic for us with the ability to start/stop applications.

Is this still an issue? Are there any other limitations or short-comings to be aware of?

MichalLytek:
@bajtos There is a feature request for the dataloader integration (batching):
https://github.com/MichalLytek/type-graphql/issues/51
It's on my priority list after the 1.0 release.

But generally speaking, I still prefer doing that at the repository layer (like https://github.com/Mando75/typeorm-graphql-loader), rather than inside resolvers 🧐

raymondfeng:
@bajtos Here is how I see it from the programming model perspective:

1. The type-graphql resolvers can be treated as `controllers` for `graphql`. It supports a `bottom-up` approach that exposes TypeScript classes as graphql operations via decorators (which supply additional mapping metadata). The layer is similar as REST controllers with `openapi-v3` decorators.

2. The `@loopback/graphql` extension so far has added the following values:
- Discover/load resolver classes using a booter
- Expose endpoints via a middleware
- Enable type-graphql resolvers with LoopBack dependency injection so that the `resolution` logic can be fulfilled by repositories and service proxies.

3. I agree with @MichalLytek that the repository layer should be responsible for optimizing data access efficiency such as tackling the N+1 issue. The problem is not specific for graphql, but also applies for LB relations/inclusions too. 

4. As @MichalLytek commented, the global registry of resolvers has been resolved. See https://github.com/strongloop/loopback-next/pull/5545#issuecomment-634203179.

IMO, the code we have is good enough for the 1st experimental release. 

We should continue to explore the following areas afterward:

1. Synergy between GraphQL schemas/types and LoopBack models/properties
2. How to map LB models/relations/inclusions to GraphQL query/mutation
3. Improve/optimize data access for GraphQL style, including the N+1 issue

bajtos:
> @bajtos There is a feature request for the dataloader integration (batching):
> [MichalLytek/type-graphql#51](https://github.com/MichalLytek/type-graphql/issues/51)
> It's on my priority list after the 1.0 release.

👍 

> But generally speaking, I still prefer doing that at the repository layer (like https://github.com/Mando75/typeorm-graphql-loader), rather than inside resolvers 🧐
> Some tools like Prisma 2 has even simple dataloader built-in.

I haven't considered the option of batching the queries at repository/ORM layer, it's an interesting option.

One concern that comes to my mind is tenant isolation - how to ensure the batching logic is applied only on a single request and does not batch queries from multiple requests into a single database query? Wouldn't that introduce a lot of complexity? Essentially, we need the repository layer to keep a per-request state (similar to thread-local storage known from old school web frameworks in .NET/Java, where each request had a dedicated thread to handle it).

See also https://github.com/Mando75/typeorm-graphql-loader#gotchas-:

> This is not a complete replacement for Facebook's dataloader. package. While it does provide some batching, it's primary purpose is to load the relations and fields needed to resolve the query. In most cases, you will most likely not need to use dataloader when using this package. However, I have noticed in my own use that there are occasions where this may need to be combined with dataloader to remove N + 1 queries. One such case was a custom resolver for a many-to-many relation that existed in the GraphQL Schema but not on a database level. In order to completely remove the N+1 queries from that resolver, I had to wrap the TypeORM GraphQL loader in a Facebook DataLoader. If you find that you are in a situation where the TypeORM GraphQL loader is not solving the N+1 problem, please open an issue and I'll do my best to help you out with it.

I did a quick search for dataloader support in Prisma, found the following material:

- https://github.com/prisma/specs/issues/242
- https://dev.to/ahmedelywa/improve-your-graphql-performance-with-prisma-2jia

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "README.md",
+    "dist",
+    "src",
+    "!*/__tests__"
+  ],
+  "dependencies": {
+    "@loopback/boot": "^2.3.1",
+    "@loopback/context": "^3.8.1",
+    "@loopback/core": "^2.7.0",
+    "@loopback/openapi-v3": "^3.4.1",
```
bajtos, line 37: Please remove this dependency and use `@loopback/rest` instead (see https://github.com/strongloop/loopback-next/issues/5692).

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "license": "MIT",
+  "files": [
+    "README.md",
+    "dist",
+    "src",
+    "!*/__tests__"
+  ],
+  "dependencies": {
+    "@loopback/boot": "^2.3.1",
+    "@loopback/context": "^3.8.1",
```
bajtos, line 35: Please remove this dependency and use `@loopback/context` instead (see your own #5625).

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "graphql-test",
```
bajtos, line 2: I find it weird to have a standalone package (`graphql-test`) inside `src` folder. Have you considered changing this package into a proper first-class package in our monorepo and moving the files e.g. to `/fixtures`? Considering that this package is also contributing tests you want to run as part of graphql test suite, perhaps the package should live in `/acceptance` or even `/examples` folder?

extensions/graphql/README.md
```diff
@@ -0,0 +1,245 @@
...
+      "ratingsCount": 1,
+      "averageRating": 1.3333333333333333,
+      "ingredients": ["one", "two", "three"],
+      "numberInCollection": 1
+    }
+  }
+}
+```
+
+## Installation
```
bajtos, line 223: I believe `Installation` goes before `Basic use` - see https://loopback.io/doc/en/contrib/README-guidelines.html

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
raymondfeng: The `package.json` is actually not used.

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
raymondfeng: It's a chicken-egg issue here. On one hand, such application serves as both test fixture and example. Moving it alone to a separate package will create circular dependency (the app pkg has dep on `@loopback/graphql`, which has dev dep on the app).

Let me move it to `src/__examples__` for now.

raymondfeng:
@bajtos PTAL

extensions/graphql/README.md
```diff
@@ -0,0 +1,245 @@
...
+      },
+    };
+  }
+}
+```
+
+## Add a resolver
+
+Please add `recipe-resolver.ts` to `src/graphql-resolvers`.
+
```
bajtos, line 92: As a person new to GraphQL and LoopBack, I would find it difficult to understand what's going on in the code snippet below. It would be nice to provide a brief explanation of concepts here.

1. Individual query/mutation/resolver functions are grouped to resolver classes, similarly to how REST API endpoints are grouped to controller classes.

2. We use TypeScript decorators to provide metadata about individual resolvers. 

3. Explain what is the purpose (effect) of the class-level `@resolver`, method-level `@query`, `@mutation` and `@fieldResolver` decorators.

You may want to consider adding a link to type-graphql documentation too, to provide readers with resources where they can learn more.

extensions/graphql/README.md
```diff
@@ -0,0 +1,245 @@
...
+    private readonly recipeRepo: RecipeRepository,
+    @service(RecipeService) private readonly recipeService: RecipeService,
+  ) {}
+}
+```
+
+## Discover and load GraphQL resolvers
+
+The `GraphQLComponent` contributes a booter that discovers and registers
+resolver classes from `src/graphql-resolvers` during `app.boot()`.
```
bajtos, line 174: Please explain what is the project layout convention - where is the booter looking for files with resolver classes.

extensions/graphql/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "unit": "lb-mocha \"dist/__tests__/unit/**/*.js\"",
+    "demo": "npm run build && node dist/__examples__/graphql-test/src/index.js",
+    "verify": "npm pack && tar xf loopback-graphql*.tgz && tree package && npm run clean"
+  },
+  "author": "IBM Corp.",
+  "copyright.owner": "IBM Corp.",
+  "license": "MIT",
+  "dependencies": {
+    "@loopback/boot": "^2.4.0",
+    "@loopback/core": "^2.9.2",
```
bajtos, line 26: Please move `@loopback/core` to `peerDependencies`, see #5625 and #5698.

extensions/graphql/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@loopback/http-server": "^2.1.10",
+    "@loopback/metadata": "^2.2.3",
```
bajtos, line 28: I think we should use `@loopback/core` instead `@loopback/metadata` for consistency with other extensions.

extensions/graphql/src/__examples__/graphql-test/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) IBM Corp. 2019.
+Node module: graphql-test
+This project is licensed under the MIT License, full text below.
+
```
bajtos, line 4: Is it necessary to replicate the LICENSE file inside `src/__examples__`?

extensions/graphql/src/__tests__/fixtures/graphql-test/package.json
bajtos: I see your point. Here are my concerns:

- If the example includes `package.json`, then we need an automated test verify that the package metadata is valid, this test must be a part of our CI builds. 
- You are introducing a new way for working with test fixtures. This makes maintenance more difficult - maintainers working in the existing code have to learn one more thing, maintainers working on a new feature must decide which fixture style to use.

In general, I think it's a bad idea to use the same code as an example app and for acceptance-level testing. Example apps and test fixtures are serving different kinds of users, they have different requirements. In example apps, the focus is on illustrating the feature and making the design & implementation easy to understand, usually on a very small subset of possible use cases. In test fixtures, we usually want to cover many use cases in a flexible way, often trading ease of learning for better maintainability via code reuse, etc.

Take a look at how we are testing Repository API against connectors: the acceptance tests are not part of `@loopback/repository`, we have test-only projects in `acceptance/repository-*` that are executed as part of our CI build. Similarly we don't have any truly end-to-end test for REST APIs in `@loopback/rest`, we kind of rely on `@examples/todo` and `@examples/todo-list` for acceptance-level tests executing all layers.

I am proposing to rework your example into a proper example project in `examples/*` and break the circular dependency by NOT including example endpoints in GraphQL extension tests. Instead, the example project should come with its own test suite that will be executed as part of `npm test`. If you feel your current extension does not have sufficient test coverage without that example project, then let's find a way how to write unit or integration level tests to change that.

extensions/graphql/src/graphql.container.ts
```diff
@@ -0,0 +1,90 @@
...
+ * container
+ */
+export class LoopBackContainer implements ContainerType {
+  constructor(readonly ctx: Context) {}
+  get(
+    resolverClass: Constructor<unknown>,
+    resolverData: ResolverData<unknown>,
+  ) {
+    debug('Resolving a resolver %s', resolverClass.name, resolverData);
+    const resolutionCtx = new GraphQLResolutionContext(
```
bajtos, line 46: IIUC, you are creating a new resolution context for each resolver class we are instantiating. Personally, I would expect to have a context per incoming request, either instead or in addition to per-resolver-class context.

For example, when executing the example query shown earlier in your pull request, how many context instances are created?

```graphql
query {
  recipe(recipeId: "1") {
     title
     description
     ratings
     creationDate
     ratingsCount(minRate: 2)
     averageRating
     ingredients
     numberInCollection
   }
 }
}
```

More importantly, how can the application developer share state between different resolver instances created while handling a single incoming request? In REST API, it's possible to share state via per-request context (e.g. the authentication layer sets the current user, the authorization layer reads that value),

extensions/graphql/src/graphql.container.ts
```diff
@@ -0,0 +1,90 @@
...
+      resolverClass,
+      resolverData,
+    );
+    const resolverBinding = createBindingFromClass(resolverClass, {
+      defaultNamespace: 'resolvers',
+    });
+    // Find resolver bindings that match the class
+    const bindings = this.ctx
+      .findByTag(GraphQLTags.RESOLVER)
+      .filter(filterByServiceInterface(resolverClass));
```
bajtos, line 57: It makes me wonder what is the performance cost of scanning all bindings whenever we want to instantiate a resolver class. 

I also don't understand why is this scan needed since we already have the binding key (`resolverBinding.key`) we want to resolve/instantiate? Cannot we use `ctx.get(resolverBinding.key)` directly? You can check whether the binding exists by calling `ctx.isBound(resolverBinding.key)`.

Calling `createBindingFromClass` could be a performance problem too, but I don't see any easy solution, so let's leave this part for later.

raymondfeng: `findByTag` uses tag index internally and the overhead is minimal. 

The idea here is to use existing bindings first before we add a new binding created from the class.

extensions/graphql/src/graphql.container.ts
raymondfeng: The `GraphQLResolutionContext` is only created to contain the resolver binding if it does not exist in the chain. At the moment, we're building a GraphQL schema (and LoopBack container for DI) per application.

raymondfeng:
@bajtos I extracted the example into `examples/graphql`.

bajtos (approved):
I am afraid I don't have bandwidth to keep up with the changes. I am fine to land this as an initial version and then iterate on any improvements needed.

Please review code coverage by tests, I see few files that have lower test coverage than is our current bar:

- resolver resolution - narrow down by key: https://coveralls.io/builds/33198782/source?filename=extensions%2Fgraphql%2Fsrc%2Fgraphql.container.ts#L85
- verify handling of express options: https://coveralls.io/builds/33198782/source?filename=extensions%2Fgraphql%2Fsrc%2Fgraphql.server.ts#L76
- it looks like there is no test loading the graphql example `index.ts` file: https://coveralls.io/builds/33198782/source?filename=examples%2Fgraphql%2Fsrc%2Findex.ts#L11

Issue #4754, opened 2020-02-26
defining a nested array model property using @property.array decorator crashes the app


alyhegazy:
## Steps to reproduce
define a nested array model property like the following

```
import {Model, property} from '@loopback/repository';

class testModel extends Model {
  @property.array(Array)
  nestedArr: Array<Array<string>>;
}
```

## Current Behavior
The app crashes with the error `Cannot start the application. Error: "items" property must be present if "type" is an array`

## Expected Behavior
The nested array property should be defined without problems. Would be nice to also allow defining the type of the nested array for example `@property.array(Array<string>)`.

## Additional information
darwin x64 10.19.0
├── @loopback/authentication@3.3.1
├── @loopback/authorization@0.4.9
├── @loopback/boot@1.7.2
├── @loopback/context@2.0.0
├── @loopback/core@1.12.2
├── @loopback/openapi-v3@1.12.0
├── @loopback/rest@1.26.0
└── @loopback/rest-explorer@1.4.8

## Related Issues

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

## Acceptance Criteria
- [ ] Update documentation to clarify that @property.array() decorator does not support nested arrays in model properties ( add reasons why, Ts type issues)
- [ ] Fix validation code in https://github.com/strongloop/loopback-next/blob/master/packages/repository-json-schema/src/build-schema.ts#L221 to return appropriate error when seeing a nested array.
- [ ] Add example to achieve nested arrays with json schema directly using @param() decorator

deepakrkris:
@alyhegazy ,  are you extending from another model, is `Model` a base class you have defined ?
 
can you try :

```
import {Entity, model, property} from '@loopback/repository';

@model()
class testModel extends Entity {
  @property.array(Array)
  nestedArr: Array<Array<string>>;
}
```
This seems to work for me.

deepakrkris:
@alyhegazy please ignore my previous comment, I get the same error.

alyhegazy:
@deepakrkris the problem is in the `property.array()` decorator. If you replace `@property.array(Array)` with `@property.array(Object)` for example it works fine.

deepakrkris:
@alyhegazy we are not supporting nested arrays 

https://github.com/strongloop/loopback-next/blob/master/packages/repository-json-schema/src/build-schema.ts#L221 . 

But this is not well documented as well as the validation code above misses returning an appropriate error.

[...]

PR #5559, opened 2020-05-25
fix: array decorator

jannyHou:
Fixes https://github.com/strongloop/loopback-next/issues/4754

To define a nested array property with `@property.array()`, you must provide a json schema to describe the sub array item:

```ts
@model()
class TestModel {
  // alternatively use @property.array('array')
  @property.array(Array, {
    jsonSchema: {
      type: 'array',
      items: {type: 'string'},
    },
  })
  nestedArr: Array<Array<string>>;
}
```

A few issues this PR addresses:

- We didn't allow nested array property before due to 
https://github.com/strongloop/loopback-next/blob/ef155eeddf11895bac8ddc491cb87c1e5b7398d0/packages/repository-json-schema/src/build-schema.ts#L260
   which is not correct. I tested with todo app, when decorate the nested array property well, creating & retrieving an instance works properly.

- It's not possible to infer the sub type of nested array property, that's why the json schema field is needed(Thank you @deepakrkris for your suggestion in https://github.com/strongloop/loopback-next/issues/4754#issuecomment-592117906). I added a new check for that field to avoid triggering the subsequent error '"items" property must be present if "type" is an array' instead.
 
## Checklist

### DIFF
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -851,6 +851,27 @@ class TestModel {
 }
 ```
 
+To define a nested array property, you must provide the `jsonSchema` field to
+describe the sub-array property. For example:
+
+```ts
+@model()
+class TestModel {
+  // alternatively use @property.array('array')
+  @property.array(Array, {
+    jsonSchema: {
+      type: 'array',
+      items: {type: 'string'},
+    },
+  })
+  nestedArr: Array<Array<string>>;
+}
+```
+
+If the `jsonSchema` field is missing, you will get an error saying
+
+> You must provide the "jsonSchema" field when define a nested array property'
+
 ### Validation Rules
 
 You can also specify the validation rules in the field `jsonSchema`. For
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -251,6 +251,34 @@ describe('build-schema', () => {
         expectValidJsonSchema(jsonSchema);
       });
 
+      it('properly converts nested array property when json schema provided', () => {
+        @model()
+        class TestModel {
+          // alternatively use @property.array('array')
+          @property.array(Array, {
+            jsonSchema: {
+              type: 'array',
+              items: {type: 'string'},
+            },
+          })
+          nestedArr: Array<Array<string>>;
+        }
+
+        const jsonSchema = modelToJsonSchema(TestModel);
+        expect(jsonSchema.properties).to.deepEqual({
+          nestedArr: {
+            type: 'array',
+            items: {
+              type: 'array',
+              items: {
+                type: 'string',
+              },
+            },
+          },
+        });
+        expectValidJsonSchema(jsonSchema);
+      });
+
       it('properly converts properties with enum in json schema', () => {
         enum QueryLanguage {
           JSON = 'json',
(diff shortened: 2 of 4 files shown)


docs/site/Model.md
```diff
@@ -851,6 +851,27 @@ class TestModel {
 }
 ```
 
+To define a nested array property, you must provide the `jsonSchema` field to
```
agnes512, line 854: I just realized that our docs https://loopback.io/doc/en/lb4/Model.html barely mentions `jsonSchema` and what it does and why it is related to our model definition. Do we name it `jsonSchema` because the model definition is translated to a JSON?

jannyHou: @agnes512 good catch I didn't realize there is hidden content on this page. 
But the hidden div ends at line 776, so it doesn't affect the doc of array decorator :)

agnes512: But I don't see that part under https://loopback.io/doc/en/lb4/Model.html#data-mapping-properties.

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -251,6 +251,34 @@ describe('build-schema', () => {
         expectValidJsonSchema(jsonSchema);
       });
 
+      it('properly converts nested array property when json schema provided', () => {
```
agnes512, line 254: Do we have any tests on querying/filtering the nested arrays?

jannyHou: @agnes512 `@loopback/repository-json-schema` doesn't have it. If we add tests, need to do it in the acceptance tests for connectors.

agnes512: I mentioned it is because when I was triaging https://github.com/strongloop/loopback-connector-postgresql/issues/441, I realized that not all connectors support arrays/nested arrays. It'd be good if we can have them documented and tested. Just some thoughts 😄

docs/site/Model.md
jannyHou: @agnes512 ah, fixes it in PR https://github.com/strongloop/loopback-next/pull/5568

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
jannyHou: > I realized that not all connectors support arrays/nested arrays.

Yeah that's the problem, the original issue https://github.com/strongloop/loopback-next/issues/4754 is not trying to support the full feature, it's a fix for the misuse of decorator. 

> It'd be good if we can have them documented and tested.

It's not a trivial thing to test across connectors, let's do it in a separate story.

docs/site/Model.md
```diff
@@ -851,6 +851,27 @@ class TestModel {
...
+describe the sub-array property. For example:
+
+```ts
+@model()
+class TestModel {
+  // alternatively use @property.array('array')
+  @property.array(Array, {
+    jsonSchema: {
+      type: 'array',
+      items: {type: 'string'},
```
agnes512, line 864: I have another question of the design. Do we need to specify these two fields? if so, why do we need `type: 'array'` as we already use the decorator?

jannyHou: @agnes512 the `jsonSchema` here is for the sub array property, 
e.g. For `Array<Array<string>>`, the json schema describes `Array<string>`

If the current doc confuses you:

> To define a nested array property, you must provide the `jsonSchema` field to describe the sub-array property.

Any suggestion on how to make it more clear?

agnes512: Ah I see, then I am good with it 👍

jannyHou:
@agnes512 sure 👍 created story in https://github.com/strongloop/loopback-next/issues/5570

PR #5365, opened 2020-05-07
[WIP] feat(repository): adding hasManyThrough repository and its factory

agnes512:
## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5562, opened 2020-05-26
feat(authentication): add a middleware for authentication

raymondfeng:
Move authentication from an explicit action to middleware to simplify registration.

## Checklist


docs/site/Loopback-component-authentication.md
```diff
@@ -33,7 +33,7 @@ Here is a **high level** overview of the authentication component.
 
 - A decorator to express an authentication requirement on controller methods
 - A provider to access method-level authentication metadata
-- An action in the REST sequence to enforce authentication
+- An middleware in the REST sequence to enforce authentication
```
hacksparrow, line 36: A middleware

packages/authentication/src/providers/auth-action.provider.ts
```diff
@@ -80,3 +86,23 @@ export class AuthenticateActionProvider implements Provider<AuthenticateFn> {
     }
   }
 }
+
+@bind(
+  asMiddleware({
+    chain: DEFAULT_MIDDLEWARE_CHAIN,
```
jannyHou, line 92: @raymondfeng what's the difference between `chain` and `group`?

raymondfeng: We allows multiple middleware chains to be invoked in the sequence. Within each chain, the `group` is used for ordering.

raymondfeng:
I'll mark this PR a breaking change. For existing applications that use the explicit authentication action, they need to remove the action by hand after upgrading.

dhmlau:
Add `breaking-change` label per @raymondfeng's comment https://github.com/strongloop/loopback-next/pull/5562#issuecomment-634064429.

raymondfeng:
Hmm, we have trouble here. The `authentication` needs to happen after `findRoute`, which is NOT a middleware at the moment until #5366 is accepted. Then we'll have to create another middleware chain after `findRoute` action. I think this is another reason that I proposed #5365. Mixing actions and middleware is tricky and confusing.

raymondfeng:
Closing it to favor #5366

Issue #3033, opened 2019-06-03
Epic: ENUM type (ORM, OpenAPI, etc.)


bajtos:
Implement a new property type - enum (enumeration).

Example model definition:

``` ts
@model()
class Person extends Entity {
  @property({
    type: 'string',
    enum: ['male', 'female', 'other'],
  })
  gender: string;
}
```

This issue is a replacement for https://github.com/strongloop/loopback/issues/1321 that's tracking ENUM feature request in LoopBack 3.x

Areas to cover:
- support both `string` and `number` as a backing datatype
-  loopback-datasource-juggler should validate property values of enum type inside DAO methods like `create` and `updateAttributes`
- automigration: connectors should map enum type to appropriate database counterpart during database schema migration
  - SQL connectors, see e.g. https://github.com/strongloop-internal/scrum-loopback/issues/603
  - MongoDB, see e.g. https://github.com/strongloop-internal/scrum-loopback/issues/604
  - Cloudant
  - etc.
- autodiscovery: connectors should add `enum` constraint to the discovered definitions where applicable
- JSON Schema and OpenAPI: code converting model definitions to JSON Schema and OpenAPI Schema should recognize `enum` and include the constraint in the schema produced
- `@loopback/rest` should validate ENUM constraints for input parameters, see https://github.com/strongloop/loopback-next/issues/1624
- documentation & blog post

luncht1me:
I'm currently looking for a way to validate properties on my model of a `string` type to only be one of an allowed value. This would be a great help as I'm having trouble finding the "loopback way" to do it. My current solution was to iterate over an enum with a custom interceptor... But I don't really like that approach.

11arn11:
any update about this feature?

zzhenryquezz:
+1

zzhenryquezz:
Works if you put the enum in jsonSchema object
```
@model()
class Person extends Entity {
  @property({
    type: 'string',
    jsonSchema: {
        enum: ['male', 'female', 'other'],
    }    
  })
  gender: string;
}
```

dougal83:
> Works if you put the enum in jsonSchema object

@zzhenryquezz Take care in using this though as it may not be validating input. [ See comment](https://github.com/strongloop/loopback-next/issues/2911#issuecomment-570732679).

ludohenin:
> Take care in using this though as it may not be validating input. [ See comment](https://github.com/strongloop/loopback-next/issues/2911#issuecomment-570732679).

[...]

PR #5584, opened 2020-05-27
jsonSchema composition for inheritance and includeRelations

InvictusMB:
Implement a workaround for enum duplication in a generated OpenAPI spec [mentioned here](https://github.com/strongloop/loopback-next/issues/3033#issuecomment-634097517).
As of `@loopback/repository-json-schema@2.4.2` both inheritance and `getModelSchemaRef(SomeModel, {includeRelations: true})` produce duplication in definitions. If an enum type property is present in `SomeModel` and a client is generated from such a spec (via [typescript-fetch](https://openapi-generator.tech/docs/generators/typescript-fetch) in my case), `SomeModel` and `SomeModelWithRelations` will have incompatible types in generated code. Same happens if a model inherits another one with an enum in it.
This PR replaces all inherited props with `allOf` composition and a `$ref` to a parent schema in both cases.

## Checklist

### DIFF
--- a/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
+++ b/packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
@@ -1019,27 +1019,54 @@ describe('build-schema', () => {
 
       const expectedSchema: JsonSchema = {
         definitions: {
+          Category: {
+            additionalProperties: false,
+            description:
+              '(tsType: Category, schemaOptions: { includeRelations: false })',
+            properties: {
+              id: {
+                type: 'number',
+              },
+            },
+            title: 'Category',
+          },
+          Product: {
+            title: 'Product',
+            description:
+              `(tsType: Product, ` +
+              `schemaOptions: { includeRelations: false })`,
+            properties: {
+              id: {type: 'number'},
+              categoryId: {type: 'number'},
+            },
+            additionalProperties: false,
+          },
           ProductWithRelations: {
             title: 'ProductWithRelations',
             description:
               `(tsType: ProductWithRelations, ` +
               `schemaOptions: { includeRelations: true })`,
+            allOf: [
+              {$ref: '#/definitions/Product'},
+              {
+                properties: {
+                  category: {$ref: '#/definitions/CategoryWithRelations'},
+                },
+              },
+            ],
+          },
+        },
+        allOf: [
+          {$ref: '#/definitions/Category'},
+          {
             properties: {
-              id: {type: 'number'},
-              categoryId: {type: 'number'},
-              category: {$ref: '#/definitions/CategoryWithRelations'},
+              products: {
+                type: 'array',
+                items: {$ref: '#/definitions/ProductWithRelations'},
+              },
             },
-            additionalProperties: false,
           },
-        },
-        properties: {
-          id: {type: 'number'},
-          products: {
-            type: 'array',
-            items: {$ref: '#/definitions/ProductWithRelations'},
-          },
-        },
-        additionalProperties: false,
+        ],
         title: 'CategoryWithRelations',
         description:
           `(tsType: CategoryWithRelations, ` +
@@ -1066,28 +1093,53 @@ describe('build-schema', () => {
       }
       const expectedSchema: JsonSchema = {
         definitions: {
-          ProductWithRelations: {
-            title: 'ProductWithRelations',
+          CategoryWithoutProp: {
+            additionalProperties: false,
             description:
-              `(tsType: ProductWithRelations, ` +
-              `schemaOptions: { includeRelations: true })`,
+              '(tsType: CategoryWithoutProp, schemaOptions: { includeRelations: false })',
+            title: 'CategoryWithoutProp',
+          },
+          Product: {
+            title: 'Product',
+            description:
+              '(tsType: Product, schemaOptions: { includeRelations: false })',
             properties: {
-              id: {type: 'number'},
-              categoryId: {type: 'number'},
-              category: {
-                $ref: '#/definitions/CategoryWithoutPropWithRelations',
+              categoryId: {
+                type: 'number',
+              },
+              id: {
+                type: 'number',
               },
             },
             additionalProperties: false,
           },
-        },
-        properties: {
-          products: {
-            type: 'array',
-            items: {$ref: '#/definitions/ProductWithRelations'},
+          ProductWithRelations: {
+            allOf: [
+              {$ref: '#/definitions/Product'},
+              {
+                properties: {
+                  category: {
+                    $ref: '#/definitions/CategoryWithoutPropWithRelations',
+                  },
+                },
+              },
+            ],
+            description:
+              '(tsType: ProductWithRelations, schemaOptions: { includeRelations: true })',
+            title: 'ProductWithRelations',
           },
         },
-        additionalProperties: false,
+        allOf: [
+          {$ref: '#/definitions/CategoryWithoutProp'},
+          {
+            properties: {
+              products: {
+                type: 'array',
+                items: {$ref: '#/definitions/ProductWithRelations'},
+              },
+            },
+          },
+        ],
         title: 'CategoryWithoutPropWithRelations',
         description:
           `(tsType: CategoryWithoutPropWithRelations, ` +
--- a/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
+++ b/packages/repository-json-schema/src/__tests__/unit/build-schema.unit.ts
@@ -313,15 +313,15 @@ describe('build-schema', () => {
         title: 'ParentWithItsChildren',
         includeRelations: true,
       });
-      expect(schema.properties).to.containEql({
-        children: {
-          type: 'array',
-          // The reference here should be `ChildWithRelations`,
-          // instead of `ParentWithItsChildren`
-          items: {$ref: '#/definitions/ChildWithRelations'},
+      expect(schema.allOf).to.containEql({
+        properties: {
+          children: {
+            type: 'array',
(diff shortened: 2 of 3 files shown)


Smixi:
Hi,
I just got the same problem with enums and Typescript Axios generator for my client.
I'm kinda new to Loopback, but I was wondering if your workaround will change the usage on mentioned show in this [post](https://github.com/strongloop/loopback-next/issues/3033#issuecomment-634097517) or you plan to go for a reusable ref, so it will not only check for inheritance. 
Because I think if two different model share the same enum, then the OpenAPI spec will still have two of them and, it will also produce duplicates in the generated code for the clients (and also incompatible types).

InvictusMB:
@Smixi It won't address reusable enums yet. Also OpenAPI generators are not good at enums either.
But it is possible work around to some extent if a shared enum property is extracted into a common ancestor model. This is essentially the use case I'm aiming to enable here.

Smixi:
> @Smixi It won't address reusable enums yet. Also OpenAPI generators are not good at enums either.
> But it is possible work to around to some extent if a shared enum property is extracted into a common ancestor model. This is essentially the use case I'm aiming to enable here.

Ok thanks for the answer !
For my side your use case is the same as mine so I'm happy for your fix :) !
For now, client side, I'm just casting to targeted model, but it will be in fact a lot cleaner with what you're fixing.
I guess it will improve overtime for both OpenAPI and LB4 for handling such items

InvictusMB:
@Smixi You could already test this PR on your setup via `patch-package`. See example [here](https://github.com/InvictusMB/ram-loopback-sample/commit/b5870bab46b2d3b6d281f0c8877fce5e941c2a93). Please let me know if there is any unexpected behavior.

Smixi:
@InvictusMB It seems my LB4 version is one minor version more than yours, and the patch thus cannot be applied (2.4.2 v 2.4.3), i'll try to downgrade if they are no breaking changes I'll stay you updated.

Smixi:
@InvictusMB I just Installed the patch but unfortunately it seems to keep the inlining of the enums.

Here is a property of my model Weapon : 

export enum CategoryType {
  A = "A",
  B = "B",
  C = "C",
  D = "D",
}

```typescript
export enum CategoryType {
  A = "A",
  B = "B",
  C = "C",
  D = "D",
}

  @property({
    type: 'string',
    required: false,
    postgresql: { columnName: 'category', dataType: 'text', dataLength: null, dataPrecision: null, dataScale: null, nullable: 'YES' },
    jsonSchema: {
      type: ["string", "null"],
      enum: Object.values(CategoryType)
    }
  })
  category?: CategoryType;
```

Here is a part of the controller :

```typescript
@authenticate('jwt')
  @post('/weapons', {
    responses: {
      '200': {
        description: 'Weapon model instance',
        content: { 'application/json': { schema: getModelSchemaRef(Weapon) } },
      },
    },
  })
  async create(
    @requestBody({
      content: {
        'application/json': {
          schema: getModelSchemaRef(Weapon, {
            title: 'NewWeapon',
            exclude: ['id'],
          }),
        },
      },
    })
    weapon: Omit<Weapon, 'id'>,
  ): Promise<Weapon> {...}

  @authenticate("jwt")
  @get('/weapons/{id}', {
    responses: {
      '200': {
        description: 'Weapon model instance',
        content: {
          'application/json': {
            schema: getModelSchemaRef(Weapon, { includeRelations: true }),
          },
        },
      },
    },
  })
  async findById(
    @param.path.string('id') id: string,
    @param.query.object('filter', getFilterSchemaFor(Weapon)) filter?: Filter<Weapon>
  ): Promise<Weapon> {
    return this.weaponRepository.findById(id, filter);
  }

```
And here is the associated part of the JSON

```json
"NewWeapon":{   "title":"NewWeapon",
   "description":"(tsType: Omit<Weapon, 'id'>, schemaOptions: { title: 'NewWeapon', exclude: [ 'id' ] })",
   "properties":{      "serialnumber":{
         "type":"string"
      
},
      "model":{
         "type":"string"
      
},
      "available":{
         "type":"boolean"
      
},
      "description":{
         "type":"string"
      
},
      "deleted":{
         "type":"boolean"
      
},
      "handtype":{         "type":"string",
         "enum":[
            "rightHanded",
            "leftHanded",
            "ambidextrous"
         
]
      
},
      "weaponType":{         "type":"string",
         "enum":[
            "rifle",
            "pistol",
            "cannon",
            "carbine",
            "crossbow"
         
]
      
},
      "caliber":{         "type":"string",
         "enum":[
            "4.5mm",
            "7.5mm",
            ".22LR",
            ".38",
            "arrow",
            "lever"
         
]
      
},
      "category":{         "type":"string",
         "enum":[
            "A",
            "B",
            "C",
            "D"
         
]
      
},
      "price":{
         "type":"string"
      
},
      "clubNumber":{
         "type":"string"
      
}
   
},
   "required":[
      "serialnumber"
   
],
   "additionalProperties":false,
   "x-typescript-type":"Omit<Weapon, 'id'>"
},
"WeaponPartial":{   "title":"WeaponPartial",
   "description":"(tsType: Partial<Weapon>, schemaOptions: { partial: true })",
   "properties":{      "id":{
         "type":"string"
      
},
      "serialnumber":{
         "type":"string"
      
},
      "model":{
         "type":"string"
      
},
      "available":{
         "type":"boolean"
      
},
      "description":{
         "type":"string"
      
},
      "deleted":{
         "type":"boolean"
      
},
      "handtype":{         "type":"string",
         "enum":[
            "rightHanded",
            "leftHanded",
            "ambidextrous"
         
]
      
},
      "weaponType":{         "type":"string",
         "enum":[
            "rifle",
            "pistol",
            "cannon",
            "carbine",
            "crossbow"
         
]
      
},
      "caliber":{         "type":"string",
         "enum":[
            "4.5mm",
            "7.5mm",
            ".22LR",
            ".38",
            "arrow",
            "lever"
         
]
      
},
      "category":{         "type":"string",
         "enum":[
            "A",
            "B",
            "C",
            "D"
         
]
      
},
      "price":{
         "type":"string"
      
},
      "clubNumber":{
         "type":"string"
      
}
   
},
   "additionalProperties":false,
   "x-typescript-type":"Partial<Weapon>"
}
```

Am I doing something wrong ?
If you need more info don't hesitate :) !

InvictusMB:
This is a slightly different use case. 
You have `schemaOptions: { partial: true }` and `schemaOptions: { exclude: [ 'id' ] }`. This requires a more complicated logic.

I've only added special handling for `schemaOptions:  {includeRelations: true}` so far.

Smixi:
@InvictusMB Ok I see ! Thus Weapon and WeaponWithRelation should have the same enum, I'll check this. Otherwise sorry, I got in fact confused with the use case fixed by your PR. Thanks you :)

packages/repository-json-schema/src/__tests__/integration/build-schema.integration.ts
```diff
@@ -1019,27 +1019,54 @@ describe('build-schema', () => {
 
       const expectedSchema: JsonSchema = {
         definitions: {
-          ProductWithRelations: {
-            title: 'ProductWithRelations',
+          Category: {
```
bajtos, line 1022: I find it very difficult to review this part of the changeset. Is there a way how to preserve the old order of `definitions` entries, to keep the amount of changed lines as small as possible please?

InvictusMB: @bajtos I don't know how to proceed with this PR tbh.
The original goal was to eliminate all duplication that blows up taxonomy in OpenAPI definitions. The reason is that each model creates 4+ definitions (`Thing`, `ThingPartial`, `ThingWithRelations`, `NewThing`, etc.) and this causes issues in the client-side generated code as TypeScript does not consider all those incarnations of the same model compatible. 

This code seems to work fine on my projects but I've observed more cases where extra definitions are created. Other variations on `JsonSchemaOptions` (e.g. excluding properties) create new definitions. I'm not sure if this code is future-proof. A proper fix would be to construct an inheritance chain/tree in terms of OpenAPI schema but I don't have enough experience with lb4 to account for all scenarios. 
 
Also I'm not sure if the `isInherited` check and `getParentModel` implemented here are sufficient.

My suggestion for moving this forward would be to expose `modelToJsonSchema` to userland customization/override instead. So that we can explore the advanced scenarios and come up with patterns that work before writing them in stone.

bajtos:
>  A proper fix would be to construct an inheritance chain/tree in terms of OpenAPI schema but I don't have enough experience with lb4 to account for all scenarios.

I like the idea of using inheritance for different variants of model schemas. I think Thing, ThingWithRelations and NewThing may be reasonably easy to implement (`NewThing` -> `Thing` -> `ThingWithRelations`). I am not sure if it's possible to include `ThingPartial` in the inheritance chain though.

> My suggestion for moving this forward would be to expose `modelToJsonSchema` to userland customization/override instead. So that we can explore the advanced scenarios and come up with patterns that work before writing them in stone.

I think that's the best approach to get started 👍🏻 . Would you like to open a new pull request for that?

Another idea to consider: https://github.com/strongloop/loopback-next/pull/4365 landed a feature that consolidates OpenAPI schema to move repeated in-place schema definitions to shared `#/components/schemas`. Can we perhaps enhance the consolidator to also rework `NewThing` -> `Thing` -> `ThingWithRelations` to use inheritance?

bajtos:
> Can we perhaps enhance the consolidator to also rework `NewThing` -> `Thing` -> `ThingWithRelations` to use inheritance?

On the second thought, I think it's better to implement this feature as a new OASEnhancer that should be executed after `ConsolidationEnhancer` finished its work.

dhmlau:

```
git commit --amend -s
git push --force-with-lease
```

PR #5585, opened 2020-05-27
set default HTTP status code for authorization error to 403

### LINKED ISSUE #5411: Authorize interceptor should return 403 HTTP status
Authentication component returns 401 errors, authorization component should return 403 errors.
See [403 status definition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403).

Offending lines:
https://github.com/strongloop/loopback-next/blob/b01bb1154346f8dd2786a13e21e291dc52ca35e1/packages/authorization/src/authorize-interceptor.ts#L106

https://github.com/strongloop/loopback-next/blob/b01bb1154346f8dd2786a13e21e291dc52ca35e1/packages/authorization/src/authorize-interceptor.ts#L121

InvictusMB:
Fixes #5411 according to the [suggested solution](https://github.com/strongloop/loopback-next/issues/5411#issuecomment-628065626)

## Checklist

### DIFF
--- a/packages/authorization/src/authorize-interceptor.ts
+++ b/packages/authorization/src/authorize-interceptor.ts
@@ -42,6 +42,7 @@ export class AuthorizationInterceptor implements Provider<Interceptor> {
     this.options = {
       defaultDecision: AuthorizationDecision.DENY,
       precedence: AuthorizationDecision.DENY,
+      defaultStatusCodeForDeny: 403,
       ...options,
     };
     debug('Authorization options', this.options);
@@ -107,7 +108,7 @@ export class AuthorizationInterceptor implements Provider<Interceptor> {
       ) {
         debug('Access denied');
         const error = new AuthorizationError('Access denied');
-        error.statusCode = 401;
+        error.statusCode = this.options.defaultStatusCodeForDeny;
         throw error;
       }
       if (
@@ -122,7 +123,7 @@ export class AuthorizationInterceptor implements Provider<Interceptor> {
     // Handle the final decision
     if (finalDecision === AuthorizationDecision.DENY) {
       const error = new AuthorizationError('Access denied');
-      error.statusCode = 401;
+      error.statusCode = this.options.defaultStatusCodeForDeny;
       throw error;
     }
     return next();
--- a/packages/authorization/src/types.ts
+++ b/packages/authorization/src/types.ts
@@ -187,4 +187,9 @@ export interface AuthorizationOptions {
    * not associated with authorization metadata.
    */
   defaultMetadata?: AuthorizationMetadata;
+  /**
+   * Default HTTP status code when the final decision is `AuthorizationDecision.DENY`.
+   * If not set, default to 403
+   */
+  defaultStatusCodeForDeny?: number;
 }
--- a/examples/access-control-migration/src/__tests__/acceptance/project.acceptance.ts
+++ b/examples/access-control-migration/src/__tests__/acceptance/project.acceptance.ts
@@ -111,7 +111,8 @@ describe('AccessControlApplication - permissions', () => {
     });
 
     it(`view returns ${permissions.view} for role ${role}`, async () => {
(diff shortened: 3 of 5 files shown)


packages/authorization/src/types.ts
```diff
@@ -187,4 +187,9 @@ export interface AuthorizationOptions {
    * not associated with authorization metadata.
    */
   defaultMetadata?: AuthorizationMetadata;
+  /**
+   * Default HTTP status code when the final decision is `AuthorizationDecision.DENY`.
+   * If not set, default to 403
+   */
+  defaultStatusCodeForDeny?: number;
```
raymondfeng, line 194: Do we want to make it even more stricter, such as `401 | 403`?

InvictusMB: I wouldn't be that strict. I can imagine a valid scenario for returning a `404`. For example, a system might have `/users/:id` returning `404` to prevent malicious agents from mining existing user ids.
Also some systems occasionally implement their own extensions in `4xx` space for whatever reasons they have.

I am happy with anything other than `401` as `401` usually triggers a token refresh and it is important to distinguish it from non-retriable failures to avoid loops.

raymondfeng:
@InvictusMB Please run `npm run prettier:fix` to fix formatting issues.

packages/authorization/src/authorize-interceptor.ts
```diff
@@ -42,6 +42,7 @@ export class AuthorizationInterceptor implements Provider<Interceptor> {
     this.options = {
       defaultDecision: AuthorizationDecision.DENY,
       precedence: AuthorizationDecision.DENY,
+      defaultStatusCodeForDeny: 403,
```
raymondfeng, line 45: This is a breaking change. Please add the following to the commit log:

```
feat(authorization): set default HTTP status code for authorization error to 403

BREAKING CHANGE: We now use http status code `403` instead of `401` to report authorization denied. The default status code can be set via `defaultStatusCodeForDeny` for the authorization options.
```

dougal83:
As long as the result can be customised or looks good. Personally I'd use 403 for when scopes are invalid and it is known that access is forbidden. Otherwise 401 is generally a better fit.

InvictusMB:
> As long as the result can be customised it looks good. Personally I'd use 403 for when scopes are invalid and it is known that access is forbidden. Otherwise 401 is generally a better fit.

The way I read [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.3) and [RFC 7235](https://tools.ietf.org/html/rfc7235#section-4.1) is:
* `403` for a generic authenticated but not authorized situation
* `401` for missing or bad credentials, when a server explicitly tells what is missing
* `404` when a resource should be hidden

A typical client implementation works like this:
* `403` -> error out or refresh token once (if stale credentials are suspected), retry and error out
* `401` -> log out or refresh token if not expired
Returning 401 indiscriminately for authorization denial leaves a client with 2 options: a forceful logout or an infinite loop of token refreshing. Because a client can no longer tell a difference between invalid credentials for the whole app and invalid credentials for a specific resource.

dougal83:
> * Because a client can no longer tell a difference between invalid credentials for the whole app and invalid credentials for a specific resource.

Do you mean if solely using the status code?  Just confirming for understanding or I'm missing something.

InvictusMB:
> Do you mean if solely using the status code? Just confirming for understanding or I'm missing something.

Yes, that's the assumption behind semantic HTTP status codes. Otherwise one could always return 400 and parse out error messages.

dougal83:
> > Do you mean if solely using the status code? Just confirming for understanding or I'm missing something.
> 
> Yes, that's the assumption behind semantic HTTP status codes. Otherwise one could always return 400 and parse out error messages.

Disagree. Status codes are not enough on their own in all circumstances. Messages exist for a reason, to give more specific information and this is more relevant especially when dealing with restful APIs. One status code can encompass more than one circumstance. Ability to customise on this PR +1.

PR #5586, opened 2020-05-27
feat(context): improve binding resolution error reporting with more contextual info

raymondfeng:
Current issue: when a binding key is not bound, a long stack trace is printed and it's not easy to find out where the failure happens. This PR introduces `ResolutionError` to capture more contextual information.

TBD:
I was planning to extend the `message` with such information, but it breaks a lot of existing test assertions, for example:

```
AssertionError: expected [Promise] to be rejected with a message matching 'bar: error', but got 'bar: error (@TestClass.prototype.bar => cb8e48c2-8d0e-4d28-bd93-0738840afdd9#bar)'
```
Should we bite the bullet to fix old assertions?

## Checklist

### DIFF
--- a/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
+++ b/packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
@@ -51,7 +51,15 @@ describe('Context bindings - Injecting dependencies of classes', () => {
         throw new Error('ctx.get() should have failed');
       },
       function onError(err) {
-        expect(err).to.match(/resolve.*InfoController.*argument 1/);
+        expect(err.message).to.match(
+          /The argument 'InfoController\.constructor\[0\]' is not decorated for dependency injection/,
+        );
+        expect(err.message).to.match(
+          /but no value was supplied by the caller\. Did you forget to apply @inject\(\) to the argument\?/,
+        );
+        expect(err.message).to.match(
+          /\(context: [\w\-]+, resolutionPath: controllers\.info\)/,
+        );
       },
     );
   });
@@ -72,7 +80,15 @@ describe('Context bindings - Injecting dependencies of classes', () => {
         throw new Error('ctx.get() should have failed');
       },
       function onError(err) {
-        expect(err).to.match(/resolve.*InfoController.*argument 1/);
+        expect(err.message).to.match(
+          /The argument 'InfoController\.constructor\[0\]' is not decorated for dependency injection/,
+        );
+        expect(err.message).to.match(
+          /but no value was supplied by the caller\. Did you forget to apply @inject\(\) to the argument\?/,
+        );
+        expect(err.message).to.match(
+          /\(context: [\w\-]+, resolutionPath: controllers\.info\)/,
+        );
       },
     );
   });
--- a/packages/context/src/context.ts
+++ b/packages/context/src/context.ts
@@ -27,6 +27,7 @@ import {JSONObject} from './json-types';
 import {ContextBindings} from './keys';
 import {
   asResolutionOptions,
+  ResolutionError,
   ResolutionOptions,
   ResolutionOptionsOrSession,
   ResolutionSession,
@@ -37,6 +38,7 @@ import {
   Constructor,
   getDeepProperty,
   isPromiseLike,
+  transformValueOrPromise,
   ValueOrPromise,
 } from './value-promise';
 
@@ -797,21 +799,27 @@ export class Context extends EventEmitter {
   ): ValueOrPromise<ValueType | undefined> {
     const {key, propertyPath} = BindingKey.parseKeyWithPath(keyWithPath);
 
-    optionsOrSession = asResolutionOptions(optionsOrSession);
+    const options = asResolutionOptions(optionsOrSession);
 
-    const binding = this.getBinding<ValueType>(key, optionsOrSession);
-    if (binding == null) return undefined;
-
-    const boundValue = binding.getValue(this, optionsOrSession);
-    if (propertyPath === undefined || propertyPath === '') {
-      return boundValue;
+    const binding = this.getBinding<ValueType>(key, {optional: true});
+    if (binding == null) {
+      if (options.optional) return undefined;
+      throw new ResolutionError(
+        `The key '${key}' is not bound to any value in context ${this.name}`,
+        {
+          context: this,
+          binding: Binding.bind(key),
+          options,
+        },
+      );
     }
 
-    if (isPromiseLike(boundValue)) {
-      return boundValue.then(v => getDeepProperty<ValueType>(v, propertyPath));
-    }
-
-    return getDeepProperty<ValueType>(boundValue, propertyPath);
+    const boundValue = binding.getValue(this, options);
+    return propertyPath == null || propertyPath === ''
+      ? boundValue
+      : transformValueOrPromise(boundValue, v =>
+          getDeepProperty<ValueType>(v, propertyPath),
+        );
   }
 
   /**
--- a/packages/context/src/inject.ts
+++ b/packages/context/src/inject.ts
@@ -133,6 +133,11 @@ export function inject(
   if (injectionMetadata.bindingComparator && !resolve) {
     throw new Error('Binding comparator is only allowed with a binding filter');
   }
+  if (!bindingSelector && typeof resolve !== 'function') {
+    throw new Error(
+      'A non-empty binding selector or resolve function is required for @inject',
+    );
+  }
   return function markParameterOrPropertyAsInjected(
     target: Object,
     member: string | undefined,
--- a/packages/context/src/resolution-session.ts
+++ b/packages/context/src/resolution-session.ts
@@ -394,3 +394,51 @@ export interface ResolutionContext<T = unknown> {
    */
   readonly options: ResolutionOptions;
 }
+
+/**
+ * Error for context binding resolutions and dependency injections
+ */
+export class ResolutionError extends Error {
+  constructor(
+    message: string,
+    readonly resolutionCtx: Partial<ResolutionContext>,
+  ) {
+    super(ResolutionError.buildMessage(message, resolutionCtx));
+    this.name = ResolutionError.name;
+  }
+
+  private static buildDetails(resolutionCtx: Partial<ResolutionContext>) {
+    return {
+      context: resolutionCtx.context?.name ?? '',
+      binding: resolutionCtx.binding?.key ?? '',
+      resolutionPath: resolutionCtx.options?.session?.getResolutionPath() ?? '',
+    };
+  }
+
+  /**
+   * Build the error message for the resolution to include more contextual data
+   * @param reason - Cause of the error
+   * @param resolutionCtx - Resolution context
+   */
+  private static buildMessage(
+    reason: string,
+    resolutionCtx: Partial<ResolutionContext>,
+  ) {
+    const info = this.describeResolutionContext(resolutionCtx);
+    const message = `${reason} (${info})`;
+    return message;
+  }
+
+  private static describeResolutionContext(
+    resolutionCtx: Partial<ResolutionContext>,
+  ) {
+    const details = ResolutionError.buildDetails(resolutionCtx);
+    const items: string[] = [];
+    for (const [name, val] of Object.entries(details)) {
+      if (val !== '') {
+        items.push(`${name}: ${val}`);
+      }
+    }
+    return items.join(', ');
+  }
+}
--- a/packages/context/src/resolver.ts
+++ b/packages/context/src/resolver.ts
@@ -15,7 +15,11 @@ import {
   describeInjectedProperties,
   Injection,
 } from './inject';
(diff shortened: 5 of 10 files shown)


packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
```diff
@@ -51,7 +51,11 @@ describe('Context bindings - Injecting dependencies of classes', () => {
         throw new Error('ctx.get() should have failed');
       },
       function onError(err) {
-        expect(err).to.match(/resolve.*InfoController.*argument 1/);
+        expect(err.message).to.eql(
+          'Cannot resolve injected arguments for InfoController.[0]: ' +
+            'The arguments[0] is not decorated for dependency injection, ' +
+            'but a value is not supplied',
```
bajtos, line 57: I find this error message a bit difficult to comprehend. How about this?

```
The arguments[0] is not decorated for dependency injection and the caller did not supply any value.
```

The parts "the arguments[0]" and "InfoController.[0]" look a bit weird too, can we make it easier to understand too? Assuming they are both referring to the first constructor argument:

```
Cannot resolve injected arguments for InfoController constructor:
The first argument is not decorated for dependency injection
and the caller did not supply any value for this argument.
```

(Positional numbers to use: `first`, `second`, `third`, `4th`, `5th` etc. )

packages/context/src/context.ts
```diff
@@ -797,21 +799,27 @@ export class Context extends EventEmitter {
...
-    const binding = this.getBinding<ValueType>(key, optionsOrSession);
-    if (binding == null) return undefined;
+    const options = asResolutionOptions(optionsOrSession);
+
+    let binding: Binding<ValueType> | undefined;
+    try {
+      binding = this.getBinding<ValueType>(key, options);
+      if (binding == null) return undefined;
+    } catch (err) {
+      throw new ResolutionError(err, {
```
bajtos, line 809: I find this approach problematic because it hides the part of the original stack trace that's pointing into `getBinding` internals and possibly into user-provided resolver.

Consider the following code:

```js
app.bind('foo').toDynamicValue(function getFoo() {
  throw new Error('something happened');
});
```

When trying to resolve `foo`, we should preserve the stack frames pointing to `getFoo`.

packages/context/src/context.ts
```diff
@@ -797,21 +799,27 @@ export class Context extends EventEmitter {
...
     }
-
-    if (isPromiseLike(boundValue)) {
-      return boundValue.then(v => getDeepProperty<ValueType>(v, propertyPath));
-    }
-
-    return getDeepProperty<ValueType>(boundValue, propertyPath);
+    return transformValueOrPromise(boundValue, v =>
+      getDeepProperty<ValueType>(v, propertyPath),
+    );
```
bajtos, line 822: Nice clean up, would you mind extracting it into a standalone PR to get it landed sooner + make future code archaeology easier?

bajtos:
> TBD:
> I was planning to extend the `message` with such information, but it breaks a lot of existing test assertions, for example:
> 
> ```
> AssertionError: expected [Promise] to be rejected with a message matching 'bar: error', but got 'bar: error (@TestClass.prototype.bar => cb8e48c2-8d0e-4d28-bd93-0738840afdd9#bar)'
> ```
> 
> Should we bite the bullet to fix old assertions?

IMO, those tests were written in a brittle way, relying too much on the particular implementation of error handling. It's a tech debt that we have to pay now.

Instead of replacing `'bar: error'` to `'bar: error (@TestClass.prototype.bar => cb8e48c2-8d0e-4d28-bd93-0738840afdd9#bar)'`, it's better to rewrite the assertion in such way that it allows both the old and the new style (while rejecting messages that don't contain the important information we are looking for in the error message).

For example:

```diff
- expect(err.message).to.equal('bar: error');
+ expect(err.message).to.match(/bar: error/);
```

packages/context/src/context.ts
raymondfeng: Please note that I intentionally only catch the error thrown from `getBinding()` (not including the error from `getValue()`.

packages/context/src/context.ts
raymondfeng: See #5329

packages/context/src/__tests__/acceptance/class-level-bindings.acceptance.ts
raymondfeng: I'll defer this fix to another PR as the latest version of my PR does not impact this code.

raymondfeng:
@bajtos PTAL.

packages/context/src/context.ts
```diff
@@ -797,21 +799,27 @@ export class Context extends EventEmitter {
...
-    const binding = this.getBinding<ValueType>(key, optionsOrSession);
-    if (binding == null) return undefined;
+    const options = asResolutionOptions(optionsOrSession);
+
+    let binding: Binding<ValueType> | undefined;
+    try {
+      binding = this.getBinding<ValueType>(key, options);
+      if (binding == null) return undefined;
+    } catch (err) {
+      throwResolutionError(err, {
```
bajtos, line 809: I still don't understand why we have to throw an error (deep) inside of `getBinding` and then catch & process it here. 

IIUC, there are only two situations where `getBinding` throws an error:
1. The binding key is invalid (it contains a property separator). https://github.com/strongloop/loopback-next/blob/e5e3d193a182b364a0877dd223be7a0d936a91cd/packages/context/src/binding-key.ts#L73-L83
2. The binding key was not found and `options.optional` is not set to `true`.

If my assessment is correct, then your proposal is removing just the one stack frame for `this.getBinding`, is it really that much of an improvement?

Setting the question aside, I think the implementation would be much simpler if this code invoked `getBinding` in a non-throwing mode and dealt with error conditions explicitly. 

Personally, I'd introduce `findBinding` that returns `undefined` when the binding was not found.

```ts
class Context {
  // ...
  getBinding<ValueType>(
    key: BindingAddress<ValueType>,
    options?: {optional?: boolean},
  ): Binding<ValueType> | undefined {
    const binding = this.findBinding(key, options);
    if (!binding && !options?.optional) {
      throw new Error(
      `The key '${key}' is not bound to any value in context ${this.name}`,
      );
    }
    return binding;
  }

  findBinding<ValueType>(
    key: BindingAddress<ValueType>,
    options?: {optional?: boolean},
  ): Binding<ValueType> | undefined {
    // the original implementation of `getBinding` goes here,
    // minus the part throwing an error
  }
}
```

Then you can rework the code in `getValueOrPromise` as follows:

```ts
    const binding = this.findBinding<ValueType>(key, options);
    if (!binding && !options?.optional) {
       throwResolutionError(/*...*/);
    }
```

Maybe that's not really needed and we just need to rework `getBinding` to throw a ResolutionError instead of plain Error?

packages/context/src/context.ts
bajtos: I am afraid I don't see how #5329 is related to his change? Did you perhaps make a typo in the issue/PR number?

bajtos:
Let's try to find a way how to avoid error rewrapping and throw ResolutionError in places where a resolution problem is detected.

packages/context/src/context.ts
raymondfeng: Sorry, it's #5593

raymondfeng:
@bajtos Thank you for the feedback. I have simplified `ResolutionError` for the 1st use case here - providing more information for `binding not found`.

The next one will be capturing errors from the value resolution and see how to expose more context in another PR.

packages/context/src/context.ts
```diff
@@ -797,21 +799,27 @@ export class Context extends EventEmitter {
...
-    if (binding == null) return undefined;
-
-    const boundValue = binding.getValue(this, optionsOrSession);
-    if (propertyPath === undefined || propertyPath === '') {
-      return boundValue;
-    }
-
-    if (isPromiseLike(boundValue)) {
-      return boundValue.then(v => getDeepProperty<ValueType>(v, propertyPath));
+    const binding = this.getBinding<ValueType>(key, {optional: true});
```
bajtos, line 804: Before this change, we were passing `optionsOrSession` to `ctx.getBinding`. With your change in place, the user-provided options are discarded. Are you sure that `getBinding` is never accessing options? What if a LB consumer overwrites `getBinding` in their application to tweak the way how bindings are resolved, perhaps using the `options` flags?

IMO, we need to preserve options here. Perhaps this would work?

```ts
const binding = this.getBinding<ValueType>(key, {...options, optional: true});
```

packages/context/src/resolver.ts
```diff
@@ -230,10 +234,10 @@ export function resolveInjectedArguments(
         return extraArgs[nonInjectedIndex++];
       } else {
         const name = getTargetName(target, method, ix);
-        throw new Error(
-          `Cannot resolve injected arguments for ${name}: ` +
-            `The arguments[${ix}] is not decorated for dependency injection, ` +
-            `but a value is not supplied`,
+        throw new ResolutionError(
+          `The value is not supplied or injected for argument '${name}'.` +
+            ` Please check if @inject is missing.`,
```
bajtos, line 239: Personally, I find the new messaging as making it more difficult to understand what's wrong. However, I do like the hint on how to fix the problem.

I am proposing the following:

```suggestion
        throw new ResolutionError(
          `The argument '${name}' is not decorated for dependency injection ` +
            'and no value was supplied by the caller. Did you forget to call '
            '@inject() decorator to configure dependency injection?'
```

packages/context/src/resolver.ts
```diff
@@ -273,9 +277,9 @@ export function resolveInjectedProperties(
   return resolveMap(injectedProperties, (injection, p) => {
     if (!injection.bindingSelector && !injection.resolve) {
       const name = getTargetName(constructor, p);
-      throw new Error(
-        `Cannot resolve injected property ${name}: ` +
-          `The property ${p} is not decorated for dependency injection.`,
+      throw new ResolutionError(
+        `The binding selector or resolve function is missing for injection of property '${name}'`,
```
bajtos, line 281: I have no idea what does it mean that a binding selector is missing or a resolve function is missing. The terms "bindingSelector" and "resolve" are speaking from the implementation details, which most LB4 users don't understand. (I don't know what do they mean either without reading the surrounding code here.)

Let's find a way how to describe this error using the terms that LB users will understand.

I think we should differentiate between the situation when the user forgot to apply `@inject()` and the situation where a decorator was applied but did not provide enough information to resolve the injected value.

Example error message for the first situation:

> The property '${name}' is not decorated for dependency injection. Did you forget to call @inject() decorator to configure dependency injection?

(Can this actually happen? Do we require any properties to be decorated for DI?)

In the second case, I think it's usually a problem in the inject-like decorator implementation that the user cannot do much about, is that right? If yes, then the message should make that clear.

For example:

> The property '${name}' is decorated for dependency injection, but the decorator did not provide enough information to resolve the value to be injected. Either `bindingSelector` or `resolve` is required.

packages/context/src/context.ts
raymondfeng: Not really, the signature of `getBinding` is as follows:
```ts
getBinding<ValueType>(
    key: BindingAddress<ValueType>,
    options?: {optional?: boolean},
  ): Binding<ValueType> | undefined;
```

As you see, the type of `options` is `{optional?: boolean}`.

packages/context/src/resolver.ts
raymondfeng: The only case for this error is:

`@inject('', { /* no resolve in options */}`

I think it's better to detect such problem at `@inject` decorator level.

raymondfeng: I improved the code to throw when `@inject` is called.

packages/context/src/resolution-session.ts
```diff
@@ -394,3 +394,51 @@ export interface ResolutionContext<T = unknown> {
...
+  ) {
+    super(ResolutionError.buildMessage(message, resolutionCtx));
+    this.name = ResolutionError.name;
+  }
+
+  private static buildDetails(resolutionCtx: Partial<ResolutionContext>) {
+    return {
+      context: resolutionCtx.context?.name ?? '',
+      binding: resolutionCtx.binding?.key ?? '',
+      resolutionPath: resolutionCtx.options?.session?.getResolutionPath() ?? '',
```
bajtos, line 414: Is there any particular reason for using `''` instead of `undefined`? By using `undefined`, we would get nicer behavior in JSON serialization too.

packages/context/src/inject.ts
```diff
@@ -133,6 +133,11 @@ export function inject(
   if (injectionMetadata.bindingComparator && !resolve) {
     throw new Error('Binding comparator is only allowed with a binding filter');
   }
+  if (!bindingSelector && typeof resolve !== 'function') {
+    throw new Error(
```
jannyHou, line 137: is it why we still throw Error here?
> I was planning to extend the message with such information, but it breaks a lot of existing test assertions

raymondfeng: This is fine as the error will be thrown when the class is loaded. There is no resolution context involved.

Issue #5565, opened 2020-05-26
Support repository mixins when scaffolding repositories and rest-crud APIs


bajtos:
## Suggestion

At the moment, we allow sharing of repository code via Repository Base Classes only. The command `lb4 repository` looks for `src/repositories/*.repository.base.ts` and treat them as repository base classes to offer in the prompt:

```
$ lb4 repository
? Please select the datasource DbDatasource
? Select the model(s) you want to generate a repository for Product
? Please select the repository base class (Use arrow keys)
  DefaultCrudRepository (Legacy juggler bridge)
  ----- Custom Repositories -----
❯ AuditingRepository
```

While easy to use, Repository Base Classes have few shortcomings too.

1. JavaScript does not support multiple inheritance, thus it's not possible to combine behavior from multiple repository base classes in the same model-specific repository class.

2. Inheritance-based reuse is considered as an anti-pattern in Object Oriented Design, it's recommended to use composition instead ("prefer composition over inheritance").

Let's enhance `lb4 repository` and `lb4 rest-crud` commands to all allow the user to choose which Repository Mixins to apply.

### Model-specific repository class:

```
$ lb4 repository
? Please select the datasource DbDatasource
? Select the model(s) you want to generate a repository for Product
? Please select the repository base class DefaultCrudRepository (Legacy juggler bridge)
? Select the mixin(s) you want to apply (Press <space> to select, <a> to toggle all, <i> to invert 
selection)
❯◉ AuditingMixin (src/mixins/auditing.repository-mixin.ts)
 ◯ TimeStampMixin (src/mixins/timestamp.repository-mixin.ts)
```

Emitted code:

```ts
// src/repositories/product.repository.ts
import {Constructor, inject} from '@loopback/core';
import {DefaultCrudRepository} from '@loopback/repository';
import {DbDataSource} from '../datasources';
import {AuditingRepositoryMixin} from '../mixins/auditing.repository-mixin';
import {Product, ProductRelations} from '../models';

export class ProductRepository extends AuditingRepositoryMixin<
  Product,
  Constructor<
    DefaultCrudRepository<
      Product,
      typeof Product.prototype.id,
      ProductRelations
    >
  >
>(DefaultCrudRepository) {
  constructor(@inject('datasources.db') dataSource: DbDataSource) {
    super(Product, dataSource);
  }
}
```

### Model API booter (rest-crud)

```
$ lb4 rest-crud
? Please select the datasource DbDatasource
[...]

PR #5591, opened 2020-05-28
docs: dynamic models, repositories, and controllers

hacksparrow:
Addresses https://github.com/strongloop/loopback-next/issues/4296.

How to dynamically add models, repositories, and controllers during
runtime.

## Checklist

### DIFF
--- a/docs/site/Dynamic-models-repositories-controllers.md
+++ b/docs/site/Dynamic-models-repositories-controllers.md
@@ -1,0 +1,259 @@
+---
+lang: en
+title: 'Dynamically adding models, repositories, and controllers during runtime'
+keywords: LoopBack 4.0, LoopBack 4, model, repository, controller, tutorial
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Dynamic-models-repositories-controllers-tutorial.html
+---
+
+## How to dynamically add models, repositories, and controllers during runtime
+
+Typically, a LoopBack app's models, repositories, and controllers are defined as
+static files in the project directory. However, there may be scenarios where all
+or some of these might have to be created after the app starts, during runtime.
+
+LoopBack provides the ability to dynamically create models, repositories, and
+controllers during runtime. This document is an explanation of how to go about
+doing it.
+
+### Defining a ModelDefinition object
+
+A `ModelDefinition` object is the first step to creating a LoopBack model, it is
+an abstraction for specifying the various attributes of a LoopBack model. A
+`ModelDefinition` object is instantiated by passing a name or a
+[ModelDefinitionSyntax](./apidocs/repository.modeldefinitionsyntax.html#modeldefinitionsyntax-interface)
+object describing the model's attributes to the `ModelDefinition`'s constructor.
+
+An example of creating a `ModelDefinition` object:
+
+```ts
+const bookDef = new ModelDefinition({
+  name: 'book',
+  properties: {
+    id: {
+      type: 'Number',
+      required: true,
+      length: null,
+      precision: 10,
+      scale: 0,
+      id: 1,
+      mysql: {
+        columnName: 'id',
+        dataType: 'int',
+        dataLength: null,
+        dataPrecision: 10,
+        dataScale: 0,
+        nullable: 'N',
+      },
+    },
+    title: {
+      type: 'String',
+      required: false,
+      length: 512,
+      precision: null,
+      scale: null,
+      mysql: {
+        columnName: 'title',
+        dataType: 'varchar',
+        dataLength: 512,
+        dataPrecision: null,
+        dataScale: null,
+        nullable: 'Y',
+      },
+    },
+  },
+  settings: {
+    idInjection: false,
+    mysql: {schema: 'test', table: 'BookStore'},
+  },
+});
+```
+
+### Defining a Model
+
+A LoopBack model class is created by passing a `ModelDefinition` object to
+`@loopback/repository`'s helper function `defineModelClass()`. It expects a base
+model to extend (typically `Model` or `Entity`), folowed by the model definition
+object. In this case it will be `Entity`.
+
+```ts
+const BookModel = defineModelClass<typeof Entity, {id: number; title?: string}>(
+  Entity,
+  bookDef,
+);
+```
+
+In case you need to use an existing Model as the base class, specify the Model
+as the base class instead of `Entity`.
+
+```ts
+import DynamicModelCtor from '@loopback/repository';
+// Assuming User is a pre-existing Model class in the app
+const StudentModel = defineModelClass<
+  typeof User,
+  {id: number; university?: string}
+>(User, studentDef);
+```
+
+For details about `ModelDefinition`, `defineModelClass`, `Model`, `Entity`, and
+`DynamicModelCtor` refer to the
+[@loopback/repository API documentation](./apidocs/repository.html).
+
+### Defining a Datasource
+
+Before a repository for this model can be set up, a datasource should be ready.
+If required, set one up dynamically by creating an instance of
+`juggler.DataSource`. It requires a name for the datasource, the connector, and
+a connection url as shown below.
+
+```ts
+const dsName = 'bookstore-ds';
+const bookDs = new juggler.DataSource({
+  name: dsName,
+  connector: require('loopback-connector-mongodb'),
+  url: 'mongodb://sysop:moon@localhost',
+});
+await bookDs.connect();
+app.dataSource(bookDs, dsName);
+```
+
+To use the newly created datasource, call its `.connect()` method and attach it
+to the app using `app.dataSource()` method. Note, this method will be available
+only on `RepositoryMixin` apps.
+
+### Defining a Repository
+
+Now that a datasource is ready, a repository can be created for `BookModel`.
+
+The `@loopback/rest-crud` package exports a helper function,
+`defineCrudRepositoryClass`, this function creates a
+`DefaultCrudRepository`-based repository class for a given model.
+
+```ts
+const BookRepository = defineCrudRepositoryClass<
+  Book,
+  typeof Book.prototype.id,
+  BookRelations
+>(BookModel);
+```
+
+In case you want to use a non-`DefaultCrudRepository` repository class or you
+want to create a custom repository, use the `defineRepositoryClass()` helper
+function instead. Pass a second parameter to this function as the base class for
+the new repository.
+
+There are two options for doing this:
+
+#### 1. Using a base repository class
+
+Create a base repository with your custom implementation, and then specify this
+repository as the base class.
+
+```ts
+class MyRepoBase<
+  E extends Entity,
+  IdType,
+  Relations extends object
+> extends DefaultCrudRepository<E, IdType, Relations> {
+  // Custom implementation
+}
+
+const BookRepositoryClass = defineRepositoryClass<
+  typeof BookModel,
+  MyRepoBase<BookModel, typeof BookModel.prototype.id, BookRelations>
+>(BookModel, MyRepoBase);
+```
+
+#### 2. Using a Repository mixin
+
+Create a repository mixin with your customization as shown in the "
+[Defining A Repository Mixin Class Factory Function](https://loopback.io/doc/en/lb4/migration-models-mixins.html#defining-a-repository-mixin-class-factory-function)
+" example, apply the mixin on the base repository class (e.g.
+`DefaultCrudRepository`) then specify this combined repository as the base class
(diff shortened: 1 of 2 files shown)


docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -254,6 +254,10 @@ children:
     url: Serving-static-files.html
     output: 'web, pdf'
 
+  - title: 'Dynamic models, repositories, and controllers'
+  url: Dynamic-models-repositories-controllers.html
```
dhmlau, line 258: missing indent. I think this .yml needs to have the right indentation in order for it to work.

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+Now that a datasource is ready, a repository can be created for `BookModel`.
+
+The `@loopback/rest-crud` package exports a helper function,
+`defineCrudRepositoryClass`, this function creates a
+`defineCrudRepositoryClass`-based entity CRUD repository class for the given
+model.
+
+```ts
+const BookRepository = defineCrudRepositoryClass(BookModel);
+```
```
bajtos, line 130: This will create a repository with `id` typed as `unknown` and empty relations (no navigational properties). Again, this may be ok if users don't need to access `findById` or included related models from code. However, in many cases, users will need the repository class to understand the real id type and the actual navigational properties. We should show them how to accomplish that:

```ts
const BookRepository = defineCrudRepositoryClass<
  Book,
  typeof Book.prototype.id,
  BookRelations
>(BookModel);
```

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+
+If you want to customize the repository, you can create a copy of
+`defineCrudRepositoryClass`'s
+[implementation](https://github.com/strongloop/loopback-next/blob/00917f5a06ea8a51e1f452f228a6b0b7314809be/packages/rest-crud/src/repository-builder.ts#L25-L47)
+and modify it according to your requirements.
```
bajtos, line 135: I don't think this is the best advice. `defineCrudRepositoryClass` is just a thin wrapper for `defineRepositoryClass`, which is in turn a small function creating a named repository class.

There are several other options how users can customize their repositories.

1. Put the custom behavior into a repository base class. Use that base class as the base to inherit from.

    ```ts
    class MyRepoBase<
      E extends Entity,
      IdType,
      Relations extends object
    > extends DefaultCrudRepository<E, IdType, Relations> {
      // put your custom code here
    }

    const BookRepositoryClass = defineRepositoryClass<
      typeof BookModel,
      MyRepoBase<BookModel, typeof BookModel.prototype.id, BookRelations>
    >(BookModel, MyRepoBase);
    ```

2. Use Repository Mixins to implement custom behavior and then apply the mixins when creating the repository class.

   Assuming we already have [FindByTitleRepositoryMixin from our docs](https://loopback.io/doc/en/lb4/migration-models-mixins.html#defining-a-repository-mixin-class-factory-function):

    ```ts
    const BookRepositoryClass = FindByTitleRepositoryMixin(
      defineCrudRepositoryClass<
        BookModel,
        typeof BookModel.prototype.id,
        BookRelations
      >(BookModel),
    );
    ```

    The code above is simple, but may define a class constructor with incorrect name, because we create `BookRepository` first and then apply the mixin second. Here is a bullet-proof approach that is unfortunately more complex:

    ```ts
    const BookRepositoryClass = defineRepositoryClass<
      typeof BookModel,
      DefaultCrudRepository<
        BookModel,
        typeof BookModel.prototype.id,
        BookRelations
      > &
        FindByTitle<BookModel>
    >(BookModel, FindByTitleRepositoryMixin(DefaultCrudRepository));
    ```

Please rework the paragraph about repository customization to show both Repository Base Class and Repository Mixin options.

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+
+If you want to use a non-`defineCrudRepositoryClass` repository class as the
+base for your repository, you can use the `defineRepositoryClass()` function
+instead.
+
+```ts
+import defineRepositoryClass from '@loopback/repository';
+// Assuming UserRepository is a pre-existing Repository class in the app
+const StudentRepository = defineRepositoryClass(BookModel, UserRepository);
```
bajtos, line 144: FYI:

https://github.com/strongloop/loopback-next/blob/20d01a28ed4f192042ff43e2a8dcb3ea98892ab3/packages/repository/src/define-repository-class.ts#L73-L75

Fortunately, if `UserRepository` is not a generic class (which I assume is your intention here), then the compiler infers all types correctly and the comment above does not apply.

```ts
class UserRepository extends DefaultCrudRepository<User, typeof User.prototype.id, UserRelations> {
}

const StudentRepository = defineRepositoryClass(StudentModel, UserRepository);
// StudentRepository has type ModelRepositoryClass<StudentModel, UserRepository>
```

Please add a short note (or a tip?) to provide this explanation to readers too. For example:
1. Explain that in your code snippet, `UserRepository` is not a generic class and therefore the compiler is able to infer everything automatically.
2. If the base class was a generic one, then the user would need to provide both generic parameters (`typeof BookModel` and e.g. `DefaultCrudRepository<BookModel, number, BookRelations>`).

bajtos, line 144:
```suggestion
const StudentRepository = defineRepositoryClass(StudentModel, UserRepository);
```

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+```
+
+Dependency injection has to be manually done for the datasource as shown below.
```
bajtos, line 147: The code below is not performing (doing) dependency injection (injecting the dependencies to the constructor), it's configuring dependency injection to be performed by the framework.

```suggestion
Dependency injection has to be configured for the datasource as shown below.
```

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+
+Note, the `app.repository()` method will be available only on `RepositoryMixin`
+apps.
+
+### Defining a Controller
+
+Once a repository is set up, a controller can be created for the model using
+the `defineCrudRestController` helper function from the `@loopback/rest-crud`
+package. It accepts a Model class and a `CrudRestControllerOptions` object.
+Dependency injection for the controlller has to be manually done.
```
bajtos, line 162: Ditto.

```suggestion
You will also have to configure dependency injection for the controller by applying the `inject` decorator manually.
```

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+LoopBack provides the ability to dynamically create models, repositories, and
+controllers during runtime. This document is an explanation of how to go about
+doing it.
+
+### Defining a ModelDefinition object
+
+A `ModelDefinition` object is the first step to creating a LoopBack model, it is
+an abstraction for specifying the various attributes of a LoopBack model. A
+`ModelDefinition` object is instantiated by passing a name or a
+[ModelDefinitionSyntax](https://loopback.io/doc/en/lb4/apidocs.repository.modeldefinitionsyntax.html#modeldefinitionsyntax-interface)
```
bajtos, line 25: Please use a relative link. I think the following should work, please double check:

```suggestion
[ModelDefinitionSyntax](./apidocs/repository.modeldefinitionsyntax.html#modeldefinitionsyntax-interface)
```

You will most likely need to run `npm run tsdocs` to populate API doc files, otherwise the link will return 404.

The same comment applies to all other places where you are linking to apidocs, please find and update them yourself.

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Model discovery
+
+Some datasource connectors provide methods for discovering model definitions
+from existing database schema. The following APIs can be very useful when
+defining LoopBack models dynamically.
+
+```ts
+// List database tables and/or views
+ds.discoverModelDefinitions({views: true, limit: 20}, cb);
```
bajtos, line 196: Uh oh, isn't there a Promise based version of this method? All public API in LB4 is based on Promises, we don't use callbacks anywhere. If discovery-related DataSource methods are not Promise-enabled yet, then please open a follow-up issue to improve that and update this part of our docs too.

Having said that, AFAICT, `discoverModelDefinitions` does implement support for Promise style invocation and the typings describe both callback and promise styles.

Please:
1. Check which discovery-related methods support Promises and update examples here in docs to use `await`.
2. If there are any methods you would like to show here but that don't support Promises, then open a follow-up issue to fix that.

```suggestion
await ds.discoverModelDefinitions({views: true, limit: 20});
```

docs/site/Dynamic-models-repositories-controllers.md
bajtos: Loosely related: https://github.com/strongloop/loopback-next/issues/5565 and [How to reuse LoopBack repository code](https://youtu.be/hRb70_JcNWI)

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+  connector: require('loopback-connector-mongodb'),
+  url: 'mongodb://sysop:moon@localhost',
+});
+await bookDs.connect();
+app.dataSource(bookDs, dsName);
+```
+
+Connect the datsource by calling its `.connect()` method and attach it to the
+app using `app.dataSource()` method. Note, this method will be available only
+on `RepositoryMixin` apps.
```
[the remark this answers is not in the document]
hacksparrow: Updated it to:

_To use the newly created datasource, call its `.connect()` method and attach it
to the app using `app.dataSource()` method._

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,218 @@
...
+
+A LoopBack model class is created by passing a `ModelDefinition` object to
+`@loopback/repository`'s helper function `defineModelClass()`. It expects a
+base model to extend (typically `Model` or `Entity`), folowed by the model
+definition object. In this case it will be `Entity`.
+
+```ts
+const BookModel = defineModelClass(Entity, bookDef);
+```
+
```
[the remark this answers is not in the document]
hacksparrow: Let's use the format number 2 throughout the docs, no need for number 1.

hacksparrow:
> Can you please update the pages listed in #4296 (comment) to mention the different ways how a Model/Repository/Controller can be created?

Shall we do it in a separate PR? 

I have applied all your feedback. Thanks a lot @bajtos.

bajtos:
> Shall we do it in a separate PR?

I am fine either way, as long as the changes are made as part of #4296 in the next few days.

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -112,39 +118,73 @@ await bookDs.connect();
...
+  MyRepoBase<BookModel, typeof BookModel.prototype.id, BookRelations>
+>(BookModel, MyRepoBase);
+```
+
+#### 2. Using a Repository mixin
+
+Create a repository mixin with your customization as shown in the
+"
+[Defining A Repository Mixin Class Factory Function](https://loopback.io/doc/en/lb4/migration-models-mixins.html#defining-a-repository-mixin-class-factory-function)
+" example, and then specify this repository as the base class.
```
bajtos, line 173:
```suggestion
" example, apply the mixin on the base repository class (e.g. `DefaultCrudRepository`) then specify this combined repository as the base class to be used.
```

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -254,6 +254,10 @@ children:
     url: Serving-static-files.html
     output: 'web, pdf'
 
+  - title: 'Dynamic models, repositories, and controllers'
```
dhmlau, line 257: maybe `Dynamically adding models, repositories, and controllers`?  to match the title of the docs page.

docs/site/Dynamic-models-repositories-controllers.md
```diff
@@ -0,0 +1,259 @@
+---
+lang: en
+title: 'Dynamically add models, repositories, and controllers during runtime'
```
dhmlau, line 3:
```suggestion
title: 'Dynamically adding models, repositories, and controllers during runtime'
```

Issue #5314, opened 2020-05-03
Fileds filter not seem to be working properly with nested relations


pratikjaiswal15:
> Note
> This issue was spun off from a Slack discussion and StackOverflow question:
> - https://loopbackio.slack.com/archives/C01177XQN8N/p1588431058303400
> - https://stackoverflow.com/questions/61560572/loopback-4-how-to-use-fields-filter-with-custom-scope-in-nested-relations

---

I have four models- purchase, purchase products, vendors, and product.

The purchase has many purchase products and vice versa
purchase products belong to product and vice versa
Vendors has many purchase and vice vera
Now with the following code, I am getting all purchases, purchase-products, and vendors(with field filter) but not able to use fields filter with the product model.
Is it a bug or I am doing it wrong way 

```
async find(

  ): Promise<Purchase[]> {
    return this.purchaseRepository.find(
      {

        include: [{
          relation: 'purchaseProducts',
          scope: {
            fields: { name: false, product_id: false}, // not working
            include: [{ relation: 'products' }],
          }

        }, {

          relation: 'vendors',
          scope: {
            fields: { address_line1: false, city: false, state: false, pincode: false, gst_number: false, // working }
          }
        }]
      }
    );

  }


```

And for relation vendors fields `name : true` is not working, so I have to fields `name : false`.
Thank you in advance

purchase.model.ts

```
import { Entity, model, property, hasMany, belongsTo } from '@loopback/repository';
import { PurchaseProduct } from './purchase-product.model';
import { Vendor } from './vendor.model';

@model({ settings: { strict: true } })
export class Purchase extends Entity {


  @property({
    type: 'string',
    id: true,

  })
  purchase_id: string;


  @property({
    type: 'date',
  })
  date?: string;


  @property({
    type: 'number',
    required: true,

  })
  totalprice?: number;
  @hasMany(() => PurchaseProduct, { keyTo: 'purchase_id' })
  purchaseProducts: PurchaseProduct[];

  @belongsTo(() => Vendor, { name: 'vendors' })
  vendor_id: number;
  // Define well-known properties here

  // Indexer property to allow additional data
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  [prop: string]: any;

  constructor(data?: Partial<Purchase>) {
    super(data);
  }
}

export interface PurchaseRelations {
  // describe navigational properties here
}

export type PurchaseWithRelations = Purchase & PurchaseRelations;


```

purchase-products.model.ts

```
[...]

Issue #4992, opened 2020-03-29
Failed to exclude properties by setting false in `fields` filter via url


agnes512:
## Description

From the docs https://loopback.io/doc/en/lb3/Fields-filter.html ( it's the same for LB4), filter `fields` can be set to `true/false` in a query to include/exclude model properties. `fields` works with API Explorer and Node. But it couldn't set properties to false (exclude) via url.
 
## Current Behavior

For example, model `Order` has three properties: `id`, `name`, and `price`. To exclude the `name` property from the result, as the docs suggested, the query can be written as `?filter[fields][name]=false`. But it returns:
```
[{name: 'order 1'}, {name: 'order 2'}..]
```
i.e it includes only the `name` property instead. I checked the filter sent to the controller, it shows that the filter is being converted to 
```
{fields: { name: 'false'}}  // string
```
while with API Explorer, the filter is converted to 

```
{fields: { name: false}}  // boolean
```

The wrong type might cause the problem.

In order to make it work with urls, the query needs to be set as `?filter[fields][id]=true&filter[fields][price]=true`

The issue might relate to https://github.com/strongloop/loopback-next/issues/2208

## Expected Behavior

`fields` should be able to exclude properties by setting properties to false via url just like what API Explorer/Node does.

## Acceptance

- [ ] ~check `where` clause if it has the same issue ( see [this](https://github.com/strongloop/loopback-next/pull/4620/files#diff-cff884992863b17e546339fb250d2811R113)).~
- [x] fix it so that API Explorer/Node/url all have the same behavior 
- [x] add tests to check the filter that is sent to the controllers
- [x] fix docs in `LB4/Working-with-data/fields-filter` page
- [x] Based on the [code](https://github.com/strongloop/loopback-next/blob/a81ce7e1193f7408d30d984d0c3ddcec74f7c316/packages/repository/src/query.ts#L508-L523), it seems accept arrays `{fields: ['foo', 'bar']}`. Check if array works for fields. Add proper normalization and error checking if needed. (explained in comment https://github.com/strongloop/loopback-next/issues/4992#issuecomment-651862792)

dhmlau:
@agnes512, could you please add acceptance criteria? Thanks.

agnes512:
can start at https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/coercion/coerce-parameter.ts#L172

agnes512:
Note: stringified JSON works well e.g `?filter[fields][name]=false` works

[...]

Issue #3455, opened 2019-07-26
Reject queries with incompatible "filter.fields" and "filter.include"


bajtos:
The filter property `fields` allows callers to limit which model properties are returned by the database. This creates a problem when the primary or the foreign key is excluded from the data, because then we cannot resolve included models.

```ts
categoryRepo.find({
  fields: ['name'],
  include: [{relation: 'products'}],
});
```

In LB3, I think we were silently ignoring `include` filter in such case.

I am proposing to be more explicit in LB4 and reject such queries with a 4xx (Bad Request) error.

Later, we can improve our implementation to automatically add PK/FK properties to the `fields` configuration and remove PK/FK properties from the data returned back to the user, so that inclusions are resolved as expected and yet the data contains only the specified properties.

See also https://github.com/strongloop/loopback-next/pull/3387

## Acceptance criteria

TBD - will be filled by the team.

InvictusMB:
@bajtos Do you think #5592 is a viable solution for this?

bajtos:
> Do you think #5592 is a viable solution for this?

Yes, if we can fix resolvers to transparently add the necessary `fields`, then there is no need reject queries excluding PK/FK fields.

PR #5354, opened 2020-05-07
feat(repository): adding hasManyThrough (through model) to hasMany and its helpers

agnes512:
Extract hasManyThrough relation definition and its helpers from #4438 

Adding unit tests for `hasManyThrough.help`

Changes: 
- Move the through member to `HasManyDefinition` and get rid of `hasManyThroughDefinition`.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5592, opened 2020-05-28
Transparently handle foreign keys in inclusion resolvers

InvictusMB:
`findByForeignKeys` adds the foreign key to the filter fields if it's not there.
Concerned inclusion resolvers prune the foreign key value from entities if it has not been requested in the original filter.

Fixes: #5314
Reasoning: [comment](https://github.com/strongloop/loopback-next/issues/5314#issuecomment-626804560)
Obsoletes: #3455
Also addresses: [comment](https://github.com/strongloop/loopback-next/issues/4992#issuecomment-635285988)

## Checklist


packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
```diff
@@ -54,6 +54,7 @@ export function createBelongsToInclusionResolver<
     const targetKey = relationMeta.keyTo as StringKeyOf<Target>;
     const dedupedSourceIds = deduplicate(sourceIds);
 
+    const pruneTargetKey = !hasFields([targetKey], inclusion.scope);
```
agnes512, line 57: I have a question about the design. I believe the problem we are trying to solve here is that to make sure the inclusion works for cases that some properties are excluded in the `fields` filter. 

What if the fk is excluded at the first level? e.g for hasMany relation, `Customer.id` is excluded:
```
await customerRepo.find({
        fields: {id: false},
        include: [
          {
            relation: 'orders',
            scope: {
              // fields: {customerId: false},
          }
        ]
      });
```
Should we also check `sourceKey` here? ( but then to me it becomes kind of hard-coded)

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -81,6 +81,88 @@ describe('HasMany relation', () => {
...
+      description: "someone else's order desc",
+      customerId: existingCustomerId + 1, // a different customerId,
+    });
+
+    const orders = await customerOrderRepo.find({
+      fields: {
+        description: true,
+        customerId: false,
+      },
+    });
```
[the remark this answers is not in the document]
InvictusMB: Will do, the two broken tests seem to highlight that too.

packages/repository/src/query.ts
```diff
@@ -526,6 +527,13 @@ export class FilterBuilder<MT extends object = AnyObject> {
     assert(order.match(/^[^\s]+( (ASC|DESC))?$/), 'Invalid order: ' + order);
   }
 
+  private normalizeFields() {
+    const fields = this.filter.fields;
+    if (Array.isArray(fields)) {
+      this.filter.fields = Object.fromEntries(fields.map(k => [k, true]));
+    }
+  }
+
```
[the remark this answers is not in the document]
InvictusMB: I'd like to hear from from @raymondfeng or @bajtos regarding this. It is possible that this piece should go away if the `@repository` Filter is not intended to follow the `loopback-datasource-juggler` Filter definition. 
In that case, there should be another bug filed for silently misbehaving with `fields: string[]` filters.
At the moment it is possible to pass an array inside `model()` decorator or via REST API and observe weird output.

packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
InvictusMB: Indeed, however we cannot work around this in scope of inclusion resolver. The only thing we could do here is error out if `sourceKey` is not set on entities. Maybe add this check here:
https://github.com/strongloop/loopback-next/blob/bfbea58dd441e92a022147fc43c8e9ffefcd81a4/packages/repository/src/relations/relation.helpers.ts#L96

I believe these 3 guys could do a similar trick on the `sourceKey`:
https://github.com/strongloop/loopback-next/blob/bfbea58dd441e92a022147fc43c8e9ffefcd81a4/packages/repository/src/repositories/legacy-juggler-bridge.ts#L388
https://github.com/strongloop/loopback-next/blob/bfbea58dd441e92a022147fc43c8e9ffefcd81a4/packages/repository/src/repositories/legacy-juggler-bridge.ts#L400
https://github.com/strongloop/loopback-next/blob/bfbea58dd441e92a022147fc43c8e9ffefcd81a4/packages/repository/src/repositories/legacy-juggler-bridge.ts#L418

packages/repository/src/query.ts
```diff
@@ -633,6 +641,30 @@ export class FilterBuilder<MT extends object = AnyObject> {
...
+/**
+ * Checks if all fields are present in a target filter
+ *
+ * @param fields - array of fields to search in a filter
+ * @param filter - target filter
+ *
+ */
+export function hasFields<T extends object = AnyObject>(
+  fields: (keyof T)[],
+  filter?: Filter<T>,
```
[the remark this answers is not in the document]
InvictusMB: I'd rather rename the helper to `matchesFields`. It checks whether `fields` property of filter is a superset of `fields` array passed as an argument.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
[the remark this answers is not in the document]
InvictusMB: Done

packages/repository/src/query.ts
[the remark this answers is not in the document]
InvictusMB: Please check if you like the updated version better

packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
agnes512: After considering the consistency and the effort it might need, personally I think allowing the excluded fk to be added in the query is good enough. If they would like to hide the source key from the response, the hidden property feature is a better solution.

InvictusMB: I just went ahead and implemented this as well. It wasn't as hard given the utilities I had created already. The only extra effort was exposing relation metadata on the resolver itself.

InvictusMB:
> Do you mind also update the bottom part of the docs https://loopback.io/doc/en/lb4/Fields-filter.html that excluding the source key for relations might cause errors? Thanks!

That's not necessary anymore. The latest commit addresses source keys too.

InvictusMB:
@raymondfeng @agnes512 @hacksparrow 
Could you guys check what else do we need for this to land?

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -385,34 +391,67 @@ export class DefaultCrudRepository<
     }
   }
 
+  /**
+   * Template method to amend filters with foreign key sources
```
agnes512, line 395: Could you add tests for this function and make the description more clear? for example what issues it tries to address, and maybe add an example like [this](https://github.com/strongloop/loopback-next/pull/5354/files#diff-4eac471411475eed05f0d6bcb873e607R29-R53)

packages/repository/src/repositories/legacy-juggler-bridge.ts
```diff
@@ -385,34 +391,67 @@ export class DefaultCrudRepository<
...
+        );
+        if (!model) {
+          return [] as (T & Relations)[];
+        }
+        const entity = this.toEntity(model);
+        return this.includeRelatedModels([entity], include, options);
+      },
     );
-    return resolved[0];
+    return resolved[0] || null;
```
agnes512, line 454: I am not sure if we should have ` || null` here. I think the empty value is handled before retuning here?

bajtos:
Thank you @InvictusMB for the pull request 👏 and @agnes512 for reviewing the changes 💪 

I need to find out more time to build a better understanding of the wider context and asses the implications of the proposed changes. I'll try to do that early next week 🤞 Sorry for the delay! 🙈

InvictusMB:
@bajtos Sorry, the PR grew out of control. Initially it was only about targets of a relation.
Here are the key points:
* We have to detect all the fields that are involved in all of the inclusions of a particular `filter` and ensure those fields are included in a filter we use to resolve both the target entity and the inclusions.
* We fill the `targetKey` in a resolver.
* We fill the `sourceKey` in `find*` methods of a repository.
* Pruning empties the fields that were added to the filter here but had not been requested originally or had been disabled intentionally.
* We must expose relation metadata because there are two sides to the relation. Resolver only knows about the target entity it resolves and it can only fill the `targetKey`. It receives the source entities as is and if the `sourceKey` was filtered out there is no way to get it back without an extra query. That's why right now `sourceKey` has to be handled in `find*` methods of a repository before passing entities into a resolver.

I don't think we can fully encapsulate this behavior inside a resolver the way you wish.
We could probably do something very clever and delay the source entity resolution until we collect all the filter amendments from inclusion resolvers via callbacks. But I'm not sure this async ping pong would be comprehensible.

<details>
  <summary>Very clever code</summary>
  
```ts
  async find(
    filter?: Filter<T>,
    options?: Options,
  ): Promise<(T & Relations)[]> {
    const include = filter?.include;

    const entityResolver = async (newFilter?: Filter<T>) => {
      const models = await ensurePromise(
        this.modelClass.find(this.normalizeFilter(newFilter), options),
      );
      return this.toEntities(models);
    }

    let amendFilter;
    const entityPromise = new Promise(resolve => {
      let count = 0;
      amendFilter = (newFilter: Filter<T>) => {
        if (count === include?.length) {
          resolve(entityResolver(newFilter));
        }
        count++;
      };
    });

    return this.includeRelatedModels(
      entityPromise,
      amendFilter,
      include,
      options,
    );
  }
```
 
</details>

InvictusMB:
@bajtos I made a change that allows the resolvers to stay self sufficient. Instead of entities they receive an async callback that will return the entities. This callback allows resolvers to communicate a list of fields to include in those entities. See if you like this approach better.

I still consider this code too clever and not very reliable. If one of the resolvers does not invoke the callback to get entities, the whole async thread will remain hanging.

<details>
  <summary>Somewhat clever code</summary>

https://github.com/strongloop/loopback-next/blob/0f83b1a0e7355cb5f09f7754930b232bfd8cfa8d/packages/repository/src/relations/relation.helpers.ts#L113-L148
 
</details>

Could you also address [this comment](https://github.com/strongloop/loopback-next/pull/5592#discussion_r432038757)?

packages/repository/src/repositories/legacy-juggler-bridge.ts
InvictusMB: It's gone in the latest iteration

packages/repository/src/repositories/legacy-juggler-bridge.ts
InvictusMB: Not really, we can get an empty array we guard against undefined here.

InvictusMB:
I have refined this a bit more. It looks even more awful but works. Now it is guaranteed that all the promises resolve even if inclusion resolvers don't query source entities.

<details>
  <summary>Latest version</summary>
  
https://github.com/strongloop/loopback-next/blob/89934bb38f1f063d4d4eaf08c415a8fcbfea2c52/packages/repository/src/relations/relation.helpers.ts#L113-L155
 
</details>

agnes512:
I will review it tomorrow 💪

packages/repository/src/query.ts
```diff
@@ -698,6 +698,70 @@ export class FilterBuilder<MT extends object = AnyObject> {
...
+  }
+  return true;
+}
+
+/**
+ * Ensures that a filter always picks target fields.
+ *
+ * @param targetFields - fields to include
+ * @param filter - a target filter
+ */
```
agnes512, line 733: IIUC, `Ensures that a filter always picks target fields.` means 
`queries that apply the returned filter would always include the target fields. If the field in the targetFields is supposed to be returned as undefined, it will be added to the pruneFields for latter processing.` Could you make API docs more descriptive and add the explanation for what is being returned ( `@return ......`)?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -81,6 +81,190 @@ describe('HasMany relation', () => {
...
+  it('includes only the fields set in filter', async () => {
+    await customerOrderRepo.create({
+      description: 'an order desc',
+    });
+    const customer = await customerRepo.findById(existingCustomerId, {
+      include: [
+        {
+          relation: 'orders',
+          scope: {
+            fields: {},
```
agnes512, line 247: If the fields is `{}`, I'd expect that all the properties would be included. Same for the test at line 486.

InvictusMB: It is the current behavior for both lb4 and juggler filters. I only verify here that it was not affected.

packages/repository/src/query.ts
InvictusMB: Fixed.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
agnes512: oh wow I didn't know this. Thanks for verifying.

bajtos:
> Here are the key points:
> 
> * We have to detect all the fields that are involved in all of the inclusions of a particular `filter` and ensure those fields are included in a filter we use to resolve both the target entity and the inclusions.
> * We fill the `targetKey` in a resolver.
> * We fill the `sourceKey` in `find*` methods of a repository.
> * Pruning empties the fields that were added to the filter here but had not been requested originally or had been disabled intentionally.
> * We must expose relation metadata because there are two sides to the relation. Resolver only knows about the target entity it resolves and it can only fill the `targetKey`. It receives the source entities as is and if the `sourceKey` was filtered out there is no way to get it back without an extra query. That's why right now `sourceKey` has to be handled in `find*` methods of a repository before passing entities into a resolver.

Thank you for the explanation, @InvictusMB. I did not realize that the first query is performed by `find*` and that we need to ensure relation's `sourceKey` is included in the `fields` part of that first query.

In that light, it makes sense to let inclusion resolvers to provide information about the fields that must be fetched from the database in the first query.

What I find as possible problematic is exposing the entire relation metadata object. If you look at [RelationDefinitionBase](https://github.com/strongloop/loopback-next/blob/56fed8861b2758400b5e01755d2bbe650beb63ed/packages/repository/src/relations/relation.types.ts#L21-L53), it does not list any keys. There may be relation types that have different metadata than `sourceKey` and `targetKey` and because the contract "resolver provides full relation metadata" does not make it clear which parts are actually used, such relations would be difficult to implement. I'd prefer to find a different way how an inclusion resolver can communicate the list of fields to include, one that's not tied to relation metadata. Instead, the contract should provides explicit description of what is expected from the resolver and it should allow the resolver to build that information independently on how is relation metadata structured.

For example, instead of exposing `resolver.relationMeta`, we can ask resolvers to provide a list of fields that must be included in the source entities provided to the resolver function - e.g. `resolver.requiredSourceFields = ['categoryId']`. IMO, such contract/interface is more explicit about what part of metadata is needed and how it will be interpreted, and allows resolvers to build that information in any way they like.

> Pruning empties the fields that were added to the filter here but had not been requested originally or had been disabled intentionally.

This step makes sense now, I was confused about the implementation details and variable/property naming.

Here is my thinking about this problem:

- The method `ensureFields` is modifying the query to ensure given fields are always included. I think we should not limit this method to inclusion resolving only (where we want to prune those fields afterwards) and we should think about it in more generic terms. Instead of returning `pruneFields` object that's specific to inclusion resolvers, I'd prefer to return a list of field names that were added to the query on top of what was requested by the user. Example names: `fieldsAdded`, `extraFields`, etc.

- When resolving relations, we want to prune the extra fields added on top of the user provided query. We can assign the value from `fieldsAdded` to `pruneFields`. We usually start variable names with a noun (verbs are for functions), so something like `fieldsToPrune`, `extraFields` or `fieldsAddedForInclusionResolvers` may be a better name.

> I don't think we can fully encapsulate this behavior inside a resolver the way you wish.
> We could probably do something very clever and delay the source entity resolution until we collect all the filter amendments from inclusion resolvers via callbacks. But I'm not sure this async ping pong would be comprehensible.

Yeah, I agree the callback-based version is much more complex and thus difficult to understand & reason about.

I think the callback-based version will also not work when including more than one relation, e.g. when geting all `TodoList` instances and including both the related `Todo` items (via hasMany) and the related `TodoListImage` (via hasOne).

I am proposing to get back to your original proposal and improve it a bit. I am not sure what was the latest version before you rewrote the code & git history, maybe https://github.com/strongloop/loopback-next/commit/e29027560e30b82b3675f515ffdd4ceeef700233 is a good commit to use? 

I apologize for pointing you on a wrong track and wasting your effort on a dead end 🙈

packages/repository/src/query.ts
```diff
@@ -688,6 +698,72 @@ export class FilterBuilder<MT extends object = AnyObject> {
...
+  const pruneFields = _.fromPairs(
+    prune.map(f => [f, undefined]),
+  ) as PruningMask<T>;
+
+  const newFilter = builder.build();
+  // if the filter only hides the fields, unset the entire fields clause
+  if (isDisablingOnly) {
+    delete newFilter.fields;
+  }
+  return [newFilter, pruneFields] as const;
```
bajtos, line 762: Let's return an object with named properties instead of tuple please, that way it's easier to understand what values are returned by this helper. See e.g. one of the older versions of your patch (e29027560):

```ts
  return {
    filter: builder.build(),
    pruneFields,
  };
```

bajtos: (But please choose a different property name than `pruneFields`, see https://github.com/strongloop/loopback-next/pull/5592#issuecomment-646665314):

> The method `ensureFields` is modifying the query to ensure given fields are always included. I think we should not limit this method to inclusion resolving only (where we want to prune those fields afterwards) and we should think about it in more generic terms. Instead of returning `pruneFields` object that's specific to inclusion resolvers, I'd prefer to return a list of field names that were added to the query on top of what was requested by the user. Example names: `fieldsAdded`, `extraFields`, etc.

dhmlau:

```
git commit --amend -s
git push --force-with-lease
```

packages/repository/src/query.ts
InvictusMB: Fixed

InvictusMB:
> BTW would you mind to extract the implementation of `ensureFields` including the tests to a new pull request, so that we can land that building block first and then have less code to review when discussing how to fix inclusion resolvers?

Separated into #6495

InvictusMB:
FK source  part split into #6496

packages/repository/src/query.ts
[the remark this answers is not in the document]
InvictusMB: Removed this piece. Maybe I'll create a separate PR dedicated to FilterBuilder later.

packages/repository/src/relations/belongs-to/belongs-to.inclusion-resolver.ts
InvictusMB: FK source keys are now handled in #6496

InvictusMB:
@agnes512 @hacksparrow 
Could you review once more? I think it's good to merge now. 

#6495 has to land first though.

PR #5593, opened 2020-05-28
feat(context): add `tryCatchFinally` helper function for value or promise

raymondfeng:
Add a helper function to apply `try-catch-finally` to value or promise.

Extracted from https://github.com/strongloop/loopback-next/pull/5586.

## Checklist

### DIFF
--- a/packages/context/src/__tests__/unit/value-promise.unit.ts
+++ b/packages/context/src/__tests__/unit/value-promise.unit.ts
@@ -10,45 +10,7 @@ import {
   resolveMap,
   resolveUntil,
   transformValueOrPromise,
-  tryWithFinally,
 } from '../..';
-
-describe('tryWithFinally', () => {
-  it('performs final action for a fulfilled promise', async () => {
-    let finalActionInvoked = false;
-    const action = () => Promise.resolve(1);
-    const finalAction = () => (finalActionInvoked = true);
-    await tryWithFinally(action, finalAction);
-    expect(finalActionInvoked).to.be.true();
-  });
-
-  it('performs final action for a resolved value', () => {
-    let finalActionInvoked = false;
-    const action = () => 1;
-    const finalAction = () => (finalActionInvoked = true);
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    tryWithFinally(action, finalAction);
-    expect(finalActionInvoked).to.be.true();
-  });
-
-  it('performs final action for a rejected promise', async () => {
-    let finalActionInvoked = false;
-    const action = () => Promise.reject(new Error('error'));
-    const finalAction = () => (finalActionInvoked = true);
-    await expect(tryWithFinally(action, finalAction)).be.rejectedWith('error');
-    expect(finalActionInvoked).to.be.true();
-  });
-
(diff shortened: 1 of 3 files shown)


packages/context/src/__tests__/unit/value-promise.unit.ts
```diff
@@ -31,6 +32,29 @@ describe('tryWithFinally', () => {
...
+    const action = () => 1;
+    const errorAction = (err: unknown) => {
+      errorActionInvoked = true;
+      throw err;
+    };
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    tryCatchFinally(action, errorAction, () => {});
+    expect(errorActionInvoked).to.be.false();
+  });
+
```
bajtos, line 57: Please add tests to verify that `finally` callback is invoked both on success and on error.

I find it confusing that the tests for `tryCatchFinally` are in multiple places in the test file. Please reorganize them into their own `describe` block. Considering the already large size of this test file, it would be best to put the new tests into a new file, e.g. `packages/context/src/__tests__/unit/try-catch-finally.unit.ts`.

packages/context/src/value-promise.ts
```diff
@@ -191,32 +191,57 @@ export function resolveList<T, V>(
 export function tryWithFinally<T>(
   action: () => ValueOrPromise<T>,
   finalAction: () => void,
+): ValueOrPromise<T> {
+  const errorAction = (err: unknown): never => {
+    throw err;
+  };
+  return tryCatchFinally(action, errorAction, finalAction);
```
bajtos, line 198: Can we move the default error action to `tryCatchFinally` so that the implementation of `tryWithFinally` can be simplified as follows?

```suggestion
  return tryCatchFinally(action, undefined, finalAction);
```

packages/context/src/value-promise.ts
```diff
@@ -191,32 +191,57 @@ export function resolveList<T, V>(
...
+ * @param errorAction - A function to be called once the action
+ * is rejected (synchronously or asynchronously). It must throw an error.
+ * @param finalAction - A function to be called once the action
+ * is fulfilled or rejected (synchronously or asynchronously)
+ */
+export function tryCatchFinally<T>(
+  action: () => ValueOrPromise<T>,
+  errorAction: (err: unknown) => never = err => {
+    throw err;
+  },
```
bajtos, line 215: What if the user wants to handle the error in such way that the entire operation resolves the final promise with a value? For example, we may want to catch "not found error" and convert it to "undefined" result.

raymondfeng:

packages/context/src/value-promise.ts
```diff
@@ -187,36 +187,63 @@ export function resolveList<T, V>(
...
-        return val;
-      },
-      err => {
-        finalAction();
-        throw err;
-      },
-    );
-  } else {
-    finalAction();
+    return (result = result.then(resolve, reject));
```
bajtos, line 227: I find this a bit cryptic. Why are you updating `result` to the promise returned by `result.then()`? Which code is going to access the updated `result` value? 

Can we simplify this line as follows - would it work the same way as now?

```suggestion
    return result.then(resolve, reject);
```

PR #5605, opened 2020-05-29
feat(repository): add more helpers for HasManyThrough

agnes512:
Add two more helpers for HasManyThrough relation

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
+++ b/packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2019. All Rights Reserved.
+// Copyright IBM Corp. 2019,2020. All Rights Reserved.
 // Node module: @loopback/repository
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
@@ -9,6 +9,7 @@ import {
   constrainDataObjects,
   constrainFilter,
   constrainWhere,
+  constrainWhereOr,
   Entity,
   Filter,
   FilterBuilder,
@@ -74,6 +75,23 @@ describe('constraint utility functions', () => {
       expect(result).to.deepEqual({
         and: [inputWhere, constraint],
       });
+    });
+  });
+
+  context('constrainWhereOr', () => {
+    const inputWhere: Where<{x: string; y: string; id: string}> = {
+      x: 'x',
+    };
+    it('enforces a constraint', () => {
+      const constraint = [{id: '5'}, {y: 'y'}];
+      const result = constrainWhereOr(inputWhere, constraint);
+      expect(result).to.deepEqual({...inputWhere, or: constraint});
+    });
+
+    it('enforces constraint with dup key', () => {
+      const constraint = [{y: 'z'}, {x: 'z'}];
+      const result = constrainWhereOr(inputWhere, constraint);
+      expect(result).to.deepEqual({...inputWhere, or: constraint});
     });
   });
 
--- a/packages/repository/src/relations/has-many/has-many-through.helpers.ts
+++ b/packages/repository/src/relations/has-many/has-many-through.helpers.ts
@@ -7,10 +7,12 @@ import debugFactory from 'debug';
 import {camelCase} from 'lodash';
 import {
   DataObject,
+  deduplicate,
   Entity,
   HasManyDefinition,
   InvalidRelationError,
   isTypeResolver,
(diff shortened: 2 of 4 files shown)


packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -109,6 +117,45 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
...
+ * };
+ * createThroughConstraint(resolvedMetadata, {id: 3, name: 'a product'});
+ *
+ * >>> {'productId': 1}
+ *
+ *  * createThroughConstraint(resolvedMetadata, {id: {inq:[3,4]}});
+ *
+ * >>> {'productId': {inq:[3,4]}}
+ */
+export function createThroughFkConstraint<Target, Through extends Entity>(
```
agnes512, line 147: This new helper is created for deleting through models once the target instances get deleted

agnes512:
Added two helpers. They are all for deleting through models as we decided to have the delete method this way https://github.com/strongloop/loopback-next/pull/4438/files#r402105078

packages/repository/src/__tests__/unit/repositories/constraint-utils.unit.ts
```diff
@@ -77,6 +78,23 @@ describe('constraint utility functions', () => {
...
+    };
+    it('enforces a constraint', () => {
+      const constraint = [{id: '5'}, {y: 'y'}];
+      const result = constrainWhereOr(inputWhere, constraint);
+      expect(result).to.deepEqual({...inputWhere, or: constraint});
+    });
+
+    it('enforces constraint with dup key', () => {
+      const constraint = [{y: 'z'}];
+      const result = constrainWhereOr(inputWhere, constraint);
```
bajtos, line 93: I am little bit confused. The test name says "with dup key", but IIUC, the code is using keys `x` and `y`. Where is the duplicated key?

agnes512: @bajtos My bad, forgot to add the key back. Fixed.

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -109,6 +119,45 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
...
+ *  keyTo: 'id',
+ *  through: {
+ *    model: () => CategoryProductLink,
+ *    keyFrom: 'categoryId',
+ *    keyTo: 'productId',
+ *  },
+ * };
+ * createThroughConstraint(resolvedMetadata, {id: 3, name: 'a product'});
+ *
+ * >>> {'productId': 1}
```
jannyHou, line 143: IIUC the constraint here should be `{productId: 1}`?

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -109,6 +119,45 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
...
+ *  // .. other props
+ *  keyFrom: 'id',
+ *  keyTo: 'id',
+ *  through: {
+ *    model: () => CategoryProductLink,
+ *    keyFrom: 'categoryId',
+ *    keyTo: 'productId',
+ *  },
+ * };
+ * createThroughConstraint(resolvedMetadata, {id: 3, name: 'a product'});
```
jannyHou, line 141: The additional query field reminds me sth(sorry don't remember where is the original chat :p) : 
 
We want to distinguish the query on through model(CategoryProductLink) and target model(Product), how do we achieve that? 

For example, `{name: 'a product'}` seems for target model, and `createThroughConstraint` applies for through model, and that's why `name` is ignored here. But:

- where is the place that processes the target model's filter?
- how to specify the through model's filter?

(I started from the helper function, might missing the code/docs if they are created other places.)

agnes512: @jannyHou I think these helpers are just for handling the relation metadata. And filters are handled in the hasManyThrough repository. Here is an example in the original design: [find method](https://github.com/strongloop/loopback-next/pull/2359/files#diff-bd2fafeb3bd8f873174e06856a1d7471R124-R141). They won't be included in this PR. But I'd love to discuss it: 

So far from those two community PRs(#2359 , #4438 ), I only see the target  model is capable of applying filters. Not sure if we need it for through models ( I don't see the reason to allow filtering through models tho). Do we need it?

PR #5606, opened 2020-05-29
feat(repository): add repo factory and tests for hasManyThrough

agnes512:
Add repository factory for hasManyThrough

## Checklist

### DIFF
--- a/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
+++ b/packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
@@ -1,0 +1,93 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/repository
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {
+  DataObject,
+  Entity,
+  EntityCrudRepository,
+  Getter,
+  HasManyDefinition,
+} from '../..';
+import {
+  createTargetConstraint,
+  createThroughConstraint,
+  createThroughFkConstraint,
+  resolveHasManyThroughMetadata,
+} from './has-many-through.helpers';
+import {
+  DefaultHasManyThroughRepository,
+  HasManyThroughRepository,
+} from './has-many-through.repository';
+
+/**
+ * a factory to generate hasManyThrough repository class.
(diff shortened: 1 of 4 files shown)


agnes512:
For the coverage decreasing: 
- we already have unit tests for those helpers in hasManyThrough factory file. Not sure if they need to be tested again.
- those CRUD methods will be tested in hasManyThrough repository PR, so they are just retuning void atm.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,93 @@
...
+  createThroughFkConstraint,
+  resolveHasManyThroughMetadata,
+} from './has-many-through.helpers';
+import {
+  DefaultHasManyThroughRepository,
+  HasManyThroughRepository,
+} from './has-many-through.repository';
+
+/**
+ * a facroty to generate hasManyThrough repository class.
```
bajtos, line 25: typo

```suggestion
 * a factory to generate hasManyThrough repository class.
```

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -0,0 +1,93 @@
...
+export type HasManyThroughRepositoryFactory<
+  TargetEntity extends Entity,
+  TargetID,
+  ThroughEntity extends Entity,
+  ForeignKeyType
+> = (
+  fkValue: ForeignKeyType,
+) => HasManyThroughRepository<TargetEntity, TargetID, ThroughEntity>;
+
+export function createHasManyThroughRepositoryFactory<
```
bajtos, line 41: Do we need a new function? Have you considered to extend the existing `createHasManyRepositoryFactory` to support both "direct" and "through" variants? Maybe that's not a good idea, I just wanted to bring it forward for consideration.

agnes512: Good point... I considered a bit. Since we only have one hasMany type, I think it might be more consistent if we only expose `createHasManyRepositoryFactory` as the entry for hasMany* relations, and let `createHasManyRepositoryFactory` to decide which factory to create.

PR #5462, opened 2020-05-16
feat: run eslint/prettier fix against files as a pre-commit hook with lint-staged

raymondfeng:
Run prettier/eslint fix before `git commit`.

I have seen a lot of lint-related issues in PRs. Hopefully this pre-push
hook can decrease such problems and save us from keep telling contributors
to run `npm run lint:fix`.

A `pre-push` hooks is also added to run `npm run build` to prevent code that does not compile from being pushed to remote.

@frbuceta Thank you for the inspiration at #4330 

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5616, opened 2020-05-31
chore: lint staged files before commit

InvictusMB:
Adds precommit hook to lint staged files. Also fixes prettier EOL handling on Windows.

Running linting on the whole project manually takes a while. It is much more convenient to lint only modified files automatically every time.

Or is there a good reason not to do so?

**EDIT:** Just saw #5462. This one is slightly different though.

## Checklist

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,9 @@
+{
+  "*.{ts,js}": [
+    "node packages/build/bin/run-prettier --write",
+    "node packages/build/bin/run-eslint --fix --report-unused-disable-directives --cache"
+  ],
+  "*.{md}": [
+    "node packages/build/bin/run-prettier --write"
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -26,7 +26,9 @@
     "eslint-plugin-mocha": "^7.0.1",
     "fs-extra": "^9.0.1",
     "husky": "^4.2.5",
+    "if-env": "^1.0.4",
     "lerna": "^3.22.1",
+    "lint-staged": "^10.2.7",
     "markdown-table": "^2.0.0",
     "open-cli": "^6.0.1",
     "typescript": "~3.9.6"
@@ -81,6 +83,7 @@
   "copyright.owner": "IBM Corp.",
   "husky": {
     "hooks": {
+      "pre-commit": "if-env LINT_STAGED=0 && echo \"lint-staged disabled via LINT_STAGED env var\" || lint-staged",
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
   }
--- a/docs/site/DEVELOPING.md
+++ b/docs/site/DEVELOPING.md
@@ -229,6 +229,9 @@ npm script `lint:fix`.
 npm run lint:fix
 ```
 
+Files staged for commit are linted automatically. If necessary, pre-commit
+linting can be bypassed by setting the environment variable `LINT_STAGED=0`.
+
 ## Working with dependencies
 
 We use npm's
--- a/package-lock.json
+++ b/package-lock.json
@@ -2547,6 +2547,16 @@
         "humanize-ms": "^1.2.1"
       }
     },
+    "aggregate-error": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/aggregate-error/-/aggregate-error-3.0.1.tgz",
+      "integrity": "sha512-quoaXsZ9/BLNae5yiNoUz+Nhkwz83GhWwtYFglcjEQB2NDHCIpApbqXxIFnm4Pq/Nvhrsq5sYJFyohrrxnTGAA==",
+      "dev": true,
+      "requires": {
+        "clean-stack": "^2.0.0",
+        "indent-string": "^4.0.0"
+      }
+    },
     "ajv": {
       "version": "6.12.3",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.3.tgz",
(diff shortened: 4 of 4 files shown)


.prettierrc
```diff
@@ -3,6 +3,7 @@
   "singleQuote": true,
   "printWidth": 80,
   "trailingComma": "all",
+  "endOfLine": "auto",
```
bajtos, line 6: This should go to a standalone commit. A new PR would be even better, if you don't mind.

InvictusMB: Done in #5673

InvictusMB:
@bajtos 
Updated the PR as requested. Also implemented bypassing `lint-staged` altogether via `LINT_STAGED=false`.

InvictusMB:
Added remarks about pre-commit linting to the documentation.

.lintstagedrc
```diff
@@ -0,0 +1,9 @@
+{
+  "*.{ts,js}": [
+    "node packages/build/bin/run-prettier --write",
```
raymondfeng, line 3: Can we only fix changed files for `prettier`?

.lintstagedrc
```diff
@@ -0,0 +1,9 @@
...
+    "node packages/build/bin/run-eslint --fix --report-unused-disable-directives --cache"
+  ],
+  "*.{md}": [
+    "node packages/build/bin/run-prettier --write"
```
raymondfeng, line 7: Ditto.

.lintstagedrc
raymondfeng: Can we also add a print out to describe how to skip the check if the command fails?

InvictusMB: `lint-staged` runs command only on changed files that were staged for commit by definition.

This is the output when pre-commit hook fails:

<details>
  <summary>output</summary>
  
```
~ git commit -a -m "chore: fail prettier"
husky > pre-commit (node v12.16.0)
[STARTED] Preparing...
[SUCCESS] Preparing...
[STARTED] Running tasks...
[STARTED] Running tasks for *.{ts,js}
[STARTED] Running tasks for *.{md}
[STARTED] node packages/build/bin/run-prettier --write
[FAILED] node packages/build/bin/run-prettier --write [FAILED]
[FAILED] node packages/build/bin/run-prettier --write [FAILED]
[SUCCESS] Running tasks...
[STARTED] Applying modifications...
[STARTED] Reverting to original state because of errors...
[SUCCESS] Reverting to original state because of errors...
[STARTED] Cleaning up...
[SUCCESS] Cleaning up...

× node packages/build/bin/run-prettier --write:
[error] packages\repository\src\repositories\legacy-juggler-bridge.ts: SyntaxError: ';' expected. (464:8)
[error]   462 |   }
[error]   463 |
[error] > 464 |   fail here
[error]       |        ^
[error]   465 |
[error]   466 |   async updateById(
[error]   467 |     id: ID,
husky > pre-commit hook failed (add --no-verify to bypass)
``` 
</details>

Is this sufficient?

raymondfeng: @InvictusMB I understand the command is only triggered for changes files. But `node packages/build/bin/run-prettier --write` itself scans all files.

raymondfeng: I'm referring the possibility of https://github.com/strongloop/loopback-next/blob/63e8f6e1f91b0808696a11b123318d5089f2b2d0/lint-staged.config.js#L13.

raymondfeng: The output is good enough to understand what's failing. My comment was to print out an option to skip the failure so that the pre-commit hooks can be skipped.

InvictusMB: > I'm referring the possibility of
> 
> https://github.com/strongloop/loopback-next/blob/63e8f6e1f91b0808696a11b123318d5089f2b2d0/lint-staged.config.js#L13

Lint staged literally does that for us already.
That's what gets passed into `run-prettier`
```
run-prettier argv
[
  'C:\\DevTools\\nodejs\\node.exe',
  'D:\\Projects\\loopback-next\\packages\\build\\bin\\run-prettier',
  '--write',
  'D:/Projects/loopback-next/packages/build/bin/run-prettier.js',
  'D:/Projects/loopback-next/packages/repository/src/repositories/legacy-juggler-bridge.ts'
]
```

> The output is good enough to understand what's failing. My comment was to print out an option to skip the failure so that the pre-commit hooks can be skipped.

It is the last line in the output:
```
husky > pre-commit hook failed (add --no-verify to bypass)
```
So far there does not seem to be a way to customize that message.
If you mean highliting `LINT_STAGED=false` option, I'm not sure we should also promote a permanent switch here. It is described in the docs and visible in pre-commit hook definition in `package.json`.

package.json
```diff
@@ -80,6 +82,7 @@
   "copyright.owner": "IBM Corp.",
   "husky": {
     "hooks": {
+      "pre-commit": "if-env LINT_STAGED=false && echo \"lint-staged disabled via LINT_STAGED env var\" || lint-staged",
```
bajtos, line 85: Let's use `0` instead of `false`, it's shorter to type :)

```suggestion
      "pre-commit": "if-env LINT_STAGED=0 && echo \"lint-staged disabled via LINT_STAGED env var\" || lint-staged",
```

.lintstagedrc
```diff
@@ -0,0 +1,9 @@
+{
```
raymondfeng, line 1: Can we switch to `lint-staged.config.js` as the default config for `lint-staged`. The JS style gives us more flexibility.

package.json
InvictusMB: Updated

.lintstagedrc
InvictusMB: I'd rather not. But it's only my personal preference to introduce more flexibility when needed. Feel free to incorporate the changes from this PR into #5462.

raymondfeng:
@InvictusMB Thanks for the patience!

Issue #5550, opened 2020-05-25
RFC: Reorganize code & docs along abstraction levels


bajtos:
In my view, loopback-next is currently a mixed bag of low-level building blocks like `@loopback/metadata` and high-level framework packages like `@loopback/core` and `@loopback/rest`. Our documentation is often describing concepts from different abstraction layers in the same place, which makes it IMO more difficult for newcomers to learn how to use the framework.

I am proposing to reorganize our monorepo structure and the documentation in such way, that the documentation for framework consumers describes primarily framework-level APIs and deemphasizes the lower-level building blocks.

For example, we should stop referring to `@loopback/context` in our documentation and tell users to import DI/IoC features directly from `@loopback/core`. This way new-to-intermediate users don't need to understand the complexity of how LoopBack is composed internally. Of course, we should provide enough information for expert users to allow them to understand our building blocks and how they are composed together, but that should be in the part of the documentation that's clearly aiming at very advanced developers.

I am proposing the following rule of thumb: if a higher-level package  is re-exporting all public APIs from another package (e.g. `@loopback/core` re-exports `@loopback/context`) or is encapsulating another package as a building block (e.g. `@loopback/rest` uses `@loopback/express`), then the later package should be moved to "building blocks" and our documentation should be updated to reference the former "framework" package in most places.

Proposed taxonomy of monorepo packages (not a complete list):

**Framework**

- `@loopback/core`
- `@loopback/rest` and related extensions (rest-explorer, rest-crud)
- `@loopback/repository`
- `@loopback/boot` and booter extensions 
- `@loopback/authentication`, `@loopback/authorization`, `@loopback/security`

I am proposing to move these packages from `packages/{name}` to `framework/{name}`.

**Building blocks**

- `@loopback/context` (and its dependencies)
- `@loopback/http-server`, `@loopback/express`, etc.
- `@loopback/openapi-v3`

I am proposing to move these packages from `packages/{name}` to `foundation/{name}`. (I am not entirely happy with the name `foundation`, feel free to propose a better one.)

The documentation for this packages should primarily aim at people using these packages *outside* of LoopBack context, see the concept of microsites I mentioned in https://github.com/strongloop/loopback-next/issues/5113.

**Utilities**

- `@loopback/build`
- `@loopback/testlab`
- `@loopback/eslint-config`
- `@loopback/http-caching-proxy`
- `@loopback/model-api-builder`

[...]

PR #5618, opened 2020-05-31
refactor: move non-core booters to their own packages

raymondfeng:
Inspired by the discussion at https://github.com/strongloop/loopback-next/issues/5550, this PR cleans up `@loopback/boot` by moving non-core booters to their own packages.

The current `@loopback/boot` is becoming a kitchen sink for all booters and it drags in unnecessary dependencies to prevent itself from being a standalone package.

## Checklist


bajtos (dismissed):
+1 to make `@loopback/boot` more lightweight.

-1 to make other existing packages more heavy weight.

So far, packages likes `@loopback/repository` and `@loopback/rest-crud` can be used independently from `@loopback/boot`, allowing applications to define their own bootstrapping strategy. I'd like to preserve that behavior.

If we want to extract the booter to their own packages, then please do that literally and create new packages to host the booters. For example: `@loopback/repository-booters` and `@loopback/rest-crud-booters`.

bajtos:
@raymondfeng can you please post a short overview of booters to show which booters are going to stay in `@loopbakc/boot` and which are going to be moved elsewhere?

raymondfeng:
> So far, packages likes @loopback/repository and @loopback/rest-crud can be used independently from @loopback/boot, allowing applications to define their own bootstrapping strategy. I'd like to preserve that behavior.

Please note that the booters are contributions to the boot extension point. I don't think having their own booters in the same package will interfere with its functions.

> If we want to extract the booter to their own packages, then please do that literally and create new packages to host the booters. For example: @loopback/repository-booters and @loopback/rest-crud-booters.

I wonder if it's over-engineering here. Consider the user experience - I would expect `RepositoryComponent` and `CrudRestComponent` to contribute their specific booters. Even if we create separate packages for booters, do you expect to have separate components and require all applications to import such components for `app.boot` to work?

The core booters left in `@loopback/boot` are:

- base-artifact.booter.ts		
- application-metadata.booter.ts	
- interceptor.booter.ts
- lifecyle-observer.booter.ts
- component-application.booter.ts
- controller.booter.ts	
- service.booter.ts

bajtos:
> > So far, packages likes @loopback/repository and @loopback/rest-crud can be used independently from @loopback/boot, allowing applications to define their own bootstrapping strategy. I'd like to preserve that behavior.
> 
> Please note that the booters are contributions to the boot extension point. I don't think having their own booters in the same package will interfere with its functions.
> 
> > If we want to extract the booter to their own packages, then please do that literally and create new packages to host the booters. For example: @loopback/repository-booters and @loopback/rest-crud-booters.
> 
> I wonder if it's over-engineering here. Consider the user experience - I would expect `RepositoryComponent` and `CrudRestComponent` to contribute their specific booters. Even if we create separate packages for booters, do you expect to have separate components and require all applications to import such components for `app.boot` to work?

I see your points. I agree there is a value in `RepositoryComponent`/`RepositoryMixin` providing a booter for repositories as part of the deal and that having a new package for each booter may be too much.

My main concern is about dependencies. As a user adding `@loopback/repository` to my `@loopback/core` based project using my custom bootstrapping conventions, I don't want to be forced to add a dependency on `@loopback/boot` that I am not going to use.

When I was designing Model API Builder/Booter, I solved this issue as follows:

- The package `@loopback/model-api-builder` provides type definitions and implementation helpers for Model API Builder extension point.
- Individual implementations (e.g. `@loopback/rest-crud`) depend on `@loopback/model-api-builder` directly, they treat model-api-builder as a building block.
- `@loopback/boot` depends on `@loopback/model-api-builder` too, and exposes a Model API Booter that queries the context for all extensions.

Can we perhaps adopt a similar solution for `@loopback/boot` too?

- Introduce a new package `@loopback/booter` to provide type definitions and implementation helpers for individual Booters. This package is meant to be used as a direct dependency.
- Move RepositoryBooter to `@loopback/repository`. Instead of adding a peer dependency on `@loopback/boot`, we can add a regular dependency on the new package `@loobpack/booter`.
- Unfortunately, there is no way how to express which `@loopback/boot` versions can be used together with `@loopback/repository`, because npm does not provide optional peer dependencies. (See also Rust Cargo's [Features](https://doc.rust-lang.org/cargo/reference/features.html) that allow consumers to pick features and thus sub-dependencies.) A possible solution is to check the compatibility at runtime: the base booter class can provide a version field, `@loopback/boot` can report an error when it encounters a booter with an incompatible version. I think we can leave this out for now, until there is a need to break the Booter contract in a backwards-incompatible way.

> 
> The core booters left in `@loopback/boot` are:
> 
> * base-artifact.booter.ts
> * application-metadata.booter.ts
> * interceptor.booter.ts
> * lifecyle-observer.booter.ts
> * component-application.booter.ts
> * controller.booter.ts
> * service.booter.ts

This looks reasonable :+1:

I have a question regarding the datasource booter. Both repositories (`@loopback/repository`) and service proxies (`@loopback/service-proxy`) need datasources. IIUC, your current proposal is to put the datasource booter to `@loopback/repository` only, therefore applications using `@loopback/service-proxy` without `@loopback/repository` will NOT be able to bootstrap datasources. I find that as too limiting and probably difficult to discover & understand for our users too. Can we find a better solution please?

raymondfeng:
> My main concern is about dependencies. As a user adding @loopback/repository to my @loopback/core based project using my custom bootstrapping conventions, I don't want to be forced to add a dependency on @loopback/boot that I am not going to use.

I think this is a good use case of `peerDependencies`. If the `@loopback/boot` is declared as a peer dependency (ideally optional peer dependency) for `@loopback/repository`, we'll have the following benefits:

- If another package uses `@loopback/repository`, it does not transitively drags in `@loopback/boot`
- When `@loopback/boot` is present for the application, booters from `@loopback/repository` will be discovered and executed for `app.boot`.

raymondfeng:
> I have a question regarding the datasource booter. Both repositories (@loopback/repository) and service proxies (@loopback/service-proxy) need datasources. IIUC, your current proposal is to put the datasource booter to @loopback/repository only, therefore applications using @loopback/service-proxy without @loopback/repository will NOT be able to bootstrap datasources. I find that as too limiting and probably difficult to discover & understand for our users too. Can we find a better solution please?

I agree. There are quite a bit common code that should be ideally extracted from `@loopback/repository`.

raymondfeng:
> Introduce a new package @loopback/booter to provide type definitions and implementation helpers for individual Booters. This package is meant to be used as a direct dependency.

I initially had `@loopback/bootstrap` for this purpose and it turned out making things unnecessarily complicated, especially for backward compatibility.

If `@loopback/boot` becomes a peer dependency for those packages that contribute booters, having interfaces and bootstrapper (extension point) is not bad and heavy.

raymondfeng:
@bajtos I have rebased this PR against the latest master. I also adopted the peerDependencies for `@loopback/boot`, which should remove the key concerns you had. PTAL.

bajtos:
> I have rebased this PR against the latest master. I also adopted the peerDependencies for `@loopback/boot`, which should remove the key concerns you had.

I am not convinced `peerDependencies` are addressing my concerns. Quoting from https://github.com/strongloop/loopback-next/pull/5618#issuecomment-661075195:

> My main concern is about dependencies. As a user adding `@loopback/repository` to my `@loopback/core` based project using my custom bootstrapping conventions, I don't want to be forced to add a dependency on `@loopback/boot` that I am not going to use.
> (...)
> npm does not provide optional peer dependencies

Let's say my project is using `@loopback/core` and `@loopback/repository` but I don't want to use `@loopback/boot`. 

- What is going to happen when I run `npm install`? I remember that some `npm` versions are printing a warning about missing peer dependencies. IIUC, npm v7 will install peer dependencies automatically again (see https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major).

- `@loopback/repository` is exporting `RepositoryBooter` that's extending `BaseArtifactBooter` from `@loopback/boot`. As a result, when runtime calls `require('@loopback/repository')` but also when TypeScript loads type definitions (`.d.ts` files) for `@loopback/repository`, then they will need to load `@loopback/boot` too and the operation is going to fail if `@loopback/boot` is not present. (This is based on my understanding of the problem, it would be best to create a small app in sandbox to verify.)

As I see it, your design is still forcing all `@loopback/repository` users to install `@loopback/boot` too.

>  If another package uses `@loopback/repository`, it does not transitively drags in `@loopback/boot`

I don't think that's true - if another package uses `@loopback/repository` and don't add `@loopback/boot`, then you will either get `npm install` warning, or npm (v7) is going to install it for you anyways. Also if I my understanding is correct, then that other package will not work at runtime if `@loopback/boot` is not installed.

> I initially had `@loopback/bootstrap` for this purpose and it turned out making things unnecessarily complicated, especially for backward compatibility.

Can you please elaborate more?

>  If `@loopback/boot` becomes a peer dependency for those packages that contribute booters, having interfaces and bootstrapper (extension point) is not bad and heavy.
> 
> There is an interesting connection between `boot` and `core`. Do we want to have `core-booters` in its own package?

Interesting. Isn't the idea of having `core-booter` in opposition to what we agreed for other packages, where e.g. `@loopback/repository` should include the booters too?

OTOH, maybe this could be a way how to solve the peer dependency problem: Remove core booters from `@loopback/boot` to make the boot package smaller in scope (and less likely to change in backwards incompatible ways). In that case it's less bad that we are forcing users of packages like `@loopback/repository` to install `@loopback/boot` even if they are not going to use it. It's not ideal, but better than now.




> > I have a question regarding the datasource booter. Both repositories (@loopback/repository) and service proxies (@loopback/service-proxy) need datasources. IIUC, your current proposal is to put the datasource booter to @loopback/repository only, therefore applications using @loopback/service-proxy without @loopback/repository will NOT be able to bootstrap datasources. I find that as too limiting and probably difficult to discover & understand for our users too. Can we find a better solution please?
> 
> I agree. There are quite a bit common code that should be ideally extracted from `@loopback/repository`.

Yeah. So what's your proposal and the plan for this pull request? Are we going to force `@loopback/service-proxy` users to install `@loopback/repository` too? If that's the case then please make sure the project templates are handling this requirement. Test scenario:

1. Run `lb4 app`, select "enable services" feature but de-select "enable repositories" feature
2. Add a datasource using e.g. REST connector
3. Add a service via `lb4 service`
4. Run the app, the service must be picked up during boot

raymondfeng:
@bajtos Thank you for the input. I have refactored common types/interfaces for booters into `@loopback/booter` so that other packages can use it as the peer dependency to implement custom booters.

PTAL.

jannyHou (commented):
Following discussions, the latest refactor looks reasonable to me 👍  Had a chat with Raymond, the next step could be extracting the datasource booter to a standalone package.

bajtos:
Thank you @raymondfeng for addressing my comments. I am afraid I won't be able to continue the review in a timely manner and at the same time, I don't want to block progress. I am ok to leave the final review and approval to @jannyHou.

One use case to consider for testing - I think it's showing that the current design of dependencies & peer dependencies is not going to work as desired.

The idea is to create a small application that depends only on `@loopback/core` and `@loopback/repository` (see https://github.com/strongloop/loopback-next/pull/5618#issuecomment-702579136 for longer explanation). Because the new version of `@loopback/repository` is not published to npm yet, we need to use a local copy. However, we also must not use `lerna` to setup dependencies, because lerna is not reporting missing peer dependencies as `npm` does.

1. Pack the repository package:

    ```
    $ (cd packages/repository && npm pack)
    ```

2. Create a new sandbox directory `sandbox/test` and put the following `package.json` file there:

    ```json
    {
      "name": "test",
      "dependencies": {
        "@loopback/core": "^2.11.0",
        "@loopback/repository": "file:../../packages/repository/loopback-repository-3.1.0.tgz"
      }
    }
   ```

3. Install dependencies in our test project:

    ```
    $ (cd sandbox/test && npm i)
    ```

4. Observe a warning about a missing peer dependency:

    ```
    npm WARN @loopback/repository@3.1.0 requires a peer of @loopback/booter@^1.0.0 but none is installed. You must install peer dependencies yourself.
    ```

PR #5628, opened 2020-06-01
feat(context): use a faster UID generator

bajtos:
Replace `uuid/v4` with `hyperid` for auto-generated unique identifiers used in Context name and Interceptor binding keys.

This change improves the performance of `@loopback/rest` from handling ~2.4k requests/second to ~3.5k requests/second on average. The average latency is improved from 3.8ms to 2.6ms.

---

I considered several alternatives:
- `uuid/v1`, which is about 5x faster than `uuid/v4`
- a hand-written counter-based solution that's 6x faster than `uuid/v1`, but produces IDs that are not universally unique, which could be a problem in Cloud Native enviroment
- [hyperid](https://www.npmjs.com/package/hyperid), which is 30x faster than `uuid/v4` and only ~9% slower than my counter-based solution, while preserving the "universally unique" property of the generated ids.

You can find more details including example ID values here: https://gist.github.com/bajtos/6343512b171156872d3b701b2c10a682

---

Discussion point: I am changing the format of generated ids. I am not sure if that format was a part of our public contract and therefore this PR should be treated as a breaking change? Personally, I think the id format is an internal implement detail that nobody should rely on.

## Checklist


raymondfeng:
@bajtos Nice fix. 

A few more things to clarify:

1. The `name` is for information/debugging purpose. It's uniqueness is NOT required by the functionality.

2. We should pay more attentions to dynamically created contexts such as `Request` and `InvocationContext` to make them as light as possible.

bajtos:
Thank you @raymondfeng for the clarification.

> 1. The `name` is for information/debugging purpose. It's uniqueness is NOT required by the functionality.

Well, I can see how it can be useful to have a unique name. For example when reading through debug logs aggregated from a service mesh, it would be rather unhelpful if the same id was used for different context instances.

Considering that `hyperid` is only ~10% slower than a non-unique counter-based solution, I think it's fine to use that module?

> 2. We should pay more attentions to dynamically created contexts such as `Request` and `InvocationContext` to make them as light as possible.

Spot on 👍 ✖️ 💯 

Do you have any more comments on the proposed changes? Do they LGTY?

raymondfeng:
I read the `hyperid` code which uses a `v4` uuid + counter. It should be good enough for now.

Issue #5113, opened 2020-04-15
Documentation improvements 2020Q3


dhmlau:
We'd like to use this issue to capture how we can improve documentation. 

From the discussion with @raymondfeng and @bajtos: 

### Adding new documentation pages 
- Add/modify useful LB3 docs to LB4 docs #4848
   - Add "Running and Debugging apps" #4844 
- how to use @loopback/context outside of LB
- how to use @loopback/http-server outside of LB
- how to use @loopback/testlab both inside and outside of LB
- how to build microservice using LoopBack
- What users can do after scaffolding the LB app
   - potentially after the project template will contain the link to this. Related to #4981
- Add a page for extensions by the community
    - e.g. https://loopback.io/doc/en/lb3/Community-connectors.html.

### Reorganizing the documentation to make content easier to find
- [x] Rework loopback.io navbar 
   - possibly find a better place for "GraphQL" and "Test your APIs". 
   - possibly add blog, github, slack icons on the top right
   - details to be discussed
- [x] Reorganize sidebar: https://github.com/strongloop/loopback-next/issues/5295
- Hide `@loopback/openapi-v3` from docs, examples and CLI templates #5692

### Adopt the four-quandrant documentation system

- [x] docs: improve the structure of "Behind the scenes" sidebar + find a better name #5776
- [x] docs: improve the structure of "Usage scenarios" (how-to guides) sidebar #5777
- [ ] docs: move API references to tsdoc #5778
- [ ] docs: move recipes to how-to guides #5783
- [ ] docs: find a better place for "Extending LoopBack 4" in the sidebar [SPIKE] #5785
- [x] docs: untangle connector reference into reference, how-to guides and tutorials #5786
- [ ] docs: add a high-level framework overview that's easy to find #5788

Related: https://github.com/strongloop/loopback-next/issues/4717

dhmlau:
@achrinza, would like to get your inputs on this. Thanks.

achrinza:
I think this is a great idea, and would possibly help users better find the information they're looking for. It'll also make the responsibilities of each LoopBack package clearer.

To add on, I think there's 2 main, overreaching user stories that we need the docs to cater towards:

1. I want to create an API using LoopBack 4

2. I want to use individual LoopBack 4 components (in an existing application)

So it may be a good idea to separate the docs as such.

To make it even clearer, the [Overview](https://loopback.io/doc/en/lb4/index.html) page should show the users these 2 options and redirect them to the correct sections of the docs.

[...]

PR #5630, opened 2020-06-01
docs: dynamic datasource, model, repository, and controller

hacksparrow:
Describe how to create datasources, models, repositories, and
controllers at runtime. Part of addressing https://github.com/strongloop/loopback-next/issues/4296.

First half addressed by https://github.com/strongloop/loopback-next/pull/5591.

### DIFF
--- a/docs/site/Controllers.md
+++ b/docs/site/Controllers.md
@@ -196,6 +196,25 @@ export class HelloController {
 - `@param.query.number` specifies in the spec being generated that the route
   takes a parameter via query which will be a number.
 
+## Class factory to allow parameterized decorations
+
+Since decorations applied on a top-level class cannot have references to
+variables, you can create a class factory that allows parameterized decorations
+as shown in the example below.
+
+```ts
+function createControllerClass(version: string, basePath: string) {
+  @api({basePath: `${basePath}`})
+  class Controller {
+    @get(`/${version}`) find() {}
+  }
+}
+```
+
+For a complete example, see
+[parameterized-decoration.ts](https://github.com/strongloop/loopback-next/blob/master/examples/context/src/parameterized-decoration.ts)
+.
+
 ## Handling Errors in Controllers
 
 In order to specify errors for controller methods to throw, the class
@@ -269,3 +288,34 @@ export class HelloController {
   }
 }
 ```
+
+## Creating Controllers at Runtime
+
+A controller can be created for a model at runtime using the
+`defineCrudRestController` helper function from the `@loopback/rest-crud`
+package. It accepts a Model class and a `CrudRestControllerOptions` object.
+Dependency injection for the controller has to be configured by applying the
+`inject` decorator manually as shown in the example below.
+
+```ts
+const basePath = '/' + bookDef.name;
+const BookController = defineCrudRestController(BookModel, {basePath});
+inject(repoBinding.key)(BookController, undefined, 0);
+```
+
+The controller is then attached to the app by calling the `app.controller()`
+method.
+
+```ts
+app.controller(BookController);
+```
+
+The new CRUD REST endpoints for the model will be available on the app now.
+
+If you want a customized controller, you can create a copy of
+`defineCrudRestController`'s
+[implementation](https://github.com/strongloop/loopback-next/blob/00917f5a06ea8a51e1f452f228a6b0b7314809be/packages/rest-crud/src/crud-rest.controller.ts#L129-L269)
+and modify it according to your requirements.
+
+For details about `defineCrudRestController` and `CrudRestControllerOptions`,
+refer to the [@loopback/rest-crud API documentation](./apidocs/rest-crud.html).
--- a/docs/site/DataSources.md
+++ b/docs/site/DataSources.md
@@ -52,3 +52,32 @@ export class DbDataSource extends juggler.DataSource {
   }
 }
 ```
+
+### Creating a DataSource at Runtime
+
+A datasource can be created at runtime by creating an instance of
+`juggler.DataSource`. It requires a name for the datasource, the connector, and
+the connection details.
+
+```ts
+import {juggler} from '@loopback/repository';
+const dsName = 'bookstore-ds';
+const bookDs = new juggler.DataSource({
+  name: dsName,
+  connector: require('loopback-connector-mongodb'),
+  url: 'mongodb://sysop:moon@localhost',
+});
+await bookDs.connect();
+app.dataSource(bookDs, dsName);
+```
+
+For details about datasource options, refer to the [DataSource
+documentation])(https://apidocs.strongloop.com/loopback-datasource-juggler/#datasource)
+.
+
+Attach the newly created datasource to the app by calling `app.dataSource()`.
+
+{% include note.html content="
+The `app.datasource()` method is available only on application classes
+with `RepositoryMixin` applied.
+" %}
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -112,6 +112,52 @@ export class Customer {
 }
 ```
 
+## Defining a Model at runtime
+
+Models can be created at runtime using the `defineModelClass()` helper function
+from the `@loopback/repository` class. It expects a base model to extend
+(typically `Model` or `Entity`), followed by a `ModelDefinition` object as shown
+in the example below.
+
+```ts
+const bookDef = new ModelDefinition('Book')
+  .addProperty('id', {type: 'number', id: true})
+  .addProperty('title', {type: 'string'});
+const BookModel = defineModelClass<typeof Entity, {id: number; title?: string}>(
+  Entity, // Base model
+  bookDef, // ModelDefinition
+);
+```
+
+You will notice that we are specifying generic parameters for the
+`defineModelClass()` function. The first parameter is the base model, the second
+one is an interface providing the TypeScript description for the properties of
+the model we are defining. If the interface is not specified, the generated
+class will have only members inherited from the base model class, which
+typically means no properties.
+
+In case you need to use an existing Model as the base class, specify the Model
+as the base class instead of `Entity`.
+
+```ts
+// Assuming User is a pre-existing Model class in the app
+import {User} from './user.model';
+import DynamicModelCtor from '@loopback/repository';
+const StudentModel = defineModelClass<
+  typeof User,
+  // id being provided by the base class User
+  {university?: string}
+>(User, studentDef);
+```
+
+If you want make this new Model available from other parts of the app, you can
+call `app.model(StudentModel)` to create a binding for it.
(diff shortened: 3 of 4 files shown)


docs/site/Repositories.md
```diff
@@ -300,6 +300,77 @@ Please See [Testing Your Application](Testing-your-application.md) section in
...
+
+const BookRepositoryClass = defineRepositoryClass<
+  typeof BookModel,
+  MyRepoBase<BookModel, typeof BookModel.prototype.id, BookRelations>
+>(BookModel, MyRepoBase);
+```
+
+#### 2. Using a Repository mixin
+
+Create a repository mixin with your customization as shown in the "
```
dhmlau, line 346: Suggestion: since it will be a link, i think we don't need the double quotes.

docs/site/Repositories.md
```diff
@@ -300,6 +300,77 @@ Please See [Testing Your Application](Testing-your-application.md) section in
...
+
+Dependency injection has to be configured for the datasource as shown below.
+
+```ts
+inject(`datasources.${dsName.name}`)(BookRepository, undefined, 0);
+const repoBinding = app.repository(BookRepository);
+```
+
+Note, the `app.repository()` method will be available only on `RepositoryMixin`
+apps
```
dhmlau, line 372: Application class?

hacksparrow: Nah, it's not available on the `Application` class, it has to be composed with `RepositoryMixin`.

dhmlau: Right. I was thinking whether there's a better description than "RepositoryMix apps". I think it can be read it as Application class with RepositoryMixin?

docs/site/Controllers.md
```diff
@@ -269,3 +269,34 @@ export class HelloController {
...
+
+The new CRUD REST endpoints for the model will be available on the app now.
+
+If you want a customized controller, you can create a copy of
+`defineCrudRestController`'s
+[implementation](https://github.com/strongloop/loopback-next/blob/00917f5a06ea8a51e1f452f228a6b0b7314809be/packages/rest-crud/src/crud-rest.controller.ts#L129-L269)
+and modify it according to your requirements.
+
+For details about `defineCrudRestController` and `CrudRestControllerOptions`,
+refer to the [@loopback/rest-crud API documentation](./apidocs/rest-crud.html).
```
raymondfeng, line 302: Maybe we should mention Class factory to allow parameterized decorations per https://loopback.io/doc/en/lb4/core-tutorial-part10.html (Class factory to allow parameterized decorations).

docs/site/Model.md
```diff
@@ -112,6 +112,32 @@ export class Customer {
...
+
+```ts
+import DynamicModelCtor from '@loopback/repository';
+// Assuming User is a pre-existing Model class in the app
+const StudentModel = defineModelClass<
+  typeof User,
+  {id: number; university?: string}
+>(User, studentDef);
+```
+
```
raymondfeng, line 140: We should mention https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/mixins/repository.mixin.ts#L167

docs/site/Controllers.md
bajtos: +1

docs/site/DataSources.md
```diff
@@ -52,3 +52,25 @@ export class DbDataSource extends juggler.DataSource {
   }
 }
 ```
+
+### Creating a DataSource at Runtime
+
+A datasource can be created at runtime by creating an instance of
+`juggler.DataSource`. It requires a name for the datasource, the connector, and
+a connection url as shown below.
```
bajtos, line 60: I find this misleading. The configuration object accepts any valid fields, e.g. `host`, `port` and `user`. The connection url is just one of the many valid ways how to configure a datasource.

Please clarify.

I think there is no need to repeat all details here, we should instead add a link to an existing documentation page showing different datasource configuration options.

docs/site/DataSources.md
```diff
@@ -52,3 +52,25 @@ export class DbDataSource extends juggler.DataSource {
...
+  connector: require('loopback-connector-mongodb'),
+  url: 'mongodb://sysop:moon@localhost',
+});
+await bookDs.connect();
+app.dataSource(bookDs, dsName);
+```
+
+To use the newly created datasource, call its `.connect()` method and attach it
+to the app using `app.dataSource()` method. Note, this method will be available
+only on `RepositoryMixin` apps.
```
bajtos, line 76: There is no need to call `.connect()` method, datasources are starting the connection process automatically on creation. Please improve the text to be more accurate.

docs/site/Model.md
```diff
@@ -112,6 +112,32 @@ export class Customer {
...
+  bookDef,
+);
+```
+
+In case you need to use an existing Model as the base class, specify the Model
+as the base class instead of `Entity`.
+
+```ts
+import DynamicModelCtor from '@loopback/repository';
+// Assuming User is a pre-existing Model class in the app
```
bajtos, line 134:
```suggestion
// Assuming User is a pre-existing Model class in the app
import {User} from './user.model';
```

docs/site/Model.md
```diff
@@ -112,6 +112,32 @@ export class Customer {
...
+const StudentModel = defineModelClass<
+  typeof User,
+  {id: number; university?: string}
```
bajtos, line 137: Typically, the primary key (`id` property) is provided by the base class (`User`), therefore it's not necessary to repeat its definition in subclasses.

```suggestion
  {university?: string},
```

docs/site/Model.md
```diff
@@ -112,6 +112,32 @@ export class Customer {
 }
 ```
 
+## Defining a Model at runtime
+
+Models can be created at runtime using the `defineModelClass()` helper function
+from the `@loopback/repository` class. It expects a base model to extend
+(typically `Model` or `Entity`), folowed by the model definition object as shown
+in the example below.
```
bajtos, line 120: Can you please explain what is the second type parameter for?

docs/site/Repositories.md
```diff
@@ -300,6 +300,77 @@ Please See [Testing Your Application](Testing-your-application.md) section in
...
```
bajtos, line 372: Per the discussion in https://github.com/strongloop/loopback-next/pull/5630/files#r433340133, I am proposing the following:

```suggestion
Note, the `app.repository()` method is available only application classes with `RepositoryMixin` applied.
```

Please repeat this comment for Models and Repositories too.

docs/site/DataSources.md
hacksparrow: Updated to make the statement more generic and linked to https://apidocs.strongloop.com/loopback-datasource-juggler/#datasource.

There's no LB4 doc with details about `DataSource` options. Ideally it should be on https://loopback.io/doc/en/lb4/DataSources.html.

docs/site/Controllers.md
hacksparrow: Not really convinced. If I am understanding correctly, the content at the link describes a general programming pattern, not something specific to Controllers. If that's the case, we should avoid adding non-specific content to avoid noise.

docs/site/DataSources.md
```diff
@@ -71,6 +71,10 @@ await bookDs.connect();
...
 
-To use the newly created datasource, call its `.connect()` method and attach it
-to the app using `app.dataSource()` method. Note, this method will be available
-only on `RepositoryMixin` apps.
+For details about datasource options, refer to the
+[DataSource documentation])(https://apidocs.strongloop.com/loopback-datasource-juggler/#datasource)
+.
+
+Attach the newly created datasource to the app by calling `app.dataSource()`.
+Note, the `app.repository()` method is available only on application classes
```
bajtos, line 79:
```suggestion
Note, the `app.dataSource` method is available only on application classes
```

Have you considered using our `note.html` alert? https://loopback.io/doc/en/contrib/alerts.html

docs/site/DataSources.md
bajtos: > There's no LB4 doc with details about `DataSource` options. Ideally it should be on https://loopback.io/doc/en/lb4/DataSources.html.

Perhaps you can open a new issue for that and add it to "Documentation improvements" epic #5113?

docs/site/Model.md
```diff
@@ -116,8 +116,9 @@ export class Customer {
 
 Models can be created at runtime using the `defineModelClass()` helper function
 from the `@loopback/repository` class. It expects a base model to extend
-(typically `Model` or `Entity`), folowed by the model definition object as shown
-in the example below.
+(typically `Model` or `Entity`), followed by a `ModelDefinition` object as shown
+in the example below. The `ModelDefinition` object is an abstraction for
+specifying the various attributes of a LoopBack model.
```
bajtos, line 121: Sorry for not being clear enough.

> Can you please explain what is the second type parameter for?

In my comment ☝️, I was referring to the second item in the list of generic parameters for `defineModelClass`, i.e. the inline anonymous interface `{id: number; title?: string}`.

Now that I am looking at the example below again, I think may be good to show how `bookDef` is created too.

```ts
const bookDef = new ModelDefinition('Book')
  .addProperty('id', {type: 'number', id: true})
  .addProperty('title', {type: 'string'});
const BookModel = // ...
```

docs/site/DataSources.md
hacksparrow: Opened yesterday - https://github.com/strongloop/loopback-next/issues/5646.

docs/site/DataSources.md
hacksparrow: Did for other instances too.

hacksparrow:
@bajtos thanks for the feedback, applied them.

docs/site/Model.md
bajtos: I am afraid this comment is still not addressed.

As a user not familiar with defineModelClass, I don't understand what properties it should describe and why. Where is id and title coming from, why they have the particular type (number, string) and why title is optional while id is required?

I am looking for an explanation that
- the interface is providing TypeScript description for the properties of the model we are defining
- when omitted, the class will have only members inherited from the base model class, which typically means no properties

docs/site/Controllers.md
raymondfeng: The class factory is part of the programming model pattern and it's being asked by our users. It gives our users more control. The change is applying to `Controllers.md` and I don't think that's noise.

hacksparrow: > The change is applying to `Controllers.md`

Got the context, I thought it was about runtime controllers.

Added **Class factory to allow parameterized decorations** at an appropriate location.

berlingoqc:
Hi will there be a section about creating dynamic relation for repo with inclusion resolver ? Is this even possible ?

PR #5642, opened 2020-06-02
chore: add generating docs/site/MONOREPO.md file

spirAde:
This makes it possible to generate MONOREPO.md file automatically with sorting of entries
Comment: https://github.com/strongloop/loopback-next/pull/5485#discussion_r429221869
## Checklist


spirAde:
This idea was mentioned in the comment https://github.com/strongloop/loopback-next/pull/5485#discussion_r429221869.

If you are still considering this functionality useful I added generating `MONOREPO.md` file. I created another file `MONOREPO-draft.md` while we discuss the solution. And I ran into with some concerns here:

1) Difference between description column text in `MONOREPO.md` file and description in `package.json`. Can I use package.json description field text to put it in the description column in `MONOREPO.md` file? Or I should take column description and update all `package.json` where it requires? Some packages.json doesn't have a description field at all and as result in the `MONOREPO-draft.md` file.

2) I used `markdown-table` package for the generating markdown table. Is it ok from your perspective?

3) Current `MONOREPO.md` file has the mention of sorting packages: `PLEASE KEEP THE TABLE ROWS SORTED ALPHABETICALLY BY PACKAGE NAME`. But it's wrong real sorted table with rows sorted alphabetically by package name in `MONOREPO-draft.md` file. Which one is preferable?

raymondfeng:
@spirAde Is it possible to patch only? For example, do not change existing rows and simply add a new row for a new package? This way, we can preserve manual updates.

spirAde:
@raymondfeng I guess yes it's possible. I can parse the current `MONOREPO.md` file line by line and check if the current row is a table row with information about the package. Does it make sense? Could you also answer other questions, especially the last question, please

spirAde:
@raymondfeng What is your opinion about the best way to match packages between `MONOREPO.md` file and packages from `lerna/project package`. I supposed that the best way is the `package.name` field, but some rows in the `MONOREPO.md` table doesn't contain this name, it contains `_(private)_` value.

spirAde:
@raymondfeng What is your opinion about the best way to match packages between MONOREPO.md file and packages from lerna / project package. I supposed that the best way is the package.name field, but some rows in the MONOREPO.md table doesn't contain this name, it contains _ (private) _ value.

raymondfeng:
> @raymondfeng What is your opinion about the best way to match packages between MONOREPO.md file and packages from lerna/project package. I supposed that the best way is the package.name field, but some rows in the MONOREPO.md table doesn't contain this name, it contains _(private)_ value.

@spirAde Feel free to normalize the columns as needed. For example:

Column 1: relative location of the package, such as `packages/core`. 
Column 2: npm package name
Column 3: private/public flag
Column 4: description

spirAde:
@raymondfeng Awesome, thank you

raymondfeng:
> Difference between description column text in MONOREPO.md file and description in package.json. Can I use package.json description field text to put it in the description column in MONOREPO.md file? Or I should take column description and update all package.json where it requires? Some packages.json doesn't have a description field at all and as result in the MONOREPO-draft.md file.

Let's take this opportunity to refine the `description` in `package.json`. Feel free to pick the better one to update package.json as the source of truth.

> I used markdown-table package for the generating markdown table. Is it ok from your perspective?

Sure.

> Current MONOREPO.md file has the mention of sorting packages: PLEASE KEEP THE TABLE ROWS SORTED ALPHABETICALLY BY PACKAGE NAME. But current MONOREPO.md table has wrong order. MONOREPO-draft.md file has correct sequence(sorted by package.name field). Which order is preferable?

Personally, I think sorting them by the package location is a better approach.

bin/build-monorepo-file.js
```diff
@@ -0,0 +1,79 @@
...
+  const packages = await getSortedPackages();
+  const markdownTable = buildTable(packages);
+
+  const content = [
+    '# Monorepo overview',
+    '',
+    'The [loopback-next](https://github.com/strongloop/loopback-next) repository uses',
+    '[lerna](https://lernajs.io/) to manage multiple packages for LoopBack 4.',
+    '',
+    '<!-- PLEASE KEEP THE TABLE ROWS SORTED ALPHABETICALLY BY PACKAGE NAME -->',
```
bajtos, line 57: Now that the content of this file is auto-generated, we should change this comment to ask people to use your new script to update the contents in this file. The sorting no longer matters in this comment, because the script will take care of keeping consistent order.

spirAde:
@raymondfeng @bajtos Since we decided the `package.json` description is source of the truth I chose the most appropriate description between `MONOREPO.md` file and description field in `package.json` file and updated them. Fixed some text. Could you check, please. Thanks in advance

PR #5598, opened 2020-05-28
Auth doc refactor

jannyHou:
Links to https://github.com/strongloop/loopback-next/issues/5294

The first auth doc refactor PR: overview page

Please review file **docs/site/Authentication-overview.md ONLY**, the others files are merged from other PRs.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5643, opened 2020-06-02
Auth doc component

jannyHou:
Sub PR for #5598

Update the authentication-component page with new layout.

Given the length of this page, I break it down into sub pages for each topic. I run doc preview to make sure links all work.

Screenshots:

New menu:
Screen Shot 2020-06-05 at 11 19 45 AM

Pages:

**Overview**
Screen Shot 2020-06-05 at 11 28 09 AM
**Decorator**
Screen Shot 2020-06-05 at 11 28 18 AM
**Action**
Screen Shot 2020-06-05 at 11 28 23 AM
**Strategy**
Screen Shot 2020-06-05 at 11 28 27 AM
**Option**
Screen Shot 2020-06-05 at 11 28 32 AM

## Checklist


docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,251 @@
...
+    } catch (error) {
+      this.reject(context, error);
+    }
+  }
+}
+```
+
+</details>
+
+By default, authentication is **not** part of the sequence of actions, so you
```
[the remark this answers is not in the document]
jannyHou: good catch!

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,251 @@
...
+@inject(AuthenticationBindings.AUTH_ACTION)
+protected authenticateRequest: AuthenticateFn,
+```
+
+The binding key `AuthenticationBindings.AUTH_ACTION` gives us access to the
+authentication function `authenticateRequest` of type `AuthenticateFn` provided
+by `AuthenticateActionProvider`.
+
+Now the authentication function `authenticateRequest` can be called in our
+custom sequence anywhere `before` the `invoke` action in order secure the
```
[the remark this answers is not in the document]
jannyHou: there shouldnt' be a `

docs/site/Authentication-component-options.md
```diff
@@ -0,0 +1,263 @@
...
+  }
+}
+```
+
+### Override Default Options In Authentication Decorator
+
+Specify overriding options in the `@authenticate` decorator only when necessary.
+
+In this example, we only specify an **overriding** option `gatherStatistics`
+with a value of `false` for the `/scareme` endpoint. We use the **default**
```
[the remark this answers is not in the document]
jannyHou: > we only specify an overriding option gatherStatistics: false for the /scareme endpoint.

👍 
> Does this mean if we have several options, we only need to specify the ones that are different from the default values?

Yes.

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,252 @@
+---
+lang: en
+title: 'Concept Two - Authentication Action'
```
dhmlau, line 3: Not sure if we need "concept one", "concept two". Personally I'd prefer not to. 
If the parent page mention there are 3 concepts you need to know for authentication in LoopBack, and list the three with links. I think this would be good.

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,252 @@
...
+---
+
+## Adding an Authentication Action to a Custom Sequence
+
+In a LoopBack 4 application with REST API endpoints, each request passes through
+a stateless grouping of actions called a [Sequence](Sequence.md).
+
+The default sequence which injects and invokes actions `findRoute`,
+`parseParams`, `invoke`, `send`, `reject` could be found in the
+[Todo example's sequence file](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/sequence.ts)
```
dhmlau, line 16:
```suggestion
[Todo example's sequence file](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/sequence.ts).
```

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,252 @@
...
+    } catch (error) {
+      this.reject(context, error);
+    }
+  }
+}
+```
+
+</details>
+
+By default, `authenticate` is **not** part of the sequence of actions, so you
```
dhmlau, line 83: `authenticate` == the authenticate action?

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,252 @@
...
+must create a custom sequence and add the authentication action.
+
+An authentication action `AuthenticateFn` is provided by the
+`AuthenticateActionProvider` class.
+
+`AuthenticateActionProvider` is defined as follows:
```
dhmlau, line 89: When reading the original documentation (before the refactoring), I was wondering why we need to include the source code of the provider. If we want to describe what the AuthenticateActionProvider does, would it be better to list them out?

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,252 @@
...
+authentication strategy **was** specified for this endpoint, its
+`authenticate(request)` function is called. If a user profile is returned, this
+means the user was authenticated successfully, and the user profile is added to
+the request context (via the `SecurityBindings.USER` binding); otherwise an
+error is thrown.
+
+Here is an example of a custom sequence which utilizes the `authentication`
+action.
+
+```ts
```
dhmlau, line 158: in the todo-jwt example, we are using the default sequence that comes with the scaffolded LB app. Maybe we use that instead? Or explain when we want to create a separate sequence?

docs/site/Authentication-component-action.md
jannyHou: yep, seems the word is confusing, any better suggestion? maybe "authenticate"?

docs/site/Authentication-component-action.md
jannyHou: I feel it aims to describe the provider with comments in code: there are 3 comments in the code snippet and the 1st one above strategy getter has many lines.

docs/site/Authentication-component-action.md
jannyHou: @dhmlau not really :p the todo-jwt example still uses the custom sequence with "authenticate" injected and invoked.

See [code](https://github.com/strongloop/loopback-next/blob/master/examples/todo-jwt/src/sequence.ts#L34).

docs/site/Authentication-component-action.md
jannyHou: I was trying to distinguish them from the "options" section cuz option is not really a concept. Hmm, maybe rename them to 

- Authentication Decorator
- Authentication Action
- Authentication Strategy
- Configuring Strategy Options

?

docs/site/Authentication-component-action.md
```diff
@@ -0,0 +1,254 @@
...
+Now the authentication function `authenticateRequest` can be called in our
+custom sequence anywhere before the `invoke` action in order secure the
+endpoint.
+
+There are two particular protocol-agnostic errors
+`AUTHENTICATION_STRATEGY_NOT_FOUND` and `USER_PROFILE_NOT_FOUND` which must be
+addressed in the sequence, and given an HTTP status code of 401 (UnAuthorized).
+
+It is up to the developer to throw the appropriate HTTP error code from within a
+custom authentications strategy or its custom services.
```
agnes512, line 219: Is L180 one of these options?

docs/site/Authentication-component-decorator.md
```diff
@@ -0,0 +1,101 @@
...
+
+or
+
+```ts
+@authenticate(metadata: AuthenticationMetadata)
+```
+
+The **strategyName** is the **unique** name of the authentication strategy.
+
+When the **options** object is specified, it must be relevant to that particular
```
agnes512, line 33: Nitpick: maybe wrap `strategyName` and `options`?

docs/site/Authentication-component-decorator.md
```diff
@@ -0,0 +1,101 @@
...
+available on the request context. You can obtain it via dependency injection by
+using the `SecurityBindings.USER` binding key.
+
+Parameters of the `@authenticate` decorator can be obtained via dependency
+injection using the `AuthenticationBindings.METADATA` binding key. It returns
+data of type `AuthenticationMetadata` provided by `AuthMetadataProvider`. The
+`AuthenticationStrategyProvider`, discussed in a later section, makes use of
+`AuthenticationMetadata` to figure out what **name** you specified as a
+parameter in the `@authenticate` decorator of a specific controller endpoint.
+
```
agnes512, line 95: I feel like this paragraph contains too much information. Is it possible to just provide minimum information to explain the mechanism?

For example ( I might not understand it correctly):
With  the binding key `AuthenticationBindings.METADATA`, the
`AuthenticationStrategyProvider`, which will be discussed in a later section, can figure out what strategy name is specified in the `@authenticate` decorator.

Cause I assume that at this point the read should know how binding key works, and `@authenticate` decorates the endpoint.

docs/site/Authentication-component-action.md
dhmlau: What i meant is that the sequence.ts is created by default when we're scaffolding the app. In the snippet here seems to guide the users to create a new sequence?

docs/site/Authentication-component-action.md
jannyHou: 👌 changed to "authenticate"

docs/site/Authentication-component-action.md
jannyHou: @dhmlau I rephrased it to be:

> By default, "authenticate" is **not** part of the sequence of actions, so you
must **modify the default sequence** and add the authentication action.

And

> Here is an example of a **modified** sequence which utilizes the `authentication`
action.

docs/site/Authentication-component-action.md
jannyHou: 👌 changed to `authenticate`

docs/site/Authentication-component-action.md
jannyHou: @agnes512 Yep that's correct.

docs/site/Authentication-component-decorator.md
jannyHou: @agnes512 Not really, people don't always read all docs in sequence...personally I feel this paragraph is good and necessary, so I tend to keep it as is. 
(but thank you for writing the draft 👍 )

Issue #5549, opened 2020-05-25
RFC: Adopt the four-quandrant documentation system


bajtos:
I am proposing to adopt the documentation system described at https://documentation.divio.com and successfully used by [Django](https://www.djangoproject.com) and others, and reorganize our documentation along the following four directions:
1. Tutorials
2. How-to guides
3. Reference
4. Background/Key concepts/Discussion/Explanation

![Overview of the documentation system](https://documentation.divio.com/_images/overview.png)

> Documentation needs to include and be structured around its four different functions: tutorials, how-to guides, technical reference and explanation. Each of them requires a distinct mode of writing. People working with software need these four different kinds of documentation at different times, in different circumstances - so software usually needs them all, and they should all be integrated into your documentation.

You can also watch the talk from PyCon Australia 2017 where this system was explained: https://youtu.be/t4vKPhjcMZg

In this issue, I'd like to keep the discussion at high-level and reach consensus whether such documentation system is something we all agree to asses and if we decide to implement it, then we will be all following it going forward.

An example of an existing doc page that's mixing different directions:
- [Key concepts >> Server](https://loopback.io/doc/en/lb4/Server.html) is a mixture of 
  - Some Background explanation of what is a Server.
  - How-to guides for customizing CORS, configuring REST API Explorer, etc.
  - Reference of [`rest` options](https://loopback.io/doc/en/lb4/Server.html#rest-options)

## Next steps

- [x] Discuss the proposal and reach consensus on whether we want to pursue this direction or not. If yes:
- [ ] A spike story to identify the first few steps of what will become a longer journey. For example, how to re-arrange the sidebar and existing pages, where to keep reference guides (inside `docs` or as tsdoc comments in code?), etc.
- [ ] Many small stories to review each existing doc page and rework it to follow the new system - extract How-to guides, extract Reference material, rework Explanations, and so on.

dhmlau:
@bajtos, I agree with you that some of our docs page are mixed with how-tos and explanation. 
Are you thinking of each page (if applicable) to have 4 sections, or there will be 4 categories in the docs? 

> small stories to review each existing doc page and rework it to follow the new system

+1

raymondfeng:
@bajtos In addition to what you proposed, I would like all of us to think about the following perspectives:

- Persona (Platform developer, API developer, Microservice developers, Integration Developer)
[...]

PR #5659, opened 2020-06-04
docs: add a new section on content organization

bajtos:
Based on the discussion in #5550, I am updating the documentation for developers with the definition of "building-blocks", a list of packages considered as building blocks, and finally instructions on how to treat these packages in our docs, CLI templates and examples.

After this PR is approved, I'll created follow-up stories to ensure proper use of the following packages:

- packages/context - already done in #5625
- packages/openapi-v3

The following foundation-level packages are already not depended on directly:
- packages/express
- packages/http-server
- packages/metadata
- packages/repository-json-schema

## Checklist

### DIFF
--- a/docs/site/DEVELOPING.md
+++ b/docs/site/DEVELOPING.md
@@ -286,6 +286,69 @@ LoopBack 4 documentation is hosted inside this monorepo in the
 This allows us to change both implementation and the documentation in a single
 pull request.
 
+### Organization of content
+
+LoopBack 4 has a highly modular design, the codebase is organized into dozens of
+packages. This arrangement provides great flexibility for package consumers,
+supporting many different ways how to compose individual packages into larger
+blocks. However, the growing number of packages also increases the complexity
+for LoopBack users, as they need to know which packages to choose and how to use
+them. As a result, the learning curve becomes very steep for new developers
+coming to LoopBack. Newcomers get quickly overwhelmed by the amount of concepts
+they need to understand and the different packages they need to know about.
+
+To prevent cognitive overload, we have categorized all monorepo packages into
+two groups:
+
+1. Foundation-level packages are lower-level packages that are typically not
+   consumed by LoopBack applications directly. Instead, there are higher-level
+   packages exposing the functionality and/or the public API of these building
+   blocks.
+
+   Examples:
+
+   - `@loopback/core` is re-exporting all public API provided by
+     `@loopback/context`.
+
+   - `@loopback/rest` is internally using `@loopback/http-server` to manage the
+     life cycle of an HTTP server.
+
+2. Framework-level packages are used directly by LoopBack applications and
+   provide API that LoopBack consumers use.
+
+The rule of thumb: a lower-level package is considered as foundation-level if
+
+- a higher-level package is re-exporting all public APIs of the lower-level
+  package (e.g. `@loopback/core` re-exports `@loopback/context`); or
+
+- another package is using the lower-level package as a implementation building
+  block (e.g. `@loopback/rest` uses `@loopback/express`).
+
+In our documentation, CLI templates and example applications, you should always
+refer to framework-level packages.
+
+Foundation-level packages should have their own documentation describing how to
+use the package independently of LoopBack, for example in their `README.md` file
+or in a dedicated section of [loopback.io](https://loopback.io).
+
+#### Foundation-level packages
+
+List of packages that are considered as building blocks:
+
+- [packages/context](https://github.com/strongloop/loopback-next/tree/master/packages/context)
+- [packages/express](https://github.com/strongloop/loopback-next/tree/master/packages/express)
+- [packages/http-server](https://github.com/strongloop/loopback-next/tree/master/packages/http-server)
+- [packages/metadata](https://github.com/strongloop/loopback-next/tree/master/packages/metadata)
+- [packages/openapi-v3](https://github.com/strongloop/loopback-next/tree/master/packages/openapi-v3)
+- [packages/repository-json-schema](https://github.com/strongloop/loopback-next/tree/master/packages/repository-json-schema)
+
+#### Framework-level packages
+
+All packages not listed above are considered as framework-level.
+
+We consider utilities like `@loopback/testlab` and example projects like
+`@loopback/todo` as framework-level too.
+
 ### Publishing changes
 
 To prevent documentation changes going live before the changes in the


bajtos:
Please note this is just small incremental step on a longer journey, I expect us to do more doc-related improvement in follow-up stories and pull requests, depending on how the discussions in #5113, #5550 and #5549 evolve.

docs/site/DEVELOPING.md
```diff
@@ -286,6 +286,46 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
+[loopback.io](https://loopback.io).
+
+The rule of thumb: if a higher-level package is re-exporting all public APIs
+from a lower-level package (e.g. `@loopback/core` re-exports
+`@loopback/context`) or is using another package as a implementation building
+block (e.g. `@loopback/rest` uses `@loopback/express`), then the lower-level
+package should be moved to "building blocks" and our documentation should be
+updated to reference the higher-level "framework" instead.
+
+List of packages that are considered as building blocks:
```
hacksparrow, line 320: And the rest are framework-level pacakges? If yes, we should mention that, else it feels like the list of framework-level pacakges is missing.

bajtos: Done in 69dcaa6

There is one thing that bugs me a bit: we also have utility packages like `@loopback/build` or `@loopback/testlab` that are technically not part of the framework, but they are not building blocks either in the sense that they should be mentioned in our framework documentation.

I guess "framework-level" is not the best term to use. Can we come up with a more accurate name?

dhmlau: What about the `extensions` and `examples` packages? Those seem more like an adds-on.

bajtos: > What about the `extensions` and `examples` packages? Those seem more like an adds-on.

Sure :+1:

What I am trying to accomplish here is to distinguish between "building blocks" packages that should not be referred to in our docs, CLI templates and example apps; and everything else. 

I used the term "framework-level" for "everything else", which is clearly misleading based on your feedback.

So, what would be a better term?

hacksparrow: > I guess "framework-level" is not the best term to use. Can we come up with a more accurate name?

They need not belong in the framework packages. They can be classified as helper/utility packages.

docs/site/DEVELOPING.md
```diff
@@ -286,6 +286,48 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
+writing docs. While this is good for capturing the knowledge in a text form, it
+also makes the documentation difficult to follow by people new to LoopBack.
+Newcomers get quickly overwhelmed by the amount of concepts they need to
+understand and all the different packages they need to know about.
+
+To prevent that issue, we have categorized all monorepo packages into two
+groups:
+
+1. Building-blocks are lower-level packages that are typically not consumed by
+   LoopBack projects directly, instead there is a higher-level package exposing
```
raymondfeng, line 302: `LoopBack projects` -> `LoopBack applications` or `LoopBack application projects`.

docs/site/DEVELOPING.md
```diff
@@ -286,6 +286,48 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
 pull request.
 
+### Organization of content
+
+LoopBack 4 is a large codebase organized into dozens of packages. We,
+developers, are often tempted to include a lot of details and wider context when
+writing docs. While this is good for capturing the knowledge in a text form, it
+also makes the documentation difficult to follow by people new to LoopBack.
+Newcomers get quickly overwhelmed by the amount of concepts they need to
+understand and all the different packages they need to know about.
```
raymondfeng, line 296: We should also mention the balance between `modular design` and `ease of use`.

- Module design leads to great flexibility and composability. As the number of modules grows, it also increases the complexity as developers need to know what to choose and how to use.
- We need to target high-level packages for application developers without dragging in lower-level details.

docs/site/DEVELOPING.md
```diff
@@ -286,6 +286,48 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
+
+To prevent that issue, we have categorized all monorepo packages into two
+groups:
+
+1. Building-blocks are lower-level packages that are typically not consumed by
```
raymondfeng, line 301: `not consumed` -> `not directly consumed` or `not explicitly consumed`?

raymondfeng: I wonder if we can find a better name for `building blocks`.  Maybe `foundation-level` packages to be symmetric with `framework-level`?

raymondfeng: Or `library-level`?

docs/site/DEVELOPING.md
raymondfeng: We may want to identify several personas of different interest/entry to LB here:

- Platform/extension developers
- Application developers

docs/site/DEVELOPING.md
bajtos: > `not consumed` -> `not directly consumed` or `not explicitly consumed`?

English is not my mother tongue, but I believe that my current version is grammatically correct while conveying the same meaning.

```
lower-level packages that are typically not consumed by LoopBack projects directly
```

bajtos: > I wonder if we can find a better name for `building blocks`. Maybe `foundation-level` packages to be symmetric with `framework-level`?

I like "foundation-level" + "framework-level".

In https://github.com/strongloop/loopback-next/pull/5659#discussion_r435239706, we were discussing with @hacksparrow that "framework-level" may not be the best description. However, I think with building blocks renamed to "foundation-level", the term "framework-level" is ok.

docs/site/DEVELOPING.md
bajtos: > We may want to identify several personas of different interest/entry to LB here:
>
> - Platform/extension developers
> - Application developers

Yeah, I agree that at some point, we need to step back and reorganize our docs to cater for different personas. As I see it, our current docs is mostly aimed at application developers. In this pull request, I am proposing how to make the docs easier to read for app developers. We can discuss how to better address other personas in https://github.com/strongloop/loopback-next/issues/5113.

docs/site/DEVELOPING.md
bajtos: Well, I think this applies to most extension projects too, but let's not dwell on details. I am going to use "LoopBack applications" for now, we can improve this later as we get a better picture about the documentation for extension authors.

docs/site/DEVELOPING.md
bajtos: I am adding the following clarification to the docs:

```
#### Framework-level packages

All packages not listed above are considered as framework-level.

We consider utilities like `@loopback/testlab` and example projects like
`@loopback/todo` as framework-level too.```

**Comment by bajtos** [comment]
@hacksparrow @dhmlau @raymondfeng thank you for the suggestions. I rewrote most of the new content to incorporate your feedback, does the new version LGTY now?

**Comment by hacksparrow** [comment]
How do we handle "breaking changes" in docs organization? For example a framework-level package underwent some refactoring and ended up becoming a foundation-level package.

File: docs/site/DEVELOPING.md (hunk 1 above)
Comment (hacksparrow) [reply docs/site/DEVELOPING.md:336]: > We consider utilities like `@loopback/testlab` and example projects like
`@loopback/todo` as framework-level too.

Doesn't feel right. Anything that LoopBack developers would not (maybe even never) use in their project feels very odd to be classified under framework-level packages. We need a third category - helpers, utility, along that line.

**Comment by bajtos** [comment]
> How do we handle "breaking changes" in docs organization? For example a framework-level package underwent some refactoring and ended up becoming a foundation-level package.

At the moment, we have a single version of docs only ("the latest").

I think we will eventually need to introduce versioning (e.g. docs for `@loopback/rest` version from 2020 vs. version from 2021`), especially once we introduce LTS versions (see https://github.com/strongloop/loopback-next/issues/4398). Perhaps such versioning would help?

In the current state, where there is a single version of docs only: When a framework-level package becomes foundation-level, I am envisioning the following process:
- docs, CLI templates, example apps are updated to framework-level APIs instead of the package that was moved to foundation-level
- if & where appropriate, we may add additional sections to explain how to migrate from the "old" to the "new" style - see e.g. https://loopback.io/doc/en/lb4/DataSource-generator.html#legacy-json-based-configuration

Is there any specific concern you have, @hacksparrow?

**Comment by hacksparrow** [comment]
My concern was looking at outdated references. Versioning should help.

> docs, CLI templates, example apps are updated to use framework-level APIs instead of referring to he package that was moved to foundation-level
> if & where appropriate, we may add additional sections to explain how to migrate from the "old" to the "new" style - see e.g. https://loopback.io/doc/en/lb4/DataSource-generator.html#legacy-json-based-configuration

Sounds good for now.

File: docs/site/DEVELOPING.md (hunk 1 above)
Comment (bajtos) [reply docs/site/DEVELOPING.md:336]: I guess this is subjective.

Let's land my PR as the first step, and then you can open a follow-up PR to discuss how to improve this initial categorization.

Comment (hacksparrow) [reply docs/site/DEVELOPING.md:336]: Sounds good.

File: docs/site/DEVELOPING.md [hunk 5]
```diff
@@ -286,6 +286,69 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
+#### Foundation-level packages
+
+List of packages that are considered as building blocks:
+
+- [packages/context](https://github.com/strongloop/loopback-next/tree/master/packages/context)
+- [packages/express](https://github.com/strongloop/loopback-next/tree/master/packages/express)
+- [packages/http-server](https://github.com/strongloop/loopback-next/tree/master/packages/http-server)
+- [packages/metadata](https://github.com/strongloop/loopback-next/tree/master/packages/metadata)
+- [packages/openapi-v3](https://github.com/strongloop/loopback-next/tree/master/packages/openapi-v3)
+- [packages/repository-json-schema](https://github.com/strongloop/loopback-next/tree/master/packages/repository-json-schema)
```
raymondfeng, line 343: This package depends on `@loopback/core` and `@loopback/repository`. It should be moved out.

docs/site/DEVELOPING.md
```diff
@@ -286,6 +286,69 @@ LoopBack 4 documentation is hosted inside this monorepo in the
...
+
+#### Foundation-level packages
+
+List of packages that are considered as building blocks:
+
+- [packages/context](https://github.com/strongloop/loopback-next/tree/master/packages/context)
+- [packages/express](https://github.com/strongloop/loopback-next/tree/master/packages/express)
+- [packages/http-server](https://github.com/strongloop/loopback-next/tree/master/packages/http-server)
+- [packages/metadata](https://github.com/strongloop/loopback-next/tree/master/packages/metadata)
+- [packages/openapi-v3](https://github.com/strongloop/loopback-next/tree/master/packages/openapi-v3)
```
raymondfeng, line 342: Not sure about this one either:
```
"dependencies": {
    "@loopback/core": "^2.7.1",
    "@loopback/repository-json-schema": "^2.4.3",
}
```

docs/site/DEVELOPING.md
bajtos: > This package depends on `@loopback/core` and `@loopback/repository`. It should be moved out.

Please note the criteria is not whether a package depends on something else from `@loopback/*` namespace, but whether LB application developers should depend on that package directly.

Both `@loopback/repository-json-schema` and `@loopback/openapi-v3` are not meant to be used directly. They are building blocks used by `@loopback/rest`. Applications are importing their APIs and functionality from `@loopback/rest`, as can be seen e.g. in [`examples/todo/src/controllers/todo.controller.ts`](https://github.com/strongloop/loopback-next/blob/master/examples/todo/src/controllers/todo.controller.ts)

PR #5668, opened 2020-06-05
feat(context): add strongly typed `on` and `once` methods

bajtos:
Declare `on` and `once` overload methods to describe event parameters in a strongly typed way:

```ts
context.on('bind' | 'unbind', ContextEventListener)
context.once('bind' | 'unbind', ContextEventListener)
binding.on('changed', BindingEventListener)
binding.once('changed', BindingEventListener)
view.on('refresh', listener);
view.once('refresh', listener);
// and so on
```

The change preserves the generic variant provided by `EventEmitter` too and thus is fully backwards compatible.

## Checklist

### DIFF
--- a/packages/context/src/binding.ts
+++ b/packages/context/src/binding.ts
@@ -201,7 +201,7 @@ export type BindingEvent = {
   /**
    * Event type
    */
-  type: string;
+  type: 'changed' | string;
   /**
    * Source binding that emits the event
    */
@@ -209,7 +209,7 @@ export type BindingEvent = {
   /**
    * Operation that triggers the event
    */
-  operation: string;
+  operation: 'tag' | 'scope' | 'value' | string;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/context/src/binding.ts
```diff
@@ -873,6 +873,34 @@ export class Binding<T = BoundValue> extends EventEmitter {
   }
 }
 
+export declare interface Binding {
```
raymondfeng, line 876: We use declare here, but not for Context.

raymondfeng: What’s the impact on tsdocs?

raymondfeng: Personally, I don’t like to have the same name for class and interface.

raymondfeng: As far as I understand, we can simply use `export interface Binding`. The interface will be merged with the `typing` perspective of `Binding` class.

bajtos: > As far as I understand, we can simply use `export interface Binding`. The interface will be merged with the `typing` perspective of `Binding` class.

It works like a charm 👍 

I am not sure why I used `export declare interface` here when I am already using `export interface` in `src/context.ts` 🤷 

I discovered that `ContextView` is emitting events too, I'll add types for those events in this PR too.

bajtos:
> What’s the impact on tsdocs?

It looks like our tsdoc processor is not merging class + interface declarations correctly 😠 In http://127.0.0.1:4001/doc/en/lb4/apidocs.context.html, I see `Binding` listed under `Classes` and `Interfaces` sections. Both entries point to the same URL http://127.0.0.1:4001/doc/en/lb4/apidocs.context.binding.html where only the interface members are described.

Screen Shot 2020-06-08 at 10 52 45

I tried to swap the order in which the class & interface is defined, but that did not help.

I reported the problem to api-extractor, see https://github.com/microsoft/rushstack/issues/1921

bajtos:
I pushed a new version, can you @raymondfeng PTAL again please?

The impact of the api-extractor bug looks pretty major to me, I think we will have to wait with this pull request until a fix is available. WDYT?

raymondfeng:
> The impact of the api-extractor bug looks pretty major to me, I think we will have to wait with this pull request until a fix is available. WDYT?

I agree. We should wait.

bajtos:
I found a workaround that preserves the current API docs for class members and excludes members that are added via the interface, see 06abcdf. 

@raymondfeng LGTY? Are you good to proceed to landing this PR (after I squash commits)?

Screenshot from the rendered docs:

Screen Shot 2020-06-09 at 16 47 38

raymondfeng:
@bajtos Why don't we add the following the `Context` class itself? The following is working for me.

```ts
export class Context {
  // ...
  on(eventName: 'bind' | 'unbind', listener: ContextEventListener): this;
  on(event: string | symbol, listener: (...args: any[]) => void): this;
  on(event: string | symbol, listener: (...args: any[]) => void): this {
    return super.on(event, listener);
  }
}
```

bajtos:
> Why don't we add the following the Context class itself? The following is working for me.

It didn't work for me, let me try again.

bajtos:
Nice, the proposed solution works :+1:

I find it a bit suboptimal that we have to implement `on`/`once` methods calling `super.on`/`super.once`, but I guess it's the cost we have to accept now.

Here is a screenshot of API docs rendered for Binding class:

Screen Shot 2020-06-11 at 09 52 50

@raymondfeng LGTY now?

PR #5673, opened 2020-06-05
chore: prevent prettier from overwriting line endings on Windows

InvictusMB:
As requested in [comment](https://github.com/strongloop/loopback-next/pull/5616#discussion_r435281611
)
## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


bajtos:
Here is Prettier's documentation for EOL config: https://prettier.io/docs/en/options.html#end-of-line

I think we should carefully consider the implications of this change. Quoting from Prettier docs:
> _First available in v1.15.0, default value changed from auto to lf in v2.0.0_
> When people collaborate on a project from different operating systems, it becomes easy to end up with mixed line endings in a shared git repository. It is also possible for Windows users to accidentally change line endings in a previously committed file from LF to CRLF. Doing so produces a large git diff and thus makes the line-by-line history for a file (git blame) harder to explore.
>
> If you want to make sure that your entire git repository only contains Linux-style line endings in files covered by Prettier:
> (...)

bajtos:
IIUC, the default Prettier configuration ensures that all files are always using Unix line ends, to avoid accidental mixing of different line-end styles.

Instead of enabling the less strict `auto` option, can we perhaps configure `git` to always use Unix line ends, on Windows too? Here is what Prettier recommends:

> Add `* text=auto eol=lf` to the repo's `.gitattributes` file. You may need to ask Windows users to re-clone your repo after this change to ensure git has not converted LF to CRLF on checkout.

@InvictusMB would that work for you?

@dougal83 I believe you are running Windows too, do you have any opinion on what line endings we should use?

dougal83:
@bajtos I'd keep Unix line endings enforced primarily for consistency across platforms.  No one wants a messy git diff... 👍 

Using pure windows I remember configuring git to handle:
https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

Probably best to go for `.gitattributes` route in my experience as mentioned above.

InvictusMB:
`.gitattributes` is probably a better way as it covers more things than `prettier`.

I'll have to test this for some time once it lands on master to see if anything pops up. There is always something emitting OS specific line endings.

dougal83:
> `.gitattributes` is probably a better way as it covers more things than `prettier`.
> 
> I'll have to test this for some time once it lands on master to see if anything pops up. There is always something emitting OS specific line endings.

Remember to wipe/re-clone or you'll be scratching you head like I was for a while!

bajtos:
@InvictusMB @dougal83 are we good to land this pull request?

PR #5674, opened 2020-06-05
feat(repository): CRUD implementation and tests for hasManyThrough Repository

agnes512:
continuation of #4438 and #2359 

Add implementation and tests for CRUD operations. 

Methods `link` and `unlink` need more discussions.

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
+++ b/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
@@ -3,12 +3,13 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {expect} from '@loopback/testlab';
+import {expect, toJSON} from '@loopback/testlab';
 import {
   BelongsToAccessor,
   BelongsToDefinition,
   createBelongsToAccessor,
   createHasManyRepositoryFactory,
+  createHasManyThroughRepositoryFactory,
   DefaultCrudRepository,
   Entity,
   EntityCrudRepository,
@@ -17,6 +18,8 @@ import {
   HasManyDefinition,
   HasManyRepository,
   HasManyRepositoryFactory,
+  HasManyThroughRepository,
+  HasManyThroughRepositoryFactory,
   juggler,
   ModelDefinition,
   RelationType,
@@ -26,6 +29,11 @@ import {
 let db: juggler.DataSource;
 let customerRepo: EntityCrudRepository<Customer, typeof Customer.prototype.id>;
 let orderRepo: EntityCrudRepository<Order, typeof Order.prototype.id>;
+let cartItemRepo: EntityCrudRepository<CartItem, typeof CartItem.prototype.id>;
+let CustomerCartItemLinkRepo: EntityCrudRepository<
+  CustomerCartItemLink,
+  typeof CustomerCartItemLink.prototype.id
+>;
 let reviewRepo: EntityCrudRepository<Review, typeof Review.prototype.id>;
 
 describe('HasMany relation', () => {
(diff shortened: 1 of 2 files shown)


packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -141,36 +144,105 @@ export class DefaultHasManyThroughRepository<
...
+    );
+    const targetConstraint = this.getTargetConstraint(throughInstances);
+    return targetRepository.updateAll(
+      constrainDataObject(dataObject, targetConstraint),
+      constrainWhere(where, targetConstraint as Where<TargetEntity>),
+      options,
+    );
   }
 
   async link(
```
agnes512, line 248: I'd like to bring up discussion here for `link` and `unlink` before making any further changes. 

IMO `link` and `unlink` are for creating/deleting the through instances i.e no target instances should be involved.
If so, the question is how can we make sure the target exists? Or is it users' responsibility?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,247 @@ describe('BelongsTo relation', () => {
...
+
+    expect(orders).have.length(2);
+    expect(orders).to.deepEqual([order1, order2]);
+    const group1 = await hasManyThroughRepo.find({
+      where: {description: 'group 1'},
+    });
+    expect(group1).to.deepEqual([order1]);
+  });
+
+  it('can delete an instance then delet the through model', async () => {
```
jannyHou, line 305:
```suggestion
  it('can delete an instance then delete the through model', async () => {
```

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,247 @@ describe('BelongsTo relation', () => {
...
+    await hasManyThroughRepo.create(
+      {
+        description: 'customer 1',
+      },
+      {
+        throughData: {id: 98},
+      },
+    );
+    const anotherHasManyThroughRepo = hasManyThroughFactory(
```
jannyHou, line 314: hmm, how to make sure the customer with id as `existingCustomerId + 1` exists?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,247 @@ describe('BelongsTo relation', () => {
...
+    );
+    let orders = await orderRepo.find();
+    let links = await customerOrderLinkRepo.find();
+
+    expect(orders).have.length(2);
+    expect(links).have.length(2);
+
+    await hasManyThroughRepo.delete();
+    orders = await orderRepo.find();
+    orders = await orderRepo.find();
```
jannyHou, line 333: a dup one

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,247 @@ describe('BelongsTo relation', () => {
...
+    );
+    const order2 = await anotherHasManyThroughRepo.create(
+      {
+        description: 'customer 2',
+      },
+      {
+        throughData: {id: 99},
+      },
+    );
+    // when order1 gets deleted, this through instance should be deleted too.
```
jannyHou, line 363: Hmm any reason to have a second order here? We can delete `order1`, then check link with `id: 98` is also gone, to verify the through instance is deleted.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -141,36 +145,105 @@ export class DefaultHasManyThroughRepository<
       throughData?: DataObject<ThroughEntity>;
       throughOptions?: Options;
     },
-  ): Promise<void> {
-    throw new Error('Method not implemented.');
+  ): Promise<TargetEntity> {
+    const targetRepository = await this.getTargetRepository(); //{id: inq:[]}
+    const throughRepository = await this.getThroughRepository(); // {category: 1}
+    const targetInstance = await targetRepository.create(
```
jannyHou, line 151: I feel there should be a try catch here. Say if the targetInstance creation fails, e.g. it already exists, then the rest of code should stop and return proper error.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
agnes512: I don't think we check the existence of the source key in any relation repositories, or it would need to access the database again.

packages/repository/src/relations/has-many/has-many-through.repository.ts
agnes512: I think repositories themselves ( through repo and target here) would throw error if anything goes wrong. Do we need extra try catch for them?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
agnes512: You're right, they are not necessary 🙈 I just wanted to make sure that the implementation satisfies [this comment](https://github.com/strongloop/loopback-next/pull/4438/files#r402105078). I can remove it if the test is confusing.

jannyHou: I see it makes sure the constraint only applies to that deleted source model. Then the test case is valid 👍 , my simplified one doesn't check that.

> Here are two scenario to consider & cover by tests, I'll use "Source", "Target" and "Through" as model names instead of Customer/Seller/Order.

1. Given through links: `{sourceId: 1, targetId: 1}` and `{sourceId: 2, targetId: 2}`.
   When I delete all models belonging to source id 1,
   then the target model id 1 and the first link is deleted.

2. Given trough links `{sourceId: 1, targetId: 1}`, `{sourceId: 2, targetId: 1}` and `{sourceId: 2, targetId: 2}`. 
  When I delete all models belonging to source id 1, 
  then the target model id 1 and the first two links are deleted. Only `{sourceId: 2, targetId: 2}` link is preserved.


--- 

I think scenario 2 covers scenario 1, I couldn't think of a reason why 2 pass while 1 fails 🤔 .

agnes512: Haha I think 1 is a general case and 2 is just a more complicated case.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,246 @@ describe('BelongsTo relation', () => {
...
+  before(givenPersistedCustomerInstance);
+  before(givenConstrainedRepositories);
+
+  beforeEach(async function resetDatabase() {
+    await customerRepo.deleteAll();
+    await customerOrderLinkRepo.deleteAll();
+    await orderRepo.deleteAll();
+  });
+
+  it('can create an target instance alone with the corresponding through model', async () => {
```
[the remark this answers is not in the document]
agnes512: Done!

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -141,36 +145,105 @@ export class DefaultHasManyThroughRepository<
...
       throughOptions?: Options;
     },
   ): Promise<Count> {
-    throw new Error('Method not implemented.');
-  }
+    const targetRepository = await this.getTargetRepository();
+    const throughRepository = await this.getThroughRepository();
+    const throughConstraint = this.getThroughConstraint();
+    const throughInstances = await throughRepository.find(
+      constrainFilter(undefined, throughConstraint),
```
jannyHou, line 207: hmm, why the first param is `undefined`?

packages/repository/src/relations/has-many/has-many-through.repository.ts
jannyHou: My take is we should check target instance exists, if not throw error properly.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -141,36 +145,105 @@ export class DefaultHasManyThroughRepository<
...
 
+  // 1. Given through links: {sourceId: 1, targetId: 1} and {sourceId: 2, targetId: 2}.
+  // When delete all models belonging to source id 1,
+  // then the target model id 1 and the first link is deleted.
+
+  // 2. Given trough links {sourceId: 1, targetId: 1}, {sourceId: 2, targetId: 1}
+  // and {sourceId: 2, targetId: 2}.
+  // When delete all models belonging to source id 1,
+  // then the target model id 1 and the first two links are deleted.
+  // Only {sourceId: 2, targetId: 2} link is preserved
```
bajtos, line 196: These comments seem to describe test cases, I feel they don't belong to "production" code.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -13,6 +16,7 @@ import {
   Options,
   Where,
 } from '../..';
+import {constrainWhereOr} from '../../repositories/constraint-utils';
```
bajtos, line 19: Should we import `constrainWhereOr` from `../..` as we already do with other symbols?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -213,7 +460,7 @@ class Order extends Entity {
   static definition = new ModelDefinition('Order')
     .addProperty('id', {type: 'number', id: true})
     .addProperty('description', {type: 'string', required: true})
-    .addProperty('customerId', {type: 'number', required: true})
+    .addProperty('customerId', {type: 'number', required: false})
```
bajtos, line 463: Interesting. How can an order exist when there is no customer placing it? Is it a limitation of the current implementation of has-many-through relation that it's not possible to enforce "required" constraint on foreign keys?

Maybe I am misunderstanding this change. I see that you are adding `CustomerOrderLink` model to establish N-N relation between customers and orders. IMO, `Order` should not have `customerId` property in that scenario.

I think it's not a good idea to reuse customer-order relation for has-many-through and we should introduce new models & relations for that.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +210,246 @@ describe('BelongsTo relation', () => {
...
+      {
+        throughData: {id: 99},
+      },
+    );
+    const order2 = await hasManyThroughRepo.create(
+      {
+        description: 'group 1',
+      },
+      {
+        throughData: {id: 98},
```
bajtos, line 402: I like the fact that it's possible to specify "throughData" when creating new instances of the related model :+1:

packages/repository/src/relations/has-many/has-many-through.repository.ts
bajtos: Can we leave `link` and `unlink` out of scope of this pull request and open a follow-up issue to discuss the desired behavior and implement it?

packages/repository/src/relations/has-many/has-many-through.repository.ts
agnes512: Because there is no original filter to apply to the through repo. Here's the API docs:
```
export function constrainFilter<T extends object>(
  originalFilter: Filter<T> | undefined,
  constraint: AnyObject,
): Filter<T> 
...
```
I think either `{}` or `undefined` works.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
agnes512: 👍 make sense. Added a new model `CartItem`. A `Customer` has many `CartItem`s through `CustomerCartItemLink`.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +211,245 @@ describe('BelongsTo relation', () => {
...
+  before(givenPersistedCustomerInstance);
+  before(givenConstrainedRepositories);
+
+  beforeEach(async function resetDatabase() {
+    await customerRepo.deleteAll();
+    await CustomerCartItemLinkRepo.deleteAll();
+    await cartItemRepo.deleteAll();
+  });
+
+  it('creates an target instance alone with the corresponding through model', async () => {
```
hacksparrow, line 239: `creates an ` -> `creates a `

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +211,245 @@ describe('BelongsTo relation', () => {
...
+
+    expect(items).have.length(2);
+    expect(items).to.deepEqual([item1, item2]);
+    const group1 = await hasManyThroughRepo.find({
+      where: {description: 'group 1'},
+    });
+    expect(group1).to.deepEqual([item1]);
+  });
+
+  it('deletes an instance then delete the through model', async () => {
```
hacksparrow, line 306: `deletes an instance then delete the through model` -> `deletes an instance, then deletes the through model`.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -203,6 +211,245 @@ describe('BelongsTo relation', () => {
...
+    links = await CustomerCartItemLinkRepo.find();
+
+    expect(items).have.length(1);
+    expect(links).have.length(1);
+    expect(items).to.deepEqual([item2]);
+    expect(links[0]).has.property('itemId', item2.id);
+    expect(links[0]).has.property('customerId', existingCustomerId + 1);
+  });
+
+  it('through models get deleted once the related target gets deleted', async () => {
```
hacksparrow, line 343: `through model get deleted once the related target gets deleted` -> `deletes through model when corresponding target gets deleted`

hacksparrow:
Let's use a repo name based on the model instead of `hasManyThroughRepo`, like how we do for the rest of the relations. Simply using `hasManyThroughRepo` is confusing and feels like it is a magical mind-reading repo.

hacksparrow:
LGTM now.

Please open an issue to re-consider the key names `throughData` and `throughOptions`:

```ts
  create(
    targetModelData: DataObject<Target>,
    options?: Options & {
      throughData?: DataObject<Through>;
      throughOptions?: Options;
    },
  ): Promise<Target>;
```

my suggestion is:

```ts
  create(
    targetModelData: DataObject<Target>,
    options?: Options & {
      through?: DataObject<Through>;
      options?: Options;
    },
  ): Promise<Target>;
```

We can do this after everything lands.

PR #5680, opened 2020-06-06
[RFC] refactor: switch to `got` as http client library

raymondfeng:
## Checklist

### DIFF
--- a/benchmark/package-lock.json
+++ b/benchmark/package-lock.json
@@ -4,6 +4,19 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@sindresorhus/is": {
+			"version": "2.1.1",
+			"resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-2.1.1.tgz",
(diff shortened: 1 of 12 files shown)


bajtos:
What is the motivation for this change? Are there any known problems with `axios`? Are there any notable features provided by `got` but not available in `axios`?

My primary requirement for an http client is to preserve stack traces of low-level Node.js errors thrown e.g. when TLS handshake or proxy connection setup fails. Historically, many modern clients were wrapping Node.js errors into custom error instances described in Fetch specification, which made troubleshooting pretty impossible.

`axios` fixed that problem long time ago. I checked `got` sources and it looks like they are most likely not wrapping low-level errors any more, but I am not sure. Can you please double check?

Assuming that `got` is preserving original error stack traces, I am fine start using it, **as long as we use it consistently everywhere**. I really don't want to end up with the situation where some of LoopBack packages are using axios and other are using got.

raymondfeng:
I'm still struggling with `packages/http-caching-proxy` to use `got`, which does not have built-in proxy support.

bajtos:
> I'm still struggling with `packages/http-caching-proxy` to use `got`, which does not have built-in proxy support.

I have encountered similar troubles in the past too. Can you perhaps check how is axios implementing proxy support?

bajtos (approved):
I guess it's better to use `got` in at least few places rather than stick to `request` everywhere.

Can you @raymondfeng please update `got` to the latest version and rebase this pull request on top of the latest `master`, to ensure all is still good?

PR #5681, opened 2020-06-07
feat(pooling): add an extension to provide pooling service

raymondfeng:
https://github.com/strongloop/loopback-next/blob/binding-pool/extensions/pooling/README.md

## Checklist


bajtos:
+1 to implement this as an experimental extension.

I don't have bandwidth to review the proposal. Can you @raymondfeng please pick somebody from the team to become a co-owner of the new package and let that person review the initial version.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+  });
+
+  it('creates a resource pool as singleton', async () => {
+    const poolService = await givenPoolService();
+    const result = await ctx.get(POOLING_SERVICE);
+    expect(result).to.be.exactly(poolService);
+  });
+
+  it('acquires/releases a resource from the pool', async () => {
+    const poolService = await givenPoolService({max: 5});
```
hacksparrow, line 50: Why not leave it out `{max: 5}`? It makes it look like the test has something to do with the parameter, when it is not. Same applies to other locations.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+    poolService.start();
```
hacksparrow, line 51: `poolService.start();` is not called in some places. How does it work? This behavior should be documented.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+  PoolFactory,
+} from '../../pooling';
+
+describe('Resource pool', () => {
+  const POOLING_SERVICE = BindingKey.create<PoolingService<ExpensiveResource>>(
+    'services.pooling',
+  );
+  let ctx: Context;
+
+  beforeEach(givenContext);
```
hacksparrow, line 36: Should we not reset `ExpensiveResource.id` too?

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
hacksparrow: My assumption is that `poolService` will not be usable for anything (eg: `poolService.run`) unless it is `start`ed.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+      },
+      'invalid',
+    );
+    const res = await poolService.acquire();
+    expect(res.status).to.eql('validated');
+    // The first creation fails and generic-pool calls `create` again
+    expect(res.id).to.eql(2);
+  });
+
+  it.skip('validates a resource during release', async () => {
```
hacksparrow, line 154: Why are we skipping this?

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  /**
+   * This simulates a resource that is expensive to create/start. For example,
+   * a datasource has overhead to connect to the database. There will be performance
+   * penalty to use `TRANSIENT` scope and creates a new instance per request.
+   * But it is not feasible to be a singleton for some use cases, for example,
+   * each request may have different security contexts.
+   */
+  class ExpensiveResource implements Poolable {
```
hacksparrow, line 279: We should move this and other artifacts to an artifacts dir.

extensions/pooling/src/pooling.ts
```diff
@@ -0,0 +1,314 @@
...
+      value,
+    );
+    await invokePoolableMethod(value, 'stop');
+  }
+}
+
+function invokePoolableMethod(resource: Poolable, method: keyof Poolable) {
+  if (typeof resource[method] === 'function') {
+    return resource[method]!();
+  }
```
hacksparrow, line 313: `else`: should we throw an error? Failing silently (eg: in case of a typo) doesn't sound good to me.

extensions/pooling/src/pooling.ts
```diff
@@ -0,0 +1,314 @@
...
+    let factory = this.options.factory;
+    if (typeof factory === 'function') {
+      factory = factory(this.context);
+    }
+    this.factory = factory;
+    this.options = {...options, factory};
+    this.pool = createPool(factory, {
+      max: 8, // Default to 8 instances
+      ...this.options.poolOptions,
+      autostart: false,
```
hacksparrow, line 149: If `autostart` is to be configurable, it must be specified before `...this.options.poolOptions`.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
```diff
@@ -0,0 +1,370 @@
...
+    res.status = 'invalid';
+    await poolService.release(res);
+    expect(poolService.pool.isBorrowedResource(res)).to.be.false();
+  });
+
+  it('invokes resource-level acquire/release methods', async () => {
+    const poolService = await givenPoolService(
+      {max: 5},
+      undefined,
+      ExpensiveResourceWithHooks,
```
hacksparrow, line 169: Should the signature of `givenPoolService()` be changed to make this possible:

```
givenPoolService(
      {max: 5},
      ExpensiveResourceWithHooks
);
```

?

extensions/pooling/src/__tests__/unit/README.md
```diff
@@ -0,0 +1 @@
+# Unit tests
```
hacksparrow, line 1: How about some unit tests?

hacksparrow:
Looks mostly good, left some comments.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
raymondfeng: Quoted from [generic-pool](https://github.com/coopernurse/node-pool#createpool):

> autostart: boolean, should the pool start creating resources, initialize the evictor, etc once the constructor is called. If false, the pool can be started by calling pool.start, otherwise the first call to acquire will start the pool. (default true)

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
raymondfeng: Removed.

extensions/pooling/src/__tests__/acceptance/pooling.acceptance.ts
raymondfeng: We're not reusing the class for multiple test files. It's probably better to keep it inside the test file itself.

extensions/pooling/src/pooling.ts
raymondfeng: Not really in this case as the hook method can be optional. I added more `debug` statements.

extensions/pooling/src/pooling.ts
raymondfeng: I intentionally disable `autostart` to fit into our life cycles. Added more comments in the code.

extensions/pooling/src/__tests__/unit/README.md
raymondfeng: The pooling service is a wrapper of `generic-pool`. I believe the extra functions have been covered by the acceptance tests.

raymondfeng:
@hacksparrow Thank you for the comments. PTAL.

PR #5714, opened 2020-06-11
chore(tsdocs): move fixture build out of main build task

bajtos:
Rework tsdocs build scripts as follows:

Introduce a new npm task `build:fixtures` that can be used to rebuild files in `fixtures/monorepo`.

Move this new task out of the main `build` task for the following
reasons:

1. The global `npm run build` command no longer invokes `build` scripts in individual packages, we use Project References to build the entire loopback-next monorepo as a whole. I find it confusing to have different behavior when a package is built from monorepo or locally.

2. It is very rare that we need to (re)build tsdoc fixtures. The files created by the compiler are versioned by `git`, rebuild is needed only to update JavaScript files to match the new `tsc` output.

3. Sometimes, I would see tsdoc fixtures build triggered by unrelated actions, e.g. `npm run tsdocs`, which adds unnecessary friction to our dev process.

## Checklist


packages/tsdocs/package.json
```diff
@@ -13,7 +13,8 @@
     "extract-apidocs": "node bin/extract-apis",
     "document-apidocs": "api-documenter markdown -i ../../docs/apidocs/models -o ../../docs/site/apidocs",
     "update-apidocs": "node bin/update-apidocs",
-    "build": "lb-tsc && npm run build --prefix fixtures/monorepo",
+    "build": "lb-tsc",
+    "build:fixtures": "npm run build --prefix fixtures/monorepo",
```
raymondfeng, line 17: Do we want to run it as part the `npm run build:full` for `loopback-next`? I'm worried that we forget to rebuild the fixtures during TypeScript upgrade.

bajtos: Sure, why not. 

I feel such change is out of scope of this pull request though.

In this PR, I am removing the inconsistency where `npm run build` in monorepo does not build tsdocs fixtures while `npm run build` in `packages/tsdocs` does.

I'd like to land this PR first, and then discuss what process we want to use to ensure fixtures are rebuilt after TypeScript upgrades.

One option that comes to my mind is to re-build the fixtures from `postinstall` hook, which is honored by `lerna bootstrap` and is not executed as often as the build task.

packages/tsdocs/package.json
```diff
@@ -13,7 +13,8 @@
...
     "clean": "lb-clean loopback-tsdocs*.tgz dist *.tsbuildinfo package",
     "pretest": "npm run build",
```
raymondfeng, line 19: Let's add `npm run build:fixtures` to `pretest`.

bajtos: I don't like the idea of rebuilding fixtures before every test run. I want the tests to finish as quickly as possible, adding a build step that's a no-op in most times is not optimal.

bajtos:
Opened a follow-up PR: https://github.com/strongloop/loopback-next/pull/5767

PR #5719, opened 2020-06-11
feat(repository): add link and unlink methods for hasManyThrough repository

agnes512:
Implemented `link` and `unlink` methods.

_(continuation of #4438, #5674 and #2359)_

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
+++ b/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
@@ -30,7 +30,7 @@ let db: juggler.DataSource;
 let customerRepo: EntityCrudRepository<Customer, typeof Customer.prototype.id>;
 let orderRepo: EntityCrudRepository<Order, typeof Order.prototype.id>;
 let cartItemRepo: EntityCrudRepository<CartItem, typeof CartItem.prototype.id>;
-let CustomerCartItemLinkRepo: EntityCrudRepository<
+let customerCartItemLinkRepo: EntityCrudRepository<
   CustomerCartItemLink,
   typeof CustomerCartItemLink.prototype.id
 >;
@@ -232,40 +232,30 @@ describe('HasManyThrough relation', () => {
 
   beforeEach(async function resetDatabase() {
     await customerRepo.deleteAll();
-    await CustomerCartItemLinkRepo.deleteAll();
+    await customerCartItemLinkRepo.deleteAll();
     await cartItemRepo.deleteAll();
   });
 
-  it('creates a target instance alone with the corresponding through model', async () => {
-    const cartItem = await customerCartItemRepo.create(
-      {
-        description: 'an item hasManyThrough',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
+  it('creates a target instance along with the corresponding through model', async () => {
+    const cartItem = await customerCartItemRepo.create({
+      description: 'an item hasManyThrough',
+    });
     const persistedItem = await cartItemRepo.findById(cartItem.id);
-    const persistedLink = await CustomerCartItemLinkRepo.find();
+    const persistedLink = await customerCartItemLinkRepo.find();
+
     expect(cartItem).to.deepEqual(persistedItem);
     expect(persistedLink).have.length(1);
     const expected = {
-      id: 99,
       customerId: existingCustomerId,
       itemId: cartItem.id,
     };
-    expect(toJSON(persistedLink[0])).to.deepEqual(toJSON(expected));
+    expect(toJSON(persistedLink[0])).to.containEql(toJSON(expected));
   });
 
   it('finds an instance via through model', async () => {
-    const item = await customerCartItemRepo.create(
-      {
-        description: 'an item hasManyThrough',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
+    const item = await customerCartItemRepo.create({
+      description: 'an item hasManyThrough',
+    });
     const notMyItem = await cartItemRepo.create({
       description: "someone else's item desc",
     });
@@ -277,22 +267,10 @@ describe('HasManyThrough relation', () => {
   });
 
   it('finds instances via through models', async () => {
-    const item1 = await customerCartItemRepo.create(
-      {
-        description: 'group 1',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
-    const item2 = await customerCartItemRepo.create(
-      {
-        description: 'group 2',
-      },
-      {
-        throughData: {id: 98},
-      },
-    );
+    const item1 = await customerCartItemRepo.create({description: 'group 1'});
+    const item2 = await customerCartItemRepo.create({
+      description: 'group 2',
+    });
     const items = await customerCartItemRepo.find();
 
     expect(items).have.length(2);
@@ -304,34 +282,24 @@ describe('HasManyThrough relation', () => {
   });
 
   it('deletes an instance, then deletes the through model', async () => {
-    await customerCartItemRepo.create(
-      {
-        description: 'customer 1',
-      },
-      {
-        throughData: {id: 98},
-      },
-    );
+    await customerCartItemRepo.create({
+      description: 'customer 1',
+    });
     const anotherHasManyThroughRepo = customerCartItemFactory(
       existingCustomerId + 1,
     );
-    const item2 = await anotherHasManyThroughRepo.create(
-      {
-        description: 'customer 2',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
+    const item2 = await anotherHasManyThroughRepo.create({
+      description: 'customer 2',
+    });
     let items = await cartItemRepo.find();
-    let links = await CustomerCartItemLinkRepo.find();
+    let links = await customerCartItemLinkRepo.find();
 
     expect(items).have.length(2);
     expect(links).have.length(2);
 
     await customerCartItemRepo.delete();
     items = await cartItemRepo.find();
-    links = await CustomerCartItemLinkRepo.find();
+    links = await customerCartItemLinkRepo.find();
 
     expect(items).have.length(1);
     expect(links).have.length(1);
@@ -341,33 +309,23 @@ describe('HasManyThrough relation', () => {
   });
 
   it('deletes through model when corresponding target gets deleted', async () => {
-    const item1 = await customerCartItemRepo.create(
-      {
-        description: 'customer 1',
-      },
-      {
-        throughData: {id: 98},
-      },
-    );
+    const item1 = await customerCartItemRepo.create({
+      description: 'customer 1',
+    });
     const anotherHasManyThroughRepo = customerCartItemFactory(
       existingCustomerId + 1,
     );
-    const item2 = await anotherHasManyThroughRepo.create(
-      {
-        description: 'customer 2',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
+    const item2 = await anotherHasManyThroughRepo.create({
+      description: 'customer 2',
+    });
     // when order1 gets deleted, this through instance should be deleted too.
-    const through = await CustomerCartItemLinkRepo.create({
+    const through = await customerCartItemLinkRepo.create({
       id: 1,
       customerId: existingCustomerId + 1,
       itemId: item1.id,
     });
     let items = await cartItemRepo.find();
-    let links = await CustomerCartItemLinkRepo.find();
+    let links = await customerCartItemLinkRepo.find();
 
     expect(items).have.length(2);
     expect(links).have.length(3);
@@ -375,7 +333,7 @@ describe('HasManyThrough relation', () => {
     await customerCartItemRepo.delete();
 
     items = await cartItemRepo.find();
-    links = await CustomerCartItemLinkRepo.find();
+    links = await customerCartItemLinkRepo.find();
 
     expect(items).have.length(1);
     expect(links).have.length(1);
@@ -386,22 +344,12 @@ describe('HasManyThrough relation', () => {
   });
 
   it('patches instances that belong to the same source model (same source fk)', async () => {
-    const item1 = await customerCartItemRepo.create(
-      {
-        description: 'group 1',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
-    const item2 = await customerCartItemRepo.create(
-      {
-        description: 'group 1',
-      },
-      {
-        throughData: {id: 98},
-      },
-    );
+    const item1 = await customerCartItemRepo.create({
+      description: 'group 1',
+    });
+    const item2 = await customerCartItemRepo.create({
+      description: 'group 1',
+    });
 
     const count = await customerCartItemRepo.patch({description: 'group 2'});
     expect(count).to.match({count: 2});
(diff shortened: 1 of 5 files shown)


agnes512:
> My take is we should check target instance exists, if not throw error properly.
Not necessarily in the first implementation but make sure we create a story for it.

@jannyHou commented in https://github.com/strongloop/loopback-next/pull/5674#discussion_r440455004. The current design is to take a whole instance ( not just the ID) and link it without checking the existence. I don't have strong opinion on either of them. Would like to also hear from @hacksparrow and @bajtos .

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -238,21 +228,38 @@ export class DefaultHasManyThroughRepository<
   }
 
   async link(
-    targetModelId: TargetID,
+    targetInstance: TargetEntity,
```
bajtos, line 231: I find the requirement to provide the entire target entity as suboptimal.

Consider the following REST API call intending to create a link between Category 123 and Product 569.

```
POST /api/categories/123/products/link/569
```

I believe our current design does not require the controller to make any extra database calls before it can create the link. IIRC:

```ts
await this.categoryRepository.products(123).link(569);
```

In your new design, the controller would have to talk to the database first. I find this suboptimal performance-wise.

Can we revert the API change back to link/unlink accepting an id only?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -413,6 +413,32 @@ describe('HasManyThrough relation', () => {
       ]),
     );
   });
+
+  it('links a target instance to the source instance', async () => {
+    const item = await cartItemRepo.create({description: 'an item'});
+    let targets = await customerCartItemRepo.find();
+    expect(targets).to.deepEqual([]);
+
+    await customerCartItemRepo.link(item, {throughData: {id: 98}});
```
bajtos, line 422: I am confused about the particular `throughData` value used here. Is it necessary to set the primary key `id` in order to make `link` work? IMO, `id` should be almost always generated by the database. If your intention is to verify that `throughData` parameter is honored, then please set a different property - one that's neither primary nor foreign key.

agnes512: Oh I thought the id couldn't be auto-generated. Will remove it from all the integration tests.

packages/repository/src/relations/has-many/has-many-through.repository.ts
agnes512: Code changes in: https://github.com/strongloop/loopback-next/pull/5719/commits/c499b5b3eb83f5aa2bccd0093ab6bbf3710c0efc and https://github.com/strongloop/loopback-next/pull/5719/commits/f23f2d6195eb50f9871327bb13d03f55d0ee1365.

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -138,24 +139,33 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
  *    keyTo: 'productId',
  *  },
  * };
+ * createThroughConstraint(resolvedMetadata, {3, name: 'a product'});
```
bajtos, line 142: This isn't a valid JavaScript object initializer, did you perhaps mean the following?

```suggestion
 * createThroughConstraint(resolvedMetadata, {id: 3, name: 'a product'});
```

AFAICT, this example is repeated twice, is that intentional?

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -147,21 +147,11 @@ export class DefaultHasManyThroughRepository<
...
-    const throughConstraint = this.getThroughConstraint();
-    const constraints = {...targetConstraint, ...throughConstraint};
-    await throughRepository.create(
-      constrainDataObject(
-        options?.throughData ?? {},
-        constraints as DataObject<ThroughEntity>,
-      ),
-      options?.throughOptions,
-    );
+    await this.link(targetInstance.getId(), options);
```
bajtos, line 154: Nice 👏

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -138,24 +139,33 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
...
+ * >>> {productId: 3}
  *
  * createThroughConstraint(resolvedMetadata, {id: {inq:[3,4]}});
  *
  * >>> {productId: {inq:[3,4]}}
  */
 export function createThroughFkConstraint<Target, Through extends Entity>(
   relationMeta: HasManyThroughResolvedDefinition,
-  targetInstance: Target,
+  targetInstance: Partial<Target> | undefined,
```
bajtos, line 155: I am concerned that supporting both the target instance and the target primary key is brittle. What if the target key is already an object (e.g. `ObjectID` instance or a composite key consisting of multiple SQL database columns), how are we going to tell whether the value is a key or a full model instance?

Let's rework this function to accept the key only, in the same way as `createThroughConstraint` already does.

```suggestion
  fkValue: ForeignKeyType,
```

packages/repository/src/relations/has-many/has-many-through.helpers.ts
bajtos: Also I think you should use `createThroughFkConstraint` name in these examples, not `createThroughConstraint`.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -65,7 +65,7 @@ export function createHasManyThroughRepositoryFactory<
     }
 
     function getThroughFkConstraint(
-      targetInstance: Target,
+      targetInstance: Partial<Target>,
```
bajtos, line 68: Interesting. If are accepting `Partial<Target>`, then how can we be sure that the partial object contains the properties we need to build the constraint? What if `targetInstance` is just `{title: 'Pencil'}`? (See #5592.)

Should we perhaps change this function to accept the key only?

packages/repository/src/relations/has-many/has-many-through.helpers.ts
agnes512: Ah I meant to show ` * createThroughConstraint(resolvedMetadata, 3)`, the case that it only take the id value. 

I don't know if it's a good design to make it take both JSON object and the target id. But I think it's redundant to have another helper function to handle this case. WDYT?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -564,11 +513,7 @@ class CustomerCartItemLink extends Entity {
   customerId: number;
   itemId: number;
   static definition = new ModelDefinition('CustomerCartItemLink')
-    .addProperty('id', {
-      type: 'number',
-      id: true,
-      required: true,
-    })
```
bajtos, line 545: This is actually an interesting discussion point. Should "through" models have their own primary key (`id`), or should we use a composite primary key composed from source & target keys (e.g. `itemId` + `customerId`)?

@raymondfeng What's your opinion?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -237,35 +237,25 @@ describe('HasManyThrough relation', () => {
...
 
   it('creates a target instance alone with the corresponding through model', async () => {
-    const cartItem = await customerCartItemRepo.create(
-      {
-        description: 'an item hasManyThrough',
-      },
-      {
-        throughData: {id: 99},
-      },
-    );
```
bajtos, line 247: I like how the test suite is much simpler now.

However, do we have any tests to verify `throughData` behavior after this commit removed `throughData` from test cases where it's not explicitly needed/tested?

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
raymondfeng: Imagine we have `Appointment` as the `through` model to connect `Doctor` and `Patient`, a patient may have multiple appointments with the same doctor. So `doctorId + patientId` won't be unique in `Appointment` model, which should have its PK. 

For the relation, we need to use two FKs (doctorId and patientId) instead of PK for `Appointment`.

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
agnes512: Added a test to verify the specified `throughData`.

agnes512:
Thank @bajtos for the detailed review 🙇‍♀️

agnes512:
Thank @bajtos for the detailed review 🙇‍♀️ ( I accidentally closed the issue)

Since we'd like `link` and `unlink` to just take in id values, I modified/added some helpers. I'd like to summarize them there to help review the PR:

To allow these CRUD functions, we need helpers to generate different kind of constraints to meet our requirements. For relationMetadata:
```
  keyFrom: 'id', // Source model
  keyTo: 'id',  // Target model
  through: {
    model: () => Link, // through model
    keyFrom: 'sourceId',
    keyTo: 'targetId',
  }
```
- `createTargetConstraint` creates -> `{id: 1}` for the Target model
- `createFkValues` extracts fks -> `1` or `[1, 2]`
- `createThroughConstraint` creates -> `{sourceId: 5}` for the through model
- `createThroughFkConstraint` creates -> `{targetId: 9}` for the through model based on the passed in **fks**

packages/repository/src/__tests__/unit/repositories/relations-helpers/resolve-has-many-through-metadata.unit.ts
```diff
@@ -28,6 +29,33 @@ describe('HasManyThroughHelpers', () => {
...
   });
+  context('createFkValues', () => {
+    it('can return the target fk value of a given through instance', () => {
+      const resolved = resolvedMetadata as HasManyThroughResolvedDefinition;
+      const through1 = createCategoryProductLink({
+        id: 1,
+        categoryId: 2,
+        productId: 9,
+      });
+      const result = createFkValues(resolved, through1);
```
bajtos, line 40: Nitpick: 

I find the name `resolved` as confusing because it does not say what is being resolved (and what kind of data are we passing in the first argument). Have you considered using `relationMeta` or something similar?

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -138,24 +187,31 @@ export function createThroughConstraint<Through extends Entity, ForeignKeyType>(
...
+  Through extends Entity,
+  ForeignKeyType
+>(
   relationMeta: HasManyThroughResolvedDefinition,
-  targetInstance: Target,
+  fkValue: ForeignKeyType | ForeignKeyType[],
 ): DataObject<Through> {
-  const targetKey = relationMeta.keyTo as StringKeyOf<Target>;
+  if (fkValue === undefined) {
+    throw new Error('"targetInstance" cannot be undefined');
```
bajtos, line 207: Please update the error message to describe the actual error condition.

packages/repository/src/relations/has-many/has-many-through.helpers.ts
```diff
@@ -68,21 +67,70 @@ export function createTargetConstraint<
...
+        id: 1,
+        categoryId: 2,
+        productId: 9,
+      }
+  ]);
+  >>> [8, 9]
+ */
+export function createFkValues<Through extends Entity, TargetID>(
+  relationMeta: HasManyThroughResolvedDefinition,
+  throughInstances: Through | Through[],
```
bajtos, line 119: Is it necessary to support both `Trough` and `Through[]` variants? I noticed this design is affecting multiple helpers. Wouldn't it be simpler to support a single value only (`Through`) and let the caller of this function to deal with the array case by calling `instances.map(i => createFkValues(relationMeta, i))`?

agnes512: I think `Through[]` is still needed cause `createTargetConstraint` calls it as well, and if I remove it, I need to do a lot type conversions to have the complier happy D: but I think the returned type can be simplified to `TargetID`.

agnes512:
Thank @bajtos . It makes more sense with the new names!
I rename them as follows:
| Before        | After (helper)    |After (factory)    |
| ------------- |:-------------:| :-------------:|
| createTargetConstraint            | createTargetConstraintOnThrough | getTargetConstraintOnThrough |
| createFkValues                 | getTargetKeyFromThroughModel | getTargetKey |
| createThroughConstraint.     |  createThroughConstraintOnSource | getThroughConstraintOnSource |
| createThroughFkConstraint   | createThroughConstraintOnTarget | getThroughConstraintOnTarget |

- Improvements of helpers are in https://github.com/strongloop/loopback-next/pull/5719/commits/a6237541c5fc6c95a4d8eeb74c44f8baae79585f
- Improvements of repository/factory are in https://github.com/strongloop/loopback-next/pull/5719/commits/8ad58d43b2e9704b0407843431448ea3402ec0d9

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -413,6 +361,51 @@ describe('HasManyThrough relation', () => {
       ]),
     );
   });
+
+  it('links a target instance to the source instance', async () => {
+    const item = await cartItemRepo.create({description: 'an item'});
+    let targets = await customerCartItemRepo.find();
+    expect(targets).to.deepEqual([]);
```
jannyHou, line 368: I might misunderstand the test case here...My understanding is, link/unlink will add/remove the **through instance**, so I am confused why the targets are empty here?

E.g. There is 
- a doctor with id=1
- a patient with id=9
- **`patient.link(doctor)`** will create an appointment with doctorId=1 & patientId=9

In this case, both target(patient) and source(doctor) exist, they are not empty.

agnes512: Line 366 only creates a "Patient" instance through "patientRepository. It is not linked to any "Doctor" yet. That's why it couldn't find any "patient" through the appointments. Once it is linked, the patient and the appointment should be found ( L 370-375).

jannyHou: Thank you @agnes512 , had a quick chat, and I realized `customerCartItemRepo` is a relation repository instead of model's repository, then all the tests here make perfect sense.

packages/repository/src/relations/has-many/has-many-through-repository.factory.ts
```diff
@@ -51,27 +51,34 @@ export function createHasManyThroughRepositoryFactory<
...
         ForeignKeyType
       >(meta, fkValue);
-      return constriant;
+      return constraint;
     }
 
-    function getThroughFkConstraint(
-      targetInstance: Target,
+    function getThroughConstraintOnTarget(
+      fkValues: TargetID,
```
jannyHou, line 74: Is the `fkValues` an array? if so I think the type should be `TargetID[]` :p

agnes512: I think the `any` type escaped from the type checking. Converted it to `TargetID[]` in https://github.com/strongloop/loopback-next/pull/5719/commits/975f25ec03866f4b65c41d380392214735b7fceb.

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -238,21 +235,38 @@ export class DefaultHasManyThroughRepository<
...
+    targetId: TargetID,
     options?: Options & {
       throughOptions?: Options;
     },
   ): Promise<void> {
-    throw new Error('Method not implemented.');
+    const throughRepository = await this.getThroughRepository();
+    const throughConstraint = this.getThroughConstraintOnSource();
+    const targetConstraint = this.getThroughConstraintOnTarget([targetId]);
+    const constraints = {...targetConstraint, ...throughConstraint};
```
bajtos, line 266: Is it safe to combine constraints this way? Should we use WhereBuilder instead to ensure complex syntax like `or` is handled correctly?

```suggestion
    const constraints = new WhereBuilder(targetConstraint).and(throughConstraint).build();
```

@raymondfeng you may have better insight into what's safe and what's not, could you PTAL?

bajtos:
> - Improvements of helpers are in a623754
> - Improvements of repository/factory are in 8ad58d4

Thank you for the links to individual commits. I started to post review comments there and then realized they don't show as a regular PR review 😱 I re-posted my suggestions via regular review and deleted the comments attached to individual commits. You may get few weird notifications pointing to that deleted content, sorry for that! 🙈

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -232,40 +232,30 @@ describe('HasManyThrough relation', () => {
 
   beforeEach(async function resetDatabase() {
     await customerRepo.deleteAll();
-    await CustomerCartItemLinkRepo.deleteAll();
+    await customerCartItemLinkRepo.deleteAll();
     await cartItemRepo.deleteAll();
   });
 
   it('creates a target instance alone with the corresponding through model', async () => {
```
hacksparrow, line 239: `alone` -> `along`

packages/repository/src/__tests__/unit/repositories/relations-helpers/resolve-has-many-through-metadata.unit.ts
```diff
@@ -13,22 +13,52 @@ import {
...
     it('can create constraint for searching through models', () => {
-      const resolved = resolvedMetadata as HasManyThroughResolvedDefinition;
-      const result = createThroughConstraint(resolved, 1);
+      const result = createThroughConstraintFromSource(relationMetaData, 1);
       expect(result).to.containEql({categoryId: 1});
     });
   });
-  context('createTargetConstraint', () => {
+  context('getTargetKeysFromThroughModels', () => {
+    it('can return the target fk value of a given through instance', () => {
```
hacksparrow, line 32: `can return the target fk value of a given through instance` -> `returns the target fk value of a given through instance`

Similarly, update wherever applicable.

PR #5735, opened 2020-06-14
feat(authentication): add support for multiple strategies on same method

### LINKED ISSUE #5310: Many Auth on same method.
```
@authenticate(['basicAuth', 'localAuth']) // We allow either basic or local auth
```
As mentioned in :@authenticate(['basicAuth', 'localAuth']) // We allow either basic or local auth
by @raymondfeng 
Is this has been implemented. If not can we put in feature-list?

nflaig:
Resolves #5310 

This change allows to specify multiple strategies, e.g. `@authenticate(['basicAuth', 'localAuth'])`

**Edit:** authenticate decorator signature was updated to `@authenticate('basicAuth', 'localAuth')`

The logic on how the strategies are executed is similiar to how [passportjs](http://www.passportjs.org/) does it.

- The authentication strategies will be executed in the provided order.
- If at least one authentication strategy succeeds, the request will be further processed without throwing an error.
- Once a strategy succeeds or redirects, all subsequent strategies will not be evaluated.
- if all strategies fail, a error will be thrown with the error message of the first provided strategy
- it is not feasible to use multiple strategies that redirect (e.g. oauth login redirects) since the first redirect will halt the execution chain

I tried to make this change backwards compatible and therefore I did not rename bindings and it is still possible to set the strategy as a single string.

**Edit:** changes are no longer backwards compatible, see [commit](https://github.com/strongloop/loopback-next/pull/5735/commits/a2aea1145ec0a32a813947126840dce54f8c4434)

The API Documentation in code and the Documentation in [/docs/site](../tree/master/docs/site) still needs to be updated. And probably some examples need to be updated or added.

**Edit:** API documentation in code is updated

## Checklist

### DIFF
--- a/packages/authentication/src/decorators/authenticate.decorator.ts
+++ b/packages/authentication/src/decorators/authenticate.decorator.ts
@@ -29,7 +29,7 @@ class AuthenticateClassDecoratorFactory extends ClassDecoratorFactory<
  * @param options - Additional options to configure the authentication.
  */
 export function authenticate(
-  strategyNameOrMetadata: string | AuthenticationMetadata,
+  strategyNameOrMetadata: string | string[] | AuthenticationMetadata,
   options?: object,
 ) {
   return function authenticateDecoratorForClassOrMethod(
@@ -43,7 +43,10 @@ export function authenticate(
     methodDescriptor?: TypedPropertyDescriptor<any>,
   ) {
     let spec: AuthenticationMetadata;
-    if (typeof strategyNameOrMetadata === 'object') {
+    if (
+      typeof strategyNameOrMetadata === 'object' &&
+      !Array.isArray(strategyNameOrMetadata)
+    ) {
       spec = strategyNameOrMetadata;
     } else {
       spec = {strategy: strategyNameOrMetadata, options: options ?? {}};
--- a/packages/authentication/src/decorators/authenticate.decorator.ts
+++ b/packages/authentication/src/decorators/authenticate.decorator.ts
@@ -18,19 +18,17 @@ import {
 import {AuthenticationMetadata} from '../types';
 
 class AuthenticateClassDecoratorFactory extends ClassDecoratorFactory<
-  AuthenticationMetadata
+  AuthenticationMetadata[]
 > {}
 
 /**
  * Mark a controller method as requiring authenticated user.
  *
- * @param strategyNameOrMetadata - The name of the authentication strategy to use
- * or the authentication metadata object.
- * @param options - Additional options to configure the authentication.
+ * @param strategies - The names of the authentication strategies to use
+ * or authentication metadata objects.
  */
 export function authenticate(
-  strategyNameOrMetadata: string | string[] | AuthenticationMetadata,
-  options?: object,
+  ...strategies: (string | AuthenticationMetadata)[]
 ) {
   return function authenticateDecoratorForClassOrMethod(
     // Class or a prototype
@@ -42,34 +40,35 @@ export function authenticate(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     methodDescriptor?: TypedPropertyDescriptor<any>,
   ) {
-    let spec: AuthenticationMetadata;
-    if (
-      typeof strategyNameOrMetadata === 'object' &&
-      !Array.isArray(strategyNameOrMetadata)
-    ) {
-      spec = strategyNameOrMetadata;
-    } else {
-      spec = {strategy: strategyNameOrMetadata, options: options ?? {}};
+    const specs: AuthenticationMetadata[] = [];
+
+    for (const strategy of strategies) {
+      if (typeof strategy === 'object') {
+        specs.push(strategy);
+      } else {
+        specs.push({strategy: strategy});
+      }
     }
+
     if (method && methodDescriptor) {
       // Method
-      return MethodDecoratorFactory.createDecorator<AuthenticationMetadata>(
+      return MethodDecoratorFactory.createDecorator<AuthenticationMetadata[]>(
         AUTHENTICATION_METADATA_KEY,
-        spec,
+        specs,
         {decoratorName: '@authenticate'},
       )(target, method, methodDescriptor);
     }
     if (typeof target === 'function' && !method && !methodDescriptor) {
       // Class
-      return AuthenticateClassDecoratorFactory.createDecorator(
-        AUTHENTICATION_METADATA_CLASS_KEY,
-        spec,
-        {decoratorName: '@authenticate'},
-      )(target);
+      return AuthenticateClassDecoratorFactory.createDecorator<
+        AuthenticationMetadata[]
+      >(AUTHENTICATION_METADATA_CLASS_KEY, specs, {
+        decoratorName: '@authenticate',
+      })(target);
     }
     // Not on a class or method
     throw new Error(
-      '@intercept cannot be used on a property: ' +
+      '@authenticate cannot be used on a property: ' +
         DecoratorFactory.getTargetName(target, method, methodDescriptor),
     );
   };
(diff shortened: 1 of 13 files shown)


raymondfeng:
@nflaig Thank you for the contribution.

packages/authentication/src/decorators/authenticate.decorator.ts
```diff
@@ -29,7 +29,7 @@ class AuthenticateClassDecoratorFactory extends ClassDecoratorFactory<
  * @param options - Additional options to configure the authentication.
  */
 export function authenticate(
-  strategyNameOrMetadata: string | AuthenticationMetadata,
+  strategyNameOrMetadata: string | string[] | AuthenticationMetadata,
```
raymondfeng, line 32: Let's improve the user experience here. I would like to see the following styles:

1. Allow one or more strategy names
```ts
@authenticate('my-strategy-1', 'my-strategy-2')
```
2. Allow one or more meta data objects
```
@authenticate(metaForStrategy1, metaForStrategy2)
```
3. Allow multiple applications of `@authenticate` on the same method: (nice-to-have)
```ts
@authenticate('my-strategy-1');
@authenticate(metaForMyStrategy1);
```

The signature can be:
```ts
export function authenticate(...strategies: (string | AuthenticationMetadata) []) { ... }
```
It allows one or more `string` or `AuthenticationMetadata`.

Please note that different strategy may have different metadata.

packages/authentication/src/types.ts
```diff
@@ -22,7 +22,7 @@ export interface AuthenticationMetadata {
   /**
    * Name of the authentication strategy
    */
-  strategy: string;
+  strategy: string | string[];
```
jannyHou, line 25: according to https://github.com/strongloop/loopback-next/pull/5735/files#r439852618, let's still keep the name as type `string`. One metadata object configures one strategy.

nflaig:
@jannyHou I agree, the signature indeed needs to be changed. I will take a look at it once I find the time or maybe someone else wants to look into it. 

>  A failed oauth authentication should not execute the redirect, but let other strategies try instead.

This is unlikely to happen unless there is a configuration or programming error. Usually, strategies such as saml or oauth (oidc) will immediately redirect the user to the identity provider to authenticate the user and once this redirect happens you can not execute further strategies.

One more thing I was wondering about that is unrelated to this PR is that a strategy etiher needs to redirect or return a UserProfile else a [error will be thrown](https://github.com/strongloop/loopback-next/blob/master/packages/authentication/src/providers/auth-action.provider.ts#L73). I feel like this limits the amount of strategies you can use since some just don't return anything.

raymondfeng:
> One more thing I was wondering about that is unrelated to this PR is that a strategy etiher needs to redirect or return a UserProfile else a error will be thrown. I feel like this limits the amount of strategies you can use since some just don't return anything.

In case that a strategy does not return a user profile, do you expect it to pass control to other strategies or simply return?

nflaig:
I think it should simply return and only in case the strategy throws an error it should execute the next strategy in the chain. 

An example could be a strategy using api keys which might only check if the key is valid and the key itself might not correlate to a user.

packages/authentication/src/decorators/authenticate.decorator.ts
nflaig: I did look into that but this means that the changes will be breaking for sure since there are no more options in the decorator, e.g. this will fail now
```ts
it('can add authenticate metadata to target method with options', () => {
      class TestClass {
        @authenticate('my-strategy', {option1: 'value1', option2: 'value2'})
        whoAmI() {}
      }

      const metaData = getAuthenticateMetadata(TestClass, 'whoAmI');
      expect(metaData).to.eql({
        strategy: 'my-strategy',
        options: {option1: 'value1', option2: 'value2'},
      });
    });
```

One more thing I am not sure about is how the metadata for each strategy should be injected into a strategy since it is now an array of metadata object.

```ts
 private metadata: AuthenticationMetadata[], // this is now an array
```

The solution could be to iterate through the metadata array and find the according metadata based on the strategy name, maybe a helper function could be added or can this be done in the inject somehow?

raymondfeng:
> I think it should simply return and only in case the strategy throws an error it should execute the next strategy in the chain.
> An example could be a strategy using api keys which might only check if the key is valid and the key itself might not correlate to a user.

So the return value should be `UserProfile | Redirect | undefined`. That's fine with you as long as we make it clear in documentation.

raymondfeng:
@nflaig Ping. Do you need further input to move forward?

nflaig:
@raymondfeng I have few local changes that I did not push yet becasue of the problem mentioned [above](https://github.com/strongloop/loopback-next/pull/5735#discussion_r443993216).

packages/authentication/src/decorators/authenticate.decorator.ts
raymondfeng: Great question.

For `@inject`, I don't think we can resolve it to the metadata for a given strategy name, which is unknown.

Maybe we have to use `AuthenticationMetadata[]` as the type and introduce a helper function to get `AuthenticationMetadata` by strategy name.

raymondfeng:
@nflaig Great progress.

For the commit message:
> BREAKING CHANGE: signature of authenticate decorator changed, options are no longer a separate input
parameter but instead have to be provided in metadata object if required

We probably need to make it clear there are a few breaking changes:

- The `@authenticate` signature
- The metadata value is now `AuthenticationMetadata[]`

nflaig:
@raymondfeng where exactly should I document the breaking changes? Should I amend the commit message?

raymondfeng:
> where exactly should I document the breaking changes? Should I amend the commit message?

Yes, amending the commit message. Please also make sure related `.md` files are updated in `/docs/site` .

raymondfeng:
@nflaig I have rebased your PR against latest master and pushed it back to your branch.

packages/authentication/src/types.ts
```diff
@@ -133,3 +133,16 @@ export const asAuthStrategy: BindingTemplate = binding => {
...
   });
 };
+
+/**
+ * Get the authentication metadata object for the specified strategy.
+ *
+ * @param metadata - Array of authentication metadata objects
+ * @param strategyName - Name of the authentication strategy
+ */
+export function getAuthenticationMetadataForStrategy(
```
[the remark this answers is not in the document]
nflaig: @jannyHou Its just a helper function to be used by the consumer of the library. One drawback now is that the injected metadata will be an array which means it is required to get the specific metadata for the strategy and this function can be used to do this. I just put it into `types.ts` because there were most other functions as well but its probably a good idea to create a file called `utils.ts` or `helpers.ts` were those functions should be placed. Same goes for the `registerAuthenticationStrategy` function.

nflaig:
Im might find time this weekend to finish the PR the todos I noted for myself are:
- update docs (required)
- add acceptance tests (required?)
- move function from `types.ts` to `helpers.ts`

Let me know if I missed something

packages/authentication/src/types.ts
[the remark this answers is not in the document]
jannyHou: > since there are more functions that should be moved to different files it might make sense to do it in another refactoring PR

@nflaig Thank you for the explanation, the plan sounds good to me.

dhmlau:
_Landing this PR as there are 2 approvals already and I'll be switching to DCO soon._

@nflaig, thanks for your contribution. Your PR has landed. 
Please open a new PR for any follow up work. Thanks!

jannyHou:
Thank you @nflaig again for the contribution!

> I looked through the documentation and not sure how to go about updating it. There are a lot of code samples which are outdated now and even images that include code.
Probably a new section needs to be added which describes how to use multiple strategies and what kind of implications and limitations exist.

I am submitting a PR to update the auth action doc, and I can add new behaviour for this PR too.

PR #5742, opened 2020-06-15
feat(repository): add helpers to create relations

### LINKED ISSUE #5384: ModelDefinition helpers for defining relations
_Moved from [my comment](https://github.com/strongloop/loopback-next/pull/5358/files/da9a530d408a4199fb9c51895df35397530ab846#r421315010) in #5358_

Building relations using `ModelDefinition` API is too involved at the moment, the user has to specify metadata that can be automatically inferred from the relation type.

For example:

```ts
  .addRelation(<BelongsToDefinition>{
    name: 'customer',
    type: RelationType.belongsTo,
    targetsMany: false,
    source: Order,
    target: () => User,
    keyFrom: 'customerId',
  });
```

## Suggestion

Provide higher-level builders for each individual relation type.

For example:

```ts
new ModelDefinition('Order')
  .belongsTo('customer', {
    source: Order,
    target: () => User,
    keyFrom: 'customerId', // can we infer this value?
  });
```

## Acceptance criteria

Implementation, tests and TSDOC API docs for all built-in relation types.

 - [ ] belongsTo 
 - [ ] hasOne 
 - [ ] hasMany 
 - [ ] embedsOne 
 - [ ] embedsMany 
 - [ ] referencesOne 
 - [ ] referencesMany 

Based on the list of relation types as of 2020-05-11:

https://github.com/strongloop/loopback-next/blob/4cc8eba853232213f9f82408568587d81103142e/packages/repository/src/relations/relation.types.ts#L11-L19

jimmy1wu:
* create helpers for belongsTo, hasOne, and hasMany relations
* infer the keyFrom value for belongsTo relation

Fixes https://github.com/strongloop/loopback-next/issues/5384

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/model/model.unit.ts
+++ b/packages/repository/src/__tests__/unit/model/model.unit.ts
@@ -508,4 +508,66 @@ describe('model', () => {
       ).to.throw(/Navigational properties are not allowed in model data/);
     });
   });
+
+  describe('relation helpers', () => {
+    it('adds belongsTo relation', () => {
+      const definition = new ModelDefinition('Phone').belongsTo('customer', {
+        source: Phone,
+        keyFrom: 'customerId',
+
+        target: () => Customer,
+        keyTo: 'id',
+      });
+
+      expect(definition.relations.customer).to.eql({
+        name: 'customer',
+        source: Phone,
+        keyFrom: 'customerId',
+        target: () => Customer,
+        keyTo: 'id',
+        type: RelationType.belongsTo,
+        targetsMany: false,
+      });
+    });
+
+    it('adds hasOne relation', () => {
+      const definition = new ModelDefinition('Customer').hasOne('address', {
+        source: Customer,
+        keyFrom: 'id',
+
+        target: () => Address,
+        keyTo: 'customerId',
+      });
+
+      expect(definition.relations.address).to.eql({
+        name: 'address',
+        source: Customer,
+        keyFrom: 'id',
+        target: () => Address,
+        keyTo: 'customerId',
+        type: RelationType.hasOne,
+        targetsMany: false,
+      });
+    });
+
+    it('adds hasMany relation', () => {
+      const definition = new ModelDefinition('Customer').hasMany('phones', {
+        source: Customer,
+        keyFrom: 'id',
+
+        target: () => Phone,
+        keyTo: 'customerId',
+      });
+
+      expect(definition.relations.phones).to.eql({
+        name: 'phones',
+        source: Customer,
+        keyFrom: 'id',
+        target: () => Phone,
+        keyTo: 'customerId',
+        type: RelationType.hasMany,
+        targetsMany: true,
+      });
+    });
+  });
 });
--- a/packages/repository/src/model.ts
+++ b/packages/repository/src/model.ts
@@ -4,8 +4,14 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {AnyObject, DataObject, Options, PrototypeOf} from './common-types';
-import {JsonSchema} from './index';
-import {RelationMetadata} from './relations';
+import {
+  BelongsToDefinition,
+  HasManyDefinition,
+  HasOneDefinition,
+  JsonSchema,
+  RelationMetadata,
+  RelationType,
+} from './index';
 import {TypeResolver} from './type-resolver';
 import {Type} from './types';
 
@@ -143,6 +149,60 @@ export class ModelDefinition {
   addRelation(definition: RelationMetadata): this {
     this.relations[definition.name] = definition;
     return this;
+  }
+
+  /**
+   * Define a new belongsTo relation.
+   * @param name - The name of the belongsTo relation.
+   * @param definition - The definition of the belongsTo relation.
+   */
+  belongsTo(
+    name: string,
+    definition: Omit<BelongsToDefinition, 'name' | 'type' | 'targetsMany'>,
+  ): this {
+    const meta: BelongsToDefinition = {
+      ...definition,
+      name,
+      type: RelationType.belongsTo,
+      targetsMany: false,
+    };
+    return this.addRelation(meta);
+  }
+
+  /**
+   * Define a new hasOne relation.
+   * @param name - The name of the hasOne relation.
+   * @param definition - The definition of the hasOne relation.
+   */
+  hasOne(
+    name: string,
+    definition: Omit<HasOneDefinition, 'name' | 'type' | 'targetsMany'>,
+  ): this {
+    const meta: HasOneDefinition = {
+      ...definition,
+      name,
+      type: RelationType.hasOne,
+      targetsMany: false,
+    };
+    return this.addRelation(meta);
+  }
+
+  /**
+   * Define a new hasMany relation.
+   * @param name - The name of the hasMany relation.
+   * @param definition - The definition of the hasMany relation.
+   */
+  hasMany(
+    name: string,
+    definition: Omit<HasManyDefinition, 'name' | 'type' | 'targetsMany'>,
+  ): this {
+    const meta: HasManyDefinition = {
+      ...definition,
+      name,
+      type: RelationType.hasMany,
+      targetsMany: true,
+    };
+    return this.addRelation(meta);
   }
 
   /**
--- a/packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
+++ b/packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
@@ -4,6 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import debugFactory from 'debug';
+import {camelCase} from 'lodash';
 import {InvalidRelationError} from '../../errors';
 import {isTypeResolver} from '../../type-resolver';
 import {BelongsToDefinition, RelationType} from '../relation.types';
@@ -14,7 +15,10 @@ const debug = debugFactory('loopback:repository:belongs-to-helpers');
  * Relation definition with optional metadata (e.g. `keyTo`) filled in.
  * @internal
  */
-export type BelongsToResolvedDefinition = BelongsToDefinition & {keyTo: string};
+export type BelongsToResolvedDefinition = BelongsToDefinition & {
+  keyFrom: string;
+  keyTo: string;
+};
 
 /**
  * Resolves given belongsTo metadata if target is specified to be a resolver.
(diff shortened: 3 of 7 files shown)


packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
```diff
@@ -49,12 +48,24 @@ export function resolveBelongsToMetadata(relationMeta: BelongsToDefinition) {
...
+  let keyFrom;
+  if (
+    relationMeta.keyFrom &&
+    relationMeta.source.definition.properties[relationMeta.keyFrom]
+  ) {
+    keyFrom = relationMeta.keyFrom;
+  } else {
+    keyFrom = camelCase(targetName + '_id');
+  }
+
```
agnes512, line 60: I think the current design allows users to customize the `keyFrom` and `keyTo` value. And `keyFrom` is default to the decorated key (e.g `customerId`), see [code](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/relations/belongs-to/belongs-to.decorator.ts#L50). Even it throws when `keyFrom` is undefined, I don't think this situation would happen as it would default the decorated key anyway. So I don't think the change is necessary.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,63 @@ export class ModelDefinition {
     return this;
   }
 
+  /**
```
agnes512, line 152: What't the intention of adding these helpers? And I think relations should not be the part of the model definition.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,63 @@ export class ModelDefinition {
...
+   * Define a new "belongs to" relation.
+   * @param name - The name of the "belongs to" relation.
+   * @param definition - The definition for the "belongs to" relation.
+   */
+  belongsTo(
+    name: string,
+    definition: Partial<BelongsToDefinition> &
+      Pick<BelongsToDefinition, 'source' | 'target'>,
```
bajtos, line 160: I find it a bit difficult to understand which properties are required and which are optional. Relation definitions have most properties optional, do we need to apply `Partial` here?

It makes me wonder if something like the following would work too:

```ts
definition: Omit<BelongsToDefinition, 'name' | 'type' | 'targetsMany'>,
```

Feel free to ignore this comment if my suggestion requires too many changes to make it work.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,63 @@ export class ModelDefinition {
...
+  ): this {
+    const belongsToDefinition: BelongsToDefinition = {
+      ...definition,
+      name,
+      type: RelationType.belongsTo,
+      targetsMany: false,
```
bajtos, line 166: Is it possible to infer `keyFrom` value from the target model name? Let's say `target` is `() => Customer` and we are defining the relation "order belongs to a customer", would something like this work?

```ts
if (! belongsToDefinition.keyFrom) {
  const targetName = definition.target().modelName;
  // ^^ should be "Customer"
  belongsToDefinition.keyFrom = camelCase(`${targetName}_Id`);
  // ^^ should produce "customerId"
}
```

A similar comment applies to other relation types too - `hasOne` and `hasMany` can use `definition.source.modelName` to build the foreign key name.

Since this is an incremental addition, I am ok to leave it out of this initial version if you prefer.

packages/repository/src/relations/relation.types.ts
```diff
@@ -117,7 +117,7 @@ export interface BelongsToDefinition extends RelationDefinitionBase {
   /*
    * The foreign key in the source model, e.g. Order#customerId.
    */
-  keyFrom: string;
+  keyFrom?: string;
```
bajtos, line 120: I have mixed feelings about this change. On one hand, I think it allows us to simplify `belongsTo` method to accept `definition: Omit<BelongsToDefinition, 'name' | 'type' | 'targetsMany'>` and is kind of consistent with other relation types that have `keyFrom` optional too.

On the other hand, I am not able to fully asses ramifications of this change, I have no idea what it could break.

On the third hand :) Since this is backwards compatible and you have code to handle the case where `keyFrom` is not provided, I think it should be fine.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,63 @@ export class ModelDefinition {
...
+
+  /**
+   * Define a new "has one" relation.
+   * @param name - The name of the "has one" relation.
+   * @param definition - The definition of the "has one" relation.
+   */
+  hasOne(
+    name: string,
+    definition: Partial<HasOneDefinition> &
+      Pick<HasOneDefinition, 'source' | 'target'>,
```
bajtos, line 179: Similarly here, would the following work?

```suggestion
    definition: Omit<HasOneDefinition, 'name' | 'type' | 'targetsMany'>,
```

packages/repository/src/model.ts
```diff
@@ -145,6 +149,63 @@ export class ModelDefinition {
...
+
+  /**
+   * Define a new "has many" relation.
+   * @param name - The name of the "has many" relation.
+   * @param definition - The definition of the "has many" relation.
+   */
+  hasMany(
+    name: string,
+    definition: Partial<HasManyDefinition> &
+      Pick<HasManyDefinition, 'source' | 'target'>,
```
bajtos, line 198: ditto

```suggestion
    definition: Omit< HasManyDefinition, 'name' | 'type' | 'targetsMany'>,
```

packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
bajtos: I think the intention here is to allow users to omit `keyFrom` when calling `modelDefinition.belongsTo()` - that's a different code path from the one used by `@belongsTo` decorator. When calling `belongsTo`, I think it's useful to be able to leave out `keyFrom` and let the framework to compute it.

packages/repository/src/model.ts
bajtos: Oh, I see that you are computing `keyFrom` in `resolveBelongsToMetadata`, that's a good solution too 👍

packages/repository/src/model.ts
jimmy1wu: I agree this is a lot more clear. I will update the PR soon.

packages/repository/src/model.ts
jimmy1wu: I tried your suggestion and adding the logic inside the belongsTo function may lead to reference errors if they define the target class after the class that is using the belongsTo method, so adding the logic in `resolveBelongsToMetadata` is more flexible.

e.g.,
https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts#L464

The `Customer` class variable is available in the scope, but has not been initialized yet so it will throw a reference error if we try to access it.

jimmy1wu:
@agnes512 @bajtos 
Thank you for the early review and suggestions :)
I will update the PR shortly.

Looking into tests today / this weekend

packages/repository/src/relations/belongs-to/belongs-to.helpers.ts
```diff
@@ -49,12 +48,24 @@ export function resolveBelongsToMetadata(relationMeta: BelongsToDefinition) {
...
   debug('Resolved model %s from given metadata: %o', targetName, targetModel);
 
+  let keyFrom;
+  if (
+    relationMeta.keyFrom &&
+    relationMeta.source.definition.properties[relationMeta.keyFrom]
+  ) {
+    keyFrom = relationMeta.keyFrom;
+  } else {
+    keyFrom = camelCase(targetName + '_id');
```
bajtos, line 58: Please add a test case to execute this line.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,60 @@ export class ModelDefinition {
     return this;
   }
 
+  /**
+   * Define a new "belongs to" relation.
+   * @param name - The name of the "belongs to" relation.
+   * @param definition - The definition for the "belongs to" relation.
+   */
+  belongsTo(
```
bajtos, line 157: Please add a test to verify this new API.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,60 @@ export class ModelDefinition {
...
+    };
+    return this.addRelation(belongsToDefinition);
+  }
+
+  /**
+   * Define a new "has one" relation.
+   * @param name - The name of the "has one" relation.
+   * @param definition - The definition of the "has one" relation.
+   */
+  hasOne(
```
bajtos, line 175: Please add a test to verify this new API.

packages/repository/src/model.ts
```diff
@@ -145,6 +149,60 @@ export class ModelDefinition {
...
+    };
+    return this.addRelation(hasOneDefinition);
+  }
+
+  /**
+   * Define a new "has many" relation.
+   * @param name - The name of the "has many" relation.
+   * @param definition - The definition of the "has many" relation.
+   */
+  hasMany(
```
bajtos, line 193: Please add a test to verify this new API.

jimmy1wu:
@agnes512 ready for review 😁

packages/repository/src/__tests__/unit/model/model.unit.ts
```diff
@@ -313,11 +307,7 @@ describe('model', () => {
     Category.definition = new ModelDefinition('Category')
       .addSetting('strict', false)
       .addProperty('id', {type: 'number', id: true, required: true})
-      .addRelation(<HasManyDefinition>{
```
agnes512, line 316: Let's revert this part and the test at line 479, or there is no tests cover `addRelation`.

packages/repository/src/__tests__/unit/model/model.unit.ts
```diff
@@ -508,4 +495,66 @@ describe('model', () => {
       ).to.throw(/Navigational properties are not allowed in model data/);
     });
   });
+
+  describe('test adding relations through helpers', () => {
```
agnes512, line 499: Nitpick: `describe('relation helpers', ..`

jimmy1wu:
@agnes512 
while addressing your feedback, i noticed that i wasn't always importing from the common index file and added a commit to fix it. sorry about that; trusted my IDE too much! 

the rest of your comments are addressed.

packages/repository/src/__tests__/unit/model/model.unit.ts
```diff
@@ -263,11 +263,7 @@ describe('model', () => {
     Category.definition = new ModelDefinition('Category')
       .addSetting('strict', true)
       .addProperty('id', {type: 'number', id: true, required: true})
-      .addRelation(<HasManyDefinition>{
```
agnes512, line 266: Please revert this part as well.

PR #5744, opened 2020-06-15
fix: clean up sandbox directories after tests

raymondfeng:
This is rework of https://github.com/strongloop/loopback-next/pull/5720.

There are 3 commits:

~1. Use unique sandbox subdirs for parallel testing (must-have)~ Extracted to https://github.com/strongloop/loopback-next/pull/5747

2. Add `KEEP_TEST_SANDBOX` env var to keep files for sandbox.delete (nice-to-have)

3. Add `after` hook to delete test sandbox directories (should-have)

## Checklist

### DIFF
--- a/packages/testlab/src/test-sandbox.ts
+++ b/packages/testlab/src/test-sandbox.ts
@@ -107,10 +107,13 @@ export class TestSandbox {
   }
 
   /**
-   * Deletes the TestSandbox.
+   * Deletes the TestSandbox. If `KEEP_TEST_SANDBOX` env variable is set, we
+   * leave the sandbox directory as-is on the file system.
    */
   async delete(): Promise<void> {
-    await remove(this.path);
+    if (!process.env.KEEP_TEST_SANDBOX) {
+      await remove(this.path);
+    }
     delete this._path;
   }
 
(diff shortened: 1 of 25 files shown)


packages/testlab/src/test-sandbox.ts
```diff
@@ -106,10 +106,13 @@ export class TestSandbox {
...
   /**
-   * Deletes the TestSandbox.
+   * Deletes the TestSandbox. If `KEEP_TEST_SANDBOX` env variable is set, we
+   * leave the sandbox directory as-is on the file system.
    */
   async delete(): Promise<void> {
-    await remove(this.path);
+    if (!process.env.KEEP_TEST_SANDBOX) {
+      await remove(this.path);
+    }
```
bajtos, line 115: I am little bit reluctant to implement env-based behavior inside delete, I think it could be surprising. Can we implement a new method please and find a descriptive name that will make it more clear what is happening or where it should be used?

Since our motivation is to run this code in after/afterEach hooks, maybe `cleanupAfterTests` could be a better name?

```ts
after('delete sandbox', () => sandbox.cleanupAfterTests());
```

> When I was trying to troubleshoot test failures, I found there were many left-over subdirectories under `.sandbox`. This prevented me from knowing which directory I should look into.

I would like to investigate a way how to let the sandbox cleanup routine detect when a test failed and print a helpful message (something along the lines "sandbox for test XYZ is in directory ABC"). This is probably out of scope of this PR, but if we introduce a new method like `cleanupAfterTests`, then it will be easier to enhance the behavior in the future.

> Do we want to honor this env var in `@loopback/testlab` for both `reset` and `delete`?

I think we should apply ENV-specific behavior in `delete` only. `reset` is meant to be called before the tests, I think the test would fail when the sanbox was not cleared because of ENV config.

PR #5721, opened 2020-06-11
refactor: move benchmark and cli acceptance tests into separate travis tasks

raymondfeng:
1. The benchmark test can be time-consuming and it doesn't involve functional testing. 

2. The CLI acceptance test for `lb4 app` is very slow too as it involves `npm install`.

This PR moves them out of the top-level `npm test` and creates separate Travis tasks to run them as part of CI.

See https://github.com/strongloop/loopback-next/issues/5670

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [x] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [x] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5745, opened 2020-06-15
test(cli): only run `lb4 app` acceptance test with CI

raymondfeng:
This is a follow-up to https://github.com/strongloop/loopback-next/pull/5721#issuecomment-643345496.

## Checklist


bajtos:
Thank you for opening this follow-up pull request!

> > I feel it's important to keep running CLI acceptance tests on all platforms we support (Windows, MacOS, Linux) as we were doing before this pull request. @raymondfeng can you please fix that?
> 
> Good point. Do we want to make the test conditional on `process.env.CI` flag?

IMO, the developers should have an easy way how to run the test locally. For example, when the test fails on CI, we don't want to debug the failure by repeatedly triggering a new CI run, that's impractical and very time consuming. 

I have mixed feelings about using `process.env.CI` flag to decide which tests to run or skip (unless there are technical reasons, e.g. Travis not supporting IPv6). On one hand, it's an easy solution to implement. On the other hand, it can be confusing for developers when `npm test` passes on their machine and then the CI build fails on a test that's skipped outside of CI.

Let's agree on the intended dev experience first and then discuss implementation details. I am thinking about the following user stories:

1. As a developer working on loopback-next, I want an easy way how to run "fast" tests, so that I can frequently re-run the test suite during development (think of red-green-refactor cycle).
2. As a developer working on loopback-next, I want an easy way how to run all tests that are part of the CI test suite, including the slower end-to-end tests. First, I want to ensure that all is good before pushing to github. Secondly, I want an easy solution for debugging a failing (end-to-end) test.
3. As framework maintainers, we want our CI builds to run both "fast" and "end-to-end" tests on all relevant platforms. Some tests are ok to run on a single platform only (e.g. repository acceptance tests), others must be executed on all platforms (CLI tests).
4. Everybody wants fast feedback from CI builds. The build pipeline/workflow should be optimized to avoid repeated work but also run jobs in parallel when possible.

.travis.yml
```diff
@@ -129,12 +129,3 @@ jobs:
...
         - cd benchmark && npm test
-
-    - node_js: "12"
-      os: linux
-      env:
-        - TASK=test-cli-lb4-app
-      script:
-        - lerna bootstrap
-        - npm run build
-        - cd packages/cli && npm run ci-test
```
jannyHou, line 140: do we still need `ci-test` script in https://github.com/strongloop/loopback-next/blob/master/packages/cli/package.json#L94 ?

raymondfeng: Good catch.

raymondfeng:
> As a developer working on loopback-next, I want an easy way how to run "fast" tests, so that I can frequently re-run the test suite during development (think of red-green-refactor cycle).

+1. In this case, we skip the heavy `lb4 app` test.

> As a developer working on loopback-next, I want an easy way how to run all tests that are part of the CI test suite, including the slower end-to-end tests. First, I want to ensure that all is good before pushing to github. Secondly, I want an easy solution for debugging a failing (end-to-end) test.

+1. Use `CI=true npm t` in this case.

> As framework maintainers, we want our CI builds to run both "fast" and "end-to-end" tests on all relevant platforms. Some tests are ok to run on a single platform only (e.g. repository acceptance tests), others must be executed on all platforms (CLI tests).

+1. The `lb4 app` test will be run on all platforms.

> Everybody wants fast feedback from CI builds. The build pipeline/workflow should be optimized to avoid repeated work but also run jobs in parallel when possible.

I was thinking about a separate job for all platforms in this case, but it seems to be too complex for one test.

raymondfeng:
@bajtos PTAL

bajtos:
I have mixed feelings about using `CI` env var to enable tests, but I guess it's the best option we have found so far.

Please update our developer/contrib documentation to explain users they need to run `CI=1 npm t` before submitting a pull request, and provide instructions for Windows users too (`CI=1 npm t` does not work on Windows). Few places where to start the search:

- https://github.com/strongloop/loopback-next/blob/862072b82ad88ec4b248e0b243dda58b53640f94/.github/PULL_REQUEST_TEMPLATE.md#L15

- https://github.com/strongloop/loopback-next/blob/862072b82ad88ec4b248e0b243dda58b53640f94/docs/site/DEVELOPING.md#L135-L141

An idea to consider: add a new `npm` script and use [cross-env](https://www.npmjs.com/package/cross-env) to set `CI=1`. Something along the following lines:

```
"test:full": "cross-env CI=1 npm test
```

packages/cli/test/acceptance/app-run.acceptance.js
```diff
@@ -14,8 +14,9 @@ const utils = require('../../lib/utils');
 const appGenerator = path.join(__dirname, '../../generators/app');
 const rootDir = path.join(__dirname, '../../../..');
 const sandboxDir = path.join(rootDir, 'sandbox');
+const {skipIf} = require('@loopback/testlab');
 
-describe('app-generator (SLOW)', function () {
+skipIf(process.env.CI == null, describe, 'app-generator (SLOW)', function () {
```
bajtos, line 19: Please add a code comment to explain what's happening here, why are we doing that and how can users run this test locally.

raymondfeng:
> Please update our developer/contrib documentation to explain users they need to run CI=1 npm t before submitting a pull request

We should let CI to catch any issues. Forcing developers to run `CI=1` before submitting a PR is not ideal - otherwise we should run this test locally without condition.

raymondfeng:
Added `cross-env` support.

PR #5747, opened 2020-06-16
fix: ensure unique sandbox directories for boot tests

raymondfeng:
Extracted from https://github.com/strongloop/loopback-next/pull/5744 for ease of review.

## Checklist

### DIFF
--- a/packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
@@ -10,12 +10,7 @@ import {BooterApp} from '../fixtures/application';
 
 describe('application metadata booter acceptance tests', () => {
   let app: BooterApp;
-  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'), {
-    // We intentionally use this flag so that `dist/application.js` can keep
-    // its relative path to satisfy import statements
-    subdir: false,
-  });
-
+  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'));
   beforeEach('reset sandbox', () => sandbox.reset());
   beforeEach(getApp);
 
@@ -33,14 +28,19 @@ describe('application metadata booter acceptance tests', () => {
     // Add the following files
     // - package.json
     // - dist/application.js
-    await sandbox.copyFile(resolve(__dirname, '../fixtures/package.json'));
+
     await sandbox.copyFile(
       resolve(__dirname, '../fixtures/application.js'),
       'dist/application.js',
+      // Adjust the relative path for `import`
+      content => content.replace('../..', '../../..'),
     );
+
+    await sandbox.copyFile(resolve(__dirname, '../fixtures/package.json'));
 
     const MyApp = require(resolve(sandbox.path, 'dist/application.js'))
       .BooterApp;
+
     app = new MyApp({
       rest: givenHttpServerConfig(),
     });
--- a/packages/boot/src/__tests__/acceptance/controller.booter.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/controller.booter.acceptance.ts
@@ -46,10 +46,6 @@ describe('controller booter acceptance tests', () => {
   }
 
   async function stopApp() {
-    try {
-      await app.stop();
-    } catch (err) {
-      console.log(`Stopping the app threw an error: ${err}`);
-    }
+    await app?.stop();
   }
 });
--- a/packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
@@ -165,7 +165,6 @@ module.exports = {
   }
 
   async function stopApp() {
-    if (app.state !== 'started') return;
-    await app.stop();
+    if (app?.state === 'started') await app?.stop();
   }
 });
--- a/packages/boot/src/__tests__/acceptance/component-application.booter.acceptance.ts
+++ b/packages/boot/src/__tests__/acceptance/component-application.booter.acceptance.ts
@@ -13,11 +13,7 @@ import {BooterApp} from '../fixtures/application';
 
 describe('component application booter acceptance tests', () => {
   let app: BooterApp;
-  const sandbox = new TestSandbox(resolve(__dirname, '../../.sandbox'), {
-    // We intentionally use this flag so that `dist/application.js` can keep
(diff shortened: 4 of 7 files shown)


packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
```diff
@@ -33,14 +29,18 @@ describe('application metadata booter acceptance tests', () => {
     // Add the following files
     // - package.json
     // - dist/application.js
+
+    const appJsFile = resolve(__dirname, '../fixtures/application.js');
+    let appJs = fs.readFileSync(appJsFile, 'utf-8');
+    // Adjust the relative path for `import`
+    appJs = appJs.replace('../..', '../../..');
```
bajtos, line 36: Can we extract this logic into `sandbox.copyFile` please? IIRC, `copyFile` is already detecting JS files and updating sourcemap URL to accommodate the new location, it would be great if we could do the same for import paths.

If it's too work to detect which paths to update, then perhaps `copyFile` can accept a new `options` to specify the mapping? Example usage:

```ts
await sandbox.copyFile(
  resolve(__dirname, '../fixtures/application.js'),
  'dist/application.js',
  {
    transform: text => text.replace('../..', '../../..'),
  },
});
```

WDYT?

packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
```diff
@@ -165,7 +165,6 @@ module.exports = {
   }
 
   async function stopApp() {
-    if (app.state !== 'started') return;
-    await app.stop();
+    if (app?.state === 'started') await app?.stop();
```
bajtos, line 168: Can you please preserve the reverse logic? I find such code easier to understand. I am also not sure how easy it is to set a breakpoint on `app?.stop` when it's on the same line as the `if` statement.

```ts
if (app?.state !== 'started') return;
await app?.stop();
```

The same comment applies to `packages/boot/src/__tests__/acceptance/model-api.booter.acceptance.ts` below too.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
raymondfeng: Maybe we should defer it. The current `copyFile` does not read the source file and we don't even know if the source is a text or binary file.

packages/boot/src/__tests__/acceptance/crud-rest.api-builder.acceptance.ts
raymondfeng: It's easy to set up a breakpoint on `if (app?.state === 'started') await app?.stop();` and step into `app.stop()`.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
bajtos: For context, here is the current `copyFile` implementation, you are right that we are not reading the source file.

https://github.com/strongloop/loopback-next/blob/862072b82ad88ec4b248e0b243dda58b53640f94/packages/testlab/src/test-sandbox.ts#L135-L146

I have two concerns about your current version:

(1)
The code is a bit involved and at different level of abstraction when compared the surrounding lines.

I'd like to read `getApp` as a sequence of instructions for copying files, I don't want to be distracted with implementation details of transforming source file content into target.

```ts
await sandbox.copyFile(resolve(__dirname, '../fixtures/application.js'), {
  // Adjust the relative path for `import`
  transform: text => text.replace('../..', '../../..'),
});
await sandbox.copyFile(resolve(__dirname, '../fixtures/package.json'));
```

(2) 
The code modifying `application.js` is duplicated in two test suites. These two places can easily get out of sync and it's likely additional copies will be introduced in the future as more test files are added.

Can you at least extract the code dealing with `application.js` into a shared test helper, to avoid duplication?

packages/boot/src/__tests__/acceptance/controller.booter.acceptance.ts
```diff
@@ -46,10 +46,6 @@ describe('controller booter acceptance tests', () => {
   }
 
   async function stopApp() {
-    try {
-      await app.stop();
-    } catch (err) {
-      console.log(`Stopping the app threw an error: ${err}`);
-    }
+    await app?.stop();
```
bajtos, line 49: Should we check the state first, for consistency with other `stopApp` implementations?

```suggestion
    if (app?.state === 'started') await app?.stop();
```

bajtos:
@raymondfeng FYI, it looks like these improvements are not enough to fix all booter tests when executed in parallel.

Can you please rebase your patch on top of the latest `master` and run `npm t -- -p` to check yourself?

Here are the test failures I am seeing: https://gist.github.com/bajtos/c2f45df9c41384c710d3712186910661

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
bajtos, line 36: IIUC, we are trying to fix the following line:

https://github.com/strongloop/loopback-next/blob/4e5f721ca666ab5047f4c36fbcb36a39b76966bb/packages/boot/src/__tests__/fixtures/application.ts#L10

An idea to consider: instead of changing the import/require path in the application file copied to sandbox, can we put a short `index.js` file into the directory where `../..` points to, to make `require('../..')` work again?

```js
// index.js
module.exports = require('../');
```

raymondfeng:
@bajtos I added `transform` option for `TestSandbox.copyFile`.

raymondfeng:
> @raymondfeng FYI, it looks like these improvements are not enough to fix all booter tests when executed in parallel.
Can you please rebase your patch on top of the latest master and run npm t -- -p to check yourself?
Here are the test failures I am seeing: https://gist.github.com/bajtos/c2f45df9c41384c710d3712186910661

I cannot reproduce the issue. With this PR, `npm t -- -p` works for me.

packages/boot/src/__tests__/acceptance/application-metadata.booter.acceptance.ts
raymondfeng: I prefer to use `transform` to make it clear - having a magic `index.js` makes it difficult to understand.

packages/boot/src/__tests__/acceptance/controller.booter.acceptance.ts
raymondfeng: In this case, app.start() is called.

PR #5705, opened 2020-06-10
feat(docs): add troubleshooting docs

dhmlau:
Related https://github.com/strongloop/loopback-next/issues/5701

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [x] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

Issue #5701, opened 2020-06-09
[Docs] Add Troubleshooting guide 


dhmlau:
Add troubleshooting guide to the docs.
https://github.com/dhmlau/learning-loopback/blob/master/troubleshooting.md might be a good starting point.

cc @raymondfeng @bajtos, per our meeting today.

## Acceptance criteria
- [ ] create a new section "Troubleshooting" - mention different perspectives
- [ ] add pages for different topics
  - [x] Tips from https://github.com/dhmlau/learning-loopback/blob/master/troubleshooting.md
  - [x] move https://loopback.io/doc/en/lb4/Running-and-debugging-apps.html to "Troubleshooting"
  - [x] VSC debugging
  - [x] `DEBUG=*` usage
  - [x] LB4 Sandbox usage - esp for users submitting issues - https://github.com/strongloop/loopback-next/blob/master/docs/site/DEVELOPING.md#using-monorepo-packages-as-dependencies
  - [ ] Open follow up story covering the following
    - [ ] Mocha related tips - running specific test, env vars, debugging, setting db
    - [ ] Memory profiling, CPU usage
    - [ ] Debugging in Docker

maaaaat:
Tips on profiling loopback’s startup would be huge right now. My company is running into cold start issues with our server less architecture that cause a bunch of 503s sent to clients as loopback gets ready to respond to requests. It would be monumental to know where to look and how to even start figuring out what needs fine tuning. 

Got a timeline on these guides or have any ideas we could try ahead of you guys publishing these guides?

dhmlau:
@maaaaat, we're not working on the profiling at the moment. If you have any tips or guides, feel free to publish it and reference it here. Or you can also enhance our docs by submitting a PR to add that info too. Thanks.

bajtos:
@maaaaat 

> Tips on profiling loopback’s startup would be huge right now. My company is running into cold start issues with our server less architecture that cause a bunch of 503s sent to clients as loopback gets ready to respond to requests. It would be monumental to know where to look and how to even start figuring out what needs fine tuning

I have had a good experience with https://www.npmjs.com/package/require-so-slow, you may want to give it a try.

jannyHou:
Created follow stories:
#6466 
#6465 
#6464 
#6463

jannyHou:
Closing as done. Let's use https://github.com/strongloop/loopback-next/issues/6465 for the further discussions about memory profiling and cpu usage.

PR #5748, opened 2020-06-16
feat(docs): add troubleshooting guide

dhmlau:
With the new acceptance criteria in https://github.com/strongloop/loopback-next/issues/5701, I'm taking a slightly different approach and making this PR as the first one.
- add the troubleshooting guide in the sidebar
- add "running and debugging app" and "setting debug strings" under it. 

Screen Shot 2020-06-15 at 9 56 24 PM

Related to https://github.com/strongloop/loopback-next/pull/5705. 

## Checklist

### DIFF
--- a/docs/site/Troubleshooting-guide.md
+++ b/docs/site/Troubleshooting-guide.md
@@ -1,0 +1,17 @@
+---
+title: 'Troubleshooting Guide'
+lang: en
+layout: page
+keywords:
+  LoopBack 4.0, LoopBack 4, Node.js, TypeScript, OpenAPI, Troubleshooting
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Troubleshooting-guide.html
+summary: Transactional API usage in LoopBack 4
+---
+
+The following documentation pages can help you troubleshoot problems you
+encountered.
+
+- [Running and debuggin apps](Running-and-debugging-apps.md)
+- [Setting debug strings](Setting-debug-strings.md)
--- a/docs/site/sidebars/lb4_sidebar.yml
+++ b/docs/site/sidebars/lb4_sidebar.yml
@@ -233,19 +233,6 @@ children:
       url: Database-migrations.html
       output: 'web, pdf'
 
-  - title: 'Running and debugging apps'
-    url: Running-and-debugging-apps.html
-    output: 'web, pdf'
-    children:
-
-    - title: 'Setting debug strings'
-      url: Setting-debug-strings.html
-      output: 'web, pdf'
-
-    - title: 'Using strong-error-handler'
-      url: Using-strong-error-handler.html
-      output: 'web, pdf'
-
   - title: 'Call Other Services'
     url: Calling-other-APIs-and-web-services.html
     output: 'web, pdf'
@@ -503,6 +490,11 @@ children:
   - title: 'Error handling'
     url: Error-handling.html
     output: 'web, pdf'
+    children:
+
+    - title: 'Using strong-error-handler'
+      url: Using-strong-error-handler.html
+      output: 'web, pdf'
 
   - title: 'Context'
     url: Context.html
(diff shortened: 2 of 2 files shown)


docs/site/Troubleshooting-guide.md
```diff
@@ -0,0 +1,17 @@
...
+layout: page
+keywords:
+  LoopBack 4.0, LoopBack 4, Node.js, TypeScript, OpenAPI, Troubleshooting
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Troubleshooting-guide.html
+summary: Transactional API usage in LoopBack 4
+---
+
+The following documentation pages can help you troubleshoot problems you
```
hacksparrow, line 13: "A list of troubleshooting guides for commonly encountered problems."?

bajtos:
Thank you @dhmlau for starting this effort!

It makes me wonder if we can align the troubleshooting guide with the new documentation system (#5549). I think a lot of the content is (and will be) in "how-to guide" style, in which case it should go to "Usage scenarios". Except nobody would look for troubleshooting guides there. Should we perhaps rename "Usage scenarios" to "How-to guides" first, and then nest troubleshooting guides under "How-to guides"?

agnes512:
> I think a lot of the content is (and will be) in "how-to guide" style, in which case it should go to "Usage scenarios". 

+1

dhmlau:
> Should we perhaps rename "Usage scenarios" to "How-to guides" first, and then nest troubleshooting guides under "How-to guides"?

+1 on renaming. 

I'm not sure about moving the troubleshooting guides under "how-to guides".  To me, the "how-to guides" helps the users to do certain thing; and the "troubleshooting guide" is used when something goes wrong. I guess we may not want to mix them up? 

Having said that, we can find the instruction for running a LB application a better home. Probably not "how-tos" either, because that's probably very first few things users need to know.

agnes512:
Maybe we can also rename the guide to something like "How to troubleshoot when something goes wrong" x)

bajtos:
Thank you for thoughtful comments.

I agree to keep "Troubleshooting" as the top-level sidebar entry then. Can we somehow document that these guides should be practical (describe how to troubleshoot a specific problem) and that content explaining troubleshooting-related content should go elsewhere?

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -807,6 +799,19 @@ children:
     url: Deploy-for-GDPR-readiness.html
     output: 'web, pdf'
 
+- title: 'Troubleshooting Guide'
```
bajtos, line 802: Personally, I would use a shorter title: "Troubleshooting", but I am ok with the longer version too.

If we decide to keep the longer form: Considering there are multiple pages nested under this entry, I think we should use plural (Troubleshooting Guides).

docs/site/Troubleshooting-guide.md
```diff
@@ -0,0 +1,17 @@
...
+encountered.
+
+- [Running and debuggin apps](Running-and-debugging-apps.md)
+- [Setting debug strings](Setting-debug-strings.md)
```
bajtos, line 17: Do we want the manual process of keeping this list up to date? I am will be one of the first people to forget to update it.

Is it possible to generate this list automatically? I vaguely remember that I saw a markdown page using Jekyll syntax to enumerate the list of child nodes, but I cannot find it now.

Do we actually need to list the pages here at all? I mean we already have a list of child nodes in the left sidebar. Is it perhaps possible to remove this page entirely and keep the "Troubleshooting Guide" item in the sidebar as an expander (a group name) only?

docs/site/sidebars/lb4_sidebar.yml
dhmlau: I like the shorter title 👍

docs/site/Troubleshooting-guide.md
dhmlau: > Is it possible to generate this list automatically? I vaguely remember that I saw a markdown page using Jekyll syntax to enumerate the list of child nodes, but I cannot find it now.

Let me find out. 

> Do we actually need to list the pages here at all? I mean we already have a list of child nodes in the left sidebar. Is it perhaps possible to remove this page entirely and keep the "Troubleshooting Guide" item in the sidebar as an expander (a group name) only?

I vaguely remember Jekyll expects each entry in the sidebar to have the `url`, i.e. need a page for each entry. But I can double check.

dhmlau:
Closing in favor of #6371

PR #5749, opened 2020-06-16
fix(repository-json-schema): added type 'object' to model json schema

### LINKED ISSUE #3804: Validation passes when an required complex property value is set to an empty string.
## Steps to reproduce

Create a model which references another model. e.g.

```ts
export class Order extends Entity {
  @property({
    type: 'number',
    id: true,
    generated: false,
  })
  id: number;

  @property({
    type: 'string',
  })
  orderNumber: string;

  @property({
    type: Customer,
    required: true,
  })
  customer: Customer;


  constructor(data?: Partial<Order>) {
    super(data);
  }
}

export class Customer extends Entity {
  @property({
    type: 'string',
    required: true,
  })
  name: string;
}

export class Customer extends Entity {
  @property({
    type: 'number',
    id: true,
    required: true,
    generated: false,
  })
  id: number;

  @property({
    type: 'string',
    required: true,
  })
  name: string;


  constructor(data?: Partial<Customer>) {
    super(data);
  }
}
```

Generate a datasource, repository and controller for `Order`.

## Current Behavior

Posting : 

```json
{
  "orderNumber": "123413",
  "customer": {} 
}
```

to the `/orders` endpoint correctly validates that customer is required, however posting:

```
{
  "orderNumber": "123413",
  "customer": ""
}
```

successfully creates a new order with an empty customer object.

## Expected Behavior

I would expect that the second post above would not pass validation and throw a 422 error and not create an order with an invalid customer object.

## Link to reproduction sandbox

https://github.com/SilexConsulting/loopback-next/tree/validation-should-fail-empty-required-object
npm test will have a single failing test to demonstrate the issue.

## Additional information
I'm sure the issue is not relevant to the platform or node/npm version but it is:
linux x64 11.8.0

loopback version tested is: 
├── @loopback/boot@1.5.6
├── @loopback/context@1.23.0
├── @loopback/core@1.10.2
├── @loopback/openapi-v3@1.9.7
├── @loopback/repository@1.14.0
├── @loopback/rest@1.19.0
├── @loopback/rest-explorer@1.3.7
└── @loopback/service-proxy@1.3.6

I have confirmed that the behaviour is the same with the head of master.

I believe that this behaviour is due to the fact that the JSON schema generated by repository-json-schema omits the type: object for LB models. I think from the documentation here: https://json-schema.org/understanding-json-schema/structuring.html that complex shemas should have a type:object property.

I have made a change in my fork of LB here: https://github.com/SilexConsulting/loopback-next/commit/444407603035eb284675026cee75716192149e2e which produces the expected results and all the modules tests pass.

ShaunSpinelli:
#3804

Added type 'object' to generated model schema to allow `ajv` validation to work correctly.

Fixes #3804 

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
@@ -135,6 +135,14 @@ describe('Validation at REST level', () => {
       await client.post('/products').type('json').send('{}').expect(422);
     });
 
+    it('rejects requests with empty string body', async () => {
+      await client.post('/products').type('json').send('""').expect(422);
+    });
+
+    it('rejects requests with string body', async () => {
+      await client.post('/products').type('json').send('"pencils"').expect(422);
+    });
+
     it('rejects requests with null body', async () => {
       await client.post('/products').type('json').send('null').expect(400);
     });
@@ -590,6 +598,52 @@ describe('Validation at REST level', () => {
       });
     });
   });
+
+  context(
+    'for request body specified via model definition with strict false',
+    () => {
+      @model({settings: {strict: false}})
+      class ProductNotStrict {
+        @property()
+        id: number;
+
+        @property({required: true})
+        name: string;
+
+        constructor(data: Partial<ProductNotStrict>) {
+          Object.assign(this, data);
+        }
+      }
+
+      class ProductController {
+        @post('products')
+        async create(
+          @requestBody({required: true}) data: ProductNotStrict,
+        ): Promise<ProductNotStrict> {
+          return new ProductNotStrict(data);
+        }
+      }
+
+      before(() => givenAnAppAndAClient(ProductController));
+      after(() => app.stop());
+
+      it('rejects requests with empty string body', async () => {
+        await client.post('/products').type('json').send('""').expect(422);
+      });
+
+      it('rejects requests with string body', async () => {
+        await client
+          .post('/products')
+          .type('json')
+          .send('"pencil"')
+          .expect(422);
+      });
+
+      it('rejects requests with null body', async () => {
+        await client.post('/products').type('json').send('null').expect(400);
+      });
+    },
+  );
 
   // A request body schema can be provided explicitly by the user
   // as an inlined content[type].schema property.
--- a/packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
+++ b/packages/openapi-v3/src/__tests__/integration/controller-spec.integration.ts
@@ -76,17 +76,20 @@ describe('controller spec', () => {
         schemas: {
           Bar: {
             title: 'Bar',
+            type: 'object',
             properties: {name: {type: 'string'}},
             additionalProperties: false,
           },
           Baz: {
             title: 'Baz',
+            type: 'object',
             properties: {name: {type: 'string'}},
             additionalProperties: false,
           },
           Foo: {
             // guarantee `definition` is deleted
             title: 'Foo',
+            type: 'object',
             properties: {
               bar: {$ref: '#/components/schemas/Bar'},
               baz: {$ref: '#/components/schemas/Baz'},
@@ -562,6 +565,7 @@ describe('controller spec', () => {
       },
       additionalProperties: false,
       title: 'MyModel',
+      type: 'object',
     };
 
     it('generates schema for response content', () => {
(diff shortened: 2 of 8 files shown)


bajtos:
From https://github.com/strongloop/loopback-next/pull/3828#pullrequestreview-294984805

- Loopback generated JSON Schema omits this declaration. build-schema.ts:353.
- Omitting this declaration causes ajv validation to pass when a required element (which is a reference to a different object) is set to null.
- Adding the declaration: "type: 'object'" to generated schemas enables ajv validation to correctly fail when a required referenced property is present but the value set to null.

From https://github.com/strongloop/loopback-next/pull/3828#issuecomment-537060985

I was thinking about adding a new test to `packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts`. Apparently, we already have a test to verify that request bodies containing only `null` are rejected:

https://github.com/strongloop/loopback-next/blob/512d37d4f8165ab02c522f071eb1862bf57125b8/packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts#L95-L101

If I understand your scenario correctly, then you want to reject requests when a required property is set to `null`. I think the simplest option is to define such property in the existing `Product` model and then add a new test to verify what happens when it's set to `null`. Just make sure to make the new property optional so that existing tests pass.

Alternatively, you can create a new `context()` block where you define a model & a controller that are specific to your scenario only. This is better in the long run, because we don't want the existing `Product` controller to grow super large to cover all different property types & definitions.

From https://github.com/strongloop/loopback-next/pull/3828#discussion_r334774531

I think the change in behavior may be caused by https://github.com/strongloop/loopback-next/pull/3895, where we started to add `additionalProperties: false` to schema emitted for models in strict mode (which is the default).

When I change the definition of `Product` model in the acceptance test and disable strict mode, then your new test fails as expected. (Unfortunately, there is one more test that fails too.)

Obviously, we don't want to change all existing tests to start testing the scenario with `strict: false` and no longer test the default setup with `strict: true`. To support non-strict mode, we need to create a new group of tests in `validation.acceptance.ts` that will use a new model with `strict: false`.

Let me ask you a question: how are your models configured? Do you use the default "strict" model behavior, or are you disabling it via model settings? If you are using strict mode, then I think you should be find a this pull request is not needed for your app.

bajtos:
IMO, we need to add tests to cover the following scenarios:

- For an Entity (model) defined with `strict: false`, test validation of values `''` (an empty string - #3804), `null` (#3828).
- For an Entity (model) defined with `strict: true`, test validation of values `''` and `null`.

I feel it's crucial to verify the outcome of AJV validation before the proposed changes and after the change to ensure we are getting the desired behavior. I consider the changes made in the existing tests as "unit-level" - we know what schema is emitted, but don't verify that is give the desired behavior during validation.

ShaunSpinelli:
Hi @bajtos, Thanks for the feedback. I have updated the tests for 

> - For an Entity (model) defined with strict: false, test validation of values '' (an empty string - #3804), null (#3828).
> - For an Entity (model) defined with strict: true, test validation of values '' and null.

And in addition added tests for when the body of the request is a string, which was the initial reason I discovered the issue.

Let me know if these tests are what you were expecting and if there is any other work that required. Cheers

packages/rest/src/__tests__/acceptance/validation/validation.acceptance.ts
```diff
@@ -591,6 +599,52 @@ describe('Validation at REST level', () => {
...
+          Object.assign(this, data);
+        }
+      }
+
+      class ProductController {
+        @post('products')
+        async create(
+          @requestBody({required: true}) data: ProductNotStrict,
+        ): Promise<ProductNotStrict> {
+          return new ProductNotStrict({id: 1, name: 'pencil'});
```
bajtos, line 623: Can you please add a code comment to explain why you are using hard-coded model properties instead of passing the `data` argument? I assume this is needed to bypass validation at juggler level and ensure we reject the request at REST API layer.

jannyHou: @ShaunSpinelli same question as @bajtos mentioned, any reason why not use:

```ts
return new Product(data);
```
?

ShaunSpinelli: Hey @bajtos and @jannyHou. Thanks for the feedback,  that was a mistake and it has been corrected to: 

```javascript
return new ProductNotStrict(data);
```

Issue #5789, opened 2020-06-19
Does not support Node <=v10.15.0


andrey-abramow:
Steps to reproduce:
- use node v10.13.0
- lb4 pr_name (@loopback/cli version: 2.9.0)
- cd  pr_name
- yarn start
```
Cannot start the application. TypeError: events_1.once is not a function
    at HttpServer.start (/Users/andreyabramow/Development/loopback-4-db-test/node_modules/@loopback/http-server/src/http-server.ts:122:20)
    at RestServer.start (/Users/andreyabramow/Development/loopback-4-db-test/node_modules/@loopback/rest/src/rest.server.ts:958:28)
    at process._tickCallback (internal/process/next_tick.js:68:7)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
`events.once` added to a node in version 10.16,  so this code

```
 /**
   * Starts the HTTP / HTTPS server
   */
  public async start() {
    console.log('SSS')

    this.server.listen(this.serverOptions);
    try {
      await once(this.server, 'listening'); // error
      this._listening = true;

      const address = this.server.address();
      assert(address != null);
      this._address = address!;
    } catch(e) {
        console.log(e)
    }
  }
```
from https://github.com/strongloop/loopback-next/blob/master/packages/http-server/src/http-server.ts#L7

does not work in older versions of node (10 < v < 10.6).

Can you please just update documentation something like
**Node.js >= 10.16**
And adding some checking of the version to `cli` would be great as well.
Thanks!

bajtos:
@andrey-abramow good catch, thank you for reporting the problem! I was not aware that `events.once` was added in a later version of Node.js 10.x when I removed `p-once` in https://github.com/strongloop/loopback-next/pull/5429.

Would you like to contribute this change yourself? I would start with updating `engines` field in all packages changed by #5429:

```diff
- "node": ">=10" 
+ "node": ">=10.16"
```

- `benchmark/package.json`
- `examples/express-composition/package.json`
- `examples/lb3-application/package-lock.json`
- `examples/lb3-application/package.json`
- `examples/passport-login/package.json`
- `examples/rpc-server/package.json`
- `extensions/cron/package.json`
- `extensions/health/package.json`
- `extensions/logging/package.json`
- `extensions/metrics/package.json`
- `packages/booter-lb3app/package.json`
- `packages/booter-lb3app/package.json`
- `packages/cli/package.json`
[...]

PR #5795, opened 2020-06-20
fix: set node version to >=10.16 to support events.once

andrey-abramow:
In PR https://github.com/strongloop/loopback-next/pull/5429 was added `require('events').once` which is only available in Node  >=10.16.

Fixes https://github.com/strongloop/loopback-next/issues/5789

## Checklist

     (**for node >=10.16**)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ for more details.
 
 Make sure you have the following installed:
 
-- [Node.js](https://nodejs.org/en/download/) >= 10
+- [Node.js](https://nodejs.org/en/download/) >= 10.16
 
 Install LoopBack 4 CLI to help create new projects as follows:
 
--- a/acceptance/extension-logging-fluentd/package.json
+++ b/acceptance/extension-logging-fluentd/package.json
@@ -6,7 +6,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "engines": {
-    "node": ">=10"
+    "node": ">=10.16"
   },
   "scripts": {
     "build": "lb-tsc",
--- a/acceptance/repository-cloudant/package.json
+++ b/acceptance/repository-cloudant/package.json
@@ -6,7 +6,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "engines": {
-    "node": ">=10"
+    "node": ">=10.16"
   },
   "scripts": {
     "build": "lb-tsc",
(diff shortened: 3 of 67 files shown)


andrey-abramow:
@aharbis @bajtos 
I can find node version check  only in `cli`
```
// Make sure node version meets the requirement. This code intentionally only
// uses ES5 features so that it can be run with lower versions of Node
// to report the version requirement.
const nodeVer = process.versions.node;
const requiredVer = pkg.engines.node;
const ok = semver.satisfies(nodeVer, requiredVer);
if (!ok) {
  const format = 'Node.js %s is not supported. Please use a version %s.';
  console.error(format, nodeVer, requiredVer);
  process.exit(1);
}
```
Should the `@loopback/core` have this check as well? Or maybe,  as all the packages can be used separately we should add this everywhere?

andrey-abramow:
@bajtos Thanks for the comment. 

You listed 17 packages because a duplication of `packages/booter-lb3app/package.json` and a` package-lock`

It looks like I have updated all files form the PR.

You can check it:
```
 https://github.com/strongloop/loopback-next/pull/5795/files
 https://github.com/strongloop/loopback-next/pull/5429/files
```
`[].map.call(document.getElementsByClassName('link-gray-dark'), function (el) { return el.text } ).filter((file) => file && file.match(/package.json/)).sort()`
Will provide the same results.

Any way I can update all packages by the script:
**node ./bin/update-monorepo-node-version.js '>=10.16'**
```

'use strict';

const Project = require('@lerna/project');
const fs = require('fs');
async function main() {
  const version = process.argv.slice(2)[0];
  if (!version) throw new Error('Specify new version in args!');

  const project = new Project(process.cwd());
  const packages = await project.getPackages();
  for (const p of packages) {
    const packageJson = require(p.manifestLocation);
    if (packageJson.engines.node !== version) {
      const newPackageJson = {
        ...packageJson,
        engines: {
          ...packageJson.engines,
          node: version,
        },
      };
      fs.writeFile(
        p.manifestLocation,
        JSON.stringify(newPackageJson, null, "\t") + '\n', {
          encoding: 'utf-8',
        },
        () => {
          console.log(
            `${p.manifestLocation} node version change${packageJson.engines.node} to ${version}`,
          );
        },
      );
    }
  }
}

main().catch(err => {
  console.error(err);
  process.exit(1);
});

```
Should I do this? Maybe it is better not to update all packages because some of them can be used separately? Maybe I am wrong.

raymondfeng:
@andrey-abramow @bajtos I have fixed all occurrences using `VSCode - Replace in files`. The latest changes are forced-pushed back to your branch.

PR #5801, opened 2020-06-22
feat: add TypeORM support

hacksparrow:
Taking the work in https://github.com/strongloop/loopback-next/pull/4794 further, this PR adds TypeORM support in LB4. You can now use TypeORM entities and repositories in LB4.

The feature is implemented and working, for usage details refer to https://github.com/strongloop/loopback-next/blob/c7157f85cb8b05a43fdf59eabfd72892bf9a6662/packages/typeorm/README.md.

What's remaining?

- CLEAN UP the repository
- Write tests
- Generate OpenAPI spec
- Convert filters to TypeORM query (everything not required in the first version, just need to develop a functional high-level idea about how to go about it)
- Add documentation to website

## Checklist


packages/typeorm/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) IBM Corp. 2017,2019.
```
achrinza, line 1: Inconsistent copyright.

Ditto for the other copyright headers.

packages/typeorm/LICENSE
```diff
@@ -0,0 +1,25 @@
...
+Node module: @loopback/repository
```
achrinza, line 2: Inconsistent node module name header.

Ditto for the other module name headers.

hacksparrow:
> Also, should we have a booter for TypeORM entities, similar to the ones for lb4 models?

Yes, there's a booter - https://github.com/strongloop/loopback-next/pull/5801/files#diff-2a23c8ed814e1a88b05bcacbeec6689c.

packages/typeorm/LICENSE
hacksparrow: Haha, yea need to clean up.

tsconfig.json
```diff
@@ -174,6 +174,12 @@
     },
     {
       "path": "packages/tsdocs/tsconfig.json"
+    },
+    {
+      "path": "packages/typeorm/tsconfig.json"
+    },
+    {
+      "path": "sandbox/appa/tsconfig.json"
```
raymondfeng, line 182: Please `rm -rf sandbox/appa`.

extensions/typeorm/src/typeorm.component.ts
```diff
@@ -3,18 +3,51 @@
...
+//   ContextObserver,
+//   CoreBindings,
+//   createBindingFromClass,
+//   filterByKey,
+//   filterByTag,
+//   inject,
+//   Server,
+//   Subscription,
+// } from '@loopback/core';
+
```
raymondfeng: Please remove the commented out code.

packages/typeorm/src/decorators/typeorm.decorator.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import {Context, inject, Injection, ResolutionSession} from '@loopback/core';
+import {QueryRunner} from 'typeorm';
+import {TypeOrmBindings} from '../keys';
+
+export namespace typeorm {
+  export function connection(connectionName?: string) {
+    return inject(
+      '',
+      {decorator: '@typeormConnection'},
```
raymondfeng, line 14: We typically use `typeorm.connection` as the name.

tsconfig.json
hacksparrow: Yea, need to clean up.

packages/typeorm/src/decorators/typeorm.decorator.ts
hacksparrow: Indeed. That's a remnant of something I was trying, will clean it up.

extensions/typeorm/src/typeorm.component.ts
hacksparrow: Will do.

packages/typeorm/src/types.ts
```diff
@@ -0,0 +1,120 @@
...
+  CockroachConnectionOptions,
+  CockroachDriver,
+  CockroachQueryRunner,
+  CordovaConnectionOptions,
+  CordovaDriver,
+  CordovaQueryRunner,
+  ExpoConnectionOptions,
+  ExpoDriver,
+  ExpoQueryRunner,
+};
```
bajtos, line 64: Do we really need that long list of exports 👆 and the long list of imports 👇 ? Isn't it enough to `export * from 'typeorm';`?

packages/typeorm/src/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+import {JsonSchemaOptions, SchemaRef} from '@loopback/rest';
+
+/**
+ * Describe the provided Entity as a reference to a definition shared by multiple
+ * endpoints. The definition is included in the returned schema.
+ *
+ * @param entityCtor - The entity constructor (e.g. `Product`)
+ * @param options - Additional options
+ */
+export function getModelSchemaRef<T extends object>(
```
bajtos, line 15: As discussed in the call, I am proposing to drop `Ref` part and emit regular JSON schema. The `ref`-based schemas were needed before OpenAPI schema consolidation was introduced. I believe with consolidation in place, we no longer need to worry about avoiding duplicated schema definitions in controller methods.

```ts
export function modelToJsonSchema<T extends object>(
  ctor: Function & {prototype: T},
  jsonSchemaOptions: JsonSchemaOptions<T> = {},
): JsonSchema {
  const allColumns: ColumnMetadataArgs[]= getMetadataArgsStorage().columns;
  const modelColumns = allColumns.filter(c => c.target === modelCtor);
  const schema: JsonSchema = {
    // set title (to modelCtor.modelName?), etc.
    properties: {}
  };
  for (c of modelColumns) {
    schema.properties[c.propertyName] = {
      type: c.options.type,
    }
  }
  return schema;
}
// getJsonSchema(modelCtor) --> call modelToJsonSchema under the hood + cache the result
// getModelSchema(modelCtor) --> convert getJsonSchema() to OpenApiSpec format
```

bajtos:
> Convert filters to TypeORM query (everything not required in the first version, just need to develop a functional high-level idea about how to go about it)

Posting few ideas from slack chat:

Juggler implementation starts here: https://github.com/strongloop/loopback-connector/blob/7d43b461d13e2e67fd07cee4ae22c1e19d0ea848/lib/sql.js#L1332-L1362

For TypeORM, we want something like this:

```
{where: {name: 'miroslav'}, limit: 10} => createQueryBuilder().where('user.name = :name', {name: 'miroslav'}).limit(10);
```

IMO, TypeORM users should not be forced to require `@loopback/repository` to get the `Filter` object (and also `getFilterSchemaFor`). In the spike, it may be ok to copy the relevant type descriptions & schema builder from `@loopback/repository` to your new extension. After the spike, I think we should extract `Filter`, `getFilterSchemaFor` and other related code into a new shared package, e.g. `@loopback/filter`.

In LB3, we created [loopback-filters](https://github.com/strongloop/loopback-filters) package with a similar intention, but it was half-baked and quickly got out of sync with the main filtering logic implemented in the memory connector. So I think it's a valid move to start building a shared filtering package again, and get it right this time :)

packages/boot/src/booters/typeorm-connection.booter.ts
```diff
@@ -0,0 +1,59 @@
...
+    @inject(CoreBindings.APPLICATION_INSTANCE)
+    public app: ApplicationUsingTypeOrm,
+    @inject(BootBindings.PROJECT_ROOT) projectRoot: string,
+    @config()
+    public entityConfig: ArtifactOptions = {},
+  ) {
+    super(
+      projectRoot,
+      // Set TypeORM connection options if passed in via bootConfig
+      Object.assign({}, ConnectionDefaults, entityConfig),
```
raymondfeng, line 36: This is wrong - all properties are merged into the 1st arg and we won't be able to access it afterward.

packages/typeorm/src/mixins/typeorm.mixin.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Type alias for Node.js options object
+ */
+export type Options = AnyObject;
+
+/**
+ * Objects with open properties
+ */
+export interface AnyObject {
```
raymondfeng, line 62: Let's use `Record<string, any>` instead.

packages/boot/src/booters/typeorm-connection.booter.ts
```diff
@@ -0,0 +1,59 @@
...
+    for (const file of this.discovered) {
+      const exported = require(file);
+      const connectionKey = Object.keys(exported)[0];
+      if (connectionKey) {
+        const options = exported[connectionKey];
+        this.app.typeormConnectionOptions.push(options);
+      }
+    }
+  }
+}
```
raymondfeng, line 50: I don't see test coverage for the booter.

packages/boot/src/booters/typeorm-connection.booter.ts
```diff
@@ -0,0 +1,59 @@
...
+      projectRoot,
+      // Set TypeORM connection options if passed in via bootConfig
+      Object.assign({}, ConnectionDefaults, entityConfig),
+    );
+  }
+
+  async load() {
+    for (const file of this.discovered) {
+      const exported = require(file);
+      const connectionKey = Object.keys(exported)[0];
```
raymondfeng, line 43: This is not clear to me how a `connection` artifact look like.

packages/typeorm/src/types.ts
bajtos: Fair enough 🤷

packages/boot/src/booters/typeorm-connection.booter.ts
bajtos: I believe `Object.assign` returns the 1st argument and therefore this line works as expected

```
$ node 
> console.log(Object.assign({}, {foo: 'bar'}))
{ foo: 'bar' }
undefined
```

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+**NOTE**: This module is experimental and evolving.
+
+## Installation
+
+```sh
+npm install --save @loopback/typeorm
+```
+
+## Usage
```
bajtos, line 15:
```suggestion
## Basic use
```

See https://loopback.io/doc/en/contrib/README-guidelines.html

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+databases. Define the connection in files with a `.connection.ts` extension and
+keep them in the `connections` directory of the project.
+
+```ts
+// src/connections/sqlite.connection.ts
+import path from 'path';
+import {SqliteConnectionOptions} from '@loopback/typeorm';
+import {Book} from '../entities/';
+
+export const SQLLITE_CONNECTION: SqliteConnectionOptions = {
```
bajtos, line 75: Can we use `ConnectionOptions` so that `@loopback/typeorm` does not have to re-export internal TypeORM files like `typeorm/driver/mysql/MysqlConnectionOptions`?

Note that `ConnectionOptions` is a [discriminated union type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions), it includes `SqliteConnectionOptions` as the variant used when `type` is `sqlite`. As a result, the compiler is able to verify that all SQLite-related fields are correctly set, even when using `ConnectionOptions` as the type.

Example file:

```ts
import {ConnectionOptions} from 'typeorm';
export const config: ConnectionOptions = {
  type: 'sqlite',
};
```

Compiler output:

```
$  tsc x.ts
x.ts:2:14 - error TS2322: Type '{ type: "sqlite"; }' is not assignable to type 'ConnectionOptions'.
  Property 'database' is missing in type '{ type: "sqlite"; }' but required in type 'SqliteConnectionOptions'.

2 export const config: ConnectionOptions = {
               ~~~~~~

  node_modules/typeorm/driver/sqlite/SqliteConnectionOptions.d.ts:13:14
    13     readonly database: string;
                    ~~~~~~~~
    'database' is declared here.


Found 1 error.
```

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+Controllers continue to work as usual. And you don't have to create repositories
+since TypeORM creates them for you; just inject them in the controllers. The
+repository API is 100% TypeORM
+[repository API](https://typeorm.io/#/repository-api).
+
+```ts
+// src/controllers/book.controller.ts
+import {get, post, Request, requestBody} from '@loopback/rest';
+import {getModelSchemaRef, Repository, typeorm} from '@loopback/typeorm';
```
bajtos, line 104: As discussed before, this should be just `getModelSchema`. (No need to use `$ref` schemas anymore, OpenAPI consolidation will take care of de-duplicating schemas.)

IMO, `@loopback/typeorm` should not be coupled with OpenAPI, I think we should either tell users to apply `jsonToSchemaObject` on the schema produced by `@loopback/typeorm`, or else introduce a new package for things that are specific to TypeORM+OpenAPI.

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+  constructor() {}
+
+  // Create a new book
+  @post('/book')
+  async create(@requestBody() data: Book) {
+    const book = new Book();
+    book.title = data.title;
+    book.published = false;
+    return await this.bookRepo.save(book);
```
bajtos, line 118: Can we use `repository.create` API?

```ts
this.bookRepo.create(data);
```

See https://typeorm.io/#/repository-api

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  // Find book by title
+  @get('/note/{title}')
+  async findByTitle(@param.path.string('title') title: string) {
+    return await this.bookRepo.find({title});
```
bajtos, line 124: Please add a comment to make it clear to readers that we are using TypeORM filtering syntax here, not LoopBack's style.

Based on what I found in the docs ([Find Options](https://typeorm.io/#/find-options)), I think this line should use `where`.

```suggestion
    return await this.bookRepo.find({where:{title}});
```

packages/typeorm/README.md
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+```
+
```
bajtos, line 128: Please add a new section to list known limitations - things/features/use-cases we know about but not implemented yet.

Few candidates:

- Full JSON/OpenAPI schema for models, including variants like with/without id, with/without relations, partial, etc.
- Json/OpenAPI schema to describe the supported `filter` format
- Support for LoopBack-style filters
- Custom repository classes (e.g. to implement `bookRepo.findByTitle(title)`).
- Database migration

Scenarios we may need to verify and document:

- When using multiple connections, how to specify which model is backed by which connection? E.g. `Category` and `Product` is accessed via `db` connection, while `Customer` and `Address` is accessed via `crm` connection.

packages/typeorm/src/decorators/typeorm.decorator.ts
```diff
@@ -0,0 +1,67 @@
...
+      async (
+        ctx: Context,
+        injection: Readonly<Injection>,
+        session: ResolutionSession,
+      ) => {
+        return getConnection(ctx, connectionName);
```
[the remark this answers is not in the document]
bajtos: > IIUC, this creates "virtual" bindings that can be injected but that are not tracked by Context and therefore not visible in context-inspector UI.
> 
> Can we find a way how to create regular bindings for individual connections please?

☝️ Please make sure this comment is either addressed as part of this original PR or there is a follow-up issue to fix binding inspection ASAP.

packages/typeorm/src/mixins/typeorm.mixin.ts
```diff
@@ -0,0 +1,57 @@
...
+import debugFactory from 'debug';
+import {ConnectionOptions} from 'typeorm';
+
+const debug = debugFactory('loopback:typeorm:mixin');
+
+export function TypeOrmMixin<T extends MixinTarget<Application>>(
+  superClass: T,
+) {
+  return class extends superClass {
+    typeormConnectionOptions: ConnectionOptions[] = [];
```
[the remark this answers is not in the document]
bajtos: > As we discussed in the call, I find the design where we keep an index-ordered list of connection options as suboptimal. Instead, I am proposing to store TypeORM's connection manager instance on the application object and create a method like `app.connection` to register new connections with the app. The API should also allow the user to give each connection a name.

❗ IMO, this comment must be resolved before this pull request can be landed.

packages/typeorm/src/mixins/typeorm.mixin.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
+      super(...args);
+      this.component(TypeOrmComponent);
+    }
+
+    async migrateSchema(options: SchemaMigrationOptions = {}): Promise<void> {
+      // TODO: implement using TypeORM
```
bajtos, line 26: Please throw an error instead of skipping the migration silently.

packages/typeorm/src/services/connections.ts
```diff
@@ -0,0 +1,42 @@
...
+  lifeCycleObserver,
+} from '@loopback/core';
+import {ConnectionManager} from 'typeorm';
+import {ApplicationUsingTypeOrm, TypeOrmBindings} from '../';
+
+@lifeCycleObserver('datasource', {
+  tags: {[ContextTags.KEY]: TypeOrmBindings.CONNECTIONS},
+  scope: BindingScope.SINGLETON,
+})
+export class TypeOrmConnections implements LifeCycleObserver {
```
bajtos, line 21: As mentioned elsewhere, a class should have a singular name. At minimum, use something like `TypeOrmConnectionManager`. Preferably, get rid of this class entirely as discussed in https://github.com/strongloop/loopback-next/pull/5801/files#r444168243

packages/typeorm/src/services/connections.ts
```diff
@@ -0,0 +1,42 @@
...
+    for (const option of connectionOptions) {
+      const connection = this.manager.create(option);
+      await connection.connect();
+    }
+  }
+
+  async stop() {
+    for (const connection of this.manager.connections) {
+      await connection.close();
+    }
```
bajtos, line 40: IMO, if we follow the direction outlined in https://github.com/strongloop/loopback-next/pull/5801/files#r444168243, then shutdown should be implemented as follows:

- TypeOrmMixin registers a new LifeCycleObserver
- The observer obtains the connection manager via `app.manager` (or `this.manager`, depending on how is the observer implemented)
- The rest remains the same (all connections are iterated and closed).

packages/typeorm/src/services/connections.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
```
bajtos, line 41: Should we close the connections in parallel?

```suggestion
  stop() {
    return Promise.all(
      this.manager.connections.map(c => c.close())
    );
  }
```

packages/typeorm/src/utils.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright IBM Corp. 2017,2020. All Rights Reserved.
+// Node module: @loopback/typeorm
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {JsonSchemaOptions, SchemaRef} from '@loopback/rest';
```
bajtos, line 6: ❗ `@loopback/typeorm` should not require consumers to use `@loopback/rest`.

At minimum, change this line to import OpenAPI types from `@loopback/openapi-v3`.

As I mentioned above, I think it would be best to move OpenAPI-related parts into a new package, but I guess that can be done in a follow-up pull request.

packages/typeorm/src/utils.ts
bajtos: > As discussed in the call, I am proposing to drop `Ref` part and emit regular JSON schema. The `ref`-based schemas were needed before OpenAPI schema consolidation was introduced. I believe with consolidation in place, we no longer need to worry about avoiding duplicated schema definitions in controller methods.

❗I'd like you to address my comment before the PR is landed.

bajtos: If it's too much work to get even a minimal conversion working, then I am proposing the following:

- rename `getModelSchemaRef` to `getModelSchema` 
- change the return type from `SchemaRef` to `SchemaObject`
- change the implementation to return a "dummy" schema describing any object value

```ts
export function getModelSchemaRef<T extends object>(
  entityCtor: Function & {prototype: T},
  options?: JsonSchemaOptions<T>,
): SchemaRef {
  // TODO: generate proper OpenAPI schema describing entity properties, etc.
  return {
    title: entityCtor.name,
    type: 'object'
  };
}
```

Personally, I'll try to get rid of `SchemaRef` (and thus the dependency on `@loopback/rest`) by removing the description of the return type or using `object` instead. For the first iteration, we don't need `options` either, since they are not supported anyways.

```ts
export function getModelSchemaRef<T extends object>(
  entityCtor: Function & {prototype: T},
) {
  return { 
    // ...
  };
}
```

packages/typeorm/src/utils.ts
hacksparrow: Note to self: Create follow up task.

packages/typeorm/README.md
hacksparrow: The object is not being saved for some reason; don't want to troubleshoot right now, let's handle the highest priority items first. Will give a try again later today, if not, will open a follow up task.

Note to self: Create follow up task.

packages/typeorm/README.md
hacksparrow: > IMO, @loopback/typeorm should not be coupled with OpenAPI,

Need elaboration.

packages/boot/src/booters/typeorm-connection.booter.ts
hacksparrow: Done.

packages/typeorm/src/decorators/typeorm.decorator.ts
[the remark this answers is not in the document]
hacksparrow: Note to self: Create follow up task.

packages/typeorm/README.md
hacksparrow: TypeORM syntax looks like `"title": title`, `name: "Yahoo"`.

packages/typeorm/src/utils.ts
```diff
@@ -3,22 +3,70 @@
...
+function getStringifiedType(func: ColumnType): string {
+  if (func === Number) {
+    return 'number';
+  } else if (func === String) {
+    return 'string';
+  } else if (func === Boolean) {
+    return 'boolean';
+  } else {
+    throw new Error(
+      `Type conversion of ${func} to OpenAPI format not implemented.`,
```
bajtos, line 69: UX improvement:

- Please include the entity name and the property name in the error message.

Discussion point:

In your proposal, applications cannot use unsupported types at all. Isn't that too limiting, considering that those types works at ORM layer and we just don't know how to describe them in OpenAPI? How about changing this line to print a warning and return a value saying the property does not have any type constraints? (I think `undefined` should work.)

packages/typeorm/README.md
bajtos: `@loopback/openapi-v3` provides things like `@get` and `@api` decorators.

As a user building gRPC or Pub/Sub application using `@loopback/typeorm` for persistence, why do I have to install dependencies for OpenAPI, when I am not using any of that?

bajtos: Also if we add `gRPC` support in the future and implement code to generate gRPC schemas (protocol buffer message type definitions) from TypeORM entities, are we going to add `gRPC` to `@loopback/typeorm` dependencies and thus force REST/OpenAPI users to install gRPC they are not using?

Anyhow, I think this is out of scope of this pull request, but please do create a follow-up task to continue this discussion and rework the code accordingly.

packages/typeorm/README.md
bajtos: Perhaps add a code comment to this code snippet, to explain the problem to our readers too?

packages/typeorm/README.md
bajtos: Weird, their documentation is showing something else. https://typeorm.io/#/find-options

Did you verify that your implementation returns only the books with the given title and no other book entries?

I don't mind how we write the filter as long as it works.

packages/typeorm/README.md
bajtos: @hacksparrow I think your list of limitations is missing the following items:

- Json/OpenAPI schema to describe the supported filter format
- Support for LoopBack-style filters
- Custom repository classes (e.g. to implement bookRepo.findByTitle(title)).
- When using multiple connections, how to specify which model is backed by which connection? E.g. Category and Product is accessed via `db` connection, while Customer and Address is accessed via `crm` connection.

packages/typeorm/src/utils.ts
hacksparrow: Done.

```ts
  } else {
    debug(
      `${entity}.${property}: Type conversion of ${func} to OpenAPI format not implemented.`,
    );
    return undefined;
  }
```

packages/typeorm/src/mixin.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface ApplicationUsingTypeOrm extends Application {
+  typeormConnectionOptions: ConnectionOptions[];
+  connection(options: ConnectionOptions): void;
+  migrateSchema(): Promise<void>;
+}
+
+@lifeCycleObserver('datasource', {
+  scope: BindingScope.SINGLETON,
+})
+export class StartStop implements LifeCycleObserver {
```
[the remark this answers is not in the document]
hacksparrow: Using `TypeOrmLifeCycleManager`.

packages/typeorm/README.md
hacksparrow: > When using multiple connections, how to specify which model is backed by which connection? E.g. Category and Product is accessed via db connection, while Customer and Address is accessed via crm connection.

Like this:

```ts
export const AssetConnection: ConnectionOptions = {
  name: 'db',
  type: 'sqlite',
  database: 'assets',
  entities: [Category, Product],
  synchronize: true,
};
```

```ts
export CrmConnection: ConnectionOptions = {
  name: 'crm',
  type: 'mysql',
  database: 'users',
  entities: [Customer, Address],
  synchronize: true,
};
```

This is more of a TypeORM documentation thing.

packages/typeorm/README.md
hacksparrow: Note to self: Create follow up task.

packages/typeorm/README.md
hacksparrow: > I think your list of limitations is missing the following items:

Updated.

packages/typeorm/src/decorators.ts
```diff
@@ -62,6 +62,6 @@ export namespace typeorm {
  * @param connectionName - Optional connection name
  */
 async function getConnection(ctx: Context, connectionName?: string) {
-  const connections = await ctx.get(TypeOrmBindings.CONNECTIONS);
-  return connections.manager.get(connectionName);
+  const manager = await ctx.get(TypeOrmBindings.MANAGER);
```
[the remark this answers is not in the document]
hacksparrow: Note to self: Create follow up task.

docs/site/Using-TypeORM-with-LoopBack.md
```diff
@@ -0,0 +1,132 @@
+---
+lang: en
+title: 'How to use TypeORM with LoopBack'
+keywords: LoopBack 4.0, LoopBack 4, TypeORM
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Using-TypeORM-with-LoopBack.html
```
raymondfeng, line 6: Let's use `Using-typeorm-with-loopback.html` to be consistent with other links.

docs/site/Using-TypeORM-with-LoopBack.md
```diff
@@ -0,0 +1,132 @@
...
+---
+
+## Overview
+
+If you would like to use TypeORM with LoopBack, it is possible with the use of
```
raymondfeng, line 11: Let's introduce TypeORM and have a link here.

docs/site/Using-TypeORM-with-LoopBack.md
```diff
@@ -0,0 +1,132 @@
...
+
+  @Column()
+  title: string;
+
+  @Column()
+  isPublished: boolean;
+}
+```
+
+### Creating Connections
```
raymondfeng, line 51: Maybe it's better to have `Creating Connections` come before `Creating Entities`?

docs/site/Using-TypeORM-with-LoopBack.md
```diff
@@ -0,0 +1,132 @@
...
+
+```sh
+npm install sqlite3
+```
+
+Refer to the
+[TypeORM documentation](https://github.com/typeorm/typeorm#installation) for the
+supported databases and the underlying drivers.
+
+### Creating controllers
```
raymondfeng, line 84: `Creating Controllers`

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -181,6 +181,13 @@ children:
   output: 'web, pdf'
   children:
 
+  - title: 'Using TypeORM'
+    output: 'web, pdf'
+    children:
+    - title: 'How to use TypeORM with LoopBack'
+      url: Using-TypeORM-with-LoopBack.html
+      output: 'web, pdf'
+
```
raymondfeng, line 190: Can we move it to be after `Accessing Databases`?

packages/typeorm/README.md
```diff
@@ -0,0 +1,152 @@
+# @loopback/typeorm
+
+This module enables TypeORM support in LoopBack. For pending features, refer to
+the [Limitations](#limitations) section below.
+
+## Overview
+
+**NOTE**: This module is experimental and evolving.
```
raymondfeng, line 8: Let's use the same disclaimer as https://github.com/strongloop/loopback-next/tree/master/extensions/cron#stability-%EF%B8%8Fexperimental%EF%B8%8F

packages/typeorm/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@loopback/typeorm",
+  "version": "0.0.1",
+  "description": "Adds support for TypeORM in LoopBack",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "engines": {
+    "node": ">=10"
```
raymondfeng, line 8: >=10.16.

packages/typeorm/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "devDependencies": {
+    "@loopback/boot": "^2.3.5",
+    "@loopback/build": "^5.4.3",
+    "@loopback/eslint-config": "^8.0.1",
+    "@loopback/repository": "^2.9.0",
+    "@loopback/rest": "^5.2.0",
+    "@loopback/service-proxy": "2.3.4",
+    "@loopback/testlab": "^3.1.7",
+    "@types/json-schema": "^7.0.5",
```
raymondfeng, line 32: Let's use the following commands to update its deps on other LB4 packages:

```sh
cd packages/typeorm
lb4 update --yes --force --skip-install
cd ../..
node bin/rebuild-package-locks.js @loopback/typeorm
```

packages/typeorm/package.json
```diff
@@ -0,0 +1,53 @@
+{
```
raymondfeng, line 1: Maybe we should move it to `extensions/typeorm`?

packages/typeorm/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "verify": "npm pack && tar xf loopback-typeorm*.tgz && tree package && npm run clean"
+  },
+  "author": "IBM Corp.",
+  "copyright.owner": "IBM Corp.",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@loopback/boot": "^2.3.5",
```
raymondfeng, line 25: This should be a regular dependency.

packages/typeorm/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@loopback/rest": "^5.2.0",
+    "@loopback/service-proxy": "2.3.4",
+    "@loopback/testlab": "^3.1.7",
+    "@types/json-schema": "^7.0.5",
+    "@types/node": "^10.17.26",
+    "sqlite3": "^4.2.0"
+  },
+  "dependencies": {
+    "@loopback/core": "^2.4.1",
+    "@loopback/openapi-v3": "3.4.4",
```
raymondfeng, line 38: Do we need it as a regular dependency?

packages/typeorm/src/typeorm.booter.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+  }
+}
+
+/**
+ * Default ArtifactOptions for TypeORM connection.
+ */
+export const ConnectionDefaults: ArtifactOptions = {
+  dirs: ['connections'],
```
raymondfeng, line 70: Let's prefix the directory, such as `typeorm-connections`.

packages/typeorm/package.json
hacksparrow: Yes, in https://github.com/strongloop/loopback-next/blob/typeorm-yaapa/packages/typeorm/src/typeorm.utils.ts.

packages/typeorm/src/__tests__/fixtures/connections/sqlite.connection.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import path from 'path';
+import {ConnectionOptions} from '../../../';
+import {Book} from '../entities/book.entity';
+
+export const SqliteConnection: ConnectionOptions = {
+  name: 'my-db',
+  type: 'sqlite',
+  database: path.join(__dirname, './mydb.sql'),
+  entities: [Book],
```
raymondfeng, line 14: I think it makes sense to have a booter for TypeORM entities too.

hacksparrow: That's the TypeORM convention, I think we should leave it alone. We just provide the glue, everything else - Entities, Connections, Respository methods - remain 100% pure TypeORM.

extensions/typeorm/README.md
```diff
@@ -0,0 +1,162 @@
...
+2. [Full JSON/OpenAPI schema for entities](https://github.com/strongloop/loopback-next/issues/5894),
+   including variants like with/without id, with/without relations, partial,
+   etc.
+3. [Support for LoopBack-style filters](https://github.com/strongloop/loopback-next/issues/5895)
+4. [JSON/OpenAPI schema to describe the supported filter format](https://github.com/strongloop/loopback-next/issues/5896)
+5. [Custom repository classes](https://github.com/strongloop/loopback-next/issues/5897)
+   (e.g. to implement bookRepo.findByTitle(title)).
+6. [Database migration](https://github.com/strongloop/loopback-next/issues/5898)
+
+We will be implementing them progressively.
```
dhmlau, line 144: We might want to make it as "Community contribution is welcome".

docs/site/Using-TypeORM-with-LoopBack.md
```diff
@@ -0,0 +1,134 @@
+---
+lang: en
+title: 'How to use TypeORM with LoopBack'
+keywords: LoopBack 4.0, LoopBack 4, TypeORM
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Using-typeorm-with-loopback.html
+---
+
+## Overview
```
dhmlau, line 9: Perhaps we can use the README.md content directly instead of duplicating the content? 
See the [JWT authentication extension docs](https://github.com/strongloop/loopback-next/blob/master/docs/site/JWT-authentication-extension.md ) as an example.

docs/site/sidebars/lb4_sidebar.yml
```diff
@@ -242,6 +242,13 @@ children:
       url: Database-migrations.html
       output: 'web, pdf'
 
+  - title: 'Using TypeORM'
```
dhmlau, line 245: any reason we need a parent item instead of having "How to use TypeORM with LoopBack" as the top level?

hacksparrow: There will be more articles under it later. Eg: "How to create custom repositories", etc.

extensions/typeorm/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "files": [
+    "README.md",
+    "dist",
+    "src",
+    "!*/__tests__"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/strongloop/loopback-next.git",
+    "directory": "packages/typeorm"
```
raymondfeng, line 50: extensions/typeorm

extensions/typeorm/src/__tests__/fixtures/application.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright IBM Corp. 2019. All Rights Reserved.
+// Node module: @loopback/boot
```
raymondfeng, line 2: Please fix license header

raymondfeng: You can use `lb4 copyright` command under `extensions/typeorm`.

extensions/typeorm/src/__tests__/fixtures/controllers/book.controller.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/rest
```
raymondfeng, line 2: Ditto.

extensions/typeorm/src/__tests__/fixtures/entities/book.entity.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
raymondfeng, line 1: `entities` -> `typeorm-entities`?

extensions/typeorm/src/typeorm.booter.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: @loopback/graphql
```
raymondfeng, line 2: Ditto.

extensions/typeorm/src/typeorm.booter.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
```
raymondfeng, line 1: typeorm-connection.booter.ts?

PR #5808, opened 2020-06-23
feat: coerce query object with schema

jannyHou:
Connect to #4992 

Filter with non-string properties like `?filter[fields][name]=false` are not coerced, so `false` turns to be `'false'`. This PR applies coercion for parameter query.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
+++ b/packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
@@ -27,6 +27,21 @@ describe('Coercion', () => {
     if (spy) spy.restore();
   });
 
+  const filterSchema = {
+    type: 'object',
+    title: 'filter',
+    properties: {
+      where: {
+        type: 'object',
+        properties: {
+          id: {type: 'number'},
+          name: {type: 'string'},
+          active: {type: 'boolean'},
+        },
+      },
+    },
+  };
+
   class MyController {
     @get('/create-number-from-path/{num}')
     createNumberFromPath(@param.path.number('num') num: number) {
@@ -49,7 +64,14 @@ describe('Coercion', () => {
     }
 
     @get('/object-from-query')
-    getObjectFromQuery(@param.query.object('filter') filter: object) {
+    getObjectFromQuery(
+      @param.query.object('filter', filterSchema) filter: object,
+    ) {
+      return filter;
+    }
+
+    @get('/random-object-from-query')
+    getRandomObjectFromQuery(@param.query.object('filter') filter: object) {
       return filter;
     }
   }
@@ -84,6 +106,8 @@ describe('Coercion', () => {
   });
 
   it('coerces parameter in query from nested keys to object', async () => {
+    // Notice that numeric and boolean values are coerced to their own types
+    // because the schema is provided.
     spy = sinon.spy(MyController.prototype, 'getObjectFromQuery');
     await client
       .get('/object-from-query')
@@ -94,9 +118,28 @@ describe('Coercion', () => {
       })
       .expect(200);
     sinon.assert.calledWithExactly(spy, {
-      // Notice that numeric and boolean values are converted to strings.
-      // This is because all values are encoded as strings on URL queries
-      // and we did not specify any schema in @param.query.object() decorator.
+      where: {
+        id: 1,
+        name: 'Pen',
+        active: true,
+      },
+    });
+  });
+
+  it('coerces parameter in query from nested keys to object - no schema', async () => {
+    // Notice that numeric and boolean values are converted to strings.
+    // This is because all values are encoded as strings on URL queries
+    // and we did not specify any schema in @param.query.object() decorator.
+    spy = sinon.spy(MyController.prototype, 'getRandomObjectFromQuery');
+    await client
+      .get('/random-object-from-query')
+      .query({
+        'filter[where][id]': 1,
+        'filter[where][name]': 'Pen',
+        'filter[where][active]': true,
+      })
+      .expect(200);
+    sinon.assert.calledWithExactly(spy, {
       where: {
         id: '1',
         name: 'Pen',
--- a/packages/rest/src/coercion/coerce-parameter.ts
+++ b/packages/rest/src/coercion/coerce-parameter.ts
@@ -3,9 +3,19 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {isReferenceObject, ParameterObject} from '@loopback/openapi-v3';
+import {
+  isReferenceObject,
+  ParameterObject,
+  ReferenceObject,
+  SchemaObject,
+} from '@loopback/openapi-v3';
 import debugModule from 'debug';
-import {RestHttpErrors} from '../';
+import {
+  RequestBodyValidationOptions,
+  RestHttpErrors,
+  validateValueAgainstSchema,
+  ValueValidationOptions,
+} from '../';
 import {parseJson} from '../parse-json';
 import {
   DateCoercionOptions,
@@ -27,18 +37,14 @@ const debug = debugModule('loopback:rest:coercion');
  *
  * @param data - The raw data get from http request
  * @param schema - The parameter's schema defined in OpenAPI specification
+ * @param options - The ajv validation options
  */
-export function coerceParameter(
+export async function coerceParameter(
   data: string | undefined | object,
   spec: ParameterObject,
+  options?: ValueValidationOptions,
 ) {
-  let schema = spec.schema;
-
-  // If a query parameter is a url encoded Json object, the schema is defined under content['application/json']
-  if (!schema && spec.in === 'query' && spec.content) {
-    const jsonSpec = spec.content['application/json'];
-    schema = jsonSpec.schema;
-  }
+  const schema = extractSchemaFromSpec(spec);
 
   if (!schema || isReferenceObject(schema)) {
     debug(
@@ -72,7 +78,7 @@ export function coerceParameter(
     case 'boolean':
       return coerceBoolean(data, spec);
     case 'object':
-      return coerceObject(data, spec);
+      return coerceObject(data, spec, options);
     case 'string':
     case 'password':
       return coerceString(data, spec);
@@ -158,19 +164,53 @@ function coerceBoolean(data: string | object, spec: ParameterObject) {
   throw RestHttpErrors.invalidData(data, spec.name);
 }
 
-function coerceObject(input: string | object, spec: ParameterObject) {
+async function coerceObject(
+  input: string | object,
+  spec: ParameterObject,
+  options?: RequestBodyValidationOptions,
+) {
   const data = parseJsonIfNeeded(input, spec);
 
-  if (data === undefined) {
+  if (data == null) {
     // Skip any further checks and coercions, nothing we can do with `undefined`
-    return undefined;
+    return data;
   }
 
   if (typeof data !== 'object' || Array.isArray(data))
     throw RestHttpErrors.invalidData(input, spec.name);
 
-  // TODO(bajtos) apply coercion based on properties defined by spec.schema
+  const schema = extractSchemaFromSpec(spec);
+  if (schema) {
+    // Apply coercion based on properties defined by spec.schema
+    await validateValueAgainstSchema(
+      data,
+      schema,
+      {},
+      {...options, coerceTypes: true, source: 'parameter'},
+    );
+  }
+
   return data;
+}
+
+/**
+ * Extract the schema from an OpenAPI parameter specification. If the root level
+ * one not found, search from media type 'application/json'.
+ *
+ * @param spec The parameter specification
+ */
+function extractSchemaFromSpec(
+  spec: ParameterObject,
+): SchemaObject | ReferenceObject | undefined {
+  let schema = spec.schema;
+
+  // If a query parameter is a url encoded Json object,
+  // the schema is defined under content['application/json']
+  if (!schema && spec.in === 'query') {
+    schema = spec.content?.['application/json']?.schema;
+  }
+
+  return schema;
 }
 
 function parseJsonIfNeeded(
--- a/packages/rest/src/parser.ts
+++ b/packages/rest/src/parser.ts
@@ -83,7 +83,7 @@ async function buildOperationArguments(
     }
     const spec = paramSpec as ParameterObject;
     const rawValue = getParamFromRequest(spec, request, pathParams);
-    const coercedValue = coerceParameter(rawValue, spec);
+    const coercedValue = await coerceParameter(rawValue, spec);
     paramArgs.push(coercedValue);
   }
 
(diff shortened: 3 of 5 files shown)


packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
```diff
@@ -98,9 +115,9 @@ describe('Coercion', () => {
       // This is because all values are encoded as strings on URL queries
       // and we did not specify any schema in @param.query.object() decorator.
       where: {
-        id: '1',
```
agnes512, line 101: Do we want to add `fields` filter to this test case? You might want to remove the comment above as it is fixed.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -169,10 +169,37 @@ function coerceObject(input: string | object, spec: ParameterObject) {
   if (typeof data !== 'object' || Array.isArray(data))
     throw RestHttpErrors.invalidData(input, spec.name);
 
-  // TODO(bajtos) apply coercion based on properties defined by spec.schema
+  const schema = extractSchemaFromSpec(spec);
+  if (schema) {
+    // apply coercion based on properties defined by spec.schema
+    const ajv = new Ajv({coerceTypes: true});
```
raymondfeng, line 175: It's a bit concerning that we hard-code the instantiation of `Ajv` here. We typically want to inject the ajv options.

jannyHou: @raymondfeng The ajv here is not used for validation, but for coercion purpose, I know this sounds a bit weird, but I don't find a better&simpler way to coerce based on schema.

See ajv doc in https://github.com/ajv-validator/ajv#coercing-data-types

If you search `options` in file [coerce parameter](https://github.com/strongloop/loopback-next/blob/4d22c725d2fe9e35663f0f50296602a7b6fa3d3f/packages/rest/src/coercion/coerce-parameter.ts), there is no options available for coercing object.

packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
jannyHou: @agnes512 I started adding feature by manually verify the field filter in todo app, so the fields filter does work.

`where: {active: true}`
and 
`fields: {name: true}`
doesn't make any difference for testing purpose, so I would say the current test case is good enough to cover our case(field filter) :p

agnes512: Oh I mentioned it because I remember when I was testing it, it works for `where` for some reasons 😂

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -18,6 +23,7 @@ import {
   matchDateFormat,
 } from './utils';
 import {Validator} from './validator';
+const Ajv = require('ajv');
```
raymondfeng, line 26: Please use `import` instead of `require`.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -169,10 +169,37 @@ function coerceObject(input: string | object, spec: ParameterObject) {
   if (typeof data !== 'object' || Array.isArray(data))
     throw RestHttpErrors.invalidData(input, spec.name);
 
-  // TODO(bajtos) apply coercion based on properties defined by spec.schema
+  const schema = extractSchemaFromSpec(spec);
+  if (schema) {
+    // apply coercion based on properties defined by spec.schema
+    const ajv = new Ajv({coerceTypes: true});
+    const validate = ajv.compile(schema);
```
raymondfeng, line 176: We should not create the `Ajv` instance per call. It's fairly expensive.

packages/rest/src/coercion/coerce-parameter.ts
raymondfeng: We should add an optional `options` parameter for `coerceParameter`. Its main usage is in `parse.ts` which has access to the configuration.

packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
jannyHou: @agnes512 I removed my change and tested with todo app, filter `?[where][isComplete]=false` does return correct result for some reason, weird. 

But the filter it self is not coerced, see my screenshot in:
Screen Shot 2020-06-23 at 4 13 56 PM

isComplete is `"false"` not `false`.

I need more time to investigate the reason, but the updated test case here should be correct. For the parameter coerce function, it wouldn't make sense if

`where: {active: true}` get coerced while `fields: {name: true}` not.

packages/rest/src/coercion/coerce-parameter.ts
jannyHou: True, I refactored to use one global instance. Or should I use a singleton scope provider instead?

packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
```diff
@@ -98,9 +115,9 @@ describe('Coercion', () => {
       // This is because all values are encoded as strings on URL queries
       // and we did not specify any schema in @param.query.object() decorator.
```
agnes512, line 116: I think the comment can be removed.

jannyHou: ah, true

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -169,10 +175,39 @@ function coerceObject(input: string | object, spec: ParameterObject) {
   if (typeof data !== 'object' || Array.isArray(data))
     throw RestHttpErrors.invalidData(input, spec.name);
 
-  // TODO(bajtos) apply coercion based on properties defined by spec.schema
+  const schema = extractSchemaFromSpec(spec);
+  if (schema) {
+    // Apply coercion based on properties defined by spec.schema
+    const validate = AJV.compile(schema);
```
raymondfeng, line 181: See more details at https://github.com/strongloop/loopback-next/pull/1757. The `compile` call is probably the most expensive one.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -21,6 +27,12 @@ import {Validator} from './validator';
 const isRFC3339 = require('validator/lib/isRFC3339');
 const debug = debugModule('loopback:rest:coercion');
 
+/**
+ * The AJV instance is created for coercing object parameters
+ * with provided schema. Not for validation.
+ */
+const AJV = new AjvCtor({coerceTypes: true});
```
raymondfeng, line 34: We have an `AjvFactory` being injected and it's available via `options`.

packages/rest/src/coercion/coerce-parameter.ts
raymondfeng: We probably have to maintain a weak map (schema -> ajv).

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -169,10 +177,45 @@ function coerceObject(input: string | object, spec: ParameterObject) {
   if (typeof data !== 'object' || Array.isArray(data))
     throw RestHttpErrors.invalidData(input, spec.name);
 
-  // TODO(bajtos) apply coercion based on properties defined by spec.schema
+  const schema = extractSchemaFromSpec(spec);
+  if (schema) {
+    // Apply coercion based on properties defined by spec.schema
+    await validateValueAgainstSchema(
```
bajtos, line 183: `validateValueAgainstSchema` throws an error when AJV validation fails. Is it ok for `coerceObject` to throw validation errors? 

`validateValueAgainstSchema` prints debug logs referring to `request body`. If we are going to use `validateValueAgainstSchema` for non-body parameters (which consider as a good idea), then the debug logs need to be updated accordingly.

We should also take a look at the validation errors produced for non-body parameters and make sure they provide enough information about which parameter failed the validation.

Another aspect to consider: should we report the first validation error only (I think that's the current design)? IMO, it's easier for users when all errors are reported at once. A possible solution for this is to build a "composite" object from all parameters (e.g. `{filter: <value from query string>, authorization: <value from headers>}`), a "composite" schema from Operation Object, and then call AJV only once. I think this way the error messages will include the parameter name ("filter", "authorization") as part of the description (path to the property with a wrong value).

packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
```diff
@@ -94,13 +111,10 @@ describe('Coercion', () => {
       })
       .expect(200);
     sinon.assert.calledWithExactly(spy, {
-      // Notice that numeric and boolean values are converted to strings.
-      // This is because all values are encoded as strings on URL queries
-      // and we did not specify any schema in @param.query.object() decorator.
```
bajtos, line 99: Please preserve the "notice" comment and update it to describe (point out) the new behavior.

bajtos: On the second though, I think we should preserve the current test as-is to describe how are schema-less values handled, and add a new test to verify how does schema-based coercion work.

packages/rest/src/parser.ts
```diff
@@ -83,7 +83,7 @@ async function buildOperationArguments(
     }
     const spec = paramSpec as ParameterObject;
     const rawValue = getParamFromRequest(spec, request, pathParams);
-    const coercedValue = coerceParameter(rawValue, spec);
+    const coercedValue = await coerceParameter(rawValue, spec);
```
bajtos, line 86: Building on top of my previous comments, perhaps this is the place where to call validation?

```suggestion
    const coercedValue = coerceParameter(rawValue, spec);
    const finalValue = await validateParameterValue(rawValue, spec);
    paramArgs.push(finalValue);
```

(Of course, this will not help with collecting all validation error together and other aspects mentioned in my other comments, but it may be good enough as a short-term fix.)

bajtos: We have also access to `globalSchemas` and `options` here, which is needed to resolve `{$ref}` schemas and use the same AJV factory as the request body validator does.

```suggestion
    const coercedValue = await coerceParameter(rawValue, spec);
    const coercedValue = coerceParameter(rawValue, spec);
    const finalValue = await validateParameterValue(rawValue, spec, globalSchemas, options);
    paramArgs.push(finalValue);
```

packages/rest/src/__tests__/acceptance/coercion/coercion.acceptance.ts
jannyHou: > I think we should preserve the current test as-is to describe how are schema-less values handled, and add a new test to verify how does schema-based coercion work.

Good point, I will edit the test.

packages/rest/src/parser.ts
jannyHou: Thanks @bajtos I tried calling `validateValueAgainstSchema` in the new place, and got 20+ test failures. This story is to coerce the **object** based on schema, but `coerceParameter` handles all types, and same to `validateValueAgainstSchema` if I call it after `coerceParameter`. 

I am a bit reluctant to investigate and verify why other types fail, and considering `schema` is usually used for object, can we deal with other types next time?

packages/rest/src/coercion/coerce-parameter.ts
raymondfeng: Having an aggregated schema that stitch all incoming params (including the body) is an interesting idea that we should explore. But let's leave it out of scope for this PR.

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -115,7 +115,7 @@ function getKeyForOptions(options: RequestBodyValidationOptions) {
  * @param globalSchemas - Schema references.
  * @param options - Request body validation options.
  */
-async function validateValueAgainstSchema(
+export async function validateValueAgainstSchema(
```
raymondfeng, line 118: Let's change the reference to `body` with a generic one, such as `value`.

jannyHou: done.

packages/rest/src/coercion/coerce-parameter.ts
jannyHou: > Having an aggregated schema that stitches all incoming params (including the body) is an interesting idea that we should explore.

@bajtos @raymondfeng Thanks for suggestion, I also tend to create a separate story for it. In this PR, I turned `validateValueAgainstSchema` a generic function that validates any value:

- rename 1st parameter: `data` --> `value`
- `ValueValidationOptions` extends `RequestBodyValidationOptions` with one more property `position`
- throws invalid request body error when `position` is 'body', throws invalid data error otherwise.

The new story can create a new rest http error for other invalid data. 
WDYT?

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -158,21 +163,55 @@ function coerceBoolean(data: string | object, spec: ParameterObject) {
...
-function coerceObject(input: string | object, spec: ParameterObject) {
+async function coerceObject(
+  input: string | object,
+  spec: ParameterObject,
+  options?: RequestBodyValidationOptions,
+) {
   const data = parseJsonIfNeeded(input, spec);
 
-  if (data === undefined) {
+  if (data === undefined || data === null) {
```
raymondfeng, line 173: Nitpick: it can be simplified as `if (data == null)`

packages/rest/src/types.ts
```diff
@@ -108,6 +108,17 @@ export type AjvKeyword = KeywordDefinition & {name: string};
...
 
+/**
+ * Options for any value validation using AJV
+ */
+export interface ValueValidationOptions extends RequestBodyValidationOptions {
+  /**
+   * Where the data comes from. It can be 'body', 'path', 'header',
+   * 'query', 'cookie', etc...
+   */
+  position?: string;
```
raymondfeng, line 119: Maybe `source`?

packages/rest/src/validation/request-body.validator.ts
```diff
@@ -145,10 +152,10 @@ async function validateValueAgainstSchema(
 
   let validationErrors: ajv.ErrorObject[] = [];
   try {
-    const validationResult = await validate(body);
-    // When body is optional & values is empty / null, ajv returns null
+    const validationResult = await validate(value);
+    // When value is optional & values is empty / null, ajv returns null
     if (validationResult || validationResult === null) {
-      debug('Request body passed AJV validation.');
+      debug('Value passed AJV validation.');
```
raymondfeng, line 158: We can use `options.position` for debugging too.

jannyHou: Good point. Added.

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -28,17 +38,12 @@ const debug = debugModule('loopback:rest:coercion');
  * @param data - The raw data get from http request
  * @param schema - The parameter's schema defined in OpenAPI specification
```
agnes512, line 39: Let's update it.

packages/rest/src/coercion/coerce-parameter.ts
jannyHou: creating the follow-up story, post here soon.

PR #5813, opened 2020-06-23
Rework snapshot matcher to support parallel mocha

bajtos:
Extracted from #5011 to land the changes in isolation from the actual enablement of parallel testing.

The first commit is cherry-picked 20da28ef99a5b895c5315365397f4620553238d2, then I added more commits to clean up the code more to my liking.

My intention is to further clean up the commit history once we agree on the implementation direction & details.

To verify parallel tests, one can run the test suite via `npm test -- --parallel`. Some of the tests are failing in parallel mode on a timeout, I think that's something to investigate as part of enabling parallel mode (outside of this PR).

## Checklist


.mocharc.js
```diff
@@ -0,0 +1,13 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+const {mergeMochaConfigs} = require('./packages/build');
+const defaultConfig = require('./packages/build/config/.mocharc.json');
+
+module.exports = mergeMochaConfigs(
+  defaultConfig,
+  require('./packages/cli/.mocharc.js'),
+  // TODO: add Mocha configs for other packages
+);
```
bajtos, line 13: I am not sure about this design - at one hand it feels a bit like overkill for our current needs, on the other hand I dislike the version proposed in https://github.com/strongloop/loopback-next/pull/5011 where we were using the CLI mocha options directly. Maybe instead of a generic merge algorithm it would be enough to merge `timeout` and `require` entries only?

Also in #5011, `process.env.MOCHA_JOBS` was handled by CLI's mocha file - I don't think that's correct. I would expect that flag to be handled either at monorepo level or inside `packages/build/config/.mocharc.json`. The latter would be a bit more tricky because we don't want to enable `parallel` mode for everybody. Anyhow, we can discuss that elsewhere, it's out of scope of this PR.

raymondfeng:
@bajtos Please fix [code formatting](https://travis-ci.com/github/strongloop/loopback-next/jobs/352721043).

packages/build/src/merge-mocha-configs.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = mergeMochaConfigs;
+
+/**
+ * Merge multiple Mocha configuration files into a single one.
+ *
+ * @param {MochaConfig[]} configs A list of Mocha configuration objects
+ * as provided by `.mocharc.js` files.
+ */
+function mergeMochaConfigs(...configs) {
```
raymondfeng, line 18: Can we use https://lodash.com/docs/4.17.15#mergeWith instead?

bajtos: IDK, do we want to introduce dependency on `lodash`? Personally, I try to keep our dependencies small.

raymondfeng:
`lodash` is a much smaller dependency than other ones we already have for `@loopback/build`, such as typescript, nyc, mocha, prettier.

bajtos:
> `lodash` is a much smaller dependency than other ones we already have for `@loopback/build`, such as typescript, nyc, mocha, prettier.

Fair enough.

> Can we use https://lodash.com/docs/4.17.15#mergeWith instead?

I am not comfortable with using `mergeWith`'s deep merge algorithm, I am concerned that non-scalar config options (like `require`) may require specific handling and the default deep-merge implementation may create unwanted outcomes.

Let's use `assignWith` for the first version, learn more about real-world usage and then improve the merging algorithm  as needed.

bajtos:
I rebased the patch on top of the latest master, addressed review comments and made few more improvements/cleanups:

- fc0efd4 fix linter issues 
- 525a8c9 simplify mergeMochaConfigs using lodash 
- 3216985 clean up packages/cli/.mocharc.js 
- 9968063 improve comment in root .mocharc.js 
- f7d26cfff cleanup snapshot matcher 

    1. Rename "state" to "snapshotMatcher" ("matcher"). I find "state" as too ambiguous, let's think about the new design as having one SnapshotMatcher instance for each snapshot directory. It is possible to further refactor the code from current functional-programming style into OOP by introducing `SnapshotMatcher` class, but I feel such change is not needed now.
    2. Move definition of root hook functions out of the hooks object, to make it easier to build a high-level understanding of the hooks without having to skip through implementation details.
    3. Move "currentTest" from individual states (matcher) into a single global variable, there is no need to duplicate it in every snapshot matcher instance.

@raymondfeng I consider the PR as ready for final review, PTAL.

bajtos:
(I'll clean up the commit history before merging.)

PR #5858, opened 2020-06-30
fix(repository): hasManyThrough can delete correct target n through based on filter

### LINKED ISSUE #5852: should be able to delete target instances based on filters in HasManyThrough relation
## Steps to reproduce

a `Doctor` has many `Patient`s through `Appointment`, the relation called `patients`.

```
Doctor :           [{id: 1, dec: 'doctor'}]
Appointment:       [{doctorId: 1, patientId: 1}, {doctorId: 1, patientId: 2}]
Patient:           [{id: 1, name: 'p1'}, {id: 2, name: 'p2'}]
``` 
## Current Behavior

```
// delete all patients that belongs to doctor id 1 where the patient name is p2
doctorRepo.patients(1).delete({name: 'p2'});
```

The target `p2` gets deleted, but patient `p1` is no longer traversable as the through model `Appointment{doctorId: 1, patientId: 1}` gets deleted too.

## Expected Behavior

Should delete the related through models instead of deleting **all** throughs that belong to the same source key.

## Acceptance Criteria
- [ ] redesign the `delete` method in hasManyThrough repository
- [ ] add tests to verify the new behaviour

agnes512:
fixes #5852 

## Checklist

### DIFF
--- a/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
+++ b/packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
@@ -66,7 +66,7 @@ export function hasManyThroughRelationAcceptance(
       existingCustomerId = (await givenPersistedCustomerInstance()).id;
     });
 
-    it('can create an instance of the related model alone with a through model', async () => {
+    it('creates an instance of the related model alone with a through model', async () => {
       const item = await customerRepo
         .cartItems(existingCustomerId)
         .create(
@@ -93,7 +93,7 @@ export function hasManyThroughRelationAcceptance(
       );
     });
 
-    it('can find instances of the related model', async () => {
+    it('finds instances of the related model', async () => {
       const item = await customerRepo
         .cartItems(existingCustomerId)
         .create(
@@ -108,7 +108,7 @@ export function hasManyThroughRelationAcceptance(
       expect(toJSON(result[0])).to.not.containEql(toJSON(notMyItem));
     });
 
-    it('can patch instances', async () => {
+    it('patches instances', async () => {
       const item1 = await customerRepo
         .cartItems(existingCustomerId)
         .create({description: 'group 1'});
@@ -129,7 +129,7 @@ export function hasManyThroughRelationAcceptance(
       );
     });
 
-    it('can patch an instance based on the filter', async () => {
+    it('patches an instance based on the filter', async () => {
       const item1 = await customerRepo
         .cartItems(existingCustomerId)
         .create({description: 'group 1'});
@@ -162,7 +162,7 @@ export function hasManyThroughRelationAcceptance(
       ).to.be.rejectedWith(/Property "id" cannot be changed!/);
     });
 
-    it('can delete many instances and their through models', async () => {
+    it('deletes many instances and their through models', async () => {
       await customerRepo
         .cartItems(existingCustomerId)
         .create({description: 'group 1'});
@@ -182,7 +182,7 @@ export function hasManyThroughRelationAcceptance(
       expect(cartItems).have.length(0);
     });
 
-    it('can delete corresponding through models when the target gets deleted', async () => {
+    it('deletes corresponding through models when the target gets deleted', async () => {
       const item = await customerRepo
         .cartItems(existingCustomerId)
         .create({description: 'group 1'});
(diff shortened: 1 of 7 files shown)


packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -191,15 +192,29 @@ export class DefaultHasManyThroughRepository<
...
-    );
-    await throughRepository.deleteAll(
-      constrainWhereOr({}, [sourceConstraint, throughFkConstraint]),
-    );
-
+    if (where) {
+      // only delete related through models
+      const targets = await targetRepository.find({where: where});
+      const targetIds = this.getTargetIds(targets);
+      if (targetIds.length !== 0) {
```
jannyHou, line 199: nitpick:
```suggestion
      if (targetIds.length > 0) {
```

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -205,7 +205,37 @@ export function hasManyThroughRelationAcceptance(
       expect(cartItems).have.length(0);
     });
 
-    //FIXME(Agnes): should be able to do deletion based on filters
+    it('can delete instances based on the filter', async () => {
```
hacksparrow, line 208: `can delete instances based on the filter` -> `deletes instances based on the filter`

Please use this format on all test descriptions.

packages/repository-tests/src/crud/relations/acceptance/has-many-through.relation.acceptance.ts
```diff
@@ -205,9 +205,39 @@ export function hasManyThroughRelationAcceptance(
...
+        .create({description: 'group 2'});
+
+      let links = await customerCartItemLinkRepo.find();
+      let cartItems = await cartItemRepo.find();
+      expect(links).have.length(2);
+      expect(cartItems).have.length(2);
+
+      await customerRepo
+        .cartItems(existingCustomerId)
+        .delete({description: 'group 3'});
```
bajtos, line 223: IIUC, here you are trying to delete a record that does not exist. That fact is not immediately clear from the code. Let's use a value that makes the intent more obvious.

```suggestion
        .delete({description: 'record-does-not-exist'});
```

packages/repository/src/__tests__/integration/repositories/relation.factory.integration.ts
```diff
@@ -343,6 +343,36 @@ describe('HasManyThrough relation', () => {
...
+    const item2 = await customerCartItemRepo.create({
+      description: 'customer 2',
+    });
+
+    let items = await cartItemRepo.find();
+    let links = await customerCartItemLinkRepo.find();
+    expect(items).have.length(2);
+    expect(links).have.length(2);
+
+    await customerCartItemRepo.delete({description: 'does not exist'});
```
bajtos, line 359: > `description: 'does not exist'` 

👍

packages/repository/src/relations/has-many/has-many-through.repository.ts
```diff
@@ -191,15 +192,29 @@ export class DefaultHasManyThroughRepository<
...
-    const throughFkConstraint = this.getThroughConstraintFromTarget(
-      targetFkValues,
-    );
-    await throughRepository.deleteAll(
-      constrainWhereOr({}, [sourceConstraint, throughFkConstraint]),
-    );
-
+    if (where) {
+      // only delete related through models
+      const targets = await targetRepository.find({where: where});
```
bajtos, line 197: Since we already have a list of `troughInstances`, is there a way how to limit this query to find only target models that are matching the user-provided `where` and are related to the source model(s)?

I am concerned about performance implication of the current solution, because it may fetch too many records. Consider the case Doctor-Appointment-Patient case, where we want to delete all appointment+patient records of the given doctor where the patient has `{died: true}`. Let's say our database is large (e.g. ~100k doctors). For each doctor, there will be only few patients that have passed away (e.g. 1-10). Ideally, we should fetch only those 1-10 patients, not ~100k-1000k passed-away patients of all doctors.

I am fine to leave this performance optimization out of scope of the initial implementation, as long as we have some documentation for this known performance issue (a code comment is good enough).

A nitpick on coding style:
```suggestion
      const targets = await targetRepository.find({where});
```

agnes512: Agreed 👍. I couldn't find a way to fix it without accessing to the database, but we can improve it by limiting how many records it fetches.

I think it needs more helpers to build such a where filter. I left a TODO comment.

PR #5884, opened 2020-07-06
chore: use defineProperty to name a generated class

raymondfeng:
Use Object.defineProperty to name generated classes

## Checklist

### DIFF
--- a/packages/express/src/middleware-interceptor.ts
+++ b/packages/express/src/middleware-interceptor.ts
@@ -285,30 +285,20 @@ export function defineInterceptorProvider<
   className = buildName(middlewareFactory, className);
   assert(className, 'className is missing and it cannot be inferred.');
 
-  const defineNamedClass = new Function(
-    'middlewareFactory',
-    'defaultMiddlewareConfig',
-    'MiddlewareInterceptorProvider',
-    'createInterceptor',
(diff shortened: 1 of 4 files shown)


bajtos:
Can you please verify that the `name` property is used in the following places?
- Error stack traces (`new Error().stack`)
- Debugger stack frames

In the past, a custom `name` property was not picked everywhere, that's why we are using `new Function()` approach in juggler and elsewhere.

bajtos:
Interesting. Based on [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name), your trick may be a valid solution that should work :+1:

> The name property is read-only and cannot be changed by the assigment operator.
> To change it, use `Object.defineProperty()`.

If we adopt this new approach, then it would be great to (eventually) update other places where we are generating dynamic classes at runtime, e.g. repository & controller class factories.

raymondfeng:
> If we adopt this new approach, then it would be great to (eventually) update other places where we are generating dynamic classes at runtime, e.g. repository & controller class factories.

I was trying but there are some tricky TS complaints (about subclassing classes with generic type parameters) that I don't know how to fix them.

Issue #5872, opened 2020-07-04
Unable to perform nested include in filter since @loopback-rest 5.2.0


hessman:
Hi loopback team ! 

The last publish seems to have introduced a regression.

## Steps to reproduce

- Upgrade to @loopback-rest 5.2.0 
- Use PostgreSQL datasource with the loopback connector
- Create models, repositories with relations.
- Create simple controllers to perform a find or findById
- Do a query with nested include :
```
{
  "include": [
    {
      "relation": "relation0",
      "scope": {
        "include": [
          {
            "relation": "relation0relation0",
            "scope": {
              "include": [
                {
                  "relation": "relation0relation0relation0"
                }
              ]
            }
          }
        ]
      }
    },
    {
      "relation": "relation1"
    },
    {
      "relation": "relation2"
    }
  ]
}
```

## Current Behavior

A Bad Request 400 error :

```
{
  "error": {
    "statusCode": 400,
    "name": "BadRequestError",
    "message": "Invalid value.",
    "details": [
      {
        "path": "/include/0/scope",
        "code": "additionalProperties",
        "message": "should NOT have additional properties",
        "info": {
          "additionalProperty": "include"
        }
      }
    ]
  }
}
```

We can not use "include" inside "scope".

## Expected Behavior

The queried model with all his queried relations (and relation's relations).

## Link to reproduction sandbox

in progress

dhmlau:
@agnes512, I believe you were looking into some issues on the filter. Could you please take a look as well? Thanks.

agnes512:
@dhmlau this is causing by the change we made in #5808 . cc @jannyHou

jannyHou:
I think the generated `TodoList.ScopeFilter` should also have `include` property. It passed before because we only validate request body but not object in query.
PR https://github.com/strongloop/loopback-next/pull/5808 introduced coercion with spec schema for query input, that's why it starts to fail. 
Submitting a fix for the scope filter.

jannyHou:
Submitted a PR to fix it: https://github.com/strongloop/loopback-next/pull/5908

PR #5908, opened 2020-07-09
fix: nested scope filter

jannyHou:
Fixes https://github.com/strongloop/loopback-next/issues/5872

Nested inclusion filter like 

```ts
{
  include: [{
    relation: 'todos',
    scope: {
      // after enabling coercion against spec schema in PR 
      // https://github.com/strongloop/loopback-next/pull/5808
      // it breaks cuz scope schema doesn't have property include.
      include: [{
        relation: 'someOtherRelation',
      }]
    }
  }]
}
```

Fails. Because the `scope` schema is generated from `EmptyModel`, which doesn't have a relation:

https://github.com/strongloop/loopback-next/blob/683e162de982acde34b45101383a5d69cf1c0df1/packages/repository-json-schema/src/filter-json-schema.ts#L38

and thus does not have `include` generated in its `properties`:

https://github.com/strongloop/loopback-next/blob/683e162de982acde34b45101383a5d69cf1c0df1/packages/repository-json-schema/src/filter-json-schema.ts#L118-L120

This PR adds an any inclusion schema for the empty model's scope filter schema.

## Checklist

### DIFF
--- a/docs/site/Include-filter.md
+++ b/docs/site/Include-filter.md
@@ -64,6 +64,38 @@ You can also use
 [stringified JSON format](Querying-data.md#using-stringified-json-in-rest-queries) in
 a REST query.
 
+#### Scope Filter
+
+Please note if the scope filter contains non-string data, they won't be coerced
+if you use the plain query. Only the stringified JSON format works as expected.
+
+For example, the following REST query which includes related `users` without
+their names:
+
+`/modelName?filter[include][0][relation]=users&filter[include][0][scope][fields][name]=false`
+
+won't hide the `name` field for `users`, because `false` is a boolean value. It
+won't be coerced and will present as a string when passed to the controller
+function.
+
+A solution is to use the stringified JSON query instead to include data with
+scope specified:
+
+```ts
+// Define the inclusion filter and get its encoded format
+const inclusionFilter = {
+  include: {
+    relation: 'users',
+    scope: {
+      fields: {name: false},
+    },
+  },
+};
+const encodedFilter = encodeURIComponent(JSON.stringify(filter));
+```
+
+and call `/modelName?filter=<encodedFilter>`
+
 ### Examples
 
 - Return all customers with their orders:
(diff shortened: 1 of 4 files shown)


agnes512:
Just wondering what was the reason it `excluded scope.include`?

jannyHou:
@agnes512 good question, if you look at the last screenshot, 

`if (hasRelations && !excluded.includes('include'))` decides not add the include property because `EmptyModel` DOES NOT have relations.

agnes512:
Will the scope be coerced? If so, can we have a test for it? Not sure if this is a valid case :p

jannyHou:
@agnes512 
> Will the scope be coerced? 

No it won't be coerced, it's an any object...

agnes512:
> No it won't be coerced, it's an any object...

Does that mean for filters like 
```
/orders?filter[include][0][relation]=shipments&filter[include][0][scope]filter[fields][name]=false
```
would not have the expected result? (`fields` inside of the scope will be a string `false`)

jannyHou:
@agnes512 Right the false will be a string. The scope filter schema is generated from `EmptyModel` which doesn't have the ability to generate fields schema according to `TodoList` model. I verified

> http://127.0.0.1:3000/todo-lists?filter[include][0][relation]=todos&filter[include][0][scope][fields][desc]=false

returns wrong result. But the corresponding API flavor call returns correctly.

hacksparrow:
> The scope filter schema is generated from EmptyModel which doesn't have the ability to generate fields schema according to TodoList model. I verified.

Let's do the verification with a test.

agnes512:
>  I verified nhttp://127.0.0.1:3000/todo-lists?filter[include][0][relation]=todos&filter[include][0][scope][fields][desc]=false
returns wrong 

It seems a limitation to me. Can we also document it?

jannyHou:
cc @bajtos @hacksparrow Thanks for review 👍 , added tests in https://github.com/strongloop/loopback-next/pull/5908/commits/fa097ea8810ff24074a332e11ba1744096b94524

cc @agnes512 
> Can we also document it?

Sure I added notes in the test case, I will mention it in doc as well, next commit coming soon.

PR #5914, opened 2020-07-10
docs: remove duplicate entry

jannyHou:
There are 2 entries for authentication and authorization as you can see from

Screen Shot 2020-07-10 at 11 05 24 AM

Two on the first level and two under "Integrate with Infrastructures", so I'd suggest only keep the first level ones.

Also, the current first level authentication doesn't point to the overview, but the component instead.

---

The new layout:

Screen Shot 2020-07-10 at 3 50 08 PM

and fold them:

Screen Shot 2020-07-10 at 3 52 01 PM

## Checklist


agnes512:
Remove https://github.com/strongloop/loopback-next/blob/master/docs/site/Integrate-with-infrastructures.md as well?

Issue #5692, opened 2020-06-09
Hide `@loopback/openapi-v3` from docs, examples and CLI templates


bajtos:
In a follow-up to #5659 and #5550, we should update our docs, examples and CLI templates to load openapi-v3 APIs from `@loopback/rest`, so that framework users do not need to learn about `@loopback/openapi-v3` package.

## Acceptance criteria

In the search commands listed below, `rg` refers to [RipGrep](https://github.com/BurntSushi/ripgrep).

- [x] Remove `@loopback/openapi-v3` from all `package.json` files **EXCEPT** `packages/rest/package.json`. Update `package-lock.json` files where necessary. Search command: 

    ```sh
    $ rg loopback/openapi-v3 -g package.json
    ```

  --> #5927, #5932

- [x] Remove `@loopback/openapi-v3` in CLI templates and update tests as necessary.
    - `packages/cli/generators/project/templates/package.json.ejs` 
    - `packages/cli/generators/project/templates/package.plain.json.ejs`.

  --> https://github.com/strongloop/loopback-next/pull/5926

- [x] Update code in examples, replace imports from `@loopback/openapi-v3` to import from `@loopback/rest` instead.    Search command: 

    ```sh
    $ rg loopback/openapi-v3 examples/ -g '!package.json'
    ```

  --> #5927, #5932

- [x] Update code snippets in our documentation, replace imports from `@loopback/openapi-v3` to import from `@loopback/rest` instead. Search command:

    ```sh
    rg loopback/openapi-v3 docs/site
    ```

  --> #5933

- [x] Final check: review remaining references to `@loopback/openapi-v3` and verify they are all desired. Search command:

    ```sh
    $ rg loopback/openapi-v3 -g '!CHANGELOG.md'
    ```

bajtos:
Closing as done 🎊

Issue #5336, opened 2020-05-05
Clean up extension template and docs


bajtos:
_This is a follow-up for #5337 and #5122_

The documentation for component authors (e.g. [Creating components](https://loopback.io/doc/en/lb4/Creating-components.html)) and the template used by `lb4 extension` are outdated. We learned a lot since those artifacts were written and introduced new conventions and features, e.g. `@config` decorator. It's time to modernize both the CLI template and the docs to the latest and greatest.

### CLI templates

Update the template used by [`lb4 extension`](https://github.com/strongloop/loopback-next/tree/master/packages/cli/generators/extension) to follow the current style as seen e.g. in [`extensions/metrics`](https://github.com/strongloop/loopback-next/blob/master/extensions/metrics)

- [x] inject the target app and component config to the component class
- [x] add a new file `src/keys.ts` exporting `{ComponentName}Keys.COMPONENT` binding key
- [x] decorate the component class with `@bind()`,
- [x] include generic Basic Use instructions in the scaffolded README file

_Nice to have:_

- [ ] use the component name in the file names to disambiguate them from similar files in other extensions, e.g. `src/{name}.keys.ts` and `src/{name}.component.ts`

### Docs

Revisit the documentation, make it consistent with the current style for writing extensions and the CLI template.

- [x] [Creating components](https://loopback.io/doc/en/lb4/Creating-components.html)
  - Rework "Configuring components" at the end to explain how to receive and provide component config via `@config` and
`this.configure()` API
  - Review the content - are the advices and code snippets still current and consistent with recommendations we are giving elsewhere? Are we missing any major information? (It's probably best to open follow-up stories to make non-trivial docs updates.)
- [x] [Extending LoopBack 4](https://loopback.io/doc/en/lb4/Extending-LoopBack-4.html),
  - "Learn from existing ones" - can we add more of our existing components to the list?
  - Review the content - are the advices and code snippets still current and consistent with recommendations we are giving elsewhere? Are we missing any major information? (It's probably best to open follow-up stories to make non-trivial docs updates.)
- [x] [Using components](https://loopback.io/doc/en/lb4/Using-components.html) 
  - Update the list of components to include all extensions developed in loopback-next monorepo (look for components both in [`extensions`](https://github.com/strongloop/loopback-next/tree/master/extensions) and [`packages`](https://github.com/strongloop/loopback-next/tree/master/packages))
- [x] [Components](https://loopback.io/doc/en/lb4/Components.html)

[...]

PR #5927, opened 2020-07-13
feat: remove openapi-v3 from apiconnect dependencies

bajtos:
Use `@loopback/rest` to access openapi-related types and helpers.

This pull request is a part of #5692.

~~BREAKING CHANGE: `@loopback/core` and `@loopback/rest` were moved to peer dependencies, to ensure the extension uses application's version of these modules. If this change breaks your application (which is unlikely), then simply add both `@loopback/core` and `@loopback/rest` to your project dependencies.~~

~~_I am observing timeouts when running the tests locally (even on master), I think they are not related to my changes and will rely on CI builds to confirm._~~

## Checklist

### DIFF
--- a/extensions/apiconnect/package.json
+++ b/extensions/apiconnect/package.json
@@ -40,13 +40,12 @@
   },
   "dependencies": {
     "@loopback/core": "^2.9.1",
-    "@loopback/openapi-v3": "^3.4.5",
+    "@loopback/rest": "^5.2.0",
     "tslib": "^2.0.0"
   },
   "devDependencies": {
     "@loopback/build": "^6.1.0",
     "@loopback/eslint-config": "^8.0.3",
-    "@loopback/rest": "^5.2.0",
     "@loopback/testlab": "^3.2.0",
     "@types/node": "^10.17.27"
   }
--- a/extensions/apiconnect/src/apiconnect.spec-enhancer.ts
+++ b/extensions/apiconnect/src/apiconnect.spec-enhancer.ts
@@ -10,7 +10,7 @@ import {
   CoreBindings,
   inject,
 } from '@loopback/core';
-import {asSpecEnhancer, OASEnhancer, OpenAPIObject} from '@loopback/openapi-v3';
+import {asSpecEnhancer, OASEnhancer, OpenAPIObject} from '@loopback/rest';
 
 /**
  * Configuration for IBM API Connect extensions to the OpenAPI spec
--- a/extensions/apiconnect/tsconfig.json
+++ b/extensions/apiconnect/tsconfig.json
@@ -14,9 +14,6 @@
       "path": "../../packages/core/tsconfig.json"
     },
     {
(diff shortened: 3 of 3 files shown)


extensions/apiconnect/package.json
```diff
@@ -39,12 +39,15 @@
...
   },
   "dependencies": {
-    "@loopback/core": "^2.9.1",
-    "@loopback/openapi-v3": "^3.4.5",
     "tslib": "^2.0.0"
   },
+  "peerDependencies": {
+    "@loopback/core": "^2.9.1",
+    "@loopback/rest": "^5.2.0"
+  },
```
raymondfeng, line 47: Should we wait for https://github.com/npm/rfcs/blob/latest/accepted/0025-install-peer-deps.md to be available in `npm@7`? 

At the moment, we have mixed usage of `peerDependencies` and `dependencies` for extension-like packages. It would be nice to make them consistent.

BTW, I also dislike the fact that we have to duplicate `peerDependencies` and `devDependencies`.

bajtos: > Should we wait for https://github.com/npm/rfcs/blob/latest/accepted/0025-install-peer-deps.md to be available in `npm@7`?

I think that's not needed. `lb4  app` always adds `@loopback/core` to project dependencies and `@loopback/rest` is usually added too (unless the user explicitly opts out).

BTW does it make any sense to use `@loopback/apiconnect` in a LB project (application) that does not have any controllers (no REST API)? Because if an app has REST API, then I think it must also have a dependency on `@loopback/rest`, right?

> At the moment, we have mixed usage of `peerDependencies` and `dependencies` for extension-like packages. It would be nice to make them consistent.

I agree! I thought we have a tracking issue for that, but could not find any. The closest one is https://github.com/strongloop/loopback-next/issues/5336. Should I open a new issue?
 
> BTW, I also dislike the fact that we have to duplicate entries in `peerDependencies` and `devDependencies`.

Yeah, I don't like that either. I guess it's the price we have to pay for using Node.js & npm.

bajtos:
@raymondfeng thank you for the comment above. How to move this pull request forward? Are you fine with the proposed approach or would you like me to rework the patch to avoid introduction of peer dependencies? Something else?

raymondfeng:
Here is my personal preference but I won't block you from merging this PR as is.

1. Simply remove `@loopback/openapi-v3` without changing to `peerDependencies` in this PR.
2. Having a follow-up PR to universally switch to `peerDependencies` for all extension-like packages.

bajtos:
> 1. Simply remove `@loopback/openapi-v3` without changing to `peerDependencies` in this PR.
> 2. Having a follow-up PR to universally switch to `peerDependencies` for all extension-like packages.

That's a good plan 👍 

I cannot simply remove `@loopback/openapi-v3` from dependencies, I need to move `@loopback/rest` from dev-dependencies to regular dependencies to avoid the situation when `import` statement is referencing a dependency that is not be available.

@raymondfeng LGTY now?

PR #5937, opened 2020-07-15
fix: enable custom AJV validation errors

hacksparrow:
The current documentation is missing a crucial piece of information on how to enable custom validation and error messages, added the details.

### DIFF
--- a/docs/site/Model.md
+++ b/docs/site/Model.md
@@ -918,10 +918,12 @@ If the `jsonSchema` field is missing, you will get an error saying
 
 > You must provide the "jsonSchema" field when define a nested array property'
 
-### Validation Rules
+### Custom Validation Rules and Error Messages
 
-You can also specify the validation rules in the field `jsonSchema`. For
-example:
+You can also specify the validation rules in the `jsonSchema` field of the
+property option and configure them with custom error messages.
+
+The validation rules and custom error messages are configured this way.
 
 ```ts
 @model()
@@ -930,15 +932,29 @@ class Product extends Entity {
     name: 'name',
     description: "The product's common name.",
     type: 'string',
-    // Specify the JSON validation rules here
+    // JSON validation rules
     jsonSchema: {
-      maxLength: 30,
       minLength: 10,
-      errorMessage:
-        'name must be at least 10 characters and maximum 30 characters',
+      maxLength: 30,
+      errorMessage: 'Name should be between 10 and 30 characters.',
     },
   })
   public name: string;
+}
+```
+
+In case you want to send an error message specific to the validation rule that
+did not pass, you can configure `errorMessage` this way.
+
+```ts
+jsonSchema: {
+  minLength: 10,
+  maxLength: 30,
+  errorMessage: {
+    // Corresponding error messages
+    minLength: 'Name should be at least 10 characters.',
+    maxLength: 'Name should not exceed 30 characters.',
+  }
 }
 ```
 
--- a/docs/site/Parsing-requests.md
+++ b/docs/site/Parsing-requests.md
@@ -208,28 +208,9 @@ A full list of validation keywords could be found in the
 
 ##### Custom Error Messages
 
-You can also specify custom error messages for the JSON schema validation rules
-in the model property decorator. The messages are added in field called
-`errorMessage` inside `jsonSchema` like:
-
-```ts
-@model()
-class Product extends Entity {
-  @property({
-    name: 'name',
-    description: "The product's common name.",
-    type: 'string',
-    // Specify the JSON validation rules here
-    jsonSchema: {
-      maxLength: 30,
-      minLength: 10,
-      errorMessage:
-        'name must be at least 10 characters and maximum 30 characters',
-    },
-  })
-  public name: string;
-}
-```
+You can also specify custom error messages for JSON schema validation rules, as
+explained in
+[Custom Validation Rules and Error Messages](Model.md#custom-validation-rules-and-error-messages).
 
 A full list of options & usage scenarios could be found in the
 [documentation of AJV errors](https://github.com/epoberezkin/ajv-errors).
--- a/examples/validation-app/README.md
+++ b/examples/validation-app/README.md
@@ -19,22 +19,26 @@ instance with the in-memory storage.
    @property({
     type: 'string',
     required: true,
-    // --- add jsonSchema -----
+    // Add jsonSchema
     jsonSchema: {
       maxLength: 10,
-      minLength: 1,
+      minLength: 5,
+      errorMessage: 'City name should be between 5 and 10 characters',
     },
-    // ------------------------
    })
    city: string;
    ```
 
 2. [ValidatePhoneNumInterceptor](src/interceptors/validate-phone-num.interceptor.ts):
-   interceptor that checks whether the area code of the phone number matches
-   with the city name
+   an interceptor for custom validation - it checks whether the area code of the
+   phone number matches with the city name.
 
 3. [CoffeeShopController](src/controllers/coffee-shop.controller.ts): controller
    where the `ValidatePhoneNumInterceptor` is applied.
+
+4. [MySequence](src/sequence.ts): a custom sequence for creating custom error
+   messages. This is not a requirement, just a demonstration of how to customize
+   error messages during runtime.
 
 ## Use
 
@@ -62,7 +66,7 @@ is happening. An example of valid request body for `CoffeeShop` is:
 ```
 
 However, the following examples of request body are not valid, and you'll be
-getting an error with status code 422.
+getting an error with status code 422 and their corresponding error messages.
 
 ### Example 1: CoffeeShop.city length exceeds limit of 10
 
--- a/examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
+++ b/examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
@@ -32,24 +32,71 @@ describe('validate properties', () => {
   });
 
   it('fails when creating a CoffeeShop with the city name exceeds length limit', async () => {
-    const invalidCityNameCS = validCoffeeShop;
+    const invalidCityNameCS = {...validCoffeeShop};
     invalidCityNameCS.city = 'Tooooooooooronto';
-    await client.post('/coffee-shops').send(invalidCityNameCS).expect(422);
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidCityNameCS)
+      .expect(422);
+    expect(response.body.error.details.length).to.equal(1);
+    expect(response.body.error.details[0].message).to.equal(
+      'City name should be between 5 and 10 characters',
+    );
   });
 
   it('fails when creating a CoffeeShop with an invalid phone number', async () => {
-    const invalidPhoneNumCS = validCoffeeShop;
+    const invalidPhoneNumCS = {...validCoffeeShop};
     invalidPhoneNumCS.phoneNum = '1111111111';
-    await client.post('/coffee-shops').send(invalidPhoneNumCS).expect(422);
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidPhoneNumCS)
+      .expect(422);
+    expect(response.body.error.details.length).to.equal(1);
+    expect(response.body.error.details[0].message).to.equal(
+      'Invalid phone number',
+    );
   });
 
   it('fails when creating a CoffeeShop with the capacity exceeds limit', async () => {
-    const invalidCapacityCS = validCoffeeShop;
+    const invalidCapacityCS = {...validCoffeeShop};
     invalidCapacityCS.capacity = 10000;
-    await client.post('/coffee-shops').send(invalidCapacityCS).expect(422);
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidCapacityCS)
+      .expect(422);
+    expect(response.body.error.details.length).to.equal(1);
+    expect(response.body.error.details[0].message).to.equal(
+      'Capacity cannot exceed 100',
+    );
   });
 
-  it('fails when creating a CoffeeShop with an invalid area code', async () => {
+  it('should report all validation errors', async () => {
+    const invalidCapacityTypeCS = {
+      city: '',
+      phoneNum: '',
+      capacity: 0,
+      rating: '',
+    };
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidCapacityTypeCS)
+      .expect(422);
+
+    expect(response.body.error.message).to.equal(
+      'The request body is invalid. See error object `details` property for more info.',
+    );
+    expect(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      response.body.error.details.map((d: any) => d.message),
+    ).to.deepEqual([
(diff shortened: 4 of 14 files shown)


docs/site/Model.md
```diff
@@ -918,10 +918,35 @@ If the `jsonSchema` field is missing, you will get an error saying
...
+  constructor(options: ApplicationConfig = {}) {
+    super(options);
+
+    const validationOptions: RequestBodyValidationOptions = {
+      $data: true,
+      ajvKeywords: true,
+      ajvErrors: true
+    };
+
+    this.bind(RestBindings.REQUEST_BODY_PARSER_OPTIONS).to({validation: validationOptions});
```
jannyHou, line 945: We initially set up the options in [here](https://github.com/strongloop/loopback-next/blob/a24ea4f7fb9a0333f16282a8586d3ab75a77bd6c/packages/rest/src/validation/ajv-factory.provider.ts#L54-L60), when the error message is a string, there is no need to configure the options...I am also confused why the options are needed as an object.

docs/site/Model.md
```diff
@@ -930,18 +955,25 @@ class Product extends Entity {
...
+        minLength: 'name should be at least 10 characters',
+        maxLength: 'name should not exceed 30 characters',
+      },
     },
   })
   public name: string;
 }
 ```
 
+Note: You can also specify a string instead of an object as a value for
```
jannyHou, line 973: nitpick: I'd suggest keep both examples to explain the note here
```ts
jsonSchema: {
      maxLength: 30,
      minLength: 10,
      errorMessage:
        'name must be at least 10 characters and maximum 30 characters'
}
```

docs/site/Model.md
hacksparrow: You will notice until https://github.com/hacksparrow/lb4.new/blob/master/src/application.ts#L27 is uncommented, custom errors don't work as expected.

So, maybe it is actually a bug than bad documentation?

docs/site/Model.md
```diff
@@ -930,18 +955,25 @@ class Product extends Entity {
...
+`errorMessage`. In this case the same string will be used as the error message
+for all the errors.
```
bajtos, line 975: Can you please add an example how to use a single `errorMessage` field? (See also @jannyHou comment above). IMO, the second example does not have to repeat the entire model/property definition, just show the value for `jsonSchema` field.

docs/site/Parsing-requests.md
```diff
@@ -208,9 +208,33 @@ A full list of validation keywords could be found in the
 
 ##### Custom Error Messages
 
-You can also specify custom error messages for the JSON schema validation rules
-in the model property decorator. The messages are added in field called
-`errorMessage` inside `jsonSchema` like:
+You can also specify the validation rules in the `jsonSchema` field of the
```
bajtos, line 211: +1 to update all relevant places.

It makes me wonder though, why are we duplicating this content? Can we change one of these two pages and replace "custom error messages" section with a link to the other page?

For example:

```diff
- ##### Custom Error Messages
- You can also specify custom error messages for the JSON schema validation rules
- (...)
+ You can also specify custom error messages for JSON schema validation rules, as explained in
+ [Custom Validation Rules and Error Messages](Model.md#custom-validation-rules-and-error-messages).
```

examples/validation-app/README.md
```diff
@@ -19,12 +19,10 @@ instance with the in-memory storage.
    @property({
     type: 'string',
     required: true,
-    // --- add jsonSchema -----
```
dhmlau, line 22: you don't want to keep the comment? :) 
I was thinking that it might be clearer to users what has been added apart from the model generation bits. 
Perhaps:
```
// --- add jsonSchema for validation -----
```

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
```diff
@@ -62,14 +101,14 @@ describe('validate properties', () => {
...
-  it('fails when creating a CoffeeShop with an invalid type', async () => {
+  it('should use sequence-level custom error handler', async () => {
     const invalidCapacityTypeCS = {
       city: 'Toronto',
       phoneNum: '416-111-1111',
       capacity: '10',
     };
     const response = await client
-      .post('/coffee-shops')
+      .patch('/coffee-shops')
```
dhmlau, line 111: just wondering why we're changing this to `patch`? the intention was to create a CoffeeShop.

examples/validation-app/README.md
jannyHou, line 22: nitpick: The comment hightlight the new code in the entire file, so better keep them.

examples/validation-app/src/models/coffee-shop.model.ts
```diff
@@ -17,38 +17,47 @@ export class CoffeeShop extends Entity {
   @property({
     type: 'string',
     required: true,
-    // --- add jsonSchema -----
```
jannyHou, line 20: ditto

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
jannyHou, line 111: hmm why it changes to patch? I guess it's also related to the code change in sequence?

examples/validation-app/src/sequence.ts
```diff
@@ -74,7 +74,7 @@ export class MySequence implements SequenceHandler {
     // 2. customize error for particular endpoint
     if (context.request.url === '/coffee-shops') {
       // if this is a validation error
-      if (err.statusCode === 422) {
+      if (err.statusCode === 422 && context.request.method === 'PATCH') {
```
jannyHou, line 77: same question as https://github.com/strongloop/loopback-next/pull/5937#discussion_r455960327

docs/site/Model.md
jannyHou: @hacksparrow Thanks for adding the default options, then I think we can delete the extra step in doc?

examples/validation-app/README.md
hacksparrow: Re-added with some modification.

hacksparrow: Re-added with some modification.

examples/validation-app/src/models/coffee-shop.model.ts
hacksparrow: Re-added with some modification.

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
hacksparrow: In the process of creating custom error messages, the custom sequence was gobbling up the error messages generated by AJV validation and emitting its own version, and the tests had been written for this behavior. Tests using `errorMessage` would simply fail.

While technically not incorrect, it was an incomplete example of validation example. Validation and customization of errors can be done at different levels.

1. Model level: validation using `jsonSchema` and custom error messages using `errorMessage`
2. Controller-level validation (which, the example app was doing)
3. Sequence-level custom error messages (which, the example app was doing)

We cannot continue to use the custom sequence error generator as-is, since it will emit some other error in place of AJV error, and the tests will fail. However, since it shows a way to create custom errors, it makes sense to keep it - but test it in a non-disruptive manner - so let handle all errors from `PATCH` request.

examples/validation-app/src/sequence.ts
hacksparrow: https://github.com/strongloop/loopback-next/pull/5937#discussion_r456410548

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
hacksparrow: https://github.com/strongloop/loopback-next/pull/5937#discussion_r456410548

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
```diff
@@ -32,24 +32,63 @@ describe('validate properties', () => {
...
+    };
+    const response = await client
+      .post('/coffee-shops')
+      .send(invalidCapacityTypeCS)
+      .expect(422);
+
+    expect(response.body.error.message).to.equal(
+      'The request body is invalid. See error object `details` property for more info.',
+    );
+    expect(response.body.error.details.length).to.equal(4);
```
bajtos, line 88: When this assertion fails on CI, how can we learn more information about the `details` array?

```
Uncaught Error [AssertionError]: expected 3 to be 4
```

At minimum, use something like this:

```suggestion
    expect(response.body.error.details).to.have.length(4);
```

```
Error [AssertionError]: expected Array [ 1, 2, 3 ] to have property length of 4 (got 3)
```

A better solution is to map `details` item into something that will be useful when inspecting test failures. For examples:

```ts
expect(response.body.error.details.map(d => d.code)).to.deepEqual([
  // error code 1
  // error code 2
  // error code 3
  // error code 4
]);
```

This way when the validation fails, we know exactly which validation errors were not reported (or reported erroneously).

packages/rest/src/validation/default-options.ts
```diff
@@ -0,0 +1,7 @@
+import {RequestBodyValidationOptions} from '../types';
+
+export const defaultValidationOptions: RequestBodyValidationOptions = {
+  $data: true,
+  ajvKeywords: true,
+  ajvErrors: true,
+};
```
bajtos, line 7: Is this fully backwards compatible? When existing projects update to the new version of `@loopback/rest`, will the validation behave the same way (accept the same set of input values, reject the same set of input values) and report the same validation errors as before?

packages/rest/src/coercion/coerce-parameter.ts
```diff
@@ -167,7 +168,7 @@ function coerceBoolean(data: string | object, spec: ParameterObject) {
 async function coerceObject(
   input: string | object,
   spec: ParameterObject,
-  options?: RequestBodyValidationOptions,
+  options: RequestBodyValidationOptions = defaultValidationOptions,
```
bajtos, line 171: I have mixed feeling about injecting `defaultValidationOptions` in so many places. Shouldn't we apply the default at the top level only? I think that means either in `parse-params.provider.ts` and `request-body.validator.ts`; or in `ajv-factory.provider.ts`.

@raymondfeng what's your opinion on this architectural aspect?

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
dhmlau: hm. Not sure if I understand the intention of changing from `POST` to `PATCH` yet. 
- If we want to show how to custom error messages using `errorMessage` and we want varieties of not just calling POST, I'm good with it, because all the tests here are POST.
- If we change it because `errorMessage` won't work with `POST`, then that would be something we need to look into.

hacksparrow: > If we change it because errorMessage won't work with POST, then that would be something we need to look into.

That test is especially written for the sequence modifying the error object and creating custom error object. It is not a part of AJV validation.

If we have a consensus on this test not belonging here, then we can remove it along with the custom sequence handler. I'd like to especially hear from @jannyHou.

packages/rest/src/validation/default-options.ts
hacksparrow: Not fully compatible, but the incompatibility will be only at the "message text" level, AFAIK.

If someone had this code:

```ts
@property({
  type: 'string',
  required: true,
  jsonSchema: {
    maxLength: 10,
    minLength: 1,
    errorMessage: 'Name should be between 10 and 30 characters.',
  },
})
```
They would have never gotten "'Name should be between 10 and 30 characters.'", instead they would see the default AJV validation error message.

With the changes in the PR, they will now see "Name should be between 10 and 30 characters.'".

Technically it is a breaking change, but the impact should be minimal for average scenarios.

packages/rest/src/coercion/coerce-parameter.ts
raymondfeng: I agree with @bajtos: let's try to apply the default options at the root of the call stack.

hacksparrow: It's now defined on `AjvFactoryProvider` as a static `readonly` property.

examples/validation-app/src/__tests__/acceptance/validate.acceptance.ts
jannyHou: Had a chat with @hacksparrow, thanks for the explanation and it's fair enough to me to specify an operation to test the customized error, and let other operations return the ajv validation error as is 👍 
Probably add a comment in this test and the sequence file to explain the reason.

packages/rest/src/parser.ts
```diff
@@ -60,7 +61,7 @@ async function buildOperationArguments(
   pathParams: PathParameterValues,
   body: RequestBody,
   globalSchemas: SchemasObject,
-  options: RequestBodyValidationOptions = {},
+  options: RequestBodyValidationOptions,
```
jannyHou, line 64: can we keep `={}` if not using `AjvFactoryProvider.defaultValidationOptions`?

hacksparrow: It is not obvious here, but the option is of a non-exported function which is called in only one place with the default as `AjvFactoryProvider.defaultValidationOptions`, so I removed it to reduce redundancy.

Making the default `{}` will undo the changes of this PR, IF this function happens to be called without an `option` in another place in future, so let's keep it as it is. Remember, default `{}` is what caused the configured custom error messages to not be used 😄

packages/rest/src/validation/ajv-factory.provider.ts
```diff
@@ -29,12 +30,17 @@ const ajvErrors = require('ajv-errors');
  */
 @bind({scope: BindingScope.SINGLETON})
 export class AjvFactoryProvider implements Provider<AjvFactory> {
+  static readonly defaultValidationOptions: RequestBodyValidationOptions = {
+    $data: true,
+    ajvKeywords: true,
+    ajvErrors: true,
+  };
```
raymondfeng, line 37: IMO, a constant is probably simpler here:
```ts
export DEFAULT_AJV_VALIDATION_OPTIONS: RequestBodyValidationOptions = {
  $data: true,
  ajvKeywords: true,
  ajvErrors: true,
};
```

raymondfeng: Now the ajv validation is used beyond request body. We probably should rename `RequestBodyValidationOptions` to `ValidationOptions` or `AjvValidationOptions`. For backward compatibility:

```ts
export type RequestBodyValidationOptions = ValidationOptions;
```

packages/rest/src/parser.ts
hacksparrow: Btw, I have made `DEFAULT_AJV_VALIDATION_OPTIONS` the default now.

packages/rest/src/types.ts
```diff
@@ -225,3 +225,6 @@ export interface Session {
 export interface RequestWithSession extends Request {
   session: Session;
 }
+
+// For backwards compatibility
```
bajtos, line 229: 👍🏻 

Perhaps add a `TODO(SEMVER-MAJOR)` comment to make it easy to find this opportunity for backwards-incompatible cleanup when working on a new major version?

PR #5959, opened 2020-07-17
Refactor extensions to move framework modules to peerDependencies

bajtos:
As discussed several times in the past (most recently in https://github.com/strongloop/loopback-next/pull/5927#discussion_r453727653), extensions should use framework modules from the target application via `peerDependencies`. This pull request makes that change happen.

I split the changes into three commits:

1.  Document the best practice in "Creating components"
2. Update existing packages (breaking change 💥)
3. Update CLI template

## Checklist

### DIFF
--- a/docs/site/Creating-components.md
+++ b/docs/site/Creating-components.md
@@ -36,6 +36,59 @@ export class MyComponent implements Component {
 }
 ```
 
+## Working with dependencies
+
+Extensions should preferably use LoopBack modules installed by the target
+application, to allow application developers to choose the version of framework
+modules they want to use. For example, they may want to hold to an older version
+until a regression is fixed in the latest version, or even use their own fork to
+have a bug fix available before it's officially published.
+
+We recommend to use `peerDependencies` to specify what packages is your
+extension expecting in the target application and `devDependencies` to make
+these packages available for extension tests.
+
+For example:
+
+```json
+{
+  "name": "my-lb4-extension",
+  "version": "1.0.0",
+  "dependencies": {
+    "tslib": "^2.0.0"
+  },
+  "peerDependencies": {
+    "@loopback/core": "^2.9.1",
+    "@loopback/rest": "^5.2.0"
+  },
+  "devDependencies": {
+    "@loopback/build": "^6.1.0",
+    "@loopback/core": "^2.9.1",
+    "@loopback/eslint-config": "^8.0.3",
+    "@loopback/rest": "^5.2.0",
+    "@loopback/testlab": "^3.2.0"
+  }
+}
+```
+
+It is also possible for a single extension version to support multiple versions
+of a framework dependency:
+
+```json
+{
+  "peerDependencies": {
+    "@loopback/core": "^1.9.0 || ^2.0.0"
+  }
+}
+```
+
+{% include important.html content=' Shrinking the supported version range of a
+peer dependency is a breaking change that should trigger a semver-major release.
+
+For example, changing the range from `"^1.5.3"` to `"^1.5.3 || ^2.0.0"` is
+backwards compatible, while changing the range from `"^1.5.3"` to `"^2.0.0"` is
+a breaking change. ' %}
+
 ## Injecting the target application instance
 
 You can inject anything from the context and access them from a component. In
--- a/extensions/apiconnect/package.json
+++ b/extensions/apiconnect/package.json
@@ -38,14 +38,18 @@
     "url": "https://github.com/strongloop/loopback-next.git",
     "directory": "extensions/apiconnect"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@loopback/core": "^2.9.1",
-    "@loopback/rest": "^5.2.0",
+    "@loopback/rest": "^5.2.0"
+  },
+  "dependencies": {
     "tslib": "^2.0.0"
   },
   "devDependencies": {
     "@loopback/build": "^6.1.0",
+    "@loopback/core": "^2.9.1",
     "@loopback/eslint-config": "^8.0.3",
+    "@loopback/rest": "^5.2.0",
     "@loopback/testlab": "^3.2.0",
     "@types/node": "^10.17.27"
   }
--- a/extensions/authentication-jwt/package.json
+++ b/extensions/authentication-jwt/package.json
@@ -20,22 +20,28 @@
   "publishConfig": {
     "access": "public"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@loopback/authentication": "^4.2.9",
     "@loopback/core": "^2.9.1",
     "@loopback/rest": "^5.2.0",
+    "@loopback/service-proxy": "^2.3.4"
+  },
+  "dependencies": {
     "@loopback/rest-explorer": "^2.2.6",
     "@loopback/security": "^0.2.14",
-    "@loopback/service-proxy": "^2.3.4",
     "@types/bcryptjs": "2.4.2",
     "bcryptjs": "^2.4.3",
     "jsonwebtoken": "^8.5.1"
   },
   "devDependencies": {
+    "@loopback/authentication": "^4.2.9",
     "@loopback/boot": "^2.3.5",
     "@loopback/build": "^6.1.0",
+    "@loopback/core": "^2.9.1",
     "@loopback/eslint-config": "^8.0.3",
     "@loopback/repository": "^2.9.0",
+    "@loopback/rest": "^5.2.0",
+    "@loopback/service-proxy": "^2.3.4",
     "@loopback/testlab": "^2.0.2",
     "@types/lodash": "^4.14.157",
     "@types/node": "^10.17.27",
--- a/packages/boot/package.json
+++ b/packages/boot/package.json
@@ -23,8 +23,10 @@
   "publishConfig": {
     "access": "public"
   },
+  "peerDependencies": {
+    "@loopback/core": "^2.9.1"
+  },
   "dependencies": {
-    "@loopback/core": "^2.9.1",
     "@loopback/model-api-builder": "^2.1.9",
     "@loopback/repository": "^2.9.0",
     "@loopback/service-proxy": "^2.3.4",
@@ -36,6 +38,7 @@
   },
   "devDependencies": {
     "@loopback/build": "^6.1.0",
+    "@loopback/core": "^2.9.1",
     "@loopback/eslint-config": "^8.0.3",
     "@loopback/rest": "^5.2.0",
     "@loopback/rest-crud": "^0.8.9",
--- a/extensions/authentication-passport/package.json
+++ b/extensions/authentication-passport/package.json
@@ -39,20 +39,25 @@
     "url": "https://github.com/strongloop/loopback-next.git",
     "directory": "extensions/authentication-passport"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@loopback/authentication": "^4.2.9",
     "@loopback/core": "^2.9.1",
-    "@loopback/rest": "^5.2.0",
+    "@loopback/rest": "^5.2.0"
+  },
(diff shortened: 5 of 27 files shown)


packages/boot/package.json
```diff
@@ -52,5 +52,10 @@
     "type": "git",
     "url": "https://github.com/strongloop/loopback-next.git",
     "directory": "packages/boot"
+  },
+  "peerDependencies": {
+    "@loopback/core": "^2.9.1",
+    "@loopback/repository": "^2.9.0",
+    "@loopback/service-proxy": "^2.3.4"
```
bajtos, line 59: This is potentially tricky.

Before my change, it was possible to use `@loopback/boot` in applications that don't have `@loopback/service-proxy` in their dependencies. `@loopback/boot` would install it's own copy that would be pretty much ignored at runtime.

With the change in place, apps must install both `@loopback/repository` and `@loopback/service-proxy` to be able to use `@loopback/boot`. I wish there was something like `optionalPeerDependencies`!

I see this as a sign that `@loopback/boot` is providing too many features and should be split into smaller packages.
1. It defines contract for Booter implementations and few base classes to help with implementation.
2. It implements the bootstrapping algorithm (`BootMixin`).
3. It provides booters for common artifacts (controllers, repositories, services, etc.).

I think the following structure would be better:
1. A new module to define Booter contract and possibly base classes. Prior art: https://github.com/strongloop/loopback-next/tree/master/packages/model-api-builder
2. `@loopback/boot` should provide `BootMixin` and artifact mixins that can be loaded using `@loopback/core` only.
3. Artifact booters requiring optional extensions (repository booter, service booter, etc.) should be moved either to those extensions (`@loopback/repository`, `@loopback/service-proxy`) or to a new package. A question to consider: where to put datasource booter that's shared by repositories and service proxies?

Such change is out of scope of this pull request though. The question is how to address the issue above inside this pull request:

1. Are we ok to require apps to install both `repository` and `service-proxy` if they want to use `@loopback/boot`? By default, `lb4 app` includes both, so it may be ok. But then it feels a kind of lame to force users to install packages they don't need.
2. Should I revert the repository & service-proxy part of this pull request and keep those two packages as regular `dependencies`?
3. We can also investigate how to rework booters to not require direct dependency on repository & service-proxy packages. They mostly call application methods mixed in by the user (e.g. `app.repository`) and import only type definitions for internal consumption from repository/service-proxy, so maybe it's ok to declare them in `devDependencies` only? The only exception I found is `ModelMetadataHelper.getModelMetadata`, it's called by `ModelBooter` to check if a class is decorated with `@model()`.

extensions/authentication-jwt/package.json
```diff
@@ -21,21 +21,21 @@
     "access": "public"
   },
   "dependencies": {
-    "@loopback/authentication": "^4.2.9",
-    "@loopback/core": "^2.9.1",
-    "@loopback/rest": "^5.2.0",
     "@loopback/rest-explorer": "^2.2.6",
```
bajtos, line 24: While we are reviewing dependencies, is it intentional that the extension `@loopback/authentication-jwt` installs `@loopback/rest-explorer` as a dependency? Shouldn't the application developer decide if they want to use API Explorer or not?

bajtos:
Added two more commits to reorganize order of package.json fields and move `peerDependencies` before regular and dev dependencies, I feel the information about required peer deps is important when reviewing package metadata and therefore should be specified early on.

packages/boot/package.json
raymondfeng: That's my intention for https://github.com/strongloop/loopback-next/pull/5618.

bajtos: To move this pull request forward, I am going to pick the option 2:

>  revert the repository & service-proxy part of this pull request and keep those two packages as regular dependencies

It's not ideal, but it will give us more time to consider different options and implement a proper long-term fix.

bajtos:
@raymondfeng updated per https://github.com/strongloop/loopback-next/pull/5959#discussion_r457440387, LGTY now?

docs/site/Creating-components.md
```diff
@@ -36,6 +36,59 @@ export class MyComponent implements Component {
...
+
+Extensions should preferably use LoopBack modules installed by the target
+application, to allow application developers to choose the version of framework
+modules they want to use. For example, they may want to hold to an older version
+until a regression is fixed in the latest version, or even use their own fork to
+have a bug fix available before it's officially published.
+
+We recommend to use `peerDependencies` to specify what packages is your
+extension expecting in the target application and `devDependencies` to make
+these packages available for extension tests.
```
raymondfeng, line 49: We should mention that `npm 6` does not install `peerDependencies`. As a workaround, we need to add duplicate entries in `devDependencies` so that they can be installed and used for testing.

extensions/authentication-jwt/package.json
raymondfeng: +1 to remove it.

extensions/authentication-jwt/package.json
```diff
@@ -20,22 +20,28 @@
   "publishConfig": {
     "access": "public"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@loopback/authentication": "^4.2.9",
     "@loopback/core": "^2.9.1",
     "@loopback/rest": "^5.2.0",
+    "@loopback/service-proxy": "^2.3.4"
```
raymondfeng, line 27: Should `@loopback/core` stay as a regular dependency?

extensions/apiconnect/package.json
```diff
@@ -38,14 +38,18 @@
     "url": "https://github.com/strongloop/loopback-next.git",
     "directory": "extensions/apiconnect"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@loopback/core": "^2.9.1",
-    "@loopback/rest": "^5.2.0",
+    "@loopback/rest": "^5.2.0"
```
raymondfeng, line 43: Should `@loopback/core` stay as a regular dependency? We need to clarify the criteria between `dependencies` and `peerDependencies`. I'm thinking about the following:

1. If the dependency needs to be consistent with the application level, it's a peer dependency (always use the one from the application level)
2. Otherwise, it should be a regular dependency (and possibly with a different version than the application level)

raymondfeng: Another criteria:

If the dependency is the target extension point for this package to contribute extensions, it's a peer dependency. 

Using this rule, `@loopback/core` should be a regular dependency while `@loopback/rest` should be a peer dependency.

extensions/authentication-jwt/package.json
bajtos: AFAICT, API Explorer is used by a test application, see 

https://github.com/strongloop/loopback-next/blob/cb9e6d79d4db265f39de551f81a824a5cb5c3dcb/extensions/authentication-jwt/src/__tests__/fixtures/application.ts#L38

I am going to move it to `devDependencies` - see https://github.com/strongloop/loopback-next/pull/6009

docs/site/Creating-components.md
```diff
@@ -36,6 +36,59 @@ export class MyComponent implements Component {
...
+}
+```
+
+It is also possible for a single extension version to support multiple versions
+of a framework dependency:
+
+```json
+{
+  "peerDependencies": {
+    "@loopback/core": "^1.9.0 || ^2.0.0"
```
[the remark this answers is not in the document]
bajtos: See the discussions in https://github.com/strongloop/loopback-next/pull/4937 and your pull request https://github.com/strongloop/loopback-next/pull/4972.

Quoting from https://github.com/strongloop/loopback-next/pull/4972#discussion_r399998539:

> Let's be conservative to use the `^<pkgVersion>` for peer dependencies as a release is cut.

The current implementation:

https://github.com/strongloop/loopback-next/blob/6e916494f1bcbdf5ba5ecd20ac8cbadcbae4fc9f/bin/update-peer-deps.js#L38-L41

extensions/apiconnect/package.json
bajtos: > - If the dependency needs to be consistent with the application level, it's a peer dependency (always use the one from the application level)
> - If the dependency is the target extension point for this package to contribute extensions, it's a peer dependency.
> - Otherwise, it should be a regular dependency (and possibly with a different version than the application level)

Your criteria make sense to me, however I have a different view on how to interpret them.

In my opinion, there should be only one version of `@loopback/core` installed in any application. Is see the core packages as a cornerstone, a "LoopBack runtime". Among other things, it provides foundation for decorators and manipulation with metadata (`MetadataInspector`, `ParameterDecoratorFactory`, etc.) I am concerned that allowing multiple versions of `@loopback/core` can introduce subtle bugs that will be difficult to troubleshoot, for example when a decorator uses different code for setting metadata than is used by the application to look up that metadata.

Also, if the application developer decides to use their forked version of `@loopback/core` (e.g. to fix a bug), then I want extensions to use the fixed version of core, not the vanilla version from npm.

I would also like to add one more criteria:

- If the dependency is typically installed in target applications, and the extension can support multiple major versions of that dependency at the same time, then the dependency should be a peer dependency.

The use case to consider: where a single version of an extension (e.g. `@loopback/extension-health`) wants to support multiple version of the framework, so that the maintainers can have a single branch and don't have to maintain different branches (major-version release lines) for each framework version. Let's say the extension depends on `@loopback/core@2` and the application depends on `@loopback/core@1`. This is very reasonable, since the only breaking change listed in `@loopback/core@2.0.0` changelog is dropped support for Node.js 8.x. When such extension is installed into the target app, we end up with two copies of `@loopback/core`, where the extension uses a different "LoopBack runtime" than the target application it's installed to.

Let's discuss!

raymondfeng: That's a valid use case that one extension package can support multiple versions of the LoopBack runtime. But it's hard to ensure that with our CI and release process though.

In addition to the containing application, we probably should also take into consideration that certain LoopBack packages can be used standalone outside a scaffolded LB4 application.

bajtos: > That's a valid use case that one extension package can support multiple versions of the LoopBack runtime. But it's hard to ensure that with our CI and release process though.

I now. Personally, I am fine to keep the support for older major versions on a "best effort" basis with no CI check for now. We can setup a more comprehensive CI solution once there are complaints from users about incompatibilities.

> In addition to the containing application, we probably should also take into consideration that certain LoopBack packages can be used standalone outside a scaffolded LB4 application.

That's a good argument. IIRC, packages that can be used standalone outside of LB4 apps are usually more low level (`@loopback/metadata`, `@loopback/context` etc.) and don't participate in peer dependencies, because they are accessed via higher-level wrappers (`@loopback/core`).

If a LB package is used by a LB4 application that's not scaffolded via `lb4 app`, then I believe the app is still going to depend on `@loopback/core`, therefore my arguments above still apply.

Are you envisioning a project where the developer wants to use an extension depending on `@loobpack/core` but the application does not depend on `@loopback/core`? Would that work at all?

raymondfeng: > Are you envisioning a project where the developer wants to use an extension depending on @loobpack/core but the application does not depend on @loopback/core? Would that work at all?

No. At minimum, we'll use `@loopback/core`.

bajtos: > No. At minimum, we'll use @loopback/core.

In that case, I think it's reasonable to for extensions to declare `@loopback/core` as a peer dependency, right?

Do you agree to keep the currently proposed direction? May I proceed to rebase this PR on top of latest master (there are going to be many merge conflicts to resolve, so I don't want to do it too many times) and get this PR landed soon?

raymondfeng: Sure.

bajtos:
I'll open a series of smaller PRs to land these changes in a way that will minimize merge conflicts.

The first one: https://github.com/strongloop/loopback-next/pull/6243

PR #5590, opened 2020-05-28
[DO NOT MERGE] rework eslint config to use proper TS projects

bajtos:
Disable `createDefaultProgram` option and ensure every linted file belongs to some `tsconfig`, see https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/MONOREPO.md#getting-started---monorepo-configuration

Unfortunately, eslint is sometimes running out of memory with this new setup. I think we have to wait until https://github.com/typescript-eslint/typescript-eslint/issues/1192 and/or https://github.com/typescript-eslint/typescript-eslint/issues/2094 is resolved by typescript-eslint.

The pull request contains a single "all-in-one" commit right now, we will need to split it into multiple smaller commits before landing. I think the change in eslint-config is breaking backwards compatibility and needs to be described as such.

Migration guide for LB users:

> Make sure all `.ts` and `.js` files in your repository are either included by `tsconfig.json` or excluded via `.eslintignore` file.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-next) 👈

- [ ] `npm test` passes on your machine
- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)
- [ ] API Documentation in code was updated
- [ ] Documentation in [/docs/site](../tree/master/docs/site) was updated
- [ ] Affected artifact templates in `packages/cli` were updated
- [ ] Affected example projects in `examples/*` were updated

👉 [Check out how to submit a PR](https://loopback.io/doc/en/lb4/submitting_a_pr.html) 👈

PR #5983, opened 2020-07-21
eslint-config: disable `createDefaultProgram`

bajtos:
While playing with eslint (https://github.com/strongloop/loopback-next/pull/5590, typescript-eslint/typescript-eslint#1192 and typescript-eslint/typescript-eslint#2094), I realized we don't have to migrate our eslint setup to project references in order to get rid of `createDefaultProgram` & its performance penalty.

In this pull request, I am proposing an alternative improvement and fixing few problems discovered by eslint along the way.

1. Add missing `await` statement to `testlab` code checking if a `.map` file exists when copying `.js` files to sandbox. The problem was discovered by the linter.
2. Update the `tsconfig` file used by `eslint` to include all files we are checking.
3. Remove `createDefaultProgram` flag from the shared/default eslint config (this is a breaking change).

Timing info (eslint with no cache):

- Before: ~3.5min
- After: ~1min

## Checklist

### DIFF
--- a/packages/testlab/src/test-sandbox.ts
+++ b/packages/testlab/src/test-sandbox.ts
@@ -151,7 +151,7 @@ export class TestSandbox {
       await outputFile(dest, content, {encoding: 'utf-8'});
     }
 
-    if (parse(src).ext === '.js' && pathExists(src + '.map')) {
+    if (parse(src).ext === '.js' && (await pathExists(src + '.map'))) {
       const srcMap = src + '.map';
       await appendFile(dest, `\n//# sourceMappingURL=${srcMap}`);
     }
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -3,18 +3,22 @@
     "tsconfig.build.json is used by eslint to constrain files to be checked",
     "tsconfig.json is used by tsc"
   ],
-  "extends": "./packages/build/config/tsconfig.common.json",
+  "compilerOptions": {
+    "skipLibCheck": true,
+    "allowJs": true,
+    "noEmit": true
+  },
   "include": [
-    "acceptance",
-    "fixtures",
-    "benchmark",
-    "examples",
-    "packages",
-    "extensions"
+    "**/.eslintrc.js",
+    "**/*.js",
+    "**/*.ts"
   ],
   "exclude": [
     "**/node_modules/**",
     "**/dist/**",
-    "**/*.d.ts"
+    "**/*.d.ts",
+    "coverage",
+    "docs/_loopback.io",
+    "docs/_preview"
   ]
 }
--- a/packages/cli/generators/project/templates/.eslintignore
+++ b/packages/cli/generators/project/templates/.eslintignore
@@ -1,3 +1,4 @@
 node_modules/
 dist/
 coverage/
+.eslintrc.js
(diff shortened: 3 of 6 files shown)


packages/testlab/src/test-sandbox.ts
```diff
@@ -151,7 +151,7 @@ export class TestSandbox {
       await outputFile(dest, content, {encoding: 'utf-8'});
     }
 
-    if (parse(src).ext === '.js' && pathExists(src + '.map')) {
+    if (parse(src).ext === '.js' && (await pathExists(src + '.map'))) {
```
bajtos, line 154: `pathExists` returns a `Promise`, therefore this condition was always resolved as truthy.

bajtos:
Timing data from Travis CI:

- `npm run lint` on master: 164.38s ([link](https://travis-ci.com/github/strongloop/loopback-next/jobs/363180167#L237))
- `npm run lint` in my PR: 95.17s ([link](https://travis-ci.com/github/strongloop/loopback-next/jobs/363322216#L240))

raymondfeng:
@bajtos Nice gain! Please investigate CI failures.

tsconfig.build.json
```diff
@@ -3,18 +3,40 @@
...
+    "packages/eslint-config/eslintrc.js",
+    "packages/repository/examples/**/*.js",
+    "packages/repository/examples/**/*.ts",
+    "packages/tsdocs/bin/*.js"
   ],
   "exclude": [
-    "**/node_modules/**",
-    "**/dist/**",
-    "**/*.d.ts"
+    "node_modules"
```
raymondfeng, line 40: Is `node_modules` the same as `**/node_modules/**`?

tsconfig.build.json
```diff
@@ -3,18 +3,40 @@
...
+    "extensions/context-explorer/public/**/*.js",
+    "packages/booter-lb3app/fixtures/**/*.js",
+    "packages/build/**/*.js",
+    "packages/build/**/*.ts",
+    "packages/cli/**/*.js",
+    "packages/cli/**/*.ts",
+    "packages/eslint-config/eslintrc.js",
+    "packages/repository/examples/**/*.js",
+    "packages/repository/examples/**/*.ts",
+    "packages/tsdocs/bin/*.js"
```
raymondfeng, line 37: Hmm, do we have a better way to track js/ts files?

tsconfig.build.json
bajtos: I believe so, they are showing `node_modules` in the official docs (https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#examples). I used `tsc --listFilesOnly` to verify exactly which files are included by `tsconfig.build.json`.

tsconfig.build.json
bajtos: I can use a wider wildcard like `**/.js`, but I am not sure if such wildcard would not pick up files we don't want to pick. I can take another look.

tsconfig.build.json
bajtos: Reverted back to the original exclude list in 6daeb75

tsconfig.build.json
bajtos: Implemented a simpler solution in 6daeb75.

bajtos:
I added two commits:
- 6daeb75 greatly simplifies the list of files specified in our tsconfig file
- 6e5cb08 fixes the failures of `npm test` in scaffolded projects

@raymondfeng LGTY now?

PR #6025, opened 2020-07-28
fix(rest): skip operations from api spec that matches explicit routes

raymondfeng:
`RestServer.api()` registers an open api spec with handlers/controller methods. `RestServer.getApiSpec()` populates paths from routes. When a route is added/removed from the server, it triggers recreation of `httpHandler`, which tries to set up routes from `API_SPEC` (which used to be mutated by `RestServer.getApiSpec()`). As a result, some routes does not have `x-operation` or `x-controller-name` and an error is thrown.

## Checklist

### DIFF
--- a/packages/rest/src/__tests__/integration/rest.server.integration.ts
+++ b/packages/rest/src/__tests__/integration/rest.server.integration.ts
@@ -1220,6 +1220,60 @@ paths:
     await server.stop();
   });
 
+  it('registers handler routes after start and getApiSpec', async () => {
+    const server = await givenAServer();
+    await server.start();
+    const client = createClientForHandler(server.requestHandler);
+    // No `/greet` is present
+    await client.get('/greet').expect(404);
+
+    // Add DummyController after server.start
+    const greetSpec = {
+      responses: {
+        200: {
+          content: {'text/plain': {schema: {type: 'string'}}},
+          description: 'greeting of the day',
+        },
+      },
+    };
+
+    // Add `GET /greet` route
+    server.route('get', '/greet', greetSpec, function greet() {
+      return 'Hello';
+    });
+
+    // Now `GET /greet` is available
+    await client.get('/greet').expect(200, 'Hello');
+
+    // Update api specs to verify serving API explorer does not mess up
+    await server.getApiSpec();
+
+    // Add `GET /greet1`
+    const binding = server.route('get', '/greet1', greetSpec, function greet() {
+      return 'Hello1';
+    });
+
+    // `GET /greet` is still available
+    await client.get('/greet').expect(200, 'Hello');
+
+    // The newly added `GET /greet1` is available
+    await client.get('/greet1').expect(200, 'Hello1');
+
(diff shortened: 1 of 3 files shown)


raymondfeng:
The issue was reported from the Slack channel:

```
Hi, I am trying to add APIs dynamically at runtime.  When I use app.api(spec), it works.  When I use app.route(verb, path, spec, handler), it seems to work when nothing happen between consecutive app.route calls.  If something happen after the first app.route call, the second app.route call will cause an error.  For example
npm start
hit http://localhost:3000/explorer/
call app.route
hit http://localhost:3000/explorer/
call app.route
#5 will cause the following error.  If I skip #4, then #5 will be fine.  Is this supported?  Thanks.
(node:12452) UnhandledPromiseRejectionWarning: Error: There is no handler configured for operation "post /4/v1/coachflow/service/1.4bce11e8-45fd-4c3a-b728-c62f5f4069ad
    at RestServer._setupOperation (W:\test-openapi4\node_modules\@loopback\rest\src\rest.server.ts:500:11)
    at RestServer._createHttpHandler (W:\test-openapi4\node_modules\@loopback\rest\src\rest.server.ts:451:14)
    at Object.observe (W:\test-openapi4\node_modules\@loopback\rest\src\rest.server.ts:382:14)
```

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -1220,6 +1220,42 @@ paths:
...
+
+    // Now update api specs
+    await server.getApiSpec();
+
+    server.route('get', '/greet1', greetSpec, function greet() {
+      return 'Hello';
+    });
+
+    // Now `/greet` is available
+    await client.get('/greet1').expect(200, 'Hello');
```
jannyHou, line 1254: I think you wanted to test `/greet` instead of `/greet1` here? probably both.

raymondfeng:
@jannyHou I had to improve the fix to handle the case when a route is unbound. PTAL.

wnwaynelee:
I verified both app.route and restServer.unbind are fixed with the latest update.  Thanks.

packages/rest/src/__tests__/integration/rest.server.integration.ts
```diff
@@ -1220,6 +1220,57 @@ paths:
...
+
+    server.route('get', '/greet', greetSpec, function greet() {
+      return 'Hello';
+    });
+
+    // Now `/greet` is available
+    await client.get('/greet').expect(200, 'Hello');
+
+    // Now update api specs
+    await server.getApiSpec();
```
jannyHou, line 1248: a bit confused here, why have to call `await server.getApiSpec()` before binding, instead of after?

same question for the next test when unbind the route.

raymondfeng: It's to verify that getApiSpec() won't mess up the rebuild of http handler. The original issue was caused by that.

Issue #6001, opened 2020-07-24
Error in the "Dynamically adding models, repositories, and controllers during runtime" section


kvbogdanov:
There is a section "Dynamically adding models, repositories, and controllers during runtime" https://loopback.io/doc/en/lb4/Dynamic-models-repositories-controllers.html in the docs where we can find a way to add a model to API during runtime.

But. Following these steps I cannot create a CRUD controller with fields described in a model definition

```
    const bookDef = new ModelDefinition({
      name: 'book',
      properties: {
        id: {
          type: 'Number',
          required: true,
          id: 1,
        },
        title: {
          type: 'String',
          required: false,
        },
      },
    });

    const BookModel = defineModelClass<typeof Entity, {id: number; title?: string}>(
      Entity,
      bookDef,
    );

    const BookRepository = defineCrudRepositoryClass(BookModel);

    inject(`datasources.mongo`)(BookRepository, undefined, 0);
    const repoBinding = app.repository(BookRepository);

    const basePath = '/booktest';

    const DynamicController0 = defineCrudRestController(BookModel, {basePath});
    inject(repoBinding.key)(DynamicController0, undefined, 0);
    app.controller(DynamicController0);
```

CRUD controller created successfully but with no params (actually, only with params of the base Entity/Model). 

![Снимок экрана 2020-07-24 в 14 15 31](https://user-images.githubusercontent.com/2289017/88368612-9f638f00-cdb8-11ea-9c77-b7b3821deb67.png)

Expected that API could add object Book with params _id_ and _title_

There are no errors or warnings in console.

hacksparrow:
Where are you executing that code? Can you share a bare minimal app reproducing the behavior?

hacksparrow:
This is a working example, add it in the constructor of your Application class. Please make sure to import the necessary dependencies.

```ts

    @model()
    class Product extends Entity {
      @property({id: true})
      id: number;

      @property({required: true})
      name: string;

      @property()
      description?: string;

      constructor(data: Partial<Product>) {
        super(data);
      }
    }

    const db = new juggler.DataSource({connector: 'memory'});
    const ProductRepository = defineCrudRepositoryClass(Product);
    const repo = new ProductRepository(db);

    const CrudRestController = defineCrudRestController<
      Product,
      typeof Product.prototype.id,
      'id'
    >(Product, {basePath: '/products'});

    class ProductController extends CrudRestController {
      constructor() {
        super(repo);
      }
    }

[...]

PR #6036, opened 2020-07-31
fix(repository): apply `@model` to generated model class

hacksparrow:
Schema was not generated for Model classes generated by `defineModelClass()`, since it did not contain model definition details. This PR fixes it.

Addresses https://github.com/strongloop/loopback-next/issues/6001.

## Checklist

### DIFF
--- a/packages/repository/src/__tests__/unit/model/define-model-class.unit.ts
+++ b/packages/repository/src/__tests__/unit/model/define-model-class.unit.ts
@@ -6,6 +6,7 @@
 import {expect} from '@loopback/testlab';
 import {defineModelClass, Entity, Model, ModelDefinition} from '../../..';
 import {AnyObject} from '../../../common-types';
+import {ModelMetadataHelper} from '../../../decorators';
 
 describe('defineModelClass', () => {
   it('creates a Model class', () => {
@@ -21,7 +22,7 @@ describe('defineModelClass', () => {
 
     // Verify that typedefs allows us to access static Model properties
     expect(DataTransferObject.modelName).to.equal('DataTransferObject');
-    expect(DataTransferObject.definition).to.equal(definition);
+    expect(DataTransferObject.definition).to.deepEqual(definition);
 
     // Verify that typedefs allows us to create new model instances
     const instance = new DataTransferObject({title: 'a title'});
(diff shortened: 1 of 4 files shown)


packages/repository/src/define-model-class.ts
```diff
@@ -60,6 +61,9 @@ export function defineModelClass<
     Props
   >;
   assert.equal(modelClass.name, modelName);
+
+  // Apply `@model(definition)` to the generated class
+  model(definition)(modelClass);
   modelClass.definition = definition;
```
bajtos, line 67: Do we still need to set `modelClass.definition`? I believe `@model` decorator handles that part automatically.

hacksparrow: I just commented the line and ran the tests, three of them fail. So looks like there's more to it.

hacksparrow:
@bajtos I have added the suggested unit test in `@loopback/repositiory`.

packages/repository/src/define-model-class.ts
raymondfeng: Can you check why they are failing? The `@model` calls `buildModelDefinition()`, which should set `definition` to the class.

hacksparrow: Using `deepEqual` instead of `equal` fixed it.

packages/repository/src/__tests__/unit/model/define-model-class.unit.ts
```diff
@@ -106,4 +107,19 @@ describe('defineModelClass', () => {
...
+    });
+
+    const Book = defineModelClass<typeof Model, {title: string}>(
+      Model,
+      definition,
+    );
+
+    const meta = ModelMetadataHelper.getModelMetadata(Book);
+
+    expect(meta).to.be.instanceOf(ModelDefinition);
```
bajtos, line 123: Please check that we have the expected model definition in the metadata.

```suggestion
    expect(meta).to.deepEqual(definition);
```

bajtos:
(Please remember to clean up the commit history before landing.)

packages/repository/src/define-model-class.ts
raymondfeng: `equal` is for the same object.

raymondfeng:
Hmm, CI is failing.

PR #6038, opened 2020-07-31
chore(example-todo): ignore coverage for geocode

nabdelgadir:
I'm proposing we ignore coverage for this section, because for unrelated PRs, this section will cause the coverage to decrease if the geocoder service is not up and running.

## Checklist

### DIFF
--- a/examples/todo/src/controllers/todo.controller.ts
+++ b/examples/todo/src/controllers/todo.controller.ts
@@ -47,6 +47,9 @@ export class TodoController {
     if (todo.remindAtAddress) {
       const geo = await this.geoService.geocode(todo.remindAtAddress);
 
+      // ignoring because if the service is down, the following section will
+      // not be covered
+      /* istanbul ignore next */
       if (!geo[0]) {
         // address not found
         throw new HttpErrors.BadRequest(


examples/todo/src/controllers/todo.controller.ts
```diff
@@ -47,6 +47,7 @@ export class TodoController {
     if (todo.remindAtAddress) {
       const geo = await this.geoService.geocode(todo.remindAtAddress);
 
+      /* istanbul ignore next */
       if (!geo[0]) {
```
bajtos, line 51: I am not sure if this is the right fix. Can we please add a test to verify handling of this error case instead? Adding "istanbul ignore next" would be ok if such test was too difficult to write.

nabdelgadir: There is a test for this case [here](https://github.com/strongloop/loopback-next/blob/geo-coverage/examples/todo/src/__tests__/acceptance/todo.acceptance.ts#L103), but the issue is when the service is down, it won't run that test, which is causing the coverage to decrease.

bajtos: I see, makes sense 👍🏻

bajtos: @nabdelgadir Can you please add a code comment to capture your explanation for future readers?

nabdelgadir: Added, thanks!

PR #6105, opened 2020-08-13
feat(rest): improve validation errors for invalid parameter value

bajtos:
Rework the validation code to use exiting `RestHttpErrors.invalidData` error instead of a custom `BadRequest` error. This way the error object includes the parameter name in the error message & properties and has a machine-readable `code` property.

See also #5808 which introduced parameter coercion.

Surprisingly, all existing tests are passing with my change in place, so I didn't make any changes in the test suite. Please let me know if you think I should write a new test 🤷🏻 

## Checklist

### DIFF
--- a/packages/rest/src/rest-http-error.ts
+++ b/packages/rest/src/rest-http-error.ts
@@ -11,7 +11,7 @@ export namespace RestHttpErrors {
     name: string,
     extraProperties?: Props,
   ): HttpErrors.HttpError & Props {
-    const msg = `Invalid data ${JSON.stringify(data)} for parameter ${name}!`;
+    const msg = `Invalid data ${JSON.stringify(data)} for parameter "${name}".`;
     return Object.assign(
       new HttpErrors.BadRequest(msg),
       {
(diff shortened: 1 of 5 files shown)


packages/rest/src/rest-http-error.ts
```diff
@@ -51,11 +51,15 @@ export namespace RestHttpErrors {
 
   export const INVALID_REQUEST_BODY_MESSAGE =
     'The request body is invalid. See error object `details` property for more info.';
-  export function invalidRequestBody(): HttpErrors.HttpError {
+
+  export function invalidRequestBody(
+    details: ValidationErrorDetails[],
```
dhmlau, line 56: would there be backward compatibility issue for adding a parameter?

hacksparrow:
What about tests?

packages/rest/src/rest-http-error.ts
bajtos: Good catch! 

In the strict interpretation, this is a breaking change because existing callers of `invalidRequestBody` must be updated. However, I consider `invalidRequestBody` as an internal helper that's not a part of our public API, therefore I think it's ok to make this change in a semver-minor release.

bajtos:
@hacksparrow 

> What about tests?

bajtos:
@hacksparrow 

> What about tests?

See PR description:

> Surprisingly, all existing tests are passing with my change in place, so I didn't make any changes in the test suite. Please let me know if you think I should write a new test 🤷🏻

Do you have any suggestions on where to put the new tests?

hacksparrow:
@bajtos A test showing the expected behavior for invalid parameter values. If the code works the same for "valid" and "invalid" values, it means there is no distinction between them for the program.

It seems to me, we'll have to dig deeper and make some changes somewhere, and add a test that will fail for invalid values.

bajtos:
@hacksparrow added a new test, see https://github.com/strongloop/loopback-next/pull/6105/files#diff-f1805dfae45a4ba243a038f5794b800cR257

LGTY now?

Issue #1789, opened 2018-10-01
lb4 controller - inferring the ID type 


marioestradarosa:
## Description / Steps to reproduce / Feature proposal
I was reviewing this weekend the developer experience when creating an application from scratch and found a possible missing feature in the `lb4 controller`. It asks the user about the ID property something necessary for internal coerce system. 

## Current Behavior
- Asks for the ID property type, even when now we can infer it from the model itself

## Expected Behavior
- Shouldn't ask for this ID property type if it can infer it?. This allows the developer for a better experience when generating a controller.

thoughts?

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

dhmlau:
+1 on inferring the id property when creating the controller.

marioestradarosa:
@dhmlau I am working on this PR (trying to catchup with JS coding again 😄 )

PR #6125, opened 2020-08-16
chore: order of id type prompt in controller generator should be in sync with model's

hacksparrow:
The `id` type order for model generator looks like this:

```sh
? Enter the property name: id
? Property type: (Use arrow keys)
❯ string
  number
  boolean
  object
  array
  date
  buffer
```

When we revisit the `id` in the controller generator, we are prompted with this: 

```sh
? What is the type of your ID? (Use arrow keys)
❯ number
  string
  object
```

Let's follow a consistent order, it makes for a better UX. Small thing, big difference.

## Checklist

### DIFF
--- a/packages/cli/generators/controller/index.js
+++ b/packages/cli/generators/controller/index.js
@@ -181,9 +181,9 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
         type: 'list',
         name: 'idType',
         message: g.f('What is the type of your ID?'),
-        choices: ['number', 'string', 'object'],
+        choices: ['string', 'number', 'object'],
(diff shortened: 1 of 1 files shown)


dhmlau:
There was a discussion in https://github.com/strongloop/loopback-next/issues/1789 to infer the type of the id. I think this might be better in terms of developer experience, though it involves more work. WDYT, @hacksparrow @bajtos?

hacksparrow:
@dhmlau agreed, that would be the preferred experience. I will work on a separate PR. Let's keep this one open till I open the new PR.

hacksparrow:

PR #6126, opened 2020-08-16
chore: get rid of default banner in model and repo generators

hacksparrow:
**Before**

```sh
? Please select the model base class (Use arrow keys)
❯ Entity (A persisted model with an ID)
  Model (A business domain object)
  ----- Custom Models -----
  Book
  Color
```

```sh
? Please select the repository base class (Use arrow keys)
❯ DefaultCrudRepository (Juggler bridge)
  ----- Custom Repositories -----
```

**After**

```sh
? Please select the model base class (Use arrow keys)
❯ Entity (A persisted model with an ID)
  Model (A business domain object)
  Book
  Color
```

```sh
? Please select the repository base class (Use arrow keys)
❯ DefaultCrudRepository (Juggler bridge)
```

The `----- Custom Models -----` and ` ----- Custom Repositories -----` banners are not required for users to know that the items under it as user-created (custom). They are a distraction and add 1 more step to getting to where the user wants to reach if they decide to select a user-created class.

## Checklist

### DIFF
--- a/packages/cli/generators/controller/index.js
+++ b/packages/cli/generators/controller/index.js
@@ -118,13 +118,15 @@ module.exports = class ControllerGenerator extends ArtifactGenerator {
     let modelList, repositoryList;
 
     try {
-      modelList = await utils.getArtifactList(
+      modelList = await utils.getArtifactListWithPath(
         this.artifactInfo.modelDir,
+        this.artifactInfo.relPath,
         'model',
       );
 
-      repositoryList = await utils.getArtifactList(
+      repositoryList = await utils.getArtifactListWithPath(
         this.artifactInfo.repositoryDir,
+        this.artifactInfo.relPath,
         'repository',
         true,
       );
--- a/packages/cli/generators/model/index.js
+++ b/packages/cli/generators/model/index.js
@@ -27,11 +27,10 @@ const ERROR_NO_MODELS_FOUND = g.f('Model was not found in');
 const BASE_MODELS = ['Entity', 'Model'];
 const CLI_BASE_MODELS = [
   {
-    name: `Entity ${chalk.gray('(A persisted model with an ID)')}`,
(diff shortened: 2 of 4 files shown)


bajtos:
I like the current behavior, but I can also see how the proposal can be seen as an improvement for many. I am fine either way 🤷🏻

bajtos:
I remember that we have many places in our documentation that are showing CLI output. Please search the docs and update all places to match the new prompts.

agnes512:
I like the current behavior too. It shows which ones are provided by LB4, but I don't have strong opinions. 

> Please search the docs and update all places to match the new prompts.

+1

bajtos:
As an alternative to the banner, can we perhaps show the path to the file where a custom model is defined? We can also change the description for built-in models to make it more obvious that they are coming from the framework. We can also add some colors to descriptions (e.g. print the path in grey; print "built-in" in grey too, or maybe in green, whatever).

```
? Please select the model base class (Use arrow keys)
❯ Entity (a persisted model with an ID - built-in)
  Model (a business domain object - built-in)
  Book (src/models/book.model.ts)
  Color (src/models/color.model.ts)
```

hacksparrow:
I have updated the PR, this is how it looks like now.

image

The grey color does the job of separating the context and de-emphasizing, so we don't need the brackets. Let's not add "- built-in)", the absence of the path should be interpreted as the class being built-in.

CLI should have minimal info and and UI should be as distraction free as possible.

hacksparrow:
The same UI change is applied to controller generator too.

image

image

It is actually good to be able to see which file we are dealing with.

hacksparrow:
Just noticed a side-effect in the controller files. It can be fixed, but more important is to have tests which would catch such side effects.

hacksparrow:

PR #6132, opened 2020-08-17
fix: appropriate type prompt for lb4 extension command

hacksparrow:
Use "Extension" instead of "Project" with `lb4 extension`.

**Before**

```sh
$ lb4 extension search
? Project description: search
? Project root directory: search
```

**After**

```sh
$ lb4 extension search
? Extension description: search
? Extension root directory: search
```

"Project" is used for creating apps. Using it also with extensions makes the user wonder if they are creating another app within the current app. Even if the command is run outside the app dir, it should make itself obvious that it is creating an extension. 

For users who have frequently used `lb4 app`, "Project" is app. Running `lb4 extension` and seeing "Project" is very confusing.

## Checklist

### DIFF
--- a/packages/cli/generators/extension/index.js
+++ b/packages/cli/generators/extension/index.js
@@ -33,12 +33,12 @@ module.exports = class ExtensionGenerator extends ProjectGenerator {
 
   promptProjectName() {
     if (this.shouldExit()) return;
-    return super.promptProjectName();
+    return super.promptProjectName('Extension');
   }
 
   promptProjectDir() {
     if (this.shouldExit()) return;
-    return super.promptProjectDir();
+    return super.promptProjectDir('Extension');
   }
 
   promptComponent() {
--- a/packages/cli/lib/project-generator.js
+++ b/packages/cli/lib/project-generator.js
@@ -131,13 +131,19 @@ module.exports = class ProjectGenerator extends BaseGenerator {
     });
   }
 
-  promptProjectName() {
+  promptProjectName(projectType) {
     if (this.shouldExit()) return false;
+    let namePrompt = 'Project name:';
+    let descriptionPrompt = 'Project description:';
+    if (projectType) {
+      namePrompt = `${projectType} name:`;
+      descriptionPrompt = `${projectType} description:`;
+    }
     const prompts = [
       {
         type: 'input',
         name: 'name',
-        message: g.f('Project name:'),
+        message: g.f(namePrompt),
         when: this.projectInfo.name == null,
         default:
           this.options.name || utils.toFileName(path.basename(process.cwd())),
@@ -146,7 +152,7 @@ module.exports = class ProjectGenerator extends BaseGenerator {
       {
         type: 'input',
         name: 'description',
-        message: g.f('Project description:'),
+        message: g.f(descriptionPrompt),
         when: this.projectInfo.description == null,
         default: this.options.name || this.appname,
       },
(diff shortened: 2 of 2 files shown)


packages/cli/generators/extension/index.js
```diff
@@ -33,12 +33,12 @@ module.exports = class ExtensionGenerator extends ProjectGenerator {
 
   promptProjectName() {
     if (this.shouldExit()) return;
-    return super.promptProjectName();
+    return super.promptProjectName('Extension');
```
raymondfeng, line 36: Let's keep `project`, maybe `Extension project`.

packages/cli/generators/extension/index.js
```diff
@@ -33,12 +33,12 @@ module.exports = class ExtensionGenerator extends ProjectGenerator {
...
   }
 
   promptProjectDir() {
     if (this.shouldExit()) return;
-    return super.promptProjectDir();
+    return super.promptProjectDir('Extension');
```
raymondfeng, line 41: `Extension project`?

packages/cli/lib/project-generator.js
```diff
@@ -131,13 +131,19 @@ module.exports = class ProjectGenerator extends BaseGenerator {
     });
   }
 
-  promptProjectName() {
+  promptProjectName(projectType) {
```
raymondfeng, line 134: Can be simplified as `projectType = 'Project'`.

packages/cli/lib/project-generator.js
```diff
@@ -157,13 +163,17 @@ module.exports = class ProjectGenerator extends BaseGenerator {
     });
   }
 
-  promptProjectDir() {
+  promptProjectDir(projectType) {
```
raymondfeng, line 166: Ditto.

packages/cli/lib/project-generator.js
```diff
@@ -131,13 +131,19 @@ module.exports = class ProjectGenerator extends BaseGenerator {
...
+    if (projectType) {
+      namePrompt = `${projectType} name:`;
+      descriptionPrompt = `${projectType} description:`;
+    }
     const prompts = [
       {
         type: 'input',
         name: 'name',
-        message: g.f('Project name:'),
+        message: g.f(namePrompt),
```
agnes512, line 146: I don't think `g.f` works for parameters. Please globalize the string when assigning the parameters.

packages/cli/generators/extension/index.js
bajtos: +1 to keep "project".

@hacksparrow if you find it confusing that both `lb4 app` and `lb4 extension` are using the same term, then perhaps use `Application project` and `Extension project`?

packages/cli/lib/project-generator.js
bajtos: +1, `g.f` does not support dynamically created inputs (AFAIK).

The best option is to keep two sets of strings, so that they can be fully translated to other languages.

```js
g.f('Project name');
g.f('Project description');
g.f('Application project name');
// ...
```

The second best option is to pass `projectType` as a parameter to `g.f`, but that means the project type will be always printed in English - not great for people who don't understand English!

```js
namePrompt = g.f('%s name:', projectType);
descriptionPrompt = g.f('%s description:', projectType);
```

packages/cli/generators/extension/index.js
hacksparrow: @bajtos Yes, that would be better.

hacksparrow:

Issue #6136, opened 2020-08-17
Warn if BootMixin app is started without booting


hacksparrow:
```ts
export class MyApplication extends BootMixin(RestApplication) {
...
}

const app = new MyApplication();
...
await app.start();
```

This app will start successfully with no booted artifacts. Since the usage of `BootMixin()` indicates a desire for booting capabilities, starting the app without booting is very likely an accidental omission. The user should be warned about the missing booting step.

This may not be a problem for users limited to using the scaffolded files, but for someone using LoopBack in a customized manner, it can be a frustrating experience to have the app running successfully with none of the artifacts loaded.  

I made a quick attempt of resolving this by adding a `start()` method in the class returned by `BootMixin` but am hit by this error:

_Class '{ readonly options: ApplicationConfig; readonly state: string; controller: <T>(controllerCtor: Constructor<T>, nameOrOptions?: string | BindingFromClassOptions | undefined) => Binding<...>; ... 53 more ...; listenerCount: (type: string | symbol) => number; }' defines instance member property 'start', but extended class '(Anonymous class)' defines it as instance member function._

@raymondfeng @bajtos

raymondfeng:
Use the following to work around the TypeScript issue:

```ts
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
```

hacksparrow:
Thank you @raymondfeng, the PR - https://github.com/strongloop/loopback-next/pull/6141.

hacksparrow:
Closed via https://github.com/strongloop/loopback-next/pull/6141.

PR #6141, opened 2020-08-18
feat: warn if app is started without booting

hacksparrow:
Warns if app is started without booting. Ref: https://github.com/strongloop/loopback-next/issues/6136.

## Checklist

### DIFF
--- a/packages/boot/src/mixins/boot.mixin.ts
+++ b/packages/boot/src/mixins/boot.mixin.ts
@@ -71,6 +71,24 @@ export function BootMixin<T extends MixinTarget<Application>>(superClass: T) {
       );
     }
 
+    booted: boolean;
+
+    /**
+     * Override to detect and warn about starting without booting.
+     */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    public async start(): Promise<void> {
+      await super.start();
+      if (!this.booted) {
+        process.emitWarning(
+          'App started without booting. Did you forget to call ' +
+            '`await app.boot()`?',
+          'LoopBackWarning',
+        );
+      }
+    }
+
     /**
      * Convenience method to call bootstrapper.boot() by resolving bootstrapper
      */
@@ -101,6 +119,7 @@ export function BootMixin<T extends MixinTarget<Application>>(superClass: T) {
 
       // @ts-ignore
       this.setState('booted');
+      this.booted = true;
 
       /* eslint-enable @typescript-eslint/ban-ts-comment */
     }
--- a/packages/boot/src/__tests__/unit/bootstrapper.unit.ts
+++ b/packages/boot/src/__tests__/unit/bootstrapper.unit.ts
@@ -5,7 +5,7 @@
 
 import {Application} from '@loopback/core';
 import {RepositoryMixin} from '@loopback/repository';
-import {expect} from '@loopback/testlab';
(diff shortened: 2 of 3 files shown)


packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -71,6 +71,23 @@ export function BootMixin<T extends MixinTarget<Application>>(superClass: T) {
...
+
+    /**
+     * Override to detect and warn about starting without booting.
+     */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    public async start(): Promise<void> {
+      await super.start();
+      if (!this.booted) {
+        console.warn(
```
raymondfeng, line 84: I wish we have a better way for warnings. `console.warn` is not very flexible.

raymondfeng: Let's use `process.emitWarning()` - https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor

hacksparrow: Wouldn't the warning be lost if they are not being actively listened for? If it requires an additional step for the developer to be able to be warned of problems, it is not en effective solution. The `console.log()` method is also used by [datasource](https://github.com/strongloop/loopback-next/blob/master/packages/boot/src/booters/datasource.booter.ts#L56) and [repository](https://github.com/strongloop/loopback-next/blob/master/packages/boot/src/booters/repository.booter.ts#L52) booters.

If `console.log()` is not flexible enough, we should be come up with a solution for it instead of passing on the pain to the user.

raymondfeng: `process.emitWarning(...)` will print the message to `stderr` as I tested:

```
process.emitWarning('xyz');
> (node:85982) Warning: xyz
```

hacksparrow: I see, updated.

packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -71,6 +65,23 @@ export function BootMixin<T extends MixinTarget<Application>>(superClass: T) {
...
+     */
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    public async start(): Promise<void> {
+      await super.start();
+      if (!this.booted) {
+        process.emitWarning(
+          'App started without booting. Did you forget to call ' +
+            '`await app.boot()`?',
+        );
```
raymondfeng, line 81: Let's pass in `type` (such as 'LoopBackWarning') in `options` so that the printout will have `LoopBackWarning` prefix.

packages/boot/src/mixins/boot.mixin.ts
```diff
@@ -24,10 +22,6 @@ import {Bootstrapper} from '../bootstrapper';
 import {BootBindings, BootTags} from '../keys';
 import {Bootable, Booter, BootOptions, InstanceWithBooters} from '../types';
 
-// FIXME(rfeng): Workaround for https://github.com/microsoft/rushstack/pull/1867
-/* eslint-disable @typescript-eslint/no-unused-vars */
-import * as loopbackContext from '@loopback/core';
-import * as loopbackCore from '@loopback/core';
```
raymondfeng, line 30: Please revert this change.

